PR #739, opened 2016-07-05
Enable windows to register to get notified of pointer button events

deklov:
Tested the keyboard related changes by running swaylock and making sure I could unlock the screen. (As far as I can tell, swaylock seems to be the only client that handles keyboard events.)

### DIFF
--- a/include/client/window.h
+++ b/include/client/window.h
@@ -12,19 +12,19 @@
 struct window;
 
 struct buffer {
-        struct wl_buffer *buffer;
-        cairo_surface_t *surface;
-        cairo_t *cairo;
-        PangoContext *pango;
-        uint32_t width, height;
-        bool busy;
+	struct wl_buffer *buffer;
+	cairo_surface_t *surface;
+	cairo_t *cairo;
+	PangoContext *pango;
+	uint32_t width, height;
+	bool busy;
 };
 
 struct cursor {
-        struct wl_surface *surface;
-        struct wl_cursor_theme *cursor_theme;
-        struct wl_cursor *cursor;
-        struct wl_poitner *pointer;
+	struct wl_surface *surface;
+	struct wl_cursor_theme *cursor_theme;
+	struct wl_cursor *cursor;
+	struct wl_poitner *pointer;
 };
 
 struct pointer_input {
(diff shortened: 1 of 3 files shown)


include/client/window.h
```diff
@@ -25,6 +27,31 @@ struct cursor {
         struct wl_poitner *pointer;
 };
 
+struct xkb {
+        struct xkb_state *state;
+        struct xkb_context *context;
+        struct xkb_keymap *keymap;
+        xkb_mod_mask_t masks[MASK_LAST];
+};
```
SirCmpwn: .

SirCmpwn:
I don't really like the change of ownership over the keyboard. The window doesn't have ownership over it, the registry does. Same for pointers afaik.

deklov:
Here are a few reasons why having windows "own" the mouse and the keyboard

1. Before my change the mouse was "owned" by the window so someone probably thought this was a good idea for the mouse. This is not a very strong argument as this might be coincidental.
2. Is the registry really "owning" the keyboard and mouse? Having a registry is standard pattern and does not necessarily imply ownership, the registry is just a place where you can find stuff which might be owned by other entities.
3. The way I see it is not about ownership of the input devices but about how manages the state required to use them (e.g., last_code, last_x and last_y). This state is passed in as the auxiliary data to wl_pointer_add_listener and wl_keyboard_add_listener. The way I understand Sway's client support code, each client has one instance of struct registry. If a client has multiple windows, it seems reasonable that each window maintain their mouse/keyboard state. It could even be a bug if they don't.

I have more to say about this but not enough time to write it up. Please correct me if I'm wrong about any of this

SirCmpwn:
You're making the incorrect assumption that one wayland client == one wayland window. The registry is something the client owns, and the client can spawn many windows. The client is only given one keyboard via the registry. We could open several connections and get the registry several times as several clients but that's ineffecient and not the Wayland way.

deklov:
I'm sorry, I'm saying that: 1) one client can have many windows, 2) we need to maintain input state for each window (e.g., last_code) and 3) each client has a single instance of the registry. If these three points are correct it implies that a multi window client needs more than one instance of the input state.

SirCmpwn:
I think that implementing that behavior in the shared code misleads its users about the nature of the keyboard on wayland. It makes sense to have the pointer state live on the window since we get coordinates relative to the surface for that.

deklov:
Why are the two different?

SirCmpwn:
Because they are? That's how the API is written. Keypresses don't matter relative to a window's coordinates, but mouse movements do.

deklov:
Fair enough.  If I drop 06b63fe will you take the rest of the changes?

SirCmpwn:
Yes. And sync with master and fix the formatting issues.

SirCmpwn:
Thanks!

PR #740, opened 2016-07-05
get_tree json fixes

zandrmartin:
- null pointer check against container names
- use actual JSON null values instead of "null" strings
- add "visible" boolean property to all containers
- remove "nodes", "floating_nodes" and "unmanaged_nodes" arrays from views as they cannot have children


SirCmpwn:
Thanks!

PR #743, opened 2016-07-06
Set panels/backgrounds' geometries correctly and don't render them ex…

deklov:
…plicitly

Panels and backgrounds were explicitly rendered by calling wlc_surface_render
in handle_output_pre_render. Calling wlc_surface_render does not set the
surface's geometry (like wlc_view_set_geometry does). Sway did not call
wlc_view_set_geometry for panels/backgrounds, so wlc defaulted their geometry
to be at the origin. This is not correct for bars unless their location is top.

Furthermore, for a surface to receive pointer events, its mask has to be
set to visible. This causes wlc to render these surfaces, causing panels
and backgrounds to be rendered twice.

This commit makes panels and surfaces visible, sets the correct geometries
and removes the code that explicitly rendered them.

### DIFF
--- a/include/bar/bar.h
+++ b/include/bar/bar.h
@@ -32,6 +32,9 @@ struct workspace {
 	bool urgent;
 };
 
+/** Global bar state */
+extern struct bar swaybar;
+
 /**
  * Setup bar.
  */
--- a/include/extensions.h
+++ b/include/extensions.h
@@ -11,8 +11,6 @@ struct background_config {
 	wlc_resource surface;
 	// we need the wl_resource of the surface in the destructor
 	struct wl_resource *wl_surface_res;
-	// used to determine if client is a background
-	struct wl_client *client;
 };
 
 struct panel_config {
--- a/include/extensions.h
+++ b/include/extensions.h
@@ -23,6 +23,8 @@ struct panel_config {
 	enum desktop_shell_panel_position panel_position;
 	// used to determine if client is a panel
 	struct wl_client *client;
+	// wlc handle for this panel's surface, not set until panel is created
+	wlc_handle handle;
 };
 
 struct desktop_shell_state {
--- a/sway/handlers.c
+++ b/sway/handlers.c
@@ -33,6 +33,66 @@
 // Event handled by sway and should not be sent to client
 #define EVENT_HANDLED true
 
+static struct panel_config *if_panel_find_config(struct wl_client *client) {
+	int i;
+	for (i = 0; i < desktop_shell.panels->length; i++) {
+		struct panel_config *config = desktop_shell.panels->items[i];
+		if (config->client == client) {
+			return config;
+		}
+	}
+	return NULL;
+}
+
+static struct wlc_geometry compute_panel_geometry(struct panel_config *config) {
+	const struct wlc_size resolution = *wlc_output_get_resolution(config->output);
+	const struct wlc_geometry *old = wlc_view_get_geometry(config->handle);
+	struct wlc_geometry new;
+
+	switch (config->panel_position) {
+	case DESKTOP_SHELL_PANEL_POSITION_TOP:
+		new.origin.x = 0;
+		new.origin.y = 0;
+		new.size.w = resolution.w;
+		new.size.h = old->size.h;
+		break;
+	case DESKTOP_SHELL_PANEL_POSITION_BOTTOM:
+		new.origin.x = 0;
+		new.origin.y = resolution.h - old->size.h;
+		new.size.w = resolution.w;
+		new.size.h = old->size.h;
+		break;
+	case DESKTOP_SHELL_PANEL_POSITION_LEFT:
+		new.origin.x = 0;
+		new.origin.y = 0;
+		new.size.w = old->size.w;
+		new.size.h = resolution.h;
+		break;
+	case DESKTOP_SHELL_PANEL_POSITION_RIGHT:
+		new.origin.x = resolution.w - old->size.w;
+		new.origin.y = 0;
+		new.size.w = old->size.w;
+		new.size.h = resolution.h;
+		break;
+	}
+
+	return new;
+}
+
+static void update_panel_geometry(struct panel_config *config) {
+	struct wlc_geometry geometry = compute_panel_geometry(config);
+	wlc_view_set_geometry(config->handle, 0, &geometry);
+}
+
+static void update_panel_geometries(wlc_handle output) {
+	for (int i = 0; i < desktop_shell.panels->length; i++) {
+		struct panel_config *config = desktop_shell.panels->items[i];
+		if (config->output == output) {
+			update_panel_geometry(config);
+		}
+	}
+}
+
 /* Handles */
 
 static bool handle_input_created(struct libinput_device *device) {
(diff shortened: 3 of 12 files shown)


deklov:
Mouse pointer looks like the X default and mouse events are received by the bar.

sway/handlers.c
```diff
@@ -109,42 +109,6 @@ static void handle_output_destroyed(wlc_handle output) {
...
-				break;
-			case DESKTOP_SHELL_PANEL_POSITION_RIGHT:
-				geo.origin = (struct wlc_point){ resolution.w - size.w, 0 };
-				break;
-			}
-			wlc_surface_render(config->surface, &geo);
-			break;
-		}
-	}
 }
```
SirCmpwn: Don't leave this empty, if we don't use it anymore just remove it and unregister the handler.

SirCmpwn:
How well does this work with:

* Output resizing (either via `swaymsg output ... res ...` or the x backend resizing the windows)
* Z ordering when the layout is arranged, the bg should always be at the bottom and the bar should be on top of all tiled windows but below floating and fullscreened windows

SirCmpwn:
Bump

deklov:
Not very well unfortunately. I got so excited when the bar started getting mouse events that I forgot to test re-sizing and swaybg. I'll try to work on this some more maybe tonight or tomorrow.

SirCmpwn:
No worries, take your time.

deklov:
Think I got this working but before you merge to master could you please test it. I wanted to focus on the bar first so I reverted some of my code related to background. I will come back to this later but just in case I don't, I don't want to leave half-done, non-working behind.

deklov:
I forgot to commit something..

deklov:
btw. I have not tested this on dual head since I only have one monitor at home.

deklov:
One more thing. I didn't to anything to try to set the z ordering. Fullscreen does work, but that might be coincidental. Do we need to set it explicitly?

SirCmpwn:
Not sure how it'd shake out over time, restarts, etc.

deklov:
[handlers.c:521] Command 'restart' failed: Unknown/invalid command

Get this before any of my changes as well.

SirCmpwn:
Sorry, I meant reload. Which causes the swaybg and swaybar processes to restart.

deklov:
Reload seems to work just fine.

deklov:
This last commits require the previous commits to be tested and I was to lazy to create a branch with only these commits and too lazy to figure out how to create a PR with only these new commits. Hope you don't mind that I amended them to this PR.

swaybar/ipc.c
```diff
@@ -7,6 +7,10 @@
 #include "bar/config.h"
 #include "bar/ipc.h"
 
+void ipc_send_workspace_command(const char *workspace_name) {
+	sway_log(L_DEBUG, "Clicked on window %s", workspace_name);
```
deklov: This log message should be: "Clicked on workspace button %s".

include/bar/bar.h
```diff
@@ -32,6 +32,9 @@ struct workspace {
 	bool urgent;
 };
 
+/** Global bar state */
+extern struct bar swaybar;
```
deklov: No need to have a global struct config since swaybar which points to it was already global.

swaybar/render.c
```diff
@@ -172,7 +172,7 @@ static void render_block(struct window *window, struct config *config, struct st
 
 }
 
-static char *handle_workspace_number(bool strip_num, const char *ws_name) {
+static const char *strip_workspace_name(bool strip_num, const char *ws_name) {
```
deklov: Don't need for the in strip_num argument since all call sites pass in swaybar.config->strip_workspace_numbers which is available here.

deklov:
Last piece of the puzzle. You can now change workspace by clicking on the bar.

swaybar/ipc.c
```diff
@@ -8,7 +8,17 @@
...
+	size_t size = strlen("workspace ") + strlen(workspace_name) + 1;
+
+	char *command = malloc(size);
+	if (!sway_assert(command != NULL, "Failed to allocate memory for workspace command")) {
+		return;
+	}
+	snprintf(command, size, "workspace %s", workspace_name);
+
+	uint32_t reply_size;
+	ipc_single_command(swaybar.ipc_socketfd, IPC_COMMAND, command, &reply_size);
```
deklov: Oops, memory leak forgot to free command.

deklov: And another bug, the last argument to ipc_single_command is both input and output, even though the comment says:

`
/**             
 * Issues a single IPC command and returns the buffer. len will be updated with
 * the length of the buffer returned from sway.
 */ 
`

deklov:
Fixed the the memory leak and misuse of ipc_single_command.

SirCmpwn:
How about changing workspaces with the scroll wheel?

deklov:
I look into that once this PR has been merged. Can you please open a ticket with all the mouse related features you want to see for swaybar and assign it to me.

SirCmpwn:
Sure.

include/extensions.h
```diff
@@ -25,6 +23,8 @@ struct panel_config {
 	enum desktop_shell_panel_position panel_position;
 	// used to determine if client is a panel
 	struct wl_client *client;
+        // wlc handle for this panel's surface, not set until panel is created
+        wlc_handle handle;
```
SirCmpwn: tab issues

swaybar/bar.c
```diff
@@ -58,6 +58,37 @@ struct output *new_output(const char *name) {
...
+	}
+
+	if (!sway_assert(clicked_output != NULL, "Got pointer event for non-existing output")) {
+		return;
+	}
+
+	double button_x = 0.5;
+	for (int i = 0; i < clicked_output->workspaces->length; i++) {
+		struct workspace *workspace = clicked_output->workspaces->items[i];
+		int button_widht, button_height;
```
SirCmpwn: typo

swaybar/ipc.c
```diff
@@ -7,6 +7,16 @@
 #include "bar/config.h"
 #include "bar/ipc.h"
 
+void ipc_send_workspace_command(const char *workspace_name) {
+	sway_log(L_DEBUG, "Clicked on workspace %s", workspace_name);
```
SirCmpwn: This log thing is not necessarily true. Anything could have invoked this function, not just mouse click things.

deklov:
We know what output each panel should be on: https://github.com/SirCmpwn/sway/blob/master/sway/extensions.c#L88. Could it be as simple as telling wlc what output to put the panels on?

SirCmpwn:
Perhaps!

deklov:
Wait a minute. When a bar connects to sway, it does not

SirCmpwn:
Can you just push more commits to this branch instead of making a fourth pull request?

deklov:
Yes, I didn't  mean to close this PR, must have done so by mistake.

deklov:
The dual head problem was easy to fix: git grep -n 'wlc.*output' in sway's source tree, found wlc_view_set_outpu and added a call to it in handle_view_created. Patch on the way. It will take longer to fix the spelling and tab errors ;)

include/extensions.h
deklov: Fixed

swaybar/bar.c
deklov: Fixed

swaybar/ipc.c
deklov: Fixed

deklov:
Bump. Can we please try and get this merged.

SirCmpwn:
Can you sync up with master?

deklov:
Done

SirCmpwn:
Thanks!

Issue #742, opened 2016-07-05
No window titles in tabbed layout


DaanDeMeyer:
Using tabbed layout I have no separate window titles when opening applications. When I open a window a black bar appears at the top of the screen containing the name of the workspace. Opening multiple windows doesn't change anything. As soon as the latest window in a workspace is removed the black bar disappears as well. Window titles do work in the default layout. 

I'm using the Arch sway package (not sway-git).

DaanDeMeyer:
Found the issue, I was using the i3 workspace_layout instead of just layout. Changing to layout in the config file fixed the issue.

DaanDeMeyer:
nvm I was wrong, layout can't be used in the config file. I found out that activating tabbed layout via the keyboard shortcut fixes the issue but enabling tabbed layout via the config file gives the above issue.

PR #744, opened 2016-07-07
Fix tabbed/stacked corner case #742

thejan2009:
Tabbed/stacked containers are now created only if a view is present on
the workspace. If a view is created on previously empty tabbed/stacked
workspace, it gets wrapped in a container.

I might have broken something with the change in container.c/swayc_tabbed_stacked_parent, as it's used in some other places, though it seems to work.


SirCmpwn:
LGTM, thanks!

PR #749, opened 2016-07-12
Enable using non-annotated tags in git-describe

Hummer12007:
This fixes the version string, as 0.7+ tags are lightweight (see `git-describe(1)`).


SirCmpwn:
Thanks!

Issue #725, opened 2016-06-27
swaylock behaves differently than i3lock


tasn:
Hey,

The swaylock process exits after it has been unlocked, i3lock exits once the lock screen is up. This makes i3lock much easier to use, because you can do things like: "i3lock && suspend" and use it as a key binding, while with swaylock it is not possible.

Reproduce:
swaylock && date
// Unlock
i3lock && date
// Unlock

See that in i3lock the date is much earlier (time of lock compared to time of unlock).

Thanks.

sleep-walker:
It behaves the same was as slock so it fits me. You could do something like
```
swaylock & sleep 1; date
```

SirCmpwn:
I don't know if I really appreciate this behavior of i3lock. I would merge a PR that adds -f (fork) to swaylock, though.

tasn:
That's a hack, no the same. What if locking takes 10 seconds due to system
load? Or fails?

sleep-walker:
It's hack, I agree. Clean solution is the one @SirCmpwn already proposed. Nice to see using sway ;)

tasn:
I've always been meaning to give i3 and wayland a proper go, so I thought I'd kill two birds with one stone. :)

Unfortunately though, I've encountered a lot of bugs, mostly with xwayland, some with sway, and one with sway/gtk3 (gnome-calculator). I haven't gotten around to reporting them all, because still unsure what I'm supposed to report as I guess everything is still WIP.

SirCmpwn:
Please report the bugs you find. As for gnome-calculator, does this fix it for you?

    for_window [id="gnome-calculator"] floating enable; border pixel

I've been thinking about distributing some default window rules with sway, anyone else have some thoughts?

tasn:
@SirCmpwn, OK, will try to get around to reporting them.

As for gnome-calculator, yes, making it float fixes it. I wrote the tiling module for www.enlightenment.org so I'm all too familiar with these issues. When I wrote it, I only had X in mind, not sure about all of the Wayland equivalents, but I'm sure there are.

[...]

PR #750, opened 2016-07-13
swaylock: Add -f option to detach from the controlling terminal, fixe…

deklov:
…s #725


swaylock/main.c
```diff
@@ -421,6 +424,9 @@ int main(int argc, char **argv) {
 		case 'u':
 			show_indicator = false;
 			break;
+                case 'f':
```
SirCmpwn: style issue here

PR #752, opened 2016-07-14
Change workspace with mouse wheel

deklov:


swaybar/bar.c
```diff
@@ -89,6 +89,13 @@ static void mouse_button_notify(struct window *window, int x, int y, uint32_t bu
 	}
 }
 
+static void mouse_scroll_notify(struct window *window, enum scroll_direction direction) {
+	sway_log(L_DEBUG, "Mouse wheel scrolled %s", direction == SCROLL_UP ? "up" : "down");
+
+	const char *workspace_name = direction == SCROLL_UP ? "prev" : "next";
```
deklov: Not sure what the correct behavior is here: prev vs. prev_on_output.

include/client/window.h
```diff
@@ -27,11 +27,18 @@ struct cursor {
...
+	SCROLL_DOWN
+};
+
 struct pointer_input {
 	int last_x;
 	int last_y;
 
 	void (*notify_button)(struct window *window, int x, int y, uint32_t button);
+
+	void (*notify_scroll)(struct window *window, enum scroll_direction direction);
```
deklov: I guess we could pass time as well but I didn't needed it..

wayland/window.c
```diff
@@ -47,7 +47,17 @@ static void pointer_handle_button(void *data, struct wl_pointer *pointer, uint32
 }
 
 static void pointer_handle_axis(void *data, struct wl_pointer *pointer,
-		    uint32_t time, uint32_t axis, wl_fixed_t value) {
+				uint32_t time, uint32_t axis, wl_fixed_t value) {
```
deklov: Not exactly sure what axis and value. Documentation is lacking. Testing suggests that on a mouse with a standard scroll wheel axis is zero and value is -10.0/10.0 when scrolled up/down.

deklov:
This is more of a  RFC pull request.

swaybar/bar.c
deklov: Btw. One annoying thing with prev/next is that when changing to a workspace on a different output the mouse pointer follows.

Hummer12007:
@deklov there's horizontal scrolling too (axis == 0 -> vertical, 1 -> horizontal)

swaybar/bar.c
SirCmpwn: The correct behavior is prev_on_output.

wayland/window.c
```diff
@@ -41,13 +41,23 @@ static void pointer_handle_button(void *data, struct wl_pointer *pointer, uint32
...
+		window->pointer_input.notify_button(window, input->last_x, input->last_y, button);
 	}
 }
 
 static void pointer_handle_axis(void *data, struct wl_pointer *pointer,
-		    uint32_t time, uint32_t axis, wl_fixed_t value) {
+				uint32_t time, uint32_t axis, wl_fixed_t value) {
+	struct window *window = data;
+
+	if (!sway_assert(!axis, "Only support horizontal axis")) {
```
SirCmpwn: Assertions are to detect things that should never happen. Horizontal scrolling doesn't qualify.

SirCmpwn:
LGTM

wayland/window.c
```diff
@@ -47,7 +47,26 @@ static void pointer_handle_button(void *data, struct wl_pointer *pointer, uint32
...
+		direction = wl_fixed_to_double(value) < 0 ? SCROLL_UP : SCROLL_DOWN;
+		break;
+	case 1:
+		direction = wl_fixed_to_double(value) < 0 ? SCROLL_LEFT : SCROLL_RIGHT;
+		break;
+	default:
+		if (!sway_assert(false, "Only support horizontal axis")) {
+			return;
+		}
+	}
```
deklov: @Hummer12007 something like this?

Hummer12007: Yeah, but why should you add a default clause? The axis enum in libinput/wayland has only two directions.

SirCmpwn: I don't have a strong preference for or against a default here, but the log message needs to be updated.

SirCmpwn:
Thanks!

PR #754, opened 2016-07-16
Clarify which tabbed/stacked container we want (fixes #751)

### LINKED ISSUE #751: Third level of window hierarchy crashes sway
Sway crashes on my machine when I try to introduce a third level of window hierarchy. The steps I need to reproduce are:

1. Open a terminal with `$mod + d`
2. Open a second terminal
3. Switch to tabbed mode with `$mod + w`
4. Switch back to the first one and hit `$mod + b` to create the second level.
5. Open a third terminal
6. Again, switch back to the first terminal and hit `$mod +b` or `$mod + v` to create the third level of hierarchy
7. Sway crashes

I haven't tested all the edge cases, probably sway crashes for other applications as well (I personally can confirm Geany). Also, I have not tried all possible combinations of vertical and horizontal splitting and/or splitting on the first or second opened application, but I can confirm that the problem is reproducable with the steps outlined above. If you're interested in some specific tests, please let me know.

[`sway -d &> sway.log`](https://gist.github.com/Ferruck/9ffced70da84e90c1c9d0516e69dfca6)

thejan2009:
Refactored swayc_tabbed_stacked_parent into 2 functions - one returns the root tabbed/stacked ancestor (which was how the original function worked), the other return just the immediate tabbed/stacked parent.

### DIFF
--- a/include/container.h
+++ b/include/container.h
@@ -262,6 +262,12 @@ bool swayc_is_empty_workspace(swayc_t *container);
  * Returns the top most tabbed or stacked parent container. Returns NULL if
  * view is not in a tabbed/stacked layout.
  */
+swayc_t *swayc_tabbed_stacked_ancestor(swayc_t *view);
+
+/**
+ * Returns the immediate tabbed or stacked parent container. Returns NULL if
+ * view is not directly in a tabbed/stacked layout.
+ */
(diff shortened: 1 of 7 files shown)


SirCmpwn:
Thanks!

PR #764, opened 2016-07-18
Do not crash when handling the command 'move position'

deklov:


SirCmpwn:
What was wrong here, and how does this help? Reading the code makes it look like it breaks move position mouse.

deklov:
Sway segfaults if you send it 'move position' because it tries to access argv[1] when argc == 1.

PR #762, opened 2016-07-17
Add support to move command to move floating containers

deklov:
This is an RFC review, not sure if I did this the right way.

PR #765, opened 2016-07-18
Refactor cmd_move

deklov:
The idea is to use existing cmd_handlers infrastructure to make parsing arguments more consistent and therefore easier to read, change and extend. I have tested each and every move command. (Some automated tests would have been nice.)


sway/commands.c
```diff
@@ -35,6 +35,8 @@
 #include "input.h"
 #include "border.h"
 
+#define ARRAY_SIZE(array) (sizeof(array) / sizeof(array[0]))
```
SirCmpwn: :-1:

sway/commands.c
```diff
@@ -1003,145 +1010,226 @@ static struct cmd_results *cmd_mouse_warping(int argc, char **argv) {
...
+		return cmd_results_new(CMD_INVALID, "move", cmd_move_expected_syntax);
+	}
+        /* Skip the 'to' part of the command */
+        argc--;
+        argv++;
+
+	static const struct cmd_handler handlers[] = {
+            {"output", cmd_move_container_to_output},
+            {"workspace", cmd_move_container_to_workspace},
+        };
```
SirCmpwn: indentation is screwed up here

SirCmpwn: And in several other places

sway/commands.c
deklov: I know you are not a fan of macros but I was hoping you would make an exception here on the basis that this macro improves code quality, is one of the most commonly used macros and that it cannot be implemented as a function. I'm not going to argue with religion though so if you still want me to get rid of it just give it another thumbs down.

SirCmpwn: I'm familiar with this macro and its uses, and I've seen it elsewhere. :-1:

deklov: How about this: Change all definitions of struct cmd_handler arrays to:

```
static const size_t FOO_HANLDERS_SIZE = XXX;
static const struct cmd_handler foo_handlers[FOO_HANLDERS_SIZE] = {
...
};
```

This way we can replace ARRAY_SIZE with FOO_HANLDERS_SIZE and we don't have to duplicate the expansion of ARRAY_SIZE all over the place.

sway/commands.c
deklov: Apparently clang-format is not enough. Guess I'll create a local vimrc which I probably should have done long time age.

sway/commands.c
SirCmpwn: :-1:

deklov: Do you have any other suggestions or do you want me expand the ARRAY_SIZE macro everywhere? If you don't mind what's wrong with having the size constants?

SirCmpwn: Expand the macro everywhere. The size constant would have to be manually updated when we add new commands.

deklov:
Please don't merge this just yet, I'd like to do one more round of testing once we're happy with this PR.

SirCmpwn:
What's the status of this, and #762?

deklov:
I was waiting for you to sign off on this one so that I can do a second round of manual testing before merging it. #762 is pretty much ready to go but I think it might be better if we get this one merged first and then I can rebase  #762 on top of this one before merging.

SirCmpwn:
Alright. Can you resolve the merge conflicts and get this up to date with master? I'll review+test this tomorrow and we can go from there.

toogley:
what's the current status of this?

Issue #766, opened 2016-07-18
Container titles - edge cases


thejan2009:
Sometimes, view title is not the right thing to use in container title. In i3wm, the X equivalent of app_id is used instead of title for tabbed/stacked layouts that have more complex (split/tabbed/stacked) child layouts.

What should be "Sway: V[urxvtc urxvtc]" is instead "Sway: V[js@h:~ js@h:~]". With a bit more complex layout and longer view titles, window title becomes completely unreadable.

PR #767, opened 2016-07-18
Construct titles of nested tabbed/stacked containers from app_id inst…

thejan2009:
…ead of title

As per  #766 - it turned out to be pretty easy to fix.


SirCmpwn:
Thanks!

PR #777, opened 2016-07-21
Arrange backgrounds only when required

Hummer12007:
More invocations of `arrange_backgrounds()` to be inserted, if/when the need arises.


SirCmpwn:
When you move between stacked or tabbed views, does the background occlude the views?

Hummer12007:
It doesn't

PR #780, opened 2016-07-22
fix regression in hide_edge_borders

zandrmartin:
origin x & y are not needed because this is the only child in the workspace. if swaybar location is top, view's y is not zero, so y + view's height > workspace height.


thejan2009:
This sometimes breaks hiding right and bottom borders (when there is more than one child in the workspace).

Issue #678, opened 2016-05-29
Bar invisible on restart until 2nd window opened


tuncer:
If you restart while only one window exists, the bar will stay invisible until you open a 2nd window.

SirCmpwn:
I suspect WLC issues here.

SirCmpwn:
Possibly related: swaylock issues on multiple outputs.

tilpner:
On `reload`, the bar only shows if there are windows with `border normal` for me. Maybe this is related?

PR #782, opened 2016-07-22
Fix swaybar take 2

zandrmartin:
A less CPU-intensive ;) fix for #678. Since Swaybar renders itself, Sway doesn't know when that has happened. This PR adds a signal Swaybar can send after it has rendered itself so Sway knows to adjust the layout.

### DIFF
--- a/include/bar/ipc.h
+++ b/include/bar/ipc.h
@@ -13,6 +13,10 @@ void ipc_bar_init(struct bar *bar, const char *bar_id);
  */
 bool handle_ipc_event(struct bar *bar);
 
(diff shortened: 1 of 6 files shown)


mikkeloscar:
I don't think this is the right approach. wlc should know that the swaybar (client) exists and sway should be able to use this info without getting a signal from swaybar.

zandrmartin:
@Hummer12007 has pointed out this at least needs improvements before merged, and I'm sure @mikkeloscar is correct in that there is most likely a better approach.

PR #785, opened 2016-07-23
don't passthrough a floating_scroll event to focused window

### LINKED ISSUE #784: floating_scroll passes scroll events to the focused window
Example configuration:

> floating_scroll up gaps inner workspace plus 1
> floating_scroll down gaps inner workspace minus 1

When floating modifier is pressed, only the defined actions should execute. In this example, inner gaps change, but the focused window also scrolls the content.

thejan2009:
Fixes #784. Tested with xev.

### DIFF
--- a/sway/handlers.c
+++ b/sway/handlers.c
@@ -876,15 +876,21 @@ bool handle_pointer_scroll(wlc_handle view, uint32_t time, const struct wlc_modi
 	if (!(modifiers->mods ^ config->floating_mod)) {
 		int x_amount = (int)_amount[0];
 		int y_amount = (int)_amount[1];
-		if (x_amount > 0) {
+
+		if (x_amount > 0 && strcmp(config->floating_scroll_up_cmd, "")) {
 			handle_command(config->floating_scroll_up_cmd);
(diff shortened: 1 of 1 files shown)


sway/handlers.c
```diff
@@ -874,17 +874,28 @@ static bool handle_pointer_button(wlc_handle view, uint32_t time, const struct w
...
 			handle_command(config->floating_scroll_right_cmd);
-		} else if (y_amount < 0) {
+			handled = true;
+		} else if (y_amount < 0 && strcmp(config->floating_scroll_left_cmd, "")) {
 			handle_command(config->floating_scroll_left_cmd);
+			handled = true;
+		}
+
+		if (handled) {
+			return EVENT_HANDLED;
```
SirCmpwn: Why not just `return EVENT_HANDLED;` on each line you set handled = true on, then eliminate the variable and tighten up the performance a tad?

thejan2009: I wasn't sure how sideway scrolling would register (if both vertical and horizontal components would be != 0).

PR #789, opened 2016-07-28
bugfix: cmd focus parent don't set focus above ws

acrisci:
In the `focus parent` command, do not set focus above the workspace level.
These containers are not meant to be focused.

This prevents a crash on repeated `focus parent` commands.


SirCmpwn:
Thanks!

acrisci:
Oops my commit message wasn't good. I implemented it for all movement not just focus parent.

SirCmpwn:
It's cool.

acrisci:
And I introduced a nasty bug I'm fixing now

PR #790, opened 2016-07-28
bugfix: set focus to workspace of an output

acrisci:
In `move_focus()`, when given an output, set the focus to the workspace of that
output instead of the output itself.

This fixes a bug that did not allow users to switch between outputs introduced
in afc6ad6.

It also fixes other issues before that commit when a workspace with children
was selected and the user tried to switch focus in the direction of another
output.

### DIFF
--- a/sway/focus.c
+++ b/sway/focus.c
@@ -67,9 +67,11 @@ bool move_focus(enum movement_direction direction) {
 	swayc_t *new_view = get_swayc_in_direction(old_view, direction);
 	if (!new_view) {
 		return false;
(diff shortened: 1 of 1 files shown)


acrisci:
Here's what didn't work before:

Suppose you are on the right monitor.

1. Go to a workspace and open up two windows side by side.
2. Focus the right window
3. `focus parent, focus left`

It should go to the workspace on the left output, but it doesn't.

SirCmpwn:
Thanks again!

PR #793, opened 2016-07-28
bugfix: invisible windows on empty workspace with tabbed/stacked layout

thejan2009:
This commit just fixes the bug that sometimes made initial workspace_layout unusable.

Sway is actually quite different in how workspace_layout functions, compared to i3. The i3 behavior isn't exactly logical, so I raised an issue in https://github.com/i3/i3/issues/2402.  When I get some response over in the i3 repo, I'll try to match its correct functionality.

### DIFF
--- a/sway/layout.c
+++ b/sway/layout.c
@@ -799,7 +799,9 @@ static void arrange_windows_r(swayc_t *container, double width, double height) {
 		}
 
 		// update container size if it's a direct child in a tabbed/stacked layout
-		if (swayc_tabbed_stacked_parent(container) != NULL) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Do you want to wait for the i3 discussion to shake out before this is merged (so you can update this PR) or do you want to make a seperate PR if any fixes are necessary?

thejan2009:
I think this should be merged, even if it's a temporary fix. While it doesn't match i3 behaviour exactly (yet), at least workspace_layout tabbed/stacked is usable again. I'll make another PR when I get some response over at i3.

SirCmpwn:
Sounds good to me.

PR #810, opened 2016-07-31
swaymsg: pretty print ipc response

### LINKED ISSUE #809: Pretty print swaymsg json output by default
This feature was rejected from i3 because apparently michael wanted to not break programs that checked for exact json strings.

I could do it with a wrapper script, but it's nicer to have it built in.

What do you think?

acrisci:
fixes #809

### DIFF
--- a/swaymsg/CMakeLists.txt
+++ b/swaymsg/CMakeLists.txt
@@ -4,7 +4,8 @@ add_executable(swaymsg
 
 target_link_libraries(swaymsg
 	sway-common
-	)
+	${JSONC_LIBRARIES}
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Looks good, but gonna leave the PR unmerged until sway 0.9 drops.

acrisci:
You should use development/release branches.

SirCmpwn:
I could, but I still haven't fully defined the process for how this rc/release/etc stuff should be put together. I'll formalize the process before 0.10 and then we'll stick to it.

PR #811, opened 2016-07-31
Implement focus handling for containers

acrisci:
The previous implementation of focus handling assumed that only views can be
focused. Containers can also be focused with a command like `focus parent` or
`focus child`.

Change `set_focused_container()` to handle the case of the given container
being a container with children and update borders accordingly.

### DIFF
--- a/include/border.h
+++ b/include/border.h
@@ -22,7 +22,6 @@ void border_clear(struct border *border);
 void update_container_border(swayc_t *container);
 
 void render_view_borders(wlc_handle view);
-void update_view_border(swayc_t *view);
 void map_update_view_border(swayc_t *view, void *data);
 int get_font_text_height(const char *font);
 bool should_hide_top_border(swayc_t *con, double y);
--- a/include/border.h
+++ b/include/border.h
@@ -16,6 +16,11 @@ struct border {
  */
 void border_clear(struct border *border);
 
+/**
+ * Recursively update all of the borders within a container.
+ */
+void update_container_border(swayc_t *container);
+
 void render_view_borders(wlc_handle view);
 void update_view_border(swayc_t *view);
 void map_update_view_border(swayc_t *view, void *data);
--- a/sway/focus.c
+++ b/sway/focus.c
@@ -115,7 +115,7 @@ bool set_focused_container(swayc_t *c) {
 
 	// Get workspace for c, get that workspaces current focused container.
 	swayc_t *workspace = swayc_active_workspace_for(c);
-	swayc_t *focused = get_focused_view(workspace);
+	swayc_t *focused = get_focused_container(workspace);
 
 	if (swayc_is_fullscreen(focused) && focused != c) {
 		// if switching to a workspace with a fullscreen view,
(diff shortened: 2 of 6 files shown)


acrisci:
I'm getting kind of confused by the vocabulary here. Should I refer to containers and views as separate things or is a view a type of container?

On i3, everything is a container, including workspaces and outputs.

sway/focus.c
```diff
@@ -140,33 +140,32 @@ bool set_focused_container(swayc_t *c) {
...
 	// get new focused view and set focus to it.
-	p = get_focused_view(c);
-	if (p->type == C_VIEW && !(wlc_view_get_type(p->handle) & WLC_BIT_POPUP)) {
+	if (c->type == C_CONTAINER || (c->type == C_VIEW && !(wlc_view_get_type(p->handle) & WLC_BIT_POPUP))) {
 		// unactivate previous focus
 		if (focused->type == C_VIEW) {
 			wlc_view_set_state(focused->handle, WLC_BIT_ACTIVATED, false);
-			update_view_border(focused);
 		}
+		update_container_border(focused);
```
acrisci: Does this need a guard for if view focused is locked? (I'm not sure when view focus can be locked).

Also, does the order that we do things here matter much? It could be a bit simpler if not.

acrisci:
It's WIP because there are a few bugs that this introduces from other parts of the code that assume this function will automatically find the view within the container and set focus on that.

You can test by opening two windows, splitting one and adding a new window to that container, and then do `focus parent` and `focus child` to see wlc focus being set correctly and borders being updated to show the container is focused by making all the children have focused borders.

acrisci:
As soon as you OK my approach here, I'll continue working in this branch to fix all the bugs this creates and find the edge cases.

include/border.h
```diff
@@ -16,6 +16,11 @@ struct border {
  */
 void border_clear(struct border *border);
 
+/**
+ * Recursively update all of the borders within a container.
+ */
+void update_container_border(swayc_t *container);
```
acrisci: This is intended to be a replacement for `update_view_border()` I will eventually make that function private and replace all occurrences of that function with this new function.

SirCmpwn:
>I'm getting kind of confused by the vocabulary here. Should I refer to containers and views as separate things or is a view a type of container?

No, containers and views are the same thing. A view is a type (literally, swayc_t->type) of container, or rather a type of swayc_t. So are outputs and workspaces and so on.

sway/focus.c
SirCmpwn: Focus is locked during view creation sometimes to deal with an edge case, iirc. @zandrmartin would know more. I'm not sure if the order matters.

SirCmpwn:
Approach here seems fine.

include/border.h
SirCmpwn: Sounds good.

sway/focus.c
zandrmartin: I think @SirCmpwn is mixing me up with someone else, I'm not too familiar with this code. Git blame says @mikkeloscar and @taiyu-len wrote the bulk of that section.

SirCmpwn: Struggling to find the thread where this was discussed.

sway/handlers.c
```diff
@@ -415,7 +415,7 @@ static bool handle_view_created(wlc_handle handle) {
 		// we were on one workspace, switched to another to add this view,
 		// now let's return to where we were
 		workspace_switch(current_ws);
-		set_focused_container(current_ws->focused);
+		set_focused_container(get_focused_container(current_ws));
```
acrisci: Does this look ok to you @zandrmartin?

zandrmartin: Makes sense to me. I see how the old way could potentially mis-set the focus to a higher-level container than it was previously.

acrisci:
I'm ok with how this is right now. I'm sure there are more bugs, but nothing major on my end. You can review/merge this whenever.

thejan2009:
Sometimes borders don't update (render) at the exact moment, but only when the mouse is moved or some other event occurs, so it looks like the compositor is lagging very much. You might want to add somewhere some explicit command for wlc to render the borders immediately.

PR #813, opened 2016-08-02
Remove dead global locked_view_focus

acrisci:
The value of `locked_view_focus` is always false. Remove dead code associated
with this variable to simplify things.


SirCmpwn:
Thanks!

PR #814, opened 2016-08-02
Refactor logic to determine if stacked/tabbed con

acrisci:
Add a new function `swayc_is_tabbed_or_stacked()` to determine whether or not
the given container has layout of stacked or tabbed. This is done commonly
enough that it should have its own method to help readability and code quality.

No logic changes.

### DIFF
--- a/include/container.h
+++ b/include/container.h
@@ -142,6 +142,11 @@ enum visibility_mask {
 } visible;
 
 /**
+ * Whether this con is tabbed or stacked
+ */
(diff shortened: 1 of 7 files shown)


SirCmpwn:
NAK. I don't want to formalize a special case for layouts like these. At some point I'd like to see things refactored to avoid having tests for specific types of containers at all (i.e. implementations of each kind of container provide their own code for handling this and that, rather than having the shared code know about each kind of container).

PR #815, opened 2016-08-02
bugfix: focus container on output when focused

acrisci:
Focus the container on the output (not the output itself) when an output is
focused.

This is intended to fix a bug where borders are not updated correctly when
switching the vt away/back to sway.


acrisci:
I'm trying to get at a crash that happens to me sometimes when switching vts.

SirCmpwn:
FWIW vt switching is buggy as hell, which mostly comes from wlc issues. Change looks good, will merge once it's all green.

PR #822, opened 2016-08-03
Add resize mode to the default config file

mayhewj:


SirCmpwn:
Thanks!

Teolha:
Wait, there now is a mode although mode_switching is not yet there? oO

It took me quite some time (and a thread in the archlinux forums) to find out why that was not working :(

https://bbs.archlinux.org/viewtopic.php?pid=1646570

SirCmpwn:
Modes work fine.

Issue #343, opened 2015-12-17
i3bar compatability


SirCmpwn:
These features need to be implemented in swaybar:
- [x] colors
- [x] status_command
- [x] hidden_state / bindsym / modifier
- [x] output
- [x] position
- [x] id
- [x] separator_symbol
- [x] strip_workspace_numbers
- [x] height
- [x] workspace_buttons
- [x] binding_mode_indicator
- [x] mode
- [x] font
- [x] Parse JSON status (with colors etc)
- [x] Mouse events
- [x] click_events
- [ ] Tray icons
  - [x] SNI
  - [ ] Xembed

mikkeloscar:
I think font is missing as well.

SirCmpwn:
Thanks, added that.

mikkeloscar:
I don't know if this belong here, but the bar doesn't seem to get fresh bar configuration if you edit the config and reload it from within sway.

SirCmpwn:
Probably worth having a second issue for that.

StephenBrown2:
Looking forward to the day when I can use i3blocks seamlessly with swaybar...

Keep up the great work!

progandy:
I guess sway should also support a custom command for the swaybar. i3 has ```i3bar_command <command>```, so maybe ```swaybar_command <command>``` works.

SirCmpwn:
Good call.

StephenBrown2:
Rather important feature missing from the list: click events

http://i3wm.org/docs/i3bar-protocol.html#_click_events

Drakulix:
What about trayicon support? I might be difficult to do, but it is a quite major feature.

SirCmpwn:
I plan on addressing that at some point, but no Wayland compositors currently support tray icons, so we'd have to make up our own protocol extension and try to land it in the various GUI toolkits, as well as in xwayland.

progandy:
Regarding trayicons, I suggest you use the StatusNotifier / AppIndicator protocol which is promoted heavily as the new way to display status icons

SirCmpwn:
My intention is to support both.

DerpyCrabs:
Can you also add smart_hide mode when swaybar is visible only if there is any window on workspace?

SirCmpwn:
Is that a request for a sway-exclusive feature, rather than one for a feature already in i3? If so, please create a new Github issue for it.

that1communist:
[...]

PR #831, opened 2016-08-04
common: use strtoul in parse_color to avoid clamp

progandy:
Some implementations of strtol may clamp the values to LONG_MAX instead of
wrapping around to negative values, so use strtoul instead to parse colors.

(Mentioned in #343)


SirCmpwn:
Thanks!

PR #832, opened 2016-08-04
Fix 'workspace back_and_forth' and workspace_auto_back_and_forth clash

thejan2009:
When workspace_auto_back_and_forth is enabled, workspaces get switched
twice with previously mentioned command, which is not the expected
behavior.

Removes one redundant creation of previous workspace.


SirCmpwn:
Thanks!

PR #838, opened 2016-08-05
Fix switch to prev ws on handle_view_created

thejan2009:
Switching back to original workspace should be done only if workspace
was switched while appending new view.

### DIFF
--- a/sway/handlers.c
+++ b/sway/handlers.c
@@ -410,7 +410,7 @@ static bool handle_view_created(wlc_handle handle) {
 	}
 	wlc_view_set_mask(handle, VISIBLE);
 
-	if (return_to_workspace && current_ws) {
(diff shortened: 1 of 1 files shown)


Hummer12007:
Are you sure that it influences anything? If I recall correctly, the cases in which this occurs were dealt before. (Or not?)

thejan2009:
I had a clash if workspace_auto_back_and_forth was on and I issued command 'workspace back_and_forth'. It usually worked, but after creating a new view, this would switch from current workspace to view's workspace and back. Switching to view's workspace was done via set_focused_container, so prev_workspace_name wasn't changed.  When workspace_switch was called, prev_workspace_name would be overwritten by current workspace, which messed up back_and_forth functionality.

It only happened under drm backend though, the issue didn't appear under x11.

PR #840, opened 2016-08-05
added get_inputs hint to man 5 sway-input

minus7:
Making it more obvious where to get the proper input device names.


SirCmpwn:
Thanks!

PR #843, opened 2016-08-07
Fix container move when workspace is focused

### LINKED ISSUE #819: Moving a container to a different workspace doesn't work
To reproduce:

* Start two programs
* `focus parent` (`$mod+a` per default)
* `move container to workspace other_workspace` (e.g. `$mod+shift+2`)

and nothing happens. On the command line, I see

    [handlers.c:576] Command 'move' failed: Can only move containers and views.

I just installed `sway-git` from the AUR, `sway --version` says

    sway version 0.9-5-g947570e (2016-08-02, branch "makepkg")

thejan2009:
Fixes #819. If workspace is focused and command 'move container to
workspace/output' is issued, workspace child containers are wrapped in a
new container and moved according to command.


SirCmpwn:
Thanks!

PR #846, opened 2016-08-09
document all xkb settings handled by wlc

mmlb:
see https://github.com/Cloudef/wlc/blob/09146b68163978aa0a49c2d271c9f891e1538d42/src/compositor/seat/seat.c#L311-L315


SirCmpwn:
Thanks!

mmlb:
np, thanks for sway!

PR #853, opened 2016-08-11
Implement indicator color

guiniol:
I implemented the indicator color as discussed in the IRC channel tonight.
I'm not sure what the testing guidelines are. I tested on my machine and it (seems) to work.


SirCmpwn:
Thanks!

PR #862, opened 2016-08-18
Add input example section to config

goetzc:
Also improve sway-input(5) description to be more consistent with sway-bar(5).


SirCmpwn:
Thanks!

PR #866, opened 2016-08-20
Use default black background if no wallpaper is specified

kb100:
Fixes: https://github.com/SirCmpwn/sway/issues/865


SirCmpwn:
How about just not launching swaybg unless a background is specified?>

kb100:
New pull request taking your suggested approach https://github.com/SirCmpwn/sway/pull/867

PR #869, opened 2016-08-22
Fix a format string

Hummer12007:


SirCmpwn:
You should use `%lu`

Hummer12007:
RIP, it's sometimes useful to try and compile stuff. (and it shall be fine now)

PR #874, opened 2016-08-30
Fix to make ipc client code FreeBSD compatible.

yohanesu75:
Calculating size of struct using "sizeof(addr.sun_family) + strlen(addr.sun_path)" is not valid on FreeBSD.


SirCmpwn:
Thanks!

PR #878, opened 2016-09-01
Correct `hide_edge_borders` command

lukaslihotzki:
`vertical` and `horizontal` should be swapped.

If border_left and border_right were set to 0, the vertical borders
instead of the horizontal borders would be hidden.

i3 handles this command equally.


SirCmpwn:
Thanks!

PR #879, opened 2016-09-02
refactor commands.c

zandrmartin:
This doesn't include a couple of recent commits against master. I just wanted some feedback before I bring it up to date.


SirCmpwn:
:+1:

zandrmartin:
This is ok to merge now if there are no further concerns I need to address.

SirCmpwn:
Thanks!

PR #889, opened 2016-09-10
Use indicator border color only when needed.

thejan2009:
Indicator border color is now used only when a container is the only
child. Reason for the change? i3 does it this way. Sway container
borders are now a bit more similar to i3 ones.


SirCmpwn:
Sync up with master?

thejan2009:
Sorry, didn't notice.

SirCmpwn:
No worries.

SirCmpwn:
Thanks!

PR #903, opened 2016-09-17
Fix swaybar when running on named outputs.

RyanDwyer:
When using a bar on a named output, load_swaybars() requires the
output to be active (ie. in the root container), but this is not the case if
the bar is added to the last output. To fix this, load_swaybars() is now
called after the output has been added to the root container.

After fixing that, swaybar would segfault due to using the wrong index
variable when loading outputs and config.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -936,9 +936,6 @@ void apply_output_config(struct output_config *oc, swayc_t *output) {
 			execvp(cmd[0], cmd);
 		}
 	}
-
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Thanks! Can you associate the email you committed with on your Github account?

RyanDwyer:
Done.

SirCmpwn:
Thanks!

PR #904, opened 2016-09-17
fix view focus property in get_tree json

zandrmartin:
This changes how the `focused` property is set on views in the `get_tree` JSON response. Previously, all views in a focused parent container were marked as focused; now only the focused view is.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -132,11 +132,9 @@ static void ipc_json_describe_output(swayc_t *output, json_object *object) {
 
 static void ipc_json_describe_workspace(swayc_t *workspace, json_object *object) {
 	int num = (isdigit(workspace->name[0])) ? atoi(workspace->name) : -1;
-	bool focused = root_container.focused == workspace->parent && workspace->parent->focused == workspace;
 	const char *layout = ipc_json_layout_description(workspace->layout);
 
 	json_object_object_add(object, "num", json_object_new_int(num));
-	json_object_object_add(object, "focused", json_object_new_boolean(focused));
 	json_object_object_add(object, "output", (workspace->parent) ? json_object_new_string(workspace->parent->name) : NULL);
 	json_object_object_add(object, "urgent", json_object_new_boolean(false));
 	json_object_object_add(object, "type", json_object_new_string("workspace"));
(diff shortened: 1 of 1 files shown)


zandrmartin:
Hm, this may need further work, actually. I'm not sure I fully understand how the focus chain works.

zandrmartin:
Updated. To get the focused view from the tree, recursively search from root to a view where every child has a `"focused": true` attribute. If any container in the chain is not focused, keep searching. In this (shortened) example, `view4` is the focused view:

```json
{
   "name": "root",
   "nodes": [
     {
       "name": "output1",
       "nodes": [
         {
           "name": "workspace1",
           "focused": true,
           "nodes": [
             {
               "name": "view1",
               "focused": false,
               "nodes": []
             },
             {
               "name": "container1",
               "focused": true,
               "nodes": [
                 {
                   "name": "container2",
                   "focused": false,
                   "nodes": [
                     {
                       "name": "view2",
                       "focused": true,
                       "nodes": []
                     },
                     {
                       "name": "view3",
                       "focused": false,
                       "nodes": []
                     }
                   ]
                 },
                 {
                   "name": "view4",
                   "focused": true,
                   "nodes": []
                 }
               ]
             }
           ]
         }
       ]
     },
     {
       "name": "output2",
       "nodes": [
         {
           "name": "workspace2",
           "focused": false,
           "nodes": []
         }
       ]
     }
   ]
}   
```
This behavior makes sense to me but should definitely be reviewed to make sure it is "correct."

zandrmartin:
Discussion on IRC with @Hummer12007 has revealed i3's tree response has only a single focused container. So our tree response is not currently i3-compatible and this PR will not make it so.

zandrmartin:
Third revision: tree now only has a single focused container which should match the i3 tree model.

SirCmpwn:
This solution isn't very effecient. It's going to traverse the tree for every container in the response.

Just do this: if this container isn't a C_VIEW, then false. Otherwise, go up the parents until you either hit the root container (true) or a parent whose focused property doesn't match where you just came from (false).

zandrmartin:
The focused container is not always a view though. Sometimes it's a workspace or a non-view container.

SirCmpwn:
Fair. Could you instead make this function take in the focused container as an argument and compute it before calling the function?

zandrmartin:
I get where you're coming from but passing the focused container as an argument seems a bit clunky, just because this function is called from a few different places. It could be done though - I count 4 different locations this is called. Ideally we'd have some kind of mechanism to cache the value on the first call but knowing when to invalidate the cache could be tricky.

Apparently i3 has a single global pointer to the focused container. Adding that to Sway would be non-trivial but might it be worthwhile? Maybe a `root->focused_container` or `config->focused_container`?

PR #906, opened 2016-09-17
add global `current_focus` pointer

zandrmartin:
Based on some test logging I think setting this pointer in `set_focused_container()` should be sufficient. I also set it to the root container in `init_layout()`; I do not know if that is a useful thing to do.


zandrmartin:
Hm. The whitespace changes I made looked right in Neovim but look wrong here. ¯\_(ツ)_/¯

PR #909, opened 2016-09-18
implement "focused container" option for swaygrab

zandrmartin:
`swaygrab -f` will grab only the focused container instead of the whole output. Tested with both the image and video options. Requires [this WLC commit](https://github.com/Cloudef/wlc/commit/5c4528bd3ff7e4bdcd7356b5f902b10d5be9a27f), otherwise the geometry gets wonky.

### DIFF
--- a/include/swaygrab/json.h
+++ b/include/swaygrab/json.h
@@ -1,0 +1,10 @@
+#include <json-c/json.h>
+#include "wlc/wlc.h"
+
+void init_json_tree(int socketfd);
+void free_json_tree();
+char *get_focused_output();
+char *create_payload(const char *output, struct wlc_geometry *g);
+struct wlc_geometry *get_container_geometry(json_object *container);
+json_object *get_focused_container();
+json_object *get_output_container(const char *output);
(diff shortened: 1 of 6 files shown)


SirCmpwn:
Can we add support for a specified container?

What happens if that container is moved or resized while recording?

zandrmartin:
If a container is moved or resized while recording, it just won't be in frame. The geometry is set from Swaygrab once, and it stays that way. Not sure how big a concern this is - I'd assume if you're going to be moving containers around you'd just capture the whole output.

It would be possible to add support for a specified container I suppose but I'm not sure how you'd identify said container. The "id" field we have in the tree response is the pointer to a given container so I guess that could work.

SirCmpwn:
>If a container is moved or resized while recording, it just won't be in frame. The geometry is set from Swaygrab once, and it stays that way. Not sure how big a concern this is - I'd assume if you're going to be moving containers around you'd just capture the whole output.


Alright

>It would be possible to add support for a specified container I suppose but I'm not sure how you'd identify said container. The "id" field we have in the tree response is the pointer to a given container so I guess that could work.

Yes

Issue #884, opened 2016-09-06
Missing call to setgroups before setuid in sway/main.c


Fale:
In https://github.com/SirCmpwn/sway/blob/master/sway/main.c#L159 there is a call to setgroups before setuids.
There is a high probability this means it didn't relinquish all groups, and this would be a potential security issue to be fixed.
See https://www.securecoding.cert.org/confluence/display/c/POS36-C.+Observe+correct+revocation+order+while+relinquishing+privileges

Fale:
Closed via #911

PR #911, opened 2016-09-20
Split setgid and setuid, add privilege check [RFC]

thejan2009:
This commit deals with issue #884. I consulted the following sources:
https://www.securecoding.cert.org/confluence/display/c/POS36-C.+Observe+correct+revocation+order+while+relinquishing+privileges
and
https://www.securecoding.cert.org/confluence/display/c/POS37-C.+Ensure+that+privilege+relinquishment+is+successful

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -156,10 +156,18 @@ int main(int argc, char **argv) {
 			exit(EXIT_FAILURE);
 		}
 		if (getuid() != geteuid() || getgid() != getegid()) {
-			if (setgid(getgid()) != 0 || setuid(getuid()) != 0) {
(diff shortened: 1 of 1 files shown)


sway/main.c
```diff
@@ -156,10 +156,18 @@ int main(int argc, char **argv) {
 			exit(EXIT_FAILURE);
 		}
 		if (getuid() != geteuid() || getgid() != getegid()) {
-			if (setgid(getgid()) != 0 || setuid(getuid()) != 0) {
+			if (setgid(getgid()) != 0){
```
SirCmpwn: Style error here, space between ) and {

PR #915, opened 2016-09-24
Implement default name for workspace command

RyanDwyer:
This implements commands such as:

    workspace number 9: test

If a workspace with the given number exists then it will be focused,
otherwise a new workspace with the given name will be created.

This functionality is as per i3's.

I'm not an expert at C, so if there's an easier way to get argv[1] onwards as a string then let me know and I'll amend.

### DIFF
--- a/sway/commands/workspace.c
+++ b/sway/commands/workspace.c
@@ -5,21 +5,23 @@
 #include "sway/workspace.h"
 #include "list.h"
 #include "log.h"
+#include "stringop.h"
 
 struct cmd_results *cmd_workspace(int argc, char **argv) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
https://github.com/SirCmpwn/sway/blob/master/include/stringop.h#L31

RyanDwyer:
Amended.

SirCmpwn:
Do you need to free the string? Not sure if create_workspace dupes it.

RyanDwyer:
Yeah, it dupes it, so I've added free() to my commit.

SirCmpwn:
Thanks!

PR #921, opened 2016-10-01
fix pointer format string in sway_log() call

### LINKED ISSUE #919: Compiling problem on x86
Trying to compile v0.10-rc1 on x86, I get the following error.

```
/builddir/build/BUILD/sway-0.10-rc1/sway/focus.c: In function 'set_focused_container':
/builddir/build/BUILD/sway-0.10-rc1/sway/focus.c:117:47: error: format '%lu' expects argument of type 'long unsigned int', but argument 5 has type 'wlc_handle {aka unsigned int}' [-Werror=format=]
  swayc_log(L_DEBUG, c, "Setting focus to %p:%lu" PRIuPTR, c, c->handle);
                                               ^
```

For more info on the build environment: http://koji.fedoraproject.org/koji/taskinfo?taskID=15850465 (interesting log files are probably root.log and build.log. Those two lines I copied here are toward the end of the build.log one)

zandrmartin:
this _might_ fix #919

### DIFF
--- a/sway/focus.c
+++ b/sway/focus.c
@@ -114,7 +114,7 @@ bool set_focused_container(swayc_t *c) {
 		c = get_focused_container(workspace);
 	}
 
-	swayc_log(L_DEBUG, c, "Setting focus to %p:%lu" PRIuPTR, c, c->handle);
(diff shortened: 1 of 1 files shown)


SirCmpwn:
@Fale does it?

PR #923, opened 2016-10-02
moved ipc_init above config file processing.

gutsoo:
This is necessary because commands in the config file (mode for
instance) emit ipc events, and if ipc_init has not been called the
ipc_clients_list is not initialized, and we segfault.  This fixes that
bug.


SirCmpwn:
Thanks!

PR #933, opened 2016-10-06
add click on title_bar to focus a container

### LINKED ISSUE #929: Clicking tabs does not switch focus
In i3wm, one of the laziest ways to switch to another tab is to click its "decoration" (for the lack of a better term, I mean the rectangle rendered above it with the title inside).  In sway, this doesn't do anything (yet).

alkino:
Fix #929

### DIFF
--- a/sway/handlers.c
+++ b/sway/handlers.c
@@ -776,6 +776,14 @@ static bool handle_pointer_motion(wlc_handle handle, uint32_t time, const struct
 	return EVENT_PASSTHROUGH;
 }
 
+bool swayc_border_check(swayc_t *c, const void *_origin) {
+	const struct wlc_point *origin = _origin;
+	if (origin->x >= c->title_bar_geometry.origin.x && origin->y >= c->title_bar_geometry.origin.y
+		&& origin->x < (c->title_bar_geometry.origin.x + (int)c->title_bar_geometry.size.w) && origin->y < (c->title_bar_geometry.origin.y + (int)c->title_bar_geometry.size.h)) {
+		return true;
+	}
+	return false;
+}
 
 static bool handle_pointer_button(wlc_handle view, uint32_t time, const struct wlc_modifiers *modifiers,
 		uint32_t button, enum wlc_button_state state, const struct wlc_point *origin) {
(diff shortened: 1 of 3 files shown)


sway/handlers.c
```diff
@@ -776,6 +776,32 @@ static bool handle_pointer_motion(wlc_handle handle, uint32_t time, const struct
...
+		if (f(container->children->items[i], data))
+			return container->children->items[i];
+
+		swayc_t *find = container_find(container->children->items[i], f, data);
+		if (find != NULL)
+			return find;
+	}
+
+	return NULL;
+}
```
SirCmpwn: Please put this in containers.c and expose it from containers.h

sway/handlers.c
```diff
@@ -840,6 +866,13 @@ static bool handle_pointer_button(wlc_handle view, uint32_t time, const struct w
 	// Check whether to change focus
 	swayc_t *pointer = pointer_state.view;
 	if (pointer) {
+		swayc_t *ws = swayc_parent_by_type(focused, C_WORKSPACE);
+		swayc_t *find = container_find(ws, &swayc_border_check, origin);
+		if (find != NULL) {
+			set_focused_container(find);
+			return EVENT_HANDLED;
+		}
+
```
SirCmpwn: This should probably live outside of the if statement, right?

sway/handlers.c
```diff
@@ -776,6 +776,32 @@ static bool handle_pointer_motion(wlc_handle handle, uint32_t time, const struct
 	return EVENT_PASSTHROUGH;
 }
 
+bool swayc_border_check(swayc_t *c, const void *_origin)
+{
+	const struct wlc_point *origin = _origin;
+	if (origin->x >= c->title_bar_geometry.origin.x && origin->y >= c->title_bar_geometry.origin.y
+		&& origin->x < (c->title_bar_geometry.origin.x + (int)c->title_bar_geometry.size.w) && origin->y < (c->title_bar_geometry.origin.y + (int)c->title_bar_geometry.size.h)) {
```
SirCmpwn: this line is too wide, split it up a bit

sway/handlers.c
alkino: I don't think so.
If the user doesn't click on any title_bar process should continue.

SirCmpwn:
Thanks!

PR #934, opened 2016-10-06
fix for click on title_bar

alkino:
- Check null pointer for swayc_parent_by_type
- Split line to follow coding style


SirCmpwn:
Thanks!

PR #936, opened 2016-10-07
parse pango markup in workspace names (and bugfix)

### LINKED ISSUE #897: Support for numeric character reference for workspace names
In i3(or at least i3-gaps version) we have support for this characters(example[with spaces, because github replace code with one char just like i3bar]: & # x0C8D;), which are great when we using icon fonts like font awesome. In sway we have just plain text. Any chance for support that?
https://en.wikipedia.org/wiki/Numeric_character_reference

thejan2009:
Fixes #897. This change allows using numeric character references in workspace names
- for example &#230; which stands for sharp s. A fix was necessary in
get_pango_layout, since markup and parsed markup had different width.


SirCmpwn:
Thanks!

PR #942, opened 2016-10-08
fix for_window [] move scratchpad

thejan2009:
- move wlc_view_set_mask before applying window criteria
 - check parent_container because container in scratchpad doesn't have a
 parent


SirCmpwn:
Thanks!

PR #945, opened 2016-10-10
Fixes dealing with workspace_layout and related bugs [rfc]

### LINKED ISSUE #901: Crash with workspace_layout tabbed on floating container creation
When there is only one direct child of workspace with setting ```workspace_layout tabbed```, sway crashes if a floating window is created.

The issue is caused in file handlers.c, just below line 401. Conditions are done wrong - there is a missing check for floating view, and a check if a wrapper container already exists. I'd suggest postponing fixing this issue until we get a more detailed clarification about workspace_layout from i3 (that was referenced in PR #793).

thejan2009:
Fix #901. Minimal working example: default config, on empty workspace $mod+w, ```swaymsg exec nautilus```, Ctrl+i. Previously you'd get a segfault, after fixing that the popup would appear, but nautilus would become invisible. That's fixed in the second commit.

### DIFF
--- a/sway/commands/layout.c
+++ b/sway/commands/layout.c
@@ -22,10 +22,10 @@ struct cmd_results *cmd_layout(int argc, char **argv) {
 	enum swayc_layouts old_layout = parent->layout;
 
 	if (strcasecmp(argv[0], "default") == 0) {
-		parent->layout = parent->prev_layout;
+		swayc_change_layout(parent, parent->prev_layout);
 		if (parent->layout == L_NONE) {
 			swayc_t *output = swayc_parent_by_type(parent, C_OUTPUT);
-			parent->layout = default_layout(output);
+			swayc_change_layout(parent, default_layout(output));
 		}
 	} else {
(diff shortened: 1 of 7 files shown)


sway/commands/layout.c
```diff
@@ -37,22 +37,22 @@ struct cmd_results *cmd_layout(int argc, char **argv) {
...
-			parent->layout = L_VERT;
+			swayc_change_layout(parent,L_VERT);
 		} else if (strcasecmp(argv[0], "toggle") == 0 && argc == 2 && strcasecmp(argv[1], "split") == 0) {
 			if (parent->layout == L_VERT) {
-				parent->layout = L_HORIZ;
+				swayc_change_layout(parent,L_HORIZ);
 			} else {
-				parent->layout = L_VERT;
+				swayc_change_layout(parent,L_VERT);
 			}
```
SirCmpwn: style: add spaces after the commas in function calls

thejan2009:
I found a better way to do it. Views under workspaces should now be encapsulated pretty much every time.

PR #948, opened 2016-10-11
also check floating cons in container_find

### LINKED ISSUE #944: Title_bar should be part of container
Hi o/

Today if you click on the title_bar of a floating windows it clicks on container behind it. It will be better if it click on it.

Regards

thejan2009:
fix #944

### DIFF
--- a/sway/container.c
+++ b/sway/container.c
@@ -727,14 +727,29 @@ swayc_t *container_find(swayc_t *container, bool (*f)(swayc_t *, const void *), 
 		return NULL;
 	}
 
+	swayc_t *con;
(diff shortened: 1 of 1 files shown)


sway/container.c
```diff
@@ -727,14 +727,29 @@ swayc_t *container_find(swayc_t *container, bool (*f)(swayc_t *, const void *),
 		return NULL;
 	}
 
+	swayc_t *con;
+	if (container->type == C_WORKSPACE) {
+		for (int i = 0; i < container->floating->length; ++i) {
```
alkino: you don't check that container->floating is not NULL.

PR #961, opened 2016-10-25
Add left_handed support for input devices

t3hknr:
Some users may want to switch buttons on their input devices, turns out
libinput already supports it. Let's add a support for it in our config.

Signed-off-by: Michał Winiarski <knr@hardline.pl>


SirCmpwn:
Can you fix your email on your github account so the commit gets associated with your user?

SirCmpwn:
Looks good save for that.

t3hknr:
Done

SirCmpwn:
Thanks!

PR #963, opened 2016-11-01
workspace_layout and container wrapping

thejan2009:
Let's try this again.

First commit introduces workspace_layout and ensures that layout of workspace container will always be horizontal or vertical, not tabbed or stacked. The second commit wraps views in containers when needed.

### DIFF
--- a/include/sway/container.h
+++ b/include/sway/container.h
@@ -324,7 +324,7 @@ void close_views(swayc_t *container);
 
 /**
  * Assign layout to a container. Needed due to workspace container specifics.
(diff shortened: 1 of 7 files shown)


thejan2009:
Needs additional testing on multi-head setup: observe how views and containers move between workspaces on different outputs when you issue "move (right|left|up|down)" command. If that works as expected, this should be good to merge.

SirCmpwn:
Rebased and merged: https://github.com/SirCmpwn/sway/commit/35b8d185ac58eaaeb12a95d6947cdcb78a2ee487

PR #965, opened 2016-11-02
Fix some colour settings in swaybar

### LINKED ISSUE #932: Resize indicator looks different from i3
[i3](http://i.imgur.com/HQwdmYK.png)
[sway](http://i.imgur.com/E94utJi.png)

Both use the same config and do otherwise look the same.

thejan2009:
This should fix #932 - if binding_mode colours are not defined, use urgent_workspace colours as a fallback.

Three missing settings were also implemented: focused_(workspace|statusline|separator)
They provide different colours for swaybar on the focused output.


include/sway/config.h
```diff
@@ -163,6 +166,14 @@ struct bar_config {
...
 		char binding_mode_bg[10];
 		char binding_mode_text[10];
+
+		bool has_focused_background;
+		bool has_focused_statusline;
+		bool has_focused_separator;
+
+		bool has_binding_mode_border;
+		bool has_binding_mode_bg;
+		bool has_binding_mode_text;
```
SirCmpwn: Is there any way we could encode these into the char arrays? i.e. "inherit" or something.

sway/ipc-json.c
```diff
@@ -328,9 +346,23 @@ json_object *ipc_json_describe_bar_config(struct bar_config *bar) {
...
 	json_object_object_add(colors, "urgent_workspace_text", json_object_new_string(bar->colors.urgent_workspace_text));
 
-	json_object_object_add(colors, "binding_mode_border", json_object_new_string(bar->colors.binding_mode_border));
-	json_object_object_add(colors, "binding_mode_bg", json_object_new_string(bar->colors.binding_mode_bg));
-	json_object_object_add(colors, "binding_mode_text", json_object_new_string(bar->colors.binding_mode_text));
+	if (bar->colors.has_binding_mode_border) {
+		json_object_object_add(colors, "binding_mode_border", json_object_new_string(bar->colors.binding_mode_border));
+	} else {
+		json_object_object_add(colors, "binding_mode_border", json_object_new_string(bar->colors.urgent_workspace_border));
+	}
```
mikkeloscar: Does i3 do this, or why would you fall back to `urgent_workspace_border` ?

thejan2009: It does, https://github.com/i3/i3/blob/next/i3bar/src/xcb.c#L430

sway/config.c
```diff
@@ -1109,6 +1131,12 @@ struct sway_binding *sway_binding_dup(struct sway_binding *sb) {
 	return new_sb;
 }
 
+static void set_bar_color(char **name, char *value) {
+	// every color has 9 characters plus \0
+	*name = malloc(10);
+	strcpy(*name, value);
+}
```
SirCmpwn: How about just using strdup everywhere instead of this function?

minus7:
Why not keep the static buffers? Is there any need for them to be dynamically allocated?

SirCmpwn:
Putting them on the heap is a bit cleaner imo

PR #981, opened 2016-12-02
Security features

SirCmpwn:
* [x] Default security config for /etc/sway/config.d/
* [x] Run config files through sed to replace PREFIX and SYSCONFDIR
* [x] Config structure changes & support code
* [x] Implement the permit and reject commands
* [x] Feature policies
  * [x] background
  * [x] panel
  * [x] lock
  * [x] lock
  * [x] fullscreen
  * [x] keyboard
  * [x] mouse
  * [x] ipc
* [x] Command policies
* [x] IPC security
  * [x] Enable/disable features
  * [x] Enable/disable events
* [x] Startup sanity check
  * [x] Check permissions on /etc/sway
  * [x] Check procfs is available
  * [x] Check for CAP_SYS_PTRACE
  * [x] Check for security sensitive command policies
* [x] Write sway-security(7)
* [x] Drop -Denable-binding-event from cmake

To do seperately:

* [ ] Optimize permission checks
* [ ] Feature policies
  * [ ] screenshot
* [ ] New Wayland protocol extensions
  * [ ] Screenshot
* [ ] Startup sanity check
  * [ ] Check permissions of all config files
  * [ ] Inform user visually of sanity check failures
* [ ] Gracefully handle permission errors in swaybar, swaygrab, swaymsg, etc
* [ ] Paranoid mode for swaygrab
* [ ] Standard mechanisms for sandboxing (collab with wayland-devel)
* [ ] Configure IPC features per-executable
* [ ] Match on /proc/[pid]/cmdline in addition to exe for permissions

I'm accepting PRs against the security branch if anyone wants to help. Feedback welcome.


SirCmpwn:
Paging @yohanesu75 for FreeBSD review

SirCmpwn:
Alright, this pull request should be good to go. Going to leave it open for a couple of days to gather feedback and reviews - please take a look at this and let me know what you think. Can you find some flaws?

Will tag 0.12-rc1 once this is merged (will grab #963 too).

minus7:
Regarding default policies: Imho the default, without explicit configuration, should be to deny everything. This way there's no possibity to miss denying some permission when trying to create a secure setup. The config then has to explicity have to grant "default" permissions (like keyboard/mouse/fullscreen) explicitly via wildcard match. From a usage perspective this isn't a problem because an example ("default") config is shipped anyway.

(Careful with the word "default" here, as it's used to refer to two different contexts)

PR #985, opened 2016-12-08
Fix build on FreeBSD

myfreeweb:
Depends on Cloudef/wlc/pull/211 & Cloudef/chck/pull/26

- Make sure CMake always finds absolute paths for Cairo, Pango and GdkPixbuf (because /usr/local/lib is not on /usr/bin/ld's default path)
- Add forgotten json-c include path to swaymsg/CMakeLists.txt
- Disable -Werror because of assert-related warnings
- Add correct /proc/pid/file path for FreeBSD
- Use [libepoll-shim](https://github.com/jiixyj/epoll-shim) on FreeBSD
- Only use Linux capabilities on, well, Linux

Current status:
- XWayland works!! (`libGL` built with `DRI3`, `xorg-server` built with `DRI3`, `XWL` and `UDEV`)
- for some reason (security thing broken?) `swaybar` gets a full window, not a bar
- window manipulation (e.g. closing a window with a sway shortcut, sometimes even opening windows) usually results in the compositor freezing in an unkillable `kqflxw` state, sometimes with a 100% CPU loop before that (but that's in wlc or something, confirmed on way-cooler) :( but maybe i built some library wrong? @yohanesu75


SirCmpwn:
>for some reason (security thing broken?) swaybar gets a full window, not a bar

Does sound like security related. Check the logs. Are you including the default security rules correctly?

SirCmpwn:
>Disable -Werror because of assert-related warnings

What warnings? I'm not disabling Werror unconditionally.

sway/ipc-server.c
```diff
@@ -125,6 +125,8 @@ struct sockaddr_un *ipc_user_sockaddr(void) {
 	return ipc_sockaddr;
 }
 
+// FreeBSD supports getting uid/gid, but not pid
```
SirCmpwn: Is there something else we could use to get the pid of the client? Have you filed a bug with FreeBSD asking for this feature? This is crucial to IPC security.

sway/ipc-server.c
```diff
@@ -154,6 +157,7 @@ int ipc_handle_connection(int fd, uint32_t mask, void *data) {
 		return 0;
 	}
 
+#ifdef __linux__
```
SirCmpwn: Don't comment this out, just have the FreeBSD version of get_client_pid return -1.

sway/ipc-server.c
yohanesu75: I'll look into it.

yohanesu75:
I also had problems with freezing at kqflxw. Haven't really dug into it yet. I've been away for a while but I have some time this week to experiment. Is the hang using EGL, shared memory or both?  @myfreeweb

myfreeweb:
So yeah, the whole security thing relies on `get_client_pid` and that's why swaybar can't get the permission to be a bar. I'm going to try a super hacky solution — do what `sockstat` does.

BTW, seems like the warning errors are just gone… o_0

yohanesu75:
How about using cmsgcred? 
http://stackoverflow.com/questions/8104904/identify-program-that-connects-to-a-unix-domain-socket

SirCmpwn:
>`permit * fullscreen keyboard mouse ipc lock panel background screenshot` is the way to go for now.

Definitely not. The only problem is finding the pid for a unix socket client, right? We can still find pids for wayland clients, I hope. If so, we can enforce everything *but* IPC:

    permit * fullscreen keyboard mouse ipc

We definitely don't want to give sway shell permissions to everything by default.

+1 to @yohanesu75's solution for getting the pid of the socket client, too, then we shouldn't have to make any compromises.

myfreeweb:
> We definitely don't want to give sway shell permissions to everything by default.

Worked fine for Xorg :D I didn't mean changing the default config file, of course — just let security stay broken on FreeBSD *for now*. Fixing the unkillable freezes that literally make the compositors unusable is more important.

cmsgcred seems like a solution, nice find @yohanesu75! But it introduces a new problem: it requires a message. In the case of extension permissions, the pid comes from `wl_client_get_credentials`, which…

> We can still find pids for wayland clients, I hope

…does the exact same thing (`LOCAL_PEERCRED`) which is uid/gid only, pid = 0. (In the current [patch](https://github.com/FreeBSDDesktop/freebsd-ports-graphics/blob/6feeb91cb27a5601b0bea6a3427024e110d33555/wayland/wayland/files/patch-src_wayland-server.c)).

At least in the sway IPC, I see where the `recv` happens. I don't know where it happens for Wayland protocol messages.

Anyway, again — I think that fixing security should be a separate PR.

SirCmpwn:
>Anyway, again — I think that fixing security should be a separate PR.

I don't.

yohanesu75:
> …does the exact same thing (LOCAL_PEERCRED) which is uid/gid only, pid = 0. (In the current patch).

This is interesting. Exactly what implications does this have for Wayland? Maybe we need to fix this patch to use correct pid as well. Can it be related to the freezing?

myfreeweb:
I don't think it's related to the freezing. But it's definitely related to sway's security thing.

Once again, it's only possible to use `cmsgcred` IN THE CODE THAT DOES THE MESSAGE SEND/RECV. I don't know where that is. wlc?

yohanesu75:
> IN THE CODE THAT DOES THE MESSAGE SEND/RECV

Send/Recv between what exactly?  WLC do open a socketpair and use for communication between parent and child process, is this what you're looking for?

myfreeweb:
> I don't know where it happens for Wayland protocol messages

> Wayland protocol messages

> Wayland protocol

the one between the wayland client and wayland server

SirCmpwn:
So what's the current status of this? Going to ship rc2 without it tomorrow if it's not fixed up in time.

Regarding pids, sway needs them in two situations:

* From wayland clients via wl_client_get_credentials (from wayland-server) or wlc_view_get_pid (from wlc) to enforce all permissions except for IPC access
* From Unix socket clients via getsockopt/SO_PEERCRED to enforce IPC access

The former is a concern of sway's dependencies, wayland-server and wlc. The latter is sway's concern.

This PR should fix the latter, if possible, on FreeBSD. My understanding of the discussion so far is that it can do so.

As for the former, this PR needn't provide a solution (the problem doesn't live in our code anyway). I *strongly* encourage you to also send patches to the dependencies if these interfaces do not work on FreeBSD. If these interfaces do not work on FreeBSD, I will include a warning in the release notes about it, and maybe in the man page, when I ship sway 0.11.

Sounds good?

myfreeweb:
Do you think that

- the client sending a hello message (with the credentials) as soon as it opens the socket
- and the server *synchronously waiting* for that message in `get_client_pid`

is an acceptable solution?

(Not blocking there would require more complex functionality — asynchronously elevating permissions after receiving messages.)

SirCmpwn:
Couldn't the client just pretend to be a different pid than it actually is?

myfreeweb:
No, the pid is written by the kernel

SirCmpwn:
The kernel writes the hello message then? How does that work? Does this break compatability with i3 IPC?

SirCmpwn:
Possibly helpful: https://lists.freebsd.org/pipermail/freebsd-bugs/2006-September/019991.html

myfreeweb:
The client sends a message with an empty `cmsgcred` struct as ancillary data. The kernel fills it in. The kernel guarantees that any messages with that struct have data from the kernel.


> Possibly helpful: https://lists.freebsd.org/pipermail/freebsd-bugs/2006-September/019991.html

`linux_ucred.pid = 0;`

SirCmpwn:
Yeah, that seems to require changes on the client level, which I'm not fond of at all. I suppose we can move forward with FreeBSD support being insecure for now. Will you start a process with FreeBSD to make interfaces that support our needs possible? It shouldn't be too terribly difficult to add support for getting the pid of a unix socket client (or server!) to the kernel.

SirCmpwn:
Is there a sway package for FreeBSD? Not entirely familiar with how software is distributed on FreeBSD. Should make sure the default security config it installs has this line:

    permit * fullscreen mouse keyboard background panel lock ipc

Which will fix the issues with swaybar, swaybg, etc (at the expense of not being able to lock down access to those features).

myfreeweb:
There's [a port](https://github.com/FreeBSDDesktop/freebsd-ports-graphics/blob/xserver-mesa-next/wayland/sway/Makefile) by @yohanesu75 in the graphics (FreeBSDDesktop) fork. (If you're familiar with Arch, ports are like PKGBUILDs, packages are built from that). Wayland is not yet supported in upstream FreeBSD.

Filed https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=215202

yohanesu75:
@myfreeweb Will you handle these issues or is there something you want me to do (since I been working on this before..)? Ps, I like your rusty-sandbox :) Rust FTW!

myfreeweb:
@yohanesu75 yeah, please update the ports (sway, wlc, wayland-protocols…)

yohanesu75:
> yeah, please update the ports (sway, wlc, wayland-protocols…)

@myfreeweb  Do you mean only the ports (Makefile etc), or required changes to the source codes as well?

myfreeweb:
The patches in the ports for sure. I don't think anything else needs changing for now.

SirCmpwn:
Thanks!

PR #996, opened 2016-12-17
Use CMAKE_INSTALL_FULL_DATA{,ROOT}DIR

woutershep:
Exherbo installs architecture dependent data in a different place than architecture
independent data. More concretely: binaries go in /usr/$chost/{bin,lib},
data goes in /usr/share and configs in /etc, /etc is already configurable
through CMAKE_INSTALL_FULL_SYSCONFDIR but the datadir was not. This
patch fixes it so that things can be pushed in the right places.


SirCmpwn:
Thanks!

PR #1001, opened 2016-12-18
Put floating views in center, fallback on top left

alkino:
Sometimes views are created before we got its size

I propose this because I got a little screen and this is boring.

### DIFF
--- a/sway/container.c
+++ b/sway/container.c
@@ -340,9 +340,18 @@ swayc_t *new_floating_view(wlc_handle handle) {
 	struct wlc_geometry geometry;
 	wlc_view_get_visible_geometry(handle, &geometry);
 
-	// give it requested geometry, but place in center
-	view->x = (swayc_active_workspace()->width - geometry.size.w) / 2;
-	view->y = (swayc_active_workspace()->height- geometry.size.h) / 2;
+	// give it requested geometry, but place in center if possible
+	// in top left otherwise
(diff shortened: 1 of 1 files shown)


sway/container.c
```diff
@@ -340,9 +340,16 @@ swayc_t *new_floating_view(wlc_handle handle) {
 	struct wlc_geometry geometry;
 	wlc_view_get_visible_geometry(handle, &geometry);
 
-	// give it requested geometry, but place in center
-	view->x = (swayc_active_workspace()->width - geometry.size.w) / 2;
-	view->y = (swayc_active_workspace()->height- geometry.size.h) / 2;
+	// give it requested geometry, but place in center if possible
+	// in top left otherwise
+	if (geometry.size.w != 0)
```
zandrmartin: Add braces to `if` and `else` please, even though they only contain one statement.

sway/container.c
```diff
@@ -340,9 +340,16 @@ swayc_t *new_floating_view(wlc_handle handle) {
...
+		view->x = (swayc_active_workspace()->width - geometry.size.w) / 2;
+	else
+		view->x = 0;
+	if (geometry.size.h != 0)
+		view->y = (swayc_active_workspace()->height- geometry.size.h) / 2;
```
zandrmartin: Please separate `height` and `-`.

alkino:
Updated. Thanks for review.

PR #1014, opened 2016-12-26
sway: extensions: make locking give back focus

### LINKED ISSUE #998: Swaylock does not give back the focus
If I run swaylock from terminal, when I unlock the display, the terminal window that launched swaylock does not have the focus (it seems like no window has the focus) until I change workspace

cyphar:
Don't switch the internal tracking of focus to the swaylock surface,
to allow for switching back to the previously active window (or the
currently active window, if some new process changed).

Closes #998
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/sway/extensions.c
+++ b/sway/extensions.c
@@ -69,7 +69,15 @@ void lock_surface_destructor(struct wl_resource *resource) {
 		}
 	}
 	if (desktop_shell.lock_surfaces->length == 0) {
+		sway_log(L_DEBUG, "Desktop shell unlocked");
 		desktop_shell.is_locked = false;
+
+		// We need to now give focus back to the focus which we internally
+		// track, since when we lock sway we don't actually change our internal
+		// focus tracking.
+		swayc_t *focus = get_focused_container(swayc_active_workspace());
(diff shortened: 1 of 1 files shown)


sway/extensions.c
```diff
@@ -157,10 +167,6 @@ static void set_lock_surface(struct wl_client *client, struct wl_resource *resou
 		desktop_shell.is_locked = true;
 		input_init();
 		arrange_windows(workspace, -1, -1);
-		swayc_t *focus_output = swayc_active_output();
-		if (focus_output == output) {
-			set_focused_container(view);
-		}
```
SirCmpwn: Does the lock surface still get all of the keypresses if you remove this?

SirCmpwn: Worked when I tested it ¯\_(ツ)_/¯

SirCmpwn:
Thanks!

sway/extensions.c
cyphar: Yeah, it worked when I tested it. I think it's because the keypresses are sent to the frontmost wlc view?

PR #1019, opened 2016-12-28
Add redshift support

SirCmpwn:
Depends on https://github.com/Cloudef/wlc/pull/216 and on https://github.com/giucam/redshift


Arnaudv6:
Note to my future self: refer to [this wiki entry](https://github.com/swaywm/sway/wiki#redshift):
this is actually [Minus7](https://github.com/minus7/redshift/tree/wayland)'s fork we're using now.

PR #1032, opened 2017-01-06
Recompressed PNG files

Jesin:
zopflipng -m --lossy_transparent


SirCmpwn:
No thanks to lossy compression. Wouldn't mind lossless, though.

SirCmpwn:
Via discussion in IRC, --lossy_transparent just loses RGB data from pixels where A = 0.

PR #1039, opened 2017-01-09
Improve configuration file processing

DarkDeepBlue:
Currently there are multiple problems with the processing of the configuration file. Gerenrally, the impementation makes many assumptions that are not valid for all users. This pull request addresses two of them.

First, the current implementation fails to read options in the configuration file correctly if they contain too much whitespace between arguments. Errors emerges if the user aligns options for example. With this changes, the line is stripped from unnecessary whitespace so the assumtions of the following processing stay valid.

The second problem arises if the user places opening parentheses on a separate line. Although it is not intuitive, but in such case the current impementation fails in a very bad manner. This pull request introduces checks if the opening parentheses can be found in the next relevant line and appends it to the procesed line.

Although the behaviour is not perfect even after this changes, the rebustness of sway will be increased.

### DIFF
--- a/common/readline.c
+++ b/common/readline.c
@@ -1,51 +1,45 @@
-#include "readline.h"
-#include "log.h"
-#include <stdlib.h>
+#include <assert.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 
-char *read_line(FILE *file) {
-	size_t length = 0, size = 128;
-	char *string = malloc(size);
-	char lastChar = '\0';
-	if (!string) {
-		sway_log(L_ERROR, "Unable to allocate memory for read_line");
-		return NULL;
-	}
-	while (1) {
-		int c = getc(file);
-		if (c == '\n' && lastChar == '\\'){
-			--length; // Ignore last character.
-			lastChar = '\0';
-			continue;
-		}
-		if (c == EOF || c == '\n' || c == '\0') {
-			break;
-		}
-		if (c == '\r') {
-			continue;
-		}
-		lastChar = c;
-		if (length == size) {
-			char *new_string = realloc(string, size *= 2);
-			if (!new_string) {
-				free(string);
-				sway_log(L_ERROR, "Unable to allocate memory for read_line");
-				return NULL;
+#include "log.h"
+#include "readline.h"
+#include "stringop.h"
+
+/***************************************************************************//**
+*
+* \brief			Read one line of the configuration file.
+*
+* \param			file	Pointer to the already open configuration file. If NULL is
+*										passed, no operation is performed.
+*
+* \return			String containing one single line from the configuration file.
+*							This string is not necessarily in the format how it is expected
+*							for further processing of configuration commands. NULL is
+*							returned in case of failure or if NULL is passed for 'file'.
+*
+******************************************************************************/
+inline char* read_line(FILE* restrict const file) {
+	char* line = NULL;
+	if (file) {
+		size_t line_length = 0u;
+		const ssize_t bytes_read = getline(&line, &line_length, file);
+		if (line) {
+			assert(line_length);
+			if (0 > bytes_read) {
+				free(line);
+				line = NULL;
+			} else {
+				strip_whitespace(line);
+				if (line[0] == '#') {
+					free(line);
+					line = NULL;
+				}
 			}
-			string = new_string;
 		}
-		string[length++] = c;
 	}
-	if (length + 1 == size) {
-		char *new_string = realloc(string, length + 1);
-		if (!new_string) {
-			free(string);
-			return NULL;
-		}
-		string = new_string;
-	}
-	string[length] = '\0';
-	return string;
+	return line;
 }
 
 char *read_line_buffer(FILE *file, char *string, size_t string_len) {
--- a/common/readline.c
+++ b/common/readline.c
@@ -7,33 +7,16 @@
 #include "readline.h"
 #include "stringop.h"
 
-/***************************************************************************//**
-*
-*	\brief	Read one line of the file. Lines beginning with '#' are skipped.
-*		Whitespace is stripped.
-*
-*	\param[in]	file	Pointer to the already open configuration file. If NULL is
-*		passed, no operation is performed.
-*
-*	\return		String containing one single line from the file.
-*
-******************************************************************************/
 inline char* read_line(FILE* restrict const file) {
 	char* line = NULL;
 	if (file) {
 		size_t line_length = 0u;
 		const ssize_t bytes_read = getline(&line, &line_length, file);
-		if (line) {
-			assert(line_length);
+		if (line && line_length) {
 			if (0 > bytes_read) {
 				free(line);
 				line = NULL;
 			} else {
-				strip_whitespace(line);
-				if (line[0] == '#') {
-					free(line);
-					line = NULL;
-				}
 			}
 		}
 	}
--- a/common/readline.c
+++ b/common/readline.c
@@ -9,15 +9,13 @@
 
 /***************************************************************************//**
 *
-* \brief			Read one line of the configuration file.
+*	\brief	Read one line of the file. Lines beginning with '#' are skipped.
+*		Whitespace is stripped.
 *
-* \param			file	Pointer to the already open configuration file. If NULL is
-*										passed, no operation is performed.
+*	\param[in]	file	Pointer to the already open configuration file. If NULL is
+*		passed, no operation is performed.
 *
-* \return			String containing one single line from the configuration file.
-*							This string is not necessarily in the format how it is expected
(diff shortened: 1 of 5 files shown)


common/readline.c
```diff
@@ -1,51 +1,43 @@
...
+*
+*	\brief	Read one line of the file. Lines beginning with '#' are skipped.
+*		Whitespace is stripped.
+*
+*	\param[in]	file	Pointer to the already open configuration file. If NULL is
+*		passed, no operation is performed.
+*
+*	\return		String containing one single line from the file.
+*
+******************************************************************************/
```
SirCmpwn: -1 to all the changes in this file.

- Don't add this sort of doc comments thing. If you want to add doc comments, you should propose it in a seperate pull request. Note that we already use doxygen style comments informally in some other places as well, so if you wanted to do that sort of thing that'd be fine with me, but you should put it in the header.
- Making read_line strip whitespace is beyond what this function should do. If the caller wants whitespace stripped, the caller should strip it themselves.

common/readline.c
```diff
@@ -1,51 +1,43 @@
...
+inline char* read_line(FILE* restrict const file) {
+	char* line = NULL;
+	if (file) {
+		size_t line_length = 0u;
+		const ssize_t bytes_read = getline(&line, &line_length, file);
+		if (line) {
+			assert(line_length);
```
SirCmpwn: Use sway_assert, or better yet handle the error intelligently and return NULL. SIGABRT should never happen at runtime on sway, as a general rule.

common/readline.c
```diff
@@ -1,51 +1,43 @@
...
+*	\param[in]	file	Pointer to the already open configuration file. If NULL is
+*		passed, no operation is performed.
+*
+*	\return		String containing one single line from the file.
+*
+******************************************************************************/
+inline char* read_line(FILE* restrict const file) {
+	char* line = NULL;
+	if (file) {
+		size_t line_length = 0u;
```
SirCmpwn: Not necessary to specify u here

common/stringop.c
```diff
@@ -10,22 +10,32 @@
...
+ * \brief		Strips the whitespace in front and rear of the string. If whitespace
+ *					is found in the middle of the string, all but one space is removed
+ *					between words.
+ *
+ * \param		str	Pointer to the char string that should be stripped. The string
+ *							is modified. If NULL is passed, no operation is performed.
+ *
+ ******************************************************************************/
+void strip_whitespace(char * restrict const str)
+{
```
SirCmpwn: Join this { on the previous line

common/stringop.c
```diff
@@ -10,22 +10,32 @@
...
+	if (NULL != str) {
+		unsigned int count = 0u;
```
SirCmpwn: Not necessary to specify u everywhere

common/stringop.c
```diff
@@ -10,22 +10,32 @@
...
+ *					is found in the middle of the string, all but one space is removed
+ *					between words.
+ *
+ * \param		str	Pointer to the char string that should be stripped. The string
+ *							is modified. If NULL is passed, no operation is performed.
+ *
+ ******************************************************************************/
+void strip_whitespace(char * restrict const str)
+{
+	if (NULL != str) {
```
SirCmpwn: Just use if (str) here

common/stringop.c
```diff
@@ -10,22 +10,32 @@
...
+		unsigned int count = 0u;
+		for (unsigned int index = 0u; str[index]; ++index) {
+			if (' ' != str[index] && '\t' != str[index]) {
+				str[count++] = str[index];
+			} else if(0u < count && ' ' != str[count-1u] && '\t' != str[count-1u]) {
```
SirCmpwn: Add spaces around mathematical operators

common/stringop.c
```diff
@@ -10,22 +10,32 @@
...
+				str[count++] = ' ';
+			}
+		}
+		str[count] = '\0';
+		if (0u < count && (' ' == str[count-1u] || '\t' == str[count-1u])) {
+			str[count-1u] = '\0';
+		}
```
SirCmpwn: This is a better approach overall, though, +1

common/stringop.c
```diff
@@ -283,6 +293,24 @@ static bool has_whitespace(const char *str) {
...
 
+
+/*
+ * Returns true if the string contains only whitespace and false otherwise.
+ */
+bool is_empty(const char *str) {
+	bool ret = true;
+	if (str) {
+		while (*str != '\0') {
+			if (!isspace(*(const unsigned char*)str)) {
```
SirCmpwn: Does isspace(*str) not just work?

common/stringop.c
```diff
@@ -283,6 +293,24 @@ static bool has_whitespace(const char *str) {
...
+				ret = false;
+				break;
```
SirCmpwn: There's no need for the ret variable to exist. During this branch you can just return false directly, and return true at the end, and eliminate the variable.

common/stringop.c
```diff
@@ -10,22 +10,32 @@
...
+ *
+ * \param		str	Pointer to the char string that should be stripped. The string
+ *							is modified. If NULL is passed, no operation is performed.
+ *
+ ******************************************************************************/
+void strip_whitespace(char * restrict const str)
+{
+	if (NULL != str) {
+		unsigned int count = 0u;
+		for (unsigned int index = 0u; str[index]; ++index) {
```
SirCmpwn: Use size_t for the index and count, please.

sway/config.c
```diff
@@ -620,23 +621,127 @@ bool load_include_configs(const char *path, struct sway_config *config) {
...
+*	happens to be on the next line.
+*
+*	\param[in]	file The pointer to the already open file.
+* \param[in/out]	line The current line read from the configuration file.
+* \param[in]	line_number Counter for the line number. Optional.
+*
+*	\return	A string to the corrected 'line' is returned.
+*
+*******************************************************************************/
+static inline char* handle_parentheses(FILE* const file, char* line,
```
SirCmpwn: Technically the word is braces, not parenthesis. () are parenthesis, {} are braces.

sway/config.c
```diff
@@ -620,23 +621,127 @@ bool load_include_configs(const char *path, struct sway_config *config) {
...
+					} else {
+						sway_log(L_ERROR, "Error: get_config_line(): Allocation of %zu bytes "
+							"was not successful.", new_size);
+					}
+				}
+			}
+		}
+	}
+	return line;
+}
```
SirCmpwn: To be honest, rather than take this approach, I would prefer to make the whole config parser character oriented instead of line oriented, and make it a state machine, like a more typical parser would look. It would be a bit more complicated, but less sloppy and more things would work. It's my fault for not designing the config parser like this to begin with, for the record.

@minus7 might be interested in making that change. You can even make it syncronous if you like!

common/readline.c
DarkDeepBlue: Removing the comments is fine with me.

After the discussion on the IRC: Should I use `getline()` here or not?

SirCmpwn: Yeah getline is fine

common/readline.c
DarkDeepBlue: Should I read it as: "Do not mark unsigned integers explicitely"?

common/stringop.c
DarkDeepBlue: I want to avoid `isspace()` being called with a negative integer.

common/readline.c
SirCmpwn: Yes.

common/stringop.c
SirCmpwn: Better to do (unsigned char)*str, then

common/stringop.c
DarkDeepBlue: I write my functions with one single exit.
Should I change that for your code base?

sway/config.c
DarkDeepBlue: Yes, good point.

common/stringop.c
SirCmpwn: You don't need to change it everywhere, but going forward do not use that principle.

sway/config.c
DarkDeepBlue: To be honest, the code is too complicated and too buggy. I see it the exact same way. The whole thing needs to be rewritten.

common/stringop.c
minus7: Might be nice to run clang-tidy/clang-format or some other formatter/format verification tool over commits automatically?

JerziKaminsky:
@SirCmpwn, [lemon](http://www.hwaci.com/sw/lemon/)? [Ragel](https://en.wikipedia.org/wiki/Ragel)? I might go for this, If it isn't too yaccy.

SirCmpwn:
I don't think we need a parser generator for a grammar this simple.

PR #1040, opened 2017-01-10
Add template for issues

DarkDeepBlue:
Currently issues are written relying on the user providing all the necessary data.

With an issue template it is ensured that all the data of interest is actually available right from the start. The risk of the user not providing necessary information is reduced.


SirCmpwn:
No thanks.

PR #1041, opened 2017-01-11
add mouse_scroll_workspaces bar setting

zandrmartin:


acrisci:
The i3 way to do this is

```
bar {
   bindsym button4 nop
}
```

zandrmartin:
We don't have a way to do `bindsym` within a `bar {}` block (AFAIK) but I'm open to that behavior. I don't think Swaybar even processes keyboard events currently.

acrisci:
i3bar doesn't listen to keyboard events either and probably never will. This is just the config approach we took for this behavior because it is more extensible to other use cases, but I don't think this feature is very widely used because it's kind of buried in the docs.

This approach is much more semantic and easier to find. However, this feature is not requested very much because putting the bar on the top is a much better solution to this problem.

SirCmpwn:
How does putting the bar on top solve this problem at all?

acrisci:
This complaint arises only when the user accidentally scrolls their workspaces at some time. This accidental scrolling only has one cause that I am aware of.

The user is on a web page and scrolling through content with the scroll wheel. They click something on the web page and then move the mouse down to get it out of the way of the text, but they don't keep track of exactly how far they move the mouse down. Sometimes, they move the mouse all the way down to the bottom of the screen, which makes sense because the web browser and the bar share a lower boundary. Then after they read some more and it's time to scroll down again, they grab the mouse and use the scroll down button. This action itself naturally moves the cursor even lower on the screen. After you scroll down on the page enough, the cursor will get lower and lower until it finally is on the bar. This is when the accidental scroll happens.

The reason why putting the bar on the top is this:

* For some reason, people will never move the mouse *up* to get the pointer out of the way of the text.
* People do not scroll up on their text for very long periods of time on a single page which is the main source of the cursor drifting.

zandrmartin:
I keep the bar at the top. I tend to move my mouse up to get it out of the way. So moving it to the top doesn't help.

acrisci:
> Maybe the user just wants that feature, ya?

Yes, absolutely. That's why I implemented that way to disable the scrolling.

> I keep the bar at the top. I tend to move my mouse up to get it out of the way. So moving it to the top doesn't help.

Implementing a feature where you can put the bar on the side of the screen (often requested in i3, but difficult to implement) might be an even better solution to your problem.

zandrmartin:
Except that I like the bar at the top, I just don't want it to move workspaces when I accidentally scroll.

acrisci:
Just to be clear, I'm not opposing this PR at all. I've dealt with the angry mob of accidental scrollers before and I'm just providing you with context for how we solved the issue in i3 and why we did it that way.

SirCmpwn:
If we implemented it the way @zandrmartin has proposed (with bindsym for compatability reasons) what are the odds that a similar feature would make its way into i3?

acrisci:
They don't like multiple ways of doing the same thing in the config so you have that against you. They hate mouse features, but this is disabling a mouse feature, so that's a wild card. You'll have to ask. I'd say it's 50/50.

PR #1043, opened 2017-01-12
Keep CAP_SYS_PTRACE with suid binary

Hummer12007:
This fixes sway when run with suid binary. (suid does not work with filecaps in an intuitive manner, so breakage occurs).

Note: suid binary will not require CAP_SYS_PTRACE.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -331,6 +331,7 @@ int main(int argc, char **argv) {
 		// Drop every cap except CAP_SYS_PTRACE
 		cap_t caps = cap_init();
 		cap_value_t keep = CAP_SYS_PTRACE;
+		sway_log(L_INFO, "Dropping extra capabilities");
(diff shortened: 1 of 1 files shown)


sway/main.c
```diff
@@ -311,6 +326,20 @@ int main(int argc, char **argv) {
 	}
 	register_extensions();
 
+#ifdef __linux__
+	if (suid) {
+		// Drop every cap except CAP_SYS_PTRACE
+		cap_t caps = cap_init();
+		cap_value_t keep = CAP_SYS_PTRACE;
```
SirCmpwn: Don't we also want that TTY one?

Hummer12007: No, it was required because a suid binary would not gain all caps as effective, and besides, all the privileged stuff is dealt with in a child process of wlc (forked in wlc_init, so it keeps everything), so it isn't.

Issue #733, opened 2016-07-04
"output right" doesn't follow i3's behaviour


regnat:
On i3, using "focus output right" while on the outer right monitor focuses the outer left monitor (and the same for "output {left,above,below}" of course).
On sway it does nothing, and the ipc returns (when trying `swaymsg focus right` on the cli) :

``` json
{ "success": false, "input": "focus output", "error": "Can't find output with name\/at direction 'right' @ (2240,885)" }
```

EDIT: It may be not clear on the message, but of course, this isn't specific to the "focus" command : others commands like "move container to" or "move workspace to" follow the same behaviour.

acrisci:
I can confirm this.

PR #1048, opened 2017-01-14
Add output wrapping

dkess:
This fixes issue #733 . Now if the user focuses output right but is at
the rightmost monitor, the focus will wrap the the leftmost monitor
(or whatever direction was chosen).

This commit adds a new function, swayc_opposite_output, which selects
the opposite output given a position and a direction. Now, when calling
output_by_name, we first check if there is an adjacent output to switch
to. If that fails, we call swayc_opposite_output to handle wrapping.


SirCmpwn:
Thanks!

PR #1051, opened 2017-01-17
add hide_edge_borders smart to sway.5 manual

ametisf:
I forgot to add the option to manual page, so here is the fix.


SirCmpwn:
Thanks!

PR #1052, opened 2017-01-19
Add window instance support

Hummer12007:
At the very least, get_tree output and instance criteria work and are tested.

Whether or not to include the second commit is to your discretion.


SirCmpwn:
Thanks!

PR #1080, opened 2017-02-19
Revise IPC security configuration

SirCmpwn:
This only contains changes to the docs and default config. Implementation to come later. Opening the pull request now to gather feedback on the proposed changes.


SirCmpwn:
Basically, this changes IPC policies to be per-executable (based on the process that's connecting to the socket). It also moves the configuration to `/etc/sway/security.d/00-defaults`. Users will be discouraged from touching this file so we can continue to update it with the latest recommended default security policy between releases, and they can override it by writing their own files in `/etc/sway/security.d`. Third party packages that integrate with sway and need special permissions (third party bars, lock screens, etc) can also ship a file for this directory to configure their permissions.

Hummer12007:
One thing bothered me, when I took a glance: this will probably break clients on freebsd. A default policy to allow everything may be required there.

SirCmpwn:
FreeBSD will have to take care of that themselves, presumably by shipping some rules in their package.

PR #1092, opened 2017-02-28
i3 feature support: Moving flotaing containers

4e554c4c:
This commit lets the 'move' command apply to floating containers as well
as tiled ones. The command may be appended with a number of pixels and
then the string `px` (like '10 px') in order to move the container more
or fewer than the standard ten pixels.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -9,30 +9,40 @@
 
 struct cmd_results *cmd_move(int argc, char **argv) {
 	struct cmd_results *error = NULL;
+	int move_amt = 10;
+
 	if (config->reading) return cmd_results_new(CMD_FAILURE, "move", "Can't be used in config file.");
 	if ((error = checkarg(argc, "move", EXPECTED_AT_LEAST, 1))) {
 		return error;
 	}
-	const char* expected_syntax = "Expected 'move <left|right|up|down|next|prev|first>' or "
+	const char* expected_syntax = "Expected 'move <left|right|up|down|next|prev|first> <[px] px>' or "
 		"'move <container|window> to workspace <name>' or "
 		"'move <container|window|workspace> to output <name|direction>' or "
 		"'move position mouse'";
 	swayc_t *view = get_focused_container(&root_container);
 
+	if (argc == 2 || (argc == 3 && strcasecmp(argv[2], "px") == 0 )) {
+		char *inv;
+		move_amt = (int)strtol(argv[1], &inv, 10);
+		if (*inv != '\0' && strcasecmp(inv, "px") != 0) {
+			move_amt = 10;
+		}
+	}
+
 	if (strcasecmp(argv[0], "left") == 0) {
-		move_container(view, MOVE_LEFT);
+		move_container(view, MOVE_LEFT, move_amt);
 	} else if (strcasecmp(argv[0], "right") == 0) {
-		move_container(view, MOVE_RIGHT);
+		move_container(view, MOVE_RIGHT, move_amt);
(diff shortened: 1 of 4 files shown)


4e554c4c:
I know that sway does not use macros, but I saw that `MAX` and `MIN` were used across layout.c already and decided to use them. Please tell me if this should be changed.

SirCmpwn:
Macros defined by standard headers are fine.

sway/commands/move.c
```diff
@@ -9,30 +9,36 @@
...
 	}
-	const char* expected_syntax = "Expected 'move <left|right|up|down|next|prev|first>' or "
+	const char* expected_syntax = "Expected 'move <left|right|up|down|next|prev|first> <[px] px>' or "
 		"'move <container|window> to workspace <name>' or "
 		"'move <container|window|workspace> to output <name|direction>' or "
 		"'move position mouse'";
 	swayc_t *view = get_focused_container(&root_container);
 
+	if (argc == 3 && strcasecmp(argv[2], "px") == 0 ) {
+		move_len = (int)strtol(argv[1], NULL, 10);
```
SirCmpwn: I would prefer to call this move_amt. Not really a length.

SirCmpwn:
Does i3 support `move left 10px`?

SirCmpwn:
How about `move left 10`?

sway/sway.5.txt
```diff
@@ -100,11 +100,12 @@ They are expected to be used with **bindsym** or at runtime through **swaymsg**(
 **layout** toggle split::
 	Cycles between available split layouts.
 
-**move** <left|right|up|down|next|prev|first>::
-	Moves the focused container _left_, _right_, _up_, or _down_. Moving to _prev_
-	or _next_ swaps the container with its sibling in the same container. Move
-	_first_ exchanges the focused element in an auto layout with the first
-	element, i.e. promotes  the focused element to master position.
+**move** <left|right|up|down|next|prev|first> <[px] px>::
```
SirCmpwn: Since `move prev` et al don't work with pixel arguments, split them into a separate doc thing.

4e554c4c:
Yes, i3 allows both `move left 10px` and `move left 10`. This has been ammended.

sway/sway.5.txt
```diff
@@ -100,11 +100,14 @@ They are expected to be used with **bindsym** or at runtime through **swaymsg**(
...
 	Cycles between available split layouts.
 
-**move** <left|right|up|down|next|prev|first>::
-	Moves the focused container _left_, _right_, _up_, or _down_. Moving to _prev_
-	or _next_ swaps the container with its sibling in the same container. Move
-	_first_ exchanges the focused element in an auto layout with the first
-	element, i.e. promotes  the focused element to master position.
+**move** <left|right|up|down> <[px] px>::
+	Moves the focused container _left_, _right_, _up_, or _down_. If the window
+	is floating it moves it _px_ in that direction.
```
SirCmpwn: Can you clarify that px is optional and changes the behavior? Maybe this should be split again into two sections. Current version is unclear about the distinction between `move left` and `move left 10px`

4e554c4c: Sure, I can see how this needs clarification. Both `move left` and `move left 10px` affect both floating and tiled containers. Why should they be split into different sections?

SirCmpwn: Just proposing that two sections might make sense. If you can sufficiently clarify it without that's fine.

SirCmpwn:
Thanks!

PR #1101, opened 2017-03-01
Fix #1099: Allow spaces in worspace names

4e554c4c:
This commit allows unquoted spaces in worspace names in order to keep
compatability with i3. The names _must not_ contain the string "output"
which is documented in 'sway.5' because how sway detects the `move
<workspace> output <output>` command. Also I documented that "number"
may be used before the worspace name without affecting how the name is
evaluated.

PR #1105, opened 2017-03-08
fix workspace output assignment

zandrmartin:
I don't know if this is going to break any of the functionality introduced in #1101 but it does make workspace -> output assignment work again.


SirCmpwn:
I'd rather break #1101 than let this stay broken. Will leave it up to @4e554c4c to sort it out.

4e554c4c:
I don't believe that this will break anything, it was just a bug in my pr. Thanks for catching this @zandrmartin.

PR #1115, opened 2017-03-14
Superfluous include causes failures if WLC headers aren't installed globally

snoack:
If the WLC headers aren't installed globally (e.g. somewhere in your home directory rather than in `/usr/includes`), despite `-DWLC_INCLUDE_DIRS` being passed correctly to `cmake`, `make` fails.

Apparently, the `CMakeList.txt` in some subdirectories are missing `WLC_INCLUDE_DIRS` in their `include_directories()` directive.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -3,7 +3,6 @@
 
 #include <stdint.h>
 #include <unistd.h>
-#include <wlc/wlc.h>
 #include <xkbcommon/xkbcommon.h>
 
 /**


SirCmpwn:
None of these subdirectories use WLC...

snoack:
Then why do I get following error, unless I add `WLC_INCLUDE_DIRS` to `swaybg/CMakeList.txt`?

```
[ 89%] Building C object swaybg/CMakeFiles/swaybg.dir/main.c.o
In file included from /tmp/sway/swaybg/main.c:14:0:
/tmp/sway/include/util.h:6:21: fatal error: wlc/wlc.h: No such file or directory
 #include <wlc/wlc.h>
                     ^
compilation terminated.
swaybg/CMakeFiles/swaybg.dir/build.make:62: recipe for target 'swaybg/CMakeFiles/swaybg.dir/main.c.o' failed
make[2]: *** [swaybg/CMakeFiles/swaybg.dir/main.c.o] Error 1
CMakeFiles/Makefile2:625: recipe for target 'swaybg/CMakeFiles/swaybg.dir/all' failed
make[1]: *** [swaybg/CMakeFiles/swaybg.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

(It seems that code in these subdirectories includes headers that in return include WLC headers.)

SirCmpwn:
The actual problem then seems to be that "include/util.h" includes wlc/wlc.h without actually needing it.

snoack:
If that is the case, feel free to remove the redundant includes. That should work as well.

SirCmpwn:
I'm suggesting your remove them and update your pull request.

snoack:
There you go.

SirCmpwn:
Thanks!

PR #1117, opened 2017-03-16
Allow also 444 for security file mode

jnsaff:
Allow also 444 as security file mode in addition to 644.

This helps with systems like NixOS where nothing in /nix/store is writable to anyone.


SirCmpwn:
Thanks!

PR #1126, opened 2017-03-18
prevent fullscreen focus thievery

### LINKED ISSUE #1102: Opening windows while another is fullscreen gives them focus
And it shouldn't.

zandrmartin:
I believe this fixes #1102


SirCmpwn:
Thanks!

PR #1147, opened 2017-04-03
Downgrade nvidia proprietary driver warning

SirCmpwn:
This is shipping no earlier than 0.13 and not unless I hear positive feedback from nvidia users about the new EGL support in wlc.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -63,8 +63,7 @@ void detect_proprietary() {
 			break;
 		}
 		if (strstr(line, "nvidia")) {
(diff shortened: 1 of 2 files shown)


soredake:
Quicky tested, browser works (but popups bug still present), mpv segfaults on most backends.

4e554c4c:
I've been using the proprietary driver for a few weeks and I haven't seen any problems yet. Having to enable the experimental DRM kms driver is annoying, but that's pretty much it. I'd give this a thumbs up.

PR #1149, opened 2017-04-05
Improve criteria handling

4e554c4c:
This commit changes how commands decide what container to act on.
Commands get the current container though `current_container`, a global
defined in sway/commands.c. If a criteria is given before a command,
then the following command will be run once for every container the
criteria matches with a reference to the matching container in
'current_container'. Commands should use this instead of
`get_focused_container()` from now on.

This commit also fixes a few (minor) mistakes made in implementing marks
such as non-escaped arrows in sway(5) and calling the "mark" command
"floating" by accident. It also cleans up `criteria.c` in a few places.

I have tested the changed commands and things seem to work.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -43,6 +43,8 @@ struct cmd_handler {
 
 int sp_index = 0;
 
+swayc_t *current_container = NULL;
+
 // Returns error object, or NULL if check succeeds.
 struct cmd_results *checkarg(int argc, const char *name, enum expected_args type, int val) {
 	struct cmd_results *error = NULL;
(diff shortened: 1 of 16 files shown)


SirCmpwn:
I was just thinking about this the other day, thanks for working on it!

sway/commands.c
```diff
@@ -43,6 +43,8 @@ struct cmd_handler {
 
 int sp_index = 0;
 
+swayc_t *current_container;
```
SirCmpwn: Initialize to NULL

sway/criteria.c
```diff
@@ -373,3 +373,25 @@ list_t *criteria_for(swayc_t *cont) {
 	}
 	return matches;
 }
+
+void container_for(swayc_t *container, list_t *tokens, list_t *list) {
```
SirCmpwn: Use container_map instead

sway/sway.5.txt
```diff
@@ -426,6 +426,20 @@ The string contains one or more (space separated) attribute/value pairs and they
 are used by some commands filter which views to execute actions on. All attributes
 must match for the criteria string to match.
 
+Criteria may be used with either the **for_window** or **assign** commands to
+specify operations to perform on new views. A criteria may also be used to
+perform specific commands (ones that normally act upon one window) on all views
+that match that criteria. Here are some ways to use criteria in this way.
```
SirCmpwn: s/Here are some ways to use criteria in this way/For example:/

SirCmpwn:
The switch to current_container is a good idea, but it almost certain to create some bugs. Please carefully check to make sure you've found every case.

Looks good overall. Let me know when you've fixed the issues you mentioned and I'll take a closer look.

sway/criteria.c
4e554c4c: I wanted to use container map, but that only takes two arguments when I would need three: tokens, container and final list. I could solve that by passing a (container list,tokens list) struct or use a global, but both felt somewhat jurry-rigged.

SirCmpwn: Use a struct.

4e554c4c:
I've gone back over the code and fixed a few things. Turns out the "title matching bugs" was just a consequence of how I was testing with swaymsg. One of the things that might not be 100% wanted is moving (non-flotaing)containers always focuses them, I can work on this if you'd like.

SirCmpwn:
Please do, yes. But we can address that separately. Thanks!

PR #1150, opened 2017-04-06
Add libcap check to CMake

JerziKaminsky:
- Moved <sys/capability.h> include inside `__linux__` guard, because all uses are similarly guarded.
- <sys/capability.h> is a part of optional devel package on at least fedora, so CMake now explicitly checks that libcap devel files are available.
- Added libcap to list of arch packages in .travis.yml, to make the dependency explicit. travis-ci installs the package by default, which is why this hasn't surfaced previously).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ arch:
     - cairo
     - gdk-pixbuf2
     - wlc-git
+    - libcap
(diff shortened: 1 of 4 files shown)


SirCmpwn:
Thanks!

PR #1151, opened 2017-04-06
Change regex to use PCRE

ascent12:
This changes the regex matching for criterias to use PCRE instead of POSIX. i3 uses PCRE, so this would be increase compatibility with it.
PCRE is already pulled in as a dependency so I don't think this will cause any issues.


SirCmpwn:
Thanks!

PR #1154, opened 2017-04-07
Removed explicitly setting file descriptors to -1

ascent12:
Doing this before calling ipc_client_disconnect is calling file descriptors to leak.


SirCmpwn:
Thanks!

Issue #98, opened 2015-08-20
IPC feature support


minus7:
Message types and their implementation status:
- [x] command
- [x] subscribe
  - [x] workspace
  - [x] barconfig_update
  - [x] output
  - [x] mode
  - [x] window
  - [x] binding
- [x] get_workspaces
- [x] get_outputs
- [x] get_tree
- [x] get_marks
- [x] get_bar_config
- [x] get_version -- somewhat done, needs better version numbers
- [x] json string encoding (quotation marks, utf-8) -- using json-c

PR #1155, opened 2017-04-07
Add `-t get_marks` and use more i3-like marks

4e554c4c:
In i3 every mark is unique and one mark cannot be used in more than one
window, sway behavior has been amended to match this.
`swaymsg -t get_marks` will now return an array of all marks used in sway.

See #98


SirCmpwn:
Thanks!

PR #1169, opened 2017-04-12
Bar Hiding

calvinkosmatka:
Hi - new to this project. I'm switching over from i3 and noticed this was missing, so I went ahead and added it.

I did it by allowing swaybar to recieve modifier presses and releases over ipc, and handling that appropriately on the swaybar side of things. Maybe it would make more sense to handle show/hide on the sway side then push that to the bar over ipc with a barconfig update. Now that I think about it, that approach could be used to unify show/hide and binding mode updates...

One issue I ran into is that bar display mode and key binding mode were being conflated, so I added display_mode in the bar configs to distinguish, and left mode for binding modes.

This still isn't exactly like it is in i3, the majorgit fetch upstream difference being that there is always some space allocated to the bar, rather than having it display over the top of the workspace. I'm not entirely sure how to resolve this. It also doesn't fade in or out.

What do you think?


swaybar/ipc.c
```diff
@@ -322,8 +335,12 @@ void ipc_bar_init(struct bar *bar, const char *bar_id) {
...
 	free(res);
 	json_object_put(outputs);
-
-	const char *subscribe_json = "[ \"workspace\", \"mode\" ]";
+	// Will need to be redone for display mode toggling compatibility
+	// Will require setting bar->config->hidden_state to "show" as default
+	const char *subscribe_json = strcmp(bar->config->display_mode, "hide") == 0 ?
+		"[ \"workspace\", \"mode\", \"modifier\" ]"
+		: "[ \"workspace\", \"mode\" ]";
+	sway_log(L_ERROR, subscribe_json);
```
Hummer12007: sway_log(L_ERROR, "%s", subscribe_json);

swaybar/render.c
```diff
@@ -286,59 +286,67 @@ void render(struct output *output, struct config *config, struct status_line *li
...
-		double pos = (window->width * window->scale) - 0.5;
-		bool edge = true;
-		for (i = line->block_line->length - 1; i >= 0; --i) {
-			struct status_block *block = line->block_line->items[i];
-			if (block->full_text && block->full_text[0]) {
-				render_block(window, config, block, &pos, edge, is_focused);
-				edge = false;
+	// Could also be done by always rendering then conditionally clearing and drawing alpha
+	// That may be preferable down the line
+	if ((!strcmp(config->display_mode, "hide")) == 0 || strcmp(config->hidden_state, "show") == 0) {
```
Hummer12007: Isn't the first clause implied by the second one?

swaybar/ipc.c
calvinkosmatka: Thanks - meant to take that out

swaybar/render.c
calvinkosmatka: No - in display_mode "hide" when the modifier is held down, hidden_state changes to "show", so the first clause will evaluate to false, while the second evaluates to true.
We could (and potentially should) make it so that we only need the second clause. I added the first because I think hidden_state defaults to "hide" but we want the bar to show anyway as long as the display_mode isn't "hide"

SirCmpwn:
Nice work!

>This still isn't exactly like it is in i3, the majorgit fetch upstream difference being that there is always some space allocated to the bar, rather than having it display over the top of the workspace. I'm not entirely sure how to resolve this. It also doesn't fade in or out.

Panel layout handling is a bit complicated. The way the bar is positioned is via the desktop-shell extension, which allows the swaybar client to indicate that it is a desktop panel. Swaybar first becomes a panel, then positions itself on screen (top, bottom, left, right) [here](https://github.com/SirCmpwn/sway/blob/master/swaybar/bar.c#L162-L165). Sway handles the first request [here](https://github.com/SirCmpwn/sway/blob/master/sway/extensions.c#L113-L134), and the second request [here](https://github.com/SirCmpwn/sway/blob/master/sway/extensions.c#L152-L163). When it calls `arrange_windows`, the panel is positioned on screen and windows are offset to avoid occulding that space (this is the behavior that you need to change).

Panel geometry is calculated by [this function](https://github.com/SirCmpwn/sway/blob/master/sway/handlers.c#L59-L93) and applied by [these functions](https://github.com/SirCmpwn/sway/blob/master/sway/handlers.c#L95-L107), which are invoked upon *creation* of the panel view, which happens [here](https://github.com/SirCmpwn/sway/blob/master/sway/handlers.c#L352-L361). Currently this is the only place the panel geometry is assigned and it's left that way for the duration of the session. Panels are not added to the main layout tree, so they aren't rearranged during `arrange_windows`.

However, `arrange_windows` *is* responsible for moving other windows out of the panel's way. Basically, `arrange_windows_r` is a recursive function that does most of the work in calculating the tree's on-screen layout. It's recursively invoked for each child of each container, and is called with the space that child has been allocated within the parent. The child then arranges itself within this space. For example, L_HORIZ is a horizontal split container so it will allocate its children areas tiled horizontally and recurse into them for each such area. Views will simply assign the wayland window geometry to the allocated space. Each kind of layout will handle this differently. Accomodating for panels is done in the output layout [here](https://github.com/SirCmpwn/sway/blob/master/sway/layout.c#L1004-L1024). This just changes the geometry allocated to the output's children before recursing. Makes sense?

The solution is probably to switch from having swaybar use the desktop-shell extension to become a panel and instead using a custom wayland extension. This extension should probably have three functions:

* Indicate that a surface should become a panel
* Indicate where to position the panel (top, bottom, left, right)
* Toggle the panel's visibility

Simply making the panel transparent isn't going to cut it once you start placing it on top of other windows, because it will still receive mouse events. You need to have sway make it disappear in the compositor - Sway has a visibility toggle function somewhere that you can use. The process for making a wayland protocol extension is fairly straightforward:

1. Add an XML file describing your protocol [to this directory](https://github.com/SirCmpwn/sway/tree/master/protocols)
2. Add it to [the relevant CMakeListst.txt](https://github.com/SirCmpwn/sway/blob/master/protocols/CMakeLists.txt)
3. Create a compositor [implementation](https://github.com/SirCmpwn/sway/blob/master/sway/extensions.c#L278-L282) of the interface in extensions.c and [register a global for it](https://github.com/SirCmpwn/sway/blob/master/sway/extensions.c#L329)

This whole process might seem intimidating, but it should be easier than you think.

include/sway/config.h
```diff
@@ -110,12 +110,19 @@ struct bar_config {
 	 * Always visible in dock mode. Visible only when modifier key is held in hide mode.
 	 * Never visible in invisible mode.
 	 */
-	char *mode;
+	char *display_mode;
```
SirCmpwn: Can you make this an enum?

include/sway/config.h
```diff
@@ -110,12 +110,19 @@ struct bar_config {
...
 	 * One of "show" or "hide".
 	 *
 	 * In "show" mode, it will always be shown on top of the active workspace.
 	 */
+	char *mode;
+	/**
+	 * Current keybinding mode
+	 *
+	 * Names should probably be reworked at some point, but I'm keeping it for now
+	 * Don't think this does anything at the moment
```
SirCmpwn: If the mode is set, it renders this in the bar.

include/sway/config.h
SirCmpwn: Sorry, nevermind. In the compositor it's fine to keep this a string. We'll want an enum on the swaybar side, though.

include/sway/config.h
SirCmpwn: Yeah this can probably be removed

include/swaybar/config.h
```diff
@@ -26,6 +26,8 @@ struct config {
 	char *font;
 	char *sep_symbol;
 	char *mode;
+	char *display_mode;
+	char *hidden_state;
```
SirCmpwn: This is the one I'd rather have as an enum. Two enums.

swaybar/ipc.c
```diff
@@ -322,8 +335,11 @@ void ipc_bar_init(struct bar *bar, const char *bar_id) {
 	}
 	free(res);
 	json_object_put(outputs);
-
-	const char *subscribe_json = "[ \"workspace\", \"mode\" ]";
+	// Will need to be redone for display mode toggling compatibility
+	// Will require setting bar->config->hidden_state to "show" as default
+	const char *subscribe_json = strcmp(bar->config->display_mode, "hide") == 0 ?
+		"[ \"workspace\", \"mode\", \"modifier\" ]"
+		: "[ \"workspace\", \"mode\" ]";
```
SirCmpwn: Can you make a JSON object here and populate it instead of using string constants?

calvinkosmatka:
Thanks!

> Simply making the panel transparent isn't going to cut it once you start placing it on top of other windows, because it will still receive mouse events. You need to have sway make it disappear in the compositor
 
Hadn't even thought of that. In i3, I have a gap around the edge and the bar stays within that, so that didn't occur to me as a problem.

Would it be possible/easier to implement the bar as a floating window that moves across workspaces rather than writing a wayland extension?

include/swaybar/config.h
calvinkosmatka: Sounds good.

SirCmpwn:
>Would it be possible/easier to implement the bar as a floating window that moves across workspaces rather than writing a wayland extension?

No,I don't think so.

include/sway/config.h
calvinkosmatka: It looked like binding modes are passed directly to the bar, so this never gets changed. If instead we wanted to pass entire bar configs to the bar, we would need to handle that bar's binding modes on the sway side, so this would be necessary. In theory this would have the advantage of keeping sway up to date about the bar's state without requiring swaybar to send its state over ipc.

calvinkosmatka:
> No,I don't think so.

Alright - time for me to start learning about wayland extensions then.

swaybar/render.c
Hummer12007: Sorry, got confused. It's different fields.

calvinkosmatka: Oh yeah, no worries. Good reason to make those into enums.

JerziKaminsky:
@calvinkosmatka , thanks for working on this. I was just about to open my own PR for this when I saw you beat me to it. Looking at the code was a bit uncanny, some telepathy might have been involved :)

include/sway/config.h
JerziKaminsky: Would it make more sense to stick with `mode` as the varname for `mode` config keyword? that seems like how the rest of the vars work. 

binding mode is not really a configuration option for bar and there's no config word for it, it's just a label published by sway itself for display. Just incidental that its stored in bar_config, because it doubles as a state object as well as a config store.

Would also make the patch smaller.

include/sway/config.h
```diff
@@ -110,12 +110,19 @@ struct bar_config {
...
+	 */
 	char *hidden_state;
```
JerziKaminsky: `hidden_state` is naturally a bool value, and as its a state variable rather than a storing config option value maybe it's better to change its type? Also avoid redundant strcmp in the code.

swaybar/ipc.c
JerziKaminsky: Alternative: always sending modifier events and ignore them.
If the user doesn't configure a bar modifier key, they won't be sent anyway, I think.

swaybar/ipc.c
```diff
@@ -19,13 +20,15 @@ void ipc_send_workspace_command(const char *workspace_name) {
...
 	json_object *font, *bar_height, *wrap_scroll, *workspace_buttons, *strip_workspace_numbers;
 	json_object *binding_mode_indicator, *verbose, *colors, *sep_symbol, *outputs;
 	json_object *markup;
 	json_object_object_get_ex(bar_config, "tray_output", &tray_output);
-	json_object_object_get_ex(bar_config, "mode", &mode);
-	json_object_object_get_ex(bar_config, "hidden_bar", &hidden_bar);
+	// "mode" as in the bar's display mode and "mode" as in binding mode
+	// should be distinguished more carefully
+	json_object_object_get_ex(bar_config, "mode", &display_mode);
+	json_object_object_get_ex(bar_config, "hidden_state", &hidden_state);
```
JerziKaminsky: IIUC, this upgrades `hidden_state` from an internal state variable into a config option, and I'm not sure I understand why. `hidden_state` is just used a state variable for "modifier pressed" , isn't it? maybe I'm missing something.

swaybar/ipc.c
```diff
@@ -44,6 +47,16 @@ static void ipc_parse_config(struct config *config, const char *payload) {
 		free(config->status_command);
 		config->status_command = strdup(json_object_get_string(status_command));
 	}
+	
+	if (display_mode) {
+		free(config->display_mode);
+		config->display_mode = strdup(json_object_get_string(display_mode));
+	}
```
JerziKaminsky: In my attempt at a PR for this, `hidden_state` was always used as a sole state variable for hide/unhiding the bar. if the mode isn't `hide`, it simply always holds a "visible" value. If you want to go that route (up to you of course), then a site like this where the display mode value is set is where the default value for `hidden_state` should also be set.

swaybar/ipc.c
```diff
@@ -364,6 +380,30 @@ bool handle_ipc_event(struct bar *bar) {
...
+			sway_log(L_ERROR, "failed to parse payload as json");
+			return false;
+		}
+		json_object *json_change;
+		if (json_object_object_get_ex(result, "change", &json_change)) {
+			const char *change = json_object_get_string(json_change);
+			free(bar->config->hidden_state);
+			if (strcmp(change, "pressed") == 0) {
+				bar->config->hidden_state = strdup("show");
+			} else { //must be "released"
```
JerziKaminsky: defensive: assert instead of comment. Check for expected value and log unexpected values with `L_ERROR` instead?

JerziKaminsky:
> The solution is probably to switch from having swaybar use the desktop-shell extension to become a panel and instead using a custom wayland extension. This extension should probably have three functions:

Would it makes sense to also contact wayland upstream to try and get a compatible feature into a new version of the desktop-shell protocol? this isn't sway-specific, and although wayland protocols are great way for extending functionality, it's also a big risk for fragmentation when it comes to common functionality being exposed as a custom protocol for every wayland compositor. Someone will want to create a dock application that talks to more than one compositor.

> Simply making the panel transparent isn't going to cut it once you start placing it on top of other 
> windows, because it will still receive mouse events. 

I'm also very new to wayland, but isn't that what input regions are for?

```
A.14.1.6. wl_surface::set_input_region - set input region
This request sets the region of the surface that can receive pointer and touch events.
Input events happening outside of this region will **try the next surface in the server surface stack**. The compositor ignores the parts of the input region that fall outside of the surface.
```

I'm thinking of a more general case where a dock isn't all-opaque or all-transparent (hidden), but hovers over a window has see-through areas. Like some of the more baroque Enlightenment contraptions. So unless I misunderstood the spec, using input regions + transperancy would be more powerful and potentially require less new code.

include/sway/config.h
SirCmpwn: The binding mode can probably be removed from this struct entirely. I don't think it's used anywhere.

swaybar/ipc.c
SirCmpwn: Aye, should probably go. Unless the i3 protocol does this or something.

swaybar/ipc.c
SirCmpwn: We have a sway_assert for this purpose

SirCmpwn: But in this case w/e tbh

SirCmpwn:
>Would it makes sense to also contact wayland upstream to try and get a compatible feature into a new version of the desktop-shell protocol? this isn't sway-specific, and although wayland protocols are great way for extending functionality, it's also a big risk for fragmentation when it comes to common functionality being exposed as a custom protocol for every wayland compositor. Someone will want to create a dock application that talks to more than one compositor.

No. This is just a stop-gap measure until we implement the surface-layers protocol later on.

>I'm also very new to wayland, but isn't that what input regions are for?

Fair point, but I'm not sure we have support for that.

JerziKaminsky:
> This is just a stop-gap measure until we implement the surface-layers protocol later on.

[postponed indefinitely](https://github.com/SirCmpwn/sway/issues/1027#issuecomment-279834595). Also the solution you described is basically an extension that does desktop-shell(::panel) + a floating hint. So it's not crazy to just get floating into the desktop-shell protocol.

SirCmpwn:
postponed until we have wlroots, rather, which is coming soon. You're welcome to pitch this feature upstream to desktop-shell but don't hold your breath.

swaybar/ipc.c
calvinkosmatka: i3bar has this as a config option. Looks like it's used for an initial default state (i.e. to show bar on startup before modifier has been pressed/released) and possibly as target for bindsyms. Honestly, I had just updated the name from what was there previously, but it looks like it should stay.

SirCmpwn: Alright, let's leave it. i3 compatability is the most important factor here.

swaybar/ipc.c
calvinkosmatka: @JerziKaminsky I had considered doing it that way specifically for the reason that I commented - if the user wants to toggle the display mode between "hide" and "dock" the bar will need to be able to get modifier events even if it started as a "dock" bar. I'll go ahead and do that. Not entirely sure what governs which modifiers get sent when. In any case, subscribe and ignore will work for now.

SirCmpwn: Don't take that approach. This way the user can choose not to use this feature and adjust their security configuration accordingly.

calvinkosmatka: Ok. Display mode toggling isn't even in play right now anyway. Thinking ahead, the best way to implement that would be to subscribe bars to barconfig updates and send that ipc event from the sway side (as opposed to subscribing bars to bindings) right?

SirCmpwn: No, bar config updates isn't an event we currently have, and I would rather not have the compositor updating the bar config to send over the wire every time the mod key is depressed. Definitely use the modifier event.

calvinkosmatka: Sorry, I just mean for display mode toggling. In i3 toggling between "dock" and "hide" is bindable, so the bar needs to respond to that binding. So what you're saying is this should be done by subscribing the bar to binding events?

calvinkosmatka: Wait, ok I see code in commands/bar/mode.c for handling bar display mode commands that sends an ipc barconfig event but the event never makes it to the socket because there's no security mapping there. I guess the question is whether to make bars subscribed to barconfig or bindings as a whole to do display mode switching.

swaybar/ipc.c
JerziKaminsky: I'll yield to compatibility concerns.

... but it's only compatiblity with i3 being stupid here. Setting `mode=hide` + `hidden_state=show` appears to effectively disabling hiding altogether. Probably because (this PR does this too) the i3 event handler doesn't *really* invert the state in response to a toggle event but sets the state to hard-coded on/off values. The result is that `hidden_state=show` sets it to `on` when in the released state, and then pressed event switches it from `on->on`.

There's no reason for anyone to ever use this option in their config. It's there because remote control requires it to be an IPC command (not a config option) and my guess is i3 doesn't distinguish the two because it treats config as scripted IPC (which has its benefits). Strictly speaking, it should distinguish the two cases.

There **could** be a legitimate use case for this as a selector between modifier-hides-bar and modifier-unhides-bar behavior. Not something I Imagine ever using, and i3 doesn't use it for that.
Instead, it's more like an override for hide mode (hence bindable).

https://i3wm.org/docs/userguide.html, section 5.3
```
In order to control whether i3bar is hidden or shown in hide mode, 
there exists the hidden_state option, which has no effect in dock 
mode or invisible mode. It indicates the current hidden_state of the bar:

1. The bar acts like in normal hide mode, it is hidden and is only unhidden 
   in case of urgency hints or by pressing the modifier key (hide state), or 
2. it is drawn on top of the currently visible workspace (show state).

Like the mode, the hidden_state can also be controlled through i3, this can be 
done by using the bar hidden_state command.
```

swaybar/ipc.c
JerziKaminsky: > Don't take that approach. This way the user can choose not to use this feature and adjust their security configuration accordingly.

I don't think security mandates one approach over another, the client must be free to ask for whatever it likes, It's the server that enforces policy and the client should deal. I think (and so did @calvinkosmatka, independently) that having the client ask for everything it wants, but use only what the server gives corresponds perfectly to the security model and would simplify matters.

This raises another point about client handling of security denials. In #1172 I saw that the server declined IPC, and swaybar carried on as if nothing happened, which I think is a problem.

That suggests that maybe the client should not subscribe to all events at once, but group them by their "severity". Some events are optional (modifier events), and some may be critical (abort if not available). 

For example, being assigned mode `hide` and then denied modifier events is an error event IMO.

Sorry for causing so much discussion and of course I don't mean calvin has to deal with all the secondary issues himself simply because his PR exposed them. I'm glad to do a followup for the things calvin wants to consider out of scope here, assuming @SirCmpwn is interested.

I'll switch to quiet mode now... :)

SirCmpwn: >I don't think security mandates one approach over another, the client must be free to ask for whatever it likes, It's the server that enforces policy and the client should deal. I think (and so did @calvinkosmatka, independently) that having the client ask for everything it wants, but use only what the server gives corresponds perfectly to the security model and would simplify matters.

The server returns errors if you try to bind to events you don't have permission for. So either that needs to be handled with a retry mechanism that removes the unpermitted events, disables the relevant UI in the bar, and tries again. But I don't think that's important, can just ask for only what the user needs instead. We shouldn't be binding to events we aren't going to use anyway, if there's a bug that leads to that information being exposed from swaybar it reduces the impact of that vulnerability if it didn't bind to some events in the first place.

>This raises another point about client handling of security denials. In #1172 I saw that the server declined IPC, and swaybar carried on as if nothing happened, which I think is a problem.

Yes, the clients need to handle permissions errors more gracefully. I have done this for swaymsg but haven't gotten around to it for everyone else. Patches welcome

>That suggests that maybe the client should not subscribe to all events at once, but group them by their "severity". Some events are optional (modifier events), and some may be critical (abort if not available).

No, i3 compatability

>Wait, ok I see code in commands/bar/mode.c for handling bar display mode commands that sends an ipc barconfig event but the event never makes it to the socket because there's no security mapping there. I guess the question is whether to make bars subscribed to barconfig or bindings as a whole to do display mode switching.

I don't think the barconfig is maintained in the compositor beyond the initial allocation, but that's probably the best approach.

JerziKaminsky: > No, i3 compatability

sway introduces a security mechanism which doesn't exist in i3, so "i3 compatibility" can really apply only if security is off, otherwise it's new ground and subject to good judgment.

SirCmpwn: That's not true. We can introduce our security policies without making backwards-incompatible changes to the IPC protocol.

JerziKaminsky: What I suggested was that the client send a separate SUBSCRIBE requests for `[modifier]` 
apart from `[mode,workspaces]`. There's no wire-format or semantic changes involved, unless
the i3 protocol state machine requires that the client only send a single subscribe request during its lifetime. Is that the case?

SirCmpwn: I see what you mean now. That makes more sense, but I still don't want swaybar asking for events that it doesn't need.

JerziKaminsky: But the IPC protocol doesn't support UNSUBSCRIBE. So if you want to allow mode switching in the future, you'll have to deal with the situation where swaybar gets events it doesn't need (consider hide->dock). Might as well implement it to begin with. In fact this way, you *avoid* having to break compat with i3 protocol, because you won't need to extend its command set.

And because i3 allows mode switching, but no unsubscribe, we can guess that this is the way i3bar handles matters as well.

SirCmpwn: I don't understand how supporting mode switching requires subscribing to the modifier event.

SirCmpwn:
@calvinkosmatka what's an ETA on this? I'm getting ready to ship 0.13-rc1 - would you like me to wait for you so it lands in 0.13, or should this be postponed to 0.14?

swaybar/ipc.c
JerziKaminsky: You start in `hide` mode and subscribe to the events. You then switch into dock and the protocol doesn't allow unsubscribing. Presto, swaybar is forced to deal with events it doesn't use *anyway*.

SirCmpwn: Is that something people do? I wasn't thinking you could change that at runtime. Currently you can't change anything about the bar at runtime without restarting it (reloading your sway config restarts the bar)

JerziKaminsky: well, I never have, but I understood @calvinkosmatka [to say](https://github.com/SirCmpwn/sway/pull/1169#discussion_r111654634) i3 allows it.

SirCmpwn: Bleh. That's going to be annoying. Alright, here's the approach we should take: subscribe to only what we need, then add more subscriptions later if necessary as features are toggled on/off. We don't have to unsubscribe but we also don't have to subscribe until it's necessary.

calvinkosmatka:
@SirCmpwn Probably best not to count on having this in 0.13. Schedule is fairly busy now, but I'll see what I can do. I'll push some updates tonight.

swaybar/ipc.c
calvinkosmatka: Right. And the bar needs to be subscribed to an event that will allow the display mode change to take place in the first place (i.e. a barconfig update event)

include/swaybar/config.h
```diff
@@ -65,6 +76,16 @@ uint32_t parse_position(const char *position);
...
 /**
+ * Parse display mode dock|hide|invisible.
+ */
+enum display_mode_types parse_display_mode(const char *display_mode);
+
+/**
+ * Parse hidden state show|hide.
+ */
+enum hidden_states parse_hidden_state(const char *hidden_state);
+
```
JerziKaminsky: @calvinkosmatka, suggestion:
```
int parse_enum_string(const char** strings, int n, const char* s);
```
Then put the variable part of the problem into static initialized arrays, and both these function (and future ones) can just call into the one function, catch any error, and cast the return value.

You can use `sizeof(char* mode_enums[])/sizeof(void*)` to have the compiler compute `n`  for you
in individual parse functions when they delegate. The compiler knows this value at compile-time.

SirCmpwn: >You can use `sizeof(char* mode_enums[])/sizeof(void*)` to have the compiler compute n for you

`sizeof(mode_enums) / sizeof(mode_enums[1])` is more correct

protocols/desktop-shell.xml
```diff
@@ -116,6 +116,32 @@
...
+      </description>
+      <arg name="mode" type="uint"/>
+    </request>
+    
+    <request name="set_panel_hide" since="3">
+      <description summary="set if panel is currently hidden">
+        Tell the shell whether or not panel is currently hidden
+      </description>
+      <arg name="mode" type="uint"/>
+    </request>
```
JerziKaminsky: Hmm. 

Questions:
1. Is this the best API looking ahead? How will someone add support for other modes to this? how much work will it entail? 
2. Is this restricted to panels only? remember, this is a wayland protocol. The object could potentially be any window. It might be cool to be able to set hide_mode for arbitrary windows (someday...). Must this mode be coupled to the "placement" semantics of a "panel"?
3. Since the protocol supports enums. Is it necessary to tie the parameter to a specific mode? How about an API that takes a mode_id + param1/2/3/4. That's more extensible and keeps the protocol smaller.
3.

SirCmpwn: Don't update desktop-shell - make a new protocol and add this to it.

include/swaybar/config.h
JerziKaminsky: True. And `sizeof(mode_enums) / sizeof(mode_enums[0])` is even more correct.

And `sizeof(mode_enums) / sizeof(char *)` still more correct because C allows empty arrays to be declared.

Also, there's no need for specialized functions at all. You can use a macro that takes
a name and picks out the right string array:

```
static const char *ss_enumA[];

#define ENUM_STR(NAME, S) parse (ss_enum ## NAME, sizeof(ss_enum ## NAME) / sizeof(char *), S)

enum A val = ENUM_STR (A, "FOOBAR");
```

SirCmpwn: Macros - fuck no

`sizeof(mode_enums) / sizeof(char *)` - no, doesn't work if the array uses anonymous structs (so I'd rather not establish a pattern that doesn't work everywhere)

`sizeof(mode_enums) / sizeof(mode_enums[0])` - yes

JerziKaminsky: > Macros - fuck no

BS. boilerplate - fuck no.

Nothing eliminates bugs faster than deleting code. Don't be a zealot -- when exceptions make sense, make them.

> sizeof(mode_enums) / sizeof(char *) - no, doesn't work if the array uses anonymous structs 
> (so I'd rather not establish a pattern that doesn't work everywhere)

I don't understand how a `char* enum_strings[]` can "use anonymous structs". We're not talking about generic data structure here. And remember that the function we call type-checks that this is the case.

SirCmpwn: >BS. boilerplate - fuck no.

I take a hardline stance on macros. I will reject pull requests that use them. I've been pretty successful at making a codebase that's accessible to new contributors so far.

>I don't understand how a char* enum_strings[] can "use anonymous structs". We're not talking about generic data structure here. And remember that calling the function means the compiler type-checks for this.

I was talking about the generalized solution for getting the length of an array in items. It doesn't necessarily bring any significant benefits to the table in this case but it works for more situations, and if one solution works in many places then it should be preferred over a solution that only works for a few.

JerziKaminsky: >  take a hardline stance on macros. I will reject pull requests that use them

Regardless of context? I think that's wrong. Just because macros can be misused [doesn't mean](https://en.wikipedia.org/wiki/Ecological_fallacy) every particular use of them is misuse. cf. goto.  

> I've been pretty successful at making a codebase that's accessible to new contributors so far.

No argument here, but banning every use of macros had very little to do with it is my guess.

If you had 100 enums, would you prefer 100 boilerplate functions to a single macro?  Would that make for a cleaner codebase?

> I'd rather not establish a pattern that doesn't work everywhere
> It doesn't necessarily bring any significant benefits to the table in this case but it works for 
> more situations

Using the correct type of a variable also works in all cases and is certainly not an anti-pattern.

For the record, `sizeof` is a compile-time operator so you can use whatever index you like when sizing an array element. 0,1, 2**31-1, it doesn't matter (I overlooked that earlier).

SirCmpwn: Regarding macros, I don't really want to justify my approach here. Take it on faith that I know what I'm doing.

>Using the correct type of a variable also works in all cases and is certainly not an anti-pattern.

Well, if the variable is an anonymous struct, then you either have to redefine it in the sizeof argument (i.e. `sizeof(arr) / sizeof(struct { ... })`) or use `sizeof(arr) / sizeof(arr[0])`. You can't use `sizeof(array) / (sizeof(void *) + sizeof(void *))`, for example, if the struct is just two pointers, because of alignment issues. So `sizeof(arr) / sizeof(arr[0])` is the most general case.

JerziKaminsky: > Take it on faith that I know what I'm doing.

Hell no. If you can't/won't defend a contrary viewpoint, I'm not going to take that as proof you're right. But I do accept that in your project you get to decide such things (as I do in mine).

> Well, if the variable is an anonymous struct,

By which you mean "But I implemented enum->string using anonymous structs in 60ce81e".
Well, my view is that you probably shouldn't have. Translating an enum into a string is a common operation so relying on an anon struct is a doubtful choice.

SirCmpwn: This is getting out of hand.

>But I do accept that in your project you get to decide such things (as I do in mine).

Good. Leave it at that.

JerziKaminsky: Thank god.

PR #1171, opened 2017-04-14
Misc fixes

JerziKaminsky:
(Was "Prep Work for swaybar hide mode", but #1169 beat me to it so I just kept the cleanup fixes)

I think b463fb8 (taken together with #1150)  shows a current weakness in the way version management/CI is handled. I'd suggest that an explicit version baseline for dependencies be defined and documented. The CI should then test at least against that baseline to make sure sway is in sync with the shipped versions of its dependencies in recent distros releases. my 2c.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Install dependencies:
 * [wlc](https://github.com/Cloudef/wlc)
 * wayland
 * xwayland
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Thanks!

PR #1173, opened 2017-04-14
FOR_REVIEW: IPC security - Allow policy targets to be symlinks

### LINKED ISSUE #1172: IPC security `permit` silently fails for symlinks
I hit this when I made `/usr/bin/swaybar` into a symlink into my build directory and things got very weird.
Is there a security concern behind this or is it an oversight? @SirCmpwn I can send a PR if describe desired behavior, I think it should either work or log an error (maybe even abort).

JerziKaminsky:
closes #1172

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -1,3 +1,7 @@
+#define _XOPEN_SOURCE 500
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
 #include <math.h>
 #include <stdint.h>
 #include <stdio.h>
@@ -117,4 +121,41 @@ uint32_t parse_color(const char *color) {
 		res = (res << 8) | 0xFF;
 	}
 	return res;
+}
+
+char* resolve_path(const char* path) {
+    struct stat sb;
+    ssize_t r;
+    int i;
+    char *current = NULL;
+    char *resolved = NULL;
+
+    if(!(current = strdup(path))) {
+           return NULL;
+    }
+    for (i = 0; i < 16; ++i) {
+        if (lstat(current, &sb) == -1) {
+           goto failed;
+        }
+        if((sb.st_mode & S_IFMT) != S_IFLNK) {
+            return current;
+        }
+        if (!(resolved = malloc(sb.st_size + 1))) {
+           goto failed;
+        }
+        r = readlink(current, resolved, sb.st_size);
+        if (r == -1 || r > sb.st_size) {
+           goto failed;
+        }
+        resolved[r] = '\0';
+        free(current);
+        current = strdup(resolved);
+        free(resolved);
+        resolved = NULL;
+    }
+
+failed:
+    free(resolved);
+    free(current);
+    return NULL;
 }
--- a/sway/commands/permit.c
+++ b/sway/commands/permit.c
@@ -38,25 +38,6 @@ static enum secure_feature get_features(int argc, char **argv,
 	return features;
 }
 
-static struct feature_policy *get_feature_policy(const char *name) {
-	struct feature_policy *policy = NULL;
-	for (int i = 0; i < config->feature_policies->length; ++i) {
-		struct feature_policy *p = config->feature_policies->items[i];
-		if (strcmp(p->program, name) == 0) {
-			policy = p;
-			break;
-		}
-	}
-	if (!policy) {
-		policy = alloc_feature_policy(name);
-		if (!policy) {
-			sway_abort("Unable to allocate security policy");
-		}
-		list_add(config->feature_policies, policy);
-	}
-	return policy;
-}
-
 struct cmd_results *cmd_permit(int argc, char **argv) {
 	struct cmd_results *error = NULL;
 	if ((error = checkarg(argc, "permit", EXPECTED_MORE_THAN, 1))) {
--- a/sway/commands/permit.c
+++ b/sway/commands/permit.c
@@ -38,7 +38,7 @@ static enum secure_feature get_features(int argc, char **argv,
 	return features;
 }
 
-static struct feature_policy *get_policy(const char *name) {
+static struct feature_policy *get_feature_policy(const char *name) {
 	struct feature_policy *policy = NULL;
 	for (int i = 0; i < config->feature_policies->length; ++i) {
 		struct feature_policy *p = config->feature_policies->items[i];
@@ -66,7 +66,7 @@ struct cmd_results *cmd_permit(int argc, char **argv) {
 		return error;
 	}
 
-	struct feature_policy *policy = get_policy(argv[0]);
+	struct feature_policy *policy = get_feature_policy(argv[0]);
 	policy->features |= get_features(argc, argv, &error);
 
 	sway_log(L_DEBUG, "Permissions granted to %s for features %d",
@@ -84,7 +84,7 @@ struct cmd_results *cmd_reject(int argc, char **argv) {
 		return error;
 	}
 
-	struct feature_policy *policy = get_policy(argv[0]);
+	struct feature_policy *policy = get_feature_policy(argv[0]);
 	policy->features &= ~get_features(argc, argv, &error);
 
 	sway_log(L_DEBUG, "Permissions granted to %s for features %d",
--- a/sway/commands/permit.c
+++ b/sway/commands/permit.c
@@ -1,7 +1,9 @@
+#define _XOPEN_SOURCE 500
 #include <string.h>
 #include "sway/commands.h"
 #include "sway/config.h"
 #include "sway/security.h"
+#include "util.h"
 #include "log.h"
 
 static enum secure_feature get_features(int argc, char **argv,
@@ -47,12 +49,29 @@ struct cmd_results *cmd_permit(int argc, char **argv) {
 		return error;
 	}
 
-	struct feature_policy *policy = get_feature_policy(argv[0]);
-	policy->features |= get_features(argc, argv, &error);
+	bool assign_perms = true;
+	char *program = NULL;
 
+	if (!strcmp(argv[0], "*")) {
+		program = strdup(argv[0]);
+	} else {
+		program = resolve_path(argv[0]);
+	}
+	if (!program) {
+		sway_assert(program, "Unable to resolve IPC permit target '%s'."
+			" will issue empty policy", argv[0]);
+		assign_perms = false;
+		program = strdup(argv[0]);
+	}
+
+	struct feature_policy *policy = get_feature_policy(program);
+	if (assign_perms) {
+		policy->features |= get_features(argc, argv, &error);
+	}
 	sway_log(L_DEBUG, "Permissions granted to %s for features %d",
 			policy->program, policy->features);
 
+	free(program);
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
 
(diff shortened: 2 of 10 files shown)


SirCmpwn:
Before I consider this I think we also need to enforce that the program that ends up being executed is owned by root and not world-writable.

JerziKaminsky:
Gladly, but can you clue me in on the reasoning? I can't make out the attack vector.
If you're going to nuke it from space, what's the purpose of having such fine-grain control over IPC interactions per program?

SirCmpwn:
If you can write to a program that has security policies, then you can just overwrite it with a program whose behavior is more sinister.

common/util.c
```diff
@@ -118,3 +122,30 @@ uint32_t parse_color(const char *color) {
...
+        return strdup(path);
+    }
+    if (!(resolved = malloc(sb.st_size + 1))) {
+       return NULL;
+    }
+    r = readlink(path, resolved, sb.st_size);
+    if (r == -1 || r > sb.st_size) {
+       goto failed;
+    }
+    resolved[r] = '\0';
```
SirCmpwn: What about recursive symlinks?

sway/commands/permit.c
```diff
@@ -40,9 +41,14 @@ static enum secure_feature get_features(int argc, char **argv,
 
 static struct feature_policy *get_policy(const char *name) {
 	struct feature_policy *policy = NULL;
+	char *program = NULL;
+
+	if (!(program = resolve_path(name))) {
+		sway_abort("Unable to allocate security policy");
```
SirCmpwn: abort is probably not the correct action here, use sway_assert and return a policy with no permissions.

sway/commands/permit.c
```diff
@@ -54,6 +60,10 @@ static struct feature_policy *get_policy(const char *name) {
 		}
 		list_add(config->feature_policies, policy);
 	}
+	sway_assert(!strcmp(program, policy->program),
+			"!! DANGER !! symlink modified during policy allocation.");
```
SirCmpwn: In this case you need to return no permissions as well.

SirCmpwn: Actually I don't think this ever happens?

JerziKaminsky:
> If you can write to a program that has security policies, then you can just overwrite it with a program 
> whose behavior is more sinister.

Sure but if you can't write to the program, why go through all the effort to sandbox it ? defense in depth I guess.

sway/commands/permit.c
JerziKaminsky: Time of check/Time of use. The symlink could be switched under us from the time we validate to the time we create the policy.

JerziKaminsky: wait. sway_assert doesn't abort?

common/util.c
JerziKaminsky: Good catch, I was assuming `readlink` handled this, but that doesn't make sense I now realize.

sway/commands/permit.c
JerziKaminsky: resolve_path returns error only if malloc fails...

SirCmpwn:
>Sure but if you can't write to the program, why go through all the effort to sandbox it ? defense in depth I guess.

A number of reasons. First, because we can limit what the program can access in the first place, so you can give limited permissions to mixed trust programs. More importantly, though, if we didn't have access controls at all then you could leverage bash or sed or something similar to arbitrarily work around restrictions.

sway/commands/permit.c
SirCmpwn: No, it logs the error and returns the first parameter. Sway should never abort at runtime, because that would kill your whole session and cause data loss and things like that. It's just a utility function to add conditional logging, the user has to implement a safe exit path.

SirCmpwn: Also, you should be storing the fully resolved symlink. When we validate the policy we're going to get the fully resolved path from `/proc/[pid]/exe`, so we need to compare against that to get the appropriate policy. All symlink resolution should happen when the policy is parsed and never again.

common/util.c
```diff
@@ -118,3 +122,38 @@ uint32_t parse_color(const char *color) {
...
+        if (r == -1 || r > sb.st_size) {
+           goto failed;
+        }
+        resolved[r] = '\0';
+        free(current);
+        current = strdup(resolved);
+        free(resolved);
+        resolved = NULL;
+    }
+failed:
```
SirCmpwn: You should probably return NULL in case of failure.

JerziKaminsky: This was deliberate because it simplified handling of "*",  but I've changed it.

sway/commands/permit.c
JerziKaminsky: > Also, you should be storing the fully resolved symlink.
> All symlink resolution should happen when the policy is parsed and never again.

That's what's stored in policy->program, yeah.

> Sway should never abort at runtime, 

got it.

sway/security.c
```diff
@@ -1,13 +1,50 @@
...
+	}
+	if (sb.st_mode & S_IWOTH) {
+		sway_log(L_ERROR, "IPC target '%s' MUST NOT be world writable", program);
+		goto failed;
+	}
+
+	return true;
+
+failed:
+	return false;
```
SirCmpwn: No need for the failed label and all the gotos in this function, just return false wherever you would goto failed

sway/security.c
```diff
@@ -94,7 +141,32 @@ static const char *get_pid_exe(pid_t pid) {
...
+			policy = p;
+			break;
+		}
+	}
+	if (!policy) {
+		policy = alloc_feature_policy (program);
+		if (!policy) {
+			sway_abort ("Unable to allocate security policy");
+		}
+		list_add (config->feature_policies, policy);
```
SirCmpwn: Style nit, don't put a space between function name and parenthesis

JerziKaminsky:
@SirCmpwn, ready for review.

I reworked the whole thing and some positive collateral cleanup to existing code resulted.
I've now spent quite a bit of effort on this, but that shouldn't prevent you from NAKing if you don't like it.
I've reorganized the commit series so that everything up to and including `validate_ipc_target()` should be a no-brainer, and it implements the checks you wanted (root/chmod o-w) without doing anything else.

`Add resolve_path util function` is needed by the last commit, but separaed because it's a handy util function you might want to cherry-pick regardless.

The final commit "Handle symlinks as IPC security targets" is really the only commit that deals with the symlink issue at all. See the commit message for description and a longinh discussion about the race condition I mentioned earlier.

Here's the test script I used (run from build directory):
```
#!/bin/sh

sudo rm -f /usr/bin/swaybar2 /usr/bin/swaybar; sudo make -j 4 install > /dev/null
echo "Standard config, should work"
sway

sudo rm -f /usr/bin/swaybar2 ; sudo mv /usr/bin/swaybar /usr/bin/swaybar2; sudo ln -sf /usr/bin/swaybar2 /usr/bin/swaybar
echo "symlink, should work"
sway

echo "symlink to symlink, should work"
sudo rm -f /usr/bin/swaybar3 ; sudo mv /usr/bin/swaybar2 /usr/bin/swaybar3; sudo ln -sf /usr/bin/swaybar3 /usr/bin/swaybar2
sway

sudo rm /usr/bin/swaybar2
echo "dangling symlink, should abort"
sway

sudo rm -f /usr/bin/swaybar2 /usr/bin/swaybar; sudo make -j 4 install  > /dev/null
echo "bad owner, should abort"
sudo chown user1:user1 /usr/bin/swaybar
sway

sudo rm -f /usr/bin/swaybar2 /usr/bin/swaybar; sudo make -j 4 install > /dev/null
echo "bad perms, should abort"
sudo chmod o+w /usr/bin/swaybar
sway

sudo rm -f /usr/bin/swaybar2 /usr/bin/swaybar; sudo make -j 4 install > /dev/null
echo "symlink is not a file, should abort"
sudo rm -f /usr/bin/swaybar2 ; sudo mv /usr/bin/swaybar /usr/bin/swaybar2; sudo ln -sf /usr/bin/swaybar /usr/bin/NOSUCHFILE
sway

sudo rm -f /usr/bin/swaybar*
sudo make -j 4 install > /dev/null
```

I've spent this much time on it because it got delicate and therefore interesting, so fun. But I'd be the first to admit the result was more subtle than anticipated and maybe just barfing a log error on symlink (which was the alternative I suggested in #1172) might have been a suitable and far simpler solution.
But the resulting code cleanups are a win anyway, so something good did come out of it.

That said, as this point I think this PR does what it claims, and the test script should provide some confidence.

SirCmpwn:
```
- Renamed validate_ipc_target to validate_ipc_program_target
  to emphasize the fact that it should be called with
  a "program target" not a "policy target". This has
  the unfortunate side-effect that errors report the "program
  target" at fault, not the "policy target" as it appears in
  the config file. This is necessary in order to avoid the
  possible race condition, described below.
```

I'm not really a huge fan of this distinction, I think it's not very important. It would be better to just resolve symlinks on policy creation and call it a day. No one is going to carry around references to unresolved paths.

SirCmpwn:
Also, for development purposes it probably makes sense not to enforce that the binary is owned by root for debug builds. Useful for i.e. hacking on swaybar.

common/util.c
JerziKaminsky: Fixed.

sway/commands/permit.c
JerziKaminsky: Done

sway/security.c
JerziKaminsky: Done.

sway/security.c
JerziKaminsky: Muscle memory. Until Sway becomes my focus project, my default clang-format belongs to someone else. 

Fixed.

JerziKaminsky:
> I'm not really a huge fan of this distinction,

it's really only there for documentation purposes, but I eliminated the renaming and played down the issue. Smaller patch.

> Also, for development purposes it probably makes sense not to enforce that the binary is owned by 
> root for debug builds. Useful for i.e. hacking on swaybar.

Wrapped in `#ifndef NDEBUG` and otherwise emits L_INFO and allows.

sway/commands/permit.c
```diff
@@ -84,7 +84,7 @@ struct cmd_results *cmd_reject(int argc, char **argv) {
 		return error;
 	}
 
-	struct feature_policy *policy = get_policy(argv[0]);
+	struct feature_policy *policy = get_feature_policy(argv[0]);
```
SirCmpwn: Don't you need to resolve symlinks here?

SirCmpwn:
Looks good aside from minor nits, will merge once resolved. Thanks!

sway/commands/permit.c
JerziKaminsky: Yes, I missed that. 

Scenario: suppose a policy file indicates a reject target against a file which does not exist. if we ignore the resolve fail and use the verbatim policy target, and the target is then created as a symlink to elsewhere, the reject policy won't be found when get_pid_exe is used to look it up. 

How do you want to handle that?

SirCmpwn: I think that's niche enough that we don't need to bother handling it.

JerziKaminsky: Done

SirCmpwn:
Thanks!

PR #1175, opened 2017-04-16
Fix multiple issues in sway_assert

JerziKaminsky:
- `_sway_assert` is a variadic function which tries to delegate to another variadic function. This requires a vprintf-style variant of the delegate. https://stackoverflow.com/a/150616
- DRY: _sway_log now delegates to _sway_vlog. 
- `sway_assert` did not correctly propagate the location information for the assert callsite.

### DIFF
--- a/common/log.c
+++ b/common/log.c
@@ -63,7 +63,8 @@ void sway_abort(const char *format, ...) {
 	sway_terminate(EXIT_FAILURE);
 }
 
-void _sway_log(const char *filename, int line, log_importance_t verbosity, const char* format, ...) {
+void _sway_vlog(const char *filename, int line, log_importance_t verbosity,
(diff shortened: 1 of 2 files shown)


common/log.c
```diff
@@ -63,52 +63,56 @@ void sway_abort(const char *format, ...) {
...
+        }
+
+        vfprintf (stderr, format, args);
+
+        if (colored && isatty (STDERR_FILENO)) {
+            fprintf (stderr, "\x1B[0m");
+        }
+        fprintf (stderr, "\n");
+    }
+}
```
SirCmpwn: Tabs, not spaces

JerziKaminsky: Fixed

SirCmpwn:
Thanks!

PR #1178, opened 2017-04-18
Prevent sway from duplicating on a failed fork

4e554c4c:
Also remove a useless `sway_log` and replace it with a pipe

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -912,8 +912,16 @@ void merge_output_config(struct output_config *dst, struct output_config *src) {
 }
 
 static void invoke_swaybar(struct bar_config *bar) {
+	// Pipe to communicate errors
+	int filedes[2];
+	if (pipe(filedes) == -1) {
+		sway_log(L_ERROR, "Pipe setup failed! Cannot fork into bar");
+		return;
+	}
+
 	bar->pid = fork();
 	if (bar->pid == 0) {
+		close(filedes[0]);
 		if (!bar->swaybar_command) {
 			char *const cmd[] = {
 				"swaybar",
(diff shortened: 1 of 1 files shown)


ascent12:
You should probably be using _Exit() instead of exit() when exiting from a child.

sway/config.c
```diff
@@ -922,14 +930,17 @@ static void invoke_swaybar(struct bar_config *bar) {
...
 			execvp(cmd[0], cmd);
+			exit(EXIT_SUCCESS);
 		} else {
 			// run custom swaybar
 			int len = strlen(bar->swaybar_command) + strlen(bar->id) + 5;
 			char *command = malloc(len * sizeof(char));
 			if (!command) {
-				sway_log(L_ERROR, "Unable to allocate swaybar command string");
-				return;
+				write(filedes[1], "Unable to allocate swaybar command string", 42);
```
SirCmpwn: Use a constant and sizeof

sway/config.c
```diff
@@ -940,10 +951,22 @@ static void invoke_swaybar(struct bar_config *bar) {
...
 			};
 
+			close(filedes[1]);
 			execvp(cmd[0], cmd);
 			free(command);
+			exit(EXIT_SUCCESS);
+		}
+	}
+	close(filedes[0]);
+	char *buf = malloc(1024);
```
SirCmpwn: Can you write an int with the length, then allocate an appropriately sized buffer for it?

4e554c4c:
@ascent12 I changed it. What's the difference?

sway/config.c
```diff
@@ -940,8 +954,22 @@ static void invoke_swaybar(struct bar_config *bar) {
...
+			close(filedes[1]);
 			execvp(cmd[0], cmd);
 			free(command);
+			_exit(EXIT_SUCCESS);
+		}
+	}
+	close(filedes[0]);
+	int len;
+	if(read(filedes[1], &len, sizeof(int)) == sizeof(int)) {
+		char *buf = malloc(len);
```
SirCmpwn: You need to free buf when you're done with it

4e554c4c: Thanks for catching that.

PR #1184, opened 2017-04-20
More Logging Fixes

JerziKaminsky:


SirCmpwn:
No thanks to L_WARN. The other change is fine. Can you rebase?

JerziKaminsky:
Sad to hear it, I needed it. Pushed.

SirCmpwn:
We generally just log warnings with L_INFO or L_ERROR if they're severe. Not necessary to add another thing.

JerziKaminsky:
Logging has conventions, warning is a standard thing. I needed something which signals a problem and is stronger than info. For example "Security policy violated but this is a debug build" feels like a warning to me, or "Can't validate target, creating empty policy". 

`-V` is verbose and it's hard to pick out the wheat from the chaff.  I'll live.

SirCmpwn:
Well, I suppose I would be more open to L_WARN if you went about the codebase and changed everything that should use it (but currently does not). Should be done separately though.

JerziKaminsky:
That's a shitton of work. No thanks.

SirCmpwn:
Really? Doesn't seem that bad to me.

JerziKaminsky:
Yeah, well you know the codebase backwards, but I'd have to be confident that I understand the context for every suspect log message. Too steep. If WARN makes sense -- merge it, if it doesn't -- don't.

4e554c4c:
Don't change something that affects the whole codebase if you don't want to change the whole codebase then?

JerziKaminsky:
How does making a new log level available "affect the whole codebase"? what code will it break exactly?

SirCmpwn:
Well, everything that should use the new log level and doesn't becomes a bug. So without doing all the work you're basically introducing a few dozen bugs.

JerziKaminsky:
You're right. Having all the messages which should be a warning report instead as INFO/ERROR would be a real problem, so I guess it's better to leave things as they are and have the messages which should be warning a report instead as INFO/ERROR.

Seriously, there are 134 log INFO/ERROR messages in the codebase and auditing them one by one is simply not something I want to spend time on. If you can't live with stuff being printed exactly as they have been for a couple of years, ok. I hope someone else will step up in the future.

SirCmpwn:
If you just have a trash bin, then everything you throw out is already in the right place. If you add a recycle bin, however, lots of items in your trash bin are suddenly in the wrong place.

Most of the logging in place is pretty clear. If you don't want to put in the work then that's fine, but it really wouldn't be very difficult. An hour's work at worst.

JerziKaminsky:
Cute analogy but I think it misses some important points. Here's my own version:

A man walks up to your house and offers to install a brand new shiny recycle bin, for free. "It's eco-friendly," he says, "and once installed you and your family and visitors can start using it for your recyclables instead of throwing them in the trash". You reply "That is indeed a handsome recycle bin, but I absolutely refuse to have it installed until you, the guy who is offering to install for free, go through my garbage can and place whatever recyclables are there into the new recycle bin.". The man replies "I don't like that idea at all," "I had nothing to do with putting that garbage there, and I have better things to do then go through your garbage. Why don't you just accept the recycle bin, use it from now on, and if anyone walks by your house and sees this or that recyclable in your trash can, they can always fish it out and put it in the bin, if it bothers them." "No," you say "I wouldn't be able to sleep nights knowing that my trash possibly contains recyclables which are going to the landfill instead of the recycling center". "But you've been doing that for years!" the man says, a little alarmed,  "what's the difference? isn't it better to just start recycling from now on?", "No, I don't think so," you say, "I'd simpler to not have the recycle bin installed in the first place and then I don't have to think about it." The man shrugs and walks away.... 


 :man_shrugging:   :walking_man:  ...

PR #1185, opened 2017-04-20
Refactor IPC target validation

JerziKaminsky:
Followup to #1173 

- `ipc` config option shouldn't fail if target doesn't exist (don't refuse to create an ipc policy just because the file doesn't exist. Borks startup)
- Added resolve_ipc_path() to reduce boilerplate
- Policy allocators now store the result of validate_ipc_target()  under policy->validated.
- cmd_permit/_reject use policy->validated to make decisions.
- sway_log instead of sway_assert on unvalidated target, we don't want to abort, even in debug builds.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -208,6 +208,7 @@ enum secure_feature {
 struct feature_policy {
 	char *program;
 	uint32_t features;
+	bool validated;
 };
 
 enum ipc_feature {
@@ -235,6 +236,7 @@ enum ipc_feature {
 struct ipc_policy {
 	char *program;
 	uint32_t features;
+	bool validated;
 };
 
 /**
--- a/include/sway/security.h
+++ b/include/sway/security.h
@@ -15,4 +15,5 @@ struct feature_policy *alloc_feature_policy(const char *program);
 struct ipc_policy *alloc_ipc_policy(const char *program);
 struct command_policy *alloc_command_policy(const char *command);
 
+char *resolve_ipc_path(const char* program);
 #endif
--- a/sway/commands/ipc.c
+++ b/sway/commands/ipc.c
@@ -19,14 +19,6 @@ struct cmd_results *cmd_ipc(int argc, char **argv) {
 		return error;
 	}
 
-	char *program = NULL;
-
-	if (!strcmp(argv[0], "*")) {
-		program = strdup(argv[0]);
-	} else if (!(program = resolve_path(argv[0]))) {
-		return cmd_results_new(
-				CMD_INVALID, "ipc", "Unable to resolve IPC Policy target.");
-	}
 	if (config->reading && strcmp("{", argv[1]) != 0) {
 		return cmd_results_new(CMD_INVALID, "ipc",
 				"Expected '{' at start of IPC config definition.");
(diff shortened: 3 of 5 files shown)


SirCmpwn:
>ipc config option shouldn't fail if target doesn't exist (don't refuse to create a an ipc just because the file doesn't exist. Borks startup)

Why? In what situation does this occur? Seems to incur extra complexity to handle an unlikely scenario imo.

A lot of changes seem to stem from that. I don't like the idea of pushing validation deeper into runtime, I'd rather do it all on startup - give warnings for bad configurations straight away and simplify checks later on.

JerziKaminsky:
> Why? In what situation does this occur? Seems to incur extra complexity to handle an unlikely scenario imo.

Right now, a policy targeting a missing file prevents sway from starting up:
```
nonexistent ipc target, should warn
04/20/17 20:17:55 - [permit.c:62] cmd_permit:Unable to resolve IPC permit target '/usr/bin/swaybar'. will issue empty policy
04/20/17 20:17:55 - [security.c:156] Unable to allocate security policy
./test.sh: line 20: 19485 Segmentation fault      (core dumped) sway -V
```
The policy allocators can currently only communicate to a caller by returning a policy or not. 
On the other hand allocators don't have the context to perform policy decisions, that's the caller's job  (warn, don't give perms, ignore). So instead, the newly created policy is returned to the caller with information about how validation fared. I think we do want to integrate validation down into the allocators, so that it's impossible to "forget" to do it.

> A lot of changes seem to stem from that. 

not really, I just took the opportunity to also tidy up.

SirCmpwn:
We probably shouldn't abort, but we shouldn't defer validation. We should just ignore the policy and print a warning.

SirCmpwn:
>Policy allocators now store the result of validate_ipc_target() under policy->validated.

>cmd_permit/_reject use policy->validated to make decisions.

These changes are ones that make me concerned. By the time you obtain a reference to a policy, that policy should have already been validated.

JerziKaminsky:
> We should just ignore the policy and print a warning.

What does "ignore" actually mean? Earlier you asked that permit return an empty policy. It did and does.

> We probably shouldn't abort, but we shouldn't defer validation

We don't.

> that policy should have already been validated.

"checked" or "enforced"? 
 
Right now, if validation fails, the **allocator** tells the caller it's [out of memory](https://github.com/SirCmpwn/sway/commit/bcf9338ce7dec6c83564c92567d14443d0467806#diff-ea05a0ab668f60879022dfb6ef9be459R48). That's wrong. This PR fixes it. That's it.

SirCmpwn:
>Anyway, right now if validation fails the allocator tells the caller it's out of memory. That's wrong. This PR fixes it. That's it.

Returning NULL doesn't signal OOM. It just means it didn't work. It would set errno = ENOMEM if we wanted to specify the specific failure, but that's not important here.

SirCmpwn:
I still haven't looked over the code, but the only change I think is necessary is to avoid aborting if the target doesn't exist.

JerziKaminsky:
> Returning NULL doesn't signal OOM. It just means it didn't work.

> the only change I think is necessary is to avoid aborting if the target doesn't exist.

How will the caller tell the difference?

SirCmpwn:
Why does the caller need to tell the difference?

JerziKaminsky:
> Why does the caller need to tell the difference?

Because an OOM is non-recoverable and a bad validation potentially is.

You seem to feel that the change needed to fix this is trivial and what I've done unnecessary, so I'm going to just close this and open an issue instead. You can deal with it as you think best.

PR #1194, opened 2017-04-26
Add -DVERSION flag for release version numbers

snoack:
Currently, the version number (as shown by `sway --version` and used in other places) is based on
`git describe`. But when packaging sway for Debian, where it is built outside of git, these information aren't available. Moreover, the version string contains information, such as date and branch, that aren't relevant for release builds, not to mention that the boilerplate to generate that version string is duplicated across the code base, which I fixed as well by having `CMakeList.txt` generate the version string. Finally, while on it, I also fixed the JSON version information in the IPC code.

Note that I removed the code path for when `SWAY_GIT_VERSION` and `SWAY_GIT_BRANCH` aren't defined. This was dead code anyway, as `CMakeList.txt` always defined these macros, but merely with an empty string if running outside of git. Moreover, I think it is preferable having the build fail if neither `-DVERSION` was given nor when building from git. Otherwise, package maintainers will easily forget to use `-DVERSION`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,10 @@ endif()
 list(INSERT CMAKE_MODULE_PATH 0
 	${CMAKE_CURRENT_SOURCE_DIR}/CMake
 	)
-if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
+
+if (VERSION)
+	add_definitions(-DSWAY_VERSION=\"${VERSION}\")
+else()
 	execute_process(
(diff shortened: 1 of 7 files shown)


SirCmpwn:
You're removing information here, we still need the full git describe string for off-release builds.

snoack:
I don't know what you mean. If `-DVERSION` isn't used, it is using the exact same version information as before my changes. If it does not, that wasn't indented, can you point me where information are removed?

SirCmpwn:
You're right, I misread it. Thanks!

PR #1201, opened 2017-04-29
Support specifying fewer than 5 colors

SirCmpwn:
This doesn't work, I'm not sure why. The color structs definitely get
changed but if you specify fewer than 5, it renders with the defaults.


SirCmpwn:
@zandrmartin points out that child border is the one you tend to see most of the time. Which seems weird to me but w/e

PR #1209, opened 2017-05-04
Add mouse button bindings

janza:
Adds support for bindings like:

bindsym button3 floating toggle
bindsym $mod+button3 floating toggle


SirCmpwn:
Thanks!

PR #1216, opened 2017-05-09
New list implementation.

ascent12:
This is a new implementation for the common list_t structure. It changes from a void ** implementation to using ~C99 flexible array members~ to a single void *. This is reduce the total number of allocations and to increase the locality of the data. My (very basic) testing has shown that it is roughly 2 to 3 times faster than the current implementation, depending on how it's used.

list_t is a bit of a misleading name: it's a dynamic array, not a linked list. Perhaps renaming this type would be appropriate.

Currently this doesn't signal if a realloc fails, and just doesn't modify this list. Maybe a different behaviour would be more appropriate.

This PR doesn't change any of the current usage of list_t. I want to see if this sort of change is welcome so I don't bother wasting my time changing a lot of code if it's not.

# Advantages

- Better locality.
- Allows you to have arrays of primitives and structs without needing to malloc each of them, meaning fewer allocations.

# Drawbacks

- ~You cannot have a list_t allocated on the stack~. This never happens in the sway codebase.
- ~You cannot have an array of list_t~. This also never happens.
- ~reallocs are _slightly_ more expensive, as the structure overhead also needs to be copied~.
- ~You can't "share" a list_t in multiple places, and one of them might realloc the structure when something is added~. I also don't think this happens anywhere in sway, but I'm not certain. Pointers to pointers still work, though.

# ~alignas~

~In order for this code to be portable and correct, it needs to use a feature introduced in C11 (alignas). The 'data' member of list_t needs to be correctly aligned to accept any type. Unaligned accesses will work on x86/x86_64, but they are slower. Without the alignas, it would be sitting on the boundary of a size_t, which is probably more than enough for most usages, but types with stricter alignments do exist. There are ways to deal with this in C99, but it would be non-portable~.

~Because of this, C11 is enabled for the project. This is quite a significant change to make, and something the maintainers might not want to do. GCC and Clang have both supported C11 for quite a while (and even default to it), and C11 doesn't add anything too major/offensive over C99, so maybe this change is permissible. If it's not, the alignas could be removed, and you could just hope that the alignment issue never becomes a real problem~.

# Interface changes
It has a slightly different interface than the current list_t, some of which are inherent to the design, so usage of it needs to be changed.

- Direct access through list->items no longer works. **list_get** needs to be used, as there are some offset calculations that have to happen.
- ~**list_free** removed: Since it's now a single allocation, we just need to use **free**()~.
- ~**list_add** and **list_insert** changed: It needs to take a list ** now~.
- **list_cat** removed: This was never used in the sway codebase.
- ~**list_bsearch** added: We have qsort, so I thought bsearch would be helpful too.~
- **list_seq_find** renamed **list_lsearch**: Just to make the name more consistent. Also, another argument is added, to (optionally) retrieve the item you're looking for.
- **list_stable_sort** renamed **list_isort**: The implementation was changed to insertion sort from an in-place merge sort. This function was only used in one place: https://github.com/SirCmpwn/sway/blob/master/sway/output.c#L276. Since this is a pretty small list, we probably don't a function with a lot of overhead like in-place merge sort. While insertion sort is a "bad" (O(n^2)) algorithm, it works pretty well for small and almost-sorted inputs.
- ~**list_foreach** renamed **list_foreach_cb**~
- ~**list_foreach** now is a macro which acts like a normal for-loop. I know how much @SirCmpwn  doesn't like using macros, but I believe this one adds a lot of useful utility. It can allow you to not have to define a simple callback function that you're just going to use in one place~.
- **list_end** added: This is just a quick function to get a pointer past of end of the list, making writing for loops for this implementation easier.


SirCmpwn:
Initial impression: I like it.

>list_t is a bit of a misleading name: it's a dynamic array, not a linked list. Perhaps renaming this type would be appropriate.

It's not called `llist_t`.

>In order for this code to be portable and correct, it needs to use a feature introduced in C11 (alignas).

This is probably okay. Let me think about it. I don't think we lose most of the benefits this PR comes with if we put the data buffer on the heap, do we? Could also just handle alignment manually.

>You can't "share" a list_t in multiple places, and one of them might realloc the structure when something is added. I also don't think this happens anywhere in sway, but I'm not certain. Pointers to pointers still work, though.

Putting the buffer on the heap also fixes this. It also makes sizeof(list_t) not misleading, which I prefer (I'm not keen on dynamically sized structs in general).

>list_free removed: Since it's now a single allocation, we just need to use free().

I still prefer this for the sake of making it cheaper to change later.

>list_foreach now is a macro which acts like a normal for-loop. I know how much @SirCmpwn doesn't like using macros, but I believe this one adds a lot of useful utility. It can allow you to not have to define a simple callback function that you're just going to use in one place.

Nope.

Also, the build failed.

SirCmpwn:
FTR you can (and we already do in many places) just write a for loop yourself to avoid making the extra function.

ascent12:
> I don't think we lose most of the benefits this PR comes with if we put the data buffer on the heap, do we?
>Putting the buffer on the heap also fixes this. It also makes sizeof(list_t) not misleading, which I prefer (I'm not keen on dynamically sized structs in general).

I was REALLY just trying to optimise the crap out of the cache by removing as much indirection as I could. Using a heap buffer would remove the alignof requirement, and remove some of drawbacks. However, from my (still very basic) testing, this extra layer of indirection is about 25% slower than this implementation. It depends how far you want to go for the sake of speed.

>Nope.

With this new implementation, writing such a loop would be a much more of a hassle, hence the macro.

>Also, the build failed.

I know. The code hasn't been updated to use this list yet. This PR isn't ready to be accepted.

ascent12:
Also, I should note that for my testing, I've been filling the list with a million ints and accessing them randomly. This may not reflect on the real world cases that the lists are used for in sway, but it's something.

SirCmpwn:
>I was REALLY just trying to optimise the crap out of the cache by removing as much indirection as I could. Using a heap buffer would remove the alignof requirement, and remove some of drawbacks. However, from my (still very basic) testing, this extra layer of indirection is about 25% slower than this implementation. It depends how far you want to go for the sake of speed.

Well, list_t isn't the bottleneck in sway's performance. I won't sacrifice convenience and simplicity to fix a non-issue.

>With this new implementation, writing such a loop would be a much more of a hassle, hence the macro.

I don't think it's worth it. Still a no.

ascent12:
>I won't sacrifice convenience and simplicity to fix a non-issue.

Yeah, that's good point. I just adapted this from some toy code I was trying to make as fast as possible (just to see how far I could get it to go).
I'll change it to a heap allocated buffer implementation.

SirCmpwn:
Cool, thanks.

ascent12:
@SirCmpwn 
Ok, the implementation has changed.

common/list.c
```diff
@@ -1,134 +1,202 @@
...
-	list->length++;
-	list->items[index] = item;
-}
+void list_add(list_t *list, const void *data) {
+	if (!data || !list || !resize(list)) {
+		return;
+	}
+
+	size_t size = list->memb_size;
+	unsigned char (*array)[size] = list->data;
```
SirCmpwn: What with this syntax? Also, why unsigned char over uint8_t?

SirCmpwn: `memcpy(&list->data[list->length], ...)` should work the same, right?

common/list.c
```diff
@@ -1,134 +1,202 @@
...
-	void *tmp = list->items[src];
-	list->items[src] = list->items[dest];
-	list->items[dest] = tmp;
+void *list_get(list_t *list, size_t index) {
+	if (!list || index >= list->length) {
+		return NULL;
+	}
+
+	size_t size = list->memb_size;
+	unsigned char (*array)[size] = list->data;
```
SirCmpwn: Makes a little more sense here. Still a bit weird though.

SirCmpwn: `return &list->data[list->memb_size * index]` would probably be simpler

common/list.c
```diff
@@ -1,134 +1,202 @@
...
+	} else {
+		if (ret) {
+			memcpy(ret, ptr, list->memb_size);
+		}
+		return (ptr - (unsigned char *)list->data) / list->memb_size;
 	}
+}
+
+ssize_t list_lsearch(const list_t *list, int compare(const void *, const void *),
+	const void *key, void *ret) {
```
SirCmpwn: Indent this line once more

common/list.c
ascent12: >What with this syntax?

It's a pointer to a VLA. The syntax seems a bit tricky and clever, but it really cleans up the other parts of the code. We can treat it like a normal array, where pointer arithmetic will work properly.

>Also, why unsigned char over uint8_t?

I don't think uint8_t really adds much here. It's all basically "typeless" data that we're not going to touch. "unsigned char" seems good enough for a byte type.

>memcpy(&list->data[list->length], ...) should work the same, right?

No, because list->data is a void *, so we can't do pointer arithmetic. It would require casting to a char type and then doing the offsets yourself.

include/list.h
```diff
@@ -1,27 +1,107 @@
...
+ * This can be used to write for loops more easily.
+ *
+ * Example (for a list of char *):
+ * 	char **end = list_end(list);
+ * 	for (char **ptr = list->items; ptr < end; ++ptr) {
+ * 		printf("%s\n", *ptr);
+ * 	}
+ */
+// Consider making this inline or __attribute__((pure))__
+void *list_end(list_t *list);
```
SirCmpwn: :+1: 

No thanks for inline/__attribute__((pure)) though.

SirCmpwn:
Looks much better. Just addressing semantics now.

common/list.c
SirCmpwn: >I don't think uint8_t really adds much here. It's all basically "typeless" data that we're not going to touch. "unsigned char" seems good enough for a byte type.

Well, you need the size to be 8 bits. I would use a uint8_t for that.

>It's a pointer to a VLA. The syntax seems a bit tricky and clever, but it really cleans up the other parts of the code. We can treat it like a normal array, where pointer arithmetic will work properly.
>
>No, because list->data is a void *, so we can't do pointer arithmetic. It would require casting to a char type and then doing the offsets yourself.

Ah. Well, alright then.

include/list.h
ascent12: It's just so that if you use it inside a loop condition, the compiler won't keep pointlessly calling the function over and over again.

SirCmpwn: Not important.

common/list.c
ascent12: >you need the size to be 8 bits

Not really. It's just making use of the fact that sizeof(char) is guaranteed to equal 1, and that every other type is guaranteed to be a multiple of char. Nothing actually relies on anything being 8 bits.

SirCmpwn: `sizeof` measures in octets, so `sizeof(uint8_t)` is guaranteed to be 1 too, but it's more explicit about the nature of the type.

ascent12: >sizeof measures in octets

I don't want to be too deep into the semantics of C, as this is a pretty pointless debate, but C uses char as its definition of byte. sizeof says "how many chars big is this". It can be changed to uint8_t if you want, but I feel that it doesn't really convey any useful information to a reader, as there is nothing which relies on anything being 8 bits.

I've always considered unsigned char to be the "byte type" in C, and the stdint types are for cases where you explicitly need a certain size, for some reason.

SirCmpwn: I know char is defined as a byte. I still consider it obsolete in favor of uint8_t.

common/list.c
```diff
@@ -1,134 +1,224 @@
...
 
-void list_insert(list_t *list, int index, void *item) {
-	list_resize(list);
-	memmove(&list->items[index + 1], &list->items[index], sizeof(void*) * (list->length - index));
-	list->length++;
-	list->items[index] = item;
+void list_insert(list_t *list, size_t index, const void *data) {
+	if (!sway_assert(list && data && index <= list->length, "Invalid argument") ||
+		!resize(list)) {
+
```
SirCmpwn: style nit: indent the second line of the condition and remove the extra newline

SirCmpwn:
Looks good. Will merge once everything has been updated to use it correctly.

ascent12:
>Will merge once everything has been updated to use it correctly.

Sweet. I've gotten started, but it's more gargantuan of a task than I would thought it would be. It turns out there are a lot of cases where it's not as simple as changing a function call, and requires restructuring some code.
I'm also going to be busy in the near future, so it might be a while before it gets finished.

SirCmpwn:
No problem, take your time. Ping me if you want any help migrating things.

ascent12:
I've changed the old usage of list_t to the new implementation.
Due to the sheer amount of code of code I changed, and the fact I don't understand a lot of it very well, the chance that I've introduced bugs is pretty high.
I've done basic testing to make sure sway itself works, but I haven't testing all of the features, so there may be bugs and crashes lurking around. This code probably needs some pretty strict review and testing before it gets merged. Some help with that would be nice.

I also added some new functions to the new API, to reduce some of the extremely cast-heavy code when working with lists of pointers (which is all of them at the moment).

I haven't updated a lot of the code to use all of these new functions. The types in the lists haven't changed, the "capacity" argument isn't being taken advantage of anywhere.
There is just far too much for me to do. Hopefully over time as the code gets updated, they can be changed to take advantage of the new features.

One particular area of attention is the bsearch/lsearch functions. The old list_seq_find took functions that are incompatible with bearch (cmp_to/key was passed second), and bsearch gives you pointers to pointers. The lsearch API was changed to match bsearch, as they're supposed to be drop in replacements for each other. I've tried to modify some of the code to fit the new API, but it's pretty likely I've missed some or screwed it up. I don't even know if having bsearch and possibly introducing all of these potentially buggy changes is worth it. It might not be the worst idea to remove this functionality and just keep the old API for this.

common/stringop.c
```diff
@@ -69,26 +69,22 @@ int lenient_strcmp(char *a, char *b) {
...
 	free(copy);
 	return res;
 }
 
 void free_flat_list(list_t *list) {
-	int i;
-	for (i = 0; i < list->length; ++i) {
-		free(list->items[i]);
-	}
-	list_free(list);
```
SirCmpwn: If you're gutting free_flat_list, just drop it and replace its usages with `list_free_withp`.

common/stringop.c
```diff
@@ -69,26 +69,22 @@ int lenient_strcmp(char *a, char *b) {
...
 list_t *split_string(const char *str, const char *delims) {
-	list_t *res = create_list();
+	list_t *res = list_new(sizeof(char *), 0);
 	char *copy = strdup(str);
 	char *token;
 
 	token = strtok(copy, delims);
 	while(token) {
 		token = strdup(token);
-		list_add(res, token);
```
SirCmpwn: This seems incorrect.

common/stringop.c
```diff
@@ -315,22 +311,25 @@ char *join_list(list_t *list, char *separator) {
...
 
-	char *p = res + strlen(list->items[0]);
-	strcpy(res, list->items[0]);
+	char *item = *(char **)list_get(list, 0);
+	char *p = res + strlen(item);
+	strcpy(res, item);
 
-	for (int i = 1; i < list->length; i++) {
+	for (size_t i = 1; i < list->length; i++) {
+		item = *(char **)list_get(list, i);
```
SirCmpwn: Is this sort of thing necessary? Why can't we just put `char*` into the list?

ascent12: Whoops, I get I forgot to run my search-and-replace on the common code. Yeah, it's supposed to use the list_getp function.

ascent12:
I've removed bsearch and changed the lsearch callback interface to match the old one.
It just becomes a mess of pointers to pointers trying to follow bsearch's interface.

SirCmpwn:
Looks good. Going to start running this for a while and merge it in a week or so. I encourage others to run this patch as well and submit feedback.

ascent12:
After talking to @SirCmpwn about this, we've decided that this PR has too much potential to introduce bugs without enough payoff.
This PR ended up being a lot more work than I expected it to be, and I really don't have the motivation to try and try and iron out all of the very subtle bugs this may have caused.

PR #1232, opened 2017-06-05
FreeBSD fixes

johalun:
Increase _POSIX_SOURCE value for localtime_r access.
Increase _XOPEN_SOURCE value for lstat() and various printf() functions access.
Conditionally link to libcap (only on Linux).
Possibly some trailing whitespace fixes (automatic).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,10 @@ find_package(GdkPixbuf)
 find_package(PAM)
 
 find_package(LibInput REQUIRED)
-find_package(Libcap REQUIRED)
+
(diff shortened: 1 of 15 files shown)


SirCmpwn:
Please don't increase _XOPEN_SOURCE globally, but rather only in files where the newer functions are required.

johalun:
After fixing 6-7 files manually I gave up and replace everywhere. I'll give it another shot.

johalun:
Fixed.

SirCmpwn:
Thanks!

Issue #986, opened 2016-12-12
Bounties for sway development


SirCmpwn:
Note: this isn't final! This thread is a proposal that *may* be carried out, but it's not set in stone. Please leave comments in here to discuss the idea.

So I've accepted donations for sway for a while now, and they've now accumulated to a fair sum. I was asked to put some of that towards beer and pizza (which was delicious), but the majority of it remains unspent. So let's spend it! I'd like to start a bounty program for sway to encourage the development of difficult/important features. Not all changes will be eligible for a bounty, but we'll have a few important issues (tracked in this GitHub thread) that will be. I also don't want to limit the bounty pools strictly to sway features - anything that improves the sway experience, even if it's in other projects in the ecosystem, can be considered for a bounty.

# Funds

Currently, sway's pile of money is $201, and grows by $30/mo in automatic donations. I've donated $199 of my own money to bring that up to $400. Not sure what bucket to drop recurring donations in - for now I'll reach out to the donors and see if they want to choose, and if not I'll top up the bounties at my discretion.

I propose to set the initial bounties as such:

# Current bounties

**sway features**

* Layout save/restore: $50
* Real time video capture: $50
* ARM support: $50
* i3-style marks: $20

**wlc features**

* Redshift support: $50
* Touchscreen support: $50
* Modesetting support: $20
* Drag and drop: $20
* Mouse capture: $20

**other**

* Xembed (tray icons) in Xwayland and wlc: $50
* wayland/x11 clipboard sync damon: $20

# Contributing to a bounty

If you'd like to fund support for a specific feature, donate normally and mention what you'd like your donation to go towards:

[![Donate with fosspay](https://drewdevault.com/donate/static/donate-with-fosspay.png)](https://drewdevault.com/donate?project=4)

Feel free to donate towards something there isn't already a bounty for - I'll add one for you.

# Awarding bounties

A pull request that addresses a bounty will not be merged on that merit alone - it will go through the same review process as any other pull request. Anyone interested in claiming a bounty is strongly encouraged to stop by the [sway IRC channel](http://webchat.freenode.net/?channels=sway&uio=d4) or open an issue discussing the problem further to make sure everyone's on the same page and work isn't wasted.

If several people participate in a feature, the bounty will be split between them. The awardees can agree amongst themselves how to split the bounty. If an agreement can't be reached, I'll have the final say on how to distribute a bounty.

[...]

PR #1234, opened 2017-06-08
Implement Tray Icons

4e554c4c:
There are still issues with this commit, mostly in dealing with `xembedsniproxy`. It creates dummy windows for placing XEmbed items which sway renders even though it shouldn't. Also `xembedsniproxy` is a super buggy program, so there still may be undiscovered problems with it.

This commit implements the StatusNotifierItem protocol, and enables
swaybar to show tray icons. It also uses `xembedsniproxy` in order to
communicate with xembed applications.
The tray is completely optional, and can be disabled on compile time
with the `enable-tray` option. Or on runtime with the bar config option
`tray_output none`.

Overview of changes:
In swaybar very little is changed outside the tray subfolder except
that all events are now polled in `event_loop.c`, this creates no
functional difference.

Six bar configuration options were added, these are detailed in
sway-bar(5)

The tray subfolder is where all protocol implementation takes place and
is organised as follows:

tray/sni_watcher.c:
  This file contains the StatusNotifierWatcher. It keeps track of
  items and hosts and reports when they come or go.
tray/tray.c
  This file contains the StatusNotifierHost. It keeps track of
  sway's version of the items and represents the tray itself.
tray/sni.c
  This file contains the StatusNotifierItem struct and all
  communication with individual items.
tray/icon.c
  This file implements the icon theme protocol. It allows for
  finding icons by name, rather than by pixmap.
tray/dbus.c
  This file allows for asynchronous DBus communication.

Note: Package maintainers sholud add xembedsniproxy and libdbus as optional dependencies of sway.

See #986 #343


SirCmpwn:
One commit ;_;

swaybar/bar.c
```diff
@@ -92,6 +120,30 @@ static void mouse_button_notify(struct window *window, int x, int y,
...
+			} else if (button == swaybar.config->context_button) {
+				sni_context_menu(item->ref, x, y);
+			} else if (button == swaybar.config->secondary_button) {
+				sni_secondary(item->ref, x, y);
+			}
+			break;
+		}
+		tray_width -= icon_width;
+	}
+#endif
```
SirCmpwn: Can you move this into another file?

swaybar/bar.c
```diff
@@ -50,18 +57,39 @@ static void spawn_status_cmd_proc(struct bar *bar) {
...
+			if (pid == 0) {
+				execlp("xembedsniproxy", "xembedsniproxy", NULL);
+				_exit(EXIT_FAILURE);
+			}
+			waitpid(pid, &wstatus, 0);
+		} while (!WIFEXITED(wstatus));
+		_exit(EXIT_FAILURE);
+	}
+}
+#endif
```
SirCmpwn: Consider moving this as well, just have a tray upkeep function that checks this and refires it if not running

swaybar/bar.c
```diff
@@ -178,23 +233,54 @@ void bar_setup(struct bar *bar, const char *socket_path, const char *bar_id) {
...
+		/* Start the SNI watcher */
+		init_sni_watcher();
+
+		/* Start the SNI host */
+		init_tray();
+
+		/* Start xembedsniproxy */
+		spawn_xembed_sni_proxy();
+	}
+#endif
```
SirCmpwn: This one too, just make it a simple

```c
#ifdef ENABLE_TRAY
    init_tray();
#endif
```

Then make init_tray do all this work. Try to limit the tray code in this file to simple calls to another function.

swaybar/render.c
```diff
@@ -297,6 +301,72 @@ void render(struct output *output, struct config *config, struct status_line *li
...
+
+		item->dirty = false;
+	}
+
+
+	if (tray_width != window->width * window->scale) {
+		tray_width -= tray_padding;
+	}
+
+no_tray:
```
SirCmpwn: Ditto here. Move it all into a separate function in another file.

SirCmpwn:
Problems I noticed while testing:

* SNI works quite well, nice job. The exception is that I can't seem to select context menu items that are occluded by other windows. Does the compositor need to do something special to let them keep focus? (i.e. open konversation's context menu and try to use any option that is displayed on top of another window, such as the konversation main window). Clicking outside of the context menu also doesn't seem to close it, and should. Making sure the context menu windows are getting focus should fix both of these issues.
* xembed doesn't work at all for me. The issues with it need to be ironed out or it needs to be removed before I'm comfortable merging this.

SirCmpwn:
I also pushed a commit fixing up the cairo blending operator so that partially transparent icons wouldn't overwrite the background

SirCmpwn:
Status?

4e554c4c:
Still working on getting tray windows working, pretty much everything else is in order.

SirCmpwn:
Should I hold off on rc1 for this?

4e554c4c:
I'd say not, unless you really want it there. I can't work on this 100% right now, so it might be good to delay things. Also I'm not even sure if it'd be good to release sni support without xembed included. People who don't know the difference will file reports.

SirCmpwn:
I'm okay with shipping SNI support alone, we already get plenty of stupid support requets and I doubt this will add much. I'll have a look at this in the morning and see if I can clean it up and still feel good about it.

SirCmpwn:
Spent some time poking on this. Results: focus is working good enough for now. Also I figured out what was wrong witht he firefox address bar dropdown and fixed that. Going to include this in the next release and maybe iterate on it a bit during rc's to fix the edge cases.

4e554c4c:
Thanks!

4e554c4c:
I'll continue work on xembed, hopefully we can have that in the next release.

nemanjan00:
Any progress?

SirCmpwn:
Yep.

PR #1235, opened 2017-06-08
include stdint to be compaitable with musl

vaartis:
It doesn't comlile with musl if you don't include it explicitly


SirCmpwn:
Thanks!

PR #1236, opened 2017-06-12
swaybar: Group child processes for signalling

ykrivopalov:
Fixes child proccess killing, when status_command is a complex command
like "i3status | wrapper.sh".

Set the process group id of the child process by calling `setpgid` after
forking and before calling `exec`.

The process group ID will be set to the process ID of the forked
process. Processes spawned by this child process will also have this
group ID.

Send signals to the process group with `killpg`. This will send the
signal to all of the process group.

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -34,6 +34,7 @@ static void spawn_status_cmd_proc(struct bar *bar) {
 			close(pipefd[0]);
 			dup2(pipefd[1], STDOUT_FILENO);
 			close(pipefd[1]);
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

SirCmpwn:
I have reverted this and it's not going to land in 0.14 unless you get it fixed within the next week. It was the cause of #1253

SirCmpwn:
It's not immediately apparent to me how this caused #1253, but there's enough information there to reproduce it so we'll see.

Issue #1233, opened 2017-06-06
Allow audio capture in swaygrab -c


clarcharr:
Would be nice if there were an easy option for this. This would mostly just add `-f alsa -i pulse` or something similar to `ffmpeg`.

Although obviously a more seasoned user could just pass options to `ffmpeg` directly, it'd be nice for usability to just add this option to `swaygrab` itself.

SirCmpwn:
I think I would rather add an environment variable or a flag like `SWAYGRAB_FFMPEG_OPTS` or `--ffmpeg-opts`

clarcharr:
I agree that it makes sense to allow arbitrary ffmpeg opts, but it would also be nice to have a simple default that includes audio so that someone can include audio in the captured video with minimal effort.

SirCmpwn:
I disagree. Depending on your setup there could be multiple ways of capturing audio, each different, plus there'd have to be driver-specific ways of picking the appropriate device and things like mixing in desktop audio with microphone audio... it's not going to happen.

PR #1241, opened 2017-06-18
FFmpeg options for swaygrab

Blajda:
Related to #1233


SirCmpwn:
Document this in the man page too

Blajda:
Will that work? If it does I'll squash the commit

SirCmpwn:
Yes, thanks

SirCmpwn:
Thanks!

PR #1247, opened 2017-06-20
Convert a popen into a fork.

### LINKED ISSUE #1245: swaygrab appears to interpret spaces in filename
When running swaygrab with a filename that contains filenames, it appears to pass the argument unquoted to ImageMagick, as it produces the following kinds of errors (I quote the filename here, so this error is internal to swaygrab):

```
 swaygrab 'test image.png'
convert: unable to open image `test': No such file or directory @ error/blob.c/OpenBlob/2701.
convert: no decode delegate for this image format `' @ error/constitute.c/ReadImage/504.
```

I end up having to escape the space myself inside the string:

```
 swaygrab 'test\ image.png'
```

clarcharr:
It's been a long time since I've done C but I'm fairly sure that this works.

Fixes #1245.

### DIFF
--- a/swaygrab/main.c
+++ b/swaygrab/main.c
@@ -9,6 +9,7 @@
 #include <stdint.h>
 #include <math.h>
 #include <time.h>
+#include <sys/wait.h>
 #include <json-c/json.h>
 #include "log.h"
 #include "ipc-client.h"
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Build failure, please address

clarcharr:
It's been a very long time but I just fixed the build failure.

SirCmpwn:
Thank you!

PR #1254, opened 2017-07-01
Fix #926

### LINKED ISSUE #926: Interact with Wayland clipboard via cli
As long as there's no clarification on #913 (how to manage interaction between X and Wayland clipboard) I'd love to interact with sway's clipboard via a cli application such as swaymsg. I'm using [pass](https://passwordstore.org) which is using xclip by default. It would be lovely to pipe passwords from pass (or any other cli application) into the Wayland clipboard.

nyorain:
Adds the 'clipboard' command that can currently only be used
like this: 'swaymsg clipboard Some clipboard text here'. In future it could be extended to set non-text data or read from the clipboard but this would require extensions to wlc (or has to wait for wlroots).
Note that due to sway command arguments passing this will
not preserve whitespace in the clipboard text correctly (i.e. two spaces will be interpreted as one).
Requires Cloudef/wlc#272 to work correctly.

Still not too familiar with the sway codebase, I am not sure this is the correct way to add new commands to sway, or if they have to be added somewhere else.

### DIFF
--- a/sway/commands/clipboard.c
+++ b/sway/commands/clipboard.c
@@ -1,0 +1,38 @@
+#include <wlc/wlc.h>
+#include <unistd.h>
+#include <string.h>
+#include "sway/commands.h"
+#include "stringop.h"
+
+static void send_clipboard(void *data, const char *type, int fd) {
+	if (strcmp(type, "text/plain") != 0
+			&& strcmp(type, "text/plain;charset=utf-8") != 0) {
+		close(fd);
+		return;
+	}
+
+	const char *str = data;
+	write(fd, str, strlen(str));
+	close(fd);
+}
+
+struct cmd_results *cmd_clipboard(int argc, char **argv) {
+	static char *current_data = NULL;
+
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "clipboard", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+
+	static const char *types[2] = {
+		"text/plain",
+		"text/plain;charset=utf-8"
+	};
+
+	char *str = join_args(argv, argc);
+	wlc_set_selection(str, types, 2, &send_clipboard);
+
+	free(current_data);
+	current_data = str;
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+}
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -103,6 +103,7 @@ sway_cmd cmd_client_unfocused;
 sway_cmd cmd_client_urgent;
 sway_cmd cmd_client_placeholder;
 sway_cmd cmd_client_background;
+sway_cmd cmd_clipboard;
 sway_cmd cmd_commands;
 sway_cmd cmd_debuglog;
 sway_cmd cmd_default_border;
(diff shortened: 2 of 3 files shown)


sway/commands/clipboard.c
```diff
@@ -0,0 +1,37 @@
+#include <wlc/wlc.h>
+#include <unistd.h>
+#include <string.h>
+#include "sway/commands.h"
+#include "stringop.h"
+
+static void send_clipboard(void *data, const char *type, int fd) {
+	if (strcmp(type, "text/plain") != 0 && strcmp(type, "text/plain;charset=utf-8") != 0) {
```
SirCmpwn: Limit this to 80 columns please

SirCmpwn:
This looks right, but I don't want to close #926 until we have the ability to read from the clipboard, too.

sway/commands/clipboard.c
```diff
@@ -0,0 +1,38 @@
+#include <wlc/wlc.h>
+#include <unistd.h>
+#include <string.h>
+#include "sway/commands.h"
+#include "stringop.h"
+
+static void send_clipboard(void *data, const char *type, int fd) {
+	if ((strcmp(type, "text/plain") != 0) &&
+			(strcmp(type, "text/plain;charset=utf-8") != 0)) {
```
SirCmpwn: Sorry, more style nits. You don't need these extra parens and you should put `&&` on the next line.

sway/commands/clipboard.c
```diff
@@ -0,0 +1,38 @@
...
+		close(fd);
+		return;
+	}
+
+	const char *str = data;
+	write(fd, str, strlen(str));
```
SirCmpwn: I presume that you don't need to write the null delimiter?

sway/commands/clipboard.c
```diff
@@ -0,0 +1,38 @@
...
+
+	static const char *types[2] = {
+		"text/plain",
+		"text/plain;charset=utf-8"
+	};
+
+	const char *str = join_args(argv, argc);
+	wlc_set_selection((void*) str, types, 2, &send_clipboard);
+
+	free((void*) current_data);
```
SirCmpwn: Casting to `void*` here is unneccessary and probably on `wlc_set_selection` as well.

nyorain: did it because str is a const pointer and wlc_set_selection/free require a non-const pointer

sway/commands/clipboard.c
nyorain: ugh... actually not sure it writes to the fd of a client, is there a null delimiter needed/expected?

sway/commands/clipboard.c
SirCmpwn: Well why did you make it const then?

sway/commands/clipboard.c
nyorain: it works without at least for all common applications... the clients can determine the length from the number of bytes read

SirCmpwn: It could work both ways and I'd err on the side of no null delimiter, but so long as you've tested it this way then I trust that it works.

sway/commands/clipboard.c
nyorain: oh, good point...

sway/commands/clipboard.c
nyorain: Oversaw this, is it ok now? Still getting used to this codestyle

SirCmpwn:
Looks good, thank! Think you could squash the commits for me?

Issue #895, opened 2016-09-12
Setting wallpaper silently fails


qlyoung:
On a fresh Ubuntu 16.04.1 netinstall, after build + install + config of latest wlc and sway, the wallpaper remains unset and displays only a black screen. This occurs when using a single color wall as well.

Log: https://spit.mixtape.moe/view/27812180
Conf: https://spit.mixtape.moe/view/0066fe57

Relevant snippets:

```
09/11/16 20:47:35 - handling config command 'output LVDS-1 res 1366x768 pos 0,0 bg ~/pictures/wallpapers/eva-01.jpg fill'
09/11/16 20:47:35 - find_handler(output) 0
09/11/16 20:47:35 - Config stored for output LVDS-1 (enabled:-1) (1366 x 768 @ 0, 0 scale 1) (bg /home/hardware/pictures/wallpapers/eva-01.jpg fill)
09/11/16 20:47:35 - handling config command 'output HDMI-A-1 res 1920x1080 pos 1366,0 bg ~/pictures/wallpapers/eva-01.jpg fill'
09/11/16 20:47:35 - find_handler(output) 0
09/11/16 20:47:35 - Config stored for output HDMI-A-1 (enabled:-1) (1920 x 1080 @ 1366, 0 scale 1) (bg /home/hardware/pictures/wallpapers/eva-01.jpg fill)
```

later...

```
09/11/16 20:47:36 - Setting background for output 0 to /home/hardware/pictures/wallpapers/eva-01.jpg
09/11/16 20:47:36 - Setting background for output 1 to /home/hardware/pictures/wallpapers/eva-01.jpg
```

but background remains solid black.

serussell:
Hi @qlyoung,

I'm finding it nigh-on impossible to install Sway on Ubuntu; I can't get WLC to install.  There are no packages for it, and the requirements information is skeletal at best.  I did my best dredging up the various dependencies, and am now stuck at a bunch of undefined references during linking libwlc, and Google fails to find any references to most of the missing symbols.

So: any pointers on how you got as far as getting Sway to actually install?  This is beyond bleeding edge, and I understand that. It's built on a foundation of bleeding edges, from Wayland on up.  Any pointers to how you got past the worst stumbling blocks you encountered would be a service.

P.S.  This question has nothing to do with this ticket, but since Github removed any ability to direct message a user, this is about the only way to get a message to somebody.  Sorry about that.

4e554c4c:
[...]

PR #1256, opened 2017-07-02
Do not fail silently when background is not PNG

lheckemann:
#895 is caused by unclear documentation, which this fixes.

EDIT: actually swaybg should probably also report problems loading the image
EDIT2: actually it already does, but cairo doesn't seem to fail loading when given a jpeg instead of a png
EDIT3: Made swaybg break if background image can't be loaded

### DIFF
--- a/swaybg/main.c
+++ b/swaybg/main.c
@@ -97,6 +97,14 @@ int main(int argc, const char **argv) {
 		if (!image) {
 			sway_abort("Failed to read background image.");
 		}
+		if (cairo_surface_status(image) != CAIRO_STATUS_SUCCESS) {
+			sway_abort("Failed to read background image: %s."
+#ifndef WITH_GDK_PIXBUF
+					"\nSway was compiled without gdk_pixbuf support, so only"
+					"\nPNG images can be loaded. This is the likely cause."
+#endif //WITH_GDK_PIXBUF
+					, cairo_status_to_string(cairo_surface_status(image)));
+		}
 		double width = cairo_image_surface_get_width(image);
 		double height = cairo_image_surface_get_height(image);
 


sway/sway.5.txt
```diff
@@ -361,7 +361,8 @@ The default colors are:
 
 **output** <name> <background|bg> <file> <mode>::
 	Sets the wallpaper for the given output to the specified file, using the given
-	scaling mode (one of "stretch", "fill", "fit", "center", "tile").
+	scaling mode (one of "stretch", "fill", "fit", "center", "tile"). This image
+	must be in PNG format unless sway was built with gdk_pixbuf.
```
SirCmpwn: I don't like the change in the docs here

swaybg/main.c
```diff
@@ -97,6 +97,9 @@ int main(int argc, const char **argv) {
 		if (!image) {
 			sway_abort("Failed to read background image.");
 		}
+		if (cairo_surface_status(image) != CAIRO_STATUS_SUCCESS) {
+			sway_abort("Failed to read background image: %s", cairo_status_to_string(cairo_surface_status(image)));
+		}
```
SirCmpwn: This change is fine, though.

sway/sway.5.txt
lheckemann: What's wrong with it? What should I change?

SirCmpwn: Just don't include this change.

lheckemann: Then how are future occurrences of #895 meant to be avoided? Trying to load a JPEG with the PNG loader gives the rather unhelpful "out of memory" error message.

lheckemann:
I've removed the doc change, but would still like to know how reoccurrences of #895 will be avoided if the requirement for the image to be a PNG is not documented.

sway/sway.5.txt
SirCmpwn: Well, you also improved the message in this pull request, so I'm not really concerned. You could extend the error message to include the likely root cause if you like.

lheckemann:
Added reference to PNG in the error message.

SirCmpwn:
Can you clarify that this is because sway was compiled without gdk_pixbuf support, and that other formats could be available if you were to recompile?

lheckemann:
Done.

SirCmpwn:
Looks good, but could you narrow the scope of the "extra" message and just use a new #ifdef block? Also, How about this message: "Sway was compiled without gdk_pixbuf support, which means only PNG images are supported. This is the likely cause."

Thanks for your patience with all of these requests.

lheckemann:
How about just removing the variable entirely? (see last commit)

SirCmpwn:
I like it. Thanks!

qlyoung:
This would have been nice to see in the docs for those of us who aren't familiar with sway's compile time options.

I guess an error message is a decent breadcrumb though...

Either way, thanks for the improvement.

PR #1258, opened 2017-07-03
Fix zsh completion for _swaylock

refacto:
There was a missing ' in _swaylock, prevents zsh from providing completion.


SirCmpwn:
Thanks!

PR #1259, opened 2017-07-03
Disable -Wunused-result diagnostic

Hummer12007:
It caused unpredictable build failures with different build environments


SirCmpwn:
To further clarify the intention of this, glibc is way too stringent with these and in many cases we just don't give a shit about i.e. ignoring the return value of `write`. And gcc fails to implement the generally-accepted means of explicitly silencing this on a case-by-case basis with `(void)write(...)`.

PR #1263, opened 2017-07-07
Implement get_clipboard ipc message

nyorain:
At the moment get_clipboard will fail if the clipboard is empty or does not contain textual data (saying this in its error message) and the pretty print of it will simply output nothing. Is this expected behaviour or should e.g. an empty string be returned or should pretty print output the error message?

As this is my first larger contribution to sway there might be things wrong with it, i tested it and it works for me but better have a closer look. I may have forgotten a place the new command must be mentioned/handled.

Edit: this pr finally finishes #926 for now (which i have seemed to close already with the wlc pr, didn't know that github is doing this, sorry).


include/sway/config.h
```diff
@@ -235,8 +235,9 @@ enum ipc_feature {
 	IPC_FEATURE_EVENT_WINDOW = 2048,
 	IPC_FEATURE_EVENT_BINDING = 4096,
 	IPC_FEATURE_EVENT_INPUT = 8192,
+    IPC_FEATURE_GET_CLIPBOARD = 16384,
```
SirCmpwn: Should be tab, not spaces

sway/ipc-server.c
```diff
@@ -55,6 +55,11 @@ struct get_pixels_request {
 	struct wlc_geometry geo;
 };
 
+struct get_clipboard_request {
+   struct ipc_client *client;
+   struct wlc_event_source *event_source;
```
SirCmpwn: Ditto

sway/ipc-server.c
```diff
@@ -322,6 +327,46 @@ void ipc_get_pixels(wlc_handle output) {
 	ipc_get_pixel_requests = unhandled;
 }
 
+static int ipc_selection_data_cb(int fd, uint32_t mask, void *data)
+{
```
SirCmpwn: Also join this { onto the previous line.

sway/ipc-server.c
```diff
@@ -322,6 +327,46 @@ void ipc_get_pixels(wlc_handle output) {
...
+	assert(data);
+	struct get_clipboard_request *req = (struct get_clipboard_request*) data;
```
SirCmpwn: `(struct get_clipboard_request *)data;`

sway/ipc-server.c
```diff
@@ -322,6 +327,46 @@ void ipc_get_pixels(wlc_handle output) {
...
+		sway_log(L_ERROR, "Selection data fd error");
+		const char *error = "{ \"success\": false, \"error\": "
+			"\"Could not receive text data from clipboard\" }";
+		ipc_send_reply(req->client, error, (uint32_t)strlen(error));
+		goto cleanup;
+	}
+
+	if (mask & WLC_EVENT_READABLE) {
+		char buf[512];
+		int ret = read(fd, buf, 511);
```
SirCmpwn: What if the clipboard is longer than 512 bytes?

sway/ipc-server.c
```diff
@@ -322,6 +327,46 @@ void ipc_get_pixels(wlc_handle output) {
...
+			const char *error = "{ \"success\": false, \"error\": "
+				"\"Could not receive text data from clipboard\" }";
+			ipc_send_reply(req->client, error, (uint32_t)strlen(error));
+			goto cleanup;
+		}
+
+		buf[ret] = '\0';
+		json_object *obj = json_object_new_object();
+		json_object_object_add(obj, "success", json_object_new_boolean(true));
+		json_object_object_add(obj, "content", json_object_new_string(buf));
```
SirCmpwn: Also include the mimetype, and for mimetypes that aren't `text/*`, base64 encode the content.

https://github.com/SirCmpwn/aerc/blob/master/src/util/base64.c

SirCmpwn: It might also make sense to update `ipc_send_reply` to handle large payloads that cannot (or should not) be written all in one `write`

sway/ipc-server.c
```diff
@@ -565,6 +610,72 @@ void ipc_client_handle_command(struct ipc_client *client) {
...
 	}
 
+   case IPC_GET_CLIPBOARD:
+   {
+		if (!(client->security_policy & IPC_FEATURE_GET_CLIPBOARD)) {
+			goto exit_denied;
+		}
+
+		size_t size;
+		const char **types = wlc_get_selection_types(&size);
```
SirCmpwn: Can everything from this point forward be moved into a separate function?

SirCmpwn:
Noted issues aside, looks good.

sway/ipc-server.c
nyorain: at the moment this does only support text. Not sure in which way other targets should be chosen, if e.g. image/png image/jpg and other non-text targets are supported which one should be used?will fix the style issues soon.

SirCmpwn: You can make the IPC command accept a preferred mimetype as input, and if omitted you could just send all of them (or the list, and wait for a second request to get the content?)

nyorain: ok will look into it, returning the list of supported types as default sounds good (querying all of them does probably not make much sense imagine e.g. a large image offered in 5 different formats...)

SirCmpwn: Would be nice to offer wildcards, so that you could i.e. use `swaymsg -t get_clipboard "text/*"` in a script.

SirCmpwn: While you're at this, can you make swaymsg use a nonzero exit code if a command resulted in an error?

nyorain: ok. yes, wildcards sound possible and are a good idea to not query for "text/plain" and "text/plain;charset=utf-8" every time.

nyorain:
I hope the new commit fixes all notes.
Support for receiving multiple clipboard targets at once can be easily enabled later on. The wildcard matching currently only supports "*" and is implemented rather simple, so it might struggle with more complex use cases (it works for e.g. "*" or "text/*" etc).
Targets imported from x11 like STRING or UTF8_STRING will currently be encoded in base64 but they are no real wayland targets so should not matter in the end (everything that does not begin with "text/" is base64 encoded).

SirCmpwn:
>Targets imported from x11 like STRING or UTF8_STRING will currently be encoded in base64 but they are no real wayland targets so should not matter in the end (everything that does not begin with "text/" is base64 encoded).

We should special case these and treat them like `text/plain` (or `text/plain;charset=utf-8`)

sway/ipc-server.c
```diff
@@ -322,6 +331,203 @@ void ipc_get_pixels(wlc_handle output) {
...
+		data[current] = '\0';
+
+		if (strncmp(req->type, "text/", 5) == 0) {
+			json_object_object_add(req->json, req->type,
+				json_object_new_string(data));
+		} else {
+			size_t outlen;
+			char *b64 = b64_encode(data, current, &outlen);
+			json_object_object_add(req->json, req->type,
+					json_object_new_string(b64));
```
SirCmpwn: Can you add `;base64` to the mimetype? i.e. `image/png` becomes `image/png;base64`

sway/ipc-server.c
```diff
@@ -322,6 +331,203 @@ void ipc_get_pixels(wlc_handle output) {
...
+		int ret = 0;
+		int current = 0;
+
+		// read data as long as there is data avilable
+		// grow the buffer step_size in every iteration
+		do {
+			if (data == NULL) {
+				data = malloc(step_size);
+			} else {
+				data = realloc(data, current + step_size);
```
SirCmpwn: Test for failure here

SirCmpwn: Otherwise a malicious client could just never stop sending us data until we OOM and then blow up

SirCmpwn: Let's also set a sane upper limit on the size of this in general, don't go higher than 8MiB. And double the size rather than adding 512 to it.

sway/ipc-server.c
```diff
@@ -322,6 +331,203 @@ void ipc_get_pixels(wlc_handle output) {
...
+			}
+
+			ret = read(fd, data + current, step_size - 1);
+			if (ret < 0) {
+				sway_log_errno(L_ERROR, "Reading from selection data fd failed");
+				goto cleanup;
+			}
+
+			current += ret;
+		} while (ret == step_size - 1);
```
SirCmpwn: Yeah this whole loop could be made better in general.

```
len = 512, i = 0, buf = malloc(len)
do {
  amt = read(fd, &buf[i], len - i - 1) // note handling of null delimiter here
  i += amt
  if i > len {
    if len == 8MiB { free buf, error out }
    _buf = realloc(buf, len *= 2)
    if !_buf { free buf, error out }
    buf = _buf
  }
}
```

SirCmpwn: Also put a timeout on this to avoid a slow post attack

SirCmpwn: And instead of trying to do this all at once, let the readable signal get hit several times and build the buffer up over time

nyorain: as long as there is any data to read from the fd the wayland event loop will probably call this function in a loop so that this would only introduce additional complexity without making it more secure i believe. Shouldn't the 8mb data limit be enough? the data is only sent once so such an attack could only happen from the ipc side (request loop) in which case a malicious client got access to the clipboard ipc feature which should not happen in the first place...

nyorain: a timeout seems reasonable, sometimes clients just don't answer a request (or an error gets lost somewhere in the x11 <-> wayland conversion, will work on it but it might take a while).

SirCmpwn: >as long as there is any data to read from the fd the wayland event loop (using poll)

That isn't how it works. You `read` until EoF here, you don't just read what's available (note, I haven't read your updated commit yet). You block until the whole thing is read and then you destroy the event source.

sway/ipc-server.c
```diff
@@ -349,43 +349,53 @@ static int ipc_selection_data_cb(int fd, uint32_t mask, void *data) {
 	}
 
 	if (mask & WLC_EVENT_READABLE) {
-		static const int step_size = 512;
-		char *data = NULL;
-		int ret = 0;
-		int current = 0;
+		static const int max_size = 8192 * 1000;
```
SirCmpwn: MiB, use * 1024

sway/ipc-server.c
```diff
@@ -349,43 +349,53 @@ static int ipc_selection_data_cb(int fd, uint32_t mask, void *data) {
...
+				break;
+
+			i += amt;
+			if (i >= len - 1) {
+				if (len >= max_size) {
+					sway_log(L_ERROR, "selection data too large");
+					free(buf);
+					goto cleanup;
+				}
+				char *next = realloc(buf, (len *= 2));
```
SirCmpwn: The extra parens are not necessary

sway/ipc-server.c
```diff
@@ -349,43 +349,53 @@ static int ipc_selection_data_cb(int fd, uint32_t mask, void *data) {
...
-		do {
-			if (data == NULL) {
-				data = malloc(step_size);
-			} else {
-				data = realloc(data, current + step_size);
-			}
+		for(;;) {
+			int amt = read(fd, buf + i, len - i - 1);
+			if (amt <= 0)
+				break;
```
SirCmpwn: Use {} here

sway/ipc-server.c
```diff
@@ -349,43 +349,53 @@ static int ipc_selection_data_cb(int fd, uint32_t mask, void *data) {
...
+					sway_log(L_ERROR, "selection data too large");
+					free(buf);
+					goto cleanup;
+				}
+				char *next = realloc(buf, (len *= 2));
+				if (!next) {
+					sway_log_errno(L_ERROR, "relloc failed");
+					free(buf);
+					goto cleanup;
+				}
```
SirCmpwn: Your error messages could be more descriptive. L_ERROR is bubbled up to the user by default, they should be able to understand them.

SirCmpwn:
bump

nyorain:
Had not had time the last few days to reimplement the data reading and add a timeout callback, will hopefully get this done at some time during the next days.

SirCmpwn:
Sounds good. Take your time, I'm just checking in.

nyorain:
Does this fix the issues? Currently waiting 1 second for the clipboard owner to send the data.
What to return to the ipc client when a requested format cannot be received? Currently, an array with all requested formats is returned (although it is limited to one format atm), how to signal an error? the current implementation just leaves the failed format out of the array but this may not be expected behavior (just an empty array on failure atm).

SirCmpwn:
1 second is a bit strict. It could be like a video clip or something. Maybe 30 seconds is better? Will review the code tonight.

SirCmpwn:
Signal an error with a standard IPC error response.

SirCmpwn:
Also... not gonna make you implement this but you might consider writing partial JSON blobs and base64 encoding the data from the user on the fly, so that we can avoid copying it all to memory.

sway/ipc-server.c
```diff
@@ -322,6 +336,266 @@ void ipc_get_pixels(wlc_handle output) {
...
+			json_object_object_add(obj, "content", json_object_new_string(req->buf));
+			json_object_object_add(req->json, req->type, obj);
+		} else {
+			size_t outlen;
+			char *b64 = b64_encode(req->buf, req->buf_position, &outlen);
+			json_object_object_add(obj, "content", json_object_new_string(b64));
+			free(b64);
+
+			char *type = malloc(strlen(req->type) + 8);
+			strcat(type, ";base64");
```
nyorain: By the way i don't think it is a good idea to add ";base64" to the mimetype after some considerations. The client e.g. requests the data with "image/png" mimetype and the answer will only contain the "image/png;base64" mimetype. The current implementation only adds ";base64" if the data is successfully retrieved - could be extended with additional checks - but i think we shouldn't do it at all.

SirCmpwn: We could also change the offer strings?

nyorain:
Yeah, i will not make this any more complicated. I already think features like the base64 (basically all non-text data) encoding will rarely be used and they have already make this whole implementation huge and rather complicated (it might make sense to move it to a separate file by now).

sway/ipc-server.c
nyorain: yeah but then we would have convert them back when the client requests the data with this mimetype. And how to know if the ;base64 was added by sway when sending the supported types or was already present? We would have to guess or try 2 attempts... Probably not worth it

SirCmpwn:
Well, non-text clipboard data is not uncommon at all. Maybe among hackers it's less common but the typical user may find themselves using it a lot more often. Graphics peeps too.

sway/ipc-server.c
SirCmpwn: Well, JSON does not support binary data. We must convert it to a string somehow.

SirCmpwn: I guess we could drop the JSON act and just dump the data directly into the socket like we do for screenshots, but screenshots are a hack too and I'd rather not proliferate hacky approaches.

nyorain:
But what are usecases to receive the base64 encoded; json ecoded data using an ipc program for sway? I mean it is implemented now but don't make this more complicated please.

SirCmpwn:
I mean, if you want to get any data other than text data out of the clipboard, you need to encode it in a form that can be used in JSON.

SirCmpwn:
I can work on finishing this feature up if you're burned out on it.

sway/ipc-server.c
nyorain: No i meant only the mimetype (but yes using binary data would be nice at some point). When we receive a data request for "image/jpg;base64" we would have to make sure somehow that the base64 was manually added by ourselves and not already present in the mime type offered by the clipboard owner.

SirCmpwn: I think that would be an unusual situation, but if it does occur, `image/jpeg;base64;base64` is a valid mimetype afaik

SirCmpwn: Still, not that hard to just test if a string ends in `;base64` and skip adding it ourselves if that's the case.

nyorain:
For now i will probably not have time to see to it, it does not sound that simple to implement.

sway/ipc-server.c
nyorain: yes that could be implemented that way (if it is really important enough). It would make the request and answer mimetypes at least match each other again, which was my primary point.

SirCmpwn: We can't just not handle binary data, even if it's a bitch.

nyorain: Well it would be possible to use a mechanism comparable to the one used in wayland data requests, i.e. passing an fd over the socket. But probably not worth it.

SirCmpwn:
Let's merge this for now and make it more sophisticated later.

PR #1270, opened 2017-07-10
fix use-after-free bug

tobiasblass:
The _bindsym_ command accesses the token list after it has been freed. This patch prepares the error message before freeing the list.


SirCmpwn:
Thanks!

PR #1272, opened 2017-07-12
Use WLC v2 pointer interface

ascent12:
Uses Cloudef/wlc#274, which fixes the slow cursor movements not working.
I have tested this, but it wasn't intensive. I basically checked that the slow cursor thing was fixed, and that was about it.
Not a whole lot of code needed to change, so it's probably fine.


SirCmpwn:
Thanks!

PR #1276, opened 2017-07-13
Don't trust SNI names, fixes #1274

### LINKED ISSUE #1274: swaybar crashes with wicd-gtk
Steps to reproduce: compile swaybar with tray support and run wicd-gtk
Result: swaybar crashes.

The log of swaybar -d is attached.

[swaybar_crash.txt](https://github.com/SirCmpwn/sway/files/1143968/swaybar_crash.txt)

4e554c4c:
If an item doesn't have a well-formed name, it will not be added to the
tray.


SirCmpwn:
Thanks!

PR #1283, opened 2017-07-17
Fix memory leaks in swaybar tray

4e554c4c:
These are a few memory leaks in the tray I found while on the plane. Of the leaks that seem to come from the tray, there is one that occurs on `swaybar/tray/sni.c:81` that I could not figure out, but that's it. There are also many leaks that come from pango calls, and a few in glib, but I did not have the resources to investigate and fix these.


SirCmpwn:
Thanks!

PR #1285, opened 2017-07-20
[swaybar] fix non-dbus build

tobiasblass:
Swaybar's CMakeLists.txt uses the enable-tray option directly to decide whether to build the tray. This leads to a compilation error if dbus is not installed.
This patch uses the ENABLE_TRAY variable instead, which is only true if the user enabled the tray *and* dbus is available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,7 @@ endif()
 
 if (enable-tray)
 	if (DBUS_FOUND)
-		set(ENABLE_TRAY)
+		set(ENABLE_TRAY TRUE)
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Seems additional work is required to make this work correctly.

tobiasblass:
I'm sorry, apparently I changed the default value of enable-tray during my debugging. This made my "all variables on default" test build rather pointless :-)

It turns out that CMake interprets _set(VARIABLE)_ as _unset(VARIABLE)_ (probably for some legacy reason). Setting the variable to true solves the issue.

SirCmpwn:
Thanks!

PR #1288, opened 2017-07-24
security: enabled ipc on freebsd

johalun:


SirCmpwn:
Command security *should* work fine on FreeBSD, best to leave that alone.

SirCmpwn:
Thanks!

PR #1294, opened 2017-07-31
Allow swaylock indicator size to be configurable

4e554c4c:


swaylock/main.c
```diff
@@ -383,6 +386,8 @@ int main(int argc, char **argv) {
 		{"separatorcolor", required_argument, NULL, 0},
 		{"keyhlcolor", required_argument, NULL, 0},
 		{"bshlcolor", required_argument, NULL, 0},
+		{"indicator_radius", required_argument, NULL, 0},
+		{"indicator_thickness", required_argument, NULL, 0},
```
SirCmpwn: Use dash-case here.

SirCmpwn:
tyvm

PR #1303, opened 2017-08-04
Fix #892

### LINKED ISSUE #892: Freeze when scrolling fast between workspaces
Sway freezes when I open multiple workspaces and then use the mousewheel in the bar to switch between them very quickly.
Setting `wrap_scroll` does not make a difference.
I can not reproduce the issue when switching via bash like this (so I assume it's not the speed but something with the mousewheel):

```
for i in `seq 1 1000`; do
  swaymsg workspace prev_on_output
  swaymsg workspace next_on_output
done
```

The only options I have when the freeze happens is to hibernate the laptop using the power key (which does not help because it's still frozen when resuming) and to reboot.

Do you need additional information?

Versions (all Arch Linux packages):
sway version: 0.9-1:
wlc: 0.0.5-1
wayland: 1.11.0-1
[sway.log.txt](https://github.com/SirCmpwn/sway/files/465554/sway.log.txt)
Thanks for a awesome project :)

ilyaluk:
Removed additional event socket to reduce synchronization problems
and limited workspace update requests count.

### DIFF
--- a/include/ipc-client.h
+++ b/include/ipc-client.h
@@ -24,6 +24,10 @@ char *get_socketpath(void);
  */
 int ipc_open_socket(const char *socket_path);
 /**
+ * Issues a single IPC command without reading response.
+ */
+void ipc_single_command_no_response(int socketfd, uint32_t type, const char *payload, uint32_t *len);
+/**
  * Issues a single IPC command and returns the buffer. len will be updated with
  * the length of the buffer returned from sway.
  */
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -88,7 +88,7 @@ void free_ipc_response(struct ipc_response *response) {
 	free(response);
 }
 
-char *ipc_single_command(int socketfd, uint32_t type, const char *payload, uint32_t *len) {
+void ipc_single_command_no_response(int socketfd, uint32_t type, const char *payload, uint32_t *len) {
 	char data[ipc_header_size];
 	uint32_t *data32 = (uint32_t *)(data + sizeof(ipc_magic));
 	memcpy(data, ipc_magic, sizeof(ipc_magic));
@@ -102,6 +102,10 @@ char *ipc_single_command(int socketfd, uint32_t type, const char *payload, uint3
 	if (write(socketfd, payload, *len) == -1) {
 		sway_abort("Unable to send IPC payload");
 	}
+}
+
+char *ipc_single_command(int socketfd, uint32_t type, const char *payload, uint32_t *len) {
+	ipc_single_command_no_response(socketfd, type, payload, len);
 
 	struct ipc_response *resp = ipc_recv_response(socketfd);
 	char *response = resp->payload;
--- a/include/swaybar/bar.h
+++ b/include/swaybar/bar.h
@@ -15,6 +15,8 @@ struct bar {
 	int ipc_socketfd;
 	int status_read_fd;
 	pid_t status_command_pid;
+
+	int pending_ipc_requests;
 };
 
 struct output {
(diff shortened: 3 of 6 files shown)


include/ipc-client.h
```diff
@@ -24,6 +24,10 @@ char *get_socketpath(void);
  */
 int ipc_open_socket(const char *socket_path);
 /**
+ * Issues a single IPC command without reading response.
+ */
+void ipc_single_command_no_response(int socketfd, uint32_t type, const char *payload, uint32_t *len);
```
SirCmpwn: Don't we need to read the response for further reads to make sense? Also, if chaning the client fixes this then I'm skeptical of the fix, malicious IPC clients shouldn't be able to crash sway.

SirCmpwn:
Yeah, this needs to be fixed at the sway level, though I'm not opposed to throttling swaybar too.

include/ipc-client.h
ilyaluk: Response is being read by `handle_ipc_event`.

SirCmpwn: Is there an advantage to reading it then instead of here? The response is generally meant to be specific to that request.

ilyaluk: Probably request rate should be limited on sway side too.

SirCmpwn: We can probably avoid the deadlock without resorting to rate limiting, no? Is that really the problem?

ilyaluk: Yep, this solves deadlock problems. In i3 it's implemented like this too.

SirCmpwn: Alright then. Can you add the sway side to this PR?

ilyaluk: In i3 there are one socket and requests made without reading response. Then, one handler handles events and responses.

ilyaluk: Okay, I'll try.

ilyaluk: I'm not sure how to throttle on sway side. Since all requests processed synchronous, we cannot limit request count (without counting requests in last _n_ seconds), but we could make socket non-blocking, which will help with locks, but on high request rates there may be unexpected problems with data loss.

SirCmpwn: Data loss sounds better than crashing to me.

ilyaluk:
Looks like after making socket non-blocking scrolling works fine without throttling on swaybar side.

SirCmpwn:
I figured that might work. Can you rollback the other changes?

ilyaluk:
Well, most changes are still needed (it hangs with event socket), but I can rollback throttling.

SirCmpwn:
Well, throttling is fine. I'm still not on board with ipc_single_command_no_response.

ilyaluk:
Okay, I'll try to fix it with two sockets and without _no_response, but I'm not sure if it will work.

ilyaluk:
I've played a bit with nonblocking sockets on sway side. Data loss could lead to client crashes, for example if ipc response lost. Probably it will be okay with throttling on client side, but nonblocking I/O may cause other unexpected bugs with current implementation.

Generally, the problem is that if client subscribed to some events and not reading them (e.g. swaybar not reading from event socket while processing scroll events), socket buffer fills and then on next `write` to that socket sway blocks.

[Example](https://gist.github.com/ilyaluk/2f74588975499ce44ca6e164a8c17700)
Open at least two workspaces and run this script.

ilyaluk:
Ah, and non-blocking patch is pretty simple. [Link](https://gist.github.com/ilyaluk/4cba205f94b71e5f62d412a615706c20)

SirCmpwn:
Do you also have to deal with E_AGAIN?

ilyaluk:
Uh huh. I've never worked with async IO before, though.

SirCmpwn:
grep for async in the read(3) man page, it's pretty straightforward. Are you going to update this PR?

ilyaluk:
Yeah, I'll figure it out.

No, I'll make another one.

SirCmpwn:
Sounds good. Many thanks!

PR #1313, opened 2017-08-08
Implement nonblocking IO in IPC server

### LINKED ISSUE #892: Freeze when scrolling fast between workspaces

ilyaluk:
Added client write buffer and handler for writable status on client
socket.

Also fixes #892.

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -40,11 +40,15 @@ static const char ipc_magic[] = {'i', '3', '-', 'i', 'p', 'c'};
 
 struct ipc_client {
 	struct wlc_event_source *event_source;
+	struct wlc_event_source *writable_event_source;
 	int fd;
 	uint32_t payload_length;
 	uint32_t security_policy;
 	enum ipc_command_type current_command;
 	enum ipc_command_type subscribed_events;
+	size_t write_buffer_len;
+	size_t write_buffer_size;
+	char *write_buffer;
 };
 
 static list_t *ipc_get_pixel_requests = NULL;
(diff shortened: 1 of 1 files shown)


sway/ipc-server.c
```diff
@@ -249,6 +275,48 @@ int ipc_client_handle_readable(int client_fd, uint32_t mask, void *data) {
...
+		return 0;
+	}
+
+	memmove(client->write_buffer, client->write_buffer + written, client->write_buffer_len - written);
+	client->write_buffer_len -= written;
+
+	if (client->write_buffer_len == 0 && client->writable_event_source) {
+		wlc_event_source_remove(client->writable_event_source);
+		client->writable_event_source = NULL;
+	}
```
SirCmpwn: This looks like it would cause issues if you didn't managed to clear out the write buffer all in one go, or for long lasting clients. No?

ilyaluk: If client managed to fill its socket buffer, it might happen again. But we can shrink buffer in this case.

SirCmpwn: Gotcha, I was actually misreading some of this code.

SirCmpwn:
LGTM, will test it later today.

SirCmpwn:
Thanks!

Issue #143, opened 2015-08-27
Doesn't work on armv7h (RPi2)


ArchieT:
Hi,

I am running Arch Linux on my Raspberry Pi 2 B, and I have installed sway from the AUR.

When I try to run it in tty, it does nothing.

```
$ sway
$
```

When I try to run it on i3, a white window appears for a split second.

Also, apparently returns 0:

```
$ sway && echo teststring
teststring
$
```

Are you gonna think of support for the ARM architecture in the near future? I would love to have sway working on my RPi, as Wayland is much more optimized for it then X11.

SirCmpwn:
Can you run with `sway -d 2> sway.log` to enable debug logging, and post the contents of sway.log?

ARM support is mostly up to wlc.

ArchieT:
Hm, after update which I performed few minutes ago I didn't try -d, and now it outputs some messages.

Contents of `sway.log`:

```
Loading config
Handling command 'set $home /home/mf'
Handling command 'set $config /home/mf/.config'
Checking for config at /home/mf/.sway/config
Checking for config at /home/mf/.config/sway/config
Handling command 'set $mod Mod4'
Handling command 'set $term urxvt'
Handling command 'set $menu dmenu_run'
Handling command 'bindsym $mod+Return exec $term'
bindsym - Bound Mod4+Return to command exec urxvt
Handling command 'bindsym $mod+Shift+q kill'
bindsym - Bound Mod4+Shift+q to command kill
Handling command 'bindsym $mod+d exec $menu'
bindsym - Bound Mod4+d to command exec dmenu_run
Handling command 'floating_modifier $mod'
Handling command 'bindsym $mod+Shift+c reload'
bindsym - Bound Mod4+Shift+c to command reload
Handling command 'bindsym $mod+Shift+r restart'
bindsym - Bound Mod4+Shift+r to command restart
Handling command 'bindsym $mod+Shift+e exit'
bindsym - Bound Mod4+Shift+e to command exit
Handling command 'bindsym $mod+h focus left'
bindsym - Bound Mod4+h to command focus left
Handling command 'bindsym $mod+j focus down'
bindsym - Bound Mod4+j to command focus down
Handling command 'bindsym $mod+k focus up'
bindsym - Bound Mod4+k to command focus up
Handling command 'bindsym $mod+l focus right'
bindsym - Bound Mod4+l to command focus right
Handling command 'bindsym $mod+Left focus left'
bindsym - Bound Mod4+Left to command focus left
[...]

PR #1327, opened 2017-08-14
Detecting Raspberry PIs and advising to use correct video mode

refacto:
This is a pull request meant to help with people experiencing the problems detailed in #143 by detecting if sway is being run on a Raspberry Pi and if the vc4 module is loaded. If the module isn't loaded, it's safe to assume the Pi isn't in the correct video mode, and the error message contains instructions to boot with the correct video mode.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -53,6 +53,46 @@ static void wlc_log_handler(enum wlc_log_type type, const char *str) {
 	}
 }
 
+void detect_raspi() {
+	bool raspi = false;
+	FILE *f = fopen("/sys/firmware/devicetree/base/model", "r");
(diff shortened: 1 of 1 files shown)


refacto:
I can add some explanation in the FAQ or wiki if there's demand.

SirCmpwn:
Thanks!

refacto:
Just a quick question -- Anything I could do to close the bounty for ARM support? ARM support is essentially here now.

SirCmpwn:
Gotta wait until it ships in a stable version and has a couple weeks of user testing. Need an incentive for you to stick around and help fix bugs!

PR #1322, opened 2017-08-14
Translate README.md to Spanish / Traduce el README.md al Español

jdiez17:
Lo único con lo que no estoy completamente satisfecho es la traducción de "release" a "versión lanzada" pero no he encontrado ninguna otra palabra mejor.

The only thing I'm not entirely happy with is the translation of "release" to "versión lanzada" (literal TL: launched version) but I have not found a better phrase for it.

Issue #1318, opened 2017-08-13
Documentation i18n


SirCmpwn:
* [ ] ja_JP: @SirCmpwn
  * [x] README (#1317)
  * [ ] Man pages
    * [ ] sway(1)
    * [ ] sway(5)
    * [ ] sway-bar
    * [ ] sway-input
    * [ ] swaymsg
    * [ ] swaygrab
    * [ ] swaylock
    * [ ] sway-security
  * [ ] FAQ

abdelq:
fr_FR

SirCmpwn:
Thanks, added it to the list.

SirCmpwn:
I wrote up some TL guidelines in the OP, please review them.

Hummer12007:
- [ ] uk_UA:
  - [x] README (#1325) @Hummer12007 (not up for anything else)
  - [ ] Man pages:
    - [ ] sway(1)
    - [ ] sway(5)
    - [ ] sway-bar
    - [ ] sway-input
    - [ ] swaymsg
    - [ ] swaygrab
    - [ ] swaylock
    - [ ] sway-security
  - [ ] FAQ

kon14:
- [ ] el_GR: @kon14
    - [x] README (#1317)
    - [ ] Man pages
        - [ ] sway(1)
        - [ ] sway(5)
        - [ ] sway-bar
        - [ ] sway-input
        - [ ] swaymsg
        - [ ] swaygrab
        - [ ] swaylock
        - [ ] sway-security
    - [ ] FAQ

I plan on doing the rest soon. Usually afking in IRC (kon14), GMT+2

SirCmpwn:
Standard "this language is not English so here's how to get support" README subsection:

>## [Language] help
>
>[Username] provides support in [language] on IRC and GitHub in [timezone].

In Japanese I'm going to write:

>## 日本語助けの
>
>SirCmpwnはIRCとGitHubでUTC-4に日本語助けがある。

Please include similar information in your translations.

SirCmpwn:
Sorry for frequent updates to these guidelines, i18n is a WIP.

SirCmpwn:
I pushed a branch with some support code for localizing man pages:

https://github.com/SirCmpwn/sway/tree/localized-manpages

Unfortunately I've run into an issue with how fucking shit asciidoc is. I think I'm just going to write a new man page authoring tool this weekend.

kon14:
[...]

PR #1329, opened 2017-08-14
Translate README to spanish

hobarrera:
Regarding #1318

I started this last night, but looks like #1322 beat me to it. I'm not sure how much localization is acceptable (it might be too much to keep both variations), so I'll leave it up to you maintainers if you want to keep both or not.


SirCmpwn:
Double check this matches the latest English readme in terms of information content and read through the updated TL guidelines - it looks like you're good on the latter but does your last section match the new English section?

SirCmpwn:
@jdiez17 @hobarrera can we get one of these squared away please?

PR #1351, opened 2017-09-11
fixes a hanging swaygrab

Ongy:
https://github.com/SirCmpwn/sway/issues/1350 is fixed with this.
The commit that change swaygrab to use fork instead of Popen tried to
write to the read end of the pipe in the child branch and exec in the
parent branch.
It also didn't add "convert" for `arg[0]` of the child, which caused it to skip the first argument.
This commit fixes those and closes the write fd after writing,
so convert actually exits.


SirCmpwn:
Thanks!

PR #1368, opened 2017-09-23
swaybar protocol: handle button 4 and 5

johnae:
i3 bar handles scroll events as button 4 and 5. An example use case would be volume up/down.
This probably has some effect on the "scrolling between workspaces" feature - now one probably has to put the pointer over the workspace buttons for it to work.


SirCmpwn:
Thanks for the patch!

>now one probably has to put the pointer over the workspace buttons for it to work.

This isn't going to work - can you fix this?

johnae:
Yeah, I'm very unfamiliar with the code base atm - just started looking at it today. Should be quite easy to fix I believe.

johnae:
@SirCmpwn There. Much better overall i.m.o and workspace switching should work as before (unless over the status bar items ofc).

SirCmpwn:
Yes, that looks much better, thanks!

Issue #1287, opened 2017-07-22
Variable prefix of another variable


emersion:
I have this config file:

```
set bg #323232
set bgu #000000
background $bgu
```

It expands to:

```
background #323232u
```

Which leads to this error:

```
[config.c:716] Error on line 189 'background $bgu': Invalid color definition #000000u (/home/simon/.config/sway/config)
```

I'm pretty sure this config file used to work before.

Note that i3 had this bug too, but it's now fixed: https://github.com/i3/i3/issues/2235

RedSoxFan:
This appears to be working in both `0.15` and `master`.

PR #1379, opened 2017-10-02
Fix issue #1287

karjonas:
The variable matching is now looping through all variables trying to find the longest possible match.

### DIFF
--- a/sway/commands/set.c
+++ b/sway/commands/set.c
@@ -30,7 +30,7 @@ struct cmd_results *cmd_set(int argc, char **argv) {
 		if (!tmp) {
 			return cmd_results_new(CMD_FAILURE, "set", "Not possible to create variable $'%s'", argv[0]);
 		}
-		snprintf(tmp, size, "$%s", argv[0]);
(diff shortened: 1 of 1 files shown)


4e554c4c:
The variables should already be added to the list longest->shortest, (see sway/commands/set.c:10)

SirCmpwn:
That hasn't been working right for some time, I think.

4e554c4c:
Yeah, but the fix should either fix that code or remove it imo

SirCmpwn:
Aye.

karjonas:
It seems like the sorting of variables is already handled so this pull request is pointless. **But**, there is a bug that is due to the code that handles variables with a $ prefix, it missing the last character of the variable name. The issue is sway/commands/set.c:33 should have 'size+1' instead of 'size'. Should I update this PR with the correct fix?

SirCmpwn:
Yes, please do.

karjonas:
That should do it.

SirCmpwn:
Thanks!

PR #1384, opened 2017-10-06
Fix two issues with spaces in workspace names

lbonn:
In two cases, the workspace name was cropped until the first space:

* when moving a container to a workspace
* when finding a suitable name for a workspace (by parsing the configuration file)

The `workspace xx` command is already behaving correctly but there might be some other instances of this error.


SirCmpwn:
Thanks!

PR #1385, opened 2017-10-06
Security config: skip hidden files

lbonn:
I think it's safe to ignore dot files there.

What happened: sway crashed because I kept a file there, opened in vim (it creates a `.xxx.swp`).


SirCmpwn:
Thanks!

PR #1387, opened 2017-10-08
ipc: various fixes for i3 compat

### LINKED ISSUE #1386: ipc: harmonize `get_tree` with i3
The tree contains most of the information but does not respect the latest i3 spec: https://i3wm.org/docs/ipc.html#_tree_reply

* no `focus` array (might be difficult?)
* in i3 responses, all of these fields are mandatory for all nodes (even `root` and `output`), whereas sway adds them selectively. In many cases, a default value of `null` or `[]` should suffice.

These issues may hinder interoperability with some existing ipc libraries.

lbonn:
Fixes #1386 + other inconsistencies with i3's behavior.

The goal was to be able to use [focus-last.py](https://github.com/acrisci/i3ipc-python/blob/master/examples/focus-last.py) (and [my rust version](https://github.com/lbonn/i3-focus-last)) with sway. It seems to work now, if permissions are set up correctly.

The program is not that trivial and uses window events, `get_tree` and commands with criteria: this was a good way to test the ipc implementation.

Major highlight: 3 criteria missing in sway, taken from i3

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -114,6 +114,7 @@ static char *crit_tokens(int *argc, char ***buf, const char * const criteria_str
 
 	char **argv = *buf = calloc(max_tokens, sizeof(char*));
 	argv[0] = base; // this needs to be freed by caller
+	bool quoted = true;
 
 	*argc = 1; // uneven = name, even = value
 	while (*head && *argc < max_tokens) {
@@ -134,7 +135,8 @@ static char *crit_tokens(int *argc, char ***buf, const char * const criteria_str
 				if (*(namep) == ' ') {
 					namep = strrchr(namep, ' ') + 1;
 				}
-				argv[(*argc)++] = namep;
+				argv[*argc] = namep;
+				*argc += 1;
 			}
 		} else if (*head == '"') {
 			if (*argc % 2 != 0) {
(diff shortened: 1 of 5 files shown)


sway/criteria.c
```diff
@@ -137,21 +144,35 @@ static char *crit_tokens(int *argc, char ***buf, const char * const criteria_str
...
-				"Names must be unquoted, values must be quoted");
+		} else if (*argc % 2 == 0 && *head != ' ') {
+			// parse unquoted values
+			if (!valp) {
+				quoted = false;
+				valp = head;  // value starts here
+			}
+		} else if (valp && !quoted && *head == ' ') {
+			// value ends here
+			argv[(*argc)++] = valp;
```
SirCmpwn: Can you split this on to two lines? I don't like the dereference-increment operation going on here.

sway/criteria.c
```diff
@@ -137,21 +144,35 @@ static char *crit_tokens(int *argc, char ***buf, const char * const criteria_str
...
+			*head = '\0';
+			valp = NULL;
+			namep = head + 1;
 		}
 		head++;
 	}
+
+	// catch last unquoted value if needed
+	if (valp && !quoted && !*head) {
+		argv[(*argc)++] = valp;
```
SirCmpwn: Ditto

sway/criteria.c
lbonn: Yes, I agree. I picked the existing style of the function.

SirCmpwn:
Thanks!

PR #1394, opened 2017-10-12
Interlink swaymsg to sway(5) and hint that command is interchargeable

madblobfish:
Added a hint where to find the commands.
Also indicate that command is itself not a command.


SirCmpwn:
Thanks!

madblobfish:
You are fast :) thank you!

PR #1397, opened 2017-10-14
Change install folder for man pages

johalun:
This should make man pages install in correct place across platforms.


SirCmpwn:
Thanks!

Issue #1396, opened 2017-10-14
_XOPEN_SOURCE and _POSIX_C_SOURCE causing trouble on FreeBSD


johalun:
Even the highest values on any these cause 
```
#define SOCK_CLOEXEC 0x10000000
#define SOCK_NONBLOCK 0x20000000
```
(and maybe more) to be hidden on FreeBSD (__BSD_VISIBLE = 0).
Optimal for FreeBSD is not using these macros at all. Is it possible to move all these declarations to some Linux specific header?

johalun:
For now it's only one file, ipc-server.c. Maybe can make some exception for that file. The other files are ok with bumping xopen source to 700.

SirCmpwn:
You can add ifdefs to that file to avoid defining them on FreeBSD if you like.

johalun:
Maybe this is the culprit? 
`set(CMAKE_C_STANDARD 99)`
Setting c99 basically hides everything on FreeBSD.. 
```
#elif defined(_C99_SOURCE)	/* Localism to specify strict C99 env. */
#define	__POSIX_VISIBLE		0
#define	__XSI_VISIBLE		0
#define	__BSD_VISIBLE		0
#define	__ISO_C_VISIBLE		1999
#define	__EXT1_VISIBLE		0
```

SirCmpwn:
You can bump it to 11 if that makes it easier?

johalun:
No, 11 is the same. 
Here's a Q/A about similar issues. 
https://lists.freebsd.org/pipermail/freebsd-standards/2004-March/000474.html
and the source
https://github.com/freebsd/freebsd/blob/master/sys/sys/cdefs.h#L748-L773

SirCmpwn:
This is good now, right?

johalun:
It's working for now, although not an elegant solution.

johalun:
Maybe it would be nice if these macros could be defined in the cmake config files instead of headers.

ascent12:
I agree that setting it on a per-file base is pretty pointless.
It's fully expected that POSIX.1-2008 is supported. Having it defined per-file isn't going to change anything.
If it removes a lot of clutter and makes it easier for other platforms, it's even more reason to put it in the build system.

SirCmpwn:
It's not like I'm unaware that you can define POSIX macros when you invoke the compiler. This was a deliberate design decision.

ascent12:
[...]

PR #1398, opened 2017-10-14
Fix build on FreeBSD adjusting/removing _XOPEN_SOURCE declaration.

johalun:
Reference #1396

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -1,6 +1,10 @@
 // See https://i3wm.org/docs/ipc.html for protocol information
 
+#ifndef __FreeBSD__
+// Any value will hide SOCK_CLOEXEC on FreeBSD (__BSD_VISIBLE=0)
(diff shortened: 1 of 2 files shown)


Zer0-One:
You probably don't want to avoid defining that on *everything* except Linux, you just want to avoid defining it on FreeBSD. And before doing that, you might want to ask the FreeBSD people about their SUSv4 compliance.

johalun:
Changed to not freebsd and added a comment so we'll know why. maybe there's a better solution but this is what I got so far..

myfreeweb:
Uhhh I guess you missed what I ended up doing in wlroots:

```c
#ifdef __FreeBSD__
#define __BSD_VISIBLE 1
#endif
```

johalun:
@myfreeweb 
> #define __BSD_VISIBLE 1
It solves the problem but I don't think we should define things that are internal to the implementation.

PR #1400, opened 2017-10-14
Various manpage edits

madblobfish:
Like last time, only more fixes and I completely rewrote the config file paragraph. 
Thanks for the good software :)

Also I'm unsure if you want to change "criteria" to "criterion" ([wictionary](https://en.wiktionary.org/wiki/criterion) for reference).
And there is [this paragraph](https://github.com/swaywm/sway/blob/master/sway/sway.5.txt#L386-L388) which I'm not sure of what it means.

### DIFF
--- a/sway/sway-bar.5.txt
+++ b/sway/sway-bar.5.txt
@@ -11,8 +11,8 @@ sway-bar - bar configuration file and commands
 Description
 -----------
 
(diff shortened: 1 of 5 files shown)


SirCmpwn:
Thanks!

PR #1401, opened 2017-10-15
Fixes to changing scale factor with output command

sideshowdave7:
Poking around with the output command with `swaymsg -t command output` I noticed a couple oddities with setting scaling on the fly.

Without this change, only the first output command would cause the scaling factor to change.  This change allows subsequent output commands to adjust scaling, as well as revert the scaling factor back to 1 (the default).

Tested by:

untitled


SirCmpwn:
LGTM, thanks! Will mege this once the build is happy.

PR #1412, opened 2017-10-17
handlers: preserve parent pointer value

Hummer12007:


SirCmpwn:
Why?

Hummer12007:
https://github.com/Hummer12007/sway/blob/9e36204b230fb9ecc7eaab55c4eacb457e071cdd/sway/handlers.c#L580 is called with the NULL pointer otherwise.

SirCmpwn:
We should rearrange the container for which fullscreen was set (likely the workspace).

Hummer12007:
@SirCmpwn ?

SirCmpwn:
aight, thanks.

PR #1419, opened 2017-10-20
Print log level even if STDERR is not a tty

4e554c4c:
Makes reading debug logs much easier, debug lines will start with `D`,
info lines with `I` and error lines with `E`.


SirCmpwn:
Thanks!

PR #1424, opened 2017-10-22
swaygrab: Prevent segfault if IPC response can't be parsed.

ggreer:
The default max nesting depth of json-c is 32, which can cause some valid trees to fail to be parsed. Instead of segfaulting, just abort and print the error returned by json-c.

Ideally this code would increase the nesting depth or manually recurse if the JSON was too deep, but this behavior is better than segfaulting.

### DIFF
--- a/swaygrab/json.c
+++ b/swaygrab/json.c
@@ -4,6 +4,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include "log.h"
 #include "ipc-client.h"
 #include "swaygrab/json.h"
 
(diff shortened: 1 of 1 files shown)


ggreer:
In case you're curious, here's the JSON response that caused swaygrab to segfault for me: https://gist.github.com/ggreer/11efa2ddb0e39b4604e9e00e5a728b04

I'm not sure how, but I accidentally created a bunch of nested splits.

swaygrab/json.c
```diff
@@ -12,7 +13,11 @@ static json_object *tree;
 void init_json_tree(int socketfd) {
 	uint32_t len = 0;
 	char *res = ipc_single_command(socketfd, IPC_GET_TREE, NULL, &len);
-	tree = json_tokener_parse(res);
+  enum json_tokener_error error = json_tokener_success;
```
emersion: Tabs!

ggreer: I have no idea how that happened. I blame sublime text.

SirCmpwn:
Hmm, can we increase the limit to something more reasonable? Maybe 256?

ggreer:
OK, increased to 256.

BTW while testing this, I found that swaygrab will segfault if IPC returns a message such as ` { "success": false, "error": "Permission denied" }`

SirCmpwn:
Thanks! LGTM, will merge when the build completes.

Handling permission failure is covered in a separate ticket, feel free to send another patch along if you like.

PR #1425, opened 2017-10-23
swaygrab: Add some error handling.

ggreer:
- If IPC response contains `success: false`, abort and print error message.
- If tree has no nodes, abort with error msg instead of segfaulting.

### DIFF
--- a/swaygrab/json.c
+++ b/swaygrab/json.c
@@ -21,6 +21,14 @@ void init_json_tree(int socketfd) {
 	if (!tree || tok->err != json_tokener_success) {
 		sway_abort("Unable to parse IPC response as JSON: %s", json_tokener_error_desc(tok->err));
 	}
+	json_object *success;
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Looks good. Would you consider adding similar errors to the other IPC utilities?

ggreer:
I was thinking about that. A `json_object* ipc_parse_and_check(char *res, uint32_t len)` might make sense. Give it an IPC response, and it parses with a 256 max depth. Then it either returns a JSON object or aborts with an error message (can't parse JSON, success is false, etc).

Another option would be for the function to return an error (or take a pointer to an error struct/string/whatever). That way any calling code could choose to abort or recover.

This codebase is new to me though, so I'm erring on the side of making smaller changes.

SirCmpwn:
https://en.wikipedia.org/wiki/Wikipedia:Be_bold

PR #1427, opened 2017-10-24
Use add_compile_options instead of interpolation

ranisalt:
Uses CMake's `add_compile_options` directive instead of interpolating  `CMAKE_C_FLAGS` for adding compilation flags.

Also, I have a question: why adding `-g` to all build types instead of recommending `RelWithDebInfo` (essentially `Release` with `-g`) as default?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,16 +2,12 @@ cmake_minimum_required(VERSION 3.1.0)
 
 project(sway C)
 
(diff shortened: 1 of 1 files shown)


SirCmpwn:
>why adding -g to all build types instead of recommending RelWithDebInfo (essentially Release with -g) as default?

Cause distro maintainers who remove debugging symbols should be drawn and quartered, that's why. We gotta make life difficult for thme.

SirCmpwn:
Thanks!

Issue #1429, opened 2017-10-24
detect_proprietary does not check nvidia_drm module parameters


dlrobertson:
Currently [`detect_proprietary`] uses `/proc/cmdline` to check if `nvidia-drm.modeset=1` was passed as a kernel parameter ([sway/main.c:138-148]). This approach works, but the following cases are not covered with this approach.

When module options are set using `modprobe` but `nvidia-drm.modeset` is not set via a kernel parameter, this check will fail despite the module being loaded with the correct parameters. For example, the following will fail.

```
cat > /etc/modprobe.d/nvidia-drm.conf << EOF
options nvidia_drm modeset=1
EOF
# or
modprobe -r nvidia_drm
modprobe nvidia_drm modeset=1
sway
```

In addition, given a kernel build with `CONFIG_DRM` or `CONFIG_DRM_KMS_HELPER` not set, the modeset option should not be recognized by `nvidia_drm`, but the current checks will pass.

I think checking `/sys/module/nvidia_drm/parameters/modeset` and ensuring it contains `Y` should cover the above edge cases. I have a local branch using this approach, and it seems to work. I can submit a PR later if this approach seems like the right direction.

I'm not sure what the equivalent would be for FreeBSD. Maybe keep the current method of checking `/proc/cmdline` for `nvidia-drm.modeset=1` on FreeBSD, but use `/sys/module/nvidia_drm/parameters/modeset` for Linux?

[sway/main.c:138-148]: https://github.com/swaywm/sway/blob/master/sway/main.c#L138-L148
[`detect_proprietary`]: https://github.com/swaywm/sway/blob/master/sway/main.c#L96

SirCmpwn:
@johalun for FreeBSD questions. Does sway+nvidia blob even work on FreeBSD?

So far as Linux is concerned that seems like an acceptable approach to me, but we're dropping support for the proprietary driver the release after next. We can merge a fix for 0.15... but it'll be dropped right away.

johalun:
I'm not sure about the state of Nvidia. It's been a long time since I used it... I can ask around.

PR #1430, opened 2017-10-24
nvidia: Validate the nvidia_drm module options

dlrobertson:
When the proprietary nvidia driver is used, ensure the modeset option is
set instead of checking /proc/cmdline for nvidia-drm.modeset=1.

Fixes: #1429

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -135,6 +135,29 @@ void detect_proprietary() {
 				"You need nvidia, nvidia_modeset, nvidia_uvm, and nvidia_drm."
 				"\x1B[0m\n");
 		}
+#ifdef __linux__
+		f = fopen("/sys/module/nvidia_drm/parameters/modeset", "r");
+		if (f) {
+			char *line = read_line(f);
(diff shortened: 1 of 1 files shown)


sway/main.c
```diff
@@ -135,6 +135,25 @@ void detect_proprietary() {
...
 		}
+#ifdef __linux__
+        f = fopen("/sys/module/nvidia_drm/parameters/modeset", "r");
+        if (f) {
+            char *line = read_line(f);
+            if (line && strstr(line, "Y")) {
+                // nvidia-drm.modeset is set to 0
+				fprintf(stderr, "\x1B[1;31mWarning: You must load "
+					"nvidia-drm with the modeset option on to use "
+					"the proprietary driver.\x1B[0m\n");
```
SirCmpwn: This feedback should be more actionable.

dlrobertson: Updated. I used something similar to the original help string since it shouldn't be affected by the users choice of init system

SirCmpwn:
Thanks!

PR #1434, opened 2017-10-25
Add option to select geometry on swaygrab

### LINKED ISSUE #1421: swaygrab: support region selection
swaygrab is working nicely but it would be nice to be able to select a window or a region with the mouse instead of taking the full screen for producing the screenshot (like [maim/slop](https://github.com/naelstrof/maim))

ranisalt:
This fixes #1421 and I use it as `swaygrab -g $(slop)` as slop works fine on Wayland.

I had to shift other options padding a bit to fit the geometry model (`width x height +- x +- y`).

I could make it work with slop directly like maim does, but I don't feel it's necessary.

### DIFF
--- a/swaygrab/main.c
+++ b/swaygrab/main.c
@@ -168,6 +168,7 @@ int main(int argc, char **argv) {
 	char *output = NULL;
 	int framerate = 30;
 	bool grab_focused = false;
+	char *geometry = NULL;
 
 	init_log(L_INFO);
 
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Sorry, no new features are being accepted at this time.

cauebs:
Wouldn't it be best to leave this open?

SirCmpwn:
No.

luctrudeau:
Seeing that screenshot options under wayland are limited, This feature would be a great addition to swaygrab. Is there another way to take a screenshot for only a region of the screen?

ReK2Fernandez:
@SirCmpwn so when does the new version with the merge coming out? is there an approx ETA?

SirCmpwn:
>is there an approx ETA?

No.

ammgws:
@luctrudeau For now you could try something like [this](https://github.com/ammgws/dotfiles/blob/master/fish/functions/screenshot.fish#L70-L73) or [this](https://github.com/ammgws/dotfiles/blob/master/fish/functions/screenshot.fish#L75-L7) to screenshot a specific region.

PR #1436, opened 2017-10-26
Add Address Sanitized build type

dlrobertson:
- Add `-DCMAKE_BUILD_TYPE=ASAN` to possible build types

I've found address sanitized builds to be super helpful for finding leaks etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,17 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 add_compile_options(-Wall -Wextra -Wno-unused-parameter -Wno-unused-result -Werror)
 
+# Add Address Sanitiezed build type
+set(CMAKE_C_FLAGS_ASAN
+    "${CMAKE_C_FLAGS_DEBUG} -fsanitize=address -fno-omit-frame-pointer"
+    CACHE STRING "Flags used by the C compiler during address sanitizer builds."
+    FORCE )
+mark_as_advanced(
+    CMAKE_C_FLAGS_ASAN
+    CMAKE_EXE_LINKER_FLAGS_DEBUG
+    CMAKE_SHARED_LINKER_FLAGS_DEBUG
+    )
+
 list(INSERT CMAKE_MODULE_PATH 0
 	${CMAKE_CURRENT_SOURCE_DIR}/CMake
 	)
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,11 @@ compiler:
   - gcc
   - clang
 
+env:
+  - BUILD_TYPE=Release
+  - BUILD_TYPE=Debug
+  - BUILD_TYPE=ASAN
+
 arch:
   packages:
     - cmake
@@ -19,7 +24,7 @@ arch:
     - wlc-git
     - libcap
   script:
-    - "cmake ."
+    - "cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE ."
     - "make"
 
 script:


SirCmpwn:
Can you summarize any issues you've found arise from this change?

dlrobertson:
http://clang.llvm.org/docs/AddressSanitizer.html is a good overview of address sanitizers. Running an address sanitized build is similar to running a process with `valgrind`. For example, the following is a snippet from running an address sanitized build of sway.

```
Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x7f666d0ffe60 in __interceptor_malloc /var/tmp/portage/sys-devel/gcc-6.4.0/work/gcc-6.4.0/libsanitizer/asan/asan_malloc_linux.cc:62
    #1 0x4131c5 in cmd_exec_always ../sway/commands/exec_always.c:76
    #2 0x412ba1 in cmd_exec ../sway/commands/exec.c:14
    #3 0x40db72 in handle_command ../sway/commands.c:471
    #4 0x4509ce in handle_binding_command ../sway/handlers.c:715
    #5 0x450cd9 in handle_bindsym ../sway/handlers.c:740
    #6 0x451960 in handle_key ../sway/handlers.c:847
    #7 0x7f666ce071c2  (/usr/lib64/libwlc.so.0+0x141c2)

Direct leak of 22 byte(s) in 9 object(s) allocated from:
    #0 0x7f666d093160 in __interceptor_strdup /var/tmp/portage/sys-devel/gcc-6.4.0/work/gcc-6.4.0/libsanitizer/asan/asan_interceptors.cc:560
    #1 0x475812 in workspace_next_name ../sway/workspace.c:69
    #2 0x43e6fb in new_output ../sway/container.c:201
    #3 0x44d104 in handle_output_created ../sway/handlers.c:171
    #4 0x7f666ce02dff  (/usr/lib64/libwlc.so.0+0xfdff)
```

You'll run across a few false positives, but as a whole they tend to be pretty good.

CMakeLists.txt
```diff
@@ -9,6 +9,31 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
 
+# Add Address Sanitiezed build type
+set(CMAKE_C_FLAGS_ASAN
+    "-g -ggdb -fsanitize=address -fno-omit-frame-pointer"
+    CACHE STRING "Flags used by the C compiler during address sanitizer builds."
+    FORCE )
+set(CMAKE_CXX_FLAGS_ASAN
+    "${DEFAULT_DEBUG} ${DEV_WARNINGS} -std=c++11 -fsanitize=address -fno-omit-frame-pointer"
+    CACHE STRING "Flags used by the C++ compiler during address sanitizer builds."
+    FORCE )
```
SirCmpwn: We don't use C++ in Sway.

dlrobertson: I'm pretty sure it still has to be specified. Event if it must be specified, this should be a blank string. I'll look into this and potentially update this later.

dlrobertson: > I'm pretty sure it still has to be specified. 

I was wrong. Removed in the update.

CMakeLists.txt
```diff
@@ -9,6 +9,27 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+    ""
+    CACHE STRING "Flags used for linking binaries during address sanitizer builds."
+    FORCE )
+set(CMAKE_SHARED_LINKER_FLAGS_ASAN
+    ""
+    CACHE STRING "Flags used by the shared libraries linker during address
+    sanitizer builds."
+    FORCE )
+mark_as_advanced(
+    CMAKE_CXX_FLAGS_ASAN
```
SirCmpwn: .

CMakeLists.txt
```diff
@@ -9,6 +9,26 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 add_compile_options(-Wall -Wextra -Wno-unused-parameter -Wno-unused-result -Werror)
 
+# Add Address Sanitiezed build type
+set(CMAKE_C_FLAGS_ASAN
+    "-g -ggdb -fsanitize=address -fno-omit-frame-pointer"
```
ranisalt: I wouldn't add `-ggdb` to a build since it is not portable - obviously, it adds info that is gdb-only. Also, I'd base it on `CMAKE_C_FLAGS_DEBUG` and add just `-fsanitize=address`, and you don't need any of the below added lines.

dlrobertson: Updated

CMakeLists.txt
```diff
@@ -9,6 +9,17 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+# Add Address Sanitiezed build type
+set(CMAKE_C_FLAGS_ASAN
+    "${CMAKE_C_FLAGS_DEBUG} -fsanitize=address -fno-omit-frame-pointer"
+    CACHE STRING "Flags used by the C compiler during address sanitizer builds."
+    FORCE )
+mark_as_advanced(
+    CMAKE_C_FLAGS_ASAN
+    CMAKE_EXE_LINKER_FLAGS_DEBUG
+    CMAKE_SHARED_LINKER_FLAGS_DEBUG
+    )
```
ranisalt: Why `mark_as_advanced` means here?

SirCmpwn:
Thanks!

CMakeLists.txt
dlrobertson: There is no need for these variables to be visible any cmake GUIs, so we can mark them as advanced. Actually `CMAKE_EXE_LINKER_FLAGS_DEBUG` and `CMAKE_SHARED_LINKER_FLAGS_DEBUG` will already be marked as advanced, so this could be compressed to `mark_as_advanced(CMAKE_C_FLAGS_ASAN)`

PR #1438, opened 2017-10-27
Explicitly setting the version of json-c required

### LINKED ISSUE #1355: Problems with git master versions of json-c
Short version:  It won't compile out of the box.  If do you change the code in order to compile it, it will crash whenever you use any key binding.  I eventually decided to see if there were earlier tags of json-c that wouldn't have this problem and it seems that even with the latest one, version 0.12.1, sway compiles and runs just fine.  Maybe the maximum version of json-c with which sway will work should be indicated in README.

Long version:
When compiling, you get type mismatch errors in five files, all involving json_object_array_length in for loops:
`
sway/ipc-server.c: In function ‘ipc_client_handle_command’:
sway/ipc-server.c:449:21: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   for (int i = 0; i < json_object_array_length(request); i++) {
`

The five files are:
* sway/ipc-server.c
* swaybar/status_line.c
* swaybar/tray/tray.c
* swaygrab/json.c
* swaylock/main.c

The solution for that is very simple.  Merely change the type of the counter variables from int to size_t.

On pressing keys, I've traced the problem to the line in ipc-server.c:
```c
json_object_object_add(obj, "binding", sb_obj);
```

My guess is that the call to `json_object_put(obj);` later in the function frees the objects recursively and then sb_obj is used (or more likely freed, as the stdout error I get from glibc is `*** Error in 'sway': corrupted double-linked list: `) later, causing an Earth-shattering kaboom.

Here's the log file.  I used Lynx because I couldn't figure out how to upload a file with my usual browser (though, I know, Lynx should be be usual browser; with Lynx, all I had to do was ^Xe:e ~/.sway.log&lt;Return&gt;VGy:e \#1&lt;Return&gt;p:wq) and I seem not to have been logged into to github, so it's anonymous gist.  Hmm.
 https://gist.github.com/anonymous/fe18bb44c7d40d5d63fbf15427e70fdc

mariusor:
to 0.12.1. This is needed because the development

 version breaks the existing API of json_object_array_length()

 by moving the return from int to size_t.

This would fix #1355

### DIFF
--- a/CMake/FindJsonC.cmake
+++ b/CMake/FindJsonC.cmake
@@ -8,10 +8,22 @@
 #
 
 find_package(PkgConfig)
-pkg_check_modules(PC_JSONC QUIET JSONC)
+
+if (JsonC_FIND_REQUIRED)
+	set(_pkgconfig_REQUIRED "REQUIRED")
+else()
+	set(_pkgconfig_REQUIRED "")
+endif()
+
+if(JsonC_FIND_VERSION)
+    pkg_check_modules(PC_JSONC ${_pkgconfig_REQUIRED} json-c=${JsonC_FIND_VERSION})
+else()
+    pkg_check_modules(PC_JSONC ${_pkgconfig_REQUIRED} json-c)
+endif()
+
 find_path(JSONC_INCLUDE_DIRS NAMES json-c/json.h HINTS ${PC_JSONC_INCLUDE_DIRS})
 find_library(JSONC_LIBRARIES NAMES json-c HINTS ${PC_JSONC_LIBRARY_DIRS})
-
 include(FindPackageHandleStandardArgs)
+
 find_package_handle_standard_args(JSONC DEFAULT_MSG JSONC_LIBRARIES JSONC_INCLUDE_DIRS)
 mark_as_advanced(JSONC_LIBRARIES JSONC_INCLUDE_DIRS)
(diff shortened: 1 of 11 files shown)


SirCmpwn:
Can you file a bug with json-c asking them not to ship the change? Breaking API changes are typically not a good idea.

mariusor:
@SirCmpwn the changes required for changing the iterators to size_t's are minimal. I think it's a better idea just to follow the project and when they release the new version, we update.

SirCmpwn:
Yeah, but distros are going to have to deal with the fallout when we ship it and I imagine many of their users will come to us, not their distros, with the complaints.

mariusor:
OK. Then maybe we should create some wrapper functions? 

I doubt somebody else dictating what they release would go well with the people from json-c. :D

SirCmpwn:
Wrapper functions wouldn't really help.

We wouldn't be dictating what they release, we'd be letting them know as users of their library that the change would cause us headaches and politely asking that they roll it back before shipping it in a stable version. This is a perfectly reasonable thing for us to do.

herrold:
As predicted, as I read the thread, a topic of discussion at"

    https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/BGLLIFYNCJ6N5JNIWZNRH2OE6P23KAZX/

and their bug:

    https://bugzilla.redhat.com/show_bug.cgi?id=1525020

PR #1440, opened 2017-10-28
Convert FindLibcap.cmake to modern cmake

HenryMiller1:
I was looking for a FindLibcap.cmake to use in icecream, and yours was the only one I found.  I notice it doesn't support modern cmake so I added that for you.


HenryMiller1:
Note, as far as I can tell this is never actually used, (nothing seem to link to the library) even though you require it on linux.  Since I know almost nothing about your project I'm not willing to tell you to remove it, but you might want to consider if you still need this at all.

SirCmpwn:
I appreciate the thought, but we're likely leaving CMake soon and there's not a good reason to make this change anyway. We aren't using any newer CMake features and we'd just leave older users behind senselessly.

PR #1441, opened 2017-10-28
Removed trailing comma from swaybar click event json

ptMuta:
Trailing comma was violating JSON validity and caused issues with deserialization in different status_command implementations (eg. [i3status-rust](https://github.com/greshake/i3status-rust))


SirCmpwn:
Please associate your committer email with your GitHub account.

SirCmpwn:
Hmm, that's not quite right.

ptMuta:
There we go... First time I had to correct commiting with a wrong profile

SirCmpwn:
Thanks!

PR #1454, opened 2017-11-05
Fix init_tray function declaration

etam:
This fixes compilation failure:
error: call to function 'init_tray' without a real prototype


SirCmpwn:
Thanks!

PR #1456, opened 2017-11-06
Add wlc version

etam:


SirCmpwn:
We always depend on the latest version of wlc (master). We ship sway and wlc releases together. I'd rather not specify this here because it'll keep changing all the time.

PR #1472, opened 2017-11-22
Fix build with recent wlroots

martinetd:
A couple of trivial fix ; this should build with wlroots/master and remove the last need for wlc includes that I didn't have in a standard path.

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -8,8 +8,8 @@
 #include <stdlib.h>
 #include <string.h>
 #include <strings.h>
-#include <wlc/wlc.h>
 #include <xkbcommon/xkbcommon-names.h>
+#include <wlr/types/wlr_keyboard.h>
 #include "log.h"
 #include "readline.h"
 #include "util.h"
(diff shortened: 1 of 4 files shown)


martinetd:
I don't really get why the build fails, here's log excerpt:
```
$ git clone https://github.com/swaywm/wlroots
Cloning into 'wlroots'...
remote: Counting objects: 10272, done.
remote: Compressing objects: 100% (18/18), done.
remote: Total 10272 (delta 4), reused 8 (delta 2), pack-reused 10252
Receiving objects: 100% (10272/10272), 2.15 MiB | 11.97 MiB/s, done.
Resolving deltas: 100% (7558/7558), done.
$ cd wlroots
$ mkdir build
$ cd build
$ meson --prefix=/usr ..
Error during basic setup:
Neither directory contains a build file meson.build.
```

Doing the same thing here works, I can't see why it doesn't find meson.build. Fresh git clone should checkout the master branch which has a meson.build file.... pass, sorry.

SirCmpwn:
The travis build had been failing since before your patch. Don't worry about it.

Thanks!

PR #1477, opened 2017-11-28
set mode on output

acrisci:
Makes it work with drm. Future work will involve getting the configured mode and setting that.


SirCmpwn:
Thanks!

PR #1478, opened 2017-11-28
remove nvidia support claim

acrisci:


ascent12:
I think this should probably be moved over to wlroots itself.

SirCmpwn:
>I think this should probably be moved over to wlroots itself.

Nah, sway's the place for this imo.

PR #1479, opened 2017-11-29
meson build

acrisci:


Ongy:
On IRC we talked about the version that should go into this.

The ideas were date (form meson, or C macro) or git ref.

IMO git ref is strictly supperior, since 
a) a date is useful if one expects the other party to have an up to date source tree
b) it doesn't break reproducable builds. I'm not sure if they are currently possible, but it's something I think should be considered.

The downside: If a build is done without git, this needs some sane fallback. (Stable version for releases?)

acrisci:
>  The downside: If a build is done without git, this needs some sane fallback. (Stable version for releases?)

I added an option called `sway_version` that you can set to something and git will not be invoked to get the version in this case.

ascent12:
There is a 'version' argument to project (which I guess is going to be 1.0.0 for the wlroots branch), and you can access it later with meson.project_version().
I think it makes sense to just use that in the absence of git information.

acrisci:
That's a good approach, but version is not settable by the user in this way. Does anyone rely upon setting the project version?

acrisci:
@SirCmpwn do we want to do the meson build for the sway utilities (like the bar) now or are those getting totally redone?

SirCmpwn:
Waiting on the subprojects is fine.

acrisci:
Ok, we basically have the same functionality as before. Ready for review.

PR #1482, opened 2017-12-01
refactor: use frame done

acrisci:


sway/desktop/output.c
```diff
@@ -72,12 +72,7 @@ static void output_frame_view(swayc_t *view, void *data) {
 	struct timespec now;
 	clock_gettime(CLOCK_MONOTONIC, &now);
```
SirCmpwn: This can be removed now

SirCmpwn:
This code comes straight from rootston, we should yoink it there too

PR #1490, opened 2017-12-03
remove unused timepsec func

acrisci:
Should fix travis. I think we used this before to send surface callbacks.


SirCmpwn:
Thanks!

PR #1492, opened 2017-12-03
use app_id instead of class

acrisci:
The app_id and class are essentially the same thing and should be treated the same. Pick one name and stick with it.

### DIFF
--- a/include/sway/view.h
+++ b/include/sway/view.h
@@ -29,9 +29,8 @@ enum sway_view_type {
 
 enum sway_view_prop {
 	VIEW_PROP_TITLE,
-	VIEW_PROP_CLASS,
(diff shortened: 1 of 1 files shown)


SirCmpwn:
This was intentional. We should use whatever name is appropriate to each shell and try not to conflate different shell concepts.

acrisci:
It's the same concept, it just got renamed for some reason. Weston treats them as the same thing as well in its desktop abstractions.

Will criteria matching and the ipc strictly differentiate between these?

SirCmpwn:
Yep.

acrisci:
I can see that causing some problems for people, but I think I'm ok with it because it's more encouragement for people to not use the legacy shells.

PR #1497, opened 2017-12-04
Add exec and exec_always commands

emersion:
Remaining TODOs to add to wlroots issue:
- [ ] Add `exec` PIDs to active workspace

Test plan: `swaymsg exec gnome-dictionary`

### DIFF
--- a/sway/commands/exec.c
+++ b/sway/commands/exec.c
@@ -1,0 +1,16 @@
+#include <string.h>
+#include "sway/commands.h"
+#include "log.h"
+#include "stringop.h"
+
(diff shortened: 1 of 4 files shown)


sway/commands/exec.c
```diff
@@ -0,0 +1,16 @@
+#include <string.h>
+#include "sway/commands.h"
+#include "log.h"
+#include "stringop.h"
+
+struct cmd_results *cmd_exec(int argc, char **argv) {
+	// TODO: config
```
SirCmpwn: Speaking of which, we should get that subsystem spun back up sooner rather than later.

SirCmpwn:
Thanks!

PR #1499, opened 2017-12-05
meson wrap wlroots

acrisci:
Easy way to include wlroots dependency. Thanks @ascent12 .


SirCmpwn:
I don't like this approach.

PR #1503, opened 2017-12-06
Add output config

emersion:
- [x] Merge new output config params
- [x] Init new output config params
- [x] wlr_output_layout_add

Test plan:
- [x] Configure each property (resolution, refresh rate, position, scale, transform)
- [x] Test hotplugging
- [x] Test IPC


sway/config/output.c
```diff
@@ -0,0 +1,181 @@
...
+		if (mode->width == width && mode->height == height) {
+			if (mode->refresh == mhz) {
+				best = mode;
+				break;
+			}
+			best = mode;
+		}
+	}
+	if (!best) {
+		sway_log(L_ERROR, "Configured mode for %s not available", output->name);
```
acrisci: If no mode is ever set, the drm backend won't ever start up and (I think) the user loses the ability to switch VT so this is a pretty hard crash for a single monitor setup.

The obvious solution is to just set the default mode, but that's a silent error if you don't know to check the logs.

This is an excellent use case for system notifications but we don't have those yet.

sway/config/output.c
```diff
@@ -0,0 +1,181 @@
...
+	oc->width = oc->height -1;
+	oc->refresh_rate = -1;
+	oc->x = oc->y = -1;
+	oc->scale = -1;
+	oc->transform = -1;
+}
+
+void merge_output_config(struct output_config *dst, struct output_config *src) {
+	if (src->name) {
+		if (dst->name) {
```
acrisci: No need for this guard, just free.

sway/config/output.c
emersion: This is already done here: https://github.com/swaywm/sway/blob/wlroots/sway/desktop/output.c#L111

sway/tree/container.c
```diff
@@ -23,6 +25,30 @@ void swayc_descendants_of_type(swayc_t *root, enum swayc_types type,
 	}
 }
 
+static void update_root_geometry() {
```
acrisci: You can get output layout extents with `wlr_output_layout_get_box(layout, NULL)`.

sway/commands/output.c
```diff
@@ -0,0 +1,251 @@
...
+			"%d, %d scale %d transform %d refresh_rate %f) (bg %s %s)",
+			output->name, output->enabled, output->width,
+			output->height, output->x, output->y, output->scale,
+			output->transform, output->refresh_rate,
+			output->background, output->background_option);
+
+	if (output->name) {
+		// Try to find the output container and apply configuration now. If
+		// this is during startup then there will be no container and config
+		// will be applied during normal "new output" event from wlc.
```
acrisci: event from what?

sway/commands/output.c
```diff
@@ -0,0 +1,251 @@
...
+};
+
+struct cmd_results *cmd_output(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "output", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	const char *name = argv[0];
+
+	struct output_config *output = calloc(1, sizeof(struct output_config));
```
acrisci: Would be simpler to just have a create func here rather than exposing `output_config_defaults()`.

sway/tree/container.c
```diff
@@ -139,6 +167,34 @@ static void free_swayc(swayc_t *cont) {
...
 }
 
+swayc_t *destroy_output(swayc_t *output) {
+	if (!sway_assert(output, "null output passed to destroy_output")) {
+		return NULL;
+	}
+	if (output->children->length > 0) {
+		// TODO save workspaces when there are no outputs.
+		// TODO also check if there will ever be no outputs except for exiting
+		// program
```
acrisci: I think if there are no outputs, sway should exit.

sway/tree/container.c
```diff
@@ -139,6 +167,34 @@ static void free_swayc(swayc_t *cont) {
 	free(cont);
 }
 
+swayc_t *destroy_output(swayc_t *output) {
```
acrisci: Do we need to explicitly take it out of the output layout?

acrisci:
Code looks ok, will test it in a few hours. Can you provide an example config snippet?

sway/commands/output.c
SirCmpwn: Don't just copy+paste more than a few lines at a time, @emersion - make sure you read everything you import.

emersion:
Example config:

```
output WL-1 pos 10,10 transform 90 scale 2
```

Try changing these at runtime.

sway/config/output.c
```diff
@@ -14,13 +14,18 @@ int output_name_cmp(const void *item, const void *data) {
 	return strcmp(output->name, name);
 }
 
-void output_config_defaults(struct output_config *oc) {
+struct output_config *new_output_config() {
```
acrisci: We'll need to standardize our create function names.

include/sway/config.h
```diff
@@ -6,6 +6,7 @@
 #include <libinput.h>
 #include <stdint.h>
 #include <string.h>
+#include <wayland-server.h>
```
acrisci: Unused?

sway/commands/output.c
```diff
@@ -0,0 +1,250 @@
...
+struct cmd_results *cmd_output(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "output", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	const char *name = argv[0];
+
+	struct output_config *output = new_output_config();
+	if (!output) {
+		return cmd_results_new(CMD_FAILURE, "output", "Unable to allocate output config");
```
acrisci: If we're out of memory, this is probably going to fail too.

acrisci:
Instead of this `update_root_geometry()` madness, why not listen for the change event of the output layout and update the root geometry then?

sway/config/output.c
```diff
@@ -0,0 +1,181 @@
...
+			oc->y);
+	} else {
+		wlr_output_layout_add_auto(root_container.output_layout, wlr_output);
+	}
+	struct wlr_box *output_layout_box =
+		wlr_output_layout_get_box(root_container.output_layout, wlr_output);
+	output->x = output_layout_box->x;
+	output->y = output_layout_box->y;
+	output->width = output_layout_box->width;
+	output->height = output_layout_box->height;
```
acrisci: The output container could listen to the output layout change event and automatically keep these in sync.

sway/config/output.c
```diff
@@ -0,0 +1,181 @@
...
+		// Look for a * config for background
+		int i = list_seq_find(config->output_configs, output_name_cmp, "*");
+		if (i >= 0) {
+			oc = config->output_configs->items[i];
+		} else {
+			oc = NULL;
+		}
+	}
+
+	int output_i;
```
acrisci: Unused?

acrisci:
`output_add_notify()` needs to handle the case where the new output is disabled.

sway/config/output.c
emersion: Not used yet, but will be when we implement background (see commented code below).

emersion:
>output_add_notify() needs to handle the case where the new output is disabled.

This is already the case, but this is handled in `apply_output_config`.

sway/desktop/output.c
```diff
@@ -98,27 +98,53 @@ static void output_resolution_notify(struct wl_listener *listener, void *data) {
 	arrange_windows(soutput->swayc, -1, -1);
 }
 
+static void output_scale_notify(struct wl_listener *listener, void *data) {
+	struct sway_output *soutput = wl_container_of(
+			listener, soutput, scale);
+	arrange_windows(soutput->swayc, -1, -1);
+}
+
+static void output_transform_notify(struct wl_listener *listener, void *data) {
```
acrisci: Have you considered not listening to scale/transform/resolution but only output layout changes for determining when to rearrange the windows?

SirCmpwn:
Initial feedback

- Xwayland crashes on changes
- I would prefer to replace refresh_rate/resolution with a single mode command: `WIDTHxHEIGHT@RATE`, and have res just be an alias for `mode`

emersion:
>Xwayland crashes on changes

I don't believe this is due to invalid code. I added some code in rootston that changes the output scale on click and Xwayland doesn't crash. I think it's related to having no seat. Plus, the segfault happens in `CloseInput`.

sway/commands/output.c
```diff
@@ -0,0 +1,277 @@
...
+	const char *name = argv[0];
+
+	struct output_config *output = new_output_config();
+	if (!output) {
+		sway_log(L_ERROR, "Failed to allocate output config");
+		return NULL;
+	}
+	output->name = strdup(name);
+
+	// TODO: atoi doesn't handle invalid numbers
```
SirCmpwn: Can we resolve this now?

sway/desktop/output.c
```diff
@@ -92,40 +91,51 @@ static void output_frame_notify(struct wl_listener *listener, void *data) {
...
+	if (!output) {
+		return;
+	}
 	output->wlr_output = wlr_output;
 	output->server = server;
-	output->swayc = new_output(output);
 
-	if (wl_list_length(&wlr_output->modes) > 0) {
-		struct wlr_output_mode *mode = NULL;
-		mode = wl_container_of((&wlr_output->modes)->prev, mode, link);
```
SirCmpwn: :+1:

sway/tree/container.c
```diff
@@ -139,6 +161,36 @@ static void free_swayc(swayc_t *cont) {
...
+			int p = root_container.children->items[0] == output;
+			// Move workspace from this output to another output
+			while (output->children->length) {
+				swayc_t *child = output->children->items[0];
+				remove_child(child);
+				add_child(root_container.children->items[p], child);
+			}
+			sort_workspaces(root_container.children->items[p]);
+			// TODO WLR: is this needed anymore?
+			//update_visibility(root_container.children->items[p]);
```
SirCmpwn: It's not, pull it out

sway/tree/container.c
```diff
@@ -139,6 +161,36 @@ static void free_swayc(swayc_t *cont) {
 	free(cont);
 }
 
+swayc_t *destroy_output(swayc_t *output) {
+	if (!sway_assert(output, "null output passed to destroy_output")) {
+		return NULL;
+	}
+
+	if (output->children->length > 0) {
+		// TODO save workspaces when there are no outputs.
```
SirCmpwn: Note to self, put these in the wlroots ticket

SirCmpwn:
LGTM :+1:

SirCmpwn:
Thanks!

PR #1505, opened 2017-12-06
input management and seat

acrisci:
It's probably going to look a lot like rootston. An input manager that manages device add/remove and passes them to seats.

* [x] pointer device add/remove
* [x] pointer motion
* [x] pointer button/axis
* [x] pointer config
* [x] keyboard device add/remove
* [x] keyboard key
* [x] keyboard modifiers
* [x] keyboard config
* [x] sway seat
* [x] sway cursor
* [x] focus

Tests:

* `weston-clickdot`
* focus windows with left click
* type into focused windows (with modifiers)
* change keyboard layout at runtime: `swaymsg input 0:0:X11_keyboard xkb_layout de`.
* hotplug devices
* attach devices to seat at runtime: `swaymsg seat seat0 attach 0:0:X11_keyboard`

Future work:

* input for popups and subsurfaces
* detach devices
* keycodes to keysym conversion
* other device config (libinput, mappings)
* touch, tablet tool, tablet pad


SirCmpwn:
```
input device_id {
    # no change from existing input blocks
    layout us
    options caps:escape
}

cursor "cursor-0" {
    attach device_id...
    attach *
    map device_id (to)? output HDMI-A-1
    map device_id (to)? region 100x100@10,10
}

seat "seat-0" {
    attach device_id...
    attach cursor_id...
    attach *
    attach --layout=us --options=caps:escape keyboard_id
    attach --as-pointer touch_id
    attach --as-touch pointer_id...
}
```

Attaching a touch device as a pointer would present a wl_pointer to the client and simulate that. Attaching n pointers as a touch device simulates a wl_touch with n touchpoints. Attaching a keyboard with layout/options/etc overrides that keyboard's default layout/options for any interactions that go through this seat.

After config parsing, if no cursors were configured, add a default:

```
cursor "cursor-0" {
    attach *
}
```

If no seats were configured, add a default:

```
seat "seat-0" {
    attach *
}
```

SirCmpwn:
Oh, I should mention that `detach` should work at runtime for all of these.

SirCmpwn:
Also, another concern we should address is that previously sway input devices were identified with a consistent identifier based on libinput. We need to carry this identifier over, but there are additional constraints - we should be able to specify `sway_id/keyboard` for example to refer just to the keyboard capability of a multi-capable device.

SirCmpwn:
```
bindsym --to=keyboard_id|seat_id ...
```

acrisci:
I don't think it makes sense to have cursor config and seat config in separate blocks since they are conceptually bound to each other. There is a 1:1 relationship between seat and cursor (necessary because there can only be a single pointer location within a surface per seat).

It won't work well if seats share cursors because then they would be sharing the devices as well. If they share devices, surfaces then get double input events. For instance, two button clicks on the surface when the user clicks once, two scrolls, two motion events, two dnd targets which is going to lead to a lot of nasty bugs that can't be fixed. Just picking a seat and going with that will break the clipboard and is no different from just putting the device on that seat but with less predictability. I think this is a protocol limitation.

Devices can live on multiple seats, but this is a really dumb thing to do and I doubt anyone would ever us this configuration except by mistake. In this case, the single device will control both cursors at the same time.

acrisci:
> Also, another concern we should address is that previously sway input devices were identified with a consistent identifier based on libinput.

And what if libinput is not in use for that backend? Should we just ignore all input configuration in that case?

SirCmpwn:
>It won't work well if seats share cursors because then they would be sharing the devices as well. If they share devices, surfaces then get double input events (one per seat). For instance, two button clicks on the surface when the user clicks once, two scrolls, two motion events, two dnd targets which is going to lead to a lot of nasty bugs that can't be fixed. Clipboard for cursor events is broken because only the first seat that wins the race to click the "copy" button gets the contents. Just picking a seat and going with that will break the clipboard for the same reason and is no different from just putting the device on that seat but with less predictability. I think this is a protocol limitation.

Cursors can be closely associated with seats I guess. However, I should note that I want you to be able to tie a keyboard into multiple seats, and we should work it out so that whichever seat got input last (from devices other than that keyboard) would propegate the keyboard events.

>And what if libinput is not in use for that backend? Should we just ignore all input configuration in that case?

We should make up some other predictable device naming schema for that device.

acrisci:
I'm messing with the rendering a little bit.

It looks like we were trying to cram the buffer into the space of the desired window geometry. Window geometry is always smaller than the size of the buffer so this leads to scaling issues which 1) makes the windows look ugly when scaled and 2) completely screws up surface coordinate calculations.

To avoid this, the approach is to always render the surface buffer as it is intended to be rendered. When the shell specifies window geometry, we place the top left corner of the window geometry at the top left corner of the sway container.

Shell configure requests for size work on window geometry size and *not* buffer size so this gives us a good result with xdg-shell. wl-shell does not seem to have a way to get window geometry coordinates so this shell will never be completely supported by sway and we need to do crappy workarounds. Hopefully window maximization will clear this up a bit because it won't render shadows so window geometry offsets should be close to zero (untested).

A consequence is we have to trust the shell surface to actually listen to our configure requests. If it doesn't, then I think we should call that a client bug and just clip the contents to the container extents (which should be done anyway to avoid shadows spilling over to adjacent containers).

acrisci:
Ok, we have basic input now so I'll stop here and do the rest in smaller PRs because this one is getting very big and the others might need closer supervision.

This is your wm so I'll do the config your way even though I don't like the approach because I just want to get sway out to the people. I'll do my config my way when it is time to make my own wm 😄 .

SirCmpwn:
>This is your wm so I'll do the config your way even though I don't like the approach

Can you enumerate your concerns, if only for posterity's sake?

sway/commands/input/xkb_layout.c
```diff
@@ -0,0 +1,24 @@
...
+	if ((error = checkarg(argc, "xkb_layout", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	if (!current_input_config) {
+		return cmd_results_new(CMD_FAILURE, "xkb_layout", "No input device defined.");
+	}
+	struct input_config *new_config =
+		new_input_config(current_input_config->identifier);
+
+	new_config->xkb_layout = strdup(argv[0]);
```
SirCmpwn: If you only use the first arg, should the checkarg call expect exactly 1 instead of at least 1? Does it make sense to join_args on argv and use all of them?

SirCmpwn: Same comment applies to a few files.

include/sway/commands.h
```diff
@@ -39,6 +42,9 @@ enum expected_args {
 	EXPECTED_EQUAL_TO
 };
 
+void input_cmd_apply(struct input_config *input);
+void seat_cmd_apply(struct seat_config *seat);
```
emersion: Can this be moved in `config.h`, just like `apply_output_config`? (Maybe rename the latter)

sway/commands/input/xkb_layout.c
acrisci: Yeah I think you're right.

sway/config.c
```diff
@@ -226,6 +229,227 @@ static int qstrcmp(const void* a, const void* b) {
 	return strcmp(*((char**) a), *((char**) b));
 }
 
+struct input_config *new_input_config(const char* identifier) {
```
emersion: Just like `config/output.c`, can you move these in `config/input.c`?

include/sway/commands.h
acrisci: Yeah we need to do a big name cleanup because there are like 3 different conventions going on in the project right now.

sway/commands/input/xkb_layout.c
acrisci: None of these seem to accept spaces. When you specify multiple options/layouts, it is a comma separated list and adding spaces does not work. I will use checkarg to specify exactly 1 argument.

sway/desktop/output.c
```diff
@@ -38,19 +40,33 @@ static void output_frame_view(swayc_t *view, void *data) {
...
 	//}
 
+	// if the shell specifies window geometry, make the top left corner of the
+	// window in the top left corner of the container to avoid arbitrarily
+	// sized gaps based on the attached buffer size
+	int window_offset_x = 0;
+	int window_offset_y = 0;
+
+	if (view->sway_view->type == SWAY_XDG_SHELL_V6_VIEW) {
+		window_offset_x = view->sway_view->wlr_xdg_surface_v6->geometry->x;
```
emersion: Multiply this by the output scale to get output-local coordinates (`renmder{width,height}` are in output-local coordinates).

acrisci:
Something that's kind of awkward with how this works is there is no "default seat".

So I can't say "put these specific devices on seat1 and all the other devices on seat0".

So in this setup:

```
seat "seat0" {
    attach *
}

seat "seat1" {
    attach keyboard1
}
```

We get keyboard1 on both seats. If I don't use the `attach *` on seat0, then I have to specifically enumerate all the devices I want on that seat which is annoying because I want to be able to plug in anything and have it just work.

Should we put all devices that do not match an attachment on seat0? This is conceptually easier because it is unifying with the case where no seats are configured.

SirCmpwn:
Maybe `attach --exclusive *` would only attach devices which are not attached on any other seat? Would have to be delayed to the end of config file parsing.

sway/desktop/output.c
acrisci: I'm going to defer fixing this until my next effort, which will be to render popups and subsurfaces and have them work with inputs.

acrisci:
In English, that reads like "attach this device to this seat and not any other seat" which isn't logically what's happening. It seems a little cryptic for what will be necessary for the most common use case of multiseat.

SirCmpwn:
```
seat {
    fallback true
}
```

acrisci:
That's a little better. What happens if there is more than one fallback?

SirCmpwn:
Set the fallback to the last one to be set to it, and migrate any relevant input devices to that seat?

acrisci:
Multiple fallbacks seems like a mistake for the user, so we might as well punish them in a way that makes the error clear from looking at the config and put all the devices not specifically configured on all the fallback seats instead of silently interpreting their intent based on the order fallbacks appear in the config.

SirCmpwn:
That works for me too.

SirCmpwn:
Something I'd like to see done shortly after this PR is some commands that rely on focus, such as move and split.

SirCmpwn:
I'd like to see `swaymsg -t get_inputs` supported in this PR.

SirCmpwn:
Hmm, I am not well equipped to test some of these features because I only have one kind of mouse. I tested multi-seat stuff in wlroots with a tablet tool and touchscreen.

SirCmpwn:
:+1: from testing

sway/input/input-manager.c
```diff
@@ -0,0 +1,315 @@
...
+		if (*p == ' ') {
+			*p = '_';
+		}
+	}
+
+	int len =
+		(strlen(name) +
+		strlen_num(device->vendor) +
+		strlen_num(device->product) +
+		3) * sizeof(char);
```
SirCmpwn: Can you just use `int len = snprintf(NULL, 0, ...)` here

sway/commands/input/xkb_variant.c
```diff
@@ -0,0 +1,25 @@
...
+	}
+	struct input_config *new_config =
+		new_input_config(current_input_config->identifier);
+
+	new_config->xkb_variant = strdup(argv[0]);
+
+	sway_log(L_DEBUG, "apply-xkb_variant for device: %s",
+		current_input_config->identifier);
+	apply_input_config(new_config);
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
```
SirCmpwn: I tried `xkb_variant dvorak` and it had no effect.

sway/input/input-manager.c
```diff
@@ -0,0 +1,315 @@
...
+		seat = input_manager_get_seat(input, default_seat);
+		sway_seat_add_device(seat, input_device);
+		return;
+	}
+
+	bool added = false;
+	wl_list_for_each(seat, &input->seats, link) {
+		if (seat->config &&
+				(seat_config_get_attachment(seat->config,
+											input_device->identifier) ||
```
SirCmpwn: Woah

sway/input/keyboard.c
```diff
@@ -0,0 +1,112 @@
...
+	if (!sway_assert(context, "cannot create XKB context")) {
+		return;
+	}
+
+	xkb_keymap_unref(keyboard->keymap);
+	keyboard->keymap =
+		xkb_keymap_new_from_names(context, &rules, XKB_KEYMAP_COMPILE_NO_FLAGS);
+	wlr_keyboard_set_keymap(wlr_device->keyboard, keyboard->keymap);
+	wlr_keyboard_set_repeat_info(wlr_device->keyboard, 25, 600);
+	xkb_context_unref(context);
```
SirCmpwn: Changing XKB settings at runtime does not seem to have any effect.

sway/tree/container.c
```diff
@@ -218,3 +225,63 @@ swayc_t *swayc_parent_by_type(swayc_t *container, enum swayc_types type) {
...
+						&sview->surface->current->input,
+						view_sx, view_sy, NULL)) {
+				*sx = view_sx;
+				*sy = view_sy;
+				*surface = swayc->sway_view->surface;
+				list_free(queue);
+				return swayc;
+			}
+		} else {
+			list_cat(queue, swayc->children);
```
SirCmpwn: -1 to this approach. Use `swayc_descendants_of_type` instead.

SirCmpwn:
Thanks for your patience in implementing this and waiting on review. It's looking great, excellent work.

SirCmpwn:
VT switching

sway/tree/container.c
acrisci: I did this in my head earlier and determined that the correct way to do this will actually be reverse BFS to get the ordering correct. We don't really have a tree to test it with though.

acrisci:
> VT switching

This requires keycodes to keysyms which belongs in the effort for bindings.

sway/input/input-manager.c
acrisci: This is how vim indented it with `==`.

sway/input/keyboard.c
acrisci: I only tested changing the layout, which works for me.

sway/tree/container.c
acrisci: This approach is definitely going to change once we have popups/subsurfaces within a tree, and I'll probably need to do something like this to `swayc_descendents_of_type` to get the ordering right. The abstractions to tree searches might change since rendering and input handling need to happen in reverse order. There are a lot of subtleties here, so I think we should leave this in and wait to do all the search abstractions in the effort to render a tree with arbitrary depth later.

acrisci:
Ok, changing the layout at runtime is a wlroots bug. The issue is the layout won't change while keys on the keyboard are depressed. You can make it work by trying to change the layout from outside the backend like on another VT.

acrisci:
> Something I'd like to see done shortly after this PR is some commands that rely on focus, such as move and split.

Yeah, the tree stuff will all happen next. This was only meant to get us to the point where input events could be tested.

acrisci:
This is a tricky issue. Like I said, dynamic layout setting works in the X11 backend. The only difference I see there is that the X11 backend only has one keyboard while the drm backend has many (for instance, your power button is a keyboard).

I verified that the keymap is being sent by the seat in the expected way and the clients are receiving the event, and it seems to actually be for the new keymap, but the clients don't seem to be actually using the new keymap.

If I remove the device and readd it (like what happens when you switch VT), it will pick up on the new keymap.

sway/tree/container.c
SirCmpwn: Sounds good.

SirCmpwn:
>Update: it seems like adding a sleep 1 && <swaymsg command> doesn't work, so maybe that theory isn't correct.

I also tested setting the layout from a separate keyboard (on a different computer via SSH) than the one I was configuring the layout for to no avail.

acrisci:
The issue is that libinput is giving me two devices for the same keyboard and I was only configuring the first one which wasn't giving events because of an assumption that there would only be one device for the identifier.

SirCmpwn:
Can you take a look at this @emersion?

sway/commands/input.c
```diff
@@ -0,0 +1,65 @@
...
+#include "sway/input/input-manager.h"
+#include "log.h"
+
+struct cmd_results *cmd_input(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "input", EXPECTED_AT_LEAST, 2))) {
+		return error;
+	}
+
+	if (config->reading && strcmp("{", argv[1]) == 0) {
```
emersion: The `if (argc > 2)` check should be done before this line and return an error.

emersion: Oh, there's already `checkarg(argc, "input", EXPECTED_AT_LEAST, 2))`. So just remove the check, I guess?

sway/commands/seat.c
```diff
@@ -0,0 +1,37 @@
...
+		return error;
+	}
+
+	if (config->reading && strcmp("{", argv[1]) == 0) {
+		current_seat_config = new_seat_config(argv[0]);
+		sway_log(L_DEBUG, "entering seat block: %s", current_seat_config->name);
+		return cmd_results_new(CMD_BLOCK_SEAT, NULL, NULL);
+	}
+
+	if (argc > 2) {
```
emersion: Same here: useless because of `checkarg`.

sway/input/cursor.c
```diff
@@ -0,0 +1,184 @@
...
+#include "sway/view.h"
+#include "list.h"
+#include "log.h"
+
+static void cursor_update_position(struct sway_cursor *cursor) {
+	double x = cursor->cursor->x;
+	double y = cursor->cursor->y;
+
+	wlr_xcursor_manager_set_cursor_image(cursor->xcursor_manager,
+		"left_ptr", cursor->cursor);
```
emersion: There's a bug in wlroots about setting a hardware cursor's image twice (it fallbacks to software cursors, I'll fix that).

Setting the cursor image when it isn't needed is inefficient because it involves uploading the image to the GPU each time. This needs to be fixed here (maybe keep the name of the currently selected cursor in the struct?).

sway/input/input-manager.c
emersion: +1

sway/input/keyboard.c
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
+	if (!sway_assert(context, "cannot create XKB context")) {
+		return;
+	}
+
+	xkb_keymap_unref(keyboard->keymap);
+	keyboard->keymap =
+		xkb_keymap_new_from_names(context, &rules, XKB_KEYMAP_COMPILE_NO_FLAGS);
```
emersion: Need to check if `keyboard->keymap == NULL` here in case the keymap doesn't exist.

acrisci:
@martinetd yeah that's a wlroots issue related to a recent change. Make an issue there and we'll investigate it.

sway/commands/input.c
acrisci: I think the checkarg needs to expect 3 arguments for `seat <name> <command>` or `seat <name> {`, then the argc check will be useless. Maybe this was copied over from the back block, which doesn't require an identifier.

emersion: I don't think the command name is included in `argc`, is it?

martinetd: argc is 1 for a command with no argument with just argv[0] set, so yes it is. You'll have 1 + number of arguments.

emersion: I'm not sure about it. Right below, `argv[0]` is used as the device identifier and `argv[1]` is `{`.

acrisci: Nevermind, it gets the original args minus the first command. Going to fix it by doing another checkarg so a command like `seat <name> attach` is not valid.

sway/input/cursor.c
acrisci: I dunno I'm not going to do the cursor until last and all of this will probably change. I'm happy if it just shows up.

emersion: Okay just add a TODO then

acrisci:
Ok I think I fixed everything.

emersion:
What about that `checkarg` stuff in `seat` and `input` commands?

sway/commands/input.c
SirCmpwn: Yeah this is confusing. In sway's command handlers argv[0] is the first argument and argv[-1] is the command name.

emersion:
My bad, for some reason I was still looking at the previous diff.

sway/config/seat.c
```diff
@@ -0,0 +1,132 @@
...
+struct seat_config *new_seat_config(const char* name) {
+	struct seat_config *seat = calloc(1, sizeof(struct seat_config));
+	if (!seat) {
+		sway_log(L_DEBUG, "Unable to allocate seat config");
+		return NULL;
+	}
+
+	sway_log(L_DEBUG, "new_seat_config(%s)", name);
+	seat->name = strdup(name);
+	if (!sway_assert(seat->name, "could not allocate name for seat")) {
```
martinetd: should free seat here

sway/config/seat.c
```diff
@@ -0,0 +1,132 @@
...
+		return NULL;
+	}
+
+	seat->fallback = -1;
+	seat->attachments = create_list();
+	if (!sway_assert(seat->attachments,
+				"could not allocate seat attachments list")) {
```
martinetd: should free seat->name and seat here (free_seat_config isn't safe to use here because it doesn't check seat->attachments before using it)

sway/input/seat.c
```diff
@@ -0,0 +1,252 @@
...
+struct sway_seat *sway_seat_create(struct sway_input_manager *input,
+		const char *seat_name) {
+	struct sway_seat *seat = calloc(1, sizeof(struct sway_seat));
+	if (!seat) {
+		return NULL;
+	}
+
+	seat->wlr_seat = wlr_seat_create(input->server->wl_display, seat_name);
+	if (!sway_assert(seat->wlr_seat, "could not allocate seat")) {
+		return NULL;
```
martinetd: ditto should free seat

SirCmpwn:
:tada:

PR #1513, opened 2017-12-13
Swaylock: Clear password buffer after use.

ggreer:
After a user enters or clears a password, try to overwrite the buffer containing it. That way it's not sitting around in RAM for something else to read later. Also, mlock() the password buffer so that it is never swapped out.

This also replaces the dynamically allocated buffer with a char[1024]. Any characters past 1023 are discarded until the user hits enter or escape. (For comparison, i3lock uses a 512 byte buffer.)


swaylock/main.c
```diff
@@ -115,7 +115,7 @@ int function_conversation(int num_msg, const struct pam_message **msg,
 		switch (msg[i]->msg_style) {
 		case PAM_PROMPT_ECHO_OFF:
 		case PAM_PROMPT_ECHO_ON:
-			pam_reply[i].resp = password;
+			pam_reply[i].resp = strdup(password); // PAM clears and frees this
```
SirCmpwn: Source?

swaylock/main.c
```diff
@@ -127,9 +127,14 @@ int function_conversation(int num_msg, const struct pam_message **msg,
...
 }
 
-/**
- * Note: PAM will free() 'password' during the process
- */
+void clear_password_buffer() {
+	volatile char *pw = password;
+	for (size_t i = 0; i < sizeof(password); ++i) {
+		// show_indicator is set at runtime, so compiler can't optimize this out
+		pw[i] = show_indicator;
```
SirCmpwn: I would rather just do

```c
volatile char *pw = password;
volatile char zero = 0;
for (size_t i = ...) {
    pw[i] = zero;
}
```

PR #1515, opened 2017-12-13
Fix `swaylock -f`

4e554c4c:
This commit makes swaylock poll the registry the first time in its
fork in order for sway to recognize it and grant it lock privileges


SirCmpwn:
Thanks!

Issue #1507, opened 2017-12-09
Containers nested in tabbed/stacked containers do not function properly.


AGausmann:
## Steps to reproduce 
(I have done these steps multiple times with various applications)

1. Open a blank workspace, set it to tabbed or stacked mode `$mod+w`/`$mod+s` - this will be called "parent container".
2. Open two windows, they should open as tabs
3. Create a container ("child container") from one of the windows by selecting it and choosing vertical/horizontal tiling, tabs, or stack `$mod+v`/`$mod+b`/`$mod+w`/`$mod+s`
4. (OPTIONAL) Open another window in the child container.

## Problems

After performing the steps above, clicking on the child container's tab has no effect, and clicking in the title bar regions of the child container's windows (even when the child container is _not_ in focus) will switch focus to that window.

I have also confirmed that this occurs with any number of tabs in the parent container. There can be any number of "child containers", they all will have this issue. The example above demonstrates the issue with more than 1 window in the child container, but 1 window also works fine.

## Additional information
I am running Arch Linux `Linux exams 4.13.12-1-ARCH #1 SMP PREEMPT Wed Nov 8 11:54:06 CET 2017 x86_64 GNU/Linux`, using the [officially distributed package](https://www.archlinux.org/packages/community/x86_64/sway/) for sway 0.15.0-1.

<details/>
  <summary>.config/sway/config</summary>

```
# Default config for sway
#
# Copy this to ~/.config/sway/config and edit it to your liking.
#
# Read `man 5 sway` for a complete reference.

### Variables
#
# Logo key. Use Mod1 for Alt.
set $mod Mod4
# Home row direction keys, like vim
set $left h
set $down j
set $up k
set $right l
# Your preferred terminal emulator
set $term termite
# Your preferred application launcher
set $menu dmenu_run
# Lock dialog
set $lock swaylock -c 000000

### Output configuration
#
# Default wallpaper (more resolutions are available in /usr/share/backgrounds/sway/)
output * bg /usr/share/backgrounds/sway/Sway_Wallpaper_Blue_1920x1080.png fill
#
# Example configuration:
#
#   output HDMI-A-1 resolution 1920x1080 position 1920,0
#
# You can get the names of your outputs by running: swaymsg -t get_outputs

### Input configuration
#
# Example configuration:
#
#   input "2:14:SynPS/2_Synaptics_TouchPad" {
#       dwt enabled
#       tap enabled
#       natural_scroll enabled
#       middle_emulation enabled
#   }
[...]

PR #1516, opened 2017-12-14
Patch #1507: Clicking on invisible title bars switches focus

AGausmann:
The initial problem that sparked #1507 was when I had views inside a container inside a tabbed container and could focus those views by clicking in their title bar region when the tab wasn't focused. This patch resolves that bug, but the issue remains open because there is a lot more to fix in those scenarios.

### DIFF
--- a/sway/handlers.c
+++ b/sway/handlers.c
@@ -931,7 +931,7 @@ static bool swayc_border_check(swayc_t *c, const void *_origin) {
 	const struct wlc_point *origin = _origin;
 	const struct wlc_geometry title_bar = c->title_bar_geometry;
 
-	if (c->border_type != B_NORMAL) {
+	if (c->border_type != B_NORMAL || !c->visible) {
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Can you retarget this against the 0.15 branch instead of master? Also, I'm pretty busy right now, it would save me review time if you could write up a test plan for verifying this change.

AGausmann:
Rebasing, give me a second.

AravK:
Bump! Please integrate this, as till then rock-solid tab splitting functionality is not quite rock-solid.
Tab splitting is awesome but without this patch we cannot really do too many kind-of-largu things at a time.

SirCmpwn:
Have you tested it, @AravK?

AravK:
My original Firefox issue isn't occuring any more, so yay! Nothing's broken, as far as I can tell.
Keeping multiple splits of terminals open behind a Firefox session and clicking on tabs isn't clicking through. This patch works!

SirCmpwn:
Thanks!

PR #1439, opened 2017-10-27
Warning fixes for json-c json_object_array_length

mariusor:
A better patch for quieting the warnings related to json_object_array_length
This grants some future-proofing for the next json-c version which contains this API breaking change.

PR #1517, opened 2017-12-14
Fix issues with json-c v0.13 (Fixes #1355)

### LINKED ISSUE #1355: Problems with git master versions of json-c

besser82:
This superseeds #1439 and fixes #1355.

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -15,7 +15,6 @@
 #include <stdlib.h>
 #include <sys/ioctl.h>
 #include <fcntl.h>
-#include <json-c/json.h>
 #include <list.h>
 #include <libinput.h>
 #ifdef __linux__
@@ -25,6 +24,7 @@ struct ucred {
 	gid_t gid;
 };
 #endif
+#include "sway_json_helper.h"
 #include "sway/ipc-json.h"
 #include "sway/ipc-server.h"
 #include "sway/security.h"
@@ -724,7 +724,7 @@ void ipc_client_handle_command(struct ipc_client *client) {
 		}
 
 		// parse requested event types
-		for (int i = 0; i < json_object_array_length(request); i++) {
+		for (json_ar_len_t i = 0; i < json_object_array_length(request); i++) {
 			const char *event_type = json_object_get_string(json_object_array_get_idx(request, i));
 			if (strcmp(event_type, "workspace") == 0) {
 				client->subscribed_events |= event_mask(IPC_EVENT_WORKSPACE);
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -1126,7 +1126,8 @@ static void ipc_event_binding(json_object *sb_obj) {
 	sway_log(L_DEBUG, "Sending binding::run event");
 	json_object *obj = json_object_new_object();
 	json_object_object_add(obj, "change", json_object_new_string("run"));
-	json_object_object_add(obj, "binding", sb_obj);
+	// sb_obj gets owned by the temporary json_object, too.
+	json_object_object_add(obj, "binding", json_object_get(sb_obj));
 
 	const char *json_string = json_object_to_json_string(obj);
 	ipc_send_event(json_string, IPC_EVENT_BINDING);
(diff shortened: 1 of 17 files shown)


besser82:
This superseeds https://github.com/swaywm/sway/pull/1439.

sway/ipc-server.c
```diff
@@ -1126,7 +1126,12 @@ static void ipc_event_binding(json_object *sb_obj) {
 	sway_log(L_DEBUG, "Sending binding::run event");
 	json_object *obj = json_object_new_object();
 	json_object_object_add(obj, "change", json_object_new_string("run"));
-	json_object_object_add(obj, "binding", sb_obj);
+	// When adding a referenced json_object with an unknown lifetime
+	// to another json_object, it must be done with a wrapped call to
+	// json_object_get().  Otherwise the object's refcount does not
+	// get properly incremented and it will be deleted together with
+	// the object it got added to.
```
SirCmpwn: This comment belongs in the commit message

SirCmpwn:
I hate that we have to do this terrible workaround, but I don't see a better way. This sucks.

besser82:
@SirCmpwn This is actually *NOT* a workaround, but well documented behaviour of json_object_object_add() in the json-c API…

See:  https://github.com/json-c/json-c/blob/master/json_object.h#L379

```
Remember, when using
json_object_object_add or json_object_array_put_idx, ownership will
transfer to the object/array.  Call json_object_get if you want to maintain
shared ownership or also add this object as a child of multiple objects or
arrays.  Any ownerships you acquired but did not transfer must be released
through json_object_put.
```

and:  https://github.com/json-c/json-c/blob/master/json_object.h#L408

```
The reference count will *not* be incremented. This is to make adding
fields to objects in code more compact. If you want to retain a reference
to an added object, independent of the lifetime of obj, you must wrap the
passed object with json_object_get.

Upon calling this, the ownership of val transfers to obj.  Thus you must
make sure that you do in fact have ownership over this object.
```

SirCmpwn:
I'm referring to the typedef and the version check header, not the refcounting.

besser82:
Ahh, I see…  Well, the typedef is not needed, if we remove `-Werror`…  Converting from size_t to int is safe in all cases, since there are no json_array_objects bigger than 4 GiB used somewhere…

SirCmpwn:
Nah, I prefer this to dropping Werror, since Werror is useful beyond this one json-c problem.

SirCmpwn:
This has been merged into the 0.15 branch and will make the next bugfix release. Thanks!

besser82:
@SirCmpwn What about the master-branch?

SirCmpwn:
Don't worry about it.

PR #1519, opened 2017-12-15
Swaylock: Clear password buffer after use.

ggreer:
After a user enters or clears a password, try to overwrite the buffer containing it. That way it's not sitting around in RAM for something else to read later. Also, mlock() the password buffer so that it is never swapped out.

This also replaces the dynamically allocated buffer with a static char[1024]. Any characters past that are discarded until the user hits enter or escape.

I had to close #1513 because I opened it against the wrong base. This PR addresses your comments over there.

Regarding PAM zeroing-out the `strdup()`'d password: I figured that out with gdb. Here's a screenshot of my debug session:

screenshot from 2017-12-14 20-40-10

I'm too lazy to set up debug symbols, and the Linux PAM code is really hard to follow, so I can't link to the exact line where the zeroing happens.

I also used the debugger to verify that swaylock's copy of the password is zeroed-out.

### DIFF
--- a/swaylock/main.c
+++ b/swaylock/main.c
@@ -7,6 +7,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/mman.h>
 #include <sys/types.h>
 #include <pwd.h>
 #include <getopt.h>
@@ -56,8 +57,7 @@ void sway_terminate(int exit_code) {
 	exit(exit_code);
 }
 
-char *password;
-int password_size;
+char password[1024];
 enum line_source line_source = LINE_SOURCE_DEFAULT;
 
 struct lock_config *init_config() {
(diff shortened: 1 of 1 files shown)


swaylock/main.c
```diff
@@ -127,9 +127,14 @@ int function_conversation(int num_msg, const struct pam_message **msg,
...
 }
 
-/**
- * Note: PAM will free() 'password' during the process
- */
+void clear_password_buffer() {
+	volatile char *pw = password;
+	for (size_t i = 0; i < sizeof(password); ++i) {
+		// show_indicator is set at runtime, so compiler can't optimize this out
+		pw[i] = show_indicator;
```
SirCmpwn: I would rather just do

```c
volatile char *pw = password;
volatile char zero = 0;
for (size_t i = ...) {
    pw[i] = zero;
}
```

ggreer: Oops. That was the code I meant to have in this PR, but I'd forgotten to add `-f` to my `git push`.

SirCmpwn:
Thanks!

qaisjp:
> I had to close #1513 because I opened it against the wrong base. This PR addresses your comments over there.

@ggreer: Got here from your blog, you can use the edit button to change the base :)

PR #1531, opened 2017-12-22
[wip] multibackend

### LINKED ISSUE #1529: Multibackend
This issue is adapted from #1390 

### Example use-cases

- Support tests #1528 
- Add an RDP backend at runtime so your buddy can pair program with you
- Add a headless backend for scripting with
- Add a nested Wayland backend to your DRM session cause you're weird
- Add an x11 backend to your DRM session so you can switch TTYs to your x session and still manipulate your sway session

Backend configuration spec:

`backend add <type> [--name <name>] [args...]`: spins up a backend of the given type with the given name (or a generated name: `type-number` e.g. `libinput-0`). `[args...]` is backend-specific, such as a path to a DRM device for the DRM backend, a DISPLAY for x11, an initial number of outputs for the wayland backend, etc.

`backend del <name>`: does what it says on the tin

`backend <name> <subcommand> [args...]`: runtime manipulation of backend-specific features

### Subcommands

**wayland**

`backend wayland-0 add-output`

`backend wayland-0 del-output <id>`

**x11**

See wayland

**headless**

See wayland, plus

`backend headless-0 add-input <caps...>`

`backend headless-0 del-input <id>`

`backend headless-0 input <id> <send-keys|depress|release|click|move> [...]`

acrisci:
fixes #1529

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -1,9 +1,11 @@
 #define _POSIX_C_SOURCE 200112L
+#include <strings.h>
 #include <stdlib.h>
 #include <stdbool.h>
 #include <wayland-server.h>
 #include <wlr/backend.h>
 #include <wlr/backend/session.h>
+#include <wlr/backend/multi.h>
 #include <wlr/render.h>
 #include <wlr/render/gles2.h>
 #include <wlr/types/wlr_compositor.h>
@@ -20,6 +22,7 @@ bool server_init(struct sway_server *server) {
 	server->wl_display = wl_display_create();
 	server->wl_event_loop = wl_display_get_event_loop(server->wl_display);
 	server->backend = wlr_backend_autocreate(server->wl_display);
+	wl_list_init(&server->subbackends);
 
 	server->renderer = wlr_gles2_renderer_create(server->backend);
 	wl_display_init_shm(server->wl_display);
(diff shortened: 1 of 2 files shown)


acrisci:
This is stalled because wlroots does not support multiple renderers.

SirCmpwn:
I see a lot of this spammed in the console when I add a second headless output:

```
libEGL warning: FIXME: egl/x11 doesn't support front buffer rendering.
```

SirCmpwn:
Should not have any outputs added to the headless backend by default imo

Also, I attempted to add a wayland backend at runtime to no avail

SirCmpwn:
`swaymsg backend add wayland` gave an error about the # of args, `swaymsg backend add wayland wayland` segfaulted

SirCmpwn:
I wonder if we should start it up with an *empty* multi-backend and wait to add more backends to it (including headless) until we're asked for (or they're specified in the config file).

sway/commands/backend.c
```diff
@@ -0,0 +1,121 @@
...
+	}
+	sway_subbackend_remove_output(&server, backend, argv[0]);
+
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+}
+
+static struct cmd_results *backend_cmd_add_input(int argc, char **argv,
+		struct sway_subbackend *backend) {
+	return cmd_results_new(CMD_INVALID, "backend <cmd> [args]",
+			"add_output is not implemented");
```
SirCmpwn: error message here is wrong

sway/commands/backend.c
```diff
@@ -0,0 +1,121 @@
...
+}
+
+static struct cmd_results *backend_cmd_del_input(int argc, char **argv,
+		struct sway_subbackend *backend) {
+	return cmd_results_new(CMD_INVALID, "backend <cmd> [args]",
+			"del_output is not implemented");
```
SirCmpwn: here too

sway/main.c
```diff
@@ -272,6 +274,7 @@ int main(int argc, char **argv) {
 		"  -c, --config <config>  Specify a config file.\n"
 		"  -C, --validate         Check the validity of the config file, then exit.\n"
 		"  -d, --debug            Enables full logging, including debug information.\n"
+		"  -H, --headless         Starts sway with the headless backend.\n"
```
SirCmpwn: If we do the "no backends at all until configured explicitly" thing this should probably be `--no-auto-backend`

sway/server.c
```diff
@@ -32,16 +42,29 @@ static void server_ready(struct wl_listener *listener, void *data) {
...
+	wl_list_init(&server->subbackends);
+
+	if (headless) {
+		server->backend = wlr_multi_backend_create(server->wl_display);
+		struct sway_subbackend *subbackend =
+			sway_subbackend_create(SWAY_SUBBACKEND_HEADLESS, "headless");
+		sway_server_add_subbackend(server, subbackend);
+		sway_subbackend_add_output(server, subbackend, "headless");
+	} else {
+		// TODO add whatever this function creates to the subbackends
```
SirCmpwn: Will need a test here eventually to see if autocreate gave us a multibackend that we can just reuse or if we need to create a multibackend ourselves.

SirCmpwn: Maybe wlroots should provide some helper functions for merging multibackends together?

sway/server.c
```diff
@@ -101,3 +124,289 @@ void server_run(struct sway_server *server) {
 	setenv("WAYLAND_DISPLAY", server->socket, true);
 	wl_display_run(server->wl_display);
 }
+
+static void sway_subbackend_destroy(struct sway_subbackend *subbackend) {
+	if (subbackend->backend) {
+		wl_list_remove(&subbackend->backend_destroy.link);
+	}
+	wl_list_remove(&subbackend->link);
+	// free(name)?
```
SirCmpwn: ?

sway/server.c
```diff
@@ -101,3 +124,289 @@ void server_run(struct sway_server *server) {
...
+	free(subbackend);
+}
+
+struct sway_subbackend *sway_server_get_subbackend(struct sway_server *server,
+		char *name) {
```
SirCmpwn: This can be const

sway/server.c
```diff
@@ -101,3 +124,289 @@ void server_run(struct sway_server *server) {
...
+		if (strcasecmp(subbackend->name, name) == 0) {
+			return subbackend;
+		}
+	}
+
+	return NULL;
+}
+
+struct sway_subbackend *sway_subbackend_create(enum sway_subbackend_type type,
+		char *name) {
```
SirCmpwn: Should be const and strdup'd probably

sway/server.c
```diff
@@ -101,3 +124,289 @@ void server_run(struct sway_server *server) {
...
+		backend = drm_backend_create(server);
+		break;
+	case SWAY_SUBBACKEND_HEADLESS:
+		backend = headless_backend_create(server);
+		break;
+	}
+
+	if (backend == NULL) {
+		wlr_log(L_ERROR, "could not create subbackend '%s'", subbackend->name);
+		sway_subbackend_destroy(subbackend);
```
SirCmpwn: I don't like "subbackend" everywhere, just call it backend (or backend_config maybe for this type)

sway/server.c
```diff
@@ -101,3 +124,289 @@ void server_run(struct sway_server *server) {
...
+		break;
+	case SWAY_SUBBACKEND_X11:
+		wlr_log(L_DEBUG, "TODO: create x11 subbackend output");
+		break;
+	case SWAY_SUBBACKEND_DRM:
+		wlr_log(L_DEBUG, "creating DRM subbackend outputs is not supported");
+		break;
+	case SWAY_SUBBACKEND_HEADLESS:
+		wlr_output =
+			wlr_headless_add_output(subbackend->backend, 1280, 960);
```
SirCmpwn: I know I'm picking nits here but we should probably default to 1280x720. Also, we should have `add-output headless-2 [WIDTHxHEIGHT|WIDTH HEIGHT]`

SirCmpwn: I wonder if we can set the mode on headless outputs to an arbitrary resolution via `output HEADLESS-2 ...`, hmm...

acrisci:
Let me just set expecations by saying that adding and removing backends at runtime probably isn't going to work yet because of limitations in wlroots. That's going to be a pretty big effort to support. 

This PR is primarily for tests. Tests just need to start with a headless backend with an output.

The rest of the interface is stubbed out in the way we determined it should be in the issue. This will help us to develop the wlroots features because now we have a way to test the multibackend support and see how the wlr interface to backends works.

acrisci:
> Also, we should have add-output headless-2 [WIDTHxHEIGHT|WIDTH HEIGHT]

I think I'd rather have the output added in a disabled state, then use the normal output commands to do whatever you want to it, and finally enable it. That way we don't need to have every output param as part of this command.

acrisci:
> I wonder if we should start it up with an empty multi-backend and wait to add more backends to it (including headless) until we're asked for (or they're specified in the config file).

How do you want to handle the case of no backends? A backend renderer is needed to create a surface texture. What do we do with the surfaces that were created when no backend was present, and then the new backend appears? The surface interface doesn't fit this use case very well at all.

Another strange case is a backend with no outputs. In this case, the only container in the tree is the root container. What do we do with shell surfaces in this case? There's nowhere to put them in the tree.

SirCmpwn:
>How do you want to handle the case of no backends?

This is a bit crazy, but what if we waited until at least one backend was created to initialize the wl_compositor et al?

emersion:
What about merging support for adding outputs and input devices now, and add multibackend support later?

SirCmpwn:
Until we can do this right I'd rather just sit on it

PR #1535, opened 2017-12-24
Libinput config

martinetd:
Restore the code that used to be in config.c:apply_input_config() pre-wlroots.

Lines were very long in the original code, do we want to use the heavy rewrite as a chance to wrap these?

### DIFF
--- a/sway/input/input-manager.c
+++ b/sway/input/input-manager.c
@@ -6,6 +6,7 @@
 #include <string.h>
 #include <libinput.h>
 #include <math.h>
+#include <wlr/backend/libinput.h>
 #include "sway/config.h"
 #include "sway/input/input-manager.h"
 #include "sway/input/seat.h"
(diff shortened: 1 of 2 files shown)


martinetd:
FWIW. I only tested 'tap_enabled' and selected some text in a term that I spawned from a ssh session, I might be missing something here but at least the config block is properly parsed from previous PR merged earlier this week, the pointer finds it and it does apply something.

As a further thought, one thing that might or might not make sense is allow '*' as device name like we do with outputs? I kind of just assumed it would work, then saw in code that it doesn't.

acrisci:
> Lines were very long in the original code, do we want to use the heavy rewrite as a chance to wrap these?

Yeah please consider this a rewrite and make your best judgement on the new code.

> As a further thought, one thing that might or might not make sense is allow '*' as device name like we do with outputs? I kind of just assumed it would work, then saw in code that it doesn't.

If it's a sway feature currently and it doesn't work now, then it should be done eventually. Otherwise make an issue for new features.

acrisci:
For the approach, libinput config is a special case because it is not per-seat configuration but truly per-device configuration. If the device is on  more than one seat, this configuration will happen multiple times, so it seems to me to be an input-manager responsibility.

The input-manager is intended to handle configuration and input state that is independent of the seat, so I think it makes more sense conceptually to bring this up a level to `sway_input_manager_apply_input_config()`.

martinetd:
Careful that sway_input_manager_apply_input_config is only called when we modify the config block, not when a new device is added (we listen with input_add_notify, which currently calls sway_seat_add_device).
I'm not comfortable duplicating that code.
I could add another function that is called in both paths, though.

On the rest:
 - ok, will rewarp a bit.
 - I do not *think* that sway currently handles wildcards on input block. I honestly have no idea but that's my take on the old code in 2 minutes, could very well be wrong
 - please see @emersion comment in the parallel rootston PR about renaming `wlr_libinput_get_device_handle` to something else, I'd rather do that first if we do.

acrisci:
Yeah, it should be called when the device is added as well after the config is assigned.

SirCmpwn:
No wildcards on the input blocks, please. File that as a separate feature request if you'd like.

sway/input/seat.c
```diff
@@ -58,9 +61,64 @@ struct sway_seat *sway_seat_create(struct sway_input_manager *input,
...
+		}
+		if (ic->tap != INT_MIN) {
+			sway_log(L_DEBUG, "apply_input_config(%s) tap_set_enabled(%d)", ic->identifier, ic->tap);
+			libinput_device_config_tap_set_enabled(libinput_device, ic->tap);
+		}
+	}
+
 	wlr_cursor_attach_input_device(seat->cursor->cursor,
-		sway_device->input_device->wlr_device);
+		wlr_device);
```
SirCmpwn: For wrapping purposes put these on the same line now that it's shorter.

martinetd:
> Yeah, it should be called when the device is added as well after the config is assigned.
Not being very helpful as the current place the code is in does this (as well as is where you put a comment to do it :P)
Can a device actually be in multiple 'seats' like you suggest? I'm not sure I fully grasp what sense this would make; I guess you could have two seats with a single pointer shared and two separate keyboards?
I guess it can be, moved it into a separate function called in both paths as I offered to do.

> No wildcards on the input blocks, please. File that as a separate feature request if you'd like.
Ok, was more a surprised input behaves differently than output kind of thing, I'm not fussy. This is out of scope anyway.

> Indent/wrapping
Agreed long fonction names + long variables can be a pain with line length. I've taken a middle ground, wrapped the log calls that can be split semi-easily (still over 80 but now no longer than the function calls)

acrisci:
> Can a device actually be in multiple 'seats' like you suggest?

Yes. Each seat has independent focus and will send input events from the device to its focused resources. This means if that the device is on five seats, and they all have the same window focused, and you press the 'f' key, the window will recieve 'fffff'.

We don't really know what we're going to do with this feature yet.

sway/input/input-manager.c
```diff
@@ -82,6 +83,75 @@ static bool input_has_seat_configuration(struct sway_input_manager *input) {
 	return false;
 }
 
+static void sway_input_manager_libinput_config_pointer(struct sway_input_device *input_device) {
```
acrisci: Do we know this is only pointer configuration or does this apply to any other devices like tablet tool or touch devices?

martinetd: setting like accel don't make sense to me for non-pointers. I have no idea about tap and other settings though, but I check that the input device is of type WLR_*_POINTER before calling this function... 

According to https://wayland.freedesktop.org/libinput/doc/latest/group__config.html#details we seem to only set options that make sense for pointers and touchpads (which is pointer for us), but there are functions to probe if the device is capable so we could remove the wlr type check and use these instead.
I think it's a bit more ambitious than just reviving the sway-input feature though, probably should get its own ticket/PR if we want to be that modular?

acrisci: Touch devices and tablet tool devices will now be treated separately. Look at the rootston code for a preview of what touch and tablet tool support will look like.

If this configuration applies to tablet tool and touch, then it should be applied to those kinds of devices as well.

PR #1537, opened 2017-12-27
Update to json-c 0.13

jelly:
Update json-c dependency to 0.13, which changed the return type of
json_object_array_length to support only a size_t instead of an int as
fallback.

Not sure if you want to take this patch, otherwise I'll apply it manually in the Arch Linux package.


jelly:
Just noticed this is a duplicate of https://github.com/swaywm/sway/pull/1439/files

SirCmpwn:
Sway 0.15.1 is coming out today with this patch applied.

PR #1539, opened 2017-12-27
Bindings

acrisci:
Make bindings work.

TODO

* [x] parse binding command
* [x] overwrite old bindings
* [x] translate keysyms
* [x] compositor bindings
* [x] user keysym bindings
* [x] user keycode bindings
* [x] bindings with `--release`

**Tests**

Example config:

These are not valid commands, so you'll have to check the logs to see if they run.

```
# bindsym translated
bindsym Alt+at pressed shift-alt-2

# bindsym raw
bindsym Alt+Shift+1 pressed shift-alt-1

# bindsym release
bindsym --release a do-it-on-release

# make sure these get overwritten and only cmd3 runs when you press "s"
bindsym s cmd1
bindsym s cmd2
bindsym s cmd3

# Check bindcode with and without modifiers (use xev or similar to check keycodes)
bindcode 39 bindcode no modifier
bindcode Alt+39 bindcode one modifier
bindcode Alt+Shift+39 bindcode two modifiers
bindcode --release 40 bindcode-release
```

**Open questions**

Getting the keysym from the name during parsing the bindings is currently done in a case insensitive way. So `bindsym A <cmd>` is equivalent to `bindsym a <cmd>` in that the keysym that is returned from the parsing is `a` and neither will respond to the raw `Shift+a`.

What order should bindings run in if they are equivalent? Before, when there was only one keyboard layout during the lifetime of the wm, we could do this test when we parse the config, but now there is the possibility that many keyboards can have many layouts so the check is no longer possible until the key is pressed. Right now, we prioritize `bindcode` bindings (because that was the easiest to do), but we could also run them in the order they were applied in the config with just a little more complexity.

Should hardcoded compositor bindings (currently only VT switching, but there will be more) run before or after? Current thinking is no, but an interesting case that will come up is that i3 has a hardcoded binding to Esc which cancels an interactive resize. If this feature is implemented, then the consequence will be that you can't bind to Esc.

### DIFF
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -1,0 +1,173 @@
+#ifdef __linux__
+#include <linux/input-event-codes.h>
+#elif __FreeBSD__
+#include <dev/evdev/input-event-codes.h>
+#endif
+#include <xkbcommon/xkbcommon.h>
+#include <xkbcommon/xkbcommon-names.h>
+#include <strings.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "list.h"
+#include "log.h"
+#include "stringop.h"
+#include "util.h"
+
+int binding_order = 0;
+
+void free_sway_binding(struct sway_binding *binding) {
+	if (binding->keys) {
+		for (int i = 0; i < binding->keys->length; i++) {
+			free(binding->keys->items[i]);
+		}
+		list_free(binding->keys);
+	}
+	if (binding->command) {
+		free(binding->command);
+	}
+	free(binding);
+}
+
+struct cmd_results *cmd_bindsym(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "bindsym", EXPECTED_MORE_THAN, 1))) {
+		return error;
+	}
+
+	struct sway_binding *binding = malloc(sizeof(struct sway_binding));
+	if (!binding) {
+		return cmd_results_new(CMD_FAILURE, "bindsym",
+				"Unable to allocate binding");
+	}
+	binding->keys = create_list();
+	binding->modifiers = 0;
+	binding->release = false;
+	binding->bindcode = false;
+
+	// Handle --release
+	if (strcmp("--release", argv[0]) == 0) {
+		if (argc >= 3) {
+			binding->release = true;
+			argv++;
+			argc--;
+		} else {
+			free_sway_binding(binding);
+			return cmd_results_new(CMD_FAILURE, "bindsym",
+				"Invalid bindsym command "
+				"(expected more than 2 arguments, got %d)", argc);
+		}
+	}
+
+	binding->command = join_args(argv + 1, argc - 1);
+
+	list_t *split = split_string(argv[0], "+");
+	for (int i = 0; i < split->length; ++i) {
+		// Check for a modifier key
+		uint32_t mod;
+		if ((mod = get_modifier_mask_by_name(split->items[i])) > 0) {
+			binding->modifiers |= mod;
+			continue;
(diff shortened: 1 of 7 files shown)


martinetd:
Basic things seem to work.
Problems with drm backend
 - I got Mod1+shift+2 to work, but not Mod1+@. Looks like Setting a binding Mod1+@ also broke VT switching somehow?!
 - I couldn't get Mod1+button1 to work despite seeing code that looks like it should be handled

I read code quickly, looks okish, too tired for proper review & will re-read tomorrow.

acrisci:
> I got Mod1+shift+2 to work, but not Mod1+@

That binding won't validate because apparently xkb wants to call it "at". I can make `Alt+at` work.

martinetd:
> That binding won't validate because apparently xkb wants to call it "at". I can make Alt+at work.

Doh, ok. Could we catch that '@' is not valid xkb-speak somehow and log something? Should I look further into why that made ctrl+alt+Fx stop working altogether by setting an invalid keybind.

martinetd:
```
# bindsym release
bindsym --release a do-it-on-release
```
is not done on key release but on keypress here (still drm)

> Should I look further into why that made ctrl+alt+Fx stop working altogether by setting an invalid keybind.

Obviously can't reproduce anymore with last patches, so I guess this part is ok...

martinetd:
tested wayland backend in rootston, no problem. X11 backend in rootston got no keybindings to work, but X11 backend on wmii worked fine key-binding wise so I guess that's a rootston problem.

(Same --release bindsym being executed on keypress with both, though)

sway/commands/bind.c
```diff
@@ -0,0 +1,242 @@
...
+	return true;
+}
+
+struct cmd_results *cmd_bindsym(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "bindsym", EXPECTED_MORE_THAN, 1))) {
+		return error;
+	}
+
+	struct sway_binding *binding = malloc(sizeof(struct sway_binding));
```
martinetd: this needs to be a calloc, because we can pass this to free_sway_binding halfway on invalid length with --release argument later on and this will segfault on free(binding->command) that is invalid pointer in this case.

sway/commands/bind.c
```diff
@@ -0,0 +1,242 @@
...
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+}
+
+struct cmd_results *cmd_bindcode(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "bindcode", EXPECTED_MORE_THAN, 1))) {
+		return error;
+	}
+
+	struct sway_binding *binding = malloc(sizeof(struct sway_binding));
```
martinetd: ditto, calloc needed.

sway/input/keyboard.c
```diff
@@ -1,18 +1,343 @@
...
+		sway_log(L_DEBUG, "@@ checking binding: %s", binding->command);
+		if (modifiers ^ binding->modifiers || n != binding->keys->length) {
+			sway_log(L_DEBUG, "@@ modifiers or key num dont match");
+			continue;
+		}
+
+		bool match = true;
+		for (int j = 0; j < binding->keys->length; ++j) {
+			match =
+				pressed_keysyms_index(pressed_keysyms,
```
martinetd: (nitpick) weird indentation?

acrisci: This looks right to me.

acrisci:
Ok I think everything works now.

SirCmpwn:
Thanks!

PR #1542, opened 2017-12-29
Pretty-print all output info in swaymsg -t get_outputs

emersion:


martinetd:
Here's the new interactive output for me:
```
Output eDP-1 'Unknown 0x0641 0x00000000'
  Mode: 0x0 @ 59.999001 Hz
  Position: 0,0
  Scale factor: 1x
  Transform: normal
  Workspace: 1

```
I'd say it's hard to tell what the make/model/serials are when it's as badly provided as this laptop... Not sure what to do about it though, I guess it's ok, and the json format still works. LGTM.

emersion:
Yeah, for internal laptop screens it's generally not very helpful. External screens are usually better at this (mine displays as `Samsung SyncMaster HS3P505873`).

The mode seems to be invalid though...

martinetd:
> The mode seems to be invalid though...

Good point. It's not present in the json output at all for me, so explains why it's 0 here.

acrisci:
Yeah, kind of agree with @martinetd here. It looks great when the fields exist, but when they don't, it looks a little weird. Here is my X11 output edgecase:

```
Output X11-1 '  '
  Mode: 634x1039 @ 0.000000 Hz
  Position: 0,0
  Scale factor: 1x
  Transform: normal
  Workspace: 1
```

I would like to see refresh rounded to two decimal places.

emersion:
Yeah, I tried putting it that way because I wanted to support configuring outputs with their make/model/serial next, and using the same format for `swaymsg` and for the configuration file seemed like a good idea. Users would be able to copy-paste the string in their config file.

What we could do:
* Hide this information when make/model/serial is empty (and remove Wayland backend's make/model)
* Always populate this information in all backends
* Use a fallback value (`unknown`)
* Display partial information (only make/model when serial is empty), but then users won't always be able to copy-paste the output name directly in the config (but when serial is missing, what's the point of not using the connector name instead?)

What do you think? Any other ideas?

martinetd:
Why not populate this in X11/wayland/headless backend, this would let us test "config by make/model/serial" more easily as well.
I'd also be tempted to hide this if unset for ugly cases like my laptop but... It _is_ set with an arbitrary string, who are we to say 0x0000 is not a valid serial? So I'd just leave it as it is for these.

emersion:
>Why not populate this in X11/wayland/headless backend, this would let us test "config by make/model/serial" more easily as well.

Yes but configuring by make/model/serial in these situations doesn't make a lot of sense. Maybe if we allow to configure per make without model/serial it could be more useful (to say something like, "apply this configuration to all Wayland outputs").

>I'd also be tempted to hide this if unset for ugly cases like my laptop but... It is set with an arbitrary string, who are we to say 0x0000 is not a valid serial? So I'd just leave it as it is for these.

It's still useful to be able to configure an output with `Unknown 0x0641 0x00000000`, because it's unlikely that you have two screens with the same make/model, even if the serial is unset.

SirCmpwn:
Thanks!

PR #1543, opened 2017-12-29
Allow to configure outputs by their identifier

emersion:
Test plan:

```
output "wayland wayland " scale 2
```


acrisci:
Tested it and it works. Code looks ok too.

acrisci:
Should update the documentation to tell people about the feature. Also, identifier should be included in the raw ipc output and maybe made a bit more clear in the pretty output.

PR #1544, opened 2017-12-29
partial fix #1056

### LINKED ISSUE #1056: swaylock allow popup to come through
VirtualBox likes to remember the user when a new version is available. Their popup come through swaylock, and pressing enter while the screen is locked does dismiss the popup. Afterward, it is not possible to login anymore.

CedricCabessa:
this patch allow the user to type its password when a modal window is shown
while the desktop is locked by swaylock.
It is a partial fix of the bug since the window should not be visible (see bug
https://github.com/swaywm/sway/issues/1056 for details).


SirCmpwn:
Thanks!

PR #1550, opened 2018-01-04
json-c: add support for both 0.12.1 and 0.13

martinetd:
json-c 0.13 is still *very* new (less than a month old) and even "up to date" distros like the latest fedora don't have it yet.

Given the difference is minor I'd like to support both for a short while until the distros slightly behind arch follow up with it ; I don't mind keeping this in my local branch (I test wlroots/sway both on arch and non-arch so I need something like this), but I figured this might be useful for others.

I've tested this commit with both json-c 0.13 and 0.12.1

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -18,7 +18,7 @@ datadir = get_option('datadir')
 sysconfdir = get_option('sysconfdir')
 prefix = get_option('prefix')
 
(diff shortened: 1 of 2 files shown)


SirCmpwn:
json-c will be shipping in plenty of distros by the time we finish the wlroots branch.

PR #1551, opened 2018-01-04
sway main: init old log system until end of switch to wlr_log

martinetd:
We need to keep the old log init if we do not want all current calls to sway_log to be silent

The function declaration next to log headers is fugly but I don't have a better idea for something quick, I guess I can try to fix sway's log.h to be compatible if we care but it'd probaby be just as fast to finish converting all sway_logs to wlr_logs...

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -16,6 +16,7 @@
 #include <sys/prctl.h>
 #endif
 #include <wlr/util/log.h>
+void init_log(log_importance_t verbosity); // log.h, but can't include because of conflict
 #include "sway/config.h"
 #include "sway/server.h"
 #include "sway/layout.h"
@@ -331,10 +332,13 @@ int main(int argc, char **argv) {
 	// TODO: switch logging over to wlroots?
 	if (debug) {
 		wlr_log_init(L_DEBUG, NULL);
+		init_log(L_DEBUG);
 	} else if (verbose || validate) {
 		wlr_log_init(L_INFO, NULL);
+		init_log(L_INFO);
 	} else {
 		wlr_log_init(L_ERROR, NULL);
+		init_log(L_ERROR);
 	}
 
 	if (optind < argc) { // Behave as IPC client


4e554c4c:
Why do we need `sway_log` at all? when files are ported `sway_log` should be just changed to `wlr_log`, or if that's too difficult to do, `sway_log` could just be defined as a macro that gives its args to `wlr_log`.

SirCmpwn:
I agree, this should instead be ripping out the old logger entirely.

martinetd:
The problem is I wanted to test the previously done binding PR, that only had sway_log, which wasn't reworked to use wlr_log (and couldn't have because it wasn't landed at that point)
I think it doesn't hurt to have a bit of leftover and proper removal when we're done until then.

Once again I already have local patches because of json-c so I don't really care what's done with this. I feel this might help other developers. Just close this if you want.

SirCmpwn:
>I think it doesn't hurt to have a bit of leftover and proper removal when we're done until then.

I disagree

>Once again I already have local patches because of json-c so I don't really care what's done with this.

FTR a patch landed the other day to update the wlroots branch json-c version.

martinetd:
> FTR a patch landed the other day to update the wlroots branch json-c version.

And that's precisely my problem because I can no longer compile anything I'm pushing a this point.
I compile with an extra patch and spend ten seconds rebasing everytime I do new commits (because I still care about testing what I do a bit) until my distro bothers upgrading, which might be in weeks or months.
It might take a few minutes to remember to cleanup later but it'd save way more than that for me in the end... Anyway, that's not up to discussion so I'll stop ranting about this and get back to work.

4e554c4c:
@martinetd to test an old pr like that you could do `#define sway_log(VERBOSITY, FMT, ...) wlr_log(VERBOSITY, FMT, ##__VA_ARGS__)`

martinetd:
> @martinetd to test an old pr like that you could do #define sway_log(VERBOSITY, FMT, ...) wlr_log(VERBOSITY, FMT, ##__VA_ARGS__)

sure that's another possible fix, doing that in our log.h would have worked as well.

4e554c4c:
if it's wanted I'll send a PR :p

martinetd:
It doesn't look like it is, thanks though :)

SirCmpwn:
>I compile with an extra patch and spend ten seconds rebasing everytime I do new commits (because I still care about testing what I do a bit) until my distro bothers upgrading, which might be in weeks or months.

Why can't you just upgrade it yourself and send your distro a patch? Your distro isn't run by magicians who bestow packages upon the populace at a whim, it's run by volunteers in the open who accept patches.

martinetd:
@SirCmpwn: I'm using four different distros daily (arch, debian, centos and fedora), and I seriously don't give a damn about json-c.
If it's something I care about I will do it (and have done it on the three open distros), but this package has maintainers who will do it when they feel like it. For "slow" distros like debian even if I send a patch it won't make it into testing for a few weeks, possibly months, and I'm pretty sure centos doesn't accept patches in the first place.
I'll just keep maintaining the compat patch for my dev branch, and not bring it up again.

PR #1552, opened 2018-01-04
config cleanup & implement free_config

martinetd:
Some typos but it's mostly leak fixing, was waiting for bindings to land to push

### DIFF
--- a/sway/commands/input.c
+++ b/sway/commands/input.c
@@ -24,7 +24,11 @@ struct cmd_results *cmd_input(int argc, char **argv) {
 	char **argv_new = argv+2;
 
 	struct cmd_results *res;
+	struct input_config *old_input_config = current_input_config;
 	current_input_config = new_input_config(argv[0]);
+	if (!current_input_config) {
+		return cmd_results_new(CMD_FAILURE, NULL, "Couldn't allocate config");
+	}
 	if (strcasecmp("accel_profile", argv[1]) == 0) {
 		res = input_cmd_accel_profile(argc_new, argv_new);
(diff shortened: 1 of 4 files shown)


4e554c4c:
I know that some of these lists contain pointers that should be freed too. Should we just add that when implementing those parts of the config?

martinetd:
Yes, there is no implementation currently so the various free functions are in headers but do not exist yet.
I have an example with cmd_set I'll push after this; a problem of this approach is that it serializes things quite a bit... We can add other cleanups later if that's a problem.

sway/commands/input.c
```diff
@@ -24,7 +24,11 @@ struct cmd_results *cmd_input(int argc, char **argv) {
 	char **argv_new = argv+2;
 
 	struct cmd_results *res;
+	struct input_config *old_input_config = current_input_config;
 	current_input_config = new_input_config(argv[0]);
+	if (!current_input_config) {
+		return cmd_results_new(CMD_FAILURE, NULL, "Couldn't allocate config");
```
emersion: If we can't allocate an input config, we probably shouldn't allocate a command result.

martinetd: Yeah, I thought about that, but I have no idea how to return something else? We can't return something on the stack or a null..

SirCmpwn: This is an unsolved problem bigger than this PR.

SirCmpwn:
We should implement the reload command soon.

martinetd:
I've got set in the pipe, can do reload as well this weekend.

PR #1555, opened 2018-01-05
commands: add 'reload' command

martinetd:
This is actually pretty straightforward, and doesn't seem to leak anything.

I've tested:
 - key bindings: added/removed/changed, no problem seen
 - input config: changed some libinput config option without problem.

Not sure how the other options work, I'm not using them.

Note that when I open an urxvt before reloading, sometimes reloading will resize it, but I think this is working as intended as it's called arrange_windows(). I'm not sure why it doesn't happen everytime, but I'm not going to look at window placements right now as there is nothing done there (is there?)
I've also had the choice of commenting out load_swaybars() in commands or adding a stub in config.c and decided to add a stub, happy to go the other way, but we'll probably want that function eventually.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -21,6 +21,7 @@
 #endif
 #include <wlr/types/wlr_output.h>
 #include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
 #include "sway/commands.h"
 #include "sway/config.h"
 #include "sway/layout.h"
@@ -107,6 +108,23 @@ void free_config(struct sway_config *config) {
 	free(config->font);
 	free((char *)config->current_config);
 	free(config);
+}
+
+static void destroy_removed_seats(struct sway_config *old_config,
+		struct sway_config *new_config) {
+	struct seat_config *seat_config;
+	struct sway_seat *seat;
+	int i;
+	for (i = 0; i < old_config->seat_configs->length; i++) {
+		seat_config = old_config->seat_configs->items[i];
+		/* Also destroy seats that aren't present in new config */
+		if (new_config && list_seq_find(new_config->seat_configs,
+				seat_name_cmp, seat_config->name) < 0) {
+			seat = input_manager_get_seat(input_manager,
+				seat_config->name);
+			sway_seat_destroy(seat);
+		}
+	}
 }
 
 static void config_defaults(struct sway_config *config) {
@@ -382,6 +400,7 @@ bool load_main_config(const char *file, bool is_active) {
 	}
 
 	if (old_config) {
+		destroy_removed_seats(old_config, config);
 		free_config(old_config);
 	}
 	config->reading = false;
--- a/sway/config.c
+++ b/sway/config.c
@@ -698,3 +698,7 @@ char *do_var_replacement(char *str) {
 	}
 	return str;
 }
+
+void load_swaybars() {
+	/* stub function for reload commnd, to fill when we restore swaybars */
+}
--- a/sway/config.c
+++ b/sway/config.c
@@ -717,7 +717,3 @@ char *do_var_replacement(char *str) {
 	}
 	return str;
 }
-
-void load_swaybars() {
-	/* stub function for reload commnd, to fill when we restore swaybars */
-}
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -25,6 +25,7 @@ struct sway_cursor {
 	struct wl_listener request_set_cursor;
 };
 
+void sway_cursor_destroy(struct sway_cursor *cursor);
 struct sway_cursor *sway_cursor_create(struct sway_seat *seat);
 
 #endif
--- a/include/sway/input/input-manager.h
+++ b/include/sway/input/input-manager.h
@@ -46,4 +46,6 @@ void sway_input_manager_apply_seat_config(struct sway_input_manager *input,
 struct sway_seat *sway_input_manager_get_default_seat(
 		struct sway_input_manager *input);
 
+struct sway_seat *input_manager_get_seat(struct sway_input_manager *input,
+		const char *seat_name);
 #endif
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -29,6 +29,8 @@ struct sway_seat {
 struct sway_seat *sway_seat_create(struct sway_input_manager *input,
 		const char *seat_name);
 
+void sway_seat_destroy(struct sway_seat *seat);
+
 void sway_seat_add_device(struct sway_seat *seat,
 		struct sway_input_device *device);
 
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -136,6 +136,7 @@ static struct cmd_handler handlers[] = {
 	{ "input", cmd_input },
 	{ "kill", cmd_kill },
 	{ "output", cmd_output },
+	{ "reload", cmd_reload },
 	{ "seat", cmd_seat },
 	{ "set", cmd_set },
 };
--- a/sway/commands/reload.c
+++ b/sway/commands/reload.c
@@ -1,0 +1,21 @@
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/layout.h"
+
+struct cmd_results *cmd_reload(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if (config->reading) {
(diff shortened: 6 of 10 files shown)


SirCmpwn:
Can you also test changing output settings and committing them with reload?

martinetd:
I've tried "output * transform 90" and the display got rotated as expected; but removing the whole line doesn't transform it back to normal (changing the 90 to normal does work), I suppose some default values need to be restored into outputs somehow, not sure how tricky this will be to fix, will have a look.

martinetd:
I've also tested input has the same problem, changing the value does work, but removing the line (even part of a block) entierely will just not re-apply the setting so for example if my system's libinput defaults to tap_disabled it won't re-enforce the default.
To do that, we'd have to call all the get_default functions to compare with the actual value and re-enforce things appropriately, e.g. `libinput_device_config_tap_get_default_enabled ` vs. `libinput_device_config_tap_get_enabled` and then re-set with `libinput_device_config_tap_set_enabled `... And that's just tap, we have a dozen of these settings.

After some thought I'd say fiddling with removed lines isn't our problem anymore, even if that means that a given config will not produce the same environment depending on what it's been before reload - this is basically the same as users running xmodmap or xinput manually.
If someone changed tap setting with xinput, and we reload sway config, we don't want to erase their tap setting do we?
For output, when we do a *randr tool, I don't think reload should reset the output to normal either if there's no output transform option.

If that's correct I believe this PR works as expected and is ready.

acrisci:
Some things may hold onto elements of the config. When the old config is freed, they may still have references which become invalid. Input manager does this for devices.

So we probably need a "destroy" event for the config which gives everything an opportunity to dump their old references.

martinetd:
Right now I do not see anything that does, for example the strings allocated in input config for xkb used in sway_keyboard_configure aren't referenced anymore once the function exits so there is no risk (even if now I looked I can see that free_input_config() is far from complete in its freeing, leak sanitizer doesn't catch fields I don't allocate, but that's another problem orthogonal to this PR)

The free_config() calls appropriate subcomponent functions for each bit which each are responsible to do their own cleanup, I'm not even sure if a more global event would be more convenient?
Anyway I don't really want to implement an event that's not used - whoever needs it first will add it somewhere appropriate on first use. Ultimately config is just a few values and strings, it's probably more convenient to just say config is config and whatever else is an independant copy that has its own independant lifespan.

SirCmpwn:
>Some things may hold onto elements of the config. When the old config is freed, they may still have references which become invalid when not overwritten. Input manager does this for devices.

Whoops, I should have caught this in the review. This is invalid, nothing should store references to the config but should instead copy the data out.

acrisci:
Code looks ok, but it seems to break bindsym bindings for some reason, and there's still the issue with input holding onto configs instead of copies.

martinetd:
You seem to know exactly where the copy is held, and I said I didn't see it -- I've just looked again and I still don't, and this doesn't crash for me (with ASAN) so it must be some setting I'm not using. Could you please point at where is holding a copy? input is fairly vague...
I'll have a look at bindsym, I thought I checked that but will recheck tomorrow morning.

martinetd:
nm, found it; odd this doesn't trigger asan though I do have blocks there

martinetd:
Ok, this isn't trivial to copy the config as it's got nested pointer, and this code just sucks (seriously 1h is way too long to understand 600 lines of code, and I'm pretty sure I still don't understand it -- I'm not a genius and I'm pretty sure being tired doesn't help, but this is horrible)
In practice it looks like nothing is needed, as we re-assign the newly allocated config on reload before the old config is free, so the pointer is always valid. I've just started hating this code. Please fix that yourself later if you're not happy with this conclusion or just drop this PR.

FWIW, bindsyms work fine for me with drm, what backend are you using?
One thing I hadn't noticed when playing with it last time is that reload seems to trigger the same weird "X11 scaling changes" as switching VT does; haven't taken time to work on that but it is separate enough to be fixed in another PR imo.

acrisci:
I agree with you that the config parser probably needs a rewrite.

It is my responsibility to fix my code and this is pretty high on my prioirty list, so don't worry about it if you don't have time. I know where the references are.

To reproduce, issue a `reload` command then your bindsyms might stop working. I need to investigate that too.

martinetd:
> In practice it looks like nothing is needed, as we re-assign the newly allocated config on reload before the old config is free, so the pointer is always valid. I've just started hating this code.

Having slept on this, I thought we had a problem if the input block is in conf so it's attached to a device, then we remove it entierly so new conf isn't linked through, then we use swaymsg to modify *another* input as that will call sway_input_manager_apply_input_config which walks all input_device configs including invalid ones; but the identifier used is device->identifier and not device->config->identifier so this works. I really think this stuff is working "fine" with regards to reload, can't come up with any way to access invalid pointer even if we sometimes have invalid references.
I won't disagree a rewrite would be useful, I was going to suggest I could do it now I'm less angry at this stuff but happy to let you do, and this is pretty much unrelated to this PR. We really really don't need to store the config anywhere, just parse the config tree on device plug only for applying conf instead of storing it, and assign seats once at the end of config parsing by walking through the input_configs/seat attach rules instead of trying to attach everytime a rule changes; pretty sure we don't really need to much with current_input_config around sway_input_manager_apply_input_config that way as well.

Re-`bindsym` - really need more info, I've reloaded a hundred times during input tests and my main bindsym to spawn a shell (alt-enter) still works everytime, rest works when I occasionally try. I also had tried removing or adding bindsyms and such without problem. What backend are you using?

martinetd:
Been playing with seat thanks to that input stuff; it looks like removing seat from config+reload doesn't remove the pointer either, so I assume it doesn't kill the wlr seat -- I think that'd make sense to do?

Most other config elements destroy everything after having instanciated new ones, for seat I could agree to not doing that as there is a lot bound to them in wlr, but we need to compare the list of seats in old and new configs and destroy any removed seat. Should I do that now?

acrisci:
Yeah that would be helpful.

martinetd:
Done that, appears to work for me (cursor poofs; on re-adding the seat block the cursor doesn't reappear but that might be related to the pending seat cursor PR I have in wlroots I think that's related ; there is a ghost cursor though as clicks/selection moves separately)

This already needed to poke into the input manager, insert generic grumbling as appropriate... FWIW we should also listen to seat destroy events and cleanup as appropriate, I don't think we do.

sway/config.c
```diff
@@ -53,53 +54,68 @@ static void free_mode(struct sway_mode *mode) {
 	free(mode);
 }
 
-void free_config(struct sway_config *config) {
+void free_config(struct sway_config *old_config) {
```
martinetd: This is pretty ugly but I need to access both old_config and the global variable 'config' inside this now; I'd usually be adding a second "optional" argument to compare to but that doesn't seem to be the style currently.
On final free_config at sway shutdown we won't be destroying seats (as they're all in the current config) but this is fine because the display going down should free stuff appropriately.
. . . now I'm writing that though this actually isn't fine for another reason, we aren't removing from the list as we process it so that should have been accessing freed names, I'll repush something to kludge that out, although this really makes me want to go for extra argument...

sway/config.c
```diff
@@ -53,7 +54,7 @@ static void free_mode(struct sway_mode *mode) {
 	free(mode);
 }
 
-void free_config(struct sway_config *config) {
+void free_config(struct sway_config *config, struct sway_config *new_config) {
```
martinetd: Done the extra argument instead, not sure if you can access the old code somewhere as force pushed but alternatiive to that was renaming config to old_config or something to access the global variable.

SirCmpwn: Don't have time to do a proper review but -1 to whatever's going on here

martinetd: Can do a separate pass instead of trying to force my way into accessing the new config here I guess, renaming this variable is just as bad in my opinion

SirCmpwn:
Anyone have test results on this?

sway/config.c
```diff
@@ -687,3 +711,7 @@ char *do_var_replacement(char *str) {
 	}
 	return str;
 }
+
+void load_swaybars() {
+	/* stub function for reload commnd, to fill when we restore swaybars */
```
SirCmpwn: Nah just take this out

SirCmpwn:
Code LGTM

PR #1556, opened 2018-01-05
Cleanup logging

martinetd:
This finishes cleaning up all sway_log that are currently compiled.

It's the first time I'm actually looking at wlroots logging, I'm not sure I like the way we call _strip_path in the macro because this calls _strip_path all the time even if log level isn't verbose enough or if condition doesn't match in sway_assert.
I guess ultimately it doesn't matter much, but was that a conscious decision when wlr_log was implemented? I'd be happy to transform it back like sway's by adding file/line as separate arguments to callback if there are afterthoughts.

### DIFF
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -79,7 +79,7 @@ struct ipc_response *ipc_recv_response(int socketfd) {
 error_2:
 	free(response);
 error_1:
-	sway_log(L_ERROR, "Unable to allocate memory for IPC response");
+	wlr_log(L_ERROR, "Unable to allocate memory for IPC response");
 	return NULL;
 }
 
--- a/common/log.c
+++ b/common/log.c
@@ -1,167 +1,26 @@
-#define _POSIX_C_SOURCE 199506L
-#include <errno.h>
-#include <libgen.h>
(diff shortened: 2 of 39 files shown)


4e554c4c:
Shouldn't things like `common/*.c` and `swaymsg/*` still use the sway_log logger? I don't think that we should make all the sway utilites link to wlroots.

SirCmpwn:
There's no reason not to link everything to wlroots. These tools don't function without sway.

acrisci:
Agree, there are legitimate reasons to keep them separate so you can toggle one or the other off. In the future, most sway issues will not require wlroots logging.

martinetd:
I don't care which, but let's not take too long to decide as right now the sway log init has been removed and none of the sway_log calls output anything.

Some more complex logging systems have a 2nd parameter orthogonal to log level (say "component") so we could toggle various components log levels independantly, if you think we need that.
wlroots could have a couple of components (backends, renderer, and types? maybe one for Xwayland too) and we could make extendable to stack other components on top for sway/whatever.
Anything's possible... We just need to agree on something.

SirCmpwn:
Thanks!

PR #1559, opened 2018-01-07
Prevent invalid free of workspace name

### LINKED ISSUE #1548: Invalid memory freeing
```
λ sway -v
sway version 0.15-rc1-48-g7c3b0ff (2018-01-03, branch "master")
```
After manual installation sway crashes at startup with the following error log

```
*** Error in `sway': free(): invalid pointer: 0x000055ff2aa11641 ***
```

Assuming that it tries to free the memory pointed to invalid location and then immediately crashes. Meanwhile weston works perfectly so I don't think that the problem with my wayland setup.

4e554c4c:
An allocated pointer was incremented before being freed in
`sway/workspace.c` which led to an invalid free. This has been fixed by
keeping the pointer in place and moving the data instead.

Fixes #1548

### DIFF
--- a/sway/workspace.c
+++ b/sway/workspace.c
@@ -68,8 +68,9 @@ char *workspace_next_name(const char *output_name) {
 			sway_log(L_DEBUG, "Got valid workspace command for target: '%s'", name);
 			char *_target = strdup(name);
 			strip_quotes(_target);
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #1563, opened 2018-01-10
Fix out of focus when using mouse to change workspace

nearffxx:
When you try to change the workspace with the mouse, upon mouse release event it will try focus on the OUTPUT 'WLC-1' handler for some reason, we just prevent it. Additionally it seems wlc_view_focus should be called with C_VIEW only.

[debug_log.c:98] Setting focus to 0x5649148455a0:1 (OUTPUT 'WLC-1')
[main.c:50] [wlc] compositor/view.c: 574 @ wlc_view_focus(): Tried to retrieve handle of wrong type (output != view)

[log.txt](https://github.com/swaywm/sway/files/1619627/log.txt)

### DIFF
--- a/sway/focus.c
+++ b/sway/focus.c
@@ -162,12 +162,12 @@ bool set_focused_container(swayc_t *c) {
 		}
 		if (c->type == C_VIEW) {
 			wlc_view_set_state(c->handle, WLC_BIT_ACTIVATED, true);
-		}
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Would appreciate some extra help testing this.

SirCmpwn:
Thanks!

PR #1570, opened 2018-01-17
Copy input config references

acrisci:
Don't hang onto copies of config pointers in the input manager or the seat.

For #1555.

### DIFF
--- a/sway/config/input.c
+++ b/sway/config/input.c
@@ -90,6 +90,16 @@ void merge_input_config(struct input_config *dst, struct input_config *src) {
 	}
 }
 
+struct input_config *copy_input_config(struct input_config *ic) {
+	struct input_config *copy = calloc(1, sizeof(struct input_config));
+	if (copy == NULL) {
+		wlr_log(L_ERROR, "could not allocate input config");
+		return NULL;
+	}
+	merge_input_config(copy, ic);
+	return copy;
+}
+
 void free_input_config(struct input_config *ic) {
 	if (!ic) {
 		return;
--- a/sway/config/seat.c
+++ b/sway/config/seat.c
@@ -99,6 +99,17 @@ void merge_seat_config(struct seat_config *dest, struct seat_config *source) {
 	}
 }
 
+struct seat_config *copy_seat_config(struct seat_config *seat) {
+	struct seat_config *copy = new_seat_config(seat->name);
+	if (copy == NULL) {
+		return NULL;
+	}
+
+	merge_seat_config(copy, seat);
+
+	return copy;
+}
+
 void free_seat_config(struct seat_config *seat) {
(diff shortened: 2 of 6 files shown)


SirCmpwn:
Dear god

martinetd:
Can't say I like the idea of making this even more complex.
See my explanation on #1555 on why the dangling pointers are never accessed, so we're fine at the moment (so it could get merged if I was convincing enough) and let me do a bigger rework during the next few weeks for this?

We don't need to store anything config related in seat;just parse the global config tree when something gets plugged in (as we currently do) and instead of storing it just apply it.
I had other things bothering me when I looked at this (and added to the mess with the seat removal), can't recall most of it right now but wanted to simplify how we apply seat config as well so it's less tangled together and hopefully will be a bit clearer to read.

sway/config/input.c
```diff
@@ -90,6 +90,16 @@ void merge_input_config(struct input_config *dst, struct input_config *src) {
 	}
 }
 
+struct input_config *copy_input_config(struct input_config *ic) {
+	struct input_config *copy = calloc(1, sizeof(struct input_config));
+	if (copy == NULL) {
+		wlr_log(L_ERROR, "could not allocate input config");
+		return NULL;
+	}
+	merge_input_config(copy, ic);
```
martinetd: merge should be the other way around?

sway/config/seat.c
```diff
@@ -99,6 +99,17 @@ void merge_seat_config(struct seat_config *dest, struct seat_config *source) {
 	}
 }
 
+struct seat_config *copy_seat_config(struct seat_config *seat) {
+	struct seat_config *copy = new_seat_config(seat->name);
+	if (copy == NULL) {
+		return NULL;
+	}
+
+	merge_seat_config(copy, seat);
```
martinetd: ditto

sway/config/input.c
acrisci: It looks correct to me. The destination should be `copy` and the source should be `ic`.

acrisci:
I think you're right that the we don't need to hold onto the config in the current implementation. The intention is future proofing since seat/device configuration is very incomplete and some very ambitious things are planned to that end.

sway/config/input.c
martinetd: doh, you're right, I was convinced it was the other way around for some reason... Even opened its implementation and didn't read what I had under my eyes, I need more sleep -_-

martinetd:
hmm, unless we access it very frequently I think it'll always be simpler to just look it up whenever we need it occasionally, but I'll admit I don't know what ambitions you have so I'll leave it to you.
I'll regroup all the tidbits I wanted to change and open an issue so we can discuss it

PR #1574, opened 2018-01-20
Command criteria

acrisci:
Make command criteria work.

Add a `kill` command, demonstrates new conventions for running bindings and includes general config refactoring.

Config refactor:

* Remove globals `current_input_config` and `current_seat_config` and be more careful about how these are managed.
* Set config handler context for the seat of the keyboard that executes the command
* When criteria matches, run each command with `current_container` config context for each container that matches.

Implemented criteria:

* class
* con_id
* app_id
* instance
* title

Test:

Kill the currently focused window.

```
bindsym Alt+w kill
```

Kill *only* the gnome-terminals.

```
bindsym Alt+w [app_id="gnome-terminal-server"] kill
```

Test the other criteria listed above similarly.

### DIFF
--- a/sway/commands/kill.c
+++ b/sway/commands/kill.c
@@ -15,7 +15,13 @@ struct cmd_results *cmd_kill(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, NULL, "no seat context given");
 	}
 
-	struct sway_view *view = seat->focus->sway_view;
+	struct sway_view *view = NULL;
+
+	if (config->handler_context.current_container) {
+		view = config->handler_context.current_container->sway_view;
+	} else {
+		view = seat->focus->sway_view;
+	}
 
 	if (view->iface.close) {
 		view->iface.close(view);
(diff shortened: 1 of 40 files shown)


acrisci:
Ready for testing.

sway/commands/kill.c
```diff
@@ -0,0 +1,23 @@
+#include <wlr/util/log.h>
+#include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
+#include "sway/view.h"
+#include "sway/commands.h"
+
+struct cmd_results *cmd_kill(int argc, char **argv) {
+	if (!config->handler_context.current_container) {
```
SirCmpwn: This should probably use sway_assert

SirCmpwn:
I like it aside from that sway_assert issue. Would appreciate someone else looking into testing it, I still don't have logind set up properly.

sway/commands/kill.c
```diff
@@ -0,0 +1,24 @@
+#include <wlr/util/log.h>
+#include "log.h"
+#include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
+#include "sway/view.h"
+#include "sway/commands.h"
+
+struct cmd_results *cmd_kill(int argc, char **argv) {
+	if (!sway_assert(config->handler_context.current_container,
```
emersion: Add a `config->reading` check just before

PR #1584, opened 2018-02-04
Handle swaybar status line errors

### LINKED ISSUE #1583: High cpu usage from swaybar when status_command argument is not installed
If I try to add `i3status` to `bar` but it is not installed, then `swaybar` runs at full 100% CPU and eventually shoots temperature to 100°C.

My bar:
```
bar {
    status_command i3status
}
```
It seems to be polling like mad:
```
poll([{fd=14, events=POLLIN}, {fd=4, events=POLLIN}, {fd=13, events=POLLIN}, {fd=5, events=POLLIN}, {fd=9, events=POLLIN}], 5, -1) = 1 ([{fd=13, revents=POLLHUP}])
```
Trace is not too useful without symbols but might be helpful?
```
(gdb) bt
#0  0x00007f7361ac3944 in poll () at /usr/lib/libc.so.6
#1  0x0000556f015705e3 in event_loop_poll ()
#2  0x0000556f0156de27 in bar_run ()
#3  0x0000556f0156bc9f in main ()
```
I can more or less reproduce this, so I can take further steps if needed (hopefully with some guidance).

4e554c4c:
The event loop API was redesigned to avoid race conditions as well.
Fixes #1583

### DIFF
--- a/swaybar/event_loop.c
+++ b/swaybar/event_loop.c
@@ -20,6 +20,19 @@ struct timer_item {
 	void *data;
 };
 
+enum state_item_flags {
+	ITEM_IS_FD,
+	ITEM_IS_TIMER,
+};
+
+struct state_item {
+	enum state_item_flags flags;
+	union {
+		int fd;
+		timer_t timer;
+	} inner;
+};
+
 static struct {
 	// The order of each must be kept consistent
 	struct {   /* pollfd array */
@@ -31,6 +44,9 @@ static struct {
 
 	// Timer list
 	list_t *timers;
+
+	// List of state changes at the end of each iteration
+	list_t *state;
 } event_loop;
 
 void add_timer(timer_t timer,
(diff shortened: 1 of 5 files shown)


swaybar/event_loop.c
```diff
@@ -20,6 +20,19 @@ struct timer_item {
 	void *data;
 };
 
+enum state_item_flags {
+	ITEM_IS_FD = 0 << 0,
+	ITEM_IS_TIMER = 1 << 0,
+};
```
ascent12: These should be `1 << 0` and `1 << 1` if this is trying to be a bitmask.
But seeing as it's a tag for a tagged union, I don't think that's actually what you're going for, so there would be no reason for the weird `<< 0`.

4e554c4c: For awhile I was thinking about making this a bitmask so different operations could be added to the flags (change mask, add event, etc), but now there isn't any of that so I'll remove it

PR #1585, opened 2018-02-04
focus overhaul

acrisci:
Changes:

No new features added or bugs fixed, just refactoring changes.

* Keep an up-to-date list of all the containers of the seat in focus order
* Add `sway_seat_get_focus` to get the "focus inactive" of the container (returns true focus if the seat has focus and container is the root)
* Add `sway_seat_container` to store seat-specific data about a container (here, the focus stack list link and destroy handler).

Features:

* [x] static bfs queue
* [x] separate config/command command handlers

Tests:

* [x] open several windows and close one and focus should revert to the last container you focused
* [x] move focus with swaymsg/binding in the basic directions (left, right, up, down).
* [x] move focus across outputs
* [x] test on vertical layouts
* [ ] bug: focused popups don't render on top when in the middle of the children list.

Hard problems:

* [ ] are view coordinates in the output coordinate system or the layout coordinate system?
* [ ] how do I iterate a list of containers in rendering order? (BFS, DFS, or something else?)
* [ ] how do I iterate a list of containers in input order? (reverse of rendering order)
* [ ] are there any use cases where the focus-inactive algorithm gives the wrong answer or is inefficient?

Test config:

```
bindsym Alt+h focus left
bindsym Alt+j focus down
bindsym Alt+k focus up
bindsym Alt+l focus right

bindsym Alt+z layout splith
bindsym Alt+x layout splitv
bindsym Alt+c layout toggle split
```

Feedback on the approach is welcome.

### DIFF
--- a/include/sway/container.h
+++ b/include/sway/container.h
@@ -164,4 +164,6 @@ swayc_t *swayc_at(swayc_t *parent, double lx, double ly,
 void container_for_each_bfs(swayc_t *con, void (*f)(swayc_t *con, void *data),
 		void *data);
 
+swayc_t *swayc_change_layout(swayc_t *container, enum swayc_layouts layout);
+
 #endif
--- a/include/sway/container.h
+++ b/include/sway/container.h
@@ -163,9 +163,9 @@ swayc_t *swayc_at(swayc_t *parent, double lx, double ly,
 		struct wlr_surface **surface, double *sx, double *sy);
 
 /**
- * Get a list of containers that are descendents of the container in rendering
- * order
+ * Apply the function for each child of the container breadth first.
  */
-list_t *container_list_children(swayc_t *con);
+void container_for_each_bfs(swayc_t *con, void (*f)(swayc_t *con, void *data),
+		void *data);
 
 #endif
--- a/include/sway/container.h
+++ b/include/sway/container.h
@@ -162,4 +162,10 @@ void container_map(swayc_t *container,
 swayc_t *swayc_at(swayc_t *parent, double lx, double ly,
 		struct wlr_surface **surface, double *sx, double *sy);
 
+/**
+ * Get a list of containers that are descendents of the container in rendering
+ * order
+ */
+list_t *container_list_children(swayc_t *con);
+
 #endif
--- a/include/sway/container.h
+++ b/include/sway/container.h
@@ -106,10 +106,6 @@ struct sway_container {
 	 * The parent of this container. NULL for the root container.
 	 */
 	struct sway_container *parent;
-	/**
-	 * Which of this container's children has focus.
-	 */
-	struct sway_container *focused;
 
 	/**
 	 * Number of master views in auto layouts.
--- a/include/sway/workspace.h
+++ b/include/sway/workspace.h
@@ -1,7 +1,7 @@
 #ifndef _SWAY_WORKSPACE_H
 #define _SWAY_WORKSPACE_H
 
-struct sway_container;
+#include <sway/container.h>
 
 extern char *prev_workspace_name;
 
@@ -12,9 +12,9 @@ bool workspace_switch(swayc_t *workspace);
 struct sway_container *workspace_by_number(const char* name);
 swayc_t *workspace_by_name(const char*);
 
-struct sway_container *workspace_output_next(struct sway_container *current);
-struct sway_container *workspace_next(struct sway_container *current);
-struct sway_container *workspace_output_prev(struct sway_container *current);
-struct sway_container *workspace_prev(struct sway_container *current);
+struct sway_container *workspace_output_next(swayc_t *current);
+struct sway_container *workspace_next(swayc_t *current);
+struct sway_container *workspace_output_prev(swayc_t *current);
+struct sway_container *workspace_prev(swayc_t *current);
 
 #endif
--- a/include/sway/workspace.h
+++ b/include/sway/workspace.h
@@ -1,7 +1,7 @@
 #ifndef _SWAY_WORKSPACE_H
 #define _SWAY_WORKSPACE_H
 
-#include <sway/container.h>
+#include "sway/container.h"
 
 extern char *prev_workspace_name;
 
--- a/sway/commands/kill.c
+++ b/sway/commands/kill.c
@@ -10,11 +10,16 @@ struct cmd_results *cmd_kill(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "kill",
 			"Command 'kill' cannot be used in the config file");
 	}
+	if (config->handler_context.current_container == NULL) {
+		wlr_log(L_DEBUG, "no container to kill");
+		return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+	}
 	enum swayc_types type = config->handler_context.current_container->type;
-	if (type != C_VIEW || type != C_CONTAINER) {
+	if (type != C_VIEW && type != C_CONTAINER) {
 		return cmd_results_new(CMD_INVALID, NULL,
 				"Can only kill views and containers with this command");
 	}
+
 	// TODO close arbitrary containers without a view
 	struct sway_view *view =
 		config->handler_context.current_container->sway_view;
--- a/sway/commands/kill.c
+++ b/sway/commands/kill.c
@@ -6,14 +6,6 @@
 #include "sway/commands.h"
 
 struct cmd_results *cmd_kill(int argc, char **argv) {
-	if (config->reading) {
-		return cmd_results_new(CMD_FAILURE, "kill",
-			"Command 'kill' cannot be used in the config file");
-	}
-	if (config->handler_context.current_container == NULL) {
-		wlr_log(L_DEBUG, "no container to kill");
-		return cmd_results_new(CMD_SUCCESS, NULL, NULL);
-	}
 	enum swayc_types type = config->handler_context.current_container->type;
 	if (type != C_VIEW && type != C_CONTAINER) {
 		return cmd_results_new(CMD_INVALID, NULL,
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -219,23 +219,11 @@ static void output_frame_notify(struct wl_listener *listener, void *data) {
 }
 
 static void handle_output_destroy(struct wl_listener *listener, void *data) {
+	struct sway_output *output = wl_container_of(listener, output, output_destroy);
 	struct wlr_output *wlr_output = data;
 	wlr_log(L_DEBUG, "Output %p %s removed", wlr_output, wlr_output->name);
 
-	swayc_t *output_container = NULL;
-	for (int i = 0 ; i < root_container.children->length; ++i) {
-		swayc_t *child = root_container.children->items[i];
-		if (child->type == C_OUTPUT &&
-				child->sway_output->wlr_output == wlr_output) {
-			output_container = child;
-			break;
-		}
-	}
-	if (!output_container) {
-		return;
-	}
-
-	destroy_output(output_container);
+	destroy_output(output->swayc);
 }
 
 void handle_new_output(struct wl_listener *listener, void *data) {
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -57,46 +57,7 @@ static void render_surface(struct wlr_surface *surface,
 	};
 	if (wlr_output_layout_intersects(layout, wlr_output, &render_box)) {
 		float matrix[16];
-
-		float translate_center[16];
-		wlr_matrix_translate(&translate_center,
-			(int)ox + render_width / 2, (int)oy + render_height / 2, 0);
-
-		float rotate[16];
-		wlr_matrix_rotate(&rotate, rotation);
-
-		float translate_origin[16];
-		wlr_matrix_translate(&translate_origin, -render_width / 2,
-			-render_height / 2, 0);
-
-		float scale[16];
-		wlr_matrix_scale(&scale, render_width, render_height, 1);
-
-		float transform[16];
-		wlr_matrix_mul(&translate_center, &rotate, &transform);
-		wlr_matrix_mul(&transform, &translate_origin, &transform);
-		wlr_matrix_mul(&transform, &scale, &transform);
-
-		if (surface->current->transform != WL_OUTPUT_TRANSFORM_NORMAL) {
-			float surface_translate_center[16];
-			wlr_matrix_translate(&surface_translate_center, 0.5, 0.5, 0);
-
-			float surface_transform[16];
-			wlr_matrix_transform(surface_transform,
-				wlr_output_transform_invert(surface->current->transform));
-
-			float surface_translate_origin[16];
-			wlr_matrix_translate(&surface_translate_origin, -0.5, -0.5, 0);
-
-			wlr_matrix_mul(&transform, &surface_translate_center,
-				&transform);
-			wlr_matrix_mul(&transform, &surface_transform, &transform);
-			wlr_matrix_mul(&transform, &surface_translate_origin,
-				&transform);
-		}
-
-		wlr_matrix_mul(&wlr_output->transform_matrix, &transform, &matrix);
-
+		wlr_matrix_project_box(&matrix, &render_box, surface->current->transform, 0, &wlr_output->transform_matrix);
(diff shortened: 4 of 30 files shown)


include/sway/container.h
```diff
@@ -162,4 +162,10 @@ void container_map(swayc_t *container,
 swayc_t *swayc_at(swayc_t *parent, double lx, double ly,
 		struct wlr_surface **surface, double *sx, double *sy);
 
+/**
+ * Get a list of containers that are descendents of the container in rendering
+ * order
```
SirCmpwn: Can you define "rendering order" for me

sway/commands/kill.c
```diff
@@ -10,11 +10,16 @@ struct cmd_results *cmd_kill(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "kill",
 			"Command 'kill' cannot be used in the config file");
 	}
+	if (config->handler_context.current_container == NULL) {
+		wlr_log(L_DEBUG, "no container to kill");
+		return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+	}
 	enum swayc_types type = config->handler_context.current_container->type;
-	if (type != C_VIEW || type != C_CONTAINER) {
+	if (type != C_VIEW && type != C_CONTAINER) {
```
SirCmpwn: derp

sway/input/input-manager.c
```diff
@@ -282,7 +282,7 @@ bool sway_input_manager_has_focus(struct sway_input_manager *input,
 		swayc_t *container) {
 	struct sway_seat *seat = NULL;
 	wl_list_for_each(seat, &input->seats, link) {
-		if (seat->focus == container) {
+		if (seat->has_focus && sway_seat_get_focus(seat, &root_container) == container) {
```
SirCmpwn: `sway_seat_get_focus` can just return null and then you don't have to check has_focus. Does has_focus need to exist at all?

sway/input/seat.c
```diff
@@ -52,6 +90,24 @@ struct sway_seat *sway_seat_create(struct sway_input_manager *input,
...
+	}
+	for (int i = containers->length - 1; i >= 0; --i) {
+		swayc_t *con = containers->items[i];
+		seat_container_from_container(seat, con);
+	}
+	free(containers);
+
+	wl_signal_add(&root_container.sway_root->events.new_container,
+		&seat->new_container);
+	seat->new_container.notify = handle_new_container;
```
SirCmpwn: Is it necessary for every container to be represented in every focus stack? If not found, can't we just go up the parents until we find a container that works?

SirCmpwn: (or return null if nothing is focused)

sway/input/seat.c
```diff
@@ -211,42 +270,74 @@ static void handle_focus_destroy(struct wl_listener *listener, void *data) {
 }
 
 void sway_seat_set_focus(struct sway_seat *seat, swayc_t *container) {
-	swayc_t *last_focus = seat->focus;
+	swayc_t *last_focus =
+		(seat->has_focus ? sway_seat_get_focus(seat, &root_container) : NULL);
```
SirCmpwn: Yeah seeing all of this boilerplate everywhere I am super on board with `sway_seat_get_focus` returning NULL and ditching `has_focus`

sway/tree/container.c
```diff
@@ -378,3 +380,26 @@ void container_map(swayc_t *container, void (*f)(swayc_t *view, void *data), voi
...
+		current = queue->items[0];
+		list_del(queue, 0);
+		list_add(list, current);
+		// TODO floating containers
+		list_cat(queue, current->children);
+	}
+
+	list_free(queue);
+	return list;
+}
```
SirCmpwn: Uh, why is this allocating memory? Maybe make this generator-like and instead of returning a list, take a callback?

sway/tree/layout.c
```diff
@@ -48,17 +48,47 @@ void init_layout(void) {
...
+	wl_signal_init(&root_container.sway_root->events.new_container);
 
 	root_container.sway_root->output_layout_change.notify =
 		output_layout_change_notify;
 	wl_signal_add(&root_container.sway_root->output_layout->events.change,
 		&root_container.sway_root->output_layout_change);
 }
 
+int index_child(const swayc_t *child) {
+	// TODO handle floating
```
SirCmpwn: FYI floating is probably what should be done next

sway/tree/layout.c
```diff
@@ -48,17 +48,47 @@ void init_layout(void) {
...
 	wl_list_init(&root_container.sway_root->unmanaged_views);
+	wl_signal_init(&root_container.sway_root->events.new_container);
 
 	root_container.sway_root->output_layout_change.notify =
 		output_layout_change_notify;
 	wl_signal_add(&root_container.sway_root->output_layout->events.change,
 		&root_container.sway_root->output_layout_change);
 }
 
+int index_child(const swayc_t *child) {
```
SirCmpwn: Can this be static?

sway/tree/workspace.c
```diff
@@ -102,7 +103,8 @@ swayc_t *workspace_create(const char *name) {
 	}
 	// Otherwise create a new one
 	struct sway_seat *seat = input_manager_current_seat(input_manager);
-	parent = seat->focus;
+	swayc_t *focus = sway_seat_get_focus(seat, &root_container);
```
SirCmpwn: Let's not call this multiple times, cache the call up above instead

SirCmpwn:
Looking pretty good to me so far, the approach seems good

sway/input/input-manager.c
acrisci: The design intent is a bit subtle because I'm trying to design for a specific future use case.

The "focus inactive" is emphasized because it is needed for rendering ("focused inactive" has different border colors) and determining where focus goes when entering a new container.

For instance, if I enter a container with some children and want to focus a leaf, focus should go to the "focus inactive" as defined as the last child container to receive focus within the container. In this case you use `sway_seat_get_focus(seat, &con)` for that specific container to get the right answer. In this case, it will not actually give you the absolute focused container.

The two use cases might actually be separate enough to warrant different functions. Then you can get rid of those checks for the main use case and have `sway_seat_get_focus(seat)` which acts on the root container and implicitly checks the `has_focus`, then another function what we have here which would be renamed `sway_seat_get_focus_inactive(seat, con)`.

> Does has_focus need to exist at all?

I put this in because lots of places in the code right now checks if there is actually a focused window, but I'm not sure if we can ever actually get into this situation. For instance, if there are no containers below the workspace level, then the workspace itself could get implicit focus.

If you don't see a problem with assuming that there is *always* a focused container, then we can take it out. Otherwise the algorithm we use here will always return the first thing in the focus stack.

sway/tree/container.c
acrisci: What about making it one of those cool things with a block like `wl_list_for_each()`?

include/sway/container.h
acrisci: "rendering order" means the order in which containers go through the rendering loop. Prior elements in rendering order are rendered "on top of" later elements. Parent containers are rendered below their children, so they will be earlier in the list. Floating containers appear on top of tiled containers, so they go in the list earlier.

For a given cursor coordinate, the cursor position might be under several containers at the same time, such as when the cursor is over a floating window with a tiled window underneath it. To determine which window gets the mouse click, we find the first window in "reverse rendering order". This is the window the user sees on top of the others.

I'm sort of inventing this vocabulary right now, but I think it fits the domain.

sway/input/seat.c
acrisci: My reasoning is that if we're ever in a situation where a new container doesn't get focused right away, this can be helpful to determine focus inactive because it encodes in what order the containers appeared. It might not be so easy though.

SirCmpwn: I suppose that's reasonable.

sway/tree/container.c
SirCmpwn: While I think that heavy reliance on wl_list is appropriate for wlroots, it's not how Sway was written and I don't want to introduce new patterns. The map approach is more appropriate here. Also, macros are the devil.

sway/tree/layout.c
```diff
@@ -346,3 +347,179 @@ void apply_vert_layout(swayc_t *container,
...
+		switch (dir) {
+		case MOVE_LEFT:
+			// get most right child of new output
+			return ws->children->items[ws->children->length-1];
+		case MOVE_RIGHT:
+			// get most left child of new output
+			return ws->children->items[0];
+		case MOVE_UP:
+		case MOVE_DOWN:
+			{
```
SirCmpwn: This probably comes from old code but can you move this { over to the same line as the case and deindent the whole thing so it's not as crazy

acrisci:
Windows aren't being arranged correctly in the multi output case (including DRM and wayland) but actually I don't think that ever completely worked correctly.

include/sway/workspace.h
```diff
@@ -1,7 +1,7 @@
 #ifndef _SWAY_WORKSPACE_H
 #define _SWAY_WORKSPACE_H
 
-struct sway_container;
+#include <sway/container.h>
```
martinetd: local includes should use double-quotes?

sway/desktop/output.c
```diff
@@ -216,11 +216,12 @@ static void output_frame_notify(struct wl_listener *listener, void *data) {
 	struct sway_output *soutput = wl_container_of(listener, soutput, frame);
 	struct wlr_output *wlr_output = data;
 	struct sway_server *server = soutput->server;
-
 	float clear_color[] = {0.25f, 0.25f, 0.25f, 1.0f};
 	struct wlr_renderer *renderer = wlr_backend_get_renderer(wlr_output->backend);
 	wlr_renderer_clear(renderer, &clear_color);
```
themindfulcoder: Why is this called twice?

acrisci: 👍

sway/tree/container.c
```diff
@@ -151,19 +151,19 @@ swayc_t *new_output(struct sway_output *sway_output) {
 	char *ws_name = workspace_next_name(output->name);
 	wlr_log(L_DEBUG, "Creating default workspace %s", ws_name);
 	swayc_t *ws = new_workspace(output, ws_name);
-	output->focused = ws;
 	// Set each seat's focus if not already set
 	// TODO FOCUS: this is probably stupid, we shouldn't define focus in two
 	// places. We should probably put the active workspace on the sway_output
 	// struct instead of trying to do focus semantics like this
```
SirCmpwn: This comment can be removed

sway/tree/container.c
```diff
@@ -378,3 +381,25 @@ void container_map(swayc_t *container, void (*f)(swayc_t *view, void *data), voi
...
+	while (queue->length) {
+		current = queue->items[0];
+		list_del(queue, 0);
+		f(current, data);
+		// TODO floating containers
+		list_cat(queue, current->children);
+	}
+
+	list_free(queue);
+}
```
SirCmpwn: Just iterate over the children twice

```c
for child : con.children {
    f(child, data);
}
for child : con.children {
    container_for_each_bfs(child, f, data);
}
```

I still don't like allocating a list here

sway/tree/layout.c
```diff
@@ -321,3 +347,247 @@ void apply_vert_layout(swayc_t *container,
...
 	}
 }
+
+/**
+ * Get swayc in the direction of newly entered output.
+ */
+static swayc_t *get_swayc_in_output_direction(swayc_t *output,
+		enum movement_direction dir, struct sway_seat *seat) {
+	// XXX is this really a seat function or can we do it with the default
+	// seat?
```
SirCmpwn: We should keep it explicit

sway/tree/layout.c
```diff
@@ -321,3 +347,247 @@ void apply_vert_layout(swayc_t *container,
...
+		default:
+			break;
+		}
+	}
+
+	return ws;
+}
+
+static void get_layout_center_position(swayc_t *container, int *x, int *y) {
+	// FIXME view coords are inconsistently referred to in layout/output systems
```
SirCmpwn: Ticket for this

sway/tree/container.c
acrisci: I'm not sure how to do a bfs without allocating a queue. Maybe the queue could be allocated once and used for the lifetime of the application?

It gets worse when we want to do reverse bfs because that requires two lists: one for the queue, and one for the result in reverse order.

Event worse, I'm also thinking that BFS with containers in `children` array order might not truly be rendering order, but rather based on the focus-inactive stack. So a child later in the stack might need to be rendered on top even though they are siblings (if for instance it is the focused window). You can see this bug right now when the popups of the focused window will render below a sibling.

SirCmpwn: Static queue sounds good to me

acrisci:
I fixed the easy problems and updated the original post with the hard problems.

emersion:
>are view coordinates in the output coordinate system or the layout coordinate system?

Layout coordinates. You don't want to deal with overlapping outputs with a different scale when using output coordinates. You want to keep the views at the same position when dynamically changing an output's scale. Output coordinates should only be used for rendering stuff.

SirCmpwn:
>are view coordinates in the output coordinate system or the layout coordinate system?

Agree with emersion

> how do I iterate a list of containers in rendering order? (BFS, DFS, or something else?)

Either one should be correct. It'll get weird when we start doing tabbed/stacked layouts though.

>how do I iterate a list of containers in input order? (reverse of rendering order)

Either one should be correct. It'll get weird when we start doing tabbed/stacked layouts though.

>are there any use cases where the focus-inactive algorithm gives the wrong answer or is inefficient?

Let's cross that bridge when we get to it?

acrisci:
Can you guys summarize your reviews so I can make a list of todos for this pr?

SirCmpwn:
>It's going to get pretty tiresome checking this everywhere. Maybe we should pull this higher up and put contextual commands into a different command list? Will file a ticket.
￼
This

>Maybe the queue could be allocated once and used for the lifetime of the application?

This

acrisci:
> It's going to get pretty tiresome checking this everywhere. Maybe we should pull this higher up and put contextual commands into a different command list?

This is a general problem with the approach of setting things on a global. It's kind of hard to reason about.

Seat and container context should always be there for every command or that's an application error. That's an easy check to do before hand since running a command can only happen in two places.

Then to get rid of the rest, it would be an obvious split to partition the commands into 1) commands that can run in the config and 2) commands that can't and then do the check accordingly earlier.

acrisci:
Another idea along those lines is to instead of passing a single container to the command, pass a list (for every container that was matched by the criteria). That way, commands might choose to behave a bit differently with a list. For instance `[class=".*"] focus` will get a list of all the windows and can choose to only focus the first one, or maybe do your multi-focus thing or whatever.

SirCmpwn:
>Another idea along those lines is to instead of passing a single container to the command, pass a list (for every container that was matched by the criteria). That way, commands might choose to behave a bit differently with a list. For instance [class=".*"] focus will get a list of all the windows and can choose to only focus the first one, or maybe do your multi-focus thing or whatever.

Eh, I'm not feeling this (at least for the time being)

acrisci:
> It's going to get pretty tiresome checking this everywhere. Maybe we should pull this higher up and put contextual commands into a different command list? Will file a ticket.

Ok this solution partitions the handlers into three groups. 1) `command_handlers` can run *after* the config has been parsed, 2) `config_handlers` can run *only* when the config is parsing, and 3) `handlers` can run at any time.

The `find_handler()` function now uses the config reading state to find the correct handler. For instance, if the config is reading, it will ignore the command handlers and if config is not reading, it will ignore the config commands.

```c
static struct cmd_handler handlers[] = {
	{ "bindcode", cmd_bindcode },
	{ "bindsym", cmd_bindsym },
	{ "exec", cmd_exec },
	{ "exec_always", cmd_exec_always },
	{ "include", cmd_include },
	{ "input", cmd_input },
	{ "output", cmd_output },
	{ "seat", cmd_seat },
	{ "workspace", cmd_workspace },
};

static struct cmd_handler config_handlers[] = {
	{ "set", cmd_set },
};

static struct cmd_handler command_handlers[] = {
	{ "exit", cmd_exit },
	{ "focus", cmd_focus },
	{ "kill", cmd_kill },
	{ "layout", cmd_layout },
	{ "reload", cmd_reload },
};
```

PR #1590, opened 2018-02-11
Fix memory errors

dlrobertson:
- `swaylock` does not free the password buffer on ENTER
 - `read_line`: OOB write when a line in /proc/modules contains a
   terminating character at `size` position.

Some errors noted on a sanitized build.

### DIFF
--- a/common/readline.c
+++ b/common/readline.c
@@ -36,7 +36,7 @@ char *read_line(FILE *file) {
 		}
 		string[length++] = c;
 	}
-	if (length + 1 == size) {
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Thanks - please retarget this patch for the 0.15 branch.

dlrobertson:
:+1: looks like one is fixed in `0.15`.

dlrobertson:
Also was there a reason sanitized builds were removed? They can be super helpful for finding memory errors.

dlrobertson:
See #1591

PR #1591, opened 2018-02-11
Fix memory errors

dlrobertson:
- `read_line`: OOB write when a line in `/proc/modules` contains a
   terminating character at `size` position.
 - `handle_view_created`: Ensure that the `list_t` returned by `criteria_for`
   is `free`'d after use
 - `ipc_event_binding_keyboard`/`ipc_event_binding`: Properly handle
   `json_object` reference counting and ownership.

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -1127,7 +1127,7 @@ static void ipc_event_binding(json_object *sb_obj) {
 	json_object *obj = json_object_new_object();
 	json_object_object_add(obj, "change", json_object_new_string("run"));
 	// sb_obj gets owned by the temporary json_object, too.
-	json_object_object_add(obj, "binding", json_object_get(sb_obj));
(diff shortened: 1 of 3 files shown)


sway/ipc-server.c
```diff
@@ -1127,7 +1127,7 @@ static void ipc_event_binding(json_object *sb_obj) {
 	json_object *obj = json_object_new_object();
 	json_object_object_add(obj, "change", json_object_new_string("run"));
 	// sb_obj gets owned by the temporary json_object, too.
-	json_object_object_add(obj, "binding", json_object_get(sb_obj));
+	json_object_object_add(obj, "binding", sb_obj);
```
dlrobertson: `json_object_get` bumps the ref count. As a result, ownership does not actually transferr to `obj`.

SirCmpwn:
Thanks!

PR #1592, opened 2018-02-11
Fix more leaks

dlrobertson:
- `get_parent_pid`: free buffer returned from `read_line` after use.
 - `workspace_for_pid`: ensure `free_pid_workspace` is called when
   `pid_workspace`s are removed from `config->pid_workspaces`.
 - `cmd_split`: return the `cmd_results` from `_do_split`, so that the
   parent function may free it.


SirCmpwn:
Thanks!

PR #1595, opened 2018-02-19
Fix segfault if swaybg is run without Wayland

nickbp:
Mirrors a similar check in `swaylock/main.c`


SirCmpwn:
Did you mean to make this for the 0.15 branch?

nickbp:
Wasn't sure what branches to apply it to. Which one(s) would you like it in? It seems to apply to both `wlroots` and `master`.

SirCmpwn:
0.15.

nickbp:
:+1: be right back

PR #1615, opened 2018-03-03
Enable docker service in travis

mikkeloscar:
Adds the docker service in travis which is now required by arch-travis.

See: https://github.com/mikkeloscar/arch-travis/pull/40


SirCmpwn:
Thanks!

PR #1616, opened 2018-03-03
xwayland: do not send surface configure when no width/height

martinetd:
The code in `apply_horiz_layout` systematically does `set_position`
then `set_size`, so for new windows there is an invalid call.

For old windows there are two calls when only one is needed,
with the current code set_position could not send any surface
configure without impact, but in the future it might be needed?
Native wayland surfaces do not need to know where they are (the
set_position handled only updates the sway internal view variable),
why does X11 window need that?

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -67,6 +67,10 @@ static void set_position(struct sway_view *view, double ox, double oy) {
 	view->swayc->x = ox;
 	view->swayc->y = oy;
 
+	if (view->width == 0 || view->height == 0) {
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


SirCmpwn:
>why does X11 window need that?

It's useful for getting a few things working like ibus or dmenu.

emersion:
>why does X11 window need that?

It seems to me that popups/menus are also positioned in the global compositor space by the client - not relative to their parent like in Wayland.

martinetd:
> It seems to me that popups/menus are also positioned in the global compositor space by the client - not relative to their parent like in Wayland.

Makes sense, can't really slack on notifiying Xwayland then.

PR #1622, opened 2018-03-12
meson: allow wlroots to be a subproject

emersion:
Usage: make `subprojects/wlroots` a symbolic link to wlroots.

This allows to edit and build sway and wlroots at the same time. This can be useful for features like multibackend.


emersion:
Added the subproject to `.gitignore`.

PR #1626, opened 2018-03-19
Update wlroots API

emersion:
Breaking changes in wlr_xdg_shell_v6 and wlr_renderer have been
made upstream.


SirCmpwn:
Not sure what happened with the build. Built and tested this locally.

Issue #1624, opened 2018-03-13
swaybar tray: scalable icons not found


mckinney-subgraph:
Hi,

I am working on a status tool that sends icons (using SNI) to the swaybar tray. It basically works but I am having some troubles getting the tray to pick up some icons. 

In the theme I am using (Paper), it does match some files for certain icons in directories like  `24x24` but not for other directories. So what I have is some icons showing up on the swaybar but not others. In particular, icons in the `scalable` directories are missing.

Is there any known reason icons in this directory may not be picked up? This directory is usually where icon themes dump their symbolic icons in 'SVG' format (in various subdirs). 

To test this further, I have modified `swaybar/tray/icon.c` to print the directories found by `find_theme_subdirs()`. The `scalable` directories do show up in the debug output. If I copy the `.svg` files manually to a directory such as `24x24`, they are also displayed properly -- so it is not a problem with the icons. However, I don't want to do this (or make users do this), the stock theme should be usable outside of the box IMO.

[index.theme.txt](https://github.com/swaywm/sway/files/1808395/index.theme.txt)

I have attached `index.theme` for Paper if that helps any.

SirCmpwn:
cc @4e554c4c

4e554c4c:
Thank you very much for debugging, I'll look into this.

mckinney-subgraph:
Closing -- sent a fix and it was merged :)

PR #1632, opened 2018-03-24
Fixed missing icons bug in index.theme parsing...

mckinney-subgraph:
This PR fixes the missing scalable icons issue I reported (#1624). The issue is that the directories were parsed from `index.theme` using `getline()` without chopping the newline at the end -- so the last sub-directory on the list could not be opened (`scalable/status` in the Paper icon theme).

Additionally I noticed that the `find_inherits()` function was concatenating the 'index.theme' string onto the directory without a forward-slash so I fixed that too.


SirCmpwn:
Please rebase your patch against the 0.15 branch.

PR #1633, opened 2018-03-24
Fixed missing icons bug in index.theme parsing (2)

mckinney-subgraph:
(re-submitting with rebase to 0.15 -- not sure if I did that correctly... if not I am happy to re-apply the patch cleanly against 0.15)

This PR fixes the missing scalable icons issue I reported (#1624). The issue is that the directories were parsed from `index.theme` using `getline()` without chopping the newline at the end -- so the last sub-directory on the list could not be opened (`scalable/status` in the Paper icon theme).

Additionally I noticed that the `find_inherits()` function was concatenating the 'index.theme' string onto the directory without a forward-slash so I fixed that too.


SirCmpwn:
>not sure if I did that correctly

No, doesn't look like you did. You're merging everything from the master branch into 0.15. Which isn't actually a terrible thing, as I'm looking over this commit log. So don't sweat it.

mckinney-subgraph:
Okay, thanks! :) I will be more careful next time.

SirCmpwn:
Thanks!

PR #1638, opened 2018-03-27
Reimplement swaybg using surface layers

SirCmpwn:
The last commit tears out some obsolete Wayland client code sway 0.x used to have. Diff is easier to read if you ignore that.


client/buffer-pool.c
```diff
@@ -98,37 +91,33 @@ static void destroy_buffer(struct buffer *buffer) {
...
+		if (!create_buffer(shm, buffer, width, height,
 					WL_SHM_FORMAT_ARGB8888)) {
 			return NULL;
 		}
 	}
-
-	window->cairo = buffer->cairo;
-	window->buffer = buffer;
 	return buffer;
 }
```
SirCmpwn: I've removed the "window" abstraction but kept this file for wl_buffer management. Refactored the old "window" abstraction out of it.

common/cairo.c
```diff
@@ -101,7 +92,9 @@ cairo_surface_t* gdk_cairo_image_surface_create_from_pixbuf(const GdkPixbuf *gdk
 		 * ------
 		 * tested as equal to lround(z/255.0) for uint z in [0..0xfe02]
 		 */
-#define PREMUL_ALPHA(x,a,b,z) G_STMT_START { z = a * b + 0x80; x = (z + (z >> 8)) >> 8; } G_STMT_END
+#define PREMUL_ALPHA(x,a,b,z) \
+		G_STMT_START { z = a * b + 0x80; x = (z + (z >> 8)) >> 8; } \
+		G_STMT_END
 		int i;
 		for (i = h; i; --i) {
 			const guint8 *gp = gdkpix;
```
SirCmpwn: This file contains formatting fixes but also addresses a long-time TODO in Sway 0.x to bring the cairo helpers into libsway-common so we could use them in both clients and the compositor

client/meson.build
```diff
@@ -0,0 +1,21 @@
+deps = [
+	cairo,
+	pango,
+	pangocairo,
+	wlroots,
+	wayland_client,
+]
+
+if gdk_pixbuf.found()
+	deps += [gdk_pixbuf]
```
emersion: You don't need to check if the dependency is found. Since it's an optional dep, adding it to the list will just work.

>You can pass the opt_dep variable to target construction functions whether the actual dependency was found or not. Meson will ignore non-found dependencies.

http://mesonbuild.com/Dependencies.html

SirCmpwn: woohoo

include/buffer_pool.h
```diff
@@ -0,0 +1,21 @@
+#ifndef _SWAY_BUFFERS_H
+#define _SWAY_BUFFERS_H
+#include <cairo/cairo.h>
+#include <pango/pangocairo.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <wayland-client.h>
+
+struct pool_buffer {
```
emersion: Rename this file to `pool_buffer.c`?

swaybg/main.c
```diff
@@ -52,163 +68,264 @@ bool is_valid_color(const char *color) {
...
+	wlr_log_init(L_DEBUG, NULL);
+
+	if (argc != 4) {
+		wlr_log(L_ERROR, "Do not run this program manually. "
+				"See man 5 sway and look for output options.");
+		return 1;
+	}
+	args.output_idx = atoi(argv[1]);
+	args.path = argv[2];
+	args.mode = atoi(argv[3]);
```
emersion: Should be removed

swaybg/main.c
```diff
@@ -52,163 +68,264 @@ bool is_valid_color(const char *color) {
...
+	wl_display_roundtrip(state.display);
+
+	if (!state.compositor) {
+		wlr_log(L_DEBUG, "wl-compositor not available");
+		return 1;
+	}
+	if (!state.layer_shell) {
+		wlr_log(L_ERROR, "layer-shell not available");
+		return 1;
+	}
```
emersion: Also check for `shm`, `output`

swaybg/main.c
```diff
@@ -52,163 +68,264 @@ bool is_valid_color(const char *color) {
...
+			ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT);
+	zwlr_layer_surface_v1_add_listener(state.layer_surface,
+			&layer_surface_listener, &state);
+	wl_surface_commit(state.surface);
+	wl_display_roundtrip(state.display);
 
-	for (i = 0; i < surfaces->length; ++i) {
-		struct window *window = surfaces->items[i];
-		window_teardown(window);
+	state.run_display = true;
```
emersion: We might want to move this before the first commit in case the compositor closes the layer surface immediately.

PR #1639, opened 2018-03-27
save errno to avoid issues with it being overwritten in sway_log_errno

taiyu-len:
was fiddling around with the logging to make it a bit more convenient to debug
and noticed a small mistake with `sway_log_errno`, as `errno` can get overwritten before we use it.

its been a while since i contributed, but i assume 0.15 is where development is going for now

### DIFF
--- a/common/log.c
+++ b/common/log.c
@@ -153,6 +153,7 @@ void sway_log_errno(log_importance_t verbosity, char* format, ...) {
 		}
 		fprintf(stderr, "\n");
 	}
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Should probably restore it too

ascent12:
It's probably not worth bothering.
You'd normally consider errno useless after any function you call.

SirCmpwn:
Thanks!

PR #1640, opened 2018-03-28
fix remove_view_from_scratchpad

### LINKED ISSUE #1468: Scratchpad show then floating toggle crashes sway.
The following sequence of commands on a window makes sway to crach : 
move to scratchpad
scratchpad show
floating toggle.

Can someone confirm? 
(And question : what is the correct way to get a window out of scratchpad?)

taiyu-len:
Fixes #1468 `Scratchpad show then floating toggle crashes sway.`
Could not reproduce with original instructions,
but can reproduce with instructions provided by nickbatsaras

1. Move a window to scratchpad
2. Move another window to scratchpad
3. Scratchpad show
4. Floating toggle
5. Scratchpad show
6. Scratchpad show

The first sign of the issue starts at step 4.
Given a scratchpad with 2 windows (1,2), with window 1 being a visible container,
turning off floating via `toggle floating` will call `remove_view_from_scratchpad`.
this function was incorrect and resulted in removing window 2,
leaving window 1 in the scratchpad when it shouldnt be and losing window 2 permanently.
the segfault was later caused in update_visibility_output due to window 1 having a NULL parent and dereferencing it.

this fix correctly removes the right container from the scratchpad list,
and  checks for null parent just in case,
also added printing out of scratchpad to layout_log

there might be something more to this that i missed, but for now it seems to work

### DIFF
--- a/sway/debug_log.c
+++ b/sway/debug_log.c
@@ -71,11 +71,11 @@ void layout_log(const swayc_t *c, int depth) {
 				layout_log(c->floating->items[i], depth + 1);
 			}
 		}
-	}
-	if (c->type == C_ROOT) {
-		fprintf(stderr,"ScratchPad\n");
-		for (int i = 0; i < scratchpad->length; ++i)
+	} else if (c->type == C_ROOT && scratchpad->length > 0) {
+		fprintf(stderr, "ScratchPad\n");
+		for (i = 0; i < scratchpad->length; ++i) {
 			container_log(scratchpad->items[i], 0);
+		}
 	}
 }
 
--- a/sway/debug_log.c
+++ b/sway/debug_log.c
@@ -72,6 +72,11 @@ void layout_log(const swayc_t *c, int depth) {
 			}
 		}
 	}
+	if (c->type == C_ROOT) {
+		fprintf(stderr,"ScratchPad\n");
+		for (int i = 0; i < scratchpad->length; ++i)
+			container_log(scratchpad->items[i], 0);
+	}
 }
 
 const char *swayc_type_string(enum swayc_types type) {
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -141,12 +141,12 @@ void remove_view_from_scratchpad(swayc_t *view) {
 	int i;
 	for (i = 0; i < scratchpad->length; i++) {
 		if (scratchpad->items[i] == view) {
+			list_del(scratchpad, sp_index);
 			if (sp_index == 0) {
 				sp_index = scratchpad->length - 1;
 			} else {
 				sp_index--;
 			}
-			list_del(scratchpad, sp_index);
 			sp_view = NULL;
 		}
 	}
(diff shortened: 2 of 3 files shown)


sway/debug_log.c
```diff
@@ -72,6 +72,11 @@ void layout_log(const swayc_t *c, int depth) {
 			}
 		}
 	}
+	if (c->type == C_ROOT) {
+		fprintf(stderr,"ScratchPad\n");
```
SirCmpwn: doop

sway/debug_log.c
```diff
@@ -71,6 +71,11 @@ void layout_log(const swayc_t *c, int depth) {
 				layout_log(c->floating->items[i], depth + 1);
 			}
 		}
+	} else if (c->type == C_ROOT && scratchpad->length > 0) {
+		fprintf(stderr, "ScratchPad\n");
```
SirCmpwn: doop

taiyu-len: Im guessing you want that line removed?
i left it in, so the scratchpad list is more clear. the output for example
```
focus:R      |(0x5555557a9320)(p:(nil)   )(f:0x55555583d4a0)(h:18446744073709551615)Type:root|layout:-    |w:1680|h:1818|x:   0|y:   0|g:  0|vis:t|children: 1|name:root
|focus:X     |(0x55555583d4a0)(p:0x5555557a9320)(f:0x55555583d820)(h: 1)Type:op  |layout:-    |w:1014|h: 746|x:   0|y:   0|g: -1|vis:t|children: 1|name:WLC-1
|-focus:W    |(0x55555583d820)(p:0x55555583d4a0)(f:0x555555fd6430)(h:18446744073709551615)Type:ws  |layout:Vert |w:1014|h: 746|x:   0|y:   0|g: -1|vis:t|children: 2|name:1
|--focus:X   |(0x555555fd4f90)(p:0x55555583d820)(f:(nil)   )(h: 4)Type:view|layout:-    |w:1014|h: 373|x:   0|y:   0|g: -1|vis:t|children: 0|name:Wayland Terminal
|--focus:K   |(0x555555fd6430)(p:0x55555583d820)(f:(nil)   )(h: 2)Type:view|layout:-    |w:1014|h: 373|x:   0|y: 373|g: -1|vis:t|children: 0|name:Wayland Terminal
ScratchPad
focus:X      |(0x5555560e69f0)(p:(nil)   )(f:(nil)   )(h: 3)Type:view|layout:-    |w: 507|h: 559|x: 253|y:  93|g: -1|vis:f|children: 0|name:Wayland Terminal
focus:K      |(0x555555fd6430)(p:0x55555583d820)(f:(nil)   )(h: 2)Type:view|layout:-    |w:1014|h: 373|x:   0|y: 373|g: -1|vis:t|children: 0|name:Wayland Terminal
```

SirCmpwn: Ah. Alright.

SirCmpwn:
Thanks!

PR #1647, opened 2018-03-29
Refactor tree

acrisci:
Various tree refactorings to prepare for split containers, decoration, and damage. Will add todos as I see them.

TODO:

* [x] organize tree includes
* [x] consistent naming conventions for functions
* [x] remove typedefs (`swayc_t`)
* [x] rewrite docstrings

NEXT:

* [ ] use destroy listeners for properties
* [ ] view x/y in layout coordinate system (not output)
* [ ] reevaluate use of `list_t`
* [ ] include coordinates in x/y names (surface, output, layout)

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -88,41 +88,38 @@ struct sway_container {
 };
 
 // TODO only one container create function and pass the type?
-struct sway_container *sway_container_output_create(
+struct sway_container *container_output_create(
 		struct sway_output *sway_output);
 
-struct sway_container *sway_container_workspace_create(
+struct sway_container *container_workspace_create(
 		struct sway_container *output, const char *name);
 
-struct sway_container *sway_container_view_create(
+struct sway_container *container_view_create(
 		struct sway_container *sibling, struct sway_view *sway_view);
 
-struct sway_container *sway_container_output_destroy(
+struct sway_container *container_output_destroy(
 		struct sway_container *output);
 
-struct sway_container *sway_container_view_destroy(struct sway_container *view);
+struct sway_container *container_view_destroy(struct sway_container *view);
 
-struct sway_container *sway_container_set_layout(
+struct sway_container *container_set_layout(
 		struct sway_container *container, enum sway_container_layout layout);
 
-void sway_container_descendents(struct sway_container *root,
+void container_descendents(struct sway_container *root,
 		enum sway_container_type type,
 		void (*func)(struct sway_container *item, void *data), void *data);
-
-// XXX: what is this?
-struct sway_container *next_view_sibling(struct sway_seat *seat);
 
 /**
  * Finds a container based on test criteria. Returns the first container that
  * passes the test.
  */
-struct sway_container *sway_container_find(struct sway_container *container,
+struct sway_container *container_find(struct sway_container *container,
 		bool (*test)(struct sway_container *view, void *data), void *data);
 
 /**
  * Finds a parent container with the given struct sway_containerype.
  */
-struct sway_container *sway_container_parent(struct sway_container *container,
+struct sway_container *container_parent(struct sway_container *container,
 		enum sway_container_type type);
 
 /**
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -20,60 +20,43 @@ struct sway_seat;
  * it on this list.
  */
 enum swayc_types {
-	C_ROOT,      /**< The root container. Only one of these ever exists. */
-	C_OUTPUT,    /**< An output (aka monitor, head, etc). */
-	C_WORKSPACE, /**< A workspace. */
-	C_CONTAINER, /**< A manually created container. */
-	C_VIEW,      /**< A view (aka window). */
+	C_ROOT,
+	C_OUTPUT,
+	C_WORKSPACE,
+	C_CONTAINER,
+	C_VIEW,
 
 	C_TYPES,
 };
 
-/**
- * Different ways to arrange a container.
- */
 enum swayc_layouts {
-	L_NONE,     /**< Used for containers that have no layout (views, root) */
+	L_NONE,
 	L_HORIZ,
 	L_VERT,
 	L_STACKED,
 	L_TABBED,
-	L_FLOATING, /**< A psuedo-container, removed from the tree, to hold floating windows */
-
-	/* Awesome/Monad style auto layouts */
-	L_AUTO_LEFT,
-	L_AUTO_RIGHT,
-	L_AUTO_TOP,
-	L_AUTO_BOTTOM,
-
-	L_AUTO_FIRST = L_AUTO_LEFT,
-	L_AUTO_LAST = L_AUTO_BOTTOM,
+	L_FLOATING,
 
 	// Keep last
 	L_LAYOUTS,
 };
 
 enum swayc_border_types {
-	B_NONE,   /**< No border */
-	B_PIXEL,  /**< 1px border */
-	B_NORMAL, /**< Normal border with title bar */
+	B_NONE,
+	B_PIXEL,
+	B_NORMAL,
 };
 
 struct sway_root;
 struct sway_output;
 struct sway_view;
 
-/**
- * Stores information about a container.
- *
- * The tree is made of these. Views are containers that cannot have children.
- */
 struct sway_container {
 	union {
 		// TODO: Encapsulate state for other node types as well like C_CONTAINER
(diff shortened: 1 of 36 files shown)


include/sway/tree/container.h
```diff
@@ -0,0 +1,143 @@
...
+ * it on this list.
+ */
+enum sway_container_type {
+	C_ROOT,
+	C_OUTPUT,
+	C_WORKSPACE,
+	C_CONTAINER,
+	C_VIEW,
+
+	C_TYPES,
```
emersion: Any reason why these aren't prefixed?

include/sway/tree/container.h
```diff
@@ -0,0 +1,143 @@
...
+	size_t id;
+
+	char *name;
+
+	enum sway_container_type type;
+	enum sway_container_layout layout;
+	enum sway_container_layout prev_layout;
+	enum sway_container_layout workspace_layout;
+
+    // TODO convert to layout coordinates
```
emersion: Indentation is weird here

include/sway/tree/container.h
```diff
@@ -0,0 +1,143 @@
...
+		struct sway_container *sibling, struct sway_view *sway_view);
+
+struct sway_container *container_output_destroy(struct sway_container *output);
+
+struct sway_container *container_view_destroy(struct sway_container *view);
+
+struct sway_container *container_set_layout(struct sway_container *container,
+		enum sway_container_layout layout);
+
+void container_descendents(struct sway_container *root,
```
emersion: Why are some of these prefixed with `container` and some others with `sway_container`?

include/sway/tree/container.h
SirCmpwn: They get typed a lot so it's more ergonomic. Since sway isn't a library I prefer it this way.

sway/tree/container.c
```diff
@@ -185,34 +151,38 @@ swayc_t *new_output(struct sway_output *sway_output) {
 	return output;
 }
 
-swayc_t *new_workspace(swayc_t *output, const char *name) {
-	if (!sway_assert(output, "new_workspace called with null output")) {
+struct sway_container *container_workspace_create(struct sway_container
+		*output, const char *name) {
+	if (!sway_assert(output,
+				"container_workspace_create called with null output")) {
```
emersion: Are you trying to save one character here?

sway/tree/container.c
```diff
@@ -185,34 +151,38 @@ swayc_t *new_output(struct sway_output *sway_output) {
...
-	if (!sway_assert(sibling, "new_view called with NULL sibling/parent")) {
+struct sway_container *container_view_create(struct sway_container *sibling,
+		struct sway_view *sway_view) {
+	if (!sway_assert(sibling,
+				"container_view_create called with NULL sibling/parent")) {
 		return NULL;
 	}
 	const char *title = view_get_title(sway_view);
-	swayc_t *swayc = new_swayc(C_VIEW);
+	struct sway_container *swayc = new_swayc(C_VIEW);
```
emersion: `new_swayc` should be `container_create` right?

sway/tree/container.c
```diff
@@ -294,7 +314,31 @@ swayc_t *swayc_parent_by_type(swayc_t *container, enum swayc_types type) {
 	return container;
 }
 
-swayc_t *swayc_at(swayc_t *parent, double lx, double ly,
+void sway_container_for_each(struct sway_container *container,
+		void (*f)(struct sway_container *view, void *data), void *data) {
+	if (container) {
```
emersion:
```c
if (container == NULL) {
   return;
}
```

emersion: Is this even needed?

acrisci: probably not

include/sway/tree/container.h
acrisci: oops

include/sway/tree/container.h
```diff
@@ -0,0 +1,137 @@
...
+	size_t id;
+
+	char *name;
+
+	enum sway_container_type type;
+	enum sway_container_layout layout;
+	enum sway_container_layout prev_layout;
+	enum sway_container_layout workspace_layout;
+
+	// TODO convert to layout coordinates
```
SirCmpwn: Can you get all of the output layout stuff in order next

sway/desktop/output.c
```diff
@@ -307,5 +307,5 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 	output->mode.notify = handle_output_mode;
 
 	arrange_layers(output);
-	arrange_windows(&root_container, -1, -1);
+	container_arrange_windows(&root_container, -1, -1);
```
SirCmpwn: -1 to renaming this to container_arrange_windows, I don't feel like arrange_windows is a logical child of containers

include/sway/tree/container.h
```diff
@@ -0,0 +1,137 @@
...
+ * Find a container at the given coordinates.
+ */
+struct sway_container *container_at(struct sway_container *parent,
+		double lx, double ly, struct wlr_surface **surface,
+		double *sx, double *sy);
+
+/**
+ * Apply the function for each child of the container breadth first.
+ */
+void container_for_each(struct sway_container *container,
```
SirCmpwn: `container_for_each_bfs -> container_for_each` and `container_map -> container_descendants` is a confusing change. The distinction between these two functions is important. It'd be better to do `container_map_descendants` and `container_map_descendants_bfs`.

sway/tree/container.c
```diff
@@ -185,34 +151,38 @@ swayc_t *new_output(struct sway_output *sway_output) {
 	return output;
 }
 
-swayc_t *new_workspace(swayc_t *output, const char *name) {
-	if (!sway_assert(output, "new_workspace called with null output")) {
+struct sway_container *container_workspace_create(
+		struct sway_container *output, const char *name) {
+	if (!sway_assert(output,
+				"container_workspace_create called with null output")) {
```
SirCmpwn: deindent this a tab

sway/tree/container.c
```diff
@@ -185,34 +151,38 @@ swayc_t *new_output(struct sway_output *sway_output) {
...
 	notify_new_container(workspace);
 	return workspace;
 }
 
-swayc_t *new_view(swayc_t *sibling, struct sway_view *sway_view) {
-	if (!sway_assert(sibling, "new_view called with NULL sibling/parent")) {
+struct sway_container *container_view_create(struct sway_container *sibling,
+		struct sway_view *sway_view) {
+	if (!sway_assert(sibling,
+				"container_view_create called with NULL sibling/parent")) {
```
SirCmpwn: ditto

sway/tree/layout.c
```diff
@@ -570,8 +591,9 @@ static swayc_t *get_swayc_in_direction_under(swayc_t *container,
 					}
 				}
 			} else {
-				wlr_log(L_DEBUG, "%s cont %d-%p dir %i sibling %d: %p", __func__,
-						idx, container, dir, desired, parent->children->items[desired]);
+				wlr_log(L_DEBUG,
+					"%s cont %d-%p dir %i sibling %d: %p", __func__, idx,
```
SirCmpwn: It occurs to me that __func__ here is not useful

sway/tree/view.c
```diff
@@ -45,6 +45,7 @@ void view_set_size(struct sway_view *view, int width, int height) {
 	}
 }
 
+// TODO make view coordinates
```
SirCmpwn: Do you mean layout coordinates?

include/sway/tree/container.h
acrisci: https://en.wikipedia.org/wiki/Map_(higher-order_function)

> In many programming languages, map is the name of a higher-order function that applies a given function to each element of a list, returning a list of results in the same order. 

This function is not "map".

sway/desktop/output.c
acrisci: The first and only pointer is a container and it affects the state of the tree of the container. If that's not a container function, then what does it take to be a container function?

include/sway/tree/container.h
SirCmpwn: Fair. Call them `container_descendants_for_each` and `container_descendants_for_each_bfs` then?

acrisci: what about `container_for_each_descendant`? sounds more like english.

SirCmpwn: +1

PR #1653, opened 2018-03-30
Revert "Refactor tree"

SirCmpwn:
Reverts swaywm/sway#1647


SirCmpwn:
Problems:

depth-first and breadth-first tree iteration were separate for a reason

It's spelt "descendant"

PR #1659, opened 2018-03-30
Add default_orientation command

SirCmpwn:


acrisci:
how to test?

sway/commands/default_orientation.c
```diff
@@ -0,0 +1,21 @@
...
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "default_orientation", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	if (strcasecmp(argv[0], "horizontal") == 0) {
+		config->default_orientation = L_HORIZ;
+	} else if (strcasecmp(argv[0], "vertical") == 0) {
+		config->default_orientation = L_VERT;
+	} else if (strcasecmp(argv[0], "auto") == 0) {
+		// Do nothing
```
acrisci: reset to `L_NONE`?

SirCmpwn:
`man 5 sway`

grep for default_orientation

????

emersion:
Please always provide a test plan!

include/sway/commands.h
```diff
@@ -125,7 +126,6 @@ sway_cmd cmd_move;
 sway_cmd cmd_new_float;
 sway_cmd cmd_new_window;
 sway_cmd cmd_no_focus;
-sway_cmd cmd_orientation;
```
emersion: Why this s/orientation/default_orientation/?

sway/tree/layout.c
```diff
@@ -123,12 +123,11 @@ struct sway_container *container_remove_child(struct sway_container *child) {
...
 enum sway_container_layout container_get_default_layout(
 		struct sway_container *output) {
-	/* TODO WLR
 	if (config->default_layout != L_NONE) {
-		//return config->default_layout;
+		return config->default_layout;
 	} else if (config->default_orientation != L_NONE) {
 		return config->default_orientation;
-	} else */if (output->width >= output->height) {
+	} else if (output->width >= output->height) {
```
emersion: Can't get this to work. Tried to resize the output so that height > width with `default_orientation auto`, it still opens new views on the right.

include/sway/commands.h
SirCmpwn: The command is named default_orientation, so it was poorly named in the first place.

sway/tree/layout.c
acrisci: is it supposed to update the orientation of the workspace when this becomes true on layout/mode changes?

Issue #1602, opened 2018-02-25
Pointer movements are always output-local


SirCmpwn:
Focused clients receive output-local pointer events even if the pointer is on another output

acrisci:
Yeah pointer events are not workspace aware yet.

PR #1660, opened 2018-03-30
Handle set_cursor requests from clients

emersion:
Allow clients to set a custom cursor if they have the seat's
pointer focus.


emersion:
>This sets the cursor even if the view is on an invisible workspace

Not my bug™. The current code will give seat pointer focus to these (and send surface enter/motion events).

Will send another PR to fix it.

acrisci:
:+1: good so far

SirCmpwn:
Relevant

https://github.com/swaywm/sway/issues/1602

emersion:
>This sets the cursor even if the view is on an invisible workspace

Fixed in #1669

SirCmpwn:
tyvm

PR #1665, opened 2018-03-30
Add lite damage tracking

emersion:
This skips the renderer if nothing has changed, and renders
everything otherwise.

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -113,4 +113,8 @@ void view_close(struct sway_view *view);
 
 void view_update_outputs(struct sway_view *view, const struct wlr_box *before);
 
+void view_damage_whole(struct sway_view *view);
+
+void view_damage_from(struct sway_view *view);
+
 #endif
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -15,6 +15,8 @@ struct sway_xdg_surface_v6 {
 	struct wl_listener request_move;
 	struct wl_listener request_resize;
 	struct wl_listener request_maximize;
+	struct wl_listener map;
+	struct wl_listener unmap;
 	struct wl_listener destroy;
 
 	int pending_width, pending_height;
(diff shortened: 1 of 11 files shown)


sway/desktop/output.c
```diff
@@ -195,82 +196,134 @@ static void render_layer(struct sway_output *output,
 	}
 }
 
-static void handle_output_frame(struct wl_listener *listener, void *data) {
-	struct sway_output *soutput = wl_container_of(listener, soutput, frame);
-	struct wlr_output *wlr_output = data;
+static void render_output(struct sway_output *output, struct timespec *when,
+		pixman_region32_t *damage) {
+	wlr_log(L_DEBUG, "render");
```
SirCmpwn: doop

sway/desktop/output.c
```diff
@@ -195,82 +196,134 @@ static void render_layer(struct sway_output *output,
...
 
-	wlr_output_make_current(wlr_output, NULL);
 	wlr_renderer_begin(renderer, wlr_output->width, wlr_output->height);
 
+	if (!pixman_region32_not_empty(damage)) {
+		// Output isn't damaged but needs buffer swap
+		goto renderer_end;
+	}
+
+	// TODO
```
SirCmpwn: TODO what?

SirCmpwn:
When testing this I just see a white screen

sway/desktop/output.c
acrisci: TODO: use vim Kappa

include/sway/tree/view.h
```diff
@@ -113,4 +113,8 @@ void view_close(struct sway_view *view);
 
 void view_update_outputs(struct sway_view *view, const struct wlr_box *before);
 
+void view_damage_whole(struct sway_view *view);
+
+void view_damage_from(struct sway_view *view);
```
acrisci: view damage from what?

emersion: I don't know how to name this one. In rootston it's `view_damage` but I don't like this name because it doesn't imply the damage is taken from the _view damage_ (the function doesn't damage the whole view). I'd like to call it something like `desktop_damage_from_view` but it isn't consistent with other names.

sway/desktop/output.c
SirCmpwn: doop x2

acrisci:
Closing the last view on a workspace doesn't clear the screen.

SirCmpwn:
>Closing the last view on a workspace doesn't clear the screen.

Cannot reproduce

emersion:
Eh, we're missing map/unmap listeners for xdg-shell-v6 views. Adding them now.

PR #1668, opened 2018-03-30
Basic split containers

acrisci:
Todo:

* [x] basic split containers
* [x] move focus around in the split tree
* [x] kill split containers
* [x] handle container destruction in the seat
* [x] arrange windows properly on container destruction
* [ ] implement organization plan
* [ ] flatten the tree correctly on tree structure changes
* [ ] don't destroy view containers until their surface goes away
* [ ] unify container destroy functions

Organization plan:

* confine all tree structure (attach/move/remove) changes to tree.h
* confine all layout changes (width/height/x/y) changes to layout.h
* confine all traversals, queries, create, and destroy to container.h
* move view and workspaces out of the tree folder

Known bugs: (will be fixed in the next effort)

* Some places in the code destroy the container immediately instead of waiting for the view to close.

Tests:

1. Open two windows and split the right one. Open a new window and you should see it split. Do that for `splith` and `splitv`.
2. Test `focus <direction>` descends into the focus stack. Also test `focus parent` and `focus child` works.
3. Split a container and try to kill it and sway should not crash.
4. destroy a split container with focus and focus should revert to the focus inactive of the first nonempty parent up to the workspace level
5. Open two windows and split the one on the right. Then close the window on the right. The windows should be arranged normally, with the left one taking up the whole screen.


sway/commands/kill.c
```diff
@@ -3,21 +3,30 @@
...
+		break;
+	case C_CONTAINER:
+		con = container_destroy(con);
+		con = container_reap_empty(con);
+		arrange_windows(con, -1, -1);
+		break;
+	case C_VIEW:
+		view_close(con->sway_view);
+		break;
+	default:
```
emersion: Remove this `default`

sway/input/cursor.c
```diff
@@ -20,9 +22,27 @@ static void cursor_update_position(struct sway_cursor *cursor) {
 	cursor->y = y;
 }
 
+static struct wlr_surface *layer_surface_at(struct sway_output *output,
```
emersion: This is from another PR right? Why does it show up here?

sway/commands/kill.c
emersion: Bump

acrisci:
Ready for review.

include/sway/tree/container.h
```diff
@@ -87,54 +87,77 @@ struct sway_container {
...
-// TODO only one container create function and pass the type?
-struct sway_container *container_output_create(
-		struct sway_output *sway_output);
+// TODO make private and use the container-specific create functions
+struct sway_container *container_create(enum sway_container_type type);
 
-struct sway_container *container_workspace_create(
-		struct sway_container *output, const char *name);
+/**
+ * Create a new container container. A container container can be a a child of
```
SirCmpwn: We should consider renaming containers to nodes

sway/commands/split.c
```diff
@@ -0,0 +1,107 @@
...
+	if (strcasecmp(argv[0], "v") == 0 || strcasecmp(argv[0], "vertical") == 0) {
+		_do_split(argc - 1, argv + 1, L_VERT);
+	} else if (strcasecmp(argv[0], "h") == 0 ||
+			strcasecmp(argv[0], "horizontal") == 0) {
+		_do_split(argc - 1, argv + 1, L_HORIZ);
+	} else if (strcasecmp(argv[0], "t") == 0 ||
+			strcasecmp(argv[0], "toggle") == 0) {
+		struct sway_container *focused =
+			config->handler_context.current_container;
+		if (focused->parent->layout == L_VERT) {
```
SirCmpwn: Hm, what if focused is a workspace?

sway/input/seat.c
```diff
@@ -35,30 +37,90 @@ void sway_seat_destroy(struct sway_seat *seat) {
...
 		void *data) {
 	struct sway_seat_container *seat_con =
 		wl_container_of(listener, seat_con, destroy);
 	struct sway_seat *seat = seat_con->seat;
 	struct sway_container *con = seat_con->container;
+	struct sway_container *parent = con->parent;
+	struct sway_container *focus = sway_seat_get_focus(seat);
 
-	bool is_focus = (sway_seat_get_focus(seat) == con);
+	// TODO handle workspace switch in the seat?
```
SirCmpwn: Eh

sway/input/seat.c
```diff
@@ -311,23 +373,7 @@ void sway_seat_set_focus_warp(struct sway_seat *seat,
...
-				wlr_seat_get_keyboard(seat->wlr_seat);
-			if (keyboard) {
-				wlr_seat_keyboard_notify_enter(seat->wlr_seat,
-						view->surface, keyboard->keycodes,
-						keyboard->num_keycodes, &keyboard->modifiers);
-			} else {
-				wlr_seat_keyboard_notify_enter(
-						seat->wlr_seat, view->surface, NULL, 0, NULL);
-			}
+			seat_send_focus(seat, container);
```
SirCmpwn: +1 good refactoring here

sway/tree/container.c
```diff
@@ -395,3 +405,17 @@ bool container_has_anscestor(struct sway_container *descendant,
 	}
 	return false;
 }
+
+bool find_child_func(struct sway_container *con, void *data) {
```
SirCmpwn: Can be static

sway/tree/layout.c
```diff
@@ -100,14 +101,16 @@ void container_add_child(struct sway_container *parent,
 			parent, parent->type, parent->width, parent->height);
 	list_add(parent->children, child);
 	child->parent = parent;
+	// TODO: set focus for this container?
+	sway_input_manager_set_focus(input_manager, child);
```
SirCmpwn: Will this always be correct? Example: start a process in a different workspace, pid tracking picks it up, switch away, window opens off-screen

Probably best to only set focus if the parent has focus?

sway/tree/layout.c
```diff
@@ -100,14 +101,16 @@ void container_add_child(struct sway_container *parent,
...
 }
 
 struct sway_container *container_reap_empty(struct sway_container *container) {
-	if (!sway_assert(container, "reaping null container")) {
```
SirCmpwn: Is this now okay? We should assert if it's a programming error to pass in a NULL container

sway/tree/layout.c
```diff
@@ -100,14 +101,16 @@ void container_add_child(struct sway_container *parent,
...
+	if (container == NULL) {
 		return NULL;
 	}
 	wlr_log(L_DEBUG, "reaping %p %s", container, container->name);
-	while (container != &root_container && container->children->length == 0) {
+	while (container->type != C_VIEW && container != &root_container && container->children->length == 0) {
```
SirCmpwn: 80 col limit

SirCmpwn:
```
set ruler
set cc=80
```

sway/tree/layout.c
```diff
@@ -619,30 +626,94 @@ static struct sway_container *get_swayc_in_direction_under(
 						wrap_candidate = parent->children->items[0];
 					}
 					if (config->force_focus_wrapping) {
-						return wrap_candidate;
+						 return sway_seat_get_focus_by_type(seat, wrap_candidate, C_VIEW);
```
SirCmpwn: Extra space slipped in here

sway/tree/layout.c
```diff
@@ -619,30 +626,94 @@ static struct sway_container *get_swayc_in_direction_under(
...
+	cont->width = child->width;
+	cont->height = child->height;
+	cont->x = child->x;
+	cont->y = child->y;
+
+	/* Container inherits all of workspaces children, layout and whatnot */
+	if (child->type == C_WORKSPACE) {
+		struct sway_container *workspace = child;
+		// reorder focus
+		int i;
```
SirCmpwn: Can move this into the loop

sway/tree/layout.c
```diff
@@ -619,30 +626,94 @@ static struct sway_container *get_swayc_in_direction_under(
...
+	wlr_log(L_DEBUG, "creating container %p around %p", cont, child);
+
+	cont->prev_layout = L_NONE;
+	cont->layout = layout;
+	cont->width = child->width;
+	cont->height = child->height;
+	cont->x = child->x;
+	cont->y = child->y;
+
+	/* Container inherits all of workspaces children, layout and whatnot */
```
SirCmpwn: Doesn't splitting a container just change the container layout? This is being handled at the command level iiuc

SirCmpwn: Probably better to handle here than at the command level though tbh

include/sway/tree/container.h
acrisci: The word `container` is used in the documentation and some commands. Renaming it will either break people's configs or split the name for it between the code and the user docs which would be confusing.

SirCmpwn: I think it would be wise to rename it internally, even if we don't update the docs.

acrisci: Ok let's do it in another pr.

sway/commands/split.c
SirCmpwn: Still have this question

sway/tree/layout.c
SirCmpwn: Can still move this into the loop

sway/tree/layout.c
acrisci: i3 is actually creating a container around the containers in the workspace in this case.

SirCmpwn: hm

acrisci:
Ok ready for another review.

sway/commands/split.c
```diff
@@ -0,0 +1,87 @@
...
+#include <strings.h>
+#include "sway/commands.h"
+#include "sway/tree/container.h"
+#include "sway/tree/layout.h"
+#include "sway/tree/view.h"
+#include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
+#include "log.h"
+
+static struct cmd_results *_do_split(int argc, char **argv, int layout) {
```
ascent12: What's with the leading underscore?

acrisci: I dunno this is kind of janky now that I look at it.

ascent12: Yeah, it's a static function, so you don't need to worry about any name/linker collisions.

I'll often do something similar to this when rewriting a function but reusing a name, just to get compiler errors for the old usages, and then remove it when I'm done.

PR #1669, opened 2018-03-30
Fix pointer events for hidden workspaces

emersion:
To test, add this at the end of `cursor_send_pointer_motion`:

```c
	static struct sway_container *old = NULL;
	if (old != swayc) {
		wlr_log(L_DEBUG, "focus change: %p", swayc);
	}
	old = swayc;
```

Test plan: open a window on a workspace, switch to an empty workspace. Pointer events shouldn't be sent to the window on the hidden workspace.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -20,7 +20,11 @@ static void cursor_update_position(struct sway_cursor *cursor) {
 	cursor->y = y;
 }
 
-static struct sway_container *cursor_at(struct sway_cursor *cursor,
+/**
+ * Returns the container at the cursor's position. If the container is a view,
+ * stores the surface at the cursor's position in `*surface`.
+ */
+static struct sway_container *container_at_cursor(struct sway_cursor *cursor,
 		struct wlr_surface **surface, double *sx, double *sy) {
 	// check for unmanaged views first
 	struct wl_list *unmanaged = &root_container.sway_root->unmanaged_views;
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Also segfaults: just clicking an empty workspace in general.

sway/input/cursor.c
```diff
@@ -19,16 +20,12 @@ static void cursor_update_position(struct sway_cursor *cursor) {
 	cursor->y = y;
 }
 
-static void cursor_send_pointer_motion(struct sway_cursor *cursor,
-		uint32_t time) {
-	struct wlr_seat *seat = cursor->seat->wlr_seat;
-	struct wlr_surface *surface = NULL;
-	double sx, sy;
-
+static struct sway_container *cursor_at(struct sway_cursor *cursor,
```
acrisci: The function is `cursor_at`, but it doesn't return a cursor. Should be `surface_at_cursor_position()` or something similar.

sway/input/cursor.c
```diff
@@ -39,19 +36,42 @@ static void cursor_send_pointer_motion(struct sway_cursor *cursor,
...
 			if (wlr_box_contains_point(&box, cursor->x, cursor->y)) {
-				surface = xsurface->surface;
-				sx = cursor->x - box.x;
-				sy = cursor->y - box.y;
-				wlr_seat_pointer_notify_enter(seat, surface, sx, sy);
-				wlr_seat_pointer_notify_motion(seat, time, sx, sy);
-				return;
+				*surface = xsurface->surface;
+				*sx = cursor->x - box.x;
+				*sy = cursor->y - box.y;
```
acrisci: If I recall correctly, xwayland surface x/y coordinates are already in layout coordinates. Is that correct?

sway/input/cursor.c
```diff
@@ -19,16 +20,12 @@ static void cursor_update_position(struct sway_cursor *cursor) {
...
-	double sx, sy;
-
+static struct sway_container *cursor_at(struct sway_cursor *cursor,
+		struct wlr_surface **surface, double *sx, double *sy) {
 	// check for unmanaged views first
+	struct wl_list *unmanaged = &root_container.sway_root->unmanaged_views;
 	struct sway_view *view;
-	wl_list_for_each_reverse(view, &root_container.sway_root->unmanaged_views,
-		unmanaged_view_link) {
+	wl_list_for_each_reverse(view, unmanaged, unmanaged_view_link) {
```
acrisci: I think the logic to find unmanaged views should be within the `container_at` function. They are on top of everything, so it should probably run no matter what the parent is.

sway/input/cursor.c
```diff
@@ -39,19 +36,42 @@ static void cursor_send_pointer_motion(struct sway_cursor *cursor,
...
+}
+
+static void cursor_send_pointer_motion(struct sway_cursor *cursor,
+		uint32_t time) {
+	struct wlr_seat *seat = cursor->seat->wlr_seat;
+	struct wlr_surface *surface = NULL;
+	double sx, sy;
+	struct sway_container *cont = cursor_at(cursor, &surface, &sx, &sy);
+
+	if (cont) {
```
acrisci: The way this will work (like it works in rootston), you need to handle the case where you get a container but no surface. In this case, the user has clicked on the decorations of the window. The surface should receive keyboard focus but not get any pointer events.

sway/input/cursor.c
emersion: Problem is, `container_at` takes a container to search views in. Unmanaged views are always outside it.

sway/input/cursor.c
emersion: Done

emersion:
Tested on DRM, works fine. Someone else: please test too!

SirCmpwn:
Works!

PR #1674, opened 2018-03-31
Pass pointer events to surface layers

SirCmpwn:


sway/input/cursor.c
```diff
@@ -20,9 +22,35 @@ static void cursor_update_position(struct sway_cursor *cursor) {
...
+		struct wlr_box box = {
+			.x = sway_layer->geo.x,
+			.y = sway_layer->geo.y,
+			.width = wlr_surface->current->width,
+			.height = wlr_surface->current->height,
+		};
+		// TODO: Test popups/subsurfaces
+		if (wlr_box_contains_point(&box, ox, oy) &&
+				pixman_region32_contains_point(
+					&wlr_surface->current->input, _sx, _sy, NULL)) {
```
emersion: We'll probably want to wrap this into a wlr function

SirCmpwn: meh

SirCmpwn: alright you convinced me

SirCmpwn:
Will switch to wlr_surface_point_accepts_input later so I can avoid some annoying rebasing

PR #1676, opened 2018-03-31
Swaybar input

### LINKED ISSUE #1104: Focus problem when clicking on a workspace on swaybar
When I click on a workspace with the mouse it looks like all the windows are focused and I have to click inside a window to get that one focused. If there is only one window on a workspace, for example a browser, it looks like it has the focus but the window (browser) doesn't get "activated", doesn't respond to any key. I have to switch to another desktop and back _with the keyboard_  to get a proper focus.

I thought it's something with my config, but I tried with the default config and I have the same problem with it. 

I'm using sway on Arch Linux, I tried with the stable version (0.11-2) and then upgraded to the git version (r2170.d4aa604-1), but I still have this problem.

### LINKED ISSUE #1457: The bottom pixel of sway-bar doesn't accept scrolling inputs
Title says it all, really weird.

My theory is that my mouse is allowed to go one pixel further down than its supposed to.

SirCmpwn:
Fixes #1104, fixes #1457


swaybar/bar.c
```diff
@@ -56,12 +57,156 @@ struct zwlr_layer_surface_v1_listener layer_surface_listener = {
...
+		uint32_t serial, struct wl_surface *surface) {
+	struct swaybar *bar = data;
+	bar->pointer.current = NULL;
+}
+
+static void wl_pointer_motion(void *data, struct wl_pointer *wl_pointer,
+		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
+	struct swaybar *bar = data;
+	bar->pointer.x = wl_fixed_to_int(surface_x);
+	bar->pointer.y = wl_fixed_to_int(surface_y);
```
emersion: Doesn't work with multiseat, but w/e

SirCmpwn: w/e indeed

SirCmpwn: We'll have to do a big refactoring project-wide to deal with multiseat

emersion: Might be cool to check that `pointer->current == wl_pointer` and do nothing if it isn't the case

PR #1684, opened 2018-03-31
Implement focus_follows_mouse, mouse_warping

SirCmpwn:
Superceeds #1680, #1683 

Test plan:

- `mouse_warping output` (default) should make the mouse follow the focused output.
- `mouse_warping none` should not warp the mouse at all.
- `focus_follows_mouse yes` (default) should move focus as the mouse moves over new views/containers
- `focus_follows_mouse no` should not move focus as the mouse moves over new views/containers
- Should not break when mouse is moved over popups, layers, etc

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -333,14 +333,28 @@ void sway_seat_set_focus(struct sway_seat *seat,
 	if (last_focus) {
 		struct sway_container *last_ws = last_focus;
 		if (last_ws && last_ws->type != C_WORKSPACE) {
-			last_ws = container_parent(last_focus, C_WORKSPACE);
+			last_ws = container_parent(last_ws, C_WORKSPACE);
 		}
 		if (last_ws) {
-			wlr_log(L_DEBUG, "sending workspace event");
 			ipc_event_workspace(last_ws, container, "focus");
 			if (last_ws->children->length == 0) {
 				container_workspace_destroy(last_ws);
 			}
+		}
+		struct sway_container *last_output = last_focus;
+		if (last_output && last_output->type != C_OUTPUT) {
+			last_output = container_parent(last_output, C_OUTPUT);
+		}
+		struct sway_container *new_output = container;
+		if (new_output && new_output->type != C_OUTPUT) {
+			new_output = container_parent(new_output, C_OUTPUT);
+		}
+		if (new_output != last_output && config->mouse_warping) {
+			struct wlr_output *output = new_output->sway_output->wlr_output;
+			// TODO: Change container coords to layout coords
+			double x = container->x + output->lx + container->width / 2.0;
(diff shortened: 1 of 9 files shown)


emersion:
For me `mouse_warping none` is the same as the default.

SirCmpwn:
derp

emersion:
On the X11/Wayland backend with `mouse_warping output` and `focus_follows_mouse no`, clicking on the other output moves the cursor in the middle of the output. If the cursor is already on the right output, we shouldn't move it in the middle (even if focus has changed).

sway/input/seat.c
```diff
@@ -291,8 +291,8 @@ void sway_seat_configure_xcursor(struct sway_seat *seat) {
 		seat->cursor->cursor->y);
 }
 
-void sway_seat_set_focus(struct sway_seat *seat,
-		struct sway_container *container) {
+void _sway_seat_set_focus(struct sway_seat *seat,
```
emersion: This should be `static`

emersion: Err, this is confusing. Maybe pick another name?

sway/input/seat.c
```diff
@@ -333,15 +333,31 @@ void sway_seat_set_focus(struct sway_seat *seat,
...
+		if (last_output && last_output->type != C_OUTPUT) {
+			last_output = container_parent(last_output, C_OUTPUT);
+		}
+		struct sway_container *new_output = container;
+		if (new_output && new_output->type != C_OUTPUT) {
+			new_output = container_parent(new_output, C_OUTPUT);
+		}
+		if (new_output && last_output && new_output != last_output
+				&& config->mouse_warping && warp) {
+			wlr_log(L_DEBUG, "warpin the mouse baby");
```
emersion: Nice debug line

sway/input/seat.c
SirCmpwn: nah

SirCmpwn:
Comments address, please review

sway/input/seat.c
```diff
@@ -351,12 +352,14 @@ void _sway_seat_set_focus(struct sway_seat *seat,
...
 		if (new_output && last_output && new_output != last_output
 				&& config->mouse_warping && warp) {
-			wlr_log(L_DEBUG, "warpin the mouse baby");
 			struct wlr_output *output = new_output->sway_output->wlr_output;
-			// TODO: Change container coords to layout coords
 			double x = container->x + output->lx + container->width / 2.0;
 			double y = container->y + output->ly + container->height / 2.0;
-			wlr_cursor_warp(seat->cursor->cursor, NULL, x, y);
+			if (!wlr_output_layout_contains_point(
+						root_container.sway_root->output_layout,
```
emersion: Extra space here

PR #1691, opened 2018-03-31
Views redesign

emersion:
Test plan: test various shells, test xwayland unmanaged views.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -70,7 +70,9 @@ struct sway_container {
 	enum sway_container_layout prev_layout;
 	enum sway_container_layout workspace_layout;
 
-	// in output-local coordinates
+	// For C_ROOT, this has no meaning
+	// For C_OUTPUT, this is the output position in layout coordinates
+	// For other types, this is the position in output-local coordinates
 	double x, y;
 	// does not include borders or gaps.
 	double width, height;
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -70,9 +70,8 @@ struct sway_container {
 	enum sway_container_layout prev_layout;
 	enum sway_container_layout workspace_layout;
 
-	// TODO convert to layout coordinates
+	// in output-local coordinates
 	double x, y;
-
 	// does not include borders or gaps.
 	double width, height;
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -39,6 +39,13 @@ struct sway_xwayland_surface {
 	int pending_width, pending_height;
 };
 
+struct sway_xwayland_unmanaged {
+	struct wlr_xwayland_surface *wlr_xwayland_surface;
+	struct wl_list link;
+
+	struct wl_listener destroy;
+};
+
 struct sway_wl_shell_surface {
 	struct sway_view *view;
 
@@ -126,9 +133,6 @@ void view_damage_from(struct sway_view *view);
 // view implementation
 
 void view_map(struct sway_view *view, struct wlr_surface *wlr_surface);
-
-void view_map_unmanaged(struct sway_view *view,
-	struct wlr_surface *wlr_surface);
 
 void view_unmap(struct sway_view *view);
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -66,10 +66,23 @@ enum sway_view_prop {
 	VIEW_PROP_INSTANCE,
 };
 
+struct sway_view_impl {
+	const char *(*get_prop)(struct sway_view *view,
+			enum sway_view_prop prop);
+	void (*set_size)(struct sway_view *view,
+			int width, int height);
+	void (*set_position)(struct sway_view *view,
+			double ox, double oy);
+	void (*set_activated)(struct sway_view *view, bool activated);
+	void (*close)(struct sway_view *view);
+};
+
 struct sway_view {
 	enum sway_view_type type;
-	struct sway_container *swayc;
-	struct wlr_surface *surface;
+	const struct sway_view_impl *impl;
+
+	struct sway_container *swayc; // NULL for unmanaged views
+	struct wlr_surface *surface; // NULL for unmapped views
 	int width, height;
 
 	union {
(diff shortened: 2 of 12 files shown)


emersion:
This PR is ready for review.

include/sway/tree/view.h
```diff
@@ -64,10 +66,23 @@ enum sway_view_prop {
 	VIEW_PROP_INSTANCE,
 };
 
+struct sway_view_impl {
+	const char *(*get_prop)(struct sway_view *view,
+			enum sway_view_prop prop);
+	void (*set_size)(struct sway_view *view,
+			int width, int height);
+	void (*set_position)(struct sway_view *view,
+			double ox, double oy);
```
acrisci: I think the main advantage of this is that now the output layout can change and we don't have to adjust all of the coordinates of all the containers in the tree.

sway/tree/view.c
```diff
@@ -6,94 +7,122 @@
...
+		.y = view->swayc->y,
+		.width = view->width,
+		.height = view->height,
+	};
+	struct wlr_output_layout_output *layout_output;
+	wl_list_for_each(layout_output, &output_layout->outputs, link) {
+		bool intersected = before != NULL && wlr_output_layout_intersects(
+			output_layout, layout_output->output, before);
+		bool intersects = wlr_output_layout_intersects(output_layout,
+			layout_output->output, &box);
```
acrisci: This function takes layout coordinates and you have given it output coordinates.

sway/tree/view.c
```diff
@@ -6,94 +7,122 @@
...
+	struct wlr_output_layout *output_layout =
+		root_container.sway_root->output_layout;
+	struct wlr_box box = {
+		.x = view->swayc->x,
+		.y = view->swayc->y,
+		.width = view->width,
+		.height = view->height,
+	};
+	struct wlr_output_layout_output *layout_output;
+	wl_list_for_each(layout_output, &output_layout->outputs, link) {
```
acrisci: A simplifying assumption you might be able to make here is that the main surface of a tiling window will never intersect with any output except the one it is on.

sway/tree/view.c
```diff
@@ -6,94 +7,122 @@
...
+		return view->impl->get_prop(view, VIEW_PROP_INSTANCE);
 	}
 	return NULL;
 }
 
+static void view_update_outputs(struct sway_view *view,
+		const struct wlr_box *before) {
+	struct wlr_output_layout *output_layout =
+		root_container.sway_root->output_layout;
+	struct wlr_box box = {
```
acrisci: Does this check need to be done for every surface that is contained in the view such as popups?

sway/tree/view.c
```diff
@@ -6,94 +7,122 @@
...
 			.height = view->height,
 		};
-		view->iface.set_size(view, width, height);
+		view->impl->set_size(view, width, height);
 		view_update_outputs(view, &box);
 	}
 }
 
 // TODO make view coordinates in layout coordinates
 void view_set_position(struct sway_view *view, double ox, double oy) {
```
acrisci: Did you consider to unify `set_size` and `set_position` and take a box?

sway/desktop/wl_shell.c
```diff
@@ -47,10 +47,6 @@ static void set_position(struct sway_view *view, double ox, double oy) {
 	view->swayc->y = oy;
 }
 
-static void set_activated(struct sway_view *view, bool activated) {
-	// no way to activate wl_shell
-}
-
 static void close(struct sway_view *view) {
```
acrisci: Can we not call these functions `close`? That name conflicts with another commonly used function.

include/sway/tree/container.h
```diff
@@ -70,9 +70,8 @@ struct sway_container {
 	enum sway_container_layout prev_layout;
 	enum sway_container_layout workspace_layout;
 
-	// TODO convert to layout coordinates
+	// in output-local coordinates
 	double x, y;
```
acrisci: change to `sx` and `sy` and you can remove the comment.

sway/desktop/output.c
```diff
@@ -378,6 +401,8 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 	output->mode.notify = handle_mode;
 	wl_signal_add(&wlr_output->events.transform, &output->transform);
 	output->transform.notify = handle_transform;
+	wl_signal_add(&wlr_output->events.scale, &output->scale);
```
acrisci: This signal is never removed.

sway/tree/view.c
emersion: Yeah, this function is broken anyway. I'll let it as is and fix it in the HiDPI PR.

include/sway/tree/view.h
emersion: Does it make sense to have a `double` for the view position?

include/sway/tree/container.h
emersion: Mhh, this isn't even correct since output containers have layout coordinates.

include/sway/tree/view.h
```diff
@@ -69,10 +69,8 @@ enum sway_view_prop {
 struct sway_view_impl {
 	const char *(*get_prop)(struct sway_view *view,
 			enum sway_view_prop prop);
-	void (*set_size)(struct sway_view *view,
-			int width, int height);
-	void (*set_position)(struct sway_view *view,
-			double ox, double oy);
+	void (*configure)(struct sway_view *view, double ox, double oy, int width,
+		int height);
```
SirCmpwn: I wonder if we should be passing layout coords to this

emersion: I think so. It's only used for Xwayland and floating, both need layout coords.

include/sway/tree/container.h
```diff
@@ -70,7 +70,9 @@ struct sway_container {
 	enum sway_container_layout prev_layout;
 	enum sway_container_layout workspace_layout;
 
-	// in output-local coordinates
+	// For C_ROOT, this has no meaning
+	// For C_OUTPUT, this is the output position in layout coordinates
+	// For other types, this is the position in output-local coordinates
```
SirCmpwn: How about using a wlr_box while you're at it

include/sway/tree/view.h
emersion: Since `configure` is also responsible for updating the container view, I'm not sure about this change yet

sway/desktop/xwayland.c
```diff
@@ -14,6 +14,33 @@
...
+
+	sway_surface->wlr_xwayland_surface = xsurface;
+
+	wl_signal_add(&xsurface->events.destroy, &sway_surface->destroy);
+	sway_surface->destroy.notify = unmanaged_handle_destroy;
+
+	wl_list_insert(&root_container.sway_root->xwayland_unmanaged,
+		&sway_surface->link);
+
+	// TODO: damage tracking
```
SirCmpwn: Do we need to at least damage the output whole here for this to work?

emersion:
One moar review plz

SirCmpwn:
I like it but let's use wlr_box

sway/desktop/xwayland.c
emersion: Actually we don't (yet) track damage for popups, subsurfaces and unmanaged views (which are all the same kind of surface). It works because we're lucky. Will definitely fix that in an upcoming PR.

include/sway/tree/container.h
emersion: No, because HiDPI.

emersion:
Fixed conflicts

acrisci:
:+1:

PR #1693, opened 2018-04-01
Partially implement move command

SirCmpwn:
Works:

- move [container|window] to workspace <name>
    - Note, this should be able to move C_CONTAINER but this is untested
- move [workspace] to output [left|right|up|down|<name>]

Not implemented yet:

- move [left|right|up|down]
- move scratchpad
- move position

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -84,8 +84,13 @@ struct sway_container {
 
 	struct {
 		struct wl_signal destroy;
+		// Raised after the tree updates, but before arrange_windows
+		// Passed the previous parent
+		struct wl_signal reparent;
 	} events;
 };
+
+const char *container_type_to_str(enum sway_container_type type);
 
 // TODO only one container create function and pass the type?
 struct sway_container *container_output_create(
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -1,0 +1,181 @@
+#include <string.h>
+#include <strings.h>
+#include <wlr/types/wlr_output.h>
+#include <wlr/types/wlr_output_layout.h>
+#include <wlr/util/log.h>
+#include "sway/commands.h"
+#include "sway/input/seat.h"
+#include "sway/output.h"
+#include "sway/tree/container.h"
+#include "sway/tree/layout.h"
+#include "sway/tree/workspace.h"
+#include "stringop.h"
+#include "list.h"
+
+static const char* expected_syntax = 
+	"Expected 'move <left|right|up|down> <[px] px>' or "
+	"'move <container|window> to workspace <name>' or "
+	"'move <container|window|workspace> to output <name|direction>' or "
+	"'move position mouse'";
+
+static struct sway_container *output_in_direction(const char *direction,
+		struct wlr_output *reference, int ref_ox, int ref_oy) {
+	int ref_lx = ref_ox + reference->lx,
+		ref_ly = ref_oy + reference->ly;
+	struct {
+		char *name;
+		enum wlr_direction direction;
+	} names[] = {
+		{ "up", WLR_DIRECTION_UP },
+		{ "down", WLR_DIRECTION_DOWN },
+		{ "left", WLR_DIRECTION_LEFT },
+		{ "right", WLR_DIRECTION_RIGHT },
+	};
+	for (size_t i = 0; i < sizeof(names) / sizeof(names[0]); ++i) {
+		if (strcasecmp(names[i].name, direction) == 0) {
+			struct wlr_output *adjacent = wlr_output_layout_adjacent_output(
+					root_container.sway_root->output_layout,
+					names[i].direction, reference, ref_lx, ref_ly);
+			if (adjacent) {
+				struct sway_output *sway_output = adjacent->data;
+				return sway_output->swayc;
+			}
+			break;
+		}
+	}
+	return output_by_name(direction);
+}
+
+static struct cmd_results *cmd_move_container(struct sway_container *current,
+		int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "move container/window",
+				EXPECTED_AT_LEAST, 4))) {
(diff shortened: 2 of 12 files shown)


sway/commands/move.c
```diff
@@ -0,0 +1,181 @@
...
+			num_name = argv[4];
+			ws = workspace_by_number(num_name);
+		} else {
+			ws_name = join_args(argv + 3, argc - 3);
+			ws = workspace_by_name(ws_name);
+		}
+		if (ws == NULL) {
+			ws = workspace_create(ws_name ? ws_name : num_name);
+		}
+		if (ws_name) {
```
acrisci: no need for the null check

include/sway/tree/container.h
```diff
@@ -84,9 +84,14 @@ struct sway_container {
 
 	struct {
 		struct wl_signal destroy;
+		// Raised after the tree updates, but before arrange_windows
+		// Passed the previous parent
+		struct wl_signal reparent;
```
acrisci: nothing seems to listen to this.

SirCmpwn: I'll be adding something in a coming PR

sway/tree/layout.c
```diff
@@ -138,22 +140,46 @@ struct sway_container *container_remove_child(struct sway_container *child) {
...
+	if (destination->type == C_VIEW) {
+		new_parent = container_add_sibling(destination, container);
+	} else {
+		new_parent = destination;
+		container_add_child(destination, container);
+	}
+	wl_signal_emit(&container->events.reparent, old_parent);
+	if (container->type == C_WORKSPACE) {
+		struct sway_seat *seat = sway_input_manager_get_default_seat(
+				input_manager);
```
acrisci: handler context has a seat.

sway/commands/move.c
```diff
@@ -0,0 +1,181 @@
...
+		{ "up", WLR_DIRECTION_UP },
+		{ "down", WLR_DIRECTION_DOWN },
+		{ "left", WLR_DIRECTION_LEFT },
+		{ "right", WLR_DIRECTION_RIGHT },
+	};
+	for (size_t i = 0; i < sizeof(names) / sizeof(names[0]); ++i) {
+		if (strcasecmp(names[i].name, direction) == 0) {
+			struct wlr_output *adjacent = wlr_output_layout_adjacent_output(
+					root_container.sway_root->output_layout,
+					names[i].direction, reference, ref_ox, ref_oy);
```
acrisci: can confirm it takes layout coords.

sway/commands/move.c
```diff
@@ -0,0 +1,181 @@
...
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "move workspace", EXPECTED_EQUAL_TO, 4))) {
+		return error;
+	} else if (strcasecmp(argv[1], "to") != 0
+			|| strcasecmp(argv[2], "output") != 0) {
+		return cmd_results_new(CMD_INVALID, "move", expected_syntax);
+	}
+	struct sway_container *source = container_parent(current, C_OUTPUT);
+	struct sway_container *destination = output_in_direction(argv[3],
+			source->sway_output->wlr_output, current->x, current->y);
```
acrisci: it's probably better to use the center point of the view for the reference

sway/tree/layout.c
SirCmpwn: I don't want to bring the seat into this level of the code (for the time being)

sway/commands/move.c
```diff
@@ -0,0 +1,181 @@
...
+#include "sway/input/seat.h"
+#include "sway/output.h"
+#include "sway/tree/container.h"
+#include "sway/tree/layout.h"
+#include "sway/tree/workspace.h"
+#include "stringop.h"
+#include "list.h"
+
+static const char* expected_syntax = 
+	"Expected 'move <left|right|up|down|next|prev|first> <[px] px>' or "
```
emersion: `next`, `prev` and `first` seem to have been removed

SirCmpwn: Good catch

sway/commands/move.c
```diff
@@ -0,0 +1,181 @@
...
+			return cmd_results_new(CMD_FAILURE, "move workspace",
+				"Can't find output with name/direction '%s'", argv[3]);
+		}
+		struct sway_container *focus = sway_seat_get_focus_inactive(
+				config->handler_context.seat, destination);
+		if (!focus) {
+			// We've never been to this output before
+			focus = destination->children->items[0];
+		}
+		container_move_to(current, focus);
```
emersion: `sway_seat_set_focus`?

emersion: Maybe not. In which case, should probably remove line 84.

sway/commands/move.c
```diff
@@ -0,0 +1,181 @@
...
+	if ((error = checkarg(argc, "move", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	struct sway_container *current = config->handler_context.current_container;
+
+	if (argc == 2 || (argc == 3 && strcasecmp(argv[2], "px") == 0)) {
+		char *inv;
+		move_amt = (int)strtol(argv[1], &inv, 10);
+		if (*inv != '\0' && strcasecmp(inv, "px") != 0) {
+			move_amt = 10;
```
emersion: It's strange to move by 10px when the argument is invalid.

sway/commands/move.c
SirCmpwn: No, I think we need to do this here. i3 doesn't bring focus along for the ride.

SirCmpwn: Though I have always wanted an option to make moves follow, cause it's nice for moving a bunch of things at once

emersion:
A weird thing happen when you:
* Open something on workspace 1
* Move it to workspace 5
* There's nothing left on workspace 1, focus right, that deletes workspace 1
* View on workspace 5 doesn't have focus

PR #1700, opened 2018-04-02
[WIP] Implement move [left|right|up|down]

SirCmpwn:


SirCmpwn:
This is ready for review, if anyone is brave enough.

acrisci:
This notation is very hard to read, but I will do my best to parse it and make test cases and compare with the i3 behavior.

SirCmpwn:
Summary of notation follows

`(<type> <children...>)`

This is nested indefinitely

acrisci:
[https://github.com/swaywm/sway-tests/blob/master/test/test_move.py](tests)

acrisci:
[test_move_into_split](https://github.com/swaywm/sway-tests/blob/master/test/test_move.py#L64) fails.

SirCmpwn:
Should it be moved to the end or adjacent to the inactive focus?

acrisci:
It goes after the focus-inactive view.

SirCmpwn:
Latest commit fixes @acrisci's issue and also addresses another test case:

```
(root
  (output
    (workspace L_HORIZ
      (view a focus)
      (container
        (view b)
        (view c)))))
-> move left
(root
  (output
    (workspace L_HORIZ
      (view a focus)
      (container
        (view b)
        (view c)))))
# no changes

(root
  (output
    (workspace L_HORIZ
      (container
        (view b)
        (view c))
      (view a focus))))
-> move right
(root
  (output
    (workspace
      (container
        (view b)
        (view c))
      (view a focus))))
# no changes
```

In plain english, set up this layout:

```
(root
  (output
    (workspace L_HORIZ
      (view a focus)
      (container (view b) (view c)))))
```

Then move (view a) to the right, then again (it should now be to the right of the first container). Keep moving it right and the layout should not change. Move it left and it should again join the middle container, then move left again and it should be back to the original layout. Keep moving left and the layout should not change.

sway/tree/layout.c
```diff
@@ -166,9 +172,183 @@ void container_move_to(struct sway_container *container,
...
+		return false;
+	}
+}
+
 void container_move(struct sway_container *container,
 		enum movement_direction dir, int move_amt) {
-	// TODO
+	int offs = dir == MOVE_LEFT || dir == MOVE_UP ? -1 : 1;
+	// TODO: Accomodate for fullscreen view
+	int i;
```
emersion: No need to declare `i` outside the loop

sway/tree/layout.c
```diff
@@ -166,9 +172,183 @@ void container_move_to(struct sway_container *container,
...
+		return dir == MOVE_UP || dir == MOVE_DOWN;
+	default:
+		return false;
+	}
+}
+
 void container_move(struct sway_container *container,
 		enum movement_direction dir, int move_amt) {
-	// TODO
+	int offs = dir == MOVE_LEFT || dir == MOVE_UP ? -1 : 1;
```
emersion: Should assert that `container` is not the root container.

sway/tree/layout.c
```diff
@@ -166,9 +172,183 @@ void container_move_to(struct sway_container *container,
...
+		if (target->parent == container->parent) {
+			wlr_log(L_DEBUG, "Swapping siblings");
+			target->parent->children->items[i + offs] = container;
+			target->parent->children->items[i] = target;
+			arrange_windows(target->parent, -1, -1);
+		} else {
+			wlr_log(L_DEBUG, "Reparenting container");
+			container_remove_child(container);
+			container_add_sibling(target, container);
+			wl_signal_emit(&container->events.reparent, parent);
```
emersion: Maybe this signal should be triggered by the function setting `container->parent` (here `container_add_sibling`)

sway/tree/layout.c
```diff
@@ -166,9 +172,183 @@ void container_move_to(struct sway_container *container,
...
+		}
+		int dest_i = offs > 0 ? target->children->length : 0;
+		wlr_log(L_DEBUG, "%d %d", offs, dest_i);
+		container_insert_child(target, container, dest_i);
+		container_set_layout(target,
+				dir == MOVE_LEFT || dir == MOVE_RIGHT ?  L_HORIZ : L_VERT);
+		arrange_windows(target, -1, -1);
+		break;
+	}
+	default:
```
emersion: When does this exactly happen?

sway/tree/layout.c
SirCmpwn: wrong

emersion: Ugh, my bad

sway/tree/layout.c
```diff
@@ -166,9 +188,186 @@ void container_move_to(struct sway_container *container,
...
+		break;
+	}
+	case C_CONTAINER:
+		if (target->children->length == 0) {
+			container_move_to(container, target);
+			arrange_windows(target->parent, -1, -1);
+			arrange_windows(parent, -1, -1);
+		} else if (!is_perpendicular(target->layout, dir)) {
+			// Insert in nearest side in direction of movement
+			int dest_i = offs < 0 ? target->children->length - 1 : 0;
```
emersion: I think this should be `target->children->length` instead of `target->children->length - 1`, just like at line 334.

SirCmpwn: ack, this was causing a bug

acrisci:
You aren't supposed to flatten when you move out of a split. [test](https://github.com/swaywm/sway-tests/blob/c7a5baaf94edaf9faa0f1234196760725a7818be/test/test_move.py#L99)

sway/tree/layout.c
```diff
@@ -166,9 +188,186 @@ void container_move_to(struct sway_container *container,
...
+		int dest_i = offs > 0 ? target->children->length : 0;
+		wlr_log(L_DEBUG, "%d %d", offs, dest_i);
+		container_insert_child(target, container, dest_i);
+		container_set_layout(target,
+				dir == MOVE_LEFT || dir == MOVE_RIGHT ?  L_HORIZ : L_VERT);
+		arrange_windows(target, -1, -1);
+		break;
+	}
+	case C_CONTAINER:
+		if (target->children->length == 0) {
```
emersion: I don't think this special case is required

emersion: Hmm, it is for the focus-inactive stuff.

SirCmpwn:
>You aren't supposed to flatten when you move out of a split. test

Tangental: please make your s-expression syntax consistent, you use `view1` and `(view 1)` and `(view a)` in different places. Ideally just use the syntax I originally wrote in.

sway/tree/layout.c
SirCmpwn: We can fix it by making that code a bit better... that focus_get_inactive can return NULL is a tough cookie to think about

acrisci:
> Tangental: please make your s-expression syntax consistent

I simplify parens when it is not ambiguous like in math, (1) + (1) == 1 + 1.

SirCmpwn:
That makes it harder to read.

SirCmpwn:
Another test case:

```
(workspace
  (view a)
  (container
    (container
      (container
        (container
          (view b))))))
-> move left
(workspace
  (view a)
  (view b))
-> move left
(workspace
  (view b)
  (view a))
```

orestisf1993:
H is horizontal, V is vertical, lower case letters are windows/leafs

with `H [ a V[ b c ] ]`:
focus a, move right, move right => new layout: `H[ V[ b c ] a ]` (correct).
`move left` => layout doesn't change.

SirCmpwn:
>H is horizontal, V is vertical, lower case letters are windows/leafs
>
>with `H [ a V[ b c ] ]`:
>focus a, move right, move right => new layout: `H[ V[ b c ] a ]` (correct).
>`move left` => layout doesn't change.

afaict this works

orestisf1993:
This is what I get: https://gfycat.com/FarflungSizzlingEmperorpenguin

sorry if I am doing something wrong :/

SirCmpwn:
Pull and try again, please, but don't expect much else to work. There are known problems with this version.

orestisf1993:
Yeah it works now. I can test it later when you have everything sorted out if you want.

SirCmpwn:
I ~hope~ think this is all working now.

orestisf1993:
`H [ a V [ b c ] ] ` c focused more recently than b (will write as c > b from now on).

move a right, a shouldn't be placed to the top of the vertical container.

SirCmpwn:
Fuck, that breaks something else

orestisf1993:
1. open window
2. splitv
3. open window
4. move down

expected outcome: no change

actual outcome: the second window moves and gets a reduced size

SirCmpwn:
This is ready for another pass

@stapelberg, you're one crazy dude

orestisf1993:
A dissimilarity with i3:
`H[a V[b c]]`, focus c, move up, move up, move down.
Final i3 layout: `H[a V[c b]]`
Final sway layout: `H[a V[b c]]`.

A bug:
` H[V[a b c] d]`, focus a, move up, move right, move left.
Final i3 layout: `H[V[b c] a d]`
Final sway layout: `H[V[b c a] d]`

swaybar/ipc.c
```diff
@@ -331,7 +331,7 @@ bool handle_ipc_readable(struct swaybar *bar) {
 	switch (resp->type) {
 	case IPC_EVENT_WORKSPACE:
 		ipc_get_workspaces(bar);
-		break;
+		return true;
```
emersion: `true`?

SirCmpwn: true -> rerender the bar

sway/tree/layout.c
```diff
@@ -100,13 +100,31 @@ static int index_child(const struct sway_container *child) {
 	return i;
 }
 
+void container_insert_child(struct sway_container *parent,
+		struct sway_container *child, int i) {
+	struct sway_container *old_parent = NULL;
```
emersion: `old_parent` is always NULL

sway/tree/layout.c
```diff
@@ -100,13 +100,31 @@ static int index_child(const struct sway_container *child) {
...
+	if (active->parent) {
+		old_parent = active->parent;
+		container_remove_child(active);
+	}
 	struct sway_container *parent = fixed->parent;
 	int i = index_child(fixed);
 	list_insert(parent->children, i + 1, active);
 	active->parent = parent;
+	if (old_parent && old_parent != parent) {
```
emersion: I'd imagine `reparent` to be triggered even if the old parent was NULL

sway/tree/layout.c
```diff
@@ -166,9 +184,264 @@ void container_move_to(struct sway_container *container,
...
+		return MOVE_LEFT;
+	}
+}
+
+/* Gets the index of the most extreme member based on the movement offset */
+static int container_limit(struct sway_container *container, int offs) {
+	if (container->children->length == 0) {
+		return 0;
+	}
+	return offs < 0 ? 0 : container->children->length - 1;
```
emersion: Inserting an item at index `container->children->length - 1` should insert before the last item?

emersion: I mean, it should be possible to hit `container->children->length` without selecting the parent

swaybar/ipc.c
emersion: Ah, makes sense

acrisci:
@orestisf1993 thanks for checking on this. You are welcome to write some tests if you want to make sure these cases will work in the future. Look at our [sway-tests](https://github.com/swaywm/sway-tests) repository. I'm on `#sway-devel` as acrisci on freenode if you run into issues.

sway/tree/layout.c
SirCmpwn: That's not what this function is for, read the comment.

SirCmpwn:
Thanks @orestisf1993

SirCmpwn:
>A bug:
>H[V[a b c] d], focus a, move up, move right, move left.
>Final i3 layout: H[V[b c] a d]
>Final sway layout: H[V[b c a] d]

I think the sway behavior here is more intuitive, I'm going to keep it.

acrisci:
I made a crappy curses [debug console](https://github.com/acrisci/i3ipc-python/blob/master/examples/i3-debug-console.py) script to see the layout. It won't completely work on sway until we get the binding event because it won't refresh when the layout changes otherwise though.

SirCmpwn:
Yeah, it's a little weird, but it's based on the semantics of focus-inactive and follows more consistent rules. Let me ruminate on it but my inclination is to be okay with the departure from i3 in this case rather than add endless edge cases.

SirCmpwn:
So I figured out exactly why @orestisf1993's problem was happening and came to a different understanding of the workflow. I agree that it's a bug now, pushed a fix.

emersion:
```
../sway/debug-tree.c:95:6: error: variable 'focus' is used uninitialized whenever 'if' condition is false [-Werror,-Wsometimes-uninitialized]
        if (seat != NULL) {
            ^~~~~~~~~~~~
../sway/debug-tree.c:98:41: note: uninitialized use occurs here
        draw_container(cairo, &root_container, focus, 0, 0);
                                               ^~~~~
../sway/debug-tree.c:95:2: note: remove the 'if' if its condition is always true
        if (seat != NULL) {
        ^~~~~~~~~~~~~~~~~~
../sway/debug-tree.c:94:30: note: initialize the variable 'focus' to silence this warning
        struct sway_container *focus;
                                    ^
                                     = NULL
```

sway/debug-tree.c
```diff
@@ -0,0 +1,113 @@
...
+	cairo_surface_flush(surface);
+	struct wlr_renderer *renderer = wlr_backend_get_renderer(server.backend);
+	if (root_container.sway_root->debug_tree) {
+		wlr_texture_destroy(root_container.sway_root->debug_tree);
+	}
+	unsigned char *data = cairo_image_surface_get_data(surface);
+	int stride = cairo_format_stride_for_width(CAIRO_FORMAT_ARGB32, width);
+	struct wlr_texture *texture = wlr_texture_from_pixels(renderer,
+		WL_SHM_FORMAT_ARGB8888, stride, width, height, data);
+	root_container.sway_root->debug_tree = texture;
```
emersion: Damage all outputs here?

sway/tree/layout.c
```diff
@@ -166,9 +185,284 @@ void container_move_to(struct sway_container *container,
...
+		int index = index_child(current);
+		int limit = container_limit(parent, move_dir);
+
+		switch (current->type) {
+		case C_OUTPUT: {
+			enum wlr_direction wlr_dir;
+			sway_dir_to_wlr(move_dir, &wlr_dir);
+			double ref_x = current->x + current->width / 2;
+			double ref_y = current->y + current->height / 2;
+			ref_x += current->sway_output->wlr_output->lx;
```
emersion: Prefer to use `current->sway_output->swayc->x` (in the future, we might want to send a fake position to clients via `wlr_output`).

sway/tree/layout.c
```diff
@@ -166,9 +185,284 @@ void container_move_to(struct sway_container *container,
...
+/* Gets the index of the most extreme member based on the movement offset */
+static int container_limit(struct sway_container *container,
+		enum movement_direction move_dir) {
+	if (container->children->length == 0) {
+		return 0;
+	}
+	return move_offs(move_dir) < 0 ? 0 : container->children->length - 1;
+}
+
+static void workspace_rejigger(struct sway_container *ws,
```
emersion: I'd like a little doc comment explaining what this function does, it's not immediately clear to me.

sway/debug-tree.c
SirCmpwn: nup

emersion:
With two outputs:
1. Open a window on output 1
2. Move it to output 2
3. Move it back to output 1
4. Move up, an empty container gets created
5. Move around, segfault

```
#0  0x0000557cd583bb20 in container_move (container=0x557cd828d940, move_dir=MOVE_DOWN, move_amt=10)
    at ../sway/tree/layout.c:423
423						while (focus_inactive->parent != sibling) {
(gdb) bt
#0  0x0000557cd583bb20 in container_move (container=0x557cd828d940, move_dir=MOVE_DOWN, move_amt=10)
    at ../sway/tree/layout.c:423
#1  0x0000557cd5831111 in cmd_move (argc=1, argv=0x557cd828dc78) at ../sway/commands/move.c:170
#2  0x0000557cd581a7e8 in execute_command (_exec=0x557cd7fdf7d0 "move down", seat=0x557cd7fd3b30)
    at ../sway/commands.c:355
#3  0x0000557cd582c1cd in keyboard_execute_command (keyboard=0x557cd7ffb890, binding=0x557cd7fdf8a0)
    at ../sway/input/keyboard.c:98
#4  0x0000557cd582bb2e in keyboard_execute_bindsym (keyboard=0x557cd7ffb890, pressed_keysyms=0x557cd7ffb8d0, modifiers=9, key_state=WLR_KEY_PRESSED) at ../sway/input/keyboard.c:166
#5  0x0000557cd582b6b7 in handle_keyboard_key (listener=0x557cd7ffb8a0, data=0x7ffd4fdd8b98)
    at ../sway/input/keyboard.c:354
#6  0x00007f368a70fabc in wlr_signal_emit_safe (signal=0x557cd7852e40, data=0x7ffd4fdd8b98)
    at ../subprojects/wlroots/util/signal.c:29
#7  0x00007f368a6f0b24 in wlr_keyboard_notify_key (keyboard=0x557cd7852d48, event=0x7ffd4fdd8b98)
    at ../subprojects/wlroots/types/wlr_keyboard.c:136
#8  0x00007f368a6e305b in x11_handle_input_event (x11=0x557cd7852cd0, event=0x557cd820f8a0)
    at ../subprojects/wlroots/backend/x11/input_device.c:74
#9  0x00007f368a6e2e7d in handle_x11_event (x11=0x557cd7852cd0, event=0x557cd820f8a0)
    at ../subprojects/wlroots/backend/x11/backend.c:70
#10 0x00007f368a6e25a9 in x11_event (fd=4, mask=1, data=0x557cd7852cd0)
    at ../subprojects/wlroots/backend/x11/backend.c:120
#11 0x00007f3688c0cfe2 in wl_event_loop_dispatch (loop=0x557cd7849f10, timeout=<optimized out>)
    at src/event-loop.c:641
#12 0x00007f3688c0affd in wl_display_run (display=0x557cd784c700) at src/wayland-server.c:1260
#13 0x0000557cd5819cc4 in server_run (server=0x557cd5a521d0 <server>) at ../sway/server.c:134
#14 0x0000557cd5818ec2 in main (argc=2, argv=0x7ffd4fdd9028) at ../sway/main.c:418
```

orestisf1993:
Where the container is placed seems to depend on the direction of movement. The following fails:
`H[a V[b c]` move a right => `V[b a c]`
`H[V[b c] a]` move a left => `V[b c a]`
```
def test_placed_after_focused_inactive(sway):
    '''
    (view1 focus) (container L_VERT (view2) (view3 focus-inactive))
    -> move right
    (container L_VERT (view2) (view3) (view1 focus))
    '''
    view1 = sway.open_window('view1')
    view2 = sway.open_window('view2')
    view2.command('splitv')
    view3 = sway.open_window('view3')

    view1.focus()
    sway.ipc.command('move right')

    tree = sway.ipc.get_tree()
    ws = tree.workspaces()[0]
    assert len(ws.nodes) == 1
    vsplit = ws.nodes[0]
    assert vsplit.nodes[0].id == view2.id
    assert vsplit.nodes[1].id == view3.id
    assert vsplit.nodes[2].id == view1.id
    assert vsplit.nodes[2].focused
```

emersion:
Open two X11 backend outputs, focus X11-2, open two windows, repeat this sequence:
* Move down
* Move right

This creates lots of nested containers.

screenshot from 2018-04-06 16-30-45

SirCmpwn:
>Where the container is placed seems to depend on the direction of movement.

Have decided not to care

PR #1701, opened 2018-04-02
protocols: remove xdg-shell XML file

emersion:


SirCmpwn:
Can you also remove:

* desktop-shell.xml
* gamma-control.xml
* server-decoration.xml
* swaylock.xml

SirCmpwn:
Thanks!

PR #1703, opened 2018-04-03
Give keyboard input to layer surfaces that ask for it

SirCmpwn:
Test plan: same as this guy https://github.com/swaywm/wlroots/pull/804


SirCmpwn:
:tada:

Issue #1506, opened 2017-12-08
Swaylock crashes


cedrichaase:
Swaylock crashes whenever I type in just over 40 random characters into the "password prompt".

I pasted the stacktrace and output of `sway -d` here:
https://gist.github.com/cedrichaase/fd3edfa8cd728630ac5adc7b34a2278f

I'm running sway/swaylock version 0.15.0.
If there's any more information I can provide, please let me know.

cedrichaase:
Push.
I might have failed to convey this, but this is a blatant security issue. If anybody can unlock my screen by typing in random stuff for a while, it makes swaylock rather pointless.

As indicated by the stack trace, the error seems to occur during the process of rendering the "typing indicator". I've toyed around with the code a little, but seeing as I have very limited experience with C and no experience with pixman/cairo, the best I could come up with was disabling rendering of said indicator altogether. This stops the error from occuring, but yeah, disabling a feature is not the same as fixing it :)

So, can anybody reproduce this?

SirCmpwn:
Try the latest 0.15 branch.

cedrichaase:
I had to keep hammering away at my keyboard a little longer this time, but swaylock still crashes for me at `c08c805` (latest commit in 0.15 at the time of writing). I added the stack trace for this version to the gist mentioned in the original post. It does look slightly different than before, although the error still seems to occur in `cairo_paint`.

martinetd:
Sounds like a duplicate of #918, do you have at least two with one scaled and one not?
Curious if that would happen with wlroots as well, should be able to test pretty soon...

cedrichaase:
Yeah, sounds like the same issue all right. Sorry, guess I overlooked that.
Indeed, I'm using a 13" MacBook Pro Retina (scaled) with a 1080p (non-scaled) display configured:
```
output eDP-1    scale 2     pos 320 1080
output DP-1     scale 1     pos 0   0
output DP-2     scale 1     pos 0   0
```
Though, while testing, the external (non-scaled) monitor was not connected.

SirCmpwn:
Your stack traces are not terribly useful. Compile sway and wlc with debug symbols enabled.

cedrichaase:
Yeah, I thought I might have to do that. Unfortunately, I have no idea how. Can you explain or point me to some documentation?

Issue #1141, opened 2017-03-30
workspace_layout tabbed breaks screen locking


wdbw:
With `workspace_layout tabbed` set in the configuration file and at least one opened window I am not able to properly lock the screen. After running `swaylock` the screen content is frozen in place, and therefore still visible (but not updated).

okeri:
same for me

Issue #1072, opened 2017-02-07
swaylock “wrong” password only displayed on one monitor


maximiliankaul:
The red “wrong” message is only displayed on one of my monitors when using swaylock and entering a wrong password.
The blue “validating” works fine on both monitors.
I can “fix” the problem by calling `wl_dispatch_events();` in `sway/swaylock/main.c`:
```
void notify_key(enum wl_keyboard_key_state state, xkb_keysym_t sym, uint32_t code, uint32_t codepoint) {
    int redraw_screen = 0;                                                         
    char *password_realloc;                                                        
                                                                                   
    if (state == WL_KEYBOARD_KEY_STATE_PRESSED) {                                  
        switch (sym) {                                                             
        case XKB_KEY_KP_Enter: // fallthrough                                      
        case XKB_KEY_Return:                                                       
            render_data.auth_state = AUTH_STATE_VALIDATING;                        
                                                                                   
            render(&render_data);                                                  
            // Make sure our render call will actually be displayed on the screen
            wl_dispatch_events();                                                  
                                                                                   
            if (verify_password()) {                                               
                exit(0);                                                           
            }                                                                      
            render_data.auth_state = AUTH_STATE_INVALID;                           
            redraw_screen = 1;                                                     
            wl_dispatch_events(); // -----> FIXES “wrong” message only displayed on one monitor <-----                                                  
                                                                                   
            password_size = 1024;                                                  
            password = malloc(password_size);                                      
            password[0] = '\0';                                                    
            break;                                                                 
```
I don’t understand why this is necessary (`redraw_screen = 1;` should take care of this).
Unfortunately I neither understand the underlying problem nor have any experience with wayland so I might be doing something terribly wrong. Thus I’m not sending a PR.
I’m calling swaylock like this: `swaylock -i /path/to/image.png -s stretch`

fluxchief:
I worked on this some time ago and I could not come up with a working solution. Having multiple wl_dispatch_events() seems to fix it sometimes but also causes some other weird behavior (application locks up) - even if it would work, this really should not be necessary.

Just for the record: I tried manually checking if there are events in the pipeline and flushing them but it still didn't behave correctly.

Issue #918, opened 2016-09-28
Swaylock Backdoor


kleinreact:
Rapidly typing arbitrary letters during swaylock gives me a `Bus error` and exits swaylock. 

The behavior is always reproducable and it never took longer than about 10 seconds to produce it.

(btw, you should change the error message: https://www.youtube.com/watch?v=W9P69WigjIU)

[_sry, i couldn't resist_]

SirCmpwn:
I've noticed this tool. Swaylock definitely needs work.

Fale:
I'm not able to reproduce this on swaylock version  (2016-10-26, branch ""). Are you still able to reproduce this? (I tried 5 times, 30 seconds each time)

kleinreact:
I just tested it again and I'm still able to reproduce the error. My current build is from a chechout from 2016-11-17, main branch. "wlc", as a dependency, has been checked out and build at the same date. Everything else are system packages.

To reproduce the error I am pressing with all ten fingers at high speed randomly on the keyboard, but I just use the letter keys, so no special keys, especially you have to avoid pressing enter. 

The error is reproducable with the internal keywboards as well as with an external bluetooth keyboard.

Fale:
I've tried and I'm not able to reproduce the problem. I wonder if it depends on other software or hardware

Lourens-Rich:
Will you retest this problem after new commit?

FSMaxB:
Is it possible to test this with a fuzzer like [afl](http://lcamtuf.coredump.cx/afl/)?

kleinreact:
I debugged the problem a little bit and found out the following:

The error always appears while executing `cairo_paint(window->cairo)` (swaylock, line 220), which is reached via the following call trace:
```
notify_key(...)
 -> render(...)               // line 210
     -> render_color(...)     // line 577
         -> cairo_paint(...)  // line 220
```
Commenting out line 577 does not solve the problem, but swaylock now crashes with a much more verbose error message:
```
wl_buffer@17: error 2: error accessing SHM buffer
11/29/16 11:09:21 - [main.c:29] failed to run wl_display_dispatch
```
Commenting out lines 210-213 solves the problem for me. However I do not have the fancy animations any more. Note that the problem cannot fixed by just using `swaylock -u` instead.

So it looks to me like something wayland or cairo related. For comparison, here is my current configuration:
```
swaylock:          0.10-rc1-68-gcd5694f (2016-11-29, branch "HEAD")
[...]

PR #1705, opened 2018-04-03
Port swaylock to layer shell

SirCmpwn:
Fixes https://github.com/swaywm/sway/issues/1506, fixes https://github.com/swaywm/sway/issues/1141, fixes https://github.com/swaywm/sway/issues/918, fixes https://github.com/swaywm/sway/issues/605, fixes https://github.com/swaywm/sway/issues/1072

Ref https://github.com/swaywm/sway/issues/797


common/meson.build
```diff
@@ -9,6 +9,7 @@ lib_sway_common = static_library(
 		'pango.c',
 		'readline.c',
 		'stringop.c',
+		'unicode.c',
```
ascent12: Would it be possible to use mbrtoc32/c32rtomb from <uchar.h>, instead of doing it ourselves?

SirCmpwn: No, it's not portable.

ascent12: They're standard C11 functions. I'm not certain, but they may introduce some locale-specific shit, though.

SirCmpwn:
This is ready for review.

protocols/wlr-input-inhibitor-unstable-v1.xml
```diff
@@ -0,0 +1,67 @@
...
+  <interface name="zwlr_input_inhibitor_v1" version="1">
+    <description summary="inhibits input to other clients">
+      While this resource exists, input to clients other than the owner of the
+      inhibitor resource will not receive input events. The client that owns
+      this resource will receive all input events normally. The compositor will
+      also disable all of its own input processing (such as keyboard shortcuts)
+      while the inhibitor is active.
+
+      The compositor may continue to send input events to selected clients,
+      such as an on-screen keyboard (via the input-method protocol).
```
acrisci: what is the input-method protocol?

sway/input/input-manager.c
```diff
@@ -263,6 +264,32 @@ static void handle_new_input(struct wl_listener *listener, void *data) {
...
+	struct sway_input_manager *input_manager = wl_container_of(
+			listener, input_manager, inhibit_deactivate);
+	struct sway_seat *seat;
+	wl_list_for_each(seat, &input_manager->seats, link) {
+		seat_set_exclusive_client(seat, NULL);
+		struct sway_container *previous = seat_get_focus(seat);
+		if (previous) {
+			wlr_log(L_DEBUG, "Returning focus to %p %s '%s'", previous,
+					container_type_to_str(previous->type), previous->name);
+			// Hack to get seat to re-focus the return value of get_focus
```
acrisci: what

sway/input/seat.c
```diff
@@ -453,6 +474,51 @@ void seat_set_focus_layer(struct sway_seat *seat,
...
+void seat_set_exclusive_client(struct sway_seat *seat,
+		struct wl_client *client) {
+	if (!client) {
+		seat->exclusive_client = client;
+		// Triggers a refocus of the topmost surface layer if necessary
+		// TODO: Make layer surface focus per-output based on cursor position
+		for (int i = 0; i < root_container.children->length; ++i) {
+			struct sway_container *output = root_container.children->items[i];
+			if (!sway_assert(output->type == C_OUTPUT,
+						"root container has non-output child")) {
```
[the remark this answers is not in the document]
SirCmpwn: on sway it won't be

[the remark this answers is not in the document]
acrisci: It should be represented that way in the ipc `get_tree` or you'll break people's scripts.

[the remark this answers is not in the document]
SirCmpwn: Noted

sway/input/input-manager.c
acrisci: have you considered using the `send_focus` function?

acrisci: it may be better, set focus to `NULL`.

protocols/wlr-input-inhibitor-unstable-v1.xml
```diff
@@ -0,0 +1,67 @@
...
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="zwlr_input_inhibit_manager_v1" version="1">
+    <description summary="inhibits input events to other clients">
+      Clients can use this interface to prevent input events from being sent to
+      any surfaces but its own, which is useful for example in lock screen
+      software. It is assumed that access to this interface will be locked down
+      to whitelisted clients by the compositor.
```
acrisci: I think you should specify what happens to input focus of the existing clients when this gets created.

protocols/wlr-input-inhibitor-unstable-v1.xml
acrisci: I think you could qualify this a bit better.

sway/input/input-manager.c
SirCmpwn: I'm not 100% on the best way to refactor this, I would appreciate it if you filed a follow-up patch improving it if you're not happy with it

protocols/wlr-input-inhibitor-unstable-v1.xml
SirCmpwn: Please move protocol feedback to the wlr-protocols thread

sway/input/seat.c
```diff
@@ -453,6 +474,51 @@ void seat_set_focus_layer(struct sway_seat *seat,
...
+		}
+	}
+	if (seat->has_focus) {
+		struct sway_container *focus = seat_get_focus(seat);
+		if (focus->type == C_VIEW && wl_resource_get_client(
+					focus->sway_view->surface->resource) != client) {
+			seat_set_focus(seat, NULL);
+		}
+	}
+	if (seat->wlr_seat->pointer_state.focused_client) {
```
acrisci: I don't want library users to dig into `pointer_state`. What about making a getter for this now that it's important?

sway/input/seat.c
```diff
@@ -453,6 +474,51 @@ void seat_set_focus_layer(struct sway_seat *seat,
...
+			wlr_seat_pointer_clear_focus(seat->wlr_seat);
+		}
+	}
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	struct wlr_touch_point *point;
+	wl_list_for_each(point, &seat->wlr_seat->touch_state.touch_points, link) {
+		if (point->client->client != client) {
+			wlr_seat_touch_point_clear_focus(seat->wlr_seat,
+					now.tv_nsec / 1000, point->touch_id);
```
acrisci: is this way of getting the time compatible with what we get from the real events?

SirCmpwn: ya

sway/input/seat.c
SirCmpwn: +1

swaylock/render.c
```diff
@@ -0,0 +1,145 @@
...
+		cairo_arc(cairo, buffer_width / 2, buffer_height / 2,
+				ARC_RADIUS + ARC_THICKNESS / 2, 0, 2 * M_PI);
+		cairo_stroke(cairo);
+	}
+
+	wl_surface_set_buffer_scale(surface->surface, surface->scale);
+	wl_surface_attach(surface->surface, surface->current_buffer->buffer, 0, 0);
+	wl_surface_damage(surface->surface, 0, 0, buffer_width, buffer_height);
+	wl_surface_commit(surface->surface);
+	wl_display_roundtrip(state->display);
```
acrisci: why are you roundtripping here?

SirCmpwn: https://github.com/swaywm/sway/blob/e79ef491c8fbcd4ed42d47478428b0e272a03348/swaylock/password.c#L102-L108

acrisci:
It doesn't compile. I don't have `rand()`.

acrisci:
I don't have `exit()`. You should include `stdlib.h`.

swaylock/main.c
```diff
@@ -390,415 +147,123 @@ int main(int argc, char **argv) {
...
-				lock_set_lock_surface(registry->swaylock, output->output, window->surface);
-			}
-			locked = true;
-		}
+	struct swaylock_surface *surface;
+	wl_list_for_each(surface, &state.surfaces, link) {
+		surface->image = background_image;
+
+		assert(surface->surface =
+				wl_compositor_create_surface(state.compositor));
```
emersion: No side effects in assertions

swaylock/password.c
```diff
@@ -0,0 +1,125 @@
...
+fail:
+	// PAM frees this
+	pw->buffer = NULL;
+	pw->len = pw->size = 0;
+	return false;
+}
+
+static bool backspace(struct swaylock_password *pw) {
+	if (pw->len != 0) {
+		pw->buffer[--pw->len] = 0;
```
emersion: Side note: with assignments in `if`s, side effects in array brackets are practices I don't like and are error-prone. Especially for a lockscreen.

swaylock/password.c
```diff
@@ -0,0 +1,125 @@
...
+void swaylock_handle_key(struct swaylock_state *state,
+		xkb_keysym_t keysym, uint32_t codepoint) {
+	switch (keysym) {
+		case XKB_KEY_KP_Enter: /* fallthrough */
+		case XKB_KEY_Return:
+			state->auth_state = AUTH_STATE_VALIDATING;
+			render_frames(state);
+			wl_display_roundtrip(state->display);
+			if (attempt_password(&state->password)) {
+				exit(0);
```
emersion: Should cleanup all resources.

swaylock/render.c
```diff
@@ -0,0 +1,146 @@
...
+#include "background-image.h"
+#include "swaylock/swaylock.h"
+
+#define M_PI 3.14159265358979323846
+
+void render_frame(struct swaylock_surface *surface) {
+	struct swaylock_state *state = surface->state;
+	surface->current_buffer = get_next_buffer(state->shm,
+			surface->buffers,
+			surface->width * surface->scale,
```
emersion: Define `buffer_width` before this instruction, and replace `surface->width * surface->scale` with `buffer_width`

swaylock/render.c
```diff
@@ -0,0 +1,146 @@
...
+	if (state->args.mode == BACKGROUND_MODE_SOLID_COLOR) {
+		cairo_set_source_u32(cairo, state->args.color);
+		cairo_paint(cairo);
+	} else {
+		render_background_image(cairo, surface->image,
+				state->args.mode, buffer_width, buffer_height);
+	}
+	cairo_identity_matrix(cairo);
+
+	int ARC_RADIUS = 50 * surface->scale;
```
emersion: Why are these in all caps?

SirCmpwn: Used to be constants. Will lowercase them now.

swaylock/render.c
```diff
@@ -0,0 +1,146 @@
...
+				ARC_RADIUS - ARC_THICKNESS / 2, 0, 2 * M_PI);
+		cairo_stroke(cairo);
+		cairo_arc(cairo, buffer_width / 2, buffer_height / 2,
+				ARC_RADIUS + ARC_THICKNESS / 2, 0, 2 * M_PI);
+		cairo_stroke(cairo);
+	}
+
+	wl_surface_set_buffer_scale(surface->surface, surface->scale);
+	wl_surface_attach(surface->surface, surface->current_buffer->buffer, 0, 0);
+	wl_surface_damage(surface->surface, 0, 0, buffer_width, buffer_height);
```
emersion: If you want to damage `buffer_{width,height}`, use `wl_surface_damage_buffer`. Or you could just use `surface->{width,height}`.

swaylock/password.c
SirCmpwn: I don't think this is error prone, this behavior is well defined

emersion: I'm not saying this is undefined behavior, I'm just saying it's unreadable and thus error-prone, even for people used to these idioms. It's just like `if`s without brackets.

Anyway, I don't care too much, this is just a warning to say "if there's a bug because of this, don't blame me for having approved this PR".

SirCmpwn: I know what you're saying. I disagree.

SirCmpwn:
Thanks!

PR #1707, opened 2018-04-03
Implement opacity command

### LINKED ISSUE #1558: command to set container opacity
A command `opacity <percent>` to set container opacity.

Set current container to be semitransparent:

```
swaymsg 'opacity 50%'
```

Dim my terminals:

```
for_window [class="terminal"] opacity 75%
```

acrisci:
Test: `swaymsg transparency 0.25`.

fixes #1558

### DIFF
--- a/sway/commands/opacity.c
+++ b/sway/commands/opacity.c
@@ -1,0 +1,39 @@
+#include <assert.h>
+#include <stdlib.h>
+#include "sway/commands.h"
+#include "sway/tree/view.h"
+#include "log.h"
+
+static bool parse_opacity(const char *opacity, float *val) {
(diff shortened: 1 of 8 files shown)


SirCmpwn:
I think I would rather call this opacity than transparency

sway/commands/opacity.c
```diff
@@ -0,0 +1,36 @@
+#include <assert.h>
+#include "sway/commands.h"
+#include "sway/tree/view.h"
+
+static bool parse_opacity(const char *opacity, float *val) {
+	*val = atof(opacity);
```
SirCmpwn: Still should be strtof

PR #1711, opened 2018-04-03
add headless outputs option

acrisci:
Add option `-H` for headless mode for tests.

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -43,7 +43,7 @@ struct sway_server {
 
 struct sway_server server;
 
-bool server_init(struct sway_server *server);
+bool server_init(struct sway_server *server, bool headless);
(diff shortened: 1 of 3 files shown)


emersion:
I'd prefer commands to spawn new outputs via IPC instead of env variables.

SirCmpwn:
>I'd prefer commands to spawn new outputs via IPC instead of env variables.

This

We should start with no outputs and make you use IPC to add them

acrisci:
This is an extraordinary amount of work.I've already done this in the backend command PR and ran into problems. If sway starts with no outputs, it will crash because outputs are assumed all over the code.

I will abandon this in favor of #1531 and keep in for myself for the tests if you require me to do this now.

PR #1713, opened 2018-04-03
swaybg: add HiDPI support

emersion:
Seems to work fine on Sway's side


swaybg/main.c
```diff
@@ -74,8 +75,8 @@ static void render_image(struct swaybg_state *state) {
 	cairo_surface_t *image = state->context.image;
 	double width = cairo_image_surface_get_width(image);
 	double height = cairo_image_surface_get_height(image);
-	int wwidth = state->width;
```
SirCmpwn: While you're here can you rename these to `surface_{width,height}`

emersion: `surface_{width,height}` are unscaled in Wayland terms, so I picked `buffer_{width,height}`

PR #1720, opened 2018-04-04
<sway/man: Fix path typo for sway-security(7) default path>

bobsaintcool:
Update the path provided in the documentation that refer to the
default security value provided by swaywm project.


SirCmpwn:
Thanks!

Issue #1137, opened 2017-03-24
wrong position on swaybar when scale=2


motherload:
When scale=2 is enabled in the config file, the swaybar aesthetically scales correctly, everything looks fine.

But if you try to click the compartments to select some workspace, it becomes clear that the they do not coincide. Can't say by how much, but the click compartment is more than 50% wider than what you see on the screen. If you have more than 2 or 3 active workspaces, it becomes nearly impossible to select them accurately with the mouse click.

ascent12:
I've noticed that this is still an issue, even for unscaled outputs. With scale=1, when you have 10 workspaces open, clicking on the right half of '9' will select '10'.
I suspect that the 1px border between buttons is not being accounted for when working out which button has been clicked.

SirCmpwn:
Good catch.

Issue #797, opened 2016-07-28
HiDPI support meta-issue


SirCmpwn:
- [x] Specify scale in output command
- [x] Pass scale factor to wlc
- [x] Screenshots
- [x] Client-side support
  - [x] swaybg
  - [x] swaybar
  - [x] swaylock
- [x] Window borders
- [x] swaybg given the wrong size on the screen
- [x] gaps

Hummer12007:
Gaps are 2x bigger on hidpi (should be scaled).

SirCmpwn:
Client-side support done as of https://github.com/SirCmpwn/sway/compare/0913eff9157a...af441541197a

mlofjard:
I get mouse cursor size issues when adding scale 2x to my config. On the title bar and background surfaces the mouse cursor is really tiny (not scaled at all), and in urxvt or weston-terminal (i.e. both wayland and xwayland apps) the mouse cursor is 4x the size. I dont get the correct size of the cursor anywhere.

ametisf:
I have mouse cursor scale issues without any scaling specified in my config. Above Qt clients (qutebrowser, qtcreator) it appears double the expected size, anything not-Qt is ok (xterm, wlterm, chromium, firefox).

william01110111:
Programs running in XWayland are scaled correctly, but this is done (presumably) by rendering them at a low resolution and scaling them up because they are all low res and bury. A better solution would be to render them at full resolution with the proper HiDPI X11 settings or even have some mechanism to choose between the two options (since not everything works well on HiDPI). I know this is possible because Gnome does it and everything looks great. I'm willing to help implement this if you point me in the right direction.

handijk:
Choosing between the two options does not seem possible, see: https://github.com/Enerccio/ewlc/issues/82
I have created an issue to allow for setting a global DPI for xwayland in ewlc, see: https://github.com/Enerccio/ewlc/issues/99

SirCmpwn:
Note that sway doesn't use ewlc. If you'd like to see XWayland DPI improvements in Sway, they should be submitted to wlc.

Enerccio:
@SirCmpwn sway can be build with ewlc (tested):

   cmake -DWLC_LIBRARIES=<ewlc-path>/target/src/libewlc.so -DWLC_INCLUDE_DIRS=<ewlc-path>/include ..

SirCmpwn:
That doesn't surprise me, but we're still not switching right now.

[...]

PR #1722, opened 2018-04-04
Add hidpi support to swaybar

SirCmpwn:
Fixes https://github.com/swaywm/sway/issues/1137

Ref https://github.com/swaywm/sway/issues/797

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -319,7 +319,7 @@ void bar_setup(struct swaybar *bar,
 	}
 
 	pointer->cursor_theme = wl_cursor_theme_load(
-				NULL, 16 * (max_scale * 2), bar->shm);
+				NULL, 24 * max_scale, bar->shm);
 	assert(pointer->cursor_theme);
 	struct wl_cursor *cursor;
 	cursor = wl_cursor_theme_get_cursor(pointer->cursor_theme, "left_ptr");
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -69,11 +69,19 @@ static void wl_pointer_enter(void *data, struct wl_pointer *wl_pointer,
 			break;
 		}
 	}
+	int max_scale = 1;
+	struct swaybar_output *_output;
+	wl_list_for_each(_output, &bar->outputs, link) {
+		if (_output->scale > max_scale) {
+			max_scale = _output->scale;
+		}
+	}
+	wl_surface_set_buffer_scale(pointer->cursor_surface, max_scale);
 	wl_surface_attach(pointer->cursor_surface,
 			wl_cursor_image_get_buffer(pointer->cursor_image), 0, 0);
 	wl_pointer_set_cursor(wl_pointer, serial, pointer->cursor_surface,
-			pointer->cursor_image->hotspot_x,
-			pointer->cursor_image->hotspot_y);
+			pointer->cursor_image->hotspot_x / max_scale,
+			pointer->cursor_image->hotspot_y / max_scale);
 	wl_surface_commit(pointer->cursor_surface);
 }
 
@@ -103,10 +111,12 @@ static void wl_pointer_button(void *data, struct wl_pointer *wl_pointer,
 	}
 	struct swaybar_hotspot *hotspot;
 	wl_list_for_each(hotspot, &output->hotspots, link) {
-		if (pointer->x >= hotspot->x
-				&& pointer->y >= hotspot->y
-				&& pointer->x < hotspot->x + hotspot->width
-				&& pointer->y < hotspot->y + hotspot->height) {
+		double x = pointer->x * output->scale;
+		double y = pointer->y * output->scale;
+		if (x >= hotspot->x
+				&& y >= hotspot->y
+				&& x < hotspot->x + hotspot->width
+				&& y < hotspot->y + hotspot->height) {
 			hotspot->callback(output, pointer->x, pointer->y,
 					button, hotspot->data);
 		}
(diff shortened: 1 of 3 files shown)


swaybar/bar.c
```diff
@@ -262,9 +305,20 @@ void bar_setup(struct swaybar *bar,
...
-	assert(pointer->cursor_theme = wl_cursor_theme_load(NULL, 16, bar->shm));
+	int max_scale = 1;
+	struct swaybar_output *output;
+	wl_list_for_each(output, &bar->outputs, link) {
+		if (output->scale > max_scale) {
+			max_scale = output->scale;
+		}
+	}
+
+	assert(pointer->cursor_theme = wl_cursor_theme_load(
```
ascent12: You shouldn't put a side effect in an assertion.

SirCmpwn: I do this all the time, it's all over sway et al. This is fine.

SirCmpwn: Nevermind duh I see why

SirCmpwn: Pushed fix to this file but will file a follow-up ticket to go fix my fuck-ups elsewhere

SirCmpwn:
>All of this code works but is hard to read. Converting between surface-local and buffer-local coordinates at the beginning and the end of functions doesn't help. I'd prefer everything to take coordinates in surface-local coordinates (especially hotspots), and mark variables using buffer-local coordinates with a buffer_ prefix.

Part of the problem is that we need to communciate the scale to pango so it can up the font size/etc accordingly. I think that the approach used here is reasonably good considering this constraint.

swaybar/render.c
```diff
@@ -260,15 +290,23 @@ static uint32_t render_status_line(cairo_t *cairo,
...
+
+	int ws_vertical_padding = WS_VERTICAL_PADDING * output->scale;
+	int ws_horizontal_padding = WS_HORIZONTAL_PADDING * output->scale;
+	int border_width = BORDER_WIDTH * output->scale;
+
 	uint32_t ideal_height = text_height + ws_vertical_padding * 2
 		+ border_width * 2;
-	if (height < ideal_height) {
-		return ideal_height;
+	if (height < ideal_height / output->scale) {
```
emersion: Here's an example of why this PR is unreadable: here you compare `height` (in buffer coordinates) with `ideal_height / output->scale` (in surface coordinates). That doesn't make sense.

swaybar/bar.c
```diff
@@ -256,33 +299,45 @@ void bar_setup(struct swaybar *bar,
...
+	int max_scale = 1;
+	struct swaybar_output *output;
+	wl_list_for_each(output, &bar->outputs, link) {
+		if (output->scale > max_scale) {
+			max_scale = output->scale;
+		}
+	}
+
+	pointer->cursor_theme = wl_cursor_theme_load(
+				NULL, 16 * (max_scale * 2), bar->shm);
```
emersion: Cursor size changes when hovering the bar. Should be `24 * max_scale`.

emersion: What if the output scale is dynamically changed?

swaybar/render.c
```diff
@@ -14,55 +14,73 @@
...
-	return ideal_height;
+	return ideal_height / output->scale;
 }
 
 static uint32_t render_status_line_text(cairo_t *cairo,
-		struct swaybar_config *config, const char *text,
-		bool focused, double *x, uint32_t width, uint32_t height) {
+		struct swaybar_output *output, struct swaybar_config *config,
+		const char *text, bool focused, double *x,
+		uint32_t width, uint32_t surface_height) {
```
emersion: `surface_width`

emersion: It's unused actually, does it make sense to remove it?

swaybar/render.c
```diff
@@ -133,10 +157,10 @@ static uint32_t render_status_block(cairo_t *cairo,
 		if (config->sep_symbol) {
 			int _height;
 			get_text_size(cairo, config->font, &sep_width, &_height,
-					1, false, "%s", config->sep_symbol);
+					output->scale, false, "%s", config->sep_symbol);
 			uint32_t _ideal_height = _height + ws_vertical_padding * 2;
-			if (height < _ideal_height) {
-				return _height;
+			if (height < _ideal_height / output->scale) {
```
emersion: Should that be `_height`?

swaybar/render.c
```diff
@@ -242,33 +272,41 @@ static uint32_t render_status_line_i3bar(cairo_t *cairo,
 static uint32_t render_status_line(cairo_t *cairo,
 		struct swaybar_config *config, struct swaybar_output *output,
 		struct status_line *status, bool focused,
-		double *x, uint32_t width, uint32_t height) {
+		double *x, uint32_t width, uint32_t surface_height) {
```
emersion: Ditto: `surface_width`?

swaybar/render.c
```diff
@@ -330,14 +368,22 @@ static uint32_t render_workspace_button(cairo_t *cairo,
 		box_colors = config->colors.inactive_workspace;
 	}
 
+	uint32_t height = surface_height *output->scale;
```
emersion: Oh my, missing a space

swaybar/render.c
```diff
@@ -462,9 +510,11 @@ void render_frame(struct swaybar *bar, struct swaybar_output *output) {
...
 		cairo_paint(shm);
 
+		wl_surface_set_buffer_scale(output->surface, output->scale);
 		wl_surface_attach(output->surface,
 				output->current_buffer->buffer, 0, 0);
-		wl_surface_damage(output->surface, 0, 0, output->width, output->height);
+		wl_surface_damage(output->surface, 0, 0,
+				output->width, output->height);
 		wl_surface_commit(output->surface);
 		wl_display_roundtrip(bar->display);
```
emersion: I still don't get why this roundtrip is needed?

swaybar/bar.c
SirCmpwn: Aha

SirCmpwn: If the output scale is dynamically changed we're SOL, I want to file a ticket and deal with it never

swaybar/render.c
SirCmpwn: yup

swaybar/render.c
SirCmpwn: oh my

swaybar/render.c
SirCmpwn: We don't dispatch in a loop like some of our other clients do.

emersion: Oh, I see, makes sense

PR #1733, opened 2018-04-04
try to fix focus-inactive

acrisci:
Focus-inactive is broken.

[test plan](https://github.com/swaywm/sway-tests/blob/master/test/test_split.py)


SirCmpwn:
Can you elaborate on what the bug was so we can test that this fixes it?

acrisci:
Ok ready to go.

PR #1736, opened 2018-04-05
Ipc container type

acrisci:
add "type" to ipc json containers.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -132,6 +132,7 @@ static void ipc_json_describe_workspace(struct sway_container *workspace,
 static void ipc_json_describe_view(struct sway_container *c, json_object *object) {
 	json_object_object_add(object, "name",
 			c->name ? json_object_new_string(c->name) : NULL);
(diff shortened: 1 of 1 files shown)


sway/ipc-json.c
```diff
@@ -132,6 +132,7 @@ static void ipc_json_describe_workspace(struct sway_container *workspace,
 static void ipc_json_describe_view(struct sway_container *c, json_object *object) {
 	json_object_object_add(object, "name",
 			c->name ? json_object_new_string(c->name) : NULL);
+	json_object_object_add(object, "type", json_object_new_string("con"));
```
emersion: Is it possible to use "container" instead? "con" is a bit weird.

acrisci:
> Is it possible to use "container" instead? "con" is a bit weird.

> Do we want to add a "view" type?

I wish we could, but it would break compatibility.

emersion:
Oh, if it comes from i3, I guess it can't be helped.

PR #1745, opened 2018-04-05
Handle output removal on swaybar

SirCmpwn:
Test plan:

* swaybar
* output removal
* handle

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -34,6 +34,34 @@ static void bar_init(struct swaybar *bar) {
 	wl_list_init(&bar->outputs);
 }
 
+static void swaybar_output_free(struct swaybar_output *output) {
+	if (!output) {
+		return;
+	}
+	wlr_log(L_DEBUG, "Removing output %s", output->name);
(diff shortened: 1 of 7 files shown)


swaybar/bar.c
```diff
@@ -270,9 +271,44 @@ static void handle_global(void *data, struct wl_registry *registry,
...
+}
+
 static void handle_global_remove(void *data, struct wl_registry *registry,
 		uint32_t name) {
-	// who cares
+	struct swaybar *bar = data;
+	struct swaybar_output *output, *tmp;
+	wl_list_for_each_safe(output, tmp, &bar->outputs, link) {
+		if (output->wl_name == name) {
+			swaybar_output_free(output);
```
emersion: It's a good idea to handle output removal, but there's also another way swaybar can get destroyed: with the layer-surface `closed` event. We should handle this one too (since the compositor can close a swaybar without removing an output). There's a `TODO: Deal with hotplugging` about this above.

SirCmpwn: addressed in latest patch

PR #1747, opened 2018-04-05
ipc layout

acrisci:
Make the containers have the "layout" in ipc json for tests.

[layout tests](https://github.com/swaywm/sway-tests/blob/master/test/test_layout.py)

Will also be added to move tests.

Also don't send container json if no listeners are subscribed to the event.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -12,37 +12,22 @@
 #include "wlr-layer-shell-unstable-v1-protocol.h"
 
 static const char *ipc_json_layout_description(enum sway_container_layout l) {
-	const char *layout;
-
 	switch (l) {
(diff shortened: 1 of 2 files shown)


sway/ipc-json.c
```diff
@@ -11,6 +11,25 @@
...
+	case L_HORIZ:
+		return "splith";
+	case L_TABBED:
+		return "tabbed";
+	case L_STACKED:
+		return "stacked";
+	case L_FLOATING:
+		return "floating";
+	case L_NONE:
+	case L_LAYOUTS:
```
SirCmpwn: We can probably just remove L_LAYOUTS, it's not used anywhere

PR #1757, opened 2018-04-06
Break everything^W^WUse wlr_box for sway_container

SirCmpwn:


include/sway/tree/container.h
```diff
@@ -71,9 +71,7 @@ struct sway_container {
 	// For C_ROOT, this has no meaning
```
emersion: While you're at it, can you change "For C_ROOT, this has no meaning" to "this is the output layout extends"?

PR #1758, opened 2018-04-06
Revert "Break everything^W^WUse wlr_box for sway_container"

SirCmpwn:
Reverts swaywm/sway#1757


SirCmpwn:
We'll do this after we're done with stuff like move, it breaks too much shit and I'm out of steam for debugging it.

PR #1761, opened 2018-04-07
Minor changes to build against json-c 0.13.x

jhj:
I'm not convinced my change to CMakeLists is the correct way to squelch
the json-c version error when building on a system with version
numbers greater than 0.12.1. 0.13.x+ is supposed to be backwards
compatible. Rest of the changes involve replacing int with size_t
where necessary to suppress -Werror=sign-compare on gcc 8.


SirCmpwn:
Build from the 0.15 branch.

PR #1769, opened 2018-04-07
Focus inactive fixes

### LINKED ISSUE #1670: update focus-inactive on window close
When a window closes, the focus-inactive list should be updated to reflect the window closing.

For instance, if you switch workspaces, and open a window on that workspace, and then that window closes, focus-inactive should be reverted to the visible workspace and not the previous workspace.

The function `workspace_is_visible()` will not work correctly without this feature.

I think it will be enough to replace the window with the next focus inactive of the parent, or the parent itself in this case, but that will need some testing.

acrisci:
Implement the ipc focus stack and focus-inactive bug hunting.

[tests](https://github.com/swaywm/sway-tests/blob/master/test/test_focus_inactive.py) (passing)

fixes #1670.


acrisci:
Criteria cannot select containers, only views. That needs to be fixed.

sway/criteria.c
```diff
@@ -273,7 +273,7 @@ static int regex_cmp(const char *item, const pcre *regex) {
 
 // test a single view if it matches list of criteria tokens (all of them).
 static bool criteria_test(struct sway_container *cont, list_t *tokens) {
-	if (cont->type != C_VIEW) {
+	if (cont->type < C_CONTAINER) {
```
SirCmpwn: Don't do numeric comparisons on enum types, enumerate the valid options and check for equality

include/sway/input/seat.h
```diff
@@ -87,8 +87,19 @@ struct sway_container *seat_get_focus(struct sway_seat *seat);
...
-		struct sway_container *container, enum sway_container_type type);
+/**
+ * Descend into the focus stack to find the focus-inactive view. Useful for
+ * container placement when they change position in the tree.
+ */
+struct sway_container *seat_get_focus_inactive_view(struct sway_seat *seat,
+		struct sway_container *container);
+
+/**
+ * Iterate over the focus-inactive children of the container calling the function on each.
```
emersion: 80 char limit

SirCmpwn:
Thanks!

PR #1772, opened 2018-04-07
Improve dependency check for libinput

bpinto:
Feel free to close this PR if it's not something you would like to have.

I created this PR because the `README` file mentions this version is required (error prone) while the build doesn't check it.


SirCmpwn:
Thanks!

PR #1773, opened 2018-04-08
Use full ws->name in swaybar hotspot callback

thejan2009:
Problem: clicking on swaybar workspace hotspot creates new workspace if `strip_workspace_numers` is enabled. Reason is wrong name in hotspot callback.

Example config subset:
```
set $ws1 "1:&#x3B1;"
set $ws2 "2:&#x3B2;"

bindsym $mod+1 workspace $ws1
bindsym $mod+2 workspace $ws2

bar {
	strip_workspace_numbers yes
}
```

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -352,6 +352,7 @@ static uint32_t render_workspace_button(cairo_t *cairo,
 		struct swaybar_output *output, struct swaybar_config *config,
 		struct swaybar_workspace *ws, double *x, uint32_t surface_height) {
 	const char *name = ws->name;
+	const char *whole_name = ws->name;
 	if (config->strip_workspace_numbers) {
 		name = strip_workspace_number(ws->name);
 	}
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Good catch.

swaybar/render.c
```diff
@@ -411,7 +412,7 @@ static uint32_t render_workspace_button(cairo_t *cairo,
 	hotspot->height = height;
 	hotspot->callback = workspace_hotspot_callback;
 	hotspot->destroy = free;
-	hotspot->data = strdup(name);
+	hotspot->data = strdup(whole_name);
```
emersion: No need for an additional variable here right? We could just `strdup(ws->name)`?

SirCmpwn: I think you're missing the point of the patch here

SirCmpwn: Shit I see what you mean

thejan2009: @SirCmpwn you're too fast :)
@emersion It's such a small change that I don't think it really matters..

SirCmpwn: I pushed a fixup commit

PR #1775, opened 2018-04-08
Add workspace_auto_back_and_forth command

thejan2009:
This is the only missing piece - other code regarding this functionality has
already been ported from pre-wlroots source.


SirCmpwn:
Thanks!

PR #1781, opened 2018-04-08
Add input "identifier" map_to_output "identifier"

SirCmpwn:


sway/config/input.c
```diff
@@ -88,6 +88,10 @@ void merge_input_config(struct input_config *dst, struct input_config *src) {
 		free(dst->xkb_variant);
 		dst->xkb_variant = strdup(src->xkb_variant);
 	}
+	if (src->mapped_output) {
+		free(src->mapped_output);
```
thejan2009: s/src/dst/

SirCmpwn: Thanks!

PR #1783, opened 2018-04-09
Fix buffer issues in swaybar status line

RyanDwyer:
This PR fixes a null termination issue and two dangling pointers.

# Null termination issue

The string that read() returns is not null terminated, so the `while (*cur)` loop could run into garbage data. In all my testing this has sometimes resulted in a "failed to parse i3bar json" message, and sometimes an infinite loop where swaybar pegs the CPU and nothing is displayed (or updated) in the status line part of the bar.

To prevent this I'm setting the null byte manually.

To test, run swaybar with a status command several times. Without my PR one of the above symptoms occurs about once in every 5 launches. With my PR it should not display the parse error message nor become unresponsive at all.

# Dangling pointers

When increasing the buffer size, `cur` and `state->current_node` need to be updated accordingly as these point into the old buffer.

To test this, you can make Conky return a JSON element over 4096 bytes long, but it's easier to temporarily change the default buffer_size from 4096 to something small like 16 in swaybar/status_line.c line 74.

### DIFF
--- a/swaybar/i3bar.c
+++ b/swaybar/i3bar.c
@@ -113,8 +113,9 @@ bool i3bar_handle_readable(struct status_line *status) {
 	char *cur = &state->buffer[state->buffer_index];
 	ssize_t n = read(status->read_fd, cur,
 			state->buffer_size - state->buffer_index);
-	if (n == 0) {
-		return 0;
+	if (n == -1) {
+		status_error(status, "[failed to read from status command]");
+		return false;
 	}
 
 	if (n == (ssize_t)(state->buffer_size - state->buffer_index)) {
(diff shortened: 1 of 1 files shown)


ascent12:
Looks good to me.
This doesn't have anything to do with your code, but right above it, I'm seeing that it's trying to return -1 on failure in a function returning a bool, which is definitely wrong.

swaybar/i3bar.c
```diff
@@ -125,9 +125,12 @@ bool i3bar_handle_readable(struct status_line *status) {
 			status_error(status, "[failed to allocate buffer]");
 			return -1;
 		}
+		state->current_node += new_buffer - state->buffer;
+		cur += new_buffer - state->buffer;
 		state->buffer = new_buffer;
 	}
 
+	cur[n] = '\0';
```
emersion: Should read at most `state->buffer_size - state->buffer_index - 1` bytes at line 114?

emersion: Also, not related to your patch but `read` returns -1 on error.

ascent12: I think if read() ends up reading enough to fill the entire buffer, realloc would get called, and would provide the space for this null terminator.
The logic is not very obvious, though.

PR #1791, opened 2018-04-10
Fix swaybar output config

RyanDwyer:
Swaybar would crash on launch if it wasn't configured to run on all outputs. This PR fixes that and makes swaybar respect its output config.

What was previously happening was all outputs from the wl_registry were being added to `bar->outputs` regardless of whether the bar actually used them, and ipc_get_workspaces() would iterate these and run strcmp() on `output->name`. For the unused outputs, `output->name` was null which caused swaybar to segfault.

My PR makes it ignore the outputs given by the wl_registry if the bar doesn't use them, noting that we have to keep track of the output index to make sure the `bar->config->outputs` maps correctly. That's the handle_global() part.

The second part of the PR cleans up the logic in bar_setup(). Because `bar->outputs` now contains only used outputs, we can remove the double nested loop. However, we do need to assign the name to the output so we know what workspaces belong to it, and that requires a short loop through `bar->config->outputs`.

I've tested all the following:

* One bar with no outputs configured - bar appears on all outputs
* One bar with `output *` - bar appears on all outputs
* One bar with `output VGA-1` - bar appears on that output only
* One bar with `output VGA-1` and `output VGA-2` - bar appears on both outputs
* Two bars on different outputs - bars appear on their given outputs

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -264,6 +264,19 @@ struct wl_output_listener output_listener = {
 	.scale = output_scale,
 };
 
+static bool bar_uses_output(struct swaybar *bar, size_t output_index) {
+	if (bar->config->all_outputs) {
+		return true;
+	}
+	struct config_output *coutput;
+	wl_list_for_each(coutput, &bar->config->outputs, link) {
+		if (coutput->index == output_index) {
+			return true;
+		}
+	}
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #1792, opened 2018-04-10
Fix swaybar not showing all status blocks

### LINKED ISSUE #1790: swaybar only renders the last block
With this `i3status` config, `swaybar` only renders the last block: https://sr.ht/8aYM.config

Running `i3status` from the CLI correctly outputs all blocks.

RyanDwyer:
Fixes #1790.

When rendering a non-edge block, swaybar's render_status_block() was miscalculating the height needed for the separator and was returning early, expecting itself to be called again later with a new surface height.

I've redone the logic here, as well as renamed _height to sep_height.

In addition to this, the separator symbol was not positioned correctly vertically.

For anyone checking my logic, my understanding is:

* `height` is the bar height with scaling applied
* `sep_height` and `_ideal_height` are both without scaling
* The function expects the return value to be without scaling

I've tested with output scales of 0.5, 1 and 2, and the pipe character as the separator symbol.

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -159,9 +159,10 @@ static uint32_t render_status_block(cairo_t *cairo,
 		if (config->sep_symbol) {
 			get_text_size(cairo, config->font, &sep_width, &sep_height,
 					output->scale, false, "%s", config->sep_symbol);
-			uint32_t _ideal_height = sep_height + ws_vertical_padding * 2;
-			if (_ideal_height * output->scale > height) {
-				return _ideal_height;
+			uint32_t _ideal_surface_height = ws_vertical_padding * 2
+				+ sep_height;
+			if (_ideal_surface_height > surface_height) {
+				return _ideal_surface_height;
 			}
 			if (sep_width > block->separator_block_width) {
 				block->separator_block_width = sep_width + margin * 2;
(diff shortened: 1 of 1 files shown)


swaybar/render.c
```diff
@@ -154,15 +154,14 @@ static uint32_t render_status_block(cairo_t *cairo,
...
 		if (config->sep_symbol) {
-			int _height;
-			get_text_size(cairo, config->font, &sep_width, &_height,
+			get_text_size(cairo, config->font, &sep_width, &sep_height,
 					output->scale, false, "%s", config->sep_symbol);
-			uint32_t _ideal_height = _height + ws_vertical_padding * 2;
-			if ((uint32_t)_height < _ideal_height / output->scale) {
-				return _height / output->scale;
+			uint32_t _ideal_height = sep_height + ws_vertical_padding * 2;
+			if (_ideal_height * output->scale > height) {
```
emersion: Can you rename `_ideal_height` to `_ideal_surface_height` (so that it's explicit that this value is in surface-local coordinates, not in buffer-local coordinates)?

Also we should always compare sizes in surface-local coordinates to prevent rounding issues when the size is odd (see #1780). Se we should replace the comparison with `_ideal_surface_height > surface_height`.

RyanDwyer:
Done.

BTW, I wrapped the expression on line 162 because it reached the 80 character limit, and I moved sep_height to the other side of the expression so that the multiplication part can stay together.

emersion:
Thanks!

PR #1801, opened 2018-04-12
<sway/man: In sway(5) treat scale as keyword for output command>

bobsaintcool:
Signed-off-by: Quentin Bourgeois <quentin@bourgeois.eu>


SirCmpwn:
Thanks!

PR #1803, opened 2018-04-13
swaylock: Securely zero-out password.

ggreer:
- Replace char* with static array. Any chars > 1024 will be discarded.
- mlock() password buffer so it can't be written to swap.
- Clear password buffer after auth succeeds or fails.

This is basically the same treatment I gave the 0.15 branch in https://github.com/swaywm/sway/pull/1519


SirCmpwn:
Thanks!

PR #1804, opened 2018-04-13
Fix crashes

martinetd:
I'm not 100% sure about last commit but I can't seem to reproduce any problem with that, and removing the other container's parent looks wrong to me.
Rest should be ok.

I get one more crash with container_move doing random moves (`[sway/desktop/xwayland.c:154] configure:view must be within tree to set position`) that I don't have time to look at, should I open an issue? (you asked not to open issues for this branch, but this part already has been reworked?)


sway/tree/layout.c
```diff
@@ -404,7 +404,7 @@ void container_move(struct sway_container *container,
 				sibling = NULL;
 			} else {
 				wlr_log(L_DEBUG, "Reparenting container (perpendicular)");
-				container_remove_child(container);
+				container_remove_child(sibling);
```
SirCmpwn: Elaborate on this

SirCmpwn:
>you asked not to open issues for this branch, but this part already has been reworked?

If you're writing patches you can open tickets

sway/tree/layout.c
martinetd: All I have is that backtrace:
```
==1977==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000048 (pc 0x00000045368b bp 0x7fffe1220f30 sp 0x7fffe1220f10 T0)
==1977==The signal is caused by a READ memory access.
==1977==Hint: address points to the zero page.
    #0 0x45368a in container_remove_child ../sway/tree/layout.c:124
    #1 0x4555c8 in container_move ../sway/tree/layout.c:410
    #2 0x440543 in cmd_move ../sway/commands/move.c:180
    #3 0x40c3c1 in execute_command ../sway/commands.c:358
    #4 0x4312ae in keyboard_execute_command ../sway/input/keyboard.c:98
    #5 0x4317d5 in keyboard_execute_bindsym ../sway/input/keyboard.c:166
    #6 0x432fdc in handle_keyboard_key ../sway/input/keyboard.c:372
    #7 0x7f9dcd054e58 in wlr_signal_emit_safe ../util/signal.c:29
    #8 0x7f9dccffcfda in handle_keyboard_key ../backend/libinput/keyboard.c:71
    #9 0x7f9dccffbaaa in wlr_libinput_readable ../backend/libinput/backend.c:35
    #10 0x7f9dcd2bcf01 in wl_event_loop_dispatch (/lib64/libwayland-server.so.0+0x9f01)
    #11 0x7f9dcd2bb601 in wl_display_run (/lib64/libwayland-server.so.0+0x8601)
    #12 0x40abcc in server_run ../sway/server.c:140
    #13 0x409ab5 in main ../sway/main.c:418
    #14 0x7f9dcc7ae1ba in __libc_start_main (/lib64/libc.so.6+0x231ba)
    #15 0x408039 in _start (/opt/wayland/bin/sway+0x408039)
```

Now that I'm looking at it again you're right, this change is most likely wrong - I was just thinking that since we're messing with sibling it would make more sense to move it out but we are adding `container` back in so that was likely as intended, just something wrong with it.

I'll just remove that commit and open an issue, this is very close to the other crash I couldn't spend time on

PR #1811, opened 2018-04-13
Fix gcc warnings

martinetd:
Similar to wlroots as well, some string truncations
 * the -1s are harmless since we manually force the final \0 all the time;
 * the strncpy->strcpy is because we based the strncpy on the source length and gcc complained that we never wrote the final \0 then when adding 1 that `strncpy(dst, src, strlen(str)+1)` is useless..........

and another maybe-uninitialized variable

### DIFF
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -48,7 +48,7 @@ int ipc_open_socket(const char *socket_path) {
 		sway_abort("Unable to open Unix socket");
 	}
 	addr.sun_family = AF_UNIX;
-	strncpy(addr.sun_path, socket_path, sizeof(addr.sun_path));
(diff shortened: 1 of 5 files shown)


common/ipc-client.c
```diff
@@ -48,7 +48,7 @@ int ipc_open_socket(const char *socket_path) {
 		sway_abort("Unable to open Unix socket");
 	}
 	addr.sun_family = AF_UNIX;
-	strncpy(addr.sun_path, socket_path, sizeof(addr.sun_path));
+	strncpy(addr.sun_path, socket_path, sizeof(addr.sun_path)-1);
```
SirCmpwn: spaces around operators

martinetd: force pushed spaces (and rebased)

emersion:
Thanks!

Issue #1797, opened 2018-04-11
Empty workspaces are not removed when they're unfocused from another output


SirCmpwn:
Setup:

```
(output 1
  (workspace 1 focus))
(output 2
  (workspace 2 (views...))
  (workspace 3 focus-inactive))
```

Run `workspace 2` and focus will switch to it, but 3 will not be cleaned up.

PR #1816, opened 2018-04-15
remove empty workspace when focusing its output

thejan2009:
Until now, focus changing code only considered cleaning up the last focused
workspace. This commit adds removal of empty workspace from output that just
received focus on a different workspace.

Fixes: #1797


SirCmpwn:
Thanks!

Issue #1650, opened 2018-03-30
Atomic layout updates


SirCmpwn:
We want every frame to be perfect - every buffer shown at its native size, no windows overlaying (in the tiled layout), etc. Accomplishing this is non-trivial. The plan is, in discrete tasks:

# 1. Refactor arrange_windows

Currently, `arrange_windows` lays out the windows, writes the desired boxes of everything to the tree, and sends configure events to clients. This needs to be broken into three steps, which will currently just be called at once (not attempting to be atomic, just refactoring the current behavior). The three steps are:

Call `arrange_layout`, which should most of the work `arrange_windows` does today, but outputs a list of:

```
struct container_layout {
    struct sway_container *container;
    struct wlr_box box;
};
```

We will call this list the "layout". Next, call `configure_layout`, which should send configure events to clients affected by the layout update.

```
for pending in layout:
   client.send_configure <- pending.box
```

Finally, call `apply_layout` to syncronize a potential layout with the actual sway tree.

```
for pending in layout:
    swayc.box <- pending.box
```

Now, let's change `arrange_windows` to look like this:

```
void arrange_windows(struct sway_container *container) {
    struct wl_list layout; // container_layout::link
    layout <- arrange_layout <- container
    configure_layout <- layout
    apply_layout <- layout
}
```

# 2. Wait to apply layouts until all clients are ready

We now need to stash the configure serials for each client that was notified of pending changes. We define this data structure:

```
struct pending_configure {
    int serial;
    struct wl_list link;
};

struct pending_layout {
    struct wl_list layout; // container_layout::link
    struct wl_list configures; // pending_configure::link
};
```

Then we update `arrange_windows`:

```
void arrange_windows(struct sway_container *container) {
    struct pending_layout pending;
    pending.layout <- arrange_layout <- container
    pending.configures <- configure_layout <- pending.layout
    root_container.pending layouts <- pending
}
```

Notably, we don't call `apply_layout`. Instead, we stow the pending layout somewhere and we update the `ack_configure` handler for the shell to look up the pending layout.

```
handle_ack_configure(...) {
    pending_layout <- find pending layout
    pending_configure <- find pending configure
    wl_list_remove(pending_configure.link)
    if empty(pending_layout.configures):
        apply_layout <- pending_layout.layout
}
```

**Milestone**: non-interactive updates will be atomic (mode changes, switch layout from splitv to splith, etc)

# 3. Stash textures on each ack configure

[...]

PR #1817, opened 2018-04-16
Implement fullscreen

RyanDwyer:
This implements `fullscreen enable`, `fullscreen disable`, `fullscreen toggle`, and allows views to request fullscreen.

This is my first significant PR for sway and I've had to learn a lot about how everything works, so please scrutinise this carefully.

Note that sway 0.15 allows the command `fullscreen` without any arguments which acts as toggle, but this is different to the i3 user guide so I've implemented it as described above. As a result, if you're porting config from sway 0.15 you may need to update your command to be `fullscreen toggle`.

wl_shell seems a bit lacking with its fullscreen ability, and from what I understand wl_shell is deprecated anyway, so I've not implemented fullscreen for that. If someone could confirm then that'd be appreciated.

There is one known issue, and that is that popups don't show when the parent view is fullscreen, but I'd like to either resolve that in a separate PR or leave it for someone else to look into.

Tests:

* Run fullscreen enable, fullscreen disable, fullscreen toggle on an xdg_v6 view
* Run fullscreen enable, fullscreen disable, fullscreen toggle on an xwayland view
* Attempt to fullscreen a container - should be ignored 
* Fullscreen a view, then kill it
* Open two views on one workspace, fullscreen one, then attempt to focus the other - focus should remain on fullscreen view
* Open two views on different workspaces, fullscreen one, then attempt to focus the other and back again - workspaces should focus correctly
* Fullscreen view, then move view to another workspace - view should be unfullscreened
* Make an xwayland view request fullscreen, eg. launch firefox and press F11
* Make an xdg_v6 view request fullscreen, eg. launch `mpv --opengl-backend=wayland` and press f

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -72,7 +72,6 @@ struct sway_container {
 	// For C_OUTPUT, this is the output position in layout coordinates
 	// For other types, this is the position in output-local coordinates
 	double x, y;
-	double saved_x, saved_y;
 	// does not include borders or gaps.
 	double width, height;
 
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -45,6 +45,7 @@ enum sway_container_border {
 
 struct sway_root;
 struct sway_output;
+struct sway_workspace;
 struct sway_view;
 
 struct sway_container {
@@ -52,6 +53,7 @@ struct sway_container {
 		// TODO: Encapsulate state for other node types as well like C_CONTAINER
 		struct sway_root *sway_root;
 		struct sway_output *sway_output;
+		struct sway_workspace *sway_workspace;
 		struct sway_view *sway_view;
 	};
 
@@ -74,9 +76,6 @@ struct sway_container {
 	double x, y;
 	// does not include borders or gaps.
 	double width, height;
-
-	// For C_WORKSPACE only
-	struct sway_view *fullscreen;
 
 	list_t *children;
 
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -72,8 +72,12 @@ struct sway_container {
 	// For C_OUTPUT, this is the output position in layout coordinates
 	// For other types, this is the position in output-local coordinates
 	double x, y;
+	double saved_x, saved_y;
 	// does not include borders or gaps.
 	double width, height;
+
+	// For C_WORKSPACE only
+	struct sway_view *fullscreen;
 
 	list_t *children;
 
--- a/sway/commands/fullscreen.c
+++ b/sway/commands/fullscreen.c
@@ -1,0 +1,40 @@
+#include <wlr/types/wlr_wl_shell.h>
+#include "log.h"
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/tree/container.h"
+#include "sway/tree/view.h"
+#include "sway/tree/layout.h"
+
+// fullscreen toggle|enable|disable
+struct cmd_results *cmd_fullscreen(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if (config->reading) return cmd_results_new(CMD_FAILURE, "fullscreen", "Can't be used in config file.");
+	if (!config->active) return cmd_results_new(CMD_FAILURE, "fullscreen", "Can only be used when sway is running.");
+	if ((error = checkarg(argc, "fullscreen", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	struct sway_container *container =
+		config->handler_context.current_container;
+	if (container->type != C_VIEW) {
+		return cmd_results_new(CMD_INVALID, "fullscreen",
+				"Only views can fullscreen");
+	}
+	struct sway_view *view = container->sway_view;
+	bool wants_fullscreen;
+
+	if (strcmp(argv[0], "enable") == 0) {
+		wants_fullscreen = true;
+	} else if (strcmp(argv[0], "disable") == 0) {
+		wants_fullscreen = false;
+	} else if (strcmp(argv[0], "toggle") == 0) {
+		wants_fullscreen = !view->is_fullscreen;
+	} else {
+		return cmd_results_new(CMD_INVALID, "fullscreen",
+				"Expected 'fullscreen <enable|disable|toggle>'");
+	}
+
+	view_set_fullscreen(view, wants_fullscreen);
+
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+}
--- a/sway/commands/fullscreen.c
+++ b/sway/commands/fullscreen.c
@@ -6,8 +6,11 @@
 #include "sway/tree/view.h"
 #include "sway/tree/layout.h"
 
-// fullscreen toggle|enable|disable
 struct cmd_results *cmd_fullscreen(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "fullscreen", EXPECTED_LESS_THAN, 2))) {
+		return error;
+	}
 	struct sway_container *container =
 		config->handler_context.current_container;
 	if (container->type != C_VIEW) {
@@ -25,7 +28,7 @@ struct cmd_results *cmd_fullscreen(int argc, char **argv) {
 		wants_fullscreen = false;
 	} else {
 		return cmd_results_new(CMD_INVALID, "fullscreen",
-				"Expected 'fullscreen' or fullscreen <enable|disable|toggle>'");
+				"Expected 'fullscreen' or 'fullscreen <enable|disable|toggle>'");
 	}
 
 	view_set_fullscreen(view, wants_fullscreen);
--- a/sway/commands/fullscreen.c
+++ b/sway/commands/fullscreen.c
@@ -8,12 +8,6 @@
 
 // fullscreen toggle|enable|disable
 struct cmd_results *cmd_fullscreen(int argc, char **argv) {
-	struct cmd_results *error = NULL;
-	if (config->reading) return cmd_results_new(CMD_FAILURE, "fullscreen", "Can't be used in config file.");
-	if (!config->active) return cmd_results_new(CMD_FAILURE, "fullscreen", "Can only be used when sway is running.");
-	if ((error = checkarg(argc, "fullscreen", EXPECTED_AT_LEAST, 1))) {
-		return error;
-	}
 	struct sway_container *container =
 		config->handler_context.current_container;
 	if (container->type != C_VIEW) {
@@ -23,15 +17,15 @@ struct cmd_results *cmd_fullscreen(int argc, char **argv) {
 	struct sway_view *view = container->sway_view;
 	bool wants_fullscreen;
 
-	if (strcmp(argv[0], "enable") == 0) {
+	if (argc == 0 || strcmp(argv[0], "toggle") == 0) {
+		wants_fullscreen = !view->is_fullscreen;
+	} else if (strcmp(argv[0], "enable") == 0) {
 		wants_fullscreen = true;
 	} else if (strcmp(argv[0], "disable") == 0) {
 		wants_fullscreen = false;
-	} else if (strcmp(argv[0], "toggle") == 0) {
-		wants_fullscreen = !view->is_fullscreen;
 	} else {
 		return cmd_results_new(CMD_INVALID, "fullscreen",
-				"Expected 'fullscreen <enable|disable|toggle>'");
+				"Expected 'fullscreen' or fullscreen <enable|disable|toggle>'");
 	}
 
 	view_set_fullscreen(view, wants_fullscreen);
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -469,6 +469,12 @@ void output_damage_view(struct sway_output *output, struct sway_view *view,
 		return;
 	}
 
+	struct sway_container *workspace = container_parent(view->swayc,
+			C_WORKSPACE);
+	if (workspace->sway_workspace->fullscreen && !view->is_fullscreen) {
+		return;
+	}
+
 	struct damage_data data = {
 		.output = output,
 		.whole = whole,
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -22,6 +22,7 @@
 #include "sway/tree/container.h"
 #include "sway/tree/layout.h"
 #include "sway/tree/view.h"
+#include "sway/tree/workspace.h"
 
 struct sway_container *output_by_name(const char *name) {
 	for (int i = 0; i < root_container.children->length; ++i) {
@@ -275,9 +276,9 @@ static void render_output(struct sway_output *output, struct timespec *when,
 
 	struct sway_container *workspace = output_get_active_workspace(output);
 
-	if (workspace->fullscreen) {
+	if (workspace->sway_workspace->fullscreen) {
 		wlr_output_set_fullscreen_surface(wlr_output,
-				workspace->fullscreen->surface);
+				workspace->sway_workspace->fullscreen->surface);
 	} else {
 		wlr_output_set_fullscreen_surface(wlr_output, NULL);
 		render_layer(output,
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -229,7 +229,11 @@ static void render_container_iterator(struct sway_container *con,
 
 static void render_container(struct sway_output *output,
 		struct sway_container *con) {
-	container_descendants(con, C_VIEW, render_container_iterator, output);
+	if (con->type == C_VIEW) { // Happens if a view is fullscreened
+		render_container_iterator(con, output);
+	} else {
+		container_descendants(con, C_VIEW, render_container_iterator, output);
+	}
 }
 
 static struct sway_container *output_get_active_workspace(
@@ -277,10 +281,8 @@ static void render_output(struct sway_output *output, struct timespec *when,
 	struct sway_container *workspace = output_get_active_workspace(output);
 
 	if (workspace->sway_workspace->fullscreen) {
-		wlr_output_set_fullscreen_surface(wlr_output,
-				workspace->sway_workspace->fullscreen->surface);
+		render_container(output, workspace->sway_workspace->fullscreen->swayc);
 	} else {
-		wlr_output_set_fullscreen_surface(wlr_output, NULL);
 		render_layer(output,
 				&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND]);
 		render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]);
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -287,11 +287,9 @@ static void render_output(struct sway_output *output, struct timespec *when,
 		render_container(output, workspace);
 
 		render_unmanaged(output, &root_container.sway_root->xwayland_unmanaged);
-
-		render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]);
-		render_layer(output,
-				&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]);
 	}
+	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]);
+	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]);
 
 renderer_end:
 	if (root_container.sway_root->debug_tree) {
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -273,17 +273,25 @@ static void render_output(struct sway_output *output, struct timespec *when,
 	float clear_color[] = {0.25f, 0.25f, 0.25f, 1.0f};
 	wlr_renderer_clear(renderer, clear_color);
 
-	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND]);
-	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]);
-
 	struct sway_container *workspace = output_get_active_workspace(output);
-	render_container(output, workspace);
 
-	render_unmanaged(output, &root_container.sway_root->xwayland_unmanaged);
+	if (workspace->fullscreen) {
+		wlr_output_set_fullscreen_surface(wlr_output,
+				workspace->fullscreen->surface);
+	} else {
+		wlr_output_set_fullscreen_surface(wlr_output, NULL);
+		render_layer(output,
+				&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND]);
+		render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]);
 
-	// TODO: consider revising this when fullscreen windows are supported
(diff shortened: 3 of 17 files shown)


sway/desktop/output.c
```diff
@@ -273,17 +273,25 @@ static void render_output(struct sway_output *output, struct timespec *when,
...
+		render_container(output, workspace);
 
-	// TODO: consider revising this when fullscreen windows are supported
-	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]);
-	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]);
+		render_unmanaged(output, &root_container.sway_root->xwayland_unmanaged);
+
+		render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]);
+		render_layer(output,
+				&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]);
```
ascent12: I think these are still supposed to be rendered, even if there is a fullscreen surface.

atomnuker: I disagree, make it an option.

emersion: We discussed this on IRC and agreed we should add an option for this.

SirCmpwn: We shouldn't give people an option to skip the overlay layer when something is fullscreen, that's just a footgun waiting to happen (it would break lock screens, for example).

SirCmpwn: But we can give them a choice of putting fullscreen views between the shell layer and top layer, or between top and overlay.

sway/input/seat.c
```diff
@@ -448,6 +448,20 @@ void seat_set_focus_warp(struct sway_seat *seat,
...
 
+	struct sway_container *last_workspace = last_focus;
+	if (last_workspace && last_workspace->type != C_WORKSPACE) {
+		last_workspace = container_parent(last_workspace, C_WORKSPACE);
+	}
+	struct sway_container *new_workspace = container;
+	if (new_workspace && new_workspace->type != C_WORKSPACE) {
+		new_workspace = container_parent(new_workspace, C_WORKSPACE);
+	}
+
```
thejan2009: Would you consider changing `last_output` and `new_output` init a few lines below? Now that we have workspaces defined, these are just  one `new_output = container_parent(new_workspace, C_OUTPUT)` away.

include/sway/tree/container.h
```diff
@@ -72,9 +72,13 @@ struct sway_container {
 	// For C_OUTPUT, this is the output position in layout coordinates
 	// For other types, this is the position in output-local coordinates
 	double x, y;
+	double saved_x, saved_y;
 	// does not include borders or gaps.
 	double width, height;
 
+	// For C_WORKSPACE only
```
SirCmpwn: Then it should be moved to a new struct and stuck in the union at the top of this struct

include/sway/tree/container.h
```diff
@@ -72,9 +72,13 @@ struct sway_container {
 	// For C_OUTPUT, this is the output position in layout coordinates
 	// For other types, this is the position in output-local coordinates
 	double x, y;
+	double saved_x, saved_y;
```
SirCmpwn: I haven't really finished reading all of this but instead of saving the x/y it's probably better to just call arrange_windows again

sway/commands/fullscreen.c
```diff
@@ -0,0 +1,40 @@
...
+#include "sway/config.h"
+#include "sway/tree/container.h"
+#include "sway/tree/view.h"
+#include "sway/tree/layout.h"
+
+// fullscreen toggle|enable|disable
+struct cmd_results *cmd_fullscreen(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if (config->reading) return cmd_results_new(CMD_FAILURE, "fullscreen", "Can't be used in config file.");
+	if (!config->active) return cmd_results_new(CMD_FAILURE, "fullscreen", "Can only be used when sway is running.");
```
SirCmpwn: Braces are mandatory and these reading/active checks are now done at commands.c by placing it in the appropriate command block

sway/commands/fullscreen.c
```diff
@@ -0,0 +1,40 @@
...
+	if (strcmp(argv[0], "enable") == 0) {
+		wants_fullscreen = true;
+	} else if (strcmp(argv[0], "disable") == 0) {
+		wants_fullscreen = false;
+	} else if (strcmp(argv[0], "toggle") == 0) {
+		wants_fullscreen = !view->is_fullscreen;
+	} else {
+		return cmd_results_new(CMD_INVALID, "fullscreen",
+				"Expected 'fullscreen <enable|disable|toggle>'");
+	}
```
SirCmpwn: Please implement the sway behavior for `fullscreeen` so that we don't break people's stuff

sway/desktop/output.c
```diff
@@ -273,17 +273,25 @@ static void render_output(struct sway_output *output, struct timespec *when,
...
-	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND]);
-	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]);
-
 	struct sway_container *workspace = output_get_active_workspace(output);
-	render_container(output, workspace);
 
-	render_unmanaged(output, &root_container.sway_root->xwayland_unmanaged);
+	if (workspace->fullscreen) {
+		wlr_output_set_fullscreen_surface(wlr_output,
+				workspace->fullscreen->surface);
```
emersion: This will only work if the surface doesn't have subsurfaces (try right-clicking in Chromium or Firefo in fullscreen mode). However this can be left as a TODO.

sway/desktop/xwayland.c
```diff
@@ -25,6 +25,15 @@ static void unmanaged_handle_request_configure(struct wl_listener *listener,
 		ev->width, ev->height);
 }
 
+static void unmanaged_handle_request_fullscreen(struct wl_listener *listener,
+		void *data) {
+	struct sway_xwayland_view *xwayland_view =
+		wl_container_of(listener, xwayland_view, request_fullscreen);
```
emersion: Unmanaged surfaces aren't views, so this will fail. Unamanged surfaces are typically popups and menus, so I don't think we want to be able to put them fullscreen.

sway/tree/layout.c
```diff
@@ -137,6 +137,21 @@ void container_move_to(struct sway_container *container,
 			|| container_has_anscestor(container, destination)) {
 		return;
 	}
+
+	if (container->sway_view->is_fullscreen) {
```
emersion: Should check if the container is a view.

sway/tree/view.c
```diff
@@ -73,7 +74,46 @@ void view_set_activated(struct sway_view *view, bool activated) {
 	}
 }
 
+void view_set_fullscreen(struct sway_view *view, bool fullscreen) {
```
emersion: This function should work when the view is unmapped, i.e. when not visible on screen. In this case the view doesn't have a container yet.

Test case: launch mpv with the fullscreen flag.

include/sway/tree/container.h
acrisci: You're right in the tiling case, but It might be needed for the floating case.

SirCmpwn: Fair point. But in the tiling case we cannot assume that these coordinates will continue to be valid after unfullscreening.

sway/commands/fullscreen.c
```diff
@@ -0,0 +1,40 @@
...
+	if (config->reading) return cmd_results_new(CMD_FAILURE, "fullscreen", "Can't be used in config file.");
+	if (!config->active) return cmd_results_new(CMD_FAILURE, "fullscreen", "Can only be used when sway is running.");
+	if ((error = checkarg(argc, "fullscreen", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	struct sway_container *container =
+		config->handler_context.current_container;
+	if (container->type != C_VIEW) {
+		return cmd_results_new(CMD_INVALID, "fullscreen",
+				"Only views can fullscreen");
```
acrisci: Containers can be fullscreen too :D

sway/desktop/output.c
acrisci: Fullscreen should certainly hide the bar, right? This is the main difference between fullscreen and maximized.

i3 has [configuration for popups during fullscreen](https://i3wm.org/docs/userguide.html#_popups_during_fullscreen_mode) but that is for transient surfaces, which I don't think we have with xdg-shell.

The overlay should be above the fullscreen surface.

SirCmpwn: Fullscreen should indeed hide the bar. But the bar is in the bottom layer, so it'll work either way.

RyanDwyer:
@emersion Regarding the damage tracking, I'm concerned about performance. Every time a view is damaged we'd have to iterate up to the workspace to check if it contains a fullscreen view. While it would save having to rerender some parts when fullscreen, it'd be adding overhead for everything when not fullscreen.

sway/tree/view.c
```diff
@@ -73,7 +74,46 @@ void view_set_activated(struct sway_view *view, bool activated) {
...
+		view->saved_height = view->height;
+		view_configure(view, 0, 0, output->wlr_output->width, output->wlr_output->height);
+		workspace->fullscreen = view;
+	} else {
+		view_configure(view, view->swayc->saved_x, view->swayc->saved_y,
+				view->saved_width, view->saved_height);
+		workspace->fullscreen = NULL;
+	}
+
+	view->is_fullscreen = fullscreen;
```
acrisci: I don't think this is so simple. You have to wait until the surface commits with a buffer the size of the screen before you can actually set it to be fullscreen. Otherwise you will flash partially rendered content.

sway/input/seat.c
RyanDwyer: I tried this just now, but `last_workspace` can be null which means I'd have to make it conditional. It doesn't really improve the readability in that case.

include/sway/tree/container.h
RyanDwyer: I'd already removed these variables before seeing the latest two comments. I think it's best to leave them out for now and look at it once floating is implemented.

sway/desktop/output.c
emersion: >But the bar is in the bottom layer

This is true for swaybar, but would be false for a GNOME-like/Windows-like/KDE-like bar.

SirCmpwn: I mean, I think we should go between top and overlay by default for that reason. But I'm not gonna sweat it if someone wants to configure it to go elsewhere.

sway/tree/view.c
emersion: We're not doing the "frame perfect" milestone yet.

emersion:
>@emersion Regarding the damage tracking, I'm concerned about performance. Every time a view is damaged we'd have to iterate up to the workspace to check if it contains a fullscreen view. While it would save having to rerender some parts when fullscreen, it'd be adding overhead for everything when not fullscreen.

Damaging a view is very costly from the client side, it means drawing again and swapping buffers. Iterating from the view to the workspace is basically `while (container isn't a workspace) { go up; }`, which is done generally in less than 5 loop iterations. Applying damage from the view to the output is very costly: it means drawing areas that didn't change and swapping buffers. That is, not applying damage from views that are hidden is a win. Same applies to layer surfaces.

Additional comment: after rendering we send frame done events. We should only send frame done events to fullscreened views.

Also, X11 unmanaged surfaces should always be drawn if the fullscreened container contains X11 views. This is because some X11 clients don't set their popups parent to the X11 view (e.g. Chromium).

sway/commands/fullscreen.c
RyanDwyer: That requires replacing the call to `wlr_output_set_fullscreen_surface()` when rendering, and that's not something I want to do at this time.

SirCmpwn: Can you prepare a list of the limitations of your PR so we can track what work remains to be done in a separate ticket?

sway/tree/view.c
RyanDwyer: I don't think this function is called before the view is mapped, so that's not a problem.

I tested mpv with the fullscreen argument and realised that I need to adjust the `handle_map()` functions to check if the surface is requesting fullscreen while mapping. I've done this and mpv's fullscreen argument works now.

SirCmpwn:
Test setup: wayland backend with [this config](https://sr.ht/wl9o.txt)

Test issues:

* Pressing `f` on mpv does not fullscreen the video ([mpv config](https://sr.ht/3C80.txt))
* Moving a fullscreen view to another output un-fullscreens it; it should still be fullscreen there
* Moving a container to a workspace with a fullscreen view, then switching back to that workspace will have the moved container focused. You can then fullscreen _that_ container, with amusing results
* If you fullscreen an unfocused window (e.g. via `swaymsg "[title=gedit] fullscreen toggle`) it does not receive focus
* If you fullscreen a window on an unfocused workspace, I can't seem to figure out how to give it focus when returning to that workspace

Also tested a bunch of other stuff that worked fine.

sway/desktop/xdg_shell_v6.c
```diff
@@ -210,6 +224,18 @@ static void handle_destroy(struct wl_listener *listener, void *data) {
 	view_destroy(&xdg_shell_v6_view->view);
 }
 
+static void handle_request_fullscreen(struct wl_listener *listener, void *data) {
+	struct sway_xdg_shell_v6_view *xdg_shell_v6_view =
+		wl_container_of(listener, xdg_shell_v6_view, request_fullscreen);
+	struct wlr_xdg_toplevel_v6_set_fullscreen_event *e = data;
+
+	if (xdg_shell_v6_view->view.wlr_xdg_surface_v6->role != WLR_XDG_SURFACE_V6_ROLE_TOPLEVEL) {
```
SirCmpwn: This should probably be a sway_assert

SirCmpwn: Unless the protocol allows you to set popups as fullscreen surfaces... I don't think it does

sway/tree/layout.c
```diff
@@ -557,6 +572,9 @@ void arrange_windows(struct sway_container *container,
 		return;
 	case C_WORKSPACE:
 		{
+			if (container->sway_workspace->fullscreen) {
+				return;
```
SirCmpwn: We may still want to set the workspace w/h/x/y (which is done after the switch), not sure what the ramifications of not doing that will be

sway/tree/view.c
emersion: >I don't think this function is called before the view is mapped, so that's not a problem.

Are you sure about this? It seems odd to me. Are you starting mpv with the Wayland backend?

emersion:
One more comment: the input code in `cursor.c` needs to be updated not to allow surfaces under the fullscreen view to receive input events, even if the input region of the fullscreen view doesn't cover the whole surface.

sway/desktop/output.c
```diff
@@ -273,15 +274,21 @@ static void render_output(struct sway_output *output, struct timespec *when,
...
+		render_layer(output,
+				&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND]);
+		render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]);
+
+		render_container(output, workspace);
 
-	// TODO: consider revising this when fullscreen windows are supported
+		render_unmanaged(output, &root_container.sway_root->xwayland_unmanaged);
+	}
 	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]);
```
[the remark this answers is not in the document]
emersion: `wlr_output_set_fullscreen_surface` is not designed to be used with custom rendering code. It's designed to allow for direct scanout.

When swapping buffers, if a surface is fullscreened with `wlr_output_set_fullscreen_surface`, the `wlr_output` will clear the screen and render the fullscreen surface. Hence, any rendering that happened before will be discarded.

So the overlay layer cannot be rendered for now. This should be added to the list of limitations of this PR.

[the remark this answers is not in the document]
SirCmpwn: I'm not going to merge a PR that breaks swaylock.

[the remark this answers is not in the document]
emersion: Then maybe don't use `wlr_output_set_fullscreen_surface` for now? Doing the rendering manually isn't that hard. We just need to be careful about surfaces that are smaller than the output size.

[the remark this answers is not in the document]
SirCmpwn: Probably also not too hard to continue to use wlr_output_set_fullscreen_surface if there isn't anything in the overlay layer at the time

SirCmpwn:
On this branch, closing qutebrowser context menus segfaults sway.

emersion:
>>This causes my code to read the wrong dimensions for a fullscreen view on a rotated output.
>
>This is addressed by getting the width/height via wlr_output_effective_resolution

The preferred way to do this is to get this info from the `C_OUTPUT`.

sway/commands/fullscreen.c
SirCmpwn: bump

sway/tree/layout.c
SirCmpwn: bump

sway/tree/layout.c
```diff
@@ -930,6 +996,28 @@ struct sway_container *container_get_in_direction(
 	}
 }
 
+struct sway_container *container_get_in_direction(
+		struct sway_container *container, struct sway_seat *seat,
+		enum movement_direction dir) {
+	struct sway_container *result = container_get_in_direction_naive(container,
+			seat, dir);
```
SirCmpwn: Not sure splitting this is necessary

sway/tree/view.c
SirCmpwn: bump

sway/commands/fullscreen.c
RyanDwyer: * Containers need to be fullscreenable.
* Surfaces can request to be fullscreened on a specific output, which is currently ignored.
* Fullscreen is not frame perfect.

I think that's it.

atomnuker: That's kind of a big issue. Would fixing it take a lot of work?

SirCmpwn: It's part of a larger epic, we can keep this patch smaller and more focused and deal with it later.

emersion: Basically nothing is pixel perfect right now. When resizing a view you'll get background artifacts. When closing a view, same. Fixing all of these requires #1650.

sway/tree/view.c
emersion: Oh, you need the not-yet-released mpv (from Git), this is annoying. Alternatively, you can use my test client here: https://gist.github.com/emersion/34ed72adecd78bdfac9208a97ffb46a3

sway/tree/layout.c
RyanDwyer: Resolved in 45a2fad.

sway/tree/view.c
RyanDwyer: Resolved in 2acfbc5.

sway/tree/layout.c
RyanDwyer: Resolved in 7e38cc2.

sway/input/seat.c
thejan2009: I'd then settle on initially assigning `new_workspace` to `new_output`, `last_workspace` to `last_output` and leaving the `container_parent` calls inside if clauses.

SirCmpwn:
qutebrowser crash:

```
#0  0x000055f660154ba7 in seat_set_focus_warp (seat=0x55f660d485e0, container=0x0, warp=true) at ../sway/input/seat.c:462
#1  0x000055f660155121 in seat_set_focus (seat=0x55f660d485e0, container=0x0) at ../sway/input/seat.c:574
#2  0x000055f66015614b in dispatch_cursor_button (cursor=0x55f660d3dff0, time_msec=1433127, button=272, state=WLR_BUTTON_PRESSED) at ../sway/input/cursor.c:211
#3  0x000055f6601561c4 in handle_cursor_button (listener=0x55f660d3e040, data=0x7fffae5f6860) at ../sway/input/cursor.c:221
#4  0x00007fca2a81b9d9 in wlr_signal_emit_safe (signal=0x55f660d3e168, data=0x7fffae5f6860) at ../subprojects/wlroots/util/signal.c:29
#5  0x00007fca2a7f9b82 in handle_pointer_button (listener=0x55f661174e90, data=0x7fffae5f6860) at ../subprojects/wlroots/types/wlr_cursor.c:344
#6  0x00007fca2a81b9d9 in wlr_signal_emit_safe (signal=0x55f661162ae8, data=0x7fffae5f6860) at ../subprojects/wlroots/util/signal.c:29
#7  0x00007fca2a7e9dae in handle_pointer_button (event=0x55f6611ac790, libinput_dev=0x55f66108fae0) at ../subprojects/wlroots/backend/libinput/pointer.c:84
#8  0x00007fca2a7e968a in wlr_libinput_event (backend=0x55f660b62200, event=0x55f6611ac790) at ../subprojects/wlroots/backend/libinput/events.c:217
#9  0x00007fca2a7e85ad in wlr_libinput_readable (fd=27, mask=1, _backend=0x55f660b62200) at ../subprojects/wlroots/backend/libinput/backend.c:35
#10 0x00007fca28d209b2 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#11 0x00007fca28d1f14c in wl_display_run () at /usr/lib/libwayland-server.so.0
#12 0x000055f660145762 in server_run (server=0x55f66037f360 <server>) at ../sway/server.c:141
#13 0x000055f660145058 in main (argc=4, argv=0x7fffae5f6c58) at ../sway/main.c:418
```

SirCmpwn:
>Also, moving fullscreen window to another output still un-fullscreens it.

We can take this to a separate patch

@emersion merge it if you like it

emersion:
Sway segfaults with this client: https://gist.github.com/emersion/4b5aa6df269daf8e53b900eef0c11be9

This is because `request_fullscreen` can be called when the client is not yet mapped. You should ignore requests if `!xdg_shell_v6_view->view.wlr_xdg_surface_v6->mapped` (because they're handled by the `map` handler).

sway/desktop/output.c
```diff
@@ -270,18 +275,24 @@ static void render_output(struct sway_output *output, struct timespec *when,
...
-	float clear_color[] = {0.25f, 0.25f, 0.25f, 1.0f};
-	wlr_renderer_clear(renderer, clear_color);
+	struct sway_container *workspace = output_get_active_workspace(output);
 
-	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND]);
-	render_layer(output, &output->layers[ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM]);
+	if (workspace->sway_workspace->fullscreen) {
+		float clear_color[] = {0.0f, 0.0f, 0.0f, 1.0f};
+		wlr_renderer_clear(renderer, clear_color);
+		render_container(output, workspace->sway_workspace->fullscreen->swayc);
```
emersion: Can you add a

```c
// TODO: handle views smaller than the output
```

sway/desktop/output.c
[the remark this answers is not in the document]
emersion: >top layer needs to be below the fullscreen surface

bump

sway/desktop/output.c
RyanDwyer: Resolved in 6afccd0.

sway/desktop/output.c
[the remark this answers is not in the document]
RyanDwyer: Resolved in bd67342.

PR #1821, opened 2018-04-18
Swaylock: Allow per-output images

RyanDwyer:
This implements the output part of the `[<output>:]<path>` syntax for the `-i` argument.

The previous implementation of this PR used IPC, but that implementation has been replaced with one which uses the xdg-output protocol.

Tests:

* Specifying no image arguments should show blank on all outputs
* Specifying `-i ~/default.png` should show default.png on all outputs
* Specifying `-i HDMI-A-1:~/a.png` should show a.png on that output, and all others blank
* Specifying `-i HDMI-A-1:~/a.png -i ~/default.png` should show default.png on all other outputs

Unusual scenario tests:

* Specifying the same output (or lack of output) multiple times should pick one of them
* Specifying an image that is unreadable should log an error and behave as though you didn't pass that argument


SirCmpwn:
I want to wait until the xdg-output discussion on wayland-devel is resolved so we can do this without sway-specific IPC.

SirCmpwn:
https://github.com/swaywm/wlroots/issues/961

SirCmpwn:
xdg-output v2 is available in the Arch repos now and supported by wlroots.

RyanDwyer:
Okay, it's using the xdg-output protocol now. Ready for review again.

protocols/xdg-output-unstable-v1.xml
```diff
@@ -0,0 +1,206 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
emersion: Please don't embed the protocol in Sway, use the XML file from wayland-protocols instead.

swaylock/main.c
```diff
@@ -84,34 +86,70 @@ static const struct zwlr_layer_surface_v1_listener layer_surface_listener = {
 	.closed = layer_surface_closed,
 };
 
-static void output_geometry(void *data, struct wl_output *output, int32_t x,
+static void wl_output_geometry(void *data, struct wl_output *output, int32_t x,
```
emersion: The `wl_` prefix is reserved for the Wayland lib

swaylock/main.c
```diff
@@ -133,13 +171,23 @@ static void handle_global(void *data, struct wl_registry *registry,
...
 	} else if (strcmp(interface, wl_output_interface.name) == 0) {
 		struct swaylock_surface *surface =
 			calloc(1, sizeof(struct swaylock_surface));
 		surface->state = state;
 		surface->output = wl_registry_bind(registry, name,
 				&wl_output_interface, 3);
-		wl_output_add_listener(surface->output, &output_listener, surface);
+		wl_output_add_listener(surface->output, &wl_output_listener, surface);
+
+		struct zxdg_output_v1 *xdg_output =
```
SirCmpwn: This will break if the server advertises wl_outputs before advertising xdg_output_manager, or if the server doesn't support xdg_output at all. You should poll xdg_outputs after the registry roundtrip.

swaylock/main.c
```diff
@@ -154,6 +202,60 @@ static const struct wl_registry_listener registry_listener = {
...
+		*separator = '\0';
+		image->output_name = arg;
+		image->path = separator + 1;
+	} else {
+		image->output_name = NULL;
+		image->path = arg;
+	}
+
+	// Bash doesn't replace the ~ with $HOME if the output name is supplied
+	if (image->path[0] == '~') {
```
SirCmpwn: How about just using wordexp here

swaylock/main.c
```diff
@@ -154,6 +196,53 @@ static const struct wl_registry_listener registry_listener = {
...
+		} else if (!image->output_name) {
+			default_image = image->cairo_surface;
+		}
+	}
+	return default_image;
+}
+
+static void load_image(char *arg, struct swaylock_state *state) {
+	// [<output>:]<path>
+	struct swaylock_image *image = malloc(sizeof(struct swaylock_image));
```
emersion: Please use `calloc` for structs

swaylock/main.c
```diff
@@ -212,7 +212,7 @@ static cairo_surface_t *select_image(struct swaylock_state *state,
 
 static void load_image(char *arg, struct swaylock_state *state) {
 	// [<output>:]<path>
-	struct swaylock_image *image = malloc(sizeof(struct swaylock_image));
+	struct swaylock_image *image = calloc(sizeof(struct swaylock_image), 1);
```
ascent12: This doesn't _actually_ affect anything, but the arguments are in the wrong order.

emersion:
Thanks!

PR #1823, opened 2018-04-18
Make key repeat configurable

RyanDwyer:
This creates two input commands for configuring the repeat delay and rate.

    input "myidentifier" {
        repeat_delay 250
        repeat_rate 25
    }

Tests (use DRM backend):

* Start sway with variations of the above config
* swaymsg input myidentifier repeat_delay 1000
* swaymsg input myidentifier repeat_rate 10

<s>Attempting to configure it via IPC doesn't work for some reason. It passes the correct values to wlr_keyboard_set_repeat_info() but it doesn't take effect. Is there anything else I need to do or is that a bug in wlroots?</s> (EDIT: This is because I was testing via the X11 backend)


sway/commands/input/repeat.c
```diff
@@ -0,0 +1,55 @@
...
+	int repeat_rate = atoi(argv[0]);
+	if (repeat_rate < 0) {
+		return cmd_results_new(CMD_INVALID, "repeat_rate",
+			"Repeat rate cannot be negative");
+	}
+	new_config->repeat_rate = repeat_rate;
+
+	apply_input_config(new_config);
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+}
```
SirCmpwn: Split this into two files

SirCmpwn:
>Is there anything else I need to do or is that a bug in wlroots?

Sounds like bug in wlroots maybe

sway/commands/input/repeat.c
RyanDwyer: Resolved in 9d3739a.

RyanDwyer:
I've discovered there's actually nothing wrong with changing the repeat settings via swaymsg. I'm ashamed to admit I'm still using i3 (I have rotated outputs) and am doing most of my testing by using the X11 backend. I think X was controlling the repeat settings that it sent to my sway window. Changing the repeat settings via swaymsg using the drm backend works fine.

SirCmpwn:
>I'm ashamed to admit I'm still using i3 (I have rotated outputs)

Try the alpha!

emersion:
Thanks!

Issue #1820, opened 2018-04-18
swaybar leaks memory


atomnuker:
To replicate: run swaybar with valgrind and change workspaces a few times. Running swaybar for a few hours of normal usage can easily make it leak hundreds of megabytes of memory.
A short abstract after 20 seconds of changing workspaces:

>==3318== 10,800 bytes in 90 blocks are definitely lost in loss record 935 of 958
==3318==    at 0x4C2EEF5: calloc (vg_replace_malloc.c:711)
==3318==    by 0x10EDE4: i3bar_parse_json (i3bar.c:61)
==3318==    by 0x10F4B6: i3bar_handle_readable (i3bar.c:166)
==3318==    by 0x11334F: status_handle_readable (status_line.c:26)
==3318==    by 0x10E21B: status_in (bar.c:421)
==3318==    by 0x10E9BA: event_loop_poll (event_loop.c:122)
==3318==    by 0x10E2C0: bar_run (bar.c:433)
==3318==    by 0x110DD9: main (main.c:104)
==3318== 
==3318== 12,853 (1,024 direct, 11,829 indirect) bytes in 2 blocks are definitely lost in loss record 937 of 958
==3318==    at 0x4C2F0FF: realloc (vg_replace_malloc.c:785)
==3318==    by 0x99E4799: ??? (in /usr/lib/libfontconfig.so.1.11.1)
==3318==    by 0x99E4E9E: ??? (in /usr/lib/libfontconfig.so.1.11.1)
==3318==    by 0x99D6EEA: FcDefaultSubstitute (in /usr/lib/libfontconfig.so.1.11.1)
==3318==    by 0xB832855: ??? (in /usr/lib/libpangoft2-1.0.so.0.4200.1)
==3318==    by 0x5DC8EE8: ??? (in /usr/lib/libpango-1.0.so.0.4200.1)
==3318==    by 0x5DC9CD7: pango_itemize_with_base_dir (in /usr/lib/libpango-1.0.so.0.4200.1)
==3318==    by 0x5DCADBB: ??? (in /usr/lib/libpango-1.0.so.0.4200.1)
==3318==    by 0x5DCCBEB: ??? (in /usr/lib/libpango-1.0.so.0.4200.1)
==3318==    by 0x5DCDFB3: pango_layout_get_pixel_extents (in /usr/lib/libpango-1.0.so.0.4200.1)
[...]

PR #1827, opened 2018-04-19
Fix memory leak on swaybar

bpinto:
Code from @NilsBrause

Ref #1820

### DIFF
--- a/swaybar/i3bar.c
+++ b/swaybar/i3bar.c
@@ -104,6 +104,7 @@ static bool i3bar_parse_json(struct status_line *status, const char *text) {
 			json_object_get_int(border_right) : 1;
 		wl_list_insert(&status->blocks, &block->link);
 	}
+	json_object_put(results);
 	return true;
 }
 


swaybar/i3bar.c
```diff
@@ -104,6 +104,7 @@ static bool i3bar_parse_json(struct status_line *status, const char *text) {
 			json_object_get_int(border_right) : 1;
 		wl_list_insert(&status->blocks, &block->link);
 	}
+  json_object_put(results);
```
ascent12: Wrong indentation.

bpinto: In a single line commit I was able to get an issue... that's a new mark in my life :joy: 

Fixed it, I have my vim configured to use spaces and not tabs. Got it to use a tab now.

bpinto:
@martinetd :100: 

I am not a C programmer and I don't even know (yet) how to get information about leaks - I might try to work on these other leaks you referenced. It was a surprise to me that a method named `json_object_put` would do anything to free memory. :joy: 

I just noticed that code was there on a comment and felt I could do something other than ignore it by packaging that single-line patch into a PR. And here it is, after an indentation issue. :grimacing:

emersion:
3 approvals for a one-line fix, I can't leave this unmerged.

Thanks!

PR #1829, opened 2018-04-19
Keep view fullscreened when moving to another workspace

RyanDwyer:
`container_handle_fullscreen_reparent()` must be called *after* setting the container's new parent.

To test:

* Fullscreen a view, then move the view to another workspace, then switch to that workspace - the view should be fullscreen.


emersion:
Thanks!

PR #1830, opened 2018-04-20
[swaylock] Install pam module

bpinto:


emersion:
Can you rebase your patch against `master`?

bpinto:
@emersion too late. I would have otherwise. 👍 

Thanks for merging!

Issue #541, opened 2016-03-25
DPMS


crondog:
Would it be possible for sway to handle turning on and off the monitor? Orbment currently has https://github.com/Cloudef/orbment/blob/master/plugins/core-dpms/core-dpms.c to handle this. 

Could something like this be implemented inside sway with a config option or maybe with swaylock?

SirCmpwn:
I'd like to see a protocol extension written for this, and then a seperate daemon to manage it.

crondog:
Hmm interesting. That sounds like a better idea. I'll need to look into protocol extensions over the weekend and report back

SirCmpwn:
https://github.com/SirCmpwn/sway/tree/master/protocols

They're pretty straightfoward. Write an XML file describing the interface, add it to CMakeLists.txt, and you'll end up with a header to include.

SirCmpwn:
See also https://github.com/SirCmpwn/sway/blob/master/sway/extensions.c

crondog:
Thanks for the tip

Fale:
any news on this?

SirCmpwn:
Nope.

Cloudef:
While orbment's core-dpms implements this using own timer and wlc's set_sleep. I think drm has dpms api too actually.

hosaka:
Looking at DRM docs, there is an api for setting dpms on drm_connector, but no way of getting current state. Would using wlc set_sleep approach be preferable?

airstep:
The question is how to make it possible? 

For now I can advice to switch to tty(n) and it works by default - screen is going off by some time. Why I can't do this in wayland? 

My tty(n) is also in KMS mode (like wayland)... 

So how tty(n) can turn screen off, but wayland can't - if them both in KMS mode? 

What you think?

Fale:
I think the situation around this could/should/will change very soon due to #1076

flounders:
I see on the status issue for for wlroots ([wlroots/#9](https://github.com/swaywm/wlroots/issues/9)) that DPMS is checked off. I also see that DPMS is an unaddressed bounty problem as well. In a general, overview type answer, given the support in wlroots what needs to happen to bring this to Sway? This question is coming from someone not too familiar with the project or Wayland compositors. If it's not too far beyond my skill set, I wouldn't mind working on this as I am able.

SirCmpwn:
It's not currently a priority in the wlroots port. Let's swing back to it later.

PR #1831, opened 2018-04-20
Another version of idle and lock handlng by doing it in the compositor

snaggen:
Based on the discussions in https://github.com/swaywm/sway/pull/1824 I created another version where everything is handled by the compositor directly. 
So this does the same thing, namely adding dpms handling and lockscreen handling, but done within the compositor it self. 

Fixes https://github.com/swaywm/sway/issues/541

Depends on wlroots pull request:
https://github.com/swaywm/wlroots/pull/879


somdoron:
I'm excited about this PR, it is so hard to get auto-lock under X.

Wonder what do you think about also catching the "before suspend" from systemd (or any other daemon) to also lock the screen?

SirCmpwn:
>Wonder what do you think about also catching the "before suspend" from systemd (or any other daemon) to also lock the screen?

Another thing that would be great to put into the daemon instead of into sway itself

somdoron:
xss-lock can actually do this without systemd service.

https://bitbucket.org/raymonad/xss-lock

https://www.mankier.com/1/xss-lock

somdoron:
what about `PrepareForSleep` from logind dbus signal?

https://www.freedesktop.org/wiki/Software/systemd/logind/

somdoron:
and this can also be a solution:

https://www.freedesktop.org/wiki/Software/systemd/inhibit/

SirCmpwn:
>Maybe swaylock should install this by default

No

snaggen:
After looking in to this, wlroots already listens to a bunch of logind signals in its logind session code. 
wlroots/backend/session/logind.c
So maybe the PrepareForSleep should best be handled within that session code somehow since it is tied to logind. Anyway, it feels it is getting a little out of scope for this kind of pull request.

snaggen:
On a second look, it wasn't that hard to add it, since we already hade the dbus eventloop up and running due to the logind integration.

somdoron:
do you want to block the sleep until you sure swaylock is running?

somdoron:
Also maybe it should be an option

somdoron:
I tested it on my machine, works great! I never managed to have a good setup for locking my screen with i3 and this just works out of the box.

few comments:
- Dependency to systemd is missing
- Because you are not blocking the sleep operation, the swaylock is actually become active after the computer is awaken, so I have a glimpse of my desktop when awaken.

emersion:
Hmm, as @SirCmpwn said, we probably don't systemd-specific stuff in the compositor. However the systemd stuff doesn't require the idle protocol and can already be done outside the compositor. So I'm still not 100% convinced by the idle protocol approach.

SirCmpwn:
I'm still in favor of the external daemon approach. Can those who are still in favor of the in-the-compositor approach summarize their reasons succinctly in a comment in this thread?

SirCmpwn:
Rationale for external daemon: Daemon can implement complex situations (e.g. dim screen after n seconds, then lock, then DPMS, then suspend) or simple ones and be useful on many compositors. It also makes us expose DPMS to clients who might want to do novel things with it, and it keeps systemd-specific code out of sway.

I don't think the complexity add is much.

somdoron:
I actually ambivalent to the solution, however I like the configuration through sway config file. 
If it can work like swaybar that will be nice (can be configured within sway configuration), but not a must.

Also one thing that I think will be easier to do with in the compositor is handling new outputs, like if I connect my laptop to a docking station, the new output should display the lock screen immediately,  without even a glimpse of my desktop. Same for new keyboard and mouse.

SirCmpwn:
I'm open to having sway support configuring this daemon in your config file and launching it from there (though I don't want it to work like the bar, i.e. by sending the config over IPC).

SirCmpwn:
>Also one thing that I think will be easier to do with in the compositor is handling new outputs, like if I connect my laptop to a docking station, the new output should display the lock screen immediately, without even a glimpse of my desktop. Same for new keyboard and mouse.

This is unrelated to idle, more of a swaylock problem (and swaylock isn't being moved into the compositor any time soon). We already handle the case of new keyboards/mice immediately being restricted if the screen is locked, though.

ascent12:
This whole logind thing just screams that it should be implemented in wlroots, not here. wlr_session was designed to hide these systemd vs whatever else differences, and this whole business is pretty useless without the DRM backend anyway, since we wouldn't be the session owner.

In terms of the external daemon, I just don't see the value in the slightest. What "complex idle behaviour" are you even talking about? What is someone going to do beyond "dim, lock, blank"? The timings might be different, but that's simple configuration.
Anything beyond that is configuration to the point of absurdity.

And as I said in the other issue, **this is not a hard problem**. The bulk of this PR is very simple. I don't see why you're trying to solve such simple problems for other compositors, and in the wrong place to begin with (i.e. not wlroots).

SirCmpwn:
wlroots can definitely implement the (e)logind side of things. Whatever.

The rest is opinionated, and wlroots isn't allowed to be opinionated.

snaggen:
But go with which ever version you like and fits in to the sway design best, if you choose to go with https://github.com/swaywm/sway/pull/1824 I suggest to merge that as is feature wise, then we can add lockscreen detection IPC and systemd suspend stuff later.

emersion:
>since we already have the same kind of things running already in the logind code I don't see why it would be so terrible to have it in the idle code in the compositor

The logind code is only used with the DRM backend. If the Wayland backend is used, we don't want to react to the parent session suspend event. Plus, we don't like systemd-specific code inside the compositor.

It's true that with an external client the lockscreen is going to be launched multiple times (and likely fail thanks to `input-inhibit`). Also we'll need to add a special case for the lockscreen anyway when we'll implement security feature (we'll want to expose `input-inhibit` only to this client).

snaggen:
Ok, I have moved the systemd related parts to the logind backend, so please have a look at how you think the interface to this is. Also, please test this and see if the inhibit part works as expected. I tend to get error on the Inhibit dbus call, seeing "Failed to send Inhibit signal: Operation now in progress" in the logs. Is the wayland eventloop taking to long, so that the prepare for sleep signal comes to late for an inhibit to be made? It would be good if someone with deeper knowledge about this could take a look. 
Also, a second opinion on the have_lock() function, currently it is a kind of brute force approach to this, is there a more efficient and simpel way to find the lockscreen surface?

somdoron:
@snaggen you assume the prepare to sleep will be called multiple times, but from the docs it seems it will be once for sleep and one for awake. Also from the docs the delay lock should be taked before the prepare for sleep, released when you are ready and retake after the system is awake.

You are currently using the block semantic, which I actually think is more accurate in this scenario. However you must close the fd or it will prevent the system from sleeping.

somdoron:
You might want to have a timer after the prepare for sleep, checking the if the lock is active and then release the lock.

Also prepare for sleep is being call twice, once before the sleep and once after, I think you are not checking that at the moment, so the lock screen will actually be invoked twice.

agx:
I had similar stuff within our compositor first and moved it into a separate process using idle, layer-shell and input-inhibitor protocols. Main reason was to keep all the policy decisions (what to do on which event) out of the compositor.

snaggen:
Completely misread the docs... must update this to work as you say and using the block lock. Saw at the end of the document, it was stated explicitly to use block for lockscreen... my bad.

snaggen:
Ok, new version available with block and that only inhibits when going down. Using timers to check for lock... so this should be more correct according to the docs. Please have a look.

somdoron:
@snaggen looks good.

Few thoughts regarding finding out if lock screen is already active:
1. You can check if the input-inhibitor has been set.
2. You can have a global mutex, derive his name from the session id or wayland display environment variable, and check if the mutex has been acquired/acquire it from swaylock. This solution might be too specific for swaylock, and I'm not sure it's the wayland way, but it will be simple.
3. Extended the input-inhibitor protocol to lock-protocol, swaylock will send the lock message when needed. This can also help solve the problem of connecting new outputs while the lock screen is active. On lock mode, compositor might only draw lock screen surfaces, so worst case scenario new output will have black screen.

somdoron:
I just tested it with the delay lock taken when registering to prepare for sleep and the swaylock is displayed before the computer is going to sleep.

snaggen:
Thanks for the help testing and solving this. Will try to find some time to fix this during the weekend.

somdoron:
@snaggen I tested both solutions, the in compositor and daemon. Surprisingly only the daemon solution is able to acquire the delay lock. I don't know why, but with the compositor solution the lock is not being acquired. You can see that by running systemd-inhibit after logging in to sway. Only with daemon solution you actually see the lock. Also only the daemon solution is able to lock the screen before the sleep. I have no clue why this is happening. I guess it is answering the question of daemon vs compositor.

Also I think the daemon solution might be better place to implement something like global mutex to know if the swaylock is actually running.

snaggen:
That is really weird... also if you look at the debug output you will see that we actually get a lock file descriptor back even in the compositor case so it really looks like everything is find and dandy, but then it is still not listed in systemd-inhibit. So yes, as you say, that probably settles the whole in compositor vs daemon discussion!
Also, I have been thinking a bit about this. And this is mostly policy, and this kind of policy should not be handled by a compositor. Also, when you answering the question "what is a compositor?" you really do not think about polling systemd for sleep event.

SirCmpwn:
Closing in favor of #1824

PR #1835, opened 2018-04-21
Swaybar fix

ascent12:
Remove status command fd from event loop on error.
This prevents poll constantly looping on POLLHUP if the status command dies, causing 100% CPU usage.

### DIFF
--- a/swaybar/event_loop.c
+++ b/swaybar/event_loop.c
@@ -72,24 +72,18 @@ void add_event(int fd, short mask,
 }
 
 bool remove_event(int fd) {
-	int index = -1;
+	/*
+	 * Instead of removing events immediately, we mark them for deletion
+	 * and clean them up later. This is so we can call remove_event inside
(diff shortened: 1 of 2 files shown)


swaybar/event_loop.c
```diff
@@ -118,11 +112,29 @@ void event_loop_poll() {
 		struct pollfd pfd = event_loop.fds.items[i];
 		struct event_item *item = (struct event_item *)event_loop.items->items[i];
 
-		if (pfd.revents & pfd.events) {
+		// Always send these events
+		unsigned events = pfd.events | POLLHUP | POLLERR;
```
SirCmpwn: -1

ascent12: This follows the semantics of poll. You get those events, even if you don't ask for them.
That's actually what's causing the issue.

SirCmpwn: Ah.

PR #1836, opened 2018-04-21
Update cursor when workspace focus changes

emersion:
Not only this updates the cursor image, but this also (and more importantly) sends pointer enter/leave events.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -564,6 +564,12 @@ void seat_set_focus_warp(struct sway_seat *seat,
 		view_set_activated(view, false);
 	}
 
+	if (last_workspace && last_workspace != new_workspace) {
+		struct timespec now;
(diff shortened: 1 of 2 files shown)


sway/input/seat.c
```diff
@@ -564,6 +564,12 @@ void seat_set_focus_warp(struct sway_seat *seat,
 		view_set_activated(view, false);
 	}
 
+	if (last_workspace && last_workspace != new_workspace) {
+		struct timespec now;
+		clock_gettime(CLOCK_MONOTONIC, &now);
+		cursor_send_pointer_motion(seat->cursor, now.tv_nsec / 1000);
```
emersion: We do this in a lot of places. Maybe we should make zero a special value that gets the current time?

SirCmpwn: Neat idea, I like it

PR #1840, opened 2018-04-22
Fix swaybar axis event logic

### LINKED ISSUE #1807: Crash in swaybar
Random swaybar crash when scrolling on it. I wasn't performing any other action at the time.
It's happened a few times before, but I haven't been able to reliably reproduce it.
```
#0  0x00007f2514fc4860 in raise () at /usr/lib/libc.so.6
#1  0x0000559db5a0543f in _sway_assert (condition=false, format=0x559db5a06ed0 "[%s:%d] %s:axis with null workspace") at ../common/log.c:27
#2  0x0000559db59fe7ae in wl_pointer_axis (data=0x559db5c0b820 <swaybar>, wl_pointer=0x559db5c577b0, time=50423815, axis=0, value=-3840) at ../swaybar/bar.c:170
#3  0x00007f25128e61c8 in ffi_call_unix64 () at /usr/lib/libffi.so.6
#4  0x00007f25128e5c2a in ffi_call () at /usr/lib/libffi.so.6
#5  0x00007f25157e0bad in  () at /usr/lib/libwayland-client.so.0
#6  0x00007f25157dd679 in  () at /usr/lib/libwayland-client.so.0
#7  0x00007f25157de9b4 in wl_display_dispatch_queue_pending () at /usr/lib/libwayland-client.so.0
#8  0x0000559db59ff175 in display_in (fd=6, mask=1, _bar=0x559db5c0b820 <swaybar>) at ../swaybar/bar.c:406
#9  0x0000559db59ff99b in event_loop_poll () at ../swaybar/event_loop.c:122
#10 0x0000559db59ff2a1 in bar_run (bar=0x559db5c0b820 <swaybar>) at ../swaybar/bar.c:433
#11 0x0000559db5a01dbd in main (argc=3, argv=0x7ffecc2b1998) at ../swaybar/main.c:104
```
Workspaces:
```
Workspace 1 (off-screen)
  Output: DP-2
  Layout: splith

Workspace 3 (off-screen)
  Output: DP-2
  Layout: splith

Workspace 4 (off-screen)
  Output: DP-2
  Layout: splith

Workspace 5 (focused)
  Output: DP-2
  Layout: splith

Workspace 2
  Output: DVI-I-1
  Layout: splitv

Workspace 10 (off-screen)
  Output: DVI-I-1
  Layout: splitv
```
There are 2 instances of swaybar, one on DP-2 and one on DVI-I-1. The DVI-I-1 is the one that crashed, and I was on workspace 10 at the time.

Relevant parts of config:
```
output DP-2 resolution 2560x1440 position 0,96
output DVI-I-1 resolution 1920x1200 position 2560,0 transform 270

bar {
    output DVI-I-1
    position top
    status_command i3status
}
bar {
    status_command cat /dev/null
    output DP-2
}
```
Everything else is basically default. The 270 transform on the output may be relevant.

```
$2 = {link = {prev = 0x559db967af80, next = 0x559db5c66188}, num = 2, name = 0x559db9671a30 "2", focused = false, visible = false, urgent = false}
$4 = {link = {prev = 0x559db5c66188, next = 0x559db966f700}, num = 10, name = 0x559db9660f50 "10", focused = false, visible = true, urgent = false}
```
So apparently, it's possible to get into a state where no workspace is focused.

I'm not sure if it's related, but something I noticed was that very rarely, my mouse cursor would disappear for a frame as I was moving it along the bar. I was not taking it off the bar surface, but it looked the same as the cursor blip when moving between two surfaces.

ascent12:
Fixes #1807 

Uses 'visible' instead of 'focused', since we may scroll on a bar which
isn't the focused output.

We can't use "next_on_output" or "prev_on_output" to implement this,
because it only modify the focused output. So scrolling on an unfocused
output will affect the incorrect one.

We just use the "workspace name" command instead.

Test (requires multiple outputs):
- Set `focus_follows_mouse no` in sway config
- Focus a window on output A
- Scroll on the swaybar on output B

Notice no crash and the workspace change is on output B.

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -156,31 +156,56 @@ static void wl_pointer_axis(void *data, struct wl_pointer *wl_pointer,
 	if (!sway_assert(output, "axis with no active output")) {
 		return;
 	}
+
 	double amt = wl_fixed_to_double(value);
-	if (!bar->config->wrap_scroll) {
-		int i = 0;
-		struct swaybar_workspace *ws = NULL;
-		wl_list_for_each(ws, &output->workspaces, link) {
-			if (ws->focused) {
-				break;
-			}
-			++i;
-		}
-		int len = wl_list_length(&output->workspaces);
-		if (!sway_assert(i != len, "axis with null workspace")) {
-			return;
-		}
-		if (i == 0 && amt > 0) {
-			return; // Do not wrap
-		}
-		if (i == len - 1 && amt < 0) {
-			return; // Do not wrap
-		}
+	if (amt == 0.0) {
+		return;
 	}
 
-	const char *workspace_name =
-		amt < 0 ?  "prev_on_output" : "next_on_output";
-	ipc_send_workspace_command(bar, workspace_name);
+	struct swaybar_workspace *first = NULL;
+	struct swaybar_workspace *active = NULL;
+	struct swaybar_workspace *last;
+
+	struct swaybar_workspace *iter;
+	wl_list_for_each(iter, &output->workspaces, link) {
+		if (!first) {
+			first = iter;
+		}
+
+		if (iter->visible) {
+			active = iter;
+		}
+
+		last = iter;
+	}
+
+	if (!sway_assert(active, "axis with null workspace")) {
+		return;
+	}
+
+	struct swaybar_workspace *new;
+
+	if (amt > 0.0) {
+		if (active == first) {
+			if (!bar->config->wrap_scroll) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
>since we may scroll on a bar which isn't the focused output.

How?

SirCmpwn:
>Set focus_follows_mouse no in sway config

Ah

SirCmpwn:
Thanks!

Issue #1828, opened 2018-04-19
Fullscreen improvements


RyanDwyer:
- [x] Containers need to be fullscreenable.
- [ ] Surfaces can request to be fullscreened on a specific output, which is currently ignored (#1813).
- [x] Moving a fullscreen view to another workspace unfullscreens it.
- [x] Unmanaged xwayland surfaces need to be rendered if the fullscreen view is xwayland (eg. Chromium context menu).

somdoron:
Also, when using xwayland app (chromium) in full screen context menu doesn't show up.

emersion:
Yeah, that's because unmanaged xwayland surfaces need to be rendered if the fullscreen view is xwayland.

somdoron:
@RyanDwyer I think it is worth adding that to the list.

atomnuker:
I think blanking background needs to be moved to wlroots, since with direct scanout you can't avoid doing it.

SirCmpwn:
I disagree. We're going to give the user the tools to deal with direct scan-out themselves.

emersion:
It's already in wlroots if you use `wlr_output_set_full screen_surface`. And yes, it should be there.

SirCmpwn:
For the record, I'm not happy with wlr_output_set_fullscreen_surface

emersion:
It seems you're not happy with direct scan-out too.

SirCmpwn:
ding ding ding

But fwiw I'm on board with making wlroots support it, but like everything else we need to do it in an unopinionated way.

emersion:
Agreed.

petabyteboy:
The wallpaper should still be rendered in fullscreen for transparent applications.

emersion:
>The wallpaper should still be rendered in fullscreen for transparent applications.

No. Won't work with direct scan-out.

atomnuker:
A setting was agreed on to toggle that.

atomnuker:
fullscreen flipping is still not synchronized (e.g. it can tear), right? This ought to be added.
Also fullscreen mode doesn't yet stop the frame callback for all invisible surfaces below.

emersion:
>fullscreen flipping is still not synchronized (e.g. it can tear), right?

No, it is synchronized, it can't tear.

>Also fullscreen mode doesn't yet stop the frame callback for all invisible surfaces below.

It should and used to, probably a bug.

RyanDwyer:
[...]

PR #1846, opened 2018-04-22
Render unmanaged xwayland views when an xwayland view is fullscreen

RyanDwyer:
This makes Chromium context menus appear when fullscreen.

Fixes an item on #1828.

Question: This will render all unmanaged xwayland views, even if it doesn't belong to the fullscreen view. Do I need to worry about this, and if so how do I filter them?

To test:

* Launch Chromium, fullscreen it, then right click on the page or open the context menu in the top right.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -282,6 +282,11 @@ static void render_output(struct sway_output *output, struct timespec *when,
 		wlr_renderer_clear(renderer, clear_color);
 		// TODO: handle views smaller than the output
 		render_container(output, workspace->sway_workspace->fullscreen->swayc);
+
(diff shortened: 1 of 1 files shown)


ascent12:
>Do I need to worry about this, and if so how do I filter them?

I think it's fine to just leave it. I don't even think X11 gives enough information to make this possible.

emersion:
>Do I need to worry about this, and if so how do I filter them?

Some clients set the parent of their popups correctly (like Firefox), and some others don't (like Chromium). We basically have no way to know if a client will do the right thing™, so we have no choice but to render all unmanaged xwayland surfaces.

PR #1850, opened 2018-04-23
[WIP] remove output pos <X,Y> syntax

### LINKED ISSUE #1822: Remove or replace output pos <X,Y>
The new syntax for output <X,Y> is not working.
e.g., "swaymsg output eDP-1 pos 0,1200 res 1920x1200" GIves the following error:
Error: Missing position argument (y).

While it works in the legacy mode
swaymsg output eDP-1 pos 0 1200 res 1920x1200

nearffxx:
Deprecate "output <X,Y>" in faviour or the legacy  "output \<X\> \<Y\>".

Fixes #1822

### DIFF
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -73,29 +73,15 @@ static struct cmd_results *cmd_output_position(struct output_config *output,
 
 	char *end;
 	output->x = strtol(argv[*i], &end, 10);
+	// Format is 1234 4321
+	if (++*i >= argc) {
+		return cmd_results_new(CMD_INVALID, "output",
+			"Missing position argument (y).");
+	}
+	output->y = strtol(argv[*i], &end, 10);
 	if (*end) {
-		// Format is 1234,4321
-		if (*end != ',') {
-			return cmd_results_new(CMD_INVALID, "output",
-				"Invalid position x.");
-		}
-		++end;
-		output->y = strtol(end, &end, 10);
-		if (*end) {
-			return cmd_results_new(CMD_INVALID, "output",
-				"Invalid position y.");
-		}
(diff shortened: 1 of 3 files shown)


sway/commands/output.c
```diff
@@ -73,29 +73,15 @@ static struct cmd_results *cmd_output_position(struct output_config *output,
 
 	char *end;
 	output->x = strtol(argv[*i], &end, 10);
+	// Format is 1234 4321
```
emersion: Should error out if `*end`

RyanDwyer:
This will break people's configs who use commas. From what I understand, the issue is that when running it via swaymsg (IPC), some higher level code splits the swaymsg arguments on the comma and runs them as individual commands. This causes an incomplete command to be passed.

I think the best approach is to change the IPC argument parsing code so that it that avoids splitting on a comma if the comma is followed by numbers (more specifically, zero or more spaces then at least one number).

An alternative approach is to allow the comma syntax in the output command, but detect if the y value is missing and if the message was given via IPC, and if so show an error instructing the user to use spaces instead. This is kind of dodgy, because it'll still try to split on the comma and run the post-comma part as its own command which is invalid.

BTW, I have no authority here. I just wanted to throw my ideas out there.

thejan2009:
Change this too: https://github.com/SirCmpwn/sway/blob/master/swaymsg/main.c#L134

RyanDwyer:
I have discovered you can use quotes to prevent splitting on the comma. However, Bash will swallow the quotes and not pass them to swaymsg so you have to escape them.

    $ swaymsg output FOO-1 pos \'100,100\'

PR #1852, opened 2018-04-24
Implement criteria commands

RyanDwyer:
Implements the following commands:

* `for_window [...] <cmdlist>`
* `assign [...] <workspace>`

Please note this PR aims to implement the commands themselves and doesn't aim to implement any more criteria tokens than what's already implemented. The tokens that work are:

* `app_id` only for xdg_shell_v6
* `class` for wl_shell and xwayland surfaces only
* `con_id` for all surfaces
* `title` for all surfaces

You can also add criteria at runtime via swaymsg.

Testing examples:

* `for_window [title="^.*$"] opacity 0.9`
* `assign [class="^mpv$"] media`

This is branched from PR #1851, so merge that one first.

For those reviewing, assign.c and for_window.c are almost direct copies from sway 0.15 so don't pay too much attention to those.


sway/commands/assign.c
```diff
@@ -0,0 +1,57 @@
...
+
+struct cmd_results *cmd_assign(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "assign", EXPECTED_AT_LEAST, 2))) {
+		return error;
+	}
+
+	char *criteria = *argv++;
+
+	if (strncmp(*argv, "→", strlen("→")) == 0) {
```
emersion: Is this reallt some UTF-8?

sway/commands/assign.c
```diff
@@ -0,0 +1,57 @@
...
+		if (argc < 3) {
+			return cmd_results_new(CMD_INVALID, "assign", "Missing workspace");
+		}
+		argv++;
+	}
+
+	char *movecmd = "move container to workspace ";
+	int arglen = strlen(movecmd) + strlen(*argv) + 1;
```
emersion: `size_t`

sway/commands/for_window.c
```diff
@@ -0,0 +1,41 @@
...
+
+struct cmd_results *cmd_for_window(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "for_window", EXPECTED_AT_LEAST, 2))) {
+		return error;
+	}
+	// add command to a criteria/command pair that is run against views when they appear.
+	char *criteria = argv[0], *cmdlist = join_args(argv + 1, argc - 1);
+
+	struct criteria *crit = malloc(sizeof(struct criteria));
```
emersion: Can you use `calloc` here instead?

sway/commands/assign.c
RyanDwyer: This file was copied from Sway 0.15. It's an optional part of the command: https://i3wm.org/docs/userguide.html#assign_workspace

emersion: Yeah, I just didn't know i3 had optional UTF-8 in its config files.

FYI, we're going to review Sway 0.15 files even if they are copied verbatim.

sway/tree/view.c
```diff
@@ -208,6 +211,31 @@ static void view_handle_container_reparent(struct wl_listener *listener,
 	}
 }
 
+static void view_execute_criteria(struct sway_view *view) {
+	if (!sway_assert(view->swayc, "cannot run criteria for unmanaged view")) {
```
emersion: s/unmanaged/unmapped/

sway/tree/view.c
```diff
@@ -208,6 +211,28 @@ static void view_handle_container_reparent(struct wl_listener *listener,
 	}
 }
 
+static void view_execute_criteria(struct sway_view *view) {
```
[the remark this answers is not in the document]
RyanDwyer: Resolved in 30a7488.

sway/commands/assign.c
RyanDwyer: Do you want me to change anything here?

emersion: No :^)

SirCmpwn:
Something to note is that I wanted to change criteria a little bit for wlroots because we have some access we didn't have in wlc. Each shell has different properties - xdg-shell has app-id and title, xwayland has the kitchen sink, etc. On sway 0.x we had to tell users to find new criteria to use because their old X-based criteria were sometimes not possible on sway due to limitations in wlc. However, with wlroots, we have direct access to Xwayland so we should be able to fully support all i3 criteria for Xwayland windows, as well as new criteria for xdg-shell et al.

emersion:
Thanks!

PR #1854, opened 2018-04-24
Refactor arrange_windows()

RyanDwyer:
The arrange_windows() function hurts my brain. Here's a refactor for your consideration.

* I've split `arrange_windows()` into a function for each container type. So you can call `arrange_root()`, `arrange_output(output)`, `arrange_workspace(workspace)` or `arrange_children_of(workspace_or_container)`. Each one iterates its children and recurses as needed.
* For convenience, if you don't know what type of container you're dealing with and just want it to work, you can pass it to `arrange_windows(whatever)` and it'll figure it out.
* Because there's now several more functions related to arranging, I've moved this out of layout.c and into a new file.
* Passing -1, -1 to the function is now unnecessary, as it gets it from the provided parent and updates the child's dimensions before iterating into it.
* Previously, `arrange_windows()` had inconsistent behaviour as to whether it would update properties on the passed container (eg. root and output would only work with their children). Now they always update the given container's properties, except for `arrange_children_of()` which is obvious from the name.
* The `apply_{horiz,vert}_layout()` functions no longer need `start` and `end` variables passed to them, as it always uses the full children list.
* Additionally, they no longer pass the parent's dimensions separately.
* The `apply_{horiz,vert}_layout()` functions no longer recurse. They only set the child's properties, then execution is returned to `arrange_children_of()` which handles the recursion. This means the function call stack is `arrange_children_of()` => `arrange_children_of()` rather than `arrange_children_of()` => `apply_horiz_layout()` => `arrange_children_of()` which I believe is easier to follow.
* Fullscreen is implemented differently this way. It saves the view's dimensions and restores them. Previously it was calling `view_configure()` without changing the swayc's stored width and height.

Also note I had to split view_set_fullscreen() into a lower level function called view_set_fullscreen_raw(), because I had to call the lower level one from elsewhere without having it go and rearrange the windows on me.

### DIFF
--- a/include/sway/tree/arrange.h
+++ b/include/sway/tree/arrange.h
@@ -1,0 +1,20 @@
+#ifndef _SWAY_ARRANGE_H
+#define _SWAY_ARRANGE_H
+
+struct sway_container;
+
+void arrange_windows(struct sway_container *container);
+
+// Determine the root container's geometry, then iterate to everything below
+void arrange_root(void);
+
+// Determine the output's geometry, then iterate to everything below
+void arrange_output(struct sway_container *output);
+
+// Determine the workspace's geometry, then iterate to everything below
+void arrange_workspace(struct sway_container *workspace);
+
+// Arrange layout for all the children of the given workspace/container
+void arrange_children_of(struct sway_container *parent);
+
+#endif
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -1,0 +1,239 @@
+#define _POSIX_C_SOURCE 200809L
+#include <ctype.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wlr/types/wlr_output.h>
+#include <wlr/types/wlr_output_layout.h>
+#include "sway/debug.h"
+#include "sway/tree/arrange.h"
+#include "sway/tree/container.h"
+#include "sway/tree/layout.h"
+#include "sway/output.h"
+#include "sway/tree/workspace.h"
+#include "sway/tree/view.h"
+#include "list.h"
+#include "log.h"
+
+struct sway_container root_container;
+
+void arrange_windows(struct sway_container *container) {
+	switch (container->type) {
+	case C_ROOT:
+		arrange_root();
+		break;
+	case C_OUTPUT:
+		arrange_output(container);
+		break;
+	case C_WORKSPACE:
+		arrange_workspace(container);
+		break;
+	case C_CONTAINER:
+		arrange_children_of(container);
+		break;
+	case C_VIEW:
+		// ignore
+		break;
+	case C_TYPES:
+		sway_assert(
+				false, "Called arrange_windows() with container type C_TYPES");
+		break;
+	}
+}
+
+void arrange_root() {
+	if (config->reloading) {
+		return;
+	}
+	struct wlr_output_layout *output_layout =
+		root_container.sway_root->output_layout;
+	const struct wlr_box *layout_box =
+		wlr_output_layout_get_box(output_layout, NULL);
+	root_container.x = layout_box->x;
+	root_container.y = layout_box->y;
+	root_container.width = layout_box->width;
+	root_container.height = layout_box->height;
+	for (int i = 0; i < root_container.children->length; ++i) {
+		struct sway_container *output = root_container.children->items[i];
+		arrange_output(output);
+	}
+}
+
+void arrange_output(struct sway_container *output) {
+	if (config->reloading) {
+		return;
+	}
+	if (!sway_assert(output->type == C_OUTPUT,
+			"called arrange_output() on non-output container")) {
+		return;
+	}
+	const struct wlr_box *output_box = wlr_output_layout_get_box(
+			root_container.sway_root->output_layout,
+			output->sway_output->wlr_output);
+	output->x = output_box->x;
+	output->y = output_box->y;
+	output->width = output_box->width;
+	output->height = output_box->height;
+	wlr_log(L_DEBUG, "Arranging output '%s' at %f,%f",
+			output->name, output->x, output->y);
+	for (int i = 0; i < output->children->length; ++i) {
+		struct sway_container *workspace = output->children->items[i];
+		arrange_workspace(workspace);
+	}
+	container_damage_whole(output);
+}
+
+void arrange_workspace(struct sway_container *workspace) {
+	if (config->reloading) {
+		return;
+	}
+	if (!sway_assert(workspace->type == C_WORKSPACE,
+			"called arrange_workspace() on non-workspace container")) {
+		return;
+	}
+	struct sway_container *output = workspace->parent;
+	struct wlr_box *area = &output->sway_output->usable_area;
+	wlr_log(L_DEBUG, "Usable area for ws: %dx%d@%d,%d",
+			area->width, area->height, area->x, area->y);
+	workspace->width = area->width;
+	workspace->height = area->height;
+	workspace->x = area->x;
+	workspace->y = area->y;
+	wlr_log(L_DEBUG, "Arranging workspace '%s' at %f, %f",
+			workspace->name, workspace->x, workspace->y);
+	arrange_children_of(workspace);
+	container_damage_whole(workspace);
+}
+
+static void apply_horiz_layout(struct sway_container *parent) {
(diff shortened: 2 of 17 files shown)


SirCmpwn:
I like the ideas in your PR description

sway/tree/arrange.c
```diff
@@ -0,0 +1,240 @@
...
+
+struct sway_container root_container;
+
+static void apply_horiz_layout(struct sway_container *parent);
+
+static void apply_vert_layout(struct sway_container *parent);
+
+void arrange_windows(struct sway_container *container) {
+	switch (container->type) {
+		case C_ROOT:
```
ascent12: Align cases with the switch.

include/sway/tree/arrange.h
```diff
@@ -0,0 +1,20 @@
+#ifndef _SWAY_ARRANGE_H
+#define _SWAY_ARRANGE_H
+
+struct sway_container;
+
+void arrange_windows(struct sway_container *container);
+
+// Determine the root container's geometry, then iterate to everything below
+void arrange_root();
```
ascent12: Explicitly put `(void)` for the arguments.
Empty argument lists in prototypes have surprising results in C.

sway/tree/arrange.c
```diff
@@ -0,0 +1,240 @@
...
+		} else {
+			arrange_children_of(child);
+		}
+	}
+	container_damage_whole(parent);
+	update_debug_tree();
+}
+
+static void apply_horiz_layout(struct sway_container *parent) {
+	wlr_log(L_ERROR, "In apply_horiz_layout()");
```
ascent12: Superfluous log message.

sway/tree/arrange.c
```diff
@@ -0,0 +1,240 @@
...
+	size_t num_children = parent->children->length;
+	if (!num_children) {
+		return;
+	}
+	// Calculate total width of children
+	double scale = 0;
+	for (size_t i = 0; i < num_children; ++i) {
+		struct sway_container *child = parent->children->items[i];
+		wlr_log(L_ERROR, "child (%s) width was %f", child->name, child->width);
```
ascent12: This too.

sway/tree/arrange.c
RyanDwyer: Whoops, I was meant to remove that.

sway/tree/arrange.c
```diff
@@ -0,0 +1,238 @@
...
+struct sway_container root_container;
+
+static void apply_horiz_layout(struct sway_container *parent);
+
+static void apply_vert_layout(struct sway_container *parent);
+
+void arrange_windows(struct sway_container *container) {
+	switch (container->type) {
+	case C_ROOT:
+		   arrange_root();
```
ascent12: You've still got messed up indentation here. There are an extra 3 spaces at the end.

sway/tree/arrange.c
```diff
@@ -0,0 +1,238 @@
...
+#include "sway/tree/workspace.h"
+#include "sway/tree/view.h"
+#include "list.h"
+#include "log.h"
+
+struct sway_container root_container;
+
+static void apply_horiz_layout(struct sway_container *parent);
+
+static void apply_vert_layout(struct sway_container *parent);
```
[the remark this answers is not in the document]
RyanDwyer: They can be removed if I move their implementations to above where they're called. The original layout.c did the same thing (see current master).

I guess it's a tradeoff between duplicating the definition and having the flow of function calls be linear, or not duplicating the definition and having to jump around the file when reading the code. Not sure which is best.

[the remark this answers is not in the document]
ascent12: I personally don't like using prototypes when they're not necessary, and would reorder the functions to not need them.
You'd normally read a file from bottom to top anyway.

I don't know how other people feel about this.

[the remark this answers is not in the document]
emersion: In this case reordering doesn't hurt readability. It's true that files are generally read from bottom to top. I don't feel strongly about it though.

In other situations where functions operating on different types depend on each other and are in the same file, forward definitions can definitely help improve readability.

include/sway/tree/arrange.h
```diff
@@ -0,0 +1,20 @@
...
+
+struct sway_container;
+
+void arrange_windows(struct sway_container *container);
+
+// Determine the root container's geometry, then iterate to everything below
+void arrange_root(void);
+
+// Determine the output's geometry, then iterate to everything below
+void arrange_output(struct sway_container *output);
```
emersion: I'm not a fan of these "I take a generic `struct sway_container` as input but I only work on output/workspace". Basically these functions could be typed and take a `sway_output`/`sway_workspace`.

sway/tree/arrange.c
```diff
@@ -0,0 +1,234 @@
...
+	for (size_t i = 0; i < num_children; ++i) {
+		struct sway_container *child = parent->children->items[i];
+		if (child->height <= 0) {
+			if (num_children > 1) {
+				child->height = parent->height / (num_children - 1);
+			} else {
+				child->height = parent->height;
+			}
+		}
+		scale += child->height;
```
emersion: It's confusing to use the `scale` variable to accumulate child heights. Maybe we could use another variable?

sway/tree/arrange.c
```diff
@@ -0,0 +1,234 @@
...
+	}
+	scale = parent->height / scale;
+
+	// Resize
+	wlr_log(L_DEBUG, "Arranging %p vertically", parent);
+	double child_y = parent->y;
+	struct sway_container *child;
```
emersion: This can go inside the loop

sway/tree/view.c
```diff
@@ -114,11 +117,22 @@ void view_set_fullscreen(struct sway_view *view, bool fullscreen) {
...
+	if (view->is_fullscreen == fullscreen) {
+		return;
+	}
 
+	view_set_fullscreen_raw(view, fullscreen);
+
+	struct sway_container *workspace =
+		container_parent(view->swayc, C_WORKSPACE);
+	arrange_workspace(workspace);
+	output_damage_whole(workspace->parent->sway_output);
```
emersion: That should probably go into `view_set_fullscreen_raw`?

include/sway/tree/arrange.h
RyanDwyer: `arrange_children_of()` would still need to accept a `sway_container`, and that means there'd be inconsistency in how you call these functions.

Additionally, there'd be nothing stopping you from accidentally passing the wrong pointer for that container type to the function. For example, if you had a C_CONTAINER, assumed it was a workspace by mistake and passed its sway_workspace pointer to `arrange_workspace()`. That wouldn't be caught during compilation, and `arrange_workspace()` wouldn't be able to assert that this has not happened.

emersion: I don't see the value of having 4 functions taking "generic containers, but not really" instead of one.

When you access a sway_workspace field you must _always_ have an assert just above checking the type (or you already have the specialized type).

sway/tree/view.c
emersion: ACK

include/sway/tree/arrange.h
SirCmpwn: ¯\\\_(ツ)\_/¯

Sort it out between yourselves

emersion: @RyanDwyer I just don't get why there's a need for these per-type functions instead of a generic one (which is here anyway). Can you explain me why you made this change?

emersion:
I guess w/e, this isn't a big deal anyway.

include/sway/tree/arrange.h
RyanDwyer: The previous function had a lot of complexity which made it hard to understand. It was recursive, used different logic per container type, and some of the switch statements used return while others use break. Separating them into different functions improves readability because the part that is recursive is isolated to a smaller area of code.

In hindsight, I wouldn't have included the new `arrange_windows()` function at all. It's not called anywhere, and it's not a very appropriate name because it can also arrange outputs and workspaces.

emersion: >Separating them into different functions improves readability because the part that is recursive is isolated to a smaller area of code.

Sure, but these functions could be static.

>It's not called anywhere, and it's not a very appropriate name because it can also arrange outputs and workspaces.

Can you make a PR to remove it then?

PR #1860, opened 2018-04-24
Add map_from_region command

emersion:
This allows to map from a region of tablet tool.


emersion:
Updated man pages, added mm support, renamed to `map_from_region` for consistency with `map_to_{output,region}`.

PR #1861, opened 2018-04-24
Fix a bunch of swaybar memory leaks

emersion:
There are more leaks from fontconfig and cairo. I guess we can't do anything about it.


atomnuker:
4 hours in, resident memory hasn't changed a kb. I think all per-frame leakage has disappeared.

SirCmpwn:
Woot

PR #1866, opened 2018-04-27
invoke_swaybar: Only invoke swaybar via sh if user specifies a custom swaybar command

ggreer:
This reproduces 0.15's behavior. It indirectly fixes some issues where you can get multiple swaybars when reloading the config or having multiple displays. Using sh on Ubuntu (which is [dash](https://wiki.ubuntu.com/DashAsBinSh), not sh) caused terminate_swaybar() to kill sh, but sh's child swaybar was then owned by pid 1.


martinetd:
If the problem is that there are two processes instead of one (sh and swaybar), just run sh -c "exec <original command>" instead?

Otherwise you're leaving the problem open for everyone with a custom swaybar command...

ascent12:
From https://i3wm.org/docs/userguide.html#_i3bar_command, it specifically mentions going through `sh -c`.

I think all you need to do is call `setsid()` after the fork, and in `terminate_swaybar` negate the pid in `kill`, so it targets the whole process group.
This is what i3 does.

ascent12:
Actually, after looking it up, it may be more appropriate to call `setpgid` instead of `setsid`.

PR #1868, opened 2018-04-27
Run in verbose mode

bpinto:
### Proposal

This change causes display managers to launch `sway` using the `--verbose` option.

With this change, systemd-enabled systems can access sway's log using `journalctl` tool.

### DIFF
--- a/sway.desktop
+++ b/sway.desktop
@@ -1,5 +1,5 @@
 [Desktop Entry]
 Name=Sway
 Comment=SirCmpwn's Wayland window manager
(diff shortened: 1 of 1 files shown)


emersion:
I don't think this is necessary. Errors will show in the log without `--verbose`, so errors will show up in the journal. If there's an error, users can run sway manually to get a debug log.

bpinto:
You are right, I had missed the errors on the `journalctl` when running without `--verbose` mode. But I have tried it again and I can see them now.

PR #1871, opened 2018-04-29
Kill wl_shell

### LINKED ISSUE #1692: Drop wl_shell support
No one uses it, it's deprecated, and we have no way to test it.

It's just an issue to track the progress of this.

emersion:
https://github.com/hdante/hello_wayland/pull/6 has been merged.

Fixes #1692


SirCmpwn:
Blocked pending a new mpv release

atomnuker:
Why? It's not a lot of code and there are still users, and its in the baseline protocol.

SirCmpwn:
It's broken and deprecated and annoying to maintain. We can't find any examples of users who don't support xdg-shell too, but if you come up with examples we'll block this on sending them patches.

atomnuker:
The thing is its part of the base protocol. Its usually the first thing implemented by someone who just wants to show something on the screen and doesn't know any better. And usually people only go for xdg if they need more features and because they haven't heard of wl_shell being deprecated (I haven't heard of that either, and indeed the headers mention nothing).
I'd say its still too early to tell whether its really been phased completely out. It was the only stable protocol up to until barely December.

SirCmpwn:
Work is ongoing to add support for deprecation to wayland-scanner, deprecating wl_shell, and removing it. There is no reason to use wl_shell. Again, find links to software that doesn't support xdg-shell and we can send them patches.

atomnuker:
wayland-scanner printing warnings matters for nought if distributions never ship wayland.xml but only preprocessed wayland headers. Furthermore you can't actually remove it from the protocol because it'll likely break a ton of programs.

SirCmpwn:
The plan is to make use of e.g. gcc deprecation attributes. And you still haven't even shown us one of these tons of programs that will break.

SirCmpwn:
Also, you'll have to ship xml anyway because wayland-scanner isn't the only scanner in town.

emersion:
>I'd say its still too early to tell whether its really been phased completely out. It was the only stable protocol up to until barely December.

It's [explicitly mentioned](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_shell) in the protocol that nobody should use it anymore.

NilsBrause:
Before dropping `wl_shell`, its stable successor `xdg_wm_base` should be supported first. As far as I can see, this is not the case yet. There is only support for the unstable `zxdg_shell_v6`:
```
% ./dump
* Global interface wl_drm (name 1 version 2)
* Global interface wl_shm (name 2 version 1)
* Global interface wl_compositor (name 3 version 4)
* Global interface wl_subcompositor (name 4 version 1)
* Global interface wl_data_device_manager (name 5 version 3)
* Global interface orbital_screenshooter (name 6 version 1)
* Global interface gamma_control_manager (name 7 version 1)
* Global interface gtk_primary_selection_device_manager (name 8 version 1)
* Global interface zxdg_output_manager_v1 (name 9 version 1)
* Global interface zwlr_layer_shell_v1 (name 10 version 1)
* Global interface zxdg_shell_v6 (name 11 version 1)
* Global interface wl_shell (name 12 version 1)
* Global interface org_kde_kwin_server_decoration_manager (name 13 version 1)
* Global interface zwp_linux_dmabuf_v1 (name 14 version 3)
* Global interface wl_seat (name 15 version 6)
* Global interface zwlr_input_inhibit_manager_v1 (name 16 version 1)
* Global interface wl_output (name 17 version 3)
```

emersion:
>Before dropping wl_shell, its stable successor xdg_wm_base should be supported first.

This will be done in another PR, but I'd like to get the annoying https://github.com/swaywm/wlroots/issues/916 bug fixed first.

acrisci:
Having experience doing the majority of the wl_shell implementation for wlroots, I can say without a doubt it's impossible to make a working desktop client with this protocol and I support its removal everywhere.

atomnuker:
Its alright, its just very barebones, gives you nothing useful and doesn't integrate well with any shell.

emersion:
mpv 0.28 with Wayland support is available and xdg-shell stable support has been merged.

emersion:
Rebased

SirCmpwn:
RIP o7

PR #1874, opened 2018-04-30
Implement borders

RyanDwyer:
Implements rendering of borders. Title text is still to do.

Implements the following configuration directives:

* `client.focused`
* `client.focused_inactive`
* `client.unfocused`
* `client.urgent` (parsed and stored but not used)
* `border`
* `default_border`

Interesting things to note:

* It was mentioned in IRC that we should use a `wlr_texture`, but I've done it textureless using `wlr_render_rect()`.
* Colours are now stored as a `float[4]` rather than `uint32_t` so that they can be passed directly to `wlr_render_rect()`.
* I've made `default_border` a runtime command (in addition to config-time). This means you can change the default border via IPC, but this is inconsistent with `default_layout` which is config-time only. We should make these consistent, one way or the other.
* I've added x and y properties to sway_view. Now the sway_container's geometrical properties refer to the geometry including borders, and the sway_view's geometrical properties refer to the geometry excluding borders. If this sounds stupid and you have a better idea, let me know.
* I've added the `view_autoconfigure()` function. This calls the existing `view_configure()` after calculating its geometry from its swayc and border configuration.
* `view_update_position()` is now removed. I note this function previously handled damage, but I don't think this is an issue any more because it's called from the arrange functions which always damages the container.
* If you set opacity on a view, the border will also use the opacity.


SirCmpwn:
>It was mentioned in IRC that we should use a wlr_texture, but I've done it textureless using wlr_render_rect().

-1

This approach will not be sustainable as we add text to it

emersion:
>view_update_position() is now removed

-1, Xwayland views can move themselves.

SirCmpwn:
>Colours are now stored as a float[4] rather than uint32_t so that they can be passed directly to wlr_render_rect().

This was done for easier use with cairo

RyanDwyer:
It's been decided in IRC that we can stick with wlr_render_rect() and floats for now. Please review, test and submit feedback.

emersion:
>Title text is still to do.

Do we want this in this PR as well? Or is this PR not in WIP anymore?

SirCmpwn:
>Or is this PR not a WIP anymore?

You were the one who changed the title to WIP

emersion:
>You were the one who changed the title to WIP

Yeah, assumed this was WIP since there was no title text.

include/sway/tree/view.h
```diff
@@ -160,6 +167,8 @@ const char *view_get_instance(struct sway_view *view);
 void view_configure(struct sway_view *view, double ox, double oy, int width,
 	int height);
 
+void view_autoconfigure(struct sway_view *view);
```
SirCmpwn: Add comment explaining what this does

sway/desktop/output.c
```diff
@@ -217,23 +217,198 @@ static void render_unmanaged(struct sway_output *output,
...
+	box.x = con->x + con->sway_view->border_thickness;
+	box.y = con->y + 1;
+	box.width = con->width - con->sway_view->border_thickness * 2;
+	box.height = con->sway_view->y - con->y - 2;
+	wlr_render_rect(renderer, &box, color,
+			output->wlr_output->transform_matrix);
+
+	// Title text
+	// TODO
+}
```
SirCmpwn: This seems to be missing indicator color afaict

sway/desktop/output.c
```diff
@@ -217,23 +217,198 @@ static void render_unmanaged(struct sway_output *output,
...
+			output->wlr_output->transform_matrix);
+
+	// Child border - bottom edge
+	box.x = con->x;
+	box.y = con->y + con->height - con->sway_view->border_thickness;
+	box.width = con->width;
+	box.height = con->sway_view->border_thickness;
+	wlr_render_rect(renderer, &box, color,
+			output->wlr_output->transform_matrix);
+}
```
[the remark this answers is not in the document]
RyanDwyer: I was thinking perhaps I should move all the rendering into a new render.c file. Are you cool with that?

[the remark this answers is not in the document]
emersion: You could also keep the rendering in `output.c` and move the iterator stuff in another file.

Whatever you think is the best. Personally I'm cool with both.

sway/commands/client.c
```diff
@@ -0,0 +1,79 @@
+#include "log.h"
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/tree/container.h"
+
+static bool parse_color(char *hexstring, float (*dest)[4]) {
```
emersion: No need for a pointer here. Also, add `static` so that the min number of elements is 4:

```c
float dest[static 4]
```

sway/config.c
```diff
@@ -130,6 +130,13 @@ static void destroy_removed_seats(struct sway_config *old_config,
 	}
 }
 
+static void set_color(float (*dest)[4], uint32_t color) {
```
emersion: Same applies here

sway/config.c
```diff
@@ -130,6 +130,13 @@ static void destroy_removed_seats(struct sway_config *old_config,
...
+	(*dest)[0] = ((color >> 16) & 0xff) / 255.0;
+	(*dest)[1] = ((color >> 8) & 0xff) / 255.0;
+	(*dest)[2] = (color & 0xff) / 255.0;
+	(*dest)[3] = 1.0;
```
emersion: Maybe we want to parse alpha too?

emersion: Nevermind, not used here

sway/commands/client.c
```diff
@@ -0,0 +1,79 @@
...
+	uint32_t decimal = strtol(hexstring, &end, 16);
+
+	if (*end != '\0') {
+		return false;
+	}
+
+	(*dest)[0] = ((decimal >> 16) & 0xff) / 255.0;
+	(*dest)[1] = ((decimal >> 8) & 0xff) / 255.0;
+	(*dest)[2] = (decimal & 0xff) / 255.0;
+	(*dest)[3] = 1.0;
```
emersion: Parse alpha?

RyanDwyer: At the moment I take the alpha from the container (from the `opacity` command). I guess we can allow them to enter the alpha part in their config and combine the two. I'd rather this be a separate PR so this one doesn't become huge.

emersion: ACK. Please file a follow-up ticket when this PR gets merged.

SirCmpwn: imo we should disregard the client opacity when drawing the borders

emersion: You mean, the per-view opacity?

Yeah, w/e, I don't know what the use-case for opacity is in the first place.

SirCmpwn: Yeah, we should just use the border's configured alpha color.

SirCmpwn:
woot

PR #1889, opened 2018-05-02
Fixed #1888 GCC cannot compile due to uninitialized variables

### LINKED ISSUE #1888: gcc 7.3.1: sway/sway/tree/view.c:115:2: error: ‘height’ may be used uninitialized in this function
GCC version: 7.3.1
OS: Archlinux

It seems that in `sway/sway/tree/view.c:90` there are four variables (x, y, width and height) initialized without a value. GCC throws an error about this and because of this I can no longer compile sway from master on Archlinux.

I'm no expert in C, at all but I can create a pull-request if the fix is something like initializing them with a NULL value or something.

Can someone put me in the right direction?

dmeijboom:
This is a fix for: #1888 
Sway wouldn't compile on Archlinux due to uninitialized variables.

For the record, these build flags are passed automatically to GCC on Archlinux in makepkg:
```
CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt"
LDFLAGS="-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now"
CPPFLAGS="-D_FORTIFY_SOURCE=2"
```

Note that I don't have a lot of experience in C so please let me know if this is not the way it should be fixed.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -88,6 +88,9 @@ void view_autoconfigure(struct sway_view *view) {
 	}
 
 	double x, y, width, height;
+
+	x = y = width = height = 0;
+
 	switch (view->border) {
 	case B_NONE:
 		x = view->swayc->x;
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -88,9 +88,7 @@ void view_autoconfigure(struct sway_view *view) {
 	}
 
 	double x, y, width, height;
-
 	x = y = width = height = 0;
-
 	switch (view->border) {
 	case B_NONE:
 		x = view->swayc->x;


sway/tree/view.c
```diff
@@ -88,6 +88,9 @@ void view_autoconfigure(struct sway_view *view) {
 	}
 
 	double x, y, width, height;
+
```
emersion: Is it possible to remove the newlines here and after the assignment?

dmeijboom: Done!

RyanDwyer:
Out of curiosity, what commands are you running to compile it? I'm on Arch Linux with GCC 7.3.1 too and it compiles fine for me.

dmeijboom:
@RyanDwyer Nothing more than `makepkg -si` on the cloned git repository from the AUR. I checked my `makepkg.conf` but it looks like I don't have any custom CFLAGS or something so I think that those are the defaults on Archlinux but I'm not sure..

PR #1890, opened 2018-05-02
Remove C_TYPES

emersion:


RyanDwyer:
We had a short discussion on IRC and came up with three possible ways we can do this:

1. Use pointers as the PR currently does.
2. Make the container type values use 1, 2, 3, 4... and pass 0 to the `seat_get_focus_by_type()` function if you want all types.
3. Make the container type values use 1, 2, 4, 8... and pass `C_OUTPUT | C_WORKSPACE` etc to the `seat_get_focus_by_type()` function if you want all types.

A benefit of option 3 is we can use that elsewhere, like `sway_assert(container->type & (C_CONTAINER | C_VIEW))`.

SirCmpwn:
I like the bitfield approach

emersion:
Updated to use a bitfield. That was a good idea @RyanDwyer

Issue #1892, opened 2018-05-02
Continue to send pointer/touch motion outside of the surface


SirCmpwn:
If you click and drag the pointer out of the surface, we should continue to send motion to the surface you clicked on.

emersion:
Duplicate of #1862

SirCmpwn:
rip

PR #1893, opened 2018-05-02
Implement basic touch support

SirCmpwn:
This required changing container_at_cursor to container_at_coords so
that we could get the appropriate surface (and sx/xy) without moving the
cursor.

Future work:

- Simulate a pointer for clients which have not bound to wl_touch
  - With long-press to right click
- Keep sending motion events when moving outside the surface (#1892)
- Bind gestures to sway commands

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -44,7 +44,8 @@ static struct wlr_surface *layer_surface_at(struct sway_output *output,
  * Returns the container at the cursor's position. If there is a surface at that
  * location, it is stored in **surface (it may not be a view).
  */
-static struct sway_container *container_at_cursor(struct sway_cursor *cursor,
+static struct sway_container *container_at_coords(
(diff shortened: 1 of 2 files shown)


sway/input/cursor.c
```diff
@@ -44,7 +44,8 @@ static struct wlr_surface *layer_surface_at(struct sway_output *output,
  * Returns the container at the cursor's position. If there is a surface at that
  * location, it is stored in **surface (it may not be a view).
  */
-static struct sway_container *container_at_cursor(struct sway_cursor *cursor,
+static struct sway_container *container_at_coords(
```
emersion: We already have `container_at`, adding `container_at_coords` is confusing.

The doc comment should be updated.

SirCmpwn: This function does too much, we should split it into container_at and surface_at and update a bunch of stuff, but that's out of scope for this PR.

PR #1895, opened 2018-05-02
Render titles

RyanDwyer:
Also implements the `font` command.

The text isn't antialiased (smoothed) so it looks poor at smaller font sizes. Suggestions for how to improve this are welcome.

The `text` colour is now stored as a uint32_t again to make it easier to give to cairo. The others remain as floats.

The thing works by storing four textures in the sway_container struct: one for focused, one for focused_inactive and so on. They are all updated at the same time. When rendering it picks which texture to use.

If a view updates its title it will NOT update the texture. If this is relatively easy to do then I'll add it to the PR, otherwise keep it separate.

You'll be pleased to know I've actually tested it with output scaling this time :)


include/sway/config.h
```diff
@@ -206,7 +206,7 @@ struct bar_config {
 struct border_colors {
 	float border[4];
 	float background[4];
-	float text[4];
+	uint32_t text; /**< uint32_t because that's the format that cairo uses */
```
SirCmpwn: This isn't the appropriate time for a `/**<` comment

sway/config.c
```diff
@@ -130,6 +132,17 @@ static void destroy_removed_seats(struct sway_config *old_config,
 	}
 }
 
+int get_font_text_height(char *font) {
+	cairo_t *cairo = cairo_create(NULL);
+	int text_height;
+	get_text_size(cairo, font, NULL, &text_height, 1, false,
+			"ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+			"abcdefghijklmnopqrstuvwxyz"
+			"!@#$%^&*([{|");
```
SirCmpwn: This approach isn't going to cut it, you'll have to render the text and if it's taller/shorter than you expected you'll need to re-compute the border geometry and re-render. Re-render compositor-wide if this happens, it would be weird to see each window with a differently sized border.

include/sway/config.h
RyanDwyer: Does `/**<` have any special meaning over the standard `//`? I copied the comment format from a few lines below because I thought there's no special meaning and it's just a style thing.

SirCmpwn: We informally use it for documentation under the assumption that maybe someone will export them with some kind of generator

PR #1905, opened 2018-05-03
Fix title texture width

RyanDwyer:
Use minimum of the container width or text width.

Also creates a `scale` variable in this function for convenience.

Tested with scaling.


SirCmpwn:
Thanks!

PR #1908, opened 2018-05-03
[WIP] Add tabbed layout

Dudemanguy911:
This adds tabbed windows to the layout. The `apply_tabbed_layout` function certainly could be utilized for the stacked layout like in 0.15, but that can be tackled later.


sway/commands/layout.c
```diff
@@ -25,6 +25,7 @@ struct cmd_results *cmd_layout(int argc, char **argv) {
 
 	// TODO: stacks and tabs
 
+
```
SirCmpwn: This newline doesn't need to be here and the comment above should be updated

sway/desktop/output.c
```diff
@@ -429,7 +437,53 @@ static void render_container_simple(struct sway_output *output,
  */
 static void render_container_tabbed(struct sway_output *output,
 		struct sway_container *con) {
-	// TODO
+	struct sway_seat *seat = input_manager_current_seat(input_manager);
+	struct sway_container *focus = seat_get_focus(seat);
+
+	for (int i = 0; i < con->children->length; ++i) {
```
emersion: `con->children` can be NULL

sway/desktop/output.c
```diff
@@ -429,7 +437,53 @@ static void render_container_simple(struct sway_output *output,
...
+					if (child->sway_view->border == B_NORMAL) {
+						render_container_simple_border_normal(output, child,
+								colors, title_texture, i);
+					} else {
+						render_container_simple_border_pixel(output, child, colors);
+					}
+				}
+			}
+		}
+		else {
```
emersion: Style: `}` and `else {` on the same line

sway/tree/arrange.c
```diff
@@ -159,6 +159,20 @@ static void apply_vert_layout(struct sway_container *parent) {
 	child->height = parent->y + parent->height - child->y;
 }
 
+static void apply_tabbed_layout(struct sway_container *parent) {
+	size_t num_children = parent->children->length;
+	if (!num_children) {
+		return;
+	}
+	for (size_t i=0; i < num_children; ++i) {
```
emersion: Style: spaces around `=`

SirCmpwn:
Nice work.

- A split container which is a child of a tabbed container has issues here
- Clicking a tab doesn't focus it (known limitation?)

Also, note that previously the tabbed layout forced the border to be normal, disregarding pixel borders. I don't mind if we support pixel borders, but it should render them like tabs without text - currently it's just one border whose color is picked in strange ways.

sway/desktop/output.c
```diff
@@ -228,11 +228,19 @@ static void render_view(struct sway_view *view, struct sway_output *output) {
  */
 static void render_container_simple_border_normal(struct sway_output *output,
 		struct sway_container *con, struct border_colors *colors,
-		struct wlr_texture *title_texture) {
+		struct wlr_texture *title_texture, int iteration) {
```
emersion: I'd name this `depth` instead of `iteration`, but is just a personal preference.

sway/desktop/output.c
```diff
@@ -228,11 +228,19 @@ static void render_view(struct sway_view *view, struct sway_output *output) {
...
 	struct wlr_renderer *renderer =
 		wlr_backend_get_renderer(output->wlr_output->backend);
 	struct wlr_box box;
 	float color[4];
+	size_t num_tabs;
+
+	if (con->parent->layout == L_TABBED) {
+		num_tabs = con->parent->children->length;
+	}
+	else {
```
emersion: Style: `}` and `else {` on the same line. Can probably be rewritten as:

```
size_t num_tabs = 1;
if (…) {
    …
}
```

sway/desktop/output.c
```diff
@@ -429,7 +437,53 @@ static void render_container_simple(struct sway_output *output,
...
+	struct sway_container *focus = seat_get_focus(seat);
+
+	for (int i = 0; i < con->children->length; ++i) {
+		struct sway_container *child = con->children->items[i];
+
+		if (child->type == C_VIEW) {
+			if (child->sway_view->border != B_NONE) {
+				struct border_colors *colors;
+				struct wlr_texture *title_texture;
+				if (focus == child) {
```
emersion: The code after `if (child->sway_view->border == B_NORMAL) {` can be factored out of this `if`.

We may want a function that chooses `colors` and `title_texture` for a child (since this code is duplicated with [line 408](https://github.com/swaywm/sway/pull/1908/files#diff-28cc674457237875f7921b0d29d34ac2R408)).

RyanDwyer:
Good job so far.

* The titles need the widths of their textures amended - they run into each other at the moment.
* The left/bottom/right border is shown as the inactive border, unless I focus the last child.

sway/commands/default_orientation.c
```diff
@@ -11,11 +11,13 @@ struct cmd_results *cmd_default_orientation(int argc, char **argv) {
 		config->default_orientation = L_HORIZ;
 	} else if (strcasecmp(argv[0], "vertical") == 0) {
 		config->default_orientation = L_VERT;
+	} else if (strcasecmp(argv[0], "tabbed") == 0) {
```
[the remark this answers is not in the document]
thejan2009: There is a `workspace_layout` command in i3 though. And it works in quite an unexpected way (see https://github.com/i3/i3/issues/2402).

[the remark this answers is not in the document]
RyanDwyer: So to clarify, `workspace_layout` determines the layout that new containers have when they are direct children of the workspace, while `default_orientation` changes the layout of the workspace itself. I agree that this is confusing, and I'd even vote to break from i3's behaviour and simply remove both commands if it makes sense to do so. But that's not something that should be done as part of this PR (or done at all unless discussed and agreed on), so this PR should follow i3's behaviour.

Dudemanguy911:
Went ahead and rebased this (and sorry, I somehow messed up the order of my commit history in the process) since some of @RyanDwyer's last commits were useful. I've addressed some things, but there are still a couple of lingering limitations/questions.

This may not be ideal, but previously the mouse was pretty broken on the tabbed layout. Any click would automatically focus on the very first child meaning that any children past the very first one were totally unusable with the mouse. I think this is because the `container_at_coords` function grabs the surface of the parent container and the `dispatch_mouse_button` function changes the focus to the parent container. I just told `dispatch_mouse_button` not to change the focus if the layout is tabbed, but that also means that clicking the border for other children won't change the focus.

The borders aren't totally perfect (I also ripped out pixel borders for now). I know @SirCmpwn suggested building the border from just one window and that makes sense to me as well. However, I haven't thought of a simple way to hack that onto `render_border_simple_normal` and I think that would probably require a new function. At the same time, the outline borders can be fixed as well. Currently, only the left/right/bottom borders will be shown as inactive unless the last child is focused. I think those two things can be handled in a future pull request though. I don't want to bloat up this one.

SirCmpwn:
I understand the desire to avoid bloating this PR, but at +71/-21 you have a long ways to go before you get there. I'd rather not merge with known bugs when we could just fix those bugs and merge later.

Dudemanguy911:
Okay, I'll spend some time on it.

Dudemanguy911:
Another update. Since only the top bar needs to be different for the tabbed view, I pulled out the part of `render_container_simple_border_normal` that renders to the top bar and put it in its own function. I also added `render_container_top_tabbed_border_normal' and `render_container_tabbed_border_normal` which piggyback off of the other rendering functions specifically for tabbed view. The design is pretty different now and some of this could be carried over to `render_container_simple` if it's desired.

I believe all of the concerns are addressed now. The border is now only rendered on the focused child and from there it calculates the position of the textures and borders for the other children. Clicking the mouse on the tabs will not change the view. Like I said before, this is because of how `container_at_coords` works and honestly I don't really don't have a good idea how to get around fixing that. But other than the mouse not being able to focus new tabs, everything should working nicely now.

sway/desktop/output.c
```diff
@@ -427,7 +540,29 @@ static void render_container_simple(struct sway_output *output,
  */
 static void render_container_tabbed(struct sway_output *output,
 		struct sway_container *con) {
-	// TODO
+	struct sway_seat *seat = input_manager_current_seat(input_manager);
+	struct sway_container *focus = seat_get_focus(seat);
+	struct wlr_texture **title_texture = get_title_textures(con);
+
+	for (int i = 0; i < con->children->length; ++i) {
```
RyanDwyer: Could you change the logic in this function to be:

* Iterate all children and render the tab/title
** Inside the loop, if the child is focused or focus_inactive, store it in a variable `active_child`
* Outside of the title loop, render the child stored in `active_child`

You then wouldn't need `get_title_textures` at all.

sway/desktop/output.c
```diff
@@ -325,7 +425,6 @@ static void render_container_simple_border_pixel(struct sway_output *output,
 	struct wlr_box box;
 	float color[4];
 
-	// Child border - left edge
```
RyanDwyer: Why remove this comment?

sway/desktop/output.c
```diff
@@ -224,11 +224,146 @@ static void render_view(struct sway_view *view, struct sway_output *output) {
...
+	}
+}
+
+/**
+ * Render decorations for a horizontal or vertical view with "border normal".
  */
 static void render_container_simple_border_normal(struct sway_output *output,
 		struct sway_container *con, struct border_colors *colors,
-		struct wlr_texture *title_texture) {
+		struct wlr_texture *title_texture, bool top) {
```
RyanDwyer: To be honest I'm not a fan of piggybacking off this function. "simple" refers to a linear layout, being L_HORIZ or L_VERT.

It's a tradeoff between code duplication vs the complexity of having separate execution paths through the function. I prefer the code duplication (separate function) option.

sway/desktop/output.c
```diff
@@ -376,6 +475,20 @@ static void render_container_simple_border_pixel(struct sway_output *output,
 			output->wlr_output->transform_matrix);
 }
 
+/**
+ * Render decorations for a tabbed view with "border normal".
+ */
+
+static void render_container_tabbed_border_normal(struct sway_output *output,
+		struct sway_container *con, struct wlr_texture **title_texture,
+		int depth, bool active) {
```
RyanDwyer: Why not pass a pointer to the appropriate border_colors struct like the layout does? Remember, some tabs might need to be rendered with the urgent class.

sway/desktop/output.c
```diff
@@ -224,11 +224,146 @@ static void render_view(struct sway_view *view, struct sway_output *output) {
 }
 
 /**
- * Render decorations for a view with "border normal".
+ * Select a color to use to render a border based on focus.
+ */
+static struct border_colors *select_focus_color(char *name) {
```
RyanDwyer: Could you remove this function and just refer to `&config->border_colors.foo` directly as needed?

sway/desktop/output.c
```diff
@@ -224,11 +224,146 @@ static void render_view(struct sway_view *view, struct sway_output *output) {
...
+	scale_box(&box, output->wlr_output->scale);
+	wlr_render_rect(renderer, &box, color,
+			output->wlr_output->transform_matrix);
+
+	// Title background
+	memcpy(&color, colors->background, sizeof(float) * 4);
+	color[3] *= con->alpha;
+	box.x = con->x + depth * tab_width;
+	box.y = con->y + 1;
+	box.width = con->width / num_tabs + 2;
```
RyanDwyer: I don't think this is right. The width needs to account for the borders on the sides. The title texture overlaps the right border because of this.

sway/desktop/output.c
Dudemanguy911: Do you mean to render the tab/title for every child within the loop? I thought @SirCmpwn didn't want to take that approach and wanted the tab/title to be only rendered for one child and for the logic for the other tabs to come from there. Please correct me if I'm misunderstanding either one of you.

RyanDwyer: > Do you mean to render the tab/title for every child within the loop?

Yes.

> I thought @SirCmpwn didn't want to take that approach and wanted the tab/title to be only rendered for one child and for the logic for the other tabs to come from there. Please correct me if I'm misunderstanding either one of you.

Yeah, I think there's a misunderstanding. The tab/title must be rendered for all children (which it is), but he's saying that the border width should be the same, regardless of whether the children use different widths or which child is focused. Cycling through each child should make the layout remain the same (ie. one border width used for all). I think the easiest and best option is to use the default border width stored in `config->border_thickness` rather than look for one in a child.

RyanDwyer: Here's a quick example... this won't compile, but it should clarify the logic I would use:

render_container_tabbed():

	struct sway_container *active_child = NULL;
	for (int i = 0; i < con->children->length; ++i) {
		struct sway_container *child = con->children->items[i];
		if (child is focused) {
			class = &config->border_colors.focused;
			texture = &child->title_focused;
			active_child = child;
		} else if (child is focused_inactive) {
			class = &config->border_colors.focused_inactive;
			texture = &child->title_focused_inactive;
			active_child = child;
		} else if (child is unfocused) {
			class = &config->border_colors.unfocused;
			texture = &child->title_unfocused;
		}

        // Render title background for this tab here
        // Render single pixel lines above and below title here

		struct wlr_box box = {
			.x = con->x + config->border_thickness,
			.y = con->y + config->border_thickness,
			.width = con->width / con->children->length - config->border_thickness * 2,
			.height = con->title_height,
		};
		wlr_renderer_scissor(renderer, &box);
		wlr_render_texture(texture)
		wlr_renderer_scissor(renderer, NULL);
	}
	if (!sway_assert(active_child, "tabbed layout has no active child")) {
		return;
	}

	// Render left, bottom and right borders here - copy from another function and discard indicator border
	// Render actual view

SirCmpwn: I think we should probably restructure the top-level loop to live entirely in the output code, instead of relying on the filtered map kind of deal. Then when we encounter a tabbed layout we can render its borders right there and worry about the children in recursive iterations.

SirCmpwn: To accomplish this containers need to grow their own borders and some sort of bool should exist in the recursive function to indicate that child borders should no longer be drawn for subsequent iterations.

SirCmpwn: Sorry, I have no clue what I'm talking about, disregard that

Dudemanguy911: Ah, that makes sense @RyanDwyer. Thanks for the explanation.

Dudemanguy911:
Oh no, I just realized this implementation has strange mouse issues with gtk3 applications. It seems like gtk3 windows try to automatically grab the focus for some reason. Since I have disabled the setting the focus on click (to avoid the first child from automatically grabbing the focus), opening a gtk3 application will disable the clicks. qt5 applications don't seem to do this and behave normally. Very strange issue. I'll have to look into it some more.

SirCmpwn:
I still get mouse input going to the first child whenever I move over the tabbed layout. If focus_follows_mouse is on then moving the mouse over the tabbed layout focuses the first child.

SirCmpwn:
Also, changing the border of any of the children messes things up.

Dudemanguy911:
All of the widths and heights that are calculated are stored in a wlr_box before being rendered. The cast will break any case where the number of tabs don't perfectly divide into integers.

SirCmpwn:
Where is the code that does this calculation?

Dudemanguy911:
Well for example take what renders one of the tab separators.
```box.x = con->x + depth * tab_width;
box.y = con->y + 1; 
box.width = con->sway_view->border_thickness;
box.height = con->sway_view->y - con->y - 2;```

`tab_width` here is calculated from `tab_width = con->width / num_tabs` which I have stored as float (which probably should be changed to double really). `box.x` tells the renderer where to start drawing the box. In case of tabs, this is iterated through with `depth` which controls where each tab starts. However, box.x is an [int](https://github.com/swaywm/wlroots/blob/master/include/wlr/types/wlr_box.h). So the proper value will get its decimals truncated and cast into an integer which will offset the borders. I imagine it probably gets worse the more and more windows you open up.

**Comment by SirCmpwn** [comment]
What we do [elsewhere](https://github.com/swaywm/sway/blob/master/sway/tree/arrange.c#L123-L160), and should probably do here, is set aside some doubles for temporary use and sort them out into ints later on. wlr_box should not use doubles, you can't subdivide pixels.

**Comment by Dudemanguy911** [comment]
Nevermind, I'm stupid. I can just use floor and ceiling. Now there's still the mouse thing to fix.

**Comment by Dudemanguy911** [comment]
I've come up with a fix for the mouse focus problems (which in reality boiled down to fetching the correct container). Let me know what you think.

**Comment by SirCmpwn** [comment]
I still have the issue with changing the borders of the children at runtime making things break.

**Comment by Dudemanguy911** [comment]
What's breaking exactly?

**Comment by SirCmpwn** [comment]
1. Open a few windows in a tabbed layout
2. Execute `border toggle`
3. Observe breakage

**Comment by SirCmpwn** [comment]
Looking better, but I can still get things to look wrong by making the children of my tabbed layout have different borders and switching between them. For example:

1. Open two windows in tabbed layout and focus the first
2. `border toggle`
3. `focus right`

We should probably

1. Always use the first child's border setting for the whole tabbed layout
2. Make any border toggle command for the child of a tabbed container affect the first child

**Comment by SirCmpwn** [comment]
Not sure if this is directly related to your changes, but if I prepare the following layout:

```
(workspace L_VERT
  (view A)
  (container L_TABBED
    (view B focus)
    (view C)))
```

Then run `border toggle`, focus is moved to view A. It should stay on view B.

SirCmpwn:
Those problems aside this is looking really solid.

SirCmpwn:
>The other approach would be to to modify the toggle/hide command for tabbed layouts and have it change all the borders in the parent layout.

Hey I like this idea

>I've never actually used i3 before, so I don't know what it did in this case. Did it just hide the entire bar?

Actually i3 forces the border for tabbed windows to always be "normal". I like supporting alternate border styles though.

>This happens on the master branch as well if you open up two different windows and then toggle the border on the second one.

Alright I'll file a bug

Dudemanguy911:
I should have added [WIP] ages ago. Tabbed borders should share all of their settings now (thickness, type and toggle). Some things left to do. Holler at me if you find something else.

- [ ] fix crashing if you invoke a split command on a tabbed layout (this is new)
- [ ] make swaybar clickable/have the correct focus on the first workspace (works on other workspaces)

SirCmpwn:
Hey, if you get this finished up today I can ship it in 1.0-alpha.2. No rush if you don't have time, though.

PR #1915, opened 2018-05-05
Implement title_format

RyanDwyer:
This implements the `title_format` command, with a new placeholder `%shell` which gets substituted with the view type (xwayland, xdg_shell_v6 or wl_shell).

Example config:

    for_window [title=".*"] title_format %title (class=%class instance=%instance shell=%shell)

You can change the title using swaymsg:

    $ swaymsg title_format %title

Some implementation notes:
* container_update_title() has been renamed to view_update_title() and moved into view.c as it only applies to views.
* Because of the above, container_notify_child_title_changed() is now public.
* The arguments of view_update_title() have changed. It used to have the title given to it via an argument; now it reads it from view_get_title().
* Containers now have both a `name` and `formatted_title` property. The `name` is used purely for `strcmp()`, so we don't blindly recreate the title textures if the view hasn't changed its title.

I've also tested it will null titles (Firefox developer tools as per #1912).

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -63,7 +63,8 @@ struct sway_container {
 	 */
 	size_t id;
 
-	char *name;
+	char *name;            // The view's title (unformatted)
+	char *formatted_title; // The title displayed in the title bar
 
 	enum sway_container_type type;
 	enum sway_container_layout layout;
@@ -204,7 +205,6 @@ void container_update_title_textures(struct sway_container *container);
  */
 void container_calculate_title_height(struct sway_container *container);
 
-void container_update_title(struct sway_container *container,
-		const char *new_title);
+void container_notify_child_title_changed(struct sway_container *container);
 
 #endif
--- a/sway/commands/title_format.c
+++ b/sway/commands/title_format.c
@@ -2,7 +2,6 @@
 #include <string.h>
 #include "sway/commands.h"
 #include "sway/config.h"
-#include "sway/tree/arrange.h"
 #include "sway/tree/view.h"
 #include "log.h"
 #include "stringop.h"
@@ -25,7 +24,6 @@ struct cmd_results *cmd_title_format(int argc, char **argv) {
 	}
 	view->title_format = format;
 	view_update_title(view, true);
-	config_find_font_height(true);
-	arrange_root();
+	config_update_font_height(true);
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
--- a/sway/commands/title_format.c
+++ b/sway/commands/title_format.c
@@ -1,0 +1,31 @@
+#define _POSIX_C_SOURCE 200809L
+#include <string.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/tree/arrange.h"
+#include "sway/tree/view.h"
+#include "log.h"
+#include "stringop.h"
+
+struct cmd_results *cmd_title_format(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "title_format", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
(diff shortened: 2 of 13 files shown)


sway/commands/title_format.c
```diff
@@ -0,0 +1,31 @@
...
+	if (container->type != C_VIEW) {
+		return cmd_results_new(CMD_INVALID, "title_format",
+				"Only views can have a title_format");
+	}
+	struct sway_view *view = container->sway_view;
+	char *format = join_args(argv, argc);
+	if (view->title_format) {
+		free(view->title_format);
+	}
+	view->title_format = strdup(format);
```
taiyu-len: dont need the strdup here,
join_args already allocates memory for format, which is also is getting leaked here.

RyanDwyer: Thanks!

sway/tree/view.c
```diff
@@ -475,3 +493,127 @@ void view_child_destroy(struct sway_view_child *child) {
...
+
+	// Now build the title
+	format = view->title_format;
+	next = strchr(format, '%');
+	while (next) {
+		// Copy everything up to the %
+		strncat(buffer, format, next - format);
+		format = next;
+
+		if (strncmp(next, "%title", 6) == 0) {
```
emersion: Is it possible to factor this into a function that returns the formatted length and only calls `strcat` if the `buffer` parameter isn't NULL? We should be able to call it first with `buffer` set to NULL, then allocate the buffer, and then call it again. That would de-duplicate this code.

sway/tree/view.c
```diff
@@ -475,3 +493,127 @@ void view_child_destroy(struct sway_view_child *child) {
...
+		if (title && view->swayc->name && strcmp(title, view->swayc->name) == 0) {
+			return;
+		}
+		if (!title && !view->swayc->name) {
+			return;
+		}
+	}
+
+	if (title) {
+		if (view->swayc->name) {
```
emersion: No need for these `if` statements: `free(NULL)` is a no-op.

emersion: You can factor them out of the outer `if`

sway/commands/title_format.c
```diff
@@ -0,0 +1,31 @@
...
+	}
+	struct sway_view *view = container->sway_view;
+	char *format = join_args(argv, argc);
+	if (view->title_format) {
+		free(view->title_format);
+	}
+	view->title_format = format;
+	view_update_title(view, true);
+	config_find_font_height(true);
+	arrange_root();
```
emersion: Is `arrange_root` necessary here?

sway/tree/view.c
```diff
@@ -475,3 +493,127 @@ void view_child_destroy(struct sway_view_child *child) {
 		free(child);
 	}
 }
+
+static char *parse_title_format(struct sway_view *view) {
+	if (!view->title_format || strcmp(view->title_format, "%title") == 0) {
+		return strdup(view_get_title(view));
```
emersion: `strdup` will segfault if `view_get_title` returns NULL

RyanDwyer: `parse_title_format` is only called if the title is not NULL. Should I still put a check in here?

include/sway/tree/container.h
```diff
@@ -63,7 +63,8 @@ struct sway_container {
 	 */
 	size_t id;
 
-	char *name;
+	char *name;            // The view's title (unformatted)
```
emersion: Nitpick: maybe this should be renamed to `title` for consistency?

RyanDwyer: I kind of agree, but that also means we'd be changing `output->name` to `output->title` and `workspace->name` to `workspace->title`, and I feel like `name` is a better name for those.

emersion: Yes. you're right.

SirCmpwn:
Thanks!

PR #1918, opened 2018-05-05
Use scissor to render title texture

RyanDwyer:
This allows the title's texture to always be the full width of the text, and clipped at render time according to the desired width (eg. tabs...).

As an added bonus, the texture no longer needs to be updated when containers are arranged.


SirCmpwn:
Thanks!

PR #1920, opened 2018-05-05
meson.build: Try looking for `pam` when `libpam` isn't found.

yorickvP:
On my system, libpam is called `libpam.so`, which would need the compiler flag `-lpam`. This change keeps the old lookup, but adds a fallback to `cc.find_library('pam')`, which makes it work for me.


SirCmpwn:
What system is that?

yorickvP:
NixOS. Is it called `liblibpam.so` on arch?!

SirCmpwn:
Haha, no. But what's odd is that your system finds libinput, libcap, etc just fine. That it doesn't find this one is weird, and sounds like a deeper issue.

yorickvP:
libinput, libcap, etc are found with pkg-config, this one is found using `cc.find_library`.

martinetd:
This is actually meson being weird with `find_library` - since the nix store for libpam is considered a "system" path it only looks for liblibpam there but in the "normal" paths (e.g. the gcc store's libdir) it looks for both liblibpam and libpam

I haven't spent the time to figure where meson finds these paths and decide which is what, but my opinion is just to change that to 'pam' for everyone instead of making this weird check. The other libs are like that anyway so it shouldn't cause any problem and that'll save meson looking for liblibpam then libpam everywhere...

SirCmpwn:
Can you report this bug to meson?

martinetd:
I will once I figure where the paths come from, but should change both imo :)

SirCmpwn:
aye

martinetd:
Ah, actually these paths aren't exposed to meson at all but brought in through a gcc wrapper script.
meson tries to build with -llibpam then falls back to manual searching which is why it works on other distros, but there's no way that manual searching would work on nixos because it doesn't expose the library paths in a normal manner.

Once that failed it does `cc --print-search-dirs` and looks in the `libraries` path for `libfoo` and `foo` (so liblibpam and libpam) and ends up finding libpam on most systems.

I'll open a meson issue anyway (suggesting to add a second gcc call if the argument starts with lib) but it's more a RFE than a bugfix -- I wonder if we could get the nixos wrapper to add all the paths to --print-search-dirs as well...

SirCmpwn:
Thanks!

PR #1923, opened 2018-05-05
Implement full damage tracking

emersion:
Uses scissors for now, will improve this in the future.

- [x] Test scaled outputs
- [x] Test output transforms
- [x] Layer shell surface unmap

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -37,8 +37,8 @@ void output_damage_whole(struct sway_output *output);
 void output_damage_surface(struct sway_output *output, double ox, double oy,
 	struct wlr_surface *surface, bool whole);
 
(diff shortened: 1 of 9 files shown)


emersion:
This PR is ready for review.

Will run this branch as my daily driver for a while to try to find any bug.

atomnuker:
Completely breaks scaling.

RyanDwyer:
Please rebase. I was going to do some DRM testing for this but I can't compile it because wlroots has had backwards breaking changes.

atomnuker:
Works perfectly, can't find anything that breaks (which isn't already broken by the lack of atomic layout updating, like mpv randomly losing the frame callback when resizing).

SirCmpwn:
Great work, thanks!

PR #1881, opened 2018-05-01
Install swaybar and swaybg under /usr/lib/sway

nbraud:
Closes #1880

PR #1927, opened 2018-05-05
Revert PR #1881

atomnuker:
1.) has an unneeded / in meson.build which causes the PATH env variable to have an incorrectly appended libexec directory for sway
2.) breaks launching swaybar and swaybg at startup with a message "sh: swaybar[/bg]: command not found"


SirCmpwn:
@nbraud looks like this has some issues, take another crack at it and submit a new PR please

nbraud:
OK, that's exciting since it didn't come up in local testing (OTOH, syncing that branch with my `debian/patches` files was wonky, so I might just have missed some commits in one branch or the other)

PR #1928, opened 2018-05-06
Criteria improvements

RyanDwyer:
Adds support for the following criteria tokens: `id` (X11 window ID), `instance`, `tiling`, `workspace`.

Ensures each view will only execute a criteria once, by storing a list of executed criteria in the view. This is the same strategy used by i3.

This also adds a check for criteria when an xwayland view changes its title, class, instance or window type.

Some implementation notes:

* `get_prop` has been split into `get_string_prop` and `get_int_prop` because some properties like the X11 window ID, window type and role are numeric.
* I started implementing the window_type and window_role tokens, but I think modifications are needed to wlroots to complete it. I've left some supporting code in place for this (event listeners and `view_get_window_{type,role}` functions).
* I discovered xwayland views trigger the `set_title` event when they change their title, so I've moved the call to `view_update_title` out of `handle_commit` and into there. xdg_shell_v6 still does it in `handle_commit` though.

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -126,7 +126,7 @@ static struct sway_xwayland_view *xwayland_view_from_view(
 	return (struct sway_xwayland_view *)view;
 }
 
-static const char *get_prop(struct sway_view *view, enum sway_view_prop prop) {
+static const char *get_string_prop(struct sway_view *view, enum sway_view_prop prop) {
 	if (xwayland_view_from_view(view) == NULL) {
 		return NULL;
 	}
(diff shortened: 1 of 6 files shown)


sway/desktop/xwayland.c
```diff
@@ -285,6 +306,40 @@ static void handle_request_fullscreen(struct wl_listener *listener, void *data)
 	view_set_fullscreen(view, xsurface->fullscreen);
 }
 
+static void handle_set_title(struct wl_listener *listener, void *data) {
+	struct sway_xwayland_view *xwayland_view =
+		wl_container_of(listener, xwayland_view, set_title);
+	struct sway_view *view = &xwayland_view->view;
+	struct wlr_xwayland_surface *xsurface = view->wlr_xwayland_surface;
+	if (!xsurface) {
```
emersion: You probably want to check if the xwayland surface is mapped here? xsurface won't be NULL.

RyanDwyer:
On second thought, I'm going to overhaul the criteria system and submit it as a new PR.

PR #1932, opened 2018-05-07
layer-shell: use usable_area when arranging non-exclusive layer surfaces

emersion:
And arrange them in the correct order.


sway/desktop/layer_shell.c
```diff
@@ -180,16 +180,13 @@ void arrange_layers(struct sway_output *output) {
 	}
 
 	// Arrange non-exlusive surfaces from top->bottom
```
SirCmpwn: comment is outdated but I think your code is wrong, we're supposed to arrange them from top to bottom

emersion: In the end I don't care a lot about the order, I just want the non-exclusive layer surfaces to respect the exclusive ones. Pushed a new commit that does this.

PR #1935, opened 2018-05-07
Fix crash when using pango markup font

heghe:
The characters & < > ' " needs to be escaped when using pango markup

Signed-off-by: Heghedus Razvan <heghedus.razvan@protonmail.com>

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -8,6 +8,68 @@
 #include <string.h>
 #include "log.h"
 
+int escape_markup_text(const char *src, char *dest, int dest_length) {
+	int length = 0;
+
+	while (src[0]) {
+		switch (src[0]) {
+		case '&':
+			length += 5;
+			if (dest && dest_length - length >= 0) {
+				dest += sprintf(dest, "%s", "&amp;");
+			} else {
+				dest_length = -1;
+			}
+			break;
+		case '<':
+			length += 4;
+			if (dest && dest_length - length >= 0) {
+				dest += sprintf(dest, "%s", "&lt;");
+			} else {
+				dest_length = -1;
+			}
+			break;
+		case '>':
+			length += 4;
+			if (dest && dest_length - length >= 0) {
+				dest += sprintf(dest, "%s", "&gt;");
+			} else {
+				dest_length = -1;
+			}
+			break;
+		case '\'':
+			length += 6;
+			if (dest && dest_length - length >= 0) {
+				dest += sprintf(dest, "%s", "&apos;");
+			} else {
+				dest_length = -1;
+			}
+			break;
+		case '"':
+			length += 6;
+			if (dest && dest_length - length >= 0) {
+				dest += sprintf(dest, "%s", "&quot;");
+			} else {
+				dest_length = -1;
+			}
+			break;
+		default:
+			length += 1;
+			if (dest && dest_length - length >= 0) {
+				*(dest++) = *src;
+			} else {
+				dest_length = -1;
+			}
+		}
+		src++;
+	}
+	// if we could not fit the escaped string in dest, return -1
+	if (dest && dest_length == -1) {
+		return -1;
+	}
+	return length;
+}
+
 PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
 		const char *text, int32_t scale, bool markup) {
 	PangoLayout *layout = pango_cairo_create_layout(cairo);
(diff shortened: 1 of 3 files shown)


SirCmpwn:
This code is liable to cause buffer overflows

heghe:
That's true in case of a long string, which might happend when it's used by the bar. Should I have a dynamic allocated array and double it's length in case it's not enough the first 2048 bytes?

SirCmpwn:
Measure the new length ahead of time and allocate it once.

emersion:
You can do something similar to this: https://github.com/swaywm/sway/pull/1915/files#diff-819010104a02bd78a3801147b06a4d57R501

common/pango.c
```diff
@@ -8,21 +8,72 @@
...
+			length += 6;
+			break;
+		default:
+			length += 1;
+		}
+	}
+	return length;
+}
+
+void escape_markup_text(char *dest, const char* src) {
```
emersion: Can this be static?

common/pango.c
```diff
@@ -8,21 +8,72 @@
 #include <string.h>
 #include "log.h"
 
+static size_t escaped_text_length(const char *text) {
```
emersion: Can you merge these two functions, so that if we change one value we don't forget to change the other?

(ie. if `src` is NULL, don't try to write to it, and always return the size)

common/pango.c
```diff
@@ -8,21 +8,72 @@
...
+			length += 1;
+		}
+	}
+	return length;
+}
+
+void escape_markup_text(char *dest, const char* src) {
+	char c;
+
+	while ((c = *(src++))) {
```
emersion: Not a fan of this, it's getting unreadable. I'd prefer to use two lines instead.

common/pango.c
```diff
@@ -8,21 +8,72 @@
...
+			dest += sprintf(dest, "%s", "&gt;");
+			break;
+		case '\'':
+			dest += sprintf(dest, "%s", "&apos;");
+			break;
+		case '"':
+			dest += sprintf(dest, "%s", "&quot;");
+			break;
+		default:
+			*(dest++) = c;
```
emersion: Ditto

common/pango.c
ascent12: This could actually just be an iteration over `src`. Using `c` is unnecessary.

emersion:
>The characters & < > ' " needs to be escaped when using pango markup

Here you always escape the markup, making it impossible to actually _use_ the markup. Users should be able to set the title format to something like `<b>%title</b>` to make window titles bold. You should instead escape text provided by the client embedded in the markup (such as window titles).

heghe:
That's true. I didn't think about this. I see two solutions:
1. Parse the string an escape ony the characters which are not used in tags or other markup related usecases. This solution is a bit harder and more complex.
2. Escape the string before putting it in the title. This is more simple but I wouldn't take this aproach.

I'll work on the first method, but it will take some time.

emersion:
Well, we don't want an app to be able to use markup in its title, so (2) is probably better.

heghe:
Oh. ok. I'll do (2) then

include/pango.h
```diff
@@ -6,6 +6,17 @@
...
+ *
+ * If the dest parameter is NULL, then the function returns the length of
+ * of the escaped src string. The dest_length doesn't matter.
+ *
+ * If the dest parameter is not NULL then the fuction escapes the src string
+ * an puts the escaped string in dest and returns the lenght of the escaped string.
+ * The dest_length parameter is the size of dest array. If the size of dest is not
+ * enough, then the function returns -1.
+ */
+int escape_markup_text(const char* src, char *dest, int dest_length);
```
SirCmpwn: Style - move the * next to the parameter name

SirCmpwn:
Also, please associate your committer email with your GitHub account.

emersion:
Can you rebase your PR please?

common/pango.c
```diff
@@ -8,20 +8,82 @@
...
 	PangoAttrList *attrs;
 	if (markup) {
 		char *buf;
 		GError *error = NULL;
-		if (!sway_assert(pango_parse_markup(
-					text, -1, 0, &attrs, &buf, NULL, &error),
-				"pango_parse_markup '%s' -> error %s", text,
+		bool result = pango_parse_markup(text, -1, 0, &attrs, &buf,
+				NULL, &error);
+		if (!sway_assert(result, "pango_parse_markup '%s' -> error %s", text,
```
emersion: Can you remove this assertion and just print an error please?

heghe:
Yeah, I'll change that. I have done the rebase and push it before I saw the other comment

emersion:
No problem, take your time ;)

emersion:
Thanks!

PR #1943, opened 2018-05-09
Overhaul criteria implementation

RyanDwyer:
This builds off PR #1928 which I submitted several days ago, then decided to close in favour of overhauling the criteria system completely.

The criteria struct now uses **properties for each token type** rather than the list_t list of tokens. The reason for this is that different token types have different data types: pcre, string and number to name a few. Tokens like urgent and window_type are enum-ish too. A bonus of this is that criteria is now easier to understand when looking at the struct definition.

The criteria **parser has been rewritten** because the previous one didn't support valueless pairs (eg. `[class="foo" floating]`).

Criteria now has **types**. Types at the moment are `CT_COMMAND`, `CT_ASSIGN_WORKSPACE` and `CT_ASSIGN_OUTPUT`. i3 uses types as well. Previously the assign command was creating a criteria with `move to workspace <name>` as its command, but this caused the window to appear briefly on the focused workspace before being moved to the assigned workspace. It now creates the view directly in the assigned workspace.

Each view will only **execute a given criteria once**. This is achieved by storing a list of executed criteria in the view. This is the same strategy used by i3.

**Escaping** now works properly. Previously you could do things like `[class="Fire\"fox"]` and the stored value would be `Fire\"fox`, but it should be (and now is) `Fire"fox`.

The public functions in criteria.c are now all prefixed with `criteria_`.

Xwayland views now listen to the `set_title`, `set_class` and `set_window_type` events and criteria will be run when these happen. XDG shell has none of these events so it continues to update the title in
`handle_commit`.

Each view type's `get_prop` function has been split into `get_string_prop` and `get_int_prop` because some properties like the X11 window ID and window type are numeric.

The following new criteria tokens are now supported:

* id (X11 window ID)
* instance
* tiling
* workspace

Test cases:

* `for_window [class="^Firefox$"] opacity 0.9`
* `for_window [class=^Firefox$] opacity 0.9`
* `for_window [ class = ^Firefox$ ] opacity 0.9`
* `for_window [tiling] opacity 0.9`
* `assign [class="^Firefox$"] web`
* `swaymsg [class="^Firefox$"] opacity 0.9`

Try to break things:

* `swaymsg [invalid] opacity 0.9`
* `swaymsg [tiling`
* `swaymsg '[class="Firefox]"]' opacity 0.9` (should parse properly but won't match)
* `swaymsg '[class="Firefox]' opacity 0.9` (mismatched quote)
* `swaymsg [class] opacity 0.9` (value is required)
* `swaymsg [class=] opacity 0.9` (valid and will match any non-NULL class)
* `swaymsg '[class="Fire\"fox"]' opacity 0.9` (check debug log to see value is stored as `Fire"fox`)

### DIFF
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -1,42 +1,61 @@
 #ifndef _SWAY_CRITERIA_H
 #define _SWAY_CRITERIA_H
 
-#include "tree/container.h"
+#include <pcre.h>
 #include "list.h"
+#include "tree/view.h"
 
-/**
- * Maps criteria (as a list of criteria tokens) to a command list.
- *
- * A list of tokens together represent a single criteria string (e.g.
- * '[class="abc" title="xyz"]' becomes two criteria tokens).
- *
- * for_window: Views matching all criteria will have the bound command list
- * executed on them.
- *
- * Set via `for_window <criteria> <cmd list>`.
- */
-struct criteria {
-	list_t *tokens; // struct crit_token, contains compiled regex.
-	char *crit_raw; // entire criteria string (for logging)
-
-	char *cmdlist;
+enum criteria_type {
+	CT_COMMAND = 1 << 0,
+	CT_ASSIGN_OUTPUT = 1 << 1,
+	CT_ASSIGN_WORKSPACE = 1 << 2,
 };
 
-int criteria_cmp(const void *item, const void *data);
-void free_criteria(struct criteria *crit);
+struct criteria {
+	enum criteria_type type;
+	char *raw; // entire criteria string (for logging)
+	char *cmdlist;
+	char *target; // workspace or output name for `assign` criteria
 
-// Pouplate list with crit_tokens extracted from criteria string, returns error
-// string or NULL if successful.
-char *extract_crit_tokens(list_t *tokens, const char *criteria);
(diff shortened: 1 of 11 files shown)


ascent12:
>Escaping now works properly

What's the i3 behaviour for this? Having no escaping is good for writing regexes, so you're not doing things like `\\W'.

RyanDwyer:
I think i3 only unescapes backslashes the `\"` sequence: https://github.com/i3/i3/blob/1cdb1ab721deb9e2cead81d224f9f52390c73340/src/config_parser.c#L377

My implementation will only unescape sequences `\"` and `\\`. I'm happy to remove the latter so we behave exactly like i3.

SirCmpwn:
Please rebase

SirCmpwn:
This is looking great. Criteria has been in sore need of an overhaul for some time now.

If you update sway.5.txt I think this is ready to merge.

SirCmpwn:
Thanks!

SirCmpwn:
I was too quick to merge, gonna revert. After a bisect it seems that this causes both of these issues:

https://github.com/swaywm/sway/issues/1966

https://github.com/swaywm/sway/issues/1965

PR #1944, opened 2018-05-09
Swaybar: Respect pango_markup config

RyanDwyer:
Makes swaybar respect the user's `pango_markup` configuration in the workspace buttons and binding mode indicator.


SirCmpwn:
Thanks!

PR #1945, opened 2018-05-09
Fix titles on rotated outputs

RyanDwyer:
If the output is rotated, the scissor box needs to be transformed in the opposite rotation.

To test, launch with an output with a 90 degree rotation and open a view on that output. Before the fix, the title isn't shown.


SirCmpwn:
Thanks!

PR #1947, opened 2018-05-09
Fix null deref in server_init

dlrobertson:
If the backend fails to be created, log an error and immidiately return
from server_init.


emersion:
Thanks!

PR #1948, opened 2018-05-10
Highlight all child borders when using focus parent

RyanDwyer:
This adds a `bool parent_focused` argument to some of the `render_container` functions which gets passed down through the children and allows them to be rendered with focused colours.

To test, create a mix of horizontal and vertical layouts and run `focus parent`.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -65,27 +65,49 @@ static void seat_container_destroy(struct sway_seat_container *seat_con) {
 	free(seat_con);
 }
 
-static void seat_send_focus(struct sway_seat *seat,
-		struct sway_container *con) {
-	if (con->type != C_VIEW) {
-		return;
-	}
-	struct sway_view *view = con->sway_view;
-	if (view->type == SWAY_VIEW_XWAYLAND) {
-		struct wlr_xwayland *xwayland =
-			seat->input->server->xwayland;
-		wlr_xwayland_set_seat(xwayland, seat->wlr_seat);
-	}
-	view_set_activated(view, true);
-	struct wlr_keyboard *keyboard =
-		wlr_seat_get_keyboard(seat->wlr_seat);
-	if (keyboard) {
-		wlr_seat_keyboard_notify_enter(seat->wlr_seat,
-				view->surface, keyboard->keycodes,
-				keyboard->num_keycodes, &keyboard->modifiers);
+/**
+ * Activate all views within this container recursively.
+ */
+static void seat_send_activate(struct sway_container *con,
+		struct sway_seat *seat) {
+	if (con->type == C_VIEW) {
+		if (!seat_is_input_allowed(seat, con->sway_view->surface)) {
+			wlr_log(L_DEBUG, "Refusing to set focus, input is inhibited");
+			return;
+		}
+		view_set_activated(con->sway_view, true);
 	} else {
-		wlr_seat_keyboard_notify_enter(
-				seat->wlr_seat, view->surface, NULL, 0, NULL);
+		for (int i = 0; i < con->children->length; ++i) {
+			struct sway_container *child = con->children->items[i];
+			seat_send_activate(child, seat);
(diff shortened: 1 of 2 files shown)


SirCmpwn:
We also discussed activating all of the views under the focused container. Can you add that?

RyanDwyer:
Okay, have a look at that. It looks like xwayland (or maybe wlroots) doesn't allow multiple xwayland surfaces to be activated at the same time, but it works for XDG shell (using filezilla as my test).

It feels a little dirty because setting focus uses an iterator function, while unsetting focus just calls itself recursively. I prefer not using the iterator functions to be honest.

sway/input/seat.c
```diff
@@ -521,14 +538,15 @@ void seat_set_focus_warp(struct sway_seat *seat,
...
+		if (last_focus) {
+			unfocus(last_focus, seat);
 		}
 
 		if (container->type == C_VIEW) {
-			seat_send_focus(seat, container);
+			seat_send_focus_iterator(container, seat);
+		} else {
+			container_descendants(container, C_VIEW, seat_send_focus_iterator,
+					seat);
```
SirCmpwn: Something that bugs me but I don't generally mention is functions with bad balance. You don't have to split the line once you get to 80 chars, you can split it earlier too.

SirCmpwn:
Definitely see if you can make it consistently use iterators or explicit recursion.

SirCmpwn:
LGTM, but still have to test this tomorrow. We'll have to keep in mind how this interacts with stacked/tabbed as we implement those as well.

sway/input/seat.c
```diff
@@ -65,27 +65,34 @@ static void seat_container_destroy(struct sway_seat_container *seat_con) {
...
+		} else {
+			wlr_seat_keyboard_notify_enter(
+					seat->wlr_seat, view->surface, NULL, 0, NULL);
+		}
 	} else {
-		wlr_seat_keyboard_notify_enter(
-				seat->wlr_seat, view->surface, NULL, 0, NULL);
+		for (int i = 0; i < con->children->length; ++i) {
+			struct sway_container *child = con->children->items[i];
+			seat_send_focus(child, seat);
```
emersion: Hmm, when focusing a container, do we want to call `wlr_seat_keyboard_notify_enter` on each view?

RyanDwyer: I don't know. Do we? From the look of it, only one view can have keyboard focus so it ends up giving keyboard focus to the last view. People aren't likely to type stuff when using focus parent anyway.

emersion: Yeah, I was wondering if it would make more sense to clear the keyboard focus.

SirCmpwn: No, we don't want to call notify_enter.

RyanDwyer: I've pushed a commit to address this, however this causes a usability issue where if you have a single view on the workspace and then run `focus parent` (eg. by mistake), there's no visible indication that the workspace itself is focused and anything you type has no effect.

sway/input/seat.c
```diff
@@ -65,27 +65,51 @@ static void seat_container_destroy(struct sway_seat_container *seat_con) {
...
+	if (con->type == C_VIEW) {
+		if (!seat_is_input_allowed(seat, con->sway_view->surface)) {
+			wlr_log(L_DEBUG, "Refusing to set focus, input is inhibited");
+			return;
+		}
+		struct sway_view *view = con->sway_view;
+		if (view->type == SWAY_VIEW_XWAYLAND) {
+			struct wlr_xwayland *xwayland =
+				seat->input->server->xwayland;
+			wlr_xwayland_set_seat(xwayland, seat->wlr_seat);
```
emersion: This should be in `seat_send_focus`

SirCmpwn:
Please rebase

SirCmpwn:
Works great, thanks!

PR #1952, opened 2018-05-11
fix crash on fullscreen toggle

Dudemanguy911:
Currently, attempting to fullscreen while the window is inactive will crash Sway. This fixes that.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -163,13 +163,15 @@ void view_set_fullscreen_raw(struct sway_view *view, bool fullscreen) {
 		struct sway_container *focus, *focus_ws;
 		wl_list_for_each(seat, &input_manager->seats, link) {
 			focus = seat_get_focus(seat);
-			focus_ws = focus;
-			if (focus_ws->type != C_WORKSPACE) {
-				focus_ws = container_parent(focus_ws, C_WORKSPACE);
-			}
-			seat_set_focus(seat, view->swayc);
-			if (focus_ws != workspace) {
-				seat_set_focus(seat, focus);
+			if (focus) {
+				focus_ws = focus;
+				if (focus && focus_ws->type != C_WORKSPACE) {
+					focus_ws = container_parent(focus_ws, C_WORKSPACE);
+				}
+				seat_set_focus(seat, view->swayc);
+				if (focus_ws != workspace) {
+					seat_set_focus(seat, focus);
+				}
 			}
 		}
 	} else {


RyanDwyer:
What this code is doing is: If there are several views in a workspace, and two seats working in that workspace, and one of the seats fullscreens a view, it moves the other seat's focus to the fullscreen view. In other words, the intention is to iterate the seats, get the workspace they're currently on, and if it's the same workspace as the view being fullscreened then change their focus.

Changing it to be `seat_get_focus_inactive(seat workspace)` will always return a view on the workspace being fullscreened, which means the workspace will match and the seat's focus will be changed regardless of what workspace they're on.

What exactly is happening here? Is there no focus at all on the seat? Is that supposed to be possible?

Dudemanguy911:
Right now, the fullscreen function uses `seat_get_focus` which will fail if the window is inactive and thus result in sway crashing since the seat isn't technically focused. From what I understand, `seat_get_focus_inactive` will search and fetch the container that is focused or focused but inactive.

>Changing it to be seat_get_focus_inactive(seat, workspace) will always return a view on the workspace being fullscreened, which means the workspace will match and the seat's focus will be changed regardless of what workspace they're on.

I'm not sure how exactly you are using "workspace" here, but in the code it's just the parent container on the workspace. So yes, it will always return a view on a workspace if a window exists on it. I believe that's desirable behaviour. It won't return anything the workspace is empty and/or you're on a different one.

RyanDwyer:
> Right now, the fullscreen function uses `seat_get_focus` which will fail if the window is inactive

`seat_get_focus` should return the focused container across the entire tree for the given seat. If it's returning `NULL`, that's quite unusual. Perhaps that seat has never focused a container before, or all the containers it's focused have been killed. I think the correct solution here is to call `seat_get_focus` as it did, then check if focus is `NULL` and only execute the following code if focus is valid.

> From what I understand, seat_get_focus_inactive will search and fetch the container that is focused or focused but inactive.

That's correct, but it will only search within the given container, which in your case is the workspace of the fullscreen view. The code is trying to determine if the seat's focus is in the same workspace as the fullscreen view or not, so limiting the search area to the workspace is not of any benefit. It needs to be across the whole tree, which is why `seat_get_focus` is used.

> I'm not sure how exactly you are using "workspace" here, but in the code it's just the parent container on the workspace.

I don't understand what you mean by that.

> So yes, it will always return a view on a workspace if a window exists on it. I believe that's desirable behaviour.

That's not the desired behaviour. The code is trying to determine if the seat has focus in the workspace. If so, it'll force the seat's focus to be the fullscreen view. If it didn't, the seat would have focus on a view that is hidden behind the fullscreen view.

> It won't return anything the workspace is empty and/or you're on a different one.

My understanding is that `seat_get_focus_inactive` will return whatever was most recently focused within the container you pass to it. It doesn't care if you're currently on another workspace.

Dudemanguy911:
>I think the correct solution here is to call seat_get_focus as it did, then check if focus is NULL and only execute the following code if focus is valid.
Oddly enough, this even focuses the inactive workspace. I was wrong. It looks like this causes the crash, but I don't know why.
```
if (focus_ws->type != C_WORKSPACE) {
    focus_ws = container_parent(focus_ws, C_WORKSPACE);
}
```
Is that if statement really needed? Hell, is `focus_ws` needed?

RyanDwyer:
Yeah, that crash makes sense if `focus` and `focus_ws` are `NULL`.

`focus` is intended to be the actual container that has focus. `focus_ws` is the workspace that the focused container is sitting on. To find the workspace, you'd normally just use `container_parent`, but in our case `focus` might be a workspace itself (if the seat has focused an empty workspace). Hence why we need to wrap it in an if-statement.

There was talk of implementing a `container_self_or_parent` function which would operate similarly to `container_parent` but also consider the given container rather than just its parents. That would remove the need for an if-statement here and make the code easier to read. It's on my to do list.

> Hell, is focus_ws needed?

You could re-use the `focus` variable, but the meaning of the variable becomes incorrect if you do that.

Dudemanguy911:
Come to think of it, I think I'll just do  `focus && focus_ws->type ...` there. That should make everyone happy I think.

sway/tree/view.c
```diff
@@ -164,7 +164,7 @@ void view_set_fullscreen_raw(struct sway_view *view, bool fullscreen) {
 		wl_list_for_each(seat, &input_manager->seats, link) {
 			focus = seat_get_focus(seat);
 			focus_ws = focus;
-			if (focus_ws->type != C_WORKSPACE) {
+			if (focus && focus_ws->type != C_WORKSPACE) {
```
RyanDwyer: If `focus` is `NULL`, `focus_ws` will also be `NULL`. Then the if-statement on line 171 will always succeed, and you'll pass `NULL` to `seat_set_focus`. It looks like `seat_set_focus` will handle `NULL` gracefully, but I'd rather wrap lines 166-173 in an `if (focus)` check to make it obvious to the reader that this only applies if the seat has focus somewhere.

Dudemanguy911:
Oh I did change this by the way.

RyanDwyer:
Thanks!

PR #1953, opened 2018-05-11
Implement __focused__ criteria

RyanDwyer:
Merge after #1943.

This adds support for the `__focused__` keyword in criteria values.

To avoid excessive use of `strcmp`, I've added a `criteria_token` enum.


SirCmpwn:
Thanks!

PR #1954, opened 2018-05-11
Implement marks

RyanDwyer:
This implements the `mark` and `unmark` commands, including the `--add`, `--replace` and `--toggle` arguments for the `mark` command, and also implements mark matching in criteria.

To test, run a variety of these commands:

* `swaymsg mark foo` (replaces any existing marks on this view with `foo`)
* `swaymsg 'mark --add bar'` (will append `bar` to view's marks list)
* `swaymsg [con_mark="foo"] opacity 0.9`
* `swaymsg [con_mark="foo"] mark bar`
* `swaymsg 'mark --add --toggle bar'` (keep existing marks list and toggle `bar`)

Try to break it:

* `swaymsg mark`
* `swaymsg 'mark --add'`
* `swaymsg 'mark --asdf'`

Note that single quotes are required when using `--add` and friends, otherwise your shell may interpret them.

There's no easy well to get a view's marks (`show_marks` is not yet implemented), so my method is to test using the opacity example.

### DIFF
--- a/sway/commands/mark.c
+++ b/sway/commands/mark.c
@@ -57,7 +57,7 @@ struct cmd_results *cmd_mark(int argc, char **argv) {
 
 	view_find_and_unmark(mark);
 
-	if (!toggle || (toggle && !had_mark)) {
+	if (!toggle || !had_mark) {
 		list_add(view->marks, strdup(mark));
 	}
 
--- a/sway/commands/mark.c
+++ b/sway/commands/mark.c
@@ -1,0 +1,68 @@
+#define _POSIX_C_SOURCE 200809L
+#include <string.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/tree/view.h"
+#include "list.h"
+#include "log.h"
+#include "stringop.h"
+
+// mark foo                      Same as mark --replace foo
+// mark --add foo                Add this mark to view's list
+// mark --replace foo            Replace view's marks with this single one
+// mark --add --toggle foo       Toggle current mark and persist other marks
+// mark --replace --toggle foo   Toggle current mark and remove other marks
+
+struct cmd_results *cmd_mark(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "mark", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	struct sway_container *container =
+		config->handler_context.current_container;
+	if (container->type != C_VIEW) {
+		return cmd_results_new(CMD_INVALID, "mark",
+				"Only views can have marks");
+	}
+	struct sway_view *view = container->sway_view;
+
+	bool add = false, toggle = false;
+	while (argc > 0 && strncmp(*argv, "--", 2) == 0) {
+		if (strcmp(*argv, "--add") == 0) {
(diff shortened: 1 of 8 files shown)


RedSoxFan:
A few things:
1. In i3, the default method is `--replace` not `--add`
2. In i3, the command is `mark [--add|--remove] [--toggle] <identifier>`, which allows `--toggle` to be used with either `--add` or `--replace`.
   - `--add --toggle` keeps the list and toggles the mark. (what this PR does for `--toggle`)
   - `--replace --toggle` clears the marks and if it the mark wasn't in the list before clearing, adds it
3. `--add` and `--replace` are basically ignored in this implementation.
   - `--add` should append (what this PR is currently doing)
   - `--replace` should clear the marks before adding
4. This implementation allows for multiple views to have the same mark. When marking, i3 removes the mark from any view that has the mark.
5. In i3, `unmark` without any arguments removes all marks from the view.

RyanDwyer:
Alright, I'll revisit this.

RyanDwyer:
Okay, I've changed it to match i3's implementation and this is ready for review again.

sway/commands/mark.c
```diff
@@ -0,0 +1,68 @@
...
+	bool had_mark = view_has_mark(view, mark);
+
+	if (!add) {
+		// Replacing
+		view_clear_marks(view);
+	}
+
+	view_find_and_unmark(mark);
+
+	if (!toggle || (toggle && !had_mark)) {
```
RedSoxFan: This can be simplified to `!toggle || !has_mark`

sway/commands/unmark.c
```diff
@@ -0,0 +1,32 @@
...
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/tree/view.h"
+#include "list.h"
+#include "log.h"
+#include "stringop.h"
+
+struct cmd_results *cmd_unmark(int argc, char **argv) {
+	if (argc == 0) {
+		// Remove all marks from the current container
```
RedSoxFan: Sorry. I was incorrect in my statement previously. `unmark` should remove all marks from all views unless criteria is given. Example: `[con_id=__focused__] unmark` would only remove marks from the focused view

sway/commands/unmark.c
```diff
@@ -0,0 +1,32 @@
...
+		struct sway_container *container =
+			config->handler_context.current_container;
+		if (container->type != C_VIEW) {
+			return cmd_results_new(CMD_INVALID, "unmark",
+					"Only views can have marks");
+		}
+		view_clear_marks(container->sway_view);
+	} else {
+		// Remove a single mark from whichever container has it
```
RedSoxFan: Similar to above, `unmark <identifier>` should remove the mark from all views unless criteria is given

SirCmpwn:
`swaymsg mark foobar`

Did not see my terminal get the foobar mark

SirCmpwn:
Ah, I see you haven't done that part. I tested it with criteria and it worked fine.

RyanDwyer:
Just a note that I had to add `using_criteria` to the `handler_context` struct, because `unmark` needs to know if criteria is being used.

SirCmpwn:
Thanks!

PR #1955, opened 2018-05-11
Allow setting border widths for normal borders using default_border.

ggreer:
In Sway 0.15, `default_border normal 1` would set 1px wide borders. This recreates that behavior.


SirCmpwn:
Thanks!

PR #1956, opened 2018-05-11
cmd_move_container: Focus a window on the source workspace.

ggreer:
In Sway 0.15, moving a window to another workspace would cause a window on the source workspace to be focused. This restores that behavior, allowing you to quickly move a lot of windows to another workspace.


ggreer:
FYI, I didn't really take the time to understand this code. I just got it working to fix the behavior that was annoying me.

If somebody has better ideas, I'm all ears.

SirCmpwn:
Thanks!

Issue #1957, opened 2018-05-11
Workspace command segfault


SirCmpwn:
Not sure of the particular conditions that caused this:

```
#0  0x00007f786355986b in raise () at /usr/lib/libc.so.6
#1  0x0000557b8a37867e in _sway_assert (condition=false, format=0x557b8a38216c "[%s:%d] %s:container is NULL") at ../common/log.c:27
#2  0x0000557b8a3713aa in container_parent (container=0x0, type=C_OUTPUT) at ../sway/tree/container.c:402
#3  0x0000557b8a36ba25 in cmd_workspace (argc=1, argv=0x557b8ac4e668) at ../sway/commands/workspace.c:95
#4  0x0000557b8a34e8de in execute_command (_exec=0x557b8ab271a0 "workspace 3", seat=0x557b8ab8dfa0) at ../sway/commands.c:373
#5  0x0000557b8a360e99 in keyboard_execute_command (keyboard=0x557b8af4d5d0, binding=0x557b8ab27250) at ../sway/input/keyboard.c:99
#6  0x0000557b8a3610d2 in keyboard_execute_bindsym (keyboard=0x557b8af4d5d0, pressed_keysyms=0x557b8af4d610, modifiers=64, key_state=WLR_KEY_PRESSED) at ../sway/input/keyboard.c:167
#7  0x0000557b8a3616cf in handle_keyboard_key (listener=0x557b8af4d5e0, data=0x7ffefc047270) at ../sway/input/keyboard.c:355
#8  0x00007f786587afdb in wlr_signal_emit_safe (signal=0x557b8af375e8, data=0x7ffefc047270) at ../subprojects/wlroots/util/signal.c:29
#9  0x00007f786586136c in wlr_keyboard_notify_key (keyboard=0x557b8af374f0, event=0x7ffefc047270) at ../subprojects/wlroots/types/wlr_keyboard.c:125
#10 0x00007f7865847316 in handle_keyboard_key (event=0x557b8b0c0200, libinput_dev=0x557b8ac7dee0) at ../subprojects/wlroots/backend/libinput/keyboard.c:73
#11 0x00007f7865846f7b in handle_libinput_event (backend=0x557b8a9b2200, event=0x557b8b0c0200) at ../subprojects/wlroots/backend/libinput/events.c:213
[...]

PR #1959, opened 2018-05-11
Implement IPC get_seats command

RyanDwyer:
We've had a few issues lately where `seat_get_focus` returns `NULL` (see #1952 and #1957), and I'm wondering if maybe we have multiple seats at play. So I ran `swaymsg -t get_seats` and discovered the `get_seats` command doesn't exist. It probably should exist, so here it is.


include/ipc.h
```diff
@@ -14,6 +14,7 @@ enum ipc_command_type {
 	IPC_GET_VERSION = 7,
 	IPC_GET_INPUTS = 8,
 	IPC_GET_CLIPBOARD = 9,
+	IPC_GET_SEATS = 10,
```
SirCmpwn: It looks like i3 has added some new messages.

https://github.com/i3/i3/blob/next/include/i3/ipc.h

Why don't we bump our extensions up to 100+.

PR #1960, opened 2018-05-11
Implement hide_edge_borders

RedSoxFan:
Implements `hide_edge_borders <none|vertical|horizontal|both|smart>`
- `hide_edge_borders none` shows all borders
- `hide_edge_borders vertical` shows only horizontal borders
- `hide_edge_borders horizontal` show only vertical borders
- `hide_edge_borders both` hides all borders
- `hide_edge_borders smart` hides borders if the view is an only child, otherwise shows all borders


sway/commands/hide_edge_borders.c
```diff
@@ -0,0 +1,37 @@
...
+		config->hide_edge_borders = E_BOTH;
+	} else if (strcmp(argv[0], "smart") == 0) {
+		config->hide_edge_borders = E_SMART;
+	} else {
+		return cmd_results_new(CMD_INVALID, "hide_edge_borders",
+				"Expected 'hide_edge_borders "
+				"<none|vertical|horizontal|both|smart>'");
+	}
+
+	container_for_each_descendant_dfs(&root_container, _configure_view, NULL);
```
SirCmpwn: dfs is more expensive, do you actually need it?

RedSoxFan: Not really. Wasn't sure which to use. I'll switch it to bfs

SirCmpwn: Whoops, nevermind, bfs is the expensive one

sway/desktop/output.c
```diff
@@ -321,82 +322,96 @@ static void render_container_simple_border_normal(struct sway_output *output,
...
-		memcpy(&color, colors->child_border, sizeof(float) * 4);
-	}
-	color[3] *= con->alpha;
-	box.x = con->x + con->width - con->sway_view->border_thickness;
-	box.y = con->y + 1;
-	box.width = con->sway_view->border_thickness;
-	box.height = con->height - 1;
-	render_rect(output->wlr_output, output_damage, &box, color);
+	struct sway_container *workspace = container_parent(con, C_WORKSPACE);
+	int other_views = workspace->children->length - 1;
```
RyanDwyer: What if the workspace contains a single container, and that container contains several children? Wouldn't you need to count all the views recursively?

RedSoxFan: You are correct. I missed that case when testing. I'll update the code

RedSoxFan:
@RyanDwyer 
This is ready for another review. The views are now counted recursively for all descendants of the workspace and only counted when `hide_edge_borders` is set to `smart`.

emersion:
Since this PR changes the rendering code, please make sure all of this works with scaled and rotated outputs.

sway/tree/container.c
```diff
@@ -547,6 +547,20 @@ bool container_has_child(struct sway_container *con,
 	return container_find(con, find_child_func, child);
 }
 
+int container_count_descendants_of_type(struct sway_container *con,
+		enum sway_container_type type) {
+	int children = 0;
+	if (con->type == type) {
+		children++;
+	} else if (con->children) {
```
RyanDwyer: This works fine when passing `C_VIEW` to the function, but if we were to try to pass `C_CONTAINER` then it wouldn't descend into the container's children.

RedSoxFan:
> Since this PR changes the rendering code, please make sure all of this works with scaled and rotated outputs.

All modes for `hide_edge_borders` appear to be working with scaled and rotated outputs

RyanDwyer:
Thanks!

Issue #1965, opened 2018-05-12
Assertion failure when running qutebrowser on Xwayland


SirCmpwn:
This doesn't happen every time, but happens often.

```
#0  0x00007fc909a2786b in raise () at /usr/lib/libc.so.6
#1  0x000056406946c366 in _sway_assert (condition=false, format=0x564069475c40 "[%s:%d] %s:container_view_create called with NULL sibling/parent") at ../common/log.c:27
#2  0x0000564069464d8b in container_view_create (sibling=0x0, sway_view=0x56406a017360)
    at ../sway/tree/container.c:341
#3  0x0000564069469550 in view_map (view=0x56406a017360, wlr_surface=0x56406a41a560)
    at ../sway/tree/view.c:379
#4  0x000056406944fd1b in handle_map (listener=0x56406a0174d8, data=0x56406a44b540)
    at ../sway/desktop/xwayland.c:268
#5  0x00007fc90bd490ec in wlr_signal_emit_safe (signal=0x56406a44b690, data=0x56406a44b540)
    at ../subprojects/wlroots/util/signal.c:29
#6  0x00007fc90bd08eef in handle_surface_commit (wlr_surface=0x56406a41a560, role_data=0x56406a44b540) at ../subprojects/wlroots/xwayland/xwm.c:633
#7  0x00007fc90bd3af70 in surface_commit_pending (surface=0x56406a41a560)
    at ../subprojects/wlroots/types/wlr_surface.c:432
#8  0x00007fc90bd3b235 in surface_commit (client=0x564069e93670, resource=0x564069f4aba0)
    at ../subprojects/wlroots/types/wlr_surface.c:516
#9  0x00007fc9058f21c8 in ffi_call_unix64 () at /usr/lib/libffi.so.6
#10 0x00007fc9058f1c2a in ffi_call () at /usr/lib/libffi.so.6
#11 0x00007fc90a2047fd in  () at /usr/lib/libwayland-server.so.0
#12 0x00007fc90a200f79 in  () at /usr/lib/libwayland-server.so.0
#13 0x00007fc90a2029b2 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#14 0x00007fc90a20114c in wl_display_run () at /usr/lib/libwayland-server.so.0
#15 0x0000564069441e24 in server_run (server=0x5640696804a0 <server>) at ../sway/server.c:145
[...]

Issue #1963, opened 2018-05-12
sway crashes when launching program from dmenu


mulimoen:
Sway crashes with the last error message
[sway/tree/container.c:342] container_view_create:container_view_create called with NULL sibling/parent
Probably due to lack of NULL-check of "cont" from line 379 in sway/tree/view.c

Steps to reproduce:
1. start sway
2. launch dmenu
3. start program from dmenu

ascent12:
Yeah. I've run into this recently too. I was sitting on an old version for awhile, so I didn't notice it straight away.
Specifically launching things from dmenu is what's causing the crash. It doesn't happen when opening something from a hotkey or a terminal.
I've tried both Nautilus and Firefox, so it seems to affect both Wayland native and Xwayland programs.

Here is a captured stack trace:
```
#0  0x00007ff5c9fc686b in raise () at /usr/lib/libc.so.6
#1  0x000055a51a1b9338 in _sway_assert (condition=false, format=0x55a51a1c2c00 "[%s:%d] %s:container_view_create called with NULL sibling/parent") at ../common/log.c:27
#2  0x000055a51a1b1d6b in container_view_create (sibling=0x0, sway_view=0x55a51c50d350) at ../sway/tree/container.c:341
#3  0x000055a51a1b6522 in view_map (view=0x55a51c50d350, wlr_surface=0x55a51c4ee590) at ../sway/tree/view.c:377
#4  0x000055a51a19ccfb in handle_map (listener=0x55a51c50d4c8, data=0x55a51c50d150) at ../sway/desktop/xwayland.c:268
#5  0x00007ff5ca801095 in wlr_signal_emit_safe (signal=0x55a51c50d2a0, data=0x55a51c50d150) at ../util/signal.c:29
#6  0x00007ff5ca7c0f0f in handle_surface_commit (wlr_surface=0x55a51c4ee590, role_data=0x55a51c50d150) at ../xwayland/xwm.c:633
#7  0x00007ff5ca7f302a in surface_commit_pending (surface=0x55a51c4ee590) at ../types/wlr_surface.c:432
#8  0x00007ff5ca7f32ef in surface_commit (client=0x55a51c1164f0, resource=0x55a51c4de060) at ../types/wlr_surface.c:516
#9  0x00007ff5c79861c8 in ffi_call_unix64 () at /usr/lib/libffi.so.6
#10 0x00007ff5c7985c2a in ffi_call () at /usr/lib/libffi.so.6
[...]

PR #1968, opened 2018-05-12
Fix double free in criteria

RyanDwyer:
This restores the criteria implementation (after SirCmpwn's revert) and fixes the issue described in #1966. The fix is in a separate commit from the reverts for easy reviewing.

The config used in that issue uses string values for `id` which triggers errors. The caller frees the error, and when the criteria is parsed a second time it attempted to free the error again, causing the crash.

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -420,6 +420,7 @@ static void unescape(char *value) {
  * populated with an error string. It is up to the caller to free the error.
  */
 struct criteria *criteria_parse(char *raw, char **error_arg) {
+	*error_arg = NULL;
 	error = NULL;
 
 	char *head = raw;
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -190,19 +190,128 @@ static bool generate_regex(pcre **regex, char *value) {
 	return true;
 }
 
+enum criteria_token {
+	T_APP_ID,
+	T_CLASS,
+	T_CON_ID,
+	T_CON_MARK,
+	T_FLOATING,
+	T_ID,
+	T_INSTANCE,
+	T_TILING,
+	T_TITLE,
+	T_URGENT,
+	T_WINDOW_ROLE,
+	T_WINDOW_TYPE,
+	T_WORKSPACE,
+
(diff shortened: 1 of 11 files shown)


sway/criteria.c
```diff
@@ -417,10 +417,9 @@ static void unescape(char *value) {
  * criteria struct.
  *
  * If errors are found, NULL will be returned and the error argument will be
- * populated with an error string.
+ * populated with an error string. It is up to the caller to free the error.
  */
 struct criteria *criteria_parse(char *raw, char **error_arg) {
-	free(error);
 	error = NULL;
```
emersion: Maybe set `*error_arg` to NULL here, just to prevent bad surprises?

SirCmpwn:
Does this also introduce https://github.com/swaywm/sway/issues/1965 ?

RyanDwyer:
It does now.

I had somehow reached the conclusion that `seat_get_focus(seat)` was the same thing as `seat_get_focus_inactive(seat, &root_container)` and as part of the criteria overhaul, replaced it with `seat_get_focus`. I've since learned it's not the same thing. `seat_get_focus(seat)` does an additional check on the `seat->has_focus` property. So ignoring that check meant that `seat_get_focus` can rightfully return `NULL`.

I've changed it back to `seat_get_focus_inactive(seat, &root_container)` which is what the current master branch does and it works fine now.

emersion:
Can you provide a test plan for this PR?

RyanDwyer:
Sure.

Test sway won't crash on launch when criteria contains multiple errors by launching with the following config (proves #1966 is fixed):

    for_window [id="nautilus"] border pixel
    for_window [id="gnome-calculator"] floating enable; border pixel
    for_window [class="Chromium"] border none
    for_window [title=".*QEMU.*"] border none; fullscreen enable
    for_window [class="qutebrowser"] border none

Test sway doesn't crash when launching programs from dmenu (proves #1965 and #1963 are fixed):

* No special config needed, just launch a program from dmenu.

As well as the test plans in the original PR (#1943).

PR #1970, opened 2018-05-13
Add xdg-shell stable support

emersion:
Not having `default` cases in `switch` statements definitely helped.


atomnuker:
Works fine, just as well as zxdg_v6. Tested mpv, gedit and gnome-terminal.

PR #1978, opened 2018-05-14
Fix #1975

### LINKED ISSUE #1975: 1.0's hide_edge_borders behavior doesn't match 0.15's
https://github.com/swaywm/sway/pull/1960 implemented hide_edge_borders for sway 1.0. The man page says "Hide window borders adjacent to the screen edges.", but that PR hides all borders, including ones not adjacent to screen edges. Also, it hides titles if hide_edge_borders is set to "vertical" or "both".

RedSoxFan:
Fixes #1975 
- Window titles are not hidden by hide_edge_borders anymore
- Only borders on the screen edge are hidden now


RedSoxFan:
@ggreer 
Can you please test this when you get a chance?

sway/tree/view.c
```diff
@@ -120,51 +140,26 @@ void view_autoconfigure(struct sway_view *view) {
...
-				+ view->border_thickness * 2;
-			height = view->swayc->height - config->font_height
-				- view->border_thickness * 3;
-		} else {
-			y = view->swayc->y;
-			height = view->swayc->height;
-		}
+		// Height is: border + title height + border + view height + border
+		x = view->swayc->x + view->border_thickness * view->border_left;
+		y = view->swayc->y + config->font_height
```
RyanDwyer: I want to clarify something: The title bar height is border + title height + border as per the comment. The single pixel bars above and below the title are *always* 1px, and overlap the border. So if your border thickness is 3px, you'll have 1px of the "border" colour, 2px of the remaining border thickness ("background" colour), then the font/texture height for the title (also "background" colour), then the opposite for below the title.

Your implementation culls the top 1px pixel bar if border_top is false, but calculates the view's y as if it were culling one entire border (3px in the above example).

I think the best option is to not cull anything at the top if using a normal border.

RedSoxFan: If I understood your comment correctly, this should be fixed

RyanDwyer: The calculation looks good now.

You also need to change the rendering code so it always renders the top pixel row when using border normal.

RedSoxFan: Ok. I'll look at it again in a few hours after some sleep. I'm not seeing what I need to change at the moment. It looks like if border_top is false, then box.y would be con->y, which is the same y that the single pixel bar above the title is rendered at.

RyanDwyer: Yeah, I think there's still some confusion. My suggestion is to always render the full title bar and top border regardless of any hide_edge_borders setting, which means `render_container_simple_border_normal` should not contain any references to border_top.

RedSoxFan: Oh ok. Should be fixed now

SirCmpwn:
Thanks!

ggreer:
This fixes the issue. Thanks a lot.

PR #1979, opened 2018-05-14
Add hide_cursor command

### LINKED ISSUE #1471: Hide mouse cursor on inactivity
Just like `unclutter` does for X, it would be useful to have a setting for the amount of seconds of mouse inactivity after which to hide the mouse cursor.

ddevault:
Which hides the cursor if you don't use it for a while.

Fixes #1471

### DIFF
--- a/sway/commands/hide_cursor.c
+++ b/sway/commands/hide_cursor.c
@@ -1,0 +1,29 @@
+#define _POSIX_C_SOURCE 200809L
+#include <string.h>
+#include <wayland-util.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
+#include "log.h"
+#include "stringop.h"
+
+struct cmd_results *cmd_hide_cursor(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "hide_cursor", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+
+	char *end;
+	config->hide_cursor_timeout = strtol(argv[0], &end, 10);
+	if (*end) {
+		config->hide_cursor_timeout = 0;
+		return cmd_results_new(CMD_INVALID, "hide_cursor",
+				"Expected an integer timeout.");
+	}
+	if (config->hide_cursor_timeout < 100 && config->hide_cursor_timeout != 0) {
+		config->hide_cursor_timeout = 100;
+	}
+
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+}
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -342,6 +342,11 @@ The default colors are:
 	workspace (or current workspace), and _current_ changes gaps for the current
 	view or workspace.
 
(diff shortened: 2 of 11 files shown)


sway/sway.5.scd
```diff
@@ -342,6 +342,11 @@ The default colors are:
 	workspace (or current workspace), and _current_ changes gaps for the current
 	view or workspace.
 
+*hide\_cursor* <timeout>
+	Hides the cursor image after the specified _timeout_ (in milliseconds) has
+	ellapsed with no activity on that cursor. A timeout of 0 disables hiding
```
RyanDwyer: elapsed

sway/commands/hide_cursor.c
```diff
@@ -0,0 +1,26 @@
...
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
+#include "log.h"
+#include "stringop.h"
+
+struct cmd_results *cmd_hide_cursor(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "font", EXPECTED_EQUAL_TO, 1))) {
```
martinetd: s/font/hide_cursor/

martinetd:
A couple of issues when testing:
 * setting the timeout interactively doesn't affect a cursor until it moved. For example, from no timeout to some short timeout (e.g. 10ms) the cursor won't disappear until it moved.
 * setting the timeout to something really small (< 1 frame redraw time? try 10ms) will display as if the cursor didn't move when it really did move. Some slightly longer value (e.g. 100ms) work around the problem and work as expected.

SirCmpwn:
>setting the timeout interactively doesn't affect a cursor until it moved. For example, from no timeout to some short timeout (e.g. 10ms) the cursor won't disappear until it moved.

Yeah, I saw this coming and was okay with it. Fixing it will make a lot of unrelated systems talk to each other.

SirCmpwn:
Set the minimum to 100ms, also prevents you from setting a negative timeout.

sway/commands/hide_cursor.c
```diff
@@ -0,0 +1,29 @@
...
+
+	char *end;
+	config->hide_cursor_timeout = strtol(argv[0], &end, 10);
+	if (*end) {
+		config->hide_cursor_timeout = 0;
+		return cmd_results_new(CMD_INVALID, "hide_cursor",
+				"Expected an integer timeout.");
+	}
+	if (config->hide_cursor_timeout < 100) {
+		config->hide_cursor_timeout = 100;
```
martinetd: nice try, but trying to disable hide with hide_cursor 0 also sets it back to 100.

Once that's fixed I'm happy though ;)

SirCmpwn: dero

synaptiko:
Is there a chance this would be rebased and merged? I'm using xbanish on i3wm and this looks like a nice replacement. Or is there an alternative approach over ipc? I was trying to find something for Wayland and I found only a script which was moving the cursor to 0×0 position.

emersion:
Be the change you seek ;)

emersion:
Superseded by https://github.com/swaywm/sway/pull/3307

PR #1982, opened 2018-05-15
Implement show_marks

RyanDwyer:
The implementation is similar to titles: store a texture for each focus class.

Tests:

* Run `show_marks yes` and `show_marks no` in both config and swaymsg
* `swaymsg mark foo`
* Test with scaled outputs (I've already done this but doesn't hurt to double check)

If the title is long enough to intersect the mark then the mark is drawn on top. However, I've noticed an unrelated bug where the title isn't scissored properly and renders into the right side border if it's long enough. This makes a few pixels of it appear to the right of the mark which looks bad. I won't fix that as part of this PR because it's a separate issue.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -287,7 +287,6 @@ static void render_rect(struct wlr_output *wlr_output,
 		wlr_backend_get_renderer(wlr_output->backend);
 
 	struct wlr_box box = *_box;
-	scale_box(&box, wlr_output->scale);
 
 	pixman_region32_t damage;
 	pixman_region32_init(&damage);
@@ -313,6 +312,9 @@ damage_finish:
 
 /**
  * Render decorations for a view with "border normal".
+ *
+ * Care must be taken not to render over the same pixel multiple times,
+ * otherwise the colors will be incorrect when using opacity.
  */
 static void render_container_simple_border_normal(struct sway_output *output,
 		pixman_region32_t *output_damage,
@@ -320,19 +322,23 @@ static void render_container_simple_border_normal(struct sway_output *output,
 		struct wlr_texture *title_texture, struct wlr_texture *marks_texture) {
 	struct wlr_box box;
 	float color[4];
+	struct sway_view *view = con->sway_view;
+	float output_scale = output->wlr_output->scale;
 
-	if (con->sway_view->border_left) {
+	if (view->border_left) {
 		// Child border - left edge
 		memcpy(&color, colors->child_border, sizeof(float) * 4);
 		color[3] *= con->alpha;
 		box.x = con->x;
 		box.y = con->y + 1;
-		box.width = con->sway_view->border_thickness;
-		box.height = con->height - 1;
+		box.width = view->border_thickness;
+		box.height = con->height - 1
+			- view->border_thickness * view->border_bottom;
+		scale_box(&box, output_scale);
 		render_rect(output->wlr_output, output_damage, &box, color);
 	}
 
-	if (con->sway_view->border_right) {
+	if (view->border_right) {
 		// Child border - right edge
 		if (con->parent->children->length == 1
 				&& con->parent->layout == L_HORIZ) {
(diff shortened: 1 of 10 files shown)


sway/tree/view.c
```diff
@@ -770,3 +771,80 @@ bool view_has_mark(struct sway_view *view, char *mark) {
...
+static void update_marks_texture(struct sway_view *view,
+		struct wlr_texture **texture, struct border_colors *class) {
+	struct sway_container *output = container_parent(view->swayc, C_OUTPUT);
+	if (!output) {
+		return;
+	}
+	if (*texture) {
+		wlr_texture_destroy(*texture);
+	}
+	if (!view->marks->length) {
```
RedSoxFan: unmark results in a segmentation fault. Changing this to `!view->marks || !view->marks->length` fixes the issue.

RyanDwyer: Thanks. I think I've fixed this issue, but not using the suggestion you gave. `view->marks` should always be a valid list.

I replicated a segfault by marking a view then running `swaymsg unmark`. I realised I need to set `*texture` to `NULL` after destroying it. After doing that, unmark works correctly. I also copied this fix into the title texture code too - I guess it's not very common to update a title to be `NULL` which is why we haven't discovered it already.

sway/desktop/output.c
```diff
@@ -412,6 +412,25 @@ static void render_container_simple_border_normal(struct sway_output *output,
...
+		texture_box.x = (box.x + box.width) * output_scale - texture_box.width;
+		texture_box.y = (box.y + box.height)
+			* output_scale - texture_box.height;
+
+		float matrix[9];
+		wlr_matrix_project_box(matrix, &texture_box,
+			WL_OUTPUT_TRANSFORM_NORMAL,
+			0.0, output->wlr_output->transform_matrix);
+
+		render_texture(output->wlr_output, output_damage, marks_texture,
```
emersion: We should factor this into a function taking a texture and output-local coords.

sway/tree/view.c
```diff
@@ -43,6 +43,14 @@ void view_destroy(struct sway_view *view) {
 	}
 	list_free(view->marks);
 
+	if (view->marks_focused) {
+		// If one is set then all of these are set
```
emersion: No need to check for NULL

sway/tree/view.c
```diff
@@ -770,3 +779,86 @@ bool view_has_mark(struct sway_view *view, char *mark) {
...
+
+	size_t len = 0;
+	for (int i = 0; i < view->marks->length; ++i) {
+		char *mark = view->marks->items[i];
+		if (mark[0] != '_') {
+			len += strlen(mark) + 2;
+		}
+	}
+	char *buffer = calloc(len + 1, 1);
+	char *part = malloc(len + 1);
```
emersion: Should `sway_assert` these

RyanDwyer:
Okay, I've fixed a heap of opacity issues.

I had to remove the box scaling from `render_rect` and apply all the scaling in the border functions. This is because it needs to get the width of the title and marks textures to position things properly, and those widths are scaled, and we don't like unscaling them.

The title and marks textures are still slightly incorrect when using opacity. I'm not sure what's going on there. I'm not rendering anything behind it.

sway/tree/view.c
RyanDwyer: Done.

sway/desktop/output.c
RyanDwyer: I don't think you can... If you look at the title one, we have to adjust the texture box between the calls to `wlr_matrix_project_box` and `render_texture`.

Maybe I'm misunderstanding you though, so correct me if I'm wrong.

emersion: Yeah, with my patch merged in it's more complicated. We could pass the box instead, and force the box to be passed to `wlr_matrix_project_box` to have the size of the texture (to make sure the texture isn't improperly scaled).

RyanDwyer: I still don't understand. If you want to refactor it, can you do it as a separate PR? It's not a blocker to this one.

RyanDwyer:
The scaling issue is resolved. Please consider merging.

SirCmpwn:
Thanks!

PR #1991, opened 2018-05-16
sway: run commands without waiting for Xwayland

martinetd:
Xwayland is lazy now, there is no need to wait at all

Test plan: Try to add stuff to your .sway/config as 'exec' statements to run at startup, see nothing fails unexpectedly because we didn't wait enough.
(works for me: X11 app (urxvt), swayidle, swaylock)


emersion:
Thanks!

PR #1995, opened 2018-05-16
Change scale from int32_t to double for pango

### LINKED ISSUE #1985: Title text on fractionaly scaled outputs is not using the right size
E.g. try `scale 1.5`, the text seems to be rendered with `scale 1`.

RedSoxFan:
Since `pango_attr_scale_new` takes a `double`, this PR changes the pango-related function signatures from `int32_t scale` to `double scale` to allow for fractional scaling of text.

Fixes #1985


SirCmpwn:
Thanks!

PR #2005, opened 2018-05-19
Implement tabbed layout

RyanDwyer:
@Dudemanguy911's PR appears to have gone stale, so I picked up where he left off. I feel a bit bad for implementing this when he already has an open PR, but I'm super keep on tabs and want to move foward with it. Sorry @Dudemanguy911.

I had to change how `container_at` works because it simply iterated all the views and checked their position which won't work for tabbed, stacked or floating layouts. It now descends intelligently into the tree, calling a layout-specific function based on each ancestor's layout. This allows the tabbed layout to only descend into the active child, as well as implement tab-clicking to focus the container.

I changed the `container_at` function to expect output-local coords rather than global layout coords. Because of the recursion it now does, it'd be messy to make it expect global layout coords. It was only called once, and always with a workspace, so this is safe. I realise this might have to be revisited when floating is implemented, but this is heading in the right direction for now.

All the following should work correctly:

* Scaled outputs
* Doing a split within a tabbed container
* For splits, the tab's title is updated dynamically (eg. "V[Foo, Bar]") and recursively (eg. "V[H[Foo, Bar], Baz]")
* Clicking a tab should focus it (or in the case of splits, should focus whatever nested child has focus_inactive)
* `focus parent` works
* Clicking a view's surface should focus it, also in child splits


Dudemanguy911:
Nah that's cool. I got kinda stuck on splits. I'll go close mine.

sway/tree/container.c
```diff
@@ -411,80 +411,139 @@ struct sway_container *container_parent(struct sway_container *container,
...
 }
 
-struct sway_container *container_at(struct sway_container *parent,
-		double lx, double ly,
+static struct sway_container *container_at_view(struct sway_container *swayc,
+		double ox, double oy,
 		struct wlr_surface **surface, double *sx, double *sy) {
-	list_t *queue = get_bfs_queue();
-	if (!queue) {
+	struct sway_view *sview = swayc->sway_view;
```
emersion: Assert that `swayc` is a view

sway/tree/container.c
```diff
@@ -411,80 +411,139 @@ struct sway_container *container_parent(struct sway_container *container,
...
+			return container_at(child, ox, oy, surface, sx, sy);
 		}
 	}
+	return NULL;
+}
+
+struct sway_container *container_at(struct sway_container *parent,
+		double ox, double oy,
+		struct wlr_surface **surface, double *sx, double *sy) {
+	if (!sway_assert(parent->type >= C_WORKSPACE,
```
emersion: Meh

sway/tree/container.c
```diff
@@ -658,18 +717,82 @@ void container_calculate_title_height(struct sway_container *container) {
 	container->title_height = height;
 }
 
+/**
+ * Calculate and return the length of the concatenated child titles.
+ * An example concatenated title is: V[Terminal, Firefox]
+ * If buffer is not NULL, also populate the buffer with the concatenated title.
+ */
+static size_t concatenate_child_titles(struct sway_container *parent,
```
emersion: Hum, is that how i3 does it? Titles could end up being veeery long

emersion:
What about damage tracking?

sway/tree/container.c
RyanDwyer: I just checked and i3 uses the window class. I've changed it to use the class too, with fallback to app_id. Also, they don't use commas so I've removed those.

RyanDwyer:
Damage tracking is now done. I made a `view_is_visible` function to make this easier. I also made a `seat_get_active_child` function which returns the most recently active immediate child of the given container, which I use to determine which tab is active.

SirCmpwn:
Couple of issues.

This uses the border style of the focused view, but I think that makes moving between tabs jarring. It should just use the first child's border and make the border command on any child who isn't the first apply to the first instead.

If the border is `pixel [n]`, space for the full titlebar is still allocated. This causes rendering issues and cursor issues in the space it would have taken up. Screenshot - this tab is not focused:

output

I would expect this to draw the blue color all the way up to the bottom of the border along the top.

SirCmpwn:
Also, if border is set to none, I still see the left border.

RyanDwyer:
To implement the "use borders from first child" thing we would have to decide out what to do if the first child is a split container, and add code to handle that. It would also be jarring if the first view used a different thickness, and you moved it to the right or killed it. I don't think the complexity is worth the reward.

My intention was to have the border defined by `config->border_thickness` rather than the border thickness of the first view. This is still likely to match the user's desired thickness and would give a consistent size through all the tabs, no matter how you rearrange them.

Do you still want to go with the first-child solution?

RyanDwyer:
My latest commit fixes the two rendering/positioning issues SirCmpwn highlighted above. Additionally, a view's border settings (`border` and `border_thickness`) will now be respected for the left, bottom and right borders.

SirCmpwn:
>Do you still want to go with the first-child solution?

We should do first descendant, maybe?

SirCmpwn:
Or perhaps dispense with all of this and just make the tabbed container itself have borders...

RyanDwyer:
I just checked i3. The way it's implemented here is the same as i3's.

However, I did notice a rendering difference (i3's tabs have left/right borders as well as top and bottom). I'll fix that now.

SirCmpwn:
i3 forces it to use the title border style, doesn't it?

RyanDwyer:
No. It changes the left/bottom/right borders based on which view is focused.

SirCmpwn:
Alright, I guess we can live with it for now. I'll get this latest version tested shortly.

Issue #1875, opened 2018-04-30
Split swaylock into a standalone project


ddevault:

emersion:
How will we manage the common lib?

SirCmpwn:
Copying what we need and refactoring the rest away

RedSoxFan:
Related #3413

emersion:
We should also migrate issues to the new repos.

RedSoxFan:
I think github lets repo admins move issues between repos within an organization now. I'm not sure if anyone other than ddevault is a repo admin though.

However, some issues may be better off recreated and linked to in the existing issue due to spanning multiple items.

ddevault:
Yeah, I can. If anyone cares, prepare a list of issues and I'll move them.

PR #2009, opened 2018-05-20
slim client dependencies

Ongy:
This removes the wlroots dependency for clients.
The clients used wlr_log, to get rid of the dependency, the relevant
macros and functions are copied over and renamed to sway_log.
This also "fixes" that wlr_strip_path strips the wlr source path,
sway_strip_path will strip the sway source directroy.
It also makes the ipc an explicit dependency, not part of sway_common,
to allow building clients that don't require IPC without jsonc.


swayipc/meson.build
```diff
@@ -0,0 +1,8 @@
+lib_sway_ipc = static_library(
```
emersion: I'd name the IPC library just "ipc" (instead of "swayipc") or `common/ipc`, since sway* directories contain executables.

emersion:
This LGTM overall. This is a good first step towards #1875.

I wonder if we should continue to use `wlr_log` for sway itself.

ascent12:
I always thought having `wlr_log` itself as part of our API was a bit weird.
Originally, the user (sway) would use `wlr_log_init` into their own logging function.

emersion:
Yeah, that's true. +1 for doing this.

include/log.h
```diff
@@ -1,16 +1,52 @@
...
 
 void error_handler(int sig);
 
+typedef enum {
+	L_SILENT = 0,
+	L_ERROR = 1,
+	L_INFO = 2,
+	L_DEBUG = 3,
+	L_LAST,
+} log_importance_t;
```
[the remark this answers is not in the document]
Ongy: redefining errors. Which I'm not happy about, but I'm not aware of a better way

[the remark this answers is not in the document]
emersion: I mean: we should probably prefix this in wlroots with `wlr_` right?

SirCmpwn:
You need to split this up into several commits so it can be properly reviewed.

include/log.h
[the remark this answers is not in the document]
Ongy: Uff, they are just defines, so unless both headers are included it won't colide.
On the other hand that would make it easier to init things properly

sway/main.c
```diff
@@ -336,21 +337,24 @@ int main(int argc, char **argv) {
 	// TODO: switch logging over to wlroots?
```
SirCmpwn: This comment can probably go away

SirCmpwn:
Part of me is saddened by this patch. But part of me accepts it as necessary. If you don't mind killing that dead comment, I'll get this merged up.

sway/main.c
```diff
@@ -223,34 +223,35 @@ static void executable_sanity_check() {
...
 			cap_set_proc(caps)) {
-			wlr_log(L_ERROR, "Failed to drop extra capabilities");
+			sway_log(L_ERROR, "Failed to drop extra capabilities");
 			exit(EXIT_FAILURE);
 		}
 	}
 #endif
 }
 
+void wlr_log_init(log_importance_t verbosity, log_callback_t callback);
```
emersion: Should include the appropriate header instead of doing this

Ongy: that conflicts on the `L_DEBUG` etc. defines =.=

emersion: Yeah, that's why we need to prefix wlroots symbols.

emersion:
You should be able to use https://github.com/swaywm/wlroots/pull/1011 to properly fix this PR - let me know if you want me to help with that.

emersion:
Closing, there are too many conflicts and sway will continue to use `wlr_log`.

PR #2010, opened 2018-05-20
Pre-multiply alpha for non-textures

### LINKED ISSUE #1984: Semi-transparent borders look ugly
Try e.g. `client.focused #4c7899 #28557733 #ffffff #2e9ef4 #285577`

RedSoxFan:
Fixes #1984 

Background
=========
For textures, Cairo stores ARGB data with pre-multiplied alpha. Additionally, the shader is multiplying all four channels by con->alpha. So wlr_render_texture will render:

    R = SourceRed * SourceAlpha * con->alpha + DestRed * (1 - SourceAlpha * con->alpha)
    G = SourceGreen * SourceAlpha * con->alpha + DestGreen * (1 - SourceAlpha * con->alpha)
    B = SourceBlue * SourceAlpha * con->alpha + DestGreen * (1 - SourceAlpha * con->alpha)
    A = SourceAlpha * con->alpha + DestAlpha * (1 - SourceAlpha * con->alpha)

Anything rendered with wlr_render_rect does not use pre-multiplied alpha and we only multiply the alpha channel by con->alpha. So wlr_render_rect will render:

    R = SourceRed + DestRed * (1 - SourceAlpha * con->alpha)
    G = SourceGreen + DestGreen * (1 - SourceAlpha * con->alpha)
    B = SourceBlue + DestBlue * (1 - SourceAlpha * con->alpha)
    A = SourceAlpha * con->alpha + DestAlpha * (1 - SourceAlpha * con->alpha)

Summary
=======
After testing out both methods, it was obvious that wlr_render_texture is producing the correct result. wlr_render_rect has an issue with lighter colors not blending (fully opaque pure white - \#FFFFFFFF - will always render as \#FFFFFFFF regardless of con->alpha). For anyone that wants to compare it against this PR, I'll keep the following up for a bit: [redsoxfan:dont-premultiply-alpha](https://github.com/RedSoxFan/sway/commit/507ba4f8cd3148c744b49ee630d8a0542efa9513)

This PR pre-multiples the alpha channel and multiplies all channels by con->alpha before passing the colors array to wlr_render_rect. This produces the same end result as wlr_render_texture.


RyanDwyer:
> wlr_render_rect has an issue with lighter colors not blending

I'm confused. Shouldn't that mean the fix should be made in wlroots?

emersion:
>Shouldn't that mean the fix should be made in wlroots?

wlroots assumes pre-multiplied alpha for textures and colors.

RedSoxFan:
> > wlr_render_rect has an issue with lighter colors not blending

> I'm confused. Shouldn't that mean the fix should be made in wlroots?

Sorry. Let me clarify, the resulting render (using the formulas in my original post) for wlr_render_rect has an issue with lighter colors not blending. As emersion pointed out, wlroots assumes pre-multiplied alpha for textures and colors, this PR applies the that to the colors passed to wlr_render_rect so we achieve the same result as wlr_render_texture (which already has the premultiplied alpha applied by cairo). The only difference is that we pass con->alpha to the shader for wlr_render_texture, but not wlr_render_rect so I applied it to the colors passed wlr_render_rect.

emersion:
Thanks!

PR #2011, opened 2018-05-21
Fix bottom border when using hide_edge_borders

RyanDwyer:
If the user had a swaybar docked at the bottom, the view's bottom border would always be shown. The check needs to be based on the workspace height rather than the output height.

To test:

* Have `hide_edge_borders both` in config, a swaybar with `position bottom`, then start sway and open a view
* Repeat the above but with the swaybar set to `position top` to ensure I didn't break anything there


RedSoxFan:
Looks good for `hide_edge_borders both` and `position bottom`.

With `hide_edge_borders both` and swaybar with `position top`, the top border still renders with `border pixel` (it should with `border normal`, which ignores view->border_top). If you want, that could be lumped in with this PR. If not, I'll create a follow up PR. It's basically the same concept. Just changing the y check from 0 to ws->y. I guess view->border_left and view->border_right should be changed to check against ws->x and ws->x + ws->width as well. I'm not sure if it is possible to test with a left/right sidebar currently or not.

Also as an aside, sorry for some reason I thought i3 showed borders against bars in my original PR for `hide_edge_borders`. Apparently, I missed a lot when testing that PR.

martinetd:
@RedSoxFan you can create arbitrary "bars" by making an exclusive side-anchored layer-shell with the wlroots `examples/layer-shell` program.

With that said, I'd second the change for the x check agains `ws->x` and `ws->x + ws->width` to be consistent

RedSoxFan:
@martinetd Thanks for the information

Looks good for both top and bottom now

RyanDwyer:
I've updated the left and right calculations as discussed.

PR #2012, opened 2018-05-21
Fix border commands from changing focus

### LINKED ISSUE #1951: border toggle gives focus to first child on workspace
It should leave focus on whatever it was toggling the border of.

RedSoxFan:
Fixes #1951 

This allows the border commands to update the cursor without having the side effect of `focus_follows_mouse` changing the focus. All other calls to `cursor_send_pointer_motion` currently allow refocusing (no change) so `focus_follows_mouse` should continue to work as expected.


RyanDwyer:
Hah, I was about the submit my own PR. It's exactly the same, except I named the variable `allow_change_focus`. I like your name better.

SirCmpwn:
Thanks!

PR #2015, opened 2018-05-21
Implement stacked layout

RyanDwyer:
You know the drill. It's mostly a copy/paste from tabbed.

At some point (probably after floating) I will try to combine some of the rendering parts to minimise the amount of duplicate/similar code.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -221,6 +221,6 @@ void container_notify_child_title_changed(struct sway_container *container);
 /**
  * Return the height of a regular title bar.
  */
-size_t container_titlebar_height();
+size_t container_titlebar_height(void);
 
 #endif
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -11,6 +11,12 @@ extern struct sway_container root_container;
 struct sway_view;
 struct sway_seat;
 
+#define TITLEBAR_BORDER_THICKNESS 1
+
+// Padding includes titlebar border
+#define TITLEBAR_H_PADDING 3
+#define TITLEBAR_V_PADDING 4
+
 /**
  * Different kinds of containers.
  *
@@ -211,5 +217,10 @@ void container_update_title_textures(struct sway_container *container);
 void container_calculate_title_height(struct sway_container *container);
 
 void container_notify_child_title_changed(struct sway_container *container);
+
+/**
+ * Return the height of a regular title bar.
+ */
+size_t container_titlebar_height();
 
 #endif
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -269,17 +269,6 @@ static void render_unmanaged(struct sway_output *output,
 		render_surface_iterator, &data);
 }
 
-static void render_view(struct sway_view *view, struct sway_output *output,
-		pixman_region32_t *damage) {
-	struct render_data data = {
-		.output = output,
-		.damage = damage,
-		.alpha = view->swayc->alpha,
-	};
-	output_view_for_each_surface(
-			view, &data.root_geo, render_surface_iterator, &data);
-}
-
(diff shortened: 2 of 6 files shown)


emersion:
>I will try to combine some of the rendering parts to minimise the amount of duplicate/similar code.

Is it possible to do that now?

sway/desktop/output.c
```diff
@@ -790,12 +790,177 @@ static void render_container_tabbed(struct sway_output *output,
...
+		struct border_colors *colors) {
+	struct sway_view *view = con->sway_view;
+	render_view(view, output, damage);
+
+	struct wlr_box box;
+	float output_scale = output->wlr_output->scale;
+	float color[4];
+
+	if (view->border != B_NONE) {
+		if (view->border_left) {
```
emersion: Maybe we could add functions to render borders that could be reused?

RyanDwyer:
>> I will try to combine some of the rendering parts to minimise the amount of duplicate/similar code.

> Is it possible to do that now?

Done. Removed more lines of code than I added :D

Also, I checked i3's implementation and made it pixel perfect. Marks now show in titlebars for tabbed and stacked containers too.

RedSoxFan:
The only issue that I've noticed so far is that you cannot move a container out of the stack in either vertical direction. It will only move up and down the stack and stop when hitting the bounds. It appears tabs have the same issue in the horizontal direction. 

If you have multiple outputs, moving a container out of the stack/tabs in the direction of the other output does move it out, but to the other output. To match i3, you should be able to move containers out of stacks/tabs and create a split on a single output (then if you move again, it will move to another outputs, if possible).

SirCmpwn:
>The only issue that I've noticed so far is that you cannot move a container out of the stack in either vertical direction. It will only move up and down the stack and stop when hitting the bounds. It appears tabs have the same issue in the horizontal direction.

Since this affects tabs too, you should file an issue and we'll address it separately.

sway/desktop/output.c
```diff
@@ -317,108 +306,159 @@ static void premultiply_alpha(float color[4], float opacity) {
...
+	if (layout == L_HORIZ || layout == L_VERT ||
+			(layout == L_STACKED && is_last_child)) {
+		if (view) {
+			left_offset = view->border_left * view->border_thickness;
+			right_offset = view->border_right * view->border_thickness;
+			connects_sides = true;
+		}
+	}
+	box.x = x + left_offset;
+	box.y = y + config->font_height + 7;
```
emersion: Is it possible to add the 3px padding and the 1px border as constants to know why there are 7, 6, 4 and 3 floating around?

emersion:
Code LGTM, nice work!

Rendering looks good, but pointer input to select the stack titlebar doesn't work (works in tabbed mode).

include/sway/tree/container.h
```diff
@@ -212,4 +218,9 @@ void container_calculate_title_height(struct sway_container *container);
 
 void container_notify_child_title_changed(struct sway_container *container);
 
+/**
+ * Return the height of a regular title bar.
+ */
+size_t container_titlebar_height();
```
ascent12: Very small point, but you should put an explicit `(void)` as the arguments for a prototype.
An empty argument list for prototypes has surprising results in C.

RyanDwyer: Ah yeah, you or someone else has mentioned that to me before. I'll change it.

SirCmpwn:
Great work! Thanks!

include/sway/tree/container.h
ascent12: I may as well give the reason now, in case anyone stumbles on this comment again.
An empty argument list in a prototype is actually declaring a function with _unspecified_ arguments, not no arguments.
So if you tried to call it like `container_titlebar_height(100);`, you wouldn't get a warning.

PR #2016, opened 2018-05-21
Fix hide_edge_borders smart for tabs and stacks

RedSoxFan:
This fixes `hide_edge_borders smart` to handle tabbed and stacked containers.


emersion:
Thanks!

Issue #1988, opened 2018-05-15
sway should send XDG_TOPLEVEL_STATE_TILED state when appropriate


atomnuker:
Currently sway always sends the MAXIMIZED state. To be more correct, it can send XDG_TOPLEVEL_STATE_TILED_TOP/BOTTOM/LEFT/RIGHT state to inform clients as to where they are on the screen. They'll still behave the same.

emersion:
Requires xdg-shell stable 1.14.

PR #2027, opened 2018-05-24
Implement floating

RyanDwyer:
This implements basic floating features. You can float views, unfloat views, make them sticky and unsticky, and click to focus/unfocus them. You cannot move or resize them.

I was going to make it so you can float `C_CONTAINER`s in this PR, but there were some issues and I wanted to get this in front of everyone for review, so I blocked off that functionality. There's still bits in the code where it handles `C_CONTAINER`s being floated though.

This PR has a known issue with floating xdg_shell views when using multiple outputs. If floating the view on the left output, I'm experiencing a flickering copy of the container at the same output-local position on the right output. This does not happen for xwayland views. I can't see what I'm doing wrong, so if someone can assist me with that then that would be appreciated.

I went with the approach of having a `sway_workspace->floating` property, but I made it a `struct sway_container *` rather than a `list_t`. Using a container makes it easier to pass to functions that expect to be given a container. This `floating` container is created at workspace creation, destroyed at workspace destruction, and is immune to reaping. The type of this container is `C_CONTAINER` and the layout is irrelevant and never used.

Each view implementation has a new `wants_floating` function, which allows the implementation to inspect the view and determine if it should be floating or not. There's a basic/terrible implementation for xwayland already done, but there needs to be some discussion on how to do this properly. Perhaps wlroots should expose its own list of window types so we don't have to use the xcb library. At the moment I'm using magic numbers, but this is a temporary solution.

Test plan:

* Use as daily driver
* Float and unfloat things
* Make things sticky and unsticky
* Use multiple outputs

My preference is to implement the move/resize commands and floating `C_CONTAINER`s as separate PRs.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -84,6 +84,7 @@ struct sway_container {
 	// Includes borders
 	double x, y;
 	double width, height;
+	double saved_x, saved_y;
 	double saved_width, saved_height;
 
 	list_t *children;
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -40,6 +40,7 @@ enum sway_container_layout {
 	L_VERT,
 	L_STACKED,
 	L_TABBED,
+	L_FLOATING,
 };
 
 enum sway_container_border {
@@ -74,10 +75,6 @@ struct sway_container {
 	enum sway_container_type type;
 	enum sway_container_layout layout;
 	enum sway_container_layout prev_layout;
-
-	// Allow the container to be automatically removed if it's empty. True by
-	// default, false for the magic floating container that each workspace has.
-	bool reapable;
 
 	// Saves us from searching the list of children/floating in the parent
 	bool is_floating;
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -75,8 +75,13 @@ struct sway_container {
 	enum sway_container_layout layout;
 	enum sway_container_layout prev_layout;
 
+	// Allow the container to be automatically removed if it's empty. True by
+	// default, false for the magic floating container that each workspace has.
+	bool reapable;
+
 	// Saves us from searching the list of children/floating in the parent
 	bool is_floating;
+	bool is_sticky;
 
 	// For C_ROOT, this has no meaning
 	// For C_OUTPUT, this is the output position in layout coordinates
@@ -174,6 +179,13 @@ struct sway_container *container_at(struct sway_container *container,
 		double *sx, double *sy);
 
 /**
+ * Same as container_at, but only checks floating views and expects coordinates
+ * to be layout coordinates, as that's what floating views use.
+ */
+struct sway_container *floating_container_at(double lx, double ly,
+		struct wlr_surface **surface, double *sx, double *sy);
+
+/**
  * Apply the function for each descendant of the container breadth first.
  */
 void container_for_each_descendant_bfs(struct sway_container *container,
@@ -228,5 +240,15 @@ void container_notify_subtree_changed(struct sway_container *container);
  * Return the height of a regular title bar.
  */
 size_t container_titlebar_height(void);
+
+void container_set_floating(struct sway_container *container, bool enable);
+
+void container_set_geometry_from_view(struct sway_container *container);
+
+/**
+ * Determine if the given container is itself floating or has a floating
+ * ancestor.
+ */
+bool container_self_or_parent_floating(struct sway_container *container);
 
 #endif
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -246,10 +246,4 @@ void container_set_geometry_from_floating_view(struct sway_container *con);
  */
 bool container_is_floating(struct sway_container *container);
 
-/**
- * Determine if the given container is itself floating or is a child of a
- * floating container.
- */
-bool container_self_or_parent_floating(struct sway_container *container);
-
 #endif
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -79,8 +79,7 @@ struct sway_container {
 	bool is_sticky;
 
 	// For C_ROOT, this has no meaning
-	// For C_OUTPUT, this is the output position in layout coordinates
-	// For other types, this is the position in output-local coordinates
+	// For other types, this is the position in layout coordinates
 	// Includes borders
 	double x, y;
 	double width, height;
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -76,6 +76,9 @@ struct sway_container {
 	enum sway_container_layout layout;
 	enum sway_container_layout prev_layout;
 
+	// Saves us from searching the list of children/floating in the parent
+	bool is_floating;
+
 	// For C_ROOT, this has no meaning
 	// For C_OUTPUT, this is the output position in layout coordinates
 	// For other types, this is the position in output-local coordinates
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -76,8 +76,6 @@ struct sway_container {
 	enum sway_container_layout layout;
 	enum sway_container_layout prev_layout;
 
-	// Saves us from searching the list of children/floating in the parent
-	bool is_floating;
 	bool is_sticky;
 
 	// For C_ROOT, this has no meaning
@@ -243,8 +241,14 @@ void container_set_floating(struct sway_container *container, bool enable);
 void container_set_geometry_from_view(struct sway_container *container);
 
 /**
- * Determine if the given container is itself floating or has a floating
- * ancestor.
+ * Determine if the given container is itself floating.
+ * This will return false for any descendants of a floating container.
+ */
+bool container_is_floating(struct sway_container *container);
+
+/**
+ * Determine if the given container is itself floating or is a child of a
+ * floating container.
  */
 bool container_self_or_parent_floating(struct sway_container *container);
 
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -40,7 +40,6 @@ enum sway_container_layout {
 	L_VERT,
 	L_STACKED,
 	L_TABBED,
-	L_FLOATING,
 };
 
 enum sway_container_border {
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -238,7 +238,7 @@ size_t container_titlebar_height(void);
 
 void container_set_floating(struct sway_container *container, bool enable);
 
-void container_set_geometry_from_view(struct sway_container *container);
+void container_set_geometry_from_floating_view(struct sway_container *con);
 
 /**
  * Determine if the given container is itself floating.
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -32,7 +32,9 @@ struct sway_view_impl {
 	void (*configure)(struct sway_view *view, double ox, double oy, int width,
 		int height);
 	void (*set_activated)(struct sway_view *view, bool activated);
+	void (*set_maximized)(struct sway_view *view, bool maximized);
 	void (*set_fullscreen)(struct sway_view *view, bool fullscreen);
+	bool (*wants_floating)(struct sway_view *view);
 	void (*for_each_surface)(struct sway_view *view,
 		wlr_surface_iterator_func_t iterator, void *user_data);
 	void (*close)(struct sway_view *view);
@@ -49,6 +51,10 @@ struct sway_view {
 	// Geometry of the view itself (excludes borders)
 	double x, y;
 	int width, height;
+
+	// The size the view would want to be if it weren't tiled.
+	// Used when changing a view from tiled to floating.
+	int natural_width, natural_height;
 
 	bool is_fullscreen;
 
@@ -214,6 +220,8 @@ void view_autoconfigure(struct sway_view *view);
 
 void view_set_activated(struct sway_view *view, bool activated);
 
+void view_set_maximized(struct sway_view *view, bool maximized);
+
 void view_set_fullscreen_raw(struct sway_view *view, bool fullscreen);
 
 void view_set_fullscreen(struct sway_view *view, bool fullscreen);
@@ -236,7 +244,7 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface);
 
 void view_unmap(struct sway_view *view);
 
-void view_update_position(struct sway_view *view, double ox, double oy);
+void view_update_position(struct sway_view *view, double lx, double ly);
 
 void view_update_size(struct sway_view *view, int width, int height);
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -137,6 +137,7 @@ struct sway_xwayland_view {
 	struct wl_listener unmap;
 	struct wl_listener destroy;
 
+	int pending_lx, pending_ly;
 	int pending_width, pending_height;
 };
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -29,7 +29,7 @@ struct sway_view_impl {
 	const char *(*get_string_prop)(struct sway_view *view,
 			enum sway_view_prop prop);
 	uint32_t (*get_int_prop)(struct sway_view *view, enum sway_view_prop prop);
-	void (*configure)(struct sway_view *view, double ox, double oy, int width,
+	void (*configure)(struct sway_view *view, double lx, double ly, int width,
 		int height);
 	void (*set_activated)(struct sway_view *view, bool activated);
 	void (*set_tiled)(struct sway_view *view, bool tiled);
@@ -48,7 +48,7 @@ struct sway_view {
 	struct sway_container *swayc; // NULL for unmapped views
 	struct wlr_surface *surface; // NULL for unmapped views
 
-	// Geometry of the view itself (excludes borders)
+	// Geometry of the view itself (excludes borders) in layout coordinates
 	double x, y;
 	int width, height;
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -32,7 +32,6 @@ struct sway_view_impl {
 	void (*configure)(struct sway_view *view, double lx, double ly, int width,
 		int height);
 	void (*set_activated)(struct sway_view *view, bool activated);
-	void (*set_tiled)(struct sway_view *view, bool tiled);
 	void (*set_fullscreen)(struct sway_view *view, bool fullscreen);
 	bool (*wants_floating)(struct sway_view *view);
 	void (*for_each_surface)(struct sway_view *view,
@@ -220,8 +219,6 @@ void view_configure(struct sway_view *view, double ox, double oy, int width,
 void view_autoconfigure(struct sway_view *view);
 
 void view_set_activated(struct sway_view *view, bool activated);
-
-void view_set_tiled(struct sway_view *view, bool tiled);
 
 void view_set_fullscreen_raw(struct sway_view *view, bool fullscreen);
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -32,7 +32,7 @@ struct sway_view_impl {
 	void (*configure)(struct sway_view *view, double ox, double oy, int width,
 		int height);
 	void (*set_activated)(struct sway_view *view, bool activated);
-	void (*set_maximized)(struct sway_view *view, bool maximized);
+	void (*set_tiled)(struct sway_view *view, bool tiled);
 	void (*set_fullscreen)(struct sway_view *view, bool fullscreen);
 	bool (*wants_floating)(struct sway_view *view);
 	void (*for_each_surface)(struct sway_view *view,
@@ -220,7 +220,7 @@ void view_autoconfigure(struct sway_view *view);
 
 void view_set_activated(struct sway_view *view, bool activated);
 
-void view_set_maximized(struct sway_view *view, bool maximized);
+void view_set_tiled(struct sway_view *view, bool tiled);
 
 void view_set_fullscreen_raw(struct sway_view *view, bool fullscreen);
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -51,6 +51,9 @@ struct sway_view {
 	double x, y;
 	int width, height;
 
+	double saved_x, saved_y;
+	int saved_width, saved_height;
+
 	// The size the view would want to be if it weren't tiled.
 	// Used when changing a view from tiled to floating.
 	int natural_width, natural_height;
--- a/sway/debug-tree.c
+++ b/sway/debug-tree.c
@@ -22,8 +22,6 @@ static const char *layout_to_str(enum sway_container_layout layout) {
 		return "L_STACKED";
 	case L_TABBED:
 		return "L_TABBED";
-	case L_FLOATING:
-		return "L_FLOATING";
 	case L_NONE:
 	default:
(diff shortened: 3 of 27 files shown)


sway/desktop/output.c
```diff
@@ -754,9 +754,87 @@ static void render_container(struct sway_output *output,
...
+		if (focus == con) {
+			colors = &config->border_colors.focused;
+			title_texture = con->title_focused;
+			marks_texture = view->marks_focused;
+		} else {
+			colors = &config->border_colors.unfocused;
+			title_texture = con->title_unfocused;
+			marks_texture = view->marks_unfocused;
+		}
+		render_titlebar(soutput, damage, con, con->x, con->y, con->width,
```
RedSoxFan: There should be a check against the border style here. If you have a floating view with opacity set below 1.0 and `border pixel` or `border none`, the title bar is visible since it is being rendered underneath the view.

include/sway/tree/container.h
```diff
@@ -76,6 +75,14 @@ struct sway_container {
 	enum sway_container_layout layout;
 	enum sway_container_layout prev_layout;
 
+	// Allow the container to be automatically removed if it's empty. True by
+	// default, false for the magic floating container that each workspace has.
+	bool reapable;
```
emersion: In this case, why don't you keep the L_FLOATING for the floating container?

include/sway/tree/container.h
```diff
@@ -76,6 +75,14 @@ struct sway_container {
...
+
+	// Saves us from searching the list of children/floating in the parent
+	bool is_floating;
```
emersion: `is_floating` is trivially implemented with a function

RedSoxFan:
I found a couple more issues:
1. If you `fullscreen enable` a floating view and then `fullscreen disable` it, the position and geometry are not restored and the view remains to cover the entire output.
2. When you focus a sticky view, whatever workspace the view was made sticky on is switched to

include/sway/tree/view.h
```diff
@@ -32,7 +32,9 @@ struct sway_view_impl {
 	void (*configure)(struct sway_view *view, double ox, double oy, int width,
 		int height);
 	void (*set_activated)(struct sway_view *view, bool activated);
+	void (*set_maximized)(struct sway_view *view, bool maximized);
```
emersion: This should be renamed `set_tiled` (see #1988).

sway/debug-tree.c
```diff
@@ -22,8 +22,6 @@ static const char *layout_to_str(enum sway_container_layout layout) {
 		return "L_STACKED";
 	case L_TABBED:
 		return "L_TABBED";
-	case L_FLOATING:
```
emersion: Should be added back?

sway/desktop/output.c
```diff
@@ -755,8 +755,94 @@ static void render_container(struct sway_output *output,
...
+	}
+}
+
+static void render_floating_container(struct sway_output *soutput,
+		pixman_region32_t *damage, struct sway_container *con) {
+	// We need to translate the floating container's coordinates from layout
+	// coordinates into output-local coordinates. This needs to happen for all
+	// children of the floating container too.
+	struct sway_container *output = container_parent(con, C_OUTPUT);
+	container_translate(con, -output->x, -output->y);
```
emersion: Hmm, I'm really not a fan of this. Find another way.

sway/desktop/output.c
```diff
@@ -755,8 +755,94 @@ static void render_container(struct sway_output *output,
 		render_container_tabbed(output, damage, con, parent_focused);
 		break;
 	case L_FLOATING:
-		// TODO
-		break;
+		sway_assert(false, "Didn't expect to see floating here");
+	}
+}
+
+static bool floater_intersects_output(struct sway_container *floater,
```
emersion: This isn't enough, because popups/subsurfaces can be outside of this box and still need to get rendered.

sway/desktop/output.c
```diff
@@ -839,6 +923,8 @@ static void render_output(struct sway_output *output, struct timespec *when,
 			&root_container.sway_root->xwayland_unmanaged);
 		render_layer(output, damage,
 			&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_TOP]);
+
+		render_floating(output, damage);
```
emersion: Floating is under top

sway/desktop/output.c
```diff
@@ -915,6 +1001,7 @@ static void send_frame_done(struct sway_output *output, struct timespec *when) {
 
 	struct sway_container *workspace = output_get_active_workspace(output);
 	send_frame_done_container(&data, workspace);
+	send_frame_done_container(&data, workspace->sway_workspace->floating);
```
emersion: `output_view_for_each_surface` has not been updated for floating containers, so this can't work

sway/desktop/output.c
```diff
@@ -1036,7 +1123,15 @@ static void output_damage_view(struct sway_output *output,
 
 void output_damage_from_view(struct sway_output *output,
 		struct sway_view *view) {
-	output_damage_view(output, view, false);
+	if (container_self_or_parent_floating(view->swayc)) {
+		view->x -= output->swayc->x;
```
emersion: Meh, I'm sure there's a nicer way to do this

sway/desktop/xdg_shell.c
```diff
@@ -117,6 +126,11 @@ static void set_fullscreen(struct sway_view *view, bool fullscreen) {
 	wlr_xdg_toplevel_set_fullscreen(surface, fullscreen);
 }
 
+static bool wants_floating(struct sway_view *view) {
+	// TODO
```
emersion: This can return true if the view isn't resizable, ie. if the min size is equal to the max size.

sway/desktop/xdg_shell.c
```diff
@@ -163,11 +179,17 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 	struct sway_xdg_shell_view *xdg_shell_view =
 		wl_container_of(listener, xdg_shell_view, commit);
 	struct sway_view *view = &xdg_shell_view->view;
-	// NOTE: We intentionally discard the view's desired width here
-	// TODO: Store this for restoration when moving to floating plane
-	// TODO: Let floating views do whatever
-	view_update_size(view, xdg_shell_view->pending_width,
-		xdg_shell_view->pending_height);
+	struct wlr_box *geometry = &view->wlr_xdg_surface->geometry;
```
emersion: The xdg-surface geometry can be unset, in which case the surface's size should be used.

sway/desktop/xdg_shell.c
```diff
@@ -97,6 +97,7 @@ static void configure(struct sway_view *view, double ox, double oy, int width,
 	xdg_shell_view->pending_width = width;
 	xdg_shell_view->pending_height = height;
 	wlr_xdg_toplevel_set_size(view->wlr_xdg_surface, width, height);
+	view_update_position(view, ox, oy);
```
emersion: This is documented to take layout coords

sway/desktop/xdg_shell_v6.c
```diff
@@ -97,6 +97,7 @@ static void configure(struct sway_view *view, double ox, double oy, int width,
 	xdg_shell_v6_view->pending_width = width;
 	xdg_shell_v6_view->pending_height = height;
 	wlr_xdg_toplevel_v6_set_size(view->wlr_xdg_surface_v6, width, height);
+	view_update_position(view, ox, oy);
```
emersion: All my comments on xdg-shell also apply here

sway/desktop/xwayland.c
```diff
@@ -152,33 +152,43 @@ static uint32_t get_int_prop(struct sway_view *view, enum sway_view_prop prop) {
 	}
 }
 
-static void configure(struct sway_view *view, double ox, double oy, int width,
+// The x and y arguments are output-local for tiled views, and layout
+// coordinates for floating views.
```
emersion: Hmmm.

sway/desktop/xwayland.c
```diff
@@ -152,33 +152,43 @@ static uint32_t get_int_prop(struct sway_view *view, enum sway_view_prop prop) {
...
+		if (!sway_assert(output, "view must be within tree to set position")) {
+			return;
+		}
+		struct sway_container *root = container_parent(output, C_ROOT);
+		if (!sway_assert(root, "output must be within tree to set position")) {
+			return;
+		}
+		struct wlr_output_layout *layout = root->sway_root->output_layout;
+		struct wlr_output_layout_output *loutput =
+			wlr_output_layout_get(layout, output->sway_output->wlr_output);
```
emersion: No need to do all of this. Just take the `struct sway_container *output`'s coordinates.

sway/desktop/xwayland.c
```diff
@@ -197,6 +215,35 @@ static void set_fullscreen(struct sway_view *view, bool fullscreen) {
...
+	uint32_t *atom = xsurface->window_type;
+	for (size_t i = 0; i < xsurface->window_type_len; ++i) {
+		wlr_log(L_DEBUG, "xwayland window type %i", *atom);
+		// TODO: Come up with a better way of doing this
+		switch (*atom) {
+		case 36: // NET_WM_WINDOW_TYPE_UTILITY
+		case 44: // NET_WM_WINDOW_TYPE_SPLASH
+		case 276: // ? PGP passphrase dialog
+		case 337: // ? Firefox open file dialog
+		case 338: // ? Firefox open file dialog
```
emersion: Let's implement this instead of having these magic values :)

sway/desktop/xwayland.c
```diff
@@ -234,10 +283,18 @@ static void handle_commit(struct wl_listener *listener, void *data) {
...
-	// TODO: Let floating views do whatever
-	view_update_size(view, xwayland_view->pending_width,
-		xwayland_view->pending_height);
+	struct wlr_xwayland_surface *xsurface = view->wlr_xwayland_surface;
+	if (!view->natural_width && !view->natural_height) {
+		view->natural_width = xsurface->width;
+		view->natural_height = xsurface->height;
+	}
+	if (view->swayc && container_is_floating(view->swayc)) {
+		view_update_size(view, xsurface->width, xsurface->height);
```
emersion: Not sure we want to allow xwayland windows to move. We want to allow that for unmanaged windows, sure, but do we want to do this for toplevels?

sway/tree/container.c
```diff
@@ -866,3 +896,72 @@ void container_notify_child_title_changed(struct sway_container *container) {
...
+			view_set_maximized(container->sway_view, true);
+		}
+		container->is_sticky = false;
+		container_reap_empty_recursive(workspace->sway_workspace->floating);
+	}
+	arrange_workspace(workspace);
+	container_damage_whole(container);
+}
+
+void container_set_geometry_from_view(struct sway_container *container) {
```
emersion: The name should make it more obvious that it expects a floating view - maybe `container_set_geometry_from_floating_view`?

sway/tree/container.c
```diff
@@ -866,3 +896,72 @@ void container_notify_child_title_changed(struct sway_container *container) {
...
+	size_t top =
+		view->border == B_NORMAL ? container_titlebar_height() : border_width;
+
+	container->x = view->x - border_width;
+	container->y = view->y - top;
+	container->width = view->width + border_width * 2;
+	container->height = top + view->height + border_width;
+}
+
+bool container_self_or_parent_floating(struct sway_container *container) {
```
emersion: This name is weird, and doesn't match what the function really does?

sway/tree/view.c
```diff
@@ -511,32 +554,30 @@ void view_unmap(struct sway_view *view) {
 	}
 }
 
-void view_update_position(struct sway_view *view, double ox, double oy) {
-	if (view->swayc->x == ox && view->swayc->y == oy) {
+void view_update_position(struct sway_view *view, double lx, double ly) {
+	if (!container_is_floating(view->swayc)) {
```
emersion: Should still do nothing if the position hasn't changed

sway/desktop/output.c
RyanDwyer: Yeah, I'm not a fan either. The options I can think of for this are:

1) Adjust the container's position temporarily as it currently does.
2) Have file-scoped `translate_x` and `translate_y` variables, which are 0 if the container being rendered is tiled, and the inverse of the output's position if floating. The rendering functions would add the translation to the thing being rendered, so it's basically forced into output-level scope.
3) Same as above, but using a `wlr_matrix` instead of the translate variables.
4) Use layout-local coordinates for all swayc positions, then subtract the output's position when adding damage and rendering anything.

Options 2 and 3 are pretty complicated. To damage or render any view, you have to figure out if the view is itself floating or a child of floating container. If it is, its coordinates are layout-local and need to be translated to output-local coordinates.

I think the best option is option 4 - use layout-local coordinates for all swaycs. This would also resolve the "these coords might be layout-local or output-local" issue that you pointed out in https://github.com/swaywm/sway/pull/2027#pullrequestreview-123501607.

sway/desktop/xdg_shell.c
RyanDwyer: Are you sure about that? `wlr_xdg_shell` contains these:

	bool has_next_geometry;
	struct wlr_box next_geometry;
	struct wlr_box geometry;

...which suggests to me that the geometry struct is always present.

sway/desktop/output.c
RyanDwyer: I went and implemented option 4. That particular commit was +71, -130. It has simplified other areas of the codebase and appears to have fixed my damage problems too.

sway/desktop/output.c
RyanDwyer: I don't quite understand this comment. Is this still relevant after my most recent commits, and if so can you please clarify what you mean?

sway/desktop/xdg_shell.c
RyanDwyer: Can you please clarify this? The `wants_floating` function is called when a view is mapped so that the implementation can decide whether it's floating by default (eg. for dialogs).

sway/desktop/xdg_shell.c
emersion: The client doesn't always set the geometry, so this box can be empty. We should probably document this, it's written in the protocol itself.

sway/tree/container.c
```diff
@@ -596,14 +597,43 @@ struct sway_container *container_at(struct sway_container *parent,
...
+			for (int k = 0; k < ws->floating->children->length; ++k) {
+				struct sway_container *floater =
+					ws->floating->children->items[k];
+				struct wlr_box box = {
+					.x = floater->x,
+					.y = floater->y,
+					.width = floater->width,
+					.height = floater->height,
+				};
+				if (wlr_box_contains_point(&box, lx, ly)) {
```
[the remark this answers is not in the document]
RyanDwyer: I floated Firefox, then right clicked near the right side so the context menu popped up outside of the float. I can interact with the popup outside the floating area just fine. That's because layer surfaces are checked before `floating_container_at` and `container_at` are called.

sway/desktop/output.c
```diff
@@ -450,8 +454,10 @@ static void render_titlebar(struct sway_output *output,
 		wlr_texture_get_size(marks_texture,
 			&texture_box.width, &texture_box.height);
 		texture_box.x =
-			(x + width - TITLEBAR_H_PADDING) * output_scale - texture_box.width;
-		texture_box.y = (y + TITLEBAR_V_PADDING) * output_scale;
+			(x - output->wlr_output->lx + width - TITLEBAR_H_PADDING)
```
emersion: Don't use `output->wlr_output->lx`, use `output->swayc->x` instead

sway/desktop/xdg_shell.c
emersion: Yes. I'm just suggesting a way to implement this for xdg-shell.

sway/tree/container.c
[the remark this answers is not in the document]
emersion: This is completely unrelated to layer surfaces. Firefox is xwayland, which uses unmanaged surfaces that are global to the whole layout and don't have a parent. I'm talking about xdg-popups and wayland subsurfaces.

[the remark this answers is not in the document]
RyanDwyer: I can replicate this with FileZilla. However, I can also replicate it when tiled, and I can replicate in master, and I can replicate it in master from a week ago which was prior to when I split `container_at` into separate functions. So this problem has been around for a while.

I guess the solution will be to iterate the containers, check their subsurfaces first, then iterate them all again and check the toplevel surface. The same has to happen for rendering too - currently a tiled view can render over the top of its left sibling's context menu.

My preference is to make a separate PR for this because it's not specific to floating.

sway/desktop/xwayland.c
RyanDwyer: This is resolved in commit "Don't let xwayland views set position unless unmanaged".

sway/tree/container.c
RyanDwyer: With the change of swayc coordinates from output-local to layout-local, this function was no longer needed so I've deleted it.

sway/desktop/xwayland.c
RyanDwyer: wlroots needs to expose its atom names first. And I'm a bit confused with how this all works - wlroots lacks atom names `DIALOG` and `TOOLBAR` which are used by i3. And according to `wlr_xwayland_surface_is_unmanaged`, surfaces with atoms `SPLASH` and `UTILITY` will be unmanaged, but to match i3 these would need to be managed and floated.

BTW, according to Ongy the atom values are defined at runtime, so I've removed the hard coded values from my PR. Keeping them there may cause false positives.

SirCmpwn:
Having issues with move left|right and my output layout on this branch. My cursor leads me to believe that the output layout is still correct but the move commands focus the wrong outputs, I can get stuck on my transformed (270) output. Also, `workspace 1` brings me to workspace 2 (which is on the transformed output).

```
output DP-2 pos 0,0 res 1920x1080
output DP-3 pos 1920,0 scale 2
output DP-1 pos 3840,-700 transform 270 bg `wallpaper 1:1 9:16` fill
output HDMI-A-1 pos 4920,0
```

sway/tree/container.c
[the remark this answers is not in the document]
emersion: Filezilla is *not* using xdg-shell. Pick any GTK+3 app, right click, and you should get a popup.

sway/desktop/xwayland.c
emersion: I can do the wlroots PR tonight if that helps

RyanDwyer:
The gedit issue is happening because we send an unmaximize to the client. This makes gedit draw the outer glow as part of its surface. It also uses rounded corners which looks terrible when we wrap the surface in our own borders.

Is it acceptable to not send the unmaximize event? The other option would be respecting CSD and accounting for the glow offset everywhere.

SirCmpwn:
>Is it acceptable to not send the unmaximize event? The other option would be respecting CSD and accounting for the glow offset everywhere.

For now let's just always keep it maximized. We'll later give the user some control over this when we bring in the SSD protocol.

RyanDwyer:
@SirCmpwn I haven't been able to replicate your output problem, but I think I fixed it in "Fix getting adjacent output". Can you please test again?

The only outstanding work to be done in this PR is the xwayland auto floating which depends on wlroots changing things (https://github.com/swaywm/wlroots/issues/1010). I'm happy to do that later if that's easier.

I'm happy to squash all my commits into one prior to merging if you prefer a cleaner git history.

sway/tree/container.c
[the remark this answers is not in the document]
emersion: Bump

SirCmpwn:
>I'm happy to squash all my commits into one prior to merging if you prefer a cleaner git history.

No, please don't.

SirCmpwn:
Can you prepare a summary of necessary follow-up work? e.g. move/resize

SirCmpwn:
>I haven't been able to replicate your output problem, but I think I fixed it in "Fix getting adjacent output". Can you please test again?

I'm still encountering this. If I focus the leftmost output and `focus right`, I end up on the 3rd output (which is the one transformed 270). `workspace 1` still brings me to the 3rd output.

SirCmpwn:
The output issue was introduced by 38deed7fe3d8a26161c3d3697d90b8a04cdaa540/

RyanDwyer:
> Can you prepare a summary of necessary follow-up work? e.g. move/resize

I've created a floating status issue here: #2063

RyanDwyer:
I can't replicate either of your problems, and I can't determine what would be causing those behaviours by looking at the code. Can you please investigate those?

For focus jumping to the third output, I would start by checking what values we're passing to `wlr_output_layout_adjacent_output` (from `container_get_in_direction`). Keep in mind the swayc x and y is now layout-local - not that it should matter because your focused output's position is 0,0 anyway.

RyanDwyer:
I've fixed a bug with the move command, but I admit I was unable to replicate that myself.

Also, wlroot's `wlr_box_contains_point` is calculated incorrectly which could be causing you issues too.

emersion:
Ah, reminder that you need to send the enter and leave events whenever a surface enters or leaves an output for HiDPI support.

SirCmpwn:
15dd31d fixes the issue I was having with focusing the wrong outputs on focus left/right.

SirCmpwn:
Seems to work as advertised. Will run this as my daily driver for a bit.

SirCmpwn:
Experiencing weirdness with mpv (--gpu-context=wayland) on my hidpi display. Similar to what I saw with mpv on x11 before 11393af.

Also having damage issues after unfullscreening a floating mpv window.

RyanDwyer:
I fixed the damage issue when unfullscreening into floating. I was setting the swayc geometry before calling `view_configure`, then there was a race condition where we might render another full screen surface before mpv sends a smaller surface, and at that point we couldn't damage properly because we didn't have the original swayc geometry any more.

My solution of storing the view's saved geometry as well as the swayc's saved geometry isn't very elegant. I'm looking forward to doing atomic because the view configuring and damaging code will get overhauled.

I assume your weirdness with `mpv --gpu-context=wayland` is the fact that it's not scaling. We let the client choose their own dimensions for floating (within reason), and mpv is returning the same dimensions as the video. This makes sense I guess; mpv would be upscaling the video if it were to scale it. Can you confirm that's what you're experiencing?

atomnuker:
Setting each window as maximized to fix a very minor issue is way more evil to everyone than just asking people to ignore the few pixels until support for a protocol is implemented. Clients can't resize themselves if they're maximized.

SirCmpwn:
Getting the same mpv issues as before now even on lodpi displays. This is definitely not stable.

include/sway/tree/container.h
```diff
@@ -76,12 +76,14 @@ struct sway_container {
 	enum sway_container_layout layout;
 	enum sway_container_layout prev_layout;
 
+	bool is_sticky;
+
 	// For C_ROOT, this has no meaning
-	// For C_OUTPUT, this is the output position in layout coordinates
-	// For other types, this is the position in output-local coordinates
+	// For other types, this is the position in layout coordinates
```
emersion: In fact, even for C_ROOT, this is the position in layout coordinates (for C_ROOT, it's the extents of the whole output layout).

sway/tree/container.c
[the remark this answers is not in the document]
emersion: Hmm, can you just remove this `wlr_box_contains_point` check and only use `container_at`?

AmateurMemetics:
Tried using this, top borders not rendered on windows which start as floating (gnome-terminal preferences window and gnome-keyring when opening Geary).

SirCmpwn:
I don't want to hold this up forever so I'm going to file a separate ticket for the border issue.

PR #2030, opened 2018-05-24
Don't focus tabbed and stacked containers on mouseover

### LINKED ISSUE #2024: Tab layout: Switching tabs on mouseover
Tab layout is switching on mouseover (instead of just on click/keybind).

chebykinn:
Fixes #2024.


sway/input/cursor.c
```diff
@@ -162,7 +162,11 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
 				seat_set_focus_warp(cursor->seat, c, false);
 			}
 		} else {
-			seat_set_focus_warp(cursor->seat, c, false);
+			// Don't switch focus on mouseover for stacked and tabbed layouts
+			if(c->parent && c->parent->layout != L_STACKED
+						 && c->parent->layout != L_TABBED) {
```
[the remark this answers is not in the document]
chebykinn: Ok

chebykinn:
Oh, yeah, I get it, it breaks multiple outputs, if I do it like this.

SirCmpwn:
Also breaks this on a single output:

```
(workspace L_HORIZONTAL
  (view focus)
  (container L_TABBED
    (view focus-inactive)
    (view)))
```

SirCmpwn:
The correct solution is probably to check if we're about to focus a child of a tabbed container to see if it's the focus-inactive child and if not, focus the focus-inactive child instead.

chebykinn:
Makes sense, I'll try that.

chebykinn:
Is this correct?

chebykinn:
Should I rebase?

SirCmpwn:
Yes

sway/input/cursor.c
```diff
@@ -162,7 +162,29 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
...
+			struct sway_container *p = c->parent;
+			struct sway_container *first_tabbed_parent = c->parent;
+			while(p) {
+				if(p->layout == L_TABBED || p->layout == L_STACKED) {
+					do_mouse_focus = false;
+					first_tabbed_parent = p;
+				}
+				p = p->parent;
+			}
+			if(!do_mouse_focus) {
```
SirCmpwn: style issue, put a space between `if` and `(`, same for `while`

chebykinn:
Done

sway/input/cursor.c
```diff
@@ -162,7 +162,29 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
...
-			seat_set_focus_warp(cursor->seat, c, false);
+			// Don't switch focus on title mouseover for
+			// stacked and tabbed layouts
+			// If pointed container is in nested containers which are
+			// inside tabbed/stacked layout we should skip them
+			bool do_mouse_focus = true;
+			struct sway_container *p = c->parent;
+			while (p) {
+				if ((p->layout == L_TABBED || p->layout == L_STACKED)
+					&& !view_is_visible(c->sway_view)) {
```
RyanDwyer: Move the call to `view_is_visible` outside of the while loop and store it in a variable rather than recomputing it on each iteration of the loop.

Also, assuming that `c` is a view makes me a little nervous. Please add an assertion for this.

sway/input/cursor.c
```diff
@@ -162,7 +162,29 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
...
+					do_mouse_focus = false;
+					break;
+				}
+				p = p->parent;
+			}
+			if (!do_mouse_focus) {
+				struct sway_container *next_focus = seat_get_focus_inactive(
+						cursor->seat, p);
+				if (next_focus && view_is_visible(next_focus->sway_view)) {
```
RyanDwyer: Is this `view_is_visible` check needed? If so, assert that `next_focus` is a view. If not, remove it.

sway/input/cursor.c
```diff
@@ -179,6 +181,8 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
 			if (!do_mouse_focus) {
 				struct sway_container *next_focus = seat_get_focus_inactive(
 						cursor->seat, p);
+				sway_assert(next_focus->type == C_VIEW,
+							"focus inactive container is not a view");
```
RyanDwyer: `next_focus` may be `NULL` which would make this crash.

Also, when sway encounters a failing `sway_assert` in a **development** build, it'll crash with the message given. When sway encounters a failing `sway_assert` in a **release** build, the `sway_assert` will return false and it'll continue on executing code. Which in your case would lead to a crash. This is why other places in the code use a pattern of:

    if (!sway_assert(condition, "message")) {
        return;
    }

RyanDwyer:
Thanks!

PR #2033, opened 2018-05-25
Arrange views after flattening the parent

### LINKED ISSUE #2029: Tab layout: wrong windows position after some rearrangements
It seems there are new issues after fix #2025 to reproduce it do the folowing steps:
1 open 2 tabs
2 mod + v
3 mod + shift + h on the second tab
after these operations the position of "window" will be in the wrong position

RyanDwyer:
Fixes #2029.

To test:

* Created a tabbed layout with some views
* Focus a view and run `splitv` or `splith` (tree should be something like `T[view, view, V[view]]`)
* Run `move left` (tree should be `T[view, view, view]`)

Before the PR, the view would be offset from the title bar until the tabbed container is next arranged. After the PR, it shows correctly as a regular tab.

### DIFF
--- a/sway/tree/layout.c
+++ b/sway/tree/layout.c
@@ -327,9 +327,11 @@ void container_move(struct sway_container *container,
 		current = container_parent(container, C_OUTPUT);
 	}
 
-	if (parent != container_flatten(parent)) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #2034, opened 2018-05-25
Clean up container title functions

RyanDwyer:
This is mostly a code cleanup PR. The only functional change is fixing a minor bug (5th bullet point).

* Add and use `lenient_strcat` and `lenient_strncat` functions
* Rename `concatenate_child_titles` function as that's no longer what it does
* Rename `container_notify_child_title_changed` because we only need to notify that the tree structure has changed, not titles
* Don't notify parents when a child changes its title
* Update ancestor titles when changing a container's layout
    * Eg. create nested tabs and change the inner container to stacking
* No need to store tree presentation in both `container->name` and `container->formatted_title`

### DIFF
--- a/common/stringop.c
+++ b/common/stringop.c
@@ -55,6 +55,20 @@ void strip_quotes(char *str) {
 	*end = '\0';
 }
 
+char *lenient_strcat(char *dest, const char *src) {
+	if (dest && src) {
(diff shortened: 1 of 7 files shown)


SirCmpwn:
Thanks!

PR #2035, opened 2018-05-25
Fix focus bug when moving between outputs

RyanDwyer:
When moving focus left or right to an adjacent output, only select the first or last child in the new workspace if the workspace's layout is horizontalish. If it's a verticalish layout, use the last focused container.

To test:

* Have two outputs.
* On the left output, create two or more views in a vertical layout. Ensure the views are direct children of the workspace (eg. by creating two views then moving one of them down).
* Focus the topmost view.
* `focus right` to move focus to the right output (it doesn't matter if the right output is empty or has views).
* `focus left` to move back to the left output.
* Prior to the fix, sway would focus the bottom child in the left output. Because the layout is perpendicular to the movement direction, It should focus the last active child instead.

### DIFF
--- a/sway/tree/layout.c
+++ b/sway/tree/layout.c
@@ -586,11 +586,19 @@ static struct sway_container *get_swayc_in_output_direction(
 	if (ws->children->length > 0) {
 		switch (dir) {
 		case MOVE_LEFT:
-			// get most right child of new output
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #2036, opened 2018-05-25
Switch workspace if required when mapping a view

### LINKED ISSUE #1879: View doesn't render when it creates a new workspace that is immediately visible
* Have two or more outputs
* In config, `workspace media output $right_output`
* Configure your workspaces/views in such a way that the media workspace does not exist
* With a container on the left output focused, run `swaymsg move container to workspace media`

The container will be moved to the media workspace and the media workspace will have focus inactive, but the media workspace fails to show the container, instead showing the background wallpaper. You can make the container appear by focusing the media workspace, switching to another workspace one on the same output then back to media. Most likely a damage issue.

RyanDwyer:
Fixes #1879.

Have something like this in your config:

    assign [class="Firefox"] web
    workspace web output X11-2

Then launch sway with two outputs, focus output X11-1, and launch Firefox. The web workspace will be created on X11-2, Firefox will be created there and focused, the IPC workspace event will be sent, but Firefox wouldn't actually render because you hadn't switched to the workspace.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -438,10 +438,11 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface) {
 	// Check if there's any `assign` criteria for the view
 	list_t *criterias = criteria_for_view(view,
 			CT_ASSIGN_WORKSPACE | CT_ASSIGN_OUTPUT);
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #2040, opened 2018-05-25
[WIP] Implement swap command

RedSoxFan:
Implements `swap container with id|con_id|mark <arg>`.
`id` is only valid for xwayland views as it represents the X11 window ID

For reference, here is what i3's user guide says for the `swap` command: https://i3wm.org/docs/userguide.html#_swapping_containers

For the most part, this should be working. I'm just trying to fix a few things.

Should be working:
- [X] Selecting a container/view by id
- [X] Selecting a container/view by con_id
- [X] Selecting a container/view by mark
- [X] Prevent swapping an anscestor and descendant
- [X] Prevent swapping with an output or workspace
- [X] Swapping position and geometry
- [X] Swapping fullscreen state for views (containers cannot be fullscreen currently)

TODO/Known Issues:
- [ ] Prevent the workspace from switching when swapping with a container/view on a hidden workspace
- [ ] Fix tab/stack focusing when swapping the focused container/view out


SirCmpwn:
>I'd prefer the workspace not to switch at all on swap.

Is the behavior of your current patch different from i3's behavior? If so that still needs to be addressed. I guess I can take a crack at debugging it if you're lost.

RedSoxFan:
Ya. This patch generates either two or four `workspace::focus` events (2, if at least one container visible. 4, if both are hidden). i3 does not generate any `workspace::focus` events during a swap.

This should produce the same end result, there is just some noise in the IPC events. I'm not sure if it is worth worrying over or not. In either case, I have tried to modify the code to make it so the workspace does not change when inserting a child, but failed. I would not be surprised if it is simple and I just don't have a good enough handle on the codebase yet.

RedSoxFan:
Actually, it appears that `move container to workspace <name>` generates `workspace::focus` events as well. If it is decided that intermediate IPC events are an issue, then that is something for another PR since it is out of the scope of this one.

SirCmpwn:
Alright. We'll have to do a big IPC cleanup later anyway.

PR #2046, opened 2018-05-26
Implement Gaps

natesymer:
This pull request contains an implementation of gaps, something @emersion suggested I do. It's been a fun couple of hours hacking on my favorite window manager!

Included are two things:
1. An implementation of the commands to implement gaps.
2. An implementation of gaps in sway/tree/arrange.c

### Commands

These commands were crafted to match their descriptions in `$ man 5 sway`. They implement both gaps and smart gaps. They support all permutations of the commands.

### arrange.c implementation

When arranging `sway_container`s, add gaps. It's pretty simple, except when gaps have been added to specific workspaces or containers.

Container-specific gaps are implemented via the `has_gaps`, `gaps_inner`, and `gaps_outer` fields of the `sway_container` struct. These fields are passed down from parent container to child container, unless the child container already has its own gaps. When commands apply values to these fields, it triggers a re-arrangement.


natesymer:
The builds.sr.ht builds failed because build/ is in the gitignore, and thus my changes to ninja.build are null and void.

PR #2047, opened 2018-05-26
Implement Gaps

natesymer:
This pull request contains an implementation of gaps, something @emersion suggested I do. It's been a fun couple of hours hacking on my favorite window manager!

Included are two things:

An implementation of the commands to implement gaps.
An implementation of gaps in sway/tree/arrange.c
Commands
These commands were crafted to match their descriptions in $ man 5 sway. They implement both gaps and smart gaps. They support all permutations of the commands.

arrange.c implementation
When arranging sway_containers, add gaps. It's pretty simple, except when gaps have been added to specific workspaces or containers.

Container-specific gaps are implemented via the has_gaps, gaps_inner, and gaps_outer fields of the sway_container struct. These fields are passed down from parent container to child container, unless the child container already has its own gaps. When commands apply values to these fields, it triggers a re-arrangement.

(this is a redo of a previous pull request that failed building)


SirCmpwn:
>These commands were crafted to match their descriptions in $ man 5 sway. They implement both gaps and smart gaps. They support all permutations of the commands.

Can you double check that what's described in sway(5) matches the behavior of i3-gaps

SirCmpwn:
Code LGTM, will test it later.

natesymer:
I think I didn't do it perfectly, I'll look at it tomorrow morning. I looked at sway 0.1.5 and it seems inner gaps are applied on both sides of a view - I only did it on one side of a view. Let me know if you find any  bugs.

RyanDwyer:
Hey, this is neat. It's nice to have gaps with titlebars working properly.

* `swaymsg gaps lol` completes successfully but should display expected syntax
* `swaymsg gaps edge_gaps` with no more arguments makes sway crash
* I'm experiencing some damage issues if I do the following:
    * Launch without any gaps commands in config
    * Open a terminal and run `swaymsg gaps outer 10`

natesymer:
So I just fixed the command syntax. I'm tackling the damage issues as we speak.

sway/commands/gaps.c
```diff
@@ -0,0 +1,153 @@
+#include <string.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/tree/arrange.h"
+#include "log.h"
+#include "stringop.h"
+
+int read_double(char *str, double *out) {
```
ascent12: This prototype should read as `static int read_double(const char *str, double *out)`

ascent12: I'd also prefer that `bool` is used instead of `int`. It would actually be possible to just use `static double read_double(const char *str)` and use NaN as an invalid return, but it's not something that's important.

sway/commands/smart_gaps.c
```diff
@@ -0,0 +1,27 @@
...
+#include "sway/config.h"
+#include "sway/tree/arrange.h"
+#include "sway/tree/view.h"
+#include "sway/tree/container.h"
+#include "log.h"
+#include "stringop.h"
+
+struct cmd_results *cmd_smart_gaps(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "smart_gaps", EXPECTED_AT_LEAST, 1))) {
```
ascent12: Same as my other point.

sway/commands/smart_gaps.c
```diff
@@ -0,0 +1,27 @@
...
+
+	if (strcmp(argv[0], "on") == 0) {
+		config->smart_gaps = true;
+		arrange_root();
+	} else if (strcmp(argv[0], "off") == 0) {
+		config->smart_gaps = false;
+		arrange_root();
+	} else {
+		return cmd_results_new(CMD_INVALID, "smart_gaps",
+			   "Expected 'smart_gaps <on|off>' ");
```
ascent12: Indentation on this line is wrong. The style is a simple two tab indent.

sway/commands/gaps.c
```diff
@@ -0,0 +1,153 @@
...
+		*out = 0.0;
+		return 1;
+	}
+	*out = v;
+	return 0;
+}
+
+struct cmd_results *cmd_gaps(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "gaps", EXPECTED_AT_LEAST, 1))) {
```
ascent12: I know this code is basically copied from somewhere else, but I really don't see the point of forcing this assignment to happen inside the conditional.
```c
struct cmd_results *error = checkarg(...);
if (error) {
    ...
}
```
works fine and I would argue looks a lot cleaner.

sway/commands/gaps.c
```diff
@@ -0,0 +1,153 @@
...
+		} else {
+			if (amount_idx < 1 && argc == 1) {
+				return cmd_results_new(CMD_INVALID, "gaps", "gaps <amount>");
+			} else if (amount_idx < 2 && argc == 2) { // This is the gaps <inner|outer>
+				return cmd_results_new(CMD_INVALID, "gaps", "gaps inner|outer <amount>");
+			} else {
+				return cmd_results_new(CMD_INVALID, "gaps", "gaps inner|outer all|workspace|current set|plus|minus <amount>"); 
+			}
+		}
+	}
```
ascent12: This function overall has some pretty deep indentation, going up to 6 levels. It should be possible to make it a lot shallower by refactoring it and adding some early return statements.

natesymer: I didn't use early returns because I'm personally opposed to them (they're less math-y). I can refactor it.

emersion: +1 to early returns

SirCmpwn: +1 to early returns

ascent12: I would agree with you more if we were talking about a functional language, or something expression-oriented like Rust. But in a purely imperative language like C, exiting as soon as possible is generally a lot cleaner and can make the logic of the code easier to understand.

natesymer: That makes sense, I've got a haskell/clojure background!

sway/commands/gaps.c
```diff
@@ -34,34 +33,52 @@ struct cmd_results *cmd_gaps(int argc, char **argv) {
...
+			config->edge_gaps = !(config->edge_gaps);
+			arrange_root();
 		} else {
 			return cmd_results_new(CMD_INVALID, "gaps", "gaps edge_gaps on|off|toggle");
 		}
 	} else {
-		int amount_idx = 0;
-		char op = '=';
-		char scope = 'a';
+		int amount_idx = 0; // the current index in argv. Should be the index of the <amount> by the read_double call.
```
SirCmpwn: 80 column limit, but this comment is hardly necessary anyway.

sway/commands/gaps.c
```diff
@@ -34,34 +33,52 @@ struct cmd_results *cmd_gaps(int argc, char **argv) {
...
+		char op = '='; // '=', '+', '-'
+		char scope = 'a'; // 'a', 'w', 'c'
```
SirCmpwn: Maybe this should be an enum.

SirCmpwn:
Rather than littering this with comments all over the place, you should document the requirements, constraints, and approach in a single comment at the top and let the code do the rest of the talking.

sway/commands/gaps.c
```diff
@@ -6,6 +6,18 @@
 #include "stringop.h"
 #include <math.h>
 
+enum GapsOp {
```
SirCmpwn: style: `enum gaps_operation` and make the members things like `GAPS_OP_SET`

natesymer:
Before I tackle the style issue with the enum, I realized I didn't implement inner gaps quite right. I wrote a new implementation. I can't figure out why with my implementation toggling the split mode erroneously re-applies gaps.

I saw in sway/tree/layout.c that a new container is built to wrap any split container. Any advice?

natesymer:
Never mind. Took care of it myself. I'd say it's ready. @SirCmpwn, take a look at it.

sway/commands/gaps.c
```diff
@@ -0,0 +1,186 @@
...
+};
+
+enum gaps_scope {
+	GAPS_SCOPE_ALL,
+	GAPS_SCOPE_WORKSPACE,
+	GAPS_SCOPE_CURRENT
+};
+
+double read_double(char *str) {
+	if (strcmp(str, "0") == 0 || strcmp(str, "0.0") == 0) {
```
ascent12: You could just read errno for ERANGE further on instead of explicitly checking for 0.

sway/commands/gaps.c
```diff
@@ -0,0 +1,186 @@
...
+	GAPS_OP_SUBTRACT
+};
+
+enum gaps_scope {
+	GAPS_SCOPE_ALL,
+	GAPS_SCOPE_WORKSPACE,
+	GAPS_SCOPE_CURRENT
+};
+
+double read_double(char *str) {
```
emersion: `str` can be `const`

sway/commands/gaps.c
```diff
@@ -0,0 +1,186 @@
...
+			return cmd_results_new(CMD_INVALID, "gaps",
+				"gaps edge_gaps on|off|toggle");
+		}
+	} else {
+		int amount_idx = 0; // the current index in argv
+		enum gaps_op op = GAPS_OP_SET;
+		enum gaps_scope scope = GAPS_SCOPE_ALL;
+		bool inner = true;
+		
+
```
emersion: Style: no need for two blank lines - one is enough

sway/commands/gaps.c
```diff
@@ -0,0 +1,186 @@
...
+			arrange_root();
+		} else if (strcmp(argv[1], "off") == 0) {
+			config->edge_gaps = false;
+			arrange_root();
+		} else if (strcmp(argv[1], "toggle") == 0) {
+			if (!config->active) {
+				return cmd_results_new(CMD_INVALID, "gaps",
+					"Cannot toggle gaps while not running.");
+			}
+			config->edge_gaps = !(config->edge_gaps);
```
emersion: Style: no need for these parentheses

natesymer:
@SirCmpwn so when will this get merged?

SirCmpwn:
Sit tight, this is still in my review list.

sway/tree/arrange.c
```diff
@@ -71,16 +71,92 @@ void arrange_workspace(struct sway_container *workspace) {
...
+
+		c->gapped = false;
+		c->current_gaps = 0;
+		wlr_log(L_DEBUG, "Removing gaps %p", c);
+	} else {
+		wlr_log(L_DEBUG, "Removing gaps: not gapped: %p", c);
+	}
+}
+
+static void add_gaps(struct sway_container *c) {
```
emersion: The code for `add_gaps` is the same as `remove_gaps`, with s/+/-/. Is it possible to unify these in a `set_gaps` function?

sway/commands/gaps.c
```diff
@@ -0,0 +1,182 @@
...
+			config->gaps_inner = val;
+			config->gaps_outer = val;
+			arrange_root();
+			return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+		}
+		// Other variants. The middle-length variant (gaps inner|outer <amount>)
+		// just defaults the scope to "all" and defaults the op to "set".
+ 
+		double total = val;
+		if (op == GAPS_OP_SUBTRACT) {
```
ascent12: I think this would be more appropriate as a switch too, also for better warnings. The GAPS_OP_SET case would just need to be `total = val`.

sway/commands/gaps.c
```diff
@@ -0,0 +1,182 @@
...
+		}
+
+		if (scope == GAPS_SCOPE_ALL) { // All: set the config values for gaps
+			if (inner) {
+				config->gaps_inner = total;
+			} else {
+				config->gaps_outer = total;
+			}
+			arrange_root();
+		} else if (scope == GAPS_SCOPE_WORKSPACE) { // Workspace: set the current workspace's gaps
```
ascent12: These comments go over the 80 column limit (It's not a hard limit, but this is significantly over).
They should go on their own line, or even on the enum definition itself.

sway/tree/arrange.c
natesymer: I was actually going to change the way gaps were set. Right now, too much of how gaps are removed depends on where the sway_container is. With things like flattening, some ugly bugs are going to pop up. I was going to change it so that a sway_container tracks gaps on each dimension (gaps_width, gaps_y, etc). This will make the remove_gaps() function smaller and more DRY.

natesymer:
Say I launch sway and run two processes that display UI (`swaymsg exec urxvt && swaymsg exec urxvt`). Then, I focus one urxvt container. I then run `swaymsg gaps inner 200`. Then, in that container, I run `swaymsg split toggle && swaymsg exec urxvt`. Should the most recently created container have the same gaps as its sibling?

ascent12:
>Should the most recently created container have the same gaps as its sibling?

It's probably best to just see what the i3-gaps behaviour is.

natesymer:
I trie i3-gaps, and it had different gaps semantics: `all` is the same, but there's no `workspace` - instead they use `current`. Should Sway match i3-gaps, or should it continue doing what it's currently doing?

I personally wouldn't mind the hyper-precise and specific behavior I'm talking about; `currrent` should only apply to the current sway_container - not its children, siblings, etc.

natesymer:
@SirCmpwn Do things look good?

emersion:
There are still damage issues when changing gaps. The usual workflow is: damage, change something, damage - that damages both the old state and the new state, ensuring everything gets re-drawn correctly.

emersion:
BTW, if you're stuck when trying to fix a bug, don't hesitate to come on IRC and ask for help!

SirCmpwn:
>Should Sway match i3-gaps, or should it continue doing what it's currently doing?

Should match i3-gaps.

Also there are merge conflicts

Still looking good though, nice work and thanks for your patience in tying it all together.

natesymer:
@SirCmpwn Even Sway 0.15 doesn't match i3-gaps.

sway/tree/container.c
```diff
@@ -729,7 +734,7 @@ void container_damage_whole(struct sway_container *container) {
 	for (int i = 0; i < root_container.children->length; ++i) {
 		struct sway_container *cont = root_container.children->items[i];
 		if (cont->type == C_OUTPUT) {
-			output_damage_whole_container(cont->sway_output, container);
+			output_damage_whole_container(cont->sway_output, cont);
```
emersion: This should really be `container`

emersion:
I think this happens: when gaps grow, containers are correctly damaged, but then views are too big. We ask clients to reconfigure themselves with a smaller size, but we don't damage correctly borders when they update their size.

emersion:
I've been trying to debug this issue, and it's now pretty clear to me it's an existing bug in `output_damage_container`. Since this PR doesn't introduce a new bug, I'd be okay to merge it.

@natesymer: I'm sorry for the delay and for wasting your time. Can you rebase this PR, squash all these commits into one and fix the merge conflict?

natesymer:
@emersion Don't worry about it! Nothing about working on sway is a waste of my time. I squashed the merge conflict and rebased.

natesymer:
Just a question, how would I squash these commits together?

vilhalmer:
If you `git rebase -i master` you will get an editor with a file where you can change the first word of every line to select the action to take on the commit. If you change all but the first one to `squash` (or just `s`), you will end up with one commit, and it will walk you through fixing any conflicts on the way. At the end you should be good to merge! (Since you're rewriting history by doing so, it'll require a force push to update your branch after doing this.)

natesymer:
@emersion All prepared! Would you merge this?

SirCmpwn:
Thanks!

PR #2048, opened 2018-05-26
Add %app_id to title_format

RyanDwyer:
If we have `%class` and `%instance` we should probably have `%app_id` too. Makes it easier to write criteria.


emersion:
Thanks!

PR #2050, opened 2018-05-27
Focus containers only on entry.

### LINKED ISSUE #1946: Mouse randomly refocuses hovered window even after changing focus via keyboard
Steps to reproduce:
1. Open two windows
2. Hover mouse over first window.
3. Change focus to next window.
4. After some delay focus will be returned to first window.

Expected behavior: focus stays as it was set until mouse is moved to another window.

sway version: 7d707e6865cb123018157582f673711f20cda2f2

smlx:
Change the behaviour of `focus_follows_mouse` to only focus the container on entry rather than on _any_ movement.
This matches the [behaviour of i3](https://i3wm.org/docs/userguide.html#_focus_follows_mouse).

Closes #1946.

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -6,7 +6,9 @@
 struct sway_cursor {
 	struct sway_seat *seat;
 	struct wlr_cursor *cursor;
-	struct cursor_position *previous;
+	struct {
+		double x, y;
+	} previous;
 	struct wlr_xcursor_manager *xcursor_manager;
 
 	struct wl_client *image_client;
@@ -26,10 +28,6 @@ struct sway_cursor {
 	uint32_t tool_buttons;
 
 	struct wl_listener request_set_cursor;
-};
-
-struct cursor_position {
-	double x, y;
 };
 
 void sway_cursor_destroy(struct sway_cursor *cursor);
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -6,6 +6,7 @@
 struct sway_cursor {
 	struct sway_seat *seat;
 	struct wlr_cursor *cursor;
+	struct cursor_position *previous;
 	struct wlr_xcursor_manager *xcursor_manager;
 
 	struct wl_client *image_client;
@@ -27,10 +28,14 @@ struct sway_cursor {
 	struct wl_listener request_set_cursor;
 };
 
+struct cursor_position {
+	double x, y;
+};
+
 void sway_cursor_destroy(struct sway_cursor *cursor);
 struct sway_cursor *sway_cursor_create(struct sway_seat *seat);
-void cursor_send_pointer_motion(struct sway_cursor *cursor,
-	double delta_x, double delta_y, uint32_t time_msec, bool allow_refocusing);
+void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
+	bool allow_refocusing);
 void dispatch_cursor_button(struct sway_cursor *cursor, uint32_t time_msec,
 	uint32_t button, enum wlr_button_state state);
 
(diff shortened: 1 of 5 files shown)


include/sway/input/cursor.h
```diff
@@ -29,8 +29,8 @@ struct sway_cursor {
 
 void sway_cursor_destroy(struct sway_cursor *cursor);
 struct sway_cursor *sway_cursor_create(struct sway_seat *seat);
-void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
-		bool allow_refocusing);
+void cursor_send_pointer_motion(struct sway_cursor *cursor,
```
emersion: Hmmm. I'd rather keep the last position of the mouse in a field in `sway_cursor`.

smlx:
@emersion I moved the last position of the mouse into `sway_cursor` as suggested. There were also some subtle bugs in the focus switching logic, but I think those are fixed now too.

include/sway/input/cursor.h
```diff
@@ -27,10 +28,14 @@ struct sway_cursor {
 	struct wl_listener request_set_cursor;
 };
 
+struct cursor_position {
+	double x, y;
+};
```
SirCmpwn: You should just inline this

sway/input/cursor.c
```diff
@@ -162,33 +170,23 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
...
+			// Focus c if both of the following are true:
+			// - cursor is over a new view, i.e. entered a new window; and
+			// - the new view is visible, i.e. not hidden in a stack or tab.
+			if (c != prev_c && view_is_visible(c->sway_view)) {
 				seat_set_focus_warp(cursor->seat, c, false);
+			} else {
+				struct sway_container *next_focus =
+					seat_get_focus_inactive(cursor->seat, &root_container);
+				if (next_focus) {
+					if (!sway_assert(next_focus->type == C_VIEW,
```
emersion: This can't be an assert, `seat_get_focus_inactive` can return a workspace

sway/input/cursor.c
```diff
@@ -144,6 +144,14 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
 	struct wlr_seat *seat = cursor->seat->wlr_seat;
 	struct wlr_surface *surface = NULL;
 	double sx, sy;
+
+	// Find the container the pointer was previously over
+	struct sway_container *prev_c = container_at_coords(cursor->seat,
```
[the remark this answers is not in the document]
smlx: I'm not sure it matters in this case.

Consider:
cursor is over container A
layout changes, cursor is now over B
move cursor

All I'm checking here is if the previous pointer position and the current pointer position are over the same container at this point in time. The specific container it is doesn't actually matter, does it? The behaviour seems to be correct at any rate.

include/sway/input/cursor.h
smlx: Moved into the struct, as suggested.

sway/input/cursor.c
smlx: Okay, I changed this to just check the type.

sway/input/cursor.c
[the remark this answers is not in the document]
emersion: >move cursor

Should focus B right? But this won't happen because of `c != prev_c`.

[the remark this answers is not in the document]
smlx: For me in that scenario in i3, focus doesn't change to B, it remains on A. Are you able to check it to confirm?

Focus basically doesn't change until a window border is crossed.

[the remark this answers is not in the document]
emersion: Oh, you're right.

SirCmpwn:
Thanks!

chebykinn:
Thanks!

PR #2052, opened 2018-05-27
Don't move empty workspaces when output is destroyed

emersion:


emersion:
>The only issue I have (which also affects master) is that the workspaces are not moved when disconnecting an output, but when reconnecting it

Nah, it's just that the bar isn't updated. If you disconnect, then switch to another workspace, the moved workspace appears on the bar.

RedSoxFan:
> Nah, it's just that the bar isn't updated. If you disconnect, then switch to another workspace, the moved workspace appears on the bar.

Oh so a missing IPC event. That should be trivial to resolve then. I'll see what the correct IPC event is and submit a PR after this one is merged.

emersion:
Thanks for your review. By the way, feel free to hop in on IRC (#sway-devel on Freenode).

PR #2053, opened 2018-05-27
xdg-shell: make all toplevels tiled

### LINKED ISSUE #1988: sway should send XDG_TOPLEVEL_STATE_TILED state when appropriate

emersion:
Depends on https://github.com/swaywm/wlroots/pull/1007
Fixes #1988

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -3,13 +3,14 @@
 #include <stdlib.h>
 #include <wayland-server.h>
 #include <wlr/types/wlr_xdg_shell.h>
+#include <wlr/util/edges.h>
+#include "log.h"
+#include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
+#include "sway/server.h"
 #include "sway/tree/container.h"
 #include "sway/tree/layout.h"
-#include "sway/server.h"
 #include "sway/tree/view.h"
-#include "sway/input/seat.h"
-#include "sway/input/input-manager.h"
-#include "log.h"
 
 static const struct sway_view_child_impl popup_impl;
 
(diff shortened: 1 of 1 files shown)


sway/desktop/xdg_shell.c
```diff
@@ -248,7 +249,8 @@ void handle_xdg_shell_surface(struct wl_listener *listener, void *data) {
 	wlr_log(L_DEBUG, "New xdg_shell toplevel title='%s' app_id='%s'",
 		xdg_surface->toplevel->title, xdg_surface->toplevel->app_id);
 	wlr_xdg_surface_ping(xdg_surface);
-	wlr_xdg_toplevel_set_maximized(xdg_surface, true);
+	wlr_xdg_toplevel_set_tiled(xdg_surface, WLR_EDGE_LEFT | WLR_EDGE_RIGHT |
+		WLR_EDGE_TOP | WLR_EDGE_BOTTOM);
```
SirCmpwn: We should probably do both

emersion: Nah, see the wlroots PR. It falls back to `maximized` if the client doesn't support tiling.

SirCmpwn: I don't understand how that's supposed to work.

emersion: If the client xdg-shell version is >= 2, then it supports tiling. If the client version is 1, then it doesn't support tiling and we make the window maximized instead. That's the best we can do.

SirCmpwn: I see. I think we should be prepared for clients who set their xdg shell version to the latest but who don't bother implementing tiled to be pretty common.

emersion: I don't think so. GNOME has tiling too (if you move a window near a border of the screen), so apps that don't implement it will look bad on GNOME too. Also, an upside to tiling is that the app has no request to unset it.

SirCmpwn: Aight.

SirCmpwn:
Thanks!

PR #2055, opened 2018-05-27
Send workspace::move in container_output_destroy

RedSoxFan:
This sends the `workspace::move` IPC event in `container_output_destroy` when moving a workspace from one output to another. This notifies swaybar (and anything else subscribed to the IPC events) of the change.


SirCmpwn:
Thanks!

PR #2056, opened 2018-05-27
Implement bindsym/bindcode --locked

### LINKED ISSUE #592: Implement bindsym --locked
add an option so media control keys like XF86AudioRaiseVolume etc. can control music/volume in locked state

frsfnrrg:
Adds the --locked flag to bindsym and bindcode commands.

When a keyboard's associated seat has an exclusive client
(i.e, a screenlocker), then bindings are only executed if
they have the locked flag. When there is no such client,
this restriction is lifted.

-------

This should resolve #592. As my development system only has one screen, I haven't 
tested e.g. how this will interact with seat or keyboard additions while a screenlocker
is active, or with multiple screens. Example usage:

    bindsym --locked XF86AudioMute exec amixer sset Master toggle
    bindsym XF86AudioMicMute exec  amixer sset Capture toggle

### DIFF
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -83,20 +83,26 @@ struct cmd_results *cmd_bindsym(int argc, char **argv) {
 	binding->keys = create_list();
 	binding->modifiers = 0;
 	binding->release = false;
+	binding->locked = false;
 	binding->bindcode = false;
 
-	// Handle --release
-	if (strcmp("--release", argv[0]) == 0) {
-		if (argc >= 3) {
+	// Handle --release and --locked
+	while (argc > 0) {
+		if (strcmp("--release", argv[0]) == 0) {
 			binding->release = true;
-			argv++;
-			argc--;
+		} else if (strcmp("--locked", argv[0]) == 0) {
+			binding->locked = true;
 		} else {
-			free_sway_binding(binding);
-			return cmd_results_new(CMD_FAILURE, "bindsym",
-				"Invalid bindsym command "
-				"(expected more than 2 arguments, got %d)", argc);
+			break;
 		}
+		argv++;
+		argc--;
+	}
+	if (argc < 2) {
+		free_sway_binding(binding);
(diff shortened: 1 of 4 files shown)


sway/commands/bind.c
```diff
@@ -83,21 +83,33 @@ struct cmd_results *cmd_bindsym(int argc, char **argv) {
...
+		bool flag = false;
+		if (strcmp("--release", argv[0]) == 0) {
 			binding->release = true;
+			flag = true;
+		}
+		if (strcmp("--locked", argv[0]) == 0) {
+			binding->locked = true;
+			flag = true;
+		}
+		if (flag) {
```
emersion: Can be re-written without this `flag` with something more straightforward:

```c
if (it's --release) {
  …
} else if (it's --locked) {
  …
} else {
  break;
}
argv++;
argc--;
```

sway/commands/bind.c
```diff
@@ -83,21 +83,33 @@ struct cmd_results *cmd_bindsym(int argc, char **argv) {
...
 			argc--;
 		} else {
-			free_sway_binding(binding);
-			return cmd_results_new(CMD_FAILURE, "bindsym",
-				"Invalid bindsym command "
-				"(expected more than 2 arguments, got %d)", argc);
+			break;
 		}
 	}
+	if (argc < 3) {
```
emersion: Should be `<=`? `argc` doesn't include the command name.

sway/input/keyboard.c
```diff
@@ -183,6 +183,9 @@ static bool binding_matches_keycodes(struct wlr_keyboard *keyboard,
 		return false;
 	}
 
+	if (locked > binding->locked)
```
emersion: Style: missing brackets

frsfnrrg:
@RedSoxFan, try again with latest force pushed version; I incorrectly translated the length failure check in my first attempt.

emersion:
Thanks!

PR #2057, opened 2018-05-28
swaymsg: Fix exit code

RyanDwyer:
When executing commands via IPC, sway returns the result in a JSON array. swaymsg expected the result to be in the root of the JSON object, wouldn't find the `success` boolean, and would always return the fallback which is `true`.

This changes swaymsg to look in the array rather than the root, and return false if any of the commands failed.

Note that sway only ever returns an array of one item (the last executed command), but that's out of scope of this PR.

To test, run `swaymsg lol; echo $?`

### DIFF
--- a/swaymsg/main.c
+++ b/swaymsg/main.c
@@ -19,13 +19,26 @@ void sway_terminate(int exit_code) {
 	exit(exit_code);
 }
 
+// Iterate results array and return false if any of them failed
 static bool success(json_object *r, bool fallback) {
-	json_object *success;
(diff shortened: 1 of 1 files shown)


swaymsg/main.c
```diff
@@ -19,13 +19,26 @@ void sway_terminate(int exit_code) {
 	exit(exit_code);
 }
 
+// Iterate results array and return false if any of them failed
 static bool success(json_object *r, bool fallback) {
 	json_object *success;
```
ascent12: This could be moved inside the while loop now.

emersion:
Thanks!

PR #2058, opened 2018-05-28
Refactor parse_title_format

RyanDwyer:
Don't read properties until needed, and do it with less lines of code.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -641,6 +641,14 @@ void view_child_destroy(struct sway_view_child *child) {
 	}
 }
 
+static size_t append_prop(char *buffer, const char *value) {
+	if (!value) {
+		return 0;
+	}
+	lenient_strcat(buffer, value);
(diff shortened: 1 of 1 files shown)


sway/tree/view.c
```diff
@@ -641,6 +641,14 @@ void view_child_destroy(struct sway_view_child *child) {
 	}
 }
 
+static size_t append_prop(char *buffer, const char *value) {
```
RedSoxFan: Would it be worth moving this to `common/stringop.c` and renaming it to something like `lenient_strcat_size`? I would guess that this is not the only place that could make use of it.

RyanDwyer: My preference is to move it only if there's anything else that can make use of it. So I'd leave it where it is until that happens.

RedSoxFan: Ok. That works

emersion:
Thanks!

PR #2060, opened 2018-05-28
Implement focus_wrapping

RedSoxFan:
Implements `focus_wrapping yes|no|force`

i3 user guide: https://i3wm.org/docs/userguide.html#_focus_wrapping

Do we want to support the legacy `force_focus_wrapping yes|no` syntax and map those to `WRAP_FORCE` and `WRAP_YES`? I'm not sure what the current thoughts are on legacy commands.


RyanDwyer:
* Have `focus_wrapping force` in your config
* Create the layout `H[view, V[view, view focused]]`
* `move left`
* Observe focus has to changed to the left view
* Shouldn't it not change in this case? The i3 docs suggest I would have to run `focus parent` then `focus left`.

I'm all for dropping legacy syntax by the way.

RedSoxFan:
> - Have `focus_wrapping force` in your config
> - Create the layout `H[view, V[view, view focused]]`
> - `move left`
> - Observe focus has to changed to the left view
> - Shouldn't it not change in this case? The i3 docs suggest I would have to run `focus parent` then `focus left`.

I just checked i3 and it does focus the left view. If you have `H[view, V[view, H[view focused, view]]]`, it does not. It appears that force only has an effect when there is another view in the direction (or it's inverse) in that container. So you can focus left/right for `L_VERT`/`L_STACKED` and focus up/down for `L_HORIZ`/`L_TABBED` even with `focus_wrapping force`.

> I'm all for dropping legacy syntax by the way.

Sounds good to me

emersion:
Thanks!

SirCmpwn:
>Do we want to support the legacy force_focus_wrapping yes|no syntax and map those to WRAP_FORCE and WRAP_YES?

Yes

PR #2062, opened 2018-05-28
Fix popups

RyanDwyer:
This adds proper support for popups overhanging the bounds of their swayc.

An example of the problem this solves:

* Place a gedit and a terminal side by side with gedit on the left.
* Right click near the right edge of gedit, so the popup overlaps the terminal.
* Before this PR, the terminal would be rendered on top of the popup and you wouldn't be able to hover or click the popup's overhanging part.

The same problem is evident with floating views showing a popup outside the bounds of the floating swayc.

For locating a surface at a given coordinate, the new logic is to iterate all the floating popups first, then floating toplevels and decorations, then tiled popups, then tiled toplevels and decorations.

For rendering popups, it renders all toplevels and decorations first, then popups afterwards.

Some discussion points:
* If you have a tabbed container, open a popup that overhangs the tabbed container, then switch to a different tab, I've made the popup disappear until you tab back. I assume this is desired?
* We might find performance gains if we store a list of mapped popups somewhere global such as `sway_server`, rather than iterating through the tree multiple times when using `container_at` and when rendering.
* When a popup is unmapped, only the swayc's bounds are damaged. I'd like to make this a separate PR because it's not a result of my doing, and I'd rather work on atomic for now.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -174,13 +174,6 @@ struct sway_container *container_at(struct sway_container *workspace,
 		double *sx, double *sy);
 
 /**
- * Same as container_at, but only checks for popups only.
- */
-struct sway_container *popup_at(struct sway_container *workspace,
-		double lx, double ly, struct wlr_surface **surface,
-		double *sx, double *sy);
-
-/**
  * Apply the function for each descendant of the container breadth first.
  */
 void container_for_each_descendant_bfs(struct sway_container *container,
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -169,16 +169,16 @@ struct sway_container *container_parent(struct sway_container *container,
  * surface-local coordinates of the given layout coordinates if the container
  * is a view and the view contains a surface at those coordinates.
  */
-struct sway_container *container_at(struct sway_container *container,
-		double ox, double oy, struct wlr_surface **surface,
+struct sway_container *container_at(struct sway_container *workspace,
+		double lx, double ly, struct wlr_surface **surface,
 		double *sx, double *sy);
 
 /**
- * Same as container_at, but only checks floating views and expects coordinates
- * to be layout coordinates, as that's what floating views use.
+ * Same as container_at, but only checks for popups only.
  */
-struct sway_container *floating_container_at(double lx, double ly,
-		struct wlr_surface **surface, double *sx, double *sy);
+struct sway_container *popup_at(struct sway_container *workspace,
+		double lx, double ly, struct wlr_surface **surface,
+		double *sx, double *sy);
 
 /**
  * Apply the function for each descendant of the container breadth first.
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -69,8 +69,7 @@ struct render_data {
 	struct root_geometry root_geo;
 	struct sway_output *output;
 	pixman_region32_t *damage;
-	struct sway_view *view;
-	bool only_popups;
+	bool popups;
 	float alpha;
 };
 
@@ -219,10 +218,24 @@ damage_finish:
 	pixman_region32_fini(&damage);
 }
 
-static void render_surface(struct wlr_surface *surface, int sx, int sy,
+static bool surface_is_popup(struct wlr_surface *surface) {
+	if (wlr_surface_is_xdg_surface(surface)) {
+		struct wlr_xdg_surface *xdg_surface =
+			wlr_xdg_surface_from_wlr_surface(surface);
+		return xdg_surface->role == WLR_XDG_SURFACE_ROLE_POPUP;
+	}
+	if (wlr_surface_is_xdg_surface_v6(surface)) {
+		struct wlr_xdg_surface_v6 *xdg_surface_v6 =
+			wlr_xdg_surface_v6_from_wlr_surface(surface);
+		return xdg_surface_v6->role == WLR_XDG_SURFACE_V6_ROLE_POPUP;
+	}
+	return false;
+}
+
+static void render_surface_iterator(struct wlr_surface *surface, int sx, int sy,
 		void *_data) {
 	struct render_data *data = _data;
-	if (data->only_popups && surface == data->view->surface) {
+	if (data->popups != surface_is_popup(surface)) {
 		return;
 	}
 	struct wlr_output *wlr_output = data->output->wlr_output;
@@ -261,7 +274,7 @@ static void render_layer(struct sway_output *output,
 		.alpha = 1.0f,
 	};
 	layer_for_each_surface(layer_surfaces, &data.root_geo,
-		render_surface, &data);
+		render_surface_iterator, &data);
 }
 
 static void render_unmanaged(struct sway_output *output,
@@ -272,7 +285,7 @@ static void render_unmanaged(struct sway_output *output,
 		.alpha = 1.0f,
 	};
 	unmanaged_for_each_surface(unmanaged, output, &data.root_geo,
-		render_surface, &data);
+		render_surface_iterator, &data);
 }
 
 static void render_rect(struct wlr_output *wlr_output,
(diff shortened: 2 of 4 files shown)


SirCmpwn:
This is only half of the solution. In the future we'll also need to implement xdg-positioner in sway to prevent popups from being opened outside of outputs.

RyanDwyer:
I've rebased this and updated the PR description. It's ready for review now.

RedSoxFan:
With `workspace_layout stacking` or `workspace_layout tabbed` in the config, sway crashes on launch
```
AddressSanitizer:DEADLYSIGNAL
==7481==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000020 (pc 0x563bc913d8e0 bp 0x7ffc7e7d6db0 sp 0x7ffc7e7d6d50 T0)
==7481==The signal is caused by a READ memory access.
==7481==Hint: address points to the zero page.
    #0 0x563bc913d8df in popup_at_container ../sway/tree/container.c:670
    #1 0x563bc913d9db in popup_at_container ../sway/tree/container.c:677
    #2 0x563bc913dc7c in popup_at ../sway/tree/container.c:707
    #3 0x563bc910c7e0 in container_at_coords ../sway/input/cursor.c:111
    #4 0x563bc910cdfa in cursor_send_pointer_motion ../sway/input/cursor.c:152
    #5 0x563bc910db4d in handle_cursor_motion_absolute ../sway/input/cursor.c:230
    #6 0x7f01de95310f in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #7 0x7f01de9233f9 in handle_pointer_motion_absolute ../subprojects/wlroots/types/wlr_cursor.c:402
    #8 0x7f01de95310f in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #9 0x7f01de8e9a8d in x11_handle_pointer_position ../subprojects/wlroots/backend/x11/input_device.c:41
    #10 0x7f01de8eac65 in update_x11_pointer_position ../subprojects/wlroots/backend/x11/input_device.c:176
    #11 0x7f01de8ec416 in handle_x11_configure_notify ../subprojects/wlroots/backend/x11/output.c:186
    #12 0x7f01de8e7eb5 in handle_x11_event ../subprojects/wlroots/backend/x11/backend.c:56
    #13 0x7f01de8e8040 in x11_event ../subprojects/wlroots/backend/x11/backend.c:84
    #14 0x7f01dca62a41 in wl_event_loop_dispatch (/usr/lib/libwayland-server.so.0+0x9a41)
    #15 0x7f01dca6114b in wl_display_run (/usr/lib/libwayland-server.so.0+0x814b)
    #16 0x563bc90dc9d4 in server_run ../sway/server.c:126
    #17 0x563bc90dbaf5 in main ../sway/main.c:428
    #18 0x7f01dc27406a in __libc_start_main (/usr/lib/libc.so.6+0x2306a)
    #19 0x563bc90d9cb9 in _start (/home/brian/programming/forks/sway/build/sway/sway+0x27cb9)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV ../sway/tree/container.c:670 in popup_at_container
```

RedSoxFan:
I'm also not getting any popups for xwayland clients.

RyanDwyer:
Both issues are now fixed.

emersion:
Can you rebase this  branch?

include/sway/tree/container.h
```diff
@@ -169,16 +169,16 @@ struct sway_container *container_parent(struct sway_container *container,
...
 		double *sx, double *sy);
 
 /**
- * Same as container_at, but only checks floating views and expects coordinates
- * to be layout coordinates, as that's what floating views use.
+ * Same as container_at, but only checks for popups only.
  */
-struct sway_container *floating_container_at(double lx, double ly,
-		struct wlr_surface **surface, double *sx, double *sy);
+struct sway_container *popup_at(struct sway_container *workspace,
```
emersion: What about subsurfaces?

emersion:
One other possible approach would be to render the focused container last.

RyanDwyer:
I like that suggestion. We could also take the same approach for `{container,popup}_at`: Check the focused container first, then everything else. I'll adjust it.

RyanDwyer:
This is ready for review again. You may want to consider comparing it with master rather than just looking at the latest commit, because the latest commit reverts/replaces a fair amount of stuff in the earlier commits.

The latest commit changes `container_at` so that it checks subsurfaces only for the focused view, then steps through the floating tree then the tiling tree. Also, the `floating_container_at` function is now included as part of `container_at`.

The commit also changes the rendering. We run through the tree like normal but skip rendering any subsurfaces. Then after we've gone through the tree, we render the subsurfaces only for the focused view.

With this change, it means if a popup is open and you use a `focus` command to focus another container, the popup will disappear. If you then focus the original container again then the popup will reappear. I'm guessing the desired behaviour is to kill the popup when focus is lost from the view. I'll save that as an exercise for another day though.

To test, play around with popup menus overlapping other views, in both tiling and floating contexts. A good example is having one floating view and one tiled view behind it, with the tiled view focused and a popup overlapping the floating container.

emersion:
>We run through the tree like normal but skip rendering any subsurfaces.

This won't work. We need to render subsurfaces even if the view is not focused, because some clients (e.g. alacritty, weston-subsurfaces) use subsurfaces to render things that are not popups (actual content, decorations).

The most simple approach would be to render all views normally, but to force the focused one to be rendered last.

RyanDwyer:
We can't take that approach, because the focused view might be a tiling view which appears behind a floating view.

RyanDwyer:
Okay, give this one a ago. Similar to before, but it renders subsurfaces for all views, then popups for the focused view at the end. I can't fault it, having tested with weston-subsurfaces.

emersion:
>Similar to before, but it renders subsurfaces for all views, then popups for the focused view at the end.

I don't see why popups and subsurfaces should be handled differently.

emersion:
>popups only need to be rendered for the focused view

Why?

RyanDwyer:
Because, from what I understand, if you move focus away from a view which has a popup then the popup should be closed. Correct me if I'm wrong.

RyanDwyer:
Also, your comment here suggests that popups only exist on the focused container:

> one other possible approach would be to render the focused container last

...because we're trying to make popups appear on top of all other containers.

emersion:
>if you move focus away from a view which has a popup then the popup should be closed

But that's unrelated to rendering… And just creates one special case for popups, instead of doing the same thing for all view children.

RyanDwyer:
We can't render popups along with everything else because another view might get drawn on top of it. That's the problem this PR is solving. Popups must be drawn last. But subsurfaces must be drawn along with each views's surface because they might get covered by something else, such as a floating view.

emersion:
What if i have a non-focused floating view on top of a tiled focused view with a popup?

RyanDwyer:
Then it works correctly. The popup from the tiled view is displayed on top of the floating view. I even gave that as an example in an earlier comment.

> A good example is having one floating view and one tiled view behind it, with the tiled view focused and a popup overlapping the floating container.

emersion:
Yeah, I don't think that's correct. I think the popup should be rendered behind the floating view.

StringEpsilon:
In current i3

StringEpsilon:
In i3 4.15, the pop-up from a tiled focused window will be rendered in front of a unfocused floating window. So RyanDwyers implementation is correct in terms of i3-compatibility.

(I tested this with the context menu from firefox, gedit and kcachegrind and a floating terminal window)

sway/desktop/output.c
```diff
@@ -217,9 +218,26 @@ static void render_texture(struct wlr_output *wlr_output,
...
+			wlr_xdg_surface_v6_from_wlr_surface(surface);
+		return xdg_surface_v6->role == WLR_XDG_SURFACE_V6_ROLE_POPUP;
+	}
+	return false;
+}
+
 static void render_surface_iterator(struct wlr_surface *surface, int sx, int sy,
 		void *_data) {
 	struct render_data *data = _data;
+	if (data->popups != surface_is_popup(surface)) {
```
emersion: This will render subsurfaces of popups when `data->popups == false`.

RyanDwyer: Do you know of any program that does this which I can use for testing?

emersion: No, sorry

SirCmpwn:
What's the story here?

RyanDwyer:
Due to the age of this PR, I'll probably make a v2. I've been prioritising other features in the meantime.

kuon:
Is it supposed to be fixed in 1.2? I use sway 1.2 and while the contextual menu renders properly, I can only click items that are also inside the window boundaries. I don't want to open an issue if this is being tracked already.

emersion:
https://github.com/swaywm/sway/issues/4924

tmccombs:
What's the current status on this? Is someone currently working on this? If not, what would I need to do to get this in a mergeable state?

emersion:
This PR is only about rendering iirc, not about pointer events.

tmccombs:
Ah, ok. I guess I'll work on a separate PR then, and maybe pull these changes into it.

PR #2064, opened 2018-05-28
Fix sway_keyboard keysyms state

### LINKED ISSUE #2020: Switching VTs works randomly
With a fresh session, it works well. For some reason, after some time, it doesn't work anymore.

smlx:
This fixes a bug on the following keyboard input:
1. press shift
2. press another key (such as ;)
3. release shift
4. release other key

If the modifer key is not released last (as in the example above), then `keyboard->pressed_keysyms_translated` never has the translated keysym removed from it as it never sees an associated `WLR_KEY_RELEASED`. Instead it sees a `WLR_KEY_RELEASED` for the untranslated keysym.

~~My fix is simple: if the modifier key is released, clear `keyboard->pressed_keysyms_translated` because there's no easy way to translate any keysyms to their unmodified counterparts. I've done some light testing and it doesn't seem to break anything obvious but please let me know if there's a better approach.~~

Approach has changed - see below.

Closes #2020.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -66,14 +66,25 @@ static void pressed_keysyms_remove(xkb_keysym_t *pressed_keysyms,
 
 static void pressed_keysyms_update(xkb_keysym_t *pressed_keysyms,
 		const xkb_keysym_t *keysyms, size_t keysyms_len,
-		enum wlr_key_state state) {
+		enum wlr_key_state state, bool mod_release_invalidate) {
 	for (size_t i = 0; i < keysyms_len; ++i) {
-		if (keysym_is_modifier(keysyms[i])) {
-			continue;
-		}
 		if (state == WLR_KEY_PRESSED) {
+			if (keysym_is_modifier(keysyms[i])) {
+				continue;
+			}
 			pressed_keysyms_add(pressed_keysyms, keysyms[i]);
 		} else { // WLR_KEY_RELEASED
+			if (keysym_is_modifier(keysyms[i])) {
+				if (mod_release_invalidate) {
+					// clear the translated keysyms since they are no longer valid
+					for (size_t j = 0; j < SWAY_KEYBOARD_PRESSED_KEYSYMS_CAP; ++j) {
+						pressed_keysyms[j] = XKB_KEY_NoSymbol;
+						return;
+					}
+				} else {
+					continue;
+				}
+			}
 			pressed_keysyms_remove(pressed_keysyms, keysyms[i]);
 		}
 	}
@@ -357,7 +368,7 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
 		keyboard_keysyms_translated(keyboard, keycode, &translated_keysyms,
 			&keyboard->modifiers_translated);
 	pressed_keysyms_update(keyboard->pressed_keysyms_translated,
-		translated_keysyms, translated_keysyms_len, event->state);
+		translated_keysyms, translated_keysyms_len, event->state, true);
 	if (!handled && event->state == WLR_KEY_PRESSED) {
 		handled = keyboard_execute_bindsym(keyboard,
 			keyboard->pressed_keysyms_translated,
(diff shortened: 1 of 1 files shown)


emersion:
Maybe we should re-translate all keycodes into keysyms every time a key is pressed.

smlx:
Do you mean by storing the keycodes in the `sway_keyboard` struct along with the keysyms?

emersion:
The keycodes should already be stored in `wlr_keyboard`

smlx:
@emersion Thanks for the suggestion, this is a much cleaner approach.

Again, I've tried to test everything I could think of but please try this to see if it breaks anything..

sway/input/keyboard.c
```diff
@@ -335,66 +316,65 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
...
 	pressed_keysyms_update(keyboard->pressed_keysyms_translated,
-		translated_keysyms, translated_keysyms_len, event->state);
-	if (!handled && event->state == WLR_KEY_PRESSED) {
+			translated_keysyms, translated_keysyms_len);
+
+	// update raw keysyms
+	const xkb_keysym_t *raw_keysyms;
+	size_t raw_keysyms_len =
+		keyboard_keysyms_raw(keyboard, keycode, &raw_keysyms,
+			&keyboard->modifiers_raw);
```
emersion: You only generate keysyms from the last pressed keycode. You should use `wlr_keyboard.keycodes` instead of just `keycode`.

smlx:
I see what you mean about calculating keysyms for all keycodes, and I've pushed a commit to work that way.

However I'm curious about why e.g. my Mod+Shift+2 binding was working even when I was just calculating the keysyms using the single keycode?

emersion:
>However I'm curious about why e.g. my Mod+Shift+2 binding was working even when I was just calculating the keysyms using the single keycode? What binding did I break by using the single keycode?

That's because modifiers are treated differently - they are stored in a bitfield. This is due to the fact that multiple keysyms can trigger the same modifier (e.g. you have `Control_L` and `Control_R` which both trigger `Ctrl`).

In order to test this, you need to bind a keysym with more than one non-modifier key (e.g. `a+b+c`).

sway/input/keyboard.c
```diff
@@ -317,16 +298,39 @@ static size_t keyboard_keysyms_translated(struct sway_keyboard *keyboard,
...
 		keyboard->seat_device->input_device->wlr_device;
 	*modifiers = wlr_keyboard_get_modifiers(device->keyboard);
 
-	xkb_layout_index_t layout_index = xkb_state_key_get_layout(
-		device->keyboard->xkb_state, keycode);
-	return xkb_keymap_key_get_syms_by_level(device->keyboard->keymap,
-		keycode, layout_index, 0, keysyms);
+	size_t num_keysyms = 0, num_keysyms_for_keycode;
+	const xkb_keysym_t *keysyms_for_keycode;
+	xkb_keycode_t keycode;
```
emersion: Style: these can move inside the loop. We generally only define outside the loop if we need the variables after the loop.

sway/input/keyboard.c
```diff
@@ -317,16 +298,39 @@ static size_t keyboard_keysyms_translated(struct sway_keyboard *keyboard,
  * This will trigger keybinds such as Alt+Shift+2.
  */
 static size_t keyboard_keysyms_raw(struct sway_keyboard *keyboard,
-		xkb_keycode_t keycode, const xkb_keysym_t **keysyms,
+		xkb_keysym_t *keysyms,
```
emersion: Style: you can merge this line and the next one.

sway/input/keyboard.c
```diff
@@ -317,16 +298,39 @@ static size_t keyboard_keysyms_translated(struct sway_keyboard *keyboard,
...
+				xkb_state_key_get_layout(device->keyboard->xkb_state, keycode);
+		num_keysyms_for_keycode =
+				xkb_keymap_key_get_syms_by_level(device->keyboard->keymap,
+						keycode,
+						layout_index,
+						0,
+						&keysyms_for_keycode);
+		for (size_t j = 0, k; j < num_keysyms_for_keycode; j++) {
+			// search the existing keysyms for the new keysym
+			for (k = 0; k < num_keysyms; k++) {
```
emersion: I don't see `k` defined anywhere - am I missing something?

sway/input/keyboard.c
```diff
@@ -335,66 +339,64 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
...
-			keyboard->pressed_keysyms_translated,
-			keyboard->modifiers_translated,
-			event->state, input_inhibited);
-	}
+	// update translated keysyms
 	const xkb_keysym_t *translated_keysyms;
-	size_t translated_keysyms_len =
-		keyboard_keysyms_translated(keyboard, keycode, &translated_keysyms,
+	size_t translated_keysyms_len = keyboard_keysyms_translated(keyboard,
+			event->keycode + 8,
```
emersion: You need to translate keysyms for all pressed keycodes here too.

emersion:
This has been fixed in a different way in #2080.

Thanks for sending a PR anyway!

smlx:
Ah, great! I ran out of time for this.

PR #2065, opened 2018-05-28
Fix #2018 - Moving containers out of tabs/stacks

### LINKED ISSUE #2018: Containers cannot be moved from tabs/stacks
For a single output, containers cannot be split out of tabs in either horizontal direction and stack in either vertical direction. It will only move from the left/top the tabs/stack to the right/bottom and stop when hitting either bounds.

For multiple outputs, moving a container out of the tabs/stacks in the direction of the other output does move it out, but to the other output. To match i3, you should be able to move containers out of tabs/stacks and create a split on a single output (then if you move again, it will move to another output, if possible).

Tabbed containers have already been merged (#2005) and stacked containers are in the works (#2015)

RedSoxFan:
Fixes #2018 

Moving a container out of a tabbed or stacked parent container creates a new split with the container and it's former parent as siblings

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -160,12 +160,8 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 	json_object_object_add(object, "type", json_object_new_string("con"));
 
 	if (c->parent) {
-		enum sway_container_layout layout =
-			(c->parent->type == C_CONTAINER && c->type == C_VIEW) ?
-			c->parent->layout : c->layout;
-
 		json_object_object_add(object, "layout",
-			json_object_new_string(ipc_json_layout_description(layout)));
+			json_object_new_string(ipc_json_layout_description(c->layout)));
 	}
 }
 
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -160,7 +160,8 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 	json_object_object_add(object, "type", json_object_new_string("con"));
 
 	if (c->parent) {
-		enum sway_container_layout layout = (c->parent->type == C_CONTAINER) ?
+		enum sway_container_layout layout =
+			(c->parent->type == C_CONTAINER && c->type == C_VIEW) ?
 			c->parent->layout : c->layout;
 
 		json_object_object_add(object, "layout",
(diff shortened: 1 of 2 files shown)


RyanDwyer:
I get some weird behaviour here. I created layout `T[view #1, view #2, view #3 focused]` where the workspace is the tabbed container, then ran `move right`. My layout then changed to `H[T[view #1, view #3 focused, view #2]]` where the workspace is splith and the new C_CONTAINER is tabbed. But it failed to move my view into the split, and instead moved it around within the tabbed container.

RedSoxFan:
@RyanDwyer Can you try again? Should be fixed now

RyanDwyer:
Yep, that one's fixed. But I found another.

Create layout `H[view, T[view focused, view]]` and run `move left`. The result is, believe it or not, `H[view, H[view, H[view]]]` and the inner-most view itself has layout=tabbed.

RedSoxFan:
> Create layout H[view, T[view focused, view]] and run move left. The result is, believe it or not, H[view, H[view, H[view]]] and the inner-most view itself has layout=tabbed.

The layout was actually correct. It was actually the IPC that was returning the wrong layout.

RyanDwyer:
I don't understand that IPC code. Why does it show the parent's layout instead of the current node's layout? And if it *is* supposed to show the parent's layout, why doesn't it do it if the parent is the workspace?

i3 just shows the current node's layout directly: https://github.com/i3/i3/blob/26f50898fd72c22a27e5b12db83808161bc3d9cd/src/ipc.c#L327

Sorry that this is diverging a bit.

RedSoxFan:
> I don't understand that IPC code. Why does it show the parent's layout instead of the current node's layout? And if it is supposed to show the parent's layout, why doesn't it do it if the parent is the workspace?

 I have no idea. I'm actually just as confused now.

 > i3 just shows the current node's layout directly: https://github.com/i3/i3/blob/26f50898fd72c22a27e5b12db83808161bc3d9cd/src/ipc.c#L327

I changed it to match that. I'm not noticing any issues

> Sorry that this is diverging a bit.

Honestly, I'm completely fine with that. Having correct IPC information is extremely useful for debugging/testing PRs like this one. I wouldn't have included the commits in this PR if I felt it was out of the scope.

RyanDwyer:
Thanks!

PR #2066, opened 2018-05-28
Support i3's legacy force_focus_wrapping command

RedSoxFan:
Implements support for i3's legacy `force_focus_wrapping yes|no` command.

This is a follow-up to #2060. @SirCmpwn commented after the merge that he wants support for this legacy command.


emersion:
Thanks!

PR #2067, opened 2018-05-28
Force fullscreen views to be fully opaque

RedSoxFan:
This forces fullscreen views to be fully opaque.

Test:
- Open a view
- Set opacity (Ex: `swaymsg opacity 0.5`)
- Fullscreen the view (`swaymsg fullscreen`)
- Observe a fully opaque view
- Unfullscreen the view (`swaymsg fullscreen`)
- Opacity should return to 0.5 (or whatever it was set to above)

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -307,11 +307,11 @@ static void premultiply_alpha(float color[4], float opacity) {
 }
 
 static void render_view_surfaces(struct sway_view *view,
-		struct sway_output *output, pixman_region32_t *damage) {
+		struct sway_output *output, pixman_region32_t *damage, float alpha) {
 	struct render_data data = {
 		.output = output,
 		.damage = damage,
-		.alpha = view->swayc->alpha,
+		.alpha = alpha,
 	};
 	output_view_for_each_surface(
 			view, &data.root_geo, render_surface_iterator, &data);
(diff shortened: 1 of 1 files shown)


sway/desktop/output.c
```diff
@@ -808,10 +808,15 @@ static void render_output(struct sway_output *output, struct timespec *when,
 			wlr_renderer_clear(renderer, clear_color);
 		}
 
+		float opacity = workspace->sway_workspace->fullscreen->swayc->alpha;
+		workspace->sway_workspace->fullscreen->swayc->alpha = 1.0f;
```
emersion: Not a fan of this

RedSoxFan: Would a `float saved_alpha` field in `struct sway_view` and store/restore it when changing the fullscreen status be fine?

emersion: Yeah, just like saved_width/saved_height, that would be better.

RedSoxFan: Fixed

SirCmpwn:
Why not just render the fullscreen view with a constant instead?

RedSoxFan:
> Why not just render the fullscreen view with a constant instead?

Currently the same function is called (`render_view_surfaces`) for any view (fullscreen or not). If preferred, I can alter the signature to take an alpha value instead of reading it from the passed view

SirCmpwn:
That sounds better to me.

SirCmpwn:
Thanks!

PR #2070, opened 2018-05-29
Make command block implementation generic

### LINKED ISSUE #547: Making the command block implementation generic
It would be nice if we tweaked the implementation of command blocks to work for any command. That is, things like this:

```
output HDMI-A-1 {
    pos 1920,0
    res 1920x1080
    bg ~/wallpaper.png
}
```

Would be expanded and run as three seperate commands:

```
output HDMI-A-1 pos 1920,0
output HDMI-A-1 res 1920x1080
output HDMI-A-1 bg ~/wallpaper.png
```

Existing command blocks would change so that this:

```
bar {
    status_command custom_statusbar
    position top
    colors {
        statusline #ffffff
        background #323232
        focused_workspace #32323200 #32323200 #ffffff
        inactive_workspace #00000000 #32323200 #666666
        active_workspace #00000000 #32323200 #666666
    }
}
```

Would be equivalent to this:

```
bar status_command custom_statusbar
bar position top
bar colors statusline #ffffff
bar colors background #323232
bar colors focused_workspace #32323200 #32323200 #ffffff
bar colors inactive_workspace #00000000 #32323200 #666666
bar colors active_workspace #00000000 #32323200 #666666
```

I think this would simplify our implementation of command blocks, simplify the documentation, and allow for more flexibility.

### LINKED ISSUE #1034: Incorrect parsing of config file
In case someone writes
```
mode "resize"
{
[…]
```
instead of
```
mode "resize" {
[…]
```
the configuration file is not read correctly and the bindings defined in the mode are not usable in normal mode.
The case that the brace is on the next line should be covered in the implemetation.

Another case is the definition of colors. If the colors have too much whitespace between the option itself, they are not processed correctly.
Whitespace between options and arguments should be ignored.

### LINKED ISSUE #1753: Add an output command block
The list of output options is getting pretty big. Allow output configuration to be specified by blocks, just like input devices:

```
output VGA-1 {
    scale 2
    resolution 1280x720
    transform flipped-90
}
```

RedSoxFan:
Overview
-------------
Fixes #547 - Generic Code Blocks
Fixed #1034 - Braces on following line
Fixes #1753 - Output Block

This allows for any command to be given in either block form or expanded form. All commands should be working now.

`bar` command
----
For `bar`, the following syntax structures should be working:
1. Block with optional id subcommand (currently in master)
     ```
    bar {
        id bar-one
        <bar-command>
    }
    # - or, if you prefer the brace on the following line --
    bar
    {
        id bar-one
        <bar-command>
    }
    ```
2. Identifier block (similar to input and seat in master)
    ```
    bar bar-one {
        <bar-command>
    }
    ```
3. Nested
    ```
    bar {
        bar-one {
            <bar-command>
        }
        bar-two {
            <bar-command>
        }
        bar-three <bar-command>
    }
    ```
4. Expanded with identifier
    ```
    bar bar-one <bar-commad>
    ``` 
5. Expanded with identifier on bar change
    ```
    bar bar-one <bar-command>
    bar <bar-command>
    bar bar-two <bar-command>
    bar <bar-command>
    ```

TODO
----------------------------------------------
- [X] Restructure functions to support any command being a given as a block or single line
- [X] Update current blocks to work again
    - [X] `mode`
    - [X] `bar` (See Above)
    - [X] `bar colors`
    - [X] `input`
    - [X] `seat`
- [X] Test all commands and update as necessary

### DIFF
--- a/common/readline.c
+++ b/common/readline.c
@@ -58,7 +58,7 @@ char *peek_line(FILE *file, int offset, long *position) {
 		if (read < 0) {
 			break;
 		}
-		if (line[read - 1] == '\n') {
+		if (read > 0 && line[read - 1] == '\n') {
 			line[read - 1] = '\0';
 		}
 	}
--- a/common/readline.c
+++ b/common/readline.c
@@ -56,6 +56,8 @@ char *peek_line(FILE *file, int line_offset, long *position) {
 	for (int i = 0; i <= line_offset; i++) {
 		ssize_t read = getline(&line, &length, file);
 		if (read < 0) {
+			free(line);
+			line = NULL;
 			break;
 		}
 		if (read > 0 && line[read - 1] == '\n') {
--- a/common/readline.c
+++ b/common/readline.c
@@ -48,6 +48,20 @@ char *read_line(FILE *file) {
 	return string;
 }
 
+char *peek_line(FILE *file, int offset) {
+	int pos = ftell(file);
+	char *line = NULL;
+	for (int i = 0; i <= offset; i++) {
+		free(line);
+		line = read_line(file);
+		if (!line) {
+			break;
+		}
+	}
+	fseek(file, pos, SEEK_SET);
+	return line;
+}
+
 char *read_line_buffer(FILE *file, char *string, size_t string_len) {
 	size_t length = 0;
 	if (!string) {
--- a/common/readline.c
+++ b/common/readline.c
@@ -1,3 +1,4 @@
+#define _POSIX_C_SOURCE 200809L
 #include "readline.h"
 #include "log.h"
 #include <stdlib.h>
@@ -48,15 +49,21 @@ char *read_line(FILE *file) {
 	return string;
 }
 
-char *peek_line(FILE *file, int offset) {
-	int pos = ftell(file);
-	char *line = NULL;
+char *peek_line(FILE *file, int offset, long *position) {
+	long pos = ftell(file);
+	size_t length = 1;
+	char *line = calloc(1, length);
 	for (int i = 0; i <= offset; i++) {
-		free(line);
-		line = read_line(file);
-		if (!line) {
+		ssize_t read = getline(&line, &length, file);
+		if (read < 0) {
 			break;
 		}
+		if (line[read - 1] == '\n') {
+			line[read - 1] = '\0';
+		}
+	}
+	if (position) {
+		*position = ftell(file);
 	}
 	fseek(file, pos, SEEK_SET);
 	return line;
--- a/common/readline.c
+++ b/common/readline.c
@@ -49,11 +49,11 @@ char *read_line(FILE *file) {
 	return string;
 }
 
-char *peek_line(FILE *file, int offset, long *position) {
+char *peek_line(FILE *file, int line_offset, long *position) {
 	long pos = ftell(file);
-	size_t length = 1;
-	char *line = calloc(1, length);
-	for (int i = 0; i <= offset; i++) {
+	size_t length = 0;
+	char *line = NULL;
+	for (int i = 0; i <= line_offset; i++) {
 		ssize_t read = getline(&line, &length, file);
 		if (read < 0) {
 			break;
--- a/include/readline.h
+++ b/include/readline.h
@@ -4,7 +4,7 @@
 #include <stdio.h>
 
 char *read_line(FILE *file);
-char *peek_line(FILE *file, int offset);
+char *peek_line(FILE *file, int offset, long *position);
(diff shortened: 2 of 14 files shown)


RedSoxFan:
I think this is ready for review.

common/readline.c
```diff
@@ -48,6 +48,20 @@ char *read_line(FILE *file) {
 	return string;
 }
 
+char *peek_line(FILE *file, int offset) {
+	int pos = ftell(file);
```
ascent12: The correct type is `long` for ftell.

common/readline.c
```diff
@@ -48,6 +48,20 @@ char *read_line(FILE *file) {
...
+	char *line = NULL;
+	for (int i = 0; i <= offset; i++) {
```
ascent12: I feel that `getline` from the standard library would be better for this, so you're not making so many allocations and then immediately freeing them.

sway/config.c
```diff
@@ -513,37 +512,92 @@ bool load_include_configs(const char *path, struct sway_config *config) {
...
+			}
+		} else {
+			lines = 0;
+		}
+		free(peeked);
+	}
+	return lines;
+}
+
+static char *expand_line(char *block, char *line, bool add_brace) {
```
ascent12: const arguments

sway/config.c
```diff
@@ -513,37 +512,92 @@ bool load_include_configs(const char *path, struct sway_config *config) {
...
+		+ (add_brace ? 2 : 0) + 1;
+	char *expanded = calloc(1, size);
+	if (!expanded) {
+		wlr_log(L_ERROR, "Cannot allocate expanded line buffer");
+		return NULL;
+	}
+	strcat(expanded, block ? block : "");
+	strcat(expanded, block ? " " : "");
+	strcat(expanded, line);
+	strcat(expanded, add_brace ? " {" : "");
```
ascent12: I'm really not a fan of strcat. Using snprintf would be a lot better.

include/sway/commands.h
```diff
@@ -54,7 +56,12 @@ struct cmd_results *execute_command(char *command,  struct sway_seat *seat);
  *
  * Do not use this under normal conditions.
  */
-struct cmd_results *config_command(char *command, enum cmd_status block);
+struct cmd_results *config_command(char *command);
+/**
+ * Parse and handle a sub command
+ */
+struct cmd_results *subcommand(char **argv, int argc,
```
emersion: Maybe this should be renamed `config_subcommand`?

include/sway/commands.h
```diff
@@ -54,7 +56,12 @@ struct cmd_results *execute_command(char *command,  struct sway_seat *seat);
...
+		struct cmd_handler *handlers, int handlers_size);
```
emersion: `handlers_size` could be a `size_t`

common/readline.c
```diff
@@ -48,6 +49,26 @@ char *read_line(FILE *file) {
...
+char *peek_line(FILE *file, int offset, long *position) {
+	long pos = ftell(file);
+	size_t length = 1;
+	char *line = calloc(1, length);
+	for (int i = 0; i <= offset; i++) {
+		ssize_t read = getline(&line, &length, file);
+		if (read < 0) {
+			break;
+		}
+		if (line[read - 1] == '\n') {
```
emersion: What if `read == 0`?

common/readline.c
```diff
@@ -48,6 +49,26 @@ char *read_line(FILE *file) {
 	return string;
 }
 
+char *peek_line(FILE *file, int offset, long *position) {
+	long pos = ftell(file);
+	size_t length = 1;
+	char *line = calloc(1, length);
```
emersion: No need to `calloc`, you can set `length = 0` and `line = NULL` and `getline` will do that for you.

include/readline.h
```diff
@@ -4,6 +4,7 @@
 #include <stdio.h>
 
 char *read_line(FILE *file);
+char *peek_line(FILE *file, int offset, long *position);
```
emersion: Can you rename `offset` to `line_offset` so that it's obvious it's a number of lines? (vs number of bytes with `position`)

emersion:
Is it possible to refactor the `output` command to use `config_subcommand` too?

emersion:
I was able to trigger once an infinite loop of `peek_line` with this config:

```
output "WL-1"


```

(An unfinished `ouput` block at the end of the config, with empty newlines)

RedSoxFan:
> I was able to trigger an infinite loop of peek_line with this config:
Should be fixed. Forgot to set line to NULL at EOF when I switched to using getline

SirCmpwn:
Thanks!

Issue #1859, opened 2018-04-24
Use-after-free when closing sway with views


emersion:
The `container_destroy` functions doesn't call `view_destroy`. And it cannot, since `view_destroy` calls `container_destroy`.

We do want a generic way of destroying containers, so we should keep `container_destroy`. `container_destroy` should call type-specific destructors like `view_destroy`, `output_destroy`. These type-specific destructors should call a new function to destroy their `sway_container`.

```
==14836==ERROR: AddressSanitizer: heap-use-after-free on address 0x60d0000e5ab0 at pc 0x557d27ea7d05 bp 0x7ffe66a51ca0 sp 0x7ffe66a51c90
READ of size 8 at 0x60d0000e5ab0 thread T0
    #0 0x557d27ea7d04 in container_destroy ../sway/tree/container.c:269
    #1 0x557d27eb380e in view_unmap ../sway/tree/view.c:256
    #2 0x557d27e7afce in handle_unmap ../sway/desktop/xwayland.c:237
    #3 0x7fa505801fb4 in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #4 0x7fa505769876 in xsurface_unmap ../subprojects/wlroots/xwayland/xwm.c:672
    #5 0x7fa505766af4 in wlr_xwayland_surface_destroy ../subprojects/wlroots/xwayland/xwm.c:283
    #6 0x7fa50576dd77 in xwm_destroy ../subprojects/wlroots/xwayland/xwm.c:1307
    #7 0x7fa505762fc9 in wlr_xwayland_finish ../subprojects/wlroots/xwayland/xwayland.c:140
    #8 0x7fa505764a70 in wlr_xwayland_destroy ../subprojects/wlroots/xwayland/xwayland.c:343
    #9 0x7fa50576348f in handle_display_destroy ../subprojects/wlroots/xwayland/xwayland.c:198
    #10 0x7fa503c31b81  (/usr/lib/libwayland-server.so.0+0x7b81)
    #11 0x7fa503c32176 in wl_display_destroy (/usr/lib/libwayland-server.so.0+0x8176)
    #12 0x557d27e5dbb7 in server_fini ../sway/server.c:129
    #13 0x557d27e5cbcd in main ../sway/main.c:423
    #14 0x7fa5034449a6 in __libc_start_main (/usr/lib/libc.so.6+0x219a6)
[...]

PR #2072, opened 2018-05-30
WIP: Atomic layout updates

RyanDwyer:
This is a very early concept/WIP of atomic layout updates. It's so early that my code won't even compile. You should consider it a proposal with some example code attached. I'm submitting it here so everyone can see the direction I'm going in and give feedback as the feature is developed.

The new data structures I'm proposing are:

* `struct sway_transaction`
* `struct sway_transaction_instruction`
* `struct sway_container_state`

The basic idea here is that when we want to arrange layout, we create a `struct sway_transaction`. We add instructions to the transaction - one instruction per container involved in the arrangement. Then we commit the transaction which sends all the configure events. Then when all the configures are done, or a time has passed, we apply the transaction which updates the tree accordingly.

Each instruction has a pointer to a container, a pending state which it'll apply to the container, and the configure serial.

I was originally going to have a `current` and `pending` state in each container, but realised that would be tedious to update the codebase to change `container->x` to `container->current.x`, so at the moment there's just `pending` and the current stuff uses the normal properties.

When arranging layout, it builds the new pending state from the container's existing pending state. This is so a second transaction running at the same time won't revert what the first transaction is trying to change. This means if no transactions are in progress, every container's pending state must be an accurate reflection of the current state.

As for how windows are arranged by commands and so on, if you only want to do one arrange, you'll be able to call `arrange_and_commit(container)`. This creates a new transaction and commits it for you. If you need to do two arranges in the one transaction, such as moving a view from one visible workspace to another visible workspace, you can take the more manual approach:

    struct sway_transaction *transaction = transaction_create();
    arrange_windows(old_workspace, transaction);
    arrange_windows(new_workspace, transaction);
    transaction_commit(transaction);

You may notice this means there's only two entry points for the arrange functions, rather than one for each container type as we had before.


RyanDwyer:
I've updated the PR so it's now a basic working implementation. Atomic updates are implemented for the `resize` command only. Almost everything else is broken because everything else writes to the current properties/state, which get overwritten with the pending state when the transaction is applied.

The PR is ready to stash the last good surface texture away, but we need to have a good discussion on whether that's actually what we want to do, because apparently reading textures back from the GPU is expensive. So in lieu of a solution, if the transaction stalls for a view then you'll see the other surfaces resize before their borders and everything else does.

Ideas mentioned in chat for preserving textures are:

* Copy the texture as originally planned (bad due to being expensive), or
* Have two wlr_textures for each surface, and swap them when we want to preserve one, or
* Expose the pixel data to sway so we can make a new texture from it, or
* Don't damage/rerender the area being updated until everything is ready (doesn't work if there's a layer above or below that needs to be updated)

Anyway, when you resize a split, here's what's happening:

* The `resize` command calls `container_recursive_resize`, which now sets the new dimensions in the container's pending state rather than directly to the container's width and height properties.
* The `resize` command calls `arrange_children_of`. This function will be removed in favour of `arrange_and_commit`, but to make rebasing easier it exists as a wrapper around `arrange_and_commit`.
* `arrange_and_commit` does the following:
    * Creates a new transaction
    * Arranges the children as it did previously, but with the following differences:
        * Everything is saved into the pending properties rather than the swayc's main properties.
        * `view_autoconfigure` no longer calls `view_configure` and instead sets the `view_{x,y,width,height}` in the pending state.
    * Adds each processed child to the transaction.
    * Adds a single damage box to the transaction - the same box as the topmost parent's geometry (note the topmost parent does not move or change size, so only one damage box is needed and not a before and after).
    * Commits the transaction with `transaction_commit`.
* `transaction_commit` iterates through the instructions/containers/configures in the transaction.
	* For views, `view_configure` is called and the serial is stored. The view's current surface texture is copied into `view->saved_texture`.
	* For `C_CONTAINER`s and anything else, the instruction is marked as ready instantly.
* `transaction_commit` prepares a timer (currently set at 200ms). If any views have not acknowledged the configure by then, the transaction will be applied anyway.
* When rendering, the main surface of each view will use the `view->saved_texture` if it exists.
* As each view commits, it calls `transaction_notify_view_ready` with the view and the serial from that commit.
* `transaction_notify_view_ready` marks off the view in the transaction and applies the transaction if all are ready.
* `transaction_apply` copies the pending state into the main swayc properties and removes the `view->saved_texture`.
* `transaction_apply` applies the damages that is stored in the transaction.
* The transaction is then destroyed.

A good test is to change the `TIMEOUT_MS` from 200 to 2000 (2 seconds). Then download emersion's hello-wayland program and rewrite the configure handler so it ignores subsequent configures (ie. making it a misbehaving client):

    static void xdg_surface_handle_configure(void *data,
            struct xdg_surface *xdg_surface, uint32_t serial) {
        static bool done_first = false;
        if (!done_first) {
            xdg_surface_ack_configure(xdg_surface, serial);
            done_first = true;
        }
    }

Then open an xdg_shell terminal, use that terminal to launch hello-wayland, then use the `resize` command. You'll see it stall for 2 seconds before applying. You'll also see the surface resize before the borders do, due to the lack of texture preservation as mentioned in the second paragraph.

Moving forward from here:

* We'll need to change the entire codebase to write into the pending state instead of the current properties. A good way to do this would be to remove the current properties and have a `swayc->current` state. That way we know we'll have updated all references because it won't compile otherwise.
* We need to handle tree modification operations, such as creating containers, reparenting containers and deleting containers. To do this I think we'll need to store new children lists in the pending state, and that makes for ugly code like `container->pending.parent->pending.children`. I don't know of a better way to do this.
* We'll need to handle cases where transactions queue up, like when you repeatedly resize a split.

emersion:
>Copy the texture as originally planned

This seems pretty hard, there's no direct way to do it. We could:
* Download the texture data and re-upload it. Very expensive. I'd advise against it.
* Render the texture to a pbuffer/FBO and use `glCopyTexImage2D`. A little bit better, but still expensive.

>Have two wlr_textures for each surface, and swap them when we want to preserve one

I'd rather "steal" the texture from the `wlr_surface` and force the `wlr_surface` to re-create a new texture next time the client commits. I don't think this strategy works for DMA-BUFs, as the client can update its texture without making the compositor import it again?

>Expose the pixel data to sway so we can make a new texture from it

The pixel data is in the GPU, so it's tricky to just read it.

SirCmpwn:
Read through your details about how this works, seems good to me.

Can we worry about "what to do with old textures" later?

RyanDwyer:
I'm making good progress on this.

I decided to make it so the main swayc/view properties are the *pending* state rather than the current state. This is because so much of the code works with these properties, and they need to work with the pending state. The only exception is rendering code which works with the current state. So the good news is everyone can continue to write into `container->x` and so on everywhere except in the rendering code.

BTW, I consider the main/pending properties to be the "live" state, and the current state to be what the renderer displays which might be a few frames behind. So IPC events are sent when the live/pending state is changed.

I discovered we had `arrange_foo` statements in both high level functions and low level functions. I wouldn't be surprised if moving a view could trigger several arranges of the same containers. I've moved these out to be high level only. So arrange functions are typically called from command handlers or from a view's map/unmap handler.

I've made floating, fullscreen, and floating + fullscreen all work with transactions.

What's left now is:

* Handling tree operations.
* Saving the texture.
* Making the title and marks textures atomic.

emersion:
Texture discussion: https://github.com/swaywm/wlroots/issues/1046

RyanDwyer:
Update on this: It now preserves the buffer/texture using the `wlr_buffer` refcounting. I am now waiting on a way to access the buffer after the view has unmapped (see https://github.com/swaywm/wlroots/issues/1065).

emersion:
>I am now waiting on a way to access the buffer after the view has unmapped (see swaywm/wlroots#1065).

Note that this should already work with xdg-shell views.

RyanDwyer:
This is now ready for testing. I'm going to start using this as my daily driver.

My latest commit implements atomic layouts for containers that are mapping, reparenting or unmapping.

How it works is we still have the regular tree, but the properties you're currently familiar with may not reflect what is actually rendered. Each container has a `current` state, and that state contains the properties that are rendered as well as another `children` list. When rendering, the `con->current.children` list is used rather than `con->children`. In most cases we can continue working with the tree as normal, and if a transaction is applied to a section of the tree then it'll eventually get updated in the rendered/current state.

Another key point is that destroying containers is now split into two steps: destroying and freeing. Destroying removes it from the tree (but keeps it in the `current.children` list of its old parent). When the container is no longer involved in a transaction (ie. removed from on-screen) then it'll be freed via `container_free`.

For views that are unmapping and destroying, a similar system is used. An unmapped view will still contain a reference to its swayc until it's no longer in a transaction. For this reason we can't use `view->swayc` to determine if a view is mapped, so a better check is `view->surface`. `view->surface` is set to `NULL` immediately at the end of each implementation's unmap handler.

The `destroy` callback in the view implementations has been renamed to `free`, to keep the terms consistent across the project.

To test what happens during a transaction, you can set `TRANSACTION_DEBUG` to `true` at the top of transaction.c. This makes it always wait for the timeout to expire before applying the new layout, even if the views are ready. You'll also want to bump up the `TIMEOUT_MS` to something like 1000 (1 second). Doing both of these allows you to observe what's rendered while a transaction is in progress, and test what happens if you interact with things while transactions are in progress. For example, launch a view then launch another one before the first one has appeared on screen.

If you use `TRANSACTION_DEBUG` and open several views, you will notice that no view is shown as focused while the transaction is happening. This is because the focus works with the pending tree, and your focus is on the view that hasn't appeared on screen yet. I think this is desirable behaviour, because it allows you to hammer out commands (eg. splitting and opening) without having to wait, and the rendered state will catch up.

Possible issues or other points of discussion:

* When a transaction is happening, I'm still sending frame done to everything despite rendering a saved buffer. I'm not sure what the consequences of this are. I haven't seen any adverse effects and mpv seems to be fine with it.
* `container_at` uses the pending state of the tree, not what's shown on screen, but with a timeout of 200ms I don't think we need to worry about this.
* In theory, transactions that complete out of order may mess up the rendered state. I'm not sure how likely this is in practice though as there's usually only one transaction running at a time. So my strategy here is to wait and see if it's a problem, and if so I'll look at a solution. If it does happen, it'd be recoverable by doing another arrange operation (eg. switch workspace and back).

atomnuker:
Works well, haven't found faults yet. One thing is that when toggling fullscreen any clients using the layer shell don't get damaged until they flip, but I'm guessing this would be fixed by proper layer shell client management (which would fix other bugs too probably, such as the damage from layer shell clients not affecting fullscreen clients).

atomnuker:
Actually I can crash it easily. Create 2 windows, make either of them stacking, create a few tabs in the stacking container and then close all of them one by one. Closing the last one will crash the compositor.

atomnuker:
Running mpv on one workspace and quickly opening and closing another mpv instance on the same workspace crashes the compositor. Works with any video and works even if mpv hasn't had a chance to open a window yet.

SirCmpwn:
This works pretty well on a cursory test, but it occurs to me that we're not going to be able to stress test this without something like interactive resize support.

SirCmpwn:
https://sr.ht/p99v.mkv

sway/desktop/transaction.c
```diff
@@ -0,0 +1,292 @@
...
+	wlr_log(L_DEBUG, "Transaction %p committing with %i instructions",
+			transaction, transaction->instructions->length);
+	transaction->num_waiting = 0;
+	for (int i = 0; i < transaction->instructions->length; ++i) {
+		struct sway_transaction_instruction *instruction =
+			transaction->instructions->items[i];
+		struct sway_container *con = instruction->container;
+		if (con->type == C_VIEW && !con->destroying &&
+				(con->current.view_width != instruction->state.view_width ||
+				 con->current.view_height != instruction->state.view_height)) {
```
SirCmpwn: These parens aren't necessary, || has a higher precedence than &&

emersion: Removing these parens also makes the code unreadable.

SirCmpwn: nah

SirCmpwn:
Codewise this looks pretty good. I'm a little concerned about thrashing the heap all the time but I'm not sure what alternative to offer. Need to test it more but I'm gonna wait until the current round of bugs is fixed.

sway/desktop/transaction.c
SirCmpwn: lol I'm dumb

martinetd:
`render_container_stacked` can still call `render_view` with NULL `colors`, happens for me with this repr "H[S[URxvt] URxvt]" (automatically happens with stacked workspace_layout stacking + default_orientation vertical, pop two windows, move right) then move the right window to the left.

```
    #0 0x426f95 in render_view ../sway/desktop/output.c:408
    #1 0x42877c in render_container_stacked ../sway/desktop/output.c:819
    #2 0x42877c in render_container ../sway/desktop/output.c:837
    #3 0x427b63 in render_container_simple ../sway/desktop/output.c:699
    #4 0x427b63 in render_container ../sway/desktop/output.c:834
    #5 0x429d18 in render_output ../sway/desktop/output.c:984
    #6 0x429d18 in damage_handle_frame ../sway/desktop/output.c:1126
    #7 0x7f6c5319ece8 in wlr_signal_emit_safe ../util/signal.c:29
    #8 0x7f6c5319ece8 in wlr_signal_emit_safe ../util/signal.c:29
    #9 0x7f6c5340fe73 in wl_event_loop_dispatch_idle src/event-loop.c:600
    #10 0x7f6c5340ff24 in wl_event_loop_dispatch src/event-loop.c:646
    #11 0x7f6c5340e601 in wl_display_run src/wayland-server.c:1260
    #12 0x40a2e4 in main ../sway/main.c:433
    #13 0x7f6c528e918a in __libc_start_main ../csu/libc-start.c:308
    #14 0x40b739 in _start (/opt/wayland/bin/sway+0x40b739)
```

martinetd:
Also I'm not 100% sure what happens but with this crash and urxvt my whole computer locks up (who said starting sway from sway/X would let one debug easily?), even a remote ssh preconnected hangs almost immediately so it's a bit of a pain to investigate.
This doesn't happen with gnome-terminal so definitely X/Xwayland specific, and doesn't happen if I pkill -11 sway so it has to crash during one of these "transactions"
I straced Xwayland and it's stuck in "`ioctl(5, DRM_IOCTL_I915_GEM_SET_DOMAIN <unfinished ...>`" so sway crashed during that ioctl and the intel driver can't handle the compositor crashing at this point? That's gonna be fun to debug.... Well at least I fixed kdump and I know where to look now, will keep prodding at this a bit today.

martinetd:
Also2, hits sway_assert when trying to move in an empty workspace... happened by chance (urxvt not installed in my VM so windows didn't pop like I expected them to)

```
2018-06-24 10:36:37 - [sway/commands.c:166] find_handler(move)
2018-06-24 10:36:37 - [sway/tree/layout.c:340] container_move:Can only move containers and views, got C_WORKSPACE
    #0 0x471bf4 in _sway_assert ../common/log.c:27
    #1 0x464f4d in container_move ../sway/tree/layout.c:337
    #2 0x44a2b7 in move_in_direction ../sway/commands/move.c:181
    #3 0x44a90e in cmd_move ../sway/commands/move.c:211
    #4 0x40ed01 in execute_command ../sway/commands.c:291
    #5 0x43d464 in keyboard_execute_command ../sway/input/keyboard.c:128
    #6 0x43e1f7 in handle_keyboard_key ../sway/input/keyboard.c:291
    #7 0x7fe6cce48ce8 in wlr_signal_emit_safe ../util/signal.c:29
    #8 0x7fe6cce2a65f in wlr_keyboard_notify_key ../types/wlr_keyboard.c:125
    #9 0x7fe6ccdf9dd8 in handle_x11_input_event ../backend/x11/input_device.c:61
    #10 0x7fe6ccdf8efe in handle_x11_event ../backend/x11/backend.c:38
    #11 0x7fe6ccdf8efe in x11_event ../backend/x11/backend.c:84
    #12 0x7fe6cd0b9f01 in wl_event_loop_dispatch src/event-loop.c:641
    #13 0x7fe6cd0b8601 in wl_display_run src/wayland-server.c:1260
    #14 0x40a2e4 in main ../sway/main.c:433
    #15 0x7fe6cc59318a in __libc_start_main ../csu/libc-start.c:308
    #16 0x40b739 in _start (/opt/wayland/bin/sway+0x40b739)
```

RyanDwyer:
@atomnuker
> Running mpv on one workspace and quickly opening and closing another mpv instance on the same workspace crashes the compositor. Works with any video and works even if mpv hasn't had a chance to open a window yet.

The mpv issue wasn't just related to mpv. Opening any view then closing it before it appeared on screen would cause the closing transaction to complete before the opening transaction. I actually had this working during development but accidentally broke it when I fixed another issue. To fix this, I've made it use a linked list of transactions and force them to apply in order.

@SirCmpwn 

I suspect the overhead of spawning the swaymsg process 100 times and connecting to the IPC socket makes the test too inaccurate. But rather than implement interactive resize, an easier option might be to allow held keybindings to repeat, then hold your keybind for `resize shrink width 1px`. The tests also depend on how long the views take to respond to their configures, so if we're serious about performance then a good first step would be to obtain figures of how much time is spent waiting for views vs not waiting for views.

@martinetd 
> render_container_stacked can still call render_view with NULL colors

I've fixed the colours thing by initialising them to the unfocused colour. I think we'll ultimately have to track which view is visibly focused separately to which view is actually focused. I'm not sure if that should be done as part of this PR though.

I've fixed some related issues to moving in and out of tabs and stacks too.

> Also2, hits sway_assert when trying to move in an empty workspace

Fixed.

Everything I'm aware of is fixed now.

martinetd:
I still get a segfault with stacking (workspace_layout stacking + default_orientation vertical) and just moving left/right a few times randomly: right, left, left, right, right.
Representation before last right is "H[H[S[URxvt URxvt]]]".
(Yes, workspace_layout stacking is screwed up; I'll accept you saying "H[H[S[X Y]]]]" shouldn't happen if that's true)

(lines are probably a bit off in layout.c as I'm dragging along a couple of debugging prints to try to debug that, say if it's too annoying for you)

```
2018-06-24 15:06:16 - [sway/commands.c:166] find_handler(move)
2018-06-24 15:06:16 - [sway/tree/layout.c:367] Visiting 0x61400003c840 C_VIEW 'asmadeus@fenrir:~'
2018-06-24 15:06:16 - [sway/tree/layout.c:320] Moving out of tab/stack into a split
2018-06-24 15:06:16 - [sway/tree/layout.c:849] creating container 0x614000044c40 around 0x61400003da40 (C_CONTAINER)
2018-06-24 15:06:16 - [sway/tree/arrange.c:327] Removing gaps: not gapped: 0x61400003da40
2018-06-24 15:06:16 - [sway/tree/arrange.c:343] Not adding gaps: 0x614000044c40
2018-06-24 15:06:16 - [sway/tree/layout.c:132] Adding 0x61400003da40 (3, 0.000000x0.000000) to 0x614000044c40 (3, 1920.
000000x1061.000000)
2018-06-24 15:06:16 - [sway/tree/layout.c:104] Inserting id:5 at index 1
2018-06-24 15:06:16 - [sway/ipc-server.c:304] Sending window::close event
==25162==The signal is caused by a READ memory access.
==25162==Hint: address points to the zero page.
    #0 0x46266b in container_remove_child ../sway/tree/layout.c:146
    #1 0x45e25b in container_destroy ../sway/tree/container.c:402
    #2 0x45e3f8 in container_flatten ../sway/tree/container.c:387
    #3 0x4650f7 in move_out_of_tabs_stacks ../sway/tree/layout.c:329
    #4 0x44a97b in move_in_direction ../sway/commands/move.c:190
    #5 0x44b12e in cmd_move ../sway/commands/move.c:223
    #6 0x40ed01 in execute_command ../sway/commands.c:291
    #7 0x43d774 in keyboard_execute_command ../sway/input/keyboard.c:128
    #8 0x43e507 in handle_keyboard_key ../sway/input/keyboard.c:291
    #9 0x7f8c2bf2cce8 in wlr_signal_emit_safe ../util/signal.c:29
    #10 0x7f8c2bf0e65f in wlr_keyboard_notify_key ../types/wlr_keyboard.c:125
    #11 0x7f8c2becc24a in handle_keyboard_key ../backend/libinput/keyboard.c:73
    #12 0x7f8c2becac8a in handle_libinput_readable ../backend/libinput/backend.c:35
    #13 0x7f8c2c19df01 in wl_event_loop_dispatch src/event-loop.c:641
    #14 0x7f8c2c19c601 in wl_display_run src/wayland-server.c:1260
    #15 0x40a2e4 in main ../sway/main.c:433
    #16 0x7f8c2b67718a in __libc_start_main ../csu/libc-start.c:308
    #17 0x40b739 in _start (/opt/wayland/bin/sway+0x40b739)
```

RyanDwyer:
@martinetd Should be fixed now.

SirCmpwn:
>I suspect the overhead of spawning the swaymsg process 100 times and connecting to the IPC socket makes the test too inaccurate. But rather than implement interactive resize, an easier option might be to allow held keybindings to repeat, then hold your keybind for resize shrink width 1px. The tests also depend on how long the views take to respond to their configures, so if we're serious about performance then a good first step would be to obtain figures of how much time is spent waiting for views vs not waiting for views.

It would be nice if we had some code in place which measured this stuff, and printed to the debug log how long we spent calculating the layout, how long each client took to get configured, and how many frames elapsed before the first correct frame was rendered.

SirCmpwn:
Probably hidden behind a flag like the debug overlay is.

RyanDwyer:
I've added the debug option. To enable, launch sway with `SWAY_DEBUG=txn_timings`. One transaction can span multiple outputs and those outputs can have different refresh rates, so for the number of frames it assumes 60Hz.

An example log message is:

    Transaction 0x5599940d4e70: 0.0ms arranging, 24.0ms waiting, 24.0ms total (1.5 frames if 60hz)

After this is merged, I plan on standardising how we handle debugging. I don't like `strcmp`ing environment variables multiple times (see `SWAY_DAMAGE_DEBUG`). I would make `SWAY_DEBUG` a comma separated string which is used to populate a bitfield `server.debug_options`. Then in various places we can check that to enable debug features. Another benefit of this is we could enable and disable debug features at runtime using something like `swaymsg debug 'foo,bar'`.

SirCmpwn:
Something I particularly wanted to see was a measure of which clients took the longest.

>After this is merged, I plan on standardising how we handle debugging. I don't like strcmping environment variables multiple times (see SWAY_DAMAGE_DEBUG). I would make SWAY_DEBUG a comma separated string which is used to populate a bitfield server.debug_options. Then in various places we can check that to enable debug features. Another benefit of this is we could enable and disable debug features at runtime using something like swaymsg debug 'foo,bar'.

I was thinking something more like getopt debug flags:

```
sway -Ddisplay-tree -Dtxn-timings -Ddamage
```

These would naturally be unsupported options and we'd reserve the right to change/remove them at will.

RyanDwyer:
Done. New example log output (using two Firefox instances):

    Transaction 0x564224623e40 committing with 4 instructions
    Transaction 0x564224623e40: 1/2 ready in 4.9ms (Mozilla Firefox)
    Transaction 0x564224623e40: 2/2 ready in 5.1ms (Mozilla Firefox)
    Transaction 0x564224623e40 is ready
    Applying transaction 0x564224623e40
    Transaction 0x564224623e40: 0.5ms arranging, 5.2ms waiting, 5.7ms total (0.4 frames if 60hz)

SirCmpwn:
Nice!

martinetd:
Cool, I agree it's useful to be able to log.

One more (last?) nitpick for stacks:

When using stacked and spawning a new window, you can clearly see the previous window disappear (background shows) then the new window appears with its decoration where the desktop was. It would be more natural to keep showing the pervious window in the background for as long as possible.

When closing a window in a stack, something else feels off: I immediately see the window behind appear in the spot of the closed window (e.g. what would happen with focus up/down), but that window's decoration is still there, then that window decoration disappears, then current window resizes. I think in this case it would be more natural to keep showing the dead window's last frame whlie the window behind resizes in the background and then displayed the resized window behind in a single transition.


At least none of the crashes I got this morning were related to this, so I think on that part it's starting to look good :+1:

RyanDwyer:
Yeah, stacks (and tabs) are a bit weird because there are no separated pending and current states for the focus stack. This means when you add or remove a view in a stacked/tabbed container, it tries to render the new focus immediately.

I haven't given much thought to the focus stack yet. We could copy the entire focus stack, or perhaps we could store a property in each container to say whether it should be rendered with focus/focused_inactive/unfocused styling. I'll come up with a good solution, but I'd rather do this in a follow up PR.

include/sway/tree/view.h
```diff
@@ -29,15 +29,15 @@ struct sway_view_impl {
...
+	uint32_t (*configure)(struct sway_view *view, double lx, double ly,
+			int width, int height);
 	void (*set_activated)(struct sway_view *view, bool activated);
 	void (*set_fullscreen)(struct sway_view *view, bool fullscreen);
 	bool (*wants_floating)(struct sway_view *view);
 	void (*for_each_surface)(struct sway_view *view,
 		wlr_surface_iterator_func_t iterator, void *user_data);
 	void (*close)(struct sway_view *view);
-	void (*destroy)(struct sway_view *view);
+	void (*free)(struct sway_view *view);
```
martinetd: (nitpick) Even if the implementation has been refactored to separate the rest of the destroy (now in surface notify destroy) and this function that is literally a free with extra checks, I'd keep the name `destroy` here as that's what we have all around in the impl vectors.
You never know if in the future that'll need to do a bit more work again, and it won't be the first destroy function that just frees its argument :)

sway/desktop/xdg_shell.c
```diff
@@ -146,16 +143,12 @@ static void _close(struct sway_view *view) {
 	}
 }
 
-static void destroy(struct sway_view *view) {
+static void _free(struct sway_view *view) {
 	struct sway_xdg_shell_view *xdg_shell_view =
 		xdg_shell_view_from_view(view);
 	if (xdg_shell_view == NULL) {
```
martinetd: That check is technically no longer required, but I guess it doesn't hurt, just wanted to bring it up

sway/desktop/transaction.c
```diff
@@ -0,0 +1,395 @@
...
+	for (i = 0; i < transaction->damage->length; ++i) {
+		struct wlr_box *box = transaction->damage->items[i];
+		for (int j = 0; j < root_container.children->length; ++j) {
+			struct sway_container *output = root_container.children->items[j];
+			output_damage_box(output->sway_output, box);
+		}
+	}
+}
+
+static void progress_queue() {
```
martinetd: It's a problem we have a lot in sway but I think even static function names should have a semi-obvious name that says what they work on, `progress_queue` in particular might be difficult for someone to understand in a stack trace here. I'm as bad with names as anyone else though, `transaction_process` or something? (might turn out difficult to differentiate with apply?)

Well, just a thought.

RyanDwyer: Yeah, I'm open to ideas. I don't think it should be something like `transaction_process` though because that implies it "processes a transaction" when in reality it iterates the queue and processes as many as it can. I think it should be `<something>_queue`.

Some ideas:

* `advance_queue`
* `check_queue`
* `handle_queue`
* `process_queue`
* `progress_queue`
* `run_queue`
* `work_queue`

You could also also add the word `transaction` to the above so it's `advance_transaction_queue` and so on.

martinetd: just `transaction_progress_queue` is fine then, it's the main thing I find could be confusing.

include/sway/tree/view.h
RyanDwyer: I renamed it to `free` to keep it consistent with the `view_destroy` and `view_free` functions, and `container_destroy` and `container_free`. Destroying and freeing are now two different things. Destroying basically marks it as destroying, and when it's no longer in a transaction it gets freed.

When a surface wants to destroy itself, we must unregister the wayland events before returning from `handle_destroy` because after we return the surface is no longer valid. But we can't free the view until later, when the view is no longer in a transaction. The non-atomic code in master does the unregistering and freeing of the view all in the `destroy` callback, using a call stack of `handle_destroy -> view_destroy -> view->destroy()`.

My branch takes a shortcut by unregistering the events in `handle_destroy`, and `view_destroy` doesn't have to dive back into the implementation at all. I could add the `destroy` callback back and call it from `view_destroy`, but it needs to be separate from the `free` callback. The only benefit of doing it that way is the compositor could theoretically initiate a destroy without having the view requesting it. I'm not sure if that's against the protocol though, or if we'd ever use it.

atomnuker:
Leaving mpv in fullscreen mode to quit at the end of a file crashes the compositor.

RyanDwyer:
Fixed. It happened when a view is a grandchild or deeper of the workspace, is fullscreen, and unmaps. The workspace would not be included in the transaction and its pointer to the fullscreen view was left dangling.

include/sway/tree/view.h
emersion: Yeah, I also think a `destroy` function which doesn't frees is confusing. I'd rather keep destroy and pick another name for the other function.

SirCmpwn:
Couple issues:

- When fullscreened, qutebrowser's left border is shown
- When exiting fullscreen, there are damage problems on swaybar

SirCmpwn:
Seems those issues only appear if qutebrowser fullscreens itself, e.g. by fullscreening a video.

martinetd:
I just noticed a couple of crashes when destroying an output (unplugging screen with DRM, but works fine with x11/wayland backends and WLR_X11/WL_OUTPUTS=2 and forcibly closing a window)

First might not be new, cba testing and will let you do whatever you want with this:
```diff
From c7f86e7c878a9283b7c2b12939fb412244f5dd7e Mon Sep 17 00:00:00 2001
From: Dominique Martinet <asmadeus@codewreck.org>
Date: Tue, 26 Jun 2018 16:32:10 +0900
Subject: [PATCH] ipc-json: allow describe_output with no ws

Fixes an obvious segfault when destroying an output
=26150==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f9707fefa01 bp 0x7ffe6456aed0 sp 0x7ffe6456a648 T0)
==26150==The signal is caused by a READ memory access.
==26150==Hint: address points to the zero page.
    #0 0x7f9707fefa00 in __strlen_avx2 (/lib64/libc.so.6+0x15ca00)
    #1 0x7f970a8a980b  (/lib64/libasan.so.5+0x9980b)
    #2 0x7f9709b546de in json_object_new_string (/lib64/libjson-c.so.4+0x56de)
    #3 0x418e23 in ipc_json_describe_output ../sway/ipc-json.c:123
    #4 0x418e23 in ipc_json_describe_container ../sway/ipc-json.c:231
    #5 0x418348 in ipc_json_describe_container_recursive ../sway/ipc-json.c:249
    #6 0x41bb2d in ipc_event_window ../sway/ipc-server.c:307
    #7 0x45d684 in container_destroy_noreaping ../sway/tree/container.c:329
    #8 0x45e59b in container_destroy ../sway/tree/container.c:406
    #9 0x7f97087979fa in wlr_signal_emit_safe ../util/signal.c:29
    #10 0x7f9708778451 in wlr_output_damage_destroy ../types/wlr_output_damage.c:90
    #11 0x7f9708777f32 in output_handle_destroy ../types/wlr_output_damage.c:13
    #12 0x7f97087979fa in wlr_signal_emit_safe ../util/signal.c:29
    #13 0x7f970870c602 in drm_connector_cleanup ../backend/drm/drm.c:1055
    #14 0x7f970870b3a0 in scan_drm_connectors ../backend/drm/drm.c:918
    #15 0x7f9708702114 in drm_invalidated ../backend/drm/backend.c:107
    #16 0x7f97087979fa in wlr_signal_emit_safe ../util/signal.c:29
    #17 0x7f970871c8a3 in udev_event ../backend/session/session.c:51
    #18 0x7f9708a0ef01 in wl_event_loop_dispatch src/event-loop.c:641
    #19 0x7f9708a0d601 in wl_display_run src/wayland-server.c:1260
    #20 0x40a2f4 in main ../sway/main.c:433
    #21 0x7f9707eb618a in __libc_start_main ../csu/libc-start.c:308
    #22 0x40b749 in _start (/opt/wayland/bin/sway+0x40b749)

diff --git a/sway/ipc-json.c b/sway/ipc-json.c
index b9289e25..000137f7 100644
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -120,8 +120,10 @@ static void ipc_json_describe_output(struct sway_container *container,
 			ws = focus->name;
 		}
 	}
-	json_object_object_add(object, "current_workspace",
-			json_object_new_string(ws));
+	if (ws) {
+		json_object_object_add(object, "current_workspace",
+				json_object_new_string(ws));
+	}
 
 	json_object *modes_array = json_object_new_array();
 	struct wlr_output_mode *mode;
```

Second looks like more work and I don't have time to investigate (use-after-free so you'll need ASAN or valgrind):
```
==29667==ERROR: AddressSanitizer: heap-use-after-free on address 0x612000190cc8 at pc 0x00000046f9d8 bp 0x7ffe15dba130 sp 0x7ffe15dba120
WRITE of size 8 at 0x612000190cc8 thread T0
    #0 0x46f9d7 in container_free ../sway/tree/container.c:174
    #1 0x42fdcc in transaction_destroy ../sway/desktop/transaction.c:87
    #2 0x431c64 in transaction_commit ../sway/desktop/transaction.c:291
    #3 0x46e2d5 in arrange_and_commit ../sway/tree/arrange.c:322
    #4 0x475c3e in output_layout_handle_change ../sway/tree/layout.c:25
    #5 0x7f8bb23bb9fa in wlr_signal_emit_safe ../util/signal.c:29
    #6 0x7f8bb239e17c in output_layout_reconfigure ../types/wlr_output_layout.c:134
    #7 0x7f8bb239e2f5 in handle_output_destroy ../types/wlr_output_layout.c:160
    #8 0x7f8bb23bb9fa in wlr_signal_emit_safe ../util/signal.c:29
    #9 0x7f8bb23a24c2 in wlr_output_destroy ../types/wlr_output.c:284
    #10 0x7f8bb234dc61 in handle_x11_event ../backend/x11/backend.c:66
    #11 0x7f8bb234dcea in x11_event ../backend/x11/backend.c:84
    #12 0x7f8bb2632f01 in wl_event_loop_dispatch src/event-loop.c:641
    #13 0x7f8bb2631601 in wl_display_run src/wayland-server.c:1260
    #14 0x40bab8 in server_run ../sway/server.c:137
    #15 0x40ab36 in main ../sway/main.c:433
    #16 0x7f8bb1ada18a in __libc_start_main ../csu/libc-start.c:308
    #17 0x408d29 in _start (/opt/wayland/bin/sway+0x408d29)

0x612000190cc8 is located 8 bytes inside of 296-byte region [0x612000190cc0,0x612000190de8)
freed by thread T0 here:
    #0 0x7f8bb4522880 in __interceptor_free (/lib64/libasan.so.5+0xee880)
    #1 0x42f07c in handle_destroy ../sway/desktop/output.c:1273
    #2 0x7f8bb23bb9fa in wlr_signal_emit_safe ../util/signal.c:29
    #3 0x7f8bb23a24c2 in wlr_output_destroy ../types/wlr_output.c:284
    #4 0x7f8bb234dc61 in handle_x11_event ../backend/x11/backend.c:66
    #5 0x7f8bb234dcea in x11_event ../backend/x11/backend.c:84
    #6 0x7f8bb2632f01 in wl_event_loop_dispatch src/event-loop.c:641

previously allocated by thread T0 here:
    #0 0x7f8bb4522e50 in calloc (/lib64/libasan.so.5+0xeee50)
    #1 0x42f298 in handle_new_output ../sway/desktop/output.c:1304
    #2 0x7f8bb23bb9fa in wlr_signal_emit_safe ../util/signal.c:29
    #3 0x7f8bb233dcbf in new_output_reemit ../backend/multi/backend.c:113
    #4 0x7f8bb23bb9fa in wlr_signal_emit_safe ../util/signal.c:29
    #5 0x7f8bb2351dca in wlr_x11_output_create ../backend/x11/output.c:174
    #6 0x7f8bb234e989 in backend_start ../backend/x11/backend.c:173
    #7 0x7f8bb2321d55 in wlr_backend_start ../backend/backend.c:35
    #8 0x7f8bb233d7a0 in multi_backend_start ../backend/multi/backend.c:24
    #9 0x7f8bb2321d55 in wlr_backend_start ../backend/backend.c:35
    #10 0x40ba24 in server_run ../sway/server.c:132
    #11 0x40ab36 in main ../sway/main.c:433
    #12 0x7f8bb1ada18a in __libc_start_main ../csu/libc-start.c:308
```

martinetd:
Ah, actually the timerfd part of swaywm/wlroots#1090 (mostly) comes from here as well: you use wl_event_loop_add_timer a lot and each of these create a new eventfd, but it's never cleaned up properly.

I think you need to use `wl_event_source_remove(transaction->timer)` when you're done with it.

That adds up pretty quickly, explains why I ran out of fd in half a day :)

RyanDwyer:
@martinetd The output disconnect crash is fixed, and the timer source is now removed.

For the IPC response, it seems weird that an output had no workspace. Can you describe how you triggered that behaviour?

martinetd:
ACK on timer fix, that works well, we leak much less FD (I'll have fun hunting the dozen of sources we're leaking all around later, but at least it's not something happening all the time like transactions :P)

For the output crash your fix makes my IPC crash just a bit earlier so my fix is indeed no longer valid, I'm now getting a crash here:
```
==20426==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x00000041871f bp 0x7ffe702a8ab0 sp 0x7ffe702a8a60 T0)
==20426==The signal is caused by a READ memory access.
==20426==Hint: address points to the zero page.
    #0 0x41871e in ipc_json_describe_output ../sway/ipc-json.c:88
    #1 0x4195bc in ipc_json_describe_container ../sway/ipc-json.c:233
    #2 0x419607 in ipc_json_describe_container_recursive ../sway/ipc-json.c:251
    #3 0x41c320 in ipc_event_window ../sway/ipc-server.c:307
    #4 0x4713bc in container_destroy_noreaping ../sway/tree/container.c:329
    #5 0x47186e in container_destroy ../sway/tree/container.c:406
    #6 0x42f003 in damage_handle_destroy ../sway/desktop/output.c:1261
    #7 0x7f9a4bd709fa in wlr_signal_emit_safe ../util/signal.c:29
    #8 0x7f9a4bd51451 in wlr_output_damage_destroy ../types/wlr_output_damage.c:90
    #9 0x7f9a4bd50f32 in output_handle_destroy ../types/wlr_output_damage.c:13
    #10 0x7f9a4bd709fa in wlr_signal_emit_safe ../util/signal.c:29
    #11 0x7f9a4bd574c2 in wlr_output_destroy ../types/wlr_output.c:284
    #12 0x7f9a4bd02c61 in handle_x11_event ../backend/x11/backend.c:66
    #13 0x7f9a4bd02cea in x11_event ../backend/x11/backend.c:84
    #14 0x7f9a4bfe7f01 in wl_event_loop_dispatch src/event-loop.c:641
    #15 0x7f9a4bfe6601 in wl_display_run src/wayland-server.c:1260
    #16 0x40bb25 in server_run ../sway/server.c:141
    #17 0x40ab36 in main ../sway/main.c:433
    #18 0x7f9a4b48f18a in __libc_start_main ../csu/libc-start.c:308
```

I'm just starting sway with WLR_X11_OUTPUTS=2 sway and closing one of the two outputs window once it's ready. Doing things in it before doesn't seem to change much here.
I get the very same crash with WLR_WL_OUTPUTS=2 and starting on wayland if that's easier for you to play with.

martinetd:
Looks good, way better patch than mine :P

I'm getting one more use-after-free that needs patching wl_list_remove in for asan (so you could see it with valgrind), but it looks fairly unrelated I think it was here before. (output not cleaning its layer shells on output destroy)

I'll fix it and keep reporting later

emersion:
>Okay, that one did exist before and fixed (separate PR inc), but unfortunately some more with "transaction" in the backtrace so leaving this one to you...

See https://github.com/swaywm/sway/issues/1859

martinetd:
>> Okay, that one did exist before and fixed (separate PR inc), but unfortunately some more with "transaction" in the backtrace so leaving this one to you...
>
> See #1859

Fair enough, I guess these are probably the same with just a sprinkle of atomicity. I was starting to feel sorry for @RyanDwyer anyway :P
As far as runtime goes and code itself I'm happy with the PR.

sway/desktop/transaction.c
```diff
@@ -0,0 +1,398 @@
...
+ * If enabled, sway will always wait for the transaction timeout before
+ * applying it, rather than applying it when the views are ready. This allows us
+ * to observe the rendered state while a transaction is in progress.
+ */
+#define TRANSACTION_DEBUG false
+
+struct sway_transaction {
+	struct wl_event_source *timer;
+	list_t *instructions;   // struct sway_transaction_instruction *
+	list_t *damage;         // struct wlr_box *
```
emersion: The preferred datatype for a list of boxes is `pixman_region32_t`. It will merge boxes that overlap.

sway/desktop/transaction.c
```diff
@@ -0,0 +1,398 @@
...
+}
+
+static void save_view_buffer(struct sway_view *view,
+		struct sway_transaction_instruction *instruction) {
+	if (!sway_assert(instruction->saved_buffer == NULL,
+				"Didn't expect instruction to have a saved buffer already")) {
+		remove_saved_view_buffer(instruction);
+	}
+	if (view->surface && wlr_surface_has_buffer(view->surface)) {
+		wlr_buffer_ref(view->surface->buffer);
```
emersion: What do we want to do with subsurfaces and popups?

sway/desktop/transaction.c
```diff
@@ -0,0 +1,398 @@
...
+		instruction->saved_buffer = view->surface->buffer;
```
emersion: The preferred idiom when referencing a buffer is:

```c
instruction->saved_buffer = wlr_buffer_ref(view->surface->buffer);
```

sway/desktop/transaction.c
```diff
@@ -0,0 +1,398 @@
...
+		if (instruction->container == container) {
+			return true;
+		}
+	}
+	return false;
+}
+
+void transaction_add_container(struct sway_transaction *transaction,
+		struct sway_container *container) {
+	if (transaction_has_container(transaction, container)) {
```
emersion: What if the container has already been added to the transaction, but its state has changed since last time?

sway/desktop/transaction.c
```diff
@@ -0,0 +1,398 @@
...
+		clock_gettime(CLOCK_MONOTONIC, &now);
+		struct timespec *create = &transaction->create_time;
+		struct timespec *commit = &transaction->commit_time;
+		float ms_arranging = (commit->tv_sec - create->tv_sec) * 1000 +
+			(commit->tv_nsec - create->tv_nsec) / 1000000.0;
+		float ms_waiting = (now.tv_sec - commit->tv_sec) * 1000 +
+			(now.tv_nsec - commit->tv_nsec) / 1000000.0;
+		float ms_total = ms_arranging + ms_waiting;
+		wlr_log(L_DEBUG, "Transaction %p: %.1fms arranging, %.1fms waiting, "
+				"%.1fms total (%.1f frames if 60hz)", transaction,
```
emersion: Nitpick: Hz with a capital H :)

sway/desktop/transaction.c
```diff
@@ -0,0 +1,398 @@
...
+				ms_arranging, ms_waiting, ms_total, ms_total / (1000 / 60));
+	}
+	int i;
```
emersion: Style: not a fan of using the same variable for unrelated things.

sway/desktop/transaction.c
emersion: Do we even need to store damage? Can we just damage all containers in the transaction when applied?

sway/desktop/transaction.c
```diff
@@ -0,0 +1,398 @@
...
+#endif
+	}
+}
+
+void transaction_notify_view_ready(struct sway_view *view, uint32_t serial) {
+	for (int i = 0; i < view->swayc->instructions->length; ++i) {
+		struct sway_transaction_instruction *instruction =
+			view->swayc->instructions->items[i];
+		if (instruction->serial == serial && !instruction->ready) {
+			set_instruction_ready(instruction);
```
emersion: A client can receive configure n and n+1 and only ACK n+1. That means we should `set_instruction_ready` for all instructions that have a serial <= ACKed serial.

sway/desktop/transaction.c
```diff
@@ -0,0 +1,398 @@
...
+		}
+	}
+}
+
+void transaction_notify_view_ready_by_size(struct sway_view *view,
+		int width, int height) {
+	for (int i = 0; i < view->swayc->instructions->length; ++i) {
+		struct sway_transaction_instruction *instruction =
+			view->swayc->instructions->items[i];
+		if (!instruction->ready && instruction->state.view_width == width &&
```
emersion: What if a client "skips configures", ie. receives one configure with size WxH, and then immediately receives another configure with size W'xH', and finally decides to render one frame with size W'xH'?

sway/desktop/xdg_shell.c
```diff
@@ -152,10 +149,6 @@ static void destroy(struct sway_view *view) {
...
 		return;
 	}
-	wl_list_remove(&xdg_shell_view->destroy.link);
```
emersion: Hmm, why moving this in `handle_destroy`? This seems more confusing.

sway/desktop/xwayland.c
```diff
@@ -250,22 +243,27 @@ static void handle_commit(struct wl_listener *listener, void *data) {
...
+
+	if (!sway_assert(view->surface, "Cannot unmap unmapped view")) {
+		return;
+	}
+
+	view_unmap(view);
+
 	wl_list_remove(&xwayland_view->commit.link);
-	view_unmap(&xwayland_view->view);
+	view->surface = NULL;
```
emersion: Can't this be done in `view_unmap`?

sway/tree/container.c
```diff
@@ -280,14 +286,47 @@ static struct sway_container *container_output_destroy(
...
+
 	return &root_container;
 }
 
-static void container_root_finish(struct sway_container *con) {
-	wlr_log(L_ERROR, "TODO: destroy the root container");
+/**
+ * Implement the actual destroy logic, without reaping.
+ */
+struct sway_container *container_destroy_noreaping(struct sway_container *con) {
```
emersion: This can be `static`

sway/desktop/transaction.c
RyanDwyer: That's a good question. If the user has a popup open and presses a keybind then we should execute it (eg. screenshot), but maybe we should either reposition or close popups when we apply a transaction. I guess with subsurfaces we'd have to play on the safe side and keep them open.

sway/desktop/transaction.c
RyanDwyer: Theoretically it would configure and apply the first state which is bad, but this doesn't happen in practice. To make this happen you would have to make a transaction the manual way, add the container to it, modify something in the container, then add the container again. You should always use do them in the order: modify tree, add to transaction, commit transaction.

sway/desktop/transaction.c
RyanDwyer: Thinking about this more, a much more simple option would be to mark skipped instructions as ready as you suggested, but not apply any transactions at all unless the transaction queue can be completely processed.

martinetd: Possibly far fetched, but would it be possible to have transaction 1 with views A & B, transaction 2 with views B & C.
With what you just said, if A acks 1st configure, B acks 2nd configure directly, when C is done we'll redraw B & C but not A?

I like your previous idea (mark first transaction as waiting-for-2nd, then apply both at once when 2nd is ready) better

RyanDwyer: > if A acks 1st configure, B acks 2nd configure directly, when C is done we'll redraw B & C but not A

When would you redraw A? (or more correctly, when would you apply A's state?)

I've changed the implementation to one which waits for the entire queue to be ready before applying them. This is easier to implement than trying to group them based on dependencies, and should have the same result as long as there's 2 or less transactions in progress at the same time. And even if you have 3 transactions, you'd have to have 2 of them in separate parts of the tree before the dependency-group option has a benefit over the wait-for-full-queue option.

sway/desktop/transaction.c
RyanDwyer: See the comment thread above this one.

sway/desktop/transaction.c
martinetd: Yeah, I think that'll work. It's less subtle than what was suggested earlier but it should be OK in practice.

common/list.c
```diff
@@ -62,6 +62,13 @@ void list_cat(list_t *list, list_t *source) {
 	}
 }
 
+void list_empty(list_t *list) {
```
martinetd: I'd be much less subtle and just set list->length to 0, list_add does reallocs but list_del never bothers with reducing capacity either so meh.

sway/tree/container.c
```diff
@@ -132,42 +133,67 @@ struct sway_container *container_create(enum sway_container_type type) {
...
+		return;
 	}
-
+	free(cont->name);
 	wlr_texture_destroy(cont->title_focused);
 	wlr_texture_destroy(cont->title_focused_inactive);
 	wlr_texture_destroy(cont->title_unfocused);
 	wlr_texture_destroy(cont->title_urgent);
 
+	for (int i = 0; i < server.destroying_containers->length; ++i) {
```
martinetd: I didn't realize we allowed that syntax, but there are plenty around so I guess it's ok :)

sway/desktop/transaction.c
emersion: What if a client receives a configure with size WxH, and commits a W'xH' surface?

sway/desktop/xdg_shell.c
emersion: Functions named `destroy` should free. This is a convention we use everywhere. Functions that don't free should have a different name.

common/list.c
```diff
@@ -62,6 +62,13 @@ void list_cat(list_t *list, list_t *source) {
...
+	list->capacity = 10;
+	list->length = 0;
+	free(list->items);
+	list->items = malloc(sizeof(void*) * list->capacity);
```
emersion: Style: missing space between `void` and `*`

emersion:
>Do we even need to store damage? Can we just damage all containers in the transaction when applied?

My bad. I failed to see that you ACKed this one.

sway/desktop/output.c
```diff
@@ -341,64 +343,105 @@ static void render_view_surfaces(struct sway_view *view,
...
+	struct wlr_output *wlr_output = output->wlr_output;
+
+	struct wlr_texture *texture = transaction_get_texture(view);
+	if (!texture) {
+		return;
+	}
+	struct wlr_box box = {
+		.x = view->swayc->current.view_x - output->swayc->current.swayc_x,
+		.y = view->swayc->current.view_y - output->swayc->current.swayc_y,
+		.width = view->swayc->current.view_width,
```
emersion: Using the view width/height will render incorrectly if the surface size is different. This happens with unresizable windows for instance. We should crop surfaces instead.

emersion:
Thinking again about subsurfaces: this is tricky because right now subsurfaces will be hidden while applying a transaction. Subsurfaces are part of the view content (ie. most of the view can be subsurfaces). You can try `weston-subsurfaces` for instance.

sway/desktop/transaction.c
RyanDwyer: Done.

emersion:
After discussion on IRC, @RyanDwyer convinced me the destroy/free wording is a good idea.

```
11:29 <ryan-au> Let's talk about terms like destroy and free
11:29 <emersion> ryan-au: also wanted to say it's very cool how atomic layout simplifies damage tracking
11:30 <emersion> ryan-au: https://github.com/swaywm/wlroots/blob/master/CONTRIBUTING.md#constructiondestruction-functions
11:30 <ryan-au> Yeah, I think we need to extend the naming convention
11:30 <ryan-au> I like the destroy term, and used it for transactions and even in some non-sway projects of mine
11:31 <ryan-au> But the problem is destroying is now a two step process due to transactions
11:31 <emersion> just find another word
11:32 <ryan-au> The only other word that comes to mind at the moment is remove
11:33 <emersion> finalize?
11:33 <emersion> prepare_destroy?
11:33 <ryan-au> Finalise can be used in the context of setting up
11:33 <emersion> true
11:34 <ryan-au> ok, so are you happen if I rename container_destroy to container_prepare_destroy, and container_free to container_destroy?
11:34 <ryan-au> *happy
11:34 <emersion> yeah, works for me
11:34 <ryan-au> It means any time you want to destroy a container you have to call container_prepare_destroy instead of container_destroy
11:35 <emersion> hmm
11:35 <ryan-au> Also container_workspace_destroy would become container_workspace_prepare_destroy
11:36 <emersion> i don't like it, but i don't like the current name either
11:36 <emersion> neither*
11:39 <ryan-au> If you consider the *_free function to be a component of destroying, then using the destroy+free terms makes sense
11:40 <ryan-au> Any time you call *_destroy, there's no backing out and it'll eventually get freed
11:40 <emersion> you have a point
11:41 <emersion> okay, i agree with you
```

sway/desktop/transaction.c
RyanDwyer: Then it wouldn't match any instruction in the list, the transaction would time out and then the transaction would apply. I don't think there's any other option. Keep in mind this function is only used by xwayland views.

PR #2080, opened 2018-06-01
Reduce work duplication in keyboard and binding code

### LINKED ISSUE #1844: Bindsyms using Mod5 (ISO_Level3_Shift) fail
Using wlroots & sway builds from today (swaywm/wlroots@12bf39a715b2 and 48c84cd1f).

My relevant environment variables are

      export XKB_DEFAULT_OPTIONS=lv3:caps_switch

And the relevant lines of `.config/sway/sway` are `set $mod Mod5` and `bindsym $mod+Return exec st`.
Pressing keys `CapsLock+Enter`  should launch `st`; it doesn't. The cause is that `keysym_is_modifier` in 
`sway/input/keyboard.c` is overly specific in detecting modifiers; comparing with `xmodmap` output, I find

    shift       Shift_L (0x32),  Shift_R (0x3e)
    lock      
    control     Control_L (0x25),  Control_R (0x69)
    mod1        Alt_L (0x40),  Alt_R (0x6c),  Meta_L (0xcd)
    mod2        Num_Lock (0x4d)
    mod3      
    mod4        Super_L (0x85),  Super_R (0x86),  Super_L (0xce),  Hyper_L (0xcf)
    mod5        ISO_Level3_Shift (0x5c),  Mode_switch (0xcb)

so that `keysym_is_modifier` is most likely missing `XKB_KEY_Num_Lock`, `XKB_KEY_Mode_switch`, and `XKB_KEY_ISO_Level3_Shift`. (Now that I have added these cases to my local version of sway, `CapsLock+Enter` does launch `st`.)

It would be nice if sway would identify which keysyms generate modifiers from the xkb API instead of using using a hardcoded list which only works with default keyboards.

frsfnrrg:
This pull request first simplifies the bindcode and bindsym generation code, 
so that the binding generated are identical in form up to the ids used to identify
the set of pressed keys; as most of the code between `cmd_bindsym` and `cmd_bindcode`
is shared, the two functions have been merged.

(Keycode numbering is briefly adjusted to match the XKB numbering scheme, like i3 seems to. (?))

The next large change is a rewrite of the shortcut handling, which should preserve the same behavior
except in edge cases. A single shortcut model is used for keycodes, raw keysyms, and translated keysyms. Release flag shortcuts are now identified on the preceding keypress, triggered if the next keyrelease breaks the shortcut condition, and kept if the next key event doesn't change shortcut state. The hardcoded list of modifier keys has been replaced by a retroactive test.

This pull request closes  #1844 . It interferes with PR #2064, as it uses a slightly different solution to solve the same issue (#2020).

I have not yet, but probably should, seen what happens with dead, compose, or latch keys. `bindsym --release Control_L` does work, and I haven't yet discovered any bugs with my existing setup.

### DIFF
--- a/include/sway/input/keyboard.h
+++ b/include/sway/input/keyboard.h
@@ -6,7 +6,20 @@
 #define SWAY_KEYBOARD_PRESSED_KEYS_CAP 32
 
 struct sway_shortcut_state {
+	/**
+	 * A list of pressed key ids (either keysyms or keycodes),
+	 * including duplicates when different keycodes produce the same key id.
+	 *
+	 * Each key id is associated with the keycode (in `pressed_keycodes`)
+	 * whose press generated it, so that the key id can be removed on
+	 * keycode release without recalculating the transient link between
+	 * keycode and key id at the time of the key press.
+	 */
 	uint32_t pressed_keys[SWAY_KEYBOARD_PRESSED_KEYS_CAP];
+	/**
+	 * The list of keycodes associated to currently pressed key ids,
+	 * including duplicates when a keycode generates multiple key ids.
+	 */
 	uint32_t pressed_keycodes[SWAY_KEYBOARD_PRESSED_KEYS_CAP];
 	int last_key_index;
 };
--- a/include/sway/input/keyboard.h
+++ b/include/sway/input/keyboard.h
@@ -3,7 +3,13 @@
 
 #include "sway/input/seat.h"
 
-#define SWAY_KEYBOARD_PRESSED_KEYSYMS_CAP 32
+#define SWAY_KEYBOARD_PRESSED_KEYS_CAP 32
+
+struct sway_shortcut_state {
+	uint32_t pressed_keys[SWAY_KEYBOARD_PRESSED_KEYS_CAP];
+	uint32_t pressed_keycodes[SWAY_KEYBOARD_PRESSED_KEYS_CAP];
+	int last_key_index;
+};
 
 struct sway_keyboard {
 	struct sway_seat_device *seat_device;
@@ -13,11 +19,11 @@ struct sway_keyboard {
 	struct wl_listener keyboard_key;
 	struct wl_listener keyboard_modifiers;
 
-	xkb_keysym_t pressed_keysyms_translated[SWAY_KEYBOARD_PRESSED_KEYSYMS_CAP];
-	uint32_t modifiers_translated;
-
-	xkb_keysym_t pressed_keysyms_raw[SWAY_KEYBOARD_PRESSED_KEYSYMS_CAP];
-	uint32_t modifiers_raw;
+	struct sway_shortcut_state state_keysyms_translated;
+	struct sway_shortcut_state state_keysyms_raw;
+	struct sway_shortcut_state state_keycodes;
+	struct sway_binding *held_binding;
+	uint32_t last_modifiers;
 };
 
 struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -69,8 +69,9 @@ bool binding_key_compare(struct sway_binding *binding_a,
 	return true;
 }
 
-static struct cmd_results * cmd_bindsym_or_bindcode(int argc, char **argv, bool bindcode) {
-	const char* bindtype = bindcode ? "bindcode" : "bindsym";
+static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
+		bool bindcode) {
+	const char *bindtype = bindcode ? "bindcode" : "bindsym";
 
 	struct cmd_results *error = NULL;
 	if ((error = checkarg(argc, bindtype, EXPECTED_MORE_THAN, 1))) {
@@ -149,7 +150,7 @@ static struct cmd_results * cmd_bindsym_or_bindcode(int argc, char **argv, bool 
 				return ret;
 			}
 		}
-		uint32_t *key = calloc(1, sizeof(xkb_keysym_t));
+		uint32_t *key = calloc(1, sizeof(uint32_t));
 		if (!key) {
 			free_sway_binding(binding);
 			free_flat_list(split);
@@ -158,9 +159,9 @@ static struct cmd_results * cmd_bindsym_or_bindcode(int argc, char **argv, bool 
 		}
 
 		if (bindcode) {
-			*key = (uint32_t) keycode;
+			*key = (uint32_t)keycode;
 		} else {
-			*key = (uint32_t) keysym;
+			*key = (uint32_t)keysym;
 		}
 
 		list_add(binding->keys, key);
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -158,7 +158,7 @@ static struct cmd_results * cmd_bindsym_or_bindcode(int argc, char **argv, bool 
 		}
 
 		if (bindcode) {
-			*key = (uint32_t) (keycode - 8);
+			*key = (uint32_t) keycode;
 		} else {
 			*key = (uint32_t) keysym;
 		}
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -69,15 +69,17 @@ bool binding_key_compare(struct sway_binding *binding_a,
 	return true;
 }
 
-struct cmd_results *cmd_bindsym(int argc, char **argv) {
+static struct cmd_results * cmd_bindsym_or_bindcode(int argc, char **argv, bool bindcode) {
+	const char* bindtype = bindcode ? "bindcode" : "bindsym";
+
 	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "bindsym", EXPECTED_MORE_THAN, 1))) {
+	if ((error = checkarg(argc, bindtype, EXPECTED_MORE_THAN, 1))) {
 		return error;
 	}
 
 	struct sway_binding *binding = calloc(1, sizeof(struct sway_binding));
 	if (!binding) {
-		return cmd_results_new(CMD_FAILURE, "bindsym",
+		return cmd_results_new(CMD_FAILURE, bindtype,
 				"Unable to allocate binding");
 	}
 	binding->keys = create_list();
@@ -100,9 +102,9 @@ struct cmd_results *cmd_bindsym(int argc, char **argv) {
 	}
 	if (argc < 2) {
 		free_sway_binding(binding);
-		return cmd_results_new(CMD_FAILURE, "bindsym",
-			"Invalid bindsym command "
-			"(expected at least 2 non-option arguments, got %d)", argc);
+		return cmd_results_new(CMD_FAILURE, bindtype,
+			"Invalid %s command "
+			"(expected at least 2 non-option arguments, got %d)", bindtype, argc);
 	}
 
 	binding->command = join_args(argv + 1, argc - 1);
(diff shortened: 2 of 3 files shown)


sway/commands/bind.c
```diff
@@ -69,22 +69,24 @@ bool binding_key_compare(struct sway_binding *binding_a,
 	return true;
 }
 
-struct cmd_results *cmd_bindsym(int argc, char **argv) {
+static struct cmd_results * cmd_bindsym_or_bindcode(int argc, char **argv, bool bindcode) {
```
ascent12: Incorrect *

sway/commands/bind.c
```diff
@@ -69,22 +69,24 @@ bool binding_key_compare(struct sway_binding *binding_a,
...
+	const char* bindtype = bindcode ? "bindcode" : "bindsym";
```
ascent12: Incorrect *

sway/commands/bind.c
SirCmpwn: Clarification: the correct style is `static struct cmd_results *cmd_bindsym...`

sway/commands/bind.c
SirCmpwn: Clarification: the correct style is `const char *bindtype...`

sway/commands/bind.c
```diff
@@ -69,22 +69,24 @@ bool binding_key_compare(struct sway_binding *binding_a,
...
 	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "bindsym", EXPECTED_MORE_THAN, 1))) {
+	if ((error = checkarg(argc, bindtype, EXPECTED_MORE_THAN, 1))) {
 		return error;
 	}
 
 	struct sway_binding *binding = calloc(1, sizeof(struct sway_binding));
 	if (!binding) {
-		return cmd_results_new(CMD_FAILURE, "bindsym",
+		return cmd_results_new(CMD_FAILURE, bindtype,
```
SirCmpwn: Side note, you can also get this from `argv[-1]`

sway/input/keyboard.c
```diff
@@ -9,88 +9,93 @@
...
-	for (size_t i = 0; i < SWAY_KEYBOARD_PRESSED_KEYSYMS_CAP; ++i) {
-		if (pressed_keysyms[i] != XKB_KEY_NoSymbol) {
-			++n;
+/**
+ * Update the shortcut model state in response to new input
+ */
+static void update_shortcut_model(struct sway_shortcut_state* state,
+				  struct wlr_event_keyboard_key * event,
+				  uint32_t new_key,
+				  bool last_key_was_a_modifier) {
```
SirCmpwn: Style: only indent this two lines and wrap parameters after 80 columns

SirCmpwn:
This is looking good, nice work

sway/commands/bind.c
```diff
@@ -115,124 +117,63 @@ struct cmd_results *cmd_bindsym(int argc, char **argv) {
...
+			if (!keysym) {
+				struct cmd_results *ret = cmd_results_new(CMD_INVALID, "bindsym",
+						"Unknown key '%s'", (char *)split->items[i]);
+				free_sway_binding(binding);
+				free_flat_list(split);
+				return ret;
+			}
 		}
-		xkb_keysym_t *key = calloc(1, sizeof(xkb_keysym_t));
+		uint32_t *key = calloc(1, sizeof(xkb_keysym_t));
```
ascent12: Wrong type in sizeof (even if they happen to be the same size).

sway/commands/bind.c
```diff
@@ -115,124 +117,63 @@ struct cmd_results *cmd_bindsym(int argc, char **argv) {
...
-	while (argc > 0) {
-		if (strcmp("--release", argv[0]) == 0) {
-			binding->release = true;
-		} else if (strcmp("--locked", argv[0]) == 0) {
-			binding->locked = true;
+		if (bindcode) {
+			*key = (uint32_t) keycode;
 		} else {
-			break;
+			*key = (uint32_t) keysym;
```
ascent12: No space after casts.

sway/input/keyboard.c
```diff
@@ -9,88 +9,93 @@
...
-
-static size_t pressed_keysyms_length(xkb_keysym_t *pressed_keysyms) {
-	size_t n = 0;
-	for (size_t i = 0; i < SWAY_KEYBOARD_PRESSED_KEYSYMS_CAP; ++i) {
-		if (pressed_keysyms[i] != XKB_KEY_NoSymbol) {
-			++n;
+/**
+ * Update the shortcut model state in response to new input
+ */
+static void update_shortcut_model(struct sway_shortcut_state* state,
```
ascent12: Incorrect *s and incorrect indentation.

sway/input/keyboard.c
```diff
@@ -9,88 +9,93 @@
...
-		xkb_keysym_t keysym) {
-	ssize_t i = pressed_keysyms_index(pressed_keysyms, keysym);
-	if (i >= 0) {
-		pressed_keysyms[i] = XKB_KEY_NoSymbol;
+/**
+ *
+ * Returns a binding which matches the shortcut model state (ignoring the
+ * `release` flag).
+ */
+static struct sway_binding* check_shortcut_model(
```
ascent12: Incorrect *s

sway/input/keyboard.c
```diff
@@ -9,88 +9,93 @@
...
 			continue;
 		}
-		if (state == WLR_KEY_PRESSED) {
-			pressed_keysyms_add(pressed_keysyms, keysyms[i]);
-		} else { // WLR_KEY_RELEASED
-			pressed_keysyms_remove(pressed_keysyms, keysyms[i]);
+
+		bool match = true;
+		for (int j = 0; j < binding->keys->length; ++j) {
+			uint32_t key = *(uint32_t*)binding->keys->items[j];
```
ascent12: `(uint32_t *)`

sway/input/keyboard.c
```diff
@@ -339,61 +193,96 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
...
+
+	// Update shortcut models
+	update_shortcut_model(&keyboard->state_keycodes, event,
+			      (uint32_t)keycode, last_key_was_a_modifier);
+	for (size_t i=0;i<translated_keysyms_len;i++) {
+		update_shortcut_model(&keyboard->state_keysyms_translated,
+				event, (uint32_t)translated_keysyms[i],
+				last_key_was_a_modifier);
+	}
+	for (size_t i=0;i<raw_keysyms_len;i++) {
```
ascent12: Spaces around binary operators.

sway/input/keyboard.c
```diff
@@ -339,61 +193,96 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
...
+	struct wlr_input_device *device =
+		keyboard->seat_device->input_device->wlr_device;
+	uint32_t code_modifiers = wlr_keyboard_get_modifiers(device->keyboard);
+
+	bool last_key_was_a_modifier = code_modifiers != keyboard->last_modifiers;
+	keyboard->last_modifiers = code_modifiers;
+
+	// Update shortcut models
+	update_shortcut_model(&keyboard->state_keycodes, event,
+			      (uint32_t)keycode, last_key_was_a_modifier);
```
ascent12: Incorrect indentation.

include/sway/input/keyboard.h
```diff
@@ -3,7 +3,13 @@
 
 #include "sway/input/seat.h"
 
-#define SWAY_KEYBOARD_PRESSED_KEYSYMS_CAP 32
+#define SWAY_KEYBOARD_PRESSED_KEYS_CAP 32
+
+struct sway_shortcut_state {
+	uint32_t pressed_keys[SWAY_KEYBOARD_PRESSED_KEYS_CAP];
+	uint32_t pressed_keycodes[SWAY_KEYBOARD_PRESSED_KEYS_CAP];
```
emersion: Why are the keycodes duplicated in each state?

sway/input/keyboard.c
emersion: Note that some keycodes can yield one modifier keysym *and* on enon-modifier keysym.

sway/commands/bind.c
frsfnrrg: I'd prefer not to change this as it isn't immediately clear that `argv` has a valid value at `-1`; at least, no other command uses this trick. Also, the function changes the values of `argc` and `argv`, so this would only change the line which sets `bindtype`, making it more difficult to see the possible values.

include/sway/input/keyboard.h
frsfnrrg: They are necessary to keep track of the keysyms corresponding to a given keycode. (This is cleaner than trying to reverse the xkb API to find what the keysym would have been at the time the corresponding keycode was pressed; consider the shortcut `Ctrl+b+at` in which `Shift` is pressed halfway through.)

They are somewhat duplicated (up to translation by 8) for the shortcut model tracking keycodes, although even in this case, after a modifier key is erased, `pressed_keys[i]` might be `0` while `pressed_keycodes[i]` stores the event keycode.

emersion: Oh, right, that makes a lot more sense. Can you add a comment explaining the mapping between these two guys?

sway/input/keyboard.c
```diff
@@ -9,88 +9,92 @@
...
-
-static size_t pressed_keysyms_length(xkb_keysym_t *pressed_keysyms) {
-	size_t n = 0;
-	for (size_t i = 0; i < SWAY_KEYBOARD_PRESSED_KEYSYMS_CAP; ++i) {
-		if (pressed_keysyms[i] != XKB_KEY_NoSymbol) {
-			++n;
+/**
+ * Update the shortcut model state in response to new input
+ */
+static void update_shortcut_model(struct sway_shortcut_state *state,
```
emersion: I don't really like the use of "model", which makes it sound more complicated than it is. What about something like `update_shortcut_state`?

sway/input/keyboard.c
frsfnrrg: Technically there is no such thing as a "modifier keysym"; modifiers are associated to the current `xkb_state`, and the set of produced keysyms are associated to the current `xkb_state`. (Granted, the rule files governing the keymap sometimes link keysyms to modifiers; but in some cases (altwin) it links keycodes directly to modifiers. These linkages may also depend on other properties of the `xkb_state`; it's telling that the xkbcommon API never provides such an association. The mapping implied by `xmodmap` is, like its assumption that a keycode can produce no more than one keysym, incorrect in the general case.)

sway/input/keyboard.c
frsfnrrg: That might work, if not for the existence of the function `check_shortcut_model` which also acts on the shortcut state.  Then again, the shortcut model is the rule set relating key input, binding output, and the state, so `check_shortcut_model` isn't strictly correct either.

The variations `struct sway_shortcut_model_state`,  `update_shortcut_model_state`, `get_bindings_matching_shortcut_model_state` are more precise, but verbose. 

I'll think about this again in eight hours.

frsfnrrg: `update_shortcut_model` stays, because it's the most apt description I could find. The shortcut model is a reduction of the full key history; and as given by the function comment, `update_shortcut_model` does precisely that. I did rename `check_shortcut_model` to `get_active_binding`.

frsfnrrg: Done.

sway/input/keyboard.c
```diff
@@ -9,88 +9,92 @@
...
+		struct sway_binding *binding = bindings->items[i];
 
-static void pressed_keysyms_update(xkb_keysym_t *pressed_keysyms,
-		const xkb_keysym_t *keysyms, size_t keysyms_len,
-		enum wlr_key_state state) {
-	for (size_t i = 0; i < keysyms_len; ++i) {
-		if (keysym_is_modifier(keysyms[i])) {
+		if (modifiers ^ binding->modifiers ||
+			npressed_keys != binding->keys->length ||
+			locked > binding->locked) {
```
emersion: Style: we add one more level of indentation to multiline conditions, so that it's easy to distinguish between the condition and the block.

sway/input/keyboard.c
```diff
@@ -9,88 +9,92 @@
...
+static void update_shortcut_state(struct sway_shortcut_state *state,
+		struct wlr_event_keyboard_key *event, uint32_t new_key,
+		bool last_key_was_a_modifier) {
+	if (event->state == WLR_KEY_PRESSED) {
+		if (last_key_was_a_modifier && state->last_key_index >= 0) {
+			// Last pressed key before this one was a modifier. We nullify
+			// the key id but not the keycode (as that is used for erasure
+			// on release)
+			state->pressed_keys[state->last_key_index] = 0;
+			state->last_key_index = -1;
```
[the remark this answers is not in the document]
frsfnrrg: This section erases modifier keys once they have been identified as changing the modifier state.
As this is discovered on the keypress after the modifier key, `last_key_index` is stored so that the modifier key can be identified.

As for `last_key_index` is reset: it may be that more than 32 keys are pressed, and then the value
might otherwise persist and cause an incorrect removal from the pressed key list.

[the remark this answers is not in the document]
emersion: All right, thanks for the explanation

emersion:
Tested with

```patch
diff --git a/sway/input/keyboard.c b/sway/input/keyboard.c
index 420cefa6..6f58fd20 100644
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -41,6 +41,14 @@ static void update_shortcut_state(struct sway_shortcut_state *state,
                        }
                }
        }
+
+       for (size_t i = 0; i < SWAY_KEYBOARD_PRESSED_KEYS_CAP; ++i) {
+               if (state->pressed_keys[i]) {
+                       char buf[256];
+                       xkb_keysym_get_name(state->pressed_keys[i], buf, 256);
+                       wlr_log(L_DEBUG, "is pressed: %d %s", state->pressed_keys[i], buf);
+               }
+       }
 }
 
 /**
@@ -212,6 +220,8 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
        bool last_key_was_a_modifier = code_modifiers != keyboard->last_modifiers;
        keyboard->last_modifiers = code_modifiers;
 
+       wlr_log(L_DEBUG, "");
+
        // Update shortcut model state
        update_shortcut_state(&keyboard->state_keycodes, event,
                        (uint32_t)keycode, last_key_was_a_modifier);
```

emersion:
Thanks, nice work!

PR #2084, opened 2018-06-01
Runtime variable expansion

### LINKED ISSUE #548: Do variable expansion at runtime
We should do variable expansion at runtime, rather than when reading the config file.

```
  swaybot   [sway] SirCmpwn pushed 2 new commits to master: https://git.io/vVf5P
  swaybot   sway/master 59a1c5c Drew DeVault: Update sway(1)
  swaybot   sway/master eacf3d0 Drew DeVault: Loosen restrictions on commands, update docs...
    F1skr   SirCmpwn: I don't think that will work ^
    F1skr   for bindsym
 SirCmpwn   F1skr: why?
    F1skr   for instance $mod is not handled at runtime afaik
    F1skr   variables*
 SirCmpwn   going to leave the docs as-is and make a new issue to do variable expansion at runtime
 SirCmpwn   which would be nice cause you could do things like swaymsg set $mod Mod1 to switch mod keys at runtime, for example
```

One issue is that you wouldn't be able to use variables _as_ commands. We could probably figure that out, though.

Warning: @mikkeloscar says this is complicated.

RedSoxFan:
**This is forked from #2070 so that needs to be merged first**

Fixes #548

This makes it so variables are expanded at runtime. For commands like `bindsym` and `bindcode`, that need an argument (such as `<key combo>`) expanded at config time, there is `char *expand_var(char *var)`.

This also makes it so that `set` can be used at runtime.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -662,7 +662,7 @@ char *do_var_replacement(char *str) {
 			size_t length = strlen(find + 1);
 			strncpy(find, find + 1, length);
 			find[length] = '\0';
-			find += 2;
+			++find;
 			continue;
 		}
 		// Find matching variable
--- a/sway/config.c
+++ b/sway/config.c
@@ -657,6 +657,14 @@ char *do_var_replacement(char *str) {
 				continue;
 			}
 		}
+		// Unescape double $ and move on
+		if (find[1] == '$') {
+			size_t length = strlen(find + 1);
+			strncpy(find, find + 1, length);
+			find[length] = '\0';
+			find += 2;
+			continue;
+		}
 		// Find matching variable
 		for (i = 0; i < config->symbols->length; ++i) {
 			struct sway_variable *var = config->symbols->items[i];
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -112,6 +112,7 @@ static struct cmd_handler handlers[] = {
 	{ "mouse_warping", cmd_mouse_warping },
 	{ "output", cmd_output },
 	{ "seat", cmd_seat },
+	{ "set", cmd_set },
 	{ "show_marks", cmd_show_marks },
 	{ "workspace", cmd_workspace },
 	{ "workspace_auto_back_and_forth", cmd_ws_auto_back_and_forth },
@@ -120,7 +121,6 @@ static struct cmd_handler handlers[] = {
 /* Config-time only commands. Keep alphabetized */
 static struct cmd_handler config_handlers[] = {
 	{ "default_orientation", cmd_default_orientation },
-	{ "set", cmd_set },
 	{ "swaybg_command", cmd_swaybg_command },
 	{ "workspace_layout", cmd_workspace_layout },
 };
@@ -268,6 +268,13 @@ struct cmd_results *execute_command(char *_exec, struct sway_seat *seat) {
 				results = cmd_results_new(CMD_INVALID, cmd, "Unknown/invalid command");
 				free_argv(argc, argv);
 				goto cleanup;
+			}
+
+			// Var replacement, for all but first argument of set
+			for (int i = handler->handle == cmd_set ? 2 : 1; i < argc; ++i) {
+				argv[i] = do_var_replacement(argv[i]);
+				unescape_string(argv[i]);
+				strip_quotes(argv[i]);
 			}
 
 			if (!config->handler_context.using_criteria) {
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -128,12 +128,14 @@ char *workspace_next_name(const char *output_name) {
 		}
 
 		if (strcmp("workspace", cmd) == 0 && name) {
-			wlr_log(L_DEBUG, "Got valid workspace command for target: '%s'", name);
 			char *_target = strdup(name);
+			_target = do_var_replacement(_target);
 			strip_quotes(_target);
 			while (isspace(*_target)) {
 				memmove(_target, _target+1, strlen(_target+1));
 			}
+			wlr_log(L_DEBUG, "Got valid workspace command for target: '%s'",
+					_target);
 
 			// Make sure that the command references an actual workspace
 			// not a command about workspaces


RedSoxFan:
Refactored and ready for review

sway/commands/bind.c
```diff
@@ -110,6 +110,10 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 
 	binding->command = join_args(argv + 1, argc - 1);
 
+	if (config->reading) {
+		argv[0] = expand_var(argv[0]);
```
emersion: Sorry, I still don't get it: why is there a special case for this one? Isn't it expanded anyway before the handler is executed?

emersion: Oh, I see, thanks for the explanation.

emersion:
I get these:

```
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 38 'output * bg $wallpaper fill': Invalid syntax ($wallpaper). (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 188 'client.focused $bgf $bgf $fg $bgf $bgf': Unable to parse border color (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 189 'client.focused_inactive $bga $bga $fgi $bga $bga': Unable to parse border color (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 190 'client.unfocused $bg $bg $fgu $bg $bg': Unable to parse border color (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 202 'statusline $fg': Invalid color definition $fg (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 203 'background $bg': Invalid color definition $bg (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 204 'focused_background $bgf': Invalid color definition $bgf (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 205 'focused_workspace $bgf $bgf $fg': Invalid color definition $bgf (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 206 'active_workspace $bga $bga $fg': Invalid color definition $bga (/home/simon/.config/sway/config)
2018-06-03 23:00:05 - [sway/config.c:601] Error on line 207 'inactive_workspace $bg $bg $fgi': Invalid color definition $bg (/home/simon/.config/sway/config)
```

Config:

```
        set $wallpaper /home/simon/.background
        set $wallpaper_lock /home/simon/.background-lock

        set $bg #000000
        set $bgf #000000
        set $bga #000000
        set $fg #ffffff
        set $fgi #888888
        set $fgu #555555
        set $btf #000000

client.focused $bgf $bgf $fg $bgf $bgf
client.focused_inactive $bga $bga $fgi $bga $bga
client.unfocused $bg $bg $fgu $bg $bg

output * bg $wallpaper fill
```

RedSoxFan:
Ok. I think this is ready for another round of review.

I've assigned each handler a `bool` that determines whether it should wait until runtime to expand variables. I set every command with subcommands to expand a runtime so that the decision is passed on to the subcommands. For identifiers in the block itself, there are `if (config->reading) { //expand }` checks just like for `<key combo>` in `bindsym` and `bindcode`.

Additionally, since starting workspace names are generated off of bindings for `workspace` commands, I have it expand there as well.

emersion:
Hmm, I'm not a fan of this bool, which introduces a lot of complexity. At the same time, I'm not sure there's a simpler solution…

SirCmpwn:
This seems kind of silly. What if we just let you escape $

```
bindsym $mod+Return exec $$term
```

RedSoxFan:
> This seems kind of silly. What if we just let you escape $
> ```
> bindsym $mod+Return exec $$term
> ```

Works for me. I'll update the PR

RedSoxFan:
This has been switched to sircmpwn's suggestion of using `$$`

sway/config.c
```diff
@@ -657,6 +657,14 @@ char *do_var_replacement(char *str) {
 				continue;
 			}
 		}
+		// Unescape double $ and move on
+		if (find[1] == '$') {
+			size_t length = strlen(find + 1);
+			strncpy(find, find + 1, length);
+			find[length] = '\0';
+			find += 2;
```
emersion: Since we just erased one `$`, shouldn't this be `+= 1`?

RedSoxFan: Fixed

emersion:
I've tried doing this:

```
swaymsg set '$term' xterm
swaymsg exec '$$term'
```

But it's not expanding `$$term`. Am I doing something wrong? Can you give a few usage examples?

emersion:
Thanks!

aadcg:
I tried this approach to change colors at runtime. For instance, let this an excerpt from a config file.

```
set $background #ffffff
set $foreground #000000

[...]

bar {
    position top
    colors {
        statusline $$foreground
        background $$background
        inactive_workspace #32323200 #32323200 #5c5c5c
    }
}
```

If my understanding is correct, running `swaymsg set '$background' #000000` would take effect. However, when I (re)load the above mentioned config file I get an error. What am I missing?

RedSoxFan:
@aadcg 

> However, when I (re)load the above mentioned config file I get an error.

If the error is that `$foreground` and `$background` are not valid colors, this is expected. See below for more information

> What am I missing?

The variables are expanded when the command is run. For the bar colors, this is when the config is loaded so unfortunately this won't work for that use case. When the config is loaded, `$$background` will be unescaped to `$background`, but will never be expanded since that is the only time it is processed - which is why you are getting an error. This functionality is only useful for bindings or with IPC commands, which get processed when the event occurs.

As long as you have a new enough version of sway (I'd have to look to see if it made it into 1.4 or requires master), you should be able to use `swaymsg bar bar-0 colors background #000000` though.

aadcg:
> This functionality is only useful for bindings or with IPC commands, which get processed when the event occurs.

Oh, I see. That makes sense.

> [...] you should be able to use swaymsg bar bar-0 colors background #000000 though.

Yes, this is exactly what I was looking for! Many thanks.

tmpm697:
I have this setup that doesn't work as I think variable expansion doesn't apply when we set via `swaymsg`:
```
set $workroom 1
set $workspace 0
bindsym Alt+Shift+1 set $$workroom 1; workspace $$workroom$$workspace
bindsym Alt+Shift+2 set $$workroom 2; workspace $$workroom$$workspace
bindsym $mod+1 workspace $$workroom$ws1; set $$workspace $ws1
bindsym $mod+2 workspace $$workroom$ws2; set $$workspace $ws2
```
Full sample [here](https://github.com/woutdp/dotfiles/blob/master/sway/.config/sway/config#L58-L100) and [here](https://www.reddit.com/r/swaywm/comments/tsvydv/comment/i2wjw6x/?utm_source=share&utm_medium=web2x&context=3)
The above example works as I can change `workroom` and then navigate `workspace` of each `workroom`.
But If I have:
```
bindsym Alt+Shift+1 script1.sh
bindsym Alt+Shift+1 script2.sh
```
and content of `script1.sh` like (same as `script2.sh`):
```
swaymsg set $$workroom 1 \; workspace $$workroom$$workspace
```
With this setup with `script.sh`, I can't navigate to workspaces of `workroom 2` after pressing `Alt+Shift+2`. I believe the expansion doesn't work properly when you put `set $$workroom 2` in a script and call it via `swaymsg`

My question is: why and how to get this work properly via script and swaymsg? cc: @emersion

PR #2087, opened 2018-06-02
Refactor cmd_output to use config_subcommand

RedSoxFan:
**This is forked from #2070 so that needs to be merged first**

@emersion requested that the output command be refactored to use subcommands in #2070. This PR does that

This also fixes setting a solid color background. In master, there are two issues:
1. There is a logic issue that prevents the `solid_color` branch from executing (it is checking against argument before -- the color).
2. `#` needs to be quoted or escaped when included in a shell command (ie swaybg). I decided to go with quoting the argument.

### DIFF
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -1,253 +1,23 @@
-#define _XOPEN_SOURCE 500
-#include <ctype.h>
-#include <libgen.h>
-#include <stdlib.h>
-#include <string.h>
-#include <strings.h>
-#include <unistd.h>
-#include <wordexp.h>
 #include "sway/commands.h"
 #include "sway/config.h"
 #include "list.h"
 #include "log.h"
-#include "stringop.h"
 
-static char *bg_options[] = {
-	"stretch",
-	"center",
-	"fill",
-	"fit",
-	"tile",
+// must be in order for the bsearch
+static struct cmd_handler output_handlers[] = {
+	{ "background", output_cmd_background },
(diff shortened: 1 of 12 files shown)


SirCmpwn:
Needs rebase

RedSoxFan:
Refactored and ready for review

sway/commands/output.c
```diff
@@ -260,40 +30,31 @@ struct cmd_results *cmd_output(int argc, char **argv) {
...
-				"Invalid output subcommand: %s.", command);
+				"Invalid output subcommand: %s.", *argv);
 		}
 
 		if (error != NULL) {
 			goto fail;
 		}
+
+		argc = config->handler_context.leftovers.argc;
+		argv = config->handler_context.leftovers.argv;
```
emersion: If the handler "forgets" to set `leftovers` in some cases, this will become an infinite loop. Is it possible to reset them to NULL before handling and assert they're not NULL after handling?

sway/commands/output/background.c
```diff
@@ -0,0 +1,105 @@
...
+#include <libgen.h>
+#include <strings.h>
+#include <unistd.h>
+#include <wordexp.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "log.h"
+#include "stringop.h"
+
+static char *bg_options[] = {
```
emersion: This can be `const`

emersion:
Thanks, nice work!

PR #2088, opened 2018-06-02
Fix top border when a view is initially floating

### LINKED ISSUE #2085: Windows which are initially floating do not render the top border correctly

RyanDwyer:
Fixes #2085.

Those properties control whether the individual borders are rendered, and only take effect if the view's border is not `B_NONE`. Therefore it's safe to set them all to true. It initialises the values if the view is initially floating, and undoes any `hide_edge_borders` effects if the view was previously tiled.


emersion:
Thanks!

PR #2091, opened 2018-06-02
sway/config: fix use-after-free for end of block

martinetd:
Introduced in 7c810dc344c28d1876c5ee158cb0806289d0f813


emersion:
Thanks!

PR #2094, opened 2018-06-02
Schedule a frame even if damage is empty

emersion:
This fixes clients scheduling a frame by committing without damaging anything like so:

```
callback = wl_surface_frame(surface);
wl_surface_commit(surface);
```


emersion:
Tested by @atomnuker

PR #2095, opened 2018-06-02
Only send frame to visible surfaces when fullscreened

emersion:
This code mirrors `render_output`.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -969,20 +969,30 @@ static void send_frame_done(struct sway_output *output, struct timespec *when) {
 		.when = when,
 	};
 
-	send_frame_done_layer(&data,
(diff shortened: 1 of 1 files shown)


atomnuker:
Works fine, and looks fine code-wise, but I can't seem to find any difference before and after with the wip damage debug PR.

atomnuker:
Been using for 3 days now, no issues to report, should be good to push?

emersion:
>but I can't seem to find any difference before and after with the wip damage debug PR.

Yeah, this only stops the rendering loop for invisible surfaces. Rejecting damage from invisible surfaces will require more work. Since swaybar isn't using frame callbacks, it continues to damage fullscreened outputs.

>Been using for 3 days now, no issues to report, should be good to push?

Aye.

PR #2096, opened 2018-06-02
Fix view title update order

kupospelov:
The first view renders its title text incorrectly in case any font is set in the config file.

The `font` command calls `config_update_font_height(true)`, which resets `config->font_height` to 0 before iterating through child containers, but `config->font_height` is not updated anymore, because at this point there is only the root container with empty `formatted_title`.

When a new view is created after that, it relies on `config->font_height` to render its title, however it gets updated after the title is already rendered.

This commit changes the order, so that a newly created view uses the relevant font height.


emersion:
Thanks!

PR #2100, opened 2018-06-03
Fix initial workspace orientation on a rotated output

RyanDwyer:
When launching on DRM with an output rotated 90 or 270 (ie. portrait), the first workspace on that output would incorrectly have an `L_HORIZ` layout. This is because the workspace was being created before the output's dimensions were assigned, and the workspace's layout is chosen when it is created.

The PR assigns the output's dimensions before creating the workspace.


emersion:
Thanks!

PR #2103, opened 2018-06-04
Focus fixes

RyanDwyer:
This PR fixes three focus related issues.

1. Restore focus when unmapping unmanaged xwayland surfaces. To test:

    * Open a terminal and run i3-input.
    * Cancel the prompt by pressing escape.
    * Focus should be returned to the terminal.

2. Restore focus when unmapping layer shell surfaces. To test:

    * Open a view.
    * Lock the screen with swaylock.
    * Unlock the screen.
    * Focus should be returned to the view.

3. Don't set focus to NULL when clicking a surface which has no container. To test:

    * Launch chromium and configure it so the bookmarks bar is visible and it contains a bookmark.
    * Right click on a bookmark.
    * swaybar should show that the workspace has focus.

There is one thing I'm not sure about, and that's the `wlr_xwayland_surface_is_unmanaged` check in `unmanaged_handle_unmap`. I copied it from `unmanaged_handle_map` but it seems backwards to me. Not only does the check seem to be incorrectly negated, but wouldn't the surface always be unmanaged at that point anyway?

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -79,14 +79,17 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
 	wl_list_remove(&surface->commit.link);
 
 	if (!wlr_xwayland_surface_is_unmanaged(xsurface)) {
-		// Restore focus
 		struct sway_seat *seat = input_manager_current_seat(input_manager);
-		struct sway_container *previous =
-			seat_get_focus_inactive(seat, &root_container);
-		if (previous) {
-			// Hack to get seat to re-focus the return value of get_focus
-			seat_set_focus(seat, previous->parent);
-			seat_set_focus(seat, previous);
+		if (seat->wlr_seat->keyboard_state.focused_surface ==
+				xsurface->surface) {
+			// Restore focus
+			struct sway_container *previous =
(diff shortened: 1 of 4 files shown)


sway/desktop/xwayland.c
```diff
@@ -77,6 +77,18 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
 	desktop_damage_surface(xsurface->surface, xsurface->x, xsurface->y, true);
 	wl_list_remove(&surface->link);
 	wl_list_remove(&surface->commit.link);
+
+	if (!wlr_xwayland_surface_is_unmanaged(xsurface)) {
```
martinetd: Not sure about what unmanaged means in the first place so I might be off the mark, but for layer shell you check if the seat's focus is on the unmapped surface ; this might need a similar check here?
You wouldn't want focus to change because some x11 window depops in another workspace due to a timeout or something.

RyanDwyer: I've added a similar check. To test, I launched a terminal and Firefox side by side. In the terminal I ran `i3-input & sleep 5 && killall i3-input`, then while the sleep was occurring I opened the settings menu of Firefox and waited. After unmapping, the Firefox menu remained open and had keyboard focus.

emersion:
Thanks!

PR #2107, opened 2018-06-05
Remove suspicious unconditional return

tobiasblass:
This return effectively disables the log_kernel function. I don't think this is intended. Git blame suggests this was accidentally commited in 9eecbb5d8a988a0dded57ead1982dd0121071454.


emersion:
Thanks!

PR #2108, opened 2018-06-05
Allow outputs to be re-enabled

### LINKED ISSUE #1887: Output enabling does nothing in DRM backend
There is no change on screen. Output is not seen using `swaymsg -t get_outputs`. Debug logs are uninformative:
```
2018-05-02 08:20:24 - [sway/commands.c:340] Handling command 'output LVDS-1 enable'
2018-05-02 08:20:24 - [sway/commands.c:226] find_handler(output) 0
2018-05-02 08:20:24 - [sway/commands/output.c:291] Config stored for output LVDS-1 (enabled: 1) (-1x-1@-1.000000Hz position -1,-1 scale -1.000000 transform -1) (bg /home/js/media/pic/wallpaper-426081.jpg stretch)
2018-05-02 08:20:24 - [sway/ipc-server.c:657] Added IPC reply to client 64 queue: [ { "success": true } ]
2018-05-02 08:20:24 - [sway/ipc-server.c:356] Client 64 writable
2018-05-02 08:20:24 - [sway/ipc-server.c:192] Client 64 hung up
2018-05-02 08:20:24 - [sway/ipc-server.c:388] IPC Client 64 disconnected
```

I'm not sure about other backends, but it's probably similar there. Reported from #1885

RedSoxFan:
Fixes #1887

When an output is:
- _connected_ - the sway_output stored in a `list_t` in `struct sway_root`
- _disabled_ - the container is destroyed
- _enabled_ - the container is (re-)created
- _disconnected_ - the sway_output is removed from the `list_t` in `struct sway_root` and freed

Additionally, I added disabled outputs to the output of `swaymsg -t get_outputs`.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -1157,6 +1157,10 @@ void output_damage_whole_container(struct sway_output *output,
 	wlr_output_damage_add_box(output->damage, &box);
 }
 
+static int find_output(const void *output1, const void *output2) {
+	return output1 == output2 ? 0 : 1;
+}
+
 static void damage_handle_destroy(struct wl_listener *listener, void *data) {
 	struct sway_output *output =
 		wl_container_of(listener, output, damage_destroy);
@@ -1165,7 +1169,19 @@ static void damage_handle_destroy(struct wl_listener *listener, void *data) {
 
 static void handle_destroy(struct wl_listener *listener, void *data) {
 	struct sway_output *output = wl_container_of(listener, output, destroy);
-	container_destroy(output->swayc);
+	if (output->swayc) {
+		container_destroy(output->swayc);
+	}
+	int index = list_seq_find(root_container.sway_root->outputs, find_output,
+			output);
+	if (index >= 0 && index < root_container.sway_root->outputs->length) {
+		wlr_log(L_DEBUG, "Removing %s from outputs list",
+				output->wlr_output->name);
+		list_del(root_container.sway_root->outputs, index);
+		wl_list_remove(&output->destroy.link);
+		output->wlr_output = NULL;
+		free(output);
+	}
 }
 
 static void handle_mode(struct wl_listener *listener, void *data) {
@@ -1203,6 +1219,7 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 	output->wlr_output = wlr_output;
 	wlr_output->data = output;
 	output->server = server;
+	list_add(root_container.sway_root->outputs, output);
 
 	if (!wl_list_empty(&wlr_output->modes)) {
 		struct wlr_output_mode *mode =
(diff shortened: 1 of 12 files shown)


ascent12:
With this here, how much effort would it be to implement what was discussed in https://github.com/swaywm/wlroots/issues/479?

emersion:
>With this here, how much effort would it be to implement what was discussed in swaywm/wlroots#479?

This is related to moving workspaces around, I think it belongs to another PR.

ascent12:
>I think it belongs to another PR

Yeah probably. I was just thinking this PR is somewhat related to that goal.

sway/desktop/output.c
```diff
@@ -1203,18 +1219,31 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
 	}
 
-	output->damage = wlr_output_damage_create(wlr_output);
+	output_enable(output);
+}
+
+void output_enable(struct sway_output *output) {
+	struct wlr_output *wlr_output = output->wlr_output;
+
+	if (!wlr_output->data) {
```
emersion: I think this should be set when the `sway_output` is created, not here

sway/desktop/output.c
```diff
@@ -1203,18 +1219,31 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
+		wlr_output->data = output;
+	}
+
+	if (!output->damage) {
+		output->damage = wlr_output_damage_create(wlr_output);
```
emersion: Same for this one

RedSoxFan:
> Looks OK besides these two comments. I think a `wl_list` would be better suited instead of a `list_t`, but it's no big deal.

Addressed the two comments. As for `wl_list`, I'm not too familiar with `wl_list`. In what way(s) is it better suited for this than `list_t`?

emersion:
>In what way(s) is it better suited for this than list_t?

You don't need to allocate a list, you don't need to search the element in the list when removing it. Basically `wl_list` are linked lists and `list_t` is an array. Here is a simple example:

```c
struct sway_root {
  struct wl_list outputs;
};

struct sway_root root_container = {…};

struct sway_output {
  struct wl_list link;
};

// Initialize the list
wl_list_init(&root_container.outputs);

// Add an output
struct sway_output *output = calloc(…);
wl_list_insert(&root_container.outputs, &output->link);

// Iterate
struct sway_output *output;
wl_list_for_each(output, &root_container.outputs, link) {
  // Do something with `output`
}

// Remove an output
wl_list_remove(&output->link);
```

RedSoxFan:
> You don't need to allocate a list, you don't need to search the element in the list when removing it. Basically `wl_list` are linked lists and `list_t` is an array.

Thanks for the explanation and example. I agree that `wl_list` would be better. I'll update the PR.

RedSoxFan:
Updated the PR to use `wl_list` instead of `list_t`.

emersion:
Thanks!

PR #2112, opened 2018-06-06
Fix display of the "verifying" message in swaylock.

smlx:


emersion:
I'm not sure this is the correct fix. If there's an incoming event that isn't the frame `done` event, it'll still not render the updated message.

We need to wait for all surfaces to not be dirty anymore.

smlx:
Hmm. I'm well beyond my bounds of understanding how wayland rendering works in that case. There was some trial and error involved getting this far!

Could you point me towards any code or documentation that might help?

emersion:
Sure, no problem!

swaylock has a rendering loop. The idea is to make sure we don't render more frames than the outputs' frame rate. It works by:
- When swaylock's state changes, `damage_state` (or `damage_surface`) is called, marking all surfaces (there's one surface per output) as dirty (ie. needs re-rendering). It also requests a new frame from the compositor (with `wl_surface_frame` + `wl_surface_commit`). Requesting a frame means "I have something to render, let me know when it's a good time to start rendering".
- The compositor sends frame events, requesting clients to render. swaylock receives it in `surface_frame_handle_done`, calls `render_frame` (which renders + commits the new buffer) and marks the surface as not dirty anymore. It also requests a new frame, so that if the surface is marked dirty, we don't have to wait before rendering (that allows to render at exactly the output's framerate instead of half of it).

Back to the issue this PR tries to fix: normally we have an event loop (`main.c` line 495) so we don't need to do anything in particular for the rendering loop to work. However `attempt_password` blocks until the password is verified, so if we request a frame right before we won't receive the frame event from the compositor until `attempt_password` returns. So we need to make sure all surfaces are not dirty anymore before calling `attempt_password` (ie. swaylock shows "verifying" on all outputs). We can do this by copying the loop and breaking it when not dirty anymore:

```c
damage_state(state);
while (wl_display_dispatch(state->display) != -1 && state->run_display) {
	// TODO: if all surfaces are not dirty anymore, break the loop
}
attempt_password(…);
```

Let me know if something's not clear or if you have more questions.

martinetd:
FWIW the inner loop shouldn't check dirty but frame_pending - when all surfaces have no frame pending left we're good to go.

emersion:
>FWIW the inner loop shouldn't check dirty but frame_pending - when all surfaces have no frame pending left we're good to go.

Meh. `frame_pending` means that we've requested a frame callback, `dirty` means that we need to render again - so I'd rather check `dirty`.

martinetd:
Except that it doesn't work. `dirty` is cleared after the frame has been rendered/commited but before the callback came back.
I agree about the English words, but he'd need to change the logic around the bools to make it true in swaylock

martinetd:
Okay, I take it back after discussing on IRC - `dirty` is fine, but need to call `wl_display_flush(state->display)` after the loop

emersion:
Any news about this?

smlx:
Sorry, I'm not going to have time to finish this anytime soon. I'll close the PR and open a new one later when I get a chance (if it's not fixed by then).

martinetd:
@smlx: thanks for trying; I had a commit back then for testing so I'm pushing it in a new PR but if there's anything else you'd like to work on feel free to ask for some explanation again
(woah, github is scary, it's still listing here the local commit I pushed by mistake and erased right away..)

PR #2115, opened 2018-06-06
Restore workspaces to output when re-enabled

RedSoxFan:
**This is forked from #2108 so that needs to be merged first**

In #2108, it was requested that workspaces be restored to their outputs when re-enabled as discussed in swaywm/wlroots#479. This PR implements that request.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -213,6 +213,9 @@ static struct sway_container *container_workspace_destroy(
 	sway_workspace->floating->parent = NULL;
 	_container_destroy(sway_workspace->floating);
 
+	list_foreach(sway_workspace->output_priority, free);
+	list_free(sway_workspace->output_priority);
+
 	free(sway_workspace);
 
 	if (output) {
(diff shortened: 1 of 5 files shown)


RedSoxFan:
This has been rebased and is ready for review

SirCmpwn:
Not a fan of this approach. I'd rather have each workspace just keep a track of the outputs it has been on in a list of priority. That way, you can remove N (where N>1) monitors and plug them back in and have everything as it was.

RedSoxFan:
> Not a fan of this approach. I'd rather have each workspace just keep a track of the outputs it has been on in a list of priority. That way, you can remove N (where N>1) monitors and plug them back in and have everything as it was.

Sounds good to me. I'll update the PR. I'll need some help testing the implementation for that approach though since I only have two monitors.

sway/tree/container.c
```diff
@@ -433,6 +445,9 @@ void container_descendants(struct sway_container *root,
 			func(item, data);
 		}
 		container_descendants(item, type, func, data);
+		if (i < root->children->length && root->children->items[i] != item) {
```
emersion: Hmm, this is a bit hacky, and only works if the current container is removed. I'd rather copy the list when doing this sort of thing.

emersion: >I would prefer the PR be changed to use a double/nested for loop to iterate the workspaces

Well, that doesn't change the fact that the inner loop will iterate over a list that will be mutated. But I agree it's a better idea.

sway/tree/workspace.c
```diff
@@ -454,3 +456,60 @@ bool workspace_is_empty(struct sway_container *ws) {
...
+		return;
+	}
+
+	int new_index = list_seq_find(ws->output_priority, find_output,
+			output->name);
+	if (new_index < 0) {
+		list_insert(ws->output_priority, old_index, strdup(output->name));
+	} else if (new_index > old_index) {
+		char *name = ws->output_priority->items[new_index];
+		list_del(ws->output_priority, new_index);
```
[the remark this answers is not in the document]
RedSoxFan: No, because instead of using `strdup(output->name)` like I do if it's not in the list, I'm just re-inserting the same copy on the following line. I could do the following, but I felt it was a waste to make a duplicate after freeing a duplicate:

````c
char *name = ws->output_priority->items[new_index];
list_del(ws->output_priority, new_index);
free(name);
list_insert(ws->output_priority, old_index, strdup(output->name));
```

sway/tree/container.c
RedSoxFan: I switched it to using a nested for loop. Since it's only ever removes the current element, it should be fine.

sway/tree/workspace.c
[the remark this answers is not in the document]
RyanDwyer: Ah, yep. I had a derp moment. It was 12:30am local time.

RyanDwyer:
Thanks!

PR #2116, opened 2018-06-06
Fix focusing after splitting tabs/stacks

### LINKED ISSUE #2113: Splitting a tab does something weird
Create layout `T[view, view, view focused]` and run `splitv`. The following happens:

* The tab titles are correct (view title, view title, `V[view class]`).
* The second tab incorrectly has the styling of focused_inactive.
* No tab has focused styling (this is correct, because normally the child of tab 3 is focused)
* The view surface displayed is from tab 2.

It doesn't matter if the tabbed container is the workspace or a child of it.

The issue is not a damage issue, but is fixed upon arranging the workspace (switching to another workspace and back fixes it).

RedSoxFan:
Fixes #2113 

This fixes focusing issues after splitting a child of a tabbed or stacked container.

### DIFF
--- a/sway/tree/layout.c
+++ b/sway/tree/layout.c
@@ -899,10 +899,7 @@ struct sway_container *container_split(struct sway_container *child,
 
 	if (set_focus) {
 		seat_set_focus(seat, cont);
-		if (cont->parent->layout == L_TABBED
-				|| cont->parent->layout == L_STACKED) {
-			seat_set_focus(seat, child);
-		}
+		seat_set_focus(seat, child);
 	}
 
 	container_notify_subtree_changed(cont);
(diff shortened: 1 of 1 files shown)


RyanDwyer:
I found a problem. Create layout `H[view, view focused]` then run `splitv`. The indicator bar appears at the bottom as expected. Then create a new view. It appears horizontally. The final representation is `H[view, V[view], view]`

RedSoxFan:
> I found a problem. Create layout `H[view, view focused]` then run `splitv`. The indicator bar appears at the bottom as expected. Then create a new view. It appears horizontally. The final representation is `H[view, V[view], view]`

Should be fixed now

RyanDwyer:
Thanks!

PR #2118, opened 2018-06-08
Remove strip quoting and fix strncpy-overlap

### LINKED ISSUE #2117: #2084 strips single quotes from exec directives
I noticed today that swayidle wasn't running despite it being in an exec directive in [my sway config](https://github.com/ggreer/dotfiles/blob/master/home/.config/sway/config#L280)

[Here's the relevant debug log](https://gist.github.com/ggreer/72c7dc51ba803ea9444ec72262cacdd8). Notice how the single quotes are stripped from the actual command being executed. This causes swayidle to exit because it doesn't understand the flags that were meant for swaylock.

Reverting the changes in https://github.com/swaywm/sway/pull/2084 fixed the issue for me.

The specific line causing this issue seems to be the quote stripping here: https://github.com/swaywm/sway/commit/067fe9d0472ca24328a962e8704d4e843f2a8af1#diff-5125e6a182d823f6a2ad1c0b7a576892R277

Mentioning @RedSoxFan because it was his PR.

RedSoxFan:
Fixes #2117 

This also fixes a strncpy--param-overlap error by replacing strncpy with memmove.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -274,7 +274,6 @@ struct cmd_results *execute_command(char *_exec, struct sway_seat *seat) {
 			for (int i = handler->handle == cmd_set ? 2 : 1; i < argc; ++i) {
 				argv[i] = do_var_replacement(argv[i]);
 				unescape_string(argv[i]);
-				strip_quotes(argv[i]);
 			}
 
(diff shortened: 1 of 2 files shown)


emersion:
Thanks!

PR #2119, opened 2018-06-08
sway/criteria: gcc string truncation warning fix

martinetd:
gcc complains about 0-length strncpy... it's fine anyway because calloc


emersion:
Thanks!

PR #2120, opened 2018-06-08
Swayidle: doublefork to not leave zombies around

martinetd:
Also using `waitpid()` in sway for exec cmd because we can™


emersion:
Thanks!

PR #2121, opened 2018-06-08
swaylock: implement ^U to clear buffer

martinetd:
The whole state->xcb.modifiers thing didn't work at all (always 0)

The xkb doc says "[xkb_state_serialize_mods] should not be used in
regular clients; please use the xkb_state_mod_*_is_active API instead"
so here it is

The final result is still rather messy so happy to get rebuked, please give me a better idea to do it though.

### DIFF
--- a/swaylock/seat.c
+++ b/swaylock/seat.c
@@ -7,28 +7,6 @@
 #include "swaylock/swaylock.h"
 #include "swaylock/seat.h"
 
-const char *XKB_MASK_NAMES[MASK_LAST] = {
-	XKB_MOD_NAME_SHIFT,
-	XKB_MOD_NAME_CAPS,
-	XKB_MOD_NAME_CTRL,
-	XKB_MOD_NAME_ALT,
-	"Mod2",
-	"Mod3",
-	XKB_MOD_NAME_LOGO,
-	"Mod5",
-};
-
-const enum mod_bit XKB_MODS[MASK_LAST] = {
-	MOD_SHIFT,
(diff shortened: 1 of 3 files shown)


swaylock/seat.c
```diff
@@ -85,15 +63,9 @@ static void keyboard_modifiers(void *data, struct wl_keyboard *wl_keyboard,
...
-	xkb_mod_mask_t mask = xkb_state_serialize_mods(state->xkb.state,
-			XKB_STATE_MODS_DEPRESSED | XKB_STATE_MODS_LATCHED);
-	state->xkb.modifiers = 0;
 	state->xkb.caps_lock = xkb_state_mod_name_is_active(state->xkb.state, XKB_MOD_NAME_CAPS, XKB_STATE_MODS_LOCKED);
-	for (uint32_t i = 0; i < MASK_LAST; ++i) {
-		if (mask & state->xkb.masks[i]) {
-			state->xkb.modifiers |= XKB_MODS[i];
-		}
-	}
+	state->xkb.control = xkb_state_mod_name_is_active(state->xkb.state, XKB_MOD_NAME_CTRL, XKB_STATE_MODS_DEPRESSED | XKB_STATE_MODS_LATCHED);
```
SirCmpwn: 80 col limit

martinetd: crap, I actually thought of it when I wrote this because the previous line was way too long and forgot to wrap it. done in a force push.

emersion:
Thanks!

PR #2129, opened 2018-06-12
Comment floating_modifier to avoid run time warning

floatious:
Sway cannot parse config option floating_modifier.

It appears that this config option has existed before, and is currently
waiting to be reimplemented.

In the meantime commment the config option, so that the user does not
get confused if the config parsing gets halted half-way though or not.


SirCmpwn:
Suprious warnings are not important right now. The master branch is not meant for end-user use, and warnings are useful for developers.

PR #2131, opened 2018-06-12
[WIP] gaps: fixes for edge cases & better alignment with i3-gaps syntax

natesymer:
My previous PR (#2047) had a few bugs with edge cases. This PR is designed to fix those bugs, as well as provide more compatibility with the i3-gaps syntax (specifically the longest gaps command's syntax & semantics).


sway/tree/arrange.c
```diff
@@ -268,6 +284,10 @@ void arrange_children_of(struct sway_container *parent) {
 	if (config->reloading) {
 		return;
 	}
+	if (parent->type == C_OUTPUT) {
+                arrange_output(parent);
```
emersion: The indentation is a bit weird here.

Why do we need that?

sway/tree/arrange.c
```diff
@@ -284,9 +304,27 @@ void arrange_children_of(struct sway_container *parent) {
 		return;
 	}
 
+	if (config->smart_gaps && workspace->current_gaps != (workspace->children->length <= 1 ? 0 : 1)) {
+		if (workspace->current_gaps == 0) {
+			container_damage_whole(workspace);
```
emersion: Why do we need that?

sway/tree/arrange.c
```diff
@@ -111,12 +115,24 @@ void add_gaps(struct sway_container *c) {
...
-	}
+	double gaps = 0;
 
-	double gaps = c->has_gaps ? c->gaps_inner : config->gaps_inner;
+	if (c->type == C_WORKSPACE) {
+		for (int i = 0; i < c->children->length; i++) {
+			printf("Child %p, %d\n", c->children->items[i], ((struct sway_container *)c->children->items[i])->type);
+		}
+		if (!config->edge_gaps || (config->smart_gaps && c->children->length == 1)) {
+			printf("SKIPPING OUTER GAPS %s\n", c->name ? c->name : "UNKNOWN NAME");
```
emersion: If you want to log something, use `wlr_log` (and don't use caps lock!)

sway/commands/gaps.c
```diff
@@ -86,12 +85,11 @@ struct cmd_results *cmd_gaps(int argc, char **argv) {
...
 				amount_idx++;
 				scope = GAPS_SCOPE_ALL;
-			} else if (strcmp(argv[amount_idx], "workspace") == 0) {
-				amount_idx++;
-				scope = GAPS_SCOPE_WORKSPACE;
 			} else if (strcmp(argv[amount_idx], "current") == 0) {
 				amount_idx++;
 				scope = GAPS_SCOPE_CURRENT;
+			} else {
+				return cmd_results_new(CMD_INVALID, "gaps", "gaps inner|outer all|current|current_container set|plus|minus <amount>");
```
emersion: Seems like there's a typo in here

sway/commands/gaps.c
```diff
@@ -104,6 +102,8 @@ struct cmd_results *cmd_gaps(int argc, char **argv) {
 			} else if (strcmp(argv[amount_idx], "minus") == 0) {
 				amount_idx++;
 				op = GAPS_OP_SUBTRACT;
+			} else {
+				return cmd_results_new(CMD_INVALID, "gaps", "gaps inner|outer all|current|current_container set|plus|minus <amount>");
```
emersion: Ditto

sway/commands/gaps.c
```diff
@@ -13,9 +13,8 @@ enum gaps_op {
 };
 
 enum gaps_scope {
-	GAPS_SCOPE_ALL,
-	GAPS_SCOPE_WORKSPACE,
-	GAPS_SCOPE_CURRENT
+	GAPS_SCOPE_ALL, // GLOBAL
+	GAPS_SCOPE_CURRENT // CURRENT WORKSPACE
```
emersion: I don't think these comments are needed

sway/tree/arrange.c
natesymer: If smart gaps are on and a second window is added to a workspace, damage gets left in the outer borders. This clears it.

emersion: TBH I don't think this is the right fix. We should just damage before _and_ after arranging. We only damage after right now.

natesymer: The problem is that adding a container (by opening a terminal for example) to a workspace with a single container already in it *only arranges the new container*. If smart_gaps were on in this situation, the workspace would need to apply outer gaps.

RyanDwyer: Keep in mind there's going to be some significant changes once the atomic layout updates PR is completed and merged. The calls to add damage in response to layout changes will all be removed and done in a single place when the layout is applied. For that reason you might not want to try too hard to get it perfect here.

sway/commands/gaps.c
```diff
@@ -150,31 +152,37 @@ struct cmd_results *cmd_gaps(int argc, char **argv) {
...
-				c->gaps_inner = total;
-			} else {
-				c->gaps_outer = total;
-			}
 
-			if (c->parent) {
-				arrange_children_of(c->parent);
-			} else {
+		switch (scope) {
+			case GAPS_SCOPE_ALL: {
```
emersion: Style: don't intend `switch` labels

```c
switch (scope) {
case GAPS_SCOPE_ALL:
    …
```

sway/commands/gaps.c
```diff
@@ -150,31 +152,37 @@ struct cmd_results *cmd_gaps(int argc, char **argv) {
...
+					c->gaps_outer = total;
+				}
+
+				if (c->parent) {
+					arrange_children_of(c->parent);
+				}
+				break;
+			}
+			default: {
+				return cmd_results_new(CMD_INVALID, "gaps", "gaps inner|outer all|current set|plus|minus <amount>");
```
emersion: This could probably use one of the `#define`s?

sway/tree/view.c
```diff
@@ -458,6 +458,26 @@ void view_execute_criteria(struct sway_view *view) {
...
+		ws = container_parent(ws, C_WORKSPACE);
+	}
+
+	if (!(config->smart_gaps && (ws->children->length == 1 || ws->children->length == 2))) {
+		return;
+	}
+
+	for (int i = 0; i < ws->children->length; i++) {
+		struct sway_container *sibling = cont->parent->children->items[i];
+		if (sibling == cont) continue;
```
emersion: Style: missing brackets

sway/tree/arrange.c
```diff
@@ -111,12 +111,19 @@ void add_gaps(struct sway_container *c) {
...
+	double gaps = 0;
 
-	double gaps = c->has_gaps ? c->gaps_inner : config->gaps_inner;
+	if (c->type == C_WORKSPACE) {
+		if (!config->edge_gaps || (config->smart_gaps && c->children->length == 1)) {
+			return;
+		}
+		gaps = c->has_gaps ? c->gaps_outer : config->gaps_outer;
+	} else {
+		if (config->smart_gaps && container_parent(c, C_WORKSPACE)->children->length < 2) {
```
emersion: Does `config->edge_gaps` need to be checked here?

natesymer:
There's currently a bug where a single C_CONTAINER container with multiple children would not get gapped under smart gaps. I've seen your comments, and I'm working on both.

sway/tree/arrange.c
natesymer: Edge gaps must be enabled to have outer gaps. If disabled, workspaces have no gaps because workspaces only have outer gaps.

natesymer: Ergo edge_gaps must be checked. It's cheap though.

PR #2137, opened 2018-06-15
Fix compilation on Debian

atomnuker:
json-c 0.13 isn't required.
The size_t -> int changes are more correct as that's what the json functions return.


SirCmpwn:
You were right, this isn't acceptable. Leave it as a private patch for Debian.

PR #2143, opened 2018-06-17
Set pool_buffers busy when handing them out

vilhalmer:
Pool buffers keep track of whether the compositor is still using them, so we don't step on its toes. They never got marked as such, though, so toe-stepping could be occurring.


SirCmpwn:
Thanks!

Issue #2063, opened 2018-05-28
Floating status


RyanDwyer:
- [x] Basic floating support
- [x] Automatically detect and float xwayland modals/dialogs/etc
- [x] `sticky`
- [x] `floating_modifier`
- [x] `floating_minimum_size` and `floating_maximum_size`
- [x] `default_floating_border`
- [x] `focus mode_toggle`
- [x] `focus floating` and `focus tiling`
- [x] `move <direction> <px>`
- [x] `move [absolute] position`
- [x] `move position mouse`
- [x] `move container to workspace` on a floating container
- [x] `resize grow|shrink <direction> <px>`
- [x] `resize set <width> <height>`
- [x] Move window by dragging title bar
- [x] Resize floating views by dragging borders
- [x] Ensure `sway_workspace->floating->children` remains in focus order
- [x] Allow floating containers

SirCmpwn:
Note all of the interactive move/resize stuff also applies to tiled views

RedSoxFan:
All items listed above are now implemented

sharethewisdom:
I have a window assigned to a workspace that appears on a large output. I need it to float, and I'd like a larger `floating_minimum_size` and `floating_maximum_size` only for windows in that workspace.

Can this be done? My `for_window` [..] floating_minimum_size 1920 x 1000 doesn't take affect.

`sway version 1.1-rc1-3-g9dca18b6 (May  7 2019, branch 'master')`

PR #2145, opened 2018-06-18
Automatically float xwayland windows

emersion:
This PR makes the compositor connect to xwayland to get fetch atoms. These atoms are then used to automatically float xwayland windows.

Reference: https://github.com/i3/i3/blob/05a1dc0f6e1b2c963b10ff5d69789f6eb54b1b68/src/manage.c#L441

Updates #2063

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -334,8 +334,12 @@ static void handle_request_configure(struct wl_listener *listener, void *data) {
 			ev->width, ev->height);
 		return;
 	}
-	// TODO: Let floating views do whatever
-	configure(view, view->swayc->x, view->swayc->y, view->width, view->height);
+	if (container_is_floating(view->swayc)) {
+		configure(view, view->swayc->x, view->swayc->y, ev->width, ev->height);
+	} else {
+		configure(view, view->swayc->x, view->swayc->y,
+			view->width, view->height);
+	}
 }
 
 static void handle_request_fullscreen(struct wl_listener *listener, void *data) {
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -15,6 +15,14 @@
 #include "sway/tree/layout.h"
 #include "sway/tree/view.h"
 
+static const char *atom_map[ATOM_LAST] = {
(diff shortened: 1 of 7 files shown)


sway/desktop/xwayland.c
```diff
@@ -199,15 +208,32 @@ static void set_fullscreen(struct sway_view *view, bool fullscreen) {
...
-	//
-	// We also want to return true if the NET_WM_STATE is MODAL.
-	// wlroots doesn't appear to provide all this information at the moment.
+	if (xwayland_view_from_view(view) == NULL) {
+		return false;
+	}
+	struct wlr_xwayland_surface *surface = view->wlr_xwayland_surface;
+	struct sway_xwayland *xwayland = &server.xwayland;
+
+	// TODO: return true if the NET_WM_STATE is MODAL
```
RyanDwyer: What's the reason for not actioning this todo?

SirCmpwn:
Did we decide it wasn't possible to pull this information out of wlroots?

sway/desktop/xwayland.c
emersion: wlroots doesn't expose this, yet

emersion:
>Did we decide it wasn't possible to pull this information out of wlroots?

Depends what you mean. We could parse these in wlroots, store them in a bitfield in `wlr_xwayland_surface` and read them here instead. But wlroots won't always support all atoms compositors might need.

SirCmpwn:
bleh.

emersion:
Alternatively, we could decide not to expose atoms, and require wlroots to support all atoms a compositor might need.

atomnuker:
Breaks Firefox downloads (clicking on links does nothing). It does work correctly for save windows though.

emersion:
>Breaks Firefox downloads (clicking on links does nothing). It does work correctly for save windows though.

Can you elaborate? The downloads popup and the download section of the library window works for me.

emersion:
>Breaks Firefox downloads (clicking on links does nothing). It does work correctly for save windows though.

So the "Save As" window where you select a destination directory works, but the "Open with… or Save" dialog that appears before is shrunk to 0x0 (with the texture overflowing out of the border).

SirCmpwn:
Thanks!

PR #2146, opened 2018-06-18
Perform (partial) server initialization before dropping privileges.

tobiasblass:
Some operations during backend creation (e.g. becoming DRM master)
require CAP_SYS_ADMIN privileges. At this point, sway has dropped them
already, though. This patch splits the privileged part of server_init
into its own function and calls it before dropping its privileges.
This fixes the bug with minimal security implications.

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -47,6 +47,8 @@ struct sway_server {
 
 struct sway_server server;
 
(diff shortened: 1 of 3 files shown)


SirCmpwn:
@Hummer12007 I wonder if you take a look at this too?

SirCmpwn:
Apparently he can't.

PR #2152, opened 2018-06-22
Init the dmabuf exporting protocol in wlroots

atomnuker:
Allows desktop capture via the dmabuf-capture wlroots example client.


SirCmpwn:
I love that the work for the compositor is just one line of code.

emersion:
Thanks!

PR #2153, opened 2018-06-22
bugfix: avoid access after free

ael-code:
if src is NULL due to a previous error we cannot use it in the command
result string.

Moreover if `src` points to `p.we_wordv[0]` we cannot use it after
`wordfree(&p)` in the command result string.

Bonus feature: If theere was an error accessing the file, the string
rapresentation of the error is now included in the command result
string.

### DIFF
--- a/sway/commands/output/background.c
+++ b/sway/commands/output/background.c
@@ -3,6 +3,7 @@
 #include <strings.h>
 #include <unistd.h>
 #include <wordexp.h>
+#include <errno.h>
 #include "sway/commands.h"
 #include "sway/config.h"
 #include "log.h"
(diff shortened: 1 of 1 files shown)


sway/commands/output/background.c
```diff
@@ -71,21 +72,27 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
...
+				sprintf(src, "%s/%s", conf_path, p.we_wordv[0]);
 				free(conf);
 			} else {
 				wlr_log(L_ERROR, "Unable to allocate background source");
 			}
 		}
-		if (!src || access(src, F_OK) == -1) {
+
+		if (access(src, F_OK) == -1) {
+			struct cmd_results * cmd_res = cmd_results_new(CMD_FAILURE, "output",
```
emersion: Nitpick: we prefer to add a space before the `*`, and no space after:

```c
struct cmd_results *cmd_res = …
```

PR #2155, opened 2018-06-22
bugfix: avoid access after free

ael-code:
if src is NULL due to a previous error we cannot use it in the command
result string.

Moreover if `src` points to `p.we_wordv[0]` we cannot use it after
`wordfree(&p)` in the command result string.

Bonus feature: If there was an error accessing the file, the string
rapresentation of the error is now included in the command result
string.


ael-code:
sorry for the noise...

SirCmpwn:
Thanks!

PR #2159, opened 2018-06-25
dont focus-follow-mouse when keyboard grab

acrisci:
Compare to i3 behavior:

Right click something to open a popup and focus follows mouse is suspended.


SirCmpwn:
Thanks!

PR #2163, opened 2018-06-26
layer shell: fix some use after free on destroy

martinetd:
Both commits revolve around the same idea that layer shells need to die and cleanup properly before sway's output.

First one is a simple not cleaning up properly on handler being called, but second one is more subtle: both layer shell and sway's output listen to wlr's output destroy event, but there is no ordering guarantee here so I added a new event from sway's output to artificially order them.


RyanDwyer:
Thanks!

PR #2164, opened 2018-06-26
output bg command: fix memleak and refactoring

ael-code:
I'm here again with background command stuff.

I've fixed a memleak, a double free error and made some refactoring.


SirCmpwn:
Thanks!

PR #14, opened 2015-08-10
fixed 'return NULL' in 'void container_map(...)'

taiyu-len:
unless you want it to return something, then void return type could be changed.

PR #12, opened 2015-08-10
return 0 on success.

taiyu-len:
corrected return value for commands.

PR #10, opened 2015-08-10
fixed small memory leak. v2

taiyu-len:
lets try this again.
frees cmd if necessary, and removed pointless copying when exec is one word.

PR #2165, opened 2018-06-27
Implement pid->workspace tracking

SirCmpwn:
When you spawn a process with the exec command, sway now notes the
workspace you had focused and the pid of the child process, then assigns
that workspace to the child when its window appears.

Some of this is carried over from sway 0.15, but with some major
refactoring and centralization of state.


sway/tree/view.c
```diff
@@ -463,9 +464,23 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface) {
...
+	if (view->type == SWAY_VIEW_XWAYLAND) {
+		struct wlr_xwayland_surface *surf =
+			wlr_xwayland_surface_from_wlr_surface(wlr_surface);
+		pid = surf->pid;
+	} else {
+		struct wl_client *client =
+			wl_resource_get_client(wlr_surface->resource);
+		uid_t uid;
+		gid_t gid;
+		wl_client_get_credentials(client, &pid, &uid, &gid);
```
ascent12: You can use NULL for uid and gid if you're not interested in their values:
https://github.com/wayland-project/wayland/blob/0e6ac72288f884ba0321c285a65578492c706534/src/wayland-server.c#L563-L564

sway/tree/workspace.c
```diff
@@ -513,3 +514,113 @@ struct sway_container *workspace_output_get_highest_available(
...
+			if (pid == pw->pid) {
+				wlr_log(L_DEBUG,
+						"found pid_workspace for pid %d, workspace %s",
+						pid, pw->workspace);
+				_pw = pw;
+				break;
+			}
+		}
+
+		if (_pw) {
```
ascent12: I think using a goto to break out of a nested loop is cleaner than this.

SirCmpwn: It's not just to break out of the loop. If we don't find a pid_workspace at the end of wl_list_for_each, pw will still be set to the last one in the list.

SirCmpwn: I can still simplify this nonetheless. New patch coming up momentarily.

sway/tree/workspace.c
```diff
@@ -513,3 +514,108 @@ struct sway_container *workspace_output_get_highest_available(
...
+			if (pid == _pw->pid) {
+				wlr_log(L_DEBUG,
+						"found pid_workspace for pid %d, workspace %s",
+						pid, pw->workspace);
+				pw = _pw;
+				goto found;
+			}
+		}
+		pid = get_parent_pid(pid);
+	} while (pid > 0 && pid != getpid());
```
martinetd: I don't think it's worth checking for `pid != getpid()` here, we double-fork so that should never happen (except for swaybar and a few exceptions maybe), and if it does happen we'll still bail out because sway probably isn't too far from pid 1.
We could spare one iteration and stop a pid > 1 though, I don't think init will ever be in a workspace :D

sway/tree/view.c
```diff
@@ -522,6 +555,17 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface) {
...
 	container_damage_whole(cont);
 	view_handle_container_reparent(&view->container_reparent, NULL);
+
+	if (prev_focus != focus) {
+		struct sway_container *ws = workspace_by_name(prev_ws);
+		if (!ws) {
+			ws = workspace_create(prev_output, prev_ws);
+		}
+		input_manager_set_focus(input_manager, ws);
+		workspace_switch(ws);
```
martinetd: That will really do the switch and a transaction, then when `view_map()` ends it will do a second `arrange_and_commit`... then as you pointed out on irc if this was assigned through a rule the focus is changed back again in a 3rd transaction?

Actually do we ever want to switch to the workspace where a new view got created? That happens when matching parent's pid?
if e.g. firefox ever wants to open a popup in my back when I'm typing something in the terminal I sure wouldn't want switching to firefox's workspace...

sway/tree/workspace.c
```diff
@@ -513,3 +514,108 @@ struct sway_container *workspace_output_get_highest_available(
...
+		ws = workspace_by_name(pw->workspace);
+
+		if (!ws) {
+			wlr_log(L_DEBUG,
+					"Creating workspace %s for pid %d because it disappeared",
+					pw->workspace, pid);
+			ws = workspace_create(pw->output, pw->workspace);
+		}
+
+		wl_list_remove(&pw->link);
```
martinetd: Why remove pw, it can only match once?

sway/tree/workspace.c
```diff
@@ -513,3 +514,108 @@ struct sway_container *workspace_output_get_highest_available(
...
+
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+
+	// Remove expired entries
+	static const int timeout = 60;
+	struct pid_workspace *old, *_old;
+	wl_list_for_each_safe(old, _old, &pid_workspaces, link) {
+		if (now.tv_sec - old->time_added.tv_sec >= timeout) {
+			wl_list_remove(&old->link);
```
martinetd: It's almost never right to just `wl_list_remove` something. It won't be able to cleanup after itself on destroy if that happens before the timeout (and then this will be use-after-free here) and it cannot know if the list has already been cleaned up.

I'm actually not sure I understand the timeout either tbh, in my earlier example, if firefox is over one minute old then its children will not pop in its workspace? What's the idea behind the "inconsistent" behaviour?
I think the whole logic here can only work if we stop double-forking, but instead single-fork and handle sigchld and cleanup pids as they die... i3 does double-fork, how does it behave?

sway/tree/view.c
RyanDwyer: @SirCmpwn You might want to consider waiting until atomic is merged then rebasing this.

sway/tree/view.c
```diff
@@ -479,17 +492,37 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface) {
 			if (!workspace) {
 				workspace = workspace_create(NULL, criteria->target);
 			}
+			prev_focus = focus;
 			focus = seat_get_focus_inactive(seat, workspace);
 		} else {
 			// TODO: CT_ASSIGN_OUTPUT
 		}
 	}
+	free(criterias);
```
emersion: Should this be a `list_free` instead?

sway/tree/workspace.c
SirCmpwn: Aye

sway/tree/view.c
martinetd: `workspace_switch(ws)` still looks here, doesn't add up with your (outdated) comment?

SirCmpwn: Derp

martinetd:
Doesn't seem to work for me? If I try to open an urxvt, run right away 'sleep 3; urxvt' and change workspace before sleep runs out I get this in logs:
```
2018-06-30 23:42:19 - [sway/tree/workspace.c:549] Looking up workspace for pid 25285
2018-06-30 23:42:19 - [sway/tree/workspace.c:558] found pid_workspace for pid 25285, workspace 2
..
2018-06-30 23:42:27 - [sway/tree/workspace.c:549] Looking up workspace for pid 25290
2018-06-30 23:42:27 - [sway/tree/container.c:439] Adding new view 0x61400002ea40:urxvt to container 0x61400002e640 2 3
```

So it looks like the process finds itself because you just added it, removes itself from the list, and the next one doesn't find it as a result.

martinetd:
That minor thing aside it looks good to me, everything I was worried about seems fixed :)

SirCmpwn:
That's not how it works. It's only meant to work consistently with processes launched by sway itself. Try it with stuff bound to keybindings. I also set up dmenu like this:

```
bindsym $mod+d exec dmenu_path | dmenu -p "Run:" -l 10 | xargs swaymsg exec
```

So that sway gets to exec the process I ask for.

martinetd:
> That's not how it works. It's only meant to work consistently with processes launched by sway itself. Try it with stuff bound to keybindings. I also set up dmenu like this:
> ```
> bindsym $mod+d exec dmenu_path | dmenu -p "Run:" -l 10 | xargs swaymsg exec
> ```
> So that sway gets to exec the process I ask for.

Ah, so it *is* supposed to match itself, and work in case you run x in a workspace then switch workspace before the window appears?
I'm not sure about your example (works for me before the patch and doesn't seem to pop the window spawnd by the second swaymsg exec where dmenu was running in the first place), but `bindsym $mod+z exec sleep 2 && urxvt` did the trick.

Small nitpick: the workspace create might still be needed, if there was nothing ni the workspace I left between running bindsym and window creation, the workspace is destroyed when I leave it, and it doesn't show up in swaybar when it is created again when the window appears.
When I switch to it, the window is in the correct workspace thoguh, so the feature works now that I understand what this is about.

SirCmpwn:
>Ah, so it is supposed to match itself, and work in case you run x in a workspace then switch workspace before the window appears?

ya

>Small nitpick: the workspace create might still be needed, if there was nothing ni the workspace I left between running bindsym and window creation, the workspace is destroyed when I leave it, and it doesn't show up in swaybar when it is created again when the window appears.

This is probably an unrelated bug

RyanDwyer:
I managed to create a crash but I've not been able to replicate it since. I was on workspace 1 which was empty, and pressed a bindsym to launch terminator, then switched to workspace 2 before it appeared. It opened on workspace 1 and took me back to workspace 1. Then when I tried to close terminator the crash occurred.

    #0  0x000056180f708baf in container_parent (container=0x700000006, type=C_OUTPUT) at ../sway/tree/container.c:503
    503		} while (container && container->type != type);
    [Current thread is 1 (Thread 0x7f25674f0380 (LWP 21407))]
    (gdb) bt
    #0  0x000056180f708baf in container_parent (container=0x700000006, type=C_OUTPUT) at ../sway/tree/container.c:503
    #1  0x000056180f711fe1 in workspace_is_visible (ws=0x56180fe4a540) at ../sway/tree/workspace.c:437
    #2  0x000056180f6eb712 in render_floating (soutput=0x56180fc7f6f0, damage=0x7ffe89a1ca40) at ../sway/desktop/output.c:889
    #3  0x000056180f6ebc5a in render_output (output=0x56180fc7f6f0, when=0x7ffe89a1ca30, damage=0x7ffe89a1ca40)
        at ../sway/desktop/output.c:987
    #4  0x000056180f6ec241 in damage_handle_frame (listener=0x56180fc7f7e8, data=0x56180fc71250) at ../sway/desktop/output.c:1128
    #5  0x00007f25670f1cff in wlr_signal_emit_safe (signal=0x56180fc712b0, data=0x56180fc71250)
        at ../subprojects/wlroots/util/signal.c:29
    #6  0x00007f25670e330d in output_handle_frame (listener=0x56180fc71348, data=0x56180fc84920)
        at ../subprojects/wlroots/types/wlr_output_damage.c:50
    #7  0x00007f25670f1cff in wlr_signal_emit_safe (signal=0x56180fc84a38, data=0x56180fc84920)
        at ../subprojects/wlroots/util/signal.c:29
    #8  0x00007f25670e67eb in wlr_output_send_frame (output=0x56180fc84920) at ../subprojects/wlroots/types/wlr_output.c:532
    #9  0x00007f25670e682f in schedule_frame_handle_idle_timer (data=0x56180fc84920) at ../subprojects/wlroots/types/wlr_output.c:539
    #10 0x00007f256528a9ac in wl_event_loop_dispatch_idle () at /usr/lib/libwayland-server.so.0
    #11 0x00007f256528aa66 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
    #12 0x00007f256528914c in wl_display_run () at /usr/lib/libwayland-server.so.0
    #13 0x000056180f6de1d4 in server_run (server=0x56180f92a7c0 <server>) at ../sway/server.c:146
    #14 0x000056180f6dda9a in main (argc=4, argv=0x7ffe89a1cf78) at ../sway/main.c:432

Also, if I do the above launch-switch-map trick with terminator, the following gets logged which is worrying:

    [sway/tree/view.c:480] Command 'title_format' failed: Only views can have a title_format

I have `for_window [title=".*"] title_format %title (%shell) (%class%app_id)` in my config.

sway/tree/view.c
```diff
@@ -492,9 +493,21 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface) {
...
 
+	pid_t pid;
+	if (view->type == SWAY_VIEW_XWAYLAND) {
+		struct wlr_xwayland_surface *surf =
+			wlr_xwayland_surface_from_wlr_surface(wlr_surface);
+		pid = surf->pid;
+	} else {
+		struct wl_client *client =
+			wl_resource_get_client(wlr_surface->resource);
+		wl_client_get_credentials(client, &pid, NULL, NULL);
```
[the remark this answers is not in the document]
SirCmpwn: Considered it, but this is probably the only place we're going to use it. Not worth the overhead.

sway/tree/workspace.c
```diff
@@ -516,3 +517,114 @@ struct sway_container *workspace_output_get_highest_available(
...
+	char *workspace;
+	struct timespec time_added;
+
+	struct sway_container *output;
+	struct wl_listener output_destroy;
+
+	struct wl_list link;
+};
+
+static struct wl_list pid_workspaces;
```
emersion: Can't this be in `sway_root`, and initialized there?

SirCmpwn: Does it need to be? The pid namespace is inherently global on Linux, so it's not like we're gaining some feel-good ability to have two sway_roots or anything. I'd rather isolate it to the code that uses it.

emersion: Yeah but this is a little meh:

```c
	if (!pid_workspaces.prev && !pid_workspaces.next) {
		wl_list_init(&pid_workspaces);
		return NULL;
	}
```

But I guess since this is not wlroots it doesn't matter too much.

RyanDwyer:
In `view_execute_criteria`, the call to `seat_set_focus(seat, view->swayc)` at the bottom of the `for` loop needs to be moved to before the call to `execute_criteria`. Otherwise it will try to apply the criteria to the workspace (or wrong view) if you've switched workspace between the launch and map.

For the crash, I can replicate this somewhat reliably now. If it has to create the workspace for the view then the workspace is added to the main/pending tree. The output needs to be arranged so the workspace is put in the output's `current.children` list. The view implementation does the arrange on the view's parent. In this case the parent is the workspace, but the arrange needs to be on the output.

We can either:

* change `view_map` to return the container that needs to be arranged (output if workspace was created, workspace if view is fullscreen, else view's parent) and have the view implementation arrange it, or
* have `view_map` arrange it.

SirCmpwn:
Updated

>Otherwise it will try to apply the criteria to the workspace (or wrong view) if you've switched workspace between the launch and map.

I don't think this is how it works. I can give the full explanation if you want, but by way of demonstration I have a rule which removes borders from Chromium, which I just opened on an unfocused workspace via this feature. Everything behaved as expected.

>For the crash, I can replicate this somewhat reliably now. If it has to create the workspace for the view then the workspace is added to the main/pending tree. The output needs to be arranged so the workspace is put in the output's current.children list. The view implementation does the arrange on the view's parent. In this case the parent is the workspace, but the arrange needs to be on the output.

I either don't understand this or it isn't still a problem, by way of the previous proof.

RyanDwyer:
> criteria to the workspace

I made the fix in another PR and it got merged to master. So when you rebased it would have been resolved in your branch.

> crash

Yeah, it may very well be fixed by now. I'll test it within the next day and let you know.

sway/tree/workspace.c
```diff
@@ -529,3 +530,114 @@ void workspace_detect_urgent(struct sway_container *workspace) {
...
+		free(pw->workspace);
+		free(pw);
+	}
+
+	return ws;
+}
+
+static void pw_handle_output_destroy(struct wl_listener *listener, void *data) {
+	struct pid_workspace *pw = wl_container_of(listener, pw, output_destroy);
+	pw->output = NULL;
```
emersion: This should also do this:

```c
wl_list_remove(&old->output_destroy.link);
wl_list_init(&old->output_destroy.link);
```

PR #2168, opened 2018-06-27
Check if command input has at least 2 arguments

BlueGone:
Using `swaymsg "input identifier"`, was making sway crashing, no matter the validity or not of the input identifier. According to `sway-input(5)`, all input commands have the following format : `input identifier subcmd arg1 [args]`, so at least 3 arguments are expected (identifier, subcmd, and arg1).

### DIFF
--- a/sway/commands/input.c
+++ b/sway/commands/input.c
@@ -31,7 +31,7 @@ static struct cmd_handler input_handlers[] = {
 
 struct cmd_results *cmd_input(int argc, char **argv) {
 	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "input", EXPECTED_AT_LEAST, 1))) {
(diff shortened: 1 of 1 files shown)


emersion:
Hmm, shouldn't 2 arguments be enough? (We might have `input <identifier> enable|disable` in the future)

BlueGone:
Seem pertinent too. I'll go for two.

emersion:
Hmm, `argc` doesn't include "input", are you sure it fixes the segfault?

BlueGone:
Yes, of course. The segfault was caused by the call to `config_subcommand` in `commands/input.c:45`,
because the call was equivalent to `config_subcommand((char **){(char *)NULL}, 0, input_handlers, sizeof(input_handlers))`, which is not handled by `config_subcommand`.

emersion:
Thanks!

PR #2170, opened 2018-06-27
 fix accidently removing borders on XCB_CONFIGURE_REQUEST 

### LINKED ISSUE #1925: Xwayland views not configured properly
Their position is off by the title height after the first configure. To reproduce, open Firefox and open a popup or a menu right after.

If you switch to another workspace and switch back again, it reconfigures the window with the correct position.

apreiml:
The view was configured with the container coordinates.
Although they were right on the first configure, they
changed after a XCB_CONFIGURE_REQUEST, when the
border was already drawn.

May fix #1925


emersion:
Thanks!

PR #2172, opened 2018-06-28
fix handling key modifiers if not pressed at first

### LINKED ISSUE #2169: Key bindings not recognized if non-modifier keys were held first
1. Press o
2. Press mod
3. Release o
4. Press f

At this point with the default config, `fullscreen toggle` should be executed, but instead the focused application gets an `f` keystroke.

apreiml:
fixes #2169


SirCmpwn:
LGTM but I'll have to test it when I get home from work

emersion:
@frsfnrrg Thoughts?

SirCmpwn:
Got annoyed enough by this problem *at* work to test this out now :wink:

PR #2173, opened 2018-06-29
Fix floating views not receiving frame events

### LINKED ISSUE #2161: Floating gpg-pinentry doesn't seem to get frame events
This one is pretty weird. From time to time `pinentry-gnome3` doesn't seem to get frame events and is frozen to a unfocused state. This doesn't seem like a damage issue because switching workspaces doesn't make it go away. Input events are processed correctly though, as I can type my password normally (without visual feedback).

emersion:
That happened when they were in tabbed or stacked containers.

Fixes #2161

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -948,7 +948,7 @@ bool view_is_visible(struct sway_view *view) {
 	// Check view isn't in a tabbed or stacked container on an inactive tab
 	struct sway_seat *seat = input_manager_current_seat(input_manager);
 	struct sway_container *container = view->swayc;
-	while (container->type != C_WORKSPACE) {
+	while (container->type != C_WORKSPACE && container->layout != L_FLOATING) {
 		if (container->parent->layout == L_TABBED ||
 				container->parent->layout == L_STACKED) {
 			if (seat_get_active_child(seat, container->parent) != container) {


sway/tree/view.c
```diff
@@ -944,14 +944,16 @@ bool view_is_visible(struct sway_view *view) {
...
+	bool is_floating = container_is_floating(floater);
+	bool is_sticky = is_floating && floater->is_sticky;
 	// Check view isn't in a tabbed or stacked container on an inactive tab
 	struct sway_seat *seat = input_manager_current_seat(input_manager);
 	struct sway_container *container = view->swayc;
 	while (container->type != C_WORKSPACE) {
 		if (container->parent->layout == L_TABBED ||
 				container->parent->layout == L_STACKED) {
-			if (seat_get_active_child(seat, container->parent) != container) {
+			if (!is_floating && seat_get_active_child(seat,
```
RyanDwyer: If you have a floating tabbed container, and the view being checked is the active child of the tabbed container, this will return false which is incorrect. `is_floating` will be true for any child of the floating container.

I think if you change the `while` condition to be `container->type != C_WORKSPACE && container->layout != L_FLOATING` then that would solve it.

emersion: And then we'll read `workspace->sway_workspace` after the loop :\

Instead I checked if `container` is floating. If a floating container has a direct tabbed/stacked parent, it's always visible. Does that sound better to you?

emersion: My bad, you're right. Force pushed your solution.

RyanDwyer:
Thanks!

PR #2174, opened 2018-06-30
sway views: add helpers to get view and layer from wlr_surface

martinetd:
As discussed on IRC a few days ago with @emersion, get the appropriate wlr shell type depending on the surface type with wlr_surface_is_x then get the view through that `->data` field.

This is a prerequisite for idle inhibit to properly toggle itself on/off depending on surface visibility as the inhibitor only has a wlr_surface to deal with - I have an implementation that uses this and it appears to work but some of the rest depends on atomic transactions (e.g. when to check if view is still visible) so that'll hold off until atomic stuff is merged... But I wanted to get feedbacks on this first anyway.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -694,6 +694,36 @@ void view_child_destroy(struct sway_view_child *child) {
 	}
 }
 
+struct sway_view *view_from_wlr_surface(struct wlr_surface *wlr_surface) {
+	if (wlr_surface_is_xdg_surface(wlr_surface)) {
+		struct wlr_xdg_surface *xdg_surface =
+			wlr_xdg_surface_from_wlr_surface(wlr_surface);
+		return view_from_wlr_xdg_surface(xdg_surface);
+	}
+	if (wlr_surface_is_xdg_surface_v6(wlr_surface)) {
+		struct wlr_xdg_surface_v6 *xdg_surface_v6 =
+			wlr_xdg_surface_v6_from_wlr_surface(wlr_surface);
+		return view_from_wlr_xdg_surface_v6(xdg_surface_v6);
+	}
+	if (wlr_surface_is_xwayland_surface(wlr_surface)) {
+		struct wlr_xwayland_surface *xsurface =
+			wlr_xwayland_surface_from_wlr_surface(wlr_surface);
+		return view_from_wlr_xwayland_surface(xsurface);
+	}
+	if (wlr_surface_is_subsurface(wlr_surface)) {
+		struct wlr_subsurface *subsurface =
+			wlr_subsurface_from_wlr_surface(wlr_surface);
+		return view_from_wlr_surface(subsurface->parent);
+	}
+	if (wlr_surface_is_layer_surface(wlr_surface)) {
+		return NULL;
+	}
+
+	wlr_log(L_DEBUG, "Surface of unknown type (role %s): %p",
+		wlr_surface->role, wlr_surface);
+	return NULL;
+}
+
 static size_t append_prop(char *buffer, const char *value) {
 	if (!value) {
 		return 0;
--- a/include/sway/layers.h
+++ b/include/sway/layers.h
@@ -22,4 +22,7 @@ struct sway_layer_surface {
 struct sway_output;
 void arrange_layers(struct sway_output *output);
 
+struct sway_layer_surface *layer_from_wlr_layer_surface(
+	struct wlr_layer_surface *layer_surface);
+
 #endif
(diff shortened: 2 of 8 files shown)


sway/tree/view.c
```diff
@@ -683,6 +683,38 @@ void view_child_destroy(struct sway_view_child *child) {
...
+	if (wlr_surface_is_xwayland_surface(wlr_surface)) {
+		struct wlr_xwayland_surface *xsurface =
+			wlr_xwayland_surface_from_wlr_surface(wlr_surface);
+		return view_from_wlr_xwayland_surface(xsurface);
+	}
+	if (wlr_surface_is_subsurface(wlr_surface)) {
+		struct wlr_subsurface *subsurface =
+			wlr_subsurface_from_surface(wlr_surface);
+		wlr_surface = subsurface->parent;
+		goto again;
```
emersion: What about

```c
return view_from_wlr_surface(subsurface->parent)
```

?

sway/tree/view.c
```diff
@@ -683,6 +683,38 @@ void view_child_destroy(struct sway_view_child *child) {
...
+		return view_from_wlr_xdg_surface_v6(xdg_surface_v6);
+	}
+	if (wlr_surface_is_xwayland_surface(wlr_surface)) {
+		struct wlr_xwayland_surface *xsurface =
+			wlr_xwayland_surface_from_wlr_surface(wlr_surface);
+		return view_from_wlr_xwayland_surface(xsurface);
+	}
+	if (wlr_surface_is_subsurface(wlr_surface)) {
+		struct wlr_subsurface *subsurface =
+			wlr_subsurface_from_surface(wlr_surface);
```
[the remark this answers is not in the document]
emersion: +1

sway/tree/view.c
```diff
@@ -683,6 +683,38 @@ void view_child_destroy(struct sway_view_child *child) {
...
+		wlr_surface = subsurface->parent;
+		goto again;
+	}
+	if (wlr_surface_is_layer_surface(wlr_surface)) {
+		wlr_log(L_DEBUG, "Surface of type layer has no view");
+		return NULL;
+	}
+
+	wlr_log(L_DEBUG, "Surface of unknown type: %p", wlr_surface);
```
emersion: You can print `wlr_surface->role` too, might be useful

sway/tree/view.c
martinetd: I would do that in a language that optimizes tail-recursion... But I guess once again we don't really care, I'll do that.

sway/tree/view.c
martinetd: Good idea

sway/tree/view.c
emersion: Pretty sure C compilers optimize tail calls at `-O2` or `-O3`

martinetd: hm, right, I wonder why I was convinced it didn't. Guess optimisations weren't turned on properly when I last checked

sway/tree/view.c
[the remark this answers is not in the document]
martinetd: opened swaywm/wlroots#1102 ; this PR won't build until the other one landed

sway/tree/view.c
```diff
@@ -683,6 +683,37 @@ void view_child_destroy(struct sway_view_child *child) {
...
+			wlr_xwayland_surface_from_wlr_surface(wlr_surface);
+		return view_from_wlr_xwayland_surface(xsurface);
+	}
+	if (wlr_surface_is_subsurface(wlr_surface)) {
+		struct wlr_subsurface *subsurface =
+			wlr_subsurface_from_wlr_surface(wlr_surface);
+		return view_from_wlr_surface(subsurface->parent);
+	}
+	if (wlr_surface_is_layer_surface(wlr_surface)) {
+		wlr_log(L_DEBUG, "Surface of type layer has no view");
```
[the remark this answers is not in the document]
martinetd: I kind of want to know if someone tries to use idle inhibitor on something that doesn't have a view, but I can log it on that side; will remove this.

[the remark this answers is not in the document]
SirCmpwn: My fear is that this'll be used in something frequent like testing which view is under a pointer or something, then get logged every frame as you drag the pointer around the screen.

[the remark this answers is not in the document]
martinetd: I plan on using it once at idle inhibitor init rather than every frame (even if there will be a check at every transaction which will be somewhat redundant on interactive resize :/), but yeah can't know where this will be used later. It's gone (force pushed)

PR #2179, opened 2018-06-30
Fix crash with stacking layout after f42bf0ad4

atomnuker:
The "simple" rendering function only applies to tiled views.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -645,7 +645,7 @@ static void render_container_simple(struct sway_output *output,
 			if (child->sway_view->border == B_NORMAL) {
 				render_titlebar(output, damage, child, child->x, child->y,
 						child->width, colors, title_texture, marks_texture);
-			} else if (con->sway_view->border != B_NONE) {
+			} else {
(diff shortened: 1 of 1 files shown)


sway/desktop/output.c
```diff
@@ -645,7 +645,7 @@ static void render_container_simple(struct sway_output *output,
 			if (child->sway_view->border == B_NORMAL) {
 				render_titlebar(output, damage, child, child->x, child->y,
 						child->width, colors, title_texture, marks_texture);
-			} else if (con->sway_view->border != B_NONE) {
+			} else {
```
emersion: Should be `child->sway_view->border != B_NONE` instead?

RyanDwyer: Just using `else` is fine because `render_top_border` returns if the view's `border_top` is 0.

But `render_floating_container` is similar and whatever change we do here should also be done there for consistency.

atomnuker: We can't use else in the render_floating_container function because that will draw a top border. The 2 functions can't be the same, they do different things, if they did the same thing then there would be a single function to do so!

PR #2182, opened 2018-07-01
Don't send frame done to surfaces behind lockscreen

RyanDwyer:
Also, when rendering, don't descend into the tree if the lockscreen is active. Just render the lockscreen's surfaces.

This makes an assumption that an `exclusive_client` will always be a lockscreen. This assumption is also made when handling `bindsym --locked` so I think it's a safe assumption.

This PR paints black behind the lockscreen, so if your lockscreen uses transparency or doesn't cover the whole layout for some reason then you'll just see black there. I know some people like to have their lock shortcut take a screenshot, blur it and apply it as the lockscreen background, and this is still possible.

Maybe my fans will spin down when my computer is locked now.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -915,6 +915,36 @@ static struct sway_container *output_get_active_workspace(
 	return workspace;
 }
 
+bool output_has_opaque_lockscreen(struct sway_output *output,
+		struct sway_seat *seat) {
+	if (!seat->exclusive_client) {
+		return false;
+	}
+
+	struct wlr_layer_surface *wlr_layer_surface;
+	wl_list_for_each(wlr_layer_surface, &server.layer_shell->surfaces, link) {
+		if (wlr_layer_surface->output != output->wlr_output) {
+			continue;
+		}
+		struct wlr_surface *wlr_surface = wlr_layer_surface->surface;
+		if (wlr_surface->resource->client != seat->exclusive_client) {
+			continue;
+		}
+		int nrects;
+		pixman_box32_t *rects =
+			pixman_region32_rectangles(&wlr_surface->current->opaque, &nrects);
+		for (int i = 0; i < nrects; ++i) {
+			pixman_box32_t *rect = &rects[i];
+			if (rect->x1 <= 0 && rect->y1 <= 0 &&
+					rect->x2 >= output->swayc->current.swayc_width &&
+					rect->y2 >= output->swayc->current.swayc_height) {
+				return true;
+			}
+		}
+	}
+	return false;
+}
+
 static void render_output(struct sway_output *output, struct timespec *when,
 		pixman_region32_t *damage) {
 	struct wlr_output *wlr_output = output->wlr_output;
@@ -950,7 +980,7 @@ static void render_output(struct sway_output *output, struct timespec *when,
 	struct sway_view *fullscreen_view = workspace->current.ws_fullscreen;
 	struct sway_seat *seat = input_manager_current_seat(input_manager);
 
-	if (seat->exclusive_client && seat->focused_layer) {
+	if (output_has_opaque_lockscreen(output, seat)) {
 		float clear_color[] = {0.0f, 0.0f, 0.0f, 1.0f};
 
 		int nrects;
@@ -1103,7 +1133,8 @@ static void send_frame_done(struct sway_output *output, struct timespec *when) {
 	struct send_frame_done_data data = {
 		.output = output,
 		.when = when,
-		.exclusive_client = seat->exclusive_client,
+		.exclusive_client = output_has_opaque_lockscreen(output, seat) ?
+			seat->exclusive_client : NULL,
 	};
 
 	struct sway_container *workspace = output_get_active_workspace(output);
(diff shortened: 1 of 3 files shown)


SirCmpwn:
-1, I think this stretches our liberties with these protocols too far

RyanDwyer:
What if we could make it configurable? Such a feature would help with power saving.

SirCmpwn:
I might be able to get behind a general powersaving mode, which would enable quality-of-life-lowering optimizations like this.

atomnuker:
This doesn't lower quality of life. Wayland clients must be able to handle not receiving a frame callback and still function perfectly when they do. No liberties taken here.

RyanDwyer:
I think by quality of life he's referring to denying users the ability to have transparency in their lockscreens.

martinetd:
Tricky because the surface might be transparent, yeah.
Instead of the asumption that this is an exclusive client, if we can go to the output, look if there is a layer in the foreground that is totally opaque, we can then not send frame done to anything behind it -- it should be effectively the same thing for you yet much more accurate.
(We don't send frame done events to surfaces in another workspace for example, this is pretty much the same)

RyanDwyer:
We can't easily tell if a surface is opaque. That requires reading data off the GPU which is slow.

emersion:
>We can't easily tell if a surface is opaque.

What we could do is make `swaylock` set the opaque region of the surface, and make sway use this optimization only if `swaylock`'s opaque region covers the whole output.

SirCmpwn:
>>We can't easily tell if a surface is opaque.
>
>What we could do is make swaylock set the opaque region of the surface, and make sway use this optimization only if swaylock's opaque region covers the whole output.

I think you guys are on the right track. wayland.xml defines an opaque region, we don't need to go to the GPU for this.

RyanDwyer:
This now uses the opaque region. It depends on https://github.com/swaywm/wlroots/pull/1108 being merged. Without it, it won't find the opaque region.

sway/desktop/output.c
```diff
@@ -948,8 +975,30 @@ static void render_output(struct sway_output *output, struct timespec *when,
 
 	struct sway_container *workspace = output_get_active_workspace(output);
 	struct sway_view *fullscreen_view = workspace->current.ws_fullscreen;
+	struct sway_seat *seat = input_manager_current_seat(input_manager);
 
-	if (fullscreen_view) {
+	if (output_has_opaque_lockscreen(output, seat)) {
+		float clear_color[] = {0.0f, 0.0f, 0.0f, 1.0f};
+
+		int nrects;
```
emersion: If it's opaque, we shouldn't need to clear.

sway/desktop/output.c
```diff
@@ -915,6 +915,33 @@ static struct sway_container *output_get_active_workspace(
...
+		}
+		struct wlr_surface *wlr_surface = wlr_layer_surface->surface;
+		if (wlr_surface->resource->client != seat->exclusive_client) {
+			continue;
+		}
+		pixman_box32_t output_box = {
+			.x2 = output->swayc->current.swayc_width,
+			.y2 = output->swayc->current.swayc_height,
+		};
+		if (pixman_region32_contains_rectangle(&wlr_surface->current->opaque,
```
emersion: To be completely standards-compliant, `wlr_surface->current->opaque` should be clipped to the surface's size. Probably better to do this on the wlroots side, with a different field.

sway/desktop/output.c
```diff
@@ -915,6 +915,33 @@ static struct sway_container *output_get_active_workspace(
...
+	struct wlr_layer_surface *wlr_layer_surface;
+	wl_list_for_each(wlr_layer_surface, &server.layer_shell->surfaces, link) {
+		if (wlr_layer_surface->output != output->wlr_output) {
+			continue;
+		}
+		struct wlr_surface *wlr_surface = wlr_layer_surface->surface;
+		if (wlr_surface->resource->client != seat->exclusive_client) {
+			continue;
+		}
+		pixman_box32_t output_box = {
```
emersion: There's no guarantee the layer surface has position 0,0.

swaylock/main.c
```diff
@@ -110,12 +110,27 @@ static void create_layer_surface(struct swaylock_surface *surface) {
 	wl_surface_commit(surface->surface);
 }
 
+static bool image_is_opaque(cairo_surface_t *image) {
+	return cairo_surface_get_content(image) == CAIRO_CONTENT_COLOR;
+}
```
SirCmpwn: What if it's not an image, but a solid color?

emersion:
>What if it's not an image, but a solid color?

This comment as been "marked as outdated", but it's not.

We should probably check the background type, and if it's a color check if the alpha channel is 0.

sway/desktop/output.c
RyanDwyer: This is the output box in output-local coords, so it'll always be 0,0. If the surface isn't at 0,0 then the `pixman_region32_contains_rectangle` check will fail.

emersion: What I mean is: you need to translate `wlr_surface->current.opaque`

SirCmpwn: This is still relevant

SirCmpwn:
Side note

We could totally union all of the opaque regions of all of the layers above the shell layer, then compare it to individual surfaces beneath and stop sending frame events to surfaces which are occluded even if the layer surfaces don't fully cover the output

PR #2183, opened 2018-07-01
Refactor renderer

RyanDwyer:
For those reviewing, I highly recommend looking at the individual commits rather than trying to review it as a whole.

The first commit moves all the rendering related functions from output.c into a new file, render.c. Some of the functions such as `scale_box` are used by both files, so these are copied for now.

The second commit introduces a `struct render_context` which contains the output, damage, scale and output lx and ly. This is a file-scoped variable; the idea being that this file will only ever contain the one public function (`render_output`) so we populate it there and can refer to it anywhere else in the file. This saves us from passing the output and damage through several functions.

The next couple of commits remove `{output_view,unmanaged,layer,drag_icons}_for_each_surface` and make their callers call `surface_for_each_surface` directly.

Another commit removes the rotation related code, which appears to have been copied from rootston and never applied to sway.

The final commit changes how we iterate the surfaces for rendering. Instead of using `surface_for_each_surface` which requires  structs specifically for passing arguments around, we now iterate them directly and render them without requiring these structs. This commit removes `surface_for_each_surface`, `get_surface_box`, `render_view_surfaces` as well as the `render_data` and `root_geometry` structs.

I will apply a similar refactor to what's remaining in output.c, but wanted to get this out the door because this will have rebase issues the longer I leave it.


emersion:
>The next couple of commits remove {output_view,unmanaged,layer,drag_icons}_for_each_surface and make their callers call surface_for_each_surface directly.

This duplicates stuff. Also, RIP popups.

sway/desktop/render.c
```diff
@@ -0,0 +1,878 @@
...
+
+	render_texture(texture, &box, matrix, alpha);
+}
+
+static void render_popup(struct wlr_surface *surface, int sx, int sy,
+		void *data) {
+	struct sway_view *view = data;
+	double ox = view->swayc->current.view_x - context.output_lx + sx;
+	double oy = view->swayc->current.view_y - context.output_ly + sy;
+	render_surface(surface, ox, oy, view->swayc->alpha);
```
emersion: You need to render subsurfaces too.

RyanDwyer:
Rotation is now re-implemented. I used a matrix instead of the previous implementation. The matrix stays in effect for all children, so subsurfaces and any child views will be floated without us having to do anything. I say child views because eventually we'll be able to float containers.

To test, I set `con->rotation = 0.1` at the start of `render_floating_container`. There is a scissoring issue (the damage box should take into account rotation) but that's going out of scope.

sway/desktop/render.c
```diff
@@ -0,0 +1,895 @@
...
+		break;
+	case L_FLOATING:
+		sway_assert(false, "Didn't expect to see floating here");
+	}
+}
+
+static void render_floating_container(struct sway_container *con) {
+	float saved_matrix[9];
+	if (con->rotation) {
+		memcpy(saved_matrix, context.matrix, sizeof(saved_matrix));
```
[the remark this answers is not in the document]
emersion: Or will it? We used to have a `rotate_child_position` to handle this.

sway/desktop/render.c
```diff
@@ -0,0 +1,895 @@
...
+	float matrix[9];
+	enum wl_output_transform transform =
+		wlr_output_transform_invert(surface->current->transform);
+	wlr_matrix_project_box(matrix, &surface_box, transform, 0,
+		context.matrix);
+
+	render_texture(texture, &surface_box, matrix, alpha);
+}
+
+static void render_surfaces(struct wlr_surface *surface, double ox, double oy,
```
emersion: This is just a `wlr_surface_for_each_surface` away. This function will get more complicated when we'll properly support subsurfaces in wlroots (which makes it possible to place subsurfaces behind their parent for instance).

emersion: >I've mentioned this in #sway already, but I've found the iterator functions difficult to follow.

How iterators are currently used is over-complicated and could be simplified a lot.

>I know some developers who strictly believe in not having any duplicated code, and inadvertently add complexity to their code to avoid it.

I agree with you, but here I think iterators would be better.

SirCmpwn:
>I don't like this PR, but the old code is worse and refactoring output.c is necessary. @SirCmpwn merge it if you like it

I'm always down to not merge +928 −814

RyanDwyer:
Closing this for several reasons:

* I don't want the the PR to become too big
* There's conflicts which will be annoying to resolve
* It's holding up my other work like my send frame done PR and popups PR
* People don't seem overly supportive of this PR anyway

I'll keep my branch around in case we ever want to come back to it though.

PR #2186, opened 2018-07-01
Static analysis fixes

martinetd:
Like swaywm/wlroots#1101, a bunch of trivial fixes; most are stuff not done properly on error (meh) but there are a couple of real leaks/small fixes.

I didn't do this as thoroughly as wlroots but it's a start

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -428,8 +428,7 @@ struct cmd_results *config_commands_command(char *exec) {
 
 	struct cmd_handler *handler = find_handler(cmd, NULL, 0);
 	if (!handler && strcmp(cmd, "*") != 0) {
-		char *input = cmd ? cmd : "(empty)";
-		results = cmd_results_new(CMD_INVALID, input, "Unknown/invalid command");
+		results = cmd_results_new(CMD_INVALID, cmd, "Unknown/invalid command");
 		goto cleanup;
 	}
 
@@ -471,10 +470,12 @@ struct cmd_results *config_commands_command(char *exec) {
 	}
 	if (!policy) {
 		policy = alloc_command_policy(cmd);
-		sway_assert(policy, "Unable to allocate security policy");
-		if (policy) {
-			list_add(config->command_policies, policy);
+		if (!sway_assert(policy, "Unable to allocate security policy")) {
+			results = cmd_results_new(CMD_INVALID, cmd,
+					"Unable to allocate memory");
+			goto cleanup;
 		}
+		list_add(config->command_policies, policy);
 	}
 	policy->context = context;
 
--- a/sway/commands/output/mode.c
+++ b/sway/commands/output/mode.c
@@ -36,11 +36,11 @@ struct cmd_results *output_cmd_mode(int argc, char **argv) {
 		}
 	} else {
 		// Format is 1234 4321
+		argc--; argv++;
 		if (!argc) {
 			return cmd_results_new(CMD_INVALID, "output",
 				"Missing mode argument (height).");
 		}
-		argc--; argv++;
 		output->height = strtol(*argv, &end, 10);
 		if (*end) {
 			return cmd_results_new(CMD_INVALID, "output",
--- a/sway/commands/output/position.c
+++ b/sway/commands/output/position.c
@@ -27,11 +27,11 @@ struct cmd_results *output_cmd_position(int argc, char **argv) {
 		}
 	} else {
 		// Format is 1234 4321 (legacy)
+		argc--; argv++;
 		if (!argc) {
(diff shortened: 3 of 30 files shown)


sway/commands.c
```diff
@@ -472,9 +471,7 @@ struct cmd_results *config_commands_command(char *exec) {
 	if (!policy) {
 		policy = alloc_command_policy(cmd);
 		sway_assert(policy, "Unable to allocate security policy");
```
emersion: `sway_assert` is not designed not to always be an assert. We should probably do this:

```c
if (!sway_assert(policy, "Unable to allocate security policy")) {
    goto cleanup;
}
```

sway/commands/output/mode.c
```diff
@@ -36,10 +36,6 @@ struct cmd_results *output_cmd_mode(int argc, char **argv) {
 		}
 	} else {
 		// Format is 1234 4321
-		if (!argc) {
```
emersion: It seems possible to me that this happens.

sway/commands/output/position.c
```diff
@@ -27,10 +27,6 @@ struct cmd_results *output_cmd_position(int argc, char **argv) {
 		}
 	} else {
 		// Format is 1234 4321 (legacy)
-		if (!argc) {
```
emersion: Ditto

sway/config/bar.c
```diff
@@ -91,6 +87,10 @@ struct bar_config *default_bar_config(void) {
 	bar->binding_mode_indicator = true;
 	bar->verbose = false;
 	bar->pid = 0;
+	if (!(bar->mode = strdup("dock"))) goto cleanup;
```
emersion: Braces are mandatory

sway/tree/workspace.c
```diff
@@ -202,14 +201,9 @@ char *workspace_next_name(const char *output_name) {
...
-		l = 3;
-	}
+	int l = snprintf(NULL, 0, "%d", ws_num);
 	char *name = malloc(l + 1);
-	if (!name) {
-		wlr_log(L_ERROR, "Could not allocate workspace name");
+	if (!sway_assert(name, "Cloud not allocate workspace name")) {
 		return NULL;
 	}
 	sprintf(name, "%d", ws_num++);
```
emersion: `++` doesn't do anything useful here

sway/commands/output/mode.c
martinetd: This is checked at the very start of the function, and argc isn't reassigned

sway/config/bar.c
martinetd: Tell that to whoever wrote that if in the first place, but meh, sure

sway/tree/workspace.c
martinetd: Well tbh relying on list length when there can be holes is pretty silly in the first place, a proper fix would be to walk over the list and look for the smallest number not taken (keep calling get_ws_by_name until none comes up?) or the highest number actually used (which could be much smaller than list length if they use names, and only cost trying to convert strings to ints), or it could be assuming the `0`th output is set and want to print `ws_num` with a useless `++` (ok that one is far fetched given default config sets workspace 10 on 0 key)

Tempted to call an out of scope on this one, just setting +1 doesn't fix much, and I'm too lazy to decide on what kind of algorithm to pick

sway/commands.c
martinetd: probably

sway/commands/output/mode.c
emersion: Ah, we probably need to move this check after `argc--; argv++;`

sway/tree/workspace.c
emersion: I agree, it's out of scope.

sway/config/bar.c
emersion: Yeah, I saw that it was like this before too. But when we actually touch those lines, we might as well fix style issues.

sway/commands/output/mode.c
martinetd: meh, that means I need to check the other warnings like this I ignored because it's just not dead code -_- but righ, that sounds logical.

sway/config/bar.c
martinetd: Yeah I'll fix it. You underestimate how much of a pain this whole serie of patch was though :P

emersion: Yeah, I understand this is very tedious and painful work and I'm grateful you take the time to do it!

martinetd: It's done :) (force pushed after fixing each individual commits)

emersion:
Thanks!

PR #2187, opened 2018-07-01
Idle inhibit

martinetd:
Implements idle inhibit, and make it toggle when the surface goes "invisible" (hooks on every transition so whenever layouts change or workspace change)

It should be easy to adjust when the idle protocol works per-output, just need to make view_is_visible return a list of outputs where the thing is visible and disable idle for these specifically.

One thing I'm not totally happy about is I'm taking a pointer to the server struct everywhere in the idle inhibit code... To end up using the global server object in transition -_-
Should I just always use the global?
Also not happy storing the list of idle inhibitors directly in servers struct, I might want to encompass things in a sway_idle_inhibit_manager_v1 or something

### DIFF
--- a/include/sway/desktop/idle_inhibit_v1.h
+++ b/include/sway/desktop/idle_inhibit_v1.h
@@ -1,0 +1,13 @@
+
+#ifndef _SWAY_DESKTOP_IDLE_INHIBIT_V1_H
+#define _SWAY_DESKTOP_IDLE_INHIBIT_V1_H
+#include <wlr/types/wlr_idle_inhibit_v1.h>
+#include "sway/server.h"
+
+struct sway_idle_inhibitor_v1 {
+	struct sway_server *server;
+
+	struct wl_listener destroy;
+};
+
+#endif
--- a/include/sway/desktop/idle_inhibit_v1.h
+++ b/include/sway/desktop/idle_inhibit_v1.h
@@ -6,8 +6,12 @@
 
 struct sway_idle_inhibitor_v1 {
 	struct sway_server *server;
+	struct sway_view *view;
 
+	struct wl_list link;
 	struct wl_listener destroy;
 };
+
+void idle_inhibit_v1_check_active(struct sway_server *server);
 
 #endif
--- a/include/sway/desktop/idle_inhibit_v1.h
+++ b/include/sway/desktop/idle_inhibit_v1.h
@@ -1,17 +1,28 @@
-
 #ifndef _SWAY_DESKTOP_IDLE_INHIBIT_V1_H
 #define _SWAY_DESKTOP_IDLE_INHIBIT_V1_H
 #include <wlr/types/wlr_idle_inhibit_v1.h>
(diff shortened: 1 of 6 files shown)


SirCmpwn:
>Also not happy storing the list of idle inhibitors directly in servers struct, I might want to encompass things in a sway_idle_inhibit_manager_v1 or something

Yeah dot his

include/sway/desktop/idle_inhibit_v1.h
```diff
@@ -0,0 +1,17 @@
+
```
SirCmpwn: What's this line for

SirCmpwn:
Wait does this actually inhibit idle activity? I thought we'd have to somehow communicate this to swayidle

sway/desktop/idle_inhibit_v1.c
```diff
@@ -0,0 +1,57 @@
...
+
+	inhibitor->server = server;
+	inhibitor->view = view_from_wlr_surface(wlr_inhibitor->surface);
+	wl_list_insert(&server->idle_inhibitors_v1, &inhibitor->link);
+
+
+	inhibitor->destroy.notify = handle_destroy;
+	wl_signal_add(&wlr_inhibitor->events.destroy, &inhibitor->destroy);
+
+	wlr_idle_set_enabled(server->idle, NULL, false);
```
SirCmpwn: oh duh here it is

martinetd: Yeah probably, same in `handle_destroy`

martinetd:
>> Also not happy storing the list of idle inhibitors directly in servers struct, I might want to encompass things in a sway_idle_inhibit_manager_v1 or something
> 
> Yeah do this

Done.

Also fixed the direct calls to `wlr_idle_set_enabled` to use the `check_active` helper, so if someone spawns a mpv in the background inhibitor won't kick in as pointed out by emersion.

sway/desktop/idle_inhibit_v1.c
```diff
@@ -0,0 +1,79 @@
...
+			/* Cannot guess if view is visible so assume it is */
+			inhibited = true;
+			break;
+		}
+		if (view_is_visible(inhibitor->view)) {
+			inhibited = true;
+			break;
+		}
+	}
+	wlr_idle_set_enabled(manager->idle, NULL, !inhibited);
```
SirCmpwn: This seems like an odd way to go about it. What if there are multiple inhibitors?

martinetd: That's precisely what this code handles: if any inhibitor has a visible view, the wlr_idle will be inhibited (the argument for `wlr_idle_set_enabled` is `enabled` so `!inhibited`)
If no inhibitors are on visible views then `inhibited` will have its default false value and idle timers will be reactivated.

SirCmpwn: Wow I'm dumb

emersion:
Thanks!

PR #2188, opened 2018-07-01
exec_always: fix leaks

martinetd:
- child would leak in the workspace_record_pid path
 - removing malloc lets us get rid of That Comment nobody seems
to remember what it was about
 - we would leak pipe fds on first fork failling
 - we didn't return an error if second fork failed
 - the final executed process still had both pipe fds
(would show up in /proc/23560/fd in launched programs)
 - we would write twice to the pipe if execl failed for some reason
(e.g. if /bin/sh doesn't exist?!)

FWIW, this has a trivial conflict with the pid workspace stuff; happy to wait till the other one lands first but pointless sitting on the commit for such a small conflict.


SirCmpwn:
Thanks!

PR #2193, opened 2018-07-02
Fix damage on swaybar when view requests to exit fullscreen

### LINKED ISSUE #2191: Whole output not damaged when leaving fullscreen
Only the view's container is, leaving the bar with fullscreen artifacts.

RyanDwyer:
Fixes #2191


emersion:
Doesn't fix it when the view is unmapped while fullscreen?

emersion:
Thanks!

PR #2194, opened 2018-07-02
Don't return pending children in seat_get_active_current_child

### LINKED ISSUE #2192: Opening a view in a tabbed container draws it at 0,0
1. Create a tabbed container
2. Open e.g. a terminal (tried `gnome-terminal` and `termite`, also works with other apps)
3. The first frame of the terminal is drawn at the top-left corner of the screen, over the bar. When the bar updates its content, this part of the screen is cleared and the issue goes away.

RyanDwyer:
Fixes #2192.

`seat_get_active_current_child` is intended to return a **current** child of the given container (ie. a child which has finished its mapping transaction and is able to be rendered on screen). The previous implementation was capable of returning a pending child, which caused a child of a tabbed or stacked view to be rendered prematurely while it was mapping.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -832,12 +832,12 @@ struct sway_container *seat_get_active_child(struct sway_seat *seat,
 
 struct sway_container *seat_get_active_current_child(struct sway_seat *seat,
 		struct sway_container *container) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #2196, opened 2018-07-02
Fix xwayland floating views unclickable

### LINKED ISSUE #2195: Firefox download dialog unclickable
1. Open Firefox
2. Click a link that downloads a file
3. The dialog selection whether to open the file with an app or save it is unclickable (input events go to the window underneath)

Works in rootston.

emersion:
Some xwayland views are first configured with a 1x1 size, and then
resized. Since the view size isn't updated, they are unclickable.

Fixes #2195

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -283,9 +283,7 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 	if (view->swayc->instructions->length) {
 		transaction_notify_view_ready_by_size(view,
 				surface_state->width, surface_state->height);
-	}
-
-	if (container_is_floating(view->swayc)) {
+	} else if (container_is_floating(view->swayc)) {
 		view_update_size(view, surface_state->width, surface_state->height);
 	}
 
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -284,6 +284,11 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		transaction_notify_view_ready_by_size(view,
 				surface_state->width, surface_state->height);
 	}
+
+	if (container_is_floating(view->swayc)) {
+		view_update_size(view, surface_state->width, surface_state->height);
+	}
+
 	view_damage_from(view);
 }
 
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -574,6 +574,8 @@ void view_update_position(struct sway_view *view, double lx, double ly) {
 	container_damage_whole(view->swayc);
 	view->x = lx;
 	view->y = ly;
+	view->swayc->current.view_x = lx;
+	view->swayc->current.view_y = ly;
 	if (container_is_floating(view->swayc)) {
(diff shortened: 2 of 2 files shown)


sway/desktop/xwayland.c
```diff
@@ -284,6 +284,11 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		transaction_notify_view_ready_by_size(view,
 				surface_state->width, surface_state->height);
 	}
+
+	if (container_is_floating(view->swayc)) {
+		view_update_size(view, surface_state->width, surface_state->height);
```
emersion: Note that this function will damage stuff and that it's not necessary because `wlr_surface` does this automatically. Maybe we should just set `view->width` and `view->height`?

RyanDwyer: Are you sure the damage isn't necessary? What if the surface becomes smaller?

I think you'd need to set `view->swayc->current.view_{width,height}` in addition to `view->{width,height}`. The view properties are considered pending, and the current state is what's used for rendering.

However, you might not want to write them into the current state if there are transactions pending. I would try something like:

    if (view->swayc->instructions->length) {
    	transaction_notify_view_ready_by_size(view,
    			surface_state->width, surface_state->height);
    } else if (container_is_floating(view->swayc)) {
    	view_update_size(view, surface_state->width, surface_state->height);
    }

...then make `view_update_size` set both the the pending properties and current properties. Not sure about damage.

emersion: >Are you sure the damage isn't necessary? What if the surface becomes smaller?

Nah, this is handled by `wlr_surface`. Though, it's necessary if the view has borders, so in the end I'll probablu keep it.

>I think you'd need to set view->swayc->current.view_{width,height} in addition to view->{width,height}.

The input code is using pending properties, that's why it worked. We probably need to change that.

>However, you might not want to write them into the current state if there are transactions pending.

Well I don't know what to do if transactions are in progress. Once the transaction is applied the view won't have the size we want (from `surface_state`), it'll have the size of the transaction. We probably need to queue a fake transaction instead.

RyanDwyer: > Well I don't know what to do if transactions are in progress. Once the transaction is applied the view won't have the size we want (from surface_state), it'll have the size of the transaction.

This problem isn't specific to floating xwayland views. We ultimately rely on the client to set their size to what we configure them to. If they don't, the transaction times out and we render them incorrectly.

If we have an xwayland view in a transaction and it resizes to something other than what we tell it to, once the transaction has applied it'll resize the swayc anyway.

emersion: > If we have an xwayland view in a transaction and it resizes to something other than what we tell it to, once the transaction has applied it'll resize the swayc anyway.

What if it's floating?

RyanDwyer: I meant floating, and edited my comment to say so. You saw the pre-edit version.

emersion: Ah okay, this makes more sense

emersion:
Updated

SirCmpwn:
Thanks!

PR #2197, opened 2018-07-02
[WIP] Fix window geometry support

### LINKED ISSUE #2178: Window geometry issues
Window geometry isn't handled properly.

Alacritty sets a window geometry larger than the surface's size. We don't handle that properly.

Floating windows have an offset on their input region (but no offset on input events).

Should also try to un-maximize tiled views, to check that we handle window geometry for tiled views correctly.

emersion:
Please try this and help me fix it.

Test plan: open `gnome-terminal`. Open its Preferences menu (floating).

```
22:54 <emersion> ryan-au: ideas on how to handle widnow geometry?
22:55 <emersion> i've been trying to implement it, but i failed
22:56 <emersion> it's basically a box inside which is the window contents
22:56 <emersion> so we want to crop everything outside it
22:57 <emersion> we basically want to offset view->{x,y,width,height}
22:57 <emersion> problem is, these are used when sending configures
22:58 <emersion> and configures don't include these offsets
22:59 <emersion> so i either get the window geometry right but rendering and input wrong
22:59 <emersion> or the other way around
```

Fixes #2178

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -29,6 +29,7 @@ struct sway_view_impl {
 	const char *(*get_string_prop)(struct sway_view *view,
 			enum sway_view_prop prop);
 	uint32_t (*get_int_prop)(struct sway_view *view, enum sway_view_prop prop);
+	void (*get_geometry)(struct sway_view *view, struct wlr_box *box);
 	uint32_t (*configure)(struct sway_view *view, double lx, double ly,
 			int width, int height);
 	void (*set_activated)(struct sway_view *view, bool activated);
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Rebase please

RyanDwyer:
I had a look into this. From what I understand, the width and height we send in the configure becomes the geometry, and in most cases (not alacritty) the surface may end up being bigger than this. So the surface overflows what we believe the boundaries of the container to be, resulting in damage issues and misaligned pointer events.

I would think the solution would be something like:

* When doing pointer events, add geo.x/y to the coordinates we pass to `wlr_surface_surface_at` (in `container_at_view`).
* In `output_damage_whole_container`, if the container is a view then use the surface size instead of the container's size.
* When rendering a floating container, instead of rendering at view_x, render at view_x - view.geo.x and likewise for y.

emersion:
Nah, I think we should clip the view instead. Same goes for apps drawing shadows.

In alacritty the surface size is smaller than the geometry, and subsurfaces outside the main surface draw decorations (which are part of the geometry).

SirCmpwn:
I think we should clip tiled views and let floating views do whatever they want.

SirCmpwn:
Closing in favor of https://github.com/swaywm/sway/pull/2466

PR #2199, opened 2018-07-03
[WIP] Make the swaylock background only render once

### LINKED ISSUE #1857: swaylock does not blend overwrite on -c NNNNNNXX where XX != FF
In other words, each time swaylock renders a frame with a transparent background it becomes less transparent.

mucamaca:
Fixes #1857


emersion:
This won't work with transparent colors for the indicator.

mucamaca:
Turns out this is a bad idea.

PR #2200, opened 2018-07-04
Fix #1857 -> transparency in swaylock

### LINKED ISSUE #1857: swaylock does not blend overwrite on -c NNNNNNXX where XX != FF

mucamaca:


swaylock/render.c
```diff
@@ -32,7 +32,9 @@ void render_frame(struct swaylock_surface *surface) {
 
 	if (state->args.mode == BACKGROUND_MODE_SOLID_COLOR || !surface->image) {
 		cairo_set_source_u32(cairo, state->args.color);
+		cairo_set_operator (cairo, CAIRO_OPERATOR_SOURCE);
```
emersion: Style: please don't put a space before parens

martinetd:
> We'll need to do as they say in the cairo FAQ. Here's another example of how to do it: emersion/mako:render.c@master#L125

Just setting the operator out of the if ought to work for background images; artificially clearing the surface is a bit overkill :)
(I agree `cairo_save`/`cairo_restore` is better though)

emersion:
Thanks!

PR #2201, opened 2018-07-04
startup: move setenv WAYLAND_DISPLAY before config execs

martinetd:
We would previously run all config commands without the environment,
which would appear to work as our socket name is the default one, but
wayland clients would start up in the wrong sway session.

(This explains why 'sometimes' my swayidle processes wouldn't die with
sway, as they weren't listening to the correct socket)


emersion:
Thanks!

PR #2202, opened 2018-07-04
Fix focus related damage

RyanDwyer:
When you have an unfocused container (so one view is focused_inactive), and you focus any other view in that container, the view with focused_inactive was not damaged. This is because we damaged the
previous focus and new focus, but needed to damage the parent of the new focus.

For example, create layout `H[view 1, V[view 2, view 3]]` and set focus to view 2. `focus left` to focus view 1, so view 2 becomes focused_inactive. Then use the mouse to focus view 3. Do this a couple of times and you'll see damage issues.


emersion:
Thanks!

PR #2203, opened 2018-07-04
ipc-server: add display destroy listener

martinetd:
`wl_event_source_remove()` is illegal after display has been destroyed

```
==20392==ERROR: AddressSanitizer: heap-use-after-free on address 0x607000001240 at pc 0x00000048e86e bp 0x7ffe4b557e00 sp 0x7ffe4b557df0
READ of size 8 at 0x607000001240 thread T0
    #0 0x48e86d in wl_list_insert ../common/list.c:149
    #1 0x7fdf673d4d7d in wl_event_source_remove src/event-loop.c:487
    #2 0x41b742 in ipc_terminate ../sway/ipc-server.c:94
    #3 0x40b1ad in main ../sway/main.c:440
    #4 0x7fdf6664c18a in __libc_start_main ../csu/libc-start.c:308
    #5 0x409359 in _start (/opt/wayland/bin/sway+0x409359)

0x607000001240 is located 48 bytes inside of 72-byte region [0x607000001210,0x607000001258)
freed by thread T0 here:
    #0 0x7fdf692c4880 in __interceptor_free (/lib64/libasan.so.5+0xee880)
    #1 0x7fdf673d371a in wl_display_destroy src/wayland-server.c:1097

previously allocated by thread T0 here:
    #0 0x7fdf692c4c48 in malloc (/lib64/libasan.so.5+0xeec48)
    #1 0x7fdf673d4d9e in wl_event_loop_create src/event-loop.c:522
    #2 0x40acb2 in main ../sway/main.c:363
    #3 0x7fdf6664c18a in __libc_start_main ../csu/libc-start.c:308
```

Alternative fix would be to, like, just not call `wl_event_source_remove` as we're exiting sway anyway but might as well do this somewhat cleanly (static globals, meh)

Open question though: shall handle_display_destroy do more and we could just remove the `ipc_terminate` call in main?

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -31,6 +31,7 @@ static int ipc_socket = -1;
 static struct wl_event_source *ipc_event_source =  NULL;
 static struct sockaddr_un *ipc_sockaddr = NULL;
 static list_t *ipc_client_list = NULL;
+static struct wl_listener ipc_display_destroy;
 
 static const char ipc_magic[] = {'i', '3', '-', 'i', 'p', 'c'};
 
@@ -55,6 +56,22 @@ int ipc_client_handle_writable(int client_fd, uint32_t mask, void *data);
 void ipc_client_disconnect(struct ipc_client *client);
 void ipc_client_handle_command(struct ipc_client *client);
 bool ipc_send_reply(struct ipc_client *client, const char *payload, uint32_t payload_length);
+
+static void handle_display_destroy(struct wl_listener *listener, void *data) {
+	if (ipc_event_source) {
+		wl_event_source_remove(ipc_event_source);
+	}
+	close(ipc_socket);
+	unlink(ipc_sockaddr->sun_path);
+
+	list_free(ipc_client_list);
+
+	if (ipc_sockaddr) {
+		free(ipc_sockaddr);
+	}
+
+	wl_list_remove(&ipc_display_destroy.link);
+}
 
 void ipc_init(struct sway_server *server) {
 	ipc_socket = socket(AF_UNIX, SOCK_STREAM | SOCK_NONBLOCK | SOCK_CLOEXEC, 0);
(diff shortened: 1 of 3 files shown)


sway/ipc-server.c
```diff
@@ -56,6 +57,13 @@ void ipc_client_disconnect(struct ipc_client *client);
 void ipc_client_handle_command(struct ipc_client *client);
 bool ipc_send_reply(struct ipc_client *client, const char *payload, uint32_t payload_length);
 
+static void handle_display_destroy(struct wl_listener *listener, void *data) {
+	if (ipc_event_source) {
+		wl_event_source_remove(ipc_event_source);
+		ipc_event_source = NULL;
+	}
```
emersion: We should also remove the `ipc_display_destroy` listener

martinetd: Good point. I'll do that at the same time as the rest if I get an answer to my question at the end of original post

emersion:
> shall handle_display_destroy do more and we could just remove the ipc_terminate call in main?

Yeah that's probably a good idea.

martinetd:
>> shall handle_display_destroy do more and we could just remove the ipc_terminate call in main?
> 
> Yeah that's probably a good idea.

Done in a force push

sway/ipc-server.c
```diff
@@ -56,6 +57,26 @@ void ipc_client_disconnect(struct ipc_client *client);
...
+	if (ipc_event_source) {
+		wl_event_source_remove(ipc_event_source);
+	}
+	close(ipc_socket);
+	unlink(ipc_sockaddr->sun_path);
+
+	while (ipc_client_list->length) {
+		struct ipc_client *client = ipc_client_list->items[0];
+		ipc_client_disconnect(client);
+	}
```
martinetd: Ah, this loop sneaked in as I was working on leaks and I wondered why I couldn't find this chunk anymore.
It's unrelated so I'll remove it for now

emersion:
Thanks!

PR #2204, opened 2018-07-04
Fix xwayland floating views unclickable, alternative version

### LINKED ISSUE #2195: Firefox download dialog unclickable

emersion:
#2196 tried to fix this bug but ran into issues (that will need to be tackled at some point). This one also fixes the bug, and additionally fixes input for popups in floating views.

Fixes #2195

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -660,14 +660,10 @@ struct sway_container *floating_container_at(double lx, double ly,
 			for (int k = 0; k < ws->floating->children->length; ++k) {
 				struct sway_container *floater =
 					ws->floating->children->items[k];
-				struct wlr_box box = {
-					.x = floater->x,
-					.y = floater->y,
-					.width = floater->width,
-					.height = floater->height,
-				};
-				if (wlr_box_contains_point(&box, lx, ly)) {
-					return container_at(floater, lx, ly, surface, sx, sy);
+				struct sway_container *at =
+					container_at(floater, lx, ly, surface, sx, sy);
+				if (at != NULL) {
+					return at;
 				}
 			}
 		}


SirCmpwn:
What's the story here?

RyanDwyer:
│06:12:04  emersion | ryan-au: have you seen https://github.com/swaywm/sway/pull/2204
    │08:29:49   ryan-au | emersion: Can you explain why that solves the problem?                           
    │08:30:17  emersion | ryan-au: view->{width,height} is set to 1                                        
    │08:30:57  emersion | (and then the view is resized by the client to be larger, but the view size isn't
    │                   | updated)                                                                         
    │08:31:28  emersion | this patch will need to go in anyway because popups                              
    │08:31:34   ryan-au | The swayc dimensions should be resized if a floating view is resized by the client
    │08:31:54  emersion | yeah, but that's tricky                                                          
    │08:32:05   ryan-au | This might explain why my floating rofi window only has a top border :)          
    │08:33:12  emersion | not sure                                                                         
    │08:34:21   ryan-au | When I say the swayc dimensions should be resized, I mean there is code in place  
    │                   | that does that                                                                   
    │08:34:25   ryan-au | It should be doing it already                                                    
    │08:36:01  emersion | oh, i think the swayc is fine                                                    
    │08:36:05  emersion | but the view size isn't                                                          
    │08:36:21  emersion | hmm, i don't remmeber                                                            
    │08:36:26  emersion | remember*                                                                        
    │08:36:59   ryan-au | Your PR is removing the swayc box check                                          
    │08:39:09  emersion | ah, so the swayc size is incorrect

tl;dr: The swayc's dimensions are 1x1, and shouldn't be. Removing the box check is not the right solution to the problem.

emersion:
I mean, the box check shouldn't be here in the first place.

But maybe you want to change how input works, adding box checks everywhere and treating popups separately.

PR #2205, opened 2018-07-04
Fix border weirdness

RyanDwyer:
I don't know why that line is there, but it effectively forces `B_NORMAL` for all tiled views, even if I run `border toggle`. It also disables borders for floating views.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -935,7 +935,7 @@ void container_set_floating(struct sway_container *container, bool enable) {
 		container_add_child(workspace->sway_workspace->floating, container);
 		if (container->type == C_VIEW) {
 			view_init_floating(container->sway_view);
-			view_set_tiled(container->sway_view, true);
+			view_set_tiled(container->sway_view, false);
 		}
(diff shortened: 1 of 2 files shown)


atomnuker:
That's the whole point of the line - to disable borders for floating views.

atomnuker:
What's wrong with leaving the call in view_init_floating()?

RyanDwyer:
Nothing, but the other call to `view_set_tiled` is in `container_set_floating` and it makes sense for them both to be near each other.

atomnuker:
That's okay then.

sway/tree/container.c
```diff
@@ -935,6 +935,7 @@ void container_set_floating(struct sway_container *container, bool enable) {
 		container_add_child(workspace->sway_workspace->floating, container);
 		if (container->type == C_VIEW) {
 			view_init_floating(container->sway_view);
+			view_set_tiled(container->sway_view, true);
```
emersion: Should be `false`?

SirCmpwn:
Thanks!

PR #2206, opened 2018-07-04
Fix a bunch of leaks

martinetd:
This PR depends on #2203 as it touches the same function.

I'm pretty sure I'm still missing some here, I got a couple of crashes while trying to "produce activity" so I got annoyed and very likely missing some as I didn't check various patterns. This should still be a good start.

There also is tons of fontconfig stuff that seems to be alloc'd by cairo but I have no idea where that comes from, hopefully it's really just fontconfig's cache and won't grow endlessly.

wrt. what I said on IRC earlier for swaybar (bar leaking ~250 bytes every 5s) it would seem that this only happens with ASAN, so I guess that's just how much memory ASAN needs to keep track of the... 913 (!) indirect alloc/malloc calls that swaybar does every 5 seconds according to perf record. I guess I'll live with that for now but that means I won't be able to keep ASAN if I seriously start to use sway...

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -481,11 +481,10 @@ void ipc_client_handle_command(struct ipc_client *client) {
 	case IPC_COMMAND:
 	{
 		struct cmd_results *results = execute_command(buf, NULL);
-		const char *json = cmd_results_to_json(results);
-		char reply[256];
-		int length = snprintf(reply, sizeof(reply), "%s", json);
+		char *json = cmd_results_to_json(results);
+		int length = strlen(json);
+		client_valid = ipc_send_reply(client, json, (uint32_t)length);
 		free(json);
-		client_valid = ipc_send_reply(client, reply, (uint32_t)length);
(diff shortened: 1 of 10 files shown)


sway/ipc-server.c
```diff
@@ -480,6 +484,7 @@ void ipc_client_handle_command(struct ipc_client *client) {
 		const char *json = cmd_results_to_json(results);
 		char reply[256];
 		int length = snprintf(reply, sizeof(reply), "%s", json);
+		free((char *)json);
```
martinetd: ah, this is actually not correct; now that I've found the doc of json_object_to_json_string it says json_object_put should free it but that's no longer available in this scope.
I guess we should strdup this

martinetd: Done that.

martinetd:
Also took me a while but I found a way to clean most of the fontconfig clutter (although it is a bit ugh, not sure where to put that `pango_cairo_font_map_set_default(NULL)` call); it's actually looking decent.

PR #2216, opened 2018-07-05
Fix swaybar teardown when workspace buttons hidden

### LINKED ISSUE #2213: Hiding workspace buttons hides the whole bar
Setting `workspace_buttons no` for a bar causes the entire bar to be hidden

RedSoxFan:
Fixes #2213

When workspace buttons are hidden, `wl_display_dispatch` was returning `-1` with `errno = 71`. Running `swaybar` with `WAYLAND_DEBUG=1` gave `wl_shm@4: error 1: invalid size (0)`. `max_height`was `0` on the first frame render due to the status line not being read yet.

~This PR prevents `max_height` from being `0`.~ This PR prevents rendering the bar if the `height` is not greater than `0`

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -497,7 +497,7 @@ void render_frame(struct swaybar *bar, struct swaybar_output *output) {
 		// different height than what we asked for
 		wl_surface_commit(output->surface);
 		wl_display_roundtrip(bar->display);
-	} else {
+	} else if (height > 0) {
 		// Replay recording into shm and send it off
 		output->current_buffer = get_next_buffer(bar->shm,
 				output->buffers,


swaybar/render.c
```diff
@@ -435,7 +435,7 @@ static uint32_t render_to_cairo(cairo_t *cairo,
 	}
 	cairo_paint(cairo);
 
-	uint32_t max_height = 0;
+	uint32_t max_height = 1;
```
emersion: Hmm, this should be already handled by line 488?

RedSoxFan: I added a debug line before line 488 and got `bar->config->height = -1`. In `sway/config/bar.c`, there is `bar->height = -1` (line 82), which is passed across the IPC. Unless `bar <bar-id> height <height>` is set in the config, `bar->config->height` will start off at `-1` and line 488 will not force a positive height. I could change `sway/config/bar.c:82` to `bar->height = 1` instead, if that would be preferred.

emersion:
Thanks!

PR #2219, opened 2018-07-06
Send pointer button events for keyboard-interactive layer surfaces

### LINKED ISSUE #2111: Incorrect handling of wlr-layer-shell protocol with keyboard interactivity
When running the wlroots layer-shell example with keyboard interactivity
`$./layer-shell -k`
the window doesn't receive pointer button events anymore like it does in rootston.

The protocol says the surface should still receive pointer events as normal, it still receives pointer enter and leave events but no more button events.

RedSoxFan:
Fixes #2111 

The function was returning before the call to `wlr_seat_pointer_notify_button` that sends the event to the layer surface.


SirCmpwn:
Thanks!

PR #2220, opened 2018-07-06
Attach destroy handler earlier for input devices

ianyfan:
This prevents it from being bypassed when the device has no seat configuration, which resulted in segfaults when trying to access the removed input.


emersion:
Thanks!

PR #2222, opened 2018-07-07
Fix titles when container titles contain UTF-8 characters

### LINKED ISSUE #1936: Title cropped when it contains UTF-8 chars
1. Open https://www.youtube.com/watch?v=kfgFcMz-VnM&feature=youtu.be
2. The titlebar grows because the title contains UTF-8 chars, but the title text is cropped (look at the brackets for instance)

RyanDwyer:
The title and marks textures would have their height set from the config's computed max font height, but the textures were not regenerated when the config's max font height changed which made a gap appear. Rather than making it regenerate the title textures every time the config font height was changed, I've changed it to just make the textures the height of the title itself and fill any gap when rendering.

Also, the `title_width` and `marks_width` variables have been renamed to make it more obvious that they are in output-buffer-local coordinates.

To test, visit [this YouTube video](https://www.youtube.com/watch?v=kfgFcMz-VnM) or [search for "grilled pork ginger" in Japanese](https://duckduckgo.com/?q=%E8%B1%9A%E3%81%AE%E7%94%9F%E5%A7%9C%E7%84%BC%E3%81%8D+&t=ffab&ia=web).

Fixes #1936.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -521,7 +521,7 @@ static void render_titlebar(struct sway_output *output,
 	size_t inner_width = width - TITLEBAR_H_PADDING * 2;
 
 	// Marks
-	size_t marks_width = 0;
+	size_t marks_ob_width = 0; // output-buffer-local
 	if (config->show_marks && marks_texture) {
 		struct wlr_box texture_box;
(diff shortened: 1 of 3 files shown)


emersion:
Thanks!

PR #2223, opened 2018-07-07
Implement some floating move commands

RyanDwyer:
This implements the following for floating containers:

* `move <direction> <amount>`
* `move [absolute] position <x> <y>`
* `move [absolute] position mouse`

Containers are assigned to new workspaces as needed.

I moved the `strtol` parsing part out of `cmd_move` and into `move_in_direction` because it's specific to that function and was conflicting with the floating arguments.

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -97,4 +97,6 @@ void output_drag_icons_for_each_surface(struct wl_list *drag_icons,
 	struct sway_output *output, struct root_geometry *geo,
 	wlr_surface_iterator_func_t iterator, void *user_data);
 
+struct sway_container *output_get_active_workspace(struct sway_output *output);
+
 #endif
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -97,6 +97,4 @@ void output_drag_icons_for_each_surface(struct wl_list *drag_icons,
 	struct sway_output *output, struct root_geometry *geo,
 	wlr_surface_iterator_func_t iterator, void *user_data);
 
-struct sway_container *output_get_active_workspace(struct sway_output *output);
-
 #endif
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -1,11 +1,13 @@
 #define _XOPEN_SOURCE 500
 #include <string.h>
 #include <strings.h>
+#include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_output.h>
 #include <wlr/types/wlr_output_layout.h>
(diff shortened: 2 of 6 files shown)


SirCmpwn:
Needs conflicts addressed

include/sway/output.h
```diff
@@ -97,4 +97,6 @@ void output_drag_icons_for_each_surface(struct wl_list *drag_icons,
 	struct sway_output *output, struct root_geometry *geo,
 	wlr_surface_iterator_func_t iterator, void *user_data);
 
+struct sway_container *output_get_active_workspace(struct sway_output *output);
```
emersion: This is already defined line 71

sway/commands/move.c
```diff
@@ -208,31 +248,78 @@ static struct cmd_results *move_in_direction(struct sway_container *container,
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
 
+static const char* expected_position_syntax =
```
emersion: Nitpick: this can be `char[]` instead of `char *`

sway/tree/container.c
```diff
@@ -989,3 +992,82 @@ void container_get_box(struct sway_container *container, struct wlr_box *box) {
 	box->width = container->width;
 	box->height = container->height;
 }
+
+/**
+ * Translate the container's position as well as all children.
+ */
+static void container_floating_translate(struct sway_container *con,
+		double x_amount, double y_amount) {
+	con->x += x_amount;
```
emersion: We could assert that this is a floating container here

RyanDwyer: I added the assert in the public function because it seemed like a more appropriate place. Also, `container_floating_translate` will get called of children of floating containers once we implement that, so putting the assertion here wouldn't work anyway.

sway/commands/move.c
RyanDwyer: There's another `char *` near the top of this file. I don't really want to go changing that one too. I think if our preference is `char []` then we should replace them all across all commands in one commit.

emersion: `char[]` is objectively better (see https://www.akkadia.org/drepper/dsohowto.pdf section 2.4.1 Pointers vs. Arrays), so I'd prefer new code not to use `char *` if possible. It's no big deal anyway.

sway/tree/container.c
emersion: Yeah, +1

emersion:
Thanks!

PR #2225, opened 2018-07-07
swaylock: fix the displaying of "verified"

martinetd:
Displaying verified after damaging state needs more than one roundtrip,
so keep looping until surfaces are not dirty anymore

I had this commit laying around for testing since we talked about it in #2112,
might as well use that since the author closed the original PR.
(I don't remember why the flush is needed after the loop, but I said it's needed so it probably was...)


emersion:
Thanks!

Issue #2215, opened 2018-07-05
Logind stuff in swayidle could be replaced with systemd-inhibit


emersion:
https://www.freedesktop.org/software/systemd/man/systemd-inhibit.html

SirCmpwn:
Would that work with elogind?

emersion:
Yes: https://github.com/elogind/elogind/blob/278eff0f0b60eda03a3775a28d2d4a3939eb4eba/meson.build#L1926

SirCmpwn:
Coolio

emersion:
Would probably need to have a `-d` flag to ask swaylock to fork in the background when it's ready. That would be cleaner than the current solution (which is to block inhibition for some time).

emersion:
Hmm, after looking at this again, it seems that it won't be possible to remove dbus. While `systemd-inhibit` would work great to inhibit before going to sleep, we still need to listen to the logind sleep event to trigger the command.

Someone suggested listening to the lock event, considering all of this I'd say it's not out of scope for swayidle since it already listens to the sleep event...

aereaux:
I have something written up for listening to the lock event, but it might need a bit more polishing.  I think either it should listen for both or neither, but I'm not sure which would be preferred.  I think @SirCmpwn had preferred neither (on IRC), but I'm not sure.

emersion:
Yeah, right now it's inconsistent and the suspend event has nothing to do with the user being idle (even if being idle can trigger suspend). Also, it's possible to configure a systemd service to run swaylock on suspend (https://wiki.archlinux.org/index.php/Power_management#Sleep_hooks). Though, it's not possible to run swaylock on a logind lock event without a dedicated daemon.

As you said, we should choose one strategy and stick to it.

emersion:
@SirCmpwn ping?

SirCmpwn:
Yes?

emersion:
Do you want to (1) remove dbus from swayidle or (2) add lock support? See comments above < 3 days old.

SirCmpwn:
I would lean towards (1) but I don't really care so long as dbus stays optional. I don't use swayidle.

PR #2226, opened 2018-07-07
swaylock: daemonize after locking

emersion:
Will be useful for #2215.

### DIFF
--- a/swaylock/main.c
+++ b/swaylock/main.c
@@ -38,6 +38,7 @@ static void daemonize() {
 		exit(1);
 	}
 	if (fork() == 0) {
+		setsid();
 		close(fds[0]);
 		int devnull = open("/dev/null", O_RDWR);
 		dup2(STDOUT_FILENO, devnull);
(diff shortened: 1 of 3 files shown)


swaylock/main.c
```diff
@@ -510,6 +511,11 @@ int main(int argc, char **argv) {
 		create_layer_surface(surface);
 	}
 
+	if (state.args.daemonize) {
+		wl_display_roundtrip(state.display);
```
martinetd: Could this possibly have the same problem as verified, e.g. we'd need to wait till all surfaces are not dirty anymore?
(well, I guess forking is faster than password validation so it's probably fine; but we'd ideally want to make sure the compositor has created the keyboard-exclusive layer surface before forking; not just that we asked the compositor to create it)

emersion: The roundtrip ensures that the compositor has received the request.

We could probably do more stuff like checking the compositor hasn't immediately closed the layer surface (I'm not even sure what we do in this case?) and checking we have sent buffers. The latter will be a more involved change.

martinetd: Well if this change is about "waiting for the lockscreen to be "safe" before forking" then it's still lying a bit, that's what I meant yes.
Meh I guess this can wait.

PR #2228, opened 2018-07-08
Implement `focus mode_toggle`

emersion:


SirCmpwn:
sweet

emersion:
Ehh, you're right, I misunderstood it. My bad,

emersion:
Updated, and also implemented `focus floating|tiling`.

emersion:
Ah, you're right, fixed and rebased.

PR #2230, opened 2018-07-09
Replace empty workspace with moved workspace

vilhalmer:
When moving a workspace to an output with only an empty workspace, remove the empty workspace instead of leaving it until the user then switches to and away from it.

### DIFF
--- a/sway/tree/layout.c
+++ b/sway/tree/layout.c
@@ -168,25 +168,36 @@ void container_move_to(struct sway_container *container,
 	struct sway_container *old_parent = container_remove_child(container);
 	container->width = container->height = 0;
 	container->saved_width = container->saved_height = 0;
-	struct sway_container *new_parent;
+
+	struct sway_container *new_parent, *new_parent_focus;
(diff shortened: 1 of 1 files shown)


sway/tree/layout.c
```diff
@@ -168,25 +168,36 @@ void container_move_to(struct sway_container *container,
...
+	// Get the focus of the destination before we change it.
+	new_parent_focus = seat_get_focus_inactive(seat, destination);
 	if (destination->type == C_VIEW) {
 		new_parent = container_add_sibling(destination, container);
 	} else {
 		new_parent = destination;
 		container_add_child(destination, container);
 	}
+
 	wl_signal_emit(&container->events.reparent, old_parent);
```
RedSoxFan: Nitpick: Group the `wl_signal_emit` line with the block above (since that block does the reparenting) and move the blank line below.

vilhalmer:
Fixed the spacing, and also realized that it was pointless to look for the workspace manually in the focus chain since the `container_reap_empty_recursive` will figure that out.

emersion:
Thanks!

PR #2236, opened 2018-07-09
Fix title bar rendering for nested stacked containers

kupospelov:
To reproduce, open a few views and change layouts to get a stack layout inside a tabbed layout or another stacked layout, e.g.: `T[termite S[termite termite]]`.

The code to render title bar for stacked layouts currently relies on the stacked container `swayc_y` field. However, in the example above `T[termite S[termite termite]]`, and both `termite` and `S[termite termite]` will have `swayc_y` set to the same value (the top bar height in my case). This makes the stacked container title bar be rendered over the tabbed container title bar with a gap below.

The issue does not affect tabbed containers, because they use `swayc_y` of their children. The same approach fixes the behavior for stacked ones, since both `termite` in `S[termite termite]` have `swayc_y` set to a value which takes into account the title bar of the parent container (the top bar height + the tabbed container title bar height in my case).

This commit changes the code to consistently use `swayc_x` and `swayc_y` of the children to render their title bars in stacked and tabbed containers.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -599,9 +599,11 @@ static void render_container_tabbed(struct sway_output *output,
 	struct border_colors *current_colors = &config->border_colors.unfocused;
 	struct sway_container_state *pstate = &con->current;
 
+	int tab_width = pstate->swayc_width / pstate->children->length;
+
 	// Render tabs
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Code looks good to me. Will test it later.

RyanDwyer:
Thanks!

PR #2239, opened 2018-07-10
[WIP] Make swaybg standalone and able to handle multiple outputs

### LINKED ISSUE #2051: Proposal: only one swaybg instance for all outputs
Instead of starting one `swaybg` instance per output, only have one for all of them.
* This is more consistent with `swaybar`
* This removes the need for the compositor to (re-)spawn or kill `swaybg` when an output is hotplugged or when the config is updated
* This makes `swaybg` easier to run on other compositors

emersion:
- [ ] Reload swaybg when an output command is executed
- [ ] Figure out how to pass per-output scaling modes/colors

Fixes #2051

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -474,11 +474,11 @@ void free_sway_binding(struct sway_binding *sb);
 
 struct sway_binding *sway_binding_dup(struct sway_binding *sb);
 
-void load_swaybars();
+void load_swaybars(void);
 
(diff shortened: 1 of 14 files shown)


emersion:
How should the interface be for per-output colors/image scaling modes? Maybe `-o VGA-1 -i picture.jpg -o DP-1 -c 123456`?

PR #2240, opened 2018-07-11
Implement swaylock customization flags

### LINKED ISSUE #1961: Implement swaylock visual customization flags
See the man page. Let's sprinkle some dashes in, too, like `--textcolor` -> `--text-color`.

RedSoxFan:
Closes #1961 

Implements the customization flags for swaylock that were in the man page and their variants for all possible states, with the requested dashes mixed in.

I also implemented `i3lock`'s `-e, --ignore-empty-password`, which prevents empty passwords from being validated _(obviously this flag should not be given if for whatever reason the user's password is empty)_. 

Additionally, since the list of flags has grown to be relatively large, I changed the order of them to make it easier to locate them in the code/usage/man page.


swaylock/main.c
```diff
@@ -381,59 +381,213 @@ static void load_image(char *arg, struct swaylock_state *state) {
...
+		"  -v, --version                  Show the version number and quit.\n"
+		"  --bs-hl-color <color>          Sets the color of backspace"
+		" highlight segments.\n"
+		"  --font <font>                  Sets the font of the text.\n"
+		"  --indicator-radius <radius>    Sets the indicator radius.\n"
+		"  --indicator-thickness <thick>  Sets the indicator thickness.\n"
+		"  --inside-color <color>         Sets the color of the inside of the"
+		" indicator.\n"
+		"  --inside-clear-color <color>   Sets the color of the inside of the"
+		" indicator when cleared.\n"
```
RyanDwyer: Question for @SirCmpwn: Would it be acceptable to break the 80 character rule for these descriptions? I think it would improve readability, and it is only text. If you're reading the code in an 80 column terminal then you're not missing out on much by having the descriptions go off-screen.

SirCmpwn: Why not just something like

```c
        " --inside-clear-color <color>   "
            "Sets the color of the inside of the indicator when cleared.\n"
```

RedSoxFan: Updated to use @SirCmpwn's suggested format

swaylock/main.c
```diff
@@ -381,59 +381,225 @@ static void load_image(char *arg, struct swaylock_state *state) {
 			image->path, image->output_name ? image->output_name : "*");
 }
 
+static void set_default_colors(struct swaylock_colors *colors) {
+	colors->background = 0xFFFFFFFF;
+	colors->bs_highlight = 0xDB3300FF;
+	colors->key_highlight = 0x33DB00FF;
+	colors->separator = 0x000000FF;
+	colors->inside = (struct swaylock_colorset){
```
SirCmpwn: Hmm... is this really valid C99?

ascent12: Yes. It's called a compound literal.

SirCmpwn: gross

PR #2242, opened 2018-07-11
Move floating windows to front when focused

RyanDwyer:
To test, open two floating windows and position them (eg. using `move position mouse`) in a way that they overlap. Click each one and observe it move to be in front of the other one.


SirCmpwn:
Thanks!

PR #2243, opened 2018-07-11
Use saved buffer when fullscreen view is in a transaction

### LINKED ISSUE #2237: mpv/xwayland fullscreen isn't atomic
Fullscreen is not part of transactions yet.

RyanDwyer:
Fixes #2237.

To test, launch mpv, fullscreen the video then unfullscreen it. When unfullscreening, it continued to render the live buffer while the other views were resizing, which meant you can get some frames of smaller video despite being fullscreen.

An easier way to test is to set `TRANSACTION_DEBUG` and increase the `TIMEOUT_MS` to 1000 or so.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -844,7 +844,11 @@ void output_render(struct sway_output *output, struct timespec *when,
 		}
 
 		// TODO: handle views smaller than the output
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

Issue #1737, opened 2018-04-05
Complete resize command


SirCmpwn:
* [x] `resize grow|shrink width|height [<amt>] [px | ppt]`
* [x] `resize grow|shrink <direction> <px> px or <ppt> ppt`
* [x] `resize grow|shrink left|right|up|down`
* [x] `resize set <width> [px | ppt] <height> [px | ppt]`

meakio:
I will try to do it this week.
I need to understand `resize grow|shrink width|height [<amt>] [px | ppt]` and how these commands operate in i3.

PR #2244, opened 2018-07-11
Implement resize command for floating views

RyanDwyer:
Implements the following for floating views:

* `resize set <width> <height>`
* `resize <grow|shrink> <width|height|up|down|left|right> <amount>`

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -29,6 +29,11 @@ enum resize_axis {
 	RESIZE_AXIS_INVALID,
 };
 
+struct resize_amount {
+	int amount;
+	enum resize_unit unit;
+};
+
 static enum resize_unit parse_resize_unit(const char *unit) {
 	if (strcasecmp(unit, "px") == 0) {
 		return RESIZE_UNIT_PX;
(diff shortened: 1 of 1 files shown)


SirCmpwn:
While you're at it, can you implement `resize grow|shrink <direction> <px> px or <ppt> ppt`?

https://github.com/swaywm/sway/issues/1737

RyanDwyer:
My latest commit changes the following:

* `resize grow|shrink <direction> <amount> <unit> or <amount> <other_unit>` is now implemented for both tiling and floating containers.
    * The general idea is you run `resize grow 10ppt or 10px` and the ppt value will apply if you've focused a tiled container, or px for floating.
    * If you throw tricky values at it like `10 or 20px` then it'll figure out the most appropriate value to use. In this example, a floating container would use the 20px while a tiling container would use the 10 and interpret it as ppt.
* If running `resize set 0 500` on a floating container, it now ignores the 0 axis as per the i3 docs rather than setting it to 0.
* My initial commit would introduce a failed assertion if you tried to run `resize grow|shrink <direction|axis>` with an up/down/left/right direction on a tiled container. I'm not sure if we even want to support this, so I've made these directions function the same as width or height, and are marked as TODO.

I've also made some improvements to the code structure in this file:

* Parsing the amount along with the unit is a task we have to do in several places, so I made it a function and created a `struct resize_amount` to represent an amount + unit.
* The `cmd_resize` entry function previously delegated the `set` stuff to another function and handled the `grow|shrink` stuff itself. It now delegates both.
* There are now function names `resize_{adjust,set}_{tiled,floating}`. `adjust` is used for a `grow` or `shrink` argument.
* When it comes to resizing a tiled container, I chose to leave the original resize code as untouched as possible. So `resize_adjust_tiled` calls the original `resize_tiled` with the raw amount and unit rather than the passing the new struct.

SirCmpwn:
Nice work!

RyanDwyer:
Forcing the min/max size is now done.

At first glance the changes I made to `resize_adjust_floating` in this commit will look fairly significant, but all I'm doing is splitting the switch into two switches because I have to constrain the grow amount between them (calculate width/height, then constrain them, then calculate x/y).

SirCmpwn:
Thanks!

PR #2245, opened 2018-07-11
Implement floating_minimum_size and floating_maximum_size

RyanDwyer:
Relevant i3 docs: https://i3wm.org/docs/userguide.html#_constraining_floating_window_size


emersion:
Thanks!

PR #2248, opened 2018-07-11
fix crash on floating windows

Dudemanguy911:
https://github.com/swaywm/sway/commit/15dc5286e280ddd06e845dc57115243e72f2339e crashes sway when opening floating windows. The reason for this is because ``current.children`` can have a length of 0 thus causing `sway_assert` and `list_del` to fail. Since I don't think there's any purpose in moving items around in a zero-length list, I just added a return condition to ``list_move_to_end``.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -671,7 +671,9 @@ void seat_set_focus_warp(struct sway_seat *seat,
 	// This must happen for both the pending and current children lists.
 	if (container_is_floating(container)) {
 		list_move_to_end(container->parent->children, container);
-		list_move_to_end(container->parent->current.children, container);
+		if (container_has_ancestor(container, container->current.parent)) {
+			list_move_to_end(container->parent->current.children, container);
+		}
 	}
 
 	// clean up unfocused empty workspace on new output


emersion:
Hmm. This function is designed to fail if the item isn't found in the list:

```c
if (!sway_assert(i < list->length, "Item not found in list")) {
```

This probably indicates a deeper issue - ie. a floating container not yet inserted into `container->parent->children`? (see `sway/input/seat.c:672`)

Dudemanguy911:
>This probably indicates a deeper issue - ie. a floating container not yet inserted into container->parent->children?

Don't some applications do that? For example, anki opens up with a floating window first before the rest of the application opens.

emersion:
>Don't some applications do that? For example, anki opens up with a floating window first before the rest of the application opens.

This is unrelated. When an application opens a floating window, it creates a floating container which is added to the list of floating containers. When focus is given to the container we just move it in the list of floating containers.

Maybe in this case the floating container isn't yet in the `current` list? Is the first line failing or the second line failing in `seat.c`?

Dudemanguy911:
>Maybe in this case the floating container isn't yet in the current list? Is the first line failing or the second line failing in seat.c?

It's the second line.

emersion:
Ah, that's probably the issue. We can probably check that the item is in the `current` list before trying to move it.

Dudemanguy911:
So you mean just making sure that `container->parent->current.children->length` is greater than 0 in `seat.c` before doing `list_move_to_end`?

emersion:
>So you mean just making sure that container->parent->current.children->length is greater than 0 in seat.c before doing list_move_to_end?

No, because you could e.g. open a floating window when there are already other floating windows, in which case the length would be > 0 but the container would still not be in the list.

Maybe just checking if there are no transactions in progress would be enough? (`container->instructions->length == 0`)

Dudemanguy911:
>Maybe just checking if there are no transactions in progress would be enough? (container->instructions->length == 0)

This doesn't seem to work (or at least the various ways I've tried to use the instructions struct). I think the current state can have no instructions and floating windows so `list_move_to_end` gets executed anyway. For now, I changed `seat.c` to use `has_ancestor` for checking the `current.parent` since it works better than the original PR, but I don't know if this is the preferred approach.

RyanDwyer:
Misclick.

RyanDwyer:
So I think what's happening here is the view is mapping as floating, and receives focus before it has mapped which is why it's not in the current list. If we avoid moving the view to the end of the current list then it may appear behind other floating views until a transaction is made (which replaces the current children list with the pending children list). However... it's likely to be at the end of the list anyway because the view was just mapped. So this would only be a problem if two floating views have mapped at the same time, and they overlap each other.

The correct solution would be to use transaction, but we haven't determined the best way to do this from low level code like this. But I'll approve this and deal with the edge case above later if it pops up.

PR #2252, opened 2018-07-12
feature: scroll button option for input devices

rkubosz:
Hello!

I missed in sway an option to set a scrolling button for my trackball Logitech M570, so I decided to implement it myself.

This patch adds support for `scroll_button` command in user's config file. Below is an example of my working config:
```
input "1133:4136:Logitech_M570" {
	scroll_method on_button_down
	scroll_button 273
}
```
In result I can scroll with holding the right button instead of default middle button.

The scroll button identifier is the same as in `libinput debug-events`.

I decided to set min and max value of button identifier to 0 and 1000, because I don't really know how the situation will be with other input devices. I am sure, that the minimum possible value must be set to 0, because libinput accepts it and disables the `on_button_down` scroll method. [(source)](https://wayland.freedesktop.org/libinput/doc/latest/group__config.html#gac95a25055b22c3631e3c10c0463ca332)

### DIFF
--- a/sway/commands/input/scroll_button.c
+++ b/sway/commands/input/scroll_button.c
@@ -1,5 +1,6 @@
 #include <string.h>
 #include <strings.h>
+#include <errno.h>
 #include "sway/config.h"
 #include "sway/commands.h"
 #include "sway/input/input-manager.h"
@@ -18,12 +19,18 @@ struct cmd_results *input_cmd_scroll_button(int argc, char **argv) {
 	struct input_config *new_config =
 		new_input_config(current_input_config->identifier);
 
+	errno = 0;
 	char *endptr;
-	long scroll_button = strtol(*argv, &endptr, 10);
+	int scroll_button = strtol(*argv, &endptr, 10);
 	if (endptr == *argv && scroll_button == 0) {
 		free_input_config(new_config);
 		return cmd_results_new(CMD_INVALID, "scroll_button",
 				"Scroll button identifier must be an integer.");
+	}
+	if (errno == ERANGE) {
+		free_input_config(new_config);
+		return cmd_results_new(CMD_INVALID, "scroll_button",
+				"Scroll button identifier out of range.");
 	}
 	if (scroll_button < 0) {
 		free_input_config(new_config);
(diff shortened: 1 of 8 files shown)


sway/commands/input/scroll_button.c
```diff
@@ -0,0 +1,31 @@
...
+		config->handler_context.input_config;
+	if (!current_input_config) {
+		return cmd_results_new(CMD_FAILURE, "scroll_button",
+			"No input device defined.");
+	}
+	struct input_config *new_config =
+		new_input_config(current_input_config->identifier);
+
+	int scroll_button = atoi(argv[0]);
+	if (scroll_button < 0 || scroll_button > 1000) {
```
emersion: Is it really necessary to check `scroll_button > 1000`?

sway/commands/input/scroll_button.c
```diff
@@ -0,0 +1,31 @@
...
+		free_input_config(new_config);
+		return cmd_results_new(CMD_INVALID, "scroll_button",
+				"Input out of range [1, 10]");
```
emersion: This message needs to be updated.

emersion:
This looks pretty good to me. Can you update the man pages too?

sway/commands/input/scroll_button.c
rkubosz: Probably not. In the worst case the identifier will not match any physical button and the `scroll_button` option will not work. I don't see any chance to crash sway or make it unusable by setting too big identifier.

I will modify it.

rkubosz:
Yes, I will add appropriate paragraph to man pages.

sway/commands/input/scroll_button.c
```diff
@@ -0,0 +1,31 @@
...
+	struct input_config *current_input_config =
+		config->handler_context.input_config;
+	if (!current_input_config) {
+		return cmd_results_new(CMD_FAILURE, "scroll_button",
+			"No input device defined.");
+	}
+	struct input_config *new_config =
+		new_input_config(current_input_config->identifier);
+
+	int scroll_button = atoi(argv[0]);
```
SirCmpwn: Use strtol and detect errors

SirCmpwn:
Thanks!

PR #2255, opened 2018-07-12
Don't disable borders for xwayland floating views

emersion:


atomnuker:
Tested, works. Is there any way to only have borders on floating windows?

emersion:
>Is there any way to only have borders on floating windows?

Hmm, maybe a criteria can do it?

PR #2256, opened 2018-07-12
Add xdg-positioner support

emersion:
Test plan: open right-click menus in Gedit, make sure they don't show up outside the current output.


atomnuker:
Tested, works perfectly.

SirCmpwn:
I really like that this is +50 lines per shell, nice work @acrisci

PR #2260, opened 2018-07-13
Fix output_has_opaque_lockscreen

emersion:
The `pixman_region32_contains_rectangle` call was using `wlr_surface->current.opaque` instead of `surface_opaque_box` (which was then unused).

The new `wlr_surface->opaque_region` is now used, fixing clients that set an infinite region and clients using a buffer without an alpha channel.


SirCmpwn:
Thanks!

PR #2264, opened 2018-07-13
IPC Events (1.0)

### LINKED ISSUE #1730: ipc events
Window events:

* [ ] new – the window has become managed by i3
* [ ] close – the window has closed
* [ ] focus – the window has received input focus
* [ ] title – the window’s title has changed
* [ ] fullscreen_mode – the window has entered or exited fullscreen mode
* [ ] move – the window has changed its position in the tree
* [ ] floating – the window has transitioned to or from floating
* [ ] urgent – the window has become urgent or lost its urgent status
* [ ] mark – a mark has been added to or removed from the window

Workspace events:

* [x] focus
* [x] init
* [ ] empty
* [ ] urgent
* [ ] reload
* [ ] rename
* [ ] restored
* [ ] move

Output events:

* [ ] unspecified?

Mode events:

* [ ] change

Barconfig Update:

* [ ] run

Shutdown:

* [ ] restart

Tick:

* [ ] tick

ianyfan:
Fixes #1730 

- [x] workspace
    - [x] focus - not satisfied with the hack in the last commit, but also not happy with old behaviour
    - [x] init
    - [x] empty
    - [x] urgent (added in #2276)
    - [x] reload
    - [x] rename (already implemented)
    - [x] ~~restored~~
    - [x] move (already implemented but I haven't tested them yet)
- [x] output (no valid events for them currently)
- [x] mode (already implemented)
- [x] window
    - [x] new (already implemented)
    - [x] close
    - [x] focus
    - [x] title
    - [x] fullscreen_mode (already implemented)
    - [x] move
    - [x] floating (already implemented)
    - [x] urgent (added in #2276)
    - [x] mark - ~I'm not satisfied with its implementation, but with the current api it's hard to work out the best places to add the events~ rewrote it to better conform to i3 behaviour
- [x] barconfig_update
    - [x] on config update
- [x] binding
    - [x] mouse binding
- [x] shutdown - though there are some timing issues, since the server seems to shut down before it can emit the shutdown event
    - [x] ~~restart~~
- [x] tick

### DIFF
--- a/completions/bash/swaymsg
+++ b/completions/bash/swaymsg
@@ -14,7 +14,8 @@ _swaymsg()
     'get_marks'
     'get_bar_config'
     'get_version'
-    'get_clipboard'
+    'get_binding_modes'
+    'get_config'
   )
 
   short=(
--- a/completions/bash/swaymsg
+++ b/completions/bash/swaymsg
@@ -16,6 +16,7 @@ _swaymsg()
     'get_version'
     'get_binding_modes'
     'get_config'
+    'send_tick'
   )
 
   short=(
--- a/include/ipc.h
+++ b/include/ipc.h
@@ -27,8 +27,9 @@ enum ipc_command_type {
 	IPC_EVENT_WINDOW = ((1<<31) | 3),
 	IPC_EVENT_BARCONFIG_UPDATE = ((1<<31) | 4),
 	IPC_EVENT_BINDING = ((1<<31) | 5),
-	IPC_EVENT_MODIFIER = ((1<<31) | 6),
-	IPC_EVENT_INPUT = ((1<<31) | 7),
+	IPC_EVENT_SHUTDOWN = ((1<<31) | 6),
+	IPC_EVENT_MODIFIER = ((1<<31) | 16),
+	IPC_EVENT_INPUT = ((1<<31) | 17),
 };
 
 #endif
--- a/include/ipc.h
+++ b/include/ipc.h
@@ -30,8 +30,6 @@ enum ipc_command_type {
 	IPC_EVENT_BINDING = ((1<<31) | 5),
 	IPC_EVENT_SHUTDOWN = ((1<<31) | 6),
 	IPC_EVENT_TICK = ((1<<31) | 7),
-	IPC_EVENT_MODIFIER = ((1<<31) | 16),
-	IPC_EVENT_INPUT = ((1<<31) | 17),
 };
 
 #endif
--- a/include/ipc.h
+++ b/include/ipc.h
@@ -15,6 +15,7 @@ enum ipc_command_type {
 	IPC_GET_VERSION = 7,
 	IPC_GET_BINDING_MODES = 8,
 	IPC_GET_CONFIG = 9,
+	IPC_SEND_TICK = 10,
 
 	// sway-specific command types
 	IPC_GET_INPUTS = 100,
@@ -28,6 +29,7 @@ enum ipc_command_type {
 	IPC_EVENT_BARCONFIG_UPDATE = ((1<<31) | 4),
 	IPC_EVENT_BINDING = ((1<<31) | 5),
 	IPC_EVENT_SHUTDOWN = ((1<<31) | 6),
+	IPC_EVENT_TICK = ((1<<31) | 7),
 	IPC_EVENT_MODIFIER = ((1<<31) | 16),
 	IPC_EVENT_INPUT = ((1<<31) | 17),
 };
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -1,3 +1,4 @@
+#define _XOPEN_SOURCE 500
 #ifdef __linux__
 #include <linux/input-event-codes.h>
 #elif __FreeBSD__
@@ -5,9 +6,11 @@
 #endif
 #include <xkbcommon/xkbcommon.h>
 #include <xkbcommon/xkbcommon-names.h>
+#include <string.h>
 #include <strings.h>
 #include "sway/commands.h"
 #include "sway/config.h"
+#include "sway/ipc-server.h"
 #include "list.h"
 #include "log.h"
 #include "stringop.h"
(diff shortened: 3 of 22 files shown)


SirCmpwn:
bump

ianyfan:
Rebased. Still not sure about the last commit though.

SirCmpwn:
This still needs mouse bindings, right? We implemented those.

ianyfan:
I added it in when I rebased it last time, because it messed up the commit anyway. Probably should have mentioned that, sorry.

(Also, I separated out the commits rather gratuitously so that I could keep track of each event, but I guess some of them could be squashed together if that seems sensible)

SirCmpwn:
Gotcha, thanks. No worries about squashing, the history LGTM.

completions/bash/swaymsg
```diff
@@ -14,7 +14,8 @@ _swaymsg()
     'get_marks'
     'get_bar_config'
     'get_version'
-    'get_clipboard'
+	'get_binding_modes'
+	'get_config'
```
RyanDwyer: This file uses two spaces for indenting but these lines are using tabs. I prefer tabs, but let's change these to spaces for consistency.

include/ipc.h
```diff
@@ -27,8 +28,10 @@ enum ipc_command_type {
 	IPC_EVENT_WINDOW = ((1<<31) | 3),
 	IPC_EVENT_BARCONFIG_UPDATE = ((1<<31) | 4),
 	IPC_EVENT_BINDING = ((1<<31) | 5),
-	IPC_EVENT_MODIFIER = ((1<<31) | 6),
-	IPC_EVENT_INPUT = ((1<<31) | 7),
+	IPC_EVENT_SHUTDOWN = ((1<<31) | 6),
+	IPC_EVENT_TICK = ((1<<31) | 7),
+	IPC_EVENT_MODIFIER = ((1<<31) | 16),
+	IPC_EVENT_INPUT = ((1<<31) | 17),
```
RyanDwyer: Why 16 and 17?

Sway 0.15 uses the following - not sure if we need to use the same:

    IPC_EVENT_MODIFIER = ((1<<31) | 6),
    IPC_EVENT_INPUT = ((1<<31) | 7),

(`SHUTDOWN` and `TICK` aren't listed in 0.15. I guess there's no events for those.)

sway/commands/bind.c
```diff
@@ -27,6 +30,33 @@ void free_sway_binding(struct sway_binding *binding) {
 	free(binding);
 }
 
+struct sway_binding *sway_binding_dup(struct sway_binding *sb) {
+	struct sway_binding *new_sb = malloc(sizeof(struct sway_binding));
```
RyanDwyer: Use `calloc` to allocate structs.

sway/tree/container.c
```diff
@@ -319,7 +321,13 @@ static struct sway_container *container_destroy_noreaping(
 	}
 
 	wl_signal_emit(&con->events.destroy, con);
-	ipc_event_window(con, "close");
+
+	// emit IPC event
+	if (con->type == C_VIEW) {
+		ipc_event_window(con, "close");
+	} else if (con->type == C_WORKSPACE) {
+		ipc_event_workspace(NULL, con, "empty");
```
RyanDwyer: This is being sent when a workspace is being destroyed. Shouldn't it be sent when a workspace becomes empty?

include/ipc.h
SirCmpwn: We need to match the i3 dudes where they line up and we need to put our extensions far away from them so i3 has room to grow.

sway/tree/container.c
ianyfan: This seems to be the behaviour in i3. The naming confuses me too.

sway/commands/bind.c
ianyfan: Sure, I just copied the code over from the 0.15 branch and updated the field allocations.

include/ipc.h
progandy: If the extensions have to be far away, what about using ```((1<<31) | (1<<30) | 1)``` and higher for sway specific events?

RyanDwyer: I don't think SirCmpwn noticed the `1<<31`. These values are already far away from i3's.

My point is sway 0.15 already has `(1<<31) | 6` allocated to `IPC_EVENT_MODIFIER`. My comment is about being compatible with sway 0.15, not with i3.

ianyfan: I think this is a moot point anyway because as far as I can tell, `IPC_EVENT_MODIFIER` and `IPC_EVENT_INPUT` don't actually appear anywhere else in the code.

ianyfan: Eh, modifier events are actually used in 0.15, but not input events. So I think it depends how important modifier events are to be preserved.

Anyway, I'm more inclined to bring the values in line with i3.

progandy: @RyanDwyer (1<<31) is not sway specific. This bit marks an event sent from i3/sway to the client after the client did IPC_SUBSCRIBE.

ianyfan: Any further thoughts on:
1. removing or implementing `IPC_EVENT_INPUT`;
2. removing or implementing `IPC_EVENT_MODIFIER`?

RyanDwyer: I would think that these events need to be preserved/implemented, unless there's a protocol which does their job. Ping @SirCmpwn

SirCmpwn: We should keep `IPC_EVENT_INPUT` around. Does anyone know what `IPC_EVENT_MODIFIER` is used for? Also, let's assign these to '31' and '30', no need to be shy.

ianyfan: I'm actually thinking 100+ to match the non-event sway-specific values. Also, it seems that `IPC_EVENT_MODIFIER` is used to indicate when a bar modifier is pressed/released (which doesn't seem to be implemented in 1.0 yet...)

SirCmpwn: Let's drop `IPC_EVENT_MODIFIER`.

ianyfan: Ideas on what to use `IPC_EVENT_INPUT` for? Maybe I can implement it. Or leave it for now?

SirCmpwn: Actually, no, I was confusing it with `IPC_GET_INPUTS` which we should keep. We can drop `IPC_EVENT_INPUT` too.

ianyfan:
Rebased with aforementioned fixes.

RyanDwyer:
Thanks!

PR #2265, opened 2018-07-13
Implement swaylock configuration file parsing

### LINKED ISSUE #1962: Add swaylock configuration file
Should work similar to the mpv configuration, with valid command line arguments also being valid key/value pairs in the config file.

RedSoxFan:
Closes #1962 

Allows for `swaylock` to be configured from file. All long options are valid using the format `long-option[=value]`. Ex. `color=111111FF` or `ignore-empty-password`. If requested, I can add support for short options, but they are currently not supported in the config file.

The default configuration file paths are `$HOME/.swaylock/config`, `$XDG_CONFIG_HOME/swaylock/config`, and `SYSCONFDIR/swaylock/config`. The configuration file can also be manually specified using `-C, --config`.

I decided that instead of duplicating the code for configuring via the command line and a config file, to just reuse the same code and treat each line of the config file as a single argument for parsing. This makes it so that options do not need to be implemented/parsed in two locations and ensures that when an option is added/changed/removed, it effects both config methods.

Any option set on the command line will override the option read from the config file.

### DIFF
--- a/swaylock/main.c
+++ b/swaylock/main.c
@@ -789,13 +789,16 @@ int main(int argc, char **argv) {
 	wlr_log_init(WLR_DEBUG, NULL);
 
 	char *config_path = NULL;
-	for (int i = 0; i < argc; i++) {
-		if (strcmp(argv[i], "-C") == 0 || strcmp(argv[i], "--config") == 0) {
-			if (i + 1 == argc) {
-				wlr_log(WLR_ERROR, "Config file path is missing");
-				return 1;
-			}
-			config_path = strdup(argv[i + 1]);
+	static struct option long_options[] = {
+		{"config", required_argument, NULL, 'C'},
+		{0, 0, 0, 0},
+	};
+	while (1) {
+		int c = getopt_long(argc, argv, "C:", long_options, NULL);
+		if (c == -1) {
+			break;
+		} else if (c == 'C') {
+			config_path = strdup(optarg);
 			break;
 		}
 	}
--- a/swaylock/main.c
+++ b/swaylock/main.c
@@ -21,6 +21,7 @@
 #include "pool-buffer.h"
 #include "cairo.h"
 #include "log.h"
+#include "readline.h"
 #include "stringop.h"
 #include "util.h"
 #include "wlr-input-inhibitor-unstable-v1-client-protocol.h"
@@ -412,15 +413,14 @@ static void set_default_colors(struct swaylock_colors *colors) {
 	};
 }
 
-static struct swaylock_state state;
+enum line_mode {
+	LM_LINE,
+	LM_INSIDE,
+	LM_RING,
+};
 
-int main(int argc, char **argv) {
-	enum line_mode {
-		LM_LINE,
-		LM_INSIDE,
-		LM_RING,
-	};
-
+static int parse_options(int argc, char **argv, struct swaylock_state *state,
+		enum line_mode *line_mode) {
 	enum long_option_codes {
 		LO_BS_HL_COLOR = 256,
 		LO_FONT,
@@ -447,6 +447,7 @@ int main(int argc, char **argv) {
 	};
 
 	static struct option long_options[] = {
+		{"config", required_argument, NULL, 'C'},
 		{"color", required_argument, NULL, 'c'},
 		{"ignore-empty-password", no_argument, NULL, 'e'},
 		{"daemonize", no_argument, NULL, 'f'},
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Nice!

swaylock/main.c
```diff
@@ -621,78 +614,212 @@ int main(int argc, char **argv) {
...
+	return 0;
+}
+
+static bool file_exists(const char *path) {
+	return path && access(path, R_OK) != -1;
+}
+
+static char *get_config_path(void) {
+	static const char *config_paths[] = {
+		"$HOME/.swaylock/config",
```
emersion: Hmm, not a fan of this one

swaylock/main.c
```diff
@@ -621,78 +614,212 @@ int main(int argc, char **argv) {
...
+			strcat(config_home, "/.config");
+			setenv("XDG_CONFIG_HOME", config_home, 1);
+			wlr_log(WLR_DEBUG, "Set XDG_CONFIG_HOME to %s", config_home);
+			free(config_home);
+		}
+	}
+
+	wordexp_t p;
+	char *path;
+	for (int i = 0; i < (int)(sizeof(config_paths) / sizeof(char *)); ++i) {
```
emersion: `size_t` can be used here

swaylock/main.c
```diff
@@ -621,78 +614,212 @@ int main(int argc, char **argv) {
...
+		.ignore_empty = false,
+		.show_indicator = true,
+	};
+	wl_list_init(&state.images);
+	set_default_colors(&state.args.colors);
+
+	wlr_log_init(WLR_DEBUG, NULL);
+
+	char *config_path = NULL;
+	for (int i = 0; i < argc; i++) {
```
emersion: Instead, `parse_options` could be a more simple `parse_option` function which parses a single option key/value.

swaylock/main.c
RedSoxFan: The path? I can remove it. I just followed the same path formats that sway uses.

swaylock/main.c
RedSoxFan: That increases the complexity of command line arguments since `--color <rrggbb[aa]>` (which is two arguments) could be given instead of `--color=<rrggbb[aa]>`. I guess I could make the `=` mandatory for command line arguments or detect whether there is a `=` in the argument.

emersion: Eh, doesn't getopt handle that for you? We should stick to getopt's behaviour.

emersion: (I was thinking of something like `parse_option(state, key, value)`)

RedSoxFan: After discussing on IRC, we decided to keep it as is. @emersion's complaint was with the manual `-C, --config` handling. Since the config file needs to be loaded first and then have command line overrides, we agreed to keep it as is.

swaylock/main.c
RedSoxFan: On IRC, we agreed to keep the path for now to maintain consistency with other tools.

swaylock/main.c
```diff
@@ -716,7 +716,7 @@ static char *get_config_path(void) {
 
 	wordexp_t p;
 	char *path;
-	for (int i = 0; i < (int)(sizeof(config_paths) / sizeof(char *)); ++i) {
+	for (size_t i = 0; i < (int)(sizeof(config_paths) / sizeof(char *)); ++i) {
```
emersion: You can now remove that `(int)` cast

swaylock/main.c
```diff
@@ -716,7 +716,7 @@ static char *get_config_path(void) {
 
 	wordexp_t p;
 	char *path;
-	for (size_t i = 0; i < (int)(sizeof(config_paths) / sizeof(char *)); ++i) {
+	for (size_t i = 0; i < (sizeof(config_paths) / sizeof(char *)); ++i) {
```
emersion: You can now remove these extra parentheses ;)

RedSoxFan:
I changed the config flag handling to use a small `getopt_long` block to allow for `-C/path/to/config` and `--config=/path/to/config` (and possibly even others) to be handled correctly.

emersion:
Thanks!

Should we have something similar for `swaybg`?

RedSoxFan:
> Should we have something similar for swaybg?

Sounds good to me. I think it makes sense to wait for your swaybg refactor PR (#2239) to be merged first. Or if you want to include it in that PR, that would be fine too.

SirCmpwn:
Hmm, I want to pump the brakes a bit on refactoring swaybg. I like that it's configured in the output command.

PR #2267, opened 2018-07-13
Simplify popup_unconstrain

### LINKED ISSUE #2259: xdg-positioner issues with GTK3 at the rightmost edges of screens
If I open a context menu on a GTK3 app (tested: gedit and nautilus) which would go over the right edge of the output if it weren't repositioned, the menu doesn't open at all.

emersion:
Just use the parent output.

Fixes #2259


SirCmpwn:
Beautiful

PR #2268, opened 2018-07-13
Enable borders on floating SSD xdg-shell views

emersion:
- [x] View is created after decoration mode is sent


SirCmpwn:
Bump

emersion:
GTK doesn't send the decoration mode at all. I think the KDE folks fixed it, but it hasn't yet been released.

emersion:
Works with `mpv`.

SirCmpwn:
Thanks!

PR #2269, opened 2018-07-13
swaybar/bg: Fix crash on DPMS off

minus7:
When turning off displays via DPMS, swaybar and swaybg still tried to
render, but did not get a valid buffer, causing them to crash.

Not sure if that's a proper fix for the problem, but it does work


emersion:
This is a good fix, because `get_next_buffer` might fail and we should handle it. However I'm not sure why it does fail in this situation…

emersion:
Thanks!

PR #2271, opened 2018-07-14
Implement tap_button_map for input devices

### LINKED ISSUE #1569: [Feature Request] LibInput Tap Button Remapping
This is something that is available in LibInput under Xorg (see https://wiki.archlinux.org/index.php/Libinput#Button_re-mapping) but to the best of my knowledge it's not available in Sway. 

To be more specific - I'd like to be able to configure my taps so that 1 finger taps are left click, 2 finger taps are middle click, and 3 finger taps are right click (currently 2 and 3 are swapped). 

Perhaps an easier version of #1514 😄

RedSoxFan:
Closes #1569 

Implements `input <identifier> tap_button_map lmr|lrm`. This allows for the number of fingers used to be remapped to different buttons when using tap to click. There are only two options allowed by libinput: `lmr` (1=left, 2=middle, 3=right) and `lrm` (1=left, 2=right, 3=middle).

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -210,6 +210,7 @@ sway_cmd input_cmd_repeat_rate;
 sway_cmd input_cmd_scroll_button;
 sway_cmd input_cmd_scroll_method;
 sway_cmd input_cmd_tap;
+sway_cmd input_cmd_tap_button_map;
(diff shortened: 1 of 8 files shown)


SirCmpwn:
Thanks!

PR #2272, opened 2018-07-14
Simplify transactions by using a dirty flag on containers

RyanDwyer:
This PR changes the way we handle transactions to a more simple method. The new method is to mark containers as dirty from low level code (eg. arranging, or `container_destroy`, and eventually `seat_set_focus`), then call `transaction_commit_dirty` which picks up those containers and runs them through a transaction. The old methods of using transactions (`arrange_and_commit`, or creating one manually) are now no longer possible.

The highest-level code (`execute_command` and view implementation handlers) will call `transaction_commit_dirty`, so most other code just needs to set containers as dirty. This is done by arranging, but can also be done by calling `container_set_dirty`.

This PR will make transactionising the focus stack much easier as I will be able to mark containers as dirty from within `seat_set_focus`. It solves the problem of low level code needing to conditionally add a container to a transaction.

This implementation is nicer on the CPU because it doesn't have to check if a container already exists in a transaction when adding one. It comes straight from the `server.dirty_containers` list which doesn't contain duplicates.

Running multiple commands through swaymsg will now use a single transaction.

And lastly, this PR has a negative net LOC :)

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -47,7 +47,7 @@ struct sway_transaction_instruction {
 	bool ready;
 };
 
-struct sway_transaction *transaction_create() {
+static struct sway_transaction *transaction_create() {
 	struct sway_transaction *transaction =
 		calloc(1, sizeof(struct sway_transaction));
 	transaction->instructions = create_list();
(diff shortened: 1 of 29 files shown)


SirCmpwn:
Great :smile:

sway/desktop/transaction.c
```diff
@@ -19,14 +19,14 @@
...
-#define TIMEOUT_MS 200
+#define TIMEOUT_MS 1000
 
 /**
  * If enabled, sway will always wait for the transaction timeout before
  * applying it, rather than applying it when the views are ready. This allows us
  * to observe the rendered state while a transaction is in progress.
  */
-#define TRANSACTION_DEBUG false
+#define TRANSACTION_DEBUG true
```
SirCmpwn: :hankey:

RyanDwyer: Whoops. I'll replace my commit.

SirCmpwn:
LGTM aside from the one issue

SirCmpwn:
Thanks!

PR #2273, opened 2018-07-14
Add extended debugging flags

SirCmpwn:
We currently have several ways of setting debug flags, including command
line arguments, environment variables, and compile-time macros. This
replaces the lot with command line flags.

### DIFF
--- a/include/sway/debug.h
+++ b/include/sway/debug.h
@@ -1,7 +1,15 @@
 #ifndef SWAY_DEBUG_H
 #define SWAY_DEBUG_H
 
+// Tree
 extern bool enable_debug_tree;
 void update_debug_tree();
+
+// Damage
+extern const char *damage_debug;
+
+// Transactions
+extern int txn_timeout_ms;
+extern bool txn_debug;
 
 #endif
--- a/sway/main.c
+++ b/sway/main.c
@@ -251,6 +251,18 @@ static void drop_permissions(bool keep_caps) {
 #endif
 }
 
+void enable_debug_flag(const char *flag) {
+	if (strcmp(flag, "render-tree") == 0) {
+		enable_debug_tree = true;
+	} else if (strncmp(flag, "damage=", 7) == 0) {
+		damage_debug = &flag[7];
+	} else if (strcmp(flag, "txn-debug") == 0) {
+		txn_debug = true;
+	} else if (strncmp(flag, "txn-timeout=", 12) == 0) {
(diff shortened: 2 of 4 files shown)


sway/main.c
```diff
@@ -251,6 +251,18 @@ static void drop_permissions(bool keep_caps) {
...
 }
 
+void enable_debug_flag(const char *flag) {
+	if (strcmp(flag, "render-tree") == 0) {
+		enable_debug_tree = true;
+	} else if (strncmp(flag, "damage=", 7) == 0) {
+		damage_debug = &flag[7];
+	} else if (strcmp(flag, "txn-debug")) {
+		txn_debug = true;
+	} else if (strncmp(flag, "txn-timeout", 11) == 0) {
```
emersion: Missing `=`?

include/sway/debug.h
```diff
@@ -1,7 +1,15 @@
 #ifndef SWAY_DEBUG_H
 #define SWAY_DEBUG_H
 
+// Tree
 extern bool enable_debug_tree;
 void update_debug_tree();
 
+// Damage
+extern const char *damage_debug;
```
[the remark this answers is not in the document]
emersion: Well since this is `const char *` this could lead to various issues such as use-after-free…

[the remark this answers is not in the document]
SirCmpwn: It's a pointer to argv, which isn't freed until you exit (and leave user code that could keep accessing it)

This is debug code, it's not supposed to be pretty.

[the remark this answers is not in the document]
emersion: Well, one thing I'd like to do after this is set it at run-time via swaymsg. With the current API it's easy to be tricked into thinking providing a string and free-ing it afterwards is fine.

w/e

sway/main.c
RyanDwyer: Yeah, not sure why GitHub has marked this as outdated. This is still relevant; everything else looks okay to me.

SirCmpwn: Cause it's outdated, I totally fixed that

RyanDwyer: I would expect the fix to be:

    } else if (strncmp(flag, "txn-timeout=", 12) == 0) {
    		txn_timeout_ms = atoi(&flag[12]);

ie. s/txn-timeout/txn-timeout=/ and replacing 11 with 12.

SirCmpwn: oh derp I thought this comment had been referring to two lines up where I forgot to check the return value of strcmp for txn-debug

PR #2275, opened 2018-07-15
Make focus part of transactions

RyanDwyer:
Rather than maintain copies of the entire focus stack, this PR transactionises the focus by introducing two new properties to the container state and using those when rendering.

* `bool focused` means this container has actual focus. Only one container should have this equalling true in its current state.
* `struct sway_container *focus_inactive_child` points to the immediate child that was most recently focused (eg. for tabbed and stacked containers).

While implementing this PR I discovered `transaction_commit_dirty` needs to be called even earlier than in `execute_command`, so it's now called from IPC, bindsym handling, etc. The reason for this is that a mapping view with criteria would call `execute_command` which created a second/nested transaction, which when combined with the new focus code making containers dirty would ultimately cause rendering issues. This isn't a problem in the simplify-transactions PR because the focus code in that PR doesn't make containers dirty.

Because transactions must be used when focus is changed, `transaction_commit_dirty` is now called in a few more places, such as when handling cursor motion and buttons.

Another negative net LOC PR :)

To test, use the transaction debug feature:

* Open views and observe that the focus styling is preserved on the previous view until the new one appears.
* Close views and observe that the focus styling is preserved on the closing view until it disappears.
* Create tabbed and stacked containers and open and close views within them. Prior to this PR, it would render something it shouldn't during the transaction.

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -119,17 +119,6 @@ struct sway_container *seat_get_active_child(struct sway_seat *seat,
 		struct sway_container *container);
 
 /**
- * Return the immediate child of container which was most recently focused, with
- * fallback to selecting the child in the parent's `current` (rendered) children
- * list.
- *
- * This is useful for when a tabbed container and its children are destroyed but
- * still being rendered, and we have to render an appropriate child.
- */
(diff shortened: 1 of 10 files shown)


SirCmpwn:
Thanks!

PR #2276, opened 2018-07-15
Implement urgency base functionality

### LINKED ISSUE #2258: Proposal: `urgent enable|disable` command
You would execute it with criteria:

* `[app_id="foo"] urgent enable`
* `[mark="foo"] urgent disable`

An example use case would be a chat program which doesn't support urgency or requesting focus but provides desktop notifications. Your notification daemon could, upon receiving a notification from this chat program, execute `swaymsg [some criteria] urgent enable`.

Another example is having a background script monitoring your GitHub notifications or email inbox for new messages, then when it gets a result it sets an appropriate browser window to urgent. If the notifications/emails are read from another location then the script can run `urgent disable` to clear the urgency.

Another example is configuring your shell to mark the window as urgent when it completes a long running command. If your terminal supports "urgent on bell", you can do this already by echoing a bell character, but using a sway command would work for all terminals. However, this would require the shell to determine its container ID so it can use it in the criteria, and I'm not sure how you'd do that.

Lastly, it would simply be useful to test the urgency implementation.

RyanDwyer:
Introduces a command to manually set urgency, as well as rendering of urgent views, sending the IPC event, removing urgency one second after focus, and matching urgent views via criteria.

To test, run something like `swaymsg [class="Firefox"] urgent enable`.

Try it with Firefox on a different workspace too, and notice the swaybar indicator working.

Note that urgency is not yet implemented when xwayland views request it, nor when a view requests focus (depending on the value of `focus_on_window_activation`). I feel like that's out of scope for this PR.

Closes #2258.

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -46,6 +46,31 @@ static int regex_cmp(const char *item, const pcre *regex) {
 	return pcre_exec(regex, NULL, item, strlen(item), 0, 0, NULL, 0);
 }
 
+static int cmp_urgent(const void *_a, const void *_b) {
+	struct sway_view *a = *(void **)_a;
+	struct sway_view *b = *(void **)_b;
+
+	if (a->urgent.tv_sec < b->urgent.tv_sec) {
+		return -1;
+	} else if (a->urgent.tv_sec > b->urgent.tv_sec) {
+		return 1;
+	}
+	if (a->urgent.tv_nsec < b->urgent.tv_nsec) {
+		return -1;
+	} else if (a->urgent.tv_nsec > b->urgent.tv_nsec) {
+		return 1;
+	}
+	return 0;
+}
+
+static void find_urgent_iterator(struct sway_container *swayc, void *data) {
+	if (swayc->type != C_VIEW || !view_is_urgent(swayc->sway_view)) {
+		return;
+	}
+	list_t *urgent_views = data;
+	list_add(urgent_views, swayc->sway_view);
+}
+
 static bool criteria_matches_view(struct criteria *criteria,
 		struct sway_view *view) {
 	if (criteria->title) {
@@ -133,8 +158,23 @@ static bool criteria_matches_view(struct criteria *criteria,
 	}
 
 	if (criteria->urgent) {
-		// TODO
-		return false;
+		if (!view_is_urgent(view)) {
+			return false;
+		}
+		list_t *urgent_views = create_list();
+		container_for_each_descendant_dfs(&root_container,
+				find_urgent_iterator, urgent_views);
+		list_stable_sort(urgent_views, cmp_urgent);
+		struct sway_view *target;
+		if (criteria->urgent == 'o') { // oldest
+			target = urgent_views->items[0];
+		} else { // latest
+			target = urgent_views->items[urgent_views->length - 1];
+		}
+		list_free(urgent_views);
+		if (view != target) {
+			return false;
+		}
 	}
 
 	if (criteria->workspace) {
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -595,15 +595,20 @@ void view_unmap(struct sway_view *view) {
 		view->urgent_timer = NULL;
 	}
 
+	struct sway_container *parent;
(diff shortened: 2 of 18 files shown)


sway/tree/view.c
```diff
@@ -1047,3 +1053,27 @@ bool view_is_visible(struct sway_view *view) {
...
+		view->urgent = (struct timespec){ 0 };
+		if (view->urgent_timer) {
+			wl_event_source_remove(view->urgent_timer);
+			view->urgent_timer = NULL;
+		}
+	}
+	container_damage_whole(view->swayc);
+
+	struct sway_container *ws = container_parent(view->swayc, C_WORKSPACE);
+	ipc_event_workspace(ws, NULL, "urgent");
```
ianyfan: The arguments should be `NULL, ws, "urgent"` since the prototype has the old workspace first, then the current ("new") workspace. Also, do you send an event for the window too?

SirCmpwn:
Overall this LGTM

RyanDwyer:
I just realised that implementing the xwayland urgency hint was really easy, so I went and did that.

I also had a look at `focus_on_window_activation` but I think some changes would need to be made in wlroots (eg. a `request_activate` event).

RedSoxFan:
1. Splitting is working again after your last commit (woot)
2. I can't seem to get `urgent enable` to work for any floating view (it works fine for tiling)
3. If you close a window without focusing it while it's urgency flag is set, the workspace retains the urgent flag.
4. Should `urgent deny` be extended to prevent `urgent enable` or even something like `urgent block` added as a stricter option?
5. Should the urgency flag be propagated upwards for nested containers? I'd have to check i3's behavior on this

RyanDwyer:
> 2. I can't seem to get urgent enable to work for any floating view (it works fine for tiling)

Fixed. This was a problem affecting all criteria. `container_for_each_descendant_dfs` did not descend into floating views. It does now.

> 3. If you close a window without focusing it while it's urgency flag is set, the workspace retains the urgent flag.

Fixed. The same issue occurred when moving urgent views to new workspaces. When reviewing this PR, start by looking at `workspace_detect_urgent`. The rest is mostly calling that function in various places.

> 4. Should urgent deny be extended to prevent urgent enable or even something like urgent block added as a stricter option?

I don't think so. The `deny` setting is there to prevent a noisy client from claiming unwanted urgency. `swaymsg urgent enable` is only run by the user or via a script of their's and should not be affected by the `deny` setting. 

> 5. Should the urgency flag be propagated upwards for nested containers? I'd have to check i3's behavior on this

I can't check this right now, so I'll either wait for your response or I'll check it later.

RedSoxFan:
> > 5. Should the urgency flag be propagated upwards for nested containers? I'd have to check i3's behavior on this
>
> I can't check this right now, so I'll either wait for your response or I'll check it later.

Yes, i3 does propagate the urgent flag up the tree.

--

Aside from the urgency propagation, this LGTM.

sway/criteria.c
```diff
@@ -46,6 +46,31 @@ static int regex_cmp(const char *item, const pcre *regex) {
 	return pcre_exec(regex, NULL, item, strlen(item), 0, 0, NULL, 0);
 }
 
+static int cmp_urgent(const void *_a, const void *_b) {
+	struct sway_view *a = *(void **)_a;
```
ascent12: Minor point, but don't cast away the const here.

RyanDwyer: I copied this from `container_sort_workspaces`. When I tried it without the cast something weird was happening and it wasn't comparing them properly.

RedSoxFan:
Just needs the urgent border colors set when there is an urgent descendant and this should be good

SirCmpwn:
There's an off-by-one error somewhere

SirCmpwn:
Actually I think it's because the urgent colors aren't correctly pulled out of the bar config in `ipc.c`

RyanDwyer:
Fixed. Swaybar was ignoring what IPC gave it and was using default colours. This would only be noticeable if you had an `urgent_workspace_border` colour that differed from the default.

SirCmpwn:
Sweet.

PR #2277, opened 2018-07-15
Fix config buffer overflow and logic

ianyfan:
When I originally wrote the code to read the config file, I set the struct variable early and used the null check to see whether it was reading the main config or an included config. Later on, I changed my mind so that it would attach it late, which negated the null check without me realising, leading to a whole bunch of problems.

Hopefully, this fixes that, as well as making the logic a bit better. I also added in a check so that if somehow, improbably, it reads in more than it thought it needed, it just aborts. I thought about resizing the buffer but was still a bit paranoid about race conditions.

Would appreciate feedback.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -560,8 +560,8 @@ static char *expand_line(const char *block, const char *line, bool add_brace) {
 
 bool read_config(FILE *file, struct sway_config *config) {
 	bool reading_main_config = false;
-	char *this_config = NULL, *config_pos;
-	long config_size = 0;
+	char *this_config = NULL;
+	unsigned long config_size = 0;
 	if (config->current_config == NULL) {
 		reading_main_config = true;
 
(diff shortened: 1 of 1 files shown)


sway/config.c
```diff
@@ -560,16 +560,16 @@ static char *expand_line(const char *block, const char *line, bool add_brace) {
 
 bool read_config(FILE *file, struct sway_config *config) {
 	bool reading_main_config = false;
-	char *this_config = NULL, *config_pos;
-	long config_size = 0;
+	char *this_config = NULL;
+	unsigned long config_size = 0;
```
emersion: Just make this a `size_t`

sway/config.c
```diff
@@ -560,16 +560,16 @@ static char *expand_line(const char *block, const char *line, bool add_brace) {
...
 	if (config->current_config == NULL) {
 		reading_main_config = true;
 
 		fseek(file, 0, SEEK_END);
 		config_size = ftell(file);
```
emersion: While you're at it, can you add error handling for this?

ianyfan: Added some, though not sure if it's good.

ianyfan:
Note: not sure if instead this could be implemented by instead of getting the size first it just starts with a fixed size and resizes as more of the config is read. This could even allow included files to be expanded, if the `read` variable is moved outside of the function to keep track of it across function calls.

Technically outside of the scope of this PR though.

SirCmpwn:
Thanks!

ianyfan:
Np, sorry about messing it up the first time round, hopefully I didn't make any more mistakes this time

PR #2279, opened 2018-07-15
Sorted binding lists

frsfnrrg:
The motivation for this pull request is that binding lookup was O(n) per key press, and duplicated some of the code involved in checking for duplicate bindings. This PR ensures that the lists of bindings remain sorted, so that `get_active_binding` can use binary search instead of a linear scan.

As a rule of thumb, for low-cost comparisons, binary search is faster than linear search around 32 elements. With a configuration involving ~60 bindsyms, binding search time drops from 4us to 3us. This change most likely does slow down modes with <10 bindsyms. but in exchange can handle even 100000 bindsyms in 6us rather than e.g. 2ms. Using e.g. an AVL tree ([tree.c.txt](https://github.com/swaywm/sway/files/2195857/tree.c.txt), [tree.h.txt](https://github.com/swaywm/sway/files/2195858/tree.h.txt)) guarantees O(n log n) loading, but yields slower lookups for <100 bindsyms, and introduces too much complexity.

The PR also resolves a TODO in `sway/tree/workspace.c`, by extracting the first workspace name from bindcode expressions as well.

~~I'm not entirely certain on naming, especially with `list_sortedset_insert` ,  `list_sortedset_find`,  and `find_workspace_name` (e.g. is `extract_binding_workspace_name` clearer?).~~ (Edit: the former two aren't pretty, but one can guess the meaning from the name, and `workspace_name_from_binding` better describes the last one.)

### DIFF
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -20,53 +20,55 @@ void free_sway_binding(struct sway_binding *binding) {
 		return;
 	}
 
-	if (binding->keys) {
-		free_flat_list(binding->keys);
-	}
+	free(binding->keys);
 	free(binding->command);
 	free(binding);
 }
 
+static void *replace_binding(void *old_item, void *new_item) {
+	struct sway_binding *old_binding = (struct sway_binding *)old_item;
+	struct sway_binding *new_binding = (struct sway_binding *)new_item;
+	wlr_log(WLR_DEBUG, "overwriting old binding command '%s' with command '%s'",
+		old_binding->command, new_binding->command);
+	free_sway_binding(old_binding);
+	return new_item;
+}
+
 /**
- * Returns true if the bindings have the same key and modifier combinations.
- * Note that keyboard layout is not considered, so the bindings might actually
- * not be equivalent on some layouts.
+ * Comparison function for bindings.
+ *
+ * Returns -1 if a < b, 0 if a == b, and 1 is a > b
  */
-static bool binding_key_compare(struct sway_binding *binding_a,
-		struct sway_binding *binding_b) {
+int sway_binding_cmp(const void *a, const void *b) {
+	const struct sway_binding *binding_a = (const struct sway_binding *)a;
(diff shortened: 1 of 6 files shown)


sway/commands/bind.c
```diff
@@ -20,53 +20,55 @@ void free_sway_binding(struct sway_binding *binding) {
...
 
-	if (binding_a->bindcode != binding_b->bindcode) {
-		return false;
+	if (binding_a->release != binding_b->release) {
+		return binding_a->release < binding_b->release ? -1 : 1;
 	}
 
-	if (binding_a->modifiers ^ binding_b->modifiers) {
-		return false;
+	if (binding_a->modifiers != binding_b->modifiers) {
```
emersion: Replacing that XOR is probably a good thing

emersion:
>I'm not entirely certain on naming, especially with 'list_sortedset_insert , list_sortedset_find, and find_workspace_name (e.g. is extract_binding_workspace_name clearer?).

Maybe we could have a different type for sorted lists, since we use those in quite a few places. Not sure it's worth it.

frsfnrrg:
> Maybe we could have a different type for sorted lists, since we use those in quite a few places. Not sure it's worth it.

It's not worth it -- by the same argument, there'd be `unorderedlist_t`, `orderedlist_t`, `sortedlist_t`, etc. It's enough that `list_t` is a typedef (and even it ought to be renamed to `struct list`). Also, `list_sortedset_insert/find` are worse than linear scan for lists always shorter than ~30 elements, so right now the functions are only useful for the bindings.

SirCmpwn:
Thank you, but the tradeoffs don't make sense here. I'm not going to merge a hundred new lines of code to gain 1us in the common case. No one has 100000 keybindings.

frsfnrrg:
Thanks for the prompt rejection. (In retrospect, the optimization wasn't a good idea.) I'll make a separate PR for the "Also extract first workspace name from bindcodes" bit, since that part actually is useful.

emersion:
Thanks, that would be helpful!

PR #2281, opened 2018-07-16
Send clicks to swaybar blocks as X11 button ids

pvsr:
Currently click events are sent to blocks with a linux event code as the button value, but the [i3bar protocol](https://i3wm.org/docs/i3bar-protocol.html) specifies that the button should be an X11 button id.

### DIFF
--- a/swaybar/i3bar.c
+++ b/swaybar/i3bar.c
@@ -219,15 +219,15 @@ void i3bar_block_send_click(struct status_line *status,
 
 enum x11_button wl_button_to_x11_button(uint32_t button) {
 	switch (button) {
-	case (BTN_LEFT):
+	case BTN_LEFT:
 		return LEFT;
-	case (BTN_MIDDLE):
+	case BTN_MIDDLE:
 		return MIDDLE;
-	case (BTN_RIGHT):
+	case BTN_RIGHT:
 		return RIGHT;
-	case (BTN_SIDE):
+	case BTN_SIDE:
 		return BACK;
-	case (BTN_EXTRA):
+	case BTN_EXTRA:
 		return FORWARD;
 	default:
 		return NONE;


pvsr:
Actually I've got a better idea for how to do this that handles scroll events as well, ignore this PR for now.

swaybar/i3bar.c
```diff
@@ -215,3 +216,26 @@ void i3bar_block_send_click(struct status_line *status,
...
+	case (BTN_RIGHT): return RIGHT;
+	case (BTN_SIDE): return BACK;
+	case (BTN_EXTRA): return FORWARD;
+	default: return NONE;
+	}
+}
+
+enum x11_button wl_axis_to_x11_button(uint32_t axis, wl_fixed_t value) {
+	switch (axis) {
+	case 0:
```
emersion: You can use `WL_POINTER_AXIS_{VERTICAL,HORIZONTAL}_SCROLL`

swaybar/i3bar.c
```diff
@@ -215,3 +216,26 @@ void i3bar_block_send_click(struct status_line *status,
 	}
 	json_object_put(event_json);
 }
+
+enum x11_button wl_button_to_x11_button(uint32_t button) {
+	switch (button) {
+	case (BTN_LEFT): return LEFT;
```
emersion: Style:

```c
switch (thing) {
case THING:
    return OTHER_THING;
…
}
```

swaybar/bar.c
```diff
@@ -147,19 +147,34 @@ static void wl_pointer_button(void *data, struct wl_pointer *wl_pointer,
...
+		if (x >= hotspot->x
+				&& y >= hotspot->y
+				&& x < hotspot->x + hotspot->width
+				&& y < hotspot->y + hotspot->height) {
+			hotspot->callback(output, pointer->x, pointer->y,
+					wl_axis_to_x11_button(axis, value), hotspot->data);
+		}
+	}
+
+	// TODO only scroll workspaces if pointer isn't over a block
```
emersion: Can't you `return` if you execute a `hotspot->callback`?

pvsr: That does work, thanks!

swaybar/i3bar.c
```diff
@@ -215,3 +216,32 @@ void i3bar_block_send_click(struct status_line *status,
 	}
 	json_object_put(event_json);
 }
+
+enum x11_button wl_button_to_x11_button(uint32_t button) {
+	switch (button) {
+	case (BTN_LEFT):
```
emersion: Style: no need for parentheses here

PR #2282, opened 2018-07-16
Fix tab split focus bug

RyanDwyer:
Fixes a bug where if you have a tab containing a split, then switch from a non-split tab to the split tab, focus is not changed properly.

To test, create layout `T[view, V[view, view focused]]` then `focus left` then `focus right`.


emersion:
Thanks!

PR #2283, opened 2018-07-16
Implement no_focus command

RyanDwyer:
i3 docs: https://i3wm.org/docs/userguide.html#no_focus

Example config: `no_focus [class="Firefox"]`

Note that if you've focused an empty workspace when the view maps, the view should receive focus regardless of any `no_focus` directive.


SirCmpwn:
Thanks!

PR #2286, opened 2018-07-17
Implement default_floating_border command and adjust CSD behaviour

RyanDwyer:
i3 docs: https://i3wm.org/docs/userguide.html#_default_border_style_for_new_windows

To test, you need to use something which floats by default and doesn't have CSD:

* Have `default_floating_border pixel 10` in config
* Open Firefox and press `Ctrl+O` to open the file open dialog

The existing code in `view_set_tiled` that set the border to `config->border` was causing issues. I've changed this so the border values are untouched, and it instead uses a `bool using_csd` in the view to determine whether the border values should be respected or not. This allows the previous border config to be reinstated easily when toggling between floating and tiling, and also means the user can't run an IPC command to mess with the borders while CSD is active.

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -69,6 +69,7 @@ struct sway_view {
 	bool border_bottom;
 	bool border_left;
 	bool border_right;
+	bool using_csd;
 
 	struct timespec urgent;
 	bool allow_request_urgent;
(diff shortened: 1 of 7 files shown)


RedSoxFan:
Thanks!

PR #2287, opened 2018-07-17
Remove superfluous IPC urgent events

RyanDwyer:
When an xwayland view is mapped, the IPC urgent event ("this is/is-not urgent") was being sent on every surface commit.

I had intentionally ommitted the check because I figured an urgent surface could update its urgent timestamp by sending urgent a second time. But that's not how it works in xwayland's case, and it makes for more complicated code.


SirCmpwn:
Thanks!

PR #2289, opened 2018-07-17
Fix memory leaks and reference to uninitialized 

frsfnrrg:
Found using Cppcheck and AddressSanitizer. (FYI, Cppcheck also indicates

     [sway/common/stringop.c:150]: (error) Common realloc mistake: 'argv' nulled but not freed upon failure

but the file is full of unchecked allocations anyway.)


SirCmpwn:
Thanks!

PR #2294, opened 2018-07-18
Fix swaylock arguments

RedSoxFan:
I messed up the handling of arguments in `swaylock` with 7b91712416953c25f2722fedb57b474504460a48. The issue is that I only gave `C:` to `getopt_long` when getting searching for the config path. This caused the `getopt_long` to output that every other flag was invalid. Since the swaylock itself was never aborting, the arguments should still be working (aside from `-i, --image` -- see below).

I decided to alter `parse_options` to only handling options when a valid pointer is given. This makes it so the option string does not have to be updated in two places. However, it does make function longer. I think in the long run, this is still the better option.

As for `-i, --image`, only the first image for an output was used. This means if you have `image` in the config and `-i, --image` as an argument, the one in the config would be used instead of overridden. I decided to make it so the last image given is used and replaces the previous definition. This only effects the default image or images defined for the same output. Specifying on in the config and one as an argument should work fine.

### DIFF
--- a/swaylock/main.c
+++ b/swaylock/main.c
@@ -345,20 +345,23 @@ static void load_image(char *arg, struct swaylock_state *state) {
 		image->path = strdup(arg);
 	}
 
-	bool exists = false;
-	struct swaylock_image *iter_image;
-	wl_list_for_each(iter_image, &state->images, link) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #2300, opened 2018-07-18
Handle xwayland override_redirect flag change

emersion:
This fixes syncplay menus.

~~… and breaks dmenu.~~


emersion:
All right, I've been testing this with:

* dmenu
* Chromium (right click menu + app menu + bar menu)
* Firefox (right click menu + app menu + bar menu)
* GTK+ under xwayland
* syncplay
* Steam (menus are unclickable but that's a different issue reproducible on master)
* VLC under xwayland
* qutebrowser under xwayland

All of these work. Please test and report any misbehaving xwayland app wrt. menus/popups/etc.

PR #2305, opened 2018-07-19
Implement focus output command

### LINKED ISSUE #2297: Default focused output
I'm using sway version 1.0-alpha.3-299-gf516dbfb with two outputs (laptop and external).
The external is my main output with workspace 1 assigned to it.

However when starting sway the laptop output is the focused one (while I want the external one to be focused).
I tried to add to the configuration `focus output DP-1`, but it didn't help.

Primary flag for output might help but I didn't found anything in the docs.

Any ideas?

RyanDwyer:
Closes #2297.


sway/commands/focus.c
```diff
@@ -44,7 +46,47 @@ static struct cmd_results *focus_mode(struct sway_container *con,
...
+	if (!argc) {
+		return cmd_results_new(CMD_INVALID, "focus",
+			"Expected 'focus output <direction|name>'");
+	}
+	char *identifier = join_args(argv, argc);
+	struct sway_container *output = output_by_name(identifier);
+
+	if (!output) {
+		enum movement_direction direction;
+		if (strcmp(identifier, "left") == 0) {
```
emersion: You can just use `parse_movement_direction` here

emersion:
Thanks!

PR #2309, opened 2018-07-19
Allow xwayland views to become urgent when on a non-visible workspace

RyanDwyer:
Depends on https://github.com/swaywm/wlroots/pull/1147

This removes the urgency stuff from the commit handler and puts it in a new `set_hints` handler instead. This allows the xwayland surface to become urgent without having to commit (which doesn't happen if it's on a non-visible workspace).

To test, open an xwayland terminal, ensure it allows urgent on bell, then run something like `sleep 2 && echo -en '\a'`, then switch to a different workspace on the same output before the sleep finishes.


emersion:
Thanks!

PR #2310, opened 2018-07-19
Implement assign to output

RyanDwyer:
Eg. `assign [class="Firefox"] output foo`


SirCmpwn:
File a follow-up ticket for spotify

bpinto:
I think that is a bug on spotify, check this issue: https://github.com/i3/i3/issues/2060

PR #2313, opened 2018-07-19
swaybar: Fix scroll handling on workspace buttons

minus7:
As well as ignoring scroll events on status elements when click_events
is enabled.
Previously, using the scroll wheel on a workspace button would switch to
that workspace instead of scrolling through them. Clicks and scrolling
on status elements would always be processed by swaybar, too. So in case
you were using scrolling as volume control on a status item, swaybar
would additionally scroll through your workspaces.

Note: I'm unsure about the enum naming. It doesn't seem to express the intent well enough, open to suggestions.

### DIFF
--- a/include/swaybar/bar.h
+++ b/include/swaybar/bar.h
@@ -29,10 +29,15 @@ enum x11_button {
 	FORWARD,
 };
 
+enum hotspot_event_handling {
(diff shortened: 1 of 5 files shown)


SirCmpwn:
Thanks!

PR #2317, opened 2018-07-20
Implement force_display_urgency_hint

RyanDwyer:
The directive sets the timeout before an urgent view becomes normal again after switching to it from another workspace.

Docs: https://i3wm.org/docs/userguide.html#_delaying_urgency_hint_reset_on_workspace_change

Also:

* When an xwayland surface removes the urgent hint while the timer is active, we now ignore the request. This happens as soon as the view receives focus, so it was effectively making the timer pointless.
* The timeout is now only applied when switching to it from another workspace.

### DIFF
--- a/sway/commands/force_display_urgency_hint.c
+++ b/sway/commands/force_display_urgency_hint.c
@@ -1,0 +1,28 @@
+#include "log.h"
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/tree/arrange.h"
+#include "sway/tree/container.h"
+#include "sway/tree/view.h"
+#include "sway/tree/layout.h"
+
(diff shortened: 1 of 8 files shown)


sway/commands/force_display_urgency_hint.c
```diff
@@ -0,0 +1,28 @@
+#include "log.h"
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/tree/arrange.h"
```
RedSoxFan: Not all of these headers are needed.

RedSoxFan:
Sorry if I wasn't clear in my description of i3's behavior for `force_display_urgent_timeout 0`. I meant the urgent state is cleared **immediately after switching to the workspace**. Currently this is clearing it immediately without the workspace switching

RedSoxFan:
Thanks!

PR #2318, opened 2018-07-20
Fix output wildcard handling

### LINKED ISSUE #1724: Reloading config loses output positioning
Instead outputs start behaving like they were added with `_auto(...)`

### LINKED ISSUE #2127: Toggling dpms changes output layout
Run the following: `swaymsg "output * dpms off"; sleep 3; swaymsg "output * dpms on"`.

On resume, my output layout is changed to an automatically generated one. The positioning I have in my config is ignored but transformations are preserved. Running `swaymsg reload` makes it re-apply the positioning from my config.

### LINKED ISSUE #2295: Using output wildcard undoes multimonitor order
Not sure if this is expected behavior.

I took a look at https://github.com/swaywm/sway/blob/master/sway/commands/output.c but couldn't rule out why this happens.

RedSoxFan:
Fixes #1724 
Fixes #2127 
Fixes #2295 
(Any others that I'm missing?)

This PR fixes issues related to `output *`.

Instead of applying the wildcard config directly to the output, the wildcard config is merged on top of the config for the output and that is then applied (falling back to the wildcard config). Additionally, when creating a non-wildcard config, it is merged on top of a copy of the wildcard config (if existent).

Output layout is no longer lost. This was due to an `output *` command being after the `output <name>` commands. When applying the `output *` config, the following snippet was changing it to auto positioning:
```c
if (oc && (oc->x != -1 || oc->y != -1)) {
	wlr_log(WLR_DEBUG, "Set %s position to %d, %d", oc->name, oc->x, oc->y);
	wlr_output_layout_add(output_layout, wlr_output, oc->x, oc->y);
} else {
	wlr_output_layout_add_auto(output_layout, wlr_output);
}
```
Since the `output *` config is never applied directly to the output (unless there is not an output config for the output), the positions in the config should be respected.

Setting the dpms state on a specific output should also work now. Before it would only work if there was a background specified by the output due to the following block changing the config before applying dpms:
```c
if (!oc || !oc->background) {
	// Look for a * config for background
	int i = list_seq_find(config->output_configs, output_name_cmp, "*");
	if (i >= 0) {
		oc = config->output_configs->items[i];
	} else {
		oc = NULL;
	}
}
/// DPMS applied below this point
```

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -461,6 +461,8 @@ void merge_output_config(struct output_config *dst, struct output_config *src);
 void apply_output_config(struct output_config *oc,
 		struct sway_container *output);
 
+struct output_config *store_output_config(struct output_config *oc);
+
 void free_output_config(struct output_config *oc);
 
 int workspace_output_cmp_workspace(const void *a, const void *b);
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -21,6 +21,60 @@ static struct cmd_handler output_handlers[] = {
 	{ "transform", output_cmd_transform },
 };
 
+static struct output_config *get_output_config(char *name, char *identifier) {
+	int i = list_seq_find(config->output_configs, output_name_cmp, name);
+	if (i >= 0) {
+		return config->output_configs->items[i];
+	}
+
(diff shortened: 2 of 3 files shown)


emersion:
Thanks!

PR #2320, opened 2018-07-21
Reset outputs on reload

### LINKED ISSUE #1752: Reloading config should reset scale, mode, … if unspecified
At runtime, not specifying an output option means "don't change it". On reload, not specifying an output option should mean "reset it to the default".

### LINKED ISSUE #2293: Freeze when reloading config
When reloading sway (meta+shift+c), screen stops updating. I am also unable to switch to a TTY with ctrl+alt+2. From another machine (via ssh), I can pkill -f sway at which point I get kicked back to a TTY.

This began occurring when I added this line to my sway config:
```
output * bg `wallpaper` fill
```
(`wallpaper` is an executable script which returns the path to an image)

I am unable to reproduce this when running sway with the -d switch. This leads me to believe it is a race condition.

Without verbose logging, this is the only output:

```
swaybg: ../swaybg/main.c:215: main: Assertion `state.compositor && state.layer_shell && state.output && state.shm'
 failed.
```

Reproduced on 1.0-alpha.4

RedSoxFan:
Fixes #1752 
Fixes #2293

This PR creates default output configs for connected outputs when reloading. This ensures that when the output configs are applied, if an option is missing it is reset to the default instead of left unchanged.

Additionally, when reloading output configs are now applied after the entire config is read and before the deferred commands. This ensures that all workspace command bindings are read and the output does not incorrectly get assigned an automatic workspace name. This simulates the behavior that sway would have on a fresh start.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -464,7 +464,11 @@ void apply_output_config(struct output_config *oc,
 
 struct output_config *store_output_config(struct output_config *oc);
 
+void apply_output_config_to_outputs(struct output_config *oc);
+
 void free_output_config(struct output_config *oc);
+
+void create_default_output_configs(void);
 
 int workspace_output_cmp_workspace(const void *a, const void *b);
 
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -21,60 +21,6 @@ static struct cmd_handler output_handlers[] = {
 	{ "transform", output_cmd_transform },
 };
 
-static struct output_config *get_output_config(char *name, char *identifier) {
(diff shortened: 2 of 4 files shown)


RedSoxFan:
@jniedrauer If you get a chance, can you please confirm that this fixes #2293? I'm unable to replicate it on this branch.

RyanDwyer:
I just realised you asked jniedrauer to test it right after I merged it. Oh well. Thanks anyway!

RedSoxFan:
> A little late, but yes. I've confirmed that I am no longer able to reproduce the issue.

Thanks for confirming.

PR #2323, opened 2018-07-22
Implement scratchpad

RyanDwyer:
Implements the following commands:

* `move scratchpad`
* `scratchpad show`
* `[criteria] scratchpad show`

Also fixes these:

* Fix memory leak when executing command with criteria (use `list_free(views)` instead of `free(views)`)
* Fix crash when running `move to` with no further arguments

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -48,6 +48,8 @@ struct sway_server {
 
 	list_t *transactions;
 	list_t *dirty_containers;
+
+	list_t *scratchpad; // struct sway_container
 };
 
 struct sway_server server;
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -48,8 +48,6 @@ struct sway_server {
 
 	list_t *transactions;
 	list_t *dirty_containers;
-
-	list_t *scratchpad; // struct sway_container
 };
 
 struct sway_server server;
--- a/sway/scratchpad.c
+++ b/sway/scratchpad.c
@@ -1,0 +1,173 @@
+#define _XOPEN_SOURCE 700
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include "sway/scratchpad.h"
+#include "sway/input/seat.h"
+#include "sway/tree/arrange.h"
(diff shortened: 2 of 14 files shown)


include/sway/server.h
```diff
@@ -48,6 +48,8 @@ struct sway_server {
 
 	list_t *transactions;
 	list_t *dirty_containers;
+
+	list_t *scratchpad; // struct sway_container
```
RedSoxFan: I think this should be part of `struct sway_root` instead of `struct sway_server`

sway/scratchpad.c
```diff
@@ -0,0 +1,173 @@
...
+		// Maybe resize it
+		if (con->width > ws->width || con->height > ws->height) {
+			// TODO: Do this properly once we can float C_CONTAINERs
+			if (con->type == C_VIEW) {
+				view_init_floating(con->sway_view);
+				arrange_windows(con);
+			}
+		}
+
+		// Center it
```
RedSoxFan: This does not match i3's behavior. For views that were tiling that are added to the scratchpad, this does match. For views that were floating (or moved while in the scratchpad), the location is retained relative to a pseudo output.

https://github.com/i3/i3/blob/next/src/scratchpad.c#L248
https://github.com/i3/i3/blob/next/src/floating.c#L926

(I'm fine with this being changed in a follow-up PR)

RyanDwyer: > For views that were floating (or moved while in the scratchpad), the location is retained relative to a pseudo output.

Just confirming, the behaviour you're describing is what the PR currently does, but it should center the container every time it's displayed?

RedSoxFan: So currently this PR centers the view if not in bounds of the output.

In i3, if you have two outputs of the same resolution, the view will always be at the same (x, y) relative to the top-left of the output.

For outputs of different resolutions, the views may extend over the bounds of the output. I'm not exactly sure how the views are being translated on the outputs. I think the second link contains the  function doing the translation.

SirCmpwn:
Nice! Sway 1.0 now officially supports every feature I use in my config file :tada:

Issue #1779, opened 2018-04-08
Simulate input events with sway commands


ddevault:
```
seat "default" cursor move +10,0
seat "default" cursor set 100,100
seat "default" cursor press|release left|right|1|2|3...
seat "default" keyboard press|release H
seat "default" keyboard type "hello world"
```

emersion:
Additional note: if the currently focused client supports text-input, we can use that to type text.

athas:
I have interest in making this time, as well as possibly the time.  However, I have no experience with sway's codebase, or Wayland programming in general.  Can anyone suggest a good point to start looking?

NilsIrl:
Would it also be possible to have virtual seats?

Seats which don't have any input device attached to them and can only be controlled through `sway` commands.

mvdan:
I could almost get this to work with commands like:

```
seat - cursor set 50 50
seat - cursor press button1
seat - cursor release button1
```

If I understand the docs correctly, this should move the cursor at 50x50 and do a full mouse left-click.

However, it doesn't quite work. The mouse does move, but the click happens at the last position the mouse was in. I tried adding sleeps, in between the calls to `swaymsg`, but still the same.

mvdan:
Thinking about it some more, I think what I tried above would work if I was able to use a virtual seat, much like what @NilsIrl suggested.

mvdan:
I was wondering if https://github.com/swaywm/sway/pull/4674 could help here, but I don't see a way to create a virtual pointer.

paulolieuthier:
@emersion forgive my ignorance but why isn't something like `wlr_seat_keyboard_send_key` enough, instead of relying on a text input protocol?

emersion:
This depends on the keymap. If you want to send a Hangul character while having your keyboard configured as en-us, it just won't work.

paulolieuthier:
That makes sense, but I don't think the text input protocol can be used to enter keys like arrows (the [input method](https://github.com/wayland-project/wayland-protocols/blob/b0a25f26d3e8b5ad24773881a9a895dd9cc75466/unstable/input-method/input-method-unstable-v1.xml#L145-L157) protocol has a "keysym" request that looks promising).

[...]

PR #2330, opened 2018-07-23
Configure initial state for numlock/capslock

progandy:
Set numlock state to enabled by default, keep capslock disabled and make this configurable.

The initial state of numlock and capslock can be set with 
```input ... xkb_numlock|xkb_capslock disable|enable```

Implements #2311 

Note: This relies ```wlr_keyboard_notify_modifiers``` which is not meant to be called by the compositor, but I found no other option to reliably set the modifiers. 

I used the xkb_ prefix to indicate that setting the state will reload the keyboard configuration. 
The commands use enabled/disabled instead of off/on to match the other boolean input commands.

### CODE REVIEW

**Comment by SirCmpwn** [comment]
(still missing man page updates)

**Comment by progandy** [comment]
Now everything should be there. I had to clean up my local clone first.

File: sway/sway-input.5.scd [hunk 1]
```diff
@@ -33,6 +33,14 @@ For more information on these xkb configuration options, see
 *input* <identifier> xkb\_variant <variant>
 	Sets the variant of the keyboard like _dvorak_ or _colemak_.
 
+*input* <identifier> xkb\_capslock enabled|disabled
+	Enables or disables CapsLock on sway startup, the default is disabled.
+	Only meaningful if sway controls the keyboard.
+
+*input* <identifier> xkb\_numlock enabled|disabled
+	Enables or disables NumLock on sway startup, the default is enabled.
+	Only meaningful if sway controls the keyboard.
```
SirCmpwn: This line can probably be removed from both.

RedSoxFan:
Since the implementation is not restricted to startup, you could change the man page to `Enables or disables CapsLock/NumLock. The default on startup is disabled/enabled.`

progandy:
I can do that. Should I somehow document that changing the value for NumLock resets CapsLock to the value stored in xkb_capslock as well? Maybe a general note that changing one xkb_ value applies all stored xkb_ values again.

SirCmpwn:
Can't we get the current state of the modifiers and OR/AND them with only the bits we want to flip?

progandy:
The flags are currently set in the keyboard configuration and then applied with ```apply_keyboard_config``` / ```sway_keyboard_configure```. There is no way to know which flag change was requested. I think I'd need a way to do different things when
* sway detects a new keyboard (startup or hotplug)
* swaymsg is sent for an existing keyboard

SirCmpwn:
I see. Let's take the easy way out for now and just make it effective in the config file (not at runtime)?

progandy:
Is there a way to do that?

I think I could create a status flag in the input config that the next call should only update the marked modifier, but that feels like dirty code. There shouldn't be status information in a configuration structure.

SirCmpwn:
Just add it to the appropriate list of commands in commands.c.

RedSoxFan:
I think the runtime portion of this can be added to the keyboard part of #1779

SirCmpwn:
Aye.

progandy:
Somehow I missed that list, it is so short.

Should this stay an individual setting for each keyboard then or simply a global option for all keyboards? I think I'd prefer the global option then.

SirCmpwn:
I like it individual, if you want global do `input * xkb_numlock enable`

progandy:
I can't add a subcommand of input to the config_commands list. I'll mark this as WIP and keep on thinking, maybe I'll get a good idea.

RedSoxFan:
> I can't add a subcommand of input to the config_commands list. I'll mark this as WIP and keep on thinking, maybe I'll get a good idea.

See https://github.com/swaywm/sway/blob/master/sway/commands/bar.c#L38 and
https://github.com/swaywm/sway/blob/master/sway/commands/bar.c#L49

progandy:
I had another idea. What do you think about allowing the swaymsg command and just storing the new value? It will be applied only after hotplugging the keyboard or changing its layout.

SirCmpwn:
Do whatever you want, just get something mergable. This feature is not that complicated.

progandy:
@SirCmpwn, @RedSoxFan, the pull request is ready for review. I kept it restricted to the configuration file for now. Sometimes I try to perfect and polish too much.

PS: I had to rebase it against the current master.

RedSoxFan:
Thanks!

Issue #2336, opened 2018-07-23
Verify sway commands match i3 commands


SirCmpwn:
Find a list of i3 commands and make sure we have all of them in sway (at least all the ones we want to have). Double check that each sway implementation supports the same arguments as the i3 equivalents.

Criteria:
- [x] class
	- [x] \_\_focused__
- [x] instance
	- [x] \_\_focused__
- [x] window_role
	- [x] \_\_focused__
- [x] window_type
	- [x] normal, dialog, utility, toolbar, splash, menu, dropdown_menu, popup_menu, tooltip, notification
- [x] id
- [x] title
	- [x] \_\_focused__
- [x] urgent
	- [x] latest/newest/last/recent (sway: only latest)
	- [x] oldest/first (sway: only oldest)
- [x] workspace
	- [x] \_\_focused__
- [x] con_mark
- [x] con_id
	- [x] \_\_focused__
- [x] floating
- [x] tiling

Commands:
- [x] exec ~~[--no-startup-id]~~
	- [x] command chaining
	- [x] exec quoting
- [x] split
	- [x] split vertical|horizontal|toggle
	- [x] splith
	- [x] splitv
	- [x] splitt
- [x] layout
	- [x] layout default|tabbed|stacking|splitv|splith (sway: default possible but not documented)
	- [x] layout toggle [split|all] (sway: all is ignored)
	- [x] layout toggle [split|tabbed|stacking|splitv|splith] [split|tabbed|stacking|splitv|splith] (sway: anything except split is ignored)
- [x] focus
	- [x] \<criteria> focus (undocumented)
	- [x] focus left|right|down|up
	- [x] focus parent|child|floating|tiling|mode_toggle
	- [x] focus output left|right|up|down|~~primary~~|\<output> (sway: no primary)
	- [x] focus mode_toggle
- [x] move
	- [x] move <left|right|down|up> [\<px> px] (sway: documentation does not specify px can be included)
	- [x] move [absolute] position <pos_x> [px] <pos_y> [px] (sway: absolute ignored)
	- [x] move [absolute] position center (sway: absolute ignored)
	- [x] move position mouse (sway: allows "absolute")
	- [x] move window|container to mark \<mark>
	- [x] move container|workspace to output left|right|down|up|current|~~primary~~|\<output> (sway: no primary; allows "window")
- [x] swap container with id|con_id|mark \<arg>
- [x] sticky enable|disable|toggle
- [x] workspace
[...]

PR #2340, opened 2018-07-23
Switch to using a function to parse booleans

RedSoxFan:
Related to #2336.

This PR adds a function (`parse_boolean`) similar to i3's `eval_boolstr`. The only `true` value that I added was `enabled` since that was in use for various commands. The rest are identical to i3's.

I also added support for toggling values. Currently, toggling is not supported for `input` or `output` commands (since it would require a change to the way their configs work, which is out of scope for this PR), but should work for anything else that takes a boolean value. For `input` and `output` commands, `toggle` is interpreted as a non-true value.

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -124,17 +124,18 @@ uint32_t parse_color(const char *color) {
 }
 
 bool parse_boolean(const char *boolean, bool current) {
-	if (strcmp(boolean, "1") == 0
-			|| strcmp(boolean, "yes") == 0
-			|| strcmp(boolean, "on") == 0
-			|| strcmp(boolean, "true") == 0
-			|| strcmp(boolean, "enable") == 0
-			|| strcmp(boolean, "enabled") == 0
-			|| strcmp(boolean, "active") == 0) {
+	if (strcasecmp(boolean, "1") == 0
+			|| strcasecmp(boolean, "yes") == 0
+			|| strcasecmp(boolean, "on") == 0
+			|| strcasecmp(boolean, "true") == 0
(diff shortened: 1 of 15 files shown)


include/util.h
```diff
@@ -50,6 +50,14 @@ pid_t get_parent_pid(pid_t pid);
  */
 uint32_t parse_color(const char *color);
 
+/**
+ * Given a string that represents a boolean, return the boolean value. This
+ * function also takes in the current boolean value to support toggling. If
+ * toggling is not desired, pass in true for current so that toggling values
+ * get parsed as not true.
+ */
+bool parse_boolean(const char *boolean, const bool current);
```
SirCmpwn: The bool argument doesn't need to be const here

common/util.c
```diff
@@ -123,6 +123,21 @@ uint32_t parse_color(const char *color) {
 	return res;
 }
 
+bool parse_boolean(const char *boolean, bool current) {
```
emersion: Do we want to handle errors? (ie. `bool parse_boolean(const char *str, bool current, bool *out)`)

SirCmpwn: I thought this too, but it looks like the function returns false for invalid values and if that matches i3's behavior we should keep it.

ascent12: It would make sense to add a comment stating that.

RedSoxFan: > It would make sense to add a comment stating that.

Added

> What about case sensitivity? Shouldn't these use `strcasecmp`?

Fixed

SirCmpwn:
Thanks!

PR #2342, opened 2018-07-23
Fix some cases where the cursor doesn't update

RyanDwyer:
`cursor_set_image` only uploads the named image if it doesn't match the previous named image. This means when setting the cursor image to a surface as given by a client (ie. not named), we have to clear the `cursor->image` property.

To test, open a window that sets a different cursor (eg. open a terminal and see the I-beam/text cursor) then switch to a new workspace. The cursor would not update, but now it changes to the regular pointer.


emersion:
Thanks!

PR #2344, opened 2018-07-24
Make mod+resize work in any direction

RyanDwyer:
This makes it so if you hold mod and right click on a surface to resize it, the resize direction is chosen based on which quarter of the surface you've clicked. The previous implementation only resized towards the bottom right.


emersion:
Thanks!

PR #2356, opened 2018-07-25
Allow containers to be fullscreen

RyanDwyer:
Note that unmanaged xwayland surfaces are now rendered when anything is fullscreen, as opposed to only when an xwayland view is fullscreen. The alternative is searching the fullscreen container for an xwayland view, but I figured this isn't worth it.

Test plan:

* `focus parent` then `fullscreen` on different container layout types
* Move views around within a fullscreen container
* Attempt to move views out of a fullscreen container (they shouldn't move)
* Attempt to fullscreen a child of a fullscreen container - the fullscreen container should unfullscreen and the child becomes fullscreen
* In a fullscreen container, `focus parent`, `focus child`, `focus <direction>` into another output
* `focus <direction>` from a sibling output into the fullscreen container
* Move fullscreen containers to other outputs of different sizes
* Fullscreen floating views to make sure I didn't break anything there

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -60,6 +60,8 @@ struct sway_container_state {
 	double swayc_x, swayc_y;
 	double swayc_width, swayc_height;
 
+	bool is_fullscreen;
+
 	bool has_gaps;
 	double current_gaps;
 	double gaps_inner;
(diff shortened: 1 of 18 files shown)


SirCmpwn:
Can you rebase this?

RedSoxFan:
Thanks!

minus7:
On a workspace with two windows, running `swaymsg focus parent && swaymsg fullscreen` fullscreens both windows for a second, then fullscreens just one.

PR #2357, opened 2018-07-25
Refactor view_map

RyanDwyer:
* Move workspace selection into separate function
* Instead of keeping a `prev_focus` variable, do the check in `should_focus` (ie. views can only take focus if they're mapped into the active workspace)
* Fix assign-to-output - it previously set `prev_focus` but should be `target_sibling`
* Remove call to `workspace_switch` as we'll only ever focus the view if it's in the active workspace


RedSoxFan:
Thanks!

PR #2365, opened 2018-07-26
Fix crash when a deferred command destroys a workspace

RyanDwyer:
Example config that produces the crash (with a single output):

    workspace 1
    workspace 2

I'm not sure on the specifics as to why it fails, but this fixes it by committing a transaction after each deferred command.


SirCmpwn:
This doesn't fix the bug for me. I will likely be able to look into this over the weekend.

RyanDwyer:
Don't merge this - I've found a better way of resolving this.

RyanDwyer:
I've replaced my commit with a different one that addresses the root cause of the problem. Ready for review again.

SirCmpwn:
Thanks!

Issue #545, opened 2016-03-26
Inform users of configuration errors visually


SirCmpwn:
i3 shows you config file errors with nagbar, which would be nice for us to do as well at some point.

PR #2366, opened 2018-07-27
Implement swaynag

RedSoxFan:
This is the first step towards #545.

This PR contains a clone of `i3-nagbar` with some added functionality. After some discussion on IRC, it was determined that we will do things a little different from i3 for showing errors. Instead of launching a terminal to show the errors, we will have a button to show detailed errors in a multi-line text box inside the expanded layer-surface.

The objective of this PR is to create `swaynagbar` in generic manner so it can be used with sway. This will not implement any actual use-cases, that will be left for future PRs.

**Left for follow-up PRs:**
- Show nagbar when there is config errors
- Add nagbar to the binding for `exit` in the default config (to match i3)
- (any other use cases that come up)

**Test Plan:**
- Show nagbar with a basic message
  - Make sure the dismiss button dismisses nagbar
- Show nagbar with a type of warning
- Show nagbar on a specific output
- Show nagbar on the bottom edge
- Show nagbar with an action button
  - Make sure the action is run when the button is clicked
- Show nagbar with a detailed message
  - Make sure nagbar expands and collapses
  - Make sure the text area can scroll with the mouse wheel and the buttons (requires long text)
  - Make sure a detailed message can be piped in via stdin
- Make sure config file works
  - Make sure default types (error and warning) can be overridden
  - Make sure custom types can be defined
  - Make sure command line args override config options


sway/desktop/layer_shell.c
```diff
@@ -212,6 +214,9 @@ void arrange_layers(struct sway_output *output) {
 	wl_list_for_each(seat, &input_manager->seats, link) {
 		seat_set_focus_layer(seat, topmost ? topmost->layer_surface : NULL);
 	}
+
+	arrange_windows(output->swayc);
+	transaction_commit_dirty();
```
RyanDwyer: Calling `transaction_commit_dirty` here could lead to problems if there is other code that calls `arrange_layers` at a time when the tree is in an inconsistent state. I try to only call it directly in the handler functions to avoid this.

Also, there is other code in sway that calls `arrange_layers` followed by `arrange_windows`, so this would end up calling `arrange_windows` twice which isn't very efficient.

To be safe, I'd pull these out and call them from the relevant `handle_*` functions in this file.

BTW, I'd like to wire `transaction_commit_dirty` up to the event loop so it's run automatically, but I haven't looked into that yet.

swaynagbar/main.c
```diff
@@ -0,0 +1,250 @@
...
+				nagbar.type = NAGBAR_WARNING;
+			} else {
+				fprintf(stderr, "Type must be either 'error' or 'warning'\n");
+				exit_code = EXIT_FAILURE;
+				goto cleanup;
+			}
+			set_nagbar_colors();
+			break;
+		case 'v': // Version
+			fprintf(stdout, "sway version " SWAY_VERSION "\n");
```
RyanDwyer: s/sway/swaynagbar/

swaynagbar/render.c
```diff
@@ -0,0 +1,298 @@
...
+		wl_surface_commit(nagbar->surface);
+		wl_display_roundtrip(nagbar->display);
+	} else {
+		nagbar->current_buffer = get_next_buffer(nagbar->shm,
+				nagbar->buffers,
+				nagbar->width * nagbar->scale,
+				nagbar->height * nagbar->scale);
+		if (!nagbar->current_buffer) {
+			wlr_log(WLR_DEBUG, "Failed to get buffer. Skipping frame.");
+			return;
```
RyanDwyer: Surely you need to destroy the cairo surface and the cairo context here, or `goto` the end so it can be cleaned up there?

SirCmpwn:
This should be in `ZWLR_LAYER_SHELL_V1_LAYER_TOP`. I also think that instead of using `-l` we should pipe the extended details into stdin, because specifying multi-line command line args is annoying.

I also have some visual changes I want to make, I'll push the changes to your branch later.

SirCmpwn:
I'd also probably call the binary `swaynag` just for brevity's sake.

RedSoxFan:
>  I also think that instead of using -l we should pipe the extended details into stdin, because specifying multi-line command line args is annoying.

That should already be working. Just use `-l -`
(ex. `echo -e "Line1\nLine2" | swaynagbar -m Message -l -`)

--

I'll push changes for the rest of the comments in a few minutes.

SirCmpwn:
I don't think we should treat an arg as a file descriptor OR a literal string. It should be XOR.

RedSoxFan:
> I don't think we should treat an arg as a file descriptor OR a literal string. It should be XOR.

Ok. I'll remove the argument from the flag and if the flag is given, just read from stdin.

SirCmpwn:
+1

SirCmpwn:
I don't think this should apply an exclusive zone.

RedSoxFan:
> I don't think this should apply an exclusive zone.

I personally prefer it with the exclusive zone. If we have errors displaying in the detailed box, it's easy to open an editor in the available space. Without setting it as exclusive, you would have to manipulate the layout to see your editor and the detailed message.

I could make it a flag that can be decided by the use case. I'm also considering a follow-up with a config file for user preferences like default font, default output, and default edge. I guess that could have a default user preference as well.

SirCmpwn:
We should also probably limit the maximum height of the details and let the user scroll with the mouse wheel.

RedSoxFan:
>We should also probably limit the maximum height of the details

Currently the maximum height of `swaynag` as a whole is hard coded to `500px`.

> let the user scroll with the mouse wheel.

That should be working right now

SirCmpwn:
Oh neat I missed that

SirCmpwn:
Needs hidpi support

SirCmpwn:
```
./swaynag/swaynag -dm 'There are errors in your config file!' -b 'Exit sway' 'swaymsg exit' -b 'Reload config' 'swaymsg reload' -l < some-file
```

This only shows one button and no details.

SirCmpwn:
`-b` should probably be done as `-b message=action` rather than two arguments for better getopt support.

Also, this:

```
./swaynag/swaynag -dm 'Test' -b 'Fullscreen' 'swaymsg fullscreen toggle'
```

Gives me the following:

```
2018-07-28 09:52:38 - [swaynag/nagbar.c:57] $TERMINAL not found. Running directly
/bin/sh: -c: option requires an argument
```

SirCmpwn:
Oh I guess `-b` is done that way for i3 backwards compatability. Meh.

SirCmpwn:
Config should be moved into a separate file from main.c

SirCmpwn:
The following config does not seem to work:

```
[error]
font=pango:IPAGothic 10, DejaVu Sans Mono 9
```

SirCmpwn:
>What do I need to do for that?

I dug into this a little bit, the way you handle outputs needs to be reworked. You should stash some state about each of them and wait until surface_enter is sent to apply a scale factor for the relevant output.

RedSoxFan:
> Currently, only colors can be set per type. I'm working on allowing other option defaults per type.

The config now allows for per-type defaults to be defined for `font`, `edge`, `output`, and several sizing options. Additionally, all color and sizing options can also be given as command line arguments. The order of preference for options is command line argument value, per-type config default, global config default, and finally swaynag default.

RedSoxFan:
> I dug into this a little bit, the way you handle outputs needs to be reworked. You should stash some state about each of them and wait until surface_enter is sent to apply a scale factor for the relevant output.

I think this should be fixed now.

Also, I think this PR is ready for another round of review.

SirCmpwn:
On a hidpi display the cursor is lowdpi.

Otherwise tests hold up, looks pretty good. I'll take a look through the code shortly.

SirCmpwn:
`swaynag -h` segfaults

SirCmpwn:
If I only specify one color in a config for a new type, the rest of the values should probably be sane (I only set background)

RedSoxFan:
> On a hidpi display the cursor is lowdpi.

This should be fixed. I was using swaybar as a reference and apparently a PR needs to be opened to support hidpi cursors in swaybar.

> `swaynag -h` segfaults

Should be fixed now.

> If I only specify one color in a config for a new type, the rest of the values should probably be sane (I only set background)

Any color not set will now fallback to the default color for the error type.

SirCmpwn:
>This should be fixed. I was using swaybar as a reference and apparently a PR needs to be opened to support hidpi cursors in swaybar.

Hmm, swaybar supports HiDPI cursors just fine...

SirCmpwn:
>Any color not set will now fallback to the default color for the error type

I'm not sure that makes sense, what do you think about coming up with a fallback color scheme which looks like swaybar's defaults?

RedSoxFan:
> Hmm, swaybar supports HiDPI cursors just fine...

Correction: swaybar does not update cursor scale when the output scale changes.

> I'm not sure that makes sense, what do you think about coming up with a fallback color scheme which looks like swaybar's defaults?

Sure. I just used the error colors since that's the default type

swaynag/swaynag.5.scd
```diff
@@ -0,0 +1,97 @@
...
+# NAME
+
+swaynag - swaynag configuration file
+
+# SYNOPSIS
+
+$HOME/.swaynag/config, $XDG\_CONFIG\_HOME/swaynag/config,
+SYSCONFDIR/swaynag/config
+
+# CONFIG FILE
```
SirCmpwn: Add an empty line after headings

SirCmpwn:
LGTM aside from the scdoc issue

SirCmpwn:
@RyanDwyer can you take a look at this?

SirCmpwn:
Thanks!

PR #2368, opened 2018-07-27
Handle out-of-fd situations gracefully for transaction and urgent timers

RyanDwyer:
I tested it by changing the condition to `false`.


SirCmpwn:
Thanks!

PR #2370, opened 2018-07-28
Fix use after free in transactions

RyanDwyer:
In `set_instructions_ready`, calling `set_instruction_ready` may cause any number of transactions to get applied, which removes them from the list being iterated.  The iteration variables need to be adjusted accordingly.


RedSoxFan:
Thanks!

PR #2371, opened 2018-07-28
Don't enable numlock by default.

ggreer:
This fixes a confusing issue where some laptop keyboards would have 'numlock mode' enabled, remapping parts of the alphabet to numbers.

It'd be great if we could detect keyboards with dedicated numpads and enable numlock by default, but that sounds really hard. :/

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -387,7 +387,7 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
 	wlr_keyboard_set_keymap(wlr_device->keyboard, keyboard->keymap);
 
 	xkb_mod_mask_t locked_mods = 0;
-	if (!input_config || input_config->xkb_numlock != 0) {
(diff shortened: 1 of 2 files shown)


sway/input/keyboard.c
```diff
@@ -387,7 +387,7 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
 	wlr_keyboard_set_keymap(wlr_device->keyboard, keyboard->keymap);
 
 	xkb_mod_mask_t locked_mods = 0;
-	if (!input_config || input_config->xkb_numlock != 0) {
+	if (input_config && input_config->xkb_numlock) {
```
progandy: This should test for ```xkb_numlock > 0```. The variable has three states:
-1: undefined, keep xkbcommon default
0 numlock disabled
1 numlock enabled

SirCmpwn:
Thanks!

PR #2372, opened 2018-07-28
Second attempt at fixing transaction use-after-free

RyanDwyer:
The solution used in 073ac425d5bf6f6393eb91d9b5f84e3caa68f511 doesn't work in all cases because the freed instruction might be ahead in the list, not necessarily behind.

The new solution delays running the queue until after the loop has finished iterating, thus avoiding the problem completely.

To test, I opened filezilla (not firefox), floated it, then resized it violently using mod+resize. It would crash within several seconds.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -344,13 +344,11 @@ static void set_instruction_ready(
 
 	}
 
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Thanks!

PR #2375, opened 2018-07-28
[WIP] Add security features

ddevault:
- [x] sway-security(7)
- [x] Load security configs
- [x] `permit` and `reject` commands
- [x] Set up Wayland socket for clients on their behalf
- [ ] Apply global filters
- [ ] `-Dinsecure`
- [ ] Lock down fullscreen

Later:

- [ ] Add IPC security features


SirCmpwn:
Note to self: the bitfield of permitted features will be arrived at via `((default_policy.permit_features & ~default_policy.reject_features) | program_policy.permit_features) & ~program_policy.reject_features`.

SirCmpwn:
My intention is to finish this before the week is through.

SirCmpwn:
As I'm working through this, the infancy of this plan is starting to dawn on me. I think I want to remove this from the 1.0 milestone and postpone it while we put together a design which we can bring to all Wayland compositors.

c-edw:
I was recently looking into `seccomp`, could it be useful here? Though, it's a Linux kernel feature, so BSD users will miss out.

emersion:
>I was recently looking into seccomp, could it be useful here? Though, it's a Linux kernel feature, so BSD users will miss out.

No, `seccomp` are useful for sandboxes, we're not a sandbox.

c-edw:
@emersion It is certainly used in sandboxes, but it seems it can be used a process to restrict it's own capabilities.

emersion:
We don't need to do that. It's not our job.

emersion:
When this is ready, I'd like a real-world example of how this can be useful, e.g. running a restricted Wayland client in a chroot.

emersion:
This won't work well for users who want their service manager to manage clients (e.g. swaylock).

ddevault:
How so?

ddevault:
Ah.

ddevault:
I think that gets back to the "we could have a protocol for creating authorized Wayland sockets to pass to your client's child", though per the discussion before we will likely need to consider this some kind of "superuser" protocol where the user is assumed to have access to everything.

Issue #2373, opened 2018-07-28
Add virtual keyboard and text input protocols to sway


ddevault:
- [x] virtual-keyboard
- [x] text-input
- [x] input-method

SirCmpwn:
Note: text input will have to wait until Purism finishes their work with it

emersion:
All of these have been implemented.

PR #2376, opened 2018-07-29
Add virtual keyboard protocol

SirCmpwn:
Ref #2373 

To test use [virtboard](https://source.puri.sm/Librem5/virtboard)

cc @dcz-purism

### DIFF
--- a/include/sway/input/input-manager.h
+++ b/include/sway/input/input-manager.h
@@ -2,6 +2,7 @@
 #define _SWAY_INPUT_INPUT_MANAGER_H
 #include <libinput.h>
 #include <wlr/types/wlr_input_inhibitor.h>
+#include <wlr/types/wlr_virtual_keyboard_v1.h>
 #include "sway/server.h"
 #include "sway/config.h"
 #include "list.h"
@@ -25,10 +26,12 @@ struct sway_input_manager {
 	struct wl_list seats;
 
 	struct wlr_input_inhibit_manager *inhibit;
+	struct wlr_virtual_keyboard_manager_v1 *virtual_keyboard;
 
 	struct wl_listener new_input;
 	struct wl_listener inhibit_activate;
 	struct wl_listener inhibit_deactivate;
+	struct wl_listener virtual_keyboard_new;
 };
 
 struct sway_input_manager *input_manager_create(struct sway_server *server);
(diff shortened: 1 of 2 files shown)


JonnyMako:
So, I'm trying to use `virtboard` but I get this message. Is this a regression? Am I missing components?
```
No input method manager global, some functionality unavailable
Could not register to session manager: Cannot invoke method; proxy is for a well-known name without an owner and proxy was constructed with the G_DBUS_PROXY_FLAGS_DO_NOT_AUTO_START flag
```

dcz-purism:
These messages come from virtboard. The first one is talking that the input-method protocol is not implemented, and the other one is about missing `org.gnome.SessionManager` dbus interface. It should be working fine despite both, is it?

dcz-purism:
Can you check if the dbus invokation works?

```
busctl call --user sm.puri.OSK0 /sm/puri/OSK0 sm.puri.OSK0 SetVisible b true
```

JonnyMako:
@dcz-purism sorry I missed your reply. That still does not work:
```
The name sm.puri.OSK0 was not provided by any .service files
```

ianyfan:
Sanity check that you're doing the busctl call *after* running virtboard?

kosciak9:
Hey @ianyfan I'm having same issue. Running `v1.0-3` straight from Arch repos with `virtboard v0.0.5`. Running `virtboard` produces:

```
No input method manager global, some functionality unavailable
Could not register to session manager: Cannot invoke method; proxy is for the well-known name org.gnome.SessionManager without an owner, and proxy was constructed with the G_DBUS_PROXY_FLAGS_DO_NOT_AUTO_START flag

(process:8181): GLib-CRITICAL **: 17:49:32.580: g_source_remove: assertion 'tag > 0' failed

(process:8181): GLib-GObject-CRITICAL **: 17:49:32.581: g_object_set: assertion 'G_IS_OBJECT (object)' failed
wl_surface@18: error 0: Specified scale value (0) is not positive
```

and running `busctl call --user sm.puri.OSK0 /sm/puri/OSK0 sm.puri.OSK0 SetVisible b true` produces no output as if it went fine. Keyboard is not showing up though.

kosciak9:
Okay, I feel ashamed, but I was running `virtboard-git` not `virtboard`. Switched packages and it works fine now.

PR #2381, opened 2018-07-29
Implement key repeat for keybindings

### LINKED ISSUE #2332: Support key repeat for evaluating keybindings
I can't, for example, hold down XF86VolumeUp to keep increasing my volume.

frsfnrrg:
This PR resolves #2332. Have tested with

    input 1:1:Piano {
            repeat_rate 2
            repeat_delay 2000
    }

Setting 

    input 0:0:X11_keyboard {
            repeat_rate 0
            repeat_delay 0
    }

should disable key repeat. 

~I will update the documentation soon.~ (No documentation changes, as this PR now reuses the advised repeat_rate and repeat_delay from the Wayland protocol.)

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -285,10 +285,10 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
 	}
 
 	// Set up (or clear) keyboard repeat for a pressed binding
-	if (next_repeat_binding) {
+	if (next_repeat_binding && wlr_device->keyboard->repeat_info.delay > 0) {
 		keyboard->repeat_binding = next_repeat_binding;
 		if (wl_event_source_timer_update(keyboard->key_repeat_source,
-				keyboard->key_repeat_initial_delay) < 0) {
+				wlr_device->keyboard->repeat_info.delay) < 0) {
 			wlr_log(WLR_DEBUG, "failed to set key repeat timer");
 		}
 	} else if (keyboard->repeat_binding) {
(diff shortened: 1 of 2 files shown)


include/sway/commands.h
```diff
@@ -208,7 +208,7 @@ sway_cmd input_cmd_middle_emulation;
 sway_cmd input_cmd_natural_scroll;
 sway_cmd input_cmd_pointer_accel;
 sway_cmd input_cmd_repeat_delay;
-sway_cmd input_cmd_repeat_rate;
+sway_cmd input_cmd_repeat_step;
```
emersion: Any reason you renamed this? Wayland uses `repeat`, not `step` (https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_keyboard-event-repeat_info)

frsfnrrg: I was unaware of that protocol, and have consistently uses 'repeat_step' to denote the spacing (in milliseconds) between subsequent binding invocations. The protocol does not apply in this instance
because the binding repeat parameters are, as far as I know, used only for bindings, not clients.

emersion: I think `repeat` is better

frsfnrrg: It seems I've incorrectly repurposed the advisory key repeat/rate information. Would you prefer:
* To create separate `binding_repeat_rate` and  `binding_repeat_delay` commands?
* To apply the advised repeat/rate information to the bindings?

SirCmpwn: Do what you think is best so long as the language matches the Wayland protocol

frsfnrrg:
I've decided to use `repeat_rate` and `repeat_delay`; this reduces the extent of the change from this PR.

sway/input/keyboard.c
```diff
@@ -303,6 +319,26 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
...
 
+static int handle_keyboard_repeat(void* data) {
+	struct sway_keyboard *keyboard = (struct sway_keyboard *)data;
+	struct wlr_keyboard *wlr_device =
+			keyboard->seat_device->input_device->wlr_device->keyboard;
+	if (keyboard->repeat_binding) {
+		if (wlr_device->repeat_info.rate > 0) {
+			// We queue the next event first, as the command might cancel it
+			if (wl_event_source_timer_update(keyboard->key_repeat_source,
+					1000 / wlr_device->repeat_info.rate) < 0) {
```
emersion: This value is different from `handle_keyboard_key`'s.

emersion: Oh, my bad.

SirCmpwn:
Thanks!

PR #2383, opened 2018-07-30
Fix title textures on scale change

RedSoxFan:
On an output scale change, only the title texture of the focused container and all marks textures were being updated.

This PR updates the title textures for all containers.

Test Plan:
- Open any number of views in various layouts (including floating)
- Add some marks
- Change the scale of the output
- Verify all title and marks textures are updated to use the correct scale


SirCmpwn:
Thanks! At some point we also need to deal with dragging floating windows between outputs of different scales.

PR #2386, opened 2018-07-30
use standard ARRAY_SIZE() macro

1ace:
Replace all the custom `sizeof(array)/sizeof(element)` and `sizeof(array)/sizeof(type)` with the standard macro `ARRAY_SIZE()`.

This reduces the risk of passing the wrong type, especially when code gets refactored and things change type.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -6,6 +6,8 @@
 #include <sys/types.h> 
 #include <xkbcommon/xkbcommon.h>
 
+#define ARRAY_SIZE(a) (sizeof(a)/sizeof(a)[0])
+
 /**
  * Wrap i into the range [0, max[
  */
--- a/common/unicode.c
+++ b/common/unicode.c
@@ -1,6 +1,7 @@
 #include <stdint.h>
 #include <stddef.h>
 #include "unicode.h"
+#include "util.h"
 
 size_t utf8_chsize(uint32_t ch) {
 	if (ch < 0x80) {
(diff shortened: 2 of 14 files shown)


SirCmpwn:
Hell no.

include/util.h
```diff
@@ -6,6 +6,8 @@
 #include <sys/types.h> 
 #include <xkbcommon/xkbcommon.h>
 
+#define ARRAY_SIZE(a) (sizeof(a)/sizeof(a)[0])
```
minus7: Is that even correct?

1ace: What would be incorrect here?

1ace:
Not gonna push for this change or anything, but I'm curious: this is a standard macro, present in all large projects I've worked on, and its use has fixed many bugs, so what is the reason for this rather strong "hell no" reaction?

include/util.h
minus7: It should be `sizeof(a[0])` instead of `sizeof(a)[0]`

minus7:
To elaborate on the 'Hell no': Sway doesn't use any macros (except for logging) and @SirCmpwn hates them.

SirCmpwn:
It's not a "standard" macro unless it's standardized, which it is not.

1ace:
I'ts standard in the sense that it's proven by being used everywhere (starting with the linux kernel itself), and is trivial enough to be guaranteed to be bug-free.

I have to admit I don't understand the "never use macros" reasoning, but I'm not too bothered by it, so ok, no macros in sway :man_shrugging:

SirCmpwn:
The Linux kernel is far from the pinaccle of C code. We use *acutal* standards here, namely ISO C1 and POSIX 2008.

PR #2387, opened 2018-07-30
bash completion

1ace:
Similar to the existing shell completion for zsh, this adds support for completion in bash

### DIFF
--- a/completions/bash/sway
+++ b/completions/bash/sway
@@ -1,0 +1,46 @@
+# sway(1) completion
+
+_sway()
+{
+  local cur prev
+  _get_comp_words_by_ref cur prev
+
+  short=(
+    -h
+    -c
+    -C
+    -d
+    -v
+    -V
+  )
+
+  long=(
+    --help
+    --config
+    --validate
+    --debug
+    --version
+    --verbose
+    --get-socketpath
+  )
+
+  case $prev in
+    -c|--config)
+      _filedir
+      return
+      ;;
+  esac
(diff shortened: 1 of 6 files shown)


completions/bash/swaygrab
```diff
@@ -0,0 +1,48 @@
+# swaygrab(1) completion
```
emersion: `swaygrab` has been removed. We should remove completions too.

1ace: OK, sure.
I haven't followed the development of sway; why was `swaygrab` removed? Is its functionality available some other way?

emersion: Yeah, you can now use tools like [grim](https://github.com/emersion/grim).

emersion: If possible, can you open a PR to remove all `swaygrab` references in docs and completion?

```
meson.build
185:		'completions/zsh/_swaygrab',

sway/sway.1.scd
95:*sway*(5) *swaymsg*(1) *swaygrab*(1) *sway-input*(5) *sway-bar*(5)

swayidle/swayidle.1.scd
61:*sway*(5) *swaymsg*(1) *swaygrab*(1) *sway-input*(5) *sway-bar*(5)

completions/zsh/_swaygrab
1:#compdef swaygrab
6:# Completion script for swaygrab in sway wm (http://swaywm.org)
```

SirCmpwn:
Thanks!

PR #2367, opened 2018-07-27
Refactor surface iterators

emersion:
The current iterators design is very confusing. This PR will:
- Remove `struct root_geometry`
- Remove `output_get_surface_box`

Updates #2291

PR #2390, opened 2018-07-30
Fix segfault with fullscreen

emersion:
Introduced in #2367


minus7:
No more crash. I don't remotely understand what's going on here

emersion:
`send_frame_done_container_iterator` is an iterator, thus has a `void *data` argument. It expects a `struct send_frame_done_data *` but I was giving a `struct timespec *`. This patch switches to a safer function without a `void *` argument.

SirCmpwn:
Thanks!

PR #2391, opened 2018-07-31
Fix popups (v2)

### LINKED ISSUE #2210: Popups don't work in floating views
https://github.com/swaywm/sway/pull/2204

RyanDwyer:
Fixes the render and `container_at` order for popups.

Fixes #2210

For rendering:

* `render_view_surfaces` has been renamed to `render_view_toplevels`
* `render_view_toplevels` now uses `output_surface_for_each_surface` (which is now public), as that function uses `wlr_surface_for_each_surface` which doesn't descend into popups
* Views now have a `for_each_popup` iterator, which is used by the renderer to render the focused view's popups
* When rendering a popup, toplevels (xdg subsurfaces) of that popup are also rendered
* Popups are now rendered on top of all containers

For sending frame done, the logic has been updated to match the rendering logic:

* `send_frame_done_container` no longer descends into popups
* `for_each_popup` is used to send frame done to the focused view's popups and their child toplevels

For `container_at`:

* `floating_container_at` is now static, which means it had to be moved higher in the file.
* `container_at` now considers popups for the focused view before checking containers.
* `tiling_container_at` has been introduced, so that it doesn't call `container_at` recursively (it would check popups recursively if it did)

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -67,9 +67,17 @@ struct sway_container *output_get_active_workspace(struct sway_output *output);
 void output_render(struct sway_output *output, struct timespec *when,
 	pixman_region32_t *damage);
 
+void output_surface_for_each_surface(struct sway_output *output,
+		struct wlr_surface *surface, double ox, double oy,
+		sway_surface_iterator_func_t iterator, void *user_data);
+
 void output_view_for_each_surface(struct sway_output *output,
 	struct sway_view *view, sway_surface_iterator_func_t iterator,
 	void *user_data);
+
+void output_view_for_each_popup(struct sway_output *output,
+		struct sway_view *view, sway_surface_iterator_func_t iterator,
+		void *user_data);
 
 void output_layer_for_each_surface(struct sway_output *output,
 	struct wl_list *layer_surfaces, sway_surface_iterator_func_t iterator,
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -230,16 +230,9 @@ struct sway_container *container_parent(struct sway_container *container,
  * surface-local coordinates of the given layout coordinates if the container
  * is a view and the view contains a surface at those coordinates.
  */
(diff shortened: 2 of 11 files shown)


emersion:
>For sending frame done, the logic has been updated to match the rendering logic:

This is probably not _required_: while the order in which views are rendered is important, the order in which we send frame events isn't.

RyanDwyer:
In the current master you can open a popup and then focus another view using a bindsym, and the popup is still rendered and receives frame done events. In my PR I only render popups for the currently focused view, which means the frame done events should be limited to the focused view too.

I guess in the future we'll make it so it either closes the popups on focus change, or denies the focus change entirely, in which case my changes to the send frame done in this PR can be reverted.

emersion:
>In my PR I only render popups for the currently focused view, which means the frame done events should be limited to the focused view too.

I don't think that's right. We should send frame events to all visible surfaces.

RyanDwyer:
Sorry, I mean "the frame done events should be limited to *the popups for* the focused view too"

emersion:
If the popup is visible at all, we should send frame events. So I don't think this PR needs to make the `send_frame_done` code more complicated (in fact, this PR shouldn't affect it).

emersion:
Or does this PR only renders popups for the focused view? In this case we should do as you said.

RyanDwyer:
Yeah, it only renders popups for the focused view. But now I've changed it so popups are closed when changing focus, so I've reverted the send frame done code.

emersion:
Ah, sorry, I missed that (I only read the PR description, not the actual diff).

SirCmpwn:
Thanks!

SirCmpwn:
This has a regression on my desktop at work where none of the views on *one* of my displays are rendered. The borders, swaybar, and swaybg all render fine, but the windows just aren't there. Tested with Xwayland and xdg-shell.

PR #2393, opened 2018-07-31
Link xcb dependency to meson options "enable_xwayland"

ganguin:
xcb is only required with xwayland, so remove xcb dependency if xwayland is disabled.

I'm new to meson, so I'm not sure it's done the right way. But I successfully tested the build with both -Denable-xwayland=true and -Denable-xwayland=false.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -50,7 +50,7 @@ conf_data = configuration_data()
 
 if get_option('enable-xwayland')
 	conf_data.set('HAVE_XWAYLAND', true)
-	xcb    = dependency('xcb')
+	xcb = dependency('xcb')
 else
 	conf_data.set('HAVE_XWAYLAND', false)
 endif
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -44,13 +44,13 @@ systemd        = dependency('libsystemd', required: false)
...
 math           = cc.find_library('m')
 rt             = cc.find_library('rt')
-xcb            = dependency('xcb')
 git            = find_program('git', required: false)
 
 conf_data = configuration_data()
 
 if get_option('enable-xwayland')
 	conf_data.set('HAVE_XWAYLAND', true)
+	xcb    = dependency('xcb')
```
SirCmpwn: can you remove the extra whitespace before the = here

emersion:
Thanks!

PR #2396, opened 2018-08-01
Correctly track saved surfaces during multiple transactions

### LINKED ISSUE #2364: Floating resize is not atomic
1. Open a floating xdg-shell window
2. Resize it from the top left corner
3. See that the bottom right corner is moving back and forth

Try rootston to see the expected resizing behaviour.

RyanDwyer:
Fixes #2364.

Suppose a view is 600px wide, and we tell it to resize to 601px during a resize operation. We create a transaction, save the 600px buffer and send the configure. This buffer is saved into the associated instruction, and is rendered while we wait for the view to commit a 601px buffer.

Before the view commits the 601px buffer, suppose we tell it to resize to 602px. The new transaction will also save the buffer, but it's still the 600px buffer because we haven't received a new one yet.

Then suppose the view commits its original 601px buffer. This completes the first transaction, so we apply the 601px width to the container. There's still the second (now only) transaction remaining, so we render the saved buffer from that. But this is still the 600px buffer, and we believe it's 601px. Whoops.

The problem here is we can't stack buffers like this. So this commit removes the saved buffer from the instructions, places it in the view instead, and re-saves the latest buffer every time the view completes a transaction and still has further pending transactions.

As saved buffers are now specific to views rather than instructions, the functions for saving and removing the saved buffer have been moved to `view.c`.

The calls to save and restore the buffer have been relocated to more appropriate functions too, favouring `transaction_commit` and `transaction_apply` rather than `transaction_add_container` and `transaction_destroy`.

To test:

* Open an xdg shell view
* Float it
* Resize it towards the top left using mod + right click

Do not test using FileZilla. That client appears to misbehave when there's multiple configures. It commits with a new size, but the texture contains old contents. I've spent so many hours trying to find a problem in sway or wlroots in relation to this, before realising that FileZilla was the only client that does this :-\

### DIFF
--- a/include/sway/desktop/transaction.h
+++ b/include/sway/desktop/transaction.h
@@ -42,17 +42,4 @@ void transaction_notify_view_ready(struct sway_view *view, uint32_t serial);
 void transaction_notify_view_ready_by_size(struct sway_view *view,
 		int width, int height);
 
-/**
- * Get the saved texture that should be rendered for a view.
- *
- * The addresses pointed at by the width and height pointers will be populated
- * with the surface's dimensions, which may be different to the texture's
- * dimensions if output scaling is used.
- *
- * This function should only be called if it is known that the view has
- * instructions.
- */
-struct wlr_texture *transaction_get_saved_texture(struct sway_view *view,
-		int *width, int *height);
-
 #endif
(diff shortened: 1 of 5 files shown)


RedSoxFan:
This looks good for resizing floating xdg-shell views.

I'm not sure if it is in the scope of this PR, but there are still artifacts when resizing floating containers with at least one xdg-shell view. It looks like it's only happening when resizing using one of the left corners and doesn't matter what the layout is.

RyanDwyer:
I think that's related to geometry (#2197). We get the damage box wrong.

PR #2397, opened 2018-08-01
XCursor is not configured if no pointer device is available

chr0me-sh:
Small change that stops the cursor from being displayed if there is no pointer input device. Created this to address the floating cursor in the middle of my screen, even though I am a keyboard-only user.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -393,7 +393,6 @@ struct sway_seat *seat_create(struct sway_input_manager *input,
 		WL_SEAT_CAPABILITY_POINTER |
 		WL_SEAT_CAPABILITY_TOUCH);
 
-	seat_configure_xcursor(seat);
 
 	wl_list_insert(&input->seats, &seat->link);
 
@@ -438,6 +437,7 @@ static void seat_apply_input_config(struct sway_seat *seat,
 
 static void seat_configure_pointer(struct sway_seat *seat,
 		struct sway_seat_device *sway_device) {
+	seat_configure_xcursor(seat);
 	wlr_cursor_attach_input_device(seat->cursor->cursor,
 		sway_device->input_device->wlr_device);
 	seat_apply_input_config(seat, sway_device);


sway/input/seat.c
```diff
@@ -370,7 +370,7 @@ struct sway_seat *seat_create(struct sway_input_manager *input,
 		wlr_seat_destroy(seat->wlr_seat);
 		free(seat);
 		return NULL;
-	}
+	}
```
SirCmpwn: Looks like this snuck in

sway/input/seat.c
```diff
@@ -438,6 +437,7 @@ static void seat_apply_input_config(struct sway_seat *seat,
 
 static void seat_configure_pointer(struct sway_seat *seat,
 		struct sway_seat_device *sway_device) {
+	seat_configure_xcursor(seat);
```
SirCmpwn: Does this have any undesirable side-effects if there are multiple pointers on the seat?

chr0me-sh: The laptop I am testing it on has both a "nub" pointer device and a trackpad. No side effects, both of them work fine.

SirCmpwn:
Can you squash those commits together, too?

emersion:
This won't work if you have only a tablet attached to the seat (and no pointer). Touch probably shouldn't require a pointer (not sure if it should be displayed when we'll implement touch fallback? - this is for later anyway).

Follow-up issue: if you unplug all pointer devices, it should hide the cursor again. This can probably be done by always calling `seat_configure_xcursor`, but call:

```c
wlr_xcursor_manager_set_cursor_image(seat->cursor->xcursor_manager, NULL, seat->cursor->cursor);
```

When there's no pointer/tablet device attached in `seat_add_device` and `seat_remove_device` (and setting the cursor back to `left_ptr` if there's at least one pointer/tablet device).

chr0me-sh:
> Follow-up issue: if you unplug all pointer devices, it should hide the cursor again. This can probably be done by always calling seat_configure_xcursor as it's done right now, but also call...

@emersion I will have a look at addressing your follow up issue and push a fix when I can get it done.

SirCmpwn:
https://github.com/swaywm/sway/issues/2398

PR #2400, opened 2018-08-02
Show swaynag on config errors

### LINKED ISSUE #545: Inform users of configuration errors visually

RedSoxFan:
Closes #545 

This PR spawns `swaynag` when there are errors in the config. When reloading, the spawned instance is killed and if needed, a new instance is spawned.

~Additionally, this PR changes the allocation for the `error` field of `cmd_results` to get the size from `vsnprintf` instead of using a hard coded 256 bytes. This is to allow for the config error log for an included config to be passed as the error for the `include` command in the calling config.~

I also added a `swaynag_command` command so that a custom path for swaynag or additional arguments can be given.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -132,6 +132,7 @@ static struct cmd_handler handlers[] = {
 static struct cmd_handler config_handlers[] = {
 	{ "default_orientation", cmd_default_orientation },
 	{ "swaybg_command", cmd_swaybg_command },
+	{ "swaynag_command", cmd_swaynag_command },
 	{ "workspace_layout", cmd_workspace_layout },
 };
 
@@ -511,14 +512,11 @@ struct cmd_results *cmd_results_new(enum cmd_status status,
 		results->input = NULL;
 	}
 	if (format) {
+		char *error = malloc(256);
 		va_list args;
 		va_start(args, format);
-		size_t length = vsnprintf(NULL, 0, format, args) + 1;
-		char *error = malloc(length);
-		va_end(args);
-		va_start(args, format);
 		if (error) {
-			vsnprintf(error, length, format, args);
+			vsnprintf(error, 256, format, args);
 		}
 		va_end(args);
 		results->error = error;
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -511,11 +511,14 @@ struct cmd_results *cmd_results_new(enum cmd_status status,
 		results->input = NULL;
 	}
 	if (format) {
-		char *error = malloc(256);
 		va_list args;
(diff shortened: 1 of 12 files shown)


SirCmpwn:
Needs a rebase

SirCmpwn:
Rather than build up a list of errors in memory, I think we should spawn swaynag on the first error, leave the file descriptor open, and write all of the errors into it as we encounter them. Then when we're done reading the config we can close the fd and swaynag will display.

RedSoxFan:
> Rather than build up a list of errors in memory, I think we should spawn swaynag on the first error, leave the file descriptor open, and write all of the errors into it as we encounter them. Then when we're done reading the config we can close the fd and swaynag will display.

Done. In the process, I also made it so that `swaynag` can be called in a generic way for anything that gets added in the future. Additionally, I added a `swaynag_command` config directive that is used as a base.

sway/commands.c
```diff
@@ -511,14 +512,11 @@ struct cmd_results *cmd_results_new(enum cmd_status status,
...
+		char *error = malloc(256);
 		va_list args;
 		va_start(args, format);
-		size_t length = vsnprintf(NULL, 0, format, args) + 1;
-		char *error = malloc(length);
-		va_end(args);
-		va_start(args, format);
 		if (error) {
-			vsnprintf(error, length, format, args);
+			vsnprintf(error, 256, format, args);
```
SirCmpwn: What's this change all about?

sway/swaynag.c
```diff
@@ -0,0 +1,103 @@
...
+#include "sway/swaynag.h"
+
+void swaynag_clone(struct swaynag_instance *dest,
+		struct swaynag_instance *src) {
+	dest->args = src->args;
+	dest->pid = src->pid;
+	dest->fd[0] = src->fd[0];
+	dest->fd[1] = src->fd[1];
+	dest->detailed = src->detailed;
+}
```
SirCmpwn: Why does this exist when memcpy exists

SirCmpwn:
LGTM aside from those two issues

sway/commands.c
RedSoxFan: Reverted it back to what's in `master`. See the strike-through text in the PR description for the original change.

sway/swaynag.c
RedSoxFan: I'm not sure what was going through my head when I did that. Switched to using `memcpy`.

SirCmpwn:
Thanks!

PR #2404, opened 2018-08-02
Allow moving containers when workspace itself is focused

RyanDwyer:
To test, create two views in a workspace, then use `focus parent` to focus the workspace, then run something like `move to workspace 2`.


SirCmpwn:
Thanks!

PR #2405, opened 2018-08-02
Reset signal mask after fork

marienz:
To test: run i3status as swaybar status_command and/or from a terminal launched from sway, `killall -USR1 i3status`. i3status should refresh.

### DIFF
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -4,6 +4,7 @@
 #include <string.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <signal.h>
 #include "sway/commands.h"
 #include "sway/config.h"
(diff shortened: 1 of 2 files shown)


emersion:
Thanks!

emersion:
>wlroots uses wl_event_loop_add_signal to handle SIGUSR1 from Xwayland.

Maybe we should unregister this handler when we got the signal?

(In any case, because of race conditions, this is a good PR to have - maybe we should document in wlroots that wlr_xwayland registers a signal handler since this is global state)

marienz:
I didn't really look at what wlroots and Xwayland do with SIGUSR1. It seemed reasonable to just clear the mask unconditionally (and clear the entire mask instead of just SIGUSR1).

PR #2410, opened 2018-08-02
Fix crash on mouse motion on fullscreen container

minus7:
`container_at` expects a workspace, not the fullscreened container.
Fixes #2409

PR #2411, opened 2018-08-02
Fix pointer events for fullscreen views

emersion:
Leftover from https://github.com/swaywm/sway/pull/2410

To test, open mpv in fullscreen mode and try to use the UI with the mouse.


RyanDwyer:
Thanks!

Issue #2412, opened 2018-08-02
Crashes


minus7:
The following situations cause sway to crash
- [x] Using a `move container to workspace` shortcut on an empty workspace
- [x] Exiting fullscreen mpv on a hidden workspace on a non-active output

RyanDwyer:
Both fixed.

PR #2414, opened 2018-08-03
Fix crash when fullscreen view closes on inactive workspace

RyanDwyer:
When a view unmaps, normally the surviving ancestor (ie. after reaping) needs to be arranged. When a fullscreen view unmaps, it arranges the workspace rather than the surviving ancestor, but didn't handle cases where the workspace itself was reaped. This happens if the workspace is not currently shown and the fullscreen view was the last container on that workspace.

This commit rewrites this part of `view_unmap` so it's more readable, and fixes the crash by not arranging the workspace if it's been reaped. Note that it no longer arranges the output under any circumstance - this wasn't required anyway.

Fixes the second part of #2412.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -592,19 +592,18 @@ void view_unmap(struct sway_view *view) {
 		view->urgent_timer = NULL;
 	}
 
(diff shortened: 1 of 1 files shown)


emersion:
Thanks for this series of fixes!

PR #2415, opened 2018-08-03
Fix crash when moving cursor over a fullscreen split container

RyanDwyer:
Calling `container_at_view` fails an assertion if the container isn't a view. Calling `tiling_container_at` works correctly, as that function checks if the container is a view and calls `container_at_view` if so.

To test:

* Fullscreen mpv and interact with the UI
* Fullscreen a split container where one of them is mpv and interact with the UI


emersion:
Thanks!

PR #2417, opened 2018-08-03
Add missing stdlib.h includes to swaynag

marienz:
Without this swaynag does not build on my system (implicit declaration of
functions like "free"). With these includes added it builds and works.


RyanDwyer:
Thanks!

PR #2418, opened 2018-08-03
Separate root-related code

RyanDwyer:
This creates a `sway/tree/root.c` and moves bits and pieces from elsewhere into it.

* `layout_init` has been renamed to `root_create` and moved into `root.c`
* `root_destroy` has been created and is called on shutdown
* scratchpad code has been moved into `root.c`, because hidden scratchpad containers are stored in the root struct. `scratchpad.c` no longer exists.

I considered making `root_create` return the root container and making `root_destroy` accept the container as an argument, but if I did that then I'd want to make `root_container` a pointer and that involves updating all references to it. Opinions are welcome regarding this.

I was going to move the workspace_pid tracking code into `root.c` as well, but realised that the workspace_pids are not stored in the root container. So I figured I'd leave that out and get feedback.


sway/tree/root.c
```diff
@@ -0,0 +1,147 @@
...
+		output_layout_handle_change;
+	wl_signal_add(&root_container.sway_root->output_layout->events.change,
+		&root_container.sway_root->output_layout_change);
+}
+
+void root_destroy(void) {
+	// sway_root
+	wl_list_remove(&root_container.sway_root->output_layout_change.link);
+	list_free(root_container.sway_root->scratchpad);
+	//wlr_output_layout_destroy(root_container.sway_root->output_layout);
```
RyanDwyer: If I uncomment this line then I reliably get this crash on shutdown (even with no views open):

    Thread 1 "sway" received signal SIGSEGV, Segmentation fault.
    0x00007ffff7ba75f6 in output_destroy (output=0x7ffff7ba7a53 <output_manager_send_details+58>) at ../subprojects/wlroots/types/wlr_xdg_output.c:76
    76		wl_resource_for_each_safe(resource, tmp, &output->resources) {
    (gdb) bt
    #0  0x00007ffff7ba75f6 in output_destroy (output=0x7ffff7ba7a53 <output_manager_send_details+58>) at ../subprojects/wlroots/types/wlr_xdg_output.c:76
    #1  0x00007ffff7ba7cde in wlr_xdg_output_manager_destroy (manager=0x555555976308) at ../subprojects/wlroots/types/wlr_xdg_output.c:242
    #2  0x00007ffff7ba7af8 in handle_layout_destroy (listener=0x555555976350, data=0x555555975de0) at ../subprojects/wlroots/types/wlr_xdg_output.c:203
    #3  0x00007ffff7baa13e in wlr_signal_emit_safe (signal=0x555555975e18, data=0x555555975de0) at ../subprojects/wlroots/util/signal.c:29
    #4  0x00007ffff7b97031 in wlr_output_layout_destroy (layout=0x555555975de0) at ../subprojects/wlroots/types/wlr_output_layout.c:66
    #5  0x0000555555597a6d in root_destroy () at ../sway/tree/root.c:52
    #6  0x000055555556096c in main (argc=4, argv=0x7fffffffe288) at ../sway/main.c:462

Any ideas?

emersion: Don't destroy objects created with a `wl_display` argument - they destroy themselves automaticlly when the display goes down to prevent being used while the display is destroyed.

emersion: Eh, `wlr_output_layout` doesn't have a `wl_display` argument. Then this is a bug in wlroots.

SirCmpwn:
>I was going to move the workspace_pid tracking code into root.c as well, but realised that the workspace_pids are not stored in the root container. So I figured I'd leave that out and get feedback.

It'd probably make emersion happy if you moved this

SirCmpwn:
(I don't care, it's not Wrong where it is)

emersion:
Thanks!

PR #2419, opened 2018-08-03
Check correct ws for sticky and fix floating iter

### LINKED ISSUE #2416: Sticky windows disappears
**Version:**
sway version 1.0-alpha.4-259-gb49904dd (Aug  3 2018, branch 'master')

Can't extract usefull informations from debug log

Sticky windows disappears after switching some workspaces if these don't get focus
**How to reproduce**
- Create a sticky window
- Create tiling window on same workspace
- Switch several times between another workspace and the current

After some switch, sticky windows that have not received focus will disappear

RedSoxFan:
Fixes #2416 

~When switching workspaces, the wrong workspace could be checked for sticky floating containers. This is fixed by using `seat_get_focus_inactive` (followed-by `container_parent`) instead of `seat_get_active_child`.~ See comments below

Additionally, I fixed the iterator for the floating children so the current index is decremented when moving a sticky container to the workspace being switched to.

### DIFF
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -412,6 +412,11 @@ bool workspace_switch(struct sway_container *workspace) {
 			has_sticky = true;
 			container_remove_child(floater);
 			container_add_child(workspace->sway_workspace->floating, floater);
+			if (floater == focus) {
+				seat_set_focus(seat, NULL);
+				seat_set_focus(seat, floater);
+			}
+			--i;
 		}
 	}
 


RyanDwyer:
Can you clarify why `seat_get_active_child` doesn't work? That should return the most recently active direct child of the given container - ie. the active workspace.

RedSoxFan:
> Can you clarify why seat_get_active_child doesn't work? That should return the most recently active direct child of the given container - ie. the active workspace.

When moving the sticky container to the next workspace, the ancestors are not always being added to the focus stack. I'm looking into the exact situation now, but if (all on one output) you:
- open a view in workspace 1
- open another view and make it sticky
- switch to workspace 2
- switch to workspace 3

the sticky view is gone. Workspace 2 was never added to the focus stack so it checks against workspace 1 (the only workspace in the focus stack).

RedSoxFan:
> When moving the sticky container to the next workspace, the ancestors are not always being added to the focus stack

If one of the sticky containers had focus, the workspace was never being added to the focus stack. This is fixed by removing focus from the focused sticky container and then setting the focus back after moving it.

RyanDwyer:
Thanks!

PR #2420, opened 2018-08-04
Implement move to workspace on a floating container

RyanDwyer:
Also adjusts `container_floating_translate` to not change the current properties directly.

I'm not sure what i3 does with regard to centering it within the workspace. I made it retain the same position if it's being moved to a workspace on the same output. If it's being moved to a workspace on another output then it will center within that workspace.

Tested with both floating views and floating containers.


sway/tree/container.c
```diff
@@ -1241,6 +1236,50 @@ void container_floating_move_to(struct sway_container *con,
...
+	}
+	if (!new_ws) {
+		// This can happen if the user has run "move container to mark foo",
+		// where mark foo is on a hidden scratchpad container.
+		return;
+	}
+	struct sway_container *old_ws = container_parent(container, C_WORKSPACE);
+	if (old_ws != new_ws) {
+		container_remove_child(container);
+		container_add_child(new_ws->sway_workspace->floating, container);
```
RedSoxFan: Needs a reparent signal event here so that the textures get updated when moved to an output with a different scale

RyanDwyer:
This is ready for review again. BTW, I wrote some details in the commit log for "Deny moving a sticky container".

SirCmpwn:
Thanks!

PR #2422, opened 2018-08-05
Fix compiler errors.

ggreer:
- Some platforms don't expose kill() unless _POSIX_C_SOURCE is defined.
- fork(), execl(), and setsid() need unistd.h on some platforms.

Basically, this fixes some platform-specific build errors.


emersion:
Thanks!

PR #2423, opened 2018-08-05
[RFC] Fix segv when sway is terminating

manio:
I've got the following SIGSEGV when terminating sway:
```
0x00005607dc603af5 in view_unmap (view=0x5607dcb3d350) at ../sway/tree/view.c:599
599		if (surviving_ancestor->type >= C_WORKSPACE) {
```
When I was printing the surviving_ancestor, it was NULL at that time


RedSoxFan:
Thanks!

PR #2428, opened 2018-08-06
Move workspace moving code out of container_move_to

RyanDwyer:
`container_move_to` handled moving containers to new parents, as well as moving workspaces to new outputs.

This commit removes the workspace-moving code from this function and introduces `workspace_move_to_output`. Moving workspaces using `container_move_to` only happened from the move command, so it's been implemented as a static function in that file.

Simplifying `container_move_to` makes it easier for me to fix some issues in #2420.


RedSoxFan:
Thanks!

PR #2432, opened 2018-08-07
Fix crash when running `move container to workspace back_and_forth`

RyanDwyer:
The back_and_forth condition is intended to be handled in the else-if block, but this was never reached because it remained in the first block's conditions.

To test, run `move container to workspace back_and_forth` when no other workspaces exist.


SirCmpwn:
Thanks!

PR #2433, opened 2018-08-07
Fix infinite loop when focusing sticky containers via workspace command

RyanDwyer:
In a multi-output setup, if a sticky container is on one output and focus is on the other output, and you run (eg) `workspace 1` to focus the workspace containing the sticky container, an infinite loop would occur. It would loop infinitely because it would remove the sticky container from the workspace, add it back to the same workspace, and then decrement the iterator variable.

The fix just wraps the loop in a workspace comparison.


SirCmpwn:
Thanks!

PR #2436, opened 2018-08-08
Fix focus when clicking floating decorations

### LINKED ISSUE #2431: Clicking on titlebar focuses parent container

RyanDwyer:
Fixes #2431.

It's not right for `container_at_view` to only return the swayc if a surface was clicked. It needs to also return the swayc if the decorations are clicked.

Note that `container_at_view` will now always return the swayc, so it assumes you've done a box check already. This seems like a safe assumption as I can't fault it.

To test, float a view which has title decorations then click the title. The floating view should be become focused.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -429,6 +429,9 @@ static void dispatch_cursor_button_floating(struct sway_cursor *cursor,
 		struct sway_container *cont) {
 	struct sway_seat *seat = cursor->seat;
 
+	seat_set_focus(seat, cont);
+	seat_pointer_notify_button(seat, time_msec, button, state);
+
 	// Deny moving or resizing a fullscreen container
 	if (container_is_fullscreen_or_child(cont)) {
 		seat_pointer_notify_button(seat, time_msec, button, state);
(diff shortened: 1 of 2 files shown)


sway/tree/container.c
```diff
@@ -567,9 +567,8 @@ static struct sway_container *container_at_view(struct sway_container *swayc,
 		*sx = _sx;
 		*sy = _sy;
 		*surface = _surface;
-		return swayc;
 	}
-	return NULL;
+	return swayc;
```
emersion: Hmm. So maybe we should make this function return `void` now, since it doesn't have the same semantics as other functions with the same signature?

RyanDwyer: Hmm. I'd favour renaming the function to `surface_at_view` and then having it return void.

emersion: Yeah, that seems sensible.

RyanDwyer:
I renamed the `container_at_view` function, and also found and fixed another focus issue where if the floating view didn't have focus and you clicked the title, it would interpret it as a move event and wouldn't focus the view. So now it sends the focus when starting a move or resize operation.

sway/input/cursor.c
```diff
@@ -429,6 +429,9 @@ static void dispatch_cursor_button_floating(struct sway_cursor *cursor,
 		struct sway_container *cont) {
 	struct sway_seat *seat = cursor->seat;
 
+	seat_set_focus(seat, cont);
+	seat_pointer_notify_button(seat, time_msec, button, state);
```
emersion: I don't think we should send a button event here. If you want to move a window and you drag over a button, it'll click the button.

RyanDwyer: Ah, good catch.

emersion:
Thanks!

PR #2439, opened 2018-08-08
Allow a fallback color to be specified for swaybg

### LINKED ISSUE #2270: Missing background image fails the output command
When you set the background image in the same line as the output position, e.g. `output VGA-1 resolution 1920x1080 position 5120,360 background /usr/local/share/backgrounds/sway/Sway_Wallpaper_Blue_1920x1080.png fill` whole command fails when the image does not exist.

I kinda think it could ignore the image then and still configure the layout as instructed, but I wouldn't go as far as to argue for it

### LINKED ISSUE #2434: Set background color AND image
In my configuration, I have these lines:
```
output * bg #000000 solid_color
output * bg wallpaper.png fit
```
Because of the `fit`, the image does not cover the screen (intended). The rest of the screen is coloured with `#404040`. To harmonize the image and colour, I tried to change the background colour BEFORE applying the image, but it does not work. Is there a way to achieve this?

RedSoxFan:
Closes #2434 
Fixes #2270

This allows for a color to be optionally set when the wallpaper does not fill the entire output. If specified, the fallback color is also used when the image path is inaccessible.

I opted to use the following syntax
```
output <name> bg <path> <mode> [<fallback_color>]
```
instead of the following from #2434
```
output <name> bg <color> solid_color
output <name> bg <path> <mode>
```

This allows for both solid colors and images (with an optional fallback color) to be set without affecting the other when changed via commands further down the config or swaymsg.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -151,6 +151,7 @@ struct output_config {
 
 	char *background;
 	char *background_option;
+	char *background_fallback;
 	enum config_dpms dpms_state;
 };
 
--- a/sway/commands/output/background.c
+++ b/sway/commands/output/background.c
@@ -6,6 +6,7 @@
 #include <errno.h>
 #include "sway/commands.h"
 #include "sway/config.h"
+#include "sway/swaynag.h"
 #include "log.h"
 #include "stringop.h"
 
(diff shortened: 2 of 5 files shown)


SirCmpwn:
Thanks!

PR #2441, opened 2018-08-09
Finish resize commands

RyanDwyer:
Completes #1737.

* Implements `resize grow|shrink <direction> <amount>` for tiled containers
* Implements `resize set <width> <height>` for tiled containers

I made `resize set` allow px units, then realised afterwards that i3 doesn't support it. So I left it in. Use `px` to use pixels, and `ppt` or no unit to use percentage points.

Also fixes a sane size check which failed to check against the correct amount of pixels in some cases.


ianyfan:
Does the man page need to be updated?

RedSoxFan:
Thanks!

PR #2445, opened 2018-08-10
Implement resizing tiled containers via cursor

RyanDwyer:
* The `OP_RESIZE` seat operation has been renamed to `OP_RESIZE_FLOATING`, and `OP_RESIZE_TILING` has been introduced.
* Similar to the above, `seat_begin_resize` and `handle_resize_motion` have been renamed and tiling variants introduced.
* `resize.c`'s `resize_tiled` has to be used, so `container_resize_tiled` has been introduced in `resize.c` to allow external code to call it. Ideally all this would be in `container.c`, but I didn't want to make this PR unnecessarily big.

Things to test:

* Create two tiled views in horizontal/vertical layout and resize all internal edges
* Float a container with two or more views inside it and resize them
* Check for regressions in the resize/move stuff that's already implemented

BTW, it would be good if the cursor icon could be something like this: 
image

...but this would have to be added to wlroots.

### DIFF
--- a/common/list.c
+++ b/common/list.c
@@ -77,6 +77,15 @@ int list_seq_find(list_t *list, int compare(const void *item, const void *data),
 	return -1;
 }
 
+int list_find(list_t *list, void *item) {
+	for (int i = 0; i < list->length; i++) {
+		if (list->items[i] == item) {
+			return i;
+		}
+	}
+	return -1;
+}
+
 void list_swap(list_t *list, int src, int dest) {
 	void *tmp = list->items[src];
 	list->items[src] = list->items[dest];
--- a/include/list.h
+++ b/include/list.h
@@ -20,6 +20,7 @@ void list_qsort(list_t *list, int compare(const void *left, const void *right));
 // Return index for first item in list that returns 0 for given compare
 // function or -1 if none matches.
 int list_seq_find(list_t *list, int compare(const void *item, const void *cmp_to), const void *cmp_to);
+int list_find(list_t *list, void *item);
 // stable sort since qsort is not guaranteed to be stable
 void list_stable_sort(list_t *list, int compare(const void *a, const void *b));
 // swap two elements in a list
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -1,6 +1,7 @@
 #ifndef _SWAY_COMMANDS_H
 #define _SWAY_COMMANDS_H
 
+#include <wlr/util/edges.h>
 #include "config.h"
 
 typedef struct cmd_results *sway_cmd(int argc, char **argv);
@@ -83,6 +84,12 @@ char *cmd_results_to_json(struct cmd_results *results);
 
 struct cmd_results *add_color(const char *name,
 		char *buffer, const char *color);
+
+/**
+ * TODO: Move this function and its dependent functions to container.c.
+ */
+void container_resize_tiled(struct sway_container *parent, enum wlr_edges edge,
+		int amount);
 
 sway_cmd cmd_assign;
 sway_cmd cmd_bar;
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -57,7 +57,8 @@ struct sway_seat {
 	enum {
 		OP_NONE,
 		OP_MOVE,
-		OP_RESIZE,
+		OP_RESIZE_FLOATING,
+		OP_RESIZE_TILING,
 	} operation;
 
 	struct sway_container *op_container;
@@ -159,8 +160,11 @@ void drag_icon_update_position(struct sway_drag_icon *icon);
 void seat_begin_move(struct sway_seat *seat, struct sway_container *con,
 		uint32_t button);
 
-void seat_begin_resize(struct sway_seat *seat, struct sway_container *con,
-		uint32_t button, enum wlr_edges edge);
+void seat_begin_resize_floating(struct sway_seat *seat,
+		struct sway_container *con, uint32_t button, enum wlr_edges edge);
+
+void seat_begin_resize_tiling(struct sway_seat *seat,
+		struct sway_container *con, uint32_t button, enum wlr_edges edge);
 
 void seat_end_mouse_operation(struct sway_seat *seat);
 
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -158,8 +158,8 @@ static int parallel_size(struct sway_container *c, enum resize_axis a) {
 	return normalize_axis(a) == RESIZE_AXIS_HORIZONTAL ? c->width : c->height;
 }
 
-static void resize_tiled(int amount, enum resize_axis axis) {
-	struct sway_container *parent = config->handler_context.current_container;
+static void resize_tiled(struct sway_container *parent, int amount,
+		enum resize_axis axis) {
 	struct sway_container *focused = parent;
 	if (!parent) {
 		return;
@@ -297,6 +297,28 @@ static void resize_tiled(int amount, enum resize_axis axis) {
 	arrange_windows(parent->parent);
 }
 
+void container_resize_tiled(struct sway_container *parent,
+		enum wlr_edges edge, int amount) {
+	enum resize_axis axis = RESIZE_AXIS_INVALID;
+	switch (edge) {
+	case WLR_EDGE_TOP:
+		axis = RESIZE_AXIS_UP;
+		break;
+	case WLR_EDGE_RIGHT:
+		axis = RESIZE_AXIS_RIGHT;
+		break;
+	case WLR_EDGE_BOTTOM:
+		axis = RESIZE_AXIS_DOWN;
+		break;
+	case WLR_EDGE_LEFT:
+		axis = RESIZE_AXIS_LEFT;
+		break;
+	case WLR_EDGE_NONE:
+		break;
+	}
+	resize_tiled(parent, amount, axis);
+}
+
 /**
  * Implement `resize <grow|shrink>` for a floating container.
  */
@@ -398,7 +420,7 @@ static struct cmd_results *resize_adjust_tiled(enum resize_axis axis,
 		}
 	}
 
-	resize_tiled(amount->amount, axis);
+	resize_tiled(current, amount->amount, axis);
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
 
@@ -421,7 +443,8 @@ static struct cmd_results *resize_set_tiled(struct sway_container *con,
 			}
 		}
 		if (width->unit == RESIZE_UNIT_PX) {
-			resize_tiled(width->amount - con->width, RESIZE_AXIS_HORIZONTAL);
+			resize_tiled(con, width->amount - con->width,
+					RESIZE_AXIS_HORIZONTAL);
 		}
 	}
 
@@ -439,7 +462,8 @@ static struct cmd_results *resize_set_tiled(struct sway_container *con,
 			}
 		}
 		if (height->unit == RESIZE_UNIT_PX) {
-			resize_tiled(height->amount - con->height, RESIZE_AXIS_VERTICAL);
+			resize_tiled(con, height->amount - con->height,
+					RESIZE_AXIS_HORIZONTAL);
 		}
 	}
 
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -213,6 +213,18 @@ static void transaction_apply(struct sway_transaction *transaction) {
 }
 
 static void transaction_progress_queue() {
+	// Check if any transactions in the queue are in a partially ready state.
+	// If so, we shouldn't progress any transactions, even ones which are fully
+	// ready at the front of the queue, because the views in the ready
+	// transactions might have committed past it to a transaction which isn't
+	// yet ready.
+	for (int i = 0; i < server.transactions->length; ++i) {
+		struct sway_transaction *transaction = server.transactions->items[i];
+		if (transaction->num_waiting != 0 &&
+				transaction->num_waiting != transaction->num_configures) {
+			return;
+		}
+	}
 	while (server.transactions->length) {
 		struct sway_transaction *transaction = server.transactions->items[0];
 		if (transaction->num_waiting) {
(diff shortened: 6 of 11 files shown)


RyanDwyer:
I've refactored the dispatch function (see commit message for details) then implemented resizing tiled views via mod.

emersion:
>BTW, it would be good if the cursor icon could be something like this

You can try `col-resize` and `row-resize`. You also have diagonal versions like `nwse-resize`.

You can try all of this by running the "Cursors" demo in `gtk3-demo`.

SirCmpwn:
>If there is a floating view/container over a tiled one, the tiled one gets the resize event instead of the floating one.
>It's no longer possible to resize a floating container.

Unable to reproduce.

My own issues:

- Doesn't seem atomic?
- All right click menus on GTK+ and Xwayland apps no longer appear

RyanDwyer:
> You can try col-resize and row-resize.

Done.

> If there is a floating view/container over a tiled one, the tiled one gets the resize event instead of the floating one.
> It's no longer possible to resize a floating container.

I also can't reproduce either of these.

> Doesn't seem atomic?

Which clients are you testing with and what are you doing exactly?

> All right click menus on GTK+ and Xwayland apps no longer appear

Fixed. Was missing a `mod_pressed` check. And while fixing this I noticed I was missing `state == WLR_BUTTON_PRESSED` checks.

common/list.c
```diff
@@ -77,6 +77,15 @@ int list_seq_find(list_t *list, int compare(const void *item, const void *data),
...
 }
 
+int list_find(list_t *list, void *item) {
+	for (int i = 0; i < list->length; i++) {
+		if (list->items[i] == item) {
+			return i;
+		}
+	}
+	return -1;
+}
```
SirCmpwn: I think this should just be a static function in that one file.

RedSoxFan:
> If there is a floating view/container over a tiled one, the tiled one gets the resize event instead of the floating one.

- Open two views in a `spltith`
- Open a third view.
  - Float it
  - Create a `splitv`
  - Open another view in the floating container
- Using floating_mod + right click, try to change the height of the third and fourth views at their split

> It's no longer possible to resize a floating container.

Resizing a floating *_view_* is working fine. I cannot resize any floating *_container_* though.

common/list.c
RyanDwyer: This function can be used in several other places. My plan is to submit a PR after this one which changes these:

* https://github.com/swaywm/sway/blob/e474d87e42d625eb4e0a41a0559bbf0e4ec5ecc5/sway/tree/root.c#L88
* https://github.com/swaywm/sway/blob/e474d87e42d625eb4e0a41a0559bbf0e4ec5ecc5/sway/tree/workspace.c#L296
* https://github.com/swaywm/sway/blob/e474d87e42d625eb4e0a41a0559bbf0e4ec5ecc5/sway/tree/workspace.c#L334
* https://github.com/swaywm/sway/blob/e474d87e42d625eb4e0a41a0559bbf0e4ec5ecc5/sway/tree/workspace.c#L343
* https://github.com/swaywm/sway/blob/e474d87e42d625eb4e0a41a0559bbf0e4ec5ecc5/sway/tree/layout.c#L25
* https://github.com/swaywm/sway/blob/e474d87e42d625eb4e0a41a0559bbf0e4ec5ecc5/sway/tree/layout.c#L129

RyanDwyer:
@RedSoxFan I was able to reproduce with your instructions. I've made it so doing mod+resize within a floating split container will now resize the container itself rather than the split. And changing that also fixes the other issue.

emersion:
>Doesn't seem atomic?

Yeah I have this issue too. Open a bunch of tiled terminals (termite), resize them quickly with the mouse: I can see flashes of my background.

RyanDwyer:
I suspect what's happening here is some of the views are committing to transactions ahead of what the others have committed to, so when we progress the queue the "ahead" view is being rendered. Here it is step by step:

* Assume two views, both with state A.
* You start resizing a view which creates pending state B and transaction B. Buffers at the current state (A) are saved and rendered while we wait for the views to commit with state B.
* View 1 commits state B. Because we're still waiting on the other view to commit, we continue to render both saved buffers.
* Another motion of resize is dispatched, so we create pending state C and transaction C.
* View 2 commits state C. At this point transaction B is complete so we apply transaction B, drop the buffers and save the current buffers. But view 2 is already at state C, so we've saved a buffer that isn't the right size.
* We render the saved buffers while waiting for view 1 to commit state C, which is imperfect.

I've added a commit which changes the transaction queue logic. It now won't progress the queue if there are any transactions in it which have some (but not all) views committed. This prevents it from progressing past transaction B while C is partially complete, thus avoiding the scenario where we read the wrong buffer.

It appears to have fixed the atomic issue described, and I don't experience any choppy behaviour (using 4 instances of termite in a 2x2 layout).

SirCmpwn:
Thanks for the updates, pulling this locally and will be running the branch for a couple of days.

SirCmpwn:
It's no longer choppy (every frame looks right) but it is really laggy, it feels like windows are catching up to my mouse.

https://sr.ht/9rXA.mkv

RyanDwyer:
There's one more thing I can try. When there's multiple transactions, we could delay sending the configures for the subsequent transactions until the previous transaction has completed. We could be smart about it, so if there's a bunch of transactions which all involve the same views then we'd skip ahead to the last transaction. It might still be laggy, but I think it'd be better than what it currently is.

SirCmpwn:
Well, we can give it a shot.

Do you think at some point we should just give up and show a stretched window if the client hasn't responded soon enough so that it keeps up with the mouse?

RyanDwyer:
If what I'm about to implement doesn't work then the options are to a) have perfect but laggy rendering, b) have imperfect but smooth rendering, or c) resort to something like stretching the surface.

For views that are shrinking I think cropping it is better than stretching it. For views that are expanding... I guess we'd have to stretch that :-\

We'd also need to fine tune it. I don't think we'd want to use stretching for standard layout updates, only for the interactive resizing, and that means we'd have to flag those transactions/containers as such.

SirCmpwn:
>If what I'm about to implement doesn't work then the options are to a) have perfect but laggy rendering, b) have imperfect but smooth rendering, or c) resort to something like stretching the surface.

Do you think we can find a middle ground? Some of my clients respond fast enough to resize smoothly, and others don't. We could resort to being imperfect only if the lag would be severe.

>For views that are shrinking I think cropping it is better than stretching it. For views that are expanding... I guess we'd have to stretch that :-\

+1

>We'd also need to fine tune it. I don't think we'd want to use stretching for standard layout updates, only for the interactive resizing, and that means we'd have to flag those transactions/containers as such.

+1

SirCmpwn:
Sending a window to the scratchpad causes sway to crash:

```
#0  0x0000562f2a287c23 in container_has_ancestor (descendant=0x0, ancestor=0x562f2a2b0b80 <root_container>) at ../sway/tree/container.c:811
#1  0x0000562f2a26a523 in seat_get_focus_by_type (seat=0x562f2a554f60, container=0x562f2a2b0b80 <root_container>, type=C_TYPES, only_tiling=false) at ../sway/input/seat.c:148
        floating = 0x0
        current = 0x562f2aabc910
#2  0x0000562f2a26c4f7 in seat_get_focus_inactive (seat=0x562f2a554f60, container=0x562f2a2b0b80 <root_container>) at ../sway/input/seat.c:882
#3  0x0000562f2a26c5de in seat_get_focus (seat=0x562f2a554f60) at ../sway/input/seat.c:906
#4  0x0000562f2a26b93a in seat_set_focus_warp (seat=0x562f2a554f60, container=0x562f2a66bb60, warp=true, notify=true) at ../sway/input/seat.c:625
        last_focus = 0x0
        last_workspace = 0x0
        new_workspace = 0x7ffdd164b6f0
        last_output = 0x562f2a28511f <arrange_children_of+303>
        new_output = 0x562f2aabc740
        new_output_last_ws = 0x0
#5  0x0000562f2a26c043 in seat_set_focus (seat=0x562f2a554f60, container=0x562f2a66bb60) at ../sway/input/seat.c:789
#6  0x0000562f2a28d283 in root_scratchpad_add_container (con=0x562f2aabc740) at ../sway/tree/root.c:79
        __PRETTY_FUNCTION__ = "root_scratchpad_add_container"
        parent = 0x562f2a66bb60
        seat = 0x562f2a554f60
#7  0x0000562f2a27a2af in move_to_scratchpad (con=0x562f2aabc740) at ../sway/commands/move.c:456
#8  0x0000562f2a27a532 in cmd_move (argc=1, argv=0x562f2a997ae8) at ../sway/commands/move.c:486
        error = 0x0
        current = 0x562f2aabc740
#9  0x0000562f2a253508 in execute_command (_exec=0x562f2a573260 "move scratchpad", seat=0x562f2a554f60) at ../sway/commands.c:300
        res = 0x200000002
        argc = 2
        argv = 0x562f2a997ae0
        handler = 0x562f2a2af010 <command_handlers+112>
        results = 0x0
        exec = 0x562f2aac52a0 "move scratchpad"
        head = 0x0
        cmdlist = 0x0
        cmd = 0x562f2aac52a0 "move scratchpad"
        views = 0x0
        __PRETTY_FUNCTION__ = "execute_command"
#10 0x0000562f2a27500c in seat_execute_command (seat=0x562f2a554f60, binding=0x562f2a573310) at ../sway/commands/bind.c:323
        binding_copy = 0x562f2a573310
        reload = false
        results = 0x562f2a573310
#11 0x0000562f2a270709 in handle_keyboard_key (listener=0x562f2a91dd30, data=0x7ffdd164ba00) at ../sway/input/keyboard.c:281
        binding_pressed = 0x562f2a573310
        keyboard = 0x562f2a91dd20
        seat = 0x562f2a554f60
        wlr_seat = 0x562f2a555080
        wlr_device = 0x562f2a8fa540
        event = 0x7ffdd164ba00
        input_inhibited = false
        keycode = 28
        translated_keysyms = 0x562f2a92b1d8
        translated_modifiers = 64
        translated_keysyms_len = 1
        raw_keysyms = 0x562f2a92b1b8
        raw_modifiers = 65
        raw_keysyms_len = 1
        code_modifiers = 65
        handled = false
        binding_released = 0x0
        next_repeat_binding = 0x0
#12 0x00007f36eb3bd60e in wlr_signal_emit_safe (signal=signal@entry=0x562f2a91dcc8, data=data@entry=0x7ffdd164ba00) at ../util/signal.c:29
        pos = 0x562f2a91dd30
        l = 0x562f2a91dd30
        cursor = {link = {prev = 0x562f2a91dd30, next = 0x7ffdd164b980}, notify = 0x7f36eb3bd580 <handle_noop>}
        end = {link = {prev = 0x7ffdd164b960, next = 0x562f2a91dcc8}, notify = 0x7f36eb3bd580 <handle_noop>}
#13 0x00007f36eb3ad026 in wlr_keyboard_notify_key (keyboard=0x562f2a91dbd0, event=event@entry=0x7ffdd164ba00) at ../types/wlr_keyboard.c:107
        updated = <optimized out>
#14 0x00007f36eb390a17 in handle_keyboard_key (event=<optimized out>, libinput_dev=<optimized out>) at ../backend/libinput/keyboard.c:73
        wlr_dev = 0x562f2a8fa540
        kbevent = 0x562f2aac3050
        wlr_event = {time_msec = 5881289, keycode = 20, update_state = true, state = WLR_KEY_PRESSED}
        state = <optimized out>
#15 0x00007f36eb38fb4c in handle_libinput_readable (fd=<optimized out>, mask=<optimized out>, _backend=0x562f2a378ab0) at ../backend/libinput/backend.c:35
        backend = 0x562f2a378ab0
        event = 0x562f2aac3050
#16 0x00007f36eb5f1a42 in wl_event_loop_dispatch () at /home/sircmpwn/sources/sway/build/sway/../../../../../../lib64/libwayland-server.so.0
#17 0x00007f36eb5f014c in wl_display_run () at /home/sircmpwn/sources/sway/build/sway/../../../../../../lib64/libwayland-server.so.0
#18 0x0000562f2a25ea7f in server_run (server=0x562f2a2b0a00 <server>) at ../sway/server.c:163
#19 0x0000562f2a25e202 in main (argc=3, argv=0x7ffdd164bde8) at ../sway/main.c:439
        verbose = 0
        debug = 1
        validate = 0
        long_options = 
            {{name = 0x562f2a29a9cc "help", has_arg = 0, flag = 0x0, val = 104}, {name = 0x562f2a29a9d1 "config", has_arg = 1, flag = 0x0, val = 99}, {name = 0x562f2a29a9d8 "validate", has_arg = 0, flag = 0x0, val = 67}, {name = 0x562f2a29a9e1 "debug", has_arg = 0, flag = 0x0, val = 100}, {name = 0x562f2a29a9e7 "version", has_arg = 0, flag = 0x0, val = 118}, {name = 0x562f2a29a9ef "verbose", has_arg = 0, flag = 0x0, val = 86}, {name = 0x562f2a29a9f7 "get-socketpath", has_arg = 0, flag = 0x0, val = 112}, {name = 0x0, has_arg = 0, flag = 0x0, val = 0}}
        config_path = 0x0
        usage = 0x562f2a29a550 "Usage: sway [options] [command]\n\n  -h, --help", ' ' <repeats 13 times>, "Show help message and quit.\n  -c, --config <config>  Specify a config file.\n  -C, --validate         Check the validity of the config file, th"...
        c = -1
        suid = false
```

RyanDwyer:
It now does the one transaction at a time thing. I'm getting smooth resizing with termite + nautilus + gedit as you have in your video. I'm pretty pleased with this, assuming I haven't introduced any instability.

If this implementation of transactions is what we'll use, then I'll also be able to simplify it. For example, views won't need to store a list of transactions they're involved with because there's only one at a time. But I'll do that after this PR is merged.

SirCmpwn:
>It now does the one transaction at a time thing. I'm getting smooth resizing with termite + nautilus + gedit as you have in your video. I'm pretty pleased with this, assuming I haven't introduced any instability.

Much better, nice! I still get some stuttering, though. It doesn't help that I'm testing this on a 4K display. I would merge it with this level of performance.

SirCmpwn:
By the way, to stress-test this I open `(splitv (splith (natilus) (urxvt)) (gedit))` and just go fuckin nuts dragging urxvt around. Just moving my mouse in a circle around the output is enough to cause severe lag.

SirCmpwn:
Thanks!@

PR #2446, opened 2018-08-10
xwayland: fix unmanaged surfaces order

### LINKED ISSUE #2349: Tooltips of popovers are rendered out of order in Firefox
1. Open Firefox
2. Open the extensions overflow menu <kbd>»</kbd>
3. Move the mouse over an item
4. The tooltip shows under the popover

~~Additionally the popovers is not re-focused when the tooltip is destroyed, so the popover is closed immediately.~~ this part has been fixed

### LINKED ISSUE #2425: Dialogs hidden under their parent
Hi,
When a GTK application, e.g. GIMP, launches multiple pop-up windows, the "first" (larger) pop-up window always obscures the second (smaller) pop-up. For example, to reproduce:

1. Open an image in GIMP, and hit File -> Export
2. File dialog opens. Enter a *.jpg file name
3. A second, smaller pop-up to set the JPEG quality appears
4. As soon as you move your mouse, the larger pop-up window grabs focus, moves to the top (obscuring the second pop-up window).

Not sure if this is a sway / wlroots issue or not, but the same build of GIMP doesn't exhibit this problem under e.g. GNOME Wayland, KDE, or X11.

*For those stuck with same problem: I temporarily work around it by toggling the top, focused window to tiling mode (`$Mod + Shift + Space`) - which then reveals the obscured, smaller floating window.*

emersion:
Fixes #2349

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -59,7 +59,7 @@ static void unmanaged_handle_map(struct wl_listener *listener, void *data) {
 		wl_container_of(listener, surface, map);
 	struct wlr_xwayland_surface *xsurface = surface->wlr_xwayland_surface;
 
-	wl_list_insert(&root_container.sway_root->xwayland_unmanaged,
+	wl_list_insert(root_container.sway_root->xwayland_unmanaged.prev,
 		&surface->link);
 
 	wl_signal_add(&xsurface->surface->events.commit, &surface->commit);


SirCmpwn:
Does this or does it not fix #2425?

emersion:
Nope, #2425 is about the order of dialogs, not popups (and I'm not sure it's a bug).

SirCmpwn:
Thanks!

PR #2448, opened 2018-08-11
Remove container_has_child

RyanDwyer:
In all cases you can use `container_has_ancestor` with the arguments swapped, which is faster than `container_has_child`.


emersion:
Thanks!

PR #2449, opened 2018-08-11
Remove container_for_each_descendant_bfs

RyanDwyer:
The function was not used.

Also renames `container_for_each_descendant_dfs` to just `container_for_each_descendant`.


SirCmpwn:
Thanks!

PR #2452, opened 2018-08-11
Man page: Correct argument format for output position config

janza:


minus7:
Using space instead of `,` is deprecated afaik; but you'll have to put the quotes around it, otherwise the comma is interpreted as command separator when using swaymsg and binds

SirCmpwn:
Thanks!

@minus7 that's what this patch fixed

SirCmpwn:
Oh, I misunderstood your comment. Using space is not deprecated afaik...

ianyfan:
Seems like it is: https://github.com/swaywm/sway/blob/146cc0a441f6901eeba7df691098f71d0de73a53/sway/commands/output/position.c#L29

In any case, both should be documented, perhaps with a note if using a space deprecated.

SirCmpwn:
We should prefer spaces imo. I use spaces in my config. We should also not document deprecated options.

PR #2453, opened 2018-08-11
More commands

ianyfan:
Ref #2336 

- [x] nop
- [x] fix set
- [x] fix some workspace shenanigans
- [x] fix exec quoting
- [x] assign

Don't think I will do `popup_during_fullscreen`, ~`force_display_urgency_hint`~ or `force_on_window_activation` in this PR since it's mainly just fixing things

### DIFF
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -7,10 +7,11 @@
 #include "tree/view.h"
 
 enum criteria_type {
-	CT_COMMAND          = 1 << 0,
-	CT_ASSIGN_OUTPUT    = 1 << 1,
-	CT_ASSIGN_WORKSPACE = 1 << 2,
-	CT_NO_FOCUS         = 1 << 3,
+	CT_COMMAND                 = 1 << 0,
+	CT_ASSIGN_OUTPUT           = 1 << 1,
+	CT_ASSIGN_WORKSPACE        = 1 << 2,
+	CT_ASSIGN_WORKSPACE_NUMBER = 1 << 3,
+	CT_NO_FOCUS                = 1 << 4,
 };
 
 struct criteria {
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -52,6 +52,10 @@ static struct cmd_results *focus_mode(struct sway_container *con,
 	}
 	if (new_focus) {
 		seat_set_focus(seat, new_focus);
+	} else {
+		return cmd_results_new(CMD_FAILURE, "focus",
+				"Failed to find a %s container in workspace",
+				floating ? "floating" : "tiling");
 	}
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -1,4 +1,5 @@
 #define _XOPEN_SOURCE 500
+#include <ctype.h>
 #include <stdbool.h>
 #include <string.h>
 #include <strings.h>
@@ -124,7 +125,11 @@ static struct cmd_results *cmd_move_container(struct sway_container *current,
 					return cmd_results_new(CMD_INVALID, "move",
 							expected_syntax);
 				}
-				ws_name = strdup(argv[3]);
+				if (!isdigit(argv[3][0])) {
+					return cmd_results_new(CMD_INVALID, "move",
+							"Invalid workspace number '%s'", argv[3]);
+				}
+				ws_name = join_args(argv + 3, argc - 3);
 				ws = workspace_by_number(ws_name);
 			} else {
 				ws_name = join_args(argv + 2, argc - 2);
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -22,7 +22,7 @@
 static const char *expected_syntax =
 	"Expected 'move <left|right|up|down> <[px] px>' or "
 	"'move [--no-auto-back-and-forth] <container|window> [to] workspace <name>' or "
-	"'move [--no-auto-back-and-forth] <container|window|workspace> [to] output <name|direction>' or "
+	"'move <container|window|workspace> [to] output <name|direction>' or "
 	"'move <container|window> [to] mark <mark>'";
 
 static struct sway_container *output_in_direction(const char *direction,
--- a/sway/commands/sticky.c
+++ b/sway/commands/sticky.c
@@ -36,5 +36,21 @@ struct cmd_results *cmd_sticky(int argc, char **argv) {
 
 	container->is_sticky = wants_sticky;
 
+	if (wants_sticky) {
+		// move container to focused workspace
+		struct sway_container *output = container_parent(container, C_OUTPUT);
+		struct sway_seat *seat = input_manager_current_seat(input_manager);
+		struct sway_container *focus = seat_get_focus_inactive(seat, output);
+		struct sway_container *focused_workspace = container_parent(focus, C_WORKSPACE);
+		struct sway_container *current_workspace = container_parent(container, C_WORKSPACE);
+		if (current_workspace != focused_workspace) {
+			container_move_to(container, focused_workspace);
+			arrange_windows(focused_workspace);
+			if (!container_reap_empty(current_workspace)) {
+				arrange_windows(current_workspace);
+			}
+		}
+	}
+
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -51,7 +51,7 @@ The following commands may only be used in the configuration file.
 	*wordexp*(3) for details). The same include file can only be included once;
 	subsequent attempts will be ignored.
 
-*set* <name> <value>
+*set* $<name> <value>
 	Sets variable $_name_ to _value_. You can use the new variable in the
 	arguments of future commands.
 
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -219,12 +219,19 @@ They are expected to be used with *bindsym* or at runtime through *swaymsg*(1).
 The following commands may be used either in the configuration file or at
 runtime.
 
-*assign* <criteria> [→] <workspace>
+*assign* <criteria> [→] [workspace] [number] <workspace>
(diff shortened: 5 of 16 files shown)


sway/sway.5.scd
```diff
@@ -51,7 +51,7 @@ The following commands may only be used in the configuration file.
 	*wordexp*(3) for details). The same include file can only be included once;
 	subsequent attempts will be ignored.
 
-*set* <name> <value>
+*set* $<name> <value>
```
SirCmpwn: The preferred format is without the symbol

SirCmpwn: It seems the symbol is used in the default config, so I guess we'll just go with that.

ianyfan:
Ready for review, I think.

In particular, please see note about assigning sticky windows.

sway/commands/move.c
```diff
@@ -124,6 +124,10 @@ static struct cmd_results *cmd_move_container(struct sway_container *current,
 					return cmd_results_new(CMD_INVALID, "move",
 							expected_syntax);
 				}
+				if (argv[3][0] < '0' || argv[3][0] > '9') {
+					return cmd_results_new(CMD_INVALID, "move",
+							"Invalid workspace number '%s'", argv[3]);
```
SirCmpwn: Can we not use workspace 10? 11? etc?

ianyfan: This just checks that the first character is a number, and if it isn't, bail. If it is, the logic is handled by `get_workspace_by_number`.

SirCmpwn: Oh, duh. Can you use `isdigit` from `ctype.h` instead?

ianyfan:
Made a quick addition to allow `move container to workspace number <number><name>`

include/sway/criteria.h
```diff
@@ -7,10 +7,11 @@
...
 
 enum criteria_type {
-	CT_COMMAND          = 1 << 0,
-	CT_ASSIGN_OUTPUT    = 1 << 1,
-	CT_ASSIGN_WORKSPACE = 1 << 2,
-	CT_NO_FOCUS         = 1 << 3,
+	CT_COMMAND                 = 1 << 0,
+	CT_ASSIGN_OUTPUT           = 1 << 1,
+	CT_ASSIGN_WORKSPACE        = 1 << 2,
+	CT_ASSIGN_WORKSPACE_NUMBER = 1 << 3,
```
RyanDwyer: Could we keep these two as one enum value and do the `isdigit` check in `select_workspace`? Perhaps we could eventually make `workspace_by_name` do the `isdigit` check and call `workspace_by_number` for us.

sway/tree/view.c
```diff
@@ -450,13 +450,27 @@ static struct sway_container *select_workspace(struct sway_view *view) {
...
 	list_t *criterias = criteria_for_view(view,
-			CT_ASSIGN_WORKSPACE | CT_ASSIGN_OUTPUT);
+			CT_ASSIGN_WORKSPACE | CT_ASSIGN_WORKSPACE_NUMBER | CT_ASSIGN_OUTPUT);
 	struct sway_container *ws = NULL;
 	for (int i = 0; i < criterias->length; ++i) {
 		struct criteria *criteria = criterias->items[i];
-		if (criteria->type == CT_ASSIGN_WORKSPACE) {
-			ws = workspace_by_name(criteria->target);
+		if (criteria->type == CT_ASSIGN_OUTPUT) {
+			// CT_ASSIGN_OUTPUT
```
RyanDwyer: This comment was originally added because it was in the `else` block. If this code will stay here (see my other comments) then we should remove this comment and probably add a `// CT_ASSIGN_WORKSPACE` to the `else` block.

RyanDwyer:
Also FYI, `force_display_urgency_hint` is already done, and `focus_on_window_activation` depends on changes in wlroots - specifically emitting events when an xwayland view requests focus/activation.

include/sway/criteria.h
ianyfan: Not necessarily. You can't just do the check straight off because `workspace 1` has different behaviour to `workspace number 1`.

Perhaps we could save the `number` bit and check whether the name begins with that, but then that would prevent workspaces from being named 'number 1', i.e. it would cause `workspace 'number 1'` to have the same behaviour as `workspace number 1`, which it currently does not.

Admittedly very esoteric, so this could be changed. Would require an args split during command parsing though.

sway/tree/view.c
```diff
@@ -450,13 +450,27 @@ static struct sway_container *select_workspace(struct sway_view *view) {
...
+			}
+		} else {
+			ws = criteria->type == CT_ASSIGN_WORKSPACE_NUMBER ?
+				workspace_by_number(criteria->target) :
+				workspace_by_name(criteria->target);
+
 			if (!ws) {
+				// We have to create the workspace, but if the container is
+				// sticky and the workspace is going to be created on the same
+				// output, we'll bail out first.
```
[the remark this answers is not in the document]
ianyfan: I copied it over because I'm not sure whether it does apply here.

Consider you have a container that you've assigned as thus:
`for_window [<criteria>] floating enable; sticky enable`
`assign [<criteria>] 2`

and then you open it on workspace 1, then clearly it should open on workspace 1, because it's been stickied. But currently, the code selects workspace 2 to open it on (possibly creating it), before executing the prior commands and realising that it should be stickied, and then ??? before crashing, at least, it does on my machine.

ianyfan:
Okay, should check that off the list then. Does the timeout default to 500ms? Plus is it documented? If not, that could be fixed and added in this PR.

RyanDwyer:
It does default to 500ms (see config.c:209), but the command isn't documented.

ianyfan:
Added a small fix for a crash that occurs when running `focus mode_toggle; floating toggle` on a workspace with no floating containers, as well as bringing it in line with i3 behaviour (probably, not sure if all cases have been covered)

sway/commands/focus.c
```diff
@@ -50,8 +50,11 @@ static struct cmd_results *focus_mode(struct sway_container *con,
 	} else {
 		new_focus = seat_get_focus_inactive_tiling(seat, ws);
 	}
-	if (!new_focus) {
-		new_focus = ws;
+	if (!new_focus || new_focus == ws ||
+			new_focus == ws->sway_workspace->floating) {
+		return cmd_results_new(CMD_FAILURE, "focus",
+				"Failed to find a %s container in workspace",
+				floating ? "floating" : "tiling");
```
RyanDwyer: If there are no tiling containers and I'm changing focus from floating to tiled, shouldn't `focus == ws` be a valid thing?

Also the `seat_get_focus_inactive` functions shouldn't return the `L_FLOATING` container. If that's happening, we should find out why and fix it there.

ianyfan: I don't really agree, but I've rebased it to allow it.

sway/tree/view.c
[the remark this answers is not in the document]
RyanDwyer: At this point the view won't be floating sticky yet because criteria isn't run until near the end of `view_map`. Using your example, I expect the following to happen:

* `select_workspace` returns workspace 2.
* The view is added in the tree as a tiling child of workspace 2 (but no transaction has run yet, so this doesn't appear on screen).
* The criteria runs the `floating` command, which moves the view into the L_FLOATING container of workspace 2.
* The criteria runs the `sticky` command, which should move the view to the currently visible workspace on that output (eg. workspace 1). It doesn't look like the `sticky` command is doing this.
* A transaction is created which makes the final state appear on screen.

sway/tree/workspace.c
```diff
@@ -233,19 +228,19 @@ static bool _workspace_by_number(struct sway_container *view, void *data) {
 	if (view->type != C_WORKSPACE) {
 		return false;
 	}
-	struct workspace_by_number_data *wbnd = data;
-	int a = strspn(view->name, wbnd->cset);
-	return a == wbnd->len && strncmp(view->name, wbnd->name, a) == 0;
+	char *name = data;
+	char *view_name = view->name;
+	while (isdigit(*name)) {
+		if (*name++ != *view_name++) {
```
RyanDwyer: Could this block of code be replaced with something like this?

    int num = atoi(data);
    int view_num = atoi(view->name);
    return num == view_num && num > 0;

include/sway/criteria.h
RyanDwyer: I'm happy to roll with your current implementation.

sway/tree/workspace.c
SirCmpwn: -1

SirCmpwn:
bump

sway/tree/workspace.c
RyanDwyer: Why the -1? Doesn't it achieve the same thing with less code?

SirCmpwn: Less code isn't always desirable. I think your suggestion is more complex.

sway/tree/view.c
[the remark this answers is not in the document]
ianyfan: Okay, I'll remove the comment and see if it can be fixed in the `sticky` command instead.

sway/commands/sticky.c
```diff
@@ -36,5 +36,21 @@ struct cmd_results *cmd_sticky(int argc, char **argv) {
...
+		struct sway_container *output = container_parent(container, C_OUTPUT);
+		struct sway_seat *seat = input_manager_current_seat(input_manager);
+		struct sway_container *focus = seat_get_focus_inactive(seat, output);
+		struct sway_container *focused_workspace = container_parent(focus, C_WORKSPACE);
+		struct sway_container *current_workspace = container_parent(container, C_WORKSPACE);
+		if (current_workspace != focused_workspace) {
+			container_move_to(container, focused_workspace);
+			arrange_windows(focused_workspace);
+			if (!container_reap_empty(current_workspace)) {
+				arrange_windows(current_workspace);
```
ianyfan: Have I got the logic here right? It seems to work but I still don't grok it.

RyanDwyer: Looks right to me.

RyanDwyer:
Thanks!

PR #2454, opened 2018-08-12
Use assigned workspace name for output

### LINKED ISSUE #2435: Output-bound workspace not used as default
I have a workspace configured to appear on a specific output (`workspace left output $monleft`). In i3, that output will use that workspace (or any of its other assigned workspaces, if any) on startup. Sway just assigns a numeric workspace to the output instead.

minus7:
Instead of relying on bindings being configured, primarily source
new workspace names from workspace-output assignments.

Fixes #2435

### DIFF
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -120,6 +120,8 @@ static void workspace_name_from_binding(const struct sway_binding * binding,
 		name = argsep(&cmdlist, ",;");
 	}
 
+	// TODO: support "move container to workspace" bindings as well
+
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

minus7:
Now sway will start with the lastest "registered" workspace on an output (e.g. 10 instead of 1 if you have explicit output-workspace bindings for the numbers). Adding a `break` in the loop should deal with that, though doesn't necessarily guarantee ordering as well as the explicit `order` id on bindings

PR #2458, opened 2018-08-15
Simplify transactions

RyanDwyer:
Commit 4b8e3a885be74c588291c51f798de80bd81a92db makes it so only one transaction is committed (ie. configures sent) at a time. This commit removes the now-unnecessary code which was used to support concurrent committed transactions.

* Instead of containers storing a list of instructions which they've been sent, it now stores a single instruction.
* Containers now have an `ntxnrefs` property. Previously we knew how many references there were by the length of the instruction list.
* Instructions no longer need a `ready` property. It was used to avoid marking an instruction ready twice when they were in a list, but this is now avoided because there is only one instruction and we nullify the `container->instruction` pointer when it's ready.
* When a transaction applies, we no longer need to consider releasing and resaving the surface, as we know there are no other committed transactions.
* `transaction_notify_view_ready` has been renamed to `view_notify_view_ready_by_serial` to make it consistent with `transaction_notify_view_ready_by_size`.
* Out-of-memory checks have been added when creating transactions and instructions.

BTW, we should consider doing some testing while out of memory to see if/where sway crashes and make sure the user can recover from that situation.

### DIFF
--- a/include/sway/desktop/transaction.h
+++ b/include/sway/desktop/transaction.h
@@ -20,6 +20,8 @@
  * create and commits a transaction from the dirty containers.
  */
 
+struct sway_transaction_instruction;
+
 /**
  * Find all dirty containers, create and commit a transaction containing them,
  * and unmark them as dirty.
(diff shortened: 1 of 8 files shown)


SirCmpwn:
Thanks!

PR #2459, opened 2018-08-15
Use list_find in more places and refactor/fix workspace prev_next functions

RyanDwyer:
The original purpose of this commit is to replace some `for` loops with `list_find`. But while doing this I found the `workspace_prev_next_impl` functions to be difficult to read and also contained a bug, so I refactored them and fixed the bug.

To reproduce the bug:

* Have two outputs, where the left output has workspaces 1, 2, 3 and the right output has workspaces 4, 5, 6. Make workspace 2 focused_inactive and workspace 4 focused.
* Run `workspace prev`.
* Previously it would visit the left output, then apply `workspace prev` to workspace 2, which focuses workspace 1.
* Now it will focus the rightmost workspace on the left output (workspace 3).

The refactoring I made to the workspace functions are:

* Added the `static` keyword.
* They now accept an `int dir` rather than bool, to avoid an unnecessary conversion.
* Rather than preparing `start` and `end` variables for the purpose of iterating, just iterate everything.
* Replace `for` loops with `list_find`.
* Don't call `workspace_output_prev_next_impl` (this fixes the bug).

### DIFF
--- a/common/list.c
+++ b/common/list.c
@@ -77,7 +77,7 @@ int list_seq_find(list_t *list, int compare(const void *item, const void *data),
 	return -1;
 }
 
-int list_find(list_t *list, void *item) {
+int list_find(list_t *list, const void *item) {
 	for (int i = 0; i < list->length; i++) {
 		if (list->items[i] == item) {
 			return i;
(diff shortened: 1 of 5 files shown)


emersion:
Thanks!

PR #2460, opened 2018-08-15
Implement mousedown operation

RyanDwyer:
This allows you to move the cursor off the surface while dragging its scrollbar.

To test:

* Launch something which has a scrollbar such as Firefox.
* Put something to the right of it such as a terminal.
* Click and hold Firefox's scrollbar.
* With the button still held, move the cursor off Firefox and onto the terminal.
* Observe the scrollbar still responds to mouse movements.
* If using `focus_follows_mouse`, observe the terminal does not gain focus.
* Release the mouse button. Do not move the mouse afterwards.
* Observe the cursor changes to whatever the terminal uses, such as an I-beam.
* If using `focus_follows_mouse`, observe the focus is changed to the terminal.

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -70,6 +70,7 @@ struct sway_seat {
 	double op_ref_lx, op_ref_ly;         // cursor's x/y at start of op
 	double op_ref_width, op_ref_height;  // container's size at start of op
 	double op_ref_con_lx, op_ref_con_ly; // container's x/y at start of op
+	bool op_moved;                       // if the mouse moved during a down op
 
 	uint32_t last_button;
 	uint32_t last_button_serial;
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -56,6 +56,7 @@ struct sway_seat {
 	// Operations (drag and resize)
 	enum {
 		OP_NONE,
+		OP_MOUSEDOWN,
 		OP_MOVE,
 		OP_RESIZE_FLOATING,
 		OP_RESIZE_TILING,
@@ -156,6 +157,9 @@ struct seat_config *seat_get_config(struct sway_seat *seat);
 bool seat_is_input_allowed(struct sway_seat *seat, struct wlr_surface *surface);
 
 void drag_icon_update_position(struct sway_drag_icon *icon);
+
+void seat_begin_mousedown(struct sway_seat *seat, struct sway_container *con,
+		uint32_t button, double sx, double sy);
 
 void seat_begin_move(struct sway_seat *seat, struct sway_container *con,
 		uint32_t button);
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -35,6 +35,14 @@ struct sway_drag_icon {
 	struct wl_listener destroy;
 };
 
+enum sway_seat_operation {
+	OP_NONE,
+	OP_DOWN,
+	OP_MOVE,
+	OP_RESIZE_FLOATING,
+	OP_RESIZE_TILING,
+};
+
 struct sway_seat {
 	struct wlr_seat *wlr_seat;
 	struct sway_cursor *cursor;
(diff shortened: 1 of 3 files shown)


martinetd:
> * Observe the cursor changes to whatever the terminal uses, such as an I-beam.
> * If using focus_follows_mouse, observe the focus is changed to the terminal.

FWIW i3 does not change focus to the new window when you release the button (with `focus_follow_mouse`), it will on clicking or when you go out of the window and back in again (obviously)
I find this behaviour occasionally useful so unless there's a reason to not do it let's stick to what i3 does -- could make this a suboption of focus follow mouse I guess but that'd be rather obscure and probably not well tested :p

include/sway/input/seat.h
```diff
@@ -56,6 +56,7 @@ struct sway_seat {
 	// Operations (drag and resize)
 	enum {
 		OP_NONE,
+		OP_MOUSEDOWN,
```
emersion: Possible to rename this to something else? This will be useful for touch and tablet too.

sway/input/seat.c
```diff
@@ -983,6 +994,7 @@ void seat_begin_resize_tiling(struct sway_seat *seat,
 }
 
 void seat_end_mouse_operation(struct sway_seat *seat) {
+	int operation = seat->operation;
```
emersion: This can be of the enum type (just need to name the enum).

include/sway/input/seat.h
RyanDwyer: Any suggestions? Cursordown maybe?

emersion: "down" or "cursor_down" works for me

SirCmpwn:
Works :+1:

RyanDwyer:
@SirCmpwn Can I get an executive decision on whether focus should be switched? See martinetd and DarkDefender's comments.

I consider i3's behaviour to be a bug BTW.

SirCmpwn:
I like your behavior.

SirCmpwn:
>I'd really hate to have focus automatically come under the cursor instead of "what was last focused in that workspace"

This is what already happens if you have focus_follows_mouse. If you don't, the behavior also doesn't change from today.

martinetd:
> This is what already happens if you have focus_follows_mouse. If you don't, the behavior also doesn't change from today.

Erm, I just tried again to make sure, and no? I have `focus_follows_mouse` and after changing workspace, the focus is where it was last in the workspace, not under the mouse.
This is consistent with i3 and honestly I wouldn't want it any other way - while changing focus on button up is minor and easy to adapt to, moving focus on workspace change would suck really hard.

RyanDwyer:
I think there's some confusion here. This PR does not change any behaviour with workspace switching, nor has anyone proposed it. It only changes focus after you click and drag from one view to another.

martinetd:
Yes, I just used workspace switching as an example of similar behaviour to what we had prior to this PR for button up. Anyway, I'll just shut up and stop making things more confusing than they need to be.

RyanDwyer:
GTK+ comboboxes should be good now. GTK closes the combobox if it receives a motion event between the press and release.

I considered making it skip sending the motion if the cursor's coords match the ref coords (the coords at the start of the click), but that wouldn't be a perfect solution because you could move the mouse back to the original pixel where you clicked. So I've introduced `op_moved` to track if the cursor was actually moved.

I'm starting to think we should consider making a struct for each operation and use a union in the seat struct. It would make it easier to tell which properties relate to which operations.

emersion:
>GTK closes the combobox if it receives a motion event between the press and release.

Do you think this could/should be added in wlroots? Would something that tracks the last position and doesn't send no-op move events fix the issue?

>I'm starting to think we should consider making a struct for each operation and use a union in the seat struct. It would make it easier to tell which properties relate to which operations.

Yeah, that would be a nice enhancement.

RyanDwyer:
> Do you think this could/should be added in wlroots? Would something that tracks the last position and doesn't send no-op move events fix the issue?

I don't understand. All move events are relevant.

emersion:
>I don't understand. All move events are relevant.

A move event from position 10,15 to 10,15 isn't relevant, or is it?

RyanDwyer:
That's not what's happening. The events wlroots emits are all good.

We called `cursor_send_pointer_motion` so that we can select the new focus if you've moved the mouse off the surface, but this has a side effect of sending a motion event to the surface and that made GTK close the combobox on release.

emersion:
TBH that really seems like a GTK+ bug. I don't get how a pointer motion event can close a popup.

RyanDwyer:
I think it's so you can click down with the button, highlight the option you want then release the button to select it.

emersion:
Oh right. Click down + move + up closes the popup, whereas click down + up (no move between) leaves it open.

SirCmpwn:
Thanks!

PR #2461, opened 2018-08-15
[RFC] Refactor seat_get_focus

RyanDwyer:
The story here is I want to add a `seat_get_focus_inactive_floating` function to supplement `seat_get_focus_inactive_tiling`, but I don't want to add yet another argument to `seat_get_focus_by_type` and add more logic to that function.

There's two ways this can be refactored, and I can't decide which is best. The first is a callback approach, where each of the `seat_get_focus_inactive*` functions pass a callback function to `seat_get_focus_inactive_callback`. The second ditches the callback approach and puts the loop and conditional logic directly in the `seat_get_focus_inactive*` functions. Both commits are exactly the same number of lines of code.

Please refer to both commits separately and provide feedback.

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -166,6 +166,9 @@ void seat_begin_resize_floating(struct sway_seat *seat,
 void seat_begin_resize_tiling(struct sway_seat *seat,
 		struct sway_container *con, uint32_t button, enum wlr_edges edge);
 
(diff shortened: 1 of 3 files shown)


emersion:
I have a preference for 748c22e87a231cfd888b3bb704f789be8f55d08c, but I'm not feeling strongly about it.

LGTM in any case

SirCmpwn:
I also don't care either way.

PR #2465, opened 2018-08-16
Default to the previous workspace if the current one has no focused windows

DonnieWest:
This fixes a bug that I found that can be replicated as follows:

Go to a new workspace and open any app. Do this three times so you have 3 workspaces with one window each. Go to the middle workspace, close the window and attempt to use `workspace next` or `workspace prev` to navigate to either. You won't be able to

This seems to be because `workspace_by_name` only defines `current_workspace` and `current_output` if there's something focused on the current workspace as you can see here https://github.com/swaywm/sway/blob/master/sway/tree/workspace.c#L260 and in `workspace_prev_next_impl` it checks if the passed in workspace is null and stops if it is.

Since I'm not proficient in C, I just patched it in `workspace_prev_next_impl` by defaulting to the previous workspace if it exists.

This is probably not the solution you want so feel free to close this in favor of a better solution or I'm happy to work out something better with some guidance

### DIFF
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -258,7 +258,8 @@ struct sway_container *workspace_by_name(const char *name) {
 	struct sway_container *current_workspace = NULL, *current_output = NULL;
 	struct sway_container *focus = seat_get_focus(seat);
 	if (focus) {
-		current_workspace = container_parent(focus, C_WORKSPACE);
+		current_workspace = focus->type == C_WORKSPACE ?
+			focus : container_parent(focus, C_WORKSPACE);
 		current_output = container_parent(focus, C_OUTPUT);
 	}
 


RyanDwyer:
Yeah, the root issue is in `workspace_by_name` on the line you highlighted. When the workspace is empty, `seat_get_focus` will return the workspace because the workspace itself has focus. Then it runs the first `container_parent`, but the workspace can't have a parent workspace so `container_parent` returns `NULL`. A good solution would involve changing this to check if the focus is a workspace.

DonnieWest:
Aha, there's the hole in my understanding. I think I know what I need to do now. I'll try to follow up tomorrow unless someone beats me to it

DonnieWest:
This seems to work @RyanDwyer - let me know if there's anything else I can do here

sway/tree/workspace.c
```diff
@@ -258,7 +258,7 @@ struct sway_container *workspace_by_name(const char *name) {
 	struct sway_container *current_workspace = NULL, *current_output = NULL;
 	struct sway_container *focus = seat_get_focus(seat);
 	if (focus) {
-		current_workspace = container_parent(focus, C_WORKSPACE);
+		current_workspace = focus->type == C_WORKSPACE ? focus : container_parent(focus, C_WORKSPACE);
```
RyanDwyer: We have an 80 character line limit, so this line needs to be wrapped onto two. That 80 characters assumes a tab width of 4 by the way.

DonnieWest: @RyanDwyer would wrapping on the `?` be acceptable here?

RyanDwyer: Yep, that's perfect.

DonnieWest:
@RyanDwyer fixed. Let me know if you'd like me to rebase or anything :)

ascent12:
@DonnieWest It's probably worth squashing your commits, since they're basically all just fixups.

DonnieWest:
How would you like me to handle that? Just squash and force push here or open a new PR with the squashed commits?

ascent12:
Just force push it here.

DonnieWest:
Done :)

RyanDwyer:
Thanks!

PR #2466, opened 2018-08-16
Fix geometry

### LINKED ISSUE #2178: Window geometry issues

### LINKED ISSUE #2476: Correctly handle size changes for floating views
We already handle this for xwayland, but not for the other shells.

See https://github.com/swaywm/sway/pull/2466#discussion_r210662665

RyanDwyer:
The only issue I've found with this PR is alacritty flickers while resizing. I think the content area might be a subsurface, and we don't save or render subsurfaces during transactions. From what I've read in the alacritty project their Wayland support is pretty beta.

My testing includes {floating,tiling} {gedit,alacritty,urxvt} and testing {pointer events,moving,resizing}. Urxvt is included to make sure I didn't break anything with views which use normal borders and no geometry.

Fixes #2178

Fixes #2476

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -192,10 +192,12 @@ static void render_view_toplevels(struct sway_view *view,
 		.damage = damage,
 		.alpha = alpha,
 	};
+	struct wlr_box geometry;
+	view_get_geometry(view, &geometry);
 	// Render all toplevels without descending into popups
 	output_surface_for_each_surface(output, view->surface,
-			view->swayc->current.view_x - output->wlr_output->lx,
-			view->swayc->current.view_y - output->wlr_output->ly,
+			view->swayc->current.view_x - output->wlr_output->lx - geometry.x,
+			view->swayc->current.view_y - output->wlr_output->ly - geometry.y,
 			render_surface_iterator, &data);
 }
 
@@ -232,6 +234,10 @@ static void render_saved_view(struct sway_view *view,
 		.width = view->saved_buffer_width,
 		.height = view->saved_buffer_height,
 	};
+	struct wlr_box geometry;
+	view_get_geometry(view, &geometry);
+	box.x -= geometry.x;
+	box.y -= geometry.y;
 
 	struct wlr_box output_box = {
 		.width = output->swayc->current.swayc_width,
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -229,13 +229,13 @@ static void render_saved_view(struct sway_view *view,
 		return;
 	}
 	struct wlr_box box = {
-		.x = view->swayc->current.view_x - output->swayc->current.swayc_x,
-		.y = view->swayc->current.view_y - output->swayc->current.swayc_y,
+		.x = view->swayc->current.view_x - output->swayc->current.swayc_x -
+			view->saved_geometry.x,
+		.y = view->swayc->current.view_y - output->swayc->current.swayc_y -
+			view->saved_geometry.y,
 		.width = view->saved_buffer_width,
 		.height = view->saved_buffer_height,
 	};
-	box.x -= view->saved_geometry.x;
-	box.y -= view->saved_geometry.y;
 
 	struct wlr_box output_box = {
 		.width = output->swayc->current.swayc_width,
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -234,10 +234,8 @@ static void render_saved_view(struct sway_view *view,
 		.width = view->saved_buffer_width,
 		.height = view->saved_buffer_height,
 	};
-	struct wlr_box geometry;
-	view_get_geometry(view, &geometry);
-	box.x -= geometry.x;
-	box.y -= geometry.y;
+	box.x -= view->saved_geometry.x;
+	box.y -= view->saved_geometry.y;
 
 	struct wlr_box output_box = {
 		.width = output->swayc->current.swayc_width,
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -192,12 +192,12 @@ static void render_view_toplevels(struct sway_view *view,
 		.damage = damage,
 		.alpha = alpha,
 	};
-	struct wlr_box geometry;
-	view_get_geometry(view, &geometry);
 	// Render all toplevels without descending into popups
-	output_surface_for_each_surface(output, view->surface,
-			view->swayc->current.view_x - output->wlr_output->lx - geometry.x,
-			view->swayc->current.view_y - output->wlr_output->ly - geometry.y,
+	double ox =
+		view->swayc->current.view_x - output->wlr_output->lx - view->geometry.x;
+	double oy =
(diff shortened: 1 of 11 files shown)


sway/desktop/render.c
```diff
@@ -232,6 +234,10 @@ static void render_saved_view(struct sway_view *view,
 		.width = view->saved_buffer_width,
 		.height = view->saved_buffer_height,
 	};
+	struct wlr_box geometry;
+	view_get_geometry(view, &geometry);
+	box.x -= geometry.x;
+	box.y -= geometry.y;
```
emersion: Shouldn't this use the saved geometry?

RyanDwyer: Yeah, it should. I'll amend it after I get some more feedback.

sway/desktop/xwayland.c
```diff
@@ -287,8 +287,6 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 	if (view->swayc->instruction) {
 		transaction_notify_view_ready_by_size(view,
 				surface_state->width, surface_state->height);
-	} else if (container_is_floating(view->swayc)) {
```
emersion: Hmm, is there any bad consequence of removing this?

Maybe we should have common code for all shells doing that?

sway/desktop/xdg_shell.c
```diff
@@ -231,6 +242,7 @@ static void destroy(struct sway_view *view) {
 static const struct sway_view_impl view_impl = {
 	.get_constraints = get_constraints,
 	.get_string_prop = get_string_prop,
+	.get_geometry = get_geometry,
```
emersion: Need to add the same for zxdg-shell-v6.

sway/desktop/xwayland.c
RyanDwyer: Your commit that I cherry picked removed the `view_set_{position,size}` functions. Maybe it wasn't called here at the time you made your commit.

I'm also unsure if there's any bad consequences of removing it. Maybe we could leave it out and see. If anything does need it, we can put it back and add a comment saying which program needs it.

emersion: I now remember what the issue is:
1. Open Firefox
2. Initiate a file download (e.g. go to https://www.debian.org/distrib/ and click on "64-bit PC netinst iso")
3. The "Open file/save as" dialog appears. It's initially small and is then resized to a larger size. It's broken with this patch.

sway/desktop/xwayland.c
```diff
@@ -282,13 +282,19 @@ static void handle_commit(struct wl_listener *listener, void *data) {
...
-		view_update_size(view, surface_state->width, surface_state->height);
+				state->width, state->height);
+	} else if ((state->width != view->width || state->height != view->height) &&
+				container_is_floating(view->swayc)) {
+		// eg. The Firefox "Save As" dialog when downloading a file
+		// It maps at a small size then changes afterwards.
+		view->width = state->width;
+		view->height = state->height;
+		container_set_geometry_from_floating_view(view->swayc);
+		transaction_commit_dirty();
```
emersion: Hmm. So the client updated its size, and `transaction_commit_dirty` will send a new configure event with the same size?

RyanDwyer: Not quite. `transaction_commit_dirty` sees that the pending size matches the desired size, so it'll won't send a configure. The transaction will be applied immediately.

emersion: Well `current` isn't updated, is it?

RyanDwyer: It is. When the transaction is applied it copies the pending state to `current`.

emersion: I don't get it. `should_configure` has this check:

```c
	if (con->current.view_width == instruction->state.view_width &&
			con->current.view_height == instruction->state.view_height) {
```

When `transaction_commit_dirty` is called, `current` still has the old state (the transaction isn't applied yet). So it still sends a configure event right?

RyanDwyer: Yeah, `should_configure` correctly returns false, but the instruction is already added to the transaction (added in `transaction_add_container`). An instruction is not a configure. An instruction may or may not need to send a configure. If it helps, you could think of an instruction as a container state that needs to be applied when the transaction is ready.

emersion: >Yeah, should_configure correctly returns false

No, `should_configure` returns `true`because the pending state (set here) mismatches the current state (set when the transaction is applied).

RyanDwyer: Yeah, there's something wrong there. Let me mull over this.

RyanDwyer: This is ready for review again.

sway/desktop/transaction.c
```diff
@@ -169,25 +170,17 @@ static void transaction_apply(struct sway_transaction *transaction) {
...
-		};
-		for (int j = 0; j < root_container.current.children->length; ++j) {
-			struct sway_container *output = root_container.current.children->items[j];
-			if (output->sway_output) {
-				output_damage_box(output->sway_output, &old_box);
-				output_damage_box(output->sway_output, &new_box);
-			}
+		// Damage the old location
+		desktop_damage_whole_container(container);
+		if (container->type == C_VIEW && container->sway_view->saved_buffer) {
```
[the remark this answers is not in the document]
RyanDwyer: It seems to work as expected. We damage the container (using `desktop_damage_whole_container`) then apply the new state to `current` and call it again. The damage is based off the x/y properties, and these properties are changed when gaps are applied or removed.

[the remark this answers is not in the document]
emersion: Oh right, missed that `desktop_damage_whole_container` call.

emersion:
Fairly weird bug:

1. Open Gedit
2. Open the Preferences window, it's a floating window
3. Resize it smaller than the minimum size, it jumps a little at some point
4. Enter and leave the dialog a few times, each time it's configured it jumps

RyanDwyer:
I can't reproduce that. When I make the preferences window small it stops at a certain size. I don't notice any jumping when resizing, nor when switching focus between the preferences window and the main window behind it.

emersion:
https://sr.ht/WEMm.mkv

RyanDwyer:
I still can't reproduce it, even using the default config. Can you try it with the default config if you're not already?

emersion:
Tried with the default config, same issue.

SirCmpwn:
I can reproduce this too, but it's very inconsistent. Maybe on average once in every 30 times I focus the floating window, it jumps a bit.

SirCmpwn:
Note: willing to file a follow-up ticket for this issue.

emersion:
Yeah, I'm okay with a follow-up issue as well.

SirCmpwn:
Thanks! :tada:

PR #2470, opened 2018-08-16
[WIP?] Fix zsh completions

ianyfan:
This should be fixed, but there are also a bunch of completions that are missing and could be added, so I don't know whether to do so now or closer to 1.0 release in case there are more changes.


emersion:
I don't think the CLI API will change a lot.

emersion:
Thanks!

PR #2471, opened 2018-08-17
Attempt to reap old tiling parent when floating a container

RyanDwyer:
To test:

* Create layout `H[view V[view]]`
* Focus the view inside the vertical split container and float it

The vertical split container was not reaped, so the tiling view would take up half the space.


emersion:
Thanks!

PR #2472, opened 2018-08-17
Refactor seat_get_focus functions

### LINKED ISSUE #2467: Weird bug when toggling between floating and tiled
1. Open a terminal and Gedit, focus Gedit
2. Switch Gedit mode between floating and tiled a few times
3. Switch focus between floating containers and tiled containers

The result is a floating Gedit window with half-drawn borders which we can't interact with (it seems it's neither tiled nor floating).

RyanDwyer:
Fixes #2467.

This commit introduces `seat_get_focus_inactive_floating` to supplement `seat_get_focus_inactive_tiling`, and uses it during `focus mode_toggle` which fixes a focus bug.

This also refactors the `seat_get_focus_inactive` functions so that they do their selection logic themselves rather than offloading it to `seat_get_focus_by_type` which was getting bloated. `seat_get_focus_by_type` is now removed.

Lastly, this commit changes `seat_get_focus` to just return the first container in the focus stack rather than looping and calling `seat_get_focus_by_type`.

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -166,6 +166,9 @@ void seat_begin_resize_floating(struct sway_seat *seat,
 void seat_begin_resize_tiling(struct sway_seat *seat,
 		struct sway_container *con, uint32_t button, enum wlr_edges edge);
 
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Thanks!

PR #2473, opened 2018-08-17
Implement iterators per container type

RyanDwyer:
This introduces the following `for_each` functions:

* `root_for_each_workspace`
* `root_for_each_container`
* `output_for_each_workspace`
* `output_for_each_container`
* `workspace_for_each_container`
* `container_for_each_child`

And introduces the following `find` functions:

* `root_find_output`
* `root_find_workspace`
* `root_find_container`
* `output_find_workspace`
* `output_find_container`
* `workspace_find_container`
* `container_find_child`

And removes the following functions:

* `container_descendants`
* `container_for_each_descendant`
* `container_find`

This change is preparing the way for demoting `sway_container`. Eventually these functions will accept and return `sway_output`s, `sway_workspace`s and `sway_container`s (meaning a `C_CONTAINER` or `C_VIEW`), and which point their assertions can be removed and we can use typehinting.

This change also makes it easy to handle abnormalities like the workspace floating list, root's scratchpad list and (once implemented) root's saved workspaces list for when there's no connected outputs.

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -95,4 +95,16 @@ void output_drag_icons_for_each_surface(struct sway_output *output,
 	struct wl_list *drag_icons, sway_surface_iterator_func_t iterator,
 	void *user_data);
 
+void output_for_each_workspace(struct sway_container *output,
+		void (*f)(struct sway_container *con, void *data), void *data);
+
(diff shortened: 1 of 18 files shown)


SirCmpwn:
Works fine :+1: Thanks!

PR #2474, opened 2018-08-17
Add using_csd variable to transaction state

RyanDwyer:
This fixes a race condition flicker when unfloating a view which uses client side decorations.

When the view is floated it has `using_csd = true`, so the decorations are not drawn. When unfloating it it changes to false, but this change wasn't part of transactions so it could potentially render the decorations around the view while it's waiting for the transaction to apply.


emersion:
Thanks!

PR #2478, opened 2018-08-18
Standardise debug variables

RyanDwyer:
This makes all debug options stored in a single struct rather than in various places, changes/fixes the behaviour of existing options, and introduces some new options.

* Fixes damage issues with `-Drender-tree` texture (by removing scissor)
* Offsets the render tree overlay's `y` position for those who have swaybar at the top
* Replaces `-Ddamage=rerender` with `-Dnodamage`
* Replaces `-Ddamage=highlight` with `-Dhighlight-damage`
* Replaces `-Dtxn-debug` with `-Dtxn-wait`
* Introduces `-Dnoatomic`
* Removes the `create_time` and `ms_arranging` figures from transactions and the log message. Transactions are created after arranging and the create time is of no significance.
* Fixes `-Dtxn-debug` (now `-Dtxn-wait`) not working.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -345,9 +345,7 @@ static void set_instruction_ready(
 	}
 
 	instruction->container->instruction = NULL;
-	if (!txn_debug) {
-		transaction_progress_queue();
-	}
+	transaction_progress_queue();
 }
 
 void transaction_notify_view_ready_by_serial(struct sway_view *view,
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -1,5 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include <errno.h>
+#include <limits.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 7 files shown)


emersion:
>Replaces -Ddamage=rerender with -Dnodamage

This mode still uses damage. It's more like a "binary damage mode", "all or nothing mode".

RyanDwyer:
Noted, but I don't think it should be changed. There's essentially three modes of rendering:

* render only the damaged parts (as we do normally),
* render the full output when there's any damage event (as `-Dnodamage` does), and
* render the full output on every frame regardless of damage.

If we were to benchmark how much time damage is saving us, we'd compare the first two because the third would surely never be worth doing.

RyanDwyer:
Okay, do you have any suggestions?

SirCmpwn:
I liked it the way it was before, though we could stand to parse it into an enum

sway/desktop/transaction.c
```diff
@@ -345,9 +345,7 @@ static void set_instruction_ready(
 	}
 
 	instruction->container->instruction = NULL;
-	if (!txn_debug) {
-		transaction_progress_queue();
-	}
+	transaction_progress_queue();
```
RyanDwyer: Note that `txn_debug` snuck back in after a rebase, and removing it completely is the correct behaviour. We now stall transactions (when using `-Dtxn-wait`) by inflating the `num_waiting` figure rather than avoiding calls to `transaction_progress_queue`.

SirCmpwn:
Thanks!

PR #2480, opened 2018-08-18
Fix mod+resize

### LINKED ISSUE #2479: Right-click on floating windows always resizes
1. Open Gedit
2. Float it
3. Any right click will not open the right-click menu but will start a resize operation instead

RyanDwyer:
Fixes #2479.

It was missing a `mod_pressed` check.

This also moves the parent traversal into the `if` block.


emersion:
Thanks!

PR #2481, opened 2018-08-18
Rename container_sort_workspaces and container_wrap_children

RyanDwyer:
* Renames `container_sort_workspaces` to `output_sort_workspaces` and moves it to `output.c`.
* Renames `container_wrap_children` to `workspace_wrap_children` and moves it to `workspace.c`. This function is only called with workspaces.

These functions will eventually take a `sway_output` and `sway_workspace` as their arguments, hence why they're being renamed and moved.


SirCmpwn:
Thanks!

PR #2484, opened 2018-08-18
Fix double iterating in container_for_each_child

ianyfan:


SirCmpwn:
Can you elaborate on this? It doesn't seem right to me.

ianyfan:
Almost occurrence of this functions calls `f(container, data)` followed by `container_for_each_child(container, f, data)`, which causes `f` to be called twice on the container. 

The rest of the occurrences which do not call `f`, I checked to make sure they did not require it to be called.

ianyfan:
I suppose the commit message isn't completely accurate. Also could be fixed by removing the extra calls before calling the function, but then I feel like the function name should be changed.

RyanDwyer:
Thanks!

PR #2493, opened 2018-08-19
Fix popup position when parent uses geometry

### LINKED ISSUE #2489: Right-click menus in floating windows are offset by geometry
1. Open Gedit
2. Make it float
3. Open the right-click menu
4. Position is off

RyanDwyer:
Fixes #2489.


emersion:
Thanks!

PR #2494, opened 2018-08-19
Allow subsurfaces which overflow the container to be interacted with

### LINKED ISSUE #2492: Input is broken for popups in floating views
1. Open gnome-dictionary
2. Make it float, resize it to be small
3. Open the popover menu
4. Cannot click items outside the window geometry

RyanDwyer:
Fixes #2492.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -704,10 +704,11 @@ struct sway_container *container_at(struct sway_container *workspace,
 		return NULL;
 	}
 	struct sway_container *c;
-	// Focused view's popups
 	struct sway_seat *seat = input_manager_current_seat(input_manager);
 	struct sway_container *focus =
 		seat_get_focus_inactive(seat, &root_container);
+	bool is_floating = focus && container_is_floating_or_child(focus);
+	// Focused view's popups
 	if (focus && focus->type == C_VIEW) {
 		surface_at_view(focus, lx, ly, surface, sx, sy);
 		if (*surface && surface_is_popup(*surface)) {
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Yeah, I think the rendering code needs to be changed in the same way, so it'll render in this order:

* Render unfocused views in the tiling layer
* Render focused view in the tiling layer if any (non-popups)
* Render unfocused views in the floating layer
* Render focused view in the floating layer if any (non-popups)
* Render popups

This means if the focused view's subsurfaces overflow the container, it'll be rendered on top of any siblings. And if there's an unfocused floating view on top of the subsurface, the subsurface will appear behind the floating view (I think this is desired, as subsurfaces could be non-popup-like things such as borders).

If this sounds good then I'll go ahead and implement it.

emersion:
We could also just render everything in focus order.

RyanDwyer:
> We could also just render everything in focus order.

No we can't, because an unfocused floating view must be rendered on top of a focused tiling view.

> Note: this isn't processing all popups before everything else (only the focused view's popups)

Only the focused view has popups. We close all popups when focus is changed.

RyanDwyer:
How should we handle a case where two unfocused tiling siblings both have subsurfaces which overlap the other sibling? The only way to render that perfectly is to render the subsurfaces on a separate "layer" to the main surface. Meaning we'd iterate the tree and draw the main surfaces, then iterate it again and draw the subsurfaces. And if we did that, that would make this task easier as we wouldn't have to render the focused view separately.

This would change the rendering order to be:

* Render all main surfaces and decorations in the tiling layer
* Render all subsurfaces in the tiling layer
* Render all main surfaces, subsurfaces and decorations in the floating layer (ie. just render in the z-order)
* Render popups

The `container_at` code would have to be updated accordingly.

Can you think of any reason why this wouldn't work?

emersion:
>No we can't, because an unfocused floating view must be rendered on top of a focused tiling view.

Of course, render each layer (tiling and floating) individually. The goal would just be to simplify a little all of this.

>Only the focused view has popups. We close all popups when focus is changed.

We can probably stop treating popups separately then.

>How should we handle a case where two unfocused tiling siblings both have subsurfaces which overlap the other sibling?

Subsurfaces are an integral part of the surface itself and can be placed under the main surface. You can't really order subsurfaces by layers across different surfaces.

IMHO we should just render everything in focus order (tiling then floating) and stop creating that many layers.

RyanDwyer:
> IMHO we should just render everything in focus order (tiling then floating) and stop creating that many layers.

That would be pretty wasteful on CPU. If we iterate the focus stack then we need to do a `view_is_visible` check to skip rendering views which belong to different workspaces, and `view_is_visible` involves traversing up the ancestors. We also need to traverse the ancestors to see if any parent is focused so we can use the focused colours. And to render the children of a fullscreen container would involve a `container_has_ancestor` check.

Also, how would this work when it comes to multiseat and there are multiple focus stacks?

emersion:
Adding more layers makes it even less maintainable.

emersion:
Oh, and I might add: rendering in focus order is how all floating compositors work (and yes, GNOME uses focus order when it tiles views). So I wouldn't call it exactly a hack.

SirCmpwn:
Is this ready to merge?

RyanDwyer:
It's not perfect, but it solves more problems than it fixes so I'm happy for it to be merged.

Specifically, if you create `V[terminal gedit]`, resize gedit to be stupidly small, then open the `Open` menu in gedit, it appears over the terminal. If you then focus the terminal and move the cursor over the menu, the cursor events apply to the terminal instead of the menu.

SirCmpwn:
Thanks! Make sure to file a follow-up ticket for that issue.

PR #2495, opened 2018-08-19
commands: implement move absolute

ianyfan:
This allows floating containers to be moved to the centre of all outputs.

Also changes the default behaviour of `move position <pos_x> <pos_y>` to be relative to the workspace, requiring `absolute` to be used for the old behaviour.


sway/commands/move.c
```diff
@@ -396,9 +401,17 @@ static struct cmd_results *move_to_position(struct sway_container *container,
 		container_floating_move_to(container, lx, ly);
 		return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 	} else if (strcmp(argv[0], "center") == 0) {
-		struct sway_container *ws = container_parent(container, C_WORKSPACE);
-		double lx = ws->x + (ws->width - container->width) / 2;
-		double ly = ws->y + (ws->height - container->height) / 2;
+		double lx, ly;
+		if (absolute) {
+			struct wlr_output_layout *layout = root_container.sway_root->output_layout;
+			struct wlr_box *extents = wlr_output_layout_get_box(layout, NULL);
```
emersion: You can just use `root_container.{x,y,width.height}`

sway/commands/move.c
```diff
@@ -430,6 +443,11 @@ static struct cmd_results *move_to_position(struct sway_container *container,
 				"Invalid position specified");
 	}
 
+	if (!absolute) {
+		struct sway_container *ws = container_parent(container, C_WORKSPACE);
+		lx += ws->x;
```
emersion: Hmm, this seems already done line 412, or am I missing something?

PR #2496, opened 2018-08-19
Fix crash when a view destroys with pending transactions

RyanDwyer:
We were removing the saved buffer when one transaction applies, then didn't have a new buffer to save when the next transaction ran. This made the rendering code crash as it had no surface to use.

This commit makes it continue to hold the buffer if the view is destroying and has more transactions. Additionally, a check is added when saving the buffer to make sure there's no one already there.

To test, kill a view while doing an aggressive resize motion.


SirCmpwn:
Thanks!

Issue #2491, opened 2018-08-19
Geometry not enforced for floating input events


emersion:
1. Open gnome-dictionary
2. Make it float
3. Focus tiled window
4. Move mouse slowly from the top edge of the screen to the top edge of the floating gnome-dictionary window
5. gnome-dictionary is focused before the mouse reaches the edge

PR #2498, opened 2018-08-20
Fix jumping views

### LINKED ISSUE #2490: Floating gnome-calculator moves by itself
1. Open gnome-calculator
2. Make it float
3. Switch focus, it'll move

Similar to https://github.com/swaywm/sway/pull/2466#issuecomment-414039753

RyanDwyer:
Fixes #2490.

To be honest I'm not sure why this fixes the issue.

I observed that I could only make the view jump if I resized it to the smallest possible size first. Then I had a suspicion that we were accidentally factoring in the title and border sizes into the view size when it uses CSD. So I changed that and it appears to have fixed the jumping issue.

I guess when we factor the title and borders in, we send a configure to the surface with a size smaller than the minimum, and it comes back with a surface at the minimum size. We interpret this as an unexpected resize, and this somehow makes it jump.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -236,7 +236,12 @@ void view_autoconfigure(struct sway_view *view) {
 		view->border_top = false;
 	}
 
-	switch (view->border) {
+	enum sway_container_border border = view->border;
+	if (view->using_csd) {
+		border = B_NONE;
+	}
+
+	switch (border) {
 	case B_NONE:
 		x = con->x;
 		y = con->y + y_offset;


sway/tree/view.c
```diff
@@ -236,7 +236,12 @@ void view_autoconfigure(struct sway_view *view) {
 		view->border_top = false;
 	}
 
-	switch (view->border) {
+	enum sway_container_border border = view->border;
+	if (view->using_csd) {
```
emersion: Maybe switching CSD should also change `view->border` and save the previous value in `saved_border`? (I think https://github.com/swaywm/sway/issues/2491 is due to this as well)

RyanDwyer: If we did that then we'd have to change the `border` command so it either ignores the command or writes to `view->saved_border` if CSD is in effect. But on the other hand, at the moment the rendering code has to check the `using_csd` value and omit the borders if so. I'm not sure which is best.

emersion:
Thanks!

PR #2499, opened 2018-08-20
Refactor destroy functions and save workspaces when there's no outputs

RyanDwyer:
This changes the destroy functions to the following:

* `output_begin_destroy`
* `output_destroy`
* `workspace_begin_destroy`
* `workspace_destroy`
* `container_begin_destroy`
* `container_destroy`
* `view_begin_destroy`
* `view_destroy`

The terminology was `destroy` and `free`, and it has been changed to `begin_destroy` and `destroy` respectively.

When the last output is disconnected, its workspaces will now be stashed in the root. Upon connection of a new output they will be restored.

There is a new function `workspace_consider_destroy` which decides whether the given workspace should be destroyed or not (ie. empty and not visible).

Calling `container_begin_destroy` will no longer automatically reap the parents. In some places we want to reap the parents and in some we don't, so this is left to the caller.

`container_reap_empty_recursive` and `container_reap_empty` have been combined into one function and it will recurse up the tree.

This needs testing with output hotplugging, and unplugging the last output to see what happens.

Despite adding more functions, I've somehow managed to delete more code than I've added :)

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -104,209 +104,52 @@ struct sway_container *container_create(enum sway_container_type type) {
 	return c;
 }
 
-static void container_workspace_free(struct sway_workspace *ws) {
-	list_foreach(ws->output_priority, free);
-	list_free(ws->output_priority);
-	list_free(ws->floating);
-	free(ws);
-}
-
-void container_free(struct sway_container *cont) {
-	if (!sway_assert(cont->destroying,
+void container_destroy(struct sway_container *con) {
+	if (!sway_assert(con->type == C_CONTAINER || con->type == C_VIEW,
+				"Expected a container or view")) {
+		return;
+	}
+	if (!sway_assert(con->destroying,
 				"Tried to free container which wasn't marked as destroying")) {
 		return;
 	}
-	if (!sway_assert(cont->ntxnrefs == 0, "Tried to free container "
+	if (!sway_assert(con->ntxnrefs == 0, "Tried to free container "
 				"which is still referenced by transactions")) {
 		return;
 	}
-	free(cont->name);
-	free(cont->formatted_title);
-	wlr_texture_destroy(cont->title_focused);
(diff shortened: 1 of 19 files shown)


sway/tree/container.c
```diff
@@ -104,209 +104,52 @@ struct sway_container *container_create(enum sway_container_type type) {
...
-			while (output->children->length) {
-				struct sway_container *workspace = output->children->items[0];
-
-				struct sway_container *new_output =
-					workspace_output_get_highest_available(workspace, output);
-				if (!new_output) {
-					new_output = fallback_output;
-					workspace_output_add_priority(workspace, new_output);
-				}
+	free(con->name);
```
emersion: This is duplicated with:

https://github.com/swaywm/sway/pull/2499/files#diff-101a892d1de03398ff5a39989c2332dcR162

Maybe we should consider calling this function from the other?

RyanDwyer: The plan is for the output and workspace destroy functions to eventually destroy a `sway_output` and `sway_workspace` (and likewise for their create functions). So while there's some code duplication at the moment, it's only temporary.

The purpose behind all these refactor PRs is to move things into their proper locations so that I can demote `sway_container` without it being too big of a PR.

RyanDwyer:
`begin_destroy` removes it from the pending tree and marks it as destroying. Then it gets included in the next transaction. Once that transaction has finished (ie. the thing is no longer rendered on screen), the `destroy` function is called to actually free all the resources for it. We had the same thing already, but `container_destroy` was the first function and `container_free` was the second function.

SirCmpwn:
Got it.

RyanDwyer:
I've been running this branch as my daily for the last 24 hours and haven't had any problems. I also tested hotplugging and didn't notice any issues, though I didn't hotplug down to zero outputs.

RyanDwyer:
Bump.

SirCmpwn:
Thanks!

PR #2506, opened 2018-08-22
Add app_id and class to get_tree output

vilhalmer:
This makes it much easier (possible at all?) to get ahold of the values you need to create criteria for both Wayland and X clients. Wayland clients will have a null `class`, X clients will have a null `app_id`. Tested both to make sure nothing exploded when serializing the opposite.


emersion:
Thanks!

Issue #2500, opened 2018-08-20
Hotplugging outputs have undesirable behaviours


ianyfan:
1. Adding an output that changes the origin relative to the current workspace causes floating windows and the cursor to move, presumably since their global co-ordinates are still the same, so positioned to the new origin, and therefore not actually on the workspace to which they belong, which does not change in the tree; furthermore, the floating windows become unresponsive until they are moved, at which point they do in fact become a child of the new output.

2. Removing an output and then reloading a config that has output configuration causes a crash:
```
==12725==ERROR: AddressSanitizer: heap-use-after-free on address 0x614000031c40 at pc 0x55da8c5dd8bf bp 0x7ffe5f9682c0 sp 0x7ffe5f9682b0
READ of size 8 at 0x614000031c40 thread T0
    #0 0x55da8c5dd8be in apply_output_config ../sway/config/output.c:227
    #1 0x55da8c5de627 in apply_output_config_to_outputs ../sway/config/output.c:316
    #2 0x55da8c591d8f in load_main_config ../sway/config.c:454
    #3 0x55da8c5f1260 in cmd_reload ../sway/commands/reload.c:22
    #4 0x55da8c58c6f9 in execute_command ../sway/commands.c:301
    #5 0x55da8c5e404d in seat_execute_command ../sway/commands/bind.c:323
    #6 0x55da8c5d759b in handle_keyboard_key ../sway/input/keyboard.c:281
    #7 0x7f3cf8095db7 in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #8 0x7f3cf805f3d3 in wlr_keyboard_notify_key ../subprojects/wlroots/types/wlr_keyboard.c:107
    #9 0x7f3cf8002884 in handle_keyboard_key ../subprojects/wlroots/backend/libinput/keyboard.c:73
    #10 0x7f3cf8002230 in handle_libinput_event ../subprojects/wlroots/backend/libinput/events.c:213
    #11 0x7f3cf800087e in handle_libinput_readable ../subprojects/wlroots/backend/libinput/backend.c:35
    #12 0x7f3cf658ca41 in wl_event_loop_dispatch (/usr/lib/libwayland-server.so.0+0x9a41)
    #13 0x7f3cf658b14b in wl_display_run (/usr/lib/libwayland-server.so.0+0x814b)
    #14 0x55da8c5a3d3c in server_run ../sway/server.c:164
[...]

PR #2507, opened 2018-08-22
Translate floating containers when a workspace is moved

RyanDwyer:
When a workspace is moved to another output, or the output it's on changes its global layout position, the floating containers on that workspace should be translated by the same amount as the workspace. This keeps the floating containers in the same position relative to the workspace.

A check is done to make sure the floating container's center point isn't being moved off screen (eg. when moving the workspace to a smaller output). If it is, it is centered within the workspace.

Fixes part of #2500.


emersion:
Thanks!

PR #2509, opened 2018-08-25
Pre-parse binding commands

mstoeckl:
Hi there!

This PR shifts complexity from runtime to setup, so that keyboard and mouse bindings always execute their associated command, even when allocation fails. This change has the additional benefit of validating all commands on startup.

(I would have preferred a zero-allocation command parsing method, but variable substitution and string escaping make the latter task almost impossible unless you maintain a 'large-enough' temporary buffer for string rewriting and then provide either a custom allocator or move away from null-terminated strings. I really would like to find a cleaner solution.)

As criteria are rarely used (I've yet to see an i3 config with more than a few, and don't require any myself), the corresponding code has gone mostly untested. Is there a test config file that includes everything?

FYI, `list_t` managed to be ambiguous once or twice, as I forgot if it was a pointer or plain struct.

Please review.

### DIFF
--- a/common/list.c
+++ b/common/list.c
@@ -18,6 +18,7 @@ list_t *create_list(void) {
 static void list_resize(list_t *list) {
 	if (list->length == list->capacity) {
 		list->capacity += 10;
+		// TODO: safely handle realloc failure
 		list->items = realloc(list->items, sizeof(void*) * list->capacity);
 	}
 }
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -47,7 +47,7 @@ struct cmd_results *checkarg(int argc, const char *name,
 		enum expected_args type, int val);
 
 struct cmd_handler *find_handler(char *line, struct cmd_handler *cmd_handlers,
-		int handlers_size);
+		int handlers_size, bool reading, bool active);
 /**
  * Parse and executes a command.
  */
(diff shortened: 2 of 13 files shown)


mstoeckl:
I wasn't aware that that runtime variable substitution was an explicit feature. As it is, I'll close this PR.

mstoeckl:
@RedSoxFan , runtime variable substitution (RVS), is, depending on your interpretation of what it is, broken to begin with -- was anyone using it?. You would need the following workaround to bypass double evaluation (and let us not forget unescaping, either)

    set $dsubvar $var
    set $var 2
    bindsym Ctrl+a set $dsubvar A
    bindsym Ctrl+b set $dsubvar B
    bindsym Ctrl+G exec echo $dsubvar

Since I've recently worked with the area, I could fix the double-RVS effect (and/or make it compatible with this PR), were you to specify the behaviour you'd think reasonable.

mstoeckl:
OK, the following "works", as long as $var has not been set before, but there still is the issue of double unescaping.

    bindsym Ctrl+a set $var A
    bindsym Ctrl+b set $var B
    bindsym Ctrl+G exec echo $var
    set $var 2

SirCmpwn:
You would do something like `bindsym Ctrl+G exec echo $$var`.

mstoeckl:
OK, so RVS does work (via `config.c`, line 761.), although double unescaping is still an issue.
(E.g. `bindsym Ctrl+k exec echo \\\" \\\\ \\\"`).

RedSoxFan:
> OK, so RVS does work (via config.c, line 761, the $$ approach just mentioned)

Try the following steps: https://github.com/swaywm/sway/pull/2084#issuecomment-395218464

If the config contains:
```
set $mod Mod4
set $term termite
bindsym $mod+Return exec bind $$term
```

What is happening is:
- `config_command` calls `cmd_bind` with the binding expanded to `bindsym Mod4+Return exec $term`
- `cmd_bind` calls `parse_command` which expands `exec $term` to `exec termite`
- If you change `$term` via `swaymsg set \$term any-other-terminal`, it will still have the binding stored as `exec termite`

PR #2510, opened 2018-08-25
Relocate container_move, container_move_to and container_get_in_direction

RyanDwyer:
* `container_move` is only called from the move command
* `container_move_to` was called from both the move command and the sticky command, but the sticky command can easily not call it
* `container_get_in_direction` is only called from the focus command

Moving these functions to their respective commands gives better separation of code and removes bloat from `layout.c`. These functions will need to be refactored to take advantage of type safety, so separating them will make this easier to refactor.

The following static functions have also been moved:

* `is_parellel`
* `invert_movement`
* `move_offs`
* `container_limit`
* `workspace_rejigger`
* `move_out_of_tabs_stacks`
* `get_swayc_in_output_direction`

They were all used by the move functions, except for the last one which is used by focus.

Other changes:

* `index_child` has been renamed to `container_sibling_index`, moved to `container.c` and made public
* `sway_output_from_wlr` has been renamed to `output_from_wlr_output`, moved to `output.c` and made public
* `container_handle_fullscreen_reparent` has been made public
* `sway_dir_to_wlr` has been made public

No changes have been made to any of the moved functions, other than updating calls to functions that have been renamed.


SirCmpwn:
Thanks!

Issue #307, opened 2015-12-13
Airblader fork features


SirCmpwn:
This issue tracks our progress implementing the features of @Airblader's i3 fork: https://github.com/Airblader/i3

Commands:
- [x] gaps
- [x] workspace <ws> gaps [inner|outer] <px>
- [x] smart_gaps
- [ ] smart_borders
- [x] bar { border }
- [x] bar { height }
- [x] rgba colors

mikkeloscar:
Do you know what a block is in the context of a bar? I'm trying to figure out how `bar { border }` should be implemented in the config file and the [airblader/i3 docs](https://github.com/Airblader/i3#background--color) mention **individual blocks**, but I'm not sure what those blocks are?

Any idea what `bar { border }` should look like in the config?

SirCmpwn:
That's not a config file option, it's part of the JSON you can send via stdout from status_command to have formatted colors and such.

Airblader:
In your swaybar screenshot from today it seems to me that the gaps are actually implemented incorrectly. The spacing between two windows is double the spacing on the screen edges. This was a bug in the old original gaps patch which I've fixed quite a while ago.

SirCmpwn:
I have the gaps configured like so in that screenshot:

    gaps outer 0
    gaps inner 10

Incorrect?

Airblader:
Your settings are fine, but the gaps implementation is wrong. Using the same gaps settings, see the following screenshot. The gap between the two windows is just as big as the gap between a window and the screen edge, as it should be. In sway this seems to be incorrect because the gap between two windows is double the size of the gap between window and screen edge.

This is (at least in i3) a problem because if you just inset a window, you will naturally run into this problem. What you need to do is divide the gaps by two for edges between two windows.

![screenshot-2015-12-17_07-07-21](https://cloud.githubusercontent.com/assets/2392216/11869324/fd94a85c-a48c-11e5-9349-60220f34a777.png)

SirCmpwn:
I see. The gaps right now are just implemented as margins. I'll see about fixing this, thanks!

Airblader:
Sure, no problem. Does sway support fullscreening nested containers? You might want to check gap behavior there, too, as in i3-gaps I had to do some fixes for it.

SirCmpwn:
As in fullscreening the actual container, instead of views? No, that's not supported, but would probably be fairly easy to do.

Airblader:
[...]

PR #2511, opened 2018-08-25
Prepare arrange code for type safe arguments

RyanDwyer:
This commit changes the arrange code in a way that will support type safe arguments.

The `arrange_output` et al functions are now public, however I opted not to use them directly yet. I've kept the generic `arrange_windows` there for convenience until type safety is fully implemented. This means this patch has much less risk of breaking things as it would otherwise.

To be type safe, `arrange_children_of` cannot exist in its previous form because the thing passed to it could be either a workspace or a container. So it's now renamed to `arrange_children` and accepts a `list_t`, as well as the parent layout and parent's box.

There was some code which checked the grandparent's layout to see if it was tabbed or stacked and adjusted the Y offset of the grandchild accordingly. Accessing the grandparent layout isn't easy when using type safe arguments, and it seemed odd to even need to do this. I determined that this was needed because a child of a tabbed container would have a swayc Y matching the top of the tab bar. I've changed this so a child of a tabbed container will have a swayc Y matching the bottom of the tab bar, which means we don't need to access the grandparent layout. Some tweaks to the rendering and autoconfigure code have been made to implement this, and the `container_at` code appears to work without needing any changes.

`arrange_children_of` (now `arrange_children`) would check if the parent had gaps and would copy them to the child, effectively making the workspace's gaps recurse into all children. We can't do this any more without passing `has_gaps`, `gaps_inner` and `gaps_outer` as arguments to `arrange_children`, so I've changed the `add_gaps` function to retrieve it from the workspace directly.

`apply_tabbed_or_stacked_layout` has been split into two functions, as it had different logic depending on the layout.

Lastly, `arrange.h` had an unnecessary include of `transaction.h`. I've removed it, which means I've had to add it to several other files.

### DIFF
--- a/include/sway/tree/arrange.h
+++ b/include/sway/tree/arrange.h
@@ -1,12 +1,16 @@
 #ifndef _SWAY_ARRANGE_H
 #define _SWAY_ARRANGE_H
-#include "sway/desktop/transaction.h"
 
 struct sway_container;
 
-/**
- * Arrange layout for all the children of the given container.
- */
+void arrange_container(struct sway_container *container);
+
+void arrange_workspace(struct sway_container *workspace);
+
+void arrange_output(struct sway_container *output);
+
+void arrange_root(void);
+
 void arrange_windows(struct sway_container *container);
 
 #endif
--- a/sway/commands/layout.c
+++ b/sway/commands/layout.c
@@ -103,7 +103,7 @@ struct cmd_results *cmd_layout(int argc, char **argv) {
 			parent->prev_split_layout = prev;
 		}
 		container_notify_subtree_changed(parent);
-		arrange_windows(parent);
+		arrange_windows(parent->parent);
 	}
 
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -16,6 +16,7 @@
 #include "log.h"
 #include "config.h"
 #include "sway/config.h"
+#include "sway/desktop/transaction.h"
 #include "sway/input/input-manager.h"
 #include "sway/input/seat.h"
 #include "sway/layers.h"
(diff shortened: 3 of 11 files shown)


RedSoxFan:
> `arrange_children_of` (now `arrange_children`) would check if the parent had gaps and would copy them to the child, effectively making the workspace's gaps recurse into all children. We can't do this any more without passing `has_gaps`, `gaps_inner` and `gaps_outer` as arguments to `arrange_children`, so I've changed the `add_gaps` function to retrieve it from the workspace directly.

I don't personally use gaps so I could be misunderstanding something, but this doesn't sound right to me. The gaps appear to only be copied from the parent to the child *if* the child does *not* have gaps and the parent *does*. Since it looks like it is possible to override gaps on a per container/view basis, wouldn't it be possible that the gaps at the workspace-level are not the correct gaps to apply?

RyanDwyer:
Hmm. That's true. Looking at the [i3-gaps readme](https://github.com/Airblader/i3), it looks like `current` is supposed to apply them to the workspace anyway, meaning you shouldn't be able to apply them per-container. This looks like a bug to me, unless sway has intentionally deviated from upstream behaviour in a backwards-incompatible way.

RedSoxFan:
> Looking at the i3-gaps readme, it looks like `current` is supposed to apply them to the workspace anyway, meaning you shouldn't be able to apply them per-container. This looks like a bug to me, unless sway has intentionally deviated from upstream behaviour in a backwards-incompatible way.

Based on the changes to [sway/commands/gaps.c](https://github.com/swaywm/sway/pull/2131/files#diff-a499025a5adff61b81605c7d4830e6a6) in the open PR #2131, it looks like you are correct and this is actually a bug and not intentional.

I'm fine with removing the ability to change gaps at the container/view level for now. I think that if it is re-implemented in the future, it should be done as an extension (rather than a deviation) and use an unused scope (ex `container`)

SirCmpwn:
We once had @airblader explain how our gaps were supposed to be and confirmed that we got it right at some point.

https://github.com/swaywm/sway/issues/307

RyanDwyer:
Found a bug in this - please hold off on merging.

RyanDwyer:
All good now.

The bug was a rendering issue when using tabbed containers with inner gaps. The tab bar was misaligned.

SirCmpwn:
Thanks!

PR #2512, opened 2018-08-25
Workspace move cycle

apreiml:
On move workspace to direction: Try the farthest on the opposite direction if no workspace is found at given direction. 

This new behavior is closer to the i3 one now. I'd prefer it, because I'm using most of the time only one display connected to my notebook. With this addition I only need one shortcut to toggle a workspace between the two displays. I couldn't test this with more displays.

Depends on https://github.com/swaywm/wlroots/pull/1199

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -26,7 +26,20 @@ static const char *expected_syntax =
 	"'move <container|window|workspace> [to] output <name|direction>' or "
 	"'move <container|window> [to] mark <mark>'";
 
-static struct sway_container *output_in_direction(const char *direction,
+enum wlr_direction opposite_direction(enum wlr_direction d) {
+	switch (d) {
+		case WLR_DIRECTION_UP:
+			return WLR_DIRECTION_DOWN;
+		case WLR_DIRECTION_DOWN:
+			return WLR_DIRECTION_UP;
+		case WLR_DIRECTION_RIGHT:
+			return WLR_DIRECTION_LEFT;
+		default:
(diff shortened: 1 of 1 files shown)


sway/commands/move.c
```diff
@@ -37,19 +50,33 @@ static struct sway_container *output_in_direction(const char *direction,
 		{ "left", WLR_DIRECTION_LEFT },
 		{ "right", WLR_DIRECTION_RIGHT },
 	};
+
+	enum wlr_direction *direction = NULL;
```
RyanDwyer: Nitpick: This doesn't need to be a pointer.

sway/commands/move.c
```diff
@@ -37,19 +50,33 @@ static struct sway_container *output_in_direction(const char *direction,
...
+
 	for (size_t i = 0; i < sizeof(names) / sizeof(names[0]); ++i) {
-		if (strcasecmp(names[i].name, direction) == 0) {
-			struct wlr_output *adjacent = wlr_output_layout_adjacent_output(
+		if (strcasecmp(names[i].name, direction_string) == 0) {
+			direction = &names[i].direction;
```
RyanDwyer: Another nitpick: You could add a `break` here to stop iterating the loop once you've found the direction.

apreiml:
@RyanDwyer I changed the code according your suggestions.

RedSoxFan:
Can you please rebase and force push? swaywm/wlroots#1199 has been merged so it should be good to merge after.

apreiml:
@RedSoxFan done

PR #2513, opened 2018-08-25
Rename sway_root.outputs to sway_root.all_outputs

RyanDwyer:
This list includes disabled outputs.

When `sway_container` is demoted, we'll need to store the root's children (ie. enabled outputs) in the `sway_root`. It makes sense to put these in a list called `outputs`, so I'm renaming the existing list in advance.

### DIFF
--- a/include/sway/tree/root.h
+++ b/include/sway/tree/root.h
@@ -21,7 +21,7 @@ struct sway_root {
 
 	struct wlr_texture *debug_tree;
 
-	struct wl_list outputs; // sway_output::link
+	struct wl_list all_outputs; // sway_output::link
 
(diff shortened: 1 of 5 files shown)


include/sway/tree/root.h
```diff
@@ -21,7 +21,7 @@ struct sway_root {
 
 	struct wlr_texture *debug_tree;
 
-	struct wl_list outputs; // sway_output::link
+	struct wl_list all_outputs; // sway_output::link
```
SirCmpwn: Can you add a comment here clarifying what kind of outputs are included here? Then you don't need to change the variable name at all.

SirCmpwn:
Thanks!

PR #2514, opened 2018-08-26
Simplify runtime variable substitution

mstoeckl:
I've looked at #2084, and think that the difficulties in the current solution for runtime variable substitution (RVS) lie in that commands evaluated at runtime via `bindsym`, `bindcode`, and `for_window`, are preprocessed twice. By ensuring that the nested commands are kept unprocessed until runtime, RVS follows more intuitively.

Under this change, a config file might include this

    set $term program
    exec $term
    set $term other-program

When the config file is read, `exec $term`  is preprocessed to `exec program` and is run. When `$mod+Return`, the asssociated command  `exec $term` is preprocessed to be `exec other-program`, 
and run.

This change rolls back the new syntax ($$var) introduced by #2084, but as that PR is recent and not yet in any stable releases, it shouldn't break any config files for people who should not already have been expecting it :-). It also ensures that the same level of string escaping is required for top-level commands
invocations as for nested command invocations.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -51,7 +51,7 @@ struct cmd_handler *find_handler(char *line, struct cmd_handler *cmd_handlers,
 /**
  * Parse and executes a command.
  */
-struct cmd_results *execute_command(char *command,  struct sway_seat *seat);
+struct cmd_results *execute_command(const char *command, struct sway_seat *seat);
 /**
  * Parse and handles a command during config file loading.
  *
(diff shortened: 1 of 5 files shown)


SirCmpwn:
Thank you for putting in the effort, but I'm not really fond of this change. It was working as intended before.

mstoeckl:
> Thank you for putting in the effort, but I'm not really fond of this change.

I have more experience with i3, and their design opinions re simplicity[0]. May I ask about your general configuration philosophy, and recommend that it be made more prominent?

> It was working as intended before.

This change does ensure that nested commands only require one level of escaping, just like top-level 
commands. The existing behavior, in which double-escaping is required, is counterintuitive.

[0] provided at https://i3wm.org/index.html

SirCmpwn:
I find the current design *more* simple. I don't intend to argue the point, I don't have the time.

PR #2515, opened 2018-08-26
Remove layout.c

RyanDwyer:
When we have type safety we'll need to have functions for `workspace_add_tiling` and so on. This means the existing container functions will be just for containers, so they are being moved to `container.c`. At this point `layout.c` doesn't contain much else, so I've relocated everything and removed the file.

* `container_swap` and its static functions have been moved to the swap command and made static.
* `container_recursive_resize` has been moved to the resize command and made static.
* The following have been moved to `container.c`:
    * `container_handle_fullscreen_reparent`
    * `container_insert_child`
    * `container_add_sibling`
    * `container_add_child`
    * `container_remove_child`
    * `container_replace_child`
    * `container_split`
* `enum movement_direction` and `sway_dir_to_wlr` have been moved to `util.c`.

No changes have been made to any of the functions.

Side note: Several commands included `layout.h` which then included `root.h`. With `layout.h` gone, `root.h` has to be included by those commands.


sway/commands/show_marks.c
```diff
@@ -2,6 +2,8 @@
 #include <string.h>
 #include "sway/commands.h"
 #include "sway/config.h"
+#include "sway/tree/root.h"
+#include "sway/tree/root.h"
```
RedSoxFan: `sway/tree/root.h` only needs to be added once

RyanDwyer: Derp. Fixed by replacing the commit.

SirCmpwn:
Thanks!

PR #2517, opened 2018-08-26
Add fish completion

fokditkak:
This is my first ever pull request and I am very excited. 
Everything compiled successfuly.

### DIFF
--- a/completions/fish/sway.fish
+++ b/completions/fish/sway.fish
@@ -1,0 +1,10 @@
+# sway(1) completion
+
+complete -c sway -s h -l help --description "Show help message and quit."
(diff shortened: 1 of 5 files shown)


SirCmpwn:
Thanks! Can you add the email you used to author these commits to github so they show up linked to your GitHub account?

fokditkak:
There is a slight problem - the default path to the fish completions file is different from what I've put in the meson.build file. I've put absolute path but I'm not sure whether that's ok or not. Also, I haven't pushed this last change.

SirCmpwn:
Thanks!

PR #2518, opened 2018-08-26
Emit window close event earlier, before destroying

ianyfan:
A small fix to an annoying problem I had with the ipc, whereby when I closed a window, it would send out a focus event on the newly-focused container before sending out the close event for the old container. Unless I'm misunderstanding the problem, this should fix that.

Also removed the TODO on container_create_notify, since I think all cases have been covered.


RedSoxFan:
Thanks!

PR #2520, opened 2018-08-27
Destroy empty workspaces in output_evacuate

RedSoxFan:
#2499 accidentally removed the check for empty workspaces when moving workspaces on output disconnect/disable. This caused empty workspaces to also be moved. This PR adds the check back in.


SirCmpwn:
"evacuate" is such a good function name

PR #2523, opened 2018-08-27
Stop floaters from snapping on move to new output

RedSoxFan:
When moving a floating view/container to a new output using the mouse, the floating view/container would snap back to the center of the original output as soon as the center point crossed over.

This PR skips the `container_floating_move_to_center` call in `arrange_workspace` when the container is being operated on by the current seat. This makes it so the floater does not snap to the center of any workspace while being moved.

### DIFF
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -216,16 +216,18 @@ static void arrange_workspace(struct sway_container *workspace) {
 	// Adjust any floating containers
 	double diff_x = workspace->x - prev_x;
 	double diff_y = workspace->y - prev_y;
-	for (int i = 0; i < workspace->sway_workspace->floating->length; ++i) {
-		struct sway_container *floater =
-			workspace->sway_workspace->floating->items[i];
-		container_floating_translate(floater, diff_x, diff_y);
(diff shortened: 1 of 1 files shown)


RyanDwyer:
An alternative solution would be to check if `diff_x` and `diff_y` are zero and skip the entire `for` loop. This would be more performant too.

RedSoxFan:
> An alternative solution would be to check if diff_x and diff_y are zero and skip the entire for loop. This would be more performant too.

I think that idea better. I've replaced the commit and force pushed.

SirCmpwn:
>When moving a floating view/container to a new output using the mouse, the floating view/container would snap back to the center of the original output as soon as the center point crossed over.

I don't understand and/or cannot reproduce this.

RedSoxFan:
> I don't understand and/or cannot reproduce this.

I'm able to replicate with the default config. Try using `floating_mod` and dragging a floating view/container to a different output slowly and keep holding the buttons for a couple seconds afterwords. If you do it quickly, the arrange call may not happen before the move is over.

SirCmpwn:
Thanks, I was able to reproduce it.

PR #2526, opened 2018-08-27
Fix moving container to inactive workspace on different output

ianyfan:
When moving a container to an inactive workspace on a different output, this will change the focus on the destination output back to its last active workspace.

Also includes a few code cleanups.

~Edit: need to add a check that `new_output_last_ws != destination_ws`~


RedSoxFan:
Thanks!

Issue #2334, opened 2018-07-23
Verify sway's IPC matches i3's IPC


SirCmpwn:
I know at least get_tree isn't right.

ianyfan:
(Possibly non-exhaustive) list of properties missing in `get_tree`:

- border
- current_border_width
- deco_rect
- floating
- floating_nodes
- fullscreen_mode
- geometry
- last_split_layout
- orientation
- output
- percent
- scratchpad_state
- stick
- swallows
- window
- window_properties
- window_rect
- workspace_layout

All the other commands seem to match i3.

SirCmpwn:
Anyone interested in finishing this one up?

ianyfan:
I think this might actually be finished, and that the remaining properties are, as discussed in #2530, undocumented and therefore should not be added.

SirCmpwn:
Sweet!

PR #2530, opened 2018-08-28
WIP: start adding missing ipc properties

apreiml:
I started to work on the get_tree command to complete #2334


apreiml:
`orientation` is obsolete. Should I skip this?

SirCmpwn:
No, we should leave it in so it doesn't break anything old which is unmarshalling that JSON.

apreiml:
I tried to implement the window and deco rects. Can you check If I understood it correctly? I still need to figure out how the gaps effect those values.

SirCmpwn:
I'm not sure how it works on i3, you'd know that better than I

sway/ipc-json.c
```diff
@@ -46,20 +46,28 @@ json_object *ipc_json_get_version() {
...
-	json_object_object_add(rect, "height", json_object_new_int((int32_t)c->height));
+	json_object_object_add(rect, "x", json_object_new_int(x));
+	json_object_object_add(rect, "y", json_object_new_int(y));
+	json_object_object_add(rect, "width", json_object_new_int(width));
+	json_object_object_add(rect, "height", json_object_new_int(height));
 
 	return rect;
 }
 
+static json_object *ipc_json_create_empty_rect() {
```
ianyfan: `void`

SirCmpwn:
How's this coming along?

apreiml:
I stopped because of https://github.com/swaywm/sway/pull/2540 I thought I'll wait until this will be merged, but I branched of from ryans branch yesterday. I will continue there and rebase later. I will try to finish it this week.

SirCmpwn:
Cool, thanks.

apreiml:
I've added all the missing fields that where documented. There some notable differences/changes though:

* i3 requires every key present in the json object, sway get_tree will do this also now for the documented i3 fields.
* _window_ will only be set on x11 view, otherwise it's `NULL`
* The defaults for (_border_, _current_border_width_) in sway are (`"none"`, `0`) instead of the i3 values (`"normal"`, `-1`)
* The bar in i3 is part of the tree, but not in sway.

I would like this pr to include only those I did for now. The others may be a little bit more effort and I'm also not sure if the plan of sway is to be also compatible to those. The i3 docs state, that they are not finalized and may change. 

From my point of view this PR is ready for a review.

sway/ipc-json.c
```diff
@@ -30,6 +30,20 @@ static const char *ipc_json_layout_description(enum sway_container_layout l) {
 	return "none";
 }
 
+static const char *ipc_json_orientation_description(enum sway_container_layout l) {
+	switch (l) {
```
ianyfan: This case statement is a bit too heavy for my liking, since it's just a return statement and there are many fall-throughs. Personally, I'd make it a ternary, though I think even an if block would be better.

sway/ipc-json.c
```diff
@@ -123,7 +144,19 @@ static void ipc_json_describe_output(struct sway_output *output,
...
+
+	struct sway_node *parent = node_get_parent(&output->node);
+	struct wlr_box parent_box = {0, 0, 0, 0};
+
+	if (parent != NULL) {
+		node_get_box(parent, &parent_box);
+	}
+
+	if (parent_box.width != 0 && parent_box.height != 0) {
+		double percent = ((double)output->width / (double)parent_box.width)
```
ianyfan: Only one of the operands needs to be casted to double. Also, there seems to a rogue trailing space here which you should definitely get rid of.

Also, does the percent property get added when the parent box is 0? If not, maybe it should have a default value.

sway/ipc-json.c
```diff
@@ -172,35 +210,101 @@ static void ipc_json_describe_workspace(struct sway_workspace *workspace,
...
+	struct wlr_box deco_box = {0, 0, 0, 0};
+
+	if (c->current.border == B_NORMAL) {
+		deco_box.width = c->width;
+		deco_box.height = c->view->y - c->y;
 	}
 
+	json_object_object_add(object, "deco_rect", ipc_json_create_rect(&deco_box));
+
+	struct wlr_box geometry = {0, 0, c->view->natural_width, c->view->natural_height };
```
ianyfan: Remove the space before the closing brace

sway/ipc-json.c
```diff
@@ -172,35 +210,101 @@ static void ipc_json_describe_workspace(struct sway_workspace *workspace,
...
+
+	struct sway_node *parent = node_get_parent(&c->node);
+	struct wlr_box parent_box = {0, 0, 0, 0};
+
+	if (parent != NULL) {
+		node_get_box(parent, &parent_box);
+	}
+
+	if (parent_box.width != 0 && parent_box.height != 0) {
+		double percent = ((double)c->width / (double)parent_box.width)
```
ianyfan: Same as above.

sway/ipc-json.c
ianyfan: Oh wait, I missed this question. tbh I'm not sure either, because I don't really use it. Maybe we should just leave it for now and see if someone with a better use case for it has an idea.

Issue #2527, opened 2018-08-28
Crash related to opening & closing a window too quickly


ianyfan:
New problem: https://github.com/swaywm/sway/issues/2527#issuecomment-417965615

### This problem appears to be fixed

~Steps to reproduce (using termite as my terminal):~
* ~empty config except `exec termite` to open the terminal~
* ~run `swaymsg reload &; termite -e sleep 0.05 &; termite` until it crashes (most of the times it happens first go but sometimes doesn't)~

Crash log:
```
=================================================================
==18603==ERROR: AddressSanitizer: heap-use-after-free on address 0x614000038908 at pc 0x556d6aec6052 bp 0x7fff958e8a10 sp 0x7fff958e8a00
READ of size 4 at 0x614000038908 thread T0
    #0 0x556d6aec6051 in render_container_simple ../sway/desktop/render.c:567
    #1 0x556d6aec77f5 in render_container ../sway/desktop/render.c:749
    #2 0x556d6aec8cff in output_render ../sway/desktop/render.c:891
    #3 0x556d6aebd3aa in damage_handle_frame ../sway/desktop/output.c:383
    #4 0x7ff706b95657 in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #5 0x7ff706b68d53 in output_handle_frame ../subprojects/wlroots/types/wlr_output_damage.c:50
    #6 0x7ff706b95657 in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #7 0x7ff706b71a52 in wlr_output_send_frame ../subprojects/wlroots/types/wlr_output.c:533
    #8 0x7ff706afcaa5 in page_flip_handler ../subprojects/wlroots/backend/drm/drm.c:1031
    #9 0x7ff704b8f4d4 in drmHandleEvent (/usr/lib/libdrm.so.2+0x54d4)
    #10 0x7ff706afcc54 in handle_drm_event ../subprojects/wlroots/backend/drm/drm.c:1041
    #11 0x7ff705889701 in wl_event_loop_dispatch (/usr/lib/libwayland-server.so.0+0xa701)
    #12 0x7ff7058882ab in wl_display_run (/usr/lib/libwayland-server.so.0+0x92ab)
    #13 0x556d6aeb4e5c in server_run ../sway/server.c:164
    #14 0x556d6aeb3ad5 in main ../sway/main.c:445
[...]

PR #2531, opened 2018-08-28
Don't use bitfield to test for similar transactions

RyanDwyer:
When there's multiple transactions in the queue, sway can take a shortcut by checking if they all operate on the same set of containers. If they do, it can skip all but the last transaction. The way we tested for transactions which used the same containers was to exclusive-or their con IDs together, but this has proved not only to be ineffective but also has the potential to make sway crash.

This patch replaces the exclusive-or with a loop and container comparison.

This has a good chance of fixing #2527, but I've not been able to reproduce it in a nested session and can't do DRM testing right now. @ianyfan Can you please test? Otherwise I'll do DRM testing later.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -22,7 +22,6 @@ struct sway_transaction {
 	list_t *instructions;   // struct sway_transaction_instruction *
 	size_t num_waiting;
 	size_t num_configures;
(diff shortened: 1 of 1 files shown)


ianyfan:
Hasn't crashed in my testing so LGTM

SirCmpwn:
Thanks! Whether it fixes the crash or not, XORing the container IDs isn't a good design.

PR #2533, opened 2018-08-29
Fix workspace tabs

### LINKED ISSUE #2532: Broken background behind floating windows
Steps to reproduce:
1. Open gnome-terminal
2. Switch to stacking or tabbed layout
3. Open about window
result: 
2018-08-29-010355_grim
expected result: 
2018-08-29-010405_grim
issue present since #2466

RyanDwyer:
When collecting focus to save into the transaction state, the workspace needs to look in the tiling list only.

As `seat_get_focus_inactive_tiling` returns any descendant, the list also needs to be traversed back up to the direct child of the workspace.

Fixes #2532

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -122,7 +122,17 @@ static void copy_pending_state(struct sway_container *container,
 	struct sway_seat *seat = input_manager_current_seat(input_manager);
 	state->focused = seat_get_focus(seat) == container;
 
-	if (container->type != C_VIEW) {
+	if (container->type == C_WORKSPACE) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

RedSoxFan:
This PR causes rendering issue for children of tabbed and stacked containers.

For the representation `T[view#1 V[view#2 view#3]]` only `view#2` or `view#3` will be rendered depending on which is focused. The same thing happens when using stacked instead of tabbed and/or horizontal instead of vertical. Also, as the nesting gets deeper, less and less is visible since only the focused view is rendered.

PR #2535, opened 2018-08-29
Fix crash when reaping a workspace

RyanDwyer:
It wasn't returning the surviving container.

To reproduce, open a terminal in a workspace with no siblings, run `sleep 2 && exit` then switch to another workspace during the sleep.


SirCmpwn:
Thanks!

PR #2536, opened 2018-08-29
Fix rendering of nested tabs

RyanDwyer:
Fixes this comment: https://github.com/swaywm/sway/pull/2533#issuecomment-416820583

The `C_CONTAINER` check needs to apply to `C_VIEW` as well.


RedSoxFan:
Thanks!

PR #2537, opened 2018-08-29
commands: fix exec quoting

ianyfan:
The way I implemented it before was really a stop-gap measure, furthermore, there was a bug since it was comparing `argc > 0` rather than `argc > 1` so even fully-quoted commands were rejected.

Now, it allows partially-quoted commands, but does not strip quotes and split the command, only if it is fully quoted.


SirCmpwn:
Thanks!

PR #2540, opened 2018-08-30
Implement type safe arguments and demote sway_container

### LINKED ISSUE #2343: Proposal: Tree code restructure
I'm making this issue to discuss some ideas for the sway codebase. I admit the ideas I'm proposing here are somewhat ambitious. I expect this would be something we'd look at after 1.0 is released, maybe even as a sway 2.0 kind of thing, if we did it at all.

My proposal aims to improve the tree-related code, more specifically in the area of containers:

* The term container refers to both any node in the tree, as well as a node of type `C_CONTAINER`.
* There is no type-specific struct for `C_CONTAINER`s. Instead, we shove everything into `sway_container` and the other types ignore the properties which don't apply to them.
* Disabled outputs don't have a swayc, unmapped views don't have a swayc, but sway_workspaces always do. There is inconsistency here, and there's no real reason for the swayc to be optional.
* `container.c` and `layout.c` are collectively over 2000 lines of code, and it's not clear what should go in which file.
* There's little strategy or consistency to where files are placed - the tree directory contains a `layout.c` and `arrange.c`, and we have two `output.c` files in different folders, with header files that don't match.
* Containers try to be generic, but then end up having to do type-specific logic.

That last point is a big one. Containers *must* be generic for the focus stack and transactions, but in many places the container-first approach doesn't seem to be providing much of a benefit. It has issues like lack of typehinting and having to assert the type of container passed to functions. In most cases we already know what type of container we're dealing with when we call a generic container function, and the generic container functions tend to check the container type and have different code paths anyway, hence why I think the container-first approach isn't working well.

On a related note, `C_CONTAINER`s and `C_VIEW`s are actually very similar. In i3 they are the same type. In sway they are separate, which has lead to things like views being able to be fullscreened/floated/marked but not containers. Considering all this functionality should apply to containers as well, there is little reason to keep these types separated.

I think we should consider combining `C_CONTAINER`s and `C_VIEW`s into one type, and consider having a more conventional approach where outputs, workspaces and containers are our primary data types. Each of these could have a node struct which is basically a generic handle. We would prefer type-specific functions and structs rather than generic ones. I think the only places where the node struct would be used is in the focus stack and transactions.

Here's an example of what I mean:

    struct sway_node {
        enum node_type type; // root, output, workspace or con
        union {
            struct sway_root *root;
            struct sway_output *output;
            struct sway_workspace *workspace;
            struct sway_con *con;
        };
        // Nothing else in this struct
    };

    struct sway_root {
        struct sway_node node;
        list_t *outputs;          // struct sway_output
        list_t *saved_workspaces; // struct sway_workspace, for when there are no outputs connected
        list_t *scratchpad;       // struct sway_con
        ...
    };

    struct sway_output {
        struct sway_node node;
        list_t *workspaces;       // struct sway_workspace
        char *name;
        ...
    };

    struct sway_workspace {
        struct sway_node node;
        struct sway_output *output;
        list_t *tiling;             // struct sway_con
        list_t *floating;           // struct sway_con
        char *name;
        ...
    };

    struct sway_con {
        struct sway_node node;
        struct sway_workspace *workspace; // Always set for all containers
        struct sway_con *parent;          // NULL if in the root of the workspace
        list_t *children;                 // struct sway_con

        struct sway_view *view;           // NULL for actual containers
        char *title;
        ...
    };

    struct sway_view {
        struct sway_con *con;
        ...
    };

Some benefits of this structure are:

* No double meaning of the word container.
* More conventional approach makes it easier to work with.
* Because the node struct is not a pointer and is embedded within each struct, we wouldn't have to manually create or track these like we currently do for `output->swayc` and `view->swayc`.
* We wouldn't have piles of generic code in `container.c` and `layout.c`.
* Typehinting/no container-type assertions.
* We could use the more appropriate `title` property for containers/views rather than reusing the `name` property.
* We could easily find a container's workspace or output by using `con->workspace` and `con->workspace->output`.
* Functions in general would be easier to read as they wouldn't contain as much branching logic.
* Possibly less lines of code overall, due to less branching and type assertions.

Some downsides of this structure are:

* We may have similar functions like `root_for_each_container`, `output_for_each_container`, `workspace_for_each_container` and `container_for_each_container`, where each function calls the next one in the chain. This means there's potential for there to be more lines of code overall, but these functions would be simple and easy to understand.
* Changing stuff introduces risk of breakage.

I don't consider the number of lines of code to be an important measurement, but I think it'd turn out to be around the same anyway.

---

I also think we should consider restructuring the filenames so there's a `types` directory similar to wlroots. This directory would contain most code, but for the more involved stuff like rendering, `container_at` and `container_get_in_direction` we could separate those into a different directory... maybe called `tasks`?

So the `sway/desktop`, `sway/input` and `sway/tree` directories could be replaced with something like:

    sway/types/container.c
    sway/types/criteria.c
    sway/types/input/cursor.c
    sway/types/input/input-manager.c
    sway/types/input/keyboard.c
    sway/types/input/seat.c
    sway/types/node.c
    sway/types/output.c
    sway/types/root.c
    sway/types/server.c
    sway/types/shells/layer_shell.c
    sway/types/shells/xdg_shell.c
    sway/types/shells/xdg_shell_v6.c
    sway/types/shells/xwayland.c
    sway/types/transaction.c
    sway/types/view.c
    sway/types/workspace.c

    sway/tasks/arrange.c
    sway/tasks/container_at.c
    sway/tasks/get_in_direction.c
    sway/tasks/render.c

---

I'm keen to get everyone's thoughts on this. Is this worth doing? Is there anything that would make this impractical?

### LINKED ISSUE #2552: Moving app between outputs moves cursor to wrong position
With two screens:

1. Open an app on the left one
2. mod+shift+right
3. Cursor is moved between the two screens, moved app becomes unfocused

RyanDwyer:
This commit changes the meaning of `sway_container` so that it only refers to layout containers and view containers. Workspaces, outputs and the root are no longer known as containers. Instead, root, outputs, workspaces and containers are all a type of `node`, and containers come in two types: layout containers and view containers.

In addition to the above, this implements type safe variables. This means we use specific types such as `sway_output` and `sway_workspace` instead of generic containers or nodes. However, it's worth noting that in a few places places (eg. seat focus and transactions) referring to them in a generic way is unavoidable which is why we still use nodes in some places.

If you want a TL;DR, look at `node.h`, as well as the struct definitions for root, output, workspace and container. Note that `sway_output` now contains a workspaces list, and workspaces now contain a tiling and floating list, and containers now contain a pointer back to the workspace.

--------------------------------------------------------------------------------

There are now functions for `seat_get_focused_workspace` and `seat_get_focused_container`. The latter will return `NULL` if a workspace itself is focused. Most other seat functions like `seat_get_focus` and `seat_set_focus` now accept and return nodes.

In the `config->handler_context` struct, `current_container` has been replaced with three pointers: `node`, `container` and `workspace`. `node` is the same as what `current_container` was, while `workspace` is the workspace that the node resides on and `container` is the actual container, which may be `NULL` if a workspace itself is focused.

The global `root_container` variable has been replaced with one simply called `root`, which is a pointer to the `sway_root` instance.

The way outputs are created, enabled, disabled and destroyed has changed. Previously we'd wrap the `sway_output` in a container when it is enabled, but as we don't have containers any more it needs a different approach. The `output_create` and `output_destroy` functions previously created/destroyed the container, but now they create/destroy the `sway_output`. There is a new function `output_disable` to disable an output without destroying it.

Containers have a new `view` property. If this is populated then the container is a view container, otherwise it's a layout container. Like before, this property is immutable for the life of the container.

Containers have both a `sway_container *parent` and `sway_workspace *workspace`. As we use specific types now, `parent` cannot point to a workspace so it'll be `NULL` for containers which are direct children of the workspace. The `workspace` property is set for all containers except those which are hidden in the scratchpad as they have no workspace.

In some cases we need to refer to workspaces in a container-like way. For example, workspaces have layout and children, but when using specific types this makes it difficult. Likewise, it's difficult for a container to get its parent's layout when the parent could be another container or a workspace. To make it easier, some helper functions have been created: `container_parent_layout` and `container_get_siblings`.

`container_remove_child` has been renamed to `container_detach` and `container_replace_child` has been renamed to `container_replace`.

`container_handle_fullscreen_reparent(con, old_parent)` has had the `old_parent` removed. We now unfullscreen the workspace when detaching the container, so this function is simplified and only needs one argument now.

`container_notify_subtree_changed` has been renamed to `container_update_representation`. This is more descriptive of its purpose. I also wanted to be able to call it with whatever container was changed rather than the container's parent, which makes bubbling up to the workspace easier.

There are now state structs per node thing. ie. `sway_output_state`, `sway_workspace_state` and `sway_container_state`.

The `focus`, `move` and `layout` commands have been completely refactored to work with the specific types. I considered making these a separate PR, but I'd be backporting my changes only to replace them again, and it's easier just to test everything at once.

Obviously this needs a ton of testing. I'm going to run this as my daily from now on.

Closes #2343 
Fixes #2552

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -47,48 +47,36 @@ void seat_destroy(struct sway_seat *seat) {
 		seat_device_destroy(seat_device);
 	}
 	sway_cursor_destroy(seat->cursor);
-	wl_list_remove(&seat->new_container.link);
+	wl_list_remove(&seat->new_node.link);
 	wl_list_remove(&seat->new_drag_icon.link);
 	wl_list_remove(&seat->link);
 	wlr_seat_destroy(seat->wlr_seat);
 }
 
-static struct sway_seat_container *seat_container_from_container(
-		struct sway_seat *seat, struct sway_container *con);
+static struct sway_seat_node *seat_node_from_node(
+		struct sway_seat *seat, struct sway_node *node);
 
-static void seat_container_destroy(struct sway_seat_container *seat_con) {
-	struct sway_container *con = seat_con->container;
-	struct sway_container *child = NULL;
-
-	if (con->children != NULL) {
-		for (int i = 0; i < con->children->length; ++i) {
-			child = con->children->items[i];
-			struct sway_seat_container *seat_child =
-				seat_container_from_container(seat_con->seat, child);
-			seat_container_destroy(seat_child);
-		}
-	}
-
-	wl_list_remove(&seat_con->destroy.link);
-	wl_list_remove(&seat_con->link);
-	free(seat_con);
+static void seat_node_destroy(struct sway_seat_node *seat_node) {
+	wl_list_remove(&seat_node->destroy.link);
+	wl_list_remove(&seat_node->link);
+	free(seat_node);
 }
 
 /**
  * Activate all views within this container recursively.
  */
-static void seat_send_activate(struct sway_container *con,
-		struct sway_seat *seat) {
-	if (con->type == C_VIEW) {
-		if (!seat_is_input_allowed(seat, con->sway_view->surface)) {
+static void seat_send_activate(struct sway_node *node, struct sway_seat *seat) {
+	if (node_is_view(node)) {
+		if (!seat_is_input_allowed(seat, node->sway_container->view->surface)) {
 			wlr_log(WLR_DEBUG, "Refusing to set focus, input is inhibited");
 			return;
 		}
-		view_set_activated(con->sway_view, true);
+		view_set_activated(node->sway_container->view, true);
 	} else {
-		for (int i = 0; i < con->children->length; ++i) {
-			struct sway_container *child = con->children->items[i];
-			seat_send_activate(child, seat);
+		list_t *children = node_get_children(node);
+		for (int i = 0; i < children->length; ++i) {
+			struct sway_container *child = children->items[i];
+			seat_send_activate(&child->node, seat);
 		}
 	}
 }
@@ -98,14 +86,15 @@ static void seat_send_activate(struct sway_container *con,
  * If con is a container, set all child views as active and don't enable
  * keyboard input on any.
  */
-static void seat_send_focus(struct sway_container *con,
-		struct sway_seat *seat) {
-	seat_send_activate(con, seat);
+static void seat_send_focus(struct sway_node *node, struct sway_seat *seat) {
+	seat_send_activate(node, seat);
 
-	if (con->type == C_VIEW
-			&& seat_is_input_allowed(seat, con->sway_view->surface)) {
+	struct sway_view *view = node->type == N_CONTAINER ?
+		node->sway_container->view : NULL;
+
+	if (view && seat_is_input_allowed(seat, view->surface)) {
 #ifdef HAVE_XWAYLAND
-		if (con->sway_view->type == SWAY_VIEW_XWAYLAND) {
+		if (view->type == SWAY_VIEW_XWAYLAND) {
 			struct wlr_xwayland *xwayland =
 				seat->input->server->xwayland.wlr_xwayland;
 			wlr_xwayland_set_seat(xwayland, seat->wlr_seat);
@@ -114,71 +103,67 @@ static void seat_send_focus(struct sway_container *con,
 		struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(seat->wlr_seat);
 		if (keyboard) {
 			wlr_seat_keyboard_notify_enter(seat->wlr_seat,
-					con->sway_view->surface, keyboard->keycodes,
+					view->surface, keyboard->keycodes,
 					keyboard->num_keycodes, &keyboard->modifiers);
 		} else {
 			wlr_seat_keyboard_notify_enter(
-					seat->wlr_seat, con->sway_view->surface, NULL, 0, NULL);
+					seat->wlr_seat, view->surface, NULL, 0, NULL);
 		}
 	}
 }
 
-void seat_focus_inactive_children_for_each(struct sway_seat *seat,
-		struct sway_container *container,
-		void (*f)(struct sway_container *container, void *data), void *data) {
-	struct sway_seat_container *current = NULL;
+void seat_for_each_node(struct sway_seat *seat,
+		void (*f)(struct sway_node *node, void *data), void *data) {
+	struct sway_seat_node *current = NULL;
 	wl_list_for_each(current, &seat->focus_stack, link) {
-		if (current->container->parent == NULL) {
-			continue;
-		}
-		if (current->container->parent == container) {
-			f(current->container, data);
-		}
+		f(current->node, data);
 	}
 }
 
 struct sway_container *seat_get_focus_inactive_view(struct sway_seat *seat,
-		struct sway_container *ancestor) {
-	if (ancestor->type == C_VIEW) {
-		return ancestor;
+		struct sway_node *ancestor) {
+	if (ancestor->type == N_CONTAINER && ancestor->sway_container->view) {
+		return ancestor->sway_container;
 	}
-	struct sway_seat_container *current;
+	struct sway_seat_node *current;
 	wl_list_for_each(current, &seat->focus_stack, link) {
-		struct sway_container *con = current->container;
-		if (con->type == C_VIEW && container_has_ancestor(con, ancestor)) {
-			return con;
+		struct sway_node *node = current->node;
+		if (node->type == N_CONTAINER && node->sway_container->view &&
+				node_has_ancestor(node, ancestor)) {
+			return node->sway_container;
 		}
 	}
 	return NULL;
 }
 
-static void handle_seat_container_destroy(struct wl_listener *listener,
-		void *data) {
-	struct sway_seat_container *seat_con =
-		wl_container_of(listener, seat_con, destroy);
-	struct sway_seat *seat = seat_con->seat;
-	struct sway_container *con = seat_con->container;
-	struct sway_container *parent = con->parent;
-	struct sway_container *focus = seat_get_focus(seat);
+static void handle_seat_node_destroy(struct wl_listener *listener, void *data) {
+	struct sway_seat_node *seat_node =
+		wl_container_of(listener, seat_node, destroy);
+	struct sway_seat *seat = seat_node->seat;
+	struct sway_node *node = seat_node->node;
+	struct sway_node *parent = node_get_parent(node);
+	struct sway_node *focus = seat_get_focus(seat);
 
 	bool set_focus =
 		focus != NULL &&
-		(focus == con || container_has_ancestor(focus, con)) &&
-		con->type != C_WORKSPACE;
+		(focus == node || node_has_ancestor(focus, node)) &&
+		node->type == N_CONTAINER;
 
-	seat_container_destroy(seat_con);
+	seat_node_destroy(seat_node);
 
 	if (set_focus) {
-		struct sway_container *next_focus = NULL;
+		struct sway_node *next_focus = NULL;
 		while (next_focus == NULL) {
-			next_focus = seat_get_focus_inactive_view(seat, parent);
+			struct sway_container *con =
+				seat_get_focus_inactive_view(seat, parent);
+			next_focus = con ? &con->node : NULL;
 
-			if (next_focus == NULL && parent->type == C_WORKSPACE) {
+			if (next_focus == NULL && parent->type == N_WORKSPACE) {
 				next_focus = parent;
 				break;
 			}
 
-			parent = parent->parent;
+			parent = node_get_parent(parent);
 		}
 
 		// the structure change might have caused it to move up to the top of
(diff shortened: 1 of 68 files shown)


sway/input/seat.c
```diff
@@ -727,11 +711,8 @@ void seat_set_focus_warp(struct sway_seat *seat,
 		if (config->mouse_warping && warp && new_output != last_output) {
```
ianyfan: Need a `container != NULL` check here, otherwise focusing workspaces is crashing.

RyanDwyer:
I'm confident that this branch is now stable enough to be merged. I'm still running it as my daily and all known issues are fixed.

I encourage others to run it as their daily as well and start reviewing.

SirCmpwn:
Now using this as my daily driver. Will merge in 2 days if I don't run into issues.

SirCmpwn:
Well, that was fast. I opened a floating window, split it horizontally and opened another window. First problem is that it was moved to 0,0 when I did so. Then I moved it back and started interactively resizing it, and split the right half vertically, then opened a third window. I tried to change the layout of the right pane and got a segfaut:

```
#0  0x00007ff7a7850049 in __strcasecmp_l_avx () at /home/sircmpwn/sources/sway/build/sway/../../../../../../lib64/libc.so.6
#1  0x00005579ece9d3fa in parse_layout_string (s=0x0) at ../sway/commands/layout.c:11
#2  0x00005579ece9d5f0 in get_layout_toggle (argc=1, argv=0x5579eec5c2c0, layout=L_VERT, prev_split_layout=L_NONE) at ../sway/commands/layout.c:64
        i = 1
        parsed = 21881
        curr = 1
#3  0x00005579ece9d6e6 in get_layout (argc=1, argv=0x5579eec5c2c0, layout=L_VERT, prev_split_layout=L_NONE) at ../sway/commands/layout.c:92
        parsed = L_NONE
#4  0x00005579ece9d7f9 in cmd_layout (argc=2, argv=0x5579eec5c2b8) at ../sway/commands/layout.c:125
        error = 0x0
        container = 0x5579ef2dec60
        workspace = 0x5579eea74ee0
        new_layout = L_NONE
        old_layout = L_VERT
#5  0x00005579ece77583 in execute_command (_exec=0x5579eec5c790 "layout toggle split", seat=0x5579eec3f0f0) at ../sway/commands.c:317
        res = 0x5579ecec1fb8
        argc = 3
        argv = 0x5579eec5c2b0
        handler = 0x5579eced4010 <command_handlers+80>
        results = 0x0
        exec = 0x5579ef218100 "layout toggle split"
        head = 0x0
        cmdlist = 0x0
        cmd = 0x5579ef218100 "layout toggle split"
        views = 0x0
        __PRETTY_FUNCTION__ = "execute_command"
#6  0x00005579ece99a22 in seat_execute_command (seat=0x5579eec3f0f0, binding=0x5579eec5d320) at ../sway/commands/bind.c:323
        binding_copy = 0x5579eec5d320
        reload = false
        results = 0x5579eec5d320
#7  0x00005579ece950fd in handle_keyboard_key (listener=0x5579ef0187f0, data=0x7ffec1f71180) at ../sway/input/keyboard.c:281
        binding_pressed = 0x5579eec5d320
        keyboard = 0x5579ef0187e0
        seat = 0x5579eec3f0f0
        wlr_seat = 0x5579eec3f220
        wlr_device = 0x5579ef018610
        event = 0x7ffec1f71180
        input_inhibited = false
        keycode = 61
        translated_keysyms = 0x5579ef02a958
        translated_modifiers = 64
        translated_keysyms_len = 1
        raw_keysyms = 0x5579ef02a958
        raw_modifiers = 64
        raw_keysyms_len = 1
        code_modifiers = 64
        handled = false
        binding_released = 0x0
        next_repeat_binding = 0x0
#8  0x00007ff7a7d86e4e in wlr_signal_emit_safe (signal=signal@entry=0x5579ef018788, data=data@entry=0x7ffec1f71180) at ../util/signal.c:29
        pos = 0x5579ef0187f0
        l = 0x5579ef0187f0
        cursor = {link = {prev = 0x5579ef0187f0, next = 0x7ffec1f71100}, notify = 0x7ff7a7d86dc0 <handle_noop>}
        end = {link = {prev = 0x7ffec1f710e0, next = 0x5579ef018788}, notify = 0x7ff7a7d86dc0 <handle_noop>}
#9  0x00007ff7a7d766d6 in wlr_keyboard_notify_key (keyboard=0x5579ef018690, event=event@entry=0x7ffec1f71180) at ../types/wlr_keyboard.c:107
        updated = <optimized out>
#10 0x00007ff7a7d5b1a7 in handle_keyboard_key (event=<optimized out>, libinput_dev=<optimized out>) at ../backend/libinput/keyboard.c:73
        wlr_dev = 0x5579ef018610
        kbevent = 0x5579ef2c5a00
        wlr_event = {time_msec = 5895433, keycode = 53, update_state = true, state = WLR_KEY_PRESSED}
        state = <optimized out>
#11 0x00007ff7a7d5a2dc in handle_libinput_readable (fd=<optimized out>, mask=<optimized out>, _backend=0x5579eea629b0) at ../backend/libinput/backend.c:35
        backend = 0x5579eea629b0
        event = 0x5579ef2c5a00
#12 0x00007ff7a7dbe702 in wl_event_loop_dispatch () at /home/sircmpwn/sources/sway/build/sway/../../../../../../lib64/libwayland-server.so.0
#13 0x00007ff7a7dbd2ac in wl_display_run () at /home/sircmpwn/sources/sway/build/sway/../../../../../../lib64/libwayland-server.so.0
#14 0x00005579ece82ddc in server_run (server=0x5579eced5a20 <server>) at ../sway/server.c:161
#15 0x00005579ece82579 in main (argc=3, argv=0x7ffec1f71568) at ../sway/main.c:444
        verbose = 0
        debug = 1
        validate = 0
        long_options = 
            {{name = 0x5579ecebfa4c "help", has_arg = 0, flag = 0x0, val = 104}, {name = 0x5579ecebfa51 "config", has_arg = 1, flag = 0x0, val = 99}, {name = 0x5579ecebfa58 "validate", has_arg = 0, flag = 0x0, val = 67}, {name = 0x5579ecebfa61 "debug", has_arg = 0, flag = 0x0, val = 100}, {name = 0x5579ecebfa67 "version", has_arg = 0, flag = 0x0, val = 118}, {name = 0x5579ecebfa6f "verbose", has_arg = 0, flag = 0x0, val = 86}, {name = 0x5579ecebfa77 "get-socketpath", has_arg = 0, flag = 0x0, val = 112}, {name = 0x0, has_arg = 0, flag = 0x0, val = 0}}
        config_path = 0x0
        usage = 0x5579ecebf5c0 "Usage: sway [options] [command]\n\n  -h, --help", ' ' <repeats 13 times>, "Show help message and quit.\n  -c, --config <config>  Specify a config file.\n  -C, --validate         Check the validity of the config file, th"...
        c = -1
        suid = false
```

SirCmpwn:
```
#0  0x00007f2058c6d049 in __strcasecmp_l_avx () at /home/sircmpwn/sources/sway/build/sway/../../../../../../lib64/libc.so.6
#1  0x00005638d1b9c3fa in parse_layout_string (s=0x0) at ../sway/commands/layout.c:11
#2  0x00005638d1b9c5f0 in get_layout_toggle (argc=1, argv=0x5638d3de1500, layout=L_VERT, prev_split_layout=L_NONE) at ../sway/commands/layout.c:64
        i = 1
        parsed = 22072
        curr = 1
#3  0x00005638d1b9c6e6 in get_layout (argc=1, argv=0x5638d3de1500, layout=L_VERT, prev_split_layout=L_NONE) at ../sway/commands/layout.c:92
        parsed = L_NONE
#4  0x00005638d1b9c7f9 in cmd_layout (argc=2, argv=0x5638d3de14f8) at ../sway/commands/layout.c:125
        error = 0x0
        container = 0x5638d3e1bc80
        workspace = 0x5638d374c160
        new_layout = L_NONE
        old_layout = L_VERT
#5  0x00005638d1b76583 in execute_command (_exec=0x5638d36b9a20 "layout toggle split", seat=0x5638d369c380) at ../sway/commands.c:317
        res = 0x5638d1bc0fb8
        argc = 3
        argv = 0x5638d3de14f0
        handler = 0x5638d1bd3010 <command_handlers+80>
        results = 0x0
        exec = 0x5638d3e17560 "layout toggle split"
        head = 0x0
        cmdlist = 0x0
        cmd = 0x5638d3e17560 "layout toggle split"
        views = 0x0
        __PRETTY_FUNCTION__ = "execute_command"
#6  0x00005638d1b98a22 in seat_execute_command (seat=0x5638d369c380, binding=0x5638d36ba5b0) at ../sway/commands/bind.c:323
        binding_copy = 0x5638d36ba5b0
        reload = false
        results = 0x5638d36ba5b0
#7  0x00005638d1b940fd in handle_keyboard_key (listener=0x5638d3a77bc0, data=0x7fff4ba54c30) at ../sway/input/keyboard.c:281
        binding_pressed = 0x5638d36ba5b0
        keyboard = 0x5638d3a77bb0
        seat = 0x5638d369c380
        wlr_seat = 0x5638d369c4b0
        wlr_device = 0x5638d3a77b30
        event = 0x7fff4ba54c30
        input_inhibited = false
        keycode = 61
        translated_keysyms = 0x5638d3a898e8
        translated_modifiers = 64
        translated_keysyms_len = 1
        raw_keysyms = 0x5638d3a898e8
        raw_modifiers = 64
        raw_keysyms_len = 1
        code_modifiers = 64
        handled = false
        binding_released = 0x0
        next_repeat_binding = 0x0
#8  0x00007f20591a3e4e in wlr_signal_emit_safe (signal=signal@entry=0x5638d3a64fc8, data=data@entry=0x7fff4ba54c30) at ../util/signal.c:29
        pos = 0x5638d3a77bc0
        l = 0x5638d3a77bc0
        cursor = {link = {prev = 0x5638d3a77bc0, next = 0x7fff4ba54bb0}, notify = 0x7f20591a3dc0 <handle_noop>}
        end = {link = {prev = 0x7fff4ba54b90, next = 0x5638d3a64fc8}, notify = 0x7f20591a3dc0 <handle_noop>}
#9  0x00007f20591936d6 in wlr_keyboard_notify_key (keyboard=0x5638d3a64ed0, event=event@entry=0x7fff4ba54c30) at ../types/wlr_keyboard.c:107
        updated = <optimized out>
#10 0x00007f20591781a7 in handle_keyboard_key (event=<optimized out>, libinput_dev=<optimized out>) at ../backend/libinput/keyboard.c:73
        wlr_dev = 0x5638d3a77b30
        kbevent = 0x5638d3df6b30
        wlr_event = {time_msec = 20094608, keycode = 53, update_state = true, state = WLR_KEY_PRESSED}
        state = <optimized out>
#11 0x00007f20591772dc in handle_libinput_readable (fd=<optimized out>, mask=<optimized out>, _backend=0x5638d34bf9b0) at ../backend/libinput/backend.c:35
        backend = 0x5638d34bf9b0
        event = 0x5638d3df6b30
#12 0x00007f20591db702 in wl_event_loop_dispatch () at /home/sircmpwn/sources/sway/build/sway/../../../../../../lib64/libwayland-server.so.0
#13 0x00007f20591da2ac in wl_display_run () at /home/sircmpwn/sources/sway/build/sway/../../../../../../lib64/libwayland-server.so.0
#14 0x00005638d1b81ddc in server_run (server=0x5638d1bd4a20 <server>) at ../sway/server.c:161
#15 0x00005638d1b81579 in main (argc=3, argv=0x7fff4ba55018) at ../sway/main.c:444
        verbose = 0
        debug = 1
        validate = 0
        long_options = 
            {{name = 0x5638d1bbea4c "help", has_arg = 0, flag = 0x0, val = 104}, {name = 0x5638d1bbea51 "config", has_arg = 1, flag = 0x0, val = 99}, {name = 0x5638d1bbea58 "validate", has_arg = 0, flag = 0x0, val = 67}, {name = 0x5638d1bbea61 "debug", has_arg = 0, flag = 0x0, val = 100}, {name = 0x5638d1bbea67 "version", has_arg = 0, flag = 0x0, val = 118}, {name = 0x5638d1bbea6f "verbose", has_arg = 0, flag = 0x0, val = 86}, {name = 0x5638d1bbea77 "get-socketpath", has_arg = 0, flag = 0x0, val = 112}, {name = 0x0, has_arg = 0, flag = 0x0, val = 0}}
        config_path = 0x0
        usage = 0x5638d1bbe5c0 "Usage: sway [options] [command]\n\n  -h, --help", ' ' <repeats 13 times>, "Show help message and quit.\n  -c, --config <config>  Specify a config file.\n  -C, --validate         Check the validity of the config file, th"...
        c = -1
        suid = false
```

RyanDwyer:
Fixed both the 0,0 issue when splitting and the layout crash.

SirCmpwn:
This seems to have been working fine for me. Thanks!

PR #2542, opened 2018-08-30
FreeBSD fixes

sghctoma:
Hi all,

I have been using Sway on my HardenedBSD 12-CURRENT machine for a few months now, and I think maybe it's time to PR my patches.

Please be aware, that a vanilla FreeBSD won't be able to build or run Sway yet, because there are some requirements that have older versions in the ports tree than necessary. I needed to make the following modifications to my system:

- devel/meson needs to be updated to 0.47.1 (0.46.1 is currently in the port tree)
- graphics/wayland-protocols needs to be updated to 1.15 (1.12 is currently in the port tree)
- devel/epool-shim needs to be updated to commit 7fff1fc. Neither the one in the ports tree, nor the latest version works. I will probably open a PR there, but first I would like to understand a bit more about why it does not work.
- x11/libinput needs to be updated to >=1.7.0 (1.6.0 is currently in the ports tree). This can be done by installing the following ports from https://github.com/FreeBSDDesktop/freebsd-ports:
  - devel/evdev-proto
  - multimedia/v4l_compat
  - x11/libinput
  - devel/py-pyudev


SirCmpwn:
Can you chase down the relevant FreeBSD maintainers to get those bullet points addressed?

SirCmpwn:
Also, do you have the time and inclination to be the official FreeBSD point-of-contact for sway?

sway/commands/output/background.c
```diff
@@ -1,4 +1,4 @@
-#define _XOPEN_SOURCE 500
+#define _XOPEN_SOURCE 600
```
SirCmpwn: I'd rather replace these with _POSIX_C_SOURCE as appropriate

sghctoma: It's done.

sghctoma:
The libinput update is already underway, I will look into the epoll-shim issue this week, and will also try to contact the other maintainers.

As for being the official point-of-contact, I have really no idea how much more time would it consume compared to just patching whenever something fails on my own system, but I think I could try.

SirCmpwn:
Thanks!

SirCmpwn:
>It probably means getting in touch with dependencies maintainers in case something breaks.

Aye, it means "keep using sway on FreeBSD and test each release before we tag it so we don't ship releases which break FreeBSD". Also, keeping sway up to date in ports.

sghctoma:
Thanks for the merge!

> Is there a pending patch for this one?

There is now. Submitted a PR for 0.47.2

> I submitted a patch to update to 1.16 a few weeks ago. Haven't heard about it yet.

Yup, sadly things can be slow here...

> Hmm, it worked a few weeks ago with the version in the ports tree. Maybe something's changed since then.

That's strange, the ports tree has a 2016 december version, which contains a bug that causes both Weston and rootston to exit with SIGUSR1 (jiixyj/epoll-shim#2). May I ask which FreeBSD version did you use? Also, did you try it with rootston, or something else? 

> It probably means getting in touch with dependencies maintainers in case something breaks.

> Aye, it means "keep using sway on FreeBSD and test each release before we tag it so we don't ship releases which break FreeBSD". Also, keeping sway up to date in ports.

I think I can do that, if these few days long silences/delays are acceptable.

emersion:
> There is now. Submitted a PR for 0.47.2

Nice

>May I ask which FreeBSD version did you use? Also, did you try it with rootston, or something else?

Ah, I disabled Xwayland completely.

>I think I can do that, if these few days long silences/delays are acceptable.

Yeah, no problem with that. Work at your own pace.

PR #2548, opened 2018-09-01
Add test to run sway

arandomhuman:
Sets the PATH variable correctly so that it finds the auxilliary
executables.
To use: run "meson test -v sway" in the build directory.

Added an "interactive-tests" option for when actual tests are added.

---

Seems helpful for iterative testing. Input on naming/formatting welcome.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -217,3 +217,13 @@ if (get_option('fish_completions'))
 
 	install_data(fish_files, install_dir: fish_install_dir)
 endif
+
+if (get_option('interactive-tests'))
+	# A test to run the just built sway executable, with all the auxilliaries
+	# in the path.
+	sway_paths_find_cmd = 'find @0@ -maxdepth 1 -type d -name \'sway*\' -printf \':%p\''.format(meson.current_build_dir())
+	sway_paths = run_command('sh', '-c', sway_paths_find_cmd).stdout().strip()
+	current_env = environment()
+	current_env.append('PATH', sway_paths)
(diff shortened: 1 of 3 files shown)


SirCmpwn:
See https://github.com/swaywm/sway-tests

arandomhuman:
@SirCmpwn The "test" simply opens sway after setting the PATH variable. The point was simply to get a sway instance up after compilation without actually installing it, instead of replicating sway-tests. I found it convenient when having a global (usable) sway installation. YMMV.

If you think it's useful but rather be implemented in a different manner (as a script?), let me know.

SirCmpwn:
I don't think it should live in this repository at all.

SirCmpwn:
When I want to test my local sway build I just run `./sway/sway`

arandomhuman:
Fair enough.

Simply running `./sway/sway` doesn't pick up the instances of swaybg, swaymsg, etc. in the build directory, which was the motivation.

Also, quick question, since this is closed: scan-build for sway gave me ~55 bugs. Should I open a WIP pull request or send them in bunches?
Just trying my hand at low-hanging fruit to get up to speed.

SirCmpwn:
I use a separate config file, which specifies e.g. `swaybg_command`.

Regarding scan stuff, either way is fine by me.

RyanDwyer:
You might want to wait until the type safe PR is merged before actioning scan-build bugs. That PR refactors a lot of stuff.

PR #2549, opened 2018-09-01
Implement deprecated new_window and new_float commands

RyanDwyer:
May as well make it as easy as possible for users who are coming from i3.

This also changes the `border` command to accept a thickness when setting the border to normal. This makes it work the same way as the `default_border` command. Eg. `border normal 5`

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -118,6 +118,8 @@ static struct cmd_handler handlers[] = {
 	{ "input", cmd_input },
 	{ "mode", cmd_mode },
 	{ "mouse_warping", cmd_mouse_warping },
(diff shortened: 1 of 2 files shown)


ianyfan:
Should (does) sway inform (maybe nag) the user that they're using a deprecated command?

RyanDwyer:
My preference is no, because supporting the commands is just two lines while adding a nagbar warning is quite a bunch more (create new files etc). I imagine we'll just always support them.

SirCmpwn:
Thanks!

emersion:
Alternatively we could warn about deprecated config options via logs

SirCmpwn:
¯\\\_(ツ)_/¯

PR #2550, opened 2018-09-01
Implement window_type criteria token

RyanDwyer:
To test:

    for_window [window_type="dialog"] title_format "%title (lolz)"

Then open Firefox and press `Ctrl+O` to open the Open File dialog.

BTW, we can't store the actual atom value in the criteria because xwayland is initialised lazily, so I store the `enum atom_name` value instead. And because 0 is a valid enum value, I use `ATOM_LAST` as the "no value set" value.


SirCmpwn:
Thanks!

PR #2560, opened 2018-09-02
Implement focus_on_window_activation

RyanDwyer:
Depends on https://github.com/swaywm/wlroots/pull/1223

i3 docs: https://i3wm.org/docs/userguide.html#focus_on_window_activation

To test, I opened Firefox along with a terminal and ran `firefox http://github.com` from the terminal.


emersion:
Thanks!

PR #2562, opened 2018-09-02
Add meson option for LD_LIBRARY_PATH

### LINKED ISSUE #2262: Set LD_LIBRARY_PATH at compile time
Hi folks!

I'm really enjoying using the sway 1.0 alphas (which I need since my display is rotated so needed wlroots) o far but have a simple problem which likely has a simple fix.

With the older sway builds the following option did the trick for NixOS (since it uses funky paths)

`-DLD_LIBRARY_PATH=/run/opengl-driver/lib:/run/opengl-driver-32/lib`

I can't find an equivalent option for the newer meson / ninja build - can anyone point me in the right direction?

RyanDwyer:
Also makes the meson option names consistently use dashes.

Fixes #2262.


emersion:
Thanks!

PR #2563, opened 2018-09-03
fix misaligned integer stores/loads

taiyu-len:
silences releated warnings from the undefined behavior sanitizer.

### DIFF
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -78,8 +78,8 @@ struct ipc_response *ipc_recv_response(int socketfd) {
 	}
 
 	total = 0;
-	response->size = data32[0];
-	response->type = data32[1];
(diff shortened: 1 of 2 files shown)


ascent12:
An alternate solution is to use scatter/gather I/O to prevent misaligned accesses and byte-level copying in the first place.
```c
char magic[sizeof(ipc_magic)];
uint32_t data[2];

struct iovec iovec[2] = {
    { .iov_base = magic, iov_len = sizeof(magic) },
    { .iov_base = data, iov_len = sizeof(data) },
};
struct msghdr msg = {
    .msg_iov = iovec,
    .msg_iovlen = 2,
}

...

recvmsg(socket_fd, &msg, 0);
```

SirCmpwn:
Thanks!

PR #2564, opened 2018-09-03
Fix window_type with disabled xwayland support

ppascher:
There were errors building with "enable-xwayland" set to "false" which were introduced by #2550.
This commit places those changes within "#ifdef HAVE_XWAYLAND" checks.

### DIFF
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -23,15 +23,15 @@ struct criteria {
 	pcre *title;
 	pcre *shell;
 	pcre *app_id;
-	pcre *class;
-	pcre *instance;
 	pcre *con_mark;
 	uint32_t con_id; // internal ID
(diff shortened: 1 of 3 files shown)


emersion:
What about window role? Doesn't it need the same treatment?

RyanDwyer:
Only `window_type` is needed because I had changed the data type to `enum atom_name`, and that datatype is defined in `xwayland.h`.

`window_role`, `class` and `instance` are all specific to xwayland but compile fine without xwayland support. Maybe they should all be inside the `HAVE_XWAYLAND` checks? We might also want to update the documentation to say which properties won't work without xwayland support.

ppascher:
Thank you @emersion & @RyanDwyer . I placed those three criteria inside xwayland checks as you suggested.

SirCmpwn:
Thanks!

PR #2565, opened 2018-09-03
Fix crash when a splash screen opens on an empty workspace

ianyfan:
This fixes a crash I got when opening GIMP on an empty workspace, though I'm not sure why it's required.

Also, if this check is required, should it be `children->length == 0 ||` as it is here, or should it be changed to `children->length > 0 &&` (i.e. the opposite)?

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -346,7 +346,8 @@ static void render_titlebar(struct sway_output *output,
 	float output_scale = output->wlr_output->scale;
 	enum sway_container_layout layout = state->parent->current.layout;
 	list_t *children = state->parent->current.children;
-	bool is_last_child = children->items[children->length - 1] == con;
+	bool is_last_child = children->length == 0 ||
+		children->items[children->length - 1] == con;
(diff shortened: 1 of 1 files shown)


RyanDwyer:
This bug was likely introduced when the magical floating container was removed. The parent of floating views used to be the magical floating container, but now it's the workspace.

The code you changed assumes the current container will be in the parent's children list, but in this case the parent is a workspace, the view is floating and `workspace->children` is the tiling list. If there are no tiling containers in the workspace then it would crash. I don't understand why it doesn't always happen though. I can't reproduce it.

`is_last_child` is used to help determine whether the border around the title area should be fully boxed in, or whether the side borders of the container should flow into the title background. For floating views they should always flow/connect, so for the purpose of the check we want `is_last_child` to be `true` if there are no tiling children. Which means `children->length == 0 ||` is the correct expression.

emersion:
Thanks!

ianyfan:
@RyanDwyer Thanks for the explanation

PR #2571, opened 2018-09-04
Fix swaybar block background fill logic

ammgws:
When a block is padded/offset to uphold `min_width`, the background colour is now applied to the whole block rather than just the original width before padding. This brings the behaviour in-line with i3bar.

Related issue: https://github.com/greshake/i3status-rust/issues/155


SirCmpwn:
Thanks!

PR #2572, opened 2018-09-04
Call wlr_output_enable on enable/disable if needed

### LINKED ISSUE #2557: Outputs disabled in Sway config aren't disabled in wlroots
In `apply_output_config`, `wlr_output_enable(wlr_output, false)` is not called when `oc->enabled == 0`. The tricky part is that when `enabled == 1`, we don't always want to `wlr_output_enable(wlr_output, true)` (when DPMS is on, we don't want to turn on screen).

RedSoxFan:
Fixes #2557 

This PR:
- Calls `wlr_output_enable(wlr_output, false)` when disabling an output.
- Calls `wlr_output_enable(wlr_output, true)` when (re-)enabling an output **_if_** DPMS is not set to `DPMS_OFF`

I also refactored `apply_output_config` to:
- ~Take in a `struct sway_output` instead of `struct sway_container`~ (Taken care of by typesafety)
- Handle (re-)enabling the output instead of having it done before in `apply_output_config_to_outputs`
   - ~This also required setting `sway_output->swayc` in `output_create` instead of `output_enable` to avoid an infinite loop~ (No longer needed with typesafety)

Since typesafety accidentally reverted part of #2553, I also removed the modeset from `output_create`.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -178,12 +178,21 @@ void apply_output_config(struct output_config *oc, struct sway_output *output) {
 	struct wlr_output *wlr_output = output->wlr_output;
 
 	if (oc && oc->enabled == 0) {
-		if (output->bg_pid != 0) {
-			terminate_swaybg(output->bg_pid);
-			output->bg_pid = 0;
+		if (output->enabled) {
+			if (output->bg_pid != 0) {
+				terminate_swaybg(output->bg_pid);
+				output->bg_pid = 0;
+			}
+			output_disable(output);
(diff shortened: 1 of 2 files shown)


RyanDwyer:
This is going to have conflicts with the typesafety branch. You'd be doing me a favour if you could rebase this off that.

RedSoxFan:
> This is going to have conflicts with the typesafety branch. You'd be doing me a favour if you could rebase this off that.

Done. typesafety actually made this PR smaller since half the changes were already done/no longer needed.

SirCmpwn:
Thanks!

PR #2578, opened 2018-09-05
Fix management of bindings during reload

RyanDwyer:
`seat_execute_command` needs to check the flags on `binding_copy`, as `binding` will be a dangling pointer after a reload command.

`handle_keyboard_key` needs to set the `next_repeat_binding` for non-reloads prior to executing the command in case the binding is freed by the reload command.


emersion:
Thanks!

ianyfan:
This still doesn't work if the keyboard is configured to a seat, since it destroys the keyboard, leaving it a dangling pointer.

PR #2582, opened 2018-09-06
Remove __PRETTY_FUNCTION__

ascent12:
This is a non-standard extension as well as completely useless in C.
__func__ is the standard way of doing this.


SirCmpwn:
Thanks!

PR #2587, opened 2018-09-06
Fix crash when running "layout toggle"

RyanDwyer:
The `argc` and `argv` used in this function are the same ones used by the layout command itself.


RedSoxFan:
Thanks!

PR #2589, opened 2018-09-06
Introduce seat_set_focus_container and seat_set_focus_workspace

RyanDwyer:
These are the same as `seat_set_focus`, but accept a specific type rather than using nodes. Doing this adds more typesafety and lets us avoid using `&con->node` which looks a little ugly.

This fixes a crash that pretty much nobody would ever come across. If you have a bindsym for `focus` with no arguments and run it from an empty workspace, sway would crash because it assumes `container` is not `NULL` (fixed in `focus.c`).


emersion:
Thanks!

PR #2592, opened 2018-09-06
Second attempt at restoring focus when closing a fullscreen view

RyanDwyer:
To reproduce the problem this is fixing, create `H[view view view]`, fullscreen one of the views and close it. The entire workspace will be given focus rather than one of the siblings.

This happens because we emit the `destroy` event, so the seat code tries to find a new focus, but the view it finds is still believed to be hidden by the fullscreen view so it's discarded and the workspace is used instead.

This clears the workspace's fullscreen pointer prior to emitting the destroy event so that the seat code finds an appropriate new focus.


emersion:
Thanks!

PR #2594, opened 2018-09-06
Float views when min == max on either dimension

### LINKED ISSUE #2181: pinentry-gtk-2 doesn't get frame events
I figured out why I had issues with the split layout here: https://github.com/swaywm/sway/issues/2161

I was using pinentry-gtk-2, not pinentry-gnome3.

RyanDwyer:
This fixes pinentry-gtk-2, but might make other views floating which would otherwise be tiled. This patch is more of a trial which could end up becoming a permanent fix.

Fixes #2181.


emersion:
Thanks!

PR #2600, opened 2018-09-06
Fix workspace switching

### LINKED ISSUE #2599: `workspace n` where n is an inactive workspace on an inactive output does not work
swaybar updates, but the inactive workspace does not become active and keyboard focus is given to the active workspace on that output.

1. Open workspace 1 on output A, put windows in it
2. Open workspace 2 on output A, put windows in it
3. Open workspace 3 on output B, focus it
4. `workspace 1`

RyanDwyer:
The output also needs to be made dirty when focusing a new output.

Fixes #2599.


SirCmpwn:
Thanks!

PR #2602, opened 2018-09-07
Fix gaps in title textures and vertically center them

### LINKED ISSUE #2455: Gap in titlebar when some windows use higher glyphs
screenshot

Reproduce with: `urxvt -title 'Hello 世界' -e /usr/bin/sleep 20h`

Maybe related to #1936?

RyanDwyer:
There was code that attempted to fill in the gap below the title texture when the texture isn't tall enough, but this only worked when the output was positioned at 0,0. The reason is that `render_rect` expects a box passed in a hybrid layout-local/output-buffer-local system, and we were passing purely output-buffer-local. I've added a comment documenting the expected format.

By the way, we can't use layout-local coordinates for the rectangle box because in some cases we set the box based on a texture size. Texture sizes are buffer-local, and we'd have to divide them to bring it back to layout-local which means losing precision. We could use output-buffer-local coordinates for the box, but this would require translating the coordinates from layout-local to output-buffer-local in many places during rendering.

This patch also vertically centers the text inside the title bar.

I've tested this with the following:

* Outputs with a position other than 0,0
* Scaled outputs
* Marks

Fixes #2455.


SirCmpwn:
@minus7 can you try this?

minus7:
Will do

minus7:
Close but not quite right: the baseline (added in white) of both texts is different by one pixel (text without the kanji is too far up).
screen2
Otherwise looks good to me

RyanDwyer:
I've implemented baseline alignment now. Note that marks are not aligned to a baseline, nor are marks even considered when calculating the config's font height. I'll address that in a different PR.

minus7:
Looks great now, thanks. Do you think the spacing on the left side should be increased?

RyanDwyer:
I don't think so. When I implemented title bars I took pixel measurements from i3, though I can't remember if I tried it with non-default border thicknesses.

If the spacing bothers you can you use `for_window [title=".*"] title_format " %title"`.

PR #2609, opened 2018-09-08
 commands: prevent running empty seat command

ianyfan:
Fixes crash when running `seat <seat>` without a sub-command.


SirCmpwn:
Thanks!

PR #2610, opened 2018-09-09
[WIP] add support for cursor themes

kamiyaa:
- uses environment variables `SWAY_CURSOR_THEME` and `SWAY_CURSOR_SIZE`

First commit adds support in sway. Second commit adds support in swaybar
Currently, the code is also not very good (`getenv()` scattered everywhere).
Feedback is appreciated.

Thanks!

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -507,11 +507,15 @@ void seat_remove_device(struct sway_seat *seat,
 
 void seat_configure_xcursor(struct sway_seat *seat) {
 	// TODO configure theme and size
-	const char *cursor_theme = NULL;
+	const char *cursor_theme = getenv("SWAY_CURSOR_THEME");
+	int cursor_size = atoi(getenv("SWAY_CURSOR_SIZE"));
+	if (cursor_size <= 0) {
+		cursor_size = 24;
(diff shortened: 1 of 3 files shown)


SirCmpwn:
I think I would rather see this done as `input` subcommands.

kamiyaa:
I don't know when(if) I will have the time to implement this as school recently started.
Apologize in advance.

SirCmpwn:
No problem, we're not in a hurry.

kamiyaa:
By `input` subcommands, do you mean the files under `sway/commands/input`?

so is it going to end up something like:
```
input "2:7:SynPS/2_Synaptics_TouchPad" {
        dwt enabled
        natural_scroll enabled
        tap enabled
        cursor_theme Adwaita 24
}
```
But if that were the case, it would look like different input devices can have different cursors..?

SirCmpwn:
More like cursor subcommands, actually.

```
seat seat0 {
    cursor_theme Adwaita 24
}
```

SirCmpwn:
bump?

kamiyaa:
uhh, I will close this as I don't have the time or expertise right now.

Issue #2558, opened 2018-09-02
Mouse focus not switched on workspace changes


atomnuker:
While keyboard focus does get switched, mouse focus does not unless mouse is moved. So scroll wheel events get sent to the client on the other workspace.

RyanDwyer:
This was added in the past, but caused issues so it was removed temporarily.

Here's where it was originally added:

    commit 03d49490ccff3c5c81bea73622c8616fa61eb3dd
    Author: emersion
    Date:   Tue Jul 17 23:04:55 2018 +0100
    
        Update cursor on focus change
        
        This allows to send wl_pointer.enter when switching between views
        in a split/tabbed layout for instance. This (1) updates the cursor
        image accordingly (2) makes it unnecessary to move the mouse before
        scrolling. It's harmless to always call cursor_send_pointer_motion
        because in case the focused surface hasn't changed this is a no-op.
        
        The `last_focus != NULL` condition is required otherwise
        cursor_send_pointer_motion will crash when sway starts up (the
        sway_output doesn't yet have a workspace).

Here's where it was attempted to be fixed:

    commit b14bd1b0b1536039e4f46fe94515c7c44e7afc61
    Author: Ryan Dwyer
    Date:   Wed Jul 25 19:26:12 2018 +1000
    
        Fix crash when closing last child of a tabbed container
        
        The crash only occurs if the mouse cursor is above the tabbed container
        when the last child is closed.
        
        Introduced in 03d49490ccff3c5c81bea73622c8616fa61eb3dd, over a week ago
        and unnoticed until now :O
        
        The above commit changes the behaviour of a focus change. When you
        change focus, it sends pointer motion which makes the client set a new
        cursor image. We already had this behaviour for workspace switching, but
        this commit adds it for view switching too, such as in a tabbed
        container or when closing a view.
        
        The sequence of events that leads to the crash is:
        
        * The last child of a tabbed container unmaps, which triggers a
        `destroy` event before we've cleaned up the child or reaped the tabbed
        container.
        * The seat code listens to the `destroy` event and removes the seat
        container from the focus stack. As part of this, it decides to set focus
        to the parent (my fix alters this decision).
        * When setting focus to the new parent, the container motion is sent as
        per the previously mentioned commit.
        * The motion code uses `container_at`, which encounters the tabbed
        container and its child in a half destroyed state, and everything blows
        up from there.
        
        `con->parent` is needed because scratchpad containers don't have parents
[...]

PR #2612, opened 2018-09-10
Attempted pointer focus fix

### LINKED ISSUE #2401: Switching focus with keyboard doesn't update pointer focus

wmww:
Fixes #2401 (aka #2558)

I know fixes for this issue have had side effects in the past, hopefully this won't but I can't guarantee it. I'm running this on my primary machine though, and I'll report any problems I have.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -10,6 +10,7 @@
 #include "sway/desktop.h"
 #include "sway/desktop/idle_inhibit_v1.h"
 #include "sway/desktop/transaction.h"
+#include "sway/input/cursor.h"
 #include "sway/output.h"
 #include "sway/tree/container.h"
 #include "sway/tree/node.h"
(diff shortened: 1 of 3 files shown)


RyanDwyer:
Sending cursor motion doesn't really belong in `transaction_commit`. It could have unintended side effects by sending motion too frequently. For example, GTK menus can close immediately on click if they receive a motion event between the press and release. Maybe we should send motion towards the end of `view_unmap` (after arranging), again in the workspace command after focusing the new workspace, and again in the focus command after changing focus.

Also, there is value in having `node_is_view` trigger a crash if `node` is `NULL`. If it silently handles the situation then it can hide potential errors and make them go undiscovered during development. And changing the calling code to `if (node && node_is_view(node))` makes it clear to the reader that `node` might be `NULL` at this point.

wmww:
Acknowledged, Thanks for looking it over. I'll keep running it locally, and possibly update with your suggestions when I have more time. Unlikely to happen in the next few days though, so you can close this if you wan't to.

wmww:
I moved this to the above-mentioned PR.

PR #2615, opened 2018-09-10
tree/container: fix null pointer dereferencing

nearffxx:
fix issue #2614


nearffxx:
it seems it was fixed by #2401

PR #2617, opened 2018-09-10
Give windows pointer focus immediately

### LINKED ISSUE #2401: Switching focus with keyboard doesn't update pointer focus

wmww:
Fixes #2401 (aka #2558)

Based on feedback on #2612

Previously, when switching windows, pointer focus was not changed until the pointer was moved. This makes the pointer enter happen immediately, without the side effects of other attempted fixes.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -246,7 +246,7 @@ static struct sway_container *container_at_stacked(struct sway_node *parent,
 
 	// Surfaces
 	struct sway_node *current = seat_get_active_child(seat, parent);
-	return tiling_container_at(current, lx, ly, surface, sx, sy);
+	return current ? tiling_container_at(current, lx, ly, surface, sx, sy) : NULL;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


sway/tree/container.c
```diff
@@ -218,7 +218,7 @@ static struct sway_container *container_at_tabbed(struct sway_node *parent,
 
 	// Surfaces
 	struct sway_node *current = seat_get_active_child(seat, parent);
-	return tiling_container_at(current, lx, ly, surface, sx, sy);
+	return current ? tiling_container_at(current, lx, ly, surface, sx, sy) : NULL;
```
emersion: Can you add a similar check line 249?

wmww: Sure, one sec...

sway/tree/view.c
```diff
@@ -583,6 +584,7 @@ void view_unmap(struct sway_view *view) {
 	}
 
 	transaction_commit_dirty();
+	cursor_send_pointer_motion(config->handler_context.seat->cursor, 0, true);
```
emersion: Can you move this below `view->surface = NULL` so we don't have bad pointers when calling `cursor_send_pointer_motion`?

sway/tree/container.c
wmww: done

sway/tree/view.c
wmww: done

SirCmpwn:
Thanks!

PR #2618, opened 2018-09-11
Implement tiling drag

RyanDwyer:
Hold `floating_modifier` then click and drag a tiling view to move it around.

This is based of an open PR in i3: https://github.com/i3/i3/pull/3085#issuecomment-379487507

* A new config directive `tiling_drag` can disable the functionality (as per i3). It's enabled by default.
* The colour of the drop zone is the indicator colour. In i3 this is a solid colour, but in sway I gave it 50% transparency. I don't think i3 can make it transparent without introducing compositing into i3, so we have an advantage here.
* Views within a floating container cannot be moved around (we do the move-floating behaviour instead)
* This functionality intentionally cannot be triggered by a regular click and drag on the title bar. i3 doesn't do this - instead i3 starts resizing the split. We should probably do the same.

Some more specific stuff:

* While dragging, move the cursor to the edges of a another container. If the edge is parallel to the container's parent layout then the view will be inserted as a sibling, otherwise it'll split the container and insert it into the split.
* You can also drop a view into the center of another container, which adds it as a sibling regardless of the container's parent layout.
* If you hover close to the perpendicular edge of a layout container then you can insert it as a sibling of the layout container. For example, create layout `H[view V[view view]]` then move one of the right views to the left edge of the vertical container.
* You can do the same for workspaces: `H[view view]` then move a view to the top.
* You can drop a view into an empty workspace.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -279,8 +279,11 @@ void container_add_child(struct sway_container *parent,
 void container_insert_child(struct sway_container *parent,
 		struct sway_container *child, int i);
 
+/**
+ * Side should be 0 to add before, or 1 to add after.
+ */
 void container_add_sibling(struct sway_container *parent,
-		struct sway_container *child);
+		struct sway_container *child, int side);
 
 void container_detach(struct sway_container *child);
 
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -228,6 +228,141 @@ static void handle_move_floating_motion(struct sway_seat *seat,
 	desktop_damage_whole_container(con);
 }
 
+static void resize_box(struct wlr_box *box, enum wlr_edges edge,
+		size_t thickness) {
+	switch (edge) {
+	case WLR_EDGE_TOP:
+		box->height = thickness;
+		break;
+	case WLR_EDGE_LEFT:
+		box->width = thickness;
+		break;
+	case WLR_EDGE_RIGHT:
+		box->x = box->x + box->width - thickness;
+		box->width = thickness;
+		break;
+	case WLR_EDGE_BOTTOM:
+		box->y = box->y + box->height - thickness;
+		box->height = thickness;
+		break;
+	case WLR_EDGE_NONE:
+		box->x += thickness;
+		box->y += thickness;
+		box->width -= thickness * 2;
+		box->height -= thickness * 2;
+		break;
+	}
+}
+
+static void handle_move_tiling_motion(struct sway_seat *seat,
+		struct sway_cursor *cursor) {
+	struct wlr_surface *surface = NULL;
+	double sx, sy;
+	struct sway_node *node = node_at_coords(seat,
+			cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
+	// Damage the old location
+	desktop_damage_box(&seat->op_drop_box);
+
+	if (!node) {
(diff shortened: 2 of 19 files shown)


include/sway/tree/container.h
```diff
@@ -279,8 +279,11 @@ void container_add_child(struct sway_container *parent,
 void container_insert_child(struct sway_container *parent,
 		struct sway_container *child, int i);
 
+/**
+ * Side should be 0 to add before, or 1 to add after.
+ */
 void container_add_sibling(struct sway_container *parent,
-		struct sway_container *child);
+		struct sway_container *child, int side);
```
emersion: Maybe we can use a `bool` here, and rename to `after`?

sway/input/cursor.c
```diff
@@ -228,6 +228,141 @@ static void handle_move_motion(struct sway_seat *seat,
...
+	// Traverse the ancestors, trying to find a layout container perpendicular
+	// to the edge. Eg. close to the top or bottom of a horiz layout.
+	while (con) {
+		enum wlr_edges edge = WLR_EDGE_NONE;
+		enum sway_container_layout layout = container_parent_layout(con);
+		struct wlr_box parent;
+		con->parent ? container_get_box(con->parent, &parent) :
+			workspace_get_box(con->workspace, &parent);
+		if (layout == L_HORIZ || layout == L_TABBED) {
+			if (cursor->cursor->y < parent.y + 30) {
```
emersion: Maybe we can define `30` as a constant?

sway/input/cursor.c
```diff
@@ -228,6 +228,141 @@ static void handle_move_motion(struct sway_seat *seat,
 	desktop_damage_whole_container(con);
 }
 
+static void resize_box(struct wlr_box *box, enum wlr_edges edge,
+		size_t thickness) {
```
emersion: `thickness` should be an `int`

sway/input/cursor.c
```diff
@@ -751,6 +889,14 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 		}
 	}
 
+	// Handle moving a tiling container
+	if (config->tiling_drag && mod_pressed && !is_floating_or_child &&
+			!cont->is_fullscreen) {
```
emersion: `state == WLR_BUTTON_PRESSED`?

sway/input/seat.c
```diff
@@ -1015,13 +1025,78 @@ void seat_begin_resize_tiling(struct sway_seat *seat,
...
+	container_detach(con);
+	if (old_parent) {
+		container_reap_empty(old_parent);
+	}
+
+	// Moving container into empty workspace
+	if (target_node->type == N_WORKSPACE && edge == WLR_EDGE_NONE) {
+		workspace_add_tiling(new_ws, con);
+
+	// Moving container before/after another
```
emersion: This comment indentation is weird. Move it in the next branch?

emersion:
Minor bug: with a split layout with two views, dragging the left container on the center of the right one works (ie. switches views), but the other way around doesn't (is a no-op).

RyanDwyer:
> Minor bug: with a split layout with two views, dragging the left container on the center of the right one works (ie. switches views), but the other way around doesn't (is a no-op).

That's expected. Dragging into the center means "move after this one". If the view being moved is already the next sibling then it's a no op. I think this is what the i3 PR does too.

PR #2620, opened 2018-09-11
commands: when moving a container, restore focus properly

ianyfan:
Fixes 2 bugs:
1. when the moved container was the last child of a container, a new focus would not be found, causing problems if that focus was required later. This was fixed by getting a new focus from the workspace if getting a new focus from the old parent failed;
2. when a container was moved from an inactive workspace, such as by using criteria, focus would be returned to the inactive workspace rather than the active workspace. This fixes that by saving the current before moving the container, then, when restoring the focus, compares the focus with the container being moved. If the container was focused, then it finds the new focus from its parents as before, otherwise, it returns to the old focus. This further necessitated checking if the inactive workspace was then made empty and being destroyed before arranging it.


ianyfan:
Added a small change that fixes misalignment when moving a container due to gaps being present when they are assumed to not be.

PR #2622, opened 2018-09-11
i3bar: count references to blocks

ianyfan:
This prevents blocks from being destroyed before their hotspots are destroyed,
in case it is used for a pending click event that fires between the bar
receiving a new status, which destroys the block, and the bar rendering the new
status, which destroys the hotspot.

This problem can be easily produced by
scrolling on a block that immediately causes a new status to be sent, with
multiple outputs.

Whether this is the best way to fix this problem I'm not sure.

### DIFF
--- a/swaybar/i3bar.c
+++ b/swaybar/i3bar.c
@@ -8,11 +8,10 @@
 #include "swaybar/config.h"
 #include "swaybar/status_line.h"
 
-void i3bar_block_free(struct i3bar_block *block) {
+static void i3bar_block_free(struct i3bar_block *block) {
 	if (!block) {
 		return;
 	}
-	wl_list_remove(&block->link);
 	free(block->full_text);
 	free(block->short_text);
 	free(block->align);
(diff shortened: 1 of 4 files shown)


swaybar/i3bar.c
```diff
@@ -22,10 +21,17 @@ void i3bar_block_free(struct i3bar_block *block) {
 	free(block);
 }
 
+void i3bar_block_drop_ref(struct i3bar_block *block) {
```
SirCmpwn: Can you name this `i3bar_block_unref`?

ianyfan:
Rebased and changed function name

SirCmpwn:
Thanks!

ianyfan:
Just realised that I forgot to check for NULL block in the new function, so may open a new PR to address that. Can it actually be NULL? Also, should I just remove the `i3_bar_block_free` function and move its logic to the unref function?

SirCmpwn:
Yes, you should probably move that logic to unref. But make unref accept a NULL argument without breaking. We model these functions after `free`.

PR #2624, opened 2018-09-11
Update textures for cmd_client_* commands

RedSoxFan:
Since the `client.{focused,focused_inactive,unfocused,urgent}` commands change colors, the textures need to be updated otherwise the textures and the rest of the title bar may utilize different colors.


SirCmpwn:
Thanks!

PR #2628, opened 2018-09-13
Remove bad assertion in workspace_rejigger

RyanDwyer:
The assertion can be (rightfully) triggered by creating layout `V[H[view view] view]` and moving the top right view to the right.

After removing the assertion I found the container being moved needs its size reset to prevent it from being sized wrongly after arranging.


emersion:
Thanks!

PR #2629, opened 2018-09-14
Allow spaces in background file path

### LINKED ISSUE #654: Sway cannot handle spaces in file path to desktop background.
I have a background switching script (chooses a random image file from a directory, and uses swaymsg to change my background), which runs:

`swaymsg "output DP-2 bg \"${PATH}\" fill"`

It works fine, as long as there are no spaces in the filename. If there is a space in the filename, it fails. Here are the relevant messages from the sway log:

```
Handling command 'output DP-2 bg "/home/tajjada/Pictures/Wallpapers/wide/02 - Antelope Canyon Sunrise.jpg" fill'

IPC Client -1 disconnected{ "success": false, "input": "output", "error": "Background file unreadable (\/home\/tajjada\/Pictures\/Wallpapers\/wide\/02)" }

```

As you can see, it chokes on the first space in the path. I even tried wrapping the whole path in quotes (as you can see). Without the quotes, it (obviously) didn't work either.

It would be nice if sway could handle spaces in paths correctly.

RedSoxFan:
Fixes #654 

`wordexp` splits the file path into multiple words if there are spaces. In `master`, only the first word is used. This caused `foo bar` to be truncated to just `foo`. Additionally, the path given to `swaybg` is not escaped in anyway and `wordexp` strips any backslash escaped spaces.

This first commit of this PR joins the words from `wordexp` together and encloses the path given to `swaybg` in quotes. This works well if there are not consecutive spaces in the file path or if consecutive spaces are backslash escaped.

The second commit of this PR will detect non-escaped consecutive spaces and escape them for the user. This allows for the path to be enclosed in quotes.

With both commits, the following methods for supplying the path should be working regardless the number of consecutive spaces:
- The path in quotes
- The path with spaces escaped with backslashes

### DIFF
--- a/sway/commands/output/background.c
+++ b/sway/commands/output/background.c
@@ -61,7 +61,7 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 				"Missing background scaling mode.");
 		}
 
-		wordexp_t p;
+		wordexp_t p = {0};
 		char *src = join_args(argv, j);
 		while (strstr(src, "  ")) {
 			src = realloc(src, strlen(src) + 2);
@@ -123,6 +123,22 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 			}
 			free(src);
 		} else {
+			// Escape spaces and quotes in the final path for swaybg
+			for (size_t i = 0; i < strlen(src); i++) {
+				switch (src[i]) {
+					case ' ':
+					case '\'':
+					case '\"':
+						src = realloc(src, strlen(src) + 2);
+						memmove(src + i + 1, src + i, strlen(src + i) + 1);
+						*(src + i) = '\\';
+						i++;
+						break;
+					default:
+						break;
+				}
+			}
+
 			output->background = src;
 			output->background_option = strdup(mode);
 		}
--- a/sway/commands/output/background.c
+++ b/sway/commands/output/background.c
@@ -63,6 +63,12 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 
 		wordexp_t p;
 		char *src = join_args(argv, j);
+		while (strstr(src, "  ")) {
+			src = realloc(src, strlen(src) + 2);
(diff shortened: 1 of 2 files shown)


sway/commands/output/background.c
```diff
@@ -63,6 +63,12 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 
 		wordexp_t p;
```
ianyfan: I was testing this and you need to initialise this to `(0)`, because if an error occurs, the values do not get initialised and so `wordfree` gets called on junk data, e.g. a segfault occurs if you run `swaymsg 'output * bg \"'`. (Technically outside the scope of this PR but since it's such a simple change I thought it could be thrown in there since it's related anyway)

ianyfan:
I think there are still some quote-escaping issues. Test with files named `".png` and `'.png` and make sure you can get them to work. (really contrived cases, I know, but still...)

SirCmpwn:
Can you do `'"'.png` (a double quote wrapped in single quotes)?

sway/commands/output/background.c
RedSoxFan: Done

RedSoxFan:
> Test with files named `".png` and `'.png` and make sure you can get them to work. (really contrived cases, I know, but still...)

Both are now working. `'.png` can be given as `"'".png` (a single quote wrapped in double quotes) or `\\\'.png` (escaped with 3 backslashes). Similarly `".png` can now be given as `'"'.png` (a double quote wrapped in single quotes) or `\\\".png` (escaped with 3 backslashes).

> Can you do `'"'.png` (a double quote wrapped in single quotes)?

Now you can. Before my last commit, the file path would be parsed, but was not correctly passed to `swaybg`.


Additionally, I changed it so the paths stored have spaces, single quotes, and double quotes escaped with a single backslash. This makes it easier to pass to `swaybg` and works better than enclosing the path in quotes.

SirCmpwn:
Thanks!

PR #2631, opened 2018-09-14
bar: remove i3bar_block_free in favour of i3bar_block_unref

ianyfan:
Ref https://github.com/swaywm/sway/pull/2622#issuecomment-421467458


emersion:
Thanks!

PR #2633, opened 2018-09-15
Remove redundant container creation in view initialisation

ianyfan:
Since the container is already assigned in ~`view_create`~ ~`view_init`~ `view_map`.

Seems to work, hope I'm not missing something.


RyanDwyer:
Uh... where? I searched the project for `view_create` and got no results.

ianyfan:
`view_init`, sorry, mis-remembered the function name.

RyanDwyer:
The container is destroyed in `view_unmap` though, so it should be created in `view_map`. Perhaps we should remove it from `view_init`.

SirCmpwn:
Thanks!

PR #2634, opened 2018-09-15
ipc: add focus information to root node

### LINKED ISSUE #2632: (IPC) Root node contains no focus information
When running `swaymsg -t get_tree` the root node has an empty focus array. This is since the focus array is built from the focus stack, which does not contain outputs.

Possible solutions I have thought of:

1. track outputs on the focus stack (not sure what additional logic is required to make this not break anything)
2. when building the focus array for the root container, use workspaces on the focus stack to determine which outputs should be included, or something (this requires additional logic to make sure that entries are not duplicated)

ianyfan:
Fixes #2632

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -316,9 +316,12 @@ static void focus_inactive_children_iterator(struct sway_node *node,
 	struct focus_inactive_data *data = _data;
 	json_object *focus = data->object;
 	if (data->node == &root->node) {
-		int len = json_object_array_length(focus);
 		struct sway_output *output = node_get_output(node);
+		if (output == NULL) {
+			return;
+		}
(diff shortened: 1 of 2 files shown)


ianyfan:
Thanks for testing. Does this work?

Also: in what cases can `workspace->output` actually be NULL? Except for between creating the workspace and assigning it to an output, I'm supposing.

RyanDwyer:
Yes, confirming it works now.

Workspaces will have a `NULL` output when there are no outputs connected. The workspaces get stashed away in `root->saved_workspaces` until you connect an output. Admittedly I don't think anyone has tested hotplugging down to zero outputs.

SirCmpwn:
Thanks!

PR #2635, opened 2018-09-16
Handle shell special characters in bg file path

RedSoxFan:
See https://github.com/swaywm/sway/issues/654#issuecomment-421604512

This changes it back so the path given to swaybg is enclosed in quotes.

Additionally, the only character that is escaped in the path stored is
double quotes now. This makes it so we don't need to keep an exhaustive
list of characters that need to be escaped.

The end user will still need to escape these characters in their config
or when passed to swaybg.

Examples for file `~/asdf(1).png`
```
output <name> bg ~/asdf\(1\).png <mode>
output <name> bg ~/asdf"(1)".png <mode>
output <name> bg ~/"asdf(1).png" <mode>
```


RedSoxFan:
@FlyingWombat
If you get a chance, can you please test this? I think with this version all characters should be working (although some will have to be escaped).

RedSoxFan:
> So, I'm not sure if I'm missing something, but it's still not working for me.

What I posted above was for if you were putting it in your config. If you are using it with `swaymsg`, you will have to wrap the path in quotes (escaping any double quotes with a backslash) so the shell does not do any expansion on it at that stage.

Ex. `swaymsg output eDP-1 bg "~/asdf\(1\).jpg" fill` or `swaymsg output eDP-1 bg "~/asdf\"(1)\".jpg" fill`

FlyingWombat:
Ok `swaymsg output eDP-1 bg "~/asdf\(1\).jpg" fill` works
And in sway config: 
`output eDP-1 background ~/"asdf(1).jpg" ` works,
but 
`output eDP-1 background "~/asdf(1).jpg"` does not work.

Is there any way we could get `"~/asdf(1).jpg"`, or at least `"/home/user/asdf(1).jpg"` to work?
Otherwise it'll be a lot trickier to set the background with a variable,
as in: `output eDP-1 background $wallpaper fill`

emersion:
`~` isn't expanded by the shell if it's quoted. If `~/"asdf(1).jpg"` works, `"/home/user/asdf(1).jpg"` should work too.

RedSoxFan:
Quotes around entire arguments get stripped before any command is executed, they are only used for grouping. This makes it so the path passed to `wordexp` is `/home/user/asdf(1).jpg`, which will fail due to the attempt to run `1` in a subshell.

If you use the path `"'/home/user/asdf(1).jpg'"`' or `'"/home/user/asdf(1).jpg"'`,  it will work. I'll look into allowing `"\"/home/user/asdf(1).jpg\"` and `'\'/home/user/asdf(1).jpg\''` (and possibly the same without requiring the outer single/double quotes) since I think those should work as well, but I think that's the best it's going to get.

FlyingWombat:
Ok, thanks; I think I understand how quote escaping works in Sway a bit better.
It now works in the way I'd like to use it.
One thing I'd like to clarify on Sway's behavior for quotes
Expressions in commands can be quoted to use values with spaces or special chars
And you said that the quotes are stripped upon parsing of the command
Is `set $<name> <value>` an exception to that?
For example:
When I put the following in my sway config
```
set $img "/home/user/some path/asdf(1).jpg"
output eDP-1 bg '$img' fill
```
it sets the wallpaper properly, and so fits my use case.
But, when I use the following I get these errors:
Using `output eDP-1 bg "$img" fill` gives `Invalid syntax (/home/user/some path/asdf(1).jpg)`
Using `output eDP-1 bg '"$img"' fill` gives `Invalid syntax (""/home/user/some path/asdf(1).jpg"")`

It seems like the quotes are staying in the variable `$img`.
Is that supposed to happen, or should I file a bug report on that?

RedSoxFan:
> Is `set $<name> <value>` an exception to that?

Yes, `set` is the one exception. The only change to the value argument for set is variable substitution still occurs.

FlyingWombat:
Another test:
```output eDP-1 bg '"/home/user/some path/asdf !@#$%^&*()_+-=`~.jpg"' fill```
Gives a syntax error with `'"/path/to.jpg"'`, but fails silently with `"'/path/to.jpg'"`
But that's to be expected, since even zsh needs to escape ! $ \` with \\ when using double quotes.
 
Using: ```output eDP-1 bg '"/home/user/some path/asdf @#%^&*()_+-=~.jpg"' fill``` works.

Using: ``` output eDP-1 bg '"/home/user/some path/asdf !@#\$%^&*()_+-=\`~.jpg"' fill ```
fails silently (note: for some reason sway doesn't need to escape `!`, but zsh does)

As far as I'm concerned, being able to use `@#%^&*()_+-=~` is good enough.
The question is whether we want to support all characters in the path, or just most.

SirCmpwn:
iirc you can escape $ with $$

ianyfan:
Possibly naive question: why is the path handled like this instead of either being passed as-is or using something like `realpath`?

SirCmpwn:
reasons

RedSoxFan:
The following should all now be working: 
`"\"/home/user/asdf(1).jpg\""`

`'\'/home/user/asdf(1).jpg\''`

 ```
 set $img "/home/user/some path/asdf(1).jpg"
 output eDP-1 bg '"$img"' fill
 ```

Additionally, this fixes #2658. 

I think this with a bunch of different paths and I think this is ready for review again.

PR #2637, opened 2018-09-16
Make seat_get_active_child ignore floating children

RyanDwyer:
`seat_get_active_child` is used to get the active tiling child in a few places, such as outputs getting their active workspace and tabbed/stacked containers getting their visible child. When a workspace uses a tabbed or stacked layout and contains a focused floating view, calling `seat_get_active_child` on the workspace would incorrectly return the floating view. This changes it so it will return the tiling child.

This fixes the following bug:

* Create layout `T[view view]` then float one of the views
* Attempt to click the tiling view to give it focus - it wouldn't work because `seat_get_active_child` would return the floating view


emersion:
Maybe it would make sense to rename the function to include "tiling"?

PR #2638, opened 2018-09-16
Fix crash when unmapping last child of a tabbed workspace

### LINKED ISSUE #2636: Crash after closing tabbed layout
To reproduce:
1. Open two windows in a tabbed layout.
2. Close both with the cursor in the titlebar area.

Backtrace: [log.txt](https://github.com/swaywm/sway/files/2386265/log.txt)

This appears to be caused by `children` being empty in `container_at_tabbed`.  I'm not sure if the `length` of `children` should be checked or if `container_at_tabbed` should not be called, though.

RyanDwyer:
* Create layout `T[view view]`
* Move the cursor into the title bar area
* Close both views

Sway would crash because `container_at_tabbed` would attempt to divide by zero when there are no children.

The children check isn't needed for `container_at_stacked` because it doesn't divide anything by the number of children.

Fixes #2636.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -202,6 +202,9 @@ static struct sway_container *container_at_tabbed(struct sway_node *parent,
 	}
 	struct sway_seat *seat = input_manager_current_seat(input_manager);
 	list_t *children = node_get_children(parent);
+	if (!children->length) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #2640, opened 2018-09-16
Overhaul swaybar protocol handling (+fixes)

### LINKED ISSUE #2177: swaybar i3bar protocol json parsing double/all-you-can-free
I don't have the courage to do the rework needed for this right now, but the logic on error is fucked up.

- `i3_handle_readable` calls `status_error` which closes fd/frees stuff but don't remove it from the array of fd it considers so it will double-free on exit (easy to trigger with a status_command like this, any invalid json will do: `status_command sh -c 'echo "{\"version\":1}"; sleep 0.1; echo "{ invalid: json {"'`; then start sway and exit sway, when i3bar exit boom)
- furthermore even within `i3_handle_readable` it's possible to call `i3bar_parse_json` that can also use `status_error` then use `status_error` again (could actually call it as many time as we want with a properly crafted array); then one last time on exit as per previous point for the strawberry.

Anyway, if someone enjoys parsing, please have a look. A quick fix can probably be to just not close fds/free stuff on json error, for all we know the next line could be proper json again, and it'll spare us from this kind of trouble.

(As bonus if someone is motivated, I noticed while trying to reproduce this that we don't handle lines but "one-read-worth" of text at a time, so if something buffers lines and sends two at once like what I wanted to do with my echos without the sleep we don't handle things very well)

(As second bonus, the `if (state->depth > sizeof(state->nodes) / sizeof(state->nodes[0]))` checks are off by one and should be >=; it's like writing `int foo[16]; if (i <= 16) access(foo[i])`)

Meanwhile, it's not like i3status ever throws up bad/over-16-levels-deep json, so we're probably safe :>

ianyfan:
Should fix #2177

This replaces the somewhat dodgy old implementation with something a bit more sensible, I think.

Both protocols now properly handle long and multiple statuses. The old i3bar handling also allowed dodgy JSON to be sent, as long as it starts with a valid array, which this fixes.

I've tried it with many different status commands, but I may have missed something so please test if you have anything unconventional. Might also want to test performance as well, seems fine but I haven't tested it thoroughly.

~One annoying thing is that since everything works in the one buffer, it is not trivial to print received JSON for debugging (since we cannot simply add a null-terminator). Ideas for adding it back in:~
~1. copy the string each time a valid array is encountered, then print it at the end; good if we assume that most of the time, only one status with be sent at a time;~
~2. don't drop the last json object, keep it around in the buffer, then add a null-terminator at the end and print it; good if we assume that most of the time, the buffer won't in fact overflow.~
~Thoughts? I think the second option is probably better.~

In the end, I didn't go with either of the above options, instead, I printed each object as they came (using some buffer shenanigans to avoid copying the string), but printed another message for what actually gets rendered.

Also leave comments about ~function/~ variable names, they may not be the best.

One thing I did realise afterwards is that `read_line_buffer` skips over `\r` but I haven't bothered to do that, not sure if that's wanted.

Note: there is a small issue with the bar immediately exiting due to invalid JSON which I thought I had fixed but still occurs occasionally. I'm also not sure why it exits rather than just display the error message. I think it might have something to do with receiving null bytes from the stream.

### DIFF
--- a/common/readline.c
+++ b/common/readline.c
@@ -70,28 +70,3 @@ char *peek_line(FILE *file, int line_offset, long *position) {
 	fseek(file, pos, SEEK_SET);
 	return line;
 }
-
-char *read_line_buffer(FILE *file, char *string, size_t string_len) {
-	size_t length = 0;
-	if (!string) {
-		return NULL;
-	}
-	while (1) {
-		int c = getc(file);
-		if (c == EOF || c == '\n' || c == '\0') {
-			break;
-		}
-		if (c == '\r') {
-			continue;
-		}
-		string[length++] = c;
-		if (string_len <= length) {
-			return NULL;
-		}
-	}
-	if (length + 1 == string_len) {
-		return NULL;
-	}
-	string[length] = '\0';
-	return string;
-}
--- a/include/swaybar/event_loop.h
+++ b/include/swaybar/event_loop.h
@@ -19,8 +19,8 @@ bool remove_event(int fd);
 bool remove_timer(timer_t timer);
 
 // Blocks and returns after sending callbacks
-void event_loop_poll();
+void event_loop_poll(void);
 
-void init_event_loop();
+void init_event_loop(void);
 
 #endif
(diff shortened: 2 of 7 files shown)


ianyfan:
I'll have a look into it.

ianyfan:
Ready for more review

SirCmpwn:
This is looking pretty good. Is it still WIP? If not I'll start running it as my daily driver.

ianyfan:
Nope, should be good to go.

SirCmpwn:
Works for me, thanks!

Issue #1768, opened 2018-04-07
Backport swaybar tray support


emersion:

mckinney-subgraph:
Hey,

After discussing with @SirCmpwn and @emersion on IRC, I have started working on this. The first phase will be to re-implement the SNI dbus protocol using the *sd-bus* API. I'll add this as a first item to the task list.

I am also still reading through the tray code and looking at the efforts already undertaken so I will update this task list as I progress. If other people have some tasks in mind, please add them to the issue.

- [ ] Implement SNI dbus protocol using sd-bus

SirCmpwn:
cc @4e554c4c

skrps:
Is this being worked on?

I have a (not naïve, I hope) question: is it *really* necessary to
implement tray support using sd-bus API?  AFAIK, it is not present in
non-SystemD distros and could be a problem for those who want this
feature.

There is some project similar to eudev or elogind, but for sd-bus, that
I am not aware of?

SirCmpwn:
This is not being worked on, feel free to adopt it. elogind ships with sd-bus.

SirCmpwn:
Also, for the record, tray support should be compile-time optional.

mckinney-subgraph:
Actually, I will submit a pull request for "minimal viable SNI" at some point during this week.

mckinney-subgraph:
Err yeah, sorry :(. I had the "minimal viable SNI" patch in the works but ran into troubles getting it working with the event loop code and then my free time evaporated. I am happy to continue to working on this when I have some spare time but can't give an ETA.

SirCmpwn:
Can you post the WIP patch so someone else doesn't have to start from scratch?

mckinney-subgraph:
@SirCmpwn Sure, can do.

SirCmpwn:
Thanks!

4e554c4c:
I can also work on this if sd-bus isn't a requirement

SirCmpwn:
I don't think the old code is very maintainable. I know it's my fault for pushing back against sd-bus before. I'd rather not port it as-is, I think it'd be better to refactor it.

SirCmpwn:
I think I would prefer to continue to have broad support for what the ecosystem is actually doing. I think KDE is in the wrong here, it's better to improve the canonical protocol than to encourage the proliferation of proprietary protocols.

SirCmpwn:
[...]

PR #2645, opened 2018-09-16
[WIP] Added minimum viable swaybar tray SNI protocol support using sd-bus

mckinney-subgraph:
Sending this WIP patch to start off the implementation of issue #1768. This just implements `NewIcon` without doing anything else (looking up the icon, etc.). 

This PR includes a Meson build option to enable tray support. It is set to `false` under the assumption that tray support would not be enabled by default. To test this PR, it needs to be set to `true`.

I worked out one of the event loop problems I was having (see #1768 discussion) -- namely that the service would segfault when an client stopped sending it stuff. 

There are some notes (BUG comment in `tray/dbus.c`) about my other problem. The gist of it is that I was having problems attaching sd-bus slots to the SNI watcher structure. My aim was to setup multiple watchers for each protocol (freedesktop, KDE, etc.). The problem I encountered was that userdata was not being passed properly to SNI watcher handlers (such as handle_new_icon) when the slot was a field of the SNI watcher struct itself.

### DIFF
--- a/swaybar/tray/dbus.c
+++ b/swaybar/tray/dbus.c
@@ -1,0 +1,131 @@
+#define _XOPEN_SOURCE 500
+
+#ifdef __FreeBSD__
+#include <dev/evdev/input-event-codes.h>
+#else
+#include <linux/input-event-codes.h>
+#endif
+#include <poll.h>
+#include <time.h>
+#include <wlr/util/log.h>
+#include "swaybar/event_loop.h"
+#include "swaybar/tray/dbus.h"
+#include "swaybar/tray/sni_host.h"
+#include "swaybar/tray/sni_watcher.h"
+
+static const char *fd_iface = "org.freedesktop.StatusNotifierWatcher";
+static const char *fd_notifier_host = "org.freedesktop.StatusNotifierHost";
+
+bool dbus_init() {
+	int ret = 0;
+	sd_bus *bus = NULL;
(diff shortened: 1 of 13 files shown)


mckinney-subgraph:
A simple tool I wrote to test this code:
https://github.com/mckinney-subgraph/snit

swaybar/tray/dbus.c
```diff
@@ -0,0 +1,131 @@
...
+	int ret = 0;
+	sd_bus *bus = NULL;
+	/* BUG: *slot should contained in sni_watcher struct to facilitate multiple
+	 * watchers but when done in this way the sd-bus handler callbacks 
+	 * (handle_new_icon) were getting bad userdata
+	 */
+
+	sd_bus_slot *slot = NULL;
+	// TODO: Add other watchers for KDE, etc.
+	struct sni_watcher fd_sni_watcher = {0};
```
ianyfan: Your problem is (probably) not about the slot being a field of the watcher struct, it's that you've declared the watcher on the stack, meaning that when the function returns, it goes out of scope, causing its memory to become corrupted when it is used in the callback later.

The solution is to `calloc` it on the heap.

mckinney-subgraph:
Closing this PR due to the fantastic effort by @ianyfan in PR https://github.com/swaywm/sway/pull/3249

PR #2660, opened 2018-09-18
Fix double iteration of scratchpad containers

### LINKED ISSUE #2655: [Alpha 1.0] Scratchpad show command doesn't hide again if used with selector
If I use the following command:
`bindsym $mod+g [instance=$scratchterm] scratchpad show`
The scratchpad terminal is shown, but not hidden again.
But if I use that command:
`bindsym $mod+g scratchpad show`
it is hidden again.

RyanDwyer:
`root_for_each_container` and `root_find_container` were using incorrect logic to determine if a container was hidden in the scratchpad. Containers will have a `NULL` parent if they are a direct child of a workspace. Containers will have a `NULL` workspace if they are hidden in the scratchpad.

The incorrect check meant that `root_for_each_container` would run the callback on scratchpad containers twice. This meant that executing a command such as `[class="$something"] scratchpad show` would cause the command to run twice, resulting in the container being shown and hidden again which is effectively a no op.

Fixes #2655.

### DIFF
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -265,10 +265,10 @@ void root_for_each_container(void (*f)(struct sway_container *con, void *data),
 	// Scratchpad
 	for (int i = 0; i < root->scratchpad->length; ++i) {
 		struct sway_container *container = root->scratchpad->items[i];
-		// If the container has a parent then it's visible on a workspace
+		// If the container has a workspace then it's visible on a workspace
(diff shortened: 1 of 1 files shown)


emersion:
@nycex, can you confirm it fixes your issue?

nycex:
Yeah, I can confirm it fixes my issue.

emersion:
Thanks!

PR #2666, opened 2018-09-19
swaybar: handle hotplugging

emersion:
Don't kill and respawn swaybars on hotplug.


RedSoxFan:
This works great when the bar config is for all outputs, but if I specify an output name, swaybar is not appearing on any output

RedSoxFan:
If I set swaybar to show on an output, unplug that output, and plug it back in, swaybar does not show on the output until I reload the config

emersion:
>If I set swaybar to show on an output, unplug that output, and plug it back in, swaybar does not show on the output until I reload the config

I couldn't reproduce this at first, because it doesn't happen with the Wayland backend. However it does happen with the DRM backend. The bug was a lack of `wl_display_flush` in the event loop.

This was the source of numerous bugs, from hotplug events not being
received to segfaults because `wl_display_roundtrip` was making the
bar process unplug events while blocking in an iteration over all
outputs.

The last commit fixes that.

RedSoxFan:
Thanks!

PR #2669, opened 2018-09-19
Introduce create_output command (for developer use)

RyanDwyer:
Should help with testing hotplugging.

Depends on https://github.com/swaywm/wlroots/pull/1253.


koxu1996:
@RyanDwyer can it be used to make split screen, eg. two virtual screens? If yes, could you share example usage of `create_output`?

RyanDwyer:
No. The command can only be used when sway is running within another compositor (or another instance of itself). It pretends that you've just connected an output, so it creates a new window in the outer DRM session.

koxu1996:
Ok, thank you.

PR #2670, opened 2018-09-19
config: free strings fields when freeing input config

ianyfan:
Fixes memory leaks, since they are duplicated from the command to the config, then duplicated again when merged, but only freed once.


SirCmpwn:
Thanks!

PR #2672, opened 2018-09-19
swaybar: fix parsing errors when json is sent in parts

### LINKED ISSUE #2664: swaybar: failed to parse i3bar json
I'm using sway with https://github.com/greshake/i3status-rust.
I started seeing this error after upgrading to cdce604d8e166ae905b394306fece4e7f035cedd. The problem arises here: https://github.com/swaywm/sway/blob/3600b3db577298bc553c394a4b09e71163f76561/common/pango.c#L131  
`args` contains not one but multiple outputs of i3status-rust ([example](https://github.com/swaywm/sway/files/2395837/swaybar.txt)) and its length is indeed >= 2048 - is this an expected behavior?

ianyfan:
Fixes #2664 @cnt0 @Ogromny please check

This issue occurs when there is still input to be read, which did not arise when I was testing it before.

### DIFF
--- a/swaybar/i3bar.c
+++ b/swaybar/i3bar.c
@@ -204,6 +204,7 @@ bool i3bar_handle_readable(struct status_line *status) {
 					status->buffer_index -= buffer_pos;
 					memmove(status->buffer, &status->buffer[buffer_pos],
 							status->buffer_index);
+					buffer_pos = 0;
 				} else {
 					// expand buffer
 					status->buffer_size *= 2;
--- a/swaybar/i3bar.c
+++ b/swaybar/i3bar.c
@@ -117,7 +117,9 @@ bool i3bar_handle_readable(struct status_line *status) {
 				memmove(status->buffer, &status->buffer[c], status->buffer_index);
 				break;
 			} else if (!isspace(status->buffer[c])) {
-				status_error(status, "[invalid json]");
+				wlr_log(WLR_DEBUG, "Invalid i3bar json: expected '[' but encountered '%c'",
+						status->buffer[c]);
+				status_error(status, "[invalid i3bar json]");
 				return true;
 			}
 		}
@@ -155,6 +157,8 @@ bool i3bar_handle_readable(struct status_line *status) {
 					++buffer_pos;
 					break;
 				} else if (!isspace(status->buffer[buffer_pos])) {
+					wlr_log(WLR_DEBUG, "Invalid i3bar json: expected ',' but encountered '%c'",
+							status->buffer[buffer_pos]);
 					status_error(status, "[invalid i3bar json]");
 					return true;
 				}
@@ -166,7 +170,8 @@ bool i3bar_handle_readable(struct status_line *status) {
 		} else {
 			test_object = json_tokener_parse_ex(status->tokener,
 					&status->buffer[buffer_pos], status->buffer_index - buffer_pos);
-			if (json_tokener_get_error(status->tokener) == json_tokener_success) {
+			enum json_tokener_error err = json_tokener_get_error(status->tokener);
+			if (err == json_tokener_success) {
 				if (json_object_get_type(test_object) == json_type_array) {
 					if (last_object) {
 						json_object_put(last_object);
@@ -198,7 +203,7 @@ bool i3bar_handle_readable(struct status_line *status) {
 					continue; // look for comma without reading more input
 				}
 				buffer_pos = status->buffer_index = 0;
-			} else if (json_tokener_get_error(status->tokener) == json_tokener_continue) {
+			} else if (err == json_tokener_continue) {
 				json_tokener_reset(status->tokener);
 				if (status->buffer_index < status->buffer_size) {
 					// move the object to the start of the buffer
@@ -219,6 +224,10 @@ bool i3bar_handle_readable(struct status_line *status) {
 					}
 				}
 			} else {
+				char last_char = status->buffer[status->buffer_index - 1];
+				status->buffer[status->buffer_index - 1] = '\0';
+				wlr_log(WLR_DEBUG, "Failed to parse i3bar json - %s: '%s%c'",
+						json_tokener_error_desc(err), &status->buffer[buffer_pos], last_char);
 				status_error(status, "[failed to parse i3bar json]");
 				return true;
 			}
--- a/swaybar/i3bar.c
+++ b/swaybar/i3bar.c
@@ -199,6 +199,7 @@ bool i3bar_handle_readable(struct status_line *status) {
 				}
 				buffer_pos = status->buffer_index = 0;
 			} else if (json_tokener_get_error(status->tokener) == json_tokener_continue) {
+				json_tokener_reset(status->tokener);
 				if (status->buffer_index < status->buffer_size) {
 					// move the object to the start of the buffer
 					status->buffer_index -= buffer_pos;


Ogromny:
Now I get "[invalid json]"... :/

ianyfan:
That occurs when looking for the opening bracket of the infinite array, and it encounters a character that is not either `[` or whitespace. Could you share the output of your status command up to the first status?

Ogromny:
```sh
i3status-rs ~/.config/sway/i3status-rs.toml 
```
```json
{"version": 1, "click_events": true}
[[{"background":null,"color":"#424242","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#424242","color":"#ffffff","full_text":" i3status-rs  /home/ogromny ","separator":false,"separator_block_width":0},{"background":"#424242","color":"#535353FF","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#535353FF","color":"#FFFFFFFF","full_text":"   ","separator":false,"separator_block_width":0},{"background":"#535353FF","color":"#424242","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#424242","color":"#ffffff","full_text":"   ","name":"pacman","separator":false,"separator_block_width":0},{"background":"#424242","color":"#535353FF","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#535353FF","color":"#FFFFFFFF","full_text":"   ","separator":false,"separator_block_width":0},{"background":"#000000","color":"#000000","full_text":"","separator":false,"separator_block_width":0},{"background":"#000000","color":"#000000","full_text":"","separator":false,"separator_block_width":0},{"background":"#000000","color":"#424242","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#424242","color":"#ffffff","full_text":"   ","name":"e55c84fc365f46a0b6e950b72142b14f","separator":false,"separator_block_width":0},{"background":"#424242","color":"#000000","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#000000","full_text":"","separator":false,"separator_block_width":0},{"background":"#000000","color":"#424242","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#424242","color":"#ffffff","full_text":"   ","name":"661d735f70534940bf627998681122bf","separator":false,"separator_block_width":0},{"background":"#424242","color":"#000000","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#000000","full_text":"","separator":false,"separator_block_width":0},{"background":"#000000","color":"#424242","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#424242","color":"#ffffff","full_text":"   ","name":"0823e702b7a84e9fb6ae22b395469115","separator":false,"separator_block_width":0}],
Temperature (-127) outside of range ([-100, 150])
```

Ogromny:
icons = "awesome"

[theme]
name = "slick"

[theme.overrides]
separator = ""

[[block]]
block = "focused_window"
max_width = 72

[[block]]
block = "uptime"

[[block]]
block = "pacman"
interval = 30

[[block]]
block = "net"
device = "wlp3s0"
ip = true
ssid = true
speed_up = false
speed_down = false

[[block]]
block = "sound"
step_width = 0

[[block]]
block = "backlight"

[[block]]
block = "temperature"
collapsed = false
interval = 1
format = "{max}°C"

[[block]]
block = "battery"
show = "both"

[[block]]
block = "time"
format = "%d/%m/%Y %H:%M:%S"
interval = 1

RyanDwyer:
bar {
        ...
        status_command ~/.config/sway/status
    }

~/.config/sway/status:

    #!/bin/bash
    
    echo '{"version":1}'
    echo '[[],'
    i=0
    
    while true; do
        #echo "[{ \"full_text\" : \"test $i\" }],"
    
        echo '['
        echo "{ \"full_text\" : \"test $i\" }"
        echo '],'
    
        i=$(($i+1))
        sleep 0.5
    done

This usually fails within a few seconds. If I uncomment the first and comment the second then I can't make it fail. The only difference with those is line breaks.

sghctoma:
Hi,

I have started investigating another issue regarding the new swaybar JSON processing that occurs on FreeBSD, and while I don't see a clear connection to this issue, here is what I have found out so far, maybe it will help investigating this:

First, `getdelim(..., EOF, ...)` won't work on FreeBSD, because `EOF` is never reached (at least not until the bar exits), and `getdelim` on FreeBSD returns `-1` when the delimiter is not found.

On Linux it behaves a bit oddly in such a case, which might have something to do with these issues. It reads the whole content available, returns the size of it, but also sets the error indicator of the stream, and `errno` to `EAGAIN`, so subsequent reads from the stream will fail until the error indicator is reset.

POSIX does not say anything about what should happen when the delimiter is not found, so both behaviours are acceptable I guess, but setting `errno` and the error indicator while succeeding feels dodgy to me.  That said, maybe using `clearerr` here too (just like after calling `getline`) would improve things?

ianyfan:
@sghctoma From the POSIX documentation:

"The characters read, including any delimiter, shall be stored in the object, and a terminating NUL added when the delimiter or end-of-file is encountered ... getline() and getdelim() functions shall return the number of bytes written into the buffer, including the delimiter character if one was encountered before EOF ... if no characters were read and the stream is at end-of-file, the end-of-file indicator for the stream shall be set and the function shall return -1"

This does not explicitly specify the case that the delimiter is EOF, but seems to indicate that it is valid to encounter EOF before the delimiter, and that it should return what it has read so far. If FreeBSD does not do that, then I would argue that it is not POSIX-compliant.

I didn't add `clearerr` because it doesn't read from the file anymore, it reads straight from the stream. I'm not sure if adding it would fix anything.

sghctoma:
@ianyfan

> This does not explicitly specify the case that the delimiter is EOF, but seems to indicate that it is valid to encounter EOF before the delimiter, and that it should return what it has read so far. If FreeBSD does not do that, then I would argue that it is not POSIX-compliant.

This is all true, but `getdelim` won't encounter `EOF` here, because the `stdout` of the status program is a continous stream, and will contain `EOF` only when it exits. In other words, "`getdelim` not finding the delimiter" is the problem, not "`EOF` being the delimiter".

> I didn't add `clearerr` because it doesn't read from the file anymore, it reads straight from the stream. I'm not sure if adding it would fix anything.

Fair enough, I missed that.

ianyfan:
I see. So you're saying that where Linux treats the end of a stream as an EOF, FreeBSD treats it as an error, and only treats it as an EOF when the stream actually terminates?

If that is the case, then it feels like the only solution is to `fread` large chunks until no more can be read, reallocing the buffer as needed, but again I'm not sure if FreeBSD treats that any differently.

Honestly, I've been meaning to try out FreeBSD myself but I keep putting it off...

ianyfan:
@RyanDwyer I've run that several times and it hasn't failed :/

I've added a commit to help debug this, could you apply that and see if it gives any useful information?

sghctoma:
@ianyfan 

> I see. So you're saying that where Linux treats the end of a stream as an EOF, FreeBSD treats it as an error, and only treats it as an EOF when the stream actually terminates?

Yep, something like that, but not exactly, since the EOF indicator is not set on Linux either.

> If that is the case, then it feels like the only solution is to fread large chunks until no more can be read, reallocing the buffer as needed, but again I'm not sure if FreeBSD treats that any differently.

I have some free time at the weekend, will look into this a bit more, and I'll get back to you. Maybe in another issue/PR, did not want to hijack this one in the first place.

> Honestly, I've been meaning to try out FreeBSD myself but I keep putting it off...

I've tried it out ~8 years ago, and stuck wih it as my main desktop OS :)

RyanDwyer:
@ianyfan

    2018-09-21 23:32:20 - [swaybar/i3bar.c:191] Received i3bar json: '[
    { "full_text" : "test 54" }
    ]'
    2018-09-21 23:32:20 - [swaybar/i3bar.c:244] Rendering last received json
    2018-09-21 23:32:21 - [swaybar/i3bar.c:191] Received i3bar json: '[
    { "full_text" : "test 55" }
    ]'
    2018-09-21 23:32:21 - [swaybar/i3bar.c:244] Rendering last received json
    2018-09-21 23:32:21 - [swaybar/i3bar.c:224] Failed to parse i3bar json: '
    [
    { "full_text" : "test 56" }
    ],
    '

I also replaced the `status_error` line with this:

    status_error(status, json_tokener_error_desc(json_tokener_get_error(status->tokener)));

...and when it fails, the error being returned is `array value separator ',' expected`.

ianyfan:
I think the problem is that I have misunderstood `json_tokener_continue`. Will push a fix shortly.

ianyfan:
@RyanDwyer Check now. I can instantly reproduce the error by adding some sleeps between each line.

I should have tested with split json earlier, and caught this bug much earlier, my apologies.

RyanDwyer:
Yep, that seems to have worked. I can't reproduce it any more.

ianyfan:
@cnt0 @Ogromny see if it is fixed now

@sghctoma please get back to me on this and we can open another PR

cnt0:
Yes, I can't reproduce this issue anymore.  

There's another one (onclick events cause crash in i3status-rust and "error reading from status command" in swaybar), but I have to check if it is related to swaybar; I will create a separate issue if needed

SirCmpwn:
Thanks!

Issue #1685, opened 2018-03-31
Sway clients should utilize subpixel geometry for font rendering


SirCmpwn:
Can communicate this to cairo

emersion:
See `cairo_font_options_set_subpixel_order`

ianyfan:
Closed by https://github.com/swaywm/sway/pull/2677

PR #2677, opened 2018-09-21
Render fonts slightly nicer 

ggreer:
Set cairo font options to render text and lines with subpixel hinting (if available).

I've been running sway with this patch for several months and I prefer it.

### DIFF
--- a/common/cairo.c
+++ b/common/cairo.c
@@ -13,6 +13,22 @@ void cairo_set_source_u32(cairo_t *cairo, uint32_t color) {
 			(color >> (0*8) & 0xFF) / 255.0);
 }
 
+cairo_subpixel_order_t to_cairo_subpixel_order(enum wl_output_subpixel subpixel) {
+	switch (subpixel) {
+	case WL_OUTPUT_SUBPIXEL_HORIZONTAL_RGB:
+		return CAIRO_SUBPIXEL_ORDER_RGB;
+	case WL_OUTPUT_SUBPIXEL_HORIZONTAL_BGR:
+		return CAIRO_SUBPIXEL_ORDER_BGR;
+	case WL_OUTPUT_SUBPIXEL_VERTICAL_RGB:
+		return CAIRO_SUBPIXEL_ORDER_VRGB;
+	case WL_OUTPUT_SUBPIXEL_VERTICAL_BGR:
+		return CAIRO_SUBPIXEL_ORDER_VBGR;
+	default:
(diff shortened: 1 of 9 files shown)


SirCmpwn:
You need to actually provide the hints, too. Check the wlr_output's subpixel geometry.

ggreer:
Ah, yeah. The actual patch I was using had an extra line setting subpixel order to `CAIRO_SUBPIXEL_ORDER_RGB`, since that's what my monitors use.

emersion:
>Ah, yeah. The actual patch I was using had an extra line setting subpixel order to CAIRO_SUBPIXEL_ORDER_RGB, since that's what my monitors use.

Please extract this piece of information from [`wl_output::geometry`](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output).

ianyfan:
Ref https://github.com/swaywm/sway/issues/1685?

ggreer:
Before (antialiased) antialiased
After (subpixel rendering) subpixel

SirCmpwn:
Nice.

common/cairo.c
```diff
@@ -13,6 +13,22 @@ void cairo_set_source_u32(cairo_t *cairo, uint32_t color) {
...
+		case WL_OUTPUT_SUBPIXEL_HORIZONTAL_RGB:
+			return CAIRO_SUBPIXEL_ORDER_RGB;
+		case WL_OUTPUT_SUBPIXEL_HORIZONTAL_BGR:
+			return CAIRO_SUBPIXEL_ORDER_BGR;
+		case WL_OUTPUT_SUBPIXEL_VERTICAL_RGB:
+			return CAIRO_SUBPIXEL_ORDER_VRGB;
+		case WL_OUTPUT_SUBPIXEL_VERTICAL_BGR:
+			return CAIRO_SUBPIXEL_ORDER_VBGR;
+		default:
+			return CAIRO_SUBPIXEL_ORDER_DEFAULT;
```
SirCmpwn: style nit: align case with switch

SirCmpwn:
LGTM aside from one minor style issue

common/cairo.c
ggreer: Fixed. I wonder if there's a clang-format rule for that.

SirCmpwn: We've looked into clang-format before, unfortunately there are a few issues with our style that it doesn't support.

SirCmpwn:
Thanks!

PR #2678, opened 2018-09-21
Reconfigure xwayland views when repositioned

### LINKED ISSUE #2673: X11 windows are not configured when workspace is moved
1. Have two outputs
2. Open Firefox on the right one
3. `move workspace to output left`
4. Type something in the URL bar
5. The popup shows up on the right output

To "fix" it force Firefox to be re-configured, e.g. open another client on the same workspace.

RyanDwyer:
Fixes #2673.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -6,6 +6,7 @@
 #include <string.h>
 #include <time.h>
 #include <wlr/types/wlr_buffer.h>
+#include "sway/config.h"
 #include "sway/debug.h"
 #include "sway/desktop.h"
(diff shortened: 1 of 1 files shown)


sway/desktop/transaction.c
```diff
@@ -390,6 +390,16 @@ static bool should_configure(struct sway_node *node,
 	}
 	struct sway_container_state *cstate = &node->sway_container->current;
 	struct sway_container_state *istate = instruction->container_state;
+#ifdef HAVE_XWAYLAND
```
emersion: You should include `config.h` if you use this macro

emersion:
Thanks!

PR #2680, opened 2018-09-21
Fix compilation against wlroots without X11 backend

marienz:
Tested:
- built sway against wlroots built with -Dx11-backend=enabled, ran WLR_BACKENDS=x11 sway, confirmed create_output still works
- built sway against wlroots built with -Dx11-backend=disabled, ran it (wayland backend), confirmed create_output still works

The error message still mentions "X11 backend" even if built without support for one, hopefully that's ok for a developer command.


emersion:
Thanks!

PR #2686, opened 2018-09-22
Implement tab cycling using mouse wheel

### LINKED ISSUE #2684: Scrolling on the title bar of a window in tabbed mode doesn't move focus
When I mouse over the title bar of a window that is in tabbed mode in i3 and scoll up or down, the focus moves to the previous / next window in the tab container – basically the same as scrolling through the workspaces when the mouse is over the bar. This doesn't work in sway (version 1.0-alpha.5-364-g15dadaaa).

RyanDwyer:
Firstly, a change had to be made to the `container_at` functions. If you create layout `T[view H[view view]]` and hover the second tab, the `container_at` functions would return the `focus_inactive` child. They now return the split container itself. To compensate for this, `dispatch_cursor_button` has been adjusted to find the `focus_inactive` child before focusing it.

The actual implementation of wheel scrolling is pretty straightforward. This uses `handle_cursor_axis`, so I took a similar approach to `handle_cursor_button` (ie. creating a `dispatch_cursor_axis` function).

To test, create a tabbed or stacked layout, hover the title bar and scroll with the wheel.

Closes #2684.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -911,9 +911,10 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 		return;
 	}
 
-	// Handle clicking a container surface
+	// Handle clicking a container surface or decorations
 	if (cont) {
-		seat_set_focus_container(seat, cont);
+		node = seat_get_focus_inactive(seat, &cont->node);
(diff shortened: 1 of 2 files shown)


emersion:
Thanks!

PR #2687, opened 2018-09-22
swaybar: explicitly check return value of getdelim

ianyfan:
This prevents an signed-to-unsigned conversion error on buffer_index if
getdelim fails and returns -1, which caused swaybar to try to search the
header for the array and immediately failing


rubdos:
Tested and working. Solves my issue of running [`i3status-rust`](https://github.com/greshake/i3status-rust/) under wlroots.

RyanDwyer:
Thanks!

PR #2688, opened 2018-09-22
Allow running commands on containers without focusing them

### LINKED ISSUE #2656: [Alpha 1.0] Sway crash when moving to scratchpad in config
Sway crashes, if you move something directly from the config to the scratchpad. here is the command:
`for_window [instance=$scratchterm] move scratchpad`

RyanDwyer:
This adds a `con` argument to `execute_command` which allows you to specify the container to execute the command on. In most cases it leaves it as `NULL` which makes it use the focused node. We only set it when executing `for_window` criteria such as when a view maps. This means we don't send unnecessary IPC focus events, and fixes a crash when the criteria command is `move scratchpad` (because we can't give focus to a hidden scratchpad container).

Each of the shell map handlers now check to see if the view has a workspace. It won't have a workspace if criteria has moved it to the scratchpad.

To test:

* `for_window [class="Firefox"] move scratchpad` and then launch Firefox
* Try other criteria commands

Fixes #2656.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -391,8 +391,6 @@ static bool view_has_executed_criteria(struct sway_view *view,
 }
 
 void view_execute_criteria(struct sway_view *view) {
-	struct sway_seat *seat = input_manager_current_seat(input_manager);
-	struct sway_node *prior_focus = seat_get_focus(seat);
 	list_t *criterias = criteria_for_view(view, CT_COMMAND);
 	for (int i = 0; i < criterias->length; i++) {
 		struct criteria *criteria = criterias->items[i];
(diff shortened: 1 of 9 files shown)


sway/tree/view.c
```diff
@@ -405,14 +404,11 @@ void view_execute_criteria(struct sway_view *view) {
 				criteria->raw, view, criteria->cmdlist);
 		seat_set_focus_container(seat, view->container);
 		list_add(view->executed_criteria, criteria);
-		struct cmd_results *res = execute_command(criteria->cmdlist, NULL);
-		if (res->status != CMD_SUCCESS) {
-			wlr_log(WLR_ERROR, "Command '%s' failed: %s", res->input, res->error);
-		}
+		struct cmd_results *res = execute_command(
```
emersion: Maybe `seat_set_focus_container` on line 405 can be removed now?

RyanDwyer: Ah, I missed that. Removed now.

PR #2690, opened 2018-09-22
commands: remove obselete code for sticky windows when switching workspace

ianyfan:
Fixes crash when switching workspace when the only window present is the sticky window, since it tries to destroy the old workspace twice. IPC events seem fine.

I think the bug was introduced when typesafety was.


RyanDwyer:
Thanks!

PR #2691, opened 2018-09-22
Call pango_cairo_context_set_font_options().

ggreer:
Call pango_cairo_context_set_font_options() before pango_cairo_update_layout() and pango_cairo_show_layout(). It looks like Pango defaults to merging the Cario font options with its own defaults, and the end result is no full hinting.

Comparison: comparison
Master is on top and this PR is on bottom.

Notice the difference in smaller characters such as `°` and `%`. Also the top left of the `?` has a little more down curl (more accurate question mark shape). And the text is a little higher contrast.

The improvements are minor, but they're improvements.

You can see another example of pango+cairo usage in Chromium: https://chromium.googlesource.com/chromium/chromium/+/master/ui/gfx/pango_util.cc#197

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -6,6 +6,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include "cairo.h"
 #include "log.h"
 #include "stringop.h"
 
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #2693, opened 2018-09-23
Move sticky containers when switching workspace via criteria

RyanDwyer:
* Create a view on workspace 1
* Switch to workspace 2 (on the same output) and create a floating sticky view
* Use criteria to focus the view on workspace 1

In master, the sticky container remains on workspace 2 and is not rendered. In this PR it is relocated to workspace 1. This is because the code that moves the sticky containers was in `workspace_switch` which is not called when using the above method. This patch relocates the sticky-moving code into `seat_set_focus_warp`.

A side effect of this patch is that if you have a sticky container focused and then switch workspaces, the sticky container will no longer be focused. It would previously retain focus. I think unfocusing is desirable though, as the user's intention is to focus whatever's on the workspace.

In `seat_set_focus_warp`, `new_output_last_ws` was only set when changing outputs, but now it's always set. This means `new_output_last_ws` and `last_workspace` might point to the same workspace, which means we have to make sure we don't destroy it twice. It now checks to make sure they're different, and to make this more obvious I've moved both calls to `workspace_consider_destroy` to be next to each other.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -667,10 +667,8 @@ void seat_set_focus_warp(struct sway_seat *seat, struct sway_node *node,
 	}
 
 	// find new output's old workspace, which might have to be removed if empty
-	struct sway_workspace *new_output_last_ws = NULL;
-	if (new_output && last_output != new_output) {
-		new_output_last_ws = output_get_active_workspace(new_output);
-	}
+	struct sway_workspace *new_output_last_ws =
+		new_output ? output_get_active_workspace(new_output) : NULL;
 
 	// Unfocus the previous focus
 	if (last_focus) {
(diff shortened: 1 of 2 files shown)


SirCmpwn:
>This patch relocates the sticky-moving code into seat_set_focus_warp.

This scares the shit out of me

RyanDwyer:
Bump.

sway/input/seat.c
```diff
@@ -754,11 +761,14 @@ void seat_set_focus_warp(struct sway_seat *seat, struct sway_node *node,
...
 	}
 
-	if (last_focus) {
-		if (last_workspace) {
-			workspace_consider_destroy(last_workspace);
-		}
+	if (new_output_last_ws) {
+		workspace_consider_destroy(new_output_last_ws);
+	}
+	if (last_workspace && last_workspace != new_output_last_ws) {
```
emersion: Won't this comparison cause a use-after-free if `new_output_last_ws` was destroyed?

RyanDwyer: No, because `workspace_consider_destroy` does not free the workspace immediately. It is marked as destroying and the next transaction frees it.

emersion: Hmm. Not a fan of this (we could in the future for instance destroy immediately when the transaction has nothing to wait for?). I guess since it's just pointer comparison and we're not actually dereferencing pointers it's fine.

emersion:
Thanks!

PR #2694, opened 2018-09-23
Prevent sticky containers from jumping on workspace switch

RyanDwyer:
If you have swaybar docked to the top, and you create a floating sticky container and switch workspaces on the same output, the sticky container would move down by the height of swaybar on each switch.

This happens because when creating the workspace we set the dimensions to the same as the output, then the subsequent arrange corrects it. During this arrange, floating containers are translated so they stay relative to the workspace. This translation needs to not occur for the initial arrange.

This patch makes workspaces have a zero width and height when first created, so we can detect whether this is the initial arrange and avoid translating the floating containers if so.


emersion:
Thanks!

PR #2696, opened 2018-09-23
Fix GtkMenuBar always opening first item

### LINKED ISSUE #2299: When clicking the menubar of a gtk3 app, the first item always opens
1. Open lowriter
2. Click "Format" on the menubar along the top
3. "File" opens instead

RyanDwyer:
It seems like advertising that we support touch when we don't, while using SSD, makes GtkMenuBar misbehave. Please don't ask me why...

Fixes #2299.


SirCmpwn:
Thanks! This was a tough one.

Issue #2429, opened 2018-08-06
Add CSD to border modes


SirCmpwn:
When cycling through border styles for floating views, CSD should be an option. Accordingly, cycling through SSD border modes should inform the client that SSDs are in use (and maybe tile it on all four sides if the client doesn't support xdg-decoration?).

PR #2697, opened 2018-09-23
[WIP] Add CSD to border modes

RyanDwyer:
I need some help with this one. Currently implementing #2429 but the client doesn't seem to respect the new decoration mode. See `view_set_csd`. For example, run `border toggle` several times on an `xdg_shell` view.

Depends on https://github.com/swaywm/wlroots/pull/1259

### DIFF
--- a/include/sway/decoration.h
+++ b/include/sway/decoration.h
@@ -5,6 +5,7 @@
 
 struct sway_server_decoration {
 	struct wlr_server_decoration *wlr_server_decoration;
+	struct sway_view *view;
 	struct wl_list link;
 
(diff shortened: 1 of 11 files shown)


emersion:
I think GTK+ honors the server preference but doesn't support dynamically changing the decoration mode. Also note that with the KDE protocol the client decides the mode anyway.

I'd advise implementing xdg-decoration instead (you can test it with winit or Qt master, there's also a wlroots example and I think GLFW support is in progress). It works differently so it will require some thought to glue it with the current KDE protocol implementation.

Issue #1568, opened 2018-01-16
multiple seat: only make cursor appear for seat if it has a pointer configured


martinetd:
For example on a setup where you occasionally plug laptop into an extra pointer+keyboard that you'd like to see in another seat -- most of the time there's no device in 2nd seat, so it doesn't make sense to have cursor laying around at that time (and now I'm trying that, it's pretty annoying)

Could also remove cursor when pointer is unplugged, if we can tell no device left with pointer capability.

emersion:
Related: if the seat only contains a tablet, hide the cursor when the pen leaves the tablet.

PR #2698, opened 2018-09-23
Only show cursor if pointer configured

ianyfan:
Fixes https://github.com/swaywm/sway/issues/1568

The implementation works, idk if it's any good though. Also haven't dealt with tablet pads because I don't have one.

Possible FR: some sort of seat sub-command `show_cursor yes|no|auto`

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -1006,8 +1006,7 @@ static void handle_touch_down(struct wl_listener *listener, void *data) {
 	if (seat_is_input_allowed(seat, surface)) {
 		wlr_seat_touch_notify_down(wlr_seat, surface, event->time_msec,
 				event->touch_id, sx, sy);
-		cursor->image_client = NULL;
-		wlr_cursor_set_image(cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
+		cursor_set_image(cursor, NULL, NULL);
 	}
 }
 
@@ -1175,11 +1174,13 @@ static void handle_request_set_cursor(struct wl_listener *listener,
 
 void cursor_set_image(struct sway_cursor *cursor, const char *image,
 		struct wl_client *client) {
-	if (!cursor->image || strcmp(cursor->image, image) != 0) {
+	if (!image) {
+		wlr_cursor_set_image(cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
+	} else if (!cursor->image || strcmp(cursor->image, image) != 0) {
 		wlr_xcursor_manager_set_cursor_image(cursor->xcursor_manager, image,
 				cursor->cursor);
-		cursor->image = image;
 	}
+	cursor->image = image;
 	cursor->image_client = client;
 }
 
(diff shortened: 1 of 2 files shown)


sway/input/cursor.c
```diff
@@ -1119,6 +1120,14 @@ static void handle_tool_tip(struct wl_listener *listener, void *data) {
 	transaction_commit_dirty();
 }
 
+static void handle_tool_proximity(struct wl_listener *listener, void *data) {
+	wlr_log(WLR_DEBUG, "proximity event");
```
emersion: Is it possible to remove this? It happens often and will clutter debug logs?

ianyfan: Whoops, forgot to remove that after testing

sway/input/seat.c
```diff
@@ -368,11 +369,23 @@ static void seat_update_capabilities(struct sway_seat *seat) {
...
 		}
 	}
 	wlr_seat_set_capabilities(seat->wlr_seat, caps);
+
+	// Hide cursor if seat doesn't have pointer capability
+	if ((caps & WL_SEAT_CAPABILITY_POINTER) == 0) {
+		wlr_cursor_set_image(seat->cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
+	} else {
+		wlr_xcursor_manager_set_cursor_image(seat->cursor->xcursor_manager,
+			"left_ptr", seat->cursor->cursor);
```
emersion: I think we already have a function (in `cursor.c`) to update the cursor (and ensure the cursor state is correctly updated): `cursor_set_image`

sway/input/seat.c
```diff
@@ -368,11 +369,23 @@ static void seat_update_capabilities(struct sway_seat *seat) {
 			caps |= WL_SEAT_CAPABILITY_TOUCH;
 			break;
 		case WLR_INPUT_DEVICE_TABLET_TOOL:
+			if (seat->cursor->proximity == WLR_TABLET_TOOL_PROXIMITY_IN) {
```
emersion: Hmmm. This will change when we'll support the tablet protocol, not sure it's worth it to do this

ianyfan: In that case, shall I just hold off with this PR until then?

emersion: Eh, master is broken anyway because if a tablet is plugged in the seat won't advertise the POINTER capability. I'd say just add the capability if there's a tablet, no need for complicated proximity logic.

ianyfan: I removed the proximity logic

sway/input/seat.c
```diff
@@ -368,11 +368,20 @@ static void seat_update_capabilities(struct sway_seat *seat) {
...
+			break;
 		case WLR_INPUT_DEVICE_TABLET_PAD:
 			break;
 		}
 	}
 	wlr_seat_set_capabilities(seat->wlr_seat, caps);
+
+	// Hide cursor if seat doesn't have pointer capability
+	if ((caps & WL_SEAT_CAPABILITY_POINTER) == 0) {
+		wlr_cursor_set_image(seat->cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
```
emersion: This should be `cursor_set_image(seat->cursor, NULL, NULL)`

ianyfan: Changing it to that causes a segfault

emersion: Can you change it to accept a NULL `image`?

ianyfan: I had a go.

Logic seems a bit messed up anyway. I wanted to add a check to only re-show the cursor if it isn't already visible, but it doesn't work.

emersion:
Thanks!

PR #2699, opened 2018-09-23
swaybg: fix increasingly smaller bg on hotplug

### LINKED ISSUE #2667: swaybg redraws an increasingly smaller background on hotplug
lvds-1

RedSoxFan:
Fixes #2667 

render_background_image alters the scale that cairo uses. Depending on the image mode, resolution, and image size, this may cause the surface to be rendered increasingly smaller. By calling cairo_save and
cairo_restore, any changes to the cairo settings by the function are not kept as a side effect.

The surface that swaybg uses is also now cleared before rendering a frame. This is needed to avoid artifacts  with certain combinations of image modes, resolutions, and image sizes. This was also part of the increasingly smaller background visual since it made it so it was not obvious the region being rendered to was smaller and caused an increasing number of smaller images to be appear for each hotplug.

### DIFF
--- a/common/background-image.c
+++ b/common/background-image.c
@@ -58,6 +58,7 @@ void render_background_image(cairo_t *cairo, cairo_surface_t *image,
 	double width = cairo_image_surface_get_width(image);
 	double height = cairo_image_surface_get_height(image);
 
+	cairo_save(cairo);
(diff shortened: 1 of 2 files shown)


emersion:
Thanks!

PR #2701, opened 2018-09-24
Implement support for input wildcard

### LINKED ISSUE #2662: Proposal: configure input devices with *
e.g.

```
input * {
  xkb_variant altgr-intl
}
```

This is especially useful if you routinely plug in random keyboards.

RedSoxFan:
Closes #2662 

Implements support for`input *`. The code is structured similar to the way `output *` is currently implemented.

This required restructuring the input subcommand handlers. Instead of creating a new input config and applying it in each subcommand, the one created by the main input command is used and the input config is stored/merged and applied after returning from the subcommand.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -458,11 +458,9 @@ struct input_config *new_input_config(const char* identifier);
 
 void merge_input_config(struct input_config *dst, struct input_config *src);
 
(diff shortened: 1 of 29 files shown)


emersion:
Thanks!

PR #2703, opened 2018-09-24
Add CSD to border modes

### LINKED ISSUE #2429: Add CSD to border modes

RyanDwyer:
This replaces `view.using_csd` with a new border mode: `B_CSD`. This also removes `sway_xdg_shell{_v6}_view.deco_mode` and `view->has_client_side_decorations` as we can now get these from the border.

You can use `border toggle` to cycle through the modes including CSD, or use `border csd` to set it directly. The client must support the xdg-decoration protocol, and the only client I know of that does is the example in wlroots.

If the client switches from SSD to CSD without us expecting it (via the server-decoration protocol), we stash the previous border type into `view.saved_border` so we can restore it if the client returns to SSD. I haven't found a way to test this though.

Closes #2429.

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -275,7 +275,7 @@ void view_set_csd_from_server(struct sway_view *view, bool enabled);
  * Updates the view's border setting when the client unexpectedly changes their
  * decoration mode.
  */
-void view_set_csd_from_client(struct sway_view *view, bool enabled);
+void view_update_csd_from_client(struct sway_view *view, bool enabled);
 
 void view_set_tiled(struct sway_view *view, bool tiled);
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -76,8 +76,19 @@ struct sway_view {
 	int natural_width, natural_height;
 
 	char *title_format;
+
+	// Our border types are B_NONE, B_PIXEL, B_NORMAL and B_CSD. We normally
+	// just assign this to the border property and ignore the other two.
+	// However, when a view using CSD is tiled, we want to render our own
+	// borders as well. So in this case the border property becomes one of the
+	// first three, and using_csd is true.
+	// Lastly, views can change their decoration mode at any time. When an SSD
+	// view becomes CSD without our approval, we save the SSD border type so it
+	// can be restored if/when the view returns from CSD to SSD.
 	enum sway_container_border border;
 	enum sway_container_border saved_border;
+	bool using_csd;
+
 	int border_thickness;
 	bool border_top;
 	bool border_bottom;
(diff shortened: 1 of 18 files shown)


include/sway/tree/view.h
```diff
@@ -268,6 +266,17 @@ void view_set_activated(struct sway_view *view, bool activated);
...
+/**
+ * If possible, instructs the client to change their decoration mode.
+ */
+void view_set_csd_from_server(struct sway_view *view, bool enabled);
+
+/**
+ * Updates the view's border setting when the client unexpectedly changes their
+ * decoration mode.
+ */
+void view_set_csd_from_client(struct sway_view *view, bool enabled);
```
emersion: I tried to use "set" for the server-side and "update" for the client-side. Is it possible to rename this to `view_update_csd_from_client`?

sway/xdg_decoration.c
```diff
@@ -0,0 +1,73 @@
...
+static void xdg_decoration_handle_surface_commit(struct wl_listener *listener,
+		void *data) {
+	struct sway_xdg_decoration *decoration =
+		wl_container_of(listener, decoration, surface_commit);
+
+	bool csd = decoration->wlr_xdg_decoration->current_mode ==
+		WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE;
+	struct sway_view *view = decoration->view;
+
+	view_set_csd_from_client(view, csd);
```
emersion: `current_mode` isn't the client preference, it's the current mode as chosen by the compositor.

To get the client preference, listen to `request_mode`.

RyanDwyer:
Copying my latest commit message into here...

This does the following:

* Removes the xdg-decoration `surface_commit` listener. I was under the impression the client could ignore the server's preference and set whatever decoration they like using this protocol, but I don't think that's right.
* Adds a listener for the xdg-decoration `request_mode` signal. The protocol states that the server should respond to this with its preference. We'll always respond with SSD here.
* Makes it so tiled views which use CSD will still have sway decorations rendered. To do this, `using_csd` had to be added back to the view struct, and the border is changed when floating or unfloating a view.

Also note that when testing with the wlroots toplevel-decoration example, that example does not support server-decoration, so according to the server-decoration protocol we assume the client is using CSD. However, the example client believes it's using SSD as that's sway's default, so there's a mismatch there until the first time you change the border.

SirCmpwn:
This can be a separate patch, but I wonder if we should make XDG windows tiled on all four sides if they're floating with a server-side border. Then GTK for example will stop using rounded corners.

emersion:
>I wonder if we should make XDG windows tiled on all four sides if they're floating with a server-side border

If we do, we should only do so if server-decoration says the window is using CSD and xdg-decoration isn't supported.

SirCmpwn:
Thanks!

PR #2704, opened 2018-09-24
input: enable configuring tablets with libinput

ianyfan:
This allows, for example, setting left handed mode for tablets.


emersion:
Are there pointer config options not supported by tablets? Are there additional tablet config options not supported by pointers?

ianyfan:
I don't think there are any additional tablet options.

I think there are pointer options not supported by tablets, but enough that it seems it would be a lot of code duplication, and the way it is handled is already a "throw everything at the wall and see what sticks" approach.

E.g. out of 12 possible options, my tablet supports 4 and my mouse supports 6.

SirCmpwn:
Thanks!

Issue #2525, opened 2018-08-27
Rendering is inconsistent with cursor events


RyanDwyer:
Follow on from #2494.

If you create `V[terminal gedit]`, resize gedit to be stupidly small, then open the `Open` menu in gedit, it appears over the terminal. If you then focus the terminal and move the cursor over the menu, the cursor events apply to the terminal instead of the menu.

Some suggestions for fixing are in the above PR.

emersion:
I still think rendering is inconsistent with pointer input, even if the particular issue explained above is fixed. We really need to take the input code and "reverse it" when writing the rendering code.

PR #2709, opened 2018-09-25
raise floating

BuJo:
Adds a configuration option to raise/not raise floating containers on gaining focus (see also https://github.com/i3/i3/issues/2990).

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -606,6 +606,18 @@ static int handle_urgent_timeout(void *data) {
 	return 0;
 }
 
+static void container_raise_floating(struct sway_container *con) {
+	// Bring container to front by putting it at the end of the floating list.
+	struct sway_container *floater = con;
+	while (floater->parent) {
+		floater = floater->parent;
+	}
+	if (container_is_floating(floater)) {
+		list_move_to_end(floater->workspace->floating, floater);
+		node_set_dirty(&floater->workspace->node);
+	}
+}
+
 void seat_set_focus_warp(struct sway_seat *seat, struct sway_node *node,
 		bool warp, bool notify) {
 	if (seat->focused_layer) {
@@ -733,16 +745,8 @@ void seat_set_focus_warp(struct sway_seat *seat, struct sway_node *node,
 	}
 
 	// If we've focused a floating container, bring it to the front.
-	// We do this by putting it at the end of the floating list.
 	if (container && config->raise_floating) {
-		struct sway_container *floater = container;
-		while (floater->parent) {
-			floater = floater->parent;
-		}
-		if (container_is_floating(floater)) {
-			list_move_to_end(floater->workspace->floating, floater);
-			node_set_dirty(&floater->workspace->node);
-		}
+		container_raise_floating(container);
 	}
 
 	if (last_focus) {
@@ -1018,17 +1022,9 @@ void seat_begin_down(struct sway_seat *seat, struct sway_container *con,
 	seat->op_ref_con_ly = sy;
 	seat->op_moved = false;
 
-	// If we've focused a floating container, bring it to the front.
-	// We do this by putting it at the end of the floating list.
+	// In case the container was not raised by gaining focus, raise on click
 	if (con && !config->raise_floating) {
-		struct sway_container *floater = con;
-		while (floater->parent) {
-			floater = floater->parent;
-		}
-		if (container_is_floating(floater)) {
-			list_move_to_end(floater->workspace->floating, floater);
-			node_set_dirty(&floater->workspace->node);
-		}
+		container_raise_floating(con);
 	}
 }
 
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -734,7 +734,7 @@ void seat_set_focus_warp(struct sway_seat *seat, struct sway_node *node,
 
 	// If we've focused a floating container, bring it to the front.
 	// We do this by putting it at the end of the floating list.
-	if (container) {
+	if (container && config->raise_floating) {
 		struct sway_container *floater = container;
 		while (floater->parent) {
 			floater = floater->parent;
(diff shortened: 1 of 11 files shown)


BuJo:
As I'm quite unfamiliar with the code-base:  I'm really not sure about the positioning of the 'raise on click'.  But I'm in awe how fast a working implementation could be created (I'm an openbox user and trying slowly to replace it).

sway/input/seat.c
```diff
@@ -989,6 +993,10 @@ void seat_begin_down(struct sway_seat *seat, struct sway_container *con,
 	seat->op_ref_con_lx = sx;
 	seat->op_ref_con_ly = sy;
 	seat->op_moved = false;
+
+	if (con && !config->raise_floating) {
```
RyanDwyer: This condition looks wrong, but I think it's right. If you've moved the mouse over the floating container and it's given focus but hasn't raised because `raise_floating` is disabled, this part of the code raises it on click. Maybe add a comment explaining why this is here.

RyanDwyer:
It looks like i3 ended up implementing this without a `raise_floating` directive. Perhaps we should do the same.

BuJo:
Hmm unsure on how this can be handled without a configuration option, as some people might prefer one of the behaviours.  It seemed to me that i3 simply reverted what appeared to be a mistaken commit.

sway/input/seat.c
BuJo: I amended the second commit with a comment.

ianyfan:
Could you add some documentation in `sway.5.scd` please?

BuJo:
Thank you! Will do.

BuJo:
I installed the manpage locally and looked at it, it looks ok.

BuJo:
Please hold off merging the branch for now.

I've been using this branch for a bit and I'm getting a few focus related bugs (where the focus is still on windows behind the foregroung one, missing focus borders..).  I'm not sure if it's related to my configuration or my code. I'll investigate.

BuJo:
I may have found the source of my problems regarding it not focusing other floating windows above the one with the focus.  #2492 adds code to `container.c:container_at()` which seems to "short circuit" container lookup as long as the cursor is still in the area of the focused view.

```diff
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -371,6 +371,17 @@ struct sway_container *container_at(struct sway_workspace *workspace,
        }
        // If focused is floating, focused view's non-popups
        if (focus && focus->view && is_floating) {
+               // Prefer floating views above the focused one
+               double _sx = 0, _sy = 0;
+               struct wlr_surface *_surface = NULL;
+               struct sway_container *_c = floating_container_at(lx, ly, &_surface, &_sx, &_sy);
+               if (_surface && _c && _c != focus && container_is_floating(_c)) {
+                       *surface = _surface;
+                       *sx = _sx;
+                       *sy = _sy;
+                       return _c;
+               }
+
                surface_at_view(focus, lx, ly, surface, sx, sy);
                if (*surface) {
                        return focus;
```

This is what I have so far.  I'm really not sure if this goes into the right direction.  I could not reproduce the steps outlined in #2492 with this code, but I'm not feeling at ease.

BuJo:
I found #2525 which is a problem not unlike this one...

BuJo:
Thanks for the review!
I'm also good with reworking the pr to do what i3 did (which means having no configuration and the focus raising off). Let's see what the others say, then the pr needs a little bit of cleanup (wip on one commit, the `container_is_floating()` might not be necessary, ...) before a merge.

BuJo:
I marked the mr as wip, because I managed to reproduce the steps from #2492 after all.

Experimenting with the following now:
```c
        // If focused is floating, focused view's non-popups
	if (focus && focus->view && is_floating) {
		bool has_subsurfaces = wl_list_length(&focus->view->surface->subsurfaces) > 0;
		c = floating_container_at(lx, ly, surface, sx, sy);
                 // only switch to second container if focused container has no menus open
		if (!has_subsurfaces && c) {
			return c;
		}
```

This feels a little like what other window managers do.. including oddities like "when the menu is closed, the focused window will not unfocus without another click".

* create two windows, one of the something like gnome-dictionary
* open a menu in gnome-dictionary overlapping the second window
* click something in the menu
* wriggle mouse a little
* focus should move to the second window, but does not
* unfocuses only on hitting root window, not when moving outside the menu area onto the second window

And then, the simple question for subsurfaces, I'm not sure on the rules here. Is this even a legal assumption? There could be other reasons for subsurfaces I'm not seeing.
"When there is a menu, let the window keep the focus"... I'd rather it only keep it as long as the pointer is within the surface (+ subsurfaces).
I apologize for the back and forth, I'll see what I can come up with.

BuJo:
Note, that this might fix #2525 too for the non-floating case, if the check were to be expanded for the tiling case. (That is, if this is a valid way of handling it...)

SirCmpwn:
I don't have nearly as much context for this problem as you do. You should make the call as to which behavior is best, and when you've done so and this PR is ready for code review, lmk.

SirCmpwn:
Take as much time as you need and don't worry too much about getting it perfect or performant, unless it's obviously broken or very slow.

BuJo:
Thank you @SirCmpwn for the kind words, I will tag you for a review when I feel that confident that the implementation is something that I **want** to use. Until then, I'll just keep doing what I do.

BuJo:
I found that the `surface_at_view()` function should be doing what I was thinking about in my comment before. Which is go through all the surfaces within a container including sub-surfaces and if the cursor is within their area.
Right now I'm experimenting with an approach of "raycasting" through the stack of containers.

```c
for (int i = workspace->floating->length - 1; i >= 0; --i) {
  struct sway_container *cn = workspace->floating->items[i];
  surface_at_view(cn, lx, ly, surface, sx, sy);
  if (*surface) return cn;
```
The commit b356b66 shows the basic idea without the popup handling.  I used it for a little bit and it handles really nicely and feels good.
I will test it for a little while.

BuJo:
@SirCmpwn I think it's ok enough for you to have a look. I threw gperf/perf at it, it doesn't seem too expensive (at least what I could gather), valgrind didn't shout at me much and I couldn't make it crash.
(I tested mostly in the nested X11 variant.)

SirCmpwn:
Thanks!

PR #2712, opened 2018-09-25
Add a missing null check for moving tiling containers

alexbakker:
Trying to move the About window in Android Studio causes a crash without this check.


RedSoxFan:
Thanks!

PR #2716, opened 2018-09-26
Replace getdelim to make swaybar work on FreeBSD

sghctoma:
As we have discussed in #2672 with @ianyfan, the Linux (glibc) and FreeBSD implementation of `getdelim` handles the "no more data, delimiter not found, and EOF not reached" scenario differently: FreeBSD handles it as an error, while Linux does not. This causes trouble with swaybar, since it uses `getdelim` with `EOF` as delimiter on an infinite stream, where `EOF` is never reached. 

This commit replaces the `getdelim`/`getline` calls in the `PROTOCOL_UNDEF` phase of `status_handle_readable` with reading all available data from the stream in one go. This is achieved by first issuing a `FIONREAD` ioctl request to get how much data is available, and than using `read` to get it all.

I have tested this with a i3's trivial-bar-script.sh, and the default config of i3status and py3status on both FreeBSD 12.0-CURRENT and Debian Buster.

### DIFF
--- a/swaybar/status_line.c
+++ b/swaybar/status_line.c
@@ -1,5 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include <fcntl.h>
+#include <sys/ioctl.h>
 #include <json-c/json.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


swaybar/status_line.c
```diff
@@ -34,18 +35,35 @@ bool status_handle_readable(struct status_line *status) {
...
 			status_error(status, "[error reading from status command]");
 			return true;
 		}
 
+		if ((size_t)available_bytes + 1 > status->buffer_size) {
+			// need room for leading '\0' too
+			status->buffer_size = available_bytes + 1;
+			status->buffer = realloc(status->buffer, status->buffer_size);
+		}
+		if (status->buffer == NULL) {
```
ianyfan: This if block seems like it should be in the previous block

SirCmpwn:
If my interpretation of the POSIX spec is correct, this is a bug in FreeBSD. Also, FIONREAD is non-standard.

swaybar/status_line.c
sghctoma: I didn't want to nest the `if`s, and also thought that an extra check on `status->buffer` won't hurt anyone. Logging `errno` here could be misleading though. All in all, I'm not adamant on keeping it this way.

sghctoma:
> If my interpretation of the POSIX spec is correct, this is a bug in FreeBSD.

In my understanding, using `EOF` as delimiter is undefined behaviour, because POSIX states the following about `getdelim` (highlight by me):

"The delimiter argument is an int, the value of which the application shall ensure is a character **representable as an unsigned char of equal value** that terminates the read process."

The delimiter here is `EOF`, which "shall expand to an integer constant expression with type int and a **negative value**" as per POSIX. 

> Also, FIONREAD is non-standard.

Indeed. I just assumed it is, because I "stole" the idea from sway/ipc-server.c.

PR #2719, opened 2018-09-28
Fix floating views in tabbed/stacked workspaces not getting frame events

### LINKED ISSUE #2659: Some dialogs have damage issues
1. Open Atom
2. Create a new file
3. Write "asdf"
4. <kbd>Ctrl</kbd>+<kbd>W</kbd>, a dialog asking if we want to save the file pops up
5. Buttons don't change state when hovered. With focus follows mouse enabled, focus doesn't go away from the dialog. Clicking works though.

RyanDwyer:
`view_is_visible` would return false, which meant the view wouldn't receive a frame done event. `view_is_visible` needs to make an exception for floating containers.

This also moves the `workspace_is_visible` check to an earlier location for performance reasons.

Fixes #2659.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -989,12 +989,16 @@ bool view_is_visible(struct sway_view *view) {
 		floater = floater->parent;
 	}
 	bool is_sticky = container_is_floating(floater) && floater->is_sticky;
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

emersion:
Thanks!

PR #2720, opened 2018-09-28
Add support for building swaylock without PAM

SirCmpwn:
This involves setuid'ing swaylock, which then forks and drops perms on
the parent process. The child process remains root and listens on a pipe
for requests to validate passwords against /etc/shadow.

### DIFF
--- a/include/swaylock/swaylock.h
+++ b/include/swaylock/swaylock.h
@@ -101,5 +101,8 @@ void render_frame(struct swaylock_surface *surface);
 void render_frames(struct swaylock_state *state);
 void damage_surface(struct swaylock_surface *surface);
 void damage_state(struct swaylock_state *state);
+void initialize_pw_backend(void);
+bool attempt_password(struct swaylock_password *pw);
+void clear_password_buffer(struct swaylock_password *pw);
 
 #endif
--- a/swaylock/shadow.c
+++ b/swaylock/shadow.c
@@ -1,0 +1,128 @@
+#define _XOPEN_SOURCE
+#include <pwd.h>
+#include <shadow.h>
+#include <stdbool.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <wlr/util/log.h>
+#include "swaylock/swaylock.h"
+
+static int comm[2][2];
+
+void run_child(void) {
+	/* This code runs as root */
+	struct passwd *pwent = getpwuid(getuid());
+	if (!pwent) {
+		wlr_log_errno(WLR_ERROR, "failed to getpwuid");
+		exit(EXIT_FAILURE);
+	}
+	char *encpw = pwent->pw_passwd;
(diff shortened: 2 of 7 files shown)


include/swaylock/swaylock.h
```diff
@@ -101,5 +101,8 @@ void render_frame(struct swaylock_surface *surface);
 void render_frames(struct swaylock_state *state);
 void damage_surface(struct swaylock_surface *surface);
 void damage_state(struct swaylock_state *state);
+void initialize_pw_backend();
```
emersion: `void initialize_pw_backend(void);`

swaylock/shadow.c
```diff
@@ -0,0 +1,124 @@
+#define _XOPEN_SOURCE
```
emersion: We should probably use `_POSIX_C_SOURCE` instead

swaylock/shadow.c
```diff
@@ -0,0 +1,124 @@
...
+			exit(EXIT_FAILURE);
+		}
+		encpw = swent->sp_pwdp;
+	}
+	wlr_log(WLR_DEBUG, "prepared to authorize user %s", pwent->pw_name);
+
+	size_t size;
+	char *buf;
+	while (1) {
+		if (read(comm[0][0], &size, sizeof(size)) < 0) {
```
emersion: We could add proper error handling here. Zero means it's EOF, -1 means something bad happened.

swaylock/shadow.c
```diff
@@ -0,0 +1,124 @@
...
+		wlr_log(WLR_DEBUG, "received pw check request");
+		buf = malloc(size);
+		if (!buf) {
+			wlr_log_errno(WLR_ERROR, "failed to malloc pw buffer");
+			exit(EXIT_FAILURE);
+		}
+		size_t offs = 0;
+		do {
+			ssize_t amt = read(comm[0][0], &buf[offs], size - offs);
+			if (amt < 0) {
```
emersion: `<=`

swaylock/shadow.c
```diff
@@ -0,0 +1,124 @@
...
+				wlr_log_errno(WLR_ERROR, "failed to read pw");
+				exit(EXIT_FAILURE);
+			}
+			offs += (size_t)amt;
+		} while (offs < size);
+		bool result = false;
+		char *c = crypt(buf, encpw);
+		if (c == NULL) {
+			wlr_log_errno(WLR_ERROR, "crypt:");
```
emersion: Missing a `exit` here?

(Minor: no need for `:`)

swaylock/shadow.c
SirCmpwn: shadow.h is actually part of the open source extension so here it makes sense

emersion:
Thanks!

PR #2722, opened 2018-09-28
Rename workspace_outputs to workspace_configs and fix memory leak

RyanDwyer:
When we eventually implement `workspace <ws> gaps inner|outer <px>`, we'll need to store the gaps settings for workspaces before they're created. Rather than create a `workspace_gaps` struct, the approach I'm taking is to rename `workspace_outputs` to `workspace_configs` and then add gaps settings to that.

I've added a lookup function `workspace_find_config`. Note that we have a similar thing for outputs (`output_config` struct and `output_find_config`).

Lastly, when freeing config it would create a memory leak by freeing the list items but not the workspace or output names inside them. This has been rectified using a `free_workspace_config` function.

### DIFF
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -128,7 +128,7 @@ void next_name_map(struct sway_container *ws, void *data) {
 static bool workspace_valid_on_output(const char *output_name,
 		const char *ws_name) {
 	struct workspace_config *wsc = workspace_find_config(ws_name);
-	return !wsc || strcmp(wsc->output, output_name) == 0;
+	return !wsc || !wsc->output || strcmp(wsc->output, output_name) == 0;
 }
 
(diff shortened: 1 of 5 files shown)


sway/tree/workspace.c
```diff
@@ -121,17 +127,8 @@ void next_name_map(struct sway_container *ws, void *data) {
...
-		if (strcasecmp(wso->workspace, ws_name) == 0) {
-			if (strcasecmp(wso->output, output_name) != 0) {
-				return false;
-			}
-		}
-	}
-
-	return true;
+	struct workspace_config *wsc = workspace_find_config(ws_name);
+	return !wsc || strcmp(wsc->output, output_name) == 0;
```
emersion: It seems that in the future, `wsc->output` may be NULL. We probably need to check this here.

emersion:
Thanks!

PR #2723, opened 2018-09-28
swaybar: small fixes and clean-ups

ianyfan:
The only real fix is setting the binding mode indicator, the others are mostly aesthetic changes.


SirCmpwn:
Thanks!

PR #2726, opened 2018-09-29
Make gaps implementation consistent with i3-gaps

RyanDwyer:
This changes our gaps implementation to behave like i3-gaps.

Our previous implementation allowed you to set gaps on a per container basis. This is no longer possible as it isn't supported by i3-gaps and doesn't seem to have a practical use case. The `gaps_outer` and `gaps_inner` properties on containers are now removed as they just read the `gaps_inner` from the workspace.

`gaps inner|outer <px>` no longer changes the gaps for all workspaces. It only sets defaults for new workspaces.

`gaps inner|outer current|workspace|all set|plus|minus <px>` is now runtime only, and the workspace option is now removed. `current` now sets gaps for the current workspace as opposed to the current container.

`workspace <ws> gaps inner|outer <px>` is now implemented. This sets defaults for a workspace.

This also fixes a bug where changing the layout of a split container from linear to tabbed would cause gaps to not be applied to it until you switch to another workspace and back.

i3-gaps docs: https://github.com/Airblader/i3#gaps

Supersedes/closes #2131.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -173,6 +173,8 @@ struct output_config {
 struct workspace_config {
 	char *workspace;
 	char *output;
+	int gaps_inner;
+	int gaps_outer;
 };
 
 struct bar_config {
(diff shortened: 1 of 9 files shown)


SirCmpwn:
Thanks!

PR #2728, opened 2018-09-29
Move sticky containers in output_evacuate

RedSoxFan:
When evacuating an output, sticky containers need to be moved to a different output. Currently, they are not moved since `workspace_is_empty` returns `true` when there are only sticky containers.

This PR adds an additional check after the `workspace_is_empty` call that checks whether there are `floating` containers. If there are, they must all be sticky. The sticky containers get moved to another output and the workspace gets destroyed as normal.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -823,9 +823,16 @@ void container_floating_move_to_center(struct sway_container *con) {
 		return;
 	}
 	struct sway_workspace *ws = con->workspace;
+	bool full = con->is_fullscreen;
+	if (full) {
+		container_set_fullscreen(con, false);
+	}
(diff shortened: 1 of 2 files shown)


RyanDwyer:
`workspace_is_empty` is only called from `output_evacuate`, so we could make it include sticky views. This would simplify both `workspace_is_empty` and `output_evacuate` because it would move the workspace like usual. Thoughts?

RedSoxFan:
> Thoughts?

The issue with that solution is it only works if the workspace that would be moved has focus. If it does not, the sticky container(s) is/are not visible on the new output (which could be worked around, but then we would have an empty workspace that is not visible anyway).

SirCmpwn:
Thanks!

PR #2729, opened 2018-09-29
Return an error when resizing is a no op

RyanDwyer:
See https://github.com/swaywm/sway/issues/1737#issuecomment-425583060


freed00m:
Maybe there is still a issue, I clearly see no op on window and still returned true.

Images posted below are here just to demonstrate that window dimensions remained same but success was true.

```
sway version 1.0-alpha.6-191-g4bebee62 (Oct  9 2018, branch 'master')
```

2018-10-09-130227_grim
2018-10-09-130232_grim

PR #2730, opened 2018-09-29
Remove libcap/prctl artifacts

ivyl:
They seem like relics of the pasts, from when we were retaining the
ptrace cap.

Some translations still may need updates.


sway/main.c
```diff
@@ -181,28 +177,8 @@ static void log_kernel() {
...
-			wlr_log(WLR_ERROR,
-				"If unsure, strip the file caps.");
-			exit(EXIT_FAILURE);
-		}
-		free(exe);
-#endif
-}
 
-static void drop_permissions(bool keep_caps) {
+static void drop_permissions() {
```
ianyfan: `drop_permissions(void)`

ivyl: is this a convention here? I neither see other functions in the immediate neighborhood using this, nor CONTRIBUTIN.md mentions it

emersion: In C, `void call_me_maybe()` declares a function that takes an unspecified number of parameters. So a call like `call_me_maybe(1, 2, 3)` is valid.

To declare a function that takes no parameter, one should use `void call_me_maybe(void)`.

ianyfan: @ivyl Leaving it empty is a mistake and should be corrected, along with the other functions in the code.

ivyl: Oh yeah, forgot about that C quirk. Thanks!

2 additional patches to follow this one coming in 5 :-)

SirCmpwn:
Thanks!

PR #2732, opened 2018-09-30
Fix hotplugging down to zero outputs

RyanDwyer:
When the last output is disconnected, `output_disable` is called like usual and evacuates the output to the `root->saved_workspaces` list. It then calls `root_for_each_container` to remove (untrack) the output from each container's outputs list. However, `root_for_each_container` did not iterate the saved workspaces, so when the output gets freed the containers would have a dangling pointer in their outputs list. Upon reconnect, `container_discover_outputs` would attempt to use the dangling pointer, causing a crash.

This makes `root_for_each_container` check the saved workspaces list, which fixes the problem. It also adds it to `root_find_container` for good measure.

### DIFF
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -273,6 +273,12 @@ void root_for_each_container(void (*f)(struct sway_container *con, void *data),
 			container_for_each_child(container, f, data);
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #2733, opened 2018-09-30
swaybar: synchronize rendering to output frames

### LINKED ISSUE #2692: swaybar: simultaneous events may cause it to run out of buffers and not render new status
Suppose your status command sends a status upon workspace change.

Then you switch to a new workspace.

Then you get events: workspace::new, workspace::focus, status command.

Each event triggers a render, using a buffer and is committed. But there are only two buffers, so the status command event does not get a buffer, and is therefore not rendered.

Introduced by 8a5ff5a4dbcefabaaeae98c1c53befcb986ba117 since it removes the `wl_display_roundtrip`s.

ianyfan:
Fixes #2692


emersion:
Thanks!

PR #2734, opened 2018-09-30
Add support for installing binaries with DT_RPATH

ivyl:
It's better to use DT_RPATH dynamic section of the elf binary to store
the paths of libraries to load instead of overwriting LD_LIBRARY_PATH
for the whole environment, causing surprises. This solution is much more
transparent and perfectly suitable for running contained installations
of wayland/wlroots/sway.

The code unsetting the LD_LIBRARY_PATH/LD_PRELOAD was also deleted as
it's a placebo security at best - we should trust the execution path
that leads us to running sway, and it's way too late to care about those
variables since we already started executing our compositor, thus we
would be compromised anyway.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -128,7 +128,16 @@ else
 endif
 add_project_arguments('-DSWAY_VERSION=@0@'.format(version), language: 'c')
 
-add_project_arguments('-D_LD_LIBRARY_PATH="@0@"'.format(get_option('ld-library-path')), language: 'c')
+if get_option('use_rpath')
+	if get_option('custom_rpath') == ''
+		# default to platform specific libdir, one level up from the binary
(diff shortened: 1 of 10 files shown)


ivyl:
If I am missing something, I would appreciate the explanation so we can document that in the code/readme

SirCmpwn:
>The code unsetting the LD_LIBRARY_PATH/LD_PRELOAD was also deleted as it's a placebo security at best - we should trust the execution path that leads us to running sway, and it's way too late to care about those variables since we already started executing our compositor, thus we would be compromised anyway.

Eh, the reasons behind this are more subtle than you think. But ultimately this is a good argument anyway. Can you update sway-security(7) accordingly?

emersion:
Note: it seems that we don't have a sway-security(7) manpage anymore.

emersion:
Thanks!

PR #2735, opened 2018-09-30
Do not strip quotes for exec or bind commands

### LINKED ISSUE #2736: `set` options and $mode_system are broken
In my config, I set the following style options for `dmenu`:
```
set $dmopts -i -fn 'Liberation Mono-9' -nb '#303030' -nf '#FFFFFF' -sb '#338DD6'
...
bindsym $mod+d          exec --no-startup-id dmenu_run $dmopts
```
And these are my binds for `$mode_system`
```
set $mode_system (l) lock, (e) logout, (s) suspend, (r) reboot, (S) shutdown, (R) UEFI
...
mode "$mode_system" {
	bindsym l exec $Locker, mode "default"
	bindsym e exit
	bindsym s exec --no-startup-id systemctl suspend, mode "default"
	bindsym r exec --no-startup-id systemctl reboot, mode "default"
	bindsym Shift+s exec --no-startup-id systemctl poweroff -i, mode "default"
	bindsym Shift+r exec --no-startup-id systemctl reboot --firmware-setup, mode "default"

	# back to normal: Enter or Escape
	bindsym Return mode "default"
	bindsym Escape mode "default"
	bindsym q mode "default"
}
...
bindsym $mod+Shift+e mode "$mode_system"
```
Both of these worked on dc01e884. I've reproduced this issue on two different computers.

### LINKED ISSUE #2738: Config wildcards always expand
`sway version 1.0-alpha.6-105-g51f68e10 (Sep 30 2018, branch 'master')`

```
bindsym $mod+Shift+z exec swaymsg 'output * dpms off'
bindsym XF86Tools exec swaymsg "output * dpms on"
```
now always expand `*` even if they are quoted, breaking `swaymsg` and potentially other functionality.
Personally, I think it should not expand when in single quotes like how `bash` handles wildcards.

This was introduced recently, but I can't find the specific changes.

RedSoxFan:
Leave quotes intact for `cmd_exec`, `cmd_exec_always`, `cmd_bindcode`, `cmd_bindsym`, and `cmd_set`.

`cmd_exec` and `cmd_exec_always` should not have any quotes stripped from them and should be passed to `exexcl` as is.

`cmd_bindcode` and `cmd_bindsym` will have their bound command processed by `execute_command`, which will strip quotes for everything except `cmd_exec`.

Fixes #2736
Fixes #2738

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -391,7 +391,10 @@ struct cmd_results *config_command(char *exec) {
 	// Var replacement, for all but first argument of set
 	// TODO commands
 	for (i = handler->handle == cmd_set ? 2 : 1; i < argc; ++i) {
-		if (*argv[i] == '\"' || *argv[i] == '\'') {
+		if (handler->handle != cmd_exec && handler->handle != cmd_exec_always
+				&& handler->handle != cmd_bindsym
+				&& handler->handle != cmd_bindcode
+				&& (*argv[i] == '\"' || *argv[i] == '\'')) {
 			strip_quotes(argv[i]);
 		}
 		argv[i] = do_var_replacement(argv[i]);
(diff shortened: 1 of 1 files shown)


minus7:
bindsym with quoted exec works again with this for me

SirCmpwn:
Thanks!

PR #2737, opened 2018-09-30
Parse floating criteria

Ragnis:
This makes the criteria parser recognize the "floating" token. From there on the criteria seems to be handled as expected.


SirCmpwn:
Thanks!

PR #2739, opened 2018-10-01
Fix smart gaps

### LINKED ISSUE #2653: Smart Gaps not working
I enable smart_gaps and it's just like it wasn't enabled

RedSoxFan:
Fixes #2653 

This PR should make it so `smart_gaps` works identical to how it does in `i3-gaps`.

Since `i3-gaps` does not allow for floating containers (only floating views), there is no behaviour to replicate. Currently, each floating container gets treated as a separate group and will apply gaps accordingly.


RyanDwyer:
Thanks!

PR #2740, opened 2018-10-01
Handle gap borders

RedSoxFan:
Fixes `hide_edge_borders smart` when gaps are in use.
Implements `hide_edge_borders smart_no_gaps` and `smart_borders on|no_gaps|off`.

Since `smart_borders on` is equivalent to `hide_edge_borders smart` and `smart_borders no_gaps` is equivalent to `hide_edge_borders smart_no_gaps`, I opted to just save the last value set for `hide_edge_borders` and restore that on `smart_borders off`. This simplifies the conditions for setting the border.


RedSoxFan:
This has bee rebased and is ready for review

RyanDwyer:
Thanks!

PR #2745, opened 2018-10-02
Set focus_inactive on a sibling when a container closes in an inactive workspace

### LINKED ISSUE #2625: Closing windows in a stacked layout when on another workspace focuses parent container
To reproduce: create a few windows in a stacked container, for example a video player. Switch to another workspace and either leave the player to close on its own or close it remotely.
When you return to the workspace that had the player, notice how the parent stacked container is focused rather than the previously focused window in a stacked layout. This only happens if the window is closed when in another workspace and makes it harder to repeat the same operation done before.

RyanDwyer:
To reproduce the problem, create layout `H[view V[view view view-focused]]`, then switch to another workspace and have the previously focused view in the vsplit close (eg. using criteria, or an mpv video finishing). Return to the workspace using `$mod+<num>` and the entire vsplit would be focused. This happens because `handle_seat_node_destroy` would only set a new focus if the currently focused view or a parent was being destroyed. To fix it, it needs to set a sibling of the destroying container to `focus_inactive` regardless of the current focus, then restore current focus if needed.

This patch changes the function accordingly. Additionally:

* The function now makes an early return if the node being destroyed is a workspace.
* `set_focus` has been renamed to `needs_new_focus`. This variable is true if the head focus needs to be changed.

Fixes #2625

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -144,32 +144,43 @@ static void handle_seat_node_destroy(struct wl_listener *listener, void *data) {
 	struct sway_node *parent = node_get_parent(node);
 	struct sway_node *focus = seat_get_focus(seat);
 
-	bool set_focus =
-		focus != NULL &&
-		(focus == node || node_has_ancestor(focus, node)) &&
-		node->type == N_CONTAINER;
+	if (node->type == N_WORKSPACE) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #2751, opened 2018-10-02
Bar mode/hidden_state events

ianyfan:
TODO:
- [x] send stop/start signals to status
- [x] key events - cannot be done using current layer shell protocol due to focus issues, may need a new protocol or IPC event
- [x] show bar on workspace urgency
- [x] documentation

~Priority low on this one, it's mostly there but I don't think I will be working that much on it for now.~

### DIFF
--- a/include/ipc.h
+++ b/include/ipc.h
@@ -30,6 +30,9 @@ enum ipc_command_type {
 	IPC_EVENT_BINDING = ((1<<31) | 5),
 	IPC_EVENT_SHUTDOWN = ((1<<31) | 6),
 	IPC_EVENT_TICK = ((1<<31) | 7),
+
+	// sway-specific event types
+	IPC_EVENT_BAR_STATE_UPDATE = ((1<<31) | 20),
 };
 
 #endif
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -191,6 +191,7 @@ struct bar_config {
 	 * In "show" mode, it will always be shown on top of the active workspace.
 	 */
 	char *hidden_state;
+	bool visible_by_modifier; // only relevant in "hide" mode
 	/**
 	 * Id name used to identify the bar through IPC.
 	 *
--- a/include/sway/ipc-server.h
+++ b/include/sway/ipc-server.h
@@ -15,6 +15,7 @@ void ipc_event_workspace(struct sway_workspace *old,
 		struct sway_workspace *new, const char *change);
 void ipc_event_window(struct sway_container *window, const char *change);
 void ipc_event_barconfig_update(struct bar_config *bar);
+void ipc_event_bar_state_update(struct bar_config *bar);
 void ipc_event_mode(const char *mode, bool pango);
 void ipc_event_shutdown(const char *reason);
 void ipc_event_binding(struct sway_binding *binding);
--- a/include/swaybar/bar.h
+++ b/include/swaybar/bar.h
@@ -47,6 +47,8 @@ struct swaybar_hotspot {
 
 struct swaybar {
 	char *id;
+	char *mode;
+	bool mode_pango_markup;
 
 	struct wl_display *display;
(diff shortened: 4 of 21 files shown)


SirCmpwn:
> key events - cannot be done using current layer shell protocol due to focus issues, may need a new protocol or IPC event

Like I said on IRC, it *can* be - you just need sway to be paying attention and let swaybar know over IPC when the button is pressed or released.

ianyfan:
Functionality is basically all there now, though I may still make small implementation changes.

ianyfan:
I fixed mode & hidden_state, I'm not sure about the other ones though.

RedSoxFan:
The handler structs/selection need to be modified to allow the commands at runtime

ianyfan:
I'm fine with only allowing performing `mode` and `hidden_state` commands at runtime, though there's a bug with that currently: https://github.com/swaywm/sway/pull/2791#issuecomment-427852910

SirCmpwn:
Wait how are you actually supposed to confiugre the bar to be hidden until you press mode

RedSoxFan:
> I'm fine with only allowing performing mode and hidden_state commands at runtime

They should be config and runtime

> though there's a bug with that currently: #2791 (comment)

I'll submit a PR to fix it. I think I messed it up during the conversion to subcommands a while ago

> Wait how are you actually supposed to confiugre the bar to be hidden until you press mod

I think `bar mode hide [<bar-id>]` is what you are looking for

ianyfan:
This PR is ready for review, though until https://github.com/swaywm/sway/pull/2808 is merged, this requires a patch in order to test changing mode/state at runtime:

```
diff --git a/sway/commands/bar.c b/sway/commands/bar.c
index 03f4c557..b0733452 100644
--- a/sway/commands/bar.c
+++ b/sway/commands/bar.c
@@ -46,14 +46,16 @@ struct cmd_results *cmd_bar(int argc, char **argv) {
 		return error;
 	}
 
-	if (find_handler(argv[0], bar_config_handlers,
-				sizeof(bar_config_handlers))) {
-		if (config->reading) {
-			return config_subcommand(argv, argc, bar_config_handlers,
-					sizeof(bar_config_handlers));
+
+
+	if (!config->reading) {
+		if (!find_handler(argv[0], bar_config_handlers,
+					sizeof(bar_config_handlers))) {
+			return cmd_results_new(CMD_FAILURE, "bar",
+					"Can only be used in config file.");
 		}
-		return cmd_results_new(CMD_FAILURE, "bar",
-				"Can only be used in config file.");
+		return config_subcommand(argv, argc, bar_config_handlers,
+				sizeof(bar_config_handlers));
 	}
 
 	if (argc > 1) {
```

SirCmpwn:
I guess I'm probably misunderstanding this but I have no idea how to make this patch have any effect on swaybar. Can someone provide an example config and steps to reproduce the correct behavior?

ianyfan:
I'm a bit confused what you're asking. The patch in the comment is just so you can run `bar mode ...`/`bar hidden_state ...` at runtime.

As for this PR, it allows using the bar subcommands `mode`, `hidden_state` and `modifier` as expected, in the config file and during runtime.

Example:
```
bar {
    mode hide
    hidden_state hide # default, but provided here for explicitness
    modifier Mod4 # also default
}
```
hides the bar, but shows it if you press down Mod4 or when there is an urgent workspace.

`bar mode invisible` hides the bar completely.

`bar hidden_state show` which only has an effect in `bar mode hide` behaves like hide but the bar is always shown, on top of the workspace.

SirCmpwn:
I'm dumb, I was running /usr/bin/swaybar instead of ./swaybar/swaybar

SirCmpwn:
LGTM, why [WIP]?

ianyfan:
Forgot to change it

SirCmpwn:
Thanks!

PR #2760, opened 2018-10-04
Give focus to another swaylock surface when output is disconnected

RyanDwyer:
* Have multiple outputs
* Launch swaylock
* Unplug an output (possibly has to be the last "connected" one)
* The swaylock surface on the remaining output would not respond to key events

This was happening because when the output destroys, focus was not given to the other swaylock surface.

This patch makes focus be transferred to another layer surface owned by the same Wayland client, but only if input was inhibited by the surface being destroyed, and only if it's in the overlay layer. I figure it's best to be overly specific and relax the requirements later if needed.

This patch removes a check in `seat_set_focus_surface` which was preventing focus from being passed from a layer surface to any other surface. I don't know of a use case for this check, but it's possible that this change could produce issues.

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -216,12 +216,48 @@ void arrange_layers(struct sway_output *output) {
 	}
 }
 
+static struct sway_layer_surface *find_any_layer_by_client(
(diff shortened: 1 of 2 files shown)


emersion:
Thanks!

PR #2761, opened 2018-10-04
Ignore unrelated cursor buttons while doing seat operations

### LINKED ISSUE #2756: Inconsistency of pointer events when holding pointer buttons outside of a surface
This can be shown using an example program displaying pointer events to stdout, such as https://github.com/Smithay/client-toolkit/blob/master/examples/pointer_input.rs

Steps to reproduce:

- click a pointer button on a surface
- holding the button, move the pointer outside of the surface, onto the window of an other program
  - the surface continues receiving `wl_pointer.motion` events with coordinates outside of it, but still in its coordinate space
- still holding the button, also click a second pointer button while the pointer is on the other window
  - the original surface continues receiving `wl_pointer.motion` events when the pointer is moved, but now *they are in the coordinate space of the surface on top of which the pointer is*.

This does not reproduce on rootston, as rootston stop sending `wl_pointer.motion` events when the pointer leaves the surface, even if a button is still pressed.

I reproduced this on sway `4e4d0c519140ae9c7c90e358d18bd6cd75413e30` compiled with wlroots `1dd523c34c71a7d156c62bc3e3f033113b666e4d`.

RyanDwyer:
* Click and hold a scrollbar
* Drag the cursor onto another surface
* While still holding the original button, press and release another cursor button
* Things get weird

There's two ways to fix this. Either cancel the seat operation and do the other click, or continue the seat operation and ignore the other click. I opted for the latter (ignoring the click) because it's easier to implement, and I suspect a second click during a seat operation is probably unintentional anyway.

Fixes #2756

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -754,11 +754,12 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 	}
 	struct sway_seat *seat = cursor->seat;
 
-	// Handle ending seat operation
-	if (cursor->seat->operation != OP_NONE &&
-			button == cursor->seat->op_button && state == WLR_BUTTON_RELEASED) {
-		seat_end_mouse_operation(seat);
-		seat_pointer_notify_button(seat, time_msec, button, state);
+	// Handle existing seat operation
+	if (cursor->seat->operation != OP_NONE) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #2762, opened 2018-10-04
[WIP] Don't schedule unnecessary output frames, and simplify damage functions

### LINKED ISSUE #2748: sway wasting significant CPU on clients that don't refresh often
To reproduce: play any audio file with embedded artwork in mpv with the wayland backend
Result: sway wastes 60% CPU while mpv wastes 10% more than it needs to

What happens is that with embedded artwork, mpv doesn't have anything to draw hence reschedules the frame callback to the next opportunity using wl_surface_commit() without damaging anything.
sway seems to do a lot of spinlocking and locking in general as seen on perf. At least it doesn't damage the whole surface.

RyanDwyer:
The `whole` argument has been removed from `damage_surface_iterator` because it was never used. As this function now only deals with partial damaging, it's been changed to return early if there is no damage and avoids scheduling a frame unnecessarily.

With the removal of the `whole` argument, `output_damage_from_view` just called `output_damage_view` with the same arguments, so I've combined these into the one function and renamed it to `output_damage_view_surfaces`.

Lastly, `output_damage_whole_container_iterator` is now removed as it was never used. `output_damage_whole_container` just damages the container including decorations and doesn't care about surfaces.

Probably closes #2748, but I'd like to see before and after metrics...

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -392,44 +392,37 @@ void output_damage_whole(struct sway_output *output) {
 
 static void damage_surface_iterator(struct sway_output *output,
 		struct wlr_surface *surface, struct wlr_box *_box, float rotation,
-		void *_data) {
-	bool *data = _data;
-	bool whole = *data;
-
+		void *data) {
+	if (!pixman_region32_not_empty(&surface->buffer_damage)) {
+		return;
+	}
 	struct wlr_box box = *_box;
 	scale_box(&box, output->wlr_output->scale);
 
 	int center_x = box.x + box.width/2;
 	int center_y = box.y + box.height/2;
 
-	if (pixman_region32_not_empty(&surface->buffer_damage)) {
-		enum wl_output_transform transform =
-			wlr_output_transform_invert(surface->current.transform);
+	enum wl_output_transform transform =
+		wlr_output_transform_invert(surface->current.transform);
 
-		pixman_region32_t damage;
-		pixman_region32_init(&damage);
-		pixman_region32_copy(&damage, &surface->buffer_damage);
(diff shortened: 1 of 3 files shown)


emersion:
>The whole argument has been removed from damage_surface_iterator because it was never used

What happens if you map a view with popups/subsurfaces outside the window geometry?

RyanDwyer:
`handle_commit` calls `view_damage_from`, which iterates each output and calls `output_damage_view_surfaces`. That calls `output_view_for_each_surface`, therefore calling `damage_surface_iterator` with each surface. It works the same way as it did before.

emersion:
>therefore calling damage_surface_iterator with each surface

I mean, it should damage the whole surfaces, it shouldn't take the damage from the surfaces. So we need `whole`.

sway/desktop/output.c
```diff
@@ -392,44 +392,37 @@ void output_damage_whole(struct sway_output *output) {
 
 static void damage_surface_iterator(struct sway_output *output,
 		struct wlr_surface *surface, struct wlr_box *_box, float rotation,
-		void *_data) {
-	bool *data = _data;
-	bool whole = *data;
-
+		void *data) {
+	if (!pixman_region32_not_empty(&surface->buffer_damage)) {
+		return;
```
emersion: Not sure this is correct. We should still send `frame` events to surfaces that registered a frame callback even if there's no damage.

emersion: (Note: not sending frame callbacks when the view isn't damaged has caused bugs with mpv in the past)

RyanDwyer:
Superseded by #2769.

PR #2768, opened 2018-10-05
Fix crash when flattening container after moving

### LINKED ISSUE #2765: Crash when moving window out of container
To reproduce:

Create S[viewa, viewb]
Now move viewa or viewb out of the container, to get crash.

Log attached.
[sway.log](https://github.com/swaywm/sway/files/2447889/sway.log)

Edit: Simpler case

RyanDwyer:
`container_flatten` removes the container from the tree (via `container_replace`) before destroying it. When destroying, the recent changes to `handle_seat_node_destroy` incorrectly assumes that the container has a parent.

This adds a check for destroying a container which is no longer in the tree. If this is the case, focus does not need to be changed.

To test, create layout `H[view T[view]]`, focus the tabbed view and run `move left`.

Fixes #2765

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -157,6 +157,11 @@ static void handle_seat_node_destroy(struct wl_listener *listener, void *data) {
 
 	seat_node_destroy(seat_node);
 
+	if (!parent) {
+		// Destroying a container that is no longer in the tree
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #2769, opened 2018-10-05
Prevent excessive frame events when we ignore a frame

### LINKED ISSUE #2748: sway wasting significant CPU on clients that don't refresh often

RyanDwyer:
In wlroots, if the compositor doesn't swap the output's buffers in response to a frame event then wlroots would emit a frame event whenever the Wayland event loop becomes idle, and do so repeatedly, because the idle event source is only removed when the buffers are swapped.

A proper fix for this is to make wlroots request frames from the backend, but in the interim this patch disables the idle event source if we're ignoring the frame event.

This prevents sway from throttling the CPU when mpv displays album art when mpv is using the wayland backend.

This also adds a missing call to `pixman_region32_fini`.

Fixes #2748

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -369,11 +369,20 @@ static void damage_handle_frame(struct wl_listener *listener, void *data) {
 	pixman_region32_t damage;
 	pixman_region32_init(&damage);
 	if (!wlr_output_damage_make_current(output->damage, &needs_swap, &damage)) {
+		pixman_region32_fini(&damage);
 		return;
 	}
 
 	if (needs_swap) {
 		output_render(output, &now, &damage);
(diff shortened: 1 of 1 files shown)


RyanDwyer:
A better fix is here: https://github.com/swaywm/wlroots/pull/1282

PR #2771, opened 2018-10-05
swaylock: Support keyboard and pointer disconnects and reconnects

RyanDwyer:
* Lock the screen with swaylock
* Do something that triggers the seat to update its capabilities (unplug/replug the keyboard, switch TTYs, sleep/suspend...)
* Keys you then type into swaylock would be handled multiple times

This stores the keyboard and pointer in the swaylock state so they can be released when new capabilities are given.

I had to bump the seat version up to 3 because `wl_pointer_remove` isn't supported on anything earlier.

### DIFF
--- a/include/swaylock/swaylock.h
+++ b/include/swaylock/swaylock.h
@@ -58,6 +58,8 @@ struct swaylock_state {
 	struct wl_compositor *compositor;
 	struct zwlr_layer_shell_v1 *layer_shell;
 	struct zwlr_input_inhibit_manager_v1 *input_inhibit_manager;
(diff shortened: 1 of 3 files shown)


SirCmpwn:
We should also probably update the keymap handler to correctly clean up the old keymap

RyanDwyer:
Is it just me or does it already do that? It calls `xkb_keymap_unref` here: https://github.com/swaywm/sway/blob/master/swaylock/seat.c#L31

SirCmpwn:
dur

SirCmpwn:
Thanks!

PR #2772, opened 2018-10-06
Only damage popups when popups have damage

### LINKED ISSUE #2601: Damage issues with popups
To replicate: open 2 gnome-terminal windows side by side and open the right click menu such that it overlaps. Click on the window you opened the terminal on again to close it. The menu will remain partially visible on the other window until its damaged (e.g. by focus).
Also reproducible on transmission-gtk popups (the black ones when hovering over files or clients).

RyanDwyer:
The previous behaviour was to damage the entire view, which would recurse into each popup. This patch makes it damage only the popup's surface, and respect the surface damage given by the client.

This adds listeners to the popup's map and unmap events rather than doing the damage in the create and destroy functions. To get the popup's position relative to the view, a new `child_impl` function `get_root_coords` has been introduced, which traverses up the parents.

Fixes #2601

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -20,6 +20,19 @@
 
 static const struct sway_view_child_impl popup_impl;
 
+static void popup_get_root_coords(struct sway_view_child *child,
+		int *root_sx, int *root_sy) {
+	struct sway_xdg_popup *popup = (struct sway_xdg_popup *)child;
+	struct wlr_xdg_surface *surface = popup->wlr_xdg_surface;
+	*root_sx = -surface->geometry.x;
+	*root_sy = -surface->geometry.y;
+	while (surface && surface->role == WLR_XDG_SURFACE_ROLE_POPUP) {
+		*root_sx += surface->popup->geometry.x;
+		*root_sy += surface->popup->geometry.y;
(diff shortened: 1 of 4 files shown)


sway/desktop/xdg_shell.c
```diff
@@ -20,6 +20,19 @@
 
 static const struct sway_view_child_impl popup_impl;
 
+static void popup_get_root_coords(struct sway_view_child *child,
```
emersion: There's already a wlroots function that does just that.

sway/tree/view.c
```diff
@@ -639,12 +640,18 @@ static void view_subsurface_create(struct sway_view *view,
 	view_child_init(child, NULL, view, subsurface->surface);
 }
 
+static void view_child_damage(struct sway_view_child *child, bool whole) {
+	int sx, sy;
+	child->impl->get_root_coords(child, &sx, &sy);
```
emersion: What about subsurfaces?

RyanDwyer: Yeah, it segfaults. I should have checked that :\

RyanDwyer: Should be good now.

SirCmpwn:
Thanks!

Issue #1864, opened 2018-04-26
Documentation still states to use sway_log, but wlr_log should be used


gabm:
its still present in the following files

```
$ ~/d/o/sway (master)> ack sway_log
CONTRIBUTING.md
110:    sway_log(L_INFO, "Using output %d of %d", desired_output, registry->outputs->length);

swaygrab/main.c
59:             sway_log(L_ERROR, "Swaygrab failed to fork.");
70:                     sway_log(L_ERROR, "Could not fdup the pipe");
74:             sway_log(L_ERROR, "Swaygrab could not run convert.");
82:             sway_log(L_ERROR, "Raw capture data is not yet supported. Proceeding with ffmpeg normally.");

HACKING.md
3:Use `sway_log(importance, fmt, ...)` to log. The following importances are
10:`sway_log` is a macro that calls `_sway_log` with the current filename and line
```

I'll provide a PR..

ianyfan:
Closed by 85961f63bfe922831011f75860b3acde3d890a9f

PR #2775, opened 2018-10-06
Remove last vestiges of sway_log

ianyfan:
Closes https://github.com/swaywm/sway/issues/1864

Personally, I think the wlroots CONTRIBUTING.md is better, maybe this one should be updated with some information from there.


SirCmpwn:
>Personally, I think the wlroots CONTRIBUTING.md is better, maybe this one should be updated with some information from there.

Good call, I'll update it separately.

PR #2776, opened 2018-10-06
Fix swaylock w/shadow on glibc, improve security

SirCmpwn:
Today I learned that GNU flaunts the POSIX standard in yet another
creative way. Additionally, this adds some security improvements,
namely:

- Zeroing out password buffers in the privileged child process
- setuid/setgid after reading /etc/shadow


emersion:
We could also consider:

* Ensure that the password buffer isn't written to corefiles
* Ensure that the password buffer isn't swapped on disk

See https://wiki.sei.cmu.edu/confluence/display/c/MEM06-C.+Ensure+that+sensitive+data+is+not+written+out+to+disk

SirCmpwn:
setuid programs will already not generate a coredump under normal conditions

emersion:
Thanks!

Issue #1883, opened 2018-05-01
Remove output_damage_view(whole=true)


emersion:
Overlaps with `output_damage_whole_container`.

SirCmpwn:
Is this still relevant?

PR #2777, opened 2018-10-06
output: clean up output_damage_view

ianyfan:
Since it was only used once, its logic was moved to output_damage_from_view,
and the iterator changed since the variable 'whole' is now always false.
Also removed unused function output_damage_whole_container_iterator.

Closes https://github.com/swaywm/sway/issues/1883 though not sure if this was what it was going for.

Maybe `output_damage_from_view` should now be renamed to `output_damage_view`?


ianyfan:
My bad, I'm still not grokking this part of the codebase.

PR #2778, opened 2018-10-06
swaybar: fix binding to wl_pointer multiple times

### LINKED ISSUE #2741: swaybar: after a while, a click causes multiple events
I'm assuming that this is due to hotspots not being removed properly, but I can't reliably reproduce it yet.

emersion:
To reproduce, suspend multiple times your computer and try using your mouse wheel on swaybar to switch between workspaces. It switches multiple times.

Fixes #2741

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -283,6 +283,10 @@ struct wl_pointer_listener pointer_listener = {
 static void seat_handle_capabilities(void *data, struct wl_seat *wl_seat,
 		enum wl_seat_capability caps) {
 	struct swaybar *bar = data;
+	if (bar->pointer.pointer != NULL) {
+		wl_pointer_release(bar->pointer.pointer);
+		bar->pointer.pointer = NULL;
+	}
 	if ((caps & WL_SEAT_CAPABILITY_POINTER)) {
 		bar->pointer.pointer = wl_seat_get_pointer(wl_seat);
 		wl_pointer_add_listener(bar->pointer.pointer, &pointer_listener, bar);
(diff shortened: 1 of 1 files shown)


ianyfan:
Does this fix https://github.com/swaywm/sway/issues/2741? It seems like it should.

emersion:
Oh, yeah probably. Can you confirm?

ianyfan:
I'll test it

ianyfan:
You're releasing the old pointer event if `seat_handle_capabilities` is called for a non-pointer. I think you need to move that block into the next block.

emersion:
I don't get it. `seat_handle_capabilities` isn't called "for a non-pointer". The `capabilities` event is sent whenever the seat changes its capabilities, e.g. when a pointer (mouse, touchpad) is connected or disconnected. If the seat still has the "pointer" capability, the pointer will get re-created right away.

ianyfan:
Might have something to do with having multiple seats, perhaps? If I add some logging, I get this at startup:

```
2018-10-06 18:38:13 - [swaybar/bar.c:292] adding pointer for 0x607000001670
2018-10-06 18:38:13 - [swaybar/bar.c:287] releasing pointer for 0x6070000016e0
```
which results in there being no pointer at the start.
(just used pointer values for the seats because I'm not sure what other property to use)

emersion:
Yeah, multiseat isn't handled by swaybar, since we use a global `bar->pointer.pointer`. We should use a per-seat state instead. This is a separate issue though.

SirCmpwn:
Thanks!

PR #2779, opened 2018-10-06
ipc: set "type" of floating containers to "floating_con"

ianyfan:
Only just noticed this detail in i3:

> type (string)
>    Type of this container. Can be one of "root", "output", "con", "floating_con", "workspace" or "dockarea".


emersion:
Thanks!

PR #2785, opened 2018-10-07
Fix crash on reload when keyboard is configured

ianyfan:
If the keyboard was configured, it would be destroyed when reloading.
This meant that setting the repeat binding on the destroyed keyboard
would cause a segmentation fault. This commit sets the repeat binding
before running the command.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -263,8 +263,10 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
 		keyboard->held_binding = binding_released;
 	}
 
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Quick question before I look at this. When the reload command is issued, would it better to schedule an idle event in the Wayland event loop and do the reload there? Then we don't have to make all these hacks.

SirCmpwn:
That sounds like a great idea.

ianyfan:
Superseded by https://github.com/swaywm/sway/pull/2789

PR #2789, opened 2018-10-08
Reload config using idle event

RyanDwyer:
This patch makes it so when you run `reload`, the actual reloading is deferred to the next time the event loop becomes idle. This avoids several use-after-frees and removes the workarounds we have to avoid them.

When you run `reload`, we validate the config before creating the idle event. This is so the reload command will still return an error if there are validation errors. To allow this, `load_main_config` has been adjusted so it doesn't apply the config if `validating` is true rather than applying it unconditionally.

This also fixes a memory leak in the reload command where if the config failed to load, the `bar_ids` list would not be freed.

To test:

* Change config and reload, verify new config applied
* Hold the binding for reload
* Introduce a validation error and reload your config, eg. by renaming your config file to something else before reloading (it appears that command validation errors are not reported by swaymsg, even in master)

### DIFF
--- a/sway/commands/reload.c
+++ b/sway/commands/reload.c
@@ -3,15 +3,12 @@
 #include "sway/commands.h"
 #include "sway/config.h"
 #include "sway/ipc-server.h"
+#include "sway/server.h"
 #include "sway/tree/arrange.h"
 #include "list.h"
+#include "log.h"
 
-struct cmd_results *cmd_reload(int argc, char **argv) {
-	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "reload", EXPECTED_EQUAL_TO, 0))) {
-		return error;
-	}
-
+static void do_reload(void *data) {
 	// store bar ids to check against new bars for barconfig_update events
 	list_t *bar_ids = create_list();
 	for (int i = 0; i < config->bars->length; ++i) {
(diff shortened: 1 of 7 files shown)


sway/commands/reload.c
```diff
@@ -37,12 +37,26 @@ struct cmd_results *cmd_reload(int argc, char **argv) {
...
 	arrange_root();
+}
+
+struct cmd_results *cmd_reload(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "reload", EXPECTED_EQUAL_TO, 0))) {
+		return error;
+	}
+
+	if (!load_main_config(config->current_config_path, true, true)) {
```
ianyfan: This ostensibly works, but there is a bug when validating that seat configs are still being applied: it seems that they are not local to a config but global, so it still destroys the keyboard and the crash I was trying to fix before still occurs.

sway/input/keyboard.c
```diff
@@ -278,9 +278,7 @@ static void handle_keyboard_key(struct wl_listener *listener, void *data) {
 				raw_modifiers, false, input_inhibited);
 
 		if (binding_pressed) {
-			if ((binding_pressed->flags & BINDING_RELOAD) == 0) {
-				next_repeat_binding = binding_pressed;
-			}
+			next_repeat_binding = binding_pressed;
```
ianyfan: Since `next_repeat_binding` isn't being set conditionally now, one of these variables can be removed.

sway/commands/reload.c
RyanDwyer: Can you give some example config that produces the crash?

ianyfan: Actually, the problem is more subtle than I made out. It seems that if you have any seat configuration before setting a fallback, like:
```
seat s attach "fake_input"
seat seat0 fallback true
```

instead of the other way round, the first command causes all the devices to be removed from the fallback seat, destroying them, before the next command reattaches them.

My full minimal config is:
```
input "1:1:AT_Translated_Set_2_keyboard" {
    xkb_layout us
    xkb_variant dvorak
}

seat s attach "fake_input"
seat seat0 fallback true

bindsym Mod4+Shift+c reload
bindsym Mod4+Return exec termite
```
but swapping the two aforementioned lines prevents the crash.

RyanDwyer: Should be resolved now.

ianyfan:
Thanks!

Issue #2792, opened 2018-10-08
Sway crashes when setting opacity for current window without windows


johnae:
I found this while I was playing around with an IPC daemon. Can easily be reproduced by just adding this to the config file (assuming no programs are otherwise autostarted I guess):

```
exec swaymsg "opacity 0.75"
```

I guess it tries to set the opacity for something that is null and crashes...

johnae:
In `sway/commands/opacity.c` it doesn't look like it is checking whether `config->handler_context.container` is null though I'm no expert on the codebase at all.

emersion:
Can you open a pull request to fix this?

johnae:
Working on it...

johnae:
Created a pull request. Verified locally that this solves the issue.

PR #2794, opened 2018-10-08
Check if there is a current container before setting its opacity

johnae:
Fixes https://github.com/swaywm/sway/issues/2792

Issuing an opacity command when there is no current container (for example issuing via `exec swaymsg "opacity 0.75"` from the config file) causes sway to crash. Now there is a NULL check which stops this from happening.

### DIFF
--- a/sway/commands/opacity.c
+++ b/sway/commands/opacity.c
@@ -21,6 +21,10 @@ struct cmd_results *cmd_opacity(int argc, char **argv) {
 
 	struct sway_container *con = config->handler_context.container;
 
+	if (con == NULL) {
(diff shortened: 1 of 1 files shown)


RyanDwyer:
This crash could also occur when the workspace is empty and you run it from a key binding (when `handler_context.container` is `NULL` it means a workspace itself is focused).

johnae:
Rebased.

johnae:
@RyanDwyer Yeah that's also the case. I found it when playing with IPC and issuing commands immediately on startup. There may be more bugs in that area, at least that was a "feeling" I had. I'll see if I can find some more.

emersion:
Thanks!

PR #2798, opened 2018-10-08
Implement bar bindsym

### LINKED ISSUE #1521: Feature request: Allow for runtime swaybar bindsym commands
Specifically in my case I wished I could toogle the swaybar height between 0 and 17, so to get manual [hide mode](https://github.com/swaywm/sway/issues/2#issuecomment-348697982).

RedSoxFan:
Implements `bar [<bar-id>] bindsym [--release] button<n> <command>`.

Tests:
- `bar bindsym button1 nop` should prevent switching to a workspace when left clicking on the workspace buttons
-  The following should prevent switching the workspace when scrolling up and down on the workspace buttons
   ```
   bar bindsym button4 nop
   bar bindsym button5 nop
   ```
- `bar bindsym --release button1 layout  toggle all` should
   - _(If on a workspace button)_, switch to the workspace clicked (or the auto_back_and_forth workspace) and change it's layout (this tests makes sure it doesn't override the pressed behaviour)
   - _(If not on a workspace button)_, change the layout of the current workspace

Fixes #1521

### DIFF
--- a/sway/commands/bar/bindsym.c
+++ b/sway/commands/bar/bindsym.c
@@ -33,12 +33,12 @@ struct cmd_results *bar_cmd_bindsym(int argc, char **argv) {
 	binding->button = 0;
 	if (strncasecmp(argv[0], "button", strlen("button")) == 0 &&
 			strlen(argv[0]) == strlen("button0")) {
-		binding->button = argv[0][strlen("button")] - '1' + 1;
+		binding->button = argv[0][strlen("button")] - '0';
 	}
-	if (binding->button == 0) {
+	if (binding->button < 1 || binding->button > 9) {
 		free_bar_binding(binding);
 		return cmd_results_new(CMD_FAILURE, "bar bindsym",
-				"Only button<n> is supported");
+				"Only button<1-9> is supported");
 	}
 
 	binding->command = join_args(argv + 1, argc - 1);
--- a/sway/commands/bar/bindsym.c
+++ b/sway/commands/bar/bindsym.c
@@ -1,5 +1,7 @@
+#define _XOPEN_SOURCE 500
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
 #include "sway/commands.h"
 #include "sway/config.h"
 #include "list.h"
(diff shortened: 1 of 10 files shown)


sway/commands/bar/bindsym.c
```diff
@@ -1,11 +1,69 @@
...
+		argv++;
+		argc--;
+	}
+
+	binding->button = 0;
+	if (strncasecmp(argv[0], "button", strlen("button")) == 0 &&
+			strlen(argv[0]) == strlen("button0")) {
+		binding->button = argv[0][strlen("button")] - '1' + 1;
+	}
+	if (binding->button == 0) {
```
ianyfan: Should there be a bounds check here? Like `> 0 && <= 9`? I'm not sure if it's possible to do something funky like `buttonA` which probably shouldn't be allowed.

sway/commands/bar/bindsym.c
```diff
@@ -1,11 +1,69 @@
...
+		return cmd_results_new(CMD_FAILURE, "bar bindsym", "No bar defined.");
+	}
+
+	struct bar_binding *binding = calloc(1, sizeof(struct bar_binding));
+	if (!binding) {
+		return cmd_results_new(CMD_FAILURE, "bar bindsym",
+				"Unable to allocate bar binding");
+	}
+
+	binding->release = false;
```
ianyfan: I think this is technically unnecessary due to `calloc` always initialising it to false, but I guess it doesn't hurt to be explicit.

sway/config/bar.c
```diff
@@ -28,6 +28,16 @@ static void terminate_swaybar(pid_t pid) {
 	}
 }
 
+void free_bar_binding(struct bar_binding *binding) {
+	if (!binding) {
+		return;
+	}
+	if (binding->command) {
```
ianyfan: Check here is unnecessary since freeing a NULL pointer is valid.

sway/config/bar.c
```diff
@@ -39,7 +49,11 @@ void free_bar_config(struct bar_config *bar) {
 	free(bar->status_command);
 	free(bar->font);
 	free(bar->separator_symbol);
-	// TODO: Free mouse bindings
+	while (bar->bindings->length) {
+		struct bar_binding *binding = bar->bindings->items[0];
+		list_del(bar->bindings, 0);
```
ianyfan: This seems unnecessary, just iterate over the list and free each item without deleting.

swaybar/config.c
```diff
@@ -74,6 +77,13 @@ void free_config(struct swaybar_config *config) {
 	free(config->font);
 	free(config->mode);
 	free(config->sep_symbol);
+	while (config->bindings->length) {
+		struct swaybar_binding *binding = config->bindings->items[0];
+		list_del(config->bindings, 0);
```
ianyfan: Make this consistent with the way bar bindings are freed in the other place.

swaybar/bar.c
```diff
@@ -144,6 +144,22 @@ static void wl_pointer_motion(void *data, struct wl_pointer *wl_pointer,
 	bar->pointer.y = wl_fixed_to_int(surface_y);
 }
 
+static bool check_bindings(struct swaybar *bar, uint32_t x11_button,
+		uint32_t state) {
+	bool released = state == WL_POINTER_BUTTON_STATE_RELEASED;
+	for (int i = 0; i < bar->config->bindings->length; i++) {
+		struct swaybar_binding *binding = bar->config->bindings->items[i];
+		wlr_log(WLR_DEBUG, "Checking [%u, %d] against [%u, %d, %s]",
```
ianyfan: Put the command in quotes. Also maybe add the word 'binding' into the message?

Though I'm not sure if I agree with having a debug message here. It might be better to only log if the binding is actually found.

swaybar/bar.c
```diff
@@ -180,6 +201,11 @@ static void wl_pointer_axis(void *data, struct wl_pointer *wl_pointer,
 		return;
 	}
 
+	if (check_bindings(bar, wl_axis_to_x11_button(axis, value),
+				WL_POINTER_BUTTON_STATE_PRESSED)) {
```
ianyfan: This should send both a pressed event and released event. (at least, that seems to be the behaviour in i3)

ianyfan:
Closes https://github.com/swaywm/sway/issues/1521

RedSoxFan:
> > Closes #1521
>
> I don't think it does. I think all that's missing is a bar config update ipc event though. I'll add that real quick

Actually swaybar doesn't currently subscribe to `barconfig_update` so separate it from this one

ianyfan:
https://github.com/swaywm/sway/pull/2751 adds that, so I guess wait until that is merged.

ianyfan:
Actually, I'm not sure whether that PR is asking for bindings to be added at runtime, it's not very clear. In any case, it's so old that I don't actually think it's a problem to close it, and to open a new PR with  more specific request.

ianyfan:
Should we allow a way to not skip default behaviour? Just a suggestion.

RedSoxFan:
> Actually, I'm not sure whether that PR is asking for bindings to be added at runtime, it's not very clear. In any case, it's so old that I don't actually think it's a problem to close it, and to open a new PR with more specific request.

Actually, ya. That doesn't really seem related. That might be better closed by #2751 though

> Should we allow a way to not skip default behaviour? Just a suggestion.

Do you mean having both the bindsym and default behaviour get triggered? You could just use `--release` for bindsyms and leave the default behaviour on pressed. I guess a `--before-pressed` could be added to have something execute before the default behaviour.

SirCmpwn:
Thanks!

PR #2805, opened 2018-10-09
resize: Determine if anything changed using before/after check

RyanDwyer:
Returning a boolean from `container_resize_tiled` and `resize_tiled` doesn't work in all cases. This patch changes it back to void and does a before/after check to see if the container was resized.

See https://github.com/swaywm/sway/pull/2729#issuecomment-428155420


SirCmpwn:
Thanks!

PR #2809, opened 2018-10-09
Fix undesirable height change of floating views

mwenzkowski:
In view_autoconfigure the height of the view is adjusted if the parent
container has a tabbed/stacked layout. Previously this height change
would also be applied to floating views, although it is not needed for
them.

To reproduce the problem:
1.  Open a floating window with a Sway titlebar (SSD).
2.  Change the workspace layout from split to tabbed/stacked.
3. In case of a tabbed workspace layout the height of the floating window
    is reduced by the height of the titlebar.
4. In case of a stacked workspace layout the height of the floating window
    is reduced by  (height of titlebar) * (number of floating windows on this
    workspace). If there are more than one floating windows on the same
    workspace, there is a visible gap between the titlebar and the content of
    the floating window.


SirCmpwn:
Thanks!

PR #2811, opened 2018-10-10
Fix floating click events

### LINKED ISSUE #2801: Floating titlebars are not draggable anymore
1. Open a floating window with a Sway titlebar (SSD)
2. Move mouse over the titlebar
3. Focus goes to the app underneath if `focus_follow_mouse` is enabled
4. It's not possible to drag the floating window

RyanDwyer:
* Set focus to a floating container when clicking its title bar.
* Raise floating when user clicks title bar or decorations (in the `seat_begin` functions).
* In `container_at`, it only returned a floating container if the user had clicked the surface. This makes it use `floating_container_at` instead.

I have tested using `focus_follows_mouse` and `raise_floating` and it seems to work correctly.

Fixes #2801.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -877,6 +877,7 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 			while (cont->parent) {
 				cont = cont->parent;
 			}
+			seat_set_focus_container(seat, cont);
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Thanks!

PR #2812, opened 2018-10-10
Fix program name in version strings

RyanDwyer:
When running `swaymsg -v`, the version returned is actually the version of swaymsg itself, yet the message displayed was `sway version <version>`. This can create confusion if users update sway and swaymsg but don't restart sway, then use swaymsg to check the version.

This patch changes the wording to be `swaymsg version <version>` instead, and likewise for swaybar.

To get the version of a running sway instance, users should run `swaymsg -t get_version`.


emersion:
Thanks!

PR #2814, opened 2018-10-10
Add mouse_warping container

### LINKED ISSUE #2577: FR: option to move cursor to the center of window when focus changes
I've had such behavior in xmonad and was very happy with it:

1) There's no need to remember where cursor position is
2) It's very handy on floating window
3) In most of cases it reduces path to target that should be clicked

Unfortunately, Sway don't allows something like that.

As far as I see, `i3` allows it, but only when output changes: this option is called `mouse_warping`. 

There's also a [blog post](http://pushl.net/blog/18/) that describes how to script `i3` to do it. But that requires xdotool, which can't move cursor under Wayland.

-----

Anyway, it seems to be a missing `focus` value for `mouse_warping` option.

Would you consider to implement it?

Emantor:
This option always moves the cursor into the middle of the container if the warp
variable is true in seat_set_focus_warp.

Fixes #2577

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -325,6 +325,12 @@ enum focus_wrapping_mode {
 	WRAP_FORCE
 };
 
+enum mouse_warping_mode {
+	WARP_NO,
+	WARP_OUTPUT,
+	WARP_CONTAINER
+};
+
 /**
  * The configuration struct. The result of loading a config file.
  */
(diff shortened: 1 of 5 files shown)


Emantor:
This does not work after refactoring, looking into it…

Emantor:
Fixed, I just can't read.

SirCmpwn:
Thanks!

Issue #2815, opened 2018-10-10
Switching workspaces with focus_follows_mouse focuses the wrong window


SirCmpwn:
It focuses the one underneath the mouse, not the focus_inactive window.

tarmack:
Would this be as easy as just setting the `allow_refocusing` to false in the `workspace_switch()` function in `commands/workspace.c`? It does at least fix it somewhat.

I'll be looking for strange side effects from this change..

SirCmpwn:
Probably

tarmack:
Yeah no, it is not that easy. It seems that this issue is one in a long line of mouse focus issues (such as: [2401](https://github.com/swaywm/sway/issues/2401) and [2558](https://github.com/swaywm/sway/issues/2558)). I noticed that if you move a tiling window away from under the pointer, scrolling events keep going to the focused window (now no longer under the cursor). But as soon as the cursor moves a bit the focus jumps to the window now under the mouse.

I get the feeling that the easiest way to make the mouse focus issues go away completely is to move the mouse pointer over to the (center of) a newly focused window if it is not over that window already. This is not ideal however because users might get confused about the location of the cursor.
Another option is to make focus changes from the mouse movement only trigger when a window/container is actually entered (pointer is actually moving from outside the container across the border). But this may get hard to reliably implement.

I'm not sure what would be the best way forward but I'm confident that some discussion here can point us in the right direction.

SirCmpwn:
>Another option is to make focus changes from the mouse movement only trigger when a window/container is actually entered (pointer is actually moving from outside the container across the border). But this may get hard to reliably implement.

This might be a good idea

PR #2820, opened 2018-10-11
Fix mouse warping container

Emantor:
For the drm backend the pointer would warp to 0,0 on window if `mouse_warping container` was set.
Check that the pointer does not warp to 0,0 and move the arrange functions into `view_map`.
This is actually a decrease in code size, if somebody has a fancier idea than passing `fullscreen` and `decoration` as arguments, I'm all for it. The cursor now correctly warps to the middle of the newly created window.

Edit: codestyle has not been fixed thoroughly, will do that when I can remove the [WIP] comment.

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -329,7 +329,8 @@ void view_destroy(struct sway_view *view);
 
 void view_begin_destroy(struct sway_view *view);
 
-void view_map(struct sway_view *view, struct wlr_surface *wlr_surface);
+void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
+	bool fullscreen, bool decoration);
 
 void view_unmap(struct sway_view *view);
 
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -406,26 +406,18 @@ static void handle_map(struct wl_listener *listener, void *data) {
 		view->natural_height = view->wlr_xdg_surface->surface->current.height;
 	}
 
-	view_map(view, view->wlr_xdg_surface->surface);
+	bool csd = false;
 
 	if (!view->xdg_decoration) {
 		struct sway_server_decoration *deco =
-			decoration_from_surface(xdg_surface->surface);
-		bool csd = !deco || deco->wlr_server_decoration->mode ==
+				decoration_from_surface(xdg_surface->surface);
+		csd = !deco || deco->wlr_server_decoration->mode ==
 			WLR_SERVER_DECORATION_MANAGER_MODE_CLIENT;
-		view_update_csd_from_client(view, csd);
+
 	}
 
-	if (xdg_surface->toplevel->client_pending.fullscreen) {
-		container_set_fullscreen(view->container, true);
-		arrange_workspace(view->container->workspace);
-	} else {
-		if (view->container->parent) {
-			arrange_container(view->container->parent);
-		} else if (view->container->workspace) {
-			arrange_workspace(view->container->workspace);
-		}
-	}
+	view_map(view, view->wlr_xdg_surface->surface,
+		xdg_surface->toplevel->client_pending.fullscreen, csd);
 
 	transaction_commit_dirty();
 
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -9,6 +9,7 @@
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_xcursor_manager.h>
 #include <wlr/types/wlr_idle.h>
+#include <wlr/types/wlr_box.h>
 #include "list.h"
 #include "log.h"
 #include "config.h"
@@ -1271,4 +1272,44 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat) {
 	cursor->cursor = wlr_cursor;
 
 	return cursor;
+
+}
+
+/**
+ * Warps the cursor to the middle of the container argument.
+ * Does nothing if the cursor is already inside the container.
+ * If container is NULL, returns without doing anything.
+ */
+void cursor_warp_to_container(struct sway_cursor *cursor,
+		struct sway_container *container) {
+	if (!container) {
+		return;
+	}
+
+	struct wlr_box box;
+	container_get_box(container, &box);
+	if (wlr_box_contains_point(&box, cursor->cursor->x, cursor->cursor->y)) {
+		return;
+	}
+
+	double x = container->x + container->width / 2.0;
+	double y = container->y + container->height / 2.0;
+
+	wlr_cursor_warp(cursor->cursor, NULL, x, y);
+}
+
+/**
+ * Warps the cursor to the middle of the workspace argument.
+ * If workspace is NULL, returns without doing anything.
+ */
+void cursor_warp_to_workspace(struct sway_cursor *cursor,
+		struct sway_workspace *workspace) {
+	if (!workspace) {
+		return;
+	}
+
+	double x = workspace->x + workspace->width / 2.0;
+	double y = workspace->y + workspace->height / 2.0;
+
+	wlr_cursor_warp(cursor->cursor, NULL, x, y);
 }
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -771,27 +771,18 @@ void seat_set_focus_warp(struct sway_seat *seat, struct sway_node *node,
 		workspace_consider_destroy(last_workspace);
 	}
 
-	if (last_focus) {
-		if (config->mouse_warping && warp &&
-				(new_output != last_output ||
-				config->mouse_warping == WARP_CONTAINER)) {
-			double x = 0;
-			double y = 0;
+	if (last_focus && warp) {
+		if (container && config->mouse_warping == WARP_CONTAINER) {
+			cursor_warp_to_container(seat->cursor, container);
+			cursor_send_pointer_motion(seat->cursor, 0, true);
+		} else if (new_output != last_output &&
+				   config->mouse_warping >= WARP_OUTPUT) {
 			if (container) {
-				x = container->x + container->width / 2.0;
-				y = container->y + container->height / 2.0;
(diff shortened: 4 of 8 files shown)


RyanDwyer:
Can't the 0,0 check in seat.c be removed now that the focus happens after the size has been determined?

RyanDwyer:
When the focused view unmaps, the new cursor position is chosen based on the layout that was there before the container unmapped. For example, create layout `H[view view]`, focus the right view and close it. The cursor will move to the center of the left half.

The fix for this will be in `view_unmap`. Focus is chosen when the unmap signal is emitted. This will need to come after the arrange.

Emantor:
This will be a lot harder to untangle, since the cursor is refocused in `container_begin_destroy`:
```
#0  0x0000555c0236271d in seat_set_focus_warp (seat=0x555c0461fc30, node=0x555c04846090, warp=true, notify=true) at ../sway/input/seat.c:628
#1  0x0000555c02362f6b in seat_set_focus (seat=0x555c0461fc30, node=0x555c04846090) at ../sway/input/seat.c:810
#2  0x0000555c0236143d in handle_seat_node_destroy (listener=0x555c04847460, data=0x555c04862930) at ../sway/input/seat.c:184
#3  0x0000555c02380774 in wl_signal_emit (signal=0x555c04862960, data=0x555c04862930) at /usr/include/wayland-server-core.h:468
#4  0x0000555c02380ac7 in container_begin_destroy (con=0x555c04862930) at ../sway/tree/container.c:90
#5  0x0000555c0238746e in view_unmap (view=0x555c04861ce0) at ../sway/tree/view.c:626
```
And without the destroyed node, the workspace can't be arranged.

RyanDwyer:
I think the best way to do this is to move the focus setting code out of `handle_seat_node_destroy` and into `view_unmap` (or rather, make `view_map` call a new function such as `choose_focus` and put the logic there).

`handle_seat_node_destroy` would then be almost a one-liner which calls `seat_node_destroy`, so those two functions can be merged into one.

Back in `view_unmap`, you'll probably need to determine `needs_new_focus` before calling `container_begin_destroy`, then do the rest afterwards.

Emantor:
I implemented you suggestion. From cursory testing it works, I'm not entirely sure focus is still handled correctly. Please test.

RyanDwyer:
Focus needs to be changed for each seat. This is getting complicated, and I'm beginning to think that warping the cursor a second time is the lesser of two evils. Let me know what you think.

This is out of scope for this PR, but I think `seat_set_focus_warp` needs to refactored in order to fix this properly. Perhaps some of the additional logic like cursor warping could be moved into the `focus` and `workspace` commands. Or perhaps the function should be split into two, where there's a high level function that does all the additional logic and a low level function that only manipulates the focus stack. Then functions like `handle_seat_node_destroy` could call the low level function. Or perhaps a mix of the two. I'll think about this over the next few days.

Emantor:
Warping the cursor a second time also introduced other bugs since we can't simply warp the cursor if the focus was already changed.

To the overall focus vs cursor problem:
I think it maybe overall beneficial to separate focus from the cursor movement. At the moment cursor movement and focus warp are very entangled inside `seat_set_focus_warp`, imo its going to be much clearer when reading the code if the cursor is moved explicitly, instead of in the `seat_set_focus_warp` function. This also makes it possible to configure the cursor move semantics in depth, if that is requested.

My concern is that it may be hard to achieve this regression free without tests. I noticed the [swaywm/sway-tests](https://github.com/swaywm/sway-tests) repository, it may be very useful to have tests for focus and cursor behaviour to cache regressions.

RyanDwyer:
What about creating a `cursor_warp_to_container` function, and calling it from `seat_set_focus_warp` and at the end of `view_unmap`?

Emantor:
Yes, I'll do that when I'm back from work this evening.

include/sway/tree/view.h
```diff
@@ -329,7 +329,7 @@ void view_destroy(struct sway_view *view);
 
 void view_begin_destroy(struct sway_view *view);
 
-void view_map(struct sway_view *view, struct wlr_surface *wlr_surface);
+void view_map(struct sway_view *view, struct wlr_surface *wlr_surface, bool fullscreen, bool decoration);
```
RyanDwyer: This line should be wrapped.

sway/desktop/xdg_shell.c
```diff
@@ -406,27 +406,18 @@ static void handle_map(struct wl_listener *listener, void *data) {
...
 	if (!view->xdg_decoration) {
 		struct sway_server_decoration *deco =
-			decoration_from_surface(xdg_surface->surface);
-		bool csd = !deco || deco->wlr_server_decoration->mode ==
-			WLR_SERVER_DECORATION_MANAGER_MODE_CLIENT;
-		view_update_csd_from_client(view, csd);
-	}
+				decoration_from_surface(xdg_surface->surface);
+		csd = !deco || deco->wlr_server_decoration->mode ==
+									WLR_SERVER_DECORATION_MANAGER_MODE_CLIENT;
```
RyanDwyer: This line shouldn't be indented this much.

sway/desktop/xdg_shell.c
```diff
@@ -406,27 +406,18 @@ static void handle_map(struct wl_listener *listener, void *data) {
...
-		arrange_workspace(view->container->workspace);
-	} else {
-		if (view->container->parent) {
-			arrange_container(view->container->parent);
-		} else if (view->container->workspace) {
-			arrange_workspace(view->container->workspace);
-		}
 	}
 
+	view_map(view, view->wlr_xdg_surface->surface, xdg_surface->toplevel->client_pending.fullscreen, csd);
```
RyanDwyer: Wrap.

sway/input/cursor.c
```diff
@@ -1271,4 +1271,33 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat) {
 	cursor->cursor = wlr_cursor;
 
 	return cursor;
+
+}
+
+void cursor_warp_to_container(struct sway_cursor *cursor,
+								struct sway_container *container){
```
RyanDwyer: Deindent, and add space between `){`.

sway/input/cursor.c
```diff
@@ -1271,4 +1271,33 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat) {
...
+	if(!container) {
```
RyanDwyer: Space before bracket.

sway/input/cursor.c
```diff
@@ -1271,4 +1271,33 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat) {
...
+		return;
+	}
+
+	double x = 0;
+	double y = 0;
+	x = container->x + container->width / 2.0;
```
RyanDwyer: May as well initialise `x` and `y` to these values rather than setting them to zero first.

sway/input/seat.c
```diff
@@ -772,26 +772,17 @@ void seat_set_focus_warp(struct sway_seat *seat, struct sway_node *node,
...
-		if (config->mouse_warping && warp &&
-				(new_output != last_output ||
-				config->mouse_warping == WARP_CONTAINER)) {
-			double x = 0;
-			double y = 0;
+		if (container && config->mouse_warping == WARP_CONTAINER && warp) {
+			cursor_warp_to_container(seat->cursor, container);
+			cursor_send_pointer_motion(seat->cursor, 0, true);
+		} else if (warp && new_output != last_output &&
+				   config->mouse_warping >= WARP_CONTAINER) {
```
RyanDwyer: * I think you intend to use `== WARP_OUTPUT` instead of `>= WARP_CONTAINER`.
* In the warping to output part, it should just warp to the workspace and not care about containers.
* The `warp` condition exist in both statements, so it could be moved into the `if (last_focus)` one (err... why is this even conditional on having a last focus?).

sway/tree/view.c
```diff
@@ -621,7 +637,16 @@ void view_unmap(struct sway_view *view) {
 
 	struct sway_seat *seat;
 	wl_list_for_each(seat, &input_manager->seats, link) {
-		cursor_send_pointer_motion(seat->cursor, 0, true);
+		if (config->mouse_warping == WARP_CONTAINER) {
+			struct sway_node *node = seat_get_focus(seat);
```
RyanDwyer: It is possible for the seat to have no focus, or for focus to be a layer surface or xwayland unmanaged. In these cases `node` will be `NULL` which will produce a crash. You should be able to produce this by making a view unmap while the lockscreen is active.

sway/tree/view.c
```diff
@@ -621,7 +637,16 @@ void view_unmap(struct sway_view *view) {
...
+			if (node->type == N_CONTAINER) {
+				cursor_warp_to_container(seat->cursor, node->sway_container);
+			} else if (node->type == N_WORKSPACE){
```
RyanDwyer: Space between `){`.

sway/tree/view.c
Emantor: I'll add `node &&` to the conditional.

sway/input/seat.c
Emantor: Judging by `git blame`, `last_focus` was used to remove the focus from a view previously. It will probably cause no harm to remove here. What are your thoughts on this @RyanDwyer?

RyanDwyer: Yep, let's remove it. And by remove I mean replace it with `warp`.

Emantor:
Hm, I can't reproduce with your instructions. However I found another bug:
- If `focus_follows_mouse` is set to `no` and you click into a container, the mouse is warped to the middle of the container.

SirCmpwn:
It doesn't seem that this fixes this issue:

https://github.com/swaywm/sway/issues/2815

Were you going to fix these together?

Emantor:
No, I think this focus issue needs to be tackled a different way.
Tracking the relevant container on crossing a border sounds like a good idea, but is out of scope for this PR.

SirCmpwn:
Aight. Well, this patch works. Thanks!

Issue #2727, opened 2018-09-29
Fix or remove edge_gaps


RyanDwyer:
Sway 0.15 introduced a command/syntax `gaps edge_gaps on|off|toggle`. It has the following documentation:

    **gaps** edge_gaps <on|off|toggle>::
        Whether or not to add gaps between views and workspace edges if amount of
        inner gap is not zero. When _off_, no gap is added where the view is aligned to
        the workspace edge, effectively creating gaps only between views. The toggle
        argument cannot be used in the configuration file.

The feature does not exist in i3-gaps.

In the 1.0 alpha, the command exists but doesn't work correctly.

The options I see are:

1. Remove the command.
2. Remove the command and allow setting negative values for `gaps_outer` which allows the user to achieve the same effect.
3. Implement the command.

My preference is 1.

DawidLoubser:
Sway's seamless per-display scaling offers a heavenly solution to my HiDpi 14-inch laptop screen (where I set scale to `1.4`) and my standard DPi, external 4k screen (scale `1.0`).

Literally my only complaint is the ugly gaps between windows, and between a window and its decoration, while on my 1.4-scale laptop screen.

So my vote goes for #3 - implement the command, or at least build default behaviour to collapse sub-pixel gaps to 0, instead of rendering an ugly 1-pixel gap.

RyanDwyer:
What you're describing is not related to `edge_gaps`. That sounds like a bug which should be fixed regardless.

DawidLoubser:
Ah, apologies - I'll open a new thread sometime with screenshots etc.
Thanks for the clarification!

tarmack:
I went the easy route, and don't allow the negative value to move windows outside the container.

The token value of -1 in the `workspace_config` seems to not pose issues but I may not fully understand the reason and working of that piece of the sway code anyway.

PR #2823, opened 2018-10-11
Fix edge gaps

tarmack:
Fixes [Issue: Fix or remove edge_gaps](https://github.com/swaywm/sway/issues/2727) by removing them and allowing negative values for `gaps outer` just the way i3 does it.

While this change allows negative values for the outer gaps to essentially negate `gaps inner` it does not allow to set it lower than the negated `gaps inner` value to prevent windows from ending up outside the container.

This pull request may result in issues with the -1 token value in `workspace_config` struct, if you see this going wrong please give me some pointers on how to go about fixing and testing that.

### DIFF
--- a/sway/commands/gaps.c
+++ b/sway/commands/gaps.c
@@ -20,31 +20,6 @@ struct gaps_data {
 	int amount;
 };
 
-// gaps edge_gaps on|off|toggle
-static struct cmd_results *gaps_edge_gaps(int argc, char **argv) {
-	struct cmd_results *error;
-	if ((error = checkarg(argc, "gaps", EXPECTED_AT_LEAST, 2))) {
-		return error;
-	}
-
-	if (strcmp(argv[1], "on") == 0) {
-		config->edge_gaps = true;
-	} else if (strcmp(argv[1], "off") == 0) {
-		config->edge_gaps = false;
-	} else if (strcmp(argv[1], "toggle") == 0) {
-		if (!config->active) {
-			return cmd_results_new(CMD_INVALID, "gaps",
-					"Cannot toggle gaps while not running.");
-		}
-		config->edge_gaps = !config->edge_gaps;
-	} else {
-		return cmd_results_new(CMD_INVALID, "gaps",
-				"gaps edge_gaps on|off|toggle");
-	}
-	arrange_root();
-	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
-}
-
 // gaps inner|outer <px>
 static struct cmd_results *gaps_set_defaults(int argc, char **argv) {
 	struct cmd_results *error = checkarg(argc, "gaps", EXPECTED_EQUAL_TO, 2);
@@ -68,15 +43,17 @@ static struct cmd_results *gaps_set_defaults(int argc, char **argv) {
 		return cmd_results_new(CMD_INVALID, "gaps",
 				"Expected 'gaps inner|outer <px>'");
 	}
-	if (amount < 0) {
-		amount = 0;
-	}
-
 	if (inner) {
-		config->gaps_inner = amount;
+		config->gaps_inner = (amount >= 0) ? amount : 0;
 	} else {
 		config->gaps_outer = amount;
 	}
+
+	// Prevent negative outer gaps from moving windows out of the workspace.
+	if (config->gaps_outer < -config->gaps_inner) {
+		config->gaps_outer = -config->gaps_inner;
+	}
+
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
 
@@ -95,8 +72,12 @@ static void configure_gaps(struct sway_workspace *ws, void *_data) {
 		*prop -= data->amount;
 		break;
 	}
-	if (*prop < 0) {
-		*prop = 0;
+	// Prevent invalid gaps configurations.
+	if (ws->gaps_inner < 0) {
+		ws->gaps_inner = 0;
+	}
+	if (ws->gaps_outer < -ws->gaps_inner) {
+		ws->gaps_outer = -ws->gaps_inner;
 	}
 	arrange_workspace(ws);
 }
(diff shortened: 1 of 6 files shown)


SirCmpwn:
Thanks for the patch! Can you associate the email you authored these commits with and your GitHub account so they link up nicely?

sway/commands/gaps.c
```diff
@@ -68,15 +43,19 @@ static struct cmd_results *gaps_set_defaults(int argc, char **argv) {
...
-		config->gaps_inner = amount;
+		config->gaps_inner = (amount >= 0) ? amount : 0;
 	} else {
 		config->gaps_outer = amount;
 	}
+
+    // Prevent negative outer gaps from moving windows out of the container.
+    int overflow;
+    overflow = (config->gaps_inner + config->gaps_outer);
+    if (overflow < 0) {
```
RyanDwyer: You could do this without introducing a new variable:

    if (config->gaps_outer < -config->gaps_inner) {
        config->gaps_outer = -config->gaps_inner;
    }

sway/commands/gaps.c
```diff
@@ -68,15 +43,19 @@ static struct cmd_results *gaps_set_defaults(int argc, char **argv) {
...
-	}
-
 	if (inner) {
-		config->gaps_inner = amount;
+		config->gaps_inner = (amount >= 0) ? amount : 0;
 	} else {
 		config->gaps_outer = amount;
 	}
+
+    // Prevent negative outer gaps from moving windows out of the container.
```
RyanDwyer: This comment should probably say workspace instead of container.

RyanDwyer:
From the associated issue:

> I noticed the workspace_config struct introduced here: Commit: Rename workspace_outputs to workspace_configs and fix memory leak138d10d). I was wondering if the eventual intention is to have this replace the current data in the sway_workplace struct or if they are supposed to keep living side by side?

The `workspace_config` serves as stored configuration for a workspace that doesn't exist yet. When a workspace is created, it looks for a `workspace_config` for that workspace and copies the settings into the workspace struct. The `workspace_config` struct is then not used, unless the workspace is closed and recreated.

Other feedback:

* The -1 in the `workspace_config` is taking effect. You should initialise it to `INT_MIN`, then check for `INT_MIN` when it is copied into the workspace when the workspace is created.
* When using something like `gaps outer current|all set -10` to change a workspace's outer gaps at runtime, the value is set to 0 (see `configure_gaps` function).
* When using `workspace <ws> gaps outer -10`, a check needs to be done here too. This syntax is intended to be used in configuration files so you can configure a single workspace's gaps before it exists.
* There is still documentation for `edge_gaps`, but you probably didn't find it because it's written as `edge\_gaps`.
* Consider updating the gaps documentation to say that negative outer gaps are allowed.

sway/commands/gaps.c
tarmack: Done

sway/commands/gaps.c
tarmack: Done

sway/commands/gaps.c
```diff
@@ -68,15 +43,17 @@ static struct cmd_results *gaps_set_defaults(int argc, char **argv) {
...
-
 	if (inner) {
-		config->gaps_inner = amount;
+		config->gaps_inner = (amount >= 0) ? amount : 0;
 	} else {
 		config->gaps_outer = amount;
 	}
+
+    // Prevent negative outer gaps from moving windows out of the workspace.
+    if (config->gaps_outer < -config->gaps_inner) {
```
RedSoxFan: Style: Please use tabs for indentation

sway/commands/gaps.c
```diff
@@ -95,8 +72,12 @@ static void configure_gaps(struct sway_workspace *ws, void *_data) {
 		*prop -= data->amount;
 		break;
 	}
-	if (*prop < 0) {
-		*prop = 0;
+    // Prevent invalid gaps configurations.
+    if (ws->gaps_inner < 0) {
```
RedSoxFan: Style: Please use tabs for indentation

sway/commands/workspace.c
```diff
@@ -94,7 +95,16 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 			return cmd_results_new(CMD_FAILURE, "workspace gaps",
 					"Expected 'workspace <ws> gaps inner|outer <px>'");
 		}
-		*prop = val >= 0 ? val : 0;
+        *prop = val;
```
RedSoxFan: Style: Please use tabs for indentation

sway/commands/gaps.c
```diff
@@ -68,15 +43,17 @@ static struct cmd_results *gaps_set_defaults(int argc, char **argv) {
...
 	if (inner) {
-		config->gaps_inner = amount;
+		config->gaps_inner = (amount >= 0) ? amount : 0;
 	} else {
 		config->gaps_outer = amount;
 	}
+
+	// Prevent negative outer gaps from moving windows out of the workspace.
+	if (config->gaps_outer < -config->gaps_inner) {
+	    config->gaps_outer = -config->gaps_inner;
```
RedSoxFan: The second level indentation here is still spaces

sway/commands/workspace.c
```diff
@@ -94,7 +95,16 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 			return cmd_results_new(CMD_FAILURE, "workspace gaps",
 					"Expected 'workspace <ws> gaps inner|outer <px>'");
 		}
-		*prop = val >= 0 ? val : 0;
+		*prop = val;
+
+		// Prevent invalid gaps configurations.
+		if (wsc->gaps_inner < 0) {
+		    wsc->gaps_inner = 0;
```
RedSoxFan: The third level of indentation here is still spaces

sway/commands/workspace.c
```diff
@@ -94,7 +95,16 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
...
+		}
+		if (wsc->gaps_outer < -wsc->gaps_inner) {
+		    wsc->gaps_outer = -wsc->gaps_inner;
```
RedSoxFan: The third level of indentation here is still spaces

tarmack:
Sorry about the spaces, @RedSoxFan. I fixed my editor but didn't quite manage to fix the code..

tarmack:
I kind of fucked up the squash, I'll split the style fixes out again, hold on..

SirCmpwn:
`git rebase -i HEAD~3` then change everything to `fixup` except the first commit. Then you can `git push -f`

tarmack:
Is it customary for the pull requests on Sway to have just the one final commit? If so, maybe it could be added to the contribution guide.

SirCmpwn:
No, but it is desirable for the commits to be neatly arranged.

RedSoxFan:
Thanks!

PR #2825, opened 2018-10-12
Fix pixel leaks when using fractional scaling

### LINKED ISSUE #2813: Fractional scaling produces pixel gaps
For example, set an output's scale to 1.4 and create some views. There are pixel gaps in the title bar and between the view surface and border decorations.

RyanDwyer:
The basic idea here is to apply rounding after scaling. It's not as simple as this, though, and I've detailed it in the comments for a function.

In order to fix some pixel leaks in the title bar, I found it easier to change how we place rectangles to fill the area. Instead of placing two rectangles across the full width above and below the title and having shorter rectangles in the inner area, it's now pieced together in vertical chunks. This method involves drawing two less rectangles per container.

I've done some pretty thorough testing, including:

* Testing with scale 1, scale 2, scale 1.4
* Testing with semi-transparent containers to make sure we're not rendering over the same pixels twice
* Testing with output positions that are not 0,0
* Testing with containers in different parts of the screen (to clarify: some pixel gaps would appear in one spot for one container but not for another container due to its position within the output and rounding).
* Testing with tabbed and stacking layouts
* Testing with marks
* Testing with titles that are truncated due to overflowing the title bar width
* Testing with titles that are truncated due to overflowing into the mark texture

Fixes #2813

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -223,11 +223,15 @@ void output_drag_icons_for_each_surface(struct sway_output *output,
 	}
 }
 
+static int scale_length(int length, int offset, float scale) {
+	return round((offset + length) * scale) - round(offset * scale);
+}
+
 static void scale_box(struct wlr_box *box, float scale) {
-	box->x *= scale;
-	box->y *= scale;
-	box->width *= scale;
-	box->height *= scale;
+	box->width = scale_length(box->width, box->x, scale);
+	box->height = scale_length(box->height, box->y, scale);
+	box->x = round(box->x * scale);
+	box->y = round(box->y * scale);
 }
 
 struct sway_workspace *output_get_active_workspace(struct sway_output *output) {
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -33,11 +33,27 @@ struct render_data {
 	float alpha;
 };
 
+/**
+ * Apply scale to a width or height.
+ *
+ * One does not simply multiply the width by the scale. We allow fractional
+ * scaling, which means the resulting scaled width might be a decimal.
+ * So we round it.
+ *
+ * But even this can produce undesirable results depending on the X or Y offset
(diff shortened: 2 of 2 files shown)


RyanDwyer:
Can you provide a screenshot?

ianyfan:
By "the bar" I meant swaybar, not sure why I was being so ambiguous. I'll get a screenshot.

ianyfan:
screenshot

grim isn't capturing a row of grey pixels at the bottom. Looks like it's a problem with swaybar not dealing with scaling properly.

RyanDwyer:
There's a pixel gap on the right edge as well. I think that might be side effect of running it in a nested session (I assume you are?). The output is 1920px wide. 1920 / 1.4 is 1371.4 so the Wayland backend reports a width of 1371. So within sway the container's width is 1371, scaled by 1.4 becomes 1919.4, which rounds down to 1919 and leaves a gap.

ianyfan:
No, it's DRM

RyanDwyer:
I think the same situation applies though. I can see three ways to fix it, none of which I'm fond of:

1. Support `float` output modes, `float` wlr_box coordinates and `float` container/workspace/output coordinates and sizes.
2. Put an `if` statement in `scale_box` so it bumps it by one pixel if it's near the far edge.
3. When a scale is set, take what the user asked for and adjust it so it becomes evenly divisible by both the output's width and height.

My preference is to not fix this. Fractional scaling is going to have inconsistent rendering anyway - for example, a 1px border might take up one pixel in one place and 2px in another.

SirCmpwn:
>When a scale is set, take what the user asked for and adjust it so it becomes evenly divisible by both the output's width and height.

Maybe we can do this calculation and print a warning like "Warning: your scale of %f is going to cause imperfect rendering. Try %f instead if you have issues"

SirCmpwn:
Thanks!

PR #2826, opened 2018-10-13
Implement common event loop for swaybar and swaylock

RyanDwyer:
It was noticed that when you have an output with DPMS off, swaylock waits around forever for the surface on that output to update which means it never verifies the password. To fix this, @RedSoxFan and I determined that we should set a timer to handle this. However, `wl_event_loop_add_timer` appears to be available for the compositor only, not clients.

To resolve this, I took the event loop from swaybar, moved it into `common/`, refactored it a bit and made it handle timers properly. It would previously check the timers every time a Wayland event came through, but now it does it using timerfds so we don't need to wait for a Wayland event.

The following things have been implemented:

* Swaylock's indicator now disappears after 3 seconds.
* Swaylock's password buffer is now cleared after 10 seconds of inactivity.
* Swaylock now sets a 50ms timer for the surfaces to become undirty. If the timer expires it will go ahead and verify anyway. I admit I haven't tested this yet.

If you're reviewing the individual commits, please note that some of the loop code got some fixes in further commits.


ascent12:
Timerfd is a linux-specific feature, meaning this is going to break *BSD support.

RyanDwyer:
Is there a recommended alternative that works on both?

emersion:
It seems that FreeBSD uses an epoll-shim that provides timerfd emulation. It's used for libinput and Wayland.

SirCmpwn:
We need a POSIX base implementation, then a BSD and Linux specific override, then select the right file based on the platform. On the POSIX base implementation we can set a timeout on poll based on the estimated time until the next timer that's due.

SirCmpwn:
If you're not feeling up to it, we can live with just one POSIX implementation.

RyanDwyer:
I made a POSIX compliant implementation using the method you described with the intention of overriding it with timerfds for Linux, but then decided to remove the Linux variant completely and just roll with the POSIX compliant one. It works, and I don't think it's worth maintaining two variants.

My latest commit replaces timerfd. This is ready for review and testing.

common/loop.c
```diff
@@ -0,0 +1,175 @@
...
+	loop->timers = create_list();
+	return loop;
+}
+
+void loop_destroy(struct loop *loop) {
+	list_foreach(loop->fd_events, free);
+	list_foreach(loop->timers, free);
+	list_free(loop->fd_events);
+	list_free(loop->timers);
+	free(loop);
```
ianyfan: Is there a leak here of `loop->fds`?

common/loop.c
```diff
@@ -0,0 +1,175 @@
...
+			struct loop_timer *timer = loop->timers->items[i];
+			int timer_ms = (timer->expiry.tv_sec - now.tv_sec) * 1000;
+			timer_ms += (timer->expiry.tv_nsec - now.tv_nsec) / 1000000;
+			if (timer_ms < ms) {
+				ms = timer_ms;
+			}
+		}
+	}
+
+	poll(loop->fds, loop->fd_length, ms);
```
emersion: There's a race here. If `ms` is negative then `poll` will wait forever. We need to reset `ms` to zero in this case.

common/loop.c
emersion: There's also a leak because elements in the lists aren't free'd.

RyanDwyer: Not any more ;)

common/loop.c
RyanDwyer: Done.

common/loop.c
```diff
@@ -0,0 +1,179 @@
...
+
+bool loop_remove_fd(struct loop *loop, int fd) {
+	for (int i = 0; i < loop->fd_length; ++i) {
+		if (loop->fds[i].fd == fd) {
+			free(loop->fd_events->items[i]);
+			list_del(loop->fd_events, i);
+
+			loop->fd_length--;
+			memmove(&loop->fds[i], &loop->fds[i + 1],
+					sizeof(void*) * (loop->fd_length - i));
```
emersion: Shouldn't this be `sizeof(struct pollfd)`?

include/swaybar/bar.h
```diff
@@ -8,6 +8,8 @@
 struct swaybar_config;
 struct swaybar_output;
 struct swaybar_workspace;
+struct loop;
+struct loop_event;
```
ianyfan: This line seems extraneous

SirCmpwn:
Thanks!

PR #2829, opened 2018-10-14
Fix clear_password_buffer

espkk:
Just a typo that breaks buffer clearing
(I think it could be safely replaced with a memset since there is no succeeding memory freeing)


Emantor:
Yes, unless I'm mistaken `swaylock_password` is contained in `swaylock_state` which is statically allocated before running `main`. A `memset` should be fine by my analysis.

SirCmpwn:
Thanks! If anyone wants to file a follow-up patch with memset that'd be fine.

PR #2835, opened 2018-10-15
Fix crash in swap command

RyanDwyer:
When swapping containers that are in the root of the workspace, `parent` will be `NULL`.

To test:

* Create two workspaces with a terminal on each
* Mark one of them
* On the unmarked terminal, run `swaymsg swap container with mark foo`


SirCmpwn:
Thanks!

PR #2836, opened 2018-10-15
Introduce seat_set_raw_focus and remove notify argument from seat_set_focus_warp

RyanDwyer:
This introduces `seat_set_raw_focus`: a function that manipulates the focus stack without doing any other behaviour whatsoever. There are a few places where this is useful, such as where we set `focus_inactive` followed by another call to set the real focus again. With this change, the `notify` argument to `seat_set_focus_warp` is also removed as these cases now use the raw function instead.

A bonus of this is we are no longer emitting `window::focus` IPC events when setting `focus_inactive`, nor are we sending focus/unfocus events to the surface.

This also fixes the following:

* When running `move workspace to output <name>` and moving the last workspace from the source output, the `workspace::focus` IPC event is no longer emitted for the newly created workspace.
* When splitting the currently focused container, unfocus/focus events will not be sent to the surface when giving `focus_inactive` to the newly created parent, and `window::focus` events will not be emitted.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -543,7 +543,7 @@ static struct cmd_results *cmd_move_container(int argc, char **argv) {
 	if (new_output_last_ws && new_output_last_ws != new_workspace) {
 		struct sway_node *new_output_last_focus =
 			seat_get_focus_inactive(seat, &new_output_last_ws->node);
-		seat_set_focus_warp(seat, new_output_last_focus, false, false);
+		seat_set_raw_focus(seat, new_output_last_focus);
 	}
 
 	// restore focus
@@ -556,7 +556,7 @@ static struct cmd_results *cmd_move_container(int argc, char **argv) {
 			focus = seat_get_focus_inactive(seat, &old_ws->node);
 		}
 	}
-	seat_set_focus_warp(seat, focus, true, false);
+	seat_set_focus(seat, focus);
 
 	// clean-up, destroying parents if the container was the last child
 	if (old_parent) {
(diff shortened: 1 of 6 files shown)


sway/commands/move.c
```diff
@@ -556,7 +556,7 @@ static struct cmd_results *cmd_move_container(int argc, char **argv) {
 			focus = seat_get_focus_inactive(seat, &old_ws->node);
 		}
 	}
-	seat_set_focus_warp(seat, focus, true, false);
+	seat_set_focus(seat, focus);
```
ianyfan: I believe this is causing an extraneous workspace::focus event from the workspace that the container was moved to back to the original workspace.

ianyfan:
Works, though I still find it a bit weird that moving the container changes the focus so that it has to be manually moved back.

ianyfan:
Thanks!

RyanDwyer:
> Works, though I still find it a bit weird that moving the container changes the focus so that it has to be manually moved back.

Say you have a container on workspace 1 and you're going to move it to workspace 2. Before moving, your focus stack is workspace 2 > workspace 1 > container. After moving the container, the focus stack hasn't changed but you need it to be workspace 1 > workspace 2 > container, and you need to emit a `workspace::focus` event too. To do this you'd normally call `seat_set_focus(seat, container)`, but this is a no op because container is already focused. So you have to change the focus to something else then change it back.

RyanDwyer:
I wonder what would break if we removed the "already focused" check...

PR #2838, opened 2018-10-15
Sway clients: Exit gracefully when compositor is unavailable

RyanDwyer:
The previous behaviour was to show an error and then crash. The new behaviour is to show a more helpful error then exit.


SirCmpwn:
Thanks!

PR #2839, opened 2018-10-15
Fix swaylock version string

RyanDwyer:
The referenced constants were not defined so it always printed "version unknown".

Also it would exit with code 1. It now exits with code 0.


SirCmpwn:
Thanks!

PR #2843, opened 2018-10-15
swaynag: Truncate message to 2048 chars.

### LINKED ISSUE #2842: swaynag: Message buffer overflow.
A buffer overflow can be caused by passing a long enough message (`-m`) to swaynag. 

Example:
Repeating `swaywm` 1000 times and passing it with `-m` will cause `[buffer overflow]` to display. Run `"swaywm".repeat(1000)` in the JavaScript console to get this string.

Version:
```
swaynag version 1.0-alpha.6-264-g56c388b5 (Oct 15 2018, branch 'master')
```

c-edw:
This PR fixes #2842. I think this should work since `vsnprintf` truncates the `va_list` to the size specified in the second parameter. I'm probably wrong on this, so please let me know. Thanks.

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -10,6 +10,9 @@
 #include "log.h"
 #include "stringop.h"
 
+static const char *overflow = "[buffer overflow]";
+static const int max_chars = 16384;
+
 size_t escape_markup_text(const char *src, char *dest) {
 	size_t length = 0;
 	if (dest) {
@@ -84,12 +87,12 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
 
 void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
 		int *baseline, double scale, bool markup, const char *fmt, ...) {
-	static char buf[2048];
+	char buf[max_chars];
 
 	va_list args;
 	va_start(args, fmt);
-	if (vsnprintf(buf, 2048, fmt, args) >= 2048) {
-		strcpy(buf, "[buffer overflow]");
+	if (vsnprintf(buf, sizeof(buf), fmt, args) >= max_chars) {
+		strcpy(&buf[sizeof(buf) - sizeof(overflow)], overflow);
 	}
 	va_end(args);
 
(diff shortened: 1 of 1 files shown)


SirCmpwn:
If you do that I'd rather see a [...] appended to the end

c-edw:
I have reduced the truncation limit to 128, since 2048 characters would probably never fit on the screen for anyone. It will append `...` if the message is over 128 characters.

SirCmpwn:
That's not gonna fly. My screen can easily fit 128 characters. At time of writing my statusbar is 122 characters long and only taking up half the screen.

c-edw:
As discussed, this will append `[buffer overflow]` instead of overwriting.

common/pango.c
```diff
@@ -10,6 +10,9 @@
 #include "log.h"
 #include "stringop.h"
 
+#define MAX_CHARS 2048
+#define OVERFLOW "[buffer overflow]"
```
SirCmpwn: Prefer static const over macros

Also, let's go with a bigger maximum, how about 16K

common/pango.c
```diff
@@ -10,6 +10,9 @@
 #include "log.h"
 #include "stringop.h"
 
+static const char *OVERFLOW = "[buffer overflow]";
+static const int MAX_CHARS = 16384;
```
SirCmpwn: lowercase

common/pango.c
```diff
@@ -84,12 +87,12 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
...
 		int *baseline, double scale, bool markup, const char *fmt, ...) {
-	static char buf[2048];
+	char buf[max_chars + sizeof(overflow)];
 
 	va_list args;
 	va_start(args, fmt);
-	if (vsnprintf(buf, 2048, fmt, args) >= 2048) {
-		strcpy(buf, "[buffer overflow]");
+	if (vsnprintf(buf, max_chars, fmt, args) >= max_chars) {
+		strcat(buf, overflow);
```
SirCmpwn: Sorry to pick nits here, but can you make the size of the buffer `max_chars`, then strcpy the overflow text to the end? Something like this:

```c
char buf[max_chars];
// ...
strcpy(&buf[sizeof(buf) - sizeof(overflow)], overflow);
```

Use sizeof(buf) as well for vsnprintf.

SirCmpwn:
Once you address the last issue, can you also squash your commits together? Thanks for your patience.

RedSoxFan:
Thanks!

PR #2845, opened 2018-10-15
common/loop.c: add _POSIX_C_SOURCE for clock_gettime and CLOCK_MONOTONIC

colemickens:
I was getting build errors on my system/toolchain due to an implicit declaration for `clock_gettime` and `CLOCK_MONOTONIC`.

Googling indicated that defining `_POSIX_C_SOURCE` should resolve it. I've used the same value that are used in other source files in the sway source.

This seems to resolve the problem.

### DIFF
--- a/common/loop.c
+++ b/common/loop.c
@@ -1,3 +1,4 @@
+#define _POSIX_C_SOURCE 199309L
 #include <limits.h>
 #include <string.h>
 #include <stdbool.h>


SirCmpwn:
You should set it to `199309L` in this case. Our policy is to use the minimum required POSIX version for each file. You can find the minimum by consulting the `clock_gettime` man page.

colemickens:
@SirCmpwn I see, thanks. Updated, tested, pushed.

SirCmpwn:
Thanks! For more information about feature test macros, check out `man 7 feature_test_macros`

PR #2849, opened 2018-10-16
Focus workspace when swaybar is clicked

ascent12:
This matches i3's behaviour.

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -170,6 +170,14 @@ static void wl_pointer_button(void *data, struct wl_pointer *wl_pointer,
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


RyanDwyer:
It's for when you have multiple outputs and you click on a blank part of swaybar on an output which doesn't have focus. On i3 it focuses that output/workspace.

ianyfan:
Okay. Should this be put after the `check_bindings` block so that it can be overridden by bindings?

Also, would this send out an extraneous event if the user clicks on a hotspot? Not sure what i3 does.

SirCmpwn:
bump

PR #2858, opened 2018-10-17
Fix moving tiled containers to workspaces which only have floating views

### LINKED ISSUE #2857: System freeze when combining floating and tiling windows into one workspace.
This bug causes my system to freeze completely, which I believe is from high CPU usage. I have tested it using a stock Sway config.

Version (this is a branch on my fork but it's only a few commits behind):
```
sway version 1.0-alpha.6-277-gcfc533aa (Oct 17 2018, branch 'feature/2842_TruncateMessage')
```

Steps to reproduce:
1. Start Sway
2. Enter workspace 1
3. Spawn two terminals using`$mod+Enter`
4. Make both terminals floating by pressing `$mod+Shift+Space`
5. Enter workspace 2
6. Spawn one terminal using `$mod+Enter` 
7. Move the terminal into workspace 1 by pressing `$mod+Shift+1`
8. Enter workspace 1
9. Observe that the system will become unresponsive

Notes:
- It doesn't matter which application you spawn. You just need two floating and one tiling window.

Log [here](https://gist.github.com/c-edw/e7158654756f206f33fa74b51381867d). It looks like some of the keystrokes I issued after the freeze still ran.

RyanDwyer:
* Make a workspace which only contains floating views
* Switch to another workspace and create a tiled view
* Move the tiled view to the workspace with `move container to workspace N`

The container would be added as a sibling to the floating view, which makes the container floating while having the geometry of a tiled container.

This changes it so it only looks for tiled containers in the workspace with a fallback to the workspace itself.

Probably fixes #2857 but my symptoms are different to the ones reported in that issue. @c-edw Can you please test if this fixes your issue?

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -483,7 +483,8 @@ static struct cmd_results *cmd_move_container(int argc, char **argv) {
 			ws = workspace_create(NULL, ws_name);
 		}
 		free(ws_name);
-		destination = seat_get_focus_inactive(seat, &ws->node);
+		struct sway_container *dst = seat_get_focus_inactive_tiling(seat, ws);
+		destination = dst ? &dst->node : &ws->node;
 	} else if (strcasecmp(argv[1], "output") == 0) {
 		struct sway_output *new_output = output_in_direction(argv[2],
 				old_output, container->x, container->y);


c-edw:
Yes, it fixes #2857. There's a new odd behaviour I've noticed but I will open another ticket for that shortly.

SirCmpwn:
Thanks!

PR #2861, opened 2018-10-17
Fix crashes when running certain commands on an empty workspace

RyanDwyer:
This fixes crashes when running the `border`, `mark`, `unmark` and `title_format` commands on an empty workspace.


SirCmpwn:
Thanks!

PR #2864, opened 2018-10-17
FreeBSD fixes

sghctoma:
This PR fixes two issues on FreeBSD:

1. `CLOCK_MONOTONIC` appeared in IEEE Std. 1003.1-200x, it was not part of POSIX.1b (the 1993 version). FreeBSD treats it accordingly, thus `_POSIX_C_SOURCE` needs to be set to `200112L`.
2. PR #2855 basically hardcodes the config file path to /etc, which is a problem on e.g. FreeBSD, where the expected path for config files of non-base software is '/usr/local/etc'. Meson sets `sysconfdir` to '/etc' explicitly *only* when `prefix` is '/usr', so it is still possible to use '/usr/local' as `prefix`, and install the config files under '/usr/local/etc'. This commit allows to do that by setting `sysconfdir` based on whether `prefix` is '/usr', or not.

### DIFF
--- a/common/loop.c
+++ b/common/loop.c
@@ -1,4 +1,4 @@
-#define _POSIX_C_SOURCE 199309L
+#define _POSIX_C_SOURCE 200112L
 #include <limits.h>
 #include <string.h>
 #include <stdbool.h>
--- a/meson.build
+++ b/meson.build
@@ -114,7 +114,13 @@ if scdoc.found()
 	endforeach
 endif
 
-add_project_arguments('-DSYSCONFDIR="/@0@"'.format(sysconfdir), language : 'c')
+# If prefix is '/usr', sysconfdir will be explicitly set to '/etc' by Meson to
+# enforce FHS compliance, so we should look for configs there as well.
+if prefix == '/usr'
+	add_project_arguments('-DSYSCONFDIR="/@0@"'.format(sysconfdir), language : 'c')
+else
+	add_project_arguments('-DSYSCONFDIR="/@0@/@1@"'.format(prefix, sysconfdir), language : 'c')
+endif
 
 version = get_option('sway-version')
 if version != ''
@@ -157,9 +163,16 @@ subdir('swaynag')
 subdir('swaylock')
 
 config = configuration_data()
-config.set('sysconfdir', sysconfdir)
 config.set('datadir', join_paths(prefix, datadir))
 config.set('prefix', prefix)
+
+# If prefix is '/usr', sysconfdir will be explicitly set to '/etc' by Meson to
+# enforce FHS compliance, so we should look for configs there as well.
+if prefix == '/usr'
+	config.set('sysconfdir', sysconfdir)
+else
+	config.set('sysconfdir', join_paths(prefix, sysconfdir))
+endif
 
 configure_file(
 	configuration: config,


SirCmpwn:
Thanks!

SirCmpwn:
Hey, can you write a little shell script which compiles wlroots and sway from scratch on FreeBSD? I'll translate this into a build file and we can do continuous integration on FreeBSD to catch these errors earlier.

sghctoma:
Hey, @SirCmpwn 

I've put the script here: https://gist.github.com/sghctoma/df3acbe8f082c593fa53670d2836f397
You will need FreeBSD 12 for this, and the script will only build wlroots and sway, you won't be able to run it (because we don't have EVDEV support in the default kernel, and there are some issues with epoll-shim).

sghctoma:
Hey, @SirCmpwn 

I've put the script here: https://gist.github.com/sghctoma/df3acbe8f082c593fa53670d2836f397
You will need FreeBSD 12 for this, and the script will only build wlroots and sway, you won't be able to run it (because we don't have EVDEV support in the default kernel, and there are some issues with epoll-shim).

sghctoma:
Hey, @SirCmpwn 

I've put the script here: https://gist.github.com/sghctoma/df3acbe8f082c593fa53670d2836f397
You will need FreeBSD 12 for this, and the script will only build wlroots and sway, you won't be able to run it (because we don't have EVDEV support in the default kernel, and there are some issues with epoll-shim).

SirCmpwn:
Thanks @sghctoma! I'll get this working soon.

sghctoma:
My pleasure.
And sorry for sending this three times, plus in e-mail. I was getting error messages, and my comments didn't show up here when I sent them because of the GitHub outage.

SirCmpwn:
Another couple of issues:

- I hit automake issues when installing wayland-protocols from ports: [details](https://builds.sr.ht/~sircmpwn/job/8923#task-ports-190)
- Can you send me an email and/or patch with the necessary fixes for scdoc?

I can also hook you up with a sr.ht account if you want to play with this build manifest yourself.

I hope these comments are getting through...

sghctoma:
The patch agains the ports tree would be quite large (~160kB), but I've just found out GitHub also "speaks" SVN, so I can export individual directories from a repo. This version of the script uses this to get new libinput+deps: https://gist.github.com/sghctoma/d627dbb72d262a45187cda0542e1bef0
Is it ok this way?

Looking at the builds.sr.ht job, it seems you have an old ports tree (wayland-protocols 1.12 and wayland 1.14), I think that's the reason of the automake issue. You need to run `portsnap fetch update`.

As for scdoc, I was reading the Makefile of an older version, and didn't realize I can set `MANDIR` form env. Anyways, I'll send a patch that does that automatically.

An sr.ht account would be nice, thanks! I've tried setting it up in a VM, but had some troubles with the database, and didn't have much time to troubleshoot yet.

SirCmpwn:
>This version of the script uses this to get new libinput+deps: >https://gist.github.com/sghctoma/d627dbb72d262a45187cda0542e1bef0
>Is it ok this way?

Sure.

>Looking at the builds.sr.ht job, it seems you have an old ports tree (wayland-protocols 1.12 and wayland 1.14), I think that's the reason of the automake issue. You need to run portsnap fetch update.

Ah, thanks. Do you think this is something I should be running for all FreeBSD users when booting up their build VMs?

SirCmpwn:
`portsnap fetch update` - can this be done non-interactively?

SirCmpwn:
Current progress:

https://builds.sr.ht/~sircmpwn/job/9016

https://builds.sr.ht/api/jobs/9016/manifest

sghctoma:
> Do you think this is something I should be running for all FreeBSD users when booting up their build VMs?

To be honest, I would reconsider even installing the ports tree in the first place by default. Mixing ports and packages is generally not a good idea because of the version differences (installing from ports could result in rebuilding *lots* of stuff), and I would assume most users won't need it anyways. Those who do, could fetch it from manifest. 
But if you do install it, than yes, updating when booting up would be a good idea.

> `portsnap fetch update` - can this be done non-interactively?

This should be non-interactive by default (without the `--interactive` option), but will check.

Issues with job 9016:
- You try to run `portsnap` without `sudo` (also `pkg remove` and the epoll-shim.pc redirect)
- `portsnap cron` is not a good idea here, since it sleeps a random amount up to one hour.

I'll get my builds.sr.ht running, and fix the manifest today.

sghctoma:
Okay, I see why you did `portsnap cron`. I'm trying to find a solution for this.

sghctoma:
Here is my current manifest: https://builds.sr.ht/api/jobs/9140/manifest
It builds wlroots, but you need FreeBSD 12.0 (which is currently in BETA1 phase) for sway because of this bug: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=231425

SirCmpwn:
Thanks! I'll adapt this for wlroots and we'll add sway when FreeBSD 12 rolls around.

SirCmpwn:
Sorry for the wait, I had to add support for this to dispatch.sr.ht...

https://github.com/swaywm/wlroots/pull/1336

sghctoma:
> Thanks! I'll adapt this for wlroots and we'll add sway when FreeBSD 12 rolls around.

Awesome, glad I could help!
FreeBSD 12 is due in late December, I will remind you when it's released.

PR #2862, opened 2018-10-17
Fix overflow in strcpy

SpeedJack:
Compiling current master:
```
In function ‘strcpy’,
    inlined from ‘get_text_size’ at ../sway-9999/common/pango.c:95:3:
/usr/include/bits/string_fortified.h:90:10: error: ‘__builtin___memcpy_chk’ writing 18 bytes into a region of size 8 overflows the destination [-Werror=stringop-overflow=]
   return __builtin___strcpy_chk (__dest, __src, __bos (__dest));
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘strcpy’,
    inlined from ‘pango_printf’ at ../sway-9999/common/pango.c:115:3:
/usr/include/bits/string_fortified.h:90:10: error: ‘__builtin___memcpy_chk’ writing 18 bytes into a region of size 8 overflows the destination [-Werror=stringop-overflow=]
   return __builtin___strcpy_chk (__dest, __src, __bos (__dest));
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Introduced in #2843 

Cause:
```c
char* overflow = "[buffer overflow]";
sizeof(overflow); /* = sizeof(char*) = 8 (WRONG) */
```
```c
char overflow[] = "[buffer overflow]";
sizeof(overflow); /* = 18 (CORRECT) */
```
Fixes #2863

PR #2865, opened 2018-10-17
Don't overflow the buffer when buffer would overflow.

johnae:
The code was taking the pointer size rather than the length of the string when strcpy:ing into the buffer. This would overflow the buffer which, ironically, this code was meant to avoid.

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -92,7 +92,7 @@ void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
 	va_list args;
 	va_start(args, fmt);
 	if (vsnprintf(buf, sizeof(buf), fmt, args) >= max_chars) {
-		strcpy(&buf[sizeof(buf) - sizeof(overflow)], overflow);
+		strcpy(&buf[sizeof(buf) - strlen(overflow) - 1], overflow);
 	}
 	va_end(args);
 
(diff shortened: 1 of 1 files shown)


SirCmpwn:
This isn't actually a buffer overflow. Consider this case:

```c
char buf[8];
char overflow[] = {'o','v','e','r',0};
strcpy(&buf[sizeof(buf) - sizeof(overflow)], overflow);
```

`sizeof(buf)` is 8, and `sizeof(overflow)` is 5. `8 - 5` is 3, so it'll be copied starting at index 3, resulting in `[?, ?, ?, 'o', 'v', 'e', 'r', 0]` - which is 8 bytes long, no overflow.

Right?

SirCmpwn:
Note that your branch is missing this fix: https://github.com/swaywm/sway/pull/2862

johnae:
@SirCmpwn That also solves the issue. I fixed this before I ever saw the other pull request. Here's how I thought about it:

```c
#include <stdio.h>
#include <string.h>

static const char *overflow = "[buffer overflow]";
static const char overflow2[] = "[buffer overflow]";

int main() {
  printf("%zu\n", sizeof(overflow)); // gives 8 on my system
  printf("%zu\n", sizeof(overflow2)); // gives 18
  printf("%zu\n", strlen(overflow)); // gives 17
  printf("%zu\n", strlen(overflow2)); // gives 17
}
```

SirCmpwn:
In any case, thanks for the patch.

PR #2870, opened 2018-10-18
Minor refactor of input manager

RyanDwyer:
The input manager is a singleton object. Passing the `sway_input_manager` argument to each of its functions is unnecessary, while removing the argument makes it obvious to the caller that it's a singleton. This patch removes the argument and makes the input manager use `server.input` instead.

On a similar note:

* `sway_input_manager.server` is removed in favour of using the `server` global.
* `seat.input` is removed because it can get it from `server.input`.

Due to a circular dependency, creating seat0 is now done directly in `server_init` rather than in `input_manager_create`. This is because creating seats must be done after `server.input` is set.

Lastly, it now stores the default seat name using a constant and removes a second reference to seat0 (in `input_manager_get_default_seat`).

### DIFF
--- a/include/sway/input/input-manager.h
+++ b/include/sway/input/input-manager.h
@@ -7,12 +7,6 @@
 #include "sway/config.h"
 #include "list.h"
 
-/**
- * The global singleton input manager
- * TODO: make me not a global
- */
-extern struct sway_input_manager *input_manager;
-
 struct sway_input_device {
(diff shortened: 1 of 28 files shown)


SirCmpwn:
This was originally done to make it easier to move away from a singleton, but I don't really care.

emersion:
Thanks!

PR #2871, opened 2018-10-18
Remove cursor warping from seat_set_focus

RyanDwyer:
Because cursor warping was the default behaviour in `seat_set_focus`, there may be cases where we may have been warping the cursor unintentionally. This patch removes cursor warping from `seat_set_focus` and only does it in the `focus` command. This is managed by a static function in `focus.c`.

To know whether to warp or not, we need to know which node had focus previously. To keep track of this easily, `seat->prev_focus` has been introduced and is set to the previous in `seat_set_focus`.


SirCmpwn:
Why does it make so much sense to handle warping in the focus command, and why didn't I do it that way in the first place? You're a genius.

PR #2874, opened 2018-10-18
swaybar: separate input code to new file

ianyfan:
This will allow for easier extensibility in the future without bloating bar.c any more.


RedSoxFan:
Thanks!

Issue #1750, opened 2018-04-06
Swaybar doesn't update cursor on output scale change


emersion:
When updating an output scale, swaybar's cursor scale is oversized.

PR #2876, opened 2018-10-18
swaybar: remove cursor code

ianyfan:
Fixes scaling issue in https://github.com/swaywm/sway/issues/1750

Someone tell me why this is a stupid thing to do :P

### DIFF
--- a/include/swaybar/input.h
+++ b/include/swaybar/input.h
@@ -8,9 +8,6 @@ struct swaybar_output;
 
 struct swaybar_pointer {
 	struct wl_pointer *pointer;
-	struct wl_cursor_theme *cursor_theme;
-	struct wl_cursor_image *cursor_image;
-	struct wl_surface *cursor_surface;
 	struct swaybar_output *current;
 	int x, y;
 };
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -9,7 +9,6 @@
 #include <sys/wait.h>
 #include <unistd.h>
 #include <wayland-client.h>
-#include <wayland-cursor.h>
 #include <wlr/util/log.h>
 #include "swaybar/bar.h"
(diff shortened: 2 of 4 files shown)


emersion:
>Firstly, can sway distinguish between surfaces that don't use the pointer at all and ones that do, but don't set a cursor?

No

>Secondly, can there be valid cases where surfaces want to use the pointer but don't want/care to set the cursor?

Per the Wayland spec, no. If clients don't want a cursor, they can set it to NULL on enter.

ianyfan:
> Per the Wayland spec, no. If clients don't want a cursor, they can set it to NULL on enter.

Doesn't that hide the cursor? What if it doesn't care to change it from what it already is?

emersion:
"What it already is" is defined as "undefined" by the spec. It can be e.g. a resize cursor if it crossed a border before entering the surface.

ianyfan:
Then you're suggesting that if a surface opens without setting a cursor because it just wants to display stuff, sway shouldn't change the cursor from what it was before?

emersion:
Yes. Otherwise there's a race condition: sway will decide to change the cursor and maybe after the surface will set a different cursor. In the end the cursor will blink with sway's cursor before displaying the correct one.

This is sway's current behaviour. To test, open two terminals on two workspaces (caret cursor). When switching from one workspace to the other a normal cursor set by sway blinks.

ianyfan:
But if instead sway doesn't change it, but the cursor moves into a window that doesn't set the cursor, and it's not a pointer, that feels wrong to me.

emersion:
>a window that doesn't set the cursor

This is a client misbehaviour, per the spec.

ianyfan:
Seems like a lot of boilerplate to me :(

PR #2877, opened 2018-10-18
Consider cursor warp when switching workspaces

RyanDwyer:
Fixes a regression introduced in 24a90e5d86441fc345356eb3767e5a6880dcedbd.

`consider_warp_to_focus` has been renamed to `seat_consider_warp_to_focus`, moved to `seat.c` and made public. It is now called when switching workspaces via `workspace <ws>`.


SirCmpwn:
Thanks!

PR #2879, opened 2018-10-19
swaybar: disallow left and right position and print error on default

### LINKED ISSUE #2878: swaybar: Right anchor displays incorrectly.
When swaybar is configured to display on the right side of the screen, it will instead display at the top, and applications will overlay it.

Version:
```
swaybar version 1.0-alpha.6-304-g30dbb8eb (Oct 18 2018, branch 'master')
```

Steps to reproduce:
1. Define `position right` in the `bar` block of your Sway config
2. Reload the configuration file however you like
3. Observe that the bar displays at the top of the screen as if `position top` was defined
4. Observe that applications overlay the bar
5. ~~Observe that `position left` works correctly~~

Emantor:
The positions "left" and "right" are not allowed by the man page, remove them
from the allowed positions. Also print an error to stderr if we default to the
bottom position.

Fixes #2878

### DIFF
--- a/sway/commands/bar/position.c
+++ b/sway/commands/bar/position.c
@@ -12,7 +12,7 @@ struct cmd_results *bar_cmd_position(int argc, char **argv) {
 	if (!config->current_bar) {
 		return cmd_results_new(CMD_FAILURE, "position", "No bar defined.");
 	}
-	char *valid[] = { "top", "bottom", "left", "right" };
+	char *valid[] = { "top", "bottom" };
 	for (size_t i = 0; i < sizeof(valid) / sizeof(valid[0]); ++i) {
 		if (strcasecmp(valid[i], argv[0]) == 0) {
 			wlr_log(WLR_DEBUG, "Setting bar position '%s' for bar: %s",
--- a/swaybar/config.c
+++ b/swaybar/config.c
@@ -1,6 +1,7 @@
 #define _XOPEN_SOURCE 500
 #include <stdlib.h>
 #include <string.h>
+#include <wlr/util/log.h>
 #include "swaybar/config.h"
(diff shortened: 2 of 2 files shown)


Emantor:
@ianyfan We could also exit instead of defaulting to bottom, I decided to keep the existing behaviour here. What is your opinion on that?

Emantor:
I think both should be removed. I'll add the `wlr_log` for `commands/bar/position.c`, but I don't think that will work for swaybar, since afaik `wlr_log` is only available to the compositor. I think printing to `stderr` is the right call there.

emersion:
>afaik wlr_log is only available to the compositor

It's also available in clients. But it makes clients link with wlroots, which is not great.

Emantor:
Swaybar does currently not link against wlroots, so i'd like to avoid that dependency.

Emantor:
I removed the allowed positions from sways command. The command already returns an error for invalid positions, so imo no change is needed there. Should be ready for review.

RyanDwyer:
> Swaybar does currently not link against wlroots

I don't know where you're getting your information from. It's linked in `swaybar/meson.build`, and if you search the `swaybar` directory for `wlr_log` you'll see it's used dozens of times.

Emantor:
> > Swaybar does currently not link against wlroots
> 
> I don't know where you're getting your information from. It's linked in `swaybar/meson.build`, and if you search the `swaybar` directory for `wlr_log` you'll see it's used dozens of times.

I did `rg "wlr*.h"` in the swaybar directory, however `*` does not match `/` and since the include path is `wlr/util/log.h` I assumed that swaybar is not linked against wlroots.

I replaced the `fprintf` with `wlr_log`, sorry for the noise.

emersion:
Thanks!

PR #2882, opened 2018-10-19
Fix logic used for mouse_warping output

### LINKED ISSUE #2881: Mouse warping regression
Running `workspace n` where n is a workspace on a different output from the current output does not bring the mouse along for the ride.

RyanDwyer:
Turns out we don't need to store the previous focus, and it should be based on which output the cursor was in.

Fixes #2881


SirCmpwn:
Thanks!

PR #2884, opened 2018-10-19
Prevent overriding background mode after it's been set.

### LINKED ISSUE #2867: swaylock: Scaling parameter has no effect.
Passing `-s` or `--scaling` to swaylock appears to have no effect. The input image `-i` is always stretched to fill the screen. I think I recall this working recently, so the bug has probably been introduced within the past few days.

Version:
```
swaylock version 1.0-alpha.6-285-g799f285c (Oct 17 2018, branch 'master')
```

Steps to reproduce:
1. Execute `swaylock -s center -i Sway_Tree.png` (get this image from `assets/`)
2. Observe that the image is stretched to fill the screen and is not centered

c-edw:
Fixes #2867.
I have added a new variant of the `background_mode` enum called `BACKGROUND_MODE_UNSET`. This is probably not the best solution.

### DIFF
--- a/swaylock/main.c
+++ b/swaylock/main.c
@@ -385,7 +385,6 @@ static void load_image(char *arg, struct swaylock_state *state) {
 		return;
 	}
 	wl_list_insert(&state->images, &image->link);
-	state->args.mode = BACKGROUND_MODE_FILL;
 	wlr_log(WLR_DEBUG, "Loaded image %s for output %s",
 			image->path, image->output_name ? image->output_name : "*");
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Instead of an assertion failure can you do a post-arg-parse sanity check and give a better error message?

emersion:
Instead you could just set it to `BACKGROUND_MODE_FILL` before processing arguments.

SirCmpwn:
Thanks!

PR #2885, opened 2018-10-19
Fix crash when ending tiling drag

### LINKED ISSUE #2846: Reliable crash on rearranging windows w/ mouse
First, I apologize that I can't offer a script to reproduce this. I only know how to do certain things with the mouse.

Note, all of the following are done by dragging the window to the top/bottom 10% of the other panel to rearrange my panels.

Spawn three terminals - A, B, C. They should appear as three columns.

```
  |     |  
  |     |  
A |  B  |  C
  |     |  
  |     |
```

Then move C to stack on B:

```
  |    
  |  C
A |------ 
  |  B
  |   
```

Then move C to stack on A:

```
   |    
 C |
---|  B 
 A | 
   |
```

Then move B to the bottom of the first column (**thereby empying the second column** maybe relevant?):

```
 C     
---
 A 
---
 B   
```

Crashes here....

RyanDwyer:
If the container being dragged has a parent that needs to be reaped, it must be reaped after we've reinserted the dragging container into the tree. During reaping, `handle_seat_node_destroy` tries to refocus the dragging container which isn't possible while it's detached.

Fixes #2846.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1110,9 +1110,6 @@ static void seat_end_move_tiling(struct sway_seat *seat) {
 	int after = edge != WLR_EDGE_TOP && edge != WLR_EDGE_LEFT;
 
 	container_detach(con);
-	if (old_parent) {
-		container_reap_empty(old_parent);
-	}
 
 	// Moving container into empty workspace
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #2887, opened 2018-10-20
Put swaybar in overlay layer when using mode hide

### LINKED ISSUE #2832: In bar hide mode, bar is not shown above fullscreen windows
With i3, when using `bar mode hide`, the bar is shown on top of fullscreen windows when the modifier key is pressed.  Sway does not show it.

RyanDwyer:
This allows the bar to render over fullscreen views.

Fixes #2832.


SirCmpwn:
Thanks!

PR #2888, opened 2018-10-20
Remove raise_floating directive

RyanDwyer:
The directive controlled whether floating views should raise to the top when the cursor is moved over it while using `focus_follows_mouse`. The default was enabled, which is undesirable. For example, if you have two floating views where one completely covers the other, the smaller one would be inaccessible because moving the mouse over the bigger one would raise it above the smaller one.

There is no known use case for having `raise_floating` enabled, so this patch removes the directive and implements the `raise_floating` disabled behaviour instead.

Resolves some comments in #2425.


SirCmpwn:
eh...

PR #2890, opened 2018-10-20
Fix crash when defaut seat is not created

### LINKED ISSUE #2891: Unable to start sway after pull request 2870
Sway exits immediately upon startup after https://github.com/swaywm/sway/pull/2870 with message:

```
2018-10-20 11:36:35 - [swaybg/main.c:228] Unable to connect to the compositor. If your compositor is running, check or set the WAYLAND_DISPLAY environment variable.
```

I've tested several versions of sway before that pull request and all start fine, it's when running commit `75ea19c71b08e4b3dd1ad053944724fd91dc04bd` swaybg claims not being able to connect to compositor.

What more would you like me to do?

mihaicmn:
Function input_manager_get_default_seat should always return a seat.

Fixes #2891.

### DIFF
--- a/sway/input/input-manager.c
+++ b/sway/input/input-manager.c
@@ -30,6 +30,10 @@ struct sway_seat *input_manager_current_seat(void) {
 	return seat;
 }
 
+struct sway_seat *input_manager_get_default_seat(void) {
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Thanks! This PR is very much appreciated.

PR #2895, opened 2018-10-20
Fix popup damage issues when toplevel and/or popup uses geometry

### LINKED ISSUE #2893: Damage issues with some popups outside floating window geometry
1. Open Gedit
2. Ctrl+O
3. Open the "file type popup"

RyanDwyer:
The `wlr_xdg_popup_get_toplevel_coords` function has the following quirks:

* It does not do anything with the coordinates of the passed popup. Instead, we are required to add them ourselves, which we do by passing them to the function as the surface local values.
* It adds the geometry (shadows etc) of the toplevel itself, so the coordinates are surface local rather than content local. For this reason, we have to negate the toplevel's geometry (`child->view->geometry`).
* I may be wrong, but the popup positions appear to be stored in surface local coordinates rather than content local coordinates. The geometry (shadows etc) of the popup itself must be negated (`surface->geometry`).

Fixes #2893.

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -25,9 +25,12 @@ static void popup_get_root_coords(struct sway_view_child *child,
 	struct sway_xdg_popup *popup = (struct sway_xdg_popup *)child;
 	struct wlr_xdg_surface *surface = popup->wlr_xdg_surface;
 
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Thanks!

emersion:
We can probably improve `wlr_xdg_popup_get_toplevel_coords`

PR #2896, opened 2018-10-20
Abort if proprietary drivers are in use

### LINKED ISSUE #2894: Log driver info on sway startup
So we can turn nvidia users away without wasting time on troubleshooting :wink:

RyanDwyer:
The idea here is we don't want users to be blissfully unaware that they are running unsupported drivers. So we abort on startup, and force the user to add a specific argument to bypass the check.

Closes #2894.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -22,6 +22,7 @@
 #include "sway/tree/root.h"
 #include "sway/ipc-server.h"
 #include "ipc-client.h"
+#include "log.h"
 #include "readline.h"
 #include "stringop.h"
 #include "util.h"
@@ -81,7 +82,7 @@ void detect_raspi(void) {
 	}
 }
 
-void detect_proprietary(void) {
+void detect_proprietary(int allow_unsupported_gpu) {
 	FILE *f = fopen("/proc/modules", "r");
 	if (!f) {
 		return;
@@ -92,15 +93,30 @@ void detect_proprietary(void) {
 			break;
 		}
 		if (strstr(line, "nvidia")) {
-			fprintf(stderr, "\x1B[1;31mWarning: Proprietary Nvidia drivers are "
-				"NOT supported. Use Nouveau.\x1B[0m\n");
 			free(line);
+			if (allow_unsupported_gpu) {
+				wlr_log(WLR_ERROR,
+						"!!! Proprietary Nvidia drivers are in use !!!");
+			} else {
+				wlr_log(WLR_ERROR,
+					"Proprietary Nvidia drivers are NOT supported. "
+					"Use Nouveau. To launch sway anyway, launch with "
+					"--my-next-gpu-wont-be-nvidia and DO NOT report issues.");
+				exit(EXIT_FAILURE);
+			}
 			break;
 		}
 		if (strstr(line, "fglrx")) {
-			fprintf(stderr, "\x1B[1;31mWarning: Proprietary AMD drivers do "
-				"NOT support Wayland. Use radeon.\x1B[0m\n");
 			free(line);
+			if (allow_unsupported_gpu) {
+				wlr_log(WLR_ERROR,
+						"!!! Proprietary AMD drivers are in use !!!");
+			} else {
+				wlr_log(WLR_ERROR, "Proprietary AMD drivers do NOT support "
+					"Wayland. Use radeon. To try anyway, launch sway with "
+					"--unsupported-gpu and DO NOT report issues.");
+				exit(EXIT_FAILURE);
+			}
 			break;
 		}
 		free(line);
@@ -214,7 +230,7 @@ void enable_debug_flag(const char *flag) {
 }
 
 int main(int argc, char **argv) {
-	static int verbose = 0, debug = 0, validate = 0;
+	static int verbose = 0, debug = 0, validate = 0, allow_unsupported_gpu = 0;
 
 	static struct option long_options[] = {
 		{"help", no_argument, NULL, 'h'},
@@ -224,6 +240,8 @@ int main(int argc, char **argv) {
 		{"version", no_argument, NULL, 'v'},
 		{"verbose", no_argument, NULL, 'V'},
 		{"get-socketpath", no_argument, NULL, 'p'},
+		{"unsupported-gpu", no_argument, NULL, 'u'},
+		{"my-next-gpu-wont-be-nvidia", no_argument, NULL, 'u'},
 		{0, 0, 0, 0}
 	};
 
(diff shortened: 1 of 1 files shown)


clarcharr:
This check aborts even for users who run optimus and aren't actually running an nvidia display. Which is a bug, considering how sway doesn't even touch my nvidia GPU.

ascent12:
I do agree that the way this is handled is incorrect. There is already logic inside wlroots for detecting the driver name of the GPU we're actually using: https://github.com/swaywm/wlroots/blob/master/backend/drm/backend.c#L149-L153.
Any such checks belong there, instead of hackishly reading /proc/modules.

SirCmpwn:
patches welcome

clarcharr:
I currently don't have the time to work on a patch but my main reasoning here was to point out that I think that this patch hurts more than it helps and should be reverted, and the original issue should be reopened.

emersion:
>I think that this patch hurts more than it helps 

It's not like it makes it impossible to use your setup. Just start sway with the flag in the meantime.

clarcharr:
I don't see why requiring a flag in this case solves anything, and its only purpose at the moment is to be rude to me as a user because I don't have the resources to go out of my way and buy a new GPU. Plus, as stated, sway will already crash on an nvidia GPU when rendering directly, and the warning that showed before was sufficient to state that the nvidia GPU was the reason.

SirCmpwn:
If you don't like it, send a bloody patch!

clarcharr:
If there were a patch that were accepted too quickly that broke sway with absolutely no other benefits then I can imagine you'd revert it your damn self. But fine, I'll send in a revert PR later once I get back home and turn on my PC.

SirCmpwn:
What part of "the nvidia proprietary isn't supported" are you not getting? Any inconvenience we cause to users who are explicitly in an unsupported demographic isn't our problem. You don't have to buy a new graphics card, you can install nouveau. This is not our fault.

SirCmpwn:
This project is run by volunteers and if none of them are interested in your problem, you need to volunteer yourself.

SirCmpwn:
This patch was written with a workaround in the first place. It doesn't "blatantly disrespect" anyone. The only blatant disrespect is the entitlement on display here.

Yes, I could trivially merge the patch. But that's not the point. We have a policy here of making users step up and put in the work for the things they care about. This leads to more contributions and is a massive bonus to the health of sway as a whole.

>I don't use Nvidia proprietary drivers on my laptop, but I'm slowly transitioning to Sway from i3 on all my machines, and if this is your attitude toward hardware that works but isn't supported, that's a problem.

Hardware that's not supported **is not supported**. My attitude towards supporting it is not to. That's not a problem, that's a tautology.

LeoTindall:
There's a difference between not supporting something and _deliberately breaking it_.

That said, _would_ you accept a patch that just prints a warning message in this case? I'm happy to submit one.

SirCmpwn:
We already had a warning message. A better patch takes @ascent12's advice. If you want to commit to working on that and revert this patch in the meanwhile, I would merge it.

SirCmpwn:
And again, the patch didn't deliberately break anything. Its aim was to make it more obvious that already-broken systems were broken, and it included a flag which turned off the behavior. Just because you don't like it doesn't mean it's deliberately breaking anything.

clarcharr:
> Yes, I could trivially merge the patch. But that's not the point. We have a policy here of making users step up and put in the work for the things they care about. This leads to more contributions and is a massive bonus to the health of sway as a whole.

I agree with this. That's why I appreciate the fact that projects like Rust tend to prefer seasoned contributors mentoring new contributors for new features, because even though it takes more effort in the short term, it helps out in the long term.

If you had, say, responded with "I am on a time crunch for beta and would like some opt-in flag to exist, and am not willing to put in the effort to change this myself. If you can devote the time to submit a better patch before then I can revert this one" then that would have communicated that very clearly.

Right now, your responses have been nothing but hostile, and it seriously pushes me away from this project as a contributor. In fact, your general attitude in situations like this is a very large reason why I haven't thought of contributing more: I already have a limited amount of energy on days after work, and I feel like trying to contribute to sway would be a mess of constant friction rather than the supposed aid you offer.

I agree that preferring that others step up is better to help the project. But god, at least pretend that you'd like others to step up rather than constantly pushing them away.

For example, this flag itself is --my-next-gpu-wont-be-nvidia. That itself is basically making the users feel like their original choice of GPU was 100% their fault, and that they themselves *deserve* to not be supported. What about that makes it seem like you'd even give a shit about patches that make the experience better for nvidia users, even those that don't relate at all to EGLstreams?

SirCmpwn:
You'll find that I'm generally pretty amicable to users who don't blatantly disregard the stated fact that we don't support systems running the proprietary driver. I agree that I've been too rude in this thread, though, and I apologise.

That being said, it doesn't change the fact that we explicitly do not support systems running this driver in any capacity. The whole point of "unsupported" is that we don't want to hear about it on our bug tracker. The only officially supported solution to this problem is "rmmod nvidia".

clarcharr:
I appreciate your apology. I think that your mentality is still rather rude but I won't blame you for hating nvidia, and I've already stated here that I agree with you.

I'll see what I can do about submitting a patch tonight or tomorrow.

RyanDwyer:
To give some background, prior to this patch we printed a log message if the nvidia driver was loaded. This works okay if the user's only GPU was nvidia because sway would crash and they'd see the log message. But if using hybrid graphics where the primary is Intel and the secondary is nvidia, sway would continue to work (just without nvidia monitors) and the user would never see the log message. Then they're likely to create issues related to outputs not working, causing us to invest time in troubleshooting it, only to find that they are using nvidia and never saw the warning.

To resolve this, we have to make sure the user is aware that they are running an unsupported driver. We wanted a one-time warning, but that means we'd have to store a "seen warning" flag somewhere, but we don't have any facility for that hence why aborting by default and using an argument to bypass it was the chosen method. Admittedly, the argument to bypass it is a bit cheeky, but the user is likely to configure it in a script and never type it again. The intention is not just to break sway for nvidia users.

Since implementing this, it's been pointed out that the method we use to check if the nvidia driver is in use isn't perfect - it can trigger false positives. This is the same method we've used for a long time, but it's more important that we don't get false positives now. To detect it properly we have to retrieve it from DRM, and this requires changes to wlroots - which looks like it's already underway.

LeoTindall:
Yes, I think the motivation makes sense and that the correct solution is to merge the device name/version extraction changes to wlroots and then warn the user through sway ↔ they actually initialize a backend on an unsupported device. 

That warning could potentially be provided through means that are more obvious (nagbar, etc), or perhaps sway _could_ bail out by default to ensure they see the message.

However, with the additional wlroots functionality, it would be easy to implement more informative error messages: e.g. "Your only video device is unsupported; sway is unlikely to work at all, please do not open issues about this. Use --unsupported-gpus to force." vs. "One or more video devices are unsupported, but there are supported devices present as well; you may experience problems with displays connected to [device], [device], .... Please do not open issues about these problems. Use --unsupported-gpus to force."

It gives sway the opportunity to be much more informative.

Issue #2787, opened 2018-10-07
RFE: transparency of inactive windows


Hi-Angel:
As the title says.

I'm open to work on this in my spare time *(I don't have much, so don't expect it got get ready quickly, but I'm interested in that work)*, and here's the report to make sure nobody is against implementing that.

[There's a demand](https://github.com/swaywm/sway/issues/2#issuecomment-248215403), and even though there're more important issues, but it's my personal show-stopper for migrating from i3+Compton over to Sway.

I'm new to Sway codebase, and would be glad if someone outlined the steps needed to implement this.

SirCmpwn:
You could accomplish this with a little Python script that connects to IPC and issues `opacity` commands as focus changes.

Hi-Angel:
Oh, that's interesting, thanks, I'll look at that.

johnae:
@SirCmpwn Is that actually possible with the implementation of opacity as it is today? Afaik you can only ever issue opacity commands for the currently focused window. So you could set the opacity to something when the window gets focus and that's all. It doesn't seem possible to set the opacity to something else when the window looses focus.

emersion:
You can use criteria to apply opacity to other windows.

Hi-Angel:
So, for the most part I've figured how to do it *(I'll document it when it's complete, but basically one have to use `i3ipc-python` library, which needs to be modified to search for sway instead of i3; and then bind a small function to focus event to traverse windows in search for unfocused ones)*. One thing I don't see is: how to apply opacity command to a particular window? Looking at `opacity.c` it takes just a single argument, the opacity value.

SirCmpwn:
Use criteria.

emersion:
Use criteria.

Hi-Angel:
But which criteria do I have to use? There doesn't seem to be "focused" criteria — do I need to extract window id in python script, and then issue a command for that id? Is that id unique?

ianyfan:
You can use `[con_id=__focused__]`

Hi-Angel:
Thank you all!

So, in short, the steps are:

1. Install via package manager or pip the [i3ipc-python](https://github.com/acrisci/i3ipc-python) library.
2. Find in the file `i3ipc.py`of its installation the line:
	```python
	['i3', '--get-socketpath'],
	```

	and replace with 
	```python
	['sway', '--get-socketpath'],
	```
3. run the following script:
	```python
	import i3ipc

	ipc = i3ipc.Connection()

[...]

PR #2900, opened 2018-10-20
A script to make inactive windows transparent

Hi-Angel:
I was thinking whether it's worth putting into sway docs or here, and [I was advised](https://github.com/swaywm/sway/issues/2787#issuecomment-431524704) to create a contrib dir instead, which I'm doing here.


SirCmpwn:
Thanks!

PR #2903, opened 2018-10-21
Make workspace back_and_forth seat-specific

RyanDwyer:
* When using multiple seats, each seat has its own `prev_workspace_name` for the purpose of `workspace back_and_forth`.
* Removes `prev_workspace_name` global variable.
* Removes unused `next_name_map` function in `tree/workspace.c`.
* Fixes memory leak in `seat_destroy` (`seat` was not freed).

Tested using `workspace back_and_forth`, `move container to workspace back_and_forth`, and `reload`.


emersion:
Thanks!

Issue #2906, opened 2018-10-21
Unsupported window_type criteria


makepanic:
Hi,
thanks for the great work on sway.
With beta.1 i tried to checkout sway again and ported various things from my i3 config to sway.

I noticed that some of the criteria aren't supported, e.g. `for_window [window_type="menu"] floating enable` won't compile.

Looking at the source, it looks like there's a hardcoded list of supported types

https://github.com/swaywm/sway/blob/master/sway/criteria.c#L278-L290

There are [some i3 configs](https://github.com/u8sand/i3-config/blob/master/config#L97) that user additional window_type filters.

If https://metacpan.org/pod/X11::Protocol::WM#_NET_WM_WINDOW_TYPE is correct, then this is the list of possible types:

sway | metacpan
---- | ----
NORMAL | NORMAL
DIALOG | DIALOG
not implemented | DESKTOP
not implemented | DOCK
TOOLBAR | TOOLBAR
not implemented | MENU
UTILITY | UTILITY
SPLASH | SPLASH

Then there's also the freedesktop spec https://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472629520

sway | freedesktop
---- | ----
NORMAL | NORMAL
DIALOG | DIALOG
not implemented | DESKTOP
not implemented | DOCK
TOOLBAR | TOOLBAR
not implemented | MENU
UTILITY | UTILITY
SPLASH | SPLASH
not implemented | DROPDOWN_MENU
not implemented | POPUP_MENU
not implemented | TOOLTIP
not implemented | NOTIFICATION
not implemented | COMBO
not implemented | DND

If this is something sway should support, I could try to open a PR and extend the if/else condition block.
I'm not fluent enough with c so it'll also be fine if someone else does it properly :)

RyanDwyer:
Fixed by the above PR.

PR #2909, opened 2018-10-21
Parse missing i3 window types

makepanic:
fixes the parsing part of #2906 

See https://github.com/i3/i3/blob/bde4f40b67b217f98982d1b2f0d7b877d31ed5e4/src/match.c#L318-L326 for window type names of i3

I'm unsure if some of the new atoms should be added to the [list of `wants_floating`](https://github.com/swaywm/sway/blob/master/sway/desktop/xwayland.c#L227) window types.


SirCmpwn:
Thanks!

PR #2913, opened 2018-10-21
swaybar: fix hotplug

emersion:
swaybar doesn't handle hotplug correctly. On hotplug, a `wl_surface` is created and nothing else happens. This patch fixes this.

To test:
1. Run sway with the Wayland backend
2. `swaymsg create_output`
3. The bar should appear on the newly created virtual output


SirCmpwn:
Thanks!

PR #2922, opened 2018-10-22
cmd_reload: recalc font sizing + rebuild textures

### LINKED ISSUE #2905: Window titles become transparent on config reload
`1.0-beta.1-2-ga4d68358 (Oct 21 2018, branch 'master')`
[sway.log](https://gist.github.com/YaLTeR/b170ac92c63f66ed11cfea1f4826a39f#file-sway-log)
[config file](https://gist.github.com/YaLTeR/b170ac92c63f66ed11cfea1f4826a39f#file-config)

Reloading the config file causes window titles to become transparent rectangles:
image

In the `sway.log` session I opened konsole, reloaded the config (which caused konsole's title become a transparent rectangle) and exited sway.

RedSoxFan:
Fixes #2905 

When the config gets reloaded, the font height and baseline get reset to 0. If the config does not have a font command in it, the variables will remain at 0 causing a transparent area where the title would be rendered.

This makes it so the font height and baseline are recalculated. Additionally, since the font height and baseline may have changed due to the reload, the title and marks textures are rebuilt.

### DIFF
--- a/sway/commands/reload.c
+++ b/sway/commands/reload.c
@@ -5,8 +5,16 @@
 #include "sway/ipc-server.h"
 #include "sway/server.h"
 #include "sway/tree/arrange.h"
+#include "sway/tree/view.h"
 #include "list.h"
 #include "log.h"
+
+static void rebuild_textures_iterator(struct sway_container *con, void *data) {
+	if (con->view) {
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Thanks!

PR #2923, opened 2018-10-22
Fix crash when moving a container to a fullscreen workspace

### LINKED ISSUE #2915: Crash when moving program to workspace with another fullcreen program.
When I moved a program(termite) to another workspace that had another program open in fullscreen(mpv) sway crashed. After the crash I restarted sway and did the above steps and got the same result.
 
Never written an issue before so let me know if you need more information.

RyanDwyer:
Setting normal focus to the fullscreen view causes the old workspace to start destroying. We then set focus to the old workspace which is no longer attached in the tree.

As we are only setting `focus_inactive` on the fullscreen container, the fix uses `seat_set_raw_focus` to avoid all the additional behaviour that comes with it such as destroying the old workspace.

Fixes #2915.


SirCmpwn:
Thanks!

Issue #2898, opened 2018-10-20
swaylock does not exit and uses 100% CPU


Marteon:
`swaymsg version 1.0-alpha.6-313-gd855837d (Oct 20 2018, branch 'HEAD')`

Seemingly random swaylock does not terminate and uses 100% CPU.
It happened several times and at leas one time was after suspend.
When it happens again I'll attach the logs.

emersion:
Also happened to me. Maybe we're not correctly handling being disconnected from the display.

ascent12:
I ran into a somewhat similar issue with swaybar ages ago where we weren't handling POLLNVAL/POLLHUP properly, causing the poll loop to spin extremely quickly, wasting 100% CPU.

RyanDwyer:
If you run swaylock from the command line you should get log output.

You might find it easier to run sway in a nested session (use a different config file with a different mod key) and launch swaylock into the nested session by running (from your normal session) `WAYLAND_DISPLAY=wayland-1 swaylock`. You could add log statements to `common/loop.c` in `loop_poll` to see how frequently it's hitting that function. You could also log these:

* See if `pfd.revents` has the `POLLHUP` or `POLLERR` or `POLLNVAL` bits set.
* Check that `ms` doesn't sit at 0 all the time.
* Check that `poll` is not returning -1.

martinetd:
It's not easy to tell with github still buggy, but #2925 will be a fix for this, there's no need to spend more time debugging this issue.
(a first approximation to the logging you suggest adding is strace, though; much easier to run and gives all the bits you suggested to log)

RyanDwyer:
It just happened for me on 641fbe576e10064b253671acd0360b500232a235, which is a commit in a branch of mine which is branched from 46fc4ba4e3e2444ff68a894ab837d8e1e3324e0a.

After unlocking, a swaylock process was still running. I straced it and got this repeatedly:

    poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN|POLLHUP}])
    poll([{fd=3, events=POLLIN}], 1, 2147483647) = 1 ([{fd=3, revents=POLLIN|POLLHUP}])

emersion:
We need to handle `POLLHUP` and stop the event loop in that case.

ianyfan:
Closing by https://github.com/swaywm/sway/pull/2925, please re-open if you are still experiencing this issue.

xyproto:
[...]

PR #2925, opened 2018-10-22
swaylock: exit early if unable to get input inhibitor

ianyfan:
This stops the program from going into a spinlock without exiting

Steps to reproduce problem: run `swaylock` twice.
The second swaylock will display an error, go into a loop and never exit.

Note: not sure whether it is the right call to return an error. Thoughts?

I don't know if this is the error causing https://github.com/swaywm/sway/issues/2898, but it seems related. Maybe if someone could check?


swaylock/main.c
```diff
@@ -925,6 +925,12 @@ int main(int argc, char **argv) {
 	}
 
 	zwlr_input_inhibit_manager_v1_get_inhibitor(state.input_inhibit_manager);
+	wl_display_roundtrip(state.display);
+	if (wl_display_get_error(state.display) == EPROTO) {
```
emersion: Instead you can check if `wl_display_roundtrip` returns -1.

emersion: Instead you can check if `wl_display_roundtrip` returns -1.

emersion:
I'd also like to fix the loop on line 963: if the client is disconnected then we enter an infinite loop. `loop_poll` should be able to return an error, and we should break the loop in this case.

ianyfan:
I'm not very aware of this problem. Could you give a set of steps to help me test it? (or maybe push this to a separate PR/issue)

emersion:
This is just how the loop should work. Otherwise on e.g. compositor crash swaylock will enter a busy loop.

To test you can probably run swaylock, and from another tty run `swaymsg exit`.

SirCmpwn:
I think this bug is in libwayland (eeh, stretching it). But `wl_display_roundtrip` shouldn't hang on a disconnected display, it should return an error immediately.

emersion:
>I think this bug is in libwayland

Hmm? Reminder that we don't use the wayland event loop.

swaylock/password.c
```diff
@@ -97,7 +98,10 @@ void swaylock_handle_key(struct swaylock_state *state,
 				state->eventloop, 50, handle_preverify_timeout, state);
 
 		while (state->run_display && state->verify_password_timer) {
-			wl_display_flush(state->display);
+			errno = 0;
+			if (wl_display_flush(state->display) == -1 && errno != EAGAIN) {
```
emersion: I'm not sure whether checking for EAGAIN is relevant here. If we do, we should probably do it in a loop. But w/e.

SirCmpwn:
>Reminder that we don't use the wayland event loop.

I know, but maybe I misunderstood the issue here. w/e

SirCmpwn:
Thanks!

swaylock/password.c
ianyfan: If by loop, you mean, keep running `wl_display_flush`, the documentation says: "wl_display_flush() never blocks. It will write as much data as possible, but if all data could not be written, errno will be set to EAGAIN and -1 returned. In that case, use poll on the display file descriptor to wait for it to become writable again." so you can't do that.

emersion:
```c
errno = 0;
while (wl_display_flush(state->display) == -1 && errno == EAGAIN) {
    // This space is intentionally left blank
}
if (errno != 0) {
     break;
}
```

ianyfan: I thought the documentation meant you can't do that, you have to `poll` before you can `flush` again.

emersion: Oh, sorry, you're right. My bad.

emersion: However we never wait for the FD to become writable.

ianyfan: Isn't that what `poll` does?

emersion: `poll` lets you know whether an FD is readable and/or writable. When reading we make sure that the FD is readable. However we don't check if it's writable before writing.

PR #2926, opened 2018-10-22
Fix xwayland wants_floating logic

### LINKED ISSUE #2844: Floating is default for emacsclient in Sway 1.0
I don't know if this is an issue, but I open it as one to inform, and then it can be closed if this is not the case.
In version 1.0 connecting to Emacs daemon with emacsclient -nc will by default open window as floating.
This is not the case in 0.15 version. 

A work around is to add the following to config
for_window [class="Emacs"] floating disable

RyanDwyer:
Requires https://github.com/swaywm/wlroots/pull/1327 to actually fix the floating logic, but won't break anything if this is merged first.

See i3's implementation here: https://github.com/i3/i3/blob/a319860cd10cda17e176218e6dc34145fbb3bd94/src/manage.c#L444-L445

To test:

* Run `emacs`, press Alt + x and type `server-start`, then open a terminal and run `emacsclient -nc`. This window should be tiled.
* Run `echo getpin | pinenetry-gtk-2`. This window should be floating.

Fixes #2844.

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -239,7 +239,7 @@ static bool wants_floating(struct sway_view *view) {
 
 	struct wlr_xwayland_surface_size_hints *size_hints = surface->size_hints;
 	if (size_hints != NULL &&
-			size_hints->min_width != 0 && size_hints->min_height != 0 &&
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #2929, opened 2018-10-22
Fix crash when resizing container hidden in the scratchpad

### LINKED ISSUE #2928: Some apps crashes on scratchpads
First of, good job on releasing the beta.1 !
I am now using sway as my primary wm.

Here's the problem:
In i3, I used the scratchpad for some graphical apps that I want running in the background but not dedicated to a workspace, for example spotify for music. Trying to start a scratchpad dedicated app results in sway crashing.

Archlinux, Sway beta.1

Spotify (installed as a user flatpak) crashes sway at call from rofi:

```
(EE) failed to read Wayland events: Connection reset by peer
Gdk-Message: 09:33:42.109: Error reading events from display: Broken pipe
libGL error: Connection closed during DRI3 initialization failure[1022/093342.116122:ERROR:x11_util.cc(90)] X IO error received (X server probably went away)
/proc/self/exe: /app/lib/libcurl-gnutls.so.4: no version information available (required by /proc/self/exe)
```
full debug log: https://gist.github.com/avrelaun/1c10cbae7feccb874f4fbbcfdfc65842

----
Tor Browser (installed by putting the official binary on path) does not crash immediately but after few seconds (could be the time of load with splash screen):
```
Gdk-Message: 09:34:23.774: Error reading events from display: Broken pipe
(EE) failed to read Wayland events: Connection reset by peer
```
full debug log: https://gist.github.com/avrelaun/b414dac5698cd27a3d30b5f5a2b7ce9c

Both works nicely on i3.

Here's the relevant part of my sway config:

```
for_window [class="Tor Browser"] floating enable
for_window [class="Tor Browser"] move scratchpad
for_window [class="Tor Browser" instance="Navigator"] resize set 1600 900
bindsym $mod+F1 [class="Tor Browser"] scratchpad show
for_window [class="Spotify" instance="spotify"] floating enable
for_window [class="Spotify" instance="spotify"] move scratchpad
for_window [class="Spotify" instance="spotify"] resize set 1600 900
bindsym $mod+F3 [instance="spotify"] scratchpad show
```

I haven't tested in other graphical apps yet.

RyanDwyer:
Firstly, the container was wrongly identifying as a tiling container because it had no workspace.

Secondly, when calculating the maximum possible size we can't use the workspace if it's not there, so we'll allow unlimited size in this case.

We should probably use the `config->floating_maximum_{width,height}`, but I feel the resize constraints needs to be refactored anyway and I just want to fix the crash for now.

Fixes #2928.

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -94,7 +94,7 @@ static void calculate_constraints(int *min_width, int *max_width,
 		*min_height = config->floating_minimum_height;
 	}
 
-	if (config->floating_maximum_width == -1) { // no maximum
+	if (config->floating_maximum_width == -1 || !con->workspace) { // no max
 		*max_width = INT_MAX;
 	} else if (config->floating_maximum_width == 0) { // automatic
 		*max_width = con->workspace->width;
@@ -102,7 +102,7 @@ static void calculate_constraints(int *min_width, int *max_width,
 		*max_width = config->floating_maximum_width;
 	}
 
-	if (config->floating_maximum_height == -1) { // no maximum
+	if (config->floating_maximum_height == -1 || !con->workspace) { // no max
 		*max_height = INT_MAX;
 	} else if (config->floating_maximum_height == 0) { // automatic
 		*max_height = con->workspace->height;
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Thanks!

PR #2930, opened 2018-10-22
swaybg: don't abort when output doesn't exist

emersion:
This can happen because of race conditions. It's better to just
exit instead of aborting and generating a coredump.

### DIFF
--- a/swaybg/main.c
+++ b/swaybg/main.c
@@ -232,7 +232,12 @@ int main(int argc, const char **argv) {
 	struct wl_registry *registry = wl_display_get_registry(state.display);
 	wl_registry_add_listener(registry, &registry_listener, &state);
 	wl_display_roundtrip(state.display);
-	assert(state.compositor && state.layer_shell && state.output && state.shm);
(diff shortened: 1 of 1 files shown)


ianyfan:
Could you outline what the race condition is? Thanks

emersion:
1. An output is plugged in
2. swaybg is spawned
3. The output is unplugged
4. swaybg tries to find the wl_output, but fails and aborts

Can happen with e.g. docking stations.

SirCmpwn:
I'd rather fix the race condition... and also deal with no outputs more gracefully.

emersion:
This race condition can't be fixed (apart from making swaybg handle multiple outputs).

SirCmpwn:
>apart from making swaybg handle multiple outputs

:D

emersion:
>and also deal with no outputs more gracefully.

What do you mean?

emersion:
>apart from making swaybg handle multiple outputs

It was #2239, I don't remember why I closed it.

PR #2932, opened 2018-10-22
dependency on git - english and french readme

Arnaudv6:
git is used to determine the current version


SirCmpwn:
Can you ditch the parenthetical and add it to all readmes?

emersion:
#2937

PR #2933, opened 2018-10-22
xwayland: populate window_properties in json for views

Snaipe:
window_properties is documented to contain a subset of the X11 properties
of a window (its title, class, instance, role, and transient ID). This
commit adds the missing json object from the get_tree output for
xwayland windows only.

This is a follow-up of #2911.

Signed-off-by: Franklin "Snaipe" Mathieu <me@snai.pe>

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -30,6 +30,7 @@ enum sway_view_prop {
 	VIEW_PROP_WINDOW_ROLE,
 #ifdef HAVE_XWAYLAND
 	VIEW_PROP_X11_WINDOW_ID,
+	VIEW_PROP_X11_PARENT_ID,
 #endif
 };
 
@@ -256,6 +257,8 @@ const char *view_get_class(struct sway_view *view);
 const char *view_get_instance(struct sway_view *view);
 
 uint32_t view_get_x11_window_id(struct sway_view *view);
+
+uint32_t view_get_x11_parent_id(struct sway_view *view);
 
 const char *view_get_window_role(struct sway_view *view);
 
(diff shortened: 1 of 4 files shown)


Snaipe:
(paging interested parties from the other PR: @ianyfan @RyanDwyer @progandy)

RyanDwyer:
Does this match i3's output? It's weird for the title to be in the window properties, as it's probably also in the outer element.

Maybe you could have `window_properties` regardless of whether it's Xwayland or not, and for Wayland list the app id, and for Xwayland keep it as you currently have it.

Snaipe:
> Does this match i3's output? It's weird for the title to be in the window properties, as it's probably also in the outer element.

Yes. It's a bit weird, but I looked at both the output from an i3 session I ran from my laptop, and the [current bleeding ipc docs](https://build.i3wm.org/docs/ipc.html), and they seem to document that now. I'm guessing they wanted to bundle a subset of x11 window properties at the same place?

Out of that lot, the only element I didn't provide the definitions for is transient_for, which *seems* to be the ID of a window transient to the current view, but I'm both unsure if this is right, and looking through the wlroots codebase there doesn't seem to be an easy way to find that information right now.

> Maybe you could have window_properties regardless of whether it's Xwayland or not, and for Wayland list the app id, and for Xwayland keep it as you currently have it.

That would be a bit weird, wouldn't it? It really looks like that map was made for X11 properties, so it does seem to be somewhat not relevant for native wayland windows. I don't see a real benefit in duplicating app_id in there, considering it's a sway-specific field.

RyanDwyer:
Okay, I didn't realise the app ID was already included. I'm happy to leave that as is and keep `window_properties` for Xwayland windows only.

You should be able to get the transient ID using `c->view->wlr_xwayland_surface->parent->window_id`, keeping in mind that `parent` may be `NULL`.

Snaipe:
Thanks for the pointer. I checked on i3 and I can confirm that transient_for is the ID of the parent window. I updated the commit with the transient_for key, and the output matches i3's.

SirCmpwn:
Thanks!

PR #2935, opened 2018-10-22
Add references to sway-output(5) in sway(5)

mteyssier:
I noticed some missing references to sway-output(5) in the sway(5) man page while looking for the documentation of the _output_ command.

Somewhat a follow-up of #2831

I tried to keep it consistent with the documentation of _input_ and how sway-input(5) is handled.

- Update the incorrect reference to _output_ in the _swaybg_command_ description.
- Add a reference to sway-output(5) in See Also
- Add an _output_ command description referring to sway-output(5)


SirCmpwn:
Thanks!

PR #2936, opened 2018-10-23
cmd_swaynag_command: fix typo in variable

RedSoxFan:
The custom `swaynag_command` was being stored as `config->swaybg_command` instead of `config->swaynag_command`.


emersion:
Thanks!

PR #2940, opened 2018-10-23
Fix swaybar crash when there's no status_command

### LINKED ISSUE #2938: no swaybar without status_command
I believe since #2901 the swaybar segfaults if there's no status_command and the bar doesn't show up

RyanDwyer:
Fixes #2938.


emersion:
Thanks!

PR #2943, opened 2018-10-23
Detail Fedora Source build requirements

jdollar:
Adding some details on issues found when trying to build sway from source in Fedora


emersion:
I think this should be added to the wiki instead.

SirCmpwn:
Aye, this belongs on the wiki.

PR #2950, opened 2018-10-23
Implement the presentation-time protocol

emersion:


emersion:
This is done in `for_each_surface_container_iterator` which is used for both. `send_frame_done_iterator` doesn't check `view_is_visible`.

RyanDwyer:
Oh, right.

SirCmpwn:
Thanks!

PR #2954, opened 2018-10-24
Rebase the cursor after focusing in a direction

### LINKED ISSUE #2945: Scrolling one tab in a tabbed layout also scrolls another
This can be hard to replicate in certain circumstances initially and very easy in other.
First create 2 terminals in a standard workspace. Make the right one tabbed. Make another terminal in the tabbed container, for a total of 2.
Call some command with a long output, like ss. Call it multiple times in both terminals in the tabbed container, filling them with text.
Next, scroll up to a certain point (for example where the command got called last time) in both terminals in the tabbed container.
Next, try this: scroll the top terminal slightly up, then check the bottom terminal for a change in position. If it doesn't change, try scrolling down. If it doesn't change, alternate between scrolling both terminals up and down.
Eventually the terminals will become entangled such that scrolling in one will also scroll the other, in most cases.
I've used gnome-terminal. No idea if it works with other emulators.

RyanDwyer:
Prior to this patch, creating a tabbed container with two views, switching tab and then scrolling without motion would cause the scroll events to be sent to the old focus. To fix this, rebasing the cursor is needed after changing focus.

Fixes #2945.

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -294,6 +294,7 @@ struct cmd_results *cmd_focus(int argc, char **argv) {
 	if (next_focus) {
 		seat_set_focus(seat, next_focus);
 		seat_consider_warp_to_focus(seat);
+		cursor_rebase(seat->cursor);
 	}
 
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);


SirCmpwn:
Thanks!

PR #2955, opened 2018-10-24
Enable gaps in the default configuration

xyproto:
The new gaps feature is so awesome that I believe it should be enabled by default.

This pull request is just a suggestion, the decision is up to you, of course.

Thanks for creating sway! :heart:

### DIFF
--- a/config.in
+++ b/config.in
@@ -197,6 +197,14 @@ mode "resize" {
 bindsym $mod+r mode "resize"
 
 #
+# Gaps:
+#
+gaps inner 10
+gaps outer 15
+smart_gaps on
+smart_borders on
+
+#
 # Status Bar:
 #
 # Read `man 5 sway-bar` for more information about this section.


Emantor:
While imo this should not be enabled by default, there should still be a commented section in the default configuration file which shows how to configure the gaps.

xyproto:
With gaps enabled, sway looks amazing. And since these options are also enabled:

```
smart_gaps on
smart_borders on
```

...gaps will only appear for multiple windows per workspace, not for users with a single terminal window. I consider this to be enough "out of the way" for a default configuration, while still being pretty.

The thinking is that most sway users are pretty experienced, and will want to configure sway to their liking anyhow, while unexperienced users will appreciate good defaults that also look nice.

It's only the default configuration, the configuration of existing users will remain the same.

ianyfan:
I would do something like this:
```
#
### Gaps
#
# Example configuration:
#
#   gaps inner 8
#
# See `man 5 sway` for a complete reference.
```
It might be a good idea to separate the gaps commands into its own section in `sway(5)`.

xyproto:
The current suggested changes follows the comment format that is already present in the existing default configuration.

SirCmpwn:
Thanks for your enthusiasm, but we'll pass on this.

PR #2957, opened 2018-10-24
Rebase the cursor after mapping a view

### LINKED ISSUE #2946: Firefox fails to update its cursor before a click
After starting Firefox, on any webpage or on its chrome, the cursor is only updated on pointer click, not on hover.

It fixed itself once I switched workspaces and back again.

### LINKED ISSUE #2949: Give focus immediately when starting the terminal
See the [attached logs](https://github.com/swaywm/sway/files/2507647/st.txt), an `enter` event has only been sent on line 174, I waited three seconds before moving the pointer.

RyanDwyer:
I originally put the rebase at the end of view_map, but at this point the view is still at its native size and will ignore the motion event if it falls outside of its native size. The only way to do this properly is to rebase the cursor later - either after sending the configure, after the view commits with the new size, or after applying the transaction. I chose to do it after applying the transaction for simplicity.

I then attempted to just call `cursor_rebase` after applying every transaction, but this causes crashes when exiting sway (and possibly other places) because `cursor_rebase` assumes the tree is in a valid state.

So my chosen solution introduces `transaction_commit_dirty_with_callback` which allows `handle_map` to register a callback which will run when the transaction is applied.

Fixes #2949.

Might be a fix for #2946, but I can't reproduce that.

### DIFF
--- a/include/sway/desktop/transaction.h
+++ b/include/sway/desktop/transaction.h
@@ -29,6 +29,13 @@ struct sway_view;
 void transaction_commit_dirty(void);
 
 /**
+ * Same as above, but runs the specific callback when the transaction is
+ * applied.
+ */
+void transaction_commit_dirty_with_callback(
+		void (*callback)(void *data), void *data);
+
+/**
  * Notify the transaction system that a view is ready for the new layout.
  *
  * When all views in the transaction are ready, the layout will be applied.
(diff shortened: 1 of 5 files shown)


SirCmpwn:
Thanks!

PR #2960, opened 2018-10-24
Add input command xkb_current_layout

nowrep:


ianyfan:
What does this do? Add a man entry please. `sway-input.5.scd`

progandy:
This is using ```xkb_state_update_mask``` on the xkb state. Is the loss of some part of the keyboard state acceptable in this case? ```xkb_numlock``` is limited to the configuration file because of that.

> https://xkbcommon.org/doc/current/group__state.html#ga566677517a286527e05efc5680adbe6b
>
> The serialization is lossy and will not survive round trips; it must only be used to feed slave state objects, and must not be used to update the master state.

nowrep:
It doesn't seem to lose numlock state for me. After this command, I have still numlock enabled.

SirCmpwn:
Isn't this something you can just make xkb do via xkeyboard-config?

nowrep:
> Isn't this something you can just make xkb do via xkeyboard-config?

If you mean set it so you can have layout switching with some key, then yes that can be set with `xkeyboard_options`. My use-case here is to set different layout for specific windows (with script over ipc), and I am not aware of any other way to do it (without needing to press some key).

SirCmpwn:
Oh, this is an attempt at per-window layout. Got it.

SirCmpwn:
So, reviewing this, it's not really going to meet the stated goal. We need to add a per-window layout override, which would be annotated on the node in the tree. Then, upon focusing it, we would switch layouts.

nowrep:
Sure, that wasn't my intention with this actually.
I chose this approach because it is easier and more flexible I think. For me it works perfectly, I only needed small change to script I used with i3.

SirCmpwn:
I see. Well, I'm not interested in merging it in its current state.

PR #2964, opened 2018-10-24
When scrolling on a tab titlebar, set focus_inactive if not focused

RyanDwyer:
For example, create layout `H[view T[view view view]]`, focus the view in the hsplit and scroll the mouse wheel over the tab title bars. Prior to this patch, focus would be given to a descendant of the tabbed container. This patch keeps the focus on the hsplit view.

This also renames some of the variables used in this part of the code to make it be easier to follow.

Tested with regular view children, as well as split container children, and tested while focus is on the tabbed container as well as off the tabbed container.


SirCmpwn:
Thanks!

PR #2974, opened 2018-10-25
Rebase the cursor after applying transactions

RyanDwyer:
This approaches cursor rebasing from a different angle. Rather than littering the codebase with `cursor_rebase` calls and using transaction callbacks, this just runs `cursor_rebase` after applying every transaction - but only if there's outputs connected, because otherwise it causes a crash during shutdown.

There is one known case where we still need to call `cursor_rebase` directly, and that's when running `seat seat0 cursor move ...`. This command doesn't set anything as dirty so no transaction occurs.

### DIFF
--- a/include/sway/desktop/transaction.h
+++ b/include/sway/desktop/transaction.h
@@ -29,13 +29,6 @@ struct sway_view;
 void transaction_commit_dirty(void);
 
 /**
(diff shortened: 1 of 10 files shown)


emersion:
Thanks!

PR #2975, opened 2018-10-25
Deny several commands when there's no outputs connected

### LINKED ISSUE #2972: Segfault on workspace change when no output is connected
Here is a gdb backtrace: [sway.txt](https://github.com/swaywm/sway/files/2514396/sway.txt)

RyanDwyer:
Fixes #2972.

I probably haven't found them all, but these are the ones that are likely to be bound to shortcuts and which will cause crashes.


SirCmpwn:
Thanks!

PR #2979, opened 2018-10-25
Alter config variable replacement process

### LINKED ISSUE #2848: config: substitution not working as expected (output cmd)
Could be user error, but I feel like I have a case where substitution is not working as expected:

```
set $mod Mod4

set $term /home/cole/code/way.sh termite
set $bg bg #ffffff solid_color

output * $bg

```

This results in:

`Error on line 7 (...): 'output * $bg': Invalid output subcommand: bg #ffffff solid_color.`

Yet, this config works, where I've performed the substitution manually:

```
set $mod Mod4

set $term /home/cole/code/way.sh termite
output * bg #ffffff solid_color

```

### LINKED ISSUE #2968: commands in config variables doesn't seem to work
in i3 this is possible to do in the config:   
```text
set $super bindsym Mod4
$super+a exec COMMAND
$super+b exec COMMAND
```

this doesn't work in the sway config. No big deal, just wanted to let you know if you want full compatibility with i3 config. Also it's a nice way to keep the lines cleaner and shorter in the config.

tested on `sway version 1.0-beta.1-36-g7160b8aa (Oct 23 2018, branch 'master')`

i use arch btw

RedSoxFan:
**Since this heavily alters the way the config commands are processed, this needs to be heavily tested before merging**

Currently, variables cannot contain commands and cannot span more than one argument. This is due to variable replacement happening after determining the handler and after splitting the config line into arguments.

This changes the process to:
1. Check for empty lines and block boundaries
2. Split the arguments as before
3. Verify that the first argument is not a variable. If needed the following occurs
    a. Perform variable replacement on just the first argument
    b. Join the arguments back together then split the arguments again. _(This is needed when the variable contains the command and arguments for the command)_
4. Determine the handler
5. If the handler is cmd_set, escape the variable name so that it does not get replaced
6. Join the arguments back together, do variable replacement on the full command, and split the arguments again
7. Perform any needed quote stripping or unescaping on arguments
8. Run the command handler

This allows for config snippets such as:

```
# Issue 2968
set $super bindsym Mod4
$super+a exec some-command
```

and

```
# Issue #2848
set $bg bg #ffffff solid_color
output * $bg
```

Fixes #2968 
Fixes #2848

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -353,12 +353,14 @@ struct cmd_results *config_command(char *exec) {
 	struct cmd_results *results = NULL;
 	int argc;
 	char **argv = split_args(exec, &argc);
+
+	// Check for empty lines
 	if (!argc) {
 		results = cmd_results_new(CMD_SUCCESS, NULL, NULL);
 		goto cleanup;
 	}
 
-	// Start block
+	// Check for the start of a block
 	if (argc > 1 && strcmp(argv[argc - 1], "{") == 0) {
 		char *block = join_args(argv, argc - 1);
 		results = cmd_results_new(CMD_BLOCK, block, NULL);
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Bump. Anyone interested in this?

SirCmpwn:
Lemme test this

SirCmpwn:
Thanks!

Issue #2967, opened 2018-10-24
[i3-ipc compatibility] output event not implemented


Ferdi265:
the "output" subscribable event is not yet implemented in sway.

sway ipc instead answers to the ipc event with `{"success": false}` and logs

    [sway/sway/ipc-server.c:687] Failed to parse request: Resource temporarily unavailable

The "Resource temporarily unavailable" seems to come from errno, which is not relevant here. The event name just failed the `strcmp` chain, which doesn't set errno.

ianyfan:
We don't emit output events anywhere yet, since it's not documented by i3 when they are emitted, so there hasn't really been a reason to allow it to be subscribed to yet.

SirCmpwn:
I don't see any reason to implement output subscribe given it doesn't really do anything in i3.

progandy:
@ianyfan i3 just forwards notifications it receives from RandR
> https://i3wm.org/docs/3.e/ipc.html#_available_events
> output (1)
>    Sent when RandR issues a change notification (of either screens, outputs, CRTCs or output properties).

Ferdi265:
Also, independent of wether we implement it or not, the irrelevant errno formatted message should IMO be removed / changed. It just doesn't make sense here.

emersion:
Yes. Please send a patch.

Ferdi265:
Will do :)

philipflohr:
@SirCmpwn I'm writing a little tool helping to configure outputs which could use the output event. 
If I write a patch implementing the output event will you merge it?

emersion:
@philipflohr You may be interested in https://github.com/emersion/kanshi

philipflohr:
@emersion thanks but not excatly what I'm looking for as I plan to implemet workspace -> output save/restore and other features. 

But maybe there can be some mutual inspiration
https://github.com/philipflohr/SwayMonMan

@ddevault Would you merge a patch implemeting the output event?

moljac024:
How does kanshi work? I wanted to write a similar tool/script for myself and I was hoping I could rely entirely on sway's IPC (I don't need this to work for anything else right now). I guess I could still do it with polling of get_outputs but that sounds so inefficient.

This would be a useful feature to have and I don't think sway should be crippled by trying to imitate i3 perfectly. After all, Wayland is not X.org and sway already differs on those points. This would be another reasonable difference from i3 IMO.

Ferdi265:
[...]

PR #2984, opened 2018-10-25
sway-ipc: don't log errno if unneeded and add more descriptive errors

Ferdi265:
As discussed in #2967.

The error paths in the handling of the SUBSCRIBE message used `wlr_log_errno()`, which appends the textual representation of `errno`.

In the first case, after failing to parse the message with Json-C, "Failed to read request" is printed.
This is misleading, since it was read correctly, it just wasn't valid JSON. Also, Json-C doesn't set `errno`, so either we use `wlr_log()` or also switch on `json_tokener_get_error()` and build a more detailed error message. The latter is not worth the boilerplate IMO.

Secondly, if an unsupported event name is encountered, the message was "Failed to parse request". I changed this to "Unsupported event type in subscribe request". I changed to `wlr_log()` as above.


emersion:
Thanks!

PR #2985, opened 2018-10-25
Fix re-tiling for floating containers

### LINKED ISSUE #2866: Inconsistent windows sizes when re-tiling.
This is a minor bug that follows on from #2857. When making multiple windows tile after floating, one of the windows will be larger than the other three. See below for example.

out

Version:
```
sway version 1.0-alpha.6-285-g799f285c (Oct 17 2018, branch 'master')
```

Steps to reproduce:
1. Start Sway
2. Spawn three terminal windows using `$mod+Enter`
3. Make all three windows floating using `$mod+Shift+Space` (you may need to focus each window)
4. Make all three windows tiling again using `$mod+Shift+Space`
5. Observe that the rightmost window is larger than the other two

mihaicmn:
When a floating container is tiled (e.g.: 'floating toggle' or
'floating disable'), it should be placed after/below the inactive
focused container from the tiling layout.

Fixes #2866

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -676,11 +676,8 @@ void container_set_floating(struct sway_container *container, bool enable) {
 		container_detach(container);
 		struct sway_container *reference =
 			seat_get_focus_inactive_tiling(seat, workspace);
-		if (reference && reference->view) {
-			reference = reference->parent;
-		}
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

Issue #2976, opened 2018-10-25
Backward-compatible I3SOCK IPC


Snaipe:
Some libraries and programs are currently not working out-of-the-box with sway, because they expect X11-specific fields to always have a non-null value (e.g. class/instance, window).

Those fields are normally there for xwayland windows, but are otherwise null for wayland ones; this ends up crashing any script trying to regexp-match a string against a null,nil,None object, among other things.

A solution to this problem would be to serve a second IPC socket for I3SOCK, and make the output try to accommodate the quirks of i3's original X11-centric output.

Some fairly obvious changes for the I3SOCK output would include:
* set the contents of `class`, `window_properties.class`, `window_properties.instance` with the value of `app_id` for wayland windows.
* always provide `window_properties.*` for wayland windows.
* as a bonus, removing any extra json fields added by sway; not completely necessary, but keeps the output undistinguishable from a stock i3.

Other less-obvious changes:
* @RyanDwyer mentioned that dock containers exist in i3's output but not sway's.
```
2018-10-25 15:18:00     ryan-au BTW, i3's IPC has dock containers and ours doesn't
2018-10-25 15:18:09     ddevault        dock containers?
2018-10-25 15:18:21     ryan-au And in one of the calls I think the root container is returned in the output list
2018-10-25 15:18:42     ryan-au Yeah, it goes root > output > dock, or root > output > content > workspace or something like that
```
* providing an ID for `window` and `window_properties.transient_for` fields? This is a bit delicate, do we provide some fake ID, or an actual xwayland-registered placeholder that identifies the wayland window?

As a side note, I've been running a proxy server on top of I3SOCK in my local machine that does the class/instance-as-app_id change, and previously broken scripts start working again (although with some config tweaking needed, as most programs used to capitalize the instance or class, but this is to be expected anyways).

emersion:
>this ends up crashing any script trying to regexp-match a string against a null,nil,None object, among other things.

This seems strange, do all x11 programs set class and instance? We could set them to an empty string, but that wouldn't make legacy scripts work.

ianyfan:
If this gets implemented, it could also remove unnecessary xwayland properties from SWAYSOCK.

SirCmpwn:
Some of them are necessary because they, well, come from xwayland. Like window class should still be there.

Snaipe:
Well that could be a good way to actually have some sections of the json that makes sense.

[...]

PR #2986, opened 2018-10-25
ipc: add backward-compatible I3SOCK ipc socket.

Snaipe:
This is a work in progress to eventually address #2976.

### DIFF
--- a/include/sway/ipc-json.h
+++ b/include/sway/ipc-json.h
@@ -4,16 +4,27 @@
 #include "sway/tree/container.h"
 #include "sway/input/input-manager.h"
 
+typedef void (*ipc_json_descriptor)(struct sway_node *, json_object *);
+typedef ipc_json_descriptor ipc_json_descriptor_map[NUM_NODE_TYPES];
+
 json_object *ipc_json_success(void);
 json_object *ipc_json_failure(const char *error);
 
 json_object *ipc_json_get_version(void);
 
 json_object *ipc_json_describe_disabled_output(struct sway_output *o);
-json_object *ipc_json_describe_node(struct sway_node *node);
-json_object *ipc_json_describe_node_recursive(struct sway_node *node);
+json_object *ipc_json_describe_node(struct sway_node *node,
+		const ipc_json_descriptor_map *desc);
+json_object *ipc_json_describe_node_recursive(struct sway_node *node,
+		const ipc_json_descriptor_map *desc);
 json_object *ipc_json_describe_input(struct sway_input_device *device);
 json_object *ipc_json_describe_seat(struct sway_seat *seat);
 json_object *ipc_json_describe_bar_config(struct bar_config *bar);
+void ipc_json_describe_container(struct sway_container *c, json_object *object);
+void ipc_json_describe_view_common(struct sway_container *c, json_object *object);
+
+void ipc_json_describe_output(struct sway_node *node, json_object *object);
+void ipc_json_describe_root(struct sway_node *node, json_object *object);
+void ipc_json_describe_workspace(struct sway_node *node, json_object *object);
 
 #endif
--- a/include/sway/ipc-json.h
+++ b/include/sway/ipc-json.h
@@ -4,6 +4,9 @@
 #include "sway/tree/container.h"
 #include "sway/input/input-manager.h"
 
+json_object *ipc_json_success(void);
+json_object *ipc_json_failure(const char *error);
+
 json_object *ipc_json_get_version(void);
 
 json_object *ipc_json_describe_disabled_output(struct sway_output *o);
--- a/include/sway/ipc-server.h
+++ b/include/sway/ipc-server.h
@@ -1,5 +1,6 @@
 #ifndef _SWAY_IPC_SERVER_H
 #define _SWAY_IPC_SERVER_H
+#include <json-c/json.h>
 #include <sys/socket.h>
 #include "sway/config.h"
 #include "sway/tree/container.h"
@@ -19,5 +20,35 @@ void ipc_event_bar_state_update(struct bar_config *bar);
 void ipc_event_mode(const char *mode, bool pango);
 void ipc_event_shutdown(const char *reason);
 void ipc_event_binding(struct sway_binding *binding);
+void ipc_event_tick(const char *payload);
+
+struct ipc_client {
+	struct wl_event_source *event_source;
+	struct wl_event_source *writable_event_source;
+	struct sway_server *server;
+	int fd;
+	uint32_t security_policy;
+	enum ipc_command_type subscribed_events;
+	size_t write_buffer_len;
+	size_t write_buffer_size;
+	char *write_buffer;
+	// The following are for storing data between event_loop calls
+	uint32_t pending_length;
+	enum ipc_command_type pending_type;
+	const struct ipc_client_impl *impl;
+};
+
+bool ipc_send_reply(struct ipc_client *client, enum ipc_command_type payload_type,
+	const char *payload, uint32_t payload_length);
+
+typedef json_object *(*ipc_handler)(struct ipc_client *client,
+	enum ipc_command_type *type, char *data);
+
+struct ipc_client_impl {
+	size_t num_commands;
+	const ipc_handler *commands;
+};
+
+extern const struct ipc_client_impl ipc_client_sway;
 
 #endif
--- a/include/sway/ipc-server.h
+++ b/include/sway/ipc-server.h
@@ -10,7 +10,7 @@ struct sway_server;
 
 void ipc_init(struct sway_server *server);
 
-struct sockaddr_un *ipc_user_sockaddr(void);
+struct sockaddr_un *ipc_user_sockaddr(const char *suffix);
 
 void ipc_event_workspace(struct sway_workspace *old,
 		struct sway_workspace *new, const char *change);
@@ -50,5 +50,6 @@ struct ipc_client_impl {
 };
 
 extern const struct ipc_client_impl ipc_client_sway;
+extern const struct ipc_client_impl ipc_client_i3;
 
 #endif
--- a/include/sway/tree/node.h
+++ b/include/sway/tree/node.h
@@ -17,6 +17,8 @@ enum sway_node_type {
 	N_CONTAINER,
 };
 
+#define NUM_NODE_TYPES ((size_t)(N_CONTAINER + 1))
+
 struct sway_node {
 	enum sway_node_type type;
 	union {
--- a/sway/ipc-i3.c
+++ b/sway/ipc-i3.c
@@ -1,0 +1,85 @@
+#include <json-c/json.h>
+#include "sway/config.h"
+#include "sway/ipc-json.h"
+#include "sway/ipc-sway.h"
+#include "sway/tree/view.h"
+
+static void ipc_i3_json_describe_view(struct sway_container *c, json_object *object) {
+	const char *app_id = view_get_app_id(c->view);
+	const char *class = app_id;
+	const char *instance = app_id;
+	const char *role = NULL;
+	uint32_t parent_id = 0;
+	uint32_t window_id = 0;
+
+#if HAVE_XWAYLAND
+	if (c->view->type == SWAY_VIEW_XWAYLAND) {
+		class = view_get_class(c->view);
+		instance = view_get_instance(c->view);
+		role = view_get_window_role(c->view);
+		parent_id = view_get_x11_parent_id(c->view);
+		window_id = view_get_x11_window_id(c->view);
+	}
+#endif
+
+	json_object_object_add(object, "window", json_object_new_int(window_id));
+
+	json_object *window_props = json_object_new_object();
+
+	if (class) {
+		json_object_object_add(window_props, "class", json_object_new_string(class));
+	}
+	if (instance) {
+		json_object_object_add(window_props, "instance", json_object_new_string(instance));
+	}
+	if (c->title) {
+		json_object_object_add(window_props, "title", json_object_new_string(c->title));
+	}
+	if (role) {
+		json_object_object_add(window_props, "window_role", json_object_new_string(role));
+	}
+	json_object_object_add(window_props, "transient_for",
+			parent_id ? json_object_new_int(parent_id) : NULL);
+
+	json_object_object_add(object, "window_properties", window_props);
+}
+
+static void ipc_i3_json_describe_container(struct sway_node *node, json_object *object) {
+	struct sway_container *c = node->sway_container;
+	ipc_json_describe_container(node->sway_container, object);
+	if (c->view) {
+		ipc_json_describe_view_common(node->sway_container, object);
+		ipc_i3_json_describe_view(node->sway_container, object);
+	}
+}
+
+const ipc_json_descriptor_map ipc_json_i3_descriptors = {
+	[N_ROOT]        = ipc_json_describe_root,
+	[N_WORKSPACE]   = ipc_json_describe_workspace,
+	[N_OUTPUT]      = ipc_json_describe_output,
+	[N_CONTAINER]   = ipc_i3_json_describe_container,
+};
+
+static json_object *ipc_i3_get_tree(struct ipc_client *client,
+		enum ipc_command_type *type, char *buf) {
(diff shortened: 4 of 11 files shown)


sway/ipc-server.c
```diff
@@ -35,58 +33,53 @@
...
+	struct sockaddr_un *sockaddr;
+};
+
+static struct ipc sway_ipc = { .socket = -1 };
+static struct ipc i3_ipc = { .socket = -1 };
+
 static list_t *ipc_client_list = NULL;
 static struct wl_listener ipc_display_destroy;
 
 static const char ipc_magic[] = {'i', '3', '-', 'i', 'p', 'c'};
```
Snaipe: Would it make sense to keep using this magic number for I3SOCK, but use something like "sway-ipc" for SWAYSOCK?

SirCmpwn: nah

include/sway/tree/node.h
```diff
@@ -17,6 +17,10 @@ enum sway_node_type {
 	N_CONTAINER,
 };
 
+enum {
+	NUM_NODE_TYPES = N_CONTAINER + 1
+};
```
emersion: This can just be a `const size_t`

sway/ipc-i3.c
```diff
@@ -0,0 +1,89 @@
...
+
+static void ipc_i3_json_describe_view(struct sway_container *c, json_object *object) {
+	const char *app_id = view_get_app_id(c->view);
+	const char *class = app_id;
+	const char *instance = app_id;
+	const char *role = NULL;
+	uint32_t parent_id = 0;
+	uint32_t window_id = 0;
+
+	(void) role;
```
emersion: I don't think those are necessary.

sway/ipc-i3.c
```diff
@@ -0,0 +1,89 @@
...
+		window_id = view_get_x11_window_id(c->view);
+	}
+#endif
+
+	json_object_object_add(object, "window", json_object_new_int(window_id));
+
+	json_object *window_props = json_object_new_object();
+
+	if (class) {
+		json_object_object_add(window_props, "class", json_object_new_string(class));
```
emersion: Shouldn't these be NULL instead if unset?

sway/ipc-i3.c
```diff
@@ -0,0 +1,89 @@
...
+	[IPC_GET_BAR_CONFIG]    = ipc_sway_get_bar_config,
+	[IPC_GET_BINDING_MODES] = ipc_sway_get_binding_modes,
+	[IPC_GET_CONFIG]        = ipc_sway_get_config,
+	[IPC_GET_MARKS]         = ipc_sway_get_marks,
+	[IPC_GET_OUTPUTS]       = ipc_sway_get_outputs,
+	[IPC_GET_TREE]          = ipc_i3_get_tree,
+	[IPC_GET_VERSION]       = ipc_sway_get_version,
+	[IPC_GET_WORKSPACES]    = ipc_sway_get_workspaces,
+	[IPC_SEND_TICK]         = ipc_sway_send_tick,
+	[IPC_SUBSCRIBE]         = ipc_sway_subscribe,
```
emersion: Some enum values are large here (e.g. `IPC_GET_SEATS`), so maybe it's more appropriate to have a list of pairs (command type, handler). Fortunately events (which have a very large value) shouldn't appear here.

sway/ipc-json.c
```diff
@@ -203,7 +221,7 @@ static void ipc_json_describe_workspace(struct sway_workspace *workspace,
 	for (int i = 0; i < workspace->floating->length; ++i) {
 		struct sway_container *floater = workspace->floating->items[i];
 		json_object_array_add(floating_array,
-				ipc_json_describe_node_recursive(&floater->node));
+				ipc_json_describe_node_recursive(&floater->node, &ipc_json_sway_descriptors));
```
emersion: Style: 80 char limit

sway/ipc-json.c
```diff
@@ -401,26 +379,14 @@ json_object *ipc_json_describe_node(struct sway_node *node) {
...
+	if (desc->descriptors[node->type]) {
+		desc->descriptors[node->type](node, object);
 	}
-
 	return object;
 }
 
-json_object *ipc_json_describe_node_recursive(struct sway_node *node) {
-	json_object *object = ipc_json_describe_node(node);
+json_object *ipc_json_describe_node_recursive(struct sway_node *node, struct ipc_json_descriptors *desc) {
```
emersion: Style: 80 char limit

sway/ipc-json.c
```diff
@@ -452,7 +418,7 @@ json_object *ipc_json_describe_node_recursive(struct sway_node *node) {
 				struct sway_container *child =
 					node->sway_container->children->items[i];
 				json_object_array_add(children,
-						ipc_json_describe_node_recursive(&child->node));
+						ipc_json_describe_node_recursive(&child->node, desc));
```
emersion: Style: 80 char limit

sway/ipc-sway.c
```diff
@@ -0,0 +1,278 @@
...
+	struct json_object *request = json_tokener_parse(buf);
+	if (request == NULL) {
+		wlr_log(WLR_INFO, "Failed to parse subscribe request");
+		return ipc_json_failure(NULL);
+	}
+
+	bool is_tick = false;
+	// parse requested event types
+	for (size_t i = 0; i < json_object_array_length(request); i++) {
+		const char *event_type = json_object_get_string(json_object_array_get_idx(request, i));
```
emersion: Style: 80 char limit

sway/ipc-i3.c
Snaipe: The project doesn't compile if `HAVE_XWAYLAND` isn't defined without those, because the compiler ends up seeing the `if ({role,parent_id,window_id})` and decides that the condition can never be true, and that the variables are actually never used.

sway/ipc-i3.c
Snaipe: i3 omits all fields in `window_props`, except for `transient_for`, for some reason.

sway/ipc-i3.c
emersion:
```c
const char *role = NULL;
if (role) { … }
```

Should compile right?

sway/ipc-i3.c
Snaipe: I think it's mostly fine; the array is still way smaller than a page, and pages are cheap on modern systems anyway. If you're worried about having the array explode in size if somebody mistakenly adds a large value, I can make it size-bound to e.g. 128 slots. I'm not sure using an array of pairs is worth enough the trouble for such a small memory saving, when it ultimately makes the implementation slightly more complicated.

Let me know if you still want me to do so.

sway/ipc-json.c
Snaipe: Mmh, I thought going over 80 was okay as long as the line was readable and still above the 100 columns hard limit, but I just realized that the column size of a tab isn't defined in the contribution guidelines.

I'll fix this, but does a tab count for 8 columns, or 4 columns?

emersion: 4 cols. You can install an [editorconfig](http://editorconfig.org/) plugin for your text editor if it doesn't support it already.

sway/ipc-i3.c
progandy: An implementation on top of ```bsearch``` doesn't look very complicated to me. Use a structure as the array element, add a comparator, ```qsort``` it and access with ```bsearch```. That should be POSIX.1-2001 compliant.

sway/ipc-i3.c
Snaipe: On my system with GCC 8, it doesn't, because GCC thinks (correctly) that role is unused, warns, and stops due to -Werror. I guess you could throw a `volatile` in there, but you'd loose quite more than void-casting or doing a noop on the variables

progandy: Wrapping all occurrences in ```#ifdef HAVE_XWAYLAND``` should work. Is that bad style?

Snaipe: Sure, I can gate these bits with `HAVE_XWAYLAND`, it's not like we have a non-xwayland fallback for those anyway. It might lead to *slightly* larger diffs if we do something about the window_id/parent_id, but I don't think we really have to care about that.

sway/ipc-i3.c
Snaipe: Oh, I was not arguing that it was going to be complicated, just *slightly more* complicated than necessary (plus you need to call qsort, as you mentioned, while using a 1:1 array makes it without change compile-time valid; although you could order the array of pairs directly in the source, but this is really not ideal).

My point is that once again one page of .rodata for both arrays is not only cheap but also half relevant for memory usage, as it's file-backed memory anyway, and the kernel may decide to page it out if it runs out of memory.

Snaipe: On that note, I also realized that I'm missing a `const` modifier on the array and other global structures introduced.

sway/ipc-i3.c
```diff
@@ -25,7 +22,11 @@ static void ipc_i3_json_describe_view(struct sway_container *c, json_object *obj
 	}
 #endif
 
+#ifdef HAVE_XWAYLAND
```
Snaipe: I left the fixup commit as-is for review because that really didn't feel like it improved code quality here.

Should I drop this and keep the void casts, or does anyone have a better alternative?

include/sway/tree/node.h
emersion: This compiles:

```c
#include <stdlib.h>

enum abc {
	A,
	B,
	C,
};

const size_t ABC_NUM = C + 1;
```

emersion: >because it's used in a type definition

Ah right. Yeah, `#define` would be better.

progandy: Sometimes, the number of elements in an enum is simply added as the last element of the enum. Is that a better option?

```c
enum abc {
	A,
	B,
	C,
	NUM_ABC
};
```

emersion: I don't like this, because it gets in the way in `switch` statements. It's also not a valid value of the enum.

sway/ipc-i3.c
Snaipe: GitHub seems to have choked on the rebase, here is the [actual fixup commit](https://github.com/swaywm/sway/pull/2986/commits/466871a89b1ed8067c953be3d944ea9c4491f64b).

Snaipe: I'm going to drop this commit and just move the void casts in the `#else` of the HAVE_XWAYLAND check, if that's alright with everyone. That ought to be cleaner.

emersion:
Can you rebase and squash the fixup commit?

sway/ipc-i3.c
emersion: I still don't get why this is needed. If this file is built without Xwayland, then the code becomes:

```c
uint32_t window_id = 0;
…
json_object_object_add(object, "window", json_object_new_int(window_id));
```

Why doesn't this work?

Snaipe: Oh, yeah, now that you mention it, the void cast for window_id was an oversight. The only guards needed are for role and parent_id, because the compiler complains that they are effectively unused when compiling without Xwayland. I'll remove the bit about window_id though.

emersion: Then just add a guard to the declaration of those two?

Snaipe: I've updated the fixup commit without the window_id guard, and I moved parent_id and role into the first guard. Is that more in line with what you had in mind?

include/sway/ipc-json.h
```diff
@@ -4,13 +4,27 @@
 #include "sway/tree/container.h"
 #include "sway/input/input-manager.h"
 
+typedef void ipc_json_descriptor(struct sway_node *, json_object *);
```
emersion: Is it possible to set this to be a pointer to a function instead?

```c
typedef void (*ipc_json_descriptor)(struct sway_node *, json_object *);
```

include/sway/ipc-json.h
```diff
@@ -4,13 +4,27 @@
...
+typedef ipc_json_descriptor *ipc_json_descriptors[NUM_NODE_TYPES];
```
emersion: The one-char difference between `ipc_json_descriptor` and `ipc_json_descriptors` is confusing. Can we add a `_map` suffix or something?

include/sway/ipc-server.h
```diff
@@ -19,5 +20,33 @@ void ipc_event_bar_state_update(struct bar_config *bar);
...
+	enum ipc_command_type subscribed_events;
+	size_t write_buffer_len;
+	size_t write_buffer_size;
+	char *write_buffer;
+	const struct ipc_client_impl *impl;
+};
+
+bool ipc_send_reply(struct ipc_client *client, const char *payload, uint32_t payload_length);
+
+typedef json_object *ipc_handler(struct ipc_client *client, char *data);
```
emersion: Ditto

```c
typedef json_object *(*ipc_handler)(struct ipc_client *client, char *data);
```

sway/ipc-sway.c
```diff
@@ -0,0 +1,275 @@
...
+			json_object_array_add(outputs,
+					ipc_json_describe_disabled_output(output));
+		}
+	}
+	return outputs;
+}
+
+static void ipc_get_workspaces_callback(struct sway_workspace *workspace,
+		void *data) {
+	json_object *workspace_json = ipc_json_describe_node(&workspace->node, &ipc_json_sway_descriptors);
```
emersion: Style: 80-char limit

emersion:
Needs rebase

panaman67:
@Snaipe Status on the rebase?

Snaipe:
Oh geez sorry, I got busy and forgot about it. I'll see what I can do.

Snaipe:
Rebased & comments addressed.

panaman67:
@emersion Can this be merged with the new rebase and comments addressed?

include/sway/tree/node.h
```diff
@@ -17,6 +17,8 @@ enum sway_node_type {
 	N_CONTAINER,
 };
 
+#define NUM_NODE_TYPES ((size_t)(N_CONTAINER + 1))
```
ddevault: We use N_LAST instead of this

ddevault: That is, we add an enum member, not a macro

sway/ipc-i3.c
ddevault: I would rather have a switch..case here as well.

sway/ipc-i3.c
```diff
@@ -0,0 +1,85 @@
...
+	[IPC_GET_OUTPUTS]       = ipc_sway_get_outputs,
+	[IPC_GET_TREE]          = ipc_i3_get_tree,
+	[IPC_GET_VERSION]       = ipc_sway_get_version,
+	[IPC_GET_WORKSPACES]    = ipc_sway_get_workspaces,
+	[IPC_SEND_TICK]         = ipc_sway_send_tick,
+	[IPC_SUBSCRIBE]         = ipc_sway_subscribe,
+};
+
+const struct ipc_client_impl ipc_client_i3 = {
+	.num_commands = sizeof (i3_commands) / sizeof (*i3_commands),
```
ddevault: style: `sizeof(i3_commands) / sizeof(i3_commands[0])`

include/sway/tree/node.h
emersion: (Not a fan of adding an enum member TBH)

sway/ipc-sway.c
```diff
@@ -0,0 +1,331 @@
...
+#include "sway/ipc-server.h"
+#include "sway/ipc-sway.h"
+#include "sway/output.h"
+#include "sway/tree/workspace.h"
+#include "log.h"
+#include "util.h"
+
+json_object *ipc_sway_command(struct ipc_client *client,
+		enum ipc_command_type *type, char *buf) {
+
```
ddevault: style nit, this n ewline isn't necessary

include/sway/tree/node.h
ddevault: Let's discuss that out of band later, we can't be inconsistent just for this one case

ddevault:
We could also use an update to the sway-ipc man page.

ammgws:
@Snaipe are you still working on this? It seems pretty close to the goal line

PR #2989, opened 2018-10-26
Fix focus after a non-visible workspace's last container is destroyed

### LINKED ISSUE #2951: Unwanted workspace move when a surface gets closed there
How to reproduce:
- Have a terminal on workspace 1.
- Run mpv, while it’s loading move to workspace 2.
- Go back to workspace 1.
- Wait for the song to finish playing.

Expected result:
You stay in your terminal.

Actual result:
You are moved to an empty workspace 2.

RyanDwyer:
The code being changed is responsible for updating the focus stack when a container is destroyed in a different part of the tree to where the real focus is. It's attempting to set focus_inactive to a sibling (or
parent if no siblings) of the container that is being destroyed, then put our real focus back on the end of the focus stack.

The problem occurs when the container being destroyed is in a different workspace. For example:

* Have a focused view on workspace 1
* Have workspace 2 not visible with a single view that is unmapping
* The first call to `seat_set_raw_focus` sets focus to workspace 2 because it's the parent
* Prior to this patch, the second call to `seat_set_raw_focus` would set focus to the view on workspace 1
* Later, when using `output_get_active_workspace`, this function would return workspace 2 because it's the first workspace it finds in the focus stack.

To fix this, workspace 1 must be placed on the focus stack between workspace 2 and the focused view. That's what this patch does.

Lastly, it also uses `seat_get_focus_inactive` to choose the focus. This fixes a crash when a view unmaps while a non-container is focused (eg. swaylock), because `focus` is `NULL`.

Fixes #2951.
Supersedes/closes #2958.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -185,7 +185,11 @@ static void handle_seat_node_destroy(struct wl_listener *listener, void *data) {
 		seat_set_focus(seat, next_focus);
 	} else {
 		// Setting focus_inactive
+		focus = seat_get_focus_inactive(seat, &root->node);
 		seat_set_raw_focus(seat, next_focus);
+		if (focus->type == N_CONTAINER) {
+			seat_set_raw_focus(seat, &focus->sway_container->workspace->node);
+		}
 		seat_set_raw_focus(seat, focus);
 	}
 }


ianyfan:
Why doesn't `output_get_active_workspace` work properly if the top of the stack is the view on workspace 1? Wouldn't it find that view, and then take its workspace, which is 1?

RyanDwyer:
> Why doesn't `output_get_active_workspace` work properly if the top of the stack is the view on workspace 1? Wouldn't it find that view, and then take its workspace, which is 1?

It uses `seat_get_active_tiling_child`, which returns the most recently focused direct child. That function works by doing a foreach followed by a parent check on each node. If we were to change it to how you describe then it would have to use `seat_get_active_child` which does a foreach followed by an ancestor traversal on each node. It's more expensive.

Plus, workspaces must be kept in focus order in the stack because of the IPC `get_tree` command. The `focused` element contains an array of children in focus order.

ianyfan:
So why doesn't `seat_set_raw_focus` put the parents on the focus stack as well?

RyanDwyer:
The intention behind it is for it to do as little as possible. I consider this good practice when still fleshing out a project. Once everything has settled, it's easy to search for usages of `seat_set_raw_focus`, determine if we can make it do a little more (like putting the parents on the stack) then make it be. In contrast, what we don't want to do is make that decision ahead of time then decide we need a `seat_set_really_raw_focus` function that bumps a single node. This is what has already happened with `seat_set_focus` as well as `cursor_send_pointer_motion`. They did too much and we had to split both of them, and that came with some difficulty.

So yes, we might make `seat_set_raw_focus` put the parents on the stack eventually.

SirCmpwn:
Thanks!

PR #2994, opened 2018-10-27
Respect border settings when rendering lone tabbed/stacked child

### LINKED ISSUE #2912: i3-incompatible hide_edge_borders interaction with stacked or tabbed containers
In sway, stacked or tabbed layout means that the view's titlebar will always be at least partially visible. In i3-wm the title can be hidden sometimes, for example when there's a single view and setting `hide_edge_borders both`. This makes the view essentially full-screen, minus the i3bar. But when a sibling view spawns, the tabbed or stacked titlebar becomes visible.

RyanDwyer:
In i3, when a child of a tabbed or stacked container has no siblings, its border settings are respected.

This patch achieves the same effect by rendering a lone tabbed/stacked child as if it's a linear container. This makes the border settings be respected.

Over in `view_autoconfigure`, we compensate for this by only adjusting `y_offset` if there's multiple children.

When the lone child is itself a split container, I'm taking a guess that rendering the tab title (containing the tree representation) is the ideal thing to do here.

Tested with `hide_edge_borders smart`, `hide_edge_borders none`, and with the lone child being a split container.

Fixes #2912.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -781,6 +781,13 @@ static void render_containers_stacked(struct sway_output *output,
 
 static void render_containers(struct sway_output *output,
 		pixman_region32_t *damage, struct parent_data *parent) {
+	if (parent->children->length == 1) {
+		struct sway_container *child = parent->children->items[0];
+		if (child->view) {
+			render_containers_linear(output, damage, parent);
+			return;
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Thanks!

PR #2995, opened 2018-10-27
tree/view.c: fix uninitialized variables warning

colemickens:
I don't know if it's the project settings or the way the meson infra works in `nixpkgs`, but my package is broken without this change. (my toolchain is treating warnings as errors.) Plus, I figure this can only help.

Thanks.

cc: @RyanDwyer the change came from (https://github.com/colemickens/sway/commit/65328ef60c9468ae44b4b1d6316d604c47293ec3#diff-819010104a02bd78a3801147b06a4d57)

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -258,7 +258,7 @@ void view_autoconfigure(struct sway_view *view) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Thanks!

PR #3001, opened 2018-10-27
Reworked fish completions and moved stuff in swaylock manpage

madblobfish:
* Added fish completions for swaynag and swayidle.
* Updated completions for swaylock
* Moved some sections in swaylock manpage to appearance


emersion:
Thanks!

PR #3009, opened 2018-10-28
swayidle: enter idle state on SIGUSR1

### LINKED ISSUE #2914: swayidle does not light up a screen upon pressing a key
Per [this comment](https://github.com/swaywm/sway/issues/2910#issuecomment-431676249), the preferred way of lighting up the screen from dpms is through swayidle. Unfortunately, for me the screen remains dark even with swayidle running in background.

# Steps to reproduce

1. Execute in one terminal `swayidle timeout 600 'swaymsg "output * dpms off"' resume 'swaymsg "output * dpms on"'`
2. Execute in another terminal `sleep 1 && swaymsg "output * dpms off"`
3. Press a key

# Expected

Screens light up back

# Actual

Screens stay black, and I see no way to light them up.

# Additional info

sway build from today, commit a4d68358.

emersion:
This allows to e.g. turn off screens when starting `swaylock`.

Also fixes a few things:
* Removes generic idle callback infrastructure
* Only connect to systemd system bus once if needed
* Allow zero timeouts to mean no timeout in swayidle CLI
* Fixes event loop
* Use `wl_event_loop_add_signal` instead of `signal`

Depends on https://github.com/swaywm/wlroots/pull/1337

Fixes #2914

### DIFF
--- a/swayidle/main.c
+++ b/swayidle/main.c
@@ -1,4 +1,5 @@
-#define _XOPEN_SOURCE 500
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
 #include <errno.h>
 #include <getopt.h>
 #include <pthread.h>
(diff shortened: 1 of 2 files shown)


emersion:
>I know that the clients are currently massively over-linked, but do you really want a "normal" client depending on 'wayland-server'?

I'm not a fan of this, but @SirCmpwn said he's fine with it iirc. Though, switching `swayidle` to another event loop would be a separate task.

RedSoxFan:
Thanks!

PR #3010, opened 2018-10-28
input-manager: consider cursor warping on input_manager_set_focus

Emantor:
input_manager_set_focus is used to set the focus after mapping the view in
view_map. This needs to consider to warp the cursor as well, since for
WARP_CONTAINER, the cursor should warp to the newly created view.

OT:
The only place where `input_manager_set_focus` is used is inside the `view_map`function. We could remove the function and open code it into `view_map`.


SirCmpwn:
Thanks!

Issue #2952, opened 2018-10-24
Configuration reload leaves orphaned swaybg processes


CameronNemo:
After reloading my configuration, I am seeing some orphaned swaybg processes hanging about. In fact each reload generates two of them.

I am running Void Linux, glibc, kernel 4.18.16, with elogind.

Please tell me if there is any more information you need.

madblobfish:
I could not reproduce your issue, could you post your configuration?

CameronNemo:
I can reproduce with the default configuration.

madblobfish:
Please provide a debug log and the exact sway version. (see https://github.com/swaywm/sway/wiki#how-do-i-report-issues)

CameronNemo:
http://ix.io/1qfm

emersion:
Hmm, just realized I still have swaybar processes left behind when I Ctrl+C sway.

RedSoxFan:
Is anyone still able to replicate this?

CameronNemo:
FWIW I have not seen this exact behavior, but I did see a zombie swaybg recently.

emersion:
I'm going to close this, as we now disconnect swaybg instead of killing it.

PR #3013, opened 2018-10-28
Handle sigchld instead of blocking on waitpid

ascent12:
This stops any potential blocking when killing swaybg (e.g. reloading config). We also forcefully close their wayland connection, in case they're slow to die and we start a new swaybg.

This fixes an issue that was reported by someone on IRC, but doesn't have a github issue.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -164,23 +164,31 @@ static void set_mode(struct wlr_output *output, int width, int height,
 	}
 }
 
-void terminate_swaybg(pid_t pid) {
-	int ret = kill(pid, SIGTERM);
-	if (ret != 0) {
-		wlr_log(WLR_ERROR, "Unable to terminate swaybg [pid: %d]", pid);
-	} else {
-		int status;
-		waitpid(pid, &status, 0);
+static void terminate_swaybg(struct wl_display *display, pid_t pid) {
+	struct wl_client *client;
+	wl_client_for_each(client, wl_display_get_client_list(display)) {
+		pid_t p;
+		wl_client_get_credentials(client, &p, NULL, NULL);
+
+		if (p == pid) {
+			wl_client_destroy(client);
+			break;
+		}
+	}
+
+	if (kill(pid, SIGTERM) != 0) {
+		wlr_log_errno(WLR_ERROR, "Unable to terminate swaybg [pid: %d]", (int)pid);
 	}
 }
 
 void apply_output_config(struct output_config *oc, struct sway_output *output) {
 	struct wlr_output *wlr_output = output->wlr_output;
+	struct wl_display *display = output->server->wl_display;
 
 	if (oc && oc->enabled == 0) {
 		if (output->enabled) {
 			if (output->bg_pid != 0) {
-				terminate_swaybg(output->bg_pid);
+				terminate_swaybg(display, output->bg_pid);
 				output->bg_pid = 0;
 			}
 			output_disable(output);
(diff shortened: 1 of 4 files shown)


sway/config/output.c
```diff
@@ -164,23 +164,31 @@ static void set_mode(struct wlr_output *output, int width, int height,
...
-	if (ret != 0) {
-		wlr_log(WLR_ERROR, "Unable to terminate swaybg [pid: %d]", pid);
-	} else {
-		int status;
-		waitpid(pid, &status, 0);
+static void terminate_swaybg(struct wl_display *display, pid_t pid) {
+	struct wl_client *client;
+	wl_client_for_each(client, wl_display_get_client_list(display)) {
+		pid_t p;
+		wl_client_get_credentials(client, &p, NULL, NULL);
```
emersion: `wl_client_get_credentials` can fail (e.g. on FreeBSD where it's not supported). We should either check the return value or initialize `p` to 0 to make sure we don't kill random clients.

sway/config/output.c
```diff
@@ -164,23 +164,31 @@ static void set_mode(struct wlr_output *output, int width, int height,
...
+
+		if (p == pid) {
+			wl_client_destroy(client);
+			break;
```
emersion: `swaybg` can be set to a third-party client, e.g. oguri. This client can open multiple connections. We shouldn't break here.

RedSoxFan:
I think this fixes #2952

sway/config/output.c
ascent12: Looking at the freebsd patches for libwayland, it will simply set the pid to 0, which won't break the logic of this loop; it just won't close the connection.

wl_client_get_credentials has no return value. It's just a getter for values which are read right when the client connects.

The only other way I can think to track this is through the socket file descriptor. It would require us opening it ourselves and using WAYLAND_SOCKET, and would require a fair bit of reworking. Not to mention it doesn't work if a process wants multiple connections (do we want to allow that?).
Checking for the pid seems like to most reliable way, at least for Linux.

emersion: >wl_client_get_credentials has no return value

Oh, all right then, I somehow had the feeling it could fail but I've mixed things up.

sway/server.c
```diff
@@ -30,6 +32,14 @@
 #include "sway/xwayland.h"
 #endif
 
+static int sigchld_handler(int signum, void *data) {
+	while (waitpid(0, NULL, WNOHANG) > 0) {
+		// Do nothing
```
martinetd: similarily I think this should match against known pids (e.g. swaybg_pid etc) and unset the values if it died, so we don't kill a random process later if a user killed swaybg themselves.

sway/config/output.c
```diff
@@ -164,23 +164,31 @@ static void set_mode(struct wlr_output *output, int width, int height,
...
+		}
+	}
+
+	if (kill(pid, SIGTERM) != 0) {
```
martinetd: I would also do the kill before `wl_client_destroy` (before the whole loop?).
If a well-behaved client notices the destroy (that closes the socket connection right?) and exits before we send the kill, it's possible that the pid gets reused meanwhile which would make us kill a random process, or in most cases we could log an error.

martinetd:
Well, if all we care about is swaybg, we can just start it in a double-fork so we don't care about reaping (and just need to waitpid at start time); but that has the same problem we currently have is that we're killing a pid we have no idea if that still is swaybg.
If on top of that we don't kill at all but just destroy the wayland client like you do here (need some message passing on double-fork to identify the right one but we do that in sway exec anyway), that solves both problem without the need for a full blown manager. I just checked, destroying the client makes swaybg exit.

Issue #3015, opened 2018-10-28
Crash in cursor_rebase


emersion:
With the latest commit.

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000055eac3a555ea in node_at_coords (seat=0x55eac45c6790, lx=100, ly=100, surface=0x7ffe9c0ea9f0, sx=0x7ffe9c0ea9f8, sy=0x7ffe9c0eaa00) at ../sway/sway/input/cursor.c:103
103	../sway/sway/input/cursor.c: No such file or directory.
[Current thread is 1 (Thread 0x7fdb6c8a69c0 (LWP 2542))]
(gdb) bt full
#0  0x000055eac3a555ea in node_at_coords (seat=0x55eac45c6790, lx=100, ly=100, surface=0x7ffe9c0ea9f0, sx=0x7ffe9c0ea9f8, sy=0x7ffe9c0eaa00) at ../sway/sway/input/cursor.c:103
        i = 0
        con = 0xc64693a
        unmanaged = 0x55eac44af490
        unmanaged_surface = 0x55eac44af488
        wlr_output = 0x55eac44f4ab0
        output = 0x55eac434ddc0
        ox = 100
        oy = 100
        ws = 0x55eac44be3f0
        c = 0x7ffe9c0ea9f0
#1  0x000055eac3a56ea2 in cursor_rebase (cursor=0x55eac44c5b90) at ../sway/sway/input/cursor.c:590
        time_msec = 207907
        surface = 0x0
        sx = 4.6672797127006555e-310
        sy = 4.6672803195103525e-310
#2  0x000055eac3a4d4b3 in transaction_apply (transaction=0x55eac46328c0) at ../sway/sway/desktop/transaction.c:305
        seat = 0x55eac45c6790
#3  0x000055eac3a4d5f0 in transaction_progress_queue () at ../sway/sway/desktop/transaction.c:339
        transaction = 0x55eac46328c0
#4  0x000055eac3a4de79 in transaction_commit_dirty () at ../sway/sway/desktop/transaction.c:536
        transaction = 0x55eac46328c0
#5  0x000055eac3a76239 in output_layout_handle_change (listener=0x55eac44af478, data=0x55eac44af520) at ../sway/sway/tree/root.c:22
#6  0x00007fdb70039fae in wlr_signal_emit_safe (signal=signal@entry=0x55eac44af548, data=data@entry=0x55eac44af520) at ../util/signal.c:29
[...]

PR #3016, opened 2018-10-28
cursor: fix uninitialized pointer in cursor_rebase

mwenzkowski:
The missing initialization of the pointer `struct wlr_surface *surface`  can lead to the following problem: If the next function call
```
cursor->previous.node = node_at_coords(cursor->seat,
```
doesn't set `surface`, then the following function call
```
cursor_do_rebase(cursor, time_msec, cursor->previous.node, surface, sx, sy);
```
uses this uninitialized pointer (which is possibly not `NULL`). By looking at the definition of this function
```
static void cursor_do_rebase(struct sway_cursor *cursor, uint32_t time_msec,
 		struct sway_node *node, struct wlr_surface *surface,
 		double sx, double sy) {
 	// Handle cursor image
 	if (surface) {
 		// Reset cursor if switching between clients
 		struct wl_client *client = wl_resource_get_client(surface->resource);
                ...
```
one can see that this could lead to a segfault in `wl_resource_get_client(surface->resource)`.
This patch fixes the problem by initializing the pointer to `NULL`

I think the described problem could be the cause of the **2nd** crash @emersion mentioned in #3015.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -585,7 +585,7 @@ static void cursor_do_rebase(struct sway_cursor *cursor, uint32_t time_msec,
 
 void cursor_rebase(struct sway_cursor *cursor) {
 	uint32_t time_msec = get_current_time_msec();
-	struct wlr_surface *surface;
+	struct wlr_surface *surface = NULL;
 	double sx, sy;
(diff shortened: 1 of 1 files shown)


emersion:
Good catch, thanks!

PR #3018, opened 2018-10-28
Add scroll factor input command.

### LINKED ISSUE #3004: [enhancement] Support modifying scroll sensitivity
I've just started using sway 1.0-beta1 on a laptop, and one glaring issue I've noticed is that scrolling via the touchpad is _really_ sensitive, and there doesn't seem to be any configuration option for libinput itself to change this. Physical mice have a `MOUSE_WHEEL_CLICK_ANGLE` property, which according to [this stackoverflow post](https://unix.stackexchange.com/questions/307663/change-scroll-speed-with-libinput) affects scroll rate, but I can't find any equivalent for touchpads. Similarly, the [libinput API](https://wayland.freedesktop.org/libinput/doc/latest/api/group__config.html) also doesn't seem to have a function for directly modifying scroll.

I think that this could be fixed by adding a `scroll_factor` option for `input` entries in sway's config that would be used to scale the delta values of scroll events from the device (which I think happens [in pointer.c here](https://github.com/swaywm/sway/blob/9658d4bcc564dd7c5c82aa1b8795cccedfaabb70/sway/input/cursor.c#L1027)).

**I'm willing to make a PR for this myself**, but I would like a bit of guidance before doing so if possible, as I have a few questions:

1. Is this change best made in sway, or wlroots? Since the other input config options are in Sway, I assume the former.
2. All of the other input configuration options currently map directly to libinput functions (see [here](https://github.com/swaywm/sway/blob/cafb1ff4f78b0f093ab4da987174b92762696366/sway/input/input-manager.c#L141)). Adding an external scaling factor for scroll deltas would be a break from this. Perhaps I should make a PR for libinput to implement this feature, and only then add support for it to Sway?

sxmichaels:
This commit adds a `scroll_factor` input command whose value is used as a multiplier for axis event deltas received from a device. Put simply, this means that scrolling speed can now be adjusted, a configuration feature that libinput itself lacks. At least on my laptop's touchpad, such a change is sorely needed.

For example, adding `scroll_factor 0.5` under a device's `input` section will make scrolling half as fast.

Fixes #3004.

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -3,6 +3,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <float.h>
 #include <math.h>
 #include <stdint.h>
 #include <stdio.h>
@@ -138,6 +139,17 @@ bool parse_boolean(const char *boolean, bool current) {
 	}
 	// All other values are false to match i3
 	return false;
+}
+
+float parse_float(const char *value) {
+	errno = 0;
+	char *end;
+	float flt = strtof(value, &end);
+	if (*end || errno) {
+		wlr_log(WLR_DEBUG, "Invalid float value '%s', defaulting to NAN", value);
+		return NAN;
+	}
+	return flt;
 }
 
 enum wlr_direction opposite_direction(enum wlr_direction d) {
--- a/sway/commands/input/scroll_factor.c
+++ b/sway/commands/input/scroll_factor.c
@@ -1,0 +1,32 @@
+#include <errno.h>
+#include <math.h>
+#include <stdlib.h>
+#include <string.h>
+#include "sway/config.h"
+#include "sway/commands.h"
+#include "sway/input/input-manager.h"
+#include "util.h"
+
+struct cmd_results *input_cmd_scroll_factor(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "scroll_factor", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE,
+			"scroll_factor", "No input device defined.");
+	}
+
+	float scroll_factor = parse_float(argv[0]);
+	if (isnan(scroll_factor)) {
+		return cmd_results_new(CMD_INVALID, "scroll_factor",
+			"Invalid scroll factor; expected float.");
+	} else if (scroll_factor < 0) {
+		return cmd_results_new(CMD_INVALID, "scroll_factor",
+			"Scroll factor cannot be negative.");
+	}
+	ic->scroll_factor = scroll_factor;
+
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+}
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -5,6 +5,7 @@
 #elif __FreeBSD__
 #include <dev/evdev/input-event-codes.h>
 #endif
+#include <float.h>
 #include <limits.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_xcursor_manager.h>
@@ -979,6 +980,8 @@ static void handle_cursor_button(struct wl_listener *listener, void *data) {
 static void dispatch_cursor_axis(struct sway_cursor *cursor,
 		struct wlr_event_pointer_axis *event) {
 	struct sway_seat *seat = cursor->seat;
+	struct sway_input_device *input_device = event->device->data;
+	struct input_config *ic = input_device_get_config(input_device);
 
 	// Determine what's under the cursor
 	struct wlr_surface *surface = NULL;
@@ -990,6 +993,8 @@ static void dispatch_cursor_axis(struct sway_cursor *cursor,
 	enum wlr_edges edge = cont ? find_edge(cont, cursor) : WLR_EDGE_NONE;
 	bool on_border = edge != WLR_EDGE_NONE;
 	bool on_titlebar = cont && !on_border && !surface;
+	float scroll_factor =
+		(ic == NULL || ic->scroll_factor == FLT_MIN) ? 1.0f : ic->scroll_factor;
 
 	// Scrolling on a tabbed or stacked title bar
 	if (on_titlebar) {
@@ -1000,7 +1005,7 @@ static void dispatch_cursor_axis(struct sway_cursor *cursor,
 				seat_get_active_tiling_child(seat, tabcontainer);
 			list_t *siblings = container_get_siblings(cont);
 			int desired = list_find(siblings, active->sway_container) +
-				event->delta_discrete;
+				round(scroll_factor * event->delta_discrete);
 			if (desired < 0) {
 				desired = 0;
 			} else if (desired >= siblings->length) {
@@ -1024,7 +1029,8 @@ static void dispatch_cursor_axis(struct sway_cursor *cursor,
 	}
 
 	wlr_seat_pointer_notify_axis(cursor->seat->wlr_seat, event->time_msec,
-		event->orientation, event->delta, event->delta_discrete, event->source);
+		event->orientation, scroll_factor * event->delta,
+		round(scroll_factor * event->delta_discrete), event->source);
 }
 
 static void handle_cursor_axis(struct wl_listener *listener, void *data) {
--- a/sway/sway-input.5.scd
+++ b/sway/sway-input.5.scd
@@ -105,13 +105,17 @@ The following commands may only be used in the configuration file.
 *input* <identifier> repeat\_rate <characters per second>
 	Sets the frequency of key repeats once the repeat\_delay has passed.
 
-*input* <identifier> scroll\_method none|two\_finger|edge|on\_button\_down
-	Changes the scroll method for the specified input device.
-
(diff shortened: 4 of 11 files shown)


sway/commands/input/scroll_factor.c
```diff
@@ -0,0 +1,26 @@
...
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE,
+			"scroll_factor", "No input device defined.");
+	}
+
+	float scroll_factor = atof(argv[0]);
+	if (scroll_factor < 0) {
+		return cmd_results_new(CMD_INVALID, "scroll_factor",
+			"Input out of range [0, inf)");
```
RyanDwyer: The brackets aren't matching. Although I'd prefer wording such as "Scroll factor cannot be negative".

RyanDwyer:
Can you also document it in the `sway-input` man page?

sway/commands/input/scroll_factor.c
emersion: I think it's deliberate, it's a math notation. But yeah natural language seems more appropriate here.

sxmichaels:
OK, implemented all of @emersion 's fixes and documented the new option in the sway-input man page. One question: is checking `errno != 0` sufficient, or should I provide separate checks and messages for each possible type of error `strtof` can encounter, e.g. out-of-range & underflow?

sway/input/cursor.c
```diff
@@ -1025,7 +1029,8 @@ static void dispatch_cursor_axis(struct sway_cursor *cursor,
 	}
 
 	wlr_seat_pointer_notify_axis(cursor->seat->wlr_seat, event->time_msec,
-		event->orientation, event->delta, event->delta_discrete, event->source);
+		event->orientation, scroll_factor * event->delta,
+		scroll_factor * event->delta_discrete, event->source);
```
emersion: `event->delta_discrete` is an integer, so maybe it should be rounded instead of being silently floored.

sway/sway-input.5.scd
```diff
@@ -105,14 +105,18 @@ The following commands may only be used in the configuration file.
...
 
-*input* <identifier> scroll\_method none|two\_finger|edge|on\_button\_down
-	Changes the scroll method for the specified input device.
-
 *input* <identifier> scroll\_button <button\_identifier>
 	Sets button used for scroll\_method on\_button\_down. The button identifier
 	can be obtained from `libinput debug-events`.
 	If set to 0, it disables the scroll\_button on\_button\_down.
 
+*input* <identifier> scroll\_factor <floating point value >= 0>
```
emersion: That `>= 0` is strange here, Can you instead add a sentence in the description?

sway/commands/input/scroll_factor.c
```diff
@@ -0,0 +1,30 @@
...
+	if ((error = checkarg(argc, "scroll_factor", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE,
+			"scroll_factor", "No input device defined.");
+	}
+
+	char *end;
```
emersion: Can you set `errno = 0` right before calling `strtof`? Otherwise it might fail because of a previous `errno` value.

emersion:
>is checking errno != 0 sufficient, or should I provide separate checks and messages for each possible type of error strtof can encounter, e.g. out-of-range & underflow?

I think just checking `errno` is enough.

sxmichaels:
Great! Thanks for the help.

sxmichaels:
Er, hold on — it looks like something is wrong with the parsing of `scroll_factor` now. To anyone who might merge this: don't yet.

common/util.c
```diff
@@ -139,6 +140,17 @@ bool parse_boolean(const char *boolean, bool current) {
 	return false;
 }
 
+float parse_float(const char *value) {
+	errno = 0;
+	char *end;
+	float flt = strtof(value, &end);
+	if (*end || errno) {
+		wlr_log(WLR_DEBUG, "Invalid float value %s, defaulting to FLT_MIN", value);
+		return FLT_MIN;
```
ascent12: NAN would be a better sentinel value than this.
Or change the signature to `bool parse_float(const char *value, float *ret)`.

sxmichaels:
Okay, not sure what was going on there; I may not have compiled the latest version when I last tested. Everything works now. I also realized that `pointer_accel.c` was also using the `atof` method that I had used in the first commit of this PR, so I moved the float parsing code into a `parse_float` function in `common/util.c` and changed both `scroll_factor.c` and `pointer_accel.c` to use it. It takes a `const char *` and returns `FLT_MIN` on error.

emersion:
Ping

sxmichaels:
Oh dear, very sorry about the delay — somehow I didn't see your last message. I just incorporated your changes and rebased. This was a particularly ugly rebase due to a bunch of merge conflicts in `util.h` and `util.c` … looks like the final result is right, but I seem to have created a bunch of garbage intermediate commits in the process. I wasn't able to figure out a way around this.

sway/input/cursor.c
```diff
@@ -992,6 +995,7 @@ static void dispatch_cursor_axis(struct sway_cursor *cursor,
 	enum wlr_edges edge = cont ? find_edge(cont, cursor) : WLR_EDGE_NONE;
 	bool on_border = edge != WLR_EDGE_NONE;
 	bool on_titlebar = cont && !on_border && !surface;
+	float scroll_factor = (ic == NULL || ic->scroll_factor == FLT_MIN) ? 1.0f : ic->scroll_factor;
```
RedSoxFan: Style: Lines should not exceed 80 columns

ISSUE_TEMPLATE.md
```diff
@@ -1,4 +1,11 @@
-Please include the following four components in your bug report: sway version, debug log, configuration (if applicable), and an explanation of steps taken to reproduce the issue.
+If you are using the nvidia proprietary driver for any reason, you have two choices:
+
+1. Uninstall it and use nouveau instead
+2. Use X11+i3 and close your browser tab
+
+If `lsmod | grep nvidia | wc -l` shows anything other than zero, your bug report is not welcome here.
+
+Otherwise, please include the following four components in your bug report: sway version, debug log, configuration (if applicable), and an explanation of steps taken to reproduce the issue.
```
ddevault: I think you botched a merge/rebase/something

sxmichaels:
OK, I figured I might have to do that. Still not sure what the heck happened during rebasing, so I just completely redid the changes as a single comment as @RedSoxFan suggested.

sway/commands/input/scroll_factor.c
```diff
@@ -0,0 +1,32 @@
...
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE,
+			"scroll_factor", "No input device defined.");
+	}
+
+	float scroll_factor = parse_float(argv[0]);
+	if (scroll_factor == NAN) {
```
RedSoxFan: Sorry. I missed this before. You cannot compare a floating point value with `NAN` to determine whether the value is `NAN` since `NAN != NAN`. You can either use `isnan(scroll_factor)` or `scroll_factor != scroll_factor`.

sxmichaels:
Changed the two instances of `x == NAN` to `isnan(x)` (there was also one in `pointer_accel.c`) and rebased. This rebase went OK.

sxmichaels:
Oh dear. Having done that, all the commits to master since I created this PR become a part of the PR itself, which is the same problem I ran into when I first tried to rebase. I'm stuck at this point; never seen this happen before. Is there some git setting that is making my rebases behave differently than would normally be expected?

RedSoxFan:
Can you please post the output of `git remote -v`? I'll give you instructions after that. Hopefully, it will help you figure out what you are doing differently.

sxmichaels:
Sure.

```
origin    git@github.com:sxmichaels/sway.git (fetch)
origin    git@github.com:sxmichaels/sway.git (push)
upstream    https://github.com/swaywm/sway (fetch)
upstream    https://github.com/swaywm/sway (push)
```

RedSoxFan:
Try the following:
```
git fetch upstream
git checkout add-scroll-factor
git rebase upstream/master
git push -f origin add-scroll-factor
```

emersion:
It seems it's down to one commit again somehow.

RedSoxFan:
Looks like that worked. It's rebased now.

sxmichaels:
Ah! _That_ did it this time, I think, although strangely I could have sworn I'd tried that before. Out of curiosity, do you know what I was doing wrong when rebasing in the first place? I'm still not exactly sure.

Thanks very much for the help — and sorry for all this trouble for such a simple PR!

RedSoxFan:
> Out of curiosity, do you know what I was doing wrong when rebasing in the first place? 

I think either your forks `master` or what you had fetched for `upstream/master` were outdated and on a commit that was reverted. I'm not sure which since it looks like your fork is up-to-date now. Just make sure to do `git fetch upstream` before `git rebase upstream/master` (if you just do `git rebase master`, it will rebase against `origin/master` -- which is your fork).

RedSoxFan:
Thanks!

mvdan:
Thanks so much for this change! Two years later, but I just found out about it and it really made an external mouse much easier to use :)

PR #3020, opened 2018-10-29
swaymsg: fix get_outputs status code

### LINKED ISSUE #3019: swaymsg -t get_outputs returns 1
It should return 0.

This is due to the `success` function checking if each output object has a `success` field set to true. This isn't the case for output objects.

emersion:
Fixes #3019

### DIFF
--- a/swaymsg/main.c
+++ b/swaymsg/main.c
@@ -23,7 +23,7 @@ static bool success_object(json_object *result) {
 	json_object *success;
 
 	if (!json_object_object_get_ex(result, "success", &success)) {
-		return false;
+		return true;
 	}
 
 	return json_object_get_boolean(success);


catb0t:
is this not prone to regression? now `success_object` returns true when something failed

emersion:
>is this not prone to regression? now success_object returns true when something failed

Can you give an example? Tried `swaymsg idontexist` and it exits with 1 as expected.

emersion:
To clarify: the new code expects result objects to have a `success` field. Other objects won't be considered as results and won't make `swaymsg` exit with 1 anymore.

catb0t:
I don't have an example (on mobile), it just looks like the fix is with how success-lacking objects are treated in `success` rather than making failure in `get_ex` return true always.

Snaipe:
FWIW, i3-msg treats json objects without a success key as successful, so it's probably fine to assume so. Real IPC failures are going to have `"success": false` and maybe an error string anyways.

SirCmpwn:
Thanks!

PR #3021, opened 2018-10-29
binding: match single-key bindings if no multi-key binding matched

Snaipe:
This makes bindings more snappy when the user is typing faster than
his keycaps are releasing.

Signed-off-by: Franklin "Snaipe" Mathieu <me@snai.pe>

### DIFF
--- a/include/sway/input/keyboard.h
+++ b/include/sway/input/keyboard.h
@@ -24,6 +24,7 @@ struct sway_shortcut_state {
 	uint32_t last_keycode;
 	uint32_t last_raw_modifiers;
 	size_t npressed;
+	uint32_t current_key;
 };
 
 struct sway_keyboard {
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -32,6 +32,7 @@ static void state_erase_key(struct sway_shortcut_state *state,
 		state->pressed_keys[state->npressed] = 0;
 		state->pressed_keycodes[state->npressed] = 0;
 	}
+	state->current_key = 0;
 }
 
 /**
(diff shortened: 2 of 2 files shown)


SirCmpwn:
I'm not sure what this is trying to accomplish.

Snaipe:
This is something that has been biting me. Essentially, consider the following config:

```
bindsym Left focus left
bindsym Up focus up
```

If you press Left, then Up while still releasing Left, then there's a chance that the keypress for Up happens while the keypress for Left isn't quite completely released, which makes the binding resolution try to find a binding for Left+Up. When Left is completely released, you're left with the Up key pressed, and the binding for Up not running.

These sorts of things happen quite often when using movement bindings, which are pretty much used in rapid succession.

This patch essentially makes the Up binding run as soon as it is pressed if, and only if, there is no Left+Up binding already taking precedence.

Snaipe:
Another way to put it:

Press and hold Left, then press up, then release left. In i3, this makes the binding for left run, then the binding for up run. In sway, only the binding for left runs.

SirCmpwn:
Hm.

Snaipe:
Sorry, I might have not explained that too well. Is something the matter?

SirCmpwn:
Just thinking it over, even if i3 does that I'm not sure it's the right behavior.

Snaipe:
Oh, my point about i3 is more about comparison more than a compatibility thing.

I think Sway's current behaviour is correct when taking into account that bindings can have multiple non-modifiers keys, but at the same time, I'm inclined to believe that the vast majority of bindings out there are single-key (in fact, sway's default config doesn't have multi-key bindings).

Triggering single-key bindings when no multi-key binding has matched seems like a sensible middle ground.

What it does mean, however, is that if a user defines the bindings `b` and `a+b+c`, and presses+hold a, then b, then c, then both bindings will be executed. It's weird, but I think it's also consistent with the fact that if a user defines `a` and `a+b`, and presses+hold a then b, then both bindings are executed (this is what Sway is currently doing).

SirCmpwn:
Thanks!

PR #3022, opened 2018-10-29
WIP: Layout save and restore

RyanDwyer:
This introduces a Python script `sway-save-tree`, similar to `i3-save-tree` which is written in Perl. I considered using acrisci's IPC library but it's not officially part of the sway project, and I didn't really want to introduce a dependency when I can do it using core libraries.

Regarding `sway-save-tree`:

* It supports python 3, so not sure if the shebang line should list python3 or just python as it is.
* It would be nice if we can make meson copy the script into the `build` directory while substituting the version number (currently hard coded) with the constant we use.
* Not sure if there's a python style guide that I should be using.
* This assumes that `swaymsg` is on your path. Not sure if that's a bad assumption to make.

Then over in the `append_layout` command, the idea here is that it builds a list of nodes, detached from the tree, then attaches it at the very end if all is good. This is mostly done; I just have to parse the properties for containers and also parse children.

I've also added a `list_t *swallow_criteria` to containers. i3 supports having multiple lists of criteria which is why this is a list. At this stage I think it'll be a list of `struct_criteria` rather than making a separate struct for it.

Once the `append_layout` command is finished and it populates the `swallow_criteria`, the remaining things to do are to check the swallow criteria when a view maps and render the placeholders.

I'm submitting this so I can get some early review on the Python part of it, and also prevent others from starting the feature.

### DIFF
--- a/sway-save-tree/sway-save-tree
+++ b/sway-save-tree/sway-save-tree
@@ -1,0 +1,112 @@
+#!/usr/bin/env python
+
+import argparse
+import json
+import re
+import subprocess
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-s', '--socket', help='Socket')
+parser.add_argument('-w', '--workspace',
+        help='Specifies the workspace that should be dumped, e.g. 1. ' \
+                'This can either be a name or the number of a workspace.')
+parser.add_argument('-o', '--output',
+        help='Specifies the output that should be dumped, e.g. LVDS-1')
+parser.add_argument('-v', '--version', action='store_true', help='Version')
+args = parser.parse_args()
+
+if args.version:
+    print('sway-save-tree version %s' % '1.0.0')
+    exit(0)
+
+cmd = ['swaymsg', '-t', 'get_tree']
+if args.socket:
+    cmd.concat(['-s', args.socket])
+
+result = subprocess.check_output(cmd)
+tree = json.loads(result)
+
+def filter_tree(tree, args):
+    if args.workspace:
+        for output in tree['nodes']:
+            for ws in output['nodes']:
+                if ws['name'] == args.workspace:
+                    return ws
+    if args.output:
+        for output in tree['nodes']:
+            if output['name'] == args.output:
+                return outut
+
+    # Focused workspace
+    output = next(o for o in tree['nodes'] if o['id'] == tree['focus'][0])
+    ws = next(ws for ws in output['nodes'] if ws['id'] == output['focus'][0])
+    return ws
+
+root = filter_tree(tree, args)
+
+def prop(node, *keys):
+    for key in keys:
+        try:
+            node = node[key]
+        except KeyError:
+            return None
+    return re.escape('^%s$' % node)
+
+def convert_children(nodes):
+    children = []
+    for node in nodes:
+        child = {}
+        has_children = len(node['nodes'])
+        if has_children:
+            child['_comment'] = '%s container with %i children' % (
+                    node['layout'], len(node['nodes']))
+        child['border'] = node['border']
+        child['current_border_width'] = node['current_border_width']
+        child['floating'] = node['floating'] if 'floating' in node else 'auto_off'
+        if has_children:
+            child['layout'] = node['layout']
+        child['geometry'] = node['rect']
+        child['name'] = node['name']
+        child['percent'] = round(node['percent'], 2)
+        child['swallows'] = {}
+        child['swallows']['app_id'] = prop(node, ('app_id'))
+        child['swallows']['class'] = prop(node, ('window_properties', 'class'))
+        child['swallows']['instance'] = prop(node, ('window_properties', 'instance'))
+        child['swallows']['title'] = prop(node, ('name'))
+        child['swallows']['transient_for'] = prop(node, ('window_properties', 'transient_for'))
+        child['type'] = node['type']
+        if has_children:
+            child['nodes'] = convert_children(node['nodes'])
+        if len(node['floating_nodes']):
+            child['floating_nodes'] = convert_children(node['floating_nodes'])
+        if 'marks' in node and len(node['marks']):
+            child['marks'] = node['marks']
+        if 'fullscreen_mode' in node and node['fullscreen_mode']:
+            child['fullscreen_mode'] = True
+
+        children.append(child)
+
+    return children
+
+# Build new tree containing only the properties we want
+newtree = convert_children(root['nodes'])
+
+# Convert it to a JSON string. The root children list should not be an array,
+# but rather concatenated json objects.
+output = ''
+for node in newtree:
+    output += json.dumps(node, indent=4) + "\n"
+output = output.rstrip()
+
+# Replace the _comment element with an actual comment
+output = re.sub(r'"_comment": "(.*?)",', '// \\1', output)
+
+# Comment the swallows criteria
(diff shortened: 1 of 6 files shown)


StephenBrown2:
I'm quite pleased to see this make a comeback, and especially in Python 3.

I can't speak to the sway side of things, but as for formatting, I think it looks decent. For a standard style, I'd recommend [black](https://pypi.org/project/black/) and [isort](https://pypi.org/project/isort/), running the following on `sway-save-tree`:
```
$ black sway-save-tree/sway-save-tree
$ isort --multi-line=3 --trailing-comma --force-grid-wrap=0 --combine-as --line-width=88 sway-save-tree/sway-save-tree
```

ascent12:
>ascent12's IPC library

What? I don't have one of those. You must be thinking of someone else.

sway-save-tree/sway-save-tree
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python
```
SirCmpwn: `#!/usr/bin/env python3`

SirCmpwn:
I'm still not fond of this idea. I wasn't against this because I didn't want to implement it, but because I didn't want it implemented. The complexity this introduces to the codebase is remarkable.

sway-save-tree/sway-save-tree
```diff
@@ -0,0 +1,112 @@
...
+parser.add_argument('-o', '--output',
+        help='Specifies the output that should be dumped, e.g. LVDS-1')
+parser.add_argument('-v', '--version', action='store_true', help='Version')
+args = parser.parse_args()
+
+if args.version:
+    print('sway-save-tree version %s' % '1.0.0')
+    exit(0)
+
+cmd = ['swaymsg', '-t', 'get_tree']
```
ianyfan: You could also use Python sockets to perform IPC, though it would take about a dozen extra lines.

Also, this should all really be in a `main` function.

RyanDwyer:
> I'm still not fond of this idea. I wasn't against this because I didn't want to implement it, but because I didn't want it implemented. The complexity this introduces to the codebase is remarkable.

In earlier versions of the codebase this was probably the case (before the type safety refactor), but it's in a better state now. Placeholder containers can be regular containers, just with a `NULL` `view` pointer and with a populated `swallows_criteria` list pointer, and should work without any major refactoring.

There is moderate complexity in the parsing of the JSON, but this is confined to `commands/append_layout.c` and doesn't affect the main codebase.

Here's all the changes I can think of that affect the main codebase:

* When rendering, we'll need to check if the container is a placeholder and render it differently. It'll be the same as a normal view, but the surface is just black with a texture showing the raw criteria.
* `view_map` will need to check the `swallows_criteria` and place the view in the matching container. This involves either scanning all the containers for placeholders or storing a list of all placeholder criteria in the `sway_root` struct so we can iterate it in one place.
* Border properties and marks will need to be moved from views to containers so we can set them on the placeholder container, but this should probably be done anyway.
* The `kill` command will need to check if the container is a placeholder and destroy it (rather than try to send a close request to the surface).
* There might be a couple of other checks to see if the container is a placeholder, such as when arranging, doing cursor motion and responding to IPC's `get_tree`, but I'm not sure about this yet.
* When destroying a container, we'll need to free the swallows criteria and criteria texture.

If you're not going to approve of a PR that does this then there's no point in me continuing. There's several people who want this feature though.

SirCmpwn:
I'm going to step out of the discussion. If you get a consensus from the rest of the core contributors, I won't block this patch.

RyanDwyer:
> > When rendering, we'll need to check if the container is a placeholder and render it differently. It'll be the same as a normal view, but the surface is just black with a texture showing the raw criteria.
> 
> Does this involve rendering additional textures?

Yeah. There'll be one new texture on the container struct which displays the criteria string. See this screenshot:

https://i3wm.org/docs/layout-saving-1.png

The texture will be created in the `append_layout` command, destroyed when swallowed or when the container is killed without swallowing, and recreated at a different scale when moving the container to an output of a different scale.

BTW, the screenshot above shows a watch icon. I don't plan on implementing that as part of this PR.

RyanDwyer:
The title bar could be disabled by border config.

I'm happy to show it in the title bar anyway.

Ferk:
Was a decision made on this feature?

Personally, I'm not a huge fan of i3, but I was still drawn to sway because it's the most complete tiling window manager for wayland.

There are essentially only 3 points so far that for my personal use I find uncomfortable in i3, and I believe this feature combined with some scripting would help mitigate the one thing that was so far not possible to workaround sanely: the fact that the tiling in sway is manual and static instead of dynamic (unlike most tiling window managers, like dwm, awesomewm, xmonad..)

For the record, the other two are:

- The lack of a next/prev focus key. This is not complex alt-tab, just a key to cycle all windows in the workspace, I wouldn't really care much which direction as long as it's consistent (this can be scripted, although I still find it annoying that the entire json tree needs to be parsed just to go to the next window, it would be great if a command existed to get the properties and the ids of immediate childs of the currently focused window without having to deal with json, then scripting this would be simpler and could even be done in bash)

- The fact that you either need to have the window title decoration taking space in every single window or no titles at all (this annoyed me a lot 10 years ago in i3.. but since then high res screens have become cheaper and it's now not a big deal, but if the aforementioned command to get current window was added I might try in the future to set up the bar so it shows the currently active window title, like dwm does and then I could disable window titles)

This [reddit comment](https://www.reddit.com/r/archlinux/comments/20q0s0/awesome_wm_vs_i3/cg5u4mj/) highlights and explains word-by-word my mayor gripes, I'm sure there are more people sharing this thoughts. Although I could do without a floating layout mode, personally.

JonnyHaystack:
I am not convinced it's a good idea to include the `sway-save-tree` script as part of sway itself, as that is just a script wrapping existing functionality, but the `append_layout` functionality is definitely important and I would be eager to see it merged.

Basically, layout saving has been done for ages, but layout restoring has never been possible. I don't think that i3 including the `i3-save-tree` script was good because 95% of people (including my past self) didn't realise that they could save layouts properly with IPC, and made extremely hacky solutions around the `i3-save-tree` perl script instead of building the tree themselves.

I maintain [i3-resurrect](https://github.com/JonnyHaystack/i3-resurrect) and would be glad to add sway support or create an alternative for sway, but this is impossible without native support for layout restoring.

Anyways my main question regarding this PR is: Would windows be swallowed only when they are mapped, or would swallowing happen when placeholders are created as well? And would swallowing occur when a window's title changes?

JonnyHaystack:
I am not convinced it's a good idea to include the `sway-save-tree` script as part of sway itself, as that is just a script wrapping existing functionality, but the `append_layout` functionality is definitely important and I would be eager to see it merged.

Basically, layout saving has been done for ages, but layout restoring has never been possible. I don't think that i3 including the `i3-save-tree` script was good because 95% of people (including my past self) didn't realise that they could save layouts properly with IPC, and made extremely hacky solutions around the `i3-save-tree` perl script instead of building the tree themselves.

I maintain [i3-resurrect](https://github.com/JonnyHaystack/i3-resurrect) and would be glad to add sway support or create an alternative for sway, but this is impossible without native support for layout restoring.

Anyways my main question regarding this PR is: Would windows be swallowed only when they are mapped, or would swallowing happen when placeholders are created as well? And would swallowing occur when a window's title changes?

Issue #3024, opened 2018-10-29
Sway 1.0-beta.1 freezes system on startup on void linux


apilat:
* Sway Version: `sway version 1.0-beta.1 (Oct 29 2018, branch 'HEAD')`
* Debug Log: [sway.log](https://gist.github.com/apilat/5b676d983804c13100c66ae875de6b25)
* Configuration File: default

I compiled the latest release from github:
```
git clone https://github.com/swaywm/sway && cd sway
git checkout 1.0-beta.1
meson build
ninja -C build
sudo ninja -C build install
sudo chmod a+s /usr/local/bin/sway
```

Running `sway` causes the system to freeze.
SysRq does still work, however, which may mean that it is just crashing the graphics driver (I am not sure whether this is true, or how to test for it).

Extra info (I'm not sure whether any of this might be useful):
* I am running `void linux` with `musl libc`
* `wlroots-devel` is installed with `xbps` version `wlroots-devel-0.1_2`
* I have 2 GPU's (amd and nvidia)
running with `WLR_DRM_DEVICES` to `card0` or `card1` set causes the program to return having found no GPUs [swaycard.log](https://gist.github.com/apilat/28b95b7635d8a4c504fa2c62241ed46a) even though they are is `/dev/dri`
```
$ ls /dev/dri
by-path card0 card1 renderD128 renderD129
```
* cards are using `amdgpu` and `nouveau` drivers respectivelly
```
$ lspci -nnk
07:00.0 VGA compatible controller [0300]: NVIDIA Corporation GK208B [GeForce GT 710] [10de:128b] (rev a1)
	Subsystem: Micro-Star International Co., Ltd. [MSI] Device [1462:8c93]
	Kernel driver in use: nouveau
	Kernel modules: nouveau
09:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Baffin [Radeon RX 460/560D / Pro 450/455/460/555/555X/560/560X] [1002:67ef] (rev e5)
	Subsystem: Gigabyte Technology Co., Ltd Device [1458:230a]
	Kernel driver in use: amdgpu
	Kernel modules: amdgpu
```
installed `xbps` driver versions are `mesa-ati-dri-18.2.3_1` and `mesa-nouveau-dri-18.2.3_1`

emersion:
It seems that sway crashes. Can you enable coredumps (if not already enabled) and post a backtrace (`bt full` in GDB)?

[...]

PR #3028, opened 2018-10-29
Make it clear that being able to restore root is a failure

emersion:
Related to #3024

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -197,17 +197,18 @@ static void log_kernel(void) {
 static void drop_permissions(void) {
 	if (getuid() != geteuid() || getgid() != getegid()) {
 		if (setgid(getgid()) != 0) {
-			wlr_log(WLR_ERROR, "Unable to drop root");
+			wlr_log(WLR_ERROR, "Unable to drop root, refusing to start");
 			exit(EXIT_FAILURE);
 		}
 		if (setuid(getuid()) != 0) {
-			wlr_log(WLR_ERROR, "Unable to drop root");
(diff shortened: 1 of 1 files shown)


sway/main.c
```diff
@@ -206,7 +206,7 @@ static void drop_permissions(void) {
 		}
 	}
 	if (setuid(0) != -1) {
-		wlr_log(WLR_ERROR, "Root privileges can be restored.");
+		wlr_log(WLR_ERROR, "Root privileges can be restored. Refusing to start.");
```
Alphare: Wouldn't it be clearer to say something like "
```suggestion
		wlr_log(WLR_ERROR, "Dropping root privileges failed. Refusing to start.");
``` ?

File: sway/main.c [hunk 2]
```diff
@@ -194,21 +194,23 @@ static void log_kernel(void) {
...
-			exit(EXIT_FAILURE);
+			wlr_log(WLR_ERROR, "Unable to drop root, refusing to start");
+			return false;
 		}
 	}
 	if (setuid(0) != -1) {
-		wlr_log(WLR_ERROR, "Root privileges can be restored.");
-		exit(EXIT_FAILURE);
+		wlr_log(WLR_ERROR, "Unable to drop root (root privileges can be "
+			"restored, and shouldn't), refusing to start");
```
SirCmpwn: "Unable to drop root - we shouldn't be able to restore it after setuid - refusing to start"

emersion: Done

SirCmpwn:
Thanks!

Issue #2987, opened 2018-10-26
titlebar hiding not respected after closing last sibling


avrelaun:
This one is a bug and a regression from i3 behaviour.

**Required config:**

new_window pixel 2
new_float pixel 2
workspace_layout default
hide_edge_borders smart (this one is probably not be related)

**Current behaviour:**

- Open a window on an empty workspace (titlebar is hidden)
- Open a second window on the same workspace (titlebars are visible)
- Put them into stack or tabbed (titlebars are visible)
- Close one window, leaving only one -> **titlebar is visible but should be hidden**

When there is only one window in a workspace, it should have only one container (previous parents should be removed) and title bar should be hidden if configured so.

RedSoxFan:
Duplicate of #2912

avrelaun:
I still have this issue even though #3031 was merged. Can someone reopen or shall I create another ticket ?

PR #3031, opened 2018-10-30
Restore previous tabbed layout behavior

### LINKED ISSUE #3002: Unable to make a tabbed layout
To reproduce:
Go to a new workspace
Create a new terminal
Set its layout to vertical
Try to set its layout to tabbed
Nothing will happen.

What this prevents:
Having 2 containers, side by side, with one being tabbed.

atomnuker:
As discussed in issue #3002 
Fixes #3002

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -781,13 +781,6 @@ static void render_containers_stacked(struct sway_output *output,
 
 static void render_containers(struct sway_output *output,
 		pixman_region32_t *damage, struct parent_data *parent) {
-	if (parent->children->length == 1) {
-		struct sway_container *child = parent->children->items[0];
-		if (child->view) {
-			render_containers_linear(output, damage, parent);
-			return;
-		}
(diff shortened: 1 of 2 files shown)


RyanDwyer:
I just want to make it clear that what we currently have implemented matches exactly what i3 does, and by merging this we are intentionally diverging from i3's behaviour.

That said, I'm fine with this.

progandy:
It seems some people want this behaviour (#2912, #2987), maybe a configuration option is better like

    hide_lone_tabs

SirCmpwn:
Follow-up ticket for the other behavior would be nice

CodeJoule:
> I just want to make it clear that what we currently have implemented matches exactly what i3 does, and by merging this we are intentionally diverging from i3's behaviour.
> 
> That said, I'm fine with this.

Why is this? It confused me for about an hour, since my i3 config wasn't working as it should with Sway.

Issue #3030, opened 2018-10-30
focus output right doesn't loop


adouzzy:
sway version 1c2a356 (Oct 29 2018, branch 'master')

"focus output right" doesn't loop through all outputs. 
On the other hand, "move workspace to output right" loops.
From my muscle memory, i3 does loop through outputs.

I am not sure if this is a bug or feature request.
Thanks

ascent12:
>I am not sure if this is a bug or feature request.

It's a bug. We try to match i3 behaviour as much as we can, and 'left' seems to work properly.

adouzzy:
Right does work, it moves to the rightest output and then stopped.
I just tried left. It behaves similarly as right.

RyanDwyer:
I started to do this, but having both `enum movement_direction` and `enum wlr_direction` is causing some grief. We have a function in `move.c` called `opposite_direction` which is exactly what we need, but it works with `enum wlr_direction` while the `focus` command uses `enum movement_direction`.

PR #3033, opened 2018-10-30
Remove enum movement_direction

RyanDwyer:
There's no point having both `movement_direction` and `wlr_direction`. This replaces the former with the latter.

As `movement_direction` also contained `MOVE_PARENT` and `MOVE_CHILD` items, these are now checked specifically in the focus command and handled in separate functions, just like the other focus variants.

This is progress towards #3030.

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -175,24 +175,3 @@ failed:
 	free(current);
 	return NULL;
 }
-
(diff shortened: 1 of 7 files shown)


emersion:
Thanks!

PR #3037, opened 2018-10-30
Wrap to fartherest output when running focus output

### LINKED ISSUE #3030: focus output right doesn't loop

RyanDwyer:
Also moves the `opposite_direction` function into `util.c` as it's used in two places now.

Fixes #3030.

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -1,4 +1,5 @@
 #define _XOPEN_SOURCE 700
+#include <assert.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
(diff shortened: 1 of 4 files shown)


SirCmpwn:
Needs rebase

common/util.c
```diff
@@ -138,3 +138,17 @@ bool parse_boolean(const char *boolean, bool current) {
 	// All other values are false to match i3
 	return false;
 }
+
+enum wlr_direction opposite_direction(enum wlr_direction d) {
+	switch (d) {
+		case WLR_DIRECTION_UP:
```
emersion: Style: extra indentation here

common/util.c
```diff
@@ -138,3 +138,17 @@ bool parse_boolean(const char *boolean, bool current) {
...
+			return WLR_DIRECTION_DOWN;
+		case WLR_DIRECTION_DOWN:
+			return WLR_DIRECTION_UP;
+		case WLR_DIRECTION_RIGHT:
+			return WLR_DIRECTION_LEFT;
+		case WLR_DIRECTION_LEFT:
+			return WLR_DIRECTION_RIGHT;
+	}
+	return 0;
```
emersion: `assert(false)`?

RyanDwyer: Done. Note that we can't use `sway_assert` here because this is common code and it can't link to `sway_terminate`.

SirCmpwn:
Thanks!

PR #3040, opened 2018-10-31
Move view border properties to container struct

RyanDwyer:
This will be needed to implement layout saving and restoring, as we need to be able to configure borders on a placeholder container which has no view.


SirCmpwn:
Needs rebase

SirCmpwn:
Thanks!

PR #3041, opened 2018-10-31
Move view marks properties to container struct

RyanDwyer:
Like border properties, this will be needed to implement layout saving and restoring.


RyanDwyer:
This is now ready for review.

PR #3045, opened 2018-10-31
[WIP] Implement seat detach

madblobfish:
I played around with seats and was badly missing this command.
I'm unsure if this is implemented correctly.

It works, but leaves empty seats behind.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -265,6 +265,7 @@ sway_cmd output_cmd_transform;
 
 sway_cmd seat_cmd_attach;
 sway_cmd seat_cmd_cursor;
+sway_cmd seat_cmd_detach;
 sway_cmd seat_cmd_fallback;
 sway_cmd seat_cmd_hide_cursor;
 sway_cmd seat_cmd_pointer_constraint;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -571,6 +571,9 @@ struct seat_attachment_config *seat_config_get_attachment(
 
 struct seat_config *store_seat_config(struct seat_config *seat);
 
+void seat_config_remove_attachment(
+	struct seat_config *seat_config, char *identifier);
+
 int output_name_cmp(const void *item, const void *data);
 
 void output_get_identifier(char *identifier, size_t len,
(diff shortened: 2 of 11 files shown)


madblobfish:
Seems to be stable now. Please review if my memory management is fine.
Should empty seats be removed?

SirCmpwn:
I don't think empty seats should be removed.

SirCmpwn:
Actually, maybe they should.

SirCmpwn:
Does this code actually update the wlroots seat?

madblobfish:
@emersion Ok, I'll write a function for destroying/removing empty seats
@SirCmpwn Not sure, how do I check?

Edit:
I think it does because:
- apply_seat_config is called which calls
- input_manager_apply_seat_config which calls calls some methods few functions further down which seem to directly modify wlroots seats

SirCmpwn:
Okay, cool. And yeah, it should destroy the seat. I assume using any seat command on a nonexistent seat will create that seat, so you could recreate destroyed seats - right?

madblobfish:
Sorry, I could not seat cleanup get it to work properly (see my try 2e839b2).
@SirCmpwn Destroyed seats can be recreated. Not sure if this works in all places.

madblobfish:
So more information:
when removing the last device from a seat i get either a crash or other strange behavior.
Here is my config, the debug log and a stack trace: https://gist.github.com/madblobfish/f6188bf61f7eb97ccdf30a092dd41204

NilsIrl:
I don't know if I am using seats incorrectly but in my (maybe incorrect) use case, this would be nice.

Is there a way to not notify a window that it lost focus? Because the would be a solution that doesn't require seats for my use case.

ddevault:
>Is there a way to not notify a window that it lost focus

wl_keyboard.leave

NilsIrl:
> wl_keyboard.leave

Sorry I am a total noob at Wayland. Am I supposed to make a wayland client here or what configuration file do I need to change?

ddevault:
Oh, you want to be notified via IPC. You could just look to see when a different window is focused and infer that the previously focused window lost it.

NilsIrl:
> Oh, you want to be notified via IPC. You could just look to see when a different window is focused and infer that the previously focused window lost it.

I want to prevent a window from knowing it is out of focus.

ddevault:
No, there's no way to do that. I'm not really interested in adding a way, either.

floyza:
@madblobfish if you're not working on this anymore, is it fine with you if I open a new pull request using your work here?

madblobfish:
@floyza I'm not working on this. You can do what you like with my changes.

PR #3046, opened 2018-10-31
Add relative output transform

tokyovigilante:
This commit enhances the output transform
command with options for a relative transform,
i.e. the provided transform will be applied as
an offset to the current transform. Append
`clockwise` to rotate clockwise from the current
rotation, or `anticlockwise` to rotate in the
opposite direction.

For example, if the output LVDS-1 is rotated 90 degrees
clockwise, the command `output LVDS-1 transform 90 clockwise`
will rotate the display to 180 degrees. All transform
options are supported, including flipped transforms.

### DIFF
--- a/sway/commands/output/transform.c
+++ b/sway/commands/output/transform.c
@@ -1,6 +1,8 @@
 #include <string.h>
 #include "sway/commands.h"
 #include "sway/config.h"
+#include "log.h"
+#include "sway/output.h"
 
 struct cmd_results *output_cmd_transform(int argc, char **argv) {
 	if (!config->handler_context.output_config) {
@@ -10,30 +12,52 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
 		return cmd_results_new(CMD_INVALID, "output",
 			"Missing transform argument.");
 	}
-
-	struct output_config *output = config->handler_context.output_config;
+	enum wl_output_transform transform;
 	if (strcmp(*argv, "normal") == 0) {
-		output->transform = WL_OUTPUT_TRANSFORM_NORMAL;
+		transform = WL_OUTPUT_TRANSFORM_NORMAL;
 	} else if (strcmp(*argv, "90") == 0) {
-		output->transform = WL_OUTPUT_TRANSFORM_90;
+		transform = WL_OUTPUT_TRANSFORM_90;
 	} else if (strcmp(*argv, "180") == 0) {
-		output->transform = WL_OUTPUT_TRANSFORM_180;
+		transform = WL_OUTPUT_TRANSFORM_180;
 	} else if (strcmp(*argv, "270") == 0) {
-		output->transform = WL_OUTPUT_TRANSFORM_270;
+		transform = WL_OUTPUT_TRANSFORM_270;
 	} else if (strcmp(*argv, "flipped") == 0) {
-		output->transform = WL_OUTPUT_TRANSFORM_FLIPPED;
+		transform = WL_OUTPUT_TRANSFORM_FLIPPED;
 	} else if (strcmp(*argv, "flipped-90") == 0) {
-		output->transform = WL_OUTPUT_TRANSFORM_FLIPPED_90;
+		transform = WL_OUTPUT_TRANSFORM_FLIPPED_90;
 	} else if (strcmp(*argv, "flipped-180") == 0) {
-		output->transform = WL_OUTPUT_TRANSFORM_FLIPPED_180;
+		transform = WL_OUTPUT_TRANSFORM_FLIPPED_180;
 	} else if (strcmp(*argv, "flipped-270") == 0) {
-		output->transform = WL_OUTPUT_TRANSFORM_FLIPPED_270;
+		transform = WL_OUTPUT_TRANSFORM_FLIPPED_270;
 	} else {
 		return cmd_results_new(CMD_INVALID, "output",
 			"Invalid output transform.");
 	}
-
+	struct output_config *output = config->handler_context.output_config;
 	config->handler_context.leftovers.argc = argc - 1;
 	config->handler_context.leftovers.argv = argv + 1;
+	if (argc > 1 &&
+			(strcmp(argv[1], "clockwise") == 0 || strcmp(argv[1], "anticlockwise") == 0)) {
+		if (!sway_assert(output->name != NULL, "Output config name not set")) {
+			return NULL;
+		}
+		if (strcmp(output->name, "*") == 0) {
+			return cmd_results_new(CMD_INVALID, "output",
+				"Cannot apply relative transform to all outputs.");
+		}
+		struct sway_output *s_output = output_by_name(output->name);
+		if (s_output == NULL) {
+			return cmd_results_new(CMD_INVALID, "output",
+				"Cannot apply relative transform to unknown output %s", output->name);
+		}
+		if (strcmp(argv[1], "anticlockwise") == 0) {
+			transform = wlr_output_transform_invert(transform);
+		}
+		struct wlr_output *w_output = s_output->wlr_output;
+		transform = wlr_output_transform_compose(w_output->transform, transform);
+		config->handler_context.leftovers.argv += 1;
+		config->handler_context.leftovers.argc -= 1;
+	}
+	output->transform = transform;
 	return NULL;
 }
--- a/sway/sway-output.5.scd
+++ b/sway/sway-output.5.scd
@@ -59,10 +59,13 @@ must be separated by one space. For example:
 	Sets the background of the given output to the specified color. _color_
 	should be specified as _#RRGGBB_. Alpha is not supported.
 
-*output* <name> transform <transform>
+*output* <name> transform <transform> [clockwise|anticlockwise]
 	Sets the background transform to the given value. Can be one of "90", "180",
 	"270" for rotation; or "flipped", "flipped-90", "flipped-180", "flipped-270"
-	to apply a rotation and flip, or "normal" to apply no transform.
+	to apply a rotation and flip, or "normal" to apply no transform. If a single
+	output is chosen and a rotation direction is specified 
+	(_clockwise_ or _anticlockwise_) then the transform is added or 
+	subtracted from the current tranform.
 
 *output* <name> disable|enable
 	Enables or disables the specified output (all outputs are enabled by


tokyovigilante:
Fixed by converting the config to an absolute offset on apply, although possibly not the most elegant way to do so.

SirCmpwn:
I'm wondering if this is a good idea to merge. I think it might be better served with a simple IPC script, which could keep the complexity of sway down.

emersion:
Yeah, I've been wondering about this too. It shouldn't be too hard to get the current value with `swaymsg -t get_outputs` and update it afterwards.

(On the other hand, this PR doesn't introduce a lot of complexity)

include/sway/config.h
```diff
@@ -160,6 +166,7 @@ struct output_config {
 	int x, y;
 	float scale;
 	int32_t transform;
+	enum config_transform_type transform_type;
```
emersion: I really think there's no need for this field. We can use `wlr_output_transform_compose` in `output_cmd_transform`.

tokyovigilante: I'm all for simplifying, but am not sure how then you'd signal that the transform should be relative rather than absolute.

tokyovigilante:
I did think about just writing a script, but a script is probably not very generalisable to other hardware. This way you can respond to any event or have any key or other input change rotation with a one-liner. 

I'd agree with @emersion that this isn't particularly complex as an addition and mostly uses existing wlroots functions, however my patch as it is currently does add another var to the output_config struct, which may not be desirable. I can't see any other way round that though, as we do need to signal through another parameter from the command line (i.e. relative vs absolute and which direction to rotate).

include/sway/config.h
emersion: 1. Get the current output transform
2. `wlr_output_transform_compose(current, relative)`

tokyovigilante: Sure, but are you proposing doing that in `output/transform.c`?

emersion: Yes

tokyovigilante: ok, is `oc->name` set at that point? I assume I'd have to pull in `tree/output` and parse as `apply_output_config` does?

emersion: Yes, `oc->name` is set (you can `sway_assert` it). You can use `output_by_name` to get the `sway_output`.

tokyovigilante: Sounds good, thanks.

SirCmpwn:
Okay, let's merge it. But it needs man page updates first.

sway/commands/output/transform.c
```diff
@@ -32,8 +33,22 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
...
 	config->handler_context.leftovers.argv = argv + 1;
+
+	if (argc == 2) {
+		if (strcmp(argv[1], "clockwise") == 0) {
+			output->transform_type = TRANSFORM_RELATIVE_CLOCKWISE;
+		} else if (strcmp(argv[1], "anticlockwise") == 0) {
+			output->transform_type = TRANSFORM_RELATIVE_ANTICLOCKWISE;
+		} else {
+			return cmd_results_new(CMD_INVALID, "output",
+				"Invalid output transform type.");
```
emersion: This doesn't work if you chain output commands like so:

```
output XXX transform 90 scale 2
```

tokyovigilante: Oh right, will tidy the detection up.

tokyovigilante:
Thanks, I'll fix the bugs, move all the code into `config/transform.c` and add documentation.

tokyovigilante:
Do I need to worry about wildcards?

emersion:
Gah, I didn't think about this one. We can't really implement those since we only produce one output config. I'd say its fine not to support them for this command.

tokyovigilante:
From a code change point of view this should be feature complete. 

I've noticed one bug involving spurious output from swaymsg if an invalid transform command is passed, which may be due to my handling of the `argv`/`argc` offsets but not entirely sure how to fix. Handled multiple commands and wildcards however. 

Will update the man page once everyone's happy.

sway/commands/output/transform.c
```diff
@@ -1,6 +1,9 @@
+
```
emersion: Style: extra line here

sway/commands/output/transform.c
```diff
@@ -10,30 +13,49 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
 		return cmd_results_new(CMD_INVALID, "output",
 			"Missing transform argument.");
 	}
-
-	struct output_config *output = config->handler_context.output_config;
+	int32_t transform = 0;
```
emersion: This can be an enum, this doesn't need to be initialized.

sway/commands/output/transform.c
```diff
@@ -10,30 +13,49 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
...
 		return cmd_results_new(CMD_INVALID, "output",
 			"Invalid output transform.");
 	}
-
+	struct output_config *output = config->handler_context.output_config;
 	config->handler_context.leftovers.argc = argc - 1;
 	config->handler_context.leftovers.argv = argv + 1;
+
+	if (argc > 1 && (strstr(argv[1], "clockwise") != NULL)) {
+		sway_assert(output->name != NULL, "Output config name not set");
```
emersion: `sway_assert` is a little weird and its return value needs to be checked, e.g.:

```c
if (!sway_assert(output->name != NULL, "Output config name not set")) {
  return NULL;
}
```

sway/commands/output/transform.c
```diff
@@ -10,30 +13,49 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
...
+		if (strcmp(output->name, "*") == 0) {
+			return cmd_results_new(CMD_INVALID, "output",
+				"Cannot apply relative transform to all outputs.");
+		}
+		struct sway_output *s_output = output_by_name(output->name);
+		struct wlr_output *w_output = s_output->wlr_output;
+		sway_assert(output != NULL, "No output named %s", output->name);
```
emersion: This shouldn't be an assertion, otherwise you'll crash the whole compositor if you execute `swaymsg output IDONTEXIST transform 90 clockwise`.

sway/commands/output/transform.c
```diff
@@ -10,30 +13,49 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
...
 	} else {
 		return cmd_results_new(CMD_INVALID, "output",
 			"Invalid output transform.");
 	}
-
+	struct output_config *output = config->handler_context.output_config;
 	config->handler_context.leftovers.argc = argc - 1;
 	config->handler_context.leftovers.argv = argv + 1;
+
+	if (argc > 1 && (strstr(argv[1], "clockwise") != NULL)) {
```
emersion: What happens if you execute `output XXX transform 90 asdfclockwise`? It should error out.

tokyovigilante: Have altered to check specifically for `clockwise`/`anticlockwise`. I don't mind what these are called (left and right could be used, but using clockwise/anticlockwise seems to have the right amount of brevity and obviousness to me.

sway/commands/output/transform.c
```diff
@@ -10,30 +12,57 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
 		return cmd_results_new(CMD_INVALID, "output",
 			"Missing transform argument.");
 	}
-
-	struct output_config *output = config->handler_context.output_config;
+	enum wl_output_transform transform = 0;
```
emersion: No need to initialize this to zero.

sway/commands/output/transform.c
```diff
@@ -10,30 +12,57 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
...
 		return cmd_results_new(CMD_INVALID, "output",
 			"Invalid output transform.");
 	}
-
+	struct output_config *output = config->handler_context.output_config;
 	config->handler_context.leftovers.argc = argc - 1;
 	config->handler_context.leftovers.argv = argv + 1;
+
+	if (argc > 1 &&
+		(strcmp(argv[1], "clockwise") == 0 || strcmp(argv[1], "anticlockwise") == 0))
```
emersion: Style: no newline before brace, two indent levels for multiline `if`

```c
if (thing &&
		other_thing) {
	do_stuff();
}
```

sway/commands/output/transform.c
```diff
@@ -10,30 +12,57 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
...
+	{
+		if (!sway_assert(output->name != NULL, "Output config name not set")) {
+		  return NULL;
```
emersion: Style: spaces → tabs

sway/commands/output/transform.c
```diff
@@ -10,30 +12,57 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
...
+			return cmd_results_new(CMD_INVALID, "output",
+				"Cannot apply relative transform to all outputs.");
+		}
+		struct sway_output *s_output = output_by_name(output->name);
+		if (s_output == NULL) {
+			return cmd_results_new(CMD_INVALID, "output",
+				"Cannot apply relative transform to unknown output %s", output->name);
+		}
+		struct wlr_output *w_output = s_output->wlr_output;
+		if (!sway_assert(w_output != NULL, "No output named %s", output->name)) {
```
emersion: I don't think this check is necessary. `sway_output` always has a non-NULL `wlr_output`

emersion:
Ah, and can you squash all of these commits into one?

sway/commands/output/transform.c
tokyovigilante: Fixed.

sway/commands/output/transform.c
tokyovigilante: Fixed.

sway/commands/output/transform.c
tokyovigilante: Fixed, I think... Should all be tabs.

sway/commands/output/transform.c
tokyovigilante: Ok, removed.

sway/commands/output/transform.c
```diff
@@ -10,30 +12,53 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
...
 		return cmd_results_new(CMD_INVALID, "output",
 			"Invalid output transform.");
 	}
-
+	struct output_config *output = config->handler_context.output_config;
 	config->handler_context.leftovers.argc = argc - 1;
 	config->handler_context.leftovers.argv = argv + 1;
+	if (argc > 1 &&
+			(strcmp(argv[1], "clockwise") == 0 || strcmp(argv[1], "anticlockwise") == 0))
+	{
```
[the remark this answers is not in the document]
tokyovigilante: Oh right, sorry.

tokyovigilante:
Sure if you're happy with the code and using `clockwise`/`anticlockwise` as verbs.

tokyovigilante:
Man page updated.

emersion:
Thanks!

tokyovigilante:
Thank you! Much appreciated.

PR #3048, opened 2018-11-01
Reword message about Nvidia drivers to be less snarky

ianyfan:
Mostly an opinionated request (so feel free to close), but I feel like the flag 'my-next-gpu-wont-be-nvidia' is unnecessarily snarky and a bit antagonistic towards those users, especially since the flag 'unsupported-gpu' already does the same thing without feeling so condescending.

I haven't removed the flag for users who are already using it, but I've just changed the message about it to not mention it.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -88,38 +88,26 @@ void detect_proprietary(int allow_unsupported_gpu) {
 		return;
 	}
 	while (!feof(f)) {
-		char *line;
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks, but given that this only shows up when the *proprietary* driver is loaded - a situation we expressly do not support - I'm not interested in improving the messaging.

PR #3049, opened 2018-11-02
Swayidle fixes

ianyfan:


swayidle/main.c
```diff
@@ -432,10 +433,13 @@ int main(int argc, char *argv[]) {
 	struct wl_event_source *source = wl_event_loop_add_fd(state.event_loop,
 		wl_display_get_fd(state.display), WL_EVENT_READABLE,
 		display_event, NULL);
-	wl_event_source_check(source);
```
emersion: I don't think this is right.

ianyfan: I pushed a different fix, namely changing `wl_display_dispatch` to `wl_display_dispatch_pending` so that it doesn't block.

swayidle/main.c
```diff
@@ -368,7 +369,7 @@ static int display_event(int fd, uint32_t mask, void *data) {
 	if (mask & WL_EVENT_HANGUP) {
 		sway_terminate(0);
 	}
-	if (wl_display_dispatch(state.display) < 0) {
+	if (wl_display_dispatch_pending(state.display) < 0) {
```
emersion: I don't think this works either. `wl_display_dispatch_pending`'s description says:

> Dispatch default queue events without reading from the display fd. 

I think we need to use `wl_display_read_events`. See how mako does it:

https://github.com/emersion/mako/blob/master/event-loop.c#L238

VirgiliusHaufniensis:
This definitely solves the problem for me! But it introduces another bug, namely that the timeout parameters no longer work. Sorry if this goes the wrong place, I'm new to Github.

swayidle/main.c
```diff
@@ -368,11 +369,17 @@ static int display_event(int fd, uint32_t mask, void *data) {
 	if (mask & WL_EVENT_HANGUP) {
 		sway_terminate(0);
 	}
-	if (wl_display_dispatch(state.display) < 0) {
-		wlr_log_errno(WLR_ERROR, "wl_display_dispatch failed, exiting");
-		sway_terminate(0);
+	while (wl_display_prepare_read(state.display) != 0) {
```
emersion: This needs to be called before polling, ie. in the main loop. Docs say:

>Use this function before polling on the display fd

emersion: (See https://people.freedesktop.org/~whot/wayland-doxygen/wayland/Client/classwl__display.html#a40039c1169b153269a3dc0796a54ddb0 for a complete example)

ianyfan: So move the `prepare_read` to the main loop? But then how would we know whether to `cancel_read`? (which I forgot to put in my attempt either, but eh)

ianyfan: Also, why doesn't this use the default event loop?

emersion: >But then how would we know whether to cancel_read? 

We could probably have a `bool read_wl_events` variable set to `false` before dispatching, set it to `true` if we read from the Wayland socket, and call `wl_display_cancel_read` after disptaching if `read_wl_events` is still `false`.

>Also, why doesn't this use the default event loop?

Which default event loop?

We use the event loop from libwayland-server, but the `wl_display` doesn't have an event loop because we're a client. We should probably migrate to our own event loop instead.

swayidle/main.c
ianyfan: I think removing this line is right now, because it's causing double `display_event` each time.

swayidle/main.c
ianyfan: I've done that now.
And I was thinking about the default event queue, sorry.

swayidle/main.c
emersion: This is needed if an event happens somewhere else (dbus, signal) and sends Wayland messages.

ianyfan: But then why won't they cause an event on polling first time round?

The alternative is to check whether `read_wl_events` has been set in `display_event` to prevent trying to read again.

emersion: Actually I think we're not using these functions correctly. I just read how Weston does it for its Wayland backend. The event loop is just `wl_event_loop_dispatch` in a loop. The `display_event` handler uses the mask to know what to do: https://gitlab.freedesktop.org/wayland/weston/blob/master/libweston/compositor-wayland.c#L2593

They also use `wl_event_source_check`.

Can you try doing the same?

>But then why won't they cause an event on polling first time round?

Because "I need to send data on the Wayland socket" is a consequence of "I've handled an D-Bus event", and the event loop isn't aware of that without `wl_event_source_check`.

ianyfan: I did it like in the link, though I omitted checking for `WL_EVENT_WRITABLE` because I thought that we're only checking for readable events. Is my understanding incorrect?
Also, I kept the count variable in, that's correct, right?

ianyfan: Yes, because https://github.com/swaywm/sway/blob/eda3bfeed5097c71634332bfe998188b028abf02/swayidle/main.c#L433 means it only triggers on readable events? or is my thinking wrong?

emersion: Ah, right. We should add `| WL_EVENT_WRITABLE` then. It probably doesn't happen right now because we don't fill the receiver buffers, but it's probably best to handle it properly.

emersion: https://gitlab.freedesktop.org/wayland/weston/issues/163

ianyfan:
Yeah sorry, I was planning on waiting to see if the Weston issue would get any feedback but then forgot about it.

emersion:
No worries :)

Are you okay with the changes I made?

ianyfan:
LGTM

emersion:
Thanks!

PR #3052, opened 2018-11-02
Correct "Current mode" in `swaymsg -t get_outputs` for scaled outputs

### LINKED ISSUE #3051: swaymsg get_outputs show inaccurate "Current mode"
I've been re-applying my Sway config to try to figure out what is going on with position of my monitors (very specific combinations work, while other positions that look good on paper don't actually work).

Anyway, during this I noticed this:

```
Output eDP-1 'Sharp Corporation 0x148B 0x00000000'
  Current mode: 1920x1080 @ 59.997002 Hz
  Position: 0,0
  Scale factor: 2x
  Transform: normal
  Workspace: 4
  Available modes:
    3840x2160 @ 47.997002 Hz
    3840x2160 @ 59.997002 Hz
```

I can say from staring at the monitor at this moment, that display is not running at 1920x1080 mode.

Edit: debug log: [sway.log](https://github.com/swaywm/sway/files/2541320/sway.log)

colemickens:
Fixes #3051.

Fixed output example:

```
Output eDP-1 'Sharp Corporation 0x148B 0x00000000'
  Current mode: 3840x2160 @ 59.997002 Hz
  Position: 3440,0
  Scale factor: 2x
  Transform: normal
  Workspace: 1
  Available modes:
    3840x2160 @ 47.997002 Hz
    3840x2160 @ 59.997002 Hz
```

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -136,12 +136,17 @@ static void ipc_json_describe_output(struct sway_output *output,
 			json_object_new_int(mode->width));
 		json_object_object_add(mode_object, "height",
 			json_object_new_int(mode->height));
-		json_object_object_add(mode_object, "refresh",
-			json_object_new_int(mode->refresh));
 		json_object_array_add(modes_array, mode_object);
 	}
 
 	json_object_object_add(object, "modes", modes_array);
+
+	json_object *current_mode_object = json_object_new_object();
+	json_object_object_add(current_mode_object, "width",
+		json_object_new_int(wlr_output->width));
+	json_object_object_add(current_mode_object, "height",
+		json_object_new_int(wlr_output->height));
+	json_object_object_add(object, "current_mode", current_mode_object);
 
 	struct sway_node *parent = node_get_parent(&output->node);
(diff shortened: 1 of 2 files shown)


emersion:
Looks good. Can we remove the `refresh` property which is directly in the `output` JSON object? `current_mode.refresh` is the same value.

ianyfan:
Is it also possible to change the scale factor to be a double? Since it is currently an integer, it displays the wrong value when the output is fractionally scaled.

colemickens:
Made the requested changes. Maybe we should change `current_mode` in the json to something like `current_dim[ensions]` or `current_resolution`, or maybe it doesn't matter; let me know.

SirCmpwn:
Thanks!

emersion:
Why has refresh been removed from mode objects?

colemickens:
@emersion Late night edits? I did an amend where I removed "refresh" after I thought I'd already removed it from `current_mode` as requested. I must've just gotten it in the wrong place one of the times. Sending a follow-up PR now...

emersion:
I suggested to remove it from the output object. In other words to move it from the output object to the current mode object.

PR #3057, opened 2018-11-03
Render LAYER_TOP surfaces above fullscreen

vilhalmer:
This allows e.g. mako to be visible over a fullscreen container.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -936,16 +936,19 @@ void output_render(struct sway_output *output, struct timespec *when,
 	}
 
 	struct sway_container *fullscreen_con = workspace->current.fullscreen;
+
+	static const float clear_color[] = {0.25f, 0.25f, 0.25f, 1.0f};
+	static const float fullscreen_clear_color[] = {0.0f, 0.0f, 0.0f, 1.0f};
+
+	int nrects;
+	pixman_box32_t *rects = pixman_region32_rectangles(damage, &nrects);
+	for (int i = 0; i < nrects; ++i) {
+		scissor_output(wlr_output, &rects[i]);
+		wlr_renderer_clear(renderer,
+				fullscreen_con ? fullscreen_clear_color : clear_color);
+	}
+
 	if (fullscreen_con) {
-		float clear_color[] = {0.0f, 0.0f, 0.0f, 1.0f};
-
-		int nrects;
-		pixman_box32_t *rects = pixman_region32_rectangles(damage, &nrects);
-		for (int i = 0; i < nrects; ++i) {
-			scissor_output(wlr_output, &rects[i]);
-			wlr_renderer_clear(renderer, clear_color);
-		}
-
 		// TODO: handle views smaller than the output
 		if (fullscreen_con->view) {
 			if (fullscreen_con->view->saved_buffer) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
This is deliberate.

martinetd:
That's basically forcing mako to use the overlay layer if someone wants to display over fullscreen, which may or may not go above a lockscreen depending on the compositor; I'm not sure that's ideal. (I don't think they have any way of telling if a lockscreen/other overlay window covering the whole screen is active; did I miss something that'd make this possible?)


I initially thought this was on purpose as well (for e.g. easier direct scan-out) but we don't do it anyway, we even have to map unmanaged xwayland window because of X11 context menus (and with this probably come popups as well?); was this discussed somewhere I could find traces of to refresh my memory?
The protocol only says "Fullscreen shell surfaces are typically rendered at the top layer", which could arbitrarily display part of the top layer or not at the discretion of the compositor (because "ordering within a single layer is undefined"), so I believe it's not set in stone.

(I won't insist more than that if you repeat the same thing, I'm mostly curious what the rationale was)

emersion:
>That's basically forcing mako to use the overlay layer if someone wants to display over fullscreen

Yeah, this has been implemented in mako. I'm not too happy about it but I'm not sure there's a better solution.

>(I don't think they have any way of telling if a lockscreen/other overlay window covering the whole screen is active; did I miss something that'd make this possible?)

Compositors have means to decide whether a lockscreen is active or not, e.g. through the layer-shell namespace or the wlr-input-inhibit protocol.

>we don't do it anyway

It's something I'd like to work on.

>we even have to map unmanaged xwayland window because of X11 context menus

Usually there's no X11 unmanaged window.

>was this discussed somewhere I could find traces of to refresh my memory?

Hmm, I don't remember.

martinetd:
>> (I don't think they have any way of telling if a lockscreen/other overlay window covering the whole screen is active; did I miss something that'd make this possible?)
> 
> Compositors have means to decide whether a lockscreen is active or not, e.g. through the layer-shell namespace or the wlr-input-inhibit protocol.

Hmm if we can identify one of the shell in the overlay layer as a lockscreen surface and display only that one within the layer I guess that would make me less reluctant with the principle, but I'm somehow not convinced this will be done :P

>> direct scan-out
>
> It's something I'd like to work on.

Even if you do, you'll need to disable it if there are unmanaged x11 window, subshells, something in the overlay layer... Not much different to also check for the top layer; that wasn't really a good argument to not do it, just the idea I had.

emersion:
>Hmm if we can identify one of the shell in the overlay layer as a lockscreen surface and display only that one within the layer I guess that would make me less reluctant with the principle, but I'm somehow not convinced this will be done :P

It's already done if the lockscreen is opaque. We don't render/send frame events to surfaces underneath.

>Not much different to also check for the top layer; that wasn't really a good argument to not do it, just the idea I had.

Yeah, this could have been a sway setting.

SirCmpwn:
Maybe we should define a new layer.

progandy:
Or maybe there should be a mechanism to configure the window order depending on the layer-shell ```namespace```? Perhaps with some standardized values for the namespace?

vilhalmer:
My rationale for the PR was based on considering a fullscreen container to be at the same layer as all of the other shell surfaces in the workspace, rather than over top them. (It "replaces" them while in fullscreen.) Then it makes sense for TOP layer-shell surfaces to still be higher, without needing to define another layer. This doesn't match the current wording in the protocol, though, and I understand the advantages of having fullscreen surfaces "win" more often in order to gain rendering optimizations.

emersion:
It's not only for optimizations. It's also very annoying when you watch a movie/do a presentation and a notification pops in.

vilhalmer:
True, though I'd often rather be able to control that on a case-by-case basis via the notification daemon.

PR #3058, opened 2018-11-03
Implement layout-independent keysym bindings

### LINKED ISSUE #2999: bindsym does not work in some keyboard layouts
`bindsym` bindings with letters (like `$mod+w`) do not work in some keyboard layouts (e.g. `ru`).

The issue is in the way these shortcuts are processed. For `bindsym`, sway compares key syms, and key syms for `w` are different in these two keyboard layouts (119 for `w`, 1731 for `ц`).

Obviously having to switch keyboard layout to use a shortcut is inconvenient. It is possible to workaround this and use `bindcode` for all bindings, but the config file will become less readable.

I also checked i3, and `bindsym` worked fine in any keyboard layout.

It seems that the easiest way to fix it is to convert key syms into key codes while reading the config file, and then only check key codes on keyboard events.

If it sounds like a reasonable idea, I could look into this more and probably prepare a pull request, though it may take some time.

kupospelov:
Fixes #2999.

The idea is to have an instance of default xkb_keymap and try to resolve bindsyms into bindcodes while binding.

If the conversion is successful, we just treat the binding as a bindcode. If the conversion has failed, we do not change the binding, so it will be processed as a normal bindsym.

In my configuration, all of the 80 bindsyms have been converted into bindcodes. I have tested the following and it worked in any keyboard layout:
* Default bindings
* Bindings with function keys
* Uppercase keysyms
* Special ones like XF86AudioRaiseVolume, XF86AudioMute, XF86MonBrightnessUp

I also checked a keysym with no simple keycode translation (the question mark, keysym `question`) and it worked as a keysym binding, however this kind of bindings still depends on the current keyboard layout.

This makes shortcut behavior more consistent with that of i3 and other programs (e.g. terminal emulators).

### DIFF
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -24,6 +24,7 @@ void free_sway_binding(struct sway_binding *binding) {
 	}
 
 	list_free_items_and_destroy(binding->keys);
+	list_free_items_and_destroy(binding->syms);
 	free(binding->input);
 	free(binding->command);
 	free(binding);
@@ -249,31 +250,41 @@ static struct cmd_results *switch_binding_remove(
 			switchcombo);
 }
 
-static struct cmd_results *binding_add(struct sway_binding *binding,
-		list_t *mode_bindings, const char *bindtype,
-		const char *keycombo, bool warn) {
-	// overwrite the binding if it already exists
-	bool overwritten = false;
+/**
+ * Insert or update the binding.
+ * Return the binding which has been replaced or NULL.
+ */
+static struct sway_binding *binding_upsert(struct sway_binding *binding,
+		list_t *mode_bindings) {
 	for (int i = 0; i < mode_bindings->length; ++i) {
 		struct sway_binding *config_binding = mode_bindings->items[i];
 		if (binding_key_compare(binding, config_binding)) {
-			sway_log(SWAY_INFO, "Overwriting binding '%s' for device '%s' "
-					"to `%s` from `%s`", keycombo, binding->input,
-					binding->command, config_binding->command);
-			if (warn) {
-				config_add_swaynag_warning("Overwriting binding"
-						"'%s' for device '%s' to `%s` from `%s`",
-						keycombo, binding->input, binding->command,
-						config_binding->command);
-			}
-			free_sway_binding(config_binding);
 			mode_bindings->items[i] = binding;
-			overwritten = true;
+			return config_binding;
 		}
 	}
 
-	if (!overwritten) {
-		list_add(mode_bindings, binding);
+	list_add(mode_bindings, binding);
+	return NULL;
+}
+
+static struct cmd_results *binding_add(struct sway_binding *binding,
+		list_t *mode_bindings, const char *bindtype,
+		const char *keycombo, bool warn) {
+	struct sway_binding *config_binding = binding_upsert(binding, mode_bindings);
+
+	if (config_binding) {
+		sway_log(SWAY_INFO, "Overwriting binding '%s' for device '%s' "
+				"to `%s` from `%s`", keycombo, binding->input,
+				binding->command, config_binding->command);
+		if (warn) {
+			config_add_swaynag_warning("Overwriting binding"
+					"'%s' for device '%s' to `%s` from `%s`",
+					keycombo, binding->input, binding->command,
+					config_binding->command);
+		}
+		free_sway_binding(config_binding);
+	} else {
 		sway_log(SWAY_DEBUG, "%s - Bound %s to command `%s` for device '%s'",
 				bindtype, keycombo, binding->command, binding->input);
 	}
@@ -329,7 +340,6 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 	bool exclude_titlebar = false;
 	bool warn = true;
 
-	// Handle --release and --locked
 	while (argc > 0) {
 		if (strcmp("--release", argv[0]) == 0) {
 			binding->flags |= BINDING_RELEASE;
@@ -339,6 +349,10 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 			binding->flags |= BINDING_BORDER | BINDING_CONTENTS | BINDING_TITLEBAR;
 		} else if (strcmp("--border", argv[0]) == 0) {
 			binding->flags |= BINDING_BORDER;
+		} else if (strcmp("--to-code", argv[0]) == 0) {
+			if (!bindcode) {
+				binding->flags |= BINDING_CODE;
+			}
 		} else if (strcmp("--exclude-titlebar", argv[0]) == 0) {
 			exclude_titlebar = true;
 		} else if (strncmp("--input-device=", argv[0],
@@ -409,6 +423,12 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 
 	// sort ascending
 	list_qsort(binding->keys, key_qsort_cmp);
+
+	// translate keysyms into keycodes
+	if (!translate_binding(binding)) {
+		sway_log(SWAY_INFO,
+				"Unable to translate bindsym into bindcode: %s", argv[0]);
+	}
 
 	list_t *mode_bindings;
 	if (binding->type == BINDING_KEYCODE) {
@@ -564,5 +584,116 @@ void seat_execute_command(struct sway_seat *seat, struct sway_binding *binding) 
 	list_free(res_list);
 	if (success) {
 		ipc_event_binding(binding);
+	}
+}
+
+/**
+ * The last found keycode associated with the keysym
+ * and the total count of matches.
+ */
+struct keycode_matches {
+	xkb_keysym_t keysym;
+	xkb_keycode_t keycode;
+	int count;
+};
+
+/**
+ * Iterate through keycodes in the keymap to find ones matching
+ * the specified keysym.
+ */
+static void find_keycode(struct xkb_keymap *keymap,
+		xkb_keycode_t keycode, void *data) {
+	xkb_keysym_t keysym = xkb_state_key_get_one_sym(
+			config->keysym_translation.xkb_state, keycode);
+
+	if (keysym == XKB_KEY_NoSymbol) {
+		return;
+	}
+
+	struct keycode_matches *matches = data;
+	if (matches->keysym == keysym) {
+		matches->keycode = keycode;
+		matches->count++;
+	}
+}
+
+/**
+ * Return the keycode for the specified keysym.
+ */
+static struct keycode_matches get_keycode_for_keysym(xkb_keysym_t keysym) {
+	struct keycode_matches matches = {
+		.keysym = keysym,
+		.keycode = XKB_KEYCODE_INVALID,
+		.count = 0,
+	};
+
+	xkb_keymap_key_for_each(config->keysym_translation.xkb_keymap,
+			find_keycode, &matches);
+	return matches;
+}
+
+bool translate_binding(struct sway_binding *binding) {
+	if ((binding->flags & BINDING_CODE) == 0) {
+		return true;
+	}
+
+	switch (binding->type) {
+	// a bindsym to translate
+	case BINDING_KEYSYM:
+		binding->syms = binding->keys;
+		binding->keys = create_list();
+		break;
+	// a bindsym to re-translate
+	case BINDING_KEYCODE:
+		list_free_items_and_destroy(binding->keys);
+		binding->keys = create_list();
+		break;
+	default:
+		return true;
+	}
+
+	for (int i = 0; i < binding->syms->length; ++i) {
+		xkb_keysym_t *keysym = binding->syms->items[i];
+		struct keycode_matches matches = get_keycode_for_keysym(*keysym);
+
+		if (matches.count != 1) {
+			sway_log(SWAY_INFO, "Unable to convert keysym %d into"
+					" a single keycode (found %d matches)",
+					*keysym, matches.count);
+			goto error;
+		}
+
+		xkb_keycode_t *keycode = malloc(sizeof(xkb_keycode_t));
+		if (!keycode) {
+			sway_log(SWAY_ERROR, "Unable to allocate memory for a keycode");
+			goto error;
+		}
+
+		*keycode = matches.keycode;
+		list_add(binding->keys, keycode);
+	}
+
+	list_qsort(binding->keys, key_qsort_cmp);
+	binding->type = BINDING_KEYCODE;
+	return true;
+
+error:
+	list_free_items_and_destroy(binding->keys);
+	binding->type = BINDING_KEYSYM;
+	binding->keys = binding->syms;
+	binding->syms = NULL;
+	return false;
+}
+
+void binding_add_translated(struct sway_binding *binding,
+		list_t *mode_bindings) {
+	struct sway_binding *config_binding =
+		binding_upsert(binding, mode_bindings);
+
+	if (config_binding) {
+		sway_log(SWAY_INFO, "Overwriting binding for device '%s' "
+				"to `%s` from `%s`", binding->input,
+				binding->command, config_binding->command);
+		free_sway_binding(config_binding);
 	}
 }
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -604,7 +604,7 @@ struct keycode_matches {
 static void find_keycode(struct xkb_keymap *keymap,
 		xkb_keycode_t keycode, void *data) {
 	xkb_keysym_t keysym = xkb_state_key_get_one_sym(
-			config->keysym_translation.xkb_state, keycode);
+			config->keysym_translation_state, keycode);
 
 	if (keysym == XKB_KEY_NoSymbol) {
 		return;
@@ -627,7 +627,8 @@ static struct keycode_matches get_keycode_for_keysym(xkb_keysym_t keysym) {
 		.count = 0,
 	};
 
-	xkb_keymap_key_for_each(config->keysym_translation.xkb_keymap,
+	xkb_keymap_key_for_each(
+			xkb_state_get_keymap(config->keysym_translation_state),
 			find_keycode, &matches);
 	return matches;
 }
--- a/sway/commands/input.c
+++ b/sway/commands/input.c
@@ -39,6 +39,30 @@ static struct cmd_handler input_config_handlers[] = {
 	{ "xkb_numlock", input_cmd_xkb_numlock },
 };
 
+/**
+ * Re-translate keysyms if a change in the input config could affect them.
+ */
+static void retranslate_keysyms(struct input_config *input_config) {
+	bool matched = false;
+	for (int i = 0; i < config->input_configs->length; ++i) {
+		struct input_config *ic = config->input_configs->items[i];
+		matched |= ic->identifier == input_config->identifier;
+
+		// the first configured xkb_layout
+		if (ic->xkb_layout) {
+			if (matched) {
+				translate_keysyms(ic->xkb_layout);
+			}
+
+			// nothing has changed
+			return;
+		}
+	}
+
+	// no xkb_layout has been set, restore the default
+	translate_keysyms(getenv("XKB_DEFAULT_LAYOUT"));
+}
+
 struct cmd_results *cmd_input(int argc, char **argv) {
 	struct cmd_results *error = NULL;
 	if ((error = checkarg(argc, "input", EXPECTED_AT_LEAST, 2))) {
@@ -73,6 +97,7 @@ struct cmd_results *cmd_input(int argc, char **argv) {
 			store_input_config(config->handler_context.input_config);
 
 		input_manager_apply_input_config(ic);
+		retranslate_keysyms(ic);
 	} else {
 		free_input_config(config->handler_context.input_config);
 	}
--- a/sway/commands/input.c
+++ b/sway/commands/input.c
@@ -43,24 +43,17 @@ static struct cmd_handler input_config_handlers[] = {
  * Re-translate keysyms if a change in the input config could affect them.
  */
 static void retranslate_keysyms(struct input_config *input_config) {
-	bool matched = false;
 	for (int i = 0; i < config->input_configs->length; ++i) {
 		struct input_config *ic = config->input_configs->items[i];
-		matched |= ic->identifier == input_config->identifier;
-
-		// the first configured xkb_layout
 		if (ic->xkb_layout) {
-			if (matched) {
+			// this is the first config with xkb_layout
+			if (ic->identifier == input_config->identifier) {
 				translate_keysyms(ic->xkb_layout);
 			}
 
-			// nothing has changed
 			return;
 		}
 	}
-
-	// no xkb_layout has been set, restore the default
-	translate_keysyms(getenv("XKB_DEFAULT_LAYOUT"));
 }
 
 struct cmd_results *cmd_input(int argc, char **argv) {
--- a/sway/commands/input.c
+++ b/sway/commands/input.c
@@ -48,7 +48,7 @@ static void retranslate_keysyms(struct input_config *input_config) {
 		if (ic->xkb_layout) {
 			// this is the first config with xkb_layout
 			if (ic->identifier == input_config->identifier) {
-				translate_keysyms(ic->xkb_layout);
+				translate_keysyms(ic);
 			}
 
 			return;
--- a/sway/config.c
+++ b/sway/config.c
@@ -340,7 +340,7 @@ static void config_defaults(struct sway_config *config) {
 
 	// The keysym to keycode translation
 	config->keysym_translation_state =
-		keysym_translation_state_create(getenv("XKB_DEFAULT_LAYOUT"));
+		keysym_translation_state_create(NULL);
 
 	return;
 cleanup:
--- a/sway/config.c
+++ b/sway/config.c
@@ -33,7 +33,7 @@
 
 struct sway_config *config = NULL;
 
-static struct keysym_translation_data new_keysym_translation_data(
+static struct xkb_state *keysym_translation_state_create(
 		const char *layout) {
 	struct xkb_rule_names rules = {
 		.layout = layout,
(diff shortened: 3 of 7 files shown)


kupospelov:
I would say that in one layout keyboards should only produce one unique keysym per key most of the time. From what I see, even numpad keys (digits, enter, insert, etc) produce keysyms which are different from the other keys' keysyms.

In the rare case when the same keysym is producible from different keys, we could just skip the conversion. In other words, we will only convert keysyms which can be replaced with a single keycode. This should still fix bindings with Latin characters, since it is hard to imagine a keyboard with several keys to type in the same letter in one layout, and keysyms which cannot be safely replaced won't be affected.

progandy:
Maybe this should be optional, let ```bindsym``` detect a parameter like ```--code``` to switch the behaviour? Or do the reverse with ```bindcode --fromsym```

If you use two layouts with different positions for the latin letters (e.g. ```de``` and ```us```) I personally prefer the binding follows the new layout.

kupospelov:
> it seems muscle memory trumps my preference. I quickly switch z and y, but for the shortcuts I keep pressing the original position.

This is exactly what I experience from time to time (except that in my case almost all shortcuts just stop working). People do not normally think about the current layout all the time, so this inconsistency is annoying.

> Maybe this should be optional, let bindsym detect a parameter like --code to switch the behaviour? Or do the reverse with bindcode --fromsym

Making the bindsym conversion optional would mean that there are cases when people use some shortcuts in one keyboard layout and do not use in another. I cannot say for everyone, but to me it seems weird.

In my opinion, it is convenient to use keysyms in the config file, but they should not be interpreted literally. Instead, bindings should be bound to proper hardware keys and just work consistently in any layout.

This is not exactly what this pull request does, though. This change fixes the problem, but does not eliminate layout-dependent keysym comparisons altogether. I do not think I can do this until there is an agreement to drop support for layout-dependent keysym bindings.

Hi-Angel:
Hmm this is odd as the `View changes` link give not the commit I reviewed. I mean, it would probably make sense to merge the commit with style fixes into the prev. one.

kupospelov:
Squashed the style fixes and rebased the changes.

emersion:
>I would say that in one layout keyboards should only produce one unique keysym per key most of the time. From what I see, even numpad keys (digits, enter, insert, etc) produce keysyms which are different from the other keys' keysyms.
>
>In the rare case when the same keysym is producible from different keys, we could just skip the conversion. In other words, we will only convert keysyms which can be replaced with a single keycode. This should still fix bindings with Latin characters, since it is hard to imagine a keyboard with several keys to type in the same letter in one layout, and keysyms which cannot be safely replaced won't be affected.

Let's say keycodes Shift+1 gives the keysym 1 in one of my layouts (e.g. fr), and that just pressing the keycode 1 gives the keysym 1 in the other (e.g. en). What happens with this patch applied?

kupospelov:
> Let's say keycodes Shift+1 gives the keysym 1 in one of my layouts (e.g. fr), and that just pressing the keycode 1 gives the keysym 1 in the other (e.g. en). What happens with this patch applied?

It depends on how you have configured your layouts. The layout which comes first is used to resolve the bindsyms.

For `XKB_DEFAULT_LAYOUT='us,fr'`:
1. $mod+1 will switch to workspace 1 in both layouts.
2. $mod+w will work in both layouts as if you used `us`.

For `XKB_DEFAULT_LAYOUT='fr,us'`:
1. 1 cannot be resolved into a bindcode, so it will be layout-dependent: you will have to use $mod+1 in `us` and $mod+Shift+1 in `fr`.
2. $mod+w will work in both layouts as if you used 'fr'.

chebykinn:
Any updates on this? This is currently one of the biggest pain points with sway for me.

kupospelov:
> I'm really not a fan of converting keysyms into keycodes. This makes bindings depend on the first layout, and breaks other layouts (set via xkb_layout a,b,c or when changed via the IPC).

The issue with keysyms is that they depend on the _current_ layout and this is a pain. The simplest way to fix that is to use only one layout for bindings, which sounds like a reasonable requirement.

As for IPC, there are options:
* Do not allow changing the layout configuration using the IPC
* Re-translate the keysyms every time there is a change in the layout configuration

> Investigating what i3 does would be interesting.

This approach is actually inspired by what i3 does. You can take a look at `i3/bindings.c`, function `translate_keysyms`. You can also look into function `get_binding` to make sure that there are no direct keysym comparisons. The behavior is also similar: the first configured layout is used to resolve keysyms.

mstoeckl:
i3 has another feature: keybindings can be restricted to a single layout with the `Group1`, `Group2`, `Group3`, `Group4` modifiers. See https://github.com/i3/i3/pull/1857 and the discussion in https://github.com/i3/i3/issues/1835.

kupospelov:
@emersion, could you please make your point clearer:

1. Do you generally agree that layout-dependent shortcuts is a problem that should be addressed in one way or another?
2. If you do agree, but do not like the keycode translation, do you know any better ways to achieve that?
3. If, for example, the keysym translation was performed on every change of layout configuration via the IPC and it worked with the `xkb_layout` command, how would this approach look?

ojab:
>Excuse me, but an you please re-explain why it's an issue with Russian layouts?

Because pressing `$mod+x` , for instance, with `ru` layout (the physical key, that is actually a `ч` keysym) doesn't work like pressing `$mod+x` in `en` layout.
It's an issue since shortcuts are generally working in any layout (in Qt & GTK+, for example, `Alt+q` [`Alt+й`] in `ru` closes the program).
Right now `$mod-[a-z]` is not working at all in `ru` layout and that's a PITA: switching to the container on the left while using `ru` layout leads to `$mod+д`, `$mod+д`, `$mod+д`, why the fuck it doesn't work, oh, `change layout`, `$mod+l`, `change layout`, there it is.

RedBorg:
This could work like variables, you add an option to `bindcode` and then specify the layout to use it on, and when the file is processed, you convert it to a normal bindcode.

Example of implementation:

`bindcode --sym [layout]:[variant] [keybind] [command]`

`bindcode --sym us:us $mod+e reload`

YaLTeR:
> Excuse me, but an you please re-explain why it's an issue with Russian layouts?

To better demonstrate the issue, here's an example Russian keyboard:

image

It's expected to be used with the US layout interchangeably (so you toggle between the two). When you switch the Russian layout on, keys don't really move around (it's not like 1 becomes shift+something or anything like that) so it's expected that US bindings continue to work the same regardless of the currently active layout. For the keys that do move (some punctuation) I would say it's still expected that they work as in the US layout rather than as in the Russian layout.

kupospelov:
> Excuse me, but an you please re-explain why it's an issue with Russian layouts?

Folks in this thread were quite eloquent at describing issues, it is hard to add anything else.

Generally speaking, the main issue is that there is no `[a-z]` keysyms in the Russian layout. When layout changes, every such key changes its keysym to something else. The `[a-z]` and `[а-я]` keysym sets do not overlap (e.g. keysym for `а` is `Cyrilic_a`, not `a`). Because of that most of the default keysym bindings just stop working until the layout is switched back to `us`. 

I am also pretty sure that the same issue exists for a number of other layouts (Ukrainian and Greek for example, should be more).

> I should really read i3's code to understand in detail what they do.

Please take a look when you have a chance. The most interesting file there is `src/binding.c`. 

From what I see the meaning of `bindsym` is different in i3. It basically says "bind the command to all the keys the keysym can be produced from in layout N". By default N = 1, but as @mstoeckl mentioned, it is also possible to specify which layout to use with `Group[1-4]`.

> The thing that makes me skeptical about this feature is that as a user, I really don't expect $mod+1 to be bound to a completely different keysym combination when switching layouts. If I bind a keysym, then I don't expect to trigger it when pressing keys that don't match keysyms on another layout.

This sounds reasonable, but there is another expectation: any binding should consistently work. As a result, in case of two or more layouts, only keys producing the same keysyms in all of the layouts are actually usable, and the others are the source of pain and despair.

> It would be nice when implementing this to try to keep the code as simple and straightforward as possible (disclaimer: I haven't looked at the code yet).

Please consider the current implementation a prototype. It has worked for me since October, but some pieces of code will need to be changed or added to support dynamic re-translation.

It seems that the key decision that needs to be made is whether sway must be consistent with i3 in how bindings work.

In terms of implementation, I tend to believe that approach used in i3 ("translate every keysym into keycodes and then only use keycodes") is simplest possible, just because there would basically be only one type of bindings instead of two and the code to process keysyms might be dropped.

Freyr666:
This works perfectly in i3 and doesn't work with sway, causing a lot of pain when using non-latin keymap. Does 1.0 release imply feature parity with i3?

ddevault:
Tone down the entitlement. This project is run by volunteers who are just doing their best in their spare time.

I suppose a verdict of some kind is necessary here. Unless @emersion disagrees, I think that this behavior (translating keysyms to keycodes) should only be done with a flag, e.g. `bindsym --to-code`. Since sway supports more general subcommands, like this:

```
bindsym --to-code {
    Mod1+f foobar
    Mod1+b foobaz
}
```

This shouldn't be too much of a pain to configure.

Hi-Angel:
> I think that this behavior (translating keysyms to keycodes) should only be done with a flag

Why not default? What would be downsides? FWIW, this is how keybindings work in pretty much any software. The example with <kbd>Ctrl</kbd>+<kbd>q</kbd> someone put above is a bright example. Not having it as default, I think, a unique "feature" to sway, any reason why?

ddevault:
>Why not default? What would be downsides?

This was already explained earlier in the thread.

emersion:
>Why not default? What would be downsides?

1. Setup `xkb_layout en,fr`
2. Bind `q`
3. Switch to `fr`
4. The `a` key on the keyboard triggers `q`

Hi-Angel:
Thanks @emersion! Please, could you run any Qt application, and tell whether the behavior is the same when Ctrl+q is pressed *(by default it should close the application)*. It would help us to figure out whether they do something different or not.

emersion:
I'm using Firefox right now. In the en layout I need to press Ctrl+Q to quit, but when I switch to the fr layout I need to press Ctrl+A (the physical A key is bound to the Q key in the layout).

ddevault:
Which seems correct to me... no?

YaLTeR:
Seems to be correct and also the same as what people are requesting in this PR.

emersion:
>Seems to be correct

Yes.

>and also the same as what people are requesting in this PR.

No, this PR converts keysyms to keybinds with the first layout. In other words, it always uses the same physical keys, regardless of the current active layout.

The behaviour with this PR would be: in en layout, press Ctrl+Q (physical keys on my QWERTY keyboard), and in fr layout, exactly the same.

YaLTeR:
Ah, I misunderstood your reply.

Hi-Angel:
Thanks! So, to have some context, here's French layout, it has <kbd>A</kbd> at the same place as <kbd>Q</kbd> in English.
screenshot_20190212_185547

As I understand, both @emersion and @ddevault agree that pressing <kbd>Ctrl</kbd>+<kbd>A</kbd> in French should work the same as <kbd>Ctrl</kbd>+<kbd>Q</kbd> in English. Unfortunately, as emersion noted, it seems not to be the behavior that this PR request makes.

@kupospelov is that possible to have in pull request the behavior that @emersion described in his example with Firefox? It should solve the issue for everyone.

eternal-sorrow:
The question is, does it work like that in the current state of sway?

Hi-Angel:
> The question is, does it work like that in the current state of sway?

@emersion for you, does binding <kbd>mod</kbd>+<kbd>Q</kbd> to some action, currently, make the <kbd>mod</kbd>+<kbd>A</kbd> in French layout to execute the same action?

emersion:
>does binding mod+Q to some action, currently, make the mod+A in French layout to execute the same action?

Currently in Sway, we keep keysyms if the user chooses to use keysyms. Just tested to make sure, and yes, this is the case.

In the end, I feel like @ddevault's [proposal](https://github.com/swaywm/sway/pull/3058#issuecomment-462778072) is at the same time clear (no magic off-the-scenes conversion) and easy to setup. Do you think it's reasonable?

Hi-Angel:
> > does binding mod+Q to some action, currently, make the mod+A in French layout to execute the same action?
> 
> Currently in Sway, we keep keysyms if the user chooses to use keysyms. Just tested to make sure, and yes, this is the case.
>
> In the end, I feel like @ddevault's proposal is at the same time clear (no magic off-the-scenes conversion) and easy to setup. Do you think it's reasonable?

Could you please elaborate on `if the user chooses to use keysyms`. It sounds like what we discussing here should already work in sway, given this magic switch is flipped on; and there's no need for any pull request at all.

emersion:
>Could you please elaborate on `if the user chooses to use keysyms`. It sounds like what we discussing here should already work in sway, given this magic switch is flipped on; and there's no need for any pull request at all.

I believe we still need a PR, but let me explain how it works currently, just to be sure we're on the same page.

There are two ways one can refer to a particular key: keycodes and keysyms (aka. key symbols). Keycodes refer to physical keys on the keyboard, they don't depend on the active layout. For instance, a keycode could be "the second key on the third row of the keyboard". Keysyms refer to symbols, and they do depend on the currently active layout. For instance, a keysym could be "a" or "é". The physical key you press to obtain a given keysym depends on the layout.

In Sway, you have two ways to setup a binding: `bindsym` and `bindcode`. If you use `bindcode`, you'll need to press the same physical key regardless of the layout you use. If you use `bindsym`, you'll need to press the physical key producing the right keysym, and this depends on your layout.

Hi-Angel:
@emersion thank you! So, I guess the `bindcode` was exactly what people at #2999 wanted. @kupospelov  does `bindcode` solve the issue?

eternal-sorrow:
> does bindcode solve the issue?

I'm not @kupospelov, but I can answer. It does'n because you need to determine keycodes for all of your keybinds. That's possible, but not very convenient.

emersion:
>does `bindcode` solve the issue?

The issue with `bindcode` is that it takes keycodes, which are just numbers, so they're pretty hard to use and unreadable. We can't make `bindcode` support keysyms because there are conflicts between keycodes and keysyms (e.g. "1" is a valid keycode *and* a valid keysym). That's why we need something like `bindsym --to-code`.

Hi-Angel:
@emersion ah, I see. Then, why does currently binding <kbd>mod</kbd>+<kbd>Q</kbd> works for you in French layout as <kbd>mod</kbd>+<kbd>A</kbd>? They're different keysims, have you maybe occasionally bound both of them separately?

eternal-sorrow:
> That's why we need something like bindsym --to-code.
We need this behavior default. I believe, it won't cause any issues with French layout.

emersion:
>why does currently binding mod+Q works for you in French layout as mod+A? They're different keysims, have you maybe occasionally bound both of them separately?

Binding the mod+Q keysyms will require me to press the **physical** mod+A keys, because I have a QWERTY keyboard. I should've made this more explicit.

>We need this behavior default.

https://github.com/swaywm/sway/pull/3058#issuecomment-462781843

Hi-Angel:
That said, I starting to see the issue. French has `Q` keysym at a different place compared to English. When you do `bindsym Q action`, would it have translated keysym to keycode, it's unclear if user wants the `Q` from English or from French. What @ddevault suggests though wouldn't solve it either.

I suggest the following solution: how about after making a global switch, which simply would choose which layout to use for keysym → keycode translation? This would be used along with `bindkey` translating to keysym by default, since it's what folks in GUI tookits and i3 are using, and it works well.

Hi-Angel:
@emersion is what I said above correct?

Hi-Angel:
@emersion I edited the comment, had some typos

Hi-Angel:
@emersion sorry if I said something wrong, just point it out. This whole situation seems confusing, so I'm trying to understand how it looks from your point of view, to have context shared :)

eternal-sorrow:
> global switch, which simply would choose which layout to use for keysym → keycode translation

Why don't make it like in i3 - by default use the first layout and bind in all layouts, but with Group<1-4> you can bind only in one of layouts.

ddevault:
This discussion has gone on too long and in too many directions with too many people. A pull request which makes [my proposal](https://github.com/swaywm/sway/pull/3058#issuecomment-462778072) a reality will be merged, one that does not will be closed.

jeremejevs:
Bikeshedding, but `bindcode --from-sym` might be slightly more intuitive than `bindsym --to-code`, since, ultimately, a code gets bound, not a sym.

ddevault:
>Bikeshedding, but bindcode --from-sym might be slightly more intuitive than bindsym --to-code, since, ultimately, a code gets bound, not a sym.

Good call.

ddevault:
Also needs a man page update

kupospelov:
Here is the updated changeset.

Summary:
1. `bindsym --to-code` enables keysym to keycode translation.
2. If there are no `xkb_layout` commands in the config file, the translation uses the XKB_DEFAULT_LAYOUT value.
3. It there is one or more `xkb_layout` command, the translation uses the first one.
4. If the translation is unsuccessful, a message is logged and the binding is stored as BINDING_KEYSYM.
5. The binding keysyms are stored and re-translated when a change in the input configuration may affect the translated bindings.

Limitations:
1. Keysyms translated into several keycodes are not supported for now. This will complicate input processing, I am not sure this is worth it, at least right now.

@emersion, @ddevault, please take a look when you have a chance.

ddevault:
Summary sounds good. Took a very brief glance over the code and it looked about right. Will review in more detail later.

emersion:
Can you rebase against master?

sway/config.c
```diff
@@ -300,6 +326,9 @@ static void config_defaults(struct sway_config *config) {
 	if (!(config->feature_policies = create_list())) goto cleanup;
 	if (!(config->ipc_policies = create_list())) goto cleanup;
 
+	// The keysym to keycode translation
+	config->keysym_translation = new_keysym_translation_data(getenv("XKB_DEFAULT_LAYOUT"));
```
emersion: We generally call functions that create things with the `_create` suffix, and those which destroy things with the `_destroy` suffix.

emersion: The `XKB_DEFAULT_LAYOUT` env variable is deprecated. Also, what about xkb options?

Can't we just reference use the first xkb layout we create from the config instead? Why create another one?

sway/commands/input.c
```diff
@@ -39,6 +39,30 @@ static struct cmd_handler input_config_handlers[] = {
...
+ * Re-translate keysyms if a change in the input config could affect them.
+ */
+static void retranslate_keysyms(struct input_config *input_config) {
+	bool matched = false;
+	for (int i = 0; i < config->input_configs->length; ++i) {
+		struct input_config *ic = config->input_configs->items[i];
+		matched |= ic->identifier == input_config->identifier;
+
+		// the first configured xkb_layout
+		if (ic->xkb_layout) {
```
emersion: Why not use `input_config->xkb_layout` directly?

jeremejevs:
@kupospelov What about `bindcode --from-sym`?

sway/config.c
kupospelov: Yep, I will need to add xkb options.

As for the reference, ideally we need to use `xkb_keymap` from the corresponding `sway_keyboard` instance. The issue is that input devices are created after the main config is loaded. To workaround this, we might need to add some code to update config and retranslate keysyms when the keyboard is configured somewhere in `seat_add_device`, but in my opinion a separate instance of `xkb_keymap` is not so expensive to justify the complications.

sway/commands/input.c
kupospelov: For some reason, I supposed that `xkb_layout` could be removed from an input configuration. This does not seem to be true, so now keysyms are re-translated if the changed input is the first one with a configured `xkb_layout`.

kupospelov:
> @kupospelov What about `bindcode --from-sym`?

I think this should depend on the behavior of the flag. In the current implementation, `bindsym --to-code` works as an ordinary `bindsym` if the translation is not possible. This would hardly be expected when using `bindcode --from-sym`.

On the other hand, if the binding completely failed in that case, `bindcode --from-sym` would look fine.

jeremejevs:
Hmm. I personally would prefer it to fail, rather than fall back to `bindsym`, for the same reasons one would prefer not to use implicit typecasting (i.e. less magic). Is there some edge case I'm overlooking?

kupospelov:
> Hmm. I personally would prefer it to fail, rather than fall back to `bindsym`, for the same reasons one would prefer not to use implicit typecasting (i.e. less magic). Is there some edge case I'm overlooking?

I think a certain decision needs to be made regarding the command behavior in case the translation fails. I am not sure how realistic these cases are, but I have to describe them at least.

The translation is triggered and may generally fail in two commands:
1. `bindcode --from-sym` (a new binding cannot be translated)
2. `input` (some stored bindings cannot be re-translated using the new input configuration)

The current implementation has the following features:
1. If the keysym translation fails, the binding is still stored and can potentially be successfully re-translated when the input configuration changes. As a side effect, the order of `input` and `bindsym --to-code` commands does not matter, the result should be the same.
2. It may not be expected that the `input` command will remove stored bindings if they cannot be re-transated. Since we use `bindsym` as a fallback it will not.

In case of `bindcode --from-sym`, we have some options:
1. `bindcode --from-sym`:
    1. Do not store failed bindings: the order of `input` and `bindcode` commands becomes important.
    2. Introduce a separate list to store failed bindings and try to re-translate them when the input configuration changes: it may look strange to say the least that a command reported to be failed just started working after an input configuration change.
2. `input`:
    1. Remove stored bindings if they cannot be re-translated (and put them in the list of failed bindings if we decide to store them).
    2. Leave such bindings as is and just log this fact: this may lead to some inconsistency when older bindings are converted using a previously defined input configuration.

I like the fact that an explicit error message will be shown, but I think that the only viable solution with `bindcode --from-sym` is to implement 1.i and 2.ii with their pros and cons.

@emersion, what do you think?

emersion:
As much as I prefer things to hard-fail rather than to sweep things under the carpet and hope it still somehow works, `bindcode --from-sym` seems indeed to have issues when the `input` config is updated.

If we want `bindcode --from-sym`, I'd rather (1) log errors (2) cleanup successfully translated bindings and re-translate everything with the new layout. The latter conflicts with "It may not be expected that the input command will remove stored bindings if they cannot be re-transated" but IMHO it makes more sense.

All in all, not sure `bindcode --from-sym` is worth it. Thoughts?

jeremejevs:
> Thoughts?

Given that this appears to be less of a no-brainer than it at first seemed, it's probably better to just merge this PR as-is, and then consider adding `bindcode --from-sym` later, based on user feedback; the two options aren't mutually exclusive after all.

kupospelov:
> All in all, not sure bindcode --from-sym is worth it. Thoughts?

I agree. In addition to that, `bindsym --to-code` does not imply any implementation details and gives us more freedom: we can always change the fallback behavior (e.g. make it fail) and the users won't be forced to update their config files.

It looks like I have addressed all the review points, so this PR should be ready for a final review or can be merged.

sway/input/keyboard.c
```diff
@@ -470,39 +470,31 @@ struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
...
 	struct wlr_input_device *wlr_device =
 		keyboard->seat_device->input_device->wlr_device;
 
-	if (input_config && input_config->xkb_layout) {
-		rules.layout = input_config->xkb_layout;
+	struct xkb_rule_names rules;
+	if (input_config) {
+		rules = input_config_get_rule_names(input_config);
 	} else {
+		memset(&rules, 0, sizeof(rules));
```
emersion: Instead of this, you can use struct initializers:

```c
struct xkb_rule_names rules = {0};
```

sway/config/input.c
```diff
@@ -179,6 +179,18 @@ struct input_config *store_input_config(struct input_config *ic) {
 	return ic;
 }
 
+struct xkb_rule_names input_config_get_rule_names(struct input_config *ic) {
```
emersion: Can this fill a pointer to a `struct xkb_rule_names` instead of returning it?

sway/config.c
```diff
@@ -300,6 +319,11 @@ static void config_defaults(struct sway_config *config) {
 	if (!(config->feature_policies = create_list())) goto cleanup;
 	if (!(config->ipc_policies = create_list())) goto cleanup;
 
+	// The keysym to keycode translation
+	struct xkb_rule_names rules = {};
```
emersion: `= {}` creates an uninitialized value. Use `= {0}` instead.

sway/commands/input.c
emersion: Bump

kupospelov: It is possible to have several configured inputs, each with a different `xkb_layout`. We need to check if `input_config` is the first one in the list, otherwise, it should not affect the translated bindings.

emersion:
I'll try having a look again soon.

Can some other contributor look at the changes too?

mstoeckl:
Could you please rebase again? (There is a conflict is with the now-merged #3998, which introduced a `binding_add` function that plays the same role as `add_binding`. It is reasonably straightforward to fix; see for example https://github.com/mstoeckl/sway/commit/b8a6a0b96301149ad1b5e8fd2ef5b8f9fc80a489 , the result of my own attempt at resolving the rebase conflicts.)

Also, I have read through the changeset a few times and have no immediate complaints; testing layout switching with `xkb_options grp:alt_shift_toggle` did not reveal any bugs.

sway/config.c
```diff
@@ -921,3 +945,55 @@ void config_update_font_height(bool recalculate) {
 		arrange_root();
 	}
 }
+
+static void translate_bindings(list_t *bindings, list_t *bindsyms,
```
emersion: This is a little confusing, the only difference between `translate_bindings` and `translate_binding` is the final "s". Can we rename this one to `translate_bindings_list`?

kupospelov:
Rebased. `binding_add` turned out to require more logging context than the translation code could provide, so I created `binding_add_translated` and shared the code between the two in the `binding_upsert` function.

sway/config.c
kupospelov: Renamed to `translate_binding_list`

sway/commands/bind.c
```diff
@@ -360,3 +375,102 @@ void seat_execute_command(struct sway_seat *seat, struct sway_binding *binding)
...
+}
+
+bool translate_binding(struct sway_binding *binding) {
+	if ((binding->flags & BINDING_CODE) == 0) {
+		return true;
+	}
+
+	switch (binding->type) {
+		// a bindsym to translate
+		case BINDING_KEYSYM:
```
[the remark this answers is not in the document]
kupospelov: Fixed

chebykinn:
@emersion Can this be merged before 1.1 release?

emersion:
Oh, thanks for the heads-up. That would be nice indeed. Will have a look tomorrow.

emersion:
Thanks!

eternal-sorrow:
https://gist.github.com/eternal-sorrow/396d642a7c6be712157cc24376298faf

simple patch so that this behaviour was default (without --to-code)

jeremejevs:
DRYing things up without patching:

```ini
bindsym --to-code {
  $super+Return exec termite
  $super+Shift+q kill
  # ...
}
```

kindaro:
@kupospelov Thank you, great work. Also helps when moving to optimized layouts, such as Dvorak.

PR #3066, opened 2018-11-04
Improvements to the runtime only part of the gaps command

mwenzkowski:
This patch addresses two issues:
1. The command `gaps inner|outer all|current set|plus|minus <amount>` is not valid in the configuration file, hence list it accordingly in the manpage.
 2.  Always raise an error if the mentioned gaps command is found in the config file. Previously there was only an error reported on startup but not on config reload.

By working on this patch I realized that the gaps command is an exception to the other commands: All commands are classified in config file only commands, runtime only commands or commands which are valid in both cases. But the gaps command doesn't conform to this classification and actually consists  of two commands:
- `gaps inner|outer all|current set|plus|minus <amount>` (runtime only)
- `gaps inner|outer <amount>` (runtime and config)

Therefore the gaps command itself has to check if it is run at config load or at runtime. 

Instead of this patch it might be a cleaner solution to split the gaps command into two distinct commands
- `gaps inner|outer all|current set|plus|minus <amount>` (runtime only)
- `default_gaps inner|outer <amount>` (runtime and config)

with the disadvantage, that this syntax would be incompatible to i3-gaps. If we keep the current syntax, I think my patch is necessary.

### DIFF
--- a/sway/commands/gaps.c
+++ b/sway/commands/gaps.c
@@ -149,16 +149,17 @@ struct cmd_results *cmd_gaps(int argc, char **argv) {
 		return error;
 	}
 
+	bool config_loading = !config->active || config->reloading;
+
 	if (argc == 2) {
 		return gaps_set_defaults(argc, argv);
 	}
-	if (argc == 4) {
-		if (config->active) {
-			return gaps_set_runtime(argc, argv);
-		} else {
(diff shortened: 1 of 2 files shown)


mwenzkowski:
Please do not merge, I discovered a problem with my patch.

emersion:
Thanks!

PR #3067, opened 2018-11-04
Improve documentation and remove dead code

madblobfish:
This contains three unrelated things:
* Documentation for `swaymsg -t get_seats` (as well as implementation in fish completions)
* Fix in sway.5 manpage (fixes spacing after the Table)
* Deletion of a not needed code fragment in seat.c


emersion:
Thanks!

PR #3068, opened 2018-11-04
Use wlr_surface_get_effective_damage

emersion:
Required after https://github.com/swaywm/wlroots/pull/1261


emersion:
Bump, damage tracking is broken without this fix.

PR #3071, opened 2018-11-05
Fix focus_wrapping yes

### LINKED ISSUE #3064: "focus left/right" does not wrap around anymore 1.0-beta
when having tabbed windows and moving focus to the left or the right, it used to go back to the other side when one side was reached

I don't know if it's intentional or if that's a bug but it would be nice to have a way to do this again

ps: no other bugs found after using sway 1.0 for a few other but lots of things working better :grinning:

RedSoxFan:
Fixes #3064 

It appears that the focus code that handles `focus_wrapping yes` was removed during the conversion to type safety. This re-implements the focus code for when `focus_wrapping` is set to `yes` (default). Neither the `no` or `force` options appear to be effected and should be working.

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -156,6 +156,14 @@ static struct sway_node *node_get_in_direction(struct sway_container *container,
 	if (new_output) {
 		return get_node_in_output_direction(new_output, dir);
 	}
+
+	// If there is a wrap candidate, return its focus inactive view
+	if (wrap_candidate) {
+		struct sway_container *wrap_inactive = seat_get_focus_inactive_view(
+				seat, &wrap_candidate->node);
+		return &wrap_inactive->node;
+	}
+
 	return NULL;
 }
 


emersion:
>Neither the no or force options appear to be effected and should be working.

Hmm, where is the `focus_wrapping` preference checked?

sway/commands/focus.c
```diff
@@ -156,6 +156,14 @@ static struct sway_node *node_get_in_direction(struct sway_container *container,
 	if (new_output) {
 		return get_node_in_output_direction(new_output, dir);
 	}
+
+	// If there is a wrap candidate, return it's focus inactive view
```
Alphare:
```suggestion
	// If there is a wrap candidate, return its focus inactive view
```
Minor typo

RedSoxFan:
> Hmm, where is the focus_wrapping preference checked?

It is checked at https://github.com/swaywm/sway/blob/master/sway/commands/focus.c#L129 and https://github.com/swaywm/sway/blob/master/sway/commands/focus.c#L136

If `focus_wrapping` is set to `no`, `wrap_candidate` will never be set to anything other than `NULL`. If `focus_wrapping` is set to `force`, the focus_inactive view's node for `wrap_candidate` is immediately returned. At the point of the change in this PR, `wrap_candidate` will only be non-`NULL` when `focus_wrapping` is set to `yes` and there is a wrap possible.

sway/commands/focus.c
RedSoxFan: Thanks. Instead of applying the suggestion and adding a commit for the typo, I just fixed it locally before rebasing.

emersion:
Thanks!

PR #3073, opened 2018-11-05
Work around swayidle never triggering on SIGUSR1

oranenj:
SIGUSR1 is supposed to trigger swayidle; however, in my experience this never happens.

I suspect there might be some kind of a race condition between setting a 0-timeout timer and the actual signal handling, so here's a PR that kind of papers over the problem by making the timeout 50ms instead. This makes it work for me consistently

Feel free to reject this PR if this workaround is not acceptable, but I'm not sure how to actually fix the timer logic so that 0-timeout timers work.

### DIFF
--- a/swayidle/main.c
+++ b/swayidle/main.c
@@ -347,7 +347,9 @@ void sway_terminate(int exit_code) {
 
 static void register_zero_idle_timeout(void *item) {
 	struct swayidle_timeout_cmd *cmd = item;
-	register_timeout(cmd, 0);
+	// A zero timeout never actually triggers. Adding a 50ms timeout is most
+	// likely not the correct fix either, but will work
+	register_timeout(cmd, 50);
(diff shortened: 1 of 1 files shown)


emersion:
Works for me. Can you post a `WAYLAND_DEBUG` log of swayidle when it doesn't work?

SirCmpwn:
Wait why was this merged...

SirCmpwn:
Issue with dispatch.sr.ht, since resolved.

emersion:
@oranenj: can you re-submit your pull request? Sorry for the fuss.

SirCmpwn:
In any case, I'm not interested in this change as-is. If there's a race condition it should be fixed, not worked around.

oranenj:
@emersion I'll create an issue about this and put the debug output there

PR #3074, opened 2018-11-05
Add swaylock grace timer

oranenj:
This adds a small feature to allow for a moment of grace before swaylock actually activates. This allows the user to hit a key to cancel the lock, if they're quick. 

Most useful when swaylock is used from swayidle, as sometimes the lock screen can trigger when you don't want it to.

Also includes a tiny typo fix.


SirCmpwn:
No thanks. Try configuring a script with swayidle instead of calling swaylock directly.

PR #3078, opened 2018-11-05
Fix the double execution of exec_always commands on config reloads

Emantor:
While at it, fix the reload message.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -389,7 +389,8 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 	config_defaults(config);
 	config->validating = validating;
 	if (is_active) {
-		wlr_log(WLR_DEBUG, "Performing configuration file reload");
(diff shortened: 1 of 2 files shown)


sway/config.c
```diff
@@ -389,7 +389,7 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 	config_defaults(config);
 	config->validating = validating;
 	if (is_active) {
-		wlr_log(WLR_DEBUG, "Performing configuration file reload");
+		wlr_log(WLR_DEBUG, "Performing configuration file reload or validation");
```
emersion: Can you use `validating` to display a more precise debug message?

Emantor: Done.

Emantor:
Now the message should be correct :+1:

emersion:
Thanks!

Issue #3025, opened 2018-10-29
Window focus does not update when switching between workspaces.


c-edw:
When switching between workspaces, the window that the cursor is hovering over is not always focused. See below for example.

![out](https://user-images.githubusercontent.com/38229097/47686419-f7189a80-dbd2-11e8-878b-612088cc4d9a.png)

Version:
```
sway version 1.0-beta.1-108-gb90af335 (Oct 29 2018, branch 'master')
```

Steps to produce:
1. Start Sway
2. Move to workspace 1
3. Open two terminals 
4. Focus the left terminal by hovering over it with your cursor
5. Move to workspace 2
5. Open two terminals
6. Focus the right terminal by hovering over it with your cursor. Keep your cursor in that position.
7. Move back to workspace 1
8. Observe that the left terminal remains focused in workspace 1 despite the cursor hovering over the right terminal.

RyanDwyer:
This is intended. See #2815.

SirCmpwn:
This is deliberate, and yes, it comes from i3. I prefer this workflow.

c-edw:
If it's been brought it from compatibility then I can't really argue, but this has an impact on a mouse-based workflow. I've found myself having to flick my cursor over swaybar several times to refocus a window, because I don't use $mod+[arrow keys]. I would like to propose that the window will be focused upon *moving* the cursor, which should preserve both keyboard and mouse based workflows.

SirCmpwn:
Sway isn't designed for a mouse-based workflow :wink: 

I would accept a patch which extends `focus_follows_mouse` with such an option, though.

c-edw:
Something like `focus_follows_mouse always`? Could I have some guidance on where to look to implement this please?

SirCmpwn:
Yeah, something like that. Just grep for focus_follows_mouse to find the places where it's already in use.

PR #3081, opened 2018-11-06
Add focus_follows_mouse always.

c-edw:
I suggested a feature in #3025 where the focus will always follow the mouse. This can be used by specifying `focus_follows_mouse always` in the config file.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -637,7 +637,8 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor,
 	cursor->previous.y = cursor->cursor->y;
 	cursor->previous.node = node;
 
-	if (node && config->focus_follows_mouse) {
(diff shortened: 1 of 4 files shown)


sway/input/cursor.c
```diff
@@ -637,7 +637,8 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor,
 	cursor->previous.y = cursor->cursor->y;
 	cursor->previous.node = node;
 
-	if (node && config->focus_follows_mouse) {
+	if (node && (config->focus_follows_mouse == FOLLOWS_YES ||
+		config->focus_follows_mouse == FOLLOWS_ALWAYS)) {
```
emersion: Style: please add one more indentation level to this line, so we can easily tell it's the condition and not the body.

emersion:
Thanks!

Issue #3082, opened 2018-11-06
Features from i3 4.16


orestisf1993:
Like in https://github.com/swaywm/sway/issues/2336#issuecomment-435686783:

- [x] `strip_workspace_name yes|no` - https://i3wm.org/docs/userguide.html#_strip_workspace_numbers_name
- [x] `title_align left|center|right` - https://i3wm.org/docs/userguide.html#_window_title_alignment
- [x] i3-msg: add support for the SUBSCRIBE message type - https://github.com/i3/i3/pull/3435
- [x] `border toggle` now accepts an optional pixel argument - https://i3wm.org/docs/userguide.html#_changing_border_style
- [x] `resize set` now accepts the `width` and `height` keywords - https://i3wm.org/docs/userguide.html#resizingconfig
- [x] allow multiple assignments of workspaces to output - https://i3wm.org/docs/userguide.html#workspace_screen
- [x]  i3-nagbar: add option for button that runs commands without a terminal - https://github.com/i3/i3/pull/3258
- [x] ipc: introduce the sync IPC command for synchronization with i3bar - https://github.com/i3/i3/pull/3214
  - Note: It has been decided sway will not support this fully and just returns `{'success': false}` on `IPC_SYNC`

orestisf1993:
> Am I correct in thinking that having strip_workspace_numbers and strip_workspace_name enabled is invalid?

i3 will only respect one of them if they are both specified.

ianyfan:
There are some more additions outlined in [the release notes](https://github.com/i3/i3/blob/next/RELEASE-NOTES-4.16), e.g. sync command, so don't close this until they've all been sorted.

SirCmpwn:
>There are some more additions outlined in the release notes, e.g. sync command, so don't close this until they've all been sorted.

Can you enumerate these and update the OP?

PR #3083, opened 2018-11-06
Implement strip_workspace_name.

c-edw:
#3082.
~~For some reason, the `strip_workspace_name` is deemed invalid in the config file. Still trying to find out why.~~

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -1,4 +1,4 @@
-#define _XOPEN_SOURCE 700
+#define _POSIX_C_SOURCE 200809L
 #include <assert.h>
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -23,7 +23,8 @@ int numlen(int n) {
 	if (n == 0) {
 		return 1;
 	}
-	return log10(n) + 1;
+	// Account for the '-' in negative numbers.
+	return log10(abs(n)) + (n > 0 ? 1 : 2);
 }
 
 static struct modifier_key {
--- a/sway/commands/bar/strip_workspace_name.c
+++ b/sway/commands/bar/strip_workspace_name.c
@@ -1,0 +1,32 @@
+#include <string.h>
+#include <strings.h>
+#include "sway/commands.h"
+#include "log.h"
+#include "util.h"
+
+struct cmd_results *bar_cmd_strip_workspace_name(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc,
+				"strip_workspace_name", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	if (!config->current_bar) {
+		return cmd_results_new(CMD_FAILURE,
+				"strip_workspace_name", "No bar defined.");
+	}
+
+	config->current_bar->strip_workspace_name =
+		parse_boolean(argv[0], config->current_bar->strip_workspace_name);
+
+	if (config->current_bar->strip_workspace_name) {
+		config->current_bar->strip_workspace_numbers = false;
+
+		wlr_log(WLR_DEBUG, "Stripping workspace name on bar: %s",
+				config->current_bar->id);
+	} else {
+		wlr_log(WLR_DEBUG, "Enabling workspace name on bar: %s",
+				config->current_bar->id);
+	}
+
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+}
--- a/sway/commands/bar/strip_workspace_numbers.c
+++ b/sway/commands/bar/strip_workspace_numbers.c
@@ -2,6 +2,7 @@
 #include <strings.h>
 #include "sway/commands.h"
 #include "log.h"
+#include "util.h"
 
 struct cmd_results *bar_cmd_strip_workspace_numbers(int argc, char **argv) {
 	struct cmd_results *error = NULL;
@@ -13,17 +14,19 @@ struct cmd_results *bar_cmd_strip_workspace_numbers(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE,
 				"strip_workspace_numbers", "No bar defined.");
 	}
-	if (strcasecmp("yes", argv[0]) == 0) {
-		config->current_bar->strip_workspace_numbers = true;
+	
+	config->current_bar->strip_workspace_numbers =
+		parse_boolean(argv[0], config->current_bar->strip_workspace_numbers);
+
+	if (config->current_bar->strip_workspace_numbers) {
+		config->current_bar->strip_workspace_name = false;
+
 		wlr_log(WLR_DEBUG, "Stripping workspace numbers on bar: %s",
 				config->current_bar->id);
-	} else if (strcasecmp("no", argv[0]) == 0) {
-		config->current_bar->strip_workspace_numbers = false;
+	} else {
 		wlr_log(WLR_DEBUG, "Enabling workspace numbers on bar: %s",
 				config->current_bar->id);
-	} else {
-		return cmd_results_new(CMD_INVALID,
-				"strip_workspace_numbers", "Invalid value %s", argv[0]);
 	}
+
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
--- a/sway/sway-bar.5.scd
+++ b/sway/sway-bar.5.scd
@@ -50,6 +50,10 @@ Sway allows configuring swaybar in the sway configuration file.
 *workspace\_buttons* yes|no
 	Enables or disables workspace buttons on the bar. Default is _yes_.
 
+*strip\_workspace\_name* yes|no
+	If set to _yes_, then workspace names will be omitted from the workspace
+	button and only the custom number will be shown. Default is _no_.
+
 *strip\_workspace\_numbers* yes|no
 	If set to _yes_, then workspace numbers will be omitted from the workspace
 	button and only the custom name will be shown. Default is _no_.
--- a/swaybar/ipc.c
+++ b/swaybar/ipc.c
@@ -153,9 +153,10 @@ static bool ipc_parse_config(
 		return false;
 	}
 	json_object *markup, *mode, *hidden_state, *position, *status_command;
-	json_object *font, *bar_height, *wrap_scroll, *workspace_buttons, *strip_workspace_numbers;
-	json_object *binding_mode_indicator, *verbose, *colors, *sep_symbol, *outputs;
-	json_object *bindings;
+	json_object *font, *bar_height, *wrap_scroll, *workspace_buttons;
+	json_object *strip_workspace_numbers, *strip_workspace_name;
+	json_object *binding_mode_indicator, *verbose, *colors, *sep_symbol;
+	json_object *outputs, *bindings;
 	json_object_object_get_ex(bar_config, "mode", &mode);
 	json_object_object_get_ex(bar_config, "hidden_state", &hidden_state);
 	json_object_object_get_ex(bar_config, "position", &position);
@@ -165,6 +166,7 @@ static bool ipc_parse_config(
 	json_object_object_get_ex(bar_config, "wrap_scroll", &wrap_scroll);
 	json_object_object_get_ex(bar_config, "workspace_buttons", &workspace_buttons);
 	json_object_object_get_ex(bar_config, "strip_workspace_numbers", &strip_workspace_numbers);
+	json_object_object_get_ex(bar_config, "strip_workspace_name", &strip_workspace_name);
 	json_object_object_get_ex(bar_config, "binding_mode_indicator", &binding_mode_indicator);
 	json_object_object_get_ex(bar_config, "verbose", &verbose);
 	json_object_object_get_ex(bar_config, "separator_symbol", &sep_symbol);
@@ -189,6 +191,9 @@ static bool ipc_parse_config(
 	}
 	if (strip_workspace_numbers) {
 		config->strip_workspace_numbers = json_object_get_boolean(strip_workspace_numbers);
+	}
+	if (strip_workspace_name) {
+		config->strip_workspace_name = json_object_get_boolean(strip_workspace_name);
 	}
 	if (binding_mode_indicator) {
 		config->binding_mode_indicator = json_object_get_boolean(binding_mode_indicator);
@@ -298,6 +303,24 @@ bool ipc_get_workspaces(struct swaybar *bar) {
 					calloc(1, sizeof(struct swaybar_workspace));
 				ws->num = json_object_get_int(num);
 				ws->name = strdup(json_object_get_string(name));
+				ws->label = strdup(ws->name);
+				// ws->num will be -1 if workspace name doesn't begin with int.
+				if (ws->num != -1) {
+					size_t len_offset = numlen(ws->num);
+					if (bar->config->strip_workspace_name) {
+						free(ws->label);
+						ws->label = malloc(len_offset + 1 * sizeof(char));
+						ws->label[len_offset] = '\0';
+						strncpy(ws->label, ws->name, len_offset);
+					} else if (bar->config->strip_workspace_numbers) {
+						len_offset += ws->label[len_offset] == ':';
+						if (strlen(ws->name) > len_offset) {
+							free(ws->label);
+							// Strip number prefix [1-?:] using len_offset.
+							ws->label = strdup(ws->name + len_offset);
+						}
+					}
+				}
 				ws->visible = json_object_get_boolean(visible);
 				ws->focused = json_object_get_boolean(focused);
 				if (ws->focused) {
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -342,19 +342,6 @@ static uint32_t render_binding_mode_indicator(cairo_t *cairo,
 	return output->height;
 }
 
-static const char *strip_workspace_number(const char *ws_name) {
-	size_t len = strlen(ws_name);
-	for (size_t i = 0; i < len; ++i) {
-		if (ws_name[i] < '0' || ws_name[i] > '9') {
-			if (':' == ws_name[i] && i < len - 1 && i > 0) {
-				return ws_name + i + 1;
-			}
-			return ws_name;
-		}
-	}
-	return ws_name;
-}
-
 static enum hotspot_event_handling workspace_hotspot_callback(struct swaybar_output *output,
 			int x, int y, enum x11_button button, void *data) {
 	if (button != LEFT) {
@@ -368,11 +355,6 @@ static uint32_t render_workspace_button(cairo_t *cairo,
 		struct swaybar_output *output,
 		struct swaybar_workspace *ws, double *x) {
 	struct swaybar_config *config = output->bar->config;
-	const char *name = ws->name;
-	if (config->strip_workspace_numbers) {
-		name = strip_workspace_number(ws->name);
-	}
-
 	struct box_colors box_colors;
 	if (ws->urgent) {
 		box_colors = config->colors.urgent_workspace;
@@ -388,7 +370,7 @@ static uint32_t render_workspace_button(cairo_t *cairo,
 
 	int text_width, text_height;
 	get_text_size(cairo, config->font, &text_width, &text_height, NULL,
-			output->scale, config->pango_markup, "%s", name);
+			output->scale, config->pango_markup, "%s", ws->label);
 
 	int ws_vertical_padding = WS_VERTICAL_PADDING * output->scale;
 	int ws_horizontal_padding = WS_HORIZONTAL_PADDING * output->scale;
@@ -421,7 +403,7 @@ static uint32_t render_workspace_button(cairo_t *cairo,
 	cairo_set_source_u32(cairo, box_colors.text);
 	cairo_move_to(cairo, *x + width / 2 - text_width / 2, (int)floor(text_y));
 	pango_printf(cairo, config->font, output->scale, config->pango_markup,
-			"%s", name);
+			"%s", ws->label);
 
 	struct swaybar_hotspot *hotspot = calloc(1, sizeof(struct swaybar_hotspot));
 	hotspot->x = *x;
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -198,6 +198,7 @@ sway_cmd bar_cmd_separator_symbol;
 sway_cmd bar_cmd_status_command;
 sway_cmd bar_cmd_pango_markup;
 sway_cmd bar_cmd_strip_workspace_numbers;
+sway_cmd bar_cmd_strip_workspace_name;
 sway_cmd bar_cmd_swaybar_command;
 sway_cmd bar_cmd_tray_output;
 sway_cmd bar_cmd_tray_padding;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -213,6 +213,7 @@ struct bar_config {
 	bool wrap_scroll;
 	char *separator_symbol;
 	bool strip_workspace_numbers;
+	bool strip_workspace_name;
 	bool binding_mode_indicator;
 	bool verbose;
 	pid_t pid;
--- a/include/swaybar/bar.h
+++ b/include/swaybar/bar.h
@@ -68,6 +68,7 @@ struct swaybar_workspace {
 	struct wl_list link; // swaybar_output::workspaces
 	int num;
 	char *name;
+	char *label;
 	bool focused;
 	bool visible;
 	bool urgent;
--- a/include/swaybar/config.h
+++ b/include/swaybar/config.h
@@ -34,6 +34,7 @@ struct swaybar_config {
 	char *hidden_state;
 	char *modifier;
 	bool strip_workspace_numbers;
+	bool strip_workspace_name;
 	bool binding_mode_indicator;
 	bool wrap_scroll;
 	bool workspace_buttons;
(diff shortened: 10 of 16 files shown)


sway/commands/bar/strip_workspace_name.c
```diff
@@ -0,0 +1,29 @@
...
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc,
+				"strip_workspace_name", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	if (!config->current_bar) {
+		return cmd_results_new(CMD_FAILURE,
+				"strip_workspace_name", "No bar defined.");
+	}
+	if (strcasecmp("yes", argv[0]) == 0) {
```
c-edw: Copied this part from `strip_workspace_number.c`. Might be better to use `parse_boolean` here.

SirCmpwn: +1, suggest using it in both places

c-edw:
I'm almost certain that I have put a `strip_workspace_name` everywhere necessary, but it's still invalid when the config loads.

```
2018-11-06 20:18:29 - [sway/config.c:655] Read line 212:     strip_workspace_numbers yes
2018-11-06 20:18:29 - [sway/commands.c:374] handling config command 'bar strip_workspace_numbers yes'
2018-11-06 20:18:29 - [sway/commands.c:156] find_handler(strip_workspace_numbers)
2018-11-06 20:18:29 - [sway/commands.c:409] Subcommand: strip_workspace_numbers yes
2018-11-06 20:18:29 - [sway/commands.c:156] find_handler(strip_workspace_numbers)
2018-11-06 20:18:29 - [sway/commands/bar/strip_workspace_numbers.c:23] Stripping workspace numbers on bar: bar-0
2018-11-06 20:18:29 - [sway/config.c:655] Read line 213:     strip_workspace_name yes
2018-11-06 20:18:29 - [sway/commands.c:374] handling config command 'bar strip_workspace_name yes'
2018-11-06 20:18:29 - [sway/commands.c:156] find_handler(strip_workspace_name)
2018-11-06 20:18:29 - [sway/commands.c:409] Subcommand: strip_workspace_name yes
2018-11-06 20:18:29 - [sway/commands.c:156] find_handler(strip_workspace_name)
2018-11-06 20:18:29 - [sway/config.c:708] Error on line 213 'strip_workspace_name yes': Unknown/invalid command (/home/cedwards/.config/sway/config)
```

It fails to find the handler, even though I've defined it:
```
...
{ "strip_workspace_numbers", bar_cmd_strip_workspace_numbers },
{ "strip_workspace_name", bar_cmd_strip_workspace_name },
...
```

I've also tried an `rm -rf build` since I had to add a file to `meson.build`. No luck.

RedSoxFan:
> ...
> { "strip_workspace_numbers", bar_cmd_strip_workspace_numbers },
> { "strip_workspace_name", bar_cmd_strip_workspace_name },
> ...

The handlers need to be in alphabetical order. Just reverse the two lines and it should work.

c-edw:
Can't believe I missed that, thanks a lot.

swaybar/render.c
```diff
@@ -355,17 +355,8 @@ static const char *strip_workspace_number(const char *ws_name) {
...
-		if (ws_name[i] < '0' || ws_name[i] > '9') {
-			if (':' == ws_name[i] && i < len - 1 && i > 0) {
-				return &ws_name[0];
-			}
-			return ws_name;
-		}
-	}
-	return ws_name;
+static const char *strip_workspace_name(char *ws_name) {
+	return strtok(ws_name, ":");
```
c-edw: This is likely a poor way of doing it, but I am not familiar enough with C to know a better way...

sway/commands/bar/strip_workspace_name.c
```diff
@@ -0,0 +1,30 @@
...
+				"strip_workspace_name", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	if (!config->current_bar) {
+		return cmd_results_new(CMD_FAILURE,
+				"strip_workspace_name", "No bar defined.");
+	}
+
+	config->current_bar->strip_workspace_name = 
+		parse_boolean(argv[0], config->current_bar->strip_workspace_numbers);
```
RyanDwyer: s/numbers/name

sway/commands/bar/strip_workspace_name.c
```diff
@@ -0,0 +1,30 @@
...
+
+	if (config->current_bar->strip_workspace_name) {
+		wlr_log(WLR_DEBUG, "Stripping workspace name on bar: %s",
```
RyanDwyer: Should you force `strip_workspace_numbers` to `false` here?

sway/commands/bar/strip_workspace_numbers.c
```diff
@@ -13,17 +14,17 @@ struct cmd_results *bar_cmd_strip_workspace_numbers(int argc, char **argv) {
...
+	
+	config->current_bar->strip_workspace_numbers = 
+		parse_boolean(argv[0], config->current_bar->strip_workspace_numbers);
+
+	if (config->current_bar->strip_workspace_numbers) {
 		wlr_log(WLR_DEBUG, "Stripping workspace numbers on bar: %s",
 				config->current_bar->id);
-	} else if (strcasecmp("no", argv[0]) == 0) {
-		config->current_bar->strip_workspace_numbers = false;
+	} else {
```
RyanDwyer: Set `strip_workspace_name` to `false` here.

swaybar/ipc.c
```diff
@@ -153,7 +153,7 @@ static bool ipc_parse_config(
 		return false;
 	}
 	json_object *markup, *mode, *hidden_state, *position, *status_command;
-	json_object *font, *bar_height, *wrap_scroll, *workspace_buttons, *strip_workspace_numbers;
+	json_object *font, *bar_height, *wrap_scroll, *workspace_buttons, *strip_workspace_numbers, *strip_workspace_name;
```
RyanDwyer: This whole function appears to ignore the 80 character line limit, but that doesn't mean you should do the same. I'm not fussed as to whether you wrap the other lines or not.

swaybar/render.c
RyanDwyer: `strtok` allocates a new string, unlike `strip_workspace_number` which returns a pointer within the same string. As it stands there's a memory leak here.

I note this is being allocated while rendering, so I think a different approach needs to be taken. We could add a new property to the `swaybar_workspace` struct for the renderable name (ie. stripped), then update it when we receive workspace events. Then render using that property.

swaybar/render.c
```diff
@@ -371,6 +375,8 @@ static uint32_t render_workspace_button(cairo_t *cairo,
 	const char *name = ws->name;
 	if (config->strip_workspace_numbers) {
 		name = strip_workspace_number(ws->name);
+	} else if(config->strip_workspace_name) {
```
RyanDwyer: Style: Space between `if` and `(`.

swaybar/render.c
c-edw: Should I add a `ws_number` and `ws_name` field to `swaybar_workspace` then? I assume they'd both need to be `char *`.

swaybar/ipc.c
```diff
@@ -298,6 +304,19 @@ bool ipc_get_workspaces(struct swaybar *bar) {
 					calloc(1, sizeof(struct swaybar_workspace));
 				ws->num = json_object_get_int(num);
 				ws->name = strdup(json_object_get_string(name));
+				char *ws_dup = strdup(ws->name);
+				char *stripped_name = strtok(ws_dup, ":");
```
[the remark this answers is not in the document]
c-edw: There's definitely a few holes here, like the one you mentioned, and if you renamed a workspace to something like `1:a:b`, the `stripped_name` will just be `a` and not `a:b`. I could try to fix this, but I wouldn't know how to do the validity check. I'd probably need to cast `ws->num` to a char and compare it to the first character of `ws->name` right?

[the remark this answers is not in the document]
ianyfan: I don't think it needs to be that complicated. Just check if `ws->num == -1` to see if the string begins with a valid number.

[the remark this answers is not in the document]
ianyfan: Another solution I thought of is to not have `stripped_name` and instead convert `ws->num` to a string. Could also have `stripped_number` just be either an index or pointer of where the name begins in `ws->name` and it'll still work as a valid string, though I'm not sure if it is more or less hassle than duplicating the string and freeing it each time.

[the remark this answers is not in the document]
c-edw: I believe that `ws->num` is independent from `ws->name`. The `ws->num` workspace number is assigned automatically and cannot be changed, but the `stripped_number` is the 'frontend' number which can be changed.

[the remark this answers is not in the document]
ianyfan: I don't think so: https://github.com/swaywm/sway/blob/7fa7f4f48d17e0470c800b258061d188ceb705da/sway/ipc-json.c#L191

c-edw:
This commit fixes all the edge cases I'm aware of. I noticed some unrelated behaviour with custom workspace names - if you rename a workspace so that the `num` becomes 2 (`swaymsg rename workspace "1" to "2"`), using `$mod+2` won't switch to that workspace. Is this intended?

swaybar/ipc.c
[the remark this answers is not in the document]
c-edw: Thank you for your help resolving these issues :+1:

swaybar/ipc.c
```diff
@@ -298,6 +304,19 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
+				// ws->num will be 0 if workspace name doesn't begin with int.
+				if (stripped_name != NULL && 
+						strlen(ws->name) > 2 && ws->num != 0) {
```
ianyfan: Not 0, -1.
Also, using a length of 2 misses cases when num > 9.
I think you should do `strlen(ws->name) > strlen(stripped_name) + 1`. (or set some variable to `strlen(stripped_name) + 1` and use that as the offset later too)

c-edw: According to documentation, `atoi` returns 0 if invalid. Existing code already assumes that `num` is 0-9.

c-edw:
@RyanDwyer I think this commit should do it the way you suggested.

swaybar/ipc.c
ianyfan: https://github.com/swaywm/sway/blob/7fa7f4f48d17e0470c800b258061d188ceb705da/sway/ipc-json.c#L191 gives -1 for num if there isn't a number.

c-edw: Yeah you're right, thanks.

swaybar/ipc.c
```diff
@@ -298,6 +304,21 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
+				char *ws_dup = strdup(ws->name);
+				char *stripped_name = strtok(ws_dup, ":");
+				size_t num_len = strlen(stripped_name) + 1;
+				// ws->num will be 0 if workspace name doesn't begin with int.
+				if (stripped_name != NULL && 
+						strlen(ws->name) > num_len && ws->num != -1) {
+					if(bar->config->strip_workspace_name) {
+						ws->label = strdup(stripped_name);
+					} else if(bar->config->strip_workspace_numbers) {
+						// Strip [n:] using offset 2.
```
ianyfan: This comment needs to be changed.

swaybar/ipc.c
```diff
@@ -298,6 +304,21 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
 				ws->name = strdup(json_object_get_string(name));
+				ws->label = strdup(ws->name);
+				char *ws_dup = strdup(ws->name);
+				char *stripped_name = strtok(ws_dup, ":");
+				size_t num_len = strlen(stripped_name) + 1;
+				// ws->num will be 0 if workspace name doesn't begin with int.
+				if (stripped_name != NULL && 
+						strlen(ws->name) > num_len && ws->num != -1) {
+					if(bar->config->strip_workspace_name) {
+						ws->label = strdup(stripped_name);
```
RedSoxFan: You need to free `ws->label` before assigning a new value here to avoid a memory leak.

swaybar/ipc.c
```diff
@@ -298,6 +304,21 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
+					} else if(bar->config->strip_workspace_numbers) {
+						// Strip number prefix [1-?:] using offset num_len.
+						ws->label = strdup(ws->name + num_len);
```
RedSoxFan: `ws->label` also needs to be free'd here

swaybar/ipc.c
c-edw: Ah, so I need to free the pointer before reassign otherwise it will be dangling, correct?

RedSoxFan: Correct. `ws->label` held the only reference to the value.

c-edw: Got it, thanks.

swaybar/ipc.c
```diff
@@ -298,6 +304,23 @@ bool ipc_get_workspaces(struct swaybar *bar) {
 					calloc(1, sizeof(struct swaybar_workspace));
 				ws->num = json_object_get_int(num);
 				ws->name = strdup(json_object_get_string(name));
+				ws->label = strdup(ws->name);
+				char *ws_dup = strdup(ws->name);
+				char *stripped_name = strtok(ws_dup, ":");
```
ianyfan: I just did some testing and it seems like i3 strips names even if the colon is not present: i.e. "1a" will be stripped into 1 and a. So this logic will have to be changed if we want to match their behaviour.

c-edw: I think this could be a bug. The [docs](https://i3wm.org/docs/userguide.html#_strip_workspace_numbers_name) for `strip_workspace_name` and `strip_workspace_number` emphasise that the name needs to be in the form `[n]:[NAME]` to apply.

c-edw: They have confirmed that this is intended. I will see if I can get this done soon.

c-edw:
I have managed to get `strip_workspace_numbers` to work with/without a semicolon according to i3's behaviour, but having issues with `strip_workspace_name`. 

There's some kind of memory bug I can't figure out. I'm allocating 2 bytes, one for the first character and one for the null terminator, and then calling `strncat` to copy the first character, but something is going wrong here.

swaybar/ipc.c
```diff
@@ -305,22 +305,19 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
 				ws->label = strdup(ws->name);
-				char *ws_dup = strdup(ws->name);
-				char *stripped_name = strtok(ws_dup, ":");
-				size_t num_len = strlen(stripped_name) + 1;
-				// ws->num will be 0 if workspace name doesn't begin with int.
-				if (stripped_name != NULL && 
-						strlen(ws->name) > num_len && ws->num != -1) {
-					if(bar->config->strip_workspace_name) {
+				size_t len_offset = ws->name[1] == ':' ? 2 : 1;
+				// ws->num will be -1 if workspace name doesn't begin with int.
```
c-edw: This code makes the assumption that the digit will always be a single character (0-9). Not ideal, but existing code already does this.

swaybar/ipc.c
```diff
@@ -305,22 +305,19 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
+				if (ws->num != -1 && strlen(ws->name) > len_offset) {
+					if (bar->config->strip_workspace_name) {
 						free(ws->label);
-						ws->label = strdup(stripped_name);
-					} else if(bar->config->strip_workspace_numbers) {
+						ws->label = (char*) malloc(2 * sizeof(char));
+						strncat(ws->label, ws->name, 1);
```
ianyfan: Maybe you're thinking of `strncpy`

c-edw: I tried `strncpy` originally and thought that maybe the issues I've been seeing were because `strncpy` doesn't append null terminators (or does it?).

c-edw:
Ready for review again, it has the same behaviour as i3 AFAIK from testing.

swaybar/ipc.c
```diff
@@ -305,16 +305,18 @@ bool ipc_get_workspaces(struct swaybar *bar) {
 				ws->num = json_object_get_int(num);
 				ws->name = strdup(json_object_get_string(name));
 				ws->label = strdup(ws->name);
-				size_t len_offset = ws->name[1] == ':' ? 2 : 1;
+				size_t len_offset = numlen(ws->num);
```
ianyfan: This can do weird stuff if `ws->num == -1`.

i3 also strips the colon if the name is something like `1:`. I don't think this does that, though admittedly I haven't actually tested it.

c-edw: Nice catch with `numlen`, I have modified it to pass the absolute value to `log10`. Hanging colon will be stripped now too.

c-edw:
I have done some testing and these cases now work:
- strip_workspace_name `1234:` names the workspace `1234`
- strip_workspace_numbers `:abcd` names the workspace `abcd`
- strip_workspace_name `1234:abcd` names the workspace `1234`
- strip_workspace_numbers `1234:abcd` names the workspace `abcd`
- strip_workspace_name `1234abcd` names the workspace `1234`
- strip_workspace_numbers `1234abcd` names the workspace `abcd`

ianyfan:
strip_workspace_numbers `:abcd` should name the workspace `:abcd`

c-edw:
@ddevault It looks like something has gone wrong with the CI.

swaybar/ipc.c
```diff
@@ -306,18 +306,20 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
+						ws->label = malloc(len_offset + 1 * sizeof(char));
+						ws->label[len_offset] = '\0';
+						strncpy(ws->label, ws->name, len_offset);
+					} else if (bar->config->strip_workspace_numbers) {
+						len_offset += ws->label[len_offset] == ':';
+						if (strlen(ws->name) > len_offset) {
+							free(ws->label);
+							// Strip number prefix [1-?:] using len_offset.
+							ws->label = strdup(ws->name + len_offset);
+						}
```
ianyfan: If `strlen(ws->name) == len_offset` this should just strip the last colon, matching the behaviour of `strip_workspace_name`.

c-edw: Can you clarify further? Changing it to a `==` could cause problems because `len_offset` will always be one less than `strlen(ws->name)`.

ianyfan: "1:" gets displayed as "1" if `strip_workspace_number` is enabled (though not if both are disabled).

c-edw: But you said this is how it was on i3, no?

>i3 also strips the colon if the name is something like 1:. I don't think this does that, though admittedly I haven't actually tested it.

common/util.c
```diff
@@ -23,7 +23,7 @@ int numlen(int n) {
 	if (n == 0) {
 		return 1;
 	}
-	return log10(n) + 1;
+	return log10(abs(n)) + 1;
```
ianyfan: Now that we're not calling this with -1 again, could you either revert this change or reject negative numbers until we have a use case for it?

sway/commands/bar/strip_workspace_name.c
```diff
@@ -0,0 +1,32 @@
...
+	if ((error = checkarg(argc,
+				"strip_workspace_name", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	if (!config->current_bar) {
+		return cmd_results_new(CMD_FAILURE,
+				"strip_workspace_name", "No bar defined.");
+	}
+
+	config->current_bar->strip_workspace_name =
```
ianyfan: Trailing space

sway/commands/bar/strip_workspace_name.c
```diff
@@ -0,0 +1,32 @@
...
+
+	if (config->current_bar->strip_workspace_name) {
+		config->current_bar->strip_workspace_numbers = false;
+
+		wlr_log(WLR_DEBUG, "Stripping workspace name on bar: %s",
+				config->current_bar->id);
+	} else {
+		wlr_log(WLR_DEBUG, "Enabling workspace name on bar: %s",
+				config->current_bar->id);
+	}
```
ianyfan: Trailing space

sway/commands/bar/strip_workspace_numbers.c
```diff
@@ -13,17 +14,19 @@ struct cmd_results *bar_cmd_strip_workspace_numbers(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE,
 				"strip_workspace_numbers", "No bar defined.");
 	}
-	if (strcasecmp("yes", argv[0]) == 0) {
-		config->current_bar->strip_workspace_numbers = true;
+	
+	config->current_bar->strip_workspace_numbers =
```
ianyfan: Trailing space

sway/commands/bar/strip_workspace_numbers.c
```diff
@@ -13,17 +14,19 @@ struct cmd_results *bar_cmd_strip_workspace_numbers(int argc, char **argv) {
...
-	} else if (strcasecmp("no", argv[0]) == 0) {
-		config->current_bar->strip_workspace_numbers = false;
+	} else {
 		wlr_log(WLR_DEBUG, "Enabling workspace numbers on bar: %s",
 				config->current_bar->id);
-	} else {
-		return cmd_results_new(CMD_INVALID,
-				"strip_workspace_numbers", "Invalid value %s", argv[0]);
-	}
+	}
```
ianyfan: Trailing space

sway/sway-bar.5.scd
```diff
@@ -50,6 +50,10 @@ Sway allows configuring swaybar in the sway configuration file.
 *workspace\_buttons* yes|no
 	Enables or disables workspace buttons on the bar. Default is _yes_.
 
+*strip\_workspace\_name* yes|no
+	If set to _yes_, then workspace name will be omitted from the workspace
```
ianyfan: "then workspace name" is grammatically incorrect. Sucks because the i3 devs should have named it `strip_workspace_names` but that's what we got.
I'm also not sure about saying "custom number" either. Maybe just "number".

swaybar/ipc.c
```diff
@@ -153,8 +153,10 @@ static bool ipc_parse_config(
 		return false;
 	}
 	json_object *markup, *mode, *hidden_state, *position, *status_command;
-	json_object *font, *bar_height, *wrap_scroll, *workspace_buttons, *strip_workspace_numbers;
-	json_object *binding_mode_indicator, *verbose, *colors, *sep_symbol, *outputs;
+	json_object *font, *bar_height, *wrap_scroll, *workspace_buttons, 
+		*strip_workspace_numbers, *strip_workspace_name;
```
ianyfan: Follow what was done previously and start a new line of declarations `json_object *...` instead of overhanging it.

Also remember to fix trailing spaces.

swaybar/ipc.c
```diff
@@ -153,8 +153,10 @@ static bool ipc_parse_config(
...
+	json_object *binding_mode_indicator, *verbose, *colors, *sep_symbol, 
+		*outputs;
```
ianyfan: Same as above, should probably move to the line below with `bindings`

swaybar/ipc.c
```diff
@@ -298,6 +304,24 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
 				ws->num = json_object_get_int(num);
 				ws->name = strdup(json_object_get_string(name));
+				ws->label = strdup(ws->name);
+				// ws->num will be -1 if workspace name doesn't begin with int.
+				if(ws->num != -1) {
+					size_t len_offset = numlen(ws->num);
+					if (bar->config->strip_workspace_name) {
+						free(ws->label);
+						ws->label = malloc(len_offset + 1 * sizeof(char));
+						ws->label[len_offset] = '\0';
```
ianyfan: I just realised you should be able to remove the three other lines in this block and this will still work, since it'll terminate the string in the correct place, even if there's leftover characters after it.

swaybar/ipc.c
```diff
@@ -298,6 +304,24 @@ bool ipc_get_workspaces(struct swaybar *bar) {
...
+						strncpy(ws->label, ws->name, len_offset);
+					} else if (bar->config->strip_workspace_numbers) {
+						len_offset += ws->label[len_offset] == ':';
+						if (strlen(ws->name) > len_offset) {
+							free(ws->label);
```
ianyfan: You could remove this line and use `strcpy` below, though it's less of a fun solution as the one above.

swaybar/ipc.c
```diff
@@ -298,6 +304,24 @@ bool ipc_get_workspaces(struct swaybar *bar) {
 					calloc(1, sizeof(struct swaybar_workspace));
 				ws->num = json_object_get_int(num);
 				ws->name = strdup(json_object_get_string(name));
+				ws->label = strdup(ws->name);
+				// ws->num will be -1 if workspace name doesn't begin with int.
+				if(ws->num != -1) {
```
ianyfan: Style: space after `if`.

common/util.c
c-edw: I think it should remain like this. Negative numbers have lengths too, and before this change it would output a large negative number which could trigger a bug.

ianyfan: In that case, I think for negative numbers it should add an extra character to account for the '-' at the start.

emersion:
Thanks!

PR #3084, opened 2018-11-07
Use the env var for finding a socket.

aereaux:
This allows one to specify the wayland socket to use, as recommended by https://github.com/wayland-project/wayland/commit/c7dbaa1cfda68185b3b0d3665efdf4620c66531e

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -141,8 +141,15 @@ bool server_init(struct sway_server *server) {
 	wlr_export_dmabuf_manager_v1_create(server->wl_display);
 	wlr_screencopy_manager_v1_create(server->wl_display);
 
-	server->socket = wl_display_add_socket_auto(server->wl_display);
-	if (!server->socket) {
+	server->socket = getenv("WAYLAND_DISPLAY");
+	int ret = 0;
+	if (server->socket) {
+		ret = wl_display_add_socket(server->wl_display, server->socket);
+	}
+	else {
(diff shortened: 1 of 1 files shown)


RyanDwyer:
In this branch, to run sway in a nested session, I now have to either unset `WAYLAND_DISPLAY` or set it to something like `wayland-1`. Is this intentional?

aereaux:
It's not really intentional, but it is a side effect.  Do you think it would be better to get it from an argument instead?  That's the only easy fix I can think of.

ascent12:
I think adding this as a config option and/or a command line option would be better than an environment variable.
The client is the one that reads WAYLAND_DISPLAY. We're supposed to be the ones setting it.

martinetd:
Alternatively could try to use it or if it fails fallback to auto, but I agree using the same environment variable isn't ideal...

SirCmpwn:
Eh, I'm not a fan of this change.

emersion:
Me neither.

aereaux:
Is there another approach that would be better, like using an argument or a config option?  I really would like a way to do this, because I frequently run into the issue in the commit I linked.

SirCmpwn:
I'm not really interested in fixing this. You can export GDK_BACKEND=x11 or something similar in your X environment, perhaps.

emersion:
You can also set `WAYLAND_DISPLAY=''` in your X11 env.

PR #3086, opened 2018-11-08
gaps: remove duplicate inner gaps

RedSoxFan:
Removes duplicate inner gaps between two containers. See https://github.com/swaywm/sway/issues/307#issuecomment-165437419.

Instead of dividing the gaps by two between two containers as suggested, I decided to do the following:
1) Inner gaps are always added to the right and bottom sides
2) Inner gaps are only added to the top side when the container's y location matches the workspace's y location without inner gaps applied
3) Inner gaps are only added to the left side when the container's x location matches the workspace's x location without inner gaps applied

I think this solution is better since the inner gaps are guaranteed to be integer values instead of having to deal with half pixel values when dividing an odd inner gap by two.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -104,7 +104,12 @@ struct sway_container {
 	bool border_right;
 
 	// The gaps currently applied to the container.
-	double current_gaps;
+	struct {
+		int top;
+		int right;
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Thanks!

PR #3087, opened 2018-11-08
Implement per side and per direction outer gaps

### LINKED ISSUE #3059: gaps: support side-specific gap configuration
https://github.com/Airblader/i3/pull/211

RedSoxFan:
Closes #3059 

This introduces the following command extensions from `i3-gaps`:
* `gaps horizontal|vertical|top|right|bottom|left <amount>`
* `gaps horizontal|vertical|top|right|bottom|left all|current set|plus|minus <amount>`
* `workspace <ws> gaps horizontal|vertical|top|right|bottom|left <amount>`

`inner` and `outer` are also still available as options for all three of the above commands. `outer` now acts as a shorthand to set/alter all sides.

Additionally, this fixes two bugs with the prevention of invalid gap configurations for workspace configs:
1. If outer gaps were not set and inner gaps were, the outer gaps would be snapped to the negation of the inner gaps due to `INT_MIN` being less than the negation. This took precedence over the default outer gaps.
2. Similarly, if inner gaps were not set and outer gaps were, inner gaps would be set to zero, which would take precedence over the default inner gaps.

Fixing both of the above items also requires checking the gaps again when creating a workspace since the default outer gaps can be smaller than the negation of the workspace specific inner gaps.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -168,6 +168,16 @@ struct output_config {
 };
 
 /**
+ * Stores size of gaps for each side
+ */
+struct side_gaps {
+	int top;
+	int right;
+	int bottom;
+	int left;
+};
+
+/**
  * Stores configuration for a workspace, regardless of whether the workspace
  * exists.
  */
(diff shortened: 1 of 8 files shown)


RyanDwyer:
Thanks!

PR #3089, opened 2018-11-08
Fix building with clang

sghctoma:
The `struct sway_container *other` variable in swap.c was potentially used uninitialized, depending on an "if" statement.


emersion:
Thanks!

PR #3108, opened 2018-11-11
Allow multiple outputs for workspace output

RedSoxFan:
Related to #3082 
i3 documentation: https://i3wm.org/docs/userguide.html#workspace_screen
`workspace <workspace> output <outputs>`

`i3 4.16` allows users to list multiple outputs for a workspace and the first available will be used.

Additionally when the workspace is created, the outputs get added to the output priority list in the order specified. This ensures that if a higher output gets connected, the workspace will move to the higher output. This works the same way as if the user had a workspace on an output, disconnected the output, and then later reconnected the output.

### DIFF
--- a/sway/commands/workspace.c
+++ b/sway/commands/workspace.c
@@ -21,6 +21,7 @@ static struct workspace_config *workspace_config_find_or_create(char *ws_name) {
 		return NULL;
 	}
 	wsc->workspace = strdup(ws_name);
+	wsc->outputs = create_list();
 	wsc->gaps_inner = INT_MIN;
 	wsc->gaps_outer.top = INT_MIN;
 	wsc->gaps_outer.right = INT_MIN;
@@ -32,7 +33,7 @@ static struct workspace_config *workspace_config_find_or_create(char *ws_name) {
 
 void free_workspace_config(struct workspace_config *wsc) {
 	free(wsc->workspace);
-	free(wsc->output);
+	free_flat_list(wsc->outputs);
 	free(wsc);
 }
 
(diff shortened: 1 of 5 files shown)


sway/commands/workspace.c
```diff
@@ -32,7 +33,7 @@ static struct workspace_config *workspace_config_find_or_create(char *ws_name) {
 
 void free_workspace_config(struct workspace_config *wsc) {
 	free(wsc->workspace);
-	free(wsc->output);
+	free_flat_list(wsc->outputs);
```
emersion: Elements in the list need to be free'ed too.

sway/tree/workspace.c
```diff
@@ -134,8 +142,7 @@ void workspace_destroy(struct sway_workspace *workspace) {
 
 	free(workspace->name);
 	free(workspace->representation);
-	list_foreach(workspace->output_priority, free);
```
emersion: Why has this been removed?

sway/tree/workspace.c
```diff
@@ -281,10 +299,19 @@ char *workspace_next_name(const char *output_name) {
...
 		// Unlike with bindings, this does not guarantee order
 		const struct workspace_config *wsc = config->workspace_configs->items[i];
-		if (wsc->output && strcmp(wsc->output, output_name) == 0
-				&& workspace_by_name(wsc->workspace) == NULL) {
-			free(target);
-			target = strdup(wsc->workspace);
+		if (workspace_by_name(wsc->workspace)) {
+			continue;
+		}
+		int found = false;
```
emersion: This can be `bool`

emersion:
Thanks!

PR #3098, opened 2018-11-10
Use parse_boolean where possible.

c-edw:
Whilst working on #3083, I noticed some opportunities to refactor the commands arg parsing.

PR #3113, opened 2018-11-11
cmd_ws_auto_back_and_forth: fix negation

RedSoxFan:
In the conversion to `parse_boolean` for `cmd_ws_auto_back_and_forth` (#3098), the `!` was never removed (was previously using `strcasecmp`) causing the setting to be the opposite of what it should be.


emersion:
Thanks!

c-edw:
I left it there because it seemed like `yes` should yield `false` originally:
```c
!strcasecmp(argv[0], "yes");
```

PR #3117, opened 2018-11-12
Use #if instead of #ifdef for WLR_HAS_*

emersion:
Also removes a weird `setenv` declaration and a unused function.


SirCmpwn:
Thanks!

PR #3118, opened 2018-11-13
[Do not merge] fix focus_follows_mouse=always in case of not moving the mouse

madblobfish:
This patch focuses the correct window when changing workplaces with `focus_follows_mouse always` but when the mouse is not moved.
Until now I had to move the mouse for the focus to change.

I do not know if this works properly with multiseat and if there is a more simple and clean solution.

@c-edw: fyi

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -68,7 +68,8 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat);
 
 
 struct sway_node *node_at_cursor(struct sway_seat *seat,
-	struct sway_cursor *cursor, double *sx, double *sy);
+	struct sway_workspace *workspace, struct sway_cursor *cursor,
+	struct wlr_surface **surface, double *sx, double *sy);
 
 /**
  * "Rebase" a cursor on top of whatever view is underneath it.
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -66,6 +66,10 @@ struct sway_node *node_at_coords(
 void sway_cursor_destroy(struct sway_cursor *cursor);
 struct sway_cursor *sway_cursor_create(struct sway_seat *seat);
 
+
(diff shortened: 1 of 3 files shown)


c-edw:
This was intentional, because the idea was that it should be compatible with keyboard or mouse based workflows. This change would impact a keyboard based workflow, since you cant prevent the window being immediately refocused. I have no problem if you want to go ahead with this change though.

sway/commands/workspace.c
```diff
@@ -216,6 +217,9 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 			free(name);
 		}
 		workspace_switch(ws, no_auto_back_and_forth);
+		if(config->focus_follows_mouse == FOLLOWS_ALWAYS){
```
c-edw: Be sure to adhere to the style guide. 
- `if(` -> `if (
- `){` -> `) {`

madblobfish: whoops, fixed

madblobfish:
> This change would impact a keyboard based workflow
True. I'm not sure how to proceed then, another config option would probably hurt more than help.

We should at least change the documentation accordingly to mention this. Currently it is:
> If set to always, the window under the cursor will always be focused, even after switching between workspaces.

madblobfish:
Also just noticed we need another fix for floating windows which steal focus.

c-edw:
>True. I'm not sure how to proceed then, another config option would probably hurt more than help.

Yes, I don't think adding another config option is the way forward.

>We should at least change the documentation accordingly to mention this.

It is a bit misleading, yeah. Maybe you could change it to say something like `the window under the cursor will be focused upon moving the cursor, after switching between workspaces`.

madblobfish:
Thought a bit longer about this and I think we should rename the current behavior to something like `on_move`, as `always` is misleading.

I'll write another patch changing the behavior of `always` what I would expect.
This will change `seat_set_focus` instead of workspace so that we should have more future proof and clean code.

PR #3121, opened 2018-11-13
[WIP] Free unfreed resources.

c-edw:
Hunting down some unfreed resources.

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -35,18 +35,19 @@ bool criteria_is_empty(struct criteria *criteria) {
 }
 
 void criteria_destroy(struct criteria *criteria) {
+	free(criteria->raw);
+	free(criteria->cmdlist);
+	free(criteria->target);
 	pcre_free(criteria->title);
 	pcre_free(criteria->shell);
 	pcre_free(criteria->app_id);
+	pcre_free(criteria->con_mark);
 #ifdef HAVE_XWAYLAND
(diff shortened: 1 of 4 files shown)


sway/criteria.c
```diff
@@ -35,18 +35,19 @@ bool criteria_is_empty(struct criteria *criteria) {
 }
 
 void criteria_destroy(struct criteria *criteria) {
+	free(criteria->raw);
+	free(criteria->cmdlist);
+	free(criteria->target);
```
RyanDwyer: I don't think you want to free the target. I'll look into this later if no one has confirmed this.

c-edw: It's a `char*`, are you suggesting that something else might use the pointer?

RyanDwyer: Oh, right. Freeing it is fine then.

c-edw:
It was still a WIP, but I can open another PR if I find more changes.

ddevault:
Derp, sorry.

c-edw:
I've just noticed a few commits that address double-frees caused by this PR:
- 828060af73dd2f3138ad9d1bf743f57dd567723c
- c36665bc17d1950672330788040e07953d4fb638

 I didn't have a good enough understanding of C when I sent this in (and still don't). I'm thinking perhaps this should be reverted.

c-edw:
May also be the cause of #3545.

Issue #3123, opened 2018-11-13
Failed to build in the latest commit # 47cca240


adouzzy:
git commit:
47cca240 2018-11-13 (HEAD -> master, origin/master, origin/HEAD) Merge pull request #3102 from emersion/render-software-cursors [GitHub]

Error Message:
`
[78/270] Compiling C object 'sway/sway@@sway@exe/desktop_output.c.o'.
FAILED: sway/sway@@sway@exe/desktop_output.c.o
cc -Isway/sway@@sway@exe -Isway -I../sway -Iinclude -I../include -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/uuid -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/gdk-pixbuf-2.0 -I/usr/local/include -I/usr/local/include/json-c -I/usr/include/pango-1.0 -I/usr/include/fribidi -I/usr/include/libdrm -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=c11 -g -DWL_HIDE_DEPRECATED -DWLR_USE_UNSTABLE -Wno-unused-parameter -Wno-unused-result -Wundef '-DSYSCONFDIR="//usr/local/etc"' '-DSWAY_VERSION="1.0-beta.1-152-g47cca240 (" __DATE__ ", branch '"'"'master'"'"')"' -pthread  -MD -MQ 'sway/sway@@sway@exe/desktop_output.c.o' -MF 'sway/sway@@sway@exe/desktop_output.c.o.d' -o 'sway/sway@@sway@exe/desktop_output.c.o' -c ../sway/desktop/output.c
../sway/desktop/output.c: In function ‘damage_surface_iterator’:
../sway/desktop/output.c:420:3: error: implicit declaration of function ‘wlr_surface_get_effective_damage’; did you mean ‘wlr_surface_get_extends’? [-Werror=implicit-function-declaration]
   wlr_surface_get_effective_damage(surface, &damage);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   wlr_surface_get_extends
cc1: all warnings being treated as errors
[79/270] Compiling C object 'sway/sway@@sway@exe/desktop_render.c.o'.
[...]

PR #3126, opened 2018-11-14
Fix unused function in swap command

emersion:
```
Error: test_id defined but not used. [-Werror=unused-function].
```

Also fixes a `size_t` cast to `void *`.

https://github.com/swaywm/sway/issues/3123#issuecomment-438763220


ddevault:
Thanks!

PR #3142, opened 2018-11-17
Move view {x,y,width,height} into container struct

RyanDwyer:
This renames/moves the following properties:

* `sway_view.{x,y,width,height}` -> `sway_container.content_{x,y,width,height}`
    * This is required to support placeholder containers as they don't have a view.
* `sway_container_state.view_{x,y,width,height}` -> `sway_container_state.content_{x,y,width,height}`
    * To remain consistent with the above.
* `sway_container_state.con_{x,y,width,height}` -> `sway_container_state.{x,y,width,height}`
    * The `con` prefix was there to give it contrast from the view properties, and is no longer useful.

The function `container_set_geometry_from_floating_view` has also been renamed to `container_set_geometry_from_content`.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -41,8 +41,8 @@ enum wlr_direction;
 struct sway_container_state {
 	// Container properties
 	enum sway_container_layout layout;
-	double con_x, con_y;
-	double con_width, con_height;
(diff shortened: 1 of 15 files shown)


ddevault:
Thanks!

PR #3143, opened 2018-11-17
commands/resize: fix grow vars uninitialized

colemickens:
I'm not sure I understand why the PPT/PX codepaths are different and thus why these would be uninitialized. Hopefully zero is a sane default.

This fixes a build error and my packaging treats it as an error. Thanks.


colemickens:
and/or it doesn't know that `width->amount` and `height->amount` will always be non-negative, maybe?

anyway, thanks for the fast review as always.

emersion:
>and/or it doesn't know that width->amount and height->amount will always be non-negative, maybe?

I don't think it matters whether these are negative or positive.

>anyway, thanks for the fast review as always.

:)

ddevault:
Thanks!

Issue #2965, opened 2018-10-24
wayland compatible terminals don't start without xwayland


praetorian55:
Both gnome-terminal and terminator do not start without xwayland installed when called via exec in the config.  Interestingly enough, executing them from an ssh session after sway was already started does cause them to open on the display.  Both open as expected when xwayland is installed

I have also tried setting GDK_BACKEND=wayland via script, and when executing the terminal directly (GDK_BACKEND=wayland gnome-terminal)

_sway version 1.0-beta.1-31-ge33671fd (Oct 23 2018, branch 'master')_

[config.txt](https://github.com/swaywm/sway/files/2510957/config.txt)
[sway.log](https://github.com/swaywm/sway/files/2510954/sway.log)

mteyssier:
It might actually explain why I couldn't start termite on the first try, just after installing sway.
Thanks for the report!

I uninstalled xwayland to run some tests (using termite instead of gnome-terminal or terminator):

Using sway:
- indeed termite doesn't start if called via exec in the config
- weston-terminal starts fine in sway!
- lauching termite from a weston-terminal: termite doesn't return and doesn't display any window.
- strace on termite: it is stuck on `connect(... "/tmp/.X11-unix/X0")`.

Using weston:
- both weston-terminal and termite are running fine.

The key is the `DISPLAY` environment variable.
It is not set by weston. (And is not set in your ssh session, right?)

You can try to unset the `DISPLAY` variable (via script?) and your terminal should run fine, something like:
`env -u DISPLAY termite` or just `unset DISPLAY` before.

Anyway the `DISPLAY` variable shouldn't be set, should it?

I just skimmed through the code but it looks like a wlroots issue, not  handling the error producing `[xwayland/xwayland.c:144] failed to exec Xwayland: No such file or directory` and thus not  unsetting `DISPLAY`.

After a quick glance at sway/server.c, currently sway would not properly handle this kind of error anyway (if compiled with `HAVE_XWAYLAND`).

emersion:
We start Xwayland lazyly, which means we create the socket and set `DISPLAY` before starting Xwayland and start it only when an app connects.

Two things can be done:

* We should add a runtime config option to disable Xwayland
* Maybe we should check for Xwayland before starting lazily? (This would require handling `PATH`, not sure it's actually a good idea)

praetorian55:
[...]

PR #3144, opened 2018-11-17
Add xwayland command

emersion:
This allows xwayland to be disabled at runtime.

Xwayland needs to be started in `server_start`, after the config has been read.

Closes https://github.com/swaywm/sway/issues/2965

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -176,8 +176,9 @@ sway_cmd cmd_title_format;
 sway_cmd cmd_unmark;
 sway_cmd cmd_urgent;
 sway_cmd cmd_workspace;
-sway_cmd cmd_ws_auto_back_and_forth;
 sway_cmd cmd_workspace_layout;
+sway_cmd cmd_ws_auto_back_and_forth;
+sway_cmd cmd_xwayland;
 
 sway_cmd bar_cmd_activate_button;
 sway_cmd bar_cmd_binding_mode_indicator;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -393,6 +393,7 @@ struct sway_config {
 	size_t urgent_timeout;
 	enum sway_fowa focus_on_window_activation;
 	enum sway_popup_during_fullscreen popup_during_fullscreen;
+	bool xwayland;
 
 	// Flags
 	enum focus_follows_mouse_mode focus_follows_mouse;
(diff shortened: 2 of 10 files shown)


RyanDwyer:
No docs?

Also, I assume you mean config-time and not runtime.

emersion:
>No docs?

Oh right, I always forget those. Fixed.

>Also, I assume you mean config-time and not runtime.

I meant "runtime" as the opposite of "compile-time". But yeah it can only be used in the config file.

ddevault:
Hmm, I don't think I like this. I would prefer it as `-Dxwayland=no`

emersion:
Disabling Wayland is not a debug option. It should be in the config file, you don't want to do it temporarily.

ddevault:
If you want to do it long-term, just uninstall Xwayland?

emersion:
The whole point is that uninstalling Wayland doesn't work. wlroots will still try to to launch it (which will fail, but set DISPLAY to a bad value). See the original issue.

emersion:
> Seems like a bug.

Well, our only option would be to try to find Xwayland in `$PATH` which is not cool.

emersion:
It's also useful to be able to disable Xwayland even if you have it installed.

ddevault:
I dunno, I'm still -1 to this. We'll see what everyone else thinks.

emersion:
You haven't explained why you don't like it, though.

ddevault:
All features have a maintenance and complexity cost. I don't think this one pays its bill.

emersion:
I don't think a +72 −39 feature will have a high maintenance cost. Users can disable Xwayland anyway at build-time, so doing it at runtime doesn't add complexity. Eventually we'll want to have it anyway, as more applications migrate to Wayland. It's also already available in rootston, and the maintenance cost there has been zero.

ddevault:
rootston isn't a user-facing compositor, we don't have to deal with support.

RyanDwyer:
The only use case I can think of is a user who doesn't want to use Xwayland and who is using a distribution version of sway which has it enabled at compile time. It may be useful for someone who is running sway on a low memory device and wants as many resources as possible available for other things.

I don't see any maintenance or complexity overhead in having it, so +1 from me.

RedSoxFan:
> >  wlroots will still try to to launch it (which will fail, but set DISPLAY to a bad value).
>
> Seems like a bug.

I agree that this does not sound like the correct/ideal behaviour, but I am not currently familiar enough with related code to suggest a solution.

With that said, I don't think that this adds any maintenance or complexity overhead since it can only be adjusted on launch so +1 from me.

I would like to suggest a small addition/change to this implementation however. Currently on reload, the command will be executed but silently ignored. I think that if the config is being reloaded and the value changes, swaynag should be spawned to notify the user that the new value won't take effect until sway is restarted.

ddevault:
Thanks!

Issue #2361, opened 2018-07-26
Keyboard layout-per-window feature


Adonai:
Hi, was just wondering if layout-per-window feature possible or planned in swaywm.

Thanks for this great project.

SirCmpwn:
This is possible in theory, but no one has started working on it.

manonthemat:
I would love to have that feature.

SirCmpwn:
So start working on it.

Hi-Angel:
Could somebody briefly outline what to look at?

Am I correct that "windows" in the code are wrapped into `wlr_box`?

Where or how "keyboard focused window" is implemented? *([as I recently learned](https://github.com/swaywm/sway/issues/2859), currently focused window is not necessarily the one receiving keyboard input — does that need to be changed btw?)*

Hi-Angel:
Some status update:

So far I figured:

1. A window is represented by `sway_view`, so I need to keep layout state there
2. An active layout seems to be stored in an opaque `struct xkb_state`; with bunch of other stuff such as enabled LEDs and pressed keys.
3. It seems, layout change is not handled by sway at all, but rather by libxkbcommon. I.e. I got an impression that sway knows nothing about current keyboard layout.

My current problems:

1. How to get notified of layout change.
2. How to extract the new layout *(to store it into `sway_view`)*.
3. How to change layout *(e.g. when input focus changed from one `sway_view` to another)*.

Hi-Angel:
[My work-in-progress branch](https://github.com/Hi-Angel/sway/tree/kbd-layout-per-window). TODOs:

1. Where to put the code to stash keyboard layout when keyboard focus changes from one `sway_view` to another? I'm guessing it's probably the same place where I'm restoring layout *(`seat_send_focus()`)*, then I need to figure out, how to get the previously focused `sway_view`.
2. How a configuration option to enable it should look like?
3. The `xkb_layout_index_t` is probably specific to a `xkb_state`, hence I need to store it for every keyboard device.

Hi-Angel:
I kind of stuck with the following question: after I updated `xkb_state` with new layout, I need to forward the update down the stack to change state of LEDs *(if needed)* and probably something else *(not sure yet what is exactly, but without that the next keypress uses old keyboard layout)*.

So I did some research with debugger, and came to conclusion that physical keyboard state update is done by wlroots. However I didn't see a function that simply makes update, given a new `xkb_state`. Do I need to extract that functional, or am I missing something?

[...]

PR #3155, opened 2018-11-19
Kbd layout per window

Hi-Angel:
This is a work-in-progress code to fix https://github.com/swaywm/sway/issues/2361

It does work as it is, but there's one last todo that I wanted to consult about.

If you highlight "todo" in last commit, you'll see that I'm currently storing only one `xkb_layout_index_t` in every window. However I'm wondering, whether this `xkb_layout_index_t` might have a different meaning for different keyboards. In which case I'd need to keep a list of structs `struct curr_kbd_layout { wlr_keyboard *kbd_owner; xkb_layout_index_t layout };`. And then on every keyboard destroy I'd need to walk over all `sway_view`s, and remove the element in list that refers to the keyboard being destroyed.

Is it the preferred way of dealing with it? Or is there something simpler? Like, if `xkb_layout_index_t` is guaranteed to map to the same layout for any keyboard on the system, then the current code already works well.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -41,6 +41,12 @@ enum binding_flags {
 	BINDING_CODE=32,     // keyboard only; convert keysyms into keycodes
 };
 
+enum sway_keep_keyboard_layout {
+	KEYBOARD_LAYOUT_UNDEFINED = 0,
+	KEYBOARD_LAYOUT_GLOBAL,
+	KEYBOARD_LAYOUT_PER_WINDOW
+};
+
 /**
  * A key binding and an associated command.
  */
@@ -165,6 +171,7 @@ struct seat_config {
 	list_t *attachments; // list of seat_attachment configs
 	int hide_cursor_timeout;
 	enum seat_config_allow_constrain allow_constrain;
+	enum sway_keep_keyboard_layout keep_keyboard_layout;
 };
 
 enum config_dpms {
--- a/include/sway/input/keyboard.h
+++ b/include/sway/input/keyboard.h
@@ -63,9 +63,15 @@ struct sway_keyboard {
 
 	struct wl_event_source *key_repeat_source;
 	struct sway_binding *repeat_binding;
+	xkb_layout_index_t default_kbd_layout;
 };
 
 struct xkb_keymap *sway_keyboard_compile_keymap(struct input_config *ic);
+
+struct sway_layout_per_kbd {
+	struct wlr_keyboard *kbd;
+	xkb_layout_index_t layout;
+};
 
 struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
 		struct sway_seat_device *device);
@@ -73,6 +79,10 @@ struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
 void sway_keyboard_configure(struct sway_keyboard *keyboard);
 
 void sway_keyboard_destroy(struct sway_keyboard *keyboard);
+
+void sway_keyboard_set_layout(struct wlr_keyboard *kbd, xkb_layout_index_t layout);
+
+xkb_layout_index_t sway_keyboard_get_layout(struct xkb_state *kbd_state);
 
 void sway_keyboard_disarm_key_repeat(struct sway_keyboard *keyboard);
 #endif
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -12,6 +12,7 @@
 #include "sway/input/keyboard.h"
 #include "sway/input/seat.h"
 #include "sway/ipc-server.h"
+#include "sway/tree/view.h"
 #include "log.h"
 
 static struct modifier_key {
@@ -475,6 +476,20 @@ static void handle_keyboard_modifiers(struct wl_listener *listener,
 	determine_bar_visibility(modifiers);
 }
 
+static void add_kbd_to_view(struct sway_container *con, void *data) {
+	if (!con->view) {
+		return;
+	}
+
+	struct sway_layout_per_kbd *lpk = calloc(1, sizeof(struct sway_layout_per_kbd));
+	if (!sway_assert(lpk, "Failed to allocate sway_layout_per_kbd")) {
+		return;
+	}
+
+	*lpk = *(struct sway_layout_per_kbd *)data;
+	list_add(con->view->kbd_layouts, lpk);
+}
+
 struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
 		struct sway_seat_device *device) {
 	struct sway_keyboard *keyboard =
@@ -580,6 +595,32 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
 	wl_signal_add(&wlr_device->keyboard->events.modifiers,
 		&keyboard->keyboard_modifiers);
 	keyboard->keyboard_modifiers.notify = handle_keyboard_modifiers;
+
+	keyboard->default_kbd_layout =
+		sway_keyboard_get_layout(wlr_device->keyboard->xkb_state);
+	struct sway_layout_per_kbd lpk = {
+		.kbd = wlr_device->keyboard,
+		.layout = keyboard->default_kbd_layout
+	};
+	root_for_each_container(add_kbd_to_view, &lpk);
+}
+
+static void rm_kbd_from_view(struct sway_container *con, void *data) {
+	if (!con->view) {
+		return;
+	}
+
+	list_t *kbd_layouts = con->view->kbd_layouts;
+	for (int i = 0; i < kbd_layouts->length; ++i) {
+		struct sway_layout_per_kbd *lpk = kbd_layouts->items[i];
+		if (lpk->kbd == (struct wlr_keyboard *)data) {
+			free(lpk);
+			list_del(kbd_layouts, i);
+			return;
+		}
+	}
+	sway_assert(false,
+			"sway_layout_per_kbd not found in sway_view for removal");
 }
 
 void sway_keyboard_destroy(struct sway_keyboard *keyboard) {
@@ -593,5 +634,19 @@ void sway_keyboard_destroy(struct sway_keyboard *keyboard) {
 	wl_list_remove(&keyboard->keyboard_modifiers.link);
 	sway_keyboard_disarm_key_repeat(keyboard);
 	wl_event_source_remove(keyboard->key_repeat_source);
+	root_for_each_container(rm_kbd_from_view,
+			keyboard->seat_device->input_device->wlr_device->keyboard);
 	free(keyboard);
+}
+
+void sway_keyboard_set_layout(struct wlr_keyboard *kbd, xkb_layout_index_t layout) {
+	wlr_keyboard_notify_modifiers(kbd,
+			kbd->modifiers.depressed,
+			kbd->modifiers.latched,
+			kbd->modifiers.locked,
+			layout);
+}
+
+xkb_layout_index_t sway_keyboard_get_layout(struct xkb_state *kbd_state) {
+	return xkb_state_serialize_layout(kbd_state, XKB_STATE_LAYOUT_EFFECTIVE);
 }
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -79,6 +79,13 @@ static void seat_send_activate(struct sway_node *node, struct sway_seat *seat) {
 	}
 }
 
+static void set_kbd_layouts(const struct sway_view *view) {
+	for (int i = 0; i < view->kbd_layouts->length; ++i) {
+		struct sway_layout_per_kbd *lpk = view->kbd_layouts->items[i];
+		sway_keyboard_set_layout(lpk->kbd, lpk->layout);
+	}
+}
+
 /**
  * If con is a view, set it as active and enable keyboard input.
  * If con is a container, set all child views as active and don't enable
@@ -99,6 +106,12 @@ static void seat_send_focus(struct sway_node *node, struct sway_seat *seat) {
 #endif
 		struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(seat->wlr_seat);
 		if (keyboard) {
+			const struct seat_config *seat_config = seat_get_config(seat);
+			if (seat_config) {
+                if (seat_config->keep_keyboard_layout == KEYBOARD_LAYOUT_PER_WINDOW) {
+                    set_kbd_layouts(view);
+                }
+			}
 			wlr_seat_keyboard_notify_enter(seat->wlr_seat,
 					view->surface, keyboard->keycodes,
 					keyboard->num_keycodes, &keyboard->modifiers);
@@ -765,6 +778,13 @@ static void send_unfocus(struct sway_container *con, void *data) {
 	}
 }
 
+static void save_kbd_layouts(struct sway_view *view) {
+	for (int i = 0; i < view->kbd_layouts->length; ++i) {
+		struct sway_layout_per_kbd *lpk = view->kbd_layouts->items[i];
+		lpk->layout = sway_keyboard_get_layout(lpk->kbd->xkb_state);
+	}
+}
+
 // Unfocus the container and any children (eg. when leaving `focus parent`)
 static void seat_send_unfocus(struct sway_node *node, struct sway_seat *seat) {
 	sway_cursor_constrain(seat->cursor, NULL);
@@ -772,6 +792,15 @@ static void seat_send_unfocus(struct sway_node *node, struct sway_seat *seat) {
 	if (node->type == N_WORKSPACE) {
 		workspace_for_each_container(node->sway_workspace, send_unfocus, seat);
 	} else {
+		struct sway_view *view = node->type == N_CONTAINER ?
+			node->sway_container->view : NULL;
+		const struct seat_config *seat_config = seat_get_config(seat);
+
+		if (seat_config && view && seat_is_input_allowed(seat, view->surface) &&
+				seat_config->keep_keyboard_layout == KEYBOARD_LAYOUT_PER_WINDOW) {
+			save_kbd_layouts(view);
+		}
+
 		send_unfocus(node->sway_container, seat);
 		container_for_each_child(node->sway_container, send_unfocus, seat);
 	}
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -18,6 +18,7 @@
 #include "sway/desktop.h"
 #include "sway/desktop/transaction.h"
 #include "sway/input/cursor.h"
+#include "sway/input/keyboard.h"
 #include "sway/ipc-server.h"
 #include "sway/output.h"
 #include "sway/input/seat.h"
@@ -30,13 +31,36 @@
 #include "pango.h"
 #include "stringop.h"
 
-void view_init(struct sway_view *view, enum sway_view_type type,
+static bool view_setup_kbd_layouts(struct sway_view *view) {
+	struct sway_seat *seat = NULL;
+	wl_list_for_each(seat, &server.input->seats, link) {
+		struct sway_seat_device *device = NULL;
+		wl_list_for_each(device, &seat->devices, link) {
+			struct wlr_input_device *wlr_input_device = device->input_device->wlr_device;
+			if (wlr_input_device->type != WLR_INPUT_DEVICE_KEYBOARD) {
+				continue;
+			}
+			struct sway_layout_per_kbd *lpk = calloc(1, sizeof(struct sway_layout_per_kbd));
+			if (!sway_assert(lpk, "Failed to allocate sway_layout_per_kbd")) {
+				return false;
+			}
+			lpk->kbd = wlr_input_device->keyboard;
+			lpk->layout = device->keyboard->default_kbd_layout;
+			list_add(view->kbd_layouts, lpk);
+		}
+	}
+	return true;
+}
+
+bool view_init(struct sway_view *view, enum sway_view_type type,
 		const struct sway_view_impl *impl) {
 	view->type = type;
 	view->impl = impl;
 	view->executed_criteria = create_list();
+	view->kbd_layouts = create_list();
 	view->allow_request_urgent = true;
 	wl_signal_init(&view->events.unmap);
+	return view_setup_kbd_layouts(view);
 }
 
 void view_destroy(struct sway_view *view) {
@@ -55,6 +79,7 @@ void view_destroy(struct sway_view *view) {
 	list_free(view->executed_criteria);
 
 	free(view->title_format);
+	list_free_items_and_destroy(view->kbd_layouts);
 
 	if (view->impl->destroy) {
 		view->impl->destroy(view);
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -277,6 +277,7 @@ sway_cmd seat_cmd_attach;
 sway_cmd seat_cmd_cursor;
 sway_cmd seat_cmd_fallback;
 sway_cmd seat_cmd_hide_cursor;
+sway_cmd seat_cmd_keep_keyboard_layout;
 sway_cmd seat_cmd_pointer_constraint;
 
 sway_cmd cmd_ipc_cmd;
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -111,6 +111,8 @@ struct sway_view {
 	} events;
 
 	struct wl_listener surface_new_subsurface;
+
+	list_t *kbd_layouts; // struct sway_layout_per_kbd *
 };
 
 struct sway_xdg_shell_v6_view {
@@ -309,7 +311,7 @@ void view_for_each_popup(struct sway_view *view,
 
 // view implementation
 
-void view_init(struct sway_view *view, enum sway_view_type type,
+bool view_init(struct sway_view *view, enum sway_view_type type,
 	const struct sway_view_impl *impl);
 
 void view_destroy(struct sway_view *view);
--- a/sway/commands/seat.c
+++ b/sway/commands/seat.c
@@ -12,6 +12,7 @@ static struct cmd_handler seat_handlers[] = {
 	{ "cursor", seat_cmd_cursor },
 	{ "fallback", seat_cmd_fallback },
 	{ "hide_cursor", seat_cmd_hide_cursor },
+	{ "keep_keyboard_layout", seat_cmd_keep_keyboard_layout },
 	{ "pointer_constraint", seat_cmd_pointer_constraint },
 };
 
--- a/sway/commands/seat/keep_keyboard_layout.c
+++ b/sway/commands/seat/keep_keyboard_layout.c
@@ -1,0 +1,29 @@
+#include "log.h"
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/input/input-manager.h"
+
+struct cmd_results *seat_cmd_keep_keyboard_layout(int argc, char **argv) {
+	const char *STR_LAYOUT_GLOBAL = "global",
+		*STR_LAYOUT_PER_WINDOW = "per_window";
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "keep_keyboard_layout", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	if (!config->handler_context.seat_config) {
+		return cmd_results_new(CMD_FAILURE, "No seat defined");
+	}
+
+	if (!strcmp(argv[0], STR_LAYOUT_GLOBAL)) {
+		config->handler_context.seat_config->keep_keyboard_layout
+			= KEYBOARD_LAYOUT_GLOBAL;
+		return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+	} else if (!strcmp(argv[0], STR_LAYOUT_PER_WINDOW)) {
+		config->handler_context.seat_config->keep_keyboard_layout
+			= KEYBOARD_LAYOUT_PER_WINDOW;
+		return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+	} else {
+		return cmd_results_new(CMD_FAILURE, "keep_keyboard_layout",
+				"Expected 'keep_keyboard_layout global|per_window");
+	}
+}
--- a/sway/config/seat.c
+++ b/sway/config/seat.c
@@ -27,6 +27,7 @@ struct seat_config *new_seat_config(const char* name) {
 	}
 	seat->hide_cursor_timeout = -1;
 	seat->allow_constrain = CONSTRAIN_DEFAULT;
+	seat->keep_keyboard_layout = KEYBOARD_LAYOUT_UNDEFINED;
 
 	return seat;
 }
@@ -114,6 +115,10 @@ static void merge_seat_attachment_config(struct seat_attachment_config *dest,
 void merge_seat_config(struct seat_config *dest, struct seat_config *source) {
 	if (source->fallback != -1) {
 		dest->fallback = source->fallback;
+	}
+
+	if (source->keep_keyboard_layout != KEYBOARD_LAYOUT_UNDEFINED) {
+		dest->keep_keyboard_layout = source->keep_keyboard_layout;
 	}
 
 	for (int i = 0; i < source->attachments->length; ++i) {
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -504,7 +504,11 @@ void handle_xdg_shell_surface(struct wl_listener *listener, void *data) {
 		return;
 	}
 
-	view_init(&xdg_shell_view->view, SWAY_VIEW_XDG_SHELL, &view_impl);
+	if (!view_init(&xdg_shell_view->view, SWAY_VIEW_XDG_SHELL, &view_impl)) {
+		sway_assert(false, "Failed to init view");
+		return;
+	}
+
 	xdg_shell_view->view.wlr_xdg_surface = xdg_surface;
 
 	xdg_shell_view->map.notify = handle_map;
--- a/sway/desktop/xdg_shell_v6.c
+++ b/sway/desktop/xdg_shell_v6.c
@@ -492,7 +492,11 @@ void handle_xdg_shell_v6_surface(struct wl_listener *listener, void *data) {
 		return;
 	}
 
-	view_init(&xdg_shell_v6_view->view, SWAY_VIEW_XDG_SHELL_V6, &view_impl);
+	if (!view_init(&xdg_shell_v6_view->view, SWAY_VIEW_XDG_SHELL_V6, &view_impl)) {
+		sway_assert(false, "Failed to init view");
+		return;
+	}
+
 	xdg_shell_v6_view->view.wlr_xdg_surface_v6 = xdg_surface;
 
 	xdg_shell_v6_view->map.notify = handle_map;
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -595,7 +595,11 @@ void handle_xwayland_surface(struct wl_listener *listener, void *data) {
 		return;
 	}
 
-	view_init(&xwayland_view->view, SWAY_VIEW_XWAYLAND, &view_impl);
+	if (!view_init(&xwayland_view->view, SWAY_VIEW_XWAYLAND, &view_impl)) {
+		sway_assert(false, "Failed to init view");
+		return;
+	}
+
 	xwayland_view->view.wlr_xwayland_surface = xsurface;
 
 	wl_signal_add(&xsurface->events.destroy, &xwayland_view->destroy);
--- a/sway/meson.build
+++ b/sway/meson.build
@@ -88,6 +88,7 @@ sway_sources = files(
 	'commands/seat/cursor.c',
 	'commands/seat/fallback.c',
 	'commands/seat/hide_cursor.c',
+	'commands/seat/keep_keyboard_layout.c',
 	'commands/seat/pointer_constraint.c',
 	'commands/set.c',
 	'commands/show_marks.c',
--- a/sway/sway-input.5.scd
+++ b/sway/sway-input.5.scd
@@ -201,6 +201,9 @@ correct seat.
 	disables hiding the cursor. The minimal timeout is 100 and any value less
 	than that (aside from 0), will be increased to 100.
 
+*seat* <name> keep_keyboard_layout global|per_window
+	Controls whether to keep keyboard layout per window. Default is _global_.
+
 *seat* <name> pointer_constraint enable|disable|escape
 	Enables or disables the ability for clients to capture the cursor (enabled
 	by default) for the seat. This is primarily useful for video games. The


progandy:
I have another question. Will this work with multiple seats, so each seat can have a different keyboard layout in the window?

Hi-Angel:
> I have another question. Will this work with multiple seats, so each seat can have a different keyboard layout in the window?

That's a good point. Okay, then, steps to finish this PR probably would be:

1. Store in every window a collection of layouts per every keyboard
2. Store in every window a collection up to number of seats of collections of layouts per every keyboard.

emersion:
>However I'm wondering, whether this xkb_layout_index_t might have a different meaning for different keyboards.

Indeed, the keyboard layout index is per XKB context.

Hi-Angel:
Is there an implementation of a dynamically resized array, or do I have to make one? Till now I've only seen a list implementation, which is less suited here.

RyanDwyer:
Yes, use `list_t` (ie. `create_list`, `list_add`, etc). It's a dynamically resizing array.

Hi-Angel:
It's a list. It have good performance for removing/adding elements, but bad performance for traversing.

In my case operations of adding/removing are very rare, whereas traversing happens often, so I need a dynamic array.

emersion:
>It's a list.

It's not a linked list. It's an array. Yes, it's badly named.

Hi-Angel:
FTR, it seems every keyboard has one seat it's assigned to.
screenshot_20181124_193458

I think then, storing a layout per keyboard should solve the question of correct layouts per seat too.

Hi-Angel:
Oh, it must be `root_for_each_container()`.

Hi-Angel:
How do I loop over all keyboards? E.g. in `view_init` I need to loop over `sway_keyboards`, to get their default layouts.

Hopefully, it's the last thing I have to do *(well, codewise, I still need to test it though)*.

Hi-Angel:
Thanks all for your tips, it should work now. I'm hoping the last commit not too big? I really don't see any way to split it. It might help to paste my checklist I've been following yesterday to make sure I didn't miss anything:

1. ✓ add `default_kbd_layout` to `sway_keyboard`.
2. ✓ add to `sway_view`: `{*wlr_keyboard, layout}`
3. ✓ on window losing focus: go through the list of the tuples, save the layout
4. ✓ on window getting focused: go through the list of tuples, set layout
5. ✓ on keyboard destroyed: go through views, remove the tuple
6. ✓ on keyboard created:
  * set `default_kbd_layout` in `sway_keyboard`
  * go through list of `sway_view`s, add a new tuple
7. ✓ on window created: go through `sway_keyboard`s, add tuples with `default_kbd_layout`
8. ✓ on window destroyed: free the tuples.

But of course it's not a substitution for the code :)

Hi-Angel:
v2: I forgot some brackets around if-conditions.

Hi-Angel:
I just made a little update to the code: the $mod key being pressed for some reason was unpressed on layout switch.

The reason turns out that wlr_keyboard is out of sync with its xkb_state. E.g. whilst wlr_keyboard::modifiers::locked was set, the analogous field inside `xkb_state` wasn't. It looks like a wlroots bug, but I'm not entirely sure, tell me if I need to report that.

For the time being I worked around that by using the fields inside `wlr_keyboard` *(actually, that even made the code shorter, because I don't need to call xkb serialization functions anymore)*.

Adonai:
Hi! I'm the author of original issue, thanks for taking your time to implement this!

> In my case operations of adding/removing are very rare, whereas traversing happens often, so I need a dynamic array.

We know all the layouts once Sway is initialized/reloaded, can we create an array with that size and pin it to every view?

Hi-Angel:
> We know all the layouts once Sway is initialized/reloaded, can we create an array with that size and pin it to every view?

We actually don't know. New keyboards may be added any time, that's why a dynamic array is used here.

emersion:
>The reason turns out that wlr_keyboard is out of sync with its xkb_state. E.g. whilst wlr_keyboard::modifiers::locked was set, the analogous field inside xkb_state wasn't. It looks like a wlroots bug, but I'm not entirely sure, tell me if I need to report that.
>
>For the time being I worked around that by using the fields inside wlr_keyboard (actually, that even made the code shorter, because I don't need to call xkb serialization functions anymore).

I'm really not a fan of this.

Can you give a piece of code that triggers this maybe-bug?

Hi-Angel:
Sure. It was in [this version](https://github.com/swaywm/sway/commit/92b5de3d317dc80fc7422c46439335a84b1d7d6b#diff-d7aa39a56641abe2c1362ca922ba299bR523) of `sway_keyboard_set_layout` function. In there I'm serializing `xkb_state` to get state of modificators. Then I found that if I pressed and held <kbd>$mod</kbd> key trying to change focused window back and forth, I only could switch focus once, and then <kbd>$mod</kbd> became, like, "unpressed", i.e. I had to release it and press again to swith a window.

In next version of the function [I used modifiers fields from `wlr_keyboard`](https://github.com/swaywm/sway/pull/3155/commits/9c604dd07893b3594119ce3bb3ec6972d267ddfd#diff-d7aa39a56641abe2c1362ca922ba299bR567). It fixed the problem and even resulted in cleaner code, though I guess `xkb_state` isn't supposed to be out of sync with these fields.

I tried debugging the problem for some time, IIRC I couldn't reproduce it when running sway under X11 *(it's reproducible from a tty)*. But at some point I moved on for various reasons.

Hi-Angel:
Btw, if you want to reproduce it with the code from this PR, make sure you have option `keep_kbd_layout per_window` in your `~/.config/sway/config`, so the function `sway_keyboard_set_layout()` gets ran when a window is switched.

Hi-Angel:
Any updates on this PR? It's already sway RC1, would be nice to have this feature in before the first release.

ddevault:
rc1 is feature frozen, so this won't get in regardless.

NEOhidra:
v1 is out, is there anything else blocking this PR from being merged?

sway/input/keyboard.c
```diff
@@ -517,5 +560,18 @@ void sway_keyboard_destroy(struct sway_keyboard *keyboard) {
...
+	root_for_each_container(rm_kbd_from_view, wlr_keyboard);
 	free(keyboard);
 }
+
+void sway_keyboard_set_layout(struct wlr_keyboard* kbd, xkb_layout_index_t layout) {
+	wlr_keyboard_notify_modifiers(kbd,
+								  kbd->modifiers.depressed,
+								  kbd->modifiers.latched,
+								  kbd->modifiers.locked,
+								  layout);
```
ddevault: Style: we don't align args like this.

sway/input/keyboard.c
```diff
@@ -505,9 +520,37 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
...
+			list_del(kbd_layouts, i);
+			return;
+		}
+	}
+	sway_assert(false,
+				"layout_per_kbd not found in sway_view for removal");
+}
+
+void sway_keyboard_destroy(struct sway_keyboard *keyboard,
+						   struct wlr_keyboard *wlr_keyboard) {
```
ddevault: here too, and throughout

sway/input/seat.c
```diff
@@ -79,6 +80,13 @@ static void seat_send_activate(struct sway_node *node, struct sway_seat *seat) {
 	}
 }
 
+static void set_kbd_layouts(const struct sway_view* view) {
+	for(int i = 0; i < view->kbd_layouts->length; ++i) {
```
ddevault: style: space between for and (

sway/input/seat.c
```diff
@@ -79,6 +80,13 @@ static void seat_send_activate(struct sway_node *node, struct sway_seat *seat) {
...
+		struct layout_per_kbd* lpk = view->kbd_layouts->items[i];
```
ddevault: style: * next to the name, not the type

sway/tree/view.c
```diff
@@ -30,13 +31,42 @@
...
 	view->impl = impl;
 	view->executed_criteria = create_list();
+	view->kbd_layouts = create_list();
 	view->allow_request_urgent = true;
 	wl_signal_init(&view->events.unmap);
+	return view_setup_kbd_layouts(view);
+}
+
+static void free_kbd_layouts(list_t* kbd_layouts) {
+	for (int i = 0; i < kbd_layouts->length; ++i)
```
ddevault: style: braces are mandatory

ddevault:
Took a look, needs some work. I'm also not sure about the approach in general here, it seems kind of backwards. This patch makes the global layout the exception rather than the norm. This isn't a particularly important difference to the internals, but the commands for interacting with this feature should be reversed. I'd prefer something like this:

```
for_window [...] xkb_layout en:gb
bindsym $mod+g xkb_layout en:gb
```

Rinse and repeat for all other `xkb_*` commands. Would only be valid if a window is focused or some criteria is met.

Hi-Angel:
Style comments are addressed.

About man page: sure I can add docs *(so far the usage is documented in the commit)* if you tell me where the docs are. I searched over the repo for `man`, `ascii`, `txt`, to no avail. Googling gives me [this commit](https://github.com/swaywm/sway/commit/4b64e481a807bd97b12c29a7ce5044e5de5fc21d), but files being added there don't exist anymore.

> This patch makes the global layout the exception rather than the norm

Why? To me it seems, everything just as it was before, barring the new option. When option wasn't modified, layout is global *(as before)*. Otherwise, there's an option to set which one you want explicitly.

> This isn't a particularly important difference to the internals, but the commands for interacting with this feature should be reversed. I'd prefer something like this:

```
for_window [...] xkb_layout en:gb
bindsym $mod+g xkb_layout en:gb
```

I don't really understand. Do you want a `for_window` command that would set a particular layout for specified windows?

Hi-Angel:
> About man page: sure I can add docs _(so far the usage is documented in the commit)_ if you tell me where the docs are. I searched over the repo for `man`, `ascii`, `txt`, to no avail. Googling gives me [this commit](https://github.com/swaywm/sway/commit/4b64e481a807bd97b12c29a7ce5044e5de5fc21d), but files being added there don't exist anymore. Also, my ½-year old version of sway-git doesn't seem to have man pages.

Nvm, I figured to look through git-log for `man` word.

All comments are addressed.

ddevault:
>Do you want a for_window command that would set a particular layout for specified windows?

Right.

Hi-Angel:
> > Do you want a for_window command that would set a particular layout for specified windows?
> 
> Right.

I see. Well, this is a unusual feature in general, besides it would need more work. Since current pull request addresses existing issue *(and does it in rather popular manner — the "either global or per-window" option is what, I think, uses pretty much every system out there)*, I would suggest that this functional would be merged, so people could start using it.

Otherwise, if it's mandatory, I'm afraid I don't know when I find time to get back to these changes. I'm right now contributing to lots of different projects that I use every day *(Emacs, Geary, Mesa, xserver, libinput…)*; whereas Sway I haven't used, for different reasons, for some time. Besides, I'm subjectively dissatisfied with Sway code *(it's a modern project written in C — I'd definitely prefer Rust — it has tons of global variables, it stores internal state in strings that require strcmp every time instead of using enums…)*. So, I'm not sure when I find both time and motivation to add the new functional. *(although, I think, when I start using it again, it may give me motivation, because I like the project either way)*.

ddevault:
*alright, could do without the snide commentary, thanks*

progandy:
> Took a look, needs some work. I'm also not sure about the approach in general here, it seems kind of backwards. This patch makes the global layout the exception rather than the norm. This isn't a particularly important difference to the internals, but the commands for interacting with this feature should be reversed. I'd prefer something like this:

I believe this patch set does something different than what you associate with keyboard layout. 

The patches by Hi-Angel store the **layout-index** for each window when you have defined multiple layouts for a keyboard and switch between them. The layouts associated with the keyboard are not changed at all.
Your proposed ```for_window ... xkb_layout en:gb``` seems to do something different, and changes the layouts configured for each window, loading a new keymap for each window.

ddevault:
Oh, I understand now. Still, the design seems backwards to me.

progandy:
So the change you desire would be more like

    for_window [...] keep_kbd_layout yes|no

instead of a global ```keep_kbd_layout per_window|global```?

ddevault:
I'd like to be able to set the default layout based on criteria (e.g. with for_window) and also allow you to pin a window to a specific layout at runtime. Rather than just magically remembering which layout you were using, I think that's more explicit. Thoughts?

gdamjan:
> Rather than just magically remembering which layout you were using, I think that's more explicit.

As a two layout user, latin and cyrillic, I don't see the usefulness to configure static per-window layout. I use both latin and cyrllic in all applications, but when I type something in Firefox and switch to my irc app, and back to firefox I'd like it to magically remember my layout I was using last. Most often it's the layout I want to continue using.

ddevault:
Alright, fair enough. This patch looks fine.

sway/tree/view.c
```diff
@@ -30,13 +31,43 @@
...
 	view->type = type;
 	view->impl = impl;
 	view->executed_criteria = create_list();
+	view->kbd_layouts = create_list();
 	view->allow_request_urgent = true;
 	wl_signal_init(&view->events.unmap);
+	return view_setup_kbd_layouts(view);
+}
+
+static void free_kbd_layouts(list_t *kbd_layouts) {
```
RyanDwyer: You can use `list_free_items_and_destroy` instead of this function.

include/sway/config.h
```diff
@@ -41,6 +41,11 @@ enum binding_flags {
 	BINDING_TITLEBAR=16, // mouse only; trigger on container titlebar
 };
 
+enum keep_kbd_layout {
+	LAYOUT_GLOBAL,
+	LAYOUT_PER_WINDOW
```
RyanDwyer: Can you please add an additional prefix such as `KEYBOARD_LAYOUT_GLOBAL`? Just to eliminate any possibility of people getting confused between keyboard layouts and container layouts.

sway/commands/keep_kbd_layout.c
```diff
@@ -0,0 +1,25 @@
...
+
+	if (!strcmp(argv[0], STR_LAYOUT_GLOBAL)) {
+		config->keep_layout = LAYOUT_GLOBAL;
+		return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+	} else if (!strcmp(argv[0], STR_LAYOUT_PER_WINDOW)) {
+		config->keep_layout = LAYOUT_PER_WINDOW;
+		return cmd_results_new(CMD_SUCCESS, NULL, NULL);
+	} else {
+		return cmd_results_new(CMD_FAILURE, "keep_kbd_layout",
+				"Unknown value for keep_kbd_layout");
```
RyanDwyer: Please change the message to something like `Expected 'keep_kbd_layout global|per_window'` so the user knows how to fix their mistake.

sway/input/keyboard.c
```diff
@@ -449,6 +450,20 @@ static void handle_keyboard_modifiers(struct wl_listener *listener,
 	determine_bar_visibility(modifiers);
 }
 
+void add_kbd_to_view(struct sway_container *con, void *data) {
+	if (con->node.type != N_CONTAINER) {
+		return;
+	}
+
+	struct layout_per_kbd *lpk = malloc(sizeof(struct layout_per_kbd));
```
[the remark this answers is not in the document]
Hi-Angel: Why?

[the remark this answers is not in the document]
emersion: It's just too easy to forget to initialize one field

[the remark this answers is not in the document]
RyanDwyer: It protects against a situation where we add a property to the struct later, forget to initialise it to null and therefore have invalid pointers running around. Using `calloc` for structs is a good habit to get into, and one that we're applying across the whole project.

`malloc` is fine for strings though.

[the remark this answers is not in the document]
Hi-Angel: > It's just too easy to forget to initialize one field

Okay, you're maintainers, so I'll replace this one *(I'm gonna address comments a bit later, when I'll be back home)*.

However I want to note that I strictly against C-style initialization of structs to "all zeroes". All it does is hides bugs — exactly when someone forgets to initialize a field. You need to understand that "all zeroes" is not an initialized struct, because all it does is sets it to an arbitrary value. It will only be initilized, when you assign the fields explicitly — with designated initializers or otherwise.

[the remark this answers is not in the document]
emersion: Disagree. Zero is a special value for pointers as well as for almost all values.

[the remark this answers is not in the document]
Hi-Angel: > Disagree. Zero is a special value for pointers as well as for almost all values.

Not for file descriptors, and not necessarily for enums as well as not necessarily for number types.

sway/input/keyboard.c
```diff
@@ -572,9 +587,37 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
...
+			list_del(kbd_layouts, i);
+			return;
+		}
+	}
+	sway_assert(false,
+			"layout_per_kbd not found in sway_view for removal");
+}
+
+void sway_keyboard_destroy(struct sway_keyboard *keyboard,
+						   struct wlr_keyboard *wlr_keyboard) {
```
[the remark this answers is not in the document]
Hi-Angel: Compiler says "error: struct sway_input_device has no member named wlr_keyboard".

Hi-Angel:
@RyanDwyer everything is addressed except the comment about `Having to pass a second struct to sway_keyboard_destroy`, I left a comment there.

sway/input/keyboard.c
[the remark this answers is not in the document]
emersion: `keyboard->seat_device->input_device->wlr_device->keyboard`

Hi-Angel:
upd: renamed `keep_kbd_layout` to `keep_keyboard_layout` *(sorry, missed that comment; or rather, I've seen it, but couldn't find it)*.

sway/input/keyboard.c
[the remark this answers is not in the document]
Hi-Angel: Thanks, done!

Hi-Angel:
Thanks @emersion, now everything should be addressed.

sway/input/keyboard.c
```diff
@@ -572,6 +588,33 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
...
+	struct layout_per_kbd lpk = {
+		.kbd = wlr_device->keyboard,
+		.layout = keyboard->default_kbd_layout
+	};
+	root_for_each_container(add_kbd_to_view, &lpk);
+
+}
+
+static void rm_kbd_from_view(struct sway_container *con, void *data) {
+	if (con->node.type != N_CONTAINER) {
```
RyanDwyer: `if (!con->view)`

sway/input/keyboard.c
```diff
@@ -572,6 +588,33 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
...
+		return;
+	}
+
+	list_t *kbd_layouts = con->node.sway_container->view->kbd_layouts;
```
RyanDwyer: `con->view->kbd_layouts`

include/sway/config.h
```diff
@@ -41,6 +41,11 @@ enum binding_flags {
 	BINDING_TITLEBAR=16, // mouse only; trigger on container titlebar
 };
 
+enum keep_keyboard_layout {
+	KEYBOARD_LAYOUT_GLOBAL,
+	KEYBOARD_PER_WINDOW
```
RyanDwyer: I meant `KEYBOARD_LAYOUT_PER_WINDOW` as well. Sorry for not being clear.

Hi-Angel:
Thanks, done.

sway/input/keyboard.c
```diff
@@ -449,6 +450,21 @@ static void handle_keyboard_modifiers(struct wl_listener *listener,
 	determine_bar_visibility(modifiers);
 }
 
+static
+void add_kbd_to_view(struct sway_container *con, void *data) {
```
emersion: Style: no newline after `static`

include/sway/config.h
```diff
@@ -41,6 +41,11 @@ enum binding_flags {
 	BINDING_TITLEBAR=16, // mouse only; trigger on container titlebar
 };
 
+enum keep_keyboard_layout {
```
emersion: This should have the `sway_` prefix

include/sway/input/keyboard.h
```diff
@@ -63,14 +63,25 @@ struct sway_keyboard {
 
 	struct wl_event_source *key_repeat_source;
 	struct sway_binding *repeat_binding;
+	xkb_layout_index_t default_kbd_layout;
 };
 
+struct layout_per_kbd {
```
emersion: Ditto

include/sway/config.h
```diff
@@ -521,6 +526,7 @@ struct sway_config {
 			char **argv;
 		} leftovers;
 	} handler_context;
+	enum keep_keyboard_layout keep_layout;
```
emersion: I'd prefer this field to be named `keep_keyboard_layout`. There are other types of layouts, e.g. output layouts.

sway/commands/keep_keyboard_layout.c
```diff
@@ -0,0 +1,25 @@
+#define _XOPEN_SOURCE 700
```
emersion: This is probably not needed.

sway/input/keyboard.c
```diff
@@ -572,6 +588,33 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
...
 	keyboard->keyboard_modifiers.notify = handle_keyboard_modifiers;
+
+	keyboard->default_kbd_layout =
+		sway_keyboard_get_layout(wlr_device->keyboard->xkb_state);
+	struct layout_per_kbd lpk = {
+		.kbd = wlr_device->keyboard,
+		.layout = keyboard->default_kbd_layout
+	};
+	root_for_each_container(add_kbd_to_view, &lpk);
+
```
emersion: Style: extra newline here

sway/input/keyboard.c
```diff
@@ -585,5 +628,19 @@ void sway_keyboard_destroy(struct sway_keyboard *keyboard) {
...
+void sway_keyboard_set_layout(struct wlr_keyboard *kbd, xkb_layout_index_t layout) {
+	wlr_keyboard_notify_modifiers(kbd,
+			kbd->modifiers.depressed,
+			kbd->modifiers.latched,
+			kbd->modifiers.locked,
+			layout);
+}
+
+xkb_layout_index_t sway_keyboard_get_layout(struct xkb_state *kbd_state) {
+	return xkb_state_serialize_layout(kbd_state, XKB_STATE_LAYOUT_LOCKED);
```
emersion: This should probably be `XKB_STATE_LAYOUT_EFFECTIVE`

sway/input/keyboard.c
```diff
@@ -449,6 +450,21 @@ static void handle_keyboard_modifiers(struct wl_listener *listener,
...
+	if (!con->view) {
+		return;
+	}
+
+	struct layout_per_kbd *lpk = calloc(1, sizeof(struct layout_per_kbd));
+	if (!sway_assert(lpk, "Failed to allocate layout_per_kbd")) {
+		return;
+	}
+
+	*lpk = *(struct layout_per_kbd*)data;
```
emersion: Style: missing space between type and `*`

sway/input/keyboard.c
```diff
@@ -572,6 +588,33 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
...
+
+static void rm_kbd_from_view(struct sway_container *con, void *data) {
+	if (!con->view) {
+		return;
+	}
+
+	list_t *kbd_layouts = con->view->kbd_layouts;
+	for (int i = 0; i < kbd_layouts->length; ++i) {
+		struct layout_per_kbd *lpk = kbd_layouts->items[i];
+		if (lpk->kbd == (struct wlr_keyboard*)data) {
```
emersion: Style: missing space between type and `*`

Hi-Angel:
@emersion thanks, done.

sway/commands/keep_keyboard_layout.c
Hi-Angel: Oh, wait a second, I this probably didn't get into the diff because I had old file with "kbd" naming opened. Let me redo this in a sec.

Hi-Angel: Thanks, done

sway/input/seat.c
```diff
@@ -751,13 +761,28 @@ static void send_unfocus(struct sway_container *con, void *data) {
...
 	sway_cursor_constrain(seat->cursor, NULL);
 	wlr_seat_keyboard_clear_focus(seat->wlr_seat);
 	if (node->type == N_WORKSPACE) {
 		workspace_for_each_container(node->sway_workspace, send_unfocus, seat);
 	} else {
+		struct sway_view *view = node->type == N_CONTAINER ?
+			node->sway_container->view : NULL;
+
+		if (view && seat_is_input_allowed(seat, view->surface) &&
+			config->keep_keyboard_layout == KEYBOARD_LAYOUT_PER_WINDOW) {
```
emersion: Style: two indents for `if` continuation lines, so that we can tell the difference between the condition and the body.

sway/commands/keep_keyboard_layout.c
```diff
@@ -0,0 +1,24 @@
+#include "sway/config.h"
+#include "sway/commands.h"
+#include "sway/input/input-manager.h"
+#include "log.h"
+
+struct cmd_results *cmd_keep_keyboard_layout(int argc, char **argv) {
```
emersion: I just realized that it would probably be better if this command was a seat command instead of a global one.

Hi-Angel: @emersion thanks, can you describe what do I need to do to convert this command to be per-seat one? I'm not familiar with Sway codebase *(especially so half a year later since last time I worked with it)*.

emersion: It should be pretty simple: remove it from the list of global commands, add it to the list of seat commands (see `commands/seat.c`), move this file to the `commands/seat` dir and rename it to have a `seat_` prefix. Then you can use it like this:

```
seat * keep_keyboard_layout per_window
```

You'll need to move the config field to the seat config too, so that different seats can have different settings.

Hi-Angel: Ok, almost done, but how do I inside the

```c
static void seat_send_focus(struct sway_node *node, struct sway_seat *seat) {
```

the current `seat_config`. Do I have to add another argument, or can I somehow infer it from these two? `sway_seat` doesn't seem to have a pointer to its original `sway_config`.

RedSoxFan:
```c
struct seat_config *seat_config = seat_get_config(seat);  // May be NULL
```

Hi-Angel: So, I converted the command to per-seat one *(didn't update docs yet, wanted to test first)*, but when I execute `swaymsg "seat * keep_keyboard_layout per_window"` it doesn't work for me. In case somebody wants to look at it and it's something trivial, [I pushed the changes to this branch](https://github.com/Hi-Angel/sway/commits/kbd-layout-per-window-seat).Myself I don't know when I gonna debug this.

gdamjan: > I just realized that it would probably be better if this command was a seat command instead of a global one.

what would be the benefit of that?

sway/input/seat.c
Hi-Angel: done

sway/commands/keep_keyboard_layout.c
emersion: >what would be the benefit of that?

Better multi-seat support. Two seats of the same machine can have different settings.

NEOhidra: This feature is ready for a few months now. People want per window keyboard kbd layout.
I use 3 seats and exactly the same layouts - layout per window, not seat.
This PR is exactly about it - per window. Why not merge it?
Is there anyone using different kbd layouts per seat?

Hi-Angel: @emersion thanks, so, I debugged it, and the problem turns out that [in this command handling function](https://github.com/Hi-Angel/sway/commit/20dd6077d9ad59de0a92a2eb4bbd9f09771833e6#diff-e2b76690ea4454a5495b846f716883deR6) the `seat_config` is not the one that is used when a user is switching between windows. Moreover: every time I execute `swaymsg "seat * keep_keyboard_layout per_window"`, the address of `config->handler_context.seat_config` in this function is different. It's like it's copied somewhere, though unlikely by me. Any ideas?

Hi-Angel: Heh, I tried to use this as an opportunity to finally try reverse-execution, but when I stop `sway` in gdb, I'm getting spammed by `Key repeat discarded, Wayland compositor doesn't seem to be processing events fast enough!` until I continue. Oh, well.

emersion: >thanks, so, I debugged it, and the problem turns out that in this command handling function the seat_config is not the one that is used when a user is switching between windows.

Yeah, configs are copied when applied iirc. But that shouldn't matter right? Since `keep_keyboard_layout` should be copied over?

Hi-Angel: > Yeah, configs are copied when applied iirc. But that shouldn't matter right? Since `keep_keyboard_layout` should be copied over?

Thanks, right, but apparently I need to also modify the copying code to make this field copied too. Where is it copied?

emersion: In `merge_seat_config`

Hi-Angel: Thanks, I looked at it. Apparently to make this work I need to introduce an additional default enum member `KEYBOARD_LAYOUT_UNDEFINED`, because this function only copies over fields that were modified, and I need to figure out whether that happened.

I think instead of copying Sway should just apply functions that under `commands/seat/` dir directly to configs requested. In particular this would allow to get rid of "undefined" values.

Hi-Angel: …and in general this would reduce maintainance because situations like mine just won't happen.

RedSoxFan: `merge_seat_config` is needed for proper layering of wildcard and named seat configs. The seat configs are handled the way they are for a reason and the complexity wouldn't be there if it wasn't necessary.

Hi-Angel: @RedSoxFan I don't see why the wildcard and named seat resolution couldn't be done just before applying the command, thus not doing any copying.

RedSoxFan: Ignoring that the user can interlace wildcard and named seat commands, you would still need some way to merge those two configs together. Applying the wildcard config and then the named config will not result in desired behavior (especially if you are overriding the wildcard config with defaults).

Hi-Angel: Correct me if I'm wrong, but what a user does is two things: 1. defines seat configs that it wants to apply a command to, and 2. provides name of the field and the exact value to set it to.

I don't really understand you, but even if you merge 2 configs where the value was set to what user wanted — you will get back a config with the value that a user have set.

Hi-Angel: @emersion done. I also rebased on master. Note, there's a gotcha, I don't know if it's expected behavior: when I put in my config a line

```haskell
seat * keep_keyboard_layout per_window
```

it doesn't work for me. However this line does:

```haskell
exec swaymsg "seat * keep_keyboard_layout per_window"
```

emersion: >I don't really understand you, but even if you merge 2 configs where the field was set to what user wanted — you will get back a config with the value that a user have set.

Merging `seat * option1 true` and `seat * option2 true` would result in `{"option1": false, "option2": true}` if merging the second one on top of the first one without "undefined" values.

There are more related issues, e.g. `seat * option1 true` + `seat seat1 option1 false`.

>when I put in my config a line it doesn't work for me

That's unexpected, other seat commands work in the config file.

Hi-Angel: > > I don't really understand you, but even if you merge 2 configs where the field was set to what user wanted — you will get back a config with the value that a user have set.
>
> Merging `seat * option1 true` and `seat * option2 true` would result in `{"option1": false, "option2": true}` if merging the second one on top of the first one without "undefined" values.

But why would you merge them if these are separate commands? So you only get to merge `option1` with `option1` and `option2` with `option2`.

Either way, I think we just gone to a speculation, point of my text you quoted was that I don't understand why would they want to merge configs while executing `seat * option1 true` without doing the copy. You only merge the copy with existing configs, not multiple existing configs, right? So there should be nothing stopping one from executing command-handler functions directly on the existing configs.

> There are more related issues, e.g. `seat * option1 true` + `seat seat1 option1 false`.

These still end up resolved to something exact, i.e. still can be resolved before applying the command handlers, and the handlers can be applied to configs we resolved to.

> > when I put in my config a line it doesn't work for me
>
> That's unexpected, other seat commands work in the config file.

Any ideas?

Hi-Angel: @emersion I figured the problem: I expected that the `seat_config` "copy" is created by `calloc()` as everything else in sway. Unfortunately instead it is created by `new_seat_config()` *(internally it still uses `calloc`, but that's not the point)*. So what happens is that I initialize `keep_keyboard_layout` in this function to a "valid" value; and then when upon `swaymsg` call or whatever the config copy is created *(to be acted upon and then to get merged with existing configs)*, it already has *valid* value.

Then the question is: where in Sway can I initialize default values of `seat_config`?

And while on it: I wonder, if it would be acceptable to leave a comment in `new_seat_config()` that it should not be used for assigning default values? So that future contributors won't get stuck on same problem with seat configs as I do.

emersion: >where in Sway can I initialize default values of seat_config?

Sorry, I don't get it…

Hi-Angel: > Sorry, I don't get it…

Let me reword it: upon creation of the config that Sway uses for the rest of its lifetime, I need to initialize a field to a default value. The config is created with `new_seat_config()`

The problem is: the "config copy" that this whole thread is about is also created with `new_seat_config()`, however the copy should have different default value, `UndefinedKbdLayout` *(or something similar, I'm talking from my memory)*.

My question is: where in code should I initialize a field of the first config to its default value so that this value won't get carried over to the "config copies" that sway creates.

emersion: Why is this an issue that the "undefined keyboard layout" value gets carried over?

Hi-Angel: > Why is it an issue that the "undefined keyboard layout" value gets carried over?

To be more explicit, "gets carried over" the value that I'd assign inside `new_seat_config()`.

If this is the "undefined keyboard layout", then if a user didn't explicitly set the option I'm implementing, on Sway startup the main config gonna have the "undefined keyboard layout" value which is a bug.

If this is a valid value, then, due to how Sway deals with config parsing *(i.e. creating copies and stuff)* I will have no way to differentiate "default value" from the one that user has explicitly set.

emersion: Other options define a "not set" value (e.g. -1 if it's an int, or zero if it's an enum, etc). Set the field to this "not set" value in `new_seat_config`. Use this "not set" value to decide whether the existing value should be overridden when merging two configs. When actually reading the config value, handle "not set" as the value you want to be the default.

Is there an issue with that?

Hi-Angel: >When actually reading the config value, handle "not set" as the value you want to be the default.
>
>Is there an issue with that?

Per my understanding, if one would set some irrelevant option with `swaymsg` command, then upon config merge mine option would have the -1. Now, if I treat this as "set option to a default value", this gonna override the value that a user could've previously set, which is a bug.

emersion: No: in the merge function you override the current value only if it's != -1.

Hi-Angel: > No: in the merge function you override the current value only if it's != -1.

This is not possible because I can't set it to -1 in `new_seat_config()`. To quote my prev. comment why:

> If this is the "undefined keyboard layout", then if a user didn't explicitly set the option I'm implementing, on Sway startup the main config gonna have the "undefined keyboard layout" value which is a bug.

emersion: >the main config gonna have the "undefined keyboard layout" value which is a bug.

no it's not

Hi-Angel: > > the main config gonna have the "undefined keyboard layout" value which is a bug.
> 
> no it's not

So, I just double-checked: in the latest code it requires one-line change: [replace here](https://github.com/swaywm/sway/pull/3155/commits/b1f665b1c3af6ba7f9923bbf41248909d537d934#diff-12676f2e76e92011b7450ecc7b4bdc1fR30) the `KEYBOARD_LAYOUT_GLOBAL` with `KEYBOARD_LAYOUT_UNDEFINED`.

What happens next is that if I remove from my configuration the line `seat * keep_keyboard_layout per_window`, sway crashes [on this sway_assert](https://github.com/swaywm/sway/pull/3155/commits/b1f665b1c3af6ba7f9923bbf41248909d537d934#diff-57d5973ad58d843753d353531e8cb85eR111), as I expected.

emersion: >sway crashes on this sway_assert, as I expected.

Look, I really appreciate your work on this PR. However I'd like to optimize the time spent on the review. Both reviewer and contributor time is scarce, and I'm not a fan of endless discussions that go nowhere.

When I said "no it's not" I wasn't kidding: the code that reads values from the config does handle the "unset" case as if it was whatever-default-you-want (here, global). In this particular case your code checks for "per-window" so it should just work. But you're probably aware of this, so:

I can understand that you don't like how it's done, and I'm open to discuss them in separate issues. However it really shouldn't happen in this PR. The PR should just implement what it says in the title and blend with the existing code -- not rework the whole existing framework.

Let me know if something's unclear.

Hi-Angel: > > sway crashes on this sway_assert, as I expected.
> 
> Look, I really appreciate your work on this PR. However I'd like to optimize the time spent on the review. Both reviewer and contributor time is scarce, and I'm not a fan of endless discussions that go nowhere.

I'm addressing your review. I'm assuming you know the code you're talking about, hence I ask you about how should I address your comments.

If you'd left the comments half a year ago when this PR was publicly worked on and finished, I'd probably did this research on my own. But right now I'm not using Sway, and to address your review I need collaboration from you.

> When I said "no it's not" I wasn't kidding: the code that reads values from the config does handle the "unset" case as if it was whatever-default-you-want (here, global). In this particular case your code checks for "per-window" so it should just work. But you're probably aware of this, so:
> 
> I can understand that you don't like how it's done, and I'm open to discuss that in separate issues. However it really shouldn't happen in this PR. The PR should just implement what it says in the title and blend with the existing code -- not rework the whole existing framework.
> 
> Let me know if something's unclear.

I don't know why are you thinking this, but I did **exactly** as you asked in prev. comment *(except I'm using a "undefined constant" instead of -1)*. Didn't I?

And yes, I also don't like that this discussion goes so long: this is the reason I linked you to exact code lines in my prev. comment, hoping that if you think that it should work whereas it clearly doesn't, you could maybe see in code why it doesn't.

I'm not trying to rework anything, at this point I don't care about sway code, I'm sincerely just trying to pass the review.

emersion: >I don't know why are you thinking this, but I did exactly as you asked in prev. comment (except I'm using a "undefined constant" instead of -1). Didn't I?

Okay. Then remove the assert you've added, since it conflicts with:

>When I said "no it's not" I wasn't kidding: the code that reads values from the config does handle the "unset" case as if it was whatever-default-you-want (here, global). In this particular case your code checks for "per-window" so it should just work.

Hi-Angel: > > I don't know why are you thinking this, but I did exactly as you asked in prev. comment (except I'm using a "undefined constant" instead of -1). Didn't I?
> 
> Okay. Then remove the assert you've added, since it conflicts with:
> 
> > When I said "no it's not" I wasn't kidding: the code that reads values from the config does handle the "unset" case as if it was whatever-default-you-want (here, global). In this particular case your code checks for "per-window" so it should just work.

As you wish. Done [in the latest update](https://github.com/swaywm/sway/compare/b1f665b1c3af6ba7f9923bbf41248909d537d934..cc02a58177c8f7271cbc1c1b741b43919454993d).

Hi-Angel:
**upd**: conflicts with latest upstream resolved, rebased.

sway/commands/keep_keyboard_layout.c
Hi-Angel: Ah, wait, I forgot to add the update to assign undefined, let me do in a sec.

Hi-Angel: Done

Hi-Angel: @emersion I think everything is addressed. It works too, I tested. Any other comments?

emersion:
Hmm, [a comment](https://github.com/swaywm/sway/issues/2361#issuecomment-498238250) made me reconsider this PR: I'm not sure implementing it in the compositor is the right approach. A script might be better.

Hi-Angel:
> Hmm, [a comment](https://github.com/swaywm/sway/issues/2361#issuecomment-498238250) made me reconsider this PR: I'm not sure implementing it in the compositor is the right approach. A script might be better.

Well, thinking about `kbdd` on X11, theoretically it could be done. But what's the difference to having it in the core? The one I see is that you're basically asking someone to go and re-make all bunch of work that I did, whereas this PR provides a working implementation.

eternal-sorrow:
Why not both? Good to have something like this in the compositor core and also have all the IPC functions to implement similar functionality as separate application.

NEOhidra:
Merge it please!
It was ready once - then it was requested to be reworked.
Once refactored it was not merged for a few days.

If such feature is not welcomed just say not gonna happen and close this PR.

Omrigan:
Indeed, this feature would be a nice addition to sway. Lack of it is the
only thing that keeps me with i3 :)

By the way, I am not quite sure that it would be possible to implement as a
script with switching window and layout atomically. This would require a
two-phase IPC and may slow the switching process. But I might be totally
mistaking as I am not familiar with sway codebase.

emersion:
This PR introduces more complexity in sway's input code, which is already quite complex. A separate script would allow users to have more fine-grained control and would be less maintenance.

Hi-Angel:
> This PR introduces more complexity in sway's input code, which is already quite complex. A separate script would allow users to have more fine-grained control and would be less maintenance.

This PR introduces a simple option, one other any compositor has, and a few switches to work with it. This seems odd to me that such a change suddenly being called an "increase in complexity of Sway". Maybe the codebase needs something to make adding an option less complex?

emersion:
>that any other compositor has

On i3 you also need a daemon if I'm not mistaken.

>This seems odd to me that such a change suddenly being called an "increase in complexity of Sway". Maybe the codebase needs something to make adding an option less complex?

The reason why this change is taking so long to be reviewed is because it _is_ complex. And no, the new command isn't the part that is complex.

ddevault:
I'm still on board with merging this patch.

Omrigan:
Is this moving any further?

emersion:
>Is this moving any further?

https://github.com/swaywm/sway/pull/4504

emersion:
Clarification: I'm not reviewing this anymore and I'm still -1 on the approach. Feel free to convince someone else to review this PR.

PR #3161, opened 2018-11-19
[WIP] Don't modeset output to default mode when already set to a mode

tarmack:
When an output already has a mode set then the default mode does not need to be set. This change was triggered by me getting annoyed that my screen would be modeset (giving a flash even when DPMS off) for every background change (I have a periodic background change scripted).

I am not entirely sure if this is the best place to fix this but it does at least fix the issue I encountered where a completely unrelated output command triggers a modeset.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -200,7 +200,7 @@ void apply_output_config(struct output_config *oc, struct sway_output *output) {
 		wlr_log(WLR_DEBUG, "Set %s mode to %dx%d (%f GHz)", oc->name, oc->width,
 			oc->height, oc->refresh_rate);
 		set_mode(wlr_output, oc->width, oc->height, oc->refresh_rate);
-	} else if (!wl_list_empty(&wlr_output->modes)) {
+	} else if (wlr_output->current_mode == NULL && !wl_list_empty(&wlr_output->modes)) {
 		struct wlr_output_mode *mode =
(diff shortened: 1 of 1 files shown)


emersion:
Hmm. Maybe a better check would be `mode == wlr_output->current_mode`?

Maybe wlroots should check this and do nothing if the mode is already set?

tarmack:
This is the fallback that sets the previous/default mode and will only be hit when the mode is not changed explicitly by the user. In such a case I would assume that any (accidental) change to the mode would come as a surprise to the user anyway. For this reason I thought simple boolean logic should suffice.

On top of that it could indeed be argued that wlroots should just not actually set the mode when it is the same as the current mode anyway. However, that is only a good idea if re-setting the same mode does not ever make sense. I could think of things like the hardware and internal state getting out of sync (no idea if there is any real chance of this, though).

If setting the mode to the same mode as the current mode does really not ever make sense, then fixing this in wlroots is probably better. Can someone with more knowledge of the details here help me make the call?

tarmack:
Currently life is occupying my time so I don't have time to investigate further, but I found that this patch is not enough. The DPMS logic in the same function seems to trigger a flash when the background is changed when the screen is "DPMS off".
I will have time again soon and will improve the patch, please don't merge yet.

RedSoxFan:
Superseded by swaywm/wlroots#1545 (drm only) and then swaywm/wlroots#1554 (all backends)

PR #3168, opened 2018-11-21
Fix missing IPC scratchpad replies/move events

### LINKED ISSUE #3159: Missing IPC scratchpad replies/move events
I want to implement a module for a custom status bar which shows information about scratchpad containers.

For this I need a way to be notified when a container is added/removed to/from scratchpad.

My proposal would be to add a new IPC_EVENT_SCRATCHPAD event type and new events for each root_scratchpad_{add_container,remove_container,show,hide} method. This is similar to workspace events.

If this is acceptable I'll submit a PR.

Any feedback is appreciated.

mihaicmn:
This matches i3's behavior:
- triggers window::move events when containers are moved to scratchpad or when scratchpad is shown
- adds all hidden scratchpad containers to get_tree reply 

Fixes #3159

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -15,6 +15,9 @@
 #include <xkbcommon/xkbcommon.h>
 #include "wlr-layer-shell-unstable-v1-protocol.h"
 
+static const int i3_output_id = INT32_MAX;
+static const int i3_scratch_id = INT32_MAX - 1;
+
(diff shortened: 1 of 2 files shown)


mihaicmn:
> Can you please add window::move events for when the container is added to and removed from the scratchpad (see root_scratchpad_add_container and root_scratchpad_remove_container)? 

Done

RedSoxFan:
Thanks!

PR #3145, opened 2018-11-17
Adding commands for configuring titlebar borders and padding

fdlamotte:
As talked in issue #3134 titlebar border size and padding are currently hardcoded

This patch proposes two new commands, which permits to adjust the width of titlebars
 - titlebar_border_thickness \<n\>
 - titlebar_padding \<h\> [v]

Example use with swaymsg to just get a 1 px border and minimal space above and below text : 
$ swaymsg titlebar_border_thickness 1
$ swaymsg titlebar_padding 5 1

PR #3169, opened 2018-11-21
[DO NOT MERGE] Implement title alignment

RedSoxFan:
**Since this is going to have conflicts with #3145, do not merge this until after that one is merged. I'll rebase and fix merge conflicts on this one**

Related to #3082

This adds support for `i3 4.16`'s ability to set the title alignment. The command is `title_align left|center|right`. When the title is on the right, marks are moved to the left. Otherwise, they are on the right.

I think this matches i3 behavior for all three alignments, but may have missed an edge case. This should also be working on all transforms and with fractional scaling.


sway/desktop/render.c
```diff
@@ -420,29 +421,40 @@ static void render_titlebar(struct sway_output *output,
 		render_rect(output->wlr_output, output_damage, &box, color);
 	}
 
-	size_t inner_width = width - titlebar_h_padding * 2;
+	int inner_x = x - output_x + titlebar_h_padding;
 	int bg_y = y + titlebar_border_thickness;
+	size_t inner_width = width - titlebar_h_padding * 2;
+
+	int ob_inner_x = round(inner_x * output_scale);
```
emersion: Can we add `// output-buffer-local` comments for those two?

RedSoxFan: Done

PR #3176, opened 2018-11-23
Move swaylock PAM policy to /etc/pam.d.

### LINKED ISSUE #3096: swaylock: PAM doesn't use config file in /usr/local/etc/pam.d.
This is not an issue with sway or swaylock itself, but may be an installation issue on some systems. I have PAM installed on my Fedora 28 system, but it doesn't use the configuration file installed to `/usr/local/etc/pam.d`. When I use swaylock, it will lock me out because PAM isn't aware of this file. Moving it to `/etc/pam.d` fixes this.

I didn't notice this until now because I just removed the Sway (0.15) package, which places the configuration file at `/etc/pam.d`. 

I assume that anybody building master is supposed to make a symlink or a copy to `/etc/pam.d`, but this should be noted somewhere.

c-edw:
Fixes #3096.

### DIFF
--- a/swaylock/meson.build
+++ b/swaylock/meson.build
@@ -43,13 +43,13 @@ executable('swaylock',
 if is_freebsd
 	install_data(
 		'pam/swaylock.freebsd',
-		install_dir: sysconfdir + '/pam.d/',
+		install_dir: '/etc/pam.d/',
 		rename: 'swaylock'
 	)
 else
 	install_data(
 		'pam/swaylock.linux',
-		install_dir: sysconfdir + '/pam.d/',
+		install_dir: '/etc/pam.d/',
 		rename: 'swaylock'
 	)
 endif


ascent12:
I think there should be a meson option to override this.

progandy:
That looks like the option was ```sysconfdir```? Of course that moves also ```sway/config``` and ```sway/security.d``` to the same location, but why would you want that in a different location than ```pam.d```?

c-edw:
As mentioned on the related issue, I started getting locked out of my computer because PAM couldn't see the policy file. It doesn't look in `/usr/local/etc/pam.d`. One way or another, the policy file needs to be present in `/etc/pam.d`, whether that's by copying it there or symlinking it there.

I am open to any other suggestions.

progandy:
What should be fixed is the special casing of ```prefix == /usr```. Instead of prepending the prefix to ```sysconfdir```, leave the variable alone. Just prepend it while assigning ```-DSYSCONFDIR``` if ```sysconfdir``` is a relative path.

emersion:
Well, PAM isn't the only component you need to configure when you install to `/usr/local`. You also need to add `/usr/local/bin` to your PATH, `/usr/local/lib` to your `ldconfig` and `/usr/local/{lib,share}/pkgconfig` to you pkg-config path. Maybe we should just add a wiki article about how to do all of this.

c-edw:
@emersion The setup differed a bit for me, which is probably why I struggled initially compiling Sway. I would be willing to write this page and make sure it works between different distros. But - I think it's quite a pain to manually copy the PAM policy to `/etc/pam.d`.

emersion:
If `prefix` is set to `/usr/local`, I don't think we should install anything outside of `/usr/local`. Maybe PAM has the `/usr/local` prefix too, for instance.

A symlink would be better than a copy.

Yeah, the exact path might differ between distros. Not sure we can do anything about this.

progandy:
Meson has build options like

    meson builddir srcdir -Dprefix=/usr
    meson builddir srcdir -Dsysconfdir=/etc

c-edw:
@progandy Would you be able to take up this issue (#3096)? I'm quite confused as to what needs doing, but I can see you already have a good idea.

c-edw:
Closing in favor of #3178. Thanks @progandy.

PR #3178, opened 2018-11-23
Meson: Fix SYSCONFDIR definition

progandy:
```SYSCONFDIR``` was set to the wrong path if ```prefix``` is not ```'/usr'``` and
```sysconfdir``` is an absolute path. Use ```join_paths()``` to fix it.

Also remove the special case for ```prefix``` ```'/usr'```. In that case Meson
already sets ```sysconfdir``` to the absolute path ```'/etc'```, so just using
```join_paths()``` will return the correct value.

    join_paths('/usr/local', 'etc') => '/usr/local/etc'
    join_paths('/usr/local', '/etc') => '/etc'
    join_paths('/usr', '/etc') => '/etc'

This should allow building and installing to '/usr/local' while storing configuration data like pam settings in ```/etc```. (#3096)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -108,13 +108,7 @@ if scdoc.found()
 	endforeach
 endif
 
-# If prefix is '/usr', sysconfdir will be explicitly set to '/etc' by Meson to
-# enforce FHS compliance, so we should look for configs there as well.
-if prefix == '/usr'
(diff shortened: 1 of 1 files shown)


SpeedJack:
See #2855

SpeedJack:
You're right! 👍

emersion:
Thanks!

PR #3179, opened 2018-11-24
fixes pango_markup support with title_format

baloo:
The support for pango_markup was broken in title_format because the
formated title was escaped. I think only the payload should be escaped.

This commit fixes 789a877b379cd35c350610be62b971ae00feb542

The pango_markup looks like:
```
font pango:"DejaVu Sans Mono 10"
for_window [class=".*"] title_format "<span background='#ff0000'>%title</span>"
```

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -824,12 +824,30 @@ struct sway_view *view_from_wlr_surface(struct wlr_surface *wlr_surface) {
 	return NULL;
 }
 
+static char *escape_pango_markup(const char *buffer) {
+	size_t length = escape_markup_text(buffer, NULL);
+	char *escaped_title = calloc(length + 1, sizeof(char));
+	escape_markup_text(buffer, escaped_title);
+	return escaped_title;
+}
+
 static size_t append_prop(char *buffer, const char *value) {
 	if (!value) {
 		return 0;
 	}
-	lenient_strcat(buffer, value);
-	return strlen(value);
+	// if using pango_markup in font, we need to escape all markup char
+	// from values to avoid messing with pango markup
+	if (!config->pango_markup) {
+		char *escaped_value = escape_pango_markup(value);
+
+		lenient_strcat(buffer, escaped_value);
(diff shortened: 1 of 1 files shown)


sway/tree/view.c
```diff
@@ -832,6 +832,25 @@ static size_t append_prop(char *buffer, const char *value) {
...
+	escape_markup_text(buffer, escaped_title);
+	free(buffer);
+	return escaped_title;
+}
+
+static const char *escape_if_pango(const char *buffer) {
+	// if using pango_markup in font, we need to escape all markup char
+	// from values to avoid messing with pango markup
+	if (!config->pango_markup) {
+		return escape_title(strdup(buffer));
```
emersion: This leaks the value returned by `escape_title`. Maybe we should move this to `append_prop`?

baloo: fixed as suggested

sway/tree/view.c
```diff
@@ -824,12 +824,31 @@ struct sway_view *view_from_wlr_surface(struct wlr_surface *wlr_surface) {
 	return NULL;
 }
 
+static char *escape_title(char *buffer) {
```
emersion: We could make `buffer` a `const char*` and stop `free`-ing it in this function.

sway/tree/view.c
```diff
@@ -824,12 +824,31 @@ struct sway_view *view_from_wlr_surface(struct wlr_surface *wlr_surface) {
...
 static size_t append_prop(char *buffer, const char *value) {
 	if (!value) {
 		return 0;
 	}
-	lenient_strcat(buffer, value);
-	return strlen(value);
+	// if using pango_markup in font, we need to escape all markup char
+	// from values to avoid messing with pango markup
+	if (!config->pango_markup) {
+		char * escaped_value = escape_title(strdup(value));
```
emersion: Style: no space between `*` and variable name.

emersion:
Thanks!

PR #3184, opened 2018-11-25
resize set: fix units for floating containers

kupospelov:
The behavior of `resize set <width> <height>` for floating containers has changed recently. According to the documentation, floating containers should be resized in px if the units are omitted, but now ppt are used for both tiled and floating containers.

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -524,7 +524,9 @@ static struct cmd_results *resize_set_floating(struct sway_container *con,
 			grow_width = width->amount - con->width;
 			con->x -= grow_width / 2;
 			con->width = width->amount;
+			break;
 		case RESIZE_UNIT_INVALID:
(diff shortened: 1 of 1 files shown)


sway/commands/resize.c
```diff
@@ -512,34 +512,38 @@ static struct cmd_results *resize_set_floating(struct sway_container *con,
...
+			// Falls through
+		case RESIZE_UNIT_PX:
+		case RESIZE_UNIT_DEFAULT:
 			width->amount = fmax(min_width, fmin(width->amount, max_width));
 			grow_width = width->amount - con->width;
-
 			con->x -= grow_width / 2;
 			con->width = width->amount;
+		case RESIZE_UNIT_INVALID:
+			break;
```
emersion: `assert(false)` would be better

emersion:
Thanks!

PR #3185, opened 2018-11-25
Replace _XOPEN_SOURCE with _POSIX_C_SOURCE

emersion:
And make sure we don't define both in the same source file.


sway/config/bar.c
```diff
@@ -1,5 +1,5 @@
 #define _POSIX_C_SOURCE 200809L
-#define _XOPEN_SOURCE 700
+#define _POSIX_C_SOURCE 200809L
```
progandy: ```_POSIX_C_SOURCE``` is defined twice here.

emersion: Good catch. Since it's defined twice with the same value the compiler didn't complain. Fixed!

PR #3189, opened 2018-11-25
v2: do not create uninitialized sway_keyboard

Hi-Angel:
It's not good to create a uninitialized struct, because it might become
a possible error in future. I actually got hit by this while working on
https://github.com/swaywm/sway/pull/3155

v2: just move initialization inside the keyboard creation, leave the rest in place.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -403,6 +403,7 @@ struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
 	keyboard->key_repeat_source = wl_event_loop_add_timer(server.wl_event_loop,
 			handle_keyboard_repeat, keyboard);
 
+	sway_keyboard_configure(device->keyboard);
 	return keyboard;
 }
 
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -443,10 +443,12 @@ static void seat_configure_keyboard(struct sway_seat *seat,
 		struct sway_seat_device *seat_device) {
 	if (!seat_device->keyboard) {
 		sway_keyboard_create(seat, seat_device);
+	} else {
+		sway_keyboard_configure(seat_device->keyboard);
 	}
+
 	struct wlr_keyboard *wlr_keyboard =
 		seat_device->input_device->wlr_device->keyboard;
-	sway_keyboard_configure(seat_device->keyboard);
 	wlr_seat_set_keyboard(seat->wlr_seat,
 			seat_device->input_device->wlr_device);
 	struct sway_node *focus = seat_get_focus(seat);


emersion:
I don't get this change. In both cases, `sway_keyboard_configure` is already called. Also, there's no uninitialized struct because we use `calloc` everywhere.

Hi-Angel:
calloc can't initialize struct fields correctly. For example, there was some pointer which was set to zero, whilst in properly configured keyboard it should've pointed to something, which led to a crash for me until I figured that just after `sway_keyboard_create()` finished execution, the `sway_keyboard` is still uninitialized.

This change makes sure that if we ever created a `sway_keyboard`, it's properly initialized.

Hi-Angel:
There's also a more general point *(it doesn't immediately apply to this situation though)* is that when you have two separate calls for struct creation and initialization, you need to create a temporary variable, and pass it between two calls. Often such design also makes problems with constifying variables.

Hi-Angel:
> Seems like you're using overloaded terminology here. An uninitialized struct will have fields of indeterminate value. calloc does initialize the struct with all zeroes, so the struct is already initialized.

Ok, maybe I do. Although I'd argue there's little difference between a random value assigned to a field, and a wrong value assigned to the same field. But sure, a field initialized to wrong value is still initialized.

Still, terminology doesn't make this pull request less useful.

> However, it sounds like your problem is you tried to dereference a pointer before it pointed to anything. What were you trying to dereference between calls of `sway_keyboard_create` and `sway_keyboard_configure`?

[This line](https://github.com/swaywm/sway/pull/3155/files#diff-d7aa39a56641abe2c1362ca922ba299bR525):

```c
keyboard->default_kbd_layout =
￼ 		sway_keyboard_get_layout(wlr_device->keyboard->xkb_state);
```

The `xkb_state` is zero between the calls.

remyabel:
Is there any reason you can't move that code after [`wlr_keyboard_set_keymap`](https://github.com/swaywm/sway/blob/c18cd9dd0369452a8b09ac2290d4a0a5b29d6ded/sway/input/keyboard.c#L462)? Adding more calls to `sway_keyboard_configure` may or may not have additional side effects.

Hi-Angel:
> Is there any reason you can't move that code after [`wlr_keyboard_set_keymap`](https://github.com/swaywm/sway/blob/c18cd9dd0369452a8b09ac2290d4a0a5b29d6ded/sway/input/keyboard.c#L462)? Adding more calls to `sway_keyboard_configure` may or may not have additional side effects.

I kind of did, the diff I referred to in my prev. comment have this code after the `wlr_keyboard_set_keymap()`. Or do you mean, like, put it right after the call? *(I put it to the end of the `sway_keyboard_configure()`)*?

Hi-Angel:
@remyabel I think you misunderstood me. The code I referred to works right now, I just showed the line that per my memory resulted in crush back when the line was inside the `sway_keyboard_create` call.

Issue #3190, opened 2018-11-25
Swayidle high CPU usage after pull request #3049


Bryophyllum:
Swayidle sits at 100% CPU usage after a Sway session has been started.
The bug may have been introduced in https://github.com/swaywm/sway/commit/d46d221118f7b4ffeb0fe0fcba296fbbd78a0cbd.

emersion:
How do you start it? I don't have such issues.

Bryophyllum:
I use an alias `sway="export XKB_DEFAULT_LAYOUT=de && export QT_QPA_PLATFORM=wayland-egl && export QT_WAYLAND_DISABLE_WINDOWDECORATION="1" && sway"`

```This is my Swayidle config:
### Idle configuration
exec swayidle \
        timeout 600 "brightnessctl --save --device=acpi_video0 && brightnessctl set 10% --device=acpi_video0" \
        resume 'brightnessctl --restore --device=acpi_video0' \
        timeout 3600 'swaymsg "output * dpms off"' \
        resume 'swaymsg "output * dpms on"' \
```

PR #3191, opened 2018-11-25
swayidle: fix busy loop on writable FD

emersion:
The wl_event_source_fd_update docs say:

> File descriptors are usually writable to begin with, so they do not need to
> be polled for writable until a write actually fails. When a write fails,
> the event mask can be changed to poll for readable and writable, delivering
> a dispatch callback when it is possible to write more. Once all data has
> been written, the mask can be changed to poll only for readable to avoid
> busy-looping on dispatch.

So we should only poll for WL_EVENT_WRITABLE if a write fails. I'm not yet sure
how to do this properly and Weston doesn't do it, so in the meantime I'll just
fix the busy loop. I'll ask them too.

Fixes https://github.com/swaywm/sway/issues/3190

### DIFF
--- a/swayidle/main.c
+++ b/swayidle/main.c
@@ -192,8 +192,7 @@ static void setup_sleep_listener(void) {
 	acquire_sleep_lock();
 
 	struct wl_event_source *source = wl_event_loop_add_fd(state.event_loop,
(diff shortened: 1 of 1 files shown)


ianyfan:
Thanks!

Issue #3197, opened 2018-11-26
Damage issues with Firefox on Wayland menus


JonnyMako:
Some symptoms:
1. Menus are drawn over each other. (attached).
2. Context menu stays on screen even after clicking outside the menu.
![menus](https://user-images.githubusercontent.com/6559121/49039660-0238f900-f186-11e8-98ac-8cd23dc922b8.png)

emersion:
Can you try https://github.com/swaywm/sway/pull/3199?

PR #3199, opened 2018-11-26
Handle destroyed subsurfaces

emersion:
Damage subsurfaces when they are destroyed. Since subsurfaces don't have an
unmap event we need to do that on destroy.

We also don't want to keep a sway_view_child when the wlr_subsurface has been
destroyed.

Fixes https://github.com/swaywm/sway/issues/3197

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -674,6 +674,8 @@ void view_update_size(struct sway_view *view, int width, int height) {
 	container_set_geometry_from_content(view->container);
 }
 
+static const struct sway_view_child_impl subsurface_impl;
+
 static void subsurface_get_root_coords(struct sway_view_child *child,
 		int *root_sx, int *root_sy) {
 	struct wlr_surface *surface = child->surface;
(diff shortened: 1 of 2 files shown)


atomnuker:
Works fine.

sway/tree/view.c
```diff
@@ -689,18 +691,47 @@ static void subsurface_get_root_coords(struct sway_view_child *child,
...
+
+static void subsurface_handle_destroy(struct wl_listener *listener,
+		void *data) {
+	struct sway_subsurface *subsurface =
+		wl_container_of(listener, subsurface, destroy);
+	struct sway_view_child *child = &subsurface->child;
+	if (child->view->container != NULL) {
+		view_child_damage(child, true);
+	}
+	view_child_destroy(child);
```
RyanDwyer: Doesn't this mean we call `view_child_destroy` twice now? `view_child_init` already sets up a listener on the surface's destroy event and calls it from the destroy handler.

sway/tree/view.c
```diff
@@ -752,6 +783,7 @@ static void view_child_handle_surface_unmap(struct wl_listener *listener,
 		void *data) {
 	struct sway_view_child *child =
 		wl_container_of(listener, child, surface_unmap);
+	wlr_log(WLR_DEBUG, "unmap");
```
RyanDwyer: Surely you didn't mean to commit this?

emersion: Derp

sway/tree/view.c
emersion: We have two objects: the `wl_surface` itself and the `wl_subsurface`. This listener is listening on the subsurface destroy event.

Destructors remove all listeners, so we should be fine if one is called before the other.

jnpkrn:
Works, thanks.

RyanDwyer:
Thanks!

PR #3202, opened 2018-11-28
Introduce a way to show config warnings in swaynag

### LINKED ISSUE #3138: Show a swaynag message when two bindings have the same key combination
Would be nice because it's easy to add a new binding without realizing it's already bound. Need to check if `--locked` is the same too.

RedSoxFan:
Closes #3138

Adds the function `config_add_swaynag_warning(char *fmt, ...)` so that handlers can add warnings to the swaynag config log in a uniform way. The formatting is identical to errors and include the line number, line, and config path.

This also alters the background file access warning to use the function and introduces a warning for duplicate bindings.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -700,6 +700,8 @@ bool read_config(FILE *file, struct sway_config *config,
 			free(line);
 			return false;
 		}
+		config->current_config_line_number = line_number;
+		config->current_config_line = line;
 		struct cmd_results *res;
 		if (block && strcmp(block, "<commands>") == 0) {
 			// Special case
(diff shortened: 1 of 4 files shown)


sway/config.c
```diff
@@ -761,10 +763,36 @@ bool read_config(FILE *file, struct sway_config *config,
...
+			return;
+		}
+
+		va_start(args, fmt);
+		vsnprintf(temp, length, fmt, args);
+		va_end(args);
+
+		swaynag_log(config->swaynag_command, &config->swaynag_config_errors,
+			"Warning on line %i (%s) '%s': %s",
+			config->current_config_line_number, config->current_config_path,
```
emersion: What happens if this is not executed from the config, e.g. via `swaymsg`?

RedSoxFan: Added a check for `config->reading`.

emersion:
Thanks!

PR #3206, opened 2018-11-28
Implement support for swaymsg -t SUBSCRIBE [-m]

RedSoxFan:
Related to #3082

In `i3 4.16`, `i3-msg` can be used with the message type `subscribe` and has the ability to monitor for responses until killed. This adds support for both to swaymsg.

If the JSON array of event types is malformed or contains an invalid event, sway will send a response with `success` set to `false`. If swaymsg sees this, it will not display the failure and exit.

If the `subscribe` event is successful, swaymsg will wait for the first response and display that instead of the success message. If `-m/--monitor` is given, swaymsg will continue monitor for responses until killed or a malformed response is received.

For the `subscribe` event, the responses will always be printed as JSON. If `-r/--raw` is given, the JSON will not be pretty printed, which may be preferred when monitoring due to there being multiple responses.

Example: `swaymsg -t subscribe -m "['window']"`

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -668,7 +668,8 @@ void ipc_client_handle_command(struct ipc_client *client) {
 		// TODO: Check if they're permitted to use these events
 		struct json_object *request = json_tokener_parse(buf);
 		if (request == NULL) {
-			client_valid = ipc_send_reply(client, "{\"success\": false}", 18);
+			const char msg[] = "[{\"success\": false}]";
+			client_valid = ipc_send_reply(client, msg, strlen(msg));
 			wlr_log(WLR_INFO, "Failed to parse subscribe request");
 			goto exit_cleanup;
 		}
@@ -695,8 +696,8 @@ void ipc_client_handle_command(struct ipc_client *client) {
 				client->subscribed_events |= event_mask(IPC_EVENT_TICK);
 				is_tick = true;
 			} else {
-				client_valid =
-					ipc_send_reply(client, "{\"success\": false}", 18);
+				const char msg[] = "[{\"success\": false}]";
+				client_valid = ipc_send_reply(client, msg, strlen(msg));
 				json_object_put(request);
 				wlr_log(WLR_INFO, "Unsupported event type in subscribe request");
 				goto exit_cleanup;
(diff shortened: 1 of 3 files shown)


sway/ipc-server.c
```diff
@@ -663,7 +663,7 @@ void ipc_client_handle_command(struct ipc_client *client) {
 		// TODO: Check if they're permitted to use these events
 		struct json_object *request = json_tokener_parse(buf);
 		if (request == NULL) {
-			client_valid = ipc_send_reply(client, "{\"success\": false}", 18);
+			client_valid = ipc_send_reply(client, "[{\"success\": false}]", 20);
```
emersion: Can we use `strlen` instead of hardcoding the string length?

swaymsg/main.c
```diff
@@ -307,13 +307,15 @@ static void pretty_print(int type, json_object *resp) {
 int main(int argc, char **argv) {
 	static int quiet = 0;
 	static int raw = 0;
+	static int monitor = 0;
```
emersion: Maybe we can use `bool`s here?

swaymsg/main.c
```diff
@@ -422,26 +436,56 @@ int main(int argc, char **argv) {
...
-	close(socketfd);
-
 	free(command);
 	free(resp);
+
+	if (type == IPC_SUBSCRIBE && ret == 0) {
+		do {
+			struct ipc_response *reply = ipc_recv_response(socketfd);
+			if (!reply) {
+				continue;
```
emersion: We should probably `break` here

sway/ipc-server.c
RedSoxFan: Changed all replies in the case to use strlen

swaymsg/main.c
RedSoxFan: Done

swaymsg/main.c
RedSoxFan: Fixed

sway/ipc-server.c
```diff
@@ -668,7 +668,8 @@ void ipc_client_handle_command(struct ipc_client *client) {
 		// TODO: Check if they're permitted to use these events
 		struct json_object *request = json_tokener_parse(buf);
 		if (request == NULL) {
-			client_valid = ipc_send_reply(client, "{\"success\": false}", 18);
+			const char *msg = "[{\"success\": false}]";
+			client_valid = ipc_send_reply(client, msg, strlen(msg) + 1);
```
emersion: We don't want to send the terminating NULL byte.

Nitpick: `const char thing[] = …` is better than `const char *thing = …`.

emersion:
Thanks!

PR #3207, opened 2018-11-28
Implement swaynag -B/--button-no-terminal

RedSoxFan:
Related to #3082 

In `i3 4.16`, `i3-nagbar` introduces the flags `-B/--button-no-terminal` to run the action directly instead of inside a terminal. This implements the flags for swaynag for compatibility.

Since swaynag does not use an equivalent to `i3-sensible-terminal`, the flags `-b/--button` only uses a terminal when the environment variable `TERMINAL` is set, otherwise it acts the same as these new flags.

### DIFF
--- a/include/swaynag/swaynag.h
+++ b/include/swaynag/swaynag.h
@@ -44,6 +44,7 @@ struct swaynag_button {
 	int y;
 	int width;
 	int height;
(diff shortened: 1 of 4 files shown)


emersion:
Thanks!

PR #3208, opened 2018-11-28
Implement bar gaps

### LINKED ISSUE #2904: Feature Request: add gap funtionality to swaybar.
Ahoy, big fan of what you're doing and congrats on the big 1.0.

Now that you're at full feature-parody with i3 I was wondering if I might request something that I've been wanting for a while but assumed was too small and unimportant to bring up, gap functionality on swaybar. I know even i3-gaps doesn't support this on i3bar (and they say it would be a large undertaking), but I wasn't sure if maybe it would be easier for you since this is a from-scratch implementation. I don't know how much demand there would be for a feature like this, but as somebody who would only use lemonbar or polybar over the default to add gaps around the bar to match the window gaps, this is a feature I'd certainly love to see. 

Thanks.

RedSoxFan:
Closes #2904

Adds the bar subcommand `gaps <amount>|<horizontal> <vertical>|<top> <right> <bottom> <left>` to set gaps for swaybar. Due to restrictions on margins for a layer_surface, only the sides that are anchored to an edge of the screen can have gaps. Since there is support for per-side outer gaps for workspaces, those should be able to be used instead for the last side.

### DIFF
--- a/sway/sway-bar.5.scd
+++ b/sway/sway-bar.5.scd
@@ -61,6 +61,13 @@ Sway allows configuring swaybar in the sway configuration file.
 *binding\_mode\_indicator* yes|no
 	Enable or disable binding mode indicator. Default is _yes_.
 
+*gaps* <all> | <horizontal> <vertical> | <top> <right> <bottom> <left>
+	Sets the gaps from the edge of the screen for the bar. Gaps can either be
+	set all at once, per direction, or per side. Note that only sides that
+	touch an edge of the screen can have gaps. For the side that does not
+	touch an edge of the screen, per-side outer gaps for workspaces may be of
+	use.
+
 *height* <height>
 	Sets the height of the bar. Default height will match the font size.
 
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -188,6 +188,7 @@ sway_cmd bar_cmd_bindsym;
 sway_cmd bar_cmd_colors;
 sway_cmd bar_cmd_context_button;
 sway_cmd bar_cmd_font;
+sway_cmd bar_cmd_gaps;
 sway_cmd bar_cmd_mode;
 sway_cmd bar_cmd_modifier;
 sway_cmd bar_cmd_output;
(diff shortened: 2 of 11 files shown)


sway/sway-bar.5.scd
```diff
@@ -61,6 +61,13 @@ Sway allows configuring swaybar in the sway configuration file.
 *binding\_mode\_indicator* yes|no
 	Enable or disable binding mode indicator. Default is _yes_.
 
+*gaps* <amount>|<horizontal> <vertical>|<top> <right> <bottom> <left>
```
emersion: Shouldn't be 4 arguments at most here?

RedSoxFan: It is?

emersion:
```
<amount>|<horizontal> <vertical>|<top> <right> <bottom> <left>
1                     2                3       4        5
```

RedSoxFan: 1 arg: `gaps <amount>`
2 args: `gaps <horizontal> <vertical>`
4 args: `gaps <top> <right> <bottom> <left>`

Should I reword `<amount>` to something like `<all>`?

emersion: Ah, I see. Maybe we can put spaces around `|` to make it clearer (I thought it was `<amount>|<horizontal>` followed by `vertical>|<top> …`).

And yeah, "all" would be better.

emersion:
Thanks!

RedSoxFan:
> So I went to try this and due to me misunderstanding how to set it up, I stumbled on an interesting bug ... Probably not something a lot of people will screw up, but I got a chuckle out of it and figured you might want to idiot-proof it.

That's actually working as intended. Sway allows for bar commands to be given in the form `bar [<bar-id>] <bar-subcommand>`. In your example above, the first `gaps` is being parsed as the `bar-id` and the second `gaps` as the subcommand. We decided a while ago to have the only invalid `bar-id` be `id`, which will be parsed as the subcommand. This is to prevent future subcommands from breaking configs that used the subcommand as a `bar-id` before it was a subcommand.

iBurley:
> In your example above, the first `gaps` is being parsed as the `bar-id` and the second `gaps` as the subcommand.

Ah, interesting. Working good then, and once I figured out my derp the gaps are working perfectly and make me very happy, I appreciate the addition.

PR #3209, opened 2018-11-28
Do not strip quotes for cmd_mode

### LINKED ISSUE #3146: mode name require escaping
Using the minimal config below one can enter the modes `foo, bar` or `foo\,\ baz` normally. From mode `foo, bar` one can go to `foo\,\ baz`, but from there one cannot return back to `foo, bar`, it seems. In order to bind a key to it from a mode, the escapes are needed.

`swaymsg -t get_version`
```
{
   "human_readable": "1.0-beta.1-195-gc36665bc (Nov 17 2018, branch 'master')",
   "variant": "sway",
   "major": 1,
   "minor": 0,
   "patch": 0,
   "loaded_config_file_name": "\/home\/bart\/.config\/sway\/config"
 }
```

config:

```
bindsym Mod4+e exit
bindsym Mod4+r reload
bindsym Mod4+f mode "foo, bar"
bindsym Mod4+b mode "foo\,\ baz"
mode "foo, bar" {
  bindsym Return mode default
  bindsym Escape mode default
  bindsym b      mode "foo\,\ baz"
}
mode "foo\,\ baz" {
  bindsym Return mode default
  bindsym Escape mode default
  bindsym f      mode "foo, bar"
}
bar {
  position top
  font pango:DejaVu Sans Mono
}
```

[here's the log](https://pastebin.com/raw/At4MGEzJ)

RedSoxFan:
Fixes #3146

Like with cmd_bindsym and cmd_bindcode, the quotes should not be stripped for cmd_mode. cmd_mode performs its own stripping for the mode name and the only valid subcommands are cmd_bindsym and cmd_bindcode.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -420,6 +420,7 @@ struct cmd_results *config_command(char *exec) {
 	// Strip quotes and unescape the string
 	for (int i = handler->handle == cmd_set ? 2 : 1; i < argc; ++i) {
 		if (handler->handle != cmd_exec && handler->handle != cmd_exec_always
+				&& handler->handle != cmd_mode
 				&& handler->handle != cmd_bindsym
 				&& handler->handle != cmd_bindcode
 				&& handler->handle != cmd_set


emersion:
Thanks!

PR #3212, opened 2018-11-28
move to workspace: fix moving floating container to non-empty workspace

martinetd:
moving a container to a non-empty workspace will find a container to move
to in the destination workspace and call container_move_to_container,
which must not just skip floating containers


RedSoxFan:
Thanks!

PR #3214, opened 2018-11-28
Fix default_orientation auto

### LINKED ISSUE #3156: default_orientation auto does not work on transformed output
On sway version 1.0-beta.1-209-g67d24e8f (Nov 19 2018, branch 'HEAD'), a monitor rotated to portrait (`transform 90` or `transform 270`) defaults to horizontal layout for tiled containers. It should default to vertical layout instead (with the default `default_orientation` of `auto`).

It looks like this was introduced by the fix for #3015, which moved `apply_output_config` after workspace initialization to fix a crash (when `apply_output_config` triggered an output layout change, it ended up in layout code that assumed the output had a workspace). That change causes this issue because `output_get_default_layout` (called from `workspace_create`) looks at `output->height` and `output_width`, which now do not yet have the configured transform applied to them.

(The revision history is a little confusing: the relevant commit is cd0bae80b6f1cc27ffff47618b4332449468134c, but the crash referenced in its commit message was fixed differently. See discussion on #2993.)

I'm not sure what the correct fix is (if I did I'd be opening a pull request instead of an issue...):

- Keep calling `apply_output_config` after creating the default workspace, but fix up the orientation of the default workspace after that. This is probably the easiest fix to implement, but seems a little messy: we'd have to remember whether we created a default workspace, and I don't know if `apply_output_config` can end up in code that needs the default layout to already be correct (or worse: affect workspace/layout configuration, which our layout fixup then interferes with).

- Change `apply_output_config` to not synchronously trigger a layout change, and move it back before workspace initialization. I'm not sufficiently familiar with sway's event loop and/or the wlroots event system to know what this would look like. If feasible, we wouldn't have to temporarily have a workspace with the wrong layout.

If you give me some pointers on what a fix should look like, I'm happy to take a stab at it.

For completeness, see https://gist.github.com/marienz/b68168590660cef20cfa2091a40bbd0f for my config, log output, "swaymsg -t get_outputs", and "swaymsg -t get_tree" after creating two windows on the monitor with the wrong layout. The config is not very minimal, but it's mostly keybinds that should be easy enough to ignore.

(A careful reader of the config and log output will spot some "subpixel" stuff that doesn't exist in upstream sway... that's a local patch that doesn't affect layout, but is neither useful nor sane enough to submit. Let me know if that's an issue and I'll get rid of it.)

RedSoxFan:
Fixes #3156 

Since the output config is no longer applied before creating the default workspace, the layout for default workspaces on an output may not be correct. Due to the ordering of calls in output_enable being changed in several bug fix PRs, this just fixes the layout after the call to apply_output_config.

### DIFF
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -88,11 +88,12 @@ void output_enable(struct sway_output *output, struct output_config *oc) {
 
 	restore_workspaces(output);
 
+	struct sway_workspace *ws = NULL;
 	if (!output->workspaces->length) {
 		// Create workspace
 		char *ws_name = workspace_next_name(wlr_output->name);
 		wlr_log(WLR_DEBUG, "Creating default workspace %s", ws_name);
-		struct sway_workspace *ws = workspace_create(output, ws_name);
+		ws = workspace_create(output, ws_name);
 		// Set each seat's focus if not already set
 		struct sway_seat *seat = NULL;
 		wl_list_for_each(seat, &server.input->seats, link) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3216, opened 2018-11-28
Fix titlebar when no title or marks

### LINKED ISSUE #3215: Bad rect with titlebar for some X11 dialogs
1. Open gedit via Xwayland: `env GDK_BACKEND=x11 gedit`
2. Type some text
3. Ctrl+W
4. A "Are you sure you want to quit?" dialog appears with a weird rect on the left of the titlebar
4. Drag the dialog, there are damage issues

Also happens with other X11 apps (e.g. Atom).

RedSoxFan:
Fixes #3215 

This fixes the rendering of the titlebar when there is no title or marks.


emersion:
Thanks!

PR #3218, opened 2018-11-28
Fix json object ipc success return

johnae:
Pull request https://github.com/swaywm/sway/pull/3206 seems to have broken the expected json success return values so that existing ipc tooling won't work (eg. it turned an object into an object wrapped in an array).


RedSoxFan:
Thanks!

PR #3219, opened 2018-11-29
swaymsg: parse success of single object

RedSoxFan:
Related to #3218

Allows swaymsg to parse the success attribute of a single object response. This is needed for the subscribe event.


emersion:
Thanks!

PR #3222, opened 2018-11-30
Fix identifier output configs on reload

### LINKED ISSUE #3221: wildcard output order bugs out position on reload
When you have:

`output * bg .. ` after your outputs where you use the model make version output config, then on reload your positions will be wrong.
Weirdly the positions are correct and does not occur with regular output identifiers.. 

If you put the `output * bg ..` before the other output commands then reloading works as expected and positions are good.

It seems something goes wrong with merging..

faulty config:

```

 25 ### Monitors
 26 #
 27 # Layout 
 28 #
 29 #                               WorkOffice(5120,0),WorkClient(5120,390)
 30 # HomeLeft(0,1440), HomeRight(2560,1440), eDP-1 (5120,1440)
 31 #
 32 # Note: there is a bug where on reload/resume specific monitor configs don't work,
 33 #       so I use interface names here.
 34 
 35 #laptop display
 36 output eDP-1 {
 37   scale 2
 38   position 5120 1440
 39 }
 40 
 41 #work client
 42 output 'Lenovo Group Limited LT2252p Wide 6V8ABBP1' { 
 43   pos 5120 390
 44 }
 45 
 46 #output 'Lenovo Group Limited LEN LT2252pwA V1BR925' {
 47 output 'DP-1' {
 48   pos 5120 390
 49 }
 50 # work office
 51 output 'Dell Inc. DELL U2715H GH85D7B70JSS' {
 52   position 5120 0
 53 }
 54 output 'Dell Inc. DELL U2715H GH85D7B228PS' {
 55   position 5120 0
 56 }
 57 #home right
 58 output 'Dell Inc. DELL U2515H 9X2VY7470GQL' {
 59 #output DP-3 { 
 60   position 2560 1440
 61 }
 62 # home left
 63 output 'Dell Inc. DELL U2515H 9X2VY7470GXL' {
 64 #output DP-4 {
 65   position 0 1440
 66 }
 67 output DP-5 bg ~/.config/sway/DP-5.jpg fill
 68 output * bg ~/.config/sway/DP-6.jpg fill
```

correct config would be to move the final two commands at the top.

* Debug Log:

https://gist.github.com/wernerb/6c824eaa0b96e9e3c5ac18462ad787ac

RedSoxFan:
Fixes #3221

When generating default output configs to reset the outputs to their default settings on reload, the output name was being used. Additionally when determining the output config to apply, if there was an output config with the output name, that was being used without checking for an identifier config. This caused sway to completely ignore any of the users specified output configs that used identifiers.

To fix this issue, the following changes have been made:
1. Default output configs are created for the identifier now instead of name. This actually makes more sense anyway since you could hotplug multiple different outputs to the same port.
2. In get_output_config, which is only used to determine which output config to apply, output configs for both the name and identifier are queried. If both are found, a new output config is generated with the identifier config merged on top of the name config. If just one is found, a copy of that config is returned. This change also requires that the result from get_output_config be freed after use to prevent memory leaks, which required some minor changes to logic in apply_output_config_to_outputs.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -277,17 +277,51 @@ void apply_output_config(struct output_config *oc, struct sway_output *output) {
 }
 
 static struct output_config *get_output_config(char *name, char *identifier) {
+	struct output_config *oc_name = NULL;
 	int i = list_seq_find(config->output_configs, output_name_cmp, name);
 	if (i >= 0) {
-		return config->output_configs->items[i];
+		oc_name = config->output_configs->items[i];
 	}
 
+	struct output_config *oc_id = NULL;
 	i = list_seq_find(config->output_configs, output_name_cmp, identifier);
 	if (i >= 0) {
-		return config->output_configs->items[i];
+		oc_id = config->output_configs->items[i];
 	}
 
-	return NULL;
+	struct output_config *result = NULL;
+	if (oc_name && oc_id) {
+		// Generate a config named `<identifier> on <name>` which contains a
+		// merged copy of the identifier on name. This will make sure that both
+		// identifier and name configs are respected, with identifier getting
+		// priority
+		size_t length = snprintf(NULL, 0, "%s on %s", identifier, name) + 1;
+		char *temp = malloc(length);
+		snprintf(temp, length, "%s on %s", identifier, name);
+
+		result = new_output_config(temp);
+		merge_output_config(result, oc_name);
+		merge_output_config(result, oc_id);
+
+		wlr_log(WLR_DEBUG, "Generated output config \"%s\" (enabled: %d)"
+			" (%dx%d@%fHz position %d,%d scale %f transform %d) (bg %s %s)"
+			" (dpms %d)", result->name, result->enabled, result->width,
+			result->height, result->refresh_rate, result->x, result->y,
+			result->scale, result->transform, result->background,
+			result->background_option, result->dpms_state);
+
+		free(temp);
+	} else if (oc_name) {
+		// No identifier config, just return a copy of the name config
+		result = new_output_config(name);
+		merge_output_config(result, oc_name);
+	} else if (oc_id) {
+		// No name config, just return a copy of the identifier config
+		result = new_output_config(identifier);
+		merge_output_config(result, oc_id);
+	}
+
+	return result;
 }
 
 void apply_output_config_to_outputs(struct output_config *oc) {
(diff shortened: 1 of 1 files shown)


sway/config/output.c
```diff
@@ -277,17 +277,49 @@ void apply_output_config(struct output_config *oc, struct sway_output *output) {
...
 
-	return NULL;
+	struct output_config *result = NULL;
+	if (oc_name && oc_id) {
+		// Generate a config named `<identifier> on <name>` which contains a
+		// merged copy of the identifier on name. This will make sure that both
+		// identifier and name configs are respected, with identifier getting
+		// priority
+		size_t length = strlen(identifier) + strlen(" on ") + strlen(name) + 1;
+		char *temp = malloc(length);
```
emersion: Need to `free` this. Also, when computing the length it'd probably be better to use `snprintf`'s return value instead of `strlen`.

sway/config/output.c
```diff
@@ -277,17 +277,49 @@ void apply_output_config(struct output_config *oc, struct sway_output *output) {
...
+
+		wlr_log(WLR_DEBUG, "Generated output config for %s on %s (enabled: %d)"
+			" (%dx%d@%fHz position %d,%d scale %f transform %d) (bg %s %s)"
+			" (dpms %d)", name, identifier, result->enabled, result->width,
+			result->height, result->refresh_rate, result->x, result->y,
+			result->scale, result->transform, result->background,
+			result->background_option, result->dpms_state);
+	} else if (oc_name) {
+		// No identifier config, just return a copy of the name config
+		result = new_output_config(strdup(name));
```
emersion: No need to `strdup` here -- `new_output_config` takes a `const char *`.

emersion:
>Default output configs are created for the identifier now instead of name. This actually makes more sense anyway since you could hotplug multiple different outputs to the same port.

Is this really required? What are default configs sued for again?

sway/config/output.c
RedSoxFan: Fixed

sway/config/output.c
RedSoxFan: Fixed

RedSoxFan:
> Is this really required? What are default configs sued for again?

Default configs are generated on reload to reset output settings that are not specified. Without them, reloading will not change settings not in the config and anything that was removed or not in the config and changed at runtime will not be reset

emersion:
Thanks!

PR #3229, opened 2018-12-01
Fix fullscreen command and improve documentation

mwenzkowski:
The first commit fixes a problem that I have when using the following keybinding on an empty workspace

`bindsym Mod4+v fullscreen disable,exec termite -e pulsemixer`.

With this binding I want to open an application but in case that another view is currently in fullscreen mode, it should leave fullscreen mode first. This does not work on an empty Workspace, because `fullscreen disable` returns with error status `CMD_INVALID` and then the following exec command will not be executed (cf. https://github.com/swaywm/sway/blob/master/sway/commands.c#L307).
In i3 this binding works on empty workspaces.

The second commit documents the optional arguments of the fullscreen command, which can be found in i3's documentation too.

### DIFF
--- a/sway/commands/fullscreen.c
+++ b/sway/commands/fullscreen.c
@@ -13,14 +13,14 @@ struct cmd_results *cmd_fullscreen(int argc, char **argv) {
 		return error;
 	}
 	if (!root->outputs->length) {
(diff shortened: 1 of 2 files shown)


RedSoxFan:
Thanks!

PR #3239, opened 2018-12-02
Add option to bindsym/bindcode to suppress warning on override

ForTheReallys:
Recently, bindsym/bindcode was changed to give a warning when you rebind a keystroke.  This patch allows the user to suppress that warning with the --no-warn flag.

### DIFF
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -161,6 +161,7 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 	binding->type = bindcode ? BINDING_KEYCODE : BINDING_KEYSYM;
 
 	bool exclude_titlebar = false;
+	bool warn = true;
 
 	// Handle --release and --locked
 	while (argc > 0) {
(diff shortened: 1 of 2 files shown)


ForTheReallys:
Woops I forgot about that.  How's this?

sway/commands/bind.c
```diff
@@ -258,9 +261,11 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 			wlr_log(WLR_INFO, "Overwriting binding '%s' for device '%s' "
 					"from `%s` to `%s`", argv[0], binding->input,
 					binding->command, config_binding->command);
-			config_add_swaynag_warning("Overwriting binding '%s' for device "
-					"'%s' to `%s` from `%s`", argv[0], binding->input,
-					binding->command, config_binding->command);
+			if (warn) {
+				config_add_swaynag_warning("Overwriting binding '%s' for device "
```
RedSoxFan: Style: Can you please change this so it does not go past column 80?

RedSoxFan:
Thanks!

PR #3241, opened 2018-12-03
cmd_focus: show scratchpad if hidden

### LINKED ISSUE #3235: Sway crashes when focusing a scratchpad window via swaymsg [app_id="something"] focus
Sway version: `sway version 1.0-beta.1-256-ga22d0c0f (Nov 28 2018, branch 'master')`
Sway config: default config
Debug log: https://gist.github.com/camoz/8d8be0ee4bde590f375374677bcbd816
# Steps to reproduce
* Start sway with default config
* run `termite --name="sdf"`
* move the new window to scratchpad via key binding
* try to show it via `swaymsg [app_id="sdf"] focus`
# Expected behavior
Scratchpad window (termite) shows up
# What happens
Sway crashes, see debug log.

BTW I only tested that with termite for now.

RedSoxFan:
Fixes #3235 

If a scratchpad container is hidden, it is still focusable using criteria and should be shown. This fixes a segfault when attempting to rebase the cursor since previously the scratchpad container would not be on any output.

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -269,6 +269,9 @@ struct cmd_results *cmd_focus(int argc, char **argv) {
 	}
 
 	if (argc == 0 && container) {
+		if (container->scratchpad && !container->workspace) {
+			root_scratchpad_show(container);
+		}
 		seat_set_focus_container(seat, container);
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3242, opened 2018-12-03
config: set $XDG_CONFIG_HOME if empty

ianyfan:
Before, it was only set if it was not set, but it should also be done
when empty.

Source: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -316,7 +316,8 @@ static char *get_config_path(void) {
 		SYSCONFDIR "/i3/config",
 	};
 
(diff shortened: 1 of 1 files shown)


emersion:
BTW, this is kind of unrelated but we should probably not `setenv` here.

ianyfan:
Makes sense but it might make the `wordexp` logic a bit more complicated.

emersion:
Thanks!

PR #3245, opened 2018-12-03
config: do not set $XDG_CONFIG_HOME if unset or empty

### LINKED ISSUE #3243: Sway sets XDG_CONFIG_HOME
See https://github.com/swaywm/sway/pull/3242#issuecomment-443670798

ianyfan:
Instead redefine the config paths to use the default $HOME/.config

Closes #3243


sway/config.c
```diff
@@ -316,28 +316,16 @@ static char *get_config_path(void) {
...
 	}
 
-	wordexp_t p;
-	char *path;
-
-	int i;
-	for (i = 0; i < (int)(sizeof(config_paths) / sizeof(char *)); ++i) {
-		if (wordexp(config_paths[i], &p, 0) == 0) {
-			path = strdup(p.we_wordv[0]);
+	for (int i = 0; i < (int)(sizeof(config_paths) / sizeof(char *)); ++i) {
```
emersion: This can be `size_t`

emersion:
Can you remove `// Not reached` on line 337? It can be reached on `wordexp` failures.

(I guess `WRDE_UNDEF` is not strictly necessary since you overwrite entries with `XDG_CONFIG_HOME`, but I think it's still nice to have)

PR #3256, opened 2018-12-05
Fix command list execution

mwenzkowski:
Test plan:

1.  Open two terminals side by side:
 ```
+-------+-------+
|   A   |   B   |
+-------+-------+
```
2. In terminal B type `swaymsg focus left, fullscreen`

3. Terminal A is in fullscreen mode.

Before this patch terminal B was made fullscreen.


emersion:
Thanks!

PR #3257, opened 2018-12-05
Added null pointer checks to non-static functions in list.c

artiqknight:
Since create_list can return NULL, functions receiving a list
should check for null pointers I believe.
Lists are used a lot throughout sway and often are assumed not to be
NULL. It's a small thing but I haven't been using sway for that long
and I figured this is a good place to start.

### DIFF
--- a/common/list.c
+++ b/common/list.c
@@ -6,7 +6,7 @@
 
 list_t *create_list(void) {
 	list_t *list = malloc(sizeof(list_t));
-	if (!list) {
+	if (list == NULL) {
 		return NULL;
 	}
 	list->capacity = 10;
@@ -40,11 +40,17 @@ void list_foreach(list_t *list, void (*callback)(void *item)) {
 }
 
 void list_add(list_t *list, void *item) {
+	if (list == NULL || item == NULL) {
+		return;
+	}
 	list_resize(list);
 	list->items[list->length++] = item;
 }
 
 void list_insert(list_t *list, int index, void *item) {
+	if (list == NULL || item == NULL) {
+		return;
+	}
 	list_resize(list);
 	memmove(&list->items[index + 1], &list->items[index], sizeof(void*) * (list->length - index));
 	list->length++;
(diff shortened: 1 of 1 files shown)


RyanDwyer:
If we were to merge your PR as is and use the same logic everywhere (returning early if passed arguments are `NULL`) then this would leave sway in an indeterminable state after OOM situations. Instead, I think it would be better if the functions which created the lists were to check that they were created successfully, and attempt to handle the situation gracefully if possible.

artiqknight:
Thank you for your comment. I agree that this is not the best solution.
Lists are created in a lot of places and I figured it's not my place to change all that code.
As it is right now however, there are a lot of places where null pointers would be derefenced, such as in the call to list_add() in swaynag/main.c line 35.
Is it better to crash early in these cases or would it be better to check every call to create_list() and exit with a failure?

RyanDwyer:
> As it is right now however, there are a lot of places where null pointers would be derefenced, such as in the call to list_add() in swaynag/main.c line 35.
> Is it better to crash early in these cases or would it be better to check every call to create_list() and exit with a failure?

I consider it critical that sway continue running during OOM situations, but not secondary programs like swaynagbar, swaybar and swaybag. The reason is if sway crashes then the user has lost their unsaved work, but if any of the secondary programs crash the user can just restart them and hasn't lost any work. For this reason I wouldn't bother trying to prevent OOM-related crashes in the secondary programs.

For sway, it's not desirable to crash or exit gracefully if the user didn't request it. What we should do depends on the context. If the user is running a command then we can just return a failure. If a view is trying to map then I think we can reply with a "no memory" response but I'm not sure how exactly. For memory allocations within transaction code, that really needs to work in all situations, so I'm not sure how we'd get around that other than reserving some memory ahead of time.

artiqknight:
> For this reason I wouldn't bother trying to prevent OOM-related crashes in the secondary programs.
Letting them just segfault feel dirty though. Swaynag already uses returns an exit code, null pointers could be checked and if there's no memory for a list Swaynag could return a failure.

> For memory allocations within transaction code, that really needs to work in all situations, so I'm not sure how we'd get around that other than reserving some memory ahead of time.

I took a look at transactions, but I started with a secondary program precisely because I'm not familiar with the inner workings of Sway. In transaction_create() the calloc for the struct itself is checked for a np, but the list for instructions is not. This could cause issues as well.

I think allocating memory upfront and checking that would be a good idea yeah.
But that would mean the location of those allocations needs to be changed I think.
Right now for example list_add() calls list_resize() which reallocs a larger block.
list_resize sets items to the realloc and doesn't check it, which means it could set items to a null pointer I think.
list->items is then used in list_add() which would then crash I think.
That means this loop transaction_commit_dirty() could crash because it calls list_add() in transaction_add_node() in a for loop:
```
	for (int i = 0; i < server.dirty_nodes->length; ++i) {
		struct sway_node *node = server.dirty_nodes->items[i];
		transaction_add_node(transaction, node);
		node->dirty = false;
}
```
Allocating memory upfront would probably mean moving the allocations out of the internal functions in loop etc. and passing pointers to memory to them.
But like I said I might be talking out of my ass here.

PR #3260, opened 2018-12-06
cmd_split: flatten when possible

### LINKED ISSUE #3259: cmd_split does not flatten tree
**Steps to reproduce:** 
I. Open two terminals
II. Tab or stack them
III. Attempt to use horizontal split ($mod+b)
IIII. Title of the tab changes to something strange.

**Picture of what happens:**
https://imgur.com/a/WrzqNvG

**Debug log:**
 https://gist.github.com/7415963987456321/859c5b76ab0c9b992be2fa1092ba0a93

**Version:**
Latest master, also happens on older versions, I meant to report this bug a while ago.

**Config:**
https://gist.github.com/7415963987456321/809f04a30aac61ab12504d8aa0e2111f

RedSoxFan:
Fixes #3259 

Matches i3 behavior

### DIFF
--- a/sway/commands/split.c
+++ b/sway/commands/split.c
@@ -18,6 +18,10 @@ static struct cmd_results *do_split(int layout) {
 		workspace_split(ws, layout);
 	}
 
+	if (con->parent->parent) {
+		container_flatten(con->parent->parent);
+	}
+
 	arrange_workspace(ws);
 
 	return cmd_results_new(CMD_SUCCESS, NULL, NULL);


RyanDwyer:
What about checking if the focused container has a single child and is already the desired layout, and returning an error if so? It would save the hassle of creating a container only to destroy it, and allows IPC scripts to detect when this is a no-op.

RedSoxFan:
The container being created would not be what's destroyed. The parent of that container would be. Also, it may not be a no-op since the parent does not have to be the same layout. I guess it could work like a layout toggle when there is a single child though.

7415963987456321:
Just tested this. It indeed works a lot better now.

RyanDwyer:
Thanks!

PR #3264, opened 2018-12-06
list: double list capacity when resizing instead of incrementing

ianyfan:
This is the industry standard since it allows insertion to be amortized
O(1) time.

### DIFF
--- a/common/list.c
+++ b/common/list.c
@@ -17,7 +17,7 @@ list_t *create_list(void) {
 
 static void list_resize(list_t *list) {
 	if (list->length == list->capacity) {
-		list->capacity += 10;
+		list->capacity *= 2;
 		list->items = realloc(list->items, sizeof(void*) * list->capacity);
 	}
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Would there be any benefit in making it a power of 2? Eg. set the initial capacity to 8 instead of 10.

ianyfan:
Possibly, but it's more complicated than it seems so I don't think it's worth us worrying about it.
Ref: https://stackoverflow.com/questions/3190146/is-it-better-to-allocate-memory-in-the-power-of-two

progandy:
Would it make sense to cap the allocation amount at a value like a few kB? At some point amortized time might be less important than memory footprint, especially if there are a lot of partially filled arrays.

ESipalis:
@Freyr666 it checks in `list_resize` method.

RedSoxFan:
Thanks!

PR #3267, opened 2018-12-06
add support for named colors

1ace:


ddevault:
-1

emersion:
Are those supported by i3?

progandy:
@emersion No, i3 only supports hex colors in in RGB or RGBA with alpha channel.

https://github.com/i3/i3/blob/db0add0f11995d4273c76c3c1ad3ab131cea6a99/libi3/draw_util.c#L86

emersion:
I would've considered merging this if it was for i3 compatibility, but if it's a new feature, then I'm not for this.

Sorry about this! Next time feel free to ask us on IRC or open an issue about new features.

1ace:
Do you mean sway can't have features that i3 doesn't have? That's seems unnecessarily restrictive.

emersion:
No, it just means @ddevault and I don't like this feature.

1ace:
That's fair 😅

PR #3271, opened 2018-12-09
 list.c: Remove list_foreach

ianyfan:
I find this function is mostly useless since the callback has to accept a `void *`pointer. The only time I really found it useful was to free every item before freeing a list, which I then discovered already had a function dedicated for it, but it was tucked in `stringops.c`

This commit replaces those cases (6 out of 9) with `free_flat_list` and moves the function into `list.c`
The rest have been replaced with for loops.

A second commit cleans up some list-related code.


sway/ipc-server.c
```diff
@@ -72,15 +72,12 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 	close(ipc_socket);
 	unlink(ipc_sockaddr->sun_path);
 
-	while (ipc_client_list->length) {
-		struct ipc_client *client = ipc_client_list->items[0];
-		ipc_client_disconnect(client);
+	for (int i = 0; i < ipc_client_list->length; ++i) {
+		ipc_client_disconnect(ipc_client_list->items[i]);
```
RyanDwyer: `ipc_client_disconnect` removes the item from the list, which changes the length while you're iterating it. I'm pretty sure you want the while loop with `items[0]` instead of `items[i]`.

swaynag/types.c
```diff
@@ -147,10 +147,8 @@ void swaynag_type_free(struct swaynag_type *type) {
 }
 
 void swaynag_types_free(list_t *types) {
-	while (types->length) {
-		struct swaynag_type *type = types->items[0];
-		swaynag_type_free(type);
-		list_del(types, 0);
+	for (int i = 0; i < types->length; ++i) {
+		swaynag_type_free(types->items[0]);
```
RyanDwyer: Do you mean to use `i` instead of `0` here?

RyanDwyer:
BTW, I'd be in favour of renaming `free_flat_list` to something that starts with `list_`. I'd also like to see `create_list` renamed to `list_create` at some point (not in this PR though).

sway/ipc-server.c
ianyfan: Oh, I see. I'll revert this change then.

RyanDwyer:
Thanks!

PR #3275, opened 2018-12-09
Rewrite strip_whitespace and remove readline.c

ianyfan:
`strip_whitespace` was pretty awfully written.
`read_line` has been replaced by getline, hopefully I've rewritten everything correctly.
`peek_line` was only used once, and the way it was being used was quite inefficient, so I just absorbed it into `detect_brace`.


sway/commands.c
```diff
@@ -251,15 +251,15 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 			criteria_destroy(criteria);
 			config->handler_context.using_criteria = true;
 			// Skip leading whitespace
-			head += strspn(head, whitespace);
+			for (; isspace(*head); ++head) {}
 		}
 		// Split command list
 		cmdlist = argsep(&head, ";");
-		cmdlist += strspn(cmdlist, whitespace);
+		for (; isspace(*cmdlist); ++head) {}
```
emersion: `++cmdlist`?

sway/config.c
```diff
@@ -1,4 +1,5 @@
-#define _XOPEN_SOURCE 600 // for realpath
+#define _POSIX_C_SOURCE 200809L
+#define _XOPEN_SOURCE 500 // for realpath
```
emersion: Don't define both, they'll cause issues because `_XOPEN_SOURCE` defines implicitly `_POSIX_C_SOURCE`

sway/main.c
```diff
@@ -47,31 +47,27 @@ void detect_raspi(void) {
...
 		if (strstr(line, "Raspberry Pi")) {
 			raspi = true;
+			break;
 		}
-		free(line);
 	}
 	fclose(f);
 	FILE *g = fopen("/proc/modules", "r");
 	if (!g) {
 		return;
```
emersion: `free(line)`

sway/config.c
ianyfan: I did not know that, thanks

sway/main.c
```diff
@@ -146,6 +138,19 @@ static void log_env(void) {
 	}
 }
 
+static void log_file(FILE *f) {
+	char *line = NULL;
+	size_t line_size = 0;
+	ssize_t nread;
+	while ((nread = getline(&line, &line_size, f)) != -1) {
+		if (line[nread -1] == '\n') {
```
emersion: Style: missing space

ianyfan:
Ah, that's a good point. And since that functionality is needed in quite a few places, maybe this isn't a change worth doing.

emersion:
I'd still like to see readline.c go away

ianyfan:
I've added a commit to address the backslash continuation issue, hopefully this works? It's not the prettiest though...

common/stringop.c
```diff
@@ -9,24 +9,17 @@
...
-		i--;
-	} while (i >= 0 && (str[i] == ' ' || str[i] == '\t'));
-	str[i + 1] = '\0';
-	return str;
+void strip_whitespace(char *str) {
+	size_t len = strlen(str);
+	size_t start = strspn(str, whitespace);
+	memmove(str, &str[start], len + 1 - start);
+
+	if (!*str) return;
```
emersion: Style: missing brackets

emersion:
Thanks!

ianyfan:
Oh, I was going to squash the commits, but forgot, sorry.

PR #3279, opened 2018-12-10
Rework default output configs

### LINKED ISSUE #3261: Clients crash (disappear?) when reloading config w/ an output disabled
`swaymsg -t get_version`: `sway version cf6edaf26aa1461b89552727d2435d6fe9d0adf3`

Debug log: 
[swaylog.txt](https://github.com/swaywm/sway/files/2651630/swaylog.txt)

That's me launching sway, launching a couple of terminals and then reloading the configuration.

This used to work.

This is the output from termite (as it crashes):
```
Gdk-Message: 19:28:21.821: Unable to load hand2 from the cursor theme
Gdk-Message: 19:28:27.635: Error flushing display: Broken pipe
```

**THE KEY**

I was able to repro with stock configuration + one extra line:
```
output "DP-2" disable
```

The interesting thing is, the only thing that seems different than normal is
  1. possibly a newer version of Sway, but I don't think so - I can revert or try to bisect if it comes to that
  2. my external monitor is usually "DP-1". This is the first time I've noticed this bug, or the output being named otherwise. (To be clear, I updated my configs an hour ago to use "DP-2", so this ought to work).

### LINKED ISSUE #3277: Reloading the config file turns off scaling
Reloading the config file ('reload' binding in config file) turns of scaling for my display. 

sway version 1.0-beta.2-20-gcf6edaf2 (Dec  6 2018, branch 'master')
wlroots 0.2.r18.gfdb67ff6-1
(sway-git and wlroots-git packages in the arch user repository)

These are the configurations I've tried to set my scaling with:
``output eDP-1 pos 0 0 res 2560x1600 scale 1.9``
and
``output eDP-1 pos 0 0 res 2560x1600``
``output eDP-1 scale 1.9``
I've also tried other scaling factors and it makes no difference.

[debug.txt](https://github.com/swaywm/sway/files/2660917/debug.txt)

RedSoxFan:
Fixes #3277 
Fixes #3261

Default output configs were generated on reload to reset an output to its default settings. The idea was that anything that was removed from the config or changed at runtime and not in the config should be reset on reload. Originally, they were created using the output name. Recently, they were changed to use the output identifier. It turns out that there are issues of shadowing with that solution as well. This should fix those issues.

Instead of generating the default output configs on reload and storing them in the output config list to merge on top of, they are now only generated when retrieving the output config for an output during a reload. This means that the default output configs are never stored anywhere and just used as a base to merge unaltered user configs on top of during a reload.

Starting with a blank output config, merges get applied in the following order:
1. Default output config (only during a reload)
2. Wildcard config (only if neither output name or output identifier exist)
3. Output name config
4. Output identifier config

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -276,7 +276,24 @@ void apply_output_config(struct output_config *oc, struct sway_output *output) {
 	}
 }
 
-static struct output_config *get_output_config(char *name, char *identifier) {
+static void default_output_config(struct output_config *oc,
+		struct wlr_output *wlr_output) {
+	oc->enabled = 1;
+	if (!wl_list_empty(&wlr_output->modes)) {
+		struct wlr_output_mode *mode =
+			wl_container_of(wlr_output->modes.prev, mode, link);
+		oc->width = mode->width;
+		oc->height = mode->height;
+		oc->refresh_rate = mode->refresh;
+	}
+	oc->x = oc->y = -1;
+	oc->scale = 1;
+	oc->transform = WL_OUTPUT_TRANSFORM_NORMAL;
+}
+
+static struct output_config *get_output_config(char *identifier,
+		struct sway_output *sway_output) {
+	const char *name = sway_output->wlr_output->name;
 	struct output_config *oc_name = NULL;
 	int i = list_seq_find(config->output_configs, output_name_cmp, name);
 	if (i >= 0) {
@@ -289,7 +306,10 @@ static struct output_config *get_output_config(char *name, char *identifier) {
 		oc_id = config->output_configs->items[i];
 	}
 
-	struct output_config *result = NULL;
+	struct output_config *result = result = new_output_config("temp");
+	if (config->reloading) {
+		default_output_config(result, sway_output->wlr_output);
+	}
 	if (oc_name && oc_id) {
 		// Generate a config named `<identifier> on <name>` which contains a
 		// merged copy of the identifier on name. This will make sure that both
@@ -299,7 +319,8 @@ static struct output_config *get_output_config(char *name, char *identifier) {
 		char *temp = malloc(length);
 		snprintf(temp, length, "%s on %s", identifier, name);
 
-		result = new_output_config(temp);
+		free(result->name);
+		result->name = temp;
 		merge_output_config(result, oc_name);
 		merge_output_config(result, oc_id);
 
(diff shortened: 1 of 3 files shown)


RedSoxFan:
@colemickens, can you please try this and confirm that is fixes #3261 for you?

sway/config/output.c
```diff
@@ -300,6 +317,9 @@ static struct output_config *get_output_config(char *name, char *identifier) {
 		snprintf(temp, length, "%s on %s", identifier, name);
 
 		result = new_output_config(temp);
+		if (config->reloading) {
+			default_output_config(result, sway_output->wlr_output);
+		}
```
emersion: Can we move these lines to line 309, since we always to the same thing for all `if` branches?

sway/config/output.c
```diff
@@ -314,11 +334,27 @@ static struct output_config *get_output_config(char *name, char *identifier) {
...
 		merge_output_config(result, oc_id);
+	} else if (config->reloading) {
+		// Neither config exists, but we need to reset the output so create a
+		// default config for the output and if a wildcard config exists, merge
+		// that on top
+		result = new_output_config(identifier);
+		default_output_config(result, sway_output->wlr_output);
+		i = list_seq_find(config->output_configs, output_name_cmp, "*");
```
emersion: Isn't this done already by `get_output_config`'s caller? Otherwise the wildcard config wouldn't be merged in case `oc_{name,id}` is not NULL.

RedSoxFan: The caller will not have the wildcard config merged on top of the default output config. Without this outputs without their own output config will not get reset on reload and just have whatever is in the wildcard config itself changed.

sway/config/output.c
RedSoxFan: I refactored it to move those lines above the branches

sway/config/output.c
emersion: Hmm, I'm not sure why there's a need for a special case here.

There are three cases:

* There's no output config, no wildcard config: need to use the default one as is
* There's no output config, a wildcard config: need to use the default config with the wildcard config merged on top
* There's an output config: need to use the default config, merge the wildcard config on top if any, and _after_ that merge the output config on top

So I think `get_output_config` should always search for a wildcard config when reloading on line 311.

RedSoxFan: Currently, the wildcard config gets merged on to all output configs when it is modified so the output config already has the information from the wildcard config. See `merge_wildcard_on_all` (L90) and `store_output_config` (L100).

Wildcard is currently treated as a special shorthand to apply to all rather than plain defaults. Changing it to apply wildcard here instead would be a breaking change. For example, if you manually set `dpms on` for a specific output, swayidle would not be able to change the dpms state with `output * dpms off` since the output specific `dpms on` would override it.

emersion: Hmm, I see. Thanks for the explanation.

emersion:
Thanks!

PR #3280, opened 2018-12-10
Call wlr_output_enable for disabled new outputs

### LINKED ISSUE #2948: Disabling an output doesn’t DPMS it off
I disabled my laptop’s output with `output LVDS-1 disable`, expecting it would be DPMS’d off, but instead it keeps the last frame of the TTY painted on it, with backlight on.

RedSoxFan:
Fixes #2948

When a new output is detected and it is disabled by the output config, call `wlr_output_enable(output->wlr_output, false)` to DPMS off the output.


emersion:
Thanks!

PR #3284, opened 2018-12-10
[WIP] Introduce shell_expand

emersion:
This performs shell expansion without field splitting. This fixes issues with
filenames containing spaces, like https://github.com/swaywm/sway/issues/654#issuecomment-445982588.

This commit also cleans up complicated logic to join fields and to set
XDG_CONFIG_HOME.

### DIFF
--- a/common/meson.build
+++ b/common/meson.build
@@ -4,14 +4,15 @@ lib_sway_common = static_library(
 		'background-image.c',
 		'cairo.c',
 		'ipc-client.c',
+		'list.c',
 		'log.c',
 		'loop.c',
-		'list.c',
 		'pango.c',
 		'readline.c',
+		'shexp.c',
 		'stringop.c',
 		'unicode.c',
-		'util.c'
+		'util.c',
 	),
 	dependencies: [
 		cairo,
--- a/common/shexp.c
+++ b/common/shexp.c
@@ -1,0 +1,51 @@
+#define _POSIX_C_SOURCE 200809L
+#include <stdlib.h>
+#include <string.h>
+#include <wordexp.h>
+#include "shexp.h"
+
+static const char escaped_quote[] = "'\"'\"'";
+
+static char *escape(const char *str) {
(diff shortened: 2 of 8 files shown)


ianyfan (dismissed):
LGTM, may want to add `#include <stdbool.h>` to `shexp.c` if you want to be pedantic about it.
Maybe add a comment documenting its effect, like:
`Takes a pointer to a string and reallocates it with the result of its shell expansion. Returns true if it succeeds, otherwise returns false, leaving the original string unchanged.`
Should the `strdup`ed result be allocated to a local variable in case it fails? or is that being too safe? ¯\_(ツ)_/¯

RedSoxFan (dismissed):

emersion:
Unfortunately this doesn't completely work. We'll need to escape ` \t\n|&;<>(){}`.

emersion:
All right, this works for me and preserves things like `\t`.

progandy:
Should command substitution be disabled as well with ```WRDE_NOCMD```?

It looks like the only issue now would be paths that contain a literal ```$``` sign instead of a variable in background image paths and SYSCONFDIR, but that shouldn't be very common.

emersion:
Damn, nothing is expanded anymore now. I guess we need to use double quotes, and escape things that need to be escaped inside (`\` and `"`).

emersion:
Scratch that, tilde expansion won't work anymore. I'm tempted to do it by joining words together (like we do now) and ignoring the fact the there can be tabs in filenames.

progandy:
It might be easier to create your own (e.g. ragel) parser for variables and tilde instead of hacking around wordexp.

emersion:
Yes, I've been tempted to copypasta code from mrsh instead of doing this.

panaman67:
@emersion  What is the status of this PR?

emersion:
There's no good way to shell-expand an argument and preserve whitespace. The current implementation is the least broken thing we can do. It would still be useful to have a common `shell_expand` function to do that, though.

PR #3287, opened 2018-12-12
swaybar: handle block->urgent

RedSoxFan:
When `block->urgent` is set, use the urgent colors. This matches i3bar's
behavior. Previously, swaybar just ignored the property.

This also adds in rendering for right borders, which was missing.


emersion:
Thanks!

PR #3288, opened 2018-12-12
Keep focus when destroying containers on non-vis ws

### LINKED ISSUE #3281: Sway messes up focus when last view unmaps in non-visible workspace
Have two outputs, and arrange the workspaces like this:

* Output 1:
    * Workspace 1 (can be empty or have views, doesn't matter)
* Output 2:
    * Workspace 2 with at least one view
    * Workspace 3 with a single terminal (focused)

In the terminal on workspace 3, run `sleep 5 && exit`. While it's sleeping, switch to workspace 2 then focus workspace 1.

When the terminal view unmaps, sway (incorrectly) gives focus inactive to workspace 3, preventing it from destroying and causing it to show your desktop background as it's now empty. Additionally, swaybar doesn't receive a workspace::focus event so it continues to show workspace 2 as focus inactive. This makes it appear as though all the views on workspace 2 suddenly died.

Sway should not need to adjust the order of the focus stack when destroying a non-visible workspace.

RedSoxFan:
Fixes #3281

Changing the focus stack when destroying a container's node on a
non-visible workspace (on an non-focused output) incorrectly causes
the non-visible workspace to become visible. If the workspace is empty,
it will not be destroyed since it is now visible. Additionally since
there was no workspace::focus event, swaybar still shows the previous
workspace as focus-inactive. It also makes no sense to change visible
workspaces due to a container on a non-visible workspace being
destroyed.

Since the focus will either be set when switching to the non-visible
workspace or the workspace will be destroyed due to being empty, there
is no need to change the focus stack when destroying a container on a
non-visible workspace.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -159,6 +159,14 @@ static void handle_seat_node_destroy(struct wl_listener *listener, void *data) {
 		return;
 	}
 
+	if (node->type == N_CONTAINER &&
+			!workspace_is_visible(node->sway_container->workspace)) {
+		// Do not change the focus stack on a non-visible workspace. The focus
+		// will either be set when the workspace becomes visible or the
+		// workspace is now empty and will be destroyed
+		return;
+	}
+
 	// Find new focus_inactive (ie. sibling, or workspace if no siblings left)
 	struct sway_node *next_focus = NULL;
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3289, opened 2018-12-12
Fix criteria execution in view_map

### LINKED ISSUE #3285: The command "fullscreen disable" does not make focused view non-fullscreen
* Sway Version: sway version 1.0-beta.2-29-g6acbe84f (Dec 10 2018, branch 'master')

* Description:
During the few last days I have noticed that my criteria for Telegram's `Media viewer` does not longer bring all attributes into effect and discovered that it was the `fullscreen disable` attribute not being applied anymore. I tested my configuration on `i3-gaps` and it worked all fine as I expected.

* Steps to reproduce:
~ Append: `for_window [title="^Media viewer$"] fullscreen disable, border none, resize set 1920 px 1080 px` to your configuration file
~  Run Telegram _(I haven't tested it with other programs that may force fullscreen on focused view)_
~ Open any chat and click on an image to open image preview

_By the way, I would be totally fine with fullscreen image preview except for a disparity between Sway and i3, in which the case of, floating windows, such as GTK3's file picker, are opened behind fullscreen windows and, due to that, may steal focus and make a fullscreen window non-closeable (pressing $mod+f toggles fullscreen off, but does not close the window). Shall I open another issue for this?_

mwenzkowski:
This patch moves view_execute_criteria(view) at the end of the function.
Previously, if a view requested to be started in fullscreen, this was
done after execution of criteria and hence it was impossible to disable
fullscreen via criteria.

Fixes #3285 

I don't use telegram but I was able to reproduce the problem in #3285 with mpv by adding
```
for_window [app_id="^mpv$"] fullscreen disable
```
to my config and starting mpv with `mpv --fullscreen <file>`. Before this patch mpv remained in fullscreen.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -605,7 +605,6 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 
 	view_update_title(view, false);
 	container_update_representation(view->container);
-	view_execute_criteria(view);
 
 	if (decoration) {
 		view_update_csd_from_client(view, decoration);
@@ -621,6 +620,8 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 			arrange_workspace(view->container->workspace);
 		}
 	}
+
+	view_execute_criteria(view);
 
 	if (should_focus(view)) {
 		input_manager_set_focus(&view->container->node);


sway/tree/view.c
```diff
@@ -625,6 +624,8 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 	if (should_focus(view)) {
 		input_manager_set_focus(&view->container->node);
 	}
+
+	view_execute_criteria(view);
```
RyanDwyer: The `should_focus` check needs to be done after the view has run its criteria. Otherwise, criteria might move the view to another workspace which would cause it to steal focus.

emersion:
Thanks!

PR #3290, opened 2018-12-14
swaybar: fix cursor scale

### LINKED ISSUE #1750: Swaybar doesn't update cursor on output scale change

RedSoxFan:
Fixes #1750

This fixes a few issues with swaybar's cursor scaling:
1. The cursor scale is now changed when the output scale changes
2. The cursor scale is no longer bound by the max output scale when
swaybar is launched
3. Related to the previous item, the cursor is no longer tiny on low
scale outputs after the max output scale has changed

This also bumps up `wl_compositor` to version 4 to allow usage of `wl_surface_damage_buffer`.

Note: This matches the way swaynag handles cursor scaling so any changes to this PR may also need to be changed for swaynag in a follow-up PR

### DIFF
--- a/swaybar/input.c
+++ b/swaybar/input.c
@@ -55,11 +55,34 @@ static enum x11_button wl_axis_to_x11_button(uint32_t axis, wl_fixed_t value) {
 	}
 }
 
+void update_cursor(struct swaybar *bar) {
+	struct swaybar_pointer *pointer = &bar->pointer;
+	if (pointer->cursor_theme) {
+		wl_cursor_theme_destroy(pointer->cursor_theme);
+	}
+	int scale = pointer->current ? pointer->current->scale : 1;
+	pointer->cursor_theme = wl_cursor_theme_load(NULL, 24 * scale, bar->shm);
+	struct wl_cursor *cursor;
+	cursor = wl_cursor_theme_get_cursor(pointer->cursor_theme, "left_ptr");
+	pointer->cursor_image = cursor->images[0];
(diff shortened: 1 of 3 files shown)


swaybar/input.c
```diff
@@ -55,32 +55,40 @@ static enum x11_button wl_axis_to_x11_button(uint32_t axis, wl_fixed_t value) {
...
+	if (pointer->cursor_theme) {
+		wl_cursor_theme_destroy(pointer->cursor_theme);
+	}
+	int scale = pointer->current ? pointer->current->scale : 1;
+	pointer->cursor_theme = wl_cursor_theme_load(NULL, 24 * scale, bar->shm);
+	struct wl_cursor *cursor;
+	cursor = wl_cursor_theme_get_cursor(pointer->cursor_theme, "left_ptr");
+	pointer->cursor_image = cursor->images[0];
+	wl_surface_set_buffer_scale(pointer->cursor_surface, scale);
+	wl_surface_attach(pointer->cursor_surface,
```
emersion: It'd be better to damage the whole surface with `wl_surface_damage_buffer` since we're updating it.

RedSoxFan: Do I just need to add a call to `wl_surface_damage_bufer` or can you please explain what I should be changing?

emersion: Ah, we probably need to bump `wl_compositor` version to 4.

PR #3291, opened 2018-12-14
Fix focusing on non-visible workspace

### LINKED ISSUE #2625: Closing windows in a stacked layout when on another workspace focuses parent container

RedSoxFan:
Fixes #2625 
#3281 should still be fixed and have no regression

My previous attempt (#3288) was not quite right. Changing the focus stack on a non-visible workspace should only be blocked if the focus would be set to the workspace itself.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -159,14 +159,6 @@ static void handle_seat_node_destroy(struct wl_listener *listener, void *data) {
 		return;
 	}
 
-	if (node->type == N_CONTAINER &&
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Thanks!

PR #3293, opened 2018-12-15
swaynag: damage the cursor surface on update

RedSoxFan:
When the cursor surface gets updated, it should be damaged.

This also bumps up `wl_compositor` to version 4 to be able to use `wl_surface_damage_buffer`.


emersion:
Thanks!

PR #3301, opened 2018-12-17
swaynag: remove double free of details button

### LINKED ISSUE #3300: swaynag: double-free crash
* Sway Version: sway version 1.0-beta.2-39-g1442d4e6 (Dec 17 2018, branch 'master')

* Debug Log: N/A

* Configuration File: N/A

Running swaynag without any arguments crashes. It looks like `swaynag.details.button_details` is appended to the `swaynag.buttons` linked-list, and then double-freed (as it's used there, and not NULLed out in the original pointer). See `main.c:103`

```
jaeger:~ $ gdb `which swaynag`   
GNU gdb (GDB) Fedora 8.2-5.fc29
[... snip .... ]
2018-12-17 13:40:09 - [swaynag/main.c:80] No message passed. Please provide --message/-m

Program received signal SIGABRT, Aborted.
0x00007ffff789f53f in raise () from /lib64/libc.so.6
[... snip ...]
(gdb) bt
#0  0x00007ffff789f53f in raise () at /lib64/libc.so.6
#1  0x00007ffff7889895 in abort () at /lib64/libc.so.6
#2  0x00007ffff78e2927 in __libc_message () at /lib64/libc.so.6
#3  0x00007ffff78e925c in  () at /lib64/libc.so.6
#4  0x00007ffff78ead45 in _int_free () at /lib64/libc.so.6
#5  0x0000000000407b73 in swaynag_destroy (swaynag=0x412b60 <swaynag>) at ../swaynag/swaynag.c:419
#6  0x0000000000404db2 in main (argc=1, argv=0x7fffffffdb08) at ../swaynag/main.c:130
```

RedSoxFan:
Fixes #3300 

If there are no arguments or invalid arguments given, swaynag will free
`swaynag.details.button_details` under the `cleanup` label in main. It
then called `swaynag_destroy`, which would attempt to free it again.

Since `swaynag.details.button_details` is either freed on line 106 of
main (when there is no detailed message) or added to `swaynag.buttons`
on line 103 of main, there is no reason to manually free it in
`swaynag_destroy`.

Although I cannot reproduce a double free on my system, for some reason,
it should have actually resulted in a double free in all code paths.

### DIFF
--- a/swaynag/swaynag.c
+++ b/swaynag/swaynag.c
@@ -416,7 +416,6 @@ void swaynag_destroy(struct swaynag *swaynag) {
 		free(button);
 	}
 	list_free(swaynag->buttons);
-	free(swaynag->details.button_details);
 	free(swaynag->details.message);
 	free(swaynag->details.button_up.text);
 	free(swaynag->details.button_down.text);


emersion:
>I cannot reproduce a double free on my system

Not all libc error out on double-free. Even if they can, they don't always do. To make sure we don't double-free, the best tool is `-Db_sanitize=address,undefined` or Valgrind.

emersion:
Thanks!

PR #3303, opened 2018-12-17
Allow tiling views to be dragged by the titlebar

### LINKED ISSUE #3182: Cannot drag windows with titlebar anymore
1. Try to drag a window's titlebar, without any modifier key
2. Nothing happens, it should move the window

Resizing with borders still works.

RedSoxFan:
Fixes #3182

Enables titling views to be dragged by the titlebar. This is in addition to using the modifier and dragging them from anywhere on the container surface. Floating views already allow this behavior.


emersion:
Thanks!

Issue #1894, opened 2018-05-02
Cursor hiding


ddevault:
Should be able to let users hide the cursor under certain conditions:

- [x] After a timeout
- [ ] ~When they start typing~ (Decided against supporting)
- [ ] ~When they touch the screen~ (Decided against supporting)

lightyou:
+1

PR #3307, opened 2018-12-17
Implement hide_cursor <timeout> command

RedSoxFan:
Supersedes #1979 
Related to #1894

Allows the cursor to be hidden after a specified timeout in milliseconds. Unlike #1979, `handle_activity` will only set the cursor image when it was previously hidden. This version will also iterate over the seats and update the timeouts for the associated cursors when the timeout is changed.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -589,6 +589,23 @@ void cursor_rebase(struct sway_cursor *cursor) {
 	cursor_do_rebase(cursor, time_msec, cursor->previous.node, surface, sx, sy);
 }
 
+static int hide_notify(void *data) {
+	struct sway_cursor *cursor = data;
+	wlr_cursor_set_image(cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
+	cursor->hidden = true;
+	return 1;
+}
+
+static void handle_activity(struct sway_cursor *cursor) {
+	wl_event_source_timer_update(cursor->hide_source,
+			config->hide_cursor_timeout);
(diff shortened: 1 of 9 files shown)


sway/input/cursor.c
```diff
@@ -1229,20 +1246,29 @@ static void handle_request_set_cursor(struct wl_listener *listener,
...
 void cursor_set_image(struct sway_cursor *cursor, const char *image,
 		struct wl_client *client) {
 	if (!(cursor->seat->wlr_seat->capabilities & WL_SEAT_CAPABILITY_POINTER)) {
 		return;
 	}
-	if (!image) {
+	if (cursor->hidden) {
+		return;
+	}
+	if (!image && cursor->image_surface) {
```
emersion: Does hiding the cursor still work?

emersion:
Thanks!

emersion:
Damn, code looked good to me but I forgot [this comment](https://github.com/swaywm/sway/pull/3307#discussion_r243906145) -- what do you think?

RedSoxFan:
> Damn, code looked good to me but I forgot this comment -- what do you think?

I just implemented it the same as the original PR, but it could be either. Honestly, never even thought of it before. If you want, I have no problem changing this to a seat command. That would be the more flexible option since you could specify a timeout per seat, if you wanted. Without looking, I'm not sure if seat wildcards are implemented though (I also have no problem adding those, if they aren't already)

emersion:
Yeah, I think that would be nice.

RedSoxFan:
Ok. I'll create a follow-up PR to switch it over within the next day or two

PR #3308, opened 2018-12-17
Focus node before tiling drag if on titlebar

RedSoxFan:
Before attempting to drag a tiling container by its titlebar, focus it. This fixes clicking on titlebars to focus a container.


gportay:
I have tested it and it worked.

ddevault:
Thanks!

PR #3310, opened 2018-12-18
Return success when renaming a workspace to itself

### LINKED ISSUE #3309: i3 IPC behaviour mismatch: rename workspace fails when using the current name
sway version: 1.0-beta.2-47-ge726b5c4

Reproduction steps:

* Go to some workspace, e.g. workspace `1`
* Run `rename workspace` with the current name, e.g. `swaymsg rename workspace to 1`

Expected result (what i3 does): The command terminates successfully, the workspace retains its name.
Actual result: `Error: Workspace already exists`

A bit of background on why I hit this edge case and care about it being changed: I have a script (with `set -e` so it doesn't continue in case of an error) that lets me choose one of my coding projects, then opens it in my editor and renames the workspace accordingly. When I manually closed my editor after an update and wanted to re-open it with my script, it didn't work because the workspace still had the name of the project I was working on and renaming failed, making the script return early.

RyanDwyer:
Fixes #3309.

Please be sure to check my work thoroughly for backdoors (I'm from Australia).

### DIFF
--- a/sway/commands/rename.c
+++ b/sway/commands/rename.c
@@ -81,8 +81,12 @@ struct cmd_results *cmd_rename(int argc, char **argv) {
 	struct sway_workspace *tmp_workspace = workspace_by_name(new_name);
 	if (tmp_workspace) {
 		free(new_name);
-		return cmd_results_new(CMD_INVALID, "rename",
-				"Workspace already exists");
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3312, opened 2018-12-19
Terminate swaybg in output_disable

RedSoxFan:
Moves the call to `terminate_swaybg` from inside `apply_output_config` to
`output_disable`. The former was only called when an output was being
disabled. The latter is called when an output is being disabled and when
an output becomes disconnected. Without this, disconnecting an enabled
output would result in a defunct swaybg process.

### DIFF
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -225,6 +225,11 @@ void output_disable(struct sway_output *output) {
 
 	root_for_each_container(untrack_output, output);
 
+	if (output->bg_pid) {
+		terminate_swaybg(output->bg_pid);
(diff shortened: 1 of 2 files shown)


sway/tree/output.c
```diff
@@ -225,6 +225,10 @@ void output_disable(struct sway_output *output) {
 
 	root_for_each_container(untrack_output, output);
 
+	if (output->bg_pid) {
+		terminate_swaybg(output->bg_pid);
```
emersion: I think it's important to reset `output->bg_pid` here in case the output is enabled again.

PR #3314, opened 2018-12-20
Allow output identifier in move, focus, and assign

RedSoxFan:
This allows output identifiers to be used with the following commands:
- `move container|window|workspace to output`
- `focus output`
- `assign [criteria] output`

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -194,6 +194,9 @@ static struct cmd_results *focus_output(struct sway_seat *seat,
 	}
 	char *identifier = join_args(argv, argc);
 	struct sway_output *output = output_by_name(identifier);
+	if (!output) {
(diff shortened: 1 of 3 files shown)


sway/commands/focus.c
```diff
@@ -194,6 +194,9 @@ static struct cmd_results *focus_output(struct sway_seat *seat,
 	}
 	char *identifier = join_args(argv, argc);
 	struct sway_output *output = output_by_name(identifier);
+	if (!output) {
+		output = output_by_identifier(identifier);
```
emersion: Maybe we could replace `output_by_{name,identifier}` by a single function? Don't know if we use one without the other somewhere.

RedSoxFan:
Superseded by #3315

PR #3315, opened 2018-12-20
Combine output_by_name and output_by_identifier

RedSoxFan:
Supersedes #3314

This combines `output_by_name` and `output_by_identifier` into a single
function called `output_by_name_or_id`. This allows for output
identifiers to be used in all commands, simplifies the logic of the
callers, and is more efficient since worst case is a single pass through
the output list.


emersion:
Thanks!

PR #3316, opened 2018-12-20
swaylock: Add some key bindings to emulate login(1)

milkey-mouse:
There are a few key combos that do something in the standard `login` program that don't work with swaylock (sabotaging my muscle memory :wink:). Namely, Ctrl-C restarts login (i.e. clears the password buffer) and Ctrl-D (effectively EOF) submits it.

See the commit messages for more detail.


emersion:
Thanks!

Issue #3311, opened 2018-12-18
Sway crashes when using the IRC app Polari


mwenzkowski:
* Sway Version: 1.0-beta.2-28-gb61a936c (Dec 16 2018, branch 'master')
* Debug Log: [sway.log](https://gist.github.com/mwenzkowski/9e528a476a6064eca16f9fc8c722b9d4#file-sway-log)
* Configuration File: Default configuration
* Stack Trace: [stack trace](https://gist.github.com/mwenzkowski/9e528a476a6064eca16f9fc8c722b9d4#file-stack-trace)

This crash occurred while I was trying out the IRC application [Polari](https://wiki.gnome.org/Apps/Polari) (Version  3.30.2) on Arch Linux.
To reproduce the problem:

0. (`rm .cache/polari/close-confirmation-shown`)
1. Start Polari
2. Close Polari (Either via key binding or by clicking the close button on the CSD)
3. A dialog window opens, asking if Polari should be closed or if it should keep running in background
4. Press the escape key
5. The main window of Polari vanishes, but the dialog window remains (which I guess is a bug in Polari, since it should be the other way round)
6. Move the mouse cursor on the dialog window, then sway crashes

The dialog window is only shown on the first run, so to reproduce it another time step 0. is necessary. I tried this five times and sway crashed every time.

emersion:
Can you try to reproduce after updating wlroots?

mwenzkowski:
I have build both wlroots (commit swaywm/wlroots@c4c1d63a0c0d7315a1a4468d62d883371ba3ec4e) and sway (commit 477bca5e288cc3155c536265272490b413328778) from source and I can still reproduce the crash.
I also tried it in rootston and there it doesn't crash.

emersion:
Can you post a `WAYLAND_DEBUG=1` debug log of Polari?

mwenzkowski:
Here is the log: [polari.log](https://gist.github.com/mwenzkowski/9e528a476a6064eca16f9fc8c722b9d4#file-polari-log).
This time I noticed that sway crashes as soon as the mouse cursor enters the dialog window. If the cursor was on the dialog window already, the crash occurs immediately when the mouse is moved.
I have updated the description above accordingly.

PR #3323, opened 2018-12-22
Check xdg_surface's role before using its toplevel

mwenzkowski:
After some debugging I found the reason for the crash in #3311. The problem was
that I ended up with a toplevel child whos parent had unmapped.  So the child's
`xdg_surface->toplevel->parent` pointed to an `xdg_surface` with its role set
to `WLR_XDG_SURFACE_ROLE_NONE` and its `xdg_surface->toplevel` pointer freed.
The function `surface_is_popup` didn't check this case, which led to a segfault
when setting `xdg_surface = xdg_surface->toplevel->parent;`.  I searched for
similar occurrences and found one in the function `is_transient_for`, where I
also added a check (I was able to reproduce a segfault there too).

Despite that this patch fixes my crash, I know too little about how xdg shell
works, to be confident that this is the right fix. I don't know if it is normal
that a child points to its parent, even if the parent has unmapped. So I think
someone with a better understanding of xdg shell should check if this patch is
the right approach.

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -216,7 +216,7 @@ static bool is_transient_for(struct sway_view *child,
 		return false;
 	}
 	struct wlr_xdg_surface *surface = child->wlr_xdg_surface;
-	while (surface) {
+	while (surface && surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL) {
 		if (surface->toplevel->parent == ancestor->wlr_xdg_surface) {
 			return true;
 		}
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -341,7 +341,7 @@ static bool surface_is_popup(struct wlr_surface *surface) {
 	if (wlr_surface_is_xdg_surface(surface)) {
 		struct wlr_xdg_surface *xdg_surface =
 			wlr_xdg_surface_from_wlr_surface(surface);
-		while (xdg_surface) {
+		while (xdg_surface && xdg_surface->role != WLR_XDG_SURFACE_ROLE_NONE) {
 			if (xdg_surface->role == WLR_XDG_SURFACE_ROLE_POPUP) {
 				return true;
 			}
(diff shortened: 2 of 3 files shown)


sway/desktop/xdg_shell.c
```diff
@@ -217,7 +217,7 @@ static bool is_transient_for(struct sway_view *child,
 		return false;
 	}
 	struct wlr_xdg_surface *surface = child->wlr_xdg_surface;
-	while (surface) {
+	while (surface && surface->role != WLR_XDG_SURFACE_ROLE_NONE) {
```
emersion: I think we could change these checks to be `surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL`.

emersion:
>So the child's xdg_surface->toplevel->parent pointed to an xdg_surface with its role set
to WLR_XDG_SURFACE_ROLE_NONE and its xdg_surface->toplevel pointer freed.

Hmm, that's annoying. We should probably remove all children before setting the role to WLR_XDG_SURFACE_ROLE_NONE in wlroots.

sway/tree/container.c
```diff
@@ -349,7 +349,7 @@ static bool surface_is_popup(struct wlr_surface *surface) {
 	if (wlr_surface_is_xdg_surface_v6(surface)) {
 		struct wlr_xdg_surface_v6 *xdg_surface_v6 =
 			wlr_xdg_surface_v6_from_wlr_surface(surface);
-		while (xdg_surface_v6) {
+		while (xdg_surface_v6 && xdg_surface_v6->role == WLR_XDG_SURFACE_V6_ROLE_NONE) {
```
RyanDwyer: This is wrong. It should be `!=`, or `== WLR_XDG_SURFACE_ROLE_TOPLEVEL` as mentioned in another comment.

emersion:
>We should probably remove all children before setting the role to WLR_XDG_SURFACE_ROLE_NONE in wlroots.

In fact, no. The `parent` argument is an `xdg-surface`, so it doesn't make sense to reset it when the `xdg-toplevel` is destroyed.

emersion:
Thanks!

PR #3327, opened 2018-12-23
Added is_fullscreen command to swaymsg

NokiDev:
I needed to check if the current window were fullscreen or not, suprised by not seeing it in swaymsg API, I added it. 

The main purpose was to lock my screen after timeout only if I wasn't watching a movie or a youtube video (in fullscreen). 

With this patch, we can now customize a little wiki's swayidle command : 

```
exec swayidle \
    timeout 300 'if (swaymsg -t is_fullscreen | grep "false"); then swaylock ;fi' \
    timeout 600 'if (swaymsg -t is_fullscreen | grep "false"); then swaymsg "output * dpms off"' \
       resume 'swaymsg "output * dpms on"' \
    before-sleep 'swaylock'
```

What left todo, is to add in swaymsg's man docs is_fullscreen capabilities. 
I didn't do it since I don't know how this works.

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -836,6 +836,22 @@ void ipc_client_handle_command(struct ipc_client *client) {
 		goto exit_cleanup;
     }
 
+	case IPC_IS_FULLSCREEN:
+	{
+		json_object *json = json_object_new_object();
+		struct sway_container *container = config->handler_context.container;
+		if(container != NULL && container->is_fullscreen) {
+			json_object_object_add(json, "fullscreen", json_object_new_string("true"));
+		}
+		else {
+			json_object_object_add(json, "fullscreen", json_object_new_string("false"));
+		}
+		const char *json_string = json_object_to_json_string(json);
+		client_valid =
+			ipc_send_reply(client, json_string, (uint32_t)strlen(json_string));
+		json_object_put(json); // free
+		goto exit_cleanup;
+	}
 	default:
 		wlr_log(WLR_INFO, "Unknown IPC command type %i", client->current_command);
 		goto exit_cleanup;
@@ -860,7 +876,7 @@ bool ipc_send_reply(struct ipc_client *client, const char *payload, uint32_t pay
 	memcpy(&data32[1], &client->current_command, sizeof(client->current_command));
 
 	while (client->write_buffer_len + ipc_header_size + payload_length >=
-				 client->write_buffer_size) {
+		client->write_buffer_size) {
 		client->write_buffer_size *= 2;
 	}
 
--- a/swaymsg/main.c
+++ b/swaymsg/main.c
@@ -389,7 +389,9 @@ int main(int argc, char **argv) {
 		type = IPC_COMMAND;
 	} else if (strcasecmp(cmdtype, "get_workspaces") == 0) {
 		type = IPC_GET_WORKSPACES;
-	} else if (strcasecmp(cmdtype, "get_seats") == 0) {
+	} else if (strcasecmp(cmdtype, "is_fullscreen") == 0) {
+        type = IPC_IS_FULLSCREEN; 
+    } else if (strcasecmp(cmdtype, "get_seats") == 0) {
 		type = IPC_GET_SEATS;
 	} else if (strcasecmp(cmdtype, "get_inputs") == 0) {
 		type = IPC_GET_INPUTS;
--- a/include/ipc.h
+++ b/include/ipc.h
@@ -21,6 +21,8 @@ enum ipc_command_type {
 	IPC_GET_INPUTS = 100,
 	IPC_GET_SEATS = 101,
 
+	IPC_IS_FULLSCREEN = 102,
+
 	// Events sent from sway to clients. Events have the highest bits set.
 	IPC_EVENT_WORKSPACE = ((1<<31) | 0),
 	IPC_EVENT_OUTPUT = ((1<<31) | 1),


emersion:
I think you can already do this with `swaymsg -t get_tree` and `jq`.

NokiDev:
Hi, maybe, but relying on what ? window size ? + Scale factors + screen on which it is displayed ? 
 
It would be simple to get it directly as sway keeps a state on whether the container is fullscreen or not. 

However, adding it explicitly in the output of get_tree may be a better solution

emersion:
Hmm. i3 has an undocumented property `fullscreen_mode` (see https://github.com/i3/i3/blob/32188d7b2c84f880f9d1ef78e9119b7e8cdfce21/src/ipc.c#L582). I'd prefer to add support for this instead of a `is_fullscreen` command.

emersion:
We could just add the property to the IPC, without implementing other fullscreen modes.

NokiDev:
Hmm, okay, will look into it.

sway/ipc-json.c
```diff
@@ -421,6 +421,7 @@ static void ipc_json_describe_container(struct sway_container *c, json_object *o
 		view_is_urgent(c->view) : container_has_urgent_child(c);
 	json_object_object_add(object, "urgent", json_object_new_boolean(urgent));
 	json_object_object_add(object, "sticky", json_object_new_boolean(c->is_sticky));
+	json_object_object_add(object, "fullscreen_mode", json_object_new_boolean(c->is_fullscreen));
```
emersion: In i3, this is an integer, not a boolean.

sway/ipc-server.c
```diff
@@ -860,7 +860,7 @@ bool ipc_send_reply(struct ipc_client *client, const char *payload, uint32_t pay
 	memcpy(&data32[1], &client->current_command, sizeof(client->current_command));
 
 	while (client->write_buffer_len + ipc_header_size + payload_length >=
-				 client->write_buffer_size) {
+		client->write_buffer_size) {
```
emersion: Missing one indentation level

swaymsg/main.c
```diff
@@ -389,7 +389,7 @@ int main(int argc, char **argv) {
 		type = IPC_COMMAND;
 	} else if (strcasecmp(cmdtype, "get_workspaces") == 0) {
 		type = IPC_GET_WORKSPACES;
-	} else if (strcasecmp(cmdtype, "get_seats") == 0) {
+    } else if (strcasecmp(cmdtype, "get_seats") == 0) {
```
emersion: Extra indentation level

sway/ipc-json.c
```diff
@@ -421,6 +421,7 @@ static void ipc_json_describe_container(struct sway_container *c, json_object *o
 		view_is_urgent(c->view) : container_has_urgent_child(c);
 	json_object_object_add(object, "urgent", json_object_new_boolean(urgent));
 	json_object_object_add(object, "sticky", json_object_new_boolean(c->is_sticky));
+        json_object_object_add(object, "fullscreen_mode", json_object_new_int(c->is_fullscreen));
```
RedSoxFan: The correct indentation here should be a single tab

sway/ipc-server.c
```diff
@@ -860,7 +860,7 @@ bool ipc_send_reply(struct ipc_client *client, const char *payload, uint32_t pay
 	memcpy(&data32[1], &client->current_command, sizeof(client->current_command));
 
 	while (client->write_buffer_len + ipc_header_size + payload_length >=
-				 client->write_buffer_size) {
+                        client->write_buffer_size) {
```
RedSoxFan: The indentation here is wrong, but there is also no need to alter this file.

swaymsg/main.c
```diff
@@ -389,7 +389,7 @@ int main(int argc, char **argv) {
 		type = IPC_COMMAND;
 	} else if (strcasecmp(cmdtype, "get_workspaces") == 0) {
 		type = IPC_GET_WORKSPACES;
-	} else if (strcasecmp(cmdtype, "get_seats") == 0) {
+        } else if (strcasecmp(cmdtype, "get_seats") == 0) {
```
RedSoxFan: The indentation here is wrong, but there is also no need to alter this file

sway/ipc-json.c
NokiDev: Done

sway/ipc-server.c
NokiDev: Removes from commit.

swaymsg/main.c
NokiDev: Removed from commit as well.

swaymsg/main.c
NokiDev: Removed from commit.

sway/ipc-server.c
NokiDev: Removed from commit.

sway/ipc-server.c
```diff
@@ -835,7 +835,7 @@ void ipc_client_handle_command(struct ipc_client *client) {
 		json_object_put(json); // free
 		goto exit_cleanup;
     }
-
+
```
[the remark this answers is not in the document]
NokiDev: removed from commit.

sway/ipc-json.c
NokiDev: Done, by using json_object_new_int
1 = fullscreen on
0 = fullscreen off

sway/ipc-json.c
```diff
@@ -421,6 +421,7 @@ static void ipc_json_describe_container(struct sway_container *c, json_object *o
 		view_is_urgent(c->view) : container_has_urgent_child(c);
 	json_object_object_add(object, "urgent", json_object_new_boolean(urgent));
 	json_object_object_add(object, "sticky", json_object_new_boolean(c->is_sticky));
+    json_object_object_add(object, "fullscreen_mode", json_object_new_int(c->is_fullscreen));
```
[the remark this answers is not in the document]
NokiDev: Done

PR #3332, opened 2018-12-24
sway(5): document tiling_drag

RedSoxFan:
Adds documentation in sway(5) for the tiling_drag command

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -558,6 +558,13 @@ The default colors are:
 	Set the opacity of the window between 0 (completely transparent) and 1
 	(completely opaque).
 
+*tiling\_drag*  enable|disable|toggle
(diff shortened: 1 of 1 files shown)


sway/sway.5.scd
```diff
@@ -558,6 +558,13 @@ The default colors are:
 	Set the opacity of the window between 0 (completely transparent) and 1
 	(completely opaque).
 
+*tiling\_drag*  enable|disable|toggle
+	Sets whether or not tiling containers can be dragged with the mouse. If
+	enabled (default), the _floating_mod_ can be used to drag tiling, as well
```
RyanDwyer: Doesn't the middle underscore need to be escaped? Also again two lines down.

RedSoxFan: Oops. Yes, they do. Fixed.

emersion:
Thanks!

PR #3334, opened 2018-12-27
Revamp seat configs

RedSoxFan:
This makes seat configs work like output and input configs do. This also
adds support for wildcard seat configs. A seat config is still created
in the main seat command handler, but instead of creating a new one in
the subcommands and destroying the main seat command's instance, the
seat subcommands modify the main one. The seat config is then stored,
where it is merged appropriately. The seat config returned from
`store_seat_config` is then applied. When attempting to apply a wildcard
seat config, a seat specific config is queried for and if found, that is
used. Otherwise, the wildcard config is applied directly.

Additionally, instead of adding input devices to the default seat
directly when there is no seat configs, a seat config for the default
seat is created with only fallback set to true, which is more explicit.
It also fixes an issue where running a seat command at runtime (with no
seat config in the sway config), would result in all input devices being
removed from the default seat and leaving sway in an unusable state.

Also, instead of checking for any seat config, the search is for a seat
config with a fallback option seat. This makes it so if there are only
seat configs with fallback set to -1, the default seat is still created
since there is no explicit notion on what to do regarding fallbacks.
However, if there is even a single fallback 0, then the default seat is
not used as a fallback. This will be needed for seat subcommands like
hide_cursor where the user may only want to set that property without
effecting anything else.

Follow-up PRs:
- Move hide_cursor to be a seat subcommand
- Fix seat cursor to work on the specified seat instead of the current

### DIFF
--- a/sway/commands/seat.c
+++ b/sway/commands/seat.c
@@ -26,9 +26,16 @@ struct cmd_results *cmd_seat(int argc, char **argv) {
 
 	struct cmd_results *res = config_subcommand(argv + 1, argc - 1,
 			seat_handlers, sizeof(seat_handlers));
+	if (res && res->status != CMD_SUCCESS) {
+		free_seat_config(config->handler_context.seat_config);
+		config->handler_context.seat_config = NULL;
+		return res;
+	}
 
-	free_seat_config(config->handler_context.seat_config);
+	struct seat_config *sc =
+		store_seat_config(config->handler_context.seat_config);
+	input_manager_apply_seat_config(sc);
+
 	config->handler_context.seat_config = NULL;
-
-	return res;
+	return cmd_results_new(CMD_SUCCESS, NULL, NULL);
 }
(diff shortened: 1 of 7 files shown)


sway/commands/seat.c
```diff
@@ -26,9 +26,16 @@ struct cmd_results *cmd_seat(int argc, char **argv) {
...
+		free_seat_config(config->handler_context.seat_config);
+		config->handler_context.seat_config = NULL;
+		return res;
+	}
 
-	free_seat_config(config->handler_context.seat_config);
-	config->handler_context.seat_config = NULL;
+	struct seat_config *sc =
+		store_seat_config(config->handler_context.seat_config);
+	input_manager_apply_seat_config(sc);
```
emersion: It seems like `sc` is not free'd.

RedSoxFan: `sc` is stored in `config->seat_configs` in `store_seat_config` and should not be free'd here

emersion:
Thanks!

PR #3336, opened 2018-12-27
sway-output(5): doc scaling consideration for pos

### LINKED ISSUE #3268: Sway doesn't merge outputs with negative position
I want to have configs on my laptop for every monitor that I connect at home/work. So, I want my main screen to be at pos 0,0 while other screens are placed above like pos 0,-2160.

It simply doesn't work - I can't reach top screen over the top edge of my laptop with cursor.

sway version: Dec 2 2018, branch '' - actually it's a RPM package sway-1.0-0.beta.2+5.g7702943e.fc29.x86_64 from here: https://copr.fedorainfracloud.org/coprs/rzl/sway/

RedSoxFan:
Closes #3268 
Wiki documentation: https://github.com/swaywm/sway/wiki#hidpi

This copies the information regarding positioning outputs when there
is scaling involved from the wiki to sway-output(5).

### DIFF
--- a/sway/sway-output.5.scd
+++ b/sway/sway-output.5.scd
@@ -36,7 +36,19 @@ must be separated by one space. For example:
 
 *output* <name> position|pos <X> <Y>
 	Places the specified output at the specific position in the global
-	coordinate space.
+	coordinate space. If scaling is active, it has to be considered when
+	positioning. For example, if the scaling factor for the left output is 2,
+	the relative position for the right output has to be divided by 2.
+
+	Example:
+
+		output HDMI1 scale 2
+
+		output HDMI1 pos 0 0 res 3200x1800
+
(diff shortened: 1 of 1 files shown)


sway/sway-output.5.scd
```diff
@@ -36,7 +36,19 @@ must be separated by one space. For example:
...
 	Places the specified output at the specific position in the global
-	coordinate space.
+	coordinate space. If scaling is active, it has to be considered when
+	positioning. For example, if the scaling factor for the left output is 2,
+	the relative position for the right output has to be divided by 2.
+
+	Example:
+
+		output HDMI1 scale 2
+
```
emersion: Are double newlines are necessary in code blocks?

RedSoxFan: Without the double newlines, it's just all on a single line when viewing the resulting man page

PR #3337, opened 2018-12-27
seat_cmd_cursor: work on seat name provided

RedSoxFan:
~**DO NOT MERGE UNTIL AFTER #3334**~

Instead of simulating events on the current seat, this makes it so
seat_cmd_cursor respects the seat name provided by
`seat <name> cursor <args>`. It also adds support for simulating
events on all seats when the wildcard is given.

This also defers the command when reading the config, which allows the
user to set the initial position of the cursor when the command is
included in the config file.


RedSoxFan:
Rebased and ready for review

emersion:
Thanks!

Issue #2576, opened 2018-09-05
Mouse bindings don't seem to work


jplatte:
sway version: 1.0-alpha.5-243-gaa2bf98e (Sep  5 2018, branch 'master')

What I've tried (per https://github.com/swaywm/sway/pull/2303, since mouse bindings don't seem to be documented):

```
bindsym button6 workspace prev_on_output
bindsym Mod4+button1 exec $term
bindsym button1+button3 kill
```

None of these key/button bindings seem to have any effect.

milkmiruku:
`sway r5204.4a219818-1`

These don't work;
```
bindsym  --whole-window button9 kill
bindsym  --whole-window Mod4+button8 move position center
```

PR #3339, opened 2018-12-28
Remove button from state on release during op

RedSoxFan:
This fixes a bug in `dispatch_cursor_button` where if there was an
operation occurring, the button would not be removed from the state on
release. This resulted in the button appearing to be permanently pressed
and caused mouse bindings to not match correctly.

This was probably part of the problem for https://github.com/swaywm/sway/issues/2576

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -825,6 +825,11 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 			seat_end_mouse_operation(seat);
 			seat_pointer_notify_button(seat, time_msec, button, state);
 		}
+		if (state == WLR_BUTTON_PRESSED) {
+			state_add_button(cursor, button);
+		} else {
+			state_erase_button(cursor, button);
+		}
 		return;
 	}
 


sway/input/cursor.c
```diff
@@ -822,6 +822,7 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 	// Handle existing seat operation
 	if (cursor->seat->operation != OP_NONE) {
 		if (button == cursor->seat->op_button && state == WLR_BUTTON_RELEASED) {
+			state_erase_button(cursor, button);
```
RyanDwyer: Won't this still go into an unexpected state if I press or release another button during a seat operation? By another button I mean a button that isn't `cursor->seat->op_button`.

For example:

    if (button == cursor->seat->op_button && state == WLR_BUTTON_RELEASED) {
        ...
    }
    if (state == WLR_BUTTON_PRESSED) {
        state_add_button(cursor, button);
    } else {
        state_erase_button(cursor, button);
    }
    return;

RedSoxFan: Actually, it wouldn't since the state just wouldn't change during an operation and state is only used for bindings, but I think this is a better approach anyway.

sway/input/cursor.c
```diff
@@ -825,6 +825,11 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 			seat_end_mouse_operation(seat);
 			seat_pointer_notify_button(seat, time_msec, button, state);
 		}
+		if (state == WLR_BUTTON_PRESSED) {
+			state_add_button(cursor, button);
+		} else {
+			state_remove_button(cursor, button);
```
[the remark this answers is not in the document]
RedSoxFan: Derp. Fixed

emersion:
Thanks!

PR #3313, opened 2018-12-19
Change mouse buttons to x11 map and libevdev names

RedSoxFan:
Closes #2576
Related to #3278

This modifies the way mouse bindings are parsed. Instead of adding to
`BTN_LEFT`, which results in button numbers that may not be expected,
buttons will be parsed in one of the following ways:

1. `button[1-9]` will now map to their x11 equivalents. This is already
the case for bar bindings. This adds support for binding to axis events,
which was not possible in the previous approach.

2. Anything that starts with `BTN_` will be parsed as an event code name
using `libevdev_event_code_from_name`. This allows for any button to be
mapped to instead of limiting usage to the ones near `BTN_LEFT`. This also
adds a dependency on libevdev, but since libevdev is already a dependency
of libinput, this should be fine. If needed, this option can have dependency
guards added.

Binding changes:
- button1: `BTN_LEFT` -> `BTN_LEFT`
- button2: `BTN_RIGHT` -> `BTN_MIDDLE`
- button3: `BTN_MIDDLE` -> `BTN_RIGHT`
- button4: `BTN_SIDE` -> ScrollUp
- button5: `BTN_EXTRA` -> ScrollDown
- button6: `BTN_FORWARD` -> ScrollLeft
- button7: `BTN_BACK` -> ScrollRight
- button8: `BTN_TASK` -> `BTN_SIDE`
- button9: `BTN_JOYSTICK` -> `BTN_EXTRA`

Since the axis events need to be mapped to an event code, this uses the
following mappings to avoid any conflicts:
- ScrollUp: `KEY_MAX + 1`
- ScrollDown: `KEY_MAX + 2`
- ScrollLeft: `KEY_MAX + 3`
- ScrollRight: `KEY_MAX + 4`

To be done in follow-up PRs:
- Add support for `libevdev_event_code_from_name` to:
    - `bar_cmd_bindsym`
    - `input_cmd_scroll_button`
- Validate and allow event codes with  `libevdev_event_code_get_name` in:
    -  `cmd_bindcode`
    -  `bar_cmd_bindcode`
    -  `input_cmd_scroll_button` (already uses event codes, just needs validation)
- Map non-axis x11 buttons to event codes in `input_cmd_scroll_button` for consistency/convenience

PR #3341, opened 2018-12-28
Improve mouse button parsing: helpers and bind{code/sym}

RedSoxFan:
This is the first of five follow-up PRs to #3313. This extracts the parsing of
mouse buttons into helpers and then uses them for `bindcode` and `bindsym`.
The next four PRs are for `input scroll_button`, `seat cursor`, `bar bindsym`,
and `bar tray_bindsym`.

The following helper functions have been added to aid with parsing mouse
buttons from a string:

1. `get_mouse_bindsym`: attempts to parse the string as an x11 button
(button[1-9]) or as an event name (ex BTN_LEFT or BTN_SIDE)
2. `get_mouse_bindcode`: attempts to parse the string as an event code
and validates that the event code is a button (starts with `BTN_`).
3. `get_mouse_button`: this is a conveniency function for callers that
do not care whether a bindsym or bindcode are used and attempts to parse
the string as a bindsym and then bindcode.

This modifies `bindcode` and `bindsym` to use `get_mouse_bindcode` and
`get_mouse_bindsym`, respectively, to parse mouse buttons. Additionally,
the `BINDING_MOUSE` type has been split into `BINDING_MOUSECODE` and
`BINDING_MOUSESYM` to match keys and allow for mouse bindcodes to be
used. Between the two commands, all button syms and codes should be
supported, including x11 axis buttons.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -2,8 +2,10 @@
 #include <math.h>
 #include <libevdev/libevdev.h>
 #include <linux/input-event-codes.h>
+#include <errno.h>
 #include <float.h>
 #include <limits.h>
+#include <strings.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_xcursor_manager.h>
 #include <wlr/types/wlr_idle.h>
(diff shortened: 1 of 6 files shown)


mstoeckl:
Just passing by; two comments:
* In this and following PRs, `get_mouse_*` are called with nonnull `error`, in which case the branches testing `if (error)` are unnecessary.
* The code does not account for memory allocation failure, rare as it may be. In general, one should not signal errors using heap allocated values, lest `strdup` or `malloc` return NULL, and the caller falsely assume success. One workaround is a more cumbersome signature, such as `bool parse_x(const char* str, uint32_t* ret, char** err_msg)`.

RedSoxFan:
>  In this and following PRs, `get_mouse_*` are called with nonnull `error`, in which case the branches testing `if (error)` are unnecessary.

I originally designed it to make the human readable error message optional. Since all the current callers do request the error message, I have removed the NULL-checks. If the caller doesn't want it, they can just immediately free it.

>  The code does not account for memory allocation failure, rare as it may be. In general, one should not signal errors using heap allocated values, lest `strdup` or `malloc` return NULL, and the caller falsely assume success. One workaround is a more cumbersome signature, such as `bool parse_x(const char* str, uint32_t* ret, char** err_msg)`.

I slightly modified the calling code (for this PR -- I'll update the rest once this is merged) to check for a return value of `0`, which denotes a failure anyway. I also added NULL-checks around the calls to `snprintf` after calls to`malloc`.

sway/input/cursor.c
```diff
@@ -1520,3 +1522,66 @@ void cursor_warp_to_workspace(struct sway_cursor *cursor,
...
+uint32_t get_mouse_bindsym(const char *name, char **error) {
+	if (strncasecmp(name, "button", strlen("button")) == 0) {
+		// Map to x11 mouse buttons
+		int number = name[strlen("button")] - '0';
+		if (number < 1 || number > 9 || strlen(name) > strlen("button0")) {
+			*error = strdup("Only buttons 1-9 are supported. For other mouse "
+					"buttons, use the name of the event code.");
+			return 0;
+		}
+		uint32_t buttons[] = {BTN_LEFT, BTN_MIDDLE, BTN_RIGHT,
```
ianyfan: Maybe this could be `static const`?

RedSoxFan: Changed

ianyfan:
Thanks!

PR #3344, opened 2018-12-28
Improve mouse button parsing: bar mouse bindings

RedSoxFan:
~**DO NOT MERGE UNTIL AFTER #3341**~

This is the fourth in a series of follow-up PRs for #3313.

This modifies `bar_cmd_bindsym` to use `get_mouse_bindsym` for parsing
mouse buttons. This also introduces `cmd_bar_bindcode`, which will use
`get_mouse_bindcode` for parsing mouse buttons. Like sway bindings, the
two commands are encapsulated in a single file with shared code.

This also modifies swaybar to operate off of event codes rather than x11
button numbers, which allows for any mouse button to be used. Tray bindings
will be modified in a separate PR.

This introduces two new IPC properties:
- For `get_bar_config`, `event_code` has been added to the `bindings`
section and will include to event code for the button. If the event code
can be mapped to a x11 button, `input_code` will still be the x11 button
number. Otherwise, `input_code` will be `0`.
- Likewise for `click_events`, `event` has been added and will include
the event code for the button clicked. If the event code can be mapped
to a x11 button, `button` will still be the x11 button number.
Otherwise, `button` will be `0`.


ddevault:
Thanks!

PR #3348, opened 2018-12-30
Verify seat fallback settings on reload

### LINKED ISSUE #3347: Config reloading broken
Config reloading worked fine on commit:
https://github.com/swaywm/sway/commit/f50d1615748d6169da773ed8bd4679eac87f0f99

On latest commit:
https://github.com/swaywm/sway/commit/3e8f548d1d0acb3aba384842615e6528b3027283

Reloading the config results in all your windows disappearing and sway seemingly stops responding to keyboard input. It isn't "frozen" however since the status bar keeps updating. Can't log out of sway using any of my usual keybindings.

So one of those 6 commits likely is the culprit. I'll update this issue when I get some time to look at each more closely.

RedSoxFan:
Fixes #3347 

This fixes an issue where on reload, all input devices that were added
via an implicit fallback to the default seat would be removed from the
default seat and applications would crash due to the seat having no
capabilities.

On reload, there is a query for a seat config with the fallback setting
set (it can either be true or false). If no such seat config exists, the
default seat is created (if needed) and has the implicit fallback true
applied to its seat config. This is the same procedure that occurs when
a new input is detected.

### DIFF
--- a/include/sway/input/input-manager.h
+++ b/include/sway/input/input-manager.h
@@ -44,6 +44,12 @@ struct sway_seat *input_manager_get_default_seat(void);
 struct sway_seat *input_manager_get_seat(const char *seat_name);
 
 /**
+ * If none of the seat configs have a fallback setting (either true or false),
+ * create the default seat (if needed) and set it as the fallback
+ */
+void input_manager_verify_fallback_seat(void);
+
+/**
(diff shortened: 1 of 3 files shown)


emersion:
Thanks!

PR #3349, opened 2018-12-30
Set font options when computing text width.

### LINKED ISSUE #2869: Window title is cut off at the end with pango fonts
This happens in particular with my default font, Encode Sans, but it can be reproduced with long titles in other fonts, e.g. Liberation Sans as in the minimal config below. Here's a screenshot:

screenshot

`sway version 1.0-alpha.6-300-g103b7bc4 (Oct 18 2018, branch 'makepkg')`

Config and log: https://gist.github.com/jplatte/b9556b716d4cc9f30cb9da66035f2f3d

johnchen902:
Fix #2869

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -453,19 +453,26 @@ static void update_title_texture(struct sway_container *con,
 	int width = 0;
 	int height = con->title_height * scale;
 
-	cairo_t *c = cairo_create(NULL);
+	// We must use a non-nil cairo_t for cairo_set_font_options to work.
+	// Therefore, we cannot use cairo_create(NULL).
+	cairo_surface_t *dummy_surface = cairo_image_surface_create(
+			CAIRO_FORMAT_ARGB32, 0, 0);
+	cairo_t *c = cairo_create(dummy_surface);
+	cairo_set_antialias(c, CAIRO_ANTIALIAS_BEST);
+	cairo_font_options_t *fo = cairo_font_options_create();
+	cairo_font_options_set_hint_style(fo, CAIRO_HINT_STYLE_FULL);
+	cairo_font_options_set_antialias(fo, CAIRO_ANTIALIAS_SUBPIXEL);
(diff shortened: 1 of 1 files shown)


sway/tree/container.c
```diff
@@ -453,19 +453,25 @@ static void update_title_texture(struct sway_container *con,
 	int width = 0;
 	int height = con->title_height * scale;
 
-	cairo_t *c = cairo_create(NULL);
+	cairo_surface_t *fake_surface = cairo_image_surface_create(
+			CAIRO_FORMAT_ARGB32, 0, 0);
+	// somehow using NULL failed to fix #2869
```
emersion: Is this a cairo bug?

johnchen902: `cairo_create(NULL)` returns a nil:

> ```
> cairo_t *
> cairo_create (cairo_surface_t *target)
> {
>     if (unlikely (target == NULL))
> 	return _cairo_create_in_error (_cairo_error (CAIRO_STATUS_NULL_POINTER));
> ```

`cairo_set_font_options(nil)` is a no-op and `cairo_get_font_options(nil)` returns default font options.

> ```
> void
> cairo_set_font_options (cairo_t                    *cr,
> 			const cairo_font_options_t *options)
> {
>     cairo_status_t status;
> 
>     if (unlikely (cr->status))
> 	return;
> ```

> ```
> void
> cairo_get_font_options (cairo_t              *cr,
> 			cairo_font_options_t *options)
> {
>     /* check that we aren't trying to overwrite the nil object */
>     if (cairo_font_options_status (options))
> 	return;
> 
>     if (unlikely (cr->status)) {
> 	_cairo_font_options_init_default (options);
> 	return;
>     }
> ```

Therefore, I think such behavior is by-design. I'll update the (in-code) comment.

RedSoxFan: Related: https://lists.cairographics.org/archives/cairo/2005-July/004448.html

RedSoxFan:
Thanks!

PR #3353, opened 2018-12-31
Apply implicit fallback seat config

RedSoxFan:
#3348 did not completely fix the issue of input devices being
removed from the implicit fallback on reload. This should handle
the remaining case(s).

The implicit fallback seat config needs to be applied (if created).
Otherwise, the input devices will still be removed from the implicit
default seat on reload when there is any seat config.


emersion:
Thanks!

PR #3354, opened 2018-12-31
swaybar: fix double free

eiz:
status->text should not be freed here. There are two scenarios:

* status->text has been set to an error by status_error. In this case
  the value shouldn't be freed because it's always a reference to a
  constant.
* status->text has been set to status->buffer because the bar is in
  text protocol mode. In this case it's a double free because the
  buffer is already freed after.


RedSoxFan:
Thanks!

PR #3358, opened 2019-01-01
Implement tiling_drag_threshold

RedSoxFan:
This is designed to replicate https://github.com/i3/i3/pull/3559 for sway. I was going to wait for that PR to be reviewed and merged, but decided to just do it now. Unlike that PR, this one allows the user to configure the threshold to their liking.

Implements `tiling_drag_threshold <threshold>` to prevent accidental
dragging of tiling containers. If a container (and all of its
descendants) are unfocused and the tile bar is pressed, a threshold
will be used before actually starting the drag. Once the threshold has
been exceeded, the cursor will change to the grab icon and the operation
will switch from `OP_MOVE_TILING_THRESHOLD` to `OP_MOVE_TILING`.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -172,6 +172,7 @@ sway_cmd cmd_swaybg_command;
 sway_cmd cmd_swaynag_command;
 sway_cmd cmd_swap;
 sway_cmd cmd_tiling_drag;
+sway_cmd cmd_tiling_drag_threshold;
 sway_cmd cmd_title_align;
 sway_cmd cmd_title_format;
 sway_cmd cmd_titlebar_border_thickness;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -426,7 +426,9 @@ struct sway_config {
 	bool auto_back_and_forth;
 	bool show_marks;
 	enum alignment title_align;
+
 	bool tiling_drag;
+	int tiling_drag_threshold;
 
 	bool smart_gaps;
 	int gaps_inner;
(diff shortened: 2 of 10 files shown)


RyanDwyer:
Code LGTM. Not sure how I feel about having `seat_begin_move_tiling` set the seat operation to something other than `MOVE_TILING` though. I'd be neater if there was a `seat_begin_move_tiling_threshold` function.

However, after reviewing this changes I'm thinking perhaps we should use an interface for the seat operation (for a separate PR). I haven't given it much thought yet, but something with `.begin`, `.move`, `.end` and `.abort` callbacks would be a good start. We could then have one file per seat operation and remove a lot of stuff from `cursor.c`.

RedSoxFan:
> Not sure how I feel about having seat_begin_move_tiling set the seat operation to something other than MOVE_TILING though. I'd be neater if there was a seat_begin_move_tiling_threshold function.

I split it into a separate function. I just did it in one before since it was an almost identical function body. 

> However, after reviewing this changes I'm thinking perhaps we should use an interface for the seat operation (for a separate PR). I haven't given it much thought yet, but something with .begin, .move, .end and .abort callbacks would be a good start. We could then have one file per seat operation and remove a lot of stuff from cursor.c.

`sway/input/seat.h` and `sway/input/cursor.h`  could probably both be split into smaller files. However, since I haven't given it much thought, I'm not sure what the best approach is. At a glance, your idea sounds good and would move a lot of code out of both files.

emersion:
>However, after reviewing this changes I'm thinking perhaps we should use an interface for the seat operation (for a separate PR). 

Maybe pointer grabs could be used?

emersion:
Thanks!

Issue #3359, opened 2019-01-01
Crash when destroying fullscreen window when there are pending transactions


eiz:
Sway version: sway version 1.0-beta.2-72-g69e09824 (Dec 31 2018, branch 'master')
Logs: https://gist.github.com/eiz/360b22c0ee423c9f68d20bba2fa416c1
Display device: Vega64 with 3 DP monitors
Other environment stuff: running recent Mesa master due to unrelated fun facts. Pretty sure it's not relevant.

When a fullscreen window is destroyed, under specific circumstances sway can crash on a null pointer dereference. The view and container are in the 'destroying' state but waiting on pending transactions to complete and the output rendering code dereferences the view's surface after it has been cleared in view_unmap.

I first noticed this while running a fairly complex session on my 3-monitor setup with a Firefox window fullscreened on top of several other windows. My Firefox tabs crashed and shortly thereafter sway also crashed. Debugged into it a bit and was able to create a test case that reproduces it reliably for me.

To reproduce, build and run this program and follow the instructions in the comments:

```
#include <xcb/xcb.h>
#include <unistd.h>

int main()
{
	xcb_connection_t *c;
	xcb_screen_t *s;
	xcb_window_t w_resize_blocker, w_fullscreen;
	uint32_t mask = 0;
	uint32_t values[2];

	c = xcb_connect(NULL, NULL);
	s = xcb_setup_roots_iterator(xcb_get_setup(c)).data;

	w_fullscreen = xcb_generate_id(c);
	mask = XCB_CW_BACK_PIXEL;
	values[0] = s->black_pixel;
	xcb_create_window(
		c, 0, w_fullscreen, s->root, 0, 0, 100, 100,
		10, XCB_WINDOW_CLASS_INPUT_OUTPUT, s->root_visual,
		mask, values);
	xcb_map_window(c, w_fullscreen);
	xcb_flush(c);
	// During this sleep, move the created window to a different output,
	// full screen it, and move focus back to the original output (so
	// the new window gets created on a separate output). There must be
	// a native Wayland client running on the output on which the
	// fullscreen window will be displayed. Not sure why yet.
	// https://github.com/emersion/hello-wayland works.
	sleep(5);

	w_resize_blocker = xcb_generate_id(c);
	mask = XCB_CW_BACK_PIXEL | XCB_CW_EVENT_MASK;
	values[0] = s->white_pixel;
	values[1] = XCB_EVENT_MASK_RESIZE_REDIRECT;
[...]

PR #3361, opened 2019-01-01
Fix fullscreen view rendering crash

eiz:
See issue #3359 for reproduction details. When a fullscreen view is
unmapped and there's a preceding transaction waiting, there may be
neither a saved buffer or a surface to render. This change matches
the equivalent code in render_view.


RedSoxFan:
Thanks!

PR #3367, opened 2019-01-02
swaylock: Fix caps lock not updating immediately

### LINKED ISSUE #2788: swaylock: capslock indicator is unhelpful
"My friend" types his password with the Caps Lock activated. When he locks his screen and hits 'Caps Lock', the 'Caps Lock' text shows up.

Unfortunately, this text appears in the same way, with the same duration, regardless of if it is being enabled or disabled. Meaning that the user doesn't know if the next keystroke will be capitalized or not, until they've already hit the key and then see if it still says Caps Lock (meaning it was just enabled), or if the text has gone away (because it was just disabled).

I propose one of the following:
* be more explicit -- "Caps Lock ON" that persists, with "Caps Lock OFF" that displays when disabling, until next key press. Probably the smallest UX change, but I guess it leaks a tiny bit extra additional information to any potential shoulder-surfers
* allow users, such as my friend, to disable the Caps Lock text indicator, and instead do something like `--indicator-ring-caps-lock` to be able to change the ring color to indicate that Caps lock is or isn't enabled (this is more a feature request, I guess, though...)

oscarwcl:
Partially fixes #2788. This change makes it so the lock screen is redrawn whenever the caps lock modifier state changes, rather on relying on the keypress event. This didn't work because caps lock is disabled when the key is released, not pressed, so the caps lock indicator does not go away until the next keypress event.

This doesn't fully implement the suggestions in #2788, but is a more basic fix that just makes it so the caps lock indicator is always accurate. I'd be happy to have a go implementng them if anyone's interested, although it doesn't personally seem to me that they'd be that useful.

### DIFF
--- a/swaylock/password.c
+++ b/swaylock/password.c
@@ -146,14 +146,6 @@ void swaylock_handle_key(struct swaylock_state *state,
 		schedule_indicator_clear(state);
 		break;
 	case XKB_KEY_Caps_Lock:
-		/* The state is getting active after this
-		 * so we need to manually toggle it */
-		state->xkb.caps_lock = !state->xkb.caps_lock;
-		state->auth_state = AUTH_STATE_INPUT_NOP;
-		damage_state(state);
-		schedule_indicator_clear(state);
-		schedule_password_clear(state);
-		break;
 	case XKB_KEY_Shift_L:
 	case XKB_KEY_Shift_R:
 	case XKB_KEY_Control_L:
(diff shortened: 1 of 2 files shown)


RedSoxFan:
> This doesn't fully implement the suggestions in #2788, but is a more basic fix that just makes it so the caps lock indicator is always accurate. I'd be happy to have a go implementng them if anyone's interested, although it doesn't personally seem to me that they'd be that useful.

It looks like there are three things in that issue:

1. Persistent `Caps Lock ON` with `Caps Lock OFF` for the next key press
This PR mostly hits this bullet, it's just missing the `Caps Lock OFF` for a key press. Personally, I think this is good enough and if someone really wants the last bit, another PR (or even a local patch) can be submitted/applied.

2. `--indicator-ring-caps-lock` (along with something like ` --disable-caps-lock-text`)
I do like this idea, but it can be done in a separate PR.

3. Only showing caps lock on entry
If this does get implements, I think it should be optional, but I don't personally have any strong opinion either way on implementing this.

ddevault:
Thanks!

colemickens:
Cool, another thank you from me!

PR #3369, opened 2019-01-03
Add "tiling_drag_threshold_on_focused" command

David96:
Not sure whether this is wanted. But the first time I noticed the new tiling drag feature was that I accidentally dragged the focused window therefore I'd like to have the threshold also apply to focused windows simply to avoid accidentally dragging them.
I'm also happy to rename the command - I just didn't find a better name. This one seems a bit long though.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -173,6 +173,7 @@ sway_cmd cmd_swaynag_command;
 sway_cmd cmd_swap;
 sway_cmd cmd_tiling_drag;
 sway_cmd cmd_tiling_drag_threshold;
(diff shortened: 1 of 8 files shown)


David96:
I can create a pull request for that instead, but as the tiling threshold doc explicitly states that it doesn't apply to focused containers I thought this was the behaviour that most people agreed on.

RedSoxFan:
I agree that modifying `tiling_drag_threshold` (and it's documentation) would be the better solution. The only reason why it only applies to unfocused containers is because I replicated an open i3 PR.

David96:
Alright, closing this PR then.

PR #3378, opened 2019-01-06
Unbreak build on 32bit platforms

jbeich:
Found while packaging on FreeBSD e.g.,
- [12.0 armv7](https://ptpb.pw/a104)
- [12.0 i386](https://ptpb.pw/bxyk)


emersion:
Thanks!

PR #3379, opened 2019-01-06
Unhide mouse after container warp to focus

Emantor:
After @RedSoxFan implemented hide_cursor command, I noticed that the cursor was not shown between window switches, if container warping is enabled. This PR adds this feature, however I am not sure whether this should be configurable and that the code handles all cases correctly.

Please provide feedback if and how this should be configurable (IMO this only makes sense if container warping is enabled) and review the code, thanks!

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -654,6 +654,7 @@ void view_unmap(struct sway_view *view) {
 
 	struct sway_seat *seat;
 	wl_list_for_each(seat, &server.input->seats, link) {
+		seat->cursor->image_surface = NULL;
 		seat_consider_warp_to_focus(seat);
 	}
 
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -654,14 +654,7 @@ void view_unmap(struct sway_view *view) {
 
 	struct sway_seat *seat;
 	wl_list_for_each(seat, &server.input->seats, link) {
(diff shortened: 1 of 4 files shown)


sway/tree/view.c
```diff
@@ -654,14 +654,8 @@ void view_unmap(struct sway_view *view) {
...
-		if (config->mouse_warping == WARP_CONTAINER) {
-			struct sway_node *node = seat_get_focus(seat);
-			if (node && node->type == N_CONTAINER) {
-				cursor_warp_to_container(seat->cursor, node->sway_container);
-			} else if (node && node->type == N_WORKSPACE) {
-				cursor_warp_to_workspace(seat->cursor, node->sway_workspace);
-			}
-		}
+		seat->cursor->image_surface = NULL;
+		seat_consider_warp_to_focus(seat);
```
Emantor: @RyanDwyer Do you have an idea why this was not changed to `seat_consider_warp_to_focus` previously? Just want to make sure I'm fixing a leftover location and not introducing a bug.

Emantor:
Thanks, I'm fine with getting this merged as is. I left a comment for `view_unmap`, if that is answered I'll remove the `[WIP]` and we can get this merged.

RedSoxFan:
Thanks!

PR #3382, opened 2019-01-07
Proposed fix for swayidle forking behaviour

### LINKED ISSUE #3119: After resuming from suspend: desktop briefly shows up before lockscreen
* Sway version: 1.0-beta.1-174-g80a1c340 (Nov 11 2018, branch 'master')
* Debug log: https://ptpb.pw/lDoA
* Config file: https://ptpb.pw/nAMb
The config file is the default one with just the swayidle example commented in, plus different $mod and $term.

**Steps to reproduce**

Start sway, open terminal and issue 'systemctl suspend', press key for resume.
What happens: I see my desktop for 0.6 seconds, then I see the lockscreen.
What should happen: I see the lockscreen.

Unlocking works fine btw.
This is on a Samsung Chromebook Plus, running ArchlinuxARM: Linux alarm 4.19.1-2-ARCH #1 SMP Sun Nov 4 19:18:32 MST 2018 aarch64 GNU/Linux

ianyfan:
This should fix #3119, but I'm not sure whether it has the potential to leave zombie processes.

### DIFF
--- a/config.in
+++ b/config.in
@@ -34,11 +34,11 @@ output * bg @datadir@/backgrounds/sway/Sway_Wallpaper_Blue_1920x1080.png fill
 #
 # Example configuration:
 #
-#exec swayidle \
-#    timeout 300 'swaylock -c 000000' \
-#    timeout 600 'swaymsg "output * dpms off"' \
-#       resume 'swaymsg "output * dpms on"' \
-#    before-sleep 'swaylock -c 000000'
+# exec swayidle -w \
+#          timeout 300 'swaylock -f -c 000000' \
+#          timeout 600 'swaymsg "output * dpms off"' \
+#               resume 'swaymsg "output * dpms on"' \
+#          before-sleep 'swaylock -f -c 000000'
 #
 # This will lock your screen after 300 seconds of inactivity, then turn off
 # your displays after another 300 seconds, and turn your screens back on when


ianyfan:
I don't think forking should be the default, but if you want it to be, please suggest what the flag and description should be.

emersion:
I think `-f` is confusing. `swaylock -f` _daemonizes_ swaylock (ie. swaylock starts in the background). Here, the behaviour is different: swayidle either waits for child processes to finish, or runs them in the background.

ianyfan:
Makes sense. I've changed it to `-w` to wait for the command to finish executing. What do you think?

emersion:
This looks good. I think the example config still needs to be updated, to use `swayidle -w` and `swaylock -f`?

emersion:
We also need to make sure this correctly locks the screen before idling.

swayidle/main.c
```diff
@@ -97,29 +90,25 @@ static void acquire_sleep_lock(void) {
...
+		goto cleanup;
 	}
 
 	// sd_bus_message_unref closes the file descriptor so we need
 	// to copy it beforehand
 	lock_fd = fcntl(lock_fd, F_DUPFD_CLOEXEC, 3);
 	if (lock_fd < 0) {
-		wlr_log(WLR_ERROR, "Failed to copy sleep lock fd: %s",
-			strerror(errno));
+		wlr_log(WLR_ERROR, "Failed to copy sleep lock fd: %s", strerror(errno));
```
emersion: Missing `goto` here

swayidle/swayidle.1.scd
```diff
@@ -16,6 +16,10 @@ swayidle - Idle manager for Wayland
 *-d*
 	Enable debug output.
 
+*-w*
+	Wait for command to finish executing before continuing, helpful for ensuring
+	that a *before-sleep* command has finished before the system goes to sleep.
```
emersion: Maybe add a warning like "swayidle will block until the command finishes".

PR #3385, opened 2019-01-09
cursor: allow mapping to all outputs

### LINKED ISSUE #3371: Undo map_to_output
I'm testing sway 1.0-beta.2

I sometimes what to lock my pointer to my external monitor, I do that by running:
 `sway input "<input-id>" map_to_output "<output-id>"`.

But as far as I was able to find there is no command to map it back to all outputs. It would be nice to be able to do `sway input "<input-id>" map_to_output *` to reset it.

For now even reloading config doesn't reset it... I have to reconnect the device.

robertgzr:
Running `input "<input>" map_to_output *` resets the mapping to all outputs

Fixes #3371

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -428,6 +428,12 @@ static void seat_apply_input_config(struct sway_seat *seat,
 	if (mapped_to_output != NULL) {
 		wlr_log(WLR_DEBUG, "Mapping input device %s to output %s",
 			sway_device->input_device->identifier, mapped_to_output);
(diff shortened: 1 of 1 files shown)


ddevault:
Thank you!

robertgzr:
wow that was quick :)

This does not address https://github.com/swaywm/sway/issues/3371#issuecomment-451318827 though, which sounds like a good idea as well

emersion:
>This does not address #3371 (comment) though, which sounds like a good idea as well

See https://github.com/swaywm/sway/pull/3388

PR #3386, opened 2019-01-09
reload: apply seat cfgs after reading entire cfg

RedSoxFan:
Wait until all seat configs have been read before applying them on
reload. This prevents unnecessary attachment/detachment of input
devices and therefore creation/destruction of seat devices as
individual lines are read.


emersion:
Thanks!

PR #3387, opened 2019-01-09
keyboard: update repeat timer before execution

RedSoxFan:
Since the keyboard can be destroyed by executing a binding (reloading
with a different seat attachment config), update the repeat timer before
executing the binding.


emersion:
Thanks!

PR #3389, opened 2019-01-09
cmd_swap: use handler context seat

RedSoxFan:
Use the handler context seat instead of the default seat in cmd_swap


emersion:
Thanks!

Issue #2918, opened 2018-10-21
Views hijack focus from all seats on map


YaLTeR:
`sway version 1.0-beta.1-2-ga4d68358 (Oct 21 2018, branch 'master')`
[sway.log](https://gist.github.com/YaLTeR/f2187b8ca50c48ac7de7e38c940403b7#file-sway-log)
[config](https://gist.github.com/YaLTeR/f2187b8ca50c48ac7de7e38c940403b7#file-config)

I was testing multiple seats and found that a second seat keyboard behaves very strangely. Inputting mod+enter on it (my sway shortcut for spawning a terminal) spawns two terminals. Opening a wayland client and typing something types every character twice. Opening a second wayland client, switching focus to it and then typing results in the characters being typed into both windows (one character into each).

All of this was happening when typing on the second seat keyboard; the first seat keyboard didn't have any weird behaviors. Also typing into xwayland clients doesn't have any weird behaviors.

In the sway.log session I started sway, pressed mod+enter on the second seat keyboard which spawned two terminals (instead of one), then I exited sway.

emersion:
Maybe the keyboard is added to both seats?

YaLTeR:
Hmm, perhaps. I changed the seat lines to:
```
seat seat0 fallback true
seat seat1 attach "1133:49941:Logitech_Logitech_USB_Keyboard"
```
Then reloaded config, tried opening the terminal and closing it and got a [crash](http://sprunge.us/08zQjM), and then while trying to paste that to a gist got [another crash](http://sprunge.us/wjPiC6).

After reloading the session, I tried opening two wayland clients and while switching focus from seat0 resulted in me being able to correctly type in different clients with different keyboards, trying to bring focus back and then switch it from seat1 resulted in the same behavior (so seat0 typing in the newly focused client and seat1 in the old client, while it should've been the other way around).

While typing the rest of this issue from my primary seat0 keyboard with seat1 keyboard just attached and not doing anything, I got a [third crash](http://sprunge.us/ZwZULL). I later got the same crash with seat1 keyboard not even attached.

emersion:
Does https://github.com/swaywm/sway/pull/3390 fix this issue?

YaLTeR:
Yeah, these issues seem to have been fixed on the latest master. There's one more thing I'd classify as unintuitive behavior: opening a client redirects all seats' inputs into it rather than just the seat that opened the client.

emersion:
[...]

PR #3390, opened 2019-01-09
cmd_bind: pass the seat to execute_command

RedSoxFan:
I think this fixes at least part of #2918. I'm not sure if it completely fixes it

`seat_execute_command` was incorrectly setting
`config->handler_context.seat` before calling `execute_command`. Since
`execute_command` was being called with a `NULL` seat argument,
`execute_command` was setting `config->handler_context.seat` to the
default seat. This resulted in all bindings being executed on the
default seat and causing undesired behavior for devices on other seats.


emersion:
Thanks!

PR #3391, opened 2019-01-09
Simplify FreeBSD build

jbeich:
Same as swaywm/wlroots#1454. libevdev already depends on `<linux/input.h>` from `evdev-proto` package on both FreeBSD and DragonFly.


emersion:
Thanks!

PR #3392, opened 2019-01-09
Reset container dimensions when moving into workspace from direction

### LINKED ISSUE #3251: Container size wrong when moving across monitors
To reproduce:
1. Use 2 monitors
2. Create two vertical windows on Monitor-1
3. Move the container nearest to Monitor-2 to Monitor-2
4. Move it back to Monitor-1
5. You see the container the window sits in now is half the size its supposed to be.
6. Keep doing steps 3 and 4 and the container keeps halving in size with every cycle.

RyanDwyer:
Fixes #3251.


RedSoxFan:
Thanks!

PR #3397, opened 2019-01-10
swaybar: obey height if given

### LINKED ISSUE #3012: [swaybar][unicode] Bar expands vertically when displaying very wide characters and doesn't contract afterwards
# Background Info

swaybar has a feature where it expands to include the height of the font that is displayed in each block.
Interestingly, very wide unicode characters increase the height of the bar even though they are only *wider* and not *higher* than other characters.

# Observed Behaviour

the bar becomes higher than normal, with text centered vertically. After the wide characters are not displayed any more, the bar keeps its expanded size.

# Expected Behaviour

swaybar should either *not* expand the height unnecessarily, or at least contract the bar afterwards, once the wide characters are not displayed any more.

# How to Reproduce

- start a swaybar with a unicode-capable font
- display regular characters, like a few "A"s
- observe the size of the bar
- display a wide unicode character (such as 𒁏)
- observe the size of the bar

# Sway Version

sway version 1.0-beta.1-84-g5a98dfee (Oct 27 2018, branch 'master')

RedSoxFan:
Fixes #3012 

If there is a bar height given, use that as the absolute height rather than as a minimum height. This matches i3-gaps behavior.

Easy test:
1. Use the following minimal config:
```
set $mod Mod4
set $term termite

bindsym $mod+Return exec $$term
bindsym $mod+Shift+e exit

bar bar-0 position top
bar bar-0 height 0
bar bar-0 status_command while date +'%Y-%m-%d %l:%M:%S %p'; do sleep 1; done

bar bar-1 position bottom
bar bar-1 height 25
bar bar-1 status_command while date +'%Y-%m-%d %l:%M:%S %p'; do sleep 1; done
```
2. Execute the command `swaymsg workspace 𒁏` (Unicode character: 0x1204F)
3. Observe `bar-0` (top) increases in height while `bar-1` (bottom) stays the same height

### DIFF
--- a/sway/config/bar.c
+++ b/sway/config/bar.c
@@ -96,7 +96,7 @@ struct bar_config *default_bar_config(void) {
 	bar->pango_markup = false;
 	bar->swaybar_command = NULL;
 	bar->font = NULL;
-	bar->height = -1;
+	bar->height = 0;
 	bar->workspace_buttons = true;
 	bar->wrap_scroll = false;
 	bar->separator_symbol = NULL;
--- a/sway/sway-bar.5.scd
+++ b/sway/sway-bar.5.scd
@@ -69,7 +69,7 @@ Sway allows configuring swaybar in the sway configuration file.
 	use.
 
 *height* <height>
-	Sets the height of the bar. Default height will match the font size.
+	Sets the height of the bar. Default height (0) will match the font size.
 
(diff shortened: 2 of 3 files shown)


ddevault:
Thanks!

Ferdi265:
Thanks for fixing this :)

PR #3398, opened 2019-01-10
input events: toggle and ipc get_inputs

### LINKED ISSUE #2157: Make it easier to toggle inputs
From my limited knowledge of libinput, it seems like it should be possible to query the current events state of an input and toggle it. Currently `sway input` can only send the 3 state events of libinput but it would be nice to have a user-accessible `toggle` event for purposes such as keyboard shortcuts or scripting. I currently use a ThinkPad and therefore toggle the touchpad on a semi-regular basis, so this would be a great help for me.

RedSoxFan:
Closes #2157 

Implements toggling input events during runtime. This will not attempt
to toggle to a mode that is not supported by the device. When toggling
the wildcard input, the device specific input configs are altered. Each
device will cycle one supported mode.

This also adds `libinput_send_events` to the IPC output for `get_inputs`
to allow for custom scripting of mode changes/toggling.

### DIFF
--- a/sway/commands/input/events.c
+++ b/sway/commands/input/events.c
@@ -1,9 +1,68 @@
+#include <limits.h>
 #include <string.h>
 #include <strings.h>
+#include <wlr/backend/libinput.h>
 #include "sway/config.h"
 #include "sway/commands.h"
 #include "sway/input/input-manager.h"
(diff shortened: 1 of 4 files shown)


ddevault:
Thanks!

PR #3399, opened 2019-01-10
Fix segfaults on output destruction

### LINKED ISSUE #3065: Segfault on exit when an output is disabled
* Sway Version: sway version 1.0-beta.1 (Nov  2 2018, branch 'HEAD') compiled without second check in drop_permission due to #3024 

* Debug Log: [sway.log](https://gist.github.com/apilat/72f49550966eb14185060c4063c6b8a3)

* Configuration File: only change to default is `output VGA-1 disable`. the error is not reproducible with the default configuration

* Full backtrace from coredump: [btfull.log](https://gist.github.com/apilat/5ce3d2115ddcaa0b8c2e86bcae28d0b6)

I ran sway, waited for all the screens to load, then exited with `Mod+Shift+E`.

RedSoxFan:
Fixes #3065 

This fixes two causes of segfaulting when an output is destroyed.

The first occurred when an output was never enabled. The issue was that
the destroy signal was never initialized so when it was emitted, sway
segfaulted. This was fixed by moving the initialization into
`output_create` since all outputs, regardless of whether they have ever
been enabled, will be destroyed at some point.

The second occurred when the cursor was on an output that was being
destroyed. The sway output would have already been removed, but if there
are other outputs, a cursor rebase would still occur. Since the
wlr_output still existed and the sway output was destroyed, the cursor
could be over nothing, resulting in a segfault when trying to get the
sway output, which was destroyed.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -85,6 +85,10 @@ static struct sway_node *node_at_coords(
 		return NULL;
 	}
 	struct sway_output *output = wlr_output->data;
+	if (!output) {
+		// output is being destroyed
+		return NULL;
+	}
 	double ox = lx, oy = ly;
 	wlr_output_layout_output_coords(root->output_layout, wlr_output, &ox, &oy);
 
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -58,6 +58,7 @@ struct sway_output *output_create(struct wlr_output *wlr_output) {
 	wlr_output->data = output;
 
 	wl_signal_add(&wlr_output->events.destroy, &output->destroy);
+	wl_signal_init(&output->events.destroy);
 
 	wl_list_insert(&root->all_outputs, &output->link);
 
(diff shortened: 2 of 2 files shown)


emersion:
TBH I'm not a fan of dynamically adding/removing listeners when enabling/disabling outputs.

sway/input/cursor.c
```diff
@@ -87,6 +87,10 @@ static struct sway_node *node_at_coords(
 	struct sway_output *output = wlr_output->data;
 	double ox = lx, oy = ly;
 	wlr_output_layout_output_coords(root->output_layout, wlr_output, &ox, &oy);
+	if (!output) {
```
emersion: Can we move this check on line 87?

RedSoxFan: Yup, moved

RedSoxFan:
> TBH I'm not a fan of dynamically adding/removing listeners when enabling/disabling outputs.

Well this is one step towards not dynamically adding/removing them since it moves the destroy listener out of `output_enable` and into `output_create`

emersion:
Thanks!

PR #3402, opened 2019-01-10
Refactor seat operations to use an interface

RyanDwyer:
This splits each seat operation (drag/move tiling/floating etc) into a separate file and introduces a `struct sway_seatop_impl` to abstract the operation. This removes a heap of code from cursor.c and seat.c, and removes a bunch of multi-purpose properties from the seat struct.

The `move_tiling_threshold` operation has been merged into `move_tiling`.

The main logic for each operation is untouched aside from variable renames.

The following previously-static functions have been made public:
* `node_at_coords`
* `container_raise_floating`
* `render_rect`
* `premultiply_alpha`
* `scale_box`

Everything seems to work correctly, but I've only tested it hastily. Please test thoroughly.

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -50,6 +50,12 @@ struct sway_cursor {
 	size_t pressed_button_count;
 };
 
+struct sway_node;
+
+struct sway_node *node_at_coords(
+		struct sway_seat *seat, double lx, double ly,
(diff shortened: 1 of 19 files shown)


RyanDwyer:
Bump.

ddevault:
Thanks!

PR #3403, opened 2019-01-10
Remove 'input' field of IPC command return json

mstoeckl:
Which "input" field?

    $ swaymsg -r bindsym ae

    [
      {
        "success": false,
        "parse_error": true,
        "error": "Invalid bindsym command (expected at least 2 arguments, got 1)",
        "input": "bindsym"
      }
    ]

Pro:
* Not in i3 protocol; nor used by `swaymsg` pretty-printed output; nor printed by config error messages; nor yet documented anywhere
* Code only provides it if `input` field of a given `struct cmd_results` is not NULL
* Information is redundant and imprecise: in config file, location of error already given; with IPC, return list already indicates which of the commands is responsible
* A very small step toward not relying on heap allocation success for command handler return values
* Net removal of code

Con:
* Backwards compatibility? (it is not in 0.15.2, at least)
* @ddevault wrote the code a year ago in d7d21bb0f895248cafefc3d12e4 and probably can explain why this is all a dumb idea

~If people are fine with this, I will add another commit removing the second "input" argument from `cmd_results_new`.~ Done.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -75,7 +75,7 @@ struct cmd_results *config_commands_command(char *exec);
 /**
  * Allocates a cmd_results object.
  */
-struct cmd_results *cmd_results_new(enum cmd_status status, const char* input, const char *error, ...);
+struct cmd_results *cmd_results_new(enum cmd_status status, const char *error, ...);
 /**
  * Frees a cmd_results object.
  */
@@ -87,8 +87,7 @@ void free_cmd_results(struct cmd_results *results);
  */
 char *cmd_results_to_json(list_t *res_list);
 
-struct cmd_results *add_color(const char *name,
-		char *buffer, const char *color);
+struct cmd_results *add_color(char *buffer, const char *color);
 
 /**
  * TODO: Move this function and its dependent functions to container.c.
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -31,7 +31,6 @@ enum cmd_status {
  */
 struct cmd_results {
 	enum cmd_status status;
-	char *input;
 	/**
 	 * Human friendly error message, or NULL on success
 	 */
@@ -63,7 +62,7 @@ list_t *execute_command(char *command,  struct sway_seat *seat,
  *
  * Do not use this under normal conditions.
  */
-struct cmd_results *config_command(char *command);
+struct cmd_results *config_command(char *command, char **new_block);
 /**
  * Parse and handle a sub command
  */
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -36,7 +36,7 @@ struct cmd_results *checkarg(int argc, const char *name, enum expected_args type
 		}
 	}
 	return error_name ?
-		cmd_results_new(CMD_INVALID, name, "Invalid %s command "
+		cmd_results_new(CMD_INVALID, "Invalid %s command "
 				"(expected %s%d argument%s, got %d)",
 				name, error_name, val, val != 1 ? "s" : "", argc)
 		: NULL;
@@ -228,8 +228,7 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 			char *error = NULL;
 			struct criteria *criteria = criteria_parse(head, &error);
 			if (!criteria) {
-				list_add(res_list, cmd_results_new(CMD_INVALID, head,
-					"%s", error));
+				list_add(res_list, cmd_results_new(CMD_INVALID,	"%s", error));
 				free(error);
 				goto cleanup;
 			}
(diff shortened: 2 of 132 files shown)


sway/commands.c
```diff
@@ -508,11 +507,7 @@ struct cmd_results *cmd_results_new(enum cmd_status status,
 		return NULL;
 	}
 	results->status = status;
-	if (input) {
-		results->input = strdup(input); // input is the command name
-	} else {
-		results->input = NULL;
-	}
+	// NOTE: `input` argument is unused, remove
```
emersion: Indeed, please do

include/sway/commands.h
```diff
@@ -75,7 +75,8 @@ struct cmd_results *config_commands_command(char *exec);
 /**
  * Allocates a cmd_results object.
  */
-struct cmd_results *cmd_results_new(enum cmd_status status, const char* input, const char *error, ...);
+struct cmd_results *cmd_results_new(enum cmd_status status,
+		const char *error, ...);
```
emersion: Style: I think we don't hit the 80 char limit if we put this on one line

mstoeckl:
The second and significantly larger commit removes the unused `input` argument from `cmd_results_new`.

I also replaced `static const char *foo = "blah"` with `static const char foo[] = "blah"` on encounter, but beyond that (and minor format string editing), I avoided making changes to the code.

Occasionally CMD_INVALID ("Unknown command or parser error") is used where CMD_FAILURE ("The command resulted in an error") should be. Continuation lines are often two tabs instead of one.

include/sway/commands.h
mstoeckl: 85.

mstoeckl: Fixing anyway -- the 80 char limit is soft, and the line originally was 103.

progandy:
I'm all for removing it, but I want to note that this differs from existing undocumented i3 behaviour. (The current implementation doesn't conform either). 

```
$  i3-msg move aae
ERROR: Your command: move aae
ERROR:                    ^^^
ERROR: Expected one of these tokens: 'window', 'container', 'to', '--no-auto-back-and-forth', 'workspace', 'output', 'mark', 'scratchpad', 'left', 'right', 'up', 'down', 'position', 'absolute'
[{"success":false,"parse_error":true,"error":"Expected one of these tokens: 'window', 'container', 'to', '--no-auto-back-and-forth', 'workspace', 'output', 'mark', 'scratchpad', 'left', 'right', 'up', 'down', 'position', 'absolute'","input":"move aae","errorposition":"     ^^^"}]

```

emersion:
IMHO command handlers shouldn't be responsible for setting the `input` field anyway, because i3 puts there the full raw command.

mstoeckl:
I am currently rebasing, although it may take up to an hour since the change to `cmd_results_new` does not always introduce compile errors.

sway/commands.c
```diff
@@ -265,8 +265,8 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 			}
 			struct cmd_handler *handler = find_handler(argv[0], NULL, 0);
 			if (!handler) {
-				list_add(res_list, cmd_results_new(CMD_INVALID, cmd,
-							"Unknown/invalid command"));
+				list_add(res_list, cmd_results_new(CMD_INVALID,
+						"Unknown/invalid command"));
```
ianyfan: Can you format `cmd` into the string like in the other ones?

sway/commands/input/map_from_region.c
```diff
@@ -52,30 +52,30 @@ struct cmd_results *input_cmd_map_from_region(int argc, char **argv) {
...
 		ic->mapped_from_region = NULL;
-		return cmd_results_new(CMD_FAILURE, "map_from_region",
+		return cmd_results_new(CMD_FAILURE,
 			"Invalid rectangle");
 	}
 	if (mm1 != mm2) {
 		free(ic->mapped_from_region);
 		ic->mapped_from_region = NULL;
-		return cmd_results_new(CMD_FAILURE, "map_from_region",
+		return cmd_results_new(CMD_FAILURE,
```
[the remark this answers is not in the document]
mstoeckl: Assuming four spaces per tab, this should not be a single line; I instead combined lines 55-56 -> 54.

mstoeckl:
Oops, confused 'ours' with 'theirs' during rebase conflict resolution. Sorry about that.

Most of the formatting requests are consistent with 4-spaces per tab and a maximum line length of 80. As this deviates from kernel-style (8-spaces per tab), it would be helpful to have this mentioned in CONTRIBUTING.md.

RedSoxFan:
> Most of the formatting requests are consistent with 4-spaces per tab and a maximum line length of 80. As this deviates from kernel-style (8-spaces per tab), it would be helpful to have this mentioned in CONTRIBUTING.md.

Sounds like a good idea to me. I actually thought it was in there.

RedSoxFan:
Thanks!

PR #3404, opened 2019-01-10
swaylock: Add caps lock state to indicator

Robinhuett:
This implements customization for the indicator as proposed in #2788 with comments from #3367 in mind.
The default behaviour does not change exept for the caps lock text color.

@RedSoxFan Could you review this? Thank you :)

### DIFF
--- a/swaylock/render.c
+++ b/swaylock/render.c
@@ -18,7 +18,17 @@ static void set_color_for_state(cairo_t *cairo, struct swaylock_state *state,
 	} else if (state->auth_state == AUTH_STATE_CLEAR) {
 		cairo_set_source_u32(cairo, colorset->cleared);
 	} else {
-		cairo_set_source_u32(cairo, colorset->input);
+		if (state->xkb.caps_lock && state->args.show_caps_lock_indicator) {
+			cairo_set_source_u32(cairo, colorset->caps_lock);
+		} else if (state->xkb.caps_lock && !state->args.show_caps_lock_indicator &&
+				state->args.show_caps_lock_text) {
+			uint32_t inputtextcolor = state->args.colors.text.input;
+			state->args.colors.text.input = state->args.colors.text.caps_lock;
+			cairo_set_source_u32(cairo, colorset->input);
+			state->args.colors.text.input = inputtextcolor;
+		} else {
+			cairo_set_source_u32(cairo, colorset->input);
+		}
 	}
 }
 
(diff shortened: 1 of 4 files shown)


swaylock/render.c
```diff
@@ -18,7 +18,14 @@ static void set_color_for_state(cairo_t *cairo, struct swaylock_state *state,
 	} else if (state->auth_state == AUTH_STATE_CLEAR) {
 		cairo_set_source_u32(cairo, colorset->cleared);
 	} else {
-		cairo_set_source_u32(cairo, colorset->input);
+		if (state->xkb.caps_lock && state->args.show_caps_lock_indicator) {
+			cairo_set_source_u32(cairo, colorset->caps_lock);
+		} else {
+			if (state->xkb.caps_lock && !state->args.show_caps_lock_indicator) {
+				state->args.colors.text.input = state->args.colors.text.caps_lock;
```
RedSoxFan: Also, can you please reset `state->args.color.text.input` after setting the cairo source?

swaylock/render.c
```diff
@@ -18,7 +18,15 @@ static void set_color_for_state(cairo_t *cairo, struct swaylock_state *state,
 	} else if (state->auth_state == AUTH_STATE_CLEAR) {
 		cairo_set_source_u32(cairo, colorset->cleared);
 	} else {
-		cairo_set_source_u32(cairo, colorset->input);
+		if (state->xkb.caps_lock && state->args.show_caps_lock_indicator) {
+			cairo_set_source_u32(cairo, colorset->caps_lock);
+		} else if (state->xkb.caps_lock && !state->args.show_caps_lock_indicator) {
+			state->args.colors.text.input = state->args.colors.text.caps_lock;
+			cairo_set_source_u32(cairo, colorset->input);
+			state->args.colors.text.input = state->args.colors.text.input;
```
RedSoxFan: You have to store the value before changing it. Currently, this line is a nop.

swaylock/render.c
```diff
@@ -18,7 +18,15 @@ static void set_color_for_state(cairo_t *cairo, struct swaylock_state *state,
 	} else if (state->auth_state == AUTH_STATE_CLEAR) {
 		cairo_set_source_u32(cairo, colorset->cleared);
 	} else {
-		cairo_set_source_u32(cairo, colorset->input);
+		if (state->xkb.caps_lock && state->args.show_caps_lock_indicator) {
+			cairo_set_source_u32(cairo, colorset->caps_lock);
+		} else if (state->xkb.caps_lock && !state->args.show_caps_lock_indicator) {
```
RedSoxFan: Can you also add a check for show_caps_lock_text here since both can be disabled and some other form of text may be shown?

Tbh I'm not a huge fan of setting the font color here, but moving it would require changes that don't belong in this PR

swaylock/render.c
Robinhuett: Ahh, I see. This was total nonsense. Sorry for that

swaylock/render.c
Robinhuett: I'm not a fan of this either, but unfortunatly there was no "easy" way of setting just one color.
Nonetheless I've added the check

RedSoxFan:
Thanks!

PR #3411, opened 2019-01-12
config: do not reset pos when braces found

### LINKED ISSUE #3410: #3400 makes config crash
https://github.com/swaywm/sway/pull/3400 creates weird effects on my config.

i reduced it to this line that creates the fail:

`bindsym $mod+XF86AudioRaiseVolume exec amixer -D pulse sset Master 1%+ && pkill -RTMIN+1 i3blocks` 

i don't even know why i got this in my config, seems like a stupid way

RedSoxFan:
Fixes #3410 

When a brace is found, the config file should not seek back to before
the brace, otherwise the brace will be read multiple times.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -613,7 +613,9 @@ static int detect_brace(FILE *file) {
 		}
 	}
 	free(line);
-	fseek(file, pos, SEEK_SET);
+	if (ret == 0) {
+		fseek(file, pos, SEEK_SET);
+	}
 	return ret;
 }
 


ianyfan:
I think this would be better placed in the `else` block of the `if (strcmp(line, "{") == 0)`, so that it skips empty lines.

RedSoxFan:
The issue with that is the lines numbers for errors won't match the real file line number (note that this is already the case if the config file contains any continued lines since `get_line_with_cont` is assumed to be a single line.) Ideally, the numbers should be accurate to make finding and fixing the lines easier

PR #3412, opened 2019-01-12
config: fix line number with continued lines

RedSoxFan:
When the config has continued lines, `get_line_with_cont` may read more
than one line of the actual file. When displaying line numbers for error
messages, they should be the line number in the file to make it easy to
find and fix the issue.


ianyfan:
Thanks!

Issue #3253, opened 2018-12-04
resize grow/shrink <direction> is not consistent with i3


d-k-c:
* Sway Version: sway version 1.0-beta.2-17-gbca21ee0 (Dec 3 2018, branch 'maste
* Configuration File: vanilla configuration, only change is `set $term gnome-terminal`

In a workspace, open 3 terminals with horizontal splits:

```
+---------+
|    A    |
+---------+
|    B    |
+---------+
|    C    |
+---------+
```

In terminal A, typing, `swaymsg resize grow down` moves boundaries of all the windows. In i3, only the boundary between A and B moves, boundary between B and C stays at the same place.

Sidenote: while playing with resize, I got some weird "Error: Cannot resize any further" in terminal B when resize should have happened, but I could not determine what caused the issue. Just posting that here in case someone else noticed it.

PR #3418, opened 2019-01-14
Replace resize_axis with wlr_edges

RyanDwyer:
This patch removes the `resize_axis` enum in favour of `wlr_edges`. As `wlr_edges` has no `horizontal` or `vertical` value, it denotes these by bitwise `or`ing the left/right and up/down values. Two constants are defined to make it easier to refer to these.

This will allow the tiling resize seatop to utilise the functions in this file. `resize_axis` was local to the resize command and couldn't be exposed in function arguments.

This is progress towards #3253.

The function `container_resize_tiled` used to convert from a `wlr_edges` to a `resize_axis`, and now just passes the values through. This function is public and will be cleaned up as part of #3253.

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -104,7 +104,7 @@ static void calculate_constraints(int *min_width, int *max_width,
 	}
 }
 
-static enum wlr_edges parse_resize_axis(const char *axis) {
+static uint32_t parse_resize_axis(const char *axis) {
 	if (strcasecmp(axis, "width") == 0 || strcasecmp(axis, "horizontal") == 0) {
 		return AXIS_HORIZONTAL;
 	}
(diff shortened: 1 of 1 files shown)


sway/commands/resize.c
```diff
@@ -111,61 +104,48 @@ static void calculate_constraints(int *min_width, int *max_width,
 	}
 }
 
-static enum resize_axis parse_resize_axis(const char *axis) {
+static enum wlr_edges parse_resize_axis(const char *axis) {
 	if (strcasecmp(axis, "width") == 0 || strcasecmp(axis, "horizontal") == 0) {
-		return RESIZE_AXIS_HORIZONTAL;
+		return AXIS_HORIZONTAL;
```
emersion: This value is outside the enum. The return type should be changed to `uint32_t`.

RyanDwyer: Does that mean I should change everything from `enum wlr_edges` to `uint32_t` where it could contain `AXIS_HORIZONTAL` or `AXIS_VERTICAL`?

emersion: Using `enum thing` with values outside of the enum is dangerous because `switch` statements might forget some cases. Bitfields are a special case of this. So yeah, everything should be changed if necessary.

RyanDwyer: Done.

emersion:
Thanks!

PR #3419, opened 2019-01-14
swaynag: handle empty $XDG_CONFIG_HOME better

ianyfan:
Set config path to fallback instead of setting $XDG_CONFIG_HOME


emersion:
Thanks!

PR #3422, opened 2019-01-14
Resize only current and immediate siblings rather than all siblings

### LINKED ISSUE #3253: resize grow/shrink <direction> is not consistent with i3

RyanDwyer:
For example, create layout `V[view view view]` and resize the leftmost view using mod+rightclick. Previously, the edge between view 2 and 3 would be adjusted as well. Now this edge will remain constant, which matches the behaviour of i3.

To do this operation correctly, the resize tiling seatop now keeps track of two containers, as the container that resizes horizontally will be a different container to the one which resizes vertically (one will be an ancestor). The tiling resize seatop now figures out these containers during the start of the operation and keeps references to them in the event. A new function `container_find_resize_parent` has been introduced to do this. This function is also used by the resize command.

During cursor motion, the seatop logic is similar to before, but now has to choose the correct container to resize.

In `resize.c`, `container_resize_tiled` and `resize_tiled` have been merged into one. One of them originally did nothing except pass the values through to the other.

`container_resize_tiled` now takes a simplified approach where it just finds the immediate siblings on either side and resizes them without worrying about the others. The `parellel_coord` and `parallel_size` functions are no longer needed and have been removed.

Fixes #3253.

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -130,175 +130,106 @@ static bool is_horizontal(uint32_t axis) {
 	return axis & (WLR_EDGE_LEFT | WLR_EDGE_RIGHT);
 }
 
-static int parallel_coord(struct sway_container *c, uint32_t axis) {
-	return is_horizontal(axis) ? c->x : c->y;
-}
-
-static int parallel_size(struct sway_container *c, uint32_t axis) {
-	return is_horizontal(axis) ? c->width : c->height;
-}
-
-static void container_recursive_resize(struct sway_container *container,
-		double amount, enum wlr_edges edge) {
-	bool layout_match = true;
-	wlr_log(WLR_DEBUG, "Resizing %p with amount: %f", container, amount);
-	if (edge & (WLR_EDGE_LEFT | WLR_EDGE_RIGHT)) {
-		container->width += amount;
-		layout_match = container->layout == L_HORIZ;
-	} else if (edge & (WLR_EDGE_TOP | WLR_EDGE_BOTTOM)) {
-		container->height += amount;
-		layout_match = container->layout == L_VERT;
-	}
-	if (container->children) {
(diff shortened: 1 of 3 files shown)


sway/commands/resize.c
```diff
@@ -130,177 +130,108 @@ static bool is_horizontal(uint32_t axis) {
...
+	// both adjacent siblings. For RIGHT or DOWN, just select the next sibling.
+	// For LEFT or UP, convert it to a RIGHT or DOWN resize and reassign con to
+	// the previous sibling.
+	struct sway_container *prev = NULL;
+	struct sway_container *next = NULL;
+	list_t *siblings = container_get_siblings(con);
+	int index = container_sibling_index(con);
+
+	if (axis == AXIS_HORIZONTAL || axis == AXIS_VERTICAL) {
+		prev = siblings->items[index - 1];
```
RedSoxFan: When running `resize grow width 10px` or `resize grow height 10px` on the first or last child, sway segfaults since `prev` and `next` get assigned invalid values here that are later dereferenced.

RyanDwyer: Fixed.

RedSoxFan:
Thanks!

PR #3423, opened 2019-01-14
Implement fullscreen global

### LINKED ISSUE #3255: fullscreen command is missing the 'global' argument
According to i3's documentation ([here](https://i3wm.org/docs/userguide.html#_toggling_fullscreen_mode_for_a_window) and [here](https://i3wm.org/docs/userguide.html#_manipulating_layout)):
>  There is also a global fullscreen mode in i3 in which the client will span all available outputs (the command is `fullscreen toggle global`)

>To make the current window (!) fullscreen, use `fullscreen enable` (or `fullscreen enable global` for the global mode), to leave either fullscreen mode use `fullscreen disable`, and to toggle between these two states use `fullscreen toggle` (or `fullscreen toggle global`).

So the complete fullscreen command should be `fullscreen [enable|disable|toggle] [global]`. I have tested this in i3 with a second monitor and it works as described.

Personally, I don't need this feature (I didn't know it existed until recently), so if you don't want to support it, you can close this issue.

RyanDwyer:
The fullscreen container is stored in the root struct.

I'm not sure how it's supposed it behave in regards to using `fullscreen disable` or `fullscreen toggle` (without global) while the container is fullscreen global. Or whether a container can be both workspace-fullscreen and global-fullscreen at the same time.

I've added checks in the `focus` and `workspace` commands to prevent moving focus off the fullscreen global container. Not sure if similar checks are needed elsewhere or whether they should go in `seat_set_focus` instead.

Please test thoroughly.

Closes #3255.

### DIFF
--- a/sway/commands/swap.c
+++ b/sway/commands/swap.c
@@ -77,6 +77,11 @@ static void swap_focus(struct sway_container *con1,
 	} else {
 		seat_set_focus_container(seat, focus);
 	}
+
+	if (root->fullscreen_global) {
+		seat_set_focus(seat,
+				seat_get_focus_inactive(seat, &root->fullscreen_global->node));
+	}
 }
 
 static void container_swap(struct sway_container *con1,
@@ -98,13 +103,13 @@ static void container_swap(struct sway_container *con1,
 	sway_log(SWAY_DEBUG, "Swapping containers %zu and %zu",
 			con1->node.id, con2->node.id);
 
-	bool fs1 = con1->is_fullscreen;
-	bool fs2 = con2->is_fullscreen;
+	enum sway_fullscreen_mode fs1 = con1->fullscreen_mode;
+	enum sway_fullscreen_mode fs2 = con2->fullscreen_mode;
 	if (fs1) {
-		container_set_fullscreen(con1, false);
+		container_fullscreen_disable(con1);
 	}
 	if (fs2) {
-		container_set_fullscreen(con2, false);
+		container_fullscreen_disable(con2);
 	}
 
 	struct sway_seat *seat = config->handler_context.seat;
@@ -136,10 +141,10 @@ static void container_swap(struct sway_container *con1,
 	}
 
 	if (fs1) {
-		container_set_fullscreen(con2, true);
+		container_set_fullscreen(con2, fs1);
 	}
 	if (fs2) {
-		container_set_fullscreen(con1, true);
+		container_set_fullscreen(con1, fs2);
 	}
 }
 
@@ -220,9 +225,13 @@ struct cmd_results *cmd_swap(int argc, char **argv) {
 
 	container_swap(current, other);
 
-	arrange_node(node_get_parent(&current->node));
-	if (node_get_parent(&other->node) != node_get_parent(&current->node)) {
-		arrange_node(node_get_parent(&other->node));
+	if (root->fullscreen_global) {
+		arrange_root();
+	} else {
+		arrange_node(node_get_parent(&current->node));
+		if (node_get_parent(&other->node) != node_get_parent(&current->node)) {
+			arrange_node(node_get_parent(&other->node));
+		}
 	}
 
 	return cmd_results_new(CMD_SUCCESS, NULL);
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -25,6 +25,12 @@ enum sway_container_border {
 	B_CSD,
 };
 
+enum sway_fullscreen_mode {
+	FULLSCREEN_NONE,
+	FULLSCREEN_WORKSPACE,
+	FULLSCREEN_GLOBAL,
+};
+
 struct sway_root;
 struct sway_output;
 struct sway_workspace;
@@ -38,7 +44,7 @@ struct sway_container_state {
 	double x, y;
 	double width, height;
 
-	bool is_fullscreen;
+	enum sway_fullscreen_mode fullscreen_mode;
 
 	struct sway_workspace *workspace;
 	struct sway_container *parent;
@@ -85,7 +91,7 @@ struct sway_container {
 	double content_x, content_y;
 	int content_width, content_height;
 
-	bool is_fullscreen;
+	enum sway_fullscreen_mode fullscreen_mode;
 
 	enum sway_container_border border;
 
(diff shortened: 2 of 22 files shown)


RedSoxFan:
I just did some testing in i3 and found the following:
- `fullscreen global` is synonymous with `fullscreen toggle global` (similar to `fullscreen` to `fullscreen toggle`)
- `fullscreen enable` on a container that is `fullscreen global` works the same as `fullscreen disable global, fullscreen enable` (and similar for `fullscreen enable global`)
- `fullscreen disable [global]` will disable fullscreen regardless which mode it is in
    - `fullscreen disable global` disables the fullscreen, but also returns `CMD_INVALID` with `^^^` under `global`
- `fullscreen toggle [global]` on a container that is fullscreen in either mode will act as `fullscreen disable`

RedSoxFan:
Can you please also remove the `input` argument for the calls to `cmd_results_new`  since #3403 has been merged?

Also, i3 sets `fullscreen_mode` to `2` in IPC_GET_TREE for a container that is `fullscreen global`

RyanDwyer:
* Toggling logic has been adjusted.
* I've introduced a `sway_fullscreen_mode` enum and merged the two `container_set_fullscreen` functions into one. This means we can pass the enum value straight through IPC.
* As part of the above, I found several other places where we needed to check for fullscreen.
* There is a new function `container_fullscreen_disable` which is just there for convenience.

sway/commands/swap.c
```diff
@@ -99,13 +99,13 @@ static void container_swap(struct sway_container *con1,
 	wlr_log(WLR_DEBUG, "Swapping containers %zu and %zu",
 			con1->node.id, con2->node.id);
 
-	bool fs1 = con1->is_fullscreen;
-	bool fs2 = con2->is_fullscreen;
+	bool fs1 = con1->fullscreen_mode;
+	bool fs2 = con2->fullscreen_mode;
```
RedSoxFan: I think these should be `enum sway_fullscreen_mode` now

RedSoxFan:
If you fullscreen (in either mode) a container with multiple views, this prevents changing focus between them.

RedSoxFan:
A few more issues:
1. If you splith an ancestor of a fullscreen container, the container gets arranged as if it is not fullscreen.
2. You can move an ancestor out of the fullscreen container onto the workspace it is covering. If the ancestor had focus, focus will be stay on the now hidden container/view
3. When swapping a focused ancestor of a fullscreen container with one hidden, focus will stay on the now hidden container/view

RedSoxFan:
> I'm too tired at the moment to try and understand what the swap command is doing with focus (I don't use the swap feature myself, which makes it even harder). It looks like you wrote the original code. Can you offer some insight?

_(I don't use it either. It was one of my first contributions and I just picked it since it seemed like something that would let me get a feel for different things.)_

It looks like this is actually a bug on master as well _(when I implemented swap, fullscreen containers weren't supported yet)_. 

Here's what I think is going on
- If the focused container is not part of the swap, there is no focus change.
- If the focus container remains on the same workspace it retains focus, otherwise the swapped container gets focused.
   - Previously, `container_set_fullscreen` handled focusing the fullscreen view since fullscreen containers were not yet supported

I think the only issue is when swapping a descendant of a fullscreen container with a container that is not a descendant of a fullscreen container.

RedSoxFan:
Also, I found one more issue:
- Create the layout `H[view view]` and `fullscreen global` it
- From the left view, run `move right`
- The views get arranged as if it is not fullscreen
- If you attempt to run `fullscreen toggle` on either of the views, it toggles fullscreen of that view never exiting the ancestor's fullscreeen
- If you focus the parent and then `fullscreen toggle`, there is some artifacts on the other outputs

RyanDwyer:
I think I've fixed everything listed. For all the arrange issues, any time we arrange a workspace we have to check if there's a fullscreen global and arrange root instead. I don't particularly want to do that in `arrange_workspace` itself, so I'm doing it in each command.

My fix for the swap focus is a bit dirty, but I still don't understand the original logic. Specifically, I don't understand why it needs to check the parent layout type. We should consider refactoring that code and adding comments at a minimum.

RedSoxFan:
A few more issues:
1. This is a slight variation of the move one in my previous post, but without the move step
    - Create the layout H[view view] and fullscreen global it
    - Run fullscreen toggle on either of the views, it toggles fullscreen of that view never exiting the parent's fullscreeen
    - If you focus the parent and then fullscreen toggle, there is some artifacts on the other outputs
2. When in fullscreen global, the damage of xdg-shell views exclusively on an output other than the workspace's doesn't seem to be correct (for xwayland views it appears to be working)
    - With two outputs open two xdg-shell views (termite, gnome-calculator, etc)
    - `focus parent`
    - `fullscreen global`
    - focus the view on the output that does not contain the workspace
    - Try to type anywhere that accepts text
    - Move the view to the output where the workspace is and notice the text is there
3. `root_scratchpad_show` needs to handle `fullscreen global`

As for swap, it probably could use some refactoring. I think a lot of stuff has changed since I original wrote it, but that's for a different PR. Your changes appear fine to me for now.

RyanDwyer:
I've made a bunch of updates. This is from my commit message:

* `arrange_root` when doing fullscreen changes, because we don't know if normal fullscreening a normal container is unfullscreening a fullscreen global container
* Consider fullscreen global containers in `node_at_coords`
    * Note that this doesn't consider transient containers. Transient containers need to be refactored anyway, as they aren't considered when the fullscreen view is a split container anyway.
* Fix a bool that should be an enum in `container_floating_move_to_center`
* Refactor `container_set_fullscreen` so that the function only orchestrates the toggling of other containers
    * Introduces `container_fullscreen_workspace` which changes a container from FS none to FS workspace
    * Introduces `container_fullscreen_global` which changes a container from FS none to FS global
    * Repurposes `container_fullscreen_disable` which changes a container from either FS mode to FS none
* Fix some scratchpad weirdness:
    * When scratchpad hiding a fullscreen global container, nullify the root's `fullscreen_global` pointer (in `container_detach`)
    * When scratchpad showing a container with mode set to `FULLSCREEN_GLOBAL`, disable any existing fullscreen global container

The unresponsive view issue has me stumped at the moment. It happens for me with xwayland too. We send the pointer and keyboard events properly, but the client doesn't commit anything. And from what I can tell we're sending frame events properly (the view passes the `view_is_visible` check).

I'd like to squash my commits once I have this in a mergeable state. I've refactored fullscreen a few times now in this PR and there are buggy commits, so having one squashed commit would make it easy to revert if needed.

RedSoxFan:
Aside from the remaining rendering/damaging/whatever-it-ends-up-being issue, this LGTM.

RyanDwyer:
I have resolved the remaining rendering/damaging issue. `output_for_each_surface` needed to consider the fullscreen global container and its children.

If you guys are happy to merge this now, let me know and I'll squash my commits first.

ddevault:
Thanks!

PR #3430, opened 2019-01-15
Handle hidden scratchpad containers in commands

### LINKED ISSUE #3236: Some commands crash sway when operating on hidden scratchpad containers
Sorry for not digging into it further and not isolating the specific commands which cause the crash. However, the config is only minimally changed from the default config (last 4 lines are relevant).
Also, please tell me how I could improve the issue title.

Sway version: `sway version 1.0-beta.1-256-ga22d0c0f (Nov 28 2018, branch 'master')`

Sway config: https://gist.github.com/camoz/dd915707940dd72d386e1ea8b1546e11
Relevant are the four lines.

# Steps to reproduce
* start sway with attached config file (see last 4 lines)
* press $mod+Shift+backslash (which will run the command `termite --name="scratchterm_full"`)
# What happens
Sway crashes, see debug log: https://gist.github.com/camoz/99a4b09179189138163a39436634b461
# What should happen
Sway should not crash. Interestingly, when I press $mod+backslash, it works (which will run `termite --name="scratchterm_full`)
Here's a log with just $mod+backslash pressed: https://gist.github.com/camoz/9cb9d7d0719c731e4e56df42452bf8f3

The difference is obviously the size of the window:
`for_window [app_id="scratchterm"] floating enable; border none; resize set 100 ppt 50 ppt; move absolute 0 23; move scratchpad; scratchpad show`
`for_window [app_id="scratchterm_full"] floating enable; border none; resize set 100 ppt 100 ppt; move absolute 0 23; move scratchpad; scratchpad show`

RedSoxFan:
Fixes #3236

This fixes the handling of hidden scratchpad containers for some
commands. For the most part, this just prevents running the commands on
hidden scratchpad containers, but there are some commands that have some
special handling for them.

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -589,6 +589,11 @@ struct cmd_results *cmd_resize(int argc, char **argv) {
 	if (!current) {
 		return cmd_results_new(CMD_INVALID, "Cannot resize nothing");
 	}
+	if (current->scratchpad && !current->workspace) {
+		return cmd_results_new(CMD_FAILURE,
+				"Cannot resize a hidden scratchpad container");
+	}
+
 
 	struct cmd_results *error;
 	if ((error = checkarg(argc, "resize", EXPECTED_AT_LEAST, 2))) {
--- a/sway/commands/floating.c
+++ b/sway/commands/floating.c
@@ -40,6 +40,11 @@ struct cmd_results *cmd_floating(int argc, char **argv) {
 		}
 	}
 
+	if (container->scratchpad && !container->workspace) {
+		return cmd_results_new(CMD_FAILURE,
+				"Cannot set floating status on a hidden scratchpad container");
+	}
+
 	bool wants_floating = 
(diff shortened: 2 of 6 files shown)


emersion:
Thanks!

sway/commands/resize.c
```diff
@@ -589,6 +589,11 @@ struct cmd_results *cmd_resize(int argc, char **argv) {
 	if (!current) {
 		return cmd_results_new(CMD_INVALID, "Cannot resize nothing");
 	}
+	if (current->scratchpad && !current->workspace) {
+		return cmd_results_new(CMD_FAILURE,
+				"Cannot resize a hidden scratchpad container");
+	}
+
```
Emantor: This breaks a floating terminal which was resized using a `for_window [app_id="quick_scratchpad"] mark quick, move scratchpad, resize set 1600 900` directive. The resize command is no longer applied (but worked before).

PR #3432, opened 2019-01-15
cursor: send clear_focus on hide and enter event on unhide

### LINKED ISSUE #3431: Send wl_pointer_leave when hide_cursor activates
So that e.g. browsers don't show tooltips.

Emantor:
Clear the focus when we hide the cursor and show it again during the unhide
action. The unhide function will send an enter event to the currently focused
window.

Tested by looking at the WAYLAND_DEBUG=1 output of termite.

Fixes #3431

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -275,6 +275,7 @@ static int hide_notify(void *data) {
 	struct sway_cursor *cursor = data;
 	wlr_cursor_set_image(cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
 	cursor->hidden = true;
+	wlr_seat_pointer_clear_focus(cursor->seat->wlr_seat);
 	return 1;
 }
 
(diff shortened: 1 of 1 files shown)


Emantor:
Looking at the code again we sent the enter event after we move the cursor. Are there edge cases where this can be a problem?

Emantor:
I still have input issue in firefox in xwayland, not sure if thats a bug introduced by this PR or an xwayland bug...

sway/input/cursor.c
```diff
@@ -313,6 +314,13 @@ void cursor_unhide(struct sway_cursor *cursor) {
 		cursor->image = NULL;
 		cursor_set_image(cursor, image, cursor->image_client);
 	}
+	struct sway_node *focus = seat_get_focus(cursor->seat);
```
RedSoxFan: I think this should be replaced with `cursor_rebase(cursor)` so that it handles layer surfaces as well

emersion:
>I still have input issue in firefox in xwayland, not sure if thats a bug introduced by this PR or an xwayland bug...

This bug also happens when switching workspaces, so I don't think it's related.

sway/input/cursor.c
Emantor: Nice catch, this fixed the input bug I was experiencing.

RedSoxFan:
Thanks!

Issue #3433, opened 2019-01-15
sporadic crash on switching from one monitor to another


sanga:
I don't like having both my laptop screen and external monitor running at the same time so I wrote a couple of small scripts to switch between them. When switching from the external monitor to the internal laptop screen I sometimes get a crash. This has started rather recently. The script does this:
```
sway output HDMI-A-1 disable
sway output eDP-1 enable
```

Apologies, but I don't have a debug log. I just run sway from gdm. I guess I could edit wherever the conf file for the gdm entry is to get the logs too, but I'm hoping this is clear enough from the stack. If not please let me know.

version info:
```
sway version 1.0-beta.2-195-g023c9242 (Jan 15 2019, branch 'master')
wlroots @ dc9c91b0a4a973f100ead0c2b1dad9f0699a8a69
```
stack trace:

```#0  0x000055f7511b3c13 in ipc_has_event_listeners (event=IPC_EVENT_SHUTDOWN) at ../sway/ipc-server.c:264
        i = 0
#1  0x000055f7511b4167 in ipc_event_shutdown (reason=0x55f7511fc418 "exit") at ../sway/ipc-server.c:377
        json = 0x7ffca55a7e90
        json_string = 0x7f3940620e40 <_wlr_vlog+41> "\220\311\303UH\211\345H\201\354", <incomplete sequence \340>
#2  0x000055f7511b60ed in sway_terminate (exit_code=1) at ../sway/main.c:36
#3  0x000055f7511f6d6d in _sway_abort (format=0x55f751205ac0 "[%s:%d] Unable to receive IPC response") at ../common/log.c:13
        args = {{gp_offset = 24, fp_offset = 48, overflow_arg_area = 0x7ffca55a7f90, reg_save_area = 0x7ffca55a7ed0}}
#4  0x000055f7511f6984 in ipc_recv_response (socketfd=3) at ../common/ipc-client.c:71
        received = -1
        data = '\000' <repeats 21 times>
        data32 = 0x7ffca55a7f96
        total = 0
        response = 0x0
        payload = 0x55f7511f684e <ipc_open_socket+182> "\203\370\377u/H\215=", <incomplete sequence \362>
#5  0x000055f7511f6c67 in ipc_single_command (socketfd=3, type=0, payload=0x55f751c421b0 "output eDP-1 enable", len=0x7ffca55a80b8) at ../common/ipc-client.c:127
        data = "i"
        data32 = 0x7ffca55a8026
        resp = 0x0
[...]

PR #3434, opened 2019-01-15
Do not send ipc_event_shutdown from ipc client

RedSoxFan:
Related to #3433 

In `sway_terminate`, `ipc_event_shutdown` was being sent regardless
which mode sway was running in. When running as an ipc client,
`sway_terminate` should just exit


emersion:
Thanks!

PR #3437, opened 2019-01-15
seatop_begin_down: raise floating

### LINKED ISSUE #3436: Bringing floating windows to the front
Steps to reproduce:
- Open two floating windows, position so they are slightly overlapping.
- Move mouse over to window which is behind (note: as i have focus_follows_mouse yes - this will transfer focus)
- Click into the window behind to bring to the front.

Expected:
Window behind is brought to the front

Actual:
Require to alt+click into the window ( floating_modifier is Mod1 )

This seems to be have been a change in behaviour within the last couple of days, alt+clicking was not required, I shall do some digging and see if I can nail this down to a certain commit.

[neil@desktop ~]$  swaymsg -t get_version
sway version 1.0-beta.2-189-g2573606b (Jan 14 2019, branch 'master')

RedSoxFan:
Fixes #3436 

In `seatop_begin_down`, raise the floating container. This appears to
have been dropped in the transition to seatops (https://github.com/swaywm/sway/pull/3402/files#diff-57d5973ad58d843753d353531e8cb85eL1039)

### DIFF
--- a/sway/input/seatop_down.c
+++ b/sway/input/seatop_down.c
@@ -74,4 +74,6 @@ void seatop_begin_down(struct sway_seat *seat,
 	seat->seatop_impl = &seatop_impl;
 	seat->seatop_data = e;
 	seat->seatop_button = button;
+
+	container_raise_floating(con);
 }


emersion:
Thanks!

PR #3438, opened 2019-01-16
Remove usage of VLAs.

c-edw:
Kernel contributors have been discontinuing use of VLAs for a variety of reasons outlined [here](https://lwn.net/Articles/749064/). 

There aren't many usages of VLAs within Sway fortunately. Two of those usages were committed by me, which is a bit ironic.

I have also added the `-Wvla` GCC flag which will warn in case of any future usages of VLAs.


c-edw:
Fixing some bugs with this currently.

c-edw:
Since `ipc_header_size` is defined as `sizeof(ipc_magic)+8`, I don't know how how it would be declared using macros. Relying on somebody to recalculate `ipc_header_size` after changing `ipc_magic` isn't a good idea of course.

In my novice opinion, I think there's nothing wrong with the current solution.

emersion:
```c
#define IPC_HEADER_SIZE (sizeof(ipc_magic) + 8)
```

For the record, kernel devs are replacing most VLAs with static arrays.

c-edw:
Ah okay, I didn't think `sizeof` could be called in that context. Good to know, thanks.

swaybar/ipc.c
```diff
@@ -13,9 +13,10 @@
 void ipc_send_workspace_command(struct swaybar *bar, const char *ws) {
 	const char *fmt = "workspace \"%s\"";
 	uint32_t size = snprintf(NULL, 0, fmt, ws);
-	char command[size];
+	char *command = malloc(sizeof(char) * size);
```
emersion: `snprintf` returns the number of chars printed, excluding the terminating NULL byte. This bug was there before, but we need to add `+ 1` here.

swaynag/swaynag.c
```diff
@@ -28,9 +28,10 @@ static bool terminal_execute(char *terminal, char *command) {
 	fprintf(tmp, "#!/bin/sh\nrm %s\n%s", fname, command);
 	fclose(tmp);
 	chmod(fname, S_IRUSR | S_IWUSR | S_IXUSR);
-	char cmd[strlen(terminal) + strlen(" -e ") + strlen(fname) + 1];
+	char *cmd = malloc(sizeof(char) * (strlen(terminal) + strlen(" -e ") + strlen(fname) + 1));
 	sprintf(cmd, "%s -e %s", terminal, fname);
 	execl("/bin/sh", "/bin/sh", "-c", cmd, NULL);
+	free(cmd);
 	return true;
```
emersion: Again, a bug that was here before. `execl` only returns if it fails. We should log the error (`wlr_log_errno`) and return false in this case.

common/pango.c
```diff
@@ -10,8 +10,9 @@
 #include "log.h"
 #include "stringop.h"
 
+#define MAX_CHARS 16384
```
emersion: The stack size on Alpine in 80K. We should be careful not to allocate too-large objects on the stack. 16K seems kind of okay considering there are not nested calls to these functions, but anything larger would be too large.

ianyfan:
`sizeof(char)` is not technically required since it is always guaranteed to be 1, but that's splitting hairs at this point.

A related point, which is outside the scope of this PR, but may be an interesting point of discussion for the devs here in terms of standardising style is whether to use the variable name instead of the type for allocing, e.g. changing `struct c v = malloc(sizeof(struct c));` to `struct c v = malloc(sizeof(*v));`. Futhermore, whether to use sizeof as an expression more, andwhether to always use calloc, or even to use calloc when you know you're about to initialise the memory.

c-edw:
@emersion Nice job spotting those bugs.
@ianyfan While it's true that `sizeof(char)` will always be 1, I think it conveys more meaning to developers. 


```c
char *command = malloc(sizeof(char) * (size + 1));
```

common/pango.c
c-edw: 16K does seem excessive. In commit cab13 it was just 2K.

emersion: It has been increased because some strings were truncated. We may just want to use dynamic allocation depending on the length of the string in this case.

ascent12:
>While it's true that sizeof(char) will always be 1, I think it conveys more meaning to developers

The intent is already obvious from the context. `sizeof(char)` is unnecessary "noise" and usually comes off as a newbie mistake.

emersion:
>The intent is already obvious from the context. sizeof(char) is unnecessary "noise" and usually comes off as a newbie mistake.

I'm fine with either. I don't see why `char` would be a special case.

common/pango.c
c-edw: I'll address this in another PR as long as that's alright with you.

emersion: Fine by me

ianyfan:
@c-edw That's a very valid point that I'm not going to argue with, but I'll offer the counter-argument that if you're allocating on the same line as declaring, the dev can already see that it's a string, and therefore that it is allocating chars, which are equivalent size to bytes.  As I said, it's splitting hairs so I don't mind if it's left in, but I think that if writing `sizeof(char)` would cause the line to exceed the line width, it would in most cases be more readable to remove the `sizeof(char)`rather than split the line.

@emersion Your point for using calloc is a very good one that I did not think about when writing my comment, thanks. 

> whether to use sizeof as an expression more

Ignore this, I got confused at parentheses usage for sizeof.

> The first one is a little more readable, the second one ensures there's no mismatch (though I think many tools are able to catch those cases, at least Clang's static analyzer).

I believe an argument against the latter is that if you change the type of the variable, then you need to go through the code and change every occurrence of allocation. And I think that if you miss one, then it won't be reported as an error, though if this is picked up by compilers or static analysers (which clearly I should be using more) then ignore this point.

c-edw:
@ascent12 That's a shame, it really looks clearer to me. I think it would also reduce the risk of somebody coming along and changing the type, but forgetting to change the allocated size. Maybe that's something you guys already watch out for.

ianyfan:
`char` is a special case because it is the only type (I think) that is guaranteed by the C standard, as 1 no less, which means it is redundant if used as a multiplier is `sizeof`.

emersion:
Thanks!

PR #3441, opened 2019-01-16
Use noop output when there's no outputs connected

RyanDwyer:
Instead of having `NULL` `workspace->output` pointers, use a <s>headless</s> noop output. This should be safer.

I've tested this with the wayland backend by closing the output window and then using swaymsg to connect to the socket and run the `create_output` command. Upon creation, I had some issues with the alignment of the first output's contents within the output window, but I'm not sure this is related to my change.

I have not tested on DRM yet.

The output is returned in the IPC `get_outputs` list, so perhaps we should hide it. Also, we never enable the output (we never have to), so perhaps we should add a check to make sure it can't be enabled using config or IPC.

### DIFF
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -1,3 +1,4 @@
+#include <strings.h>
 #include "sway/commands.h"
 #include "sway/config.h"
 #include "sway/output.h"
@@ -24,6 +25,13 @@ struct cmd_results *cmd_output(int argc, char **argv) {
 	struct cmd_results *error = checkarg(argc, "output", EXPECTED_AT_LEAST, 1);
 	if (error != NULL) {
 		return error;
+	}
+
+	// The NOOP-1 output is a dummy output used when there's no outputs
+	// connected. It should never be configured.
+	if (strcasecmp(argv[0], "NOOP-1") == 0) {
+		return cmd_results_new(CMD_FAILURE,
+				"Refusing to configure the no op output");
 	}
 
 	struct output_config *output = new_output_config(argv[0]);
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -29,7 +29,7 @@ struct cmd_results *cmd_output(int argc, char **argv) {
 
 	// The NOOP-1 output is a dummy output used when there's no outputs
 	// connected. It should never be configured.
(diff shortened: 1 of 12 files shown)


emersion:
This sounds kind of dangerous. Maybe we could have a noop backend instead (without a renderer).

ddevault:
Rebase?

RyanDwyer:
Rebased. I'll do testing with the DRM backend this weekend. Please don't merge until then.

RyanDwyer:
DRM tested fine, no crashes. I hotplugged from two outputs down to zero and back to two, and also ran `swaymsg -t get_outputs` via SSH from another device while there were no outputs connected.

This is still relevant:

> The output is returned in the IPC get_outputs list, so perhaps we should hide it. Also, we never enable the output (we never have to), so perhaps we should add a check to make sure it can't be enabled using config or IPC.

emersion:
>The output is returned in the IPC get_outputs list, so perhaps we should hide it. Also, we never enable the output (we never have to), so perhaps we should add a check to make sure it can't be enabled using config or IPC.

+1

Do we do something similar for fake workspaces?

RyanDwyer:
Both of those are done now. I'm not sure what you mean by fake workspaces. When there's no outputs connected, `swaymsg -t get_workspaces` shows nothing.

RedSoxFan:
I think the guard needs to be added to `apply_output_config` as well to prevent against enabling with `output * enable`. It might also make sense to move the existing guard to cmd_output just to be safe for all subcommands

emersion:
>I'm not sure what you mean by fake workspaces

Oh, right. We just move workspaces in `saved_workspaces`.

sway/commands/output.c
```diff
@@ -26,6 +27,13 @@ struct cmd_results *cmd_output(int argc, char **argv) {
 		return error;
 	}
 
+	// The NOOP-1 output is a dummy output used when there's no outputs
+	// connected. It should never be configured.
+	if (strcasecmp(argv[0], "NOOP-1") == 0) {
```
emersion: Can we check against `root.noop_output->name` instead? I'm worried about someone using the noop beackend of unit tests and running into this.

emersion:
Thanks!

PR #3447, opened 2019-01-17
Remove swaylock headers and unicode.c

emersion:


ianyfan:
I think `unicode.c` can be removed as well.

ddevault:
Thanks!

PR #3451, opened 2019-01-17
Cleanup stringop.c (partly)

ianyfan:
I've removed a lot of old and unused functions.

I want to refactor some of the other functions. In particular `unescape_string`, and maybe move it to `sway/commands.c` since it is the only place it is used, but I'm not sure about its specification to be confident in rewriting it correctly.

### DIFF
--- a/common/stringop.c
+++ b/common/stringop.c
@@ -1,13 +1,13 @@
 #define _POSIX_C_SOURCE 200809L
-#include <stdlib.h>
+#include <ctype.h>
+#include <stdbool.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <strings.h>
-#include <ctype.h>
-#include "stringop.h"
-#include "log.h"
-#include "string.h"
 #include "list.h"
+#include "log.h"
+#include "stringop.h"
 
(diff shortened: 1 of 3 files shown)


emersion:
Ping: maybe we should first merge this, and do the remaining work in another PR?

ianyfan:
@emersion I'm fine with that; the branch has been updated and rebased.

common/stringop.c
```diff
@@ -64,12 +64,8 @@ char *lenient_strncat(char *dest, const char *src, size_t len) {
...
 // strcmp that also handles null pointers.
 int lenient_strcmp(char *a, char *b) {
-	if (a == b) {
-		return 0;
-	} else if (!a) {
-		return -1;
-	} else if (!b) {
-		return 1;
+	if (!a || !b || a == b) {
+		return a - b;
```
emersion: Not a fan of this. This casts a `size_t` into an `int`. (I'm not even sure the cast happens _after_ the subtraction)

ianyfan: I looked it up, and this is undefined behaviour, so I will revert it.

emersion:
Thanks!

PR #3452, opened 2019-01-17
Use Meson feature options

emersion:


emersion:
Updated.

>I also question the value of a lot of the options like sway-version, use_rpath and custom_rpath.

+1. I think sway-version is still useful if you download a tarball of sway, so I kept it.

>Make it possible to explicitly select libsystemd vs libelogind.

Will do in another PR. 

>The wlroots option for this could also be yielding. Perhaps some detection if sway/wlroots are trying to link against different ones from each other would be warranted.

I'm not sure we want to protect against linking different ones at all costs. Sway only uses these for the sd-bus lib, so I'd rather have a 'sd-bus-provider' option in sway than a 'logind-provider' one (a library might offer sd-bus without logind).

>Make 'xwayland' a yielding option inside of wlroots (This is useful when wlroots is being used as a subproject).

Makes sense.

emersion:
>Make 'xwayland' a yielding option inside of wlroots (This is useful when wlroots is being used as a subproject).

Well, how does it work if 'xwayland' is set to 'auto', it's enabled in sway because 'xcb' is found but it's disabled in wlroots because some other dependency is not found?

ascent12:
>Well, how does it work if 'xwayland' is set to 'auto', it's enabled in sway because 'xcb' is found but it's disabled in wlroots because some other dependency is not found?

Probably. We should add [some stuff](https://mesonbuild.com/Reference-manual.html#compiler-object) (`get_define`) reading wlroots' config.h to make sure we're not linking against a wlroots with optional features compiled out.
That would also be the way to see if wlroots was compiled with libsystemd or libelogind.

>I'm not sure we want to protect against linking different ones at all costs.

The point is that it's a linker error, since you'd have multiple definitions of the same symbol.

>Sway only uses these for the sd-bus lib, so I'd rather have a 'sd-bus-provider' option in sway than a 'logind-provider' one (a library might offer sd-bus without logind).

Yeah, `logind-provider` is not a good name here, but I can't really think of a good middle-ground name between the two.  It would be nice if they were the same option.
It's not like it's really that big of a deal, considering that it'd be rare for someone to have both libraries.

ddevault:
Thanks!

emersion:
`get_define` doesn't (yet) work for this use-case: https://github.com/mesonbuild/meson/issues/4800

PR #3454, opened 2019-01-17
output: remove output_add_listeners

emersion:
Simplify the code by registering signals when outputs are created and removing
signals when they are destroyed.


RyanDwyer:
What do you think of moving `output_enable` and `output_disable` into `tree/desktop/output.c` so they're in the same file as the handlers, and keeping the listener registration in those functions? It means the `if (output->enabled)` checks won't be needed, and it could register the listeners directly. Are there any other benefits to registering them on output creation?

PR #3455, opened 2019-01-17
Better handle outputs without CRTC

emersion:
~~Depends on #3454~~

This happens if you plug in more outputs than supported by your GPU.

This patch makes it so outputs without CRTCs appear as disabled. As soon as
they get a CRTC (signalled via the mode event), we can enable them.


emersion:
This is now ready for review.

Emantor:
Time to try my Intel MST setup again!

PR #3460, opened 2019-01-18
Fix segfault in output_enable

emersion:


emersion:
Actually this is still nice to have, to prevent segfaults when enabling, disabling and enabling again an output.

ddevault:
Thanks!

Issue #3462, opened 2019-01-18
sway segfaults on startup


m0rr1gan:
Sway segfaults when starting from a tty with default config file.

sway -v:
sway version 1.0-beta.2-218-g60220fce (Jan 18 2019, branch 'master')

sway -c /etc/sway/config -d:
https://ptpb.pw/3N5G

coredump gdb sway and bt full:
https://ptpb.pw/90TY

I am running the latest version of sway and wlroots.
https://github.com/swaywm/sway/pull/3460 does not fix this for me

emersion:
Can you try https://github.com/swaywm/sway/pull/3463?

emersion:
If it doesn't fix it, please run recompile sway with `meson build-asan -Db_sanitize=address && build-asan/sway/sway`

mnd999:
Does temporarily commenting out your output lines fix it? I've had the same thing and that fixes it for me.
I am able to add them back in once sway is up and running.

emersion:
@mnd999 Please try the PR if you can reproduce the crash.

mnd999:
@emersion No, that patch doesn't help me either.

My backtrace:

```
Reading symbols from /home/mark/dev/sway/build/sway/sway...done.
[New LWP 2626]
[New LWP 2632]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
Core was generated by `dev/sway/build/sway/sway'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000056339d1498ea in arrange_output (output=0x56339dd87100) at ../sway/tree/arrange.c:243
243		output->lx = output_box->x;
[Current thread is 1 (Thread 0x7eff316ef9c0 (LWP 2626))]
(gdb) bt
#0  0x000056339d1498ea in arrange_output (output=0x56339dd87100) at ../sway/tree/arrange.c:243
#1  0x000056339d119fb4 in arrange_layers (output=0x56339dd87100) at ../sway/desktop/layer_shell.c:180
#2  0x000056339d11c41d in handle_transform (listener=0x56339dd87218, data=0x56339dc83190)
    at ../sway/desktop/output.c:530
#3  0x00007eff33cd307e in wlr_signal_emit_safe (signal=0x56339dc83320, data=0x56339dc83190)
    at ../util/signal.c:29
#4  0x000056339d13181d in apply_output_config (oc=0x56339dc78b80, output=0x56339dd87100)
[...]

PR #3463, opened 2019-01-18
[DO NOT MERGE] Fix crash in cursor_rebase with multiple screens

emersion:
I tried fixing this crash in the most logical way possible. Let me know if you have other ideas.

Designing the output configuration sequence without invalid state is tricky.

We have one function, apply_output_config, that takes an output and (besides
other things) performs a modeset and inserts the output in the output layout.
The modeset can fail, in which case we don't want the output to be enabled.
We also have an output_enable function, which calls output_apply_config and
also configures the output's workspace and inserts it in the root container.

Now, we have two choices.

Either we configure the output before it's been inserted in the root container
and then, if the modeset was successful, we insert it and create the workspace.
The main issue with this approach is that configuring the output triggers a
handful of signals, namely wlr_output.mode and wlr_output_layout.change. In
those event handlers, we need to make sure to ignore these outputs in the
process of being configured.

Either we first insert the output, create the workspace and then try to
configure it. It means we need to undo everything if the modeset fails. The
main issue with this solution is that it enables and disables the output very
quickly, creates a workspace and immediately destroys it, and maybe moves
views back and forth (see output_evacuate).

I've tried to make it so an output isn't enabled then immediately disabled. We
already have code for ignoring outputs when the output is being destructed.

Fixes https://github.com/swaywm/sway/issues/3462

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -83,8 +83,8 @@ struct sway_node *node_at_coords(
 		return NULL;
 	}
 	struct sway_output *output = wlr_output->data;
-	if (!output) {
-		// output is being destroyed
+	if (!output || !output->configured) {
+		// output is being destroyed or is being configured
 		return NULL;
 	}
 	double ox = lx, oy = ly;


RyanDwyer:
My preference is the first option (configure then insert to tree) because it won't trigger a bunch of unnecessary IPC events if the configure fails.

emersion:
Okay, that's what's implemented so far. I think we could simplify the code a lot more though.

m0rr1gan:
Also works for me.
Fixes https://github.com/swaywm/sway/issues/3462

PR #3464, opened 2019-01-18
Make sure wlroots has been built with xwayland

emersion:


ascent12:
We could just guard this check with `if wlroots.type_name() == 'internal'` or (perhaps `!= 'pkgconfig'`) and not have to do most of that extra stuff.

emersion:
Nah, because we need access to the `conf_data` variable of the subproject.

ascent12:
With the xwayland option now being yielding, I don't think it's necessary to read it from the subproject.

emersion:
>With the xwayland option now being yielding, I don't think it's necessary to read it from the subproject.

Well, the issue is if xwayland is set to auto, which resolves to enabled in sway and disabled in wlroots.

ascent12:
Oh right, that is an issue. That makes sense.

I'm not really a fan of making the subproject being preferred over the system dependency, though. I'd prefer it to be consistent with Meson's behaviour.

PR #3466, opened 2019-01-19
Fix memory leaks in swaybar tray initialization

mstoeckl:
Memory leaks were found with AddressSanitizer. Two were simple unfreed allocations; the other was a call to `opendir` without a corresponding call to `closedir`.


emersion:
Thanks!

PR #3469, opened 2019-01-19
swaybg: remove assert crashing on hotplug

emersion:
When hotplugging, an output may disappear and re-appear. This commit makes it
so swaybg doesn't crash anymore in this case.


emersion:
Yes, but this is racy.

emersion:
Wait, maybe not, since we `waitpid`. So we're probably sending multiple name events for the same output...

emersion:
A better fix is here: https://github.com/swaywm/wlroots/pull/1489

PR #3473, opened 2019-01-19
Allocate minimum size necessary in pango text functions.

c-edw:
Almost forgot about this. See [here](https://github.com/swaywm/sway/pull/3438/files/338a8c051df8290eb7863ba0c566139747b7b224#r248228815) (annoyingly this doesn't link directly to the comment. It's in one of the resolved conversations...).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ install_manifest.txt
 bin/
 test/
 build/
+build-*/
 .lvimrc
 config-debug
 wayland-*-protocol.*
--- a/common/pango.c
+++ b/common/pango.c
@@ -10,10 +10,6 @@
 #include "log.h"
 #include "stringop.h"
 
-#define MAX_CHARS 16384
-
-static const char overflow[] = "[buffer overflow]";
-
 size_t escape_markup_text(const char *src, char *dest) {
 	size_t length = 0;
 	if (dest) {
@@ -88,13 +84,15 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
 
 void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
 		int *baseline, double scale, bool markup, const char *fmt, ...) {
-	char buf[MAX_CHARS];
-
 	va_list args;
 	va_start(args, fmt);
-	if (vsnprintf(buf, sizeof(buf), fmt, args) >= MAX_CHARS) {
-		strcpy(&buf[sizeof(buf) - sizeof(overflow)], overflow);
-	}
+	// Add one since vsnprintf excludes null terminator.
+	int length = vsnprintf(NULL, 0, fmt, args) + 1;
+	va_end(args);
+
+	char *buf = malloc(length);
+	va_start(args, fmt);
+	vsnprintf(buf, length, fmt, args);
 	va_end(args);
 
 	PangoLayout *layout = get_pango_layout(cairo, font, buf, scale, markup);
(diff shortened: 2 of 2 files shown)


common/pango.c
```diff
@@ -88,14 +89,22 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
...
 	va_list args;
 	va_start(args, fmt);
-	if (vsnprintf(buf, sizeof(buf), fmt, args) >= MAX_CHARS) {
-		strcpy(&buf[sizeof(buf) - sizeof(overflow)], overflow);
-	}
+	// Add one since vsnprintf excludes null terminator.
+	int length = vsnprintf(NULL, 0, fmt, args) + 1;
 	va_end(args);
+	// Limit length of string to MAX_CHARS.
+	int length_limited = imin(length, MAX_CHARS);
```
emersion: TBH I don't think limiting to `MAX_CHARS` is necessary.

c-edw: Unlimited size? I guess the only benefit is protection against oversized strings created from a bug or something like that.

emersion: Yeah, we don't do this elsewhere, let's not do it here.

.gitignore
```diff
@@ -5,6 +5,8 @@ install_manifest.txt
 bin/
 test/
 build/
+build-gcc/
```
c-edw: May I add these? It makes things easier when testing for each compiler.

c-edw:
Anomaly is causing CI build failure. Code compiles.

```
...
(1/5) Probing GDK-Pixbuf loader modules...
(2/5) Reloading system manager configuration...
(3/5) Reloading device manager configuration...
(4/5) Arming ConditionNeedsUpdate...
(5/5) Updating the MIME type database...
:: Searching databases for updates...
:: Searching AUR for updates...
Get https://aur.archlinux.org/rpc.php?arg%5B%5D=yay-bin&type=info&v=5: dial tcp 5.9.250.164:443: i/o timeout
```

include/util.h
```diff
@@ -8,7 +8,12 @@
 #include <xkbcommon/xkbcommon.h>
 
 /**
- * Wrap i into the range [0, max[
+ * Return the minimum (lowest value) of two integers.
+ */
+int imin(int a, int b);
```
mstoeckl: This function is not used anywhere?

.gitignore
emersion: Not a fan of this.

c-edw: What can I do instead? Without this, I was having to do:

```
CC=gcc meson build
rm -rf build
...
CC=clang meson build
...
rm -rf build
```

emersion: I'd prefer something like `/build-*` instead.

c-edw:
Yep, just force pushed.

include/util.h
```diff
@@ -8,7 +8,7 @@
 #include <xkbcommon/xkbcommon.h>
 
 /**
- * Wrap i into the range [0, max[
+ * Wrap i into the range [0, max]
```
emersion: I think "[0, max[" is correct.

Unrelated to this PR: I wonder why it's a function actually? Isn't `((i % max) + max) % max` the same as `i % max`?

c-edw: I believe that in C, `%` calculates the remainder. `((i % max) + max) % max` calculates the modulus.

emersion: `%` _is_ the modulus operator…

common/pango.c
```diff
@@ -88,13 +84,15 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
...
 	va_list args;
 	va_start(args, fmt);
-	if (vsnprintf(buf, sizeof(buf), fmt, args) >= MAX_CHARS) {
-		strcpy(&buf[sizeof(buf) - sizeof(overflow)], overflow);
-	}
+	// Add one since vsnprintf excludes null terminator.
+	int length = vsnprintf(NULL, 0, fmt, args) + 1;
+	va_end(args);
+
+	char *buf = malloc(length);
```
ddevault: Need to handle malloc failure

common/pango.c
```diff
@@ -104,17 +102,20 @@ void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
...
```
ddevault: Here as well

common/pango.c
```diff
@@ -88,13 +84,15 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
...
 		int *baseline, double scale, bool markup, const char *fmt, ...) {
-	char buf[MAX_CHARS];
-
 	va_list args;
 	va_start(args, fmt);
-	if (vsnprintf(buf, sizeof(buf), fmt, args) >= MAX_CHARS) {
-		strcpy(&buf[sizeof(buf) - sizeof(overflow)], overflow);
-	}
+	// Add one since vsnprintf excludes null terminator.
+	int length = vsnprintf(NULL, 0, fmt, args) + 1;
```
emersion: Nitpick: I prefer to use "length" (or "len") to mean the return value of `strlen` (aka without the terminating NULL byte) and "size" to mean the value passed to `malloc` et al (aka including the terminating NULL byte).

c-edw:
In what cases do allocation failures need to be handled? Is it always? If that's the case, is that something that wasn't enforced in the past but is now?

common/pango.c
c-edw: Yep, that's fair. I can change it if you really want.

ddevault:
Just where it's important. In this case it's possibly exploitable by an evil client which sets its title to something ridiculous that we can't draw.

emersion:
In a perfect world, the compositor would survive even if the device is running low on memory.

emersion:
Thanks!

PR #3474, opened 2019-01-19
Remove wlroots dependency for swaymsg

mstoeckl:
`swaymsg` only used the logging functions of wlroots, and those sparingly. This PR removes the dependency on the wlroots library, and its transitive dependencies on ~40 other libraries. In my case, startup time drops from ~15ms to ~3ms, and the executable size shrinks by ~10k bytes.

I'm not to certain whether is best to warn, abort, or stay silent in the various places where the ipc client functions are used.

### DIFF
--- a/common/meson.build
+++ b/common/meson.build
@@ -20,3 +20,12 @@ lib_sway_common = static_library(
 	],
 	include_directories: sway_inc
 )
+
+lib_sway_ipc_client = static_library(
+	'sway-ipc-client',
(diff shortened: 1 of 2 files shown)


ddevault:
I would rather just provide, and link to, an alternate implementation of the `wlr_*` log functions (or `sway_*`), than do things like passing an error string reference.

emersion:
This needs a rebase.

PR #3478, opened 2019-01-20
Replace wlr_log with sway_log

mstoeckl:
Following @ddevault 's recommendation in #3474 , this PR replaces `wlr_log` with `sway_log`,
and WLR_(ERROR|DEBUG|INFO) with SWAY_(ERROR|DEBUG|INFO). See also #2009.

(Note: I'd really like to replace `SWAY_SRC_DIR` +` _sway_strip_path` with a per-file `PROJECT_FILE_PATH` macro, since  the former combination leaks the path into the executable, and calls the slow function `strstr` at runtime. It's not obvious how to do this with meson.)

### DIFF
--- a/include/log.h
+++ b/include/log.h
@@ -1,7 +1,18 @@
 #ifndef _SWAY_LOG_H
 #define _SWAY_LOG_H
+
 #include <stdbool.h>
-#include <wlr/util/log.h>
+#include <stdarg.h>
+#include <string.h>
+#include <errno.h>
+
+typedef enum {
+	SWAY_SILENT = 0,
+	SWAY_ERROR = 1,
+	SWAY_INFO = 2,
+	SWAY_DEBUG = 3,
+	SWAY_LOG_IMPORTANCE_LAST,
+} sway_log_importance_t;
 
 #ifdef __GNUC__
 #define ATTRIB_PRINTF(start, end) __attribute__((format(printf, start, end)))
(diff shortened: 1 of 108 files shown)


include/log.h
```diff
@@ -1,22 +1,54 @@
...
+#include <stdarg.h>
+#include <string.h>
+#include <errno.h>
+
+typedef enum {
+	SWAY_SILENT = 0,
+	SWAY_ERROR = 1,
+	SWAY_INFO = 2,
+	SWAY_DEBUG = 3,
+	SWAY_LAST,
```
emersion: These could be `L_*` instead. `SWAY_LAST` is not a very good name.

meson.build
```diff
@@ -12,6 +12,8 @@ project(
 
 add_project_arguments(
 	[
+                                '-DSWAY_SRC_DIR="@0@"'.format(meson.current_source_dir()),
```
emersion: Please use tabs

mstoeckl: Done.

include/log.h
mstoeckl: I'm now using SWAY_LOG_IMPORTANCE_LAST, similar to WLR_LOG_IMPORTANCE_LAST.

emersion:
BTW, can you commit with your GitHub email so that the commit is linked with your account?

mstoeckl:
My usual commit email has been linked.

emersion:
Eh, sorry. Seems like a GitHub bug.

sway/main.c
```diff
@@ -303,25 +303,30 @@ int main(int argc, char **argv) {
 		}
 	}
 
+	// As the 'callback' function for wlr_log is equivalent to that for
+	// sway, we do not need to override it.
 	if (debug) {
-		wlr_log_init(WLR_DEBUG, NULL);
+		sway_log_init(SWAY_DEBUG, sway_terminate);
+		wlr_log_init(SWAY_DEBUG, NULL);
```
ddevault: This is wrong - wlr_log_init takes WLR_* arguments. You can also probably move this into sway_log_init and make the wlr log level implied by the sway log level (you'll need a mapping of sway loglevels to wlr loglevels)

mstoeckl: Thank you for catching that. As sway and wlroots are only used together by the main sway executable, not by swaymsg and others, having `sway_log_init` call `wlr_log_init` conditionally will be tricky to implement cleanly.

ddevault: Good point. Skip it, then.

emersion:
Thanks!

PR #3481, opened 2019-01-21
Log libinput_config_status errors

RedSoxFan:
This add errors from calls to `libinput_device_config_*_set` to the
debug log. This may be useful when looking at issues related to input
devices and config settings not working.


emersion:
Thanks!

PR #3486, opened 2019-01-21
systemd --user support files

gdamjan:
* adds a sway-session.target that binds to graphical-session.target
  (see man systemd.special)
* also adds a 10-systemd script that should (must?) be included from sway
  that starts imports the sway environment variables into the systemd
  --user instance, and then starts sway-session.target

The idea is that, any --user units that are wanted by either
sway-session.target or graphical-session.target will be started when
sway runs, and will have the correct environment from sway too.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -153,6 +153,9 @@ subdir('client')
 subdir('swaybg')
 subdir('swaybar')
 subdir('swaynag')
+if systemd.found()
+  subdir('systemd')
+endif
 
 config = configuration_data()
 config.set('datadir', join_paths(prefix, datadir))
--- a/systemd/10-systemd
+++ b/systemd/10-systemd
@@ -1,0 +1,4 @@
+# should go to /etc/sway/config.d/10-systemd
+#
+exec systemctl --user import-environment
+exec systemctl --user start sway-session.target
--- a/systemd/meson.build
+++ b/systemd/meson.build
@@ -1,0 +1,15 @@
+systemd_dep = dependency('systemd')
+
+if systemd_dep.found()
+  systemd_units = files(
+      'sway-session.target',
+  )
+  user_units_dir = systemd_dep.get_pkgconfig_variable('systemduserunitdir')
+  install_data(systemd_units, install_dir: user_units_dir)
+
+  sway_systemd_scripts = files(
+      '10-systemd',
+  )
+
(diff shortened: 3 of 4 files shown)


ascent12:
Instead of sway calling systemctl, you'd want to have it so systemd calls sway, like any other normal service.
Weston has an incomplete example of it: https://gitlab.freedesktop.org/wayland/weston/merge_requests/39 which you could take inspiration from.
But I also agree that we probably don't need a unit file upstream.

I don't know whether we want to add optional `sd_notify` support into sway, so that systemd is capable of knowing when sway has finished its startup and can order units after that. It'd be a trivially small amount of code, but it depends on whether you want any direct systemd things in the code.

gdamjan:
@ascent12 from my research it's not easy to start sway/weston as a --user service from a login manager, yet -- so you'd need to login via agetty, which is meh. Otherwise I agree, that would be a nicer approach.

gdamjan:
Let me just formally disagree with the decision here (before I go on with other things):

- the support I added is optional, doesn't subtract from any other OS at all
- I don't know of other init systems that have a --user session management, and both Gnome and KDE (at least), are using and improving their systemd --user support. So it's not a preference for a init system, it's about using features that exist on a platform vs not using them
- like it or not, systemd is a standard feature on Linux desktops including (Debian, Ubuntu, Arch, Fedora, RHEL and OpenSuse), but again this change doesn't subtract from any other OS.
- sending this same change to ALL distros is a huge effort, and promotes platform fragmentation. I'd claim that window managers supporting systemd linux distros should support [graphical-session.target](https://www.freedesktop.org/software/systemd/man/systemd.special.html#graphical-session.target) properly.

With that said, I'm off to fight other issues.

emersion:
>I don't know whether we want to add optional sd_notify support into sway, so that systemd is capable of knowing when sway has finished its startup and can order units after that. It'd be a trivially small amount of code, but it depends on whether you want any direct systemd things in the code.

Add this to your config:

```
exec systemd-notify --ready || true
```

jpsamaroo:
Maybe the wiki would be a better place for this script? It already has some distro-specific docs there (for Debian/Ubuntu), so I feel like putting init-specific scripts there would make sense.

ddevault:
+1, wiki would be a great place for this

PR #3488, opened 2019-01-22
root_scratchpad_remove_container: do not show 

### LINKED ISSUE #3461: Closing a window in the scratchpad crashes Sway
* Sway Version: 1.0-beta.2-218-g60220fce (Jan 18 2019, branch 'master')
* Stack Trace: https://gist.github.com/Bryophyllum/8c7e0e3366287ca48c6254b5212267bd

* Steps to reproduce:
-> Run any graphical program in a terminal
-> Move the focused window to the scratchpad
-> Press CTRL+C in the terminal to kill the program

I can't exactly pinpoint the exact day on which this bug could be introduced since the only window on my setup that closes itself while being in the scratchpad is Telegram's active call window and I haven't been making calls for over one week until today.

RedSoxFan:
Fixes #3461

This removes the call to `root_scratchpad_show` from
`root_scratchpad_remove_container` and places it in the
`cmd_move_container`. This also moved the IPC `window::move` event to
`cmd_scratchpad`.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -515,6 +515,7 @@ static struct cmd_results *cmd_move_container(int argc, char **argv) {
 	// move container
 	if (container->scratchpad) {
 		root_scratchpad_remove_container(container);
+		root_scratchpad_show(container);
 	}
 	switch (destination->type) {
 	case N_WORKSPACE:
--- a/sway/commands/scratchpad.c
+++ b/sway/commands/scratchpad.c
@@ -3,6 +3,7 @@
 #include "sway/config.h"
 #include "sway/input/input-manager.h"
 #include "sway/input/seat.h"
+#include "sway/ipc-server.h"
 #include "sway/tree/container.h"
 #include "sway/tree/root.h"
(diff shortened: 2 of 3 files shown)


RyanDwyer:
The core issue here is having `root_scratchpad_remove_container` call `root_scratchpad_show`. When destroying the container we don't really care about making it show, and this additionally triggers an unwanted `window::move` IPC event.

`root_scratchpad_remove_container` is called from three places:

* When destroying a container in `container_begin_destroy`, where we don't want to show it.
* In `container_set_floating`, where it's already showing (we don't allow unfloating a hidden scratchpad container using criteria).
* In `cmd_move_container`, where it needs to be shown.

A better fix is to remove the call to `root_scratchpad_show` from `root_scratchpad_remove_container` and add it to `cmd_move_container`. Additionally, remove the `window::move` IPC event from `root_scratchpad_show` because it's already called in the move command. It would need to go in the `scratchpad show` command though.

emersion:
Thanks!

PR #3493, opened 2019-01-22
util.c: remove numlen function

ianyfan:
Its uses have been replaced by snprintf, which is more in line with its usage.

Also cleaned headers.


ianyfan:
That's already done in https://github.com/swaywm/sway/pull/3451/

PR #3494, opened 2019-01-22
i3 command behaviour compatibility fixes

ianyfan:
1. Allow tiled containers to be stickied
This also stops stickied containers from losing its sticky status when
it is tiled, allowing it to be immediately stickied when floated again.*
2. Allow setting floating on scratchpad containers

* There's a quirk in i3 that if the container is not on the active workspace when floated (and is sticky), it is not moved to the current workspace, but sticks when switching to that workspace. This matches that behaviour, though I'm not sure if it is a bug.

I've tested it a bit, but not sure if I've missed some edge case.

Looking into it more, it seems like there are a number of commands that, if run on hidden scratchpad containers, result in failure on sway but success on i3 (though I haven't tested which ones), so if compatibility is the goal then current handling should be changed.

### DIFF
--- a/sway/commands/floating.c
+++ b/sway/commands/floating.c
@@ -45,7 +45,7 @@ struct cmd_results *cmd_floating(int argc, char **argv) {
 				"Cannot set floating status on a hidden scratchpad container");
 	}
 
-	bool wants_floating = 
+	bool wants_floating =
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks!

PR #3495, opened 2019-01-22
swaybar: fix workspace command

ianyfan:
Escape quotes and backslashes, allowing switching to workspace names
like "1" (including quotes) and \


ianyfan:
A bit strange that it was working before. Seems like there's a bug that 'workspace "1' behaves the same as 'workspace Y'

emersion:
Thanks!

PR #3496, opened 2019-01-22
Fix dead stores found by scan-build

mstoeckl:
`ninja scan-build` was used to invoke the [clang static analyzer](https://clang-analyzer.llvm.org/). 

In addition to removing unused code, two minor problems are fixed:
* `resize set` and `resize adjust` did not yield an error when given too many arguments. Now

        bindsym Ctrl+Return resize set width 400 px height 400 px
    
    still works, but the following fails:

        bindsym Ctrl+Return resize set width 400 px height 400 px oops

* `orientation` was incorrectly overridden to be 'U' for scroll events in the swaybar tray `handle_click` function.

### DIFF
--- a/client/pool-buffer.c
+++ b/client/pool-buffer.c
@@ -85,7 +85,6 @@ static struct pool_buffer *create_buffer(struct wl_shm *shm,
 	close(fd);
 	unlink(name);
 	free(name);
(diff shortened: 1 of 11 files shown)


ddevault:
Thanks!

PR #3498, opened 2019-01-22
swaybar: fix command malloc in workspace command

RedSoxFan:
Fixes https://github.com/swaywm/sway/commit/d8f3e6e19a12514ed6f8c8f470b89fde0f39dc59#commitcomment-32026156

This fixes a typo on the malloc line in ipc_send_workspace_command. The
plus one to the size for the null-terminator was outside of the malloc
call, which was causing the incorrect pointer to be freed later in the
function.


emersion:
Thanks!

PR #3500, opened 2019-01-22
swaybar: fix crash with some tray icon pixmaps

### LINKED ISSUE #3487: Launching xembedproxy crashes swaybar with tray enabled
Forgive me if I'm jumping the gun on testing the new tray with `xembedsniproxy`, but the bar crashes when I try to launch the proxy to get any of my icons to show up.

- Sway version: `sway version 1.0-beta.2-243-g410c9613 (Jan 21 2019, branch 'master')` (with a small patch to set `enable-tray` to `true`).
- xembedsniproxy version: `xembed-sni-proxy-git v5.14.90.r14.gad1a76b8-1`
- Debug log: https://termbin.com/z4kh
- Backtrace: https://termbin.com/gs12

All that I need to reproduce the problem is:

1. Start sway.
2. Start `xembedsniproxy`.
3. Start Chromium (with something which adds tray icons, i.e. running in the background or Hangouts).
4. swaybar crashes, a window is created for each tray icon, and a black box appears in the top left corner of the screen on all workspaces.

Interestingly through some combination of stopping the proxy, closing things, and starting again with Chromium already loaded, I can get it to not crash. As in, it's the same as step 4 above, but instead of crashing, I get the two icons which should have loaded in the bar (but very garbled and changing).

progandy:
Ignore pixmaps for icons where width or height is 0 and use the exact width and height to parse the image data of nonsquare icons.

Fixes #3487

---
Notes:
* Crash reproduced and fix verified with [unibrow](https://xyne.archlinux.ca/projects/unibrow/)
* This does not address the creation of visible windows for all xembed tray icons.

### DIFF
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -56,8 +56,8 @@ static int read_pixmap(sd_bus_message *msg, struct swaybar_sni *sni,
 			goto error;
 		}
 
-		int size;
-		ret = sd_bus_message_read(msg, "ii", NULL, &size);
+		int width, height;
+		ret = sd_bus_message_read(msg, "ii", &width, &height);
 		if (ret < 0) {
 			sway_log(SWAY_ERROR, "%s %s: %s", sni->watcher_id, prop, strerror(-ret));
 			goto error;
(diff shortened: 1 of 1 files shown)


swaybar/tray/item.c
```diff
@@ -419,18 +427,16 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
...
 		}
 	}
 
 	int icon_size;
 	cairo_surface_t *icon;
 	if (sni->icon) {
-		int actual_size = cairo_image_surface_get_height(sni->icon);
-		icon_size = actual_size < ideal_size ?
-			actual_size*(ideal_size/actual_size) : ideal_size;
+		icon_size = ideal_size;
```
ianyfan: -1 on this change because it causes icons to look ugly when scaling only a little, say, from 20 to 22.

progandy: Oh, right. Integer maths. I'll change it.

progandy:
> I also find it a bit silly that you can send non-square icons too.

I haven't encountered it, but user input must not cause crashes. I could simply ignore non-square icons as well.

progandy:
@ianyfan This new version is the minimal validation I consider necessary to prevent a malicious application from crashing swaybar.

ddevault:
Thanks!

PR #3502, opened 2019-01-23
Fix tray icon updates

lukaslihotzki:
Apparently, IconThemePath needs to be reread on NewIcon signal.
This fixes tray icon updates in electron apps with libappindicator-gtk3.


ddevault:
Thanks!

PR #3513, opened 2019-01-24
Don't hardcode urxvt (similar to i3)

jbeich:
Similar to https://github.com/i3/i3/commit/6420b2b102bab214ad0bf1fa0f5e190f02984e82. Sway (unlike i3) probably wants to try Wayland-native terminals first.


ddevault:
Thank you, but this is by design.

PR #3516, opened 2019-01-26
Implement pointer-constraints-unstable-v1 and relative-pointer-unstable-v1

Laaas:
It's fully functional and has been tested with the example wlroots clients, although there is one thing I'd like to fix, which I'd like input from you on how to fix it:
In `sway/input/cursor.c` I duplicate some code from wlroots, because the code from wlroots isn't always run before this function runs, since the focus will be moved to the new surface before all surface commit listeners have been executed. If you remove the duplicated code, pointer constraints will still work fine, but the cursor won't always be able to warp into the region, which can be quite annoying for the user.
The duplicated code (`sway/input/cursor.c:1433`):
```c
	// FIXME: Big hack, stolen from wlr_pointer_constraints_v1.c:121.
	// This is necessary because the focus may be set before the surface
	// has finished committing, which means that warping won't work properly,
	// since this code will be run *after* the focus has been set.
	// That is why we duplicate the code here.
	if (pixman_region32_not_empty(&constraint->current.region)) {
		pixman_region32_intersect(&constraint->region,
			&constraint->surface->input_region, &constraint->current.region);
	} else {
		pixman_region32_copy(&constraint->region,
			&constraint->surface->input_region);
	}
```

In addition, should I add any config commands to control pointer constraints? I do not know how you'd like them to be designed, so I refrained from doing that. You can however even now just shift focus to the parent (or anything else really) to break the constraint.

### DIFF
--- a/sway/commands/pointer_constraint.c
+++ b/sway/commands/pointer_constraint.c
@@ -1,0 +1,51 @@
+#include <string.h>
+#include <wlr/types/wlr_pointer_constraints_v1.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/input/cursor.h"
+#include "sway/input/seat.h"
+
+enum operation {
+	OP_ENABLE,
+	OP_DISABLE,
+	OP_ESCAPE,
+};
+
+// pointer_constraint [enable|disable|escape]
+struct cmd_results *cmd_pointer_constraint(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "pointer_constraint", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+
+	enum operation op;
+	if (strcmp(argv[0], "enable") == 0) {
+		op = OP_ENABLE;
+	} else if (strcmp(argv[0], "disable") == 0) {
+		op = OP_DISABLE;
+	} else if (strcmp(argv[0], "escape") == 0) {
+		op = OP_ESCAPE;
+	} else {
+		return cmd_results_new(CMD_FAILURE, "Expected enable|disable|escape");
+	}
+
+	if (op == OP_ESCAPE && config->reading) {
+		return cmd_results_new(CMD_FAILURE, "Can only escape at runtime.");
+	}
+
+	struct sway_cursor *cursor = config->handler_context.seat->cursor;
+	struct seat_config *seat_config = seat_get_config(cursor->seat);
+	switch (op) {
+	case OP_ENABLE:
+		seat_config->allow_constrain = true;
+		break;
+	case OP_DISABLE:
+		seat_config->allow_constrain = false;
+		/* fallthrough */
+	case OP_ESCAPE:
+		sway_cursor_constrain(cursor, NULL);
+		break;
+	}
+
+	return cmd_results_new(CMD_SUCCESS, NULL);
+}
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -1454,6 +1454,11 @@ void handle_pointer_constraint(struct wl_listener *listener, void *data) {
 
 void sway_cursor_constrain(struct sway_cursor *cursor,
 		struct wlr_pointer_constraint_v1 *constraint) {
+	struct seat_config *config = seat_get_config(cursor->seat);
+	if (!config->allow_constrain) {
+		return;
+	}
+
 	if (cursor->active_constraint == constraint) {
 		return;
 	}
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -403,6 +403,14 @@ static void handle_cursor_motion(struct wl_listener *listener, void *data) {
 	double dx = event->delta_x;
 	double dy = event->delta_y;
 
+	double dx_unaccel = event->unaccel_dx;
+	double dy_unaccel = event->unaccel_dy;
+
+	wlr_relative_pointer_manager_v1_send_relative_motion(
+		server.relative_pointer_manager,
+		cursor->seat->wlr_seat, event->time_msec, dx, dy,
+		dx_unaccel, dy_unaccel);
+
 	struct wlr_surface *surface = NULL;
 	double sx, sy;
 	struct sway_node *node = node_at_coords(cursor->seat,
@@ -437,6 +445,12 @@ static void cursor_motion_absolute(struct sway_cursor *cursor,
 	double lx, ly;
 	wlr_cursor_absolute_to_layout_coords(cursor->cursor, dev,
 		x, y, &lx, &ly);
+
+	double dx = lx - cursor->cursor->x;
+	double dy = ly - cursor->cursor->y;
+	wlr_relative_pointer_manager_v1_send_relative_motion(
+		server.relative_pointer_manager,
+		cursor->seat->wlr_seat, (uint64_t)time_msec * 1000, dx, dy, dx, dy);
 
 	struct wlr_surface *surface = NULL;
 	double sx, sy;
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -1,4 +1,5 @@
 #define _POSIX_C_SOURCE 200809L
+#include <assert.h>
 #include <math.h>
 #include <libevdev/libevdev.h>
 #include <linux/input-event-codes.h>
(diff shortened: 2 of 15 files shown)


ddevault:
Because I'm a lazy piece of shit, can you provide a test plan for me? I'm not sure what clients I should expect this to work with (or their behavior)

ddevault:
Aside from the wlroots example of course.

Laaas:
Well it should generally be useful in games, although it won't be super useful without relative pointer motion. I'm actually not aware of any non-example clients that use pointer constraints without relative pointer motion.

Laaas:
I'll try finding an example though

emersion:
Probably virtualization software, like GNOME Boxes?

ddevault:
If you don't have time it's no big deal, but if you're able to finish this by Friday it'll ship with sway 1.0 - if not, sway 1.1.

absrd:
Played successfully a FPS games with this PR and relative pointer: https://github.com/absrd/sway/tree/relative-pointer

One possible issue: In a dual monitor setup in a game with a cursor, i.e. not FPS, if I put the cursor in game at the edge between monitors and click, it starts rendering sway's cursor. Though, constraints seems to work fine, i.e. the cursor doesn't leave the game or the monitor, which the cursor is in.

Laaas:
@absrd will you make a PR?

emersion:
@absrd: please make a PR without pointer-constraints. You can test it with the relative-pointer example in wlroots.

klardotsh:
I am absolutely thrilled to see these branches happening (these are two of the last three blockers from me using `sway` as my daily driver on my desktop - I already use it on laptop). I can gladly test a few games and perhaps VirtualBox or something this evening if this needs any further testing/reassurance (most of the games I play are XWayland-based as far as I know, especially a few of the WIne-/Proton-based titles, but I'm not sure if that matters)

Any particular testing requests @ddevault / @Laaas ?

Laaas:
@emersion that example actually segfaults without pointer constraints implemented... blame me lol.

@ddevault It's honestly pretty finished and I just named it WIP because I wanted input, because I wasn't happy with the end result completely (some bad code duplication as mentioned at the top of this PR), so I thought maybe you could think of a nice way to avoid that, with your greater experience with the codebase.

I'll rename this as not WIP, since it works AFAIK.

Laaas:
@klardotsh I think it would be helpful just trying to use it as a daily driver. I will switch my machine to @absrd's branch and try playing some of my games, maybe some Wayland native games too if I can find any.

ddevault:
I'll carve out some time tomorrow for testing and reviewing this.

ddevault:
>In addition, should I add any config commands to control pointer constraints? I do not know how you'd like them to be designed, so I refrained from doing that. You can however even now just shift focus to the parent (or anything else really) to break the constraint.

￼￼Would probably be good to have these two commands:

- `pointer_constraint enabled|disabled` (config-time)
- `pointer_constraint escape` (runtime)

This is okay to add separately if necessary, but if you have the time please add it to this patch.

sway/input/cursor.c
```diff
@@ -403,6 +403,14 @@ static void handle_cursor_motion(struct wl_listener *listener, void *data) {
 	double dx = event->delta_x;
 	double dy = event->delta_y;
 
+	double dx_unaccel = event->unaccel_dx;
+	double dy_unaccel = event->unaccel_dy;
+
+        wlr_relative_pointer_manager_v1_send_relative_motion(
```
emersion: Style: bad indent here

ddevault:
Pushed the pointer_constraint command. @emersion take a look and merge this if you like it?

sway/input/cursor.c
```diff
@@ -1284,3 +1392,111 @@ const char *get_mouse_button_name(uint32_t button) {
...
+
+	struct sway_pointer_constraint *sway_constraint =
+		calloc(1, sizeof(struct sway_pointer_constraint));
+	sway_constraint->constraint = constraint;
+
+	sway_constraint->destroy.notify = handle_constraint_destroy;
+	wl_signal_add(&constraint->events.destroy, &sway_constraint->destroy);
+
+	struct sway_node *focus = seat_get_focus(seat);
+	if (focus->type == N_CONTAINER && focus->sway_container->view) {
```
emersion: `focus` can be NULL

emersion: Also, `focus` can be a workspace, so we should check its type too.

sway/commands/pointer_constraint.c
```diff
@@ -0,0 +1,54 @@
...
+	case OP_ENABLE:
+		seat_config->allow_constrain = true;
+		break;
+	case OP_DISABLE:
+		seat_config->allow_constrain = false;
+		/* fallthrough */
+	case OP_ESCAPE:
+		if (cursor->active_constraint) {
+			wlr_pointer_constraint_v1_send_deactivated(
+					cursor->active_constraint);
```
emersion: `cursor->active_constraint` needs to be unset too. It's probably better to call `sway_cursor_constrain(NULL)`.

sway/input/cursor.c
```diff
@@ -1284,3 +1392,111 @@ const char *get_mouse_button_name(uint32_t button) {
...
+			wl_list_remove(&cursor->constraint_commit.link);
+		}
+		wl_list_init(&cursor->constraint_commit.link);
+		cursor->active_constraint = NULL;
+	}
+
+	free(sway_constraint);
+}
+
+void handle_pointer_constraint(struct wl_listener *listener, void *data) {
```
emersion: This can be `static`, and be set in `sway_cursor_constrain` near `wl_signal_add`. Makes code easier to follow.

sway/input/cursor.c
```diff
@@ -401,9 +399,76 @@ static void handle_cursor_motion(struct wl_listener *listener, void *data) {
...
+
+	double dx = event->delta_x;
+	double dy = event->delta_y;
+
+	double dx_unaccel = event->unaccel_dx;
+	double dy_unaccel = event->unaccel_dy;
+
+	wlr_relative_pointer_manager_v1_send_relative_motion(
+		server.relative_pointer_manager,
+		cursor->seat->wlr_seat, (uint64_t)event->time_msec * 1000, dx, dy,
```
emersion: Don't multiply by 1000, `wlr_relative_pointer_manager_v1_send_relative_motion` expects milliseconds too.

emersion:
Thanks!

lephe:
It's awesome to see this key feature implemented. I'll be happy to use it and report any bugs. Thanks a lot!

klardotsh:
Well, there's other bugs at play for sure for my setup (going to open a separate ticket for that), but this functionality seems to work exactly as advertised! Was able to play Guild Wars 2 (via Wine+XWayland) and the cursor correctly locked to the window, and didn't get "caught" at the edge of the screen when spinning the character in endless circles.

Looks like a win to me!

Kunagi7:
Yesterday, I tried this implementation after it was merged to master.
Since I have an Intel HD 4000 I can't play fancy games to test this but:
I was able to play Euro Truck Simulator 2 (from Steam) without any camera movement glitches.
Also tried Minecraft and it worked almost perfect (camera OK, the player's inventory shows the mouse cursor as expected but chest inventories don't show the mouse cursor).
With this I can fully switch to sway. 
Great work!
_P.S: I had an issue with scdoc but I fixed it by updating to 1.8.0._

Laaas:
@Kunagi7 what about when you just open your inventory? Seems like it doesn't unhide the cursor for some reason, i.e. a buggy client.

Kunagi7:
@Laaas When I open my inventory (E key) the cursor unhide works as expected.
But if I open a chest the cursor stays hidden (but I can still move stuff inside the chest, blindly).
If I have time I'll try to play Minecraft again and try to recreate the issue.

emersion:
>I think disabling palm rejection when the mouse/pointer is captured/grabbed could be a good way to implement it. If I have a game open, then palm rejection is disabled so I can use my trackpad for camera movement. Then when I navigate to a Discord/IRC/other window, palm rejection is re-enabled for typing. Then when switching back to the game, palm rejection is disabled again and I get my trackpad back for camera movement. A keybinding could be a secondary/alternative method of disabling/enabling palm rejection.

There's a libinput setting for this. Should be easy to implement a script with the IPC to do exactly as you say.

PR #3523, opened 2019-01-28
Center surface inside container when it's too small

RyanDwyer:
The goal here is to center fullscreen views when they are both too small for the output and refuse to resize to the output's dimensions. It has the side effect of also centering the view when it's too small for its container.

Example clients that have this behaviour are emersion's hello-wayland and weston (running inside sway).

It works by introducing `surface_{x,y,width,height}` properties to the container struct. The x and y represent layout-local coordinates where the surface will be rendered. The width and height are only used to track the surface's previous dimensions so we can detect when the client has resized it and recenter and apply damage accordingly.

The new surface properties are calculated when a transaction is applied, as well as when a view resizes itself unexpectedly. The latter is done in `view_update_size`. This function was previously restricted to views which are floating, but can now be called for any views.

For views which refuse to resize *smaller* than a particular size, such as gnome-calculator, the surface is still anchored to the top left as per the current behaviour.

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -291,10 +291,9 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		wlr_xdg_surface_get_geometry(xdg_surface, &new_geo);
 		struct sway_container *con = view->container;
 
-		if ((new_geo.width != con->content_width ||
-					new_geo.height != con->content_height) &&
-				container_is_floating(con)) {
-			// A floating view has unexpectedly sent a new size
+		if ((new_geo.width != con->surface_width ||
+					new_geo.height != con->surface_height)) {
+			// The view has unexpectedly sent a new size
 			desktop_damage_view(view);
 			view_update_size(view, new_geo.width, new_geo.height);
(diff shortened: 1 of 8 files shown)


sway/desktop/xdg_shell.c
```diff
@@ -291,9 +291,8 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		wlr_xdg_surface_get_geometry(xdg_surface, &new_geo);
 		struct sway_container *con = view->container;
 
-		if ((new_geo.width != con->content_width ||
-					new_geo.height != con->content_height) &&
-				container_is_floating(con)) {
+		if ((new_geo.width != con->surface_width ||
+					new_geo.height != con->surface_height)) {
 			// A floating view has unexpectedly sent a new size
```
RedSoxFan: I don't think this is specific to floating views anymore so the comment should be updated (repeat for `xdg_shell_v6` and `xwayland`)

RyanDwyer: Fixed.

RedSoxFan:
Thanks!

ianyfan:
This makes tiling mouse resize wavy, not sure what the right fix is.

PR #3525, opened 2019-01-28
Add note about required scdoc version to README.

c-edw:
scdoc is outdated on Fedora 29 (it was ~1.3.0 or something like that) and probably is on many other distros, so I wasn't able to generate manpages until I built the latest version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Install dependencies:
 * pango
 * cairo
 * gdk-pixbuf2 \*\*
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Thanks!

c-edw:
addendum: I guess I would've ideally added that to the translations too, but there's already discrepancies which make the READMEs unreliable to follow. As long as it's noted somewhere, I'm happy.

PR #3526, opened 2019-01-28
Introduce container_is_scratchpad_hidden

RyanDwyer:
Just a convenience function that improves readability of the code.

Other things worth noting:

* `container_get_siblings` and `container_sibling_index` no longer use the `const` keyword.
* `container_handle_fullscreen_reparent` is only ever called after attaching the container to a workspace, so its `con->workspace` check has been changed to an assertion.


emersion:
Thanks!

Issue #3529, opened 2019-01-28
Output scale affects cursor movement between outputs


jpgrayson:
In my two-output configuration, using scale factors other than 1.00 can cause cursor movement between outputs to be disabled, i.e. cursor movement will stop at the edge of the current output instead of moving to the adjacent output.

My output configuration (with scale factors of 1.00):

```
Output eDP-1 'Sharp Corporation 0x144A 0x00000000'
  Current mode: 3200x1800 @ 59.981998 Hz
  Position: 0,0
  Scale factor: 1.000000
  Transform: normal
  Workspace: 2
  Available modes:
    3200x1800 @ 47.985001 Hz
    3200x1800 @ 59.981998 Hz

Output DP-1 'Goldstar Company Ltd LG HDR 4K 0x00004F99' (focused)
  Current mode: 3840x2160 @ 59.997002 Hz
  Position: 3200,0
  Scale factor: 1.000000
  Transform: normal
  Workspace: 1
  Available modes:
    640x480 @ 59.939999 Hz
    ...
    3840x2160 @ 59.997002 Hz
```
Changing the scale of `eDP-1` (the left output) to any non-1.0 value will trigger the bug (e.g. `swaymsg 'output eDP-1 scale 2'`).

Using a non-1.0 scale on `DP-1` (the right output) *does not* trigger the bug--the cursor still freely moves between outputs.

If the relative postions of `eDP-1` and `DP-1` are swapped such that `eDP-1` is on the right and `DP-1` is on the left (`swaymsg 'output eDP-1 position 3200 0; output DP-1 position 0 0;'`), then using a non-1.0 scale on `DP-1` (now the left output) triggers the bug. I.e. it seems as if the scale of the *left* output determines whether this bug manifests.

If the outputs' relative positions are modified such that they have an above/below relationship instead of left-of/right-of, using a non-1.0 scale on *either* output causes the bug to manifest.

Some other notes:
* Even if the outputs' scales match (e.g. 2.0 and 2.0), the cursor movement is restricted. I.e. it seems to only matter whether the scale is != 1.0.
* sway version 1.0-beta.2-283-g7c27d73b (Jan 28 2019, branch 'master')
* wlroots-git 0.2.r178.g018727b1-1

emersion:
If the left output has scale 2 and is 3200px wide, you need to position the other one at 3200 / 2 = 1600. Patches welcome to improve the docs if it's not clear enough.

jpgrayson:
[...]

PR #3532, opened 2019-01-28
Document cursor moving between adjacent outputs

jpgrayson:
Add a sentence to sway-output.5.scd to highlight that the cursor can
only be moved between immediately adjacent outputs.

References issue #3529


RedSoxFan:
Thanks!

PR #3535, opened 2019-01-29
Cleanup config reading failure error logs

RedSoxFan:
Related https://www.reddit.com/r/swaywm/comments/akqofn/sway_wont_start_at_all_on_intel_graphics/

This cleans up the log when sway fails to read a config file. The
following changes have been made:
- A missing error message has been added to the log when the config file
is a directory instead of a regular file
- In main, `goto` statements have been added after the `sway_terminate`
calls instead of wrapping every block in `if (!terminate_request)`
- Unnecessary NULL-checks around calls to free in `main` have been
removed
- Deferred command execution has been extracted to a separate function
and the `Running deferred commands` log message will not be shown when
there are no deferred commands.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -516,6 +516,11 @@ bool read_config(FILE *file, struct sway_config *config,
 		struct swaynag_instance *swaynag);
 
 /**
+ * Run the commands that were deferred when reading the config file.
+ */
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks!

PR #3541, opened 2019-01-30
Add details on `--whole-window` for bindsym

norcalli:
Updates sway.5 to include information on the usage of the `--whole-window` option in the context of the `bindsym` command, which modifies mouse bindings to allow them to operate over the whole window instead of just the titlebar. Also includes the disclaimer about mouse bindings only working over the title bar.

Also fixes the escaping of the `BTN_LEFT` and `BTN_RIGHT` key mention.

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -282,7 +282,7 @@ runtime.
 
 		for_window <criteria> move container to output <output>
 
-*bindsym* [--release|--locked] [--input-device=<device>] [--no-warn] <key combo> <command>
+*bindsym* [--whole-window] [--border] [--exclude-titlebar] [--release] [--locked] [--input-device=<device>] [--no-warn] <key combo> <command>
 	Binds _key combo_ to execute the sway command _command_ when pressed. You
 	may use XKB key names here (*xev*(1) is a good tool for discovering these).
 	With the flag _--release_, the command is executed when the key combo is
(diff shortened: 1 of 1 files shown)


norcalli:
I tried to add emphasis to the warning, but it didn't render how I wanted, so is there a better way to highlight that sentence, or should I just leave it un-emphasized?
image

sway/sway.5.scd
```diff
@@ -294,11 +294,17 @@ runtime.
 	the _--no-warn_ flag.
 
 	Mouse buttons can either be specified in the form _button[1-9]_ or by using
-	the name of the event code (ex _BTN_LEFT_ or _BTN_RIGHT_). For the former
+	the name of the event code (ex _BTN\_LEFT_ or _BTN\_RIGHT_). For the former
```
RedSoxFan: @ddevault With scdoc 1.8, is it intentional that the underscores have to be escaped here to underline the whole word? I'm not sure due to the other underscores.

ddevault: Yes

ddevault: 1.8.1 rather

norcalli:
Rendered 
image

RedSoxFan:
Oh actually, one small thing. Can you please add the flags to the `bindcode` syntax line?

norcalli:
@RedSoxFan donezo

RedSoxFan:
Just noticed that your last two commits are not tied to your github account. Can you please squash the three commits into one (with an email attached to your account) and force push?

norcalli:
@RedSoxFan done. Sorry, I forgot I auto configure my git based on hostname.

emersion:
Thanks!

PR #1636, opened 2018-03-27
swaybar: correctly render min_width for strings

jrouleau:
fixes #1635

PR #3542, opened 2019-01-30
Rebase #1636

v44r:
Rebase #1636 against current master.


swaybar/i3bar.c
```diff
@@ -78,7 +78,7 @@ static void i3bar_parse_json(struct status_line *status,
 				block->min_width = json_object_get_int(min_width);
 			} else if (type == json_type_string) {
 				/* the width will be calculated when rendering */
-				block->min_width = 0;
+				block->min_width_str = strdup(json_object_get_string(min_width));
```
RedSoxFan: This needs to be freed in `i3bar_block_unref`

swaybar/i3bar.c
```diff
@@ -21,6 +21,9 @@ void i3bar_block_unref(struct i3bar_block *block) {
 		free(block->full_text);
 		free(block->short_text);
 		free(block->align);
+		if (block->min_width_str) {
```
RedSoxFan: `free(NULL)` is a no-op so the condition here is actually unnecessary and can be removed

v44r:
D'oh

emersion:
Can you squash these commits together?

emersion:
Thanks!

PR #3551, opened 2019-02-01
Revert double frees.

c-edw:
Please see https://github.com/swaywm/sway/pull/3121#issuecomment-459697467.

I am sincerely sorry for this. I should've had a better understanding of what I was doing and not relied on you guys to catch my errors.


ianyfan:
It should be more obvious when resources should not be freed.

ddevault:
Thanks!

PR #3554, opened 2019-02-02
cmd_seat: allow - to be used as alias for current

RedSoxFan:
This allows for `-` (hyphen) to be used as an alias for the current seat
while sway is running. This alias was chosen since it is unlikely to
interfere with any desirable seat identifier

### DIFF
--- a/sway/commands/seat.c
+++ b/sway/commands/seat.c
@@ -2,6 +2,7 @@
 #include <strings.h>
 #include "sway/commands.h"
 #include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
 #include "log.h"
 #include "stringop.h"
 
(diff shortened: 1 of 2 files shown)


RedSoxFan:
> Maybe `-` would be better? (I don't know)

That sounds fine to me. I'll update this in a couple hours

RedSoxFan:
Changed from `.` to `-`

ianyfan:
Ha, this feels like bikeshedding, but I don't really like `-` as the alias. I think I would prefer `@`, but in the end, I don't really think it matters.

emersion:
>Ha, this feels like bikeshedding, but I don't really like - as the alias. I think I would prefer @, but in the end, I don't really think it matters.

Many programs take `-` as a special argument, for instance for reading for stdin instead of a file. That's why I suggested that.

ddevault:
I don't like `-` either but w<sup>ho <sup>car<sup>es</sup></sup></sup>

PR #3562, opened 2019-02-03
Focus workspace inactive node with focus_follows_mouse

### LINKED ISSUE #3552: focus-follows-mouse sometimes focusses on container
* Sway Version: aee9942145119444e5592846ac7e1a0d89e8a3e0

* Debug Log: https://pub.yori.cc/sway-focus-wrong.log
The relevant part is near the end. At 17:05:30, the focus was wrong and the container was focussed.

You should try to reproduce the issue with the default configuration. If you cannot, please reproduce with a minimal configuration, upload the config to a pastebin service, and link to it below.

* Configuration File: https://pub.yori.cc/i3_config

Finally, explain the steps you took in plain English to reproduce the problem below.

Steps to reproduce: make a V container on the left screen, a single window on the right screen, and stack two terminals on the left screen. Then, enter the left screen with the mouse. When hovering over the border, the entire container will light up. Usually, the focus ends up correct, but sometimes, the container will remain focussed, and output will not go to a window.

vilhalmer:
Fixes #3552. I'm pretty sure that passing through the potential NULL here is the right thing to do, since seat_set_focus handles it, but if not I can add a check for that.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -362,7 +362,7 @@ void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
 			struct sway_output *focused_output = node_get_output(focus);
 			struct sway_output *output = node_get_output(node);
 			if (output != focused_output) {
-				seat_set_focus(seat, node);
+				seat_set_focus(seat, seat_get_focus_inactive(seat, node));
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #3564, opened 2019-02-03
seat_cmd_cursor: do not create non-existing seat

RedSoxFan:
If a seat does not exist in seat_cmd_cursor, do not create it. A seat
without any attachments is useless since it will have no capabilities.

This changes `input_manager_get_seat` to have an additional argument
that dictates whether or not to create the seat if it does not exist.


ddevault:
Thanks!

PR #3567, opened 2019-02-04
[WIP] Rebase cursor on layer rearrange

### LINKED ISSUE #3080: Rebase the cursor when re-ordering layer surfaces
**I first reported this to [slurp](https://github.com/emersion/slurp/issues/15), but this is apparently a sway bug according to @emersion, so the steps below will be from that issue.**

When you run slurp, the cursor will remain regular until you move it, and then, it will turn into the crosshair.

This causes a problem when running slurp and directly doing a selection, the selection will not be highlighted and the resulting selection will have an empty x and y value, like this: 0,0 1144x753

Steps to replicate:
1. Open a terminal and run slurp
2. Don't move the cursor
3. Notice the cursor still being regular, not a crosshair
4. Press the primary mouse button and drag a selection
5. Notice the selection not being highlighted and the result having x/y values set to 0

Sway version: sway version 1.0-beta.1-151-ge9c77970 (Nov  5 2018, branch 'master')

c-edw:
Closes #3080.

I'm not sure if this is the correct way of doing it.

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -9,6 +9,7 @@
 #include "sway/desktop/transaction.h"
 #include "sway/input/input-manager.h"
 #include "sway/input/seat.h"
+#include "sway/input/cursor.h"
 #include "sway/layers.h"
 #include "sway/output.h"
 #include "sway/server.h"
@@ -211,6 +212,7 @@ void arrange_layers(struct sway_output *output) {
 
 	struct sway_seat *seat;
 	wl_list_for_each(seat, &server.input->seats, link) {
+		cursor_rebase(seat->cursor);
 		seat_set_focus_layer(seat, topmost ? topmost->layer_surface : NULL);
 	}
 }


emersion:
Maybe we should just commit a transaction in this case?

@RyanDwyer Thoughts?

RyanDwyer:
While this works, it can be wasteful because there's probably other code somewhere that arranges layers and calls `cursor_rebase` itself. This commit would cause it to be rebased twice in this case.

I'd prefer it to be rebased from `handle_layer_shell_surface`.

emersion:
What if a layer surface is resized? We need to rebase in this case too.

I think we want to detect when we need to re-arrange layers, instead of doing it at every single commit.

c-edw:
Could you distil what I need to do? I can move the `cursor_rebase` code into `handle_layer_shell_surface`, but if you're asking me to `detect when we need to re-arrange layers`, I don't know how that would be done.

emersion:
>detect when we need to re-arrange layers

Right now we re-arrange layer surfaces at each commit, even if they still keep the same size. This is unnecessary.

For this PR we could probably just rebase the cursor in `handle_layer_shell_surface`, which is necessary anyway. Not re-arranging when a layer surface doesn't resize is a follow-up issue.

c-edw:
I tried moving it into `handle_layer_shell_surface`, but that doesn't seem to fix the case I'm testing https://github.com/swaywm/swaylock/issues/32#issue-402670691.

emersion:
Superseded by https://github.com/swaywm/sway/pull/3665

PR #3570, opened 2019-02-05
execute_command: dont strip quotes for exec_always

RedSoxFan:
This removes quote stripping for `exec_always` in `execute_command`.
Since `exec_always` commands will be deferred in the config and
processed by `execute_command`, the quotes need to be left intact
like they are for `exec`.


emersion:
Thanks!

PR #3573, opened 2019-02-05
IPC_COMMAND: split on newline

### LINKED ISSUE #3566: Commands issued via swaymsg aren't split at newlines
Run `swaymsg "$(echo 'workspace 1\n workspace 2')"`

Result on i3: end up on workspace `2`

Result on sway: end up on workspace `1 workspace 2`

RedSoxFan:
Related to #3566 

This splits commands given in IPC_COMMAND on newline to match i3's behavior.

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -595,6 +595,16 @@ void ipc_client_handle_command(struct ipc_client *client) {
 	switch (client->current_command) {
 	case IPC_COMMAND:
 	{
+		char *line = strtok(buf, "\n");
+		while (line) {
+			size_t line_length = strlen(line);
+			if (line + line_length >= buf + client->payload_length) {
+				break;
+			}
+			line[line_length] = ';';
+			line = strtok(NULL, "\n");
+		}
+
 		list_t *res_list = execute_command(buf, NULL, NULL);
 		transaction_commit_dirty();
 		char *json = cmd_results_to_json(res_list);


ianyfan:
I can't seem to get backslash continuation working in i3 (still works in config). Can you?

RedSoxFan:
> I can't seem to get backslash continuation working in i3 (still works in config). Can you?

Thought I did, but nope. Removed support since it simplifies the code a bit

sway/ipc-server.c
```diff
@@ -595,6 +595,16 @@ void ipc_client_handle_command(struct ipc_client *client) {
 	switch (client->current_command) {
 	case IPC_COMMAND:
 	{
+		char *line = strtok(buf, "\n");
+		while (line) {
+			size_t line_length = strlen(line);
+			if (line + line_length >= buf + client->payload_length) {
```
ianyfan: Is this check needed?

RedSoxFan: Yes, otherwise it will replace the final null terminator with `;`. It could be changed from `>=` to `==` though.

ianyfan: I thought the loop would break if it reached the end (i.e. strtok returns null).

RedSoxFan: It will, but the following line `line[line_length] = ';'` should not be executed for the final token

ianyfan: You are correct, sorry, I misremembered how strtok behaved on the last token.

ddevault:
Thanks!

PR #3574, opened 2019-02-05
load_include_configs: fix wordexp fail condition

### LINKED ISSUE #3565: Segfaults via include command.
I am fuzzing various config commands and found some strings that can cause segfaults:

```
include '() { 0; }; touch /tmp/blns.shellshock1.fail;''
include '() { _; } >_[$($())] { touch /tmp/blns.shellshock2.fail; }'
include '<<< %s(un='%s') = %u'
```

Backtrace is [here](https://gist.github.com/c-edw/4d19ff0c20a0c2007fae546b9ceb60d9).

RedSoxFan:
Fixes #3565 

This fixes the failure condition for the wordexp call in
load_include_configs. The only success value is zero. Since the error
codes are positive, having the check be less than zero was causing
segfaults on failure when accessing the words.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -557,7 +557,7 @@ bool load_include_configs(const char *path, struct sway_config *config,
 
 	wordexp_t p;
 
-	if (wordexp(path, &p, 0) < 0) {
+	if (wordexp(path, &p, 0) != 0) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

Issue #3575, opened 2019-02-05
Using bar goes to "$workspace" rather than $workspace


nperry:
Steps:
1. Use default config
2. Open a window on workspace 1. 
3. Move to workspace two
4. Using mouse click on the workspace 1

Actual:
Goes to "1"

Expected
Goes to 1

[neil@desktop ~]$ swaymsg -t get_version
sway version 1.0-rc1-11-g1e65cc1e (Feb  5 2019, branch 'master')

v44r:
Reverting 1e65cc1e fixes it.

PR #3576, opened 2019-02-05
Fix quote stripping

emersion:
Let's not use `!strcmp(…)` anymore.

Fixes https://github.com/swaywm/sway/issues/3575


RedSoxFan:
Thanks! (Sorry for breaking it)

emersion:
Yeah, `!strcmp(a, b)` is confusing because if checks if `a` and `b` match. OTOH, `strcmp(a, b) != 0` is less so because it checks if `a` and `b` mismatch.

RedSoxFan:
Oh derp, I had it inverted

PR #3579, opened 2019-02-05
cmd_workspace_gaps: fix double free on bad amount

### LINKED ISSUE #3577: Abort due to double free via workspace command.
Using this in a config:

```
workspace "x" gaps "x" "x"
```

causes Sway to output:

```
Aborted (core dumped)
```

However, this works:

```
workspace "x" gaps "x" "1"
```

The user should be told to enter a numeric value for the fourth parameter.

Backtrace is [here](https://gist.github.com/c-edw/064c16db56c3d9072c4e990ebcb5004b).

RedSoxFan:
Fixes #3577 

This fixes a double free in cmd_workspace_gaps when the amount given is
invalid. The end pointer from strtol is part of the argument and should
not be freed. Freeing the end pointer could result in a double free or
bad free depending on whether or not the end pointer was at the start of
the argument

### DIFF
--- a/sway/commands/workspace.c
+++ b/sway/commands/workspace.c
@@ -76,7 +76,6 @@ static struct cmd_results *cmd_workspace_gaps(int argc, char **argv,
 	char *end;
 	int amount = strtol(argv[gaps_location + 2], &end, 10);
 	if (strlen(end)) {
-		free(end);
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #3581, opened 2019-02-05
If validating the config, do it as early as possible.

c-edw:
I may be missing something, but I can't see why we would initialize the server and IPC when we only want to validate the config file. It was causing a memory leak (not that it matters since Sway exits almost immediately) due to the server not being destroyed.

This change has the caveat that various debug information in wlroots code won't appear.

I also think that verbose logging when validating a config file is unnecessary, but I won't submit that change unless others agree.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -312,6 +312,18 @@ int main(int argc, char **argv) {
 		wlr_log_init(WLR_ERROR, NULL);
 	}
 
+	log_kernel();
+	log_distro();
+	log_env();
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

c-edw:
My bad, I only just noticed an issue with this. It will cause a segfault due to input manager being uninitialized. It looks like we can still skip IPC initialization though. I will send a PR in a few moments.

PR #3582, opened 2019-02-05
Initialize server so input manager is available.

c-edw:
Fixes https://github.com/swaywm/sway/pull/3581#issuecomment-460681872.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -318,12 +318,6 @@ int main(int argc, char **argv) {
 	detect_proprietary(allow_unsupported_gpu);
 	detect_raspi();
 
-	if (validate) {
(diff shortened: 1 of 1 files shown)


c-edw:
The original reason I wanted to do this is to solve detected memory leaks in the validation code path. But we can't skip server initialization, and freeing it leaks memory regardless. I think it would require quite a lot of digging to solve.

ddevault:
Thanks! But are memory leaks in the validation path really that important? We exit right away.

c-edw:
As you say, it's not really important. It just causes ugly traces to spit out with a build enabled with LeakSanitizer.

PR #3585, opened 2019-02-05
Fix close_popups for xdg-shell

emersion:
wlr_xdg_popup_destroy will destroy popups, so we need to walk the tree
carefully. It's enough to just destroy all direct children, since destroying
the parent will also destroy all children.


RedSoxFan:
Thanks!

PR #3587, opened 2019-02-05
output_cmd_background: fix no file + valid mode

### LINKED ISSUE #3580: Heap buffer overflow via output command.
In a config file:

```
output * background fill irrelevant
```

However, it does not occur with:

```
output * background irrelevant irrelevant
```

This only occurs with the flags suggested to me by emersion:
```
-Db_sanitize=address,undefined
```

Backtrace is [here](https://gist.github.com/c-edw/366c99adc9c76cfa1dbbbad6b9117db2).

RedSoxFan:
Fixes #3580 

If output_cmd_background is given a valid mode as the first argument,
then there is no file given and an error should be returned.

join_args should not be called with an argc of zero since it sets the
last character to the null terminator. With an argc of zero, the length
is zero causing a heap buffer overflow when setting the byte before the
start of argv to '\0'. This probably will not ever generate a segfault,
but may cause data corruption to whatever is directly before it in
memory. To make other such cases easier to detect, this also adds a
sway_assert in join_args when argc is zero. 

_If desired, I can split the second part into a separate commit or PR_.

### DIFF
--- a/common/stringop.c
+++ b/common/stringop.c
@@ -258,6 +258,9 @@ int unescape_string(char *string) {
 }
 
 char *join_args(char **argv, int argc) {
+	if (!sway_assert(argc > 0, "argc should be positive")) {
+		return NULL;
+	}
 	int len = 0, i;
 	for (i = 0; i < argc; ++i) {
 		len += strlen(argv[i]) + 1;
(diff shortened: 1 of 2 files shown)


c-edw:
I'm stunned at how fast you're fixing these issues. Really nice job.

ddevault:
Thanks!

PR #3592, opened 2019-02-06
load_main_config: add realpath to config_chain

### LINKED ISSUE #3589: Primary config can be interpreted twice.
In `config1`:

```
include config2
exec "echo Included config 1."
```

In `config2`:

```
include config1
```

Run `sway -c config1`. The output will be:

```
Included config 1.
Included config 1.
```

Note that you can only include the *primary* config twice. I haven't found a way to cause any issues with this yet, but I imagine it's possible.

RedSoxFan:
Fixes #3589 

Since `load_include_config` compares against the realpath of a config
file when checking if a config has already been added, the main config's
realpath has to be added to the config_chain.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -379,6 +379,14 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 		path = get_config_path();
 	}
 
+	char *real_path = realpath(path, NULL);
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3595, opened 2019-02-06
Return false if config could not be loaded.

### LINKED ISSUE #3590: Error is not thrown when including non-existent (sub) configuration file.
It looks like an error is supposed to be thrown if an included configuration file doesn't exist, but this doesn't happen.

```c
if (!load_include_configs(argv[0], config,
			&config->swaynag_config_errors)) {
	return cmd_results_new(CMD_INVALID,
			"Failed to include sub configuration file: %s", argv[0]);
}
```

I think it's because `load_include_configs` is returning true when it shouldn't. I can get an error to throw with something like `include sway\nconfig`, but this is outside of normal usage.

I'll have a look at this tomorrow morning, unless somebody beats me to it.

c-edw:
Fixes #3590.

This makes two assumptions:
- A wildcard path pointing to an empty directory or non-existent directory is OK
- An error should be thrown if the included path does not exist and it is not a wildcard path

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -574,7 +574,13 @@ bool load_include_configs(const char *path, struct sway_config *config,
 	char **w = p.we_wordv;
 	size_t i;
 	for (i = 0; i < p.we_wordc; ++i) {
-		load_include_config(w[i], parent_dir, config, swaynag);
+		bool found = load_include_config(w[i], parent_dir, config, swaynag);
+		if (!found) {
+			wordfree(&p);
+			free(parent_path);
+			free(wd);
+			return false;
+		}
 	}
 	free(parent_path);
 	wordfree(&p);
--- a/sway/config.c
+++ b/sway/config.c
@@ -558,17 +558,13 @@ bool load_include_configs(const char *path, struct sway_config *config,
 	const char *parent_dir = dirname(parent_path);
 
 	if (chdir(parent_dir) < 0) {
-		free(parent_path);
-		free(wd);
-		return false;
+		goto error_chdir;
 	}
 
 	wordexp_t p;
 
 	if (wordexp(path, &p, 0) != 0) {
-		free(parent_path);
-		free(wd);
-		return false;
+		goto error_wordexp;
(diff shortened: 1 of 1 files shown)


sway/config.c
```diff
@@ -587,7 +587,10 @@ bool load_include_configs(const char *path, struct sway_config *config,
 	}
 
 	free(wd);
-	return true;
+
+	char wildcard = path[strlen(path) - 1] == '*';
+	// If not a wildcard path, real path must point to an existing file.
+	return wildcard || file_exists(path);
```
emersion: -1, pattern could be e.g. `` ./`hostname`.conf `` or `./*.conf`.

Instead, we should check for any `load_include_config` error on line 577.

c-edw: Yep, that's a much better solution. I think I tried that originally but for some reason it didn't resolve the issue. It does now.

sway/config.c
```diff
@@ -574,7 +574,13 @@ bool load_include_configs(const char *path, struct sway_config *config,
 	char **w = p.we_wordv;
 	size_t i;
 	for (i = 0; i < p.we_wordc; ++i) {
-		load_include_config(w[i], parent_dir, config, swaynag);
+		bool found = load_include_config(w[i], parent_dir, config, swaynag);
+		if (!found) {
+			wordfree(&p);
+			free(parent_path);
+			free(wd);
+			return false;
```
emersion: I think we also need to restore the CWD here. An approach like this would probably work:

```c
bool success = true;

for (…) {
	success = load_include_config(…);
	if (!success) {
		goto out;
	}
}

out:
free(…);
chdir(…); // TODO: error handling

return success;
```

c-edw: What about the early `return false` on line 571? The CWD doesn't appear to be restored there.

emersion: Ah, right. Could you fix this one as well?

emersion:
Thanks!

ddevault:
This was by design.

PR #3600, opened 2019-02-06
Clarify error for options and positional args

RedSoxFan:
When both options and positional arguments are given, sway would print
the error `Don't use options with the IPC client`. Over the past
several months, it seems like a large number users are including this error
message in issues instead of a debug log due to not understanding that
the error message means there is an issue with their command.

This makes the error message more verbose and will hopefully make it so
more users understand that the message is not a bug in sway, but with
the command used.


emersion:
Thanks!

PR #3603, opened 2019-02-07
bar_cmd_modifier: add support for none

### LINKED ISSUE #3601: Sway-bar doesn't recoqnize 'modifier none' -option
Man page for sway-bar documents 'modifier none' -option but using it results in following error:

`[sway/config.c:756] Error on line 213 'modifier none': Unknown modifier 'none' (.config/sway/config)`

I'm using Sway and Wlroots from master.

`sway version 1.0-rc1-22-g09c2a46b (Feb  5 2019, branch 'master')`

[Swaylog](https://pastebin.com/sAJQXWNy)

RedSoxFan:
Fixes #3601 

sway-bar(5) documents `modifier none`, which comes from i3. This
implements the functionality for `modifier none` since it was not
previously implemented. The bar modifier toggles visibility of the bar
when the bar mode is set to hide. When the bar modifier is set to
`none`, the ability to toggle visibility of the bar will be disabled.

### DIFF
--- a/sway/commands/bar/modifier.c
+++ b/sway/commands/bar/modifier.c
@@ -15,19 +15,26 @@ struct cmd_results *bar_cmd_modifier(int argc, char **argv) {
 	}
 
 	uint32_t mod = 0;
-	list_t *split = split_string(argv[0], "+");
-	for (int i = 0; i < split->length; ++i) {
-		uint32_t tmp_mod;
-		if ((tmp_mod = get_modifier_mask_by_name(split->items[i])) > 0) {
-			mod |= tmp_mod;
-		} else {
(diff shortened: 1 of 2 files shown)


emersion:
Code LGTM, but I'm not sure we should merge this new feature before 1.0. @ddevault Thoughts?

RedSoxFan:
> Code LGTM, but I'm not sure we should merge this new feature before 1.0.

I viewed it as a bug since it should have been implemented for i3 compatibility. If this doesn't get merged, `none` has to get removed from sway-bar(5) since it was documented.

emersion:
Thanks!

PR #3607, opened 2019-02-07
load_main_config: use given path, store realpath

### LINKED ISSUE #3605: Regression with including config files
My config setup: `~/.config/sway/` has `config` and `local`, both symlinked from different folders. `config` contains settings shared among my machines (so it's symlinked from `/elsewhere/common/.config/sway/`), and `local` contains machine-local settings (so it's symlinked from `/elsewhere/machine-1/.config/sway/` on one machine, and so on). `config` contains a line `include local`. Since c65057ef8b03ce11d4c002ed8d1615926486b7a4, when processing the include statement in `config`, sway searches `/elsewhere/common/.config/sway/` (as opposed to `~/.config/sway/`) where it doesn't find the `local` config because it's obviously not there.

RedSoxFan:
Fixes #3605 

Since `load_include_config` compares against the realpath of a config
file when checking if a config has already been added, the main config's
realpath has to be added to the config_chain.

However, includes from the main config should be processed relative to
the path given to allow for symbolic links. This stores the realpath in
`config->config_chain`, but uses the given path for all other
operations.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -385,7 +385,6 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 		free(path);
 		return false;
 	}
-	free(path);
 
 	struct sway_config *old_config = config;
 	config = calloc(1, sizeof(struct sway_config));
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3614, opened 2019-02-07
config.in: don't include non-existent directory

attente:


ddevault:
This is by design. Is it causing errors for you?

attente:
Yes, I got a big red error message from `swaynag` saying that there was an error in the configuration file.

Does it make sense to have the include there if sway doesn't actually distribute or install that directory?

ddevault:
We should make wildcards fail more gracefully instead.

attente:
I might be a bit overly pedantic at this point, but who is placing configuration files in `@sysconfdir@/sway/config.d` if not sway itself? If it's meant to be distro maintainers/packagers, then maybe it should be them adding that include instead.

I also don't necessarily think making wildcards fail more gracefully would be a good thing since that error is actually helpful for people customizing their own configs in case they made a mistake. It's just this one case with the include in the current default config that is problematic for a new user installing sway from source.

PR #3619, opened 2019-02-08
Revert "Return false if config could not be loaded."

ddevault:
Reverts swaywm/sway#3595

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -573,13 +573,7 @@ bool load_include_configs(const char *path, struct sway_config *config,
 	char **w = p.we_wordv;
 	size_t i;
 	for (i = 0; i < p.we_wordc; ++i) {
-		bool found = load_include_config(w[i], parent_dir, config, swaynag);
-		if (!found) {
-			wordfree(&p);
-			free(parent_path);
-			free(wd);
-			return false;
-		}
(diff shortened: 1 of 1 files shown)


c-edw:
If it's by design, you should add a comment or something. That PR also fixed an unnoticed issue where the CWD wouldn't be restored in some cases.

emersion:
Can you send a PR doing these things?

c-edw:
I will.

@ddevault 
I would like to say that I'm unhappy with how this was reverted. It doesn't make you look good when you merge your own PR to revert other people's work without even a simple 'sorry', especially when you didn't intervene in my original PR (though I understand you may have been busy and weren't aware of it). If you have made *specific* decisions about how things like this function you must document them so people like me don't come along and see them as bugs.

c-edw:
I just saw the PR to remove `include @sysconfdir@/sway/config.d/*` because it was causing errors - the actual intention with my changes was to log an error if an absolute path was included but didn't exist. Wildcard paths were supposed to be unaffected and I could've sworn it worked when I tested it.

emersion:
I believe `@sysconfdir@/sway/config.d/*` expansion will fail if `@sysconfdir@/sway/config.d` doesn't exist.

PR #3623, opened 2019-02-08
Add comment about load_include_configs.

c-edw:
Related #3595, #3619.

I made `load_include_configs` void and added a comment that emphasises we don't care if config files fail to load (unless something abnormal occurs).

Also re-reverts improved error handling that fixes a case where the working directory would not be restored.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -549,43 +549,34 @@ static bool load_include_config(const char *path, const char *parent_dir,
 	return true;
 }
 
-bool load_include_configs(const char *path, struct sway_config *config,
+void load_include_configs(const char *path, struct sway_config *config,
 		struct swaynag_instance *swaynag) {
 	char *wd = getcwd(NULL, 0);
 	char *parent_path = strdup(config->current_config_path);
 	const char *parent_dir = dirname(parent_path);
 
 	if (chdir(parent_dir) < 0) {
-		free(parent_path);
-		free(wd);
-		return false;
+		sway_log(SWAY_ERROR, "failed to change working directory");
+		goto cleanup;
 	}
 
 	wordexp_t p;
-
-	if (wordexp(path, &p, 0) != 0) {
-		free(parent_path);
-		free(wd);
-		return false;
+	if (wordexp(path, &p, 0) == 0) {
+		char **w = p.we_wordv;
+		size_t i;
+		for (i = 0; i < p.we_wordc; ++i) {
+			load_include_config(w[i], parent_dir, config, swaynag);
+		}
+		wordfree(&p);
 	}
 
-	char **w = p.we_wordv;
-	size_t i;
-	for (i = 0; i < p.we_wordc; ++i) {
-		load_include_config(w[i], parent_dir, config, swaynag);
-	}
-	free(parent_path);
-	wordfree(&p);
-
-	// restore wd
+	// Attempt to restore working directory before returning.
 	if (chdir(wd) < 0) {
-		free(wd);
-		sway_log(SWAY_ERROR, "failed to restore working directory");
-		return false;
+		sway_log(SWAY_ERROR, "failed to change working directory");
 	}
-
+cleanup:
+	free(parent_path);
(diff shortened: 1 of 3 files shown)


emersion:
I think it's possible to make this code look a lot nicer, by merging the error paths with the normal path.

c-edw:
Do you mean that I am overusing labels? If so, how does this look:

```c
if (wordexp(path, &p, 0) != 0) {
	if (chdir(wd) == 0) {
		sway_log(SWAY_ERROR, "failed to change working directory");
	}
	free(parent_path);
	free(wd);
	return;
}
```
...
```c
error_chdir_end:
error_chdir_start:
	sway_log(SWAY_ERROR, "failed to change working directory");
	free(parent_path);
	free(wd);
```

emersion:
No, I mean you could remove the `return` line 584 and rename `error_*` labels to `out_*`.

c-edw:
Sorry for the delay on this. Would you mind giving it another look?

sway/config.c
```diff
@@ -549,43 +549,38 @@ static bool load_include_config(const char *path, const char *parent_dir,
...
-		return false;
-	}
-
+out_wordexp:
+	// Attempt to restore working directory before returning.
+	if (chdir(wd) == 0) {
+		goto free_all;
+	}
+out_chdir:
+	sway_log(SWAY_ERROR, "failed to change working directory");
```
RedSoxFan: Move this up and remove the `out_chdir` label

sway/config.c
```diff
@@ -549,43 +549,38 @@ static bool load_include_config(const char *path, const char *parent_dir,
...
-	// restore wd
-	if (chdir(wd) < 0) {
-		free(wd);
-		sway_log(SWAY_ERROR, "failed to restore working directory");
-		return false;
-	}
-
+out_wordexp:
+	// Attempt to restore working directory before returning.
+	if (chdir(wd) == 0) {
```
RedSoxFan: Invert this and have the sway_log error inside

sway/config.c
```diff
@@ -549,43 +549,38 @@ static bool load_include_config(const char *path, const char *parent_dir,
...
 	if (chdir(parent_dir) < 0) {
-		free(parent_path);
-		free(wd);
-		return false;
+		goto out_chdir;
 	}
 
 	wordexp_t p;
 
 	if (wordexp(path, &p, 0) != 0) {
```
RedSoxFan: Invert this, have through wordfree in here and remove the `out_wordexp` label

c-edw: Are you sure? Don't think we can call `wordfree` until later.

sway/config.c
```diff
@@ -549,43 +549,37 @@ static bool load_include_config(const char *path, const char *parent_dir,
...
+out_wordexp:
+	// Attempt to restore working directory before returning.
 	if (chdir(wd) < 0) {
-		free(wd);
-		sway_log(SWAY_ERROR, "failed to restore working directory");
-		return false;
+		sway_log(SWAY_ERROR, "failed to change working directory");
 	}
-
+free_all:
```
RedSoxFan: nitpick: can you change this to `cleanup` for consistency with the rest of the code base
(sorry missed this one before)

sway/config.c
RedSoxFan: Move everything that uses the wordexp in there (it's only 6 lines)

RedSoxFan: The blank line before the if can also be removed

c-edw:
All good points. I just made a few idiotic mistakes so had to force push a lot. I'll stick to contributing during the daytime rather than at 5am.

RedSoxFan:
Can you just remove the blank line before the wordexp if, squash the commits, and force push? Aside from those, looks good to me.

RedSoxFan:
Can you cleanup that commit message? I should have said to fixup the commits instead of squash them.

RedSoxFan:
Thanks!

PR #3627, opened 2019-02-09
container_at_stacked: skip titles when zero pixels

### LINKED ISSUE #3618: Segfault when switching to stacking layout
* Sway Version: 1.0_beta2 installed from sorrow overlay on Gentoo (fully updated)
* Debug Log: https://paste.pound-python.org/show/2HMR6KsVBiUChEU0LSrK/
* Configuration File: https://paste.pound-python.org/show/G8KwGDUXP37Ez1QV4kZu/

All I do is press **the key to enable stacking layout** (tried various mods and keys) and sway segfaults.

Running `$ swaymsg layout stacking` works fine, but using a hotkey fails.

Might be relevant, might not: I am using the dvorak keyboard layout set through the sway.desktop file.

Relevant section from dmesg:
`sway[19270]: segfault at 55cd6d375550 ip 000055d16c87d21f sp 00007ffc132872a0 error 4 in sway[55d16c84e000+3b000]`

`Code: d0 00 00 00 c5 fb 10 54 24 08 c5 f1 57 c9 c5 f3 2a ca c5 eb 5c c0 c5 fb 5e c1 c5 fb 2c d0 3b 50 04 7d 18 48 8b 40 08 48 63 d2 <48> 8b 04 d0 e9 28 fe ff ff 0f 1f 84 00 00 00 00 00 48 89 de 48 89`

RedSoxFan:
Fixes #3618

It is possible to make the title bars have a zero pixel height while
stacked, by using a blank font and no padding. This causes a division by
zero when attempting to calculate the child index in
container_at_stacked, which then results in a segfault when attempting
to access the child at that bad index (INT_MIN). This just skips the
check to see if the cursor is over a title bar of a child of a stacked
container when the title bar height is zero since there will be no title
bars.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -251,10 +251,12 @@ static struct sway_container *container_at_stacked(struct sway_node *parent,
 
 	// Title bars
 	int title_height = container_titlebar_height();
-	int child_index = (ly - box.y) / title_height;
-	if (child_index < children->length) {
-		struct sway_container *child = children->items[child_index];
-		return child;
+	if (title_height > 0) {
+		int child_index = (ly - box.y) / title_height;
+		if (child_index < children->length) {
+			struct sway_container *child = children->items[child_index];
+			return child;
+		}
 	}
 
 	// Surfaces


emersion:
Thanks!

Issue #3109, opened 2018-11-11
/etc/profile is never sourced when launching sway from GDM


asppsa:
Hi,

I found that when launching Sway from GDM on Archlinux, the `/etc/profile` file never gets sourced.  I'm not sure if this counts as a bug, as the behaviour is the same for other wayland sessions that I tried, like Weston (but not Gnome, for some reason), and its only a problem when launching from a graphical login manager, as otherwise the profile is sourced already (e.g. when logging into a TTY).  My sway version:

~~~
sway version 1.0-beta.1-157-g994c35e3 (Nov  6 2018, branch 'master')
~~~

To verify, you could do something like edit /etc/profile and add `export MY_VARIABLE=1` to that file. Then login via GDM with Sway, launch a terminal and check to see if `MY_VARIABLE` is set.  In my tests, it is not.

I have been able to solve this by changing the `Exec=sway` line in `sway.desktop` to `Exec=bash -l -c sway`. This launches a login shell (which sources the profile) and then runs sway inside of the shell.  This is enough for my purposes, but I'm not sure if this is the correct way to fix this (or even if this warrants fixing) ... at any rate I figured I would report it.

asppsa:
Hi, thanks for getting back so fast. 

The reason I wanted this sourced is that on Arch Linux, a bunch of environment vars get set by files in `/etc/profile.d` (mostly not too important, but there is the occasional thing of significance in there, like setting up Java paths). However, if you feel this is not Sway's job to fix, I understand.

emersion:
>The reason I wanted this sourced is that on Arch Linux, a bunch of environment vars get set by files in /etc/profile.d (mostly not too important, but there is the occasional thing of significance in there, like setting up Java paths). However, if you feel this is not Sway's job to fix, I understand.

This shouldn't matter, because when you run `exec` in the sway config, a shell is used so the profile is sourced.

emersion:
>To verify, you could do something like edit /etc/profile and add export MY_VARIABLE=1 to that file. Then login via GDM with Sway, launch a terminal and check to see if MY_VARIABLE is set. In my tests, it is not.

Wait, this should work. What is your `/bin/sh`?

asppsa:
[...]

PR #3634, opened 2019-02-10
Add swayrun command to exec sway using a login shell

electrickite:
When running Sway from a display manager, a user's login shell is not used. Because of this, profile scripts are not run and the environment may not be configured. As suggested in #3109, this can be solved by a wrapper that runs sway via the user's configured shell. This is similar to the approach taken by GNOME to source the standard set of profile scripts.

This could be accomplished with a shell script, but a dedicated binary avoids dependencies on specific shells.

The sway.desktop wayland session file is also updated to call the swayrun wrapper.


emersion:
I'd prefer this to be a POSIX shell script and to be outside of `$PATH` (e.g. in `/usr/libexec`).

electrickite:
@emersion A shell script is actually where I started this PR and I would prefer that solution as well. However, AFAIK, there is no POSIX-compliant way to invoke a login shell from a shell script.

`sh` (And ex: `bash` run as `sh`) can only be invoked in login mode if ARGV[0] is prefixed with a “-“. The `-l` flag is non standard, so we can’t guarantee that the user’s preferred shell supports it. But, the `exec` command has no standard way to set ARGV[0]. Bash provides a non-standard set of `exec` options, `-l` and `-a`, that could be used  to do this.  `zsh` uses an environment variable to set ARGV[0].  I’m sure other shells have their own ways.

It is comparatively easy to do this using POSIX C exec functions.

So, if we use a shell script, we would need to decide on which non-POSIX shell to use to invoke the login shell. GNOME, for example, chose `bash`. This may be a relatively safe choice since it is included in the LSB and, at least on Arch, is an indirect dependency of Sway.

Thoughts? Would you prefer a hard dependency on a non-POSIX shell, or the C route? My thought was that the C wrapper was so small (<100 lines) that it did not add too much complexity to Sway.

emersion:
Eh, I haven't considered that. I think you're right, a short C executable is probably better.

Is `allowed_shell` really necessary though?

rkanati:
I used to use LightDM and had this exact problem, so I ended up writing a custom `Xsession` script that actually used a proper login shell. Since switching to Sway, I've been using SDDM, which supports more shells natively, and I've not needed to do this.

My point is, while this might be a useful stopgap in the meantime, I feel it ought to be the responsibility of display managers and shells to play nice together.

swayrun/main.c
```diff
@@ -0,0 +1,71 @@
...
+		exec_argv[2] = "exec " SWAY_COMMAND " \"$@\"";
+		exec_argv[3] = shell;
+
+		execvp(shell, exec_argv);
+		fprintf(stderr, "Could not run %s using login shell: %s\n", SWAY_COMMAND, shell);
+	} else {
+		argv[0] = SWAY_COMMAND;
+		execvp(SWAY_COMMAND, argv);
+		fprintf(stderr, "Could not run %s\n", SWAY_COMMAND);
+	}
```
ddevault: Why do we care if they want to run it with some shell which isn't in /etc/shells?

electrickite: I modeled the approach from the `gnome-session` wrapper. It checks `/etc/shells` and also does not accept nologin or false. I suppose it is probably not necessary since by the time this gets called, the DM has already decided to allow or deny a login. 

Would you prefer the check be removed?

ddevault: I don't particularly care either way, just seems unnecessary. I'm not -1 or +1 on this

emersion: I'd prefer to drop it, it simplifies the code.

emersion:
>I feel it ought to be the responsibility of display managers and shells to play nice together.

Yeah, they're the ones responsible for login, after all.

electrickite:
> Yeah, they're the ones responsible for login, after all.

@emersion It doesn’t seem like there’s consensus on where (or if?) profile scripts should be sourced in the Wayland world. Things seemed clearer when using startx or xinitrc.

swayrun/main.c
electrickite: @emersion I agree it would simplify things. Can you see any security concerns with using an “unapproved” login shell? `chsh` and friends confirm a shell is in `/etc/shells` before running.

emersion: Yeah, but we assume the environment from which we're started is safe. If it's not the case, much more nasty things could be done, e.g. via `LD_PRELOAD` or `PATH`.

electrickite: Sounds reasonable. The `allowed_shell` check has been removed.

electrickite:
@emersion Do you still feel the `swayrun` executable should live outside of `$PATH`? If so, I would suggest `/usr/lib/sway/swayrun`. Both Debian and Arch discourage the use of `/usr/libexec` and Fedora seems to allow `/usr/lib/<pkg>`.

emersion:
Yes, libdir is fine by me. This indeed seems to be what most packages do.

electrickite:
@emersion Latest commit moves swayrun to libdir.
@ddevault I believe this is ready for review.

swayrun/main.c
```diff
@@ -0,0 +1,39 @@
...
+
+#define SWAY_COMMAND "sway"
+
+int main(int argc, char **argv) {
+	char *shell = getenv("SHELL");
+
+	if (shell) {
+		// 3 exec arguments + argc + argv[argc] NULL pointer
+		int exec_argc = 4 + argc;
+		char **exec_argv = malloc(exec_argc * sizeof(char*));
```
emersion: Style: `char *`

swayrun/main.c
```diff
@@ -0,0 +1,39 @@
...
+#include <unistd.h>
+
+#define SWAY_COMMAND "sway"
+
+int main(int argc, char **argv) {
+	char *shell = getenv("SHELL");
+
+	if (shell) {
+		// 3 exec arguments + argc + argv[argc] NULL pointer
+		int exec_argc = 4 + argc;
```
emersion: This should be `size_t`

swayrun/main.c
```diff
@@ -0,0 +1,39 @@
...
+		char **exec_argv = malloc(exec_argc * sizeof(char*));
+
+		// Prefix - to shell path to indicate login shell
+		char *login_shell = malloc(strlen(shell) + 2);
+		strcpy(login_shell, "-");
+		strcat(login_shell, shell);
+
+		// Build the argumrnts to exec
```
emersion: Typo: arguments

swayrun/main.c
```diff
@@ -0,0 +1,39 @@
...
+
+		execvp(shell, exec_argv);
+		fprintf(stderr, "Could not run %s using login shell: %s\n", SWAY_COMMAND, shell);
+	} else {
+		argv[0] = SWAY_COMMAND;
+		execvp(SWAY_COMMAND, argv);
+		fprintf(stderr, "Could not run %s\n", SWAY_COMMAND);
+	}
+
+	return errno;
```
emersion: Not a fan of returning `errno` here. It spec'ed as an `int` so it could be set to the error value 256, which can wrap-around when converted to an exit code. Instead, we can return `EXIT_FAILURE` and change the `fprintf` calls to also print `strerror(errno)`.

swayrun/main.c
electrickite: Thanks for the correction.

swayrun/main.c
electrickite: Sorry about that. Corrected.

swayrun/main.c
electrickite: Changed from `int` to `size_t`

swayrun/main.c
electrickite: Good point on `errno` as a return value. The only function that can set the return value should be `execvp`, but good to be cautious. It also improves the error messages.

ddevault:
I have to admit that this patch makes me uneasy. Like @emersion said, it's the login manager's responsibility to login. Since login managers are unsupported, the normal way of logging in is by... logging in, e.g. with agetty, which will run a login shell. I imagine that the login managers are wrong if they aren't launching their compositors/X sessions like so:

    execvp("/bin/sh", { "-sh", "/usr/bin/sway" })

Sorry for stonewalling this, but I want to understand better before we start adding binaries to sway willy-nilly to fix someone else's bug.

emersion:
Can you ask GNOME why they fixed it in gnome-shell rather than gdm?

electrickite:
@emersion Not sure this provides the full answers, but [here is the commit](https://gitlab.gnome.org/GNOME/gnome-session/commit/7e307f8ddb91db5d4051c4c792519a660ba67f35) to `gnome-session` that (re-)added this feature and the [discussion about it](https://bugzilla.gnome.org/show_bug.cgi?id=736660) on bugzilla.

@ddevault I understand: we don't want to fix the problem in the wrong place, but I could see this being argued either way, really. A login manager or (DM acting as a login manager) is responsible for authenticating the user and starting their chosen shell. The shell then sets up its environment, sources script, etc, etc.

You could argue that a Wayland compositor when run directly from the login manager is taking the place of a traditional tty shell, and so it is the responsibility of the compositor to configure its own environment sensibly.

I could also see the opposite viewpoint, that the compositor is simply an executable run at the end of a login process, and its environment should be set up properly prior to execution.

My rational for the PR was that, while login managers are not necessarily supported, Sway at least acknowledges the use-case by providing an out-of-the-box Wayland session `.desktop` file. It would be nice if Sway, when invoked via a DM, tried to "do the right thing" and mimic a traditional login environment.

ddevault:
Yeah, I'm going to pass on this patch. I think you should make your case with the login manager instead.

PR #3638, opened 2019-02-10
fix double free for mode toggle if bar was invisible

### LINKED ISSUE #3637: Setting 'bar mode toggle' after 'bar mode invisible' will crash Sway
Sway will crash when toggling bar mode two times after setting bar mode to invisible.

Steps to reproduce:

`swaymsg bar mode invisible; swaymsg bar mode toggle; swaymsg bar mode toggle`

Sway version:

`1.0-rc1-38-g995b6eda (Feb 10 2019, branch 'master')`

[Coredump](https://pastebin.com/bJwqnjER)

[Debug log](https://pastebin.com/kbPB246W)

Emantor:
If the bar was set to "invisible" and subsequently "toggle" was send twice, the
new mode was never set and the bar->mode was double freed.
Fix this by not requiring the bar->mode to be "hide" and instead show it
unconditionally, because it was either hidden or invisible.

Fixes #3637

### DIFF
--- a/sway/commands/bar/mode.c
+++ b/sway/commands/bar/mode.c
@@ -11,7 +11,7 @@ static struct cmd_results *bar_set_mode(struct bar_config *bar, const char *mode
 	if (strcasecmp("toggle", mode) == 0 && !config->reading) {
 		if (strcasecmp("dock", bar->mode) == 0) {
 			bar->mode = strdup("hide");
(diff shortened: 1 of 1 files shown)


Emantor:
The signed-off-by is a habit, I'll remove it from the commit…

emersion:
Thanks!

PR #3641, opened 2019-02-10
input/keyboard: respect solo repeat_{rate,delay}

### LINKED ISSUE #3640: repeat_delay ignored on its own
Version: 1.0 RC1

Setting either repeat_rate or repeat_delay on its own causes the property to be ignored. I want to change repeat_delay but keep the default for repeat_rate.

https://github.com/swaywm/sway/blob/f5190d1f797f5a9d0596ab6d31240a0e179ac457/sway/input/keyboard.c#L552-L558

RedSoxFan:
Fixes #3640 

If `repeat_rate` or `repeat_delay` is set without the other being set,
the default was being used for both. This changes the logic to respect
the value given and use the default for the other when only one is set.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -549,13 +549,17 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
 		wlr_keyboard_led_update(wlr_device->keyboard, leds);
 	}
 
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

Issue #3163, opened 2018-11-20
Swaybar subpixel font rendering


mrlotfi:
I am using sway version 1.0-beta.1-211-gbf7af9c6 (Nov 20 2018, branch 'master') on archlinux. I felt that the fonts are a bit blurry on the swaybar compared to i3bar. So I took a screenshot and found that the swaybar is not using subpixel rendering. On the other hand window titles are rendered properly. Is this the expected behaviour or is this a bug?

![image](https://user-images.githubusercontent.com/9977153/48803774-2761dd80-ed28-11e8-8754-b5706b758968.png)

![image](https://user-images.githubusercontent.com/9977153/48803994-b838b900-ed28-11e8-946e-6882c20f566a.png)

emersion:
I thought we had subpixel rendering in swaybar. Are you using fractional scaling?

vperilla:
I am using factional scaling

![barra](https://user-images.githubusercontent.com/211490/48807504-b56ba300-eceb-11e8-896d-2526bea49f50.png)

mrlotfi:
> I thought we had subpixel rendering in swaybar. Are you using fractional scaling?

No. I have only overridden the default font using font pango: xxxxxx option.

emersion:
We do use cairo subpixel font rendering. Not sure what's wrong here.

ggreer:
Perhaps it's a bug detecting your screen's subpixel output. In common/cairo.c, `to_cairo_subpixel_order()` defaults to `CAIRO_SUBPIXEL_ORDER_DEFAULT` if the subpixel order is `WL_OUTPUT_SUBPIXEL_UNKNOWN`. That renders text with antialiasing and no hinting. I have this bug on one of my laptops, but it uses an eDP->LVDS converter.

I guess some debug logging in that function could be useful.

lephe:
Interestingly `weston-info` fails in my case:

```
interface: 'wl_output', version: 3, name: 24
	x: 0, y: 0, scale: 1,
	physical_width: 344 mm, physical_height: 193 mm,
	make: 'Unknown', model: '0x38ED',
	subpixel_orientation: unknown, output_transform: normal,
	mode:
		width: 1920 px, height: 1080 px, refresh: 40.064 Hz,
		flags:
	mode:
		width: 1920 px, height: 1080 px, refresh: 60.096 Hz,
		flags: current
```
[...]

PR #3645, opened 2019-02-11
Add support for manually setting subpixel hinting on outputs.

ggreer:
Many laptop screens report unknown subpixel order. Allow users to manually set subpixel hinting to work around this.

Addresses https://github.com/swaywm/sway/issues/3163

~~TODO: update docs.~~

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -4,6 +4,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <strings.h>
+#include <wayland-server-protocol.h>
 #include "log.h"
 #include "util.h"
 
@@ -53,4 +54,24 @@ float parse_float(const char *value) {
 		return NAN;
 	}
 	return flt;
+}
+
+
+const char *sway_wl_output_subpixel_to_string(enum wl_output_subpixel subpixel) {
+	switch (subpixel) {
+	case WL_OUTPUT_SUBPIXEL_UNKNOWN:
+		return "unknown";
+	case WL_OUTPUT_SUBPIXEL_NONE:
+		return "none";
+	case WL_OUTPUT_SUBPIXEL_HORIZONTAL_RGB:
+		return "rgb";
+	case WL_OUTPUT_SUBPIXEL_HORIZONTAL_BGR:
+		return "bgr";
+	case WL_OUTPUT_SUBPIXEL_VERTICAL_RGB:
+		return "vrgb";
+	case WL_OUTPUT_SUBPIXEL_VERTICAL_BGR:
+		return "vbgr";
+	}
+	sway_assert(false, "Unknown value for wl_output_subpixel.");
+	return NULL;
 }
--- a/sway/commands/output/subpixel.c
+++ b/sway/commands/output/subpixel.c
@@ -1,0 +1,36 @@
+#include <string.h>
+#include "log.h"
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/output.h"
+
+struct cmd_results *output_cmd_subpixel(int argc, char **argv) {
+	if (!config->handler_context.output_config) {
+		return cmd_results_new(CMD_FAILURE, "Missing output config");
+	}
+	if (!argc) {
+		return cmd_results_new(CMD_INVALID, "Missing subpixel argument.");
+	}
+	enum wl_output_subpixel subpixel;
+
+	if (strcmp(*argv, "rgb") == 0) {
+		subpixel = WL_OUTPUT_SUBPIXEL_HORIZONTAL_RGB;
+	} else if (strcmp(*argv, "bgr") == 0) {
+		subpixel = WL_OUTPUT_SUBPIXEL_HORIZONTAL_BGR;
+	} else if (strcmp(*argv, "vrgb") == 0) {
+		subpixel = WL_OUTPUT_SUBPIXEL_VERTICAL_RGB;
+	} else if (strcmp(*argv, "vbgr") == 0) {
+		subpixel = WL_OUTPUT_SUBPIXEL_VERTICAL_BGR;
+	} else if (strcmp(*argv, "none") == 0) {
+		subpixel = WL_OUTPUT_SUBPIXEL_NONE;
+	} else {
+		return cmd_results_new(CMD_INVALID, "Invalid output subpixel.");
+	}
+
+	struct output_config *oc = config->handler_context.output_config;
+	config->handler_context.leftovers.argc = argc - 1;
+	config->handler_context.leftovers.argv = argv + 1;
+
+	oc->subpixel = subpixel;
+	return NULL;
+}
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -8,10 +8,11 @@
 #include <unistd.h>
 #include <wlr/types/wlr_output_layout.h>
 #include <wlr/types/wlr_output.h>
-#include "log.h"
 #include "sway/config.h"
 #include "sway/output.h"
 #include "sway/tree/root.h"
+#include "log.h"
+#include "util.h"
 
 int output_name_cmp(const void *item, const void *data) {
 	const struct output_config *output = item;
@@ -43,6 +44,7 @@ struct output_config *new_output_config(const char *name) {
 	oc->x = oc->y = -1;
 	oc->scale = -1;
 	oc->transform = -1;
+	oc->subpixel = WL_OUTPUT_SUBPIXEL_UNKNOWN;
 	return oc;
 }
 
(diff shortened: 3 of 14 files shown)


sway/config/output.c
```diff
@@ -225,6 +229,10 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		sway_log(SWAY_DEBUG, "Set %s scale to %f", oc->name, oc->scale);
 		wlr_output_set_scale(wlr_output, oc->scale);
 	}
+	if (oc && oc->subpixel != WL_OUTPUT_SUBPIXEL_UNKNOWN) {
+		sway_log(SWAY_DEBUG, "Set %s subpixel to %d", oc->name, oc->subpixel);
```
ggreer: If there is a toString function for wl_output_subpixel, I'd love to use it. I looked around for a few minutes but I didn't find one.

sway/commands/output/subpixel.c
```diff
@@ -0,0 +1,36 @@
+#include <string.h>
```
RedSoxFan: Style: use tabs not spaces

sway/config/output.c
RedSoxFan: I think you have to write your own, which should be trivial since it is a small enum

RedSoxFan:
Other points:
- This needs to be reset on reload (see `default_output_config`)
- This should probably be added to `IPC_GET_OUTPUTS`

RedSoxFan:
Also just as a note, since sway is currently in a feature freeze for the 1.0 tree, this will not make it into 1.0.

sway/config/output.c
ggreer: I added one to my wlroots PR, since I'm guessing both wlroots and WMs that use it will want to log subpixel order.

ggreer:
```ggreer@iron:~/code/sway% swaymsg -t get_outputs
Output eDP-1 'Unknown 0x06D0 0x00000000' (focused)
  Current mode: 2880x1920 @ 60.007000 Hz
  Position: 0,0
  Scale factor: 2.000000
  Subpixel order: rgb
  Transform: normal
  Workspace: 3
  Available modes:
    2880x1920 @ 60.007000 Hz
```

I can also change the subpixel hinting at runtime with `swaymsg output eDP-1 subpixel rgb`. Some apps don't pick up the new hinting until they're restarted, but app title bars are updated immediately.

ggreer:
Alright. I added the code to store the original subpixel value & reset it in `apply_output_config()`. In testing, I can run `swaymsg output eDP-1 subpixel rgb` and `swaymsg -t get_outputs` says hinting is rgb. Then I can reload my config and `swaymsg -t get_outputs` says unknown (which is the original value).

sway/config/output.c
```diff
@@ -306,6 +311,14 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		sway_log(SWAY_DEBUG, "Set %s scale to %f", oc->name, oc->scale);
 		wlr_output_set_scale(wlr_output, oc->scale);
 	}
+	enum wl_output_subpixel subpixel = output->original_subpixel;
+	if (oc && oc->subpixel && oc->subpixel != WL_OUTPUT_SUBPIXEL_UNKNOWN) {
+		subpixel = oc->subpixel;
+		sway_log(SWAY_DEBUG, "Set %s subpixel to %s", oc->name,
```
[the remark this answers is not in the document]
ggreer: But oc might be null, right? Then `oc->name` would cause a segfault.

sway/config/output.c
```diff
@@ -359,6 +372,7 @@ static void default_output_config(struct output_config *oc,
 	}
 	oc->x = oc->y = -1;
 	oc->scale = 1;
+	oc->subpixel = WL_OUTPUT_SUBPIXEL_UNKNOWN;
```
RedSoxFan: Oh sorry, missed this. Either use:
```c
struct sway_output *output = wlr_output->data;
oc->subpixel = output->detected_subpixel;
```
or change the function signature to take a `struct sway_output` instead of `struct wlr_output` and use
```
oc->subpixel = output->detected_subpixel;
```

sway/config/output.c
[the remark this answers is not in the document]
RedSoxFan: Actually use along with the change in `default_output_config`:
```c
if (oc && (oc->subpixel != WL_OUTPUT_SUBPIXEL_UNKNOWN || config->reloading)) {
    sway_log(SWAY_DEBUG, "Set %s subpixel to %s", oc->name,
            sway_wl_output_subpixel_to_string(oc->subpixel);
    wlr_output_set_subpixel(wlr_output, oc->subpixel);
}
```

ggreer:
Alright, changes made. Before merging, I want to test this again on a lower resolution display to make sure it still behaves correctly.

sway/config/output.c
```diff
@@ -306,6 +311,13 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		sway_log(SWAY_DEBUG, "Set %s scale to %f", oc->name, oc->scale);
 		wlr_output_set_scale(wlr_output, oc->scale);
 	}
+
+	if (oc && (oc->subpixel != WL_OUTPUT_SUBPIXEL_UNKNOWN || config->reloading)) {
+		sway_log(SWAY_DEBUG, "Set %s subpixel to %s", oc->name,
+			sway_wl_output_subpixel_to_string(oc->subpixel));
+			wlr_output_set_subpixel(wlr_output, oc->subpixel);
```
RedSoxFan: Extra indent here. Otherwise, LGTM

common/util.c
```diff
@@ -54,3 +55,23 @@ float parse_float(const char *value) {
 	}
 	return flt;
 }
+
+
+const char *sway_wl_output_subpixel_to_string(enum wl_output_subpixel subpixel) {
+	switch (subpixel) {
+		case WL_OUTPUT_SUBPIXEL_UNKNOWN:
```
emersion: Style: `case` at the same indent level as `switch`

common/util.c
```diff
@@ -54,3 +55,23 @@ float parse_float(const char *value) {
...
+			return "none";
+		case WL_OUTPUT_SUBPIXEL_HORIZONTAL_RGB:
+			return "rgb";
+		case WL_OUTPUT_SUBPIXEL_HORIZONTAL_BGR:
+			return "bgr";
+		case WL_OUTPUT_SUBPIXEL_VERTICAL_RGB:
+			return "vrgb";
+		case WL_OUTPUT_SUBPIXEL_VERTICAL_BGR:
+			return "vbgr";
+		default:
```
emersion: Please don't add `default` cases, the compiler won't tell you where your code needs an update if a new enum member is added. Instead, add `assert(false)` after the `switch`.

sway/config/output.c
```diff
@@ -306,6 +311,13 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		sway_log(SWAY_DEBUG, "Set %s scale to %f", oc->name, oc->scale);
 		wlr_output_set_scale(wlr_output, oc->scale);
 	}
+
+	if (oc && (oc->subpixel != WL_OUTPUT_SUBPIXEL_UNKNOWN || config->reloading)) {
+		sway_log(SWAY_DEBUG, "Set %s subpixel to %s", oc->name,
+			sway_wl_output_subpixel_to_string(oc->subpixel));
+		wlr_output_set_subpixel(wlr_output, oc->subpixel);
```
emersion: Don't you need to damage the whole output here?

sway/ipc-server.c
```diff
@@ -644,6 +644,8 @@ void ipc_client_handle_command(struct ipc_client *client) {
 			json_object_object_add(output_json, "focused",
 				json_object_new_boolean(focused));
 
+			const char *subpixel = sway_wl_output_subpixel_to_string(output->wlr_output->subpixel);
+			json_object_object_add(output_json, "subpixel hinting", json_object_new_string(subpixel));
```
RedSoxFan: Can you  please also change this to `subpixel` or `subpixel_hinting` and document it in `sway-ipc.7.scd`? (swaymsg pretty print will also need to be updated to reflect the change)

sway/config/output.c
ggreer: Originally I put the damage in `wlr_output_set_subpixel()`. Fixed.

common/util.c
ggreer: I see one `assert(false)` in the codebase, so I used sway_assert() with an error message instead. Anyway, default case removed.

sway/ipc-server.c
ggreer: Done.

emersion:
Thanks!

PR #3647, opened 2019-02-11
fix misc memory leaks

RedSoxFan:
This fixes a few misc memory leaks reported by asan:
- Items of `config->config_chain` are now freed instead of just the list
itself
- `bar->swaybar_command` is now freed
- The result returned by a seat subcommand is now returned instead of
leaked


emersion:
Thanks!

PR #3648, opened 2019-02-11
Abort early when XDG_RUNTIME_DIR is not set

### LINKED ISSUE #3187: sway crashes display without a clear error message when XDG_RUNTIME_DIR isn't set
After installing sway 1.0-beta.1 on Gentoo, simply running `sway` (or `sway -C`) makes my display crash (I cannot switch TTYs).
By redirecting the output to a file and viewing it through SSH, I can see an error message about XDG_RUNTIME_DIR.

Would it be possible to output the error message without locking the display?

RedSoxFan:
Fixes #3187 

This aborts sway and displays an error message about XDG_RUNTIME_DIR
not being set without initializing the wl_display or logging any other
information.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -299,6 +299,14 @@ int main(int argc, char **argv) {
 		}
 	}
 
+	// Since wayland requires XDG_RUNTIME_DIR to be set, abort with just the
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3665, opened 2019-02-12
Rebase cursor when a layer surface maps

emersion:
Also removes an extraneous arrange_outputs call, it's already called if
necessary in arrange_layers.

Updates https://github.com/swaywm/sway/issues/3080


ddevault:
Thanks!

Issue #3656, opened 2019-02-11
Incorrect --version output when built from source tarball but git is installed


jbeich:
Regressed by #3578. Affects downstream builds of both releases and snapshots.

```
$ git --version
git version 2.20.1
$ curl -o sway-1.0-rc2.tar.gz https://codeload.github.com/swaywm/sway/tar.gz/1.0-rc2
$ tar xzf sway-1.0-rc2.tar.gz
$ cd sway-1.0-rc2
$ meson _build
$ ninja -C _build
$ _build/sway/sway --version
sway version  (Feb 12 2019, branch '')
```

SpeedJack:
The code inside the `if git.found()` block should be executed only if we are in a valid git repo (maybe checking if the `.git` directory exists?). Otherwise the `else` block should be executed.

martinetd:
There are ways to use git that do not rely on a .git directory (environment variable, .git file redirecting to another directory (e.g. worktrees), etc)

I would run the git commands as long as git is found, and if their output is empty fallback to the else code instead.

PR #3667, opened 2019-02-12
Fix --version when building from tarball

emersion:
Fixes https://github.com/swaywm/sway/issues/3656


ddevault:
Thanks!

progandy:
It might be useful to use ```export-subst``` to add additional information to archives exported from git.

https://git-scm.com/docs/gitattributes#_creating_an_archive

PR #3669, opened 2019-02-12
Don't remove from scratchpad on move to workspace

athrungithub:
on move container window to another workspace,
not remove from scratchpad.

I not sure, i3 for ex, no remove


RedSoxFan:
Thanks!

PR #3672, opened 2019-02-13
subsurface_get_root_coords: break on NULL

### LINKED ISSUE #3671: Sway crashes when closing QT app
* To reproduce
Run openscad and another window in tiling mode and it should crash when closing openscad

* Sway version
sway-git February 11

* Debug Log:
https://gist.github.com/bendik/5b7a8966ac647d78120cad4b716df358

* Configuration File:
https://gist.github.com/bendik/21f4ffabd1f76aa7ba6430e1f6a7dd96

* Stack Trace:
https://gist.github.com/bendik/905da9e67d90df795189e99dc86863b1

RedSoxFan:
Fixes #3671 

It is possible for `wlr_surface_is_subsurface` to return true, but
`wlr_surface_from_wlr_surface` to be NULL. This adds a NULL check to the
value returned by `wlr_surface_from_wlr_surface` and breaks out of the
while loop in `subsurface_get_root_coords`.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -708,6 +708,9 @@ static void subsurface_get_root_coords(struct sway_view_child *child,
 	while (surface && wlr_surface_is_subsurface(surface)) {
 		struct wlr_subsurface *subsurface =
 			wlr_subsurface_from_wlr_surface(surface);
+		if (subsurface == NULL) {
+			break;
+		}
 		*root_sx += subsurface->current.x;
 		*root_sy += subsurface->current.y;
 		surface = subsurface->parent;


emersion:
>It is possible for `wlr_surface_is_subsurface` to return true, but
`wlr_surface_from_wlr_surface` to be NULL.

In which case does this happen again? We should add this to the wlroots docs.

RedSoxFan:
> > It is possible for `wlr_surface_is_subsurface` to return true, but
> > `wlr_surface_from_wlr_surface` to be NULL.
> 
> In which case does this happen again? We should add this to the wlroots docs.

To be honest, I'm not sure. I modeled this PR around one of your wlroots ones (https://github.com/swaywm/wlroots/pull/1396).

emersion:
Thanks!

RedSoxFan:
> I'll update the docs.

Somehow I completely ignored all documentation in that PR that I was referencing... It is already documentated

PR #3673, opened 2019-02-13
cursor: relative-pointer-v1 time is usec

RedSoxFan:
Similar to #3543 

In `handle_cursor_motion`, the timestamp passed to
`wlr_relative_pointer_manager_v1_send_relative_motion` should be
microseconds (not milliseconds) according to relative-pointer-v1 spec.


emersion:
Thanks!

PR #3674, opened 2019-02-13
Make sway-version overrideable again

bschacht:
It would be great if the version of sway could be set at compile time without having to patch the buildfile. This would let package maintainers override the version if there have been any changes.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -128,7 +128,10 @@ endif
 
 add_project_arguments('-DSYSCONFDIR="/@0@"'.format(join_paths(prefix, sysconfdir)), language : 'c')
 
-version = '"@0@"'.format(meson.project_version())
+version = get_option('sway-version')
+if version == ''
+	version = '"@0@"'.format(meson.project_version())
+endif
 if git.found()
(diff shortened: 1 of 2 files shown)


emersion:
>This would let package maintainers override the version if there have been any changes.

What's your use-case for this? I don't understand why you would need to patch up the version.

yorickvP:
https://github.com/colemickens/nixpkgs-wayland/blob/master/pkgs/sway-beta/default.nix#L38 building from a github archive discards all other revision information

bschacht:
On 2/13/19 12:47 PM, emersion wrote:
>> This would let package maintainers override the version if there have been any changes.
> 
> What's your use-case for this? I don't understand why you would need to patch up the version.

For example if there are backports of fixes, we can indicate the patch
level in the version. Or right now i'd like to indicate that its -rc2,
but i'd have to patch meson.build.

ddevault:
I suggest patching meson.build instead.

yorickvP:
Be prepared to receive bug reports from end-users with the wrong version, then.

PR #3676, opened 2019-02-14
meson: better version string for exported archives

progandy:
Use the gitattribute [export-subst](https://git-scm.com/docs/gitattributes#_creating_an_archive) to include some commit information in meson.build during ```git archive``` operations.

This should somewhat alleviate the concerns in #3674. The version number won't be correct, but at least the version information will contain the correct commit hash. 

This does not provide a way to add a suffix to the version number in case packagers apply patches. In that case, the version should be changed with a patch as well.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+meson.build       export-subst
--- a/meson.build
+++ b/meson.build
@@ -11,6 +11,10 @@ project(
 	],
 )
 
+gitinfo_hash='$Format:%h$'
+gitinfo_tree='$Format:%t$'
+gitinfo_refname='$Format:%D$'
+
 add_project_arguments(
 	[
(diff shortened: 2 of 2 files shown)


ddevault:
Hmmmmmmmmmm. I gotta think this one over

progandy:
If you want it, then there are some changes I'd want to make like removing ```gitinfo_tree```. It serves no purpose, since it is only available when the commit hash is filled in as well.

By the way, I think we should also make sure that git doesn't return the value for a parent git repository if sway is built from a tarball, maybe something like comparing the output of 
```git rev-parse --show-toplevel```
and
```join_paths(meson.source_root(), meson.current_source_dir())```

ddevault:
I think I'm -1 to this change, it's a bit too complex to justify considering that 99% of the time we're going to be working with a release tarball, not a HEAD tarball.

PR #3689, opened 2019-02-15
apply_output_config: dpms on before modeset

### LINKED ISSUE #3688: eDP-1 dpms on fails when thunderbolt DP-1 disconnected while dpms off
Steps:
 - `swaymsg "output * dpms off"; sleep 5; swaymsg "output * dpms on"`
 - disconnect USB-C HDMI adapter (DP-1) while displays are off
 - eDP-1 doesn't come back on

Hardware: Dell XPS-13 9360(4K) + Trianium USB-C HDMI hub
swaymsg -t get_version: sway version 1.0-rc1-45-gfb4f2928 (Feb 12 2019, branch 'master')

Logs: https://termbin.com/0te1
 - control where adapter isn't disconnected (*22:51:28*)
 - experiment where is it disconnected (*22:51:46*).

Notes:
 - Output can be restored by `swaymsg output eDP-1 disable && swaymsg output eDP-1 enable`.
 - I have had issues with this hardware and outputs not working on resume under X11 with this hardware config, so I may have quirky hardware. Still, maybe wlroots can detect the failure and reinit?

RedSoxFan:
Fixes #3688 

On the DRM backend, if an output is dpms'd off and a different output is
hotplugged, the CRTC for the output is reclaimed. When modesetting an
output without a CRTC, a CRTC will not be given to an output that is not
desired to be enabled. This splits setting the dpms state in
apply_output_config. If the output should be dpms on, the it is enabled
before attempting to modeset. Otherwise, it is dpms'd off after setting
everything else.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -199,6 +199,11 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		return true;
 	}
 
+	if (oc && oc->dpms_state == DPMS_ON) {
+		sway_log(SWAY_DEBUG, "Turning on screen");
+		wlr_output_enable(wlr_output, true);
+	}
+
 	bool modeset_success;
 	if (oc && oc->width > 0 && oc->height > 0) {
 		sway_log(SWAY_DEBUG, "Set %s mode to %dx%d (%f GHz)", oc->name, oc->width,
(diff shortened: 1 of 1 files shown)


edrex:
Tested, fixes #3688 for me.

emersion:
>When modesetting an
output without a CRTC, a CRTC will not be given to an output that is not
desired to be enabled.

But enabling the output later should give the CRTC?

(Maybe we can disable the output early too, if needed? That would make it so the DPMS code is all in one place)

RedSoxFan:
> But enabling the output later should give the CRTC?

Currently on `master` (https://github.com/swaywm/sway/commit/e8c472aee95ce1cbb6feef04d970327d1717d7fc), `apply_output_config` would return early due to `modeset_success` being false so `dpms on` would never take effect. If you did `output <identifier> disable` and `output <identifier> enable`, then it would get a CRTC.
 
> (Maybe we can disable the output early too, if needed? That would make it so the DPMS code is all in one place)

That would not work since a DPMS'd off output is still usable (blindly anyway) and should have the resolution, scale, and transform set correctly.

emersion:
Thanks!

Issue #3594, opened 2019-02-06
Mouse events not being accepted by clients if cursor is at the edge of a client


atomnuker:
Just like the description says scroll or click events are not accepted(?) by clients if the cursor is at the very edge of it. They're still present in WAYLAND_DEBUG=client however so I'm not sure where the problem is.
This seems to be a recent regression from the relative/constrained cursor changes.

PR #3694, opened 2019-02-15
Check layout before getting pointer surface coords

VincentVanlaer:
This fixes issues of clients at the edge of the screen, like swaybar, ignoring buttons.

Fixes https://github.com/swaywm/sway/issues/3594

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -425,11 +425,12 @@ static void handle_cursor_motion(struct wl_listener *listener, void *data) {
 		dx, dy, dx_unaccel, dy_unaccel);
 
 	struct wlr_surface *surface = NULL;
+	struct sway_node *node = NULL;
 	double sx, sy;
-	struct sway_node *node = node_at_coords(cursor->seat,
-		cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
-
 	if (cursor->active_constraint) {
+		node = node_at_coords(cursor->seat,
+			cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
+
 		if (cursor->active_constraint->surface != surface) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


sway/input/cursor.c
```diff
@@ -445,8 +446,13 @@ static void handle_cursor_motion(struct wl_listener *listener, void *data) {
...
 
 	wlr_cursor_move(cursor->cursor, event->device, dx, dy);
+
+	// Recalculate pointer location after layout checks
+	node = node_at_coords(cursor->seat,
+		cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
+
 	cursor_send_pointer_motion(cursor, event->time_msec, node, surface,
-		sx + dx, sy + dy);
+		sx, sy);
```
emersion: This stops confining `s{x,y}`.

VincentVanlaer: Could you give an example of how this breaks confinement? My knowledge of the sway code is almost non-existent, so I might be missing something obvious.

emersion: `wlr_region_confine` will move `sx.sy` into the constrained region. We need to use `sx_confined,sy_confined` somehow. We also need to add `dx,dy` to the cursor position, but adding them to `sx,sy` in the `cursor_send_pointer_motion` seems wrong anyway.

VincentVanlaer: At lines 444-445 `dx` and `dy` are updated with `sx_confined,sy_confined`. These are then passed to `wlr_cursor_move` and we pick those new coordinates up in the arguments of `node_at_coords`.

emersion: Oh, you're right. Forgot `wlr_cursor_move` was doing all the magic.

We should probably add `dx,dy` when calling `node_at_coords` right?

VincentVanlaer: From what I can tell the purpose of that code is to get the coordinates of the pointer in the coordinates of the surface to which the constraint is associated. It does that by assuming that the previous position of the cursor is still over the constraining surface, and if not, it bails out. I'm not a fan of that really, I'd rather have it obtain the surface local coordinates in way that doesn't make these assumptions, but couldn't really find one since the constraint doesn't know where the surface is.

emersion: Okay, that's probably good enough for now.

emersion:
Thanks!

PR #3695, opened 2019-02-15
Replace broken debuglog command with log_level command.

ggreer:
The debuglog command mentioned in the manpage wasn't hooked up at all. This patch replaces it with a log_level command that directly sets the log verbosity.


ddevault:
I'd rather just remove this entirely to be honest.

ggreer:
I think it's pretty handy to be able to set the log level in the config instead of tweaking how sway is started. I ended up making this change because someone on IRC recommended running `swaymsg debuglog on` to get more logging out of a sway that was in a buggy state. It turns out that despite being in the manpage, that's not a command that works.

Ideally, I'd also like to change the command line args to use a `--log_level` instead of `-V` and `-d`, but that seems like a whole 'nother can of worms. Long term, it'd be neat if config, `swaymsg` commands, and command line arguments were all unified in their names and behavior.

ddevault:
You generally don't get useful logs *after* the bugs have started. A partial debug log isn't very useful. I'm still in favor of removing this entirely.

emersion:
Yeah, +1 for removing this.

PR #3696, opened 2019-02-16
Spanish translation README

athrungithub:


ddevault:
Needs links to this translation in the list of translations in the other README files.

ddevault:
Got some feedback from the spanish speaking community that this isn't a very good translation. I'm going to pass on this patch, but thank you.

athrungithub:
Yes, spanish is different in diverse countries. There wil never be a consensus.
Thanks  sway comunity!

ReK2Fernandez:
is not that different as long you follow the international Spanish standards, indeed there are diff slang but is all understandable. With time you can get more into details and regional pick latin american, central American and Iberian Spanish but to start Spanish generic is better than nothing..

PR #3697, opened 2019-02-16
Add workspace {prev,next}_on_output --create

ddevault:
This creates the next workspace if you hit the end.

### DIFF
--- a/sway/commands/workspace.c
+++ b/sway/commands/workspace.c
@@ -181,7 +181,9 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 			++argv;
 		}
 
-
+		bool create = argc > 1 && strcasecmp(argv[1], "--create") == 0;
+		struct sway_seat *seat = config->handler_context.seat;
+		struct sway_workspace *current = seat_get_focused_workspace(seat);
 		struct sway_workspace *ws = NULL;
 		if (strcasecmp(argv[0], "number") == 0) {
 			if (argc < 2) {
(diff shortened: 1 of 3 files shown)


sway/commands/workspace.c
```diff
@@ -181,7 +181,12 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 			++argv;
 		}
 
-
+		bool create = false;
```
RedSoxFan: This could be simplified to the following, which does fit on a single line:
```c
bool create = argc > 1 && strcasecmp(argv[1], "--create") == 0;
```

sway/commands/workspace.c
```diff
@@ -181,7 +181,12 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
...
+		if (argc > 1) {
+			create = strcasecmp(argv[1], "--create") == 0;
+		}
+		struct sway_seat *seat = input_manager_current_seat();
```
RedSoxFan: Since GitHub won't let me add the comment to the actual lines, can you change all ocurrances of seat to be the same for consistency? seat is redefined in the body for `back_and_forth` (L211) and `config->handler_context.seat` is used directly in `seat_consider_warp_to_focus` on L227.

Since this is in a command handler, `config->handler_context.seat` should be set, but `input_manager_current_seat()` should return `config->handler_context.seat` when set. I don't really care which is used, but it should be consistent

ddevault:
We can merge new features into master, I cherry pick bugfixes into the 1.0 branch for each rc.

RedSoxFan:
> We can merge new features into master, I cherry pick bugfixes into the 1.0 branch for each rc.

Oh ok

RedSoxFan:
Thanks!

PR #3701, opened 2019-02-16
Disable unneeded wlroots subproject features

emersion:


ianyfan:
What if we want these options enabled?

emersion:
You can override them, they're just defaults. Also, this only happens when a subproject is used. Building wlroots and building wlroots as a subproject are completely separate, build files are at different locations.

If you build sway, I don't think you'll ever want to have `sway/build/subproject/wlroots/rootston/rootston`.

ianyfan:
Meh, I do this so that I can have access to rootston for testing without having to build wlroots separately.

ianyfan:
But I am not against this change

ddevault:
Thanks!

JonnyMako:
I think this is breaking the build for me:
```
Program git found: YES (/usr/bin/git)

|
|Executing subproject wlroots 
|
|Value disabled is not boolean (true or false).
|
|Subproject wlroots is buildable: NO (disabling)
Dependency wlroots found: NO (tried pkgconfig and cmake)

meson.build:73:1: ERROR:  Dependency "wlroots" not found, tried pkgconfig and cmake
```

emersion:
Which version of Meson do you have?

PR #3704, opened 2019-02-16
Fix reload freeze when not modsetting current mode

### LINKED ISSUE #3685: Display freeze on config reload.
Reloading the config ($mod+Shift+C) under ordinary conditions is freezing my display. I think the system is still responsive because I can still hear audio play after triggering it.

Can anyone else reproduce this?

RedSoxFan:
Fixes #3685 with swaywm/wlroots#1545 -- which was
reverted in swaywm/wlroots#1550 due to #3685

On master, all output configs are applied on reload. This may cause an
output to have its config applied up to three times, instead of just
once. The three cases are: output name, output identifier, and wildcard.
Not only is this inefficient, but it can cause swaybg to be spawned and
immediately killed.

However, swaybg requires two roundtrips of wl_display (to obtain needed
globals) before it enters its normal event loop. Modesetting will
roundtrip the wl_display. Without modesetting, waitpid for killing
swaybg could block infinitely due to swaybg being blocked by
wl_display_roundtrip.

This only configured an output once. It either uses the wildcard config
or creates an empty wildcard config and applies that. This also fixes a
bug where an output would not be reset when there is no output config to
apply to it.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -391,6 +391,17 @@ void apply_output_config_to_outputs(struct output_config *oc) {
 	}
 }
 
+void reset_outputs(void) {
+	struct output_config *oc = NULL;
+	int i = list_seq_find(config->output_configs, output_name_cmp, "*");
+	if (i >= 0) {
+		oc = config->output_configs->items[i];
+	} else {
+		oc = store_output_config(new_output_config("*"));
+	}
+	apply_output_config_to_outputs(oc);
+}
+
 void free_output_config(struct output_config *oc) {
 	if (!oc) {
 		return;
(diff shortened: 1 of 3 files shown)


sway/config/output.c
```diff
@@ -391,6 +391,17 @@ void apply_output_config_to_outputs(struct output_config *oc) {
 	}
 }
 
+void reset_outputs(void) {
+	struct output_config *oc = NULL;
+	int i = list_seq_find(config->output_configs, output_name_cmp, "*");
```
emersion: What about non-wildcard configs?

RedSoxFan: The wildcard config is only ever applied directly when there is no output specific config. If there is an output specific config, that is applied instead.

https://github.com/swaywm/sway/blob/77a3694a782f48ea476958bf97e3f82ebafae0b1/sway/config/output.c#L375-L381

emersion: Ah, I see

emersion:
Thanks!

Issue #3703, opened 2019-02-16
Keyboard rate/delay no changes


krnl0138:
Changed configuration of keyboard delay/rate.

**Expected Behavior.** The keystrokes become faster
**Current Behavior.** Nothing changes.

sway-git 1.0-rc1-64-g5c78e5cd (Feb 16 2019, branch 'master')
wayland-git Installed: 1.16.90.2144.d7b0b79-1
wlroots-git Installed: 0.3.r14.g9e49ceb1-1
zsh 5.7.1 (x86_64-pc-linux-gnu)
`swaymsg -t get-inputs`: https://pastebin.com/XgecAQgE

According to wiki I put this line into my `$HOME/.config/sway/config`.
I substitute an identifier with `"1:1:AT_Translated_Set_2_keyboard"` respectively.
```
input <identifier> {
    repeat_delay 500
    repeat_rate 5
} 
```
I tried different numbers and identifiers from inputs, with or without quotes.
No changes after reboot or re-login. 

I also tried to put into `$HOME/.zshenv`
```
export WLC_REPEAT_RATE=5
export WLC_REPEAT_DELAY=500
```
and also passed them with `env` command on sway boot. No changes again.

**Note:** I got my layouts and grp:caps_toggle managed via enviroments in .zshenv. 
But when i put them into sway config in a form of 
```
xkb_layout us,de
xkb_options grp:caps_toggle
```
while commenting them in `.zshenv` file accordingly, nothing happens and I can't change my layout.

What am I doing wrong?

RedSoxFan:
Can you please send a debug log (`sway -d 2>sway.log`)?

Also, sway does not use wlc anymore so any of the `WLC_` environment variables will have no effect.

krnl0138:
@RedSoxFan, here it is. https://pastebin.com/1wmhGKGV

RedSoxFan:
Just as a sanity check, can you try using `input * {` instead of the identifier.

Also, I see you are using alacritty in the log. Are you experiencing the issue in all applications or just alacritty?

krnl0138:
@RedSoxFan, thanks, I figured it out, kind of.

1. Everything is working until I have `input *` in my config, that's great news.
2. **Alacritty** inverts rate property, e.g. if I have 80 then **alacritty** types really slowly, but **xterm** and **st** are smooth and fast. However, if I nailed it to 10, **alacritty** starts to fly, but typing in chromium and other apps becomes slow as hell.

[...]

PR #3707, opened 2019-02-16
sway-input.5: document wildcard and id troubleshooting

RedSoxFan:
See https://github.com/swaywm/sway/issues/3703#issuecomment-464391058

This documents the wildcard character for both inputs and seats. There
is also a tip added on trying the wildcard to verify a setting if the
identifier does not appear to be working.


emersion:
Typo in the commit message?

RedSoxFan:
> Typo in the commit message?

Eh, I guess tshooting isn't as common of a shorthand for troubleshooting as I thought. Maybe a regional thing or something since searching for it doesn't yield results. Changed the commit message.

emersion:
Thanks!

PR #3708, opened 2019-02-16
Disconnect swaybg instead of killing it

emersion:
This is much more reliable. This also fixes race conditions when killing swaybg
while it's doing a wl_display_roundtrip.


RedSoxFan:
I'm getting defunct swaybg processes with this when the output config changes, an output is unplugged, and after reloads

emersion:
Ah, yes. Hrm, we need to double-fork and `waitpid`.

emersion:
Should be fixed now.

RedSoxFan:
Thanks!

PR #3709, opened 2019-02-17
workspace_next_name: fallback to next available number

RedSoxFan:
This changes `workspace_next_name` to use the next available number as
the workspace name instead of the number of outputs. This fixes the case
where a number that is already in use could be returned. The workspace
numbers in use have no relation to the number of outputs so it makes
more sense to use the lowest available number

### DIFF
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -334,16 +334,13 @@ char *workspace_next_name(const char *output_name) {
 	if (target != NULL) {
 		return target;
 	}
-	// As a fall back, get the current number of active workspaces
-	// and return that + 1 for the next workspace's name
-	int ws_num = root->outputs->length;
-	int l = snprintf(NULL, 0, "%d", ws_num);
-	char *name = malloc(l + 1);
-	if (!sway_assert(name, "Could not allocate workspace name")) {
-		return NULL;
-	}
-	sprintf(name, "%d", ws_num++);
-	return name;
+	// As a fall back, use the next available number
+	char name[12] = "";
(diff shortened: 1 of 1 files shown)


sway/tree/workspace.c
```diff
@@ -334,15 +334,19 @@ char *workspace_next_name(const char *output_name) {
...
-		return NULL;
-	}
-	sprintf(name, "%d", ws_num++);
+	// As a fall back, use the next available number
+	char *name = NULL;
+	int ws_num = 1;
+	do {
+		if (!name || ws_num % 10 == 0) {
+			int l = snprintf(NULL, 0, "%d", ws_num);
+			name = realloc(name, l + 1);
```
emersion: Instead, we could just have a 12-byte array on the stack.

RedSoxFan: Changed

sway/tree/workspace.c
```diff
@@ -334,16 +334,13 @@ char *workspace_next_name(const char *output_name) {
...
-	if (!sway_assert(name, "Could not allocate workspace name")) {
-		return NULL;
-	}
-	sprintf(name, "%d", ws_num++);
-	return name;
+	// As a fall back, use the next available number
+	char name[12] = "";
+	unsigned int ws_num = 1;
+	do {
+		sprintf(name, "%u", ws_num++);
```
emersion: Can we use `snprintf` here, just in case?

RedSoxFan: Done

emersion:
Thanks!

PR #3710, opened 2019-02-17
Unbreak build on FreeBSD

jbeich:
Regressed by #3708. FreeBSD (even 13.0-CURRENT) still defines `SOCK_CLOEXEC` only when `_POSIX_C_SOURCE` is **not** defined.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -1,4 +1,8 @@
 #define _POSIX_C_SOURCE 200809L
+#ifdef __FreeBSD__
+// for SOCK_CLOEXEC
+#define __BSD_VISIBLE 1
+#endif
 #include <assert.h>
 #include <fcntl.h>
 #include <stdbool.h>


ddevault:
Is the more portable solution to set CLOEXEC on the file descriptor with fnctl after the fact?

jbeich:
DragonFly, NetBSD, Solaris (unlike FreeBSD, OpenBSD) don't hide `SOCK_CLOEXEC`. Unlike GNU libc those systems enable all features by default. A more portable way would be to replace `_POSIX_C_SOURCE` with `_GNU_SOURCE` unless compatibility with an old standard is important.

See also [POSIX#411](http://austingroupbugs.net/view.php?id=411). This fix tries to keep consistency with **wlroots**.

ddevault:
>A more portable way would be to replace _POSIX_C_SOURCE with _GNU_SOURCE unless compatibility with an old standard is important

That would be *less* portable. We're a POSIX shop here. I'm wondering if we can do this POSIXly by saving the file descriptor configuration until after it's opened.

emersion:
This code was borrowed from wlroots' xwayland code. A fix would need to be made there too if we don't use `SOCK_CLOEXEC`.

We could have a `set_cloexec` function instead (we already unset CLOEXEC right before the `exec` call).

emersion:
Superseded by https://github.com/swaywm/sway/pull/3729

Issue #2377, opened 2018-07-29
Add touch support to swaybar


ddevault:

emersion:
(Or we could just use the pointer fallback)

SirCmpwn:
I want to be able to touch and drag left/right to cycle through workspaces like we scroll through them today.

SirCmpwn:
But yeah swaynag can totally just use the pointer fallback.

PR #3711, opened 2019-02-17
Initial pass on touch support for swaybar

ddevault:
- Activate hotspots on tap
- Cycle workspaces on swipe

Future patches:

- Make padding (constants in `render.c`) configurable so my fat fingers have bigger targets
- Incorporate with #3697 to create new workspaces on swipe(?)

Fixes https://github.com/swaywm/sway/issues/2377

### DIFF
--- a/swaybar/input.c
+++ b/swaybar/input.c
@@ -123,6 +123,23 @@ static bool check_bindings(struct swaybar *bar, uint32_t button,
 	return false;
 }
 
+static void process_hotspots(struct swaybar_output *output,
+		double x, double y, uint32_t button) {
+	x *= output->scale;
+	y *= output->scale;
+	struct swaybar_hotspot *hotspot;
+	wl_list_for_each(hotspot, &output->hotspots, link) {
+		if (x >= hotspot->x && y >= hotspot->y
+				&& x < hotspot->x + hotspot->width
+				&& y < hotspot->y + hotspot->height) {
+			if (HOTSPOT_IGNORE == hotspot->callback(output, hotspot,
+					x / output->scale, y / output->scale, button, hotspot->data)) {
+				return;
+			}
+		}
+	}
+}
+
 static void wl_pointer_button(void *data, struct wl_pointer *wl_pointer,
 		uint32_t serial, uint32_t time, uint32_t button, uint32_t state) {
 	struct swaybar *bar = data;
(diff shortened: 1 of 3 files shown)


swaybar/input.c
```diff
@@ -278,6 +398,10 @@ static void seat_handle_capabilities(void *data, struct wl_seat *wl_seat,
 		bar->pointer.pointer = wl_seat_get_pointer(wl_seat);
 		wl_pointer_add_listener(bar->pointer.pointer, &pointer_listener, bar);
 	}
+	if ((caps & WL_SEAT_CAPABILITY_TOUCH)) {
```
emersion: Need to release `wl_touch` in case the seat looses the capability (just like the pointer, see https://github.com/swaywm/sway/pull/3711/files#diff-00a64d3193d21f96087133944068ca3eR393).

swaybar/input.c
```diff
@@ -267,6 +277,116 @@ struct wl_pointer_listener pointer_listener = {
...
+		wl_fixed_t major, wl_fixed_t minor) {
+	// Who cares
+}
+
+static void wl_touch_orientation(void *data, struct wl_touch *wl_touch,
+		int32_t id, wl_fixed_t orientation) {
+	// Who cares
+}
+
+static struct wl_touch_listener touch_listener = {
```
emersion: This can be const

swaybar/input.c
```diff
@@ -267,6 +277,116 @@ struct wl_pointer_listener pointer_listener = {
 	.axis_discrete = wl_pointer_axis_discrete,
 };
 
+static struct touch_slot *get_touch_slot(struct swaybar_touch *touch, int32_t id) {
+	int next = -1;
```
emersion: This can be `ssize_t`

swaybar/input.c
```diff
@@ -267,6 +277,116 @@ struct wl_pointer_listener pointer_listener = {
...
+	int prev_progress = (int)((slot->x - slot->start_x)
+			/ slot->output->width * 100);
+	slot->x = wl_fixed_to_double(x);
+	slot->y = wl_fixed_to_double(y);
+	// "progress" is a measure from 0..100 representing the fraction of the
+	// output the touch gesture has travelled, positive when moving to the right
+	// and negative when moving to the left.
+	int progress = (int)((slot->x - slot->start_x)
+			/ slot->output->width * 100);
+	if (abs(progress) % 10 < abs(prev_progress) % 10) {
```
emersion: I'm not sure I get how this condition works. Why use a modulo?

ddevault: This is true each time your finger slides past a threshold of 10% of the width of the screen.

ddevault: Actually this is pretty weird, not sure what I was thinking here. Will revisit this

benutzer193: Would following condition work?
```c
if (abs(progress - prev_progress) > 0)
```

ddevault: I think so, yep

ddevault: Actually, not sure it would. That would be checking that you are moving *fast* enough, not *far* enough.

emersion:
```c
if (abs(progress) > 10)
```

?

ddevault: No, after further testing and reasoning I believe the original conditional is correct.

emersion: Well I still don't understand it.

ddevault: Progress is a value from `[0..100)` which represents how far your finger has travelled from its starting point to the starting point plus the width of the screen. We subdivide the bar into 10 segments, and passing from one segment to the next causes the workspace to change. We offset these by your starting position, so if you touch, then travel 10% of the screen, you go to workspace +1, then 10% more and workspace +2, and so on.

`progress` will be `< prev_progress` when you cross one of these thresholds - the previous value will be near the end of the last segment, so a higher number (in this example 9), and the new value will be near the start of the next segment, so a lower number (in this example 2).

ddevault: Actually, reworked this a bit to make it behave a bit more predictably. Does the new approach make more sense?

emersion:
Nice, thanks!

PR #3713, opened 2019-02-18
Use container under cursor for mouse bindings

RedSoxFan:
This matches i3's behavior of executing mouse bindings in regards to the
container under the cursor instead of what is focused.

[i3 User Guide 4.4](https://i3wm.org/docs/userguide.html#mousebindings): A mouse binding makes i3 execute a command upon pressing a specific mouse button **in the scope of the clicked container**


ddevault:
Thanks!

PR #3725, opened 2019-02-19
layout cmd: always operate on parent container, like i3

### LINKED ISSUE #3724: layout change targets focused container, while in i3 it targets parent
In i3, layout change targets the parent of the selected container always.

In sway, layout change targets the parent only if the selected container is a window, otherwise it targets the selected container.

The behavior is only different when there are three or more levels of containers, and an intermediate level container is selected.

For example, with [ a b ] selected in the following layout,

```
 -------
| a |   |
 ---| c |
| b |   |
 -------
```

pushing `Ctrl-e` would change to

```
 -------
| | |   |
|a|b| c |
| | |   |
 -------
```
while in i3 it would change to

```
 -------
|   a   |
|-------|
|   b   |
|-------|
|       |
|   c   |
|       |
 -------
```

I noticed this the first time I played with sway but of course at that time there were bigger issues. Now things are so stable I'm nitpicking :)

If it's confirmed as a bug, I'd like to fix it as a #firstpatch since it's probably a one-liner (c isn't my native lang)

edrex:
Fixes #3724

### DIFF
--- a/sway/commands/layout.c
+++ b/sway/commands/layout.c
@@ -111,10 +111,8 @@ struct cmd_results *cmd_layout(int argc, char **argv) {
 				"Unable to change layout of floating windows");
 	}
 
-	// Typically we change the layout of the current container, but if the
-	// current container is a view (it usually is) then we'll change the layout
-	// of the parent instead, as it doesn't make sense for views to have layout.
-	if (container && container->view) {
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Thanks!

Issue #2681, opened 2018-09-21
Build system: build sway with the xwayland and x11-backend meson options disabled


RyanDwyer:
Running an additional build with these disabled will make it easier to catch places where we've forgotten to guard them.

PR #3726, opened 2019-02-19
ci: add Alpine build

emersion:
Fixes https://github.com/swaywm/sway/issues/2681


ddevault:
Thanks! FreeBSD build soon?

PR #3737, opened 2019-02-20
output_evacuate: call workspace_consider_destroy

RedSoxFan:
This calls `workspace_consider_destroy` on the workspace that was
visible on an output that a workspace was just evacuated to. This
prevents having hidden empty workspaces.


emersion:
Thanks!

Issue #3658, opened 2019-02-12
Segmentation fault in output_render


Emantor:
```
#0  0x000055d86bbbcc94 in output_render (output=0x55d86c2bbca0, when=0x7ffc1f747310, damage=0x7ffc1f747320)
    at ../sway/desktop/render.c:1006
        clear_color = {0, 0, 0, 1}
        nrects = 1
        rects = 0x7ffc1f747320
        wlr_output = 0x55d86be62400
        renderer = 0x55d86be4ee20
        __PRETTY_FUNCTION__ = "output_render"
        workspace = 0x55d86c436390
        fullscreen_con = 0x3f00000077e
        seat = 0x0
        focus = 0x7ffc00000002
        width = 527725168
        height = 1
        transform = 32764
#1  0x000055d86bbb9100 in damage_handle_frame (listener=0x55d86c2bbe18, data=0x55d86c2bbe60)
    at ../sway/desktop/output.c:386
        output = 0x55d86c2bbca0
        now = {tv_sec = 244029, tv_nsec = 491214679}
        needs_swap = true
        damage = {extents = {x1 = 0, y1 = 0, x2 = 1920, y2 = 1080}, data = 0x0}
#2  0x00007fd4d7163795 in wlr_signal_emit_safe (signal=0x55d86c2bbec0, data=0x55d86c2bbe60)
    at ../subprojects/wlroots/util/signal.c:29
        pos = 0x55d86c2bbe18
        l = 0x55d86c2bbe18
        cursor = {link = {prev = 0x55d86c2bbe18, next = 0x7ffc1f747390}, notify = 0x7fd4d71636df <handle_noop>}
        end = {link = {prev = 0x7ffc1f747370, next = 0x55d86c2bbec0}, notify = 0x7fd4d71636df <handle_noop>}
#3  0x00007fd4d715037e in output_handle_frame (listener=0x55d86c2bbf58, data=0x55d86be62400)
    at ../subprojects/wlroots/types/wlr_output_damage.c:50
        output_damage = 0x55d86c2bbe60
#4  0x00007fd4d7163795 in wlr_signal_emit_safe (signal=0x55d86be62520, data=0x55d86be62400)
    at ../subprojects/wlroots/util/signal.c:29
        pos = 0x55d86c2bbf58
        l = 0x55d86c2bbf58
        cursor = {link = {prev = 0x55d86c2bbf58, next = 0x7ffc1f747430}, notify = 0x7fd4d71636df <handle_noop>}
[...]

PR #3740, opened 2019-02-21
Fix small memory leaks reported in #3658

Emantor:
I went through the Address Sanitizer Report from #3658 and fixed the missing IPC frees.


Emantor:
Sound argument, amended and pushed.

ianyfan:
Thanks!

PR #3741, opened 2019-02-21
Add configuration example for gaps

xyproto:
* Add a commented out section in `config.in` with an example for how to enable the gaps feature.


xyproto:
Just using a commented out example for configuring gaps was suggested in #2955.

ddevault:
Thank you, but we'll pass on this patch.

PR #3743, opened 2019-02-21
output_get_active_workspace: check workspaces length

RedSoxFan:
Fixes #3658 
Fixes #3654 
Fixes #3043 

If an output's node was dirty and the transaction was committed before a
workspace was moved to or created for the output, the instruction would
have a bad value for `state->active_workspace` due to a missing
length check in `output_get_active_workspace`. If there was no focus on
the output, the first workspace was being returned. If the workspace
list was currently empty, the value was either garbage, or in the case of
an output being disabled and re-enabled, a workspace that may have been
previously freed. This just adds the length check to avoid returning out
of bounds value.

PR #3744, opened 2019-02-21
Handle NULL from output_get_active_workspace

RedSoxFan:
Follow-up for #3743 

This modifies the places where output_get_active_workspace is called to
handle a NULL result. Some places already handled it and did not need a
change, some just have guard off code blocks, others return errors, and
some have sway_asserts since the case should never happen. A lot of this
is probably just safety precautions since they probably will never be
called when `output_get_active_workspace` is not fully configured with a
workspace.

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -37,6 +37,9 @@ static struct sway_node *get_node_in_output_direction(
 		struct sway_output *output, enum wlr_direction dir) {
 	struct sway_seat *seat = config->handler_context.seat;
 	struct sway_workspace *ws = output_get_active_workspace(output);
+	if (!sway_assert(ws, "Expected output to have a workspace")) {
+		return NULL;
+	}
 	if (ws->fullscreen) {
 		return seat_get_focus_inactive(seat, &ws->fullscreen->node);
 	}
--- a/sway/commands/titlebar_border_thickness.c
+++ b/sway/commands/titlebar_border_thickness.c
@@ -21,7 +21,12 @@ struct cmd_results *cmd_titlebar_border_thickness(int argc, char **argv) {
 
 	for (int i = 0; i < root->outputs->length; ++i) {
 		struct sway_output *output = root->outputs->items[i];
-		arrange_workspace(output_get_active_workspace(output));
+		struct sway_workspace *ws = output_get_active_workspace(output);
+		if (!sway_assert(ws, "Expected output to have a workspace")) {
+			return cmd_results_new(CMD_FAILURE,
+					"Expected output to have a workspace");
+		}
+		arrange_workspace(ws);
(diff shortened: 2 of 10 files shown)


sway/commands/focus.c
```diff
@@ -37,6 +37,8 @@ static struct sway_node *get_node_in_output_direction(
 		struct sway_output *output, enum wlr_direction dir) {
 	struct sway_seat *seat = config->handler_context.seat;
 	struct sway_workspace *ws = output_get_active_workspace(output);
+	if (!ws) {
```
emersion: Missing if body?

RedSoxFan: Not sure what happened there. Should be fixed

emersion:
Maybe we should just put assertions in command handlers too? I don't think there's a case in which commands should be executed in such a state? Not sure, though.

RedSoxFan:
> Maybe we should just put assertions in command handlers too? I don't think there's a case in which commands should be executed in such a state? Not sure, though.

I changed it to have sway_asserts for the command handlers. It should be impossible to handle a command while before `output_enable` returns.

sway/commands/titlebar_border_thickness.c
```diff
@@ -21,8 +21,11 @@ struct cmd_results *cmd_titlebar_border_thickness(int argc, char **argv) {
 
 	for (int i = 0; i < root->outputs->length; ++i) {
 		struct sway_output *output = root->outputs->items[i];
-		arrange_workspace(output_get_active_workspace(output));
-		output_damage_whole(output);
+		struct sway_workspace *ws = output_get_active_workspace(output);
+		if (ws) {
```
emersion: Can this be executed outside of a command handler?

sway/ipc-json.c
```diff
@@ -199,7 +199,7 @@ static void ipc_json_describe_output(struct sway_output *output,
 
 	struct sway_workspace *ws = output_get_active_workspace(output);
 	json_object_object_add(object, "current_workspace",
-			json_object_new_string(ws->name));
+			ws ? json_object_new_string(ws->name) : NULL);
```
emersion: This should probably be an assertion too

sway/commands/titlebar_border_thickness.c
RedSoxFan: Nope. Just missed it

sway/ipc-json.c
RedSoxFan: Changed

emersion:
Hrm, CI hasn't been started it seems. Can you try to force-push?

ddevault:
Linode had an outage and the *one* service I still have on Linode was affected. Should be good now.

emersion:
Thanks!

PR #3749, opened 2019-02-22
move scratchpad: hide visible scratchpad container

### LINKED ISSUE #3748: Scratchpad inconsistency with i3
Sway Version: sway version 1.0-beta.2-300-g204e1f47 (Jan 31 2019, branch 'master')

When a scratchpad window is "shown" via "scratchpad show", using "move scratchpad" does not send it back into the scratchpad like it does in i3. Minor issue, nothing significant.

RedSoxFan:
Fixes #3748

This makes it so running `move [to] scratchpad` on a container already
in the scratchpad does not return an error. To match i3's behavior, a
visible scratchpad container will be hidden and a hidden scratchpad
container will be treated as a noop.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -854,11 +854,11 @@ static struct cmd_results *cmd_move_to_scratchpad(void) {
 		}
 	}
 
-	if (con->scratchpad) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3753, opened 2019-02-22
input/cursor: allow whole-window bindings on ws

### LINKED ISSUE #3750: Make bindsym --whole-window work on empty workspace like i3
sway version 1.0-rc1-98-gd3d79565 (Feb 22 2019, branch 'master')

Using bindsym --whole-window $mod+button2 works when over an open window (e.g. termite) and any command bound to it functions as it should. In my case, I have a launcher menu (which is actually termite launched like Rofi would be) set to spawn. If you $mod+button2 on an empty workspace/desktop however, nothing happens. An enhancement was filed against i3 for this feature: https://github.com/i3/i3/issues/2097

How difficult would this be to incorporate in Sway? I am prepared to run the latest git, provide logs or provide whatever information you may need.

RedSoxFan:
Fixes #3750 

To match i3's behavior, this allows mouse bindings to be triggered over
a workspace when `--whole-window` is given.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -552,10 +552,11 @@ static void state_add_button(struct sway_cursor *cursor, uint32_t button) {
 static struct sway_binding* get_active_mouse_binding(
 		const struct sway_cursor *cursor, list_t *bindings, uint32_t modifiers,
 		bool release, bool on_titlebar, bool on_border, bool on_content,
-		const char *identifier) {
-	uint32_t click_region = (on_titlebar ? BINDING_TITLEBAR : 0) |
-			(on_border ? BINDING_BORDER : 0) |
-			(on_content ? BINDING_CONTENTS : 0);
+		bool on_workspace, const char *identifier) {
+	uint32_t click_region =
+			((on_titlebar || on_workspace) ? BINDING_TITLEBAR : 0) |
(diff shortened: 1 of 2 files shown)


sway/input/cursor.c
```diff
@@ -564,6 +565,7 @@ static struct sway_binding* get_active_mouse_binding(
 				cursor->pressed_button_count != (size_t)binding->keys->length ||
 				release != (binding->flags & BINDING_RELEASE) ||
 				!(click_region & binding->flags) ||
+				(on_workspace && click_region != binding->flags) ||
```
emersion: Should `click_region != binding->flags` be a bitwise operation instead?

RedSoxFan: `binding->flags` must be `BINDING_TITLEBAR | BINDING_BORDER | BINDING_CONTENTS` (aka whole-window) for the bindings to trigger on a workspace. Earlier in the function, `click_region` is set to that value so a simple equality check should be sufficient

RedSoxFan: Actually, nevermind. It does need to be bitwise since there are other non-region flags

RedSoxFan: Changed

emersion:
Thanks!

PR #3758, opened 2019-02-23
fish: improve completions a little bit

sanga:
Sprinkled a few `-f` (don't complete from the filesystem) and `-r` (requires param) around. Also the swaynag completion help was referring to swaylock, so fixed that too (note: I suppose those config file locations apply to swaynag?)


emersion:
Thanks!

PR #3761, opened 2019-02-23
Fix crash exiting fullscreened floating container

minus7:
container_floating_move_to_center and container_fullscreen_disable were
calling recursively when the container spawned as a fullscreen floating
container (via for_window). Such a window now doesn't crash sway anymore
but is still configured with a wrong, zero size, making it not directly
usable.

I didn't see this patch break anything that wasn't already broken.


RedSoxFan:
Thanks!

PR #3762, opened 2019-02-23
ipc: add missing fields to disabled outputs

cjbassi:
i3 requires all outputs to have certain fields, including 'primary', 'current_workspace', and 'rect' which were missing on disabled outputs.

https://i3wm.org/docs/ipc.html#_outputs_reply

Fixes a bug I was having with [i3ipc-rs](https://github.com/tmerr/i3ipc-rs) where my program would crash when calling get_outputs() because the fields were missing.


emersion:
Can you rebase? It seems wlroots is too new for this PR.

cjbassi:
Done!

emersion:
Thanks!

PR #3765, opened 2019-02-24
Add swaybar overlay mode

### LINKED ISSUE #1620: [feature request] swaybar background mode
On my system, I have window borders set up to be transparent except for the title and swaybar the same way. With this configuration, I end up with roughly double the height of the title for the top screen margin, whereas if swaybar could be in the background & windows overlaid the entire top margin would only take up one line of space.

milkey-mouse:
Fixes #1620. Overlay mode puts swaybar above normal windows (in the overlay layer) and ignores input events. In my case I'm using it to always have the time in the corner of the screen. It probably works best with a theme/color scheme like mine, where the swaybar background color is transparent.


RedSoxFan:
Thanks!

PR #3767, opened 2019-02-24
introduce hide_edge_borders lone_tab

### LINKED ISSUE #3663: Add `hide_lone_tabs` to hide titlebar when there is only one tab
This is a duplicate of https://github.com/swaywm/sway/issues/2987, which was closed even though the problem is still here on latest sway (RC2).

This one is a bug and a regression from i3 behaviour.

**Required config:**

new_window pixel 2
new_float pixel 2
workspace_layout default
hide_edge_borders smart (this one is probably not be related)

**Current behaviour:**

    * Open a window on an empty workspace (titlebar is hidden)

    * Open a second window on the same workspace (titlebars are visible)

    * Put them into stack or tabbed (titlebars are visible)

    * Close one window, leaving only one -> **titlebar is visible but should be hidden**

When there is only one window in a workspace, it should have only one container (previous parents should be removed) and title bar should be hidden if configured so.

thejan2009:
Behaves same as _both_ in i3wm with regards to tabbed/stacked containers. If
there is only one tabbed/stacked container child, title bar is hidden.

Related issues and merge requests: #3031, #3002, #2912, #2987.
Fixes #3663

### DIFF
--- a/sway/commands/hide_edge_borders.c
+++ b/sway/commands/hide_edge_borders.c
@@ -5,8 +5,14 @@
 
 struct cmd_results *cmd_hide_edge_borders(int argc, char **argv) {
 	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "hide_edge_borders", EXPECTED_EQUAL_TO, 1))) {
+	if ((error = checkarg(argc, "hide_edge_borders", EXPECTED_AT_LEAST, 1))) {
 		return error;
+	}
+
+	if (strcmp(*argv, "--i3") == 0) {
+		config->hide_lone_tab = true;
+		++argv;
+		--argc;
 	}
 
 	if (strcmp(argv[0], "none") == 0) {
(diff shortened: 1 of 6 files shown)


thejan2009:
Interesting, I wasn't aware of flags in the config. It'd certainly be a better solution because all the options can be adjusted to behave as in i3. I'm sure this will affect the `horizontal` option too, and I'll have to look into what `smart` and `smart_no_gaps` should be doing.

thejan2009:
The optional flag together with the original config command seems quite a mouthful, so I simplified it a bit.

RedSoxFan:
Can you remove the `hide_edge_borders lone_tab` remnants since the flag more powerful?

sway/commands/hide_edge_borders.c
```diff
@@ -23,6 +29,7 @@ struct cmd_results *cmd_hide_edge_borders(int argc, char **argv) {
 		config->hide_edge_borders = E_SMART_NO_GAPS;
 	} else {
 		return cmd_results_new(CMD_INVALID, "Expected 'hide_edge_borders "
+				"'[--i3]' "
```
RedSoxFan: The line wrapping is off here and there should not be single quotes around the flag. It looks like the flags does fit on the line below once the single quotes are removed.

RedSoxFan:
Thanks!

meiyopeng:
Hi.  How about adding this option to the smart_borders command?

RedSoxFan:
> Hi. How about adding this option to the smart_borders command?

Out of curiosity, is there any reason you are using `smart_borders` instead of `hide_edge_borders --i3 smart|smart_no_gaps`? I'd rather not add more complexity to something that is basically just a wrapper.

meiyopeng:
The only reason: `smart_borders` is shorter and is consistent with `smart_gaps`.

Issue #3770, opened 2019-02-25
RFC: Changes to output transform command.


c-edw:
There are a few changes I would like to make to the `transform` command possibly post-1.0. I'm willing to send a patch for these.

- The first one is 'additive' transforms. The would allow users to prefix the rotation value with a plus (`+`) or minus (`-`), and this value would be added to the current rotation value. An example usage:

```
# Each time I press $mod+R, my screen rotates 90 degrees.
bindsym $mod+R output eDP-1 transform +90
```

- The second is that I would like to improve the arg parsing for the `transform` command. Currently, it won't allow for a user to have a rotation of `0`, you have to pass `normal`. Same for `360` or `720` (obviously these are redundant, but still valid). To fix this, I would like to parse the value as an int and then apply a modulus operation to clamp it between 0-360 degrees. It looks like the command also allows `flipped-` in the value so the actual implementation will be a bit more complex.

RedSoxFan:
>  The first one is 'additive' transforms. The would allow users to prefix the rotation value with a plus (`+`) or minus (`-`), and this value would be added to the current rotation value.

`transform [flipped-](90|180|270) [clockwise|anticlockwise]` already exists for +/- 90/180/270 degree relative transformations. I'm not sure how useful arbitrary degree rotations would be, but there may be a use case

>  The second is that I would like to improve the arg parsing for the `transform` command. Currently, it won't allow for a user to have a rotation of `0`, you have to pass `normal`. Same for `360` or `720` (obviously these are redundant, but still valid). To fix this, I would like to parse the value as an int and then apply a modulus operation to clamp it between 0-360 degrees. It looks like the command also allows `flipped-` in the value so the actual implementation will be a bit more complex.

I'm fine with `0` as an alias for `normal`. However, I am not a fan of clamping the transformation. I think that whatever is supplying the transformation should be in charge of clamping the values and making sure it is supplying a valid value. My comments on the arbitrary degree rotations here are the same as above

ddevault:
>There are a few changes I would like to make to the transform command possibly post-1.0.

Note that we'll accept the patch into master whenever you want. We won't include it in 1.0 but 1.0 isn't based on master.

c-edw:
@RedSoxFan There isn't a way to have additive rotations right now. I didn't explain what I mean by that very clearly. If I specify a rotation of `+90`, then I would expect executing it four times to bring me back to the original rotation. Values without a plus/minus prefix will behave as they do now (a static rotation).

[...]

PR #3772, opened 2019-02-25
Allow 0 degree transform (normal transform).

c-edw:
There are some changes to this code still being discussed in #3770 but this is the main thing I wanted to change.


emersion:
Thanks!

PR #3774, opened 2019-02-25
fix smart_borders description in manual

carloabelli:
The wording for smart_borders was opposite the actual behavior.


RedSoxFan:
Thanks!

PR #3777, opened 2019-02-26
execute_command: do not strip qoutes for cmd_mode

### LINKED ISSUE #3775: Escape character in mode with pango markup
sway version 1.0-rc1-116-g2510e3df (Feb 24 2019, branch 'master')

I am trying to create a mode with pango markups (to add colors). Unfortunately, I noticed that the single and double quote characters (so '  and ") are escaped differently in the bindsym and in the mode commands. This is problematic since they are needed for the pango markup attributes. It took me 2 hours to figure out how to obtain something that works.  The purpose of that bug report is (1) to point out what does not work and why and (2) to give a syntax that works for the current master branch. 

So let's assume that we are trying to create a mode using the following $HELLO variable. The intuitive way does not work:  

```set $HELLO "<span bgcolor='blue'>Hello</span>"
bindsym Alt+F3 mode "$HELLO"
mode --pango_markup "$HELLO" {
   bindsym Escape mode "default"
}
```
So I tried several variants with and without double-quote and the following syntax is the only one that works:

```
set $HELLO "<span bgcolor='blue'>Hello</span>"
bindsym Alt+F3 mode "$$HELLO"
mode --pango_markup $HELLO {
   bindsym Escape mode "default"
}
```
  
So the rules are:
- The value of the variable **MUST** be "quoted" in the set command
- The variable given to bindsym mode **MUST USE A DOUBLE $**. Here the "quotes" are optional.
- The variable given to mode --pango_markup **MUST NOT BE QUOTED** and **MUST USE A SINGLE $**.

RedSoxFan:
Fixes #3775 
Related to #3209 

`cmd_mode` performs its own quote stripping for the mode string to
avoid double stripping quotes for `cmd_bindcode` and `cmd_bindsym` in
`config_command` and `execute_command`. Stripping quotes in
`execute_command` for `cmd_mode` will also result in double stripping,
which will cause issues for any mode string with spaces, such as pango
markup.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -255,7 +255,8 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 			int argc;
 			char **argv = split_args(cmd, &argc);
 			if (strcmp(argv[0], "exec") != 0 &&
-					strcmp(argv[0], "exec_always") != 0) {
+					strcmp(argv[0], "exec_always") != 0 &&
+					strcmp(argv[0], "mode") != 0) {
 				int i;
 				for (i = 1; i < argc; ++i) {
 					if (*argv[i] == '\"' || *argv[i] == '\'') {


emersion:
Thanks!

PR #3786, opened 2019-02-27
Set minimum wlroots version

emersion:
Makes it more obvious that building sway with an old copy of wlroots won't work.


ddevault:
Thanks!

PR #3787, opened 2019-02-27
Print Meson features

emersion:


RedSoxFan:
Can you change this to match the formatting that wlroots uses?

emersion:
I'm not a fan of the formatting wlroots uses.

ddevault:
Thanks!

PR #3766, opened 2019-02-24
Add sway-ipc.7.scd to document IPC protocol

RedSoxFan:
**~~This is still a WIP~~ and will ideally be completed before 1.0 is
released since it would be nice to have for release**

Closes #707 

This add `sway-ipc.7.scd` that documents the IPC protocol

Note: This increases the minimum scdoc version from 1.8.1 to 1.9.0 to
allow for table cells to be continued on the following line

PR #3788, opened 2019-02-27
meson: check scdoc version

emersion:


emersion:
@jbeich It seems scdoc's pkg-config file contains errors on FreeBSD. See https://builds.sr.ht/~sircmpwn/job/33711

emersion:
In the meantime, pushed a fix not to use pkg-config to find scdoc's path.

jbeich:
@emersion, I've fixed pkg-config's file and updated to 1.9.1. Expect FreeBSD package around 2019-03-03 ([details](https://github.com/swaywm/sway/pull/3766#issuecomment-467719559)).

emersion:
Wow, thanks a lot, that was fast :)

RedSoxFan:
Do you want this merged now or wait for FreeBSD (and possibly #3792)?

emersion:
Hmm. Problem is, we'll probably fail builds on many platforms until all of them get a patched (non-1.9.1) scdoc. We'll probably need to build scdoc from source because of this…

I think it'd be better to merge this as-is and improve it when the situation settles.

emersion:
Related: https://todo.sr.ht/~sircmpwn/scdoc/23

jbeich:
scdoc should probably migrate to meson. It's `install` target is partially broken on BSDs because `-D` flag has different semantics. In general, howegrown build systems are a constant source of pain downstream.

RedSoxFan:
> I think it'd be better to merge this as-is and improve it when the situation settles.

Sounds good

RedSoxFan:
Thanks!

emersion:
Submitted a patch to fix the pkg-config file: https://lists.sr.ht/~sircmpwn/public-inbox/%3C0k4nGuwl8cWAr8T9PlSbYQt33rnRn7rYoWEw-kKZ4WrhEK4Oco6AudUzWFmBKN8_7N_uJNcc_xY7iwzAE3s_2gD1Ie7NyDsAsdWjme3Vxss%3D%40emersion.fr%3E

eli-schwartz:
This came up today due to $REASONS (sway wasn't detecting scdoc correctly) and I cannot help but wonder

if perhaps, instead of patching the scdoc pkg-config file, it would make more sense to patch scdoc, from this:

```c
if (argc == 2 && strcmp(argv[1], "-v") == 0) {
```

to this
```c
if (argc == 2 && (strcmp(argv[1], "-v") == 0 || strcmp(argv[1], "--version") == 0)) {
```

Then from meson you can always have simply done, `find_program('scdoc', version: '>=1.9')`.

emersion:
I don't really like this approach, because `--version` is implemented differently by each program (there is no standard) and can't contain arbitrary key-value entries.

eli-schwartz:
> I don't really like this approach, because `--version` is implemented differently by each program (there is no standard)

You mean a lot of them accidentally use -v instead and then can't tell the difference between `grep -v --invert-match`, `ls -v (natural version sort)`, `scdoc -v`, and `gcc -v --verbose`?

Once you know the appropriate command flags, the version number is very easy to determine, as it is effectively always `(.* )?v?([0-9][0-9\.]+)\n` and anything else is a freaky exception.

> and can't contain arbitrary key-value entries.

Unsure why someone looking for a version number would expect key-value entries when looking for the required minimum. sway isn't looking for a key-value database of scdoc either (and scdoc.pc doesn't contain them) so I'm not really sure what the point here is.

Sure, for things a lot more complicated than "find a single standalone single-file executable" having a key-value database to store associated information is useful.

PR #3790, opened 2019-02-28
sway-ipc.7: clarify window_rect omits decorations

### LINKED ISSUE #3785: Incorrect value in window_rect.y
I want to configure grim to take a screenshot of the focused window using the output of `swaymsg -t get_tree`. The `rect` element gives the necessary information to get the whole geometry including the decorations. The `deco_rect` gives the title bar area (relative to `rect`). My problem is with `window_rect` and more specifically with its `y` value that does not really make any sense.

Am I right to assume that `window_rect` is supposed to provide the relative geometry of the window without its decorations?

In `ipc-json.c` the `window_rect` element is computed as follow:

```
	struct wlr_box window_box = {
		c->content_x - c->x,
              	(c->current.border == B_PIXEL) ? c->current.border_thickness : 0,
		c->content_width,
		c->content_height
	};
```

The value for `y` does not make sense to me since it ignores the height of the title bar (and potentially of multiple title bars in a vertically stacked container). I modified it as follow and that seems to work fine in all cases. 

```
	struct wlr_box window_box = {
		c->content_x - c->x,
                c->content_y - c->y,
		c->content_width,
		c->content_height
	};
```

I will post the python script I wrote to get the geometry for grim.

RedSoxFan:
Closes #3785 

According to the i3 ipc documentation, `window_rect` excludes the window
decorations from the calculation. This just clarifies that in
`sway-ipc.7.scd`

### DIFF
--- a/sway/sway-ipc.7.scd
+++ b/sway/sway-ipc.7.scd
@@ -317,7 +317,8 @@ node and will have the following properties:
 :  The absolute geometry of the node
 |- window_rect
 :  object
-:  The geometry of the contents inside the node
+:  The geometry of the contents inside the node. The window decorations are
+   excluded from this calculation, but borders are included.
 |- deco_rect
 :  object
 :  The geometry of the decorations inside the node


ddevault:
Thanks!

schauveau:
I do not think that the problem is fixed. Sway does not follow the I3 behavior. See my new comment in https://github.com/swaywm/sway/issues/3785#issuecomment-468253587

Issue #3635, opened 2019-02-10
Wrong colors in tray with some programs


gnidorah:
Last checked with `sway version 1.0-rc2`

KSNI applications have wrong colored icons in tray.

Demo program: https://github.com/gnidorah/ksni-tray-demo
Another example is a `kleopatra` application.

Screenshots:
i3
![screenshot_20190211_201920](https://user-images.githubusercontent.com/12064730/52581145-08607c80-2e3b-11e9-8ab6-7bbf0c004c1a.png)
sway 1.0-rc2
![1](https://user-images.githubusercontent.com/12064730/52581153-10202100-2e3b-11e9-867e-f9c605112030.png)
sway 0.15.2 doesn't have issues with colors. Perhaps we could just reuse its code?
![2019-02-13-195254_swaygrab](https://user-images.githubusercontent.com/12064730/52728917-35db3080-2fc9-11e9-9791-07716d3754c1.png)

Log:
```
2019-02-11 20:57:04 - [swaybar/tray/item.c:152] org.kde.StatusNotifierItem-18160-1/StatusNotifierItem IconName = 'kleopatra'
2019-02-11 20:57:04 - [swaybar/tray/item.c:75] org.kde.StatusNotifierItem-18160-1/StatusNotifierItem IconPixmap: found icon w:128 h:128
2019-02-11 20:57:04 - [swaybar/tray/item.c:75] org.kde.StatusNotifierItem-18160-1/StatusNotifierItem IconPixmap: found icon w:96 h:96
2019-02-11 20:57:04 - [swaybar/tray/item.c:75] org.kde.StatusNotifierItem-18160-1/StatusNotifierItem IconPixmap: found icon w:64 h:64
2019-02-11 20:57:04 - [swaybar/tray/item.c:75] org.kde.StatusNotifierItem-18160-1/StatusNotifierItem IconPixmap: found icon w:44 h:44
2019-02-11 20:57:04 - [swaybar/tray/item.c:75] org.kde.StatusNotifierItem-18160-1/StatusNotifierItem IconPixmap: found icon w:32 h:32
2019-02-11 20:57:04 - [swaybar/tray/item.c:75] org.kde.StatusNotifierItem-18160-1/StatusNotifierItem IconPixmap: found icon w:256 h:256
2019-02-11 20:57:04 - [swaybar/tray/item.c:96] org.kde.StatusNotifierItem-18160-1/StatusNotifierItem IconPixmap no. of icons = 6
```

[...]

PR #3794, opened 2019-02-28
tray: fix pixmap colors

ianyfan:
by converting from network byte order to host byte order

Fixes https://github.com/swaywm/sway/issues/3635


gnidorah:
@ianyfan Works fine for me. Thank you much!

emersion:
Thanks!

PR #3798, opened 2019-02-28
seat: don't send button release when not pressed

emersion:
All seat operations except "down" eat the button pressed event and don't send
it to clients. Thus, when ending such seat operations we shouldn't send the
button released event.

This commit moves the logic used to send pressed/released into the "down"
operation.

Related to https://github.com/swaywm/wlroots/pull/1586


RedSoxFan:
Thanks!

PR #3802, opened 2019-03-01
Allow arbitrary transforms/rotations.

### LINKED ISSUE #3770: RFC: Changes to output transform command.

c-edw:
Closes #3770.

Unfortunately this doesn't achieve the goal of having negative rotations because Sway tries to parse it as an argument.

It does however allow for arbitrary rotations such as `flipped-1080`. It also adds the ability to pass `normal-<deg>` for the sake of consistency.

### DIFF
--- a/sway/commands/output/transform.c
+++ b/sway/commands/output/transform.c
@@ -4,6 +4,13 @@
 #include "log.h"
 #include "sway/output.h"
 
+static enum wl_output_transform rotation_to_transform(int deg, bool flipped) {
+	// Calculate the modulus.
+	int mod = (((deg / 90) % 4) + 4) % 4;
+	// The last 4 variants of wl_output_transform are FLIPPED.
+	return flipped ? mod + 4 : mod;
+}
+
 struct cmd_results *output_cmd_transform(int argc, char **argv) {
 	if (!config->handler_context.output_config) {
 		return cmd_results_new(CMD_FAILURE, "Missing output config");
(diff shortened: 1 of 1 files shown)


RedSoxFan:
I'm still -1 on this. All this does is complicate the code to do a modulus on the degree given, which I still think should be done by whatever/whoever is feeding the transform to sway. Additionally, this will accept any integer and truncate it one of the valid values. I do not think giving `output <output> transform 143` should be a valid equivalent to `output <output> transform 90`. Also, this is not arbitrary rotations since there are 4 possible rotations this will map to (8 possible transforms).

Putting my opinions on this aside, in case the others disagree with me, here is an actual review:
- There is way too much duplicated code with minor changes. It needs to be cleaned up in some manor
- Although they are unlikely to change, the values of the wl_output_transform enum should not be assumed. Use an array and index it to get the transform's value.
- Consolidate the `This allows for:` comment into a command syntax instead of a list
- If the value given is not a multiple of 90, an error should be given instead of truncating it to a valid one.
- The commit message has to be changed since as I stated above, this is not arbitrary rotations

> Unfortunately this doesn't achieve the goal of having negative rotations because Sway tries to parse it as an argument.

Negative rotations should actually work with this code. Use `swaymsg -- output <output> transform -<degree>` or `swaymsg "output <output> transform -<degree>"`

emersion:
>I'm still -1 on this

I'm also -1 on this. Adding to much magic to sway commands won't do any good IMHO.

Sorry, but we'll pass on this one.

PR #3803, opened 2019-03-01
Minor fix of code duplication.

hugbubby:
This change removes 3~ lines of code that didn't need to be restated. Doesn't modify functionality, except for reordering the checks in the nested if statement to mirror its parent aesthetically.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -186,11 +186,7 @@ static void log_kernel(void) {
 
 static bool drop_permissions(void) {
 	if (getuid() != geteuid() || getgid() != getegid()) {
-		if (setgid(getgid()) != 0) {
-			sway_log(SWAY_ERROR, "Unable to drop root, refusing to start");
-			return false;
-		}
-		if (setuid(getuid()) != 0) {
+		if (setuid(getuid()) != 0 || setgid(getgid()) != 0) {
 			sway_log(SWAY_ERROR, "Unable to drop root, refusing to start");
 			return false;
 		}


sway/main.c
```diff
@@ -186,16 +186,12 @@ static void log_kernel(void) {
...
-			return false;
-		}
-		if (setuid(getuid()) != 0) {
+		if (setuid(getuid()) != 0 || setgid(getgid()) != 0) {
 			sway_log(SWAY_ERROR, "Unable to drop root, refusing to start");
 			return false;
 		}
 	}
-	if (setuid(0) != -1) {
+	if (setuid(0) == 0) {
```
emersion: -1. It's better to enter the error condition if something very bad happens, like `setuid` returning 42. Additionally, the spec says that if `setuid` fails, it must return -1, but doesn't say that if `setuid` succeeds, it should return 0.

hugbubby:
I'm a little confused on how to parse these error messages in the CI. What does "Table could not be indented" mean?

sway/main.c
hugbubby: It does say that if it succeeds, it should return 0, at least in the man page.

emersion: Yeah, sorry, skipped the important bit.

emersion:
Don't worry about it, scdoc 1.9.1 i broken and causes CI failures. Unfortunately a fix hasn't been pushed yet to scdoc master, so we can't do a lot about it.

sway/main.c
```diff
@@ -186,16 +186,12 @@ static void log_kernel(void) {
...
-			return false;
-		}
-		if (setuid(getuid()) != 0) {
+		if (setuid(getuid()) != 0 || setgid(getgid()) != 0) {
 			sway_log(SWAY_ERROR, "Unable to drop root, refusing to start");
 			return false;
 		}
 	}
-	if (setuid(0) != -1) {
+	if (setuid(0) == 1) {
```
emersion: Eh?

hugbubby: I'm not very good at computer

RedSoxFan:
> Unfortunately a fix hasn't been pushed yet to scdoc master, so we can't do a lot about it.

FYI my patch to fix scdoc was merged into master ~90min ago

emersion:
>FYI my patch to fix scdoc was merged into scdoc master ~90min ago

Oh indeed, I didn't notice because it hasn't been mentioned on the ML. Will send a patch to use `scdoc-git`.

emersion:
Can you rebase this PR?

hugbubby:
@emersion Done.

ianyfan:
You could absorb the condition into the above condition (since it's now a single if block in an if block)

emersion:
>You could absorb the condition into the above condition (since it's now a single if block in an if block)

I'd prefer not to. This is security code, it needs to be readable.

emersion:
Okay, one last thing: can you commit with the same e-mail as your GitHub account? (Or alternatively add the e-mail you use to your GitHub account)

hugbubby:
@emersion Done.

emersion:
Thanks!

PR #3804, opened 2019-03-01
ci: use scdoc-git because 1.9.1 is broken

emersion:
Follow-up issue: revert this patch when scdoc-1.9.2 is released and included in all the platforms we support.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -12,7 +12,6 @@ packages:
   - meson
   - pango-dev
   - pixman-dev
-  - scdoc
   - wayland-dev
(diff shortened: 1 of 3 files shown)


ddevault:
It's probably reasonably to use scdoc-git indefinitely for the same reasons we use wlroots-git.

emersion:
Hmm, apparently scdoc is broken on FreeBSD because `install -D` fails...

>It's probably reasonable to use scdoc-git indefinitely for the same reasons we use wlroots-git.

That's a good question.

ddevault:
Thanks!

PR #3805, opened 2019-03-02
floating_maximum_size: change default behavior

### LINKED ISSUE #3593: Popup windows get buttons cropped out
* Sway Version: 1.0-rc1

* Debug Log:

The only row I get trying to start sway with `sway -d 2`:
```
2019-02-06 18:00:40 - [sway-1.0-rc1/sway/main.c:317] Don't use options with the IPC client
```
But that's not my issue. I don't think, logs will be helpful here.

I use `transmission-remote-gtk` client for `transmission-daemon` torrent client. All popup windows of this application that have height more than some threshold (at least adding torrent, torrent properties and local preferences windows are affected) get their button row cropped out by window border. This bottom part of the window is still displayed on the screen, but it's out of window boundary. I can't click these buttons with mouse, the underlying window gets clicked instead. But I can navigate to these buttons with keyboard Tab key and activate them with enter. I can resize window, enlarging it enough to include these buttons and then they become clickable.

Screenshot of the right bottom corner of such window. You can see the window border that is higher then window buttons.

### LINKED ISSUE #3718: Wrong mpv floating window size and borders
This problem occurs when mpv is configured to run as a floating window.
Similar to the old #680, although version 0.15.2 worked as expected for me.
It is maybe related to #2176 and #3106 

intel graphics
sway version 1.0-rc1-74-g7baaa3a0 (Feb 18 2019, branch 'master')
wlroots version 0.3-fd0b625ab
mpv version 0.29.0-134-g8b563a0346

* Configuration File:
Default configuration with the addition:
`for_window [app_id="mpv"] floating enable`

* Debug Log:
https://gist.github.com/bitraid/86638d2d95c54bdb2048586b18ccd5c3

* mpv Log:
https://gist.github.com/bitraid/206dfb02cc87f32d41900cd7e7d5e02e
>[vo/gpu/wayland] Resizing due to xdg from 1024x768 to 665x499
>[vo/gpu/wayland] Handling resize on the egl side

* Steps to reproduce:
1. Open the terminal.
2. Run: `mpv /usr/local/share/backgrounds/sway/Sway_Wallpaper_Blue_768x1024.png --no-config --image-display-duration=inf -v`

* Expected behavior:
The mpv floating window should have the dimensions of the playing image/video (as it happens with version 0.15.2 and i3wm).

* Actual behavior:
The mpv window has a smaller size and its borders are incorrect. Furthermore, switching to the tiling area after scaling the mpv window to its either half, normal or double size (<kbd>Alt</kbd>+<kbd>0</kbd>, <kbd>Alt</kbd>+<kbd>1</kbd>, <kbd>Alt</kbd>+<kbd>2</kbd>), resets the window size. Manual resize of the window seems buggy, messes up the borders and can't increase after a certain size.

RedSoxFan:
Fixes #3718 
Fixes #3593 

This changes the way zero (which is the default) is interpreted for both
the width and height of `floating_maximum_size`. It now refers to the
width and height of the entire output layout, which matches i3's
behavior.

This also removes duplicated code to calculate the floating constraints
in three files. Before this, `container_init_floating` used two-thirds
of the workspace width/height as the max and the entire workspace
width/height was used everywhere else. Now, all callers use a single
function `floating_calculate_constraints`.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -212,6 +212,9 @@ void container_update_representation(struct sway_container *container);
  */
 size_t container_titlebar_height(void);
 
+void floating_calculate_constraints(int *min_width, int *max_width,
+		int *min_height, int *max_height);
+
 /**
  * Resize and center the container in its workspace.
  */
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -66,45 +66,6 @@ static int parse_resize_amount(int argc, char **argv,
 	return 2;
 }
 
-static void calculate_constraints(int *min_width, int *max_width,
-		int *min_height, int *max_height) {
-	struct sway_container *con = config->handler_context.container;
-
-	if (config->floating_minimum_width == -1) { // no minimum
-		*min_width = 0;
-	} else if (config->floating_minimum_width == 0) { // automatic
-		*min_width = 75;
-	} else {
-		*min_width = config->floating_minimum_width;
-	}
-
-	if (config->floating_minimum_height == -1) { // no minimum
-		*min_height = 0;
-	} else if (config->floating_minimum_height == 0) { // automatic
-		*min_height = 50;
-	} else {
-		*min_height = config->floating_minimum_height;
-	}
-
-	if (config->floating_maximum_width == -1 ||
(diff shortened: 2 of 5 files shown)


emersion:
Thanks!

PR #3809, opened 2019-03-03
render_floating: skip fullscreen floaters

### LINKED ISSUE #3808: Floating windows in between monitors keep borders in fullscreen
A floating window, placed between monitors

Example

and then put into fullscreen, will maintain the border on the monitor it isn't physically attached to.

Example

sway version 1.0-rc1-130-g4431ae68 (Feb 28 2019, branch 'master')

Steps to reproduce:

Float a window (with a border) in between 2 workspaces
fullscreen it

RedSoxFan:
Fixes #3808 

If a floater is fullscreen either on a workspace or globally, it
should not be rendered on any output is is not fullscreened on. When
rendering it on an output it should not be rendered on, there will be
an extraneous border along the adjacent side of the output. This adds
a check in render_floating to skip all fullscreened floaters

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -944,6 +944,9 @@ static void render_floating(struct sway_output *soutput,
 			}
 			for (int k = 0; k < ws->current.floating->length; ++k) {
 				struct sway_container *floater = ws->current.floating->items[k];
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #3810, opened 2019-03-03
cursor: refocus floater before move/resize

### LINKED ISSUE #3537: popup window in gtk and qt apps
sway version 1.0-beta.2-288-g74a4d8e9 
wlroots: commit 77c25c152629e8bd3e8287d86
this no happen in rootston.

gtk3-demo, select gtkentrycompletion
1) open popup (with 't' key)
2) move window, with keybinding or mouse.
background no drawing correct.

It sucess in all popup window that are outside the parent window. 
For example in nautilus preferences in any 'open file' dialog window.

I no get any warning in log, sometimes, get 
`"denying request to set cursor from unfocused client"` from cursor.c

if background is video (example mpv) the window is redraw ok
I read all issues with firefox, intellij, etc. 
I don't where set breakpoint in gdb

RedSoxFan:
Fixes #3537 

This refocuses floaters before moving or resizing them to close any
popups or other surfaces that extend over the bounds of the container.
This prevents artifacts from occurring.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -728,6 +728,8 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 			while (cont->parent) {
 				cont = cont->parent;
 			}
+			// refocus container to close anything that extends over the bounds
+			seat_set_focus_container(seat, NULL);
 			seat_set_focus_container(seat, cont);
 			seat_pointer_notify_button(seat, time_msec, button, state);
 			seatop_begin_move_floating(seat, cont, button);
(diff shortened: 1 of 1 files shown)


sway/input/cursor.c
```diff
@@ -725,7 +725,6 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 			while (cont->parent) {
 				cont = cont->parent;
 			}
-			seat_set_focus_container(seat, cont);
```
emersion: Any bad consequences of doing this?

emersion:
Not sure if it's worth it, but maybe we could add a comment to explain why we reset the focus?

sway/input/cursor.c
RedSoxFan: I just moved it into `seatop_begin_move_floating` so no

RedSoxFan:
> Not sure if it's worth it, but maybe we could add a comment to explain why we reset the focus?

Done

RedSoxFan:
Hold off on merging this for now. I'll rebase it against #3812 and move the focusing into cursor.c before the button press events are sent

RedSoxFan:
Rebased and refactored now that #3812 is merged

emersion:
I wonder whether we should properly damage child surfaces outside of the container's bounds instead.

RedSoxFan:
> I wonder whether we should properly damage child surfaces outside of the container's bounds instead.

I'll see if I can get that working

RedSoxFan:
Closing in favor of #3822, which has been merged

PR #3812, opened 2019-03-04
Allow concurrent clicks

### LINKED ISSUE #3539: Click ignored while holding other mouse button
sway version: 1.0-beta.2-291-g5cb3d476 (Jan 30 2019, branch 'master')

I noticed this while testing Warframe with pointer-constraints and relative-pointer merged (which works great by the way) since you hold right click to aim and then cannot shoot with left click. But this is an issue anywhere even without this merged. Easiest way to test is to hold right click anywhere you can get a dropdown menu and then try to click an option with right click still held down. In other environments (i3 / GNOME Wayland) this would work, whereas in sway it doesn't.

RyanDwyer:
If two cursor buttons are pressed at the same time, the client will now be notified of the second button press.

The main reason for not sending the concurrent presses was due to an early return in `dispatch_cursor_button` if a seatop is in progress. This patch makes it call `seat_pointer_notify_button` prior to returning. But it also has to make sure there's not a mismatch in events such as a release without a press.

Prior to this patch, the `down` seatop would send press and release events in its begin and finish functions. No other seatops did this. A press event would be sent prior to starting tiling drag, but never an associated release.

After this patch, no seatops send their own press or release events. We send them prior to calling the seatop begin functions, then the first part of `dispatch_cursor_button` handles all presses during seatops and when releasing the seatop.

To test:

* Use DRM (you can't test this with the wayland backend, as the DRM sway won't send concurrent clicks to the nested sway)
* Run `weston-eventdemo`
* Press and hold left mouse button
* Press right mouse button - note event in the weston-eventdemo stdout

Fixes #3539.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -613,6 +613,7 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 		} else {
 			state_erase_button(cursor, button);
 		}
+		seat_pointer_notify_button(seat, time_msec, button, state);
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


ddevault:
Thanks!

PR #3816, opened 2019-03-04
meson: update scdoc requirement to >= 1.9.2

RedSoxFan:
Since scdoc 1.9.1 is bugged, this updates the meson version check to
 \>= 1.9.2 and drops the version requirement from the README. This should
make it more obvious to users who have 1.9.1 that they need to update
scdoc to be able to compile man pages and hopefully cut down on the
duplicate issues

### DIFF
--- a/README.es.md
+++ b/README.es.md
@@ -38,7 +38,7 @@ Instale las dependencias:
 * pango
 * cairo
 * gdk-pixbuf2 \*\*
-* [scdoc](https://git.sr.ht/~sircmpwn/scdoc) >= 1.9.0 (optional: man pages) \*
+* [scdoc](https://git.sr.ht/~sircmpwn/scdoc) (optional: man pages) \*
 * git \*
 
 _\*Compile-time dep_
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Install dependencies:
 * pango
 * cairo
 * gdk-pixbuf2 \*\*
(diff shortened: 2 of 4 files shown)


ddevault:
Why not just drop versions from the README and let meson do it?

RedSoxFan:
We started adding the version there to attempt to reduce the bug reports for old versions, which predated the meson version checks. However, meson currently has `>= 1.9.0` and 1.9.1 will not work, so it will not display a user friendly error message. This at least documents it in the readme for now since it's fairly obvious a lot of users aren't searching through the issue tracker before submitting a new one (although there is no guarantee they'll read the readme either...)

ddevault:
Just do == 1.9.3. We have a policy of always being on the latest version of everything.

RedSoxFan:
Since you didn't update the version in the scdoc Makefile, `scdoc >= 1.9.3` will fail. I can change meson to `>= 1.9.2`  and then drop the version in the README though

RedSoxFan:
> I can change meson to >= 1.9.2 and then drop the version in the README though

Changed

ddevault:
>Since you didn't update the version in the scdoc Makefile

Sorry :< I'm going to release .4 to correct this.

RedSoxFan:
> Sorry :< I'm going to release .4 to correct this.

I'd still rather keep this >= 1.9.2 for now until 1.9.4 actually hits repos.

ddevault:
Aight

ddevault:
Thanks!

PR #3822, opened 2019-03-06
output_damage_whole_container: damage subsurfaces 

### LINKED ISSUE #3537: popup window in gtk and qt apps

RedSoxFan:
Fixes #3537 

This adds an iterative call in `output_damage_whole_container` to
damage the subsurfaces for all visible views that are inside of the
container. This is needed to damage subsurfaces that extend outside the
box of the container. Without this, those subsurfaces will create
artifacts when moving or resizing.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -473,6 +473,17 @@ void output_damage_box(struct sway_output *output, struct wlr_box *_box) {
 	wlr_output_damage_add_box(output->damage, &box);
 }
 
+static void damage_child_views_iterator(struct sway_container *con,
+		void *data) {
+	if (!con->view || !view_is_visible(con->view)) {
+		return;
+	}
+	struct sway_output *output = data;
+	bool whole = true;
+	output_view_for_each_surface(output, con->view, damage_surface_iterator,
+			&whole);
+}
+
 void output_damage_whole_container(struct sway_output *output,
 		struct sway_container *con) {
 	// Pad the box by 1px, because the width is a double and might be a fraction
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Thanks for the detailed explanation!

> Damaging view children should probably be done in `desktop_damage_whole_container`.

I ended up putting it in `output_damage_whole_container` instead, which `desktop_damage_whole_container` calls for every output.

RedSoxFan:
I just noticed #1883 is to remove `output_damage_view`. Should I use `output_view_for_each_surface` directly? If so, I could roll #1883 into this PR since `output_damage_from_view` is the only other caller of `output_damage_view`

emersion:
>I just noticed #1883 is to remove output_damage_view. Should I use output_view_for_each_surface directly?

Yeah, why not.

RedSoxFan:
> It depends how far we want to go with this, but: this will work if the container is a view, but it won't work if the container contains multiple views. We'd need to recursively traverse child containers in this case.

Oops. This should be fixed now

> > I just noticed #1883 is to remove output_damage_view. Should I use output_view_for_each_surface directly?
> 
> Yeah, why not.

Changed to using `output_view_for_each_surface` directly. I'll create a follow-up PR for #1883

emersion:
Thanks!

Issue #3821, opened 2019-03-05
Incorrect coordinates in swaymsg get_tree


schauveau:
This is a followup on issue #3785 which was IMHO closed too early.
The issue is that the coordinates given in `rect`, `window_rect` and `deco_rect` are not consistent with i3. For convenience, I will refer to the current window as WIN and to its parent container as PARENT. 

In i3:
   - `WIN.rect` provides the **absolute** window coordinates **excluding** the decoration but including the borders.
   - `WIN.window_rect` is **relative** to `WIN.rect` and provides the coordinates excluding the borders and the decorations.
   - `WIN.deco_rect` is **relative** to `PARENT.rect` and provides the coordinates of the decoration for WIN only.   

In Sway:
  - `WIN.rect` provides the **absolute** window coordinates **INCLUDING THE DECORATIONS** and the borders.
  - `WIN.window_rect` has the same value than in i3 which is useless since that makes it relative to neither `WIN.rect` nor `PARENT.rect`. 
  - `WIN.deco_rect` is **relative** to `WIN.rect`. Also, in stacked or tabbed mode, it provides the combined coordinates of all decorations in the PARENT container instead of the individual decoration as in i3.

schauveau:
I also noticed a few other differences that may affect the future implementation.
- in i3 each floating window is described by 2 nodes (the actual window and its PARENT container) but in sway only one node is used (so no usable PARENT).
- in i3 a window with a border of type `none` or `pixel` has a usable `deco_rect` when its parent layout is `stack` or `tabbed`. That makes sense since those containers are drawing a decoration for each of their elements. In Sway, the `deco_rect` is incorrectly filled with zeros.

RedSoxFan:
> This is a followup on issue #3785 which was IMHO closed too early.
>> Am I right to assume that window_rect is supposed to provide the relative geometry of the window without its decorations?

#3785 was closed because the question quoted above was answered and better documented. Since #3785 was focused on `window_rect`, this is technically a separate issue (`deco_rect` and `rect`) anyway. Sorry, I didn't get to your follow-up response sooner though

As for this issue, #3823 should correct both `rect` and `deco_rect`, but please let me know if you notice a case that I missed.

> in i3 each floating window is described by 2 nodes (the actual window and its PARENT container) but in sway only one node is used (so no usable PARENT).

I think that sway's current response is sufficient (or at least will be after #3823) in regard to this. The parent is the workspace. @ddevault @emersion thoughts?

[...]

PR #3823, opened 2019-03-06
ipc: change {,deco_}rect to match i3

RedSoxFan:
Related to #3821 

This fixes the `deco_rect` and `rect` properties in the IPC responses
to match i3's behavior.

`deco_rect` should be relative to the parent node, not the current
node. This also takes tabbed and stacked decorations into account and
will calculate `deco_rect` for all containers since tabbed and stacked
child containers will have decorations.

`rect` should exclude the window decorations.


emersion:
Thanks!

PR #3824, opened 2019-03-06
Don't send button events to surfaces when dragging or resizing

### LINKED ISSUE #3819: Mouse-click events get sent to a program while resizing
* Sway Version:
sway version 1.0-rc1-144-ga6711740 (Mar  5 2019, branch 'master')

* Debug Log:
https://0x0.st/zo24.log (nautilus-wayland.log)
https://0x0.st/zo2J.log (sway.log)

* Configuration File:
https://0x0.st/zipQ.txt (config)

* Steps to reproduce:
1. Start Nautilus, set it to floating mode.
2. When resizing notice that the mouse events are sent to nautilus. In example the context menu of nautilus appears when resizing started when the cursor was positioned over a file.

* Expected Behaviour:
No context menu should appear and no mouse click events should be send to to nautilus while resizing.

RyanDwyer:
It turns out sending button events during all seat operations is not desirable. This patch introduces a new property `seatop_impl.allows_events` which allows each operation to define whether button events should be passed to the surface or not.

The `down` seat operation is the only one that supports this. As all the other seatops don't support it, the calls to `seat_pointer_notify_button` prior to starting them have been removed.

Fixes #3819.

@SwayUser Can you please test?


emersion:
Looks good, will approve once we hear back from @SwayUser.

SwayUser:
This works, but it introduces a new bug for me. The bug is that the menubar in Nautilus and other programs doesn't work anymore. (The button in the image doesn't work anymore, for example.)
Image

RyanDwyer:
Fixed. From my commit message:

    By the time seatop_allows_events was called, seatop_impl was already
    NULL, causing the function to always return false. This means a press
    event was sent to clients without a corresponding release event.
    
    This patch moves the call to seatop_finish to after the
    seatop_allows_events check.

emersion:
@SwayUser Would be nice if you could confirm this PR fixed both issues.

SwayUser:
Works for me now, but issues are fixed.

emersion:
Thanks!

PR #3827, opened 2019-03-06
ipc: fix fullscreen deco_rect

RedSoxFan:
Related to #3821 

This fixes the deco_rect reported by the ipc for fullscreen containers
to be all zeroes. Children of the fullscreen container should still
have their decorations reported correctly


schauveau:
I confirm that rect, deco_rect and window_rect are not working as expected for fullscreen windows. Thanks

PR #3828, opened 2019-03-06
ipc: fix rect for stacked children

RedSoxFan:
Related to #3821 

This now takes all titlebars for stacked children into account for the
ipc property `rect`


ddevault:
Thanks!

schauveau:
I confirm that the rect, deco_rect and window_rect are now working as expected in stacked containers. Thanks

PR #3836, opened 2019-03-08
damage: remove output_damage_view

### LINKED ISSUE #1883: Remove output_damage_view(whole=true)

RedSoxFan:
Closes #1883
Related #3822 

This removes `output_damage_view` since it is unnecessary. The logic
has been moved into its only caller `output_damage_from_view`. When
damaging the whole view, `output_damage_whole_container` should be used
instead


emersion:
Thanks!

PR #3837, opened 2019-03-08
Add -p/--pretty option to swaymsg

jpgrayson:
This new option forces pretty (non-raw/non-JSON) output. By default, when
not using a tty, swaymsg outputs using the "raw" format. This makes it
impossible to, for example, pipe the pretty output to a pager such as
`less` since piping does not use a tty.

The new -p/--pretty option gives the user explicit control over the output
format while retaining the default tty-dependent behavior.

### DIFF
--- a/completions/bash/swaymsg
+++ b/completions/bash/swaymsg
@@ -21,6 +21,7 @@ _swaymsg()
 
   short=(
     -h
+    -p
     -q
     -r
     -s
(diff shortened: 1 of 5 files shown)


jpgrayson:
> Can you please change the ordering to retain alphabetical order in all locations

Done. And in the process of putting the new option in lexical order, I noticed that the `-m/--monitor` option is missing from the fish and bash completions. So a win for keeping things in order.

I'll submit a separate PR to correct those missing completions.

RedSoxFan:
Thanks!

PR #3838, opened 2019-03-08
Add missing swaymsg completions

jpgrayson:
The `-m/--monitor` option was missing from the bash and fish completions.

The `subscribe` IPC message type was missing from the bash, fish, and zsh
completions.


RedSoxFan:
Thanks!

PR #3839, opened 2019-03-08
New get_bindings IPC

jpgrayson:
This adds a new IPC interface to sway for interrogating the configured mouse and keyboard bindings.

My goal is to enable external tools to be able to display keybindings. I've added pretty-print for these bindings to `swaymsg`, so that's the first tool. But I'm also envisioning enabling a graphical application to be written that displays the keybindings in a nicely formatted fashion. Such an application bound to `$mod+?` or `$mod+F1` could be useful for easing the sway learning curve. As a relatively long-time i3 and sway user, I still need help remembering less frequently used bindings and consulting my config file is a little to slow/noisy when all I want is a quick reminder.

I expect that the sentiment is to be conservative with adding new IPC interfaces, so I will understand if this one is not accepted. However, I would argue that this one does not seem too exotic compared to sway's other "`get`" IPC interfaces.

Lastly, if this `get_bindings` IPC is accepted, it may be worth considering eliminating the `get_binding_modes` interface since `get_bindings` provides both the binding mode names as well as the bindings associated with those modes--i.e. it's a superset of the information provided by `get_binding_modes`.


RedSoxFan:
I think it would be better to extend `get_binding_modes` to be similar to `get_bar_config` where it lists the modes without a payload and gives mode specific bindings when the mode is given as the payload.

Either way, `get_binding_modes` is not going to be removed since it is there for i3 ipc compatibility.

Also, any IPC changes/additions should be documented in `sway-ipc.7.scd`

RedSoxFan:
This also requires investigation on whether this has been discussed by i3 and if so, what their comments/stance is/was.

If it hasn't been discussed by i3, bringing it there could be a good idea before proceeding

ddevault:
Aye, going to push back on this until it's been discussed with i3.

jpgrayson:
This is the i3 enhancement request for the existing `get_binding_modes`:

https://github.com/i3/i3/issues/2375

And I found this i3 issue from two years ago that seeks the same feature for sway:

https://github.com/i3/i3/issues/2691

The requestor (@deathlyfrantic) wanted a quick way to see the configured bindings without having to reference the config file. But AFAICT, there was no patches on the table or even a concrete plan to use the new IPC, so the conversation and issue were closed. My interpretation is that i3 might accept this if a more compelling/concrete case was made.

I like @RedSoxFan's idea of extending `get_binding_modes` to return a bindings list when provided a mode argument. I will make that change.

Perhaps the path forward here will be for me to also compose a patch to add this feature to i3 and use that to drive the discussion. Thoughts?

RedSoxFan:
> My interpretation is that i3 might accept this if a more compelling/concrete case was made.
> Perhaps the path forward here will be for me to also compose a patch to add this feature to i3 and use that to drive the discussion. Thoughts?

After reading through the issues you linked, I'm not sure if i3 will accept it, but it seems like a reasonable step. Personally, I'm not opposed to adding this to Sway even if i3 rejects it though

ddevault:
I just finished excising sway's extensions to i3 IPC, let's pump the brakes on adding more :)

RedSoxFan:
> I just finished excising sway's extensions to i3 IPC, let's pump the brakes on adding more :)

I think extensions are fine within reason. For this particular case, I can definitely see a layer shell overlay listing bindings being useful for people who are new to sway (or just want to have it for less commonly used bindings).

ddevault:
Right, but bindings are an inerited feature from i3, and feature development for that is in their domain.

ddevault:
Will consider this feature once i3 has something similar.

PR #3842, opened 2019-03-09
fix "directive argument is null" errors

jpeeler:
In sway/criteria.c - value was the offender. Moved the logging call up to where value is not guaranteed to e NULL.

In sway/input/seat.c - until cursor_theme is configurable, it shouldn't be passed to sway_assert. It's already marked with a TODO, so I'm sure the logging can be changed back later.

Full disclosure: I have not (yet) tested these changes.

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -624,8 +624,8 @@ struct criteria *criteria_parse(char *raw, char **error_arg) {
 				in_quotes = false;
 			}
 			unescape(value);
+			sway_log(SWAY_DEBUG, "Found pair: %s=%s", name, value);
 		}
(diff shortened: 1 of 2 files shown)


ddevault:
>Full disclosure: I have not (yet) tested these changes.

Let me know when this changes

jpeeler:
I tested the criteria code. I'm not sure how to check the seat code since it's only hit when there's a memory allocation failure. Because the change here is removing a parameter, it should be fine, right?

This is the first time I've used sway. I'm loving it so far! I'm especially relieved that reloading configuration doesn't cause window state to be lost.

ddevault:
Thanks!

Issue #3333, opened 2018-12-26
Windows not rendered when using negative y output position


IMFTC:
Adding the lines
```
output LVDS-1 resolution 1600x900  position 0,0
output VGA-1  resolution 1920x1080 position 1600,-400
```
to the default config file results in windows not being rendered when using tabbed or stacked layout, no matter if it's a single window or a container. The tabs in the tabbed layout and all title bars for the stacked layout are rendered correctly, but the space below only shows the background image.  Switching to split layout makes the window(s) visible.

Using the alternative configuration
```
output LVDS-1 resolution 1600x900  position 0,400 
output VGA-1  resolution 1920x1080 position 1600,0
```
everything works correctly.

sway version 9bf1b7a029afe2fdb07cb2bfe35cca4ad76098f9
wlroots version ed5296c6cf23efa07f78e1657cec1528f638f95e

jeffa5:
I've had the same issue with this but also Firefox and Qutebrowser (so potentially X applications?) become unresponsive to mouse interactions when in tiling layout. They still accept key presses though. Moving the applications back to a display without negative positions makes them work again until moved to a display with negative pos.

emersion:
>I've had the same issue with this but also Firefox and Qutebrowser (so potentially X applications?) become unresponsive to mouse interactions when in tiling layout. They still accept key presses though. Moving the applications back to a display without negative positions makes them work again until moved to a display with negative pos.

This is an Xwayland bug.

edrex:
When this issue is triggered on my setup (primary display scale 2, negative y on scale 1 display) the tab text is rendered 2x normal (like the tabs think they are on the scale 2 display). 

![negative-output-position-corruption](https://user-images.githubusercontent.com/14615/51149434-f9c07e80-1816-11e9-9a8a-b25b0cac658a.png)

![negative-output-position-corruption-stacked](https://user-images.githubusercontent.com/14615/51149352-99314180-1816-11e9-91c3-5e298d0d30b9.png)

It could be a separate issue but I'm guessing they have the same root cause.

ThomasLeister:
I'm having similar issues with negative display positions. if I use this configuration:

```
output eDP-1 {
        pos 0 0
}

output DP-4 {
	scale 1.2
        pos 1 -1800
        res 3840x2160
}
```

[...]

PR #3848, opened 2019-03-10
arrange: use int not size_t for title offsets

### LINKED ISSUE #3845: Setting output Y-coordinate to negative value will make stacked and tabbed windows transparent
Picture with negative Y-coordinate.

Setting output Y-coordinate to negative value will make stacked and tabbed windows transparent. I haven't found a consistent value to produce this. Setting the Y-coordinate to -30 with default config seems to reliable produce this but with my personal config-file any negative value will work. X-coordinate seems to be unaffected.

[Picture with default values.](https://i.postimg.cc/bpCtHqN3/default.png)

[Picture with negative Y-coordinate.](https://i.postimg.cc/7xbXMBkT/negativeposition.png)

`sway version 1.0-rc1-152-g5fab8a2a (Mar  9 2019, branch 'master')`

```
[testuser@Fedora-X220 ~]$ diff .config/sway/config /usr/local/etc/sway/config 
29c29
< output LVDS-1 resolution 1366x768 position 0,-30
---
> #   output HDMI-A-1 resolution 1920x1080 position 1920,0
```

[Swaylog](https://pastebin.com/i1xYqBkg)

RedSoxFan:
Fixes #3845 
Related to #3333

This changes `apply_tabbed_layout` and `apply_stacked_layout` to use
`int` instead of `size_t`. This is necessary for tabbed and stacked
containers to be positioned correctly when the y-location is negative.
The reasoning for this is signed plus unsigned is always an unsigned
value. This was causing the y-location of the container to be
positioned near `INT_MIN` due to an unsigned integer underflow

_`ssize_t` will also work, but I feel like `int` is better since `ssize_t`
makes it seem like the titlebar height can be negative_

### DIFF
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -99,7 +99,7 @@ static void apply_tabbed_layout(list_t *children, struct wlr_box *parent) {
 	}
 	for (int i = 0; i < children->length; ++i) {
 		struct sway_container *child = children->items[i];
-		size_t parent_offset = child->view ? 0 : container_titlebar_height();
+		int parent_offset = child->view ? 0 : container_titlebar_height();
 		container_remove_gaps(child);
 		child->x = parent->x;
 		child->y = parent->y + parent_offset;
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #3849, opened 2019-03-10
sway_view_child: add listener for view unmap

### LINKED ISSUE #3835: Firefox self update crashes sway (different backtrace)
Split from issue #3263
This one also happened when Firefox nightly running in wayland mode was performing self update.

sway version 1.0-rc1-148-g0df76ed9 (Mar  7 2019, branch 'makepkg')
wlroots: 0.4.1.r26.g2baad6eb-1

Debug log last few minutes before the crash: [sway.crash.log](https://github.com/swaywm/sway/files/2945147/sway.crash.log)
Full backtrace: [btfull.txt](https://github.com/swaywm/sway/files/2945153/btfull.txt)

This time child->view->container was invalid pointer, but not null.
```
(gdb) print *child->view
$5 = {type = (unknown: 2944062624), impl = 0x55bbaf7f5c80, container = 0x4000000c0, surface = 0x100000002, xdg_decoration = 0x37000000002, pid = 1, saved_x = 1.8673562960504278e-311, 
  saved_y = 2.121995791459338e-314, saved_width = 1, saved_height = 1, natural_width = 6744320, natural_height = 0, title_format = 0x1df000000004 <error: Cannot access memory at address 0x1df000000004>, 
  using_csd = 112, urgent = {tv_sec = 9, tv_nsec = 94264591767072}, allow_request_urgent = 225, urgent_timer = 0x0, saved_buffer = 0x400000000, saved_buffer_width = 0, saved_buffer_height = 0, geometry = {
    x = 0, y = 0, width = 0, height = 0}, saved_geometry = {x = -1350634880, y = 21947, width = 0, height = 0}, destroying = false, executed_criteria = 0x0, {wlr_xdg_surface_v6 = 0x0, wlr_xdg_surface = 0x0, 
    wlr_xwayland_surface = 0x0, wlr_wl_shell_surface = 0x0}, events = {unmap = {listener_list = {prev = 0x0, next = 0x0}}}, surface_new_subsurface = {link = {prev = 0x0, next = 0x0}, notify = 0x0}}
(gdb) print *child->view->container 
Cannot access memory at address 0x4000000c0
```

RedSoxFan:
I think this should fix #3835, but I'm not sure it is the best solution
or how to get firefox to do self-updates to test it.

Since not all child views's have an unmap event, it is possible for it
to still be mapped (default state) in the destruction handler. When
the destruction handler is called, the corresponding view may have
already been freed and the memory location reallocated. This adds a
listener for the view unmapping and removes the mapped status. This
ensures that the child view is damaged due to destruction while the
view still exists and not after.

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -202,6 +202,7 @@ struct sway_view_child {
 	struct wl_listener surface_map;
 	struct wl_listener surface_unmap;
 	struct wl_listener surface_destroy;
+	struct wl_listener view_unmap;
 };
 
 struct sway_subsurface {
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -820,6 +820,14 @@ static void view_child_handle_surface_unmap(struct wl_listener *listener,
 	child->mapped = false;
 }
 
+static void view_child_handle_view_unmap(struct wl_listener *listener,
+		void *data) {
+	struct sway_view_child *child =
+		wl_container_of(listener, child, view_unmap);
+	view_child_damage(child, true);
+	child->mapped = false;
+}
+
 void view_child_init(struct sway_view_child *child,
 		const struct sway_view_child_impl *impl, struct sway_view *view,
 		struct wlr_surface *surface) {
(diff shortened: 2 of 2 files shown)


emersion:
@cg9999 Can you try this?

cg9999:
I ran 15 self update cycles without issues. As this crash is hard to reproduce, it's not a proof it's completely fixed, but it's looking good and I don't see any regressions.

emersion:
Thanks!

Issue #3795, opened 2019-02-28
sway crashing on running slurp and then moving workspace before selecting an area


biwhite:
sway version 1.0-rc1-118-g9032be4f (Feb 27 2019, branch 'master')

Sway debug log
https://gist.github.com/biwhite/c4a3bd2175036a5ea7e985806066abd0

Backtrace
https://gist.github.com/biwhite/30c70405f417ab302bde25f805bb8a51

Finally, explain the steps you took in plain English to reproduce the problem below.

Crashes when opening sway, running slurp, and using a workspace prev_on_output, either through key bindings or swaymsg.

Emantor:
Investigating the code, the fix is going to be in `seat_get_focused_workspace` by also trying `seat_get_focus_inactive`. However a seat not having a focus or inactive focus behaviour is still valid (e.g. shortly after creation) and should imo be handled by all callers. In this case I think its also fine to abort in the case of the above command.
The other call sites of `seat_get_focused_workspace` need investigation as well.

emersion:
>the fix is going to be in `seat_get_focused_workspace` by also trying `seat_get_focus_inactive`

We should probably only do this if a layer surface is focused?

PR #3646, opened 2019-02-11
workspace_get_initial_output: handle focused layer

RedSoxFan:
Fixes #3616 

When a layer surface is focused, `seat_get_focused_workspace` will be
NULL. This changes `workspace_get_initial_output` to use output of the
focus inactive. If the focus inactive is also NULL, then either the
first output or the noop output will be used as fallbacks.

PR #3852, opened 2019-03-10
Fix crash in cmd_workspace when layer surface has focus

emersion:
This changes `seat_get_focused_workspace` to use `seat_get_focus_inactive` instead of `seat_get_focus`. This may break things, but I can't think of a situation in which it would.

We still need to check whether `seat_get_focused_workspace` is NULL.

Follow-up issue: check all other `seat_get_focused_workspace` calls.

Test plan: run `sleep 5; swaymsg workspace prev_on_output`, then run `slurp` in another terminal

Fixes https://github.com/swaywm/sway/issues/3795

### DIFF
--- a/sway/commands/workspace.c
+++ b/sway/commands/workspace.c
@@ -184,6 +184,11 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 		bool create = argc > 1 && strcasecmp(argv[1], "--create") == 0;
 		struct sway_seat *seat = config->handler_context.seat;
 		struct sway_workspace *current = seat_get_focused_workspace(seat);
+		if (!current) {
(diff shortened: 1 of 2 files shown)


ddevault:
Thanks!

emersion:
@RedSoxFan Still interested in your opinion on this:

>This changes seat_get_focused_workspace to use seat_get_focus_inactive instead of seat_get_focus. This may break things, but I can't think of a situation in which it would.

RedSoxFan:
>  This changes seat_get_focused_workspace to use  seat_get_focus_inactive instead of seat_get_focus. This may break  things, but I can't think of a situation in which it would.

I think it's fine and a few locations (such as #3646 -- I can search for all of them if needed) have already been changed to not use `seat_get_focused_workspace` due to similar issues.

Issue #2176, opened 2018-06-30
First xdg-shell configure event is invalid


emersion:
When a client creates a new xdg-surface and commits a NULL buffer, we're supposed to send a configure event with the size of the window. We currently send it with a 0x0 size (which lets the client decides). We only send the first "real" configure event when the client maps (ie. after it has drawn its first frame).

We should send the correct configure event from the beginning.

RyanDwyer:
My understanding is that the view can create itself but choose not to map immediately. During this delay the user might focus another container, in which case the size it needs could be different. So if we did try to preempt the size when it's created then we'd have to arrange it again when it sends the map request. The positive would be that when it maps, the surface will probably already be the correct size so the transaction would have one less view to wait on. This isn't a huge positive considering it still has to wait for the siblings to resize.

It would be fairly complicated to determine what size the view should be without actually arranging its siblings. You'd need a function similar to `arrange_children_of`. And to further complicate it, you don't know what size it should be until you know whether it'll have borders, will be placed on a different workspace due to criteria, or will be floating.

In my opinion the negatives greatly outweigh the positives. It's a lot of complicated code for something that the user is never going to see.

It sounds to me like the recommendation is intended for floating window managers.

emersion:
>My understanding is that the view can create itself but choose not to map immediately.

I don't think so.

RyanDwyer:
It must. After we send the configure, the client takes time to create the buffer, then commits or requests to be mapped when it's done. This could take 0ms, or it could take 10 seconds. It would be stupid for the compositor to block the event queue while waiting for the client to map, which means the user might change focus between these operations.

emersion:
>It must. After we send the configure, the client takes time to create the buffer, then commits or requests to be mapped when it's done.

It's no different than a normal resize operation. If we block when resizing, we can block when mapping.

SirCmpwn:
We can't let clients slowloris the compositor into a standstill.

SirCmpwn:
Also, letting the client choose its initial size itself lets us stow that away and restore it when we float the client later.

emersion:
>We can't let clients slowloris the compositor into a standstill.

I don't understand.

>Also, letting the client choose its initial size itself lets us stow that away and restore it when we float the client later.

The official way to do this is to send a 0x0 size.

emersion:
[...]

Issue #3029, opened 2018-10-29
Surface contents can be rendered outside the borders after a resize


c-edw:
QEMU is underlapping other windows. See below for example.

![out](https://user-images.githubusercontent.com/38229097/47685694-46f56280-dbcf-11e8-841c-6491bdffbd0b.png)

Version:
```
sway version 1.0-beta.1-108-gb90af335 (Oct 29 2018, branch 'master')
```

Steps to produce:
1. Start Sway
2. Launch two terminals and QEMU (supply a cdrom and wait for it to begin booting for the window to underlap)
3. Move one terminal to the left of QEMU and one to the right
4. Observe that QEMU underlaps the rightmost terminal, though this is only visible if your terminal is configured to be transparent

This issue might not be with Sway or wlroots, so please confirm.

SirCmpwn:
Is your terminal transparent?

c-edw:
Yes, it is.

emersion:
>Is your terminal transparent?

In the description:

>though this is only visible if your terminal is configured to be transparent

Maybe we should scissor tiled views to their container's bound (excluding popups).

SirCmpwn:
>In the description

![](https://sr.ht/8HT-.png)

:wink: 

>Maybe we should scissor tiled views to their container's bound (excluding popups).

Maybe. In my opinion this is working as designed.

RyanDwyer:
A similar thing happens with gnome-calculator when you have swaybar at the bottom and gnome-calculator doesn't have enough space to fit its minimum surface height. It overlaps swaybar. For example, create layout `V[view view gnome-calculator]`.

c-edw:
@RedSoxFan FYI this is still reproducible after your patch.

![020734-16-03-2019](https://user-images.githubusercontent.com/38229097/54469470-55b07080-4790-11e9-91d7-e11d7c4059cf.png)

RedSoxFan:
> @RedSoxFan FYI this is still reproducible after your patch.

I only touched floating clients. Nothing has been changed regarding tiling ones, which is why this remains opens.

pedrocr:
I can reproduce a similar issue by setting outer gaps and then opening several gnome-terminal windows. At some point gnome-terminal stops reducing in size and so the last window becomes out of bounds:

![gnometerminal_outofbounds](https://user-images.githubusercontent.com/246322/60760164-e1dcf100-a028-11e9-8902-81ff167641d1.png)

[...]

PR #3854, opened 2019-03-10
fullscreen: init floating on disable without size

### LINKED ISSUE #3850: Floating window created as fullscreen has zero size
When creating a fullscreen window and having a rule to make it floating, the window will have zero width/height when de-fullscreening it.

May be related to #3718.
Running c02d2a01289581fd3c32a22e1a498ffad736281d

Reproducible with config:
```
for_window [app_id="mpv"] floating enable
for_window [class="mpv"] floating enable
for_window [title="^:float:"] floating enable
```
And commands:
`mpv /usr/local/share/backgrounds/sway/Sway_Wallpaper_Blue_768x1024.png --no-config --image-display-duration=inf -v --fs=yes`
or
`xfce4-terminal --fullscreen --title=:float:test`

RedSoxFan:
Fixes #3850 

If a container gets mapped as fullscreen and set to floating by
criteria, the size and location are never set for the floating
container. This adds a check in container_fullscreen_disable for a
width or height of 0 and calls container_init_floating

_Note that the content may be smaller than the container due to the
natural size being larger than the maximum floating size and the
client giving a smaller surface of the same aspect ratio. I think
this is related to #2176 (or at least the discussion in the issue)_

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -984,6 +984,13 @@ void container_fullscreen_disable(struct sway_container *con) {
 		root->fullscreen_global = NULL;
 	}
 
+	// If the container was mapped as fullscreen and set as floating by
+	// criteria, it needs to be reinitialized as floating to get the proper
+	// size and location
+	if (container_is_floating(con) && (con->width == 0 || con->height == 0)) {
+		container_init_floating(con);
+	}
+
 	con->fullscreen_mode = FULLSCREEN_NONE;
 	container_end_mouse_operation(con);
 	ipc_event_window(con, "fullscreen_mode");


ddevault:
@minus7 does this fix your issue?

minus7:
The window size isn't zero anymore, but until I move the window, mpv's content is larger than the window: screenshot-2019-03-10T21:29:31Z

Xfce4-terminal plays nice with it tough

RedSoxFan:
@minus7 can you try again?

minus7:
No change

RedSoxFan:
@minus7 how about now?

sway/desktop/transaction.c
```diff
@@ -395,6 +395,10 @@ static bool should_configure(struct sway_node *node,
 	// Xwayland views are position-aware and need to be reconfigured
 	// when their position changes.
 	if (node->sway_container->view->type == SWAY_VIEW_XWAYLAND) {
+		if (cstate->fullscreen_mode != FULLSCREEN_NONE
```
emersion: Doesn't `content_x` change anyway in this case?

RedSoxFan: Ya, I think I just got confused while testing something

RedSoxFan:
> The window size isn't zero anymore, but until I move the window, mpv's content is larger than the window: 

On IRC, it was determined that this only happens when `--force-window=immeadiate` (and running as a xwayland client), which forces mpv to map at 640x480. After some investigation, I'm pretty sure this falls under the same core issue as #3029 so I'd like to tackle that as a separate PR

emersion:
Thanks!

PR #3856, opened 2019-03-10
meson: use pkg-config var for scdoc path

emersion:
Also make sure to specify `native: true` to use the host binaries in case of a cross-build.


RedSoxFan:
Thanks!

PR #3858, opened 2019-03-11
detect_proprietary: use strncmp

### LINKED ISSUE #1594: Warning in detect_proprietary()
I am using nouveau, but code in detect_proprietary detects nvidiafb as nvidia driver. And prints Warning about it. But nvidiafb is only: "Framebuffer driver for nVidia graphics chipset".
the check for strstr("nvidiafb", "nvidia") returns true seomehow, so it has to be fixed

RedSoxFan:
Fixes #1594

Only the main nvidia module needs to be blocked. Others such as
nvidiafb are benign and do not need to be blocked


ddevault:
Thanks!

PR #3860, opened 2019-03-11
Stop using wlr_output->{lx,ly}

emersion:
This is an update for https://github.com/swaywm/wlroots/pull/1611. It doesn't depend on it though, and can be merged before.

Also fixes sway_output->{lx,ly,width,height} not being updated. Also fixes
output_get_in_direction adding buffer coords to layout coords.


RedSoxFan:
Thanks!

Issue #3862, opened 2019-03-11
Sway does not compile


johnae:
After this https://github.com/swaywm/sway/commit/5c8424c0743c36b11baabe8b316ee6daecd0cfb8#diff-9ae010cb99f0a17a044ce6772d0e7ebc sway does not compile, compilation fails with error:

```
In function 'log_loaded_themes.isra.1',
    inlined from 'init_themes' at ../swaybar/tray/icon.c:351:2:
../swaybar/tray/icon.c:319:8: error: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Werror=alloc-size-larger-than=]
  char *str = malloc(len);
```

emersion:
Damn, that's a pretty bad bug. Working on a fix.

ianyfan:
Hmm, this seems to imply that `len == -1` (which becomes enormous due to being unsigned) but that can only ever happen if `themes->length == 0` which is checked for at the beginning of the function.

Maybe it is an optimisation issue?

A fix could be to make `len` a `ssize_t`.

emersion:
Can you try https://github.com/swaywm/sway/pull/3864?

PR #3864, opened 2019-03-11
Fix size_t temporary underflow in log_loaded_themes

emersion:
`len` will underflow but will overflow right after, so it's not as bad as it
may appear. Still better not to under/overflow at all.

Fixes https://github.com/swaywm/sway/issues/3862


emersion:
>but that's better than an underflow

Yeah, this is deliberate. I'd prefer simplicity and readability over saving 2 bytes.

RedSoxFan:
Thanks!

PR #3866, opened 2019-03-11
commands: allow tiled sticky containers to be moved to a workspace on…

ianyfan:
… the same output

But not tiled sticky children of floating containers


ianyfan:
@RedSoxFan I'll rewrite it in a bit. Do you have a suggestion?

RedSoxFan:
> @RedSoxFan I'll rewrite it in a bit. Do you have a suggestion?

`cmd_move: allow tiled sticky containers to move` or at least something shorter than the current one. A short commit message should preferably be less than 50 characters (or at least around there). The current one is just too long

RedSoxFan:
Thanks!

PR #3875, opened 2019-03-12
get_deco_rect: fix floaters on tabbed/stacked ws

### LINKED ISSUE #3870: sway chrashes in get_deco_rect()
I am using sway with the commit 783b3d6b378db0bdf2a10bfa8a6787bd4b152a5a from the master branch. With the default configuration sway crashes if I do the following after sway has started:

1.  Start a terminal (in my case termite)
2. Execute the following in this terminal
```
swaymsg layout tabbed, floating enable, border pixel, move absolute 0 0, workspace 2
```
Here is the [debug log](https://gist.github.com/mwenzkowski/5ac497937d6f39d1b6558fd947b8a4f9#file-sway-log) and the [stack trace](https://gist.github.com/mwenzkowski/5ac497937d6f39d1b6558fd947b8a4f9#file-stack-trace).

RedSoxFan:
Fixes #3870 

This fixes the decoration rects for floating containers on a workspace
that is either tabbed or stacked. Without this, the floater would
incorrectly try to calculate where it's tab or stack decorations were
on the workspace. This would cause a SIGFPE (due to a divide-by-zero)
when the floater was on a tabbed workspace without any tiling children.
Furthermore, the floater does not care what the workspace's layout is
and should just use the location relative to the workspace. This should
have no effect on children of a floating container.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -352,8 +352,9 @@ static void ipc_json_describe_workspace(struct sway_workspace *workspace,
 
 static void get_deco_rect(struct sway_container *c, struct wlr_box *deco_rect) {
 	enum sway_container_layout parent_layout = container_parent_layout(c);
-	if ((parent_layout != L_TABBED && parent_layout != L_STACKED &&
-			c->current.border != B_NORMAL) ||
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #3879, opened 2019-03-12
config.in: allow launch apps with args via dmenu

3ap:
Without this change you can't execute apps with command line arguments (e.g. firefox -P profile) because -P will be parsed as argument for "swaymsg exec".


ddevault:
Thanks!

PR #3882, opened 2019-03-12
Handle seat_get_focused_workspace returning NULL

### LINKED ISSUE #3853: seat_get_focused_workspace callers should check for NULL
This can happen if called after the output is created but before the workspace is created.

RedSoxFan:
Closes #3853 

This modifiers the callers of seat_get_focused_workspace to handle
getting NULL as the return value, if they did not already.


emersion:
Thanks!

PR #3887, opened 2019-03-13
swaybar: fix loading of malformed icon theme

stev47:
If the icon index.theme contained a key-value pair without a preceding
group header, entry_handler() would be called with a zero pointer and
lead to a segfault.
Set the error flag and break on such malformed files.


ddevault:
Thanks!

Issue #3784, opened 2019-02-27
Input configuration by input type


kennylevinsen:
# Problem

Certain parameters (e.g. "natural_scroll"), is intended and/or only used with certain types of devices (e.g. touchpads).

At the current time, input configuration can only be applied to a single device, or with wildcard to all devices. For such parameters, the user will then either have to individually add all inputs that needs the type-specific configuration, or live with a wildcard that also applies to other, unwanted device types (e.g. inverting scroll on regular scroll wheels).

# Potential solution

The ability to apply input configuration on device type would allow for users to make sensible default configurations for various device types (e.g. natural scroll on all touchpads), without interference between different device types (e.g. scroll inversion applying to both touchpads and mouse). Such functionality would ease shared configuration setups, and would allow reconfiguration-free use of new devices.

# Example configuration

```
input type:touchpad {
   natural_scroll enabled
}
```

# Potential problems

sway-input currently recognizes touchpads as "mouse", which is not granular enough. However, considering that GNOME can distinguish between such devices using libinput (i.e. it separates touchpad and mouse configuration), I suppose that the information should be available from libinput.

RedSoxFan:
> sway-input currently recognizes touchpads as "mouse", which is not granular enough. However, considering that GNOME can distinguish between such devices using libinput (i.e. it separates touchpad and mouse configuration), I suppose that the information should be available from libinput.

Looks like GNOME checks whether the output of `libinput_device_config_tap_get_finger_count` is greater than zero for the device (https://gitlab.gnome.org/GNOME/mutter/blob/master/clutter/clutter/evdev/clutter-input-device-evdev.c#L1459)

lolzballs:
I'm looking into migrating to sway and I was looking for a way to do this type of configuration. This seems like something I can tackle in the next couple of days if no one else has started working on it.

RedSoxFan:
Closing via #3895

PR #3895, opened 2019-03-14
Add support for configurations per input device type

lolzballs:
This PR implements #3784.

The "touchpad-iness" of a pointer is determined through `libinput_device_config_tap_get_finger_count`, which is referenced from [clutter](https://gitlab.gnome.org/GNOME/mutter/blob/master/clutter/clutter/evdev/clutter-input-device-evdev.c#L1459). In the future, this may be updated to a more concrete method of determining whether an input device is a touchpad.

`swaymsg` is updated to reflect this change, outputting `Type: Touchpad` or `Type: Mouse` rather than only `Type: Pointer`. In addition, `sway` now logs the matching of a type wildcard.

The behaviour is as follows:
- If a device is added and has a configuration, it is applied on top of it's type config.
- If a device is added, does not have a configuration, and it has a matching type config, the device will use the type config.
- If a device is already existing when a type config is applied, the type config will be merged on top of the existing device config.
- As expected, wildcard configs apply on top both type configs and device-specific configs.

### DIFF
--- a/sway/config/input.c
+++ b/sway/config/input.c
@@ -18,6 +18,7 @@ struct input_config *new_input_config(const char* identifier) {
 		return NULL;
 	}
 
+	input->input_type = NULL;
 	input->tap = INT_MIN;
 	input->tap_button_map = INT_MIN;
 	input->drag = INT_MIN;
@@ -140,6 +141,30 @@ static void merge_wildcard_on_all(struct input_config *wildcard) {
 			merge_input_config(ic, wildcard);
 		}
 	}
+
+	for (int i = 0; i < config->input_type_configs->length; i++) {
+		struct input_config *ic = config->input_type_configs->items[i];
+		if (strcmp(wildcard->identifier, ic->identifier) != 0) {
+			sway_log(SWAY_DEBUG, "Merging input * config on %s", ic->identifier);
+			merge_input_config(ic, wildcard);
+		}
+	}
+}
+
+static void merge_type_on_existing(struct input_config *type_wildcard) {
+	for (int i = 0; i < config->input_configs->length; i++) {
+		struct input_config *ic = config->input_configs->items[i];
+		if (ic->input_type == NULL) {
+			continue;
+		}
+
+		if (strcmp(ic->input_type, type_wildcard->identifier + 5) == 0) {
+			sway_log(SWAY_DEBUG, "Merging %s top of %s",
+				type_wildcard->identifier,
+				ic->identifier);
+			merge_input_config(ic, type_wildcard);
+		}
+	}
 }
 
 struct input_config *store_input_config(struct input_config *ic) {
@@ -148,11 +173,19 @@ struct input_config *store_input_config(struct input_config *ic) {
 		merge_wildcard_on_all(ic);
 	}
 
-	int i = list_seq_find(config->input_configs, input_identifier_cmp,
+	list_t *config_list = NULL;
+	if (strncmp(ic->identifier, "type:", 5) == 0) {
+		config_list = config->input_type_configs;
+		merge_type_on_existing(ic);
+	} else {
+		config_list = config->input_configs;
+	}
+
+	int i = list_seq_find(config_list, input_identifier_cmp,
 			ic->identifier);
 	if (i >= 0) {
 		sway_log(SWAY_DEBUG, "Merging on top of existing input config");
-		struct input_config *current = config->input_configs->items[i];
+		struct input_config *current = config_list->items[i];
 		merge_input_config(current, ic);
 		free_input_config(ic);
 		ic = current;
@@ -167,7 +200,7 @@ struct input_config *store_input_config(struct input_config *ic) {
 			free_input_config(ic);
 			ic = current;
 		}
-		list_add(config->input_configs, ic);
+		list_add(config_list, ic);
 	} else {
 		// New wildcard config. Just add it
 		sway_log(SWAY_DEBUG, "Adding input * config");
--- a/sway/input/input-manager.c
+++ b/sway/input/input-manager.c
@@ -71,6 +71,40 @@ char *input_device_get_identifier(struct wlr_input_device *device) {
 	return identifier;
 }
 
+static bool device_is_touchpad(struct sway_input_device *device) {
+	if (device->wlr_device->type != WLR_INPUT_DEVICE_POINTER ||
+			!wlr_input_device_is_libinput(device->wlr_device)) {
+		return false;
+	}
+
+	struct libinput_device *libinput_device =
+		wlr_libinput_get_device_handle(device->wlr_device);
+
+	return libinput_device_config_tap_get_finger_count(libinput_device) > 0;
+}
+
+const char *input_device_get_type(struct sway_input_device *device) {
+	switch (device->wlr_device->type) {
+	case WLR_INPUT_DEVICE_POINTER:
+		if (device_is_touchpad(device)) {
+			return "touchpad";
+		} else {
+			return "pointer";
+		}
+	case WLR_INPUT_DEVICE_KEYBOARD:
+		return "keyboard";
+	case WLR_INPUT_DEVICE_TOUCH:
+		return "touch";
+	case WLR_INPUT_DEVICE_TABLET_TOOL:
+		return "tablet_tool";
+	case WLR_INPUT_DEVICE_TABLET_PAD:
+		return "tablet_pad";
+	case WLR_INPUT_DEVICE_SWITCH:
+		return "switch";
+	}
+	return "unknown";
+}
+
 static struct sway_input_device *input_sway_device_from_wlr(
 		struct wlr_input_device *device) {
 	struct sway_input_device *input_device = NULL;
--- a/sway/input/input-manager.c
+++ b/sway/input/input-manager.c
@@ -105,6 +105,37 @@ const char *input_device_get_type(struct sway_input_device *device) {
 	return "unknown";
 }
 
+static void apply_input_type_config(struct sway_input_device *input_device) {
+	const char *device_type = input_device_get_type(input_device);
+	struct input_config *type_config = NULL;
+	for (int i = 0; i < config->input_type_configs->length; i++) {
+		struct input_config *ic = config->input_type_configs->items[i];
+		if (strcmp(ic->identifier + 5, device_type) == 0) {
+			type_config = ic;
+			break;
+		}
+	}
+	if (type_config == NULL) {
+		return;
+	}
+
+	for (int i = 0; i < config->input_configs->length; i++) {
+		struct input_config *ic = config->input_configs->items[i];
+		if (strcmp(input_device->identifier, ic->identifier) == 0) {
+			struct input_config *current = new_input_config(ic->identifier);
+			merge_input_config(current, type_config);
+			merge_input_config(current, ic);
+
+			current->input_type = device_type;
+			config->input_configs->items[i] = current;
+			free_input_config(ic);
+			ic = NULL;
+
+			break;
+		}
+	}
+}
+
 static struct sway_input_device *input_sway_device_from_wlr(
 		struct wlr_input_device *device) {
 	struct sway_input_device *input_device = NULL;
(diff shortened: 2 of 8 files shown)


sway/config/input.c
```diff
@@ -156,7 +158,7 @@ struct input_config *store_input_config(struct input_config *ic) {
 		merge_input_config(current, ic);
 		free_input_config(ic);
 		ic = current;
-	} else if (!wildcard) {
+	} else if (!wildcard && !type_wildcard) {
```
RedSoxFan: This should be executed, otherwise the following will create four input configs for `type:touchpad`:
```
input type:touchpad {
  tap enabled
  tap_button_map lmr
  scroll_method edge
  natural_scroll disabled
}
```

sway/config/input.c
```diff
@@ -156,7 +158,7 @@ struct input_config *store_input_config(struct input_config *ic) {
...
 		sway_log(SWAY_DEBUG, "Adding non-wildcard input config");
 		i = list_seq_find(config->input_configs, input_identifier_cmp, "*");
 		if (i >= 0) {
```
RedSoxFan: If this is not a type config and a matching type config exists, that should be used as the base here

sway/config/input.c
```diff
@@ -148,6 +148,8 @@ struct input_config *store_input_config(struct input_config *ic) {
 		merge_wildcard_on_all(ic);
 	}
 
+	bool type_wildcard = strncmp(ic->identifier, "type:", 5) == 0;
```
RedSoxFan: If this is a type config, it should be merged on all configs of the type similar to the wildcard. Note that this will not have any effect in the config since the inputs do not yet exist so it will have to be handled on retrieval as well

sway/input/input-manager.c
```diff
@@ -738,15 +772,25 @@ void input_manager_configure_xcursor(void) {
 
 struct input_config *input_device_get_config(struct sway_input_device *device) {
 	struct input_config *wildcard_config = NULL;
+	struct input_config *type_config = NULL;
```
RedSoxFan: This is not sufficient. This will need a somewhat simpler version of `get_output_config`

sway/config/input.c
```diff
@@ -170,10 +172,9 @@ struct input_config *store_input_config(struct input_config *ic) {
 		list_add(config->input_configs, ic);
 	} else {
 		// New wildcard config. Just add it
-		sway_log(SWAY_DEBUG, "Adding input * config");
+		sway_log(SWAY_DEBUG, "Adding input %s config", ic->identifier);
```
RedSoxFan: This should not be changed

ddevault:
Missing man page updates.

RedSoxFan:
The layering here also does not work:
```
swaymsg input <input> events disable
swaymsg input \* events enable

swaymsg input <keyboard> xkb_layout fr
swaymsg input type:keyboard xkb_layout us
```

During runtime, wildcard and type config changes should have the diffs merged on top like the `merge_wildcard_on_all` function that this removes.

lolzballs:
> During runtime, wildcard and type config changes should have the diffs merged on top like the `merge_wildcard_on_all` function that this removes.

Just a quick question about this, is it the intended behavior that a wildcard later on in the config overwrites an identifier specific one?

To illustrate, the following currently turns natural scrolling off, even on `<identifier>`. The way my code works is that wildcard < type_wildcard < identifier, kinda like CSS, where the more specific selectors are used. 

```
input <identifier> {
    tap enabled
    natural_scroll enabled
}

input * {
    natural_scroll disabled
}
```

RedSoxFan:
Yes, the config should be parsed from top to bottom where the wildcard can override the identifier. This allows for items to be overridden easier for included configs.

Obviously, type configs cannot work this way during the initial read since the device does not yet exist and the type is unknown.

lolzballs:
So how should type configs work with this model?

I see a few reasonable options:
- Keep type config rules consistent. This can be either the type config being overridden by the device configs (as there were before), or the device configs being overridden by the type config. New rules that appear follow this as well.
- Apply type config on top of the existing device config as the device appears. New rules apply on top of these.
- Keep all rules consistent, where the rules are layered on top of each other, independent of order.

RedSoxFan:
> Apply type config on top of the existing device config as the device appears. New rules apply on top of these.

I think this option would yield the best results

sway/input/input-manager.c
```diff
@@ -507,6 +541,18 @@ static void handle_new_input(struct wl_listener *listener, void *data) {
 	sway_log(SWAY_DEBUG, "adding device: '%s'",
 		input_device->identifier);
 
+	for (int i = 0; i < config->input_type_configs->length; i++) {
+		struct input_config *input_config = config->input_type_configs->items[i];
+		bool type_wildcard = strncmp(input_config->identifier, "type:", 5) == 0;
+		if (type_wildcard &&
+				strcmp(input_config->identifier + 5,
```
RedSoxFan: The line wrapping here is pretty ugly. Can we extract this into something like:
```c
static const char *extract_type_wildcard(const char *identifier) {
    if (strncmp(identifier, "type:", strlen("type:")) == 0) {
        return identifier + strlen("type:");
    }
    return NULL;
}
```
```c
const char *type = extract_type_wildcard(input_config->identifier);
if (type && strcmp(type, input_device_get_type(input_device)) == 0) {
    // ...
}
```

sway/input/input-manager.c
```diff
@@ -659,8 +705,16 @@ void input_manager_set_focus(struct sway_node *node) {
 void input_manager_apply_input_config(struct input_config *input_config) {
 	struct sway_input_device *input_device = NULL;
 	bool wildcard = strcmp(input_config->identifier, "*") == 0;
+	bool type_wildcard = strncmp(input_config->identifier, "type:", 5) == 0;
 	wl_list_for_each(input_device, &server.input->devices, link) {
+		if (type_wildcard &&
```
RedSoxFan: This should occur in `store_input_config`, not here. Merging it here will make it so any device specific options will always be shadowed by the type config.

lolzballs: Sorry, this process has been really long because I haven't been really sure about how the layering of the configs should be handled. I want to get it right this time!

So to clarify, when a type config is stored, we merge it on top of the existing device configs. When a new device appears, we use the device config, merged on top of the type config.

RedSoxFan: There is no rush. We would rather have things done right than fast. There are no deadlines or anything to meet. We are all volunteering our time after all.

That sounds correct. You'll either have to search for a device that matches the config or store the type in the input config when a device matches it.

On a related note, I just realized that `merge_wildcard_on_all` needs to be modified to merge on top of the type configs as well as the device configs

lolzballs: > There is no rush. We would rather have things done right than fast. There are no deadlines or anything to meet. We are all volunteering our time after all.

Wasn't implying there was a rush, I just felt like I was wasting your time by doing this with trial and error basically. That being said, I'm glad you guys actually take time to review things thoroughly and it makes me a lot more appreciative of the work you guys do, and me more happy to use it.

sway/config/input.c
```diff
@@ -140,19 +141,53 @@ static void merge_wildcard_on_all(struct input_config *wildcard) {
...
+            sway_log(SWAY_DEBUG, "Merging %s top of %s",
+                type_wildcard->identifier,
+                ic->identifier);
+			merge_input_config(ic, type_wildcard);
+		}
+	}
 }
 
 struct input_config *store_input_config(struct input_config *ic) {
+	list_t *config_list = NULL;
```
RedSoxFan: Can you move this down to just before the if-else on L179 where this value is set? (No blank line will be needed between this and the if)

sway/input/input-manager.c
```diff
@@ -507,6 +541,31 @@ static void handle_new_input(struct wl_listener *listener, void *data) {
...
+	}
+
+	// Merge new device config on top of type wildcard, if the prior exists
+	for (int i = 0; i < config->input_configs->length; i++) {
+		struct input_config *ic = config->input_configs->items[i];
+		if (strcmp(input_device->identifier, ic->identifier) == 0) {
+			struct input_config *current = new_input_config(ic->identifier);
+			merge_input_config(current, type_config);
+			merge_input_config(current, ic);
+			current->input_type = device_type;
```
RedSoxFan: This is just preference, but can you change the ordering to:
```c
current->input_type = device_type;
config->input_configs->items[i] = current;
free_input_config(ic);
```

It shouldn't make a difference, but it's a good idea to remove references before freeing.

sway/input/input-manager.c
```diff
@@ -659,9 +718,11 @@ void input_manager_set_focus(struct sway_node *node) {
 void input_manager_apply_input_config(struct input_config *input_config) {
 	struct sway_input_device *input_device = NULL;
 	bool wildcard = strcmp(input_config->identifier, "*") == 0;
+	bool type_wildcard = strncmp(input_config->identifier, "type:", 5) == 0;
 	wl_list_for_each(input_device, &server.input->devices, link) {
 		if (strcmp(input_device->identifier, input_config->identifier) == 0
-				|| wildcard) {
+				|| wildcard
+				|| type_wildcard) {
```
RedSoxFan: This doesn't look right. I think this should be checking to see if the device's type matches the config's type

sway/input/input-manager.c
```diff
@@ -785,7 +846,18 @@ void input_manager_configure_xcursor(void) {
 
 struct input_config *input_device_get_config(struct sway_input_device *device) {
 	struct input_config *wildcard_config = NULL;
+	struct input_config *type_config = NULL;
 	struct input_config *input_config = NULL;
+	const char *device_type = input_device_get_type(device);
+
+	for (int i = 0; i < config->input_type_configs->length; ++i) {
```
RedSoxFan: Can you move this below the input config loop? If a device specific config is found, there is no reason to check the type configs. Additionally with this change, the `type_config` variable can be removed and just return the type config when found. The `device_type` variable declaration can be moved to be just before the loop as well (no blank line).

RedSoxFan: Looks like this still needs to be done. Other than that, looks good to me

sway/input/input-manager.c
```diff
@@ -786,6 +851,7 @@ void input_manager_configure_xcursor(void) {
 struct input_config *input_device_get_config(struct sway_input_device *device) {
 	struct input_config *wildcard_config = NULL;
 	struct input_config *input_config = NULL;
+
```
RedSoxFan: Can you remove this blank line?

sway/input/input-manager.c
```diff
@@ -795,5 +861,14 @@ struct input_config *input_device_get_config(struct sway_input_device *device) {
...
+	for (int i = 0; i < config->input_type_configs->length; ++i) {
+		input_config = config->input_type_configs->items[i];
+		if (strcmp(input_config->identifier + 5, device_type) == 0) {
+			return input_config;
+		}
+	}
+
 	return wildcard_config;
 }
+
```
RedSoxFan: This one too

RedSoxFan:
Code LGTM, but the commit history needs to be polished:
- The first commit can be combined with the second as it only has one effective line change and doesn't need to be it's own commit
- Can you please also reword the commit messages?
   - First+Second combined commit:
     - Short: `Add heuristics to detect touchpads`
     - Long: The current `Use libinput_device_config_tap_get_finger_count to determine whether
a device is a touchpad.` is fine
    - The last commit:
        - Short: `Implement input type configs (#3784)`
        - Long: Something that encompasses the commit instead of an intermediate change

sway/input/input-manager.c
```diff
@@ -541,6 +541,33 @@ static void handle_new_input(struct wl_listener *listener, void *data) {
...
+	struct input_config *type_config = NULL;
+	for (int i = 0; i < config->input_type_configs->length; i++) {
+		struct input_config *ic = config->input_type_configs->items[i];
+		if (strcmp(ic->identifier + 5, device_type) == 0) {
+			type_config = ic;
+			break;
+		}
+	}
+
+	// Merge new device config on top of type wildcard, if the prior exists
```
RedSoxFan: Sorry missed this one before. This block needs to be wrapped in a `if (type_config) {` otherwise sway segfaults when trying to merge the NULL type config on top of `current` (line 559)

RedSoxFan:
Thanks!

lolzballs:
No problem! Thanks for bearing with my misunderstandings.

Also #3784 can be closed now.

RedSoxFan:
> Also #3784 can be closed now.

Just for future reference, if you add `Fixes #<num>` (generally use this one for bug fixes) or `Closes #<num>` (generally use this one for enhancements) to the original post, it will automatically close the listed issues when the PR is merged.

PR #3906, opened 2019-03-15
output_cmd_background: validate colors

RedSoxFan:
This validates the color and fallback color in `output_cmd_background`
to ensure that only colors of the form `#RRGGBB` are accepted.


emersion:
Thanks!

PR #3915, opened 2019-03-15
Fix quoting of commands passed to for_window

minus7:
E.g. `for_window [class="mpv"] move container to output "Dell Inc. ..."`
does not work because the executed move command only uses `Dell` as
output name.


RedSoxFan:
Thanks!

PR #3905, opened 2019-03-15
config/output: revamp identifier/name layering

RedSoxFan:
Fixes #3891 

This revamps the way that output configs are handled when referencing
an output by both identifier and name. If an output is always referred
to by name or by identifier, this should have no noticeable change. As
soon as there is a name output config and an identifier output config
that matches an output, an output config is generated that is named
`<identifier> on <name>` that is generated with the identifier output
config merged on top of the name output config and stored. When a
change to either is stored, the delta is merged on top of that
"id on name" output config, as well. If the "id on name" output config
exists, it has the highest precedence and will be used when applying
a config to the output.

This fixes the following case:
- `swaymsg output <name> bg /path/to/wallpaper1 fill`
- `swaymsg output <identifier> bg /path/to/wallpaper2 fill`
- `swaymsg output <name> dpms on`

Without this, the wallpaper is changed to `/path/to/wallpaper1`. With
this, the wallpaper remains `/path/to/wallpaper2`.

PR #3916, opened 2019-03-15
Fix output config retrieval for new outputs

RedSoxFan:
This removes `output_find_config`, which would take the first matching
output config it found. This is fine if only a name output config,
identifier output config, or even just wildcard exist, but if there is
a name output config and identifier output config, they are not merged.
Instead, this introduces find_output_config, which is just a wrapper
for `get_output_config`. This ensures that both the name and identifier
output configs are respected.

This fixes the following case:
- For simplicity in this example, remove all output configs from config
- Run `swaymsg output <name> bg #ff0000 solid_color`
- Run `swaymsg output <identifier> scale 2`
- (Note the bg will be removed here. That will be fixed by #3905)
- Disconnect and reconnect output

Without this, the output will have the background, but not the scale.
With this, the output will have both the background and scale

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -585,6 +585,8 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output);
 
 struct output_config *store_output_config(struct output_config *oc);
 
+struct output_config *find_output_config(struct sway_output *output);
+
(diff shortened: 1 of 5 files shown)


ddevault:
Thanks!

Issue #3813, opened 2019-03-04
Introduce default seatop


RyanDwyer:
For after 1.0 is released.

The idea is we'd add a none/idle/normal/default seatop which handles most of `dispatch_cursor_button` and `cursor_send_pointer_motion` (ie. click to focus, `focus_follows_mouse`, starting other seatops). We would then treat seatops as being more like a seat state, so there's always a seatop in progress. It might make more sense to rename seatops to seat states at this point.

This would remove a good chunk of code out of `cursor.c`, would remove some early returns, and should make it easier to manage things like sending button events and tracking which buttons are pressed.

emersion:
+1

We do the same in wlroots for grabs, we have a default grab.

PR #3917, opened 2019-03-16
Introduce default seatop

RyanDwyer:
This introduces a `default` seat operation which is used when no mouse buttons are being held. This means there is now always a seat operation in progress. It allows us to separate `default` code from the standard cursor management code.

Note that as of creation, this PR has two commits. The first is small and replaces the seatop `allows_events` boolean with a `button` callback. The second commit is somewhat significant. The rest of this description relates to the second commit.

The `sway_seatop_impl` struct has gained callbacks `axis`, `rebase` and `end`, and lost callbacks `finish` and `abort`. `axis` and `rebase` are only used by the default seatop. `end` is called when a seatop is being replaced by another one and allows the seatop to free any resources, though no seatop currently needs to do this. `finish` is no longer required, as each seatop can gracefully finish in their `button` callback. And `abort` is not needed, as calling `end` would achieve the same thing. The struct has also gained a bool named `allow_set_cursor` which allows the client to set a new cursor during `default` and `down` seatops.

Seatops would previously store which button they were started with and stop when that button was released. This behaviour is changed so that it only ends once all buttons are released. So you can start a drag with $mod+left, then click and hold right, release left and it'll continue dragging while the right button is held.

The motion callback now accepts `dx` and `dy`. Most seatops don't use this as they store the cursor position when the seatop is started and compare it with the current cursor position. This approach doesn't make sense for the default seatop though, hence why `dx` and `dy` are needed.

The `pressed_buttons` array has been moved from the `sway_cursor` struct to the default seatop's data. This is only used for the default seatop to check bindings. The total pressed button count remains in the `sway_cursor` struct though, because all the other seatops check it to know if they should end.

The `down` seatop no longer has a `moved` property. This was used to track if the cursor moved and to recheck `focus_follows_mouse`, but seems to work without it.

The logic for `focus_follows_mouse` has been refactored. As part of this I've removed the call to `wlr_seat_keyboard_has_grab` as we don't appear to use keyboard grabs.

The functions for handling relative motion, absolute motion and tool axis have been changed. Previously the handler functions were `handle_cursor_motion`, `handle_cursor_motion_absolute` and `handle_tool_axis`. The latter two both called `cursor_motion_absolute`.  Both `handle_cursor_motion` and `cursor_motion_absolute` did very similar things. These are now simplified into three handlers and a single common function called `cursor_motion`. All three handlers call `cursor_motion`. As `cursor_motion` works with relative distances, the absolute and tool axis handlers convert them to relative first.

Completes #3813.

Running this as my daily from now on, though I don't use focus follows mouse nor do I use the mouse much.

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -52,8 +52,6 @@ struct sway_cursor {
 	struct wl_event_source *hide_source;
 	bool hidden;
 
-	// Mouse binding state
-	uint32_t pressed_buttons[SWAY_CURSOR_PRESSED_BUTTONS_CAP];
 	size_t pressed_button_count;
 };
 
@@ -77,13 +75,6 @@ void cursor_rebase_all(void);
 void cursor_handle_activity(struct sway_cursor *cursor);
 void cursor_unhide(struct sway_cursor *cursor);
 int cursor_get_timeout(struct sway_cursor *cursor);
-
-/**
- * Like cursor_rebase, but also allows focus to change when the cursor enters a
- * new container.
- */
-void cursor_send_pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
-	struct sway_node *node, struct wlr_surface *surface, double sx, double sy);
 
 void dispatch_cursor_button(struct sway_cursor *cursor,
 	struct wlr_input_device *device, uint32_t time_msec, uint32_t button,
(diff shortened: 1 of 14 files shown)


RyanDwyer:
I've been using this for over 24 hours now and haven't noticed any regressions. I'm now inviting others to review and test.

ddevault:
Thanks!

PR #3918, opened 2019-03-16
Flush stdout in swaymsg when in subscribe mode

Flakebi:
When in subscribe mode **and** piping to a process, the notifications are only arriving when the stdout buffer is full.
An easy way to reproduce that is `swaymsg -m -t subscribe '["window"]' | cat`.

Flushing stdout after printing a message fixes that.


RedSoxFan:
Thanks!

PR #3920, opened 2019-03-17
config/output: handle wildcard in get_output_config

### LINKED ISSUE #3919: Sway background is grey until config reload
sway version: `sway version 1.0-rc1-184-g73605dac (Mar 16 2019, branch 'master')`
[sway.log](https://github.com/swaywm/sway/files/2974128/sway.log)
[config](https://github.com/swaywm/sway/files/2974131/config.txt)
The relevant line from the config: `output * bg "/home/freefull/wallpapers/0_a4ab4_6dab62ae_orig.jpg" stretch`

When I start up sway, the background ends up grey instead of getting set to the image in my config. As soon as I reload, the background gets set correctly.

RedSoxFan:
Fixes #3919 

In #3916, I overlooked that `get_output_config` does not handle
wildcards unless the config is reloading, which is a remnant of older
iterations of the output config handling that went unnoticed due to
`output_find_config` handling it. With the current version of the
output config handling, having `get_output_config` handle wildcard
configs is actually preferable. This fixes having only a wildcard
output config in the config file or when connecting/enabling a new
output with only a wildcard config existing.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -493,19 +493,20 @@ static struct output_config *get_output_config(char *identifier,
 		free(result->name);
 		result->name = strdup(identifier);
 		merge_output_config(result, oc_id);
-	} else if (config->reloading) {
-		// Neither config exists, but we need to reset the output so create a
-		// default config for the output and if a wildcard config exists, merge
-		// that on top
-		free(result->name);
-		result->name = strdup("*");
+	} else {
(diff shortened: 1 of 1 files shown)


RedSoxFan:
> Shouldn't this merge the wildcard config in all cases?

It's merged on to the output configs when they are stored, which has better layering. See `merge_wildcard_on_all`

emersion:
Thanks!

Issue #3907, opened 2019-03-15
Please extend the README 'running' section to suggest keeping xwayland installed


SjonHortensius:
As a new user of wayland using sway it is tempting to remove xorg-server-xwayland as it seems like a minor and optional compatibility layer. However, as #2965 and other issues prove (I have been bitten by this as well) - most applications (including `dmenu`, `sway` itself and `termite`) do not function if xwayland is not installed (unless explicitly configured - which is counter-intuitive). I would suggest to extend the "Running" section of the README on the frontpage with:

> You will probably want to have xwayland installed, since most wayland applications will not start without it

c-edw:
This is something you can send a PR for. That way you can phrase it however you see fit.

emersion:
Yeah, please send a PR.

SjonHortensius:
Sure, I just wanted to gauge the sentiment first. :)

PR #3922, opened 2019-03-17
Update README - append hint to keep xwayland installed when running sway

SjonHortensius:
as discussed in #3907, many wayland-native apps refuse to run (unless
configured as such) without having xwayland installed. This can confuse
new users who might be tempted to remove anything xorg related - only to
end up with a broken wayland setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,4 +70,6 @@ Run `man 5 sway` for information on the configuration.
 ## Running
 
 Run `sway` from a TTY. Some display managers may work but are not supported by
-sway (gdm is known to work fairly well).
+sway (gdm is known to work fairly well). You will probably want to have xwayland
+installed, since most wayland applications (including dmenu and sway) will not
+start without it.


c-edw:
Correct me if I'm wrong, but you've used a translation tool for this, yes? I'm sure not what the procedure is supposed to be but most of the time I see people just updating the English `README.md` and then those changes eventually trickle down to the translated versions. It would be nice if somebody could clarify this.

SjonHortensius:
> Correct me if I'm wrong, but you've used a translation tool for this, yes?

Yes, and I've reverse-checked a few to confirm it wasn't total gibberish.

> I'm sure not what the procedure is supposed to be but most of the time I see people just updating the English README.md and then those changes eventually trickle down to the translated versions.

I checked for any rules or guidance on this, but found none. I did find that the [last change](https://github.com/swaywm/sway/commit/3fca8b8d22ee51a934fa3b8704c5aee8a5de7689) appeared to do mass-edits to all translated READMEs as well.

> A lot of the translations right now are outdated and might just lead to further confusion. Maybe we should discuss this.

Maybe a related issue could be created to discuss how to do this in the future? It would make sense to have a list of contributors per language who would be responsible for updating their part of the docs once changes are made.

ddevault:
We don't accept machine translations. Stick to updating the english readme.

SjonHortensius:
I've pushed another commit - but I'm pretty sure [others have been updating a few readme's as well](https://github.com/swaywm/sway/commit/3fca8b8d22ee51a934fa3b8704c5aee8a5de7689) for languages they don't fully understand ;)

README.md
```diff
@@ -70,4 +70,6 @@ Run `man 5 sway` for information on the configuration.
 ## Running
 
 Run `sway` from a TTY. Some display managers may work but are not supported by
-sway (gdm is known to work fairly well).
+sway (gdm is known to work fairly well). You will probably want to have xwayland
+installed, since most wayland applications (including dmenu and sway) will not
```
Emantor: dmenu is not a wayland application

SjonHortensius: Thanks - I actually mentioned `dmenu` here because it is included in the default configuration. I'll rewrite it

ddevault:
Removing text from READMEs in other languages is fine, it's adding new text that's the problem :smile:

README.md
emersion: It's not that most Wayland apps won't start without it. It's just that if you don't want Xwayland, you need to explicitly state it in your config file.

ddevault:
I think I'd rather not bother explaining this here - not to mention that this PR still hasn't been updated to meet the criteria for merge. Thanks, though.

PR #3926, opened 2019-03-18
swaybg: one instance for all outputs

### LINKED ISSUE #2051: Proposal: only one swaybg instance for all outputs

### LINKED ISSUE #3892: swaybg is restarted when "output * dpms on" is sent, even if dpms is already on
```
sway version 1.0-rc1-178-gd64e8ba9 (Mar 14 2019, branch 'master')
```

When I run `swaymsg "output * dpms on"`, `swaybg` is restarted and the background of every output flashes for a few frames, even when the output is already on. This is mildly annoying.

This is triggered by using the below swayidle when charging, the dpms is never turned off but turning dpms on must happen every time.
```
swayidle \
    timeout 300 '[[ $(cat /sys/class/power_supply/BAT1/status) == "Discharging" ]] && swaymsg "output * dpms off"' \
    resume 'swaymsg "output * dpms on"'`
```

RedSoxFan:
Closes #2051 
Fixes #3892

This makes it so there will only be one swaybg instance running
instead of one per output. swaybg's cli has been changed to a xrandr
like interface, where you select an output and then change properties
for that output and then select another output and repeat. This also
makes it so swaybg is only killed and respawned when a background
changes or when reloading.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -228,91 +228,6 @@ static bool set_mode(struct wlr_output *output, int width, int height,
 	return wlr_output_set_mode(output, best);
 }
 
-static void handle_swaybg_client_destroy(struct wl_listener *listener,
-		void *data) {
-	struct sway_output *output =
-		wl_container_of(listener, output, swaybg_client_destroy);
-	wl_list_remove(&output->swaybg_client_destroy.link);
-	wl_list_init(&output->swaybg_client_destroy.link);
-	output->swaybg_client = NULL;
-}
-
-static bool set_cloexec(int fd, bool cloexec) {
-	int flags = fcntl(fd, F_GETFD);
-	if (flags == -1) {
-		sway_log_errno(SWAY_ERROR, "fcntl failed");
-		return false;
-	}
-	if (cloexec) {
-		flags = flags | FD_CLOEXEC;
-	} else {
-		flags = flags & ~FD_CLOEXEC;
-	}
-	if (fcntl(fd, F_SETFD, flags) == -1) {
-		sway_log_errno(SWAY_ERROR, "fcntl failed");
-		return false;
-	}
-	return true;
-}
-
-static bool spawn_swaybg(struct sway_output *output, char *const cmd[]) {
-	int sockets[2];
-	if (socketpair(AF_UNIX, SOCK_STREAM, 0, sockets) != 0) {
-		sway_log_errno(SWAY_ERROR, "socketpair failed");
-		return false;
-	}
-	if (!set_cloexec(sockets[0], true) || !set_cloexec(sockets[1], true)) {
-		return false;
-	}
-
-	output->swaybg_client = wl_client_create(server.wl_display, sockets[0]);
-	if (output->swaybg_client == NULL) {
-		sway_log_errno(SWAY_ERROR, "wl_client_create failed");
-		return false;
-	}
-
-	output->swaybg_client_destroy.notify = handle_swaybg_client_destroy;
-	wl_client_add_destroy_listener(output->swaybg_client,
-		&output->swaybg_client_destroy);
-
-	pid_t pid = fork();
-	if (pid < 0) {
-		sway_log_errno(SWAY_ERROR, "fork failed");
-		return false;
-	} else if (pid == 0) {
-		pid = fork();
-		if (pid < 0) {
-			sway_log_errno(SWAY_ERROR, "fork failed");
-			_exit(EXIT_FAILURE);
-		} else if (pid == 0) {
-			if (!set_cloexec(sockets[1], false)) {
-				_exit(EXIT_FAILURE);
-			}
-
-			char wayland_socket_str[16];
-			snprintf(wayland_socket_str, sizeof(wayland_socket_str),
-				"%d", sockets[1]);
-			setenv("WAYLAND_SOCKET", wayland_socket_str, true);
-
-			execvp(cmd[0], cmd);
-			sway_log_errno(SWAY_ERROR, "execvp failed");
-			_exit(EXIT_FAILURE);
-		}
-		_exit(EXIT_SUCCESS);
-	}
-
-	if (close(sockets[1]) != 0) {
-		sway_log_errno(SWAY_ERROR, "close failed");
-		return false;
-	}
-	if (waitpid(pid, NULL, 0) < 0) {
-		sway_log_errno(SWAY_ERROR, "waitpid failed");
-		return false;
-	}
-
-	return true;
-}
-
 bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 	if (output == root->noop_output) {
 		return false;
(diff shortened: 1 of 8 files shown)


sway/config/output.c
```diff
@@ -569,3 +465,133 @@ void free_output_config(struct output_config *oc) {
...
+	if (pid < 0) {
+		sway_log_errno(SWAY_ERROR, "fork failed");
+		return false;
+	} else if (pid == 0) {
+		pid = fork();
+		if (pid < 0) {
+			sway_log_errno(SWAY_ERROR, "fork failed");
+			_exit(EXIT_FAILURE);
+		} else if (pid == 0) {
+			if (!set_cloexec(sockets[1], false)) {
```
ascent12: Wouldn't it make more sense to not set CLOEXEC in the first place?

RedSoxFan: I think it's needed for the double fork for the wayland client socket. _Just as a note, that's currently in master. I just moved the function_

ianyfan:
Firstly, I like the idea of a wildcard config, either if you specify arguments before specifying an output for the first time, or if you explicitly specify an output of `*`.

Secondly, it doesn't seem like your code checks if the output config already exists, so say if you pass `-o X -i Y ... -o X -m M` you could merge the configs.

RedSoxFan:
> Firstly, I like the idea of a wildcard config, either if you specify arguments before specifying an output for the first time, or if you explicitly specify an output of `*`.

The output `*` should work as a wildcard. If desired, I could change it so anything before the first output gets treated as wildcard, as well
https://github.com/swaywm/sway/blob/50ef791d51f938a367f4086164b7fc9617144c63/swaybg/main.c#L205-L216

> Secondly, it doesn't seem like your code checks if the output config already exists, so say if you pass `-o X -i Y ... -o X -m M` you could merge the configs

Sounds good. I'll make the change and push

RedSoxFan:
> Secondly, it doesn't seem like your code checks if the output config already exists, so say if you pass `-o X -i Y ... -o X -m M` you could merge the configs

Done

sway/config/output.c
emersion: We need to set CLOEXEC in the compositor process, but unset it in the child process. So this is really needed.

sway/config/output.c
```diff
@@ -569,3 +465,137 @@ void free_output_config(struct output_config *oc) {
...
+				_exit(EXIT_FAILURE);
+			}
+
+			char wayland_socket_str[16];
+			snprintf(wayland_socket_str, sizeof(wayland_socket_str),
+				"%d", sockets[1]);
+			setenv("WAYLAND_SOCKET", wayland_socket_str, true);
+
+			execl("/bin/sh", "/bin/sh", "-c", command, NULL);
```
emersion: Why do we need `sh` here? This requires quoting things, which is not nice

swaybg/main.c
```diff
@@ -77,68 +75,85 @@ bool is_valid_color(const char *color) {
...
+	wl_surface_attach(output->surface, output->current_buffer->buffer, 0, 0);
+	wl_surface_damage_buffer(output->surface, 0, 0, INT32_MAX, INT32_MAX);
+	wl_surface_commit(output->surface);
 }
 
-static bool prepare_context(struct swaybg_state *state) {
-	if (state->args->mode == BACKGROUND_MODE_SOLID_COLOR) {
-		state->context.color = parse_color(state->args->path);
-		return is_valid_color(state->args->path);
+static void free_swaybg_output_config(struct swaybg_output_config *config) {
```
emersion: We generally name those `destroy`

swaybg/main.c
```diff
@@ -190,24 +202,90 @@ static void xdg_output_handle_logical_size(void *data,
...
+		find_config(output, output->identifier);
+	}
+}
+
+static void create_layer_surface(struct swaybg_output *output) {
+	output->surface = wl_compositor_create_surface(output->state->compositor);
+	assert(output->surface);
+
+	// Empty input region
+	output->input_region =
```
emersion: I don't think we need to save this. We can destroy it right away.

swaybg/main.c
```diff
@@ -296,49 +518,29 @@ int main(int argc, const char **argv) {
...
-			&layer_surface_listener, &state);
-	wl_surface_commit(state.surface);
-	wl_display_roundtrip(state.display);
 
 	state.run_display = true;
 	while (wl_display_dispatch(state.display) != -1 && state.run_display) {
 		// This space intentionally left blank
 	}
 
+	output = NULL;
```
emersion: No need to set this to NULL

swaybg/main.c
RedSoxFan: Doesn't it need to be stored so the `wl_region` can be destroyed later?

swaybg/main.c
RedSoxFan: Removed

swaybg/main.c
RedSoxFan: Changed

sway/config/output.c
RedSoxFan: What would be the best way of fxing this? The `master` version uses `execvp` with an array of arguments. If I try using that here, it is a VLA which is an error based on the compilation flags.

swaybg/main.c
emersion: We can destroy it right away after `wl_surface_set_input_region`.

sway/config/output.c
emersion: You can use an array with a large-enough size instead of a VLA.

```c
char *cmd[32];
size_t i = 0;

cmd[i++] = "swaybg";
cmd[i++] = "-o";
cmd[i++] = "my-beloved-output";
…

assert(i <= sizeof(cmd) / sizeof(cmd[0]));
```

emersion: (BTW I'm assuming you'll want to add items in loops and whatnot -- otherwise it would've been simpler)

swaybg/main.c
RedSoxFan: Changed

sway/config/output.c
RedSoxFan: Changed to use `execvp` again. I'm not really sure what the best size for the array is though.

emersion: We could always dynamically allocate it, but I think a reasonably large value will be enough (e.g. 128).

RedSoxFan: It's at 128 with a comment about how many output configs it will allow for (31 color only, 21 image only, 15 image with fallback color). If anyone complains, we can bump it up or something later

RedSoxFan: Actually, I'll just switch it to dynamic allocation in a few hours (have to go do some stuff now). For some reason, I was thinking it had to be an array with a fixed length...

RedSoxFan: Switched to dynamic allocation

sway/config/output.c
```diff
@@ -584,3 +480,150 @@ void free_output_config(struct output_config *oc) {
...
+			cmd[i++] = oc->background;
+			cmd[i++] = "-m";
+			cmd[i++] = oc->background_option;
+			if (oc->background_fallback) {
+				cmd[i++] = "-c";
+				cmd[i++] = oc->background_fallback;
+			}
+		}
+	}
+
```
emersion: Can we assert that we've not assigned past the array end here?

RedSoxFan: Done

emersion:
Thanks!

We can now think about moving `swaybg` to a standalone repo.

PR #3931, opened 2019-03-18
Add modifiers array to event_json

petoknm:
I want to make our sway i3bar implementation adhere to the [i3bar protocol](https://i3wm.org/docs/i3bar-protocol.html#_click_events) by adding missing "modifiers" field to event_json. (I'm working on an i3blocks blocklet and I'm missing the modifiers when using sway, i3 provides them)

Currently I only added an empty array because I don't know exactly how to get the modifiers. The button variable does not contain any information about the modifiers so I guess we will have to call into wlr or somehow get the information into the function. I'm not sure about the inner architecture of sway to make a decision how exactly to do this, so I would like if someone could point me in the right direction.


RedSoxFan:
> Currently, I only added an empty array because I don't know exactly how to get the modifiers.

swaybar does not receive keyboard events so it's not going to be able to access which modifiers are pressed

emersion:
Someone wanted to get the modifier information out of sway to display the current keyboard layout.

petoknm:
I assume its sway sending those events... I think we could include modifiers along with those mouse events, or at least a way to somehow get the modifier state.

kraftwerk28:
I made an implementation here: #6701, though it forced me to add a property to `get_inputs` command output, so swaybar can retrieve a list of modifiers.

Issue #3927, opened 2019-03-18
Cannot address monitors by their company name


stratosgear:
My monitors output:

```
➜ swaymsg -t get_outputs
Output eDP-1 'Panasonic Industry Company 0x96A2 0x00000000'
  Current mode: 2560x1440 @ 59.999001 Hz
  Position: 3520,300
  Scale factor: 1.200000
  Transform: normal
  Workspace: 4:
  Available modes:
    2560x1440 @ 48.000000 Hz
    2560x1440 @ 59.999001 Hz

Output DP-7 'Goldstar Company Ltd 20EN33 303NDMTMM283'
  Current mode: 1600x900 @ 60.000000 Hz
  Position: 0,300
  Scale factor: 1.000000
  Transform: normal
  Workspace: 3:
  Available modes:
    720x400 @ 70.082001 Hz
    640x480 @ 59.939999 Hz
    640x480 @ 75.000000 Hz
    800x600 @ 56.250000 Hz
    800x600 @ 60.317001 Hz
    800x600 @ 75.000000 Hz
    832x624 @ 74.551003 Hz
    1024x768 @ 60.004002 Hz
    1024x768 @ 75.028999 Hz
    1280x720 @ 60.000000 Hz
    1152x864 @ 75.000000 Hz
    1280x800 @ 59.810001 Hz
    1440x900 @ 59.887001 Hz
    1600x900 @ 60.000000 Hz

Output DP-6 'Dell Inc. DELL U2412M 0FFXD46R5Y6S' (focused)
  Current mode: 1920x1200 @ 59.950001 Hz
  Position: 1600,0
  Scale factor: 1.000000
  Transform: normal
  Workspace: 2:
  Available modes:
    720x400 @ 70.082001 Hz
    640x480 @ 59.939999 Hz
    800x600 @ 60.317001 Hz
    1024x768 @ 60.004002 Hz
    1280x960 @ 60.000000 Hz
    1280x1024 @ 60.020000 Hz
    1680x1050 @ 59.953999 Hz
    1600x1200 @ 60.000000 Hz
    1920x1080 @ 60.000000 Hz
    1920x1200 @ 59.950001 Hz
```

If I do any sway output commands with eDP-1, or DP-x, they work fine:

```
 sway output eDP-1 scale 1.5
[ { "success": true } ]
```

But using the *"make, model and serial which you can get from swaymsg -t get_outputs."* as described in **man 5 sway-output** gives me:

```
[...]

PR #3933, opened 2019-03-18
Clarify usage of "make, model and serial" string

stratosgear:
As identified in #3927


ddevault:
Like @emersion said, this isn't the right solution.

PR #3940, opened 2019-03-20
cmd_mode: allow cmd_set to be a subcommand

RedSoxFan:
Related https://github.com/swaywm/sway/issues/2336#issuecomment-474661290

This allows set to be used in mode blocks


emersion:
Thanks!

PR #3943, opened 2019-03-20
Support focus <direction> for floating containers

### LINKED ISSUE #3859: focus floating window with keyboard (mod+left or mod+right)
open two applications in floating mode, partially or fully covered.
Change the application (mod + left or mod + right) using the keyboard and the application does not pass to the front.

Sorry for my English.

RyanDwyer:
This kind of worked before in that focus would change, but it wasn't intentionally supported and had side effects such as not raising the container, and being unable to cycle through all floaters depending on the direction used.

This commit makes it properly supported. The new focus is chosen based on the distance to the center point of each floating container in the workspace, and the container is raised.

In a multi output setup, if both visible workspaces have floating containers, focus will NOT cross into the other output. It is assumed the user will use a workspace binding in this case.

If two floating containers occupy the exact same center point and you try to focus in a direction, the behaviour is undefined.

Fixes #3859.

To test, float a few views, move them around a bit so none are on the same x/y as any other, then go wild with the `focus <direction>` commands.

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -1,3 +1,4 @@
+#include <float.h>
 #include <strings.h>
 #include <wlr/types/wlr_output_layout.h>
 #include "log.h"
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

Vorsplummi:
This works great. Thanks!

PR #3956, opened 2019-03-23
Fix #3924 ( Stripping of escaped quotes in exec )

### LINKED ISSUE #3924: exec strips escaped quotes when it should not be
`swaymsg 'exec notify-send "a; b"'` execs `notify-send "a; b"` ☺

` swaymsg 'exec "notify-send \"a; b\""'` execs `notify-send a; b` ☹

In i3, both exhibit the former behaviour.

Philz69:
From what i understand, 
do_var_replacement never adds escape characters.
if the variable's value contains an escaped characther, it should stay escaped after the replacement.

Thus, unescapping the string after calling do_var_replacement is unescessary and strips the escape characters before calling the exec_always.c
And exec_always.c removes the quotes because they are not escaped anymore

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -276,7 +276,6 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 			// Var replacement, for all but first argument of set
 			for (int i = handler->handle == cmd_set ? 2 : 1; i < argc; ++i) {
 				argv[i] = do_var_replacement(argv[i]);
-				unescape_string(argv[i]);
 			}
 
 			if (!config->handler_context.using_criteria) {


Philz69:
I just realised I should've added the description to the commit message. If needed, I will resubmit with a corrected commit message.

RedSoxFan:
`execute_command` gets called for all bindings, any commands executed via IPC, for_window, assign, and any commands that get defered so I have to think more on if this would break anything.

In the meantime, can you please update the commit message and force push?

Philz69:
I have force pushed the ammended commit with the new message!

ddevault:
Thanks!

PR #3957, opened 2019-03-23
fix opening a floating view on the NOOP output

### LINKED ISSUE #3954: segfault in container_init_floating
* Sway Version: 1.0

* Debug Log: https://pub.yori.cc/sway-crash10.log

* Configuration File: https://pub.yori.cc/i3_config

Finally, explain the steps you took in plain English to reproduce the problem below.

1. disconnect from docking station

```
#0  0x000000000044819f in container_init_floating ()
#1  0x0000000000449e32 in container_set_floating ()
#2  0x000000000044dd61 in view_map ()
#3  0x00000000004235e3 in handle_map ()
#4  0x00007f96a300c9fc in wlr_signal_emit_safe ()
   from /nix/store/qy5an0g3krz886f5qvw64b03br1nlmd9-wlroots-408eca7dfa12eda0f1b0ec6050e99ee2e6a8f2b4/lib/libwlroots.so.2
#5  0x00007f96a2ff2049 in handle_xdg_surface_commit ()
   from /nix/store/qy5an0g3krz886f5qvw64b03br1nlmd9-wlroots-408eca7dfa12eda0f1b0ec6050e99ee2e6a8f2b4/lib/libwlroots.so.2
#6  0x00007f96a3006dd2 in surface_commit_pending ()
   from /nix/store/qy5an0g3krz886f5qvw64b03br1nlmd9-wlroots-408eca7dfa12eda0f1b0ec6050e99ee2e6a8f2b4/lib/libwlroots.so.2
#7  0x00007f96a3007088 in surface_commit () from /nix/store/qy5an0g3krz886f5qvw64b03br1nlmd9-wlroots-408eca7dfa12eda0f1b0ec6050e99ee2e6a8f2b4/lib/libwlroots.so.2
#8  0x00007f96a27cd05e in ffi_call_unix64 () from /nix/store/kbbiky87bsdndh63m7919sxkf3gsh5zq-libffi-3.2.1/lib/../lib64/libffi.so.6
#9  0x00007f96a27cc023 in ffi_call () from /nix/store/kbbiky87bsdndh63m7919sxkf3gsh5zq-libffi-3.2.1/lib/../lib64/libffi.so.6
#10 0x00007f96a3050b94 in wl_closure_invoke () from /nix/store/9n45jlbvwi49hlmm0kqlgllhnznq1x04-wayland-1.16.0/lib/libwayland-server.so.0
#11 0x00007f96a304d61f in wl_client_connection_data () from /nix/store/9n45jlbvwi49hlmm0kqlgllhnznq1x04-wayland-1.16.0/lib/libwayland-server.so.0
#12 0x00007f96a304ebe2 in wl_event_loop_dispatch () from /nix/store/9n45jlbvwi49hlmm0kqlgllhnznq1x04-wayland-1.16.0/lib/libwayland-server.so.0
#13 0x00007f96a304d7fa in wl_display_run () from /nix/store/9n45jlbvwi49hlmm0kqlgllhnznq1x04-wayland-1.16.0/lib/libwayland-server.so.0
#14 0x000000000040e509 in main ()
```
Release build, built without debug.

RedSoxFan:
Fixes #3954

Since the NOOP output has no size, the minimum floating size is greater
than the workspace size for the NOOP output. In this case, the floater
gets centered in the output instead of the workspace. However, the
NOOP output is not part of the output layout and thus has a NULL box.
Attempting to access the properties of this box was causing a segfault.

This fixes the issue by just setting the floater's box to all zeroes
when mapping on the NOOP output. When the workspace gets moved from the
NOOP output to a new output, any floater whose width or height is zero
or has an x or y location outside of the output, gets passed to
`container_init_floating` again. This will then set the appropriate
size and centering. For any floater that has a valid size and location,
they are preserved.

### DIFF
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -60,6 +60,27 @@ static void restore_workspaces(struct sway_output *output) {
 		struct sway_workspace *ws = root->noop_output->workspaces->items[0];
 		workspace_detach(ws);
 		output_add_workspace(output, ws);
+		
+		// If the floater was made floating while on the NOOP output, its width
+		// and height will be zero and it should be reinitialized as a floating
+		// container to get the appropriate size and location. Additionally, if
+		// the floater is wider or taller than the output or is completely
+		// outside of the output's bounds, do the same as the output layout has
+		// likely changed and the maximum size needs to be checked and the
+		// floater re-centered
+		for (int i = 0; i < ws->floating->length; i++) {
+			struct sway_container *floater = ws->floating->items[i];
+			if (floater->width == 0 || floater->height == 0 ||
+					floater->width > output->width ||
+					floater->height > output->height ||
+					floater->x > output->lx + output->width ||
+					floater->y > output->ly + output->height ||
+					floater->x + floater->width < output->lx ||
+					floater->y + floater->height < output->ly) {
+				container_init_floating(floater);
+			}
+		}
+
 		ipc_event_workspace(NULL, ws, "move");
 	}
 
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -651,6 +651,17 @@ void floating_calculate_constraints(int *min_width, int *max_width,
 
 void container_init_floating(struct sway_container *con) {
 	struct sway_workspace *ws = con->workspace;
+	struct wlr_box *ob = wlr_output_layout_get_box(root->output_layout,
+			ws->output->wlr_output);
+	if (!ob) {
+		// On NOOP output. Will be called again when moved to an output
+		con->x = 0;
+		con->y = 0;
+		con->width = 0;
+		con->height = 0;
+		return;
+	}
+
 	int min_width, max_width, min_height, max_height;
 	floating_calculate_constraints(&min_width, &max_width,
 			&min_height, &max_height);
@@ -659,8 +670,6 @@ void container_init_floating(struct sway_container *con) {
 		con->width = max_width;
 		con->height = max_height;
 		if (con->width > ws->width || con->height > ws->height) {
-			struct wlr_box *ob = wlr_output_layout_get_box(root->output_layout,
-					ws->output->wlr_output);
 			con->x = ob->x + (ob->width - con->width) / 2;
 			con->y = ob->y + (ob->height - con->height) / 2;
 		} else {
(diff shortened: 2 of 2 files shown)


sway/tree/output.c
```diff
@@ -60,6 +60,13 @@ static void restore_workspaces(struct sway_output *output) {
 		struct sway_workspace *ws = root->noop_output->workspaces->items[0];
 		workspace_detach(ws);
 		output_add_workspace(output, ws);
+		for (int i = 0; i < ws->floating->length; i++) {
```
emersion: Is it possible to add a comment here to explain in which situation we need to check for `width == 0 || height == 0 || …`?

RedSoxFan: Done

sway/tree/output.c
```diff
@@ -60,6 +60,21 @@ static void restore_workspaces(struct sway_output *output) {
...
+		// If the floater was made floating while on the NOOP output, its width
+		// and height will be zero and it should be reinitialized as a floating
+		// container to get the appropriate size and location. Additionally, if
+		// the floater's top-left corner is outside of the workspace, do the
+		// same as the output layout has likely changed and the maximum size
+		// may be different
+		for (int i = 0; i < ws->floating->length; i++) {
+			struct sway_container *floater = ws->floating->items[i];
+			if (floater->width == 0 || floater->height == 0 ||
+					floater->x < ws->x || floater->y < ws->y) {
```
emersion: I don't think this check is right/enough.

Shouldn't we also check for the right/bottom borders? Shouldn't users be able to drag the top-left corner past the workspace bounds?

RedSoxFan: I think it should be good now. I changed it to use the output bounds instead of the workspace bounds and added a check for the bottom/right borders

emersion: One question remains: do we want to allow this configuration?

```
floating
+-------------------------+
|                         |
|                         |
|                         |
|         +------------------------------------------+
|         |               |                          |
|         |               |                          |
|         |               |                          |
|         |               |                          |
+-------------------------+                          |
          |                                          |
          |                                          |
          |                                          |
          |                                          |
          |                                          |
          |                                          |
          |                                          |
          |                                          |
          |                                          |
          |                                          |
          |                                          |
          +------------------------------------------+
          workspace
```

RedSoxFan: I changed it to check for a width/height of zero, a width/height larger than the output, or being completely outside of the output. That should allow for the above to be retained.

emersion:
Thanks!

PR #3958, opened 2019-03-23
criteria: fix __focused__ when no focus or unset

### LINKED ISSUE #3953: [bug or feature ?] exec swaymsg '[non-matching-CRITERIA]' command returns non-zero on empty workspace
* Sway Version: sway version 1.0
* Archlinux: community/sway 1.0-6 [installed]

* Debug Log: http://sprunge.us/8zxts4

* Configuration File:  http://sprunge.us/orXzhF
Should be default config file + this line:
bindsym $mod+q exec swaymsg '[app_id="keepassxc" con_id="__focused__"] move to scratchpad' || xterm

* Steps you took in plain English to reproduce:
  - login on tty
  - run 'sway'
  - press '$mod+q' --> xterm appears (but it should not)
  - press '$mod+q' --> xterm does not appear a 2nd time
  - press '$mod+2' --> new workspace
  - press '$mod+q' --> xterm appears again
  - press '$mod+shift+e' --> to exit sway

Explanation: 
the (non-matching) swaymsg '<criteria>' move to scratchpad must return non-0 on empty workspace... but as the <criteria> is only a filter I guess it should not, but instead the "|| xterm" part should never be executed..

RedSoxFan:
Fixes #3953 

This fixes the behavior of `__focused__` when there is no focused view
to match i3's behavior of successfully matching no views instead of
returning an error of a missing value. It also applies the same logic
when a token is not applicable (or unset) for a view such as `app_id`
for a focused xwayland view or `class` for a focused xdg-shell view.

This adds an `autofail` boolean to `struct criteria`. If it is set to
`true`, then `criteria_matches_view` will immediately bail out as a
no match. If `autofail` is set, the criteria will also not be
considered empty by `criteria_is_empty`.

To set this new `autofail` property, `get_focused_prop` will now take
in a boolean pointer of the same name. If `__focused__` is supported
for the token and there is no focused view or the focused view does not
have a value for the token, then the boolean will be set to true. In
`parse_token`, the boolean value will be checked and if set to true,
then `criteria->autofail` will be set to true and `parse_token` will
bail successfully. Tokens will still be parsed to make sure the whole
criteria is syntactically valid, which is also why
`&criteria->autofail` is not passed to `get_focused_prop` and a local
boolean is declared in `parse_token`.

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -16,7 +16,8 @@
 #include "config.h"
 
 bool criteria_is_empty(struct criteria *criteria) {
-	return !criteria->title
+	return !criteria->autofail
+		&& !criteria->title
 		&& !criteria->shell
 		&& !criteria->app_id
 		&& !criteria->con_mark
@@ -98,6 +99,10 @@ static void find_urgent_iterator(struct sway_container *con, void *data) {
 
 static bool criteria_matches_view(struct criteria *criteria,
 		struct sway_view *view) {
+	if (criteria->autofail) {
+		return false;
+	}
+
 	if (criteria->title) {
 		const char *title = view_get_title(view);
 		if (!title || regex_cmp(title, criteria->title) != 0) {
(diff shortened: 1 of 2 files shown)


sway/criteria.c
```diff
@@ -370,46 +370,87 @@ static char *get_focused_prop(enum criteria_token token) {
...
+	// returned.
+
 	switch (token) {
 	case T_APP_ID:
-		value = view_get_app_id(view);
+		if (view) {
+			value = view_get_app_id(view);
+		}
+		if (!value) {
+			value = "\b";
```
emersion: I'm not a fan of this `\b` thing. I'd prefer a separate `bool` value.

RedSoxFan: I'll rework this and make it more elegant

RedSoxFan: Should be more elegant now

sway/criteria.c
```diff
@@ -366,50 +371,80 @@ static enum criteria_token token_from_name(char *name) {
...
 	const char *value = NULL;
 
 	switch (token) {
 	case T_APP_ID:
-		value = view_get_app_id(view);
+		if (view) {
+			value = view_get_app_id(view);
+		}
+		if (!value) {
+			*autofail = true;
```
emersion: We don't need those `if (!value)` anymore

RedSoxFan: Oh right. `parse_token` checks both the returned value and `autofail` so setting `autofail` even when there is a value is fine.

emersion:
Thanks!

Issue #3952, opened 2019-03-22
[Feature request]  Support workspace renaming during swaymsg exec handling


mliszcz:
I have a script that issues `rename workspace number <x> to '<x>:  <app_id or class>'` command upon every focus change event. It does not work well if workspace is renamed during `exec` handling (e.g. some launcher calls `swaymsg exec ...` and then exits immediately causing a focus change).

Simplest scenario that reproduces the issue:
1. In workspace "1" in terminal window type: `swaymsg exec app; swaymsg rename workspace number 1 to '1: new'`

Expected result: There is one workspace ("1: new") with two windows (terminal + app).
Actual result: There are two workspaces: "1: new" with terminal, "1" with app.

Following change in `root_workspace_for_pid` in `sway/tree/root.c` fixes the issue:
```diff
        if (pw && pw->workspace) {
                ws = workspace_by_name(pw->workspace);
+               if (!ws) {
+                   ws = workspace_by_number(pw->workspace);
+               }

                if (!ws) {
                        sway_log(SWAY_DEBUG,
                                        "Creating workspace %s for pid %d because it disappeared",                                                                           
                                        pw->workspace, pid);
                        ws = workspace_create(pw->output, pw->workspace);
                }
```

My question is: would it be possible to support described scenario? I can send a PR if you are ok with the solution proposed above.

RedSoxFan:
SGTM. Can you submit your proposed diff as a PR?

mliszcz:
Hi,

I've opened a PR (#3959), however after thinking a little more about this I'm not convinced that this is the right approach. The fix does not address a scenario where workspace changes number during rename.

In i3 following scenario always yields the same result (regardless which command completes first):

```
{ i3-msg exec chromium & } ; { i3-msg rename workspace 2 to 1 & }
```

So either `exec` is atomic or they track the the workspace by some unique id instead of a name.

Anyway, @RedSoxFan could you please help to review the PR?

PR #3959, opened 2019-03-23
Allow for workspace renaming during exec handling

mliszcz:
This change adds support for renaming workspace (without changing
workspace number) when `exec` command is being processed. If focused
workspace is renamed  after receiving `exec` command but before
the application is launched, sway will assign the application to
a workspace basing on matching number.

The change can be verified by running following command:

swaymsg exec <application>; swaymsg rename workspace number 1 to '1: new'

Fixes: #3952

### DIFF
--- a/include/sway/tree/root.h
+++ b/include/sway/tree/root.h
@@ -85,4 +85,6 @@ struct sway_container *root_find_container(
 
 void root_get_box(struct sway_root *root, struct wlr_box *box);
 
+void root_rename_pid_workspaces(const char *old_name, const char *new_name);
+
 #endif
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -390,3 +390,17 @@ void root_get_box(struct sway_root *root, struct wlr_box *box) {
 	box->width = root->width;
 	box->height = root->height;
 }
+
+void root_rename_pid_workspaces(const char *old_name, const char *new_name) {
+	if (!pid_workspaces.prev && !pid_workspaces.next) {
+		wl_list_init(&pid_workspaces);
+	}
+
+	struct pid_workspace *pw = NULL;
+	wl_list_for_each(pw, &pid_workspaces, link) {
+		if (strcmp(pw->workspace, old_name) == 0) {
+			free(pw->workspace);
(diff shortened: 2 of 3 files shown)


RedSoxFan:
From the issue:
> The fix does not address a scenario where workspace changes number during rename.
> In i3 following scenario always yields the same result (regardless which command completes first):
>    `{ i3-msg exec chromium & } ; { i3-msg rename workspace 2 to 1 & }`

Maybe `pid_workspaces` should be updated on workspace renames instead then?

include/sway/tree/root.h
```diff
@@ -85,4 +85,6 @@ struct sway_container *root_find_container(
 
 void root_get_box(struct sway_root *root, struct wlr_box *box);
 
+void root_rename_pid_workspaces(const char* old_name, const char *new_name);
```
RedSoxFan: style: `s/char\* old_name/char \*old_name/`

mliszcz:
Thanks! I've changed the code according to your suggestion. I've tested it by running:

```
swaymsg exec chromium; swaymsg rename workspace number 1 to 5
```

If you're ok with the new implementation, I'll squash the commits.

sway/tree/root.c
```diff
@@ -390,3 +390,18 @@ void root_get_box(struct sway_root *root, struct wlr_box *box) {
...
 	box->height = root->height;
 }
+
+void root_rename_pid_workspaces(const char *old_name, const char *new_name) {
+	if (!pid_workspaces.prev && !pid_workspaces.next) {
+		wl_list_init(&pid_workspaces);
+	}
+
+	struct pid_workspace *pw = NULL;
+
```
RedSoxFan: nitpick: can you just remove the blank line here and have the declaration right about the `wl_list_for_each`?

include/sway/tree/root.h
mliszcz: done

sway/tree/root.c
mliszcz: done

RedSoxFan:
Thanks!

PR #3966, opened 2019-03-25
Honor output for xdg_toplevel_set_fullscreen

### LINKED ISSUE #1813: zxdg_toplevel_v6_set_fullscreen output argument not respected
If a program wants to go fullscreen on a specific output it can set it explicitly via an argument of zxdg_toplevel_v6_set_fullscreen. This currently does nothing under sway.
Can be replicated with 2 outputs and mpv: mpv <file> --fs --fs-screen=X --gpu-context=wayland

RedSoxFan:
Fixes #1813 
Requires swaywm/wlroots#1638

This honors the fullscreen output request for
`xdg_toplevel_set_fullscreen` and `zxdg_toplevel_v6_set_fullscreen`.

If the request was sent before mapping, the fullscreen output request
will be retrieved from the client_pending state for the toplevel. The
output will be passed to `view_map` and if there is a workspace on the
output, the view will be placed on that workspace.

If the request comes in after being mapped, the view will be moved to
the workspace on the output (if there is one) before becoming
fullscreen.

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -314,7 +314,7 @@ void view_destroy(struct sway_view *view);
 void view_begin_destroy(struct sway_view *view);
 
 void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
-	bool fullscreen, bool decoration);
+	bool fullscreen, struct wlr_output *fullscreen_output, bool decoration);
 
 void view_unmap(struct sway_view *view);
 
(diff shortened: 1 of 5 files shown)


RedSoxFan:
Rebased and ready for review

RedSoxFan:
If anyone gets some time, this is waiting on a review. This doesn't do anything important though so no rush

ddevault:
Thanks!

PR #3967, opened 2019-03-25
Implement seat subcommand hide_cursor_while_typing

RedSoxFan:
Related to #1894

This implements the following seat subcommand:
  `hide_cursor_while_typing <timeout>`

If timeout is set to a positive value (of at least 100), the cursor
will be hidden when typing and will not be unhidden until the timeout
has expired after the last key has been released.

This also changes the way cursor hiding/unhiding is managed to allow
for both `hide_cursor` and `hide_cursor_while_typing` to be used in
conjunction with each other for complex cursor hiding rules.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -269,6 +269,7 @@ sway_cmd seat_cmd_attach;
 sway_cmd seat_cmd_cursor;
 sway_cmd seat_cmd_fallback;
 sway_cmd seat_cmd_hide_cursor;
+sway_cmd seat_cmd_hide_cursor_while_typing;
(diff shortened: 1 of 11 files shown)


ddevault:
I know I was the one who originally asked for this, but I'm -1. I think that `hide_cursor` is good enough and we needn't introduce the complexity of the other two features.

RedSoxFan:
> I know I was the one who originally asked for this, but I'm -1. I think that `hide_cursor` is good enough and we needn't introduce the complexity of the other two features.

I'll close this and the corresponding issue then.

PR #3969, opened 2019-03-25
Implement handling of short_text field of i3 input protocol.

DmitriK:
Matches i3bar behavior of setting all blocks to use the short_text if the full
text width does not fit.


ddevault:
Any sway users who take advantage of this feature who can share their test results?

tarmack:
I want to take advantage of this feature. I tested this pull request but I can't get the `short_text` value to be displayed.

I use Pango markup, including different font sizes (only smaller than the configured font) and I also have quite a few "icons" in my bar. Maybe these are confusing the space estimations?

tarmack:
This works wonderfully now. Thank you!

ddevault:
Thanks!

PR #3977, opened 2019-03-26
Fix order mismatch in interpolated string arguments

jnpkrn:
Spotted just thanks to the same-message-different-sink that follows.

### DIFF
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -316,7 +316,7 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 		struct sway_binding *config_binding = mode_bindings->items[i];
 		if (binding_key_compare(binding, config_binding)) {
 			sway_log(SWAY_INFO, "Overwriting binding '%s' for device '%s' "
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Would you mind fixing the same mismatch in the bindswitch function as well and then force pushing? Both changes can be a single commit, preferably with a short commit message along the lines of `bindings: fix overwrite log argument mismatch` so that it is easy to tell what was changed

jnpkrn:
@RedSoxFan, thanks, done.
I'm used to somewhat uniform commit summaries, but couldn't spot any
pre-existing pattern and didn't want to appear too exotic :-)

RedSoxFan:
> I'm used to somewhat uniform commit summaries, but couldn't spot any
pre-existing pattern and didn't want to appear too exotic :-)

We don't require a standardized pattern, just anything that gets the information across (although I do like the somewhat common `system/function: what changed` pattern for that). My comment was more about getting as much information as possible into the short commit message as your original one could have referenced an change anywhere in the code base.

RedSoxFan:
Thanks!

Issue #3972, opened 2019-03-25
[bug] sway consistently crashes if the song changes in mpv while the window is hidden in the scratchpad


norcalli:
* Sway Version:

```json
{
   "human_readable": "1.0-rc1-200-gd9de5b87 (Mar 25 2019, branch 'master')",
   "variant": "sway",
   "major": 1,
   "minor": 0,
   "patch": 0,
   "loaded_config_file_name": "\/home\/ashkan\/.config\/sway\/config"
 }
```

* Steps to reproduce:

*Consistently*, if I have `mpv` playing in a scratchpad **and hidden** and the song changes, it will crash sway.

* Stack Trace, if sway crashes:

```
(gdb) bt full 
#0  0x000056315a5452e8 in container_init_floating (con=0x56315b7c8000) at ../sway/sway/tree/container.c:655 
        ws = 0x0 
        ob = 0x7fd460a7d550 
        min_width = -517721104 
        max_width = 32766 
        min_height = 1542126080 
        max_height = -2008805100 
#1  0x000056315a5518bb in handle_request_configure (listener=0x56315b7c7050, data=0x7ffee12434f0) at ../sway/sway/desktop/xwayland.c:438 
        xwayland_view = 0x56315b7c6f00 
        ev = 0x7ffee12434f0 
        view = 0x56315b7c6f00 
        xsurface = 0x56315b7c6cb0 
#2  0x00007fd460a38ede in wlr_signal_emit_safe (signal=signal@entry=0x56315b7c6db8, data=data@entry=0x7ffee12434f0) at ../util/signal.c:29 
        pos = 0x56315b7c7050 
        l = 0x56315b7c7050 
        cursor = {link = {prev = 0x56315b7c7050, next = 0x7ffee1243480}, notify = 0x7fd460a38e50 <handle_noop>} 
        end = {link = {prev = 0x7ffee1243460, next = 0x56315b7c6db8}, notify = 0x7fd460a38e50 <handle_noop>} 
#3  0x00007fd4609f7a49 in xwm_handle_configure_request (ev=0x56315b82b830, xwm=0x56315b753790) at ../xwayland/xwm.c:878 
        surface = 0x56315b7c6cb0 
        mask = <optimized out> 
        geo_mask = 15 
        wlr_event = {surface = 0x56315b7c6cb0, x = 1407, y = 587, width = 960, height = 480} 
        count = 1 
        event = 0x56315b82b830 
        xwm = 0x56315b753790 
[...]

PR #3978, opened 2019-03-27
Fix xwayland configure request scratchpad crash

RedSoxFan:
Fixes https://github.com/swaywm/sway/issues/3972#issue-425142958

This fixes a crash in `container_init_floating` when a xwayland view
sends a configure request while in the scratchpad.

`container_init_floating` gets called so the configured minimum and
maximum sizes gets respected when resizing to the requested size. Since
the workspace was NULL, it would SIGSEGV when attempting to get the
workspace's output for the output box retrieval.

This extracts the resizing portion of `container_init_floating` into a
separate function. If the container is in the scratchpad, it will just
be resized and skip the centering

Additionally, `container_init_floating` has been renamed to
`container_floating_resize_and_center` to more accurately describe what
it does.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -215,10 +215,7 @@ size_t container_titlebar_height(void);
 void floating_calculate_constraints(int *min_width, int *max_width,
 		int *min_height, int *max_height);
 
-/**
- * Resize and center the container in its workspace.
- */
-void container_init_floating(struct sway_container *container);
+void container_floating_resize_and_center(struct sway_container *con);
 
 void container_set_floating(struct sway_container *container, bool enable);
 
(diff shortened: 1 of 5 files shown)


RyanDwyer:
It seems odd to me that xwayland's `handle_request_configure` calls `container_init_floating`. `container_init_floating` is not intended to be called on a container that is already floating, and makes a rightful assumption that the container is not in the scratchpad (because it would be floating already if it was).

Rather than make `container_init_floating` handle this situation, I think a better option would be to make it so xwayland's `handle_request_configure` doesn't call it.

RedSoxFan:
Since `container_init_floating` just makes sure the size conforms to the min/max restrictions and then center's it, I'd rather just rename the function (I have to think of a good name). It's already used in a few locations for that purpose. All of the actual setup for floating containers is done in `container_set_floating`.

RyanDwyer:
Sounds good. I'll let you come up with a name.

RedSoxFan:
I renamed `container_init_floating` to `container_floating_resize_and_center` to more accurately describe what it does

emersion:
Thanks!

PR #3979, opened 2019-03-27
Fix scratchpad fullscreen behavior and crash

RedSoxFan:
Fixes https://github.com/swaywm/sway/issues/3972#issuecomment-476576069

When setting fullscreen on a hidden scratchpad container, there was a
check to see if there was an existing fullscreen container on the
workspace so it could be fullscreen disabled first. Since the workspace
is NULL, it would cause a SIGSEGV. This adds a NULL check to avoid the
crash.

This also changes the behavior of how fullscreen is handled when adding
a container to the scratchpad or changing visibility of a scratchpad
container to match i3's. The behavior is as follows:
- When adding a container to the scratchpad or hiding a container back
  into the scratchpad, there is an implicit fullscreen disable
- When setting fullscreen on a container that is hidden in the
  scratchpad, it will be fullscreen when shown (and fullscreen disabled
  when hidden as stated above)
- When setting fullscreen global on a container that is hidden in the
  scratchpad, it will be shown immediately as fullscreen global. The
  container is not moved to a workspace and remains in the
  scratchpad. The container will be visible until fullscreen disabled
  or killed. Since the container is in the scratchpad, running
  `scratchpad show` or `move container to scratchpad` will have no
  effect

This also changes `container_replace` to transfer fullscreen and
scratchpad status.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -197,10 +197,11 @@ static bool gaps_to_edge(struct sway_view *view) {
 
 void view_autoconfigure(struct sway_view *view) {
 	struct sway_container *con = view->container;
-	if (container_is_scratchpad_hidden(con)) {
+	if (container_is_scratchpad_hidden(con) &&
+			con->fullscreen_mode != FULLSCREEN_GLOBAL) {
 		return;
 	}
-	struct sway_output *output = con->workspace->output;
+	struct sway_output *output = con->workspace ? con->workspace->output : NULL;
 
 	if (con->fullscreen_mode == FULLSCREEN_WORKSPACE) {
 		con->content_x = output->lx;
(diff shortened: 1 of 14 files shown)


RyanDwyer:
Would it be better to forbid toggling fullscreen on containers that are hidden in the scratchpad? The check would also have to be done when the view requests fullscreen.

RedSoxFan:
Just checked i3's behavior:
- fullscreen on a hidden scratchpad container will cause the container to be fullscreened when shown
- fullscreen global on a hidden scratchpad container will immediately cause the container to be shown as fullscreen global
  - If this happens, the container cannot be moved to the scratchpad
  - It will be moved back to the scratchpad on a fullscreen disable
- When moving a fullscreen (workspace or global) container to the scratchpad, the fullscreen status will be removed

I guess I'll change this to replicate i3's behavior

RedSoxFan:
Ok. I think I _finally_ got everything and this is ready for review.

RedSoxFan:
If anyone gets some time, this is waiting on a review

sway/tree/view.c
```diff
@@ -247,7 +250,8 @@ void view_autoconfigure(struct sway_view *view) {
 	// title area. We have to offset the surface y by the height of the title,
 	// bar, and disable any top border because we'll always have the title bar.
 	list_t *siblings = container_get_siblings(con);
```
norcalli: siblings can be null

ddevault:
Good for merge once the merge conflicts are resolved.

Since we only have to fix this once (fingers crossed) I'm not too worried about the long-term maintenance cost of matching i3's complicated behavior. If it becomes a problem later we can rethink it.

RedSoxFan:
> Good for merge once the merge conflicts are resolved.

Rebased and resolved merge conflicts

ddevault:
Thanks!

PR #3997, opened 2019-03-30
Fix a crash in swaybar when an icon dir is not readable

### LINKED ISSUE #3981: Swaybar segfault when /usr/local/share/icons inaccessible
Version: 1.0 (1.0-7 in Arch Linux)

This is a minor and obtuse one that i thought i'd flag up.

I had a problem with swaybar always segfaulting which i tracked down to /usr/local/share/icons existing, being root owned, and only readable by root. I can fix it on my machine by simply deleting everything in /usr/local/share (it's not really used on my system). However, it might be worth knowing about in case others experience similar.

In particular, it is the call to opendir in swaybar/tray/icon.c that fails
```
static list_t *load_themes_in_dir(char *basedir) {
    DIR *dir;
    if (!(dir = opendir(basedir))) {
        ....
```
It still fails if i replace the basedir variable with the literal string "/usr/local/share/icons".

I'm not sure what's going on here, because it shouldn't segfault. I tried a simple small program
```
#include <dirent.h>
#include <stdio.h>

int main (int argc, char **argv) {
    if (!opendir("/usr/local/share/icons")) {
        printf("Not open!\n");
    }
    return 0;
}
```
This runs fine (printing "Not open!") when compiled with both clang and gcc.

ForTheReallys:
Fixes #3981.

### DIFF
--- a/swaybar/tray/icon.c
+++ b/swaybar/tray/icon.c
@@ -348,6 +348,9 @@ void init_themes(list_t **themes, list_t **basedirs) {
 	*themes = create_list();
 	for (int i = 0; i < (*basedirs)->length; ++i) {
 		list_t *dir_themes = load_themes_in_dir((*basedirs)->items[i]);
+		if (dir_themes == NULL) {
+			continue;
+		}
 		list_cat(*themes, dir_themes);
 		list_free(dir_themes);
 	}


emersion:
Thanks!

PR #4015, opened 2019-04-03
Add deprecation warnings for new_float, new_window, and force_focus_wrapping.

c-edw:
To my knowledge, these are the only deprecated i3 commands that Sway has. 
`new_window` is not compatible with i3's syntax but it doesn't matter too much if it's due to be removed in future.


sway/commands/new_float.c
```diff
@@ -0,0 +1,7 @@
+#include "log.h"
+#include "sway/commands.h"
+
+struct cmd_results *cmd_new_float(int argc, char **argv) {
+	sway_log(SWAY_INFO, "Warning: new_float is deprecated");
```
ddevault: This should include instructions on how to fix it (e.g. "Warning: new_float is deprecated. Use floating_border instead.")

sway/meson.build
```diff
@@ -72,6 +72,8 @@ sway_sources = files(
 	'commands/layout.c',
 	'commands/mode.c',
 	'commands/mouse_warping.c',
+	'commands/new_float.c',
+	'commands/new_window.c',
```
RyanDwyer: Can you please put these below move.c so they're alphabetical?

include/sway/commands.h
```diff
@@ -146,6 +146,8 @@ sway_cmd cmd_log_colors;
 sway_cmd cmd_mark;
 sway_cmd cmd_mode;
 sway_cmd cmd_mouse_warping;
+sway_cmd cmd_new_float;
```
RedSoxFan: Can you move these to be in alphabetically order too?

sway/commands/new_float.c
```diff
@@ -0,0 +1,7 @@
+#include "log.h"
+#include "sway/commands.h"
+
+struct cmd_results *cmd_new_float(int argc, char **argv) {
+	sway_log(SWAY_INFO, "Warning: new_float is deprecated. Use default_floating_border instead.");
```
RedSoxFan: It is probably worth adding the following:
```c
if (config->reading) {
    config_add_swaynag_warning("new_float is deprecated. Use "
            "default_floating_border instead");
}
```

This will show a swaynag warning when included in the config, which is likely more visible to the user.

RedSoxFan: style: 80 column limit

sway/commands/new_window.c
```diff
@@ -0,0 +1,7 @@
+#include "log.h"
+#include "sway/commands.h"
+
+struct cmd_results *cmd_new_window(int argc, char **argv) {
+	sway_log(SWAY_INFO, "Warning: new_window is deprecated. Use default_border instead.");
```
RedSoxFan: style: 80 column limit

RedSoxFan: The swaynag warning suggestion also applies here

c-edw:
I've added a warning for `force_focus_wrapping` too, but not sure whether that was the right move. In the manpages, it says it's there for convenience. It doesn't explicitly say that it's deprecated in i3's docs either.

sway/commands/force_focus_wrapping.c
```diff
@@ -1,20 +1,12 @@
...
-	if (error) {
-		return error;
+	sway_log(SWAY_INFO, "Warning: force_focus_wrapping is deprecated. "
+		"Use focus_wrapping instead.");
+	if (config->reading) {
+		config_add_swaynag_warning("force_focus_wrapping is deprecated. "
+			"Use focus_wrapping instead.");
 	}
-
-	if (parse_boolean(argv[0], config->focus_wrapping == WRAP_FORCE)) {
```
RedSoxFan: Even with the deprecation warning, this block should be retained. The arguments for `force_focus_wrapping` are different then`focus_wrapping`.

c-edw:
`Legacy` to me means *discouraged but supported*. `deprecated` to me means *supported but due to be removed*. There is nothing in i3's docs that indicates legacy syntaxes are going to be removed in future.

Similar to the valid point you raised about `force_focus_wrapping` having different arguments, `new_window` and `new_float` on i3 can be used in the legacy syntax `[COMMAND] 1pixel`, but this isn't supported on Sway. For users, there might cause some confusion if switching to the newer counterpart of one of these commands whilst still using this legacy syntax. 

What do you think about just having the deprecation warnings without calling the newer alternative commands? 

This change overall could require more effort on the users' part to get their i3 config working smoothly with Sway.

emersion:
>Legacy to me means discouraged but supported. deprecated to me means supported but due to be removed. There is nothing in i3's docs that indicates legacy syntaxes are going to be removed in future.

I don't get the reasoning and I agree with @RedSoxFan. What's the point of marking a command as legacy if it's not going to be removed at some point?

RedSoxFan:
> Legacy to me means discouraged but supported

Under that definition, Sway should support the `1pixel` syntax for `border`, `new_window`, and `new_float` for compatibility reasons.

> What do you think about just having the deprecation warnings without calling the newer alternative commands?

To be fair, these commands are relatively trivial to support. `new_window` and `new_float` are just aliases (ignoring support for the legacy `1pixel` syntax mentioned above) and `force_focus_wrapping` just has different mappings for the arguments.

Either way, having `force_focus_wrapping` be an alias for `focus_wrapping` is not the correct solution since the arguments have different meanings.

c-edw:
Sorry for the delay on this. It's ready for review again.

sway/commands/force_focus_wrapping.c
```diff
@@ -1,9 +1,15 @@
-#include <strings.h>
```
RedSoxFan: Oops. Can you re-add these headers?

c-edw: Have I misunderstood something? They don't appear to be necessary here.

RedSoxFan: Looks like `<strings.h>` isn't needed, but `sway/config.h` is needed for `config_add_swaynag_warning`

c-edw: I see, thanks for the explanation.

RedSoxFan:
Thanks!

PR #4035, opened 2019-04-11
cmd_focus: raise floating for `<criteria> focus`

### LINKED ISSUE #4031: swaymsg: focus does not raise floating window
Focusing a floating window using swaymsg does not raise that window to the front.

I notice there's a recent change #3943 which raises the floating window if left/right/up/down is given. That works, but what about the case of just...

    swaymsg <criteria> focus

This focuses the window such that it receives keyboard input, but it is not brought to the front. The "focused" floating window can therefore still be obscured behind another floating window.

Using `sway version 1.0-rc1-222-ge0324fc8 (Apr  9 2019, branch 'master')`

RedSoxFan:
Closes #4031 

Floaters are currently raised for `focus <direction>`. This extends the
same functionality to `<criteria> focus`.

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -317,6 +317,7 @@ struct cmd_results *cmd_focus(int argc, char **argv) {
 		}
 		seat_set_focus_container(seat, container);
 		seat_consider_warp_to_focus(seat);
+		container_raise_floating(container);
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #3992, opened 2019-03-29
Fix sizing of floaters + scratchpad

RedSoxFan:
Fixes #3991 
Fixes #3935 

This fixes the sizing of floating non-view containers. On master, the
floater will get set to the maximum width and height, which by default
is the entire output layout. When setting a non-view container to
floating, this will set a sane default size of 50% of the workspace
width and 75% of the workspace height, or whatever the closest is that
the minimum and maximum floating width/height values allow for. On all
future calls to `floating_natural_resize`, the width and height will be
kept unless they need to be changed to respect the min/max floating
width/height values.

This also matches i3's behavior of setting scratchpad containers to 50% of
the workspace's width and 75% of the workspace's height, bound by the
minimum and maximum floating width/height.

PR #4037, opened 2019-04-11
Fix scratchpad logic for floating windows 

norcalli:
Only sets an initial size if the container does *not* have a view. This
allows initial sizing geometry to be respected.

This is a follow up to https://github.com/swaywm/sway/pull/3992, wherein I believe the original intent was to only set a default size for containers without a view:

> When setting a non-view container to
floating, this will set a sane default size of 50% of the workspace
width and 75% of the workspace height, or whatever the closest is that
the minimum and maximum floating width/height values allow for.

### DIFF
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -61,9 +61,14 @@ void root_scratchpad_add_container(struct sway_container *con) {
 
 	struct sway_container *parent = con->parent;
 	struct sway_workspace *workspace = con->workspace;
-	container_set_floating(con, true);
-	container_floating_set_default_size(con);
-	container_floating_move_to_center(con);
+
+	// When a tiled window is sent to scratchpad, center and resize it.
+	if (!container_is_floating(con)) {
+		container_set_floating(con, true);
+		container_floating_set_default_size(con);
+		container_floating_move_to_center(con);
+	}
+
 	container_detach(con);
 	con->scratchpad = true;
 	list_add(root->scratchpad, con);


RedSoxFan:
This was to set the initial size of scratchpad containers. When a container is made floating, it is resized with `container_floating_resize_and_center` immediately after the call to `container_floating_set_default_size` and this should not have any effect for container with views.

`container_floating_resize_and_center` calls `floating_natural_resize` which uses the `natural_width` and `natural_height` for the sizing. These are set when the application initially maps and for configure requests for floating xwayland views.

From https://github.com/swaywm/sway/pull/3992#issuecomment-482345195
> I'm not sure what the equivalent i3 behavior is supposed to be for floating scratchpad containers with a predefined size.

Here is the i3 code (it would be awesome if GitHub allowed cross repo snippets...):
https://github.com/i3/i3/blob/a574346964ce17d35df2ef9072ba16e26209ac84/src/scratchpad.c#L197

> I'm not sure what a container not having a view would mean

Containers without views are used for layout purposes. The layout is a tree where the leaves have views and the branches do not.

> I forgot to mention that if I create a window initially in tiling mode and then make it floating, it doesn't share this problem, in that it respects the initial geometry that I had specified (e.g. with st -g 150x30).

I tried adding `for_window [class="st"] floating enable` and running `st -g 150x30`, but it seems to be respecting the size specified for me. I can't seem to find a case where this fails. Can you please provide reproduction steps for the issue you are experiencing?

norcalli:
@RedSoxFan A picture is worth a thousand words, so here's a demo: https://static.kiani.io/vid-2019-04-11-201707.mp4

I have the master build which I demo first and then the build with my patch.

Just `floating enable` isn't sufficient to reproduce, a minimal config is `for_window [class="st"] floating enable, move scratchpad`, which I launched with:

```sh
./build/sway/sway -c (echo 'for_window [class="st"] floating enable, move scratchpad' | psub)
```

Then I create an `st` instance using `swaymsg` with 

```sh
swaymsg -s /run/user/1000/sway-ipc.1000.130160.sock   -- exec st -g 10x10 pulsemixer, scratchpad show
```

Two calls to that are required to get it to show up, but you can see that it starts with either 10x10 with my patch, or 50%x75% without.

RedSoxFan:
> a minimal config is `for_window [class="st"] floating enable, move scratchpad`

Ok so I figured out the issue. The i3 snippet only occurs when moving from tiled to the scratchpad, not from floating to the scratchpad. The correct fix should be to change the logic in `root_scratchpad_add_container`. Instead of unconditionally calling `container_set_floating`, `container_floating_set_default_size`, and `container_floating_move_to_center`, they should be wrapped in `if (!container_is_floating(con)) {`.

norcalli:
@RedSoxFan I tested your suggestion, and it worked in my testing, so I've updated the PR with the suggested fix.

I tested my initial scenario and additionally sending a tiled window to scratchpad, and it matched our expectations.

sway/tree/root.c
```diff
@@ -62,8 +62,13 @@ void root_scratchpad_add_container(struct sway_container *con) {
 	struct sway_container *parent = con->parent;
 	struct sway_workspace *workspace = con->workspace;
 	container_set_floating(con, true);
-	container_floating_set_default_size(con);
-	container_floating_move_to_center(con);
+
+	// When a tiled window is sent to scratchpad, center and resize it.
+	if (!container_is_floating(con)) {
```
RedSoxFan: This will always be true since it is being set to floating right before the check. The `container_set_floating` call needs to be moved inside.

norcalli: oops, I guess that's what I get for being tired and coding.

RedSoxFan: LGTM as soon as this change is pushed

norcalli:
squashed/force pushed change

RedSoxFan:
Thanks!

PR #4041, opened 2019-04-12
sway-input.5: add xkeyboard-config(7) to "see also"

tbm:
Since xkeyboard-config(7) is mentioned in the man page, add a reference
to the "see also" section.


emersion:
Thanks!

PR #4048, opened 2019-04-13
Fix potential null accesses

norcalli:
I recommend viewing this with whitespace ignored. I simply wrapped some branches in `if`s where the subject could be null. I ran into all of these while doing some testing.


ddevault:
Thanks!

PR #4049, opened 2019-04-13
Spawn swaynag as a wayland client

### LINKED ISSUE #4047: Reloading sway should clear error messages
I started sway and it showed an error bar saying that my config file has an error.  I edited the file, fixed the error and reloaded sway but the error message was still there.

At it turns out, sway doesn't remove existing error messages upon reload.  I think i3 has the same behaviour.  To me this seems strange and confusing, but maybe there's a good reason for it.

RedSoxFan:
**Until #4051 is merged, I would recommend rebasing this on top of it for testing. Otherwise, pressing the "Reload sway" button may result in sway reloading multiple times causing a noticeable delay/unresponsive period. Reloading via a binding or command line is not affected by the multiple reloads issue**

Fixes #4047 

This spawns swaynag as a wayland client similar to how swaybar and
swaybg are already done

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -1,5 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include <float.h>
+#include <fcntl.h>
 #include <math.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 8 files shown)


RedSoxFan:
Ready for review

emersion:
Thanks!

PR #4051, opened 2019-04-14
swaynag: fix pointer management

RedSoxFan:
Currently on master, swaynag will retrieve a pointer instance whenever
the capabilities change and WL_SEAT_CAPBILITY_POINTER is set. The
pointer instances were never being destroyed so swaynag received events
multiple times due to having several instances of the pointer.

This fixes it so if there is already a pointer instance, swaynag does
not attempt to retrieve another. Additionally, if the pointer
capability is removed, the pointer instance is destroyed.


emersion:
Thanks!

PR #4054, opened 2019-04-15
Fix the payload type returned by IPC

### LINKED ISSUE #4053: [bug] IPC returns incorrect payload type when issuing command while subscribed
`sway version 1.0-rc1-234-gb2cfcebe (Apr 14 2019, branch 'ashkan')`

I have a script which listens for a particular event using `subscribe` and then issues a command via `RUN_COMMAND`. The result for the `RUN_COMMAND` is given the type of `0x80000003`, which is the window event type (the type I subscribed to).

I was using the payload type to try to distinguish between results, and was wondering why my program seemed stuck, and it's because the result type isn't distinguished properly. I would expect it to be `0`.

norcalli:
If a client is subscribed and sends a subsequent ipc command which
causes event updates, then those event updates override the
`client->current_command` and send the incorrect type for the payload
associated with the command.

Example:

SUBSCRIBE {window}
RUN_COMMAND focus -> PAYLOAD_TYPE is 0x80000002 for window events

Therefore, we decouple the `client->current_command` by passing it as an
argument to the ipc_send_reply function, avoiding a data race. The same
is done for the `client->payload_length` as a precautionary measure for
the same reason.

Closes #4053

### DIFF
--- a/sway/ipc-server.c
+++ b/sway/ipc-server.c
@@ -47,13 +47,14 @@ struct ipc_client {
 	struct wl_event_source *writable_event_source;
 	struct sway_server *server;
 	int fd;
-	uint32_t payload_length;
 	uint32_t security_policy;
-	enum ipc_command_type current_command;
 	enum ipc_command_type subscribed_events;
 	size_t write_buffer_len;
 	size_t write_buffer_size;
 	char *write_buffer;
+	// The following are for storing data between event_loop calls
+	uint32_t pending_length;
+	enum ipc_command_type pending_type;
 };
 
 struct sockaddr_un *ipc_user_sockaddr(void);
@@ -61,8 +62,10 @@ int ipc_handle_connection(int fd, uint32_t mask, void *data);
 int ipc_client_handle_readable(int client_fd, uint32_t mask, void *data);
 int ipc_client_handle_writable(int client_fd, uint32_t mask, void *data);
 void ipc_client_disconnect(struct ipc_client *client);
-void ipc_client_handle_command(struct ipc_client *client);
-bool ipc_send_reply(struct ipc_client *client, const char *payload, uint32_t payload_length);
+void ipc_client_handle_command(struct ipc_client *client, uint32_t payload_length,
+	enum ipc_command_type payload_type);
+bool ipc_send_reply(struct ipc_client *client, enum ipc_command_type payload_type,
+	const char *payload, uint32_t payload_length);
 
 static void handle_display_destroy(struct wl_listener *listener, void *data) {
 	if (ipc_event_source) {
@@ -178,7 +181,7 @@ int ipc_handle_connection(int fd, uint32_t mask, void *data) {
 		return 0;
 	}
 	client->server = server;
-	client->payload_length = 0;
+	client->pending_length = 0;
 	client->fd = client_fd;
 	client->subscribed_events = 0;
 	client->event_source = wl_event_loop_add_fd(server->wl_event_loop,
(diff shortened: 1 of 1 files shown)


norcalli:
Formatting updated btw

sway/ipc-server.c
```diff
@@ -226,7 +232,12 @@ int ipc_client_handle_readable(int client_fd, uint32_t mask, void *data) {
 	// Wait for the rest of the command payload in case the header has already been read
 	if (client->payload_length > 0) {
 		if ((uint32_t)read_available >= client->payload_length) {
-			ipc_client_handle_command(client);
+			// Pass the payload_length and payload_type so that
+			// they are semantically and logically decoupled from
+			// the logic pertaining to the values stored in client.
+			ipc_client_handle_command(client,
```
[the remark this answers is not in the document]
norcalli: Unless it's inaccurate, then I'm not sure what you mean by that.

sway/ipc-server.c
```diff
@@ -657,7 +675,7 @@ void ipc_client_handle_command(struct ipc_client *client) {
 		}
 		const char *json_string = json_object_to_json_string(outputs);
 		client_valid =
-			ipc_send_reply(client, json_string, (uint32_t)strlen(json_string));
+			ipc_send_reply(client, payload_type, json_string, (uint32_t)strlen(json_string));
```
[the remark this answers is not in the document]
norcalli: I swear I thought I fixed that. My bad.

sway/ipc-server.c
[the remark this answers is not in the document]
RedSoxFan: I just don't think the comment adds anything of value. The comment in the struct declaration should be sufficient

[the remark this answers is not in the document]
RedSoxFan: > I wanted to explain to people in the future why I decided to pass in the arguments rather than the way it was before. Unless you're referring to the duplication of the comment, but I would rather be verbose than the opposite.

Can you try rewording it in a way that provides some reasoning then? If I just started looking at the code and came across that comment, I would question why the property is part of the struct.

Maybe the properties should also be renamed to something like `pending_length` and `pending_type`?

sway/ipc-server.c
```diff
@@ -843,42 +851,43 @@ void ipc_client_handle_command(struct ipc_client *client) {
...
 	default:
-		sway_log(SWAY_INFO, "Unknown IPC command type %i", client->current_command);
+		sway_log(SWAY_INFO, "Unknown IPC command type %x", payload_type);
 		goto exit_cleanup;
 	}
 
 exit_cleanup:
 	if (client_valid) {
-		client->payload_length = 0;
+		client->pending_length = 0;
```
RedSoxFan: What do you think about changing `ipc_client_handle_command` to return a success boolean and having this reset in `ipc_client_handle_readable` after the `ipc_client_handle_command` calls? That would make it so the pending properties are only touched in `ipc_client_handle_readable` (aside from the initialization in `ipc_handle_connection` anyway).

norcalli: I was trying to be conservative in the number of changes since this is a critical bug for my utilities, and I didn't want to refactor and potentially introduce bugs. That change does look fine, though. I'm not sure I understand the reason for the code, though.

RedSoxFan: If `ipc_send_reply` fails, it will disconnect and free `client`. Once that happens, `client` should not be used anymore

norcalli: In that case, let's move the assignment to 0 to the start of the function after asserting that client is not null, since there's no reason to wait. We've used the payload_length already, so it should be set to 0 afterward. Technically, we should set `pending_type` to an empty value as well.

RedSoxFan: Could also store it in a local temp variable in `ipc_client_handle_readable`, reset it before the call, and pass the local temp variable

norcalli: And with that, we can drop `client_valid` too, since it's not used.

norcalli: @RedSoxFan yup, considered that too. If I think of `handle_command` like a flush, then I don't think that setting `pending_length` to 0 there is bad practice. At worst, it should be redundant.

norcalli: And for a final cleanup, casting `strlen` to `uint32_t` is inconsistent, which indicates to me that it isn't strictly required (unless it's some platform dependent fix I'm not aware of)

RedSoxFan: I'm still not a huge fan of `client->pending_length` being altered in `handle_command` after the rest of the changes this PR makes. This PR removes all other accesses of it in the function and passes the payload length so that it does not use the property. Having `client->pending_length` modified in the function just seems to go against the rest of the PR

norcalli: @RedSoxFan changed

RedSoxFan:
Thanks!

PR #4062, opened 2019-04-16
swaymsg: add timeout and type checks

### LINKED ISSUE #4046: Better behaviour when run on X11
This is obviously a very minor issue, but would it make sense to print an error when swaymsg (and possibly other commands) are run on X11 instead of Wayland?

Currently I get:

```
$ swaymsg -t get_outputs
Output eDP-1 '(null) (null) (null)'
  Current mode: 0x0 @ 0.000000 Hz
  Position: 0,0
  Scale factor: 0.000000
  Transform: (null)
  Workspace: 4
swaymsg: json_object.c:1214: json_object_array_length: Assertion `json_object_get_type(jso) == json_type_array' failed.
```

and `swaymsg -t get_inputs` just hangs.

RedSoxFan:
*Disclaimer: I'm not really sure of the value of this, but it's not that large*

Closes #4046 

This adds a 3 second timeout to the initial reply in swaymsg. This
prevents swaymsg from hanging when `swaymsg -t get_{inputs,seats}` is
used in i3. The timeout is removed when waiting for a subscribed event
or monitoring for subscribed events.

This also adds type checks to commands where i3 does not reply with all
of the properties that sway does (such as `modes` in `get_outputs`).

This is mostly just a behavioral adjustment since swaymsg should run on
i3. When running under i3, some command reply's (such as the one for
`get_outputs) may have more useful information in the raw json than the
pretty printed version.

### DIFF
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -69,6 +69,16 @@ int ipc_open_socket(const char *socket_path) {
 	return socketfd;
 }
 
+bool ipc_set_recv_timeout(int socketfd, struct timeval tv) {
+	if (setsockopt(socketfd, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv)) == -1) {
+		sway_log_errno(SWAY_ERROR, "Failed to set ipc recv timeout");
+		return false;
+	}
+	sway_log(SWAY_DEBUG, "ipc recv timeout set to %ld.%06ld",
+			tv.tv_sec, tv.tv_usec);
+	return true;
+}
+
 struct ipc_response *ipc_recv_response(int socketfd) {
 	char data[IPC_HEADER_SIZE];
 	uint32_t *data32 = (uint32_t *)(data + sizeof(ipc_magic));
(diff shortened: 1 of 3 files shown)


emersion:
It's probably a good idea to have an I/O timeout anyway

common/ipc-client.c
```diff
@@ -69,6 +69,18 @@ int ipc_open_socket(const char *socket_path) {
 	return socketfd;
 }
 
+bool ipc_set_recv_timeout(int socketfd, struct timeval tv) {
+	int result = setsockopt(socketfd, SOL_SOCKET, SO_RCVTIMEO,
+			(const char*)&tv, sizeof(tv));
```
emersion: `man 3p setsockopt` says the value is a `const void *`, so I don't think a cast is necessary.

RedSoxFan: Removed cast. Sorry about that, I searched for how to set a timeout. Apparently Windows has the argument as `const char *` instead of `const void *` like every other platform so I was seeing the snippet with cast for compatibility. I forgot to check if the cast was actually needed.

emersion:
Thanks!

PR #4067, opened 2019-04-18
hide_edge_borders.c: add missing case for --i3

mwenzkowski:
Disable the i3-compatible behavior if the option '--i3' is not given.
Previously it was only possible to disable it by changing the config
file. Now it also works via swaymsg.


emersion:
Thanks!

Issue #3793, opened 2019-02-28
spotify escapes workspace assignment


CaptainBloodz:
A couple of apps assigned to workspaces here:
assign [class="Mixxx"] 6
assign [class="Carla"] 7
assign [class="Audacity"] 8
assign [class="Spotify"] 7
assign [class="Firefox"] 9
assign [class="Cadence"] 10

among which Spotify doesn't work.
I'm not 100% sure, but AKAIR it has been working recently,

emersion:
Can you share the output of `swaymsg -t get_tree` when Spotify is running?

Also a debug log.

CaptainBloodz:
swaymsg -t get_tree
{
   "id": 1,
   "name": "root",
   "rect": {
     "x": 1920,
     "y": 0,
     "width": 1920,
     "height": 1080
   },
   "focused": false,
   "focus": [
     26
   ],
   "border": "none",
   "current_border_width": 0,
   "layout": "splith",
   "orientation": "horizontal",
   "percent": null,
   "window_rect": {
     "x": 0,
     "y": 0,
     "width": 0,
     "height": 0
   },
   "deco_rect": {
     "x": 0,
     "y": 0,
     "width": 0,
     "height": 0
   },
   "geometry": {
     "x": 0,
     "y": 0,
     "width": 0,
     "height": 0
   },
   "window": null,
   "urgent": false,
   "floating_nodes": [
   ],
   "sticky": false,
   "type": "root",
   "nodes": [
     {
       "id": 2147483647,
       "name": "__i3",
       "rect": {
         "x": 1920,
         "y": 0,
         "width": 1920,
         "height": 1080
       },
       "focused": false,
       "focus": [
         2147483646
       ],
       "border": "none",
       "current_border_width": 0,
       "layout": "output",
       "orientation": "horizontal",
       "percent": null,
       "window_rect": {
         "x": 0,
         "y": 0,
         "width": 0,
         "height": 0
       },
       "deco_rect": {
         "x": 0,
         "y": 0,
         "width": 0,
         "height": 0
       },
       "geometry": {
         "x": 0,
         "y": 0,
         "width": 0,
         "height": 0
       },
       "window": null,
       "urgent": false,
       "floating_nodes": [
       ],
       "sticky": false,
       "type": "output",
       "nodes": [
         {
           "id": 2147483646,
           "name": "__i3_scratch",
           "rect": {
             "x": 1920,
             "y": 0,
             "width": 1920,
             "height": 1080
           },
[...]

PR #4069, opened 2019-04-19
cmd_move: allow for all i3 syntax options

RedSoxFan:
Related https://github.com/swaywm/sway/issues/3793#issuecomment-484433598

This modifies cmd_move to allow for the [syntax options allowed by i3](https://github.com/i3/i3/blob/a574346964ce17d35df2ef9072ba16e26209ac84/parser-specs/commands.spec#L316-L324).

The following syntaxes are supported:
- `move left|right|up|down [<amount> [px]]`
- `move [--no-auto-back-and-forth] [window|container] [to] workspace
   <name>|next|prev|next_on_output|prev_on_output|current|number <num>`
- `move [window|container] [to] output <name/id>|left|right|up|down`
- `move [window|container] [to] mark <mark>`
- `move workspace to [output] <name/id>|left|right|up|down`
- `move [window|container] [to] [absolute] position <x> [px] <y> [px]`
- `move [window|container] [to] [absolute] position center`
- `move [window|container] [to] position mouse|cursor|pointer`

This also allows retains the following syntax option that is not
supported by i3, but is supported in sway 1.0:
- `move workspace [to] output <name/id>|left|right|up|down`

The changes are:
- `window` and `container` are now optional
- `output` is now optional for `move workspace` when `to` is given

There is also stricter command checking now. If `absolute` or
`--no-auto-back-and-forth` are given for commands that do not support
them, it will be treated as invalid syntax instead of being silently
ignored.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -377,10 +377,11 @@ static bool container_move_in_direction(struct sway_container *container,
 	return false;
 }
 
-static struct cmd_results *cmd_move_container(int argc, char **argv) {
+static struct cmd_results *cmd_move_container(bool no_auto_back_and_forth,
+		int argc, char **argv) {
 	struct cmd_results *error = NULL;
 	if ((error = checkarg(argc, "move container/window",
-				EXPECTED_AT_LEAST, 3))) {
+				EXPECTED_AT_LEAST, 2))) {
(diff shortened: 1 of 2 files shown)


ddevault:
Thanks!

PR #4071, opened 2019-04-19
Correctly reset the cursor image after a seatop

ForTheReallys:
This fixes a minor annoyance when moving or resizing using the floating_modifier.  Currently, after finishing moving or resizing a window, the cursor is reset to "left_ptr", not the image requested by the client.  This PR saves said image and restores it correctly.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -193,11 +193,9 @@ static void cursor_do_rebase(struct sway_cursor *cursor, uint32_t time_msec,
 		struct sway_node *node, struct wlr_surface *surface,
 		double sx, double sy) {
 	struct wlr_seat *wlr_seat = cursor->seat->wlr_seat;
-	// Handle cursor image
 	if (surface) {
 		if (seat_is_input_allowed(cursor->seat, surface)) {
 			wlr_seat_pointer_notify_enter(wlr_seat, surface, sx, sy);
-			wlr_seat_pointer_notify_motion(wlr_seat, time_msec, sx, sy);
 		}
 	} else if (node && node->type == N_CONTAINER) {
 		// Try a node's resize edge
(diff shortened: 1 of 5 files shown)


emersion:
We should probably just send a `wl_surface.enter` event instead. (This would be done with `cursor_rebase`)

ForTheReallys:
You're right.  I thought we already sent `enter` in `cursor_rebase`. I didn't realize wlroots doesn't send this if the surface already has pointer focus.

emersion:
Oh. I wonder if we should clear pointer focus during a seatop. Probably?

ForTheReallys:
I would think so.  At least for the resize and move seatops right?  I just finished testing that out and it looks like it actually fixes another minor bug where pointer motion sent while moving a container.

emersion:
Yeah. Nice!

sway/input/seatop_default.c
```diff
@@ -192,12 +192,12 @@ static void state_add_button(struct seatop_default_event *e, uint32_t button) {
...
+	struct wlr_seat *wlr_seat = cursor->seat->wlr_seat;
 	// Handle cursor image
 	if (surface) {
-		// Reset cursor if switching between clients
-		struct wl_client *client = wl_resource_get_client(surface->resource);
-		if (client != cursor->image_client) {
-			cursor_set_image(cursor, "left_ptr", client);
+		if (seat_is_input_allowed(cursor->seat, surface)) {
+			wlr_seat_pointer_notify_enter(wlr_seat, surface, sx, sy);
+			wlr_seat_pointer_notify_motion(wlr_seat, time_msec, sx, sy);
```
emersion: Is the `wlr_seat_pointer_notify_motion` call really required?

ForTheReallys: It's not required here.  I can put it in `handle_motion` instead if you'd like.

emersion:
1. Build is failing, can you look into it?
2. I'm not able to test, because there were wlroots breaking changes. Can you rebase?

ForTheReallys:
Sure. Give me a minute

emersion:
Thanks!

PR #4073, opened 2019-04-20
ipc: fix criteria for emitting bar_state_update

### LINKED ISSUE #4066: Swaybar remains visible after setting hidden_state to hide.
* Sway Version:
1.0 (arch repository 1.0-8)

* Configuration File:
```
mode "test mode" {
    bindsym Escape mode default; bar hidden_state hide
}
bindsym $mod+period mode "test mode"; bar hidden_state show

...

bar {
    mode hide
    ....
}
```

* Debug Log:
https://gist.github.com/JangYeeun/e154be114fe99b74a1f71675134ca14e

After pressing `$mod+period` (and releasing `$mod`) the bar is visible. (as expected)
After pressing `Escape` the bar is still visible. (expected to disappear)
Only after pressing `Alt` or `Ctrl` or `Shift` or `$mod` the bar will disappear.

RedSoxFan:
Fixes #4066 

This fixes the criteria for emitting a `bar_state_update` event to
notify swaybar (and any other bars utilizing the event) on whether the
bar is visible by modifier. It is not enough to only emit the event
when both the bar mode and bar hidden state are `hide` since it is
possible to release the modifier while hidden state is `show` and then
change hidden state to `hide` without pressing the modifier. This also
emits the event whenever visible by modifier is set and should no
longer be regardless of the mode and state to ensure that it gets
properly cleared. If visible by modifier is not set and the bar is not
in `hide`/`hide`, then no events will be sent and visible by modifier
will not be set

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -424,11 +424,18 @@ static int handle_keyboard_repeat(void *data) {
 static void determine_bar_visibility(uint32_t modifiers) {
 	for (int i = 0; i < config->bars->length; ++i) {
 		struct bar_config *bar = config->bars->items[i];
-		if (strcmp(bar->mode, bar->hidden_state) == 0) { // both are "hide"
-			bool should_be_visible = 
-				bar->modifier != 0 && (~modifiers & bar->modifier) == 0;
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4074, opened 2019-04-20
Validate icon_struct in read_theme_file

Ppjet6:
The read_theme_file function used to return an invalid icon_struct in
some cases, for example when an empty index.theme file was read.

This makes sure the struct we're returning is always valid as per the
Icon Theme specification.

Signed-off-by: Maxime “pep” Buquet <pep@bouah.net>


RedSoxFan:
Thanks!

PR #4081, opened 2019-04-24
swaybar: add multiseat support

### LINKED ISSUE #2780: Multiseat support for swaybar
We should also check that `swaylock` handles multiseat properly.

RedSoxFan:
Closes #2780

This just adds multiseat support to swaybar


ddevault:
Thanks!

PR #4082, opened 2019-04-24
swaybar: hide mode visibility improvements

### LINKED ISSUE #3036: hidden statusbar remains hidden when using a "mode" keybinding
sway version 1.0-beta.1-108-gb90af335 (Oct 30 2018, branch 'master')

Relevant config:

```
mode "resize" {
    bindsym Left resize shrink width 10 px or 10 ppt
    bindsym Down resize grow height 10 px or 10 ppt
    bindsym Up resize shrink height 10 px or 10 ppt
    bindsym Right resize grow width 10 px or 10 ppt
    bindsym Return mode "default"
    bindsym Escape mode "default"
}
bindsym $mod+r mode "resize"

bar {
    position bottom
    mode hide
    status_command i3status
}
```

My bar is set to "hidden". 
In i3, it becomes visible when I enter into a "mode" (e.g. "resize") and remains visible until I leave the "mode". This is important because the info about the mode and about being in the mode is printed in the bar.
In sway, it remains hidden.

RedSoxFan:
Closes #3036

This allows swaybar to become visible when the mode changes (to any
mode other than the default). swaybar will be hidden again when the
modifier is pressed and released or when switching back to the default
mode.

This also applies the same logic to visible by urgency to hide swaybar
when the modifier is pressed and released.

These changes are to match i3's behavior.

### DIFF
--- a/include/swaybar/bar.h
+++ b/include/swaybar/bar.h
@@ -23,6 +23,7 @@ struct swaybar {
 	// only relevant when bar is in "hide" mode
 	bool visible_by_modifier;
 	bool visible_by_urgency;
+	bool visible_by_mode;
 	bool visible;
 
 	struct wl_display *display;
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks!

PR #4084, opened 2019-04-25
cmd_move: respect workspace layout when inserting

### LINKED ISSUE #4038: Window `move` does't preserve its layout (unlike i3)
Sway version: 1.0

Steps to reproduce:
- `layout tabbed`
- open 3 windows
- on the right-most window: `move right`
- open 1 more window, while still on the 3rd window (the one moved to the right)

Expected result (ie. `i3` behaviour):
- first two windows are on the left half of the screen, tabbed
- 4th window is on the right half of the screen, tabbed over the 3rd window

Actual result:
- first two windows are on the left third of the screen, tabbed _(good behaviour)_
- 3rd window in the middle third of the screen
- 4th window in the right third of the screen
- layout has changed to `splith`

RedSoxFan:
Fixes #4038 

When moving a container to become a direct child of the workspace and
the workspace's layout is tabbed or stacked, wrap it in a container
with the same layout. This allows for the following:
- Run `layout tabbed|stacked` on an empty workspace (or use
  `workspace_layout tabbed|stacked` in the config)
- Open some views
- Move one of the views in any direction
- Open another view
- The new container should also be `tabbed`/`stacked`

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1385,8 +1385,10 @@ void container_replace(struct sway_container *container,
 		root_scratchpad_show(container);
 		root_scratchpad_remove_container(container);
 	}
-	container_add_sibling(container, replacement, 1);
-	container_detach(container);
+	if (container->parent || container->workspace) {
+		container_add_sibling(container, replacement, 1);
+		container_detach(container);
+	}
 	if (scratchpad) {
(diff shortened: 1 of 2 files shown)


ddevault:
Thanks!

PR #4088, opened 2019-04-25
swaybg: split into standalone project

ddevault:
The new upstream is https://github.com/swaywm/swaybg


emersion:
It seems `swaybar` still needs gdk-pixbuf unfortunately, because of tray icons.

ddevault:
Hm, I'm not sure it actually uses it. It compiles fine without it and that `#ifdef` is the only reference to it I can find.

ddevault:
Nevermind, you're right.

common/background-image.c
```diff
@@ -21,6 +24,103 @@ enum background_mode parse_background_mode(const char *mode) {
 	return BACKGROUND_MODE_INVALID;
 }
 
+#if HAVE_GDK_PIXBUF
+static cairo_surface_t* gdk_cairo_image_surface_create_from_pixbuf(
```
emersion: Is it intentional to move this from `cairo.c` to this file?

ddevault: Yeah.

emersion:
Thanks!

PR #4089, opened 2019-04-25
ci: enable all features

emersion:
If unspecified, feature options are set to "auto", which means enabled only if
the required dependencies are found. In CI we don't want to miss compilation
errors because a dependency hasn't been found and code isn't built.

Leave FreeBSD out for now because it uses a subproject (haven't found a way to
make auto_features=enabled only apply to the toplevel project).


ddevault:
Thanks!

PR #4092, opened 2019-04-26
Unbreak -Werror on 32-bit platforms

jbeich:
Regressed by #4062. See error logs: [i386](https://github.com/swaywm/sway/files/3121838/sway-1.0.r1.257.log), [armv7](https://github.com/swaywm/sway/files/3121842/sway-1.0.r1.257.log).

### DIFF
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -75,7 +75,7 @@ bool ipc_set_recv_timeout(int socketfd, struct timeval tv) {
 		return false;
 	}
 	sway_log(SWAY_DEBUG, "ipc recv timeout set to %ld.%06ld",
-			tv.tv_sec, tv.tv_usec);
+			(long)tv.tv_sec, (long)tv.tv_usec);
 	return true;
 }
 


common/ipc-client.c
```diff
@@ -75,7 +75,7 @@ bool ipc_set_recv_timeout(int socketfd, struct timeval tv) {
 		return false;
 	}
 	sway_log(SWAY_DEBUG, "ipc recv timeout set to %ld.%06ld",
-			tv.tv_sec, tv.tv_usec);
+			(long)tv.tv_sec, (long)tv.tv_usec);
```
jbeich: Oops, `.tv_usec` was fine as is.

jbeich: `suseconds_t` is `long` on BSDs/Linux/Solaris, `long long` on Linux [x32](https://en.wikipedia.org/wiki/X32_ABI) and `int` on Linux sparc.
https://git.musl-libc.org/cgit/musl/tree/arch/x32/bits/alltypes.h.in?id=v1.1.22#n23
https://github.com/torvalds/linux/blob/v5.0/arch/sparc/include/uapi/asm/posix_types.h#L19
Cast both `time_t` and `suseconds_t` appears more common.
https://codesearch.debian.net/search?q=printf.*tv_usec

ddevault:
Why don't we just remove this logging statement? It's not very important.

RedSoxFan:
> Why don't we just remove this logging statement? It's not very important.

+1. I meant to remove this before submitting #4062.  Currently the function is only called from swaymsg. Since swaymsg is hard coded to the `SWAY_INFO` logging level, the statement is never be printed anyway.

Issue #4108, opened 2019-04-30
"workspace <name> output <output>" does not work as documented


mythmon:
* Sway Version: `sway version  (2019-02-07, branch "")` (Installed from Fedora 30 repositories)

* Debug Log: https://gist.github.com/mythmon/e64a7c1344fd561b30702f1e21ba0bf2

* Configuration File: Reproduced with the default configuration file.

The command `workspace <name> output <output>` is listed in `man 5 sway` as specifying "that the workspace named name should appear on the specified output". However, when running the command, the named workspace does not move to the named output, and no error is reported.

Ultimately my goal is to create a keybinding that brings a given workspace to the currently focused output, even if that workspace is currently displayed on another output. I'm not asking for the workspace to be shown on both outputs, only the current way. If there is another way to do this, I'd like to hear about it.

RedSoxFan:
`workspace <name> output <output>` does not move existing workspaces. It defines behavior for newly created workspaces of that name. The behavior stems from i3.

Unfortunately, I don't think there is currently a _good_ way to achieve your goal without an IPC script however.

It appears that sway is missing the `current` output for the `move` commands (as implemented for i3 in i3/i3#2483) so I'll submit a PR for that in a little bit.

As stated in i3/i3#2357, the `current` output would allow for `[workspace=<name>] move workspace to output current`. However, that only works for non-empty workspaces. The i3 issue appears to suggest that i3 would be open to adding the functionality, but it doesn't look like it was followed up on.

mythmon:
> workspace <name> output <output> does not move existing workspaces. It defines behavior for newly created workspaces of that name. The behavior stems from i3.

For users that aren't familiar with i3 (like me), it would be good to note that in the man page then. As it stands it probably a little over-terse. Sway seems like one of the only viable tiling window managers that supports Wayland, so I imagine a lot of users that have never used i3 will try to use it. I think that is a useful set of users to take into consideration (though being part of this group, I'm biased :smile:).

I was exploring `workspace <name> output <output>` because `move workspace to output <output>` only seemed to be able to operate on the currently focused workspace. Are you saying that you can specify the workspace to move by prefixing the command with `[workspace=<name>]`? If so, the external script to do this would be reasonable to write (and I was already planning to write a similar script anyways).

RedSoxFan:
> For users that aren't familiar with i3 (like me), it would be good to note that in the man page then.

Submitted as #4111 

[...]

PR #4109, opened 2019-04-30
cmd_move: add support for output current

RedSoxFan:
Related to https://github.com/swaywm/sway/issues/4108#issuecomment-488104938

This adds support for the following commands for i3 compatibility:
- `move [window|container] [to] output current`
- `move workspace to [output] current`
- `move workspace [to] output current`

The above commands are only useful when used with criteria.


ddevault:
Thanks!

PR #4113, opened 2019-05-01
The -w flag does not exist in swayidle

matrixise:
Remove the -w flag from the example

### DIFF
--- a/config.in
+++ b/config.in
@@ -34,7 +34,7 @@ output * bg @datadir@/backgrounds/sway/Sway_Wallpaper_Blue_1920x1080.png fill
 #
 # Example configuration:
 #
-# exec swayidle -w \
+# exec swayidle \
 #          timeout 300 'swaylock -f -c 000000' \
 #          timeout 600 'swaymsg "output * dpms off"' \
 #               resume 'swaymsg "output * dpms on"' \


emersion:
Except it does exist?

https://github.com/swaywm/swayidle/blob/master/main.c#L504

RedSoxFan:
The `-w` flag does exist: https://github.com/swaywm/swayidle/blob/3e392e31c0684854a9a145cda1bd9a44c99ef24d/main.c#L505

It was added in https://github.com/swaywm/swayidle/pull/7

matrixise:
@emersion @RedSoxFan not in the sway-1.0-0.beta.2 on fedora 29 :/

RedSoxFan:
> @emersion @RedSoxFan not in the sway-1.0-0.beta.2 on fedora 29 :/

Sway 1.0 Beta 2 was released back on November 29th of last year, which predates the `-w` flag (January 14th). Also, sway no longer includes swayidle as it was extracted into a separate project on January 12th.

Additionally, it appears that Fedora 29 does not include any 1.0 release (or pre-release) in their [repositories](https://apps.fedoraproject.org/packages/sway) so I'm guessing you are using an outdated COPR and would recommend switching to a different one (or building from source).

PR #4119, opened 2019-05-03
swaymsg: fix quiet error reporting

### LINKED ISSUE #4116: swaymsg --quiet option and return status code
Using sway 1.0 on Arch Linux

I'm using the "--quiet" option of swaymsg which is described in the manual as
"Sends the IPC message but does not print the response from sway."

I was expecting this option to only affect the output but not the return status code.
Consider the differences between the following commands:

`swaymsg xxx`
outputs "Error: Unknown/invalid command 'xxx'" 
return status code is 1

`swaymsg --quiet xxx`
outputs nothing
return status code is 0 

I'm not sure if this is to be considered a bug or not.
Reporting it just in case.

RedSoxFan:
Fixes #4116 

This makes it so swaymsg still returns the correct successful or failed
error code when in quiet mode

### DIFF
--- a/swaymsg/main.c
+++ b/swaymsg/main.c
@@ -397,6 +397,9 @@ int main(int argc, char **argv) {
 	if (!socket_path) {
 		socket_path = get_socketpath();
 		if (!socket_path) {
+			if (quiet) {
+				exit(EXIT_FAILURE);
+			}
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4122, opened 2019-05-03
Fix a race condition when respawning swaybg

Hummer12007:
The destroy handler is sometimes called after a new instance was spawned and config->swaybg_client is already assigned to a new value, so in some cases wl_client_destroy is not called for newly spawned swaybg instances after changing the wallpaper and the background image gets stuck (and a lot of useless swaybg instances are spawned and not terminated)

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -482,9 +482,11 @@ void free_output_config(struct output_config *oc) {
 
 static void handle_swaybg_client_destroy(struct wl_listener *listener,
 		void *data) {
-	wl_list_remove(&config->swaybg_client_destroy.link);
-	wl_list_init(&config->swaybg_client_destroy.link);
-	config->swaybg_client = NULL;
+	struct wl_client *client = data;
+	wl_list_remove(&listener->link);
+	if (client == config->swaybg_client) {
+		config->swaybg_client = NULL;
+	}
 }
 
 static bool _spawn_swaybg(char **command) {


sway/config/output.c
```diff
@@ -482,9 +482,10 @@ void free_output_config(struct output_config *oc) {
 
 static void handle_swaybg_client_destroy(struct wl_listener *listener,
 		void *data) {
-	wl_list_remove(&config->swaybg_client_destroy.link);
-	wl_list_init(&config->swaybg_client_destroy.link);
-	config->swaybg_client = NULL;
+	struct wl_client *client = data;
+	wl_list_remove(&listener->link);
+	if (client == config->swaybg_client)
+		config->swaybg_client = NULL;
```
RedSoxFan: Style: braces are required

emersion:
>config->swaybg_client is already assigned to a new value

That means the listener is setup twice, without being removed in-between. Something else is going wrong.

sway/config/output.c
Hummer12007: Fixed

RedSoxFan:
@Hummer12007 I think you had half of it. Can you please try https://github.com/RedSoxFan/sway/commit/f4dcb70ee0981d841d6e4a16cebb30f5ad7f0a50 and see if that fixes the issue?

Hummer12007:
Works for me too.

RedSoxFan:
Thanks @Hummer12007 for the initial look into this issue

emersion:
Seconded, that definitely helped!

Hummer12007:
Thanks for fixing the issue!

PR #4123, opened 2019-05-03
Fix xdg-decoration unconfigured if set before first commit

emersion:
In case a set_mode/unset_mode request is sent before the first commit, we need
to handle the value and send our preference accordingly.

This fixes xdg-decoration support for Qt apps.


ddevault:
Thanks!

PR #4124, opened 2019-05-04
Destroy swaybg client on reload

RedSoxFan:
Supersedes/Closes #4122  _(thanks @Hummer12007 for the initial PR)_

When reloading, this destroys the old config's swaybg client before
spawning the new config's swaybg. This fixes a race condition where the
old config's swaybg client's destroy was being called after the new
config's swaybg client was being spawned. This was causing the
reference to the new swaybg client to be removed and never destroyed.

This also modifies handle_swaybg_client_destroy to grab the config
reference using wl_container_of on the listener since the swaybg client
may be the old config swaybg client and should be used instead of the
global config instance

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -441,6 +441,10 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 		config->reloading = true;
 		config->active = true;
 
+		if (old_config->swaybg_client != NULL) {
+			wl_client_destroy(old_config->swaybg_client);
+		}
+
 		if (old_config->swaynag_config_errors.client != NULL) {
 			wl_client_destroy(old_config->swaynag_config_errors.client);
 		}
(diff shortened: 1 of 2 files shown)


emersion:
Ah, makes sense. Do we need to do something similar for `swaybar`?

RedSoxFan:
> Ah, makes sense. Do we need to do something similar for `swaybar`?

It looks like swaybar isn't even being launched as a wayland client so it actually needs the full conversion. For some reason, I thought that was already done.

I also just checked swaynag and it looks like swaynag already handles the stuff in this PR so that's good.

emersion:
>For some reason, I thought that was already done.

Yeah, me too

>I also just checked swaynag and it looks like swaynag already handles the stuff in this PR so that's good.

Yup, swaynag sounds fine

emersion:
Thanks!

PR #4128, opened 2019-05-05
[WIP] Tweaks to preserve tabbed and stack layout when moving out a view out of a container

fdlamotte:
Tries to ddress #4084 since current fix only works when workspace has a tabbed or stacked layout.

If a view is moved out of a tabbed (or stacked) container, it becomes tabbed (or stacked)

There was also other issues 
 - when the container had only two views, and where the remaining view lost the tabbed or stack layout
 - when moving perpendicularly out of a container

Some issue are still unsolved 
 - when there is only one view in a stacked or tabbed container and we try to move it out (horizontaly or verticaly)
 - still fails with some layouts

I'm not proud of this code, it feels really like a patchwork of experiments ... but it fixes one major thing I lacked from i3 in my workflow (splitting a tabbed layout by moving out one view) and while it's still not perfect I find the behaviour better and I'm currently using it ...

So I share it hoping it can help find a good solution to solve the initial issue (I'd be in favor of a refactoring of the move command).

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -320,6 +320,8 @@ static bool container_move_in_direction(struct sway_container *container,
 	int offs =
 		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_UP ? -1 : 1;
 
+	enum sway_container_layout old_layout = container_parent_layout(container);
+
 	while (current) {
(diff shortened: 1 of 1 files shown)


fdlamotte:
well ... it does not really work (only in simple cases), there are lots of edge cases :(

PR #4141, opened 2019-05-12
Spawn swaybar as a wayland client

RedSoxFan:
Follow-up to https://github.com/swaywm/sway/pull/4124#issuecomment-489297910

This just makes it so swaybar is handled as a wayland client


emersion:
Thanks!

Issue #4148, opened 2019-05-13
Invalid xkb options result in unusable keyboard and therefore no way to exit sway


jktjkt:
* Sway Version: `1.1-rc1-4-g8c8e6a6c (May 11 2019, branch 'HEAD')`

I was trying to set up a local keyboard layout. I made a mistake and confused `en` with `us`, etc. This resulted in a sway instance that I was not available to terminate (no `Ctrl-Alt-Fn` switching, no `win-shift-e`, etc). I understand that it is a config error, but it's a pity that I had to power-cycle the whole box.

- A minimal config file:
```
set $mod Mod4

input type:keyboard {
        xkb_layout "en,cs"
        xkb_variant ",qwerty"
        xkb_options "grp:alt_shift_toggle"
}

bindsym $mod+Return exec konsole

bindsym $mod+Shift+q kill

bindsym $mod+Shift+e exec swaynag -t warning -m 'You pressed the exit shortcut. Do you really want to exit sway? This will end your Wayland session.' -b 'Yes, exit sway' 'swaymsg exit'
```
(Some) invalid configurations result in an error message like this in the debug log:
```
xkbcommon: ERROR: Couldn't find file "symbols/en" in include paths
xkbcommon: ERROR: 1 include paths searched:
xkbcommon: ERROR:       /usr/share/X11/xkb
xkbcommon: ERROR: 1 include paths could not be added:
xkbcommon: ERROR:       /home/jkt/.xkb
xkbcommon: ERROR: Abandoning symbols file "(unnamed)"
xkbcommon: ERROR: Failed to compile xkb_symbols
xkbcommon: ERROR: Failed to compile keymap
2019-05-13 13:15:28 - [sway-9999/sway/input/keyboard.c:515] cannot configure keyboard: keymap does not exist
```

- Expected behavior:
  - In a perfect world, I would love to see a user-visible error message (similar to what I get when I reload config on the fly and there's a config schema error).
  - Alternatively, as a minimum, make sure that keyboard is still available. Default to an english layout "as if no config was present".

In case someone ends up here by a random web search, this is the correct config stanza:
```
input type:keyboard {
        xkb_layout "us,cz"
        xkb_variant ",qwerty"
        xkb_options "grp:alt_shift_toggle"
}
```

PR #4153, opened 2019-05-14
input/keyboard: attempt default keymap on failure

RedSoxFan:
Partial fix for #4148. The user facing error portion is still a WIP

This attempts to use the default keymap when the one defined in the
input config fails to compile. The goal is to make it so the keyboard
is always in a usable state, even if it is not the user's requested
settings as usability is more important.

This also removes the calls to `getenv` for the `XKB_DEFAULT_*` family
of environment variables. The reasoning is libxkbcommon will fallback
to using those (and then the system defaults) when any of the rule
names are `NULL` or an empty string anyway so there is no need for
sway to duplicate the efforts.

### DIFF
--- a/include/sway/input/keyboard.h
+++ b/include/sway/input/keyboard.h
@@ -65,6 +65,8 @@ struct sway_keyboard {
 	struct sway_binding *repeat_binding;
 };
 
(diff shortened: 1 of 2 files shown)


ddevault:
Thanks!

PR #4154, opened 2019-05-14
config: Fix socket/wl_client leaks when executing child processes fails

shramov:
Added spawn_wl_client helper function that is used for spawning both bar and
background processes

There are still unresolved issues so please don't merge until they are all solved.

Where to place this helper function? Config is not good place for it.

Failing on close(sockets[1]) is not wise in my opinion. It's cleanup stage and we can not do anything with it.

Also there is same code with same problems in sway/swaynag.c but it has pipe setup stage.
It can be generalized if posix_spawn is used instead of inner fork.

### DIFF
--- a/include/sway/util.h
+++ b/include/sway/util.h
@@ -1,0 +1,14 @@
+#ifndef _SWAY_SWAY_UTIL_H
+#define _SWAY_SWAY_UTIL_H
+
+#include <spawn.h>
+
+/**
+ * Close fd and log theoretical case when close(2) failed
+ */
+void close_warn(int fd);
+
+struct wl_client *spawn_wl_client(char * const cmd[], struct wl_display *display);
+struct wl_client *spawn_wl_client_fa(char * const cmd[], struct wl_display *display, posix_spawn_file_actions_t *fa);
+
+#endif//_SWAY_SWAY_UTIL_H
--- a/sway/config/bar.c
+++ b/sway/config/bar.c
@@ -13,6 +13,7 @@
 #include "sway/config.h"
 #include "sway/input/keyboard.h"
 #include "sway/output.h"
+#include "sway/util.h"
 #include "config.h"
 #include "list.h"
 #include "log.h"
@@ -187,79 +188,25 @@ static void handle_swaybar_client_destroy(struct wl_listener *listener,
 	bar->client = NULL;
 }
 
-static void invoke_swaybar(struct bar_config *bar) {
-	int sockets[2];
-	if (socketpair(AF_UNIX, SOCK_STREAM, 0, sockets) != 0) {
-		sway_log_errno(SWAY_ERROR, "socketpair failed");
-		return;
-	}
-	if (!set_cloexec(sockets[0], true) || !set_cloexec(sockets[1], true)) {
-		return;
-	}
-
-	bar->client = wl_client_create(server.wl_display, sockets[0]);
-	if (bar->client == NULL) {
-		sway_log_errno(SWAY_ERROR, "wl_client_create failed");
-		return;
-	}
-
-	bar->client_destroy.notify = handle_swaybar_client_destroy;
-	wl_client_add_destroy_listener(bar->client, &bar->client_destroy);
-
-	pid_t pid = fork();
-	if (pid < 0) {
-		sway_log(SWAY_ERROR, "Failed to create fork for swaybar");
-		return;
-	} else if (pid == 0) {
-		// Remove the SIGUSR1 handler that wlroots adds for xwayland
-		sigset_t set;
-		sigemptyset(&set);
-		sigprocmask(SIG_SETMASK, &set, NULL);
-
-		pid = fork();
-		if (pid < 0) {
-			sway_log_errno(SWAY_ERROR, "fork failed");
-			_exit(EXIT_FAILURE);
-		} else if (pid == 0) {
-			if (!set_cloexec(sockets[1], false)) {
-				_exit(EXIT_FAILURE);
-			}
-
-			char wayland_socket_str[16];
-			snprintf(wayland_socket_str, sizeof(wayland_socket_str),
-					"%d", sockets[1]);
-			setenv("WAYLAND_SOCKET", wayland_socket_str, true);
-
-			// run custom swaybar
-			char *const cmd[] = {
-					bar->swaybar_command ? bar->swaybar_command : "swaybar",
-					"-b", bar->id, NULL};
-			execvp(cmd[0], cmd);
-			_exit(EXIT_FAILURE);
-		}
-		_exit(EXIT_SUCCESS);
-	}
-
-	if (close(sockets[1]) != 0) {
-		sway_log_errno(SWAY_ERROR, "close failed");
-		return;
-	}
-
-	if (waitpid(pid, NULL, 0) < 0) {
-		sway_log_errno(SWAY_ERROR, "waitpid failed");
-		return;
-	}
-
-	sway_log(SWAY_DEBUG, "Spawned swaybar %s", bar->id);
-	return;
-}
-
 void load_swaybar(struct bar_config *bar) {
 	if (bar->client != NULL) {
 		wl_client_destroy(bar->client);
 	}
 	sway_log(SWAY_DEBUG, "Invoking swaybar for bar id '%s'", bar->id);
-	invoke_swaybar(bar);
+
+	char *const cmd[] = {
+			bar->swaybar_command ? bar->swaybar_command : "swaybar",
+			"-b", bar->id, NULL};
+	bar->client = spawn_wl_client(cmd, server.wl_display);
+
+	if (bar->client == NULL) {
+		sway_log(SWAY_ERROR, "Failed to spawn swaybar %s", bar->id);
+		return;
+	}
+
+	sway_log(SWAY_DEBUG, "Spawned swaybar %s", bar->id);
+	bar->client_destroy.notify = handle_swaybar_client_destroy;
+	wl_client_add_destroy_listener(bar->client, &bar->client_destroy);
 }
 
 void load_swaybars(void) {
--- a/sway/swaynag.c
+++ b/sway/swaynag.c
@@ -10,6 +10,7 @@
 #include "log.h"
 #include "sway/server.h"
 #include "sway/swaynag.h"
+#include "sway/util.h"
 #include "util.h"
 
 static void handle_swaynag_client_destroy(struct wl_listener *listener,
@@ -19,6 +20,40 @@ static void handle_swaynag_client_destroy(struct wl_listener *listener,
 	wl_list_remove(&swaynag->client_destroy.link);
 	wl_list_init(&swaynag->client_destroy.link);
 	swaynag->client = NULL;
+}
+
+static struct wl_client *swaynag_spawn_detailed(char *cmd[], struct swaynag_instance *swaynag) {
+	if (pipe(swaynag->fd) != 0) {
+		sway_log(SWAY_ERROR, "Failed to create pipe for swaynag");
+		return NULL;
+	}
+
+	if (!set_cloexec(swaynag->fd[1], true)) {
+		goto failed;
+	}
+
+	posix_spawn_file_actions_t fa;
+	posix_spawn_file_actions_init(&fa);
+	posix_spawn_file_actions_adddup2(&fa, swaynag->fd[0], STDIN_FILENO);
+	posix_spawn_file_actions_addclose(&fa, swaynag->fd[0]);
+
+	struct wl_client *client = spawn_wl_client_fa(cmd, server.wl_display, &fa);
+
+	posix_spawn_file_actions_destroy(&fa);
+
+	if (client == NULL) {
+		goto failed;
+	}
+
+	close_warn(swaynag->fd[0]);
+
+	return client;
+
+failed:
+	close_warn(swaynag->fd[0]);
+	close_warn(swaynag->fd[1]);
+
+	return NULL;
 }
 
 bool swaynag_spawn(const char *swaynag_command,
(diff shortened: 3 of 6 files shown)


shramov:
Also is there any reason to set CLOEXEC on sockets[1] and the removing it after fork?
If sway is single process and (I hope) is not trying to run fork/exec in sighandlers it's not required.

emersion:
>Added spawn_wl_client helper function that is used for spawning both bar and
background processes

Good idea

>Where to place this helper function? Config is not good place for it.

Maybe `util.c`?

>Failing on close(sockets[1]) is not wise in my opinion. It's cleanup stage and we can not do anything with it.

I prefer to fail rather than to leak an opened file descriptor.

>Also there is same code with same problems in sway/swaynag.c but it has pipe setup stage.
It can be generalized if posix_spawn is used instead of inner fork.

Not a fan of posix_spawn

>Also is there any reason to set CLOEXEC on sockets[1] and the removing it after fork?
If sway is single process and (I hope) is not trying to run fork/exec in sighandlers it's not required.

We fork-and-exec pretty frequently. We need to ensure all file descriptors have CLOEXEC set, otherwise we leak compositor file descriptors to child processes (which can be a security issue in addition to being a cleanup issue).

shramov:
> Maybe `util.c`?
It needs global 'server' variable, so util maybe is not best for it. However it's definetly something that belongs to util...

> Not a fan of posix_spawn
API is really a bit braindamaging but adding pipe setup or similar things will lead to recreating it localy.


> We fork-and-exec pretty frequently.
Can this function be interrupted with fork? If not - you don't need to set closeexec and then remove it

shramov:
Also if "CLOEXEC on all sockets" is sway policy it's better to create socketpair with SOCK_STREAM | SOCK_CLOEXEC argument to eliminate theoretical socket leaking

emersion:
>It needs global 'server' variable, so util maybe is not best for it. However it's definetly something that belongs to util...

Maybe add a `struct wl_display *` as first argument instead?

>API is really a bit braindamaging but adding pipe setup or similar things will lead to recreating it localy.

I'm fine with keeping the swaynag version duplicated.

>Can this function be interrupted with fork? If not - you don't need to set closeexec and then remove it

Well it's not a matter of being interrupted. It's just a matter of having a socket opened without CLOEXEC. Next time we'll spawn a process it will leak.

>Also if "CLOEXEC on all sockets" is sway policy it's better to create socketpair with SOCK_STREAM | SOCK_CLOEXEC argument to eliminate theoretical socket leaking into child processes

This isn't POSIX.

shramov:
> This isn't theoretical. Remove the CLOEXEC code, make sway spawn a new process (trigger any `exec` command in the sway config via a binding), inspect the new process FDs and notice sway's FDs are listed.

socket[1] is closed on exit from this function so cloexec or not does not matter.

> > Also if "CLOEXEC on all sockets" is sway policy it's better to create socketpair with SOCK_STREAM | SOCK_CLOEXEC argument to eliminate theoretical socket leaking into child processes
> 
> This isn't POSIX.

Both BSD and Linux supports it. If your policy is "all sockets are CLOEXEC" this is only way to go

> I'm fine with keeping the swaynag version duplicated.

That's a tradeoff between complex helper function and code duplication. Both sides have pros and cons...

emersion:
>socket[1] is closed on exit from this function so cloexec or not does not matter.

Nah, socket[0] is still there. Also socket[0] should be closed in the child (this is currently ensured via CLOEXEC).

>Both BSD and Linux supports it. If your policy is "all sockets are CLOEXEC" this is only way to go

Our policy is POSIX. Not "Linux or BSD".

>That's a tradeoff between complex helper function and code duplication. Both sides have pros and cons...

I prefer duplication in this case.

shramov:
> No, see the current implementation.

Current implementation sets CLOEXEC on sockets[1] for unknown reason.
It may be reasonable only if spawn function can be interrupted with another pair of fork+exec calls.
If it's true then you should use SOCK_STREAM | SOCK_CLOEXEC to ensure that fd never exists without CLOEXEC flag.
If it's not true (and that's our case) there is no reason to set CLOEXEC and then unset it later since sockets[1] fd never leaves that function.

emersion:
>Current implementation sets CLOEXEC on sockets[1] for unknown reason.

The reason is: not needing to make sure every error path makes sure we don't leak a non-CLOEXEC FD.

Feel free to change this if you add error handling.

shramov:
I've pushed small fixup that moved helper function into sway/util.c

Why not into common/util.c? I don't think linking with wayland-server for swaybar is good idea...

Also there is current variant with SOCK_STREAM | SOCK_CLOEXEC, I'll revert it later.

include/sway/config.h
```diff
@@ -622,6 +622,8 @@ void reset_outputs(void);
 
 void free_output_config(struct output_config *oc);
 
+struct wl_client * spawn_wl_client(char * const cmd[], struct wl_display * display, struct wl_listener *client_destroy);
```
emersion: Style: no space at the right of the star (e.g. `struct wl_display *display`)

sway/config/bar.c
```diff
@@ -188,70 +188,14 @@ static void handle_swaybar_client_destroy(struct wl_listener *listener,
...
-			execvp(cmd[0], cmd);
-			_exit(EXIT_FAILURE);
-		}
-		_exit(EXIT_SUCCESS);
-	}
-
-	if (close(sockets[1]) != 0) {
-		sway_log_errno(SWAY_ERROR, "close failed");
-		return;
+	bar->client = spawn_wl_client(cmd, server.wl_display, &bar->client_destroy);
```
emersion: TBH I'd prefer setting up the listener manually in the caller.

sway/util.c
```diff
@@ -0,0 +1,80 @@
...
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+#include <wayland-server.h>
+
+#include "log.h"
+#include "util.h"
+
+struct wl_client * spawn_wl_client(char * const cmd[], struct wl_display * display, struct wl_listener *client_destroy)
+{
```
emersion: Style: `{` on the same line as the function prototype

sway/config/bar.c
emersion: (This would also fix the listener being called in case of error)

emersion (changes requested):
Looks pretty good otherwise

sway/config/bar.c
shramov: I don't know wayland internals but in most cases it's better to add handler as early as possible.
If it's not needed to be called in case of error it may be added before return from function.
Also it's possible to check this argument for NULL and add only if it's set

sway/util.c
shramov: Fixed

sway/config/bar.c
emersion: I don't want our functions to look like this:

```c
struct thing *create_thing(struct wl_listener *event1_listener, struct wl_listener *event2_listener, …, struct wl_listener *eventN_listener);
```

emersion: (And I don't want the listener to be added in case of error inside of `spawn_wl_client` -- right now it is)

shramov: See fixup commit

shramov:
Also invoke_swaybar function looks obsolete now

shramov:
BTW posix_spawn does not look so bad as I've expected

shramov:
Finally I've satisfied with this patch.
Error handling in swaynag is much cleaner now.

shramov:
@emersion looks like github thinks that there are requested changes...
Do you have any issues with last patch?

emersion:
I haven't had time to review the new version yet. Additionally, I'm not convinced by `posix_spawn` and functions like `close_warn`.

shramov:
close_warn is useless in my opinion
However if you want to check close() result it's better to have such function then to place lot of if's

shramov:
Just FYI some links regarding close:
https://lwn.net/Articles/576478/
http://austingroupbugs.net/view.php?id=529

shramov:
Bumping
@emersion This PR is going to decay... Either it has to be closed and discarded or if you have some concerns left I'll try to address them

include/sway/util.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _SWAY_SWAY_UTIL_H
+#define _SWAY_SWAY_UTIL_H
+
+#include <spawn.h>
+
+/**
+ * Close fd and log theoretical case when close(2) failed
+ */
+void close_warn(int fd);
```
emersion: Not a fan of this

include/sway/util.h
```diff
@@ -0,0 +1,14 @@
...
+
+struct wl_client * spawn_wl_client(char * const cmd[], struct wl_display *display);
```
emersion: Style: no space on the right of `*`

sway/util.c
```diff
@@ -0,0 +1,97 @@
...
+		sigset_t set;
+		sigfillset(&set);
+		posix_spawnattr_setflags(&attr, POSIX_SPAWN_SETSIGDEF);
+		posix_spawnattr_setsigdefault(&attr, &set);
+
+		char wayland_socket_str[16];
+		snprintf(wayland_socket_str, sizeof(wayland_socket_str),
+				"%d", sockets[1]);
+		setenv("WAYLAND_SOCKET", wayland_socket_str, true);
+
```
emersion: Extra line

sway/swaynag.c
```diff
@@ -21,6 +22,41 @@ static void handle_swaynag_client_destroy(struct wl_listener *listener,
 	swaynag->client = NULL;
 }
 
+struct wl_client * swaynag_spawn_detailed(char *cmd[], struct swaynag_instance *swaynag) {
+
```
emersion: Extra line

sway/swaynag.c
```diff
@@ -21,6 +22,41 @@ static void handle_swaynag_client_destroy(struct wl_listener *listener,
 	swaynag->client = NULL;
 }
 
+struct wl_client * swaynag_spawn_detailed(char *cmd[], struct swaynag_instance *swaynag) {
```
emersion: This can be static

Style: no space on the right of `*`

emersion (changes requested):
.

include/sway/util.h
shramov: As I've already mentioned there is no point in checking return value of close()
So close_warn() can be just replaced by close()
Is it ok for you?

sway/swaynag.c
shramov: Hm, even in return type? Is not it unreadable?
I've fixed it but probably you can consider leaving space there

include/sway/util.h
emersion: No, we don't want to leak FDs.

sway/swaynag.c
emersion: No. Adding a space makes it unreadable.

More seriously, the style guide is the style guide. Fighting the style guide isn't worth your time.

shramov: No offence. I've fixed that as you've pointed to missing space.
Just tried to clarify

include/sway/util.h
shramov: Leak fd in close?!
How checking fd result can help?

emersion: If `close` fails, something is going very wrong. We should definitely hard fail.

shramov: Hard fail means something like assert
Logging and 'failing' client spawn is not hard fail.
Also if you have not already, please, check links regarding close errors
But here, when we are working with sockets/pipes (not something on NFS) close can not fail...

emersion: >Hard fail means something like assert

No. Assertions should be used if our internal state is inconsistent. That's not the case here.

We have many cases in which very bad things happen (e.g. initializing DRM fails) and we don't assert (for a good reason).

>when we are working with sockets/pipes (not something on NFS) close can not fail...

I don't see this written in POSIX.

shramov: Personally I don't like asserts and never use them in my own code so it was used just as an example.

> I don't see this written in POSIX.

Not everything is written in POSIX, check this one please http://austingroupbugs.net/view.php?id=529
It is about correcting description of posix close and adding restart variant.
Current state of close is that even if it fail you can not do anything with it.
On Linux it strictly can not be interrupted (EINTR) since it can not be restarted.
So what is point here to fail? What can you do? Restart close?

emersion: Look, this is simple: `close` can fail, thus check for errors, and fail if there's one.

Sorry, I'm getting a little tired of this discussion.

shramov: Have you checked link to austingroup and discussion there regarding failing close?

shramov: Also let me clarify my point:

close(2), even failed, closes descritpor. It's not stated in current POSIX but will be added into next revision.

All we want from close(2) here is to close fd and that's done even if (theoretical) error returned.

For us that error means nothing - fd is already closed, no leaking, correct cleanup.

emersion: >For checking close(2) return value I've added small wrapper close_warn that logs error

-1

>For us that error means nothing - fd is already closed, no leaking, correct cleanup.

No - it means FD is not a valid file descriptor, or something else went wrong. Either way, something bad happen. Let's not ignore it.

shramov: There are 2 places with call to close: cleanup in error path and closing client side of socket pair/pipe.

Cleanup path should not be interrupted - it can lead to fd leaking when we fail on first close.
We can log errors there but can not return in the middle.

Closing client side of socketpair/pipe is done after client is spawned. If we want to fail there we have to kill client process. Otherwise we report error to caller but wayland client is already started. But it's hard to kill client process - thanks to double fork we have no control over it. Looks bad to me.

Also brief overview of close errors for BSD/Linux:

EBADF: fd is not valid? When it was closed? There are ~10 lines of code between socketpair and close.
All other errors (EINTR, ECONNRESET(BSD), ENOSPC, EDQUOT, EIO(mystical Linux)) are just informational, fd is closed. So why should we hard fail?

https://www.freebsd.org/cgi/man.cgi?query=close&sektion=2
http://man7.org/linux/man-pages/man2/close.2.html

james-lawrence: *scratches head* slightly confused but the disconnect going on here.

the original code being replaced by this function was just warning anyways. the assertion that if an error occurs then fail doesn't always hold. sometimes just logging and moving on is the correct answer to handle an error.

there is a potential behavior change here because of one spot where the code doesn't return early, but @shramov has covered that in their explanation.

basically looks like everyone is bikeshedding over a non-issue.

@emersion its likely not a real issue as @shramov pointed out and provided documentation and reasonable explanations for.

@shramov you could just return a success bool and short circuit the calling function to maintain previous behavior and if you really feel strongly over not hard failing open another PR around that particular issue.

apologies for just randomly dropping my 2c.

PR #4156, opened 2019-05-14
Add infos to help using for_window to man 5

Nomeji:
Infos about how to get criteria was nowhere to be found on the man page.

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -770,17 +770,20 @@ Mark all Firefox windows with "Browser":
 [class="Firefox"] mark Browser
 ```
 
+You may like to use swaymsg -t get_tree for finding the values of these 
+properties in practice for your applications.
+
 The following attributes may be matched with:
 
 *app_id*
 	Compare value against the app id. Can be a regular expression. If value is
 	\_\_focused\_\_, then the app id must be the same as that of the currently
-	focused window.
+	focused window. _app_id_ are specific to Wayland applications.
 
 *class*
 	Compare value against the window class. Can be a regular expression. If
 	value is \_\_focused\_\_, then the window class must be the same as that of
-	the currently focused window.
+	the currently focused window. _class_ are specific to X11 applications.
 
 *con_id*
 	Compare against the internal container ID, which you can find via IPC. If


sway/sway.5.scd
```diff
@@ -546,6 +546,8 @@ The default colors are:
 
 *for_window* <criteria> <command>
 	Whenever a window that matches _criteria_ appears, run list of commands.
+	A list of _criteria_ for currently opened windows may be obtained via 
+	swaymsg -t get_tree.
```
emersion: Hmm, not really: `swaymsg -t get_tree` doesn't return criteria (ie. conditions), it returns window properties.

This addition should probably be moved to the *CRITERIA* section (since criteria isn't only used in `for_window`), and say that attributes for currently opened windows may be obtained via `swaymsg -t get_tree`.

Nomeji: Okay edited

ddevault:
I think what @emersion was getting at (and I agree with) is that this would be better mentioned just once in the opening paragraph. Something like "swaymsg -t get_tree may be useful for finding the values of these properties in practice for your applications."

Nomeji:
> I think what @emersion was getting at (and I agree with) is that this would be better mentioned just once in the opening paragraph. Something like "swaymsg -t get_tree may be useful for finding the values of these properties in practice for your applications."

Okay I think I got it. Should be good now.

sway/sway.5.scd
```diff
@@ -770,17 +770,19 @@ Mark all Firefox windows with "Browser":
...
 
+You may like to use swaymsg -t get_tree for finding the values of these properties in practice for your applications.
+
 The following attributes may be matched with:
 
 *app_id*
 	Compare value against the app id. Can be a regular expression. If value is
 	\_\_focused\_\_, then the app id must be the same as that of the currently
-	focused window.
+	focused window. _app_id_ are specific to Wayland applications.
```
ddevault: The similar comment on *class* is fine, but since sway is a Wayland compositor we needn't document that it has Wayland-specific behavior.

sway/sway.5.scd
```diff
@@ -770,17 +770,19 @@ Mark all Firefox windows with "Browser":
 [class="Firefox"] mark Browser
 ```
 
+You may like to use swaymsg -t get_tree for finding the values of these properties in practice for your applications.
```
ddevault: Wrap to 80 cols

sway/sway.5.scd
emersion: I think it's still useful to say it's only for Wayland applications, otherwise users will expect app_id to work with X11 clients too.

Nomeji: I wrote this to help users understand that if they cannot find _class_ using swaymsg -t get_tree is probably because it's a Wayland application. It something that trouble me when I tried to use for_window. I think that since this is a pretty new behaviour (especially for users coming from i3) it could be good to document it.

Is it not relevant or is it not explicit enough?

ddevault: Nevermind, this is fine as-is (not including the 80 col issue)

emersion:
For some reason the last push triggered the builds.sr.ht webhook but it replied: "Got update, but there are no new commits".

ddevault:
Thanks!

I used my powers of using a bandaid instead of debugging the real problem to fix it.

PR #4158, opened 2019-05-16
bar: fix segfault with missing or invalid bar id

alyssais:
Prior to this patch, if I ran something like this, sway would crash:

    swaymsg bar height 50

or

    swaymsg bar not-a-bar-id color bg #ff0000

This was in contrast to other bar subcommands, like status_command, which would exit with a "No bar defined message".

The difference between the subcommands that crashed and the ones that exited was that some subcommands had a check to see if a bar was specified, while others just assumed that it had been and carried on until they segfaulted.

Because this check was identical in every subcommand it was present in, and I couldn't think of a case where it would be valid to run a bar subcommand without specifying which bar to apply it to, I moved this check from individual subcommands into the bar command, which is already responsible for actually setting the specified bar. This reduced code duplication, and fixed the crash for the subcommands that were missing this check.

### DIFF
--- a/sway/commands/bar.c
+++ b/sway/commands/bar.c
@@ -82,26 +82,30 @@ struct cmd_results *cmd_bar(int argc, char **argv) {
 		++argv; --argc;
 	}
 
-	if (!config->current_bar && config->reading) {
-		// Create new bar with default values
-		struct bar_config *bar = default_bar_config();
-		if (!bar) {
-			return cmd_results_new(CMD_FAILURE,
-					"Unable to allocate bar state");
-		}
+	if (!config->current_bar) {
+		if (config->reading) {
+			// Create new bar with default values
+			struct bar_config *bar = default_bar_config();
+			if (!bar) {
+				return cmd_results_new(CMD_FAILURE,
+						"Unable to allocate bar state");
+			}
 
-		// set bar id
-		const int len = snprintf(NULL, 0, "bar-%d", config->bars->length - 1) + 1;
-		bar->id = malloc(len * sizeof(char));
-		if (bar->id) {
-			snprintf(bar->id, len, "bar-%d", config->bars->length - 1);
+			// set bar id
+			int len = snprintf(NULL, 0, "bar-%d", config->bars->length - 1) + 1;
+			bar->id = malloc(len * sizeof(char));
+			if (bar->id) {
+				snprintf(bar->id, len, "bar-%d", config->bars->length - 1);
+			} else {
+				return cmd_results_new(CMD_FAILURE, "Unable to allocate bar ID");
+			}
+
+			// Set current bar
+			config->current_bar = bar;
+			sway_log(SWAY_DEBUG, "Creating bar %s", bar->id);
 		} else {
-			return cmd_results_new(CMD_FAILURE, "Unable to allocate bar ID");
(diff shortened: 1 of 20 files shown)


RedSoxFan:
> Because this check was identical in every subcommand it was present in

The condition was being used as a non-fatal assert in each of the subcommand handlers. I think I would rather have the check added to the subcommand handlers that are missing it, but if the others are fine with this change, I'm not going to block it. If this does get merged, the change should be mirrored to the input, output, and seat subcommands for consistency.

alyssais:
> The condition was being used as a non-fatal assert in each of the subcommand handlers.

Could you explain what you mean by this?

> I think I would rather have the check added to the subcommand handlers that are missing it

I’m concerned that that would make it far more likely that this crash would be reintroduced, as soon as a new subcommand was added.

RedSoxFan:
> > The condition was being used as a non-fatal assert in each of the subcommand handlers.
> 
> Could you explain what you mean by this?

The condition is just a `NULL`-check. By non-fatal assert, I just meant it returns an error instead of aborting with `SIGABRT` like an `assert` would do.

> > I think I would rather have the check added to the subcommand handlers that are missing it
> 
> I’m concerned that that would make it far more likely that this crash would be reintroduced, as soon as a new subcommand was added.

I don't see it any different than any other `NULL`-check.

ddevault:
I'm okay with doing it in a single place

sway/commands/bar.c
```diff
@@ -82,26 +82,30 @@ struct cmd_results *cmd_bar(int argc, char **argv) {
...
+		if (config->reading) {
+			// Create new bar with default values
+			struct bar_config *bar = default_bar_config();
+			if (!bar) {
+				return cmd_results_new(CMD_FAILURE,
+						"Unable to allocate bar state");
+			}
+
+			// set bar id
+			const int len = snprintf(NULL, 0, "bar-%d", config->bars->length - 1) + 1;
```
RedSoxFan: This is over the 80 column limit. It doesn't look like the `const` is actually needed here so that can just get dropped. It'll still be over by 1 character, but to maintain readability, that's close enough.

sway/commands/bar/pango_markup.c
```diff
@@ -9,10 +9,7 @@ struct cmd_results *bar_cmd_pango_markup(int argc, char **argv) {
 	if ((error = checkarg(argc, "pango_markup", EXPECTED_EQUAL_TO, 1))) {
 		return error;
 	}
-	if (!config->current_bar) {
-		return cmd_results_new(CMD_FAILURE, "No bar defined.");
-	}
-	config->current_bar->pango_markup 
+	config->current_bar->pango_markup
```
RedSoxFan: nit: since the trailing space is being striped from this line anyway, can you move the `=` up to this line?

alyssais:
> +			const int len = snprintf(NULL, 0, "bar-%d", config->bars->length - 1) + 1;
>
> This is over the 80 column limit. It doesn't look like the `const` is actually needed here so that can just get dropped. It'll still be over by 1 character, but to maintain readability, that's close enough.

It's now exactly 80 by my count.

> -	config->current_bar->pango_markup
> +	config->current_bar->pango_markup
>
> nit: since the trailing space is being striped from this line anyway, can you move the `=` up to this line?

Done!

RedSoxFan:
Thanks!

PR #4160, opened 2019-05-17
commands/bar: improve error for invalid subcommand

alyssais:
Running a command like this produced a confusing error message:

    $ swaymsg bar bar-0 colors background #ff0000
    Error: Unknown/invalid command 'bar-0'

This patch makes the error message use argv[1] instead of argv[0] (from config_subcommand's implementation), so it actually uses the name of the command, rather than the id of the bar.

### DIFF
--- a/sway/commands/bar.c
+++ b/sway/commands/bar.c
@@ -80,6 +80,13 @@ struct cmd_results *cmd_bar(int argc, char **argv) {
 		}
 		config->current_bar = bar;
 		++argv; --argc;
+	} else if (!config->reading) {
+		if (is_subcommand(argv[0])) {
+			return cmd_results_new(CMD_INVALID, "No bar defined.");
+		} else {
+			return cmd_results_new(CMD_INVALID,
+					"Unknown/invalid command '%s'", argv[1]);
+		}
 	}
 
 	if (!config->current_bar) {
@@ -103,8 +110,6 @@ struct cmd_results *cmd_bar(int argc, char **argv) {
 			// Set current bar
 			config->current_bar = bar;
 			sway_log(SWAY_DEBUG, "Creating bar %s", bar->id);
-		} else {
-			return cmd_results_new(CMD_FAILURE, "No bar defined.");
 		}
 	}
 


alyssais:
On Thu, May 16, 2019 at 06:49:23PM -0700, Brian Ashworth wrote:
> RedSoxFan requested changes on this pull request.
>
> This produces false errors for bar blocks in the config since ids are not required in the config. Changing it to a `else if (!config->reading)` would work, but could be equally confusing when the bar id is missing at runtime. The following might fix that:
> ```c
> } else if (!config->reading) {
>     if (is_subcommand(argv[0])) {
>         // Missing bar id
>     } else {
>        // Unknown/invalid command
>     }
> }
> ```

Makes sense, thanks.

In that case I'll wait to see what happens with #4158, since it would
probably conflict otherwise.

alyssais:
> RedSoxFan requested changes on this pull request.
>
> This produces false errors for bar blocks in the config since ids are not required in the config. Changing it to a `else if (!config->reading)` would work, but could be equally confusing when the bar id is missing at runtime. The following might fix that:
> ```c
> } else if (!config->reading) {
>     if (is_subcommand(argv[0])) {
>         // Missing bar id
>     } else {
>        // Unknown/invalid command
>     }
> }
> ```

Done!

This made the missing bar check further down dead code, so I've removed
that (or moved it up here -- however you want to think about it).

sway/commands/bar.c
```diff
@@ -80,6 +80,12 @@ struct cmd_results *cmd_bar(int argc, char **argv) {
 		}
 		config->current_bar = bar;
 		++argv; --argc;
+	} else if (!config->reading) {
+		if (is_subcommand(argv[0])) {
+			return cmd_results_new(CMD_FAILURE, "No bar defined.");
```
RedSoxFan: This should actually be `CMD_INVALID` since it is a syntax issue

sway/commands/bar.c
```diff
@@ -80,6 +80,12 @@ struct cmd_results *cmd_bar(int argc, char **argv) {
...
+		} else {
+			return cmd_results_new(CMD_INVALID, "Unknown/invalid command '%s'", argv[1]);
```
RedSoxFan: style: 80 column limit

alyssais:
> RedSoxFan requested changes on this pull request.

Done.

RedSoxFan:
Thanks!

PR #4166, opened 2019-05-21
Make comment casing consistient

panaman67:
Makes the first letter of all the comments uppercase. THus making the config file more consistient


ddevault:
Thanks!

PR #4169, opened 2019-05-21
Implement wlr-output-management-v1

atx:


sway/desktop/output.c
```diff
@@ -651,4 +673,72 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
+		struct wlr_output *wlr_output = config_head->state.output;
+		struct sway_output *output = wlr_output->data;
+		if (!output->enabled || config_head->state.enabled) {
+			continue;
+		}
+		struct output_config *oc = new_output_config("temp");
+		oc->enabled = false;
+		// There is no explicit way to set DPMS through the protocol.
+		// Users probably expect that disabling a monitor will turn it off.
+		oc->dpms_state = DPMS_OFF;
```
emersion: No need to touch DPMS, setting `enabled = false` is enough. DPMS is deliberately left out of the protocol.

sway/desktop/output.c
```diff
@@ -651,4 +673,72 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
+		oc->x = config_head->state.x;
+		oc->y = config_head->state.y;
+		oc->transform = config_head->state.transform;
+		oc->scale = config_head->state.scale;
+
+		apply_output_config(oc, output);
+		free_output_config(oc);
+	}
+
+	// Handle failure somehow?
```
emersion: Could make `apply_output_config` return a bool, if it doesn't do that yet

sway/desktop/output.c
```diff
@@ -578,6 +578,28 @@ static void update_textures(struct sway_container *con, void *data) {
...
+			continue;
+		}
+		struct wlr_output_configuration_head_v1 *config_head =
+			wlr_output_configuration_head_v1_create(config, output->wlr_output);
+		struct wlr_box *output_box = wlr_output_layout_get_box(
+			root->output_layout, output->wlr_output);
+		if (output_box) {
+			config_head->state.x = output_box->x;
+			config_head->state.y = output_box->y;
+		}
```
emersion: We also need to handle DPMS here. We don't want to mark the output as disabled in the protocol if it's just DPMS'ed off.

The easiest way is probably to check whether the `sway_output` is enabled (does it have an `enabled` field?) and overwrite `config_head->enabled`.

sway/desktop/output.c
```diff
@@ -651,4 +673,72 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
+	}
+
+	// Then enable outputs that need to
+	wl_list_for_each(config_head, &config->heads, link) {
+		struct wlr_output *wlr_output = config_head->state.output;
+		struct sway_output *output = wlr_output->data;
+		if (!config_head->state.enabled) {
+			continue;
+		}
+		struct output_config *oc = new_output_config("temp");
```
emersion: Instead of "temp" we could use the output name? I don't remember what this string is used for TBH.

sway/desktop/output.c
atx: Oh, right, was reading the wrong code path in apply_output_config

sway/desktop/output.c
atx: It already does, my bad

sway/desktop/output.c
atx: Doesn't seem to be relevant here as far as I can tell - I used temp as that was what I saw elsewhere in the code.

atx:
Pushed an updated version.

The failure path is still a bit meh - maybe everything should roll back instead of partially applying a config?

sway/desktop/output.c
RedSoxFan: It's used when merging output configs in (`store_output_config`) and determining which output to apply the config to (in `apply_output_config_to_outputs`). Currently, this config is not being stored so if the output command is ran for the output (via swaymsg or a binding) it will revert everything set by this config to whatever is stored

emersion: I guess we need to store the output config too then. Not sure which identifier to use though.

emersion:
>The failure path is still a bit meh - maybe everything should roll back instead of partially applying a config?

The wlroots API won't make this easy. I'd suggest we just wait for wlroots to gain an atomic API instead: https://github.com/swaywm/wlroots/issues/1640

atx:
I have added the calls to `store_output_config`. This seems to fix output command overwriting the changes configured over this protocol.

emersion:
Thanks!

PR #4171, opened 2019-05-22
commands/bar: fix mode and hidden_state at runtime

RedSoxFan:
This fixes a regression that I missed when reviewing #4160 

For compatibility with i3, `bar mode` and `bar hidden_state` do not
require bar-ids (in the normal location) at runtime since they follow
the alternative syntax: `bar mode|hidden_state <option> [<bar-id>]`

This removes the incorrect error that the bar-id is missing for those
two bar subcommands


emersion:
Thanks!

PR #4173, opened 2019-05-23
Perform keymap validation

### LINKED ISSUE #4148: Invalid xkb options result in unusable keyboard and therefore no way to exit sway

RedSoxFan:
Closes #4148 (this is the error part; see #4153 for something usable)

**Just the first commit**: basic invalid keymap prevention

Before the delta input config is stored, this attempts to compile a
keymap with it. If the keymap fails to compile, then the first line of
the xkbcommon log entry will be included with a `CMD_FAILURE`, the
entire xkbcommon log entry will be included in the sway error log, and
the delta will not be stored.

This will only handle basic issues such as a layouts not existing. This
will NOT catch more complex issues such as when a variant does
exist, but not for the given layout (ex: `azerty` is a valid variant,
but the `us` layout does not have a `azerty` variant).

**Both commits combined**: full invalid keymap protection

This allows for an optional validation stage of the input config before
storing it. Currently, the only validation performed is a valid xkb
configuration. The validation stage will do a dry run of the storage
process and make sure that storing the input config will not result in
any invalid xkb configurations. If there are any xkb keymap compilation
errors, the input config will not be stored, the full xkbcommon log
will be rewritten to the sway log, and the first line will be stored in
the error variable passed to store_input_config. The error is then used
to provide a little bit of information for the CMD_FAILURE.

### DIFF
--- a/sway/commands/input.c
+++ b/sway/commands/input.c
@@ -88,9 +88,10 @@ struct cmd_results *cmd_input(int argc, char **argv) {
 
 	if (!res || res->status == CMD_SUCCESS) {
 		char *error = NULL;
-		struct xkb_keymap *keymap = sway_keyboard_compile_keymap(
-				config->handler_context.input_config, &error);
-		if (!keymap) {
+		struct input_config *ic =
+			store_input_config(config->handler_context.input_config, &error);
+		if (!ic) {
+			free_input_config(config->handler_context.input_config);
 			if (res) {
 				free_cmd_results(res);
 			}
@@ -99,10 +100,6 @@ struct cmd_results *cmd_input(int argc, char **argv) {
 			free(error);
 			return res;
 		}
-		xkb_keymap_unref(keymap);
-
-		struct input_config *ic =
-			store_input_config(config->handler_context.input_config);
 
 		input_manager_apply_input_config(ic);
 		retranslate_keysyms(ic);
--- a/sway/commands/input.c
+++ b/sway/commands/input.c
@@ -2,6 +2,7 @@
 #include <strings.h>
 #include "sway/commands.h"
 #include "sway/input/input-manager.h"
+#include "sway/input/keyboard.h"
 #include "log.h"
 #include "stringop.h"
 
@@ -86,6 +87,20 @@ struct cmd_results *cmd_input(int argc, char **argv) {
 	}
 
 	if (!res || res->status == CMD_SUCCESS) {
+		char *error = NULL;
+		struct xkb_keymap *keymap = sway_keyboard_compile_keymap(
+				config->handler_context.input_config, &error);
+		if (!keymap) {
+			if (res) {
+				free_cmd_results(res);
+			}
+			res = cmd_results_new(CMD_FAILURE, "Failed to compile keymap: %s",
+					error ? error : "(details unavailable)");
+			free(error);
+			return res;
+		}
+		xkb_keymap_unref(keymap);
+
 		struct input_config *ic =
 			store_input_config(config->handler_context.input_config);
 
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -495,7 +495,45 @@ struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
 	return keyboard;
 }
 
-struct xkb_keymap *sway_keyboard_compile_keymap(struct input_config *ic) {
+static void handle_xkb_context_log(struct xkb_context *context,
+		enum xkb_log_level level, const char *format, va_list args) {
+	va_list args_copy;
+	va_copy(args_copy, args);
+	size_t length = vsnprintf(NULL, 0, format, args_copy) + 1;
+	va_end(args_copy);
+
+	char *error = malloc(length);
+	if (!error) {
+		sway_log(SWAY_ERROR, "Failed to allocate libxkbcommon log message");
+		return;
+	}
+
+	va_copy(args_copy, args);
+	vsnprintf(error, length, format, args_copy);
+	va_end(args_copy);
+
+	if (error[length - 2] == '\n') {
+		error[length - 2] = '\0';
+	}
+
+	sway_log_importance_t importance = SWAY_DEBUG;
+	if (level <= XKB_LOG_LEVEL_ERROR) { // Critical and Error
+		importance = SWAY_ERROR;
+	} else if (level <= XKB_LOG_LEVEL_INFO) { // Warning and Info
+		importance = SWAY_INFO;
+	}
+	sway_log(importance, "[xkbcommon] %s", error);
+
+	char **data = xkb_context_get_user_data(context);
+	if (importance == SWAY_ERROR && data && !*data) {
+		*data = error;
+	} else {
+		free(error);
+	}
+}
+
+struct xkb_keymap *sway_keyboard_compile_keymap(struct input_config *ic,
+		char **error) {
 	struct xkb_rule_names rules = {0};
 	if (ic) {
 		input_config_fill_rule_names(ic, &rules);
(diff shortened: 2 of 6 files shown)


sway/commands/input.c
```diff
@@ -86,6 +87,15 @@ struct cmd_results *cmd_input(int argc, char **argv) {
 	}
 
 	if (!res || res->status == CMD_SUCCESS) {
+		char *error = NULL;
+		struct xkb_keymap *keymap = sway_keyboard_compile_keymap(
+				config->handler_context.input_config, &error);
+		if (!keymap) {
+			return cmd_results_new(CMD_FAILURE, "Failed to compile keymap: %s",
+					error ? error : "(null)");
```
shramov: Is not 'error' leaking here?
Maybe it's better to log errors when they appear?

RedSoxFan: > Is not 'error' leaking here?

Fixed

> Maybe it's better to log errors when they appear?

The errors are being logged when they appear. There is to provide a quick description in swaymsg/swaynag  for user convenience.

shramov: Also it's better to print something more descriptive instead of '(null)', very confusing

RedSoxFan: I'll change it to `(error allocation failed)` when I get back to my laptop

shramov: Maybe it's not allocation error, maybe callback was not called?
Or on any error it will be called?

sway/input/keyboard.c
```diff
@@ -476,13 +476,47 @@ struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
...
+
+	va_copy(args_copy, args);
+	vsnprintf(error, length, format, args_copy);
+	va_end(args_copy);
+
+	if (error[length - 2] == '\n') {
+		error[length - 2] = '\0';
+	}
+
+	sway_log(SWAY_ERROR, "(xkbcommon) %s", error);
```
emersion: Maybe we can use `level` to choose the sway log level?

emersion: Nit: so far we've prefixed error messages coming from libs with `[name]` instead of `(name)`.

sway/input/keyboard.c
```diff
@@ -476,13 +476,47 @@ struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
...
+
+	char **data = xkb_context_get_user_data(context);
+	if (data && !*data) {
+		*data = error;
```
emersion: We should probably only fill this if the log level is error

sway/commands/input.c
RedSoxFan: I used `(details unavailable)`, which isn't really helpful either, but shouldn't be confusing. The two causes would be the allocation failed or xkbcommon just didn't log anything (for whatever reason).

sway/input/keyboard.c
RedSoxFan: Done

RedSoxFan: Noted and changed

sway/input/keyboard.c
RedSoxFan: Done

RedSoxFan:
I updated the second commit to something nicer

sway/input/keyboard.c
```diff
@@ -495,13 +495,53 @@ struct sway_keyboard *sway_keyboard_create(struct sway_seat *seat,
...
+
+struct xkb_keymap *sway_keyboard_compile_keymap(struct input_config *ic,
+		char **error) {
 	struct xkb_rule_names rules = {0};
 	if (ic) {
 		input_config_fill_rule_names(ic, &rules);
 	}
 
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
+	xkb_context_set_user_data(context, error);
```
emersion: Can we unset user data at the end of the function? Just to be sure it won't write at an invalid address.

RedSoxFan: Done

emersion:
First commit LGTM. The second one seems a little bit complicated.

RedSoxFan:
> First commit LGTM. The second one seems a little bit complicated.

This first commit is simple, but will only catch basic stuff like typos. The second commit catches more (including, for example, layout and variant mismatches), but I'm not really sure how to make the implementation better. I guess it depends on how far we want to go

RedSoxFan:
>  I guess it depends on how far we want to go

Any opinions on this? I think it may be more helpful to users to have some of the information shown by swaynag/swaymsg as opposed to being "hidden" in the debug log and them thinking it is just not working. However, I do understand that the second commit adds a lot of complexity to how input configs are stored.

The first commit is pretty straight forward so that should be fine, but I can remove the second commit if needed. Alternatively, if anyone has any ideas for making it simpler, that could work too.

emersion:
Can't we merge configs, then sanity-check the whole thing, then apply it if it's fine? I don't understand why callbacks are needed.

RedSoxFan:
> Can't we merge configs, then sanity-check the whole thing, then apply it if it's fine? I don't understand why callbacks are needed.

We can't just merge the configs without some way of rolling it back. Otherwise, it leaves the input configs in an invalid state anyway.

emersion:
Can't we merge the existing config into a temporary config, then merge the new config on top, then sanity-check it, and then merge the temporary config on top on the existing config?

RedSoxFan:
> Can't we merge the existing config into a temporary config, then merge the new config on top, then sanity-check it, and then merge the temporary config on top on the existing config?

For named input configs, yes. For type and wildcard configs, no since that may partially propagate it. But with that said, the dry run can actually be restricted to just type and wildcard related logic, which can simplify the second commit. I'll update the PR shortly

RedSoxFan:
I reworked the second commit again. I think this version is better.

emersion:
Thanks!

PR #4181, opened 2019-05-24
Update output manager config on all output events

atx:
The output manager config was not properly updated when output
configuration was changed through some other means (such as a command).


emersion:
Thanks!

PR #4187, opened 2019-05-27
bindings: allow unlocked and locked bindings

### LINKED ISSUE #4183: Allow bindsym --locked for existing bindsyms
My keyboard has only one key for play/pause, and in most cases this runs `playerctl play-pause`. However, I'd like to have it run `playerctl pause` *only* when on a lock screen, e.g. let you pause the player but not start it back up again until locked.

For this, it'd be nice if we had an option to `bindsym` the version for an unlocked screen in addition to `bindsym --locked`, so that we can run different scripts when on the lock screen and not. I figure that there are more examples than this one but it'd be a cool functionality to add regardless.

RedSoxFan:
_Preface: I noticed while working on this that I missed that
`sway/input/switch.c` is space indented when I approved the PR that
added it. I'll submit a follow-up PR to correct the indentation of the
file (submitted as #4188)_

Closes #4183 

This changes the behavior of bindings to make the `BINDING_LOCKED` flag
conflicting, which will allow for both unlocked and locked bindings.

If there are two matching bindings and one has `--locked` and the other
does not, the one with `--locked` will be preferred when locked and
the one without will be preferred when unlocked.

If there are two matching bindings and one has both a matching
`--input-device=<input>` and `--locked` and the other has neither, the
former will be preferred for both unlocked and locked.

This also refactors `get_active_binding` in `sway/input/keyboard.c`
to make it easier to read.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -146,7 +146,7 @@ static void get_active_binding(const struct sway_shortcut_state *state,
 		uint32_t modifiers, bool release, bool locked, const char *input) {
 	for (int i = 0; i < bindings->length; ++i) {
 		struct sway_binding *binding = bindings->items[i];
-		bool binding_locked = binding->flags & BINDING_LOCKED;
+		bool binding_locked = (binding->flags & BINDING_LOCKED) != 0;
 		bool binding_release = binding->flags & BINDING_RELEASE;
 
 		if (modifiers ^ binding->modifiers ||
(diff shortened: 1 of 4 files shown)


tokyovigilante:
You tab-indenting heathens.... ;) Sorry about that.

emersion:
Needs rebase

RedSoxFan:
> Needs rebase

Done

sway/input/keyboard.c
```diff
@@ -178,18 +178,42 @@ static void get_active_binding(const struct sway_shortcut_state *state,
...
-		} else if (!*current_binding ||
-				strcmp((*current_binding)->input, "*") == 0) {
-			*current_binding = binding;
-
-			if (strcmp((*current_binding)->input, input) == 0) {
-				// If a binding is found for the exact input, quit searching
-				return;
+		if (*current_binding) {
+			if (*current_binding == binding) {
+				if ((*current_binding)->order != binding->order) {
```
emersion: Hmm, I'm confused. How can this happen if `*current_binding == binding`?

RedSoxFan: I just removed the check since it shouldn't happen. I think I read the previous condition wrong when refactoring

emersion:
Thanks!

PR #4189, opened 2019-05-27
common/ipc-client: remove ipc recv timeout log

RedSoxFan:
Supersedes/Closes #4092 
See #4092 for compilation logs showing the issue

This just removes the ipc recv timeout log statement in
`ipc_recv_set_timeout`. The `tv_sec` field of `struct timeval` has
varying types and/or sizes depending on the platform and architecture.
On some of these, the current format string will cause compilation
errors. Additionally, the log statement is not extremely useful and the
function is currently only used by swaymsg, which has a hardcoded log
level that will prevent it from even being shown, so there is no point
in even keeping it.


emersion:
Thanks!

PR #4202, opened 2019-06-01
Implement XCursor theme configuration

dnkl:
This adds configuration commands that configures the XCursor theme and
size. Configuration is done through a new seat command (the size is optional and defaults to 24):
```
seat * xcursor_theme Adwaita [48]
```
The configured theme name and size is used as the default pointer
theme (set in `input/seat.c`), and exported through the `XCURSOR_THEME` and `XCURSOR_SIZE`
environment variables, and used as the default XWayland xcursor theme.

Marked as WIP since it's missing, at the very least, documentation updates. I would like to get some feedback first though, before I get to that.

## Major Changes
The biggest changes (in addition to the new feature in itself) are:
* `XCURSOR_THEME` and `XCURSOR_SIZE` are now exported by the default seat, every time its xcursor configuration is changed. Thus all applications (re)started after a runtime change will see the new settings.
* The XWayland xcursor theme manager is no longer instantiated in `server_start()`. This is instead done by the default seat (and re-created when the theme is changed).

## Questions:
1) is a global configuration command the preferred option? Or should it be somewhere else?
2) I know that we're still missing the protocols needed to correctly propagate the theme/size to clients, but in the mean time, should swaybar/nag etc be updated to check the `XCURSOR_*` environment variables?

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -19,6 +19,7 @@
 #include "sway/ipc-server.h"
 #include "sway/layers.h"
 #include "sway/output.h"
+#include "sway/server.h"
 #include "sway/tree/arrange.h"
 #include "sway/tree/container.h"
 #include "sway/tree/root.h"
@@ -722,17 +723,72 @@ void seat_remove_device(struct sway_seat *seat,
 	seat_update_capabilities(seat);
 }
 
+static bool xcursor_manager_is_named(const struct wlr_xcursor_manager *manager,
+		const char *name) {
+	return (!manager->name && !name) ||
+		(name && manager->name && strcmp(name, manager->name) == 0);
+}
+
 void seat_configure_xcursor(struct sway_seat *seat) {
-	// TODO configure theme and size
+	unsigned cursor_size = 24;
 	const char *cursor_theme = NULL;
 
-	if (!seat->cursor->xcursor_manager) {
-		seat->cursor->xcursor_manager =
-			wlr_xcursor_manager_create(cursor_theme, 24);
-		if (sway_assert(seat->cursor->xcursor_manager,
-					"Cannot create XCursor manager for theme")) {
-			return;
+	const struct seat_config *seat_config = seat_get_config(seat);
+	if (!seat_config) {
+		seat_config = seat_get_config_by_name("*");
+	}
+	if (seat_config) {
+		cursor_size = seat_config->xcursor_theme.size;
+		cursor_theme = seat_config->xcursor_theme.name;
+	}
+
+	if (seat == input_manager_get_default_seat()) {
+		char cursor_size_fmt[16];
+		snprintf(cursor_size_fmt, sizeof(cursor_size_fmt), "%d", cursor_size);
+		setenv("XCURSOR_SIZE", cursor_size_fmt, 1);
+		if (cursor_theme != NULL) {
+			setenv("XCURSOR_THEME", cursor_theme, 1);
 		}
+
+#if HAVE_XWAYLAND
+		if (!server.xwayland.xcursor_manager ||
+				!xcursor_manager_is_named(server.xwayland.xcursor_manager,
+					cursor_theme) ||
+				server.xwayland.xcursor_manager->size != cursor_size) {
+
+			wlr_xcursor_manager_destroy(server.xwayland.xcursor_manager);
+
+			server.xwayland.xcursor_manager =
+				wlr_xcursor_manager_create(cursor_theme, cursor_size);
+			sway_assert(server.xwayland.xcursor_manager,
+						"Cannot create XCursor manager for theme");
+
+			wlr_xcursor_manager_load(server.xwayland.xcursor_manager, 1);
+			struct wlr_xcursor *xcursor = wlr_xcursor_manager_get_xcursor(
+				server.xwayland.xcursor_manager, "left_ptr", 1);
+			if (xcursor != NULL) {
+				struct wlr_xcursor_image *image = xcursor->images[0];
+				wlr_xwayland_set_cursor(
+					server.xwayland.wlr_xwayland, image->buffer,
+					image->width * 4, image->width, image->height,
+					image->hotspot_x, image->hotspot_y);
+			}
+		}
+#endif
+	}
+
+	/* Create xcursor manager if we don't have one already, or if the
+	 * theme has changed */
+	if (!seat->cursor->xcursor_manager ||
+			!xcursor_manager_is_named(
+				seat->cursor->xcursor_manager, cursor_theme) ||
+			seat->cursor->xcursor_manager->size != cursor_size) {
+
+		wlr_xcursor_manager_destroy(seat->cursor->xcursor_manager);
+		seat->cursor->xcursor_manager =
+			wlr_xcursor_manager_create(cursor_theme, cursor_size);
+		sway_assert(seat->cursor->xcursor_manager,
+					"Cannot create XCursor manager for theme");
 	}
 
 	for (int i = 0; i < root->outputs->length; ++i) {
--- a/sway/server.c
+++ b/sway/server.c
@@ -168,17 +168,6 @@ void server_fini(struct sway_server *server) {
 }
 
 bool server_start(struct sway_server *server) {
-	// TODO: configurable cursor theme and size
-	int cursor_size = 24;
-	const char *cursor_theme = NULL;
-
-	char cursor_size_fmt[16];
-	snprintf(cursor_size_fmt, sizeof(cursor_size_fmt), "%d", cursor_size);
-	setenv("XCURSOR_SIZE", cursor_size_fmt, 1);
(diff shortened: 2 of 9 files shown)


emersion:
Thanks for this PR! It's one of the features I get asked about pretty frequently.

>is a global configuration command the preferred option? Or should it be somewhere else?

I believe a seat command would be better.

>should swaybar/nag etc be updated to check the XCURSOR_* environment variables?

Yeah, I think that would be a good idea.

dnkl:
> I believe a seat command would be better.

I have this mostly working now, but is somewhat stuck on a seemingly small problem; in `server_start()`, the `XCURSOR_*` environment variables are set, and xwayland is configured.

What I'm trying to do is use the default seat's configuration as source:
```
bool server_start(struct sway_server *server) {
	struct sway_seat *default_seat = input_manager_get_default_seat();
	struct seat_config *default_seat_config = seat_get_config(default_seat);
```
 This works, but *only* if I explicitly configure `seat0` in the config:
```
seat seat0 xcursor_theme Adwaita
```
Using a wildcard does not work:
```
seat * xcursor_theme Adwaita
```
Any suggestions? Other seat commands appear to apply to the default seat when a wildcard is used...

dnkl:
Found it. I hadn't updated `merge_seat_config()`. Sorry about the noise :)

sway/server.c
```diff
@@ -170,10 +170,22 @@ void server_fini(struct sway_server *server) {
 }
 
 bool server_start(struct sway_server *server) {
-	// TODO: configurable cursor theme and size
+	/* This ensures we have a configuration for the default seat */
+	input_manager_verify_fallback_seat();
+
```
dnkl: @emersion looking for feedback on this; I doubt this is the best way to do it...

Background: at this point, there hasn't yet been an explicit `seat0` conf created, only the `*` is available (if you configured the cursor theme with `seat * xcursor_theme ...`).

Thus, the `seat_get_config()` call below return `NULL` without the call to `input_manager_verify_fallback_seat()`.

Note: I picked this particular call by breaking in gdb when the `seat0` conf was created, and checking the call path...

dnkl: I chose another approach, where `server_start()` only sets the default theme. When the default seat's xcursor theme is configured, we export the environment variables there instead (and update XWayland).

dnkl:
"WIP" removed. I think this is ready for review now.

sway/input/seat.c
```diff
@@ -722,17 +723,66 @@ void seat_remove_device(struct sway_seat *seat,
...
+	int cursor_size = 24;
 	const char *cursor_theme = NULL;
 
-	if (!seat->cursor->xcursor_manager) {
-		seat->cursor->xcursor_manager =
-			wlr_xcursor_manager_create(cursor_theme, 24);
-		if (sway_assert(seat->cursor->xcursor_manager,
-					"Cannot create XCursor manager for theme")) {
-			return;
+	const struct seat_config *seat_config = seat_get_config(seat);
```
RedSoxFan: `seat_get_config` will not match the wildcard config so you'll need to add
```c
if (!seat_config) {
    seat_config = seat_get_config_by_name("*");
}
```

sway/input/seat.c
```diff
@@ -722,17 +723,66 @@ void seat_remove_device(struct sway_seat *seat,
...
+		snprintf(cursor_size_fmt, sizeof(cursor_size_fmt), "%d", cursor_size);
+		setenv("XCURSOR_SIZE", cursor_size_fmt, 1);
+		if (cursor_theme != NULL) {
+			setenv("XCURSOR_THEME", cursor_theme, 1);
 		}
+
+#if HAVE_XWAYLAND
+		if (!server.xwayland.xcursor_manager ||
+			!xcursor_manager_is_named(server.xwayland.xcursor_manager,
+                                      cursor_theme)) {
```
RedSoxFan: style: don't line up the arguments. just do an extra indentation level to differentiate it from the body. also, use tabs not spaces.

sway/input/seat.c
```diff
@@ -722,17 +723,66 @@ void seat_remove_device(struct sway_seat *seat,
...
+						"Cannot create XCursor manager for theme");
+
+			wlr_xcursor_manager_load(server.xwayland.xcursor_manager, 1);
+			struct wlr_xcursor *xcursor = wlr_xcursor_manager_get_xcursor(
+				server.xwayland.xcursor_manager, "left_ptr", 1);
+			if (xcursor != NULL) {
+				struct wlr_xcursor_image *image = xcursor->images[0];
+				wlr_xwayland_set_cursor(
+					server.xwayland.wlr_xwayland, image->buffer,
+					image->width * 4, image->width, image->height, image->hotspot_x,
```
RedSoxFan: style: 80 columns

sway/input/seat.c
```diff
@@ -722,17 +723,66 @@ void seat_remove_device(struct sway_seat *seat,
...
+		char cursor_size_fmt[16];
+		snprintf(cursor_size_fmt, sizeof(cursor_size_fmt), "%d", cursor_size);
+		setenv("XCURSOR_SIZE", cursor_size_fmt, 1);
+		if (cursor_theme != NULL) {
+			setenv("XCURSOR_THEME", cursor_theme, 1);
 		}
+
+#if HAVE_XWAYLAND
+		if (!server.xwayland.xcursor_manager ||
+			!xcursor_manager_is_named(server.xwayland.xcursor_manager,
```
RedSoxFan: I think this should also check if the size changed

sway/input/seat.c
```diff
@@ -722,17 +723,66 @@ void seat_remove_device(struct sway_seat *seat,
...
+					image->hotspot_y);
+			}
+		}
+#endif
+	}
+
+	/* Create xcursor manager if we don't have one already, or if the
+	 * theme has changed */
+	if (!seat->cursor->xcursor_manager ||
+		!xcursor_manager_is_named(seat->cursor->xcursor_manager, cursor_theme)) {
```
RedSoxFan: Like above, I think this should also check if the size changed

sway/input/seat.c
dnkl: When testing, I noticed that the first time we came here, the default seat had no configuration. However, it was soon called again, this time with a `seat0` configuration available (with the `*` configuration applied/merged).

Out of curiosity (and so that I can learn); is that something that is specific to the default seat? Or what is the reason for explicitly asking for the `*` config?

sway/input/seat.c
dnkl: Absolutely yes

sway/input/seat.c
dnkl: Anyway, I've updated it to also try the `*` config. Resolving conversation for now.

RedSoxFan: After thinking about it, I think there is only one case where a seat would not have an explicit seat config and can fallback to the wildcard seat config.

If you had the following in your config (and no `seat seat0 <seat-subcommand>` lines):
```
seat seat1 fallback true
seat * xcursor_theme redglass 16
```
Then, there would be no seat config created for `seat0`. `seat1` has an explicit `fallback true` so there would not be an implicit one added for `seat0`. `seat0` would just exist as the default seat, but have no attachments and no config.

Any seat other than `seat0`, would require some `seat <name> <seat-subcommand>` to be executed, which would create a seat config before the seat itself.

sway/input/seat.c
```diff
@@ -722,17 +723,72 @@ void seat_remove_device(struct sway_seat *seat,
...
+		char cursor_size_fmt[16];
+		snprintf(cursor_size_fmt, sizeof(cursor_size_fmt), "%d", cursor_size);
+		setenv("XCURSOR_SIZE", cursor_size_fmt, 1);
+		if (cursor_theme != NULL) {
+			setenv("XCURSOR_THEME", cursor_theme, 1);
 		}
+
+#if HAVE_XWAYLAND
+		if (!server.xwayland.xcursor_manager ||
+			!xcursor_manager_is_named(server.xwayland.xcursor_manager,
```
RedSoxFan: style: there needs to be an additional level of indentation added to the wrapped lines of the condition to differentiate it visibly from the body

sway/input/seat.c
```diff
@@ -722,17 +723,72 @@ void seat_remove_device(struct sway_seat *seat,
...
+					image->hotspot_x, image->hotspot_y);
+			}
+		}
+#endif
+	}
+
+	/* Create xcursor manager if we don't have one already, or if the
+	 * theme has changed */
+	if (!seat->cursor->xcursor_manager ||
+		!xcursor_manager_is_named(
```
RedSoxFan: style: there needs to be an additional level of indentation added to the wrapped lines of the condition to differentiate it visibly from the body

sway/input/seat.c
dnkl: Ah, now I see what you mean. Hopefully fixed correctly this time.

ddevault:
Thanks!

Issue #4101, opened 2019-04-28
Remove orbital gamma-control and screenshooter


emersion:
We have improved versions under the wlr namespace.

c-edw:
Was this meant for wlroots? I see almost no references to these things in this repo.

emersion:
Nah, this is for sway. I'd like to stop exposing those after 1.1 is released.

PR #4204, opened 2019-06-02
Remove orbital gamma-control protocol

emersion:
~~Waiting for 1.1 to be released before merging this.~~

Fixes https://github.com/swaywm/sway/issues/4101


emersion:
Well I guess 1.1 has a branch, so we can merge this anyway. (Otherwise builds are broken now that wlroots has removed these protocols)

PR #4213, opened 2019-06-04
cmd_hide_edge_borders: add missing arg count check

RedSoxFan:
This adds the missing argument count check after the --i3 flag
processing in cmd_hide_edge_borders. Without the check,
`hide_edge_borders --i3` would result in a SIGSEGV instead of a syntax
error. There are some minor adjustments to make it so nothing gets
altered if this check fails


emersion:
Thanks!

PR #4218, opened 2019-06-05
criteria: reset on semicolon separation

RedSoxFan:
This matches i3's behavior of only retaining criteria across comma
separated commands. When separating commands with a semicolon, the
criteria is reset and allows for new criteria to be set, if desired.

Excerpt from [i3 user guide](https://i3wm.org/docs/userguide.html#command_criteria):
> When using multiple commands, separate them by using a , (a comma) instead of a semicolon. Criteria apply only until the next semicolon, so if you use a semicolon to separate commands, only the first one will be executed for the matched window(s).

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -227,28 +227,31 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 
 	head = exec;
 	do {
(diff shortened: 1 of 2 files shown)


ddevault:
Thanks!

PR #4220, opened 2019-06-05
swaybar/nag: use xcursor theme defined by XCURSOR_THEME/SIZE

dnkl:
If the `XCURSOR_THEME` and/or `XCURSOR_SIZE` environment variables are
set, use the theme and size they define.

If they're not set, use the same defaults as before (system default
theme, size=24).

Note: this builds upon https://github.com/swaywm/sway/pull/4202, and was suggested here: https://github.com/swaywm/sway/pull/4202#issuecomment-497972719

### DIFF
--- a/swaybar/input.c
+++ b/swaybar/input.c
@@ -67,9 +67,19 @@ void update_cursor(struct swaybar_seat *seat) {
 	if (pointer->cursor_theme) {
 		wl_cursor_theme_destroy(pointer->cursor_theme);
 	}
+	const char *cursor_theme = getenv("XCURSOR_THEME");
+	unsigned cursor_size = 24;
+	const char *env_cursor_size = getenv("XCURSOR_SIZE");
+	if (env_cursor_size) {
+		char *end;
+		unsigned size = strtoul(env_cursor_size, &end, 10);
+		if (!*end) {
+			cursor_size = size;
+		}
+	}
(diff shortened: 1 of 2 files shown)


swaybar/input.c
```diff
@@ -67,9 +67,19 @@ void update_cursor(struct swaybar_seat *seat) {
 	if (pointer->cursor_theme) {
 		wl_cursor_theme_destroy(pointer->cursor_theme);
 	}
+	const char *cursor_theme = getenv("XCURSOR_THEME");
+	unsigned cursor_size = 24;
+	const char *env_cursor_size = getenv("XCURSOR_SIZE");
+	if (env_cursor_size) {
+		char *end;
+		unsigned size = strtoul(env_cursor_size, &end, 10);
+		if (!*end) {
```
emersion: Also check errno, and also check the string isn't empty.

dnkl: Fixed. I did **not** add a check for the empty string for `XCURSOR_THEME` as I consider `""` to be just like misspelling a valid theme name. If you think this is incorrect I'll of course add a check there too,

dnkl:
> Thanks for your work, and sorry for being too slow to reply on the previous PR!

No worries. What better way to learn my way around the code than being forced to find the answers myself :)

RedSoxFan:
Thanks!

PR #4221, opened 2019-06-05
cmd_exec{,_always}: close standard in for child

### LINKED ISSUE #4214: `swaymsg exec nvim` corrupts Linux VT
If I run

    $ swaymsg exec nvim

in a terminal emulator inside sway, then exit sway, I get some really weird results. It is as if neovim has started up inside the linux terminal (e.g.: /dev/tty1) while sway is still running, and then when I exit sway I can simultaneously see neovim UI as well as my shell prompt, and I can't reliably exit either of them because they're somehow both in control of the tty.

Expected behaviour would be that neovim doesn't start, similar to when I run `nvim&` in a terminal. I would imagine this has something to do with neovim thinking it is running in 'interactive mode' inside /dev/tty1, when it really shouldn't be because /dev/tty1 is currently hidden and sway is running on top and capturing all user input.

I should mention - I know I need to start an editor like neovim inside a terminal emulator e.g.: `swaymsg exec alacritty -e nvim` if I want to actually be able to use it. However it still feels like a bug that running the 'wrong sort of executable' with `swaymsg exec` (i.e.: not a GUI app) produces such a corrupted state in my linux VT. The same thing happens with vi, and I assume vim (though I haven't tried vim because I don't have it installed)

### sway version

`sway --version` gives 

    sway version 1.0

however pacman reports `1.1.1-1`. I'm pretty sure I'm running the latest version since I needed to install `swaybg` before my background showed again.

RedSoxFan:
Fixes #4214 

This just closes the standard input for processes started with the exec
and exec_always commands. This prevents interactive programs like vi(m)
from taking over input on the getty. Standard output and standard error
are kept open, which allows for the output/errors to be captured along
with the sway log.

### DIFF
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -61,6 +61,7 @@ struct cmd_results *cmd_exec_always(int argc, char **argv) {
 		close(fd[0]);
 		if ((child = fork()) == 0) {
 			close(fd[1]);
+			close(STDIN_FILENO);
 			execl("/bin/sh", "/bin/sh", "-c", cmd, (void *)NULL);
 			_exit(0);
 		}


RedSoxFan:
I'm actually just going to close this. I cannot get anything to work unless stdout is redirected

dnkl:
Did you try redirecting stdin to a non-readable file descriptor (e.g. `/dev/null` opened in write-only mode)?

I imagine it will not work for every application (if at all) since it depends on how much (and what) they write to stdout before (hopefully) failing on stdin.

Thought it might be worth testing out anyway.

FreeFull:
In addition, making `isatty` return false for stdout and stderr might be nice, if possible. `isatty` is the common way most interactive programs check if they are running in a terminal

PR #4225, opened 2019-06-08
Unhide cursor on cursor activity after touch

ForTheReallys:
Fix a small bug where the cursor becomes hidden after receiving a touch event until the surface is unfocused or sway receives a `set_cursor` event.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -174,11 +174,15 @@ void cursor_rebase_all(void) {
 	}
 }
 
-static int hide_notify(void *data) {
-	struct sway_cursor *cursor = data;
+static void cursor_hide(struct sway_cursor *cursor) {
 	wlr_cursor_set_image(cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


sway/input/cursor.c
```diff
@@ -174,11 +174,15 @@ void cursor_rebase_all(void) {
 	}
 }
 
-static int hide_notify(void *data) {
-	struct sway_cursor *cursor = data;
+void cursor_hide(struct sway_cursor *cursor) {
```
RedSoxFan: nit: can you make this static since it isn't listed in the header (and shouldn't unless it is ever used in a different file)?

ForTheReallys: Done

RedSoxFan:
Can you squash the commits and force push?

ForTheReallys:
I actually didn't have the hardware either.  Airbar has a relatively cheap solution that provides touch capabilities via usb.  In case you're interested: https://air.bar .  It only supports certain screen sizes though

RedSoxFan:
Thanks!

PR #4228, opened 2019-06-08
Fix "xwayland false"

aidanharris:
Commit 190546fd315a24c04006fb1b177069933f4350da failed to consider the
edge case where xwayland is disabled via the sway config. This leads to
a SEGFAULT when setting the xwayland cursor since the xwayland server is
not running.


emersion:
Thanks!

PR #4229, opened 2019-06-08
swaybar: do not retry search for tray icons

### LINKED ISSUE #3789: swaybar repeatedly polling various files
sway version 1.0-rc1-123-gd016848b

Specific tray apps seem to make swaybar repeatedly poll for image/asset files.

kdeconnect-indicator doesn't do it, but [kmix](https://cgit.kde.org/kmix.git), [radiotray-ng](https://github.com/ebruck/radiotray-ng) (gtk) and soulseekqt do. the icon for kmix is a red face, but the radiotray-ng and soulseekqt icons appear as they should.

swaybar CPU usage without any tray apps is about 0.5%. For each troublesome app open, it goes up roughly by 3.5/5%.

[sway.log](https://gist.github.com/mxmilkb/fb943841e1d0ab650ed812537f202286)

[swaystrace.log](https://gist.github.com/mxmilkb/c93e3d801a5d2dbf72f23495d83598d6)

kupospelov:
In case a tray icon cannot be found or does not have a desirable size, swaybar retries the search again and again, which increases load on disk and CPU.

This commit solves it by storing `target_size` for each icon, so that swaybar does not search for an icon of some size if it already tried to.

Fixes #3789.

### DIFF
--- a/include/swaybar/tray/item.h
+++ b/include/swaybar/tray/item.h
@@ -20,6 +20,7 @@ struct swaybar_sni {
 	cairo_surface_t *icon;
 	int min_size;
 	int max_size;
+	int target_size;
 
 	// dbus properties
 	char *watcher_id;
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -1,6 +1,7 @@
 #define _POSIX_C_SOURCE 200809L
 #include <arpa/inet.h>
 #include <cairo.h>
+#include <limits.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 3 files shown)


ianyfan:
It looks like this would prevent users from replacing the icon file while swaybar is running.

kupospelov:
Yep. Wouldn't it be reasonable to require a bar restart?

The main use case here is changing `icon_theme`, but the user will have to restart the bar anyway after a config change.

I have read about the proposed alternatives, but as a user who never changes the icon theme, I do not like the idea of checking `mtime` of a bunch of directories every few seconds.

travankor:
Any chance this will be merged for sway 1.3?

ddevault:
The tradeoffs seem reasonable. This needs to be rebased before it can be merged, though.

kupospelov:
Rebased. @ddevault please take a look when you have a chance.

emersion:
Thanks!

PR #4237, opened 2019-06-09
ipc: add xkb_layout_names and xkb_active_layout_index

emersion:
These can be useful to implement per-window layouts as a script.

See https://github.com/swaywm/sway/issues/2361


ddevault:
Thanks!

PR #4238, opened 2019-06-09
Add a new xkb_switch_layout command

emersion:
This allows users to programatically change the active layout.


ddevault:
Thanks!

Issue #883, opened 2016-09-06
Support setting (dynamic) title bar location


johnp:
"Use case": 16:10 screens are unreasonably high price and nearly impossible to get in a laptop.
(Also, Firefox tabs-in-titlebar doesn't work in sway/i3, which means you can't just flick the mouse pointer to the screen edge to switch tabs)

It would therefore be nice to have a `title_bar` command in the sway config with the basic options:
- top/bottom/left/right/none

Advanced options (not sure if these are worth it, but would make some nice /r/unixporn posts):
- outer_edge_horizontal/outer_edge_vertical

Advanced options need to have a fallback, specifying which option to choose, should no outer edge or two exist in that plane.

Examples:

`title_bar outer_edge_vertical bottom`

E.g. first doing horizontal splits always places the title bar at the bottom, but upon doing a vertical split, the title bar of the upper window will be placed at the top (outer vertical edge).

`title_bar outer_edge_vertical outer_edge_horizontal bottom`

Same as above, but if a window doesn't have vertical outer edges (or two of them) it tries to find a horizontal outer edge first, before falling back to bottom.  
E.g. one horizontal split and then split the right container three times vertically. The top right will have it's title bar on the top, the bottom right on the bottom and the window in between those two will have it's title bar at the right edge (outer horizontal).

SirCmpwn:
I don't understand what you're asking for.

johnp:
Currently, setting `border pixel <n>` disables the title bar, `border normal` has it's title bar at the top.
`title_bar bottom` would now instead place the title bar at the bottom. That's pretty much the main thing I'd be asking for.

SirCmpwn:
Interesting thought. May be worth looking into. Have you pitched this feature request to i3 as well?

johnp:
No, I've just done a quick search and could only find this: https://www.reddit.com/r/i3wm/comments/463mtj/window_titlebars_on_bottom/
Should I place a feature request there too?

SirCmpwn:
Wouldn't be a bad idea.

johnp:
Opened i3/i3/issues/2441. The explanation of the dynamic placement option may be a bit better there.

alkino:
This feature has been refused in i3.

SirCmpwn:
Link to i3 discussion?

bearcatsandor:
[...]

PR #4244, opened 2019-06-10
Allow rendering titlebar on bottom

janza:
This implements part of #883 and similar issue on i3/i3/issues/2747 by adding a command `titlebar_position`. It accepts two options for now: top and bottom.

I've probably missed some edge cases but would like to know if there's any interest in adding this feature?

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -473,6 +473,11 @@ enum xwayland_mode {
 	XWAYLAND_MODE_IMMEDIATE,
 };
 
+enum titlebar_position {
+	TITLEBAR_TOP,
+	TITLEBAR_BOTTOM
+};
+
 /**
  * The configuration struct. The result of loading a config file.
  */
@@ -534,6 +539,7 @@ struct sway_config {
 	bool show_marks;
 	enum alignment title_align;
 	bool primary_selection;
+	enum titlebar_position titlebar_position;
 
 	bool tiling_drag;
 	int tiling_drag_threshold;
--- a/sway/commands/titlebar_position.c
+++ b/sway/commands/titlebar_position.c
@@ -1,0 +1,30 @@
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/output.h"
+#include "sway/tree/arrange.h"
+#include "sway/tree/root.h"
+
+struct cmd_results *cmd_titlebar_position(int argc, char **argv) {
+	struct cmd_results *error = NULL;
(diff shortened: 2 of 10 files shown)


include/sway/config.h
```diff
@@ -413,6 +413,11 @@ enum alignment {
 	ALIGN_RIGHT
 };
 
+enum position {
```
ddevault: Would prefer this to be titlebar-specific, i.e. `TITLEBAR_TOP`

ddevault:
+1

sway/commands/titlebar_position.c
```diff
@@ -0,0 +1,29 @@
...
+
+	if (strcmp(argv[0], "top") == 0) {
+		config->titlebar_position = TITLEBAR_TOP;
+	} else if (strcmp(argv[0], "bottom") == 0) {
+		config->titlebar_position = TITLEBAR_BOTTOM;
+	} else {
+		return cmd_results_new(CMD_INVALID,
+				"Expected 'titlebar_position top|bottom'");
+	}
+
```
RedSoxFan: You are going to want to add a call to `arrange_root` here

RedSoxFan (commented):
I know this is listed as a draft, but here's a few small comments.

janza:
I might need a bit of guidance on this one:

I've implemented rendering of titlebar on the bottom which works OK in cases when border is set to 0. 
But with the current implementation I see some [rendering issues](https://josip.dev/vid/recording-20191016193752.mp4) when there is a border. They appear when resizing the window using mouse, but they are not there when resizing using keyboard shortcuts (or swaymsg).

ammgws:
@janza do you still get those rendering issues?

i2:
Hi, what is the status of this PR? Thank you!

janza:
Hi, I've been running sway with this patch for a while now without issues (AUR sway-titlebar-positioning-git).
I would be interested to hear if there's additional feedback on the code and interest in merging it?

i2:
I'd be very interested if this becomes available on main branch.

gman0:
Hello, any updates on this?

herkhinah:
I would love getting this merges although it's rejected by i3. In my setup (my focus is more on the top of the display because that way I can sit more ergonomically) I want the most important information at the top of my screen and the noise like system bar and window titles at the bottom.

Dazage:
I'd love to see this merged as it makes for a cleaner interface whenever the sway bar is positioned at the top of the screen.
Rendering the titlebar for applications at the bottom reduces clutter/"noise" between important information in the sway bar and the application(s) on the screen.

PR #4246, opened 2019-06-11
Fix segfaults caused by faulty command parsing

### LINKED ISSUE #4239: crash on swaymsg command
* **Sway Version**: `sway version 1.1-rc1-34-g5b1a8d62 (Jun  9 2019, branch 'master')`

* **Debug Log**: of `./build/sway/sway -d -c /etc/sway/config &>/tmp/sway.log`: http://sprunge.us/KcmhqR

* **Default Config File** (untouched): `/etc/sway/config` http://sprunge.us/jsajhn
* **CUSTOM Config File**: `/etc/sway/config.d/99-custom.mine`: http://sprunge.us/YxWBgg

* **Steps to reproduce**:
 1.) start sway as user `me` with standard config + above 99-custom.mine config: with `./build/sway/sway -d -c /etc/sway/config`
 2.) start sakura with app_id `my-shell` using custom shortcut `$mod+shift+return`
 3.) switch to empty/fresh workspace with default shortcut `$mod+2`
 4.) try to get sakura shell over using custom shortcut `$mod+shift+s`

* **Stack Trace**, if sway crashes:
    `coredumpctl gdb sway` with `bt full`: http://sprunge.us/HIv9ZV

mcoffin:
This fixes faulty command parsing introduced by
f0f5de9a9e87ca1f0d74e7cbf82ffceba51ffbe6. When that commit allowed
criteria reset on ';' delimeters in commands lists, it failed to account
for its inner ','-parsing loop eating threw the entire rest of the
string.

This refactors argsep to use a list of multiple separators, and
(optionally) return the separator that it matched against in this
iteration via a pointer. This allows it to hint at the command parser
which separator was used at the end of the last command, allowing it to
trigger a potential secondary read of the criteria.

Fixes #4239

### DIFF
--- a/common/stringop.c
+++ b/common/stringop.c
@@ -251,37 +251,61 @@ char *join_args(char **argv, int argc) {
 	return res;
 }
 
-char *argsep(char **stringp, const char *delim) {
+static inline char *argsep_next_interesting(const char *src, const char *delim) {
+	char *special = strpbrk(src, "\"'\\");
+	char *next_delim = strpbrk(src, delim);
+	if (!special) {
+		return next_delim;
+	}
+	if (!next_delim) {
+		return special;
+	}
+	return (next_delim < special) ? next_delim : special;
+}
+
+char *argsep(char **stringp, const char *delim, char *matched) {
 	char *start = *stringp;
 	char *end = start;
 	bool in_string = false;
 	bool in_char = false;
 	bool escaped = false;
-	while (1) {
-		if (*end == '"' && !in_char && !escaped) {
-			in_string = !in_string;
-		} else if (*end == '\'' && !in_string && !escaped) {
-			in_char = !in_char;
-		} else if (*end == '\\') {
-			escaped = !escaped;
-		} else if (*end == '\0') {
-			*stringp = NULL;
-			break;
-		} else if (!in_string && !in_char && !escaped && strchr(delim, *end)) {
-			if (end - start) {
-				*(end++) = 0;
-				*stringp = end + strspn(end, delim);;
-				if (!**stringp) *stringp = NULL;
-				break;
-			} else {
-				++start;
-				end = start;
-			}
-		}
-		if (*end != '\\') {
+	char *interesting = NULL;
+
+	while ((interesting = argsep_next_interesting(end, delim))) {
+		if (escaped && interesting != end) {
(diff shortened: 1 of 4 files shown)


common/stringop.c
```diff
@@ -251,37 +251,60 @@ char *join_args(char **argv, int argc) {
...
-		} else if (*end == '\0') {
-			*stringp = NULL;
-			break;
-		} else if (!in_string && !in_char && !escaped && strchr(delim, *end)) {
+			end = interesting + 1;
+		} else if (*interesting == '\\') {
+			if (escaped) {
+				escaped = NULL;
+			} else {
+				escaped = end;
```
RedSoxFan: I'm pretty sure that this should be `escaped = interesting`.  That would also mean that `escaped` could remain a `bool` and the check above could be changed to `if (escaped && interesting != end) {`

common/stringop.c
```diff
@@ -251,37 +251,60 @@ char *join_args(char **argv, int argc) {
...
+		} else {
+			end++;
 		}
-		if (*end != '\\') {
-			escaped = false;
-		}
-		++end;
+	}
+	if (!interesting) {
+		*stringp = NULL;
```
RedSoxFan: `*matched` should also be set to `\0` since the end was reached instead of a delim. Otherwise, the previous value set will still be there (unless reset by the caller).

common/stringop.c
mcoffin: Yep. Nice catch. I've been blessed so much with using Rust to write parsers that I'd forgotten how much of a PITA it is in raw C before I built this. :rofl:

common/stringop.c
```diff
@@ -251,37 +251,61 @@ char *join_args(char **argv, int argc) {
...
 	bool in_string = false;
 	bool in_char = false;
 	bool escaped = false;
-	while (1) {
-		if (*end == '"' && !in_char && !escaped) {
+	char *interesting = NULL;
+
+	while ((interesting = argsep_next_interesting(end, delim))) {
+		if (escaped && interesting != end) {
+			escaped = NULL;
```
RedSoxFan: `escaped = false`

mcoffin:
@RedSoxFan done.

RedSoxFan:
Thanks!

PR #4249, opened 2019-06-11
bindings: defer while initializing

### LINKED ISSUE #4224: Starting sway with the laptop lid closed does not disable the output
Version: `sway version 1.1-rc1-25-g799f5a2c (Jun  5 2019, branch 'master')`

I have my laptop connected to an external monitor, so my laptop lid is closed most of the time.
After booting the laptop and starting sway, i have to manually open the lid then close it in order for it to detect that my laptop lid is closed.

I have the following bindswitch commands configured:
```
bindswitch lid:on exec swaymsg output eDP-1 disable
bindswitch lid:off exec swaymsg output eDP-1 enable
```

Im assuming this is not actually a bug, and should be an enhancement request.

RedSoxFan:
Fixes #4224 

This adds the logic to defer binding execution while sway is still
initializing. Without this, the binding command would be executed, but
the command handler would return CMD_DEFER, which was being treated as
a failure to run. To avoid partial executions, this will defer all
bindings while config->active is false.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -644,7 +644,23 @@ void run_deferred_commands(void) {
 		list_free(res_list);
 		free(line);
 	}
-	transaction_commit_dirty();
+}
+
+void run_deferred_bindings(void) {
+	struct sway_seat *seat;
+	wl_list_for_each(seat, &(server.input->seats), link) {
+		if (!seat->deferred_bindings->length) {
+			continue;
+		}
+		sway_log(SWAY_DEBUG, "Running deferred bindings for seat %s",
+				seat->wlr_seat->name);
+		while (seat->deferred_bindings->length) {
+			struct sway_binding *binding = seat->deferred_bindings->items[0];
+			seat_execute_command(seat, binding);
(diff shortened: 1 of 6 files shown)


sway/config.c
```diff
@@ -647,6 +647,24 @@ void run_deferred_commands(void) {
...
+		sway_log(SWAY_DEBUG, "Running deferred bindings for seat %s",
+				seat->wlr_seat->name);
+		while (seat->deferred_bindings->length) {
+			struct sway_binding *binding = seat->deferred_bindings->items[0];
+			seat_execute_command(seat, binding);
+			list_del(seat->deferred_bindings, 0);
+			free_sway_binding(binding);
+		}
+	}
+	transaction_commit_dirty();
```
emersion: Maybe we can move this out of this function, and add it after this function's call, so that we atomically apply the new state for both deferred commands and bindings?

RedSoxFan: Done

emersion:
Thanks!

PR #4252, opened 2019-06-13
use surface coordinates for damaging buffers

ii8:
Using the content coordinates can go wrong if the surface inside the content is actually smaller than the content space.

You can see the effect if you run [havoc](https://github.com/ii8/havoc) and enter text on the very bottom line; if the content width/height is not an exact multiple of the cell width/height then the text will be cut off.

Somebody more familiar with the sway codebase can hopefully confirm that using pending coordinates instead of current ones has no bad side-effects.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -163,12 +163,12 @@ void output_view_for_each_surface(struct sway_output *output,
 		.user_iterator = iterator,
 		.user_data = user_data,
 		.output = output,
-		.ox = view->container->current.content_x - output->lx
+		.ox = view->container->surface_x - output->lx
 			- view->geometry.x,
-		.oy = view->container->current.content_y - output->ly
+		.oy = view->container->surface_y - output->ly
 			- view->geometry.y,
-		.width = view->container->current.content_width,
-		.height = view->container->current.content_height,
+		.width = view->container->surface_width,
+		.height = view->container->surface_height,
 		.rotation = 0, // TODO
 	};
 
@@ -182,12 +182,12 @@ void output_view_for_each_popup(struct sway_output *output,
 		.user_iterator = iterator,
 		.user_data = user_data,
 		.output = output,
-		.ox = view->container->current.content_x - output->lx
+		.ox = view->container->surface_x - output->lx
 			- view->geometry.x,
-		.oy = view->container->current.content_y - output->ly
+		.oy = view->container->surface_y - output->ly
 			- view->geometry.y,
-		.width = view->container->current.content_width,
-		.height = view->container->current.content_height,
+		.width = view->container->surface_width,
+		.height = view->container->surface_height,
 		.rotation = 0, // TODO
 	};
 


sway/desktop/output.c
```diff
@@ -163,12 +163,12 @@ void output_view_for_each_surface(struct sway_output *output,
 		.user_iterator = iterator,
 		.user_data = user_data,
 		.output = output,
-		.ox = view->container->current.content_x - output->lx
+		.ox = view->container->surface_x - output->lx
```
emersion: Hmm. I think we need to add these to `sway_container_state`

ii8: Ah, I feared this.

So why does a `sway_container` not simply have two `sway_container_state` members, one for pending as well?

emersion: It's a consequence of the codebase not being designed for atomic layout updates from day one.

Though adding those to `sway_container_state` shouldn't be too hard, just need to update them from the pending state with the other fields.

ii8: Maybe I'm not understanding this right, but it looks like those existing surface variables are already being treated as if they belong in the current state(which is why this commit seems to work correctly)

They are simply calculated from the content variables when a transaction is applied.

Should I move them into the current state?

ii8: Except for some reason they are not the current content variables. I guess I'm missing something, I'll look at this again tomorrow.

emersion: >They are simply calculated from the content variables when a transaction is applied.

Hmm, that sounds confusing. At the very least we could move them to the current state struct. @RyanDwyer, do you remember what's the deal?

RyanDwyer: When I implemented transactions my approach was to only add properties to the state struct if they were required. And everything seemed to work without putting the surface coords in the struct.

It appears that the surface coords are only used for the `container_at_coords` functions and when rendering the surface. Because we update it when applying the transaction, the value reflects the rendered/current state rather than pending state. This works fine for rendering, but introduces an unnoticeable bug for `container_at_coords`. If you click a surface while a transaction is in progress which moves the surface's position then it will use the rendered coords but should be using the pending coords, which we don't have.

If it helps, you can think of the normal/pending properties being the real live state, but the current/rendered state might lag behind. So normal container clicks apply to the normal/pending state which is potentially not being rendered yet.

The "correct" approach is to add the surface coords to the container state then use `view->container->current.surface_x` here. The `surface_at_view` function can continue to use `container->surface_x` because it would then reflect the pending state which fixes the minor bug mentioned above. The rendering code would need to be updated to use `current.surface_x`.

However, the current proposed code should work without introducing any problems. I say accept this as it is, and if you want to fix the `container_at_coords` bug then it should be a separate issue/PR.

emersion: Follow-up issue: https://github.com/swaywm/sway/issues/4277

emersion:
Thanks!

Issue #3380, opened 2019-01-06
Compositor Effects


Olie440:
This issue was raised two years ago before Sway came with a compositor: https://github.com/swaywm/sway/issues/453.

Are there any plans to implement any compositor effects similar to what Compton has for X?
The main ones i'd like to see are: 

- Fading between work-spaces
- Shadows
- Opacity

emersion:
>before Sway came with a compositor

Sway has always been a Wayland compositor. Things work differently on X11 and on Wayland.

Sway has the roles of a X11 server + X11 WM + X11 compositor.

>Are there any plans to implement any compositor effects similar to what Compton has for X?

No, sorry, this isn't planned.

However, you can already set a window's opacity with the `opacity` command. See `contrib/inactive-windows-transparency.py` for example usage.

If you want effects, other compositors might suit you better (e.g. https://github.com/WayfireWM/wayfire).

Olie440:
Does Sway work with other Compositors? (ie. Serving just the roles of X11 server + X11 WM).
I like Sways layout and ability to have config be ported from i3, but I also want to be able to have a fading effect and shadows.
I imagine most users coming from i3 would feel the same.

emersion:
>Does Sway work with other Compositors? (ie. Serving just the roles of X11 server + X11 WM).

No, you can't use sway and another compositor at the same time.

Olie440:
Ok thank you for the information.

WillPower3309:
[swayfx](https://www.github.com/willpower3309/swayfx) currently has rounded corners and shadows, as well as dimming. Although, blur is currently in a draft PR, so it will be merged soon. Fade animations will be next!

saxophone-guy:
Just a suggestion, but Hyprland seems promising

emersion:
Note that Hyprland's main developer, vaxry, has been banned from the FreeDesktop.Org community.

Issue #1511, opened 2017-12-12
Shadow support?


that1communist:
My usecase is a black border when unfocused and a white border when focused, meaning when I have a floating window on a white background, i can't see the border, compton shadows fix this, and it'd be really nice to have this on sway, and I mostly post this to find out if this is planned.

SirCmpwn:
You already requested this -_-

https://github.com/swaywm/sway/issues/1426

that1communist:
Yeah but I figured I was too broad. Not supporting most compton features =/= not supporting shadows.

Issue #1065, opened 2017-01-29
Feature request: drop shadows


bearcatsandor:
How difficult would it be to add the ability to have drop-shadows on windows, for those of us that use gaps?

SirCmpwn:
Probably not that difficult.

bearcatsandor:
Anychance this might make it into the roadmap, is it an addition that you'd accept?

SirCmpwn:
I won't put it on the roadmap but I would accept a pull request.

SirCmpwn:
No.

topisani:
I implemented this in [tablecloth](https://github.com/topisani/tablecloth) (rootston based compositor i abandoned), and would like to implement it in sway. The aproach i used used GLESv2 directly (loading a GL shader). Would an aproach like this be accepted? maybe enabled by a macro if GLES is avaliable?
Otherwise, what would be the rendering options?
 
Config-wise, i would probably do it like border/titlebar configuration, allowing shadow color change on focus, urgent etc.

rhysperry111:
I think this probably depends on #3652

PR #4254, opened 2019-06-15
Configurable GLESv2 drop shadows

topisani:
drop shadows

I implemented drop shadows using GLESv2. There is a meson build option to disable them, which leaves the configs and commands in place, but disables the rendering (and thus the GLESv2 requirement).

The shadows are configurable using
```
shadows.<focused|focused_inactive|unfocused|urgent> <radius> <offset> <color>
```
Radius is the radius (duh), offset is in the bottom right direction.

It's something that's been requested a few times (#1065, #3380, #1511), and personally i think it really helps the "sexyness" (even though the above screenshot is way over the top).

Currently the main issue is that damage is not being applied when you move a window (especially apparent when dragging floating windows). I assume this should be fairly easy for someone with knowledge of the codebase to fix.


ascent12:
You should seriously consider telling people of your intentions to do such a PR before you spend a lot of potentially wasted time and effort on it.

For starters, I'm strongly against sway making any GL calls directly; it seriously messes with some stuff going on in wlroots (i.e. the vulkan renderer). Yes, wlr_renderer is stupidly limited, and compositors are free to do whatever the hell they want, but sway is the flagship. If something in wlroots is broken/crap, we fix wlroots instead of making sway work around it.

But also, I think this is fundamentally the wrong way to go about implementing a feature like this. Drop shadows should be implemented as window decorations, rather than a big rectangle behind everything. You can get all of the effects of drop shadows, along with a whole lot more flexibility to do other shit, as well as preventing the situation this may cause with double drop-shadows, where a CSD client draws its own (e.g. GTK).

---

I'm getting off-topic of the discussion of the PR itself here, but this is how I think shadows should be implemented:

If sway had a customizable system for window decorations (which some people have been asking for), you would just clamp something like 
right-border
to the edges of windows, you can get whatever effect you want, including drop shadows. Technically it only needs to be 1 pixel tall and tile the image all the way along, but I've made it longer just so you can actually see it.
You would need 8 images total, for all of the corners and edges, so you can also have offset shadows. But sway doesn't need to know anything about that, it's just clamping 8 images to the edges/corners.
You would also need some extra information to say what should be considered part of the window geometry and what isn't.

I've been thinking of pitching something like this for inclusion in wlroots as a server-side decorator helper, or possibly as its own library with some kind of "standard" theme system, but it's not something I've actually worked on, and is a low priority on my list of things I want to do. I've also considered a wayland extension which allows the compositor to tell clients what SSDs it should draw, which kind of turns it into a CSD/SSD hybrid, where you get the advantages of both (unified look, but works more nicely with things like direct scan-out).

ddevault:
Sorry, but I'll pass on this patch.

topisani:
Makes sense, wasn't really expecting it to get merged, mainly implemented it for personal use. @ascent12 the approach you're describing sounds like it makes more sense, hope we get something like this in the future!

escalade:
It works and looks great. Hopefully this functionality will be properly added soon.

WillPower3309:
[swayfx](https://www.github.com/willpower3309/swayfx) currently has rounded corners and shadows, as well as dimming. Although, blur is currently in a draft PR, so it will be merged soon. Fade animations will be next!

PR #4262, opened 2019-06-17
Add calibration_matrix config option

mebubo:
Can be used to change the orientation of a touchscreen.

Example usage with swaymsg:

    # identity
    swaymsg input type:touch calibration_matrix '"1 0 0 0 1 0"'

    # 90 degree clockwise
    swaymsg input type:touch calibration_matrix '"0 -1 1 1 0 0"'

    # 180 degree clockwise
    swaymsg input type:touch calibration_matrix '"-1 0 1 0 -1 1"'

    # 270 degree clockwise
    swaymsg input type:touch calibration_matrix '"0 1 0 -1 0 1"'

Documentation:

    https://wayland.freedesktop.org/libinput/doc/latest/absolute-axes.html#calibration-of-absolute-devices

I have never really written any C, so please don't hesitate to tell me if I did anything horrible C-wise.

Personally I'm unsure about a few things:

1) The struct I chose to represent the calibration matrix in `config.h`. I also tried using:
- A plain `float matrix[6]`, but there was no good way to represent the non-configured state.
- A `float *matrix`, with `NULL` for the non-configured state. Allocating the arrays felt messy and less safe in this case. The commit for this attempt is here: https://github.com/mebubo/sway/commit/ea7c76514d6b218d0aa53af80f07a691eb587fc0

2) I was not sure whether it was safe to pass the reference to a local array to `libinput_device_config_calibration_set_matrix`. I had to go and look at libinput sources to confirm that it does not store the reference. Checking for things like this - is it something one has to do systematically?

3) Should I have used a loop to print the array contents instead of referencing the elements one-by-one?

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -101,6 +101,11 @@ struct input_config_mapped_from_region {
 	bool mm;
 };
 
+struct calibration_matrix {
+	bool configured;
+	float matrix[6];
+};
+
 /**
  * options for input devices
  */
@@ -109,6 +114,7 @@ struct input_config {
 	const char *input_type;
 
 	int accel_profile;
+	struct calibration_matrix calibration_matrix;
 	int click_method;
 	int drag;
 	int drag_lock;
--- a/sway/commands/input/calibration_matrix.c
+++ b/sway/commands/input/calibration_matrix.c
@@ -27,7 +27,7 @@ struct cmd_results *input_cmd_calibration_matrix(int argc, char **argv) {
 	for (int i = 0; i < split->length; ++i) {
 		char *item = split->items[i];
 		float x = parse_float(item);
-		if (x != x) {
+		if (isnan(x)) {
 			return cmd_results_new(CMD_FAILURE, "calibration_matrix: unable to parse float: %s", item);
 		}
 		parsed[i] = x;
--- a/sway/commands/input/calibration_matrix.c
+++ b/sway/commands/input/calibration_matrix.c
@@ -10,7 +10,7 @@
 
 struct cmd_results *input_cmd_calibration_matrix(int argc, char **argv) {
 	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "calibration_matrix", EXPECTED_EQUAL_TO, 1))) {
+	if ((error = checkarg(argc, "calibration_matrix", EXPECTED_EQUAL_TO, 6))) {
 		return error;
 	}
 	struct input_config *ic = config->handler_context.input_config;
@@ -18,14 +18,9 @@ struct cmd_results *input_cmd_calibration_matrix(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "No input device defined.");
 	}
 
-	list_t *split = split_string(argv[0], " ");
-	if (split->length != 6) {
-		return cmd_results_new(CMD_FAILURE, "calibration_matrix should be a space-separated list of length 6");
-	}
-
 	float parsed[6];
-	for (int i = 0; i < split->length; ++i) {
-		char *item = split->items[i];
+	for (int i = 0; i < argc; ++i) {
+		char *item = argv[i];
 		float x = parse_float(item);
 		if (isnan(x)) {
 			return cmd_results_new(CMD_FAILURE, "calibration_matrix: unable to parse float: %s", item);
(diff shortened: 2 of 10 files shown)


emersion:
>I have never really written any C, so please don't hesitate to tell me if I did anything horrible C-wise.

No worries! Thanks for taking the time to dig in and write a patch!

>A plain float matrix[6], but there was no good way to represent the non-configured state.

This seems like a good option to me. `NAN` can be used to represent unset values.

>I was not sure whether it was safe to pass the reference to a local array to libinput_device_config_calibration_set_matrix. I had to go and look at libinput sources to confirm that it does not store the reference. Checking for things like this - is it something one has to do systematically?

Depends. Is the argument marked as `const`? If this is the case, this means the function will read the value when called and not access it afterwards. Otherwise when it's unclear, the docs usually explain that sort of thing.

>Should I have used a loop to print the array contents instead of referencing the elements one-by-one?

If it's just 6 elements, referencing one by one looks fine to me.

include/sway/config.h
```diff
@@ -101,6 +101,11 @@ struct input_config_mapped_from_region {
 	bool mm;
 };
 
+struct calibration_matrix {
+	bool configured;
```
emersion: Ah, I see. This looks OK to me as well.

sway/commands/input/calibration_matrix.c
```diff
@@ -0,0 +1,40 @@
...
+	list_t *split = split_string(argv[0], " ");
+	if (split->length != 6) {
+		return cmd_results_new(CMD_FAILURE, "calibration_matrix should be a space-separated list of length 6");
+	}
+
+	float parsed[6];
+	for (int i = 0; i < split->length; ++i) {
+		char *item = split->items[i];
+		float x = parse_float(item);
+		if (x != x) {
```
emersion: There is a `isnan` function

sway/commands/input/calibration_matrix.c
```diff
@@ -0,0 +1,40 @@
...
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "calibration_matrix", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE, "No input device defined.");
+	}
+
+	list_t *split = split_string(argv[0], " ");
```
emersion: Is there a reason why the strings need to be quoted like this:

```
calibration_matrix "1 2 3 4 5 6"
```

Instead of having something like this:

```
calibration_matrix 1 2 3 4 5 6
```

emersion:
Would be nice to port https://gitlab.freedesktop.org/wayland/weston/blob/master/clients/touch-calibrator.c to layer-shell

RedSoxFan:
Just as a note, either in this PR or a follow-up, the current calibration matrix should also be added to the IPC description for libinput devices in [`describe_libinput_device`](https://github.com/swaywm/sway/blob/ddad41f423db14f89f765fc134d45d1ea8a60caf/sway/ipc-json.c#L623) and documented in [`sway-ipc.7.scd`](https://github.com/swaywm/sway/blob/ddad41f423db14f89f765fc134d45d1ea8a60caf/sway/sway-ipc.7.scd#L1055)

sway/commands/input/calibration_matrix.c
mebubo: Indeed it is simpler this way. Implemented.

mebubo:
> Just as a note, either in this PR or a follow-up, the current calibration matrix should also be added to the IPC description for libinput devices in [`describe_libinput_device`](https://github.com/swaywm/sway/blob/ddad41f423db14f89f765fc134d45d1ea8a60caf/sway/ipc-json.c#L623) and documented in [`sway-ipc.7.scd`](https://github.com/swaywm/sway/blob/ddad41f423db14f89f765fc134d45d1ea8a60caf/sway/sway-ipc.7.scd#L1055)

Thanks for the reminder, implemented.

mebubo:
> > I was not sure whether it was safe to pass the reference to a local array to libinput_device_config_calibration_set_matrix. I had to go and look at libinput sources to confirm that it does not store the reference. Checking for things like this - is it something one has to do systematically?
> 
> Depends. Is the argument marked as `const`? If this is the case, this means the function will read the value when called and not access it afterwards. Otherwise when it's unclear, the docs usually explain that sort of thing.

Thanks! The argument is indeed marked as const.

mebubo:
@emersion, is there anything that needs to be done before this can be merged?

emersion:
Thanks!

ddevault:
Nice work!

PR #4267, opened 2019-06-19
cmd_seat: split action and config handlers

### LINKED ISSUE #4233: Performance issues with "seat seat0 cursor set x y"
* Sway Version: `sway version 1.1.1`

Freshly compiled from git tag "1.1.1", since weirdly sway-1.1.1 on ArchLinux gives 1.0 to swaymsg -t get_version and sway -v

* Debug Log: https://pastebin.com/f8hq6yCd

Note that I removed ~ 2 MB worth of repeating lines (replaced them by --SKIP--)

* Configuration File: Bug also happen in default configuration

I want to be able to simulate smooth mouse movement from a script.

Under i3/X, I use [xte](https://linux.die.net/man/1/xte). I thought I could just `swaymsg -t command seat seat0 cursor set x y` to achieve the same result with Sway.

Unfortunately, it doesn’t works. Sending ~60 such commands per second (for visual smoothness) makes Sway use 100% of CPU, with the cursor moving only ~1-2 times per second and the terminal instance running the script (tested both with alacritty and xterm) ending up getting killed.

Here is a Python script reproducing the issue : https://gist.github.com/sloonz/865f49654a7d64d0500debe0b6e75661

It basically tries to smoothly move the cursor on a circle of radius 300 px centered around (500, 500) at 1 turn/second. Running it results in a non-smooth movement, Sway using huge amount of CPU, and the terminal running the script getting killed.

RedSoxFan:
Fixes #4233 

This separates the logic for seat subcommand handlers that only perform
actions on the seat and handlers that alter the seat config. The former
group can immediately free the seat config after running the command as
it is only used by the subcommand to find the name of the seat to
operate on. The latter group alters the seat config so it will need to
go through the storage and application stage (assuming success).

_I chose the above solution rather than the proposed flag since this
constrains all of the logic to the seat command handler. Additionally,
a flag is useless for the remainder of the life of the seat config_

### DIFF
--- a/sway/commands/seat.c
+++ b/sway/commands/seat.c
@@ -7,14 +7,44 @@
 #include "stringop.h"
 
 // must be in order for the bsearch
+// these handlers perform actions on the seat
+static struct cmd_handler seat_action_handlers[] = {
+	{ "cursor", seat_cmd_cursor },
+};
+
+// must be in order for the bsearch
+// these handlers alter the seat config
 static struct cmd_handler seat_handlers[] = {
 	{ "attach", seat_cmd_attach },
-	{ "cursor", seat_cmd_cursor },
 	{ "fallback", seat_cmd_fallback },
 	{ "hide_cursor", seat_cmd_hide_cursor },
 	{ "pointer_constraint", seat_cmd_pointer_constraint },
 	{ "xcursor_theme", seat_cmd_xcursor_theme },
 };
+
+static struct cmd_results *action_handlers(int argc, char **argv) {
+	struct cmd_results *res = config_subcommand(argv, argc,
+			seat_action_handlers, sizeof(seat_action_handlers));
+	free_seat_config(config->handler_context.seat_config);
+	config->handler_context.seat_config = NULL;
+	return res;
+}
+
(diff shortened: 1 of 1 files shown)


emersion:
Note: we'd need a similar fix for `xkb_switch_layout`, right now it's hardcoded IIRC.

sway/commands/seat.c
```diff
@@ -7,15 +7,45 @@
...
+// these handlers perform actions on the seat
+static struct cmd_handler seat_action_handlers[] = {
+	{ "cursor", seat_cmd_cursor },
+};
+
+// must be in order for the bsearch
+// these handlers alter the seat config
 static struct cmd_handler seat_handlers[] = {
 	{ "attach", seat_cmd_attach },
-	{ "cursor", seat_cmd_cursor },
```
emersion: Adding a `bool` flag to `cmd_handler` would have been much nicer… But then it wouldn't work with `config_subcommand` and `find_handler`.

RedSoxFan: Actually, `find_handler` has some issues that need to be fixed anyway. For example, `seat * layout tabbed` works the same as `layout tabbed` (plus the unnecessary seat overhead)

PR #4270, opened 2019-06-19
config: fix find_handler logic

RedSoxFan:
Without this change, the handlers listed in the config_handlers or
command_handlers arrays (depending on reading or active) in commands.c
would be valid subcommands. To make matters worse, they would also take
precedence over the defined subcommand handlers.

This corrects find_handler to only work on the handler array given
instead of implicitly trying others.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -143,44 +143,34 @@ static int handler_compare(const void *_a, const void *_b) {
 	return strcasecmp(a->command, b->command);
 }
 
-struct cmd_handler *find_handler(char *line, struct cmd_handler *cmd_handlers,
-		int handlers_size) {
-	struct cmd_handler d = { .command=line };
-	struct cmd_handler *res = NULL;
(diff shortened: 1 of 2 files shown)


sway/commands.c
```diff
@@ -143,44 +143,34 @@ static int handler_compare(const void *_a, const void *_b) {
...
+	if (!handlers || !handlers_size) {
+		return NULL;
 	}
+	struct cmd_handler query = { .command = line };
+	return bsearch(&query, handlers,
+			handlers_size / sizeof(struct cmd_handler),
+			sizeof(struct cmd_handler), handler_compare);
+}
 
+static struct cmd_handler *find_handler_ex(char *line,
```
emersion: This is used by only one function. Would it be used by more than one in the future?

RedSoxFan: I'm thinking about changing `config_subcommand` to use this, which would simplify the logic in the command handlers with subcommands (or the ones that have config/command only handlers anyway).

PR #4271, opened 2019-06-20
cmd_mode: allow runtime creation and modification

RedSoxFan:
This allows for modes to be created, bindings to be added to modes, and
bindings to be removed from modes at runtime. Additionally, this also
allows for `mode <mode>` to be deferred in the config to set an initial
mode.

Bindings can already be added and removed from the default mode at
runtime so I don't see why manipulating other modes should be any
different.


ddevault:
Thanks!

PR #4280, opened 2019-06-23
Generate wayland protocol header/data code with custom_target

### LINKED ISSUE #4209: Migrate wayland protocols code generation to custom targets
```
13:48 <pq> mstoeckl, ybakos, emersion, using generator() in meson is sub-optimal, unless your intention really is to have the files re-generated *every* time the file is referenced somewhere. Which can mean to re-generate it multiple times during a single build.
13:48 <pq> use a custom_target() instead
```

See https://github.com/swaywm/wlroots/blob/master/protocol/meson.build

mstoeckl:
Closes  #4209. This change replaces the meson `generator` used to generate protocol source files and headers with a set of `custom_target`s. Each distinct file is now generated only once.

I've also added a fix for the dependency for `lib_server_protos`, which should depend on `wayland_server` instead of `wayland_client`; this doesn't change the dependency list for the `sway` binary, because `libwlroots.so` links both client and server libraries.

### DIFF
--- a/protocols/meson.build
+++ b/protocols/meson.build
@@ -1,85 +1,79 @@
 wl_protocol_dir = wayland_protos.get_pkgconfig_variable('pkgdatadir')
 
-wayland_scanner = find_program('wayland-scanner')
-
-# should check wayland_scanner's version, but it is hard to get
-if wayland_server.version().version_compare('>=1.14.91')
-	code_type = 'private-code'
+wayland_scanner_dep = dependency('wayland-scanner', required: false, native: true)
+if wayland_scanner_dep.found()
+	wayland_scanner = find_program(
+		wayland_scanner_dep.get_pkgconfig_variable('wayland_scanner'),
+		native: true,
+	)
 else
-	code_type = 'code'
+	wayland_scanner = find_program('wayland-scanner', native: true)
 endif
 
-wayland_scanner_code = generator(
-	wayland_scanner,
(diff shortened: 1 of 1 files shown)


protocols/meson.build
```diff
@@ -9,39 +9,14 @@ else
...
-]
-
-server_protocols = [
-	[wl_protocol_dir, 'stable/xdg-shell/xdg-shell.xml'],
-	[wl_protocol_dir, 'unstable/xdg-shell/xdg-shell-unstable-v6.xml'],
-	[wl_protocol_dir, 'unstable/xdg-output/xdg-output-unstable-v1.xml'],
-	[wl_protocol_dir, 'unstable/pointer-constraints/pointer-constraints-unstable-v1.xml'],
-	['wlr-layer-shell-unstable-v1.xml'],
-	['wlr-input-inhibitor-unstable-v1.xml'],
+protocols = [
```
ascent12: I'd much rather see an array of dictionaries used than this.

I also don't know if it's worth diverging from how wlroots does this (which has been using custom_target for ages now).

ddevault: Yeah I would rather be consistent with wlroots

mstoeckl: You're right, the wlroots approach is better. (I'd also been misreading the meson documentation, making my version more complicated than it should have been.)

emersion:
Thanks!

PR #4282, opened 2019-06-23
Use -fmacro-prefix-map to strip build path, when available

### LINKED ISSUE #3680: Build path in binary makes it unreproducible
Since 1211a81aad18bbc4d9e8fb9973238ad8e7e1f688 the _sway_strip_path function uses SWAY_SRC_DIR to strip the build path from the filename. This makes the build unreproducible. There is a note about this in https://github.com/swaywm/sway/blob/d168d65f2c0297bf5662c0f48f5f53705e54a376/include/log.h#L36 but i didn't find an issue for that and TODOs in the code tend to get forgotten.

mstoeckl:
Closes #3680. This uses `-fmacro-prefix-path` to make `__FILE__` strings independent of the filesystem location of the build directory. If the option is not available, it falls back to the second method described in https://github.com/swaywm/sway/issues/3680#issuecomment-463804964 .

The method used to identify the relative path prefix is ugly but straightforward. I am not aware of any simple ways to determine string lengths with Meson itself.

I've tested this change with gcc 9.1.0 (which supports `-fmacro-prefix-path`) and clang 8.0.0 (which does not yet support it, their patch depends on others which are not yet merged).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,8 +13,6 @@ project(
 
 add_project_arguments(
 	[
-		'-DSWAY_SRC_DIR="@0@"'.format(meson.current_source_dir()),
-
 		'-DWL_HIDE_DEPRECATED',
 		'-DWLR_USE_UNSTABLE',
 
@@ -141,6 +139,44 @@ if git.found()
 	endif
 endif
 add_project_arguments('-DSWAY_VERSION=@0@'.format(version), language: 'c')
+
+# Compute the relative path used by compiler invocations.
+source_root = meson.current_source_dir().split('/')
+build_root = meson.build_root().split('/')
+relative_dir_parts = []
+i = 0
+in_prefix = true
+foreach p : build_root
+	if i >= source_root.length() or not in_prefix or p != source_root[i]
+		in_prefix = false
+		relative_dir_parts += '..'
+	endif
+	i += 1
+endforeach
+i = 0
+in_prefix = true
+foreach p : source_root
+	if i >= build_root.length() or not in_prefix or build_root[i] != p
+		in_prefix = false
+		relative_dir_parts += p
+	endif
+	i += 1
+endforeach
+relative_dir = join_paths(relative_dir_parts) + '/'
+
+# Strip relative path prefixes from the code if possible, otherwise hide them.
+if cc.has_argument('-fmacro-prefix-map=/prefix/to/hide=')
+	add_project_arguments(
+		'-fmacro-prefix-map=@0@='.format(relative_dir),
+		language: 'c',
+	)
+else
+	add_project_arguments(
+		'-DSWAY_REL_SRC_DIR="@0@"'.format(relative_dir),
+		language: 'c',
+	)
+endif
+
 
 sway_inc = include_directories('include')
 
--- a/common/log.c
+++ b/common/log.c
@@ -93,20 +93,3 @@ void _sway_log(sway_log_importance_t verbosity, const char *fmt, ...) {
 	sway_log_stderr(verbosity, fmt, args);
 	va_end(args);
 }
-
-// strips the path prefix from filepath
-// will try to strip SWAY_SRC_DIR as well as a relative src dir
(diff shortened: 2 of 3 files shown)


meson.build
```diff
@@ -142,6 +140,46 @@ if git.found()
...
+endforeach
+relative_dir = join_paths(relative_dir_parts) + '/'
+
+# Strip relative path prefixes from the code if possible, otherwise hide them.
+if cc.has_argument('-fmacro-prefix-map=/prefix/to/hide=')
+	add_project_arguments(
+		'-fmacro-prefix-map=@0@='.format(relative_dir),
+		language: 'c',
+	)
+else
```
ascent12: I'd rather just see this fallback removed.
If a compiler that doesn't support this is being used, they'll just get whatever the path is, which 99% of the time would just be `../src/file.c` instead of `src/file.c`, unless their build directory is outside of the source directory.

That currently includes clang, but there us a patch implementing it. I'm not really sure what's holding it up.

meson.build
```diff
@@ -142,6 +140,46 @@ if git.found()
 endif
 add_project_arguments('-DSWAY_VERSION=@0@'.format(version), language: 'c')
 
+# Compute the relative path used by compiler invocations.
+source_root = meson.source_root().split('/')
```
ascent12: It would be better to use `meson.current_source_dir()` instead. It's unlikely sway would ever be used as a subproject, but I think it would return the directory of the parent project instead of sway's actual source root.

This would be more relevant for wlroots, though.

meson.build
```diff
@@ -142,6 +140,46 @@ if git.found()
...
+i = 0
+in_prefix = true
+foreach p : source_root
+	if i >= build_root.length() or not in_prefix or build_root[i] != p
+		in_prefix = false
+		relative_dir_parts += p
+	endif
+	i += 1
+endforeach
+relative_dir = join_paths(relative_dir_parts) + '/'
```
ascent12: An alternative to this algorithm is to just call into the python executable instead:
```meson
py = import('python').find_installation('python3')
relpath_cmd = '''
from os.path import relpath
from sys import argv
print(relpath(argv[1], start=argv[2]))
'''

relpath = run_command(py, '-c', relpath_cmd, meson.current_source_dir(), meson.build_root()).stdout().strip()
```
It's not a new build-time dependency, because meson already needs it.

ddevault: -1

meson.build
mstoeckl: Fixed; fortunately `meson.current_source_dir()` also returns an absolute path.

meson.build
mstoeckl: I really like having uncluttered logs, especially when testing changes from a narrow terminal; the fallback saves 8 characters per line at minimum with out of tree builds, and possibly more when sway is built with a packaging tool.

mstoeckl:
(This last rebase had no changes, and was just to restart the build checks.)

meson.build
ddevault: I'm okay with this but we need to ask meson upstream for a string.length() rather than this hack.

mstoeckl: It turns out that sizeof("string") gives you the string length at compile time. No length computation in meson is needed at all.

mstoeckl:
Updated, fixing the string length computation hack.

meson.build
```diff
@@ -142,6 +140,45 @@ if git.found()
...
+
+# Strip relative path prefixes from the code if possible, otherwise hide them.
+if cc.has_argument('-fmacro-prefix-map=/prefix/to/hide=')
+	add_project_arguments(
+		'-fmacro-prefix-map=@0@='.format(relative_dir),
+		language: 'c',
+	)
+else
+	# For lack of a string.length(), ...
```
ddevault: Does this comment still need to be here?

mstoeckl: Right, removed.

ddevault:
Thanks!

Issue #3547, opened 2019-01-31
Windows not equal sizes after leaving fullscreen.


c-edw:
Similar to #2866.

![image](https://user-images.githubusercontent.com/38229097/52090179-8190e180-25a8-11e9-9934-79113a763062.png)

Steps to produce:
1. Move to an empty workspace
2. Spawn one window and fullscreen it ($mod+F)
3. Spawn two more windows whilst in fullscreen
4. Leave fullscreen ($mod+F)
5. Observe that one of the windows is larger than the other two

On i3, the windows would be of equal size.

athrungithub:
it swaybar?.
 for my it's ok

pedrocr:
Had a look at the code to try and see what was happening. It seems the sequence of events is something like this:

1. The window is fullscreen so now has the full width as it's size
2. New windows are opened and get 0 width
3. When fullscreen is finally exited then `apply_horiz_layout()` is finally called with a full-width size for the first window and 0 for all the others. This creates the uneven layout

Fixing this requires either not keeping the fullscreen width/height in the container width/height or changing the layout calculations to operate on saved_width and saved_height when fullscreen. The first option seems less of a hack but may require deeper surgery in the display code.

PR #4283, opened 2019-06-23
Layout correctly with several new windows

pedrocr:
If there is more than one new window layout correctly by calculating the default size of the new windows using the information of how many of them there are in total.

This helps with issue #3547 but doesn't fix it in all situations. Things now work correctly if the first layout of new windows happens after leaving fullscreen. But if for some reason an `arrange_container()` gets called while we are fullscreen the windows will still be incorrectly sized after `saved_width`/`saved_height` get used to restore the first window's size before going fullscreen.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -87,6 +87,7 @@ struct sway_container {
 	double width, height;
 	double saved_x, saved_y;
 	double saved_width, saved_height;
+	bool saved_dims;
 
 	// These are in layout coordinates.
 	double content_x, content_y;
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -39,7 +39,11 @@ static void apply_horiz_layout(list_t *children, struct wlr_box *parent) {
 			}
 		}
 		container_remove_gaps(child);
-		total_width += child->width;
+		if (child->saved_dims) {
+			total_width += child->saved_width;
+		} else {
+			total_width += child->width;
+		}
 	}
 	double scale = parent->width / total_width;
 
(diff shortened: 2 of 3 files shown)


pedrocr:
I've pushed as a second commit what may be a full fix for #3547. The solution seems fairly sane so if it hasn't regressed anything #4284 can probably be abandoned.

RyanDwyer:
I wonder if it would be worth changing the approach to one which stores width/height weighting values in each container rather than doing it by pixels. So each new container would default to weighting values of 1. If you had 3 siblings with weights 2, 1, 1 then the sizes would be 50%, 25%, 25%.

This would mean the fullscreen container would retain the weighting of 1, the siblings would also have weighting 1, and everything would size equally when exiting fullscreen. You wouldn't need to track the saved width or saved height.

The resize command would need to be changed to set the weighting instead of the width and height. I don't think it would be too difficult.

pedrocr:
I don't think that changes the problem really. You'd just be replacing `saved_width` with `saved_width_weight`, no? Since the layout code already does a scale calculation on each run the end result is the same as doing it in pixels.

pedrocr:
Nevermind, I think I see what you mean. I'll try a version with that.

pedrocr:
New pull request sent as #4293. The code does look a bit better and should be less brittle. @RyanDwyer could you please review that one?

RyanDwyer:
This PR adds a significant amount of complexity to fix the problem. I'd be happy to approve the first commit only, but I understand it doesn't fix all the cases. If the first commit were merged, what would the steps be to reproduce the remaining issue? Just trying to understand what the latter commits are solving.

pedrocr:
To reproduce the issue after the first commit:

1. Launch three containers horizontally
2. In one of them switch to vertical splits
3. Make that container fullscreen
4. Launch two new containers vertically
5. Unfullscreen

I'm pretty sure this PR can be abandoned in favor of #4293 though. The weights calculation can probably be made simpler but the solution is just much more robust.

pedrocr:
Closed in favor of #4293

PR #4284, opened 2019-06-23
Don't manipulate size/pos of fullscreen containers

### LINKED ISSUE #3547: Windows not equal sizes after leaving fullscreen.

pedrocr:
In fullscreen mode position/dimensions of the fullscreen container were being changed and then restored on unfullscreen. This would break things that manipulated those dimensions while the window was fullscreen.
Instead have the rendering code handle the positioning directly and leave the container information alone.

Fixes #3547

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -85,8 +85,6 @@ struct sway_container {
 	// Includes borders
 	double x, y;
 	double width, height;
-	double saved_x, saved_y;
-	double saved_width, saved_height;
 
 	// These are in layout coordinates.
 	double content_x, content_y;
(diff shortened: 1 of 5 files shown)


pedrocr:
What broke? Everything seems to work fine for me.

pedrocr:
Had a look again and `FULLSCREEN_GLOBAL` was still manipulating container size/pos so I removed that. Don't know if that is what was broken for you.

RedSoxFan:
> What broke? 

Fullscreening a view seems to work fine. Fullscreening a container renders at the same size it was instead of using the full output.

pedrocr:
Closed in favor of #4293

PR #4287, opened 2019-06-27
tree: set correct border value before creating floater

arandomhuman:
Currently, `border` is set to `B_CSD` after setting coordinates, messing up the `content_*` fields.


ddevault:
Thanks!

PR #4293, opened 2019-06-28
Fix layout when fullscreen

### LINKED ISSUE #3547: Windows not equal sizes after leaving fullscreen.

### LINKED ISSUE #4297: As more windows are opened their sizes become unequal
* Debug Log: https://gist.github.com/pedrocr/b42eda00a349a79abe6b4f501cf3be1d

Start a default session and just open a lot of containers. Just leaving `$mod+Enter` pressed will do it in the default config. As the windows are opened they become increasingly uneven with newer windows being larger than the older ones. Something like this:

unevenlysizedwindows

pedrocr:
This is an alternative to #4283 that's implemented using width/height fractions of the parent container. It seems to make the code cleaner, although it also required changing the resize command to match. As far as I can tell everything works fine.

Fixes #3547
Fixes #4297

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -174,10 +174,14 @@ void container_resize_tiled(struct sway_container *con,
 		if (prev && prev->width - sibling_amount < MIN_SANE_W) {
 			return;
 		}
-		con->width += amount;
-		next->width -= sibling_amount;
+
+		con->width_fraction +=
+			((double)amount / con->width) * con->width_fraction;
+		next->width_fraction -=
+			((double)sibling_amount / con->width) * con->width_fraction;
 		if (prev) {
-			prev->width -= sibling_amount;
+			prev->width_fraction -=
+				((double)sibling_amount / con->width) * con->width_fraction;
 		}
 	} else {
 		if (con->height + amount < MIN_SANE_H) {
@@ -189,10 +193,14 @@ void container_resize_tiled(struct sway_container *con,
 		if (prev && prev->height - sibling_amount < MIN_SANE_H) {
 			return;
 		}
-		con->height += amount;
-		next->height -= sibling_amount;
+
+		con->height_fraction +=
+			((double)amount / con->height) * con->height_fraction;
+		next->height_fraction -=
+			((double)sibling_amount / con->height) * con->height_fraction;
 		if (prev) {
-			prev->height -= sibling_amount;
+			prev->height_fraction -=
+				((double)sibling_amount / con->height) * con->height_fraction;
 		}
 	}
 
@@ -280,10 +288,11 @@ static struct cmd_results *resize_adjust_tiled(uint32_t axis,
 		}
 	}
 
-	double old_width = current->width;
-	double old_height = current->height;
+	double old_width = current->width_fraction;
+	double old_height = current->height_fraction;
 	container_resize_tiled(current, axis, amount->amount);
-	if (current->width == old_width && current->height == old_height) {
+	if (current->width_fraction == old_width &&
+			current->height_fraction == old_height) {
 		return cmd_results_new(CMD_INVALID, "Cannot resize any further");
 	}
 	return cmd_results_new(CMD_SUCCESS, NULL);
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -18,39 +18,49 @@ static void apply_horiz_layout(list_t *children, struct wlr_box *parent) {
 		return;
 	}
 
-	// Count the number of new windows we are resizing
+	// Count the number of new windows we are resizing, and how much space
+	// is currently occupied
 	int new_children = 0;
+	double current_width_fraction = 0;
 	for (int i = 0; i < children->length; ++i) {
 		struct sway_container *child = children->items[i];
-		if (child->width <= 0) {
+		current_width_fraction += child->width_fraction;
+		if (child->width_fraction <= 0) {
 			new_children += 1;
 		}
 	}
 
-	// Calculate total width of children
-	double total_width = 0;
+	// Calculate each height fraction
+	double total_width_fraction = 0;
 	for (int i = 0; i < children->length; ++i) {
 		struct sway_container *child = children->items[i];
-		if (child->width <= 0) {
-			if (children->length > new_children) {
-				child->width = parent->width / (children->length - new_children);
+		if (child->width_fraction <= 0) {
+			if (current_width_fraction <= 0) {
+				child->width_fraction = 1.0;
+			} else if (children->length > new_children) {
+				child->width_fraction = current_width_fraction /
+					(children->length - new_children);
 			} else {
-				child->width = parent->width;
+				child->width_fraction = current_width_fraction;
 			}
 		}
-		container_remove_gaps(child);
-		total_width += child->width;
+		total_width_fraction += child->width_fraction;
 	}
-	double scale = parent->width / total_width;
+	// Normalize width fractions so the sum is 1.0
+	for (int i = 0; i < children->length; ++i) {
+		struct sway_container *child = children->items[i];
+		child->width_fraction /= total_width_fraction;
+	}
 
 	// Resize windows
 	sway_log(SWAY_DEBUG, "Arranging %p horizontally", parent);
 	double child_x = parent->x;
 	for (int i = 0; i < children->length; ++i) {
 		struct sway_container *child = children->items[i];
+		container_remove_gaps(child);
 		child->x = child_x;
 		child->y = parent->y;
-		child->width = floor(child->width * scale);
+		child->width = floor(child->width_fraction * parent->width);
 		child->height = parent->height;
 		child_x += child->width;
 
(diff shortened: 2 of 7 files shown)


RyanDwyer:
This isn't quite what I had in mind. I meant setting each weighting value to 1 when the container is created then not changing them until the user resizes them. They would not sum to 1. In all places where we zero the width and height, we would make it reset the weighting to 1 instead. There would then be no reason to have the `if (child->width <= 0)` branch when arranging, and therefore no reason to count the children to see how many are new.

I consider the above approach to be more of a refactor than a bug fix. A fix should be merged first, then we can look into how well weighting would work and whether it's worth it (that alliteration was not intentional).

pedrocr:
Merging a fix first would be merging #4283. That one is a fix without changing the way the code works. But that's just uglier and it has to manipulate `saved_width` and `saved_height`. I see what you mean about simplifying the weights calculation by not doing normalization. That may work with less special cases. I'll give it a try.

pedrocr:
I think this is what you meant. It's indeed much simpler. I started along these lines last night and then for some reason didn't think it was going to work. In my defense it was late and I was tired :)

The diff right now is probably good but the commit sequence arrives at it in a roundabout way. I'll squash/rework stuff into a cleaner history.

pedrocr:
I've condensed it into two commits. The first does the layout changes, the second implements the resize command based on the new layout method. So hopefully this is mergeable now.

pedrocr:
I've just come across a behavior that probably means normalization to the sum being 1.0 is actually needed:

1. Open two windows
2. Resize one window to be larger than the other
3. Close one of the windows
4. Open a new window

The expected result would be two equally sized windows but because the existing window has a factor that's different than 1.0 the new window will not be half size and can be larger or smaller depending on the remaining factor.

So the previous code is actually the one that retains current behavior while fixing bugs.

pedrocr:
I've pushed the normalizing solution again, as it seems needed to have correct behavior. It may still be possible to make the layout functions simpler but probably not by too much without altering behavior.

sway/tree/arrange.c
```diff
@@ -18,30 +18,51 @@ static void apply_horiz_layout(list_t *children, struct wlr_box *parent) {
...
+	}
+
+	// Calculate each height fraction
+	double total_width_fraction = 0;
+	for (int i = 0; i < children->length; ++i) {
+		struct sway_container *child = children->items[i];
+		if (child->width_fraction <= 0) {
+			if (current_width_fraction <= 0) {
+				child->width_fraction = 1.0;
 			} else {
```
RedSoxFan: Use a `if, else if, else`

sway/tree/arrange.c
```diff
@@ -58,31 +79,52 @@ static void apply_vert_layout(list_t *children, struct wlr_box *parent) {
...
+	}
+
+	// Calculate each height fraction
+	double total_height_fraction = 0;
+	for (int i = 0; i < children->length; ++i) {
+		struct sway_container *child = children->items[i];
+		if (child->height_fraction <= 0) {
+			if (current_height_fraction <= 0) {
+				child->height_fraction = 1.0;
 			} else {
```
RedSoxFan: Use a `if, else if, else`

pedrocr:
Thanks for the review. Do you have an example of what you mean with the floating containers rendering on top of each other? I just tested with GIMP and its floating windows and everything seems to work as expected.

I've pushed a fix for the issue of reducing size on going from floating to tiled.

RedSoxFan:
> Do you have an example of what you mean with the floating containers rendering on top of each other?

Float a view, split it (horizontally or vertically, doesn't matter), open another view. Instead of being split, the two views take up the full size of the floating container. Using the opacity command may help visualize it.

pedrocr:
Thanks. I've pushed a fix.

RedSoxFan:
A couple more issues:

If you create `H[viewA V[viewB viewC]` and move viewA right and left multiple times, it'll continuously shrink.

If you have one output with `H[viewA viewB viewC]` that are all equal size and move viewC to the output to the right and then move it back, the sizing is viewA 25%, viewB 25%, viewC 50%. Moving viewC further to the left retains the 50% width.

pedrocr:
I'll have a look at these but I don't know what the correct behavior is. This PR sizes things by a fraction of the total container and when a view is coming from somewhere else it's treated as if it's new. This will naturally be different behavior than using the width/height of existing views as the starting point to rearrange the container. So the behavior will never match current behavior in all situations. If that's wanted I did another PR that adjusted `saved_width` and `saved_height` as containers got their layout arranged while something is fullscreen. It didn't fix things like #4297 but maybe that's enough of an edge case to ignore and we should be going back to that. It all depends on what the wanted behavior is.

pedrocr:
Both cases are fixed now, thanks again for looking at this.

sway/commands/resize.c
```diff
@@ -280,10 +288,10 @@ static struct cmd_results *resize_adjust_tiled(uint32_t axis,
 		}
 	}
 
-	double old_width = current->width;
-	double old_height = current->height;
+	double old_width = current->width_fraction;
+	double old_height = current->height_fraction;
 	container_resize_tiled(current, axis, amount->amount);
-	if (current->width == old_width && current->height == old_height) {
+	if (current->width_fraction == old_width && current->height_fraction == old_height) {
```
RedSoxFan: style: can you split this after the `&&` so that it's within 80 characters?

sway/tree/arrange.c
```diff
@@ -18,30 +18,51 @@ static void apply_horiz_layout(list_t *children, struct wlr_box *parent) {
...
-		container_remove_gaps(child);
-		total_width += child->width;
+		total_width_fraction += child->width_fraction;
+	}
+	// Normalize width fractions so the sum is 1.0
+	for (int i = 0; i < children->length; ++i) {
+		struct sway_container *child = children->items[i];
+		child->width_fraction /= total_width_fraction;
+
+		sway_log(SWAY_DEBUG, "Child %d with %lf width_fraction", i, child->width_fraction);
```
RedSoxFan: style: 80 columns

sway/tree/arrange.c
```diff
@@ -18,30 +18,51 @@ static void apply_horiz_layout(list_t *children, struct wlr_box *parent) {
...
 	}
-	double scale = parent->width / total_width;
 
 	// Resize windows
-	sway_log(SWAY_DEBUG, "Arranging %p horizontally", parent);
+	sway_log(SWAY_DEBUG, "Arranging %p horizontally, x %d, width %d", parent, parent->x, parent->width);
```
RedSoxFan: style: 80 columns

sway/tree/arrange.c
```diff
@@ -50,6 +71,8 @@ static void apply_horiz_layout(list_t *children, struct wlr_box *parent) {
 			child->width = parent->x + parent->width - child->x;
 		}
 		container_add_gaps(child);
+
+		sway_log(SWAY_DEBUG, "Child %d at %lf width %lf", i, child->x, child->width);
```
RedSoxFan: style: 80 columns

sway/commands/resize.c
pedrocr: Done and I think I followed the style now. Only one extra tab to indent a second line instead of spaces to align with the previous line. I also generally use 2 space tabs so my editor was giving me more room to fit in 80 columns than should be the reference with 4 spaces.

sway/tree/arrange.c
pedrocr: These are all stray changes from debugging stuff. I've removed them although it may make sense to add a little bit more to the log to be able to diagnose weird things that crop up in the future. Not sure how much.

pedrocr:
I've fixed those now. The commits already have the `Fixes #xxxx` lines but I've added them to the PR as well.

sway/commands/resize.c
RedSoxFan: It actually needs an additional indentation level here to differentiate it from the body

pedrocr: Yep, misread that in `CONTRIUTING.md`. It's fixed now.

pedrocr:
Should it really be a single commit? At least the first commit is independent and the others help explain the changes.

RedSoxFan:
> Should it really be a single commit? At least the first commit is independent and the others help explain the changes.

I guess the first commit can remain separate, but the rest of the commits are a logical grouping. Without commits 3-8, commit 2 is not in a usable state and commits 3-8 don't make sense on there own as they are essentially just intermediary fixes for the second commit. Feel free to include as much information as you want in the commit message to explain the changes, but the commit history should be clean and logically grouped together.

pedrocr:
It depends on the project. Some would prefer a sequence of commits that does the work building upon each other.

RedSoxFan:
> It depends on the project. Some would prefer a sequence of commits that does the work building upon each other.

A general rule for sway's commit history is a change log should be able to be generated from the short log. Also, I appreciate the work you are putting into this and not trying to come off rude, but please don't argue sway's conventions by saying what other project use. They are set and are not going to change.

RedSoxFan:
Thanks!

pedrocr:
Thanks for the patience to review it! I'll get the gaps PR ready based on this.

PR #4298, opened 2019-06-30
Rework the gaps code so sizes are correct and to prevent broken layouts

### LINKED ISSUE #4294: Very large inner gaps break layout
* Sway Version: 1.1-rc1-51-gf5d1c272 (Jun 28 2019, branch 'master')
* Debug Log: https://gist.github.com/pedrocr/f060091cc09fb117ae6fa100504b9230

When using very large gaps the layout becomes broken. To reproduce do something like `swaymsg gaps inner all set 200` and then open a bunch of terminals horizontally. At some point the layout becomes broken with incorrectly rendered windows. Something like this:

broken_gaps_layout

These are not just incorrectly sized windows, the rendering seems actually broken.

### LINKED ISSUE #4296: Gaps create unequally sized windows
* Debug Log: https://gist.github.com/pedrocr/834aa8956c30510f2d2bb2bf17afdef4

When using gaps windows that are the same size become unequally sized. To make it obvious just open two windows side by side a do something like:

```
swaymsg gaps inner all set 500
```

The end result will be something like this:

unevengaps

While large gaps make this very noticeable it happens on any sized gap.

### LINKED ISSUE #4304: Negative outer gaps break workspaces
* Sway Version: sway version 1.1-rc1-51-gf5d1c272 (Jul  2 2019, branch 'master')
* Debug Log: https://gist.github.com/pedrocr/b5060f692f460510f2ba6960ccdf6b07

Negative outer gaps are not controlled for properly. Open a single terminal and do something like:

```
swaymsg gaps outer all minus 300
```

And you will get a window that's now covering more than the total screen instead of a no-op.

pedrocr:
Fixes for issues #4296 and #4294. They're built on top of each other because they're touching the same code and the reworking of the calculation to make the gaps even makes it easier to then write the code to cap the gaps.

The weird addition of `#include "pango.h"` in ` sway/tree/workspace.c` was just to get `MIN()` and `MAX()`. This is probably not ok and sway should instead include its own definitions of these macros.

--
Fixes #4294  
Fixes #4296
Fixes #4304

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -34,4 +34,9 @@ const char *sway_wl_output_subpixel_to_string(enum wl_output_subpixel subpixel);
 
 bool set_cloexec(int fd, bool cloexec);
 
+#undef MAX
+#define MAX(a, b)  (((a) > (b)) ? (a) : (b))
+#undef MIN
+#define MIN(a, b)  (((a) < (b)) ? (a) : (b))
+
 #endif
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1226,13 +1226,17 @@ void container_add_gaps(struct sway_container *c) {
 		}
 	}
 
-	int gaps1 = ws->gaps_inner / 2;
-	int gaps2 = ws->gaps_inner - gaps1;
+	int gaps_horizontal = MAX(0, MIN(ws->gaps_inner, c->width - MIN_SANE_W));
+	int gaps_horizontal1 = gaps_horizontal / 2;
+	int gaps_horizontal2 = gaps_horizontal - gaps_horizontal1;
+	c->current_gaps.left = gaps_horizontal1;
+	c->current_gaps.right = gaps_horizontal2;
 
-	c->current_gaps.top = gaps1;
-	c->current_gaps.bottom = gaps2;
-	c->current_gaps.left = gaps1;
-	c->current_gaps.right = gaps2;
+	int gaps_vertical = MAX(0, MIN(ws->gaps_inner, c->height - MIN_SANE_H));
+	int gaps_vertical1 = gaps_vertical / 2;
+	int gaps_vertical2 = gaps_vertical - gaps_vertical1;
+	c->current_gaps.top = gaps_vertical1;
+	c->current_gaps.bottom = gaps_vertical2;
 
 	c->x += c->current_gaps.left;
 	c->y += c->current_gaps.top;
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1226,10 +1226,13 @@ void container_add_gaps(struct sway_container *c) {
 		}
 	}
 
-	c->current_gaps.top = c->y == ws->y ? ws->gaps_inner : 0;
-	c->current_gaps.right = ws->gaps_inner;
-	c->current_gaps.bottom = ws->gaps_inner;
-	c->current_gaps.left = c->x == ws->x ? ws->gaps_inner : 0;
+	int gaps1 = ws->gaps_inner / 2;
+	int gaps2 = ws->gaps_inner - gaps1;
+
+	c->current_gaps.top = gaps1;
+	c->current_gaps.bottom = gaps2;
+	c->current_gaps.left = gaps1;
+	c->current_gaps.right = gaps2;
 
 	c->x += c->current_gaps.left;
 	c->y += c->current_gaps.top;
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -746,15 +746,19 @@ void workspace_add_gaps(struct sway_workspace *ws) {
 	} else {
 		// For tiled containers we need to add the half-gap all around the edge
 		// to match the half gaps that the children have all already added around
-		// themselves. We use the opposite gaps1/gaps2 order so that the sum adds
-		// up to the correct total gap size in all circumstances.
-		int gaps1 = ws->gaps_inner / 2;
-		int gaps2 = ws->gaps_inner - gaps1;
+		// themselves. We use the opposite order for the two halves so that the
+		// sum adds up to the correct total gap size in all circumstances.
+		int gaps_horizontal = MAX(0, MIN(ws->gaps_inner, ws->width - MIN_SANE_W));
+		int gaps_horizontal1 = gaps_horizontal / 2;
+		int gaps_horizontal2 = gaps_horizontal - gaps_horizontal1;
+		ws->current_gaps.left += gaps_horizontal2;
+		ws->current_gaps.right += gaps_horizontal1;
 
-		ws->current_gaps.top += gaps2;
-		ws->current_gaps.bottom += gaps1;
-		ws->current_gaps.left += gaps2;
-		ws->current_gaps.right += gaps1;
+		int gaps_vertical = MAX(0, MIN(ws->gaps_inner, ws->height - MIN_SANE_H));
+		int gaps_vertical1 = gaps_vertical / 2;
+		int gaps_vertical2 = gaps_vertical - gaps_vertical1;
+		ws->current_gaps.top += gaps_vertical2;
+		ws->current_gaps.bottom += gaps_vertical1;
 	}
 
 	ws->x += ws->current_gaps.left;
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -743,6 +743,18 @@ void workspace_add_gaps(struct sway_workspace *ws) {
 		ws->current_gaps.right += ws->gaps_inner;
 		ws->current_gaps.bottom += ws->gaps_inner;
 		ws->current_gaps.left += ws->gaps_inner;
+	} else {
+		// For tiled containers we need to add the half-gap all around the edge
+		// to match the half gaps that the children have all already added around
+		// themselves. We use the opposite gaps1/gaps2 order so that the sum adds
+		// up to the correct total gap size in all circumstances.
+		int gaps1 = ws->gaps_inner / 2;
+		int gaps2 = ws->gaps_inner - gaps1;
+
+		ws->current_gaps.top += gaps2;
+		ws->current_gaps.bottom += gaps1;
+		ws->current_gaps.left += gaps2;
+		ws->current_gaps.right += gaps1;
 	}
 
 	ws->x += ws->current_gaps.left;
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -54,21 +54,6 @@ struct sway_output *workspace_get_initial_output(const char *name) {
 	return root->outputs->length ? root->outputs->items[0] : root->noop_output;
 }
 
-static void prevent_invalid_outer_gaps(struct sway_workspace *ws) {
-	if (ws->gaps_outer.top < -ws->gaps_inner) {
-		ws->gaps_outer.top = -ws->gaps_inner;
-	}
-	if (ws->gaps_outer.right < -ws->gaps_inner) {
-		ws->gaps_outer.right = -ws->gaps_inner;
-	}
-	if (ws->gaps_outer.bottom < -ws->gaps_inner) {
-		ws->gaps_outer.bottom = -ws->gaps_inner;
-	}
-	if (ws->gaps_outer.left < -ws->gaps_inner) {
-		ws->gaps_outer.left = -ws->gaps_inner;
-	}
-}
-
 struct sway_workspace *workspace_create(struct sway_output *output,
 		const char *name) {
 	if (output == NULL) {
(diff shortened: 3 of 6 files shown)


sway/tree/container.c
```diff
@@ -1226,10 +1228,13 @@ void container_add_gaps(struct sway_container *c) {
 		}
 	}
 
-	c->current_gaps.top = c->y == ws->y ? ws->gaps_inner : 0;
-	c->current_gaps.right = ws->gaps_inner;
-	c->current_gaps.bottom = ws->gaps_inner;
-	c->current_gaps.left = c->x == ws->x ? ws->gaps_inner : 0;
+	int gaps_horizontal = MAX(0, ws->gaps_inner + MIN(0, c->width - ws->gaps_inner - MIN_SANE_W));
```
RedSoxFan: style: 80 columns

However, I think this would be more readable as:
```c
int gaps_horizontal = MAX(0, MIN(ws->gaps_inner, c->width - MIN_SANE_W));
```

(And similar for the others)

pedrocr:
> Considering that these are not defined in `pango.h`, but one of the headers included in that file, I'm not a fan of this approach.

Definitely. My next sentence was exactly that sway should include its own macros. `pango.h` was just a placeholder, that reused what `container.c` already did. Please let me know where in the sway tree it makes sense to include those macros.

> Another thing to consider is making it so the gap clamping considers both width and height so that they are equal in both directions. I'm not sure which method is better. Thoughts?

Whenever the gaps get clamped the result is always going to be a bit odd and this should only really happen in extreme cases. But I think this way should work better. Right now if the window doesn't have enough space horizontally but does vertically the vertical gaps to the other containers are all correct. The way you suggest would result in unequal vertical gaps even though the window fits fine vertically.

> It should also be noted in sway.5 that the inner gaps may be smaller or omitted entirely when there isn't enough space available.

Makes sense, I'll add that. I reused the sane width/height values from the resize command. I don't know if those values should be placed somewhere global in the code and possibly also referred to in the documentation somewhere.

sway/tree/container.c
```diff
@@ -1226,10 +1228,13 @@ void container_add_gaps(struct sway_container *c) {
...
+	c->current_gaps.left = gaps_horizontal / 2;
+	c->current_gaps.right = gaps_horizontal - (gaps_horizontal / 2);
```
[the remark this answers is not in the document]
pedrocr: I did that initially. But in `workspace.c` that doesn't work because the gaps are added with `+=` instead of set to the value. So to make everything parallel I did it like this. The performance impact is surely irrelevant but I can go back to calculating `gaps_horizontal1`, and `gaps_horizontal2` the same way in both locations and only then doing the assignments. It makes the code less compact but may be easier to understand.

sway/tree/container.c
pedrocr: Yep, that does look better, thanks.

pedrocr:
I've moved the macros and the static values to `include/util.h` and added a note to the manpage.

sway/tree/workspace.c
```diff
@@ -734,7 +734,29 @@ void workspace_add_gaps(struct sway_workspace *ws) {
...
-	ws->current_gaps = ws->gaps_outer;
+	// Since we can't rely on outer gaps being small enough for the specific
+	// output we are in, cap them so at least the minimum sane space is left.
+	if (ws->width - ws->gaps_outer.left - ws->gaps_outer.right > MIN_SANE_W) {
+		ws->current_gaps.left = ws->gaps_outer.left;
+		ws->current_gaps.right = ws->gaps_outer.right;
+	} else {
+		int total_gap = MAX(0, ws->width - MIN_SANE_W);
+		double left_gap_frac = ((float) ws->gaps_outer.left /
+			((float) ws->gaps_outer.left + (float) ws->gaps_outer.right));
```
RedSoxFan: This will produce a `SIGFPE` when `ws->gaps_outer.left + ws->gaps_outer.right == 0`

sway/tree/workspace.c
```diff
@@ -734,7 +734,29 @@ void workspace_add_gaps(struct sway_workspace *ws) {
...
+		ws->current_gaps.left = left_gap_frac * total_gap;
+		ws->current_gaps.right = total_gap - ws->current_gaps.left;
+	}
+	if (ws->height - ws->gaps_outer.top - ws->gaps_outer.bottom > MIN_SANE_H) {
+		ws->current_gaps.top = ws->gaps_outer.top;
+		ws->current_gaps.bottom = ws->gaps_outer.bottom;
+	} else {
+		int total_gap = MAX(0, ws->height - MIN_SANE_H);
+		double top_gap_frac = ((float) ws->gaps_outer.top /
+			((float) ws->gaps_outer.top + (float) ws->gaps_outer.bottom));
```
RedSoxFan: Like above, this can produce a `SIGFPE`

include/util.h
```diff
@@ -34,4 +34,11 @@ const char *sway_wl_output_subpixel_to_string(enum wl_output_subpixel subpixel);
 
 bool set_cloexec(int fd, bool cloexec);
 
+#undef MAX
+#define MAX(a, b)  (((a) > (b)) ? (a) : (b))
+#undef MIN
+#define MIN(a, b)  (((a) < (b)) ? (a) : (b))
+
+static const int MIN_SANE_W = 100, MIN_SANE_H = 60;
```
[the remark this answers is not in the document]
pedrocr: I just kept them as they were, they also made more sense to me as defines.

sway/tree/workspace.c
pedrocr: Only if you find a workspace that's <= than MIN_SANE_W :) But yeah it should be caught and is simple enough to fix.

pedrocr:
I've fixed all the issues and reimplemented the workspace gaps handling. The code is now correct as it clamps gaps after both outer and inner gaps are applied. Only when we have a total gap (and outer gaps can supposedly be negative) can we then clamp with the correct logic. The code also becomes simpler that way.

pedrocr:
I've found another bug with negative outer gaps filed as #4304 and added a fix for that as well.

sway/tree/container.c
```diff
@@ -1226,10 +1226,17 @@ void container_add_gaps(struct sway_container *c) {
 		}
 	}
 
-	c->current_gaps.top = c->y == ws->y ? ws->gaps_inner : 0;
-	c->current_gaps.right = ws->gaps_inner;
-	c->current_gaps.bottom = ws->gaps_inner;
-	c->current_gaps.left = c->x == ws->x ? ws->gaps_inner : 0;
+	int gaps_horizontal = MAX(0, MIN(ws->gaps_inner, c->width - MIN_SANE_W));
+	int gaps_horizontal1 = gaps_horizontal / 2;
+	int gaps_horizontal2 = gaps_horizontal - gaps_horizontal1;
```
RedSoxFan: I understand the idea of making this mirror the workspace gaps section, but I think the cleanness of the following outweighs it.
```c
int gaps_horizontal = MAX(0, MIN(ws->gaps_inner, c->width - MIN_SANE_W));
c->current_gaps.left = gaps_horizontal / 2;
c->current_gaps.right = gaps_horizontal - c->current_gaps.left;
```

sway/tree/workspace.c
```diff
@@ -743,6 +725,49 @@ void workspace_add_gaps(struct sway_workspace *ws) {
 		ws->current_gaps.right += ws->gaps_inner;
 		ws->current_gaps.bottom += ws->gaps_inner;
 		ws->current_gaps.left += ws->gaps_inner;
+	} else {
+		// For tiled containers we need to add the half-gap all around the edge
+		// to match the half gaps that the children have all already added around
+		// themselves. We use the opposite order for the two halves so that the
+		// sum adds up to the correct total gap size in all circumstances.
+		int gaps_horizontal1 = ws->gaps_inner / 2;
```
RedSoxFan: I think the following would be cleaner:
```c
int inner_gaps_left = ws->gaps_inner / 2;
ws->current_gaps.left += inner_gaps_left;
ws->current_gaps.right = ws->gaps_inner - inner_gaps_left;
```

sway/tree/workspace.c
```diff
@@ -743,6 +725,49 @@ void workspace_add_gaps(struct sway_workspace *ws) {
...
+		ws->current_gaps.top += gaps_vertical2;
+		ws->current_gaps.bottom += gaps_vertical1;
+	}
+
+	// Prevent the resulting gaps from turning out negative, since outer gaps can
+	// be negative and nothing ensures they will not be larger than the current
+	// inner gaps.
+	int mingaps = ws->layout == L_TABBED ||
+	              ws->layout == L_STACKED ? 0 : -ws->gaps_inner;
+	ws->current_gaps.left = MAX(mingaps, ws->current_gaps.left);
```
RedSoxFan: I don't think this is right. The outer gaps should be able to negate the inner gaps for that side, but this would allow for it to negate double the inner gaps. I think these should all be like `ws->current_gaps.left = MAX(0, ws->current_gaps.left)`. 

_It looks like this function was just missing a call to `prevent_invalid_outer_gaps` in between setting the outer gaps and inner gaps._

sway/tree/container.c
pedrocr: I had that like that once upon a time :)

sway/tree/workspace.c
pedrocr: How do you get a double negation? I've gotten rid of `prevent_invalid_outer_gaps` completely, it doesn't seem workable to cover all cases with that. It just does that calculation on apply instead. So it allows the outer gaps to at most be `-ws->gaps_inner` for tiled containers to then allow those to be negated by the child gaps. It also seems to make more sense to not manipulate the values the user actually sets himself as the workspace may change sizes by moving from one output to another.

This reminds me of another corner case. If the gaps are large enough that the gap clamping will apply in the children the negative gap will not be properly negated... That doesn't seem trivial to fix.

RedSoxFan: > How do you get a double negation?

Any values where outer `outer == -2 * inner` should work, but the following should make it easily visible
```
swaymsg -- gaps outer all set -400, gaps inner all set 200
```

> So it allows the outer gaps to at most be `-ws->gaps_inner`

Currently, it is being checked after inner gaps are calculated so it actually makes sure that *all* gaps are at most `-ws->gaps_inner`, which is incorrect.

> This reminds me of another corner case. If the gaps are large enough that the gap clamping will apply in the children the negative gap will not be properly negated

Can you try explaining that further? When would that occur?

pedrocr: > Currently, it is being checked after inner gaps are calculated so it actually makes sure that all gaps are at most `-ws->gaps_inner`, which is incorrect.

It's correct because the children will then apply `ws->gaps_inner` themselves and thus negate that.

> Can you try explaining that further? When would that occur?

Set outer gaps to -1000 and inner gaps to +1000.  Now start opening windows. At first it's correct with no outer gaps but gaps within windows. As you open more windows and the inner gaps get clamped the windows start falling outside the screen. This is going to be annoying to fix...

pedrocr: It's still wrong though, because clamping needs to be to half size only per edge...

But this doesn't work generally because of the other bug. I'll have to think of a better solution.

pedrocr: I've pushed a fix for the half-sized per edge, but this is not mergeable like this because of the other bug. It also doesn't even work always like this because since after the gap clamping it can be negative, the code that adjusts outer gaps can return broken results.

pedrocr:
I think I have a plan to make everything work but it's a completely different solution than this. It would go something like:

1. The workspace gaps code would add the full gap around the screen (outer and inner) and do the clamping calculations for those. That fixes #4304 and this new issue I pointed out.
2. The container gaps code would be integrated with the tiling layout code and would only add the gaps between containers. The layout code has all the information needed to do proper clamping and adjusting of sizes between containers to fix #4294 and #4296.

I haven't written the code yet but it only makes sense to do that on top of #4293.

pedrocr:
I've implemented the plan on top of #4293 and it worked really well:

https://github.com/pedrocr/sway/compare/layout_improvement...pedrocr:gaps_rework

The code ends up simpler and easier to reason about. And even after the clamping code it's a net reduction in code:

```
$ git diff layout_improvement gaps_rework | diffstat
 include/sway/tree/container.h |   12 -------
 include/sway/tree/node.h      |    3 +
 include/util.h                |    5 +++
 sway/commands/move.c          |    1 
 sway/commands/resize.c        |    2 -
 sway/tree/arrange.c           |   38 +++++++++++++++--------
 sway/tree/container.c         |   67 ------------------------------------------
 sway/tree/view.c              |   16 ++--------
 sway/tree/workspace.c         |   52 ++++++++++++++------------------
 9 files changed, 60 insertions(+), 136 deletions(-)
```

pedrocr:
Closed in favor of #4310

Issue #4308, opened 2019-07-06
Large outer gaps break layout


pedrocr:
* Sway Version: sway version 1.1-rc1-51-gf5d1c272 (Jul  2 2019, branch 'master')
* Debug Log: https://gist.github.com/pedrocr/6831ea744d7904d4aebe599888092f5f
* Configuration File: default config

Large outer gaps are not controlled for. Open a single terminal and do something like:

```
swaymsg gaps outer all set 30000
```

And all your windows will now disappear.

PR #4310, opened 2019-07-06
Rework the gaps code to make it simpler and correct

### LINKED ISSUE #4296: Gaps create unequally sized windows

### LINKED ISSUE #4304: Negative outer gaps break workspaces

pedrocr:
Rework the gaps code to fix a series of bugs. The general approach is:

1. In tiled containers apply only gaps between containers directly in the layout code. This ensures proper sizing of containers. Fixes #4296 and #4294
2. For the gaps around the outside of workspaces (both outer and inner), calculate them together and apply them in the workspace code. At the end make sure they are positive and leave enough usable space in the container. Fixes #4304 and #4308

Depends on the layout PR (#4293) as it currently builds on top of it in the parts it touches the same code. But it doesn't depend on it functionally, so it can be reworked to apply to master.

### DIFF
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -52,23 +52,39 @@ static void apply_horiz_layout(list_t *children, struct wlr_box *parent) {
 		child->width_fraction /= total_width_fraction;
 	}
 
+	// Calculate gap size
+	double inner_gap = 0;
+	struct sway_container *child = children->items[0];
+	struct sway_workspace *ws = child->workspace;
+	if (ws) {
+		inner_gap = ws->gaps_inner;
+	}
+	// Descendants of tabbed/stacked containers don't have gaps
+	struct sway_container *temp = child;
+	while (temp) {
+		enum sway_container_layout layout = container_parent_layout(temp);
+		if (layout == L_TABBED || layout == L_STACKED) {
+			inner_gap = 0;
+		}
+		temp = temp->parent;
+	}
+	double child_total_width = parent->width - inner_gap * (children->length - 1);
+
 	// Resize windows
 	sway_log(SWAY_DEBUG, "Arranging %p horizontally", parent);
 	double child_x = parent->x;
 	for (int i = 0; i < children->length; ++i) {
 		struct sway_container *child = children->items[i];
-		container_remove_gaps(child);
 		child->x = child_x;
 		child->y = parent->y;
-		child->width = floor(child->width_fraction * parent->width);
+		child->width = floor(child->width_fraction * child_total_width);
 		child->height = parent->height;
-		child_x += child->width;
+		child_x += child->width + inner_gap;
 
 		// Make last child use remaining width of parent
 		if (i == children->length - 1) {
 			child->width = parent->x + parent->width - child->x;
 		}
-		container_add_gaps(child);
 	}
 }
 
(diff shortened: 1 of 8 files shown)


sway/tree/arrange.c
```diff
@@ -58,38 +88,67 @@ static void apply_vert_layout(list_t *children, struct wlr_box *parent) {
...
 	// Resize
 	sway_log(SWAY_DEBUG, "Arranging %p vertically", parent);
+	double inner_gap = 0;
+	struct sway_container *child = children->items[0];
+	struct sway_workspace *ws = child->workspace;
+	if (ws) {
+		inner_gap = ws->gaps_inner;
+	}
+	double total_gap = MIN(MAX(0, parent->height - MIN_SANE_H * children->length),
+                         inner_gap * (children->length - 1));
```
RedSoxFan: style: indentation is off

sway/tree/arrange.c
```diff
@@ -12,44 +12,74 @@
...
 	// Resize windows
 	sway_log(SWAY_DEBUG, "Arranging %p horizontally", parent);
+	double inner_gap = 0;
+	struct sway_container *child = children->items[0];
+	struct sway_workspace *ws = child->workspace;
+	if (ws) {
+		inner_gap = ws->gaps_inner;
+	}
+	double total_gap = MIN(MAX(0, parent->width - MIN_SANE_W * children->length),
+                         inner_gap * (children->length - 1));
```
RedSoxFan: style: indentation is off

pedrocr:
The gaps should now match i3-gaps. Could you also test the floating containers issue? Not sure what that is. If it still happens how do I reproduce?

sway/tree/workspace.c
```diff
@@ -730,19 +691,39 @@ void workspace_add_gaps(struct sway_workspace *ws) {
...
-		// tabbed/stacked container is using gaps.
-		ws->current_gaps.top += ws->gaps_inner;
-		ws->current_gaps.right += ws->gaps_inner;
-		ws->current_gaps.bottom += ws->gaps_inner;
-		ws->current_gaps.left += ws->gaps_inner;
+	// Add inner gaps and make sure we don't turn out negative
+	ws->current_gaps.top = MAX(0, ws->current_gaps.top + ws->gaps_inner);
+	ws->current_gaps.right = MAX(0, ws->current_gaps.right + ws->gaps_inner);
+	ws->
+current_gaps.bottom = MAX(0, ws->current_gaps.bottom + ws->gaps_inner);
```
RedSoxFan: Can you fix this? The line splitting is weird, but it should actually fit on a single line anyway

sway/tree/workspace.c
```diff
@@ -730,19 +691,39 @@ void workspace_add_gaps(struct sway_workspace *ws) {
...
+	ws->current_gaps.left = MAX(0, ws->current_gaps.left + ws->gaps_inner);
+
+	// Now that we have the total gaps calculated we may need to clamp them in
+	// case they've made the available area too small
+	if (ws->width - ws->current_gaps.left - ws->current_gaps.right < MIN_SANE_W &&
+	    ws->current_gaps.left + ws->current_gaps.right > 0) {
```
RedSoxFan: There is mixed tabs and spaces here. Please use tabs and indent this line one additional tab to make it distinguishable from the body of the block.  Can you also move the `&&` to this line so the previous one is under 80 characters?

sway/tree/workspace.c
```diff
@@ -730,19 +691,39 @@ void workspace_add_gaps(struct sway_workspace *ws) {
...
+		int total_gap = MAX(0, ws->width - MIN_SANE_W);
+		double left_gap_frac = ((double)ws->current_gaps.left /
+			((double)ws->current_gaps.left + (double)ws->current_gaps.right));
+		ws->current_gaps.left = left_gap_frac * total_gap;
+		ws->current_gaps.right = total_gap - ws->current_gaps.left;
+	}
+	if (ws->height - ws->current_gaps.top - ws->current_gaps.bottom < MIN_SANE_H &&
+	    ws->current_gaps.top + ws->current_gaps.bottom > 0) {
```
RedSoxFan: Like above, there is mixed tabs and spaces here. Please use tabs and indent this line one additional tab to make it distinguishable from the body of the block.  Can you also move the `&&` to this line so the previous one is at the 80 character limit?

sway/tree/workspace.c
pedrocr: Yep, this was just an errant Enter that went into the text editor instead of the terminal somehow.

include/util.h
```diff
@@ -34,4 +34,9 @@ const char *sway_wl_output_subpixel_to_string(enum wl_output_subpixel subpixel);
 
 bool set_cloexec(int fd, bool cloexec);
 
+#undef MAX
+#define MAX(a, b)  (((a) > (b)) ? (a) : (b))
```
RedSoxFan: It looks like `fmin` and `fmax` are used throughout the code base so let's just use those instead

pedrocr: For some parts that can work but not all code is floating point where `fmin`/`fmax` applies.

pedrocr: The magic of automatic number type conversions actually makes the straight swap work fine. I don't really see a situation where this can break although I generally distrust these automatic conversions.

RedSoxFan:
Thanks!

PR #4311, opened 2019-07-07
Make mouse drag in tiled mode swap containers if no edge is selected

DarkDefender:
I've always had the idea that the highlighted middle area would swap the current container, so I've implemented a quick solution for this.

This will also make it so that all 5 highlightable areas have unique actions. Previously, I've run into cases where the middle area would just perform the same action as one of the others.

I'm guessing we could add a config option for this if anyone feels strongly about using the middle area for swapping.

Now, I've added some code to see if the action should trigger a swap action and then just copied the swap function from `sway/commands/swap.c`. I don't know how the right way to solve this would be. Should we move in the swap function from `swap.c` into a central place or make the swap function a public function?

This is my first patch to sway, so I'm sure this patch might be missing other things than just code de-duplication.

### DIFF
--- a/sway/commands/swap.c
+++ b/sway/commands/swap.c
@@ -84,8 +84,7 @@ static void swap_focus(struct sway_container *con1,
 	}
 }
 
-static void container_swap(struct sway_container *con1,
-		struct sway_container *con2) {
+void container_swap(struct sway_container *con1, struct sway_container *con2) {
 	if (!sway_assert(con1 && con2, "Cannot swap with nothing")) {
 		return;
 	}
--- a/sway/input/seatop_move_tiling.c
+++ b/sway/input/seatop_move_tiling.c
@@ -245,8 +245,12 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 		target_node->sway_workspace : target_node->sway_container->workspace;
 	enum wlr_edges edge = e->target_edge;
 	int after = edge != WLR_EDGE_TOP && edge != WLR_EDGE_LEFT;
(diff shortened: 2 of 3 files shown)


sway/input/seatop_move_tiling.c
```diff
@@ -245,22 +282,31 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 		target_node->sway_workspace : target_node->sway_container->workspace;
 	enum wlr_edges edge = e->target_edge;
 	int after = edge != WLR_EDGE_TOP && edge != WLR_EDGE_LEFT;
+	bool swap = false;
 
-	container_detach(con);
+	if (edge == WLR_EDGE_NONE && target_node->type == N_CONTAINER) {
```
RedSoxFan: nit: I think the following would be cleaner

```c
bool swap = edge == WLR_EDGE_NONE && target_node->type == N_CONTAINER;
if (!swap) {
    container_detach(con);
}
```

DarkDefender:
I've updated the patch with the given input.

sway/commands/swap.c
```diff
@@ -84,7 +84,7 @@ static void swap_focus(struct sway_container *con1,
 	}
 }
 
-static void container_swap(struct sway_container *con1,
+void container_swap(struct sway_container *con1,
```
RedSoxFan: nit: since it is now short enough to fit on a single line, can you please combine this and the following line?

DarkDefender: Done!

RedSoxFan:
Thanks!

PR #4312, opened 2019-07-09
cmd_bindswitch: add option to execute on reload

### LINKED ISSUE #4305: Output config reloads with 'swaymsg reload'
NVIDIA: no
version: 1.1-rc1-47-gb59139c2 (Jun 22 2019, branch 'master')
configuration file: https://gist.github.com/jasperro/aa237e0201885970f5cec45d513f11f5

When I run `swaymsg reload` and my laptop lid is closed, the laptop display is enabled again, which is not what I want. I don't know if this is a bug, or an intended feature.

I tried adding `exec_always cat /proc/acpi/button/lid/LID0/state | awk '{print $2}' | grep -q 'closed' && swaymsg output 'Unknown 0x1E3D 0x00000000' toggle &` to the end of my config file, which works, but crashes alacritty (terminal) and breaks right-click menus in firefox.

My proposal is to only load output configuration on first start and not reload it after that.

RedSoxFan:
Closes #4305

This adds a --reload flag to cmd_bindswitch that allows for the binding
to be executed on reload. One possible use case for this is to  allow
users to disable outputs when the lid closes and enable them when the
lid opens without having to open and re-close the lid after a reload.

### DIFF
--- a/include/sway/input/switch.h
+++ b/include/sway/input/switch.h
@@ -4,16 +4,20 @@
 #include "sway/input/seat.h"
 
 struct sway_switch {
-    struct sway_seat_device *seat_device;
+	struct sway_seat_device *seat_device;
+	enum wlr_switch_state state;
+	enum wlr_switch_type type;
 
-    struct wl_listener switch_toggle;
+	struct wl_listener switch_toggle;
 };
 
 struct sway_switch *sway_switch_create(struct sway_seat *seat,
-        struct sway_seat_device *device);
+		struct sway_seat_device *device);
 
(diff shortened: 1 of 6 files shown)


include/sway/input/switch.h
```diff
@@ -5,6 +5,8 @@
 
 struct sway_switch {
     struct sway_seat_device *seat_device;
```
emersion: Unrelated note: seems like this file is using mixed tabs and spaces

RedSoxFan: Not anymore. All tabs now

PR #4313, opened 2019-07-09
input/libinput: only reset supported options

RedSoxFan:
This adds checks to the `input_manager_libinput_reset_*` functions to
only attempt resetting supported options on reload. This should have no
functional difference to the user, but will remove several
`Failed to apply libinput config: Unsupported configuration option`
lines from the log that can be noisy and potential red herrings.


ddevault:
Thanks!

PR #4314, opened 2019-07-10
config/output: correctly set width/height in apply

### LINKED ISSUE #4198: dpms and font issues
* Sway Version:  1.0
* Debug Log: TODO
* Configuration File:  https://github.com/swalladge/dotfiles/blob/master/host-archdesk/config/sway/config

When dpms turns the monitor off (whether I do it manually, or swayidle calls it after swaylock is running), _and_ the monitors have completely turned themselves off, and then I dpms on/log back in, I often get font issues. 

Specifically, most text in x11 windows is missing. See screenshot of keepassxc:

screenshot of keepassxc

Additionally, fonts in wayland windows sometimes become fuzzy. Eg. this instance of termite:

screenshot of termite

Actually, I'm not sure if it's just fonts that become fuzzy, or the entire window - icons and such look fuzzy in firefox, but hard to tell because firefox has several related wayland issues anyway. I'll check some other gui apps.

I'll keep this issue updated as I do more debugging.

RedSoxFan:
Fixes #4198

In `apply_output_config`, this sets `output->{width,height}` using the
values in the output box. Previously, they were being set using
`wlr_output_transformed_resolution`, which takes the width and height
from the wlr_output and just checks whether they should be swapped
based on the transform. This did not take into account the output's
scale. `wlr_output_effective_resolution` could be used instead, which
handles both transform and scale. However, the values in the output box
have already been processed by `wlr_output_effective_resolution` so they
can just be used directly

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -308,9 +308,8 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		wlr_output_layout_get_box(root->output_layout, wlr_output);
 	output->lx = output_box->x;
 	output->ly = output_box->y;
-	wlr_output_transformed_resolution(wlr_output,
-		&output->width, &output->height);
-
+	output->width = output_box->width;
+	output->height = output_box->height;
 
 	if (oc && oc->dpms_state == DPMS_OFF) {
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4315, opened 2019-07-10
cmd_layout: toggle split for tabbed/stack default

### LINKED ISSUE #4129: from tabbed layout to split layout
Hi !

I use sway version 1.0

I use this config:

```
workspace_layout tabbed
bindsym button3 layout toggle tabbed split
bindsym $mod+w layout toggle tabbed split
```
if i press the key kombi mod+w or klick on the right mouse button over a titlebar it should toogle to split for all windows

It works but sometimes it not wok -> How is this possible ?

I can also provide a debug if nessacery !

Thanks !

kind regards

kristoferus75

RedSoxFan:
Fixes #4129

This fixes the logic of split for layout toggle when the default
layout is L_TABBED or L_STACKED. When the default layout is L_TABBED
or L_STACKED, the container/workspace may not have a prev_split_layout.
This was causing L_NONE to be returned by get_layout_toggle, which was
being handled as a syntax error. This adds logic to try
config->default_orientation when prev_split_layout is L_NONE. If that
is also L_NONE, then L_VERT is used when the output is taller than
wide, otherwise, L_HORIZ is used.

### DIFF
--- a/sway/commands/layout.c
+++ b/sway/commands/layout.c
@@ -2,6 +2,7 @@
 #include <string.h>
 #include <strings.h>
 #include "sway/commands.h"
+#include "sway/output.h"
 #include "sway/tree/arrange.h"
 #include "sway/tree/container.h"
 #include "sway/tree/workspace.h"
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4319, opened 2019-07-11
cmd_split: fix toggle split for non-split layouts

### LINKED ISSUE #4316: remember previous tile partition when exiting stacking mode
**Feature request:** Make `sway` remember the partitioning of tiles, so when you stack their parent, and then unstack it, you revert back to the original partition of the tiles as they were (so you don't spend time re-partitioning it again like they were before stacking).

This is already done in `i3`.

-----
**Details**

Just tried sway 1.1.1-3.  I noticed this behaviour:

1. I create a tile (mod+enter).  Let's call this tile _A_.
2. I create a tile underneath that tile. Let'scall this tile _B_.
3. I create a sibling tile to _B_, underneath _A_.  Let's call this sibling of _B_, _B_.

So, _A_ looks like a horizontal rectangle, underneath it two square windows (_B_ and _C_).

Now do this:

1.  Go to _A_.
2. Put int stacking mode (mod+s).

So, now you them stacked.  But what happens if you unstack it back to tiling again?  I.e. if you press mod+e on tile _A_?

What will happen is this:  all tiles, _A_, _B_, _C_ will be stacked vertically, and totally forget the previous hierarchical partitioning that we made earlier.

This does not happen in i3 as it remembers the previous configuration, so when I stack/unstack, things are neat and I don't need to rearrange stuff.

RedSoxFan:
Fixes #4316 

Previously, `layout toggle` and `layout toggle split` would set L_VERT
when layout was L_HORIZ, otherwise it would set L_HORIZ. This meant
that when the layout was L_TABBED or L_STACKED, it would always be
L_HORIZ. This extends #4315 (which corrects the handling when multiple
layouts are given) to try prev_split_layout,
config->default_orientation, and then falling back to L_VERT when the
output is taller than wide and L_HORIZ when wider than tall.

### DIFF
--- a/sway/commands/layout.c
+++ b/sway/commands/layout.c
@@ -26,19 +26,37 @@ static const char expected_syntax[] =
 	"'layout toggle [split|all]' or "
 	"'layout toggle [split|tabbed|stacking|splitv|splith] [split|tabbed|stacking|splitv|splith]...'";
 
+static enum sway_container_layout toggle_split_layout(
+		enum sway_container_layout layout,
+		enum sway_container_layout prev_split_layout,
+		struct sway_output *output) {
+	if (layout == L_HORIZ) {
+		return L_VERT;
+	} else if (layout == L_VERT) {
+		return L_HORIZ;
+	} else if (prev_split_layout != L_NONE) {
+		return prev_split_layout;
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

Issue #3999, opened 2019-03-31
Load custom xkb layout from file


InspectorMustache:
I use a custom keyboard layout defined in ~/.config/xkb/custom.xkb which I upload to X using `xkbcomp ~/.config/xkb/custom.xkb $DISPLAY`. Is there any way to do the same in Sway? I tried `input * xkb_layout ~/.config/xkb/custom.xkb` but that didn't work.

I guess I could register my layout globally and then just load it with `input * xkb_layout custom` but a) I'd like to do this from within my home folder and b) making a layout available like that is unexpectedly complicated; I edited both the .lst and .xml files in /usr/share/X11/xkb/rules and that didn't work. Those changes could also be gone with the next system update.

yory8:
Maybe it's not the same thing you mean, but I installed my own custom layout in `/usr/share/X11/xkb/symbols/`, with a custom name so it doesn't touch / isn't touched by anything else, and it works just fine. No need to register anything, just `input YOURKEYBOARD xkb_layout YOURNAME`

mdeff:
I did that as well: custom keyboard layout in `/usr/share/X11/xkb/symbols/custom`, then `input * xkb_layout custom`.

progandy:
Local user settings for keyboard layouts can be stored in ```~/.xkb/symbols/``` and then loaded by their names just like global layouts. ([source](https://github.com/xkbcommon/libxkbcommon/blob/110d17c6bee504569f68a105b2d4d06fe5e7e0b0/src/context.c#L89))

mdeff:
Confirmed to work! Thanks for the tip @progandy. I'll do that from now on.

InspectorMustache:
My apologies, I should have been more specific with my question. It's been a while since I set up my layout so I forgot about the xkb nomenclature. The thing is that my keyboard layout file not only contains an xkb_symbols section but also xkb_types and xkb_keycodes sections. This is all contained in a struct named xkb_keymap. This is the format returned by xkbcomp. I could probably get rid of the xkb_keycodes section but xkb_types is essential for me as that is where custom modifier levels are defined.

So when I try to load this file by doing `input [identifier] xkb_layout custom`, I get an error because only an xkb_symbols section is expected:
```
2019-04-04 21:40:33 - [sway-1.0/sway/input/seat.c:673] adding device 1133:16388:Logitech_K360 to seat seat0
[...]

PR #4325, opened 2019-07-13
Implement input_cmd_xkb_file (#3999)

edyounis:
Adds a new commend "xkb_file", which constructs the internal
xkb_keymap from a xkb file rather than an RMLVO configuration.
This allows greater flexibility when specifying xkb configurations.
An xkb file can be dumped with the xkbcomp program.

### DIFF
--- a/sway/commands/input/xkb_file.c
+++ b/sway/commands/input/xkb_file.c
@@ -1,0 +1,28 @@
+#define _POSIX_C_SOURCE 200809L
+#include "sway/config.h"
+#include "sway/commands.h"
+#include "sway/input/input-manager.h"
+#include "log.h"
+
+struct cmd_results *input_cmd_xkb_file(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "xkb_file", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE, "No input device defined.");
+	}
+
+	if (strcmp(argv[0], "-") == 0) {
+		free(ic->xkb_file);
+		ic->xkb_file = NULL;
+	} else {
+		ic->xkb_file = strdup(argv[0]);
+	}
+	ic->xkb_file_is_set = true;
+
+	sway_log(SWAY_DEBUG, "set-xkb_file for config: %s file: %s",
+			ic->identifier, ic->xkb_file);
+	return cmd_results_new(CMD_SUCCESS, NULL);
+}
--- a/sway/commands/input/xkb_file.c
+++ b/sway/commands/input/xkb_file.c
@@ -1,7 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include "sway/config.h"
 #include "sway/commands.h"
-#include "sway/input/input-manager.h"
 #include "log.h"
 
 struct cmd_results *input_cmd_xkb_file(int argc, char **argv) {
--- a/sway/commands/input/xkb_layout.c
+++ b/sway/commands/input/xkb_layout.c
@@ -1,7 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include "sway/config.h"
 #include "sway/commands.h"
-#include "sway/input/input-manager.h"
 #include "log.h"
 
 struct cmd_results *input_cmd_xkb_layout(int argc, char **argv) {
--- a/sway/config/input.c
+++ b/sway/config/input.c
@@ -39,6 +39,7 @@ struct input_config *new_input_config(const char* identifier) {
 	input->repeat_rate = INT_MIN;
 	input->xkb_numlock = INT_MIN;
 	input->xkb_capslock = INT_MIN;
+	input->xkb_file_is_set = false;
 
 	return input;
 }
@@ -94,6 +95,11 @@ void merge_input_config(struct input_config *dst, struct input_config *src) {
 	}
 	if (src->tap_button_map != INT_MIN) {
 		dst->tap_button_map = src->tap_button_map;
+	}
+	if (src->xkb_file_is_set) {
+		free(dst->xkb_file);
+		dst->xkb_file = src->xkb_file ? strdup(src->xkb_file) : NULL;
+		dst->xkb_file_is_set = dst->xkb_file != NULL;
 	}
 	if (src->xkb_layout) {
 		free(dst->xkb_layout);
@@ -284,6 +290,8 @@ struct input_config *store_input_config(struct input_config *ic,
 		ic = current;
 	}
 
+	ic->xkb_file_is_set = ic->xkb_file != NULL;
+
 	if (!current || new_current) {
 		list_add(config_list, ic);
 	}
@@ -307,6 +315,7 @@ void free_input_config(struct input_config *ic) {
 		return;
 	}
 	free(ic->identifier);
+	free(ic->xkb_file);
 	free(ic->xkb_layout);
 	free(ic->xkb_model);
 	free(ic->xkb_options);
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -534,11 +534,6 @@ static void handle_xkb_context_log(struct xkb_context *context,
 
 struct xkb_keymap *sway_keyboard_compile_keymap(struct input_config *ic,
 		char **error) {
-	struct xkb_rule_names rules = {0};
-	if (ic) {
-		input_config_fill_rule_names(ic, &rules);
-	}
-
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
 	if (!sway_assert(context, "cannot create XKB context")) {
 		return NULL;
@@ -546,8 +541,46 @@ struct xkb_keymap *sway_keyboard_compile_keymap(struct input_config *ic,
 	xkb_context_set_user_data(context, error);
 	xkb_context_set_log_fn(context, handle_xkb_context_log);
 
-	struct xkb_keymap *keymap =
-		xkb_keymap_new_from_names(context, &rules, XKB_KEYMAP_COMPILE_NO_FLAGS);
+	struct xkb_keymap *keymap = NULL;
+
+	if (ic && ic->xkb_file) {
+		FILE *keymap_file = fopen(ic->xkb_file, "r");
+		if (!keymap_file) {
+			if (error) {
+				size_t len = snprintf(NULL, 0, "cannot read XKB file %s: %s",
+						ic->xkb_file, strerror(errno)) + 1;
+				*error = malloc(len);
+				if (*error) {
+					snprintf(*error, len, "cannot read XKB file %s: %s",
+							ic->xkb_file, strerror(errno));
+				} else {
+					sway_log_errno(SWAY_ERROR, "cannot read XKB file %s: %s",
+							ic->xkb_file, strerror(errno));
+				}
+			} else {
+				sway_log_errno(SWAY_ERROR, "cannot read XKB file %s: %s",
+						ic->xkb_file, strerror(errno));
+			}
+			goto cleanup;
+		}
+
+		keymap = xkb_keymap_new_from_file(context, keymap_file,
+					XKB_KEYMAP_FORMAT_TEXT_V1, XKB_KEYMAP_COMPILE_NO_FLAGS);
+
+		if (!fclose(keymap_file)) {
+			sway_log_errno(SWAY_ERROR, "cannot close XKB file %s: %s",
+					ic->xkb_file, strerror(errno));
+		}
+	} else {
+		struct xkb_rule_names rules = {0};
+		if (ic) {
+			input_config_fill_rule_names(ic, &rules);
+		}
+		keymap = xkb_keymap_new_from_names(context, &rules,
+			XKB_KEYMAP_COMPILE_NO_FLAGS);
+	}
+
+cleanup:
 	xkb_context_set_user_data(context, NULL);
 	xkb_context_unref(context);
 	return keymap;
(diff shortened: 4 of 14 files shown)


sway/input/keyboard.c
```diff
@@ -534,18 +534,32 @@ static void handle_xkb_context_log(struct xkb_context *context,
...
-	}
-
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
 	if (!sway_assert(context, "cannot create XKB context")) {
 		return NULL;
 	}
 	xkb_context_set_user_data(context, error);
 	xkb_context_set_log_fn(context, handle_xkb_context_log);
 
+	if (ic->xkb_file) {
```
RedSoxFan: `ic` can be `NULL` so this should be `ic && ic->xkb_file`

sway/commands/input/xkb_file.c
```diff
@@ -0,0 +1,22 @@
...
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "xkb_file", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE, "No input device defined.");
+	}
+
+	ic->xkb_file = strdup(argv[0]);
```
RedSoxFan: There should be some way of resetting this to `NULL` so the rules can be used instead. We usually use `-` for that purpose

sway/input/keyboard.c
```diff
@@ -534,18 +534,32 @@ static void handle_xkb_context_log(struct xkb_context *context,
...
+		FILE *keymap_file = fopen(ic->xkb_file, "r");
+		if (!sway_assert(keymap_file, "cannot open XKB file")) {
```
RedSoxFan: I don't think someone specifying an invalid path should cause sway to abort. This should:
- Use `sway_log_errno` to log the error
- If `err` is not `NULL`, use `malloc` and `sprintf`to pass the error back to the user
- Return `NULL`

sway/input/keyboard.c
```diff
@@ -534,18 +534,32 @@ static void handle_xkb_context_log(struct xkb_context *context,
...
+			return NULL;
+		}
+		struct xkb_keymap *keymap =
+			xkb_keymap_new_from_file(context, keymap_file,
+				XKB_KEYMAP_FORMAT_TEXT_V1,
+				XKB_KEYMAP_COMPILE_NO_FLAGS);
+		xkb_context_set_user_data(context, NULL);
```
RedSoxFan: I think the following would be cleaner:
```c
struct xkb_keymap *keymap = NULL;
if (ic->xkb_file) {
    // Open file and error handling
    keymap = xkb_keymap_new_from_file(...);
    // Close file
} else {
    // Create rules
   keymap = xkb_keymap_new_from_names(...);
}

xkb_context_set_user_data(context, NULL);
xkb_context_unref(context);
return keymap;
```

sway/input/keyboard.c
```diff
@@ -534,18 +534,32 @@ static void handle_xkb_context_log(struct xkb_context *context,
...
-
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
 	if (!sway_assert(context, "cannot create XKB context")) {
 		return NULL;
 	}
 	xkb_context_set_user_data(context, error);
 	xkb_context_set_log_fn(context, handle_xkb_context_log);
 
+	if (ic->xkb_file) {
+		FILE *keymap_file = fopen(ic->xkb_file, "r");
```
RedSoxFan: `keymap_file` is never closed

sway/sway-input.5.scd
```diff
@@ -44,6 +44,11 @@ on top of the existing device configurations.
 For more information on these xkb configuration options, see
 *xkeyboard-config*(7).
 
+*input* <identifier> xkb_file <file_name>
+	Sets all xkb configurations from a complete .xkb file. This file can be
+	dumped from xkbcomp -xkb. This setting overrides xkb_layout, xkb_model,
```
RedSoxFan: `_xkbcomp -xkb_`

sway/commands/input/xkb_file.c
```diff
@@ -0,0 +1,22 @@
+#define _POSIX_C_SOURCE 200809L
+#include "sway/config.h"
+#include "sway/commands.h"
+#include "sway/input/input-manager.h"
```
RedSoxFan: Also, this include isn't needed

sway/input/keyboard.c
edyounis: "If err is not NULL, use malloc and sprintf to pass the error back to the user"

I am not sure what you are referring to here?

sway/commands/input/xkb_file.c
edyounis: The RMLVO commands, as well as this command, can only be run in a config file. This is the same exact format for a c file as the xkb_layout command. Is there still an issue here? Should all the RMLVO commands be changed in this way?

sway/commands/input/xkb_file.c
edyounis: Again, this file is almost an exact copy and paste from xkb_layout.c. Should these all be changed?

sway/commands/input/xkb_file.c
RedSoxFan: > The RMLVO commands, as well as this command, can only be run in a config file.

You can change them at runtime with `swaymsg` (or any IPC client). The only input subcommands that are config only are `xkb_capslock` and `xkb_numlock`.

> Should all the RMLVO commands be changed in this way?

No, they are fine as-is, but `xkb_file` shadows them when set so there needs to be a way to unset it.

edyounis: Sounds good

sway/commands/input/xkb_file.c
RedSoxFan: Yes, it's not needed in any of them. It's probably a remnant of a previous iteration.

sway/input/keyboard.c
RedSoxFan:
```c
if (err) {
	size_t len = snprintf(NULL, 0, "cannot read XKB file %s: %s",
			ic->xkb_file, strerror(errno)) + 1;
	*err = malloc(len);
	if (*err) {
		snprintf(*err, len, "cannot read XKB file %s: %s",
				ic->xkb_file, strerror(errno));
	}
}
```

This will allow for the error to be shown in swaynag or outputted by swaymsg to the user.

sway/commands/input/xkb_layout.c
```diff
@@ -1,7 +1,5 @@
 #define _POSIX_C_SOURCE 200809L
-#include "sway/config.h"
```
RedSoxFan: `sway/config.h` should be kept. It'll compile without it since `sway/commands.h` includes it, but it's better to be explicit.

sway/commands/input/xkb_file.c
```diff
@@ -0,0 +1,27 @@
...
+	if ((error = checkarg(argc, "xkb_file", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE, "No input device defined.");
+	}
+
+	if (strcmp(argv[0], "-") == 0) {
+		if (ic->xkb_file) {
```
RedSoxFan: `free(NULL)` is a no-op so the guard isn't needed

sway/input/keyboard.c
edyounis: Thanks for explaining this clearly. I haven't used errno's in c in a while, so excuse the possibly stupid questions. By `err` do you mean `error` from the function parameter? Should this be done in addition to sway_log_errno?

RedSoxFan: > By err do you mean error from the function parameter?

Oops sorry, yes had the name mixed up

> Should this be done in addition to sway_log_errno?

I guess it could be an else for the `if (*error)` in case the `malloc` fails. That way it wouldn't get double logged

edyounis: Should we do something similar for "cannot close xkb file"

RedSoxFan: Nah. Just logging that in the debug log is fine. It's not really useful from a user perspective. Failure to open could be a typo in the path or a permission issue which is much more useful to the user

sway/input/keyboard.c
```diff
@@ -534,20 +534,58 @@ static void handle_xkb_context_log(struct xkb_context *context,
...
+		if (!keymap_file) {
+			if (error) {
+				size_t len = snprintf(NULL, 0, "cannot read XKB file %s: %s",
+						ic->xkb_file, strerror(errno)) + 1;
+				*error = malloc(len);
+				if (*error) {
+					snprintf(*error, len, "cannot read XKB file %s: %s",
+							ic->xkb_file, strerror(errno));
+				} else {
+					sway_log_errno(SWAY_ERROR, "cannot read XKB file %s: %s",
```
RedSoxFan: This should also get logged if `error` is `NULL`

sway/input/keyboard.c
```diff
@@ -534,20 +534,58 @@ static void handle_xkb_context_log(struct xkb_context *context,
...
+							ic->xkb_file, strerror(errno));
+				}
+			}
+		}
+
+		keymap = xkb_keymap_new_from_file(context, keymap_file,
```
RedSoxFan: If `keymap_file` is `NULL`, this shouldn't continue, but does need to cleanup. The most readable option would be to add a `cleanup:` label before the `xkb_context_set_user_data` line and add a `goto cleanup` to the end of the `if` above.

edyounis: Do you think that if keymap_file is null that we should default to the previous setup with rules?

sway/input/keyboard.c
```diff
@@ -534,20 +534,64 @@ static void handle_xkb_context_log(struct xkb_context *context,
...
+				sway_log_errno(SWAY_ERROR, "cannot read XKB file %s: %s",
+						ic->xkb_file, strerror(errno));
+			}
+		}
+
+		keymap = xkb_keymap_new_from_file(context, keymap_file,
+					XKB_KEYMAP_FORMAT_TEXT_V1, XKB_KEYMAP_COMPILE_NO_FLAGS);
+
+		if (!fclose(keymap_file)) {
+			if (error) {
```
RedSoxFan: No need for `*error` here. Just use `sway_log_errno`.

sway/input/keyboard.c
edyounis: This would be cleaner
`if (!keymap_file) {
 //handle error
} else {
  keymap = xkb_keymap_new_from_file(...)
  //close file
}
if (!keymap) {
 // rule setup
}
//context and return`

RedSoxFan: No. One of the cases where this gets called is when storing the delta input config. If `NULL` is returned from this function, the delta input config is rejected for producing an invalid keymap. If this gets changed to return a valid keymap (from the rules), it'll result in the delta config being applied anyway and `*error` being leaked.

sway/commands/input/xkb_file.c
```diff
@@ -0,0 +1,26 @@
...
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "xkb_file", EXPECTED_EQUAL_TO, 1))) {
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE, "No input device defined.");
+	}
+
+	if (strcmp(argv[0], "-") == 0) {
```
RedSoxFan: Oops. Sorry didn't think this through. `ic->xkb_file` is always going to be `NULL` here since it's a delta input config so the `free` can be removed. Also, setting it to `NULL` means it won't get merged since there is no difference between not being set at all and being unset. There needs to be a bool or something denoting that this should be merged.

edyounis: I don't understand? I am not too far into the source of sway to understand what is meant by "delta input config?" Heads up, I am heading to bed right now, I will gladly continue working on this in the morning.

RedSoxFan: `config->handler_context.input_config` contains an empty input config. It is used as a delta input config to store the changes that should occur for the input subcommand. The whole storage process is pretty long to express in words. I'd recommend looking at the main input command handler and walking through the `store_input_config` function.

The issue here is that `merge_input_config` will see `src->xkb_file` as `NULL` in both the case that the subcommand does not touch `xkb_file` and the case when it wants to unset the `xkb_file`. When `src->xkb_file` is `NULL`, it doesn't merge it on to the destination input config because it views it as being untouched. There needs to be some way to differentiate between `untouched` and `set it to NULL`.

Adding a simple `bool xkb_file_is_set` to `struct input_config` should work. This could make the section in `merge_input_config` look something like:
```c
if (src->xkb_file_is_set) {
    free(dst->xkb_file);
    dst->xkb_file = src->xkb_file;
    dst->xkb_file_is_set = dst->xkb_file != NULL;
}
```
And then in `store_input_config`:
```c
ic->xkb_file_is_set = ic->xkb_file != NULL;

if (!current || new_current) {  // This line exists, just showing for placement reference
```

Clearing the flag on storage is probably necessary for input identifier configs to get properly merged with input type configs when the input device is originally hotplugged (the type for the identifier is unknown until the device appears).

Hopefully that helps. If you have more questions, let me know.

edyounis: I think I got this; I just pushed the relevant changes.

sway/config/input.c
```diff
@@ -95,6 +96,11 @@ void merge_input_config(struct input_config *dst, struct input_config *src) {
 	if (src->tap_button_map != INT_MIN) {
 		dst->tap_button_map = src->tap_button_map;
 	}
+	if (src->xkb_file_is_set) {
+	    free(dst->xkb_file);
```
RedSoxFan: The block has mixed tabs and spaces

RedSoxFan:
Thanks!

PR #4331, opened 2019-07-15
cmd_opacity: add relative opacity changes

jpeeler:
This enhances the opacity command to support relative assignment as well
as the currently implemented absolute assignment. The syntax is inspired
by the syntax used in pactl and looks like this in a sway config:

// relative change (new feature)
bindsym button4 opacity +.1
bindsym button5 opacity -.1

// absolute change (already implemented)
bindsym button4 opacity 1
bindsym button5 opacity .3

### DIFF
--- a/sway/commands/opacity.c
+++ b/sway/commands/opacity.c
@@ -1,21 +1,13 @@
 #include <assert.h>
 #include <stdlib.h>
+#include <strings.h>
 #include "sway/commands.h"
 #include "sway/tree/view.h"
 #include "log.h"
 
-static bool parse_opacity(const char *opacity, float *val) {
-	char *err;
-	*val = strtof(opacity, &err);
-	if (*val < 0 || *val > 1 || *err) {
-		return false;
-	}
-	return true;
-}
-
 struct cmd_results *cmd_opacity(int argc, char **argv) {
 	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "opacity", EXPECTED_EQUAL_TO, 1))) {
+	if ((error = checkarg(argc, "opacity", EXPECTED_AT_LEAST, 1))) {
 		return error;
 	}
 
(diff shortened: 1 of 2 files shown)


sway/commands/opacity.c
```diff
@@ -4,11 +4,23 @@
 #include "sway/tree/view.h"
 #include "log.h"
 
-static bool parse_opacity(const char *opacity, float *val) {
+static bool adjust_opacity(const char *opacity, float *alpha) {
 	char *err;
-	*val = strtof(opacity, &err);
-	if (*val < 0 || *val > 1 || *err) {
-		return false;
+	float val = 0.0f;
```
panaman67: Is this temp assignment really necessary? should be able to combine this and actual assignment below.

jpeeler: Good point, I'll fix this.

RedSoxFan:
The code looks fine. `sway.5.scd` needs to be updated though.

My only concern is the syntax. The rest of the commands use a keyword for relative changes. For example,
```
gaps inner|outer|<dir>|<edge> all|current set|plus|minus <amount>
output <name-or-id> transform <rotation> [clockwise|anticlockwise]
resize shrink|grow width|height [<amount> [px|ppt]]
```

I think it may be more consistent to do something like:
```
opacity [set] <fixed-amount>
opacity plus|minus <delta-amount>
```

@ddevault  @emersion: any opinions on this?

jpeeler:
I considered rewriting the gaps syntax to support the newly proposed syntax. However, I had been testing in the sway config and not on the CLI. Using the minus character "-" doesn't seem to parse so well, so it seems the "set|plus|minus" syntax is better. Updated to look like:
// relative change
bindsym button4 opacity plus .1
bindsym button5 opacity minus .1

// absolute change
bindsym button4 opacity set 1
bindsym button5 opacity set .

---
Should the old syntax work in addition to the above? Currently it doesn't and I'm not sure what is expected here.

RedSoxFan:
> Using the minus character "-" doesn't seem to parse so well

For reference, you have to quote the command or add `--` after all of the options. For example, the original syntax for this PR would have parsed with: `swaymg "opacity -0.5"` or `swaymsg -- opacity 0.5`

> Should the old syntax work in addition to the above? 

Yes, this should not be a breaking change. If there is only one argument, it should be treated as an implicit `set`

jpeeler:
Updated to support the old implicit set as well.

sway/commands/opacity.c
```diff
@@ -1,21 +1,48 @@
...
 #include "log.h"
 
-static bool parse_opacity(const char *opacity, float *val) {
+static bool adjust_opacity(int argc, char * const argv[], float *alpha) {
 	char *err;
-	*val = strtof(opacity, &err);
-	if (*val < 0 || *val > 1 || *err) {
+	float val;
+
+	if (argc == 1) {
```
RedSoxFan: I'm not a huge fan of the repetition in this function. Something like the following would be better:
```c
char *err;
float val = strtof(argc == 1 ? argv[0] : argv[1], &err);
if (*err) {
    return false;
}

if (strcasecmp(argv[0], "plus") == 0) {
    val = *alpha + val;
} else if (strcasecmp(argv[0], "minus") == 0) {
    val = *alpha - val;
} else if (argc > 1 && strcasecmp(argv[0], "set") != 0) {
    return false;
}

if (val < 0 || val > 1) {
    return false;
}

*alpha = val;
return true;
```

sway/sway.5.scd
```diff
@@ -650,7 +650,7 @@ The default colors are:
 	Any mark that starts with an underscore will not be drawn even if
 	*show_marks* is yes. The default is _yes_.
 
-*opacity* <value>
+*opacity* set|plus|minus <value>
```
RedSoxFan:
```
*opacity* [set|plus|minus] <value>
    Adjusts the opacity of the window between 0 (completely transparent) and
    1 (completely opaque). If the operation is omitted, _set_ will be used
```

jpeeler: Copied this verbatim.

sway/commands/opacity.c
jpeeler: Nearly copied this exactly.

RedSoxFan:
Thanks!

PR #4336, opened 2019-07-16
Make fullscreen check for fullscreen parents

### LINKED ISSUE #4332: Using split in fullscreen causes odd behavior
* Sway Version: 1.0
* Debug Log: https://pastebin.com/Ru6mH58w (relevant section, will upload full log if req'd)

Running `split {v,h}` while an app is in fullscreen starts to tile the fullscreen container. This also resizes the window when fullscreen is exited by pressing the fullscreen binding two times.

I'd be happy to try to fix this if somebody pointed me in the right general file region for the changes.

sauyon:
This ensures that a child of a fullscreen container cannot be fullscreened.

Fixes #4332.

### DIFF
--- a/sway/commands/fullscreen.c
+++ b/sway/commands/fullscreen.c
@@ -33,8 +33,15 @@ struct cmd_results *cmd_fullscreen(int argc, char **argv) {
 		}
 	}
 
-	bool is_fullscreen = container &&
-		container->fullscreen_mode != FULLSCREEN_NONE;
+	bool is_fullscreen = false;
+	for (struct sway_container *curr = container; curr; curr = curr->parent) {
+		if (curr && curr->fullscreen_mode != FULLSCREEN_NONE) {
+			container = curr;
+			is_fullscreen = true;
+			break;
+		}
+	}
+
 	bool global = false;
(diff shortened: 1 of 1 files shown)


pedrocr:
Shouldn't you be able to set the child fullscreen with the result being the same as going out of fullscreen in the parent and going into fullscreen in the child? Is the idea to be less confusing and whenever you've already entered a fullscreen state if you call fullscreen toggle on a child it's interpreted as leaving fullscreen on the parent? I can see both behaviors being desirable if you really want to fullscreen containers with tiled windows inside.

sauyon:
I think this is more sane default behavior, though I'm sure it breaks some people's workflows. Maybe an extra option should be added to `fullscreen`?

RedSoxFan:
I think the behavior in this PR makes more sense, but I also don't use fullscreen containers much.

@RyanDwyer since you were the one to implement the current behavior, any thoughts on this change?

RyanDwyer:
I think the behaviour in this PR makes more sense too. I'm happy for it to be changed.

ddevault:
Thanks!

PR #4337, opened 2019-07-16
Update output manager on layout change

atx:
The output manager config was not properly updated if the position
of the output got changed.


sway/desktop/output.c
```diff
@@ -688,6 +688,14 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 	update_output_manager_config(server);
 }
 
+void handle_output_layout_change(struct wl_listener *listener,
+		void *data)
+{
```
ddevault: style

PR #4342, opened 2019-07-17
ipc: add an input event

### LINKED ISSUE #4334: Feature request: Allow subscribing to xkb layout changes
Version: 1.0

I am trying to write a script that would display the selected keymap, and print a new line every time the layout is changed. In X, I use xkb-switch to accomplish this task, as it has a subscribe option.

I can read it fine according to docs; `swaymsg -t get_inputs | jq '.[] | select(."xkb_active_layout_name") | ."xkb_active_layout_name"` works fine. However, right now this can only be polled in a script, which is not desirable (time delay between change events) An event that can be subscribed to on active keyboard layout change (Something like `swaymsg -t subscribe '["xkb_active_layout_name"]' -w` would be useful.

Or, I currently use `xkb_options "grp:caps_toggle"`; using caps lock to cycle through the keymaps. I cannot bindsym to this key now however. 'bindsym Caps_Lock "swaymsg -t send_tick 'keymap_changed'"' does not send a tick.

RedSoxFan:
Closes #4334 

This adds an ipc event related to input devices. Currently the
following changes are supported:
- [x] added: when an input device becomes available
- [x] removed: when an input device is no longer available
- [x] xkb_keymap_changed: (keyboards only) the keymap changed
- [x] xkb_layout_changed: (keyboards only) the effective layout changed
- [x] libinput_config_changed: (libinput devices only) a libinput config option changed


RedSoxFan:
This is ready for review. I kept the `input::xkb_config_changed` event in a separate commit since it includes a major refactor of the libinput config management code.

sway/sway-ipc.7.scd
```diff
@@ -1702,6 +1705,58 @@ event is a single object with the following properties:
...
+:  (Keyboards only) The keymap for the keyboard has changed
+|- xkb_layout_changed
+:  (Keyboards only) The effective layout in the keymap has changed
+|- libinput_config_changed
+:  (libinput device only) A libinput config option for the device changed
+
+*Example Event:*
+```
+{
+	"change": "xkb_layout_changed",
```
emersion: It seems like i doesn't add a `_changed` suffix when properties change (see the `window` event for instance). For consistency this should probably just be the name of the property.

RedSoxFan: Changed

sway/sway-ipc.7.scd
```diff
@@ -1371,6 +1371,9 @@ available:
 |- 0x80000014
 :  bar_status_update
 :  Send when the visibility of a bar should change due to a modifier
+|- 0x80000015
```
emersion: Since i3 won't implement such an event, it may be wise to leave 0x80000015 for a future i3 event, and choose a much higher event code.

sway/input/keyboard.c
```diff
@@ -603,8 +608,23 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
 		}
 	}
 
+	bool keymap_changed = false;
+	bool effective_layout_changed = keyboard->effective_layout != 0;
```
emersion: Hmm, not sure I understand the logic here...

emersion: Hmm, nevermind. It's the effective layout that gets reset.

However if the keymap changed the effective layout changes too, even if `effective_layout` is still zero. The effective layout depends on the keymap.

sway/sway-ipc.7.scd
RedSoxFan: `bar_status_update` is also Sway specific. I just added this one right after. Should they both be changed?

sway/input/keyboard.c
RedSoxFan: Yes, but if the effective layout is 0 before and after, then either the entire keymap changed or the keymap is identical and there is no change to the effective layout. See #4343 for a follow-up for this

PR #4343, opened 2019-07-17
 input/keyboard: don't reset layout for same keymap 

RedSoxFan:
~**This is built on #4342 so that'll have to be merged first.**~ #4342 has been merged

In sway_keyboard_config, do not change the keymap when the new keymap
is unchanged, unless this is during a config reload. The reasoning for
this is to prevent the effective layout from being reset to index 0 for
input config changes unrelated to the keymap.

For example,
```sh
swaymsg "input type:keyboard xkb_layout 'us,us'"
swaymsg "input type:keyboard xkb_variant ',dvorak'"
swaymsg "input type:keyboard xkb_options caps:toggle"
# Switch to English (Dvorak) by pressing caps
swaymsg "input type:keyboard repeat_delay 1000"
# On master, the layout will be reset to `English (US)`
# With this PR, the layout remains `English (Dvorak)`
```

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -622,34 +622,42 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
 		keymap_changed = true;
 	}
 
-	xkb_keymap_unref(keyboard->keymap);
-	keyboard->keymap = keymap;
-	keyboard->effective_layout = 0;
-	wlr_keyboard_set_keymap(wlr_device->keyboard, keyboard->keymap);
+	if (keymap_changed || config->reloading) {
+		xkb_keymap_unref(keyboard->keymap);
+		keyboard->keymap = keymap;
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Rebased and ready for review

sway/input/keyboard.c
```diff
@@ -678,9 +685,6 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
 	if (keymap_changed) {
 		ipc_event_input("xkb_keymap",
 			keyboard->seat_device->input_device);
-	} else if (effective_layout_changed) {
-		ipc_event_input("xkb_layout",
```
emersion: This is not sent anymore. Is this deliberate?

RedSoxFan: I thought so, but it is actually still needed for config reloads. Re-added

PR #4348, opened 2019-07-17
cmd_output: support current output alias

### LINKED ISSUE #4345: Feature request: Transform current output
## What

Ability to transform the current screen without specifying the name explicitly. Perhaps by allowing the `<name>` to be ommited. 

## Why

This would make it easier to bindsym to simple changes like `transform 90 clockwise`.

## Background

I currently have the following in my config to handle screen rotation and moving workpaces between outputs:

```bash
    # Move workspace to other screens
    bindsym $mod+Shift+w exec $term_run "source ~/.fzf.zsh && swaymsg -t get_outputs | jq -r '.[] | select(.focused!=true) | [.name, .make, .model] | @tsv' | fzf --select-1 --reverse --prompt='Move workspace to:' | awk '{print $1}' | xargs -r swaymsg move workspace to"
    bindsym $mod+Alt+w exec swaymsg -t get_outputs | jq '.[] | select(.focused!=true) | .name' | head -n1 | xargs swaymsg move workspace to

    # Rotate current workspace
    bindsym $mod+Shift+r exec swaymsg -t get_outputs | jq '.[] | select(.focused==true) | .name' | xargs -I% swaymsg output % transform 90 clockwise
```

RedSoxFan:
Closes #4345 

Similar to seat command, this provides an alias for the current output.
Instead of the output name or identifier, `-` can be used to operate on
the focused output by name and `--` can be used to operate on the
focused output by its identifier. This will prevent operating on the
no-op output when using either alias.

### DIFF
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -36,7 +36,32 @@ struct cmd_results *cmd_output(int argc, char **argv) {
 				"Refusing to configure the no op output");
 	}
 
-	struct output_config *output = new_output_config(argv[0]);
+	struct output_config *output = NULL;
+	if (strcmp(argv[0], "-") == 0 || strcmp(argv[0], "--") == 0) {
+		if (config->reading) {
+			return cmd_results_new(CMD_FAILURE,
+					"Current output alias (%s) cannot be used in the config",
+					argv[0]);
+		}
+		struct sway_output *sway_output = config->handler_context.node ?
+			node_get_output(config->handler_context.node) : NULL;
+		if (!sway_output) {
+			return cmd_results_new(CMD_FAILURE, "Unknown output");
+		}
+		if (sway_output == root->noop_output) {
+			return cmd_results_new(CMD_FAILURE,
+					"Refusing to configure the no op output");
+		}
+		if (strcmp(argv[0], "-") == 0) {
+			output = new_output_config(sway_output->wlr_output->name);
+		} else {
+			char identifier[128];
+			output_get_identifier(identifier, 128, sway_output);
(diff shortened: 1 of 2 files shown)


sway/commands/output.c
```diff
@@ -36,7 +36,32 @@ struct cmd_results *cmd_output(int argc, char **argv) {
 				"Refusing to configure the no op output");
 	}
 
-	struct output_config *output = new_output_config(argv[0]);
+	struct output_config *output = NULL;
+	if (strcmp(argv[0], "-") == 0 || strcmp(argv[0], "--") == 0) {
```
ddevault: The right side of the `||` will always be true when the left side is true and can be omitted

ddevault:
I'm not sure I like string munging here. Can't we just get the current sway_output and pass it along?

sway/commands/output.c
ascent12: strcmp will take the null terminator into account, so it won't. That would be true with strncmp.

RedSoxFan:
> I'm not sure I like string munging here. Can't we just get the current sway_output and pass it along?

The output commands operate on an output config not a sway_output. Although most of the code here is just error handling

sway/commands/output.c
ddevault: derp

ddevault:
Alright, fair enough.

ddevault:
Thanks!

Jamedjo:
@RedSoxFan Thanks for this! Especially for such a fast turnaround :medal_sports:

PR #4351, opened 2019-07-19
bindsym/code: add group support

### LINKED ISSUE #3662: Add support for Group in bindsym/bindcode
This is orthogonal to https://github.com/swaywm/sway/pull/3058

It allows to enable a keybinding for a specific keyboard layout only.

RedSoxFan:
~**This is built on #4342 so that'll have to be merged first.**~ #4342 has been merged

Closes #3662 

This adds support for specifying a binding for a specific group. Any
binding without a group listed will be available in all groups. The
priority for matching bindings is as follows: input device, group, and
locked state.

For full compatibility with i3, this also adds Mode_switch as an alias
for Group2. Since i3 only supports this for backwards compatibility
with older versions of i3, it is implemented here, but not documented.

### DIFF
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -78,6 +78,10 @@ static bool binding_key_compare(struct sway_binding *binding_a,
 		return false;
 	}
 
+	if (binding_a->group != binding_b->group) {
+		return false;
+	}
+
 	if (binding_a->modifiers ^ binding_b->modifiers) {
 		return false;
 	}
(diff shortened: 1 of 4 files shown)


RedSoxFan:
Rebased and ready for review

sway/commands/bind.c
```diff
@@ -387,6 +392,34 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
...
+		if (strncmp(split->items[i], "Group", strlen("Group")) == 0) {
+			if (binding->group != XKB_LAYOUT_INVALID) {
+				free_sway_binding(binding);
+				list_free_items_and_destroy(split);
+				return cmd_results_new(CMD_FAILURE,
+						"Only one group can be specified");
+			}
+			char *end;
+			int group = strtol(split->items[i] + strlen("Group"), &end, 10);
+			if (group < 1 || group > 4) {
```
emersion: Could also check `end[0] != '\0' || str[0] == '\0'`

RedSoxFan: Added

Issue #2897, opened 2018-10-20
bindsym with multimedia key


megamann:
Hy.
**bindsym** with multimedia key not work properly.
My config (example one key):
`bindsym XF86HomePage exec apps-1`
`bindsym $mod+XF86HomePage exec apps-2`
`bindsym Mod1+XF86HomePage exec apps-3`
When press first or second or third combination **sway** always runs apps-1

SirCmpwn:
https://github.com/SirCmpwn/sway/wiki#how-do-i-report-issues

megamann:
`sway version 1.0-alpha.6-329-gc75098e9 (Oct 20 2018, branch 'master')`

[log.sway.txt](https://github.com/swaywm/sway/files/2498854/log.sway.txt)

adabru:
I have the same issue with XF86TouchpadOn(keycode 200), XF86TouchpadOff(201), XF86MonBrightnessDown(232), XF86MonBrightnessUp(233), XF86WebCam(220), XF86Launch7(194), etc. but not with XF86AudioRaiseVolume, XF86AudioLowerVolume, XF86AudioNext(171), XF86AudioPlay/XF86AudioPause(172), XF86AudioPrev(173), XF86AudioStop/XF86Eject(174). At least in my case all meida keys above a certain keycode threshold (~194) don't propagate modifiers. In case of e.g. XF86AudioPlay the keysym changes from the shiftkey, though.

@megamann you can try removing all bindsyms from your sway.conf (so they are not absorbed) and running `xev` to see which media keys propagate modifiers and which ones don't. I ended up using the media keys that work with modifiers, violating the function that is printed on them.

I looked a little bit at the `/usr/share/X11/symbols/*` files but couldn't find a clue there (I looked superficially though).

stsquad:
I seem to have a similar problem (sway version 1.0-beta.2) with the XF86AudioPlay not making it to Emacs. Running xev I see it as:

    KeymapNotify event, serial 30, synthetic NO, window 0x0,
    keys:  4294967199 0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
           0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0

I have the raise and lower volume keys captured by sway but other "media" keys do arrive, for example XF86Mail:

    KeyPress event, serial 32, synthetic NO, window 0x1c00001,
    root 0x39f, subw 0x1c00002, time 11035962, (33,43), root:(2584,541),
[...]

PR #4373, opened 2019-07-26
Add support for wlr_keyboard_group

### LINKED ISSUE #3847: Some keyboard button combinations do not work
With sway 1.0-rc4: 
On a swedish keyboard layout I have configured sway as such:

```code
input * {
	xkb_layout se
	xkb_model pc105
}
```
On this keyboard, the symbols "less than", "greater than" and the "pipe" character are all available on the "102nd" key (scan code KEY_102ND). "<" is produced by simply pressing the button. "greater than" is produced by holding shift and pressing the button. The pipe character is produced by holding AltGr and pressing the button. This is how the button works in most cases, but when I try doing this in sway it doesn't work. Every button combination produces "<" and no other character.

Similarly, the "backslash" key (scan code KEY_BACKSLASH) produces " ' " when pressed, and is supposed to produce an asterisk character when holding shift and pressing button. But every button combination simply produces the ' character instead.

I do not know if there are other inaccessible characters, but these symbols are all pretty important when working in terminal emulators. I do not know if this problem is local to sway, wlroots, or another library dependency.

RedSoxFan:
**Depends on wlroots PR swaywm/wlroots#1771**

Related swaywm/wlroots#1217
Related #2897
Closes #3847

A wlr_keyboard_group allows for multiple keyboard devices to be
combined into one logical keyboard. This is useful for keyboards that
are split into multiple input devices despite appearing as one physical
keyboard in the user's mind.

This adds support for wlr_keyboard_groups to sway. There are two
keyboard groupings currently supported, which can be set on a per-seat
basis. The first keyboard grouping is none, which disables all grouping
and provides no functional change. The second is keymap, which groups
the keyboard devices in the seat by their keymap. With this grouping,
the effective layout and repeat info is also synced across keyboard
devices in the seat. Device specific bindings will still be executed as
normal, but everything else related to key and modifier events will be
handled by the keyboard group's keyboard.

Future improvements:
- Handle keyboards attached to multiple seats

### DIFF
--- a/sway/sway-input.5.scd
+++ b/sway/sway-input.5.scd
@@ -218,6 +218,16 @@ correct seat.
 	disables hiding the cursor. The minimal timeout is 100 and any value less
 	than that (aside from 0), will be increased to 100.
 
+*seat* <name> keyboard_grouping none|keymap
+	Set how the keyboards in the seat are grouped together. Currently, there
+	are two options. _none_ will disable all keyboard grouping. This will make
+	it so each keyboard device has its own isolated state. _keymap_ will
+	group the keyboards in the seat by their keymap. This is useful for when
+	the keyboard appears as multiple separate input devices. In this mode,
+	the effective layout and repeat info are also synced between the keyboards
+	in the group. The default is _keymap_. To restore the behavior of older
+	versions of sway, use _none_.
+
 *seat* <name> pointer_constraint enable|disable|escape
 	Enables or disables the ability for clients to capture the cursor (enabled
 	by default) for the seat. This is primarily useful for video games. The
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -281,6 +281,7 @@ sway_cmd seat_cmd_attach;
 sway_cmd seat_cmd_cursor;
 sway_cmd seat_cmd_fallback;
 sway_cmd seat_cmd_hide_cursor;
+sway_cmd seat_cmd_keyboard_grouping;
 sway_cmd seat_cmd_pointer_constraint;
 sway_cmd seat_cmd_xcursor_theme;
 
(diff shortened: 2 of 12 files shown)


sway/sway-input.5.scd
```diff
@@ -214,6 +214,16 @@ correct seat.
...
 	than that (aside from 0), will be increased to 100.
 
+*seat* <name> keyboard_grouping none|keymap
+	Set how the keyboards in the seat are grouped together. Currently, there
+	are two options. _none_ will disable all keyboard grouping. This will make
+	it so each keyboard device has it's own isolated state. _keymap_ will
+	group the keyboards in the seat by their keymap. This is useful for when
+	the keyboard appears as multiple separate input devices. In this mode,
+	the effective layout and repeat info are also synced between the keyboards
+	in the group. The default is _keymap_. The restore the behavior of older
```
Emantor: Typo: "To Resotre the behviour of older…"

RedSoxFan: Fixed

sway/sway-input.5.scd
```diff
@@ -214,6 +214,16 @@ correct seat.
 	disables hiding the cursor. The minimal timeout is 100 and any value less
 	than that (aside from 0), will be increased to 100.
 
+*seat* <name> keyboard_grouping none|keymap
+	Set how the keyboards in the seat are grouped together. Currently, there
+	are two options. _none_ will disable all keyboard grouping. This will make
+	it so each keyboard device has it's own isolated state. _keymap_ will
```
cyberhuman: one more typo, it must be `its`

RedSoxFan: Fixed

ddevault:
Thanks!

PR #4377, opened 2019-07-26
Avoid adding duplicate criteria for no_focus and command

### LINKED ISSUE #4376: Don't add `no_focus` with duplicate criteria
If someone executes `no_focus $CRITERIA` with the same `$CRITERIA` multiple times, then, from looking at the source which does a simple `list_add`, it will add a new criteria handler even though it's a duplicate.

I suggest we search through the list for matching criteria first before adding it. I would've liked something like this for the `for_window` variant of criteria commands, but those are slightly more complex since they can have any command. However, you could do the same thing there if they are *exact* matches.

norcalli:
Fixes #4376

### DIFF
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -40,6 +40,9 @@ struct criteria {
 };
 
 bool criteria_is_empty(struct criteria *criteria);
+bool criteria_is_equal(struct criteria *left, struct criteria *right);
+
+bool criteria_already_exists(struct criteria *criteria);
 
 void criteria_destroy(struct criteria *criteria);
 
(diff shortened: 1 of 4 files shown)


norcalli:
Obviously this is a naive version, and ordering of criteria would cause duplicates to be formed, so I can improve this as needed, but I honestly didn't know how to compare `pcre` instances, so I went for the easy solution for now.

RyanDwyer:
Yeah, I'm happy with your implementation. Scripts aren't likely to change their order of criteria tokens in each invocation so it wouldn't be worth the effort to handle that in sway.

PR #4384, opened 2019-07-27
Remove all wayland-server.h includes

emersion:
The documentation for wayland-server.h says:

> Use of this header file is discouraged. Prefer including
> wayland-server-core.h instead, which does not include the server protocol
> header and as such only defines the library PI, excluding the deprecated API
> below.

Replacing wayland-server.h with wayland-server-core.h allows us to drop the
WL_HIDE_DEPRECATED declaration.

This commit si similar to wlroots' ca45f4490ccc ("Remove all wayland-server.h
includes").

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -2,7 +2,7 @@
 #define _SWAY_OUTPUT_H
 #include <time.h>
 #include <unistd.h>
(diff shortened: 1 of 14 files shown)


ddevault:
Thanks!

PR #4388, opened 2019-07-28
Fix resize sibling amount calculations

### LINKED ISSUE #4386: Resize broken for tiling windows
After updating to master after a few weeks, I noticed that my resize command was completely broken for tiled windows (floating were exempt). Upon running `resize set width 500px height 500px`, the window would not resize to `500px`. It resized to something smaller (432px), then after a second rerun, 489px, and after a final run 490px (which is the correct size accounting for my 5px window border).

Using this script, I ran a git bisect and found the offending commit.

I use `st` because it launches very quickly, so you can use any application for this purpose if you want to recreate the output. It:

- launches sway
- waits 2 seconds for sway to launch
- gets the swaysock (newest sway socket)
- launches two windows to create a tiled situation where resizing can take place
- resizes, gets the width, checks that against the expected result (`490 = 500 - 5 * 2`) 

The first good commit I used was `2bccb387d81298ffd3b88a193deb3e32c0b6c494`.

```fish
#!/usr/bin/env fish

ninja -C build


sed 's/^exec/# &/' ~/.config/sway/config > /tmp/swayconfig

build/sway/sway -d -c /tmp/swayconfig &

sleep 2
set -x SWAYSOCK (ls -t1 /var/run/user/1000/sway-ipc.1000.* | head -n1)

swaymsg -- exec st
swaymsg -- exec st
sleep 0.2
swaymsg -- resize set width 500px height 500px
sleep 0.1
set WIDTH (swaymsg -t get_tree | jq '.. | select(.focused?) | .window_rect.width')
echo width = $WIDTH
swaymsg exit
wait
if test $WIDTH -eq 490
	echo "SUCCESS"
	exit 0
else
	echo "FAILURE"
	exit 1
end
```

The result was that `e3a3917d3afb66fc8ba3eebb7aed603d3b7ce844` was the first commit which broke this.

pedrocr:
Sibling amounts for resizing were being calculated after the original fraction had been altered. This led to broken resize amounts.

Fixes #4386

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -160,7 +160,7 @@ void container_resize_tiled(struct sway_container *con,
 	}
 
 	// Apply new dimensions
-	int sibling_amount = prev ? amount / 2 : amount;
+	int sibling_amount = prev ? ceil((double)amount / 2.0) : amount;
 
 	if (is_horizontal(axis)) {
 		if (con->width + amount < MIN_SANE_W) {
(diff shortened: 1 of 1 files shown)


norcalli:
I can confirm that this passes my bisection script, and so it seems like it's ok.

RedSoxFan:
Thanks!

PR #4392, opened 2019-07-28
Fix numerical issues in resize

### LINKED ISSUE #4391: resize has numerical instability in tiled containers
* Sway Version: sway version 1.1-rc1-82-gcefc608c (Jul 28 2019, branch 'master')
* Debug Log: https://gist.github.com/pedrocr/b4b9ed288b906540fb5d23ed69d8bb78
* Configuration File: default

Resizing a window with a fixed number of pixels has off-by-one issues. Asking to resize the window to `500px` will sometimes result in 499 or 501 pixel window. Doing it repeatedly converges on the correct value.

pedrocr:
Because the layout code rounds down the dimensions of the windows resizing would often be off by one pixel. The width/height fraction would not exactly reflect the final computed width and so the resize code would end up calculating things wrong.
    
To fix this recompute the width/height fractions after rounding down and adjust the resize amount by 0.1 pixels to always survive the rounding down correctly when we end up on the fence.

Also adjust the gaps calculation to not end up with fractions of a pixel in the case where the gaps were sized down for lack of space.

Fixes #4391

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -172,9 +172,19 @@ void container_resize_tiled(struct sway_container *con,
 		if (prev && prev->width - sibling_amount < MIN_SANE_W) {
 			return;
 		}
+		if (con->child_total_width <= 0) {
+			return;
+		}
 
-		double amount_fraction =
-			((double)amount / con->width) * con->width_fraction;
+		// We're going to resize so snap all the width fractions to full pixels
+		// to avoid rounding issues
+		list_t *siblings = container_get_siblings(con);
+		for (int i = 0; i < siblings->length; ++i) {
+			struct sway_container *con = siblings->items[i];
+			con->width_fraction = con->width / con->child_total_width;
+		}
+
+		double amount_fraction = (double)amount / con->child_total_width;
 		double sibling_amount_fraction =
 			prev ? amount_fraction / 2.0 : amount_fraction;
 
@@ -193,9 +203,19 @@ void container_resize_tiled(struct sway_container *con,
 		if (prev && prev->height - sibling_amount < MIN_SANE_H) {
 			return;
 		}
+		if (con->child_total_height <= 0) {
+			return;
+		}
 
-		double amount_fraction =
-			((double)amount / con->height) * con->height_fraction;
+		// We're going to resize so snap all the height fractions to full pixels
+		// to avoid rounding issues
+		list_t *siblings = container_get_siblings(con);
+		for (int i = 0; i < siblings->length; ++i) {
+			struct sway_container *con = siblings->items[i];
+			con->height_fraction = con->height / con->child_total_height;
+		}
+
+		double amount_fraction = (double)amount / con->child_total_height;
 		double sibling_amount_fraction =
 			prev ? amount_fraction / 2.0 : amount_fraction;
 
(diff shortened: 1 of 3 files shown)


pedrocr:
I'm not quite sure I like the current code. Manipulating the width and height fractions to reflect the pixel rounding will then result in the container dimensions not roundtripping if the tiled container is resized to and from a different size. I think I have a slightly different solution that doesn't suffer from that.

sway/commands/resize.c
```diff
@@ -173,8 +173,10 @@ void container_resize_tiled(struct sway_container *con,
 			return;
 		}
 
+		// Use an extra 0.1 pixels for the calculation to avoid rounding down
+		// by the layout code eating a pixel
 		double amount_fraction =
-			((double)amount / con->width) * con->width_fraction;
+			(((double)amount + 0.1) / con->width) * con->width_fraction;
```
emersion: Would be nice to find a better way to fix this…

pedrocr: Using `round()` instead of `floor()` in the layout code would eliminate the need for that but would make it possible in extreme cases for the children to end up larger than the parent.

Depending on how correct the layout code should be things could get even more complex. For example right now the last container always gets whatever leftover space there is from doing `floor()` to all the other widths. That means that if you resize that last container to a fixed px size it won't be correctly sized.

If that kind of pixel correctness is not wanted then maybe PR isn't needed at all.

pedrocr:
This is a slightly more complex way of doing things but that keeps the width/height fractions intact when we're not resizing so that when the parent container gets resized the sizes will roundtrip correctly. When resizing everything gets snapped to the pixel grid to create a sane starting point.

Using `round()` instead of `floor()` to get rid of the `+0.1` and avoid some more corner cases should be fine in almost all situations but does change how the last container is sized in pathological cases where the containers are much too small anyway.

pedrocr:
Any reason not to merge this now?

pedrocr:
Is there a specific objection to this PR? 1.2 has ended up shipping with broken layout code which kind of bums me out.

ddevault:
I think the problem with these layout patches is that they're risky patches which solve an unclear problem. This is some important code and changes to it will probably have unforseen consequences. Can you please provide specific reproduction steps for the problem which this patch solves?

pedrocr:
The PR solves the specific bug linked, it includes a description and a log.

I know these patches need reviewing to make sure issues are not introduced but then maybe 1.2 should have been released with the rest of the resize/layout stuff reverted. In the absence of a test suite having this go through master and a release test cycle seems like the only option unless there's a specific objection.

pedrocr:
I've been using this plus #4393 locally for months now with no issues. Meanwhile sway 1.2 was released and now 1.3 is close and this bug is still unpatched that was an after effect of fixing the other layout bugs. Could this at least be moved into master after 1.3 to see if anyone finds any issues with it?

ddevault:
Okay, let's move it into master and work on getting it into 1.3 (it's a bugfix, so it's still eligible to ship with 1.3). Sorry for overlooking this.

pedrocr:
Thanks for merging. I'm happy to look into any layout/resize/gaps issues that show up from this or even in general.

PR #4393, opened 2019-07-28
Make all the container dimensions integers

pedrocr:
Containers are always fixed to the pixel grid so position and size them with integers instead of doubles.

Functionally this should be no different since rounding down is already being done on things like layout. But it makes it clear what the intention is and avoids bugs where fractional pixels are used. The translating and moving code is still using doubles because the cursors can have fractional pixels and thus the code is plumbed that way. But that could also probably be changed easily by doing the integer
conversions earlier and plumbing with int.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -41,8 +41,8 @@ enum wlr_direction;
 struct sway_container_state {
 	// Container properties
 	enum sway_container_layout layout;
-	double x, y;
-	double width, height;
+	int x, y;
+	int width, height;
 
 	enum sway_fullscreen_mode fullscreen_mode;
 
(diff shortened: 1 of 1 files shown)


ddevault:
Does this fix a specific problem?

pedrocr:
No, I just came across it while fixing other stuff.

ddevault:
I feel like this is going to have unforseen consequences.

pedrocr:
While that may be true those consequences would be bugs as containers are not supposed to be located on fractional locations. So finding those would not be bad.

RyanDwyer:
I considered doing this at one point, but never got around to it. You'll want to test fractional scaling as well as resizing floating containers using the mouse while holding shift.

pedrocr:
I've tried a 1.3 fractional scale and a few resizes with no issues. I've been running the code for a week now also without issues so merging to master to get more testing might be feasible.

Another option would be to first add `-Wfloat-conversion` or even `-Wconversion` to the build and fix the cases that show up. Only then changing the dimensions code. From a quick test there may not be that many cases of implicit conversions today so it might even be a prudent way to go.

ddevault:
Thanks!

PR #4394, opened 2019-07-28
input/libinput: fix typo in set_middle_emulation

### LINKED ISSUE #4387: Touchpad buttons have reversed
## What

Using the left touchpad button has randomly started causing a right click and vice versa. 

## When

I'm not sure of the cause and will close this issue if I can't reproduce it after a restart. It didn't occur previously so could have started after updating sway, but I think it was working when I first started sway this session.

My current commit is formed by merging origin/layer-shell-popups into master at e4bba906b68b9347c085a7833f6dbc7ddb1a41b4, so it could also be to do with that. Previously I was on 36aa67e5 (master from 17th July) with wlroots at df3f0ffb (master June 28th).

RedSoxFan:
Fixes #4387 

This fixes a typo in set_middle_emulation where it would set left
handed instead of middle emulation. _Apparently, I didn't do a 
good enough job double checking my refactor in #4342. Just
looked it over again and this looks like the only issue..._

### DIFF
--- a/sway/input/libinput.c
+++ b/sway/input/libinput.c
@@ -127,7 +127,7 @@ static bool set_middle_emulation(struct libinput_device *dev,
 		return false;
 	}
 	sway_log(SWAY_DEBUG, "middle_emulation_set_enabled(%d)", mid);
-	log_status(libinput_device_config_left_handed_set(dev, mid));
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4408, opened 2019-08-02
Allow for empty criteria

RPigott:
Some folks on the irc were wondering if there was a cleaner way to use for_window on all windows than [app_id=".*"] and I couldn't think of any. This little change seems like it works to allow empty criteria, [], which match all windows and might be useful to some. Allowing empty criteria doesn't seem so bad to me. I feel like I'm missing something -- anyone know why this check is here in the first place?

Just a thought.

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -668,11 +668,6 @@ struct criteria *criteria_parse(char *raw, char **error_arg) {
 		goto cleanup;
 	}
 
-	if (criteria_is_empty(criteria)) {
-		*error_arg = strdup("Criteria is empty");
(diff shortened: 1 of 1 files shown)


ddevault:
Does i3 have something for this purpose?

RPigott:
Not as far as I know. Maybe that's reason enough not to allow this.

ddevault:
Would prefer to pitch it to them first.

RPigott:
Makes sense. I'll see about that then.

ddevault:
Please re-open this once some progress is made on that front. Thanks!

PR #4416, opened 2019-08-04
cmd_mode: make modes case sensitive

RedSoxFan:
This mirrors a change in i3 4.17 that makes binding modes case
sensitive


ddevault:
...why on Earth would they do this :/

Thanks.

RedSoxFan:
> ...why on Earth would they do this :/

https://github.com/i3/i3/issues/3587

PR #4418, opened 2019-08-04
swaymsg: return 2 for sway errors

RedSoxFan:
This mirrors a change in i3 4.17 that returns 2 for errors from sway,
including invalid command, command failed, and invalid subscription
requests

https://github.com/i3/i3/pull/3614


ddevault:
Thanks!

PR #4424, opened 2019-08-05
input/cursor: do not hide when buttons are pressed

### LINKED ISSUE #4420: Feature request: Don't hide cursor when holding down buttons
The current behavior for `hide_cursor` is quite buggy for multiple applications. Some examples:
- Firefox: Hold down left click on a text input with text, wait for the cursor to disapear, release left mouse button, now your cursor moves the selection as if you were still holding the left mouse button.
- Minecraft: break blocks in a line without releasing the left button and without moving the mouse. When the cursor 'disapears', you can no longer move the view and you have to release the left mouse button

Preventing the cursor from hiding when any mouse button is held would prevent those weird behaviors.

My config is available here https://github.com/cafehaine/dotfiles/blob/master/.config/sway/config

RedSoxFan:
Closes #4420

This just adds a small quality of life improvement to the cursor hiding
functionality. The cursor will no longer be hidden unless all buttons
are released.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -186,7 +186,12 @@ static int hide_notify(void *data) {
 	return 1;
 }
 
-int cursor_get_timeout(struct sway_cursor *cursor){
+int cursor_get_timeout(struct sway_cursor *cursor) {
+	if (cursor->pressed_button_count > 0) {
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4425, opened 2019-08-05
cmd_swap: add floating support

RedSoxFan:
For feature parity with i3 4.17, this just adds floating container
support to the swap command


ddevault:
Thank you!

PR #4426, opened 2019-08-06
sway.5: explain how to enable pango markup in font

### LINKED ISSUE #4411: Document more explicit howto enable pango rendering of window titles
Little nitpick here

In the manpage: https://github.com/swaywm/sway/blob/8ee054b1b95b7466c0dd89bfc9026f4083fb0016/sway/sway.5.scd#L562 
We should maybe document that the default value for font *is*: monospace 10 and to turn on pango rendering: "pango:Monospace 10" should be used.

Took me some time to figure out that starting the string with pango was the magic trick :)

RedSoxFan:
Closes #4411

This clarifies the syntax to use for the font command to enable pango
markup support.

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -559,8 +559,11 @@ The default colors are:
 	to the opposite edge of the container, even if there are other containers
 	in the direction. Default is _yes_.
 
-*font* <font>
-	Sets font for use in title bars in Pango format.
+*font* [pango:]<font>
+	Sets font to use for the title bars. To enable support for pango markup,
+	preface the font name with _pango:_. For example, _monospace 10_ is the
+	default font. To enable support for pango markup, _pango:monospace 10_
+	should be used instead.
 
 *titlebar_border_thickness* <thickness>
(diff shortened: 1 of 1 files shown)


ddevault:
Wait a minute... since when do we use anything other than Pango for text rendering?

Emantor:
Sway doesn't, `pango:` decides whether the markup is enabled or not, which gets passed into the various pango functions.

RedSoxFan:
Pango is always used. Pango markup is not always interpreted.

https://github.com/swaywm/sway/blob/19d05d3ae0eb2545986c275341250845c84f0f30/sway/commands/font.c#L16-L22

https://github.com/swaywm/sway/blob/19d05d3ae0eb2545986c275341250845c84f0f30/sway/tree/container.c#L499-L500

https://github.com/swaywm/sway/blob/19d05d3ae0eb2545986c275341250845c84f0f30/common/pango.c#L57-L69

ddevault:
Ah, I see.

ddevault:
Thanks!

PR #4428, opened 2019-08-06
container_replace: copy {width,height}_fraction

### LINKED ISSUE #4422: Repeated splith/v increase window size
* Sway Version: `1.1-rc1-91-g90d8a4df`
* Debug Log: https://gist.github.com/YaLTeR/551d975e487493af87ec02d10989a95b
* Configuration File: default with terminal set to `alacritty`

1. Spawn two windows in a || layout (default).
2. Press `splith` a bunch of times.
3. The window size increases.

Same can be done with `splitv` when the initial layout is vertical instead of horizontal.

RedSoxFan:
Fixes #4422

This copies the width and height fractions from the container to the
container replacing it. Without setting these values, the container
is treated as a new container and throws off the existing sizing. Since
one container is replacing the other, it makes sense for the sizing to
remain the same.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1324,8 +1324,12 @@ void container_replace(struct sway_container *container,
 		root_scratchpad_remove_container(container);
 	}
 	if (container->parent || container->workspace) {
+		float width_fraction = container->width_fraction;
+		float height_fraction = container->height_fraction;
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Ping?

ddevault:
Thanks!

PR #4429, opened 2019-08-06
config/output: rebase cursors after config applied

### LINKED ISSUE #4410: Rotating output doesnt update mouse image
When rotating the output with `swaymsg output ... transform 180 `  the screen rotates but the mouse pointer stays at the same not rotated location. Only after moving it again jumps the to rotated position.

RedSoxFan:
Fixes #4410

When applying an output config, an output may transform or be altered
in some way that effects the cursor. In order for the cursor images to
be updated properly, all cursors need to be rebased after applying
output configs.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -5,9 +5,11 @@
 #include <sys/socket.h>
 #include <sys/wait.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4440, opened 2019-08-10
Patches

MisterDA:
Hello,
This is a bunch of patches for bugs reported by [PVS-Studio](https://www.viva64.com/en/pvs-studio/).

It mostly removes some dead code and checks, and fixes some memory leaks.

PVS-Studio also reports a whole lot of calls to malloc and the family, or strdup calls that are not checked for null results.
Another warning reported are expressions of the form `double x = y / 2` where `y` is of type int, that could be written as `double x = y / 2.` to avoid the loss of a fractional part. This is not fixed in this PR, let me know if it is more-or-less needed.
*Edit: this was fixed in 1372d61: Avoid loss of fractional part*
 

Thanks for the good work!

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -709,7 +709,7 @@ void view_update_size(struct sway_view *view, int width, int height) {
 		con->surface_y = fmax(con->surface_y, con->content_y);
 	}
 	con->surface_width = width;
-	con->surface_width = height;
+	con->surface_height = height;
 }
 
 static const struct sway_view_child_impl subsurface_impl;
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -117,7 +117,6 @@ struct ipc_response *ipc_recv_response(int socketfd) {
 	return response;
 error_2:
 	free(response);
-	free(payload);
 error_1:
 	sway_log(SWAY_ERROR, "Unable to allocate memory for IPC response");
 	return NULL;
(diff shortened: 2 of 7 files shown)


sway/tree/view.c
```diff
@@ -709,7 +709,7 @@ void view_update_size(struct sway_view *view, int width, int height) {
 		con->surface_y = fmax(con->surface_y, con->content_y);
 	}
 	con->surface_width = width;
-	con->surface_width = height;
+	con->surface_height = height;
```
MisterDA: Please double-check this. It looks like a copy-mistake to me, but it's possible that I don’t know what I'm doing.

MisterDA:
Hi,
I’ve changed the commit messages with your suggestions.
I also added a commit that divides integers by `2.` (double) when storing the result in a double, to keep a fractional part, although I don’t know if that’s a good idea.

progandy:
> I also added a commit that divides integers by `2.` (double) when storing the result in a double, to keep a fractional part, although I don’t know if that’s a good idea.

There is another pull request that converts the variables to integers #4393

MisterDA:
> > I also added a commit that divides integers by `2.` (double) when storing the result in a double, to keep a fractional part, although I don’t know if that’s a good idea.
> 
> There is another pull request that converts the variables to integers #4393

Ok, so 1372d61 is unneeded. I’m removing it.

sway/tree/view.c
ddevault: I agree that this is a copy/paste mistake.

ddevault:
Thanks!

PR #4445, opened 2019-08-11
layer_shell: Guard against negative exclusive zone

dos1:
This can happen with surfaces that set negative margins.


ddevault:
Thanks!

PR #4447, opened 2019-08-12
workspace: prefer identifiers for output priority

### LINKED ISSUE #4432: Workspaces shuffle after idle on multi-output setups
Workspaces often shuffle when displays wake from idle

# How to reproduce

1. Run a setup with multiple outputs that have swayidle dpms off'ing displays
2. Wait for displays to be disabled, and subsequently hardware suspend
3. Wake the system up
4. Observe workspace locations

# Expected result

Workspaces should be located on the outputs that they were on previously

# Actual result

Workspaces are often shuffled, so that whatever was on output 1 is now on output 2 and vice versa.

While this may seem like a minor issue, it can in fact be extremely frustrating.

RedSoxFan:
Fixes #4432 

Since output names can change in various configurations, including
DisplayPort MST, prefer output identifiers for the output priority.
Users can still use `workspace <ws> output <names-or-ids>`, but any
output that is programmatically added to the list will be added under
the output identifier. If the output name exists in the list (from the
user workspace output configs), then that will be retained instead of
switching to the output identifier for that output.

### DIFF
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -499,19 +499,28 @@ static int find_output(const void *id1, const void *id2) {
 	return strcmp(id1, id2);
 }
 
+static int workspace_output_get_priority(struct sway_workspace *ws,
+		struct sway_output *output) {
+	char identifier[128];
+	output_get_identifier(identifier, sizeof(identifier), output);
+	int index_id = list_seq_find(ws->output_priority, find_output, identifier);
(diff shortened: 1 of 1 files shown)


RedSoxFan:
@kennylevinsen can you please confirm that this fixes the issue for you?

kennylevinsen:
It appears to do the trick—workspaces end up in their correct spaces even after output order has been shuffled.

ddevault:
Thanks!

PR #4446, opened 2019-08-12
handle_seat_node_destroy: do not focus own node

RedSoxFan:
Fixes #4444 

In handle_seat_node_destroy, it was possible to focus the node attached
to the seat node that is being destroyed when an empty workspace was
being destroyed in a multiple seat environment. This resulted in
infinite recursion when attempting to destroy the workspace. This just
moves the seat node destruction higher so it cannot be the focus
inactive for the seat. This is the same ordering that is applied to
destruction of seat nodes for containers

--

On a related note, since seats have their own focus, should the workspace
not actually be destroyed until all seats have switched focus away (even if
the workspace is hidden) instead of forcing all seats to vacate the empty
workspace?

PR #4448, opened 2019-08-13
workspace: do not destroy if any seat is focusing

RedSoxFan:
Follow-Up for #4446 

Since each seat has its own focus, do not destroy a workspace until it
is no longer focused by any seat. This prevents seats from being forced
to evacuate the workspace just because another seat switched focus away
from it


ddevault:
Thanks!

PR #4449, opened 2019-08-13
swaymsg.1: add tip about two hyphens for commands

### LINKED ISSUE #4443: swaymsg:getopt can't parse negative integers, failing 'calibration_matrix'
When rotating the screen, I would like for its touchscreen to be rotated too. Trying out the  command with `swaymsg` fails due to its use of `getopt` interpreting a negative value (e.g. `-1`) as a command option.

**Steps to reproduce**
```bash
$ swaymsg input '0:0:Elan_Touchscreen' calibration_matrix '-1 0 1 0 -1 1'
```
```
swaymsg: invalid option -- '1'
Usage: swaymsg [options] [message]

  -h, --help             Show help message and quit.
  -m, --monitor          Monitor until killed (-t SUBSCRIBE only)
  -p, --pretty           Use pretty output even when not using a tty
  -q, --quiet            Be quiet.
  -r, --raw              Use raw output even if using a tty
  -s, --socket <socket>  Use the specified socket.
  -t, --type <type>      Specify the message type.
  -v, --version          Show the version number and quit.
```
The code fails at that point [(link)](https://github.com/swaywm/sway/blob/a8a239e2f1ca6236c2b8f7cebf0cc37890e77ec0/swaymsg/main.c#L362)
```c
c = getopt_long(argc, argv, "hmpqrs:t:v", long_options, &option_index);
```
This is only an issue when `-1` is the first negative number. For example, `'0 -1 1 1 1 0 0'` is fine 

**Sway version**
```
swaymsg -t get_version
sway version 1.1-rc1-99-g724926ea (Aug  8 2019, branch 'master')
```

**Workarounds**
* Adding `--`to signal that no more options will follow [(solution source)](https://www.thecodingforums.com/threads/getopt-with-negative-numbers.540466/)
```bash
$ swaymsg -- input '0:0:Elan_Touchscreen' calibration_matrix '-1 0 1 0 -1 1'
```
* Add a leading space to the argument string
```bash
# Space added before `-1`
$ swaymsg input '0:0:Elan_Touchscreen' calibration_matrix ' -1 0 1 0 -1 1'
```

**Suggested fix**
Clarify the man page of swaymsg to suggest passing `--` when writing plain sway commands to avoid parsing issues.  
In hindsight I should have guessed better, unfortunately it's only after having installed Eclipse and figuring out how to compile and debug the code that I understood what was the problem.

This is my first contribution to an open source project, Sway is amazing, thanks for what you're doing!

RedSoxFan:
Closes #4443

This adds a tip to the command section about using two hyphens before a
command to signal that no swaymsg options will follow to allow for sway
commands with leading hyphens.

### DIFF
--- a/swaymsg/swaymsg.1.scd
+++ b/swaymsg/swaymsg.1.scd
@@ -46,6 +46,11 @@ _swaymsg_ [options...] [message]
 
 	See **sway**(5) for a list of commands.
 
+	Tip: If you are proving a command that contains a leading hyphen (_-_),
+	insert two hyphens (_--_) before the command to signal to swaymsg not to
+	parse anything beyond that point as an option. For example, use
+	_swaymsg -- mark --add test_ instead of _swaymsg mark --add test_
+
 *get\_workspaces*
 	Gets a JSON-encoded list of workspaces and their status.
 


ddevault:
Fuck you, GNU.

Thanks @RedSoxFan!

progandy:
An alternative would be to use the POSIX-like mode of getopt_long I think.

```man 3 getopt_long```
> If  the first character of optstring is '+' or the environment variable POSIXLY_CORRECT is set, then option  processing  stops as soon as a nonoption argument is encountered.

ddevault:
That was discussed on IRC and ultimately decided against.

PR #4452, opened 2019-08-15
input/keyboard: send released only if pressed sent

### LINKED ISSUE #3063: Key events of sway bindings are eaten on keypress but not key release
Using `sway version 1.0-beta.1-138-gd19648a3 (Nov  4 2018, branch 'master')`

When executing some keybinding while a client is focused, this client does not receive the key press event of the last key, as expected, but it will receive the release event.

For example, if I execute the `super`+`e` key binding, the underlying client receives the following events:

- key `super` pressed
- key `e` release
- key `super` released

meaning the client received a release event for a key it never received a pressed event for. Depending on the concerned keys, this may corrupt their xkbcommon state.

This can be seen with any client dumping its keyboard events, such as the keyboard exemple from [SCTK](https://github.com/smithay/client-toolkit).

### LINKED ISSUE #3196: Moving between containers will change the timeline of a video on Udemy
sway version 1.0-beta.1-239-gaf7a97623bf1 (Nov 26 2018, branch 'master')

When I'm watching videos on udemy using firefox and I swtich to another containter using for example the shortcut SUPER+3 to code something then I switch back to my browser container using, for example, the shortcut SUPER+2 the video on firefox will move to another time on the timeline. Moving back to the container using SUPER+2 will have the same behavior as when I'm watching the video and press the key 2 at any time.

RedSoxFan:
Fixes #3063
Fixes #3196

This keeps track of whether surfaces received a key press event and
will only send a key release event if the pressed event was sent. This
also requires changing the keycodes that are sent via wl_keyboard_enter
to only include those that were previously sent. This makes it so
surfaces do not receive key release events for keys that they never
received a key press for and makes it so switching focus doesn't leak
keycodes that were consumed by bindings.

### DIFF
--- a/include/sway/input/keyboard.h
+++ b/include/sway/input/keyboard.h
@@ -60,6 +60,7 @@ struct sway_keyboard {
 	struct sway_shortcut_state state_keysyms_translated;
 	struct sway_shortcut_state state_keysyms_raw;
 	struct sway_shortcut_state state_keycodes;
+	struct sway_shortcut_state state_pressed_sent;
 	struct sway_binding *held_binding;
 
 	struct wl_event_source *key_repeat_source;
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks! A little hacky but it'll do.

PR #4453, opened 2019-08-15
cmd_xwayland: add force for immediate launch

### LINKED ISSUE #4359: Sway hangs for a few seconds before first window opens
version: 1.0

I installed sway on two laptops, and one of them has the problem.  I was trying to nail down the cause but didn't come up with anything.

What I did was to launch sway, wait for 10 seconds, and then launch termite.  There would be a 2-8 seconds lag between the command and response.  After that, further commands work without any lags.

[Gist](https://gist.github.com/leafpile/99b40033b1742fea7b96040bbd36f645)

(Line 1141 was when I launched termite)

RedSoxFan:
Closes #4359

This just adds a force option to cmd_xwayland that allows for xwayland
to be immediately launched instead of lazily launched. This is useful
for slower machines so it can be part of the startup time instead of
when the user is actively trying to use it

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -424,6 +424,12 @@ enum alignment {
 	ALIGN_RIGHT
 };
 
+enum xwayland_mode {
+	XWAYLAND_MODE_DISABLED,
+	XWAYLAND_MODE_LAZY,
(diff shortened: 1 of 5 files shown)


ddevault:
Thanks!

PR #4460, opened 2019-08-17
Optionally show all workspace buttons in the bar, even the ones on other outputs

pedrocr:
Add a bar command that makes it so all workspaces are shown, even the ones that do not belong to the bar's output. I find this useful when disabling the bar on external screens and yet wanting to figure out what workspaces have been created in those outputs. Waybar includes a similar option.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -222,6 +222,7 @@ sway_cmd bar_cmd_unbindcode;
 sway_cmd bar_cmd_unbindsym;
 sway_cmd bar_cmd_wrap_scroll;
 sway_cmd bar_cmd_workspace_buttons;
+sway_cmd bar_cmd_workspace_buttons_all_outputs;
 
 sway_cmd bar_colors_cmd_active_workspace;
 sway_cmd bar_colors_cmd_background;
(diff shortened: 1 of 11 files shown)


ddevault:
Have you discussed this with i3?

pedrocr:
No, I've never used i3. Are all new features supposed to be shared?

pedrocr:
Feature request submitted to i3 here:

https://github.com/i3/i3/issues/3773

pedrocr:
This was closed without a reason. i3 considered this in the past but didn't want it. This together with the other PR just means that swaybar is replicating the i3 bar which is stuck at a state with missing features for historical reasons. I guess the message from sway is "we've replicated i3, warts and all, and now if you want something else swap the bar"?

ddevault:
This isn't a wart, it's just a feature you want. The goal of sway is not to implement every idea anyone has.

emersion:
>stuck at a state with missing features for historical reasons

It's not really historical reasons. They explained why they don't want to support that.

pedrocr:
The wart is what the other PR shows. Since it's not possible with the i3bar protocol to left align something it's not possible to implement both these PRs as external scripts. If someone was designing i3 today there's no reason for mode and workspace to be builtin and everything else to be a script.

PR #4463, opened 2019-08-17
Optionally add focused window title to the bar

pedrocr:
When using sway without title bars it's useful to be able to display window titles in the bar.  Add a "window_title yes|no" bar command that displays the title of the focused window on the bar by reacting to window events.

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -136,3 +136,33 @@ void pango_printf(cairo_t *cairo, const char *font,
 	g_object_unref(layout);
 	free(buf);
 }
+
+void pango_printf_ellipsized(cairo_t *cairo, const char *font,
+		double scale, bool markup, int max_width, const char *fmt, ...) {
+	va_list args;
+	va_start(args, fmt);
+	// Add one since vsnprintf excludes null terminator.
+	int length = vsnprintf(NULL, 0, fmt, args) + 1;
+	va_end(args);
+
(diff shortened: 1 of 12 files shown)


pedrocr:
Feature request submitted to i3 here:

https://github.com/i3/i3/issues/3774

ddevault:
No thanks. I agree with @airblader's arguments against this for inclusion in i3.

pedrocr:
This is disappointing. Note that the argument for i3 was that this could be done with the i3bar protocol and then it turned out it couldn't.

emersion:
It's not possible because you want to align the title. This is a separate issue.

pedrocr:
So a PR to extend the i3bar protocol with an alignment property would be acceptable?

emersion:
Ref https://github.com/i3/i3/issues/1148

The question is whether or not the added complexity is acceptable, especially regarding clickable blocks.

pedrocr:
That's why I'm asking before hacking on things. I'm leaning towards just keeping a local fork.

pedrocr:
Note that clickable blocks are already part of the protocol:

https://i3wm.org/docs/i3bar-protocol.html

All that's really needed to replicate the current bar generically is a property that says if modules should be placed on the left or right of the bar. That doesn't seem too hard to do really.

PR #4465, opened 2019-08-18
render: set surface as sampled for presentation

### LINKED ISSUE #4451: Presentation time is reported to wrong surface commits
* Sway Version: `1.1-rc1-99-g724926ea (Aug  7 2019, branch 'makepkg')`

As far as I can understand the logic between sway and wlroots, `presentation-time` is delivered to the last surface commit at the time of the presentation. However, that's not entirely correct: it should be delivered to the surface commit that was displayed in the presentation, so in sway that would be something along the last surface commit at the time of `output_render`.

In particular, the current behavior results in timestamps of up to one frame off for surfaces which render their contents on the `frame` callback. To see why, look at this diagram (that I recorded from `rootston`, but as far as I can tell the relevant parts work the same way in sway):

timestamps

Red lines are presentation timestamps (yes, they are not aligned with the other lines, I noted that in https://github.com/swaywm/wlroots/issues/655#issuecomment-511098029, but this isn't relevant here), the top half of the graph depicts the time from the page flip to `output_render` start and end (green and blue dots) when the surface contents are sampled, with the next `frame` callback send right afterwards. At the bottom half of the graph you can see the life-cycle of surface contents, being rendered and committed on `frame`, represented by the triangle (shortly after the blue dot because `frame` is currently sent right after `output_render`), then sampled during the next `output_render` (represented by the circle) and finally presented on screen during the presentation after that.

In this case, the last commit before a presentation will be presented only on the next presentation, so effectively the reported timestamp is one frame earlier than it should be.

YaLTeR:
Fixes #4451, requires https://github.com/swaywm/wlroots/pull/1798.

There's one issue: it looks like the very first drawn frame frequently gets a discarded presentation feedback. I can't quite figure out why. Is there any other path where a surface can get rendered? I tried adding a similar call to `render_saved_view` but that didn't help.

Update: might actually not be an issue, but just how new tiled views are handled and how the timings align.

Update 2: I recorded the screen with my phone and it seems that the discarded/presented notifications correspond exactly to frames shown on screen, so yeah, looks like no issue with this PR.

Update 3: Actually, sometimes it does seem to deliver discarded events to presented frames, most notably when resizing a floating window with mouse fast enough.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -120,6 +120,8 @@ static void render_surface_iterator(struct sway_output *output,
 		wlr_output->transform_matrix);
 
 	render_texture(wlr_output, output_damage, texture, &box, matrix, alpha);
+
+	wlr_presentation_surface_sampled(server.presentation, surface);
 }
 
 static void render_layer(struct sway_output *output,
@@ -267,6 +269,10 @@ static void render_saved_view(struct sway_view *view,
 
 	render_texture(wlr_output, damage, view->saved_buffer->texture,
 			&box, matrix, alpha);
+
+	// FIXME: we should set the surface that this saved buffer originates from
+	// as sampled here.
+	// https://github.com/swaywm/sway/pull/4465#discussion_r321082059
 }
 
 /**


YaLTeR:
Seems like updating in `render_saved_view` solves "Update 3" so I added a respective commit.

sway/desktop/render.c
```diff
@@ -267,6 +269,10 @@ static void render_saved_view(struct sway_view *view,
 
 	render_texture(wlr_output, damage, view->saved_buffer->texture,
 			&box, matrix, alpha);
+
+	if (view->surface) {
+		wlr_presentation_surface_sampled(server.presentation, view->surface);
```
emersion: Hmm. I don't think this is correct, because it displays a previous version of the view. So sending presented for this surface makes the client think its last update has been presented,

YaLTeR: Where's the correct place to mark the surface as sampled in this case? Probably somewhere where the buffer is saved?

YaLTeR: Actually I guess in this case the saved buffer needs to store the presentation feedback object associated with the commit from which the buffer was saved, and then either discard it if it's overridden with another saved buffer, or signal it as sampled in this spot.

YaLTeR: I tried approaching this problem by adding a third "saved" state to the presentation feedback. The idea was that a feedback can become "saved" and then saved feedbacks could be either sampled or discarded (in addition to directly sampling feedbacks). I would then mark the feedback as "saved" in `view_save_buffer`, sample the "saved" feedback in `render_saved_view` and discard the "saved" feedback in `view_remove_saved_buffer`. However, I could not get it to work, because:
- when the surface is dragged around as a floating window, `view_save_buffer` and `view_remove_saved_buffer` are called multiple times on the same buffer, which leads to the feedback being discarded in my model;
- when the surface is first created and committed I observed that it's possible for `view_save_buffer` to be called before the commit handler of the feedback, which means that the feedback that should have been "saved" by `view_save_buffer` wasn't "saved" because it wasn't "committed" yet at that point in time.

I'm not sure how else to approach this problem in a way which is reasonable. Any help would be appreciated.

YaLTeR:
I rebased and left a FIXME in `render_saved_view()`. This patch works well enough without that implemented (the only issue I can see is not sending presented events during resizing), and implementing that properly seems to require a refactor of the wlroots presentation time handling and further thought, and can be done in a subsequent patch.

emersion:
Can you fill a follow-up issue for saved views?

YaLTeR:
Yeah, I was just planning to do that: https://github.com/swaywm/sway/issues/4572

Issue #3690, opened 2019-02-15
Remove xdg-shell-unstable-v6 support


emersion:
Probably not anytime soon. This is just an issue to track various toolkits/clients progress.

- [x] EFL ([since 1.21.0](https://git.enlightenment.org/core/efl.git/commit/src/lib/ecore_wl2/ecore_wl2_display.c?id=ffa276fce3b812372a063431903fa0569d71fcb2))
- [x] GLFW ([since 3.3](https://github.com/glfw/glfw/pull/1152), doesn't support v6, only wl-shell and xdg-shell stable)
- [x] Gstreamer ([since 1.16](https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/commit/67bb17e4fae76f8f662fdd69bba6cdaaf37a90fa))
- [x] GTK3 ([since 3.22.30](https://gitlab.gnome.org/GNOME/gtk/commit/2e7ac262be15fb402c4103d28533c6c6d955bf65))
- [x] Qt (since 5.12)
- [x] mpv ([since 0.29.0](https://github.com/mpv-player/mpv/commit/76211609e3c589dafe3ef9a36cacc06e8f56de09), doesn't support v6, only xdg-shell stable)
- [ ] Ozone ([only supports v5 and v6](https://chromium.googlesource.com/chromium/src/+/master/ui/ozone/platform/wayland/host/wayland_connection.cc), but Wayland support is experimental anyway)
- [x] SDL2 ([since 2.0.8](https://hg.libsdl.org/SDL/rev/bc702c5a7c3a))
- [x] Weston ([since 6.0.0](https://gitlab.freedesktop.org/wayland/weston/merge_requests/103))
- [x] winit

nielsdg:
Just a heads up that the next GStreamer release will have support for XDG shell stable in at least [glimagesink](https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/merge_requests/66) and [vaapisink](https://gitlab.freedesktop.org/gstreamer/gstreamer-vaapi/merge_requests/39)

nielsdg:
Annnd if [this](https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/merge_requests/225) gets merged in time, GStreamer 1.16 will also have support for XDG shell in `waylandsink`

emersion:
Seems like Gstreamer 1.16 has been released with xdg-shell stable support.

ddevault:
[...]

PR #4466, opened 2019-08-18
Remove xdg-shell v6 support

emersion:
All major toolkits and apps have gained xdg-shell stable support.

Closes: https://github.com/swaywm/sway/issues/3690


ddevault:
Thanks! It's about time :)

Issue #4407, opened 2019-08-01
layer_shell does not obtain keyboard focus in certain corner cases


kennylevinsen:
Experienced using https://github.com/kennylevinsen/wldash at commit 9ac29b2a94e406bfa6442bb549b773b4a073ef19, which creates a top layer shell surface that is keyboard interactive.

# To reproduce

Note that this issue is finicky, and depends on what applications are currently running and have focus.

1. Start sway with the following configuration:
```
bar {
    position bottom
    mode hide
    modifier $mod
    ...
}

$mod+t exec wldash
```
2. Start a terminal emulator, such as alacritty
3. Launch wldash with $mod+t
4. Press escape

# Expected result

wldash receives escape and terminates

# Actual result

wldash does not have keyboard focus until focus is changed, causing sway to acknowledge wldash's keyboard_interactive top layer_shell surface.

This seems to be due to interaction with swaybar, as removing the bar resolves the issue.

ddevault:
Can you test this again now that #4467 has been merged?

kennylevinsen:
Verified fixed. Thanks!

PR #4467, opened 2019-08-18
A layer-shell will only be focused if it is non-null

### LINKED ISSUE #4461: Focus fighting with status bars
I'm currently developing a layer-shell based program that uses keyboard input and if it's open on one monitor and there is a status bar open on a different monitor then it focus fights with that status bar. If both are on the same monitor then it's not an issue. This is most noticeable if you have status bars on more than one monitor as it will focus fight with one of them no matter what. This happens with both swaybar and rootbar.

EDIT:
This issue might actually be a duplicate of #4407

asdfjkluiop:
This should fix issue #4407 and definitely fixes #4461

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -212,7 +212,9 @@ void arrange_layers(struct sway_output *output) {
 
 	struct sway_seat *seat;
 	wl_list_for_each(seat, &server.input->seats, link) {
-		seat_set_focus_layer(seat, topmost ? topmost->layer_surface : NULL);
+		if (topmost != NULL) {
+			seat_set_focus_layer(seat, topmost->layer_surface);
+		}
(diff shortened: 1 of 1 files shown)


sway/desktop/layer_shell.c
```diff
@@ -212,7 +212,9 @@ void arrange_layers(struct sway_output *output) {
 
 	struct sway_seat *seat;
 	wl_list_for_each(seat, &server.input->seats, link) {
-		seat_set_focus_layer(seat, topmost ? topmost->layer_surface : NULL);
+		if(topmost != NULL) {
```
ddevault: style nit: need a space between if and (

ddevault:
Please squash these commits together.

ddevault:
Thanks!

PR #4471, opened 2019-08-19
Allow moving a container hidden in scratchpad

lbonn:
i3 allows this and it seemed easy to support here as well.


lbonn:
Thanks, that was quite an oversight.

I'll try to think of something.

lbonn:
Ok here is a second version, I've opted to base the positions on the currently focused workspace. I think it's a reasonable fallback.

RedSoxFan:
Thanks!

PR #4474, opened 2019-08-20
cmd_move: fix move to scratchpad hidden container

### LINKED ISSUE #4473: Segfault moving window from scratchpad to mark in focused workspace during for_window
This one is a bit convoluted, but I've gotten the repro as small as I can. I recently wrote a script to automatically arrange qutebrowser windows on workspaces with other QB containers present. I do this by sending all QB windows to the scratchpad, then execing the script within the same for_window command. It uses python-i3ipc to look at the scratchpad, find similar windows on by comparing the app_id, mark one of the windows, and move the new window out of the scrachpad to the mark. This works fine, until I reload sway. After the reload, the next container to try to leave the scratchpad causes a segfault.

I got this down to a two-line sway config which will crash immediately after start. It defines a for_window, then runs a termite which sleeps a second, reloads, and execs another termite (matching the `app_id=__focused__` criteria). I tried to eliminate the external script entirely but it was making it more complicated to follow rather than less.

It seems to be important that the marked container also matches the criteria of the for_window. For example, matching app_id=qutebrowser and running it from the termite does not cause the crash. Additionally, the `sleep 1` is necessary to trigger the crash.

* Sway Version:
sway version cf92f2a (Aug 17 2019, branch '1.2-rc1-patch')

(The -patch in the branch and the hash are from building inside an arch package repo, the "patch" is to the PKGBUILD to build the rc + debug symbols. Sway itself is at the 1.2-rc1 tag.)

* Debug Log:

https://paste.sr.ht/~vilhalmer/d00ed12c8e959eaa421dc251dac25cf45c554441

* Configuration File:
```
for_window [app_id=__focused__] move scratchpad, exec sway-crash
exec termite -e 'sh -c "sleep 1 && swaymsg reload && termite"'
```
* `sway-crash` script:
```
#!/usr/bin/env python
import i3ipc

tree = i3ipc.Connection().get_tree()
tree.find_focused().command('mark _target')  # Mark window in current workspace
scratch = [c for c in tree.descendents() if c.name == '__i3_scratch'][0].floating_nodes[0]  # Grab the termite on the scratchpad
scratch.command("move to mark _target")  # Boom
```

https://paste.sr.ht/~vilhalmer/078cada92d1dbaa525d82fb7c178f0c1b941c412

RedSoxFan:
**Behavior matches i3 PR: https://github.com/i3/i3/pull/3776**

Fixes #4473 

When moving to a scratchpad hidden container (using
`move [window|container] [to] mark <mark>`), this moves the container to
the scratchpad (equivalent to `move [window|container] [to] scratchpad`).

Previously, this would crash since the destination did not have a
workspace.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -388,6 +388,8 @@ static bool container_move_in_direction(struct sway_container *container,
 	return false;
 }
 
+static struct cmd_results *cmd_move_to_scratchpad(void);
+
 static struct cmd_results *cmd_move_container(bool no_auto_back_and_forth,
 		int argc, char **argv) {
 	struct cmd_results *error = NULL;
@@ -503,6 +505,11 @@ static struct cmd_results *cmd_move_container(bool no_auto_back_and_forth,
 		destination = &dest_con->node;
 	} else {
 		return cmd_results_new(CMD_INVALID, expected_syntax);
+	}
+
+	if (destination->type == N_CONTAINER &&
+			container_is_scratchpad_hidden(destination->sway_container)) {
+		return cmd_move_to_scratchpad();
 	}
 
 	if (container->is_sticky && container_is_floating_or_child(container) &&


RedSoxFan:
The corresponding i3 PR has been merged

ddevault:
Thanks!

Issue #4483, opened 2019-08-21
ipc command results incompatible with i3


lbonn:
**Sway Version**: master

sway and i3 differ in the way they handle responses to commands with criterias:

- i3 will apply the operation to all matching nodes and return a global response, usually returning failure if the criteria match is empty
- sway will return one response per criteria match

This makes sway's ipc harder to use reliably. Example on my session where I have 5 open chromium:

```
$ waymsg -t command '[class="Chromium-browser"], focus; move up'
[
   { "success": true},
   { "success": true},
   { "success": true},
   { "success": true},
   { "success": true},
   { "success": true}
 ]
```

whereas i3 would return only 2 objects in the array.

More annoying, if there is no match (`[class="Bogus"]`), an empty response will be returned instead of a failure status.

PR #4484, opened 2019-08-21
ipc: collapse multi-container command results

lbonn:
Fix for #4483.

That's the quick and dirty solution instead of refactoring all commands to process lists of nodes.

It might be enough or not but it should be closer to the intended behavior than what we have now.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -284,17 +284,32 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 				free_argv(argc, argv);
 				goto cleanup;
 			}
+		} else if (views->length == 0) {
+			list_add(res_list,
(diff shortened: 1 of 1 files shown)


sway/commands.c
```diff
@@ -284,17 +284,30 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 				free_argv(argc, argv);
 				goto cleanup;
 			}
+		} else if (views->length == 0) {
+			list_add(res_list, cmd_results_new(CMD_FAILURE, "No matching node."));
```
RedSoxFan: style: can you wrap this so it is under 80 columns?

lbonn:
Thanks for the review. Just refreshed the styling.

RedSoxFan:
Thanks!

PR #4485, opened 2019-08-21
sway.5: remove mention of floating_scroll

RedSoxFan:
It looks like floating_scroll was still in the sway(5) man page as a
remnant of the 0.x era. This just removes it from the man page since it
is no longer a valid command. Mouse bindings with Button4-7 can be used
instead


ddevault:
Thanks!

PR #4495, opened 2019-08-24
Fix formatting for title_format in man 5 sway

MichaelAquilina:
Use explicit linebreaks to make scdoc use a separate line for each entry
listed


ddevault:
Thanks!

PR #4500, opened 2019-08-26
Remove extraneous load_swaybars call

Hummer12007:
Swaybars are already loaded after calling load_main_config

Due to this call, swaybar/status_command was spawned twice (with the second instance of status_command being kept alive for some reason) on each config reload.


emersion:
>Swaybars are already loaded after calling load_main_config

Where are swaybars spawned exactly?

Hummer12007:
When processing the `bar` command.

Hummer12007:
Well, this call would still be extraneous even after implementing the correct solution.

PR #4501, opened 2019-08-27
cmd_mode: don't reset to default after subcommand

RedSoxFan:
In cmd_mode, the mode is currently being reset to the default mode after
a mode subcommand handler is executed. This stores and restores the mode
instead


ddevault:
Thanks!

PR #4504, opened 2019-08-27
contrib: add a keyboard layout per window script

emersion:
Fixes https://github.com/swaywm/sway/issues/2361

### DIFF
--- a/contrib/keyboard-layout-per-window.py
+++ b/contrib/keyboard-layout-per-window.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+
+# This script keeps track of active keyboard layouts per window.
+#
+# This script requires i3ipc-python package (install it from a system package
+# manager or pip).
+
+import i3ipc
+
+ipc = i3ipc.Connection()
+prev_focused = -1
+windows = {}
+
+def on_window_focus(ipc, event):
+    global windows, prev_focused
+
+    # Save current layout
+    layouts = {}
+    for input in ipc.get_inputs():
+        layouts[input.identifier] = input.xkb_active_layout_index
+    windows[prev_focused] = layouts
+
+    # Restore layout of the newly focused window
+    if event.container.id in windows:
+        for (input_id, layout_index) in windows[event.container.id].items():
+            if layout_index != layouts[input_id]:
+                ipc.command("input \"{}\" xkb_switch_layout {}".format(
+                    input_id, layout_index))
+
+    prev_focused = event.container.id
+
+def on_window_close(ipc, event):
+    global windows
+    if event.container.id in windows:
+        del(windows[event.container.id])
+
+def on_window(ipc, event):
+    if event.change == "focus":
+        on_window_focus(ipc, event)
+    elif event.change == "close":
+        on_window_close(ipc, event)
+
+ipc.on("window", on_window)
+ipc.main()


progandy:
I believe stored layouts should be removed on ```window::close``` and ```prev_focused``` cleared if it references the closed window.

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,33 @@
...
+    layouts = {}
+    for input in ipc.get_inputs():
+        layouts[input.identifier] = input.xkb_active_layout_index
+    windows[prev_focused] = layouts
+
+    # Restore layout of the newly focused window
+    if event.container.id in windows:
+        for (input_id, layout_index) in windows[event.container.id].items():
+            if layout_index != layouts[input_id]:
+                ipc.command("input \"" + input_id + "\" " +
```
eternal-sorrow: better use format string here

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,33 @@
...
+
+# This script keeps track of active keyboard layouts per window.
+#
+# This script requires i3ipc-python package (install it from a system package
+# manager or pip).
+
+import i3ipc
+
+ipc = i3ipc.Connection()
+prev_focused = -1
```
eternal-sorrow: why not use current window id `ipc.get_tree().find_focused().id`?

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,33 @@
...
+    global windows, prev_focused
+
+    # Save current layout
+    layouts = {}
+    for input in ipc.get_inputs():
+        layouts[input.identifier] = input.xkb_active_layout_index
+    windows[prev_focused] = layouts
+
+    # Restore layout of the newly focused window
+    if event.container.id in windows:
```
eternal-sorrow: also better add `else` statement here switching to the default layout (index 0) if current window id not in `windows`

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,33 @@
...
+prev_focused = -1
+windows = {}
+
+def on_window_focus(ipc, event):
+    global windows, prev_focused
+
+    # Save current layout
+    layouts = {}
+    for input in ipc.get_inputs():
+        layouts[input.identifier] = input.xkb_active_layout_index
```
Hi-Angel: Creating `layouts` uninitialized is unnecessary here, you can create them initialized as:

```python
    layouts = {input.identifier : input.xkb_active_layout_index
               for input in ipc.get_inputs()}
```

And it's faster as a bonus.

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,33 @@
...
+    windows[prev_focused] = layouts
+
+    # Restore layout of the newly focused window
+    if event.container.id in windows:
+        for (input_id, layout_index) in windows[event.container.id].items():
```
Hi-Angel: I think `input_id` may be better named as `kbd_id`, or perhaps `input_dev_id` *(I don't know can the device be anything else besides a keyboard?)*.

Which btw raises a question: what happens if we detached a keyboard? I guess the dictionary now gonna have stale entries, that needs to be cleaned up.

contrib/keyboard-layout-per-window.py
emersion: Not a big win, I prefer simplicity

emersion:
>By the way, does this work with multiple seats?

No

>Which btw raises a question: what happens if we detached a keyboard? I guess the dictionary now gonna have stale entries, that needs to be cleaned up.

I don't see this happening often enough to cause an issue. Unless you have multiple thousand keyboards.

contrib/keyboard-layout-per-window.py
emersion: >I think input_id may be better named as kbd_id, or perhaps input_dev_id (I don't know can the device be anything else besides a keyboard?).

-1

Hi-Angel: > -1

-1, because… ?

> This could even be seen as a feature: if you re-plug the keyboard, you don't loose your window → layout mappings.

Fair enough, but are you sure this ID is always the same for the same keyboard?

Hi-Angel:
> > By the way, does this work with multiple seats?
> 
> No

If we consider this PR as a replacement to https://github.com/swaywm/sway/pull/3155, then per-seat functional needs to be implemented, because, clearly, the replacement needs to be at least as good as the one supposed to get replaced.

eternal-sorrow:
I think, this script isn't supposed to be the catch-all solution for everyone. Personally, I'm going to create another one that saves the layout data on disk and restores it after restart and also has some other nice features (uses asyncio with i3ipc-python instead of old API for example).

emersion:
> the replacement needs to be at least as good as the one supposed to get replaced.

This is not supposed to be a replacement. This is supposed to be an alternate solution. To be honest, I won't even be using it myself, the goal here is just to give a hint for people who want this feature. Feel free to improve it if you want multi-seat or other features.

Full disclaimer: I don't personally care if this gets merged or not. But if you're so salty about it, I'd suggest to completely ignore this PR.

Hi-Angel:
> Full disclaimer: I don't personally care if this gets merged or not. But if you're so salty about it, I'd suggest to completely ignore this PR.

I don't understand what you mean. I didn't say anything wrong or bad, did I? My assumption that this PR was supposed to be a replacement didn't appear out of pure air, it were your words https://github.com/swaywm/sway/pull/3155#issuecomment-525393122, quoting:

> > Is this moving any further?
> 
> #4504

I'm sorry if I was confused, and these 2 PRs supposed to co-exist. Although I wonder then, why would you post this reply *(the one I quoted)*.

emersion:
Sorry, I've had enough of this. I'm dropping the ball.

Hi-Angel:
> Sorry, I've had enough of this. I'm dropping the ball.

Do we merge https://github.com/swaywm/sway/pull/3155 then? Sorry to ask it here, but it seems relevant, so far you were the only one against.

contrib/keyboard-layout-per-window.py
Hi-Angel: @emersion please un-resolve this discussion. For some reason I can't do this, and this one is ongoing.

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,44 @@
...
+            if layout_index != layouts[input_id]:
+                ipc.command("input \"{}\" xkb_switch_layout {}".format(
```
gdamjan:
```suggestion
                ipc.command("""input "{}" xkb_switch_layout {}""".format(
```
maybe this is more readable, the back slashes really threw me off

PR #4509, opened 2019-08-29
criteria: fullscreen mode criteria, fix #4222

### LINKED ISSUE #4222: Feature request: add fullscreen criteria
would be really nice to match against fullscreen (or fullscreen_mode?)
like:
`for_window [fullscreen] <command>`

versusvoid:


emersion:
What is your use-case for this feature?

versusvoid:
With status bar on top `move position 0 0` on fullscreen window will move it down a bar height, leaving black bar on top and cutting window's bottom.

emersion:
Isn't this a bug? How does i3 behave?

versusvoid:
I don't think it's a bug. User asks WM to move window, WM moves window.

Never used i3, so wouldn't know.

emersion:
Please check how i3 behaves. We are an i3-compatible compositor, we should do what i3 does.

PR #4520, opened 2019-08-31
Fix markup syntax in man pages

tbm:


emersion:
Can you explain this change? The old syntax looks good to me.

ddevault:
scdoc is not markdown, \` means something different here.

Thanks!

tbm:
> Can you explain this change? The old syntax looks good to me.

Backticks are Markdown syntax; they are put into the manpage as it is by scdoc instead of creating bold/code formating.

(The exception seems to be 3 backticks for code blocks; that seems allowed in scdoc)

PR #4521, opened 2019-08-31
seat: set cursor image only if no pointer cap previously

xdavidwu:
This make seat_update_capabilities set cursor image only if
there was no pointer cap before update. This avoid resetting
cursor to left_ptr if an input device is removed.


ddevault:
Thanks!

PR #4522, opened 2019-08-31
seat: avoid unneeded reloading xcursor theme

xdavidwu:
Load xcursor theme on configuring pointer or tablet tool only if
there was no pointer cap before.

With #4521 , this also avoid resetting cursor to left_ptr if new
device is added but there was pointer cap already.


ddevault:
Thanks!

Issue #4517, opened 2019-08-31
Sway 1.2 crashes after copy/paste with `wl-copy` in neovim


arolle:
- sway version 1.2 (Aug 29 2019, Arch Linux)`
- `nvim --version` v0.3.8

## Explanation of steps taken to reproduce the issue.

```
$ nvim -u NONE -V10nvimdebug.log \
  -c "set clipboard=unnamed,unnamedplus" \
  -c "silent normal ii" -c "silent normal yypp"
```
Then neovim is not responsive and sway (with default config) crashes when opening
new terminals. Earlier this had worked.

```
line 0: sourcing "/usr/share/nvim/runtime/autoload/provider/clipboard.vim"
finished sourcing /usr/share/nvim/runtime/autoload/provider/clipboard.vim
continuing in command line
[...]
clipboard: error invoking wl-copy: wl_display_dispatch: Broken pipe 
clipboard: error invoking wl-copy: wl_display_dispatch: Broken pipe 
Writing ShaDa file [...]
```

[Sway stack trace and debug log](https://gist.github.com/arolle/6567c582e68ec854c6da8dfb266025a6)

```
                #0  0x00007f270a04eb58 wlr_seat_keyboard_enter (libwlroots.so.3)
                #1  0x0000559ccb609368 n/a (sway)
                #2  0x0000559ccb5ebee3 n/a (sway)
                #3  0x0000559ccb61530c n/a (sway)
                #4  0x00007f270a075cfe n/a (libwlroots.so.3)
                #5  0x00007f270a05a4da n/a (libwlroots.so.3)
                #6  0x00007f270a0710ba n/a (libwlroots.so.3)
                #7  0x00007f270a0713b8 n/a (libwlroots.so.3)
                #8  0x00007f27095246d0 ffi_call_unix64 (libffi.so.6)
                #9  0x00007f27095240a0 ffi_call (libffi.so.6)
                #10 0x00007f270a0bc82f n/a (libwayland-server.so.0)
                #11 0x00007f270a0b9193 n/a (libwayland-server.so.0)
                #12 0x00007f270a0ba7f2 wl_event_loop_dispatch (libwayland-server.so.0)
                #13 0x00007f270a0b939c wl_display_run (libwayland-server.so.0)
                #14 0x0000559ccb5de592 n/a (sway)
                #15 0x00007f270a6b8ee3 __libc_start_main (libc.so.6)
                #16 0x0000559ccb5de8be n/a (sway)
```

[...]

Issue #4502, opened 2019-08-27
Segfault in wl_resource_get_client


kennylevinsen:
Triggered randomly while testing wldash. Seems to be the focus logic breaking. Unfortunately lost the debug log.

sway 1a8b9fda8a81d71ec9de65fd4c3f1df40e03f971, wlroots a20bb38763877410a80a0c02ddfbfaa906c288fd

Dump below is debugoptimized, not a release build.

```
GNU gdb (GDB) Fedora 8.3-6.fc30
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /usr/local/bin/sway...
[New LWP 26525]
[New LWP 26527]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/usr/local/bin/sway -d'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fa4e04349e4 in wl_resource_get_client () from /lib64/libwayland-server.so.0
[Current thread is 1 (Thread 0x7fa4df569a80 (LWP 26525))]
[...]

PR #4525, opened 2019-08-31
layer-shell: don't give focus to unmapped layer surfaces

emersion:
Focused layers are not cleared when destroyed, they are cleared on unmap.
Giving focus to an unmapped layer surface is (1) incorrect and (2) triggers a
use-after-free.

Closes: https://github.com/swaywm/sway/issues/4517
Closes: https://github.com/swaywm/sway/issues/4502


vilhalmer:
This fixes the crash for me.

ddevault:
Thanks!

PR #4528, opened 2019-09-01
man: document how to specify multiple xkb layouts and options

emersion:
This isn't described in xkeyboard-config(7).


tbm:
fwiw, this is based on a question I just asked on IRC.  This patch makes things much clearer!

ddevault:
Thanks!

PR #4531, opened 2019-09-02
seatop_default: only focus container on press

### LINKED ISSUE #4529: `bindsym button1 nop` doesn't work
Sway 1.2

With `bindsym button1 nop` in my config, I expect that if I left-click on the titlebar of an unfocused window, nothing will happen (as in [i3](https://i3wm.org/docs/userguide.html#_nop)). Instead, it gets focused.

RedSoxFan:
Fixes #4529

This matches i3's behavior of only focusing a container when pressed.
This allows for `bindsym button1 nop`, `bindsym BTN_LEFT nop`, or
`bindcode 272 nop` to be used to disable focusing when clicking on the
title (or with additional flags to `bind{code,sym}` other portions of
the container).

Without this additional condition, the user would need both
`bindsym button1 nop` and `bindsym --release button1 nop` to override
both the pressed and released behavior.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -400,7 +400,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 	}
 
 	// Handle clicking a container surface or decorations
-	if (cont) {
+	if (cont && state == WLR_BUTTON_PRESSED) {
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4532, opened 2019-09-02
Fully search icon_theme_path if icon isn't found in expected paths

benjamb:
I felt it inappropriate to mutate the `sni->tray->basedirs` list, though I'm sure this isn't the right way of going about this, but I'm new to the code base and my C is a little rusty.

This (or some variation of this) results in the proper icon being displayed in the tray for Quassel, I've not tested against anything else.

### DIFF
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -411,9 +411,15 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
 		char *icon_name = sni->status[0] == 'N' ?
 			sni->attention_icon_name : sni->icon_name;
 		if (icon_name) {
-			char *icon_path = find_icon(sni->tray->themes, sni->tray->basedirs,
+			list_t *icon_search_paths = create_list();
+			list_cat(icon_search_paths, sni->tray->basedirs);
+			if (sni->icon_theme_path) {
+				list_add(icon_search_paths, sni->icon_theme_path);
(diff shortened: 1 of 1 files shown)


benjamb:
This currently causes a crash in i3status-rust when the icon changes (e.g. if I disconnect from the core), I need to investigate further.

benjamb:
Fixed the above. I was incorrectly using `list_free_items_and_destroy`, when I just wanted to free the list.

benjamb:
@RedSoxFan Cheers, I've gone ahead and amended my patch given the above.

swaybar/tray/item.c
```diff
@@ -411,9 +411,14 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
 		char *icon_name = sni->status[0] == 'N' ?
 			sni->attention_icon_name : sni->icon_name;
 		if (icon_name) {
-			char *icon_path = find_icon(sni->tray->themes, sni->tray->basedirs,
+			list_t *icon_search_paths = create_list();
+			list_cat(icon_search_paths, sni->tray->basedirs);
+			if (sni->icon_theme_path)
```
RedSoxFan: style: braces are required

benjamb: Ah, cheers. I've updated.

RedSoxFan:
Thanks!

PR #4534, opened 2019-09-03
swaybar: complete barconfig_update event handling

### LINKED ISSUE #4533: i3blocks Spawning duplicates
* Sway Version: sway version 1.2

After config reload, i3blocks spawns another process every time. Not sure if Sway issue or i3blocks
Before reload:
before
After reloading a few times:
aftermany

RedSoxFan:
Supersedes/Closes #4159
Supersedes/Closes #4500
Closes #4533 

This adds complete support for the barconfig_update ipc event. This also
changes the bar command and subcommand handlers to correctly emit the
event. This makes it so all bar subcommands other than id and
swaybar_command are dynamically changeable at runtime. sway-bar.5 has
been updated accordingly

_Note: Although this does provide a solution for #4500/#4533, it does not handle
the generic case of the status command process not being terminated when
swaybar segfaults or aborts. I'm not sure what the best solution for that would be_

### DIFF
--- a/include/swaybar/bar.h
+++ b/include/swaybar/bar.h
@@ -41,6 +41,7 @@ struct swaybar {
 	int ipc_socketfd;
 
 	struct wl_list outputs; // swaybar_output::link
+	struct wl_list unused_outputs; // swaybar_output::link
 	struct wl_list seats; // swaybar_seat::link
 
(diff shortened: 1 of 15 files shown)


ddevault:
Thanks!

PR #4536, opened 2019-09-03
Do not search for edges on subsurfaces, fix #4381

### LINKED ISSUE #4381: Popup clicks don't work when popup is drawn over another container
* Sway Version: sway version 1.1-rc1-80-ge4bba906 (Jul 27 2019, branch 'master')
* Debug Log: https://gist.github.com/pedrocr/535e2380a303660c426505e56b489de6
* Configuration File: default

When a popup is over another container, clicks on menu items don't work. To reproduce use firefox or a gnome-terminal with tabs and right-click on a tab such that the popup opens over the container to the right. Clicking a menu item will then not work over the other container but will work over the parent container. The situation is something like this:

popup

Clicking on "Close Terminal" works over the parent container but not over the other container on the right.

I tested this with 1.0 and 1.1.1 as well, using wlroots 0.5 and 0.6. All of them are broken but in 1.0 nothing happens when you click. With 1.1.1 and master when you click it's as if the mouse has left and reentered the popup as the highlight over the menu item being hovered disappears and reappears.

versusvoid:
Subsurfaces (in most cases popups) aren't decorated by sway
and will never have any borders, but may be drawn beyond container
boundaries producing false positive when searching for edge.
So we want to skip edge search when handling mouse event on subsurface.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -58,8 +58,8 @@ static bool edge_is_external(struct sway_container *cont, enum wlr_edges edge) {
 }
 
 static enum wlr_edges find_edge(struct sway_container *cont,
-		struct sway_cursor *cursor) {
-	if (!cont->view) {
+		struct wlr_surface *surface, struct sway_cursor *cursor) {
+	if (!cont->view || cont->view->surface != surface) {
 		return WLR_EDGE_NONE;
 	}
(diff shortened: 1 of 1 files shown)


ddevault:
Can you add this email address to your github account, or commit again with an address which already is?

versusvoid:
@ddevault done

PR #4537, opened 2019-09-03
pthread_sigmask()-out SIGUSR1 in main() so GPU drivers (i.e. libMali) that spawn threads don't get sway killed by xwayland's SIGUSR1

a-m-joseph:
Sway needs to mask SIGUSR1 in main() like Weston does.  Otherwise the compositor dies a grisly death as soon as the user does anything that triggers xwayland loading if the GPU driver shared library uses threads.  Tested on rk3399/libMali with claws-mail, which works now.

Apologies if this pull request is entered or formatted improperly.  I've been using git for quite a while but this is my first time using github to submit a patch instead of just emailing it.

Fixes https://github.com/swaywm/sway/issues/4538

### DIFF
--- a/sway/meson.build
+++ b/sway/meson.build
@@ -210,5 +210,6 @@ executable(
 	include_directories: [sway_inc],
 	dependencies: sway_deps,
 	link_with: [lib_sway_common],
-	install: true
+	install: true,
+	link_args : '-lpthread'
 )
--- a/sway/main.c
+++ b/sway/main.c
@@ -214,6 +214,19 @@ void enable_debug_flag(const char *flag) {
 }
 
 int main(int argc, char **argv) {
+
+	// Shared libraries (such as libMali.so, the Mali OpenGL
+	// driver) may create their own threads.  To ensure that
+	// Xwayland's SIGUSR1 does not hit those threads and kill
+	// sway, we must call pthread_sigmask() as early as possible.
+	// See also weston commit
+	// f59dc1112be50467b7c0f8aeba68f3aa10d36725 which does the
+	// same thing.
+	sigset_t mask;
+	sigemptyset(&mask);
+	sigaddset(&mask, SIGUSR1);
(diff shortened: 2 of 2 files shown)


a-m-joseph:
builds.sr.ht: alpine.yml — builds.sr.ht job failed 

../sway/server.c:14:10: fatal error: wlr/types/wlr_gamma_control.h: No such file or directory
 #include <wlr/types/wlr_gamma_control.h>

Hrm, I don't think I did that...

emersion:
Thanks for your PR!

>Hrm, I don't think I did that...

You'll need to open a PR against master.

>Sway needs to mask SIGUSR1 in main() like Weston does.

I see. Wouldn't be a better idea to do this in `wlr_xwayland_create`? This function sets up the global signal handler anyway. It would be nice to document that behaviour in the header doc comment, too.

emersion:
Also, can you make sure to author your commits with a valid email address? Thanks

sway/meson.build
```diff
@@ -210,5 +210,6 @@ executable(
 	include_directories: [sway_inc],
 	dependencies: sway_deps,
 	link_with: [lib_sway_common],
-	install: true
+	install: true,
+	link_args : '-lpthread'
```
ascent12: The correct way to do this in meson is with the 'threads' dependency.

a-m-joseph:
> I see. Wouldn't be a better idea to do this in `wlr_xwayland_create`?

No, that's too late.

The pthread_sigmask() call needs to happen before the server_privileged_prepare() call in sway's main.c.

sway/meson.build
a-m-joseph: Ah thanks.  I've never used meson/ninja before.

a-m-joseph:
Superseded by https://github.com/swaywm/wlroots/pull/1815, closing.

a-m-joseph:
Reopening per https://github.com/swaywm/wlroots/pull/1815#issuecomment-527903621

PR #4539, opened 2019-09-03
Fix bar_state_update docs

xPMo:
`man sway-ipc` incorrectly showed a `bar_status_update` event, when the event is `bar_state_update`.


RedSoxFan:
Thanks!

Issue #4542, opened 2019-09-05
Marked Containers Lose Their Marks


asahaf:
Already marked containers lose their mark when `for_window` applies a new mark to a newly created container, leaving only the new container with that mark, removing it from all other matched containers.

Sway Version: 1.2

Steps to reproduce:
1. add this to the config file: `for_window [app_id="firefox"] mark browser`, or call `swaymsg for_window '[app_id="firefox"] mark browser'`
2. open a new firefox window. the new container gets marked with `brwoser`
3. open a new firefox window. the new container gets marked with `browser`, but the previous one loses its mark

According to the documentation, `mark` by default do `--replace`, but `mark` in this case seems always doing toggle

RedSoxFan:
A mark can only be set for a single window at a time since they act as unique identifiers.

`--add` and `--replace` alter the behavior for the marks list for a single window. By default, a window will only be have a single mark, but `--add` can be used to allow it to have multiple. Neither of these options have anything to do with how many windows can have the same mark.

I've submitted #4545  to clarify the documentation

PR #4545, opened 2019-09-05
sway.5: clarify that a marks are unqiue

RedSoxFan:
Related to #4542 

This just clarifies that a mark can only be set for a single window
since they are used as unique identifiers


asahaf:
Thanks for the clarification.

It needs to be fixed here too: [https://github.com/swaywm/sway/blob/2c1a11016ccb7646a74de51eff003c18e5aa7902/sway/sway.5.scd#L789](https://github.com/swaywm/sway/blob/2c1a11016ccb7646a74de51eff003c18e5aa7902/sway/sway.5.scd#L789)

PR #4546, opened 2019-09-05
Fix typo (an -> and)

NilsIrl:
While I'm at it could I ask a question?

I have a device that provides absolute events (a wacom tablet), how can I make it relative?


emersion:
>I have a device that provides absolute events (a wacom tablet), how can I make it relative?

Not possible. Why do you need that?

(Better discuss this in a separate issue)

PR #4555, opened 2019-09-09
swaybar: make status block text render in the same way as others

xdavidwu:
Other components like workspace button, status line (error or plain
text) already render text at integer coords. This make status block
also render text at integer coords.


RedSoxFan:
Thanks!

PR #4576, opened 2019-09-21
Add zsh sway command completions for swaymsg

RPigott:
So this started as just a personal thing when I could never remember the order of some arguments to sway commands in swaymsg and I added to it over time. Now I've cleaned it up and made it more complete. The main new things are:
- Complete sway commands and their arguments where appropriate
- Complete criteria attributes where appropriate
- Complete sway-ipc event types for subscribe where appropriate
- A boat load of descriptions mostly copy/pasted from the man pages for you to read if you have verbose menu completions on

Some nouns, like input device identifiers, are only offered completions if the user has jq installed, as sway/swaymsg doesn't really have an output format that's particularly suitable for parsing in the shell. Users without jq shouldn't have any issues though.

I had to read the man pages a lot and I noticed some errors. Looks like few are already fixed (I'm on 1.2) but I included the --to-code flag for unbindsym in sway(5) (and in the completion definition) because it appears to be valid in the case and worked as expected for me when I tried it. Also in sway-input(5), repeat_rate and repeat_delay are documented twice: once up front and again under "libinput configuration". I left the man page alone because I'm not sure which instance should be removed.

So yeah if you're a zsh user try these out. I tried to test them a lot, but the completions are quite long now so I might have missed something. Thanks to the folks on #zsh freenode and the zsh-users mailing list who helped answer my questions.

### DIFF
--- a/completions/zsh/_swaymsg
+++ b/completions/zsh/_swaymsg
@@ -1,39 +1,754 @@
 #compdef swaymsg
-#-----------------
-# Description
-# -----------
-#
-# Completion script for swaymsg in sway wm (http://swaywm.org)
-#
-# ------------------------------------------------------
-# Author
-# --------
-#
-#  * Seth Barberee <seth.barberee@gmail.com>
-#
-#  -------------------------------------------
 
-types=(
-'get_workspaces'
-'get_seats'
-'get_inputs'
-'get_outputs'
-'get_tree'
-'get_marks'
-'get_bar_config'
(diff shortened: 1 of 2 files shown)


ddevault:
Err, we'll take the man page if you want, but I don't really want to shove a full list of sway's commands into the swaymsg zsh script.

RPigott:
Okay. I'll close this and make a different PR later.

PR #4577, opened 2019-09-21
output: mode: fix log message; refresh rate is in Hz, not GHz

dnkl:


sway/config/output.c
```diff
@@ -260,7 +260,7 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 
 	bool modeset_success;
 	if (oc && oc->width > 0 && oc->height > 0) {
-		sway_log(SWAY_DEBUG, "Set %s mode to %dx%d (%f GHz)", oc->name, oc->width,
+		sway_log(SWAY_DEBUG, "Set %s mode to %dx%d (%f Hz)", oc->name, oc->width,
```
emersion: I think it's mHz. It's copied directly from the `wlr_output_mode`.

dnkl: Hmm, no I don't think so. `oc->refresh_rate` is passed as is to `set_mode()`, where it is multiplied with 1000 and from there on it's called mHz.

emersion: When I grep for places where it's set, I only get:

```
oc->refresh_rate = mode->refresh;
```

`mode->refresh` is in mHz.

dnkl: Also, this log message only appears when a "mode" command is used. If the mode command has left out the refresh rate, it is `-1.0`. If not, the user specified it (something like 1920x1080@60Hz), and then it appears like this in the log:

```
2019-09-21 14:17:39 - [sway/config/output.c:263] Set eDP-1 mode to 1920x1080 (60.056999 GHz)
```

dnkl: There's also this log message (also in `sway/config/output.c`, which also uses **Hz**:
```
        sway_log(SWAY_DEBUG, "Config stored for output %s (enabled: %d) (%dx%d@%fHz "
                "position %d,%d scale %f subpixel %s transform %d) (bg %s %s) (dpms %d)",
                oc->name, oc->enabled, oc->width, oc->height, oc->refresh_rate,
                oc->x, oc->y, oc->scale, sway_wl_output_subpixel_to_string(oc->subpixel),
                oc->transform, oc->background, oc->background_option, oc->dpms_state);
```
Is it possible that the line `oc->refresh_rate = mode->refresh` is wrong (should be divided with 1000)? Or the opposite, that all other instances of `oc->refresh_rate` got it wrong?

emersion: Yeah it's perfectly possible that there are serious mistakes here.

dnkl: I think the line `oc->refresh_rate = mode->refresh;` is wrong. It is from the function `default_output_config`. This function is **only** called when **reloading** a configuration https://github.com/swaywm/sway/blob/00c4c7e8cc437c6f905c6fbe94cf8eaf508f8644/sway/config/output.c#L370

I instrumented sway a bit, to see what happens with regard to the refresh rate when the config doesn't specify one:
```diff
diff --git a/sway/config/output.c b/sway/config/output.c
index 9da009bb..2148a340 100644
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -213,6 +213,7 @@ static bool set_mode(struct wlr_output *output, int width, int height,
 	wl_list_for_each(mode, &output->modes, link) {
 		if (mode->width == width && mode->height == height) {
 			if (mode->refresh == mhz) {
+				sway_log(SWAY_ERROR, "matched refresh rate: %d", mhz);
 				best = mode;
 				break;
 			}
@@ -329,6 +330,7 @@ static void default_output_config(struct output_config *oc,
 			wl_container_of(wlr_output->modes.prev, mode, link);
 		oc->width = mode->width;
 		oc->height = mode->height;
+		sway_log(SWAY_ERROR, "setting default refresh rate: %d", mode->refresh);
 		oc->refresh_rate = mode->refresh;
 	}
 	oc->x = oc->y = -1;
```

This resulted in the following output when reloading the configuration:
```
2019-09-28 22:03:25 - [sway/config/output.c:333] setting default refresh rate: 120114
2019-09-28 22:03:25 - [sway/config/output.c:258] Turning on screen
2019-09-28 22:03:25 - [sway/config/output.c:264] Set * mode to 1920x1080 (120114.000000 GHz)
2019-09-28 22:03:25 - [sway/config/output.c:228] Assigning configured mode to eDP-1
```

We can see that `default_output_config` is being called. What we **don't** see is the line `sway_log(SWAY_ERROR, "matched refresh rate: %d", mhz);` But we **do** see _Assigning configured mode to eDP-1_ which is from the same function. This means we did **not** match the refresh rate, but used the "best" one found (the last one for this particular width x height).

If I apply this patch:
```diff
@@ -329,7 +330,8 @@ static void default_output_config(struct output_config *oc,
 			wl_container_of(wlr_output->modes.prev, mode, link);
 		oc->width = mode->width;
 		oc->height = mode->height;
-		oc->refresh_rate = mode->refresh;
+		oc->refresh_rate = mode->refresh / 1000.0;
```

The log instead shows:
```
2019-09-28 22:16:49 - [sway/config/output.c:333] setting default refresh rate: 120114
2019-09-28 22:16:49 - [sway/config/output.c:258] Turning on screen
2019-09-28 22:16:49 - [sway/config/output.c:264] Set * mode to 1920x1080 (120.113998 GHz)
2019-09-28 22:16:49 - [sway/config/output.c:216] matched refresh rate: 120114
2019-09-28 22:16:49 - [sway/config/output.c:228] Assigning configured mode to eDP-1
```

I.e. we now actually match it, instead of falling back to the last one (which happens to be the same).

For reference, this is the `set_mode` function where we see the refresh-rate (in HZ) being converted to mHZ and then matched against the available refresh rates.

```c
static bool set_mode(struct wlr_output *output, int width, int height,
		float refresh_rate) {
	int mhz = (int)(refresh_rate * 1000);
	if (wl_list_empty(&output->modes)) {
		sway_log(SWAY_DEBUG, "Assigning custom mode to %s", output->name);
		return wlr_output_set_custom_mode(output, width, height, mhz);
	}

	struct wlr_output_mode *mode, *best = NULL;
	wl_list_for_each(mode, &output->modes, link) {
		if (mode->width == width && mode->height == height) {
			if (mode->refresh == mhz) {
				sway_log(SWAY_ERROR, "matched refresh rate: %d", mhz);
				best = mode;
				break;
			}
			best = mode;
		}
	}
	if (!best) {
		sway_log(SWAY_ERROR, "Configured mode for %s not available", output->name);
		sway_log(SWAY_INFO, "Picking default mode instead");
		best = wl_container_of(output->modes.prev, mode, link);
	} else {
		sway_log(SWAY_DEBUG, "Assigning configured mode to %s", output->name);
	}
	return wlr_output_set_mode(output, best);
}

```

Is this enough proof?

Should I open up a second pull request for setting the default refresh rate correctly? Or just fix it in this one?

ammgws:
What happened with this?

PR #4579, opened 2019-09-22
Error out on unknown debug flag

emersion:
Otehrwise it's pretty easy to misremember a flag and think damage=rerender is
enabled when it's not.


ddevault:
Thanks!

Issue #4578, opened 2019-09-22
segfault in container_is_floating when assigning with tiling criteria.


RPigott:
* Sway Version: sway version 1.2-rc1-51-gda0ad4c3 (Sep 22 2019, branch 'makepkg')

* Debug Log: [sway.log](https://gist.githubusercontent.com/RPigott/998852adb061935a17504037181926fd/raw/5b947a143d82d092dc16aa07f328bd3c58ef1d5c/sway.log)

<details>
<summary>added sway config lines</summary>

```
assign [app_id="transmission-gtk" tiling] 5
bindsym $mod+z exec transmission-gtk
```
</details>

* Steps to reproduce:

1. Assign a view with 'app_id' and 'tiling' criteria to a (non-focused?) workspace (in sway config)
2. Create a view matching the criteria
3. Segfault

I think that's it actually. I used transmission-gtk. I haven't tested with other applications. Maybe xdg-shell specific looking at the stack trace, but I don't know. Let me know if you can/can't reproduce.

<details>
<summary>
stack trace
</summary>

```
#0  0x0000557d6c7e7d61 in container_is_floating (container=0x0) at ../sway/sway/tree/container.c:805
#1  0x0000557d6c7b473f in criteria_matches_view (view=0x557d6d564900, criteria=0x557d6d037e30) at ../sway/sway/criteria.c:190
#2  0x0000557d6c7b473f in criteria_matches_view (criteria=0x557d6d037e30, view=0x557d6d564900) at ../sway/sway/criteria.c:100
#3  0x0000557d6c7b4915 in criteria_for_view (view=view@entry=0x557d6d564900, types=types@entry=(CT_ASSIGN_OUTPUT | CT_ASSIGN_WORKSPACE | CT_ASSIGN_WORKSPACE_NUMBER)) at ../sway/sway/criteria.c:229
        criteria = 0x557d6d037e30
        i = 1
        criterias = 0x557d6d009450
        matches = 0x557d6d039d20
#4  0x0000557d6c7ed47a in select_workspace (view=0x557d6d564900) at ../sway/sway/tree/view.c:475
        seat = 0x557d6d0057c0
        criterias = <optimized out>
        ws = <optimized out>
        node = <optimized out>
        __PRETTY_FUNCTION__ = "select_workspace"
        __PRETTY_FUNCTION__ = "view_map"
        ws = <optimized out>
        seat = <optimized out>
        node = <optimized out>
        target_sibling = <optimized out>
[...]

PR #4580, opened 2019-09-22
Create container before selecting workspace

RPigott:
This is my attempt at fixing #4578 . This is my first time working on sway, so sorry if there is a glaring issue with the patch.

My understanding is this: some criteria like 'tiling' and 'floating' predicates need a container to determine if they match. When container is null it causes a segfault. All I've done is move container_create earlier in view_map so that it precedes select_workspace and all the criteria can check matches without error.

This avoids the segfault for me, and it doesn't seem that view is modified in the intermediate code, so I don't think there will be unintended side-effects.


ddevault:
LGTM, thanks!

PR #4583, opened 2019-09-23
build: update wlroots version, be more strict

emersion:
wlroots versions are incompatible with each other. Often our users struggle
with figuring out that their wlroots version is too old after a new release.
Use a more strict version check to prevent building sway with incompatible
wlroots versions.


ddevault:
Thanks!

Issue #4586, opened 2019-09-24
Criteria [workspace=__focuses__] has unexpected behavior.


andremarcais:
#### Info
sway version 1.2 (Aug 29 2019, Arch Linux)
[log](https://gist.github.com/andremarcais/32fe005b5ae8f72eefe426f791ba7aca#file-sway-log)
[cfg](https://gist.github.com/andremarcais/32fe005b5ae8f72eefe426f791ba7aca#file-test-conf)

#### Description
First I go to workspace 10 and add some windows. Then I go to workspace 1 and add some windows. If I then execute `[workspace=__focused__] kill` (while on workspace 1), the windows in both workspace 1 and 10 are killed.

RPigott:
I think this can be closed now.

PR #4587, opened 2019-09-24
criteria: use a strict regex match string for __focused__ value

RPigott:
My attempt at fixing #4586 .

`__focused__` directly substitutes the workspace name for matches, but the value is compared as a regex, so it may unexpectedly match other unfocused workspaces. Instead, a strict regex match should be substituted for `__focused__` value in workspace criteria.

Also, it seems this is the case for several other criteria properties including: app_id, shell, title, instance, and window_role, so I have updated those as well. This patch fixes the issue described for me.

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -55,6 +55,17 @@ static int regex_cmp(const char *item, const pcre *regex) {
 	return pcre_exec(regex, NULL, item, strlen(item), 0, 0, NULL, 0);
 }
 
+static char *strict_regex(const char *str) {
+	size_t size = snprintf(NULL, 0, "^%s$", str) + 1;
+	char *reg = malloc(size);
+	if (!*reg) {
+		sway_log(SWAY_ERROR, "Unable to allocate regex for criteria");
+		return NULL;
+	}
+	snprintf(reg, size, "^%s$", str);
+	return reg;
+}
+
 #if HAVE_XWAYLAND
 static bool view_has_window_type(struct sway_view *view, enum atom_name name) {
 	if (view->type != SWAY_VIEW_XWAYLAND) {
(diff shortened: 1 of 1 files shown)


common/stringop.c
```diff
@@ -309,3 +309,10 @@ char *argsep(char **stringp, const char *delim, char *matched) {
 	}
 	return start;
 }
+
+char *strict_regex(const char *str) {
```
RedSoxFan: This can just be `static` in `sway/criteria.c`

common/stringop.c
```diff
@@ -309,3 +309,10 @@ char *argsep(char **stringp, const char *delim, char *matched) {
...
+	size_t size = strlen(str) + 3;
```
RedSoxFan: Using `size_t size = snprintf(NULL, 0, "^%s$", str) + 1` is preferred for new additions

common/stringop.c
```diff
@@ -309,3 +309,10 @@ char *argsep(char **stringp, const char *delim, char *matched) {
...
+	char *reg = malloc(size);
```
RedSoxFan: Need to check if allocation was successful. Otherwise, log an error and return `NULL`.

RedSoxFan:
Also, the `strdup` at the end of `get_focused_prop` needs to be removed. The only case that doesn't need the `strict_regex` wrapping is `T_CON_ID`, which is already allocating data anyway. That case is actually resulting in a memory leak currently due to that never being freed after the `strdup`

RPigott:
Okay I have made the requested changes.

There is one other issue to consider. For example, try:

```
swaymsg workspace '.*'
swaymsg '[workspace=__focused__]' kill
```
and observe _all_ windows on _all_ workspace being killed. I suppose workspace names should be escaped.

Otherwise the comparison should be reworked to make a non-regex comparison for the `__focused__` case. It seems that is what i3 does : https://github.com/i3/i3/blob/7db0d179a3aa659f1dc109012d253403094c72a2/src/match.c#L175

RedSoxFan:
> I suppose workspace names should be escaped.
> 
> Otherwise the comparison should be reworked to make a non-regex comparison for the `__focused__` case. It seems that is what i3 does : https://github.com/i3/i3/blob/7db0d179a3aa659f1dc109012d253403094c72a2/src/match.c#L175

Good point. I'm not a huge fan of adding the complexity of escaping everything. I think mirroring i3's approach would be better for this case

RPigott:
Closing this since #4627 exists.

PR #4588, opened 2019-09-25
Repaint scheduling

YaLTeR:
Requires https://github.com/swaywm/wlroots/pull/1832.

This PR establishes the necessary infrastructure to delay output renders and frame callbacks to reduce the visual latency between a surface committing new contents and the contents being shown on-screen. Additionally, it adds configurable **constant** delays to individual output renders, and to individual view frame callbacks. The way the code is structured makes it possible to subsequently add some kind of an _adaptive_ mode which figures out the delays automatically for lowest possible latency with minimal dropped frames.

For more details on how this works see [how it's done in Weston](https://www.collabora.com/about-us/blog/2015/02/12/weston-repaint-scheduling/): this PR does pretty much the exact same thing, except it also adds frame callback delaying. Also see https://github.com/swaywm/wlroots/issues/655#issuecomment-511098029 for my comments after implementing a proof-of-concept of this in rootston.

Currently undergoing dogfooding, thus far seems to work well. I have a test client which renders on frame callbacks and prints the delay from receiving a frame callback and the subsequent presentation time. Using this client, I was able to monitor a drop in latency from a little less than 2 frames on the default configuration to about half a frame on two different setups.

How to set up the constants:
1. `output max_render_time`. Run something animating and full-screen, like `glxgears` (as I noted in https://github.com/swaywm/wlroots/issues/655#issuecomment-511098029, the damaged region affects the commit time). Start with `1`, then go up if you see frame drops. On my laptop with an integrated Intel graphics card this works out to be about `5`–`6` ms (Weston went with `7` by default). My desktop with an AMD GPU can easily do `1` (the lowest value).
1. View `max_render_time`. This one will vary per application (since every application obviously takes a different time to render). Run the application full-screen, start with `1`, increase until it stops dropping frames. My test client runs well at `2`, and a game I'm playing runs well at `3`. I haven't tried adjusting this for anything else because I don't really need those extra couple of milliseconds of reduced latency for anything else.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -212,6 +212,7 @@ struct output_config {
 	float scale;
 	int32_t transform;
 	enum wl_output_subpixel subpixel;
+	int max_render_time; // In milliseconds
 
 	char *background;
 	char *background_option;
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -72,7 +72,7 @@ struct sway_output *output_get_in_direction(struct sway_output *reference,
 void output_add_workspace(struct sway_output *output,
 		struct sway_workspace *workspace);
 
-typedef void (*sway_surface_iterator_func_t)(struct sway_output *output,
+typedef void (*sway_surface_iterator_func_t)(struct sway_output *output, struct sway_view *view,
 	struct wlr_surface *surface, struct wlr_box *box, float rotation,
 	void *user_data);
 
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -51,6 +51,11 @@ struct sway_output {
 	struct {
 		struct wl_signal destroy;
 	} events;
+
+	struct timespec last_presentation;
+	uint32_t refresh_nsec;
+	int max_render_time; // In milliseconds
+	struct wl_event_source *repaint_timer;
 };
 
 struct sway_output *output_create(struct wlr_output *wlr_output);
@@ -70,6 +75,8 @@ void output_add_workspace(struct sway_output *output,
 typedef void (*sway_surface_iterator_func_t)(struct sway_output *output,
 	struct wlr_surface *surface, struct wlr_box *box, float rotation,
 	void *user_data);
+
+int output_repaint_timer_handler(void *data);
 
 void output_damage_whole(struct sway_output *output);
 
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -108,6 +108,8 @@ struct sway_view {
 	} events;
 
 	struct wl_listener surface_new_subsurface;
+
+	int max_render_time; // In milliseconds
 };
 
 struct sway_xdg_shell_view {
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -508,6 +508,10 @@ static bool scan_out_fullscreen_view(struct sway_output *output,
 
 int output_repaint_timer_handler(void *data) {
 	struct sway_output *output = data;
+	if (output->wlr_output == NULL) {
+		return 0;
+	}
+
 	output->wlr_output->block_idle_frame = false;
 
 	struct sway_workspace *workspace = output->current.active_workspace;
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -480,19 +480,13 @@ static bool scan_out_fullscreen_view(struct sway_output *output,
 	return wlr_output_commit(wlr_output);
 }
 
-static void damage_handle_frame(struct wl_listener *listener, void *data) {
-	struct sway_output *output =
-		wl_container_of(listener, output, damage_frame);
-	if (!output->enabled || !output->wlr_output->enabled) {
-		return;
-	}
-
-	struct timespec now;
-	clock_gettime(CLOCK_MONOTONIC, &now);
+int output_repaint_timer_handler(void *data) {
+	struct sway_output *output = data;
+	output->wlr_output->block_idle_frame = false;
 
 	struct sway_workspace *workspace = output->current.active_workspace;
 	if (workspace == NULL) {
-		return;
+		return 0;
 	}
 
 	struct sway_container *fullscreen_con = root->fullscreen_global;
@@ -515,7 +509,7 @@ static void damage_handle_frame(struct wl_listener *listener, void *data) {
 		last_scanned_out = scanned_out;
 
 		if (scanned_out) {
-			goto frame_done;
+			return 0;
 		}
 	}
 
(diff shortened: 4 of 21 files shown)


ddevault:
>How to set up the constants

This should probably make it into the man pages

ddevault:
Needs updating per #4584 

Did some testing, seems to work pretty well. Nice work!

YaLTeR:
Done, although I can't get anything to scan-out directly for some reason (as indicated by absence of `Scanning out fullscreen view` in `sway.log`).

emersion:
>I can't get anything to scan-out directly for some reason

This highly depends on the planes IN_FORMATs. You can check the format used by a client with `WAYLAND_DEBUG=1` and check which formats are supported by your planes with `drm_info`.

ddevault:
Needs rebasing once more.

YaLTeR:
Rebased and introduced `sway_surface` to store `frame_done_timer`, which was moved from the wlroots PR.

YaLTeR:
Just found a crash that can happen on output disconnect. In the beginning of `output_repaint_timer_handler` there's this line:
```c
output->wlr_output->block_idle_frame = false;
```
and `wlr_output` was be `NULL`. Should `output_repaint_timer_handler` just return on `output->wlr_output == NULL`? Is it possible for the output itself to get de-allocated while `output_repaint_timer_handler` is scheduled to run?

ddevault:
>Should output_repaint_timer_handler just return on output->wlr_output == NULL

Probably, yes.

ddevault:
Future direction: is it possible to detect dropped frames and tune these numbers automatically?

ddevault:
Been testing this branch, mpv basically hoses the whole thing. Video playback will drop frames at just about any value of max_render_time under 15, and those values basically eliminate the value-add of this feature for all other windows.

YaLTeR:
> Future direction: is it possible to detect dropped frames and tune these numbers automatically?

That's what the idea was for the next step; the properties even allow for a new value like "adaptive" for this potential future behavior.

> Video playback will drop frames at just about any value of max_render_time under 15, and those values basically eliminate the value-add of this feature for all other windows.

I'm assuming you're talking about the output property here? I wonder why mpv behaves weirdly with this, maybe they do their own frame delaying which is interfering or something?

I've been noticing mpv freezing video output on me lately (until I move the mouse for example), but I haven't investigated whether it was related to max_render_time. In any case it seems like a bug in mpv if it stops rendering like that. My test client that renders only by subscribing to a frame callback from within the previous frame callback is able to continue rendering fine, even if max_render_time is set to a value that's too low.

tokyovigilante:
I took this for a test drive on my X1Y4 (Intel UHD 620) and only dropped frames in mpv at 7ms and above (ie 7ms `output max_render_time`). I’m using the native Wayland mpv video output, HW accel using `vaapi` and Vulkan.

Emantor:
Looking at https://github.com/mpv-player/mpv/commit/ea4685b233090bb155cfa4ec6bf87f421a553168 There have been some changes within the mpv wayland backend rather recently, maybe its a good idea to test with a current build from master.

ddevault:
mpv from master is definitely an improvement, I can get it down to 8ms without noticable frame dropping. It's still the worst client I've tested yet, though.

YaLTeR:
> I've been noticing mpv freezing video output on me lately (until I move the mouse for example), but I haven't investigated whether it was related to max_render_time.

After watching a couple hours of videos with max_render_time switched off without issues, I can say that it's most likely related. I'll note that I'm using mpv 0.29.1 with wayland, gpu-hq and display-resample. I still believe it's not a sway issue as everything else works fine.

sway/desktop/output.c
```diff
@@ -401,18 +408,31 @@ bool output_has_opaque_overlay_layer_surface(struct sway_output *output) {
...
 		struct wlr_surface *surface, struct wlr_box *box, float rotation,
 		void *data) {
-	struct timespec *when = data;
-	wlr_surface_send_frame_done(surface, when);
+	int view_max_render_time = 0;
+	if (view != NULL) {
+		view_max_render_time = view->max_render_time;
+	}
+
+	int delay = output->refresh_nsec / 1000000 - output->max_render_time
```
emersion: This delay doesn't take into account the current time. I think we should compute it relative to the presentation timestamp. Otherwise if a client submits a content update pretty late we'll miss vblank.

sway/desktop/output.c
```diff
@@ -522,17 +540,40 @@ static void damage_handle_frame(struct wl_listener *listener, void *data) {
...
+		return;
+	}
+
+	int delay = output->refresh_nsec / 1000000 - output->max_render_time;
+
+	if (output->max_render_time == 0 || delay < 1) {
+		output_repaint_timer_handler(output);
+	} else {
+		output->wlr_output->block_idle_frame = true;
+		wl_event_source_timer_update(output->repaint_timer, delay);
```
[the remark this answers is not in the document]
YaLTeR: So, the convenient thing about doing it here is that it's a single spot where the frames are either drawn normally (with the delay disabled) or scheduled (with the delay enabled). It is also where surfaces are sent the frame callback.

If the output frames are scheduled from the present event handler, then:
- Assuming that the present event handler, if it's called, is called right before the frame handler, the logic for present would be "schedule a frame" and the logic for frame would be "if a frame isn't scheduled, render a frame right away". However, if it's possible for the present event handler to be called later, I'm not sure how to handle the interaction, since we always want to render from the frame handler, and we wouldn't know there if the present event handler will be called shortly after.
- There are cases where the present event handler is called, but the frame event handler isn't, for example, I'm assuming, when the output is disabled (the last frame is presented, but no new frame is requested). The present event handler doesn't know about this, so how should it know whether to schedule a frame?
- The surface frame callback sending would also then be duplicated between the frame handler (if surface delay is disabled) and the present handler (if the delay is enabled, since then it should also be computed from the presentation time), which raises all the same issues.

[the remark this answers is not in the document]
emersion: Sorry about the delay. I'll prioritize this pull request from now on so that we can get this merged.

>It is also where surfaces are sent the frame callback.

Or where the frame callback is scheduled, I assume.

>Assuming that the present event handler, if it's called, is called right before the frame handler, the logic for present would be "schedule a frame" and the logic for frame would be "if a frame isn't scheduled, render a frame right away". However, if it's possible for the present event handler to be called later, I'm not sure how to handle the interaction, since we always want to render from the frame handler, and we wouldn't know there if the present event handler will be called shortly after.

The present event is fired on presentation, so always before the frame callback.

>There are cases where the present event handler is called, but the frame event handler isn't, for example, I'm assuming, when the output is disabled (the last frame is presented, but no new frame is requested). The present event handler doesn't know about this, so how should it know whether to schedule a frame?

I don't think it's a big deal if we schedule a frame and rendering ends up being cancelled because the output is disabled. We already check whether the output is disabled in the rendering function.

>The surface frame callback sending would also then be duplicated between the frame handler (if surface delay is disabled) and the present handler (if the delay is enabled, since then it should also be computed from the presentation time), which raises all the same issues.

Isn't this just a matter of calling `send_frame_done` from these two places, with a `if`?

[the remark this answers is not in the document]
YaLTeR: > The present event is fired on presentation, so always before the frame callback.

So, then, a simple solution is possible of storing the presentation info in the presentation callback, and using that info to compute the delays in the frame callback, otherwise leaving the code as is, right?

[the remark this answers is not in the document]
emersion: All right, let's do this. It'll be easy enough to change if needed anyway.

sway/desktop/output.c
```diff
@@ -765,6 +806,8 @@ static void handle_present(struct wl_listener *listener, void *data) {
 		return;
 	}
 
+	output->refresh_nsec = output_event->refresh;
```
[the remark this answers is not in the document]
YaLTeR: Where exactly should it be zeroed? Along with the `struct timespec last_presentation` that I'm adding.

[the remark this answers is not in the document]
emersion: Right after we render or scan-out (`wlr_output_commit`), probably?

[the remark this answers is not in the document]
YaLTeR: Actually, looks like there's no need to zero them. They provide valid "last known" presentation data. If it's out of date (for example, if sway running as a wayland application was hidden and hasn't received presentation feedback in a while) then the delay computation will simply fall back to no delay, whereas some other part of the code, potentially interested in the refresh rate, could still use the value, which likely hasn't changed since the last time sway received presentation feedback.

YaLTeR:
Rebased and updated as per comments. Thus far testing showed it working as intended both on the DRM backend and on the Wayland backend (although that one needs https://github.com/swaywm/wlroots/pull/1916 for 100% proper testing).

YaLTeR:
> (although that one needs swaywm/wlroots#1916 for 100% proper testing)

Tested using emersion's patches, everything seems to work as intended.

include/sway/config.h
```diff
@@ -203,6 +203,7 @@ struct output_config {
 	float scale;
 	int32_t transform;
 	enum wl_output_subpixel subpixel;
+	int max_render_time;
```
emersion: Maybe add a `// ms` comment to make it clear what the unit is?

include/sway/tree/view.h
```diff
@@ -108,6 +108,8 @@ struct sway_view {
 	} events;
 
 	struct wl_listener surface_new_subsurface;
+
+	int max_render_time;
```
emersion: Maybe add a `// msec` comment to make it clear what the unit is?

sway/desktop/output.c
```diff
@@ -524,18 +548,84 @@ static void damage_handle_frame(struct wl_listener *listener, void *data) {
...
+	if (delay < 1) {
+		output_repaint_timer_handler(output);
+	} else {
+		output->wlr_output->block_idle_frame = true;
+		wl_event_source_timer_update(output->repaint_timer, delay);
+	}
+
 	// Send frame done to all visible surfaces
-	send_frame_done(output, &now);
+	struct send_frame_done_data data;
```
emersion: Can we zero-initialize this, just in case we add a new member? (`= {0}`)

include/sway/output.h
```diff
@@ -51,6 +51,11 @@ struct sway_output {
 	struct {
 		struct wl_signal destroy;
 	} events;
+
+	struct timespec last_presentation;
+	uint32_t refresh_nsec;
+	int max_render_time;
```
emersion: `// msec`

sway/desktop/output.c
```diff
@@ -524,18 +548,84 @@ static void damage_handle_frame(struct wl_listener *listener, void *data) {
...
+	}
+
+	int delay = msec_until_refresh - output->max_render_time;
+
+	// If the delay is less than 1 millisecond (which is the least we can wait)
+	// then just render right away.
+	if (delay < 1) {
+		output_repaint_timer_handler(output);
+	} else {
+		output->wlr_output->block_idle_frame = true;
```
emersion: Instead of having this workaround in wlroots, I'd be more comfortable having it in Sway only, until we find a better solution. Do you think adding a `sway_output.block_frame_events` field and a `if (output->block_frame_events) { return; }` to the frame handler would be equivalent?

emersion:
Build failure is unrelated. Opened a PR to fix it: https://github.com/swaywm/sway/pull/4733

sway/desktop/output.c
YaLTeR: I can't say if it would be equivalent from just looking at the code, since `block_idle_frame` actually blocks the call to backend's `schedule_frame`, which contains some non-trivial code in case of DRM and commits the surface right away in case of Wayland.

emersion: I think this sequence of events happen:

* We get a frame event. We setup our repaint timer.
* A surface commits, `wlr_output_damage` schedules a frame. This sets `wlr_output.frame_pending` to true.
* Our frame timer fires, but a frame is already pending.

I think moving `block_idle_frame` into `wlr_output_damage` would be better, since it's the helper scheduling frame events behind our backs. This will help reducing the complexity added to the already-complex `wlr_output` code.

YaLTeR: > I think moving `block_idle_frame` into `wlr_output_damage` would be better, since it's the helper scheduling frame events behind our backs. This will help reducing the complexity added to the already-complex `wlr_output` code.

I'm worried about the following sequence of events:
1. A surface commits, a frame is scheduled using the idle timer.
2. Immediately after we get a frame event which sets up the repaint timer.
3. The idle frame renders and breaks it.

Or is the 1-2 transition like this impossible? As in, can there be a situation where a surface commit is directly followed by a frame event, without idle timers getting the chance to run in the meantime?

emersion: >I think moving block_idle_frame into wlr_output_damage would be better, since it's the helper scheduling frame events behind our backs.

Actually, this isn't quite right. One other wlroots module calls `wlr_output_schedule_frame`: the screencopy implementation. `wlr_output.block_idle_frame` is a way for Sway to say to wlroots "don't worry, I'll send a frame shortly".

Well, let's just merge this and I'll open an issue about all of these problems. The `wlr_output` interface hasn't been designed with repaint scheduling in mind.

>I'm worried about the following sequence of events

The 1-2 transition is impossible because if a frame event is pending (e.g. because we've swapped the output buffers) then `wlr_output_schedule_frame` is a no-op.

emersion:
Thanks!

PR #4589, opened 2019-09-26
Fix direct scan-out flickering

emersion:
Sometimes when using direct scan-out, some flickering between the
fullscreen app and the regular desktop could be seen.

This happened because we called wlr_output_attach_render and then
wlr_output_attach_buffer for direct scan-out. wlr_output_attach_render
makes the OpenGL context current but also attaches the OpenGL buffer to
the primary plane apparently (all of this happens inside
eglMakeCurrent).

This patch moves the scan-out logic outside of output_render, before
wlr_output_attach_render. This lines it up with rootston's
implementation. This also makes more sense since no rendering is
involved when using direct scan-out.

Sorry about that, I should've tested this with more clients. The new
code has been tested with mpv and a GLFW demo.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -412,6 +412,72 @@ static void send_frame_done(struct sway_output *output, struct timespec *when) {
 	output_for_each_surface(output, send_frame_done_iterator, when);
 }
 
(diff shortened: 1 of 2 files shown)


ddevault:
Thanks!

PR #4593, opened 2019-09-27
xdg_shell: remember size supplied by surface

versusvoid:
Evince sends new (cached?) size when opening document from recents. It hits `unexpected new size handling` for floating and fullscreen views in xdg_shell and produces weird results for tiled view. New size isn't stored in window properties and old size is sent in next `configure` (for example, when focus leaves the window), fixing the glitched state

I see three possible fixes sway could implement:
1) Immediately resend configure hoping view will be happy with it (presented here)
2) Save requested size and relayout workspace (I don't know how sway handles tiled windows refusing to resize)
3) Make window floating

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -290,6 +290,7 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 			desktop_damage_view(view);
 			view_update_size(view, new_geo.width, new_geo.height);
 			memcpy(&view->geometry, &new_geo, sizeof(struct wlr_box));
+			wlr_xdg_toplevel_set_size(xdg_surface, new_geo.width, new_geo.height);
 			desktop_damage_view(view);
 			transaction_commit_dirty();
 		} else {


sway/desktop/xdg_shell.c
```diff
@@ -269,6 +269,20 @@ static const struct sway_view_impl view_impl = {
...
+static void handle_unexpected_new_size(struct sway_view *view,
+		struct sway_container* con, struct wlr_box *new_geo) {
+	if (container_is_floating(con) || container_is_fullscreen_or_child(con)) {
+		desktop_damage_view(view);
+		view_update_size(view, new_geo->width, new_geo->height);
+		memcpy(&view->geometry, new_geo, sizeof(struct wlr_box));
+		desktop_damage_view(view);
+		transaction_commit_dirty();
+	} else {
+		wlr_xdg_toplevel_set_size(view->wlr_xdg_surface,
```
emersion: I'm worried about an infinite loop here

emersion:
Shouldn't this be fixed in evince?

versusvoid:
I expected you to say that =)
Yeah, it should. But I think sway also should behave consistently: either accept view size and resize window (somehow) or force calculated size for window.

Now it accepts view size, doesn't change the window which looks glitchy, but forces old calculated size after focus change or any other event triggering xdg_shell `configure` event.

I'm also worried about infinite loop, so maybe making view float would be a better solution. User then can tile it again manually.

emersion:
When the client doesn't obey the size, I think the best we can do is to just center it as we do currently. I don't want to add workarounds for clients that don't obey, but if you insist end up using the right size.

versusvoid:
@emersion now sway will remember and use surface's geometry until container's size changed

emersion:
I don't understand why this change is necessary.

emersion:
Without this change, we always send `configure` with the available space and we always center the client in case its buffer is smaller. I'd don't understand how configuring the client with the size it provided would be better.

versusvoid:
It would break loop in first scenario.

PR #4602, opened 2019-10-02
Introduce pid criteria token

RyanDwyer:
One of my barriers to switching to a Wayland native terminal emulator is the lack of urgency support. I have my shell configured to echo a bell character when commands are completed, which makes most X terminal emulators set the window to urgent. This is generally not possible in Wayland due to no native urgency support.

With this patch I can apply this behaviour in a Wayland native terminal using the following in my `~/.zshrc`:

    precmd() {
        swaymsg "[pid=$PPID] urgent enable"
    }

So when a shell command completes, zsh runs that swaymsg command with the PID of the terminal emulator. Sway then sets the terminal emulator to urgent.

You can test it by putting the above in your zsh config, then run `sleep 1` in a Wayland native terminal and focus away before the sleep completes. This assumes your terminal emulator uses one process per window.

### DIFF
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -46,6 +46,7 @@ struct criteria {
 	bool tiling;
 	char urgent; // 'l' for latest or 'o' for oldest
 	struct pattern *workspace;
+	pid_t pid;
 };
 
 bool criteria_is_empty(struct criteria *criteria);
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -31,7 +31,8 @@ bool criteria_is_empty(struct criteria *criteria) {
 		&& !criteria->floating
 		&& !criteria->tiling
 		&& !criteria->urgent
-		&& !criteria->workspace;
+		&& !criteria->workspace
+		&& !criteria->pid;
 }
 
 // The error pointer is used for parsing functions, and saves having to pass it
@@ -370,6 +371,12 @@ static bool criteria_matches_view(struct criteria *criteria,
 		}
 	}
 
+	if (criteria->pid) {
+		if (criteria->pid != view->pid) {
+			return false;
+		}
+	}
+
 	return true;
 }
 
@@ -458,6 +465,7 @@ enum criteria_token {
 	T_TITLE,
 	T_URGENT,
 	T_WORKSPACE,
+	T_PID,
 
 	T_INVALID,
 };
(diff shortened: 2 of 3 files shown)


RedSoxFan:
Sounds reasonable to me. I think `__focused__` would make sense to implement for `pid` criteria to allow applying to all windows in the same process.

Not sure if this has been discussed for i3 (can look later), but I'm fine with merging this even if i3 is not interested.

ddevault:
Hm, the feature itself seems fine but for the stated problem can't we add an urgency hit to xdg-shell?

norcalli:
Does this mean you could do `for_window [pid=$PID] ...`?

RedSoxFan:
> Does this mean you could do `for_window [pid=$PID] ...`?

Yes, it would be able to be used wherever criteria is allowed

norcalli:
Once a pid has been used, though, in the `for_window` context, it seems like it would make sense to remove it, since it could/should never be triggered again, right?

I could see this being very useful, regardless, for the launcher script I use. Currently, I have to run a subscribe loop on window events and check everything myself, but I could potentially just do a `for_window` once I've launched it. Then again, I could also see it being useful to target the process group id in case a command launches a subprocess with a UI...

vilhalmer:
It could technically be reused for the life of the program, rather than adding a new one each time. Perhaps a `--once` flag could be added to specify whether the criteria should vanish after the first application? I've wished for this flag in other contexts before.

tadeokondrak:
Another usecase: switching between Kakoune editor windows on X11 [uses xdotool to focus its own window by ID](https://github.com/mawww/kakoune/blob/5239288f57907a3ae7458c0b3a680546d079878c/rc/windowing/x11.kak#L61), I don't see a way to replicate this with sway right now.

ammgws:
@tadeokondrak Does the pid show up in get_tree? If so then can't you use swaymsg to focus the window using the id of that same node?

tadeokondrak:
@ammgws Thanks, I'll use this. Having this feature would still be a nice improvement over checking the tree, though.

ddevault:
This change is in good shape once the merge conflicts are resolved. However, I would like to see a proposal for urgency in xdg shell posted to wayland-devel before this is merged.

emersion:
The proposal has already been sent: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/9

ddevault:
Do you want to explicitly add "It is not implied that the compositor would raise the window or give it focus when the urgency hint is given", then bump it for a vote?

RyanDwyer:
Rebased.

ddevault:
Thanks!

PR #4627, opened 2019-10-12
criteria: make literal comparison for __focused__ values

RPigott:
This is another fix for #4586 which replaces #4587.

This time, I have modified criteria to make a literal comparison rather than a regex comparison when the attribute value is `__focused__`. The behavior should now better match [i3's logic](https://github.com/i3/i3/blob/7db0d179a3aa659f1dc109012d253403094c72a2/src/match.c) .

previously, the criteria->autofail field was used to disqualify criteria that had a `__focused__` attribute value, but had no focused container at the time of parsing. Criteria now compare against the focused container at match time, like i3 does, so the field is no longer necessary, but criteria should have the same matching behavior as before.

con_id never made a regex comparison in the first place, so I kept the logic the same. When there is no focused container, criteria->con_id will be 0 and always fail, like before.

### DIFF
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -14,29 +14,38 @@ enum criteria_type {
 	CT_NO_FOCUS                = 1 << 4,
 };
 
+enum pattern_type {
+	PATTERN_PCRE,
+	PATTERN_FOCUSED,
+};
+
+struct pattern {
+	enum pattern_type match_type;
+	pcre *regex;
+};
+
 struct criteria {
 	enum criteria_type type;
 	char *raw; // entire criteria string (for logging)
 	char *cmdlist;
 	char *target; // workspace or output name for `assign` criteria
 
-	bool autofail; // __focused__ while no focus or n/a for focused view
-	pcre *title;
-	pcre *shell;
-	pcre *app_id;
-	pcre *con_mark;
+	struct pattern *title;
+	struct pattern *shell;
+	struct pattern *app_id;
+	struct pattern *con_mark;
 	uint32_t con_id; // internal ID
 #if HAVE_XWAYLAND
-	pcre *class;
+	struct pattern *class;
 	uint32_t id; // X11 window ID
-	pcre *instance;
-	pcre *window_role;
+	struct pattern *instance;
+	struct pattern *window_role;
 	enum atom_name window_type;
 #endif
 	bool floating;
 	bool tiling;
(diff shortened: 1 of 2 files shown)


include/sway/criteria.h
```diff
@@ -14,29 +14,33 @@ enum criteria_type {
 	CT_NO_FOCUS                = 1 << 4,
 };
 
+struct pattern {
+	char *literal;
```
ddevault: Can we instead make this an enum?

```c
enum pattern_type {
    PATTERN_PCRE,
    PATTERN_FOCUSED,
    // etc
}
```

RPigott: Yeah that does sound better. I was trying a little too hard to match the i3 code.

I've switched it to an enum, PTAL.

sway/criteria.c
```diff
@@ -99,36 +146,75 @@ static void find_urgent_iterator(struct sway_container *con, void *data) {
...
 
 	if (criteria->title) {
 		const char *title = view_get_title(view);
-		if (!title || regex_cmp(title, criteria->title) != 0) {
+		if (!title) {
 			return false;
 		}
+
+		switch (criteria->title->match_type) {
+			case PATTERN_FOCUSED:
```
ddevault: style: align case with switch

sway/criteria.c
```diff
@@ -99,36 +146,75 @@ static void find_urgent_iterator(struct sway_container *con, void *data) {
...
 
 	if (criteria->shell) {
 		const char *shell = view_get_shell(view);
-		if (!shell || regex_cmp(shell, criteria->shell) != 0) {
+		if (!shell) {
 			return false;
 		}
+
+		switch (criteria->shell->match_type) {
+			case PATTERN_FOCUSED:
```
ddevault: ditto

sway/criteria.c
```diff
@@ -99,36 +146,75 @@ static void find_urgent_iterator(struct sway_container *con, void *data) {
...
 
 	if (criteria->app_id) {
 		const char *app_id = view_get_app_id(view);
-		if (!app_id || regex_cmp(app_id, criteria->app_id) != 0) {
+		if (!app_id) {
 			return false;
 		}
+
+		switch (criteria->app_id->match_type) {
+			case PATTERN_FOCUSED:
```
ddevault: .

RPigott:
Fixed switch/case style.

ddevault:
Thanks!

Issue #4631, opened 2019-10-12
Version number is unreliable


emersion:
The version we get from Git is unreliable. At the time of writing it says 1.2-rc1 although the latest release is 1.2. This happens because 1.2 has been tagged on the v1.2 branch, not on master. When building from master the latest tag is 1.2-rc1.

Not sure how to fix this.

dnkl:
Merging the 1.2-branch into master will make the 1.2 tag "visible" in master too.

PR #4634, opened 2019-10-14
build: always use the project version

emersion:
Don't use the latest tag, always use the project version for the version
string. Because of version branches, getting the version from Git can be
unreliable.

Closes: https://github.com/swaywm/sway/issues/4631


ddevault:
Thanks!

PR #4646, opened 2019-10-17
view_from_wlr_xwayland_surface(): fix miscasting of types

a-m-joseph:
The existing code works by accident because the "view" field (of type
sway_view) just happens to be the first field in the
sway_xwayland_view struct.  Depending on that is error-prone; all it
takes is a simple reordering of fields in that struct to trigger a
nonlocal bug.  Adding the correct casting sequence will ensure that
reodering fields won't cause problems.

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -574,7 +574,7 @@ static void handle_set_hints(struct wl_listener *listener, void *data) {
 
 struct sway_view *view_from_wlr_xwayland_surface(
 		struct wlr_xwayland_surface *xsurface) {
-	return xsurface->data;
+	return &((struct sway_xwayland_view*)xsurface->data)->view;
 }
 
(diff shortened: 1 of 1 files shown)


emersion:
>all it takes is a simple reordering of fields in that struct to trigger a nonlocal bug

This won't happen, the xwayland view struct extends the generic view struct.

a-m-joseph:
> > all it takes is a simple reordering of fields in that struct to trigger a nonlocal bug
> 
> This won't happen

I'm sure somebody thought something similar about the issue that causes #4648.

This PR is basic defensive programming.  If defensive programming isn't "sway style" I won't waste anybody's time arguing further.

ddevault:
It's not necessary to be abrasive. This is a pattern we use extensively throughout sway and wlroots and is guaranteed to meet our expectations per the C specification. A "simple reordering of the fields" would be an egregious error and easily caught in code review.

PR #4648, opened 2019-10-17
xwayland.c handle_map(): NULL out xsurface->data() to prevent crashing.

### LINKED ISSUE #4647: xwayland.c handle_map() leaves wlr_xwayland_surface.data pointing at garbage, crashes sway
Pull request to fix the bug will follow shortly.

When changing a surface from managed to unmanaged in handle_map(), the call to handle_destroy(.., view) causes the sway_xwayland_view pointed to by the untyped wlr_xwayland_surface.data field to become invalid garbage, yet the untyped wlr_xwayland_surface.data continues to point at it.  In particular: view_get_*(view_from_wlr_surface(..)), even with appropriate NULL checking, will crash sway when this codepath is exercised (reliable test case: drop-down menus in Google Earth).

The validity condition of the wlr_xwayland_surface.data field is very, very error-prone and undocumented.  This bug will probably be inadvertently reintroduced in the future.  To prevent that, I suggest simplifying the condition to "in sway, wlr_xwayland_surface.data always points to a valid sway_xwayland_view" and synthesizing a fake/no-op sway_xwayland_view for "unmapped" surfaces.

a-m-joseph:
Fixes #4647

When changing a surface from managed to unmanaged in handle_map(), the
call to handle_destroy(.., view) causes the sway_xwayland_view pointed
to by the untyped wlr_xwayland_surface.data field to become invalid
garbage, yet the untyped wlr_xwayland_surface.data continues to point
at it.  In particular: view_get_*(view_from_wlr_surface(..)), even
with appropriate NULL checking, will crash sway when this codepath is
exercised (reliable test case: drop-down menus in Google Earth).

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -401,6 +401,7 @@ static void handle_map(struct wl_listener *listener, void *data) {
 		// This window used not to have the override redirect flag and has it
 		// now. Switch to unmanaged.
 		handle_destroy(&xwayland_view->destroy, view);
+		xsurface->data = NULL;
 		struct sway_xwayland_unmanaged *unmanaged = create_unmanaged(xsurface);
 		unmanaged_handle_map(&unmanaged->map, xsurface);
 		return;


emersion:
Shouldn't `handle_destroy` do this?

emersion:
Hmm, maybe it's clearer that way, because `handle_destroy` may assume the xsurface will be destroyed (making it unnecessary to reset the field). I was just worried about other callers forgetting to do it.

a-m-joseph:
I can see a lot of ways to fix this.  My PR was not necessarily the best fix -- just the shortest one.

PR #4651, opened 2019-10-18
Support man page localization

krayfaus:
This pull request updates meson.build to:
* build and install localized man pages.

Notes:
Current scdoc files were moved to ease
the management of locales in the future.
The new doc directory in sway, swaybar,
swaymsg and swaynag shall have a similar
structure to that of man: where English
files are kept in root dir and localized
ones are nested in their respective dir.


meson.build
```diff
@@ -96,33 +96,62 @@ if scdoc.found()
...
+			'sway.5.scd',
+			'sway-bar.5.scd',
+			'sway-input.5.scd',
+			'sway-ipc.7.scd',
+			'sway-output.5.scd'
+		],
+		'swaybar': [
+			'swaybar-protocol.7.scd'
+		],
+		'swaymgs': [
```
ascent12: Typo: `swaymsg`

meson.build
```diff
@@ -96,33 +96,62 @@ if scdoc.found()
...
+				endif
+
+				foreach filename: sources
+					input = join_paths(projdir, filename)
+					if run_command('[', '-e', input, ']').returncode() == 0
+						topic = input.split('.')[-3].split('/')[-1]
+						section = input.split('.')[-2]
+						output = '@0@/@1@.@2@'.format(outdir, topic, section)
+						cmd = '@0@ < @1@ > @2@'.format(scdoc_prog.path(), input, output)
+						if run_command(sh, '-c', cmd).returncode() == 0
```
ascent12: This should be using custom_target rather than run_command. We want to actually build things at build time, not configure time.

meson.build
```diff
@@ -96,33 +96,62 @@ if scdoc.found()
...
+				projdir = join_paths(meson.source_root(), proj, 'doc')
+				outdir = join_paths(meson.build_root(), 'doc')
+			else
+				docdir = join_paths(mandir, locale)
+				projdir = join_paths(meson.source_root(), proj, 'doc', locale)
+				outdir = join_paths(meson.build_root(), 'doc', locale)
+			endif
+
+			# Checks if locale dir exists before proceeding with scdoc compilation
+			if run_command('[', '-d', projdir, ']').returncode() == 0
```
ascent12: I really don't like the approach of using shell commands to do this. Meson cannot track dependencies, and can lead to dirty builds and otherwise not building things properly if files are changed.
There is a reason all of the source files are listed explicitly instead of using things like globbing.

ddevault:
Can you translate one or two man pages into another language (one you're fluent in, no google translate) to prove that the changes are suitable for a future multi-lingual sway manual? Also, I'd like to see a meson option for enabling or disabling specific locales for generation/installation.

ddevault:
The commit history will also need to be cleaned up before this can be merged.

meson.build
krayfaus: I've actually tried to use meson's custom_target interface but it doesn't support generating multiple targets with the same name, and the localized man pages need to have the same name as the default one. Still I'll try to workaround this issue, for a cleaner approach.

krayfaus:
I'm currently translating some docs to Portuguese  but thought it would be appropriate not to bloat this pull request, I'll make another branch with my translation in my personal sway repo where the changes can be tested, is that good?

About the commit history, is it needed to create an orphan branch to pull or just one clean commit suffices the problem?

ddevault:
It wouldn't be bloat, it's an essential part of proving that these meson changes work and should be part of the patchset. We generally shy away from changing the code for the code's sake, it's best if we can prove the correctness of the approach with an actual use-case.

krayfaus:
Sorry for leaving this unattended for so long.
Due to personal reasons I didn't have time nor energy to continue this.
Good luck!

PR #4654, opened 2019-10-19
Various small fixes

lbonn:
Some memory leaks in more or less pathological code paths, some simplifications and defensive null checks.

Most of these were found by running clang-tidy/analyzer.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -198,9 +198,6 @@ static void set_config_node(struct sway_node *node) {
 
 list_t *execute_command(char *_exec, struct sway_seat *seat,
 		struct sway_container *con) {
-	list_t *res_list = create_list();
-	char *exec = strdup(_exec);
-	char *head = exec;
 	char *cmd;
 	char matched_delim = ';';
 	list_t *views = NULL;
@@ -213,9 +210,16 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 		}
 	}
 
+	char *exec = strdup(_exec);
+	char *head = exec;
+	list_t *res_list = create_list();
+
+	if (!res_list || !exec) {
+		return NULL;
+	}
+
 	config->handler_context.seat = seat;
 
-	head = exec;
 	do {
 		for (; isspace(*head); ++head) {}
 		// Extract criteria (valid for this command list only).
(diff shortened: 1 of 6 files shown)


sway/commands.c
```diff
@@ -209,6 +209,7 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 		// passing a NULL seat means we just pick the default seat
 		seat = input_manager_get_default_seat();
 		if (!sway_assert(seat, "could not find a seat to run the command on")) {
+			free(exec);
```
emersion: Can we move the `exec` declaration after this `return` instead?

Same for `res_list`?

sway/criteria.c
```diff
@@ -376,31 +376,31 @@ static char *get_focused_prop(enum criteria_token token, bool *autofail) {
...
 	struct sway_view *view = focus ? focus->view : NULL;
-	const char *value = NULL;
+	char *value = NULL;
 
 	switch (token) {
 	case T_APP_ID:
 		*autofail = true;
 		if (view) {
-			value = view_get_app_id(view);
+			value = strdup(view_get_app_id(view));
```
emersion: If `view_get_app_id` can return NULL, it will segfault.

sway/main.c
```diff
@@ -259,6 +259,9 @@ int main(int argc, char **argv) {
 			exit(EXIT_SUCCESS);
 			break;
 		case 'c': // config
+			if (config_path) {
+				free(config_path);
```
emersion: No need for the `if`, `free(NULL)` is a no-op.

sway/criteria.c
lbonn: Indeed! This one is a bit akward to fix without breaking the pattern of other cases. Please let me know what you think of the new solution.

sway/commands.c
```diff
@@ -213,9 +210,16 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 		}
 	}
 
+	char *exec = strdup(_exec);
+	char *head = exec;
+	list_t *res_list = create_list();
+
+	if (!sway_assert(res_list && exec, "allocation failure")) {
```
emersion: This shouldn't be an assertion. Allocation failures shouldn't make the process abort in debug mode.

ddevault:
Thank you!

PR #4657, opened 2019-10-20
Fix refresh rate scale of output

DanySpin97:
When applying config, value mode->refresh is mHz; convert it to Hz before
assigning it to the temporary output config. oc->refresh_rate will
be converted back to mHz in set_mode function.

Fix debug log printing GHz instead of Hz.

Based on #4577, fixes https://github.com/emersion/kanshi/issues/51 .

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -260,7 +260,7 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 
 	bool modeset_success;
 	if (oc && oc->width > 0 && oc->height > 0) {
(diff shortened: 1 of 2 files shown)


sway/config/output.c
```diff
@@ -329,7 +329,7 @@ static void default_output_config(struct output_config *oc,
 			wl_container_of(wlr_output->modes.prev, mode, link);
 		oc->width = mode->width;
 		oc->height = mode->height;
-		oc->refresh_rate = mode->refresh;
+		oc->refresh_rate = mode->refresh / 1000;
```
emersion: All of these are integers, should probably convert either `mode->refresh` or `1000` to float.

DanySpin97: Fixed.

PR #4668, opened 2019-10-23
Fix tiled containers resize with mouse

### LINKED ISSUE #4659: Resize of tiled containters with mouse doesn't work anymore
Previously(sway 1.2) I was able to resize tiled windows with mouse when moving mouse to a window border, holding LMB and dragging the border. Window swapping works fine.

sway version 1.2-3ee3a9ef (Oct 18 2019, branch 'master')

If you need more info let me know.

kchibisov:
Fixes regression introduced in 2c1a11016ccb7646a74de51eff003c18e5aa7902

Fixes #4659

cc @versusvoid

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -59,7 +59,7 @@ static bool edge_is_external(struct sway_container *cont, enum wlr_edges edge) {
 
 static enum wlr_edges find_edge(struct sway_container *cont,
 		struct wlr_surface *surface, struct sway_cursor *cursor) {
-	if (!cont->view || cont->view->surface != surface) {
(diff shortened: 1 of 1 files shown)


emersion:
@RyanDwyer Do you remember the details?

RyanDwyer:
@versusvoid You're right. The function call could be avoided if `surface` is `NULL`. However `find_edge` is called three times, so it's a weigh up between a) wrapping each invocation in an `if (!surface)` check (ie. code duplication) and removing `surface` from `find_edge`'s signature, or b) changing the first condition within `find_edge` to be `if (!cont->view || surface)` and keeping `surface` in the signature. I don't have a preference, so I'll leave that choice to you.

Issue #4638, opened 2019-10-15
Port new i3 focus changes (v4.18?)


lbonn:
i3 has gained new focus commands here https://github.com/i3/i3/pull/3407.

In the spirit of #2336 and #3082, here is a list of tasks:

- [x] `focus_wrapping workspace`
- [x] `focus next|prev` (check that behavior matches)
- [x] `focus next|prev sibling`

loserMcloser:
Upvote for "focus next|prev" --- I was just coming here to make a feature request for that, and here it was as the most recently opened issue!

PR #4671, opened 2019-10-24
Support focus next|prev [sibling] i3 commands

RPigott:
Support `focus prev|next [sibling]` from i3 -- steps 2+3 for #4638 .

PTAL. If the behavior does not match i3, let me know.


ddevault:
Nice, thanks!

PR #4673, opened 2019-10-26
Fix output manager segfault

emersion:
See individual commit messages.

Closes: https://github.com/swaywm/wlroots/issues/1867


ddevault:
Thanks!

Issue #4679, opened 2019-10-28
broken scroll_factor behavior in sway and xwayland clients


RPigott:
* Sway Version: sway version 1.2-1a253ca7 (Oct 28 2019, branch 'master')

So this is an interesting issue with discrete scroll events. Sway introduced a scroll_factor feature with #3018 that allows the user to scale scroll events to control scroll sensitivity for their scrolling devices. 

At the moment, scroll_factor scales the axis value and rounds the axis_discrete * scroll_factor product to the nearest integer. I understand the behavior is application dependent, but in my observation xwayland clients that implement a scroll tab/slot behavior will skip a tab/slot when a scroll_factor > 1.5 causes sway to emit a discrete value greater than 1, and wayland clients tend to work fine.

You can observe in firefox, with the toolkit.tabbox.switchByScrolling pref enabled, the correct behavior on the wayland backend and the skipping behavior on the x11 backend. Swaybar is a wayland client that will not skip, chromium is an xwayland client that will skip.

Sway itself also exhibits the skipping behavior when scrolling to switch windows with scroll_factor in tabbed/stacked layouts. scroll_factor <= 0.5 makes scrolling to switch windows impossible altogether.

I'm not sure how universal the behavior is, so I'd be interested to hear about clients that do not follow the trend.

I don't know why xwayland clients behave this way, but I'm thinking we should just not scale the axis discrete event. It's an easy change, fixes those skipping issues and doesn't appear to have any ill effects on native wayland clients. I've observed in chromium not scaling axis discrete causes scroll_factor to have no effect for my mouse, but of course it still works to scale scroll events from my touchpad, which doesn't emit discrete events.

I understand there is a [new scrolling api in the works](https://gitlab.freedesktop.org/libinput/libinput/merge_requests/139) that should land in libinput soon, and [proposed parallel changes to wl_pointer](https://lists.freedesktop.org/archives/wayland-devel/2019-April/040377.html) which I expect is relevant, but I'm not quite sure how it fits in.

Thoughts on not scaling axis_discrete? Anyone know why only xwayland clients appear to be buggy?

sammoth:
[...]

PR #4680, opened 2019-10-28
Do not apply scroll_factor for switching windows in tabbed layout.

RPigott:
Partial fix for #4679. Only removes discrete scaling for switching windows in tabbed/stacked layouts, and leaves the axis_discrete scaling for clients alone.

I figure this should be changed anyway, as switching windows by scrolling a touchpad that doesn't emit axis_discrete doesn't work at all in the first place, so the feature is just broken in all cases with scroll_factor.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -546,7 +546,7 @@ static void handle_axis(struct sway_seat *seat,
 				seat_get_active_tiling_child(seat, tabcontainer);
 			list_t *siblings = container_get_siblings(cont);
 			int desired = list_find(siblings, active->sway_container) +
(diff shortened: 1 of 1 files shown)


RPigott:
Closes #4679.

RPigott:
Ping @emersion, or any project member. I guess this one got lost by the wayside last year.

emersion:
I'm not convinced this is a good fix, and I'd like `scroll_factor`'s behavior to be consistent everywhere.

asegarra:
just wanted to respectfully ask reconsideration for this PR... e.g. Chromium tab scrolling works even with a scroll_factor of 10, not changing this behavior effectively means setting scroll_factor breaks tabbed layout scrolling.

PR #4681, opened 2019-10-29
Implement input map_to_region command

tadeokondrak:
This has been in the documentation for a while but never got implemented.

### DIFF
--- a/sway/commands/input/map_to_region.c
+++ b/sway/commands/input/map_to_region.c
@@ -15,9 +15,7 @@ struct cmd_results *input_cmd_map_to_region(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "No input device defined");
 	}
 
-	// This is used to clear the current output mapping.
-	ic->mapped_to_output = strdup("");
-
+	ic->mapped_to = MAPPED_TO_REGION;
 	ic->mapped_to_region = calloc(1, sizeof(struct wlr_box));
 
 	const char *errstr;
--- a/sway/commands/input/map_to_region.c
+++ b/sway/commands/input/map_to_region.c
@@ -1,0 +1,56 @@
+#define _POSIX_C_SOURCE 200809L
+#include <stdlib.h>
+#include <string.h>
+#include <wlr/types/wlr_box.h>
+#include "sway/commands.h"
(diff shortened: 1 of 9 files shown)


sway/sway-input.5.scd
```diff
@@ -88,7 +88,7 @@ The following commands may only be used in the configuration file.
 	Maps inputs from this device to the specified output. Only meaningful if the
 	device is a pointer, touch, or drawing tablet device.
 
-*input* <identifier> map_to_region <WxH@X,Y>
+*input* <identifier> map_to_region <X> <Y> <width> <height>
```
tadeokondrak: It seems like you can't have commas in command invocations, so I had to change this. It's similar to `grim`'s output format, but with spaces as all separators.

sway/commands/input/map_to_region.c
```diff
@@ -0,0 +1,56 @@
...
+	if ((error = checkarg(argc, "map_to_region", EXPECTED_EQUAL_TO, 4))) {
+		return error;
+	}
+	struct input_config *ic = config->handler_context.input_config;
+	if (!ic) {
+		return cmd_results_new(CMD_FAILURE, "No input device defined");
+	}
+
+	// This is used to clear the current output mapping.
+	ic->mapped_to_output = strdup("");
```
ddevault: This is weird. Can we set it to NULL and then update other code to not segfault? Do we have to free the old string?

sway/input/seat.c
```diff
@@ -576,19 +576,39 @@ static void seat_reset_input_config(struct sway_seat *seat,
...
+	struct wlr_box *mapped_to_region = NULL;
 
 	struct input_config *ic = input_device_get_config(
 			sway_device->input_device);
 	if (ic != NULL) {
 		sway_log(SWAY_DEBUG, "Applying input config to %s",
 			sway_device->input_device->identifier);
 
+		// We use an empty string as a marker to clear the mapped_to_output
+		// property, because a NULL set in a handler_context isn't preserved.
```
ddevault: Oh. This is still weird, though. Can it be fixed?

ddevault:
This looks pretty good.

Issue #4677, opened 2019-10-27
Sway static compilation failure on set_cloexec


sheenobu:
* Sway Version: master 
* Wlroots version: master

Compiling sway in static mode fails due to conflicting symbols exposed by both wlroots and sway. The symbol is set_cloexec. This function is duplicated a few times across both code bases.

Simple version of my build commands:

```
> meson --default-library static --buildtype release build
> ninja -C build
```

Full version of my build script: https://gist.github.com/sheenobu/bfa8f758878184a4c5b72066a938bd8f

Expected output: 

a binary with sway and wlroots statically linked together

Actual output: 

```
/usr/bin/ld: subprojects/wlroots/libwlroots.a(sockets.c.o): in function `set_cloexec':
sockets.c:(.text+0x0): multiple definition of `set_cloexec'; common/libsway-common.a(util.c.o):util.c:(.text+0x290): first defined here
```

I was able to fix this by inlining set_cloexec within include/util.h but i think fixing this in wlroots might be better? just make set_cloexec a wlroots util function that can be imported in?

HACK, don't use:

```
diff --git a/include/util.h b/include/util.h
index 6d9454e0..930162f0 100644
--- a/include/util.h
+++ b/include/util.h
@@ -1,9 +1,12 @@
 #ifndef _SWAY_UTIL_H
 #define _SWAY_UTIL_H
 
+#include <stdlib.h>
 #include <stdint.h>
 #include <stdbool.h>
 #include <wayland-server-protocol.h>
+#include <fcntl.h>
+#include "log.h"
 
 /**
  * Wrap i into the range [0, max[
@@ -32,6 +35,22 @@ float parse_float(const char *value);
 
 const char *sway_wl_output_subpixel_to_string(enum wl_output_subpixel subpixel);
 
-bool set_cloexec(int fd, bool cloexec);
+static inline bool set_cloexec(int fd, bool cloexec) {
+       int flags = fcntl(fd, F_GETFD);
+       if (flags == -1) {
+               sway_log_errno(SWAY_ERROR, "fcntl failed");
+               return false;
+       }
+       if (cloexec) {
+               flags = flags | FD_CLOEXEC;
+       } else {
+               flags = flags & ~FD_CLOEXEC;
+       }
+       if (fcntl(fd, F_SETFD, flags) == -1) {
+               sway_log_errno(SWAY_ERROR, "fcntl failed");
+               return false;
+       }
+       return true;
+}
```

emersion:
[...]

PR #4682, opened 2019-10-29
Rename symbol set_cloexec to sway_set_cloexec, remove duplicates.

sheenobu:
set_cloexec is defined by both sway and wlroots (and who-knows-else),
so rename the sway one for supporting static linkage. We also remove
the duplicate version of this in client/.

Fixes: https://github.com/swaywm/sway/issues/4677


ddevault:
Thanks!

PR #4690, opened 2019-10-31
Skip line continuation when it is a comment

### LINKED ISSUE #4585: Line continuation on comments breaks the next line
- sway v1.2

If you have `\` as the last character of a commented line, this seems to be joined internally with the next (possible not comented line), 

* Configuration File:

```
# Some commented line \
some command
```

The `some command` will not execute. Not sure this is also a problem on i3, but I can test it.

robertgzr:
Currently commented lines ending in the backslash character will be
concatenated with the following line.

```
# with this comment \
exec swaynag -m 'will not run'
```

This change modifies `getline_with_cont` to stop reading when the
initial character is a '#'.

Closes #4585


emersion:
Thanks!

PR #4699, opened 2019-11-04
seatop_move_floating: make container respect pointer constraint

### LINKED ISSUE #4133: Losing mouse grasp moving floating windows towards edges
* Sway Version: `sway version 1.1-rc1-3-g9dca18b6 (May  7 2019, branch 'master')`
* Debug Log: [sway.log](https://gist.github.com/l4rzy/74d0df4f91119d42d37ac478d44b159a)
* Configuration File: default

**Description**: Moving a floating window towards screen edges e.g. `URxvt` (XWayland) and `gtk3-demo` (Wayland) caused the window to lose its mouse grasp (tested with all 4 edges), though the window itself continued to move along that direction and it's no help trying to move back the window once I release the mouse. Interestingly, the moved window appeared to be rendered on another workspace through the workspace indicator, but it was actually not rendered on that workspace by switching to that workspace and check.

**Expectation**: The window should stop moving as the mouse stop. I'm fairly new to sway, so I'm not sure this is a bug or feature :), but as a sane expectation, this is weird.

RPigott:
Closes #4133

### DIFF
--- a/sway/input/seatop_move_floating.c
+++ b/sway/input/seatop_move_floating.c
@@ -6,6 +6,7 @@
 
 struct seatop_move_floating_event {
 	struct sway_container *con;
+	double dx, dy; // cursor offset in container
 };
 
 static void handle_button(struct sway_seat *seat, uint32_t time_msec,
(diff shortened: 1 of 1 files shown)


sway/input/seatop_move_floating.c
```diff
@@ -6,6 +6,7 @@
 
 struct seatop_move_floating_event {
 	struct sway_container *con;
+	double ox, oy; // cursor offset in container
```
emersion: Could add "when seatop started" to make it clear it's not the current offset. Also we generally use `ox`/`oy` for output-local coords, so maybe we should rename that to `cx`/`cy` (container) or `dx`/`dy` (delta).

RPigott: Well the whole point of the patch is that the offset stays the same, even if you try to move the container somewhere where the pointer is constrained, so I think it makes sense as is.

I see why `ox, oy` are bad names, noted. I hadn't made the connection to output-local coords. I guess I'll use dx/dy here then.

emersion: >Well the whole point of the patch is that the offset stays the same, even if you try to move the container somewhere where the pointer is constrained, so I think it makes sense as is.

Eh, indeed!

PR #4700, opened 2019-11-04
smart_borders: separate smartness from edge types

### LINKED ISSUE #4698: hide_edge_borders and smart_borders mutually exclusive
# Summary:

`smart_borders` overrides `hide_edge borders`.

I know the sway(5) man page says, `smart_borders on` is identical to `hide_edge_borders smart`, but that's not how it's implemented in i3(-gaps 4.17.1). Hence, I'm not sure if this is considered working as expected, so forgive me my naivety.

* Sway Version: 1.2

* Debug Log: *Don't think this is necessary here, as this seems a design decision. Will file subsequently if you insist.*

# Configuration File:

1. 
       hide_edge_borders vertical
       smart_borders on
       [...]
1. 
       smart_borders on
       hide_edge_borders vertical
       [...]

# Steps to reproduce:
  1. set config as above
  1. on a workspace with no containers, open one window/container
  1. open a second window/container
  1. swap `smart_borders` and `hide_edge_borders` lines as above
  1. reload config and repeat

# Expected result:
  With one container, no border is shown. With >1 containers, only the horizontal borders are shown.

# Actual result:
1. vertical edge borders aren't hidden with >1 container
2. horizontal borders are always shown, also with just one container

  Whatever of the two configs come first is overridden/ignored.

RPigott:
Closes #4698 ?

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -333,8 +333,12 @@ enum edge_border_types {
 	E_VERTICAL,     /**< hide vertical edge borders */
 	E_HORIZONTAL,   /**< hide horizontal edge borders */
 	E_BOTH,		/**< hide vertical and horizontal edge borders */
-	E_SMART, /**< hide both if precisely one window is present in workspace */
-	E_SMART_NO_GAPS, /**< hide both if one window and gaps to edge is zero */
+};
+
+enum edge_border_smart_types {
+	ESMART_OFF,
+	ESMART_ON, /**< hide edges if precisely one window is present in workspace */
+	ESMART_NO_GAPS, /**< hide edges if one window and gaps to edge is zero */
 };
 
 enum sway_popup_during_fullscreen {
@@ -509,7 +513,7 @@ struct sway_config {
 	int border_thickness;
 	int floating_border_thickness;
 	enum edge_border_types hide_edge_borders;
-	enum edge_border_types saved_edge_borders;
+	enum edge_border_smart_types hide_edge_borders_smart;
 	bool hide_lone_tab;
 
 	// border colors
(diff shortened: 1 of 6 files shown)


include/sway/config.h
```diff
@@ -333,8 +333,11 @@ enum edge_border_types {
 	E_VERTICAL,     /**< hide vertical edge borders */
 	E_HORIZONTAL,   /**< hide horizontal edge borders */
 	E_BOTH,		/**< hide vertical and horizontal edge borders */
-	E_SMART, /**< hide both if precisely one window is present in workspace */
-	E_SMART_NO_GAPS, /**< hide both if one window and gaps to edge is zero */
+};
+enum edge_border_smart_types {
```
RedSoxFan: Can you please add a blank line before this?

sway/commands/hide_edge_borders.c
```diff
@@ -32,14 +32,13 @@ struct cmd_results *cmd_hide_edge_borders(int argc, char **argv) {
 	} else if (strcmp(argv[0], "both") == 0) {
 		config->hide_edge_borders = E_BOTH;
 	} else if (strcmp(argv[0], "smart") == 0) {
-		config->hide_edge_borders = E_SMART;
+		config->hide_edge_borders_smart = E_SMART;
```
RedSoxFan: I think this and `smart_no_gaps` should also set `config->hide_edge_borders = E_NONE` as well. It appears that `hide_edge_borders smart|smart_no_gaps` is the same behavior as `hide_edge_borders none; smart_borders on|no_gaps`.

include/sway/config.h
```diff
@@ -333,8 +333,11 @@ enum edge_border_types {
...
+	E_DUMB,
```
RedSoxFan: nit: can you change the prefix of these so they differ from the other enum? Maybe `ESMART_{OFF,ON,NO_GAPS|`?

sway/commands/hide_edge_borders.c
RPigott: Makes sense. I put that equivalence in the doc.

RedSoxFan:
Thanks

PR #4702, opened 2019-11-05
Prefer higher refresh rate default modes

RPigott:


RPigott:
Oh the user can also just not specify a mode at all, not just omit refresh rate.

Should it still pick the last one in that case too, or try to find the "best" mode? Is just highest resolution -> highest refresh the "best" in that case?

emersion:
Screens indicate their "preferred" mode, which is the one we choose in case the user doesn't specify the mode explicitly.

RPigott:
I thought it might be something like that, but I wasn't sure where it would be documented.

Is there only one "preferred" mode? (the last one) Or are they sorted by preference?

Does it make sense to prefer higher resolution modes if the user selects a resolution that doesn't match the "preferred" mode?

emersion:
>Is there only one "preferred" mode? (the last one) Or are they sorted by preference?

I think there is at most one preferred mode reported by the kernel. `drm_info` will show this information. In practice I've always seen it being the last one in `wlr_output` (first in the kernel list), but I don't think this is ensured by the kernel. To prevent issues related to this: https://github.com/swaywm/sway/pull/4704

>Does it make sense to prefer higher refresh rate modes if the user selects a resolution that doesn't match the "preferred" mode?

Probably, but I'm not 100% sure. On laptops it might be better to choose a 60Hz mode for battery savings for instance. OTOH when the user doesn't specify anything they let Sway choose the mode, so we can do whatever we want. Choosing the mode with the highest refresh rate is probably the most sensible thing to do, and the user can override it anyway if necessary.

All in all I'm +1 for this PR. I just suspect it doesn't fix swaywm/wlroots#1887 (because the preferred mode is probably the last one → this isn't a bug).

RPigott:
> All in all I'm +1 for this PR. I just suspect it doesn't fix [swaywm/wlroots#1887](https://github.com/swaywm/wlroots/issues/1887) (because the preferred mode is probably the last one → this isn't a bug).

Yeah it definitely doesn't.

PR #4703, opened 2019-11-05
Do not always modeset when applying output config

### LINKED ISSUE #4687: Display configuration lost on sway reload
sway version 1.2, default configuration (plus a few extra key bindings), Fedora 30 (copr gumieri/sway)

I have a Lenovo X1 laptop with Intel HD Graphics 520. It only has enough bandwidth to drive two DP 4k monitors at 30Hz, plus its internal panel.

It's driving these outputs nicely:
```
Output eDP-1 'Unknown 0x183D 0x00000000'
  Current mode: 1920x1080 @ 60.049000 Hz
Output DP-5 'Samsung Electric Company U28E590 HTPJ203148'
  Current mode: 3840x2160 @ 30.000000 Hz
Output DP-6 'Unknown MAG321CURV DA2A019160072' (focused)
  Current mode: 3840x2160 @ 30.000000 Hz
```
If I reload (mod+Shift+c), sway defaults back to driving the monitors at 60Hz, which doesn't work.
There is no output configuration in my sway config file. That's what I want - I will configure via kanshi because I need hotplugging to work nicely.

Steps to reproduce:
```
$ swaymsg output DP-5 mode 3840x2160@30Hz
$ swaymsg output DP-6 mode 3840x2160@30Hz
```
everything looks good...
```
$ swaymsg reload
```
everything is broken.

In the log file below I start sway, run the two swaymsg output lines above, then reload sway.
`swaymsg -t get_outputs >/tmp/get_outputs.log` [get_outputs.log](https://github.com/swaywm/sway/files/3790835/get_outputs.log)
`sway -d 2> /tmp/sway-monitor-reset.log` [sway-monitor-reset.log](https://github.com/swaywm/sway/files/3790795/sway-monitor-reset.log)
`cat ~/.config/sway/config > config.txt` [config.txt](https://github.com/swaywm/sway/files/3790962/config.txt)

RPigott:
For most output config options, applying a config with the option unset
does not change the current value, but a modeset is always attempted
even if the output has no mode configured. This unexpectedly causes a
config reload to reset output modes, even if none are configured.

Change it to leave unconfigured outputs that already have a mode alone.

Fixes #4687

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -269,7 +269,7 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 			oc->height, oc->refresh_rate);
 		modeset_success = set_mode(wlr_output, oc->width, oc->height,
 			oc->refresh_rate, oc->custom_mode == 1);
-	} else if (!wl_list_empty(&wlr_output->modes)) {
+	} else if (!wl_list_empty(&wlr_output->modes) && !wlr_output->current_mode) {
 		struct wlr_output_mode *mode =
 			wl_container_of(wlr_output->modes.prev, mode, link);
 		modeset_success = wlr_output_set_mode(wlr_output, mode);


RPigott:
Ah I see now this behavior is pretty clearly intentional. When sway reloads the config, it makes sure to set the default values for outputs again, even if they are unconfigured in the config. Also means the patch does nothing for the linked issue.

I think we should change the man page to document reload behavior instead then. Right now it reads:
```
reload
   Reloads the sway config file and applies any changes.
```
As is, I wouldn't expect `swaymsg reload` to have any effect on outputs that have no configuration in the config file.

RPigott:
Well either way, this patch is wrong. Closing.

emersion:
>I think we should change the man page to document reload behavior instead then. 

Yes, that's a good point. We should definitely expand the `reload` description.

PR #4704, opened 2019-11-05
Use wlr_output_preferred_mode instead of the last mode

emersion:
Instead of relying on the order of modes, use wlr_output_preferred_mode
to get the preferred mode.


RedSoxFan:
Thanks

Issue #4705, opened 2019-11-05
Sway fails to start after the recent "prefer high refresh rate" commit


colemickens:
```diff
- rev = "3975ca28c2e870eb3f40bbd43a90354743f7ccf1";
-  sha256 = "01lhclh5l6shd0624m5dm3j1irbwwq3ziin5ki5kzzgraqbfqamp";
+  rev = "f33dcd4c604fa24058f6cbaf692831420dc5db67";
+  sha256 = "18xfwpgldn9psf6662anf9zq43b23zxgpb1r2xlp38hzjs3ca1qw";
```

Between there, Sway stopped being able to start with my external monitor attached. I think it's just the most recent commit f33dcd4c604fa24058f6cbaf692831420dc5db67.

[sway.log](https://github.com/swaywm/sway/files/3808884/sway.log)

emersion:
Is this segfaulting? Do you get a coredump?

colemickens:
No, it just fails to start. I normally just get a couple of lines of output, this is stdout and err mixed.

```
2019-11-05 04:11:15 - [backend/drm/atomic.c:57] eDP-1: Atomic commit failed (modeset): Invalid argument
```

eDP-1 is the internal panel, but I'm starting to see quite the pattern. I'm on the very latest kernel, latest DRM, nearly latest mesa. My impressions of Intel graphics support was higher, though maybe it's just Sway bugs, I don't really know much about this area.

Anyway, thanks. Let me know if I should do the usual drm debug dance and post that too.

emersion:
Ah, I missed the atomic commit failures.

Yeah, some DRM logs would be helpful.

JonnyMako:
I get a segfault at startup. Log attached.

[sway.log](https://github.com/swaywm/sway/files/3810022/sway.log)
[gdb.txt](https://github.com/swaywm/sway/files/3810043/gdb.txt)


```
[   74.962174] sway[1002]: segfault at 8 ip 000056060faae0ac sp 00007fffabdf4bc0 error 4 in sway[56060fa80000+5d000]
[   74.962179] Code: f0 8b 40 04 39 45 d0 75 36 48 8b 45 f0 8b 40 08 39 45 ec 75 0a 48 8b 45 f0 48 89 45 f8 eb 47 48 8b 45 f0 8b 50 08 48 8b 45 f8 <8b> 40 08 39 c2 7e 06 48 8b 45 f0 eb 04 48 8b 45 f8 48 89 45 f8 48
[...]

PR #4706, opened 2019-11-05
Fix segfault in set_mode

emersion:
best is NULL prior to being assigned to a mode.

Closes: https://github.com/swaywm/sway/issues/4705


JonnyMako:
works for me. thanks.

RedSoxFan:
Thanks

PR #4727, opened 2019-11-13
render: use nearset neighbor scaling for outputs with an integer scale

RPigott:
Proof of concept. See swaywm/wlroots#1914.


ddevault:
This should be a configurable option. Maybe just an environment variable would be enough?

RPigott:
@ddevault Yes I also think it should be configurable. A config option is more convenient, I'd say. I was just waiting for thoughts on the wlroots patch first.

emersion:
I'd prefer not to add another env variable. I'd be fine with a debug option.

RPigott:
I think it makes the most sense as an output config option. I can flesh out this patch later today.

RPigott:
Added a general config flag 'integer_scale_filter' that applies to all outputs and also rebased to avoid conflicts.

emersion:
How about making this an output command `scale_filter`, which defaults to nearest if integer scale is used but lets the user override this behaviour?

RPigott:
Like a linear/nearest/smart kinda thing? Where smart is integer scale => nearest?

I can't imagine any user would want to force nearest for all scaling factors, but seems sensible anyway. I kinda like having a 'smart' option similar to many other user config options.

It's a general config here just because it was suggested and easy, but I'll try that, it seems nice.

luispabon:
My 2p, need to be able to have different scaling choice per output because a lot of 13" laptop users are going to be using fractional on their hidpi laptop panel and integer on external displays.

RPigott:
@luispabon The option already only affects displays with integer scale, so that works. A per output config is needed to allow the user to disable nearest neighbor scaling on some outputs but not others, even when they both have integer scale.

Personally, I dont find NN to be an improvement, so I'd rather leave it on linear, but I don't have a hi-dpi display.

I'd be interested to hear from users that want this feature if there is a way to make the smart behavior smarter than just integer scale => NN. Like, are there other display qualities that make NN more or less preferable? Or am I weird for preferring linear.

tokyovigilante:
I have a Lenovo X1Y4 with a 4K 14" screen as well as a Dell 24" 4K display, I run sway @2x HiDPI scaling on both with changes to font sizes in Wayland apps. 

The only XWayland app I use is my browser (Vivaldi) which looks significantly better with GL_NEAREST scaling compared with GL_LINEAR, which is blurry and tends to wash out black text to grey.

+1 for this to be optional (and opt-in) for integer scaling, either per output or globally. I think an "auto" mode might not be ideal as people who are running mixed-DPI or with variable scaling factors across outputs will probably notice the difference and be surprised, but those who want NN scaling can just opt in on whatever display.

I'd say something like 

```
output eDP-1 {
    resolution 3840x2160
    scale 2.0
    integer_scale_filter <linear | nearest>
}
```

with linear as default if not specified would work well. Calling it `integer_scale_filter` might make it clear it applies to 2x, 3x etc only.

RPigott:
Okay. Old "integer_scale_filter" flag is now "scale_filter" output config option.

Values are [linear|nearest|smart] like above. linear is the default so the feature is entirely opt-in, like @tokyovigilante  says, because I figure the people who want this will know what to look for more than the people who don't. I think that should work for everyone. Added a man page blurb.

Obviously still requires the wlroots patch, but PTAL.

luispabon:
In my case most of my xwayland apps are all IDEs (jetbrains and vs code), and working with blurry text 8 hours a day is a massive no-no due to eye fatigue. NN on a 15.6" 4k display (scale 2) looks sharp enough that sometimes I forget it's not native resolution.

I quite like the way Windows 10 deals with this issue when you have a mix of scale factors on apps that aren't hidpi aware. It will use NN on integer scale displays and linear on non integer scales. It doesn't give any option to override though, which is windows-typical.

I honestly think `smart` should be the sane default here, instead of `linear`.

ddevault:
+1 for "smart" as the default

sway/desktop/render.c
```diff
@@ -70,6 +72,33 @@ static void scissor_output(struct wlr_output *wlr_output,
...
+
+	switch (scale_filter) {
+	case SCALE_FILTER_LINEAR:
+		glTexParameteri(attribs.target, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+		break;
+	case SCALE_FILTER_NEAREST:
+		glTexParameteri(attribs.target, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+		break;
+	case SCALE_FILTER_SMART:
+		if (ceilf(wlr_output->scale) == wlr_output->scale) {
```
emersion: We need to call `glTexParameteri` with `GL_LINEAR` in the other case to correctly react to live config changes.

I'd suggest setting `scale_filter` to either linear or nearest before the switch if it's set to smart.

sway/desktop/render.c
```diff
@@ -70,6 +72,33 @@ static void scissor_output(struct wlr_output *wlr_output,
...
+			glTexParameteri(attribs.target, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+		}
+		break;
+	default:
+		return;
+	}
+	return;
```
emersion: Unnecessary return

sway/desktop/render.c
```diff
@@ -70,6 +72,33 @@ static void scissor_output(struct wlr_output *wlr_output,
...
+	case SCALE_FILTER_NEAREST:
+		glTexParameteri(attribs.target, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+		break;
+	case SCALE_FILTER_SMART:
+		if (ceilf(wlr_output->scale) == wlr_output->scale) {
+			glTexParameteri(attribs.target, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+		}
+		break;
+	default:
+		return;
```
emersion: We should probably add an assertion here

sway/desktop/render.c
RPigott: I'm not sure I understand. You mean when the output scale changes? In my observation that works already, though I'm not actually certain why (or if it's reliable).

Either way, why not set it in an `else` after the integer check from the "smart" case if we want?

emersion: Yeah, when the scale changes from being an integer to a fraction, we need to set `GL_LINEAR`.

>Either way, why not set it in an else after the integer check from the "smart" case if we want?

Because it duplicates code. Suggested alternative:

```c
if (scale_filter == SCALE_FILTER_SMART) {
	scale_filter = ceilf(wlr_output->scale) == wlr_output->scale ? SCALE_FILTER_NEAREST :
		SCALE_FILTER_LINEAR;
}

switch (scale_filter) {
case SCALE_FILTER_LINEAR:
	glTexParameteri(attribs.target, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
	break;
case SCALE_FILTER_NEAREST:
	glTexParameteri(attribs.target, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
	break;
default:
	assert(false); // unreachable
}
```

sway/desktop/render.c
RPigott: Hmm, this actually is reachable, although I didn't expect it.

When sway starts it gets to render with scale_filter still at SCALE_FILTER_NONE. I intended NONE as a sentinal value for unset displays so output config merging is correct.

sway/desktop/render.c
```diff
@@ -110,6 +139,11 @@ static void render_surface_iterator(struct sway_output *output, struct sway_view
 		return;
 	}
 
+	struct output_config *oc = find_output_config(output);
```
progandy: How expensive is this call and Is it acceptable to call it in the iterator?  Should the ```scale_filter``` property be a member of ```sway_output```?

sway/desktop/render.c
```diff
@@ -119,6 +153,7 @@ static void render_surface_iterator(struct sway_output *output, struct sway_view
 	wlr_matrix_project_box(matrix, &box, transform, rotation,
 		wlr_output->transform_matrix);
 
+	set_scale_filter(wlr_output, texture, oc->scale_filter);
```
emersion: We should probably call `set_scale_filter` from `render_texture`

RPigott: The logic was originally inside `render_texture`, but I moved it out when I switched to a per-output config. My reasoning was:

1. `render_texture` doesn't already maintain a reference to the relevant sway_output
2. `set_scale_filter` is only required, I think, when rendering a texture that corresponds to an application buffer, so I call it before those calls to `render_texture` and do not call it in others such as those in `render_titlebar`.

sway/desktop/render.c
RPigott: I did think about that, but I don't know the answer.

progandy: If you do that, you could move the calculation of the required filter from ```set_scale_filter``` to ```apply_output_config``` as well. That removes almost all of the new code from the rendering path.

RPigott: That's a great idea, I'll rework it to that.

sway/config/output.c
```diff
@@ -299,6 +318,21 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		wlr_output_set_scale(wlr_output, oc->scale);
 	}
 
+	if (oc) {
+		enum scale_filter_mode scale_filter;
+		if (oc->scale_filter == SCALE_FILTER_SMART ||
+				oc->scale_filter == SCALE_FILTER_UNKNOWN) {
```
RPigott: This feels like a hack, but without it I hit the assert in `set_scale_filter` when sway starts up because output->scale_filter is SCALE_FILTER_UNKNOWN.

I include the unkown state to make output config merging work, and I treat unknown the same as smart because smart is the default. Let me know if you've got another idea.

RPigott:
`scale_filter` is now a sway_output property so we don't have to fetch the config in the rendering code.

Added the relevant ipc stuff for the output property, SCALE_FILTER_NONE is now SCALE_FILTER_UNKNOWN to match the subpixel enum, "smart" logic is now in apply_output_config.

Commits should be squashed, probably. Just helps me to keep them separate for a bit.

RPigott:
Okay, I think I've addressed all the review comments.

Fixed up the default option, squashed, rebased, and once again ready for review. PTAL.

sway/config/output.c
```diff
@@ -297,6 +316,21 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
...
+					SCALE_FILTER_NEAREST : SCALE_FILTER_LINEAR;
+				break;
+			case SCALE_FILTER_LINEAR:
+			case SCALE_FILTER_NEAREST:
+				output->scale_filter = oc->scale_filter;
+				break;
+		}
+		sway_log(SWAY_DEBUG, "Set %s scale_filter to %s", oc->name,
+			sway_output_scale_filter_to_string(output->scale_filter));
+		output_damage_whole(output);
```
RPigott: quick fix: this is only necessary if scale_filter actually changed.

RPigott:
Hmmm. Here's a thought... Why not move some of this to wlroots?

So, no more backend specific set_scale_filter in sway's render.c. Instead let wlroots gles2 backend set it and have a wlr_output_set_scale_filter for sway to use?

Other backends could do whatever they want with that (not sure if they care), and if wlroots had it's own enum corresponding to linear/nearest sway wouldn't even have to link to glesv2.

The config bits would stay here, and compositors that never call wlr_output_set_scale_filter will still get linear like before. Thoughts?

emersion:
Yeah, I'd prefer not to add this to wlroots. Maybe some compositors want to have more control over many more parameters (not just MAG_FILTER), maybe they want to have per-surface rules. Non-GLES2 renderers might support less (e.g. software renderer?) or more (e.g. Vulkan?) settings.

wlroots already links to GLES2, so there's not a lot of benefit from not linking to it in Sway: the Sway binary will end up linking against GLES2 anyway.

hypergig:
does this look like it will make it to 1.3?

ddevault:
Yes.

PR #4731, opened 2019-11-16
Use new presentation-time API

emersion:
See https://github.com/swaywm/wlroots/pull/1919

This has the advantage to (1) reduce boilerplate and (2) make us
correctly handle wlr_output_event_present.commit_seq.


ddevault:
Needs rebasing

ddevault:
Thanks!

PR #4732, opened 2019-11-17
grimshot: fix branching on command exit status

mmlb:
The previous behavior was incorrect because `if` was checking the return
status of the `[` command which was never going to be an error. `[`
seems to only return an error if no args are provided. This was
basically a useless use of `[` anyway since it was just meant as a
straight interpretation of command exit, something that `if` can do
itself.

Compare:

```sh
[ ]; echo ?=$?
[ /bin/false ]; echo ?=$?
if [ /bin/false ]; then echo this is the unintended bug; fi
if /bin/false; then echo this will not be printed; fi
```

### DIFF
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -51,8 +51,7 @@ die() {
 
 check() {
   COMMAND=$1
-  command -v "$COMMAND" > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


mmlb:
:+1: no problemo!

PR #2580, opened 2018-09-05
Restore focus correctly when closing a fullscreen view

RyanDwyer:
We weren't calling `seat_send_focus`. I think this was previously called by `seat_set_focus_warp`.

Fixes #2534.

PR #4740, opened 2019-11-20
Port input method and text input from rootston

Leo1003:
This ports basic input method support from rootston to sway.
Origin from: 
- swaywm/wlroots#1203
- swaywm/wlroots#1303
- swaywm/wlroots#1308
- swaywm/wlroots#1759

### DIFF
--- a/sway/input/text_input.c
+++ b/sway/input/text_input.c
@@ -59,15 +59,13 @@ static void text_input_set_pending_focused_surface(
 		struct sway_text_input *text_input, struct wlr_surface *surface) {
 	wl_list_remove(&text_input->pending_focused_surface_destroy.link);
 	text_input->pending_focused_surface = surface;
-	wl_signal_add(&surface->events.destroy,
-		&text_input->pending_focused_surface_destroy);
-}
 
-static void text_input_clear_pending_focused_surface(
-		struct sway_text_input *text_input) {
-	wl_list_remove(&text_input->pending_focused_surface_destroy.link);
-	wl_list_init(&text_input->pending_focused_surface_destroy.link);
-	text_input->pending_focused_surface = NULL;
+	if (surface) {
+		wl_signal_add(&surface->events.destroy,
+			&text_input->pending_focused_surface_destroy);
+	} else {
+		wl_list_init(&text_input->pending_focused_surface_destroy.link);
+	}
 }
 
 static void handle_im_destroy(struct wl_listener *listener, void *data) {
@@ -157,13 +155,12 @@ static void handle_text_input_destroy(struct wl_listener *listener,
 	if (text_input->input->current_enabled) {
 		relay_disable_text_input(text_input->relay, text_input);
 	}
-	text_input_clear_pending_focused_surface(text_input);
+	text_input_set_pending_focused_surface(text_input, NULL);
 	wl_list_remove(&text_input->text_input_commit.link);
 	wl_list_remove(&text_input->text_input_destroy.link);
 	wl_list_remove(&text_input->text_input_disable.link);
 	wl_list_remove(&text_input->text_input_enable.link);
 	wl_list_remove(&text_input->link);
-	text_input->input = NULL;
 	free(text_input);
 }
 
@@ -187,6 +184,8 @@ struct sway_text_input *sway_text_input_create(
 	}
 	input->input = text_input;
 	input->relay = relay;
+
+	wl_list_insert(&relay->text_inputs, &input->link);
 
 	input->text_input_enable.notify = handle_text_input_enable;
 	wl_signal_add(&text_input->events.enable, &input->text_input_enable);
@@ -215,12 +214,7 @@ static void relay_handle_text_input(struct wl_listener *listener,
 		return;
 	}
 
-	struct sway_text_input *text_input = sway_text_input_create(relay,
-		wlr_text_input);
-	if (!text_input) {
-		return;
-	}
-	wl_list_insert(&relay->text_inputs, &text_input->link);
+	sway_text_input_create(relay, wlr_text_input);
 }
 
 static void relay_handle_input_method(struct wl_listener *listener,
@@ -250,7 +244,7 @@ static void relay_handle_input_method(struct wl_listener *listener,
 	if (text_input) {
 		wlr_text_input_v3_send_enter(text_input->input,
 			text_input->pending_focused_surface);
-		text_input_clear_pending_focused_surface(text_input);
+		text_input_set_pending_focused_surface(text_input, NULL);
 	}
 }
 
@@ -276,7 +270,7 @@ void sway_input_method_relay_set_focus(struct sway_input_method_relay *relay,
 		if (text_input->pending_focused_surface) {
 			assert(text_input->input->focused_surface == NULL);
 			if (surface != text_input->pending_focused_surface) {
-				text_input_clear_pending_focused_surface(text_input);
+				text_input_set_pending_focused_surface(text_input, NULL);
 			}
 		} else if (text_input->input->focused_surface) {
 			assert(text_input->pending_focused_surface == NULL);
--- a/sway/input/text_input.c
+++ b/sway/input/text_input.c
@@ -174,6 +174,8 @@ static void handle_pending_focused_surface_destroy(struct wl_listener *listener,
 	struct wlr_surface *surface = data;
 	assert(text_input->pending_focused_surface == surface);
 	text_input->pending_focused_surface = NULL;
+	wl_list_remove(&text_input->pending_focused_surface_destroy.link);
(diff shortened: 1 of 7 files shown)


ddevault:
Looks pretty good at a first glance. Are there any new clients we can look at for testing this, or are we still stuck with the basic demo clients?

Leo1003:
We currently use the [input-method example in wlroots](https://github.com/swaywm/wlroots/blob/master/examples/input-method.c) and [wlchewing](https://github.com/xdavidwu/wlchewing) to test it. But the former is just an example and can only do some basic tests. The other one is an very early stage implement using libchewing.

SpencerMichaels:
What work remains before this PR can be merged? I am interested in getting this functionality into Sway and would be happy to help test or patch as needed. @Leo1003  My use case is also Chinese input methods (I use Pinyin, but am familiar with Zhuyin, although rusty); maybe there is something I could do to get wlchewing or another Chinese IME into a more usable state for this?

For testing, does ibus not support Wayland? It looks like it has support for V1 of this protocol, at least — see [here](https://github.com/ibus/ibus/tree/master/client/wayland).

ohk2kt3t4:
@SpencerMichaels For what it's worth, I cannot get `ibus` to run without `xwayland`, despite the [package ](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/ibus)I installed it with building it with the `--enable-wayland` flag. The input itself does work in wayland containers though (because it's using the V1 protocol mentioned above?).

xdavidwu:
I faced some hangs but @Leo1003 could not reproduce, and I don't have time to look into it (yet). For testing, [input-method example in wlroots](https://github.com/swaywm/wlroots/blob/master/examples/input-method.c) can be used (with `GTK_IM_MODULE=wayland` on gtk3 apps, for example). If you want to test with [wlchewing](https://github.com/xdavidwu/wlchewing) (a Chinese Zhuyin input method which is lacks some features but I would consider it feasible now for testing), you will also need [keyboard grab support](https://github.com/swaywm/wlroots/pull/1864) on wlroots. (Cursor may be drawn in wrong position on gtk though, because of a gtk [bug](https://gitlab.gnome.org/GNOME/gtk/merge_requests/1251))

We don't support input-method-unstable-v1. If ibus works, it is likely running under toolkit (gtk or qt) im modules, and if panels appears, it is by xwayland.

David96:
Is there any reason this is not getting merged? I tried it with squeekboard and it makes the on screen keyboard reliably show up on focus of a text field, which is really useful for the tablet mode of a convertible.

Leo1003:
@David96 I think this PR is ready to merge, but we are making it better.
We currently work on 7195ddc6d87299fa51bbcd9e40d19d5ff5baaa04 to eliminate some duplicate call, the already focused issue seems related to #2580. This issue may need some time to investigate.
If we cannot fix it, I would just remove the debug message and waiting for merge.

David96:
Alright, no rush and thanks for the update. As long as it hasn't been forgotten, I'm happy :)

Leo1003:
The final issue is resolved in #5130, I think this is ready for review.
P.S. The hangs in the early conversions are some "use after free" issues, and have been fixed :wrench:

dcz-purism:
Is there a reason for making text-input listeners per-text input? While I don't think this was leveraged so far, the relay was the choke point for detecting invalid states, like multiple text-inputs active at the same time. It could prevent broken calls to wlr_input_method_v2_send_activate if needed. Putting that in text-input handler might not be a good idea.

There is some overlap in lifetimes of text-inputs (esp. when switching clients) that should be carefully analyzed.

As a note not pertaining the actual patch, it seems that the text-input v3 protocol, and by extension, the input-method v2 were botched, and i would advise to have a way to drop them in the future in favor of the fixed version. Here's some background: 

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/19

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/16

and, to a lesser extent, conditional on not having virtual-keyboard widely accepted: https://gitlab.freedesktop.org/wayland/wayland-protocols/issues/12

I would appreciate if those MRs could get more reviews, to justify releasing a fixed revision and moving on to the fix without having to implement the bad version first.

sway/input/text_input.c
```diff
@@ -0,0 +1,302 @@
...
+	if (relay->seat->wlr_seat != wlr_text_input->seat) {
+		return;
+	}
+
+	struct sway_text_input *text_input = sway_text_input_create(relay,
+		wlr_text_input);
+	if (!text_input) {
+		return;
+	}
+	wl_list_insert(&relay->text_inputs, &text_input->link);
```
emersion, line 223: This should be done in `sway_text_input_create`

sway/input/text_input.c
```diff
@@ -0,0 +1,302 @@
...
+	if (text_input->input->current_enabled) {
+		relay_disable_text_input(text_input->relay, text_input);
+	}
+	text_input_clear_pending_focused_surface(text_input);
+	wl_list_remove(&text_input->text_input_commit.link);
+	wl_list_remove(&text_input->text_input_destroy.link);
+	wl_list_remove(&text_input->text_input_disable.link);
+	wl_list_remove(&text_input->text_input_enable.link);
+	wl_list_remove(&text_input->link);
+	text_input->input = NULL;
```
emersion, line 166: No need to do this, since we're free'ing `text_input` right after.

sway/input/text_input.c
```diff
@@ -0,0 +1,302 @@
...
+	struct sway_input_method_relay *relay = wl_container_of(listener, relay,
+		input_method_destroy);
+	struct wlr_input_method_v2 *context = data;
+	assert(context == relay->input_method);
+	relay->input_method = NULL;
+	struct sway_text_input *text_input = relay_get_focused_text_input(relay);
+	if (text_input) {
+		// keyboard focus is still there, so keep the surface at hand in case
+		// the input method returns
+		text_input_set_pending_focused_surface(text_input,
```
emersion, line 83: This will blow up if `text_input->input->focused_surface` is NULL.

I think it would be better to merge `text_input_clear_pending_focused_surface` into `text_input_set_pending_focused_surface`, and make `text_input_set_pending_focused_surface` accept a NULL `surface`.

emersion:
Can you also rebase against master?

Leo1003:
Ok, I will apply those changes and rebase ASAP.

Leo1003:
Rebased against master and changes are applied!

fosskers:
Sorry to dig up an old thread, but how is this actually used? The release for Sway 1.5 states:

> Input method editors (IME) are now supported via the input-method and text-input protocols

but doesn't provide an example. Nor does it seem that anything was added to the manpages or Wiki.

Does anyone have a working example?

fosskers:
Note to others. Adding the following to your swag `config` "just works":
```
exec_always fcitx5 -d --replace
```
Also  make sure to set these environment variables somewhere:
```fish
# Fish shell
set -x GTK_IM_MODULE 'fcitx'
set -x QT_IM_MODULE 'fcitx'
set -x XMODIFIERS '@im=fcitx'
```
I can now type 日本語 and other things in my Wayland windows. Unfortunately it doesn't work (not even the language switching) if I'm focused on an `alacritty` window.

12101111:
@fosskers You are still using the dbus frontend of fcitx5, run this command

```
$ dbus-send --session --type=method_call --print-reply --dest=org.fcitx.Fcitx5 /controller org.fcitx.Fcitx.Controller1.DebugInfo
method return time=1614562501.869255 sender=:1.7 -> destination=:1.39 serial=2629 reply_serial=2
   string "Group [wayland:] has 4 InputContext(s)
  IC [d2ab939f5c4d4f18b45bcf4df5df4fe8] program:firefox-wayland frontend:dbus cap:e001000012 focus:0
  IC [fd5d833c24d6448e875510a3871de9a4] program:firefox-wayland frontend:dbus cap:e001000012 focus:0
  IC [20e600d1d1354f8b87d7f5f38929a8a4] program:firefox-wayland frontend:dbus cap:e001000012 focus:0
  IC [7dcea1603b2f46bf9bf8f3139d8af51e] program:firefox-wayland frontend:dbus cap:e001000052 focus:0
Group [x11::0] has 2 InputContext(s)
  IC [8a8666158c624e6fba303472090d3c27] program:goldendict frontend:dbus cap:800072 focus:0
  IC [0b5c913b9cb74a9288218f8386d916b0] program:goldendict frontend:dbus cap:800072 focus:0
Input Context without group
"
```

There are 5 frontend provided by fcitx, `GTK_IM_MODULE=fcitx` select `dbus` frontend. With https://github.com/swaywm/sway/pull/5890, you can use waylandim frontend and type on alacritty.

Issue #4637, opened 2019-10-15
Cannot compile custom xkb keymap (key type must be integer)


InspectorMustache:
I'm trying to make sway use my xkb file (many thanks for implementing that feature by the way).  I can't seem to get it to work, however.  To isolate the problem, I've used this minimal config:

```
input * xkb_file /home/nicolai/.config/xkb/custom.xkb
bindsym Mod4+Return exec termite
```

I then start sway by running `sway -c config`, open termite, run `swaymsg exit` and see these errors related to xkb:
```
2019-10-15 21:01:32 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:32 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:32 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:32 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:32 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:32 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:32 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:32 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:33 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:33 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
2019-10-15 21:01:33 - [sway/input/keyboard.c:606] cannot close XKB file /home/nicolai/.config/xkb/custom.xkb: Success: Success
[...]

PR #4743, opened 2019-11-21
input/keyboard: cleanup xkb_file error handing

RedSoxFan:
Fixes https://github.com/swaywm/sway/issues/4637#issuecomment-542271957

This fixes an inverted fclose return value check and simplifies the
error handling and logging for xkb_file in sway_keyboard_compile_keymap


ddevault:
Thanks!

PR #4757, opened 2019-11-27
config: improvements to the reload validation

### LINKED ISSUE #4744: Crash on `xkb_state_update_key` when reloading config
Sway crashes when I attempt to reload the config since https://github.com/swaywm/sway/commit/5d882cb5fc2d9d9fd68439021e48a90aa2e50e79. Building from 2f858a1a works.

## Steps to reproduce

1. ??? - I haven't narrowed this down much so don't know if it just applies to my setup
2. Build sway from master
3. Use keyboard shortcut to reload sway
4. Crashes on `xkb_state_update_key` via `wlr_keyboard_notify_key`
5. Build sway from commit before `Add support for wlr_keyboard_group`
6. Reload works

## Versions & Logs

- Sway Version: `sway version 1.2-5d882cb5 (Nov, branch 'reload-red')`
- Debug Log: https://paste.ubuntu.com/p/576MJwtQ3B/
- Configuration File: https://github.com/Jamedjo/dotfiles/blob/fb35804f7a002fd85edd5f32b80053071ff3e77d/sway/config
* Stack Trace, if sway crashes: https://paste.ubuntu.com/p/jWQ8YgdcyC/

RedSoxFan:
Fixes #4744 

For the validation pass of reloading, there is no need to touch swaybg,
swaynag, inputs, outputs, or seats. This drastically improves the speed
of a reload by skipping over the expensive I/O configuration and
handling of wayland clients. As long as the syntax is valid, the
CMD_FAILURE's can be relayed during the actual reload.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -447,15 +447,17 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 				old_config->xwayland ? "enabled" : "disabled");
 		config->xwayland = old_config->xwayland;
 
-		if (old_config->swaybg_client != NULL) {
-			wl_client_destroy(old_config->swaybg_client);
-		}
+		if (!config->validating) {
+			if (old_config->swaybg_client != NULL) {
+				wl_client_destroy(old_config->swaybg_client);
+			}
 
-		if (old_config->swaynag_config_errors.client != NULL) {
-			wl_client_destroy(old_config->swaynag_config_errors.client);
-		}
+			if (old_config->swaynag_config_errors.client != NULL) {
+				wl_client_destroy(old_config->swaynag_config_errors.client);
+			}
 
-		input_manager_reset_all_inputs();
+			input_manager_reset_all_inputs();
+		}
 	}
 
 	config->current_config_path = path;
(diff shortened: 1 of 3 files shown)


sway/config.c
```diff
@@ -520,8 +522,13 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 		return success;
 	}
 
-	if (is_active) {
+	if (is_active && !validating) {
 		input_manager_verify_fallback_seat();
+
+		for (int i = 0; i < config->input_configs->length; i++) {
+			input_manager_apply_input_config(config->input_configs->items[i]);
```
emersion: This call seems to have been added. Is there a reason why we didn't need it and need it now?

RedSoxFan: It's not being handled in `cmd_input` on reload anymore

https://github.com/swaywm/sway/blob/33be52b26a2b83add6be469de8840ecb42f3002c/sway/commands/input.c#L92-L94

emersion: Nevermind, the old call now has a `if (config->reloading)` guard

PR #4764, opened 2019-11-28
input/keyboard: reset seat keyboard on destroy

### LINKED ISSUE #4758: Assertion raised when new seat attaches all inputs in config
**version**: `sway version 1.2-1a579780 (Nov 27 2019, branch 'master')`
Looks like creating `seat1` in the config screws up with the default `seat0` which loses its `sway_keyboard *`, which triggers an assertion

### Reproduction steps
1. config:
   ```
   # config.config
   exec_always xterm
   seat seat1 attach *
   ```
2. Command
    `sway -d -c config.config`
3. That's it, assertion raised

### Stack trace
```
Thread #1 [sway] 76145 [core: 5] (Suspended : Signal : SIGABRT:Aborted)	
	raise() at 0x7ffff770bf25	
	abort() at 0x7ffff76f5897	
	__assert_fail_base.cold() at 0x7ffff76f5767	
	__assert_fail() at 0x7ffff7704526	
	seat_keyboard_notify_enter() at seat.c:126 0x555555586e32	
	seat_send_focus() at seat.c:160 0x555555586f98	
	seat_set_focus() at seat.c:1,051 0x555555589401	
	input_manager_set_focus() at input-manager.c:357 0x55555557f650	
	view_map() at view.c:649 0x5555555b6724	
	handle_map() at xwayland.c:431 0x5555555bccd6	
	<...more frames...>	
```
### Cause
The assertion triggered [is this one](https://github.com/swaywm/sway/blob/1a5797803a48fb71515e6bd6335d684a6cae6b6d/sway/input/seat.c#L126), coming from `seat_keyboard_notify_enter()`:
```c
static void seat_keyboard_notify_enter(struct sway_seat *seat,
	struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(seat->wlr_seat);
	if (!keyboard) {
		wlr_seat_keyboard_notify_enter(seat->wlr_seat, surface, NULL, 0, NULL);
		return;
	}

	struct sway_keyboard *sway_keyboard =
		sway_keyboard_for_wlr_keyboard(seat, keyboard);
//////////
// That one
	assert(sway_keyboard && "Cannot find sway_keyboard for seat keyboard");

	struct sway_shortcut_state *state = &sway_keyboard->state_pressed_sent;
	wlr_seat_keyboard_notify_enter(seat->wlr_seat, surface,
			state->pressed_keycodes, state->npressed, &keyboard->modifiers);
}
```
The function `seat_keyboard_notify_enter()` is called twice, once for each seat: `seat1` and `seat0`. Turns out `seat0` has no keyboard -> assertion raised

**Suggested solution:** not sure!

_seat1_ is fine but _seat0_ is not. The function raising the assertion looks that `seat->wlr_seat` has a `wlr_keyboard *`, which it as, and then a `sway_keyboard *`, which _seat0_ has not anymore... Bad seat initialisation?

I tried checking how `seat_add_device()` gets called but I kind of stopped investigating there.

RedSoxFan:
Fixes #4758 

If a sway keyboard is being destroyed, then the keyboard is being
removed from a seat. If the associated wlr_keyboard is the currently
set keyboard for the wlr_seat, then we need to reset the wlr_seat's
keyboard to NULL so it doesn't reference an invalid device for the seat.
The next configured keyboard from the seat or the next keyboard from
that seat that has an event will then become the seat keyboard.
Similarly, this needs to be done for a wlr_keyboard_group's keyboard
when the wlr_keyboard_group is being destroyed.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -675,9 +675,7 @@ static void sway_keyboard_group_remove(struct sway_keyboard *keyboard) {
 		struct sway_keyboard_group *sway_group = wlr_group->data;
 		wlr_group->data = NULL;
 		wl_list_remove(&sway_group->link);
-		wl_list_remove(&sway_group->keyboard_key.link);
-		wl_list_remove(&sway_group->keyboard_modifiers.link);
-		free(sway_group->seat_device->keyboard);
+		sway_keyboard_destroy(sway_group->seat_device->keyboard);
 		free(sway_group->seat_device->input_device);
 		free(sway_group->seat_device);
 		free(sway_group);
@@ -920,6 +918,11 @@ void sway_keyboard_destroy(struct sway_keyboard *keyboard) {
 	}
 	if (keyboard->seat_device->input_device->wlr_device->keyboard->group) {
 		sway_keyboard_group_remove(keyboard);
+	}
+	struct wlr_seat *wlr_seat = keyboard->seat_device->sway_seat->wlr_seat;
+	struct sway_input_device *device = keyboard->seat_device->input_device;
+	if (wlr_seat_get_keyboard(wlr_seat) == device->wlr_device->keyboard) {
+		wlr_seat_set_keyboard(wlr_seat, NULL);
 	}
 	if (keyboard->keymap) {
 		xkb_keymap_unref(keyboard->keymap);


sway/input/keyboard.c
```diff
@@ -921,6 +919,11 @@ void sway_keyboard_destroy(struct sway_keyboard *keyboard) {
 	if (keyboard->seat_device->input_device->wlr_device->keyboard->group) {
 		sway_keyboard_group_remove(keyboard);
 	}
+	struct wlr_seat *wlr_seat = keyboard->seat_device->sway_seat->wlr_seat;
+	struct sway_input_device *device = keyboard->seat_device->input_device;
+	if (wlr_seat_get_keyboard(wlr_seat) == device->wlr_device->keyboard) {
+		wlr_seat_set_keyboard(wlr_seat, NULL);
```
emersion: Hmm, this should happen automatically inside wlroots because the seat should listen to the keyboard's destroy event.

RedSoxFan: The wlr_keyboard can still exist. This needs to occur when changing the keyboard changes seats in sway (which destroys and creates a new sway_keyboard) as well. The wlr_keyboard doesn't get destroyed in that case

paul-ri:
Yes, thanks a lot

PR #4766, opened 2019-11-28
output: Ensure that frame_done is delayed on max_render_time

kennylevinsen:
max_render_time can be set on output, view, or both. However, if only
applied to the output, send_frame_done_iterator would erroneously send
frame_done immediately, ignoring the output max_render_time. As
damage_handle_frame processed max_render_time correctly, idle frames
would be blocked in anticipation of the delay that was meant to happen.

Without the delay, frame events would be dispatched during the idle
frame block, and some clients would never receive the frame done events
they had requested, at least not until something else actively drove
another render.

Respecting both view and output max_render_time in
send_frame_done_iterator ensures that the frame events are always
correctly delayed.

Fixes #4756

PR #4772, opened 2019-11-30
output: Really fix max render time

kennylevinsen:
The previous fix (#4766) was discovered to just treat a symptom rather than the cause, and did not account for proper use of per-view max_render_time.

This PR restores the original behavior of view max_render_time, and instead ensures that surface commits always result in a frame being scheduled, even if it occurred during our frame schedule block.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -426,7 +426,7 @@ static void send_frame_done_iterator(struct sway_output *output, struct sway_vie
 	int delay = data->msec_until_refresh - output->max_render_time
 			- view_max_render_time;
 
-	if ((output->max_render_time == 0 && view_max_render_time == 0) || delay < 1) {
+	if (output->max_render_time == 0 || view_max_render_time == 0 || delay < 1) {
 		wlr_surface_send_frame_done(surface, &data->when);
 	} else {
 		struct sway_surface *sway_surface = surface->data;
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -515,6 +515,9 @@ int output_repaint_timer_handler(void *data) {
 
 	output->wlr_output->frame_pending = false;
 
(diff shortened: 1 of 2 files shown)


sway/desktop/output.c
```diff
@@ -513,7 +513,10 @@ int output_repaint_timer_handler(void *data) {
 		return 0;
 	}
 
-	output->wlr_output->block_idle_frame = false;
+	output->wlr_output->frame_pending = false;
+
+	bool surface_needs_frame = output->surface_needs_frame;
+	surface_needs_frame = false;
```
YaLTeR: Seems like a left-over debug line which actually disables the behavior?

sway/desktop/output.c
```diff
@@ -552,11 +555,14 @@ int output_repaint_timer_handler(void *data) {
 		return 0;
 	}
 
+
```
YaLTeR: Stray blank line?

sway/desktop/output.c
kennylevinsen: Ah, my latest push had not gone through. Weird.

The correct form is `output->surface_needs_frame = false`.

kennylevinsen:
Feel free to give it a go.

`frame_pending` serves the same purpose as `block_idle_frame`, stopping `wlr_output_schedule_frame` from scheduling any idle frames. It is cleared immediately before our frame event, but we can just reset it.

tokyovigilante:
Seems to be working pretty well in limited testing with output max_render_time of 3ms and view max_render_time of 2ms (Radeon Vega VII, Mesa 20-git). Any finer than that and glxgears which I've been using to test and my browser start dropping frames. Latency feels significantly improved.

YaLTeR:
Yeah, looks like this is working well on my tests.

PR #4775, opened 2019-12-01
fix bar_state_update/input event in sway-ipc docs

JayceFayne:
While implementing rust bindings for the sway-ipc I noticed this small bug in the sway-ipc docs.
Hope this will prevent future confusion if someone wants to implement other bindings.


ddevault:
Thanks!

RedSoxFan:
I'm pretty sure this was right before. `((1<<31) | 20)` is `0x80000014`. `20` decimal is `14` hex

JayceFayne:
Damn u are right, sorry!
Forgot the hex conversion...

Issue #4684, opened 2019-10-29
Draw layer-shell popups on top of regular windows


emersion:
xdg-shell popups are already drawn on top of other toplevels, regardless of the toplevels z-index. We should do the same for layer-shell popups.

Original issue: https://github.com/swaywm/wlr-protocols/issues/60

ammen99:
I do not like this solution, because I do not see such behavior specified anywhere in the wayland protocols. Can you elaborate why Sway even started drawing in this way for xdg-shell windows?

David96:
@ammen99 And what would you suggest about the problem I described in the issue linked by emersion? I think it is reasonable for popups to be positioned over regular windows. Popups very often overlap other windows and usually you want to be able to see them (just like for the popups of waybar that show additional information when hovering some item).

ammen99:
The solution would be that the compositor lets you configure order of surfaces by their namespace. This idea will solve your problem, yes, but doesn't fix the fundamental issue - layer-shell's layers are nowhere near enough.

emersion:
> I do not see such behavior specified anywhere in the wayland protocols

It's a compositor policy. xdg-popup doesn't specify that popups should be drawn over everything else.

>Can you elaborate why Sway even started drawing in this way for xdg-shell windows?

Let's say you have two overlapping windows: a text editor on top of a web browser. Let's say the user moves the mouse over a link in the web browser and the web browser shows a tooltip with xdg-popup. You want to draw the tooltip on top of both windows.

David96:
> The solution would be that the compositor lets you configure order of surfaces by their namespace. Drawing popups on top will solve your problem, yes, but doesn't fix the fundamental issue - layer-shell's layers are nowhere near enough.

But I really don't want to start thinking about exactly at which layer what application should be drawn - that'd be additional work that has to be done for every panel or application starter I want to try out. On the other hand, I can't imagine a usecase where popups shouldn't be drawn over other windows. This would really make life easier.

ammen99:
@David96 You don't want to but you'll have to think it as soon as you want to add a third application which belongs to the top layer :)

Anyway I see your point, maybe this isn't a bad idea after all.

David96:
I would be willing to spend some time trying to implement this but I have zero experience with Wayland and basically none with wlroots and the sway codebase, so some pointer as to where to start would be appreciated.
[...]

PR #4781, opened 2019-12-03
Render layer shell popups over the top layer

David96:
As talked about in #4684, this renders popups separately from the corresponding windows, to make sure they are always displayed above other windows. I'm not sure though, whether they should be rendered before or after the top layer.

Furthermore, I don't have a good understanding of sways rendering code. This seems to work for me, but I cannot really figure out the side effects those changes may have. Splitting the output_layer_for_each_surface in three functions also gives some code duplication but I wanted to make sure to not break anything using output_layer_for_each_surface and expecting it to also include popups.
So it would be nice, if someone with an actual understanding of this code could give it a look and it's also why it is marked [WIP].

Edit: something I just realised: this changes the rendering order, but it doesn't change the order used for giving focus, therefore e.g. a context menu doesn't get any input events outside of the parent window.

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -125,6 +125,14 @@ void output_layer_for_each_surface(struct sway_output *output,
 	struct wl_list *layer_surfaces, sway_surface_iterator_func_t iterator,
 	void *user_data);
 
+void output_layer_for_each_surface_toplevel(struct sway_output *output,
+	struct wl_list *layer_surfaces, sway_surface_iterator_func_t iterator,
+	void *user_data);
+
+void output_layer_for_each_surface_popup(struct sway_output *output,
+	struct wl_list *layer_surfaces, sway_surface_iterator_func_t iterator,
+	void *user_data);
+
 #if HAVE_XWAYLAND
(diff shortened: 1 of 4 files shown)


alebastr:
To finish this PR matching changes to `sway/input/cursor.c` are needed. I made a PoC change that seems to be working as expected: https://github.com/alebastr/sway/commit/a9bd645fd576a6da81f7580dc5008e07cce15f33
However the changes required to make this working made me think that rendering layer shell toplevel and popups on a different layers is way outside of initial assumptions. Maybe there's a cleaner way to do that, like changing layer-shell surface storage to be a single list sorted in a desired rendering/focus order and having only one place in the code that defines the order. Or 3 lists, overlay, above normal surfaces and below normal surfaces.

emersion:
Then you need to sort the list, and find a way to store mixed layer surfaces and popups in the list. I'm not sure it'll be simpler in the end.

David96:
Rebased to current master and added @alebastr 's commit, removing WIP since it now at least does everything it should.

PR #4782, opened 2019-12-03
config: apply input type configs on reload

### LINKED ISSUE #4780: Input 'type:' settings not applied on reload
I have in my config a touchpad config like so:

    input type:touchpad {
        tap enabled
        pointer_accel 0.8
        dwt disabled
    }

When `reload`ing, these settings are reverted to their libinput defaults, and then when executing any `swaymsg input type:touchpad ...` command they are restored as written in the config (with whatever config just applied by swaymsg, of course).

Observed on `sway version 1.2-ad189d27 (Dec  3 2019, branch 'master')`
Appears to be caused by 1a579780. Previous commit does _not_ share the erroneous behavior.

RedSoxFan:
Fixes #4780 

When making the reload validation improvements (#4757), I forgot that
input type configs are stored in a separate list. This makes it so input
type configs are correctly applied on reload.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -529,6 +529,11 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 			input_manager_apply_input_config(config->input_configs->items[i]);
 		}
 
+		for (int i = 0; i < config->input_type_configs->length; i++) {
+			input_manager_apply_input_config(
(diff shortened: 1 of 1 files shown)


RPigott:
Yup looks like that was it. Thanks for the fix!

PR #4786, opened 2019-12-05
Improve and add contrib scipts

xenrox:
I changed `inactive-windows-transparency.py` so that the transparency gets removed when the script is terminated and added a command line option to set the transparency value.

`autoname-workspaces.py` is a script that adds icons to the workspace name for each open application.


ddevault:
Thanks!

PR #4789, opened 2019-12-05
Fix segfault on toggling a scratchpad hidden view

### LINKED ISSUE #4788: Sway crashes on show scratchpad when focused window already is inside the scratchpad
I use a rofi to focus windows, including those in scratchpad
It does this in the background `swaymsg [con_id=<id of client> focus`
This works good for top level windows in the scratchpad, and shows the client.
For "nested" clients, such as a terminal in a tabbed view, it does not show the window
If you then tried to show the scratchpad sway crashes.

* Sway Version: 1.2
* Debug Log: https://gist.github.com/hakeee/3e64be7964ed26ee38c31052502b28a4
* Configuration File: `default` but with `termite` as terminal

* Steps to reproduce the crash:
1. Start sway
2. Open a terminal 
   This terminal gets con_id=5
3. Create a tabbed view
4. Open another terminal
5. Select parent (both tabs)
6. Move to scratchpad `swaymsg move scratchpad`
7. Open a new terminal
8. Execute `swaymsg [con_id=5] focus` 
   I believe the tabbed view should appear here with the terminal foused but it does not. 
   It would be nice if it did.
9. Show scratchpad `swaymsg scratchpad show`

```
#0  0x00005574b8a91f5e in  ()
#1  0x00005574b8a92fde in  ()
#2  0x00005574b8a9c1bf in  ()
#3  0x00005574b8ac32c7 in  ()
#4  0x00005574b8aa7477 in  ()
#5  0x00005574b8ab026a in  ()
#6  0x00007fc240f5cb0e in  () at /usr/lib/libwlroots.so.3
#7  0x00007fc240f4afdd in wlr_keyboard_notify_key () at /usr/lib/libwlroots.so.3
#8  0x00007fc240f213b7 in  () at /usr/lib/libwlroots.so.3
#9  0x00007fc240f2032c in  () at /usr/lib/libwlroots.so.3
#10 0x00007fc240fa37f2 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#11 0x00007fc240fa239c in wl_display_run () at /usr/lib/libwayland-server.so.0
#12 0x00005574b8a86592 in  ()
#13 0x00007fc2415a5153 in __libc_start_main () at /usr/lib/libc.so.6
#14 0x00005574b8a868be in  ()
```

RPigott:
Fixes #4788 

It's possible to focus scratchpad hidden views (does i3 allow this?). Previously, sway would crash trying to hide an already hidden view. Now focused scratchpad hidden views will be made visible on the active workspace like other scratchpad hidden views.

### DIFF
--- a/sway/commands/scratchpad.c
+++ b/sway/commands/scratchpad.c
@@ -28,7 +28,7 @@ static void scratchpad_toggle_auto(void) {
 
 	// Check if the currently focused window is a scratchpad window and should
 	// be hidden again.
-	if (focus && focus->scratchpad) {
+	if (focus && focus->scratchpad && focus->workspace) {
 		sway_log(SWAY_DEBUG, "Focus is a scratchpad window - hiding %s",
 				focus->title);
 		root_scratchpad_hide(focus);


RedSoxFan:
The proper solution would be to have the top-level floating container show when focusing a descendant of a scratchpad hidden container. To answer your question, i3 does not support nested floating containers so there isn't functionality to match. However, it is possible to show a scratchpad hidden top-level by focusing it so this would be the natural progression

RPigott:
Oh good, that's much more reasonable. Was gonna test if i3 really allowed this cause it seemed so weird. I didn't think to check the behavior when focusing an unnested view.
Makes sense focusing a nested hidden container should unhide the container. That's where the real bug is.

Issue #1904, opened 2018-05-03
Touch bindings


SirCmpwn:
```
touch {
    region along-the-top 20px x 1.0 @ .5 x 0

    region --window win-bottom-half 1.0 x .5 @ 0 x .5
    region --window win-right-half .5 x 1.0 @ .5 x 0

    # five finger tap to spawn a terminal
    gesture exec $term {
        down 5
        delay < 10ms
        up all # optional, all guestures implicitly end with "up all"
        # equivalent shorter version:
        # tap 5
    }

    # five finger long press to lock screen
    gesture exec swaylock {
        down 5
        50ms < delay < 1s
    }

    # slide along the top to right to spawn dmenu
    gesture exec dmenu_run {
        down @along-the-top 1
        horiz +100px < move
    }

    # two finger slide across bottom half of window to splith
    gesture splith {
        down @win-bottom-half
        horiz +/-0.5 < move
    }

    gesture splitv {
        down @win-top-half
        vert +/-0.5 < move
    }

    # rotate screen with 5 finger rotate
    gesture output LVDS-1 transform +90 {
        down 5
        45deg < rotation < 90deg
        up all
    }

    # drag from right/left to switch workspaces
    # --create to make a new one if you hit the end
    gesture workspace --create next_on_output {
        down "100px x 1.0 @ 1.0-100px x 0" 1
        horiz -0.25 < move
    }

    gesture workspace --create prev_on_output {
        down "100px x 1.0 @ 0 x 0" 1
        horiz +0.25 < move
    }
}
```

SirCmpwn:
Extended examples with some gestures for interactive move/resize/rotate of windows

ascent12:
Having the command after 'gesture' seems odd.

SirCmpwn:
Where else would it go?

grahnen:
I have a suggestion for an API, built on the API sent to me by @ddevault over on IRC. (Thanks a lot for the guidance!) It's a little crude maybe, so if it could receive a second opinion from someone that'd be incredibly helpful! For now it's over on https://github.com/grahnen/libtouch but if you want to adopt it, it can just be transferred to swaywm/libtouch.

Seeing this is my first attempt at any actual FLOSS contribution, I feel like I need the approval of at least one more before I proceed, hope that's not too much to hope for ^^

ddevault:
This looks close... I wonder if you could write a little demo program?

grahnen:
[...]

PR #4794, opened 2019-12-07
Pass gesture events to clients

### LINKED ISSUE #4724: Implement pointer-gestures wayland protocol [Former: Trackpad (pinch) gestures are not passed to clients]
Steps to reproduce:
 - start evince (or any other app that uses pinch gestures. Mostly gtk3 apps. eyg, inkscape 1.0 ...)
 - pinch your trakpad... (I am using a mbp 2015) nothing happens
 - On gnome shell the document gets zoomed in/out

The pointer-gestures protocol is not yet implementd for sway.
It is part of wlroots: https://github.com/swaywm/wlroots/pull/1731/files
Maybe there is only very little additional work to do?

lolzballs:
Fixes #4724.

This change can be tested quite easily if you use a backend which supports gestures (i.e. libinput). Open evince and your favourite pdf and you shall now see that pinch-to-zoom works properly. Prior to this, pinch-to-zoom would not do anything since the pinch and swipe events generated by the libinput backend are not handled anywhere (i.e. passed to the active client).

I'm not too sure how this will affect #1904 but I think there shouldn't be a huge change, as this change is solely passing pointer gestures to clients, rather than detecting a configurable gesture and executing a sway routine.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -735,6 +735,60 @@ static void handle_request_set_cursor(struct wl_listener *listener,
 			event->hotspot_y, focused_client);
 }
 
+static void handle_pointer_pinch_begin(struct wl_listener *listener, void *data) {
+	struct sway_cursor *cursor = wl_container_of(
+			listener, cursor, pinch_begin);
+	struct wlr_event_pointer_pinch_begin *event = data;
+	wlr_pointer_gestures_v1_send_pinch_begin(
+			cursor->pointer_gestures, cursor->seat->wlr_seat,
+			event->time_msec, event->fingers);
+}
+
+static void handle_pointer_pinch_update(struct wl_listener *listener, void *data) {
+	struct sway_cursor *cursor = wl_container_of(
+			listener, cursor, pinch_update);
+	struct wlr_event_pointer_pinch_update *event = data;
+	wlr_pointer_gestures_v1_send_pinch_update(
+			cursor->pointer_gestures, cursor->seat->wlr_seat,
+			event->time_msec, event->dx, event->dy,
+			event->scale, event->rotation);
+}
+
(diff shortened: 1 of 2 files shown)


ddevault:
Thanks! This doesn't effect touch gestures, which are a different animal entirely.

sway/input/cursor.c
```diff
@@ -825,6 +879,20 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat) {
 	cursor->hide_source = wl_event_loop_add_timer(server.wl_event_loop,
 			hide_notify, cursor);
 
+	cursor->pointer_gestures = wlr_pointer_gestures_v1_create(server.wl_display);
+	cursor->pinch_begin.notify = handle_pointer_pinch_begin;
+	wl_signal_add(&wlr_cursor->events.pinch_begin, &cursor->pinch_begin);
```
emersion: These signals are not removed when the cursor is destroyed.

lolzballs: Ah sorry about that. I made a change to correct this, although since the pull request was merged already I'm not sure what the best path to fixing this is.

Also on a sidenote, `wlr_pointer_gestures_v1` doesn't have a destroy or free function. It seems it handles destroying itself when the display is destroyed. Is this something worth worrying about?

emersion: >Ah sorry about that. I made a change to correct this, although since the pull request was merged already I'm not sure what the best path to fixing this is.

If you could submit a new PR to fix this, that would be great.

>Also on a sidenote, wlr_pointer_gestures_v1 doesn't have a destroy or free function. It seems it handles destroying itself when the display is destroyed. Is this something worth worrying about?

Yeah, wlr_pointer_gestures_v1 is destroyed when the display is.

PR #4795, opened 2019-12-07
input/cursor: remove gesture listeners in destroy

lolzballs:
Part of #4794. Forgot to remove gesture listeners when the
cursor is destroyed.


ddevault:
Thanks!

Issue #1800, opened 2018-04-12
Detect HiDPI displays and automatically set scale=2


ddevault:
Unless the user specifies it manually, of course. What are the criteria for identifying a HiDPI display (e.g. a minimum ppi)?

NilsBrause:
Since sway has fractional scaling, you could even make the scaling factor a function of the PPI, e.g. scale=ppi/100. That way, everything would have consistent physical dimensions on every display.

SirCmpwn:
We're not going to set a non-integral scale factor by default.

NilsBrause:
Okay. Does fractional scaling have any downsides in sway?

SirCmpwn:
Yes, it has downsides in all situations. We have to either copy pixels or drop pixels, it's impossible to faithfully render clients at a fractional scale factor.

NilsBrause:
I see. In that case, I'd say about 200 ppi. Also scale=3 for 300 ppi and so on.

clarfon:
Perhaps it'd be helpful to show the "real" fractional scale factor in `swaymsg -t get_outputs` at least so people don't have to compute it themselves?

emersion:
I don't see the point. If you don't want to compute it, don't specify it and (when this feature will be implemented) sway will automatically use it.

clarfon:
@emersion I'm specifically talking about the case of fractional scaling; @ddevault already mentioned that fractional scaling will not be the default. So, for people who want to override their display to use fractional scales, rather than rounding to the nearest integer, they can easily figure it out rather than having to compute it.

emersion:
I'd rather not promote fractional scaling.

pat-s:
Coming from i3 since a few days, `sway` already does a perfect job on my mixed DPI setup. I got a 13' Full HD laptop besides a 4k 27' one. The 4k one looks good at around 160 DPI and the laptop at 96 DPI.

I never managed to get a working setup in X before (one was always sharp, the other one blurry due to scaling). With `sway` and `wayland` both look sharp by default. Applying `scale = 2` by default to the 4k would not make my world better.

Or do I overlook something here?

emersion:
You mean "it looks sharp with fractional scaling, why isn't this the default?"

This is definitely not happening. We've already discussed this countless times, fractional scaling has performance issues and is a big lie (no matter how it's implemented). Here's an article that explains the issue: https://medium.com/elementaryos/what-is-hidpi-and-why-does-it-matter-b024eabea20d

[...]

PR #4803, opened 2019-12-11
Auto-detect output scale

emersion:
If the screen DPI is high enough, auto-enable scale=2 (if the user
hasn't set the scale).

Uses heuristics based on [1].

[1]: https://gitlab.gnome.org/GNOME/mutter/blob/05217066171992b0bc50882869aad6385285c878/src/backends/meta-monitor.c#L1590

Closes: https://github.com/swaywm/sway/issues/1800

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -263,6 +263,49 @@ static bool set_mode(struct wlr_output *output, int width, int height,
 	return wlr_output_set_mode(output, best);
 }
 
+/* Some manufacturers hardcode the aspect-ratio of the output in the physical
+ * size field. */
+static bool phys_size_is_aspect_ratio(struct wlr_output *output) {
+	return (output->phys_width == 1600 && output->phys_height == 900) ||
+		(output->phys_width == 1600 && output->phys_height == 1000) ||
+		(output->phys_width == 160 && output->phys_height == 90) ||
+		(output->phys_width == 160 && output->phys_height == 100) ||
+		(output->phys_width == 16 && output->phys_height == 9) ||
+		(output->phys_width == 16 && output->phys_height == 10);
+}
+
+// The minimum DPI at which we turn on a scale of 2
+#define HIDPI_DPI_LIMIT (2 * 96)
+// The minimum screen height at which we turn on a scale of 2
+#define HIDPI_MIN_HEIGHT 1200
+// 1 inch = 25.4 mm
+#define MM_PER_INCH 25.4
+
+static int compute_default_scale(struct wlr_output *output) {
+	int width, height;
+	wlr_output_transformed_resolution(output, &width, &height);
+	if (height < HIDPI_MIN_HEIGHT) {
+		return 1;
+	}
+
+	if (output->phys_width == 0 || output->phys_height == 0) {
+		return 1;
+	}
+
+	if (phys_size_is_aspect_ratio(output)) {
+		return 1;
+	}
+
+	double dpi_x = (double) width / (output->phys_width / MM_PER_INCH);
+	double dpi_y = (double) height / (output->phys_height / MM_PER_INCH);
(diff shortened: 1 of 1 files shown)


tokyovigilante:
@emersion Looks great and very straightforward, however is there any consensus (from Gnome or elsewhere) about the HIDPI_LIMIT DPI of 192ppi? I have a Dell P2415Q which is a 24" 4K monitor at 183 DPI, and think 2x scaling is highly appropriate. Obviously I'm setting a scale factor anyway, but I'd argue 180 dpi at 40-50cm is reasonable to qualify for automatic scaling.

emersion:
>however is there any consensus (from Gnome or elsewhere) about the HIDPI_LIMIT DPI of 192ppi?

192 = 96*2 (96 DPI is the standard LoDPI)

This PR's goal isn't to automatically scale *all* monitors, it's best-effort.

>The call to wlr_output_set_scale (currently L368 just inside the if (scale != wlr_output->scale)) needs to use scale not oc->scale

Oh, my bad. Fixed.

>oc->scale should now be set to 0 in default_output_config otherwise the outputs will still default to 1 after a reload

Hmm I'm not sure I follow. Wouldn't we have the same issue with `enabled` for instance?

RedSoxFan:
> > oc->scale should now be set to 0 in default_output_config otherwise the outputs will still default to 1 after a reload
> 
> Hmm I'm not sure I follow. Wouldn't we have the same issue with `enabled` for instance?

`oc->scale` is set to `1` in `default_output_config`, which is only called when reloading. If the scale is not set in the config, then the scale will be `1` and avoid the detection until something causes the output config to be applied again. In `default_output_config`, it should either be `-1` or `0` not `1` so that the detection is used

emersion:
Oh, silly mistake on my end again, I was reading `new_output_config` instead. Fixed.

tokyovigilante:
> 192 = 96*2 (96 DPI is the standard LoDPI) 
> This PR's goal isn't to automatically scale _all_ monitors, it's best-effort.

No I understand that, just pointing out that's a relatively historical figure and seems to have been chosen for that reason alone (that it's double 96 DPI) by Gnome and whether this is a valid assumption in 2019. 

As noted elsewhere and in that Gnome bug/feature request, Windows and Apple (who arguably popularised the 2x/"Retina" standard for HIDPI scaling) both consider a lower DPI appropriate to scale to 2x, even as low as 150 DPI.  Obviously it also depends on your visual acuity and distance from screen too.

emersion:
Yeah, that's fair. I personally ave a 161 DPI screen and 2x is too much.

tokyovigilante:
Or would you consider making this configurable? I guess at that point you may as well just be setting it as an output option though,

emersion:
The whole point is to not need to configure anything. It's an heuristic, if it doesn't work for some particular cases but still works for most others it's fine.

tokyovigilante:
Yeah of course, I guess that's my point, that using 192 DPI as the cut-off for HIDPI is too high for my monitor, and using 160 DPI would be too low for yours, so I'm just asking if you'd agree that a 24" 4K screen should be scaled by default, and if so would you consider using 180 DPI as the cutoff.  If you want to stick with 192 DPI that's fine. 

The table "High Resolution/High Pixel Density Displays" [here (eizoglobal.com)](https://www.eizoglobal.com/library/basics/pixel_density_4k/) has a pretty big gap between desktop/external monitors at 100-150 dpi, other than 24" 4K (at ~185) and then HIDPI laptop monitors all over 200 dpi.

Unfortunately the Steam HW survey doesn't list DPI, only resolution, and only a handful of users have a screen > 1440p. Obviously gaming vs desktop use is different, but I haven't found any other useful resource for desktop DPI statistics.

RedSoxFan:
Thanks!

PR #4805, opened 2019-12-11
Add seat <seat> idle_{inhibit,wake} <sources...>

ddevault:
This adds seat configuration options which can be used to configure what
events affect the idle behavior of sway.

An example use-case is mobile devices: you would remove touch from the
list of idle_wake events. This allows the phone to stay on while you're
actively using it, but doesn't wake from idle on touch events while it's
sleeping in your pocket.

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -4,6 +4,7 @@
 #include <wlr/types/wlr_layer_shell_v1.h>
 #include <wlr/types/wlr_seat.h>
 #include <wlr/util/edges.h>
+#include "sway/config.h"
 #include "sway/input/input-manager.h"
 
 struct sway_seat;
@@ -79,6 +80,8 @@ struct sway_seat {
 	void *seatop_data;
 
 	uint32_t last_button_serial;
+
+	uint32_t idle_inhibit_sources, idle_wake_sources;
 
 	list_t *deferred_bindings; // struct sway_binding
 
@@ -195,6 +198,9 @@ void seat_apply_config(struct sway_seat *seat, struct seat_config *seat_config);
 struct seat_config *seat_get_config(struct sway_seat *seat);
 
 struct seat_config *seat_get_config_by_name(const char *name);
+
+void seat_idle_notify_activity(struct sway_seat *seat,
+		enum sway_input_idle_source source);
 
 bool seat_is_input_allowed(struct sway_seat *seat, struct wlr_surface *surface);
 
--- a/sway/commands/seat/idle.c
+++ b/sway/commands/seat/idle.c
@@ -1,0 +1,73 @@
+#define _POSIX_C_SOURCE 200809L
+#include <limits.h>
+#include <string.h>
+#include <strings.h>
+#include <stdint.h>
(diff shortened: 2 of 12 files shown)


include/sway/input/seat.h
```diff
@@ -80,6 +81,8 @@ struct sway_seat {
 
 	uint32_t last_button_serial;
 
+	uint32_t idle_inhibit_sources, idle_wake_sources;
```
emersion: `idle_wake_sources` doesn't seem to be used

sway/commands/seat/idle.c
```diff
@@ -0,0 +1,79 @@
+#define _POSIX_C_SOURCE 200809L
+#include <limits.h>
+#include <string.h>
+#include <strings.h>
+#include <stdint.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/input/seat.h"
+
+struct {
```
emersion: Nit: this array can be `static const`

sway/input/seat.c
```diff
@@ -71,6 +72,21 @@ static void seat_node_destroy(struct sway_seat_node *seat_node) {
 	free(seat_node);
 }
 
+void seat_idle_notify_activity(struct sway_seat *seat,
+		enum sway_input_idle_source source) {
+	uint32_t mask = seat->idle_inhibit_sources;
+	struct wlr_idle_timeout *timeout;
+	wl_list_for_each(timeout, &server.idle->idle_timers, link) {
+		if (timeout->idle_state) {
+			mask = seat->idle_wake_sources;
```
emersion: This doesn't work well with multiple timers, each having a different timeout. A 5s screen dim timeout combined with a 30s dpms timeout would break this for instance.

ddevault: The idea behind this code was just to find out if any timer was idle, which it does. So, by design?

emersion: I'm just saying this design breaks timers that are not idle (going to sleep while they shouldn't).

ddevault: I don't understand.

emersion: Let's say touch is not in `idle_wake_sources`. Let's say you want to setup one timer to dim the screen after 5s and another timer to blank the screen after 30s. It's a pretty standard setup people have on phones.

You stop touching the screen for 10s. The screen is dimmed. Then you start touching the screen again. Because the dim timer has expired, the touches don't wake timers up to un-dim the screen. Worse, the blank timer isn't ping'ed and no matter how much you touch your screen it'll blank the screen after 20 more seconds.

ddevault: Got it, thanks for explaining. I'll update this to kick in once all idle timers have expired instead.

emersion: Yeah, that's probably a good solution. I was thinking about more complicated solutions as usual :P

emersion: Maybe at some point we'll need to get to the more complicated solutions though. If you add in a messaging client using the idle protocol to mark the user as away after 1min, things get more complicated. Needs more thought.

PR #4807, opened 2019-12-12
Fix lingering workspace with scratchpad show

### LINKED ISSUE #4806: Lingering empty workspace on scratchpad show
```
$ swaymsg -t get_version
sway version 1.2-ad189d27 (Dec  3 2019, branch 'master')
```
Steps to reproduce:

1. Send a window to the scratchpad
2. Switch to an empty workspace
3. scratchpad show
4. Switch to another workspace
5. scratchpad show

The workspace from step 2 is now empty, but is not destroyed.

RPigott:
Showing a window in the scratchpad can move a visible scratchpad window
from another workspace to the current one. If the scratchpad window was
the last visible container in that workspace, the old workspace should
be destroyed.

Fixes #4806

### DIFF
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -131,6 +131,7 @@ void root_scratchpad_show(struct sway_container *con) {
 	// Show the container
 	if (old_ws) {
 		container_detach(con);
+		workspace_consider_destroy(old_ws);
 	}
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Thanks!

Issue #4811, opened 2019-12-14
Input config merging is broken for repeat_delay


RPigott:
```
swaymsg -pt get_version
sway version 1.2-f9ce8749 (Dec 13 2019, branch 'master')
```
I have the following keyboard
```
Input device: AT Translated Set 2 keyboard
  Type: Keyboard
  Identifier: 1:1:AT_Translated_Set_2_keyboard
  Product ID: 1
  Vendor ID: 1
  Active Keyboard Layout: English (US)
  Libinput Send Events: enabled
```
and in my config I have the following section:
```
input type:keyboard {
    repeat_rate 25
    repeat_delay 200
}
```
The setting works, but it is fragile. It unexpectedly breaks in the following cases:

1. `swaymsg input 1:1:AT_Translated_Set_2_keyboard xkb_layout us`
  - This command _should_ have no effect, but instead repeat delay is reset to the default 600.
  - The repeat_delay can be corrected by a reload
  - Using `type:keyboard` doesn't reset repeat_delay, and in fact will restore the previously reset repeat_delay to the configured 200 ms.
2. Replace `type:keyboard` with an the identifier in the config, then `swaymsg reload`.
  - This _should_ have no effect, but instead the repeat delay is reset to the default 600. (Very surprising!)
  - Cannot be corrected by a reload.
  - Can be corrected by exiting and restarting sway.

I can't find a concise explanation for both (1) and (2), so I'm not sure how they're related.

RedSoxFan:
1. #4138 should fix that issue
2. This one is an oversight in my keyboard group implementation. I'm working on a fix for it

PR #4813, opened 2019-12-14
config/input: set type for new identifier configs

RedSoxFan:
Fixes the first bullet of #4811 

When an input becomes available, the input type config for that device
type will be merged underneath the input identifier config, provided
they both exist. If an input type config gets added or modified at a
later point, then those changes get merged onto the input identifier
configs for that type. However there was a missing case of the input
identifier config being added after the device is already available and
the input type config existing. This makes it so that the first time an
input identifier config gets stored, there will be a check to see if it
matched any of the available devices. If it does, then there will be a
search for the associated input type config, which will be merged
underneath the input identifier config if found.

### DIFF
--- a/sway/config/input.c
+++ b/sway/config/input.c
@@ -249,6 +249,17 @@ static void merge_type_on_existing(struct input_config *type_wildcard) {
 	}
 }
 
+static const char *set_input_type(struct input_config *ic) {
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #4814, opened 2019-12-15
seat_cmd_keyboard_grouping: change keymap to smart

### LINKED ISSUE #4811: Input config merging is broken for repeat_delay

RedSoxFan:
Closes #4811 (this indirectly fixes the second bullet)

This removes `seat <seat> keyboard_grouping keymap` and replaces it with
`seat <seat> keyboard_grouping smart`. The smart keyboard grouping will
group based on both the keymap and repeat info. The reasoning for this
is that deciding what the repeat info should be for a group is either
arbitrary or non-deterministic when multiple keyboards in the group have
repeat info configured (unless somehow exposed to the user in a
reproducible uniquely identifiable fashion).

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -177,9 +177,9 @@ enum seat_config_allow_constrain {
 };
 
 enum seat_keyboard_grouping {
-	KEYBOARD_GROUP_DEFAULT,  // the default is currently keymap
+	KEYBOARD_GROUP_DEFAULT,  // the default is currently smart
 	KEYBOARD_GROUP_NONE,
-	KEYBOARD_GROUP_KEYMAP
+	KEYBOARD_GROUP_SMART  // keymap and repeat info
 };
 
 enum sway_input_idle_source {
--- a/include/sway/input/keyboard.h
+++ b/include/sway/input/keyboard.h
@@ -54,6 +54,9 @@ struct sway_keyboard {
 	struct xkb_keymap *keymap;
 	xkb_layout_index_t effective_layout;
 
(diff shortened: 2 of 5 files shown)


ddevault:
Thanks!

PR #4815, opened 2019-12-16
input/seatop_default: fix focusing floating titles

### LINKED ISSUE #4728: stack in scratchpad: clicking window title selects stack
Clicking a window title in a stack or tab set in the scratchpad selects the parent container instead:

Screencap-2019-15-15-13:28:49

RedSoxFan:
Fixes #4728 

When clicking on the titlebar of a floating container (or descendant of
a floating container), the top-level floating container was being
focused and then allowing you to move the top-level floating container.
This made it so you couldn't switch to a different tab/stack within the
floating container. With this patch, the focus inactive view for the
container that the titlebar is associated with is focused, then the
traversal to the top-level floating container is performed to use with
the move floating operation.


ddevault:
Thanks!

Issue #4065, opened 2019-04-17
Crash on scratchpad move


thepenguinthatwants:
This has happened to me quite many times. 

So if I use few times.

`    bindsym $mod+Shift+minus move scratchpad`

It just logs me out.

RedSoxFan:
Sway is probably segfaulting. If you are using sway 1.0, it may be fixed on master depending on the issue.

After sway crashes, can you try running `coredumpctl gdb sway` followed by `bt full` and posting the results? If you are not using systemd, consult your distros documentation on how to obtain a core dump and use gdb to get the back trace

Also, can you provide a debug log? Start sway with `sway -d >sway.log 2>&1`, reproduce the issue (as quick as possible), and post the log.

norcalli:
Also, make sure to post the version of sway that you were using at the time of the crash. `sway -v` and copy the output.

thepenguinthatwants:
Sorry I had no clue how to copy terminal in wayland and without mouse on alacritty.
[logs](https://snag.gy/WVOcSh.jpg)

sway version 1.0

thepenguinthatwants:
If I start sway with  `sway -d >sway.log 2>&1 ` 
it just open it in wlroots? But i cant reproduce the bug in it?

only way to reproduce is to things up and quickly move multiple things to scratchpad and then it logs me out.

`    bindsym $mod+Shift+minus move scratchpad`

RedSoxFan:
Unfortunately, it looks like sway and wlroots were not compiled with debug symbols so the backtrace is useless. If you could rebuild with debug symbols, that would be helpful.

> If I start sway with `sway -d >sway.log 2>&1` it just open it in wlroots? But i cant reproduce the bug in it?

I'm guessing you are running that from within sway and getting a nested session. It would probably be easier to reproduce from a TTY instead

RedSoxFan:
Is this still an issue?

edrex:
I've been seeing an infrequent crash for the past couple weeks under similar circumstances, while interacting with the scratchpad (showing it, at least the last time). Here's the last few minutes of log for the last crash:

https://clbin.com/NjrHr

`sway version 1.2-rc1-48-ga085e649 (Sep 15 2019, branch 'master')`

zsugabubus:
Seems related, but different stack trace and log to @edrex.

* Sway Version: 1.2

* Debug Log: https://gist.github.com/zsugabubus/22e77e3b02b0bb5d6907cc129345878b

```
[...]

PR #4816, opened 2019-12-16
root_scratchpad_hide: disable fullscreen descendants

RedSoxFan:
I think this fixes @edrex's [crash](https://github.com/swaywm/sway/issues/4065#issuecomment-532890928) in #4065. Either way, this does fix a crash when hiding a scratchpad container when any of it's descendants are fullscreen

Any descendant of a scratchpad container may be fullscreen so checking
to see if the top-level scratchpad container is fullscreen in
root_scratchpad_hide is not sufficient. This iterates through all
descendants of the scratchpad container

### DIFF
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -149,6 +149,12 @@ void root_scratchpad_show(struct sway_container *con) {
 	seat_set_focus(seat, seat_get_focus_inactive(seat, &con->node));
 }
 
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

edrex:
oh, it's fullscreening descendants before hiding scratchpad that causes it, great catch and thanks!

PR #4833, opened 2019-12-22
Have commands return cmd_results on the stack

mstoeckl:
> sway commands are implemented using functions that return an error code and (if the function was not successful) an error string. The two are bundled together by the type `struct cmd_results`. This
patch alters the command handler prototype so that the cmd_results objects are returned by value (on the stack), instead of by a pointer to a heap allocated instance of a cmd_results.

> The latter method had the flaw that, if the heap allocation of the cmd_results object failed, the exact return code would be lost. Furthermore, for some command handlers (such as those in sway/commands/output), returning NULL signified success, so an allocation failure could lead to an ignored error. This change prevents both classes of errors.

I've been sitting on this patch for a while, as allocation failures are very rare in practice and require nonstandard system configuration to achieve. This isn't the largest allocation-related issue to fix (as it won't even cause a crash), but one must start somewhere.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -6,7 +6,7 @@
 
 struct sway_container;
 
-typedef struct cmd_results *sway_cmd(int argc, char **argv);
+typedef struct cmd_results sway_cmd(int argc, char **argv);
 
 struct cmd_handler {
 	char *command;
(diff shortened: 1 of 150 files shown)


ddevault:
I'm not sure that this change is worth such a large diff, which is likely to introduce bugs to otherwise stable code. I think we'll pass. Behaving well under memory pressure is not something most applications can cope with anyway.

PR #4846, opened 2019-12-27
[WIP] Output-namespaced workspaces

### LINKED ISSUE #4771: Two Alternative Multi-Monitor Setups for Sway
This is a **suggestion** for two alternate ways of switching between workspaces (virtual desktops) in Sway.

So far, as much as I want to like Sway, I greatly dislike how the workspaces are handled. Right now, they closely follow the implementation of vanilla i3, where workspaces are shared, but separated between monitors (or screens).

For example, I have two monitors, and both workspace 1 and workspace 2 are automatically created on each monitor when I start Sway.

When the user presses Mod+1 it will focus on the workspace on the "first monitor" and if you press Mod+2, it will focus on the "second monitor".

If anything else but Mod+1 and Mod+2 are pressed, a new workspace is created on the currently "focused monitor", and empty workspaces are removed if there are no windows in the workspace, which makes sense.

However, you can imagine how lost someone would be if 4 workspaces were automatically created among 4 monitors, and since there are only 10 digits in most English/cardinal languages, having 10 workspaces only accessible with a minimum of two key-presses can be very limiting, especially when divided among 4 monitors.

Also, new users have to keep remembering which workspaces are created on which screens, which are removed, which exist, and which windows are open in each workspace, on each screen! It's frustrating to say the least!

With all that in mind, here's some alternatives to, in my opinion, this very counter-intuitive, default method.

## Alternative 1 - Grouped workspaces per-monitor

Each monitor can be treated as a separate group. For example, they could be called **Workspace Groups**. Let's say that workspace 1-1 is workspace 1 on the first monitor, and workspace 2-1 is workspace 1 on the second monitor. You can see where this is going....

Applications/Windows can easily be separated between screens, yet still, because you can display whichever workspaces you want for each monitor, it is very flexible for the user to show and hide whichever workspaces you want, and it offers an option for 10 easily accessible workspaces per-monitor (or to be more precise, per-group)!

In order to switch between workspace groups, you could use some unused keybinds, that involve arrow keys or numbering, or perhaps you could just have your mouse cursor on whichever monitor you wanted, in order to specify the group.

Either way, it just makes a lot of sense. Even if it may be difficult to implement.

## Alternative 2 - Workspaces are shared among all screens

Instead of separating workspaces between screens, why not use each workspace for all screens? This is actually somewhat similar to how most DEs and WMs work, such as Openbox and XFCE, and it should be easier to implement!

Like the default implementation, the same amount of keystrokes are required to switch between groups. But you would have a maximum of 10 groups before needing additional keypresses to access more workspaces.

Both the default and this implementation do not scale well the more screens you have, but it should be a lot easier to implement in the meantime.

## Conclusion

I honestly think all 3 methods should be implemented, to give the power-user more options, but the way it works right now is great for veteran i3 users. I am a new user of Sway/i3, so I have a fresh perspective of how these things should work.

I was just really disappointed of how it handles right now. Sure, it's great for single-monitor use, but Sway still extremely fails in terms of dual-monitor, or higher, and I keep going back to other WMs because of it.

I hope the developers consider these options, and I hope non-developers and other uses can support decisions like these, at the very least.

Thank you for reading!

kennylevinsen:
This PR implements configurable workspace namespacing. There are two namespace options, `global` and `output`. `global`, the default, behaves exactly like sway has always done.

`output`, on the other hand, behaves a bit more like tags from awesomewm, in that workspaces are local to an output.

I have been running this for a while, and although a very simple change, it has made multiple workspaces on my laptop (which docks to a varying number of screens) *much* more pleasant, where it previously was in fact very frustrating.

Fixes #4771

# A bit of explanation

## Global

Imagine a setup with 3 displays. With `global` namespacing, these will be born with workspaces "1", "2" and "3", with the order appearing from a user-perspective to be somewhat arbitrary. if you run `workspace 2`, focus will go to display 2. If you run `workspace 4`, a workspace named `4` will be created on whatever output is focused, perhaps leaving you with:

display | workspaces
----------|-----------------
1          | 2
2          | 1
3          | 3, 4

With dynamic screen setups and frequent use of namespaces, this ends up chaotic rather fast. This is made much worse should one decide to hide or not use a bar, leaving them with no map of the mess of workspaces that they have created.

This can be managed for static dual-screen setups by creating output-specific navigation binds (e.g. `mod+NUM` for output 1, `mod+shift+NUM` for output 2), but this is merely a workaround and does not scale to arbitrary or dynamic display configurations.

`prev_on_output`/`next_on_output` tries to assist with this, but falls short as it will not create new workspaces and only allows `alt-tab`-style navigation.

## Output namespacing

With `output` namespacing, each output has its own set of workspaces, and thus all displays start with its very own workspace "1". If you run `workspace 2`, a workspace named "2" will be created on the current output. Thus, you'd end up with:

display | workspaces
----------|-----------------
1          | 1
2          | 1
3          | 1, 2

This provides for very predictable behavior regardless of display configuration. This 

## `output` behaviors worth noting

- Navigating to a workspace will never change what output is in focus. It will either go to the workspace if it exists on the focused output, or create it on the focused output.
- Moving a workspace between outputs can result in name conflicts, which are automatically solved by using `workspace_next_name`. This also applies to move done due to an output disconnecting.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -190,6 +190,7 @@ sway_cmd cmd_unmark;
 sway_cmd cmd_urgent;
 sway_cmd cmd_workspace;
 sway_cmd cmd_workspace_layout;
+sway_cmd cmd_workspace_namespace;
 sway_cmd cmd_ws_auto_back_and_forth;
 sway_cmd cmd_xwayland;
 
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -471,6 +471,11 @@ enum xwayland_mode {
 	XWAYLAND_MODE_IMMEDIATE
 };
 
+enum sway_workspace_namespace {
+	WORKSPACE_NAMESPACE_GLOBAL,
+	WORKSPACE_NAMESPACE_OUTPUT,
+};
+
 /**
  * The configuration struct. The result of loading a config file.
  */
@@ -512,6 +517,7 @@ struct sway_config {
 	enum sway_fowa focus_on_window_activation;
 	enum sway_popup_during_fullscreen popup_during_fullscreen;
 	enum xwayland_mode xwayland;
+	enum sway_workspace_namespace workspace_namespace;
 
 	// swaybg
 	char *swaybg_command;
--- a/include/sway/tree/workspace.h
+++ b/include/sway/tree/workspace.h
@@ -58,14 +58,14 @@ void workspace_begin_destroy(struct sway_workspace *workspace);
 
 void workspace_consider_destroy(struct sway_workspace *ws);
 
-char *workspace_next_name(const char *output_name);
+char *workspace_next_name(struct sway_output *output);
 
 bool workspace_switch(struct sway_workspace *workspace,
 		bool no_auto_back_and_forth);
 
-struct sway_workspace *workspace_by_number(const char* name);
+struct sway_workspace *workspace_by_number(struct sway_output *output, const char* name);
 
-struct sway_workspace *workspace_by_name(const char*);
+struct sway_workspace *workspace_by_name(struct sway_output *output, const char* name);
 
 struct sway_workspace *workspace_output_next(
 		struct sway_workspace *current, bool create);
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -105,6 +105,7 @@ static struct cmd_handler config_handlers[] = {
 	{ "swaybg_command", cmd_swaybg_command },
 	{ "swaynag_command", cmd_swaynag_command },
 	{ "workspace_layout", cmd_workspace_layout },
+	{ "workspace_namespace", cmd_workspace_namespace },
 	{ "xwayland", cmd_xwayland },
 };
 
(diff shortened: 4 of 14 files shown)


kennylevinsen:
From @ddevault in #4771:
> We're not interested in diverging this far from i3.

Assuming i3 isn't under complete feature-freeze these days (official or not), then I might be willing to implement it for i3 as well, should its existence there be a hard requirement. In sway at least, the core of the feature is only 4 lines written twice (signature changes generate some noise, though).

With that in mind, could we re-open for discussion? I don't really see there being any other good approach to handling workspaces across dynamic screen counts, which I believe is a problem we should make *some* solution for.

dbrgn:
Thanks @kennylevinsen for the implementation, this is exactly the behavior I'm missing from dwm, and I find the sway workspace management quite impractical.

Are you maintaining your patch somewhere? Did you start implementing this feature for i3?

kennylevinsen:
This patch still applies cleanly. i3 and sway are both in feature freeze though.

Script approaches are also available in the other thread.

fort-ified:
It doesn't appear to apply anymore. Hunk 14 fails to apply, creating /sway/tree/workspace.c.rej containing:
```diff
--- sway/tree/workspace.c
+++ sway/tree/workspace.c
@@ -816,4 +831,4 @@ size_t workspace_num_tiling_views(struct sway_workspace *ws) {
    size_t count = 0;
    workspace_for_each_container(ws, count_tiling_views, &count);
    return count;
-}
+}
```
It looks like the external methods are probably the safest going forward.

PR #4847, opened 2019-12-28
Refactor color parsing related code

RedSoxFan:
There are currently multiple color parsing functions in use by sway. This purpose of this PR is to remove some duplicated code and refactor some of the calling code along the way. This PR also ensures that although the colors may be stored in different formats (uint32_t, float[4], const char *), they are all parsed in the same way.

See the individual commits for the full commit messages, but here's a summary:
- First commit: adds color validation to parse_color and refactors swaybar's ipc_parse_colors
- Second commit: remove duplicated float array color code in cmd_client_* and config
- Third commit: remove add_color and refactor bar_cmd_colors color parsing

_Note: I am aware that this PR does not touch the color validation for colors in `output_cmd_background`. For that command, the colors must be rgb, not rgba. Since all other locations allow for rgb or rgba, it would introduce unnecessary code complexity to remove this duplication._

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -1,4 +1,5 @@
 #define _POSIX_C_SOURCE 200809L
+#include <ctype.h>
 #include <float.h>
 #include <fcntl.h>
 #include <math.h>
@@ -13,21 +14,21 @@ int wrap(int i, int max) {
 	return ((i % max) + max) % max;
 }
 
-uint32_t parse_color(const char *color) {
+bool parse_color(const char *color, uint32_t *result) {
 	if (color[0] == '#') {
 		++color;
 	}
-
 	int len = strlen(color);
-	if (len != 6 && len != 8) {
-		sway_log(SWAY_DEBUG, "Invalid color %s, defaulting to color 0xFFFFFFFF", color);
-		return 0xFFFFFFFF;
+	if ((len != 6 && len != 8) || !isxdigit(color[0]) || !isxdigit(color[1])) {
+		return false;
 	}
-	uint32_t res = (uint32_t)strtoul(color, NULL, 16);
-	if (strlen(color) == 6) {
-		res = (res << 8) | 0xFF;
+	char *ptr;
+	uint32_t parsed = strtoul(color, &ptr, 16);
+	if (*ptr != '\0') {
+		return false;
 	}
-	return res;
+	*result = len == 6 ? ((parsed << 8) | 0xFF) : parsed;
+	return true;
 }
 
 bool parse_boolean(const char *boolean, bool current) {
--- a/common/util.c
+++ b/common/util.c
@@ -31,6 +31,13 @@ bool parse_color(const char *color, uint32_t *result) {
 	return true;
 }
 
+void color_to_rgba(float dest[static 4], uint32_t color) {
+	dest[0] = ((color >> 24) & 0xff) / 255.0;
+	dest[1] = ((color >> 16) & 0xff) / 255.0;
+	dest[2] = ((color >> 8) & 0xff) / 255.0;
+	dest[3] = (color & 0xff) / 255.0;
+}
+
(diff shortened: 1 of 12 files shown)


common/util.c
```diff
@@ -13,21 +13,28 @@ int wrap(int i, int max) {
...
 	if (color[0] == '#') {
 		++color;
 	}
-
 	int len = strlen(color);
-	if (len != 6 && len != 8) {
-		sway_log(SWAY_DEBUG, "Invalid color %s, defaulting to color 0xFFFFFFFF", color);
-		return 0xFFFFFFFF;
+	char *ptr;
+	uint32_t parsed = strtoul(color, &ptr, 16);
```
mstoeckl: [`strtoul`](https://en.cppreference.com/w/c/string/byte/strtoul) ignores whitespace before the number, so `parse_color` would accept "`# fffff`" (with one space) as a valid color. "`#0x0000`" also passes.

RedSoxFan: Thanks. `strtoul` was already being used so I didn't think about it. I changed it so it verifies the first two characters (excluding the `#`) are hex digits. That will catch signs, leading whitespace, and `0x`. Anything further into the string will be caught by the `*ptr != '\0'`.

sway/commands/bar/colors.c
```diff
@@ -16,38 +18,58 @@ static struct cmd_handler bar_colors_handlers[] = {
...
+
+	char buffer[10] = {0};
+	if (!hex_to_rgba_hex(argv[0], buffer)) {
+		return cmd_results_new(CMD_INVALID, "Invalid color: %s", argv[0]);
 	}
-	error = add_color(*color, argv[0]);
-	if (error) {
-		return error;
+
+	if (!*color && !(*color = malloc(sizeof(buffer)))) {
```
emersion: We can just use `strdup` here

emersion: `hex_to_rgba_hex` could return the dup'ed string btw, this would keep the static buffer + dup logic inside this function.

RedSoxFan: Moved the malloc inside `hex_to_rgba_hex` and removed the static buffers

sway/commands/bar/colors.c
```diff
@@ -16,38 +18,53 @@ static struct cmd_handler bar_colors_handlers[] = {
 	{ "urgent_workspace", bar_colors_cmd_urgent_workspace },
 };
 
+static char *hex_to_rgba_hex(const char *hex) {
+	uint32_t color;
+	if (!parse_color(hex, &color)) {
+		return NULL;
+	}
+	char *rgba = malloc(10);
```
craigbarnes: No `NULL` check?

emersion: Send a PR to add one?

craigbarnes: Sure, I can do. Do `ENOMEM` errors have to be propagated via `cmd_results`? If so, it seems calling `cmd_results_new()` would fail with `ENOMEM` too. Is there a way to signal failure without allocating?

emersion: Hmm, that's an issue. Could we return a NULL result to indicate an allocation failure?

craigbarnes: I guess so. It seems this pattern is just being copied from elsewhere in the codebase though.

For example:

https://github.com/swaywm/sway/blob/31a83bd48d9aad2a039565fc39f1a52bd0cdf17d/sway/criteria.c#L50-L53

https://github.com/swaywm/sway/blob/31a83bd48d9aad2a039565fc39f1a52bd0cdf17d/sway/criteria.c#L504-L507

A quick search with `git grep '(re|c|m)alloc *\(' -- sway/` shows up dozens more.

emersion: Yes, NULL checks are missing throughout the whole codebase.

PR #4858, opened 2019-12-31
Fix wlroots version dependency

bitraid:


ddevault:
I just pushed a proper fix for this to master. Thanks anyway!

PR #4876, opened 2020-01-04
Set DesktopNames in sway.desktop to have `XDG_CURRENT_DESKTOP`

morxa:
Setting DesktopNames in the desktop file sets the environment variable
`XDG_CURRENT_DESKTOP`, which can be used to find out the name of the
currently running session.

See also https://bugzilla.gnome.org/show_bug.cgi?id=727546 and https://bugzilla.redhat.com/show_bug.cgi?id=1787407.

We may want to allow additional names, e.g., Sway. I can update this PR if desired.

### DIFF
--- a/sway.desktop
+++ b/sway.desktop
@@ -3,3 +3,4 @@ Name=Sway
 Comment=An i3-compatible Wayland compositor
 Exec=sway
 Type=Application
+DesktopNames=sway


ddevault:
This isn't our responsibility, this is something the login manager should do. Arguably it shouldn't do it, because this isn't a spec, no one has written it down, and software which relies on it is wrong.

thebitstick:
According to the [XDG spec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys), the login manager sets the variable found in `DesktopNames` in the session file.
So this should be re-opened.

morxa:
Thanks for finding it in the spec! I agree, but I'll leave this up to @ddevault.

emersion:
>XDG_CURRENT_DESKTOP otoh takes a list of compatible desktop identifiers, used for checking what to show of .desktop file entries.

People shouldn't use this to detect the current desktop.

alebastr:
I was trying to make `xdg-desktop-portal-wlr` package work out of the box and found another reason to reconsider this PR:

`xdg-desktop-portal-wlr` requires both `XDG_CURRENT_DESKTOP=sway` and a call to `dbus-update-activation-environment --systemd WAYLAND_DISPLAY XDG_CURRENT_DESKTOP` from sway config.
Sway config part could be shipped downstream (at least Arch [already does that](https://git.archlinux.org/svntogit/community.git/tree/trunk/50-systemd-user.conf?h=packages/sway), although they don't copy `XDG_CURRENT_DESKTOP`), but initialization of the variable should be addressed here.

Darkspirit:
Either you reopen this or [Firefox](https://searchfox.org/mozilla-central/rev/df4c90d4b8c92c99f76334acfe4813c573c12661/widget/GfxInfoX11.cpp#348) can't detect you.

emersion:
>Either you reopen this or Firefox can't detect you.

This is actually a very good reason *not* to reopen this. Clients like Firefox should really not detect the current compositor.

Darkspirit:
Firefox whitelists features for desktop environments that are proven bug free. It does not blacklist broken ones as it causes more trouble for users. WebRender (OpenGL rendering) rollout will happen incrementally by desktop environment. If you don't merge this, you won't get hardware rendering by default at some point.

Darkspirit:
Also, if your users report bugs and upload their about:support page, their desktop environment entry is empty. It's unlikely that every user is manually asked for it, so bugs might get mistriaged and not further debugged. All Sway bugs should ideally end up here: https://bugzilla.mozilla.org/show_bug.cgi?id=wayland-sway

emersion:
This is really broken behaviour and a Firefox bug. Reminds me of obnoxious websites whitelisting some browsers. (Not to mention that XDG_CURRENT_DESKTOP is unreliable and only set when users have a display manager enabled.)

emersion:
Reported https://bugzilla.mozilla.org/show_bug.cgi?id=1640115

Darkspirit:
Yes, it might not be nice, but it's the common way. I guess systems without display manager are not officially supported, so it's up to the user to enable and disable certain preferences to achieve the correct feature set.

emersion:
>the common way

The _common_ way? How is this _common_? Firefox is the very first Wayland client doing this.

> I guess systems without display manager are not officially supported

How has a display manager anything to do with Firefox? Firefox doesn't interact at all with the display manager.

By the way, display managers are unsupported by Sway (which does interact directly with them). So I guess that makes it impossible for Firefox to officially support Sway.

This is getting quite ridiculous.

Darkspirit:
Good to have brought this up, I almost wouldn't have reported this to you.
(I'm not a Firefox developer. Now I feel too anxious and probably won't help you further.)

fwsmit:
Is there still a good reason to not set `XDG_CURRENT_DESKTOP`? From what I can tell the only reason for not setting it is the fear of landing on a whitelist or blacklist. But there is a legitimate use case for it in xdg-desktop-portal-wlr. Why make the configuration of this more difficult than it has to be?

I think we should reconsider this issue at least.

kennylevinsen:
People might have other preferences (e.g. setting it to KDE to have Qt rather than Gtk portals for file opening/printing, to have launchers show the proper applications, etc.). We only set environment variables that are needed to make clients able to talk to us (`SWAYSOCK`, `WAYLAND_DISPLAY`). That `xdg-desktop-portal` chose to rely on this is unfortunate.

Just set it to what you want before starting sway, or before starting the application in question. There are plenty of other environment variables that you might want to set (`MOZ_ENABLE_WAYLAND`, `QT_QPA_PLATFORM` , whatever else you fancy), no need for us to hardcode just one of them.

smlx:
`XDG_CURRENT_DESKTOP` is documented in [xdpw's FAQ](https://github.com/emersion/xdg-desktop-portal-wlr/wiki/FAQ#how-do-i-run-xdpw) and [Troubleshooting guide](https://github.com/emersion/xdg-desktop-portal-wlr/wiki/%22It-doesn't-work%22-Troubleshooting-Checklist).

bl4ckb0ne:
@jessesimpson36 you can edit the wiki to add that

jessesimpson36:
I tried editing the wiki, but it seems I do not have permission.  If you have feedback about the description, I'd be happy to change it to better reflect the findings of this discussion.
The diff I tried changing was in Home.md after line 490


### XDG_CURRENT_DESKTOP environment variable is not being set 

Some applications may rely on the XDG_CURRENT_DESKTOP environment variable to be set so that the application can accommodate for being ran under sway. One example of this is using Screensharing in Firefox. To set this variable, one option is to modify your shell's profile file to export this environment variable before manually running `sway`. If you use a login manager, the login manager will use the DesktopNames attribute inside the `sway.desktop` file.  This file may be modified to add `DesktopNames=sway`. 

It's not good practice for applications to make use of XDG_CURRENT_DESKTOP in this way, and Firefox has an issue open to try to fix it. Refer to (xpdw's FAQ)[https://github.com/emersion/xdg-desktop-portal-wlr/wiki/FAQ#how-do-i-run-xdpw] and (xdpw's Troubleshooting guide)[https://github.com/emersion/xdg-desktop-portal-wlr/wiki/%22It-doesn't-work%22-Troubleshooting-Checklist] if your issue is specifically related to screensharing.

emersion:
Ah, yes, the wiki was locked down because of recent abuse. I've unlocked it again.

jessesimpson36:
Oh, I'm sorry to hear it.
I edited the wiki. Thanks for allowing me to make this change.

PR #4879, opened 2020-01-05
Add missing items to documentation for GET_TREE response in ipc.

tmccombs:
A few items in the response for the get_tree ipc call were missing from
the documentation. This adds some documentation for them.


tmccombs:
Looking at these, I'm also wondering if fullscreen_mode should be a string instead of an integer for the enumeration. Changing would be backward incompatible though...

emersion:
>Looking at these, I'm also wondering if fullscreen_mode should be a string instead of an integer for the enumeration. Changing would be backward incompatible though...

Would be incompatible with i3.

PR #4883, opened 2020-01-05
render: do not render indicator on floating views

### LINKED ISSUE #4860: Floating client.focused in front of full window has the right border use the indicator color
> sway version 1.2-ed022615 (Jan  1 2020, branch 'master')

I have had this issue since sway 1.1 (when I first started using sway), across multiple systems (both using amdgpu and nouveau). This seems to happen with every application ran. I used a custom config to highlight the issue further.

swaybug1
swaybug3

In both of these I am focusing the floating alacritty window, and the red color is "client.focused indicator".

* [Config](https://pastebin.com/5vsk4rjr)
* [Log](https://pastebin.com/hmiyJ5H5) of me just moving a few windows around

RedSoxFan:
Fixes #4860 

This adds a check to make it so the indicator is only rendered on views
with a parent, which floating views do not. Since floating views do not
have a parent, the workspace layout was being incorrectly used to
determine whether to show the split indicator previously. This has no
impact on floating containers and a view within a floating container
will still have indicators rendered appropriately.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -339,7 +339,7 @@ static void render_view(struct sway_output *output, pixman_region32_t *damage,
 		container_current_parent_layout(con);
 
 	if (state->border_right) {
-		if (siblings->length == 1 && layout == L_HORIZ) {
+		if (con->current.parent && siblings->length == 1 && layout == L_HORIZ) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #4888, opened 2020-01-06
Update licence year

ReekyMarko:
Signed-off-by: Marko Korhonen <marko.korhonen@reekynet.com>


emersion:
Thanks for taking the time to open a PR, but wdon't do that.

ReekyMarko:
What's the reasoning behind that? Perhaps it would be better to leave out the year from the license file altogether then

Issue #4892, opened 2020-01-07
Weird behavior with Celeste


YaLTeR:
* Sway Version: 1.2-1e2a2b07 (Jan  7 2020, branch 'master')
* Debug Log: https://gist.github.com/YaLTeR/47c1de76ed2393eb0a1e2802670b6dae
* Configuration File: [1](https://github.com/YaLTeR/dotfiles/blob/9f3322e0b46063309b551f6b018796a1e4b77580/common/.config/sway/config) [2](https://github.com/YaLTeR/dotfiles/blob/9f3322e0b46063309b551f6b018796a1e4b77580/desktop/.config/sway/local)

When starting [Celeste](https://store.steampowered.com/app/504230/Celeste/) with full-screen (default), _sometimes_:
- it starts on the first workspace instead of the current one,
- it becomes impossible to switch to the first workspace until it's killed.

I could not reproduce it with the default config. Also removing most auto-started applications from my sway config makes it harder to reproduce. In the attached log I start sway, then Steam is started, I launch Celeste (from the "9 — steam" workspace), which opens on the "1 — browser" workspace, then I fruitlessly try switching to that workspace multiple times.

YaLTeR:
The second point was indeed fixed, but the game still spawns on the first workspace instead of the current one most of the time for some reason.

RedSoxFan:
So on a second look, it appears this is what's happening with the first bullet:
- Steam being launched via `exec steam` in your config
- There is a PID mapping created for steam to the first workspace (`1 - Browser`) since that's what is focused when `exec steam` is ran
- Steam maps without the X11 property  `_NET_WM_PID` set
- Steam is assigned to workspace `11 - Steam` due to the `assign` in your config
- Since the view has a PID of 0 (unknown), the PID mapping for Steam remains
- Celeste is mapped with the X11 property  `_NET_WM_PID` set
- Celeste is assigned to workspace `1 - Browser` because there is a PID mapping for it's ancestor (Steam) to `1 - Browser`

I'm not really sure what can be done for this. For xwayland views, if the `_NET_WM_PID` property is not set, there is no way to identity the owning process. There are several applications that do fork before mapping so removing matching for ancestor PIDs isn't desirable.

Perhaps a flag to `exec`/`exec_always` to tell Sway not to store a PID mapping for that command would work?

emersion:
What is i3's behavior, @YaLTeR?

YaLTeR:
Not sure, I've never used i3.

[...]

PR #4893, opened 2020-01-08
view: remove workspace pid mapping for assigns

RedSoxFan:
Related #4892 (fixes first bullet)

If a view is mapped to a workspace using an assign, the pid should still
be removed from the pid mapping list. This prevents child processes from
matching against it and mapping a view to a likely undesired workspace.

### DIFF
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -225,6 +225,13 @@ static pid_t get_parent_pid(pid_t child) {
 	return -1;
 }
 
+static void pid_workspace_destroy(struct pid_workspace *pw) {
+	wl_list_remove(&pw->output_destroy.link);
+	wl_list_remove(&pw->link);
+	free(pw->workspace);
+	free(pw);
+}
+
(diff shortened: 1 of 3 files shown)


sway/tree/root.c
```diff
@@ -320,6 +320,23 @@ void root_record_workspace_pid(pid_t pid) {
...
 
+void root_remove_workspace_pid(pid_t pid) {
+	if (!pid_workspaces.prev || !pid_workspaces.next) {
+		return;
+	}
+
+	struct pid_workspace *pw, *tmp;
+	wl_list_for_each_safe(pw, tmp, &pid_workspaces, link) {
+		if (pid == pw->pid) {
+			wl_list_remove(&pw->output_destroy.link);
```
emersion: It seems like we do this multiple times. Would be nice to extract it into a `pid_workspace_destroy` function.

RedSoxFan: Done

PR #4894, opened 2020-01-09
(contrib/inactive-windows-transparency.py): Use transparency arg in the window focus event

jboyens:
Currently the transparency passed in on the command line is not being used in the `on_window_focus` event. This is unexpected since upon start-up, all the windows will be set to the correct transparency, but then switching between windows will set it to the default of 0.8.

It can be tested by running the script with `-o 0.1` to view the incorrect behavior, apply the patch, and then run the same test again to see the now correct behavior.

### DIFF
--- a/contrib/inactive-windows-transparency.py
+++ b/contrib/inactive-windows-transparency.py
@@ -9,9 +9,9 @@ import argparse
 import i3ipc
 import signal
 import sys
+from functools import partial
 
-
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks! Nice use of partial.

tmpm697:
I have issue with killing this script but firefox-beta window still keeps its transparency property.

PR #4897, opened 2020-01-10
integer scaling, not integral

7596ff:
Integral scaling doesn't make any sense. If this is actually how the downscaling works, ignore, but I have a feeling it is simply integer scaling.


ddevault:
Thanks!

Issue #4798, opened 2019-12-10
Nearest/smart scale_filter flickering back and forth linear/nearest


luispabon:
* Sway Version:
sway: ad189d27f97076c11e87ecd3b89859a49f58b031
wlroots: 8681e4ab8a2d4c95abd34abf0e0eed3351d11bf0
* Debug Log:
n/a

* Configuration File:
```
sway output eDP-1 pos 0 0 res 3840x2160 scale_filter nearest
```

* What happened

I've been testing out the recent work to xwayland scaling filters (https://github.com/swaywm/sway/pull/4727) for the last couple of weeks.

I have found one issue. Scenario: 
 * `sway output eDP-1 pos 0 0 res 3840x2160 scale_filter nearest`, also happens with smart)
 * Open two xwayland apps side by side (for some reason it won't work if the other app is wayland native). It seems to happen more when two apps are side by side, not just one

When you click on app A (for instance pycharm), you can see app B changing its scaling from `nearest` to `linear`. Causes a very noticeable flickering. Sometimes it's only portions of that window that will go blurry, sometimes it's the whole screen. On Pycharm, you sometimes it's the currently selected line in unison with the cursor blinking.

It's impossible to screenshot. I'll look for a way to record a vid of it.

Before running this new feature, I had simply patched wlroots to always use nearest. As I understand, this patch works in conjunction with https://github.com/swaywm/wlroots/pull/1914/files (which is the line I was patching to use NEAREST) in the understanding that the default is linear anyway. When I patched that line (eg was always set explicitly) there was no flickering at any point.

* Stack Trace, if sway crashes:
n/a

emersion:
Maybe we're missing a call to `set_scale_filter` prior to `render_texture`. To fix this, I'd prefer to move the `set_scale_filter` call from inside `render_texture` so that we're sure not to miss new calls.

luispabon:
https://www.youtube.com/watch?v=CBKAGahU8yw&feature=youtu.be --> make sure you're viewing the 4k stream on a 4k display, the flickering is a much harder to see at 1080p or less.

luispabon:
@emersion do you mean calling `set_scale_filter` from `render_texture`? I tried it here and that didn't do it: https://github.com/swaywm/sway/commit/39a2e3998e47bc74a0095f15cd5bb5051fa2211b

tokyovigilante:
[...]

PR #4903, opened 2020-01-13
Fix unreliable scaling filter

emersion:
Two changes were made:

- Bind the texture before glTexParameteri
- Set the scaling filter before each wlr_render_texture_with_matrix call

Logging in wlroots allows to check that the scaling filter is properly
set prior to rendering.

Fixes: 6968fb3123e6 ("add scale_filter output config option")
Closes: https://github.com/swaywm/sway/issues/4798

cc @RPigott


RPigott:
Awesome! Looks good, and I cannot reproduce the flickering once applied. Thank you!

ddevault:
Thanks!

PR #4905, opened 2020-01-14
container: add inactive fullscreen to focus stack

### LINKED ISSUE #4823: Fullscreen containers do not receive keyboard input after move
```
swaymsg -pt get_version
sway version 1.2-218b5b9d (Dec 15 2019, branch 'master')
```
When moving a fullscreen container from an offscreen workspace to the active workspace, the active workspace retains keyboard focus, leaving the fullscreen container to cover the screen in an unusual split focus state.

Possibly related to what the user in #4753 describes.

Steps to reproduce:
1. Open something on workspace 1. `swaymsg exec $term`.
2. Open something else and fullscreen it. `swaymsg exec mpv, fullscreen`
3. Move the fullscreen client to an offscreen workspace. `swaymsg move to workspace 2`.
- as expected: The fullscreen container remains fullscreen, and focus on workspace 1 returns to its rightful place on the focus stack, $term.
4. Move the fullscreen container back to the active workspace. `swaymsg move '[app_id=mpv]' to workspace 1`.
- as expected: The fullscreen container returns to the active workspace and covers the screen.
5. Send any key, e.g. 'q'.
- expected:
Pressing 'q' (mpv keybind for quit) will quit the client.
- actual:
Pressing 'q' inserts 'q' into the previously focused $term behind mpv.
6. Kill the fullscreen client, e.g. `swaymsg kill`.
- expected:
Given the results of (5), expect that a `kill` keybind (or inserting `swaymsg kill<RETURN>` in the background $term) will kill the focused $term.
- actual
`kill` keybind or `swaymsg kill` in focused $term kills the fullscreen client instead.

I also observed more strange behavior with split focus fullscreen containers. To reproduce, create a split focus fullscreen container as in (1-4) and:

1. Because the ipc still affects the fullscreen window, it is possible to retain the split focus even when no container is fullscreen. Use `swaymsg fullscreen` to return the fullscreen container to the tiling area, and observe it is still possible to type `swaymsg kill` in $term, but have it affect the previously fullscreened container. Other commands that unfullscreen as a side effect, such as `scratchpad show` also retain the split focus behavior.

2. With `layer=overlay` in mako config, showing a desktop notification (`notify-send some thing`) while an unfocused fullscreen container is visible leaves the desktop a shattered mess. The tiling and floating containers are visible, the fullscreen container is no longer visible but the layer surface area of mako is filled with the content of the fullscreen container and the mako notification composited on top of that.

### LINKED ISSUE #4892: Weird behavior with Celeste

RedSoxFan:
Fixes #4892 
Fixed #4823 

When a container was being made fullscreen and it is on the focused
workspace for a seat, focus was being set to the container. However,
when the container was on a non-focused workspace, the focus stack
wasn't being touched. When assigning a fullscreen container to a
workspace or moving a fullscreen container to a different workspace,
this would make it so the fullscreen container was never added to the
focus stack for the workspace thus preventing access to the workspace.
This adds the container to the top of the focus stack, behind the
container on the focused workspace.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -958,6 +958,11 @@ static void container_fullscreen_workspace(struct sway_container *con) {
 			focus_ws = seat_get_focused_workspace(seat);
 			if (focus_ws == con->workspace) {
 				seat_set_focus_container(seat, con);
+			} else {
+				struct sway_node *focus =
+					seat_get_focus_inactive(seat, &root->node);
+				seat_set_raw_focus(seat, &con->node);
+				seat_set_raw_focus(seat, focus);
 			}
 		}
 	}


ddevault:
Thanks

PR #4910, opened 2020-01-15
Typos

RPigott:
Here are some typos I noticed while working on sway. 

1. Error message copy/paste error in commands/split.c
2. wlr_output is already defined wlr_output = output->wlr_output 3 lines above.
3. The repeat_rate and repeat_delay descriptions are duplicated in the sway-input man page. I think the ones in the libinput configuration section are a mistake, but I liked their wording a bit better so I put those descriptions in the xkb section in place of the ones already there.
4. 1. unbindsym supports the same bind flags as bindsym, but some were missing in the man page.
    2. intermediate indents in the sway.5 markup cause an unintended space in the command arguments in the formatted document. Remove them.

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -356,8 +356,8 @@ runtime.
 		for_window <criteria> move container to output <output>
 
 *bindsym* [--whole-window] [--border] [--exclude-titlebar] [--release] [--locked] \
-		[--to-code] [--input-device=<device>] [--no-warn] [Group<1-4>+]<key combo> \
-		<command>
+[--to-code] [--input-device=<device>] [--no-warn] [Group<1-4>+]<key combo> \
+<command>
 	Binds _key combo_ to execute the sway command _command_ when pressed. You
 	may use XKB key names here (*xev*(1) is a good tool for discovering these).
 	With the flag _--release_, the command is executed when the key combo is
@@ -409,7 +409,7 @@ runtime.
 ```
 
 	*bindcode* [--whole-window] [--border] [--exclude-titlebar] [--release] \
-	[--locked] [--input-device=<device>] [--no-warn] [Group<1-4>+]<code> <command>
+[--locked] [--input-device=<device>] [--no-warn] [Group<1-4>+]<code> <command>
 	is also available for binding with key/button codes instead of key/button names.
 
 *bindswitch* [--locked] [--no-warn] [--reload] <switch>:<state> <command>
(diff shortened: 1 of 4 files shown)


sway/sway.5.scd
```diff
@@ -725,12 +725,14 @@ The default colors are:
...
-*unbindsym* [--whole-window] [--border] [--exclude-titlebar] [--release] [--input-device=<device>] <key combo>
+*unbindsym* [--whole-window] [--border] [--exclude-titlebar] [--release] [--locked] \
+		[--to-code] [--input-device=<device>] <key combo>
 	Removes the binding for _key combo_ that was previously bound with the
 	given flags.  If _input-device_ is given, only the binding for that
 	input device will be unbound.
 
-	*unbindcode* [--whole-window] [--border] [--exclude-titlebar] [--release] [input-device=<device>] <code>
+	*unbindcode* [--whole-window] [--border] [--exclude-titlebar] [--release] \
+	[--locked] [input-device=<device>] <code>
```
emersion: Style: missing indent here

RPigott: So the formatting in the man page is off here, but the style is the same as for bindcode description, following bindsym above.

Now that I take another look, and this may be a bug in scdoc, the current man page actually has a full tab-width gap between the [--release] and [--locked] args in bindcode - where the newline break and indent is in the markup. unbindcode used to be one line in the markup with fewer args so it doesn't currently have that formatting error.

You can get what I presume is the correct formatting with a _dedent_ following bindcode. It seems the tab is inserted literally there. Still looks kinda ugly in the markup though.

So now I've actually pushed the version with the corrected (?) formatting with the ugly dedent. Let me know what you think we should do here.

sway/sway.5.scd
```diff
@@ -725,12 +725,14 @@ The default colors are:
 *unbindswitch* <switch>:<state>
 	Removes a binding for when <switch> changes to <state>.
 
-*unbindsym* [--whole-window] [--border] [--exclude-titlebar] [--release] [--input-device=<device>] <key combo>
+*unbindsym* [--whole-window] [--border] [--exclude-titlebar] [--release] [--locked] \
+		[--to-code] [--input-device=<device>] <key combo>
```
emersion: We need strip leading whitespace here too, I believe.

RPigott: Yeah looks like it. The specific width of my terminal hid this bit of formatting from me in the man page.

PR #4916, opened 2020-01-17
output: apply oc to outputs using wl_for_each_safe

### LINKED ISSUE #4904: crash when re-enabling monitors (after dpms off)
* Sway Version: sway version 1.2-8b251f06 (Jan 11 2020, branch 'HEAD')

* Debug Log: From a couple of months ago, git master, same error

https://gist.github.com/2eab9ca9ca6e7ae7eaba7ec2c76abf86

* Configuration File: Base (very little changed)

* Stack Trace, if sway crashes: https://gist.github.com/91f7149735a436b3835cbbfdedddbba9

* Env info:

thunderbolt -> thunderbolt displays (dual dp adapter) (daisy chain), 2 4k monitors

after dpms off sometimes one of them will not return all modes (or something, lower res when it crashes back to shell at least), power cycling that display makes it re-report in.

integrated intel gpu i7-7600U (kaby lake)

RedSoxFan:
Fixes #4904 

If applying an output config to an output fails, the output may be
destroyed by wlroots. To be able to handle this situation correctly,
apply_output_config_to_outputs needs to use wl_list_for_each_safe.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -550,8 +550,8 @@ void apply_output_config_to_outputs(struct output_config *oc) {
 	// will be applied during normal "new output" event from wlroots.
 	bool wildcard = strcmp(oc->name, "*") == 0;
 	char id[128];
-	struct sway_output *sway_output;
-	wl_list_for_each(sway_output, &root->all_outputs, link) {
+	struct sway_output *sway_output, *tmp;
+	wl_list_for_each_safe(sway_output, tmp, &root->all_outputs, link) {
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

emersion:
>If applying an output config to an output fails, the output may be
destroyed by wlroots.

I think this is a wlroots bug. This is unexpected from an API user point-of-view.

This happens because wlroots fails to start rendering an output. Instead of destroying the `wlr_output` we should keep the output off.

PR #4923, opened 2020-01-18
commands/move: Fix crash when required args not provided

### LINKED ISSUE #4919: Sway crashes when issuing move [container|window] command without additional arguments
Steps to reproduce:

`swaymsg move window` or
`swaymsg move container`

[swaylog.TXT](https://github.com/swaywm/sway/files/4075320/swaylog.TXT)
[backtrace.TXT](https://github.com/swaywm/sway/files/4075321/backtrace.TXT)

Sway and Wlroots are both from master branch and up-to-date.

ammgws:
Fixes #4919.

Crash was occurring because ++argv was going out of bounds.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -918,11 +918,17 @@ struct cmd_results *cmd_move(int argc, char **argv) {
 
 	if (strcasecmp(argv[0], "window") == 0 ||
 			strcasecmp(argv[0], "container") == 0) {
-		--argc; ++argv;
+		--argc;
+		if (argc > 0) {
+			++argv;
+		}
 	}
 
 	if (strcasecmp(argv[0], "to") == 0) {
-		--argc; ++argv;
+		--argc;
+		if (argc > 0) {
+			++argv;
+		}
 	}
 
 	if (!argc) {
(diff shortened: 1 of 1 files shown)


sway/commands/move.c
```diff
@@ -918,11 +918,17 @@ struct cmd_results *cmd_move(int argc, char **argv) {
...
 	if (strcasecmp(argv[0], "window") == 0 ||
 			strcasecmp(argv[0], "container") == 0) {
-		--argc; ++argv;
+		--argc;
+		if (argc > 0) {
+			++argv;
+		}
 	}
 
 	if (strcasecmp(argv[0], "to") == 0) {
```
emersion: I think we could keep the previous logic, but add a `argc > 0` check in front of this `if`

sway/commands/move.c
```diff
@@ -918,11 +918,17 @@ struct cmd_results *cmd_move(int argc, char **argv) {
 
 	if (strcasecmp(argv[0], "window") == 0 ||
```
emersion: Technically we need to make sure `argc > 0` before calling `strcasecmp` here, to prevent commands ending with `--no-auto-back-and-forth` from segfaulting.

ammgws:
Sorry one last thing, I wanted to further test if there were also issues with `--no-auto-back-and-forth` but I keep getting `swaymsg: unrecognized option '--no-auto-back-and-forth'`.

e.g. `swaymsg move --no-auto-back-and-forth window to workspace 1`

PR #4930, opened 2020-01-19
Add ability to remove the floating modifier

ForTheReallys:


ddevault:
Needs updated man pages as well.

ddevault:
Thanks!

PR #4932, opened 2020-01-21
Implement input method keyboard grab

xdavidwu:
This implements input-method-unstable-v2 keyboard grab.

Depends on: ~~#4740~~, ~~swaywm/wlroots#1864~~

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -378,6 +378,21 @@ static void update_keyboard_state(struct sway_keyboard *keyboard,
 	}
 }
 
+/**
+ * Get keyboard grab of the seat from sway_keyboard.
+ *
+ * Returns NULL if the keyboard is not grabbed by an input method.
+ */
+static struct wlr_input_method_keyboard_grab_v2 *keyboard_get_im_grab(
+		struct sway_keyboard *keyboard) {
+	struct wlr_input_method_v2 *input_method = keyboard->seat_device->
+		sway_seat->im_relay.input_method;
+	if (!input_method) {
+		return NULL;
+	}
+	return input_method->keyboard_grab;
+}
+
 static void handle_key_event(struct sway_keyboard *keyboard,
 		struct wlr_event_keyboard_key *event) {
 	struct sway_seat *seat = keyboard->seat_device->sway_seat;
@@ -489,38 +504,35 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 	}
 
 	if (!handled || event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
-		if (seat->im_relay.input_method) {
-			struct wlr_input_method_keyboard_grab_v2 *kb_grab =
-				seat->im_relay.input_method->keyboard_grab;
-			struct wlr_virtual_keyboard_v1 *virtual_keyboard =
-				wlr_input_device_get_virtual_keyboard(wlr_device);
+		struct wlr_input_method_keyboard_grab_v2 *kb_grab = keyboard_get_im_grab(keyboard);
+		struct wlr_virtual_keyboard_v1 *virtual_keyboard =
+			wlr_input_device_get_virtual_keyboard(wlr_device);
 
-			// If event is from virtual keyboard of the same client as grab,
-			// do not send it back. TODO see swaywm/wlroots#2322
-			if (kb_grab && !(virtual_keyboard &&
-					wl_resource_get_client(virtual_keyboard->resource) ==
-					wl_resource_get_client(kb_grab->resource))) {
-				// Do not send release event to grab if the press event was not
-				// sent to grab.
-				if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
-					bool pressed_sent = update_shortcut_state(
-						&keyboard->state_pressed_sent, event->keycode,
-						event->state, keyinfo.keycode, 0);
-					if (pressed_sent) {
-						wlr_seat_set_keyboard(wlr_seat, wlr_device);
-						wlr_seat_keyboard_notify_key(wlr_seat, event->time_msec,
-							event->keycode, event->state);
-						goto end;
-					}
-
+		// If event is from virtual keyboard of the same client as grab,
+		// do not send it back. TODO see swaywm/wlroots#2322
+		if (kb_grab && !(virtual_keyboard &&
+				wl_resource_get_client(virtual_keyboard->resource) ==
+				wl_resource_get_client(kb_grab->resource))) {
+			// Do not send release event to grab if the press event was not
+			// sent to grab.
+			if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
+				bool pressed_sent = update_shortcut_state(
+					&keyboard->state_pressed_sent, event->keycode,
+					event->state, keyinfo.keycode, 0);
+				if (pressed_sent) {
+					wlr_seat_set_keyboard(wlr_seat, wlr_device);
+					wlr_seat_keyboard_notify_key(wlr_seat, event->time_msec,
+						event->keycode, event->state);
+					goto end;
 				}
 
-				wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
-					wlr_device->keyboard);
-				wlr_input_method_keyboard_grab_v2_send_key(kb_grab,
-					event->time_msec, event->keycode, event->state);
-				goto end;
 			}
+
+			wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
+				wlr_device->keyboard);
+			wlr_input_method_keyboard_grab_v2_send_key(kb_grab,
+				event->time_msec, event->keycode, event->state);
+			goto end;
 		}
 
 		bool pressed_sent = update_shortcut_state(
@@ -648,29 +660,23 @@ static void handle_modifier_event(struct sway_keyboard *keyboard) {
 	struct wlr_input_device *wlr_device =
 		keyboard->seat_device->input_device->wlr_device;
 	if (!wlr_device->keyboard->group) {
-		struct sway_seat *seat = keyboard->seat_device->sway_seat;
-		bool sent_to_kb_grab = false;
-		if (seat->im_relay.input_method) {
-			struct wlr_input_method_keyboard_grab_v2 *kb_grab =
-				seat->im_relay.input_method->keyboard_grab;
-			struct wlr_virtual_keyboard_v1 *virtual_keyboard =
-				wlr_input_device_get_virtual_keyboard(wlr_device);
+		struct wlr_input_method_keyboard_grab_v2 *kb_grab = keyboard_get_im_grab(keyboard);
+		struct wlr_virtual_keyboard_v1 *virtual_keyboard =
+			wlr_input_device_get_virtual_keyboard(wlr_device);
 
-			// If event is from virtual keyboard of the same client as grab,
-			// do not send it back. TODO see swaywm/wlroots#2322
-			if (kb_grab && !(virtual_keyboard &&
-					wl_resource_get_client(virtual_keyboard->resource) ==
-					wl_resource_get_client(kb_grab->resource))) {
-				wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
-						wlr_device->keyboard);
-				wlr_input_method_keyboard_grab_v2_send_modifiers(kb_grab,
-						&wlr_device->keyboard->modifiers);
-				sent_to_kb_grab = true;
-			}
-		}
-		if (!sent_to_kb_grab) {
-			wlr_seat_set_keyboard(seat->wlr_seat, wlr_device);
-			wlr_seat_keyboard_notify_modifiers(seat->wlr_seat,
+		// If event is from virtual keyboard of the same client as grab,
+		// do not send it back. TODO see swaywm/wlroots#2322
+		if (kb_grab && !(virtual_keyboard &&
+				wl_resource_get_client(virtual_keyboard->resource) ==
+				wl_resource_get_client(kb_grab->resource))) {
+			wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
+					wlr_device->keyboard);
+			wlr_input_method_keyboard_grab_v2_send_modifiers(kb_grab,
+					&wlr_device->keyboard->modifiers);
+		} else {
+			struct wlr_seat *wlr_seat = keyboard->seat_device->sway_seat->wlr_seat;
+			wlr_seat_set_keyboard(wlr_seat, wlr_device);
+			wlr_seat_keyboard_notify_modifiers(wlr_seat,
 					&wlr_device->keyboard->modifiers);
 		}
 
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -489,6 +489,40 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 	}
 
 	if (!handled || event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
+		if (seat->im_relay.input_method) {
+			struct wlr_input_method_keyboard_grab_v2 *kb_grab =
+				seat->im_relay.input_method->keyboard_grab;
+			struct wlr_virtual_keyboard_v1 *virtual_keyboard =
+				wlr_input_device_get_virtual_keyboard(wlr_device);
+
+			// If event is from virtual keyboard of the same client as grab,
+			// do not send it back. TODO see swaywm/wlroots#2322
+			if (kb_grab && !(virtual_keyboard &&
+					wl_resource_get_client(virtual_keyboard->resource) ==
+					wl_resource_get_client(kb_grab->resource))) {
+				// Do not send release event to grab if the press event was not
+				// sent to grab.
+				if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
+					bool pressed_sent = update_shortcut_state(
+						&keyboard->state_pressed_sent, event->keycode,
+						event->state, keyinfo.keycode, 0);
+					if (pressed_sent) {
+						wlr_seat_set_keyboard(wlr_seat, wlr_device);
+						wlr_seat_keyboard_notify_key(wlr_seat, event->time_msec,
+							event->keycode, event->state);
+						goto end;
+					}
+
+				}
+
+				wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
+					wlr_device->keyboard);
+				wlr_input_method_keyboard_grab_v2_send_key(kb_grab,
+					event->time_msec, event->keycode, event->state);
+				goto end;
+			}
+		}
+
 		bool pressed_sent = update_shortcut_state(
 				&keyboard->state_pressed_sent, event->keycode, event->state,
 				keyinfo.keycode, 0);
@@ -499,7 +533,7 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 		}
 	}
 
-
+end:
 	free(device_identifier);
 }
 
@@ -614,10 +648,31 @@ static void handle_modifier_event(struct sway_keyboard *keyboard) {
 	struct wlr_input_device *wlr_device =
 		keyboard->seat_device->input_device->wlr_device;
 	if (!wlr_device->keyboard->group) {
-		struct wlr_seat *wlr_seat = keyboard->seat_device->sway_seat->wlr_seat;
-		wlr_seat_set_keyboard(wlr_seat, wlr_device);
-		wlr_seat_keyboard_notify_modifiers(wlr_seat,
-				&wlr_device->keyboard->modifiers);
+		struct sway_seat *seat = keyboard->seat_device->sway_seat;
+		bool sent_to_kb_grab = false;
+		if (seat->im_relay.input_method) {
+			struct wlr_input_method_keyboard_grab_v2 *kb_grab =
+				seat->im_relay.input_method->keyboard_grab;
+			struct wlr_virtual_keyboard_v1 *virtual_keyboard =
+				wlr_input_device_get_virtual_keyboard(wlr_device);
+
+			// If event is from virtual keyboard of the same client as grab,
+			// do not send it back. TODO see swaywm/wlroots#2322
+			if (kb_grab && !(virtual_keyboard &&
+					wl_resource_get_client(virtual_keyboard->resource) ==
+					wl_resource_get_client(kb_grab->resource))) {
+				wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
+						wlr_device->keyboard);
+				wlr_input_method_keyboard_grab_v2_send_modifiers(kb_grab,
+						&wlr_device->keyboard->modifiers);
+				sent_to_kb_grab = true;
+			}
+		}
+		if (!sent_to_kb_grab) {
+			wlr_seat_set_keyboard(seat->wlr_seat, wlr_device);
+			wlr_seat_keyboard_notify_modifiers(seat->wlr_seat,
+					&wlr_device->keyboard->modifiers);
+		}
 
 		uint32_t modifiers = wlr_keyboard_get_modifiers(wlr_device->keyboard);
 		determine_bar_visibility(modifiers);
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -503,11 +503,10 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 				keyinfo.raw_keysyms_len);
 	}
 
-	bool pressed_sent = false;
 	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
 		// If the pressed event was sent to a client, also send the released
 		// event. In particular, don't send the released event to the IM grab.
-		pressed_sent = update_shortcut_state(
+		bool pressed_sent = update_shortcut_state(
 			&keyboard->state_pressed_sent, event->keycode,
 			event->state, keyinfo.keycode, 0);
 		if (pressed_sent) {
@@ -536,7 +535,9 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 		}
 	}
 
-	if (!handled && !pressed_sent) {
+	if (!handled && event->state != WL_KEYBOARD_KEY_STATE_RELEASED) {
+		// If a released event failed pressed sent test, and not in sent to
+		// keyboard grab, it is still not handled. Don't handle released here.
 		update_shortcut_state(
 			&keyboard->state_pressed_sent, event->keycode, event->state,
 			keyinfo.keycode, 0);
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -503,7 +503,22 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 				keyinfo.raw_keysyms_len);
 	}
 
-	if (!handled || event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
+	bool pressed_sent = false;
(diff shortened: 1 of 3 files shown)


xhebox:
There're two problems of handle_key patch in https://github.com/swaywm/sway/pull/4932/commits/7471b43fdcdd7161c746fba188a3581c79c07344, i've modified a patch/

1. Update_short_cuts or what else code probably messed around the state, so that i can not receive release key event from the grab interface.

2. Even if i move the check to the most front, bypass all sway code. There's something wrong with it:
```
-       if (!handled && wlr_device->keyboard->group) {
+       if (!handled && wlr_device->keyboard->group && !virtual_keyboard) {
                // Only handle device specific bindings for keyboards in a group
```
As virtual_keyboard and real keyboard compose a keyboard_group, sway early return here.

The workaround patch is attached at the end.

**EDIT**: There are problems in https://github.com/swaywm/wlroots/pull/1864 too.

```
@@ -329,6 +329,18 @@
        seat_idle_notify_activity(seat, IDLE_SOURCE_KEYBOARD);
        bool input_inhibited = seat->exclusive_client != NULL;
 
+       struct wlr_input_method_keyboard_grab_v2 *kb_grab =
+               seat->im_relay.input_method ? seat->im_relay.input_method->im_keyboard_grab : NULL;
+       struct wlr_virtual_keyboard_v1 *virtual_keyboard =
+               wlr_device->keyboard->virtual_keyboard;
+
+       if (seat->im_relay.input_method && kb_grab && !virtual_keyboard) {
+               wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
+                       wlr_device->keyboard);
+               wlr_input_method_keyboard_grab_v2_send_key(kb_grab,
+                       event->time_msec, event->keycode, event->state);
+       } else {
+
        // Identify new keycode, raw keysym(s), and translated keysym(s)
        xkb_keycode_t keycode = event->keycode + 8;
 
@@ -423,7 +435,7 @@
                handled = true;
        }
 
-       if (!handled && wlr_device->keyboard->group) {
+       if (!handled && wlr_device->keyboard->group && !virtual_keyboard) {
                // Only handle device specific bindings for keyboards in a group
                free(device_identifier);
                return;
@@ -451,6 +463,8 @@
                }
        }
 
+       }
+
        transaction_commit_dirty();
 
        free(device_identifier);
```

xdavidwu:
> Update_short_cuts or what else code probably messed around the state, so that i can not receive release key event from the grab interface.

Not all release events are missing, right? According to my tests, release events are sent successfully at most cases. Sometime release events seems still somewhat missing like I uses Mod4 as my binding modifiers, but when I switches workspace, sometime Mod4 seems like get stuck and needs an extra press to solve it after this patch. I am still investigating this situation.

> 2.

The problems with keyboard grouping is explained at https://github.com/swaywm/wlroots/pull/1864#issuecomment-586103504

xhebox:
Nope, all events are missing, with keyboard_group. I pressed the `f`, and seems it's marked as `pressed`, and I could not press `f` again.

And now with `seat * keyboard_grouping none`, it works almost perfectly. The problem must be related to the behavior of keyboard_group.

And the imperfect part is that, when I started IME(forward grab by virtual_keyboard) in a terminal, and press CTRL+C, I missed the release event of CTRL+C. So the corresponding release event was not forwarded by virtual_keyboard. The result is that, I am seeing lots of `ccccccccccccc` in the terminal untile i pressed `c` again.

This could be a responsibility of clients, but it will be nice if sway handle it.

xdavidwu:
> And the imperfect part is that, when I started IME(forward grab by virtual_keyboard) in a terminal, and press CTRL+C, I missed the release event of CTRL+C. So the corresponding release event was not forwarded by virtual_keyboard. The result is that, I am seeing lots of ccccccccccccc in the terminal untile i pressed c again.

Terminals kill the IME at the time it received press event, and when release event arrives, IME is already gone. Currently virtual-keyboard implementation does not send missing release event when it is gone. I think IME should handle SIGINT to be safe.

xhebox:
Yep, if virtual-keyboard handle it, then we need some records at the side of wlroots. Sounds like what you say `pressed event sent state tracking`. And with `pressed event sent state tracking`, grab could be moved to wlroots too.

Whatever it is not a big problem compared to missing events. Both solutions are ok though. It is just, program could be killed by various signals, it's hard to handle all cases. Exceptions could kill IME too.

Even IME handle this smoothly. If SIGKILL signal the program at similar time point, then program has no way to handle it at all.

xdavidwu:
I have just made virtual keyboard not to group when from same client as input method and also check and remove on grab start. It should make it work with keyboard grouping enabled but I think this can still be improved.

emersion:
Tested with [wlhangul](https://github.com/emersion/wlhangul), seems to work fine.

emersion:
This PR needs a rebase.

xdavidwu:
Commit to ungroup virtual keyboard is dropped. It is not needed after #5204.

emersion:
Is this still WIP?

sway/input/keyboard.c
```diff
@@ -485,6 +485,40 @@ static void handle_key_event(struct sway_keyboard *keyboard,
...
+				wlr_input_device_get_virtual_keyboard(wlr_device);
+
+			// If event is from virtual keyboard of the same client as grab,
+			// do not send it back. TODO see swaywm/wlroots#2322
+			if (kb_grab && !(virtual_keyboard &&
+					wl_resource_get_client(virtual_keyboard->resource) ==
+					wl_resource_get_client(kb_grab->resource))) {
+				// Do not send release event to grab if the press event was not
+				// sent to grab.
+				if (event->state == WLR_KEY_RELEASED) {
```
emersion, line 501: Hmm, I'm not quite sure I understand this part. Doesn't this mean we only forward RELEASED events to regular clients during a input-method grab?

kchibisov:
I've tested it a bit, and it seems that grab isn't applied correctly in certain case:

Start wlhangul. Open firefox and something like vte based terminal next to it, trigger input method in firefox and type something, focus vte terminal and try to type into it as well with enabled input method. As a result wlhangul won't get any key events that you'll be typing into that vte terminal, and will only get ime state updates. Firefox isn't required here, just the easiest one to trigger that.

One more issue is that I get infinite spam with  wl_keyboard::keymap for a new opened window

```
export WAYLAND_DEBUG=1
vte-2.91/sakura
```

You'll see something like

```
[3289908.847] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289914.839] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289917.601] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289924.767] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289927.864] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289932.609] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289939.246] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289941.921] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289948.479] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289953.429] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289958.070] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289962.844] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289967.558] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289970.735] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289976.048] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289982.531] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289988.132] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289990.925] wl_keyboard@27.keymap(1, fd 14, 51158)
[3289997.223] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290000.274] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290006.650] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290011.050] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290016.090] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290020.543] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290025.387] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290030.007] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290033.125] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290037.736] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290045.438] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290051.053] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290053.937] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290058.771] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290065.301] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290070.333] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290074.816] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290079.631] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290084.231] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290089.191] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290092.278] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290097.837] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290103.857] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290108.850] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290113.359] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290118.104] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290123.425] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290126.405] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290133.011] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290136.258] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290142.650] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290148.189] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290151.573] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290157.970] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290163.437] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290166.944] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290173.915] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290177.550] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290185.348] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290190.021] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290192.979] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290199.682] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290205.146] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290209.553] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290213.008] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290219.475] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290225.028] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290229.500] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290234.860] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290239.864] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290242.718] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290247.572] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290252.368] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290256.924] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290263.275] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290268.651] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290272.117] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290278.449] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290283.178] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290288.324] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290290.999] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290295.832] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290301.717] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290305.112] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290311.660] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290316.124] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290321.600] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290326.865] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290330.432] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290335.106] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290341.785] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290345.163] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290349.909] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290354.706] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290361.043] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290364.851] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290371.775] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290376.455] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290380.273] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290386.840] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290391.794] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290394.642] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290401.437] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290406.523] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290411.350] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290414.191] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290421.348] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290425.045] wl_keyboard@27.keymap(1, fd 14, 51158)
[3290429.590] wl_keyboard@27.keymap(1, fd 14, 51158)

```

which won't stop unless you press anything.

xhebox:
> Start wlhangul

You could also test it with wlchewing or my wlpinyin. I've started to use wlpinyin months ago. Sometimes it does has problems, but generally speaking, it works with this PR.

st3r4g:
> One more issue is that I get infinite spam with wl_keyboard::keymap for a new opened window ( you need wlhangul running)

This is pretty nasty, it seems this check isn't working, by reading the comment:
https://github.com/swaywm/wlroots/blob/master/types/wlr_input_method_v2.c#L240-L244i

It doesn't happen on wlchewing because it performs a similar check on its side, while wlhangul doesn't.

sway/input/keyboard.c
xdavidwu: It was to send RELEASED to regular client instead of keyboard grab if its corresponding PRESSED was sent to regular client (pressed before grab happened).

tadeokondrak:
> I've tested it a bit, and it seems that grab isn't applied correctly in certain case:
> Start wlhangul. Open firefox and something like vte based terminal next to it, trigger input method in firefox and type something, focus vte terminal and try to type into it as well with enabled input method. As a result wlhangul won't get any key events that you'll be typing into that vte terminal, and will only get ime state updates. Firefox isn't required here, just the easiest one to trigger that.

I think this issue and the one described [here](https://github.com/Riey/kime/pull/129#issuecomment-767000048) is solved by adding
```c
	if (text_input->input->focused_surface == NULL) {
		sway_log(SWAY_DEBUG, "Disabling text input, but no longer focused");
		return;
	}
```
in `relay_disable_text_input`.

tadeokondrak:
This PR doesn't work with layer-shell surfaces with keyboard interactivity. If your focus goes from a surface supporting text-input to a layer-shell surface with keyboard interactivity, input events are still routed to the input method and text-input events are still routed to the previous surface.

simnalamburt:
Despite the effort of @tadeokondrak, there are still two remaining bugs with this PR. Me and @guswns0528 are trying to solve this problem but as we're not very familiar with the codebase of sway and wlroots, we're having trouble. Any help would be welcomed!

&nbsp;

## Current bugs
First you must install sway 1.5.1, wlroots 0.12.0 and setup the Korean IME with input-method-unstable-v2 support. There bugs are reproducible with [kime], [fcitx5], or [wlhangul].

### 1. Key repeat bug
This bug is not related with key repeat feature of IME. For example, this bug is reproducible with IMEs which lacks the key repeat feature like old version of kime without or wlhangul. Sway's input-method-unstable-v2 implementation of wlroots is the suspected cause.

- Error reproduction steps 1: Hold <kbd>Alt</kbd> for 3 seconds → Press <kbd>1</kbd> → Immediately release <kbd>Alt</kbd> and <kbd>1</kbd>
  - Expected behavior: <kbd>Alt</kbd> + <kbd>1</kbd> is typed once
  - Actual behavior: <kbd>1</kbd> is typed infinitely
- Error reproduction steps 2: Hold <kbd>Ctrl</kbd> for 3 seconds → Press <kbd>Tab</kbd> → Immediately release <kbd>Ctrl</kbd> and <kbd>Tab</kbd>
  - Expected behavior: <kbd>Ctrl</kbd> + <kbd>Tab</kbd> is typed once
  - Actual behavior: <kbd>Tab</kbd> is typed infinitely
- Error reproduction steps 2: Hold <kbd>a</kbd> for 3 seconds → Press <kbd>b</kbd> → Immediately release <kbd>a</kbd> and <kbd>b</kbd>
  - Expected behavior: Repeat of <kbd>a</kbd> stops as soon as user release <kbd>a</kbd> and <kbd>b</kbd>
  - Actual behavior: <kbd>b</kbd> is typed infinitely

### 2. Focus change bug
I think it's caused by the timing of wlroots' deactivate, done event. But not sure though.

- Error reproduction steps:
  1. Create two alacritty instance. Make sure that they use the wayland backend.
  2. In A terminal, type "한[글]". ("글" means string which is still in preedit mode)
  3. Change the focus to the B terminal
- Expected behavior: "한글" in terminal A, nothing in B
- Actual behavior: "한" in A, "글" in B

## Reference
- https://github.com/korean-input/issues/issues/7

xhebox:
@simnalamburt I made a personal Chinese IME wlpinyin.  I could explain some backgrounds to you:

1. When IME grabs the input from sway/wlroots, it gains full control in literal. IME could handle the infinite typing case you mentioned, but they may not handle it correctly for whatever reason. This is what I get from wlpinyin in Chinese input mode:

```
song song song song sen fen fen fen fen fen fen fen fen fen fen fen fen fen fen fen fen fen fen fen f
// Press s first to produce 's' and 'ong'
// When I press f, it started to produce 'f' and 'en', and 's'/'ong' disapeared
// When release, it stoped
```
2.  Yes, there is such a problem. If you already leaves the surface, so you can not commit the preedit string in the previous surface. You can not explicitly manage the preedit/commit state of different surfaces.

xdavidwu:
> 2. Focus change bug

I think this is a bug in IME. Focus changes deactivates the input-method and I think there is no way to know it is going to be deactivated in advance. Which means IME does not have chance to commit string to old text-input. Also, by the protocol, there is no way to distinguish the text-input behind. With these, I think IME should reset their internal states (for example, to-be-committed string) on deactivate. This is what my wlchewing does, and so it does not have this bug.

> 1. Key repeat bug

Also sounds like IME bug to me. input-method-keyboard-grab, like in wayland core protocol for normal clients, sends pressed and released events and leave key repeat mechanism to clients. It looks like the IME you use does not handle released events correctly. In my experience of using my wlchewing daily, I did not encounter any problem of missing released event. You can check if there are released event by setting environment variable `WAYLAND_DEBUG=1`, which turns on printing of wayland messages in libwayland.

@tadeokondrak would you make a seperated PR for your patch for layer-shell support? It is not related to keyboard grab. Without keyboard grab, it also may help for usecases like automatically bringing up a on-screen virtual keyboard.

simnalamburt:
Thank you! I could workaround the "Focus change bug" and fix the "Key repeat bug". This PR is currently being thoroughly tested by bunch of Korean sway users and now we're all experiencing no more bugs!

Is there any left blocker before merging this PR? Looks like the keymap spamming issue has been fixed with latest commit since it doesn't occur anymore!

Riey:
I'm maintain new Korean IME kime and I met several repeat bugs of course it's IME bug but problem is this repeat was triggered in client side which is unexpected behavior for me

Why client get `RepeatInfo` even keyboard is grabbed? is it necessary?

tadeokondrak:
I think there still is a focus change bug in sway, but it's related to text-input and not input-method, and is not related to keyboard grab.

simnalamburt:
Yeah I think this patch itself is enough tested and reviewed by bunch of people who desperately want to use IME with sway. :)

fosskers:
I've seen the patched AUR package with this added. How is this actually configured and used? Please and thanks :)

simnalamburt:
Just install patched version of sway and turn on the IME which have input-method-unstable-v2 supports like [kime](https://github.com/Riey/kime), fcitx5. Then it automatically works.

For Arch Linux with kime:

```
# Install two AUR: sway-im and kime-bin
paru -S sway-im kime-bin

# Run kime
systemd-run --user -du kime kime-wayland

# Run sway
sway
```

###### Reference
- https://aur.archlinux.org/packages/sway-im/

fosskers:
I did manage to get `fcitx5` working without a patched `sway`. Although I still can't type in `alacritty` for some reason.

tadeokondrak:
> I did manage to get fcitx5 working without a patched sway. Although I still can't type in alacritty for some reason.

This is because some applications also support IME input without going through the window system, using (for example) D-Bus instead of X11 or Wayland. Alacritty only supports X11 (xim) and Wayland (text-input-v3) IMEs directly, as far as I know.

sway/input/keyboard.c
```diff
@@ -489,6 +489,40 @@ static void handle_key_event(struct sway_keyboard *keyboard,
...
+						goto end;
+					}
+
+				}
+
+				wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
+					wlr_device->keyboard);
+				wlr_input_method_keyboard_grab_v2_send_key(kb_grab,
+					event->time_msec, event->keycode, event->state);
+				goto end;
```
emersion, line 522: I think we could simplify these `goto`s a little Maybe something among the lines of:

```c
	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
		// If the press event was sent to a client, also send the release
		// event. In particular, don't send the release event to the IM grab.
		bool pressed_sent = update_shortcut_state(
			&keyboard->state_pressed_sent, event->keycode,
			event->state, keyinfo.keycode, 0);
		if (pressed_sent) {
			wlr_seat_set_keyboard(wlr_seat, wlr_device);
			wlr_seat_keyboard_notify_key(wlr_seat, event->time_msec,
				event->keycode, event->state);
			handled = true;
		}
	}

	if (!handled) {
		struct wlr_input_method_keyboard_grab_v2 *kb_grab =
			keyboard_get_im_grab(struct sway_keyboard *keyboard);
		if (kb_grab) {
			wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
				wlr_device->keyboard);
			wlr_input_method_keyboard_grab_v2_send_key(kb_grab,
				event->time_msec, event->keycode, event->state);
			handled = true;
		}
	}

	if (!handled) {
		update_shortcut_state(
			&keyboard->state_pressed_sent, event->keycode, event->state,
			keyinfo.keycode, 0);
		wlr_seat_set_keyboard(wlr_seat, wlr_device);
		wlr_seat_keyboard_notify_key(wlr_seat, event->time_msec,
			event->keycode, event->state);
	}
```

emersion (changes requested):
I've been trying to review this for quite some time. I'm really sorry for the huge delay, once again.

Overall the logic looks correct to me.

I have trouble with the changes in `handle_key_event`. This function is already a little bit tricky to follow, and the changes introduce more complexity (duplicated `wlr_seat_keyboard_notify_key`, `goto end`). I've left some comments to see if we can improve that a bit.

In other news, we're trying to revive the IM standardization efforts in the wayland-protocols repo. If you have feedback about text-input or input-method, please feel free to leave a comment:

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/39

sway/input/keyboard.c
xdavidwu: Applied the idea. Thanks!
The code above would send released event to client even if no pressed event was sent before. I've fixed it.

sway/input/keyboard.c
```diff
@@ -614,10 +648,31 @@ static void handle_modifier_event(struct sway_keyboard *keyboard) {
...
+			struct wlr_input_method_keyboard_grab_v2 *kb_grab =
+				seat->im_relay.input_method->keyboard_grab;
+			struct wlr_virtual_keyboard_v1 *virtual_keyboard =
+				wlr_input_device_get_virtual_keyboard(wlr_device);
+
+			// If event is from virtual keyboard of the same client as grab,
+			// do not send it back. TODO see swaywm/wlroots#2322
+			if (kb_grab && !(virtual_keyboard &&
+					wl_resource_get_client(virtual_keyboard->resource) ==
+					wl_resource_get_client(kb_grab->resource))) {
```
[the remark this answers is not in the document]
xdavidwu: Applied, but I am not sure about hiding the check for virtual keyboard behind that function, so I left the check outside.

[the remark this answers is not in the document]
emersion: Ah, I'd prefer to move it over to the function if possible. Or do you have a concern with that?

sway/input/keyboard.c
```diff
@@ -488,17 +503,47 @@ static void handle_key_event(struct sway_keyboard *keyboard,
...
+				wl_resource_get_client(kb_grab->resource))) {
+			wlr_input_method_keyboard_grab_v2_set_keyboard(kb_grab,
+				wlr_device->keyboard);
+			wlr_input_method_keyboard_grab_v2_send_key(kb_grab,
+				event->time_msec, event->keycode, event->state);
+			handled = true;
+		}
+	}
+
+	if (!handled && !pressed_sent) {
```
emersion, line 539: Why do we need to check for `!pressed_sent` here?

If the event was not handled, then `pressed_sent` will always be false, or am I missing something?

> The code above would send released event to client even if no pressed event was sent before. I've fixed it.

Hm. If a pressed event was sent to the client, the first block `if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED)` is responsible for sending the release event (and set `handled = true`).

If no pressed event was sent, then the first block will not handle the event, nor will the IM block. Maybe we need to restore the original condition in this block to fix it?

```c
if (pressed_sent || event->state == WL_KEYBOARD_KEY_STATE_PRESSED)
```

xdavidwu: I was trying to match that original check but I was wrong. I am sorry for that.

I just came up with a much simpler way to restore the original behavior. Not handling released events at that last block should do the trick. I have updated this PR for it.

emersion: Oh, yeah, sounds good to me and simpler than my proposal.

emersion (approved):
Looks good to me, thanks a lot!

sway/input/keyboard.c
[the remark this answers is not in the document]
xdavidwu: I was thinking if that was too much for a function named like that, but it should be fine with detailed comments added.

I've applied it.

Issue #4906, opened 2020-01-14
[Feature Request] option to move to mouse without going off screen


tmccombs:
# Use case

I have some buttons on my status bar (using waybar) which have click handlers that launch applications that have small floating windows. For example, launching gsimplecal when I click on the date/time module. Currently they launch in the middle of the screen. Setting a specific x,y coordinate doesn't work for me, because I have two monitors with different resolutions. and my statusbar is on the bottom. using `for_window [app_id="gsimplecal"] move position mouse` results in the window being partially off-screen. 

# Suggestion

Add an option to the move command like `move --constrain position mouse` which moves the targeted window to the mouse, but if centering would result in the window being partially off screen (including hidden by a layer-shell such as the statusbar) it is adjusted to be fully visible. 

# Alternative

My use case would also be solved by having a way to move to a relative position other than "center". for example `move position bottom right`.

# Workaround

I think I could accomplish this by using `for_window [<match>] exec position-script` where position-script finds the window, computes the proper location and moves it there, but that is substantially more complicated than the above solutions. 

If one of these solutions is acceptable, I'd be happy to take a stab at implementing it.

p.s. is there a difference between cursor, mouse, and pointer in the move command?

emersion:
Please open a feature request on the i3 bug tracker. If it gets accepted there, we'll consider implementing this feature.

tmccombs:
So, I actually just tried `move position mouse` in i3, and it has the behavior described in my first suggestion (that is it takes the edges of the screen into account). So sway actually differs from i3 in it's behavior. Would that make this a bug?

PR #4934, opened 2020-01-21
Keep windows in bounds on move to position mouse

tmccombs:
If the mouse/cursor/pointer is near the edge of an output when a "move
position to pointer" command is run, then the floating container will be
constrained to fit inside the bounds of the output as much as possible.

This behavior matches what i3 does in this scenario. I also think it is
a better user experience.

Relates to #4906

The logic for the bounds check follows the implementation in i3: https://github.com/i3/i3/blob/733077822302d8b77eacb606a26fd002a42f534f/src/floating.c#L536

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -760,19 +760,18 @@ static struct cmd_results *cmd_move_to_position_pointer(
 	double ly = cursor->y - container->height / 2;
 
 	/* Correct target coordinates to be in bounds (on screen). */
-	for (int i = 0; i < root->outputs->length; ++i) {
-		struct wlr_box box;
-		output_get_box(root->outputs->items[i], &box);
-		if (wlr_box_contains_point(&box, cursor->x, cursor->y)) {
-			lx = fmax(lx, box.x);
-			ly = fmax(ly, box.y);
-			if (lx + container->width > box.x + box.width) {
-				lx = box.x + box.width - container->width;
-			}
-			if (ly + container->height > box.y + box.height) {
-				ly = box.y + box.height - container->height;
-			}
-			break;
+	struct wlr_output *output = wlr_output_layout_output_at(
+			root->output_layout, cursor->x, cursor->y);
+	if (output) {
+		struct wlr_box *box =
+				wlr_output_layout_get_box(root->output_layout, output);
+		lx = fmax(lx, box->x);
+		ly = fmax(ly, box->y);
+		if (lx + container->width > box->x + box->width) {
+			lx = box->x + box->width - container->width;
+		}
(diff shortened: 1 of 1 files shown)


sway/commands/move.c
```diff
@@ -747,6 +748,38 @@ static struct cmd_results *cmd_move_in_direction(
...
+	if (!seat->cursor) {
+		return cmd_results_new(CMD_FAILURE, "No cursor device");
+	}
+	struct wlr_cursor *cursor = seat->cursor->cursor;
+	/* Determine where to put the window. */
+	double lx = cursor->x - container->width / 2;
+	double ly = cursor->y - container->height / 2;
+
+	/* Correct target coordinates to be in bounds (on screen). */
+	for (int i = 0; i < root->outputs->length; ++i) {
```
tmccombs: I was unable to find a better way to get the output from the cursor, if there is a better way, I'd be happy to change this. (Maybe add a function to wlroots to get the output from the cursor's state?)

ddevault:
There's a wlroots function which is probably better for this: `wlr_output_layout_closest_point`.

tmccombs:
I'm not sure how `wlr_output_layout_closest_point` would be helpful, perhaps you could exapnd on that? 

I did find `wlr_output_layout_output_at` which does seem useful.

ddevault:
Given a position which is out of bounds, it returns the nearest position which is in bounds.

tmccombs:
I'm sorry, I understand what `wlr_output_layout_closest_point` does. I'm just not sure how you are proposing I would use it in this context. 

Are you suggesting that rather than finding the output that the cursor is on, and constraining the container to the bounding box of that output, I should use `wlr_output_layout_closest_point` with `NULL` as the second argument to find the closest points inside _some_ output for the top-lelft and bottom-right corners, then use those to calculate the final position? I think that would result in slightly different behavior than i3 in the case of multiple outputs. If the cursor is near the boundary between two outputs the window will overlap between the two outputs. Personally, I don't think that is as nice of a user experience in most cases, but maybe that is just me?

Or are you saying I should find the output containing the cursor, then pass that as the second argument to `wlr_output_layout_closest_point` and do something like this:

```c
wlr_output_layout_closest_point(root->output_layout, output, lx, ly, &lx, &ly);
double lx2, ly2;
wlr_output_layout_closest_point(root->output_layout, output, lx + container->width, ly + container->height, &lx2, &ly2);
lx -= lx + container->width - lx2;
ly -= ly + container->height - ly2;
```

That doesn't seem a whole lot cleaner to me, but if you prefer it, I'd be fine with changing it.

tmccombs:
Is this waiting on anything from me? or just waiting on a review?

PR #4935, opened 2020-01-22
Implement split none command

RPigott:
While browsing the i3 issue tracker I saw issue i3/i3#3808 "split none" command with an accepted label. 

Not sure if we want it yet, but I figured this one was pretty easy to implement for sway so here it is. i3 doesn't have an implementation yet, so I can't really compare it to theirs, but I expect this is the intended behavior.

### DIFF
--- a/sway/commands/split.c
+++ b/sway/commands/split.c
@@ -36,6 +36,24 @@ static struct cmd_results *do_split(int layout) {
 	return cmd_results_new(CMD_SUCCESS, NULL);
 }
 
+static struct cmd_results *do_unsplit() {
+	struct sway_container *con = config->handler_context.container;
+	struct sway_workspace *ws = config->handler_context.workspace;
+
+	if (con && con->parent && con->parent->children->length == 1) {
+		container_flatten(con->parent);
+	} else {
+		return cmd_results_new(CMD_FAILURE, "Can only flatten a child container with no siblings");
+	}
+
+	if (root->fullscreen_global) {
+		arrange_root();
+	} else {
+		arrange_workspace(ws);
+	}
+	return cmd_results_new(CMD_SUCCESS, NULL);
+}
+
 struct cmd_results *cmd_split(int argc, char **argv) {
 	struct cmd_results *error = NULL;
 	if ((error = checkarg(argc, "split", EXPECTED_EQUAL_TO, 1))) {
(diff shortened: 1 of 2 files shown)


sway/commands/split.c
```diff
@@ -59,6 +59,11 @@ struct cmd_results *cmd_split(int argc, char **argv) {
 		} else {
 			return do_split(L_VERT);
 		}
+	} else if (strcasecmp(argv[0], "n") == 0|| strcasecmp(argv[0], "none") == 0) {
```
RedSoxFan: style: space between `0` and `||`

RPigott:
Just updated since this was stale and had a conflict. No change in the i3 situation AFAIK.

scippio:
pls... merge it.. :pray:

jhonrocha:
Hey! i3 currently have a workaround for this: https://www.youtube.com/watch?v=AWA8Pl57UBY&t=278s
However, I am not able to reproduce that behavior on sway, trying to move a single child against its parent is actually moving the parent towards the selected direction. Is there a way to reproduce that?

RPigott:
@jhonrocha That should be fixed in master. If not, please file a new bug for it. i3 has entered a feature freeze and I'm not sure of the status of this feature for i3.

emersion:
The i3 devs have given a provisional ACK for this API. I'd be fine with merging even if there's a risk we need to add back compat for i3 later. This is a pretty small feature.

RPigott:
I'm fine with merging this. I think its unlikely for i3 to merge an incompatible feature, and even if they do it's unlikely to be a disaster if we change ours to match.

emersion:
Alright, let's do this then, thanks!

RPigott:
@emersion Uh, this doesn't actually build. I guess the CI hadn't been run in ages and it was never updated for a047b5ee4.

sim590:
How is this supposed to work? I tried:

```
bindsym $mod+u split none
```

But it just didn't change anything. The `V` marker in front of the window title indicating that the window is in vertical split mode is still there. Isn't `split none` supposed to remove it so that the next window I spawn is under the same parent as the focused window?

RPigott:
> Isn't split none supposed to remove it so that the next window I spawn is under the same parent as the focused window?

That sounds right. It will only work on singleton containers though, so if you've spawned another window already it will do nothing. Maybe try running `swaymsg split none` in your terminal from that configuration and see if it reports an error?

sim590:
@RPigott : you're right. If I remove the windows from the container, it seems to work. That's what I was missing I think!

PR #4936, opened 2020-01-22
Proper installation of shell completions

eli-schwartz:
Primarily motivated by the fact that fish is currently being installed to the wrong place, but let's do pkg-config since it is available.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -54,6 +54,8 @@ libinput       = dependency('libinput', version: '>=1.6.0')
 systemd        = dependency('libsystemd', version: '>=239', required: false)
 elogind        = dependency('libelogind', version: '>=239', required: false)
 xcb            = dependency('xcb', required: get_option('xwayland'))
+bash_comp      = dependency('bash-completion', required: false)
+fish_comp      = dependency('fish', required: false)
 math           = cc.find_library('m')
 rt             = cc.find_library('rt')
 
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -260,7 +262,11 @@ if get_option('bash-completions')
 		'completions/bash/swaybar',
 		'completions/bash/swaymsg',
 	)
-	bash_install_dir = datadir + '/bash-completion/completions'
+	if bash_comp.found()
+		bash_install_dir = bash_comp.get_pkgconfig_variable('completionsdir')
+	else
+		bash_install_dir = datadir + '/bash-completion/completions'
```
ascent12: Use join_paths instead of +

meson.build
```diff
@@ -271,7 +277,11 @@ if get_option('fish-completions')
 		'completions/fish/swaymsg.fish',
 		'completions/fish/swaynag.fish',
 	)
-	fish_install_dir = datadir + '/fish/completions'
+	if fish_comp.found()
+		fish_install_dir = fish_comp.get_pkgconfig_variable('completionsdir')
+	else
+		fish_install_dir = datadir + '/fish/vendor_completions.d'
```
ascent12: Here too.

meson.build
eli-schwartz: I suppose I could do that, but the current implementation is just shifting over existing code into a conditional. I wasn't sure what the original intent of using string concat rather than join_paths() was. :)

eli-schwartz: I've gone ahead and fixed this as well as a few other instances in a preliminary patch.

PR #4947, opened 2020-01-24
Docs: Add max_render_time field to get_tree response

ammgws:
This was added to the get_tree output in bd9a53f1a3e7dba247aab0a4e4268724acc12c38.

If it's not intended to be documented then perhaps we could add a disclaimer in the docs as per [i3](https://i3wm.org/docs/ipc.html#_tree_reply):

> While the nodes might have more properties, please do not use any properties which are not documented here. They are not yet finalized and will probably change!

### DIFF
--- a/sway/sway-ipc.7.scd
+++ b/sway/sway-ipc.7.scd
@@ -355,6 +355,10 @@ node and will have the following properties:
 |- floating_nodes
 :  array
 :  The floating children nodes for the node
+|- max_render_time
+:  integer
+:  Number of milliseconds a surface has to render and commit new contents before
+   being sampled by the compositor for the next presentation. "off" means none.
 |- representation
 :  string
 :  (Only workspaces) A string representation of the layout of the workspace


emersion:
cc @YaLTeR 

I'm not sure we're ready to document this yet. Maybe we should make it a ns value (240Hz displays make ms not precise enough). Also need to check whether using the literal `"off"` string is fine (not a fan of making this field both an int and string).

YaLTeR:
I used ms because that's the resolution the wayland timers provide. If there are higher resolution timers then we can use ns, sure.

I used the literal "off" string with the intention of extending it with other strings like "adaptive" later.

progandy:
> Also need to check whether using the literal `"off"` string is fine (not a fan of making this field both an int and string).

Otherwise you'd need two fields, e.g. ```render_strategy``` / ```max_render_time``` or define some negative values as magic numbers. There is also precedent with the configuration command for ```max_render_time```.

ammgws:
Closing due to inactivity

PR #4949, opened 2020-01-25
criteria: match containers without view

### LINKED ISSUE #4929: Command criteria only match views
Some criteria match container attributes (e.g. `con_mark`, `con_id`, `floating`) and many commands operate on containers, but there's no way to apply commands to non-leaf containers using criteria.
```
$ swaymsg 'splitv'
$ xterm &
$ swaymsg 'focus parent; mark test; focus child'
$ swaymsg '[con_mark="test"] focus'
Error: No matching node.
```
In `i3` this will correctly focus the parent container.

From what I understand criteria are expanded using `criteria_get_views` which, as the name suggests, only returns views. `handler_context.container` is then set to `view->container`. I believe this can be solved by matching on containers instead of views. I can try to make a patch if that is of interest.

* Sway Version:
```
$ swaymsg -t get_version
sway version 1.2
```
but doesn't appear to have changed on current master.

* Debug Log:
I don't believe the relevant code paths do much logging. Please let me know otherwise.

* Configuration File:
Reproducible with empty/default configuration.

ANerd:
Closes #4929

Replaces `criteria_get_views` with `criteria_get_containers` which can
return containers without views when the criteria only contains
container properties.

### DIFF
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -73,8 +73,8 @@ struct criteria *criteria_parse(char *raw, char **error);
 list_t *criteria_for_view(struct sway_view *view, enum criteria_type types);
 
 /**
- * Compile a list of views matching the given criteria.
+ * Compile a list of containers matching the given criteria.
  */
-list_t *criteria_get_views(struct criteria *criteria);
+list_t *criteria_get_containers(struct criteria *criteria);
 
 #endif
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -144,6 +144,35 @@ static void find_urgent_iterator(struct sway_container *con, void *data) {
 	list_add(urgent_views, con->view);
 }
 
+static bool has_container_criteria(struct criteria *criteria) {
+	return criteria->con_mark || criteria->con_id;
+}
+
+static bool criteria_matches_container(struct criteria *criteria,
+		struct sway_container *container) {
+	if (criteria->con_mark) {
+		bool exists = false;
+		struct sway_container *con = container;
+		for (int i = 0; i < con->marks->length; ++i) {
+			if (regex_cmp(con->marks->items[i], criteria->con_mark->regex) == 0) {
+				exists = true;
+				break;
+			}
+		}
+		if (!exists) {
+			return false;
+		}
+	}
+
+	if (criteria->con_id) { // Internal ID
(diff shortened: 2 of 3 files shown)


ddevault:
What does i3 do about this?

ANerd:
i3 supports this, as can be tested by the following and similar examples:
```
$ i3-msg 'splitv'
$ xterm &
$ i3-msg 'focus parent; mark test; focus child'
$ i3-msg '[con_mark="test"] focus'
```

I made this patch because this feature was part of my i3 workflow, and it broke when I moved to sway.

include/sway/criteria.h
```diff
@@ -6,6 +6,8 @@
 #include "list.h"
 #include "tree/view.h"
 
+struct sway_container;
```
RedSoxFan: I don't think this is needed. `struct sway_container` is not referenced in the header and  `sway/criteria.c` already includes `sway/tree/container.h`.

sway/criteria.c
```diff
@@ -144,8 +158,72 @@ static void find_urgent_iterator(struct sway_container *con, void *data) {
 	list_add(urgent_views, con->view);
 }
 
+static bool criteria_matches_container(struct criteria *criteria,
```
RedSoxFan: It looks like i3 only matches containers by `con_id` and `con_mark`. The rest of the options will only match windows/views.

https://github.com/i3/i3/blob/fc01ece3657e334d9067c40c0e7274ee955345e5/src/commands.c#L190

ANerd: It also looks like i3 ignores any other terms when the container has no window. I verified this by doing `i3-msg '[con_mark="test" class="do_not_match"] focus` and it still focused the container. I find this inconsistent, but I'm happy to do the same for compatibility.

sway/criteria.c
```diff
@@ -144,6 +144,35 @@ static void find_urgent_iterator(struct sway_container *con, void *data) {
 	list_add(urgent_views, con->view);
 }
 
+bool has_container_criteria(struct criteria *criteria) {
```
RedSoxFan: Since this isn't used outside of the file and is not included in the header, can you please make this `static`?

RedSoxFan:
Thanks!

PR #4952, opened 2020-01-25
Pointer gestures: Add three- and four-finger swipes to sway and swaybar

michaelweiser:
Having been a Mac user for some years I can't quite let go of touchpad gestures yet, even in sway. My use-case is switching workspaces using horizontal three-finger swipes and maybe triggering screen lock with an upwards swipe.

I've tried my hand at hacking this into sway, keeping it as generic as possible in the process. This ended up as pseudo-keysyms analogous to scrolling to which actions can be bound using bindsym as usual:
```
bindsym --whole-window --release SWIPE_3_LEFT workspace prev_on_output
bindsym --whole-window --release SWIPE_3_RIGHT workspace next_on_output
```

Would this be of interest to anyone but me? If so, what would need improving to get this included? Thanks!

BTW: Further usages I can think of would be moving windows about with an ongoing swipe or resizing using pinches. But this would be considerably harder to implement and IMO of seriously qestionable benefit.


michaelweiser:
Minor comment style change and commit reordering. No other changes. Using this daily without problems. Is there anything that needs improving to get this functionality included?

coderkalyan:
@michaelweiser Although I'm not a maintainer, I just took a look at this PR. The only things I would change refer to the configs - personally, I would think a new binding command such as `bindgesture` would be cleaner than adding a special case to bindsym. Additionally, instead of something like SWIPE_3_UP, I would use <gesture>:<direction>:<fingercount> (`swipe:up:3`). This would also give more flexibility for adding other gestures later, as well as directions. Also, I think it would be a good idea to implement up to 5 fingers for fingercount (by the looks of the code only 3 and 4 fingers is implemented) and additional directions (I think libinput supports left_up, etc; you can name them something like `upleft`).

Would love to have this feature in sway!

michaelweiser:
Updated for current master.

@coderkalyan: I appreciate all your comments. I intentionally tried to stay very close to what and how it's implemented currently because changes are routinely rejected if they stray too far from what i3 is doing. So I'd wait for a maintainer to get around to looking at this to see what changes are requested/suggested. Your ideas will likely be reflected in that now they're in this PR.

Regarding finger count: Adding five fingers would be trivial programmatically but I quickly tried to do a swipe with all my fingers and thumb on the trackpad and it seems quite unnatural to me. Is anyone else doing that? Also, I tried a four+one swipe with `libinput debug-events` and libinput still reports it with finger count four. I know it has a thumb detection, which is zone-based (contact in the lower third of the surface counts as thumb I think) but I don't know if it's applied to swipes as well. I tried to trick it and almost broke my fingers in doing so to no avail. It may well be that libinput is currently limited to four finger swipes. :/

marcusatiliusregulus:
5-finger gestures for swiping might not be so useful but i can imagine it for pinching gestures especially on very larg touchpads and on tablets. for example think about a drawing app letting you zoom in and out of the sheet with 5 fingers and pinching with less fingers changes stroke width or something else...

chron-isch:
5-finger swipes might not be very practical, but other 5-finger gestures are, I'd love to use a  5-finger rotation to trigger something like `output eDP-1 transform 90` for example on my tablet.

michaelweiser:
Pinches are handled as separate gestures by libinput and sway which could have their own (potentially very similar) PR. As stated in the description, I find this of questionable benefit in the context of a tiling window manager and likely quite hard to implement which is why I've left them out for now from this one.

coderkalyan:
@michaelweiser I agree with @chron-isch, I feel it's a better idea to add 5 finger gesture (its trivial to implement) rather than debate whether or not it's currently being used. If theres one thing I've learned from wayland, its that use cases will always surprise you :smile:

FlorianFranzen:
@michaelweiser Any update on this? It seems you are keeping this branch up to date, so I assume this works well for you.

Would love to see this merged.

michaelweiser:
@FlorianFranzen: Well yes, I use it daily and it works well. As its footprint in the code is quite low and not very conflict-prone I was otherwise content to let it sit here waiting for one of the maintainers to get around to it for review.

@emersion: As you stuck your neck out adding a label way back when: Could I interest you in this? :) I'm happy to work on it to get it ready for inclusion.

RPigott (commented):
I haven't looked at the swaybar parts yet, so just a few comments on the sway parts.

sway/input/seatop_default.c
```diff
@@ -747,6 +750,208 @@ static void handle_pointer_axis(struct sway_seat *seat,
 	}
 }
 
+static uint32_t wl_swipe_button_base(uint8_t fingers) {
```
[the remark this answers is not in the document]
michaelweiser: Done. I also removed it from a bunch of already present functions in sway and swaybar while at it and put that into separate commits for consideration.

sway/input/cursor.c
```diff
@@ -1194,6 +1191,34 @@ uint32_t get_mouse_bindsym(const char *name, char **error) {
 			SWAY_SCROLL_UP, SWAY_SCROLL_DOWN, SWAY_SCROLL_LEFT,
 			SWAY_SCROLL_RIGHT, BTN_SIDE, BTN_EXTRA};
 		return buttons[number - 1];
+	} else if (strncmp(name, "SWIPE_", strlen("SWIPE_")) == 0) {
+		static const struct {
+			uint32_t button;
+			const char *dir; } swipes[] = {
```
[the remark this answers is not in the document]
michaelweiser: Done.

sway/input/seatop_default.c
```diff
@@ -747,6 +750,208 @@ static void handle_pointer_axis(struct sway_seat *seat,
...
+		button_end = button_base + SWAY_SWIPE_DIR_COUNT;
+	}
+
+	/**
+	 * Handle mouse bindings - if there's any swipe bindings for this
+	 * context. Here we must swallow all swipes if there's even one
+	 * possible binding because we have no way of knowing yet what kind of
+	 * swipe this is going to be. It's in the nature of the thing because
+	 * the user may turn a left into an upwards swipe mid-way.
+	 */
```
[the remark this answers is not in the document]
michaelweiser: At least Safari and Firefox seem to do two- and three-finger gestures ever since 2009: http://www.macyourself.com/2009/09/18/mac-browser-showdown-multi-touch-gestures-in-safari-vs-firefox/. I think to have played around with that in Firefox at one point. The mail app (which I've never used) seems to have (at least at one point had) a swipe to archive function: https://osxdaily.com/2015/12/02/change-mail-swipe-gesture-mac-os-x/.

sway/input/seatop_default.c
```diff
@@ -747,6 +750,208 @@ static void handle_pointer_axis(struct sway_seat *seat,
...
+	struct sway_binding *binding = NULL;
+	for (uint32_t button = button_base; button < button_end; button++) {
+		state_add_button(e, button);
+		binding = get_active_mouse_binding(e,
+				config->current_mode->mouse_bindings,
+				modifiers, false, on_titlebar, on_border,
+				on_contents, on_workspace, dev_id);
+		// here we execute all press commands at swipe start because we
+		// just don't know what it's going to end up as - admittedly
+		// limits usefulness
```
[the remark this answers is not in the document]
michaelweiser: I couldn't think of a sensible use-case either. That's why I sprinkled those comments around so someone would tell me. ;) 
I've added new commits that switch behaviour to preferably executing a defined on-press binding on release and falling back on a --release binding if there is none. man pages are updated as well. This seems straightforward and well-fitting to me.
What would you envision that new binding type to look like in the config and code?

michaelweiser:
Out of reflex I rebased to current master. But I left the old commits unchanged, so no need to review them again. New commits implement the requested changes. I applied your suggestions to swaybar as well.

sway/input/seatop_default.c
```diff
@@ -642,7 +645,7 @@ static void handle_tablet_tool_motion(struct sway_seat *seat,
  * Functions used by handle_pointer_axis  /
  *--------------------------------------*/
 
-static uint32_t wl_axis_to_button(struct wlr_event_pointer_axis *event) {
+static uint32_t axis_to_button(struct wlr_event_pointer_axis *event) {
```
RPigott, line 648: Ehh, I wouldn't worry about the functions in swaybar like `wl_axis_to_button`, because "wl_axis" is an actual enum from the wayland protocol, so a function that translates a "wl_axis" value to something is appropriately named "wl_axis_to_something", even though it's not a symbol from libwayland. Similar for the wl_pointer object below.

On the contrary a "swipe_button" isn't anything related to the wayland protocol so it's just not appropriate here. I think just drop this commit; it's not necessary.

sway/input/seatop_default.c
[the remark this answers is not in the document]
RPigott: So if you use a swipe over the mail app it will archive something instead of switch workspaces? That's counter to how sway normally does bindings, but if we wanted we could coerce wlroots into telling us which apps care about swipes and give them first chance. It might be fine if only a minority of clients really care about swipes. Though if some apps got priority on swipes it might be better to just have a command the user can use with `for_window` to choose which ones...

I guess the current behavior is fine, I'm just not used to swipes.

sway/input/seatop_default.c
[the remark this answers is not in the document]
RPigott: Well the new semantics are good enough for me. Let's just leave it for now.

sway/input/seatop_default.c
michaelweiser: That explains it. Dropped.

sway/input/seatop_default.c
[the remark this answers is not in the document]
michaelweiser: I'd love to have some gestures in Firefox for navigation but was not able to get it on Linux working yet. This discussion has sparked my interest again. On the other hand I do not want to drag this out. So I'd suggest merging as-is and perhaps follow up with a separate PR regarding pass-through to swipe-enabled clients. Right now this would leave the user with a binary decision for each type of swipe: Either have a compositor-level binding or pass-through to clients. What d'you think?

michaelweiser:
I've squashed the review changes into the original commits. From my side this is ready for merge.

swaybar/input.c
```diff
@@ -463,6 +464,115 @@ static const struct wl_touch_listener touch_listener = {
...
+		int32_t cancelled) {
+	struct swaybar_seat *seat = data;
+	struct swaybar_swipe *swipe = &seat->swipe;
+
+	if (!swipe->swiping) {
+		sway_log(SWAY_DEBUG, "Ignoring swipe end without begin");
+		return;
+	}
+
+	swipe->swiping = false;
```
RPigott, line 551: This isn't sufficient to track a swipe beginning to end.

Basically, it should not be possible to leave swaybar in an indefinite "swiping" state. If a user begins a swipe on the bar and moves the cursor away, the swiping state remains. Then if the user begins a new swipe and moves the cursor back to the bar before ending it, swaybar will execute a binding, but the begin and end events it saw actually belong to two different swipe events.

It was possible to trick clients this way before, but this PR also introduces another method by putting the swiping state in the default seatop (instead of it's own seatop), because now entering any non-default seatop will trash the swiping state in sway.

Either sway needs to be consistent and cancel the swipe in the compositor when we can no longer deliver the real end event, or clients should be consistent in understanding that not all swipes have an end.

The protocol isn't really clear which answer is correct and I don't know of any clients that actually use it anyway, so I guess we do whatever seems reasonable until told otherwise. I think the easiest solution here would be to cancel the ongoing swipe in swaybar in the pointer leave event. As a result, a swipe must begin and end on the bar surface to count.

RPigott (changes requested):
Okay, had a look at the swaybar half this time. Looking good I think, other than review comments.

FlorianFranzen:
@michaelweiser Any update on this? I would not mind taking this over, but you have brought it this far and it seems we are getting close to having this merged.

michaelweiser:
@FlorianFranzen Yes, sorry for letting this sit and thanks for poking me. I had some trouble reproducing the indefinite swiping state at first and have been procrastinating since.

swaybar/input.c
michaelweiser: @RPigott I added the cancel to the pointer leave event as you suggested and can confirm that this avoids the "dangling" swipe state in swaybar. This is certainly cleaner than leaving the flag set indefinitely.

I was not able to reproduce the triggering of a binding by returning to the bar amidst a new swipe, though. Is it possible this is caused by the bar having its own layer?

RPigott: Hmm, I'm not sure why it wouldn't work for you but I'm pretty sure I didn't have trouble reproducing. It was a while ago and I didn't write down any other notes of my observations though. Make sure that sway itself isn't capturing the off-bar swipe since it will need to pass through the completion of that swipe to the bar at the end.

It shouldn't matter now since the new commit LGTM. Without the dangling swiping state I'm confident it's fixed. I'll try it out shortly.

RPigott (approved):

RPigott:
This all LGTM now, but I will wait for another review before merging.

swaybar/input.c
michaelweiser: > Make sure that sway itself isn't capturing the off-bar swipe

Right, that was the reason. Once I disabled the swipe bindings in sway proper, I was able to reproduce the issue. Thanks!

michaelweiser:
Squashed the bar review changes into the base commit. No other changes.

ErikReider:
Would the waybar project need to implement something similar to what swaybar? I've only tested this for a few minutes but I found that it didn't work when swiping over waybar

michaelweiser:
I remember to have had the same problem after first implementing swipe support in sway only. That's why I set out to add swipe support to swaybar in the first place. I have a dim recollection to have researched and understood at the time what the reason was for sway's swipe handling not applying to swaybar as well. I think it had to do with the fact of bars being special in some way but can not recall the details nor find them in the code. @RPigott Can you tell off the top of your head? Otherwise I'll dig some more.

RPigott:
Yes, waybar would need to implement swipe gestures.  It's unrelated to this PR, so it could add swipe gestures anytime.

swaybar and waybar are Wayland clients, and sway already delivers swipe events to clients that indicate they care about them. If clients want to react to swipe events, they must use the gestures client protocol. This PR implements:

1. A mechanism for sway to prevent delivering swipe events, so that sway can handle bindings on it's own without disturbing clients.
2. Implementation of gestures protocol for swaybar.

FlorianFranzen:
@michaelweiser It seems it currently fails to build because `wlr_box` has been moved to `wlr/util/box.h`.

michaelweiser:
@FlorianFranzen Yes. Unrelated to this PR though. Rebase resolved the issue.

emersion:
Thanks for working on this @michaelweiser, and thanks for the great review @RPigott! The nicely split changes and the well-written commit messages really makes the review a breeze.

The first 3 commits all look good to me.

Sorry if this has come up before, but while reviewing "input: Extend swipe gesture support to finger count" I've been wondering whether putting the number of fingers in the "key" value is the right way to go about this. Maybe we should consider putting the number of fingers in a new field in `struct sway_binding`, just like modifiers are a separate field?

From a command syntax point of view, would it be better to have a separate flag to describe the number of fingers? e.g. `bindsym --fingers=3 SWIPE_UP`? If we ever wanted to extend Sway to also handle other gestures like pinch and (upcoming) hold gestures, how would the syntax look like? Pinch gestures have a scale and rotation.

sway/sway.5.scd
```diff
@@ -440,6 +440,17 @@ runtime.
 	is over an empty workspace. Using a mouse binding over a layer surface's
 	exclusive region is not currently possible.
 
+	Special events SWIPE\_{3,4}\_{UP,DOWN,LEFT,RIGHT} respond to three- and
```
emersion, line 443: Nit: add `_` around `SWIPE\_{3,4}\_{UP,DOWN,LEFT,RIGHT}`

emersion: We should probably also make it clearer that the number of fingers is optional here?

michaelweiser: @emersion Thanks for the review! I added the underscores to the man page. Currently, the finger count isn't optional though since it's part of the special "key" identifier. It could be if we went the route of having a `--fingers` option to `bindsym`. I'll look into that and moving the finger count to struct `sway_binding` to see what it feels like. I won't be able to get to that before the week after next though. Bear with me. :)

emersion: > Currently, the finger count isn't optional though since it's part of the special "key" identifier

Ah, I see. I guess I just assumed 2-finger swipes were going to be available as well, but these are interpreted as scrolling instead.

> I won't be able to get to that before the week after next though. Bear with me. :)

No problem, we aren't in a hurry. I'm also guilty of dragging this for too long, I'll try to be more responsive in the future. :)

sevensidedmarble:
Is this going to include smooth transitions between workspaces like on MacOS? That would be a killer feature for Sway.

emersion:
Nope, no animations, no eye candy.

emersion:
@michaelweiserm, if you have time to update this PR, it would be nice to ship it in Sway 1.7. If not, no worries :)

FlorianFranzen:
> @michaelweiser Although I'm not a maintainer, I just took a look at this PR. The only things I would change refer to the configs - personally, I would think a new binding command such as `bindgesture` would be cleaner than adding a special case to bindsym. Additionally, instead of something like SWIPE_3_UP, I would use `<gesture>:<direction>:<fingercount>` (`swipe:up:3`). 

I implemented something very similar to this in #6778. Would love to get some feedback.

The0x539:
> No animations, no eye candy.

I disagree that all animations are eye candy. In particular, an animation for a swipe when switching desktops provides meaningful feedback before committing to the switch.

michaelweiser:
@FlorianFranzen : I've tried #6778. It works like a charm and I like the new command and syntax. Thanks for taking over and sorry for dropping the ball here. I believe this PR can be closed in favour of #6778. Let me know if there's anything I can do to help out - although I recently realised when trying to rebase some of my local sway customisations to 1.7 that I've gotten quite out of practice.

andraxin:
@FlorianFranzen
I've been using the original gesture patches for what it feels like time immemorial.

When switching to Sway 1.8, I noticed the "official" gesture support with a dedicated
keyword bindgesture instead of bindsym. Since this causes loss of functionality
compared to the original patch (modifiers are no longer supported), I'm wondering

*    if this change was necessary and why
     -   personally, I don't particularly care what it's called, but I have several actions
         configured with modifiers that no longer work
 *   if adding modifier support to bindgesture would be difficult

BTW, don't get me wrong, I'm grateful for the work you (and @michaelweiser) put in
to get this crucial functionality accepted; I'm just hoping to get my modifier+gesture
actions back soonish...  :smirk:

andraxin:
@FlorianFranzen 
Soon a year will have passed; so far, without any answers.
Is my question regarding the difficulty of adding support for
modifiers to `bindgesture` that cumbersome (or trivial)?

I'd appreciate any input, including potential guidance on how
one might go about reinstating that lost functionality.

PR #4958, opened 2020-01-27
Add support for wlr-output-power-management-unstable-v1

emersion:
See https://github.com/swaywm/wlroots/pull/1875

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -977,3 +977,21 @@ void handle_output_manager_test(struct wl_listener *listener, void *data) {
 	wlr_output_configuration_v1_send_succeeded(config);
 	wlr_output_configuration_v1_destroy(config);
 }
+
+void handle_output_power_manager_set_mode(struct wl_listener *listener,
+		void *data) {
+	struct wlr_output_power_v1_set_mode_event *event = data;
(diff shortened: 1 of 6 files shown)


sway/desktop/output.c
```diff
@@ -981,3 +981,20 @@ void handle_output_manager_test(struct wl_listener *listener, void *data) {
...
+	struct output_config *oc = new_output_config(output->wlr_output->name);
+	switch (event->mode) {
+	case ZWLR_OUTPUT_POWER_V1_MODE_OFF:
+		oc->dpms_state = DPMS_OFF;
+		break;
+	case ZWLR_OUTPUT_POWER_V1_MODE_ON:
+		oc->dpms_state = DPMS_ON;
+		break;
+	}
+	apply_output_config(oc, output);
```
RedSoxFan: This should be:
```c
oc = store_output_config(oc);
apply_output_config(oc, output);
```
Without the call to store_output_config, any changes to the output config will override this. Additionally, the config defaults would be used. That could alter aspects of the output such as scale and position, which could result in the output layout changing with possible gaps that the pointer cannot cross.

emersion: Right! Done.

ddevault:
Thanks!

PR #4959, opened 2020-01-27
Add 'output' mode to grimshot

cyrinux:
Hi,

This adds the possibility to capture only the focused screen(s) if several are used

Regards,

### DIFF
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -23,7 +23,7 @@ SUBJECT=${2:-screen}
 FILE=${3:-$(xdg-user-dir PICTURES)/$(date +'Grimshot %Y-%m-%d %H-%M-%S.png')}
 if [ "$ACTION" = "usage" ] ; then
   echo "Usage:"
-  echo "  grimshot copy|save win|screen|area [FILE]"
+  echo "  grimshot copy|save win|screen|focused_screen|area [FILE]"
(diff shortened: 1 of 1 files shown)


contrib/grimshot
```diff
@@ -23,7 +23,7 @@ SUBJECT=${2:-screen}
 FILE=${3:-$(xdg-user-dir PICTURES)/$(date +'Grimshot %Y-%m-%d %H-%M-%S.png')}
 if [ "$ACTION" = "usage" ] ; then
   echo "Usage:"
-  echo "  grimshot copy|save win|screen|area [FILE]"
```
emersion: Can we name this `output` instead?

ohk2kt3t4:
Would this still capture more than one screen in a multiseat sway session?

cyrinux:
> Would this still capture more than one screen in a multiseat sway session?

I don't know sorry ...

contrib/grimshot
cyrinux: Yeah good idea, i changed it.

ohk2kt3t4:
> I don't know sorry ...

Was just asking because if that was the case the description might be better as
 "This adds the possibility to capture only the focused screen(s) if several are used"

emersion:
Thanks!

PR #4962, opened 2020-01-30
Resolve bindswitch events causing crash when listener is registered

nathanrossi:
The `ipc_event_binding` function does not handle encoding of bindswitch events, and there is currently no documented format for these events in the ipc documentation.

To avoid sway crashing when any binding event listeners are registered and a bindswitch event occurs, simply ignore the event. I have implemented it this way as it resolves the crash. However it might be useful to implement support for bindswitch events?


nathanrossi:
I have replaced the patch with one that replaces the if as suggested.

PR #4964, opened 2020-01-31
Fix typos

torstehu:


include/sway/criteria.h
```diff
@@ -66,7 +66,7 @@ void criteria_destroy(struct criteria *criteria);
 struct criteria *criteria_parse(char *raw, char **error);
 
 /**
- * Compile a list of criterias matching the given view.
+ * Compile a list of criterion matching the given view.
```
emersion: I think this should be "criteria" actually, since it's a list of multiple criterion items. We say e.g. "list of strings" (plural form).

sway/criteria.c
```diff
@@ -361,10 +361,10 @@ static bool criteria_matches_view(struct criteria *criteria,
 }
 
 list_t *criteria_for_view(struct sway_view *view, enum criteria_type types) {
-	list_t *criterias = config->criteria;
+	list_t *criterion = config->criteria;
```
emersion: Ditto, should be "criteria"

torstehu:
I have update the commit after the review. Thanks for the feedback!

emersion:
Actually, "criterias" is incorrect. "criteria" should be used instead as it's already plural.

ddevault:
Thanks!

PR #4979, opened 2020-02-05
input: Map virtual-pointer to the requested output

any1:
Depends on https://github.com/swaywm/wlroots/pull/2019


artizirk:
For me `wlr_cursor_map_input_to_output` fails with `Cannot map device "virtual pointer" to output(not found in this cursor)`

emersion:
Thanks!

PR #4985, opened 2020-02-07
fix i3bar relative coordinates when scaling is used

hvenev:
The first commit fixes the mistakes I made in 24e8ba048aef4751c6fa1d5982ee634f921e6cf6. The second commit is the one that actually increases the precision of the reported values. Pixel coordinates are sent, not logical coordinates. Do you think this makes sense, or should logical coordinates be sent as floating-point values?

I haven't tested multi-monitor setups.

### DIFF
--- a/swaybar/input.c
+++ b/swaybar/input.c
@@ -138,21 +138,23 @@ static bool check_bindings(struct swaybar *bar, uint32_t button,
 	return false;
 }
 
-static void process_hotspots(struct swaybar_output *output,
+static bool process_hotspots(struct swaybar_output *output,
 		double x, double y, uint32_t button) {
-	x *= output->scale;
-	y *= output->scale;
+	double px = x * output->scale;
+	double py = y * output->scale;
 	struct swaybar_hotspot *hotspot;
 	wl_list_for_each(hotspot, &output->hotspots, link) {
-		if (x >= hotspot->x && y >= hotspot->y
-				&& x < hotspot->x + hotspot->width
-				&& y < hotspot->y + hotspot->height) {
-			if (HOTSPOT_IGNORE == hotspot->callback(output, hotspot,
-					x / output->scale, y / output->scale, button, hotspot->data)) {
-				return;
+		if (px >= hotspot->x && py >= hotspot->y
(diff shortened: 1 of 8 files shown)


emersion:
>Pixel coordinates are sent, not logical coordinates. Do you think this makes sense, or should logical coordinates be sent as floating-point values?

I think logical coords would make more sense. Current bar programs don't support the "scale" parameter.

Do current clients support floating point coords?

hvenev:
> I think logical coords would make more sense. Current bar programs don't support the "scale" parameter.

You are right, logical coordinates are better.

> Do current bar programs support floating point coords?

Most of the ones written in python probably will, but I have no idea what JSON implementations in statically typed programming languages do.

We could have the bar program announce that it supports floating-point coordinates, like we do with click events.

swaybar/status_line.c
```diff
@@ -85,6 +85,13 @@ bool status_handle_readable(struct status_line *status) {
 				}
 			}
 
+			json_object *float_coords;
+			if (json_object_object_get_ex(header, "float_coords", &float_coords)
```
emersion: Maybe name this `float_event_coords`, to make it clear it's about events.

swaybar/input.c
```diff
@@ -229,19 +231,8 @@ static void wl_pointer_axis(void *data, struct wl_pointer *wl_pointer,
...
-		if (x >= hotspot->x
-				&& y >= hotspot->y
-				&& x < hotspot->x + hotspot->width
-				&& y < hotspot->y + hotspot->height) {
-			if (HOTSPOT_IGNORE == hotspot->callback(output, hotspot, x, y,
-					button, hotspot->data)) {
-				return;
-			}
-		}
+	if(process_hotspots(output, pointer->x, pointer->y, button)) {
```
emersion: Style: missing space after `if`

swaybar/status_line.c
hvenev: Done.

swaybar/input.c
hvenev: Fixed.

emersion:
Thanks!

PR #4996, opened 2020-02-10
Fix inverted transforms

emersion:
This patch makes it so users that have configured their screen with a
transform don't have to update their config after the wlroots breaking
change.

References: https://github.com/swaywm/wlroots/pull/2023

### DIFF
--- a/sway/commands/output/transform.c
+++ b/sway/commands/output/transform.c
@@ -4,6 +4,22 @@
 #include "log.h"
 #include "sway/output.h"
 
+static enum wl_output_transform invert_rotation_direction(
+		enum wl_output_transform t) {
+	switch (t) {
+	case WL_OUTPUT_TRANSFORM_90:
+		return WL_OUTPUT_TRANSFORM_270;
+	case WL_OUTPUT_TRANSFORM_270:
+		return WL_OUTPUT_TRANSFORM_90;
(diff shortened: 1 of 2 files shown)


sway/commands/output/transform.c
```diff
@@ -32,6 +33,11 @@ struct cmd_results *output_cmd_transform(int argc, char **argv) {
 	} else {
 		return cmd_results_new(CMD_INVALID, "Invalid output transform.");
 	}
+
+	// Sway uses clockwise transforms, while WL_OUTPUT_TRANSFORM_* describe
+	// anti-clockwise transforms
+	transform = wlr_output_transform_invert(transform);
```
mstoeckl: While `wlr_output_transform_invert` correctly fixes the rotation direction for pure rotations, the inverse transformation of `WL_OUTPUT_TRANSFORM_FLIPPED_90` is itself, not the value `WL_OUTPUT_TRANSFORM_FLIPPED_270` that the wlroots patch swapped it with.

emersion: Oh, right. I guess the clockwise/anticlockwise handling needs to be fixed as well.

emersion:
Updated, also added a note in the docs.

ddevault:
Thanks!

Issue #4960, opened 2020-01-28
Scale factor defaults to 2.0 upon reload


kdryja:
Whenever I execute reload directive (e.g. via `$mod+Shift+c` from default config), the scaling of my screen gets bumped to 2.0 (from 1.0), even though nothing sets it explicitly in my config file.

Steps to repro:
1. Launch sway with default config from /etc/sway/config
2. Run the command:
```
$ swaymsg -t get_outputs
> Output eDP-1 'Unknown 0x0000 0x00000000' (focused)
  Current mode: 2560x1440 @ 59.998001 Hz
  Position: 0,0
  Scale factor: 1.000000
  Scale filter: nearest
  Subpixel hinting: unknown
  Transform: normal
  Workspace: 3
  Max render time: off
  Available modes:
    2560x1440 @ 59.998001 Hz
```
3. Reload sway (e.g. using `$mod+Shift+c`)
4. Run the command:
```
$ swaymsg -t get_outputs
> Output eDP-1 'Unknown 0x0000 0x00000000' (focused)
  Current mode: 2560x1440 @ 59.998001 Hz
  Position: 0,0
  Scale factor: 2.000000
  Scale filter: nearest
  Subpixel hinting: unknown
  Transform: normal
  Workspace: 3
  Max render time: off
  Available modes:
    2560x1440 @ 59.998001 Hz
```
5. Scale factor is now doubled

FWIW, I can work around this by putting `output eDP-1 scale 1` in my config file, thus forcing the scale to always be 1, but I'm pretty sure it wasn't necessary before. Is it intended?

* Sway Version:
```
$ swaymsg -t get_version
> sway version 1.4
```
* Debug Log: [sway.log](https://gist.github.com/kdryja/051b4d00a30af3c6bfba99ceaaf9673d#file-sway-log)
* Configuration File: Reproducible with the default config from /etc/sway/config Pasting just in case anyway: [config](https://gist.github.com/kdryja/051b4d00a30af3c6bfba99ceaaf9673d#file-config)

emersion:
This log has been obtained with Sway running nested inside another Wayland compositor. Please provide a debug log with Sway started from a TTY.

kdryja:
Apologies, see here:
[sway.log](https://gist.github.com/kdryja/051b4d00a30af3c6bfba99ceaaf9673d#file-sway-log-tty)

emersion:
Seems like the output scale auto-detection isn't working as expected here.

```
[...]

PR #4998, opened 2020-02-10
Fix output scale auto-detection

emersion:
Closes: https://github.com/swaywm/sway/issues/4960


emersion:
Indeed! Fixed.

RedSoxFan:
I think scale filter also needs to be moved back below scale and changed to use the pending scale.

emersion:
Yeah, this seems broken as well. Pushed a bunch of other commits to simplify `apply_output_config`.

RedSoxFan:
Thanks!

PR #5000, opened 2020-02-11
config: fix validation exit code and log level

### LINKED ISSUE #4976: Config validation always exits with 0
When I try to validate a configuration file `sway -c sway.conf -C` always exits with 0, independent of whether or not there is an error. It should exit with something >0 if an error occured.

This makes it unfit for automatic config checking, which would be helpful over at https://github.com/rycee/home-manager/pull/829.

Tested on sway v1.4.

RedSoxFan:
Fixes #4976

This makes it so invalid configs will return the exit code 1 when the
validation flag is given. This also reduces the log level to SWAY_ERROR,
which makes it so only the errors are shown. If someone wants more
verbose output, they can use the -V/--verbose or -d/--debug flags.
Additionally, this also makes it so swaybg will not be spawned when
validating the config.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -404,7 +404,7 @@ static bool load_config(const char *path, struct sway_config *config,
 		sway_log(SWAY_ERROR, "Error(s) loading config!");
 	}
 
-	return true;
+	return config->active || !config->validating || config_load_success;
 }
 
 bool load_main_config(const char *file, bool is_active, bool validating) {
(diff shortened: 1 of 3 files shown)


sway/config.c
```diff
@@ -404,7 +404,7 @@ static bool load_config(const char *path, struct sway_config *config,
 		sway_log(SWAY_ERROR, "Error(s) loading config!");
 	}
 
-	return true;
+	return config->active || !config->validating || config_load_success;
```
emersion: What's the reason why we ignore errors when not validating or when the config is already active?

Do we want not to make the `reoad` command fail on error?

RedSoxFan: I just left it at the current behavior for every case other than validating only. The current config options don't really need to be fatal. The errors just show in swaynag. The fatal ones would be the currently not implemented security ones.

emersion: The return value seems to just passed through down to the `reload` command handler. I don't know whether i3 makes the `reload` command fail on config error.

In any case, I agree the current patch is enough to fix the original issue.

PR #5009, opened 2020-02-12
i3compat: add `window_type` to IPC response

ammgws:
i3 added these in i3/i3#3797

Todo:
- update docs
- convert int to string? e.g. `...WINDOW_TYPE_NORMAL`

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -85,6 +85,35 @@ static const char *ipc_json_output_transform_description(enum wl_output_transfor
 	return NULL;
 }
 
+#if HAVE_XWAYLAND
+static const char *ipc_json_xwindow_type_description(enum atom_name window_type) {
+	switch (window_type) {
+	case NET_WM_WINDOW_TYPE_NORMAL:
+		return "normal";
+	case NET_WM_WINDOW_TYPE_DIALOG:
+		return "dialog";
+	case NET_WM_WINDOW_TYPE_UTILITY:
+		return "utility";
+	case NET_WM_WINDOW_TYPE_TOOLBAR:
+		return "toolbar";
+	case NET_WM_WINDOW_TYPE_SPLASH:
+		return "splash";
+	case NET_WM_WINDOW_TYPE_MENU:
+		return "menu";
+	case NET_WM_WINDOW_TYPE_DROPDOWN_MENU:
+		return "dropdown_menu";
+	case NET_WM_WINDOW_TYPE_POPUP_MENU:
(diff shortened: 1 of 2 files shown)


emersion:
i3 docs: https://github.com/i3/i3/pull/3931/files

> convert int to string?

Yes, we need to use strings rather than ints. A switch with values from `parse_window_type` should do it.

We need to guard the whole thing behind `HAVE_XWAYLAND`.

ammgws:
On mobile now, but pretty sure that whole code block is guarded already.

Will get onto the other stuff soon, cheers.

emersion:
Oh right, it's already guarded indeed.

emersion:
Can you also update the IPC man page?

emersion:
Looks good, can you squash everything into a single commit?

ammgws:
P.S.
For peace of mind I wanted to test with an X program that sets window type but haven't been able to find one yet and don't have the time at the moment to make a test program either.

emersion:
Testing with the "Save changes to XXX before closing?" dialog of `evn GDK_BACKEND=x11 gedit`, this doesn't seem to work. Sway reports "unknown", but `xprop` reports `_NET_WM_WINDOW_TYPE_DIALOG`.

ammgws:
Yeah I can reproduce that. After doing some printf debugging, I found that view_get_window_type is returning 280 for _NET_WM_WINDOW_TYPE_NORMAL and 305 for _NET_WM_WINDOW_TYPE_DIALOG.

RedSoxFan:
Referencing this should help https://github.com/swaywm/sway/blob/447967ad3a68b5a515581e4db76fb232cf56caa3/sway/desktop/xwayland.c#L230-L241

ammgws:
That did it, thanks.

20200216_15h04m32s

sway/ipc-json.c
```diff
@@ -85,6 +85,42 @@ static const char *ipc_json_output_transform_description(enum wl_output_transfor
 	return NULL;
 }
 
+#if HAVE_XWAYLAND
+static const char *ipc_json_xwindow_type_description(struct sway_view *view) {
```
ddevault: This needs an alternative implementation for when wlroots is built without Xwayland support.

emersion: The call to this function is already guarded behind `#if HAVE_XWAYLAND`.

ddevault: So it seems.

ddevault:
Thanks!

PR #5010, opened 2020-02-12
Remember if direct scanout fails

ascent12:
If a view fails direct scanout, we remember if it fails so we don't
bother trying again. This removes a lot of log spam of the atomic DRM
test failing in this case.

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -75,6 +75,7 @@ struct sway_view {
 	char *title_format;
 
 	bool using_csd;
+	bool failed_scan_out;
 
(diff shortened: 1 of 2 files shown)


emersion:
The view could change its size or buffer format and it could work again.

escalade:
Needs update, my log is getting spammed.

emersion:
This should help with the issue as well: https://github.com/swaywm/wlroots/pull/2818

emersion:
Superseded by https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3370

PR #5015, opened 2020-02-14
output: fix updating output manager config

### LINKED ISSUE #5014: Starting kanshi causes failed assertion in `wlr_output_management_v1.c` [already bisected]
## Problem Description

`sway` master crashes via failed assertion when starting `kanshi` with an empty config file.

## Versions and Logs

* `sway` Version: 1.4-f5fa0c71 (Feb 13 2020, branch 'HEAD'), AUR r6219.f5fa0c71-1
* `wlroots` Version: AUR 0.10.0.r9.g273b280f-1
* `kanshi` Version: AUR r80.86d6aff-1
* Debug Log, Backtrace, and Configuration File: https://gist.github.com/Ferdi265/60c62196aa8ce518bcbd4e0d46b947d9

## Additional Information

* Failed Assertion:
```
sway: types/wlr_output_management_v1.c:689: head_send_state: Assertion `head->state.mode != NULL || wl_list_empty(&head->state.output->modes)' failed.
```
Using a debugger I confirmed, that `head->state.mode` is `NULL` and `&head->state.output->modes` is not empty.

* Triggering Commit: f5fa0c7124dde3438a22550aae017f824d20c93b "Stop calling apply_output_config from output_enable"

RedSoxFan:
Fixes #5014 

The output manager config is created when the output is created. It is
updated when the mode, transform, scale, or layout for the output
changes, as well as, when the output is destroyed.

Since the output->enabled property was not being set before calling
apply_output_config, the output event handlers were early returning and
never updating the output manager config when the output state was
committed.

This fixes the issue by setting output->enabled in apply_output_config
below the output disabling section. There are also a few other minor
changes that are required to function.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -345,6 +345,9 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		return wlr_output_commit(wlr_output);
 	}
 
+	bool was_enabled = output->enabled;
+	output->enabled = true;
+
 	if (!oc || oc->dpms_state != DPMS_OFF) {
 		sway_log(SWAY_DEBUG, "Turning on output %s", wlr_output->name);
 		wlr_output_enable(wlr_output, true);
@@ -393,6 +396,7 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		// output disabled for now and try again when the output gets the mode
 		// we asked for.
 		sway_log(SWAY_ERROR, "Failed to modeset output %s", wlr_output->name);
+		output->enabled = was_enabled;
 		return false;
 	}
 
@@ -432,8 +436,8 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 	output->width = output_box->width;
 	output->height = output_box->height;
 
-	if ((!oc || oc->enabled) && !output->enabled) {
-		output_enable(output);
+	if ((!oc || oc->enabled) && !output->configured) {
+		output_configure(output);
 	}
 
 	if (oc && oc->dpms_state == DPMS_OFF) {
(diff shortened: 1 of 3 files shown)


sway/tree/output.c
```diff
@@ -102,20 +102,19 @@ struct sway_output *output_create(struct wlr_output *wlr_output) {
...
 
+	size_t len = sizeof(output->layers) / sizeof(output->layers[0]);
+	for (size_t i = 0; i < len; ++i) {
+		wl_list_init(&output->layers[i]);
+	}
+
 	return output;
 }
 
 void output_enable(struct sway_output *output) {
```
emersion: Hmm, maybe rename this to `configure_output`?

sway/config/output.c
```diff
@@ -393,6 +395,7 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		// output disabled for now and try again when the output gets the mode
 		// we asked for.
 		sway_log(SWAY_ERROR, "Failed to modeset output %s", wlr_output->name);
+		output->enabled = output->configured;
```
emersion: Maybe it makes more sense to save the previous value of `output->enabled` prior to the `apply_output_config` call, and change it back here?

sway/tree/output.c
RedSoxFan: Renamed `output_enable` to `output_configure`

sway/config/output.c
RedSoxFan: Changed. Although I'm pretty sure it's a logic error if the value did not match output->configured.

emersion: Yeah, the previous version worked. It just required the reader to understand that previously enabled outputs were always configured.

Issue #2333, opened 2018-07-23
Implement security configuration


SirCmpwn:
The commands and config file details can be the same as it was on 0.15, but we should change how we enforce them - only commands run by sway (i.e. via the exec command) should be able to have access to most secure features. We should do this by opening the wayland socket on their behalf and transferring the file descriptors through exec via the `WAYLAND_SOCKET` environment variable.

emersion:
Prior art: https://github.com/mupuf/libwsm (config files: https://github.com/mupuf/libwsm/tree/master/data/security/wsm/default)

emersion:
https://gitlab.freedesktop.org/wayland/weston/issues/206

BackSlasher:
Seeing that https://gitlab.freedesktop.org/wayland/weston/-/issues/206 is closed, do you have any idea how the followup on Sway looks like?
I'm migrating from i3wm, and happy to contribute some code if it can help

emersion:
The status quo is that sockets created via security-context-v1 (used by sandboxing engines such as Flatpak) don't get access to privileged protocols.

Still missing is a way to configure the compositor such that sandboxed clients get access to allow-listed privileged protocols. It's unclear how this is to be implemented.

BackSlasher:
I see. I was hoping there was a direction I can help implement. Thank you for explaining

Wolf480pl:
Hmm... the `sandbox_*` properties from security-context-v1 are already available through criteria in sway.conf...

what if:
```
security_context [sandbox_engine="org.flatpak" sandbox_appid="org.example.app"] {
    allow_protocol wlr_output_manager
    allow_protocol xdg_output_manager
    allow_drm_lease <some_device>
    deny_protocol wp_cursor_shape
}
```

(the criteria here would only be able to use `security_*` properties plus maybe other properties available at connection time, which is why I'm not proposing reusing `for_window`)
would that be a good direction?

chovanecadam:
That seems good to me. The user should also be able to change the default list of protocols denied for all security contexts by omitting the stuff in the square brackets.

PR #5021, opened 2020-02-17
Keyboard shortcuts inhibit

michaelweiser:
Adding support for the keyboard shortcuts inhibit protocol allows remote
desktop and virtualisation software to receive all keyboard input in
order to pass it through to their clients so users can fully interact
the their remote/virtual session. The software usually provides its own
key combination to release its "grab" to all keyboard input. The
inhibitor can be deactivated by the user by removing focus from the
surface using another input device such as the pointer.

Use support for the procotol in wlroots (swaywm/wlroots#2026) to add support to sway. Extend
the input manager with handlers for inhibitor creation and destruction
and appropriate bookkeeping. Attach the inhibitors to the seats they
apply to to avoid having to search the list of all currently existing
inhibitors on every keystroke and passing the inhibitor manager around.
Add a helper function to retrieve the inhibitor applying to the
currently focused surface of a seat, if one exists.

Extend bindsym with a flag for bindings that should be processed even if
an inhibitor is active. Conversely this disables all normal shortcuts if
an inhibitor is found for the currently focused surface in
keyboard::handle_key_event() since they don't have that flag set. Use
above helper function to determine if an inhibitor exists for the
surface that would eventually receive input.

Also, add a command to influence keyboard shortcuts inhibitors. In its current
form it can be used to activate, deactivate or toggle an existing
inhibitor on the surface currently receiving input. This can be used to
define an escape shortcut such as:
```
bindsym --inhibited $mod+Ctrl+Shift+q shortcuts_inhibitor deactivate
```

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -39,7 +39,7 @@ enum binding_flags {
 	BINDING_CONTENTS=8,  // mouse only; trigger on container contents
 	BINDING_TITLEBAR=16, // mouse only; trigger on container titlebar
 	BINDING_CODE=32,     // keyboard only; convert keysyms into keycodes
-	BINDING_RELOAD=62,   // switch only; (re)trigger binding on reload
+	BINDING_RELOAD=64,   // switch only; (re)trigger binding on reload
 };
 
 /**
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -177,6 +177,12 @@ enum seat_config_allow_constrain {
 	CONSTRAIN_DISABLE
 };
 
+enum seat_config_shortcuts_inhibit {
+	SHORTCUTS_INHIBIT_DEFAULT,  // the default is currently enabled
+	SHORTCUTS_INHIBIT_ENABLE,
+	SHORTCUTS_INHIBIT_DISABLE
+};
+
 enum seat_keyboard_grouping {
 	KEYBOARD_GROUP_DEFAULT,  // the default is currently smart
 	KEYBOARD_GROUP_NONE,
@@ -201,6 +207,7 @@ struct seat_config {
 	list_t *attachments; // list of seat_attachment configs
 	int hide_cursor_timeout;
 	enum seat_config_allow_constrain allow_constrain;
+	enum seat_config_shortcuts_inhibit shortcuts_inhibit;
 	enum seat_keyboard_grouping keyboard_grouping;
 	uint32_t idle_inhibit_sources, idle_wake_sources;
 	struct {
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -40,6 +40,7 @@ enum binding_flags {
 	BINDING_TITLEBAR=16, // mouse only; trigger on container titlebar
 	BINDING_CODE=32,     // keyboard only; convert keysyms into keycodes
 	BINDING_RELOAD=64,   // switch only; (re)trigger binding on reload
+	BINDING_INHIBITED=128,	// keyboard only: ignore shortcut inhibitor
 };
 
 /**
--- a/sway/commands/seat/shortcuts_inhibitor.c
+++ b/sway/commands/seat/shortcuts_inhibitor.c
@@ -1,0 +1,96 @@
+#include "log.h"
+#include "sway/commands.h"
+#include "sway/input/seat.h"
+#include "sway/input/input-manager.h"
+#include "util.h"
+
+static struct cmd_results *handle_action(struct seat_config *sc,
+		struct sway_seat *seat, const char *action) {
+	struct sway_keyboard_shortcuts_inhibitor *sway_inhibitor = NULL;
+	if (strcmp(action, "disable") == 0) {
+		sc->shortcuts_inhibit = SHORTCUTS_INHIBIT_DISABLE;
+
+		wl_list_for_each(sway_inhibitor,
+				&seat->keyboard_shortcuts_inhibitors, link) {
+			wlr_keyboard_shortcuts_inhibitor_v1_deactivate(
+					sway_inhibitor->inhibitor);
+		}
+
+		sway_log(SWAY_DEBUG, "Deactivated all keyboard shortcuts inhibitors");
+	} else {
+		sway_inhibitor = keyboard_shortcuts_inhibitor_get_for_focused_surface(seat);
+		if (!sway_inhibitor) {
+			return cmd_results_new(CMD_FAILURE,
+					"No inhibitor found for focused surface");
+		}
+
+		struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor =
+			sway_inhibitor->inhibitor;
+		bool inhibit;
+		if (strcmp(action, "activate") == 0) {
+			inhibit = true;
+		} else if (strcmp(action, "deactivate") == 0) {
+			inhibit = false;
+		} else if (strcmp(action, "toggle") == 0) {
+			inhibit = !inhibitor->active;
+		} else {
+			return cmd_results_new(CMD_INVALID, "Expected enable|"
+					"disable|activate|deactivate|toggle");
+		}
+
+		if (inhibit) {
+			wlr_keyboard_shortcuts_inhibitor_v1_activate(inhibitor);
+		} else {
+			wlr_keyboard_shortcuts_inhibitor_v1_deactivate(inhibitor);
+		}
+
+		sway_log(SWAY_DEBUG, "%sctivated keyboard shortcuts inhibitor",
+				inhibit ? "A" : "Dea");
+	}
+
+	return cmd_results_new(CMD_SUCCESS, NULL);
+}
+
+// shortcuts_inhibitor [enable|disable|activate|deactivate|toggle]
+struct cmd_results *seat_cmd_shortcuts_inhibitor(int argc, char **argv) {
+	struct cmd_results *error =
+		checkarg(argc, "shortcuts_inhibitor", EXPECTED_EQUAL_TO, 1);
+	if (error) {
+		return error;
+	}
+
+	struct seat_config *sc = config->handler_context.seat_config;
+	if (!sc) {
+		return cmd_results_new(CMD_FAILURE, "No seat defined");
+	}
+
+	if (strcmp(argv[0], "enable") == 0) {
+		sc->shortcuts_inhibit = SHORTCUTS_INHIBIT_ENABLE;
+	// at runtime disable is an action that also deactivates all active
+	// inhibitors handled in handle_action()
+	} else if (strcmp(argv[0], "disable") == 0 && !config->active) {
+		sc->shortcuts_inhibit = SHORTCUTS_INHIBIT_DISABLE;
+	} else if (!config->active) {
+		return cmd_results_new(CMD_INVALID, "only enable and disable "
+				"can be used in the config");
+	} else {
+		if (strcmp(sc->name, "*") != 0) {
+			struct sway_seat *seat = input_manager_get_seat(sc->name, false);
+			if (!seat) {
+				return cmd_results_new(CMD_FAILURE,
+						"Seat %s does not exist", sc->name);
+			}
+			error = handle_action(sc, seat, argv[0]);
+		} else {
+			struct sway_seat *seat = NULL;
+			wl_list_for_each(seat, &server.input->seats, link) {
+				error = handle_action(sc, seat, argv[0]);
+				if (error && error->status != CMD_SUCCESS) {
+					break;
+				}
+			}
+		}
+	}
+
+	return error ? error : cmd_results_new(CMD_SUCCESS, NULL);
+}
--- a/sway/input/input-manager.c
+++ b/sway/input/input-manager.c
@@ -296,6 +296,46 @@ static void handle_inhibit_deactivate(struct wl_listener *listener, void *data) 
 	}
 }
 
+static void handle_keyboard_shortcuts_inhibitor_destroy(
+		struct wl_listener *listener, void *data) {
+	struct sway_keyboard_shortcuts_inhibitor *sway_inhibitor =
+		wl_container_of(listener, sway_inhibitor, destroy);
+
+	sway_log(SWAY_DEBUG, "Removing keyboard shortcuts inhibitor");
+
+	// sway_seat::keyboard_shortcuts_inhibitors
+	wl_list_remove(&sway_inhibitor->link);
+	wl_list_remove(&sway_inhibitor->destroy.link);
+	free(sway_inhibitor);
+}
+
+static void handle_keyboard_shortcuts_inhibit_new_inhibitor(
+		struct wl_listener *listener, void *data) {
+	struct sway_input_manager *input_manager =
+		wl_container_of(listener, input_manager,
+				keyboard_shortcuts_inhibit_new_inhibitor);
+	struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor = data;
+
+	sway_log(SWAY_DEBUG, "Adding keyboard shortcuts inhibitor");
+
+	struct sway_keyboard_shortcuts_inhibitor *sway_inhibitor =
+		calloc(1, sizeof(struct sway_keyboard_shortcuts_inhibitor));
+	if (!sway_assert(sway_inhibitor, "could not allocate keyboard "
+				"shortcuts inhibitor")) {
+		return;
+	}
+	sway_inhibitor->inhibitor = inhibitor;
+
+	sway_inhibitor->destroy.notify = handle_keyboard_shortcuts_inhibitor_destroy;
+	wl_signal_add(&inhibitor->events.destroy, &sway_inhibitor->destroy);
+
+	// attach inhibitor to the seat it applies to
+	struct sway_seat *seat = inhibitor->seat->data;
+	wl_list_insert(&seat->keyboard_shortcuts_inhibitors, &sway_inhibitor->link);
+
+	wlr_keyboard_shortcuts_inhibitor_v1_activate(inhibitor);
+}
+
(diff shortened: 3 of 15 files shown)


include/sway/config.h
```diff
@@ -40,6 +40,7 @@ enum binding_flags {
 	BINDING_TITLEBAR=16, // mouse only; trigger on container titlebar
 	BINDING_CODE=32,     // keyboard only; convert keysyms into keycodes
 	BINDING_RELOAD=62,   // switch only; (re)trigger binding on reload
```
michaelweiser: OT: Shouldn't this be 64?

ddevault: Yep.

michaelweiser: Separate PR or drive-by fix in this one? Have no idea how to and no real inclination of going down the rabbithole of retesting it, TBH. :-/ Had a quick glance where it's used and seems trivial enough, though.

ddevault: A second commit in this PR would be fine imo.

michaelweiser:
I'm struggling with what to do with the compositor built-in bindings in `keyboard::handle_key_event()`. Initially I thought that switching VTs (which is all they're implemeting currently) is such a fundamental recovery technique that I shouldn't allow inhibitors to suppress them by default. So I made inhibitors not apply to them for now.

Instead I wanted to add an additional oneshot passthrough command similar to what some VM software does, i.e. have a keybinding that, if activated, will make sway ignore the next binding and instead pass it through to the client. That way you could press e.g. Ctrl+Alt+Space which would activate oneshot passthrough and then press Ctrl+Alt+F2 which would switch the VT in the remote or virtual session instead of locally.

I got a first version implemented but it was quite cumbersome and literally just made sway ignore the next binding, no matter how far in the future that may be and how many non-binding keystrokes came inbetween.

To be intuitive, oneshot passthrough should IMO also disable itself on any non-binding key combo that *was* indeed passed to the client. However, the way `keyboard::handle_key_event()` processes key events appears particularly ill suited for this kind of thing (or my brain is). Also, looking at the existing complexity my guess is that any additional complexity in that code path would better be avoided.

Long story short: Should I bother or could we instead have a default/recommended escape key binding for uncooperative clients/inhibitors?

include/sway/config.h
michaelweiser: Done. Thanks!

ddevault:
Would it make sense to shelve that for a future patch?

michaelweiser:
> Would it make sense to shelve that for a future patch?
Indeed. After some discussion and brainstorming (Thanks @wolf-z!) I came to realize that most virtualization and remote desktop software has virtual keyboards, shortcut sending menus or special key combos to work around exactly this issue. So I think the current implementation is in line with existing ones. I quickly retested with GNOME and virt-manager and found that it also does not inhibit Ctrl+Alt+Fx and virt-manager provides a menu to send them to the guest.

BTW: I also came across this article which mentions a warning dialog in GNOME that an application wants to inhibit the keyboard and that the escape key combo is Super+Escape: https://unix.stackexchange.com/questions/417670/virtual-machine-manager-wants-to-inhibit-shortcuts-again-and-again-on-waylan. I'm not saying we should have a warning dialog but I like the Super+Escape key combination.

sway/commands/shortcuts_inhibitor.c
```diff
@@ -0,0 +1,40 @@
+#include "sway/commands.h"
+#include "sway/input/seat.h"
+#include "sway/input/input-manager.h"
+
+struct cmd_results *cmd_shortcuts_inhibitor(int argc, char **argv) {
+	struct cmd_results *error = checkarg(argc, "shortcuts_inhibitor",
```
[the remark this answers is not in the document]
michaelweiser: Done.

sway/commands/shortcuts_inhibitor.c
```diff
@@ -0,0 +1,40 @@
...
+			EXPECTED_EQUAL_TO, 1);
+	if (error) {
+		return error;
+	}
+
+	struct sway_seat *seat = config->handler_context.seat;
+	struct sway_keyboard_shortcuts_inhibitor *sway_inhibitor =
+		keyboard_shortcuts_inhibitor_get_for_focused_surface(seat);
+	if (!sway_inhibitor) {
+		return cmd_results_new(CMD_SUCCESS, NULL);
```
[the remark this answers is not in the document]
michaelweiser: Done.

sway/input/input-manager.c
```diff
@@ -296,6 +296,48 @@ static void handle_inhibit_deactivate(struct wl_listener *listener, void *data)
...
+				"shortcuts inhibitor")) {
+		return;
+	}
+	sway_inhibitor->inhibitor = inhibitor;
+
+	sway_inhibitor->destroy.notify = handle_keyboard_shortcuts_inhibitor_destroy;
+	wl_signal_add(&inhibitor->events.destroy, &sway_inhibitor->destroy);
+
+	// attach inhibitor to the seat it applies to
+	struct sway_seat *seat = NULL;
```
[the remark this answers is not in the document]
michaelweiser: Uncanny. :) Done.

sway/input/keyboard.c
```diff
@@ -150,16 +150,18 @@ static bool update_shortcut_state(struct sway_shortcut_state *state,
...
 	for (int i = 0; i < bindings->length; ++i) {
 		struct sway_binding *binding = bindings->items[i];
 		bool binding_locked = (binding->flags & BINDING_LOCKED) != 0;
+		bool binding_inhibited = (binding->flags & BINDING_INHIBITED) != 0;
 		bool binding_release = binding->flags & BINDING_RELEASE;
 
 		if (modifiers ^ binding->modifiers ||
 				release != binding_release ||
 				locked > binding_locked ||
+				inhibited != binding_inhibited ||
```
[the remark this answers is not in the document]
michaelweiser: I had this programmed out once already but then decided that I did not want --inhibited bindsyms to leak into normal modes. But I agree that I certainly also want to be able to mark my brightness and volume bindsyms as non-inhibited. So: Done. Thanks for making me rethink this.

sway/commands/shortcuts_inhibitor.c
```diff
@@ -0,0 +1,40 @@
...
+#include "sway/input/input-manager.h"
+
+struct cmd_results *cmd_shortcuts_inhibitor(int argc, char **argv) {
+	struct cmd_results *error = checkarg(argc, "shortcuts_inhibitor",
+			EXPECTED_EQUAL_TO, 1);
+	if (error) {
+		return error;
+	}
+
+	struct sway_seat *seat = config->handler_context.seat;
```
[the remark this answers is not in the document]
michaelweiser: Done. While at it I took inspiration from `pointer_constraint` and added a per-`seat_config` (and thus with `*` potentially global) enable/disable default. This actually allows the user to decide when to activate an inhibitor.

The inhibitor actions don't seem to fit into this all that well. I had to redefine the wildcard `seat *` to mean "current seat"  to keep semantics. What do you think of this?

What's also left is the question how the user could get some indication of an existing inhibitor on a window and its state. Maybe a border color change? I had thought of a criteria for that. What do you think and how would I go about it?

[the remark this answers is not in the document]
michaelweiser: I have the criteria coded up but it doesn't fully work because criteria are only executed once. So a config like:
```
for_window [shortcuts_inhibitor=active] title_format !+%title
for_window [shortcuts_inhibitor=inactive] title_format !-%title
for_window [shortcuts_inhibitor=absent] title_format %title
```
will happily upgrade the window title to '!+%title` as an inhibitor becomes active but not back down if it goes inactive or gets destroyed.

Now I'm thinking of executing a configurable command as the inhibitor state changes but I can't find any example for such a `command_to_exec_on_some_condition` hook-like functionality. What would the sway-native approach be here? Have a script listen to sway-ipc and send the `title_format` command?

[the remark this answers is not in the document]
RedSoxFan: > The inhibitor actions don't seem to fit into this all that well. I had to redefine the wildcard seat * to mean "current seat" to keep semantics. What do you think of this?

There is already `seat -` for the current seat. `seat *` should work fine for `enable|disable`. Either return a result with the status `CMD_INVALID` for `activate|deactivate` on the wildcard seat or activate/deactivate all shortcut inhibitors. I'm not really sure which is better.

> What's also left is the question how the user could get some indication of an existing inhibitor on a window and its state. Maybe a border color change? I had thought of a criteria for that. What do you think and how would I go about it?
> [..]
> What would the sway-native approach be here? Have a script listen to sway-ipc and send the title_format command?

Information for none of the inhibitor types is currently exposed via IPC. It could be added to get_tree response though. Any other change would be out of scope for this PR and would have to be discussed as a separate one.

sway/input/keyboard.c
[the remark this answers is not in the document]
RedSoxFan: Just to explain the reason why we have it this way, it is to remove duplication of bindings. You can always specify different actions (or no action with `nop`) for a more specific case.
```
# Both
bindsym $mod+Escape --inhibited exec notify-send "Both Normal and Inhbited"

# Just Normal
bindsym $mod+Escape exec notify-send "Only Normal Shortcuts"

# Just Inhibited (order doesn't matter)
bindsym $mod+Escape nop
bindsym $mod+Escape --inhibited notify-send "Only Inhibited"
```

sway/commands/seat/shortcuts_inhibitor.c
```diff
@@ -0,0 +1,121 @@
+#include "sway/commands.h"
```
RedSoxFan: Not a huge fan of this entire file. I'd recommend taking inspiration from https://github.com/swaywm/sway/blob/e1eda8a09c15780e7f6ab0b6a1ab7e811a3d8764/sway/commands/urgent.c

Something like:
```c
// Validate config->handler_context.seat_config is not NULL
if (strcmp(argv[0], "enable") == 0) {  // Possibly rename to `allow` for consistency
    sc->allow_shortcuts_inhibit = true;
} else if (strcmp(argv[0], "disable") == 0) {  // Possibly rename to `deny` for consistency
    sc->allow_shortcuts_inhibit = false;
} else if (strcmp(sc->name, "*") == 0) {
    return cmd_results(CMD_INVALID, "cannot activate/deactivate for wildcard seat");
} else if (!config->active) {
    return cmd_results(CMD_INVALID, "only enable and disable can be used in the config");
} else {
    // Validate seat exists
    // Validate inhibitor exists
    // Use parse_boolean(argv[0], inhibitor->active) to get state to set
}
```

RedSoxFan:
I think `disable` should also deactivate all existing inhibitors on that seat. I don't think the inverse is true though. The proper place to deactivate them would be in https://github.com/swaywm/sway/blob/master/sway/input/seat.c#L1348

sway/commands/seat/shortcuts_inhibitor.c
michaelweiser: Done. Lost the operations enum. I stuck with enable/disable instead of allow/deny because deny in particular suggests that we have something to say in the matter, i.e. would actively refuse inhibitors. In think "disable" makes it more clear that we're just not going to honour the inhibitors. I pruned all other uses of "allow" to be consistent.

michaelweiser:
> I think `disable` should also deactivate all existing inhibitors on that seat. I don't think the inverse is true though.

Agreed and Done.

> The proper place to deactivate them would be in https://github.com/swaywm/sway/blob/master/sway/input/seat.c#L1348

I did that but then, once inhibitors being disabled has become part of the config, it re-deactivates all inhibitors on any succeeding seat config command, including those that just activated one. I think it's a conflict between *changing the config* and *executing an action on a config change*. I agree that we should disable all active inhibitors once the user decides not to honour any new ones either but from then on the user should IMO also be able to explicitly (re-)activate them. So I made this part of the seat command for now, which somewhat complicates control flow in there. Alternatively I could think of somehow tracking changes through the config merge so that deactivation in `seat_apply_config()` only happens if the config item just changed *to* disabled from another value. Have I missed some existing infrastructure that permits that already?

sway/commands/shortcuts_inhibitor.c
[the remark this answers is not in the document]
michaelweiser: Done. I had totally missed `seat -`. I've gone the route of applying the active/inactive/toggle actions to the current/named/all seat(s) because it seems most consistent to me.

Is there a way to avoid iterating all seats and all inhibitors when trying to add an inhibitor indicator to `get_tree`? For now I've added a note to the manpage that indicator indication is up to the client.

michaelweiser:
Is there anything I should be doing here?

sway/sway-input.5.scd
```diff
@@ -239,6 +239,25 @@ correct seat.
 	by default) for the seat. This is primarily useful for video games. The
 	"escape" command can be used at runtime to escape from a captured client.
 
+*seat* <name> shortcuts_inhibitor enable|disable|active|inactive|toggle
```
emersion: Maybe we should rename `active` to `activate`, just to make it clear the command performs an action and isn't read-only.

sway/commands/seat/shortcuts_inhibitor.c
```diff
@@ -0,0 +1,86 @@
...
+	if (error) {
+		return error;
+	}
+
+	struct seat_config *sc = config->handler_context.seat_config;
+	if (!sc) {
+		return cmd_results_new(CMD_FAILURE, "No seat defined");
+	}
+
+	if (strcmp(argv[0], "enable") == 0) {
```
emersion: When used at runtime, this should probably also deactivate any active inhibitor

michaelweiser: I had thought about *activating* all existing inhibitors when enabling the overall functionality but that seems overzealous to me. I agree with @RedSoxFan that disabling the functionality should also deactivate all active inhibitors for good measure and consistency but why should enabling it do the same?

emersion: Oh, sorry, I misread your code. I agree, only `disable` should be a special-case.

emersion:
Follow-up issue: decide how to let users allow/deny keyboard shortcut inhibition requests on a per-client basis.

* `for_window […] shortcuts_inhibitor enable|disable`?
* Somehow add a way to run a command which returns zero to allow the client to inhibit shortcuts?
* Leave this feature for the more general Wayland protocol security framework (https://github.com/swaywm/sway/issues/2333)?

sway/sway-input.5.scd
michaelweiser: @RedSoxFan suggested the use of `parse_boolean()` which handles the special `toggle` case and accepts `yes`, `on`, `true`, `enable`, `enabled`, and `active`. Neither of those matches particularly well and I also don't like the fact that it silently accepts anything else as false. So I would eliminate it again and explicitly match for `activate`, `deactivate` and `toggle` as before, perhaps without the enum boilerplate. Would that be acceptable? Alternatively, we could extend `parse_boolean()`...

sway/commands/seat/shortcuts_inhibitor.c
michaelweiser: Ah, well, if it's easily misread then maybe control flow is too complicated now. Do you have a suggestion where I maybe didn't fully understand @RedSoxFan's suggestion of doing the disabling int `seat_apply_config()`?

michaelweiser:
> Follow-up issue: decide how to let users allow/deny keyboard shortcut inhibition requests on a per-client basis.
> 
>     * `for_window […] shortcuts_inhibitor enable|disable`?

Apart from issue of addressing the currently focused window with `seat -`, the following should allow inhibitors in general but immediately deactivate them for specific kinds of windows:
```
seat * shortcuts_inhibitor enable
for_window [...] seat - shortcuts_inhibitor deactivate
```
But since criteria seems to be applied only once, this would not guard against the client destroying and reestablishing a new inhibitor, which they seem to do quite aggressively.
So we'd need a per-window/view, non-per-seat version of the command and a per-window/view flag for it, right?

>     * Somehow add a way to run a command which returns zero to allow the client to inhibit shortcuts?
I don't get this one. Can you elaborate?

>     * Leave this feature for the more general Wayland protocol security framework (#2333)?
I'd rather do this now and port to a more generic framework later than to defer (indefinitely).

emersion:
>So we'd need a per-window/view, non-per-seat version of the command and a per-window/view flag for it, right?

Yeah, that's what I meant. I think we should leave it for a follow-up PR though, to keep this one easier to review.

sway/sway-input.5.scd
michaelweiser: Switched from parse_boolean back to explicit parsing. Also simplified the control flow. I could comfortably live with this latest incarnation, fwiw.

michaelweiser:
> I think we should leave it for a follow-up PR though, to keep this one easier to review.

Agreed.

sway/commands/seat/shortcuts_inhibitor.c
```diff
@@ -0,0 +1,96 @@
...
+			if (!seat) {
+				return cmd_results_new(CMD_FAILURE,
+						"Seat %s does not exist", sc->name);
+			}
+			error = handle_action(sc, seat, argv[0]);
+		} else {
+			struct sway_seat *seat = NULL;
+			wl_list_for_each(seat, &server.input->seats, link) {
+				error = handle_action(sc, seat, argv[0]);
+				if ((error && error->status != CMD_SUCCESS)) {
```
RedSoxFan: only one set of parenthesis are needed

michaelweiser: Removed, also in `cursor.c` where they came from by separate patch. Squashed the commits, quickly retested that everything still works as expected. Ready to go from my side.

RedSoxFan:
Thanks!

PR #5022, opened 2020-02-17
tree/view: do not attempt enter on hidden child

RedSoxFan:
If a subsurface is created for a surface that is associated with a
scratchpad hidden view, do not attempt to send an enter to it. The
subsurface is not on any output and since there is no workspace
associated with the view, attempting to get the output for the NULL
workspace will result in a SIGSEGV.

There is likely other ways to reproduce, but I ran into this with the following for_window rule and attempting to do a video conference in firefox-nightly using wayland backend for firefox.
```
for_window [title="Nightly - Sharing Indicator"] move to scratchpad
```


emersion:
Thanks!

PR #5046, opened 2020-02-27
Only destroy swaybar surface via ipc when needed

vilhalmer:
Currently when changing properties via ipc like `workspace_buttons`, which don't affect the bar's surface at all, the surface is still destroyed and recreated. This causes sway to recalculate the workspace height twice in rapid succession, which is rather unpleasant. This fix only requests destruction if the mode was changed.


ddevault:
Thanks!

PR #5048, opened 2020-02-27
input: implement cycling through keyboard layout list

mucamaca:
This change enables users to change layout relative to the current position on the layout list. This way one can for example `bindsym Mod1+space input * xkb_layout_switch next` and iterate through their entire list of keymaps with a single shortcut.

### DIFF
--- a/sway/commands/input/xkb_switch_layout.c
+++ b/sway/commands/input/xkb_switch_layout.c
@@ -1,4 +1,5 @@
 #define _POSIX_C_SOURCE 200809L
+#include <assert.h>
 #include "sway/config.h"
 #include "sway/commands.h"
 #include "sway/input/input-manager.h"
@@ -11,6 +12,26 @@ static void switch_layout(struct wlr_keyboard *kbd, xkb_layout_index_t idx) {
 	}
 	wlr_keyboard_notify_modifiers(kbd, kbd->modifiers.depressed,
 		kbd->modifiers.latched, kbd->modifiers.locked, idx);
+}
+
+static xkb_layout_index_t get_current_layout_index(struct wlr_keyboard *kbd) {
+	xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(kbd->keymap);
+	assert(num_layouts > 0);
+
+	xkb_layout_index_t layout_idx;
+	for (layout_idx = 0; layout_idx < num_layouts; layout_idx++) {
+		if (xkb_state_layout_index_is_active(kbd->xkb_state,
+				layout_idx, XKB_STATE_LAYOUT_EFFECTIVE)) {
+			break;
+		}
+	}
+	return layout_idx;
+}
+
+static void switch_layout_relative(struct wlr_keyboard *kbd, int dir) {
+	xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(kbd->keymap);
+	xkb_layout_index_t idx = get_current_layout_index(kbd);
+	switch_layout(kbd, (idx + num_layouts + dir) % num_layouts);
 }
 
 struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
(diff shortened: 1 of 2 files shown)


ddevault:
Doesn't xkb have this feature built-in?

emersion:
It does, but only for a pre-defined set of keybinds. This prevents issues with pressed keys not being released while switching the keymap (e.g. press "A", switch keymap, then A's keycode doesn't map to the same keysym and the wrong keysym is released).

BTW, we should really make sure to only switch the keymap when all keys are released.

mucamaca:
> It does, but only for a pre-defined set of keybinds. This prevents issues with pressed keys not being released while switching the keymap (e.g. press "A", switch keymap, then A's keycode doesn't map to the same keysym and the wrong keysym is released).

So the way I implemented this is not correct?

sway/commands/input/xkb_switch_layout.c
```diff
@@ -13,6 +13,26 @@ static void switch_layout(struct wlr_keyboard *kbd, xkb_layout_index_t idx) {
...
+	}
+	return layout_idx;
+}
+
+static void switch_layout_relative(struct wlr_keyboard *kbd, int dir) {
+	xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(kbd->keymap);
+	xkb_layout_index_t idx = get_current_layout_index(kbd);
+	wlr_keyboard_notify_modifiers(kbd, kbd->modifiers.depressed,
+		kbd->modifiers.latched, kbd->modifiers.locked,
+		((idx + num_layouts) + dir) % num_layouts);
```
emersion: Style: unnecessary parentheses here

sway/commands/input/xkb_switch_layout.c
```diff
@@ -13,6 +13,26 @@ static void switch_layout(struct wlr_keyboard *kbd, xkb_layout_index_t idx) {
 		kbd->modifiers.latched, kbd->modifiers.locked, idx);
 }
 
+static xkb_layout_index_t get_current_layout_index(struct wlr_keyboard *kbd) {
+	xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(kbd->keymap);
```
emersion: I'd like to add `assert(num_layouts > 0)` here to ensure we don't return garbage from this function.

sway/commands/input/xkb_switch_layout.c
```diff
@@ -13,6 +13,26 @@ static void switch_layout(struct wlr_keyboard *kbd, xkb_layout_index_t idx) {
...
+			break;
+		}
+	}
+	return layout_idx;
+}
+
+static void switch_layout_relative(struct wlr_keyboard *kbd, int dir) {
+	xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(kbd->keymap);
+	xkb_layout_index_t idx = get_current_layout_index(kbd);
+	wlr_keyboard_notify_modifiers(kbd, kbd->modifiers.depressed,
```
emersion: Can't we re-use `switch_layout` in this function?

sway/commands/input/xkb_switch_layout.c
```diff
@@ -28,11 +48,21 @@ struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
 	}
 
 	const char *layout_str = argv[0];
+	int relative, layout;
 
-	char *end;
-	int layout = strtol(layout_str, &end, 10);
-	if (layout_str[0] == '\0' || end[0] != '\0' || layout < 0) {
-		return cmd_results_new(CMD_FAILURE, "Invalid layout index.");
+	if (!strcmp(layout_str, "next")) {
```
emersion: Style: I'd prefer to use `strcmp(…) == 0` instead. `!strcmp(…)` is error-prone.

sway/sway-input.5.scd
```diff
@@ -71,6 +71,11 @@ For more information on these xkb configuration options, see
 	layouts are configured with *xkb_layout*. A list of layouts you can switch
 	between can be obtained with *swaymsg -t get_inputs*.
 
+*input* <identifier> xkb_switch_layout next|prev
```
emersion: This section should be merge with the previous one.

emersion:
>So the way I implemented this is not correct?

No, it's fine, as the incorrect behavior would come from the `xkb_switch_layout` command itself. So it's a separate issue.

sway/commands/input/xkb_switch_layout.c
mucamaca: This was to prevent potential overflow with `idx=0`, `dir=-1` but it probably doesn't really matter.

Issue #1930, opened 2018-05-06
swaybar: use wl_pointer.axis_discrete when available


emersion:
Scrolling on a touchpad is not possible because it's way too fast. Using `wl_pointer.axis_discrete` should fix the issue.

Depends on https://github.com/swaywm/wlroots/issues/905

c-edw:
Just found a bug that I think might be related to this. If you scroll on the bar using a mouse, it will reach a 'boundary' and stop. However, if you use a touchpad to scroll, it will wrap back around (though you have to scroll quite fast). @emersion Please confirm whether this would be fixed with this change.

emersion:
Yeah, I think it's related. I haven't tested if libinput correctly sends correct discrete events, but I hope so.

emersion:
Closed by https://github.com/swaywm/sway/pull/5067

PR #5067, opened 2020-03-03
swaybar: Fix scrolling to switch workspace with precise trackpads

mortie:
This patch implements some logic to try to emulate scrolling in discrete steps for devices which scroll continuously, such as modern touchpads. It keeps track of whether the axis event is discrete or not, so it shouldn't have any effect on switching workspaces with scroll wheels (assuming an axis_discrete event is always fired before an axis event for devices which scroll wheels, which seems to be the case).

This is related to https://github.com/swaywm/sway/issues/1930.

Note that this patch makes swaybar depend on wl_seat v5 instead of v3.

Here's exactly how the decision about whether to switch workspaces or not is done:

``` C
bool do_scroll =
	(axis == WL_POINTER_AXIS_VERTICAL_SCROLL && seat->scroll_is_discrete_vertical) ||
	(axis == WL_POINTER_AXIS_HORIZONTAL_SCROLL && seat->scroll_is_discrete_horizontal) ||
	fabs(amt) >= 10 ||
	(fabs(amt) >= 1 && time >= seat->scroll_time + 200);
```

Switch workspace if:

* The event is from a device with discrete scroll steps.
* The movement in this one event is really big. This allows for making a large motion to scroll through lots of workspaces at once, typically to get to the first or last workspace.
* The time since the last workspace switch is at least 200ms (and the event's delta isn't tiny). This makes a slow, controlled motion first switch workspace instantly, then every 200ms.

This feels pretty good to me on my laptop with my touchpad, but I haven't tested it on others. Maybe some tuning of the values is necessary (or maybe they should even be configurable).

### DIFF
--- a/swaybar/input.c
+++ b/swaybar/input.c
@@ -211,18 +211,17 @@ static void workspace_next(struct swaybar *bar, struct swaybar_output *output,
 
 	if (new) {
 		ipc_send_workspace_command(bar, new->name);
+
+		// Since we're asking Sway to switch to 'new', it should become visible.
+		// Marking it visible right now allows calling workspace_next in a loop.
+		new->visible = true;
+		active->visible = false;
 	}
 }
 
-static void wl_pointer_axis(void *data, struct wl_pointer *wl_pointer,
-		uint32_t time, uint32_t axis, wl_fixed_t value) {
-	struct swaybar_seat *seat = data;
-	struct swaybar_pointer *pointer = &seat->pointer;
-	struct swaybar_output *output = pointer->current;
-	if (!sway_assert(output, "axis with no active output")) {
-		return;
-	}
-
+static void process_discrete_scroll(struct swaybar_seat *seat,
+		struct swaybar_output *output, struct swaybar_pointer *pointer,
+		uint32_t axis, wl_fixed_t value) {
 	// If there is a button press binding, execute it, skip default behavior,
 	// and check button release bindings
 	uint32_t button = wl_axis_to_button(axis, value);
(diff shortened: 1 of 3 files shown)


mortie:
I see the FreeBSD CI build failed, but it failed with:

```
pkg: wrong architecture: FreeBSD:12.0:amd64 instead of FreeBSD:12:amd64
pkg: repository FreeBSD contains packages with wrong ABI: FreeBSD:12.0:amd64
Unable to update repository FreeBSD
Error updating repositories!
```

emersion:
Yeah, don't worry about it, that's a FreeBSD bug. They're working on fixing it.

ianyfan:
It's been a while since I've worked on this, so I may be wrong, but I'm not sure if this is very robust in the case that a seat has both a trackpad and a mousewheel attached. I think a complete solution would utilise discrete axis events separately.

mortie:
We discussed that kind of thing in #sway-devel, I think we concluded that this way of implementing it should be safe. This solution works in the case where we receive an axis_discrete event for axis A, followed by an axis event for axis B, followed by the axis event for axis A; A is correctly treated as discrete, while B is treated as continuous.

The only way for it to fail is if we receive an axis_discrete for axis A, followed by an axis event for the same axis A which should've been treated as continuous, followed by the axis event for axis A which should've been treated as discrete - and I think that sequence is illegal.

I'm not 100% confident in this interpretation though, and I'm ready to be proven wrong if you disagree. Also, if you have a suggestion for a better way to figure out whether an axis event is for a discrete or continuous scroll, I'd be happy to implement it.

Synthetica9:
If I understand correctly, this would only fire scroll events every 200ms when scrolling continuously? On Waybar, we keep an internal counter to see how far we've scrolled yet, which seems to work very well. (See https://github.com/Alexays/Waybar/pull/381)

mortie:
What's the status on this? Would people prefer if I removed the timing component and re-introduced the counter, doing something like waybar? Or is this way fine and people are just busy with other things?

Either is ok, but I think I can fairly easily change my patch to use a counter in a way which still feels good. I just need to know if it's worthwhile or not.

swaybar/input.c
```diff
@@ -246,7 +246,24 @@ static void wl_pointer_axis(void *data, struct wl_pointer *wl_pointer,
...
+		(axis == WL_POINTER_AXIS_VERTICAL_SCROLL && seat->scroll_is_discrete_vertical) ||
+		(axis == WL_POINTER_AXIS_HORIZONTAL_SCROLL && seat->scroll_is_discrete_horizontal) ||
+		fabs(amt) >= 10 ||
+		(fabs(amt) >= 1 && time >= seat->scroll_time + 200);
+
+	// Clear flags in case the next event isn't discrete
+	if (axis == WL_POINTER_AXIS_VERTICAL_SCROLL)
+		seat->scroll_is_discrete_vertical = false;
+	else if (axis == WL_POINTER_AXIS_HORIZONTAL_SCROLL)
+		seat->scroll_is_discrete_horizontal = false;
```
ddevault: Braces are mandatory

ddevault:
This code should be reworked to use the frame event instead. We should be accumulating pointer events into a single struct, then defer processing it until we get a frame event.

swaybar/input.c
```diff
@@ -268,7 +285,12 @@ static void wl_pointer_axis_stop(void *data, struct wl_pointer *wl_pointer,
 
 static void wl_pointer_axis_discrete(void *data, struct wl_pointer *wl_pointer,
 		uint32_t axis, int32_t discrete) {
-	// Who cares
+	struct swaybar_seat *seat = data;
```
emersion: Why not use the `discrete` value here?

mortie:
Alright, I reworked my patches to accumulate events and then act on them in the 'frame' event. I'm also respecting the 'discrete'' value in 'axis_discrete', and removed my weird timing behaviour in favour of a plain "accumulate scroll distance, act when it crosses some threshold" system.

ddevault:
Thanks!

PR #5071, opened 2020-03-04
Null check sway_layer_surface when checking damage

tadeokondrak:
Before this change, an overlay layer-shell surface without a buffer
attached would segfault the compositor on screen damage.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -388,6 +388,9 @@ bool output_has_opaque_overlay_layer_surface(struct sway_output *output) {
 		struct wlr_surface *wlr_surface = wlr_layer_surface_v1->surface;
 		struct sway_layer_surface *sway_layer_surface =
 			layer_from_wlr_layer_surface_v1(wlr_layer_surface_v1);
+		if (sway_layer_surface == NULL) {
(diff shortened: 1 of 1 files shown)


emersion:
The `sway_layer_surface` is created when the `wlr_layer_surface_v` is created, in `handle_layer_shell_surface`. How can it be NULL?

tadeokondrak:
The `wlr_layer_surface_v1` is created and added to the `wlr_layer_shell_v1` surface list before `handle_layer_shell_surface` is called: https://github.com/swaywm/wlroots/blob/41f9916ae59e8fc06e4e00dd3fc9b57baf5ed537/types/wlr_layer_shell_v1.c#L437
It might be a better change for sway to maintain its own list of layer surfaces instead of using the list in `wlr_layer_shell_v1`.

emersion:
Oh, I see. Okay, this solution sounds reasonable. It's the only place where we iterate over `layer_shell->surfaces`.

Alternatively, we could iterate over `&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]`.

emersion:
Thanks!

PR #5073, opened 2020-03-05
Add contrib/keyboard-layout-per-window.py

DrPandemic:
Re-opening https://github.com/swaywm/sway/pull/3155 since I'm currently using it and I think could be useful to others. It adds a contrib script to keep track of active keyboard layouts per window.

I think it would be nice to have this feature directly in sway (https://github.com/swaywm/sway/pull/3155), but I don't have enough knowledge about sway's internals to do that PR.

This PR uses the code from the initial PR but tries to fix all the comments.

### What about `seats`
This script doesn't support them. I think the current state of the script make sense by itself even without seats.

I unplugged my usb-c hub containing, among other things, my keyboard. It kept the same ID even after being reconnected. This means that the script kept working as expected.

### DIFF
--- a/contrib/keyboard-layout-per-window.py
+++ b/contrib/keyboard-layout-per-window.py
@@ -37,7 +37,11 @@ def on_window(ipc: i3ipc.connection.Connection, event: i3ipc.events.WindowEvent)
 
 if __name__ == "__main__":
     ipc = i3ipc.Connection()
-    prev_focused = ipc.get_tree().find_focused().id
+    focused = ipc.get_tree().find_focused()
+    if focused:
+        prev_focused = focused.id
+    else:
+        prev_focused = None
     windows = {}
 
     ipc.on("window", on_window)
--- a/contrib/keyboard-layout-per-window.py
+++ b/contrib/keyboard-layout-per-window.py
@@ -7,11 +7,7 @@
 
 import i3ipc
 
-ipc = i3ipc.Connection()
-prev_focused = ipc.get_tree().find_focused().id
-windows = {}
-
-def on_window_focus(ipc, event):
+def on_window_focus(ipc: i3ipc.connection.Connection, event: i3ipc.events.WindowEvent):
     global windows, prev_focused
 
     # Save current layout
@@ -23,21 +19,26 @@ def on_window_focus(ipc, event):
     if event.container.id in windows:
         for (kdb_id, layout_index) in windows[event.container.id].items():
             if layout_index != layouts[kdb_id]:
-                ipc.command("""input "{}" xkb_switch_layout {}""".format(
-                    kdb_id, layout_index))
+                ipc.command(f"input \"{kdb_id}\" xkb_switch_layout {layout_index}")
+                break
 
     prev_focused = event.container.id
 
-def on_window_close(ipc, event):
+def on_window_close(ipc: i3ipc.connection.Connection, event: i3ipc.events.WindowEvent):
     global windows
     if event.container.id in windows:
         del(windows[event.container.id])
 
-def on_window(ipc, event):
+def on_window(ipc: i3ipc.connection.Connection, event: i3ipc.events.WindowEvent):
     if event.change == "focus":
         on_window_focus(ipc, event)
     elif event.change == "close":
         on_window_close(ipc, event)
 
-ipc.on("window", on_window)
-ipc.main()
+if __name__ == "__main__":
+    ipc = i3ipc.Connection()
+    prev_focused = ipc.get_tree().find_focused().id
+    windows = {}
+
+    ipc.on("window", on_window)
+    ipc.main()
--- a/contrib/keyboard-layout-per-window.py
+++ b/contrib/keyboard-layout-per-window.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+
+# This script keeps track of active keyboard layouts per window.
+#
+# This script requires i3ipc-python package (install it from a system package
+# manager or pip).
+
+import i3ipc
+
+ipc = i3ipc.Connection()
+prev_focused = ipc.get_tree().find_focused().id
+windows = {}
+
+def on_window_focus(ipc, event):
+    global windows, prev_focused
+
+    # Save current layout
+    layouts = {input.identifier : input.xkb_active_layout_index
+               for input in ipc.get_inputs()}
+    windows[prev_focused] = layouts
+
+    # Restore layout of the newly focused window
+    if event.container.id in windows:
+        for (kdb_id, layout_index) in windows[event.container.id].items():
+            if layout_index != layouts[kdb_id]:
+                ipc.command("""input "{}" xkb_switch_layout {}""".format(
+                    kdb_id, layout_index))
+
+    prev_focused = event.container.id
+
+def on_window_close(ipc, event):
+    global windows
+    if event.container.id in windows:
+        del(windows[event.container.id])
+
+def on_window(ipc, event):
+    if event.change == "focus":
+        on_window_focus(ipc, event)
+    elif event.change == "close":
+        on_window_close(ipc, event)
+
+ipc.on("window", on_window)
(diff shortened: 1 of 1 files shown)


vyachkonovalov:
Thank you for the script. From my perspective, it would be good to set default layout for new windows. Maybe optional, through an argument.

vvrein:
My version of script that can set default kb layout per app_id.

<details><summary>keyboard-layout-per-window.py</summary>


```python

#!/usr/bin/env python

# This script keeps track of active keyboard layouts per window.
#
# This script requires i3ipc-python package (install it from a system package
# manager or pip).

import i3ipc

ipc = i3ipc.Connection()

legal_inputs = ['1118:1957:Microsoft_Microsoft®_2.4GHz_Transceiver_v9.0',]
any_keyboard = False
app_ids_kb   = {"Alacritty": 0, "telegramdesktop": 1}
debug = 0


prev_focused = -1
windows_ids  = {}
#app_ids_kb  = {}



def on_window_focus(ipc, event):

    # Set specific vars
    global windows_ids, prev_focused, app_ids_kb, legal_inputs
    app_id = event.container.app_id
    window_id = event.container.id


    # Get inputs current state only for legal inputs
    inputs = []
    for input in ipc.get_inputs():
        if input.identifier in legal_inputs or any_keyboard:
            inputs.append(input)


    # Save current layout
    layouts = {}
    for input in inputs:
        layouts[input.identifier] = input.xkb_active_layout_index
    if debug:
        print("Layout saved                          app_id => {}; id => {}; layout => {}"
                    .format(app_id, window_id, input.xkb_active_layout_index))
    windows_ids[prev_focused] = layouts


    # Restore layout of the newly focused window
    if app_id in app_ids_kb:
        layout_index = app_ids_kb[app_id]
        for input in inputs:
            if input.xkb_active_layout_index != app_ids_kb[app_id]:
                input_id     = input.identifier
                ipc.command("input \"{}\" xkb_switch_layout {}".format(input_id, layout_index))
            if debug:
                print("Layout restored, cause: app_ids_kb,   app_id => {}; id => {}; layout => {};"
                                         .format(app_id, window_id, layout_index))
    elif window_id in windows_ids:
        for (input_id, layout_index) in windows_ids[window_id].items():
            if layout_index != layouts[input_id]:
                ipc.command("input \"{}\" xkb_switch_layout {}".format(input_id, layout_index))

            if debug:
                print("Layout restored, cause: windows_ids,  app_id => {}; id => {}; layout => {};"
                                         .format(app_id, window_id, layout_index))
    prev_focused = window_id
    if debug:
        print("END on_window_focus END\n")


def on_window_close(ipc, event):
    global windows
    if event.container.id in windows_ids:
        del(windows_ids[event.container.id])

def on_window(ipc, event):
    if event.change == "focus":
        on_window_focus(ipc, event)
    elif event.change == "close":
        on_window_close(ipc, event)

ipc.on("window", on_window)
ipc.main()

```

</details>

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
+
+# This script keeps track of active keyboard layouts per window.
+#
+# This script requires i3ipc-python package (install it from a system package
+# manager or pip).
+
+import i3ipc
+
+ipc = i3ipc.Connection()
```
Anakael, line 10: I suggest to move such things to bottom of script and wrap with __main__ check

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,43 @@
...
+prev_focused = ipc.get_tree().find_focused().id
+windows = {}
+
+def on_window_focus(ipc, event):
```
Anakael, line 14: How about type hints? Not necessary here, but nice python feature.

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,43 @@
...
+    global windows, prev_focused
```
Anakael, line 15: Use global is bad practice. Better explicitly pass through args.

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,43 @@
...
+    # Save current layout
+    layouts = {input.identifier : input.xkb_active_layout_index
+               for input in ipc.get_inputs()}
+    windows[prev_focused] = layouts
+
+    # Restore layout of the newly focused window
+    if event.container.id in windows:
+        for (kdb_id, layout_index) in windows[event.container.id].items():
+            if layout_index != layouts[kdb_id]:
+                ipc.command("""input "{}" xkb_switch_layout {}""".format(
```
Anakael, line 26: Why triple quotes here?
F-string with single quote would be more readable.

emersion:
It's just a single-file script, I don't think globals/__main__/types are important at all.

Anakael:
> It's just a single-file script, I don't think globals/**main**/types are important at all.

Not really important. Just suggest to implement it by pythonic-way.

contrib/keyboard-layout-per-window.py
DrPandemic: I agree with you, but here it's part of a callback function call chain. AFAIK, there's no clean way of passing those variables in the callback invocations. So, I think this is a good balance between being simple and readable.

DrPandemic:
## Modifications
- Added some type hinting, why not?
- Moved things to "main".
- Use f-strings, nice syntax.

## What I didn't do
- Letting the user change any kind of default layout. I didn't find any clean/simple way of being able to received a string like `ca-fr` and transform it into the right `xkb_layout`. Also, the first layout is seems to always be the default, so if you really want to change the default, change the order of your layouts. That's not a nice solution, but it should work.
- Stop using globals. Since the code runs in callback, it would be complicated to actually pass the other variables.

@emersion Do you know what I could do to help this PR get merged?

Anakael:
-> Letting the user change any kind of default layout. I didn't find any clean/simple way of being able to received a string like ca-fr and transform it into the right xkb_layout.

https://wiki.archlinux.org/index.php/Xorg/Keyboard_configuration
Look at .../base.lst and localectl commands.

contrib/keyboard-layout-per-window.py
```diff
@@ -0,0 +1,44 @@
...
+
+def on_window(ipc: i3ipc.connection.Connection, event: i3ipc.events.WindowEvent):
+    if event.change == "focus":
+        on_window_focus(ipc, event)
+    elif event.change == "close":
+        on_window_close(ipc, event)
+
+if __name__ == "__main__":
+    ipc = i3ipc.Connection()
+    prev_focused = ipc.get_tree().find_focused().id
```
gdamjan, line 40: this can fail if the script is run as soon as sway is started and there's nothing focused.

gdamjan, line 40:
```suggestion
    focused = ipc.get_tree().find_focused()
    if focused:
        prev_focused = focused.id
    else:
        prev_focused = None
```

ony (commented):
Looks good to me.

I agree that it is ok to use globals in a script not intended to be used as a library. Though in this case I would also drop detection if this is run as script or as a module via `__name__` as well.

> Stop using globals. Since the code runs in callback, it would be complicated to actually pass the other variables.

If dropping globals is wanted it easily achieved with something like:

```python
class WindowListener:
    def __init__(self, ipc);
        # ...
        ipc.on('window', self._on_window)

    def _on_window(self, ipc, event):
        # ...

    def _on_focus(self, event):
        # ...

# ...
    ipc = i3ipc.Connection()
    WindowListener(ipc)
    ipc.main()
```

gdamjan:
there's also https://github.com/artemsen/swaykbdd

DrPandemic:
I personally think this PR is in an acceptable state. I've been using this script for maybe a year (minus some small tweaks through the year) and it's been working fine.

@emersion what would it take to get it merged?

PR #5074, opened 2020-03-05
Only schedule a frame if client has requested a frame callback

emersion:
When a client hasn't damaged its surface, we only need to schedule an
output frame if the client has requested a frame callback.


ddevault:
Thanks!

PR #5075, opened 2020-03-05
common/log: log current time with milliseconds

emersion:
Same as [1].

[1]: https://github.com/swaywm/wlroots/pull/2052


emersion:
Updated to match the wlroots change.

ddevault:
Thanks!

Issue #4924, opened 2020-01-18
Sway does not handle subsurfaces that extend beyond the parent's bounding box very well (common in gtk applications)


tmccombs:
* Sway Version: 1.2

* Debug Log:

Relevant portion:

```
2020-01-18 02:15:28 - [sway/desktop/transaction.c:489] Transaction 0x558676e2c7f0 is ready
2020-01-18 02:15:28 - [sway/desktop/transaction.c:280] Applying transaction 0x558676e2c7f0
2020-01-18 02:15:32 - [types/seat/wlr_seat_pointer.c:362] button_count=1 grab_serial=0 serial=132
2020-01-18 02:15:32 - [sway/desktop/transaction.c:411] Transaction 0x558676e2c7f0 committing with 1 instructions
2020-01-18 02:15:32 - [sway/desktop/transaction.c:280] Applying transaction 0x558676e2c7f0
2020-01-18 02:15:32 - [types/seat/wlr_seat_pointer.c:362] button_count=0 grab_serial=132 serial=133
2020-01-18 02:15:38 - [sway/input/cursor.c:576] denying request to set cursor from unfocused client
2020-01-18 02:15:46 - [sway/ipc-server.c:334] Sending window::focus event
```

* Configuration File: Happens with default configuration

Steps to reproduce:

Run the popover example code found here: https://python-gtk-3-tutorial.readthedocs.io/en/latest/popover.html#menu-popover (reproduced here):

It is more obvious if you change the border width on line 46 (for the app window) to something smaller, like 10.

Once the window opens switch to floating mode (Mod+Shift+space in default config). Then click on the button to make the popover appear. Although the popover displays outside of the parent window, it doesn't seem to receive mouse events, and if focus_follows_mouse is enabled, as soon as you move over the popover, the parent window (and the popover itself) loses focus. Clicking also seems to send the click event to the window underneath the popover.

# Update

After more investigation I've determined this is because GTK uses subsurfaces, rather than popup surfaces or toplevel surfaces for menus and popovers. And sway seems to have some assumptions that subsurfaces fit inside the bounding box of the parent top-level surface. This results in the following undesirable behavior:

 - If a subsurface extends beyond the bounding box it does not receive cursor events
 - If a subsurface extends beyond the bounding box of a floating window, the border is drawn over the subsurface
 - If a subsurface extends beyond the bounding box of a tiling window, it is cut off by the bounding box.

emersion:
Can you try master?

[...]

PR #5082, opened 2020-03-09
Subsurface overlap

tmccombs:
I believe this mostly fixes problems with subsurfaces extending beyond the views bounding box as described in #4924. 

What this does:

- Handles input events for subsurfaces extending beyond the border for focused views that are floating (this was already working if the view was tiled and not floating)
- Render the border before the view, so that subsurfaces that overlap the border are drawn on top
- Render the active child of linear tiled containers last, so that subsurfaces are not covered up by other children while that child is active. 

What it doesn't fix:

- If a floating view has subsurfaces that extend beyond its bounding box but the view isn't focused, those subsurfaces will be rendered but won't receive input events. Honestly I think the best solution in this case might be to clip the view to its bounding box if it isn't focused, or something like that.
- Possibly other edge conditions?

I would welcome feedback on this PR.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -307,22 +307,8 @@ static void render_saved_view(struct sway_view *view,
 	// https://github.com/swaywm/sway/pull/4465#discussion_r321082059
 }
 
-/**
- * Render a view's surface and left/bottom/right borders.
- */
-static void render_view(struct sway_output *output, pixman_region32_t *damage,
+static void render_view_border(struct sway_output *output, pixman_region32_t *damage,
 		struct sway_container *con, struct border_colors *colors) {
-	struct sway_view *view = con->view;
-	if (view->saved_buffer) {
-		render_saved_view(view, output, damage, view->container->alpha);
-	} else if (view->surface) {
-		render_view_toplevels(view, output, damage, view->container->alpha);
-	}
-
-	if (con->current.border == B_NONE || con->current.border == B_CSD) {
-		return;
-	}
-
 	struct wlr_box box;
 	float output_scale = output->wlr_output->scale;
 	float color[4];
@@ -371,6 +357,25 @@ static void render_view(struct sway_output *output, pixman_region32_t *damage,
 		box.height = state->border_thickness;
 		scale_box(&box, output_scale);
 		render_rect(output, damage, &box, color);
+	}
+}
+
+/**
+ * Render a view's surface and left/bottom/right borders.
+ */
+static void render_view(struct sway_output *output, pixman_region32_t *damage,
+		struct sway_container *con, struct border_colors *colors) {
+	// Draw border first, so that if subsurfaces overlap the border is drawn under it
+	if (con->current.border != B_NONE && con->current.border != B_CSD) {
+		render_view_border(output, damage, con, colors);
+	}
+
+	struct sway_view *view = con->view;
+
+	if (view->saved_buffer) {
+		render_saved_view(view, output, damage, view->container->alpha);
+	} else if (view->surface) {
+		render_view_toplevels(view, output, damage, view->container->alpha);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


emersion:
I'm not sure about this. Sub-surfaces and the main surface are supposed to form a single layer, there's no guarantee that sub-surfaces spanning outside of the window geometry will be displayed. Sub-surfaces aren't meant to be used for menus/popups, instead xdg-popup should be used.

tmccombs:
>  there's no guarantee that sub-surfaces spanning outside of the window geometry will be displayed

Do you have a reference for this? According to the [freedesktop documentation for subsurfaces](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_subsurface), it should be:

> An additional interface to a wl_surface object, which has been made a sub-surface. A sub-surface has one parent surface. A sub-surface's size and position are not limited to that of the parent. Particularly, a sub-surface is not automatically clipped to its parent's area.

> Sub-surfaces aren't meant to be used for menus/popups, instead xdg-popup should be used.

Maybe, but Gtk definitely uses them this way, and I think Qt does as well (although I haven't dug into Qt's menus enough to know for sure). And I know that in the case of Gtk it was done very intentionally (although that was a while ago).

See https://gitlab.gnome.org/GNOME/gtk/-/commit/96f7c11f4840d7425ab19290ea18d3904ab939a1 and https://bugzilla.gnome.org/show_bug.cgi?id=759738

emersion:
>According to the freedesktop documentation for subsurfaces, it should be:

window geometry != surface geometry

The window geometry is an xdg-shell concept.

>Maybe, but Gtk definitely uses them this way, and I think Qt does as well (although I haven't dug into Qt's menus enough to know for sure). And I know that in the case of Gtk it was done very intentionally (although that was a while ago).

I'm not sure xdg-shell popups were stabilized in 2015.

I consider this a GTK bug. Maybe it's not a very visible bug on GNOME because there are no SSDs and no real tiling.

tmccombs:
> window geometry != surface geometry

I'm not sure what your point is. The documentation for wl_subsurface explicitly says it shouldn't be clipped, and I can't find anything in the documentation for xdg-shell that says subsurfaces should or can be clipped to the window geometry.  I could very well be missing something, but so far, sway is the only compositor I've found that has this behavior. 

> I consider this a GTK bug

It looks like Gtk 4 will use xdg-popups: https://gitlab.gnome.org/GNOME/gtk/-/issues/1097#note_756464. Although, this doesn't really help with existing Gtk 3 apps.

> Sub-surfaces and the main surface are supposed to form a single layer

With this PR they still are a single layer, it just makes sure that the active tiled window on a workspace is rendered on top, and floating windows receive events for overflowing subsurfaces (which are currently rendered).

Regardless of  whether this PR is the correct fix, the current behavior is very inconsistent. Whether overflowing subsurfaces are visible currently depends on if the window is floating, or where it is in a tiling layout. And even if it is visible, it will only accept mouse interaction if the view is tiled and no parent containers are floating.

emersion:
>The documentation for wl_subsurface explicitly says it shouldn't be clipped

It says that it shouldn't be clipped to the parent surface geometry. However there are many situations where the surface will be clipped if it goes outside of the window geometry. For instance if the sub-surface goes off-screen (clients don't know where the edges of the screen are).

tmccombs:
> However there are many situations where the surface will be clipped if it goes outside of the window geometry. For instance if the sub-surface goes off-screen 

Sure. I do think that going off screen is a little different than extending beyond the bounding box of the tile in  a tiling window manager. I can see the argument  there though. But in that case, I would expect it to _always_ clip to the bounding box of the tile, not the current behaviour where it clips it to the bottom and right, but not to the top or left. And I still think that for floating views at least, it should be allowed to draw outside of the bounding box.

emersion:
Closing because the clipping behaviour is getting implemented in https://github.com/swaywm/sway/pull/5451. I think this is a better approach.

tmccombs:
I'm still unclear. Is there a technical problem with this PR, or is it simply that you think that using subsurfaces for popups is wrong (even if the code was written before xdg-popup and the xdg-popup positioning existed), and therefore any app that does so should be broken in sway?

Issue #5093, opened 2020-03-11
Sway keeps adding the same tablet PAD device multiple times


ofourdan:
* Sway Version:

sway version 1.4-fafc0609 (Mar 11 2020, branch 'master')

* Debug Log:

[sway.log.gz](https://github.com/swaywm/sway/files/4318863/sway.log.gz)

* Configuration File:

Irrelevant

* Stack Trace, if sway crashes:

[sway-bt.txt.gz](https://github.com/swaywm/sway/files/4318882/sway-bt.txt.gz)

* Steps to reproduce:

https://gitlab.freedesktop.org/xorg/xserver/issues/995#note_433460

* Additional info

This was investigated as a part of an issue reported to Xwayland on instructions from sway developpers.

https://gitlab.freedesktop.org/xorg/xserver/issues/995

As it turns out, I reckon this is a bug with `sway` instead, and that following patch in `sway` fixes it:

[0001-input-Avoid-creating-the-PAD-device-multiple-times.patch.gz](https://github.com/swaywm/sway/files/4318898/0001-input-Avoid-creating-the-PAD-device-multiple-times.patch.gz)

That fixes both the crash in `sway` but also he crash in `Xwayland` and also the same pad devices being added several times on reload.

Xyene:
This seems like it might also be the root cause of https://github.com/swaywm/wlroots/issues/2062, since the reproduction steps are identical (though the backtraces differ).

emersion:
@Xyene Can you try out this patch?

Xyene:
It fixes the issue I filed against `Xwayland` (thanks!) on a real non-emulated device. I didn't have a very reproducible setup for the wlroots bug, but I wasn't able to reproduce it with this patch.

Firefox continues to crash if it has focus while the tablet is disconnected (matches the destroy events, I think), though this doesn't happen in GNOME. This sounds like it might be a related, but separate, issue.

PR #5097, opened 2020-03-12
input: Avoid creating the PAD device multiple times

emersion:
If a pad device for a tablet exists, reloading the configuration,
removing/reading the device or even suspending the system will recreate
the same Wayland input device multiple times.

Make sure we don't re-create the same Wayland device more than
necessary.

Signed-off-by: Olivier Fourdan <ofourdan@redhat.com>

Closes: https://github.com/swaywm/sway/issues/5093

Thanks @ofourdan for the patch!


ddevault:
Thanks!

PR #5098, opened 2020-03-12
Fix big workspace numbers overflow

### LINKED ISSUE #5078: Big workspace numbers overflow
Ref: https://github.com/i3/i3/pull/3950/commits/83c7aff089a6728b6e522d934d656a8e09463112

ammgws:
Fixes #5078, which also brings us inline with i3 behaviour.﻿

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -384,8 +384,19 @@ static json_object *ipc_json_describe_scratchpad_output(void) {
 
 static void ipc_json_describe_workspace(struct sway_workspace *workspace,
 		json_object *object) {
-	int num = isdigit(workspace->name[0]) ? atoi(workspace->name) : -1;
-
+	int num;
+	if (isdigit(workspace->name[0])) {
+		errno = 0;
+		char *endptr = NULL;
+		long long parsed_num = strtoll(workspace->name, &endptr, 10);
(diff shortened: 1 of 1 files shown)


sway/ipc-json.c
```diff
@@ -384,8 +384,19 @@ static json_object *ipc_json_describe_scratchpad_output(void) {
 
 static void ipc_json_describe_workspace(struct sway_workspace *workspace,
 		json_object *object) {
-	int num = isdigit(workspace->name[0]) ? atoi(workspace->name) : -1;
-
+	int num;
+	if isdigit(workspace->name[0]) {
```
RedSoxFan: There are missing parentheses around the condition. (gcc should not be accepting this...)

ammgws: @RedSoxFan added parentheses. Either way it built and ran so not sure how to catch this next time.

RedSoxFan: Apparently gcc implements isdigit as a macro that starts with a paren, which is why it was accepting it. The FreeBSD CI build (https://builds.sr.ht/~sircmpwn/job/167521), which uses clang, did fail and showed the error. Anyway, thanks for fixing it.

RedSoxFan:
Thanks!

Issue #4496, opened 2019-08-25
swaybar shows multiple status lines at once


shibe2:
* Sway Version:

sway version 1.2-rc2 (Aug 23 2019, Arch Linux)

* Configuration File:

```
bar {
    status_command tail -f ~/sway-status.txt
}
```

When status command initially immediately outputs multiple lines, they all are shown on the swaybar. When after that the command outputs more lines, only the last line is shown.

Swaybar should always display only the last status line.

To reproduce, create a file `~/sway-status.txt` with multiple short lines of text. Set status_command to output that file:

    swaymsg bar bar-0 status_command "tail -f ~/sway-status.txt"

samerickson:
I cannot seem to reproduce your issue. 
[screenshot of what I managed to reproduce](https://imgur.com/gallery/NRAVIIo)

Also, could we get a little bit more information regarding what you are trying to achieve? It seems a little unorthodox to be creating a status bar in this way (running tail on a text file).

shibe2:
@samerickson Your screenshot shows exactly the issue. You should only see the last line, "sizth line" as the status.

### What I'm trying to achieve

Status command may want to output information as soon as it's available. If shortly after that another piece of information is available, it will output another line. Normally, it works fine. Only if multiple updates happened during the bar startup, both old and new status appear together.

As an example, my program is affected. https://github.com/shibe2/sklt

samerickson:
I misunderstood the issue, I thought you were having multiple status bars. Status bars will display newline characters inline. This is useful if you want something like `lsblk` to be displayed, which is normally a block of text, but in a status bar, you need that information to be in a single line. 

In your case, if you are trying to only display the last line of the file, you need to change your tail command. Tail defaults to showing the last 10 lines of a file, to make it display **Only** the last line, use: `tail -n 1`. 

As for getting your project to function correctly, I have created an issue https://github.com/shibe2/sklt/issues/1, as I cannot seem to make your project function.

samerickson:
You can change your bar config to link to a script containing the following:
```
#!/bin/bash
while true;
    printf "%s %s\n" "$(date +%D)" "$(date +%r)"
    sleep 1
done
```
As for getting it to display the current keyboard layout, you can have a look at this project as well:
https://gitlab.com/racy/sway-keyboard-layout/blob/master/sway-keyboard-layout. This makes your code footprint a lot smaller, but I am unsure of the resource consumption differences.

samerickson:
[...]

PR #5103, opened 2020-03-14
swaybar: only show last status line

ammgws:
Solves #4496

### DIFF
--- a/swaybar/status_line.c
+++ b/swaybar/status_line.c
@@ -113,6 +113,11 @@ bool status_handle_readable(struct status_line *status) {
 
 		sway_log(SWAY_DEBUG, "Using text protocol.");
 		status->protocol = PROTOCOL_TEXT;
+		char *last_newline = strrchr(status->buffer, '\n');
+		if (last_newline != NULL) {
+			status->buffer_index = strlen(last_newline + 1);
+			memmove(status->buffer, last_newline + 1, status->buffer_index + 1);
+		}
 		status->text = status->buffer;
 		// intentional fall-through
 	case PROTOCOL_TEXT:
--- a/swaybar/status_line.c
+++ b/swaybar/status_line.c
@@ -113,19 +113,26 @@ bool status_handle_readable(struct status_line *status) {
 
 		sway_log(SWAY_DEBUG, "Using text protocol.");
 		status->protocol = PROTOCOL_TEXT;
-		char *last_newline = strrchr(status->buffer, '\n');
-		if (last_newline != NULL) {
-			status->buffer_index = strlen(last_newline + 1);
-			memmove(status->buffer, last_newline + 1, status->buffer_index + 1);
-		}
 		status->text = status->buffer;
 		// intentional fall-through
 	case PROTOCOL_TEXT:
 		errno = 0;
 		while (true) {
-			if (status->buffer[read_bytes - 1] == '\n') {
-				status->buffer[read_bytes - 1] = '\0';
+			bool trailing_newline = true;
+			while (trailing_newline == true) {
+				if (status->buffer[read_bytes - 1] == '\n') {
+					status->buffer[read_bytes - 1] = '\0';
+					read_bytes--;
+				} else {
+					trailing_newline = false;
+				}
 			}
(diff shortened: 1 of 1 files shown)


ianyfan:
Have you tested whether this will work if the last line ends with a new-line character?

ammgws:
I did before and after these changes, and both output just a new line character. I wasnt sure whether to go further and make it ignore the last line if it's a new line character

ianyfan:
I mean: if you send `first line\nsecond line\n` what does this output?

ammgws:
From memory it would just output a newline, so the status line would be empty. I will check again once #5106 is fixed.

ammgws:
In sway config: status_command tail -f /tmp/1
`printf "XXX" >/tmp/1 from my terminal to output to the file, and then running swaybar (killing swaybar in between runs):

| XXX  | bar output on startup |
| ------------- | ------------- |
| aaa | aaa  |
| aaa\n  | (empty)  |
| aaa\nbbb  | bbb  |
| aaa\nbbb\n  | (empty)  |

However the second time `aaa\nbbb\n` or `aaa\n` is sent, it displays the last value.

Hmm

ianyfan:
I think that it should show the line before the `\n` character. However, i3bar seems to try to display all the lines if you send it multiple lines at once, but I'm not sure if we should deviate from its behaviour. We currently, do, since it only the last line is shown, but no-one seems to have complained so far.

FreeFull:
It is very common to produce output from shell scripts using `echo`, which by default appends a newline at the end.

As far as i3bar goes, maybe it'd be a good idea to ask the i3 developers if its behaviour is intentional.

ammgws:
Yeah if no one else does then I will try find some time to install i3 and verify the existing behaviour before asking the guys over there.

ammgws:
Testing so far

| XXX  | sway | i3 |
| ------------- | ------------- | ------------- |
| aaa | aaa  | aaa |
| aaa\n  | (empty)  | aaa |
| aaa\nbbb  | bbb | aaa\nbbb (newline rendered, bbb is partially cutoff) |
| aaa\nbbb\n  | (empty)  | aaa\nbbb (newline rendered, bbb is partially cutoff) |

ammgws:
OK so i3 has confirmed it's a bug, so we can go ahead with our changes.

`printf "XXX" >/tmp/1` and start i3/sway with bar config of `status_command tail -f /tmp/1`:

| XXX  | i3 | sway | sway (this PR) |
| ------------- | ------------- | ------------- | ---------- |
| aaa | aaa | aaa | aaa |
| aaa\n | aaa | aaa | aaa |
| aaa\nbbb | aaa<br/>bbb | aaa⏎bbb | bbb |
| aaa\nbbb\n | aaa<br/>bbb | aaa⏎bbb | bbb |

`printf "XXX" >>/tmp/1` after i3/swaybar is already running:
| XXX  | i3 | sway | sway (this PR) |
| ------------- | ------------- | ------------- | ---------|
| aaa | aaa | aaa | aaa |
| aaa\n | aaa<br/>| aaa | aaa |
| aaa\nbbb | aaa<br/>bbb| bbb | bbb |
| aaa\nbbb\n | aaa<br/>bbb | bbb | bbb |


One final thing that may be a problem:
On startup, `aaa\nbbb\n\n` (line ending with two or more newline chars), the bar will output `bbb`, however if sent again it will output a blank line (probable `\n`)
Can't figure out out where it's coming from yet...

ammgws:
I think I'll need to modify the behaviour of the code around `getline` to deal with multiple newline characters, then this should be complete.

ianyfan:
I'd prefer to get changes merged into i3 first so that its behaviour can be confirmed (I thin we can afford to wait since this seems to be fairly low priority).

Also: could you see what the intended behaviour is if you first send an incomplete line and then complete it later? e.g. if you send `aaa` followed by `bbb\n` should it complete the line (show `aaabbb`) or start a new line (show `bbb`)

Thanks.

ammgws:
Good points. Code should be more or less the same so I'll take it over to i3 first.

ammgws:
Closing as don't have time to work on this

Feel free to pick it up

PR #5108, opened 2020-03-14
input/cursor: correctly transfer focus when using tablet pen

### LINKED ISSUE #4819: Focus does not follow mouse when using tablet (among other misbehaviors)
Version 1.2-aa8fe584 (Dec 17 2019 'master')
Default config file

When `focus_follows_mouse` is set, a window is expected to get focus when hovered by the cursor. It works when moving the cursor with a regular touchpad or mouse. It does not when using the pen of a **Wacom Intuos S Tablet**. The cursor moves over the unfocused window but its focus status is unchanged (or it happens after a delay depending on the application). Both native and xwayland clients are affected. All the input device are attached to the same default seat.

# Steps

1. Open a nautilus client using dmenu
2. Press CTRL+n to open another. The new client, on the right, is focused.
3. Approach the pen close to the right half of the tablet: the cursor spawns on the already focused window.
4. Translate the pen to the left half: the cursor moves over the client on the left but it remains unfocused (undesired behavior).

Here is the debug trace: https://gist.github.com/florentc/ad8655f13207be7b6626fca5b8b9b379

* When using a mouse or touchpad, the client is immediately focused
* Sometimes, the hovered client gets focused but after a small delay whereas it is instantaneous using a mouse or touchpad

# Additional remarks

I have noticed several other misbehaviors I suspect share the same cause:
* in Chromium (xwayland) the shape of the cursor rarely changes to reflect links or text fields when moved using the pen
* in pcmanfm (gtk3) the right click menu opens where the cursor used to be the last time it was moved with a regular mouse instead of where it appears to be after being moved using the pen
* in many gtk3 applications (including nautilus), the cursor may flicker when moved using the pen and right click menus (when triggered using the pen) might fail to open
* borderless gtk3 floating windows with client side decoration cannot be dragged by the pen: they stay in place
* floating windows with sway titlebars can be dragged using the pen but the window does not follow the mouse during the movement (like when moving them using a regular mouse or touchpad), it gets displaced instantly once the pen is lifted

Shall I submit a new issue for each one of the above?

It is **as if the cursor moved using the tablet were not the same as the cursor moved by classic devices** although all the input devices are attached to the same seat.

I have detailed this issue because it is the easiest to reproduce among a bunch of misbehaviors that make it hard for now to go full tablet and get rid of regular pointing devices, which is impractical.

I am willing to provide additional information or perform other experiments if needed.

Xyene:
Tentatively fixes #4819.

This commit ensures that `seatop_motion` is called to transfer focus when a window is selected via a pen. Previously, it would race with `node_at_coords`, and only properly transfer focus if its returned `surface` was `NULL`.

I've tested this locally, and it fixes the issues I was experiencing.

Calling `seatop_motion` directly here works as well, but my impression is that the confine handling inside `cursor_motion` should be done for tablet devices too. Applications see the following sequence of events:

```
[912231.664] wl_pointer@3.motion(7987491, 281.539062, 991.007812)
[912231.673] zwp_tablet_tool_v2@4278190083.motion(281.539062, 991.007812)
[912231.676] zwp_tablet_tool_v2@4278190083.frame(7987491)
```

In a GNOME session on the same application, the `wl_pointer` event is not present. I don't know enough to say which is the correct behaviour, or if it even matters. Sway infrequently omits the `wl_pointer` event, too.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -515,8 +515,8 @@ static void handle_tablet_tool_position(struct sway_cursor *cursor,
 	double sx, sy;
 	struct wlr_surface *surface = NULL;
 	struct sway_seat *seat = cursor->seat;
-	node_at_coords(seat, cursor->cursor->x, cursor->cursor->y,
-		&surface, &sx, &sy);
+	struct sway_node *focused_node = node_at_coords(seat, cursor->cursor->x,
+		cursor->cursor->y, &surface, &sx, &sy);
 	struct sway_tablet_tool *sway_tool = tool->data;
 
 	if (!surface || !wlr_surface_accepts_tablet_v2(tablet->tablet_v2, surface)) {
@@ -527,6 +527,10 @@ static void handle_tablet_tool_position(struct sway_cursor *cursor,
 
 	wlr_tablet_v2_tablet_tool_notify_proximity_in(sway_tool->tablet_v2_tool,
 		tablet->tablet_v2, surface);
+
+	if (focused_node && config->focus_follows_mouse != FOLLOWS_NO) {
+		seat_set_focus(seat, focused_node);
+	}
 
 	wlr_tablet_v2_tablet_tool_notify_motion(sway_tool->tablet_v2_tool, sx, sy);
 }


emersion:
>Calling seatop_motion directly here works as well, but my impression is that the confine handling inside cursor_motion should be done for tablet devices too.

Yes. The `wl_pointer.motion` event is incorrect.

Xyene:
Thanks! So, to confirm:

* We don't want `wlr_relative_pointer_manager_v1_send_relative_motion` to be called for tablet events.
* We _do_ want `wlr_region_confine` to be called.
* We don't want `wlr_cursor_move` to be called. (Is this what triggers `wl_pointer.motion`?)

In that case, a follow-up question is: is https://github.com/swaywm/sway/blob/fcd524bb0d6f3c6cd4f2ddc180520bc56a89aa4a/sway/input/cursor.c#L481-L484 correct? The branch is taken even if the surface supports tablet v2, if the `surface` returned by `node_at_coords` is `NULL`, and ends up sending `wl_pointer.motion`. That branch is how tablet focus seems to work right now (even if erratically).

emersion:
>and ends up sending wl_pointer.motion

Just like touch, we should use `seat_set_focus` to transfer focus.

Xyene:
I've updated this PR to call `seat_set_focus`. I also added a check for `focus_follows_mouse`, which while not semantically correct (perhaps a new option should be better), matches i3 behavior.

PR #5110, opened 2020-03-15
sway/input/cursor.c: move NULL check to proper location

ammgws:
This check appears a couple times in the code, but it was out of place for one of them.


ddevault:
Thanks!

Issue #5106, opened 2020-03-14
Segfault on startup


ammgws:
sway version 1.4-7dd9cd29 (Mar 14 2020, branch 'makepkg')

[sway --config /etc/sway/config --debug 2&> /tmp/swaydebuglog](https://gist.github.com/ammgws/563e779108bb70b9affe48c97b99479f)

[backtrace](https://gist.github.com/d686def74f3219e9fbd117bf10a68b98)

have downgraded to sway 1.4 release for now

TheConfuZzledDude:
I've been getting the same problem. Interestingly, when run from a TTY it doesn't always segfault, it just crashes silently

[sway -c /etc/sway/config -d 2&> sway.log](https://gist.github.com/TheConfuZzledDude/91de2553d8a90f919faaf4860c65d493)

Also, reverting to 1.4 doesn't work for me, it launches, and I can see the cursor, and programs are clearly running (redshift works for example, and focus changes when I open a window in the scratchpad), but both screens are completely grey.
[The same, but on 1.4](https://gist.github.com/TheConfuZzledDude/6fec75925ceb934546a2b5008d0b2eed)

I'm going to assume that it's probably a graphics driver issue?

ammgws:
> I'm going to assume that it's probably a graphics driver issue?

Not sure since my last driver update was around a week before this (opencl-amd), and the ones before that (mesa stuff) were in February. I thought that maybe I could find some recent commits in sway/wlroots that might cause this but haven't seen anything that looks relevant.

TheConfuZzledDude:
Well we both have AMD GPUs, and this only started happening to me after updating Linux to 5.8, so it could be a kernel issue. I'll try downgrading to 5.7 or something and see

TheConfuZzledDude:
Actually, updating to the latest wlroots-git and sway-git seems to have fixed this for me

ammgws:
Yeah it's fixed for me now too.

Things that changed in between:

**sway related**
wlroots-git: `0.10.0.r48.g1282c3b1-1` `to 0.10.0.r49.g5ee52a3a-1`
Just [this](https://github.com/swaywm/wlroots/commit/5ee52a3ab9a12adb157738dc3fba72137c3c6879), but don't think it's relevant

sway-git: `r6258.55016729-1` to `r6261.6498542b-1`
Only new commits are from me and they're irrelevant

swaybg-git, swayidle-git, swaylock-git unchanged

**other**
[...]

PR #5111, opened 2020-03-15
sway/input/cursor.c: fix undefined behaviour (printing NULL)

ammgws:
Since `libevdev_event_code_get_name` may return NULL

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -1059,11 +1059,11 @@ uint32_t get_mouse_bindcode(const char *name, char **error) {
 	const char *event = libevdev_event_code_get_name(EV_KEY, code);
 	if (!event || strncmp(event, "BTN_", strlen("BTN_")) != 0) {
 		size_t len = snprintf(NULL, 0, "Event code %d (%s) is not a button",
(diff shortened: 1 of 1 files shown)


ammgws:
Hmm, I'll come back to this later

sway/input/cursor.c
```diff
@@ -1058,12 +1058,18 @@ uint32_t get_mouse_bindcode(const char *name, char **error) {
...
 	if (!event || strncmp(event, "BTN_", strlen("BTN_")) != 0) {
+		char *_event;
+		if (!event) {
+			_event = strdup("null");
+		} else {
+			_event = strdup(event);
+		}
 		size_t len = snprintf(NULL, 0, "Event code %d (%s) is not a button",
-				code, event) + 1;
+				code, _event) + 1;
```
ddevault: Can you just use a ternary expression? `event ? event : "(null)"`

ammgws: That is way nicer, thanks! I added a commit but I cannot test compilation until #5106 is fixed.

ddevault:
Thanks!

PR #5088, opened 2020-03-09
Update for new wlr_buffer API

emersion:
Update for https://github.com/swaywm/wlroots/pull/2044

PR #5113, opened 2020-03-15
Add support for linux-explicit-synchronization-unstable-v1

emersion:
Add support for linux-explicit-synchronization-unstable-v1. See https://github.com/swaywm/wlroots/pull/2070

Depends on https://github.com/swaywm/sway/pull/5088

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -57,6 +57,8 @@ struct sway_output {
 	uint32_t refresh_nsec;
 	int max_render_time; // In milliseconds
 	struct wl_event_source *repaint_timer;
+
+	list_t *textured_buffers;
 };
 
 struct sway_output *output_create(struct wlr_output *wlr_output);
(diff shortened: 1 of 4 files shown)


Sunderland93:
What does this protocol do? and what are the benefits to the user?

Emantor:
> What does this protocol do? and what are the benefits to the user?

Refer to the [protocol description](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/master/unstable/linux-explicit-synchronization/linux-explicit-synchronization-unstable-v1.xml) for details.

ascent12:
@Sunderland93 This is a "infrastructure" type of protocol, so we can support some upstream graphics things like Vulkan want. Mesa is expected to be the primary clients for this, and probably not a lot else.
End users and other clients wouldn't notice or really care.

PR #5124, opened 2020-03-18
Keyboard shortcuts inhibitors criteria

michaelweiser:
Introduction of keyboard shortcuts inhibitors in #5021 postponed two areas of functionality:

1. Per-client configurability of default activation of new inhibitors, particularly through criteria.
2. Visual feedback to users.

Here are some attempts to address those:

Add a separate per-view `shortcuts_inhibitor` command that can be used with criteria to override the per-seat defaults. This allows to e.g. disable shortcuts inhibiting globally but enable it for specific, known-good virtualization and remote desktop software or, alternatively, to blacklist that one slightly broken piece of software that just doesn't seem to get it right but insists on trying.

Allow re-execution of criteria. This allows its action to be re-executed once the criterion once again matches the view. Add a criteria selector that allows to match presence and state of keyboard shortcuts inhibitors on views. This allows to give visual feedback to the user on the state of an inhibitor through e.g.:
```
for_window [shortcuts_inhibitor=inactive] title_format !-%title
for_window [shortcuts_inhibitor=active] title_format !+%title
for_window  shortcuts_inhibitor=present] title_format !%title
for_window  shortcuts_inhibitor=absent] title_format %title
```

Add additional criteria execution call sites right after keyboard shortcuts inhibitor state changes. This allows to immediately and efficiently visually reflect inhibitor state changes to the user (or trigger any other desired action).

Add an additional criteria selector for idle inhibitors to showcase the functionality (and because I want to be able to visually tell that a client will prevent screen lock from coming on).

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -1176,3 +1176,45 @@ bool view_is_transient_for(struct sway_view *child,
 	return child->impl->is_transient_for &&
 		child->impl->is_transient_for(child, ancestor);
 }
+
+struct view_surface_iterator_context {
+	struct sway_view *view;
+	struct wlr_surface *wlr_surface;
+};
+
+static void view_criteria_execution_surface_iterator(
+		struct wlr_surface *surface, int sx, int sy, void *data) {
+	struct view_surface_iterator_context *context = data;
+
+	if (surface != context->wlr_surface) {
+		return;
+	}
+
+	view_execute_criteria(context->view);
+}
+
+static void view_criteria_execution_container_iterator(
+		struct sway_container *container, void *data) {
+	if (container->view) {
+		struct view_surface_iterator_context context = {
+			.view = container->view,
+			.wlr_surface = data,
+		};
+
+		view_for_each_surface(container->view,
+				view_criteria_execution_surface_iterator,
+				&context);
+	}
+}
+
+static void view_execute_criteria_from_wlr_surface(void *data) {
+	// here we intentionally go looking for a surface matching our argument
+	// because that surface might have since been freed
+	root_for_each_container(view_criteria_execution_container_iterator, data);
+}
+
+void view_schedule_criteria_execution_from_wlr_surface(
+		struct wlr_surface *wlr_surface) {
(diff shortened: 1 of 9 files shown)


sway/tree/view.c
```diff
@@ -437,7 +437,9 @@ void view_execute_criteria(struct sway_view *view) {
 		}
 		sway_log(SWAY_DEBUG, "for_window '%s' matches view %p, cmd: '%s'",
 				criteria->raw, view, criteria->cmdlist);
-		list_add(view->executed_criteria, criteria);
+		if (!criteria->continuous) {
```
michaelweiser: Here we might need some mechanism to prevent flapping, i.e. one continuous criterion matching and executing an action that makes another continuous criterion match and execute an action that in turn makes the first criterion match again and so on.

RedSoxFan:
Sorry for the delayed review, but can you please split this into one PR per commit. The commits appear to be separated logically, but each of these commits really should have it's own PR instead of being grouped together. If any of the commits depend on changes in another, you can mark them as a draft, which can be rebased and marked ready for review after the PR(s) it depends on is/are merged.

michaelweiser:
The per-view command and criteria reexecution are self-contained and now at #5322 and #5323, respectively, for individual review. The per-view command is touched again when adding additional call-sites for continuous criteria updates here. Conversely, this PR's changes depend on criteria rexecution to be useful. So I'll need to keep and update them here following their review as to not break this PR. I'll see about the draft status.

michaelweiser:
Sorry, I had a slight change of mind: The idle inhibitor criterion is now as #5324 and references criteria reexecution for usefulness. Dropped the idle inhibitor criterion and criteria reexecution here since they are self-contain in that regard. Some merging will now be required to see the functionality in its full splendour though, and the keyboard shortcuts and idle inhibitor criteria will have a merge conflict. See https://github.com/michaelweiser/sway/tree/ksi-criteria-save for a full-fledged experience. :)

michaelweiser:
Useless without #5323.

PR #5126, opened 2020-03-21
change apply_exclusive() to match layer shell protocol

Leon-Plickat:
As briefly discussed in IRC.

Sway currently respect the the positive exclusive zone of surfaces anchored
to three or four edges. With these changes, sway will respect positive
exclusive zones of layer surfaces anchored to one or three edges.

This matches the protocol, which states that a positive exclusive zone
should be respected, "if the surface is anchored to one edge or an
edge and both perpendicular edges". If the surfaces is "anchored to
only two perpendicular edges (a corner), anchored to only two
parallel edges or anchored to all edges a positive value will be
treated the same as zero".

---

I am not entirely sure if this is the best way to do this, but I think two anchor-cases to compare to per struct and a single struct per edge is clearer than either complex bit logic or having two structs per edge with a single anchor case each. However, I am not happy about the names `singular_anchor` and `anchor_triplet`.

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -25,13 +25,16 @@ static void apply_exclusive(struct wlr_box *usable_area,
 		return;
 	}
 	struct {
-		uint32_t anchors;
+		uint32_t singular_anchor;
+		uint32_t anchor_triplet;
 		int *positive_axis;
 		int *negative_axis;
 		int margin;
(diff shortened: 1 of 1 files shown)


wmww:
This bug had been mildly annoying me for a while, but not enough to dig into. I just noticed it got fixed. Nice work @Leon-Plickat!

PR #5127, opened 2020-03-21
Fix output mode usage

emersion:
The "Hz" part is required.


montvid:
C'mon lets drop the Hz part as there is no Mhz or anything else. :P

ddevault:
Thanks!

montvid:
How am i off topic? I am the one who raised this issue.
Instead of fixing the code documentation was fixed. Who on earth would like to write Hz - it is obvious as it gets it should not be mandatory. Xorg does not require that.

PR #5130, opened 2020-03-23
Fix redundant call to seat_send_focus()

Leo1003:
This commit fixes calling to `seat_send_focus()` twice when a view is closed.
This PR basically revert #2580.
I have tested that the original issue no longer exists. Closing a fullscreen view can successful focus to another view even after reverting.


ddevault:
Is there a related ticket that describes a problem this fixes?

Leo1003:
I found this problem during debugging #4740

ddevault:
Cool, thanks!

PR #5132, opened 2020-03-23
add --no-repeat option for bindings

### LINKED ISSUE #3151: [Feature Request] bindsym --no-repeat
It would be great to disable the key-repeat of `bindsym`'s

lheckemann:
This allows e.g. triggering one command while a key is held, then
triggering another to undo the change performed by it afterwards. One
use case for this is triggering push-to-talk functionality for VoIP
tools without granting them full access to all input events.

Fixes #3151

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -41,6 +41,7 @@ enum binding_flags {
 	BINDING_CODE=32,     // keyboard only; convert keysyms into keycodes
 	BINDING_RELOAD=64,   // switch only; (re)trigger binding on reload
 	BINDING_INHIBITED=128,	// keyboard only: ignore shortcut inhibitor
+	BINDING_NOREPEAT=256, // keyboard only; do not trigger when repeating a held key
 };
 
 /**
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -373,6 +373,8 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 			binding->input = strdup(argv[0] + strlen("--input-device="));
 		} else if (strcmp("--no-warn", argv[0]) == 0) {
 			warn = false;
+		} else if (strcmp("--no-repeat", argv[0]) == 0) {
+			binding->flags |= BINDING_NOREPEAT;
 		} else {
 			break;
 		}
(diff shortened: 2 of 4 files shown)


mweinelt:
Tested against Sway1 .4 and works very well, thank you. This is scratching a big itch right now.

mweinelt:
@lheckemann Small issue I just came across, hope it's fine to talk about it here vs. opening an issue.

When I change to another desktop while having such a key pressed the release event does not get triggered, when I release the key. 

Specifically for mumble that means:
- From that point on PTT is enabled in mumble and won't disable on key release.
- PTT stays enabled until I press "Apply" its audio input setting
- after which the key starts working again

```
bindsym --no-repeat Scroll_lock exec /nix/store/5fm0p77lx0c7dw8k7lhzb8hjynbfch6a-dbus-1.12.16/bin/dbus-send --session --type=method_call --dest=net.sourceforge.mumble.mumble / net.sourceforge.mumble.Mumble.startTalking
bindsym --release Scroll_Lock exec /nix/store/5fm0p77lx0c7dw8k7lhzb8hjynbfch6a-dbus-1.12.16/bin/dbus-send --session --type=method_call --dest=net.sourceforge.mumble.mumble / net.sourceforge.mumble.Mumble.stopTalking
```

lheckemann:
@mweinelt aah, it seems that a release binding is triggered only if no other bindings were triggered after the initial press of the key — maybe someone else here knows why? That's pretty much independent of this change AFAIUI.

https://github.com/swaywm/sway/blob/dffc184a68c2d6839d4e601be141fe1adcd385aa/sway/input/keyboard.c#L400-L401

lheckemann:
@mweinelt right, I think I've got it.
- In sway: if another binding is executed after the press, the release binding isn't activated, so the `stopTalk` call is never executed
- In mumble: pushing any push-to-talk button _increments a counter_, which is decremented when the button is released — and transmission only stops when the counter reaches 0. This might be to support multiple people on the same mic, each with a keyboard/push-to-talk button? I'm a bit uncertain on that.

In any case, the right solution, I think, would be to trigger release bindings regardless of what happens between pressing and releasing… but maybe there's a good reason why this doesn't already happen.

PR #5143, opened 2020-03-26
Add focus following for touch_down

gg-rewrite:
The issue is similar to #4819: while the application windows correctly handle touch events when tapped, the keyboard input focus stays on the window the mouse is hovered upon. The solution is also similar, call move_cursor with relative coordinates.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -368,6 +368,12 @@ static void handle_touch_down(struct wl_listener *listener, void *data) {
 				event->touch_id, sx, sy);
 		cursor_hide(cursor);
 	}
+
+	//move cursor so focus follows touch
+	float delta_x = lx - cursor->cursor->x;
+	float delta_y = ly - cursor->cursor->y;
+	cursor_motion(cursor, 0, event->device, delta_x,
+		      delta_y, delta_x, delta_y);
 }
 
 static void handle_touch_up(struct wl_listener *listener, void *data) {
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -352,7 +352,7 @@ static void handle_touch_down(struct wl_listener *listener, void *data) {
 	wlr_cursor_absolute_to_layout_coords(cursor->cursor, event->device,
 			event->x, event->y, &lx, &ly);
 	double sx, sy;
-	node_at_coords(seat, lx, ly, &surface, &sx, &sy);
+	struct sway_node *focused_node = node_at_coords(seat, lx, ly, &surface, &sx, &sy);
 
 	seat->touch_id = event->touch_id;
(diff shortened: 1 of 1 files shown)


emersion:
Hmm, I don't think touch should move the cursor at all. With multi-touch, touch input is completely separate from pointer input.

gg-rewrite:
> Hmm, I don't think touch should move the cursor at all. With multi-touch, touch input is completely separate from pointer input.

As far as I see in the contents of cursor_motion there's no specific method to transfer the input focus, or at least I can't find it. If possible, can you point me to it so I could make corrections? I'll keep looking in the meantime.

Xyene:
@gg-rewrite from my work on https://github.com/swaywm/sway/pull/5108 (which I can unfortunately no longer test, since I no longer have the tablet in question), the actual focus transferring happens [on calling `seatop_motion`](https://github.com/swaywm/sway/blob/ac5ffdaf4890333474e5dd4fe6b9271e3b824d5b/sway/input/cursor.c#L263) (which happens in `cursor_motion`).

However, it looks like `seatop_motion`, through some levels of abstraction, ends up firing off a `wl_pointer.motion` event, which shouldn't happen for tablet (and I imagine touch) input.

@emersion I thought this would be something standard, so I went and tried on my devices, but it looks like different OSes do things in different ways.

* In Chrome OS, on touch the real cursor is hidden, and only reappears (in its original position) the next time a touchpad event is received.

* In Windows 10, touch down hides the cursor, but does not move it. Touch _move_ moves the real cursor (while still hidden), *unless* there are currently multiple touch points active, in which case the cursor does not move. The cursor reappears on a touchpad event.

The proposed behaviour sounds most similar to that from Chrome OS (and seems easiest to implement). Hiding the cursor would be a nice-to-have, though.

gg-rewrite:
Almost. The correct answer seems to be using seat_set_focus (*5 minutes of headdesking skipped*) with the node provided by node_at_coords.

This way the cursor behaves the same as in Chrome OS (as in hiding when touch event happends and reappearing at the old position after mouse events).

Xyene:
Nice, that does seem to be what

https://github.com/swaywm/sway/blob/2b51c5b2942fbedfe34a912b47616e98cfcea02e/sway/input/seatop_default.c#L475-L480

is doing as well.

One possible remaining concern here is that I think the code as written will ignore the value of `focus_follows_mouse` in config. This may or may not match `i3` behaviour (I can't test this, though, my touchscreen machines don't run Linux). I don't see any special handling of touch events in their code, so I'd guess they treat touch input as a "mouse" for the purposes of `focus_follows_mouse`.

ddevault:
This looks like the right solution, thanks a lot!

PR #5148, opened 2020-03-27
Fix hit test for floaters that cross outputs

### LINKED ISSUE #2315: Floating z-index is inverted
1. Open two floating windows
2. Focus is messed up

joshua-shone:
In the case of multiple overlapping floating windows, this commit fixes an issue where the wrong window would be focused in response to a cursor if one of the windows came from a different output (overhanging).

cross-output_floater_hittest_bug

This fix is similar to 8789ceea8 (fixes #2315) which switched to reverse iteration on the floating list, whereas this fix does it on the output list.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -289,7 +289,10 @@ static struct sway_container *container_at_linear(struct sway_node *parent,
 
 static struct sway_container *floating_container_at(double lx, double ly,
 		struct wlr_surface **surface, double *sx, double *sy) {
-	for (int i = 0; i < root->outputs->length; ++i) {
+	// For outputs with floating containers that overhang the output bounds,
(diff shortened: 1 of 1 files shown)


emersion:
Hmm, I don't think our output list is ordered at all. Am I missing something?

joshua-shone:
@emersion yes, this issue does appear to be a consequence of [render_floating](https://github.com/swaywm/sway/blob/854497b242b78f14f990a04d2c1fed14b03ee33b/sway/desktop/render.c#L934) which iterates over outputs in ascending order, like this:

```c
for (int i = 0; i < root->outputs->length; ++i) {
    [...]
    for (int k = 0; k < ws->current.floating->length; ++k) {
        [...]
        render_floating_container(soutput, damage, floater);
    }
}
```

Issue #5139, opened 2020-03-25
RFC: change focus_on_window_activation default to urgent


emersion:
Right now the default is "smart". This means a visible app can steal focus whenever it wants to. This can be an issue since having focus allows for e.g. clipboard access.

We could change the default to "urgent" instead. This would potentially break some users' config, but users who care about it can just change the setting in their config file.

ddevault:
+1

PR #5150, opened 2020-03-27
Change focus_on_window_activation default to urgent

emersion:
Before the default was "smart". This means a visible app could steal focus
whenever it wanted to. This can be an issue since having focus allows for
e.g. clipboard access.

This commit changes the default to "urgent" instead.

Closes: https://github.com/swaywm/sway/issues/5139


ddevault:
Thanks!

PR #5156, opened 2020-03-29
build: make completions respect install prefix

martinetd:
paths straight from pkgconfig will sometimes be outside of the users
build prefix, we should never try to install anything outside of that
directory.

filter the value obtained from pkgconfig and use some fallback if it
is not inside

----
There might be a cleaner way of doing that, but I didn't manage to split the if e.g. something like the following fails because bash_install_dir isn't defined if it didn't get in first loop and meson doesn't see to have lazy evaluation?
@ascent12 might have a better idea, happy to change it as long as we respect prefix.

```meson
        if bash_comp.found()
                bash_install_dir = bash_comp.get_pkgconfig_variable('completionsdir')
        endif
        if not bash_comp.found() or not bash_install_dir.startswith(prefix)
                bash_install_dir = join_paths(datadir, 'bash-completion', 'completions')
        endif
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -264,6 +264,9 @@ if get_option('bash-completions')
 	)
 	if bash_comp.found()
 		bash_install_dir = bash_comp.get_pkgconfig_variable('completionsdir')
+		if not bash_install_dir.startswith(prefix)
+			bash_install_dir = join_paths(datadir, 'bash-completion', 'completions')
+		endif
 	else
 		bash_install_dir = join_paths(datadir, 'bash-completion', 'completions')
 	endif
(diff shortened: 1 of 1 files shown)


martinetd:
ah, actually realized the way I tested this was flawed, if meson has lazy evaluation that might work. Still a bit ugly but better than current patch...

emersion:
See https://github.com/swaywm/swaybg/issues/13. I think this should be fixed by making pkg-config files use the `${prefix}` variable instead.

martinetd:
Didn't look in other repos & had missed the swaybg issue, sorry.

Hm, looking at the fish thread it isn't going quite well, but I understand sway's position of not wanting workarounds so closing this.
I'll try to poke at bash lists if you haven't yet (didn't check yet), let's see how this goes...

PR #5162, opened 2020-03-31
Use bitshifts for bitfields

emersion:
Instead of hardcoded power of 2 values, use bitshifts. This makes the
enums more readable, avoids mistakes, and makes it clear how much of the
int32_t bit space we have left.

While at it, fix other minor style issues.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -33,15 +33,15 @@ enum binding_input_type {
 };
 
 enum binding_flags {
-	BINDING_RELEASE=1,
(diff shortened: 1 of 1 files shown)


ianyfan:
Would the style changes add noise to the blame history?

emersion:
Can leave these out if you prefer.

RedSoxFan:
Thanks! Sorry for the delayed review.

emersion:
No worries, we aren't in a hurry :)

PR #5177, opened 2020-04-04
build: make completions respect install prefixes

martinetd:
Tell pkgconfig about prefix and datadir as required in the .pc files, so
if the prefix isn't standard nothing is installed outside of it.

For fish, this requires https://github.com/fish-shell/fish-shell/pull/6778

Fixes swaywm/swaybg#13

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -263,7 +263,10 @@ if get_option('bash-completions')
 		'completions/bash/swaymsg',
 	)
 	if bash_comp.found()
-		bash_install_dir = bash_comp.get_pkgconfig_variable('completionsdir')
+		bash_install_dir = bash_comp.get_pkgconfig_variable(
+			'completionsdir',
+			define_variable: ['datadir', datadir]
+		)
 	else
 		bash_install_dir = join_paths(datadir, 'bash-completion', 'completions')
 	endif
@@ -278,7 +281,10 @@ if get_option('fish-completions')
 		'completions/fish/swaynag.fish',
 	)
 	if fish_comp.found()
-		fish_install_dir = fish_comp.get_pkgconfig_variable('completionsdir')
+		fish_install_dir = fish_comp.get_pkgconfig_variable(
+			'completionsdir',
+			define_variable: ['datadir', datadir]
+		)
 	else
 		fish_install_dir = join_paths(datadir, 'fish', 'vendor_completions.d')
 	endif


meson.build
```diff
@@ -278,7 +281,10 @@ if get_option('fish-completions')
 		'completions/fish/swaynag.fish',
 	)
 	if fish_comp.found()
-		fish_install_dir = fish_comp.get_pkgconfig_variable('completionsdir')
+		fish_install_dir = fish_comp.get_pkgconfig_variable(
+			'completionsdir',
+			define_variable: ['datadir', datadir]
```
emersion, line 286: `datadir` is relative to the `prefix`. We need to use the full path here: `prefix / datadir`.

meson.build
```diff
@@ -263,7 +263,10 @@ if get_option('bash-completions')
 		'completions/bash/swaymsg',
 	)
 	if bash_comp.found()
-		bash_install_dir = bash_comp.get_pkgconfig_variable('completionsdir')
+		bash_install_dir = bash_comp.get_pkgconfig_variable(
+			'completionsdir',
+			define_variable: ['prefix', prefix]
```
emersion, line 268: We'll want to define `datadir` here, not `prefix`. Here's the bash pc file:

```
datadir=/usr/share
completionsdir=${datadir}/bash-completion/completions
```

martinetd: Hm, my bash-completion.pc uses prefix: `completionsdir=${prefix}/share/bash-completion/completions`

I guess we can just send both variables...

meson.build
martinetd: bah stupid github doesn't send comment when you resolve... retyping -_-

`meson configure` does display a relative path, but it sends the full path here -- I tested it, fish completion is properly installed within the prefix.

meson.build
martinetd: arg, I can't find a working syntax to send multiple variable?! what the heck, meson.

```
                bash_install_dir = bash_comp.get_pkgconfig_variable(
                        'completionsdir',
                        define_variable: [
                                'prefix', prefix,
                                'datadir', datadir,
                        ]
                )
->
../meson.build:267:2: ERROR: define_variable must be made up of 2 strings for VARIABLENAME and VARIABLEVALUE
```

emersion: It seems like upstream uses `datadir`:

https://github.com/scop/bash-completion/blob/master/bash-completion.pc.in#L6

meson.build
emersion: Ah, that's likely because the resulting string is fed into `install_dir`, which accepts paths relative to `prefix`. Should be fine then!

meson.build
martinetd: Yeah I just checked that and was testing. pushed update into datadir

PR #5189, opened 2020-04-08
WIP: Use a static size for titlebar height

### LINKED ISSUE #4992: Japanese/Chinese characters change title height which can cause screen flashing

oranenj:
This patch makes the maximum text height configurable instead of trying to change titlebar height dynamically.

should fix #4992, though some fine-tuning is still needed. Is it preferable to get rid of vertical padding and have one variable only concerned with titlebar height?

The text ascent information is used to move the Cairo coordinates so that there will not be a ton of empty space in the texture; essentially when done this way the texture only contains the actual painted pixels, so positioning it is easier to reason about.

I've been running this patch on top of 1.4 for a while now and I haven't run into any issues.

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -82,6 +82,40 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
 	return layout;
 }
 
+void get_text_physical_size(cairo_t *cairo, const char *font, int* ascent, int* descent,
+		int* ink_size, bool markup, const char *fmt, ...) {
+	va_list args;
+	va_start(args, fmt);
+	// Add one since vsnprintf excludes null terminator.
+	int length = vsnprintf(NULL, 0, fmt, args) + 1;
+	va_end(args);
+
+	char *buf = malloc(length);
+	if (buf == NULL) {
+		sway_log(SWAY_ERROR, "Failed to allocate memory");
+		return;
+	}
+	va_start(args, fmt);
+	vsnprintf(buf, length, fmt, args);
+	va_end(args);
+
+	PangoLayout *layout = get_pango_layout(cairo, font, buf, 1, markup);
+	pango_cairo_update_layout(cairo, layout);
+	PangoRectangle ir, lr;
+	pango_layout_get_pixel_extents(layout, &ir, &lr);
+	if (ascent) {
+		*ascent = PANGO_ASCENT(ir);
+	}
+	if (descent) {
+		*descent = PANGO_DESCENT(ir);
+	}
+	if (ink_size) {
+		*ink_size = ir.height;
+	}
+	g_object_unref(layout);
+	free(buf);
+}
+
 void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
 		int *baseline, double scale, bool markup, const char *fmt, ...) {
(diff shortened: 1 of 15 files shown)


vvrein:
Setting max font height is great, and I think this shoud be done, but what about max title bar height?

thestr4ng3r:
What is missing here for this to be not WIP anymore?

WhyNotHugo:
Seems that this requires users to manually compute and set this value, rather than have sway calculate it itself.

That feels somewhat of changing one problem for another -- I'd now have to figure out how to compute that, and need to re-do that math when I change the font size.

I'd rather expect sway to do the right thing automatically -- eg: use a default max size that can fit most "normal" text (latin/cjk/emoji).

emersion:
>I'd rather expect sway to do the right thing automatically -- eg: use a default max size that can fit most "normal" text (latin/cjk/emoji).

Nobody has found a way to do this without doing Latin-centric computations.

thestr4ng3r:
I would argue that it is a valid thing for the user to specify a maximum height in pixels if desired, even if automatic computation was practical. A default maximum size might be larger or much less predictable than what is desired by the user.

oranenj:
I'll just note that if someone feels like picking up this work to get it into a mergeable state, please go ahead. I'm currently not very likely to have either the time or mental resources to work on refining my patch.

SL-RU:
I'm frustrated without that feature. With my lovely font titles in browser or in other applications are always changing height when there are russian letters or there are only latin.

madblobfish:
> Nobody has found a way to do this without doing Latin-centric computations.
How about letting the use define a string of which the height is then evaluated?

MarkPash:
What needs to be done to get this into a mergeable state? Are there objections to how this PR addresses the issue?
I'm also very much interested in this patch being in master.

emersion:
>What needs to be done to get this into a mergeable state?

1. Implement a fixed size option as suggested in https://github.com/swaywm/sway/issues/4992#issuecomment-610967189
2. Submit a PR.

This PR seems to do something more complicated than what's necessary.

madblobfish:
Hi @emersion,
what about my idea of specifying a string of which the height is then evaluated?
That might be more user friendly and does not need to be adjusted to the font.

emersion:
I don't like this solution. It doesn't let you add/remove the hard-coded padding.

vvrein:
@emersion 
Could you take a look at [this](https://github.com/swaywm/sway/compare/b5b628cb41a4a316456d5ef3014f1a2cbdc5cfc2...vvrein:master) solution?
I'm using it almost a year, without almost any annoyance.
Some tweaks are needed to get this to mergeable state, but if this approach is ok in general - I can do it.

emersion:
Looks fine to me from a quick look, although I think there could be a better way to pick the baseline. Feel free to open a PR so that we can discuss!

oranenj:
@vvrein can you make a PR with that change? I think this PR can be closed, then.

vvrein:
@oranenj done https://github.com/swaywm/sway/pull/6059

oranenj:
I'll close this then.

PR #5190, opened 2020-04-08
Add subtitles to video on home page

danshick:
I put the vtt subtitle file in the assets directory, with the same name as the video. The video itself is hosted at sr.ht, so it may be more appropriate to put this vtt file there instead.

I'm also not entirely sure this will work correctly hosted from github pages, as some browsers may need the MIME type to be "text/vtt" and I'm not sure how github pages will serve it. Might work correctly out of the box. My local testing webserver applied the correct MIME type based on the extension automatically.

### DIFF
--- a/assets/lJ9C.vtt
+++ b/assets/lJ9C.vtt
@@ -1,0 +1,250 @@
+WEBVTT
+
+1
+00:00:00.000 --> 00:00:01.000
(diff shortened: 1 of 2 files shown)


danshick:
Big thanks to leon-p from #sway for writing these out.

Issue #5193, opened 2020-04-09
Fails to display if config has any output commands


jbeich:
Regressed by https://github.com/swaywm/sway/commit/0cdcf66bbccf

Environment:
```
$ cc --version
FreeBSD clang version 10.0.0 (git@github.com:llvm/llvm-project.git llvmorg-10.0.0-0-gd32170dbd5b)
Target: x86_64-unknown-freebsd13.0
Thread model: posix

$ pkg info -x libffi xkbcommon \^wayland drm-.\*kmod
libffi-3.2.1_3
libxkbcommon-0.10.0_1
wayland-1.18.0_3
wayland-protocols-1.20
drm-devel-kmod-5.0.g20200320

$ sway --version                                                                                                             ~
sway version 1.2-rc1-291-gc9fa7510
wlroots version 0.10.0-67-gd10f8a98
```

Steps to reproduce:
```
$ echo 'output * subpixel rgb' >/tmp/sway.conf
$ sway -c /tmp/sway.conf
```
Debug log:
```diff
--- good
+++ bad
@@ -36,6 +36,12 @@
 [types/wlr_idle.c:247] idle manager created
 [wayland] unable to lock lockfile /var/run/holo/wayland-0.lock, maybe another compositor is running
 [sway/config.c:387] Loading config from /tmp/sway.conf
+[sway/config.c:796] Read line 1: output * subpixel rgb
+[sway/commands.c:372] Config command: output * subpixel rgb
+[sway/commands.c:392] After replacement: output * subpixel rgb
+[sway/commands.c:423] Subcommand: subpixel rgb
+[sway/config/output.c:223] Adding output * config
+[sway/config/output.c:233] Config stored for output * (enabled: -1) (-1x-1@-1.000000Hz position -1,-1 scale -1.000000 subpixel rgb transform -1) (bg (null) (null)) (dpms 0) (max render time: -1)
 [sway/server.c:188] Initializing Xwayland (lazy=1)
 [sway/server.c:206] Starting backend on wayland display 'wayland-1'
 [backend/drm/drm.c:1323] Scanning DRM connectors
@@ -79,12 +85,9 @@
 [backend/drm/drm.c:1234]   'HDMI-A-3' crtc=-1 state=0 desired_enabled=0
 [backend/drm/drm.c:1535] Requesting modeset for 'DP-1'
 [sway/desktop/output.c:866] New output 0x802291f80: DP-1
-[sway/config/output.c:349] Turning on output DP-1
[...]

PR #5194, opened 2020-04-09
Fix initial modeset

emersion:
An if branch takes care of the case where the output needs to be turned
off (DPMS'ed or disabled). The other branch needs to unconditionally
enable the output.

output->current_mode is already taken care of in apply_config.

Sorry about that, probably made a bad change by mistake after my DRM testing.

Closes: https://github.com/swaywm/sway/issues/5193


ddevault:
Thanks!

PR #5197, opened 2020-04-09
Don't transfer focus to NULL node on touch

### LINKED ISSUE #5185: On screen keyboard steals focus on touch
Hello everyone,

recently there has been a change to transfer focus on touch events, this is all great, but it also applies to onscreen keyboards, which is not so great. It causes the on screen keyboard (squeekboard in my case), which uses the input method protocol to show up whenever a text field is focused, to instantly hide again as soon as it is being touched.

Steps to reproduce:
- install an on screen keyboard like squeekboard
- start it
- focus any text field for it to show up
- try typing using the touchscreen

Result:
- the keyboard instantly hides, since the focus moved away from the textfield

The expected result obviously is, that it should be possible to type using the touch keyboard.

My sway version: Sway version 1.4-3c96a1d3 (Apr  4 2020, branch 'master')

Best regards and thanks in advance :)

David96:
Fixes #5185

The issue is fixed when checking for NULL already, but I think the N_WORKSPACE check also makes sense.

Edit: removed the workspace part as proposed by emersion.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -408,7 +408,9 @@ static void handle_touch_down(struct wl_listener *listener, void *data) {
 				event->touch_id, sx, sy);
 	}
 
-	seat_set_focus(seat, focused_node);
+	if (focused_node) {
+		seat_set_focus(seat, focused_node);
+	}
 }
 
 static void handle_touch_up(struct wl_listener *listener, void *data) {


emersion:
>I think the N_WORKSPACE check also makes sense.

Why? I think a touch event on an empty workspace should focus the workspace.

David96:
Thinking about it again you're probably right.
I wasn't sure how sway handles the on screen keyboard thing so I started looking at the function for focus follows mouse, as this wasn't causing the keyboard to steal focus. There I found this https://github.com/swaywm/sway/blob/2b51c5b2942fbedfe34a912b47616e98cfcea02e/sway/input/seatop_default.c#L444
That's where I got the idea of checking for N_WORKSPACE from in the first place. After using it sway segfaulted that's how I found out that node could be NULL. And then I left in both checks thinking about the gaps but for touch this probably really doesn't make any sense, as you're not accidentally hovering over them. So I'll remove that check when I get to my computer.

David96:
Anything else? Otherwise I think this should be ready to merge.

PR #5200, opened 2020-04-10
input/seatop_default: release on empty workspace

### LINKED ISSUE #3932: wl_data_source.cancelled not sent when dropping outside any surface
* Sway Version: 1.0

* Reproduction Steps:

1. start weston-dnd, make it floating
2. drag to outside of weston-dnd and release mouse

Emantor:
Instead of handling presses and releases on empty workspaces as setting
focus to the workspace, handle releases by notifying the seat of a
pointer action. This way DnDs are correctly released if the button is
released over an empty workspace.

Fixes #3932

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -280,7 +280,10 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 
 	// Handle clicking an empty workspace
 	if (node && node->type == N_WORKSPACE) {
-		seat_set_focus(seat, node);
+		if (state == WLR_BUTTON_PRESSED) {
+			seat_set_focus(seat, node);
+		}
+		seat_pointer_notify_button(seat, time_msec, button, state);
 		return;
 	}
 


sway/input/seatop_default.c
```diff
@@ -279,8 +279,12 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 	}
 
 	// Handle clicking an empty workspace
-	if (node && node->type == N_WORKSPACE) {
-		seat_set_focus(seat, node);
+	if (node && node->type == N_WORKSPACE ) {
+		if ( state == WLR_BUTTON_PRESSED) {
```
emersion, line 283: Style: stray space here

sway/input/seatop_default.c
```diff
@@ -279,8 +279,12 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
...
+			seat_set_focus(seat, node);
+		} else {
+			seat_pointer_notify_button(seat, time_msec, button, state);
```
emersion, line 286: Hmm, we should probably call `seat_pointer_notify_button` unconditionally here?

sway/input/seatop_default.c
```diff
@@ -279,8 +279,12 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 	}
 
 	// Handle clicking an empty workspace
-	if (node && node->type == N_WORKSPACE) {
-		seat_set_focus(seat, node);
+	if (node && node->type == N_WORKSPACE ) {
```
emersion, line 282: Style: another stray space here

sway/input/seatop_default.c
Emantor: In that case I simply removed the return from the function, `handle_button()` will call `seat_pointer_notify_button()` at the very end, and none of the other functions should run on empty workspaces. Pushed the new version.

emersion:
Hmm. Why the `state == WLR_BUTTON_PRESSED` check?

I'm a little worried about removing that `return`. This is inconsistent with the rest of the function.

Emantor:
> Hmm. Why the `state == WLR_BUTTON_PRESSED` check?

The comment states that focus should change on click, not on release.

> I'm a little worried about removing that `return`. This is inconsistent with the rest of the function.

IMO this function and the pointer handling will require more cleanups to fix all issues with DnD. But I did not have the time to do a deeper analysis. I can change this back to include the return.

emersion:
>I can change this back to include the return.

I'd prefer to keep this if branch consistent with the rest of the function. It can be surprising that all if branches `return` except this one.

Emantor:
Pushed a version which includes the return.

PR #5204, opened 2020-04-10
Virtual keyboard grouping fix

any1:


emersion:
@RedSoxFan Do you remember why we don't use `libinput_device_get_device_group` to group keyboards?

RedSoxFan:
> 
> 
> @RedSoxFan Do you remember why we don't use `libinput_device_get_device_group` to group keyboards?

I think we wanted something that was that was backend independent. It's been a while. I can check over the weekend. _(I'm currently assembling a new computer otherwise I'd check now)_

RedSoxFan:
@emersion I looked into the libinput device groupings. On my systems, the inputs that should be group together are all in separate libinput device groups.

As for this PR, it looks fine to me. The alternative would be to change the wlr_keyboard_group the device is in when the keymap is about to change (as is done for physical keyboards). That should allow the virtual keyboards to be grouped with physical keyboards with the same keymap, but also not effect the keymap of other keyboards.

emersion:
>On my systems, the inputs that should be group together are all in separate libinput device groups.

Ah, that explains it. I wonder if we should ask users to send libinput patches to fix this upstream instead, this would benefit other compositors as well. I've asked whot about this, let's see how it goes.

emersion:
>That should allow the virtual keyboards to be grouped with physical keyboards with the same keymap, but also not effect the keymap of other keyboards.

I don't think this is something we want. If two VNC sessions have the same keymap and one of them decides to change it, it shouldn't affect the other one.

emersion:
@RedSoxFan Can you share a `libinput list-devices` dump?

RedSoxFan:
@emersion Here you go: https://gist.github.com/RedSoxFan/3e6cc09f61b179630ccde0db3d2a41ed

It appears that the inputs for the (terrible) keyboard that I have plugged into my desktop are actually registering in a single group today _I'm pretty sure they were not the other day..._

The inputs on my laptop are not being grouped though

PR #5215, opened 2020-04-14
Implement create_output command for headless backend

emersion:


ddevault:
Thanks!

genofire:
Is it now possible to run sway without any real screen? (maybe for vnc use by wayvnc?)

emersion:
Was possible before with `WLR_BACKENDS=headless`.

PR #5216, opened 2020-04-14
Add a secondary headless backend

emersion:
This allows the create_output command to work on DRM too.

Depends on ~~https://github.com/swaywm/wlroots/pull/2111~~, ~~https://github.com/swaywm/sway/pull/5215~~ and ~~https://github.com/swaywm/wlroots/pull/2063~~

cc @any1

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -4,6 +4,8 @@
 #include <stdlib.h>
 #include <wayland-server-core.h>
 #include <wlr/backend.h>
+#include <wlr/backend/headless.h>
+#include <wlr/backend/multi.h>
 #include <wlr/backend/noop.h>
 #include <wlr/backend/session.h>
 #include <wlr/render/wlr_renderer.h>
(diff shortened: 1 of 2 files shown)


sway/server.c
```diff
@@ -153,9 +154,15 @@ bool server_init(struct sway_server *server) {
 		return false;
 	}
 
+	server->noop_backend = wlr_noop_backend_create(server->wl_display);
+
 	struct wlr_output *wlr_output = wlr_noop_add_output(server->noop_backend);
 	root->noop_output = output_create(wlr_output);
 
+	server->headless_backend =
```
any1, line 162: What happens here if a headless backend was already created in `wlr_backend_autocreate`?

emersion: Then we get two headless backends, but it's no big deal, since they share the renderer.

any1: Might this be a potential source of confusion for future contributors?

emersion: I think conditionally creating the headless backend would be even more confusing.

ddevault:
Rebase?

sway/server.c
emersion: Added a comment to disambiguate

emersion:
Rebased again

ddevault:
Thanks!

PR #5222, opened 2020-04-16
input/seatop_default: properly notify pointer leave

### LINKED ISSUE #5220: wl_data_device.leave events are not sent
**Steps to repro:**
- Open 2 nautilus windows and put them in floating mode
- From window 1, start dragging a file until the pointer enters the window 2;
- Drag it out of window 2 into desktop shell (outside any toplevel surface) and drop the file;

**Expected result:**
- The dnd is canceled and keeps the dragged file on the window 1's folder

**Actual result:**
- The file is moved to window 2's folder

From the [debug logs](https://gist.github.com/nickdiego/fd10a91d9598baa4a6df470b4e798d18), it seems like `wl_data_device::leave` events are not sent when the pointer actually leaves the surface, instead they are sent when it enters another surface.

It's also reproducible in tiling mode, for example, when the drag enters a 2nd surface and then leaves to a 2nd screen (with no surface) and drop the file there.

**Version:**
```
➜  ~ swaymsg -t get_version                                                                                     
sway version 1.4-0efc28e2f58a (Apr 15 2020, branch 'master')
```
This already includes fix for #3932 .

nickdiego:
Currently, clients receive wl_data_device::leave events only when the
pointer enters another surface, which leads to issues, such as #5220.
This happens because wlr_seat_pointer_notify_enter() is called when
handling motion events only for non-NULL surfaces.

Fixes #5220

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -219,6 +219,7 @@ static void cursor_do_rebase(struct sway_cursor *cursor, uint32_t time_msec,
 	}
 
 	if (surface == NULL) {
+		wlr_seat_pointer_notify_enter(wlr_seat, NULL, 0, 0);
 		wlr_seat_pointer_clear_focus(wlr_seat);
 	}
 }


sway/input/seatop_default.c
```diff
@@ -200,25 +200,26 @@ static void cursor_do_rebase(struct sway_cursor *cursor, uint32_t time_msec,
...
-		enum wlr_edges edge = find_resize_edge(node->sway_container, surface, cursor);
-		if (edge == WLR_EDGE_NONE) {
-			cursor_set_image(cursor, "left_ptr", NULL);
-		} else if (container_is_floating(node->sway_container)) {
-			cursor_set_image(cursor, wlr_xcursor_get_resize_name(edge), NULL);
-		} else {
-			if (edge & (WLR_EDGE_LEFT | WLR_EDGE_RIGHT)) {
-				cursor_set_image(cursor, "col-resize", NULL);
+	} else {
+		if (node && node->type == N_CONTAINER) {
```
emersion, line 204: Thanks for you PR!

Why is this block changed? It would make git-bisect's life easier if only one line was modified to add a call to `wlr_seat_pointer_notify_enter`.

nickdiego: I guess you mean `git-blame` here, right?

My motivation for this was readability, but TBH `cursor_do_rebase()` doesn't seem to be the correct place to notify enter/leave events. So, as an improvement I'd propose to either (1) move event notification out of this function to where it's called from, or (2) rename this function to better reflect what it actually does. wdyt?
Anyways, if you still prefer just to add the missing call, I'll proceed and do it :)

Thanks

emersion: >I guess you mean git-blame here, right?

Yeah, git-blame too. When searching for the cause of a bug, it'll be harder if a commit contains unrelated refactoring.

>move event notification out of this function to where it's called from

Interestingly, @Xyene suggested the same thing.

"Rebasing" here means that "update the pointer to be consistent with whatever is under it". It's used when changing the layout for instance, or after a TTY switch. If we remove enter/leave events from this function, we could rename it to something like `cursor_update_image` and re-use it from the tablet code.

nickdiego: > Yeah, git-blame too. When searching for the cause of a bug, it'll be harder if a commit contains unrelated refactoring.

Ack and done.

> If we remove enter/leave events from this function, we could rename it to something like cursor_update_image and re-use it from the tablet code.

Sounds like something worth doing. Will send it in a followup patch soon.
Thanks

Xyene: :+1: 

To summarize a discussion from IRC yesterday (since I imagine you might find it interesting):

* the cursor/seat code predates tablet and touch specs, so is tightly coupled with pointers
* this is "surprising", e.g. you'd expect functions like `cursor_motion` or `cursor_do_rebase` to not send pointer events, but they do
   * this means they're not used for tablet/touch support, instead both tablet and touch re-implement ad-hoc parts of that functionality
   * since fewer people use those input methods, it's easier for the code to fall behind without anyone noticing
* tablet support, at least, currently has [some issues](https://github.com/swaywm/sway/issues?q=is%3Aopen+is%3Aissue+author%3AXyene+tablet)<sup>1</sup> that would likely best be fixed by decoupling the `*cursor*` functions from sending pointer events, and moving those events to the call sites

The plan proposed looks something like this:

1. move the pointer events out of the cursor functions
2. reuse them for tablet support &mdash; this'd fix a bunch of tablet issues
3. for future maintainability, add handlers for tablet and touch into seatop, so that the logic for those devices can be migrated from `cursor.c`
4. once everything's in one place rather than two and confirmed "not broken by refactoring", it'll be easier to see what abstractions make sense

<sup>1</sup> There are more if you just search for "tablet", but these are the ones I know the cause of so far :wink:

nickdiego: Hi @Xyene, thanks for the info. It sounds like a good plan for me.

nickdiego:
> This one-liner already fixes a bug. Can you rebase and squash the commits?

Done. Thanks

emersion:
Thanks for the fix!

tchebb:
This change caused wlroots to segfault on pointer leave when an xdg_shell popup is active. I've written up an analysis at swaywm/wlroots#2162, but the tl;dr is I think we should revert this until we fix the underlying issue.

emersion:
Let's wait a little bit, I'll revert if we don't come up with a fix in a few days.

travankor:
revert?

PR #5229, opened 2020-04-18
Implement pointer simulation if client hasn't bound to touch

David96:
This implements simulating cursor events if the surface under the cursor hasn't bound to touch events.

A touch down is interpreted as button down, movement moves the cursor with the button pressed and touch up releases it.

I'm not exactly sure about the check for the pointer constraints - my idea was to ignore the touch if it's outside the constraint.

**Edit:** this should now correctly determine whether a surface supports touch events. It depends on https://github.com/swaywm/wlroots/pull/2151
I also removed all pointer constraints checks and added some multi touch handling (all touches except the first one are be ignored).

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -50,6 +50,8 @@ struct sway_cursor {
 	struct wl_listener touch_down;
 	struct wl_listener touch_up;
 	struct wl_listener touch_motion;
+	bool simulating_pointer_from_touch;
+	int32_t pointer_touch_id;
 
 	struct wl_listener tool_axis;
 	struct wl_listener tool_tip;
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -402,7 +402,6 @@ static void handle_pointer_frame(struct wl_listener *listener, void *data) {
 
 static void handle_touch_down(struct wl_listener *listener, void *data) {
 	struct sway_cursor *cursor = wl_container_of(listener, cursor, touch_down);
-	cursor_handle_activity(cursor, IDLE_SOURCE_TOUCH);
 	struct wlr_event_touch_down *event = data;
 
 	struct sway_seat *seat = cursor->seat;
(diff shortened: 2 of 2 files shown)


Xyene:
:+1:

I don't have any authority regarding this, but w.r.t. the pointer constraints, it might be best to do nothing from a consistency standpoint &mdash; I ran into a similar issue with tablet input. The constraint comes from `pointer-constraints-unstable-v1`, and is meant for mouse pointers. In particular, tablet input "ignores" any pointer constraints, since the tablet tool is not considered a pointer by the spec.

sway/input/cursor.c
```diff
@@ -450,12 +466,12 @@ static void handle_touch_motion(struct wl_listener *listener, void *data) {
 		}
 	}
 
-	if (!surface) {
-		return;
-	}
-
-	// TODO: fall back to cursor simulation if client has not bound to touch
-	if (seat_is_input_allowed(cursor->seat, surface)) {
```
Xyene, line 458: I don't *think* that's what `seat_is_input_allowed` means. I'm pretty sure it'll return true regardless of if touch is bound, so long as the client is accepting input (of any kind?) That's my impression from playing with the tablet code, at least.

David96: Hmm yeah, I was just assuming from the comment that this is what it does - and for my main use case (using touch to switch windows using their titlebars) it works. As basically all apps I use, use gtk they all bind to touch anyway so it's hard to test. But I'll have a closer look at that.

David96:
Yes, the constraints come from the mouse pointer but that's exactly what this is supposed to be emulating right?
If my mouse cursor were constrained to a window, I wouldn't expect touch to be able to move it out of its bounds.

sway/input/cursor.c
David96: Looking at the code, I definitely think you're right - now I "only" have to find the right way to check for that, if anyone has an idea, it would be nice to know :)

emersion: `seat_is_input_allowed` is used for lockscreens: when a lockscreen is active input events can only be sent to the lockscreen client.

Xyene:
Oh, I definitely agree that it would be nice if emulated pointer input was subject to pointer constraints. For my _personal_ usage at least, it'd even be preferable :smile: 

At the same time, I think if you were to try e.g. GNOME you'd find it also ignores the pointer constraints for emulated touch (I'm guessing here, since that's what it does for emulated tablet input). The pre-existing tablet code also ignores pointer constraints for emulated input.

I have no idea how well that plays with apps &mdash; I can imagine a situation where an app that `assert`s that all inputs come from within the constraint crashing, but I don't know if any such apps exist.

I think regardless of whether sway should match the behavior of other compositors, it should at least be consistent with itself. Whether that means both touch and tablet input should be subject to constraints or whether both _shouldn't_ is an interesting question. There were some concerns when I brought this up ~a month ago, but that discussion petered out due to IRL constraints :wink: ... so I don't know.

sway/input/cursor.c
Xyene: At least Wayland-native don't bind touch, e.g. kitty or alacritty, but I don't know what the check would look like. Tablets support has `wlr_surface_accepts_tablet_v2`, but I couldn't find an touch analog in a quick grep.

emersion: Yeah, i think we need to add `wlr_surface_accepts_touch` in wlroots.

include/sway/input/cursor.h
```diff
@@ -67,6 +67,9 @@ struct sway_cursor {
 	bool hidden;
 
 	size_t pressed_button_count;
+
+	bool simulating_pointer;
```
Xyene, line 71: Two small nits about this:
* this would be best to keep right after the declarations for `touch_down`, `touch_up` etc. so they're in one group
* `simulating_pointer` is a bit of a misleading name. Since we might be emulating pointer input for both tablet and touch devices _at the same time_, a single field isn't sufficient the way this implies. Tablets now use `cursor->simulated_tool_tip_down`, so a more specific name would be good here.

David96: Yeah I wasn't sure about the naming either, so I just went with whatever came to my mind first.
How about simulated_touch_down? To me this sounds like the touch down is being simulated though but it's more in line with the tablet version (which again to me sounds like the tool tip is being simulated not the pointer but that might be just me). I personally would probably prefer simulated_by_touch and simulated_by_tablet.

Xyene: Yeah, I'm not a huge fan of `simulated_tool_tip_down` either, since what's being simulated is "backwards". `cursor->simulated_by_touch/tablet` is a bit better, but it makes it sound like the _cursor_ is being simulated.

Maybe `simulating_pointer_from_touch` and `simulating_pointer_from_tool_tip` (or `simulating_touch_as_pointer` and `simulating_tool_tip_as_pointer`)? A bit of a mouthful, but more "clear" (and only used in one or two places).

David96: I went with simulating_pointer_from_touch since I feel it's the clearest and as you said, it's only used a few times.

PR #5233, opened 2020-04-19
Add missing include

eknoes:
I was wondering if I oversee something, I just wanted to compile sway with debug symbols but could not start it without including `#include <wlr/types/wlr_output_power_management_v1.h>` in `server.c`.

Otherwise I got an error message "undefined symbol wlr_output_power_manager_v1".


eknoes:
Okay, sorry. Strange that it didn't work for me.

Issue #5232, opened 2020-04-19
Resizing windows with tablet pen is extremely jerky


Xyene:
I'm running sway master. If I open a blank workspace and open two apps, resizing them via mouse is smooth/instant. If I do the same with my tablet pen, the resizing is very jerky, and trails the pen position by over a second. Pen movement is otherwise very smooth.

My mouse sends events (as per `evhz`) at 1000 Hz, while my tablet only at 250 Hz, so if anything it seems like the mouse should place a higher load on sway. I'm not too sure how to go about tracking this down.

I'm happy to provide other information that might be useful.

Xyene:
The PR I've opened addresses the simple case of resizing a non-v2 surface. With it applied, the problem persists for v2 surfaces (a simple test-case is `gedit`, which binds tablet v2 handlers).

I've found the cause, but don't know what the most accepted fix here would be. The root cause is here:

https://github.com/swaywm/sway/blob/2b15cf453e4b28324e9012515011a705c2960b30/sway/input/cursor.c#L524-L528

When we're emulating pointer events _or_ are e.g. in the gap between two windows, we call `cursor_motion`. That eventually calls the seat's motion callback.

https://github.com/swaywm/sway/blob/2b15cf453e4b28324e9012515011a705c2960b30/sway/input/seatop_default.c#L485-L488

...which calls `cursor_do_rebase` that actually handles the resizing.

The problem seems to be that the drag is started with the emulated pointer, but when the tablet pen enters a tablet v2 surface, emulated pointer events stop being generated... so the resize is "stuck". `handle_motion` etc. all have calls to `wlr_seat_pointer_*` in them, so unless I'm missing something a bunch of that code would need to be refactored.

PR #5238, opened 2020-04-20
input/cursor: fix jerky resizing of non-tablet_v2 surfaces

Xyene:
This fixes resizing containers being very jerky under pointer emulation.

Refs #5232.


imyxh:
Was excited to see this as I face the same problem with my tablet. Happy to say that it's made the resizing (and moving) of windows very smooth.

Xyene:
@imyxh happy to hear that :smile: 

Unfortunately, I don't think this fixes resizing tablet v2 surfaces. I think if you open two gedit windows (which are GTK so by default seem to accept it), resizing is still jerky. (I'm currently trying to fix this; I know why it happens but it seems to existing code can't easily be changed to have it not happen.)

In the "tablet fixes not yet merged" category, you might also be interested in https://github.com/swaywm/sway/pull/5108 and https://github.com/swaywm/sway/pull/5237 :smile:

Issue #5228, opened 2020-04-17
Can't disable subpixel rendering in title bars


imyxh:
So, I asked a [question](https://www.reddit.com/r/swaywm/comments/cocyw9/how_do_i_turn_off_subpixel_rendering_in_title_bars/) about this on Reddit a few months ago, but I guess I never reported a GitHub issue. I'm now fairly certain it's a bug so here we go.

![image](https://user-images.githubusercontent.com/12703838/79621361-f83fcf00-80c7-11ea-87f2-5953df45e92d.png)

The image above is a snapshot of my title bar. It's very clearly using subpixel rendering—which in my opinion, on my display, looks awful. To make it clearer, here's a screenshot of a screenshot in GIMP:

![image](https://user-images.githubusercontent.com/12703838/79621411-27564080-80c8-11ea-9bc4-dc38b4934295.png)

I have set my sway config like so:

```
# (in: ~/.config/sway/config)
output * subpixel none
```

I have also disabled it in conf.d:

```
$ file /etc/fonts/conf.d/10-no-sub-pixel.conf
/etc/fonts/conf.d/10-no-sub-pixel.conf: symbolic link to ../conf.avail/10-no-sub-pixel.conf
```

And for the most part, it works. Screenshot from terminal:

![image](https://user-images.githubusercontent.com/12703838/79621469-671d2800-80c8-11ea-8293-01dae080bdd5.png)

You can open that up in GIMP and tell that indeed, there is only grayscale antialiasing, and no subpixel rendering.

Mako, however ...

![image](https://user-images.githubusercontent.com/12703838/79621522-96cc3000-80c8-11ea-940d-ae67db776014.png)

... also displays the same issue as the sway window titlebars. These tests are all using the same font (Iosevka 10). I've checked plenty of applications; for the most part, they all use grayscale antialiasing; even black-on-white text in Firefox uses grayscale. Everything—wayland and xwayland—seems to use my grayscale setting except Mako and the titlebars.

How may I help debug further to find the cause of this subpixel rendering? You would expect the `output * subpixel none` command to turn it off.

progandy:
The renderers in sway and mako translate ```WL_OUTPUT_SUBPIXEL_NONE``` to subpixel antialias with ***default*** subpixel order. It should probably be changed to choose grayscale antialiasing instead.

* [mako](https://github.com/emersion/mako/blob/175b8d8ab752bcc29da39ff89eca0b274af66628/render.c#L59) (you can create a bug over there)
[...]

PR #5239, opened 2020-04-20
interpret "subpixel none" as CAIRO_ANTIALIAS_GRAY

imyxh:
Fixes issue #5228. Currently, WL_OUTPUT_SUBPIXEL_NONE is ignored and CAIRO_ANTIALIAS_SUBPIXEL is still set. This commit checks if subpixel is set to none and if so, calls set_antialias with CAIRO_ANTIALIAS_GRAY. This mirrors the functionality in Mako's [PR261](https://github.com/emersion/mako/pull/261).

@progandy brought up that we could alternatively "extend it to 'none' = no antialias, 'gray' = grayscale" and/or read from the fontconfig setup.


emersion:
> we could alternatively "extend it to 'none' = no antialias, 'gray' = grayscale" and/or read from the fontconfig setup.

Subpixel none doesn't mean no antialias. Subpixel none means that the physical pixels don't have a geometry, ie. antialiasing shouldn't make any assumption about subpixel geometry.

PR #5240, opened 2020-04-20
input/cursor: fix hide cursor timeouts on tablets and touchscreens

### LINKED ISSUE #5169: Cursor gets hidden even it is currently being used with a drawing tablet (wacom)
* Sway Version: 1.4

* Configuration File:

```
input 1386:788:Wacom_Intuos_Pro_S_(WL)_Pen {
    map_to_output DP-1
    map_from_region 0.0x0.0 1.0x0.901
}

# Define seat for wacom tablet
seat seat_wacom attach "1386:788:Wacom_Intuos_Pro_S_(WL)_Pen"

### Seat configurations
# For the default seat
seat seat0 hide_cursor 10000

# For wacom seat
seat seat_wacom hide_cursor 10000
seat seat_wacom idle_inhibit tablet_tool
seat seat_wacom idle_wake tablet_tool
```

I have a "Wacom Intuos Pro S" which I want to use with its own _seat_. I also want the cursor corresponding to this input device to hide after 10 sec if it has not been used. This seems to work... but not quite. If I use the pen of the tablet on an empty monitor (no program opened) then the cursor does not get hidden. However, if I use a program (e.g. MyPaint or Xournal++) and draw something, then the cursor gets hidden after 10 sec, even if I am constantly moving the pen.

There are other problems I have been experiencing with this configuration, like swaylock getting started even if am using the keyboard (it happened also while I was writing this). 

This is also the first time that I am trying to configure different seats, so it might well be something that I am missing.

Xyene:
This commit refactors `cursor_handle_activity` to also take the idle
source, so that it can be reused for tablet and touch activity.
Previously, the timeouts would be tracked, but the cursor would never be
un-hidden for anything but pointers (including emulated pointers on
non-tablet_v2 surfaces).

Fixes #5169.

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -85,7 +85,8 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat);
 void cursor_rebase(struct sway_cursor *cursor);
 void cursor_rebase_all(void);
 
-void cursor_handle_activity(struct sway_cursor *cursor);
+void cursor_handle_activity(struct sway_cursor *cursor,
+	enum sway_input_idle_source idle_source);
 void cursor_unhide(struct sway_cursor *cursor);
 int cursor_get_timeout(struct sway_cursor *cursor);
 
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -241,11 +241,12 @@ int cursor_get_timeout(struct sway_cursor *cursor) {
 	return timeout;
 }
 
-void cursor_handle_activity(struct sway_cursor *cursor) {
+void cursor_handle_activity(struct sway_cursor *cursor,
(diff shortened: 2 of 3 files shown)


include/sway/input/cursor.h
```diff
@@ -85,7 +85,7 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat);
 void cursor_rebase(struct sway_cursor *cursor);
 void cursor_rebase_all(void);
 
-void cursor_handle_activity(struct sway_cursor *cursor);
+void cursor_handle_activity(struct sway_cursor *cursor, int idle_source);
```
Emantor, line 88: Please keep the type, i.e. ``
```suggestion
void cursor_handle_activity(struct sway_cursor *cursor, enum sway_input_idle_source source);
```

Issue #5236, opened 2020-04-19
Sway re-applies transforms on each config reload


icefoxen:
* Sway Version: sway version 1.4
* Debug log: http://paste.debian.net/1141361/
* Configuration File: Default config file, with the following additions for 3-monitor setup:

```
output DP-1 resolution 1920x1080 position 0 0 transform 90 clockwise
output DVI-D-1 resolution 1920x1080 position 1080 300
output DP-3 resolution 1920x1080 position 3000 0 transform 90 anticlockwise
```

 * Summary

Trying to configure a 3-monitor setup, with the two outside monitors rotated 90 degrees like so:

```
+----+            +-----+
|    | +--------+ |     |
|    | |        | |     |
|    | +--------+ |     |
+----+            +-----+
```

With the above configuration it works, but the transform rotations get re-applied every time the sway config file is reloaded.  So upon starting sway everything is correct.  Upon reloading the config with Mod4+Shift+c it rotates the outer two monitors again, so that "up" on each is facing inwards.  Reload again and they're upside-down, reload again and "up" is facing outwards, and reload again and everything faces upwards again.  So, it seems like the transform state isn't getting reset properly when the file is reloaded.  This isn't really a *critical* problem, but it does make playing with a sway config file a bit tedious, not to mention surreal.

Thank you!

progandy:
Do not use an incremental transform in your configuration file, but a fixed value.
90 and 270 instead of 90 (anti)clockwise.

emersion:
I know it's confusing, but `clockwise` means "add the transform to the current value". Maybe we should make it an error to use it from the config file.

icefoxen:
Thanks, at least it's an easy thing to fix.  What is `clockwise` and `anticlockwise` for then anyway, if not for use from a config file?

Emantor:
It's to be used via `swaymsg` where it will rotate the display `clockwise` or `anticlockwise`. Since `swaymsg` is for interactive use, it makes sense there.

PR #5241, opened 2020-04-20
Restrict relative output transforms to runtime changes

emersion:
Prevent them from being used in the config file.

This is a breaking config file change.

References: https://github.com/swaywm/sway/issues/5236


ddevault:
Thanks!

PR #5254, opened 2020-04-25
input/cursor: refactor cursor logic for reuse in tablet/touch

### LINKED ISSUE #5232: Resizing windows with tablet pen is extremely jerky

Xyene:
Fixes #5232 and hopefully other tablet bugs.

Will need rebasing after https://github.com/swaywm/sway/pull/5222 is merged.

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -220,6 +220,9 @@ bool seat_is_input_allowed(struct sway_seat *seat, struct wlr_surface *surface);
 
 void drag_icon_update_position(struct sway_drag_icon *icon);
 
+enum wlr_edges find_resize_edge(struct sway_container *cont,
+		struct wlr_surface *surface, struct sway_cursor *cursor);
+
 void seatop_begin_default(struct sway_seat *seat);
 
 void seatop_begin_down(struct sway_seat *seat, struct sway_container *con,
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -7,6 +7,7 @@
 #include <wlr/util/edges.h>
 #include "sway/config.h"
 #include "sway/input/input-manager.h"
+#include "sway/input/tablet.h"
 #include "sway/input/text_input.h"
 
 struct sway_seat;
@@ -19,6 +20,8 @@ struct sway_seatop_impl {
 			double dx, double dy);
 	void (*axis)(struct sway_seat *seat, struct wlr_event_pointer_axis *event);
 	void (*rebase)(struct sway_seat *seat, uint32_t time_msec);
+	void (*tablet_tool_motion)(struct sway_seat *seat, struct sway_tablet *tablet,
+			struct sway_tablet_tool *tool, uint32_t time_msec, double dx, double dy);
 	void (*end)(struct sway_seat *seat);
 	void (*unref)(struct sway_seat *seat, struct sway_container *con);
 	void (*render)(struct sway_seat *seat, struct sway_output *output,
@@ -260,6 +263,10 @@ void seatop_button(struct sway_seat *seat, uint32_t time_msec,
  */
 void seatop_motion(struct sway_seat *seat, uint32_t time_msec,
 		double dx, double dy);
+
+void seatop_tablet_tool_motion(struct sway_seat *seat,
+		struct sway_tablet *tablet, struct sway_tablet_tool *tool,
+		uint32_t time_msec, double dx, double dy);
 
 void seatop_axis(struct sway_seat *seat, struct wlr_event_pointer_axis *event);
 
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -289,7 +289,7 @@ void cursor_unhide(struct sway_cursor *cursor) {
 	cursor_rebase(cursor);
 }
 
-static void cursor_motion(struct sway_cursor *cursor, uint32_t time_msec,
+static void pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
 		struct wlr_input_device *device, double dx, double dy,
 		double dx_unaccel, double dy_unaccel) {
 	cursor_handle_activity(cursor, IDLE_SOURCE_POINTER);
@@ -324,17 +324,17 @@ static void cursor_motion(struct sway_cursor *cursor, uint32_t time_msec,
 	seatop_motion(cursor->seat, time_msec, dx, dy);
 }
 
-static void handle_cursor_motion_relative(
+static void handle_pointer_motion_relative(
 		struct wl_listener *listener, void *data) {
 	struct sway_cursor *cursor = wl_container_of(listener, cursor, motion);
 	struct wlr_event_pointer_motion *e = data;
 
-	cursor_motion(cursor, e->time_msec, e->device, e->delta_x, e->delta_y,
+	pointer_motion(cursor, e->time_msec, e->device, e->delta_x, e->delta_y,
 			e->unaccel_dx, e->unaccel_dy);
 	transaction_commit_dirty();
 }
 
-static void handle_cursor_motion_absolute(
+static void handle_pointer_motion_absolute(
 		struct wl_listener *listener, void *data) {
 	struct sway_cursor *cursor =
 		wl_container_of(listener, cursor, motion_absolute);
@@ -347,7 +347,7 @@ static void handle_cursor_motion_absolute(
 	double dx = lx - cursor->cursor->x;
 	double dy = ly - cursor->cursor->y;
 
-	cursor_motion(cursor, event->time_msec, event->device, dx, dy, dx, dy);
+	pointer_motion(cursor, event->time_msec, event->device, dx, dy, dx, dy);
 	transaction_commit_dirty();
 }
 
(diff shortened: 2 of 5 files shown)


sway/input/cursor.c
```diff
@@ -212,6 +212,27 @@ void cursor_rebase_all(void) {
 	}
 }
 
+void cursor_update_image(struct sway_cursor *cursor,
+		struct sway_node *node) {
+	if (node && node->type == N_CONTAINER) {
+		// Try a node's resize edge
+		enum wlr_edges edge = find_resize_edge(node->sway_container, NULL, cursor);
```
Xyene, line 219: Note for reviewer: the original code had

```c
enum wlr_edges edge = find_resize_edge(node->sway_container, surface, cursor);
```

...but following the control flow, `surface` was guaranteed to always be `NULL` at that point.

include/sway/input/seat.h
```diff
@@ -220,6 +223,9 @@ bool seat_is_input_allowed(struct sway_seat *seat, struct wlr_surface *surface);
 
 void drag_icon_update_position(struct sway_drag_icon *icon);
 
+enum wlr_edges find_resize_edge(struct sway_container *cont,
```
Xyene, line 226: I'm not super big on the semantics of having this function exported from here; if anyone has a better home for it I'm happy to move it.

Xyene:
I think this should be ready for review, but maybe not merging without a bit more testing. I'm running the following patches on my system, based on the latest master (which has the rest of the tablet fixes merged now):

* this
* https://github.com/swaywm/sway/pull/5258
* https://github.com/swaywm/sway/pull/5259

...and hopefully if there are issues I'll run into them in the next few days. Together, these fix most of the issues I've seen with tablets, though. @David96 or @imyxh, if either of you are feeling particularly brave, some more tablet user testing of these patches would go a long way :smile: 

Copy-pasteable commands if you choose to do so:

```shell
$ cd /path/to/sway
$ curl -L https://github.com/swaywm/sway/pull/5254.diff | patch -p1
$ curl -L https://github.com/swaywm/sway/pull/5258.diff | patch -p1
$ curl -L https://github.com/swaywm/sway/pull/5259.diff | patch -p1
```

Currently, this only adds a seatop handler for tablet motion. I think we ought to have one more at least for `seatop_down`, so that you can move a floating tablet v2 surface by tablet pen (currently, you can't, even without this patchset) &mdash; but I think that would best be done in a separate PR.

I would also like to rename the existing `button`, `motion`, and `axis` handlers to `pointer_button`, `pointer_motion` etc. to be unambiguous, but the resulting diff will be quite large. @emersion if you're OK with that, I can go ahead and do it.

I think there are definitely some opportunities for reducing code duplication here, but I think it'd be best to figure out whether the current logic is buggy or not before adding abstractions to it.

emersion:
>I would also like to rename the existing button, motion, and axis handlers to pointer_button, pointer_motion etc. to be unambiguous, but the resulting diff will be quite large. @emersion if you're OK with that, I can go ahead and do it.

Yes, please do in a separate commit.

sway/input/seatop_default.c
```diff
@@ -471,6 +473,43 @@ static void handle_motion(struct sway_seat *seat, uint32_t time_msec,
...
+	if (surface) {
+		if (seat_is_input_allowed(seat, surface)) {
+			struct sway_tablet_tool *sway_tool = tool->data;
+			wlr_tablet_v2_tablet_tool_notify_proximity_in(sway_tool->tablet_v2_tool,
+				tablet->tablet_v2, surface);
+			wlr_tablet_v2_tablet_tool_notify_motion(sway_tool->tablet_v2_tool, sx,
+				sy);
+		}
+	} else {
+		cursor_update_image(cursor, node);
```
emersion, line 500: Maybe we should do an equivalent of `wlr_seat_pointer_clear_focus` but for the tablet here?

include/sway/input/seat.h
```diff
@@ -19,6 +20,8 @@ struct sway_seatop_impl {
 			double dx, double dy);
 	void (*axis)(struct sway_seat *seat, struct wlr_event_pointer_axis *event);
 	void (*rebase)(struct sway_seat *seat, uint32_t time_msec);
+	void (*tablet_tool_motion)(struct sway_seat *seat, struct sway_tablet *tablet,
+			struct wlr_tablet_tool *tool, uint32_t time_msec, double dx, double dy);
```
emersion, line 24: Can we change the `tool` parameter to be a `struct sway_tablet_tool` directly?

sway/input/cursor.c
```diff
@@ -787,7 +787,7 @@ static void handle_constraint_commit(struct wl_listener *listener,
 	check_constraint_region(cursor);
 }
 
-static void handle_request_set_cursor(struct wl_listener *listener,
+static void handle_pointer_set_cursor(struct wl_listener *listener,
```
emersion, line 782: I'd prefer to keep "request" in the handler name here, to make it clear we react to a client request and not to a cursor change.

sway/input/seatop_default.c
Xyene: I think the closest equivalent would be `wlr_tablet_v2_tablet_tool_notify_proximity_out`. I did consider calling it here, but in practice it doesn't change the behavior (since it's implicitly called during the next frame anyway). That said, it's idempotent, so might as well be explicit and call it here too.

David96:
Is there any specific operations you would like us to test?

I just pulled your changes and am running sway with it right now - so far resizing windows appears to be smooth and from a superficial test, using the pen buttons seems to work better than ever (I have one eraser button and one that I bound to a select tool in my note taking app, the 2nd one never quite seemed to work right, now it does).

Only thing I'm not sure about: that one button seems to be supposed to work as right click? At least in FF it does. I also was able to at one point use it together with my modifier key to resize a window without grabbing its border (a 1px one in my case, quite hard to grab with the pen ^^) - but in this case the resizing wasn't quite working. As soon as the pen touched the screen it stopped and in general it was quite sluggish (maybe the offset between cursor and border was added to the window size upon resize start? Felt like something like that). And after playing around with it a bit I'm now no longer able to resize a window with this method at all. The window I was using is alacritty so I think without native tablet tool support. But this might be a general thing and not related to this pull request - I never played around with resizing windows so far, as I usually use my keyboard for that.

Also, I don't seem to be able to move windows supporting tablet tool events (tried with nemo and FF) using Mod+Pen. The events are just "eaten" by the client acting as if no modifier key was pressed.

sway/input/seatop_default.c
```diff
@@ -622,13 +634,23 @@ static void handle_rebase(struct sway_seat *seat, uint32_t time_msec) {
...
 			cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
-	cursor_do_rebase(cursor, time_msec, e->previous_node, surface, sx, sy);
+
+	if (surface) {
+		if (seat_is_input_allowed(seat, surface)) {
+			wlr_seat_pointer_notify_enter(seat->wlr_seat, surface, sx, sy);
+		}
+	} else {
+		cursor_update_image(cursor, e->previous_node);
+		wlr_seat_pointer_notify_enter(seat->wlr_seat, NULL, 0, 0);
```
emersion, line 644: Isn't this the same as `wlr_seat_pointer_clear_focus`?

emersion: Hmm I guess the old code had this too. Let's keep it like this in this PR.

Xyene: `wlr_seat_pointer_clear_focus` calls `wlr_seat_pointer_enter`, while this calls `wlr_seat_pointer_notify_enter`. But I agree it's confusing, I had a double-take at this too.

Xyene:
Thanks for testing @David96, I didn't really have much of a test plan other than "it seems to work better" :)

> I just pulled your changes and am running sway with it right now - so far resizing windows appears to be smooth and from a superficial test, using the pen buttons seems to work better than ever

Great!

> Only thing I'm not sure about: that one button seems to be supposed to work as right click?

I don't actually use the buttons on my pen so I've never noticed this, but I can confirm there's something weird going on here. My pen has two buttons. If I start `gedit` (by default, it binds tablet input) and press the "bottom" one, the right-click context menu opens. If I start `GDK_BACKEND=x11 gedit`, which *doesn't* bind tablet input, then I need to press the "top" button on my pen to trigger the same. So, there's some inconsistency there.

> Also, I don't seem to be able to move windows supporting tablet tool events (tried with nemo and FF) using Mod+Pen. The events are just "eaten" by the client acting as if no modifier key was pressed.

This is since we map tool tip down to left button press, and only enter the "move" state on left button press. On a v2 surface, we don't do this mapping, so never enter the move move. I've opened https://github.com/swaywm/sway/issues/5293 to track this.

PR #5255, opened 2020-04-25
[RFC] Add each view's shell to JSON description

tchebb:
This is a criteria you can use to select windows since commit 484cc189e909 ("Add shell criteria token"), but there's no way to query it for an existing window. This exposes its value in the output of `swaymsg -t get_tree`.

I don't know if this makes sense. Regarding criteria, `sway(5)` says "[y]ou may like to use `swaymsg -t get_tree` for finding the values of these properties", but shell isn't currently reported there. Some of the others aren't obviously included either, though, so maybe the man page just needs changing.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -493,6 +493,8 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 
 	json_object_object_add(object, "max_render_time", json_object_new_int(c->view->max_render_time));
 
+	json_object_object_add(object, "shell", json_object_new_string(view_get_shell(c->view)));
+
 #if HAVE_XWAYLAND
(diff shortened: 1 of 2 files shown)


RedSoxFan:
This makes sense to me and is trivial to maintain. I think this may help new users who are not familiar with the `app_id` vs `class` difference.

Can you please add an entry to [`sway-ipc.7.scd`](https://github.com/swaywm/sway/blob/5cf5349cd79f57433c2539742f087259eee985a1/sway/sway-ipc.7.scd#L285) for this? Once that is done, you can squash the two commits together and force push.

tchebb:
Done! Thanks for the review.

RedSoxFan:
Thanks!

PR #5265, opened 2020-04-27
[WIP] Make a tap on a windows title bar focus it

### LINKED ISSUE #4182: basic touch support for container tabs
now that https://github.com/swaywm/sway/issues/2377 is implemented, container title tabs in tab/stack layouts are the only things that don't respond to touch support. Nothing fancy, just activate the tab on touch like with a standard pointer. Why doesn't this work as a generic pointer? Is it because absolute input sources are treated differently?

ArenM:
This is my first attempt, which is basically a simplified copy of the
mouse down even handling.
this should fix #4182

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -17,6 +17,7 @@ struct sway_seatop_impl {
 			enum wlr_button_state state);
 	void (*motion)(struct sway_seat *seat, uint32_t time_msec,
 			double dx, double dy);
+	void (*touch_down)(struct sway_seat *seat, uint32_t time_msec);
 	void (*axis)(struct sway_seat *seat, struct wlr_event_pointer_axis *event);
 	void (*rebase)(struct sway_seat *seat, uint32_t time_msec);
 	void (*end)(struct sway_seat *seat);
@@ -257,6 +258,8 @@ void seatop_button(struct sway_seat *seat, uint32_t time_msec,
  */
 void seatop_motion(struct sway_seat *seat, uint32_t time_msec,
 		double dx, double dy);
+
+void seatop_touch_down(struct sway_seat *seat, uint32_t time_msec);
 
 void seatop_axis(struct sway_seat *seat, struct wlr_event_pointer_axis *event);
 
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -615,7 +615,6 @@ static void handle_axis(struct sway_seat *seat,
  * ----------------------------------*/
 
 static void handle_touch_down(struct sway_seat *seat, uint32_t time_msec) {
-	struct seatop_default_event *e = seat->seatop_data;
 	struct sway_cursor *cursor = seat->cursor;
 
 	// Determine what's under the cursor
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -610,6 +610,35 @@ static void handle_axis(struct sway_seat *seat,
 	}
 }
 
+/*-------------------------------------\
+ * Functions used by handle_touch_down /
+ * ----------------------------------*/
+
+static void handle_touch_down(struct sway_seat *seat, uint32_t time_msec) {
+	struct seatop_default_event *e = seat->seatop_data;
+	struct sway_cursor *cursor = seat->cursor;
+
+	// Determine what's under the cursor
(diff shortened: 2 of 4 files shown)


sway/input/seatop_default.c
```diff
@@ -610,6 +610,34 @@ static void handle_axis(struct sway_seat *seat,
...
+	enum wlr_edges edge =
+			cont ? find_edge(cont, surface, cursor) : WLR_EDGE_NONE;
+	bool on_border = edge != WLR_EDGE_NONE;
+	bool on_titlebar = cont && !on_border && !surface;
+
+	if (on_titlebar) {
+		sway_log(SWAY_ERROR, "Touch Event on Titlebar");
+		node = seat_get_focus_inactive(seat, &cont->node);
+		seat_set_focus(seat, node);
+		return;
```
Xyene, line 637: Don't need this `return`, since this is the end of the function anyway.

sway/input/seatop_default.c
```diff
@@ -610,6 +610,34 @@ static void handle_axis(struct sway_seat *seat,
...
+
+	struct sway_container *cont =
+			node && node->type == N_CONTAINER ? node->sway_container : NULL;
+	enum wlr_edges edge =
+			cont ? find_edge(cont, surface, cursor) : WLR_EDGE_NONE;
+	bool on_border = edge != WLR_EDGE_NONE;
+	bool on_titlebar = cont && !on_border && !surface;
+
+	if (on_titlebar) {
+		sway_log(SWAY_ERROR, "Touch Event on Titlebar");
```
Xyene, line 634: This isn't an error :)

include/sway/input/seat.h
```diff
@@ -17,6 +17,7 @@ struct sway_seatop_impl {
 			enum wlr_button_state state);
 	void (*motion)(struct sway_seat *seat, uint32_t time_msec,
 			double dx, double dy);
+	void (*touch_down)(struct sway_seat *seat, uint32_t time_msec);
```
Xyene, line 20: Just in terms of placement, this is not a good line to put this on. `button`/`motion`/`axis` all refer to pointers, so having a touch handler in the middle is a bit weird. Same with the decl of the prototype, and `.touch_down = handle_touch_down,`.

Xyene (changes requested):
I'd double-check that https://github.com/swaywm/sway/pull/5229 doesn't accomplish the same thing, in a different way. This PR adds a seatop for touch down, but I suspect applications that don't bind touch won't receive any event. https://github.com/swaywm/sway/pull/5229 adds pointer simulation to surfaces that don't bind touch, which by extension includes titlebars.

ArenM:
You're right, #5229 does have the the same effect.

Does it make sense to reuse this code to get moving windows, to work with that, using mod + dragging a window, for example?

Xyene:
I suspect that #5229 enables mod + dragging a window that doesn't bind touch (like a Wayland-native terminal, e.g. alacritty or kitty), but not one that does (e.g. I expect `gedit`). This would be good to confirm.

If that's the case, then adding a seatop handler for touch should allow for that case too. That's roughly the stage tablet support is at currently. The motion handler, for instance, [can be shimmed](https://github.com/swaywm/sway/pull/5254/files#diff-57d5973ad58d843753d353531e8cb85eR1461) for seatops that don't care about the distinction between pointer and touch motion.

David96:
@Xyene yes it does, the position is not updated on motion though, only after lifting the finger, I have to check that out.

Alacritty btw does seem to bind to touch, but not use it (https://github.com/alacritty/alacritty/blob/33abfe34a86863958e70a6b5109eab5740a6bc81/alacritty/src/event.rs#L619, I think it's simply part of the library they are using) - kitty doesn't though, that's what I'm using to test the cursor simulation stuff.

emersion:
>Alacritty btw does seem to bind to touch, but not use it

We can't do anything about these clients. Would you mind opening a winit issue to let them know about this issue?

Xyene:
> Alacritty btw does seem to bind to touch, but not use it

That's good to know, I made a false assumption here that because it didn't bind tablet it would also not be binding touch, but that's not the case.

David96:
> We can't do anything about these clients. Would you mind opening a winit issue to let them know about this issue?

Done: https://github.com/rust-windowing/winit/issues/1555

PR #5270, opened 2020-04-28
Redirect questions to wiki/IRC

emersion:
The issue tracker has already enough traffic with bug reports and
feature requests.


RedSoxFan:
Thanks!

PR #5281, opened 2020-04-30
Fix typo in swaybar-protocol.7.scd

jaehl:


swaybar/swaybar-protocol.7.scd
```diff
@@ -7,8 +7,8 @@ swaybar-protocol - JSON status line protocol for swaybar
 # SUMMARY
 
 swaybar defines the following JSON protocol to specify what information is
-displayed in the status line on the right side of swaybar. The protocol is
-compromised of a header, that is a JSON object, followed by an infinite JSON
+displayed in the status line on the right side of swaybar. The protocol
+compromises a header, that is a JSON object, followed by an infinite JSON
```
emersion, line 11: Hmm, I'm no native speaker, but I don't think "compromises" is the right word here. Maybe "includes" would be better?

jaehl: Whoops, I failed to actually change the original word. It should be "comprises".

PR #5282, opened 2020-04-30
build: improve feature summary

emersion:
Use the Meson summary() built-in instead of manually formatting a
message.

wlroots already depends on Meson 0.54.0.


ddevault:
Thanks!

PR #5287, opened 2020-05-01
Grimshot improvements

WhyNotHugo:
I've had a pretty large script for taking screenshots for a while, and made it [its own thing](https://gitlab.com/WhyNotHugo/swaygrab) yesterday.

Someone [pointed me](https://www.reddit.com/r/swaywm/comments/gb6per/swaygrab1_an_easy_to_use_screenshot_tool_for_sway/fp57uh2/) to `grimshot`, which seems to have pretty similar goals.

This PR bring a couple of minor features to grimshot making my alternative project moot (there's really no use in having "yet another screenshot tool). I've also improved the usage output (we all forget the exact syntax to commands!), and added a man page.

While commit messages are pretty verbose, here's a quick summary:

- Use `$XDG_SCREENSHOTS_DIR` to save screenshots, otherwise, fall back to the previous behaviour.
- Dropped one dependency that didn't really work for the previous point anyway.
- Avoid screenshots taken quickly one after the other from overwriting each other.
- Show more detailed `usage` output.
- Improve handling of typos and bad commands -- it was hard to get to the help page if you forgot it was `usage`.
- Added a separate flag for notifications -- I do think the majority of the screenshot taking is done interactively, so there's no need to notify the user.
- Added a man page.


emersion:
`XDG_SCREENSHOTS_DIR` is not part of the XDG user dirs spec. I'd prefer to use a `GRIMSHOT_DIR` variable instead.

contrib/grimshot
```diff
@@ -19,29 +19,63 @@
...
   echo "Usage:"
-  echo "  grimshot copy|save win|screen|output|area [FILE]"
-  echo "Troubleshoot:"
+  echo "  grimshot [--notify] (copy|save) [win|screen|output|area|window] [FILE]"
   echo "  grimshot check"
+  echo "  grimshot usage"
+  echo ""
+  echo "Commands:"
+  echo "  copy: Copy the screenshot data into the clipboard."
+  echo "  save: Save the screenshot into a regular file."
```
jhalmen, line 48: i'd go with copy and save _to_ vs _into_

contrib/grimshot
```diff
@@ -19,29 +19,63 @@
...
+  echo "  check: Verify if required tools are intalled and exit."
```
jhalmen, line 49: typo: installed

contrib/grimshot.1.scd
```diff
@@ -0,0 +1,77 @@
...
+	Save the screenshot into a regular file. Grimshot will write images
+	files to *$XDG_SCREENSHOTS_DIR* if this is set (or defined in
+	`user-dirs.dir`), or otherwise fall back to `$XDG_PICTURES_DIR`
+
+*copy*
+	Copy the screenshot data (as image/png) into the clipboard.
+
+# SYNOPSIS
+
+Grimshot is a easy to use screenshot tool for sway. It relies on grim, slurp
```
jhalmen, line 26: an?

contrib/grimshot.1.scd
```diff
@@ -0,0 +1,77 @@
...
+# Screenshots:
+# Cmd+P: Current window
+# Cmd+Shift+p: Select region
+# Cmd+Alt+p Current output
+# Cmd+Ctrl+p Select a window
+
+bindsym Mod4+p       exec grimshot active
+bindsym Mod4+Mod1+p  exec grimshot region
+bindsym Mod4+Shift+p exec grimshot output
+bindsym Mod4+Ctrl+p  exec grimshot window
```
jhalmen, line 41: these don't seem to match with the comments above

contrib/grimshot.1.scd
```diff
@@ -0,0 +1,77 @@
...
+
+_region_
+	Allows manually selecting a rectangular region, and screenshots that.
+
+_window_
+	Allows manually selecting a single window (by clicking on it), and
+	screenshots that.
+
+_output_
+	Screenshots the currently active output.
```
jhalmen, line 62: maybe 'captures' is a better verb to be used in the target descriptions?
'to screenshoot' sounds weird

contrib/grimshot.1.scd
```diff
@@ -0,0 +1,77 @@
...
+
+bindsym Mod4+p       exec grimshot active
+bindsym Mod4+Mod1+p  exec grimshot region
+bindsym Mod4+Shift+p exec grimshot output
+bindsym Mod4+Ctrl+p  exec grimshot window
+```
+
+# TARGETS
+
+grimshot can grab a few easy to remember named targets:
```
jhalmen, line 46: i'd leave the 'easy to remember' out. 
doesn't sound right

contrib/grimshot
WhyNotHugo: Updated "Save [..] to a regular file.  
`into the clipboard` makes more sense I think.

WhyNotHugo:
Thanks for the feedback, I've addressed all the typos and doc inconsistencies.

Re: `XDG_SCREENSHOTS_DIR`: There's no solid spec for these, it just seems to be a convention.

I opted for `XDG_SCREENSHOTS_DIR` since, hopefully, other tools can follow suit and have an actual standard for this. It's also configued in the same place as all other `XDG_*_DIR`s. This approach allows configuring this either via env vars, or via a config file (the same file where all the other paths are configured too), which I think gives the most flexibility to users.

Note that the upstream tool for figuring these out actually works:
```
➜ grep SCREEN .config/user-dirs.dirs
XDG_SCREENSHOTS_DIR="$HOME/screenshots"
➜ xdg-user-dir SCREENSHOTS
/home/hugo/screenshots
```

progandy:
> Re: `XDG_SCREENSHOTS_DIR`: There's no solid spec for these, it just seems to be a convention.

What convention? I haven't seen any tool use it, web searches only show some configuration files with that variable, no code. The tool works because it accepts an ```XDG_${CUSTOM_NAME}_DIR``` that has been defined for your username. 

If you want something standardized, maybe use e.g. ```$XDG_PICTURES_DIR\screenshots```

WhyNotHugo:
The "convention" is to use `$XDG_NAME_DIR` for user directories, where `NAME` can be, well... anything?

It's a very lose convention, different apps use different made up name, and there's not hard list of "official" ones. I've seen tools use the following amongst others:

```
XDG_DESKTOP_DIR
XDG_DOWNLOAD_DIR
XDG_DOCUMENTS_DIR
XDG_PICTURES_DIR
XDG_PUBLICSHARE_DIR
```

`SCREENSHOTS` makes perfect sense for, well, screenshots.

> If you want something standardized, maybe use e.g. `$XDG_PICTURES_DIR\Screenshots`

`$XDG_PICTURES_DIR` is already used for apps that save image files in general, and by photo-management tools (and people actively used them for both). Re-using the same environment variable for two different things lacks flexibility.

If you don't like `$XDG_SCREENSHOTS_DIR`, you can just _not_ use it; the fallback is the previous location anyway: `$XDG_PICTURES_DIR`.

contrib/grimshot.1.scd
```diff
@@ -0,0 +1,77 @@
...
+bindsym Mod4+Mod1+p  exec grimshot output
+bindsym Mod4+Ctrl+p  exec grimshot window
+```
+
+# TARGETS
+
+grimshot can capture the following named targets:
+
+_active_
+	Captures the currently active window.
```
ikalnytskyi, line 49: It seems like the actual script uses `win` target name, not `active`.

WhyNotHugo: Good catch, fixed.

travankor:
Should the build style optionally compile the man page? Most people will probably download the script, but forget the man page that needs to be manually created.

WhyNotHugo:
Uhm... I can include a Makefile for it, and move `grimshot`, `grimshot.1.scd` and the Makefile into `contrib/grimshot/` if that makes sense.

Didn't really think of it that much; for Arch in particular, I'm just creating a package for grimshot.

WhyNotHugo:
I'd wouldn't mind including a `Makefile` with the following, but given that the rest of the project uses meson, is sounds... wrong.

```
PREFIX?=/usr/local
BINDIR?=$(DESTDIR)$(PREFIX)/bin
MANDIR?=$(DESTDIR)$(PREFIX)/share/man

grimshot.1: grimshot.1.scd
	scdoc < $< > $@

all: grimshot.1

clean:
	rm -rf grimshot.1

install: all
	mkdir -p $(BINDIR) $(MANDIR)/man1
	install -m755 grimshot $(BINDIR)/grimshot
	install -m644 grimshot.1 $(MANDIR)/man1/grimshot.1

.PHONY: all clean install
```

contrib/grimshot.1.scd
```diff
@@ -0,0 +1,77 @@
...
+# OUTPUT
+
+Grimshot will always print the filename of the captured screenshot to
+stdout.
+
+# AUTHORS
+
+Maintained by Drew DeVault <sir@cmpwn.com>, who is assisted by other open
+source contributors. For more information about sway development, see
+<https://github.com/swaywm/sway>.
```
ddevault, line 73: NACK, the contrib directory is not maintained by Drew DeVault :wink:

ddevault:
I might just run the man page through scdoc in advance and drop the compiled roff file into contrib.

progandy:
> I'd wouldn't mind including a `Makefile` with the following, but given that the rest of the project uses meson, it sounds... wrong.

Maybe use a ninja build file? Meson uses it as well.

WhyNotHugo:
> I might just run the man page through scdoc in advance and drop the compiled roff file into contrib.

This would only really help users manually build from source -- I expect those to be a minority. But I've added this to meson. I'm not very experience with meson, so lemme know if I did a no-go there.

ddevault:
Don't add this to meson. We don't build things in contrib, as a rule, as part of the normal build.

WhyNotHugo:
Oh, I must've misunderstood you previous comment. Undid that change in meson.

Did you mean committing the compiled roff into vcs?

ddevault:
>Did you mean committing the compiled roff into vcs?

Yes.

Can you also [rebase this](https://git-rebase.io) to fix up all of the commit soup? Otherwise this LGTM.

WhyNotHugo:
Rebased into sensible commits. Or did you want it to be just one?

ddevault:
This commit log looks good to me. Are you going to commit the roff?

WhyNotHugo:
Done!

ddevault:
Thanks!

WhyNotHugo:
Thanks for the feedback and merging this in!

eoli3n:
Is there a way to be able to change default screenshots filenames ?
Because i use curl to upload screenshots and default grimshot filename contains ``,`` which is problematic.
```
$ curl -F "file=@2020-11-15T16:41:25,109868379+01:00.png" 'https://x0.at/'
curl: (26) Failed to open/read local data from file/application

$ mv 2020-11-15T16:41:25,109868379+01:00.png test.png

$ curl -F "file=@test.png" 'https://x0.at/'
https://x0.at/lSy.png
```

WhyNotHugo:
Hmm... might make sense to make that configurable.

Maybe an environment variable with the command to run to generate the file name...?

eoli3n:
> Maybe an environment variable with the command to run to generate the file name...?
Why not

rieje:
It's just minor refactoring. But for `active` window:

`swaymsg -t get_tree | jq -r 'recurse(.nodes[]?, .floating_nodes[]?) | select(.focused)' | jq -r '.rect | "\(.x),\(.y) \(.width)x\(.height)"'`

can just be 

`swaymsg -t get_tree | jq -j '.. | select(.type?) | select(.focused).rect | "\(.x),\(.y) \(.width)x\(.height)"'`?

Issue #4769, opened 2019-11-29
Repaint scheduling documentation unclear


kennylevinsen:
The documentation of max_render_time is currently unclear as to the purpose and interaction between view and output max_render_time.

For example, a view max_render_time is only active if an output max_render_time is enabled, but the man page only says that this is needed for "optimal latency".

PR #5289, opened 2020-05-01
Simplify repaint scheduling documentation

kennylevinsen:
The current documentation for repaint scheduling is very technical and
somewhat confusing.

Closes: https://github.com/swaywm/sway/issues/4769

### DIFF
--- a/sway/sway-output.5.scd
+++ b/sway/sway-output.5.scd
@@ -117,22 +117,17 @@ must be separated by one space. For example:
 	(ie. blank the screen but keep workspaces as-is), one can set DPMS to off.
 
 *output* <name> max_render_time off|<msec>
-	When set to a positive number of milliseconds, enables delaying output
-	rendering to reduce latency. The rendering is delayed in such a way as
-	to leave the specified number of milliseconds before the next
-	presentation for rendering.
+	Controls when sway composites the output, as a positive number of
+	milliseconds before the next display refresh. A smaller number leads to
+	fresher composited frames and lower perceived input latency, but if set too
+	low, sway may not finish compositing in time for display refresh, leading to
+	delayed frames.
 
-	The output rendering normally takes place immediately after a
-	presentation (vblank, buffer flip, etc.) and the frame callbacks are
-	sent to surfaces immediately after the rendering to give surfaces the
-	most time to draw their next frame. This results in slightly below 2
-	frames of latency between the surface rendering and committing new
-	contents, and the contents being shown on screen, on average. When the
-	output rendering is delayed, the frame callbacks are sent immediately
-	after presentation, and the surfaces have a small timespan (1 /
-	(refresh rate) - max_render_time) to render and commit new contents to
-	be shown on the next presentation, resulting in below 1 frame of
-	latency.
+	When set to off, sway composites immediately after display refresh,
+	maximizing time available for compositing.
+
+	To adjust when applications are instructed to render, see *max_render_time*
+	in *sway*(5).
 
 	To set this up for optimal latency:
 	. Launch some _full-screen_ application that renders continuously, like
@@ -140,14 +135,8 @@ must be separated by one space. For example:
 	. Start with *max_render_time 1*. Increment by *1* if you see frame
 	  drops.
 
-	To achieve even lower latency, see the *max_render_time* surface
-	property in *sway*(5).
-
-	Note that this property has an effect only on backends which report the
-	presentation timestamp and the predicted output refresh rate—the DRM
-	and the Wayland backends. Furthermore, under the Wayland backend the
-	optimal max_render_time value may vary based on the parent compositor
-	rendering timings.
+	This setting only has an effect on Wayland and DRM backends, as support for
+	presentation timestamps and predicted output refresh rate is required.
 
 *output* <name> adaptive_sync on|off
 	Enables or disables adaptive synchronization (often referred to as Variable
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -186,20 +186,28 @@ set|plus|minus <amount>
 	Cycles the layout mode of the focused container through a list of layouts.
 
 *max_render_time* off|<msec>
-	Works together with *output max_render_time* to reduce the latency even
-	further by delaying the frame callbacks sent to a surface. When set to
-	a positive number of milliseconds, delays the frame callback in such a
-	way that the surface has the specified number of milliseconds to render
-	and commit new contents before being sampled by the compositor for the
-	next presentation. See *max_render_time* in *sway-output*(5) for
-	further details.
+	Controls when the relevant application is told to render this window, as a
+	positive number of milliseconds before the next time sway composites the
+	output. A smaller number leads to fresher rendered frames being composited
+	by sway and lower perceived input latency, but if set too low, the
+	application may not finish rendering before sway composites the output,
+	leading to delayed frames.
+
+	When set to off, the relevant application is told to render this window
+	immediately after display refresh. How much time is left for rendering
+	before sway composites the output at that point depends on the output
+	*max_render_time* setting.
 
 	To set this up for optimal latency:
-	. Set up *output max_render_time*.
+	. Set up *output max_render_time* (see *sway-output*(5)).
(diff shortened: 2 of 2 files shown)


sway/sway-output.5.scd
```diff
@@ -117,37 +117,29 @@ must be separated by one space. For example:
...
-	most time to draw their next frame. This results in slightly below 2
-	frames of latency between the surface rendering and committing new
-	contents, and the contents being shown on screen, on average. When the
-	output rendering is delayed, the frame callbacks are sent immediately
-	after presentation, and the surfaces have a small timespan (1 /
-	(refresh rate) - max_render_time) to render and commit new contents to
-	be shown on the next presentation, resulting in below 1 frame of
-	latency.
+	Controls when sway composites the output, as a positive number of
+	milliseconds before next vblank. A smaller number leads to fresher
```
travankor, line 121: vblank still sounds technical

sway/sway.5.scd
```diff
@@ -186,21 +186,28 @@ set|plus|minus <amount>
...
-	next presentation. See *max_render_time* in *sway-output*(5) for
-	further details.
+	Controls when the relevant application is told to render this window, as a
+	positive number of milliseconds before next time sway composites the output.
+	A smaller number leads to fresher rendered frames being composited by sway
+	and lower perceived input latency, while a larger number gives the
+	application more time to render and lowers the likelihood of delayed frames.
+
+	When set to off, the relevant application is told to render this window
+	immediately after the output has been composited, to maximize the time
```
travankor, line 196: Does it use the output `max_render_time`? Or are they not related at all?

sway/sway-output.5.scd
```diff
@@ -117,37 +117,29 @@ must be separated by one space. For example:
...
-
-	The output rendering normally takes place immediately after a
-	presentation (vblank, buffer flip, etc.) and the frame callbacks are
-	sent to surfaces immediately after the rendering to give surfaces the
-	most time to draw their next frame. This results in slightly below 2
-	frames of latency between the surface rendering and committing new
-	contents, and the contents being shown on screen, on average. When the
-	output rendering is delayed, the frame callbacks are sent immediately
-	after presentation, and the surfaces have a small timespan (1 /
-	(refresh rate) - max_render_time) to render and commit new contents to
```
travankor, line 133: Is using zero is equivalent to having it off?

sway/sway-output.5.scd
```diff
@@ -117,37 +117,29 @@ must be separated by one space. For example:
...
-	be shown on the next presentation, resulting in below 1 frame of
-	latency.
+	Controls when sway composites the output, as a positive number of
+	milliseconds before next vblank. A smaller number leads to fresher
+	composited frames and lower perceived input latency, while a larger number
+	gives sway more time to composite and lowers the likelihood of delayed
+	frames.
```
travankor, line 124: "fresher composited frames"  and "lower likelihood of delayed frames" doesn't seem like a good way to differentiate the two.

sway/sway-output.5.scd
emersion: "Display refresh"?

sway/sway-output.5.scd
kennylevinsen: It's the relevant (only?) pro and con. You lower it to get fresher frames, you increase it to avoid frames that missed vblank and got delayed till next display refresh.

Open to suggestions.

sway/sway-output.5.scd
kennylevinsen: The `max_render_time` commands to not accept zero as a value, only positive integers greater than zero and the word `off`.

(Internally, `off` maps to zero.)

sway/sway.5.scd
kennylevinsen: Output `max_render_time` sets when sway composites the output, which view `max_render_time` is relative to.

However, I do see that I have missed an intricacy in view `max_render_time` when set to zero, in which case frame callbacks are sent on display refresh, while if set to a positive value, it is relative to composite.

I'll need to double-check why that is, and add this to the documentation.

kennylevinsen:
Changes:
1. vblank -> display refresh
2. Now covers the "corner-case" of view max_render_time = off, where frame callbacks are issued after vblank regardless of when compositing occurs.
3. Uses a bit different wording for the downside of low values.
4. The summary of view max_render_time in the output section was replaced by a reference, as this corner case ended up making it a bit too verbose.

sway/sway.5.scd
```diff
@@ -186,21 +186,29 @@ set|plus|minus <amount>
...
 
 	To set this up for optimal latency:
-	. Set up *output max_render_time*.
+	. Set up *output max_render_time* (see *sway-output*(5)).
 	. Put the target application in _full-screen_ and have it continuously
 	  render something.
 	. Start by setting *max_render_time 1*. If the application drops
 	  frames, increment by *1*.
 
+	This setting only has an effect if a per-output *max_render_time* is in
```
travankor, line 208: Applications need to support the presentation time protocol for this setting to work, right? Adding this to applications without presentation time support will do nothing?

sway/sway-output.5.scd
travankor: higher input latency if too high vs dropped frames if set too low?

sway/sway-output.5.scd
```diff
@@ -117,37 +117,26 @@ must be separated by one space. For example:
...
+	low, sway may not finish compositing in time for display refresh, leading to
+	delayed frames.
+
+	When set to off, sway composites immediately after display refresh,
+	maximizing time available for compositing.
+
+	To adjust when applications are instructed to render, see *max_render_time*
+	in *sway*(5).
 
 	To set this up for optimal latency:
```
travankor, line 132: Is this example correct? Can't different applications give different results?

sway/sway-output.5.scd
kennylevinsen: For a user to see a smooth framerate, the application has to be rendering smoothly, which is outside of our control. The opposite of delayed frames is just *not* delaying frames/showing frames on time.

Sway and wayland never *drops* frames. That's up to the application. If you miss the deadline, they just end up being one frame period late.

(Maybe I'm too stubborn on the semantics - if there ends up being consensus for "dropped", we can take that instead.)

sway/sway.5.scd
kennylevinsen: No, this affects frame callbacks. The presentation time protocol informs you when your frame ended up being shown on screen, and is unaffected.

(The previous docs noted that the *backend* needs to have some form of presentation feedback and ability to estimate next frame time for any of this to work at all, but I simplified that to just say that it only works for wayland and drm.)

sway/sway-output.5.scd
kennylevinsen: Do you mean the example calibration routine?

What you are benchmarking here is not application render time, but sway composite time. Having something like glxgear fullscreen (assuming it is not direct scanout capable), mans that sway is copying the entire screen. Having many instances of glxgears so that sway is copying many small sections would be more work, but unsure how much it would affect timings.

This PR is mostly about simplifying the documentation. I have not touched the calibration routines.

sway/sway.5.scd
travankor: So do applications like mpv that use presentation time benefit from more configurable frame callbacks?

sway/sway-output.5.scd
travankor: Maybe desynchronized or inconsistent frames as alternatives? The phrasing sounded weird with the original commit, but it is fine now after reading it over.

sway/sway.5.scd
kennylevinsen: Frame callbacks are not configurable from the application perspective. Either it asks for them or it doesn't. You can use per-view `max_render_time` to control when sway will send them.

An application like mpv could use presentation time to know how long it took after it submitted its frame until it actually ended up on screen. Then on next frame callback, it could compensate for the delay by sending video frames from further ahead in the video, so everything lines up correctly when display finally shows the content.

(It is also possible to not use frame callbacks at all, but using only presentation time would likely lead to delayed frames as it does not expose the composition time, only presentation time.)

PR #5292, opened 2020-05-02
Fix resizing floating windows at top border

David96:
Adding the container_titlebar_height here causes the window height to
instantly add the titlebar height to the window when resizing is
started.

### DIFF
--- a/sway/input/seatop_resize_floating.c
+++ b/sway/input/seatop_resize_floating.c
@@ -76,7 +76,6 @@ static void handle_motion(struct sway_seat *seat, uint32_t time_msec,
 	struct sway_container_state *state = &con->current;
 	width += state->border_thickness * 2;
 	height += config->titlebar_border_thickness * 2;
-	height += container_titlebar_height();
 	height += config->titlebar_v_padding;
 
 	// Recalculate these, in case we hit a min/max limit


emersion:
Was added in https://github.com/swaywm/sway/pull/5250. 

cc @adee: thoughts?

David96:
Just try resizing a window at the top border, it will instantly add the height of the titlebar to the window height. I tried that with and without titlebar and in both cases it happened.
I'm also not sure about adding any value that comes from the config here because as I understand it the state of the current container could be completely different from what's written to the config (changed by swaymsg or whatever)? But I haven't noticed any bugs related to that yet so I only removed the one line that definitely caused buggy behaviour for me.

adee:
container_titlebar_height() is used in render_titlebar() as well, so no height change with swaymsg will be reflected in the rendering either. There are probably multiple small issues regarding border size and resize not resolved by my change. 
Maybe better to revert https://github.com/swaywm/sway/pull/5250 for now until a more thorough investigation can be done.

adee:
Okay. I've begun to rework this. I'll open a new pull request. Please don't merge this.
I'll do more thorough testing this time.

David96:
Closing in favor of #5299

PR #5295, opened 2020-05-02
Don't unhide cursor on touch events

David96:
Touch events hide the cursor so unhiding it again only causes it to
flicker (and not hide at all).

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -268,7 +268,7 @@ void cursor_handle_activity(struct sway_cursor *cursor,
 			cursor->hide_source, cursor_get_timeout(cursor));
 
 	seat_idle_notify_activity(cursor->seat, idle_source);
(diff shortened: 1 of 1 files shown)


emersion:
We should probablu guard the `wl_event_source_timer_update` call too right? This would make it so the timer isn't started on touch events.

David96:
Not sure, but I think that doesn't really matter? If the cursor gets hidden by a touch event, it only gets unhidden on a pointer event (mouse, tablet tool, whatever) which in turn then updates the timer anyway.
But if you think it's more correct I can add that of course.

emersion:
>Not sure, but I think that doesn't really matter?

I just wanted to guard all of the "hide cursor timeout" logic and avoid having a running timer when touch hides the cursor, to make the code easier to reason about. But it doesn't matter that much indeed.

PR #5299, opened 2020-05-02
Really fix floating window border resize problems

adee:
Same problem, different approach.
Fixes: https://github.com/swaywm/sway/pull/5250

I've done testing of all combinations of settings and runtime border changes with resize I could come up with. So far I'm unable to see it brake. More testing is welcome. 

@David96 : please test this instead https://github.com/swaywm/sway/pull/5292, it should resolve your problems.

### DIFF
--- a/sway/input/seatop_resize_floating.c
+++ b/sway/input/seatop_resize_floating.c
@@ -7,6 +7,7 @@
 #include "sway/tree/arrange.h"
 #include "sway/tree/view.h"
 #include "sway/tree/workspace.h"
+#include "sway/tree/container.h"
 
 struct seatop_resize_floating_event {
 	struct sway_container *con;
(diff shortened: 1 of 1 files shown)


David96:
I can confirm this also fixes the issue I had.

emersion:
Let me know when this is ready for review.

adee:
Thanks for the reminder. Ready for review now.

sway/input/seatop_resize_floating.c
```diff
@@ -55,29 +56,37 @@ static void handle_pointer_motion(struct sway_seat *seat, uint32_t time_msec,
...
+
 	// Determine new width/height, and accommodate for floating min/max values
 	double width = e->ref_width + grow_width;
 	double height = e->ref_height + grow_height;
 	int min_width, max_width, min_height, max_height;
 	floating_calculate_constraints(&min_width, &max_width,
 			&min_height, &max_height);
-	width = fmax(min_width, fmin(width, max_width));
-	height = fmax(min_height, fmin(height, max_height));
+	width = fmax(min_width + border_width, fmin(width, max_width));
```
emersion, line 78: We should probably take `border_width` into account for the max too? Something like `max_width - border_width`? We should also be careful that the result is > 0.

emersion: Let's merge this PR as-is. I've created a follow-up issue: https://github.com/swaywm/sway/issues/5337

Issue #5268, opened 2020-04-28
Pointer constraints are applied inconsistently to tablet pen input


Xyene:
This issue is largely an RFC regarding what sway should do regarding tablet/touch pointer emulation in the face of pointer constraints. Any input or feedback would be much appreciated :)

On non-tablet-v2 surfaces, pen input is emulated as pointer input, it currently reuses the existing pointer code. As a side-effect, this makes tablet input over a non-v2 surface constrained by `pointer-constraints-unstable-v1` confines.

However, this is pretty surprising, and leads to seemingly buggy (though actually expected) behavior. Consider the below window layout: `gedit` is on the left and binds tablet, `kitty` is in the top right and does not, and the game osu! (which sets up a pointer confine, and does not bind tablet) is in the bottom right.

![](https://user-images.githubusercontent.com/1403503/80442378-bf2f0800-88da-11ea-9676-ab2949e387e6.png)

If the game window is focused, input via mouse cannot leave the window. However, tablet input _can_ leave the window through the left side (because it hits the tablet v2 handler for `gedit`), but it can't leave through the top (because `kitty` requires pointer emulation as well, so the pen input remains confined).

This is inconsistent. I think that either pointer constraints should be uniformly applied to tablet (and touch!) input, or not at all.

**Pros of confining tablet input:**
* An application that _only_ expects pointer events to come from inside whatever confine it set would become confused by the emulated pointer events, which could land outside its confine. Hypothetically it could seriously misbehave, fail an `assert`, or something similar. I don't know of any such application, though it's _possible_ to write one that does.
* Not confining is consistent with `XGrabPointer` from X11, and XWayland translates `XGrabPointer` into a pointer constraint. XWayland apps that expected to grab all input, including tablet input, will fail to do so.

**Cons of confining tablet input:**
* It's semantically wrong; the pointer constraint spec is explicitly about pointers. See also [this](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/16).
* GNOME doesn't confine tablet input, so it would be inconsistent between compositors.

Neither option seems flawless. A middle ground might be to not confine the tablet input, but to stop sending events of any form to the surface owning the confine if the pen has left it. That would solve the issue of the pathological application that crashes... but it still doesn't feel like _great_ solution. 

Either implementation would be easy enough to do, but I think sway should pick one way and stick with it.

emersion:
[...]

PR #5300, opened 2020-05-02
Fix up pointer constraint interaction with tablet input

Xyene:
Fixes https://github.com/swaywm/sway/issues/5268.

The less-obvious part of this change is in the second commit, with the (unexpected) problem described in its message. By my understanding of the pointer constraint spec this should be correct, but I wouldn't bet on my understanding :)

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -292,16 +292,19 @@ void cursor_unhide(struct sway_cursor *cursor) {
 static void pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
 		struct wlr_input_device *device, double dx, double dy,
 		double dx_unaccel, double dy_unaccel) {
-	cursor_handle_activity(cursor, IDLE_SOURCE_POINTER);
+	if (device->type == WLR_INPUT_DEVICE_POINTER) {
+		cursor_handle_activity(cursor, IDLE_SOURCE_POINTER);
+	}
 
 	wlr_relative_pointer_manager_v1_send_relative_motion(
 		server.relative_pointer_manager,
 		cursor->seat->wlr_seat, (uint64_t)time_msec * 1000,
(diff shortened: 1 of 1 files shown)


sway/input/cursor.c
```diff
@@ -743,11 +746,11 @@ static void handle_tool_button(struct wl_listener *listener, void *data) {
 		(enum zwp_tablet_pad_v2_button_state)event->state);
 }
 
-static void check_constraint_region(struct sway_cursor *cursor) {
+static void check_constraint_region(struct sway_cursor *cursor, bool do_warp) {
```
Xyene, line 749: Note for reviewer: this flag is a bit of an awkward way to do this, in theory the entire warp logic could be moved to line 1285 (which is the only caller with `do_wrap = true`), but I opted for this way to minimize future churn, since the call on line 1285 is a workaround for another issue.

emersion:
>This commit addresses this behavior by only warping on the first surface commit since the pointer constraint was added.

Hmm, I don't think this is correct. If the pointer is confined and the surface changes the confine region, we should probably wrap.

Xyene:
I think you're right, I missed that there's a `set_region` request and assumed a new region would create a new confine, but that's not the case.

Part of the issue here is that wlroots handles confine regions transparently,  and doesn't explicitly notify sway when they change. Instead, since it has no information to the contrary, sway performs warping logic on every surface commit regardless of if the region has changed.

I think a solution here would be to make wlroots signal sway when the region changes, so sway doesn't have to warp every commit. I'll take a look at doing that in a separate patch, and base this one off that if it ends up working.

Xyene:
Alright, I've split off the pointer region warping into https://github.com/swaywm/wlroots/pull/2191 and https://github.com/swaywm/sway/pull/5334; this PR contains just the remaining logic to allow tablet input to leave the confine. With the other two commits applied, the issue remains resolved.

PR #5303, opened 2020-05-04
input/tablet: add seatop_down entry for tablet input

### LINKED ISSUE #5302: Tablet implicit grab stops sending motion events when outside surface
For example, this is problematic when trying to drag a scrollbar, and exiting the window &mdash; the scrollbar motion stops. Additionally, without `focus_follows_mouse no`, the tablet passes focus to whatever surface it goes over regardless of if there is an active implicit.

If the tablet is over a surface that does not bind tablet handlers, sway will fall back to pointer emulation, and all of this works fine. It probably should have consistent behavior between emulated and not-emulated input, though.

Xyene:
Depends on https://github.com/swaywm/wlroots/pull/2158.

Currently, when tablet input exits a window during an implicit grab, it
passes focus to another window.

For instance, this is problematic when trying to drag a scrollbar, and
exiting the window &mdash; the scrollbar motion stops. Additionally,
without `focus_follows_mouse no`, the tablet passes focus to whatever
surface it goes over regardless of if there is an active implicit grab.

If the tablet is over a surface that does not bind tablet handlers, sway
will fall back to pointer emulation, and all of this works fine. It
probably should have consistent behavior between emulated and
not-emulated input, though.

This commit adds a condition for entering `seatop_down` when a tablet's
tool tip goes down, and exiting when it goes up. Since events won't be
routed through `seatop_default`, this prevents windows losing focus during
implicit grabs.

Closes #5302.

Some more logic should be refactored into these new handlers, but I feel
that is best left outside the scope of this PR.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -572,7 +572,19 @@ static void handle_tablet_tool_position(struct sway_cursor *cursor,
 	struct sway_seat *seat = cursor->seat;
 	node_at_coords(seat, cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
 
-	if (surface && wlr_surface_accepts_tablet_v2(tablet->tablet_v2, surface)) {
+	// The logic for whether we should send a tablet event or an emulated pointer
+	// event is tricky. It comes down to:
+	// * If we began a drag on a non-tablet surface (simulated_tool_tip_down),
+	//   then we should continue sending emulated pointer events regardless of
+	//   whether the surface currently under us accepts tablet or not.
+	// * Otherwise, if we are over a surface that accepts tablet, then we should
+	//   send tablet events.
+	// * If we began a drag over a tablet surface, we should continue sending
+	//   tablet events until the drag is released, even if we are now over a
+	//   non-tablet surface.
+	if (!cursor->simulated_tool_tip_down &&
+			((surface && wlr_surface_accepts_tablet_v2(tablet->tablet_v2, surface)) ||
+				wlr_tablet_tool_v2_has_implicit_grab(tool->tablet_v2_tool))) {
 		seatop_tablet_tool_motion(seat, tool, time_msec, dx, dy);
 	} else {
 		wlr_tablet_v2_tablet_tool_notify_proximity_out(tool->tablet_v2_tool);
@@ -680,6 +692,8 @@ static void handle_tool_tip(struct wl_listener *listener, void *data) {
 
 		wlr_tablet_v2_tablet_tool_notify_up(sway_tool->tablet_v2_tool);
 	}
+
+	seatop_tablet_tool_tip(seat, sway_tool, event->time_msec, event->state);
 }
 
 static struct sway_tablet *get_tablet_for_device(struct sway_cursor *cursor,
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -23,6 +23,8 @@ struct sway_seatop_impl {
 	void (*rebase)(struct sway_seat *seat, uint32_t time_msec);
 	void (*tablet_tool_motion)(struct sway_seat *seat,
 			struct sway_tablet_tool *tool, uint32_t time_msec, double dx, double dy);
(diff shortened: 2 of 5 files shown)


sway/input/cursor.c
```diff
@@ -544,7 +544,19 @@ static void handle_tablet_tool_position(struct sway_cursor *cursor,
...
+	//   then we should continue sending emulated pointer events regardless of
+	//   whether the surface currently under us accepts tablet or not.
+	// * Otherwise, if we are over a surface that accepts tablet, then we should
+	//   send tablet events.
+	// * If we began a drag over a tablet surface, we should continue sending
+	//   tablet events until the drag is released, even if we are now over a
+	//   non-tablet surface.
+	if (!cursor->simulated_tool_tip_down &&
+			((surface && wlr_surface_accepts_tablet_v2(tablet->tablet_v2, surface)) ||
+				wlr_tablet_tool_v2_has_implicit_grab(sway_tool->tablet_v2_tool))) {
```
Xyene, line 559: Note for reviewer: this looks a lot better in an editor that renders tabs properly. I'm open to suggestions, though.

sway/input/cursor.c
```diff
@@ -544,7 +544,19 @@ static void handle_tablet_tool_position(struct sway_cursor *cursor,
...
+	// The logic for whether we should send a tablet event or an emulated pointer
+	// event is tricky. It comes down to:
+	// * If we began a drag on a non-tablet surface (simulated_tool_tip_down),
+	//   then we should continue sending emulated pointer events regardless of
+	//   whether the surface currently under us accepts tablet or not.
+	// * Otherwise, if we are over a surface that accepts tablet, then we should
+	//   send tablet events.
+	// * If we began a drag over a tablet surface, we should continue sending
+	//   tablet events until the drag is released, even if we are now over a
+	//   non-tablet surface.
```
emersion, line 556: Can't we send events in the `tablet_tool_tip` handlers instead?

emersion:
Needs rebase.

sway/input/cursor.c
Xyene: Hmm, I'm not sure I follow... do you mean within the seatop handlers for `tablet_tool_tip`? We do send the actual motion events in the seatop handler for `tablet_tool_motion`, but deciding whether to emulate input as pointer events is a task that seems best left to `cursor.c`, so that all the special-casing can happen in one place as opposed to over different seatop implementations.

That leaves the seatops receiving only pointer and tablet input, without having to worry about whether it's _emulated_ pointer input they're receiving, or "real" input.

Xyene:
Rebased, CI failure appears unrelated.

sway/input/cursor.c
```diff
@@ -680,6 +692,8 @@ static void handle_tool_tip(struct wl_listener *listener, void *data) {
 
 		wlr_tablet_v2_tablet_tool_notify_up(sway_tool->tablet_v2_tool);
 	}
+
+	seatop_tablet_tool_tip(seat, sway_tool, event->time_msec, event->state);
```
emersion, line 696: Shouldn't this be inside the condition above, right after `seatop_tablet_tool_motion`?

emersion: Err, my bad.

sway/input/cursor.c
emersion: I agree that emulating pointer events is something that should be done in `cursor.c`. However calling functions that directly send events like `wlr_tablet_v2_tablet_tool_notify_proximity_out` is a little fragile here. The seatop may or may not forward events to `wlr_seat`, in case it doesn't, we shouldn't notify proximity out.

But I see we don't have `sway_seatop_impl.tablet_tool_proximity_out` yet, so my point is moot anyway. Also looking at this PR now, the diff doesn't touch the code near `wlr_tablet_v2_tablet_tool_notify_proximity_out` (though maybe that has changed with the rebase? can't remember).

tl;dr LGTM!

emersion:
> CI failure appears unrelated

Indeed, a job restart fixed it.

PR #5307, opened 2020-05-05
Add views idle inhibition status in get_tree output

### LINKED ISSUE #5286: Add idle inhibition status in a swaymsg command response
Here is my use case: I'd like to display in the status bar whether the idle state is currently inhibited or not (given the current state of the tree).

However, at the moment, the idle info is nowhere to be found in any of the swaymsg command response.

There are several options I could think of to add this info:

* Add a new command like `swaymsg -t get_idle_status` or so that would directly return whether it is inhibited as a whole or not.

* In the get_tree response add a new "inhibit_idle" key for containers that would store a boolean.

* Same as previous option but make the "inhibit_idle" key contain any of the possible values ("focus", "fullscreen", "none",...) it was set with.

All those options should make some scripting possible around this idle info.

tardypad:
Fixes #5286

Following command is an example of scripting to determine whether any view is currently inhibiting the idle state.
Exits with 0 if idle is inhibited, >0 otherwise.
```
swaymsg -t get_tree |
jq -e 'recurse(.floating_nodes[]?,.nodes[]?) | select(.inhibit_idle == true)' > /dev/null
```

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -18,6 +18,7 @@
 #include <wlr/types/wlr_output.h>
 #include <xkbcommon/xkbcommon.h>
 #include "wlr-layer-shell-unstable-v1-protocol.h"
+#include "sway/desktop/idle_inhibit_v1.h"
 
 static const int i3_output_id = INT32_MAX;
 static const int i3_scratch_id = INT32_MAX - 1;
@@ -133,6 +134,22 @@ static const char *ipc_json_xwindow_type_description(struct sway_view *view) {
 	return "unknown";
 }
 #endif
+
+static const char *ipc_json_user_idle_inhibitor_description(enum sway_idle_inhibit_mode mode) {
+	switch (mode) {
+	case INHIBIT_IDLE_FOCUS:
+		return "focus";
+	case INHIBIT_IDLE_FULLSCREEN:
+		return "fullscreen";
+	case INHIBIT_IDLE_OPEN:
+		return "open";
+	case INHIBIT_IDLE_VISIBLE:
+		return "visible";
+	case INHIBIT_IDLE_APPLICATION:
+		return NULL;
+	}
+	return NULL;
+}
 
 json_object *ipc_json_get_version(void) {
 	int major = 0, minor = 0, patch = 0;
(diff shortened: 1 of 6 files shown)


emersion:
That sounds like a complicated JSON scheme. Is this because you plan to add more features in the future?

RedSoxFan:
> That sounds like a complicated JSON scheme. Is this because you plan to add more features in the future?

Yeah sorry. It probably is overly complex for this. I got used to more complicated object-oriented serializations at work. Maybe something more like the following would be appropriate:
```
"inhibit_idle": {
  "application": true|false, (or possibly "blocked|yes|no" - "blocked" is not currently implemented)
  "user": "focus|fullscreen|open|visible|none",
  "active": true|false (or possibly "no|application|user")
}
```

This should give all the useful information, but is flatter.

tardypad:
@RedSoxFan How about the following variation:
```
"inhibit_idle": true|false,
"inhibitors": {
  "application": enabled|disabled|none,
  "user": "focus|fullscreen|open|visible|none"
}
```
First, I find it clearer when splitting the different types of info there:
- "inhibitors" only contains the config/state of each inhibitor linked to the view
- "inhibit_idle" is global to the view and reports whether it inhibits the idle state or not.

Second, I would made the inhibitors "application" possible values more similar to the "user" one. I believe it is also better matching the other tree fields (There is no field using yes|no from what I can see)

RedSoxFan:
> @RedSoxFan How about the following variation:
> ```
> "inhibit_idle": true|false,
> "inhibitors": {
>   "application": enabled|disabled|none,
>   "user": "focus|fullscreen|open|visible|none"
> }
> ```
> First, I find it clearer when splitting the different types of info there:
> - "inhibitors" only contains the config/state of each inhibitor linked to the view
> - "inhibit_idle" is global to the view and reports whether it inhibits the idle state or not.

Change `inhibitors` to `idle_inhibitors` or add idle somehow to the user/application keys and that works for me.

> Second, I would made the inhibitors "application" possible values more similar to the "user" one. I believe it is also better matching the other tree fields (There is no field using yes|no from what I can see)

:+1:

tardypad:
> Change inhibitors to idle_inhibitors or add idle somehow to the user/application keys and that works for me.

Ah right, "inhibitors" alone had no meaning... I'll name it "idle_inhibitors"

sway/ipc-json.c
```diff
@@ -134,6 +135,22 @@ static const char *ipc_json_xwindow_type_description(struct sway_view *view) {
...
 
+static const char *ipc_json_user_idle_inhibitor_description(enum sway_idle_inhibit_mode mode) {
+	switch (mode) {
+	case INHIBIT_IDLE_FOCUS:
+		return "focus";
+	case INHIBIT_IDLE_FULLSCREEN:
+		return "fullscreen";
+	case INHIBIT_IDLE_OPEN:
+		return "open";
+    case INHIBIT_IDLE_VISIBLE:
```
RedSoxFan, line 146: style: indentation - use tabs not spaces

sway/ipc-json.c
```diff
@@ -134,6 +135,22 @@ static const char *ipc_json_xwindow_type_description(struct sway_view *view) {
...
+		return "visible";
+    case INHIBIT_IDLE_APPLICATION:
```
RedSoxFan, line 148: style: indentation - use tabs not spaces

sway/sway-ipc.7.scd
```diff
@@ -379,6 +379,14 @@ node and will have the following properties:
 |- shell
 :  string
 :  (Only views) The shell of the view, such as _xdg\_shell_ or _xwayland_
+|- inhibit_idle
+:  boolean
+:  (Only views) Whether the view is inhibiting the idle state
+|- idle_inhibitors
+:  object
+:  (Only views) An object containing the state of the _application_ and _user_ idle inhibitors.
+    _application_ can be _enabled_, _disabled_ (not supported yet) or _none_.
```
RedSoxFan, line 388: Can you just drop the disabled bit instead of documenting something that isn't currently supported? If/when support for disabling an application idle inhibitor gets implemented, this can be added in.

The second example below will also need to be updated.

RedSoxFan:
Thanks!

PR #5316, opened 2020-05-08
input/keyboard: use wlr_keyboard_keymaps_match from wlroots

Xyene:
Added in swaywm/wlroots#2172, so that sway doesn't need to maintain
an independent copy of this function.

Minor space &rarr; tab conversion also included for two lines.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -5,6 +5,7 @@
 #include <wlr/backend/session.h>
 #include <wlr/interfaces/wlr_keyboard.h>
 #include <wlr/types/wlr_idle.h>
+#include <wlr/types/wlr_keyboard.h>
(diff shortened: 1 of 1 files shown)


sway/input/keyboard.c
```diff
@@ -5,6 +5,7 @@
 #include <wlr/backend/session.h>
 #include <wlr/interfaces/wlr_keyboard.h>
 #include <wlr/types/wlr_idle.h>
+#include <wlr/types/wlr_keyboard.h>
```
Xyene, line 8: Note for reviewer: already implicitly included through `wlr_keyboard_group.h` below, but probably best to make it explicit.

RedSoxFan:
Thanks!

Issue #5167, opened 2020-04-02
reload crashes sway


LamaRaz:
hardware: eeepc 1005pe (old netbook with N450/GMA3150  GL 1.4)
linux void musl
sway 1.4

default config file with just some additional key bindgings
no systemd

crash happens often but not always, in this log the first reload was successful

perhaps related: when first starting sway some garbage is displayed on the swaybar (almost like uninitialized memory) for a fraction of a second, happens always

log:

https://paste.ubuntu.com/p/3GRTSGHGx8/

emersion:
Please provide a stack trace. You can do so by compiling from source, reproducing the crash and then running `coredumpctl gdb` and then `bt full`.

Here are some instructions to compile from source: https://github.com/swaywm/sway/wiki/Development-Setup#compiling-as-a-subproject

Xyene:
I coincidentally had this happen to me just now while running `swaymsg reload` in a terminal (as a workaround for https://github.com/swaywm/sway/issues/5029, which crashes reload by keybinding). I've uploaded the log [here](https://gist.github.com/Xyene/b711db092e09f4dd60be69ef0a9b8d6f).

tchebb:
I experienced this several times on sway 1.4, and just now again using a master build built from 3078f232581d1dcd548810370c193c6d235d2e82. Here is the stack trace from the master build:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fdc836051e5 in __strlen_avx2 () from /usr/lib/libc.so.6
[Current thread is 1 (Thread 0x7fdc827d7dc0 (LWP 20354))]
(gdb) bt
#0  0x00007fdc836051e5 in __strlen_avx2 () at /usr/lib/libc.so.6
#1  0x00007fdc83510676 in __vfprintf_internal () at /usr/lib/libc.so.6
#2  0x00007fdc835115e6 in buffered_vfprintf () at /usr/lib/libc.so.6
#3  0x00007fdc8350e774 in __vfprintf_internal () at /usr/lib/libc.so.6
#4  0x0000562167f4ee6a in sway_log_stderr
    (verbosity=SWAY_DEBUG, fmt=0x562167f58ef0 "[%s:%d] running command for binding: %s", args=0x7ffc1df10ae0) at ../common/log.c:89
#5  0x0000562167f4efee in _sway_log (verbosity=SWAY_DEBUG, fmt=0x562167f58ef0 "[%s:%d] running command for binding: %s")
[...]

Issue #4975, opened 2020-02-04
Character is printed when reload configuration


Axenntio:
I have this in my configuration
```
# Reload the configuration file
	bindsym $mod+Shift+r reload
```
When I press the shortcut, sway reload the configuration, then put a `R` in my current window.
I'm on sway version 1.4

cryptoluks:
Same issue for me. Same version.

ohk2kt3t4:
Does it happen with other letters too?

RedSoxFan:
Related https://github.com/swaywm/wlroots/pull/1973
I haven't gotten around to looking into a better solution

cryptoluks:
> 
> 
> Does it happen with other letters too?

Yes, I reload my config with $mod+Shift+c and a capital "C" is printed when reloading.

declanqian:
Yeah. It happens to me too.

damneq:
Same thing here, using $mod+Shift+c and capital "C" is printed

the-antz:
Same here, for me it prints a 'C' with the default bind, version is 1.4.

dwma:
sway version 1.4, I reload through a menu by pressing the return key and 2 returns are printed.

l4tte:
I can't replicate, it's probably a config issue like @dwma suggested

cryptoluks:
Yes, I can confirm the issue is solved when removing the keyboard block. But that's obviously not an option. :smile: 
```
input type:keyboard {
    xkb_layout de
    xkb_variant nodeadkeys
    xkb_options caps:hyper
}
```

craigbarnes:
I get the same bug in 1.4 when my config contains:

```
input type:keyboard {
    xkb_options caps:ctrl_modifier,compose:ralt
    repeat_delay 350
    repeat_rate 25
}

bindsym $mod+r reload
```

It seems like it might be because `cmd_reload()` resets inputs and at some point the key is re-interpreted as being pressed, but not yet bound to anything.

The behaviour when running `swaymsg reload` from a terminal also seems wierd. It seems to insert several line feeds into the terminal (possibly due to the enter key being down when the command gets run).

schauveau:
I found a simple workaround. The idea is to bind the reload command on a key release and to also bind a fake action (nop) on the key press. 
```                                                                               
bindsym $mod+Shift+c nop                                                                            
bindsym --release $mod+Shift+c reload
```

xPMo:
Just want to add that this is not exclusive to reload.

[...]

PR #5317, opened 2020-05-08
input: reset keyboard groups keyboard on reset all

### LINKED ISSUE #5029: Crash in handle_keyboard_repeat
`$swaymsg -t get_version`
`sway version 1.4`

[sway.1.log](https://github.com/swaywm/sway/files/4232504/sway.1.log)
[config.zip](https://github.com/swaywm/sway/files/4232519/config.zip)

coredumpctl gdb sway:
https://gist.github.com/dsreyes1014/ad1e2d7513e01f58e0fe398645b2b7c4

 bt full:
https://gist.github.com/dsreyes1014/7e9e0a1c1fbac03318e9965b8ae3750e

RedSoxFan:
Fixes #5029 

If the keyboard that triggers the reload binding is using the default
keymap, default repeat delay, and default repeat rate, the associated
keyboard group is never being destroyed on reload. This was causing the
keyboard group's keyboard not to get disarmed and result in a
use-after-free in handle_keyboard_repeat.

If the keyboard was not using the defaults for all three settings, then
it's associated keyboard group would get destroyed during the reset -
which did disarm the keyboard group's keyboard. In this case, the
use-after-free would not occur.

This adds a block to input_manager_reset_all_inputs that disarms the
keyboard for all keyboard groups in all seats. Since the inputs are all
being reset anyway, which will disarm all individual keyboards, it is
not necessary to be selective on which ones get disarmed.

### DIFF
--- a/sway/input/input-manager.c
+++ b/sway/input/input-manager.c
@@ -4,15 +4,17 @@
 #include <string.h>
 #include <math.h>
 #include <wlr/backend/libinput.h>
+#include <wlr/types/wlr_cursor.h>
+#include <wlr/types/wlr_keyboard_group.h>
 #include <wlr/types/wlr_input_inhibitor.h>
 #include <wlr/types/wlr_virtual_keyboard_v1.h>
 #include <wlr/types/wlr_virtual_pointer_v1.h>
-#include <wlr/types/wlr_cursor.h>
 #include "sway/config.h"
+#include "sway/input/cursor.h"
 #include "sway/input/input-manager.h"
+#include "sway/input/keyboard.h"
 #include "sway/input/libinput.h"
 #include "sway/input/seat.h"
-#include "sway/input/cursor.h"
 #include "sway/ipc-server.h"
 #include "sway/server.h"
 #include "stringop.h"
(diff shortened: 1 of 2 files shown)


sway/input/input-manager.c
```diff
@@ -539,6 +541,18 @@ void input_manager_reset_all_inputs() {
...
+
+	// If there is at least one keyboard using the default keymap and repeat
+	// delay and rate, then it is possible that there is a keyboard group that
+	// needs to be disarmed. Since all inputs are being reset anyway, just
+	// disarm all of them.
+	struct sway_seat *seat;
+	wl_list_for_each(seat, &server.input->seats, link) {
+		struct sway_keyboard_group *group;
+		wl_list_for_each(group, &seat->keyboard_groups, link) {
+			sway_keyboard_disarm_key_repeat(group->seat_device->keyboard);
```
Xyene, line 553: This fixes the issue, thanks. One question is, should this also be calling `seat_configure_keyboard`? There are exit/re-enter events that it's not clear that this patch is sending (or, does it not matter for keyboard group keyboards?)

https://github.com/swaywm/sway/blob/726d187d3ce83431a2565cc1b423af99da5f7b6b/sway/input/seat.c#L700

LaserEyess:
I can confirm this fixes #5167, previously the crash on reloading was 100% reproducible.

sway/input/input-manager.c
RedSoxFan: I don't think it's strictly needed now, but may be for #4975 - which is related to keyboards being added/removed to a keyboard group. 

I changed it to call `seat_reset_device`, which is the same method normal input devices use, instead of just `sway_keyboard_disarm_key_repeat`. This required adding a condition in `seat_configure_keyboard` to not call `sway_keyboard_configure` for keyboard group keyboards.

LaserEyess:
I rebuilt sway with the new patch, and I'm still getting the same crash as in #5167. Repeatedly calling `reload` using a keybind triggers it. Sometimes you have to do it for 5 seconds, sometimes 30 seconds.

RedSoxFan:
> I rebuilt sway with the new patch, and I'm still getting the same crash as in #5167. Repeatedly calling `reload` using a keybind triggers it. Sometimes you have to do it for 5 seconds, sometimes 30 seconds.

Oops. Should be fixed now.

LaserEyess:
Ok, after repeated attempts with a clean config, I can't get it to crash again. Spent about 3 minutes trying.

PR #5323, opened 2020-05-10
criteria: Allow re-execution of criteria

michaelweiser:
Currently, criteria are executed only once. This limits their usefulness
for reacting to properties of windows that change over time, for example
existance and state of idle or keyboard shortcuts inhibitors (for which
criteria are to be implemented in a separate change).

This change keeps the principal behaviour of not re-executing criteria
while their matching state has not changed. But as soon as a criterion
no longer matches, it will be removed from the list of already executed
criteria. This allows its action to be re-executed once the criterion
once again matches the view.

Since function criteria_for_view() returns a list of all currently
matching criteria, this can be trivially implemented and is even
slightly more efficient than before: Instead of checking each matching
criterion against the list of already executed criteria and potentially
executing and adding it, we just make the list returned by
criteria_for_view() the new list of executed criteria. Then we walk
through it and execute every criterion not already in the old list of
exectued criteria. Adding the criterion to the list becomes redundant.
The old list is freed at the end.

With this approach we implicitly forget about any criteria we may have
executed in the past which no longer match the view right now.

As an additional simplification, function view_has_executed_criteria()
is replaced by list_find() since it does the same thing.

Since execution behaviour was not documented explicitly before, we add a
note to sway(5) on how it is behaving now.

Signed-off-by: Michael Weiser <michael.weiser@gmx.de>

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -841,6 +841,14 @@ Kill all windows with the title "Emacs":
 [class="Emacs"] kill
 ```
 
+Criteria are applied on state changes of a view. These currently are mapping
+(i.e. opening of a window), adding or removing marks and setting title, app_id,
+class, role or window type.
+
+Criteria will be executed once when first applied or when the view changes its
+attributes to match. They will be re-executed only after the view did not match
+a least once.
+
 You may like to use swaymsg -t get_tree for finding the values of these
 properties in practice for your applications.
 
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -414,29 +414,22 @@ static void view_handle_surface_new_subsurface(struct wl_listener *listener,
 	view_subsurface_create(view, subsurface);
 }
 
-static bool view_has_executed_criteria(struct sway_view *view,
-		struct criteria *criteria) {
-	for (int i = 0; i < view->executed_criteria->length; ++i) {
-		struct criteria *item = view->executed_criteria->items[i];
-		if (item == criteria) {
-			return true;
-		}
-	}
-	return false;
-}
-
 void view_execute_criteria(struct sway_view *view) {
-	list_t *criterias = criteria_for_view(view, CT_COMMAND);
-	for (int i = 0; i < criterias->length; i++) {
-		struct criteria *criteria = criterias->items[i];
+	// replacing list of executed criteria removes those that do no longer
+	// match the view, opening them up for re-execution once they match
+	// again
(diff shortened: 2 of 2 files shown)


michaelweiser:
Can be seen in action with keyboard shortcuts inhibitors and idle inhibitors criteria selectors from #5124.

emersion:
Is this consistent with i3's behavior?

michaelweiser:
Looking at the i3 code (while not being familiar with it) I can not find an explicit list of already executed criteria guarding against re-execution as currently implemented in sway. In fact, https://github.com/i3/i3/blob/dd13398e046cd613778d5cf1f911c8a5163de732/src/commands.c#L176 reads surprisingly familiar to what I'm proposing. So it would seem that if criteria are executed only once in i3 it is an implicit result of criteria only being applied upon window creation rather than being explicitly tracked.

Similar to sway, i3's documentation does nowhere state that criteria are executed only once.

That being said: This feature only really makes sense with criteria that *can* change their matching state as in #5124 and #5325 (or the `mark` command). If these have no chance of getting in because of adherence to i3 behaviour we can save us some effort here as well.

Also, there would be a user-visible change in criteria behaviour for sway users for criteria matching values that can change over the lifecycle of a view. These certainly include `con_mark`/`mark` and `title`/`set_title`. I'm not sure whether `class`, `app_id`, `window_role` and `window_type` can change. Other criteria might benefit from additional calls to `view_execute_criteria()` for consistent behaviour, e.g. when a view changes state from tiling to floating and back so that a respective `for_window` criterion can match and execute an additional action for certain windows. I admit, I had not thought of that.

emersion:
>So it would seem that if criteria are executed only once in i3 it is an implicit result of criteria only being applied upon window creation rather than being explicitly tracked.

Indeed. It seems i3 doesn't execute `for_window` rules on properties changes (e.g. title changes). I'd prefer not to diverge from i3's behavior because that could break some users' config.

i3 devs say that a script can monitor window property changes.

If we do want to support `for_window` rules, we'd need to add a special flag for it, e.g. `for_window --on-update …`. I'm not 100% convinced this is a good idea because we generally don't want to add more window management features than i3 has.

michaelweiser:
Okay, back to the drawing board then. :) I gather, adding idle inhibitor and keyboard shortcuts inhibitor states to the tree so a script can monitor them would be the way to go then.

michaelweiser:
Underway for idle inhibitors in #5307.

PR #5334, opened 2020-05-12
input/pointer: only warp cursor when the confine region has changed

Xyene:
Refs #5268.

Sway counterpart to https://github.com/swaywm/wlroots/pull/2191.


sway/input/cursor.c
```diff
@@ -1222,6 +1231,9 @@ void handle_pointer_constraint(struct wl_listener *listener, void *data) {
 	sway_constraint->destroy.notify = handle_constraint_destroy;
 	wl_signal_add(&constraint->events.destroy, &sway_constraint->destroy);
 
+	sway_constraint->set_region.notify = handle_pointer_constraint_set_region;
+	wl_signal_add(&constraint->events.set_region, &sway_constraint->set_region);
```
emersion, line 1235: We need to `wl_list_remove` this listener in the constraint destructor handler.

Xyene: Missed that, thanks :+1:

PR #5339, opened 2020-05-14
security.d: remove sway-security(7) mention

### LINKED ISSUE #5335: Default config files mention non existent sway-security man pages
`security.d/00-defaults.in` mentions the `sway-security(7)` man pages, which seem to have been deleted in 2018, if I understood the git log correctly. It probably makes sense to remove that mention, at least for now, as the file is non existent.

ericonr:
Fix #5335


emersion:
Actually I think we can delete `security.d` altogether.

ericonr:
Is the commit message well explained?

ericonr:
Huh, it seems to be called in meson build, and `sway/config.c`, in line 467 tries to load it as well. How should the removal work here?

emersion:
Can you try removing both `configure_file` calls?

ericonr:
I made the change in meson.build. What are the plans for the current security code?

emersion:
Ideally we should remove all security-related stuff. We have changed our plans for security as we want to lock down Wayland protocols instead of the Sway IPC.

ericonr:
I can try to make another PR cleaning that up. Would you be interested, then?

emersion:
>I can try to make another PR cleaning that up. Would you be interested, then?

Definitely :)

PR #5348, opened 2020-05-16
Fix typos

tbm:


protocols/wlr-output-power-management-unstable-v1.xml
```diff
@@ -50,7 +50,7 @@
 
     <request name="get_output_power">
       <description summary="get a power management for an output">
-        Create a output power management mode control that can be used to
+        Create an output power management mode control that can be used to
```
emersion, line 53: Please submit this change to wlr-protocols.

tbm: Ok, I removed that part of the patch.

Issue #5350, opened 2020-05-16
swaybar crashes with a segfault


houseofsuns:
On my amd64 Gentoo unstable machine I see the following crash of swaybar since today. This is a bit confusing as it worked yesterday and I did not find any relevant changes in the logs.

Other information: I'm running sway-1.4 on kernel 5.6.12 on Intel graphics.

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7f2a2ea in ?? () from /lib64/libc.so.6
(gdb) bt full
#0  0x00007ffff7f2a2ea in  () at /lib64/libc.so.6
#1  0x000055555555fb0b in ipc_get_workspaces (bar=0x555555568340 <swaybar>) at ../sway-1.4/swaybar/ipc.c:363
        ws_output = <optimized out>
        i = 0
        output = 0x5555555829b0
        len = <optimized out>
        res = 0x5555555fdde0 "[ { \"id\": 4, \"name\": \"1\", \"rect\": { \"x\": 0, \"y\": 0, \"width\": 1920, \"height\": 1200 }, \"focus\": [ 7, 61 ], \"border\": \"none\", \"current_border_width\": 0, \"layout\": \"splith\", \"orientation\": \"horizontal\", \""...
        results = 0x5555555841b0
        length = 6
        ws_json = <optimized out>
        num = 0x0
        name = 0x5555555fe7e0
        visible = 0x0
        focused = 0x5555555ff670
        out = 0x0
        urgent = 0x0
#2  0x000055555555a178 in bar_setup
    (bar=0x555555568340 <swaybar>, socket_path=0x555555579960 "/run/user/1001/sway-ipc.1001.1872.sock")
    at ../sway-1.4/swaybar/main.c:86
        registry = <optimized out>
        seat = 0x555555568068 <xdg_shell_types+72>
        socket_path = 0x555555579960 "/run/user/1001/sway-ipc.1001.1872.sock"
        debug = <optimized out>
        usage = 0x555555564148 "Usage: swaybar [options...]\n\n  -h, --help", ' ' <repeats 13 times>, "Show help message and quit.\n  -v, --version          Show the version number and quit.\n  -s, --socket <socket>  Connect to sway via socket.\n  -b, "...
        c = <optimized out>
        long_options = 
[...]

Issue #5349, opened 2020-05-16
swaymsg is incompatible with json-c-0.14


JusterK:
On sway version 1.4 swaymsg -t get_outputs shows no outputs, when it's compiled with json-c-0.14. The version json-c-0.13.1 works fine though.
```
% swaymsg -t get_outputs
Output DP-1 '(null) (null) (null)' (inactive)

Output HDMI-A-1 '(null) (null) (null)' (inactive)
```
Also see this [https://github.com/Alexays/Waybar/issues/709](url) bug.

sfar9999:
I just hit this. It causes swaybar to crash on startup inside `ipc_get_workspaces`.

The problem is with the generated JSON from sway. The value of `res` in
```
        char *res = ipc_single_command(bar->ipc_socketfd,
                        IPC_GET_WORKSPACES, NULL, &len);
```
looks like this:
[
  {
    "id": 8,
    "name": "j",
    "rect": {
      "x": 32,
      "y": 32,
      "width": 3776,
      "height": 2096
    },
    "focus": [
      7
    ],
    "border": "none",
    "current_border_width": 0,
    "layout": "splith",
    "orientation": "horizontal",
    "percent": null,
    "window_rect": {
      "x": 0,
      "y": 0,
      "width": 0,
      "height": 0
    },
    "focused": false
  },
  {
    "id": 14,
    "name": "f",
    "rect": {
      "x": 32,
      "y": 32,
      "width": 3776,
      "height": 2096
    },
    "focus": [
      15
    ],
    "border": "none",
    "current_border_width": 0,
    "layout": "splith",
    "orientation": "horizontal",
    "percent": null,
    "window_rect": {
      "x": 0,
      "y": 0,
      "width": 0,
      "height": 0
    },
    "focused": false
  }
  ...
]
("f" and "j" are workspace names)

The JSON object properties aren't anything like what swaybar is expecting. Looks more like a top level container than a workspace perhaps?

sfar9999:
https://github.com/json-c/json-c/issues/599
"cannot add more than 11 objects. Is this a known issue?"

Yeah that'll be it :)

JusterK:
@sfar9999 Indeed, just compiled sway with json-c master and everything works again.

Sway still needs an update though. A direct access has been deprecated:
`* Added a json_tokener_get_parse_end() function to replace direct access of tok->char_offset.`

which is used in i3bar.c
```
% find -name '*.c' -type f -exec grep -n char_offset {} \+
[...]

PR #5367, opened 2020-05-20
swaybar: add NULL check when listing workspaces

milkey-mouse:
For some reason my version of sway doesn't show workspace names:

    $ swaymsg -t get_outputs
    Output HDMI-A-1 '(null) (null) (null)' (inactive)
    Output HDMI-A-2 '(null) (null) (null)' (inactive)

Which is weird, but it's no reason to crash swaybar. The field is
totally missing from the JSON, so it ends up doing `strcmp(NULL, name)`
which is undefined behavior.

### DIFF
--- a/swaybar/ipc.c
+++ b/swaybar/ipc.c
@@ -360,7 +360,7 @@ bool ipc_get_workspaces(struct swaybar *bar) {
 
 		wl_list_for_each(output, &bar->outputs, link) {
 			const char *ws_output = json_object_get_string(out);
-			if (strcmp(ws_output, output->name) == 0) {
+			if (ws_output != NULL && strcmp(ws_output, output->name) == 0) {
 				struct swaybar_workspace *ws =
 					calloc(1, sizeof(struct swaybar_workspace));
 				ws->num = json_object_get_int(num);


milkey-mouse:
That's the weird thing, none of my outputs are disabled. I'm writing this in a window on `HDMI-A-1` right now. I think it's a graphics driver issue, but everything is working fine. I know `get_outputs` has worked correctly before so something else must have changed.

I do have workspaces turned off though (or at least, I only have one workspace enabled, and workspaces disabled on the swaybar config). Maybe that has something to do with it?

I didn't consider this worth opening an issue for because it seems harmless and I have no idea what the problem is (and it probably isn't sway's fault anyway).

emersion:
Do you have debug logs?

Make sure to grab the latest wlroots commit, it contains some maybe-related fixes.

milkey-mouse:
Here are my debug logs: https://gist.github.com/milkey-mouse/dac49c33fb66a55d558a3ef278a7c3f7

The problem shows up the same in sway/wlroots 1.4 and master, although master also takes ~10 seconds to exit (which sway 1.4 doesn't).

Xyene:
This is caused by a bug upstream in json-c. https://github.com/swaywm/sway/issues/5349 and https://github.com/swaywm/sway/issues/5350 have more info; it's been fixed upstream but not yet repackaged everywhere.

Issue #5294, opened 2020-05-02
input/seatop_default: properly notify pointer leave causing SEGV in wlroots.so


adee:
kernel: sway[45781]: segfault at 0 ip 00007f7dcb2cabd2 sp 00007ffdd967fa10 error 4 in libwlroots.so.5[7f7dcb2a4000+4f000]`

To reproduce:
- open GHex
- set it floating
- open the rightmost dropdown menu (Help)
- move the cursor to the left to open the menus there
- when it reaches the leftmost menu sway crashes

The commit is: 091f580b7cb0df92f0f6e0be34b52f8abfdbe97a.
This does not generate a coredump. Any ideas how can I make it create one?

CC: @nickdiego

Xyene:
Here's a trace from ASan, from the latest master:

```
Thread 1 "sway" received signal SIGSEGV, Segmentation fault.
AddressSanitizer:DEADLYSIGNAL
=================================================================
==15827==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7ffff6e6cf80 bp 0x7fffffffd340 sp 0x7fffffffd310 T0)
==15827==The signal is caused by a READ memory access.
==15827==Hint: address points to the zero page.
    #0 0x7ffff6e6cf7f in xdg_pointer_grab_enter ../types/xdg_shell/wlr_xdg_popup.c:21
    #1 0x7ffff6e4eb8d in wlr_seat_pointer_notify_enter ../types/seat/wlr_seat_pointer.c:335
    #2 0x5555555fd40a in handle_motion ../sway/input/seatop_default.c:463
    #3 0x5555555f9bd4 in seatop_motion ../sway/input/seat.c:1455
    #4 0x5555555dfe80 in pointer_motion ../sway/input/cursor.c:324
    #5 0x5555555e0090 in handle_pointer_motion_relative ../sway/input/cursor.c:332
    #6 0x7ffff6ecee92 in wlr_signal_emit_safe ../util/signal.c:29
    #7 0x7ffff6e8156e in handle_pointer_motion ../types/wlr_cursor.c:355
    #8 0x7ffff6ecee92 in wlr_signal_emit_safe ../util/signal.c:29
    #9 0x7ffff6e2bf40 in relative_pointer_handle_relative_motion ../backend/wayland/seat.c:469
    #10 0x7ffff61bbccc  (/lib/x86_64-linux-gnu/libffi.so.7+0x6ccc)
[...]

PR #5368, opened 2020-05-21
Use new wlroots API for clearing keyboard/pointer focus during grabs

tchebb:
We are not allowed to do what we did in #5222 and pass a `NULL` surface wlr_seat_pointer_notify_enter(), and it's causing crashes when an xdg-shell popup is active (see #5294 and swaywm/wlroots#2161).

Instead, solve #5220 using the new wlroots API introduced in swaywm/wlroots#2217.


Emantor:
I tested this in conjunction with the swaywm/wlroots#2217 and can confirm that this fixes the annoying crashes I observe when changing windows while a right-click menu is open.

Issue #5361, opened 2020-05-19
seat_cmd_cursor is missing wl_pointer frame events


siyia2:
```
### Control Mouse with Keyboard

bindsym $mod+KP_2 seat "*" cursor move 0 10
bindsym $mod+KP_8 seat "*" cursor move 0 -10
bindsym $mod+KP_9 seat "*" cursor move 10 -10
bindsym $mod+KP_7 seat "*" cursor move -10 -10
bindsym $mod+KP_4 seat "*" cursor move -10 0
bindsym $mod+KP_1 seat "*" cursor move -10 10
bindsym $mod+KP_3 seat "*" cursor move 10 10
bindsym $mod+KP_6 seat "*" cursor move 10 0
bindsym $mod+KP_5 seat "*" cursor press button1; seat "*" cursor release button1
```
For the life of me i cannot simulate a mouse click with my keyboard, when i press the cursor kb hotkey it seems like the mouse click is working, but it doesnt do anything.

Kommynct:
swaymsg seat "*" cursor press button1 && swaymsg seat "*" cursor release button1

works fine for me. you wouldn't want ; here because you don't want the actions to occur at the same time, && does it when the last action completes.

siyia2:
Can you switch tabs in Firefox with KB emulated mouse clicks?

Xyene:
The click actions get generated fine, the issue you're facing is that they get sent while `$mod` is still pressed, and `$mod+button1` is the shortcut for moving windows around. You briefly start moving the window under the cursor (so the cursor flashes to the hand grab icon) before releasing it.

Kommynct:
Ohh, that explains why I had no such issue, as I have my floating modifier set otherwise.

Xyene:
Closing as a configuration issue, since Sway's behavior here is correct (even if perhaps surprising, in this case) and the events are being generated.

siyia2:
omg thx so much for this clarification!!!!

siyia2:
Actually i changed to:
`
KP_2 seat "*" cursor move 0 10
KP_8 seat "*" cursor move 0 -10
KP_9 seat "*" cursor move 10 -10
KP_7 seat "*" cursor move -10 -10
KP_4 seat "*" cursor move -10 0
KP_1 seat "*" cursor move -10 10
KP_3 seat "*" cursor move 10 10
KP_6 seat "*" cursor move 10 0
KP_5 seat "*" cursor press button1; seat "*" cursor release button1
`
[...]

PR #5370, opened 2020-05-21
input/pointer: send pointer enter event on confine warp

Xyene:
The spec has this to say about sending events on confine creation:

 > Whenever the confinement is activated, it is guaranteed that the
surface the pointer is confined to will already have received pointer
focus and that the pointer will be within the region passed to the
request creating this object.

...and on region update:

> If warped, a wl_pointer.motion event will be emitted, but no
wp_relative_pointer.relative_motion event.

Prior to this patch, sway did neither, and updated the hardware cursor
position without notifying the underlying surface until the next motion
event. This led to inconsistent results, especially in applications that
draw their own software cursor.

([Related SDL patch.](https://bugzilla.libsdl.org/show_bug.cgi?id=5151))

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -791,6 +791,8 @@ static void check_constraint_region(struct sway_cursor *cursor) {
 				wlr_cursor_warp_closest(cursor->cursor, NULL,
 					sx + con->content_x - view->geometry.x,
 					sy + con->content_y - view->geometry.y);
+
+				cursor_rebase(cursor);
 			}
 		}
 	}
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -637,6 +637,7 @@ static void handle_rebase(struct sway_seat *seat, uint32_t time_msec) {
 	if (surface) {
 		if (seat_is_input_allowed(seat, surface)) {
 			wlr_seat_pointer_notify_enter(seat->wlr_seat, surface, sx, sy);
+			wlr_seat_pointer_notify_motion(seat->wlr_seat, time_msec, sx, sy);
 		}
 	} else {
 		cursor_update_image(cursor, e->previous_node);


sway/input/cursor.c
```diff
@@ -791,6 +791,11 @@ static void check_constraint_region(struct sway_cursor *cursor) {
 				wlr_cursor_warp_closest(cursor->cursor, NULL,
 					sx + con->content_x - view->geometry.x,
 					sy + con->content_y - view->geometry.y);
+
+				// The specification says we should send pointer enter/motion events as
+				// necessary, but no relative motion events.
+				seatop_pointer_motion(cursor->seat, get_current_time_msec(), 0, 0);
```
emersion, line 797: We generally rebase the cursor in this case. Would that work here?

Xyene: Almost, but I think there's a very corner case that that wouldn't handle. Cursor rebase only sends a pointer enter event:

https://github.com/swaywm/sway/blob/5d13f647f9384e59012c0f829651911564bb5365/sway/input/seatop_default.c#L656-L659

But, after https://github.com/swaywm/sway/commit/6f0a0bd3852c6d92f30d43b9a39dc743987584ac, we may also need to warp when the region changes on the same surface, in which case `wlr_seat_pointer_notify_enter` will be a no-op (if the surface already has pointer focus), but we probably do want to send a motion event to the client anyway.

emersion: Should rebasing the cursor also send a motion event? I'd say yes, because rebasing may update the cursor position without changing the focus.

Xyene: Actually, that sounds like a good idea, done. And then in the case where it doesn't have focus already, it behaves as it has until now, since wlroots won't send a motion event after a real enter.

Xyene: Turns out this was a very good call, since `swaymsg seat * cursor move dx dy` didn't notify the underlying surface of the motion (https://github.com/swaywm/sway/issues/5361), and this fixes that too.

PR #5391, opened 2020-05-27
Log Xwayland startup failures

Xyene:
Though unlikely, this commit makes server initialization totally
explicit in the logs.


emersion:
Hm, small typo

Xyene:
Oops, should have at least tried compiling locally...

PR #5393, opened 2020-05-28
input/keyboard: wlr_keyboard_group enter and leave

### LINKED ISSUE #4975: Character is printed when reload configuration

RedSoxFan:
Fixes #4975 
Requires swaywm/wlroots#2229

This adds support for wlr_keyboard_group's enter and leave events. The
enter event just updates the keyboard's state. The leave event updates
the keyboard's state and if the surface was notified of a press event
for the key code, then it is notified of a key release.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -124,7 +124,7 @@ static void state_add_key(struct sway_shortcut_state *state,
  * Update the shortcut model state in response to new input
  */
 static bool update_shortcut_state(struct sway_shortcut_state *state,
-		struct wlr_event_keyboard_key *event, uint32_t new_key,
+		uint32_t keycode, enum wlr_key_state keystate, uint32_t new_key,
 		uint32_t raw_modifiers) {
 	bool last_key_was_a_modifier = raw_modifiers != state->last_raw_modifiers;
 	state->last_raw_modifiers = raw_modifiers;
@@ -134,12 +134,12 @@ static bool update_shortcut_state(struct sway_shortcut_state *state,
 		state_erase_key(state, state->last_keycode);
 	}
 
-	if (event->state == WLR_KEY_PRESSED) {
+	if (keystate == WLR_KEY_PRESSED) {
 		// Add current key to set; there may be duplicates
-		state_add_key(state, event->keycode, new_key);
-		state->last_keycode = event->keycode;
+		state_add_key(state, keycode, new_key);
+		state->last_keycode = keycode;
 	} else {
-		return state_erase_key(state, event->keycode);
+		return state_erase_key(state, keycode);
 	}
 
 	return false;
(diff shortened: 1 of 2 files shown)


emersion:
We need to properly update the clients' state too.

sway/input/keyboard.c
```diff
@@ -494,6 +514,52 @@ static void handle_keyboard_group_key(struct wl_listener *listener,
...
+		struct wlr_seat *wlr_seat = keyboard->seat_device->sway_seat->wlr_seat;
+		struct wlr_input_device *wlr_device =
+			keyboard->seat_device->input_device->wlr_device;
+
+		struct timespec now;
+		clock_gettime(CLOCK_MONOTONIC, &now);
+		int64_t time_msec = (int64_t)now.tv_sec * 1000 + now.tv_nsec / 1000000;
+
+		wlr_seat_set_keyboard(wlr_seat, wlr_device);
+		wlr_seat_keyboard_notify_key(
```
emersion, line 558: We need to properly update the clients' state too, without causing any "real" key press/release which would trigger an in-app shortcut for instance.

sway/input/keyboard.c
```diff
@@ -349,51 +394,25 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 	bool shortcuts_inhibited = sway_inhibitor && sway_inhibitor->inhibitor->active;
 
 	// Identify new keycode, raw keysym(s), and translated keysym(s)
```
Xyene, line 396: I *think* this comment should be carried over to `update_keyboard_state`?

sway/input/keyboard.c
```diff
@@ -494,6 +514,52 @@ static void handle_keyboard_group_key(struct wl_listener *listener,
...
+			continue;
+		}
+
+		struct wlr_seat *wlr_seat = keyboard->seat_device->sway_seat->wlr_seat;
+		struct wlr_input_device *wlr_device =
+			keyboard->seat_device->input_device->wlr_device;
+
+		struct timespec now;
+		clock_gettime(CLOCK_MONOTONIC, &now);
+		int64_t time_msec = (int64_t)now.tv_sec * 1000 + now.tv_nsec / 1000000;
```
Xyene, line 555: I noticed we were going to duplicate this logic that `input/cursor.c` also has, so I refactored it in https://github.com/swaywm/sway/pull/5425. If that PR ends up getting merged, it'd remove the need to do this here as well.

sway/input/keyboard.c
```diff
@@ -334,9 +336,52 @@ void sway_keyboard_disarm_key_repeat(struct sway_keyboard *keyboard) {
...
+
+	const xkb_keysym_t *translated_keysyms;
+	uint32_t translated_modifiers;
+	size_t translated_keysyms_len;
+};
+
+static void update_keyboard_state(struct sway_keyboard *keyboard,
+		uint32_t raw_keycode, enum wlr_key_state keystate,
+		struct key_info *keyinfo) {
+	keyinfo->keycode = raw_keycode + 8;
```
Xyene, line 355: I know this is just pre-existing code that was moved around, but what does this 8 represent?

RedSoxFan: xkb keycodes are 8 higher than event codes (https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h). I'm blanking out on the reasoning.

sway/input/keyboard.c
RedSoxFan: I think the only way to do this would be a wl_keyboard leave immediately followed by an enter. I think this could be done once and only if any of the keys had the pressed event sent. I'll switch it over to do that tomorrow.

RedSoxFan: The focused surface is now being refocused when at least one keycode had a pressed event sent. This should update the surface and prevent in app bindings from being triggered.

sway/input/keyboard.c
RedSoxFan: Copied the comment

sway/input/keyboard.c
RedSoxFan: Sounds good. I didn't end up needing this after the changes, but it may come in handy in the future

sway/input/keyboard.c
```diff
@@ -494,6 +513,49 @@ static void handle_keyboard_group_key(struct wl_listener *listener,
...
+
+		pressed_sent |= update_shortcut_state(&keyboard->state_pressed_sent,
+				*keycode, WLR_KEY_RELEASED, keyinfo.keycode, 0);
+	}
+
+	if (!pressed_sent) {
+		return;
+	}
+
+	// Refocus the focused node so the keyboard state gets synced
```
Xyene, line 552: Does this need to consider unmanaged Xwayland surfaces as well? I recall from [here](https://github.com/Xyene/sway/blob/a7bcfc56021c192975c7a5ae5f22d1d0770e37ea/sway/desktop/xwayland.c#L102-L118) that the focus for these are handled differently since they don't have `sway_node`s associated with them.

RedSoxFan: I pushed a new version which handles when the currently focused surface is a layer surface or unmanaged surface.

PR #5394, opened 2020-05-28
ipc: invert output transformation when necessary

### LINKED ISSUE #5356: output transform doesn't agree with configuration
I have a dual monitor setup like this: `|-`

To get the outputs to be oriented correctly I issue:
```
swaymsg output DP-1 position 0,-600 transform 270
swaymsg output DP-2 position 1440,0 transform normal
```

`swaymsg -t get_outputs` will return the following:
```
Output DP-1 (focused)
  Current mode: 2560x1440 @ 59.951000 Hz
  Position: 0,-600
  Scale factor: 1.000000
  Scale filter: nearest
  Subpixel hinting: unknown
  Transform: 90
  Workspace: 1
  Max render time: 1 ms
  Adaptive sync: disabled
  Available modes:
[...]

Output DP-2
  Current mode: 2560x1440 @ 59.951000 Hz
  Position: 1440,0
  Scale factor: 1.000000
  Scale filter: nearest
  Subpixel hinting: unknown
  Transform: normal
  Workspace: m
  Max render time: 1 ms
  Adaptive sync: disabled
  Available modes:
[...]
```

The transform values seem to not match. This leads to problems with `grim` which will display the vertical output rotated the wrong way.

Interestingly when using `kanshi` the applied value matches the one returned by `swaymsg -t get_outputs`. 

---

* Sway Version: `sway version 1.4-aa48b926 (May 14 2020, branch 'HEAD')`
* Debug log: https://paste.sr.ht/blob/4b9e8b69fcf25abb4505c88cf46bd8d0dd4d0ed5

Xyene:
After swaywm/wlroots#2023, #4996 inverted config inputs to
`transform`. For consistency, we should also communicate inverted
transformations via IPC.

Closes #5356.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -68,19 +68,24 @@ static const char *ipc_json_output_transform_description(enum wl_output_transfor
 	case WL_OUTPUT_TRANSFORM_NORMAL:
 		return "normal";
 	case WL_OUTPUT_TRANSFORM_90:
-		return "90";
+		// Sway uses clockwise transforms, while WL_OUTPUT_TRANSFORM_* describes
+		// anti-clockwise transforms.
+		return "270";
 	case WL_OUTPUT_TRANSFORM_180:
 		return "180";
 	case WL_OUTPUT_TRANSFORM_270:
-		return "270";
+		// Transform also inverted here.
+		return "90";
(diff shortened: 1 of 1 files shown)


Xyene:
Done :+1:

PR #5400, opened 2020-05-29
seat_cmd_cursor: emit frame events

### LINKED ISSUE #5361: seat_cmd_cursor is missing wl_pointer frame events

RedSoxFan:
Fixes #5361 

This emits frame events for the seat_cmd_cursor subcommands. The
wl_pointer frame event are required to notify clients when to process
the events. It will now be emitted after cursor movement, button press,
button release, and axis events.

### DIFF
--- a/sway/commands/seat/cursor.c
+++ b/sway/commands/seat/cursor.c
@@ -24,6 +24,7 @@ static struct cmd_results *handle_command(struct sway_cursor *cursor,
 		int delta_y = strtol(argv[2], NULL, 10);
 		wlr_cursor_move(cursor->cursor, NULL, delta_x, delta_y);
 		cursor_rebase(cursor);
+		wlr_seat_pointer_notify_frame(cursor->seat->wlr_seat);
 	} else if (strcasecmp(argv[0], "set") == 0) {
 		if (argc < 3) {
 			return cmd_results_new(CMD_INVALID, expected_syntax);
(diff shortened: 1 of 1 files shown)


sway/commands/seat/cursor.c
```diff
@@ -32,6 +33,7 @@ static struct cmd_results *handle_command(struct sway_cursor *cursor,
 		float x = strtof(argv[1], NULL) / root->width;
 		float y = strtof(argv[2], NULL) / root->height;
 		wlr_cursor_warp_absolute(cursor->cursor, NULL, x, y);
+		wlr_seat_pointer_notify_frame(cursor->seat->wlr_seat);
 		cursor_rebase(cursor);
```
Xyene, line 37: Hmm, perhaps this might be better placed in `cursor_rebase`, or at least come after the call to it since it's the one that's sending enter/motion events here, not `wlr_cursor_warp_absolute`.

Xyene: Sorry, meant in the seatop implementing `cursor_rebase`.

RedSoxFan: I moved it to after `cursor_rebase`. I don't think emitting it inside `cursor_rebase` (or the seatop handler) is a good idea since we call `cursor_rebase` in places where there is no cursor event - such as when a view gets resized or unmapped.

Xyene: That sounds good then. I was concerned about whether we sent a frame to the surface underneath a floating window when the floating window is unmapped while having pointer focus, but it looks like we do anyway.

PR #5402, opened 2020-05-30
input/cursor: only consider in-bounds floaters for fullscreen hit test

Xyene:
This makes transient containers have logic more in-line with the test from `floating_container_at`. This fixes bugs where a floating container would take input way past its borders when its parent was fullscreened.

---

This was particularly noticeable in Xwayland apps. For instance, a fullscreened Chrome or Jetbrains IDE dialog would be resizable (and show the resize cursor image) everywhere outside its bounds. It's also present in Wayland-native apps, e.g. Firefox's <kbd>Ctrl</kbd>+<kbd>O</kbd> file browser can be moved by dragging around anywhere, not just on the dialog itself.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -274,14 +274,10 @@ static struct sway_container *container_at_linear(struct sway_node *parent,
 	list_t *children = node_get_children(parent);
 	for (int i = 0; i < children->length; ++i) {
 		struct sway_container *child = children->items[i];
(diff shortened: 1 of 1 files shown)


Xyene:
> ...all verify this condition already so this is the one missing case.

Good call, done.

RedSoxFan:
Thanks!

Issue #4441, opened 2019-08-10
sctk CSD disappears during resize


YaLTeR:
* Sway Version: `1.1-rc1-99-g724926ea (Aug  7 2019, branch 'makepkg')`
* Debug Log: https://gist.github.com/YaLTeR/399b7e98d9f1d78b4a69b2992812adae
* Configuration File: default

`smithay-client-toolkit` CSD briefly disappears in sway during resize. This can be seen on both sctk examples and `winit` windows, such as `alacritty`. Doesn't happen on Weston. [Video.](https://streamable.com/ngorc)

1. Open `alacritty` or a sctk example.
1. Make it floating.
1. `swaymsg border csd`
1. Try resizing by the border or with a shortcut, see CSD disappear.

kchibisov:
Fixed by #5403.

PR #5403, opened 2020-05-30
Improve resize transactions

kennylevinsen:
This PR contains a few improvements to the resizing logic:

1. Save all view buffers, including subsurfaces, rather than only the main surface. Fixes subsurfaces disappearing during resize.
2. Ensure that "current" geometry is not trashed, and is used when appropriate. Fixes border/titlebar and content disagreeing on dimensions during resize.
3. Update centering logic to account for complete and current dimensions. Fixes positioning during resize, especially with CSDs.
4. Remove bogus dimensions being tracked on the container state that was no longer useful.

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -55,6 +55,13 @@ struct sway_view_impl {
 	void (*destroy)(struct sway_view *view);
 };
 
+struct sway_saved_buffer {
+	struct wlr_client_buffer *buffer;
+	int x, y;
+	int width, height;
+	struct wl_list link; // sway_view::saved_buffers
+};
+
 struct sway_view {
 	enum sway_view_type type;
 	const struct sway_view_impl *impl;
@@ -64,9 +71,6 @@ struct sway_view {
 	struct sway_xdg_decoration *xdg_decoration;
 
 	pid_t pid;
-
-	double saved_x, saved_y;
-	int saved_width, saved_height;
 
 	// The size the view would want to be if it weren't tiled.
 	// Used when changing a view from tiled to floating.
@@ -80,8 +84,7 @@ struct sway_view {
 	bool allow_request_urgent;
 	struct wl_event_source *urgent_timer;
 
-	struct wlr_client_buffer *saved_buffer;
-	int saved_buffer_width, saved_buffer_height;
+	struct wl_list saved_buffers; // sway_saved_buffer::link
 
 	// The geometry for whatever the client is committing, regardless of
 	// transaction state. Updated on every commit.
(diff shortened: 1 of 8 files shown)


kennylevinsen:
cc @emersion, @kchibisov

Note: A damage issue can currently be seen if one resizes e.g. firefox aggressively enough. Unsure if this is new, as firefox currently had nothing to render during resize, but worth a look regardless.

kchibisov:
Seems to fix https://github.com/swaywm/sway/issues/4441 and firefox flickering on resize.

kennylevinsen:
Another note: I am questioning the use of these saved buffers. This PR fixes them, but I fail to see a scenario where they *actually* help over just rendering the latest buffers, and not dealing with saved buffers also fixes the issue.

emersion:
>I am questioning the use of these saved buffers

This is used for atomic transactions. When re-configuring the layout we need to send new configure events to all clients. We don't update anything till all clients have ack'ed the configure event (with a timeout). When a client submits an updated buffer but another client is still pending, we need to render the previous buffer. That's why we save it.

kennylevinsen:
So the point is to not change anything till everything is ready? That would make sense, but not sure how effective this is at the moment. E.g., in local tests with a 2x2 grid, resizing with the mouse so that all views change size leads to buffers being all over the place, with no visible difference with and without saved buffers.

Have not debugged further.

emersion:
> Expected size in transactions are not accounting for borders/decorations, so they often assume the new buffer is bad.

That should only affect xwayland right? Wayland has a serial mechanism.

kennylevinsen:
I have managed to obtain pretty good resize behavior. I took the liberty of expanding the scope of the PR slightly to cover these.

kennylevinsen:
I think this is good enough for now. More improvements (e.g. moving pending states to state structs like is done with current state, maybe clarifying the different sizes) can be done in a later PR.

include/sway/tree/view.h
```diff
@@ -80,8 +83,7 @@ struct sway_view {
 	bool allow_request_urgent;
 	struct wl_event_source *urgent_timer;
 
-	struct wlr_client_buffer *saved_buffer;
-	int saved_buffer_width, saved_buffer_height;
+	list_t *saved_buffers;
```
emersion, line 86: Can we use `wl_list` instead of `list_t` here? It seems more appropriate.

kennylevinsen:
1. Converted `list_t` to `struct wl_list`
2. Fixed a multi-output bug where saved buffer rendering would stop on first failed output intersect instead of proceeding to the next buffer that might intersect
3. Added some cleanup on view_destroy that we were technically missing.

PR #5415, opened 2020-06-03
Save transform during transaction

### LINKED ISSUE #5412: transactions: saved buffers aren't transformed
Saved buffers are rendered with `WL_OUTPUT_TRANSFORM_NORMAL` even if their original wl_surface was transformed.

coderkalyan:
Closes #5412.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -1199,6 +1199,7 @@ static void view_save_buffer_iterator(struct wlr_surface *surface,
 		saved_buffer->height = surface->current.height;
 		saved_buffer->x = sx;
 		saved_buffer->y = sy;
+		saved_buffer->transform = wlr_output_transform_invert(surface->current.transform);
 		wl_list_insert(&view->saved_buffers, &saved_buffer->link);
 	}
 }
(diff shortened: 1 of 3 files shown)


sway/tree/view.c
```diff
@@ -1199,6 +1199,7 @@ static void view_save_buffer_iterator(struct wlr_surface *surface,
 		saved_buffer->height = surface->current.height;
 		saved_buffer->x = sx;
 		saved_buffer->y = sy;
+		saved_buffer->transform = wlr_output_transform_invert(surface->current.transform);
```
emersion, line 1202: Nit: can we do the `wlr_output_transform_invert` call when rendering? This would make the rendering logic more consistent with the regular path:

https://github.com/swaywm/sway/blob/f7e050c58f61a15bf5987e2e3d280a72666f6381/sway/desktop/render.c#L150

coderkalyan: Ok, I'll just save the current.transform then.

coderkalyan:
@emersion Ok, I just moved the inversion to the render routine (just stashed the current.transform instead of its inverted counterpart).

PR #5420, opened 2020-06-04
Patches to address static analysis warnings

MisterDA:
Hi!
This is a collection of small patches fixing some warnings reported by PVS-Studio.
Happy to change things if I got the "bug fix" wrong.

### DIFF
--- a/common/loop.c
+++ b/common/loop.c
@@ -117,9 +117,15 @@ void loop_add_fd(struct loop *loop, int fd, short mask,
 	struct pollfd pfd = {fd, mask, 0};
 
 	if (loop->fd_length == loop->fd_capacity) {
-		loop->fd_capacity += 10;
-		loop->fds = realloc(loop->fds,
-				sizeof(struct pollfd) * loop->fd_capacity);
+		int capacity = loop->fd_capacity + 10;
+		struct pollfd *tmp = realloc(loop->fds,
+				sizeof(struct pollfd) * capacity);
+		if (!tmp) {
+			sway_log(SWAY_ERROR, "Unable to allocate memory for pollfd");
+			return;
+		}
+		loop->fds = tmp;
+		loop->fd_capacity = capacity;
 	}
 
 	loop->fds[loop->fd_length++] = pfd;
--- a/sway/commands/bar/colors.c
+++ b/sway/commands/bar/colors.c
@@ -24,6 +24,9 @@ static char *hex_to_rgba_hex(const char *hex) {
 		return NULL;
 	}
 	char *rgba = malloc(10);
+	if (!rgba) {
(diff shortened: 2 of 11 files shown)


sway/commands/bar/colors.c
```diff
@@ -36,7 +36,7 @@ static struct cmd_results *parse_single_color(char **color,
 	}
 
 	char *rgba = hex_to_rgba_hex(argv[0]);
-	if (!*rgba) {
+	if (!rgba) {
```
Xyene, line 39: In the same train of thought, there's an unchecked `malloc` that's `snprintf`'d into as part of `hex_to_rgba_hex`. Let's include that as part of this patch so we fix things entirely :)

common/loop.c
```diff
@@ -118,8 +118,13 @@ void loop_add_fd(struct loop *loop, int fd, short mask,
 
 	if (loop->fd_length == loop->fd_capacity) {
 		loop->fd_capacity += 10;
-		loop->fds = realloc(loop->fds,
+		struct pollfd *tmp = realloc(loop->fds,
 				sizeof(struct pollfd) * loop->fd_capacity);
+		if(!tmp) {
```
Xyene, line 123: Style: leave a space after the `if` here.

sway/commands/bar/colors.c
```diff
@@ -24,6 +24,8 @@ static char *hex_to_rgba_hex(const char *hex) {
 		return NULL;
 	}
 	char *rgba = malloc(10);
+	if (!rgba)
```
Xyene, line 27: Style: braces around all if statements, even if they're single line.

common/loop.c
```diff
@@ -118,8 +118,13 @@ void loop_add_fd(struct loop *loop, int fd, short mask,
 
 	if (loop->fd_length == loop->fd_capacity) {
 		loop->fd_capacity += 10;
-		loop->fds = realloc(loop->fds,
+		struct pollfd *tmp = realloc(loop->fds,
 				sizeof(struct pollfd) * loop->fd_capacity);
+		if (!tmp) {
```
Xyene, line 123: Sorry, this dropped off my review queue -- last nit here, rather than assigning to `tmp` first, you can just assign to `loop->fds` and `!loop->fds`.

MisterDA: I don’t think so, not at least in the general case. If `realloc` fails, it returns `NULL`. If we write to the `loop->fds` pointer in that case, the memory that was pointed to is lost and there’s a leak.
The patch I’m proposing is idiomatic and lets the possibility of free'ing the memory if realloc fails.

MisterDA: Hey, what do you think of it now?

PR #5425, opened 2020-06-05
common/util: fix get_current_time_msec returning microseconds

Xyene:
This commit makes `get_current_time_msec` correctly return milliseconds
as opposed to microseconds. It also considers the value of `tv_sec`, so
we don't lose occasionally go back in time by one second. Finally, the
function is moved into `util.c` so that it can be reused elsewhere
without having to consider these pitfalls.

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -1,14 +1,20 @@
 #define _POSIX_C_SOURCE 200809L
 #include <ctype.h>
-#include <float.h>
 #include <fcntl.h>
 #include <math.h>
+#include <time.h>
 #include <stdlib.h>
 #include <string.h>
 #include <strings.h>
 #include <wayland-server-protocol.h>
 #include "log.h"
 #include "util.h"
+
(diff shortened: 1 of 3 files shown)


emersion:
>This commit makes get_current_time_msec correctly return milliseconds
as opposed to microseconds. It also considers the value of tv_sec, so
we don't lose occasionally go back in time by one second.

Good catch!

>Finally, the function is moved into util.c so that it can be reused elsewhere
without having to consider these pitfalls.

If we only need to use it in this file, nothing wrong with keeping it static. Do you plan to use it elsewhere?

Xyene:
I was planning on suggesting its usage here: https://github.com/swaywm/sway/pull/5393/files#diff-d7aa39a56641abe2c1362ca922ba299bR555

emersion:
Related: https://github.com/swaywm/sway/pull/5077

PR #5429, opened 2020-06-07
input: tweak resize behavior to not change tab focus on border click

Xyene:
If a resize is triggered on a tabbed or stacked container, change focus to the tab which already had inactive focus, rather than to the tab whose border was clicked -- otherwise, we'd change the active tab when the user probably just wanted to resize.

---

Tested with tabbed and stacked containers, with both `focus_follows_mouse yes` and `no`.


RedSoxFan:
Thanks!

PR #5430, opened 2020-06-07
swaynag: allow specifying more buttons which execute and dismiss

grahamc:
I don't love -z / -Z, but I figure this patch is far from being
accepted for other reasons too.


emersion:
@RedSoxFan, what do you think of the idea of this PR?

grahamc:
I updated the PR to update swaynag.1.scd, since swaynag.5.scd does not seem related.

PR #5431, opened 2020-06-07
input/cursor: don't send wl_pointer.motion event on pointer unlock warp

### LINKED ISSUE #5405: Pointer warp issue in World of Warcraft after commit 6ea4539
The behaviour change with pointer warping in commit 6ea4539 causes issues with World of Warcraft's use of warp. Now, the game is already a little odd with pointer warp and at the moment requires an Xwayland patch to function as it does in X11: https://gitlab.freedesktop.org/xorg/xserver/-/issues/734 . In the game, holding the right mouse button hides and warps the cursor to center and starts rotating the camera. Releasing right mouse button shows the cursor, warps it back to start and disables camera rotation. With the change in commit 6ea4539 the camera is jumping when releasing the right mouse button. My assumption is that because the warp back now causes an event, it is still taken as camera movement.

I'm a little unsure where this issue should be directed to. Under X11 everything works so the issue is in wine -> Xwayland -> sway. The breaking change is in sway, but it seems the change was made to conform to spec. The game itself is non-free and as this has been the behaviour for decades it is unlikely to change. The example program in Xwayland issue 734 shows the behaviour when the motion event print is uncommented. The position change on the first 2 motion events after button release is large.

* Sway Version: sway version 1.4-6ea45395 (May 31 2020, branch 'makepkg')
* Debug Log: https://gist.github.com/Morgus/a9105cae3f005245dfbefedfb12452ea

Xyene:
On warping to a cursor hint, update the pointer position we track as well, so that on the next pointer rebase we don't send an unexpected synthetic motion event to clients.

Fixes #5405.

---

Refs https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/20, awaiting clarification on that end but figured I should bring up a patch for testing/review earlier. I don't have any games/applications that require this patch, so it hasn't been tested by me (hoping to get someone affected by the issue to give it a shot).

If the plan is to tag Sway 1.5 in the near-ish future, I think we should either (temporarily) revert https://github.com/swaywm/sway/commit/6ea45395c70939a6d855736cabfe75ad9cf4a0ae or apply something along the lines of this patch. The slight cosmetic improvement https://github.com/swaywm/sway/commit/6ea45395c70939a6d855736cabfe75ad9cf4a0ae provides (+ making clicks post-`swaymsg cursor move` click the right location) is probably not worth breaking applications that rely on not seeing a motion event post-unlock.

(Tacking onto 1.5 milestone so that it doesn't get lost.)

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -1244,6 +1244,10 @@ static void warp_to_constraint_cursor_hint(struct sway_cursor *cursor) {
 		double ly = sy + con->content_y - view->geometry.y;
 
 		wlr_cursor_warp(cursor->cursor, NULL, lx, ly);
+
+		// Warp the pointer as well, so that on the next pointer rebase we don't
+		// send an unexpected synthetic motion event to clients.
+		wlr_seat_pointer_warp(constraint->seat, sx, sy);
 	}
 }
 


emersion:
Hm, this writes into the internal seat state. I wonder if there's a better way to do this…

Xyene:
Yeah, I'm not a huge fan of poking into wlroots like this either. The fundamental problem is we do seem to need to be able to change that state without notifying clients of motion events (for this and #5459, at least), but knowledge of that state existing breaks the abstraction (as far as I can tell, it's just there as an optimization to prevent sending duplicate motion events).

Perhaps we could add `wlr_seat_pointer_get_last_position` and `wlr_seat_pointer_warp` to wlroots, and use those instead of directly poking wlroots?

sway/input/cursor.c
```diff
@@ -1234,6 +1234,11 @@ static void warp_to_constraint_cursor_hint(struct sway_cursor *cursor) {
 		double ly = sy + con->content_y - view->geometry.y;
 
 		wlr_cursor_warp(cursor->cursor, NULL, lx, ly);
```
emersion, line 1236: I'd expect `wlr_cursor_warp` to call `wlr_seat_pointer_notify_motion` and update the seat pointer state. Is this not the case?

Xyene: No, `wlr_cursor_warp` and friends only update the position of the cursor image. For instance, they're also used by tablet input, where sending pointer events would be incorrect.

emersion: Ah, I see.

emersion:
+1 to `wlr_seat_pointer_warp`

emersion:
Thanks!

PR #5440, opened 2020-06-11
swaybar: ensure correct init order for status_line

ldrumm:
`$WAYLAND_SOCKET` is unset by `wl_display_connect` after it has
    successfully connected to the wayland socket.
    
    However, subprocesses spawned by swaybar (status-command) don't have
    access to waybar's fds as $WAYLAND_SOCKET is O_CLOEXEC. This means any
    status command which itself tries to connect to wayland will fail if
    this environment variable is set.
    
    Reorder display and status-command initialization so that this variable
    is not set and add an assert so we can enforce this invariant in future.

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -413,10 +413,6 @@ bool bar_setup(struct swaybar *bar, const char *socket_path) {
 	if (!ipc_initialize(bar)) {
 		return false;
 	}
-	if (bar->config->status_command) {
-		bar->status = status_line_init(bar->config->status_command);
-		bar->status->bar = bar;
-	}
 
 	bar->display = wl_display_connect(NULL);
 	if (!bar->display) {
@@ -443,6 +439,11 @@ bool bar_setup(struct swaybar *bar, const char *socket_path) {
 		pointer->cursor_surface =
 			wl_compositor_create_surface(bar->compositor);
 		assert(pointer->cursor_surface);
+	}
+
+	if (bar->config->status_command) {
+		bar->status = status_line_init(bar->config->status_command);
+		bar->status->bar = bar;
 	}
 
 #if HAVE_TRAY
--- a/swaybar/status_line.c
+++ b/swaybar/status_line.c
@@ -1,4 +1,5 @@
 #define _POSIX_C_SOURCE 200809L
+#include <assert.h>
 #include <fcntl.h>
 #include <sys/ioctl.h>
 #include <json.h>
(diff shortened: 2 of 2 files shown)


swaybar/main.c
```diff
@@ -91,6 +91,9 @@ int main(int argc, char **argv) {
 		free(socket_path);
 		return 1;
 	}
+	// We might have been spawned as a wayland client of sway, but
+	// any children we spawn won't have access to our fds
+	unsetenv("WAYLAND_SOCKET");
```
emersion, line 96: `wl_display_connect` already does this.

ldrumm: Thanks. I read through [`wl_display_connect`](https://gitlab.freedesktop.org/wayland/wayland/-/blob/a61ae8eca858f866636eb2721f42ca0fa7fd06f8/src/wayland-client.c#L1135-1161) and can see that `WAYLAND_SOCKET` should not be set, but it is set in the child spawned by status line

e.g.:
```
bar {
    status_command while printenv WAYLAND_SOCKET; do sleep 1; done
}
```
screenshot of this bar is below.

I also made the following changes
```diff
diff --git a/swaybar/status_line.c b/swaybar/status_line.c
index fb9271f8..86ddc943 100644
--- a/swaybar/status_line.c
+++ b/swaybar/status_line.c
@@ -153,6 +153,9 @@ struct status_line *status_line_init(char *cmd) {
                exit(1);
        }
 
+    if (getenv("WAYLAND_SOCKET")) {
+        sway_log(SWAY_ERROR, "WAYLAND_SOCKET should not be set");
+    }
        status->pid = fork();
        if (status->pid == 0) {
                dup2(pipe_read_fd[1], STDOUT_FILENO);
```
...and can see the error in the log:
```
2020-06-11 20:35:16 - [swaybar/status_line.c:151] WAYLAND_SOCKET should not be set
2020-06-11 20:35:16 - [swaybar/bar.c:173] Sending stop signal to status command
2020-06-11 20:35:16 - [swaybar/tray/tray.c:37] Initializing tray
2020-06-11 20:35:16 - [swaybar/tray/watcher.c:196] Registered org.freedesktop.StatusNotifierWatcher
2020-06-11 20:35:16 - [swaybar/tray/watcher.c:196] Registered org.kde.StatusNotifierWatcher
```

I will have to debug this further. Perhaps `unsetenv` is failing for some reason, or this is a glibc bug. I'll update this patch with my findings
image

ldrumm: Found it!
`status_line_init` is [called before the display is initalized](https://github.com/swaywm/sway/blob/master/swaybar/bar.c#L417-L430)

I'll update the patch

ldrumm:
descript edit

PR #5443, opened 2020-06-11
contrib: Add simple battery monitoring status-command

ldrumm:
This status displays a clock and monitors the laptop battery, nagging
via swaynag when the battery capacity drops below the configured
threshold. It nags once each subsequent percentage point drop.
Also displays the local time which is updated roughly each second.

### DIFF
--- a/contrib/laptop_status.c
+++ b/contrib/laptop_status.c
@@ -1,0 +1,78 @@
+#include <signal.h>
+#include <stdbool.h>
+#include <unistd.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#if !defined(__linux__)
+#error This program uses Linux-specific sysfs features
+#endif
+
+/* Simple sway status_command for laptops. Prints the date and the current
+ * battery status. Warns using swaynag when the battery drops to an unreasonable
+ * level. For every percentage point demise below that level, renags.
+ *
+ * Should compile without arguments with any reasonable Linux compiler.
+ * It should be easy to modify the battery status handling code for other
+ * systems
+ *
+ * example installation and configuration:
+ *      cc laptop_status.c -o ~/.local/bin/laptop-status
+ *      sed -i 's/status_command.+/status_command ~/.local/bin/laptop-status' \
+ *          ~/.config/sway/config'
+ */
+
+static const char *cap_file = "/sys/class/power_supply/BAT0/capacity";
+static const char *stat_file = "/sys/class/power_supply/BAT0/status";
+static const int bat_warn_below = 10;
+
+int main(void)
+{
+    setvbuf(stdout, NULL, _IOLBF, 0);
+    int status = open(stat_file, O_RDONLY);
+    int cap = open(cap_file, O_RDONLY);
+    if (cap < 0 || status < 0) {
+        fprintf(stderr, "no battery found\n");
+        exit(EXIT_FAILURE);
+    }
+
+    char timestamp[128] = {0}, bat_status[32] = {0}, bat_capacity[32] = {0};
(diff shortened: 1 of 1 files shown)


MisterDA:
You could make the battery path configurable with a command-line argument.

ldrumm:
On Fri Jun 12, 2020 at 10:43 AM BST, Antonin Décimo wrote:
> You could make the battery path configurable with a command-line
> argument.

Is the battery ever at a different path? As far as I know
`/sys/class/power_supply/BAT<N>` is a stable user-space interface. The only case
I can see that being useful is if there are multiple batteries, in which case
this example program is of limited utility as the core logic only supports a
single battery. Given the user already has to compile this themself I think
keeping it simple is of better value.

That said, if you really think supporting multiple batteries is important we can
add such support, but I don't think it should be a priority.

illiliti:
How about connecting to netlink and listen battery uevents instead of using read() and sleep() ?

ldrumm:
On Fri Sep 11, 2020 at 11:04 AM BST, illiliti wrote:
> How about connecting to netlink and listen battery uevents instead of
> using read() and sleep() ?

Because I have no idea such a thing existed!

Thanks for the tip. I'll take a look. My guess is that because we have to sleep
a maximum of a second for the time to change it may only buy us one fewer system
call (poll with a timeout of 1 second) at the expense of a lot of extra
complexity, but I'll see what I can do.

emersion:
I'm not sure this is in scope for `contrib/`.

Geo25rey:
This might be more revealent if this was a shell script. You might want to check out the i3/i3status project.

ldrumm:
On Sat Oct 31, 2020 at 3:14 PM GMT, Geoffrey Casper wrote:
> This might be more revealent if this was a shell script. You might want
> to check out the i3/i3status project.

@Geo25rey I wrote the present implementation because there's unnecessary
overhead in a shell script. Forking at least once per second is not free, and
has a non-negligible effect on the battery.

This shell version (carefully written to minimize forks)

```shell
#!/bin/sh

nag_pid=
while : ; do
    t="$(date)"
    read bat_percent < /sys/class/power_supply/BAT0/capacity
    read bat_status < /sys/class/power_supply/BAT0/status
    if [ "$bat_percent" -lt 10 ] ; then
        if [ "$nag_pid" -ne "" ] ; then
            kill "$nag_pid"
        fi
        swaynag "battery is low" &
        nag_pid=$?
    fi
    if [ "$status" = Discharging ] ; then
        status=" "
    else
        status="+"
    fi
    printf "%s[% 3d%%]%s\n" "$t" $bat_percent $status
    sleep 1
done
```

...results in 2 `clone/execve` pairs per second and 152 syscalls per loop
iteration. The C version results in 2 `pread`s, a `writev`, and a `nanosleep`
per loop iteration. My shell is `dash`, where `[` and `printf` are shell
builtins.

```
You're welcome to adopt the shell version above, but resource usage
matters so I think personally think this is a better approach than a shell
script.

@illiliti I did some investigation into uevents, and found that the kernel only
sends uevents for power supply connection/disconnection events, but not for
energy level changes. Thanks for pointing them out though. I lear

@emersion what criteria do you have for acceptance into `contrib/`? I'm not
especially concerned if you say "no", but you're the maintainer so please let
me know what `contrib/` is for. I know ddevault encourages maintainers to be
fairly liberal with a contrib directory [1], so perhaps you can clarify your
expectations with a `README` that future contributors know better when something
is out of scope.

[1] https://drewdevault.com/2020/06/06/Add-a-contrib-directory.html
>
> --
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/swaywm/sway/pull/5443#issuecomment-719947118

emersion:
Well, the thing is, this script has almost nothing to do with Sway itself. It doesn't use swaymsg unlike most `contrib/` scripts.

PR #5451, opened 2020-06-13
Clip surfaces to container

kennylevinsen:
This draft PR attempts to clip surfaces to their configured size if applicable.

Works by clamping box dimensions to the view size with fmin. Maybe we could just use the view width/height instead and flat out ignore the surface dimensions.

## TODO
- [X] ~~Squeeze toplevels (unintentional feature)~~ fixed
- [X] Clip toplevels
- [X] Verify transformations (tested 90 degrees at least)
- [X] Verify layer shell/xwayland/popups

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -32,6 +32,7 @@
 struct render_data {
 	pixman_region32_t *damage;
 	float alpha;
+	struct sway_container *container;
 };
 
 /**
@@ -149,14 +150,22 @@ static void render_surface_iterator(struct sway_output *output, struct sway_view
 	struct wlr_fbox src_box;
 	wlr_surface_get_buffer_source_box(surface, &src_box);
 
-	struct wlr_box dst_box = *_box;
-	scale_box(&dst_box, wlr_output->scale);
+	struct wlr_box proj_box = *_box;
+	scale_box(&proj_box, wlr_output->scale);
 
 	float matrix[9];
 	enum wl_output_transform transform =
 		wlr_output_transform_invert(surface->current.transform);
-	wlr_matrix_project_box(matrix, &dst_box, transform, rotation,
+	wlr_matrix_project_box(matrix, &proj_box, transform, rotation,
 		wlr_output->transform_matrix);
+
+	struct wlr_box dst_box = *_box;
+	struct sway_container *container = data->container;
+	if (container != NULL) {
+		dst_box.width = fmin(dst_box.width, container->current.content_width - surface->sx);
+		dst_box.height = fmin(dst_box.height, container->current.content_height - surface->sy);
+	}
+	scale_box(&dst_box, wlr_output->scale);
 
 	render_texture(wlr_output, output_damage, texture,
 		&src_box, &dst_box, matrix, alpha);
@@ -256,6 +265,9 @@ static void render_view_toplevels(struct sway_view *view,
 		.damage = damage,
 		.alpha = alpha,
 	};
+	if (!container_is_current_floating(view->container)) {
+		data.container = view->container;
+	}
 	// Render all toplevels without descending into popups
 	double ox = view->container->surface_x -
 		output->lx - view->geometry.x;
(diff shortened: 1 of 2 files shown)


tmccombs:
What problem is this trying to fix? This will make https://github.com/swaywm/sway/issues/4924 even worse than it currently is?

kennylevinsen:
> What problem is this trying to fix?

That surfaces are not constrained to their container, leading to graphical glitches. E.g., a tiling setup where a tile is made smaller than the application intends to draw, leading to content overlap.

This is a pretty common complaint.

>  This will make #4924 even worse than it currently is?

It will likely behave as @emersion described in the GTK issue, clipping everything except in a few scenarios around tiling windows.

dimitris-triantafyllidis:
@kennylevinsen Hi! How is this going? I gave it a try by clipping `damage` in `render_view`and then restoring it back after the view is rendered, and it seems to work nicely. No glitches so far, scaling and transforms seem to work fine. It is 15-20 lines of pixman boilerplate. If such an approach is acceptable I can test it a bit better and make a PR so that you can check it out. Just let me know.

kennylevinsen:
> @kennylevinsen Hi! How is this going?

Haven't spent time looking at it. Priorities and all. :/

> I gave it a try by clipping `damage` in `render_view`and then restoring it back after the view is rendered, and it seems to work nicely. No glitches so far, scaling and transforms seem to work fine. It is 15-20 lines of pixman boilerplate. If such an approach is acceptable I can test it a bit better and make a PR so that you can check it out. Just let me know.

You're always free to make a draft PR, the best code wins. :)

dimitris-triantafyllidis:
@kennylevinsen I opened #5957, in case you want to check it out or leave a comment.

kennylevinsen:
Seems to work now, including some quick tests with transforms and scales.

dimitris-triantafyllidis:
@kennylevinsen I tried this and I believe a couple of things need fixing.

Shadows are not accounted for and parts of the right and bottom sides of the view get clipped. I think shadows in floating views should be drawn anyway. If using csd and sway does not draw borders, the view could be "lost" in the background without its shadows. They would also not be clipped totally in windows with rounded corners and it wouldn't look nice.

Secondly, parts of popups that are clickable get clipped and the clipping is also not aligned with the container's bounds, for example in gnome-terminal.

tmccombs:
> parts of popups that are clickable get clipped and the clipping is also not aligned with the container's bounds, for example in gnome-terminal

Related to https://github.com/swaywm/sway/issues/4924?

kennylevinsen:
> Shadows are not accounted for and parts of the right and bottom sides of the view get clipped.

Ah, yes. The quick fix is the same as yours, just not doing this for floating windows. Not sure about the "parts missing" - I'll have to take a look at that one myself.

> Secondly, parts of popups that are clickable get clipped and the clipping is also not aligned with the container's bounds, for example in gnome-terminal.

Ah, adding the clipping in popup rendering was probably a brainfart. Let's not do that.

kennylevinsen:
@dimitris-triantafyllidis The issues you found should be fixed now. Thank you for testing. :)

dimitris-triantafyllidis:
@kennylevinsen I did a very quick check and yes they are. But, while resizing floating views such as gedit and nautilus, the right / bottom parts that were being clipped are not drawn continuously, resulting in some flicker.

kennylevinsen:
Forgot to add the floating exception to saved views as well.

sway/desktop/render.c
```diff
@@ -158,6 +159,14 @@ static void render_surface_iterator(struct sway_output *output, struct sway_view
 	wlr_matrix_project_box(matrix, &dst_box, transform, rotation,
 		wlr_output->transform_matrix);
 
+	struct sway_container *container = data->container;
+	if (container != NULL) {
+		dst_box.width = fmin(dst_box.width,
+				wlr_output->scale * (container->current.content_width - surface->sx));
```
emersion, line 165: Is there a reason why we can't do this before the `scale_box` call? Is it because the projection needs the full non-clipped box?

kennylevinsen: Clipping the box before the projection leads to a transformation matrix that squeezes the surface to fit rather than clipping it on overflow...

Which is a shame, as the code was a lot prettier when I could clip before scale_box. :(

dimitris-triantafyllidis: Maybe the scaling in the `fmin` calls needs some adjustment by the way, in order to be scaled equivalently to how `dst_box_width` is scaled?

kennylevinsen: Can you elaborate on what your mean by this?

dimitris-triantafyllidis: > Ah, you mean the rounding logic in scale_box?
Yep!

kennylevinsen:
ping @emersion

emersion (approved):
LGTM, thanks!

PR #5453, opened 2020-06-14
input/tablet: focus appropriate node on tip down

Xyene:
This commit makes tablet input more usable when `focus_follows_mouse` is
set to `no`.

Previously, tapping down on surfaces that bound tablet input would not
switch focus, whereas tapping on surfaces that didn't (and hence went
through pointer emulation) did.

This misbehavior is masked when `focus_follows_mouse` is set to `yes`,
which is probably why this hasn't come up before.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -220,7 +220,14 @@ static void handle_tablet_tool_tip(struct sway_seat *seat,
 	struct sway_container *cont = node && node->type == N_CONTAINER ?
 		node->sway_container : NULL;
 
-	if (cont) {
+	if (wlr_surface_is_layer_surface(surface)) {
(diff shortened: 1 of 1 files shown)


sway/input/seatop_default.c
```diff
@@ -220,7 +220,14 @@ static void handle_tablet_tool_tip(struct sway_seat *seat,
 	struct sway_container *cont = node && node->type == N_CONTAINER ?
 		node->sway_container : NULL;
 
-	if (cont) {
+	if (wlr_surface_is_layer_surface(surface)) {
```
Xyene, line 223: Note for reviewer: the mouse button down path first tries to focus an empty workspace. However, we don't need to do so here because if `surface` is null, we'll be hitting the pointer emulation path anyway.

PR #5457, opened 2020-06-14
input/cursor: send idle events based off device type, not input type

Xyene:
Previously, a tablet or touch device could report activity as a pointer
device if it went through pointer emulation. This commit refactors idle
sources to be consistently reported based on the type of the device that
generated an input event, and now how that input event is being
processed.


sway/input/cursor.c
```diff
@@ -358,6 +377,7 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 static void handle_pointer_button(struct wl_listener *listener, void *data) {
 	struct sway_cursor *cursor = wl_container_of(listener, cursor, button);
 	struct wlr_event_pointer_button *event = data;
+	cursor_handle_activity(cursor, event->device);
```
Xyene, line 380: Note for reviewer: these have been refactored out of `pointer_motion`.

sway/input/cursor.c
```diff
@@ -369,7 +389,6 @@ static void handle_pointer_button(struct wl_listener *listener, void *data) {
 		}
 	}
 
-	cursor_handle_activity(cursor, IDLE_SOURCE_POINTER);
```
Xyene, line 372: Note for reviewer: I moved all calls to `cursor_handle_activity` to the start of functions, so figuring out if one ends up reporting an idle source change doesn't require scanning the whole function.

sway/input/seat.c
```diff
@@ -1380,9 +1380,9 @@ void seat_apply_config(struct sway_seat *seat,
 
 	wl_list_for_each(seat_device, &seat->devices, link) {
 		seat_configure_device(seat, seat_device->input_device);
+		cursor_handle_activity(seat->cursor,
+			seat_device->input_device->wlr_device);
 	}
-
-	cursor_handle_activity(seat->cursor, IDLE_SOURCE_POINTER);
```
Xyene, line 1385: Note for reviewer: I'm not sure what this call was supposed to accomplish here, so I've just made it trigger for all devices on the seat rather than just the pointer.

sway/input/cursor.c
```diff
@@ -383,20 +402,21 @@ void dispatch_cursor_axis(struct sway_cursor *cursor,
...
 	struct wlr_event_pointer_axis *event = data;
-	cursor_handle_activity(cursor, IDLE_SOURCE_POINTER);
+	cursor_handle_activity(cursor, event->device);
 	dispatch_cursor_axis(cursor, event);
 	transaction_commit_dirty();
 }
 
 static void handle_pointer_frame(struct wl_listener *listener, void *data) {
 	struct sway_cursor *cursor = wl_container_of(listener, cursor, frame);
-	cursor_handle_activity(cursor, IDLE_SOURCE_POINTER);
```
Xyene, line 393: We don't have a reference to the device here, so I just removed the call. We shouldn't really be seeing frame events without axis/button events, which would already update the idle status.

sway/input/cursor.c
```diff
@@ -322,6 +339,7 @@ static void handle_pointer_motion_relative(
 		struct wl_listener *listener, void *data) {
 	struct sway_cursor *cursor = wl_container_of(listener, cursor, motion);
 	struct wlr_event_pointer_motion *e = data;
+	cursor_handle_activity(cursor, e->device);
```
Xyene, line 342: Note for reviewer: these have been refactored out of `pointer_motion`.

sway/input/cursor.c
```diff
@@ -253,11 +253,32 @@ int cursor_get_timeout(struct sway_cursor *cursor) {
 	return timeout;
 }
 
+static enum sway_input_idle_source idle_source_from_device(
+		struct wlr_input_device *device) {
+	switch (device->type) {
+		case WLR_INPUT_DEVICE_KEYBOARD:
```
emersion, line 259: Style: `case` statements should be on the same indentation level as `switch`

sway/input/cursor.c
```diff
@@ -253,11 +253,32 @@ int cursor_get_timeout(struct sway_cursor *cursor) {
...
+			return IDLE_SOURCE_TOUCH;
+		case WLR_INPUT_DEVICE_TABLET_TOOL:
+			return IDLE_SOURCE_TABLET_TOOL;
+		case WLR_INPUT_DEVICE_TABLET_PAD:
+			return IDLE_SOURCE_TABLET_PAD;
+		case WLR_INPUT_DEVICE_SWITCH:
+			return IDLE_SOURCE_SWITCH;
+	}
+
+	assert(false);
```
emersion, line 273: `abort()` is preferred because some release builds make assertions a no-op and that makes build fail (missing return value).

sway/input/seat.c
emersion: This was introduced in 4d88c957905e7.

Probably the intent was to reset/restart the idle timers in case their configuration changed?

cc @RedSoxFan

Xyene:
Fixed!

sway/input/seat.c
RedSoxFan: emersion is correct on the reasoning. It predated the source argument, which is why it was only called once for the seat.

PR #5458, opened 2020-06-14
input/pointer: don't trigger pointer bindings for emulated input

Xyene:
Prior to this commit, a tablet device could trigger mouse button down
bindings if the pen was pressed on a surface that didn't bind tablet
handlers -- but it wouldn't if the surface did bind tablet handlers.

We should expose consistent behavior to users so that they don't have to
care about emulated vs. non-emulated input, so stop triggering bindings
for any non-pointer devices.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -253,10 +253,47 @@ static void handle_tablet_tool_tip(struct sway_seat *seat,
  * Functions used by handle_button  /
  *--------------------------------*/
 
+static bool trigger_pointer_button_binding(struct sway_seat *seat,
+		struct wlr_input_device *device, uint32_t button,
+		enum wlr_button_state state, uint32_t modifiers,
+		bool on_titlebar, bool on_border, bool on_contents, bool on_workspace) {
+	// We can reach this for non-pointer devices if we're currently emulating
+	// pointer input for one. Emulated input should not trigger bindings.
+	if (device->type != WLR_INPUT_DEVICE_POINTER) {
+		return false;
+	}
+
+	struct seatop_default_event *e = seat->seatop_data;
+
+	char *device_identifier = device ? input_device_get_identifier(device)
+		: strdup("*");
+	struct sway_binding *binding = NULL;
+	if (state == WLR_BUTTON_PRESSED) {
+		state_add_button(e, button);
+		binding = get_active_mouse_binding(e,
+			config->current_mode->mouse_bindings, modifiers, false,
+			on_titlebar, on_border, on_contents, on_workspace,
+			device_identifier);
+	} else {
+		binding = get_active_mouse_binding(e,
+			config->current_mode->mouse_bindings, modifiers, true,
+			on_titlebar, on_border, on_contents, on_workspace,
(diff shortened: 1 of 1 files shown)


sway/input/seatop_default.c
```diff
@@ -342,8 +362,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 	}
 
 	// Handle tiling resize via mod
-	bool mod_pressed = keyboard &&
```
Xyene, line 345: Minor style changed batched here; `modifiers` is guaranteed to be `0` if `keyboard` is null.

sway/input/seatop_default.c
```diff
@@ -253,10 +253,47 @@ static void handle_tablet_tool_tip(struct sway_seat *seat,
...
+	// We can reach this for non-pointer devices if we're currently emulating
+	// pointer input for one. Emulated input should not trigger bindings.
+	if (device->type != WLR_INPUT_DEVICE_POINTER) {
+		return false;
+	}
+
+	struct seatop_default_event *e = seat->seatop_data;
+
+	char *device_identifier = device ? input_device_get_identifier(device)
+									 : strdup("*");
```
emersion, line 269: Style: never align, instead use one or two tabs to indent continuation lines.

sway/input/seatop_default.c
```diff
@@ -253,10 +253,47 @@ static void handle_tablet_tool_tip(struct sway_seat *seat,
...
+		state_add_button(e, button);
+		binding = get_active_mouse_binding(e,
+			config->current_mode->mouse_bindings, modifiers, false,
+			on_titlebar, on_border, on_contents, on_workspace,
+			device_identifier);
+	} else {
+		binding = get_active_mouse_binding(e,
+			config->current_mode->mouse_bindings, modifiers, true,
+			on_titlebar, on_border, on_contents, on_workspace,
+			 device_identifier);
```
emersion, line 281: Style: stray space here

sway/input/seatop_default.c
```diff
@@ -278,29 +315,12 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
...
 
-	// Handle mouse bindings
 	struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(seat->wlr_seat);
 	uint32_t modifiers = keyboard ? wlr_keyboard_get_modifiers(keyboard) : 0;
 
-	char *device_identifier = device ? input_device_get_identifier(device)
-		: strdup("*");
-	struct sway_binding *binding = NULL;
-	if (state == WLR_BUTTON_PRESSED) {
-		state_add_button(e, button);
```
emersion, line 289: Do we need to add/remove buttons to the state for emulated input?

emersion: Sounds like that's not the case -- the state is only used for bindings.

sway/input/seatop_default.c
Xyene: Thanks for noticing, looks like I had "Reformat on paste" enabled in my editor by default :/

PR #5459, opened 2020-06-15
input/cursor: restore pointer position after emulating touch input

### LINKED ISSUE #1896: [sway master latest touch] mouse pointer-event doesn't keep its coordinates after touch-events
Hi, I've tried the recent master with the basic touch support - one thing I've experienced on my touch laptop:

tapping on my touch screen and afterwards clicking the mouse pointer results in a click at the last position of the touch pointer, though the mouse pointer is visible on a different position.

I'm not able to look at the code, but wanted to mention this, it's a small thing and can easily be overseen

Thanks for your great work on sway !

Xyene:
Fixes #1896.

---

Disclaimer: I have no touch device with which to test this patch. It *seems* correct by reading through it, but some real testing from someone who can would be helpful :)

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -304,35 +304,51 @@ void cursor_unhide(struct sway_cursor *cursor) {
 static void pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
 		struct wlr_input_device *device, double dx, double dy,
 		double dx_unaccel, double dy_unaccel) {
+	double sx, sy;
+	struct wlr_seat *wlr_seat = cursor->seat->wlr_seat;
 	wlr_relative_pointer_manager_v1_send_relative_motion(
 		server.relative_pointer_manager,
-		cursor->seat->wlr_seat, (uint64_t)time_msec * 1000,
+		wlr_seat, (uint64_t)time_msec * 1000,
 		dx, dy, dx_unaccel, dy_unaccel);
 
-	// Only apply pointer constraints to real pointer input.
-	if (cursor->active_constraint && device->type == WLR_INPUT_DEVICE_POINTER) {
-		struct wlr_surface *surface = NULL;
-		double sx, sy;
-		node_at_coords(cursor->seat,
-			cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
+	if (device->type == WLR_INPUT_DEVICE_POINTER) {
+		// Only apply pointer constraints to real pointer input.
+		if (cursor->active_constraint) {
+			struct wlr_surface *surface = NULL;
+			node_at_coords(cursor->seat,
+				cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
 
-		if (cursor->active_constraint->surface != surface) {
-			return;
+			if (cursor->active_constraint->surface != surface) {
+				return;
+			}
+
+			double sx_confined, sy_confined;
+			if (!wlr_region_confine(&cursor->confine, sx, sy, sx + dx, sy + dy,
+					&sx_confined, &sy_confined)) {
+				return;
+			}
+
+			dx = sx_confined - sx;
+			dy = sy_confined - sy;
 		}
 
-		double sx_confined, sy_confined;
-		if (!wlr_region_confine(&cursor->confine, sx, sy, sx + dx, sy + dy,
-				&sx_confined, &sy_confined)) {
-			return;
-		}
-
-		dx = sx_confined - sx;
-		dy = sy_confined - sy;
+		// The expectation is that touch input will not update the cursor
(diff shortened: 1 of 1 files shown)


emersion:
Ref https://github.com/swaywm/sway/pull/5431#issuecomment-644859617

sway/input/cursor.c
```diff
@@ -287,35 +287,53 @@ static void pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
...
-				&sx_confined, &sy_confined)) {
-			return;
+			dx = sx_confined - sx;
+			dy = sy_confined - sy;
 		}
 
-		dx = sx_confined - sx;
-		dy = sy_confined - sy;
+		// The expectation is that touch input will not update the cursor
+		// image position at all, even if it's being emulated as pointer input.
```
emersion, line 319: This is a little misleading tbh. Why shouldn't we update the cursor position when emulating touch input?

Xyene: When we're using touch input, the cursor image is hidden. Updating the cursor position when on a touch surface is rather useless due to multi-touch, so we don't -- but that makes it confusing when the user moves over a non-touch surface, where we currently _do_ move the cursor image despite it not being visible. I think we should be consistent in what we do, so that users have a good idea of where their pointer is even when it's hidden.

For reference, other OSes I've tried:

* In Chrome OS, on touch the real cursor is hidden, and only reappears (in its original position) the next time a touchpad event is received.

* In Windows 10, touch down hides the cursor, but does not move it. Touch _move_ moves the real cursor (while still hidden), *unless* there are currently multiple touch points active, in which case the cursor does not move. The cursor reappears on a touchpad event.

The proposed behaviour is most similar to that from Chrome OS.

emersion: >I think we should be consistent in what we do, so that users have a good idea of where their pointer is even when it's hidden

Agreed.

What I'd expect is:

- Hide the pointer when using touch on a client that supports touch
- Show the pointer when using touch on a client that doesn't support it (emulated input)
- Show the pointer once a mouse/touchpad on the same seat starts being used again

Wdyt?

emersion: Also, I thought we were sending a pointer leave event when hiding the cursor while sending touch input?

Xyene: > Show the pointer when using touch on a client that doesn't support it 

Hmm, this sounds like it offloads understanding about touch/non-touch surfaces to the user (e.g. "why does my pointer show up in application X but not in Y?").

I'll admit that I'm a bit biased to "just doing what other OSes do" since they probably had a bunch of UX people think more about this than I have, but I'm not sure I see the value in doing this. The UX between apps that handle touch events and those that don't is homogenous on Windows/Chrome OS -- the cursor is hidden unconditionally.

> Also, I thought we were sending a pointer leave event when hiding the cursor while sending touch input?

Yes, this matches my understanding too.

`handle_touch_down` [calls `cursor_hide`](https://github.com/swaywm/sway/blob/master/sway/input/cursor.c#L419) which [calls `wlr_seat_pointer_notify_clear_focus`]( https://github.com/swaywm/sway/blob/master/sway/input/cursor.c#L230).

This will cause the `pointer_state.focused_client` to become `NULL`, so any calls to `wlr_seat_pointer_send_motion` will [short circuit](https://github.com/swaywm/wlroots/blob/122318a5d5605e272d888e9a8b00a17e26b571c4/types/seat/wlr_seat_pointer.c#L235).

However, during pointer emulation we end up calling `handle_pointer_motion`, which [re-enters the surface](https://github.com/swaywm/sway/blob/master/sway/input/seatop_default.c#L558) (possibly immediately after having exited it -- now that I'm writing this out this sounds like another "bug" that should be addressed separately, later: we should only leave the surface if the surface supports touch).

emersion: >this sounds like it offloads understanding about touch/non-touch surfaces to the user (e.g. "why does my pointer show up in application X but not in Y?")

Correct, but it might also be less confusing: e.g. "why does multi-touch doesn't work on this app?" or "why can't I scroll by moving my finger on my touchscreen on this app?".

And also it feels like this behaviour would be less of a pain to implement. :P

>The UX between apps that handle touch events and those that don't is homogenous on Windows/Chrome OS -- the cursor is hidden unconditionally.

Ah, didn't know that.

Ultimately, I don't feel that strongly about it, I'm fine either way. What I'm more interested in is a way to properly fix this bug. The hack in this PR doesn't feel right -- do you have an idea to make this less hacky?

>now that I'm writing this out this sounds like another "bug" that should be addressed separately, later: we should only leave the surface if the surface supports touch

Yeah, created https://github.com/swaywm/sway/issues/5510 to track it.

PR #5466, opened 2020-06-17
grimshot: fix POSIX compliance

blaggacao:
`source` is not POSIX compliant, '`.`' is.
https://stackoverflow.com/a/11588629

Take together with #5465 

The reason this didn't bubble up before might be that `/bin/sh` resolves to `/bin/bash` for some people?


blaggacao:
image
Just had to try it out while supporting the argument :wink:

PR #5467, opened 2020-06-18
input/keyboard: use $XDG_CONFIG_HOME/xkb as first lookup path for the XKB context

1ace:
Simply porting @whot's Mutter patch to Sway:  
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/936

@whot: could you have a quick look to confirm this is correct? thanks!

### DIFF
--- a/include/meta-keymap-utils.h
+++ b/include/meta-keymap-utils.h
@@ -1,0 +1,28 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */
+/*
+ * Utilities for use with libxkbcommon
+ *
+ * Copyright 2020 Red Hat, Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef META_KEYMAP_UTILS_H
+#define META_KEYMAP_UTILS_H
+
+#include <xkbcommon/xkbcommon.h>
+
+struct xkb_context * meta_create_xkb_context(void);
+
+#endif /* META_KEYMAP_UTILS_H */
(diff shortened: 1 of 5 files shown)


include/meta-keymap-utils.h
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef META_KEYMAP_UTILS_H
+#define META_KEYMAP_UTILS_H
+
+#include <xkbcommon/xkbcommon.h>
+
+struct xkb_context * meta_create_xkb_context(void);
```
whot, line 26: the `meta` prefix comes from mutter's predecessor metacity and is the standard prefix for internal functions. probably not the best prefix to use here :)

sway/meta-keymap-utils.c
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+#include "meta-keymap-utils.h"
+
+#include <glib.h>
+#include <linux/limits.h>
+
+struct xkb_context *
+meta_create_xkb_context (void)
+{
```
whot, line 28: gnome projects have a quirky indentation style, you'll likely need to fix this for sway to match the rest.

sway/meta-keymap-utils.c
```diff
@@ -0,0 +1,53 @@
...
+	char xdg[PATH_MAX] = {0};
+	const char *env;
+
+	/*
+	 * We can only append search paths in libxkbcommon, so we start with an
+	 * emtpy set, then add the XDG dir, then add the default search paths.
+	 */
+	ctx = xkb_context_new(XKB_CONTEXT_NO_DEFAULT_INCLUDES);
+
+	env = g_getenv("XDG_CONFIG_HOME");
```
whot, line 39: quick `git grep` says that sway isn't using glib so you'll have to replace all of these with the normal calls.

emersion:
Would it be a good idea to write a patch for libxkbcommon to do this by default instead?

whot:
> Would it be a good idea to write a patch for libxkbcommon to do this by default instead?

It already does but libxkbcommon uses `secure_getenv()` and that returns NULL if you have capabilities (or setuid root). As said above, I don't know whether sway needs this patch. If `secure_getenv()` works fine then you don't.

emersion:
Right, we have suid root if we're running without systemd. Though we drop our permissions pretty early, not sure dropping permissions makes `secure_getenv` work properly.

emersion:
Sway never has capabilities fwiw.

whot:
> Though we drop our permissions pretty early, not sure dropping permissions makes secure_getenv work properly.

a quick read of the glibc sources suggests that `__libc_init_secure` is set on startup and if set, `secure_getenv()` returns NULL. So dropping permissions won't help. Easy to verify though: just add a `secure_getenv("HOME")` before the libxkbcommon context is created and check if that's NULL.

emersion:
Note, we're phasing out the suid approach, it's being superseded by seatd-launch.

kennylevinsen:
We do actually support CAP_SYS_NICE now...

PR #5468, opened 2020-06-19
input: disable events for map_to_output devices when output not present

### LINKED ISSUE #3449: map_to_output when output is disabled
I map my touch devices (touch screen and wacom pen) to my laptop screen. It works as expected when both my laptop screen and external screens are enabled.  However, when I disable my laptop screen (i.e., only the external screens are enabled): the touch screen gets disabled as expected, but the wacom pen again works on the whole output surface (i.e., my external screens). ~~That may be another instance of #3443.~~

Xyene:
Fixes #3449.

---

Tested with a pointer device, since I do not have a libinput recent enough to contain https://gitlab.freedesktop.org/libinput/libinput/-/commit/0d06bfc4e27bf4a56a7f34b3f20c4a61570e765c.

### DIFF
--- a/include/sway/input/input-manager.h
+++ b/include/sway/input/input-manager.h
@@ -44,9 +44,11 @@ void input_manager_configure_xcursor(void);
 
 void input_manager_apply_input_config(struct input_config *input_config);
 
+void input_manager_configure_all_inputs(void);
+
 void input_manager_reset_input(struct sway_input_device *input_device);
 
-void input_manager_reset_all_inputs();
+void input_manager_reset_all_inputs(void);
 
 void input_manager_apply_seat_config(struct seat_config *seat_config);
 
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -481,14 +481,7 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 	// Reconfigure all devices, since input config may have been applied before
 	// this output came online, and some config items (like map_to_output) are
 	// dependent on an output being present.
-	struct sway_input_device *input_device = NULL;
-	wl_list_for_each(input_device, &server.input->devices, link) {
-		struct sway_seat *seat = NULL;
-		wl_list_for_each(seat, &server.input->seats, link) {
-			seat_configure_device(seat, input_device);
-		}
-	}
-
+	input_manager_configure_all_inputs();
 	return true;
 }
 
--- a/sway/input/input-manager.c
+++ b/sway/input/input-manager.c
@@ -521,6 +521,22 @@ static void retranslate_keysyms(struct input_config *input_config) {
 	}
 }
 
+static void input_manager_configure_input(
+		struct sway_input_device *input_device) {
+	sway_input_configure_libinput_device(input_device);
+	struct sway_seat *seat = NULL;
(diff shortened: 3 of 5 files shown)


include/sway/input/input-manager.h
```diff
@@ -44,9 +44,13 @@ void input_manager_configure_xcursor(void);
 
 void input_manager_apply_input_config(struct input_config *input_config);
 
+void input_manager_configure_input(struct sway_input_device *input_device);
```
RedSoxFan, line 47: It doesn't look like this is being used outside of `input-manager.c`. We generally try to keep all functions static until they are used elsewhere.

sway/config/output.c
```diff
@@ -478,17 +481,7 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 		output->max_render_time = oc->max_render_time;
 	}
 
-	// Reconfigure all devices, since input config may have been applied before
```
RedSoxFan, line 481: I think it makes sense to keep this comment here

sway/input/input-manager.c
```diff
@@ -542,6 +542,21 @@ void input_manager_apply_input_config(struct input_config *input_config) {
 	retranslate_keysyms(input_config);
 }
 
+void input_manager_configure_input(struct sway_input_device *input_device) {
```
RedSoxFan, line 545: Can you replace the matching lines in `input_manager_apply_input_config` right above this with a call to this new function?

include/sway/input/input-manager.h
Xyene: For symmetry, I based this off `input_manager_reset_input` below which is also exported despite not being used elsewhere. I've dropped this one from the header, though.

sway/input/input-manager.c
Xyene: Done :+1:

sway/config/output.c
Xyene: I moved (and reworded, since now we care about both enabling/disabling where the comment had been written for enabling only) the comment to https://github.com/swaywm/sway/pull/5468/files#diff-4f65f4327e987fef8ec3796cdb07644eR429-R430; I think that branch contains the more tricky `map_to_output` logic the comment refers to (though this call to `input_manager_configure_all_inputs` is still necessary).

RedSoxFan: I think it makes sense to have both comments. The branch above handles the case of disabling the input since the output is now disabled. This call here handles (re-)enabling the input due to the output becoming available (again).

Xyene: Alright, restored.

include/sway/input/input-manager.h
RedSoxFan: I think it may have been used at some point and just never converted to static. I could be wrong and it was just an oversight though. Either way, thanks. We do try to follow that, but it's not a strict requirement.

sway/config/output.c
RedSoxFan: Thanks.

RedSoxFan:
Oh actually, one more thing. This is missing the case where an enabled output gets disconnected.

Xyene:
Good catch, addressed that case as well.

emersion:
Thanks!

Issue #4139, opened 2019-05-10
Add a configuration option to set a tablet as a relative input


Eragonfr:
I have a Wacom intuos M Bluetooth, and i like using it as a relative input.
I haven't view any sway configuration option to set that.
It's possible to add that into sway ? Or into wlroots ?

ascent12:
libinput does give us relative coordinates for tablet events, but sway isn't hooked up to use it.

https://github.com/swaywm/sway/blob/master/sway/input/cursor.c#L440-L469
(`struct wlr_event_tablet_tool_axis` has that `dx` and `dy` as a member).

I'll admit that the input code would be the part I'm the least familiar with, but I don't think it would be that hard for someone to add support for that.

Eragonfr:
Progress with the issue #2288  ?

Eragonfr:
I'm not a good dev, and I don't know how i can add my modification in a good way. I want to add an option similar to `option "Mode Relative"`(from X libinput configuration) into main sway configuration file; so users can choose between relative and absolute input.
Do you have any idea how I can create that ?

PR #5472, opened 2020-06-19
Add a configuration option to set a tablet as a relative input

Xyene:
Closes https://github.com/swaywm/sway/issues/4139.

This commit adds a new option to set whether movement of a tablet tool should be treated as absolute or relative; the default is absolute.

Configuration is per-tool, with pen/eraser/brush/pencil/airbrush tools being independently configurable. Mouse and lens tools ignore this setting and are always treated as relative.

Included is a change to switch lens tool input to relative. I don't have such a device to test this, but [the libinput docs](https://wayland.freedesktop.org/libinput/doc/1.11.3/tablet-support.html#tablet-relative-motion) recommend having it as relative by default.

Tested with the pen tool, since that's all I have :)

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -6,11 +6,13 @@
 #include <time.h>
 #include <wlr/interfaces/wlr_switch.h>
 #include <wlr/types/wlr_box.h>
+#include <wlr/types/wlr_tablet_tool.h>
 #include <xkbcommon/xkbcommon.h>
 #include "../include/config.h"
 #include "list.h"
 #include "swaynag.h"
 #include "tree/container.h"
+#include "sway/input/tablet.h"
 #include "sway/tree/root.h"
 #include "wlr-layer-shell-unstable-v1-protocol.h"
 
@@ -116,6 +118,11 @@ enum input_config_mapped_to {
 	MAPPED_TO_REGION,
 };
 
+struct input_config_tool {
+	enum wlr_tablet_tool_type type;
+	enum sway_tablet_tool_mode mode;
+};
+
 /**
  * options for input devices
  */
@@ -159,6 +166,8 @@ struct input_config {
 	enum input_config_mapped_to mapped_to;
 	char *mapped_to_output;
 	struct wlr_box *mapped_to_region;
+
+	list_t *tools;
 
 	bool capturable;
 	struct wlr_box region;
--- a/sway/commands/input/tool_mode.c
+++ b/sway/commands/input/tool_mode.c
@@ -1,0 +1,73 @@
+#include <strings.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+
+static void set_tool_mode(struct input_config *ic,
+		enum wlr_tablet_tool_type type, enum sway_tablet_tool_mode mode) {
+	for (int i = 0; i < ic->tools->length; i++) {
+		struct input_config_tool *tool = ic->tools->items[i];
+		if (tool->type == type) {
+			tool->mode = mode;
+			return;
+		}
+	}
+
+	struct input_config_tool *tool = calloc(1, sizeof(*tool));
+	if (tool) {
+		tool->type = type;
+		tool->mode = mode;
+		list_add(ic->tools, tool);
+	}
+}
+
+struct cmd_results *input_cmd_tool_mode(int argc, char **argv) {
+	struct cmd_results *error;
+	if ((error = checkarg(argc, "tool_mode", EXPECTED_AT_LEAST, 2))) {
+		return error;
+	}
+
+	struct input_config *ic = config->handler_context.input_config;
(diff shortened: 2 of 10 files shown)


sway/sway-input.5.scd
```diff
@@ -94,6 +94,17 @@ The following commands may only be used in the configuration file.
...
+## TABLET CONFIGURATION
+
+*input* <identifier> tool_mode <tool> <absolute|relative>
+	Sets whether movement of a tablet tool should be treated as absolute or
+	relative; the default is absolute.
+
+	Valid values for _\<tool\>_ are currently "pen", "eraser", "brush",
+	"pencil", "airbrush", and the wildcard _\*_, which matches all tools.
+
+	Mouse and lens tools ignore this setting and are always treated as relative.
```
Xyene, line 106: There's nothing really blocking allowing this to be changed for mouse and lens as well, except I don't have a device with any so I don't know if they also report absolute position... if they do, then I don't see a reason for not allowing them here.

sway/commands/input/tool_mode.c
```diff
@@ -0,0 +1,73 @@
...
+		enum wlr_tablet_tool_type type, enum sway_tablet_tool_mode mode) {
+	for (int i = 0; i < ic->tools->length; i++) {
+		struct input_config_tool *tool = ic->tools->items[i];
+		if (tool->type == type) {
+			tool->mode = mode;
+			return;
+		}
+	}
+
+	struct input_config_tool *tool = calloc(1, sizeof(*tool));
```
RedSoxFan, line 15: style: `sizeof(struct input_config_tool)`

sway/config/input.c
```diff
@@ -153,6 +154,22 @@ void merge_input_config(struct input_config *dst, struct input_config *src) {
...
+		struct input_config_tool *src_tool = src->tools->items[i];
+		for (int j = 0; j < dst->tools->length; j++) {
+			struct input_config_tool *dst_tool = dst->tools->items[j];
+			if (src_tool->type == dst_tool->type) {
+				dst_tool->mode = src_tool->mode;
+				goto tool_merge_outer;
+			}
+		}
+
+		struct input_config_tool *dst_tool = malloc(sizeof(*dst_tool));
```
RedSoxFan, line 167: style: `sizeof(struct input_config_tool)`

include/sway/config.h
```diff
@@ -11,8 +11,10 @@
 #include "list.h"
 #include "swaynag.h"
 #include "tree/container.h"
+#include "sway/input/tablet.h"
 #include "sway/tree/root.h"
 #include "wlr-layer-shell-unstable-v1-protocol.h"
+#include "wlr/types/wlr_tablet_tool.h"
```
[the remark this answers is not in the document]
Xyene: Oops, fixed, thanks.

sway/commands/input/tool_mode.c
Xyene: Hmm, the kernel style says this in section 14 ([link](https://www.kernel.org/doc/Documentation/process/coding-style.rst)):

> The preferred form for passing a size of a struct is the following:
>
 >     p = kmalloc(sizeof(*p), ...);
>
> The alternative form where struct name is spelled out hurts readability and
> introduces an opportunity for a bug when the pointer variable type is changed
> but the corresponding sizeof that is passed to a memory allocator is not.

This makes sense to me. While I can change this code to use `sizeof(struct input_config_tool)` etc. instead, in that case I think the divergence from the kernel style should be documented as part of CONTRIBUTING.md alongside other divergences.

daanjderuiter:
Not sure what kind of testing this patch is still waiting for, but building running it just now gives the expected functionality with my XP-PEN tablet (though only on @Xyene 's branch, which is a bit out of date). Only the config option in conjunction with a wildcard to select all tool modes does not work in my shell (zsh), though '*' does work with quotes (single and double). If you'd like me to do any additional testing let me know :)

sway/commands/input/tool_mode.c
RedSoxFan: I personally don't have any strong opinion one way or the other, but I do value consistency. Since we do use the `sizeof(struct input_config_tool)` style throughout the codebase, I'd rather continue with that. I can submit a PR tomorrow to add a note to CONTRIBUTING,md with this divergence.

RedSoxFan (commented):
Aside from the sizeof style, LGTM (from a code perspective - no hardware to test with)

sway/commands/input/tool_mode.c
RedSoxFan: Submitted #5672

emersion (approved):
LGTM, thanks!

Let's continue discussing the style issue in the other PR, and adapt the style if deemed necessary.

PR #5474, opened 2020-06-19
Introduce list_for_each utility macro

Xyene:
This PR introduces a `list_for_each` macro to make dealing with `list_t` more ergonomic (and less prone to swapping indices). It's based off `wl_array_for_each` from `wayland-util.h`, modified to accommodate `list_t`'s indirection.

Some existing places that I thought could benefit from the readability have been ported over, but this was best-effort and there are probably places that I've missed.

Alternative syntax possible here would be e.g.

```c
list_for_each(struct sway_container *child, container->children) {
    ...
}
```

which is one line shorter, but not symmetric with `wl_list_for_each`.

### DIFF
--- a/include/list.h
+++ b/include/list.h
@@ -27,6 +27,11 @@ void list_swap(list_t *list, int src, int dest);
 // move item to end of list
 void list_move_to_end(list_t *list, void *item);
 
+#define list_for_each(e, list)											\
+	for (__typeof__(e) *p = (__typeof__(e) *) (list)->items, (e) = *p;	\
+			(void **) p < ((list)->items + (list)->length);				\
+			p++, (e) = *p)
+
 /* Calls `free` for each item in the list, then frees the list.
  * Do not use this to free lists of primitives or items that require more
  * complicated deallocation code.
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -294,8 +294,8 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 					cmd_results_new(CMD_FAILURE, "No matching node."));
 		} else {
 			struct cmd_results *fail_res = NULL;
-			for (int i = 0; i < containers->length; ++i) {
-				struct sway_container *container = containers->items[i];
+			struct sway_container *container;
+			list_for_each(container, containers) {
 				set_config_node(&container->node);
 				struct cmd_results *res = handler->handle(argc-1, argv+1);
 				if (res->status == CMD_SUCCESS) {
@@ -527,8 +527,8 @@ void free_cmd_results(struct cmd_results *results) {
 
 char *cmd_results_to_json(list_t *res_list) {
 	json_object *result_array = json_object_new_array();
-	for (int i = 0; i < res_list->length; ++i) {
-		struct cmd_results *results = res_list->items[i];
+	struct cmd_results *results;
+	list_for_each(results, res_list) {
 		json_object *root = json_object_new_object();
 		json_object_object_add(root, "success",
 				json_object_new_boolean(results->status == CMD_SUCCESS));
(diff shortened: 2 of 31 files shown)


include/list.h
```diff
@@ -27,6 +27,11 @@ void list_swap(list_t *list, int src, int dest);
 // move item to end of list
 void list_move_to_end(list_t *list, void *item);
 
+#define list_for_each(e, list)											\
+	for (__typeof__(e) *p = (__typeof__(e) *) (list)->items, (e) = *p;	\
```
emersion, line 31: This uses non-standard C extensions and is complicated to read.

emersion:
In general we try to avoid using macros altogether because they tend to hide logic. I'm not 100% against a helper, but I'm not sure it's worth it: a simple loop with a counter is readable enough IMHO.

RedSoxFan:
This would be a -1 from me. It's too easy to misread as `wl_list_for_each`. I'm also not a huge fan of adding macros where they are not needed (as an example, logging is one of the few cases where it is useful since it allows for line numbers to be inserted). There just isn't enough benefit here to warrant the use of a macro.

Also just as a note, in the past we discussed transitioning anything that doesn't benefit from being an array list away from list_t and refactoring them to use wl_list. We haven't really executed this, but we do try to follow it for anything new.

RedSoxFan:
I'm going to close this since I doubt this will get approved. Thanks for taking the time to work on it and all the other contributions you've done though.

Xyene:
I wanted to put together a well-reasoned response to this even in the event that it didn't get merged, since there are some meta-points that I wanted to discuss -- but shortly after opening this PR I got DOS'd by real-life events and haven't had time to reply.

Anyway, here goes:

> This uses non-standard C extensions

Agreed, though I'm not sure how helpful being limited to a stricter subset of C (extensions) than libwayland itself uses is. While I understand the philosphical standpoint, when the most direct dependency of sway already requires `__typeof__`, it feels incongruous.

libwayland has a macro to hide list-traversing logic, as does the kernel which sway borrows most of its style guidelines from. Lists are ubiquitous; making their traversal standard and succint seems like an easy win for readability.

> It's too easy to misread as `wl_list_for_each`.

I can see that being the case, but I see that as a feature rather than a bug. It behaves like `wl_list_for_each` from the perspective of the caller, and if you were to accidentally use the wrong one, it'd fail to compile. And as a reader, you don't have to care.

It's a bit of a cliche that code is read more than it is written, but nonetheless I'd argue the similarity is a positive thing, since it reduces the effort a *reader* has to expend to understand what's going on. As a reader, whether it's `wl_list_for_each` or `list_for_each` is largely irrelevant; the semantics are the same.

To illustrate, the two examples below are cases where it'd be a bit *too* easy as-is to do the wrong thing and not notice, since they'll compile without warnings. They're pulled from existing code, and edited to be wrong.

```cpp
for (i = 0; i < config->input_type_configs->length; i++) {
    struct input_config *tc = config->input_type_configs->items[i];
    ...
}

for (i = 0; i < config->input_configs->length; i++) {
    struct input_config *tc = config->input_type_configs->items[i];
    ...
}
```

```cpp
for (int i = 0; i < mode->keysym_bindings->length; i++) {
    free_sway_binding(mode->keycode_bindings->items[i]);
```

These are all typos I have either made myself, or could see myself missing in any nontrivial review. Given how often code is copy-pasted with minor modifications, these examples are not that far-fetched.

If as a general software engineering heuristic we try to reduce the cognitive complexity of reading code to make reviews and future debugging sessions easier, this sort of verbosity runs contrary to that.

> we discussed transitioning anything that doesn't benefit from being an array list away from list_t and refactoring them to use wl_list.

I can see this helping in some places, but having extra intermediary structs or `link` members for structs that will ever be placed in a list doesn't seem like a clear-cut improvement over the status quo.

I'm not going to push further on this since I'm outnumbered, but did want to express these points anyway. It's my largest peeve when working on sway :)

PR #5407, opened 2020-05-31
tree/view: fix smart gaps when ancestor container is tabbed or stacked

Xyene:
Fixes #5406.

PR #5485, opened 2020-06-27
tree/view: fix smart borders with tabbed/stacked ancestor

### LINKED ISSUE #5484: smart_borders adds borders for split tab with single child
This looks like the `smart_borders` version of https://github.com/swaywm/sway/issues/5406.

Relevant config:

```
smart_borders on
```

A reproducing layout is `T[V[app] app]`, which looks like this in sway:

image

and like this in i3-gaps:

image

i3-gaps does not draw the border on the vertical container until it has more than one child.

FWIW, the current behavior does follow the [_documentation_](https://github.com/Airblader/i3/blob/6e04b07c3b2b72f1763ef3415564804ab39a076d/README.md#smart-borders) of the feature on i3's side, but not its actual behavior:

>  Smart borders will draw borders on windows only if there is more than one window in a **workspace**. This feature can also be enabled only if the gap size between window and screen edge is `0`.

(Emphasis mine.)

Still, I think the _behavior_ that i3-gaps implements makes more sense.

Xyene:
Fixes #5484.

---

Tested states (all match i3-gaps now):

image

image

image

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -210,6 +210,23 @@ bool view_ancestor_is_only_visible(struct sway_view *view) {
 	return only_visible;
 }
 
+static bool view_is_only_visible(struct sway_view *view) {
+	struct sway_container *con = view->container;
+	while (con) {
+		enum sway_container_layout layout = container_parent_layout(con);
+		if (layout != L_TABBED && layout != L_STACKED) {
+			list_t *siblings = container_get_siblings(con);
+			if (siblings && siblings->length > 1) {
+				return false;
+			}
+		}
+
+		con = con->parent;
+	}
+
+	return true;
+}
+
 static bool gaps_to_edge(struct sway_view *view) {
 	struct side_gaps gaps = view->container->workspace->current_gaps;
 	return gaps.top > 0 || gaps.right > 0 || gaps.bottom > 0 || gaps.left > 0;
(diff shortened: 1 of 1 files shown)


sway/tree/view.c
```diff
@@ -244,25 +261,32 @@ void view_autoconfigure(struct sway_view *view) {
...
-				|| config->hide_edge_borders == E_HORIZONTAL || hide_smart) {
+				|| config->hide_edge_borders == E_HORIZONTAL) {
 			con->border_top = con->y != ws->y;
 			int bottom_y = con->y + con->height;
 			con->border_bottom = bottom_y != ws->y + ws->height;
 		}
 
+		if (smart && view_ancestor_is_only_visible(view)) {
+			bool show_border = !view_is_only_visible_in_tab_or_stack(view);
+			con->border_left = show_border;
```
RedSoxFan, line 284: I believe these should be `con->border_left &= show_border`. See #4698 for more information.

sway/tree/view.c
```diff
@@ -244,25 +261,32 @@ void view_autoconfigure(struct sway_view *view) {
...
+
 		if (config->hide_edge_borders == E_BOTH
-				|| config->hide_edge_borders == E_HORIZONTAL || hide_smart) {
+				|| config->hide_edge_borders == E_HORIZONTAL) {
 			con->border_top = con->y != ws->y;
 			int bottom_y = con->y + con->height;
 			con->border_bottom = bottom_y != ws->y + ws->height;
 		}
 
+		if (smart && view_ancestor_is_only_visible(view)) {
```
RedSoxFan, line 282: I'm also not sure that both `view_ancestor_is_only_visible` and `view_is_only_visible_in_tab_or_stack` need to be called here. The two functions are only slightly different. It looks like the former should be used for smart gaps (#5407 - which also appears to be what caused this issue) and the latter for smart borders.

sway/tree/view.c
Xyene: Thanks, confirmed that the current code is misbehaving re: #4698 and that `&`-ing fixes it.

sway/tree/view.c
Xyene: Replacing the `view_ancestor_is_only_visible` with `view_is_only_visible_in_tab_or_stack` makes the top border still shown, since `con->y != ws->y` due to the tabs being in the way:

image

RedSoxFan: That makes sense for the original code. It doesn't look like the call to `view_ancestor_is_only_visible` is needed with this iteration though. It doesn't hit the case above anyway.

Xyene: Thanks, confirmed that's the case and removed the extra call.

RedSoxFan:
Thanks!

Issue #5116, opened 2020-03-17
Drag & drop doesn't focus target window


WhyNotHugo:
When I drag something (eg: an image) between two windows (A->B), once I drop/release, my mouse cursor is hovering on window B, but that window doesn't have focus. I need to move the mouse out and back again.

Oddly though, the source window doesn't seem to have the focus either.

I've verified this by drag-and-dropping...

- From `nemo` to `telegram`.
- From `nemo` to `firefox`.

These are all wayland-native apps.

I'm running sway 1.4.

Xyene:
I've been able to reproduce this on sway/wlroots master (currently 99897a6). The window borders change colors as if the window should have focus, but keyboard focus is not passed on drop.

PR #5489, opened 2020-06-28
seat: Refocus seat when wlr_drag is destroyed

kennylevinsen:
wlr_drag installs grabs for the full duration of the drag, leading to
the drag target not being focused when the drag ends. This leads to
unexpected focus behavior, especially for the keyboard which requires
toggling focus away and back to set.

We can only fix the focus once the grabs are released, so refocus the
seat when the wlr_drag destroy event is received.

Closes: https://github.com/swaywm/sway/issues/5116

Depends on: https://github.com/swaywm/wlroots/pull/2306

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -402,6 +402,31 @@ static void drag_icon_handle_destroy(struct wl_listener *listener, void *data) {
 	free(icon);
 }
 
+static void drag_handle_destroy(struct wl_listener *listener, void *data) {
+	struct sway_drag *drag = wl_container_of(listener, drag, destroy);
+
+	// Focus enter isn't sent during drag, so refocus the focused node, layer
+	// surface or unmanaged surface.
+	struct sway_seat *seat = drag->seat;
+	struct sway_node *focus = seat_get_focus(seat);
+	if (focus) {
+		seat_set_focus(seat, NULL);
+		seat_set_focus(seat, focus);
+	} else if (seat->focused_layer) {
+		struct wlr_layer_surface_v1 *layer = seat->focused_layer;
+		seat_set_focus_layer(seat, NULL);
+		seat_set_focus_layer(seat, layer);
+	} else {
+		struct wlr_surface *unmanaged = seat->wlr_seat->keyboard_state.focused_surface;
+		seat_set_focus_surface(seat, NULL, false);
+		seat_set_focus_surface(seat, unmanaged, false);
+	}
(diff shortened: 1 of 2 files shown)


sway/input/seat.c
```diff
@@ -402,6 +402,22 @@ static void drag_icon_handle_destroy(struct wl_listener *listener, void *data) {
...
 }
 
+static void drag_handle_destroy(struct wl_listener *listener, void *data) {
+	struct sway_drag *drag = wl_container_of(listener, drag, destroy);
+
+	// Focus doesn't update during the drag, so focus the drag target
+	if (drag->wlr_drag->focus) {
+		struct sway_view *focused = view_from_wlr_surface(drag->wlr_drag->focus);
+		if (focused != NULL && focused->container != NULL) {
+			seat_send_focus(&focused->container->node, drag->seat);
```
Xyene, line 412: Can a drag end with a layer surface or an Xwayland unmanaged surface as the target? It sounds to me like they can, and if that's the case this should have [some more logic](https://github.com/swaywm/sway/blob/b7a6cc506a952cfa20918ddc9304e580a0371b05/sway/input/keyboard.c#L556) for those cases (since they won't have `sway_view`s associated).

kennylevinsen: Yeah, good point. Ended up changing the logic to just mimic the keyboard leave logic. Just refocusing what is already focused is sufficient, and better than overriding focus logic locally.

sway/input/seat.c
```diff
@@ -402,6 +402,31 @@ static void drag_icon_handle_destroy(struct wl_listener *listener, void *data) {
 	free(icon);
 }
 
+static void drag_handle_destroy(struct wl_listener *listener, void *data) {
+	struct sway_drag *drag = wl_container_of(listener, drag, destroy);
+
+	// Focus enter isn't sent during drag, so refocus the focused node, layer
+	// surface or unmanaged surface.
```
emersion, line 409: Would it be more appropriate to call `cursor_rebase` here?

I guess maybe that'd only work if focus-follows-mouse is enabled?

kennylevinsen: Won't that only affect the cursor? The cursor *is* affected by this issue, but the biggest problem is keyboard focus, which requires a full leave and re-enter to fix.

emersion: Right, I see, this makes sense.

PR #5498, opened 2020-07-01
 commands/move: unwrap workspace container on move to new workspace 

### LINKED ISSUE #5426: Extra container created when moving tab groups with `workspace_layout tabbed`
This doesn't match i3 behavior. 

To reproduce, create a layout like this one:

image

Then, <kbd>mod</kbd>+<kbd>a</kbd>, <kbd>mod</kbd>+<kbd>a</kbd> to select the topmost container.

image

Finally, <kbd>Ctrl</kbd>+<kbd>mod</kbd>+<kbd>empty workspace number</kbd>, and switch to that workspace.

Result in Sway:

image

Result in i3:

image

Sway creates the extra `H[T[kitty kitty] T[kitty kitty]]` container, where i3 doesn't. To replicate i3 behavior, <kbd>mod</kbd>+<kbd>a</kbd>, <kbd>mod</kbd>+<kbd>a</kbd>, <kbd>mod</kbd>+<kbd>e</kbd> has to be executed after the workspace switch.

I think the particular case of moving to an empty workspace is special-cased by i3 for ergonomic purposes. Sway behaves the same as i3 when the target workspace already has a tiling container in it.

Xyene:
If moving e.g. `T[app app]` into a new workspace with
`workspace_layout tabbed`, then post-`arrange_workspace` the tree in
that workspace will be `T[T[app app]]`. This still happens with
horizontal or vertical workspace layout, but is less visible since those
containers have no decorations.

Fixes #5426.

### DIFF
--- a/include/sway/tree/workspace.h
+++ b/include/sway/tree/workspace.h
@@ -105,6 +105,9 @@ struct sway_container *workspace_find_container(struct sway_workspace *ws,
  */
 struct sway_container *workspace_wrap_children(struct sway_workspace *ws);
 
+void workspace_unwrap_children(struct sway_workspace *ws,
+		struct sway_container *wrap);
+
 void workspace_detach(struct sway_workspace *workspace);
 
 void workspace_add_tiling(struct sway_workspace *workspace,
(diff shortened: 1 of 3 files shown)


RedSoxFan:
Thanks!

PR #5500, opened 2020-07-01
commands/move: maintain workspace_layout when moving

### LINKED ISSUE #5157: Different behavior of workspace_layout option from i3
Running sway 1.4_1 on Void Linux.
One of the features which I was used to in i3 was the ability to set the default workspace layout as stacking/tabbed because I only need tiling when I need to look at two or more windows at once. However, while using sway I encountered several different behavior which cause inconvenience while using this feature.

## Steps to reproduce:
1. Set 'workspace_layout stacking' in config.
2. Starting from a fresh workspace, spawn two or more windows in the same stack.
3. Move one of the windows right of the screen.
4. Move the same window left.
5. Close all windows.
6. Spawn multiple windows.

## What happens in Sway:
After moving the window left (step 4),
* In case two windows were spawned in step 2, the window on the right swaps positions with the window on the left. (Unexpected behavior)
* In case more than two windows were spawned in step 2, the window on the right gets merged back into the window stack on the left.

After spawning multiple windows (step 6), the windows spawn in vertical split. (Unexpected behavior)

## What happens in i3/What should happen in sway:
After moving the window left (step 4), the window on the right always gets merged back into the window stack on the left.

After spawning multiple windows (step 6), the windows spawn in stacking mode.

Xyene:
Fixes #5157.

---

I think there are definitely cases I've missed here, but at least Sway does generate compatible layout trees as per the linked issue.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -271,12 +271,11 @@ static void workspace_rejigger(struct sway_workspace *ws,
 		return;
 	}
 	container_detach(child);
-	struct sway_container *new_parent = workspace_wrap_children(ws);
+	workspace_wrap_children(ws);
 
 	int index =
 		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_UP ? 0 : 1;
 	workspace_insert_tiling(ws, child, index);
-	container_flatten(new_parent);
 	ws->layout =
 		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_RIGHT ?
 		L_HORIZ : L_VERT;
@@ -345,8 +344,11 @@ static bool container_move_in_direction(struct sway_container *container,
 							container_insert_child(current->parent, container,
 									index + (offs < 0 ? 0 : 1));
 						} else {
-							workspace_insert_tiling(current->workspace, container,
-									index + (offs < 0 ? 0 : 1));
+							struct sway_workspace *ws = current->workspace;
+							workspace_insert_tiling(ws,
+								container_split(container,
+									output_get_default_layout(ws->output)),
+								index + (offs < 0 ? 0 : 1));
 						}
 						return true;
 					}
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1329,6 +1329,14 @@ void container_detach(struct sway_container *child) {
 		container_update_representation(old_parent);
 		node_set_dirty(&old_parent->node);
 	} else if (old_workspace) {
+		// We may have removed the last tiling child from the workspace. If the
+		// workspace layout was e.g. tabbed, then at this point it may be just
+		// H[]. So, reset it to the default (e.g. T[]) for next time.
+		if (!old_workspace->tiling->length) {
+			old_workspace->layout =
+				output_get_default_layout(old_workspace->output);
+		}
+
 		workspace_update_representation(old_workspace);
 		node_set_dirty(&old_workspace->node);
 	}


sway/commands/move.c
```diff
@@ -271,12 +271,11 @@ static void workspace_rejigger(struct sway_workspace *ws,
...
 	}
 	container_detach(child);
-	struct sway_container *new_parent = workspace_wrap_children(ws);
+	workspace_wrap_children(ws);
 
 	int index =
 		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_UP ? 0 : 1;
 	workspace_insert_tiling(ws, child, index);
-	container_flatten(new_parent);
 	ws->layout =
```
Xyene, line 279: We shouldn't be doing this flattening: if we had `T[app app]` to begin with, by now we have `H[T[app] T[app]]`. If we flatten the old workspace, we end up with `H[app T[app]]`, which is incorrect.

Xyene: `git blame` shows that this flattening was added in https://github.com/swaywm/sway/pull/2695. However, the two test cases in that PR continue to work now even without this flattening.

sway/commands/move.c
```diff
@@ -127,8 +127,11 @@ static void container_move_to_container_from_direction(
 			if (destination->parent) {
 				container_insert_child(destination->parent, container, index);
 			} else {
-				workspace_insert_tiling(destination->workspace,
-						container, index);
+				struct sway_workspace *ws = destination->workspace;
```
Xyene, line 130: This branch needed changing because if we had `T[focused unfocused]` and ran `move left`, we'd generate `H[H[focused] T[unfocused]]`. Now we generate `H[T[focused] T[unfocused]]`. The other split change covers the `move right` case.

Xyene:
> I'm getting a split instead of a stack.

Me too, I'll see what I can do. I think here i3's "hack" of always having the workspace be `H` comes in handy, but I think changing that behavior here without tests is asking for trouble.

> Another case to consider

Hmm, I can't reproduce this. In i3 with `workspace_layout stacking`, I'm getting `H[S[app1] S[app2]]`, after `move right` with `app2` having focus.

<details><summary>Visuals</summary>

* `S[app1 app2]`

image

* `move right` &rarr; `H[S[app1] S[app2]]`

image

</details>

> _Aside: I kinda wish we had automated testing for this._

Same, I saw there was https://github.com/swaywm/sway-tests -- what happened to that?

RedSoxFan:
> > Another case to consider
> 
> Hmm, I can't reproduce this. In i3 with `workspace_layout stacking`, I'm getting `H[S[app1] S[app2]]`, after `move right` with `app2` having focus.

Oh sorry that was with the default `workspace_layout`  for both. Forgot to mentioned that. Regardless what the `workspace_layout` is set to, I get a stack on the right in Sway.

> > _Aside: I kinda wish we had automated testing for this._
> 
> Same, I saw there was https://github.com/swaywm/sway-tests -- what happened to that?

Those were never maintained/supported. It was just a side project by one of the larger contributors at the time. I don't think it really went anywhere.

sway/tree/workspace.c
```diff
@@ -657,11 +657,18 @@ static void set_workspace(struct sway_container *container, void *data) {
 	container->workspace = container->parent->workspace;
 }
 
+static void workspace_reset_layout(struct sway_workspace *workspace) {
+	if (!workspace->tiling->length) {
+		workspace->layout = output_get_default_layout(workspace->output);
```
Xyene, line 662: Hooked this into adding a tiling child to the workspace. This means that when we've emptied the workspace, we might still have its representation be `H[]` until the next child is added, when it'll turn back into `S[app]`. This is not ideal, but as far as I can tell we don't have a reliable "workspace has emptied" or "workspace container deleted" signal.

Xyene:
> Oh sorry that was with the default `workspace_layout` for both.

Hmm, I still can't reproduce this... maybe it incidentally got fixed through my last force push?

<details><summary>Visuals</summary>

* `S[app app]`

image

* `move right` &rarr; `H[S[app] app]`

image

</details>

sway/tree/workspace.c
RedSoxFan: `container_detach` may be a better spot? I *think* that every tiling container get removed using that.

RedSoxFan:
> Hmm, I still can't reproduce this... maybe it incidentally got fixed through my last force push?

Yup. Looks like it did. I can reproduce on  2d8e453, but not  ea46ac1

sway/tree/workspace.c
Xyene: I was worried that some places may be poking into `workspace->tiling` directly, but it looks like everywhere that ends up calling `workspace_update_representation` does use `container_detach` before that, so I've moved that logic there.

RedSoxFan:
Thanks!

Issue #5488, opened 2020-06-28
Maximized X11 app crashes Sway


jbeich:
Regressed by 8d5e627bc98f376f84e7f5b0a7caed791351c577

```c
$ mkdir -p /tmp/ffprofile/xulstore
$ mdb_load -s db -T /tmp/ffprofile/xulstore <<\EOF
chrome://browser/content/browser.xhtml\09main-window\09width
\07\04\00\00\00\00\00\00\001024
chrome://browser/content/browser.xhtml\09main-window\09height
\07\03\00\00\00\00\00\00\00768
chrome://browser/content/browser.xhtml\09main-window\09sizemode
\07\09\00\00\00\00\00\00\00maximized
EOF

$ cat >/tmp/sway.conf
exec env -u MOZ_ENABLE_WAYLAND firefox --profile /tmp/ffprofile about:blank

$ sway --version
sway version 1.5-rc1-1-gc8224270
wlroots version 0.10.0-216-g6c8f66ff

$ sway -c /tmp/sway.conf
Process 62085 stopped
* thread #1, name = 'sway', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
    frame #0: 0x0000000800bb457f libc.so.7`strlen(str=0x0000000000000000) at strlen.c:101:8
   98            * boundaries is integral multiple of word size.
   99            */
   100          lp = (const unsigned long *)((uintptr_t)str & ~LONGPTR_MASK);
-> 101          va = (*lp - mask01);
   102          vb = ((~*lp) & mask80);
   103          lp++;
   104          if (va & vb)
(lldb) bt
* thread #1, name = 'sway', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
  * frame #0: 0x0000000800bb457f libc.so.7`strlen(str=0x0000000000000000) at strlen.c:101:8
    frame #1: 0x0000000800abe182 libc.so.7`strdup(str=0x0000000000000000) at strdup.c:48:8
    frame #2: 0x0000000800903d56 libwlroots.so.5`wlr_foreign_toplevel_handle_v1_set_title(toplevel=0x00000008039e0700, title=0x0000000000000000) at wlr_foreign_toplevel_management_v1.c:197:20
    frame #3: 0x000000000028aa81 sway`view_update_title(view=0x0000000802d59b00, force=false) at view.c:1148:3
[...]

PR #5506, opened 2020-07-02
tree/view: fix segfault in view_update_title

emersion:
xdg-shell doesn't allow clients to set the title to NULL, so we
shouldn't need to call wlr_foreign_toplevel_handle_v1_set_title with an
empty string to reset the old one.

Closes: https://github.com/swaywm/sway/issues/5488

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -1171,7 +1171,7 @@ void view_update_title(struct sway_view *view, bool force) {
 
 	ipc_event_window(view->container, "title");
 
-	if (view->foreign_toplevel) {
+	if (view->foreign_toplevel && title) {
 		wlr_foreign_toplevel_handle_v1_set_title(view->foreign_toplevel, title);
 	}
 }


sway/tree/view.c
```diff
@@ -1171,7 +1171,7 @@ void view_update_title(struct sway_view *view, bool force) {
 
 	ipc_event_window(view->container, "title");
 
-	if (view->foreign_toplevel) {
+	if (view->foreign_toplevel && title != NULL) {
```
jbeich, line 1174: Why mix 2 different styles of `NULL` check?
```suggestion
	if (view->foreign_toplevel && title) {
```

jbeich: Elsewhere in the file `title` is more common than `title != NULL`:
```
sway/tree/view.c
1141:           if (title && view->container->title &&
1145:           if (!title && !view->container->title) {
1152:   if (title) {
```

Xyene: Sway code in general is not super consistent on this (I suspect to avoid bikeshedding given there's no linter), but I agree it would be nice to keep things consistent within the same statement. @emersion, up to you, the PR looks functionally correct to me :)

emersion: In general we don't have a preferred NULL check style in Sway/wlroots. However I agree that mixing two styles in the same condition is too inconsistent, applied your suggestion.

PR #5517, opened 2020-07-04
Reload Command loads config file in order of priority described in man pages

Geo25rey:
Reload Command now loads config file with most priority instead of currently loaded config file

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -500,6 +500,7 @@ struct sway_config {
 	struct side_gaps gaps_outer;
 
 	list_t *config_chain;
+	bool user_config_path;
 	const char *current_config_path;
 	const char *current_config;
 	int current_config_line_number;
--- a/sway/commands/reload.c
+++ b/sway/commands/reload.c
@@ -22,7 +22,12 @@ static void do_reload(void *data) {
 		list_add(bar_ids, strdup(bar->id));
 	}
 
-	if (!load_main_config(config->current_config_path, true, false)) {
+	const char *path = NULL;
+	if (config->user_config_path) {
+		path = config->current_config_path;
+	}
+
+	if (!load_main_config(path, true, false)) {
 		sway_log(SWAY_ERROR, "Error(s) reloading config");
 		list_free_items_and_destroy(bar_ids);
 		return;
@@ -55,7 +60,12 @@ struct cmd_results *cmd_reload(int argc, char **argv) {
 		return error;
 	}
 
-	if (!load_main_config(config->current_config_path, true, true)) {
+	const char *path = NULL;
+	if (config->user_config_path) {
+		path = config->current_config_path;
+	}
+
+	if (!load_main_config(path, true, true)) {
 		return cmd_results_new(CMD_FAILURE, "Error(s) reloading config.");
 	}
 
--- a/sway/config.c
+++ b/sway/config.c
@@ -447,6 +447,7 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 		}
 	}
 
+	config->user_config_path = file ? true : false;
 	config->current_config_path = path;
 	list_add(config->config_chain, real_path);
 
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -272,7 +272,9 @@ set|plus|minus <amount>
 	optional comment argument is ignored, but logged for debugging purposes.
 
 *reload*
-	Reloads the sway config file and applies any changes.
+	Reloads the sway config file and applies any changes. The config file is
+	located at path specified by the command line arguments when started,
+	otherwise according to the priority stated in *sway*(1).
 
 *rename workspace* [<old_name>] to <new_name>
 	Rename either <old_name> or the focused workspace to the <new_name>


sway/commands/reload.c
```diff
@@ -22,7 +22,9 @@ static void do_reload(void *data) {
 		list_add(bar_ids, strdup(bar->id));
 	}
 
-	if (!load_main_config(config->current_config_path, true, false)) {
+//	if (!load_main_config(config->current_config_path, true, false)) {
```
TheAvidDev, line 25: This line should be removed instead of commented out, and I don't think the explanatory comment below it is necessary either.

sway/commands/reload.c
```diff
@@ -55,7 +57,9 @@ struct cmd_results *cmd_reload(int argc, char **argv) {
 		return error;
 	}
 
-	if (!load_main_config(config->current_config_path, true, true)) {
+//	if (!load_main_config(config->current_config_path, true, true)) {
```
TheAvidDev, line 60: Same as above.

jhalmen:
why would you want this?

consider starting a nested instance with a different config, reloading would mess it up if this got merged.

Geo25rey:
> why would you want this?
> 
> consider starting a nested instance with a different config, reloading would mess it up if this got merged.

@jhalmen What do you mean by "reloading would mess it up if this got merged"? What is "it"?

jhalmen:
"it" would be the nested instance.
consider i have my main config for my main sway instance, and a secondary config for my nested instance (with different keybindings, so that i can actually do stuff in this nested instance)
if i now update this secondary config and then reload my nested instance this pr would make it so my nested instance has all the same keybindings as my main config (because this main config is in the 'highest priority location'). this means that i cannot interact with my nested instance through keybindings anymore, because they all get swallowed by my main instance.

does this pr actually solve an issue you have?

Geo25rey:
> "it" would be the nested instance.
> consider i have my main config for my main sway instance, and a secondary config for my nested instance (with different keybindings, so that i can actually do stuff in this nested instance)
> if i now update this secondary config and then reload my nested instance this pr would make it so my nested instance has all the same keybindings as my main config (because this main config is in the 'highest priority location'). this means that i cannot interact with my nested instance through keybindings anymore, because they all get swallowed by my main instance.
> 
> does this pr actually solve an issue you have?

Yeah I launch sway for the first time with the default config in /etc/sway/config, then from within sway I should just be able to reload and get the config from ~/.sway/config.

Geo25rey:
@jhalmen Let me know what you think

TheAvidDev:
> Yeah I launch sway for the first time with the default config in /etc/sway/config, then from within sway I should just be able to reload and get the config from ~/.sway/config.

What's the reason for doing this?

Geo25rey:
> > Yeah I launch sway for the first time with the default config in /etc/sway/config, then from within sway I should just be able to reload and get the config from ~/.sway/config.
> 
> What's the reason for doing this? Shouldn't you only have to do it once?

I suppose the reasoning is user ignorance. I just started using sway (started from a tty) without a making a config. I figured like many other Unix-like applications that the config was optional and would just use the default in /etc, which is does. But I figured after "reloading" the config, sway would load the custom config in my home directory.

TheAvidDev:
> Does that answer your questions? @TheAvidDev

Yep, that makes sense, not really sure if this is a good idea since others bring up good argument against -- I'm not a core contributor or anything though, so take my opinion with a grain of salt.

stevenengler:
Fwiw, I had the same problem, and it wasn’t a good introduction to Sway :) After spending a good hour or so trying to figure out what was wrong, it wasn’t until I logged out and back in again before I realized why it wasn’t working. You don’t run into this problem with i3 since it gives you a configuration file by default on first launch. I don’t know the best solution to this, but I think it’s a problem worth thinking about.

Geo25rey:
As of my most recent commit, the implementation works as follows: if there is a specific config file set from the command line arguments, then it will reload that from that specific location, otherwise, sway will load a config file according to the priority described in `man sway`.

Geo25rey:
@RedSoxFan can you take a look?

RedSoxFan:
> if there is a specific config file set from the command line arguments, then it will reload that from that specific location, otherwise, sway will load a config file according to the priority

This actually does appear to match i3's implementation so this would be i3-compat. However, this does not match our code style (https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#style-reference). Please use tabs for indentation.

When you have that resolved, can you please squash the commits into one and force push?

include/sway/config.h
```diff
@@ -499,6 +499,8 @@ struct sway_config {
 	int gaps_inner;
 	struct side_gaps gaps_outer;
 
+    bool user_config_path;
```
RedSoxFan, line 502: This can get moved down into the group below. Probably in between `config_chain` and `current_config_path`.

sway/commands/reload.c
```diff
@@ -22,7 +22,14 @@ static void do_reload(void *data) {
 		list_add(bar_ids, strdup(bar->id));
 	}
 
-	if (!load_main_config(config->current_config_path, true, false)) {
+    const char *path;
```
RedSoxFan, line 25: This could be simplified to the following (a ternary would also work, but this looks better due to the line breaks):
```c
const char *path = NULL;
if (config->user_config_path) {
    path = config->current_config_path
}
```

sway/commands/reload.c
```diff
@@ -55,7 +62,14 @@ struct cmd_results *cmd_reload(int argc, char **argv) {
 		return error;
 	}
 
-	if (!load_main_config(config->current_config_path, true, true)) {
+    const char *path;
```
RedSoxFan, line 65: Same as above:
```c
const char *path = NULL;
if (config->user_config_path) {
    path = config->current_config_path
}
```

sway/config.c
```diff
@@ -447,6 +447,8 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 		}
 	}
 
+    config->user_config_path = file ? true : false;
+
```
RedSoxFan, line 451: No need for the empty line

Geo25rey:
@RedSoxFan I got it all done. Thanks for your input.

include/sway/config.h
```diff
@@ -500,6 +500,7 @@ struct sway_config {
 	struct side_gaps gaps_outer;
 
 	list_t *config_chain;
+ 	bool user_config_path;
```
RedSoxFan, line 503: There is a space at the beginning of this line. Other than that, LGTM

Geo25rey:
@RedSoxFan I promise it's good this time

RedSoxFan:
Thanks!

PR #5521, opened 2020-07-06
config/output: reconfigure on commit failure

### LINKED ISSUE #5483: 1.5-rc1: Segfault in ipc_json_describe_workspace
* Sway Version: sway version 1.5-rc1-ba0232e6 (Jun 25 2020, branch 'master')
* Debug Log: https://gist.githubusercontent.com/mortie/05dffd169580706b6bd3c396ac35e1a1/raw/344e3db673b4fde235ed4aed93f76fe88e18a859/sway-crash.log
* Configuration File: https://gist.githubusercontent.com/mortie/05dffd169580706b6bd3c396ac35e1a1/raw/ab25515e3474e0db895d6535f04b137d0a23c6e5/config
* Stack Trace, if sway crashes: https://gist.githubusercontent.com/mortie/05dffd169580706b6bd3c396ac35e1a1/raw/344e3db673b4fde235ed4aed93f76fe88e18a859/sway-crash.backtrace

I was playing around with the new headless output feature. I created an output with `swaymsg create_output`, then played around a bit, removed the output with `swaymsg output HEADLESS-1 disable`, reloaded the configuration, then repeated the last two steps a couple of times.

If you want a coredump and the exact executable I built, just ask and I will send it in a PM.

RedSoxFan:
Fixes #5483 

*Note that this only addresses the crash. I opened #5522 to provide a
way to destroy the headless output.*

If the wlr_output_commit fails and the output was previously enabled, we
re-enable the output. If the output was previously enabled, then it was
also configured. Without reenabling or reconfiguring the output, the
output is left in an invalid state. This adds the missing reconfiguring
step.

This also adds safe guards in output_disable to prevent removing an
output index that is not in the list.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -397,13 +397,21 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 
 	struct wlr_output *wlr_output = output->wlr_output;
 
+	// Store some information so we can restore in case the commit fails.
 	bool was_enabled = output->enabled;
+	bool was_configured = output->configured;
+	double old_x = 0, old_y = 0;
+
 	if (oc && !oc->enabled) {
 		// Output is configured to be disabled
 		sway_log(SWAY_DEBUG, "Disabling output %s", oc->name);
 		if (output->enabled) {
 			output_disable(output);
-			wlr_output_layout_remove(root->output_layout, wlr_output);
+			if (was_configured) {
+				wlr_output_layout_output_coords(
+						root->output_layout, wlr_output, &old_x, &old_y);
+				wlr_output_layout_remove(root->output_layout, wlr_output);
+			}
 		}
 	} else {
 		output->enabled = true;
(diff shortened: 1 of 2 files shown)


emersion:
Hmm, this is getting a little bit complicated. I'm not a huge fan of this approach because we need to save/restore more output properties than before. This logic would be nicer if we only marked an output as enabled once the commit was successful. What was the reason why we don't do this? Because of `output.events.mode/commit/etc` events maybe?

RedSoxFan:
> Hmm, this is getting a little bit complicated. I'm not a huge fan of this approach because we need to save/restore more output properties than before. 

Yeah. Unfortunately.

> This logic would be nicer if we only marked an output as enabled once the commit was successful. What was the reason why we don't do this? Because of `output.events.mode/commit/etc` events maybe?

The case where an output is disabled and the commit fails for enabling it is easy. We just flip the enabled boolean back, which does need to be enabled for the output event handlers.

The issue and complication is actually the other way - the output was enabled and the commit to disable it fails. This case should be much more rare, but still should be handled. If we don't re-enable and reconfigure the output, then we are left in a state where wlroots thinks the output is enabled, but sway doesn't. This leads to the output being frozen on the last frame or just the default swaybg gray, which can cascade to more problems due to the invalid state if the output gets disconnected or the user issues an enable for the output.

sway/config/output.c
```diff
@@ -397,13 +397,21 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
...
+	// Store some information so we can restore in case the commit fails.
 	bool was_enabled = output->enabled;
+	bool was_configured = output->configured;
+	double old_x = 0, old_y = 0;
+
 	if (oc && !oc->enabled) {
 		// Output is configured to be disabled
 		sway_log(SWAY_DEBUG, "Disabling output %s", oc->name);
 		if (output->enabled) {
 			output_disable(output);
```
emersion, line 409: What about only calling `output_disable` once the output commit succeeds?

emersion:
>The issue and complication is actually the other way - the output was enabled and the commit to disable it fails. This case should be much more rare, but still should be handled. If we don't re-enable and reconfigure the output, then we are left in a state where wlroots thinks the output is enabled, but sway doesn't. This leads to the output being frozen on the last frame or just the default swaybg gray, which can cascade to more problems due to the invalid state if the output gets disconnected or the user issues an enable for the output.

Sure. I was mainly asking why we need to flip the `enabled` flag _before_ the output commit. If we can flip it _after_, then no need for complicated rollback logic.

It turns out we have a `handle_mode` listener which needs this flag to know whether the backend turned the output on on its own or not. Ref https://github.com/swaywm/wlroots/issues/2300 and https://github.com/swaywm/wlroots/issues/2094.

PR #5525, opened 2020-07-07
input: implement xdg_toplevel interactive resize hints

### LINKED ISSUE #5411: Sway doesn't set the XDG_TOPLEVEL_STATE_RESIZING flag when actively resizing
Clients simply get the new size via handle_toplevel_config without the resize flag being set.
While that's normal when switching between different sizes (e.g. floating to tiled), when doing a live resize, such as when in resize mode, the XDG_TOPLEVEL_STATE_RESIZING flag ought to be signaled in order to allow clients to perform less expensive resizes (like not rebuilding a suboptimal swapchain with Vulkan).
The spec isn't entirely clear on whether the client should get a new handle_toplevel_config after stopping live resizing to clear their state, or whether the client should do a proper resize after 1 or N frames and clear their state, but getting a new handle_toplevel_config event with the same size as the current but the flag cleared would be perfectly fine and not break anything.
Mutter implements this, and mpv also implements this as a client.

RPigott:
This PR implements interactive resize hints for xdg_toplevels by setting the resize state on all clients participating in an interactive resize. Should resolve #5411. The issues made it sounds ambiguous, but by my reading of xdg-shell.xml clients _should_ receive a non-resizing configure at the end:

> xdg_toplevel resize request:
> [...]
> After the resize is completed, the client will receive another "configure" event without the resize state.

these resizes don't involve the request, but I think it makes sense.

I made a little test client that turns green when resize hint is set, gif below. PTAL.

output

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -746,6 +746,28 @@ void container_floating_set_default_size(struct sway_container *con) {
 	free(box);
 }
 
+
+/**
+ * Indicate to clients in this container that they are participating in (or
+ * have just finished) an interactive resize
+ */
+void container_set_resizing(struct sway_container *con, bool resizing) {
+	if (!con) {
+		return;
+	}
+
+	if (con->view) {
+		if (con->view->impl->set_resizing) {
+			con->view->impl->set_resizing(con->view, resizing);
+		}
+	} else {
+		for (int i = 0; i < con->children->length; ++i ) {
+			struct sway_container *child = con->children->items[i];
+			container_set_resizing(child, resizing);
+		}
+	}
+}
+
 void container_set_floating(struct sway_container *container, bool enable) {
 	if (container_is_floating(container) == enable) {
 		return;
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -219,6 +219,8 @@ void container_floating_resize_and_center(struct sway_container *con);
 
 void container_floating_set_default_size(struct sway_container *con);
 
+void container_set_resizing(struct sway_container *con, bool resizing);
+
 void container_set_floating(struct sway_container *container, bool enable);
 
 void container_set_geometry_from_content(struct sway_container *con);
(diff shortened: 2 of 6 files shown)


sway/tree/container.c
```diff
@@ -746,6 +747,29 @@ void container_floating_set_default_size(struct sway_container *con) {
...
+ * Indicate to clients in this container that they are participating in (or
+ * have just finished) an interactive resize
+ */
+void container_set_resizing(struct sway_container *con, bool resizing) {
+	if (!con) {
+		return;
+	}
+
+	if (con->view) {
+		if (con->view->type == SWAY_VIEW_XDG_SHELL &&
```
RedSoxFan, line 761: I think we have all of the `xdg_shell` specific code in `sway/desktop/xdg_shell.c`. There should not be any shell specific code in this file. So to follow that pattern, `set_resizing` should be added to `sway_view_impl`, This block would then be moved to `sway/desktop/xdg_shell.c` and would be called from here with:

```c
if (con->view->impl->set_resizing) {
    con->view->impl->set_resizing(con->view, resizing);
}
```

*Aside: actually, it looks like `surface_is_popup` and `surface_at_view` both do have shell specific code. Both of those should probably be refactored in the same way, but that's a different issue.*

sway/tree/container.c
```diff
@@ -675,6 +675,7 @@ static void floating_natural_resize(struct sway_container *con) {
 	}
 }
 
+
```
RedSoxFan, line 678: Accidental add?

sway/tree/container.c
RPigott: Yeah, that sounds good. I'll rearrange it this way.

Only slightly related, but do you know if there's something to implement for `set_resizing` on other views? Like, do X clients have something similar that would work with Xwayland? I don't really know about X.

sway/tree/container.c
RPigott: Oops, yeah.

sway/tree/container.c
RedSoxFan: I have no idea. For this PR, you can just implement it for xdg-shell and xwayland can be implemented later if needed.

PR #5535, opened 2020-07-10
config/output: don't change output state before commit

emersion:
Previously, we called output_disable prior to wlr_output_commit. This
mutates Sway's output state before the output commit actually succeeds.
This results in Sway's state getting out-of-sync with wlroots'.

An alternative fix [1] was to revert the changes made by output_disable
in case of failure. This is a little complicated. Instead, this patch
makes it so Sway's internal state is never changed before a successful
wlr_output commit.

We had two output flags: enabled and configured. However enabled was set
prior to the output becoming enabled, and was used to prevent the output
event handlers (specifically, the mode handler) from calling
apply_output_config again (infinite loop).

Rename enabled to enabling and use it exclusively for this purpose.
Rename configured to enabled, because that's what it really means.

[1]: https://github.com/swaywm/sway/pull/5521

Closes: https://github.com/swaywm/sway/issues/5483

cc @RedSoxFan

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -36,7 +36,7 @@ struct sway_output {
 	// last applied mode when the output is DPMS'ed
 	struct wlr_output_mode *current_mode;
 
-	bool enabled, configured;
+	bool enabling, enabled;
 	list_t *workspaces;
 
 	struct sway_output_state current;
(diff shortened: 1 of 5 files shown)


RedSoxFan:
Thanks!

PR #5537, opened 2020-07-11
swaybar: allow status line cleanup to proceed when hidden

### LINKED ISSUE #5536: Hidden swaybar with running status command can’t exit
- **Sway Version:** 1.4

- **Configuration:**

    ```
    bar {
        ⋮
        mode invisible
        ⋮
    }
    ```

     (or `mode hide`, or any means of hiding the bar)

- **Description:**

    Reloading or exiting Sway leaves behind a running swaybar and stopped status command each time.

charmander:
`determine_bar_visibility` stops and starts the status command process according to the bar’s visibility. If the bar was hidden during teardown, teardown would stall while waiting for the stopped status command process to exit.

This resumes a stopped status command during teardown and allows, for example, sway to reload or quit without leaving a swaybar instance behind each time.

Fixes #5536.

### DIFF
--- a/swaybar/status_line.c
+++ b/swaybar/status_line.c
@@ -185,6 +185,7 @@ struct status_line *status_line_init(char *cmd) {
 
 void status_line_free(struct status_line *status) {
 	status_line_close_fds(status);
+	kill(status->pid, status->cont_signal);
 	kill(status->pid, SIGTERM);
 	waitpid(status->pid, NULL, 0);
 	if (status->protocol == PROTOCOL_I3BAR) {


swaybar/status_line.c
```diff
@@ -186,6 +186,7 @@ struct status_line *status_line_init(char *cmd) {
 void status_line_free(struct status_line *status) {
 	status_line_close_fds(status);
 	kill(status->pid, SIGTERM);
+	kill(status->pid, SIGCONT);
```
emersion, line 189: Can we send `status->cont_signal` instead of `SIGCONT`? It seems like the continuation signal is configurable.

Nit: I'd argue it makes a little bit more sense to send `SIGCONT` before `SIGTERM`.

RedSoxFan: I'm going to unresolve this conversation since it looks like the following still needs to be addressed:

> Nit: I'd argue it makes a little bit more sense to send SIGCONT before SIGTERM

charmander: Thanks for pointing it out!

PR #5539, opened 2020-07-12
Make the default workspace commands compatible with i3

whhone:
**Problem**:
When I rename the workspace to something like "1:web", `$mod+1` does not move to the "1:web" with the default config. This breaks the expectation of i3 users.

**Cause**:
The default Sway binding for `$mod+1` does not have the `number` keyword:

```
bindsym $mod+1 workspace 1
```

Instead, the default Sway binding for `$mod+1` is

```
bindsym Mod1+1 workspace number $ws1
```

https://github.com/i3/i3/commit/e6662df114329ba45bd1d117c731b3dc8bdd13fb is the corresponding commit from i3.

### DIFF
--- a/config.in
+++ b/config.in
@@ -112,27 +112,27 @@ output * bg @datadir@/backgrounds/sway/Sway_Wallpaper_Blue_1920x1080.png fill
 # Workspaces:
 #
     # Switch to workspace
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Thanks!

PR #5543, opened 2020-07-12
Assert output is found before removing from list

emersion:
References: https://github.com/swaywm/sway/issues/5483

### DIFF
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -251,6 +251,11 @@ void output_disable(struct sway_output *output) {
 	if (!sway_assert(output->enabled, "Expected an enabled output")) {
 		return;
 	}
+	int index = list_find(root->outputs, output);
+	if (!sway_assert(index >= 0, "Output not found in root node")) {
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


sway/tree/output.c
```diff
@@ -259,6 +259,9 @@ void output_disable(struct sway_output *output) {
 	root_for_each_container(untrack_output, output);
 
 	int index = list_find(root->outputs, output);
+	if (!sway_assert(index >= 0, "Output not found in root node")) {
```
RedSoxFan, line 262: After thinking about it more, I have a couple concerns:
- I wonder if this assertion should actually be moved up in the function to be before the calls to `output_evacuate` and `root_for_each_container`.
- I'm also questioning whether `output->enabled` should be set before the return.

Any thoughts?

emersion: That's a good point. I've decided to just move the assert at the beginning of the function. This makes it so we don't assume anything in case of broken Sway state, and don't change anything. This is probably the safest thing to do, and would probably make it more obvious to the user that something's going wrong.

RedSoxFan:
Thanks!

PR #5544, opened 2020-07-13
xdg_shell: schedule configure on maximize requests

RPigott:
OK, so I'm pretty sure this is a hack, but hear me out.

I saw some users get caught up about buggy Evince behavior and when searching the old reports found #4593. I think the discussion there is correct, however when I tried to run Evince and reproduce, I can see with `WAYLAND_DEBUG=1 evince` that just before evince hangs (until next configure) it tries `set_maximized`, which obviously shouldn't do anything for a tiled view, BUT sway doesn't send back a configure event. Evince waits until the next configure and continues fine.

Now, I'm pretty sure we're _supposed_ to send one, even if it doesn't do anything, based on the description of set_maximized from the xml:

> After requesting that the surface should be maximized, the compositor
    will respond by emitting a configure event. Whether this configure
    actually sets the window maximized is subject to compositor policies.

I make sway respond with this patch. I can now see sway emit the configure, and it does fix the hang in Evince for me. However I don't think I'm supposed to call `wlr_xdg_surface_schedule_configure` here.

Is this the right behavior? Anyone know a better way to do it?

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -360,6 +360,11 @@ static void handle_request_fullscreen(struct wl_listener *listener, void *data) 
 	transaction_commit_dirty();
 }
 
+static void handle_request_maximize(struct wl_listener *listener, void *data) {
+	struct wlr_xdg_surface *surface = data;
+	wlr_xdg_surface_schedule_configure(surface);
+}
+
 static void handle_request_move(struct wl_listener *listener, void *data) {
 	struct sway_xdg_shell_view *xdg_shell_view =
 		wl_container_of(listener, xdg_shell_view, request_move);
@@ -402,6 +407,7 @@ static void handle_unmap(struct wl_listener *listener, void *data) {
 	wl_list_remove(&xdg_shell_view->commit.link);
 	wl_list_remove(&xdg_shell_view->new_popup.link);
 	wl_list_remove(&xdg_shell_view->request_fullscreen.link);
+	wl_list_remove(&xdg_shell_view->request_maximize.link);
 	wl_list_remove(&xdg_shell_view->request_move.link);
 	wl_list_remove(&xdg_shell_view->request_resize.link);
 	wl_list_remove(&xdg_shell_view->set_title.link);
@@ -449,6 +455,10 @@ static void handle_map(struct wl_listener *listener, void *data) {
 	xdg_shell_view->request_fullscreen.notify = handle_request_fullscreen;
 	wl_signal_add(&xdg_surface->toplevel->events.request_fullscreen,
 			&xdg_shell_view->request_fullscreen);
+
+	xdg_shell_view->request_maximize.notify = handle_request_maximize;
+	wl_signal_add(&xdg_surface->toplevel->events.request_maximize,
+			&xdg_shell_view->request_maximize);
 
 	xdg_shell_view->request_move.notify = handle_request_move;
 	wl_signal_add(&xdg_surface->toplevel->events.request_move,


progandy:
The same language is used for UnsetMaximized, SetFullscreen, UnsetFullscreen, so those might have to be fixed as well.

It is probably better to explicitly call ```wlr_xdg_toplevel_set_maximized(surface, false)``` instead of directly scheduling a configure event.

emersion:
>evince hangs (until next configure) it tries set_maximized

Nice find!

>It is probably better to explicitly call wlr_xdg_toplevel_set_maximized(surface, false) instead of directly scheduling a configure event.

Not sure. wlroots will notice the pending state is the same as the current state, and not send any event.

>Is this the right behavior?

Yes, this PR implements the right behaviour.

It'd be a lot better to handle this is wlroots, so that compositors don't have to worry about it. We _could_ call `wlr_xdg_surface_schedule_configure` in functions that are supposed to trigger a configure event, however that breaks one use-case: delaying configuration (e.g. do something asynchronously prior to deciding whether to maximize or not). If we schedule a configure event in wlroots, first wlroots will send an empty configure event, then the compositor will asynchronously send another configure event. This is incorrect.

I wonder if there's a better way to handle this in wlroots. If we don't find a better solution, we'd need to check whether there are other places where we should send empty configure events, and document this in the wlroots header.

progandy:
> I wonder if there's a better way to handle this in wlroots. If we don't find a better solution, we'd need to check whether there are other places where we should send empty configure events, and document this in the wlroots header.

wlroots could at least be allowed to send empty events if the callback is undefined.

emersion:
>wlroots could at least be allowed to send empty events if the callback is undefined.

Do you mean send an empty configure event when there is no listener?

I'm not sure this is a good idea:

- It's unexpected that registering a listener would have a side-effect and change wlroots' behaviour.
- If there is a listener it doesn't necessarily means it's the compositor's listener. It's common for wlroots modules to register listeners without necessarily handling the event like the compositor would.

RPigott:
> It'd be a lot better to handle this is wlroots, so that compositors don't have to worry about it. [...] however that breaks one use-case: delaying configuration.

Yeah! I thought the same.

I feel like set_maximized should have an event in wlroots, like set_fullscreen does, rather than just setting the maximized state right away. `wlr_xdg_toplevel` seems to have a set of states that do and don't need compositor approval to modify, and maximized seems like it should be in the former group.

As is, sway only sees the "new" state, which also makes it hard to comply with the protocol even when it does want to permit the client to change state, because it is required to reply with a configure in case the client makes the request while already maximized/un-maximized as well

> unset_maximize
> [...]
> If the surface was already not maximized, the compositor will still
emit a configure event without the "maximized" state.

An event doesn't solve the problem of wlroots refusing to send a configure, though. Maybe there could be an extra bit of state in wlroots that only the compositor modifies, like a nonce sway can increment that just exists to perturb the state and effectively let wlroots know the compositor is "done" considering the request? So that wlroots state has changed and wlroots knows a configure is required.

emersion:
>I feel like set_maximized should have an event in wlroots, like set_fullscreen does, rather than just setting the maximized state right away. wlr_xdg_toplevel seems to have a set of states that do and don't need compositor approval to modify, and maximized seems like it should be in the former group.

That's already done. wlroots will emit a `request_maximize` event without updating the effective state.

https://github.com/swaywm/wlroots/blob/842df2bd6c940c5f4933622987fc9a296f404d67/types/xdg_shell/wlr_xdg_toplevel.c#L394

>An event doesn't solve the problem of wlroots refusing to send a configure, though. Maybe there could be an extra bit of state in wlroots that only the compositor modifies, like a nonce sway can increment that just exists to perturb the state and effectively let wlroots know the compositor is "done" considering the request? So that wlroots state has changed and wlroots knows a configure is required.

Actually, I don't think we need the compositor to be involved here. When the client sends a `set_maximized` request, wlroots knows the client expects a configure event even if it doesn't actually maximize the toplevel. So we could have a flag indicating that the client expects a configure event in wlroots, and make wlroots send (possibly no-op) configure events in this case.

https://github.com/swaywm/wlroots/blob/842df2bd6c940c5f4933622987fc9a296f404d67/types/xdg_shell/wlr_xdg_toplevel.c#L28

emersion:
>I can now see sway emit the configure, and it does fix the hang in Evince for me.

Actually that's weird, I'd expect `compare_xdg_surface_toplevel_state` to prevent a no-op configure event from being sent…

progandy:
> Actually that's weird, I'd expect `compare_xdg_surface_toplevel_state` to prevent a no-op configure event from being sent…

The exported ```wlr_xdg_surface_schedule_configure``` doesn't seem to call that function. Only the wlroots-internal ```schedule_xdg_surface_configure``` cares about the previous state.

RPigott:
> That's already done. wlroots will emit a request_maximize event without updating the effective state.

Uh, that's right. I think mixed up the server_pending and client_pending states... I only really scanned wlr_xdg_surface.c long enough to find something to fix the bug I saw when I realized arrange_container didn't do the trick.

> Actually that's weird, I'd expect compare_xdg_surface_toplevel_state to prevent a no-op configure event from being sent…

Well now I'm confused too :/. Before you pointed out the above I was under the impression it would work because wlroots would send a configure with the maximized flag set, but I guess that's not the case.

I really do see a new configure in WAYLAND_DEBUG output though, and it that changes the behavior in Evince for me. I never thought to check the actual state that was sent because WAYLAND_DEBUG=1 doesn't print it. I've only tested it in an embedded sway if that matters.

It's way too late here, but I guess I'll take another look tomorrow.

emersion:
>The exported wlr_xdg_surface_schedule_configure doesn't seem to call that function. Only the wlroots-internal schedule_xdg_surface_configure cares about the previous state.

Ah, that's correct. This was done initially because some third-party protocols (like xdg-decoration) are add-ons for xdg-toplevel and need to schedule configure events (for their own state) even if the xdg-toplevel state is unchanged.

emersion:
Thanks!

PR #5545, opened 2020-07-13
Add note on quoting to swaymsg manpage

### LINKED ISSUE #5531: `swaymsg output` doesn't respect quoted display name
after starting sway, the command `swaymsg output "Ancor Communications Inc ASUS VS247 E2LMTF120338" mode 1920x1080 scale 1 position 0,100 enable` returns `Error: Invalid output subcommand: Communications.`.

* Tried with single quotes for the display identifier
* Tried with zsh, bash, and sh shells in case it was a shell expansion thing.

According to the `sway-output` manpage this is the correct way to pass the identifier.

- **Sway Version:** 
  - `sway version 1.4-a8a6ed66 (May 12 2020, branch 'master')`

- **Debug Log:**
probably not that helpful but still: https://gist.github.com/ohthehugemanatee/50c79e5acd22d22fdcf6319cb44199d1
I see lots of spam from Variety wallpaper manager, but nothing related to the issue

ohthehugemanatee:
Argument expansion happens twice when sending commands through `swaymsg`, leading to confusing behavior like #5531 . This PR adds a note to the swaymsg manpage to clarify.

I'm not sure that it's sufficient... may be better to add one on sway-output as well, since that's a common place to see multi-word strings as a single argument (display identifiers).

Closes #5531 .

### DIFF
--- a/swaymsg/swaymsg.1.scd
+++ b/swaymsg/swaymsg.1.scd
@@ -44,12 +44,18 @@ _swaymsg_ [options...] [message]
 	The message is a sway command (the same commands you can bind to keybindings
 	in your sway config file). It will be executed immediately.
 
-	See **sway**(5) for a list of commands.
+	See *sway*(5) for a list of commands.
 
-	Tip: If you are proving a command that contains a leading hyphen (_-_),
-	insert two hyphens (_--_) before the command to signal to swaymsg not to
-	parse anything beyond that point as an option. For example, use
-	_swaymsg -- mark --add test_ instead of _swaymsg mark --add test_
+	Tips:
+	- Command expansion is performed twice: once by swaymsg, and again by sway.
+	  If you have quoted multi-word strings in your command, enclose the entire
+	  command in single-quotes. For example, use
+	  _swaymsg 'output "Foobar Display" enable'_ instead of
+	  _swaymsg output "Foobar Display" enable_.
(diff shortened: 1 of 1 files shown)


swaymsg/swaymsg.1.scd
```diff
@@ -47,14 +47,14 @@ _swaymsg_ [options...] [message]
...
 	Tips:
-	- Command expansion is performed twice: once by swaymsg, and again by sway.
-	If you have quoted multi-word strings in your command, enclose the entire command
-	in single-quotes. For example, use _swaymsg 'output "Foobar Display" enable'_
-	instead of _swaymsg output "Foobar Display" enable_
-	- If you are proving a command that contains a leading hyphen (_-_),
-	insert two hyphens (_--_) before the command to signal to swaymsg not to
-	parse anything beyond that point as an option. For example, use
-	_swaymsg -- mark --add test_ instead of _swaymsg mark --add test_
+		- Command expansion is performed twice: once by swaymsg, and again by sway.
```
emersion, line 50: The hyphen shouldn't be indented. You can test your changes with:

```sh
scdoc <swaymsg/swaymsg.1.scd | man -l -
```

ohthehugemanatee: whoops, thanks. Now it renders better:

image

PR #5546, opened 2020-07-14
added ppt unit to move position command

Schnilz:
You could specify the width and height of a container in percent when resizing it but not when moving it.
I also updated the docs to reflect the changes. 
First time coding in C, hopefully my formatting is ok :)

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -71,6 +71,40 @@ float parse_float(const char *value) {
 	return flt;
 }
 
+enum movement_unit parse_movement_unit(const char *unit) {
+	if (strcasecmp(unit, "px") == 0) {
+		return MOVEMENT_UNIT_PX;
+	}
+	if (strcasecmp(unit, "ppt") == 0) {
+		return MOVEMENT_UNIT_PPT;
+	}
+	if (strcasecmp(unit, "default") == 0) {
+		return MOVEMENT_UNIT_DEFAULT;
+	}
+	return MOVEMENT_UNIT_INVALID;
+}
+
+int parse_movement_amount(int argc, char **argv,
+		struct movement_amount *amount) {
+	char *err;
+	amount->amount = (int)strtol(argv[0], &err, 10);
+	if (*err) {
+		// e.g. 10px
+		amount->unit = parse_movement_unit(err);
+		return 1;
+	}
+	if (argc == 1) {
+		amount->unit = MOVEMENT_UNIT_DEFAULT;
+		return 1;
+	}
+	// Try the second argument
+	amount->unit = parse_movement_unit(argv[1]);
+	if (amount->unit == MOVEMENT_UNIT_INVALID) {
+		amount->unit = MOVEMENT_UNIT_DEFAULT;
+		return 1;
+	}
+	return 2;
+}
 
 const char *sway_wl_output_subpixel_to_string(enum wl_output_subpixel subpixel) {
 	switch (subpixel) {
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -820,37 +820,81 @@ static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, expected_position_syntax);
 	}
 
-	double lx, ly;
-	char *inv;
-	lx = (double)strtol(argv[0], &inv, 10);
-	if (*inv != '\0' && strcasecmp(inv, "px") != 0) {
-		return cmd_results_new(CMD_FAILURE, "Invalid position specified");
-	}
-	if (strcmp(argv[1], "px") == 0) {
-		--argc;
-		++argv;
+	struct movement_amount lx = { .amount = 0, .unit = MOVEMENT_UNIT_INVALID };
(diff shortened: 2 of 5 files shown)


emersion:
Thanks for sending a PR! Is this an i3 feature we're missing?

Schnilz:
From quickly looking at i3's docs, i don't think so. Is this a problem ?

emersion:
Unfortunately we don't accept new window management features unless i3 adds them.

Schnilz:
Hm, it doesn't break compatability with i3 Syntax, I don't think. But if this realy is a strictly enforced rule, then I will look into implementing it for i3 too... 
it's already an approved feature request from 2018
(https://github.com/i3/i3/issues/3519).

emersion:
Thanks for finding the i3 issue.

>it's already an approved feature request from 2018

Ah, the feature request has been accepted. Since this is a small feature adding it to Sway sounds fine to me.

sway/commands/move.c
```diff
@@ -822,23 +823,49 @@ static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
 
 	double lx, ly;
 	char *inv;
+	bool ppt_x = false;
+	bool ppt_y = false;
 	lx = (double)strtol(argv[0], &inv, 10);
-	if (*inv != '\0' && strcasecmp(inv, "px") != 0) {
-		return cmd_results_new(CMD_FAILURE, "Invalid position specified");
+	if (*inv != '\0') {
+			if (strcasecmp(inv, "px") != 0) {
```
emersion, line 830: How about adding a small helper to parse the unit?

```c
// Parses a unit string and populates ppt, returns false on failure
static bool parse_unit(const char *str, bool *ppt);
```

sway/commands/move.c
```diff
@@ -847,8 +874,19 @@ static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
 			struct sway_seat *seat = config->handler_context.seat;
 			ws = seat_get_focused_workspace(seat);
 		}
-		lx += ws->x;
-		ly += ws->y;
+		if (!ppt_x) {
+			lx += ws->x;
+		} else {
+			lx = ws->x + lx / 100.0 * ws->width;
+			}
```
emersion, line 881: Style: extra indent here

sway/commands/move.c
```diff
@@ -847,8 +874,19 @@ static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
...
+			if (!ppt_y) {
+			ly += ws->y;
+		} else {
+			ly = ws->y + ly / 100.0 * ws->height;
+			}
```
emersion, line 886: Ditto

sway/commands/move.c
Schnilz: still learning that whole vim thing ;)

Schnilz:
thanks for the quick review. I added the suggested helper function and cleaned up the formatting.

common/util.c
```diff
@@ -71,6 +71,16 @@ float parse_float(const char *value) {
 	return flt;
 }
 
+bool parse_unit(const char *str, bool *ppt) {
+	if (strcmp(str, "px") == 0) {
+		if (ppt) *ppt = false;
```
emersion, line 76: Style: braces are mandatory

But I don't think we need this check, we never call the function with a NULL `ppt`.

common/util.c
```diff
@@ -71,6 +71,16 @@ float parse_float(const char *value) {
 	return flt;
 }
 
+bool parse_unit(const char *str, bool *ppt) {
```
emersion, line 74: I don't think this needs to live in common code, since this is just used in `move.c`. A `static` function in the same file would be enough.

sway/commands/move.c
```diff
@@ -822,11 +823,15 @@ static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
 
 	double lx, ly;
 	char *inv;
+	bool ppt_x = false;
+	bool ppt_y = false;
 	lx = (double)strtol(argv[0], &inv, 10);
-	if (*inv != '\0' && strcasecmp(inv, "px") != 0) {
-		return cmd_results_new(CMD_FAILURE, "Invalid position specified");
+	if ((*inv != '\0') && !parse_unit(inv,&ppt_x)) {
```
emersion, line 829: Style: unnecessary parentheses, missing space after comma

sway/commands/move.c
```diff
@@ -822,11 +823,15 @@ static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
...
+		return cmd_results_new(CMD_FAILURE,
+			"Invalid position specified");
 	}
-	if (strcmp(argv[1], "px") == 0) {
+
+	if (parse_unit(argv[1],&ppt_x)) {
```
emersion, line 834: Style: missing space after comma (also applies to the rest of the patch)

Schnilz:
I found that in the resize command the functions are already implemented, so i moved those to common/utils to the other `parse_*` functions (renamed to a more general name) and used them in the  move command too. I hope this is ok.

PR #5549, opened 2020-07-15
Add check for empty GEOM variable

vkasljevic:
In case when slurp is used to select part of screen, if user aborts the selection, grimshot will capture the whole screen instead of exiting. This is fixed with check for empty variable.

### DIFF
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -111,6 +111,10 @@ if [ "$ACTION" = "check" ] ; then
   exit
 elif [ "$SUBJECT" = "area" ] ; then
   GEOM=$(slurp -d)
+  # Check if user exited slurp without selecting the area
+  if [ -z "$GEOM" ]; then
+    exit
+  fi
   WHAT="Area"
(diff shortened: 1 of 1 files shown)


vkasljevic:
There was same issue with windows. I've added a check there as well.

emersion:
>There was same issue with windows. I've added a check there as well.

Forgot to push changes?

vkasljevic:
I did but for some reason first comit is detached from branch and second is not visible. I'm working on it.

emersion:
Actually, it seems like GitHub is having issues.

Please squash both commits together.

vkasljevic:
And push new commit on top of these? I could overwrite history, but never ends well.

emersion:
Yes, please force-push with a single updated commit, so that history is clean when merged into master.

vkasljevic:
It didn't work. I'll just create a new pull request from new branch. There is not point spending so much time debugging over a small patch.

Issue #4324, opened 2019-07-12
Black screen on some Wine games after they lose focus


junglerobba:
I observed this on Dark Souls 1, Zero Escape VLR and Earth Defense Force 5. After losing focus (moving mouse out of game window, switching workspace) the game window turns black without any way of getting an image back besides restarting the game. EDF still shows a cursor though. 
I tested Dark Souls 1 on GNOME Wayland and could not reproduce this behavior.

sway version 1.1-rc1-61-gc312a10c (Jul 11 2019, branch 'master')

[Debug Log with default config](https://gist.github.com/junglerobba/2c077d5cbd09e4b221c93c3997b3ac33)

Steps to reproduce:

1. Start any of the listed games
2. Switch to a different workspace, the game turns black at this point

emersion:
I notice this as well with other games.

cafehaine:
Might be unrelated, but something similar happens with Dota 2 (sdl, xwayland). It doesn't happen when the game is in "borderless windowed" mode though

junglerobba:
FWIW, this also happens with Wayfire.

p4block:
Also Car Mechanic Simulator 2018. Sometimes it doesn't happen the first time, but will happen the second time.

Lost a bit of progress in the game due to this >:(

ammgws:
> It doesn't happen when the game is in "borderless windowed" mode though

Cheers, experienced this with The Walking Dead game and turning off fullscreen in the settings did the trick.

mikesbytes:
I have also had this same bug with Doom Eternal (which uses vulkan, if that makes a difference)

p4block:
Note that this problem **can be worked around** by running the uniconize daemon or a program/script with similar functionality:

https://github.com/neXromancers/uniconize

antoyo:
I'm having this issue with sway 1.5.1 on Arch.
The PR #5566 should be included in this version, no?

kennylevinsen:
No, 1.5 was tagged 2020-07-16, that PR was opened 2020-07-21. 1.5.1 is just a wlroots compatibility bump.

PR #5566, opened 2020-07-21
Fix X11 clients getting stuck minimized

junglerobba:
Usually it should be enough to simply not grant a client's
minimize request, however some applications (Steam, fullscreen
games in Wine) don't wait for the compositor and minimize anyway,
getting them stuck in an unrecoverable state.
Restoring them immediately lead to heavy flickering ( https://streamable.com/imhyj7 ) when unfocused
on my test application (Earth Defense Force 5 via Steam), so it's
preferable to grant their request without actually minimizing and
then restoring them once they are in focus again.

Should fix https://github.com/swaywm/sway/issues/4324

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -222,6 +222,11 @@ static void set_activated(struct sway_view *view, bool activated) {
 		return;
 	}
 	struct wlr_xwayland_surface *surface = view->wlr_xwayland_surface;
+
+	if (activated && surface->minimized) {
+		wlr_xwayland_surface_set_minimized(surface, false);
+	}
+
 	wlr_xwayland_surface_activate(surface, activated);
 }
 
@@ -406,6 +411,7 @@ static void handle_destroy(struct wl_listener *listener, void *data) {
 	wl_list_remove(&xwayland_view->destroy.link);
 	wl_list_remove(&xwayland_view->request_configure.link);
 	wl_list_remove(&xwayland_view->request_fullscreen.link);
+	wl_list_remove(&xwayland_view->request_minimize.link);
 	wl_list_remove(&xwayland_view->request_move.link);
 	wl_list_remove(&xwayland_view->request_resize.link);
 	wl_list_remove(&xwayland_view->request_activate.link);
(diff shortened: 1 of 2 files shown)


sway/input/seat.c
```diff
@@ -1197,6 +1197,15 @@ void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
 		// the workspace needs to be arranged
 		arrange_workspace(new_workspace);
 	}
+
+	#if HAVE_XWAYLAND
```
RedSoxFan, line 1201: Do not indent preprocessor directives.

sway/input/seat.c
```diff
@@ -1197,6 +1197,15 @@ void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
...
+	// Some X11 clients minimize without waiting for the compositor to allow it
+	if (container && container->view && container->view->type == SWAY_VIEW_XWAYLAND
```
RedSoxFan, line 1203: I'm not a huge fan of xwayland specific code bleeding into this file. The other two cases (which are pre-existing and unrelated to this PR) are for the entire server so those are a little harder to mask away.

However, this is directly related to a view so I'd prefer to see this get moved to `sway/desktop/xwayland.c`. I think it makes sense to just add the following to `set_activated`.
```c
if (activated && surface->minimized) {
  wlr_xwayland_surface_set_minimized(surface, false);
}
```

Otherwise `set_minimized` should be added to `sway_view_impl`, but I'd rather not do that unless the previous solution has some oversight.

sway/desktop/xwayland.c
```diff
@@ -653,6 +673,10 @@ void handle_xwayland_surface(struct wl_listener *listener, void *data) {
 		&xwayland_view->request_fullscreen);
 	xwayland_view->request_fullscreen.notify = handle_request_fullscreen;
 
+	wl_signal_add(&xsurface->events.request_minimize,
```
RedSoxFan, line 676: This needs to be removed in `handle_destroy`

sway/desktop/xwayland.c
```diff
@@ -508,6 +508,26 @@ static void handle_request_fullscreen(struct wl_listener *listener, void *data)
...
+	if (!xsurface->mapped) {
+		return;
+	}
+
+	struct wlr_xwayland_minimize_event *e = data;
+	struct sway_seat *seat = input_manager_current_seat();
+	bool focused =
+			seat->wlr_seat->keyboard_state.focused_surface == xsurface->surface;
+	if (!focused && e->minimize) {
+		wlr_xwayland_surface_set_minimized(xsurface, true);
```
RedSoxFan, line 525: I think the following would be cleaner:
```c
bool focused = seat_get_focus(seat) == &view->node;
wlr_xwayland_surface_set_minimized(xsurface, !focused && e->minimize);
```

sway/input/seat.c
```diff
@@ -1197,6 +1197,15 @@ void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
...
+			&& container->view->wlr_xwayland_surface->minimized) {
+		wlr_xwayland_surface_set_minimized(
+			container->view->wlr_xwayland_surface, false);
+	}
+	#endif
```
RedSoxFan, line 1208: Do not indent preprocessor directives.

sway/input/seat.c
junglerobba: I didn't really like doing this here either, but I guess I just missed that it could be done in `set_activated`, so thanks for pointing that out.

RedSoxFan:
Thanks!

Issue #5569, opened 2020-07-23
include is a valid runtime command.


RPigott:
Filing this as a bug since I think it is unintentional, but `include` is a valid runtime command, even though sway(5) files it under the "only in configuration file" commands.

What I mean is, you can include additional configuration commands at runtime like so: `swaymsg include ~/.config/sway/additional` and sway will run them, so long as the file contains only runtime commands (or other includes).

It's not such a bad _feature_ actually, though it has some strange behavior. First, it works only once since a single file can be included only once. Second, relative paths are always interpreted relative to the main sway config, not the PWD of swaymsg. Both sensible behaviors for a config command, but highly unusual of a runtime command.

It also gives a silly new way to write sway "scripts":

```
$ cat ~/.config/sway/lol # must use an absolute path
#!/usr/bin/swaymsg include
rename workspace to shebang
$ chmod +x ~/.config/sway/lol
$ ~/.config/sway/lol # it works!
```

Having `include` work but only once and only with absolute paths at runtime is pretty strange, so I think it should probably be disallowed at runtime to match the current documentation.

If we want to change the docs and keep runtime includes, I think the behavior should be corrected to be more sensible when invoked with swaymsg.

PR #5570, opened 2020-07-23
commands: disallow runtime include

RPigott:
Simple fix for #5569 if this is what we want.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -72,7 +72,6 @@ static struct cmd_handler handlers[] = {
 	{ "fullscreen", cmd_fullscreen },
 	{ "gaps", cmd_gaps },
 	{ "hide_edge_borders", cmd_hide_edge_borders },
-	{ "include", cmd_include },
 	{ "input", cmd_input },
 	{ "mode", cmd_mode },
 	{ "mouse_warping", cmd_mouse_warping },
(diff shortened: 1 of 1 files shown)


synaptiko:
I was relying on this behavior to have a switcher of color schemes during the runtime, I overcome the limitation of loading a file only once by generating unique file names to be included: https://github.com/synaptiko/.files/blob/master/0-theme/switch-sway.js#L45

I would welcome this back but otherwise I need to find a different method how to achieve the same.

Geo25rey:
@synaptiko I believe the following can work for you.

Add the following lines to your config:
```
set $chosen-theme-link ~/.sway/theme
set $theme-path ~/.sway/theme-to-include
set $theme-keybind $mod+i

include $chosen-theme-link

bindsym $theme-keybind exec 'rm -f $chosen-theme-link; ln -s $theme-path $chosen-theme-link; swaymsg reload'
```

`$chosen-theme-link` is a symbolic link to your current theme config file. `$theme-path` is the actual theme config file. `$theme-keybind` is a keybind to switch to the the theme.

All you need to do is change the variables to the paths and keybinds you want.

synaptiko:
@Geo25rey Thank you, I had this initially but I didn't like the need to call `reload` (it was messing up with some things previously). I already implemented a solution without `include`, I just "parse" the config and run the commands to change colors one by one (which is good enough for my needs): https://github.com/synaptiko/.files/blob/7e571a48f303dc608e2aca0fdcce8333d1f6cd76/0-theme/switch-sway.js#L43-L65

Geo25rey:
@synaptiko it seems between the JavaScript code and the source code of swaymsg there is redundant argument parsing. Maybe you can base your code off of the swaymsg source code for increased efficiency.

PR #5477, opened 2020-06-22
Rebase of wlr-foreign-toplevel-management-v1

johan-bjareholt:
A rebase of #4476 as requested by @ddevault 

There were no risky merge conflicts. Compiled and shortly tested as well.

PR #5589, opened 2020-07-31
Implement foreign toplevel fullscreen

RPigott:
This pull adds support for the foreign toplevel fullscreen event and request.

By event I mean that previously FTM would not indicate the fullscreen state as set on fullscreen views, so the first commit fixes it to indicate the fullscreen state for fullscreen views.

The second commit handles the FTM set_fullscreen request. It doesn't comply with the output hint, since it's not necessary and I don't have an easy way to test multi-output stuff on my laptop anway. PTAL.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -978,6 +978,10 @@ static void set_fullscreen_iterator(struct sway_container *con, void *data) {
 	if (con->view->impl->set_fullscreen) {
 		bool *enable = data;
 		con->view->impl->set_fullscreen(con->view, *enable);
+		if (con->view->foreign_toplevel) {
+			wlr_foreign_toplevel_handle_v1_set_fullscreen(
+				con->view->foreign_toplevel, *enable);
+		}
 	}
 }
 
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -670,6 +670,17 @@ static void handle_foreign_close_request(
 	view_close(view);
 }
 
+static void handle_foreign_destroy(
+		struct wl_listener *listener, void *data) {
+	struct sway_view *view = wl_container_of(
+			listener, view, foreign_destroy);
+
+	wl_list_remove(&view->foreign_activate_request.link);
+	wl_list_remove(&view->foreign_fullscreen_request.link);
+	wl_list_remove(&view->foreign_close_request.link);
+	wl_list_remove(&view->foreign_destroy.link);
+}
+
 void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 			  bool fullscreen, struct wlr_output *fullscreen_output,
 			  bool decoration) {
@@ -709,6 +720,9 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 	view->foreign_close_request.notify = handle_foreign_close_request;
 	wl_signal_add(&view->foreign_toplevel->events.request_close,
 			&view->foreign_close_request);
+	view->foreign_destroy.notify = handle_foreign_destroy;
+	wl_signal_add(&view->foreign_toplevel->events.destroy,
+			&view->foreign_destroy);
 
 	// If we're about to launch the view into the floating container, then
 	// launch it as a tiled view in the root of the workspace instead.
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -636,6 +636,33 @@ static void handle_foreign_activate_request(
 	}
 }
 
+static void handle_foreign_fullscreen_request(
+		struct wl_listener *listener, void *data) {
+	struct sway_view *view = wl_container_of(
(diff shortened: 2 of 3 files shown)


sway/tree/view.c
```diff
@@ -676,6 +703,9 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 	view->foreign_activate_request.notify = handle_foreign_activate_request;
 	wl_signal_add(&view->foreign_toplevel->events.request_activate,
 			&view->foreign_activate_request);
+	view->foreign_fullscreen_request.notify = handle_foreign_fullscreen_request;
+	wl_signal_add(&view->foreign_toplevel->events.request_fullscreen,
```
emersion, line 707: Need to remove the listener somewhere

emersion:
>It doesn't comply with the output hint

I'm fine with keeping this as a follow-up issue. But if you want, you could just copy-pasta the xdg-shell code:

https://github.com/swaywm/sway/blob/5c67d997948309d881ed94387309865c76ecddcb/sway/desktop/xdg_shell.c#L354

sway/tree/view.c
RPigott: When the signal is destroyed along with the toplevel handle it's just as good, right? Because the view owns the FT handle _and_ all the listeners in this case. It would only be a problem if a signal could be emitted on a FT handle that outlived it's view, I think, but the handle is destroyed first.

Or should I remove _all_ the foreign toplevel listeners before it's destroyed here perhaps?

https://github.com/swaywm/sway/blob/5c67d997948309d881ed94387309865c76ecddcb/sway/tree/view.c#L767-L770

Because now that I check, the activate and close listeners don't appear to be removed either, if it matters. Did you make the same observation in [this comment](https://github.com/swaywm/sway/pull/5477#pullrequestreview-436128287)?

RPigott:
> you could just copy-pasta the xdg-shell code and test with WLR_WL_OUTPUTS=2

Oh I didn't know about WLR_WL_OUTPUTS, that's pretty neat.

That's probably what I'll do, and it looks like it'll work fine,  however I've got to implement it in my little test client as well, so if it's alright with you I'll fix the output hints in a follow-up issue.

sway/tree/view.c
emersion: >When the signal is destroyed along with the toplevel handle it's just as good, right? Because the view owns the FT handle and all the listeners in this case. It would only be a problem if a signal could be emitted on a FT handle that outlived it's view, I think, but the handle is destroyed first.

No, because if another listener removes itself from the list, it'll need to write to other listeners' prev/next pointers. If the other listeners are free'd, then this causes a use-after-free.

This is also a requirement of [`wlr_signal_emit_safe`](https://github.com/swaywm/wlroots/blob/aaf490d794bd077330c306face305f4a87de5809/util/signal.c#L7).

RPigott: I see. So all the foreign toplevel listeners should be removed then.

RPigott:
OK, I tried two different strategies to remove the listeners. I remove the active and close listeners as well since they were not previously removed.

It seems the destroy signal is only emitted when the compositor calls destroy, and this is the only place it's called, so they should be functionally equivalent, and they don't cause issues for me. However I haven't been able to create a crash without removing them either so I'm not totally certain this is the best way.

sway/tree/container.c
```diff
@@ -978,6 +978,11 @@ static void set_fullscreen_iterator(struct sway_container *con, void *data) {
 	if (con->view->impl->set_fullscreen) {
 		bool *enable = data;
 		con->view->impl->set_fullscreen(con->view, *enable);
+		if (con->view->foreign_toplevel) {
+			wlr_foreign_toplevel_handle_v1_set_fullscreen(
+				con->view->foreign_toplevel, *enable
+			);
```
emersion, line 984: Style: no newline before `);`

sway/tree/view.c
```diff
@@ -636,13 +636,51 @@ static void handle_foreign_activate_request(
...
 
+static void handle_foreign_destroy(
+		struct wl_listener *listener, void *data) {
+	struct sway_view *view = wl_container_of(
+			listener, view, foreign_destroy);
+
+	wl_list_remove(&view->foreign_activate_request.link);
+	wl_list_remove(&view->foreign_fullscreen_request.link);
+	wl_list_remove(&view->foreign_close_request.link);
+	wl_list_remove(&listener->link);
```
emersion, line 681: Nit: can we remove `&view->foreign_destroy.link` instead? Makes it easier to compare it with the function registering the listeners and check we aren't missing anything.

emersion:
Looks good apart from these minor issues!

RPigott:
Okay, should be good now. I rebased to drop the intermediate commit that removed listeners without the destroy signal, and to address the style nits in their original commits.

PR #5599, opened 2020-08-02
document force_display_urgency_hint

### LINKED ISSUE #5573: force_display_urgency_hint is undocumented
The `force_display_urgency_hint` command is not documented in [sway(5)](https://github.com/swaywm/sway/blob/master/sway/sway.5.scd).

JayceFayne:
fixes #5573

### DIFF
--- a/sway/commands/force_display_urgency_hint.c
+++ b/sway/commands/force_display_urgency_hint.c
@@ -1,5 +1,6 @@
 #include "sway/commands.h"
 #include "sway/config.h"
+#include <errno.h>
 
 struct cmd_results *cmd_force_display_urgency_hint(int argc, char **argv) {
 	struct cmd_results *error = NULL;
@@ -8,13 +9,16 @@ struct cmd_results *cmd_force_display_urgency_hint(int argc, char **argv) {
 		return error;
 	}
 
-	char *err;
-	int timeout = (int)strtol(argv[0], &err, 10);
-	if (*err) {
-		if (strcmp(err, "ms") != 0) {
-			return cmd_results_new(CMD_INVALID,
-					"Expected 'force_display_urgency_hint <timeout> ms'");
-		}
+	errno = 0;
+	char *end;
+	int timeout = (int)strtol(argv[0], &end, 10);
+	if (errno || end == argv[0] || (*end && strcmp(end, "ms") != 0)) {
+		return cmd_results_new(CMD_INVALID, "timeout integer invalid");
+	}
+
+	if (argc > 1 && strcmp(argv[1], "ms") != 0) {
+		return cmd_results_new(CMD_INVALID,
+				"Expected 'force_display_urgency_hint <timeout> [ms]'");
 	}
 
 	config->urgent_timeout = timeout > 0 ? timeout : 0;
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -632,6 +632,13 @@ The default colors are:
 	information on pango font descriptions, see
 	https://developer.gnome.org/pango/stable/pango-Fonts.html#pango-font-description-from-string
 
+*force_display_urgency_hint* <timeout> [ms]
+	If an application on another workspace sets an urgency hint, switching to this
+	workspace may lead to immediate focus of the application, which also means the
+	window decoration color would be immediately resetted to *client.focused*. This
+	may make it unnecessarily hard to tell which window originally raised the
+	event. This option allows to set a _timeout_ in ms to delay the urgency hint reset.
+
 *titlebar_border_thickness* <thickness>
 	Thickness of the titlebar border in pixels
 


sway/commands/force_display_urgency_hint.c
```diff
@@ -13,7 +13,7 @@ struct cmd_results *cmd_force_display_urgency_hint(int argc, char **argv) {
 	if (*err) {
 		if (strcmp(err, "ms") != 0) {
 			return cmd_results_new(CMD_INVALID,
-					"Expected 'force_display_urgency_hint <timeout> ms'");
```
emersion, line 16: Hmm, why replace the space with a `|`?

emersion:
LGTM otherwise!

sway/commands/force_display_urgency_hint.c
JayceFayne: I thought it should be consistent with the manual, I documented the command as `force_display_urgency_hint <timeout>|ms` and having the command return a different desired syntax seemed confusing to me. The i3 [userguide](https://github.com/i3/i3/blob/next/docs/userguide#L1176) documents `force_display_urgency_hint` as `force_display_urgency_hint <timeout> ms` since `force_display_urgency_hint ms` is not valid syntax (but is in sway, maybe the implementation should be changed?). I also just noticed that the `force_display_urgency_hint` command is config only on i3 while it can be used over ipc with sway.

emersion: >maybe the implementation should be changed?

Hmm, yeah, Sway's implementation is buggy. `force_display_urgency_hint ms` doesn't make sense.

JayceFayne: Atm it just sets `force_display_urgency_hint` to zero. I could create a follow up pull request that changes the syntax to `force_display_urgency_hint <timeout> ms`. Or maybe `force_display_urgency_hint <timeout> [ms]`? I am not sure.

emersion: > `force_display_urgency_hint <timeout> [ms]`

Yeah this sounds like the correct one. See the i3 tests: https://github.com/i3/i3/blob/ae757c6848b49d7a0423973a39791ba6eca29308/testcases/t/201-config-parser.t#L400

JayceFayne: Ahh yes, I see. Should I create a separate pull request and add the documentation in this one?

emersion: If you're willing to change the code to fix it, feel free to edit this PR.

JayceFayne: Alright!

sway/commands/force_display_urgency_hint.c
```diff
@@ -11,10 +11,12 @@ struct cmd_results *cmd_force_display_urgency_hint(int argc, char **argv) {
 	char *err;
```
emersion, line 11: In fact, I'd prefer to name this variable `end` rather than `err`. `strtol` will set it to the first non-digit character.

sway/commands/force_display_urgency_hint.c
```diff
@@ -11,10 +11,12 @@ struct cmd_results *cmd_force_display_urgency_hint(int argc, char **argv) {
...
 	int timeout = (int)strtol(argv[0], &err, 10);
 	if (*err) {
```
emersion, line 13: There are a number of errors not handled here:

* We should error out if `end == argv[0]` (the first character isn't a digit, nothing has been parsed)
* We should set `errno = 0` before the `strtol` call, and error out if `errno != 0` after the call.

emersion:
Thanks for splitting this up in two commits :+1:

sway/commands/force_display_urgency_hint.c
JayceFayne: yes, I agree!

sway/commands/force_display_urgency_hint.c
JayceFayne: ahh...right :smile:

JayceFayne:
I had some trouble rebasing the commits :sweat_smile:

JayceFayne:
I pushed some code. Let me know if that covers the errors cases you mentioned :+1:

emersion:
The new error handling LGTM, except that it doesn't accept `force_display_urgency_hint 500ms` (without a space).

JayceFayne:
Yes I noticed that limitation as well while looking trough the i3 tests. But since its not really documented behaviour I'd say we could keeping it simple for now and maybe support that syntax later if needed?

emersion:
The issue is that it was supported before this PR. We can't break existing user configurations.

JayceFayne:
Ofc, you are right, I'll try to fix it.

sway/commands/force_display_urgency_hint.c
```diff
@@ -8,13 +9,16 @@ struct cmd_results *cmd_force_display_urgency_hint(int argc, char **argv) {
...
-	int timeout = (int)strtol(argv[0], &err, 10);
-	if (*err) {
-		if (strcmp(err, "ms") != 0) {
-			return cmd_results_new(CMD_INVALID,
-					"Expected 'force_display_urgency_hint <timeout> ms'");
-		}
+	errno = 0;
+	char *end;
+	int timeout = (int)strtol(argv[0], &end, 10);
+	if (errno || end == argv[0] || (*end && strcmp(end, "ms") != 0) {
```
emersion, line 15: Missing a parens, otherwise looks good!

JayceFayne: Yeah, just noticed while looking at the ci output :smile:

PR #5602, opened 2020-08-03
Document required '\n' in swaybar-protocol

hhirtz:
The following statusbar output is not considered by sway to be following
the swaybar-protocol:

    {"version":1}[[{"full_text":"2.89","urgent":false}],

However this one is:

    {"version":1}\n[[{"full_text":"2.89","urgent":false}],

Both outputs contain a header with the required values and an unfinished
array of objects with the required values, but the first one is showed
verbatim.

### DIFF
--- a/swaybar/swaybar-protocol.7.scd
+++ b/swaybar/swaybar-protocol.7.scd
@@ -8,10 +8,10 @@ swaybar-protocol - JSON status line protocol for swaybar
 
 swaybar defines the following JSON protocol to specify what information is
 displayed in the status line on the right side of swaybar. The protocol
-comprises a header, that is a JSON object, followed by an infinite JSON
-array that represents the information to display. All communication is done by
-writing the status line to standard output and reading events from standard
-input.
+comprises a header, that is a JSON object, followed by a newline (*0x0A*),
+followed by an infinite JSON array that represents the information to display.
+All communication is done by writing the status line to standard output and
+reading events from standard input.
 
 # HEADER
 


emersion:
It doesn't seem like i3 has this limitation. Maybe it is a Sway bug?

alebastr:
The '\n' after the header is already a part of [i3bar protocol](https://i3wm.org/docs/i3bar-protocol.html).
While i3 itself may implement a correct streaming JSON parser and handle such cases, it's not a requirement for any compatible status bars.

swaybar/swaybar-protocol.7.scd
```diff
@@ -8,10 +8,10 @@ swaybar-protocol - JSON status line protocol for swaybar
 
 swaybar defines the following JSON protocol to specify what information is
 displayed in the status line on the right side of swaybar. The protocol
-comprises a header, that is a JSON object, followed by an infinite JSON
-array that represents the information to display. All communication is done by
-writing the status line to standard output and reading events from standard
-input.
+comprises a header, that is a JSON object, followed by a newline (*0x0A*),
```
Geo25rey, line 11: You might want to replace 'a newline' with 'an ASCII newline character'.

Geo25rey:
Something to consider regarding this newline bit. What if '\n' was '\r\n' from a Windows PC? Should that be considered an error during JSON parsing? Should it be ignored? What actually happens in a situation like this?

emersion:
>The '\n' after the header is already a part of i3bar protocol.

Oh, missed that, my bad.

swaybar/swaybar-protocol.7.scd
emersion: The current wording sounds explicit enough for me, especially since 0x0A is mentioned.

PR #5603, opened 2020-08-03
rm 3 useless if statements and merge them into one

DRG-developer:
very small optimization, but it improves the code a little.

makes 3 if statements one

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -244,19 +244,22 @@ static uint32_t render_status_block(cairo_t *cairo,
 
 	uint32_t border_color = block->urgent
 		? config->colors.urgent_workspace.border : block->border;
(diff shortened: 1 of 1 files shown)


emersion:
Thanks for taking the time to send a PR, but I don't think this improves readability.

DRG-developer:
@emersion 
It might not improve code readability, but it will give a very slight performance enhancement, and for sure does *not* decrease readability.

Anyway, have a good day

Emantor:
> It might not improve code readability, but it will give a very slight performance enhancement, and for sure does _not_ decrease readability.

I don't think so, modern compilers will optimize this correctly regardless of the notation. If you have concrete examples in assembly where this is not the case, please post those, I'd be interested.

PR #5607, opened 2020-08-05
set XDG_CURRENT_DESKTOP environment for systemd login

denisok:
https://github.com/emersion/xdg-desktop-portal-wlr requires XDG_CURRENT_DESKTOP to be set, login managers should set it from DesktopNames

### DIFF
--- a/sway.desktop
+++ b/sway.desktop
@@ -3,3 +3,4 @@ Name=Sway
 Comment=An i3-compatible Wayland compositor
 Exec=sway
 Type=Application
+DesktopNames=sway


emersion:

denisok:
@emersion is this documentation issue for the xdg-desktop-portal-wlr ?
`Make sure XDG_CURRENT_DESKTOP=sway is set.`
Why it is there ?

Actually I setup clean environment and that var doesn't affect screen sharing in the Firefox. The only thing is needed is correct XDG_SESSION_TYPE set to `wayland`.

emersion:
Likely happens because you don't have any other portal implementations installed.

denisok:
> Likely happens because you don't have any other portal implementations installed.

OK, so does it make sense to set it? I didn't get it from the discussion from dup issues.

BTW is there an issue to set `XDG_SESSION_TYPE` ?
https://github.com/systemd/systemd/pull/14925#issuecomment-621687957

denisok:
> BTW is there an issue to set `XDG_SESSION_TYPE` ?
> [systemd/systemd#14925 (comment)](https://github.com/systemd/systemd/pull/14925#issuecomment-621687957)

answering my question:
https://github.com/swaywm/wlroots/pull/2304

emersion:
>so does it make sense to set it? 

If you're using xdg-desktop-portal-wlr yeah

PR #5609, opened 2020-08-06
Add trigger information to criteria

danieldg:
This allows writing commands like "for_window [trigger="map" workspace="1"]"
that only execute if the criteria matches during the initial mapping of the
window.

Currently implemented trigger names:
 - command: explicit invocation by a command, not very useful
 - map: initial window creation
 - mark, title, app_id, class, window_role, window_type:
   the given property changed

### DIFF
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -45,6 +45,7 @@ struct criteria {
 #endif
 	bool floating;
 	bool tiling;
+	bool always;
 	char urgent; // 'l' for latest or 'o' for oldest
 	struct pattern *workspace;
 	pid_t pid;
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -34,6 +34,7 @@ struct criteria {
 	struct pattern *shell;
 	struct pattern *app_id;
 	struct pattern *con_mark;
+	char *trigger; // event that triggered this criteria match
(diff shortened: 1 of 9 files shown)


emersion:
Sorry, we're not implementing any new window management features unless i3 implements them too. Please consider searching for or opening an i3 feature request.

Geo25rey:
@emersion since when is that the case? I know that Sway needs to remain i3 compatible, but I thought that there are features in Sway that aren't in i3. Are people discouraged from implementing new features until i3 is 100% compatible?

emersion:
There are several reasons to this:

- We don't want Sway to keep growing. Letting too many features in make it difficult to maintain Sway.
- We don't want to implement new features that could become redundant or inconsistent with i3.

Geo25rey:
So what kind of features are ok to add?

emersion:
Features that aren't about window management (e.g. input/DRM/etc), and features that are implemented/approved by i3.

danieldg:
If someone else finds this feature useful, feel free to make the i3 pull request and/or patch. I'll just maintain this patch locally.

danieldg:
Sadly, it appears that github does not understand how to remove a commit from a merge request. Re-created as https://github.com/swaywm/sway/pull/5662.

PR #5623, opened 2020-08-14
swaybar: use text subpixel antialias only where it would look good

### LINKED ISSUE #5605: swaybar/swaynag: no subpixel text antialiasing
Using Sway 1.5, on Gentoo and Arch linux

```
output DP-1 {
  subpixel rgb
}
font pango:monospace 10
bar {
  status-command ~/mystatus.sh
}
```

At this point, subpixel blending will be used by sway for window titles/decoration, but not by swaybar to draw text. Swaybar keeps using grayscale AA, even though subpixel seems to be what's wanted by render code in both cases

Testing shows that Cairo will bail out of subpixel AA if the cairo_operator is set to SOURCE. Changing swaybar to use the default OVER "fixes" it, and results in use of subpixel blending for text

Is it Cairo that could be wrong ? Changing to OVER "feels right" because AA will require blending anyway, yet the manual lets one feel that active operator and anti-aliasing modes shouldn't interfere with each other (internal requirements of AA not caller's problem)

Fenveireth:
Closes #5605

Text Subpixel antialiasing is :

- FreeType makes glyph bitmaps containing coverage percentage for each
  subpixel, instead of pixel
- Then draw by performing the blend for each subpixel, instead of pixel (e.g.
  dual-source blending in opengl)

And there's only one Alpha channel, so this extra coverage data can't
leave Cairo to reach the compositor through there.
Therefore, it can't work as intended if output text alpha != bar background
alpha. Disable it for those cases, enable it elsewhere

As for color emojis, they are RGBA bitmaps. If drawn with text alpha=1.0
and background alpha=1.0 (should be completely opaque bar), then with
'CAIRO_OPERATOR_SOURCE' texels with alpha < 1.0 result in a blend with
whatever's behind the bar, instead of the bar background

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -23,8 +23,31 @@ static const int WS_HORIZONTAL_PADDING = 5;
 static const double WS_VERTICAL_PADDING = 1.5;
 static const double BORDER_WIDTH = 1;
 
-static uint32_t render_status_line_error(cairo_t *cairo,
-		struct swaybar_output *output, double *x) {
+struct render_context {
+	cairo_t *cairo;
+	struct swaybar_output *output;
+	cairo_font_options_t *textaa_sharp;
+	cairo_font_options_t *textaa_safe;
+	uint32_t background_color;
+};
+
+static void choose_text_aa_mode(struct render_context *ctx, uint32_t fontcolor) {
+	uint32_t salpha = fontcolor & 0xFF;
+	uint32_t balpha = ctx->background_color & 0xFF;
+
+	// Subpixel antialiasing requires blend be done in cairo, not compositor
+	cairo_font_options_t *fo = salpha == balpha ?
+			ctx->textaa_sharp : ctx->textaa_safe;
+	cairo_set_font_options(ctx->cairo, fo);
+
+	// Color emojis, being semitransparent bitmaps, are leaky with 'SOURCE'
+	cairo_operator_t op = salpha == 0xFF ?
+			CAIRO_OPERATOR_OVER : CAIRO_OPERATOR_SOURCE;
+	cairo_set_operator(ctx->cairo, op);
+}
+
+static uint32_t render_status_line_error(struct render_context *ctx, double *x) {
+	struct swaybar_output *output = ctx->output;
 	const char *error = output->bar->status->text;
 	if (!error) {
 		return 0;
@@ -32,6 +55,7 @@ static uint32_t render_status_line_error(cairo_t *cairo,
 
 	uint32_t height = output->height * output->scale;
 
+	cairo_t *cairo = ctx->cairo;
 	cairo_set_source_u32(cairo, 0xFF0000FF);
 
 	int margin = 3 * output->scale;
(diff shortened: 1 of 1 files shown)


Fenveireth:
I would rather have 'OVER' for everything, but it would ruin it for setups setting all colors with a same <1 opacity, to produce the same result as an opaque bar drawn whole with that opacity value.
And that's the normal way to get a translucent bar, right ?

Xyene (commented):
Thanks for working on this! This matches my (albeit rudimentary) understanding of how subpixel aliasing works, and sounds like a good compromise. I've nitted a few things in the code, but on the most part this looks good. Please also squash the commits into one when you're done.

/ping @emersion from original thread

swaybar/render.c
```diff
@@ -477,32 +514,33 @@ static uint32_t render_status_line_i3bar(cairo_t *cairo,
...
+		return render_status_line_i3bar(ctx, x);
 	case PROTOCOL_UNDEF:
 		return 0;
 	}
 	return 0;
 }
 
-static uint32_t render_binding_mode_indicator(cairo_t *cairo,
-		struct swaybar_output *output, double x) {
+static uint32_t render_binding_mode_indicator(struct render_call* ctx,
```
Xyene, line 540: Nit: `struct render_call *ctx`.

swaybar/render.c
```diff
@@ -635,15 +680,18 @@ static uint32_t render_workspace_button(cairo_t *cairo,
 	return output->height;
 }
 
-static uint32_t render_to_cairo(cairo_t *cairo, struct swaybar_output *output) {
+static uint32_t render_to_cairo(struct render_call* ctx) {
```
Xyene, line 683: (same)

swaybar/render.c
```diff
@@ -23,8 +23,32 @@ static const int WS_HORIZONTAL_PADDING = 5;
 static const double WS_VERTICAL_PADDING = 1.5;
 static const double BORDER_WIDTH = 1;
 
-static uint32_t render_status_line_error(cairo_t *cairo,
-		struct swaybar_output *output, double *x) {
+struct render_call {
```
Xyene, line 26: Given that this is replacing `cairo_t *cairo` in most places, perhaps `struct render_context`/`cairo_context`/`cairo_state` (or similar) is a better name?

swaybar/render.c
```diff
@@ -706,26 +754,35 @@ void render_frame(struct swaybar_output *output) {
 
 	free_hotspots(&output->hotspots);
 
+	struct render_call ctx;
```
Xyene, line 757: Nit: `ctx = { 0 }` so we don't get garbage pointers in coredumps in case we forget to initialize a field.

swaybar/render.c
```diff
@@ -53,21 +77,24 @@ static uint32_t render_status_line_error(cairo_t *cairo,
...
+	choose_text_aa_mode(ctx, 0xFF0000FF);
 	pango_printf(cairo, font, output->scale, false, "%s", error);
 	*x -= margin;
 	return output->height;
 }
 
-static uint32_t render_status_line_text(cairo_t *cairo,
-		struct swaybar_output *output, double *x) {
+static uint32_t render_status_line_text(struct render_call *ctx, double *x) {
+	cairo_t *cairo = ctx->cairo;
```
Xyene, line 87: Tiny nit: `cairo` is not used until later, so this declaration could be moved to ~line 94.

swaybar/render.c
```diff
@@ -146,9 +176,10 @@ static void i3bar_block_unref_callback(void *data) {
 	i3bar_block_unref(data);
 }
 
-static uint32_t render_status_block(cairo_t *cairo,
-		struct swaybar_output *output, struct i3bar_block *block, double *x,
-		bool edge, bool use_short_text) {
+static uint32_t render_status_block(struct render_call *ctx,
+		struct i3bar_block *block, double *x, bool edge, bool use_short_text) {
+	cairo_t *cairo = ctx->cairo;
```
Xyene, line 181: (same, try to keep declarations close to where they're used)

Xyene (changes requested):
Argh, sorry, this got lost in my emails... just some stylistic things and I think this should be good to go.

Feel free to ping me if I I don't reply, it probably means I've lost the notification.

swaybar/render.c
```diff
@@ -457,8 +493,9 @@ static uint32_t predict_binding_mode_indicator_length(cairo_t *cairo,
 	return width;
 }
 
-static uint32_t render_status_line_i3bar(cairo_t *cairo,
-		struct swaybar_output *output, double *x) {
+static uint32_t render_status_line_i3bar(struct render_call* ctx, double *x) {
```
emersion, line 496: Style: `*` on the right

Xyene (approved):
LGTM, thanks for working on this!

Issue #3693, opened 2019-02-15
When setting a wallpaper image, the background switches to gray and then to the image


xyproto:
Hi,

I wrote a utility for being able to set Timed Wallpapers (wallpaper slideshows, with crossfade between images) that also works for Wayland/Sway.

This works nicely, but when the wallpaper is changed, the background blinks by first switching to gray and then to the selected image.

Steps to reproduce:
1. Install the [Mojave timed wallpaper](https://github.com/japamax/gnome-mojave-timed-wallpaper), either from GitHub or from the AUR package named `gnome-mojave-timed-wallpaper` (if you use Arch Linux).
2. Install the [monitor tools](https://github.com/xyproto/monitor), either from GitHub or from the AUR package named `monitor` (if you use Arch Linux).
3. Run this command `settimed mojave-timed`.

Then just wait a bit, and the background will blink every few minutes.

The command being run for setting the wallpaper looks like this:

    swaymsg 'output "*" background /usr/share/backgrounds/gnome/mojave/mojave_dynamic-1320.jpg fill'

This is a feature request for the background image to be switched smoothly, without the background blinking.

I'm using `sway version 1.0-beta.1-245-gdbf8e1ce (Nov 27 2018, branch 'master')` and have also tested with the latest master of sway (which fixes the problem with the windows blinking too).

Thanks for creating Sway!

emersion:
This would require to stop killing swaybg, and replacing it with some kind of IPC.

ddevault:
If we're feeling spicy we could wait for the new swaybg instance to come up before killing the old one. The z-ordering should place it on top

kennylevinsen:
I tried swapping things around so that the old swaybg client is destroyed after a new one is forked, but that still leaves some blanking.

I also discovered that sway gets rather upset if apply_output_config hangs (everything freezes until apply_output_config is done), so I'm not sure how good an idea it is to wait for swaybg for this use-case—it'll be rather annoying if sway hangs a brief moment every time wallutils swaps the background. It'll mess with fast typing. I at least concluded that my sleep(1) hack was a *terrible* idea. :)

I could create a thread to kill swaybg in the background, but everything is starting to feel a bit hacky. If only I could write `go func() { ... }`...

michaelKurowski:
+1 very annoying while using pscircle

0xmatthias:
Hello,

I would like to implement an IPC feature (on both sides), is there some work in process?

emersion:
[...]

PR #5629, opened 2020-08-18
Allow all glob patterns in output configs

### LINKED ISSUE #5632: Setting output config by name after it has been set by identifier doesn't work on sway startup but does on reload
- **Sway Version:** sway version v1.5-108b8e97 (Aug 19 2020, branch 'HEAD')

- **Debug Log:** https://gist.github.com/pedrocr/d618d47d76e7196882f65b014f6bbbac

- **Configuration File:** Default config plus some output configs like these:

```
output "Goldstar Company Ltd LG HDR 4K 0x0000333A" mode 3840x2160
output DP-1 mode 1920x1080@60hz
```

The identifier is for the monitor attached to the `DP-1` output.

- **Description:** Adjusting the output mode of the connection after the identifier doesn't work on sway startup. If I then do a reload of the config the correct mode is applied.

pedrocr:
Instead of special casing `output * ...` just make output take all glob patterns. This is useful if you want to change a config in all your HDMI or DP outputs, or want to match any monitor of a given model.

This is implemented by just storing all output commands unaltered and then matching the glob patterns to the outputs whenever needed. Match order is the same as command introduction order so that any newer commands are always respected.

Code becomes significantly simpler by doing it this way, with a bunch of complex matching logic across several functions being replaced with a straightforward merge sequence in `find_output_config(sway_output)`.

Fixes #5632

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -472,6 +472,10 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 	// this output came online, and some config items (like map_to_output) are
 	// dependent on an output being present.
 	input_manager_configure_all_inputs();
+
+	// If we've changed the config of an output restart the backgrounds as well
+	spawn_swaybg();
+
 	return true;
 }
 
@@ -507,13 +511,13 @@ static void default_output_config(struct output_config *oc,
 
 struct output_config *find_output_config(struct sway_output *sway_output) {
 	// Start with a default config for this output
-	struct output_config *result = new_output_config("merge");
+	char *name = sway_output->wlr_output->name;
+	struct output_config *result = new_output_config(name);
 	default_output_config(result, sway_output->wlr_output);
 
 	// Apply all matches in order
 	char id[128];
 	output_get_identifier(id, sizeof(id), sway_output);
-	char *name = sway_output->wlr_output->name;
 	for (int i = 0; i < config->output_configs->length; ++i) {
 		struct output_config *oc = config->output_configs->items[i];
 		if (!fnmatch(oc->name, name, 0) || !fnmatch(oc->name, id, 0)) {
@@ -640,21 +644,8 @@ bool spawn_swaybg(void) {
 		return true;
 	}
 
-	size_t length = 2;
-	for (int i = 0; i < config->output_configs->length; i++) {
-		struct output_config *oc = config->output_configs->items[i];
-		if (!oc->background) {
-			continue;
-		}
-		if (strcmp(oc->background_option, "solid_color") == 0) {
-			length += 4;
-		} else if (oc->background_fallback) {
-			length += 8;
-		} else {
-			length += 6;
-		}
-	}
-
+	// At most the program, 8 arguments per output and the terminating NULL
+	size_t length = 2 + wl_list_length(&root->all_outputs) * 8;
 	char **cmd = calloc(length, sizeof(char *));
 	if (!cmd) {
 		sway_log(SWAY_ERROR, "Failed to allocate spawn_swaybg command");
@@ -663,8 +654,13 @@ bool spawn_swaybg(void) {
 
 	size_t i = 0;
 	cmd[i++] = config->swaybg_command;
-	for (int j = 0; j < config->output_configs->length; j++) {
-		struct output_config *oc = config->output_configs->items[j];
+
+	// Iterate all the outputs and write the command
+	list_t *output_configs = create_list();
+	struct sway_output *sway_output, *tmp;
+	wl_list_for_each_safe(sway_output, tmp, &root->all_outputs, link) {
+		struct output_config *oc = find_output_config(sway_output);
+		list_add(output_configs, oc);
 		if (!oc->background) {
 			continue;
 		}
@@ -685,7 +681,7 @@ bool spawn_swaybg(void) {
 				cmd[i++] = oc->background_fallback;
 			}
 		}
-		assert(i <= length);
+		assert(i < length);
 	}
 
 	for (size_t k = 0; k < i; k++) {
@@ -693,6 +689,12 @@ bool spawn_swaybg(void) {
 	}
 
 	bool result = _spawn_swaybg(cmd);
+
 	free(cmd);
+	for (int k = 0; k < output_configs->length; k++) {
+		free_output_config(output_configs->items[k]);
+	}
+	list_free(output_configs);
+
 	return result;
 }
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -119,7 +119,101 @@ void merge_output_config(struct output_config *dst, struct output_config *src) {
 	}
 }
 
+enum output_config_changes {
+	OUTPUT_CONFIG_ENABLED = 1 << 1,
+	OUTPUT_CONFIG_WIDTH = 1 << 2,
+	OUTPUT_CONFIG_HEIGHT = 1 << 3,
+	OUTPUT_CONFIG_X = 1 << 4,
+	OUTPUT_CONFIG_Y = 1 << 5,
+	OUTPUT_CONFIG_SCALE = 1 << 6,
+	OUTPUT_CONFIG_SCALE_FILTER = 1 << 7,
+	OUTPUT_CONFIG_SUBPIXEL = 1 << 8,
+	OUTPUT_CONFIG_REFRESH_RATE = 1 << 9,
+	OUTPUT_CONFIG_CUSTOM_MODE = 1 << 10,
+	OUTPUT_CONFIG_TRANSFORM = 1 << 11,
+	OUTPUT_CONFIG_MAX_RENDER_TIME = 1 << 12,
+	OUTPUT_CONFIG_ADAPTIVE_SYNC = 1 << 13,
+	OUTPUT_CONFIG_BACKGROUND = 1 << 14,
+	OUTPUT_CONFIG_BACKGROUND_OPTION = 1 << 15,
+	OUTPUT_CONFIG_BACKGROUND_FALLBACK = 1 << 16,
+	OUTPUT_CONFIG_DPMS_STATE = 1 << 17,
+};
+
+static int output_config_changes(struct output_config *src) {
+	int changes = 0;
+
+	if (src->enabled != -1) {
+		changes |= OUTPUT_CONFIG_ENABLED;
+	}
+	if (src->width != -1) {
+		changes |= OUTPUT_CONFIG_WIDTH;
+	}
+	if (src->height != -1) {
+		changes |= OUTPUT_CONFIG_HEIGHT;
+	}
+	if (src->x != -1) {
+		changes |= OUTPUT_CONFIG_X;
+	}
+	if (src->y != -1) {
+		changes |= OUTPUT_CONFIG_Y;
+	}
+	if (src->scale != -1) {
+		changes |= OUTPUT_CONFIG_SCALE;
+	}
+	if (src->scale_filter != SCALE_FILTER_DEFAULT) {
+		changes |= OUTPUT_CONFIG_SCALE_FILTER;
+	}
+	if (src->subpixel != WL_OUTPUT_SUBPIXEL_UNKNOWN) {
+		changes |= OUTPUT_CONFIG_SUBPIXEL;
+	}
+	if (src->refresh_rate != -1) {
+		changes |= OUTPUT_CONFIG_REFRESH_RATE;
+	}
+	if (src->custom_mode != -1) {
+		changes |= OUTPUT_CONFIG_CUSTOM_MODE;
+	}
+	if (src->transform != -1) {
+		changes |= OUTPUT_CONFIG_TRANSFORM;
+	}
+	if (src->max_render_time != -1) {
+		changes |= OUTPUT_CONFIG_MAX_RENDER_TIME;
+	}
+	if (src->adaptive_sync != -1) {
+		changes |= OUTPUT_CONFIG_ADAPTIVE_SYNC;
+	}
+	if (src->background) {
+		changes |= OUTPUT_CONFIG_BACKGROUND;
+	}
+	if (src->background_option) {
+		changes |= OUTPUT_CONFIG_BACKGROUND_OPTION;
+	}
+	if (src->background_fallback) {
+		changes |= OUTPUT_CONFIG_BACKGROUND_FALLBACK;
+	}
+	if (src->dpms_state != 0) {
+		changes |= OUTPUT_CONFIG_DPMS_STATE;
+	}
+
+	return changes;
+}
+
 struct output_config *store_output_config(struct output_config *oc) {
+	// Get rid of a previous output config that has the same name and changes
+	// the same things, as it has been superseded. Since we do this on every
+	// insert there can only ever be one of these. This prevents the output
+	// config list of commands growing unbounded.
+	for (int i = 0; i < config->output_configs->length; ++i) {
+		struct output_config *list_oc = config->output_configs->items[i];
+		if (!strcmp(oc->name, list_oc->name) &&
+			output_config_changes(oc) == output_config_changes(list_oc)) {
+			free_output_config(config->output_configs->items[i]);
+			list_del(config->output_configs, i);
+			break;
+		}
+	}
+
+	// Then simply add the command to the bottom of the list where it will be
+	// applied last and thus have the biggest priority.
 	list_add(config->output_configs, oc);
 
 	sway_log(SWAY_DEBUG, "Config stored for output %s (enabled: %d) (%dx%d@%fHz "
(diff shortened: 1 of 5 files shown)


sway/config/output.c
```diff
@@ -125,104 +119,8 @@ void merge_output_config(struct output_config *dst, struct output_config *src) {
...
-			free_output_config(oc);
-			oc = current;
-		}
-		list_add(config->output_configs, oc);
-	} else {
-		// New wildcard config. Just add it
-		sway_log(SWAY_DEBUG, "Adding output * config");
-		list_add(config->output_configs, oc);
-	}
+	list_add(config->output_configs, oc);
```
emersion, line 123: This keeps adding new items to the list instead of merging configs. This means having a script that executes an output command will make Sway allocate more and more memory, resulting in very long output config lists.

pedrocr: Indeed. I considered doing a merge there but opted to simplify. If that's a problem just merging when there are exact matches on the pattern will fix that and be contained within that function.

pedrocr: I had to bring back the simple `list_add(new_oc)` because to keep the existing semantics of newer commands always overriding older ones that is needed.

Minimizing memory usage would require a somewhat more involved merge, where only commands that affect the same thing (e.g., resolution) are merged because they're a strict superset of the older command. That's doable if this is considered very important but requires quite a bit more code.

pedrocr: Fixed the memory usage problem by just deleting old commands that have the same name/pattern and change the same things in the output config. This bounds the memory usage. It can still go higher than current master but not by any huge amount.

pedrocr:
At this point this PR is really a refactor of the code that then allows glob matching to be done easily. The old behavior could be brought back just by changing the line that does the two `fnmatch()` calls into three `strcmp()` calls. I can put the introduction of glob matching into a separate commit at the end if it makes more sense in the commit log.

pedrocr:
I've split out the patch series as I've found that the initial refactor fixes at least one bug with the existing code, unrelated to the matching itself.

pedrocr:
I've been running this locally with no issues. Any comments on it?

sway/config/output.c
```diff
@@ -513,125 +504,43 @@ static void default_output_config(struct output_config *oc,
...
-		struct sway_output *sway_output) {
-	const char *name = sway_output->wlr_output->name;
-
-	struct output_config *oc_id_on_name = NULL;
-	struct output_config *oc_name = NULL;
-	struct output_config *oc_id = NULL;
+struct output_config *find_output_config(struct sway_output *sway_output) {
+	// Start with a default config for this output
+	struct output_config *result = new_output_config("merge");
+	default_output_config(result, sway_output->wlr_output);
```
RedSoxFan, line 510: Just as a note, this is currently only used on reload. I think it is fine for this to be applied unconditionally, but there may be an unintentional side effect that I'm overlooking.

RedSoxFan (changes requested):
I'm not sure how I feel about this as a whole, but here are some comments.

Something to consider is swaybg (as well as anything that is compatible with `swaybg_command` like [oguri](https://github.com/vilhalmer/oguri)). We build a command based on the output configs. swaybg currently does not support globbing and would fail to correctly set the wallpaper. Additionally, the output command is unique in that it allows for multiple subcommands to be provided on the same line (for backwards compatibility). This may result in multiple output configs having a wallpaper for the same output glob pattern, which would result in the output being specified multiple times in the swaybg spawn command.

pedrocr:
I hadn't looked at `swaybg_command` but if I'm understanding things correctly it's easy to solve with full backwards compatibility. Currently `spawn_swaybg()` is iterating over the `output_configs` and building the command from that. Instead it should iterate over the outputs, call `find_output_config()` with each of them to get the background options in effect, and add those to the command line with the output name. No globbing support is needed at all. How does that sound?

sway/config/output.c
```diff
@@ -513,125 +504,43 @@ static void default_output_config(struct output_config *oc,
...
 }
 
-struct output_config *find_output_config(struct sway_output *output) {
-	char id[128];
-	output_get_identifier(id, sizeof(id), output);
-	return get_output_config(id, output);
-}
-
-void apply_output_config_to_outputs(struct output_config *oc) {
+void apply_output_config_to_outputs(void) {
```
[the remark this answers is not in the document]
pedrocr: If this is a problem, it would be simple to add a `struct output_config` to each output that stores whatever is the current output configuration so that if the new one is the same the change is turned into a no-op.

What would be the worry? Flickering if the underlying wlroots code does something with the hardware even though no config has been changed?

sway/config/output.c
pedrocr: `default_output_config()` itself doesn't have any side effects, it just initializes the struct. Having the same exact code path for a config change and for reload seems like an advantage. Bug #5632 is the kind of thing that can happen when there are two code paths.

sway/config/output.c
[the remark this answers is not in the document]
pedrocr: wlroots seems to already turn into no-ops changes that don't change anything so flickering shouldn't be possible:

https://github.com/swaywm/wlroots/blob/28cedb56237efdf388fed713ddb401a5d8d07434/types/wlr_output.c#L164-L166

Doing this at the wlroots level seems reasonable too. Instead of trying to have sway keep more state to try and avoid issuing unneeded commands.

pedrocr:
Thanks for the review @RedSoxFan. Let me know if you think these two remaining things are ok and I'll do the `swaybg_command` changes as well if so.

sway/config/output.c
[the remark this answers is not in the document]
emersion: Yes, re-applying the same configuration should be perfectly fine. wlroots handles this so that compositors don't need to worry about it.

RedSoxFan:
> I hadn't looked at `swaybg_command` but if I'm understanding things correctly it's easy to solve with full backwards compatibility. Currently `spawn_swaybg()` is iterating over the `output_configs` and building the command from that. Instead it should iterate over the outputs, call `find_output_config()` with each of them to get the background options in effect, and add those to the command line with the output name. No globbing support is needed at all. How does that sound?

That would require also restarting swaybg everytime an output is connected or disconnected. Currently, we only restart it when there would be a change.

pedrocr:
I've implemented the swaybg change with the restart on any output config change. Is the restart problem that swaybg doesn't even need to be restarted if the resolution changes because it can adapt by itself? It should only be restarted if we're altering the command line commands it receives?

emersion:
>It should only be restarted if we're altering the command line commands it receives?

Correct.

pedrocr:
The simplest way I can think of to do that is to keep the command line between calls and just do a no-op if it's the same. It would mirror what happens on mode change with wlroots turning a mode set into a no-op. Seems simple to implement too.

pedrocr:
I've implemented the check that the command is not the same, and confirmed that it eliminates the restarts when the background config hasn't changed. I think that takes care of the last issue discussed, but maybe there are other doubts about this.

RedSoxFan:
I'm not a huge fan of this PR as a whole.
1. I'm not convinced that full glob matching is actually needed.
2. I realize it will have no noticeable impact for most, but this does result in higher memory usage due to each (non shadowed) delta being stored long term as well as the swaybg command.
3. Yes, wlroots will prevent output commits without any effect from touching the outputs, but it is still unnecessary to process the output configs of every single output any time that any output config changes or any output is hotplugged.
4. Similar to the previous point, regenerating and comparing the swaybg command every single output config change or hotplug is also additional computation that will be unnecessary in most cases.
5. As mentioned above, having a swaybg command that has to be regenerated for hotplugging will make #3693 more noticeable. If I had to guess, the most common use case is probably `output * bg </path/to/wallpaper> <mode>`. This does not require any knowledge of the active outputs to pass this information off to swaybg. Depending on the output, it's possible that the two cycles occur before the output is even presenting anything to the user so they don't even see the default background.

If others disagree with me, I won't block the PR from being merged (and will clear my requested changes review), but I doubt this will be getting an approval from me.

@emersion I'm curious what your thoughts are on this PR

pedrocr:
I'll try to answer the objections if it helps with the discussion, but I can also close the PR if this is not really wanted.

> I'm not convinced that full glob matching is actually needed.

My use case is for applying the same config to several screens of the same make and model. It's a common situation in hot-seat offices that all have the same monitors. That specific use case could potentially be implemented by itself by adding make/model match but I can see people wanting to do things like set variable refresh on all DisplayPort outputs and other conveniences. I agree this isn't a must have feature though, so if it has many other downsides it can just be scrapped.

> I realize it will have no noticeable impact for most, but this does result in higher memory usage due to each (non shadowed) delta being stored long term as well as the swaybg command.

From a quick calculation it seems to be less than 1000 bytes per output in the absolute worst case scenario. It can also result in less memory when you can use a glob whereas before several configs are needed. I don't think there's any platform sway is interested where it makes a difference either way.

> Yes, wlroots will prevent output commits without any effect from touching the outputs, but it is still unnecessary to process the output configs of every single output any time that any output config changes or any output is hotplugged.
> Similar to the previous point, regenerating and comparing the swaybg command every single output config change or hotplug is also additional computation that will be unnecessary in most cases.

There's a bit of extra calculation. If the processing cost was too high caching could be done relatively easily but there's no calculation that's particularly expensive here, so the simplicity seems like a win.

>As mentioned above, having a swaybg command that has to be regenerated for hotplugging will make #3693 more noticeable. If I had to guess, the most common use case is probably output * bg </path/to/wallpaper> <mode>. This does not require any knowledge of the active outputs to pass this information off to swaybg. Depending on the output, it's possible that the two cycles occur before the output is even presenting anything to the user so they don't even see the default background.

This should allow less cases of showing #3693 actually. It only ever restarts swaybg if it needs to be restarted because the command line has changed. With the existing code if you issue the exact same background command over swaymsg the background flickers even though nothing has changed. The `output * bg` case could be special cased if that's wanted.

As I said I'm happy to just close this if there's no interest. The globbing itself I've found helpful but the PR has ended up reworking the code. Maybe too much? It's less code that to me seems simpler with at least one bug fixed.

pedrocr:
Just noticed that the same feature request and equivalent PR also exists for kanshi:

https://github.com/emersion/kanshi/issues/72
https://github.com/emersion/kanshi/pull/73

pedrocr:
I've been running this for nearly 6 months with no issues now. Is there anything else I can help with in the discussion?

emersion:
As I stated elsewhere, I'm still not convinced globs are a good feature to have for output configs.

pedrocr:
@emersion hadn't seen your opinion on that then. I was asking because both this and the kanshi PRs have been left open but without any further discussion. Feel free to close this if you don't want the feature then. I'll have to make do without it.

emersion:
Ref https://github.com/emersion/kanshi/issues/72

pedrocr:
To be able to implement the glob patterns I had to refactor the code and fixed at least one bug in the process. The glob support is a single line change after that. If it would help I can submit a PR that doesn't change functionality at all and then make the glob pattern support a trivial PR on top of that.

emersion:
Sure, feel free to split the refactoring part up if it makes sense to merge it standalone. Disclaimer, I haven't looked again at the refactoring yet.

pedrocr:
The `fnmatch()` change is a one liner either way (plus documentation) so the PR is already representative of what that would be.

WhyNotHugo:
I'd also really appreciate this feature. A bit of background:

My monitor changes id depending on which ports I use on my laptop and the monitor, so without globbing, I need to keep all permutations of the ids around. Any time I make a change, I usually have to change all of them too.

There are around 10 desks at my office all which all have monitors of the same model, so glob matching would allow one line to adapt all 10 of them. So far, I've to reconfigure sway each time I use a new one, and keep around a pile of identifiers in my config.

I understand the desire for optimisation, but 1kb memory and a few string comparisons when there's an output change seems a bit extreme; I'm sure there's much juicier targets for optimisations.

h-2:
I really need this feature!

In my shared office, the screens have individual IDs, so I need to configure each one individually, which is super annoying, e.g.:
```
output 'LG Electronics LG ULTRAFINE 308MALFA5952' {
...
}
output 'LG Electronics LG ULTRAFINE 211MAFCQ5123' {
...
}
...
```

I would like to be able to do just:
```
output 'LG Electronics LG ULTRAFINE *' {
}
```

josecastillolema:
I'd also really appreciate this feature. Shared office, lots of individual IDs in the screens.

hw0lff:
I just stumbled upon this thread.
I too have shared offices spaces and had the need for a more advanced output matching.
I used kanshi for a while but it didn't offer the functionality I needed (to manage >30 profiles deterministically).

I've pondered about a _complete_ solution over the last 2 years.
Globbing cannot and will never be able to describe all monitor configurations due to being to unspecific in certain cases.
A single regex alone will not be enough either since it could match to much.
A complete solution must be able to distinguish between several (read: all) attributes of a monitor.

I came up with a matching procedure that allows full text matching, substring comparison and regex matching for every monitor attribute individually.
It needed a separate syntax to be able to describe an output configuration completely.
With that procedure it is possible to express **every** output configuration.

However, I don't think this procedure belongs into sways codebase for several reasons.
1. It would probably break i3 compatibility.
2. The code that implements this will be very complex[^1].
   It needs an extended version of a full cardinality graph matching algorithm.
   It gets more complex when specifying modes add constraints to the procedure.
3. I would prefer to keep sway simple rather to add this complexity.
   The current way of configuring outputs with sway is simple.
   simple to use, simple to understand. This is good.
   For most people the simple way is enough to get around.
4. This advanced matching feature can be achieved with an external program using the wlroots output-management protocol.
   It won't add a maintenance burden to the sway maintainers **and** it can be used with every other wlroots compositor.


P.S.:
This complete matching procedure is already implemented in [shikane].
For anyone interested, the custom syntax can be found in the [manual] under the search field.

Reading through this myself, it may seem like a self-promotion at first glance.
However, it is meant as a kind warning that the globbing support of this PR will not cover all configuration needs.
I've racked my brain way too much about this topic and abandonded the idea of relying solely on extended globbing a long time ago.
It will only push the threshold of perceptibility of the underlying problem further back. 


[^1]: my implementation has ~2kloc for the procedure alone

[shikane]: https://gitlab.com/w0lff/shikane
[manual]: https://w0lff.gitlab.io/shikane/shikane.5.html

hw0lff:
The issue https://github.com/swaywm/sway/issues/6410 is a perfect example of what I meant with the shift of perceptibility of the underlying problem.
This globbing PR and #6410 talk about different ways to find a solution to their specific use cases.

When we look at them more broadly then both are about advanced output matching.
However, both see only their specific subset of the problem.
Implementing one (e.g. the glob match) would be just reducing the set of affected people.
It'd shift the focus on the "better & advanced output matching" feature further away.
When I said "affected people", I mean affected by the lack of advanced output matching.

What about matching a serial number?
What if any output should be chosen if the supplied mode can be applied?
What if an output should be chosen only if the supplied mode fits && the vendor matches to a given regex?
What about this weird mode which is @ 50,863Hz? Will the decimals be truncated or rounded when matching?

A _complete_ solution can handle all of the above.

pedrocr:
The part of this PR that actually implements glob patterns is a single line. The more extensive changes were needed to replace the complex and error prone config merge code with a much simpler method of just applying all the rules all the time in the order the user specified them. That's how at least one bug is fixed by this. I've already offered to do a PR without the actual glob pattern single line change if that makes it easier. It would fix at least one bug and not have any functionality changes. The additional memory overhead is minimal.

Since the glob pattern change is then a one-liner match difference I'd still have that over more complex stuff. It doesn't cover all cases but covers quite a few that are useful, most commonly the "office full of identical monitors with different serial numbers" setup.

WhyNotHugo:
In principle, I agree with approaches like [kanshi](https://git.sr.ht/~emersion/kanshi/), [swww](https://github.com/Horus645/swww/) or [shikane](https://gitlab.com/w0lff/shikane), where an external client takes care of output-specific configuration. There's no _need_ for the compositor to handle all this itself.

In practice, the main issue is that these clients start _after_ the compositor, so the compositor renders first with the default scale/resolution/rotation/etc _and then_ the external client applies the correct settings.

I can't think of a sensible way to work around this. The only thing that comes to mind is telling sway to delay rendering until a client has explicitly configured outputs... but honestly, this sounds like it could have a million edge cases where sway becomes unusable.

kennylevinsen:
If we felt a need, we could do something like `output_config_cmd ...` to tell sway that what it's starting is an output management client, or `output_control manual`.

But right now, the output management protocol only handles a subset of our output configuration options. Render bit depth, render time, color profile, tearing, subpixels, scale filtering, background, soon hdr...

kennylevinsen:
> This design forces the output manager to be a child of the compositor, so it can't be managed by the user service manager (e.g.: restart/reload/etc), and it won't even be properly supervised.

The command could just be a start command or launching a supervisor, whereas the latter option (`output_control manual`) in the same sentence makes no assumptions about how the configuration manager is executed.

PR #5641, opened 2020-08-23
exec: fix validation during config reload

kupospelov:
Repro steps:
1. Add `exec` (with no arguments) to your config.
2. Reload sway.

Sway crashes with the "argc should be positive" error from `join_args`.

The `exec` command just calls `exec_always`, however a few shared validation checks should be performed before this call.


emersion (approved):
LGTM, thanks!

Issue #3007, opened 2018-10-27
Jetbrains autocomplete popup loosing focus immediately


Ogromny:
Hi,

When I do a "fuzzy search" ( Shift Shift ) or a "search file" ( Ctrl + N ), the popup appear and disappear in 1ms, I actually tried to uncheck "Hide navigation popups on focus loss" but it doesn't works :/.

Has anyone ever encountered this problem?

sxmichaels:
Both of those work normally for me in CLion. What Jetbrains product are you using? Do you have focus-follows-mouse enabled (I don't)?

Ogromny:
With CLion it's doesn't works, but, I tried with Intellij and it's works... I will try to reinstall CLion

ESipalis:
This should be an issue with jetbrains 2018.1 products: https://youtrack.jetbrains.com/issue/IDEA-188824

I60R:
The same problem occurs for me.  
I use Intellij IDEA EAP.  
Both, sway and IDEA are up to date.  
Graphic driver - nouveau (if that matter).  
Under GNOME Wayland everything works fine.

pimeys:
Having the same problem with intel drivers and intellij fuzzy search. Did anybody find any workaround yet?

I60R:
Yes, switching boot JDK to OpenJDK as suggested in Jetbrains issue tracker solved issue for me. IDEA now looks ugly, but at least it's usable

tindzk:
Thanks for the workaround, @I60R. For everyone else affected by this issue, you can change IDEA's JDK by setting the following environment variable:

```shell
export IDEA_JDK=/usr/lib/jvm/java-8-openjdk  # Substitute path to the official JDK
```

francma:
"Search everywhere" works for me on PHPStorm 2018.2.6.

2018.3 introduced [New search popup](https://www.jetbrains.com/phpstorm/whatsnew/#v2018-3-new-search-popup) and that is broken. It looks like sway isn't able to recognize that it is a dialog and spawns the popup next to the PHPStorm instance and then immediately closes it as it has no focus.

[PHPStorm 2018.3.log](https://gist.github.com/francma/b0f67ed95052311b0e167ff2f28df9dc)

Adding also PHPStorm 2018.2 where it is working fine for comparasion.

[PHPStorm 2018.2.log](https://gist.github.com/francma/02172e6c3d28ba57612d01060f8145c3)

[...]

PR #5646, opened 2020-08-27
Re-focus on parent surface if it is available

BrassyPanache:
First I must say that I am not sure if my approach is sound.

My primary issue was IntelliJ IDEA's code suggestion pop-up not returning focus
to the active editing window.

I have spent some time looking at the changes of @Xyene (#5398) and
@RyanDwyer (#2103). I think my proposed change maintains the status
quo for the most part whilst fixing my focus issue.

I have verified that @Xyene's fix for IntelliJ sub-menus still works.

I have done basic testing which consists of:

- Chrome
- IntelliJ IDEA 2020.2.1
- VSCode
- Alacritty

It seems to hold up. I at least didn't see any obvious errors.

Relates to #3007

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -92,20 +92,15 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
 	wl_list_remove(&surface->commit.link);
 
 	struct sway_seat *seat = input_manager_current_seat();
-	if (seat->wlr_seat->keyboard_state.focused_surface ==
-			xsurface->surface) {
-
-		// Try to find another unmanaged surface from the same process to pass
-		// focus to. This is necessary because some applications (e.g. Jetbrains
-		// IDEs) represent their multi-level menus as unmanaged surfaces, and
-		// when closing a submenu, the main menu should get input focus.
-		struct sway_xwayland_unmanaged *current;
-		wl_list_for_each(current, &root->xwayland_unmanaged, link) {
-			struct wlr_xwayland_surface *prev_xsurface =
-				current->wlr_xwayland_surface;
-			if (prev_xsurface->pid == xsurface->pid &&
-					wlr_xwayland_or_surface_wants_focus(prev_xsurface)) {
-				seat_set_focus_surface(seat, prev_xsurface->surface, false);
+	if (seat->wlr_seat->keyboard_state.focused_surface == xsurface->surface) {
+		// This simply returns focus to the parent surface if there's one available.
+		// This seems to handle JetBrains issues.
+		if (xsurface->parent &&
+				wlr_surface_is_xwayland_surface(xsurface->parent->surface)) {
+			struct wlr_xwayland_surface *next_surface =
+				wlr_xwayland_surface_from_wlr_surface(xsurface->parent->surface);
+			if (wlr_xwayland_or_surface_wants_focus(next_surface)) {
+				seat_set_focus_surface(seat, xsurface->parent->surface, false);
 				return;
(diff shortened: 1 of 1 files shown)


sway/desktop/xwayland.c
```diff
@@ -92,22 +92,12 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
...
-		wl_list_for_each(current, &root->xwayland_unmanaged, link) {
-			struct wlr_xwayland_surface *prev_xsurface =
-				current->wlr_xwayland_surface;
-			if (prev_xsurface->pid == xsurface->pid &&
-					wlr_xwayland_or_surface_wants_focus(prev_xsurface)) {
-				seat_set_focus_surface(seat, prev_xsurface->surface, false);
-				return;
-			}
+	if (seat->wlr_seat->keyboard_state.focused_surface == xsurface->surface) {
+		// This simply returns focus to the parent surface if there's one avaialble.
```
Xyene, line 96: Small nit here.

```suggestion
		// This simply returns focus to the parent surface if there's one available.
```

sway/desktop/xwayland.c
```diff
@@ -92,22 +92,12 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
...
+		// This seems to handle JetBrains issues.
+		if (xsurface->parent && xsurface->parent->surface) {
```
Xyene, line 98: I think this should still be calling `wlr_xwayland_or_surface_wants_focus`, does doing so break this?

BrassyPanache: I've just tested a new build with this suggestion and all seems to work as expected. I'm not sure that I've done it all correctly as I'm new to the code base :smile:

BrassyPanache:
Thanks for the super quick review.  In my testing I added quite a bit of logging to verify that the same surfaces were given focus.  As far as I could tell it was the case.  I removed the additional logging before I committed.  I've attached one of my annotated [logs](https://github.com/swaywm/sway/files/5134103/sway.1.log) for reference.

sway/desktop/xwayland.c
```diff
@@ -92,20 +92,15 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
...
-			struct wlr_xwayland_surface *prev_xsurface =
-				current->wlr_xwayland_surface;
-			if (prev_xsurface->pid == xsurface->pid &&
-					wlr_xwayland_or_surface_wants_focus(prev_xsurface)) {
-				seat_set_focus_surface(seat, prev_xsurface->surface, false);
+	if (seat->wlr_seat->keyboard_state.focused_surface == xsurface->surface) {
+		// This simply returns focus to the parent surface if there's one available.
+		// This seems to handle JetBrains issues.
+		if (xsurface->parent &&
+			wlr_surface_is_xwayland_surface(xsurface->parent->surface)) {
```
Xyene, line 99:
```suggestion
				wlr_surface_is_xwayland_surface(xsurface->parent->surface)) {
```

(Otherwise it's easy to miss that this statement is part of the if condition and not its body.)

Xyene (commented):
Thanks, apart from a small nit this looks good to me.

Also, don't forget to rebase + squash your commits into one, then we can merge this.

sway/desktop/xwayland.c
BrassyPanache: Thanks again for the review. This should now be resolved. I've also rebased and squished into a single commit.

Xyene (approved):
Thanks!

PR #5649, opened 2020-08-27
Use wlr_output_event_commit

emersion:
Instead of listening to both transform and scale events, we can listen
to the commit event and use the new wlr_output_event_commit struct to
decide what to do.

This de-duplicates some of the work we were doing twice when an output
was re-configured.

Depends on [1].

[1]: https://github.com/swaywm/wlroots/pull/2315


ddevault:
Thanks!

Issue #4600, opened 2019-10-01
Hitman game launcher is not clickable


jtheoof:
* Sway Version:

sway version 1.2 (Sep 26 2019, Arch Linux)

* Debug Log:

No useful log in `journalctl`:

```
Sep 30 22:35:40 ansan /usr/lib/gdm-wayland-session[1650]: >>> Adding process 31201 for game ID 863550
Sep 30 22:35:40 ansan /usr/lib/gdm-wayland-session[1650]: >>> Adding process 31213 for game ID 863550
Sep 30 22:35:40 ansan /usr/lib/gdm-wayland-session[1650]: Fontconfig warning: "/etc/fonts/conf.avail/05-reset-dirs-sample.conf", line 6: unknown element "reset-dirs"
Sep 30 22:35:41 ansan /usr/lib/gdm-wayland-session[1650]: ERROR: ld.so: object '/home/jattali/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
Sep 30 22:35:41 ansan /usr/lib/gdm-wayland-session[1650]: Game update: AppID 863550 "", ProcID 31184, IP 0.0.0.0:0
Sep 30 22:35:41 ansan /usr/lib/gdm-wayland-session[1650]: Setting breakpad minidump AppID = 863550
Sep 30 22:35:41 ansan /usr/lib/gdm-wayland-session[1650]: Steam_SetMinidumpSteamID:  Caching Steam ID:  76561198001543971 [API loaded no]
Sep 30 22:35:41 ansan /usr/lib/gdm-wayland-session[1650]: ERROR: ld.so: object '/home/jattali/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
Sep 30 22:35:41 ansan /usr/lib/gdm-wayland-session[1650]: ERROR: ld.so: object '/home/jattali/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
Sep 30 22:35:41 ansan /usr/lib/gdm-wayland-session[1650]: Game update: AppID 863550 "", ProcID 31229, IP 0.0.0.0:0
[...]

PR #5650, opened 2020-08-29
xwayland: support views that change override-redirect status

junglerobba:
This fixes https://github.com/swaywm/sway/issues/4600 by destroying the view and recreating it with the correct type (same as this Wayfire patch https://github.com/WayfireWM/wayfire/pull/706 )

I have tested with the Warframe launcher and Dark Souls 3, which both work fine with this.

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -118,10 +118,34 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
 static void unmanaged_handle_destroy(struct wl_listener *listener, void *data) {
 	struct sway_xwayland_unmanaged *surface =
 		wl_container_of(listener, surface, destroy);
+	wl_list_remove(&surface->request_configure.link);
 	wl_list_remove(&surface->map.link);
 	wl_list_remove(&surface->unmap.link);
 	wl_list_remove(&surface->destroy.link);
+	wl_list_remove(&surface->override_redirect.link);
 	free(surface);
+}
+
+static void handle_map(struct wl_listener *listener, void *data);
+
+struct sway_xwayland_view *create_xwayland_view(struct wlr_xwayland_surface *xsurface);
+
+static void unmanaged_handle_override_redirect(struct wl_listener *listener, void *data) {
+	struct sway_xwayland_unmanaged *surface =
+		wl_container_of(listener, surface, override_redirect);
+	struct wlr_xwayland_surface *xsurface = surface->wlr_xwayland_surface;
+
+	bool mapped = xsurface->mapped;
+	if (mapped) {
+		unmanaged_handle_unmap(&surface->unmap, NULL);
+	}
+
+	unmanaged_handle_destroy(&surface->destroy, NULL);
+	xsurface->data = NULL;
+	struct sway_xwayland_view *xwayland_view = create_xwayland_view(xsurface);
+	if (mapped) {
+		handle_map(&xwayland_view->map, xsurface);
+	}
 }
 
 static struct sway_xwayland_unmanaged *create_unmanaged(
@@ -144,6 +168,8 @@ static struct sway_xwayland_unmanaged *create_unmanaged(
 	surface->unmap.notify = unmanaged_handle_unmap;
 	wl_signal_add(&xsurface->events.destroy, &surface->destroy);
 	surface->destroy.notify = unmanaged_handle_destroy;
+	wl_signal_add(&xsurface->events.set_override_redirect, &surface->override_redirect);
+	surface->override_redirect.notify = unmanaged_handle_override_redirect;
 
 	return surface;
 }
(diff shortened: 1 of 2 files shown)


sway/desktop/xwayland.c
```diff
@@ -470,6 +489,18 @@ static void handle_map(struct wl_listener *listener, void *data) {
 	transaction_commit_dirty();
 }
 
+static void handle_override_redirect(struct wl_listener *listener, void *data) {
+	struct sway_xwayland_view *xwayland_view =
+		wl_container_of(listener, xwayland_view, map);
```
emersion, line 494: s/map/override_redirect/

kqzkqz:
thanks for this fix! been playing around with it applied for the past week and it seems to work well. fixes one of the last remaining xwayland issues i've been experiencing (that i've run into at least) 

just as some user feedback, one quirk I am seeing is that for launchers like the league of legends client, it seems that when the view is destroyed and re-created (happens a lot in champ select), it garbles up something with the pointer coordinates in the app itself, i.e the pointer is actually offset 3 inches to the left. however when applying a move event to it (i.e super + mouse1 and drag) looks like the state is updated again and the cursor maps correctly. i'll make a follow-up issue for this if needed.

sway/desktop/xwayland.c
```diff
@@ -465,6 +484,18 @@ static void handle_map(struct wl_listener *listener, void *data) {
...
+static void handle_override_redirect(struct wl_listener *listener, void *data) {
+	struct sway_xwayland_view *xwayland_view =
+		wl_container_of(listener, xwayland_view, override_redirect);
+	struct wlr_xwayland_surface *xsurface = data;
+	struct sway_view *view = &xwayland_view->view;
+
+	handle_destroy(&xwayland_view->destroy, view);
+	xsurface->data = NULL;
+	struct sway_xwayland_unmanaged *unmanaged = create_unmanaged(xsurface);
+	unmanaged_handle_map(&unmanaged->map, xsurface);
```
emersion, line 496: We should probably only do this if the `xsurface` is already mapped?

sway/desktop/xwayland.c
```diff
@@ -465,6 +484,18 @@ static void handle_map(struct wl_listener *listener, void *data) {
 	transaction_commit_dirty();
 }
 
+static void handle_override_redirect(struct wl_listener *listener, void *data) {
+	struct sway_xwayland_view *xwayland_view =
+		wl_container_of(listener, xwayland_view, override_redirect);
+	struct wlr_xwayland_surface *xsurface = data;
+	struct sway_view *view = &xwayland_view->view;
+
+	handle_destroy(&xwayland_view->destroy, view);
```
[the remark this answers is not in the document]
junglerobba: Rebased and updated, will squash commits if the changes make sense.

sway/desktop/xwayland.c
```diff
@@ -121,9 +121,35 @@ static void unmanaged_handle_destroy(struct wl_listener *listener, void *data) {
...
+	struct sway_xwayland_unmanaged *surface =
+		wl_container_of(listener, surface, override_redirect);
+	struct wlr_xwayland_surface *xsurface = surface->wlr_xwayland_surface;
+
+	bool mapped = xsurface->mapped;
+	if (mapped) {
+		unmanaged_handle_unmap(&surface->unmap, NULL);
+	}
+
+	wl_list_remove(&surface->request_configure.link);
```
emersion, line 142: One last comment: why do we need to handle this listener as a special case?

junglerobba: I didn't find it in unmanaged_handle_destroy or unmanaged_handle_unmap and wasn't sure whether to add it in one of those. For this case it needs to be removed though, otherwise sway crashes when trying to move/close or otherwise interact with a view that switched from unmanaged to managed.
I can move if it to one of those functions if you think that's more appropriate.

emersion: Ah, it's probably missing then and not crashing "by chance". If the listener is added in `create_xwayland_view`, it should be removed in `handle_destroy`. If it's added in `handle_map`, it should be removed in `handle_unmap`.

junglerobba: Ok, I moved it to `unmanaged_handle_destroy` then. For managed views the listener was already handled correctly.

emersion (approved):
LGTM, thanks!

microo8:
When will there be a new release with this patch? This is the only reason why I have installed i3 (for gaming)

emersion:
We don't do ETAs, sorry. There'll be a new release when we'll decide it's time for a new release.

damianatorrpm:
> When will there be a new release with this patch? This is the only reason why I have installed i3 (for gaming)

It's merged, you can always build it from source.

microo8:
So I compiled it from https://aur.archlinux.org/packages/sway-git/
And still can't click on the fallout4 launcher executed by steam

```
sway --version                                                                                                                      ~
sway version 1.5-0cb9282a (Oct 26 2020, branch 'master')
```

PR #5656, opened 2020-09-02
swaynag: adds option to separately specify the text color for buttons

oliver-giersch:
adds the `button_text` option, which allows setting a separate RRGGBB[AA] value for the text color of buttons. If not set, the generic text color is used.

example:
```bash
$ ls -a ~/ | ./swaynag -t warning -m "files in home" -L "show..." -l --text=FF0000 --button-text=000000
```


emersion (approved):
LGTM, thanks!

PR #5658, opened 2020-09-02
input/libinput: remove input type property bias

### LINKED ISSUE #5654: Left-handedness and calibration matrix won't stick to XP Pen tablet
- **Sway Version:**
sway version 1.5-6991ac8c (Aug 22 2020, branch 'master')

- **Debug Log:**
 http://0x0.st/iEd8.txt

- **Configuration File:**
http://0x0.st/iEdN.txt
If needed, full configuration can be found [here](https://github.com/unresolvedsymbol/dotfiles/blob/master/.config/sway/config)

- **Description:**
Setting left handed mode or the calibration matrix on my XP Pen G430S has no effect yet returns success.
https://streamable.com/oy1i9s
Note: The video shows left handedness enabled for the mouse input of the tablet but this has no effect and coordinates are still inverted. Trying to use calibration matrix on either of the two doesn't change it at all, however it can change it on my touchscreen.

RedSoxFan:
Closes #5654 

This changes it so all libinput config options are set on any device
that supports it. Previously, only a subset of libinput config options
were being considered depending on the input type. Instead of trying to
guess which properties the device may support, attempt to set any
configured property regardless of the device type. All of the functions
already have early returns in them for when the device does not actually
support the property. This brings the configuration side inline with
describe_libinput_device for the IPC side.This change was prompted
by a tablet tool showing the calibration matrix property in the IPC
message, but not being able to actually change it since that property
was only being considered for the touch input type.

### DIFF
--- a/sway/input/libinput.c
+++ b/sway/input/libinput.c
@@ -186,17 +186,23 @@ static bool set_calibration_matrix(struct libinput_device *dev, float mat[6]) {
 	return changed;
 }
 
-static bool config_libinput_pointer(struct libinput_device *device,
-		struct input_config *ic, const char *device_id) {
-	sway_log(SWAY_DEBUG, "config_libinput_pointer('%s' on  '%s')",
-			ic->identifier, device_id);
-	bool changed = false;
+void sway_input_configure_libinput_device(struct sway_input_device *input_device) {
+	struct input_config *ic = input_device_get_config(input_device);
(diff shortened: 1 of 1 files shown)


unresolvedsymbol:
@RedSoxFan Ah, thank you! Just compiled and can confirm it works now us lefties can use `swaymsg -- input type:tablet_tool calibration_matrix -1.0 0.0 1.0 0.0 -1.0 1.0`!

PR #5663, opened 2020-09-06
Fix #5643: in sway's render.c, use floor for rounding

### LINKED ISSUE #5643: Right border renders incorrectly when left border is offscreen
- **Sway Version:** 1.5-6991ac8c (Aug 16 2020, branch 'master')

- **Debug Log:** https://gist.github.com/FreeFull/dacb4be28e01048b87b35a19fb878edb

- **Configuration File:** Default sway config.

- **Description:**
When a floating window's left border overlaps with the left edge of the screen or is entirely off the left edge, the right border's rendering will be off by one pixel to the left. I have only tested this with a single monitor. Dragging the top, right or bottom border of the window offscreen doesn't seem to make any of the other borders misrender.

dimitris-triantafyllidis:
Currently, various floating-point expressions involving the coordinates of borders, titlebars and content surfaces are directly assigned to integral values before being passed to lower-level functions, and so are rounded towards zero.

Some of these floating-point expressions can have fractional negative values, so the coordinates of the top-left corners of titlebars, borders and surfaces are rounded towards bottom-right, while the coordinates of the bottom-right corners are rounded towards top-left.

These glitches can be best observed with semi-transparent titlebars and borders. To reproduce:

1) Move a floating container to a position where the left border touches the left edge of the screen
2) Drag the container one pixel to the left
The content surface and the right border will move one pixel to the left while the left border, the titlebar/top border and the bottom border will not. The content surface will overlap with the left border.
3) Drag the container further to the left so that the content surface (which is one pixel inside the left border already) touches the left edge of the screen
4) Drag the container one pixel to the left
The right border will move one pixel towards the left while the content surface will not and will overlap with the right border
5) Perform the same steps with the top edge of the screen or both edges at the same time

I used floor to preserve the rounding for positive coordinates and fix it for the negative ones. I could not observe any undesirable side-effects.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -261,7 +261,7 @@ static void render_view_toplevels(struct sway_view *view,
 		output->lx - view->geometry.x;
 	double oy = view->container->surface_y -
 		output->ly - view->geometry.y;
-	output_surface_for_each_surface(output, view->surface, ox, oy,
+	output_surface_for_each_surface(output, view->surface, floor(ox), floor(oy),
 			render_surface_iterator, &data);
 }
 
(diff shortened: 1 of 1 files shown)


dimitris-triantafyllidis:
I want to add that I noticed another single pixel glitch (within monitor bounds this time) with floating windows that this PR seems to have fixed.

I will look at it a bit better when I can and I might add one more commit to this branch.

TheAvidDev:
Hey, was just wondering why this PR was closed?

dimitris-triantafyllidis:
There is another issue I noticed that is not caused by this commit, yet I felt like closing this to first check whether they could be related.

PR #5664, opened 2020-09-06
Security v2

danieldg:
This is an alternative to the old security mechanism based on commands that was never merged.

This adds one command (security) which is used to allow or deny access to privileged interfaces on a client-by-client basis. If no security configuration it present, all privileged operations are allowed to all clients.
`security deny * *` will deny all clients access to all privileged operations
`security permit pid=12345 screencopy_manager` will allow all _existing_ connections created by PID 12345 access to the wlr_screencopy_manager interface.

PID and CID are accepted.  Because specifying PID is prone to attacks involving PID reuse, only existing connections from a given PID have their permissions modified.  To prevent PID reuse attacks, a process should only send commands addressed at its own PID or one of its children, and should wait for a reply from sway before `wait`ing on the child or exiting. If an application needs access immediately on startup, use a wrapper script that opens a socket, sends a `security permit` message directed at its own PID, and then passes the socket directly to the target program using `WAYLAND_SOCKET`.

Other possible methods to address sockets for security operations:

1. Send one end of a newly created socketpair over `SWAYSOCK` as a reply to `security create_socket`
2. Submit a socket via `SWAYSOCK` as part of a `security submit_socket` command.  This would an external program to create a secondary wayland server socket, launch programs with WAYLAND_DISPLAY=wayland-0-priv, and then pass any incoming sockets to sway with elevated permissions.
3. A two-step protocol where an initial command (`swaymsg -t get_clients`?) gets unique handles to existing client sockets, and the security command accepts this handle. This is significantly more complex on both sides.

Draft to solicit feedback; I will add real documentation if this is a valid path forward.

Example launch script: `perl -MIO::Socket::UNIX -e '$^F = 10; $s = IO::Socket::UNIX->new(Peer=>"$ENV{XDG_RUNTIME_DIR}/$ENV{WAYLAND_DISPLAY}"); $ENV{WAYLAND_SOCKET} = fileno $s; system "swaymsg security permit pid=$$ screencopy_manager"; exec "grim"'`


emersion:
>To avoid attacks from PID reuse, only existing connections from a given PID have their permissions modified

This doesn't avoid the race: by the time Sway receives the IPC command, the old process may have exited and a new process might have re-used its PID.

danieldg:
True, but a process elevating its own pid's sockets would need to crash between sending the message and receiving the response. I changed the wording slightly to reflect this.

Since sway's IPC socket is a unix socket, it is possible to use FD passing to have sway create a socketpair with the requested permissions and send that socket to the requesting client. I didn't do this initially because the ipc_client struct is not visible outside ipc-server.c.

So other than the pid-reuse race, is this a reasonable security implementation?

danieldg:
This version is a bit hard to use, I have a better one.

PR #5665, opened 2020-09-06
Add an option to hide the cursor when typing

tamirzb:
Add an option for the `hide_cursor` command to hide the cursor when typing, i.e. whenever a key is pressed.

The format of this new option is `seat <name> hide_cursor keyboard enable|disable`. This is in addition to the existing format of this command for setting the timeout, which still works the same. Combining both to hide the cursor either when typing or after a some time of inactivity is also possible.

This is supposed to have a similar behavior to using [xbanish](https://github.com/jcs/xbanish) or `unclutter -keystroke` on X.

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -6,6 +6,7 @@
 #include <wlr/types/wlr_pointer_gestures_v1.h>
 #include <wlr/types/wlr_surface.h>
 #include "sway/input/seat.h"
+#include "config.h"
 
 #define SWAY_CURSOR_PRESSED_BUTTONS_CAP 32
 
@@ -68,6 +69,10 @@ struct sway_cursor {
 
 	struct wl_event_source *hide_source;
 	bool hidden;
+	// This field is just a cache of the field in seat_config in order to avoid
+	// costly seat_config lookups on every keypress. HIDE_WHEN_TYPING_DEFAULT
+	// indicates that there is no cached value.
+	enum seat_config_hide_cursor_when_typing hide_when_typing;
 
 	size_t pressed_button_count;
 };
@@ -94,6 +99,7 @@ void cursor_handle_activity(struct sway_cursor *cursor,
 		struct wlr_input_device *device);
 void cursor_unhide(struct sway_cursor *cursor);
 int cursor_get_timeout(struct sway_cursor *cursor);
+void cursor_notify_key_press(struct sway_cursor *cursor);
 
 void dispatch_cursor_button(struct sway_cursor *cursor,
 	struct wlr_input_device *device, uint32_t time_msec, uint32_t button,
--- a/sway/commands/seat/hide_cursor.c
+++ b/sway/commands/seat/hide_cursor.c
@@ -3,26 +3,48 @@
 #include "sway/commands.h"
 #include "sway/config.h"
 #include "sway/input/seat.h"
+#include "sway/input/cursor.h"
+#include "sway/server.h"
 #include "stringop.h"
+#include "util.h"
 
 struct cmd_results *seat_cmd_hide_cursor(int argc, char **argv) {
 	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "hide_cursor", EXPECTED_EQUAL_TO, 1))) {
+	if ((error = checkarg(argc, "hide_cursor", EXPECTED_AT_LEAST, 1))) {
 		return error;
 	}
-	if (!config->handler_context.seat_config) {
+	if ((error = checkarg(argc, "hide_cursor", EXPECTED_AT_MOST, 2))) {
+		return error;
+	}
+	struct seat_config *seat_config = config->handler_context.seat_config;
+	if (!seat_config) {
 		return cmd_results_new(CMD_FAILURE, "No seat defined");
 	}
 
-	char *end;
-	int timeout = strtol(argv[0], &end, 10);
-	if (*end) {
-		return cmd_results_new(CMD_INVALID, "Expected an integer timeout");
+	if (argc == 1) {
+		char *end;
+		int timeout = strtol(argv[0], &end, 10);
+		if (*end) {
+			return cmd_results_new(CMD_INVALID, "Expected an integer timeout");
+		}
+		if (timeout < 100 && timeout != 0) {
+			timeout = 100;
+		}
+		seat_config->hide_cursor_timeout = timeout;
+	} else {
+		if (strcmp(argv[0], "when-typing") != 0) {
+			return cmd_results_new(CMD_INVALID,
+				"Expected 'hide_cursor <timeout>|when-typing [enable|disable]'");
+		}
+		seat_config->hide_cursor_when_typing = parse_boolean(argv[1], true) ?
+			HIDE_WHEN_TYPING_ENABLE : HIDE_WHEN_TYPING_DISABLE;
+
+		// Invalidate all the caches for this config
+		struct sway_seat *seat = NULL;
+		wl_list_for_each(seat, &server.input->seats, link) {
+			seat->cursor->hide_when_typing = HIDE_WHEN_TYPING_DEFAULT;
+		}
 	}
-	if (timeout < 100 && timeout != 0) {
-		timeout = 100;
-	}
-	config->handler_context.seat_config->hide_cursor_timeout = timeout;
 
 	return cmd_results_new(CMD_SUCCESS, NULL);
 }
--- a/sway/sway-input.5.scd
+++ b/sway/sway-input.5.scd
@@ -229,11 +229,16 @@ correct seat.
 	Set this seat as the fallback seat. A fallback seat will attach any device
 	not explicitly attached to another seat (similar to a "default" seat).
 
-*seat* <name> hide_cursor <timeout>
-	Hides the cursor image after the specified _timeout_ (in milliseconds)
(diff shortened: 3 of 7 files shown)


emersion:
Ref https://github.com/swaywm/sway/pull/3967

tamirzb:
Oh, I didn't realize there was already a similar PR in the past. I tried to look but I guess I missed it.

Anyway, I'm not entirely sure why was that PR closed. I feel that this can be useful, same as xbanish is a useful tool on X11. Personally, I can say that after switching from i3 to sway, this is the only functionality I'm still missing in sway that I had in i3.

I understand the idea of not introducing too much complexity, so I think it should be mentioned that this PR is much less complex than #3967, as it only attempts to provide the core functionality of xbanish and nothing more.

RedSoxFan:
> I understand the idea of not introducing too much complexity, so I think it should be mentioned that this PR is much less complex than #3967, as it only attempts to provide the core functionality of xbanish and nothing more.

I agree that this is more simple. My PR reshowed the cursor after a period of inactivity and had to keep track of why the cursor was hidden.

I personally wouldn't be opposed to this. There are a few style issues if we do move forward with it though.

tamirzb:
So, what exactly would you say is needed at this point in order to get this merged? I'd be happy to fix any style issues and/or other issues you point out in this PR.

Geo25rey:
It would be nice if you can manually show and hide the cursor with a runtime command rather than a timeout being force upon us.

sway/commands/seat/hide_cursor.c
```diff
@@ -7,22 +7,41 @@
...
+		int timeout = strtol(argv[0], &end, 10);
+		if (*end) {
+			return cmd_results_new(CMD_INVALID, "Expected an integer timeout");
+		}
+		if (timeout < 100 && timeout != 0) {
+			timeout = 100;
+		}
+		seat_config->hide_cursor_timeout = timeout;
+	} else {
+		if (strncmp(argv[0], "keyboard", 9)) {
```
kennylevinsen, line 32: `strcmp(argv[0], "keyboard") != 0`

sway/commands/seat/hide_cursor.c
```diff
@@ -7,22 +7,41 @@
...
+			return cmd_results_new(CMD_INVALID,
+				"Expected 'hide_cursor <timeout>|keyboard [enable|disable]'");
+		}
+		if (!strncmp(argv[1], "enable", 7)) {
```
kennylevinsen, line 36: `strcmp(argv[1], "enable") == 0`

sway/commands/seat/hide_cursor.c
```diff
@@ -7,22 +7,41 @@
...
+			seat_config->hide_cursor_when_typing = true;
+		} else if (!strncmp(argv[1], "disable", 8)) {
```
kennylevinsen, line 38: same

sway/input/seat.c
```diff
@@ -73,6 +73,24 @@ static void seat_node_destroy(struct sway_seat_node *seat_node) {
...
+
+	const struct seat_config *seat_config = seat_get_config(seat);
+	if (!seat_config) {
+		seat_config = seat_get_config_by_name("*");
+		if (!seat_config) {
+			return;
+		}
+	}
+
+	if (seat_config->hide_cursor_when_typing == true) {
```
kennylevinsen, line 89: Don't explicitly compare with `true`.

sway/input/seat.c
```diff
@@ -73,6 +73,24 @@ static void seat_node_destroy(struct sway_seat_node *seat_node) {
 	free(seat_node);
 }
 
+void seat_notify_key_press(struct sway_seat *seat) {
+	if (seat->cursor->hidden) {
+		return;
+	}
+
+	const struct seat_config *seat_config = seat_get_config(seat);
```
kennylevinsen, line 81: Looking up the config for every keypress seems excessive, especially as it happens twice in the worst case where the config is for `*` and hide when typing isn't enabled,

Cache the value on the seat or the cursor, so the function just becomes something like:
```
if (!seat->cursor->hidden && seat->cursor->hide_when_typing) {
    cursor_hide(seat->cursor);
}
```

sway/sway-input.5.scd
```diff
@@ -218,11 +218,16 @@ correct seat.
...
-	than that (aside from 0), will be increased to 100.
+*seat* <name> hide_cursor <timeout>|keyboard [enable|disable]
+	Hides the cursor image after the specified event occured.
+
+	If _timeout_ is specified, then the cursor will be hidden after _timeout_
+	(in milliseconds) has elapsed with no activity on the cursor. A timeout of 0
+	(default) disables hiding the cursor. The minimal timeout is 100 and any
+	value less than that (aside from 0), will be increased to 100.
+
+	If _keyboard_ is enabled, then the cursor will be hidden when typing
```
kennylevinsen, line 229: What if I want it to hide when typing *or* after a timeout? I wonder if that should be supported.

kennylevinsen (changes requested):
Left a few review comments.

sway/sway-input.5.scd
tamirzb: It is supported. The following configuration would cause the cursor to be hidden either when typing or after a 5 seconds timeout:
```
seat * hide_cursor keyboard enable
seat * hide_cursor 5000
```

sway/commands/seat/hide_cursor.c
tamirzb: Actually now that I'm looking at the code again, I see that there is a `parse_boolean` function in `util.h` which I think fits here perfectly. I will replace the checks here and for "disable" with that function.

sway/commands/seat/hide_cursor.c
tamirzb: Ok, will change

sway/input/seat.c
tamirzb: The thing is, `hide_cursor_when_typing` actually has three possible value: `true`, `false` and `-1` which stands for default/unset (every field in `seat_config` supports this in order to support merging configs). Therefore, if I remove the explicit comparison with `true` it means that a value of `-1` will also pass the `if` check, which is something I don't want.
I was thinking about making this an enum, but I saw that `fallback` is another field in `seat_config` which is kind of a boolean and basically has these three possible states, so I kinda copied from it.

Perhaps it would make sense to add a comment here explaining why this is an explicit comparison with `true`? Or maybe still use an enum even though it would be different from the `fallback` field? What do you think?

tamirzb: The thing is, it actually has three possible values: `true`, `false`, and `-1` (default). Now that I'm looking at the code again I see that some others fields in `seat_config` that have similar three values use enums, so I'll use an enum here as well.

sway/input/seat.c
tamirzb: I agree that it's excessive. The issue of course with having a cache is that you need to invalidate it whenever the configuration changes. Right now I'm thinking to just go over all the seats and invalidate their cache whenever a `hide_cursor_when_typing` configuration changes. Not the prettiest solution but I'm not sure there is a better one.

tamirzb:
@kennylevinsen Ok so I fixed all your comments, let me know if you have anything else to add :)

kennylevinsen:
ping @emersion

include/sway/input/cursor.h
```diff
@@ -68,6 +69,10 @@ struct sway_cursor {
 
 	struct wl_event_source *hide_source;
 	bool hidden;
+    // This field is just a cache of the field in seat_config in order to avoid
```
emersion, line 72: Style: we use tabs

sway/commands/seat/hide_cursor.c
```diff
@@ -3,26 +3,48 @@
...
+			return cmd_results_new(CMD_INVALID, "Expected an integer timeout");
+		}
+		if (timeout < 100 && timeout != 0) {
+			timeout = 100;
+		}
+		seat_config->hide_cursor_timeout = timeout;
+	} else {
+		if (strcmp(argv[0], "keyboard") != 0) {
+			return cmd_results_new(CMD_INVALID,
+				"Expected 'hide_cursor <timeout>|keyboard [enable|disable]'");
```
emersion, line 37: Bikeshed: maybe we could rename the command to `hide_cursor when-typing enable|disable`?

sway/input/seat.c
```diff
@@ -73,6 +73,32 @@ static void seat_node_destroy(struct sway_seat_node *seat_node) {
 	free(seat_node);
 }
 
+void seat_notify_key_press(struct sway_seat *seat) {
```
emersion, line 76: Hm, I think I'd prefer to leave this in `cursor.c`, alongside the current hide-cursor-after-timeout implementation.

emersion: (This function can take a `sway_cursor` instead of a `sway_seat` I believe.)

emersion (changes requested):
Overall looks good, just a few minor comments

sway/commands/seat/hide_cursor.c
tamirzb: Done

include/sway/input/cursor.h
tamirzb: Oops, changed

sway/input/seat.c
tamirzb: Done

emersion (approved):
LGTM, thanks!

PR #5669, opened 2020-09-09
Allow running as root with an opt-in CLI flag

yujiri8:


emersion:
Why?

yujiri8:
Because it can be very inconvenient for this to be absolutely disallowed. I understand the reasons not to run a window manager as root, but I don't see why it should be made actually impossible without editing the source.

Geo25rey:
I can see making this a compile time option, but it's a major security concern if it's a runtime option.

yujiri8:
If it requires recompiling, one might as well just edit the source.

Allowing users to opt-*in* to insecure things should not be considered a security failure; it's not software's responsibility to protect users from their own intentional choices.

TheAvidDev:
Just a note, I don't think this has a chance unless you provide a sufficiently good reason.

> Because it can be very inconvenient for this to be absolutely disallowed.

What is this situation where you are running a window manager as root?

yujiri8:
I don't think it's reasonable to try to guess all possible uses of a piece of software and artificially prevent something just because it seems like a bad idea.

Personally, I run things normally as root because for the amount of tinkering I do, having to constantly switch accounts or use `sudo` is very cumbersome. (I built from canonical source and didn't run as root for the other issue I opened.)

I did not expect this to be accepted; I'm only waiting for a final no before I fork.

Geo25rey:
@yujiri8 It would make sense to have the daily driver version for the average software user which has root access disabled by default at compile time. And if you are a tinkerer who requires root permissions for debugging or something, you could compile your tinkerer's version with root access which you might not run most of the time unless you're tinkering. 

It's very important that this feature is a compile time feature rather than a runtime feature since it may allow for easier input logging or potential hardware damage. I do agree though that Sway shouldn't specifically limit the users even for security reasons as long as the user knows what they're doing, i. e. specifically opting in to a security vulnerability.

danieldg:
The patch as-is does not correctly handle the case where sway itself is setuid root.  Just disabling the "setuid(0)!=-1" double-check part of drop_permissions would be better.

You could also require an environment variable checked using `secure_getenv` instead of a command line argument (as long as sway is fine with using that maybe-glibc-specific function): that will return NULL if there is any kind of permission elevation involved in running sway, not just suid bits.

Other possible uses for this: running sway in a container that maps an unprivileged user to uid=0 (this mapping can be required if your system and/or user doesn't have subuids configured and you need to do anything involving capabilities inside the container).

yujiri8:
Thanks for pointing out the missed case. I have updated the commit so that it still drops root if a non-root user runs it with the flag and setuid.

kennylevinsen:
> Personally, I run things normally as root because for the amount of tinkering I do, having to constantly switch accounts or use `sudo` is very cumbersome. (I built from canonical source and didn't run as root for the other issue I opened.)

Note: If you start a terminal as root (setting WAYLAND_DISPLAY correctly and all that), then you should be able to freely start whatever you want from there without needing to sudo all the time, nor needing to switch over to a root sway instance.

yujiri8:
Interesting. I can get Fractal and `eog` to work that way, but not other apps (various errors about Qt plugins and X displays).

L-as:
A reason for why this would be useful:
When I run programs in a new user namespace with bubblewrap I often set uid and gid to 0, and sway works in this new namespace if I patch the source. IMHO using users to manage privileges and security is an outdated approach (which I don't think is even used much in practice), and because of that sway shouldn't assume that running as root is "insecure".

kennylevinsen:
Running as uid/gid 0 in a container is not secure. See LXC's rationale/documentation for unprivileged containers for more information.

Running as root is always unsafe, so the question here is whether to allow unsafe use. In this case, it's more a matter of it conflicting with the direct session backend's handling of SUID.

danieldg:
> Running as uid/gid 0 in a container is not secure. See LXC's rationale/documentation for unprivileged containers for more information.

It is possible to set up an unprivileged container that runs as non-root but which appears to be running as root from *inside* the container - try running `unshare --user --map-root-user id` for a simple example.  Running sway inside such a container is one possible case that is no less secure than running as non-root outside the container but is denied by the current code.  (I suspect that sway might have problems accessing DRM backends if you do this - but wayland or X11 backends should work fine).

> And in this case, it's also a matter of it conflicting with the direct session backend's handling of SUID.

It is perfectly possible to have sway be secure when running as suid and still be able to run as root; that's what the current version of the patch does.  Requiring an explicit argument even defends against a theoretical badly written login manager that can be convinced to forget to drop root but which still limits the available commands to (say) display manager .desktop files.

Note that without this patch, you can still run sway as root: you can either patch it directly or just load a helper (say, LD_PRELOAD) that fakes the returns of setuid/setgid as needed.  While that technique is usually used to fake being root, it's just as easy to fake not being root.

kennylevinsen:
> It is possible to set up an unprivileged container that runs as non-root but which appears to be running as root from inside the container

Yes, that's user privilege separation as the UID is an arbitrarily high number through UID re-mapping. User privilege separation was incorrectly stated to be an "outdated" mechanism for handling privileges.

If they wanted to run an unprivileged container, then I misread.

> It is perfectly possible to have sway be secure when running as suid and still be able to run as root;

I never said it wasn't possible. I said that running as root (that is, real uid 0) is *unsafe*, and that it conflicts with the direct backend's handling of SUID. The current block being in place because it signals a faulty attempt at using the direct backend.

We would need one of the following:
1. To remove error handling around the privilege drop code for direct usage.
2. Have a SWAY_UNSAFE_YES_I_REALLY_WANT_TO_RUN_AS_ROOT=1 flag to disable privilege drop.
3. A dedicated *truly* direct session backend that is never picked automatically that, if selected, signals privileges shouldn't be dropped.

Personally, I don't really care about protecting people from shooting themselves in the feet as long as it's their own active decision and doesn't end up being my problem.

I do, however, find the overlap with SUID-case of running sway as root and then dropping to a user to be a possible problem area, with innocent users wanting to use the direct backend possibly ending up with a fully-root sway session by running it with sudo (like the good old days with privileged X servers).

emersion:
Superseded by #7044

Scorpion2185:
Is this option going to be provided?

bl4ckb0ne:
Not likely.

kennylevinsen:
sway shouldn't complain about being run as "true" root anymore, so such a flag is unnecessary.

Scorpion2185:
It just did on Debian Bookworm: version 1.7-3.
How can you fix it?

@yujiri8 are you going to fork it?

Scorpion2185:
@kennylevinsen  It just did on Debian Bookworm: version 1.7-3.
How can you fix it?

@yujiri8 are you going to fork it?

yujiri8:
I have lost interest in the issue since I don't run as root anymore, which is only because I found it too inconvenient to fight against so many pieces of software trying to prevent it, all very misguided.

yujiri8:
> @kennylevinsen It just did on Debian Bookworm: version 1.7-3.

Looks to me like that's because the change from #7044 hasn't been released yet?

kennylevinsen:
Indeed, 1.7 is from January while the #7044 is from June.

> I found it too inconvenient to fight against so many pieces of software trying to prevent it, all very misguided.

We still very much do not recommend running a user session or your compositor as root.

... However, the previous behavior was not meant to raise a finger at users wanting to disregard our advice, but rather to avoid simple but very critical yet simple mistakes during use of the SUID feature that normal users relied on. Flexibility vs. footguns.

With the advent of libseat and seatd we could drop the SUID feature and thus this safeguard[^1]. So, you do you. Heck, run passwordless root telnet open to the public internet if you want - just don't try blame us for what follows. :)

[^1]: Do note that we will error out if SUID operation is detected during a transitional period, so that SUID users that failed to read the release notes across multiple releases do not end up becoming root by surprise. Principle of least surprise and all that.

Scorpion2185:
> Indeed, 1.7 is from January while the #7044 is from June.

Could you make a new (minor) release for this?
For an easier update, for example for `uscan`.

kennylevinsen:
> Could you make a new (minor) release for this?

A minor release would be 1.8, and we do not make such a release just for a single change. This is also a breaking change, so it cannot qualify for a patch release as it is not critically necessary.

The next wlroots release is slowly materializing, and I imagine sway will follow not too long after.

Issue #5066, opened 2020-03-03
Simple clients lose their borders when floated


lephe:
Simple (not to say trivial) Wayland clients seem to lose their border configuration at the moment they go floating. I first noticed this with a hand-written program then with emersion's [`hello_wayland`](https://github.com/emersion/hello-wayland).

Here's an example with the cat. Before floating, the window looks like this (notice the 1 pixel border around the right half of the screen):

![out-1](https://user-images.githubusercontent.com/33732078/75812704-f5468480-5d8e-11ea-8933-6ffa69cbcddd.png)

After floating and tiling back, the border is gone, the `border` attribute is also gone, and the corresponding space has been redistributed to the client:

![out-2](https://user-images.githubusercontent.com/33732078/75812711-f972a200-5d8e-11ea-8fcf-7f056fa9fcbb.png)

The border disappears immediately when floating. I have only observed this problem with such simple programs. Gtk/Qt applications and the like all keep their borders.

Sway version:
```
sway version 1.4-476773dd (Mar  3 2020, branch 'master')
```
Minimal configuration with which I could manipulate my keyboard to reproduce the bug:
```
output * bg #000000 solid_color

bindsym F12 exec alacritty
bindsym Shift+Mod1+C reload
bindsym Mod1+Space floating toggle

bindsym F1 workspace 1
bindsym F2 workspace 2
```
Steps to reproduce:
1. Compile and execute emersion's [`hello_wayland`](https://github.com/emersion/hello-wayland)
2. Make the window float

Also, obligatory big thank you for the project and the work that goes into it. ^^

somini:
Can confirm this happens with alacritty configured for `decorations: none` (to show only sway decorations).

patcoll:
This seems to happen to me using kitty, whether I have `hide_window_decorations` set to either yes or no. Haven't been able to nail down a simple reproducible case for myself yet. Using `sway-launcher-desktop` and looking to have a border on a floating window.

dimitris-triantafyllidis:
I just looked a bit into this and it seems to have something to do with CSD and some logic around view->using_csd. If you print the tree when the window is floating, is the border set to "csd"? I plan to try and fix it one of these days along with a couple of other border related things.

lephe:
> If you print the tree when the window is floating, is the border set to "csd"?

Yes, the border is set to `csd` while the window in my initial example is floating. Well spotted!

emersion:
[...]

PR #5670, opened 2020-09-09
Issue #5066: Fix saved_border initialization in view_map

dimitris-triantafyllidis:
In view_map, the function view_update_csd_from_client is called conditionally before the container's border types are initialized from the configuration. This results in saved_border being initialized inside view_update_csd_from_client from garbage values.

This results in applications that take this path not having their borders restored when set to floating and then back to tiled.

Moved this conditional call just after the next block that initializes the containers borders.

In my system this path is taken by most of the GTK applications I use, including gnome-terminal, nautilus, totem, gnome-calculator and others, vscode, and emersions's hello_wayland example application.

To reproduce the bug, open an application and verify from sway's debug log that the message
`View 0xSOMETHING updated CSD to 1` appears in sway.log.
Then toggle the view to floating and back to tiled. The borders will be set to none.
While floated the borders will be set to csd and saved_border will be none.
You need to not have explicitly set the window's borders from the keyboard
or with criteria.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -743,10 +743,6 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 		&view->surface_new_subsurface);
 	view->surface_new_subsurface.notify = view_handle_surface_new_subsurface;
 
-	if (decoration) {
-		view_update_csd_from_client(view, decoration);
-	}
-
 	if (view->impl->wants_floating && view->impl->wants_floating(view)) {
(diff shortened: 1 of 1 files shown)


RedSoxFan:
Related #4287. Maybe there are issues with both orderings and requires a larger change?

dimitris-triantafyllidis:
You are right, I am closing this PR! There are a couple of other related spots I intended to check anyway.

PR #5672, opened 2020-09-10
CONTRIBUTING.md: Add note about sizeof style

RedSoxFan:
See https://github.com/swaywm/sway/pull/5472#discussion_r458520053

Throughout the code base, sizeof is used with the struct name. To
maintain consistency, this should continue being used. It does however
differ from the kernel style guide and should be mentioned.


emersion:
I'm fine with either style (just like we allow both `if (asdf)` and `if (asdf != NULL)` I guess).

PR #5583, opened 2020-07-28
Allow to disable insecure protocols at compile-time

ya-isakov:
This pull-request is a way to allow build sway without insecure ptotocols, such as text-input (which allow to keylog user input from other windows) or screencopy (which allows any application to capture other apps' windows)
This is just a temporary solution, until proper security model is in place. Also, everything is enabled by default (like it was before this PR)

PR #5675, opened 2020-09-12
Client security labels

danieldg:
This is a rework of https://github.com/swaywm/sway/pull/5664 because I found that a bit awkward to use in practice.

Instead of specifying the interfaces each command needs, create security labels for various types of clients. A simple configuration might have two labels (default and privileged), while a minimum-privilege system could specify a distinct label for each type of program.  If no configuration commands are present, all clients are allowed access to all interfaces.

All wayland clients normally get the "default" label. There are two ways to specify a different label for a specific socket: the `exec --use-wayland-socket` command starts a process with `WAYLAND_SOCKET` set to a wayland client FD, and `sandbox_socket` creates another listening socket that accepts new clients.  Both of these commands accept `--label <name>` to set the label of the client or clients they create. The `sway{nag,bar,bg}_command` commands also accept `--label` arguments.


danieldg:
The build failure on FreeBSD is due to meson.build "add_project_arguments('-D_C11_SOURCE', language: 'c')" under if_freebsd. This causes `__BSD_VISIBLE` to be false, which hides the `CMSG_SIZE` macro.  This fact is not well-documented in the freebsd manpage for cmsg.

I could remove the offending line in meson.build, but 3e924f23451a3bcbad68d5250234e1eb1a987855 seems to imply that it might cause problems (unfortunately it does not say what problems that would be). I could also `#undef _C11_SOURCE` at the top of the file. Any preference?

emersion:
Have you read https://gitlab.freedesktop.org/wayland/weston/-/issues/206?

danieldg:
I read it a while ago, but didn't refer to it when creating this.  After re-reading, I think switching to socket submission via the wayland interfaces talked about there would be better than the `submit_socket` patch, so drop that one.  The changes to exec and the basic labelling seems to be compatible with the static security policies in that discussion.

sway/commands/sandbox_socket.c
```diff
@@ -0,0 +1,119 @@
...
+		while (i < argc) {
+			if (strcmp(argv[i], "--label") == 0) {
+				if (i + 1 >= argc)
+					return cmd_results_new(CMD_INVALID, "--label requires an argument");
+				label = argv[i + 1];
+				i += 2;
+			} else {
+				return cmd_results_new(CMD_INVALID, "Unknown argument to sandbox_socket");
+			}
+		}
```
ony, line 73: This is a bit unexpected ordering. I think options usually go before positional arguments. Just to consider:

```suggestion
		int i = 1;
		while (i < argc) {
			if (strcmp(argv[i], "--label") == 0) {
				if (i + 1 >= argc)
					return cmd_results_new(CMD_INVALID, "--label requires an argument");
				label = argv[i + 1];
				i += 2;
			} else if (strcmp(argv[i], "--") == 0) {  // after this any argument should be treated as positional
				++i;
				break;
			} else if (strncmp(argv[i], "-", 1) == 0) {
				return cmd_results_new(CMD_INVALID, "Unknown option to sandbox_socket");
			} else {
				break;  // end of options, now only positional arguments
			}
		}
		char *path = argv[i++];
		if (!path) {
				return cmd_results_new(CMD_INVALID, "Missing mandatory argument \"path\" to sandbox_socket");
		} else if (argv[i]) {
				return cmd_results_new(CMD_INVALID, "Unknown argument to sandbox_socket");
		}
```

sway/commands/sandbox_socket.c
```diff
@@ -0,0 +1,119 @@
...
+
+		if (path_len > sizeof(name.sun_path)) {
+			return cmd_results_new(CMD_INVALID, "Invalid socket path: %s", path);
+		}
+
+		name.sun_family = AF_UNIX;
+		memcpy(name.sun_path, path, path_len);
+		size_t name_len = sizeof(name) - sizeof(name.sun_path) + path_len;
```
ony, line 81: I would be careful with expressions like `sizeof(x) - sizeof(x.a)` - they look weird. Even if `sun_path` is last field in struct, there is no guarantee there will be no padding for alignment purposes that will ruin this calculation. 

What is suggested in [unix(7) man](https://www.man7.org/linux/man-pages/man7/unix.7.html):

```suggestion
		size_t name_len = offsetof(struct sockaddr_un, sun_path) + path_len;
```

sway/commands/sandbox_socket.c
```diff
@@ -0,0 +1,119 @@
...
+};
+
+static list_t *sandbox_sockets;
+
+static int fd_accept(int srv_fd, uint32_t mask, void *data) {
+	struct sandbox_socket *sock = data;
+	struct sockaddr_un name;
+	socklen_t slen = sizeof(name);
+
+	int cli_fd = accept(srv_fd, (struct sockaddr *)&name, &slen);
```
ony, line 32: It doesn't look like you use `name` after `accept`. I think you can simply drop it.

```suggestion
	int cli_fd = accept(srv_fd, NULL, NULL);
```

sway/commands/sandbox_socket.c
```diff
@@ -0,0 +1,119 @@
...
+	} else if (strcmp(op, "delete") == 0) {
+		int i;
+		for(i = 0; i < sandbox_sockets->length; ++i) {
+			struct sandbox_socket *sock = sandbox_sockets->items[i];
+			if (strcmp(sock->path, path) == 0) {
+				wl_event_source_remove(sock->src);
+				unlink(path);
+				free(sock->path);
+				free(sock);
+				list_del(sandbox_sockets, i);
```
ony, line 111: I might be wrong, but I think none of these will actually close service socket.
```suggestion
				wl_event_source_remove(sock->src);
				unlink(path);
				close(sock->fd);
				free(sock->path);
				free(sock);
				list_del(sandbox_sockets, i);
```

ony (changes requested):

sway/commands/sandbox_socket.c
danieldg: At least in my copy of wayland, `wl_event_source_remove` does close the fd as long as `wl_event_source->fd >= 0`.

sway/commands/sandbox_socket.c
danieldg: yeah, this is a saner ordering, not sure why I didn't do that in the first place

sway/commands/sandbox_socket.c
danieldg: I think for some reason I thought offsetof wasn't available here

sway/commands/security_label.c
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	for(int i = 2; i < argc; i++) {
+		char* arg = argv[i];
+		if (strcmp(arg, "*") == 0)
+			privs |= ~0;
+		else if (strcmp(arg, "layer_shell") == 0)
+			privs |= 1 << PRIV_LAYER_SHELL;
+		else if (strcmp(arg, "output_manager") == 0)
+			privs |= 1 << PRIV_OUTPUT_MANAGER;
```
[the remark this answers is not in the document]
danieldg: If done this way, that table probably belongs in security.c and exposed via the header, so there's only one place to update when new permission bits are added.

sway/security.c
```diff
@@ -0,0 +1,218 @@
...
+
+uint32_t security_get_label_id(const char* name)
+{
+	size_t i;
+	if (!security_labels) {
+		security_label_capacity = 8;
+		security_label_len = 1;
+		security_labels = calloc(sizeof(security_labels[0]), security_label_capacity);
+		security_labels[0].allowed = ~0;
+		security_labels[0].name = "default";
```
ony, line 201: Do we really need "default"? I don't think we implicitly associate everything that is not explicitly labeled with "default" label.

ony (commented):
I'm mostly biased here since I'm interested in ability of sway to serve on multiple sockets (e.g. `sandbox_socket` command).

I would consider splitting this PR in:
1. Creating new listen sockets.  
This might be extended further to be able to tag clients without relying on window titles, pids, etc. My use-case is to place clients in pre-defined layout.
2. Security labels and making them effective.
3. Ability to execute commands with ad-hoc socket with security labels applied.

I expect this to increase probability of individual parts to be merged :)

sway/commands/exec_always.c
```diff
@@ -26,14 +30,39 @@ struct cmd_results *cmd_exec_validate(int argc, char **argv) {
...
+			seclabel = argv[skip_argc];
+			skip_argc++;
+		} else if (strcmp(argv[skip_argc], "--") == 0) {
+			skip_argc++;
+			break;
+		} else {
+			return cmd_results_new(CMD_INVALID, "Unknown switch %s", argv[skip_argc]);
 		}
 	}
+	if ((error = checkarg(argc - skip_argc, argv[-1], EXPECTED_AT_LEAST, 1))) {
```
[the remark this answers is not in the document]
danieldg: The `argv[-1]` was already there, and acts like `argv[0]` in a C program: it should contain the correct choice "exec" or "exec_always".  If this isn't true, you should probably submit this as a fix to sway itself.

sway/commands/exec_always.c
```diff
@@ -26,14 +30,39 @@ struct cmd_results *cmd_exec_validate(int argc, char **argv) {
...
+		return error;
+	}
+	argc -= skip_argc;
+	argv += skip_argc;
+
+	if (seclabel && !use_wl_socket)
+		return cmd_results_new(CMD_INVALID, "--label requires --use-wayland-socket");
```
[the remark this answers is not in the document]
danieldg: I was considering adding `--use-wayland-socket` as an independent option to this patch, but didn't split them like that.

I do think that it is needed, though having `--label` imply `--use-wayland-socket` is probably nicer to users writing the config.

sway/commands/exec_always.c
```diff
@@ -94,6 +151,11 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "Second fork() failed");
 	}
 
+	if (client && seclabel) {
+		uint32_t label_id = security_get_label_id(seclabel);
+		security_set_client_label_id(client, label_id);
+	}
```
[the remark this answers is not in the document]
danieldg: Since `security_get_label` allocates a label, I wanted it as late as possible.

sway/security.c
danieldg: We need some way to set the permissions of clients that don't get a label from somewhere else. This lets you do things like disable protocols you don't want to use at all (as discussed in #5583).

danieldg:
@ony -
> I would consider splitting this PR in:
>     1. Creating new listen sockets.
>        This might be extended further to be able to tag clients without relying on window titles, pids, etc. My use-case is to place clients in pre-defined layout.

You should be able to cherry-pick the sandbox_socket commit and remove the security_ function calls as a starting point on your idea.  Maybe rename the command to `alt_socket`? The ability to have additional listening sockets isn't very useful alone, but if you were already planning to add a socket_id or similar match-item it would make sense to merge.  Adding label support to `alt_socket` would then become a part of this series like it does to `exec`.

I've sometimes wanted the ability to apply a mark to multiple windows; automatically applying a mark to all clients coming in on an alt_socket seems like a reasonable way to do as you describe.  Might need to come up with a different name to avoid confusing it with the existing marks. And then you get to fight about if this feature needs to get added to i3 first :).

danieldg:
> > If no configuration commands are present, all clients are allowed access to all interfaces.
> 
> This doesn't seem like a good default security configuration. Clipboard, screen, and keyboard input may contain sensitive financial information(e.g. cryptocurrency seed and password) or otherwise sensitive personal data.

This default is needed so that existing setups continue to work; it's how sway works today, and I don't think it's useful to merge a security setup that breaks everything for no good reason. I would expect anyone who cares about security would restrict clients by either labelling them or changing the default permissions and labelling the trusted programs that do need such access.  It may be useful to change the default config to include security labels.

Note that this mechanism alone doesn't make things secure - you generally need to combine it with one or more of sandboxing (flatpak, docker, etc), security modules (selinux), or running things as distinct UIDs (android) to avoid other methods of grabbing the secrets in question.

crocket:
If I can change the defaults through user configuration, I think it's okay.

By the way, if you want to do application-level sandboxing, look into linux namespaces and firejail which combines linux namespaces and seccomp-bpf. Seccom-bpf can block unused system calls.

danieldg:
> If I can change the defaults through user configuration, I think it's okay.

It does.

> By the way, if you want to do application-level sandboxing, look into linux namespaces and firejail which combines linux namespaces and seccomp-bpf. Seccom-bpf can restrict system calls.

Yes, I'm aware of most of the security mechanisms available there; namespaces are the foundation of the sandboxing mechanisms I mentioned.  I haven't found restricting system calls on arbitrary applications to be very useful unless it's already inside a good sandbox, because everyone needs read/write/open, and that's already enough to cause havoc if you're malicious. The best place to use seccomp is as part of the application (web browser rendering processes, for example).

For sway and this branch in particular, access to either the sway IPC socket or config file will let you bypass security.  I don't think that's worth changing, because both of those are adjacent to other important sockets (dbus) and config files you already need to protect.

crocket:
@danieldg I don't know where else to contact you. So, I am going to ask you an off-topic question here. What is the best linux sandboxing method that's not too inconvenient?

crocket:
I wish there was explanation for each security privilege so that users can make informed choices. It can be hard to know how a certain protocol is going to threaten security.

danieldg:
> I wish there was explanation for each security privilege so that users can make informed choices. It can be hard to know how a certain protocol is going to threaten security.

This information should probably be included in the documentation for this feature. I don't like writing documentation until a design is finalized, so that isn't done yet.  I think @emersion had a few "why" explanations for some of the less-obvious protocols in https://github.com/swaywm/sway/pull/5583's comments.  Documenting which ones can only be used to "break" the session (set gamma so everything is black, cover the screen, etc) versus those that can actually be used to take over would be a start there.

> @danieldg I don't know where else to contact you. So, I am going to ask you an off-topic question here. What is the best linux sandboxing method that's not too inconvenient?

The least inconvenient way is to build off someone else's sandbox, but then you have to agree with what they decide to allow. I think I override or change something in almost every sandbox I use for real; I currently use flatpak for most of my desktop sandboxing, but I haven't really done a wide survey.  I'm generally on freenode IRC (including sway's channel).

emersion:
FWIW, somehow related is https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/68.

emersion:
I believe this is largely superseded by the security-context-v1 protocol. Please re-open if not.

PR #5678, opened 2020-09-13
Make focus_follows_mouse work when hovering a layer-shell surface on …

### LINKED ISSUE #5668: Output is not focused when cursor is over status bar
- **Sway Version:**
sway version 1.5-7ca9ef12 (Sep  5 2020, branch 'master')

- **Description:**
I'm using sway with 2 monitors positioned directly above each other and waybar positioned at the top of the screen. When moving the cursor from the top one to the bottom one, the output becomes focused only if I move the cursor beyond waybar (either on a window or an empty workspace). The same happens when using swaybar so it's not something related to waybar.

The log was taken by starting sway, then moving the cursor down from top to bottom and then closing sway, not sure whether it's useful though.

[sway_debug.log](https://github.com/swaywm/sway/files/5194357/sway_debug.log)

David96:
…another output

Fixes #5668

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -7,7 +7,9 @@
 #include "sway/input/cursor.h"
 #include "sway/input/seat.h"
 #include "sway/input/tablet.h"
+#include "sway/output.h"
 #include "sway/tree/view.h"
+#include "sway/tree/workspace.h"
 #include "log.h"
 #if HAVE_XWAYLAND
 #include "sway/xwayland.h"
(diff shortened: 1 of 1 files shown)


David96:
Just squashed a null check commit.
Been running this since opening the pull request and haven't noticed any problems - is there anything missing for this to get merged?

I'd be happy about a shorter commit name, but couldn't think of any that nicely describes what the commit does.

Xyene (approved):
Looks good, thanks!

PR #5681, opened 2020-09-14
input/cursor: reset event source after unhide

Emantor:
Reset the event source after unhiding the cursor, to ensure that the
timeout starts after showing the cursor. Also remove the open coded
variant in seat_consider_warp_to_focus().

Fixes #5679

PR #5682, opened 2020-09-14
input/pointer: update cursor activity after updating button counts

### LINKED ISSUE #5679: hide_cursor stops countdown after left clicking once (but not moving the cursor)
- **Sway Version:**
sway version v1.5

- **Debug Log:**
[sway.log](https://gist.github.com/antisocrates/118a0a106e6be87fde9d000541ac144e)

- **Configuration File:**
[config](https://gist.github.com/antisocrates/0595d8908fee19dab3b38d6325497ca7)

- **Description:**
  Using `seat * hide_cursor 3000` in sway config to hide the mouse cursor after 3 seconds of inactivity.
The countdown (towards timeout to hide the cursor) stops after clicking anywhere indefinately (until the cursor moves again)

  _Steps to reproduce:_

 1. Left click anywhere
 2. Dont move the cursor (stay on the same spot you clicked, otherwise countdown resets and hides as normal)

     **Here is a 60second demonstration of the issue i have uploaded:**
     [Video Demonstration](https://streamable.com/03ejjf)

**_Clarification:_**
_Clicking will stop the countdown, showing the cursor, but won't restart it so the cursor will remain visible indefinitely until moved._

Xyene:
Otherwise, Sway will not re-hide a cursor after the last button has been
released.

Needed alongside afa890e to fix #5679.

/cc @Emantor -- <del>I had tested #5681 by scrolling, which worked correctly
after that patch.</del> Looking a bit deeper into this, this was regressed by 82c439c4f116b36d7d171bcc923b50ead59b4ab2 ([diff](https://github.com/swaywm/sway/commit/82c439c4f116b36d7d171bcc923b50ead59b4ab2#diff-40846134db734aabaf9eb9e98bccab5eL358-L375)),
so just reverting that bit of the patch (which is what this PR does) fixes the issue
by itself.

I think afa890e8e9f10667a56e896a114bf81fbc3ff54a still makes sense, though.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -378,7 +378,6 @@ void dispatch_cursor_button(struct sway_cursor *cursor,
 static void handle_pointer_button(struct wl_listener *listener, void *data) {
 	struct sway_cursor *cursor = wl_container_of(listener, cursor, button);
 	struct wlr_event_pointer_button *event = data;
-	cursor_handle_activity(cursor, event->device);
 
 	if (event->state == WLR_BUTTON_PRESSED) {
 		cursor->pressed_button_count++;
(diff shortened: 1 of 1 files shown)


RedSoxFan (approved):

RedSoxFan:
Thanks!

PR #5683, opened 2020-09-15
Fix minor typos in German README

Vetm:


emersion:
cc @ParadoxSpiral: does that sound right to you?

Emantor:
As a german native speaker, this looks good to me.

emersion:
Thank you both!

Issue #5687, opened 2020-09-15
Stop binding to wayland-0


emersion:
Port the Weston MR to stop binding to wayland-0: https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/486

PR #5688, opened 2020-09-15
server: Avoid using "wayland-0" as WAYLAND_DISPLAY

Geo25rey:
Closes: #5687

See https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/486

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -1,7 +1,8 @@
-#define _POSIX_C_SOURCE 200112L
+#define _POSIX_C_SOURCE 200809L
 #include <assert.h>
 #include <stdbool.h>
 #include <stdlib.h>
+#include <string.h>
 #include <wayland-server-core.h>
 #include <wlr/backend.h>
 #include <wlr/backend/headless.h>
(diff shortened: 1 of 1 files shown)


sway/server.c
```diff
@@ -151,13 +152,30 @@ bool server_init(struct sway_server *server) {
...
+		}
+	}
+
 	if (!server->socket) {
 		sway_log(SWAY_ERROR, "Unable to open wayland socket");
 		wlr_backend_destroy(server->backend);
 		return false;
 	}
 
+	if (setenv("WAYLAND_DISPLAY", server->socket, true)) {
```
emersion, line 171: The `setenv` already happens in `main.c` on line 390.

panaman67:
If this is best practice, should a method be made in wlroots?

Geo25rey:
> If this is best practice, should a method be made in wlroots?

That makes sense. Are you thinking adding a function wlr_renderer_init_wl_display_socket to render/wlr_renderer.c?

emersion (approved):
LGTM, thanks!

PR #5689, opened 2020-09-15
cursor: update hide timer during config apply

### LINKED ISSUE #5686: sway crashes instantly (segmentation fault)
- **Sway Version:**
sway version 1.5-7aea4692 (Sep 15 2020, branch 'master')

- **Debug Log:**
[sway.log](https://gist.github.com/antisocrates/5c2db46e43234b98ff64f47b8e7de248)

- **Configuration File:**
[config](https://gist.github.com/antisocrates/85062c5f45acea51675ce7ffbfe7f84f)

- **Stack Trace:**
[journalctl](https://gist.github.com/antisocrates/c10265186126edda35470ac2e9f72b19)

    [coredump](https://ufile.io/j0nj3mfc)

- **Description:**
 Starting sway from tty, instantly crashes with segmentation fault.

Emantor:
We can't arm the timer during cursor creation since the config may not
be ready yet. Instead arm the timer while applying the input
configuration, cursor_get_timeout() can now handle this even if the
configuration is not ready yet:

Return NULL in calls to seat_get_config*() if the config is not
initialized yet. This is required since we try to retrieve the hide
timeout from the cursor config, even if the config is not initialized.

Return 0 if we can't find any input configuration. This is okay even if
a timeout is configured, since the config will reapply the seat
configuration after load and sets up the correct timer with the correct
timeout.

Fixes #5686

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -1041,8 +1041,6 @@ struct sway_cursor *sway_cursor_create(struct sway_seat *seat) {
 
 	cursor->hide_source = wl_event_loop_add_timer(server.wl_event_loop,
 			hide_notify, cursor);
-	wl_event_source_timer_update(
-			cursor->hide_source, cursor_get_timeout(cursor));
 
 	wl_list_init(&cursor->image_surface_destroy.link);
 	cursor->image_surface_destroy.notify = handle_image_surface_destroy;
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -722,6 +722,8 @@ static void seat_configure_pointer(struct sway_seat *seat,
 	wlr_cursor_attach_input_device(seat->cursor->cursor,
 		sway_device->input_device->wlr_device);
 	seat_apply_input_config(seat, sway_device);
+	wl_event_source_timer_update(
+			seat->cursor->hide_source, cursor_get_timeout(seat->cursor));
 }
 
 static void seat_configure_keyboard(struct sway_seat *seat,


Emantor:
Apologies for the breakage.

antisocrates:
Fixes both the segmentation fault crash bug, and timeout is honored after login. Thanks

sway/input/seat.c
```diff
@@ -1430,6 +1432,9 @@ void seat_apply_config(struct sway_seat *seat,
 
 struct seat_config *seat_get_config(struct sway_seat *seat) {
 	struct seat_config *seat_config = NULL;
+	if(!config) {
```
emersion, line 1435: Style: missing space after `if`

sway/input/seat.c
```diff
@@ -1442,6 +1447,9 @@ struct seat_config *seat_get_config(struct sway_seat *seat) {
 
 struct seat_config *seat_get_config_by_name(const char *name) {
 	struct seat_config *seat_config = NULL;
+	if(!config) {
```
emersion, line 1450: Ditto

Emantor:
Fixed both style issues.

sway/input/cursor.c
```diff
@@ -246,6 +246,9 @@ int cursor_get_timeout(struct sway_cursor *cursor) {
 	if (!sc) {
 		sc = seat_get_config_by_name("*");
 	}
+	if (!sc) {
```
emersion, line 249: This is already accounted for with the ternary expression below.

sway/input/seat.c
```diff
@@ -1430,6 +1432,9 @@ void seat_apply_config(struct sway_seat *seat,
 
 struct seat_config *seat_get_config(struct sway_seat *seat) {
 	struct seat_config *seat_config = NULL;
+	if (!config) {
```
emersion, line 1435: Hmm. But `seat_apply_input_config` unconditionally accesses `config`. Does something make `cursor_get_timeout` different?

Emantor: No, I just tested again and have now removed the line and verified it by using some additional logging. I applied the changes for the crashes first and moved the event source at the very last. All that is required is moving the arm of the timer.

Emantor:
This should be ready for merge, thanks for the review.

sway/input/seat.c
emersion: OK. If removing completely the checks make you uncomfortable, we can add assertions instead.

emersion (approved):
LGTM, thanks!

PR #5685, opened 2020-09-15
cursor: arm cursor hide timer immediately

Emantor:
According to the wayland docs, wayland timers are disarmed on creation.
This leads to the cursor not being hidden if there is no activity after
creation, since the timer is armed on activity, but not at creation.
Arm the timer after creation to ensure the cursor is hidden even if
there is no cursor activity after creation.

Fixes #5684

PR #5695, opened 2020-09-17
input/seat: arm cursor hide timeout for all devices rather than only pointers

Geo25rey:
Depends on [wlroots#2403](https://github.com/swaywm/wlroots/pull/2403)

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -722,8 +722,6 @@ static void seat_configure_pointer(struct sway_seat *seat,
 	wlr_cursor_attach_input_device(seat->cursor->cursor,
 		sway_device->input_device->wlr_device);
 	seat_apply_input_config(seat, sway_device);
-	wl_event_source_timer_update(
-			seat->cursor->hide_source, cursor_get_timeout(seat->cursor));
 }
 
 static void seat_configure_keyboard(struct sway_seat *seat,
@@ -822,6 +820,9 @@ void seat_configure_device(struct sway_seat *seat,
 		case WLR_INPUT_DEVICE_TABLET_PAD:
 			seat_configure_tablet_pad(seat, seat_device);
 			break;
+	}
+	if (wlr_cursor_is_input_device_attached(seat->cursor->cursor, input_device->wlr_device)) {
+		cursor_handle_activity(seat->cursor, input_device->wlr_device);
 	}
 }
 


Geo25rey:
Any suggestion for consolidating the commit message?

Geo25rey:
@emersion Have you had time to look at this?

TheAvidDev:
> Have you had time to look at this?

You should wait around a week before bumping.

Geo25rey:
I just figured I'd ask since he responded on my wlroots PR

sway/input/seat.c
```diff
@@ -823,6 +821,9 @@ void seat_configure_device(struct sway_seat *seat,
 			seat_configure_tablet_pad(seat, seat_device);
 			break;
 	}
+	if (wlr_cursor_is_input_device_attached(seat->cursor->cursor, input_device->wlr_device)) {
```
emersion, line 824: Why isn't `if (input_device->wlr_device->type != WLR_INPUT_DEVICE_KEYBOARD && input_device->wlr_device->type != WLR_INPUT_DEVICE_SWITCH)` enough?

emersion:
Can you describe what this PR is trying to fix?

Geo25rey:
This does the same thing as #5685, but for all devices that are can control the cursor in a given seat; this includes touch screens and tablet devices, rather than just pointer devices.

sway/input/seat.c
Geo25rey: That doesn't guarantee that the corresponding device is associated with the seat's cursor

emersion: `seat_configure_device` is not called if the device doesn't belong to the seat.

Geo25rey: `wlr_cursor_is_input_device_attached` should take that into consideration

emersion: I don't get it.

The whole function is not called if the device isn't attached to the seat, so there's no point in checking that here.

Geo25rey: All the devices are attached to the seat as a prerequisite. `wlr_cursor_is_input_device_attached` checks if the specified device controls the seat's cursor. Then function uses the wlr_cursor as the "object" not the seat itself.

emersion: There's no need to check that.

For instance, let's say a pointer gets added to the seat. `seat_configure_pointer` gets called. It itself calls `wlr_cursor_attach_input_device`. Then `wlr_cursor_is_input_device_attached` will always return true.

Geo25rey: But `seat_configure_device` runs for all devices, so non-pointing devices have to be ruled out or only pointing devices should be included. To prevent code duplicates in the device specific configure functions of pointing devices, this will emulate cursor activity to arm the hide timeouts. 

I chose this method since it is more easily maintained in such a case where a new kind of pointing device is added.

emersion: I'd prefer not to add a new wlroots function just because of how the Sway code is architectured.

Geo25rey: Well the functionality was already in wlroots. It just wasn't publicly accessible.

Issue #5667, opened 2020-09-08
Qt apps crash Sway


J0nnyMak0:
Clicking on any menu in Zoom crashes Sway. Log and trace attached.
`sway version 1.5-7ca9ef12 (Sep  4 2020, branch 'master')`
[gdb.txt](https://github.com/swaywm/sway/files/5191308/gdb.txt)
[sway.log](https://github.com/swaywm/sway/files/5191309/sway.log)

emersion:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f5e122d0063 in wlr_surface_is_xwayland_surface (surface=0x0) at ../subprojects/wlroots/xwayland/xwm.c:89
89		return surface->role == &xwayland_surface_role;
[Current thread is 1 (Thread 0x7f5e10db1940 (LWP 896))]
(gdb) bt full
#0  0x00007f5e122d0063 in wlr_surface_is_xwayland_surface (surface=0x0) at ../subprojects/wlroots/xwayland/xwm.c:89
#1  0x0000558f03e7b14b in unmanaged_handle_unmap (listener=0x558f053cd470, data=0x558f053cd190) at ../sway/desktop/xwayland.c:99
        previous = 0x3d5
        surface = 0x558f053cd3f0
        xsurface = 0x558f053cd190
        seat = 0x558f04e5cba0
#2  0x00007f5e122c7cd4 in wlr_signal_emit_safe (signal=0x558f053cd318, data=0x558f053cd190) at ../subprojects/wlroots/util/signal.c:29
        pos = 0x558f053cd470
        l = 0x558f053cd470
        cursor = {link = {prev = 0x558f053cd470, next = 0x7fff1eb8a270}, notify = 0x7f5e122c7c1e <handle_noop>}
        end = {link = {prev = 0x7fff1eb8a250, next = 0x558f053cd318}, notify = 0x7f5e122c7c1e <handle_noop>}
#3  0x00007f5e122d211e in xsurface_unmap (surface=0x558f053cd190) at ../subprojects/wlroots/xwayland/xwm.c:835
#4  0x00007f5e122d27d1 in xwm_handle_unmap_notify (xwm=0x558f0529cdc0, ev=0x558f053cdfe0) at ../subprojects/wlroots/xwayland/xwm.c:987
        xsurface = 0x558f053cd190
[...]

PR #5700, opened 2020-09-19
check parent surface before it is destroyed

J0nnyMak0:
Fixes https://github.com/swaywm/sway/issues/5667


emersion:
Does this fix the original bug? (Please add a link to it)

J0nnyMak0:
> Does this fix the original bug? (Please add a link to it)

Done.

emersion (approved):
OK, thanks!

PR #5705, opened 2020-09-23
config: check the return value of xkb_context_new

jhofstee:
When xkb_context_new returns NULL and that is passed on to
xkb_keymap_new_from_names, sway will crash. This happens for example
if xkeyboard-config is not installed. Exit with an error message instead.

While at it, check xkb_keymap as well.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -38,10 +38,15 @@ struct sway_config *config = NULL;
 static struct xkb_state *keysym_translation_state_create(
 		struct xkb_rule_names rules) {
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
+	if (!context)
+		sway_abort("Failed to create XKB context");
+
 	struct xkb_keymap *xkb_keymap = xkb_keymap_new_from_names(
 		context,
 		&rules,
 		XKB_KEYMAP_COMPILE_NO_FLAGS);
+	if (!xkb_keymap)
+		sway_abort("Failed to compile global XKB keymap");
 
 	xkb_context_unref(context);
 	return xkb_state_new(xkb_keymap);


sway/config.c
```diff
@@ -38,10 +38,15 @@ struct sway_config *config = NULL;
 static struct xkb_state *keysym_translation_state_create(
 		struct xkb_rule_names rules) {
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
+	if (!context)
```
emersion, line 41: Style: braces are mandatory

sway/config.c
```diff
@@ -38,10 +38,15 @@ struct sway_config *config = NULL;
...
+		sway_abort("Failed to create XKB context");
+
 	struct xkb_keymap *xkb_keymap = xkb_keymap_new_from_names(
 		context,
 		&rules,
 		XKB_KEYMAP_COMPILE_NO_FLAGS);
+	if (!xkb_keymap)
```
emersion, line 48: Ditto

sway/config.c
```diff
@@ -38,10 +38,15 @@ struct sway_config *config = NULL;
 static struct xkb_state *keysym_translation_state_create(
 		struct xkb_rule_names rules) {
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
+	if (!context)
+		sway_abort("Failed to create XKB context");
```
emersion, line 42: We shouldn't crash on this, instead we should log the error and return NULL.

emersion (changes requested):
Thanks for your PR! Here are some comments.

george-hopkins:
@jhofstee Thank you for the patch, it helped me narrowing down an issue I had. Do you intend to keep working on it or would you mind if someone else would take over?

jhofstee:
> @jhofstee Thank you for the patch, it helped me narrowing down an issue I had. Do you intend to keep working on it or would you mind if someone else would take over?

A bit late, but feel to take over.

PR #5709, opened 2020-09-26
commands/move: fix single-split escaping on move

Xyene:
Prior to this commit, having a layout like `T[app1 V[app2]]`, focusing
`app2`, and then doing `move left` would result in `T[app2 app1]`. Now, the
resulting layout is `T[app1 app2]`, which matches i3 behavior.

`container_flatten` updates `container->parent`, meaning that the
existing check would never be true.


RedSoxFan (approved):

RedSoxFan:
Thanks!

Issue #5064, opened 2020-03-02
Floating window content is drawn one pixel too high on an output with a negative y coordinate after move


YSelfTool:
* Sway Version: sway version 1.4
* Debug Log: https://gist.github.com/YSelfTool/9283eabf5c9d414ea81dc349a3eb7cda
* Configuration File: https://gist.github.com/YSelfTool/eb2d60136cb50442bf1b20ee882049ee

1. Enable an output with a negative y coordinate, i.e. above the normal 0,0 output.
2. Open floating window on that output.
3. Move the window a bit by grabbing the header.

The window content will be drawn one pixel too high, resulting in a white (i.e. background color) line above the bottom decoration and an overlap with the top decoration: ![Screenshot](https://www.rsonn.de/downloads/sway-bug.png)
This matches the attached config file, the bottom window is on an output at 0,0, the top on an output  at 0,-1200.

~~I haven't seen the problem with other programs than alacritty.~~ The problem is independent of te application. I have tested and seen it with alacritty, firefox (wayland) and xclocks (using xwayland). I haven't tested negative x coordinates, but they might be affected as well.

This doesn't happen on outputs with positive positions, and it doesn't happen if the window is moved with key bindings, i.e. grabbing with the mouse is required. The offset only appears when moving the window.

dimitris-triantafyllidis:
Hi. I can't reproduce this issue in my branch which is mentioned in this pull request: #5663. The keyboard probably hides this because it moves by an integer amount of pixels.

Xyene:
Fixed in https://github.com/swaywm/sway/commit/aac1582ea90897a25408a520b8f3bde0552a69cc.

PR #5711, opened 2020-09-27
Fix #5643, #5064: rounding issues in floating-point -> integer conversions

### LINKED ISSUE #5643: Right border renders incorrectly when left border is offscreen

dimitris-triantafyllidis:
Currently, various floating-point expressions involving the coordinates of borders, title bars and content surfaces are directly assigned to integers, and so they are rounded towards zero.

This results in off-by-one distances between these elements when the signs of their coordinates differ.

The glitches are easier to see with semitransparent borders and title bars, where one can see that they can overlap with the content surface when it enters an area with a negative coordinate. 

Fixed by wrapping these expressions with a call to floor before the assignment.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -105,8 +105,8 @@ static bool get_surface_box(struct surface_iterator_data *data,
 		data->rotation);
 
 	struct wlr_box box = {
-		.x = data->ox + _sx,
-		.y = data->oy + _sy,
+		.x = floor(data->ox + _sx),
+		.y = floor(data->oy + _sy),
 		.width = sw,
 		.height = sh,
 	};
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -346,8 +346,8 @@ static void render_view(struct sway_output *output, pixman_region32_t *damage,
 	if (state->border_left) {
 		memcpy(&color, colors->child_border, sizeof(float) * 4);
 		premultiply_alpha(color, con->alpha);
-		box.x = state->x;
-		box.y = state->content_y;
+		box.x = floor(state->x);
+		box.y = floor(state->content_y);
 		box.width = state->border_thickness;
 		box.height = state->content_height;
 		scale_box(&box, output_scale);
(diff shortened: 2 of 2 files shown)


dimitris-triantafyllidis:
Any feedback on this?

emersion:
>directly assigned to integers, and so they are rounded towards zero.

Doesn't `floor` result in exactly the same thing?

dimitris-triantafyllidis:
For positive floats sure, but negative ones will be rounded towards negative infinity with `floor` instead of towards zero without it. The point of this PR is to round everything towards the same direction to preserve distances when positive and negative coordinates are mixed.

dimitris-triantafyllidis:
@emersion Gentle ping

Xyene:
Hi, sorry for the long review cycle...

Your commit text makes sense and the patch is straightforward, but I'm not sure how to reproduce this locally. Is it necessary for part of the window to be in negative coordinates, and part in positive? Some simple "ELI'm-not-sure-how-to-reproduce-this-in-order-to-test-this-patch" instructions, and maybe some before-after screenshots would be helpful.

dimitris-triantafyllidis:
@Xyene No worries, I understand :-)

Before / after sample:
before
after

In both screenshots the smaller window renders well, but the larger one (which goes partly off the top-left corner of the monitor) has a titlebar, a bottom border, a left border, and a content whose top-left corners have negative coordinates. So the title bar is rendered one pixel off to the right and to the bottom (see overlap with the top edge of the right border), the bottom border is rendered one pixel off to the right, and if you squint a bit you can see that the window content is also rendered one pixel off to the right, overlapping with the right border. This last one is actually not very easy to see in this screenshot...

Note 1: You may have to "shake" the window with the mouse a bit so that the coordinates become fractional, to see the issue.
Note 2: If you try with semi transparent borders and a stacked or tabbed floating container, you may notice that there is a one pixel "jog" between the titlebar(s) and the rest of the window. This is unrelated to the current PR and happens all over the monitor when dragging the window.

If you need further info or screenshots I will be happy to provide them.

emersion:
Thanks a lot of the screenshots, that's very helpful.

Can you rebase on top of master as a single commit? See https://git-rebase.io/.

dimitris-triantafyllidis:
I resolved a conflict here and the commits became two, then I just redid the branch locally and force-pushed as a single commit, am I missing something?

emersion (approved):
Oh, sorry, I didn't notice you pushed a new commit.

This LGTM, thanks. I'll let @Xyene merge this if he agrees.

Xyene (approved):
Thanks for working on this!

PR #5712, opened 2020-09-28
layer-shell: error on 0 dimension without anchors

ifreund:
The protocol requires clients to set opposing anchors when requesting
a width or height of 0.

The goal of this patch is not to break clients that rely on this
behavior but to improve the consistency of the layer shell ecosystem
through adherence to the protocol.


wmww (approved):

ifreund:
For reference:
https://github.com/swaywm/sway/blob/c150177a94e1b3df3710d1b2b3db1d59fb396685/protocols/wlr-layer-shell-unstable-v1.xml#L114-L117

emersion:
Can we send the protocol errors in wlroots?

ifreund:
> Can we send the protocol errors in wlroots?

Of course, not sure why I didn't do that in the first place. https://github.com/swaywm/wlroots/pull/2422

PR #5716, opened 2020-10-02
tiling_drag: emit window move ipc events

RPigott:
At the moment moving windows with the tiling drag feature doesn't emit ipc events, but I think it probably should.

It looks like the swap command doesn't currently emit a window move event, so I've just added a window move event to the actual move action of a tiling drag, not the swap.

### DIFF
--- a/sway/input/seatop_move_tiling.c
+++ b/sway/input/seatop_move_tiling.c
@@ -5,6 +5,7 @@
 #include "sway/desktop.h"
 #include "sway/input/cursor.h"
 #include "sway/input/seat.h"
+#include "sway/ipc-server.h"
 #include "sway/output.h"
 #include "sway/tree/arrange.h"
 #include "sway/tree/node.h"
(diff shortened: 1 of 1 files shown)


0x7D2B:
Why not have floating window drags emit an event as well: https://github.com/RPigott/sway/pull/1

RPigott:
@0x7D2B do they do that on i3?

I think the idea is that the tiling drag actions each correspond to some of the usual window management commands that already emit window move events. i.e. `swaymsg move right` on a tiling window emits an ipc window event, so the equivalent drop zone in a tiling drag should as well. `swaymsg move right 10px` on a floating window doesn't emit an ipc window event, so dragging a floating window 10px to the right shouldn't either.

Unless i3 already emits move events for floating drags, I think you'd need to take it up with them if you want the event emitted in the floating case. FWIW, I think I'd prefer the introduction of a new floating_move event over using the move event in that case.

i3 doesn't implement tiling drag, so we use our best judgement there.

0x7D2B:
@RPigott Ah, I see now - I somehow didn't connect it to the original window move action's IPC event.

I think it's weird that `swaymsg move right` on a tiled window would emit an event but `swaymsg move right 10px` on a floating window wouldn't, but if this is done for consistency with i3, then I think this is something I should raise as a separate issue.

emersion (approved):
Thanks!

PR #5717, opened 2020-10-03
require systemd/elogind >=242

nmschulte:
for sd-bus macro `SD_BUS_METHOD_WITH_NAMES` introduced in v242
https://github.com/systemd/systemd/commit/856ad2a86bd9b3e264a090fcf4b0d05bfaa91030#diff-0664187e1dac5dd69ed30c8577c27d3cR121

see conversation following https://github.com/swaywm/sway/pull/5161#issuecomment-702991211


TheAvidDev:
#5161 isn't merged, or even marked ready for review, so this is probably better inside of that PR than standalone.

nmschulte:
Oh so it is.  I thought these were introduced a while back on `master` but must've misunderstood.

TheAvidDev:
To be honest, I don't know much about the situation. From what I see though, the `SD_BUS_METHOD_WITH_NAMES` is being introduced in #5161. If an error occurs outside of #5161 on plain master then this is totally valid.

Issue #5690, opened 2020-09-15
Missing option from i3: workspace_min_width


palb91:
- **i3 PR:**
  - i3/i3@94228fd9 (Jun 25, 2019).

- **Description:**
  - An option is missing in sway, `workspace_min_width <px> [px]`;
  - It is just a visual effect, but I'm currently migrating my configuration to sway and noticed it;
  - This option sets a **minimal size for workspace buttons**;
  - In the following picture, you can see the difference in both environments:
    ![i3_sway_bar](https://user-images.githubusercontent.com/29063375/93268771-13fded00-f7ae-11ea-84d1-fc3976a10bcc.png)
    Both screenshot were taken from the top-left of the screen, just to show why I would like to have this feature in sway too :wink:

- **From i3 Userguide:**
  -  https://i3wm.org/docs/userguide.html#_minimal_width_for_workspace_buttons
  - > **5.13. Minimal width for workspace buttons**
    >
    > By default, the width a workspace button is determined by the width of the text showing the workspace name. If the name is too short (say, one letter), then the workspace button may look too small.
    > 
    > This option specifies the minimum width for workspace buttons. If the name of a workspace is too short to cover the button, an additional padding is added on both sides of the button so that the text is centered.
    > 
    > The default value of zero means that no additional padding is added.
    > 
    > The setting also applies to the current binding mode indicator.
    > 
    > Note that the specified pixels refer to logical pixels, which may translate into more pixels on HiDPI displays.
    > 
    > **Syntax:**
    > ```
    > workspace_min_width <px> [px]
    > ```
    > 
    > **Example:**
    > ```
    > bar {
    >     workspace_min_width 40
    > }
    > ```

Geo25rey:
Does the `bar tray_padding` command execute the desired behavior?

palb91:
No, and `bar tray_padding` doesn't seem to do exactly what the man says:
```
TRAY
       tray_padding <px> [px]
           Sets the pixel padding of the system tray. This padding will sur‐
           round the tray on all sides and between each item. The default
           value for px is 2.
```
On my swaybar, the padding is effective only on the top and the bottom, not on **all sides and between each item**. I tried with something quite visual: `tray_padding 50 px`.

palb91:
Aaand… it works! Thanks for the implementation !

PR #5719, opened 2020-10-03
Add support for workspace_min_width bar option.

tarmack:
https://github.com/swaywm/sway/issues/5690 requests support for workspace_min_width in swaybar. This pull requests implements that support.

I am not sure if there are any guidelines to the order/position of items added to existing structs. I tried to make them logical but there is already some discrepancy between the sway and swaybar versions of the config struct.

### DIFF
--- a/sway/commands/bar.c
+++ b/sway/commands/bar.c
@@ -36,6 +36,7 @@ static struct cmd_handler bar_handlers[] = {
 	{ "unbindcode", bar_cmd_unbindcode },
 	{ "unbindsym", bar_cmd_unbindsym },
 	{ "workspace_buttons", bar_cmd_workspace_buttons },
+	{ "workspace_min_width", bar_cmd_workspace_min_width },
 	{ "wrap_scroll", bar_cmd_wrap_scroll },
 };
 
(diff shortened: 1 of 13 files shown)


sway/commands/bar.c
```diff
@@ -37,6 +37,7 @@ static struct cmd_handler bar_handlers[] = {
 	{ "unbindsym", bar_cmd_unbindsym },
 	{ "workspace_buttons", bar_cmd_workspace_buttons },
 	{ "wrap_scroll", bar_cmd_wrap_scroll },
+	{ "workspace_min_width", bar_cmd_workspace_min_width },
```
RedSoxFan, line 40: This must be in alphabetical order for bsearch to work

tarmack: I knew this.. :-/ sorry for making you repeat a comment right above the change. It is fixed now.

emersion (changes requested):
Code LGTM! Can we add this to the sway-bar man page as well?

tarmack:
@emersion, good point! Added the setting to the sway-bar man page as well.

emersion (approved):
Thanks!

PR #5738, opened 2020-10-15
seat: use default output mapping if there is no input config

### LINKED ISSUE #5638: Cursor position is wonky with multiple outputs in Wayland backend
```
$ swaymsg -pt get_version
sway version 1.5-6991ac8c (Aug 18 2020, branch 'master')
$ swaymsg -s $SWAYEMBED -pt get_version
sway version 1.5-6991ac8c (Aug 22 2020, branch 'master')
```

When starting an embedded sway session with `WLR_WL_OUTPUTS=$n` where n > 1, I observe the the cursor position in those outputs doesn't match the drawn cursor. Specifically in an interactive move the y position matches, but the x position is n*x with respect to the offset from 0,0 in the top left.

i.e.
```
$ WLR_WL_OUTPUTS=2 sway
$ swaymsg -s SWAYEMBED exec '$term', floating enable, move abosolute position 0 0
```

Now begin an interactive move using the floating modifier. Observe:

1. moving the cursor to 0,100 moves the client to 0,100
2. moving the cursor to 100,0 moves the client to 200,0.

For clarity, I'm starting sway from a terminal in the parent session that is the only client in a horizontal layout workspace. So I got from H[termite] to H[termite WL-0 WL-1], with each output taking a third of my screen. But even when I move those outputs to their own workspaces the linear offset remains.

Regardless of the geometry of the Wayland output windows in the parent session, using the `swaymsg move` commands moves clients to the intended position. e.g. `swaymsg -s $SWAYEMBED move absolute position 100 0` moves a client to a 100 pixel offset. The position is reported as 100,0 from get_tree and the real distance moved by the client is also 100 pixels, matched by moving the cursor in the parent session 100 pixels over with the ipc.

If I once again place my terminal at 0,0 and invoke slurp with the client geometry from the get_tree response, the client area is highlighted correctly by slurp in the drawn overlay, but only when my cursor is in the left half of the rectangle. I must also click in the left half of the rectangle for slurp to consider the area selected.

RPigott:
This causes sway to use the default output mapping for input device even when there is no applicable input config.

The motivation here is to fix #5638, in which pointer coordinates in individual Wayland backed outputs are incorrectly interpreted as in the coordinate space of the abstract output plane. With this commit each pointer is mapped to it's own output which fixes the offset.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -660,17 +660,16 @@ static void seat_apply_input_config(struct sway_seat *seat,
 		struct sway_seat_device *sway_device) {
 	struct input_config *ic =
 		input_device_get_config(sway_device->input_device);
-	if (ic == NULL) {
-		return;
-	}
 
 	sway_log(SWAY_DEBUG, "Applying input config to %s",
 		sway_device->input_device->identifier);
 
-	const char *mapped_to_output = ic->mapped_to_output;
-	struct wlr_box *mapped_to_region = ic->mapped_to_region;
+	const char *mapped_to_output = ic == NULL ? NULL : ic->mapped_to_output;
+	struct wlr_box *mapped_to_region = ic == NULL ? NULL : ic->mapped_to_region;
+	enum input_config_mapped_to mapped_to =
+		ic == NULL ? MAPPED_TO_DEFAULT : ic->mapped_to;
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

PR #5742, opened 2020-10-18
transaction: don't reconfigure X views unless integral coords changed

### LINKED ISSUE #5035: Dragging a floating XWayland window lags and snaps
### Sway Version:

`sway version 1.4`

### Debug Log:

[sway.log](http://0x0.st/iZY9.log)

### Configuration File:

Default with the following additions:

```
input type:pointer
{
	accel_profile flat
	pointer_accel -0.5
}
```

`accel_profile flat` is optional but it will be easier to see the issue than using `adaptive`.

### Reproduction:

Drag a floating XWayland window. Either through the titlebar or by holding `Super`.

When the cursor moves the window will lag and then snap to the cursor's position. It will keep doing this the whole time.

### Additional information:

I tested this with Weston where this didn't happen.

`.config/weston.ini`
```
[core]
xwayland=true

[libinput]
accel-profile=flat
accel-speed=-0.5
```

I only notice this with _negative_ `pointer_accel` values. It is also far more noticable with a flat profile.

The faster you move the mouse the less it happens. You can see this better with `accel_profile adaptive` where you have to move the mouse really slow to notice this. `swaymsg 'input type:pointer pointer_accel -1'` makes it more obvious.

Xyene:
Sway logical coordinates are doubles, but they get truncated to integers
when sent to Xwayland through `xcb_configure_window`. X11 apps will not
respond to duplicate configure requests (from their truncated point of
view) and cause transactions to time out.

Fixes #5035.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -397,8 +397,12 @@ static bool should_configure(struct sway_node *node,
 	// Xwayland views are position-aware and need to be reconfigured
 	// when their position changes.
 	if (node->sway_container->view->type == SWAY_VIEW_XWAYLAND) {
-		if (cstate->content_x != istate->content_x ||
-				cstate->content_y != istate->content_y) {
+		// Sway logical coordinates are doubles, but they get truncated to
+		// integers when sent to Xwayland through `xcb_configure_window`.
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

PR #5745, opened 2020-10-19
transaction: stop matching X transactions by geometry

Xyene:
JetBrains IDEs really keep providing the most pathological cases for
Xwayland...

The "Search everywhere" dialog can be resized, but it responds to a
configure request with a configure notify including double the growth in
width.

So, if the user drags the left edge by 5px, we send Xwayland a configure
request based off the current geometry, -5px on the x axis, and +5px
width. JetBrains IDEs then respond with -5px on the x axis, and +10px
width.

Presumably they do this to keep the dialog symmetric. Still, it breaks
Sway because we match commits to transactions by geometry, so any
resizes of the "Search everywhere" dialog will cause Sway's transactions
to time out.

Instead, we can just assume that the first commit following a
transaction is in response to that transaction. This is obviously
race-y, but I don't think we can do anything particularly smarter in the
face of crazy applications.

### DIFF
--- a/include/sway/desktop/transaction.h
+++ b/include/sway/desktop/transaction.h
@@ -37,15 +37,6 @@ void transaction_notify_view_ready_by_serial(struct sway_view *view,
 		uint32_t serial);
 
 /**
- * Notify the transaction system that a view is ready for the new layout, but
- * identifying the instruction by geometry rather than by serial.
- *
- * This is used by xwayland views, as they don't have serials.
- */
-void transaction_notify_view_ready_by_geometry(struct sway_view *view,
-		double x, double y, int width, int height);
-
-/**
  * Unconditionally notify the transaction system that a view is ready for the
  * new layout.
  */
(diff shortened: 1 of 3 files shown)


Xyene:
Actually, this is probably not good for merging yet. PyCharm behaves differently in Wayfire (i.e. does not do the weird resizing). Nor does it do so on Windows. Marking as draft until the discrepancy is tracked down.

Xyene:
Aha, I've tracked down the culprit:

https://github.com/swaywm/sway/blob/8355884fbd4ea04203614172424b27c5b74018ab/sway/desktop/xwayland.c#L528

Due to the `_and_center` part, the container's (x, y) position will change, so the transaction will not match up.

/cc @RedSoxFan from git blame (https://github.com/swaywm/sway/commit/dd28e6a6d6abf06d2d16e6c91aeaf942bf225af7), since I don't really understand how resizing is supposed to work here. The centering causes the transactions to hang, but resizing doesn't work at all without that call.

PR #5749, opened 2020-10-21
commands/move: fix crash when moving sphsc child

### LINKED ISSUE #5707: Sway crashes when moving children of container currently in scratchpad
###### Setup
- Sway Version: 1.5
- [Debug Log](https://gist.github.com/QuentinI/edfc9e000a3fd344441b62187c549a73)
- [Configuration File](https://gist.github.com/QuentinI/3f4702e63a3b586872063c93fccf3b9c) (only difference from default is removal of default wallpaper)

###### Description
Sway crashes when moving children of a container residing in scratchpad to any workspace. Steps to reproduce with default config (debug log is of me doing this):
- Create a new terminal emulator window (`$mod+Return`)
- Execute `swaymsg 'splith'` in it
- Switch to another workspace (`$mod+2`)
- Create yet another terminal emulator window (`$mod+Return`)
- Run `swaymsg '[con_id=6] move to scratchpad'` in it
- Return to workspace 2 (`$mod+2`)
- Run `swaymsg '[con_id=5] move to workspace 1'`
- Sway crashes

Sway still crashes if there are more children in parent container (i.e. moving one child doesn't leave it empty). Sway doesn't seem to crash if parent container isn't in scratchpad.

RPigott:
Fixes #5707.

Yet another bug with scratchpad hidden split containers. Children of sphsc need to be shown before we move them. This change also stops calling root_scratchpad_show for non-hidden scratchpad containers, but afaict that's redundant anyway so the behavior should be the same.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -537,7 +537,8 @@ static struct cmd_results *cmd_move_container(bool no_auto_back_and_forth,
 	struct sway_node *focus = seat_get_focus(seat);
 
 	// move container
-	if (container->scratchpad) {
+	if (container_is_scratchpad_hidden_or_child(container)) {
+		container_detach(container);
 		root_scratchpad_show(container);
 	}
 	switch (destination->type) {


RPigott:
Actually, showing the parent window is a pretty unfortunate side-effect. Don't merge this and I'll come up with something better.

RPigott:
Better.

Xyene (approved):
Thanks!

PR #5751, opened 2020-10-21
commands/resize: don't consider 1px resizes to be invalid

danieldg:
A "resize shrink width 1px" will cause grow_x to be 0 while grow_width is -1,
incorrectly rejecting the command even though the resize is not a noop.  Fix
this by checking width/height instead of x/y.


emersion (approved):
LGTM, thanks!

PR #5752, opened 2020-10-22
input: tweak focus behavior to allow focusing parent containers

Xyene:
Sway focuses the inactive child when focusing split containers. However,
there is currently no way to focus the parent container itself by mouse.
A user must use the keyboard (`focus parent`) to do so.

This commit maintains the current behavior, but makes it such that a
second click on the split container titlebar (i.e., after its children
are visible) focuses the split container itself.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -429,13 +429,31 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 		}
 	}
 
+	// Handle changing focus when clicking on a container
+	if (cont && state == WLR_BUTTON_PRESSED) {
+		// Default case: focus the container that was just clicked.
+		node = &cont->node;
+
+		// If the container is a tab/stacked container and the click happened
+		// on a tab, switch to the tab. If the tab contents were already
+		// focused, focus the tab container itself. If the tab container was
+		// already focused, cycle back to focusing the tab contents.
+		if (on_titlebar) {
+			struct sway_container *focus = seat_get_focused_container(seat);
+			if (focus == cont || !container_has_ancestor(focus, cont)) {
+				node = seat_get_focus_inactive(seat, &cont->node);
+			}
+		}
+
+		seat_set_focus(seat, node);
+		transaction_commit_dirty();
+	}
+
 	// Handle beginning floating move
 	if (cont && is_floating_or_child && !is_fullscreen_or_child &&
 			state == WLR_BUTTON_PRESSED) {
 		uint32_t btn_move = config->floating_mod_inverse ? BTN_RIGHT : BTN_LEFT;
 		if (button == btn_move && (mod_pressed || on_titlebar)) {
-			seat_set_focus_container(seat,
-					seat_get_focus_inactive_view(seat, &cont->node));
 			seatop_begin_move_floating(seat, container_toplevel_ancestor(cont));
 			return;
 		}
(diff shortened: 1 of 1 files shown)


Xyene:
Initial state:

image

After clicking the second tab:

image

After clicking the second tab again:

image

travankor:
Does i3 have this feature, too?

RedSoxFan (commented):
This concept itself seems fine to me. It's been a while since I looked at this block of code and I'm just questioning things now.

Currently, this would be gated behind `config->tiling_drag`. I think it may make sense in the  `// Handle clicking a container surface or decorations` block below as well (L492). 

I also think that block should be `border or decorations` (if there is a surface, it should go into the block before) and there shouldn't be a call to `seat_pointer_notify_button`. but that's pre-existing code that's not directly related so that could be a separate issue.

Thoughts?

Xyene:
> Currently, this would be gated behind `config->tiling_drag`. I think it may make sense in the // Handle clicking a container surface or decorations block below as well (L492).

Good catch. Rather than duplicate the code, I've refactored this bit of input handling. This leaves the `// Handle clicking a container surface or decorations` practically empty now.

sway/input/seatop_default.c
```diff
@@ -452,38 +452,49 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
...
-		struct sway_container *focus = seat_get_focused_container(seat);
-		bool focused = focus == cont || container_has_ancestor(focus, cont);
-		if (on_titlebar && !focused) {
-			node = seat_get_focus_inactive(seat, &cont->node);
-			seat_set_focus(seat, node);
-		}
-
-		// If moving a container by it's title bar, use a threshold for the drag
+			cont && cont->fullscreen_mode == FULLSCREEN_NONE &&
+			cont == seat_get_focused_container(seat)) {
```
Xyene, line 478: This condition is new. It's meant to prevent a "weird" UI state where the current tab container is focused, and a drag is initiated. The new block above will have switched the focus to the child, so now the tab container looks "inactive" but still responds to the drag. This isn't super intuitive or obvious, so I've disallowed the case.

RedSoxFan: While testing, I kept focusing the parent and then dragging, which is what this prevents. This isn't functionality that I use though so I'm not sure how common that order would be.

Maybe it would actually be better to always focus `cont` here and then do the focus inactive handling in `finalize_move` in `seatop_move_tiling.c`?

RedSoxFan (commented):
Just as a note, this does not currently effect tabs/stacks in floating containers.

sway/input/seatop_default.c
Xyene: I like that idea a lot, actually. It'd add some complexity to the logic, since `seatop_begin_move_tiling` would have to now take a `struct sway_container *` to restore focus to at the end, iff there was no move... but the UX certainly would be much better.

Xyene:
I haven't had a chance to revisit this PR (but have not forgotten about it). https://github.com/i3/i3/pull/4230 got merged today, implementing the same behavior for i3 as this PR does for Sway.

Xyene:
I've gone and implemented this for floating tabs/stacks as well. It does feel a _bit_ weird because floating decoration clicks immediately trigger `seatop_move_floating` with no threshold (so we can have both this new behavior and a move trigger in the same click), but this feels less bad than the inconsistency of not doing this for floating windows.

Haven't yet addressed the UX around double-clicking a tab before initiating a drag.

Xyene:
i3 has added drags, so I've made this PR match i3's behavior in all cases. Should be good to go, almost 2 years later!

Xyene:
Gentle ping.

emersion (approved):
Thanks!

Issue #5754, opened 2020-10-23
Extra split when escaping container


RPigott:
```
$ swaymsg -pt get_version
sway version 1.5-cc8d318a (Oct 20 2020, branch 'master')
$ i3-msg -t get_version
4.18.2 (2020-07-26)
```

Begin with layout:

`H[H[app1 app2]]`

Steps to reproduce:

1. `swaymsg '[con_mark=app1]' move left`

i3's result:

`H[app1 H[app2]]`

sway's result:

`H[H[app1] H[app2]]`

RPigott:
Regressed by #5500. ping @Xyene.

Xyene:
I guess my analysis in https://github.com/swaywm/sway/pull/5500/files#diff-fcf5a35c88e94d24870f5242b57cc1120fd954e9849ed7dce9e05efed9be4748L279 was not entirely correct; I suspect that's what's breaking things here.

Xyene:
I think this was fixed in https://github.com/swaywm/sway/pull/5756, please re-open if I'm wrong.

PR #5756, opened 2020-10-23
Rework split, layout, workspace_layout, and move commands to work like i3

### LINKED ISSUE #5755: Extra split on singletons
```
$ swaymsg -pt get_version
sway version 1.5-cc8d318a (Oct 20 2020, branch 'master')
$ i3-msg -t get_version
4.18.2 (2020-07-26)
```

Begin with layout `H[app1]`

Steps to reproduce:

1. `swaymsg splith`

i3 result: `H[app1]` (no change)
sway result: `H[H[app1]]`

RPigott:
EDIT: Hijacking top comment b/c this PR is barely related to what it started as anyway

Hey you. Lurker. Please test.

---
Fix #5755 

This is intended to match i3's behavior here:

https://github.com/i3/i3/blob/3cd1c45eba6de073bc4300eebb4e1cc1a0c4479a/src/tree.c#L354

### DIFF
--- a/include/sway/tree/workspace.h
+++ b/include/sway/tree/workspace.h
@@ -116,6 +116,13 @@ struct sway_container *workspace_add_tiling(struct sway_workspace *workspace,
 void workspace_add_floating(struct sway_workspace *workspace,
 		struct sway_container *con);
 
+/**
+ * Adds a tiling container to the workspace without considering
+ * the workspace_layout, so the con will not be split.
+ */
+void workspace_insert_tiling_direct(struct sway_workspace *workspace,
+		struct sway_container *con, int index);
+
 struct sway_container *workspace_insert_tiling(struct sway_workspace *workspace,
 		struct sway_container *con, int index);
 
@@ -133,5 +140,13 @@ void workspace_get_box(struct sway_workspace *workspace, struct wlr_box *box);
 size_t workspace_num_tiling_views(struct sway_workspace *ws);
 
 size_t workspace_num_sticky_containers(struct sway_workspace *ws);
+
+/**
+ * workspace_squash is container_flatten in the reverse
+ * direction. Instead of eliminating redundant splits that are
+ * parents of the target container, it eliminates pairs of
+ * redundant H/V splits that are children of the workspace.
+ */
+void workspace_squash(struct sway_workspace *workspace);
 
 #endif
--- a/include/sway/tree/workspace.h
+++ b/include/sway/tree/workspace.h
@@ -110,13 +110,13 @@ void workspace_unwrap_children(struct sway_workspace *ws,
 
 void workspace_detach(struct sway_workspace *workspace);
 
-void workspace_add_tiling(struct sway_workspace *workspace,
+struct sway_container *workspace_add_tiling(struct sway_workspace *workspace,
 		struct sway_container *con);
 
 void workspace_add_floating(struct sway_workspace *workspace,
 		struct sway_container *con);
 
-void workspace_insert_tiling(struct sway_workspace *workspace,
+struct sway_container *workspace_insert_tiling(struct sway_workspace *workspace,
 		struct sway_container *con, int index);
 
 void workspace_remove_gaps(struct sway_workspace *ws);
--- a/sway/commands/layout.c
+++ b/sway/commands/layout.c
@@ -162,6 +162,12 @@ struct cmd_results *cmd_layout(int argc, char **argv) {
 			}
 			container->layout = new_layout;
 			container_update_representation(container);
+		} else if (config->handler_context.container) {
+			// i3 avoids changing workspace layouts with a new container
+			// https://github.com/i3/i3/blob/3cd1c45eba6de073bc4300eebb4e1cc1a0c4479a/src/con.c#L1817
+			container = workspace_wrap_children(workspace);
+			container->layout = new_layout;
+			container_update_representation(container);
 		} else {
 			if (old_layout != L_TABBED && old_layout != L_STACKED) {
 				workspace->prev_split_layout = old_layout;
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -309,16 +309,6 @@ static bool container_move_in_direction(struct sway_container *container,
 		return false;
 	}
 
-	// If container is in a split container by itself, move out of the split
-	if (container->parent) {
-		struct sway_container *old_parent = container->parent;
-		struct sway_container *new_parent =
-			container_flatten(container->parent);
-		if (new_parent != old_parent) {
-			return true;
-		}
-	}
-
 	// Look for a suitable *container* sibling or parent.
 	// The below loop stops once we hit the workspace because current->parent
 	// is NULL for the topmost containers in a workspace.
@@ -721,10 +711,18 @@ static struct cmd_results *cmd_move_in_direction(
 		return cmd_results_new(CMD_SUCCESS, NULL);
 	}
 	struct sway_workspace *old_ws = container->workspace;
+	struct sway_container *old_parent = container->parent;
 
 	if (!container_move_in_direction(container, direction)) {
 		// Container didn't move
 		return cmd_results_new(CMD_SUCCESS, NULL);
+	}
+
+	// clean-up, destroying parents if the container was the last child
+	if (old_parent) {
+		container_reap_empty(old_parent);
+	} else if (old_ws) {
+		workspace_consider_destroy(old_ws);
 	}
 
 	struct sway_workspace *new_ws = container->workspace;
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -133,6 +133,7 @@ static void container_move_to_container_from_direction(
 			}
 			container->width = container->height = 0;
 			container->width_fraction = container->height_fraction = 0;
+			workspace_squash(destination->workspace);
 		}
 		return;
 	}
@@ -145,6 +146,7 @@ static void container_move_to_container_from_direction(
 		container_insert_child(destination, container, index);
 		container->width = container->height = 0;
 		container->width_fraction = container->height_fraction = 0;
+		workspace_squash(destination->workspace);
 		return;
 	}
 
@@ -396,6 +398,7 @@ static bool container_move_in_direction(struct sway_container *container,
 		if (old_parent) {
 			container_reap_empty(old_parent);
 		}
+		workspace_squash(container->workspace);
 		return true;
 	}
 }
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -276,11 +276,12 @@ static void workspace_rejigger(struct sway_workspace *ws,
 		return;
 	}
 	container_detach(child);
-	workspace_wrap_children(ws);
+	struct sway_container *new_parent = workspace_wrap_children(ws);
 
 	int index =
 		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_UP ? 0 : 1;
 	workspace_insert_tiling(ws, child, index);
+	container_flatten(new_parent);
 	ws->layout =
 		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_RIGHT ?
 		L_HORIZ : L_VERT;
@@ -340,11 +341,8 @@ static bool container_move_in_direction(struct sway_container *container,
 							container_insert_child(current->parent, container,
 									index + (offs < 0 ? 0 : 1));
 						} else {
-							struct sway_workspace *ws = current->workspace;
-							workspace_insert_tiling(ws,
-								container_split(container,
-									output_get_default_layout(ws->output)),
-								index + (offs < 0 ? 0 : 1));
+							workspace_insert_tiling(current->workspace, container,
+									index + (offs < 0 ? 0 : 1));
 						}
 						return true;
 					}
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -395,6 +395,8 @@ static bool container_move_in_direction(struct sway_container *container,
 			workspace_insert_tiling(ancestor->workspace, container,
 					index + (offs < 0 ? 0 : 1));
 		}
+		ancestor->height = ancestor->width = 0;
+		ancestor->height_fraction = ancestor->width_fraction = 0;
 		if (old_parent) {
 			container_reap_empty(old_parent);
 		}
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -263,126 +263,141 @@ static void container_move_to_container(struct sway_container *container,
 	}
 }
 
-/* Takes one child, sets it aside, wraps the rest of the children in a new
- * container, switches the layout of the workspace, and drops the child back in.
- * In other words, rejigger it. */
-static void workspace_rejigger(struct sway_workspace *ws,
-		struct sway_container *child, enum wlr_direction move_dir) {
-	if (!child->parent && ws->tiling->length == 1) {
-		ws->layout =
-			move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_RIGHT ?
-			L_HORIZ : L_VERT;
-		workspace_update_representation(ws);
-		return;
+static bool container_move_to_next_output(struct sway_container *container,
+		struct sway_output *output, enum wlr_direction move_dir) {
+	struct sway_output *next_output =
+		output_get_in_direction(output, move_dir);
+	if (next_output) {
+		struct sway_workspace *ws = output_get_active_workspace(next_output);
+		if (!sway_assert(ws, "Expected output to have a workspace")) {
+			return false;
+		}
+		switch (container->fullscreen_mode) {
+		case FULLSCREEN_NONE:
+			container_move_to_workspace_from_direction(container, ws, move_dir);
+			return true;
+		case FULLSCREEN_WORKSPACE:
+			container_move_to_workspace(container, ws);
+			return true;
+		case FULLSCREEN_GLOBAL:
+			return false;
+		}
 	}
-	container_detach(child);
-	struct sway_container *new_parent = workspace_wrap_children(ws);
-
-	int index =
-		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_UP ? 0 : 1;
-	workspace_insert_tiling(ws, child, index);
-	container_flatten(new_parent);
-	ws->layout =
-		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_RIGHT ?
-		L_HORIZ : L_VERT;
-	workspace_update_representation(ws);
-	child->width = child->height = 0;
+	return false;
 }
 
 // Returns true if moved
 static bool container_move_in_direction(struct sway_container *container,
 		enum wlr_direction move_dir) {
 	// If moving a fullscreen view, only consider outputs
-	if (container->fullscreen_mode == FULLSCREEN_WORKSPACE) {
-		struct sway_output *new_output =
-			output_get_in_direction(container->workspace->output, move_dir);
-		if (!new_output) {
-			return false;
-		}
-		struct sway_workspace *ws = output_get_active_workspace(new_output);
-		if (!sway_assert(ws, "Expected output to have a workspace")) {
-			return false;
-		}
-		container_move_to_workspace(container, ws);
-		return true;
-	}
-	if (container->fullscreen_mode == FULLSCREEN_GLOBAL) {
+	switch (container->fullscreen_mode) {
+	case FULLSCREEN_NONE:
+		break;
+	case FULLSCREEN_WORKSPACE:
+		return container_move_to_next_output(container,
+				container->workspace->output, move_dir);
+	case FULLSCREEN_GLOBAL:
 		return false;
 	}
 
-	// Look for a suitable *container* sibling or parent.
-	// The below loop stops once we hit the workspace because current->parent
-	// is NULL for the topmost containers in a workspace.
-	struct sway_container *current = container;
 	int offs =
 		move_dir == WLR_DIRECTION_LEFT || move_dir == WLR_DIRECTION_UP ? -1 : 1;
+	int index = -1;
+	int	desired = -1;
+	list_t *siblings = NULL;
+	struct sway_container *target = NULL;
 
-	while (current) {
-		list_t *siblings = container_get_siblings(current);
-		if (siblings) {
-			enum sway_container_layout layout = container_parent_layout(current);
-			int index = list_find(siblings, current);
-			int desired = index + offs;
+	// Look for a suitable ancestor of the container to move within
+	struct sway_container *ancestor = NULL;
+	struct sway_container *current = container;
+	bool wrapped = false;
+	while (!ancestor) {
+		// Don't allow containers to move out of their
+		// fullscreen or floating parent
+		if (current->fullscreen_mode || container_is_floating(current)) {
+			return false;
+		}
 
-			// Don't allow containers to move out of their
-			// fullscreen or floating parent
-			if (current->fullscreen_mode || container_is_floating(current)) {
-				return false;
+		enum sway_container_layout parent_layout = container_parent_layout(current);
+		if (!is_parallel(parent_layout, move_dir)) {
+			if (!current->parent) {
+				// No parallel parent, so we reorient the workspace
+				current = workspace_wrap_children(current->workspace);
+				current->workspace->layout =
+					move_dir == WLR_DIRECTION_LEFT ||
+					move_dir == WLR_DIRECTION_RIGHT ?
+					L_HORIZ : L_VERT;
+				container->height = container->width = 0;
+				container->height_fraction = container->width_fraction = 0;
+				workspace_update_representation(current->workspace);
+				wrapped = true;
+			} else {
+				// Keep looking for a parallel parent
+				current = current->parent;
 			}
+			continue;
+		}
 
-			if (is_parallel(layout, move_dir)) {
-				if (desired == -1 || desired == siblings->length) {
-					if (current->parent == container->parent) {
-						current = current->parent;
-						continue;
-					} else {
-						// Reparenting
-						if (current->parent) {
-							container_insert_child(current->parent, container,
-									index + (offs < 0 ? 0 : 1));
-						} else {
-							workspace_insert_tiling(current->workspace, container,
-									index + (offs < 0 ? 0 : 1));
-						}
-						return true;
-					}
-				} else {
-					// Container can move within its siblings
-					container_move_to_container_from_direction(container,
-							siblings->items[desired], move_dir);
-					return true;
-				}
+		// Only scratchpad hidden containers don't have siblings
+		// so siblings != NULL here
+		siblings = container_get_siblings(current);
+		index = list_find(siblings, current);
+		desired = index + offs;
+		target = desired == -1 || desired == siblings->length ?
+				NULL : siblings->items[desired];
+
+		// If the move is simple we can complete it here early
+		if (current == container) {
+			if (target) {
+				// Container will swap with or descend into its neighbor
+				container_move_to_container_from_direction(container,
+						target, move_dir);
+				return true;
+			} else if (!container->parent) {
+				// Container is at workspace level so we move it to the
+				// next workspace if possible
+				return container_move_to_next_output(container,
+						current->workspace->output, move_dir);
+			} else {
+				// Container has escaped its immediate parallel parent
+				current = current->parent;
+				continue;
 			}
 		}
 
-		current = current->parent;
+		// We found a suitable ancestor, the loop will end
+		ancestor = current;
 	}
 
-	// Maybe rejigger the workspace
-	struct sway_workspace *ws = container->workspace;
-	if (ws) {
-		if (!is_parallel(ws->layout, move_dir)) {
-			workspace_rejigger(ws, container, move_dir);
-			return true;
-		} else if (ws->layout == L_TABBED || ws->layout == L_STACKED) {
-			workspace_rejigger(ws, container, move_dir);
-			return true;
+	if (target) {
+		// Container will move in with its cousin
+		container_move_to_container_from_direction(container,
+				target, move_dir);
+		return true;
+	} else if (!wrapped && !container->parent->parent &&
+			container->parent->children->length == 1) {
+		// Treat singleton children as if they are at workspace level like i3
+		// https://github.com/i3/i3/blob/1d9160f2d247dbaa83fb62f02fd7041dec767fc2/src/move.c#L367
+		return container_move_to_next_output(container,
+				ancestor->workspace->output, move_dir);
+	} else {
+		// Container will be promoted
+		struct sway_container *old_parent = container->parent;
+		if (ancestor->parent) {
+			// Container will move in with its parent
+			container_insert_child(ancestor->parent, container,
+					index + (offs < 0 ? 0 : 1));
+		} else {
+			// Container will move to workspace level,
+			// may be re-split by workspace_layout
+			workspace_insert_tiling(ancestor->workspace, container,
+					index + (offs < 0 ? 0 : 1));
 		}
-
-		// Try adjacent output
-		struct sway_output *output =
-			output_get_in_direction(container->workspace->output, move_dir);
-		if (output) {
-			struct sway_workspace *ws = output_get_active_workspace(output);
-			if (!sway_assert(ws, "Expected output to have a workspace")) {
-				return false;
-			}
-			container_move_to_workspace_from_direction(container, ws, move_dir);
-			return true;
+		if (old_parent) {
+			container_reap_empty(old_parent);
 		}
-		sway_log(SWAY_DEBUG, "Hit edge of output, nowhere else to go");
+		return true;
 	}
-	return false;
 }
 
 static struct cmd_results *cmd_move_to_scratchpad(void);
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1355,16 +1355,6 @@ void container_detach(struct sway_container *child) {
 		container_update_representation(old_parent);
 		node_set_dirty(&old_parent->node);
 	} else if (old_workspace) {
-		// We may have removed the last tiling child from the workspace. If the
-		// workspace layout was e.g. tabbed, then at this point it may be just
-		// H[]. So, reset it to the default (e.g. T[]) for next time.
-		// But if we are evacuating a workspace with only sticky floating
-		// containers, the workspace will already be detached from the output.
-		if (old_workspace->output && !old_workspace->tiling->length) {
-			old_workspace->layout =
-				output_get_default_layout(old_workspace->output);
-		}
-
 		workspace_update_representation(old_workspace);
 		node_set_dirty(&old_workspace->node);
 	}
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1410,6 +1410,28 @@ void container_replace(struct sway_container *container,
 
 struct sway_container *container_split(struct sway_container *child,
 		enum sway_container_layout layout) {
+	// i3 doesn't split singleton H/V containers
+	// https://github.com/i3/i3/blob/3cd1c45eba6de073bc4300eebb4e1cc1a0c4479a/src/tree.c#L354
+	if (child->parent || child->workspace) {
+		list_t *siblings = container_get_siblings(child);
+		if (siblings->length == 1) {
+			enum sway_container_layout current = container_parent_layout(child);
+			if (container_is_floating(child)) {
+				current = L_NONE;
+			}
+			if (current == L_HORIZ || current == L_VERT) {
+				if (child->parent) {
+					child->parent->layout = layout;
+					container_update_representation(child->parent);
+				} else {
+					child->workspace->layout = layout;
+					workspace_update_representation(child->workspace);
+				}
+				return child;
+			}
+		}
+	}
+
 	struct sway_seat *seat = input_manager_get_default_seat();
 	bool set_focus = (seat_get_focus(seat) == &child->node);
 
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1630,3 +1630,65 @@ bool container_is_sticky(struct sway_container *con) {
 bool container_is_sticky_or_child(struct sway_container *con) {
 	return container_is_sticky(container_toplevel_ancestor(con));
 }
+
+static bool is_parallel(enum sway_container_layout first,
+		enum sway_container_layout second) {
+	switch (first) {
+	case L_TABBED:
+	case L_HORIZ:
+		return second == L_TABBED || second == L_HORIZ;
+	case L_STACKED:
+	case L_VERT:
+		return second == L_STACKED || second == L_VERT;
+	default:
+		return false;
+	}
+}
+
+static bool container_is_squashable(struct sway_container *con,
+		struct sway_container *child) {
+	enum sway_container_layout gp_layout = container_parent_layout(con);
+	return (con->layout == L_HORIZ || con->layout == L_VERT) &&
+		(child->layout == L_HORIZ || child->layout == L_VERT) &&
+		!is_parallel(con->layout, child->layout) &&
+		is_parallel(gp_layout, child->layout);
+}
+
+static void container_squash_children(struct sway_container *con) {
+	for (int i = 0; i < con->children->length; i++) {
+		struct sway_container *child = con->children->items[i];
+		i += container_squash(child);
+	}
+}
+
+int container_squash(struct sway_container *con) {
+	if (!con->children) {
+		return 0;
+	}
+	if (con->children->length != 1) {
+		container_squash_children(con);
+		return 0;
+	}
+	struct sway_container *child = con->children->items[0];
+	int idx = container_sibling_index(con);
+	int change = 0;
+	if (container_is_squashable(con, child)) {
+		// con and child are a redundant H/V pair. Destroy them.
+		while (child->children->length) {
+			struct sway_container *current = child->children->items[0];
+			container_detach(current);
+			if (con->parent) {
+				container_insert_child(con->parent, current, idx);
+			} else {
+				workspace_insert_tiling_direct(con->workspace, current, idx);
+			}
+			change++;
+		}
+		// This will also destroy con because child was its only child
+		container_reap_empty(child);
+		change--;
+	} else {
+		container_squash_children(con);
+	}
+	return change;
+}
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -804,9 +804,10 @@ void container_set_floating(struct sway_container *container, bool enable) {
 			container->width = reference->width;
 			container->height = reference->height;
 		} else {
-			workspace_add_tiling(workspace, container);
-			container->width = workspace->width;
-			container->height = workspace->height;
+			struct sway_container *other =
+				workspace_add_tiling(workspace, container);
+			other->width = workspace->width;
+			other->height = workspace->height;
 		}
 		if (container->view) {
 			view_set_tiled(container->view, true);
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -627,6 +627,21 @@ struct sway_container *workspace_find_container(struct sway_workspace *ws,
 	return NULL;
 }
 
+static void set_workspace(struct sway_container *container, void *data) {
+	container->workspace = container->parent->workspace;
+}
+
+static void workspace_attach_tiling(struct sway_workspace *ws,
+		struct sway_container *con) {
+	list_add(ws->tiling, con);
+	con->workspace = ws;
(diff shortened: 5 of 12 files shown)


Xyene:
Wow, you beat me to the punch here -- I was going to make the same change tonight.

However, this isn't quite right (I know you haven't marked it as ready for review yet etc., but anyway). The check for the parent layout to be a [vh]split is actually necessary, because if you have a `T[app1]` and vsplit, this *should* result in `T[V[app1]]`, _not_ `V[app1]`.

Probably a good idea to leave a comment linking to the i3 source here, because this is pretty non-obvious.

RPigott:
Fixes #5755, #5754, Re-fixes #5157,  #4038.

I ended up going through a bunch of i3-compat issues but I think this matches i3 in the above cases. I don't think 5755 is related to 5754 just my fixes ended up on the same branch.

Yeah thanks @Xyene I forgot to copy that part. I added a comment as suggested.

RPigott:
For #4038 OP seems to describe i3 behavior as `T[1 2 3] -> H[T[1 2] T[3]]` but I can't reproduce this. In i3 I get `T[1 2 3] -> T[T[1 2] T[3]]`.

Xyene:
> In i3 I get `T[1 2 3]` -> `T[T[1 2] T[3]]`.

I'm not entirely sure how... `workspace_layout normal`, open two windows:

image

`layout tabbed`:

image

`move right`:

image

This state is not correct. Subsequent `move right`s from here toggle between `T[app1 T[app2]]` and `T[app1 app2]`. Instead, the transition should be `T[app1 app2]` &rarr; `H[T[app1] app2]`

RPigott:
Yeah this is wrong in that case.

> This only differs in i3 in that i3 always has the wrapping H for the workspace, whereas Sway often changes the workspace layout itself.

i3 doesn't always wrap workspaces, just when changinging the layout on a single container. Otherwise they change the workspace layout. The difference here is in the behavior of the layout command, which I guess I should also update to match:

https://github.com/i3/i3/blob/3cd1c45eba6de073bc4300eebb4e1cc1a0c4479a/src/con.c#L1817

The case I was trying to get at (and how I interpreted #4038) is

Begin: empty workspace

`layout tabbed`, open apps, `move right`. Then the workspace layout changes and I get the `T[1 2 3] -> T[T[1 2] T[3]]` behavior.

Xyene:
Getting better :+1: 

The case I described still happens with `workspace_layout tabbed`, though it has been fixed for `workspace_layout normal`.

RPigott:
> i3 workspace_layout doesn't change the layout of the root workspace container; it changes that of its first-level children.

Wow no kidding. In i3 `workspace_layout tabbed` doesn't make tabbed workspaces but `layout tabbed` on the workspace does.

The sway(5) man page will probably need to be updated with a proper description of `workspace_layout` too.

Xyene:
Yeah, it's tricky. Bit of a meta-comment, and you may already have a testing setup that works for you, but I've found the following to be very helpful in comparing i3 with Sway:

* script to dump layout, works on both i3 and Sway
```
$ cat ~/gettree.sh                                                                                                                                    
swaymsg -t get_tree|jq "def f: {name, type, layout, nodes: [.nodes[]|f]}; recurse(.nodes[])|select(.name == \"$1\")|f"
$ ~/gettree.sh 1   # get tree of workspace 1
```

* start Xephyr

```
$ Xephyr +iglx -br -ac -resizeable -noreset -keybd ephyr,,xkbmodel=evdev :2 &
```

* start i3 within Xephyr

```
$ unset XDG_RUNTIME_DIR && unset WAYLAND_DISPLAY && DISPLAY=:2 i3
```

* passthrough mode in Sway config so you can use conflicting keybinds within Xephyr i3

```
mode "passthrough" {
	bindsym $mod+p mode default
}

bindsym $mod+p mode passthrough
```

RPigott:
So far I've been running i3 in a separate session and sway embedded in gdb. It's a workable setup but could be improved. The i3 layout dump will certainly be handy, I had a basic one but it was still bit of a hassle to interpret. I need to test more i3 layouts

Quick take at matching i3 workspace_layout behavior like https://github.com/i3/i3/blob/e54e88b9e58579b406c5de94b68cb9ad3211eb4e/src/workspace.c#L899, but I'm guessing code will need to be moved somewhere more appropriate than view_map to match all the cases.

Xyene:
The latest commit certainly improves things. I'm generally a fan of keeping the workspace node layout intact (i3-style) over what Sway currently does; it makes debugging i3-compat issues easier.

With `workspace_layout tabbed`, starting with this layout and `move right`:

<details><summary>Start state</summary>

```json
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        },
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    }
  ]
}
```
</details>

<details><summary>Sway result</summary>

```json
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    },
    {
      "name": "xyene@lynwood: /code/sway-src/sway",
      "type": "con",
      "layout": "none",
      "nodes": []
    }
  ]
}
```
</details>

<details><summary>i3 result</summary>

```json
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/i3",
          "type": "con",
          "layout": "splith",
          "nodes": []
        }
      ]
    },
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/i3",
          "type": "con",
          "layout": "splith",
          "nodes": []
        }
      ]
    }
  ]
}
```
</details>

RPigott:
Hopefully improved. I Only tested it in a few cases and had #5761 applied on top.

I moved the change into workspace_add_tiling where it obviously belongs, but lots of places aren't happy to have workspace_add_tiling put their con as a child of a new container instead of the workspace they requested. So, most calls of workspace_add_tiling are changed to something like `con = workspace_add_tiling(ws, con)` and some of the following references are changed to modify the new con, if there is one.

Xyene:
That case works much better now. Quick testing: starting with a `T[app]` and doing repeated `splitv`, you end up with `V[V[V[...[V[app]]]]]`.

Xyene:
One more weird case with `workspace_layout tabbed` where a blank container was created:

<details><summary>Log</summary>

```
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1                                                                                                                                                                                                                                                        
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    }
  ]
}
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ swaymsg move up
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1                                                                                                                                                                                                                                                        
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": "xyene@lynwood: /code/sway-src/sway",
      "type": "con",
      "layout": "none",
      "nodes": []
    }
  ]
}
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ swaymsg move up
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1
{                                                                                                                                                          
   "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    },
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": null, <--------------------------------- this container is blank
          "type": "con",
          "layout": "splith",
          "nodes": []
        }
      ]
    }
  ]
}
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ swaymsg move up                                                                                    
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": "xyene@lynwood: /code/sway-src/sway",
      "type": "con",
      "layout": "none",
      "nodes": []
    },
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": null,
          "type": "con",
          "layout": "splith",
          "nodes": []
        }
      ]
    }
  ]
}
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ swaymsg move up                                                                                    
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    },
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": null,
          "type": "con",
          "layout": "splith",
          "nodes": []
        }
      ]
    }
  ]
}
```

</details>

RPigott:
> Quick testing: starting with a `T[app]` and doing repeated `splitv`, you end up with `V[V[V[...`

Oops, pushed the corrected version of that commit.

> One more weird case with workspace_layout tabbed where a blank container was created

I don't observe that with your #5761 applied. I guess it gets reaped.

Xyene:
Exciting, it's getting there :)

> I don't observe that with your #5761 applied. I guess it gets reaped.

You're right, oops.

It does seem that some extraneous containers are still being created with `workspace_layout tabbed`:

<details><summary>Log</summary>

```
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1                                                                                        
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        },
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    }
  ]
}
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ swaymsg move right
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1
{                                                                          
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    },
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    }
  ]
}
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ swaymsg move up
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1   
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    },
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": null,
          "type": "con",
          "layout": "splith",
          "nodes": [
            {
              "name": null,
              "type": "con",
              "layout": "tabbed",
              "nodes": [
                {
                  "name": "xyene@lynwood: /code/sway-src/sway",
                  "type": "con",
                  "layout": "none",
                  "nodes": []
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ swaymsg move up    
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1    
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    },
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": null,
          "type": "con",
          "layout": "splith",
          "nodes": [
            {
              "name": null,
              "type": "con",
              "layout": "tabbed",
              "nodes": [
                {
                  "name": null,
                  "type": "con",
                  "layout": "splith",
                  "nodes": [
                    {
                      "name": null,
                      "type": "con",
                      "layout": "tabbed",
                      "nodes": [
                        {
                          "name": "xyene@lynwood: /code/sway-src/sway",
                          "type": "con",
                          "layout": "none",
                          "nodes": []
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ swaymsg move up    
(RPigott-consplit) xyene@lynwood:/code/sway-src/sway$ ~/gettree.sh 1    
{
  "name": "1",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": "xyene@lynwood: /code/sway-src/sway",
          "type": "con",
          "layout": "none",
          "nodes": []
        }
      ]
    },
    {
      "name": null,
      "type": "con",
      "layout": "tabbed",
      "nodes": [
        {
          "name": null,
          "type": "con",
          "layout": "splith",
          "nodes": [
            {
              "name": null,
              "type": "con",
              "layout": "tabbed",
              "nodes": [
                {
                  "name": null,
                  "type": "con",
                  "layout": "splith",
                  "nodes": [
                    {
                      "name": null,
                      "type": "con",
                      "layout": "tabbed",
                      "nodes": [
                        {
                          "name": null,
                          "type": "con",
                          "layout": "splith",
                          "nodes": [
                            {
                              "name": null,
                              "type": "con",
                              "layout": "tabbed",
                              "nodes": [
                                {
                                  "name": "xyene@lynwood: /code/sway-src/sway",
                                  "type": "con",
                                  "layout": "none",
                                  "nodes": []
                                }
                              ]
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

</details>

RPigott:
The commit I removed was just one that had commented out some code. Now workspace_wrap_children now doesn't double wrap, that was creating some extra containers.

So here's an interesting layout with `workspace_layout tabbed`:

`begin` -> `H[T[app app*]]`
`move right` -> `H[T[app] T[app*]]` (matches i3)
`move up` -> `V[T[app*] H[T[app]]]` (matches i3, your scenario above)
`move right` -> `H[V[H[T[app]]] T[app*]]` (sway), `H[T[app] T[app*]]` (i3)

In the last move, though it's pretty unwieldy, the sway layout is the one that matches my mental model of the i3 tree. I'd guess it's special cased somewhere, but atm I'm not sure what i3 is doing here to get the simpler layout.

Xyene:
I'd guess it's this:

https://github.com/i3/i3/blob/1f0c628cde40cf87371481041b7197344e0417c6/src/tree.c#L638-L651

```c
/*
 * tree_flatten() removes pairs of redundant split containers, e.g.:
 *       [workspace, horizontal]
 *   [v-split]           [child3]
 *   [h-split]
 * [child1] [child2]
 * In this example, the v-split and h-split container are redundant.
 * Such a situation can be created by moving containers in a direction which is
 * not the orientation of their parent container. i3 needs to create a new
 * split container then and if you move containers this way multiple times,
 * redundant chains of split-containers can be the result.
 *
 */
void tree_flatten(Con *con) {
```

Called right at the end of `tree_move` on `croot`, which looks like, well, the root of the workspace: https://github.com/i3/i3/blob/1f0c628cde40cf87371481041b7197344e0417c6/src/move.c#L388

RPigott:
Hmm, thanks for the tip.

I implemented "workspace_squash" which is basically equivalent to i3's tree_flatten. I called it squash since container flatten was already taken to go the other way. Now I squash the workspace after moving and I get the same layout as i3 :tada:.

sway/tree/workspace.c
```diff
@@ -817,3 +833,10 @@ size_t workspace_num_tiling_views(struct sway_workspace *ws) {
 	workspace_for_each_container(ws, count_tiling_views, &count);
 	return count;
 }
+
+void workspace_squash(struct sway_workspace *workspace) {
+	for (int i = 0; i < workspace->tiling->length; i++) {
+		struct sway_container *child = workspace->tiling->items[i];
```
Xyene, line 839: Can this access out of bounds memory if `workspace->tiling` shrinks in length during `container_squash`?

Xyene: Separately, I'm not really sure whether we should be relying on the order containers appear in `workspace->tiling` here (for the skipping). It seems fragile if we don't already do this elsewhere.

sway/commands/move.c
```diff
@@ -366,21 +330,23 @@ static bool container_move_in_direction(struct sway_container *container,
 				}
 			}
 		}
-
-		current = current->parent;
+		if (!current->parent && !escaped) {
+			// No parallel parent, container is trying to escape it's workspace
```
Xyene, line 334:
```suggestion
			// No parallel parent, container is trying to escape its workspace
```

Xyene (commented):
Awesome! I can give this more thorough review / testing tonight. Could you rebase on master? A few conflicts have accumulated.

sway/tree/workspace.c
RPigott: I don't think workspace->tiling can shrink in length during the squash. container_squash only modifies the parent by adding containers to it, which is why I keep track of the number added. Containers added to the parent are not themselves eligible to be squashed because the recursion ends when it's done adding them and destroys the container pair. Basically each child has the opportunity to squash one pair of containers in its family, but it stops there.

Regardless, unless I'm sorely mistaken about C (possible), the condition statement is evaluated on each loop, so if length changes the condition value will reflect that and the loop will end rather then set `i >= workspace->length->tiling`. I expect that the length will grow because I am adding containers to it, so I actually need that for correctness. Otherwise the loop will end early.

Actually there is a bug though...  the reap_empty call should decrement change before returning because at that point con and child are certainly empty and will be removed. Atm I guess it skips the immediate sibling of children it successfully squashed.

RPigott:
Rebased. I tried to add some more helpful commit messages since they were previously not too helpful.

I absorbed your commit because I think it should apply before the workspace_layout changes so that the intermediate commits  don't produce weird null containers. Hope that's alright.

I'll continue testing it for a while.

include/sway/tree/workspace.h
```diff
@@ -110,14 +110,14 @@ void workspace_unwrap_children(struct sway_workspace *ws,
...
+struct sway_container *workspace_add_tiling(struct sway_workspace *workspace,
 		struct sway_container *con);
 
 void workspace_add_floating(struct sway_workspace *workspace,
 		struct sway_container *con);
 
-void workspace_insert_tiling(struct sway_workspace *workspace,
-		struct sway_container *con, int index);
+struct sway_container *workspace_insert_tiling(struct sway_workspace *workspace,
+		struct sway_container *con, int index, enum sway_container_layout layout);
```
RPigott, line 120: I added this layout argument because I needed a fucntion that performed the insert but without the potential split for workspace_layout.

Since this ended up only being used in one place with a layout other than config->default_layout I'd rather remove it and make that a different function.

Xyene:
> I absorbed your commit because I think it should apply before the workspace_layout changes so that the intermediate commits don't produce weird null containers. Hope that's alright.

That's great, I've closed the other PR.

One immediate thing I've noticed is that I can no longer move containers across workspaces when one workspace is on another monitor (with `move left/right` when at the edge of a monitor).

Xyene:
I believe the "resizing based off siblings" logic regressed a bit, namely in this layout:

<details><summary>Screenshots</summary>

image

`move down`

image

</details>

include/sway/tree/workspace.h
Xyene: Yeah, that sounds like a good idea.

include/sway/tree/workspace.h
```diff
@@ -134,4 +134,6 @@ size_t workspace_num_tiling_views(struct sway_workspace *ws);
 
 size_t workspace_num_sticky_containers(struct sway_workspace *ws);
 
+void workspace_squash(struct sway_workspace *workspace);
```
Xyene, line 137: Let's add a description of both what this and `container_flatten` does to the headers.

sway/tree/container.c
```diff
@@ -797,16 +797,16 @@ void container_set_floating(struct sway_container *container, bool enable) {
...
-			container->height = reference->height;
+		if (other) {
+			container_add_sibling(other, container, 1);
+			container->width = other->width;
+			container->height = other->height;
 		} else {
-			workspace_add_tiling(workspace, container);
-			container->width = workspace->width;
-			container->height = workspace->height;
+			other = workspace_add_tiling(workspace, container);
```
Xyene, line 807: Let's just declare a new variable for this here rather than taking over `reference`/`other` from above to mean a different thing depending on what branch you're in.

sway/tree/container.c
```diff
@@ -1609,3 +1619,68 @@ bool container_is_scratchpad_hidden_or_child(struct sway_container *con) {
 	con = container_toplevel_ancestor(con);
 	return con->scratchpad && !con->workspace;
 }
+
+static enum sway_container_layout layout_category(enum sway_container_layout layout) {
```
Xyene, line 1623: Nit: `container_...` naming scheme to match the rest of this file.

sway/tree/container.c
```diff
@@ -1609,3 +1619,68 @@ bool container_is_scratchpad_hidden_or_child(struct sway_container *con) {
...
+		struct sway_container *child = con->children->items[i];
+		i += container_squash(child);
+	}
+}
+
+// Returns the number of new containers added to the parent
+int container_squash(struct sway_container *con) {
+	int change = 0;
+	if (!con->children) {
+		return change;
```
Xyene, line 1660: Always 0?

sway/tree/container.c
```diff
@@ -1609,3 +1619,68 @@ bool container_is_scratchpad_hidden_or_child(struct sway_container *con) {
...
+	}
+	if (con->children->length != 1) {
+		container_squash_children(con);
+		return change;
```
Xyene, line 1664: Looks like this is also always 0; let's make that explicit here and move the `int change = 0;` declaration to L1668.

sway/tree/container.c
```diff
@@ -1609,3 +1619,68 @@ bool container_is_scratchpad_hidden_or_child(struct sway_container *con) {
...
+				workspace_insert_tiling(con->workspace, current, idx,
+						L_NONE);
+			}
+			change++;
+		}
+		container_reap_empty(child);
+		change--;
+		return change;
+	}
+	container_squash_children(con);
```
Xyene, line 1684: The `if` above ends with a `return`, so we can put this line in an `else` block, and deduplicate the return by removing the one on L1682 and keeping the one on L1685.

include/sway/tree/workspace.h
Xyene: Yeah, that sounds like a good idea.

sway/commands/layout.c
```diff
@@ -162,6 +162,13 @@ struct cmd_results *cmd_layout(int argc, char **argv) {
 			}
 			container->layout = new_layout;
 			container_update_representation(container);
+		} else if (config->handler_context.container) {
+			// i3 avoids changing workspace layouts with a new container
+			// https://github.com/i3/i3/blob/3cd1c45eba6de073bc4300eebb4e1cc1a0c4479a/src/con.c#L1817
+			container = workspace_wrap_children(workspace);
+			container->layout = new_layout;
+			container_update_representation(container);
+			workspace_update_representation(workspace);
```
Xyene, line 171: I think this call is implied in `container_update_representation`, is it not? https://github.com/swaywm/sway/blob/39328ca4e4b97bd76153869f6a0a20fef3c17112/sway/tree/container.c#L612-L616

sway/tree/container.c
RPigott: Seems good. Technically I could just declare it as `int change = -1` instead of ending with `change--` since I know in advance I will reap con. I'll just keep it as 0 though I think. I prefer that the value matches its meaning throughout.

RPigott:
> I believe the "resizing based off siblings" logic regressed a bit, namely in this layout:

Yes I noticed that too. I'll look into it.

> One immediate thing I've noticed is that I can no longer move containers across workspaces when one workspace is on another monitor (with move left/right when at the edge of a monitor).

Still works for me when the container is at workspace level, but obviously that's not as common anymore with all the layout changes. I'll check how i3 does it.

sway/tree/container.c
RPigott: I replaced this helper with a new helper function "is_parallel" similar to commands/move

RPigott:
Aight, I'm back.

I had some trouble getting the move behavior to match i3 within the existing control flow, so in the end I basically just looked at i3's tree_move and translated it to sway. So a lot of the `container_move_in_direction` function has changed, but it seems to work.

For the geometry I forgot to reset the container height/width after wrapping a container – that should be better now.

For the moving between outputs i3 does special case those lone children of workspace level splits, so I added that as well in commands/move.c:371.

sway/tree/container.c
```diff
@@ -1609,3 +1620,79 @@ bool container_is_scratchpad_hidden_or_child(struct sway_container *con) {
...
+		is_parallel(gp_layout, child_layout);
+}
+
+static void container_squash_children(struct sway_container *con) {
+	for (int i = 0; i < con->children->length; i++) {
+		struct sway_container *child = con->children->items[i];
+		i += container_squash(child);
+	}
+}
+/*
```
Xyene, line 1657: I think this description is good, but ought to be moved to the header file since that's where most of the docs are.

sway/tree/workspace.c
```diff
@@ -830,3 +851,14 @@ size_t workspace_num_sticky_containers(struct sway_workspace *ws) {
 	workspace_for_each_container(ws, count_sticky_containers, &count);
 	return count;
 }
+
+// workspace_squash is container_flatten in the reverse
+// direction. Instead of eliminating redundant splits that are
+// parents of the target container, it eliminates pairs of
+// redundant H/V splits that are children of the workspace.
```
Xyene, line 858: Same comment re: header here.

sway/tree/workspace.c
Xyene: You're right; I wasn't paying close enough attention.

Xyene (commented):
Thanks again for working on this -- it's getting better and better. I've left a couple of code style nits, and have started running on this branch. I haven't run into _major_ issues _yet_, so that's a good sign :)

Here's at least one small one, though: if you have a display layout like `| output 1 | output 2 |` with `T[H[app1*] app2]` on output 2, `move left` should result in `T[app1 app2]`. Instead, however, `app1` escapes onto output 1, as if the original layout had been `T[app1 app2]`.

Xyene:
Actually, one more thing: if you have a layout like `S[S[S[S[H[app]]]]]` and `move left`, only one level of the `S[` should be collapsed. Currently it collapses the entire chain.

RPigott:
Hmm yeah. I may be "squashing" too liberally. i3's move can return early and skip the squash sometimes, so I may need to move the squash call into specific branches of the move function.

RPigott:
> Actually, one more thing: if you have a layout like S[S[S[S[H[app]]]]] and move left, only one level of the S[ should be collapsed.

Wait are you sure? I just tried this in i3 and it collapsed all the S for me.

Still, I'll have to correct it incorrectly flattening the new ws if the con switches workspaces or unnecessarily flattening parents for simple moves.

Xyene:
> Wait are you sure? I just tried this in i3 and it collapsed all the S for me.

Uhhh... I _was_, but clearly I'm wrong. However, I did just confirm that if you had `H[V[H[V[app]]]]` and did `move left`, the result in i3 is `H[V[app]]`, but in this branch is just `H[app]`. This does seem like a case of over-squashing, as you say. (Also, I think this is not all running in the same transaction, or something: I saw flashes of `H[]` titlebars when moving the last child of a container out.)

The geometry issues also seem mostly fixed, though I was able to generate a "bad" geometry:

<details><summary>Screenshots</summary>

initial

image

`move down`

image

`move up`

image

`move down` again -- space ratio is wrong

image



</details>

RPigott:
> I did just confirm that if you had `H[V[H[V[app]]]]` and did move left, the result in i3 is `H[V[app]]`, but in this branch is just `H[app]`.

I just tried this in i3 and got `H[app]` as well...

<details>
<summary>
layouts
</summary>

```
$ gettree.sh 2 # HVHV[
{
  "name": "2",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "splitv",
      "nodes": [
        {
          "name": null,
          "type": "con",
          "layout": "splith",
          "nodes": [
            {
              "name": null,
              "type": "con",
              "layout": "splitv",
              "nodes": [
                {
                  "name": null,
                  "type": "con",
                  "layout": "splith",
                  "nodes": []
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
$ i3-msg move left
[{"success":true}]
$ gettree.sh 2 # H[
{
  "name": "2",
  "type": "workspace",
  "layout": "splith",
  "nodes": [
    {
      "name": null,
      "type": "con",
      "layout": "splith",
      "nodes": []
    }
  ]
} 
```
</details>

Could it be a config difference?

> `H[T[V[app1*]] T[app2]]` + `move right` is a no-op, though the result should be `H[T[app1] T[app2]]` [workspace_layout=tabbed]

Oops, accidentally caught by the wrong branch. I checked for workspace when I meant !parent – that probably messed up a lot of similar layouts.

I think this branch is getting much closer to a working state. I'll try to push smaller fixup commits and squash/rebase/whatever them later rather than re-writing half the branch each time. Hopefully that makes it easier to see what's changed.

Changing when it squashes doesn't fix any previously mentioned bugs I don't think, but should correct cases where you perform a simple move like a sibling swap within a deeply nested squashable layout. i3 leaves those layouts be if it didn't do anything that could introduce new splits during the move.

RPigott:
If it helps, my understanding of the `H[V[H[V[` case step-by-step is

`H[V[H[V[app]]]]` -`move left`-> `H[V[H[app V[]]]` -squash/flatten-> `H[app V[]]` -reap-> `H[app]`

You could get `H[V[app]]` with

`H[V[H[V[app]]]]` -`move up`-> `H[V[app H[V[]]]` -reap-> `H[V[app]]`

That's what I observe in both this branch and i3.

sway/commands/move.c
```diff
@@ -316,72 +306,88 @@ static bool container_move_in_direction(struct sway_container *container,
...
+			// Container will move to workspace level,
+			// may be re-split by workspace_layout
+			workspace_insert_tiling(current->workspace, container,
+					index + (offs < 0 ? 0 : 1));
 			return true;
 		}
-		sway_log(SWAY_DEBUG, "Hit edge of output, nowhere else to go");
 	}
+	// Unreachable
 	return false;
```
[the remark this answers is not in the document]
RPigott: So I added that there because the compiler gets upset that it reaches the end of the function without a return, and it does that because it doesn't know that since I changed the loop, "current" will actually never be null.

So the real problem here is I silenced the compiler warning without correcting the code. The loop condition should actually just be `while(true)` – the behavior is the same but it doesn't mislead readers, and the compiler, to think there is an invariant there.

Instead I think I'll move the declarations from the middle of the loop to before the loop, and the if block at the end to outside the loop. Once it gets to that final block it can't repeat anyway since every branch of the if has a return.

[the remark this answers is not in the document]
RPigott: Actually better yet I can just introduce a variable to be the loop invariant.

Xyene:
> I just tried this in i3 and got H[app] as well...

Hmm, I think you're probably right and I wasn't running the same command, but at any rate I can no longer generate weird behavior after your fixups.

> Oops, accidentally caught by the wrong branch. I checked for workspace when I meant !parent – that probably messed up a lot of similar layouts.

> I think this branch is getting much closer to a working state.

Agreed, good job :)

> I'll try to push smaller fixup commits and squash/rebase/whatever them later rather than re-writing half the branch each time. Hopefully that makes it easier to see what's changed.

This sounds fine, I was using the link under "force-pushed" to view the changes, but this sounds easier on your end for now.

At the moment, the only remaining issue I've seen is the geometry one from https://github.com/swaywm/sway/pull/5756#issuecomment-723710972. I'll continue running on this branch for a while (and with `workspace_layout tabbed`), just in case new things pop up.

RPigott:
> This sounds fine, I was using the link under "force-pushed" to view the changes, but this sounds easier on your end for now.

Wow I didn't know about that feature. Anyway, thanks a million for reviewing and testing stuff.

> At the moment, the only remaining issue I've seen are:
> - the geometry one from #5756 (comment), and
> - the transaction one from the same comment

The geometry one at least is an existing issue in sway, not introduced by this PR.

It's not super easy to get the H[T[V[ layout from your example on master but if you do set it up and try moving the second app down/up/down (it takes a few more movements than it does on this branch b/c of the changes) you can get the same stunted lower window.

I'm not sure about the issue with container titles, it may have been introduced by b149e46 since it now squashes before reaping any empty containers. Maybe reaping should be part of squash's duties.

RPigott:
I don't know why it included the wlroots 0.12 commit, that was a mistake. The only change is the new commit intended to address the geometry of promoted containers. I squashed the fixup commits from before.

sway/commands/move.c
```diff
@@ -263,138 +265,141 @@ static void container_move_to_container(struct sway_container *container,
...
+		enum sway_container_layout parent_layout = container_parent_layout(current);
+		if (!is_parallel(parent_layout, move_dir)) {
+			if (!current->parent) {
+				// No parallel parent, so we reorient the workspace
+				current = workspace_wrap_children(current->workspace);
+				current->workspace->layout =
+					move_dir == WLR_DIRECTION_LEFT ||
+					move_dir == WLR_DIRECTION_RIGHT ?
+					L_HORIZ : L_VERT;
+				container->height = container->width = 0;
```
Xyene, line 332: Somewhat an aside from the meat of this PR, but I think we should have something like a `container_reset_geometry` for this since these lines appear in a lot of other places.

sway/tree/container.c
```diff
@@ -1609,3 +1620,69 @@ bool container_is_scratchpad_hidden_or_child(struct sway_container *con) {
...
+	default:
+		return false;
+	}
+}
+
+static bool container_is_squashable(struct sway_container *con,
+		struct sway_container *child) {
+	enum sway_container_layout gp_layout = con->parent ?
+		con->parent->layout : con->workspace->layout;
+	enum sway_container_layout con_layout = con->layout;
```
Xyene, line 1642: Hmm, is there a reason why these are stored in variables here? `is_parallel` is pure, and the logic below mixes `con->layout` with `con_layout`, and `child->layout` with `child_layout`.

include/sway/tree/workspace.h
```diff
@@ -110,13 +110,16 @@ void workspace_unwrap_children(struct sway_workspace *ws,
...
 
-void workspace_add_tiling(struct sway_workspace *workspace,
+struct sway_container *workspace_add_tiling(struct sway_workspace *workspace,
 		struct sway_container *con);
 
 void workspace_add_floating(struct sway_workspace *workspace,
 		struct sway_container *con);
 
-void workspace_insert_tiling(struct sway_workspace *workspace,
+void workspace_insert_tiling_direct(struct sway_workspace *workspace,
```
Xyene, line 119: We should have a comment here explaining the difference in nuance between `workspace_insert_tiling_direct` and `workspace_insert_tiling`. I only understand it since I've been following along this thread :)

Xyene (commented):
> Wow I didn't know about that feature. 

It's very well-hidden :)

> Anyway, thanks a million for reviewing and testing stuff.

Thank _you_ for working on this -- I know it expanded way further than you probably imagined it would, but thanks for sticking through it.

> I'm not sure about the issue with container titles, it may have been introduced by b149e46 since it now squashes before reaping any empty containers. Maybe reaping should be part of squash's duties.

I _feel_ like some function the code is calling is calling `transaction_commit_dirty` when it shouldn't. Alternate explanation: some internal state (probably geometry!) is trashed prematurely, causing a transaction to time out. This is a snippet of the Sway log on executing a `move up` that resulted in a container title flash:

```
00:15:36.793 [sway/desktop/transaction.c:428] Transaction 0x55d3cbabb900 committing with 10 instructions
00:15:36.993 [sway/desktop/transaction.c:389] Transaction 0x55d3cbabb900 timed out (1 waiting)
00:15:36.993 [sway/desktop/transaction.c:281] Applying transaction 0x55d3cbabb900
00:15:36.994 [sway/desktop/transaction.c:428] Transaction 0x55d3cb99f880 committing with 8 instructions
00:15:36.994 [sway/desktop/transaction.c:281] Applying transaction 0x55d3cb99f880
```

That looks like 3 calls to `transaction_commit_dirty`, one of which times out after 200ms. That causes the title to show the incorrect result for 200ms. I think either not scheduling that failing transaction in the first place, or figuring out why it's hanging would be sufficient to address this issue.

> The only change is the new commit intended to address the geometry of promoted containers.

Awesome!

RPigott:
- Fixed a bug in the first commit where floating containers couldn't be split.
- Fixed a bug where parents of promoted containers could sometimes survive being emptied

Actually, @Xyene does that fixup commit fix the titles for you? I may have been getting lucky and those accidental containers were just getting reaped later.

Xyene:
> Actually, @Xyene does that fixup commit fix the titles for you? 

Yes, the titles are fine now and transactions are no longer timing out. Great!

sway/tree/container.c
```diff
@@ -1410,6 +1410,28 @@ void container_replace(struct sway_container *container,
 
 struct sway_container *container_split(struct sway_container *child,
 		enum sway_container_layout layout) {
+	// i3 doesn't split singleton H/V containers
+	// https://github.com/i3/i3/blob/3cd1c45eba6de073bc4300eebb4e1cc1a0c4479a/src/tree.c#L354
+	if (child->parent || child->workspace) {
+		list_t *siblings = container_get_siblings(child);
+		if (siblings->length == 1) {
+			enum sway_container_layout current = container_parent_layout(child);
+			if (container_is_floating(child)) {
```
RPigott, line 1410: If #5785 gets merged I won't need this.

RPigott:
Rebased on master again so the CI can build.

sway/commands/move.c
RPigott: You're probably right, but I think I'll hold off adding that in this PR.

Xyene:
So far, so good. I haven't encountered any major issues in the last 2 days, so that's encouraging.

Here's a small one, though:

* start with `T[H[app* app]]`
* `focus parent`
* `splith` (multiple times)

The resulting layout is something like `T[H[H[H[H[...H[app app]]]]]`. In i3, the `splith`s remain a no-op.

@RedSoxFan, do you have any thoughts on this PR? I think it's roughly in a state where it could be merged: I think that some corner cases might still be wrong, but by and large it's less finicky than current master.

ChrisLane:
Just tested this branch, it seems to fix https://github.com/nwg-piotr/autotiling/issues/19

RPigott:
> The resulting layout is something like T[H[H[H[H[...H[app app]]]]]. In i3, the spliths remain a no-op.

Hmm I'm not able to reproduce. After the focus parent the H split is focused, right? In this branch and in i3 I get T[H[H[app app]*]] after the split and then all further splith's are no-ops. Actually it seems if you focus parent again in both this branch and i3 you can get more H[ by splith'ing the topmost H.

Something is fishy with workspace layouts still. Without setting `workspace_layout` in my config if I `layout tabbed` an empty workspace in i3 and open a new window I actually get `T[T[app]]` as if `workspace_layout tabbed` was set. I made several mistakes trying to create `T[H[app app]]` in i3 because of that.

Xyene:
> Hmm I'm not able to reproduce. After the focus parent the H split is focused, right? 

Yes, though oddly I haven't been able to reproduce this since. What I feel (since my memory isn't that good) was happening was the new container was getting focus, as opposed to the old one keeping it. At any rate, without being able to reproduce this, it's not worth worrying about now.

> Without setting workspace_layout in my config if I layout tabbed an empty workspace in i3 and open a new window I actually get T[T[app]] as if workspace_layout tabbed was set.

But... not quite? On i3 with `workspace_layout tabbed` and opening a single application, you get `H[T[app]]` (the workspace node layout is `splith`). But if you `layout tabbed` before opening the application (still `workspace_layout tabbed`), you get `T[T[app]]`. This matches the behavior in this PR, AFAICT.

Xyene:
I've been running this for the past week with no major issues. In general, I'd be comfortable with merging this before we get an authoritative response regarding `layout tabbed` on an empty workspace &mdash; that's a pretty niche scenario. Probably worth another round of rebasing to get this in a mergeable state.

On the "of some concern" list, I've noticed Firefox (84.0b3) geometries are wrong when floated out of the tiling layer. I'm not sure if this is caused by this patch (the geometry resetting?) or by a regression in Firefox, but at least I wasn't able to (easily) reproduce this issue in Sway master.

<details><summary>Screenshot</summary>

image

</details>

RPigott:
> On the "of some concern" list, I've noticed Firefox (84.0b3) geometries are wrong when floated out of the tiling layer.

I tried FF83 and 84 but it doesn't seem to happen for me. Is it reproducible everytime for you? Is there a chance it's fixed by #5812? I think I'm running that one too.

Xyene:
> Is it reproducible everytime for you? Is there a chance it's fixed by #5812? I think I'm running that one too.

Unfortunately yes, and partially.

#5812 does seem to prevent the issue from being permanent (once Firefox is in a bad state, it sends another configure a bit later, which after #5812 fixes the borders). Here's a [video](https://tbrindus.ca/files/dist/sway/5756-1.mp4).

There are at least two issues:

* When the window is floated, the border is drawn at the wrong position for 200ms (presumably it's waiting for Firefox to ack a configure?), and Firefox never does(?) so the transaction times out. With #5812, the original transaction still times out, but Firefox sends another one and "fixes" itself.

```
00:06:55.442 [sway/desktop/transaction.c:505] Transaction 0x5603c03bbae0 is ready
00:06:55.442 [sway/desktop/transaction.c:281] Applying transaction 0x5603c03bbae0
00:06:55.443 [sway/desktop/transaction.c:428] Transaction 0x5603c03d72b0 committing with 1 instructions
00:06:55.443 [sway/input/cursor.c:901] denying request to set cursor from unfocused client
00:06:55.443 [sway/input/cursor.c:901] denying request to set cursor from unfocused client
00:06:55.643 [sway/desktop/transaction.c:389] Transaction 0x5603c03d72b0 timed out (1 waiting)
00:06:55.643 [sway/desktop/transaction.c:281] Applying transaction 0x5603c03d72b0
```

* When the window is un-floated, there is a frame where the split indicator border is drawn incorrectly (my indicator color is red)

image

This is frame 118/550 of the video. I suspect this is happening as the floated container is being split on being added back into the tiling layer, prior to being squashed again.

RPigott:
Latest commit fixes #5783 but workspace {prev,next}_on_output will need the same treatment. Normally I'd make a different PR but it's just easier to have one big i3 compat branch for testing, especially when they might conflict. It should be easy enough to split later if we want.

As for firefox, it turns out #5812 was pretty buggy and I think it's related. I'm not sure yet how #5848 and #5857 will be resolved but I have a hunch their closure might address the firefox issue as well, or at least make it easier to debug.

Xyene (approved):
In the interest of getting more eyes on this and preventing this branch getting stale, I'll go ahead and merge this now. Irrespective of any outstanding bugs this is already a great improvement over the status quo. Thanks for working on this!

PR #5764, opened 2020-10-24
container: Fix NULL pointer dereference

### LINKED ISSUE #5762: Sway crashes when unplugging an output with only sticky, floating container
- **Sway Version:**
  - sway version 1.5-d06c1ac1 (Oct 24 2020, branch 'HEAD')

- **Debug Log:**
  - https://gist.github.com/mwenzkowski/cc0400d92e17bf8dd56652ee9d5e0aaa#file-debug-log

- **Configuration File:**
  - Default configuration file
 
- **Stack Trace:**
  - https://gist.github.com/mwenzkowski/cc0400d92e17bf8dd56652ee9d5e0aaa#file-stack-trace

- **Description:**
  - After starting sway on my laptop with an external monitor attached, I opened a terminal on the external monitor, made it floating and sticky and unplugged the external monitor, then sway chrashed.

I have begun investigating this issue and will prepare a patch myself.

mwenzkowski:
Reset the workspace layout to the output's default only if the workspace
is actually attached to an output.

Fixes #5762

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1358,7 +1358,9 @@ void container_detach(struct sway_container *child) {
 		// We may have removed the last tiling child from the workspace. If the
 		// workspace layout was e.g. tabbed, then at this point it may be just
 		// H[]. So, reset it to the default (e.g. T[]) for next time.
-		if (!old_workspace->tiling->length) {
+		// But if we are evacuating a workspace with only sticky floating
+		// containers, the workspace will already be detached from the output.
+		if (old_workspace->output && !old_workspace->tiling->length) {
 			old_workspace->layout =
 				output_get_default_layout(old_workspace->output);
 		}


sway/tree/container.c
```diff
@@ -1358,7 +1358,7 @@ void container_detach(struct sway_container *child) {
 		// We may have removed the last tiling child from the workspace. If the
 		// workspace layout was e.g. tabbed, then at this point it may be just
 		// H[]. So, reset it to the default (e.g. T[]) for next time.
-		if (!old_workspace->tiling->length) {
+		if (old_workspace->output && !old_workspace->tiling->length) {
```
RedSoxFan, line 1361: Can you please add to the comment above something like "If we are evacuating a workspace with only sticky floating containers, the workspace will already be detached from the output."?

mwenzkowski: Done.

RedSoxFan (approved):

RedSoxFan:
Thanks!

PR #5765, opened 2020-10-25
Enable variable override using /etc/sway/config-vars.d/

bisco2:
We use this in Debian to allow other packages to override default variables- I thought that this maybe also helps other distributors. (Based on this change it would be great to also define the wallpaper path using a variable defined in the variables section)

### DIFF
--- a/config.in
+++ b/config.in
@@ -20,6 +20,11 @@ set $term alacritty
 # on the original workspace that the command was run on.
 set $menu dmenu_path | dmenu | xargs swaymsg exec --
 
+#### Variable override
+#
+# Files in this directory can be used to override variables
+include /etc/sway/config-vars.d/*
+
 ### Output configuration
 #
 # Default wallpaper (more resolutions are available in @datadir@/backgrounds/sway/)


emersion:
I'd prefer something like `include /etc/sway/config.d/*`. But I'm not sure it makes sense in general, and it's easy for distributions to append it to the default config.

>Based on this change it would be great to also define the wallpaper path using a variable defined in the variables section

If you want to customize the Sway config, I think it would make more sense to edit it instead of hiding variable changes in another file. The default config aims to be simple to read for new users, if they need to spot the include and navigate multiple files to change the default wallpaper they'll have an harder time understanding what happens.

bisco2:
> I'd prefer something like `include /etc/sway/config.d/*`. But I'm not sure it makes sense in general, and it's easy for distributions to append it to the default config.

Yeah, we are doing this in Debian by patching `config.in` - but it would be simpler for us if the change was merged upstream. But I can understand if its not a priority.

> 
> > Based on this change it would be great to also define the wallpaper path using a variable defined in the variables section
> 
> If you want to customize the Sway config, I think it would make more sense to edit it instead of hiding variable changes in another file. The default config aims to be simple to read for new users, if they need to spot the include and navigate multiple files to change the default wallpaper they'll have an harder time understanding what happens.

Our use case would be to have various packages put their configuration snippets in the `/etc/sway/config-vars.d/` folder. For example the Debian `desktop-base` package contains wallpapers - the package could set the default Debian wallpaper as sway desktop background by overriding the `$background` variable (*if* it is installed- if not, the default would be to use upstreams sway background).
But if we patch the default config to allow the variable override we can also introduce a `$background` variable ;)

RPigott:
Am I missing something? config.in already contains `include @sysconfdir@/sway/config.d/*`.

Arch Linux also uses this to ship distro specific configuration
```
$ pacman -Flq sway | grep config.d
etc/sway/config.d/
etc/sway/config.d/50-systemd-user.conf
```

Can't debian ship a 50-debian.conf or similar? You can override the default background by just putting in another `output * bg` command – the later one one will override the previous one in the main config file. Feel free to make a $background variable for it and ship your own defaults in a `20-debian-defaults.conf` if you want.

emersion:
> config.in already contains include @sysconfdir@/sway/config.d/*.

Eh, indeed, I missed that.

>You can override the default background by just putting in another output * bg command – the later one one will override the previous one in the main config file. 

Yeah, and variables are evaluated at run-time, so assigning to the variable after the `output * bg` command should work too.

Closing because `@sysconfdir@/sway/config.d` can be used instead of introducing another default dir.

bisco2:
> Am I missing something? config.in already contains include @sysconfdir@/sway/config.d/*

The PR was about overriding variables before they get used in the configuration

> Yeah, and variables are evaluated at run-time, so assigning to the variable after the output * bg command should work too.

It doesn't - overriding existing variables in `@sysconfdir@/sway/config.d` has no effect on configuration settings defined before.

But yeah, I can see how it is better to have a simple configuration file. I'll find a way to hack around this ;) Sorry for the noise

RPigott:
> The PR was about overriding variables before they get used in the configuration

The only vars in the default config are $mod, $term, and $menu. I don't think you could have a good case to override any of them. Really all three of these vars are fundamentally user choice. Users _will_ want to edit them. I don't see how moving them to another directory benefits anyone, since it just wouldn't make sense for a package to override their values.

>>Yeah, and variables are evaluated at run-time, so assigning to the variable after the output * bg command should work too.

> It doesn't - overriding existing variables in @sysconfdir@/sway/config.d has no effect on configuration settings defined before.

Variables can be made to be evaluated at runtime with the `$$var` syntax. However, I want to reiterate that sway doesn't spawn your bg until it has completed reading the config. Using a second `output bg * <blah>` command has exactly the same effect. Your backgrounds package can just ship a file with that command to `/etc/sway/config.d`.

PR #5766, opened 2020-10-25
output: evacuate sticky containers only if new output has a workspace

mwenzkowski:
Sticky floating containers on an otherwise empty workspace can only be
evacuated if the new output has an active workspace. The noop output may
not have one and in that case we have to move the whole workspace to the
new output.

### DIFF
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -152,21 +152,18 @@ void output_enable(struct sway_output *output) {
 	arrange_root();
 }
 
-static void evacuate_sticky(struct sway_workspace *old_ws,
-		struct sway_output *new_output) {
-	struct sway_workspace *new_ws = output_get_active_workspace(new_output);
-	if (!sway_assert(new_ws, "New output does not have a workspace")) {
-		return;
+static void evacuate_sticky(struct sway_container *con, void *data) {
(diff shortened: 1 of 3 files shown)


sway/tree/output.c
```diff
@@ -195,18 +195,25 @@ static void output_evacuate(struct sway_output *output) {
...
+			output_get_active_workspace(new_output);
+
+		// workspace_is_empty() only checks that no tiling containers and no
+		// non-sticky floating containers exist
+		bool ws_is_empty = workspace_is_empty(workspace);
+
+		// If the workspace has only sticky floating containers and the new
+		// output has an active workspace (the noop output may not have one),
+		// move the containers to it.
+		if (ws_is_empty && workspace->floating->length && new_output_ws) {
```
Xyene, line 208: It's a bit nonobvious (though correct) that if `ws_is_empty` is true here and `workspace->floating->length` is nonzero, then all of those floating views are sticky views.

I think introducing a `workspace_num_sticky_views` (to complement `workspace_num_tiling_views`) in `workspace.c` and using that here would make the intent more obvious.

Xyene (approved):
LGTM, thanks!

PR #5767, opened 2020-10-25
input: fully change focus when scrolling tabs/stacks

Xyene:
When scrolling on a tabbed/stacked container, i3 focuses its
inactive-focused child. Sway does the same, but then resets the
focus to whatever was focused previously.

Ref https://github.com/i3/i3/blob/e5992eed163179f5cd2715c2c212d3d757f04b31/src/click.c#L207-L219

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -704,19 +704,14 @@ static void handle_pointer_axis(struct sway_seat *seat,
 			} else if (desired >= siblings->length) {
 				desired = siblings->length - 1;
 			}
-			struct sway_node *old_focus = seat_get_focus(seat);
+
 			struct sway_container *new_sibling_con = siblings->items[desired];
 			struct sway_node *new_sibling = &new_sibling_con->node;
 			struct sway_node *new_focus =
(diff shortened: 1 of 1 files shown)


sway/input/seatop_default.c
```diff
@@ -704,19 +704,14 @@ static void handle_pointer_axis(struct sway_seat *seat,
...
+
 			struct sway_container *new_sibling_con = siblings->items[desired];
 			struct sway_node *new_sibling = &new_sibling_con->node;
 			struct sway_node *new_focus =
 				seat_get_focus_inactive(seat, new_sibling);
-			if (node_has_ancestor(old_focus, tabcontainer)) {
-				seat_set_focus(seat, new_focus);
-			} else {
-				// Scrolling when focus is not in the tabbed container at all
-				seat_set_raw_focus(seat, new_sibling);
```
Xyene, line 716: I don't actually understand what this line was doing -- I believe it was redundant due to `new_focus` below being a descendant of `new_sibling`?

RedSoxFan:
Thanks!

PR #5770, opened 2020-10-27
In container_split, set a floating container's view to tiled

dimitris-triantafyllidis:
Currently, when a floating container with a view is split and children are added to it, the new views are rendered as tiled, while the first view stays in floating style.

Here this is addressed by setting the view to tiled as soon as the container is split, by duplicating the "view part" of the logic in container_set_floating(..., false). Since the new container of the view is no longer considered floating, it makes sense to set the view to tiling at this point.

The view would have to be set back to floating if it was possible to "unsplit" the container (see "split none" when it gets implemented).

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1413,6 +1413,13 @@ struct sway_container *container_split(struct sway_container *child,
 	struct sway_seat *seat = input_manager_get_default_seat();
 	bool set_focus = (seat_get_focus(seat) == &child->node);
 
+	if (container_is_floating(child) && child->view) {
+		view_set_tiled(child->view, true);
+		if (child->view->using_csd) {
+			child->border = child->saved_border;
+		}
+	}
+
 	struct sway_container *cont = container_create(NULL);
 	cont->width = child->width;
 	cont->height = child->height;


Xyene:
How are you adding children to an already-floating container?

dimitris-triantafyllidis:
You can add children containers if you split the floating container first.
Say you launch chromium and make the container floating. Then split the floating container and launch a second chromium window in this floating container. The result is currently this:
current_master
You now have a single parent floating container in the workspace with two children tiled within it, but one view (the first) is still rendered in the "floating style", while the second one is rendered in the "tiled style".

After this commit, the result is this:
after_commit

Which I think is the appropriate one.

Xyene:
Ah, I understand now and agree that this is more reasonable behavior.

The borders don't seem to entirely work, though (not sure if this was pre-existing or caused by this PR):

image

The leftmost Chrome instance here is focused, but has no border. The rightmost, which was opened after the original view was floated, does have a border. (You can see how the UIs don't line up between the two instances.)

dimitris-triantafyllidis:
In my config I set borders with criteria too and I did not observe this. But now that I removed the criteria I get the same result as you, but the border is lost as soon the first chromium is floated, before the split and so before the code in this PR is executed. So I think this is actually related to  #5066.

Xyene (approved):
Ah, that explains why the borders were fine in your screenshots. In that case, LGTM -- thanks!

PR #5771, opened 2020-10-27
Fix some grammar in Readme to sound more formal

codebam:


emersion:
I'm comfortable with the current non-formal wording.

TheAvidDev:
I like the more relaxed wording as well, not to mention that there's a few other places in the README that still use conjunctions.

PR #5774, opened 2020-10-28
output: Revert implementation of evacuate_sticky()

mwenzkowski:
The function evacuate_sticky() was changed in commit 32788a93 to be used
by workspace_for_each_container() to make the code more readable. But I
overlooked that it is not safe to use workspace_for_each_container() to
remove container from a workspace. This commit restores the previous
implementation for evacuate_sticky().

### DIFF
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -152,18 +152,21 @@ void output_enable(struct sway_output *output) {
 	arrange_root();
 }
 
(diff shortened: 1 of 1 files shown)


Xyene (approved):
Ah yes, we don't have `_for_each_safe` variants of this function. This LGTM, thanks.

PR #5776, opened 2020-10-29
swaybar: support StatusNotifierItem ToolTip

nmschulte:
This is some work I started while investigating tool-tip support for d-bus menu in #5161.

**The updates to the d-bus message parsing logic are the most interesting thing here.**  I've not tested since rebasing but the branch builds and runs but I know it does not fully function because the paint logic is naive.  Someone more familiar with the Wayland spec or wlroots should be able to nudge this in the right direction; I think it requires an additional surface like the tray uses for popups.

As well, this isn't the kind of support I was after; not many applications use SNI's ToolTip, and instead I just want to update #5161 to paint the application names and/or (icon) descriptions in bordered boxes when hovering on the icons.

### DIFF
--- a/include/swaybar/tray/item.h
+++ b/include/swaybar/tray/item.h
@@ -52,6 +52,9 @@ struct swaybar_sni {
 	bool item_is_menu;
 	char *menu;
 	struct swaybar_sni_tool_tip *tool_tip;
+	char *title;
+	char *category;
+	char *id;
 	char *icon_theme_path; // non-standard KDE property
 
 	struct wl_list slots; // swaybar_sni_slot::link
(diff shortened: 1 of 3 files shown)


emersion:
Converting to drafts since this seems to still be WIP (judging from the last commit message). When it's ready let us know.

tinywrkb:
https://gitlab.freedesktop.org/xdg/xdg-specs/-/issues/84

nmschulte:
> https://gitlab.freedesktop.org/xdg/xdg-specs/-/issues/84

I'm happy to report there is much recent movement on this upstream issue.

PR #5779, opened 2020-10-31
input: remove motion deltas from seatop callbacks

Xyene:
Straightforward cleanup, they haven't been used for a while.


emersion (approved):
Thanks!

PR #5780, opened 2020-11-01
commands/focus: force container warp when fulfilling `focus mode_toggle`

### LINKED ISSUE #5772: mouse_warping container does't take effect when toggling focus between floating and tiled layers
- **Sway Version:**
  - `sway version v1.5`

- **Description:**

Based on the descriptions of the `mouse_warping` and `focus mode_toggle` commands, it would seem to reason that changing container focus via `focus mode_toggle` should trigger a mouse warp when `mouse_warping container` is set. Currently, the mouse cursor remains stationary.

```
mouse_warping output|container|none
  If output is specified, the mouse will be moved to new outputs as you move focus between them. If con‐
  tainer is specified, the mouse will be moved to the middle of the container on switch. Default is out‐
  put.

focus mode_toggle
  Moves focus between the floating and tiled layers.
```

Xyene:
This commit switches focusing behavior to force a warp when executing
`focus mode_toggle`.

Fixes #5772.

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -268,7 +268,16 @@ static struct cmd_results *focus_mode(struct sway_workspace *ws,
 	}
 	if (new_focus) {
 		seat_set_focus_container(seat, new_focus);
-		seat_consider_warp_to_focus(seat);
+
+		// If we're on the floating layer and the floating container area
+		// overlaps the position on the tiling layer that would be warped to,
+		// `seat_consider_warp_to_focus` would decide not to warp, but we need
+		// to anyway.
+		if (config->mouse_warping == WARP_CONTAINER) {
+			cursor_warp_to_container(seat->cursor, new_focus, true);
+		} else {
+			seat_consider_warp_to_focus(seat);
+		}
 	} else {
 		return cmd_results_new(CMD_FAILURE,
 				"Failed to find a %s container in workspace",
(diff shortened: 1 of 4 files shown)


sway/commands/focus.c
```diff
@@ -268,7 +268,16 @@ static struct cmd_results *focus_mode(struct sway_workspace *ws,
...
 	if (new_focus) {
 		seat_set_focus_container(seat, new_focus);
-		seat_consider_warp_to_focus(seat);
+
+		// If we're on the floating layer and the floating container area
+		// overlaps the position on the tiling layer that would be warped to,
+		// `seat_consider_warp_to_focus` would decide not to warp, but we need
+		// to anyway.
+		if (config->mouse_warping == WARP_CONTAINER) {
+			cursor_warp_to_container(seat->cursor, new_focus, true);
```
Xyene, line 277: This is the only place where what `seat_consider_warp_to_focus` is "wrong", so I figured it's better to special-case here than plumb a new parameter through all calls to `seat_consider_warp_to_focus` everywhere.

Emantor (approved):

RedSoxFan (commented):
`seat_consider_warp_to_focus` contains the following after the call to `cursor_warp_to_container` that I think may need to be kept.
```c
if (seat->cursor->hidden){
    cursor_unhide(seat->cursor);
}
```

Xyene:
Good point, I've pushed that into `cursor_warp_to_container` and `cursor_warp_to_workspace`.

RedSoxFan (approved):

RedSoxFan:
Thanks!

PR #5786, opened 2020-11-03
Add border class support

TheAvidDev:


TheAvidDev:
So sorry, wrong fork...

PR #5787, opened 2020-11-03
tree/container: introduce `container_is_sticky`

Xyene:
To query whether a container is sticky, checking `con->is_sticky` is
insufficient. `container_is_floating_or_child` must also return true;
this led to a lot of repetition.

This commit introduces a `container_is_sticky` function, and switches
all stickiness checks to use it. (Including ones where the container is
already known to be floating, for consistency.)

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -1244,13 +1244,9 @@ bool view_is_visible(struct sway_view *view) {
 			return false;
 		}
 	}
-	// Determine if view is nested inside a floating container which is sticky
-	struct sway_container *floater = view->container;
-	while (floater->parent) {
-		floater = floater->parent;
-	}
-	bool is_sticky = container_is_floating(floater) && floater->is_sticky;
-	if (!is_sticky && workspace && !workspace_is_visible(workspace)) {
+
+	if (!container_is_sticky_or_child(view->container) && workspace &&
+			!workspace_is_visible(workspace)) {
 		return false;
 	}
 	// Check view isn't in a tabbed or stacked container on an inactive tab
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -498,7 +498,7 @@ bool workspace_is_empty(struct sway_workspace *ws) {
 	// Sticky views are not considered to be part of this workspace
 	for (int i = 0; i < ws->floating->length; ++i) {
 		struct sway_container *floater = ws->floating->items[i];
-		if (!floater->is_sticky) {
+		if (!container_is_sticky(floater)) {
 			return false;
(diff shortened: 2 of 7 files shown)


sway/tree/view.c
```diff
@@ -1244,13 +1244,9 @@ bool view_is_visible(struct sway_view *view) {
 			return false;
 		}
 	}
-	// Determine if view is nested inside a floating container which is sticky
-	struct sway_container *floater = view->container;
-	while (floater->parent) {
-		floater = floater->parent;
```
Xyene, line 1250: This bit looks like an ad-hoc reimplementation of `container_is_floating_or_child`, so switching it to `container_is_sticky` seems valid.

sway/tree/workspace.c
```diff
@@ -819,7 +819,7 @@ size_t workspace_num_tiling_views(struct sway_workspace *ws) {
 }
 
 static void count_sticky_containers(struct sway_container *con, void *data) {
-	if (container_is_floating(con) && con->is_sticky) {
```
Xyene, line 822: This was using `container_is_floating` rather than `container_is_floating_or_child`, but the overall result should remain the same. (The result of `workspace_num_sticky_containers` is currently only compared against 0.)

RedSoxFan: Although this is only used to check for the existence of sticky containers and wouldn't cause any issues, it does make it so the function name(s) are misleading.

`if (container_is_sticky(con) && !con->parent)` would be a logical equivalent and would allow for the naming to remain the same.

RedSoxFan (commented):
I think this should be `container_is_sticky_or_child`.

Technically this and `container_is_floating_or_child` should be `or_descendant`, but I'd be fine with `or_child` to match the existing naming scheme.

sway/tree/workspace.c
Xyene: I've introduced an `container_is_sticky_or_child` variant to keep the semantics of existing code more clear.

Xyene:
Something I don't fully understand is that it seems like all children of a sticky container also have their `is_sticky` set to true. This leads to two ways to write `container_is_sticky_or_child`:

```c
bool container_is_sticky_or_child(struct sway_container *con) {
	return container_is_sticky(container_toplevel_ancestor(con));
}
```

or

```c
bool container_is_sticky_or_child(struct sway_container *con) {
	return con->is_sticky && container_is_floating_or_child(con);
}
```

Though the former seems the most reasonable option to me.

RedSoxFan (approved):

RedSoxFan:
Thanks!

PR #5788, opened 2020-11-03
Drop gtk-primary-selection support

ifreund:
See: https://github.com/swaywm/wlroots/issues/2421 https://github.com/swaywm/wlroots/pull/2460

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -14,7 +14,6 @@
 #include <wlr/types/wlr_data_control_v1.h>
 #include <wlr/types/wlr_export_dmabuf_v1.h>
 #include <wlr/types/wlr_gamma_control_v1.h>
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

luispabon:
This PR broke copying URLs from firefox's URL bar.

Emantor:
> This PR broke copying URLs from firefox's URL bar.

Unless your distribution has gtk3 version 3.24.24, this will remain broken. This will break lots of users using some kind of PPA-like infrastructure but having the main system on a "stable" version, i.e. Ubuntu, Debian, …

Nothing to be done here though, Sway depends on bleeding edge libraries on master.

luispabon:
Unfortunately not, Ubuntu 20.04 ships with  3.24.20 :/

PR #5793, opened 2020-11-04
input/seatop_default: consider fullscreen views to have no edges

### LINKED ISSUE #5792: Pointer can leave fullscreen container off bounded edge
This has been around for a while, and is still present in master.

If an app is fullscreened (from WM), and the mouse is moved all the way to the top of its output (bounded at the top -- no second output is present above it) and clicked, sometimes the application will receive a pointer leave event.

```
[2641171.164] wl_pointer@3.motion(286023954, 230.992188, 0.000000)
[2641171.179] wl_pointer@3.frame()
[2641174.174] wl_pointer@3.frame()
[2641176.178] wl_pointer@3.motion(286023959, 231.402344, 0.000000)
[2641176.190] wl_pointer@3.frame()
[2641180.287] wl_pointer@3.frame()
[2641185.173] wl_pointer@3.frame()
[2641193.203] wl_pointer@3.frame() <----------- empty frames since mouse is totally at the top of the output
[2641243.306] wl_pointer@3.frame()
[2641478.382] wl_pointer@3.motion(286024261, 231.269531, 0.000000)
[2641478.413] wl_pointer@3.frame()
[2641484.244] wl_pointer@3.frame()
[2641497.258] wl_pointer@3.leave(106841, wl_surface@96)   <---------------------- this was a left button click
[2641497.283]  -> wl_pointer@3.set_cursor(106813, wl_surface@17, 3, 3)
[2641497.295]  -> wl_surface@17.attach(wl_buffer@52, 0, 0)
[2641497.301]  -> wl_surface@17.set_buffer_scale(2)
[2641497.305]  -> wl_surface@17.damage(0, 0, 24, 24)
[2641497.312]  -> wl_surface@17.commit()
[2641497.319] wl_pointer@3.frame()
[2641497.342] xdg_toplevel@95.configure(2560, 1440, array)
[2641497.349] xdg_surface@100.configure(106839)
[2641497.360]  -> xdg_surface@100.ack_configure(106839)
[2641497.365] xdg_toplevel@46.configure(1257, 450, array)
[2641497.370] xdg_surface@45.configure(106840)
[2641497.375]  -> xdg_surface@45.ack_configure(106840)
[2641509.548] wl_buffer@137.release()
```

This confuses the application -- Firefox, for instance, will refuse to take mouse input again until it is unfullscreened.

Whether this happens or not seems to be dependent on the dimensions and position of the container prior to being fullscreened. The below layout (`H[kitty V[kitty firefox]]`, no extra resizing) reliably reproduces the issue after resizing Firefox:

image

Xyene:
Previously, `find_edge` on a single fullscreen view would occasionally
return an edge rather than `WLR_EDGE_NONE`*. This would trigger entry
into `seatop_resize_tiling`, which doesn't have meaning for a fullscreen
view.

The result was that the fullscreen container hitbox was considered to be
that of where it'd be if it were tiling, so most clicks would not go
through.

Fixes #5792.

\* I suspect this to be due to floating point imprecision in `find_edge`,
but I didn't dig too deeply.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -73,6 +73,9 @@ static enum wlr_edges find_edge(struct sway_container *cont,
 			cont->border == B_CSD) {
 		return WLR_EDGE_NONE;
 	}
+	if (cont->fullscreen_mode) {
+		return WLR_EDGE_NONE;
+	}
 
 	enum wlr_edges edge = 0;
 	if (cursor->cursor->x < cont->x + cont->border_thickness) {


sway/input/seatop_default.c
```diff
@@ -73,6 +73,9 @@ static enum wlr_edges find_edge(struct sway_container *cont,
 			cont->border == B_CSD) {
```
Xyene, line 73: `cont->border == B_NORMAL` during the reproduction of this bug, so this branch is never triggered.

emersion (approved):
Thanks!

PR #5795, opened 2020-11-07
Use automatic surface enter/leave

kennylevinsen:
This is the sway implementation of https://github.com/swaywm/wlroots/pull/2468


emersion:
This looks pretty good to me from a very quick look. Is there a reason why this is new API, and doesn't completely replace `wlr_surface_{enter,leave}`?

kennylevinsen:
I was mainly unsure if it should be done as a breaking change. I find the old naming (send_*, implying that it's going blindly to do just that) to be a bit misleading with the new behavior.

kennylevinsen:
Ready for review together with wlroots PR.

kennylevinsen:
(If wlroots PR changed to old names is merged, then this PR will not be needed.)

emersion:
Not needed anymore.

PR #5798, opened 2020-11-10
output: add phys_scale to adjust EDID-reported size

danieldg:
Some monitors report an invalid size in their EDID.  While sway ignores these sizes (see phys_size_is_aspect_ratio), some other applications do not.  Instead of requiring a change to the monitor, allow specifying a scaling factor in the configuration to bring the physical dimensions in line with reality.

**Note**: Currently this is in terms of the original (EDID-specified) phys_size; it might be useful to just throw that value out and calculate it based on a scaling factor of (square) pixels.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -282,6 +282,7 @@ sway_cmd output_cmd_dpms;
 sway_cmd output_cmd_enable;
 sway_cmd output_cmd_max_render_time;
 sway_cmd output_cmd_mode;
+sway_cmd output_cmd_phys_scale;
 sway_cmd output_cmd_position;
 sway_cmd output_cmd_scale;
 sway_cmd output_cmd_scale_filter;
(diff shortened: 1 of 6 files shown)


Emantor:
The kernel already provides infrastructure to load `EDID` data from a file instead of requesting it from the monitor, see `CONFIG_DRM_LOAD_EDID_FIRMWARE`. Documentation can be found [here](https://www.kernel.org/doc/html/latest/admin-guide/edid.html), however its unfortunately rather sparse. Here, the arch wiki comes to the [rescue](https://wiki.archlinux.org/index.php/Kernel_mode_setting#Forcing_modes_and_EDID). This has the additional advantage of not only working with sway, but supporting other wayland compositors as well.

danieldg:
Thanks for pointing me to the kernel solution.  Both the arch wiki and the kernel documentation point to a method that is only useful if you only ever have one monitor plugged in to a given output.  However, after digging in the kernel source for a bit, I was able to find `/sys/kernel/debug/dri/0/HDMI-A-2/edid_override` which allows both adjusting and undoing the override without rebooting.

I also added this fact to archwiki.

PR #5802, opened 2020-11-11
Update sway-bar manpage to reflect default config

charlesrocket:
Default bar position in current version `1.5-5ae4f650` is `top` https://github.com/swaywm/sway/blob/master/config.in#L204


TheAvidDev:
The man page refers to the position if one isn't defined which is [bottom](https://github.com/swaywm/sway/blob/989123a2a5933367e5c7c39c3793f4814e026bf1/swaybar/config.c#L28).

charlesrocket:
@TheAvidDev O I see

PR #5807, opened 2020-11-12
input/keyboard: fix Group# bindings for keyboard groups

### LINKED ISSUE #5800: Only Group1, not Group2 bindings are active when using more than one keyboard layout
### **Sway Version:**
``sway version 1.5-0df5753a (Nov  7 2020, branch 'master')``

### **Debug Log:**
https://gist.github.com/dddlr/6ecbc566d1ec6a0ea74d4f756457738b

### **Configuration File:**
https://gist.github.com/dddlr/ca76e94599d6acee9f52fd8abf18a584

### **Description:**

When using two keyboard layouts defined via ``xkb_layout``, bindings prefixed with ``Group1+`` appear to be active regardless of the current layout. On the other hand, bindings prefixed with ``Group2+`` are never active.

To reproduce the debug log:

1. Type the equivalent of Super+D in Programmer Dvorak (i.e. Super + the physical H key) to run ``i3-dmenu-desktop``. Note that Programmer Dvorak is the default keyboard layout.
2. Open termite
3. Switch keyboard layout to QWERTY using Alt+Space (which is configured in ``xkb_options``)
4. Type Super+D in QWERTY
5. ``i3-dmenu-desktop`` still runs despite the binding only being defined for Group1.

In addition:

- Super+Shift+r (defined as ``Group1+$mod+Shift+r``) works for both Programmer Dvorak and QWERTY even though it is a Group1 binding.
- Super+Shift+b (defined as ``Group2+$mod+Shift+b``) doesn't do anything regardless of whether Programmer Dvorak or QWERTY is used.

I should mention that this bug shouldn't exist due to #3662 and #4351, so I'm wondering if I'm doing something wrong here?

Thanks!

RedSoxFan:
Fixes #5800 

The keyboard group's effective keyboard layout was never being changed
due to a condition that incorrectly preventing it from being performed.
The IPC event that follows the change was correctly being prevented.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -626,10 +626,12 @@ static void handle_modifier_event(struct sway_keyboard *keyboard) {
 		determine_bar_visibility(modifiers);
 	}
 
-	if (wlr_device->keyboard->modifiers.group != keyboard->effective_layout &&
-			!wlr_keyboard_group_from_wlr_keyboard(wlr_device->keyboard)) {
+	if (wlr_device->keyboard->modifiers.group != keyboard->effective_layout) {
 		keyboard->effective_layout = wlr_device->keyboard->modifiers.group;
-		ipc_event_input("xkb_layout", keyboard->seat_device->input_device);
+
+		if (!wlr_keyboard_group_from_wlr_keyboard(wlr_device->keyboard)) {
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

PR #5822, opened 2020-11-17
 criteria: fix crash when comparing NULL properties

paul-ri:
_Copy of the main commit message:_

For each following combinations of criteria & command below, the command would
crash sway without the fix.
It's particular about the __focused__ criteria, where the view matches part of
the criteria but not the focused app, leading to a failure when calling
`strcmp` with NULL.

"xterm" is a non-wayland app (X11) and "kitty" is. Both are terminals.

    # "class" is specific to X11
    # The view is X11 (xterm) leading to the criteria checking for the
    # focused app's class, leading to a crash
    for_window [class="__focused__"] floating enable
    exec kitty -e xterm

    # Similarly, crash as the focused app (xterm) has no app_id when the view has one
    for_window [app_id="__focused__"] floating enable
    exec xterm -e kitty

    # If the view has a title but not the focused app: NULL title will crash criteria checking
    for_window [title="__focused__"] floating enable
    exec xterm -title "" -e xterm

**EDIT:** Applied more fixes as suggested

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -188,7 +188,7 @@ static bool criteria_matches_view(struct criteria *criteria,
 
 		switch (criteria->title->match_type) {
 		case PATTERN_FOCUSED:
-			if (focused && strcmp(title, view_get_title(focused))) {
+			if (focused && lenient_strcmp(title, view_get_title(focused))) {
 				return false;
 			}
 			break;
@@ -228,7 +228,7 @@ static bool criteria_matches_view(struct criteria *criteria,
 
 		switch (criteria->app_id->match_type) {
 		case PATTERN_FOCUSED:
-			if (focused && strcmp(app_id, view_get_app_id(focused))) {
+			if (focused && lenient_strcmp(app_id, view_get_app_id(focused))) {
 				return false;
 			}
 			break;
@@ -260,7 +260,7 @@ static bool criteria_matches_view(struct criteria *criteria,
 
 		switch (criteria->class->match_type) {
 		case PATTERN_FOCUSED:
-			if (focused && strcmp(class, view_get_class(focused))) {
+			if (focused && lenient_strcmp(class, view_get_class(focused))) {
 				return false;
 			}
 			break;
--- a/common/stringop.c
+++ b/common/stringop.c
@@ -64,7 +64,7 @@ char *lenient_strncat(char *dest, const char *src, size_t len) {
 }
 
 // strcmp that also handles null pointers.
-int lenient_strcmp(char *a, char *b) {
+int lenient_strcmp(const char *a, const char *b) {
 	if (a == b) {
 		return 0;
 	} else if (!a) {
--- a/include/stringop.h
+++ b/include/stringop.h
@@ -12,7 +12,7 @@ char *lenient_strcat(char *dest, const char *src);
 char *lenient_strncat(char *dest, const char *src, size_t len);
 
 // strcmp that also handles null pointers.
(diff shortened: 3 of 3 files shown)


paul-ri:
I'm not sure why above we're checking for the view's `app_id` instead of the focused view `app_id`.  
Meaning why doing:
```c
if (criteria->app_id) {
		const char *app_id = view_get_app_id(view);
                if (!app_id) {
			return false;
		}
```
instead of:
```c
if (criteria->app_id) {
		const char *app_id = view_get_app_id(focused);
                if (!app_id) {
			return false;
		}
```

I have very little knowledge of the codebase so feel free to change the fix entirely.

sway/criteria.c
```diff
@@ -228,8 +228,13 @@ static bool criteria_matches_view(struct criteria *criteria,
 
 		switch (criteria->app_id->match_type) {
 		case PATTERN_FOCUSED:
-			if (focused && strcmp(app_id, view_get_app_id(focused))) {
-				return false;
+			{
+				const char *focused_app_id = view_get_app_id(focused);
```
paul-ri, line 232: I was tempted to use `xdg_shell_view_from_view` instead to determine the view type, but apparently it's not done this way. I've copied what l.224 is doing.

emersion:
>I'm not sure why in the code, slightly above the currently suggested fix, we're checking for the view's app_id instead of the focused view app_id.

Because we should only check for the focused app_id in case the match type is `PATTERN_FOCUSED`.

sway/criteria.c
```diff
@@ -228,8 +228,13 @@ static bool criteria_matches_view(struct criteria *criteria,
 
 		switch (criteria->app_id->match_type) {
 		case PATTERN_FOCUSED:
```
emersion, line 230: Style nit: instead of using braces, can use a semicolon:

```c
case PATTERN_FOCUSED:;
	const char *focused_app_id = view_get_app_id(focused);
```

The C syntax is annoying here.

emersion:
This fix looks good to me, but it seems we'd need to do the same for class/instance too, and maybe title if that can be NULL as well? Maybe we can use our `lenient_strcmp` helper to reduce the boilerplate a bit.

paul-ri:
> Because we should only check for the focused app_id in case the match type is PATTERN_FOCUSED.

Sure. I was in fact confused about the difference between `view` (a view) and `focused` (a view too) in that context. I don't get the whole hierarchy yet. I guess `view` could have multiple containers, one of them might a focused one needing checked.  
It'll get clearer as I keep looking at the code.

paul-ri:
> This fix looks good to me, but it seems we'd need to do the same for class/instance too, and maybe title if that can be NULL as well? Maybe we can use our `lenient_strcmp` helper to reduce the boilerplate a bit.

Awesome, I fixed the following cases:
```
for_window [class="__focused__"] floating enable
exec kitty -e xterm

for_window [title="__focused__"] floating enable
exec xterm -title "" -e xterm

for_window [app_id="__focused__"] floating enable
exec xterm -e kitty
```

More details in the commit. Let me know if there's other places you can think about.

sway/criteria.c
paul-ri: Cheers, didn't know about that one.

Removed in favour of `lenient_strcmp` as suggested

sway/criteria.c
paul-ri: Using `lenient_strcmp` instead

emersion (approved):
LGTM, thanks!

PR #5830, opened 2020-11-21
Remove outdated unsupported driver checks

RPigott:
This removes the detect_proprietary function and the `--my-next-gpu-wont-be-nvidia` flag.

Correct me if I'm wrong, but my understanding of the history of this code is that way back when sway was built atop wlc, wlc had rudimentary support for EGLStreams/nvidia so it was possible to run some sway 0.x versions using the nvidia driver. The result was a poor, buggy user experience that couldn't reasonably have been addressed within sway, and sway had a warning as such for a long time. Since sway never intended to support the nvidia driver in the first place, and a release built on wlroots breaks it even further, at some point sway was made to just abort when the nvidia driver was loaded so users would not accidentally use the driver and waste everyone's time filing bug reports.

So, the flag was included to help users that may want the unsupported drivers loaded for an application other than sway. AFAICT with at least semi-modern versions of wlroots both the abort and this flag no longer serve a purpose, which means we can drop the /proc/modules hack and related flags without issue and avoid further inconvenience for those users.

With this code removed, on my optimus laptop wlroots still ignores any nvidia devices and runs using the intel device with at most a non-fatal error message, even then only if I intentionally insert the nvidia module beforehand and specify the nvidia card first in WLR_DRM_DEVICES. I don't think there is any risk of a user successfully launching sway on nvidia. Even were it possible, wlroots selecting an nvidia device over a supported one would probably be a worthwhile bug to report.

I am aware that this flag has caused some controversy over the years, and, that there may be those who feel the strong wording of the flag and the error send a message that is worth keeping in sway on principle. I think controversy arises because some users are misinformed about the purpose of the abort and the flag, unaware of the underlying reason for the incompatibility. This misconception most likely comes from sway itself. In particular, the wording of the error, "Proprietary Nvidia drivers are NOT supported." is misleading because in fact it is nvidia that does not support wayland. The aggressive snooping of /proc/modules doesn't do us any favors either. The error reporting of wlroots, which plainly states which necessary feature(s) is unsupported by the user's hardware driver, is more useful here. Combined with the removal of all references to nvidia within sway's code, it should be more clear what is wrong _and_ who is at fault.

If there is still a need to blacklist some drivers, I think it would be better to put that behavior in wlroots, or at least have wlroots provide a way for the compositor to blacklist drivers without the /proc/modules hack. I left the `--unsupported-gpu` flag in as an inert option to leave open the possibility of re-purposing it to implement this kind of blacklisting.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -83,44 +83,6 @@ void detect_raspi(void) {
 	}
 }
 
-void detect_proprietary(int allow_unsupported_gpu) {
-	FILE *f = fopen("/proc/modules", "r");
-	if (!f) {
-		return;
-	}
-	char *line = NULL;
-	size_t line_size = 0;
-	while (getline(&line, &line_size, f) != -1) {
-		if (strncmp(line, "nvidia ", 7) == 0) {
-			if (allow_unsupported_gpu) {
-				sway_log(SWAY_ERROR,
-						"!!! Proprietary Nvidia drivers are in use !!!");
-			} else {
-				sway_log(SWAY_ERROR,
-					"Proprietary Nvidia drivers are NOT supported. "
-					"Use Nouveau. To launch sway anyway, launch with "
-					"--my-next-gpu-wont-be-nvidia and DO NOT report issues.");
-				exit(EXIT_FAILURE);
-			}
-			break;
-		}
-		if (strstr(line, "fglrx")) {
-			if (allow_unsupported_gpu) {
-				sway_log(SWAY_ERROR,
-						"!!! Proprietary AMD drivers are in use !!!");
-			} else {
-				sway_log(SWAY_ERROR, "Proprietary AMD drivers do NOT support "
-					"Wayland. Use radeon. To try anyway, launch sway with "
-					"--unsupported-gpu and DO NOT report issues.");
-				exit(EXIT_FAILURE);
-			}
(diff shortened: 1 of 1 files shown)


emersion:
Nack. The check was introduced in https://github.com/swaywm/sway/pull/2896, after Sway started using wlroots.

>With this code removed, on my optimus laptop wlroots still ignores any nvidia devices

That's one reason why this flag was introduced. Some screens may be connected to the Nvidia card, Sway will run but won't be able to use them, and users will be confused because Sway half-works.

RPigott:
Can you clarify if you also mean to NACK the possibility of extending wlroots for sway to do the blacklisting by driver name directly? Looks like it was suggested in that thread too.

I didn't want to re-implement a feature if we could just as easily have simply dropped it, but apparently that's not the case.

emersion:
>Can you clarify if you also mean to NACK the possibility of extending wlroots for sway to do the blacklisting by driver name directly? Looks like it was suggested in that thread too.

Yeah, that sounds fine. You'll hit the fact that a DRM backend can have multiple DRM devices opened so there's no single driver name.

RPigott:
Oh, I don't plan to work on it anytime soon – plenty of more interesting to do for now. If someone else wants to though they can.

PR #5835, opened 2020-11-24
Add mention of 'include' ordering wildcards based on locale

sochotnicky:
This can be important in case one of included files depends on
variable definitions in another. A common way to "fix" this is to
prefix config files with numbers so that they are loaded in proper
order.


sochotnicky:
Actually - not dependent on locale. I guess the sorting is based on C locale sort (numbers, uppercase, lowercase etc). 

I got caught when I defined my outputs in one file and workspace assignments in another - but they were evaluated in wrong order.

PR #5840, opened 2020-11-26
contrib/grimshot: changing the date format used in the filename

matteoredaelli:
Some tools (like Ms Teams) do not accept some characters like : 
So I propose to change the date format or we could add a parameter option

### DIFF
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -28,7 +28,7 @@ fi
 
 ACTION=${1:-usage}
 SUBJECT=${2:-screen}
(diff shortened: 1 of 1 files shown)


TheAvidDev:
You can already override the output file so I don't think having a specific setting just for the date format makes sense. Perhaps you should file a bug with Teams as well.

matteoredaelli:
And what about an optional parameter? I'd prefer not to create a new project forked from this script...

WhyNotHugo:
I'd close this in favour of #5914, which addresses this scenarios **plus** a few others.

Issue #5848, opened 2020-11-30
Atomic resizing of tiles is broken


emersion:
https://github.com/swaywm/sway/pull/5812 breaks atomic atomic resizing of tiles.

See https://github.com/swaywm/sway/pull/5812#issuecomment-735005960

PR #5850, opened 2020-12-01
[WIP]: Fix atomic transactions

RPigott:
This is a first pass at #5848 which reserves the new transaction for "unexpected" resizes.

It's not working perfectly. In cases like #5750 I can (sometimes?) see a transaction timing out before the border is corrected. I'm not sure what's happening here b/c as I understand it the final transaction should have no clients to wait for in the first place.

### DIFF
--- a/include/sway/desktop/transaction.h
+++ b/include/sway/desktop/transaction.h
@@ -1,6 +1,10 @@
 #ifndef _SWAY_TRANSACTION_H
 #define _SWAY_TRANSACTION_H
 #include <stdint.h>
+#include "sway/output.h"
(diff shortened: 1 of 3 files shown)


kennylevinsen:
I have been poking at this myself too, and I imagine the problem is that view->geometry and con->surface_x/surface_y isn't double-buffered. They are updated immediately when you call view_update_size, while the container remains fixed until the transaction completes.

I had some progress double-buffering these locally, but that also has some quirks that I haven't solved yet.

RPigott:
Hmm, yeah that makes sense. That's probably the way to go.

Issue #595, opened 2016-04-22
IntelliJ Idea show weird blank screen


jarias:
Don't know if it apply's to other java apps but I test IntelliJ in weston and it doesn't work quite well but at least it shows the window content. Also the window should the attach screenshot should be a modal floating window but it appears to be trying to render it as a normal window.

OS: `Linux lucy 4.5.1-1-ARCH #1 SMP PREEMPT Thu Apr 14 19:19:32 CEST 2016 x86_64 GNU/Linux`

Sway Package version `0.5-2`
WLC Package version `0.0.2-1`
XWayland Package version `1.18.3-1`

![2016-04-22-090717_swaygrab](https://cloud.githubusercontent.com/assets/119519/14756712/1d56da72-08a9-11e6-918b-686c39b9b5cf.png)

ids1024:
Other java programs, like the arduino ide, seem to be fairly broken in sway, though in different ways.

ids1024:
`export _JAVA_AWT_WM_NONREPARENTING=1` seems to fix the problem.

stwa:
This is a known problem with non-re-parenting window managers:
https://awesome.naquadah.org/wiki/Problems_with_Java#Impersonate_Another_Window_Manager

`wmname LG3D` works for me.

jarias:
@ids1024 @stwa thanks I'll try that as soon as I get home. I'm closing the issue since it looks like is a known Java issue.

ricejamesh:
I tried both _JAVA_AWT_WM_NONREPARENTING=1 and wmname LG3D with neither one working.  Are there any further suggestions for fixing this?

mkubasz:
I have this issue too.

emersion:
Please check if this issue is still relevant with sway 1.0, investigate it and send a patch.

mkubasz:
Yeap, new version fixed this issue. Now I'll be waiting for android studio emulator fix patch. Thanks.

ricejamesh:
2018.1.5 did not fix it for me.

emersion:
>waiting for […] fix patch

Investigate! Ask for help on IRC! Send a patch!

>2018.2 did not fix it for me. Bummer

Now idea what "2018.2" refers to.

roaddan:
I have the same issue.

I am using Microsoft remote desktop application (ver. 10.2.1) on Mac OS X Sierra to connect to my  Linux box which is a Debian 9 i686. 
[...]

PR #5851, opened 2020-12-01
Set _JAVA_AWT_WM_NONREPARENTING in sway.desktop

ForgottenUmbrella:
Fixes issue #595 .
From what I understand, Java expects non-reparenting WMs to set this environmental variable in order for applications like IntelliJ Idea to work, since it has no way to detect this on its own.


emersion:
Sorry, we won't add toolkit-specific quirks to the .desktop file.

PR #5853, opened 2020-12-02
Route wlroots logs into Sway logging infrastructure

emersion:
Instead of letting wlroots print messages to stdout, route debugging
messages into Sway's logging functions. This allows a more consistent
output (e.g. if Sway or wlroots changes its output style, they don't get
out-of-sync).

I also added a [wlr] prefix to wlroots messages, not yet sure it's a
good thing.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -223,6 +223,25 @@ void enable_debug_flag(const char *flag) {
 	}
 }
 
+static sway_log_importance_t convert_wlr_log_importance(
+		enum wlr_log_importance importance) {
+	switch (importance) {
+	case WLR_ERROR:
+		return SWAY_ERROR;
+	case WLR_INFO:
+		return SWAY_INFO;
+	default:
+		return SWAY_DEBUG;
+	}
+}
+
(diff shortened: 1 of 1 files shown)


sway/main.c
```diff
@@ -223,6 +223,25 @@ void enable_debug_flag(const char *flag) {
...
+		return SWAY_INFO;
+	default:
+		return SWAY_DEBUG;
+	}
+}
+
+static void handle_wlr_log(enum wlr_log_importance importance,
+		const char *fmt, va_list args) {
+	static char sway_fmt[1024];
+	snprintf(sway_fmt, sizeof(sway_fmt), "[wlr] %s", fmt);
```
ericonr, line 241: Unless there's a contract on the length of `fmt`, it's probably best to check the return value of snprintf.

emersion: I think a truncated log message isn't a big deal.

Xyene (approved):
LGTM, thanks! +1 for [wlr] prefix.

PR #5865, opened 2020-12-06
Wayland socket-based labels for criteria

danieldg:
This is (an improved version of) the labelling part of https://github.com/swaywm/sway/pull/5675, which is useful alone (as @ony mentioned).  There is currently no support for handling Xwayland sockets (and I'm not sure it would make sense to do so; at least none of the current methods for setting labels would support X).  Since this isn't a feature that i3 could support (due to i3 not being the display server) and is similar to an existing sway-only feature (matching pid in criteria), I don't think i3 compatibility is relevant.

Labels are distinct from marks because multiple windows and clients can have a single label, otherwise they are similar in concept.  This merge request does not contain a way to set a label on an existing client, but that could be added later if it is useful.

Example that launches two terminals on different workspaces as part of sway startup:
```
assign [cli_label=wks1] workspace 1
assign [cli_label=wks2] workspace 2
exec --label wks1 $term
exec --label wks2 $term
```

### DIFF
--- a/include/sway/client_label.h
+++ b/include/sway/client_label.h
@@ -1,0 +1,4 @@
+#include <wayland-server-core.h>
+
+char* wl_client_label_get(struct wl_client *client);
+void wl_client_label_set(struct wl_client *client, char* label);
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -304,4 +304,6 @@ sway_cmd cmd_ipc_cmd;
 sway_cmd cmd_ipc_events;
 sway_cmd cmd_ipc_event_cmd;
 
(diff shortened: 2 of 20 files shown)


RPigott:
I think I'd rather support a feature like this with startup notifications, the way #6639 does for workspace matching. It currently falls back to pid matching automatically, but it would be nice if even that could be removed if xdga adoption is high.

ony:
@RPigott, one of the feature that comes out of this management is ability to support hierarchy of processes rather than single PID.
Back then I wanted to have split screen (like in some games) between users having own set of keyboard and mouse.

PR #5874, opened 2020-12-11
Fix for_window criteria and mouse button bindings

ftilde:
Previously, the special case handling of scratchpad and unmark commands
was (probably accidentally) limited to criteria directly handled in the
execute_command function. This would exclude: 1. for_window criteria, as
these are handled externally for views and 2. and mouse bindings which
select target the node currently under the mouse cursor.

As a concrete example `for_window [app_id="foobar"] move scratchpad,
scratchpad show` would show (or hide due to the toggling functionality)
another window from the scratchpad, instead of showing the window with
app_id "foobar".

This commit replaces the "using_criteria" flag with "node_overridden"
with the more general notion of signifying that the node (and
container/workspace) in the current command handler context of the sway
config is not defined by the currently focused node, but instead
overridden by other means, i.e., criteria or mouse position.

### for_window criteria:

Choose two applications `$app1` and `$app2` with respective app ids `$appid1` and `$appid2`. Add the following to your sway config:

```
for_window [app_id="$appid1"] move scratchpad, scratchpad show
for_window [app_id="$appid2"] move scratchpad, scratchpad show
bindsym $somekey $app1
bindsym $otherkey $app2
```

Press `$somekey`. `$app1` will be started, moved to scratchpad and then shown.
Then press `$otherkey`. `$app2` will be started and moved to scratchpad. Now the behaviour diverges:

#### master:
`$app1` will be hidden (because the criterion of the second config line of above is not applied to the `scratchpad show`).

#### with this patch:
`$app2` will be shown as expected.

### mouse button bindings:

In your sway config:
```
focus_follows_mouse no
bindsym --whole-window $mod+button2 scratchpad show
```

Again, choose two applications (`$app1`, `$app2`), move them to scratchpad, show them both on your desktop.
Now focus `$app1`, but move your mouse over onto the window of `$app2` and press `$mod` + middle mouse button.

master:
`$app1` will be hidden (since the container override of the button command is not propagated to `scratchpad show`).

with this patch:
`$app2` will be hidden as expected.

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -174,10 +174,11 @@ static const struct cmd_handler *find_core_handler(char *line) {
 			handlers, sizeof(handlers));
 }
 
-static void set_config_node(struct sway_node *node) {
+static void set_config_node(struct sway_node *node, bool node_overridden) {
 	config->handler_context.node = node;
 	config->handler_context.container = NULL;
 	config->handler_context.workspace = NULL;
+	config->handler_context.node_overridden = node_overridden;
 
 	if (node == NULL) {
 		return;
@@ -202,6 +203,7 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 	char *cmd;
 	char matched_delim = ';';
 	list_t *containers = NULL;
+	bool using_criteria = false;
 
 	if (seat == NULL) {
 		// passing a NULL seat means we just pick the default seat
@@ -225,7 +227,7 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 		for (; isspace(*head); ++head) {}
 		// Extract criteria (valid for this command list only).
 		if (matched_delim == ';') {
-			config->handler_context.using_criteria = false;
+			using_criteria = false;
 			if (*head == '[') {
 				char *error = NULL;
 				struct criteria *criteria = criteria_parse(head, &error);
@@ -239,7 +241,7 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 				containers = criteria_get_containers(criteria);
 				head += strlen(criteria->raw);
 				criteria_destroy(criteria);
-				config->handler_context.using_criteria = true;
+				using_criteria = true;
 				// Skip leading whitespace
 				for (; isspace(*head); ++head) {}
 			}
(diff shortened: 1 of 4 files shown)


ftilde:
I've just rebased on top of master and added a test test which (hopefully) will make reviewing this patch easier.

I would be very happy if this made it into the next release, since I've actually been mildly but continuously annoyed by this ever since switching to sway (and now finally enough to look into it).

sway/commands.c
```diff
@@ -278,10 +279,17 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
...
 		}
 
-		if (!config->handler_context.using_criteria) {
+
+		if (!using_criteria) {
 			// The container or workspace which this command will run on.
-			struct sway_node *node = con ? &con->node :
-					seat_get_focus_inactive(seat, &root->node);
+			struct sway_node *node;
+			if(con) {
```
Xyene, line 286: Nit: space after the `if`.

Nit: no need to define `struct sway_node *node`, can just call `set_config_node` directly in both branches.

sway/commands.c
```diff
@@ -278,10 +279,17 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
...
+				node = &con->node;
+				config->handler_context.node_overridden = true;
+			} else {
+				node = seat_get_focus_inactive(seat, &root->node);
+				config->handler_context.node_overridden = false;
```
Xyene, line 291: The internal state is only consistent so long as `node_overridden` is itself consistent with `handler_context.container`. For that to be the case, one has to remember to always set this flag appropriately after (or immediately before) a call to `set_config_node`.

Instead, I think `set_config_node` should take this bool as a parameter, to reduce chances for programmer error.

Xyene (changes requested):
Thanks for working on this -- I've left a couple comments.

Also, please rebase since it looks like `sway/commands/scratchpad.c` has acquired conflicts.

sway/commands.c
ftilde: Yes, that's probably a good idea.

ftilde:
Alright, I've implemented the requested changes (as a new version of the original patch) and rebased on top of master!

ftilde:
The last change just fixed the wording in a comment.

Xyene (approved):
LGTM, thanks!

PR #5877, opened 2020-12-13
Delay refocus until after arranging nodes

### LINKED ISSUE #5873: IPC response is incorrect. On subscription to window events, the focus type reports the wrong values for rect.height and rect.width
- **Sway Version:**
```
>> sway -v
sway version 1.5.1
```

- **Debug Log:**

`sway -c /etc/sway/config -d 2> ~/sway.log`
https://gist.github.com/jhzn/59f4ed7946bdc7a072ce32bbeaf3332d

- **Description:**

I have a program which subscribes to window events.
I don't believe the details of that program are significant because I'm using strace like below to demonstrate the data being received in my program.
Let's say I have only 1 window(1920x1080, x and y of the window) open on a workspace.
I open another one, it's placed to the right of the first window. Now I have 2 windows of size 960x1080 because 1920/2 = 960.

When I close the second window. A `Window Change` event of the `Focus` type is triggered.
This data I assume contain the information of the focused window, this being the 1st window(the only one remaining).

The fields `container.rect.height` and `container.rect.width` report the incorrect data. I believe they contain stale data.
This data being what the window was previously, before the 2nd window was closed. With the example dimensions above I would expect the x and y of these fields now to be 1920x1080 but it's 960x1080.

This is how I received the following IPC data.
```shell
#autotiling-rs is my program
strace -s 10000 -p $(pgrep autotiling-rs)
```
Output which is of relevance.
```
recvfrom(4, "i3-ipc", 6, 0, NULL, NULL) = 6
recvfrom(4, "\20\3\0\0", 4, 0, NULL, NULL) = 4
recvfrom(4, "\3\0\0\200", 4, 0, NULL, NULL) = 4
recvfrom(4, "{ \"change\": \"focus\", \"container\": { \"id\": 42, \"name\": \"Alacritty\", \"rect\": { \"x\": 10, \"y\": 10, \"width\": 948, \"height\": 1030 }, \"focused\": true, \"focus\": [ ], \"border\": \"pixel\", \"current_border_width\": 2, \"layout\": \"none\", \"orientation\": \"none\", \"percent\": 0.49894736842105264, \"window_rect\": { \"x\": 2, \"y\": 2, \"width\": 944, \"height\": 1026 }, \"deco_rect\": { \"x\": 0, \"y\": 0, \"width\": 0, \"height\": 0 }, \"geometry\": { \"x\": 0, \"y\": 0, \"width\": 800, \"height\": 600 }, \"window\": null, \"urgent\": false, \"marks\": [ ], \"fullscreen_mode\": 0, \"nodes\": [ ], \"floating_nodes\": [ ], \"sticky\": false, \"type\": \"con\", \"pid\": 33829, \"app_id\": \"Alacritty\", \"visible\": true, \"max_render_time\": 0, \"shell\": \"xdg_shell\", \"inhibit_idle\": false, \"idle_inhibitors\": { \"user\": \"none\", \"application\": \"none\" } } }", 784, 0, NULL, NULL) = 784
```

If I run a command to get the `tree` and find the node therein, the data is correct.

rpigott:
fixes #5873
    
i3 determines focus after it arranges its nodes, so the ipc events that follow have the newly updated window geometry.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -844,6 +844,7 @@ void view_unmap(struct sway_view *view) {
 		seat_consider_warp_to_focus(seat);
 	}
 
+	wl_signal_emit(&view->container->node.events.destroy, &view->container->node);
 	transaction_commit_dirty();
 	view->surface = NULL;
 }
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -97,8 +97,6 @@ void container_begin_destroy(struct sway_container *con) {
 		container_fullscreen_disable(con);
 	}
 
-	wl_signal_emit(&con->node.events.destroy, &con->node);
-
 	container_end_mouse_operation(con);
 
 	con->node.destroying = true;


sway/input/seat.c
```diff
@@ -329,7 +329,7 @@ static struct sway_seat_node *seat_node_from_node(
 	seat_node->node = node;
 	seat_node->seat = seat;
 	wl_list_insert(seat->focus_stack.prev, &seat_node->link);
-	wl_signal_add(&node->events.destroy, &seat_node->destroy);
```
RPigott, line 332: I intended to separate out the refocus logic from destroy so it could be figured out later, but on closer inspection, it seems that after this change the node::destroy event is just unused, because this was actually its only listener.

So maybe there is no need for a new event after all. Is it still worth renaming it to "postdestroy" or should I just move the *emits around?

sway/tree/workspace.c
```diff
@@ -147,6 +147,8 @@ void workspace_begin_destroy(struct sway_workspace *workspace) {
 	if (workspace->output) {
 		workspace_detach(workspace);
 	}
+	wl_signal_emit(&workspace->node.events.postdestroy, &workspace->node);
```
RPigott, line 150: Yeah, this is wrong – interferes with the logic in 9350a52c0. Since it doesn't make sense to split the signal here either, I'm going back to just the single destroy signal again.

sway/tree/view.c
```diff
@@ -844,6 +844,7 @@ void view_unmap(struct sway_view *view) {
 		seat_consider_warp_to_focus(seat);
 	}
 
+	wl_signal_emit(&view->container->node.events.destroy, &view->container->node);
```
emersion, line 847: I'm not very comfortable moving the `wl_signal_emit` call here. This makes it so the signal is no longer emitted for containers which aren't views.

Is it possible to emit it in `container_destroy` instead?

RPigott: Unfortunately emitting the signal any later segfaults in cursor_rebase. I think it doesn't expect view->surface == NULL.

PR #5879, opened 2020-12-13
ci: fix archlinux builds

### LINKED ISSUE #5878: Arch Linux CI builds broken
Following Fedora, Arch recently changed the Xwayland package from `xorg-server-xwayland` to `xorg-xwayland` to separate the Xwayland build/release from X.org [1]. As I understand it, that breaks the CI build because:

1. xorg-server-xwayland is gone
2. Our egl dependency was previously satisfied by the `xorg-server-xwayland` dependency on libgl, which the new package doesn't have.

[1] https://bugs.archlinux.org/task/68916

RPigott:
Fixes #5878

### DIFF
--- a/.builds/archlinux.yml
+++ b/.builds/archlinux.yml
@@ -3,6 +3,7 @@ packages:
   - cairo
   - gdk-pixbuf2
   - json-c
+  - libegl
   - libinput
   - libxcb
   - libxkbcommon
@@ -12,7 +13,7 @@ packages:
   - wayland
   - wayland-protocols
   - xcb-util-image
-  - xorg-server-xwayland
+  - xorg-xwayland
 sources:
   - https://github.com/swaywm/sway
   - https://github.com/swaywm/wlroots


.builds/archlinux.yml
```diff
@@ -12,7 +13,7 @@ packages:
   - wayland
   - wayland-protocols
   - xcb-util-image
-  - xorg-server-xwayland
+  - xorg-xwayland
```
emersion, line 16: I wonder if we need `xorg-xwayland` at all, actually. We don't link to it, it's purely a runtime dependency.

RPigott: I would have thought wlroots uses the xwayland.pc file for some of the Xwayland variables but it looks like it doesn't.

AFAICT the potentially useful vars provided are "have_initfd" which wlroots doesn't use but maybe could? (I'm not 100% sure what it's for) and "xwayland" which gives the path to the Xwayland binary, which wlroots could maybe make use of, but doesn't. The commit message here has some info about why we might want to use the pc file: https://gitlab.freedesktop.org/xorg/xserver/-/commit/5188603ff7d25186c9076fefeceba825f3a40d47

So I think we could drop it now, but might have to add it back in if we want to make use of those features in wlroots later. What do you think?

emersion: >AFAICT the potentially useful vars provided are "have_initfd" which wlroots doesn't use but maybe could? (I'm not 100% sure what it's for) and "xwayland" which gives the path to the Xwayland binary, which wlroots could maybe make use of, but doesn't.

Yeah. We don't use it because we don't check for Xwayland at build-time. I don't know if we should.

>So I think we could drop it now, but might have to add it back in if we want to make use of those features in wlroots later. What do you think?

I guess it's fine to keep it, doesn't hurt.

emersion (approved):
Thanks!

PR #5882, opened 2020-12-13
Fix layer shell subsurface use after free

psnszsn:
This occurs when a layer surface is destroyed , prompting an unmap event for all its children subsurfaces (if it has any). Since this happens after the parent surface is destroyed, any pointers to it become dangling.

A better approach might be to make wlroots unmap the children subsurfaces _before_ the parent is freed, but I couldn't figure out any easy way to do that.

To test this, start gtk-layer-demo, open the popover, and kill the process with the popover still open.

### DIFF
--- a/include/sway/layers.h
+++ b/include/sway/layers.h
@@ -48,6 +48,7 @@ struct sway_layer_subsurface {
 	struct wl_listener unmap;
 	struct wl_listener destroy;
 	struct wl_listener commit;
+	struct wl_listener layer_surface_destroy;
 };
 
 struct sway_output;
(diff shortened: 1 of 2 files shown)


emersion:
How does the xdg-shell code handle this?

psnszsn:
> Maybe we can iterate over the children subsurfaces when the layer surface is destroyed?

Wlroots already iterates over children subsurfaces when the parent is destroyed and triggers an unmap on them. We could iterate over them first and set the pointer to the parent to NULL, but that isn't ideal.

> How does the xdg-shell code handle this?

Xdg-shell code avoids this problem in 2 ways:
1. it does not do damage to subsurfaces on map and unmap
2. even if it did, the `view_child_damage` function does not seem to use the pointer to the parent view.

I guess I should remove the map and unmap handlers too? That would avoid the problem.

emersion:
Superseded by https://github.com/swaywm/sway/pull/6728 I believe.

PR #5890, opened 2020-12-16
Implement input_method_v2 popups 

tadeokondrak:
Depends on ~~#4932~~, ~~https://github.com/swaywm/wlroots/pull/2550~~ and ~~https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3416~~.
~~https://github.com/swaywm/wlroots/pull/2697~~, ~~#5980~~, ~~#6004~~, and ~~#6005~~ are related but not required.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -19,6 +19,7 @@
 #include "sway/config.h"
 #include "sway/input/input-manager.h"
 #include "sway/input/seat.h"
+#include "sway/input/text_input.h"
 #include "sway/layers.h"
 #include "sway/output.h"
 #include "sway/server.h"
@@ -200,6 +201,16 @@ static void render_unmanaged(struct sway_output *output,
 		render_surface_iterator, &data);
 }
 #endif
+
+static void render_input_popups(struct sway_output *output,
+		pixman_region32_t *damage, struct wl_list *input_popups) {
+	struct render_data data = {
+		.damage = damage,
+		.alpha = 1.0f,
+	};
+	output_input_popups_for_each_surface(output, input_popups,
+		render_surface_iterator, &data);
+}
 
 static void render_drag_icons(struct sway_output *output,
 		pixman_region32_t *damage, struct wl_list *drag_icons) {
@@ -1029,6 +1040,9 @@ void output_render(struct sway_output *output, struct timespec *when,
 	struct wlr_output *wlr_output = output->wlr_output;
 	struct wlr_renderer *renderer = output->server->renderer;
 
+	struct sway_seat *seat = input_manager_current_seat();
+	struct sway_container *focus = seat_get_focused_container(seat);
+
 	struct sway_workspace *workspace = output->current.active_workspace;
 	if (workspace == NULL) {
 		return;
@@ -1125,8 +1139,6 @@ void output_render(struct sway_output *output, struct timespec *when,
 
 	render_seatops(output, damage);
 
-	struct sway_seat *seat = input_manager_current_seat();
-	struct sway_container *focus = seat_get_focused_container(seat);
 	if (focus && focus->view) {
 		render_view_popups(focus->view, output, damage, focus->alpha);
 	}
@@ -1136,6 +1148,7 @@ render_overlay:
 		&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]);
 	render_layer_popups(output, damage,
 		&output->layers[ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY]);
+	render_input_popups(output, damage, &seat->im_relay.input_popups);
 	render_drag_icons(output, damage, &root->drag_icons);
 
 renderer_end:
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -37,6 +37,28 @@ static uint32_t get_current_time_msec(void) {
 	return now.tv_sec * 1000 + now.tv_nsec / 1000000;
 }
 
+static struct wlr_surface *input_popup_surface_at(struct sway_output *output,
+		struct sway_input_method_relay *relay, double ox, double oy, double *sx, double *sy) {
+	struct sway_input_popup *popup;
+	wl_list_for_each(popup, &relay->input_popups, link) {
+		int lx, ly;
+		if (!sway_input_popup_get_position(popup, &lx, &ly)) {
+			continue;
+		}
+		if (!popup->popup_surface->mapped || !popup->visible) {
+			continue;
+		}
+		double _sx = ox - lx;
+		double _sy = oy - ly;
+		struct wlr_surface *sub = wlr_surface_surface_at(
+			popup->popup_surface->surface, _sx, _sy, sx, sy);
+		if (sub) {
+			return sub;
+		}
+	}
+	return NULL;
+}
+
 static struct wlr_surface *layer_surface_at(struct sway_output *output,
 		struct wl_list *layer, double ox, double oy, double *sx, double *sy) {
 	struct sway_layer_surface *sway_layer;
@@ -136,6 +158,11 @@ struct sway_node *node_at_coords(
 	// find the focused workspace on the output for this seat
 	struct sway_workspace *ws = output_get_active_workspace(output);
 	if (!ws) {
+		return NULL;
+	}
+
+	if ((*surface = input_popup_surface_at(output,
+			&seat->im_relay, ox, oy, sx, sy))) {
 		return NULL;
 	}
 
--- a/sway/input/text_input.c
+++ b/sway/input/text_input.c
@@ -3,6 +3,11 @@
 #include "log.h"
 #include "sway/input/seat.h"
 #include "sway/input/text_input.h"
+#include "sway/tree/view.h"
+#include "sway/tree/container.h"
+#include "sway/desktop.h"
+#include "sway/output.h"
+#include "sway/layers.h"
 
 static struct sway_text_input *relay_get_focusable_text_input(
 		struct sway_input_method_relay *relay) {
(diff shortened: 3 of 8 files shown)


sway/desktop/render.c
```diff
@@ -986,6 +986,36 @@ static void render_seatops(struct sway_output *output,
...
+		if (popup == NULL || popup->popup_surface == NULL
+				|| !popup->popup_surface->mapped) {
+			continue;
+		}
+		struct render_data data = {
+			.damage = damage,
+			.alpha = 1.0f,
+		};
+		struct sway_view *view = view_from_wlr_surface(
+			popup->text_input->input->focused_surface);
```
ngkz, line 1003: I encountered NULL popup->text_input dereference at here when I insert null check at [here](https://github.com/tadeokondrak/wlroots/commit/8960e09ae207e106603b0e7db16e3876881e4857#diff-d68541b5302b73ea493573e636e082fccb19a59399139f7adf1eca439742625cR458) and activate fcitx5.

tadeokondrak: This seems to be a problem with this PR. It doesn't send frame callback events, probably due to the surface damage being incorrect. I'll keep testing with fcitx5.

tadeokondrak: Hi, it works with fcitx5 now for me.

12101111:
sway crash on exit
```
  * frame #0: 0x00007fcc2418f6a7 libwayland-server.so.0`wl_list_remove(elm=0x00007fcc1bd2dd98) at wayland-util.c:55:18
    frame #1: 0x00007fcc2413a06a libwlroots.so.7`xdg_popup_grab_handle_seat_destroy(listener=<unavailable>, data=<unavailable>) at wlr_xdg_popup.c:158:2
    frame #2: 0x00007fcc2416634c libwlroots.so.7`wlr_signal_emit_safe(signal=<unavailable>, data=0x00007fcc1c04e050) at signal.c:29:3
    frame #3: 0x00007fcc24134e88 libwlroots.so.7`wlr_seat_destroy(seat=0x00007fcc1c04e050) at wlr_seat.c:170:2
    frame #4: 0x00007fcc2418975f libwayland-server.so.0`wl_display_destroy [inlined] wl_priv_signal_final_emit(signal=0x00007fcc245fce88, data=0x00007fcc245fce20) at wayland-server.c:2157:3
    frame #5: 0x00007fcc24189728 libwayland-server.so.0`wl_display_destroy(display=0x00007fcc245fce20) at wayland-server.c:1134
    frame #6: 0x00005589ae2246a3 sway`server_fini(server=0x00005589ae2825d0) at server.c:203:2
    frame #7: 0x00005589ae223dee sway`main(argc=1, argv=0x00007ffd55a5e338) at main.c:436:2
    frame #8: 0x00007fcc247c0129 ld-musl-x86_64.so.1`libc_start_main_stage2(main=(sway`main at main.c:245), argc=<unavailable>, argv=0x00007ffd55a5e338) at __libc_start_main.c:94:7
    frame #9: 0x00005589ae216eb6 sway`_start + 22
```

emersion:
#4932 has been merged now, so this needs a rebase.

emersion:
What client can we test this PR with?

include/sway/layers.h
```diff
@@ -24,6 +24,8 @@ struct sway_layer_surface {
 
 	struct wlr_box geo;
 	enum zwlr_layer_shell_v1_layer layer;
+
+	struct wl_list input_popups;
```
emersion, line 28: Hm, why is there a list of input popups for each layer surface?

I figured input popups would be "global", not per-layer-surface. Another example of a "global" surface is drag icons.

12101111:
> What client can we test this PR with?

You can test this with fcitx5 + alacritty

Use this command to see which frontend is using by client 
```
dbus-send --session --type=method_call --print-reply --dest=org.fcitx.Fcitx5 /controller org.fcitx.Fcitx.Controller1.DebugInfo
```

You can see `progeam:alacritty frontend:waylandim` if this PR works

Shugyousha:
I have done some basic testing of this PR with https://github.com/tadeokondrak/anthywl.

include/sway/layers.h
tadeokondrak: I'm not sure if this review comment is specific to layer surfaces or includes views too, but if it's both, then:
It seems to me that having the input popup in the view/layer surface is simpler than having it global, because it needs to be positioned relative to the parent surface. If it isn't, then the popup has to be updated every time the parent surface changes.

emersion: Hm, but the IME only creates a single popup, right? And it's re-used for all text-input clients?

When the parent surface changes, the popup needs to be updated anyways, because the focused text field changes as well?

tadeokondrak: > Hm, but the IME only creates a single popup, right? And it's re-used for all text-input clients?

With the current protocol, input methods can create multiple popups, but without positioning support, this is a bit useless.
But I'm not sure if there's a usecase for multiple popups (on the same seat) even with positioning support.

> When the parent surface changes, the popup needs to be updated anyways, because the focused text field changes as well?

My wording wasn't clear, I meant the parent surface updating geometry, not the focus changing.

emersion:
This PR renders the IM popup right above the surface with the input field, below everything else. This is an issue if the surface with the input field has subsurfaces/xdg-popups above it: the IM popup will be rendered below these.

I think we should render the IM popup above everything else instead, e.g. here: https://github.com/swaywm/sway/blob/6059c744f8b24ef3321eb7896991112366896e26/sway/desktop/render.c#L1127

include/sway/layers.h
emersion: > But I'm not sure if there's a usecase for multiple popups (on the same seat) even with positioning support.

Wondered about this as well: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/40#note_847039

If the Sway implementation can be simplified by ignoring extra popups, maybe it'd be worth it.

> I meant the parent surface updating geometry, not the focus changing.

We can get the globally-active IME. From the IME we can get the `wlr_surface` with the focused text field. If we could get the position of this surface, then positioning the IME popup would be easy, correct?

fosskers:
Alright, I have a custom patched version of `wlroots-git` installed (based on https://github.com/swaywm/wlroots/pull/2550), and a build of `sway-git` based on `master`. So far everything is running. The result of the command given above by @12101111 is:
```
method return time=1618595438.050418 sender=:1.5 -> destination=:1.33 serial=2513 reply_serial=2
   string "Group [wayland:] has 6 InputContext(s)
  IC [561d5ca5dac845418f8b84dd0e3f054c] program:firefox frontend:dbus cap:e001000012 focus:0
  IC [ca565dfc223e47249f57f2df37d7ec6e] program:firefox frontend:dbus cap:e001000012 focus:0
  IC [ce9472215e87439f9c442dcc2a1915bf] program:firefox frontend:dbus cap:e001000012 focus:0
  IC [5b374d3324394819a73b486a9ec5b72e] program:firefox frontend:dbus cap:e001000012 focus:0
  IC [d0a8c154dd4e414bb90eca715ddf321e] program:firefox frontend:dbus cap:e001000052 focus:0
  IC [c99385fcb290441281077b627eac3f50] program: frontend:wayland_v2 cap:0 focus:1
Group [x11::0] has 0 InputContext(s)
Input Context without group
"
```
Now I'm going to rebuild the latter with this PR and see what happens.

fosskers:
I'm still getting the same results from the `dbus` command. Is some special configuration required?

12101111:
Yes, the code still have some bugs ( or fcitx5 have some bugs )

tadeokondrak:
fcitx5 has keyboard input issues when using input-method-v2.

As for this PR, since there are not many clients using it, I think it's worth implementing a future version of the protocol with these issues fixed instead: https://github.com/swaywm/wlroots/pull/2550#pullrequestreview-616270027

glyh:
I hope this PR will be merged soon, since IMEs are basically not usable on wayland.

tadeokondrak:
I've resolved the review comment; this should be ready for review.

wildwestrom:
I'm so hyped for this! I couldn't wait so I built it from source on my machine.

wildwestrom:
Can we re-run the job? Seems like Freedesktop's gitlab is working again.

tinywrkb:
Does this still work correctly with Sway 1.7?  
I don't have popups working with Fcitx5 anymore when applying this on top of 1.7.  
Fcitx5 is built from the latest commit in master.  

This sort of seems to work with anthywl, but I don't know how to type in Japanese, so I just hit random keys, and sometimes I get a popup, but I don't get a visual feedback for every key press, and I have no idea if this is an expected Japanese typing behavior.

wildwestrom:
@tinywrkb 

I'm on Arch Linux, using the `sway-hidpi-git` package and `fcitx5-git`. Japanese input through mozc works through Xwayland and native Wayland windows.

Not sure if those packages track changes from this PR or not.

tinywrkb:
@wildwestrom you don't pull in the PR, so

> Japanese input through mozc works through Xwayland and native Wayland windows.

means XIM, and the widget toolkit's IM modules. But when it's a native Wayland client without an IM module, then you won't have popups.  
Previously, when I used this PR, I had popups working with Alacritty and Kitty, both running as native Wayland clients, and using Fcitx5.   
Is this just Fcitx5's fault? I have no idea.

JingMatrix:
I apply this PR with the current HEAD.
The api of `wlr_output_layout_get_box` has changed, so you should change your code a bit in `sway/sway/input/text_input.c`.
After this, I compiled your code, and tested it with `alacritty` and `kitty`, both gave me the same result:
the popup showed up at the very beginning, and then never showed up again.

dragove:
I've tested it with AUR package [sway-im](https://aur.archlinux.org/packages/sway-im) with fcitx5 as input method in kitty. When the character meets the end of a line, the popup will be hidden. I don't know if it is a issue with kitty. Alacritty won't meet this issue since it won't put characters into the line before I select my needed word in IM.

JingMatrix:
Now I test it ( after fixing the output_box api problem) with the HEAD of fcitx5 and HEAD of sway, everything works!

Legogris:
@emersion Looks like this is ready for another review :)

sway/tree/container.c
```diff
@@ -395,6 +395,8 @@ static bool surface_is_popup(struct wlr_surface *surface) {
 			wlr_subsurface_from_wlr_surface(surface);
 		surface = subsurface->parent;
 	}
+	if (wlr_surface_is_input_popup_surface_v2(surface))
```
emersion, line 398: Style: braces are mandatory

sway/input/cursor.c
```diff
@@ -139,6 +161,11 @@ struct sway_node *node_at_coords(
 		return NULL;
 	}
 
+	if ((*surface = input_popup_surface_at(output,
```
emersion, line 164: We should probably move this up right before "check for unmanaged views" to allow fullscreen surfaces to work with IMEs.

sway/input/cursor.c
```diff
@@ -37,6 +37,28 @@ static uint32_t get_current_time_msec(void) {
...
+	wl_list_for_each(popup, &relay->input_popups, link) {
+		int lx, ly;
+		if (!sway_input_popup_get_position(popup, &lx, &ly)) {
+			continue;
+		}
+		if (!popup->popup_surface->mapped || !popup->visible) {
+			continue;
+		}
+		double _sx = ox - lx;
+		double _sy = oy - ly;
```
emersion, line 52: This mixes up layout-local coords and output-local coords. I think `input_popup_surface_at` should take layout-local coords instead.

sway/input/text_input.c
```diff
@@ -26,6 +31,329 @@ static struct sway_text_input *relay_get_focused_text_input(
...
+		wl_signal_add(
+			&layer->layer_surface->events.unmap, &popup->focused_surface_unmap);
+		goto update_popup;
+	}
+
+	struct sway_view *view = view_from_wlr_surface(surface);
+	wl_signal_add(&view->events.unmap, &popup->focused_surface_unmap);
+
+	// Since the focus has changed, the popup may have to adjust
+update_popup:
```
emersion, line 267: Instead of a `goto`, `else if`/`else` would be more readable.

sway/input/text_input.c
```diff
@@ -26,6 +31,329 @@ static struct sway_text_input *relay_get_focused_text_input(
...
+			layer_from_wlr_layer_surface_v1(layer_surface);
+		*lx = layer->geo.x + popup->x;
+		*ly = layer->geo.y + popup->y;
+		return true;
+	}
+
+
+	struct sway_view *view = view_from_wlr_surface(
+		text_input->input->focused_surface);
+	if (view->container == NULL) {
```
emersion, line 328: This can return NULL here.

This logic doesn't work on subsurfaces nor xdg_popups.

evanslify:
@tadeokondrak @emersion This PR seems to have some amendments pending. Can I take over by starting another PR?

tadeokondrak:
Sorry about this PR... I've left it open because I thought I could get to it eventually - but it looks like someone else is continuing the effort at https://github.com/swaywm/sway/pull/7226, which is great.

Decodetalkers:
> Sorry about this PR... I've left it open because I thought I could get to it eventually - but it looks like someone else is continuing the effort at https://github.com/swaywm/sway/pull/7226, which is great.

Thanks that you have return back..Your help is still needed..Do you still have time now?

Decodetalkers:
> Sorry about this PR... I've left it open because I thought I could get to it eventually - but it looks like someone else is continuing the effort at #7226, which is great.

```
This mixes up layout-local coords and output-local coords. I think input_popup_surface_at should take layout-local coords instead.
```

I do not know how to make change about this request, can you help me?

PR #5891, opened 2020-12-16
Fix swaybar tray for non-systemd

nothien:
Meson's generated config.h header defines false macros as 0, not
undefined.  This means that the header line, which was checking for the
definition existing, not a non-zero value, was incorrect.  Now the
swaybar tray can be used with systemd, elogind, or basu.


emersion (approved):
Good catch, thanks!

PR #5893, opened 2020-12-18
ISSUE_TEMPLATE: the wiki is community maintained

Emantor:
Mention that the github wiki may contain outdated information scripts
and old workarounds and that sway the sway project does not provide
support for it.


emersion:
Maybe we can add a note that users can just fix the wiki themselves?

emersion (approved):
Thanks!

Emantor:
> Maybe we can add a note that users can just fix the wiki themselves?

Added a note that users should adjust the wiki pages themselves :+1:

PR #5897, opened 2020-12-19
Restrict floating corners command line option (wip)

djvs:
This is definitely going to need some help, but you can see where I'm going with this.  Idea is to allow emulation of xmonad-style resizing behavior (e.g., always keep top-left corner static, resize at bottom-right corner).

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -15,6 +15,8 @@
 #include "sway/input/tablet.h"
 #include "sway/tree/root.h"
 #include "wlr-layer-shell-unstable-v1-protocol.h"
+#include <wlr/util/edges.h>
+
 
 // TODO: Refactor this shit
 
@@ -473,6 +475,7 @@ struct sway_config {
 	bool floating_mod_inverse;
 	uint32_t dragging_key;
 	uint32_t resizing_key;
+    enum wlr_edges resizing_corner;
 	char *floating_scroll_up_cmd;
 	char *floating_scroll_down_cmd;
 	char *floating_scroll_left_cmd;
--- a/sway/commands/resizing_corner.c
+++ b/sway/commands/resizing_corner.c
@@ -1,0 +1,31 @@
+#include "strings.h"
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/input/keyboard.h"
+
+struct cmd_results *cmd_resizing_corner(int argc, char **argv) {
+	struct cmd_results *error = NULL;
+	if ((error = checkarg(argc, "resizing_corner", EXPECTED_AT_LEAST, 1))) {
+		return error;
+	}
+
+    // Corresponds to WLR bitmasks (see edges.h)
+    if (strcasecmp(argv[1], "all") == 0) {
+      config->resizing_corner = WLR_EDGE_NONE;
+    } else if (strcasecmp(argv[1], "topright") == 0) {
+      config->resizing_corner = WLR_EDGE_TOP & WLR_EDGE_RIGHT;
+    } else if (strcasecmp(argv[1], "bottomright") == 0) {
+      config->resizing_corner = WLR_EDGE_BOTTOM & WLR_EDGE_RIGHT;
+    } else if (strcasecmp(argv[1], "bottomleft") == 0) {
+      config->resizing_corner = WLR_EDGE_BOTTOM & WLR_EDGE_LEFT;
+    } else if (strcasecmp(argv[1], "topleft") == 0) {
+      config->resizing_corner = WLR_EDGE_TOP & WLR_EDGE_LEFT;
+	} else {
+		return cmd_results_new(CMD_INVALID,
+				"Usage: resizing_corner [topright|bottomright|bottomleft|topleft|all]");
+	}
+
+	config->resizing_corner = corner;
+
+	return cmd_results_new(CMD_SUCCESS, NULL);
+}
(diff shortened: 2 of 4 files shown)


emersion:
Sorry, we're not implementing any new window management features unless i3 implements them too. Please consider searching for or opening an i3 feature request.

djvs:
This is not uncommon in other tiling WMs.  I just spent a good hour and a half PRing this, you want me to do it all over again for a different WM I'm not even using?

emersion:
No. I'm not asking you to implement that in i3. I'm asking you to convince i3 this is worth implementing. If it isn't, we generally don't merge these features in Sway.

djvs:
OK, well they refused it mostly on account of me not even wanting it for i3, but for sway:

https://github.com/i3/i3/issues/4286

This is seriously blocking me from using Wayland, is there really no way around this?

orestisfl:
> mostly on account of me not even wanting it for i3, but for sway

That's 100% unrelated to the reason that you feature request was rejected for i3. We don't base our decision for features on who is asking for them (with some exceptions for overwhelmingly popular support for a certain feature) but on the actual usecases of the feature vs implementation/maintenance costs.

djvs:
Apparently he did, cause that's what he said.  This decision-making is nuts, I don't really want to get involved - happy to help with this PR if this gets unfucked.

travankor:
>This is seriously blocking me from using Wayland at all, is there really no way around this?

You can use a different compositor, like [wayfire](http://github.com/wayfirewm/wayfire). From what I can tell, they're open to new window management functionality; you just need to use their plugin API and write a plugin.

djvs:
You could also just merge PRs that people want instead of locking your
featureset to a different project that yours is supposed to replace.

djvs:
Rewriting the entire featureset is not a reasonable solution for me or
pretty much anyone.

emersion:
>they refused it mostly on account of me not even wanting it for i3, but for sway

Read again the i3 thread. They said that they won't implement it because they don't think it's a goo feature to add. They also said they don't take the fact that someone requested the feature for Sway into account while deciding whether to accept it for i3.

You're free to fork Sway if you want this feature, nothing stops you from doing that. You can also contribute to other xmonad-like Wayland compositors, they need help.

In general Sway doesn't strive to merge all of the features from all existing WMs. Merging new features has a maintenance cost for us.

Airblader:
> They said that they won't implement it because they don't think it's a goo feature to add. 

For transparency, it's both. No, in its current state I don't think I'd accept the feature either way. But I also said I won't accept it because it isn't a feature request for i3.

> They also said they don't take the fact that someone requested the feature for Sway into account while deciding whether to accept it for i3.

Correct; I don't care whether the same request exists for sway. I do care, however, whether there's even an i3 user who wants a feature being brought up. :-)

emersion:
Okay, sorry for misunderstanding. We might need to change our policy here, but I don't know if there's a good way to let Sway contributors suggest new features while remaining at feature-parity with i3. In other words: I don't really want to add new WM features that i3 doesn't have, but OTOH there's no good path for Sway contributors to suggest new i3 features.

djvs:
Literally all I need is "only resize windows from the bottom right" that I've been using my entire adult life.  I have a fork ^ as you can see from the PR I took time out of my life to submit to you, I don't want to sit around pulling downstream changes onto it for the rest of my life.  Thanks a lot.

djvs:
This is an ergnomic nightmare by the way.  Have a nearly fullscreen window you dragged to the bottom right corner, that you want to resize to ~100x100?  You have to - with the touchpad! - drag it all the way to the center of the screen, move your mouse all the way past the center into the bottom left, and then drag it all the way back to the center to resize.  I'm trying to keep an eye on nerve damage to my wrist, and you're basically telling me to get fucked.

Airblader:
I can't speak for sway, but frequently using near fullscreen floating windows sounds like a tiling window manager is the wrong choice.

In any case, this topic is concluded for me. We did start an internal discussion regarding how to handle feature requests, however.

djvs:
Any large floating window.  I've been using TWMs over normal DEs since ~2008, I know it's the right choice, I'm just missing the float functionality here that actually complements the tiling functionality (this together with the snap-to-corner-on-resize that I mentioned).  No matter how big the window is you have a ton of extra added mileage for the mouse, which on the touchpad is your wrist pivoting and compressing your nerves, over and over.

smlx:
> Again, this is standard on xmonad, and for a good reason.

Maybe you should take a look at [waymonad](https://github.com/waymonad/waymonad), I'm sure they'd appreciate contributions.

djvs:
Waymonad is dead, and the author uses sway.

nmschulte:
To save on your wrist, you could probably script the behavior you're after.  I know Sway can `resize` a container, and can `floating toggle` a container.  I'm not sure but I guess you can `size` and `position` a container with the protocol as well.  I know `get_tree` will return all of the window information, but I'm sure there's a way to get just the window in focus.

```
swaymsg resize <shrink|grow> <direction> [amount] [unit]
```

djvs:
Thanks Nathan - not sure that'd work well with mouse resize though.  Already hard-coded this locally and resigned to just dealing with merge conflicts going forward.

emersion:
Our policy is now written down here: https://github.com/swaywm/sway/commit/fc1148da7f9c7bfc81f29e2b8304ae4e1592d95d

I also removed the part of our issue templates asking users to open i3 feature requests for window management features, so you shouldn't get more of these @Airblader.

PR #5899, opened 2020-12-20
grimshot: Fix man page issues

LEARAX:
This is a very minor documentation-only PR that tries to fix a few different issues in the grimshot man page. Most of the changes aim to be common sense, but I can revert or split up my reorganization/rewording changes if they aren't appreciated.

The main issues that prompted this PR are the swapped labels for the SYNOPSIS and DESCRIPTION sections and the absence of line breaks in the usage line:

https://github.com/swaywm/sway/blob/5f1fe33d369b88539c9fb434703c1c00cd26c903/contrib/grimshot.1#L14-L18
gets formatted as:
```
DESCRIPTION
       grimshot [--notify] (copy|save) [TARGET] [FILE] grimshot check grimshot usage
```

Please verify and let me know if any changes I've made are in error.


WhyNotHugo (approved):
LGTM

PR #5901, opened 2020-12-20
Add no_titlebars command

insidewhy:
I think a lot of people only use a small number of windows in each workspace and can remember which windows they have in each workspace so don't need the dedicated titlebar space.

I know this is something the i3 maintainers are not interested in so it probably won't fit the sway philosophy either. Several people are asking for it so I figured I'd open this PR at least so that people can find the code if they need it. Feel free to close.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -160,6 +160,7 @@ sway_cmd cmd_opacity;
 sway_cmd cmd_new_float;
 sway_cmd cmd_new_window;
 sway_cmd cmd_no_focus;
+sway_cmd cmd_no_titlebars;
 sway_cmd cmd_output;
 sway_cmd cmd_permit;
 sway_cmd cmd_popup_during_fullscreen;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -511,6 +511,7 @@ struct sway_config {
 
 	bool tiling_drag;
 	int tiling_drag_threshold;
+	bool no_titlebars;
 
 	bool smart_gaps;
 	int gaps_inner;
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -78,6 +78,7 @@ static struct cmd_handler handlers[] = {
 	{ "new_float", cmd_new_float },
 	{ "new_window", cmd_new_window },
 	{ "no_focus", cmd_no_focus },
+	{ "no_titlebars", cmd_no_titlebars },
 	{ "output", cmd_output },
 	{ "popup_during_fullscreen", cmd_popup_during_fullscreen },
 	{ "seat", cmd_seat },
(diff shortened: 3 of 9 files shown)


tadeokondrak:
I might be missing something, but isn't this just `border pixel`?

```
       border none|normal|csd|pixel [<n>]
           Set border style for focused window. normal includes a border of thickness n and a title
           bar. pixel is a border without title bar n pixels thick. Default is normal with border
           thickness 2. csd is short for client-side-decorations, which allows the client to draw its
           own decorations.
```

insidewhy:
@tadeokondrak I tried the `border pixel 0` command via swaymsg and the titlebars are still rendered. Nothing changed there actually.

dimitris-triantafyllidis:
Did you try `default_border` and `default_floating_border`? If you just use `border` you should at least add criteria since it is just a command.

insidewhy:
This doesn't seem to do anything for me:

```
default_border pixel 0
default_floating_border pixel 0
```

Xyene:
This definitely seems a lot like `default_border pixel 0`. Note that if you change it at runtime through `swaymsg`, already-open windows do not have their borders changed; the setting only applies to new windows.

image

This change goes a bit further than this though, since `container_titlebar_height` is also used to calculate the height of tab/stack headers, which you (probably?) don't actually want to get rid of.

If `default_border pixel 0` is not working for you, it would be worth figuring out why, because it should. Otherwise, if the intent behind this patch is to also unconditionally hide tab/stack headers, then I agree this not something we'd want to merge (but others are free to find and apply it locally if it's something they really want).

insidewhy:
The tab/stack headers are exactly what I and many others want to get rid of. Aren't those the "titlebar"? We don't need to be reminded of which windows are open on each workspace when we can remember, so we want to reclaim the wasted space.

insidewhy:
It's already possible to remove them by setting font size to 0, and border padding and size to 0. Then the code is still trying to render the headers, which feels weird.

This patch is pretty small, maybe I was mostly trying to get a feel for how dogmatic those in control are about mandating how things should work. I agree that endless configuration options can be annoying, but maybe some kind of alternative with pluggable/scriptable decorations might be something that will let people decide for themselves how things can look and work instead of being told.

I came here from KDE because I was annoyed by the endless bugs, and those very well could be due to the level of configurability that WM offers. On the other hand I really don't like how Mac OS offers so little control and configurability. I wonder if there's a nice middle ground out there.

emersion:
Sorry, we don't accept patches for this kind of feature anymore, see https://github.com/swaywm/sway/commit/fc1148da7f9c7bfc81f29e2b8304ae4e1592d95d. Feel free to keep this patch in a fork ofc.

PR #5909, opened 2020-12-25
Revert removal of the raise_floating directive

diamondburned:
This pull request reverts commit b9b1b0e5662f8e3dbbfa9bb09dd4f69aac9c2db0, and therefore pull request #2888. The reason being that with some panels that list windows (such as `wf-panel`), focusing on a window won't raise it up front, which is weird.

However, this pull request also sets the default value for `raise_floating` to `false`, to not break existing setups that expect the old behavior.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -131,6 +131,7 @@ sway_cmd cmd_floating_modifier;
 sway_cmd cmd_floating_scroll;
 sway_cmd cmd_focus;
 sway_cmd cmd_focus_follows_mouse;
+sway_cmd cmd_raise_floating;
 sway_cmd cmd_focus_on_window_activation;
 sway_cmd cmd_focus_wrapping;
 sway_cmd cmd_font;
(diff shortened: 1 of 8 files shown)


emersion:
This is a new WM feature, which we generally don't accept anymore.

Also, this doesn't explain how this is addressed:

> For example, if you have two floating views where one completely covers the other, the smaller one would be inaccessible because moving the mouse over the bigger one would raise it above the smaller one.

diamondburned:
> Also, this doesn't explain how this is addressed:
> 
> > For example, if you have two floating views where one completely covers the other, the smaller one would be inaccessible because moving the mouse over the bigger one would raise it above the smaller one.

By default, it is disabled, so those who want the feature can just enable it. I don't see why a feature should be removed if users who use it differently can misuse it, since not everyone has focus on mouse hover.

> This is a new WM feature, which we generally don't accept anymore.

Should I close the pull request then?

emersion:
>Should I close the pull request then?

As per https://github.com/swaywm/sway/commit/fc1148da7f9c7bfc81f29e2b8304ae4e1592d95d, we now see sway's window management features as complete. Sorry about that, but ofc feel free to keep this in a fork.

PR #5910, opened 2020-12-25
tray: do not render passive items

xi:
From the spec:

> Passive: The item doesn't convey important information to the user, it can be considered an "idle" status and is likely that **visualizations will chose to hide it**.

https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/#org.freedesktop.statusnotifieritem.status

### DIFF
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -461,6 +461,11 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
 		sni->target_size = target_size;
 	}
 
+	// Passive
+	if (sni->status && sni->status[0] == 'P') {
+		return 0;
+	}
+
 	int icon_size;
 	cairo_surface_t *icon;
 	if (sni->icon) {


ianyfan:
This looks good, but do you have an example of an application that uses this, just as a reference? Thanks

xi:
I personally use it to show an indicator if there are notifications: https://github.com/xi/notification-hub

Apart from that I do not really have examples, sorry.

emersion:
Is there another bar that hides passive items? Have you tried your program outside of Sway?

xi:
Before switching to sway I used lxpanel, which does hide passive indicators. I also checked mate-panel which hides passive indicators, too. I have not tested any bars/panels so far.

progandy:
A future improvement may be to create a toggle for that (e.g. click inside status area, but outside of an existing icon). 

For example windows has a popup window where it collects idle tray icons.

emersion (approved):
Sorry for the delay. Thanks!

PR #5914, opened 2020-12-26
grimshot: unbreak saving with default filename on FreeBSD

jbeich:
Regressed by 61d59180b875. CC @WhyNotHugo

```
$ sh contrib/grimshot save active
date: invalid argument 'ns' for -I
/home/foo/.png
```

`-I` is not in [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/date.html) and not supported by [DragonFly](https://www.dragonflybsd.org/cgi/web-man?command=date&section=1), [NetBSD](https://man.netbsd.org/date.1), [OpenBSD](https://man.openbsd.org/date.1). On [FreeBSD](https://man.freebsd.org/date/1) `-I` [appears](https://svnweb.freebsd.org/changeset/base/337332) since 12.0 (still missing in 11.4) but 13.0 doesn't support `-I ns` (nanoseconds) or `%N`.

Additionally, `-I` uses `:` to separate time which makes such files harder to share with Windows.

### DIFF
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -12,11 +12,20 @@
 ##
 ## See `man 1 grimshot` or `grimshot usage` for further details.
 
-getTargetDirectory() {
+get_target_directory() {
   test -f "${XDG_CONFIG_HOME:-$HOME/.config}/user-dirs.dirs" && \
     . "${XDG_CONFIG_HOME:-$HOME/.config}/user-dirs.dirs"
 
   echo "${XDG_SCREENSHOTS_DIR:-${XDG_PICTURES_DIR:-$HOME}}"
+}
+
+get_target_file() {
+  NAME=$(date +%Y%m%d_%H%M%S)
+  DIR=${1:-$(get_target_directory)}
+  while [ -e "$DIR/$NAME$SUFFIX.png" ]; do
+    SUFFIX="_$((i+=1))"
+  done
+  echo "$DIR/$NAME$SUFFIX.png"
 }
 
 NOTIFY=no
@@ -42,7 +51,7 @@ done
 
 ACTION=${1:-usage}
 SUBJECT=${2:-screen}
-FILE=${3:-$(getTargetDirectory)/$(date -Ins).png}
+FILE=${3:-$(get_target_file)}
 
 if [ "$ACTION" != "save" ] && [ "$ACTION" != "copy" ] && [ "$ACTION" != "check" ]; then
   echo "Usage:"


contrib/grimshot
```diff
@@ -28,7 +28,7 @@ fi
 
 ACTION=${1:-usage}
 SUBJECT=${2:-screen}
-FILE=${3:-$(getTargetDirectory)/$(date -Ins).png}
+FILE=${3:-$(getTargetDirectory)/$(date -Ins 2>/dev/null || date +%Y%m%d_%Hh%Mm%Ss_grim).png}
```
WhyNotHugo, line 31: This generates a very funny-looking filename and I don't see why files should be named `grim`.

```console
$ date +%Y%m%d_%Hh%Mm%Ss_grim
20201226_17h09m53s_grim
```

This also has the issue of overwriting consecutive screenshots taken within one second.

```suggestion
FILE=${3:-$(getTargetDirectory)/$(date +%Y%m%dT%H%M%S.%N).png}
```

I think this should work fine on all platforms, and avoid having diverging / per-platform variations. This looks like:

> 20201226T160853.228348830

WhyNotHugo (commented):
> Additionally, -I uses : to separate time which makes such files harder to share with Windows.

What do you mean by this? Does `sway` even actually run on windows? What exactly is the issue here?

contrib/grimshot
jbeich: `%N` is a GNU extension not supported by `strftime` anywhere but a special-cased in GNU date. See strftime in [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strftime.html), [Linux](https://www.man7.org/linux/man-pages/man3/strftime.3.html), [FreeBSD](https://man.freebsd.org/strftime/3), [DragonFly](https://www.dragonflybsd.org/cgi/web-man?command=strftime&section=3), [NetBSD](https://man.netbsd.org/strftime.3), [OpenBSD](https://man.openbsd.org/strftime.3).

jbeich:
> What do you mean by this? 

```
$ grimshot save active
/home/foo/2020-12-26T16:26:50+0000.png
$ mv ~/2020-12-26T16:26:50+0000.png /path/to/samba-share
$ ssh bhyve-windows10
$ dir \\host\samba-share
[...]
26/12/2020  16:26           220.744 2U263W~X.PNG
[...]
```

contrib/grimshot
WhyNotHugo: Oh, you're right, my bad.

Is there no way to generate a timestamp _with_ milliseconds? Taking more than one screenshot in under one second is a pretty valid use-case, and I'd rather not break that.

I do agree that `grimshot` should be compatible with POSIX's `date` though.

WhyNotHugo:
Looking at the man page for `dir`, I've no idea what your above command is supposed to do.
It seems to do the same as `ls`, right? Does `ls` work for you?

Looks to me like you're just hitting some weird bug in either `samba` or `dir`, have you tried reporting this there?

Xyene:
The behavior above is not a bug; Samba must mangle filenames containing `:` because `:` is an invalid character in Windows filenames (and SMB is, first and foremost, a Windows protocol). I assume the `ssh bhyve-windows10` gets @jbeich into a Windows machine, and `dir` is Windows' `dir` command showing the mangled names. It would be a bit sad if this use-case didn't work.

See also https://github.com/swaywm/sway/pull/5840

An (untested) suggestion:

```sh
if [ -n "$3" ]; then
	FILE="$3"
else
	NAME="date +%Y%m%d_%Hh%Mm%Ss"
	DIR="$(getTargetDirectory)"
	SUFFIX="$(ls "${DIR}" | wc -l)"
	if [ "${SUFFIX}" -gt 0 ]; then
		NAME="${NAME}_${SUFFIX}"
	fi
	FILE="${DIR}/${NAME}.png"
fi
```

WhyNotHugo:
Ooh, thanks @Xyene.

I guess the question is, do we really want to change implementations in sway to work around bugs or limitations on Windows?

The above suggestion is pretty neat though, but seems to have a few duplicate letters that end up in the filename though. This diff seems to cover all cases and issues:

```diff
diff --git a/contrib/grimshot b/contrib/grimshot
index 518d85f8..4e19de79 100755
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -12,7 +12,7 @@
 ##
 ## See `man 1 grimshot` or `grimshot usage` for further details.

-getTargetDirectory() {
+get_target_directory() {
   test -f ${XDG_CONFIG_HOME:-~/.config}/user-dirs.dirs && \
     . ${XDG_CONFIG_HOME:-~/.config}/user-dirs.dirs

@@ -28,11 +28,21 @@ fi

 ACTION=${1:-usage}
 SUBJECT=${2:-screen}
-FILE=${3:-$(getTargetDirectory)/$(date -Ins).png}
+if [ -n "$3" ]; then
+  FILE="$3"
+else
+  NAME="date +%Y%m%d_%H%M%S"
+  DIR="$(get_target_directory)"
+  SUFFIX="$(ls "${DIR}" | grep "${NAME}" | wc -l)"
+  if [ "${SUFFIX}" -gt 0 ]; then
+    NAME="${NAME}_${SUFFIX}"
+  fi
+  FILE="${DIR}/${NAME}.png"
+fi

 if [ "$ACTION" != "save" ] && [ "$ACTION" != "copy" ] && [ "$ACTION" != "check" ]; then
   echo "Usage:"
```

jbeich:
@WhyNotHugo wrote:

>  ACTION=${1:-usage}
>  SUBJECT=${2:-screen}
> -FILE=${3:-$(getTargetDirectory)/$(date -Ins).png}
> +if [ -n "$3" ]; then
> +  FILE="$3"
> +else

Looks harder to read, so I've factored it out into `get_target_file` function in order to preserve existing style.

> +  NAME="date +%Y%m%d_%H%M%S"

Obviously missing `$(...)` aka "command substitution".

> +  SUFFIX="$(ls "${DIR}" | grep "${NAME}" | wc -l)"

- ls(1) output isn't very stable
- grep(1) can be replaced by `case (pattern) esac` in sh(1) scripts
- `${NAME}` may match unrelated files with different suffix

I've replaced this part with an iterator.

contrib/grimshot
```diff
@@ -12,13 +12,22 @@
...
 }
 
+get_target_file() {
+  NAME=$(date +%Y%m%d_%H%M%S)
+  DIR=${1:-$(get_target_directory)}
+  while [ -e "$DIR/$NAME$SUFFIX.png" ]; do
+    SUFFIX="_$((i+=1))"
+  done
+  echo "$DIR/$NAME$SUFFIX.png"
+}
```
WhyNotHugo, line 29: This might be raceful if somebody is taking lots of screenshots in succession (e.g.: when trying to catch a quirk in an animation or something alike). Two instances of grimshot will run, detect no file, and then both write (at the same time) into the new filename.

I'm not saying we should block this merge due to this issue, but mostly pointing it out. I'm not currently using `grimshot`, so have no strong opinion.

PR #5915, opened 2020-12-26
ci: enable basu on freebsd

jbeich:
Avoid swaybar tray growing dependencies not supported by elogind and basu.
https://github.com/swaywm/sway/blob/7c9706de7186f32159f719e3f15fcdc017a13f35/.builds/alpine.yml#L31
https://github.com/swaywm/sway/blob/7c9706de7186f32159f719e3f15fcdc017a13f35/.builds/archlinux.yml#L28

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -1,5 +1,6 @@
 image: freebsd/latest
 packages:
+- devel/basu
 - devel/json-c
 - devel/libevdev
 - devel/meson
@@ -32,7 +33,7 @@ tasks:
     cd subprojects
     ln -s ../../wlroots wlroots
     cd ..
-    meson build
+    meson build -Dtray=enabled -Dsd-bus-provider=basu
 - build: |
     cd sway
     ninja -C build


.builds/freebsd.yml
```diff
@@ -1,5 +1,6 @@
 image: freebsd/latest
 packages:
+- devel/basu
```
jbeich, line 3: The package is still missing probably because I've recently moved it.
https://svnweb.freebsd.org/changeset/ports/559264

emersion (approved):
LGTM! Please ping when the change is propagated to pkg repos, so that I can kick off the CI again.

.builds/freebsd.yml
```diff
@@ -32,7 +33,7 @@ tasks:
     cd subprojects
     ln -s ../../wlroots wlroots
     cd ..
-    meson build
+    meson build -Dtray=enabled -Dsd-bus-provider=basu
```
jbeich, line 36: Can't use `-Dauto_features=enabled` instead of `-Dtray=enabled` due to https://github.com/mesonbuild/meson/issues/5320.

emersion:
Thanks!

PR #5921, opened 2020-12-29
Transient seat

any1:
This implements transient seats for sway. See: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3870


any1:
Ping

emersion (approved):
Fixed a minor style nit (`{` needs to be on the same line as the function decl).

LGTM, thanks!

emersion (commented):
Ah, I also added the global to the list of privileged protocols in `is_privileged()`.

PR #5934, opened 2021-01-07
Add last_focused timestamp to containers

tsdh:
A new field `struct timespec last_focused` is added to the `sway_container_state` struct holding a UNIX timestamp telling when this container last gained focus.  This timestamp is added to the IPC `get_tree` JSON output.

The rationale is that it allows users writing their own container manipulation scripts (e.g., window switchers) to sort containers by last-recently-used order.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -52,6 +52,7 @@ struct sway_container_state {
 
 	struct sway_container *focused_inactive_child;
 	bool focused;
+	struct timespec last_focused;
 
 	enum sway_container_border border;
 	int border_thickness;
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -152,6 +152,13 @@ static void copy_container_state(struct sway_container *container,
 
 	struct sway_seat *seat = input_manager_current_seat();
 	state->focused = seat_get_focus(seat) == &container->node;
+	if (state->focused) {
+		struct timespec ts;
+		timespec_get(&ts, TIME_UTC);
+		state->last_focused = ts;
+	} else {
+		state->last_focused = container->current.last_focused;
+	}
 
 	if (!container->view) {
 		struct sway_node *focus =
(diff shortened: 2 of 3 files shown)


emersion:
Sorry, we're not accepting any new window management features anymore. See https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#scope-of-future-changes-to-sway

tsdh:
@emersion Yes, I've read that including that only "**most** new window management features" won't be accepted and my opinion was that this is no actual window management feature but rather a tiny and trivial bit of "window management self-explanation".  Of course, YMMV.

Anyway, what would be the way to go?  Trying to make a similar PR for i3 so that this PR becomes a compatibility issue or do you think this "feature" is unworthy (e.g., because the information can be deduced differently already, in which case I happily let go of this PR; essentially, I just want a reliable, easy way to switch between the two least recently used windows)?

emersion:
I think your use-case doesn't need this feature. An IPC client can track container focus changes and record timestamps itself.

tsdh:
Ok, I'll close this PR then.

But I'm keen to hear more about the IPC client approach.  Do you have in mind that I should write a program that subscribes to window events and uses that to maintain a mapping from container id to last-focused timestamp which then my window switcher (which currently is just `swaymsg -t get_tree | jq ... | wofi ... | sed ... | swaymsg "[con_id=$id]" focus`) can utilize for sorting in LRU order?

emersion:
> Do you have in mind that I should write a program that subscribes to window events and uses that to maintain a mapping from container id to last-focused timestamp which then my window switcher (which currently is just swaymsg -t get_tree | jq ... | wofi ... | sed ... | swaymsg "[con_id=$id]" focus) can utilize for sorting in LRU order?

Yup, something like this.

tsdh:
@emersion Ok, I see.  Quite involved task for such a rudimentary requirement.  I guess I need to leave shell scripting territory for implementing that...

emersion:
Yeah, people often use a Python script to interact with the Sway IPC. Libraries for other programming languages are available as well.

Issue #5902, opened 2020-12-21
Respond to changed modes on DRM connectors (e.g. resized/maximised virtual machine window)


hedgepigdaniel:
When running Sway in a virtual machine (unlike on a physical machine), it is common for the preferred DRM mode to change. This happens when resizing/maximizing the virtual machine window in the host system.

Currently, the DRM backend chooses a mode at startup and never changes it without user intervention [unless the DRM connector is disabled and reenabled](https://github.com/swaywm/wlroots/blob/de896caceb9f6354a99c9435c3a429ff52b6ca4a/backend/drm/drm.c#L1311-L1312), even though there is a udev event and `scan_drm_connectors` is called.

Instead, I think it would be better if the mode was automatically changed when necessary/appropriate as a result of changes to available DRM modes. I'm not sure exactly what the criteria should be for when the mode should change automatically, but I think my case is clearly one where it should:
 - The current mode is no longer in the list of modes on the DRM connector
 - There is no configured mode, and no mode was set manually - the current mode is set only by default because it was the preferred mode at the time

Would this be a welcome improvement?

Emantor:
We would need code to detect that we are running in a virtual machine, register for drm invalidation events and than change mode to the new preferred mode.

emersion:
Duplicate of https://github.com/swaywm/wlroots/issues/2188, I think.

PR #5942, opened 2021-01-10
output: reapply output config when fixed modes change

hedgepigdaniel:
Fixes https://github.com/swaywm/sway/issues/5902

Depends on wlroots change: https://github.com/swaywm/wlroots/pull/2628

Add a handler for the `available_modes` signal from wlroots and reapply the output config in the handler.

The effect is that if there is no configured output mode, then the mode is updated to match the new preferred mode of the output. If the user has configured an output mode, then there is no effect.

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -44,6 +44,7 @@ struct sway_output {
 	struct wl_listener destroy;
 	struct wl_listener commit;
 	struct wl_listener mode;
+	struct wl_listener available_modes;
 	struct wl_listener present;
 	struct wl_listener damage_destroy;
 	struct wl_listener damage_frame;
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -872,6 +872,17 @@ static void handle_mode(struct wl_listener *listener, void *data) {
 	update_output_manager_config(output->server);
 }
 
+static void handle_available_modes(struct wl_listener *listener, void *data) {
+	struct sway_output *output = wl_container_of(listener, output, available_modes);
+	struct output_config *oc = find_output_config(output);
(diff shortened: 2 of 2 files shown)


emersion:
I'm worried about a real driver changing the mode list. I don't want that to trigger an automatic modeset.

hedgepigdaniel:
I'm assuming that reapplying the existing config will never trigger a modeset unless the hardware situation has changed? Wouldn't it be a driver bug, if the modes changed for a physical output with a hardware edid?

We could instead only reapply the config if the output has the `hotplug_mode_update` DRM prop set - (e.g. by not updating the available modes in wlroots unless that prop was set, or alternatively by adding a member `hotplug_mode_update` to `struct wlr_output` so that Sway can make the decision). Would either of those be better?

emersion:
>Wouldn't it be a driver bug, if the modes changed for a physical output with a hardware edid?

Not necessarily. The kernel can prune modes, e.g. when the physical link has partially failed and cannot support higher resolutions anymore.

- If the link has partially failed and the current mode is still usable, the kernel will prune some modes and everything will continue to work normally. The current mode could disappear from the list but still continue to work, I think, because of mode flag changes for instance.
- If the link has failed and the current mode can't be used anymore, the kernel will set `link-status = BAD` to signal it. We already handle that in wlroots.

So yeah, I'd prefer to perform a modeset only if `hotplug_mode_update = 1`. Related: https://github.com/swaywm/wlroots/issues/2300#issuecomment-761819673.

hedgepigdaniel:
Closed in favour of https://github.com/swaywm/wlroots/pull/2712

PR #5944, opened 2021-01-10
output: use suggested position for output by default

hedgepigdaniel:
Depends on and includes https://github.com/swaywm/sway/pull/5942 (could also be independently rebased onto master)

Depends on wlroots API: https://github.com/swaywm/wlroots/pull/2629

Some virtualised outputs provide a suggested position which indicates the relative position of the outputs in the host WM. This change uses these positions for outputs with no configured position.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -331,6 +331,18 @@ static int compute_default_scale(struct wlr_output *output) {
 	return 2;
 }
 
+static float get_effective_scale(struct output_config *oc,
+		struct wlr_output *wlr_output) {
+	float scale;
+	if (oc && oc->scale > 0) {
+		scale = oc->scale;
+	} else {
+		scale = compute_default_scale(wlr_output);
+		sway_log(SWAY_DEBUG, "Auto-detected output scale: %f", scale);
+	}
+	return scale;
+}
+
 static void queue_output_config(struct output_config *oc,
 		struct sway_output *output) {
 	if (output == root->noop_output) {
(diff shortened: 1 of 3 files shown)


sway/config/output.c
```diff
@@ -451,9 +457,24 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 	}
 
 	// Find position for it
-	if (oc && (oc->x != -1 || oc->y != -1)) {
-		sway_log(SWAY_DEBUG, "Set %s position to %d, %d", oc->name, oc->x, oc->y);
-		wlr_output_layout_add(root->output_layout, wlr_output, oc->x, oc->y);
+	int32_t effective_x, effective_y = -1;
+	float effective_scale = get_effective_scale(oc, output->wlr_output);
+	if (output->wlr_output->suggested_x != -1) {
+		effective_x = output->wlr_output->suggested_x / effective_scale;
```
hedgepigdaniel, line 463: This assumes that all the outputs have the same scale. If this output has a position relative to another output, its the other output's scale that is relevant.

There's no good way to know which output, so if the outputs don't all have the same scale, then the suggested position properties don't make sense. My working theory is that for virtualisation software that provides the suggested position, the outputs are also scaled by the host system, and hence it would not be desirable to have different scales in a compositor in the virtualised system.

I've tested this on VMware Fusion with a combination of high and low DPI displays. In this case the virtualised outputs are all high DPI (the resolution of the low DPI physical display is doubled for the corresponding virtual display).

So a path forward could be to make this assumption explicit, and only use the suggested position if all outputs have the same scale.

PR #5945, opened 2021-01-11
swaybar: Fallback to focused_statusline instead of statusline on focused output

### LINKED ISSUE #5940: swaybar: focused_statusline seems to be ignored when using swaybar-protocol
- **Sway Version:** `sway version 1.5.1`
- **Debug Log:** http://ix.io/2LqH
- **Configuration File:**
  - **Sway Config:** http://ix.io/2LqE (default config with slight changes to the bar section)
  - **Statusbar Script:** http://ix.io/2LqG
- **Description:**
`focused_statusline` seems to be ignored when the `status_command` uses the swaybar-protocol. This problem does not seem to happen when the `status_command` is set to something like `while date +'%Y-%m-%d %l:%M:%S %p'; do sleep 1; done` (which does not use the swaybar-protocol).

### Screenshots

Bars when `status_command` **uses** swaybar-protocol:
bar_3
bar_2
bar_1

Bars when `status_command` **does not use** swaybar-protocol:
bar_3
bar_2
bar_1

Mukundan314:
Fixes #5940

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -269,7 +269,9 @@ static uint32_t render_status_block(cairo_t *cairo,
 	}
 	double text_y = height / 2.0 - text_height / 2.0;
 	cairo_move_to(cairo, offset, (int)floor(text_y));
-	uint32_t color = block->color_set ? block->color : config->colors.statusline;
+	uint32_t color = output->focused ?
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

PR #5947, opened 2021-01-12
Support cursor capture in grimshot

thecaralice:
Uses grim `-c` flag

Cursor capture is enabled by `--cursor` flag. When used with `--notify` it can be either `--notify --cursor` or `--cursor --notify`, but only the the former is documented.


contrib/grimshot
```diff
@@ -26,13 +26,24 @@ else
...
+if [ "$1" = "--cursor" ]; then
+  CURSOR="-c"
+  shift 1
+fi
+
+
+if [ "$1" = "--notify" ] && [ "$NOTIFY" = "no" ]; then
+  NOTIFY=yes
+  shift 1
+fi
```
WhyNotHugo, line 38: In order to avoid repeating this block (that's already checked above), I'd recommend doing something like:

```bash
POSITIONAL=()
NOTIFY=no
CURSOR=no

while [[ $# -gt 0 ]]; do
  key="$1"

  case $key in
    -n|--notify)
      NOTIFY=yes
      shift # past argument
      ;;
    -c|--cursor)
      CURSOR=yes
      shift # past argument
      ;;
    *)    # unknown option
      POSITIONAL+=("$1") # save it in an array for later
      shift # past argument
      ;;
  esac
done
```

Inspired by https://stackoverflow.com/a/14203146

I also think that we could make `--copy` and `--save` into flags following the same pattern -- I recall another PR or issue wanting to allow a "copy + save" action, and this would be a good approach/

thecaralice: Sorry for late reply

> I also think that we could make `--copy` and `--save` into flags following the same pattern -- I recall another PR or issue wanting to allow a "copy + save" action, and this would be a good approach/

Would it make sense to do that in this PR or I should make a new one?

contrib/grimshot
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
WhyNotHugo, line 1: Is bash really necessary? Given the variety of users of this, I'm sure there's people who won't have it installed (e.g.: some alpine and BSD users).

(in any case, this should be `/usr/bin/env bash`)

contrib/grimshot
WhyNotHugo: I think it's best to address that in a separate PR to keep thing in scope here.

contrib/grimshot
```diff
@@ -13,10 +13,10 @@
 ## See `man 1 grimshot` or `grimshot usage` for further details.
 
 getTargetDirectory() {
-  test -f ${XDG_CONFIG_HOME:-~/.config}/user-dirs.dirs && \
-    . ${XDG_CONFIG_HOME:-~/.config}/user-dirs.dirs
+  test -f "${XDG_CONFIG_HOME:-~/.config}/user-dirs.dirs" && \
+    . "${XDG_CONFIG_HOME:-~/.config}/user-dirs.dirs"
 
-  echo ${XDG_SCREENSHOTS_DIR:-${XDG_PICTURES_DIR:-$HOME}}
+  echo "${XDG_SCREENSHOTS_DIR:-${XDG_PICTURES_DIR:-$HOME}}"
```
WhyNotHugo, line 19: :+1:

contrib/grimshot
```diff
@@ -13,26 +13,46 @@
...
+while [[ $# -gt 0 ]]; do
+  key="$1"
+
+  case $key in
+    -n|--notify)
+      NOTIFY=yes
+      shift # past argument
+      ;;
+    -c|--cursor)
+      CURSOR=yes
```
WhyNotHugo, line 35: How about setting this to `CURSOR="-c"`...

contrib/grimshot
```diff
@@ -91,12 +111,12 @@ takeScreenshot() {
   FILE=$1
   GEOM=$2
   OUTPUT=$3
-  if [ ! -z "$OUTPUT" ]; then
-    grim -o "$OUTPUT" "$FILE" || die "Unable to invoke grim"
+  if [ -n "$OUTPUT" ]; then
+    grim ${CURSOR:+-c} -o "$OUTPUT" "$FILE" || die "Unable to invoke grim"
```
WhyNotHugo, line 115: ... And then just using `$CURSOR` here?

I believe this syntax here is the only bashism, right?

WhyNotHugo (commented):
LGTM, I only have one minor comment pending...

contrib/grimshot
thecaralice: Yeah, you are right, but I can't think of a shebang for something like "POSIX sh-based shell supporting arrays" (see [SC3054](https://github.com/koalaman/shellcheck/wiki/SC3054))

contrib/grimshot
thecaralice: As I wrote above bash is used because of the array, so this expansion is not *the* bashism or the most significant one; actually I'm not sure if it's a bashism

thecaralice:
I'll try to get rid of any sh extensions in a separate commit

contrib/grimshot
```diff
@@ -13,26 +13,46 @@
...
+      shift # past argument
+      ;;
+    *)    # unknown option
+      POSITIONAL+=("$1") # save it in an array for later
+      shift # past argument
+      ;;
+  esac
+done
+
+printf "NOTIFY: %s\nCURSOR: %s\nPOSITIONAL: %s" "$NOTIFY" "$CURSOR" "$(args "${POSITIONAL[@]}")"
```
thecaralice, line 45: oops, i forgot to remove this

contrib/grimshot
thecaralice: Well, it doesn't seem to be a bashism — at least it's supported on `dash` and `shellcheck` doesn't report it as an extension

contrib/grimshot
```diff
@@ -13,26 +13,40 @@
...
+  echo "${XDG_SCREENSHOTS_DIR:-${XDG_PICTURES_DIR:-$HOME}}"
 }
 
-if [ "$1" = "--notify" ]; then
-  NOTIFY=yes
-  shift 1
-else
-  NOTIFY=no
-fi
+NOTIFY=
```
WhyNotHugo, line 22: You're setting notify to an empty string here, but the [existing] check below is `"$NOTIFY" = "no"`.

Running `grimshot --cursor save active` showed a notification, when it shouldn't.

WhyNotHugo (changes requested):
I tried using `dash` as a shell and it work :+1:

contrib/grimshot
thecaralice: oops

WhyNotHugo (approved):
LGTM. Thanks for all the cleanups done too!

emersion:
This PR makes some changes then undoes them in later commits. Can you cleanup the Git history?

thecaralice:
Sure.

WhyNotHugo (approved):
Rebase/squash still LGTM.

PR #5948, opened 2021-01-12
Adhere to X11 focus requirements

BrassyPanache:
When mapping an X11 view the focus state should be retrieved via the
existing mechanisms to determine focus. For certain views if this check
is not completed they will have initial focus when it is undesired.

Relates to swaywm/wlroots#2604

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -785,7 +785,18 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 
 	view_execute_criteria(view);
 
-	if (should_focus(view)) {
+	bool set_focus = should_focus(view);
+
+#if HAVE_XWAYLAND
+	if (wlr_surface_is_xwayland_surface(wlr_surface)) {
+		struct wlr_xwayland_surface *xsurface =
+			wlr_xwayland_surface_from_wlr_surface(wlr_surface);
+    	set_focus = (wlr_xwayland_icccm_input_model(xsurface) !=
+			WLR_ICCCM_INPUT_MODEL_NONE) && set_focus;
+	}
+#endif
+
+	if (set_focus) {
 		input_manager_set_focus(&view->container->node);
 	}
 


tsdh:
@BrassyPanache I have test-driven this PR in combination with https://github.com/swaywm/wlroots/pull/2605.  Now the "New Class" popup has input focus again, and also "Navigate > File Structure" doesn't vanish immediately.  That's a big relief!

Aside from this positive effects which enable me to use sway at work where I have to use IntelliJ, I haven't encountered any issues / bad side-effects so far.  I'll keep using those two PR-branches (and update them daily in case you push changes) and report back if I find quirks.

Also, feel free to ask for specific tests I should try, if there are any.  AFAIK, these changes only affect XWayland clients, no?  I'm not using too many of those except for Emacs, IntelliJ, and sometimes SquirrelSQL.

BrassyPanache:
Thanks for the detailed feedback @tsdh!  That's really appreciated.  I have no specifics to ask for - but if you see anything out of the ordinary in your typical workflow it'd be great if you could document it.

sway/tree/view.c
```diff
@@ -785,9 +785,21 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 
 	view_execute_criteria(view);
 
+
+#if HAVE_XWAYLAND
+	if (wlr_surface_is_xwayland_surface(wlr_surface)) {
+		if (wlr_xwayland_or_surface_wants_focus(
```
emersion, line 791: `wlr_xwayland_or_surface_wants_focus` is for override-redirect windows only (that's why there's "or" in the name), so I'm not comfortable using it for regular windows.

Maybe we could add a wlroots helper that returns the input model for a surface (as described in https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.7)?

sway/tree/view.c
```diff
@@ -785,9 +785,21 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 
 	view_execute_criteria(view);
 
+
+#if HAVE_XWAYLAND
```
emersion, line 789: I feel like the logic could be simplified here, perhaps with something like:

```c
bool set_focus = should_focus(view);
if (wlr_surface_is_xwayland_surface(wlr_surface)) {
    set_focus = set_focus || wlr_xwayland_or_surface_wants_focus(…);
}
if (set_focus) {
    input_manager_set_focus(…);
}
```

sway/tree/view.c
BrassyPanache: Hey no worries.  I missed that comment.  I read it simply as "or" :smile:

sway/tree/view.c
BrassyPanache: Yep! Agree.

sway/tree/view.c
```diff
@@ -785,7 +785,18 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
...
 	view_execute_criteria(view);
 
-	if (should_focus(view)) {
+	bool set_focus = should_focus(view);
+
+#if HAVE_XWAYLAND
+	if (wlr_surface_is_xwayland_surface(wlr_surface)) {
+		struct wlr_xwayland_surface *xsurface =
+			wlr_xwayland_surface_from_wlr_surface(wlr_surface);
+    	set_focus = (wlr_xwayland_icccm_input_state(xsurface) !=
```
emersion, line 794: This overwrites the previous value of `set_focus`.

Do we want to always focus X11 windows that accept it? I don't think so, Sway can be configured to not give focus to new windows by default. So maybe we can add `&& set_focus` to the `if`?

BrassyPanache: I've tested it both ways.  The fix only works if we override.

emersion (approved):
LGTM, thanks!

PR #5950, opened 2021-01-13
Set the current seat in seatop handlers

### LINKED ISSUE #5939: Switching windows places after a reload kills Sway
- **Sway Version:** 1.5.1

- **Debug Log:** http://0x0.st/-sC7.txt I commented out the Discord Canary autostart for this log, I guess that makes it easier

- **Configuration File:** https://raw.githubusercontent.com/PowerMan2206/dots/master/.config/sway/config, happens with the default config 

- **Stack Trace:** https://0x0.st/-sCd.txt is the `coredumpctl` output, and https://0x0.st/-sCR.txt of `bt full`

- **Description:** I opened 2 windows (Alacritty), reloaded Sway with Meta+Shift+C and used my mouse to switch their places (Meta+LMB), pulling one "into" the other. After this Sway dies and throws me back into the tty login screen.

rpigott:
Fixes #5939, see [comment](https://github.com/swaywm/sway/issues/5939#issuecomment-758492315).

The functions in container.c all like to reference the current seat, so we better make sure it is set before we do anything that may deal with focus. Currently, when the seatop_ commands run, the handler_context.seat can be anything, even NULL, so I've chosen to set it in all the seatop_begin* functions. I don't suppose it matters for some, like floating_resize, but I set it in all of them anyway for symmetry.


sway/tree/container.c
```diff
@@ -1614,6 +1614,181 @@ void container_raise_floating(struct sway_container *con) {
...
+	enum sway_fullscreen_mode fs1 = con1->fullscreen_mode;
+	enum sway_fullscreen_mode fs2 = con2->fullscreen_mode;
+	if (fs1) {
+		container_fullscreen_disable(con1);
+	}
+	if (fs2) {
+		container_fullscreen_disable(con2);
+	}
+
+	struct sway_seat *seat = input_manager_current_seat();
```
RPigott, line 1738: This is the only line changed in this commit. Otherwise these three functions were lifted directly from commands/swap.c

sway/tree/container.c
```diff
@@ -1423,7 +1423,7 @@ struct sway_container *container_split(struct sway_container *child,
 		}
 	}
 
-	struct sway_seat *seat = input_manager_get_default_seat();
+	struct sway_seat *seat = input_manager_current_seat();
```
RPigott, line 1426: I don't see why container_split should use the default seat while the reset use the current seat, so I've changed it.

emersion:
`config->handler_context` is mostly meant for command handlers, not for tree manipulation functions. Can we instead pass the seat via an argument to `container_swap`?

RPigott:
> `config->handler_context` is mostly meant for command handlers, not for tree manipulation functions. Can we instead pass the seat via an argument to container_swap?

Technically yes, but I think this way is better. My reasoning is as follows:

The container\* functions need a reference to a seat whenever they may mess with the focus, but none accept a seat argument. They all get a seat reference currently with `input_manager_current_seat()` with the exception of container_split, which I believe is an error, hence the 3rd commit.

The implementation of `input_manager_current_seat` is:

```c
struct sway_seat *input_manager_current_seat(void) {
	if (!seat) {
		seat = input_manager_get_default_seat();
	}
	return seat;
}
```

So they do already reference the `handler_context.seat` to use the correct seat when invoked from the command handlers.

Well the seatop actions are just like commands, and setting `handler_context.seat` in seatop_begin\* is analogous to setting it in execute_command. This change basically just treats the seatop actions like the commands that they are. If anything I think it should go further and set the handler_context node/container etc.

RPigott:
Actually I suppose it would be better to analogize each tiling dropzone with a command, rather than the whole tiling_move operation. In that case seatop_begin* is the wrong place to set `handler_context.seat`, it should be when the action ends before we start modifying the tree. Similarly the other seatops should only set it when they actually do something.

emersion:
I'm not a fan of setting global state like this. We did it because it was the less intrusive way to do it for commands, but I'd prefer not to make more stuff rely on it.

RPigott:
I generally agree, and I think there are still bugs in the PR as written. I'll need to think about this one a bit, so I've converted to a draft for now.

PR #5952, opened 2021-01-15
swaymsg: use 3 digits for fractional part of the refresh rate

RPigott:
The fractional part of the real number we want to represent never has more than 3 decimal digits, so use 3 decimal digits of precision.

e.g. 'swaymsg -t get_outputs' would show a refresh rate of 59934 mHz
as 59.933998 Hz, now correctly as 59.934 Hz.

It was previously cast as a float, probably because that's how the output_config represents it, but printf only prints doubles anyway so that seems kind of pointless.


emersion (approved):
Thanks!

PR #5955, opened 2021-01-16
Implement foreign toplevel fullscreen output hint

### LINKED ISSUE #5637: Implement foreign toplevel fullscreen output hints
This was intentionally left out of #5589. Creating this issue so I don't neglect it.

fwSmit:
Closes #5637
This feels a bit too simple, but I think this should be enough.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -651,6 +651,18 @@ static void handle_foreign_fullscreen_request(
 		}
 	}
 
+	if (event->fullscreen && event->output && event->output->data) {
+		struct sway_output *output = event->output->data;
+		struct sway_workspace *ws = output_get_active_workspace(output);
+		if (ws && !container_is_scratchpad_hidden(view->container)) {
+			if (container_is_floating(view->container)) {
+				workspace_add_floating(ws, view->container);
+			} else {
+				workspace_add_tiling(ws, view->container);
+			}
+		}
+	}
+
 	container_set_fullscreen(container,
 		event->fullscreen ? FULLSCREEN_WORKSPACE : FULLSCREEN_NONE);
 	if (event->fullscreen) {


RPigott:
Hi fwSmit,

I think the fullscreen event should already work. It is sent here:

https://github.com/swaywm/sway/blob/7cf25d3b987ff00a644bf5b6910a366f93029bbe/sway/tree/container.c#L982-L985

Or is there some case where a fullscreen handle doesn't receive the event?

#5637 is about the "output" argument of the FTM set_fullscreen request which should hint to the compositor on which output the client should be fullsceened. At the moment sway just ignores it. See [#5589 (comment)](https://github.com/swaywm/sway/pull/5589#issuecomment-667104171) if you still want to tackle it.

fwSmit:
> EDIT: It occurs to me that [a015730](https://github.com/swaywm/sway/commit/a01573016a0a18476afbc2fa7c467406338a7672) doesn't appear in any release version of sway. Could that be the confusion?

You're right, it was a combination of that and me thinking that the toplevel handler and toplevel manager versions are completely independent (why is the handler version provided each time a client gets one. It seems to be the same version as the manager?).

> Or is there some case where a fullscreen handle doesn't receive the event?

Nope, it seems to be working great.

> #5637 is about the "output" argument of the FTM set_fullscreen request which should hint to the compositor on which output the client should be fullscreened. At the moment sway just ignores it. See [#5589 (comment)](https://github.com/swaywm/sway/pull/5589#issuecomment-667104171) if you still want to tackle it.

Thanks for the explanation, I'll give it a go!

RPigott:
> why is the handler version provided each time a client gets one. It seems to be the same version as the manager?

All Wayland interfaces are versioned. The global is incremented whenever one of it's interfaces changes. Some info here: https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-Versioning

fwSmit:
> All Wayland interfaces are versioned. The global is incremented whenever one of it's interfaces changes. Some info here: https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-Versioning

So you're always guaranteed to get the same version of toplevel handler as toplevel manger from the compositor? Or can the compositor decide to give you a V3 manager and a V2 handler?

Anyways, I've copy pasta'd the linked code. I haven't been able to test it yet, because WLR_WL_OUTPUTS doesn't seem to do much.
I have a second display, but when I just tested it, I got only a black screen. This is a regression in sway/wlroots that may have been fixed in master already. I'll rebase and test later today.

fwSmit:
> No, the versions will always match.

Okay, good to know.

> > WLR_WL_OUTPUTS doesn't seem to do much
> 
> Works for me...

What should it do? The documentation doesn't tell me much.

emersion:
When running with the Wayland backend (nested in another Wayland compositor), it should open two windows (ie. two virtual outputs) instead of a single one.

fwSmit:
> When running with the Wayland backend (nested in another Wayland compositor), it should open two windows (ie. two virtual outputs) instead of a single one.

Oh I assumed it would work from tty as well. I now got it working and tested this PR. Found no problems. See https://github.com/swaywm/wlroots/pull/2681 for test code.

RPigott (approved):
Sorry for the delay! Patch LGTM, thanks!

PR #5957, opened 2021-01-20
Clip title bars to container (screenshots inside)

dimitris-triantafyllidis:
Title bars can be rendered outside the container in stacked containers
with too many children or in deep splits where a lot of title bars are rendered.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -439,6 +439,19 @@ static void render_titlebar(struct sway_output *output,
 	int titlebar_v_padding = config->titlebar_v_padding;
 	enum alignment title_align = config->title_align;
 
+	struct wlr_box clip_box = {
+		.x = floor(con->current.x - output_x),
+		.y = 0,
+		.width = con->current.width,
+		.height = con->current.y - output_y + con->current.height,
+	};
+	scale_box(&clip_box, output_scale);
+
+	pixman_region32_t clipped_output_damage;
+	pixman_region32_init(&clipped_output_damage);
+	pixman_region32_intersect_rect(&clipped_output_damage, output_damage,
+		clip_box.x, clip_box.y, clip_box.width, clip_box.height);
+
 	// Single pixel bar above title
 	memcpy(&color, colors->border, sizeof(float) * 4);
 	premultiply_alpha(color, con->alpha);
@@ -447,7 +460,7 @@ static void render_titlebar(struct sway_output *output,
 	box.width = width;
 	box.height = titlebar_border_thickness;
 	scale_box(&box, output_scale);
-	render_rect(output, output_damage, &box, color);
+	render_rect(output, &clipped_output_damage, &box, color);
 
 	// Single pixel bar below title
 	box.x = x;
@@ -455,7 +468,7 @@ static void render_titlebar(struct sway_output *output,
 	box.width = width;
 	box.height = titlebar_border_thickness;
 	scale_box(&box, output_scale);
-	render_rect(output, output_damage, &box, color);
+	render_rect(output, &clipped_output_damage, &box, color);
 
 	// Single pixel left edge
 	box.x = x;
@@ -463,7 +476,7 @@ static void render_titlebar(struct sway_output *output,
 	box.width = titlebar_border_thickness;
 	box.height = container_titlebar_height() - titlebar_border_thickness * 2;
 	scale_box(&box, output_scale);
-	render_rect(output, output_damage, &box, color);
+	render_rect(output, &clipped_output_damage, &box, color);
 
 	// Single pixel right edge
 	box.x = x + width - titlebar_border_thickness;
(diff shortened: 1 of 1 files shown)


sway/desktop/render.c
```diff
@@ -423,31 +461,31 @@ static void render_titlebar(struct sway_output *output,
 	box.width = width;
 	box.height = titlebar_border_thickness;
 	scale_box(&box, output_scale);
-	render_rect(output, output_damage, &box, color);
+	render_rect(output, &clipped_output_damage, &box, color);
```
kennylevinsen, line 464: We should no't need to clip the titlebar. Instead, just use the right dimensions when drawing it.

sway/desktop/render.c
```diff
@@ -330,10 +330,35 @@ static void render_saved_view(struct sway_view *view,
...
+		}
+
+		pixman_region32_fini(&clipped_damage);
+
+	} else {
+		if (!wl_list_empty(&view->saved_buffers)) {
+			render_saved_view(view, output, damage, view->container->alpha);
+		} else if (view->surface) {
+			render_view_toplevels(view, output, damage, view->container->alpha);
+		}
```
kennylevinsen, line 361: This basically does the same as #5451, but on the other side of the equation.

`render_texture` takes a destination box, which it turns into a pixman rect (using union_rect), and then intersects it with the output damage. #5451 approach limits the destination box size (which is just 2 fmin calls), where this approach limits the output damage (which is an arbitrarily large number of random rectangles).

I think the approach in #5451 lacked handling of output scale, but I wonder if that's all it's missing.

sway/desktop/render.c
dimitris-triantafyllidis: I used this approach because it contains almost zero extra logic and code, and is very localized. If an error occurs it can be tracked somewhere in these few lines of code. Would there be any benefit in going all over this function and modifying code for this corner case?

sway/desktop/render.c
dimitris-triantafyllidis: Same as above, I did it this way because of simplicity and locality. The intersection call is not free but is the overhead even measurable? If it indeed is, maybe the intersection calls could be moved to an outer function and save us some math operations.

dimitris-triantafyllidis:
I would also like to add a further note regarding this. There are some rounding issues and an extreme scenario where I can still cause damage tracking artifacts and I would like to fix these. Yeah they are not critical by any means but why not? Regarding the rounding issues, I partially address them in #5711. It that PR gets merged, I will also have to tweak the code in this PR, and as it is structured now it will be a trivial change. In any case, thanks for taking the time to look at this and feel free to not merge this if you think you want to try another approach. This also needs to be tested with multiple outputs by the way, I may do it today If I have time.

sway/desktop/render.c
kennylevinsen: render_titlebar takes a single "width" argument. It's a bug if the caller sets the "width" argument so that we'd render outside the bounds we've set for the container. We shouldn't add extra logic and code to work around such bugs.

Did you have an observed case where this happens?

(I don't personally think we need to consider/add code for the scenario where the titlebar is *taller* than the container)

dimitris-triantafyllidis: No, I have not observed any issues with the width and I agree that in such a case, the width should be correctly specified by the caller. The only issue that the title bar clipping part of this PR is aiming to address is in stacked containers.

sway/desktop/render.c
kennylevinsen: > The intersection call is not free but is the overhead even measurable?

All execution time is measurable. It's always a good idea to avoid work in code that is called many times for every frame. That is not to suggest excessive premature optimization, but if it is trivial to pick a *much* cheaper option, then it would make sense to do so.

I am not extremely opposed to your current approach, but I just like the idea of just giving a smaller wlr_box to render_texture to copy instead of adding additional intersects on the output damage.

> If it indeed is, maybe the intersection calls could be moved to an outer function and save us some math operations. In any case, some math has to be done to also account for scale and transformations. GTK AFAIK only uses integer scaling factors, other GUI kits may use floats, so this might be an extra thing to account for, to get the roundings right etc.

This is all accounted for in `render_texture`. Also, wayland only exposes integer scaling factors for buffers.

dimitris-triantafyllidis: > All execution time is measurable. It's always a good idea to avoid work in code that is called many times for every frame. That is not to suggest excessive premature optimization, but if it is trivial to pick a much cheaper option, then it would make sense to do so.

Agreed. My thinking was that if it works well and no alternative is going to be merged soon, it could be good for now and I would come back and improve it after I fixed some other things I wanted fixed. BTW I just saw that you just updated your PR, going to check it out!

dimitris-triantafyllidis:
Removed the view clipping part of this PR since #5451 got merged. Posting a before / after to illustrate the title bars issue. The damage artifacts when moving the overflowed container are not visible in the screenshots.

before
after

PR #5960, opened 2021-01-20
focus: beyond fullscreen when focused explicitly

lbonn:
When issuing a focus command on a specific container, users expect to proceed it even if is hidden by a fullscreen window.

This matches the behavior of i3.

Reported by @markstos: https://github.com/lbonn/i3-focus-last/issues/5

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -377,6 +377,13 @@ struct cmd_results *cmd_focus(int argc, char **argv) {
 		if (container_is_scratchpad_hidden_or_child(container)) {
 			root_scratchpad_show(container);
 		}
+		// if we are switching to a container under a fullscreen window, we first
+		// need to exit fullscreen so that the newly focused container becomes visible
+		struct sway_container *obstructing = container_obstructing_fullscreen_container(container);
+		if (obstructing) {
+			container_fullscreen_disable(obstructing);
+			arrange_root();
+		}
 		seat_set_focus_container(seat, container);
 		seat_consider_warp_to_focus(seat);
 		container_raise_floating(container);
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1090,25 +1090,14 @@ void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
 	struct sway_container *container = node->type == N_CONTAINER ?
 		node->sway_container : NULL;
 
-	// Deny setting focus to a view which is hidden by a fullscreen container
-	if (new_workspace && new_workspace->fullscreen && container &&
-			!container_is_fullscreen_or_child(container)) {
-		// Unless it's a transient container
-		if (!container_is_transient_for(container, new_workspace->fullscreen)) {
-			return;
-		}
+	// Deny setting focus to a view which is hidden by a fullscreen container or global
+	if (container && container_obstructing_fullscreen_container(container)) {
+		return;
 	}
+
 	// Deny setting focus to a workspace node when using fullscreen global
 	if (root->fullscreen_global && !container && new_workspace) {
 		return;
-	}
-	// Deny setting focus to a view which is hidden by a fullscreen global
-	if (root->fullscreen_global && container != root->fullscreen_global &&
-				!container_has_ancestor(container, root->fullscreen_global)) {
-		// Unless it's a transient container
-		if (!container_is_transient_for(container, root->fullscreen_global)) {
-			return;
-		}
 	}
 
 	struct sway_output *new_output =
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -418,6 +418,28 @@ void container_for_each_child(struct sway_container *container,
 	}
 }
 
+struct sway_container *container_obstructing_fullscreen_container(struct sway_container *container)
+{
+	struct sway_workspace *workspace = container->pending.workspace;
+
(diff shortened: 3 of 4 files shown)


sway/input/seat.c
```diff
@@ -1058,11 +1058,13 @@ void seat_set_raw_focus(struct sway_seat *seat, struct sway_node *node) {
 	}
 }
 
-void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
+static void _seat_set_focus(struct sway_seat *seat, struct sway_node *node, bool force) {
+	// If force is true, first unset fullscreen on overlapping node
```
Xyene, line 1062: This comment probably belongs in the header file, distinguishing `seat_set_focus` from `seat_set_focus_force`.

sway/input/seat.c
```diff
@@ -1058,11 +1058,13 @@ void seat_set_raw_focus(struct sway_seat *seat, struct sway_node *node) {
 	}
 }
 
-void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
+static void _seat_set_focus(struct sway_seat *seat, struct sway_node *node, bool force) {
```
Xyene, line 1061: Nit: we usually shy away from `_function` naming internal functions.

sway/input/seat.c
```diff
@@ -1091,23 +1093,33 @@ void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
 		node->sway_container : NULL;
 
 	// Deny setting focus to a view which is hidden by a fullscreen container
+	// except if forced
```
Xyene, line 1096: Design question: could we refactor out this code into a new `struct sway_container *container_obstructing_fullscreen_container(struct sway_container *)`?

Then `focus.c` could check if `container_obstructing_fullscreen_container(...) != NULL` and `container_fullscreen_disable(...)` that, removing the need to plumb `force` through this code.

Xyene (commented):
Sorry for the late review; I have some general questions regarding this approach, but functionally it looks good.

sway/input/seat.c
lbonn: This sounds like a much better design, thank you.
I've made the changes accordingly.

lbonn:
After @Xyene's review, the method has changed a bit.

However the behavior is different than last time I've tried and fullscreen windows unfocused this way will still appear maximized if focused again. I'll try to see if it comes from my patch or recent changes since the PR was opened.

lbonn:
Actually it also happens on the original version... Any idea what might be going on?

Xyene:
Hmm, I'm not sure, but you might be missing a call to `arrange_root` after the disable?

sway/input/seat.c
Xyene: :+1: Ideally, we'd use this code within `seat_set_focus` too, so we don't have the same block duplicated twice.

sway/commands/focus.c
```diff
@@ -377,6 +377,10 @@ struct cmd_results *cmd_focus(int argc, char **argv) {
 		if (container_is_scratchpad_hidden_or_child(container)) {
 			root_scratchpad_show(container);
 		}
+		struct sway_container *obstructing = container_obstructing_fullscreen_container(container);
```
Xyene, line 380: Probably best to add a comment here for why `seat_set_focus_container` is insufficient, since this is the only special case.

sway/input/seat.c
lbonn: Which block are you talking about? The one added in focus.c?

Initially, that's why I added a `seat_set_focus_force` function as when I reviewed the code back in January it seemed like this was the only location we'd actually need to do that. There are a lot of calls to `seat_set_focus*` and adding the same logic there has a big chance of introducing regressions IMO.

lbonn:
> Hmm, I'm not sure, but you might be missing a call to `arrange_root` after the disable?

That did the trick, thank you! Just pushed a new version with `arrange_root()` and an explanatory comment.

sway/input/seat.c
Xyene: I meant that these two blocks are basically identical in functionality:

https://github.com/swaywm/sway/blob/63564ee11dd168e35b7e1ac7f3068fff10372cbd/sway/tree/container.c#L420-L441

https://github.com/swaywm/sway/blob/63564ee11dd168e35b7e1ac7f3068fff10372cbd/sway/input/seat.c#L1093-L1112

The one from `seat.c` can be made to call `container_obstructing_fullscreen_container` and avoid duplicating its logic.

lbonn: Ah gotcha haha :smiley: !

sway/tree/container.c
```diff
@@ -418,6 +418,32 @@ void container_for_each_child(struct sway_container *container,
 	}
 }
 
+struct sway_container *container_obstructing_fullscreen_container(struct sway_container *container)
+{
+	if (!container) {
```
Xyene, line 423: Most of the functions here don't do nullchecks internally, let's just move this check to the caller for consistency.

Xyene (commented):
Otherwise LGTM.

sway/tree/container.c
lbonn: Updated

markstos:
@emersion This bug fix to match i3 behavior appears ready to merge pending final review by @Xyene and could be a good candidate for the 1.6 release.

Xyene (approved):
Thanks!

PR #5972, opened 2021-01-25
desktop/output: Disable head if mode is NULL

kennylevinsen:
wlr_output_configuration_head_v1_create normally fills out the head
"enabled" field to match the wlr_output state. We overwrite this to also
set the head as enabled if it is only turned off with DPMS.

However, in some cases we may not have a mode for this display, in which
case setting it as enabled will lead to a segfault later on. Therefore,
enabled conditional on the presence of a mode.

Fixes: https://github.com/swaywm/wlroots/issues/2199


emersion (approved):
Hm, I guess this would happen if an output is plugged in but is configured to be DPMS off.

Thanks!

PR #5973, opened 2021-01-26
xdg-foreign: add v1 and v2 implementations

alebastr:
Updated original xdg-foreign commit with changes from swaywm/wlroots#2487.

Tested with [portal-test](https://github.com/matthiasclasen/portal-test) flatpak.

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -9,6 +9,7 @@
 #include <wlr/backend/multi.h>
 #include <wlr/backend/noop.h>
 #include <wlr/backend/session.h>
+#include <wlr/config.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_compositor.h>
 #include <wlr/types/wlr_data_control_v1.h>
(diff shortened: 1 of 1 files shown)


include/sway/server.h
```diff
@@ -84,6 +88,9 @@ struct sway_server {
 	struct wlr_input_method_manager_v2 *input_method;
 	struct wlr_text_input_manager_v3 *text_input;
 	struct wlr_foreign_toplevel_manager_v1 *foreign_toplevel_manager;
+#if WLR_HAS_XDG_FOREIGN
+	struct wlr_xdg_foreign_registry *foreign_registry;
```
emersion, line 92: Do we need to store this in the server struct? If not, maybe we can just remove this field.

alebastr: Not really, there's no reason to store it here.
Server struct already contains several objects that aren't used anywhere beyond server.c and could be replaced by variables in `server_init` so I assumed that's a part of code style.

emersion: Yeah, it's mostly cargo culting I think. I'd merge a PR cleaning that up.

emersion (approved):
Thanks!

PR #5980, opened 2021-01-29
Fix incorrect damage being applied on popups

tadeokondrak:
This is because popup iteration is based on `container->surface_x` and `container->surface_y`, which are only set after the new location is damaged.

To reproduce:

- Open a floating window and a popup that hangs over the bottom or right
- Move the window in the direction of the popup overhang
- The previous position of the popup is damaged, not the new one


Xyene (approved):
LGTM, thanks!

Issue #5857, opened 2020-12-04
Infinite transaction loop


RPigott:
In quick resize operations, some clients can get stuck in an infinite transaction loop where sway will send constant configure events. This bug was also introduced by #5812. Programming is hard.

Steps to reproduce:

1. Float a slow-reacting client, e.g. slow-ack
2. `swaymsg '[con_mark=slow]' resize shrink width 40px && swaymsg '[con_mark=slow]' resize grow width 20px`
3. Observe it transform back-and-forth between sizes.

Note the use of && instead of comma separator in swaymsg. Alternatively, it's pretty easy to accidentally trigger with an interactive resize. I can't reproduce this with termite, but I can with the imv image viewer.

I noticed that when the client is in this loop, it somehow always has the wrong configure serial (an old one) to offer transaction_notify_view_ready_by_serial, so that function never does anything.

kennylevinsen:
I suppose one could make an argument for never, or only once, sending a configure with size to floating windows. Just draw what size they give us and call it a day.

emersion:
Yeah, that sounds reasonable. Unless the user resizes the window, in case we really need to send a configure.

Figuring out which case it is (extraneous configure, or manual configure) may be an issue.

tadeokondrak:
Not sure if this is the same issue or not, but #5812 causes weston-terminal floating windows to forever grow in height on their own for me.

kennylevinsen:
> Yeah, that sounds reasonable. Unless the user resizes the window, in case we really need to send a configure.

Only if it's a server-side resize, right?

> Figuring out which case it is (extraneous configure, or manual configure) may be an issue.

I was thinking of setting a sticky bit on the container state when we are doing server-initiated resizing, which would cause the transaction logic to send a configure. The bit would be cleared in the pending state when setting up the transaction.

There's only two locations that would be exempt from setting this bit (xwayland and xdg_shell commit), and only for floating windows.

PR #5812, opened 2020-11-13
xdg_shell: allow views to change geometry anytime

RPigott:
Fixes #5750.

Currently, when sway sends a configure with some geometry and the client responds with a different geometry in a commit that acks that configure, sway ignores the new size. Sway applies the surface geometry it had requested to the container, not what was actually committed, in the following transaction.

This change allows any client commit to change its surface geometry, even if it is a response to a configure event.

PR #5981, opened 2021-01-30
Transactional resize fixes

kennylevinsen:
Closes: https://github.com/swaywm/sway/issues/5848

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -317,6 +317,7 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 void view_unmap(struct sway_view *view);
 
 void view_update_size(struct sway_view *view, int width, int height);
+void view_center_surface(struct sway_view *view);
 
 void view_child_init(struct sway_view_child *child,
 	const struct sway_view_child_impl *impl, struct sway_view *view,
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -316,7 +316,7 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 
 void view_unmap(struct sway_view *view);
 
-void view_update_size(struct sway_view *view, int width, int height);
+void view_update_size(struct sway_view *view);
 void view_center_surface(struct sway_view *view);
 
 void view_child_init(struct sway_view_child *child,
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -295,17 +295,20 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 	if (new_size) {
 		// The view has unexpectedly sent a new size
 		desktop_damage_view(view);
-		view_update_size(view, new_geo.width, new_geo.height);
 		memcpy(&view->geometry, &new_geo, sizeof(struct wlr_box));
+		if (container_is_floating(view->container)) {
+			view_update_size(view, new_geo.width, new_geo.height);
+			transaction_commit_dirty();
+			transaction_notify_view_ready_immediately(view);
+		} else {
+			view_center_surface(view);
+		}
 		desktop_damage_view(view);
-		transaction_commit_dirty();
 	}
 
 	if (view->container->node.instruction) {
 		transaction_notify_view_ready_by_serial(view,
 				xdg_surface->configure_serial);
-	} else if (new_size) {
-		transaction_notify_view_ready_immediately(view);
 	}
 
 	view_damage_from(view);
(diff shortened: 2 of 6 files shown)


kennylevinsen:
cc @RPigott @emersion

Xyene:
Thanks for working on this!

It looks like something's not quite right with titlebars during interactive resizes, though (excuse the vertical monitor):


https://user-images.githubusercontent.com/1403503/106537327-15fbea80-64c8-11eb-81b7-e4103ccd2e84.mp4

kennylevinsen:
@Xyene Hmm, something seems to have been broken in the refactor when I tried to re-enable the centering logic. I'll take a look. Thanks for the test.

kennylevinsen:
Hmm.

I think we're dealing with a bad interaction with the time of sampling and saved buffers.

The initial implementation added an extra geometry field to `struct sway_view`. Every surface commit would update `next_geometry`, with an executing transaction copying to `current_geometry` - *not* state from when the transaction was queued. This interacted poorly with surface centering logic, but otherwise *appeared* to behave correctly.

The current approach uses `struct sway_container_state`, and all the usual logic there. This means that the next geometry is copied when the transaction is queued, and moved to current when it executes. This causes occasional misrendering.

I suspect this is an interaction between when the geometry is sampled vs. e.g. when the buffer is saved. It's possible that this can be fixed by simply sampling at the time of queuing rather than when the transaction executes, not sure. The transaction system is a bit cumbersome to deal with...

kennylevinsen:
I have reworked the approach.

It turns out most of the issues were associated with the view centering logic, especially when combined with saved buffers, and I had just been lead down a wrong path with respect to double-buffering view geometries.

I'll try to daily drive this a bit, but @RPigott and @Xyene, if you have time, please do test and see how things work out. Especially if you have any weird clients, like the GNOME polkit dialog.

kennylevinsen:
As pointed out in IRC, horizontal resizes are now fine, but vertical resizes when the top of a window is moved still has some slack left proportional to the resize. Funnily enough, it does not happen when resizing from the bottom.

include/sway/tree/view.h
```diff
@@ -317,6 +317,7 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 void view_unmap(struct sway_view *view);
 
 void view_update_size(struct sway_view *view, int width, int height);
+void view_center(struct sway_view *view);
```
emersion, line 320: Maybe rename to `view_center_surface`? `view_center` sounds like the function would center the view on screen.

kennylevinsen: Agreed

kennylevinsen:
Found the issue - view_center, now view_center_surface, was erroneously referring to `con->content_y` (the pending container state), rather than `con->current.content_y` (the current container state) as intended. This was throwing off surface_y, which meant that resizes that moved a surface vertically would still have glitches.

sway/desktop/xdg_shell.c
```diff
@@ -287,10 +287,8 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 
 	struct wlr_box new_geo;
 	wlr_xdg_surface_get_geometry(xdg_surface, &new_geo);
-	bool new_size = new_geo.width != view->geometry.width ||
-			new_geo.height != view->geometry.height ||
-			new_geo.x != view->geometry.x ||
-			new_geo.y != view->geometry.y;
+	bool new_size = memcmp(&new_geo, &view->geometry,
```
emersion, line 290: Hm, I'm concerned about padding. The struct _probably_ won't have padding, but `memcmp` would make it extremely difficult to debug if it ever has. Maybe introduce a helper for this?

kennylevinsen: Ugh, you're right, undid the change. :|

emersion (approved):
LGTM. @RPigott, @Xyene, do you have anything to add?

RPigott:
Hey, sorry I'm late.

The approach makes sense. I've just tried to break it for a while with a variety of weird clients but it has held up :+1:. At the very least, the previously known issues addressed and/or caused by #5812 seem to be working. I only noticed that #5857 is still here, but I see you've already made another PR for that. I'll keep running this, but I think it could be merged.

Thanks very much for helping fix this.

Xyene:
Thanks again for working on this! The vertical resizes are much better than before, but there seems to be at least one more case where the "sliver of the desktop" appears when resizing containers with titlebars.

Consider this layout: `V[A T[H[B* C]]]`. Resizing vertically causes a gap between B/C's titlebars and the titlebar of the `T` to appear. No gap appears between the titlebar of B/C and their content, though, so that's fixed :)

emersion:
Alight, merging now then, thanks!

kennylevinsen:
@Xyene I can't seem to reproduce the last mentioned problem with a `V[foot T[H[slow-ack-configure foot]]]`. Could you create a new issue for it with more details?

PR #5982, opened 2021-01-30
desktop/layer_shell: Fix allocation type mismatch

mstoeckl:
Since `sizeof(struct sway_layer_surface) > sizeof(struct sway_layer_subsurface)`, the allocation size mismatch fixed by this PR is currently not a problem.


Xyene (approved):
Thanks!

PR #5983, opened 2021-01-30
Check the output state for recorded workspaces

kupospelov:
Steps to reproduce:
1. Run a process that takes some time to open a view (e.g. a bash script runs on workspace 2 and sleeps for a few seconds before executing something else).
2. Disable the output on which the process started.

In this case Sway re-creates the workspace on the disabled output. If you go to workspace 2 after that, there appears one more workspace 2 on the enabled output.


emersion (approved):
LGTM, thanks!

PR #5993, opened 2021-02-04
Align ordering of core node properties with i3

jpalus:
Try to better mimic JSON node structure produced by i3 which might be
relied on by already existing tools. In particular having "type" right
after "id" is quite handy for streaming high-performance JSON parsers
such as simdjson (which are handy for maintaining responsiveness on
resource constrained systems).

refer https://github.com/i3/i3/blob/ab2a22a78b25ad12fed2c177a34c44950795cf33/src/ipc.c#L338

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -23,6 +23,20 @@
 static const int i3_output_id = INT32_MAX;
 static const int i3_scratch_id = INT32_MAX - 1;
 
+static const char *ipc_json_node_type_description(enum sway_node_type node_type) {
+	switch (node_type) {
+	case N_ROOT:
+		return "root";
+	case N_OUTPUT:
+		return "output";
+	case N_WORKSPACE:
+		return "workspace";
+	case N_CONTAINER:
+		return "con";
+	}
+	return "none";
+}
+
 static const char *ipc_json_layout_description(enum sway_container_layout l) {
 	switch (l) {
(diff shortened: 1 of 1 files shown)


sway/ipc-json.c
```diff
@@ -23,6 +23,20 @@
 static const int i3_output_id = INT32_MAX;
 static const int i3_scratch_id = INT32_MAX - 1;
 
+static const char *ipc_json_node_type_description(enum sway_node_type l) {
```
Xyene, line 26: Nit: let's use something more descriptive than `l` here (`l` is used for `sway_container_layout` elsewhere in this file, which is bad as-is; making it represent a `sway_node_type` here makes things more confusing).

Xyene (changes requested):
Thanks for working on this. I'm generally happy to merge this since it cleans up this code a bit, but note that we don't -- and can't -- make any guarantees about node ordering with respect to i3. It's also possible that future changes will make the ordering differ again (especially as new i3 properties are introduced).

sway/ipc-json.c
jpalus: Renamed to `node_type`.

jpalus:
Thanks for review! As for lack of guarantees that's perfectly understandable. I'll take my chances though :).

Xyene (approved):
Thanks!

PR #5994, opened 2021-02-04
Add static+const qualifiers

mstoeckl:
There are three groups of variables which this PR makes static const.

* Lists of command handlers
* The Wayland request listeners
* The option lists passed to getopt_long

I'm fairly certain that these are all the variables which should be static const; I checked with ``objdump -t `which sway` | grep "\.data" `` to find initialized global (or function static) variables that were not linked as relocatable read-only, and did not find any others which could be fixed.


emersion (approved):
LGTM, thanks!

PR #6002, opened 2021-02-06
output: simplify layer surface iteration

vyivel:
Note that now data.{width,height} are set to the layer surface's size, not to an xdg_surface's. Not sure if it's important, but it seems to work correctly anyway.


emersion (approved):
Thanks!

Issue #5991, opened 2021-02-02
sway tools such as swaybar and swaynag do not respond to touch input


sicelo:
- **Sway Version:**
  - `1.5-40bdd0ee (Jan  6 2021, branch 'master')`

- **Debug Log:**
  - https://gist.github.com/sicelo/46e87812327e38a9334b04f3b1c41fb6

- **Description:**
  - `swaynag` and `swaybar` do not react to touch input on my phone, e.g. I cannot dismiss a `swaynag` prompt.

- **Additional info**
  - ``swaymsg -t get_inputs` output for this touchscreen (it's a single-touch, resistive touch panel):
  ```Input device: TSC2005 touchscreen
  Type: Touch
  Identifier: 0:2005:TSC2005_touchscreen 
  Product ID: 2005
  Vendor ID: 0
  Libinput Send Events: enabled
```

PR #6003, opened 2021-02-07
swaynag: Use position from wl_pointer.enter

kennylevinsen:
Only wl_pointer.motion was used to update pointer position, which would
cause issues if the pointer was not moved prior to wl_pointer.button.

This also fixes touch input through wl_pointer emulation, which fires
wl_pointer.button immediately after wl_pointer.enter.

Closes: https://github.com/swaywm/sway/issues/5991


Xyene (approved):
LGTM, thanks! This has got to be one of the most common oversights Wayland clients make... I've seen variations of it in five different projects by now :')

PR #6004, opened 2021-02-08
text_input: Add support for focusing layer-shell surfaces

tadeokondrak:
Without this, text-input focus stays on the last focused non-layer-shell surface.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1234,6 +1234,7 @@ void seat_set_focus_surface(struct sway_seat *seat,
 		wlr_seat_keyboard_notify_clear_focus(seat->wlr_seat);
 	}
 
+	sway_input_method_relay_set_focus(&seat->im_relay, surface);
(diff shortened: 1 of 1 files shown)


sway/input/seat.c
```diff
@@ -1230,8 +1230,10 @@ void seat_set_focus_surface(struct sway_seat *seat,
 
 	if (surface) {
 		seat_keyboard_notify_enter(seat, surface);
+		sway_input_method_relay_set_focus(&seat->im_relay, surface);
```
Xyene, line 1233: Nit: for symmetry, this line can go above `seat_tablet_pads_notify_enter`, which also has internal logic for handling the `surface = NULL` case. Then the second diff line below becomes unnecessary.

Xyene (changes requested):
LGTM, thanks!

Xyene (approved):
Thanks!

PR #6005, opened 2021-02-08
text_input: Ignore text_input disable from unfocused windows

tadeokondrak:
Before this commit, there would be cases where focus changes from one
window to another, the new window activates text_input, then the old
window sends a deactivate request, making text_input unfocused
completely.


Xyene (approved):
LGTM, thanks!

PR #6007, opened 2021-02-08
text_input: Only send surrounding_text and content_type if supported

tadeokondrak:
Allows the input method to detect if these are not supported.
~~Depends on https://github.com/swaywm/wlroots/pull/2735~~


Xyene (approved):
Thanks!

PR #6009, opened 2021-02-08
Avoid configure events on client-initiated resize

kennylevinsen:
Avoid transactions if a resize is client-initiated.

Closes: https://github.com/swaywm/sway/issues/5857

### DIFF
--- a/include/sway/desktop/transaction.h
+++ b/include/sway/desktop/transaction.h
@@ -51,10 +51,4 @@ void transaction_notify_view_ready_by_serial(struct sway_view *view,
 void transaction_notify_view_ready_by_geometry(struct sway_view *view,
 		double x, double y, int width, int height);
 
-/**
- * Unconditionally notify the transaction system that a view is ready for the
- * new layout.
- */
-void transaction_notify_view_ready_immediately(struct sway_view *view);
-
 #endif
--- a/include/sway/desktop/transaction.h
+++ b/include/sway/desktop/transaction.h
@@ -28,6 +28,12 @@ struct sway_view;
  */
 void transaction_commit_dirty(void);
 
+/*
+ * Same as transaction_commit_dirty, but signalling that this is a
+ * client-initiated change has already taken effect.
+ */
+void transaction_commit_dirty_client(void);
+
 /**
  * Notify the transaction system that a view is ready for the new layout.
  *
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -35,6 +35,7 @@ struct sway_transaction_instruction {
 		struct sway_container_state container_state;
 	};
 	uint32_t serial;
+	bool server_request;
 	bool waiting;
 };
 
@@ -165,7 +166,7 @@ static void copy_container_state(struct sway_container *container,
 }
 
 static void transaction_add_node(struct sway_transaction *transaction,
-		struct sway_node *node) {
+		struct sway_node *node, bool server_request) {
 	struct sway_transaction_instruction *instruction = NULL;
 
 	// Check if we have an instruction for this node already, in which case we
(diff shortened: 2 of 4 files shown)


kennylevinsen:
cc @RPigott

emersion:
> shells: Avoid txns for client-initiated resize 

I'm not sure how I feel about this commit. Can you expand on the motivation?

My current understanding is that the atomic transactions should be a no-op if no configure is needed, and that side-stepping transactions will create a new separate code-path.

kennylevinsen:
We have two paths:

1. Make it possible to communicate to the transaction subsystem that a node change was initiated by the client or by the server.
2. Redefine the purpose of the transaction subsystem to only apply to server-side initiated transactions.

I initially aimed to implement the first path, but found it to be rather clumsy. It requires more state on the transaction or nodes to indicate the initiator of the transaction or node change. This also implies handling precedence correctly when compressing transactions touching the same nodes as we currently do.

Client-initiated resizes play poorly with the transaction subsystem regardless, as the buffers for the new geometry has already been received. The buffer save logic will try to save and render a buffer for the current geometry, even though the buffer is for the new geometry. As buffer save only happens once we reach the transaction in the queue, it only works for windows which don't yet know of the new size.

The only "good" use of the transaction system for client-side resizes is when it turns out to be a no-op, so I thought it made more sense to reduce its scope and only use it for server-side resizes. Granted, `view_update_size` isn't super pretty in this approach due to copying state from con to con->current.

kennylevinsen:
Found an issue, so converting back to draft to block merge for now.

I'm considering add a client to wleird that runs a constant resizing loop to make it easier to see behavior of client-initiated buffer size changes.

kennylevinsen:
False alarm: It was not a regression from master, but pre-existing weird-ish behavior with respect to the resizor test from wleird.

https://github.com/emersion/wleird/pull/18 adds a client that stresses client-side resizing, which behaves nicely with this PR.

Xyene:
I'm not sure whether this is caused strictly by this PR (I have all your open-for-review ones patched in), but I'm sometimes seeing this when moving a window from the tiling layer to the floating:

https://user-images.githubusercontent.com/1403503/107997409-afd79300-6fb0-11eb-806d-d640bdecdd5e.mp4

kennylevinsen:
I got an easy reproduction of this with txn-wait.

It appears that the issue is that the window is behaving as floating before the last tiling transaction it is a member of completes. I imagine it's just the usual problem with pending state being used somewhere, rather than the intended current state. I'm not entirely sure if that's the only issue based on the glitches seen, but it would certainly cause weird rendering as the transaction goes through.

kennylevinsen:
Pretty sure the problem is container_is_floating, which uses pending rather than current state. A bunch of places might need to check pending though.

After https://github.com/swaywm/sway/pull/6019 we could change it to be `container_state_is_floating` if we want to make it universal rather than have a current/pending variant or a boolean arg.

kennylevinsen:
I have pushed an approach that uses the transaction system. It's not super pretty and a bit pointless, but it allows us to synchronize with the current state updates easily. That should fix the issue @xyene shows in the video clip.

The reason for this change is that the previous approach required a lot of cleanup around pending/current use which ended up being too much work for now.

However, as previously mentioned, going through the transaction system leads to at least one bug: Saved buffers are incorrect for client-initiated resize, as they're from *after* the change rather than before. Thus, if a floating window ends up waiting in the transaction system after a client-initiated resize, we *will* show imperfect frames with incorrect geometry.

There are two ways to fix this: Change buffer save logic to save the *old* buffer immediately on surface commit instead of letting new buffer content become visible, or go back to bypassing the transaction system entirely on client-initiated resizes if we get pending/current state a bit more under control.

emersion (approved):
LGTM, thanks!

Issue #6012, opened 2021-02-10
Parent container headers don't precisely track position of children


Xyene:
Continuing from https://github.com/swaywm/sway/pull/5981#issuecomment-775789768.

Video of issue:

https://user-images.githubusercontent.com/1403503/107462992-4af0e880-6b2b-11eb-8de5-58c9127124b2.mp4

There are some frames in the first half that look like this:

![image](https://user-images.githubusercontent.com/1403503/107463044-6825b700-6b2b-11eb-90ad-22c8e3e04756.png)

The issue only appears to occur if the interactive resize is triggered from the left half of the child container used in the video. The issue also occurs if the right half of the right child is used, but not the right. So, I suspect the issue only triggers if the child width changes as well as the height &mdash; if only the height changes, things work fine. (Perhaps the updates land in different transactions?)

/cc @kennylevinsen

kennylevinsen:
It doesn't reproduce under txn-wait, or with slow-ack-configure from weird? It'll be a bit annoying to debug if it's only sporadic single frame glitches. :/

Do you mean it reproduces using *left* half of the right view?

kennylevinsen:
Reproducing with `V[foot T[weston-simple-egl H[foot foot]]]`, I see at times in excess of 30 transactions queued, with the current transaction compression unable to take care of all of it, so active state is a few txn iterations behind.

However, it would seem that rendering the titlebars like that imply that `con->current.y` of the child container is very different from that of the parent, so it seems like a weird transaction. Or maybe the arrange logic occasionally messes up.

kennylevinsen:
Added some debug, occasionally see this

```
00:01:00.180 [sway/tree/arrange.c:293] Arranging workspace '1' at 20.000000, 20.000000
00:01:00.180 [sway/tree/arrange.c:156] Arranging 0x7fffdebbd650 vertically
00:01:00.180 [sway/tree/arrange.c:77] Arranging 0x7fffdebbd540 horizontally
> queuring 7 instructions, 27 queued transactions
> id 4, id 6, id 11, id 7, id 8, id 9, id 10, 
00:01:00.201 [sway/tree/arrange.c:77] Arranging 0x7fffdebbd660 horizontally
> queuring 3 instructions, 28 queued transactions
> id 7, id 8, id 9, 
```

Where the structure is:
```
3: WL-1
   4: vsplit
      6: foot
      10: tab
         11: weston-simple-gl (hidden)
         9: hsplit
            7: foot
            8: foot
[...]

PR #6018, opened 2021-02-12
transactions: Amend pending transactions

kennylevinsen:
The transaction system contains a necessary optimization where a popped
transaction is combined with later, similar transactions. This breaks
the chronological order of states, and can lead to desynchronized
geometries.

To fix this, we replace the queue with only 2 transactions: current and
pending. If a pending transaction exists, it is updated with new state
instead of creating additional transactions.

As we never have more than a single waiting transaction, we no longer
need the queue optimization that is causing problems.

Closes: https://github.com/swaywm/sway/issues/6012

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -23,6 +23,8 @@
 #include "sway/xwayland.h"
 #endif
 
+struct sway_transaction;
+
 struct sway_server {
 	struct wl_display *wl_display;
 	struct wl_event_loop *wl_event_loop;
@@ -85,8 +87,22 @@ struct sway_server {
 	struct wlr_text_input_manager_v3 *text_input;
 	struct wlr_foreign_toplevel_manager_v1 *foreign_toplevel_manager;
 
+	// The timeout for transactions, after which a transaction is applied
+	// regardless of readiness.
 	size_t txn_timeout_ms;
-	list_t *transactions;
+
+	// Stores a transaction after it has been committed, but is waiting for
+	// views to ack the new dimensions before being applied. A queued
+	// transaction is frozen and must not have new instructions added to it.
+	struct sway_transaction *queued_transaction;
+
+	// Stores a pending transaction that will be committed once the existing
+	// queued transaction is applied and freed. The pending transaction can be
+	// updated with new instructions as needed.
+	struct sway_transaction *pending_transaction;
+
+	// Stores the nodes that have been marked as "dirty" and will be put into
+	// the pending transaction.
 	list_t *dirty_nodes;
 };
 
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -86,7 +86,11 @@ static void transaction_destroy(struct sway_transaction *transaction) {
 static void copy_output_state(struct sway_output *output,
 		struct sway_transaction_instruction *instruction) {
 	struct sway_output_state *state = &instruction->output_state;
-	state->workspaces = create_list();
+	if (state->workspaces) {
+		state->workspaces->length = 0;
+	} else {
+		state->workspaces = create_list();
+	}
 	list_cat(state->workspaces, output->workspaces);
 
 	state->active_workspace = output_get_active_workspace(output);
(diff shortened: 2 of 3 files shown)


sway/desktop/transaction.c
```diff
@@ -162,14 +178,33 @@ static void copy_container_state(struct sway_container *container,
...
+
+	// Check if we have an instruction for this node already, in which case we
+	// update that.
+	if (node->ntxnrefs > 0) {
+		for (int idx = 0; idx < transaction->instructions->length; idx++) {
+			struct sway_transaction_instruction *other =
+				transaction->instructions->items[idx];
+			if (other->node == node) {
+				instruction = other;
+				goto update;
```
Xyene, line 191: This can be a `break`, and the code below can be within an `if (!instruction)` block, avoiding a `goto`.

sway/desktop/transaction.c
```diff
@@ -533,20 +527,31 @@ void transaction_commit_dirty(void) {
 	if (!server.dirty_nodes->length) {
 		return;
 	}
-	struct sway_transaction *transaction = transaction_create();
-	if (!transaction) {
-		return;
+
+	struct sway_transaction *transaction = NULL;
+	if (server.transactions->length > 1) {
```
Xyene, line 532: Hmm, I get that doing things like this gets to reuse the dequeuing code, but rather than an array of length 2, I think this would be more descriptive as two separate fields: `server.current_transaction` and `server.pending_transaction` (or whatever naming matches what we do for double-buffering state elsewhere).

Xyene (commented):
Thanks for working on this! The overall approach seems reasonable to me, and I've been running this without being able to induce issues.

sway/desktop/transaction.c
kennylevinsen: I like the goto better, but oh well. Changed.

sway/desktop/transaction.c
kennylevinsen: Agreed, I initially kept the the same just for change simplicity. Updated.

Xyene (approved):
This looks good to me and appears to work in all cases I've tried, thanks.

@emersion, @RPigott -- thoughts?

emersion:
Obligatory bikeshed: `current_transaction` isn't really "current", right? Throughout wlroots, "current" means currently displayed on screen, aka. state used for rendering. But as far as I understand there are two transactions:

- A "pending" transaction, currently being constructed.
- A transaction which has sent `configure` events to clients, but which is still waiting for client ACKs.

Maybe we can rename the latter to "queued"? That's the wording used in the DRM backend when the state has been submitted but not yet applied.

kennylevinsen:
> Obligatory bikeshed:

:)

> Maybe we can rename the latter to "queued"? That's the wording used in the DRM backend when the state has been submitted but not yet applied.

Sure. current/pending did give a bit of a double-buffering vibe, which indeed isn't what is going on here.

sway/desktop/transaction.c
```diff
@@ -472,6 +459,17 @@ static void transaction_commit(struct sway_transaction *transaction) {
 	}
 }
 
+static void transaction_commit_pending() {
```
emersion, line 462: Nit: `(void)`

sway/desktop/transaction.c
```diff
@@ -104,8 +108,16 @@ static void copy_workspace_state(struct sway_workspace *ws,
 	state->layout = ws->layout;
 
 	state->output = ws->output;
-	state->floating = create_list();
-	state->tiling = create_list();
+	if (state->floating) {
+		state->floating->length = 0;
```
emersion, line 112: Why are these changes now necessary? Is it just a bugfix for leaked memory?

If so, please split these into a separate commit.

emersion: Ah, my bad, this is required for the new `transaction_add_node` logic.

include/sway/server.h
```diff
@@ -86,7 +88,8 @@ struct sway_server {
 	struct wlr_foreign_toplevel_manager_v1 *foreign_toplevel_manager;
 
 	size_t txn_timeout_ms;
-	list_t *transactions;
+	struct sway_transaction *current_transaction;
+	struct sway_transaction *pending_transaction;
```
emersion, line 92: Maybe add comments to explain what these are

emersion:
Overall looks good!

include/sway/server.h
kennylevinsen: Added comments for the 4 transaction-related fields.

emersion (approved):
LGTM, thanks a lot!

PR #6019, opened 2021-02-13
container: Move pending state to state struct

kennylevinsen:
This PR moves pending state out of `struct sway_container` and replaces it with a `pending` instance of `struct sway_container_state`. A misuse of pending state during rendering that stuck out is also fixed.

Based on https://github.com/swaywm/sway/pull/6018.


emersion (approved):
LGTM. The first commit is almost completely mechanical.

Xyene (approved):
Looks good, thanks!

PR #6022, opened 2021-02-13
input: Only commit transactions when necessary

kennylevinsen:
There is no need to check for transactions at the end of every user
input, as the vast majority of input will not issue transactions. This
implementation can also hide where changes are made without an
appropriate transaction commit, as a future unrelated input would issue
the commit instead.

Instead, commit transactions in places where changes are made or are
likely to be made.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -9,7 +9,6 @@
 #include <wlr/types/wlr_keyboard_group.h>
 #include <xkbcommon/xkbcommon-names.h>
 #include "sway/commands.h"
-#include "sway/desktop/transaction.h"
 #include "sway/input/input-manager.h"
 #include "sway/input/keyboard.h"
 #include "sway/input/seat.h"
@@ -500,7 +499,6 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 		}
 	}
 
-	transaction_commit_dirty();
 
 	free(device_identifier);
 }
@@ -587,7 +585,6 @@ static int handle_keyboard_repeat(void *data) {
 
 		seat_execute_command(keyboard->seat_device->sway_seat,
 				keyboard->repeat_binding);
-		transaction_commit_dirty();
 	}
 	return 0;
 }
--- a/sway/input/seatop_resize_tiling.c
+++ b/sway/input/seatop_resize_tiling.c
@@ -2,6 +2,7 @@
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/util/edges.h>
 #include "sway/commands.h"
+#include "sway/desktop/transaction.h"
 #include "sway/input/cursor.h"
 #include "sway/input/seat.h"
 #include "sway/tree/arrange.h"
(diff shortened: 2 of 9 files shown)


sway/input/keyboard.c
```diff
@@ -468,6 +469,7 @@ static void handle_key_event(struct sway_keyboard *keyboard,
 
 	if (binding) {
 		seat_execute_command(seat, binding);
+		transaction_commit_dirty();
```
Xyene, line 472: What was the motivation behind the changes in `keyboard.c`? Does `keyboard_execute_compositor_binding` already cause commits internally? (If so, perhaps making `seat_execute_command` or one of its callees fire the commits makes things more symmetrical?)

In general it'd be nice to push the transaction committing as low as possible, to reduce the probability of messing up in new code. If all `seat_execute_command`s must be followed by a `transaction_commit_dirty`, I think the commands should do that themselves.

Xyene (commented):
This change looks mostly good, just a few comments here and there.

sway/input/keyboard.c
kennylevinsen: The reason to do it higher up is to combine multiple operations into a single transaction. Moving it down is a more significant change in behavior.

I don't personally have any attachment to that behavior, and would indeed find it less error prone to move it down, but whether we want that change should be considered.

(Even with the amend logic, the first transaction commits immediately.)

kennylevinsen: I moved the calls from keyboard, but the seatop ones are trickier. They call `arrange_*` and container updates arbitrarily, so we can't easily move the commit down there without causing transactions to be broken up as mentioned above.

sway/input/seatop_resize_tiling.c
```diff
@@ -7,6 +7,7 @@
 #include "sway/tree/arrange.h"
 #include "sway/tree/container.h"
 #include "sway/tree/view.h"
+#include "sway/desktop/transaction.h"
```
Xyene, line 10: Sorry, one more reorder.

sway/input/switch.c
```diff
@@ -59,11 +59,9 @@ static void execute_binding(struct sway_switch *sway_switch) {
 		dummy_binding->command = matched_binding->command;
 
 		seat_execute_command(seat, dummy_binding);
+		transaction_commit_dirty();
```
Xyene, line 62: Is this still necessary now?

sway/input/keyboard.c
Xyene: > Compositor bindings are just the hardcoded VT change keybind.

Ack.

This sounds like a good middle ground, thanks.

sway/input/switch.c
kennylevinsen: Good catch, nope. :)

sway/input/seatop_resize_tiling.c
kennylevinsen: To be fair I should've looked for more myself. :/

Xyene (approved):
Thanks!

synaptiko:
@kennylevinsen I just upgraded to this commit and it seems to be causing the borders around the tiles not to be redrawn in case mouse movements are changing focused view.

kennylevinsen:
Ah, thanks - dealt with in https://github.com/swaywm/sway/pull/6028.

Issue #6023, opened 2021-02-14
Transactions involving hidden X11 or eglSwapInterval(..., 1)-using Wayland windows time out


Xyene:
Pulling this from IRC so that it doesn't get lost...

Consider either layout:

* `V[kitty T[xeyes* xeyes]]`
* `V[kitty T[weston-simple-egl* weston-simple-egl]]`

Attempting to resize the focused window causes transactions to time out due to the hidden window.

PR #6026, opened 2021-02-14
transaction: Only wait for ack from visible views

kennylevinsen:
Transactions currently wait for all configures to be acked, regardless
fo what they were sent to. This includes views that are hidden in tabbed
or stacked containers. If these views do not ack the configure in
response to a single frame callback, they can cause transaction
timeouts.

Check if a container is hidden before registering the configure serial
and saving any view buffers.

Closes: https://github.com/swaywm/sway/issues/6023

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -35,6 +35,7 @@ struct sway_transaction_instruction {
 		struct sway_container_state container_state;
 	};
 	uint32_t serial;
+	bool waiting;
 };
 
 static struct sway_transaction *transaction_create(void) {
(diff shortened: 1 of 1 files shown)


sway/desktop/transaction.c
```diff
@@ -420,13 +420,19 @@ static void transaction_commit(struct sway_transaction *transaction) {
 		struct sway_transaction_instruction *instruction =
 			transaction->instructions->items[i];
 		struct sway_node *node = instruction->node;
+		bool hidden = node_is_view(node) &&
+			container_is_hidden(node->sway_container);
```
Xyene, line 424: Hmm, how does this differ from `!view_is_visible(node->sway_container->view)`?

Xyene:
Thanks for working on this, it does prevent the transactions from timing out. Some interesting behavior I've noticed with this patch applied:

* In `V[weston-simple-egl T[weston-simple-egl* xeyes]]`, resizing causes tearing below the titlebar of the topmost `weston-simple-egl`. This doesn't happen with `xeyes*`.
* `kitty` font size became about 2x bigger(?)

sway/desktop/transaction.c
kennylevinsen: Argh, didn't see that it was implemented on sway_view. It doesn't belong there, as all the tested properties are container properties. :/

The view version is a bit more expensive and tests for scenarios that likely don't matter here, but it's silly to have two, so I'll switch over.

kennylevinsen:
> * In `V[weston-simple-egl T[weston-simple-egl* xeyes]]`, resizing causes tearing below the titlebar of the topmost `weston-simple-egl`. This doesn't happen with `xeyes*`.

I'll take a look, maybe we need some handling for ready by geometry.

> * `kitty` font size became about 2x bigger(?) Note I also have #6018 applied.

Uh what? Just when opened normally? Can you share the WAYLAND_DEBUG?

sway/desktop/transaction.c
Xyene: Agreed it's a bit confusing. I think the logic goes like this, though: if you have `T[A* B]`, do you expect `container_is_visible(B)` to return true? In some sense it should, because the titlebar is visible. But, its contents are not, so it's not useful in the common case.

Perhaps this is just a naming thing though, and calling it `container_view_is_visible` and moving it to `container.h` would be fine.

Xyene (approved):
LGTM, thanks!

PR #6028, opened 2021-02-15
input: Commit transactions in seatop_default

kennylevinsen:
This is needed for focus_follows_mouse, and was accidentally omitted in
the previous transaction commit shuffle.


Xyene (approved):
Thanks!

PR #6033, opened 2021-02-15
build: use wlroots dependency variables

emersion:
Instead of manually parsing header files and having two different
code-paths depending on whether a subproject is used, use
dependency variables which can come from either the subproject or
pkg-config.

References: https://github.com/swaywm/wlroots/pull/2734


Xyene (approved):
Thanks!

Issue #6034, opened 2021-02-16
Tabbed layout doesn't change view instantly on mouse scroll


David96:
### Please fill out the following:
- **Sway Version:** latest master (commit a3d45c32)

- **Debug Log:**
[sway-debug.txt](https://github.com/swaywm/sway/files/5988340/sway-debug.txt)
I cut out the config parsing, the relevant part seems to be:
<details>

```
00:00:03.741 [DEBUG] [sway/commands/bind.c:617] running command for binding: layout tabbed
00:00:03.741 [INFO] [sway/commands.c:255] Handling command 'layout tabbed'
00:00:03.742 [DEBUG] [sway/tree/arrange.c:263] Usable area for ws: 1920x1055@0,25
00:00:03.742 [DEBUG] [sway/tree/arrange.c:293] Arranging workspace '1:>_' at 0.000000, 1105.000000
00:00:03.742 [DEBUG] [sway/tree/arrange.c:77] Arranging 0x7ffe230ad800 horizontally
00:00:03.742 [DEBUG] [sway/desktop/transaction.c:404] Transaction 0x563f9454d390 committing with 4 instructions
00:00:03.746 [DEBUG] [sway/desktop/transaction.c:500] Transaction 0x563f9454d390 is ready
00:00:03.746 [DEBUG] [sway/desktop/transaction.c:303] Applying transaction 0x563f9454d390
00:00:03.746 [DEBUG] [sway/tree/container.c:1194] Container 0x563f944169d0 entered output 0x563f94246750
00:00:03.800 [DEBUG] [sway/tree/view.c:491] Checking criteria [app_id=".*"]
00:00:03.800 [DEBUG] [sway/tree/view.c:493] Criteria already executed
00:00:03.802 [DEBUG] [sway/tree/view.c:491] Checking criteria [app_id=".*"]
00:00:03.803 [DEBUG] [sway/tree/view.c:493] Criteria already executed
00:00:05.102 [DEBUG] [sway/tree/arrange.c:263] Usable area for ws: 1920x1055@0,25
00:00:05.102 [DEBUG] [sway/tree/arrange.c:293] Arranging workspace '1:>_' at 0.000000, 1105.000000
00:00:05.102 [DEBUG] [sway/tree/arrange.c:77] Arranging 0x7ffe230ad8a0 horizontally
00:00:05.470 [DEBUG] [sway/desktop/transaction.c:404] Transaction 0x563f945afeb0 committing with 5 instructions
[...]

PR #6035, opened 2021-02-16
Add missing transaction commits to seatop_default

David96:
See commit description.

I hope I didn't miss anything.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -232,6 +232,7 @@ static void handle_tablet_tool_tip(struct sway_seat *seat,
 				wlr_layer_surface_v1_from_wlr_surface(surface);
 		if (layer->current.keyboard_interactive) {
 			seat_set_focus_layer(seat, layer);
+			transaction_commit_dirty();
 		}
 	} else if (cont) {
 		bool is_floating_or_child = container_is_floating_or_child(cont);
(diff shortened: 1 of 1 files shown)


sway/input/seatop_default.c
```diff
@@ -489,6 +493,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 		node = seat_get_focus_inactive(seat, &cont->node);
 		seat_set_focus(seat, node);
 		seat_pointer_notify_button(seat, time_msec, button, state);
+		transaction_commit_dirty();
```
kennylevinsen, line 496: Nit: In the other places, the order is focus, commit, pointer notify, while here it's focus, pointer notify, commit. It would be good to align the behavior.

sway/input/seatop_default.c
```diff
@@ -503,6 +508,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 			wlr_xwayland_set_seat(xwayland, seat->wlr_seat);
 			seat_set_focus_surface(seat, xsurface->surface, false);
 			seat_pointer_notify_button(seat, time_msec, button, state);
+			transaction_commit_dirty();
```
kennylevinsen, line 511: Same.

sway/input/seatop_default.c
David96: I guess in my head it was "do everything with seat_* and then commit the transaction" :sweat_smile: 
Fixed.

kennylevinsen (approved):
Did you test if it solved https://github.com/swaywm/sway/issues/6034?

LGTM, /cc @Xyene.

David96:
Yes, the last transaction_commit_dirty added fixes that.

Xyene (approved):
Thanks for the fix!

PR #6036, opened 2021-02-16
Stop destroying empty workspaces

jpalus:
Otherwise information about assigned output is lost

### DIFF
--- a/include/sway/tree/workspace.h
+++ b/include/sway/tree/workspace.h
@@ -56,8 +56,6 @@ void workspace_destroy(struct sway_workspace *workspace);
 
 void workspace_begin_destroy(struct sway_workspace *workspace);
 
-void workspace_consider_destroy(struct sway_workspace *ws);
-
 char *workspace_next_name(const char *output_name);
 
 bool workspace_switch(struct sway_workspace *workspace,
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -601,15 +601,13 @@ static struct cmd_results *cmd_move_container(bool no_auto_back_and_forth,
 	// clean-up, destroying parents if the container was the last child
 	if (old_parent) {
 		container_reap_empty(old_parent);
-	} else if (old_ws) {
-		workspace_consider_destroy(old_ws);
 	}
 
 	// arrange windows
 	if (root->fullscreen_global) {
 		arrange_root();
 	} else {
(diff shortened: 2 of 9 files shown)


jpalus:
This might be a bit controversial so let me elaborate a bit more. I have a few workspaces which most of the time have either one window or not at all like for multimedia consumption, viewing docs or reading emails. Depending on what's my current focus these workspaces are either on "primary" screen or on one of "secondary" screens. For example if my main task is to read lengthy document then relevant workspace is on primary screen, however if I need to compare code I work on with docs, then doc workspace is on secondary screen. The point is no workspace has fixed output. Now after switching from i3 I was very frustrated that workspaces do not "remember" their assigned output. Turned out it's only the case for workspaces without windows hence this PR.

I've considered additional options like:
1. Keeping current behavior if number of outputs == 1, but that would change `get_tree` result depending on number of screens, something that is not desired
2. Introducing additional option to decide whether empty workspaces should be removed but I see two issues with that
   * it is very technical and consequences might not be clear to end user
   * it's hard to tell what default value it should have -- keep removing empty workspaces for backward compatibility with sway or stop removing empty workspaces for backward compatibility with i3

Some applications will be affected by this change like waybar but fixing them should be trivial (additional check for number of child nodes).

I'm still quite new to sway so there might be things I'm overlooking.

emersion:
What does i3 do?

emersion:
Note, you can assign workspaces to outputs in the config file.

jpalus:
> What does i3 do?

My bad, looks like my perception was wrong -- i3 behaves exactly the same as sway even though I could have sworn it worked differently. Feel free to close this PR though I must admit remembering output assignment makes more sense to me.

> Note, you can assign workspaces to outputs in the config file.

Yes I'm aware that's why I specifically mentioned "The point is no workspace has fixed output."

RPigott:
@jpalus Some alternatives to swaybar, like Waybar, have a feature/setting where workspace numbers remain shown even when those workspaces don't exist. Could that be what you're thinking off

kennylevinsen:
I'm going to close this as sway's current behavior is identical to i3.

jpalus:
Yes this pull should be closed. I'm planning to implement different, less intrusive approach, but I'll start with i3 and only if accepted same will be implemented for sway.

jpalus:
Just for reference i3/i3#4391 got rejected so nothing will follow in sway.

Issue #6038, opened 2021-02-18
Subsurface damage does not account for popup position


kennylevinsen:
Firefox loves subsurfaces, and thus has subsurfaces with popups with subsurfaces.

When we render these popups, we go through output_view_for_each_popup_surface, while damage, reported through subsurface_commit and ultimately view_child_damage, goes through output_surface_for_each_surface.

The former positions the subsurface box correctly when rendering the content, but the latter positions it wrong for damage, not taking the popup position into account at all and just ends up damaging 0,0.

To reproduce, take nightly firefox in Wayland mode, and open one plugin popup (e.g. clicking on uBlock Origin toolbar icon).

kennylevinsen:
Related firefox issues:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1548747
- https://bugzilla.mozilla.org/show_bug.cgi?id=1548501

PR #6046, opened 2021-02-19
Ensure view children have their parent set

kennylevinsen:
When a popup had a subsurface, view_child_init would fail to set the parent link, leading to the subchild not being able to extract the xdg_popup location.

Includes a fix to check child mapping recursively in order to ensure that we don't try to damage a subsurface before its parent is mapped, as well as an unmapping fix to ensure that we don't try to damage a subsurface *after* its parent has been unmapped and unlinked (which is neither needed nor possible, as the parent is required for position to be known).

Closes: https://github.com/swaywm/sway/issues/6038

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -982,8 +982,18 @@ static void view_child_subsurface_create(struct sway_view_child *child,
 	view_child_damage(&subsurface->child, true);
 }
 
+static bool view_child_is_mapped(struct sway_view_child *child) {
+	while (child) {
+		if (!child->mapped) {
+			return false;
+		}
+		child = child->parent;
+	}
+	return true;
+}
+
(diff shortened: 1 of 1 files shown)


sway/tree/view.c
```diff
@@ -1088,7 +1099,7 @@ void view_child_init(struct sway_view_child *child,
 		wlr_surface_send_enter(child->surface, workspace->output->wlr_output);
 	}
 
-	view_init_subsurfaces(child->view, surface);
```
emersion, line 1081: Hm, indeed calling `view_init_subsurfaces` here doesn't make sense. Good catch!

emersion (approved):
All commits LGTM, thanks!

WhyNotHugo:
Thanks for addressing this so quickly, it was driving me crazy!

PR #6049, opened 2021-02-20
grimshot: add a pipe mode

markstos:
Swappy is a great screenshot annotation tool that expects input in STDIN.

Support tools like that by adding an option to pipe the output to STDOUT.

Example syntax:

     grimshot pipe area | swappy -f -

### DIFF
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -32,13 +32,13 @@ FILE=${3:-$(getTargetDirectory)/$(date -Ins).png}
 
 if [ "$ACTION" != "save" ] && [ "$ACTION" != "copy" ] && [ "$ACTION" != "check" ]; then
   echo "Usage:"
-  echo "  grimshot [--notify] (copy|save) [active|screen|output|area|window] [FILE]"
+  echo "  grimshot [--notify] (copy|save) [active|screen|output|area|window] [FILE|-]"
   echo "  grimshot check"
   echo "  grimshot usage"
   echo ""
   echo "Commands:"
   echo "  copy: Copy the screenshot data into the clipboard."
-  echo "  save: Save the screenshot to a regular file."
+  echo "  save: Save the screenshot to a regular file or '-' to pipe to STDOUT."
   echo "  check: Verify if required tools are installed and exit."
   echo "  usage: Show this message and exit."
   echo ""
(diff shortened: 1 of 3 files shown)


emersion:
What about accepting "-" as a file parameter to use stdout?

markstos:
That would match swappy and is a common idiom.

markstos:
@emersion I tested that `grimshot` could already pipe to STDOUT by setting FILE to "-", it was just an undocumented feature.

I've replaced my previous patch with a new one that simply documents how piping to STDOUT works with the existing code.

contrib/grimshot
```diff
@@ -32,13 +32,14 @@ FILE=${3:-$(getTargetDirectory)/$(date -Ins).png}
...
 if [ "$ACTION" != "save" ] && [ "$ACTION" != "copy" ] && [ "$ACTION" != "check" ]; then
   echo "Usage:"
-  echo "  grimshot [--notify] (copy|save) [active|screen|output|area|window] [FILE]"
+  echo "  grimshot [--notify] (copy|save) [active|screen|output|area|window] [FILE|-]"
   echo "  grimshot check"
   echo "  grimshot usage"
   echo ""
   echo "Commands:"
   echo "  copy: Copy the screenshot data into the clipboard."
   echo "  save: Save the screenshot to a regular file."
```
emersion, line 41: This line should be removed, right?

markstos: This change helps document this special case. Would you like me to remove it?

emersion: I mean, it's the same as the one below, except with less details.

markstos: That is why the line exists-- to be a one-line summary of the usage. 

Would you like me to remove the change?

emersion: The summary for "save" appears twice. Delete one, keep the other.

emersion (approved):
Thanks

contrib/grimshot
markstos: Sorry for the confusion. Your comment appeared attached to the to the "usage" line rather than the duplicated line. I've force -pushed a removal of the duplicated line. 

Thanks for the rapid feedback.

WhyNotHugo:
Nice, thanks for this!

PR #6055, opened 2021-02-22
render: scale titlebars correctly when straddling outputs

### LINKED ISSUE #6054: Titlebar incorrectly scaled when straddling outputs of different scales
This happens on sway/wlroots master, and I think has always been the case.

Screenshot below shows the issue; the thin line divides a 1440p@1 monitor on the left from a 4K@1.5 on the right. The window contents are correctly scaled, but the titlebar is not.

image

For reference, below is how this looks on either monitor.

image

quantum5:
When a container straddles multiple outputs, the title bar is only rendered
at the scale of the "effective" output. If the title bar straddles onto
another output with a different scale factor, it was drawn at the wrong size.

In this commit, we take into consideration the scale the title was rendered
at and scale it accordingly so that it appears at the right size on the other
outputs.

This fixes #6054.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -518,6 +518,10 @@ static void render_titlebar(struct sway_output *output,
 		struct wlr_box texture_box;
 		wlr_texture_get_size(title_texture,
 			&texture_box.width, &texture_box.height);
+
+		float title_scale = container_get_effective_output(con)->wlr_output->scale;
+		texture_box.width = texture_box.width * output_scale / title_scale;
+		texture_box.height = texture_box.height * output_scale / title_scale;
 		ob_title_width = texture_box.width;
 
 		// The title texture might be shorter than the config->font_height,


Xyene (approved):
Looks good and works, thanks!

dimitris-triantafyllidis:
@Xyene, @quantum5 Hi, I got some crashes caused by dereferencing a NULL returned from the call in line 522.

Stack trace:
[stack_trace.txt](https://github.com/swaywm/sway/files/6036711/stack_trace.txt)
Sway log (last lines):
[sway.log](https://github.com/swaywm/sway/files/6036715/sway.log)

I have two reproducible crash scenarios:

1. Open two terminals, float them together by focusing ancestor and move them both towards an edge of the screen. As soon as of them leaves the screen, sway crashes.
2. Open a terminal, float it and make it a bit small (for convenience), split it, set it to stacked and start adding children until the title bars overflow the container. Then split the last container and sway crashes.

Issue #6044, opened 2021-02-19
Scrolling on titlebars using touchpad doesn't change windows


David96:
In sway, using a two finger scroll on the touchpad over the titlebar of a stacked/tabbed layout doesn't change the window. The reason for that lies here:
https://github.com/swaywm/sway/blob/4e02a598a33fb9b306a1ab220438adf82cac07a6/sway/input/seatop_default.c#L713

At least for my touchpad, delta_discrete is always 0. One had to use delta to figure out to which window to change. Now that has some problems.

Simply checking whether delta is >/<0 results in extremely fast scrolling as the resolution is pretty high. Checking whether it's over some threshold isn't optimal either as the delta is a delta to the last scroll event so adding a threshold just waits until you reach a certain speed but if you do you likely scroll more windows than intended.

I'm not sure how to tackle this, to get a good user experience I think we had to keep track of the scroll sequence and sum the deltas to make the window change depend on distance scrolled. I also understand if you don't think it's worth the additional trouble. Or maybe there's an easy solution I just didn't find yet.

I just checked i3 and it did work there, though quite fast. I think this is more a sideeffect of the X input system than an intended behavior but anyway, since in i3 it works I filed it as i3-compat (relevant code seems to be this: https://github.com/i3/i3/blob/b95bab0e34365d7637cc87f11e0b3e5fa0956dfb/src/click.c#L213 so no special handling for touchpads).

For myself I'm currently using this patch:
```diff
diff --git a/sway/input/seatop_default.c b/sway/input/seatop_default.c
index f9eb8c8a..2c7e42bd 100644
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -709,8 +709,9 @@ static void handle_pointer_axis(struct sway_seat *seat,
 			struct sway_node *active =
 				seat_get_active_tiling_child(seat, tabcontainer);
 			list_t *siblings = container_get_siblings(cont);
+			float delta = event->delta_discrete == 0 ? event->delta : event->delta_discrete;
 			int desired = list_find(siblings, active->sway_container) +
-				round(scroll_factor * event->delta_discrete);
+				round(scroll_factor * delta);
 			if (desired < 0) {
 				desired = 0;
 			} else if (desired >= siblings->length) {
```
Which works okish but for the reasons explained above isn't optimal at all.

wmww:
[...]

PR #6056, opened 2021-02-22
Touchpad scroll on titlebar

David96:
This is to figure out how #6044 could look like.

It's only meant as a POC. I'm not exactly sure were to place the functions that I now just put in util.c.

The main issue I see is that I don't know when to reset the distance_scrolled because there doesn't seem to be a "two fingers down"/"two fingers up" like event from libinput.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -12,6 +12,7 @@
 #include "sway/tree/view.h"
 #include "sway/tree/workspace.h"
 #include "log.h"
+#include "util.h"
 #if HAVE_XWAYLAND
 #include "sway/xwayland.h"
 #endif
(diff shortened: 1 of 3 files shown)


sway/input/seatop_default.c
```diff
@@ -709,8 +710,17 @@ static void handle_pointer_axis(struct sway_seat *seat,
 			struct sway_node *active =
 				seat_get_active_tiling_child(seat, tabcontainer);
 			list_t *siblings = container_get_siblings(cont);
+			int delta_discrete = event->delta_discrete;
+			if (delta_discrete == 0) {
+				scroll_distance_add(event->delta);
+				float dist = scroll_distance_get();
+				delta_discrete = round(dist / 10.);
```
David96, line 717: This threshold of 10 should be user configurable I guess to make up for differences in touchpad models.

PR #6057, opened 2021-02-22
container: Add container_is_current_floating

kennylevinsen:
Needed to check if containers are currently floating from render code,
as container_is_floating checks pending state.


emersion (approved):
Thanks!

PR #6059, opened 2021-02-23
WIP: sway: add titlebar_v_height option 

vvrein:
This pr is in wip state, as there is need to fix some things:

- [x] 1) We need new name for `fixed titlebar height`, since in code there already is a `titlebar_height` variable 
- [ ] 2) I need to remind myself what for `baseline` stands and find out how it can be calculated more correctly
- [x] 3) Remove or rewrite logging messages

Base logic is:
- If there is no configured `fixed titlebar height` - then variable assigns value `-1` and actual title height goes to be calculated with cairo.
- Same if there is explicit `-1` in config.
- If `fixed titlebar height` is set to 0 and above - title height is then picked from config.

----

For 1) my variants are:
- titlebar_height_fixed  <-- and my vote for this 
- titlebar_height_bind
- titlebar_height_exact

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -578,16 +578,22 @@ void container_update_title_textures(struct sway_container *container) {
 }
 
 void container_calculate_title_height(struct sway_container *container) {
+        int height;
+        int baseline;
+
 	if (!container->formatted_title) {
 		container->title_height = 0;
 		return;
 	}
-	cairo_t *cairo = cairo_create(NULL);
-	int height;
-	int baseline;
-	get_text_size(cairo, config->font, NULL, &height, &baseline, 1,
-			config->pango_markup, "%s", container->formatted_title);
-	cairo_destroy(cairo);
+	if (config->titlebar_height >= 0) {
+		height   = config->titlebar_height;
+		baseline = 0;
+	} else {
+		cairo_t *cairo = cairo_create(NULL);
+		get_text_size(cairo, config->font, NULL, &height, &baseline, 1,
+				config->pango_markup, "%s", container->formatted_title);
+		cairo_destroy(cairo);
+	}
 	container->title_height = height;
 	container->title_baseline = baseline;
 }
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -188,6 +188,7 @@ sway_cmd cmd_title_align;
 sway_cmd cmd_title_format;
 sway_cmd cmd_titlebar_border_thickness;
 sway_cmd cmd_titlebar_padding;
+sway_cmd cmd_titlebar_height;
 sway_cmd cmd_unbindcode;
 sway_cmd cmd_unbindswitch;
 sway_cmd cmd_unbindsym;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -492,6 +492,7 @@ struct sway_config {
 	int titlebar_border_thickness;
 	int titlebar_h_padding;
 	int titlebar_v_padding;
+	int titlebar_height;
 	size_t urgent_timeout;
 	enum sway_fowa focus_on_window_activation;
 	enum sway_popup_during_fullscreen popup_during_fullscreen;
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -89,6 +89,7 @@ static const struct cmd_handler handlers[] = {
 	{ "tiling_drag_threshold", cmd_tiling_drag_threshold },
 	{ "title_align", cmd_title_align },
 	{ "titlebar_border_thickness", cmd_titlebar_border_thickness },
+	{ "titlebar_height", cmd_titlebar_height },
 	{ "titlebar_padding", cmd_titlebar_padding },
 	{ "unbindcode", cmd_unbindcode },
 	{ "unbindswitch", cmd_unbindswitch },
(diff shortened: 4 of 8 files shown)


oranenj:
Do the padding and border width variables get used with a fixed titlebar height? I think the "v" might be redundant in the option name, so perhaps just "titlebar_height"?

Your definition of the configuration option doesn't allow "-1", but I think that's fine; if people want to make it automatic, they should just remove the configuration option. Also, allowing a height less than 1 makes no sense and probably results in weird behaviour.

You're also missing documentation for the variable, don't forget that :)

sway/tree/container.c
```diff
@@ -573,16 +573,32 @@ void container_update_title_textures(struct sway_container *container) {
 }
 
 void container_calculate_title_height(struct sway_container *container) {
+	sway_log(SWAY_INFO, "===> container_calculate_title_height");
```
oranenj, line 576: I think most of this logging isn't  very useful, so to make this PR mergeable, you should remove it.

Other than that, is there anything here that's actually WIP anymore? The Alpine build failure doesn't seem related to the code.

sway/tree/container.c
```diff
@@ -573,16 +573,32 @@ void container_update_title_textures(struct sway_container *container) {
...
-	int height;
-	int baseline;
-	get_text_size(cairo, config->font, NULL, &height, &baseline, 1,
-			config->pango_markup, "%s", container->formatted_title);
-	cairo_destroy(cairo);
+	if (config->titlebar_v_height >= 0) {
+		sway_log(SWAY_INFO, "===> config->titlebar_v_height >= 0");
+
+		height   = config->titlebar_v_height;
+		baseline = round(height * 0.77);
```
oranenj, line 590: I'm not sure if I like this magic number, but I don't have any better ideas either. Any other reviewers have opinions about this?

oranenj (changes requested):
A stylistic nitpick about commit messages: generally imperative mood reads better, so I personally would prefer something akin to "sway: add titlebar_v_height option to configure static titlebar height". Might even leave it as "add titlebar_v_height option" because the name is decently self-explanatory.

vvrein:
@oranenj Sorry me for those verbose actions... That was my auto rebuld script
<details><summary>script</summary>

```bash
#!/bin/bash
pushd sway
git fetch --all
git reset origin/master --hard
meson setup --wipe ./build

pushd subprojects/wlroots
git fetch --all
git reset origin/master --hard
popd

git rebase upstream/master
meson build --reconfigure && ninja -C build && git push origin master --force
```
</details>

I'm now going to comb the code, and take it to mergeable state. Thank for your advices :)

vvrein:
I found a strange thing, which can explain why I used `_v_` in parameter name:  
https://github.com/vvrein/sway/tree/titlebar_height_fixed
This code compiles, but then sway shows an error like "unknown command `titlebar_padding 0 0`"
The same is for `titlebar_height` name.

image


Since I could not find any obvious reason for this behavior - I better leave titlebar_v_height name for now

vvrein:
About magic `baseline = round(height * 0.77);`


<details><summary>screenshot</summary> 

grimcap

</details>
 
 I looked at what cairo does, when it calculates baseline, and then picked up value that was near.
 ```
 >>> 12/15
0.8
>>> 12/16
0.75
>>> 14/19
0.7368421052631579
>>>
```
Also, in my experiments, baseline seems does not make any job, since hardcoding values like 10, 20, 30 does not change anything.

**Comment by vvrein** [comment]
A little summary for now:
- I deleted all my debug logging from container_calculate_title_height
- Added a little note to manual
- Left the old variable name

Questions:
- If another than `titlebar_v_height` variable name is desirable, then I need some help with https://github.com/swaywm/sway/pull/6059#issuecomment-812893750
- Need I somehow extend note about `titlebar_v_height`?

**Comment by oranenj** [comment]
@vvrein The `titlebar_padding` issue is because after you add titlebar_height_fixed, the array is no longer sorted. You need to add it before titlebar_padding: https://github.com/vvrein/sway/blob/b7d6ca952331261cc3c45dd51d7391a89e010fd8/sway/commands.c#L44

**Comment by emersion** [comment]
> I looked at what cairo does, when it calculates baseline, and then picked up value that was near.

I'm not following. Can you link the cairo code? Where do these values come from?

Can't we use something like this? https://www.cairographics.org/manual/cairo-cairo-scaled-font-t.html#cairo-font-extents-t

File: sway/tree/container.c [hunk 3]
```diff
@@ -573,16 +573,22 @@ void container_update_title_textures(struct sway_container *container) {
 }
 
 void container_calculate_title_height(struct sway_container *container) {
+        int height;
```
Exagone313, line 576: With project code style, these lines should be moved after the `if (!container->formatted_title)` block.

sway/tree/container.c
Exagone313: Since the value is not used elsewhere, the baseline value could also be a configuration directive.

FrostKiwi:
Really happy with what this PR brings. I can misuse this to create the broder I always wanted: Show selection based on just a line at the top, while the title is shown in the bar.
20210607_18h41m14s_grim

vvrein:
This commit changes option name from `titlebar_v_height` to `titlebar_height`,   
and sets `baseline` exactly to zero, since afterf it would be updated here https://github.com/swaywm/sway/blob/master/sway/config.c#L1007-L1009

vvrein:
Closing in favor of https://github.com/swaywm/sway/pull/6433

PR #6060, opened 2021-02-23
build: Add dependency on libdrm

tadeokondrak:
As of 66343839b146a54505b746784cd42a8efb844963, sway now uses a
libdrm header. Add this dependency to the build system so headers from
it can be used on systems where pkg-config is required to find them.

(Encountered on NixOS.)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -52,6 +52,8 @@ glesv2         = dependency('glesv2')
 libevdev       = dependency('libevdev')
 libinput       = dependency('libinput', version: '>=1.6.0')
 xcb            = dependency('xcb', required: get_option('xwayland'))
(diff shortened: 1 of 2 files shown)


emersion:
Can we use `partial_dependency` to avoid linking to libdrm?

tadeokondrak:
> Can we use partial_dependency to avoid linking to libdrm?

I've done that now. I enabled `compile_args` and `includes` from it, but pkg-config only uses `compile_args`.
I'm not sure whether to enable `includes`, it might help in case meson gets smarter about parsing pkg-config files or uses something other than pkg-config.

emersion (approved):
LGTM, thanks!

emersion:
Yeah, might be better to keep `includes` for future-proofing.

PR #6062, opened 2021-02-24
output: Reconfigure xcursor when applying output config

tadeokondrak:
Before this commit, when an output had its scale dynamically changed,
Sway would not load a cursor theme with the new scale. This results
in stale cursor images when moving the cursor into an area controlled
by the compositor, like the background or resize areas.

To reproduce:
- Using IPC, set an output scale to a value that isn't currently used
- Move the cursor into a compositor-controlled area
- The cursor will not change


emersion (approved):
Thanks!

PR #6063, opened 2021-02-24
render: handle containers without output when rendering titles

### LINKED ISSUE #6061: container_get_effective_output returning NULL in render_titlebar, leading to crashes
- **Sway Version:**
  - master

- **Debug Log:**
  - https://github.com/swaywm/sway/files/6036715/sway.log
 
- **Stack Trace:**
  - https://github.com/swaywm/sway/files/6036711/stack_trace.txt

Started appearing after #6055.

I have two reproducible crash scenarios:

1. Open two terminals, float them together by focusing ancestor and move them both towards an edge of the screen. As soon as of them leaves the screen, sway crashes.
2. Open a terminal, float it and make it a bit small (for convenience), split it, set it to stacked and start adding children until the title bars overflow the container. Then split the last container and sway crashes.

quantum5:
In e0a94bee8da3271f942c0881ee18a7e2d4138063, it was believed that if the
container is being rendered, it must have an output.

This turned out not to be the case. When rendering a container, all its
children are rendered, even if the children is positioned off screen and
thus not having any output. This is the cause of the crash in #6061.

This commit introduces a null-check, which fixes #6061.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -285,6 +285,7 @@ bool container_is_fullscreen_or_child(struct sway_container *container);
 /**
  * Return the output which will be used for scale purposes.
  * This is the most recently entered output.
+ * If the container is not on any output, return NULL.
  */
 struct sway_output *container_get_effective_output(struct sway_container *con);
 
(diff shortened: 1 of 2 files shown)


Xyene (approved):
This looks good, thanks.

I think in the general case we'd make some of this logic less surprising if we didn't render offscreen views (this is not the first time `container_get_effective_output` has been nullchecked), but this is still an improvement over crashing.

PR #6064, opened 2021-02-25
feat: add option to set focused window opacity

gbaranski:
Also i've fixed some inconsistency in variable naming

sway


gibbz00:
Why was this pull request closed? Was just thinking of implementing something similar.

PR #6075, opened 2021-03-03
input: stash current node and surface in cursor

### LINKED ISSUE #6073: Pointer constraint view hittest is wrong for floating container
Consider the following floating layout, the thin gray vertical line showing the split between two outputs:

image

The yellow region is an instance of wlroots' `examples/pointer-constraints confine persistent no-region`, **and is focused**.

The terminal window displays over `pointer-constraints` despite it not having focus, since its anchor point is on the other window. This is _maybe_ correct behavior, though I don't have a multi-monitor i3 to test. However, when the constrained pointer crosses under the terminal window, sway's "surface at" test hits the terminal window first.

This has the effect of bailing out of `pointer_motion` early:

https://github.com/swaywm/sway/blob/d358aab8d9dd6121533326423759394ca804a05a/sway/input/cursor.c#L358-L360

...and the pointer gets locked into place until window focus is changed via keybindings, as it is no longer processing motion events.

Xyene:
If an actively constrained surface is partially occluded by another node
(say, by a floating node), the cursor should remain constrained and pass
"under" the floating node.

Previously, this did not always happen, as `node_at_coords` is a simple
top-down hit test that doesn't cover constraints.

Furthermore, whether the cursor should remain constrained is a function
of the input device type: only pointer devices should be constrained.

This commit addresses this issue by stashing the surface and node that
currently have cursor input focus, and using the stashed values
everywhere.

When `wlr_cursor_move` and friends are called, one must now take care to
also call `cursor_update_current_node` with the `wlr_input_device` that
initiated the movement.

Fixes #6073.

---

The input logic is really finicky, so while this is ready for review I'd prefer
to not merge it until after 1.6. There are bound to be bugs.

Note that I have only tested this code with pointer input; I'll be able to
test with a tablet later, but I do not own a touchscreen device.

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -22,6 +22,15 @@ struct sway_cursor {
 		double x, y;
 		struct sway_node *node;
 	} previous;
+
+	// Not to be confused with `previous` above; they store entirely unrelated
+	// data.
+	struct {
+		double sx, sy;
+		struct sway_node *node;
+		struct wlr_surface *surface;
+	} current;
+
 	struct wlr_xcursor_manager *xcursor_manager;
 	struct wl_list tablets;
 	struct wl_list tablet_pads;
@@ -78,10 +87,6 @@ struct sway_cursor {
 };
 
 struct sway_node;
-
-struct sway_node *node_at_coords(
-		struct sway_seat *seat, double lx, double ly,
-		struct wlr_surface **surface, double *sx, double *sy);
 
 void sway_cursor_destroy(struct sway_cursor *cursor);
 struct sway_cursor *sway_cursor_create(struct sway_seat *seat);
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -618,11 +618,7 @@ void seat_execute_command(struct sway_seat *seat, struct sway_binding *binding) 
 	struct sway_container *con = NULL;
 	if (binding->type == BINDING_MOUSESYM
 			|| binding->type == BINDING_MOUSECODE) {
-		struct wlr_surface *surface = NULL;
-		double sx, sy;
-		struct sway_node *node = node_at_coords(seat,
-				seat->cursor->cursor->x, seat->cursor->cursor->y,
-				&surface, &sx, &sy);
+		struct sway_node *node = seat->cursor->current.node;
 		if (node && node->type == N_CONTAINER) {
 			con = node->sway_container;
 		}
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -74,7 +74,7 @@ static struct wlr_surface *layer_surface_popup_at(struct sway_output *output,
  * Returns the node at the cursor's position. If there is a surface at that
  * location, it is stored in **surface (it may not be a view).
  */
-struct sway_node *node_at_coords(
+static struct sway_node *node_at_coords(
 		struct sway_seat *seat, double lx, double ly,
 		struct wlr_surface **surface, double *sx, double *sy) {
 	// check for unmanaged views first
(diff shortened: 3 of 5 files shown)


Xyene:
I should say I'm not a fan of the extra pointer state this PR introduces. It _works_, and it does fix the bug, but it seems too easy to mess up and not call `cursor_update_current_node` everywhere it should be called. So, this isn't really "ready for review" in the sense that I'd necessarily like to get this merged, more that it's "ready for someone other than me to take a look at it and see what they think about this approach".

> Previously, this did not always happen, as `node_at_coords` is a simple top-down hit test that doesn't cover constraints.

It's possible that changing `node_at_coords` to take a device type, and short-circuiting out in the pointer device + over constrained surface case may result in a cleaner patch. Some plumbing of device types through seatops would be needed.

What, exactly, "`node_at_coords`" will actually do in that case seems hard to describe succinctly -- certainly "`node_at_coords`" becomes an insufficient name, and I haven't been able to think up a satisfactory one. This alone would indicate to me that that function does too much, but I also can't think of any sensible way to split it up.

kennylevinsen:
> I should say I'm not a fan of the extra pointer state this PR introduces. 

Yeah, it would make sense if it meant we could cache the value and save a bunch of lookups, but one would have to hammer mouse buttons while keeping the mouse entirely stationary for that to matter...

> t's possible that changing node_at_coords to take a device type, and short-circuiting out in the pointer device + over constrained surface case may result in a cleaner patch

How about a `constrained_node_at_coords`, which would be more or less identical to your current `cursor_update_current_node`? It would then be called instead of `node_at_coords` when constraints should be observed.

nowrep:
> the cursor should remain constrained and pass "under" the floating node.

Should it really work like this? I would still want to eg. dismiss the notification with mouse click.

It can also happen with a close dialog - application confines the cursor to the window and on close it shows confirm dialog. In this case, you should be able to interact with the dialog.

Xyene:
> Should it really work like this? I would still want to eg. dismiss the notification with mouse click.

That sounds like it'd be very confusing. Imagine playing a game that captures the mouse (specifically, confines it to the window surface, and hides the hardware cursor) when a notification pops up. You'd try to ignore it, but while swiveling your camera around, the real (hidden) cursor goes under the popup, and the game loses focus/you lose/etc. That seems bad?

> In this case, you should be able to interact with the dialog.

Yes, but in this case the application should release the confine. There's no guarantee that the dialog would appear over the surface of the application (and thus, over the confined region) in the first place.

nowrep:
> Imagine playing a game that captures the mouse (specifically, confines it to the window surface, and hides the hardware cursor) when a notification pops up. You'd try to ignore it, but while swiveling your camera around, the real (hidden) cursor goes under the popup, and the game loses focus/you lose/etc. That seems bad?

If the game locks your cursor and hides it (for camera control) then you won't be able to reach the notification because the mouse cursor doesn't move.
You can only enter the notification when the game has confined the cursor but not locked it. And in this case, I think the expected behavior is to send the mouse events to the notification window (that is displayed on top = under mouse cursor) - also this is how it works on i3.

I've been running this patch and it does exactly what I want, and it still seems to correctly keep the cursor inside the confined window in windowed case (non-fullscreen).

```diff
diff --git a/sway/input/cursor.c b/sway/input/cursor.c
index d8b1abeb..fd295f72 100644
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -364,18 +364,16 @@ static void pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
                node_at_coords(cursor->seat,
                        cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
 
-               if (cursor->active_constraint->surface != surface) {
-                       return;
-               }
+               if (cursor->active_constraint->surface == surface) {
+                       double sx_confined, sy_confined;
+                       if (!wlr_region_confine(&cursor->confine, sx, sy, sx + dx, sy + dy,
+                                       &sx_confined, &sy_confined)) {
+                               return;
+                       }
 
-               double sx_confined, sy_confined;
-               if (!wlr_region_confine(&cursor->confine, sx, sy, sx + dx, sy + dy,
-                               &sx_confined, &sy_confined)) {
-                       return;
+                       dx = sx_confined - sx;
+                       dy = sy_confined - sy;
                }
-
-               dx = sx_confined - sx;
-               dy = sy_confined - sy;
        }
 
        wlr_cursor_move(cursor->cursor, device, dx, dy);
```

Issue #6074, opened 2021-03-02
Changing tab with mouse click is not working anymore


Moonlight-Angel:
- **sway version:** 1.5-d358aab8 (Mar  2 2021, branch 'master')

- **Description:**
After upgrading to the latest commit on `master`, it looks like I'm not able to change tabs with mouse click anymore.
I'm not able to find a commit that would indicate a change of behavior, so I'm not sure if it's intended behavior.

- **Quick steps to reproduce**
  - Switch to tab or stacked layout
  - Open 2 or more windows
  - Try changing tab focus with mouse clicks

tsdh:
I can confirm that.  It also affects the stacked layout.

kennylevinsen:
Observations from quick local test:

1. It works for floating containers.
2. It works for *unfocused* tiling containers - sneaking the cursor in from the top without touching the surface content to click a titlebar works.
3. It does not work for *focused* tiling containers.

Sounds like an edge-case in focus handling or the default seat ops.

kuon:
Also, if grabbing/drag the tab that is not focused and moving it, it actually moves the focused tab.

Xyene:
I'm on 1afedcb94c0517b5434e2220e63b997cf01427bd and things work. So, maybe a6544f5a64e5d0899555795915d9a05f7bb09f9c or d358aab8d9dd6121533326423759394ca804a05a?

kennylevinsen:
> I'm on [1afedcb](https://github.com/swaywm/sway/commit/1afedcb94c0517b5434e2220e63b997cf01427bd) and things work. So, maybe [a6544f5](https://github.com/swaywm/sway/commit/a6544f5a64e5d0899555795915d9a05f7bb09f9c) or [d358aab](https://github.com/swaywm/sway/commit/d358aab8d9dd6121533326423759394ca804a05a)?

Indeed, the latter met a quirk in view_container_at.

https://github.com/swaywm/sway/pull/6077 should fix it.

tsdh:
I can confirm it is fixed now.

PR #6077, opened 2021-03-03
container: Use content geometry in view_container_at

kennylevinsen:
container: Add view_container_content_at

container_at checks if the position provided matches the currently
focused container with view_container_at as a fast path.

view_container_at checks using the main container geometry, which
includes the titlebar and border area. If a tabbed container is focused,
then positions over unfocused tabs are incorrectly reported as belonging
to the focused container, breaking focus on click.

Add view_container_content_at for use in the focused container fast path
which only tests container content area, and fall back to full workspace
scans for border and titlebar areas.

Closes: https://github.com/swaywm/sway/issues/6074

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -311,7 +311,30 @@ static struct sway_container *floating_container_at(double lx, double ly,
 	return NULL;
 }
 
-struct sway_container *view_container_at(struct sway_node *parent,
+static struct sway_container *view_container_content_at(struct sway_node *parent,
+		double lx, double ly,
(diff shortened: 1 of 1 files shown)


adee:
This introduces a new bug for floating windows: mouse over the titlebar does not focus it anymore.

kennylevinsen:
> This introduces a new bug for floating windows: mouse over the titlebar does not focus it anymore.

Nice catch, should be fixed now.

emersion (approved):
Looks like a good idea to me.

Thanks!

PR #6093, opened 2021-03-12
resolve #6092: silence IconThemePath message

### LINKED ISSUE #6092: Silence error message regarding IconThemePath property?
- We are not accepting any new window management features unless they get implemented by i3.

- **Description:**
  #3249 uses `IconThemePath` property from a `StatusNotifierItem` object. And as that PR denotes, `IconThemePath` is not in [XDG's specification](https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/#org.freedesktop.statusnotifieritem.tooltip). It's from [KDE](https://github.com/KDE/kde-workspace/commit/a54f9ec9dbbe7001f5a38c15c7c7371ab0932b1b). That is to say, many applications that only follow XDG's specification do not have `IconThemePath`. For example, I think those using qt5 do not have `IconThemePath` (though those using qt4 and sni-qt do).

  When Sway encounters applications that do not implement `IconThemePath`, it dumps a lot of error messages in syslog (through function `sni_get_property_async`, ~[line 182](https://github.com/swaywm/sway/blob/bc239b2f6b0cf9d29adb8df7e464e7d104cad4bb/swaybar/tray/item.c#L182)~ [line 121](https://github.com/swaywm/sway/blob/bc239b2f6b0cf9d29adb8df7e464e7d104cad4bb/swaybar/tray/item.c#L121-L123)). These messages are somehow annoying. Is it possible to silence messages regarding `IconThemePath`? Or at least lowering down the importance of the messages so that the messages are only printed when we use `--debug` flag?

Thanks!

piyueh:
`IconThemePath` is not a standard property in XDG's StatusNotifierItem specification, so missing this property should not be logged as an error.

This patch changes the log level to `SWAY_DEBUG` when `swaybar` queries the value of `IconThemePath` so that `swaybar` won't log the returned message as an error if `IconThemePath` does not exist. It's a quick fix and may not look nice though...

Now, if users want to see the return message regarding `IconThemePath`, they'll have to enable the debug flag, like `swaybar -b bar-0 --debug`.

### DIFF
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -115,10 +115,11 @@ static int get_property_callback(sd_bus_message *msg, void *data,
 	const char *prop = d->prop;
 	const char *type = d->type;
 	void *dest = d->dest;
+	sway_log_importance_t log_lv = strcmp(prop, "IconThemePath") ? SWAY_ERROR : SWAY_DEBUG;
 
 	int ret;
 	if (sd_bus_message_is_method_error(msg, NULL)) {
-		sway_log(SWAY_ERROR, "%s %s: %s", sni->watcher_id, prop,
+		sway_log(log_lv, "%s %s: %s", sni->watcher_id, prop,
 				sd_bus_message_get_error(msg)->message);
 		ret = sd_bus_message_get_errno(msg);
 		goto cleanup;
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -115,12 +115,16 @@ static int get_property_callback(sd_bus_message *msg, void *data,
 	const char *prop = d->prop;
 	const char *type = d->type;
 	void *dest = d->dest;
-	sway_log_importance_t log_lv = strcmp(prop, "IconThemePath") ? SWAY_ERROR : SWAY_DEBUG;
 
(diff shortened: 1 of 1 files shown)


alkino (approved):

swaybar/tray/item.c
```diff
@@ -115,10 +115,11 @@ static int get_property_callback(sd_bus_message *msg, void *data,
 	const char *prop = d->prop;
 	const char *type = d->type;
 	void *dest = d->dest;
+	sway_log_importance_t log_lv = strcmp(prop, "IconThemePath") ? SWAY_ERROR : SWAY_DEBUG;
```
emersion, line 118: Can we also check the value of `sd_bus_message_get_errno()` as well here? Just to make sure the error is because of a missing property and not something else.

piyueh: I updated the code. Now using the `name` member of the error object to determine whether this error is about missing property or not. I didn't use `sd_bus_message_get_errno()` because I'm not sure if the error number will be changed or not in the future.

emersion (approved):
LGTM, thanks!

PR #6094, opened 2021-03-12
xwayland: simplify override-redirect focus restoration

emersion:
No need to grab the xwayland surface from the wlr_surface, the
parent is already an xwayland surface.


kennylevinsen (approved):
LGTM

PR #6098, opened 2021-03-15
Add icon to .desktop file

throwaway1037:
I chose the third version of the logo with text in SVG format. Feel free to change this if you want.

Also, I don't know whether the icon filepath will need to be changed after installation; I put it in relative to the `assets/` directory.

### DIFF
--- a/sway.desktop
+++ b/sway.desktop
@@ -3,3 +3,4 @@ Name=Sway
 Comment=An i3-compatible Wayland compositor
 Exec=sway
 Type=Application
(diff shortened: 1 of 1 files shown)


markstos:
Test the "Compiling from Source" instructions through `sudo ninja -C build install` and then test that the icon works after it has been installed.

How will this icon be used? Do some display managers have the option to display the icon for the options of window managers or desktop environments you can choose?

emersion (changes requested):
As stated above, won't work when Sway is installed.

PR #6099, opened 2021-03-15
Add toggle logic inside DPMS handler

### LINKED ISSUE #5929: DPMS toggle
Outputs can be toggled without DPMS, but only `on` and `off` are provided for DPMS. Strangely, Sway does respond to `output dpms toggle` but it seems to just do the same thing as `output dpms off`.

ivanfed0t0v:
Logic that obtains current DPMS state is put inside the handler.

sway_output from which the current DPMS state will be obtained is selected by the following logic:
* For `-` and `--` the focused output is used;
* For `*` error "Cannot apply toggle to all outputs" is reported;
* For everything else all_output_by_name_or_id() is used.

Fixes #5929.

### DIFF
--- a/sway/commands/output/dpms.c
+++ b/sway/commands/output/dpms.c
@@ -1,6 +1,8 @@
 #include "sway/commands.h"
 #include "sway/config.h"
+#include "sway/output.h"
 #include "util.h"
+#include <strings.h>
 
 struct cmd_results *output_cmd_dpms(int argc, char **argv) {
 	if (!config->handler_context.output_config) {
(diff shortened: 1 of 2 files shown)


sway/commands/output.c
```diff
@@ -62,8 +62,30 @@ struct cmd_results *cmd_output(int argc, char **argv) {
...
+		} else {
+			output->dpms_state = DPMS_ON;
+		}
 	} else {
 		output = new_output_config(argv[0]);
+
+		// Obtain current DPMS state for the specified output,
+		// on wildcard just choose focused output
+		struct sway_output *sway_output = NULL;
+		if (strcmp(argv[0], "*") == 0) {
```
RPigott, line 76: This would modify the output dpms state for non-dpms wildcard output commands too though.

I don't think we should modify the output config outside of the handler, so all this logic should probably go into output/dpms.c.

RPigott:
Hi, thanks for the PR.

The new intended behavior makes sense to me, at least. I doubt it would make much sense to toggle each output state individually.

sway/commands/output.c
ivanfed0t0v: Oh, I missed that bit, thanks. Gonna move it inside the handler.

Thought it'd be a bit cleaner to put this logic here. But now, looking at output/toggle.c, it makes a lot more sense to just stay inside the handler.

ivanfed0t0v:
> I doubt it would make much sense to toggle each output state individually.

I agree and I don't think it's currently possible to toggle DPMS on each output individually from inside the handler anyway.

Actually, handler at output/toggle.c simply reports an error when called on wildcard. I think I should do the same with DPMS for consistency.

RPigott (changes requested):
Just one small thing – now that dpms has a "toggle" argument could you please add it to the sway-output man page? I think the meaning is clear, so just listing on|off|toggle for the argument of dpms is probably fine.

Other than that, this LGTM. Should probably be squashed for merging.

emersion (approved):
Looks good to me, and it seems like @RPigott is fine with this version too.

Thanks!

PR #6101, opened 2021-03-16
config: allow whitespaces in config path

columbarius:
Backport configpath building from xdg-desktop-portal-wlr.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -338,34 +338,58 @@ static bool file_exists(const char *path) {
 	return path && access(path, R_OK) != -1;
 }
 
+static char *config_path(const char *prefix, const char *config_folder) {
+	if (!prefix || !prefix[0] || !config_folder || !config_folder[0]) {
+		return NULL;
+	}
+
+	const char *filename = "config";
+
+	size_t size = 3 + strlen(prefix) + strlen(config_folder) + strlen(filename);
+	char *path = calloc(size, sizeof(char));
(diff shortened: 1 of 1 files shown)


sway/config.c
```diff
@@ -338,34 +338,62 @@ static bool file_exists(const char *path) {
...
-			if (file_exists(path)) {
-				return path;
-			}
-			free(path);
+	struct config_path {
+		const char *prefix;
+		const char *config_folder;
+	};
+
+	static struct config_path config_paths[6];
```
emersion, line 370: Hm, this is getting a little out of hand. Maybe can be simplified to:

```c
struct config_path config_paths[] = {
    { .prefix = "asdf", .config_folder = "asdf" },
    …
};
```

columbarius: I would like to do that, but that does only work for constant strings and not for variables.

columbarius: Put everything that is constant into the struct and added the variable bits afterwards. I think that's the best we can do, except having one array for all paths and one for all folders and a strange function which maps the order of the config files to the place in each array. But that will probably not make things clearer.

Also change it to calculate the number of tried config files at runtime.

emersion: Maybe try dropping the `const` qualifier for the array

columbarius: You mean the static qualifier? That solved the issue. Didn't know static has some form of constant global behaviour.

emersion (approved):
LGTM, thanks!

columbarius:
Thanks for merging!

PR #6104, opened 2021-03-16
idle_inhibit: Always listen for view_unmap

kennylevinsen:
For wlr_idle_inhibit, we'd listen to inhibitor destroy, whereas for
manual idle inhibition, we'd listen for view unmap. If a view associated
with a wlr_idle_inhibit inhibitor was destroyed, we could possibly end
up with a dangling view pointer.

Always listen to view unmap to ensure we clean up properly.

### DIFF
--- a/sway/desktop/idle_inhibit_v1.c
+++ b/sway/desktop/idle_inhibit_v1.c
@@ -11,14 +11,21 @@
 static void destroy_inhibitor(struct sway_idle_inhibitor_v1 *inhibitor) {
 	wl_list_remove(&inhibitor->link);
 	wl_list_remove(&inhibitor->destroy.link);
+	wl_list_remove(&inhibitor->view_unmap.link);
 	sway_idle_inhibit_v1_check_active(inhibitor->manager);
 	free(inhibitor);
+	sway_log(SWAY_DEBUG, "Sway idle inhibitor destroyed");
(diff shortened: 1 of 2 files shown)


sway/desktop/idle_inhibit_v1.c
```diff
@@ -58,8 +65,9 @@ void sway_idle_inhibit_v1_user_inhibitor_register(struct sway_view *view,
 	inhibitor->view = view;
 	wl_list_insert(&inhibitor->manager->inhibitors, &inhibitor->link);
 
-	inhibitor->destroy.notify = handle_destroy;
-	wl_signal_add(&view->events.unmap, &inhibitor->destroy);
```
emersion, line 62: Hm, why isn't `unmap` emitted before `destroy`?

kennylevinsen: Unmap is likely fine - hadn't dug into the wlroots side yet. However, the section for idle_inhibit only listened on inhibitor destroy, so that's the primary change.

kennylevinsen:
Awaiting feedback from someone who can reproduce the issue...

kennylevinsen:
https://github.com/swaywm/sway/pull/6121 is a better fix in my opinion, so let's go with that.

Issue #4939, opened 2020-01-23
[Feature request] swaynag: option to display on top


PeterNerlich:
I think swaynag could use some argument that makes it display on top, showing regardless whether a window is fullscreen etc. Currently, when I have something fullscreen, I don't notice swaynag appearing.

    $ swaymsg -t get_version
    sway version 1.2

distek:
To add on to this request; This would be a use case:
https://github.com/m00qek/swaynag-battery

robsmith11:
Is there any workaround that can be used now?  I also use swaynag for low battery notifications, but have missed a few when using a fullscreen window, resulting in my laptop shutting off in the middle of my work.

I run this script in the background:
```
#!/bin/bash

while true; do
    sleep 61
    ct=`cat /sys/class/power_supply/BATT/charge_now`
    c0=`cat /sys/class/power_supply/BATT/charge_full`
    (( ct*1000 / c0 < 70 )) && swaynag -m 'low battery'
done
```

PeterNerlich:
I have not found any yet.

fluix-dev:
I believe the following should work, feel free to let me know if it does.

```diff
diff --git a/swaynag/swaynag.c b/swaynag/swaynag.c
index d9bec3687..c8d97fb09 100644
--- a/swaynag/swaynag.c
+++ b/swaynag/swaynag.c
@@ -474,7 +474,7 @@ void swaynag_setup(struct swaynag *swaynag) {
 	swaynag->layer_surface = zwlr_layer_shell_v1_get_layer_surface(
 			swaynag->layer_shell, swaynag->surface,
 			swaynag->output ? swaynag->output->wl_output : NULL,
-			ZWLR_LAYER_SHELL_V1_LAYER_TOP, "swaynag");
+			ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY, "swaynag");
 	assert(swaynag->layer_surface);
 	zwlr_layer_surface_v1_add_listener(swaynag->layer_surface,
 			&layer_surface_listener, swaynag);
```
In the meantime, I'll work on an actual solution of using `--layer background|bottom|top|overlay` to select the layer shell layer.

robsmith11:
> I believe the following should work, feel free to let me know if it does.

Thanks!  I finally got around to patching sway after losing my work one too many times after missing a low battery notification while fullscreen.  It now works perfectly.

GrabbenD:
[...]

PR #6105, opened 2021-03-17
swaynag: adds option to set wayland shell layer

Jamedjo:
Uses --layer/-y set to overlay|top|bottom|background

Closes https://github.com/swaywm/sway/issues/4939

### DIFF
--- a/include/swaynag/types.h
+++ b/include/swaynag/types.h
@@ -7,6 +7,7 @@ struct swaynag_type {
 	char *font;
 	char *output;
 	uint32_t anchors;
+	int32_t layer; // enum zwlr_layer_shell_v1_layer or -1 if unset
 
 	// Colors
 	uint32_t button_text;
--- a/swaynag/swaynag.c
+++ b/swaynag/swaynag.c
@@ -476,7 +476,8 @@ void swaynag_setup(struct swaynag *swaynag) {
 	swaynag->layer_surface = zwlr_layer_shell_v1_get_layer_surface(
 			swaynag->layer_shell, swaynag->surface,
 			swaynag->output ? swaynag->output->wl_output : NULL,
-			ZWLR_LAYER_SHELL_V1_LAYER_TOP, "swaynag");
+			swaynag->type->layer,
+			"swaynag");
 	assert(swaynag->layer_surface);
 	zwlr_layer_surface_v1_add_listener(swaynag->layer_surface,
 			&layer_surface_listener, swaynag);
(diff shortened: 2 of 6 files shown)


fluix-dev (approved):

swaynag/types.c
```diff
@@ -100,6 +101,10 @@ void swaynag_type_merge(struct swaynag_type *dest, struct swaynag_type *src) {
 		dest->anchors = src->anchors;
 	}
 
+	if (src->layer > 0) {
```
emersion, line 104: `ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND == 0` so we can't use this as a guard value.

swaynag/swaynag.c
```diff
@@ -476,7 +476,8 @@ void swaynag_setup(struct swaynag *swaynag) {
 	swaynag->layer_surface = zwlr_layer_shell_v1_get_layer_surface(
 			swaynag->layer_shell, swaynag->surface,
 			swaynag->output ? swaynag->output->wl_output : NULL,
-			ZWLR_LAYER_SHELL_V1_LAYER_TOP, "swaynag");
+			swaynag->type->layer ? swaynag->type->layer : ZWLR_LAYER_SHELL_V1_LAYER_TOP,
```
emersion, line 479: Ditto

swaynag/types.c
Jamedjo: > `ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND == 0` so we can't use this as a guard value.

Thanks for spotting this, I've updated it to use a `int32_t` with an initial value of `-1`.

We could alternatively remove support for the `background` setting, as I'm unaware of the use case. From my testing the nag bar appears behind the desktop background, but if a smaller bg is centred on the page it does show and looks similar to using a value of `bottom`.

emersion:
Hm, CI is failing for some reason -- maybe rebase against master to see if that fixes it?

include/swaynag/types.h
```diff
@@ -7,6 +7,7 @@ struct swaynag_type {
 	char *font;
 	char *output;
 	uint32_t anchors;
+	int32_t layer;
```
emersion, line 10: Nit: ideally add a small `// enum zwlr_layer_shell_v1_layer or -1 if unset` comment to make it clear what the values can be.

emersion (approved):
LGTM, thanks!

Issue #6109, opened 2021-03-18
swaybar: Implement further touch / tablet (pen) events


Zocker1999NET:
I use sway / swaybar on a Thinkpad X1 Tablet Gen3 (supporting pen inputs). Sway can correctly accept touch & pen inputs and forward those to all applications appropriate. But I also use swaybar to implement a bar myself, which works quite well, but due to swaybar not forwarding events like swiping with a finger (will be forwarded as simple click), holding down with a finger (also simple click) and so on I cannot use those events for further actions. Also with the pen it is not possible to send something equivalent to a right click to swaybar (other applications seem to detect a button on my pen as "right click", but not swaybar).

It would be great if we could receive some of those events from swaybar to use in our status scripts.

My propose, which should hopefully easy to implement for already contributors:
- long touch press -> right click
- doubled touch press -> right click
- button press on pen -> right click

This would be a great step forward for more touch/tablet support IMHO.

kennylevinsen:
Try this: https://github.com/swaywm/sway/pull/6111

PR #6111, opened 2021-03-18
swaybar: Use position from wl_pointer.enter

kennylevinsen:
Only wl_pointer.motion was used to update pointer position, which would
cause issues if the pointer was not moved prior to wl_pointer.button.

This also fixes touch input through wl_pointer emulation, which fires
wl_pointer.button immediately after wl_pointer.enter.

Copied from a similar fix made to swaynag.

Closes: https://github.com/swaywm/sway/issues/6109


emersion (approved):
LGTM

PR #6113, opened 2021-03-19
layer-shell: don't arrange layers on each surface commit

emersion:
Only arrange layer surfaces if a surface has updated its layer-shell
configuration.

References: https://github.com/swaywm/wlroots/pull/2794


vyivel:
Closing as outdated.

PR #6119, opened 2021-03-19
Add sway, swaymsg and swaynag completions for pwsh

isti115:
Some basic PowerShell Core tab completion support for sway, swaymsg and swaynag


emersion:
Would be nice to get this reviewed by another PowerShell user.

These completion files are just added to the repo, not installed. Is there a place where these should be installed?

Isti115:
I personally don't know of any specific directory where these should be placed, they can be sourced from the PowerShell profile script.

That doesn't mean that there isn't a standardized location though, as I'm not an expert in PowerShell by any means.

Issue #6103, opened 2021-03-16
idle_inhibit inhibitor held while client crashes leads to use after free


kennylevinsen:
The following ASan report was given by `kenny` in IRC:
```
=================================================================
==7289==ERROR: AddressSanitizer: heap-use-after-free on address 0x61600011eb90 at pc 0x5556eecd4f8d bp 0x7ffdcd550f10 sp 0x7ffdcd550f00
READ of size 8 at 0x61600011eb90 thread T0
    #0 0x5556eecd4f8c in sway_idle_inhibit_v1_is_active ../sway/sway/desktop/idle_inhibit_v1.c:109
    #1 0x5556eecd51b8 in sway_idle_inhibit_v1_check_active ../sway/sway/desktop/idle_inhibit_v1.c:138
    #2 0x5556eecd53b3 in destroy_inhibitor ../sway/sway/desktop/idle_inhibit_v1.c:14
    #3 0x5556eecd5403 in handle_destroy ../sway/sway/desktop/idle_inhibit_v1.c:22
    #4 0x7fa71ed7e466 in wlr_signal_emit_safe ../util/signal.c:29
    #5 0x7fa71ed45ebc in idle_inhibitor_v1_destroy ../types/wlr_idle_inhibit_v1.c:36
    #6 0x7fa71ed45ebc in idle_inhibitor_v1_destroy ../types/wlr_idle_inhibit_v1.c:31
    #7 0x7fa71ed45ebc in idle_inhibitor_handle_surface_destroy ../types/wlr_idle_inhibit_v1.c:55
    #8 0x7fa71ed7e466 in wlr_signal_emit_safe ../util/signal.c:29
    #9 0x7fa71ed6d726 in surface_handle_resource_destroy ../types/wlr_surface.c:644
    #10 0x7fa71ee417ff  (/usr/lib/libwayland-server.so.0+0x77ff)
    #11 0x7fa71ee47b40  (/usr/lib/libwayland-server.so.0+0xdb40)
    #12 0x7fa71ee4804f  (/usr/lib/libwayland-server.so.0+0xe04f)
    #13 0x7fa71ee422bd in wl_client_destroy (/usr/lib/libwayland-server.so.0+0x82bd)
    #14 0x7fa71ee423a8  (/usr/lib/libwayland-server.so.0+0x83a8)
    #15 0x7fa71ee45079 in wl_event_loop_dispatch (/usr/lib/libwayland-server.so.0+0xb079)
[...]

Issue #6091, opened 2021-03-08
segmentation fault when idle is inhibited while sleeping?


VRichardJP:
- sway version 1.5.1
- log: https://gist.github.com/VRichardJP/ed123c27395119469172096e76cb57d6
- generated with command:
```
$ sway -c default_config -d 2> ~/sway_segfault.log
```
- config: https://gist.github.com/VRichardJP/6e35c4858ad42fcb6dbcdc593e98ffa8
this is the default sway config with the following changes:
```
### Idle configuration
#
# Example configuration:
#
exec swayidle -w \
         timeout 300 'swaylock -f -c 000000' \
         timeout 600 'swaymsg "output * dpms off"' resume 'swaymsg "output * dpms on"' \
         before-sleep 'swaylock -f -c 000000'
#
# This will lock your screen after 300 seconds of inactivity, then turn off
# your displays after another 300 seconds, and turn your screens back on when
# resumed. It will also lock your screen before your computer goes to sleep.
#
# inhibit_idle when any application is running fullscreen
#
for_window [class="^.*"] inhibit_idle fullscreen
for_window [app_id="^.*"] inhibit_idle fullscreen
#
# clamshell mode support (lid closed)
#
set $laptop eDP-1
bindswitch --reload --locked lid:on output $laptop disable
bindswitch --reload --locked lid:off output $laptop enable
```

This is a segfault I encountered while trying to produce a minimal example for https://github.com/swaywm/sway/issues/6072. The `inhibit_idle fullscreen` lines are not necessarily smart, but they create some weird interaction that may be worth investigating.

- What I do to reproduce:
1. start sway
2. wait few seconds
3. close the laptop lid
4. wait few seconds
5. open the laptop lid
6. sway has crashed

- What I understand of the crash:
1. When I close the lid: systemd suspends the laptop, swayidle execute swaylock, the bindswitch on lid:on disable the output and reload the configuration. On configuration reload idle is inhibited because swaylock is running fullscreen.
2. When I open the lid the laptop goes out of sleep, sway crashes before lid:off event is triggered

Issue #5680, opened 2020-09-14
heap-use-after-free sway_idle_inhibit_v1_is_active


KenMacD:
Sway from src with commit: r6478

```
==1140==ERROR: AddressSanitizer: heap-use-after-free on address 0x6160008dd290 at pc 0x55b538b8ce1c bp 0x7ffc7bb05da0 sp 0x7ffc7bb05d90
READ of size 8 at 0x6160008dd290 thread T0
    #0 0x55b538b8ce1b in sway_idle_inhibit_v1_is_active ../sway/sway/desktop/idle_inhibit_v1.c:109
    #1 0x55b538b8d047 in sway_idle_inhibit_v1_check_active ../sway/sway/desktop/idle_inhibit_v1.c:138
    #2 0x55b538ba356b in transaction_progress_queue ../sway/sway/desktop/transaction.c:354
    #3 0x55b538ba4e82 in transaction_commit_dirty ../sway/sway/desktop/transaction.c:542
    #4 0x55b538c2066a in view_unmap ../sway/sway/tree/view.c:846
    #5 0x55b538ba5629 in handle_unmap ../sway/sway/desktop/xdg_shell.c:414
    #6 0x7f9c8dbe5c76 in wlr_signal_emit_safe ../util/signal.c:29
    #7 0x7f9c8db9daff in unmap_xdg_surface ../types/xdg_shell/wlr_xdg_surface.c:39
    #8 0x7f9c8db9e334 in reset_xdg_surface ../types/xdg_shell/wlr_xdg_surface.c:474
    #9 0x7f9c8db9e3c2 in destroy_xdg_surface ../types/xdg_shell/wlr_xdg_surface.c:513
    #10 0x7f9c8db9b7f8 in xdg_client_handle_resource_destroy ../types/xdg_shell/wlr_xdg_shell.c:71
    #11 0x7f9c8dca3e8f  (/usr/lib/libwayland-server.so.0+0x8e8f)
    #12 0x7f9c8dca8850  (/usr/lib/libwayland-server.so.0+0xd850)
    #13 0x7f9c8dca8d4f  (/usr/lib/libwayland-server.so.0+0xdd4f)
    #14 0x7f9c8dca401f in wl_client_destroy (/usr/lib/libwayland-server.so.0+0x901f)
    #15 0x7f9c8dca4108  (/usr/lib/libwayland-server.so.0+0x9108)
[...]

Issue #5325, opened 2020-05-10
Heap use-after-free in sway_idle_inhibit_v1_check_active


Xyene:
The following sway crash happened while clicking a tab in Firefox with my tablet pen, sway/wlroots master.

```
38:58:31.491 [sway/desktop/idle_inhibit_v1.c:29] New sway idle inhibitor                                                
38:58:31.491 [DEBUG] [types/wlr_idle.c:186] Disabling idle timers for all seats                                         
38:58:31.491 [sway/desktop/idle_inhibit_v1.c:29] New sway idle inhibitor   
...
38:58:35.471 [sway/desktop/transaction.c:280] Applying transaction 0x6040006a0690
38:58:35.471 [sway/desktop/idle_inhibit_v1.c:21] Sway idle inhibitor destroyed
=================================================================
==2335==ERROR: AddressSanitizer: heap-use-after-free on address 0x615000000a90 at pc 0x55865b7a232d bp 0x7ffd0491a400 sp 0x7ffd0491a3f8
READ of size 8 at 0x615000000a90 thread T0
    #0 0x55865b7a232c in check_active ../sway/desktop/idle_inhibit_v1.c:96
    #1 0x55865b7a264c in sway_idle_inhibit_v1_check_active ../sway/desktop/idle_inhibit_v1.c:125
    #2 0x55865b7a1cd7 in destroy_inhibitor ../sway/desktop/idle_inhibit_v1.c:14
    #3 0x55865b7a1d30 in handle_destroy ../sway/desktop/idle_inhibit_v1.c:22
    #4 0x7fd9a3fc6f1e in wlr_signal_emit_safe ../util/signal.c:29
    #5 0x7fd9a3f8547f in idle_inhibitor_v1_destroy ../types/wlr_idle_inhibit_v1.c:36
    #6 0x7fd9a3f8553d in idle_inhibitor_handle_surface_destroy ../types/wlr_idle_inhibit_v1.c:55
    #7 0x7fd9a3fc6f1e in wlr_signal_emit_safe ../util/signal.c:29
    #8 0x7fd9a3fb6c8f in surface_handle_resource_destroy ../types/wlr_surface.c:571
    #9 0x7fd9a408e3ce  (/lib/x86_64-linux-gnu/libwayland-server.so.0+0x93ce)
    #10 0x7fd9a4092c10  (/lib/x86_64-linux-gnu/libwayland-server.so.0+0xdc10)
[...]

PR #6121, opened 2021-03-20
idle_inhibit: Store wlr inhibitor instead of view

kennylevinsen:
Should fix a dangling sway_view pointer in between view unmap and inhibitor destroy, which could cause issues if inhibition is checked in between.

As a bonus, we should now support idle inhibitors persisting across surface remap.

I still haven't managed to reproduce the original issue, so it would be nice if we could have someone experiencing problems test it.

Closes: https://github.com/swaywm/sway/issues/5325
Closes: https://github.com/swaywm/sway/issues/5680
Closes: https://github.com/swaywm/sway/issues/6091
Closes: https://github.com/swaywm/sway/issues/6103
Supersedes https://github.com/swaywm/sway/pull/6104

### DIFF
--- a/include/sway/desktop/idle_inhibit_v1.h
+++ b/include/sway/desktop/idle_inhibit_v1.h
@@ -22,6 +22,7 @@ struct sway_idle_inhibit_manager_v1 {
 
 struct sway_idle_inhibitor_v1 {
 	struct sway_idle_inhibit_manager_v1 *manager;
+	struct wlr_idle_inhibitor_v1 *wlr_inhibitor;
 	struct sway_view *view;
 	enum sway_idle_inhibit_mode mode;
 
--- a/sway/desktop/idle_inhibit_v1.c
+++ b/sway/desktop/idle_inhibit_v1.c
@@ -36,7 +36,7 @@ void handle_idle_inhibitor_v1(struct wl_listener *listener, void *data) {
 
 	inhibitor->manager = manager;
 	inhibitor->mode = INHIBIT_IDLE_APPLICATION;
-	inhibitor->view = view_from_wlr_surface(wlr_inhibitor->surface);
+	inhibitor->wlr_inhibitor = wlr_inhibitor;
 	wl_list_insert(&manager->inhibitors, &inhibitor->link);
 
 	inhibitor->destroy.notify = handle_destroy;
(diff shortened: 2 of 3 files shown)


sway/desktop/idle_inhibit_v1.c
```diff
@@ -104,10 +104,10 @@ void sway_idle_inhibit_v1_user_inhibitor_destroy(
 
 bool sway_idle_inhibit_v1_is_active(struct sway_idle_inhibitor_v1 *inhibitor) {
 	switch (inhibitor->mode) {
-	case INHIBIT_IDLE_APPLICATION:
+	case INHIBIT_IDLE_APPLICATION:;
 		// If there is no view associated with the inhibitor, assume visible
-		return !inhibitor->view || !inhibitor->view->container ||
-			view_is_visible(inhibitor->view);
+		struct sway_view *view = view_from_wlr_surface(inhibitor->wlr_inhibitor->surface);
```
emersion, line 109: Can `inhibitor->wlr_inhibitor->surface` be NULL here?

kennylevinsen: The idle inhibitor itself is destroyed on wlr_surface destroy, and the surface is not nulled anywhere, so unless inhibitors can be created with NULL surfaces somehow, the inhibitor surface pointer should be valid up until and including the inhibitor destroy signal.

emersion: Makes sense!

include/sway/desktop/idle_inhibit_v1.h
```diff
@@ -22,6 +22,7 @@ struct sway_idle_inhibit_manager_v1 {
 
 struct sway_idle_inhibitor_v1 {
 	struct sway_idle_inhibit_manager_v1 *manager;
+	struct wlr_idle_inhibitor_v1 *wlr_inhibitor;
 	struct sway_view *view;
```
emersion, line 26: Is this field still used?

kennylevinsen: Yes, it is used for manual/"user" inhibitors made with the `inhibit_idle` command. It is only accessed if `inhibitor->mode` is not `INHIBIT_IDLE_APPLICATION`.

wisp3rwind:
Appears to fix https://github.com/swaywm/sway/issues/5680 for me: With the updated reproducer, 6059c744f8b24ef3321eb7896991112366896e26 reliably crashes, but this PR does not.

emersion (approved):
Thanks!

Issue #5931, opened 2021-01-05
exec quoting different from i3 (double quoted executables not working)


mreppen:
- **Description:**
  - When running `exec` on an executable with a space in the path, i3 works with double-quoting, but sway does not. But Sway does work with `\ ` for quoting the spaces.
  - In i3, the following works: (i3-msg) `exec "\"my exectuable\""` 
  - In Sway, (swaymsg) `exec "\"my exectuable\""` fails

Is this intentional?

Because the spec for .desktop files specify quoting, it breaks using `i3-dmenu-desktop` in Sway for such .desktop files

emersion:
`swaymsg exec "\"termite\""` works fine, and `swaymsg exec "\"termite -t test\""` works fine too. So I guess we're incorrectly stripping too many quotes?

mreppen:
Indeed. I would expect your second example to fail.

RPigott:
@emersion those are both as expected. exec strips 1 layer of outer quotes, as it does in i3. https://i3wm.org/docs/userguide.html#exec_quoting

Similarly, `i3-msg exec "\"test file\""` in i3 and `swaymsg exec "\"test file\""` in sway have the same behavior for me. They both execute an executable with name "test" and argument "file".

If you want to run an executable "test file", use: `swaymsg exec '"\"test file\""'`. The `''` are stripped by the shell, the `""` are stripped by exec, and the inner `\"test file\"` is passed to the exec'd shell as `"test file"` so our command `test file` runs with no arguments.

RPigott:
@mreppen could you give an example desktop file where i3 and sway differ? A debug log could help identify how sway interprets the commands.

mreppen:
@RPigott  Sure. But you don't need a desktop file to see that they are different:

`~/test exec.sh`:
```
#!/bin/bash
notify-send Works
```

Works: `i3-msg 'exec "\"$HOME/test exec.sh\""'`
Does not work: `swaymsg 'exec "\"$HOME/test exec.sh\""'`


.desktop example:

`~/.local/share/applications/myapp.desktop`:
```
[Desktop Entry]
Name=My App
Exec="$HOME/test exec.sh"
Terminal=false
Type=Application
```

Running `i3-dmenu-desktop` and selecting "My App" in i3 works, but in Sway not.


I actually patched `i3-dmenu-desktop` to do more proper escaping before I realized that it already works in i3

RPigott:
Well at least an example wasn't hard to find.

[...]

PR #6123, opened 2021-03-20
commands: unescape '\' from IPC commands as well

### LINKED ISSUE #5652: i3bar: cannot click(move) on workspaces with \ in the name
Bug related to pango markup 

Steps to reproduce the issue:

- create a window in workspace 6
- `i3-msg 'rename workspace 6 to 6\'`
+ you will see the old workspace `6` and you will see a new empty workspace `6\`
+ then start clicking on the workspace named `6\` several times, it appends \\\\\\\\ exponentially

swaybar v1.5

fenveireth:
Fixes #5652.
Was not consistent with handling of commands read from config file

### DIFF
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -255,18 +255,10 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 			continue;
 		}
 		sway_log(SWAY_INFO, "Handling command '%s'", cmd);
+
 		//TODO better handling of argv
 		int argc;
 		char **argv = split_args(cmd, &argc);
-		if (strcmp(argv[0], "exec") != 0 &&
-				strcmp(argv[0], "exec_always") != 0 &&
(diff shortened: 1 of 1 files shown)


emersion:
Hm, shouldn't we avoid doing the escaping in the first place?

Fenveireth:
That was my first reflex, but turns out it makes it overall easier to create workspaces you are locked out of.
Invoked edge case for the escaping is using a workspace called "1" (double-quotes part of the display name). That's arguably just asking for trouble, but once it's in there, we need some way to make the name precisely round-trip

As of now, sway IPC-side lets you do that by putting problematic characters individually between single-quotes, if the first character in the argument is single-quoted.
This is not compatible with i3, which will not interpret the single-quoting, and will interpret '\' shell-style

mreppen:
Possibly relevant (also issue with quoting \\):
https://github.com/swaywm/sway/issues/5931

Issue #6115, opened 2021-03-19
swaybar can fail to parse IPC messages if the JSON is nested too deep


oranenj:
I'm running sway 1.6rc1

I noticed that sometimes, workspace events may produce messages that swaybar (at least) fails to parse, with the following message in the logs:
```
[swaybar/ipc.c:552] failed to parse payload as json
```

After debugging a bit, this turns out to be due to swaybar using `json_tokener_parse` to parse the input string. This function has a maximum nesting depth of 32. I verified that this is the issue by using json_tokener_parse_verbose instead and printing out the error, which resulted in:
```
[swaybar/ipc.c:553] bad data, error was: nesting too deep
```
Since json-c 0.11 there's a new function that allows setting the maximum depth:
https://json-c.github.io/json-c/json-c-0.11/doc/html/json__tokener_8h.html#a8c0696656dba7dde887a71e5c7686057

It seems that in the IPC messages, the "nodes" key can get nested arbitrarily deep. I don't know if it's normal for it to become this deeply nested, or if there's some other bug that's triggering this.

Looking at the top-level "representation" key, it does seem a bit silly:
```
"representation": "H[H[H[H[V[V[H[V[V[V[V[H[H[V[H[V[gnome-terminal-server] V[V[V[H[gnome-terminal-server] H[gnome-terminal-server]]]]]]]]]]]]]]]]]]]"
```

oranenj:
I think I may have found how to make sway "leak" containers. Let me see if I can reproduce the exact set of actions...

oranenj:
yeah, I managed. That's a separate issue though, so I'll open a new bug

PR #6124, opened 2021-03-21
Avoid redundant containers more aggressively

### LINKED ISSUE #6116: Specific sequences may cause sway to "leak" containers, potentially leading to unnecessary deep nesting in the representation hierarchy
Using sway 1.6rc1.

Related to #6115
I used my terminal to reproduce because it's easy to open new ones, but I suspect this will happen with any window.
```
$ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": null
}
$ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": "H[gnome-terminal-server gnome-terminal-server]"
}
# here we have the lone horizontal container that is not needed.
$ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": "V[H[gnome-terminal-server] gnome-terminal-server]"
}
$ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": "V[H[gnome-terminal-server] gnome-terminal-server gnome-terminal-server]"
}
$ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": "H[V[H[gnome-terminal-server] gnome-terminal-server] gnome-terminal-server]"
}
$ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": "H[V[H[gnome-terminal-server]] gnome-terminal-server]"
}
# and these are definitely even less necessary:
$ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": "H[V[H[gnome-terminal-server]]]"
}
$ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": "H[]"
}
```
And the steps to reproduce:
0. start with a completely fresh workspace
1. open two windows side-by-side
2. move the second window from horizontal split to vertical (shift-mod-h)
3. Open another window
4. Now move the bottom one to the right (shift-mod-l)
5. close the left bottom one. You should be left with two windows in a horizontal split, but the representation says `H[V[H[gnome-terminal-server]] gnome-terminal-server]`
6. If you close all windows, there will be a container "left over"

Apparently some sequences of opening and closing windows can result in arbitrarily deep nesting, causing #6115

oranenj:
Hopefully fixes #6116. With the previous logic, it was possible to accumulate redundant containers, potentiall causing arbitrarily deep nesting.

I'm not a 100% certain that this patch doesn't change behaviour, but I'm reasonably sure that if it does, it's unnoticeable.

I tested this locally and was unable to make things crash and burn even when mashing arbitrary window movement commands.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1705,10 +1705,16 @@ static bool is_parallel(enum sway_container_layout first,
 static bool container_is_squashable(struct sway_container *con,
 		struct sway_container *child) {
 	enum sway_container_layout gp_layout = container_parent_layout(con);
-	return (con->pending.layout == L_HORIZ || con->pending.layout == L_VERT) &&
-		(child->pending.layout == L_HORIZ || child->pending.layout == L_VERT) &&
+	bool both_tiling = (con->pending.layout == L_HORIZ || con->pending.layout == L_VERT) &&
+		(child->pending.layout == L_HORIZ || child->pending.layout == L_VERT);
+
+	// this function is only called when con has one child. Therefore
+	// it is redundant if its layout is the same as the child's
+	return  (con->pending.layout == child->pending.layout) ||
+		// We can squash eg. H[a, b, V[H[c, d]]] to H[a, b, c, d]
+		(both_tiling &&
 		!is_parallel(con->pending.layout, child->pending.layout) &&
-		is_parallel(gp_layout, child->pending.layout);
+		is_parallel(gp_layout, child->pending.layout));
 }
 
 static void container_squash_children(struct sway_container *con) {
@@ -1730,7 +1736,7 @@ int container_squash(struct sway_container *con) {
 	int idx = container_sibling_index(con);
 	int change = 0;
 	if (container_is_squashable(con, child)) {
-		// con and child are a redundant H/V pair. Destroy them.
+		// con and child are a redundant pair. Destroy them.
 		while (child->pending.children->length) {
 			struct sway_container *current = child->pending.children->items[0];
 			container_detach(current);


oranenj:
With this patch, it seems you can still end up with *some* redundancy, but it should avoid silly layouts like H[H[H[...]]] from accumulating, and the nesting will clear itself up before it can cause problems.

RPigott (commented):
This PR mutates layouts in a way that i3 does not, so we shouldn't merge this.

To be clear, addressing the initial comment of #6116:

> \# here we have the lone horizontal container that is not needed.
> "r": "V[H[gnome-terminal-server] gnome-terminal-server]"

Not a bug.

> \# and these are definitely even less necessary:
>  "r": "H[V[H[gnome-terminal-server]]]"

Not a bug.

However silly these layouts may be, they are not disallowed. We shouldn't deviate from i3 here.

> $ swaymsg -t get_workspaces --raw | jq '.[] | {num: .num, r: .representation} | select(.num==11)'
{
  "num": 11,
  "r": "H[]"
}

This is a bug, and still reproducible for me with this PR. The workspace representation shouldn't be "H[]". I would guess this is a missing/misplaced call to arrange_workspace somewhere.

oranenj:
@RPigott I admit I have no idea how i3 behaves here (I don't care about i3 compatibility, though I understand that others do), so the proposed logic may well be wrong. I've seen layouts to the effect of H[H[H[H[H[H[H[firefox]]]]]]] or something though, which is definitely not correct, and once they stack deep enough, sway starts breaking in more noticeable ways.

Xyene:
> I've seen layouts to the effect of H[H[H[H[H[H[H[firefox]]]]]]] or something though, which is definitely not correct

Depending on how these layouts are produced, they are valid under i3. An easy way to produce one is to create `H[A H[A' H[A'' B]]]` and then close {A'', A', A} to get `H[H[H[B]]]`. A single move of `B` at this point would collapse the entire chain into `H[B]`, but the collapsing wouldn't happen before then.

RPigott:
FWIW, i3's approach to tree simplification seems ad hoc to me as well. I agree both that there is no practical purpose for deeply nested parallel singleton containers and that they may be too "easy" to create accidentally in certain usage patterns. As long as i3 has them though, sway should as well. And sway should be made to handle them robustly. So it sounds like #6115 will need to be addressed in swaybar.

I'm going to close this PR, but feel free open a new one if you still want to address the "H[]" bugged case.

PR #6127, opened 2021-03-22
swaybar: fail gracefully on tokener creation fail

### LINKED ISSUE #6126: swaybar crashes on json_tokener_parse_ex
Switching the workspace after starting sway always kills the swaybar. Status command is i3bar

https://github.com/swaywm/sway/pull/6125 has introduced this. Reverting back before this commit fixes the issue.

- **Sway Version:**
master at 47f1f9b63ea04ee5d520f36d9a0ef5de04921816

- **Stack Trace:**

```
#0  0x00007f915eac514b json_tokener_parse_ex (libjson-c.so.5 + 0x814b)
#1  0x000055628f5bbceb handle_ipc_readable (swaybar + 0xbceb)
#2  0x000055628f5b7b92 ipc_in (swaybar + 0x7b92)
#3  0x000055628f5c4d4c loop_poll (swaybar + 0x14d4c)
#4  0x000055628f5b8696 bar_run (swaybar + 0x8696)
#5  0x000055628f5b7197 main (swaybar + 0x7197)
#6  0x00007f915e7e1b25 __libc_start_main (libc.so.6 + 0x27b25)
#7  0x000055628f5b72ae _start (swaybar + 0x72ae)
```

Xyene:
This commit adds missing error-handling to the creation of the tokener
instance. The stack depth parameter is used to initialize an array that
json-c prefaults ahead of time, causing INT_MAX to result in out of
memory errors.

Also drop the depth to 256 to prevent this OOM.

Though this fix is not very satisfactory -- json-c could be made to
not prefault -- it should do for now. At the very least, swaybar will
not crash.

Fixes #6126.

### DIFF
--- a/swaybar/ipc.c
+++ b/swaybar/ipc.c
@@ -547,7 +547,16 @@ bool handle_ipc_readable(struct swaybar *bar) {
 		return false;
 	}
 
-	json_tokener *tok = json_tokener_new_ex(INT_MAX);
+	// The default depth of 32 is too small to represent some nested layouts, but
+	// we can't pass INT_MAX here because json-c (as of this writing) prefaults
+	// all the memory for its stack.
+	json_tokener *tok = json_tokener_new_ex(256);
+	if (!tok) {
+		sway_log_errno(SWAY_ERROR, "failed to create tokener");
+		free_ipc_response(resp);
+		return false;
+	}
+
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks for the fix!

PR #6133, opened 2021-03-23
contrib/autoname-workspaces.py: fix type error when class_name none

mordax777:
Same as in [this line](https://github.com/swaywm/sway/blob/master/contrib/autoname-workspaces.py#L26), fix the TypeError when  class_name is of type `NoneType`.

### DIFF
--- a/contrib/autoname-workspaces.py
+++ b/contrib/autoname-workspaces.py
@@ -31,7 +31,7 @@ def icon_for_window(window):
     else:
         # xwayland support
         class_name = window.window_class
-        if len(class_name) > 0:
+        if class_name is not None and len(class_name) > 0:
             class_name = class_name.lower()
             if class_name in WINDOW_ICONS:
                 return WINDOW_ICONS[class_name]
--- a/contrib/autoname-workspaces.py
+++ b/contrib/autoname-workspaces.py
@@ -22,24 +22,17 @@ DEFAULT_ICON = "󰀏"
 
 
 def icon_for_window(window):
-    app_id = window.app_id
-    if app_id is not None and len(app_id) > 0:
-        app_id = app_id.lower()
-        if app_id in WINDOW_ICONS:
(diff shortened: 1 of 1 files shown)


contrib/autoname-workspaces.py
```diff
@@ -31,7 +31,7 @@ def icon_for_window(window):
     else:
         # xwayland support
         class_name = window.window_class
-        if len(class_name) > 0:
+        if class_name is not None and len(class_name) > 0:
```
kennylevinsen, line 34: Maybe avoid the nested else/if and duplication doing something like:

```python
name = None
if window.app_id is not None and len(window.app_id) > 0:
    name = window.app_id.lower()
elif window.window_class is not None and len(window.window_class) > 0:
    name =  window.window_class.lower()

if name in WINDOW_ICONS:
    return WINDOW_ICONS[name]

logging.info("No icon available for window with name: %s" % str(name))
return DEFAULT_ICON
```

mordax777: Thanks! I tested the refactored function and it worked. :slightly_smiling_face: 

I committed the refactored function to this PR.

kennylevinsen:
Build failing due to missing some changes from master. Please rebase. :)

mordax777:
> Build failing due to missing some changes from master. Please rebase. :)

done ^^

kennylevinsen (approved):
LGTM, thanks!

tmpm697:
This script seems to does not respect current floating property of windows, my windows seem to always `tabbed` that inherit from parent or smth.

fluix-dev:
@tmpm697 is this issue somehow related to this PR? If not, open a new issue (or discuss in other channels).

tmpm697:
seem like to be, as I use inactive-transparency-window without issue but this script change all my floating window to tabbed ones even though I've set `for_window [workspace="abc"] floating enable` in the current workspace that runs this script.

PR #6135, opened 2021-03-23
view: subsurface NULL check in view_from_wlr_surface

kennylevinsen:
Necessary NULL checks had been added to xdg_shell and xwayland surfaces,
but subsurfaces had been missed.


emersion (approved):
Thanks!

Issue #6118, opened 2021-03-19
Pango markup not rendered in swaybar as i3bar does


danielotero:
### Description

Currently, `swaybar` does not render Pango markup as i3bar does. The i3 behavior seems reasonable: the markup will be rendered if the bar font is prefixed with the "pango:" string.

These should be easily reproducible with the following config:
```
bindsym Mod4+1 workspace <b>O</b>ne
bar     font  "pango:monospace 10"
```
or this one:
```
bindsym Mod4+1 workspace <b>O</b>ne
font    "pango:monospace 10"
bar     position bottom
```

Which both results in:

![image](https://user-images.githubusercontent.com/9847857/111844161-c3a14c80-8902-11eb-8090-0f6faa8119e6.png)

### Details

This is because the [`pango_markup`](https://github.com/swaywm/sway/blob/989123a2a5933367e5c7c39c3793f4814e026bf1/include/swaybar/config.h#L29) `swaybar_config` member is not set with the [`font` command](https://github.com/swaywm/sway/blob/5fb5984e941cade797acac8f80b471c2dba58de8/sway/commands/bar/font.c#L7-L17) as with the [same command](https://github.com/swaywm/sway/blob/2a684cad5fc8e12a8e47a7fd00e2b7c66b43afb0/sway/commands/font.c#L16-L22) in the main configuration. But that is not enough as the bar font is usually inherited from the main config font, which would leave the `pango_markup` member [default initialized](https://github.com/swaywm/sway/blob/989123a2a5933367e5c7c39c3793f4814e026bf1/swaybar/config.c#L27) to false and then [forwarded to the bar](https://github.com/swaywm/sway/blob/a047b5ee4a2a67d30d93641ff86531d54b8e0879/sway/ipc-json.c#L1117-L1118) unconditionally.

### Appendix
The behavior can be fixed with the sway `pango_markup` config option, but is not backwards compatible with i3. Does this option has any advantage over the `i3` approach?

PR #6138, opened 2021-03-24
config: Fix swaybar pango_markup inconsistency

danielotero:
Until now, swaybar did not have pango markup enabled by default, even if
the sway config had it on. This patch aims to mimic the i3 behavior, but
maintaining the functionality of the "pango_markup" sway config command.

Fixes swaywm/sway#6118

More details can be found in the referenced issue.


danielotero:
Ping

emersion (approved):
LGTM, thanks

PR #6146, opened 2021-03-30
Remove advice about Firefox from issue template

emersion:
Firefox got a lot better. I think now would be a good time to remove
the advice from the issue template. We can always add it back if we
start getting invalid bug reports again.


kennylevinsen (approved):
LGTM.

kennylevinsen:
Build failing due to lacking xcb-icccm.

emersion:
Ah, pushed a fix for that.

kennylevinsen:
Thanks!

PR #6151, opened 2021-04-01
output: damage whole output when exiting scanout

### LINKED ISSUE #6149: Overlay layer layer-shell clients cause damage/rendering issues with fullscreen mpv
```
$ swaymsg -pt get_version
sway version 1.6-rc2-1d62d6bf (Mar 28 2021, branch 'makepkg')
$ pacman -Q mpv mesa
mpv 1:0.33.0-4
mesa 21.0.1-1
```

The best clients I know of to reproduce this with are mpv and mako, but any overlay client should do. Steps to reproduce:

```
# maybe relevant
$ cat mpv.conf
gpu-context=wayland
profile=gpu-hq
fs=yes
$ cat ~/.config/mako/config
layer=overlay
anchor=bottom-right
default-timeout=4000
```

1. Start video playback in mpv fullscreen and pause the video at any time.
2. Switch to another workspace and arrange for a notification to trigger mako, e.g. `sleep 5 && notify-send testing testing`
3. Switch back to mpv and wait for the overlay notification to appear.

When the notification appears, the fullscreen mpv window is replaced by whatever image was on the screen before switching to the fullscreen mpv workspace. With the steps above that would be the workspace with the terminal that sent the notify-send, but you could also switch to another workspace before switching to mpv to show that instead. It is a past image of the workspace, not a current render. If there is a bar with a clock on that workspace you can observe that the time does not change.

The entire overlay surface is displayed correctly, including the transparent border around the notification which reveals a small corner of the mpv window that should be covering the whole screen. I would attach a picture, but a screenshot doesn't seem to show the issue and in fact immediately corrects it. Here's my shitty drawing instead:

grey: image of term on other workspace
red: mpv seen through the transparent part of mako's overlay surface
blue: opaque mako notification

g43

If you play the mpv video, the issue is corrected on the next frame, so you only see a quick flicker when the notification appears and disappears while mpv is playing. You can also replace mako with `wlroots/examples/layer-shell`, and observe rapid flickering of the mpv window <-> previous workspace on every committed frame of the layer-shell client, but _only_ if you switch to fullscreen mpv _before_ the overlay client is mapped. If you start layer-shell and then switch workspaces it works fine.

Reproduction isn't quite 100%, but is fairly reliable. If it doesn't happen or only displays a short flicker you can just try again.

The obvious thing to try is `sway -D damage=rerender`, which does fix it for me. I also tried shuffling some of the logic in output_render but no luck. Really not sure why I can only reproduce with mpv.

This is not a regression – sway has had this behavior for me since as long as I can remember, but I only just tried looking into it recently.

RPigott:
Fixes #6149. See issue.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -553,6 +553,7 @@ static int output_repaint_timer_handler(void *data) {
 		if (last_scanned_out && !scanned_out) {
 			sway_log(SWAY_DEBUG, "Stopping fullscreen view scan out on %s",
 				output->wlr_output->name);
+			output_damage_whole(output);
 		}
 		last_scanned_out = scanned_out;
 


emersion (approved):
LGTM, thanks!

PR #6152, opened 2021-04-03
container: don't set fullscreen on children

ashkitten:
the original behavior set fullscreen for all descendents of a container, which causes issues when firefox is one of those children because it sends its own set_fullscreen request in response to being fullscreened.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1039,16 +1039,15 @@ void container_end_mouse_operation(struct sway_container *container) {
 	}
 }
 
-static void set_fullscreen_iterator(struct sway_container *con, void *data) {
+static void set_fullscreen(struct sway_container *con, bool enable) {
 	if (!con->view) {
 		return;
 	}
 	if (con->view->impl->set_fullscreen) {
-		bool *enable = data;
-		con->view->impl->set_fullscreen(con->view, *enable);
+		con->view->impl->set_fullscreen(con->view, enable);
 		if (con->view->foreign_toplevel) {
 			wlr_foreign_toplevel_handle_v1_set_fullscreen(
-				con->view->foreign_toplevel, *enable);
+				con->view->foreign_toplevel, enable);
 		}
 	}
 }
@@ -1058,9 +1057,7 @@ static void container_fullscreen_workspace(struct sway_container *con) {
 				"Expected a non-fullscreen container")) {
 		return;
 	}
-	bool enable = true;
-	set_fullscreen_iterator(con, &enable);
-	container_for_each_child(con, set_fullscreen_iterator, &enable);
+	set_fullscreen(con, true);
 	con->pending.fullscreen_mode = FULLSCREEN_WORKSPACE;
 
 	con->saved_x = con->pending.x;
@@ -1094,9 +1091,7 @@ static void container_fullscreen_global(struct sway_container *con) {
 				"Expected a non-fullscreen container")) {
 		return;
 	}
-	bool enable = true;
-	set_fullscreen_iterator(con, &enable);
-	container_for_each_child(con, set_fullscreen_iterator, &enable);
+	set_fullscreen(con, true);
 
 	root->fullscreen_global = con;
 	con->saved_x = con->pending.x;
@@ -1122,9 +1117,7 @@ void container_fullscreen_disable(struct sway_container *con) {
 				"Expected a fullscreen container")) {
 		return;
 	}
-	bool enable = false;
-	set_fullscreen_iterator(con, &enable);
-	container_for_each_child(con, set_fullscreen_iterator, &enable);
+	set_fullscreen(con, false);
 
 	if (container_is_floating(con)) {
 		con->pending.x = con->saved_x;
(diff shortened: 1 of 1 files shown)


ashkitten (commented):
i'm not sure what's going on here, so i didn't touch it yet

sway/tree/container.c
```diff
@@ -1380,8 +1377,8 @@ void container_add_child(struct sway_container *parent,
 	container_for_each_child(child, set_workspace, NULL);
 	bool fullscreen = child->pending.fullscreen_mode != FULLSCREEN_NONE ||
 		parent->pending.fullscreen_mode != FULLSCREEN_NONE;
-	set_fullscreen_iterator(child, &fullscreen);
-	container_for_each_child(child, set_fullscreen_iterator, &fullscreen);
+	set_fullscreen(child, &fullscreen);
+	container_for_each_child(child, set_fullscreen, &fullscreen);
```
ashkitten, line 1381: i'm not sure what's going on here, so i didn't touch it yet

RPigott:
> which causes issues when firefox is one of those children because it sends its own set_fullscreen request in response to being fullscreened.

Can you explain the issue in more detail? Why is this a problem?

ashkitten:
> Can you explain the issue in more detail (expected vs actual)? Why is this a problem?

when firefox is a child of a container, and the parent is fullscreened, the firefox window is told that it's fullscreen as well, and for whatever reason it sends its own set_fullscreen request in response, making it fullscreen on its own. even if it didn't do that, firefox has a unique fullscreen ui which is undesired when fullscreening the parent container.

when fullscreening a parent container...
expected result: the parent container becomes fullscreen, the firefox ui remains as is
actual result: firefox becomes fullscreen on its own, in its fullscreen ui

RPigott:
Oh I see. Maybe we should just ignore set_fullscreen request when we have already set fullscreen for a client, like is always done with schedule_configure in when handling a maximize request.

ashkitten:
even if firefox was not allowed to set itself fullscreen, the fullscreen ui is undesirable. it doesn't make sense to tell a client that it is fullscreen, when it isn't actually

RPigott:
I don't think whether or not FF's fullscreen UI is desirable is relevant. I do at least think the set_fullscreen request should be idempotent if we really want to consider a child of a fullscreen container as fullscreened. I don't really know the justification for setting the fullscreen state on children of a fullscreen view, because they are indeed not "fullscreen". However the fullscreen state also hints more to the client than just a size – should it render transparent regions/shadows, draggable interactive resize corner's, etc.

This probably warrants further discussion. Some of the discussion in #4255 is maybe relevant.

@emersion @Xyene any thoughts? For the record, I don't think i3 has ever supported fullscreened parent containers.

emersion:
> I do at least think the set_fullscreen request should be idempotent if we really want to consider a child of a fullscreen container as fullscreened

Agreed. We should at least be consistent in our fullscreen state handling.

> I don't really know the justification for setting the fullscreen state on children of a fullscreen view, because they are indeed not "fullscreen". However the fullscreen state also hints more to the client than just a size – should it render transparent regions/shadows, draggable interactive resize corner's, etc.

Yes. Still not sure about this.

Xyene:
> For the record, I don't think i3 has ever supported fullscreened parent containers.

It does, if you have `H[V[app nautilus*] app]` and `fullscreen`, nautilus drops its decorations (as expected). If you `focus parent` to get `H[V*[app nautilus] app]` and `fullscreen` again, you get the fullscreened `V`, and nautilus continues showing its decorations.

RPigott:
Oh interesting, I never knew. Well that saves us the trouble of thinking about it probably – we'll just do whatever i3 does, which in this case seems most similar to "don't set fullscreen state on children of a fullscreen container", right?

Xyene:
Yes, that seems reasonable. In i3, the output of `get_tree` on a workspace containing the previous example shows `fullscreen_mode: 1` for the `V`, and `fullscreen_mode: 0` for both of its children. I imagine sway same after this change, but that's something to double-check.

sway/tree/container.c
RPigott: Sway is spreading the fullscreen state to new children of the container. If we don't want children to inherit the fullscreen state, this part is no longer necessary.

I tried adding a new child to a fullscreen container in i3 and it refused with a "((null))" protocol error message, so I suppose we should just no longer allow this by just refusing the move command.

sway/tree/container.c
```diff
@@ -1049,8 +1049,7 @@ static void container_fullscreen_workspace(struct sway_container *con) {
 		return;
 	}
 	bool enable = true;
-	set_fullscreen_iterator(con, &enable);
-	container_for_each_child(con, set_fullscreen_iterator, &enable);
+	set_fullscreen(con, &enable);
```
RPigott, line 1052: We no longer need a data pointer for the iterator, so I think just `set_fullscreen(con, true)` is cleaner.

ashkitten:
i believe i've addressed the changes requested

RPigott:
Hey, sorry for the delay, looking good. Just one thing:

It's not technically necessary for this PR, but we should probably drop this bit as well https://github.com/swaywm/sway/blob/4e77bc293515ea2f00ebbc4395cb0eb4446d0195/sway/commands/fullscreen.c#L37-L43 to match i3's behavior for fullscreened split containers. It's not necessary to check the parent's fullscreen state because i3 allows you to fullscreen children of fullscreened containers just fine.

RPigott:
Also, can you rebase to appease the ci? If you want to drop the fullscreen command bit you can do it in a new commit in this PR, otherwise I can just merge this one as is and fix that later.

ashkitten:
rebased on master. i can't say i understand the fullscreen command code, so i'd prefer to just leave it out of this pr.

RPigott (approved):
Okay, that's fine. This LGTM, thanks!

PR #6161, opened 2021-04-04
desktop/layer_shell: fix centering for opposing anchors

### LINKED ISSUE #6160: Sway: anchors on opposing sides do not center the layer-shell
The documentation of the layer-shell protocol states:

```xml
<request name="set_anchor">
  <description summary="configures the anchor point of the surface">
    Requests that the compositor anchor the surface to the specified edges
    and corners. If two orthogonal edges are specified (e.g. 'top' and
    'left'), then the anchor point will be the intersection of the edges
    (e.g. the top left corner of the output); otherwise the anchor point
    will be centered on that edge, or in the center if none is specified.

    Anchor is double-buffered, see wl_surface.commit.
  </description>
  <arg name="anchor" type="uint" enum="anchor"/>
</request>
```

It isn't mentioned that two opposing anchors are not allowed, but when trying
that the window doesn't get accepted. So maybe mention something about that in
the documentation? I would suggest adding something like this:

>     If two opposing edges are specified it will result in undefined
>     behaviour.

fwSmit:
Fixes #6160. Tested it with the mako patch mentioned in the issue.

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -115,9 +115,10 @@ static void arrange_layer(struct sway_output *output, struct wl_list *list,
 		// Horizontal axis
 		const uint32_t both_horiz = ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT
 			| ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
-		if ((state->anchor & both_horiz) && box.width == 0) {
+		if (box.width == 0) {
 			box.x = bounds.x;
-			box.width = bounds.width;
+		} else if ((state->anchor & both_horiz) == both_horiz) {
+			box.x = bounds.x + ((bounds.width / 2) - (box.width / 2));
 		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT)) {
 			box.x = bounds.x;
 		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT)) {
@@ -128,9 +129,10 @@ static void arrange_layer(struct sway_output *output, struct wl_list *list,
 		// Vertical axis
 		const uint32_t both_vert = ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP
 			| ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM;
-		if ((state->anchor & both_vert) && box.height == 0) {
+		if (box.height == 0) {
 			box.y = bounds.y;
-			box.height = bounds.height;
+		} else if ((state->anchor & both_vert) == both_vert) {
+			box.y = bounds.y + ((bounds.height / 2) - (box.height / 2));
 		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP)) {
 			box.y = bounds.y;
 		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM)) {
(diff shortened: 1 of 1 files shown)


RPigott:
Are you sure this is a regression? Looks like this bit of code hasn't changed in a while, so I think it always worked like this. Should be fixed either way, but would be nice to know what caused it if it changed.

sway/desktop/layer_shell.c
```diff
@@ -115,25 +115,35 @@ static void arrange_layer(struct sway_output *output, struct wl_list *list,
...
 			| ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
+		const bool match_both_horiz =
+			(state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT) &&
+			(state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT);
 		if ((state->anchor & both_horiz) && box.width == 0) {
 			box.x = bounds.x;
 			box.width = bounds.width;
-		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT)) {
+		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT)
+				&& !match_both_horiz) {
```
RPigott, line 125: How about just adding another both left + right case above for box.width != 0? I think it makes sense to preserve the simple "both -> left -> right -> neither" flow.

I don't like the use of "both_horiz" in this function, since it sort of seems like state->anchor & both_horiz should only be truthful if the anchor contains both horizontal borders, but it's actually true if the anchor contains either. So I think you could just replace the definition of `both_horiz` with your `match_both_horiz` and the whole block would make more sense.

sway/desktop/layer_shell.c
```diff
@@ -115,25 +115,35 @@ static void arrange_layer(struct sway_output *output, struct wl_list *list,
...
 			| ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM;
+		const bool match_both_vert =
+			(state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP) &&
+			(state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM);
 		if ((state->anchor & both_vert) && box.height == 0) {
 			box.y = bounds.y;
 			box.height = bounds.height;
-		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP)) {
+		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP)
+				&& !match_both_vert) {
```
RPigott, line 143: Same.

sway/desktop/layer_shell.c
fwSmit: Sure, but that would mean duplicating the averaging code.

RPigott: I don't think that's a problem – it's only one line. It makes more sense to me to keep all the cases clear.

fwSmit:
> Are you sure this is a regression? Looks like this bit of code hasn't changed in a while, so I think it always worked like this. Should be fixed either way, but would be nice to know what caused it if it changed.

I never said there was a regression. On taking a closer look, there isn't a difference between sway 1.5.1 and master in case (unlike what I said in the issue). There's only a difference between mako and another client I tested, dunst. Mako shows the surface as if anchored to the left and dunst doesn't manage to show a surface at all, because it immediately gets closed. I didn't investigate it further though.

RPigott:
Oh, ok.

fwSmit:
I now know why the other client failed. It was applying too big of a margin which made the width < 0. I don't really agree with the way that margin is calculated now.
Now when you have left and right anchors, the margin goes immediately from the width of the surface instead of calculating from the edge of the screen.
For no anchors, no margin is applied.

sway/desktop/layer_shell.c
```diff
@@ -139,18 +143,12 @@ static void arrange_layer(struct sway_output *output, struct wl_list *list,
 			box.y = bounds.y + ((bounds.height / 2) - (box.height / 2));
 		}
 		// Margin
-		if ((state->anchor & both_horiz) == both_horiz) {
-			box.x += state->margin.left;
-			box.width -= state->margin.left + state->margin.right;
```
RPigott, line 144: If the client requests 0 width and opposing anchors we still need to honor the margin, so we can't drop this.

I'm thinking the cases should really be "if width == 0" and "if left else if right". wlroots guarantees width == 0 implies both horizontal anchors by enforcing the protocol.

fwSmit: Oh right, width = 0 means as wide as possible. I re added it and changed the check to `width = 0` (moving the width changing below).

fwSmit:
I didn't apply margins for the case where both anchors for now, because I don't know what would be the best way to apply margins there. If you think it makes  sense, I can do `box.x += state->margin.left - state->margin.right` to allow for the use case of offsetting from the center. Other options would be too complicated and less useful I think.

RPigott (approved):
Yeah, I don't think there's much point in applying margins to centered surfaces, so that seems fine.

This LGTM. Thanks!

RPigott:
Err, can you rebase to appease the ci? I'm not sure how to tell GH to ignore the test failures from the wlroots version mismatch. Then I can merge it.

fwSmit:
> Err, can you rebase to appease the ci? I'm not sure how to tell GH to ignore the test failures from the wlroots version mismatch. Then I can merge it.

I think github allows maintainers to edit PR branches by default. But anyways, you should be able to merge it after the builds have completed.

PR #6175, opened 2021-04-12
Match i3 floating enable/disable behavior

RPigott:
"Bug report" I saw on irc:

> 15:23 <FreeFull_> By the way, I bug I noticed: With tabbed layouts, often when I unfloat a window, it'll end up in a split layout rather than going in as a tab

i3 tries a little harder to restore floated windows to their original position, so let's do that too. Ref in i3: https://github.com/i3/i3/blob/9bf9cb9926402e73aa62e4c1bcc03d671c849c2e/src/floating.c#L244


Xyene (approved):
LGTM and works, thanks!

PR #6176, opened 2021-04-12
Log wlroots version on startup

emersion:
Can be useful to make sure a bugfix is included. In the future maybe
the wlroots version string could include a commit hash when built
from source, too.


kennylevinsen (approved):

kennylevinsen:
LGTM, thanks!

PR #6182, opened 2021-04-13
Use "!ptr" for comparison to NULL

ElyesH:


emersion:
Sorry, `ptr == NULL` is just as good as `!ptr`

PR #6185, opened 2021-04-14
Mandatory libseat

kennylevinsen:
See https://github.com/swaywm/wlroots/pull/2839

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -76,7 +76,6 @@ endif
 
 wlroots_features = {
 	'xwayland': false,
-	'libseat': false,
 }
 foreach name, _ : wlroots_features
 	var_name = 'have_' + name.underscorify()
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -316,7 +315,3 @@ summary({
 	'man-pages': scdoc.found(),
 }, bool_yn: true)
 
-if not wlroots_features['libseat']
-	warning('The sway binary must be setuid when compiled without (e)logind or libseat')
```
emersion, line 320: We loose this warning when libseat is compiled with the builtin backend. I wonder if it's a big deal. Probably not.

emersion (approved):
LGTM, thanks!

meson.build
kennylevinsen: I guess we could probe pkgconfig in the future.

martinetd:
note README also needs updating (still says "* libseat (optional, for (e)logind and seatd support)")... perhaps with a link to https://git.sr.ht/~kennylevinsen/seatd ? although I guess people can probably find it if they try...
(Yes, I'm this lazy!)

martinetd:
ugh, sorry posted that in the wrong repo -- that was meant for wlroots'which had a similar change.

emersion:
Good catch, done in https://github.com/swaywm/wlroots/commit/9f33d8ad39c26c00ef129419ea1c69c6966e5606

PR #6190, opened 2021-04-16
cmd_fullscreen: allow fullscreen on fullscreen split containers

RPigott:
#6152 fixes fullscreen split containers to better match i3 visually.

For commands, using the fullscreen command on a child of a fullscreen split container should fullscreen that child, rather than just unfullscreen the parent.

I've also noticed that i3 seems to refuse move commands that add a child to a fullscreened container, but I've left that alone for now.


emersion (approved):
Thanks!

PR #6191, opened 2021-04-16
Minor exec improvements

emersion:
_See individual commits._

### DIFF
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -66,7 +66,8 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 		if ((child = fork()) == 0) {
 			close(fd[1]);
 			execlp("sh", "sh", "-c", cmd, (void *)NULL);
-			_exit(0);
+			sway_log_errno(SWAY_ERROR, "execlp failed");
+			_exit(1);
 		}
 		ssize_t s = 0;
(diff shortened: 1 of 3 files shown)


sway/commands/exec_always.c
```diff
@@ -65,8 +65,9 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 		close(fd[0]);
 		if ((child = fork()) == 0) {
 			close(fd[1]);
-			execl("/bin/sh", "/bin/sh", "-c", cmd, (void *)NULL);
-			_exit(0);
+			execlp("sh", "sh", "-c", cmd, NULL);
```
codicodi, line 68: While this should work fine on most platforms, C spec allows `NULL` to be an integer, and `man 3 exec` says that `NULL` should be cast to pointer type.

emersion: Fair enough, reverted this change.

codicodi: Same applies to all the other `execlp`s (although they were missing a cast before too)

emersion: Right. I don't care enough to make the change in a separate commit, but feel free to open a PR if you want.

kennylevinsen (approved):

kennylevinsen:
Thanks!

PR #6207, opened 2021-04-18
Ignore superfluous fullscreen output hints

### LINKED ISSUE #6202: Sway spawns empty container upon entering fullscreen in qutebrowser
- **Sway Version:** 1.6

- **Debug Log:**
[sway.log](https://gist.github.com/lokesh-krishna/23d416df1a768aa25656d7eefe5c2de5l)

- **Configuration File:**
Can reproduce with default config

- **Description:**
  - Launch qutebrowser (2.2.0) and use an in-tab full screen button to enter fullscreen while in tabbed mode
  - Exit full screen 
  - Sway shifts back to tiled mode and there is now an empty container
  - Repeating the process keeps generating empty containers

RPigott:
So for whatever reason qutebrowser likes to give us an output hint for fullscreen requests, which is fine. But in response sway tries to add qutebrowser to the tiling workspace it's already in and bungles things.

There's really two bugs here I think,

1. There's no reason to remove and re-add a window to the workspace it's already in if it requests the output it's already on. This would mess up the tiling order even without the ghost containers because it get's re-added at workspace level even if it was nested.
2. Doing that shouldn't make ghost containers anyway, so let's reap them in workspace_tiling add.

Fixes #6202.

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -358,7 +358,8 @@ static void handle_request_fullscreen(struct wl_listener *listener, void *data) 
 	if (e->fullscreen && e->output && e->output->data) {
 		struct sway_output *output = e->output->data;
 		struct sway_workspace *ws = output_get_active_workspace(output);
-		if (ws && !container_is_scratchpad_hidden(container)) {
+		if (ws && !container_is_scratchpad_hidden(container) &&
+				container->pending.workspace != ws) {
 			if (container_is_floating(container)) {
 				workspace_add_floating(ws, container);
 			} else {
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -794,7 +794,11 @@ void workspace_detach(struct sway_workspace *workspace) {
 struct sway_container *workspace_add_tiling(struct sway_workspace *workspace,
 		struct sway_container *con) {
 	if (con->pending.workspace) {
+		struct sway_container *old_parent = con->pending.parent;
 		container_detach(con);
+		if (old_parent) {
+			container_reap_empty(old_parent);
+		}
 	}
 	if (config->default_layout != L_NONE) {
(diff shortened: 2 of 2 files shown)


sway/desktop/xdg_shell.c
```diff
@@ -358,7 +358,7 @@ static void handle_request_fullscreen(struct wl_listener *listener, void *data)
 	if (e->fullscreen && e->output && e->output->data) {
 		struct sway_output *output = e->output->data;
 		struct sway_workspace *ws = output_get_active_workspace(output);
-		if (ws && !container_is_scratchpad_hidden(container)) {
+		if (ws && container->pending.workspace != ws) {
```
Xyene, line 361: `container_is_scratchpad_hidden` also contains a check for `con->scratchpad` -- should we be keeping it?

Xyene (approved):
This LGTM (my question is just a question), though I don't quite understand the mechanism by which the ghost container was being created in the first place.

sway/desktop/xdg_shell.c
RPigott: `con->scratchpad` is true for containers that are shown from the scratchpad, so they could belong to a different non-scratchpad workspace. We could just retain the scratchpad hidden check though, like `!container_is_scratchpad_hidden(container) &&` which seems fine.

Leaving it out makes sway honor the output hint for scratchpad hidden containers. I don't think I care either way, but it wasn't really my intention to change the behavior in order to allow that, so I'll add it back in. Good catch.

RPigott:
> I don't quite understand the mechanism by which the ghost container was being created

It seems to be a bug in workspace_add_tiling, which will detach a container for you but not necessarily reap it's possibly empty parent. The ghosts would also have appeared if the client requested fullscreen on an active output from a different workspace too (or more realistically via foreign toplevel fullscreen request), but I think that's it. This function is sparsely used with attached containers I guess.

Issue #6218, opened 2021-04-23
Tab dragging functionality


tomKPZ:
Tab dragging is an intuitive, well understood functionality implemented by VScode [1], Chrome, Firefox, etc.  Sway currently allows detaching tabs by dragging them, but cannot join the tab to an existing tab or stack container.

I've implemented this behavior with a patch [2] and I'll open a PR shortly.

[1] https://user-images.githubusercontent.com/9124077/115812431-3ab87d80-a3a6-11eb-9f2b-e31525d75857.mp4
[2] https://user-images.githubusercontent.com/9124077/115812402-2bd1cb00-a3a6-11eb-8868-ad9db6185095.mp4

Ashe:
Really looking forward to this getting implemented, I miss this from i3!

emersion:
Implemented in https://github.com/swaywm/sway/pull/6219

PR #6219, opened 2021-04-23
Add tab dragging functionality

tomKPZ:
Implements functionality described in [1].  Please see the issue for a
video with a demonstration of the new behavior.

An issue is that titlebars cover up a significant portion of the top
edge drop area.  The solution is simply to change the edge drop area
hitbox to start at the contents instead of the container.

[1] https://github.com/swaywm/sway/issues/6218

### DIFF
--- a/sway/input/seatop_move_tiling.c
+++ b/sway/input/seatop_move_tiling.c
@@ -16,6 +16,10 @@
 // Thickness of the dropzone when dragging to the edge of a layout container
 #define DROP_LAYOUT_BORDER 30
 
+// Thickness of indicator when dropping onto a titlebar.  This should be a
+// multiple of 2.
+#define DROP_SPLIT_INDICATOR 10
+
 struct seatop_move_tiling_event {
 	struct sway_container *con;
 	struct sway_node *target_node;
@@ -23,6 +27,8 @@ struct seatop_move_tiling_event {
 	struct wlr_box drop_box;
 	double ref_lx, ref_ly; // cursor's x/y at start of op
 	bool threshold_reached;
+	bool split_target;
+	bool insert_after_target;
 };
 
 static void handle_render(struct sway_seat *seat,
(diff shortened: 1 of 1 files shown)


sway/input/seatop_move_tiling.c
```diff
@@ -92,8 +94,59 @@ static void resize_box(struct wlr_box *box, enum wlr_edges edge,
 	}
 }
 
+static bool titlebar_box(struct sway_node *node, struct wlr_cursor* cursor,
+		struct wlr_box *title_box) {
+	struct sway_container* con = node->sway_container;
+	struct sway_node *parent = &con->pending.parent->node;
+	int title_height = container_titlebar_height();
+	struct wlr_box box;
+	list_t* children;
```
Xyene, line 103: Nit: `list_t *children`.

sway/input/seatop_move_tiling.c
```diff
@@ -92,8 +94,59 @@ static void resize_box(struct wlr_box *box, enum wlr_edges edge,
...
+	int tab_width, child_index;
+	switch(parent ? node_get_layout(parent) : L_NONE) {
```
Xyene, line 105: Nit: missing space after `switch`.

sway/input/seatop_move_tiling.c
```diff
@@ -92,8 +94,59 @@ static void resize_box(struct wlr_box *box, enum wlr_edges edge,
 	}
 }
 
+static bool titlebar_box(struct sway_node *node, struct wlr_cursor* cursor,
+		struct wlr_box *title_box) {
+	struct sway_container* con = node->sway_container;
```
Xyene, line 99: Nit: `*con`.

sway/input/seatop_move_tiling.c
```diff
@@ -92,8 +94,59 @@ static void resize_box(struct wlr_box *box, enum wlr_edges edge,
 	}
 }
 
+static bool titlebar_box(struct sway_node *node, struct wlr_cursor* cursor,
```
Xyene, line 97: Nit: `*cursor`.

Xyene (commented):
I'm a fan of this idea, it's something I'd thought about before but never got to myself -- thanks for working on it. Here's a first round of review.

Some initial thoughts:

- There should be a way to cancel a drag & drop action. For all other actions, releasing the drag over the drop zone it originated is a no-op, here, releasing a grabbed tab over where it was grabbed splits the tab out of the container.
- The logic of `after_target = 2 * cursor->cursor->x > 2 * box.x + box.width` is useful, but it's also a bit visually unpredictable due to how the drop region is rendered. It would be nicer if we could render a thinner bar between the borders of the two adjacent containers that the dragged tab will be inserted between.

tomKPZ:
Latest patch addresses comments.  See [1] for new behavior.  One thing I'm wondering is if it'd be useful to have a different action when dragging on the middle third of a tab, like inserting within the nested container, or swapping the containers.  Firefox and Chrome have a similar functionality [2] [3], but maybe this'd be better as a follow-up.

[1] https://user-images.githubusercontent.com/9124077/115979853-307cb780-a53d-11eb-9b35-0c434507f5b7.mp4
[2] https://user-images.githubusercontent.com/9124077/115979916-91a48b00-a53d-11eb-975c-d112e644dc39.mp4
[3] https://user-images.githubusercontent.com/9124077/115979930-a84ae200-a53d-11eb-97ca-9d4b9d1ddd94.mp4

tomKPZ:
Friendly ping

Is any further action required?

sway/input/seatop_move_tiling.c
```diff
@@ -16,13 +16,19 @@
...
+
 struct seatop_move_tiling_event {
 	struct sway_container *con;
 	struct sway_node *target_node;
 	enum wlr_edges target_edge;
 	struct wlr_box drop_box;
 	double ref_lx, ref_ly; // cursor's x/y at start of op
 	bool threshold_reached;
+	bool split_target;
+	bool after_target;
```
Xyene, line 31: Nit: `insert_after_target` is probably a better name here.

sway/input/seatop_move_tiling.c
```diff
@@ -127,27 +201,53 @@ static void handle_motion_postthreshold(struct sway_seat *seat) {
...
+	// container is not a descendant of the source container.
+	if (!surface && !container_has_ancestor(con, e->con) &&
+			split_titlebar(node, e->con, cursor->cursor,
+			&e->drop_box, &e->after_target)) {
+		e->target_edge = WLR_EDGE_NONE;
+		// Don't allow dropping over the source container's titlebar
+		// to give users a chance to cancel a drag operation.
+		if (con == e->con) {
+			e->target_node = NULL;
+			return;
```
Xyene, line 214: Nit: let's drop this return and just put the fall-through path in an `else`, so that the `e->target_node` assignments mirror their indentation level.

sway/input/seatop_move_tiling.c
```diff
@@ -127,27 +201,53 @@ static void handle_motion_postthreshold(struct sway_seat *seat) {
 		return;
 	}
 
+	// Check if the cursor is over a tilebar only if the destination
+	// container is not a descendant of the source container.
+	if (!surface && !container_has_ancestor(con, e->con) &&
+			split_titlebar(node, e->con, cursor->cursor,
+			&e->drop_box, &e->after_target)) {
```
Xyene, line 208: Nit: I think we want an extra tab here for the hanging params of `split_titlebar`.

Xyene (commented):
Sorry, I've been swamped these days. The code looks good apart from some small nits and the demo videos look excellent (thanks for providing those!)

I'd like to test this out locally before merging, which I think I'll have time to do this weekend.

Xyene:
Had a chance to test this, and it looks good to me. Could you please squash your commits so that we can merge this?

tomKPZ:
Squashed commits, thank you for reviewing!  I had a lot of fun with this and look forward to creating more patches in the future :)

Xyene (approved):
Thanks!

Issue #6223, opened 2021-04-24
Gtk4 window gets clipped within container


keis:
### Please fill out the following:
- **Sway Version:** sway version 1.6-1a0f86be (Apr 10 2021, branch 'master')
- Also tested with current master (8529141150ef2d9a07b9c0fad193bdbc12d3ad42) and various commits after 1.5

- **Debug Log:**
https://gist.github.com/keis/3f8830b80ba470b0018166608dbabaaf

- **Configuration File:**
```
output * bg /usr/share/backgrounds/default.png fill
exec "gtk4-demo"
```

- **Description:**
In a regression from sway 1.5 Gtk4 windows (in my example the gtk4-demo) render with a weird clipping within the sway container.

I was able to `git bisect` to this commit a6544f5a64e5d0899555795915d9a05f7bb09f9c

![20210424_11h13m10s_grim](https://user-images.githubusercontent.com/125062/115954196-f20cdb80-a4ef-11eb-99fc-9cdc53a5fbe8.png)

emersion:
Probably related to a confusion between window geometry and surface geometry.

kennylevinsen:
The window seems placed wrong, look at how close the upper left corner button is to the edge compared to the right.

emersion:
I think it's not mis-placed, but it's just clipped? The close button is cut in the middle.

keis:
Indeed. For reference this is what the demo window looks like when floating

![20210424_14h40m57s_grim](https://user-images.githubusercontent.com/125062/115959164-2641c580-a50b-11eb-9de0-e5fe01065de9.png)

kennylevinsen:
Ah true, thought a thick margin was intended, didn't see a close button was cut.

nagisa:
May be relevant to this: at some point I reported a similar issue (shadows part of window decoration not disappearing when window is tiled) to #gtk on IRC and was told that gtk4 does not yet implement the xdg-shell protocol or something along those lines.

emersion:
GTK4 does implement xdg-shell, that's the only way they can display a window on screen. Maybe they don't implement the tiled states, but that would be surprising since GNOME uses them as well.

nagisa:
![Screenshot showing shadows from a gtk4 window being drawn over other windows tiled to the top and left of the gtk4-demo window](https://user-images.githubusercontent.com/679122/115989097-2ba31c80-a5ac-11eb-9ebc-420ca489e6f1.png)

[...]

PR #6229, opened 2021-04-25
Fix offset toplevel geometry issue

kennylevinsen:
The first commit removes `surface->sx`/`surface->sy` from these calculations. It doesn't work, break things if non-zero, and causes confusion.

The second commit shuffles the clip box handling around a bit, in order to compensate for the offset in the clip dimensions, which could otherwise lead to insufficient or excessive clipping.

Closes: https://github.com/swaywm/sway/issues/6223


emersion (approved):
Thanks!

PR #6230, opened 2021-04-26
Remove support for arbitrary rotations

emersion:
There was some unused code-paths for rendering surfaces with an
arbitrary rotation applied. This was imported from rootston.

Since we don't have plans to make use of this, remove it.


kennylevinsen (approved):

kennylevinsen:
Nice catch, thanks!

PR #6243, opened 2021-04-29
Display window title and icon in swaybar

FlexW:
This pull request adds an icon and the name of the focused window to swaybar. I recorded a screencast to showcase it:

https://nextcloud.felixweilbach.de/s/gZCnf5G6WfTyc8B

I think such a feature could be pretty helpful, as it helps to track which window is currently focused when using gaps without window titles. Why not use a tool like i3blocks? With i3blocks it is difficult/impossible to display the window title aligned to the left side. I think it's also not possible to display any window icons with tools like i3blocks.

I know that sway does not want to have every feature, so my first question is if this will have a chance to get merged?

The pull request is not ready yet. I think the following points need to be addressed before it could be merged:

- [x] Introduce configuration option to disable or enable the feature. I think per default it should be disabled.
- [x] Document the option.
- [x] Cache the icons, so that they don't get loaded from disk every frame. (They still get loaded from the disk if the window changes, but I think that is okay.)
- [x] Clean the code, address remaining TODO's in the code 
- [x] Make sure it compiles without `HAVE_TRAY`.

Thanks for any feedback!

Signed-off-by: Felix Weilbach <felix.weilbach@t-online.de>


swaybar/desktop.c
```diff
@@ -0,0 +1,96 @@
...
+
+		if (next_line) {
+			*next_line = '\n';
+		}
+		cur_line = next_line ? (next_line + 1) : NULL;
+	}
+	free(desktop_entry_start);
+
+	return icon_name;
+}
```
YerinAlexey, line 96: See https://gitlab.freedesktop.org/ddevault/fdicons

swaybar/render.c
```diff
@@ -681,6 +732,76 @@ static uint32_t render_workspace_button(struct render_context *ctx,
...
+	} else {
+		// TODO: Generate a image on the fly
+		icon = load_background_image(
+				"/usr/share/icons/Adwaita/16x16/apps/"
+				"utilities-terminal-symbolic.symbolic.png");
+	}
+	assert(icon);
+	if (!icon) {
+		return output->height;
+	}
```
YerinAlexey, line 780: This if statement is unreachable because `icon` is asserted before

swaybar/render.c
```diff
@@ -573,25 +621,28 @@ static uint32_t render_binding_mode_indicator(struct render_context *ctx,
...
 
 	double text_y = height / 2.0 - text_height / 2.0;
 	cairo_set_source_u32(cairo, config->colors.binding_mode.text);
-	cairo_move_to(cairo, x + width / 2 - text_width / 2, (int)floor(text_y));
+	cairo_move_to(cairo, *x + width / 2 - text_width / 2, (int)floor(text_y));
 	choose_text_aa_mode(ctx, config->colors.binding_mode.text);
 	pango_printf(cairo, config->font, output->scale,
 			output->bar->mode_pango_markup, "%s", mode);
+
+	*x += width;
```
YerinAlexey, line 644: Does it really need to mutate `x`? Maybe you could just add width with later draw calls?

swaybar/render.c
```diff
@@ -681,6 +732,76 @@ static uint32_t render_workspace_button(struct render_context *ctx,
...
+				target_size,
+				icon_theme,
+				&min_size,
+				&max_size);
+		icon = load_background_image(icon_path);
+	} else {
+		// TODO: Generate a image on the fly
+		icon = load_background_image(
+				"/usr/share/icons/Adwaita/16x16/apps/"
+				"utilities-terminal-symbolic.symbolic.png");
```
YerinAlexey, line 775: If the application doesn't have the icon set, I think it's better to not show it at all instead of hard-coding this icon which also introduces problems for people who don't have Adwaita theme.

FlexW: Yes, my plan was to generate a image with cairo. Like it is done with missing tray icons.

swaybar/render.c
FlexW: I thought I keep it consistent with the other render functions.

FlexW: Yes, for example look at `render_status_line_text` or `render_workspace_button`

swaybar/render.c
YerinAlexey: In the case of tray, it makes sense to have a dummy icon because otherwise it won't be useful. But here it would be better if application icon is not shown at all if the application doesn't define any.

swaybar/render.c
YerinAlexey: Nevermind then

pedrocr:
This was previously rejected here:

https://github.com/swaywm/sway/pull/4463

FlexW:
> This was previously rejected here:
> 
> #4463
> 
> waybar can do this and for swaybar/i3 it's possible to do it with the i3status protocol but not be left-aligned. Even the workspace feature could be implemented with the i3status protocol. If the protocol had a way to specify alignment the whole bar could just be an i3status output without any features embedded in the bar itself. Sway has decided to just replicate i3 so it's probably a better idea to look to other bars for new features, including potentially forking swaybar into that.

Didn't saw that. Thanks. Then I close this pr.

emersion:
Yeah. Thanks for working on this, but at this point we'll just stick to i3 features. You're more than welcome to create a fork of swaybar though!

Issue #6232, opened 2021-04-26
Get scroll_factor value in swaymsg


ErikReider:
Currently when running `swaymsg -r -t get_inputs`, you aren't able to see the scroll_factor value because of it not being implemented in libinput. It would be a welcome improvement for it to be visible somewhere.

emersion:
Patches welcome.

PR #6251, opened 2021-05-04
Added scroll_factor to ipc

ErikReider:
This commit adds the capability to get the current scroll_factor from the config #6232

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -1,3 +1,4 @@
+#include <float.h>
 #include <json.h>
 #include <libevdev/libevdev.h>
 #include <stdio.h>
@@ -1000,6 +1001,17 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 					layout ? json_object_new_string(layout) : NULL);
 			}
 		}
+	}
+
+	if (device->wlr_device->type == WLR_INPUT_DEVICE_POINTER) {
+		struct input_config *ic = input_device_get_config(device);
+		float scroll_factor = 1.0f;
+		if (ic != NULL && !isnan(ic->scroll_factor) && 
+				ic->scroll_factor != FLT_MIN) {
+			scroll_factor = ic->scroll_factor;
+		}
+		json_object_object_add(object, "scroll_factor", 
+				json_object_new_double(scroll_factor));
 	}
 
 	if (wlr_input_device_is_libinput(device->wlr_device)) {
--- a/sway/sway-ipc.7.scd
+++ b/sway/sway-ipc.7.scd
@@ -1131,6 +1131,9 @@ following properties:
 |- xkb_active_layout_index
 :  integer
 :  (Only keyboards) The index of the active keyboard layout in use
+|- scroll_factor
+:  floating
+:  (Only pointers) Multiplier applied on scroll event values.
 |- libinput
 :  object
 :  (Only libinput devices) An object describing the current device settings.


sway/ipc-json.c
```diff
@@ -1003,6 +1004,12 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 	}
 
 	if (wlr_input_device_is_libinput(device->wlr_device)) {
+		struct input_config *ic = input_device_get_config(device);
+		if(ic != NULL && !isnan(ic->scroll_factor)) {
```
emersion, line 1008: Style: missing space between `if` and `(`

sway/ipc-json.c
```diff
@@ -1003,6 +1004,12 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 	}
 
 	if (wlr_input_device_is_libinput(device->wlr_device)) {
+		struct input_config *ic = input_device_get_config(device);
```
emersion, line 1007: The scroll factor is also applied when the device isn't coming from libinput (e.g. virtual pointers for VNC, running with wlroots' Wayland or X11 backend, etc). I think it would make sense to have a new `if (device->wlr_device->type == WLR_INPUT_DEVICE_POINTER)` block.

sway/sway-ipc.7.scd
```diff
@@ -1131,6 +1131,9 @@ following properties:
 |- xkb_active_layout_index
 :  integer
 :  (Only keyboards) The index of the active keyboard layout in use
+|- scroll_factor
+:  floating
+:  Changes the scroll factor for the specified input device.
```
emersion, line 1136: The description reads a little bit weird. Maybe something among these lines would be better?

```
(Only pointers) Multiplier applied on scroll event values.
```

emersion (changes requested):
The meat of the patch LGTM, here are a few comments.

sway/sway-ipc.7.scd
ErikReider: Isn't scroll_factor also applied to touchpads? Maybe (Only pointers and touchpads)

emersion: A touchpad is a pointer.

sway/ipc-json.c
```diff
@@ -1003,6 +1004,14 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
...
+		if (ic != NULL && device->wlr_device->type == WLR_INPUT_DEVICE_POINTER) {
```
emersion, line 1008: This should be moved outside of the `wlr_input_device_is_libinput` block

sway/ipc-json.c
```diff
@@ -1002,6 +1003,17 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 		}
 	}
 
+	if (device->wlr_device->type == WLR_INPUT_DEVICE_POINTER) {
+		struct input_config *ic = input_device_get_config(device);
+		if (ic != NULL) {
```
emersion, line 1008: We want to expose the `scroll_factor` field even if we don't have an input config, I think. Should default to 1.0 in this case.

ErikReider: So `float scroll_factor = ic->scroll_factor`?

emersion: I mean that we should call `json_object_object_add` even if `ic == NULL`.

emersion (commented):
LGTM! Can you squash all of the changes into a single commit?

sway/ipc-json.c
```diff
@@ -1002,6 +1003,17 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
...
 
+	if (device->wlr_device->type == WLR_INPUT_DEVICE_POINTER) {
+		struct input_config *ic = input_device_get_config(device);
+		float scroll_factor = 1.0f;
+		if (ic != NULL && !isnan(ic->scroll_factor) && 
+				ic->scroll_factor != FLT_MIN) {
+			scroll_factor = ic->scroll_factor;
+		}
+		json_object_object_add(object, "scroll_factor", 
+				json_object_new_int(scroll_factor));
```
emersion, line 1014: Ah, one last thing: this should be `json_object_new_float`, not `new_int`

ErikReider: Oh, my bad 😅

emersion (approved):
Thanks!

PR #6255, opened 2021-05-05
config.in: change terminal emulator to foot

ddevault:
This is my preferred terminal emulator now. Seeing as the default config
file is basically "Drew's preferences watered down a bit for a general
audience", I reckon it should be updated accordingly :)


emersion:
Thanks

PR #6262, opened 2021-05-07
cairo: Replace <cairo/cairo.h> by <cairo.h>

concatime:
For full context, read https://github.com/Cloudef/bemenu/issues/170 and https://gitlab.freedesktop.org/cairo/cairo/-/issues/479.
TL;DR, cairo’s pc file adds `/cairo` to CFLAGS.
So namespace cairo shouldn’t be used.
It was working simply because `/usr/include` is always used, not because of Meson + pkg-config.
A similar patch will land in swaybg too.
Thank you.

### DIFF
--- a/client/pool-buffer.c
+++ b/client/pool-buffer.c
@@ -1,6 +1,6 @@
 #define _POSIX_C_SOURCE 200809
 #include <assert.h>
-#include <cairo/cairo.h>
+#include <cairo.h>
 #include <fcntl.h>
 #include <pango/pangocairo.h>
 #include <stdio.h>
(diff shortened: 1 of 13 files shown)


include/cairo_renderer.h
```diff
@@ -2,7 +2,7 @@
 #define _SWAY_CAIRO_H
```
emersion, line 2: The include guard needs to be updated as well.

Maybe `cairo_util.h` would be a better name?

concatime: I chose `cairo_renderer.h` just like @Cloudef did in https://github.com/Cloudef/bemenu/commit/ebd7338bd5653a56d319a3a879f97e867413e58b. Should I rename it to `cairo_util.h`?

emersion: Yes. "Renderer" doesn't make much sense to me for Sway.

emersion (approved):
Thanks

PR #6265, opened 2021-05-08
commands/exec: stop truncating >4095-byte commands

mstoeckl:
This commit fixes a rare edge case, in which very long commands are silently truncated. One way to test this is to compare the following two shell commands; if truncation occurs, the first will fail to set the background color.

```
swaymsg exec "swaynag -m `(head -c 4096 < /dev/zero | tr '\0' 'a')` --background 00FF00"

swaynag -m `(head -c 4096 < /dev/zero | tr '\0' 'a')` --background 00FF00
```

### DIFF
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -26,7 +26,7 @@ struct cmd_results *cmd_exec_validate(int argc, char **argv) {
 
 struct cmd_results *cmd_exec_process(int argc, char **argv) {
 	struct cmd_results *error = NULL;
-	char *tmp = NULL;
+	char *cmd = NULL;
 	if (strcmp(argv[0], "--no-startup-id") == 0) {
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks

sway/commands/exec_always.c
```diff
@@ -36,17 +36,12 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 	}
 
 	if (argc == 1 && (argv[0][0] == '\'' || argv[0][0] == '"')) {
-		tmp = strdup(argv[0]);
-		strip_quotes(tmp);
+		cmd = strdup(argv[0]);
```
Mic92, line 39: This code is missing a NULL check in case it goes out of memory.

emersion: We don't have a OOM checking policy in Sway, yet. Many other places would need to get fixed.

Patches welcome I suppose.

PR #6268, opened 2021-05-11
C: Handle _POSIX_C_SOURCE macro properly

concatime:
`_POSIX_C_SOURCE` macro should be handled properly. I had a case where `_POSIX_C_SOURCE=200809L` was in my cflags, and the build was failing with (due to `werror=true` in Meson):
```
error: '_POSIX_C_SOURCE' macro redefined [-Werror,-Wmacro-redefined]
```
We should define `_POSIX_C_SOURCE` only when required, so I wrapped each define with a condition.

Thank you.

### DIFF
--- a/client/pool-buffer.c
+++ b/client/pool-buffer.c
@@ -1,4 +1,8 @@
-#define _POSIX_C_SOURCE 200809
+#if !defined(_POSIX_C_SOURCE) || _POSIX_C_SOURCE+0 < 200809L
+#undef _POSIX_C_SOURCE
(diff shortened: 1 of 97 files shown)


emersion:
NACK from me. `_POSIX_C_SOURCE` is specified by each source file to describe the POSIX requirements. It's not something someone should be able to change with cflags.

kennylevinsen:
Also NACK, defines in this category are meant to control included symbols, and any other value than the one specified will break includes. It makes no sense as a compile time argument.

PR #6269, opened 2021-05-12
view: commit transactions for foreign toplevel requests

RPigott:
These foreign toplevel actions require a transaction, but currently none are triggered. The ipc/binding fullscreen/focus commands already have their transactions triggered later.

To see the issue, send a foreign fullscreen request for your terminal and observe it won't fullscreen until you disturb something to commit a transaction.


emersion (approved):
Thanks!

PR #6278, opened 2021-05-17
Don't apply hide_edge_borders to any floating container

9ary:
This fixes the following scenario:
- Place a floating window so its border is right at the edge of the screen
- Create a new split
- The border disappears
- Moving the window does not restore the border


9ary:
Hey, did this fly under the radar or just lack of time? No pressure, but it'd be nice if someone could take a look.

Xyene (approved):
LGTM, thanks!

PR #6282, opened 2021-05-18
Avoid no-op output reconfiguring

### LINKED ISSUE #6280: Significant input latency when processing output command even if no-op
- **Sway Version:** sway version 1.5.1-eca57594 (May 18 2021, branch 'HEAD')
- **Debug Log:** https://gist.githubusercontent.com/pedrocr/7c8868dc86d947ecf4b2bbc7353749b8/raw/b1128c9c1179fb68ae03a86ef86ad36d42de8ff5/debug.log
- **Configuration File:** default
- **Description:**

There is something in sway that creates significant input latency when outputs get reconfigured and XWayland is running. To reproduce:
 
  - Start a new sway session with the default config
  - Run `swaymsg "output * dpms on"
  - Start a XWayland using app (chrome in the debug log)
  - Run `swaymsg "output * dpms on" again

The second time around includes restarting XWayland and that seems to cause significant input latency. Without XWayland there's no latency, with XWayland running there's significant input latency even if the output command is a no-op like this one.

I noticed this because I had the following swayidle setup:

```
exec swayidle -w before-sleep '$lock' \
  timeout 5 'if pgrep --uid $(whoami) swaylock; then swaymsg "output * dpms off"; fi'\
  resume 'swaymsg "output * dpms on"'\
```

This results in very noticeable latency when typing or moving the mouse at any moment after 5 seconds as it runs the `dpms on` command expecting it to be a no-op. It's common to see these kinds of warnings in the logs because of this:

```
[ERROR] client bug: timer event5 trackpoint: scheduled expiry is in the past (-91ms), your system is too slow
```

And the pointer lag is very noticeable because of this.

pedrocr:
When an output command doesn't actually change anything in the output config the processing will still all be done and cause visible artifacts like reconfiguring inputs. To fix this save the current per-output config in the output data and on configure, if the config is the same just return without doing anything else.

Fixes #6280

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -14,6 +14,7 @@
 #include "sway/tree/root.h"
 #include "log.h"
 #include "util.h"
+#include "stringop.h"
 
 int output_name_cmp(const void *item, const void *data) {
 	const struct output_config *output = item;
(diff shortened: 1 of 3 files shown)


sway/config/output.c
```diff
@@ -125,6 +125,67 @@ void merge_output_config(struct output_config *dst, struct output_config *src) {
...
+		oc->background_option = strdup(src->background_option);
+	}
+	if (src->background_fallback) {
+		oc->background_fallback = strdup(src->background_fallback);
+	}
+	oc->dpms_state = src->dpms_state;
+	return oc;
+}
+
+bool compare_linked_strings(const char *first, const char *second) {
```
emersion, line 164: I think `lenient_strcmp` does the same thing?

Xyene:
Hmm, is this actually enough?

I seems to me like this will break out if the output config has not changed, however this seems like it'd keep us from hitting

https://github.com/swaywm/sway/pull/6282/files#diff-66d6f9399c138857218e684475e604f3710684dd9deb896b66bf0ec2cacd149fR549

such that if an _input_ config was modified to contain a `map_to_output` directive, it would not be applied.

pedrocr:
I thought about that but if an input config has changed it was the input command that ran, not the output one. When that happens inputs gets reconfigured by this call:

https://github.com/swaywm/sway/blob/12e223e7973f7d7132d95c6302328067dec732ce/sway/commands/input.c#L93-L95

I checked and setting a new `map_to_output` works fine both as an isolated `swaymsg` command as well as in the config during a reload. If a new output shows up that needs to be mapped to then that output will be configured and the input reconfigure still gets triggered.

sway/config/output.c
```diff
@@ -397,6 +454,16 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 
 	struct wlr_output *wlr_output = output->wlr_output;
 
+	if (output->current_config && compare_output_config(oc, output->current_config)) {
```
kennylevinsen, line 457: This will lead to NULL pointer dereference if `oc` is NULL

pedrocr: calling `apply_output_config(NULL, output)` is valid? I've been running this for a while without issue but I guess the rest of the code is defensive about that so might as well be too.

sway/config/output.c
```diff
@@ -125,6 +126,63 @@ void merge_output_config(struct output_config *dst, struct output_config *src) {
 	}
 }
 
+struct output_config *duplicate_output_config(const struct output_config *src) {
```
emersion, line 129: Can we simplify this to the following?

- Create a new empty `dst` config
- Merge `src` into `dst`

sway/config/output.c
```diff
@@ -125,6 +126,63 @@ void merge_output_config(struct output_config *dst, struct output_config *src) {
...
+		oc->background_option = strdup(src->background_option);
+	}
+	if (src->background_fallback) {
+		oc->background_fallback = strdup(src->background_fallback);
+	}
+	oc->dpms_state = src->dpms_state;
+	return oc;
+}
+
+bool compare_output_config(const struct output_config *dst, const struct output_config *src) {
```
emersion, line 165: I'm not super thrilled about having yet another list of output config fields to maintain.

An alternative approach would be to have some kind of `config_dirty` field, set to true when changing a field. But finding all spots where we update a field doesn't sound fun either.

emersion:
I think we've addressed this in a different way. Please re-open if not.

PR #6286, opened 2021-05-20
Add `anything` option to grimshot

RagnarGrootKoerkamp:
This enables selecting a window, output, or area within one invocation, removing the need for multiple keybindings.
- Drag for an area (note that this already works in the `window` case)
- Click on a window to select it
- Click outside of all windows (e.g. in gaps or on the statusbar) to
select the output

Ideally this needs some testing with multiple monitors, which I don't have.

~The name `selection` may need to change.~ It's `anything` now.

### DIFF
--- a/contrib/grimshot.1
+++ b/contrib/grimshot.1
@@ -5,7 +5,7 @@
 .nh
 .ad l
 .\" Begin generated content:
-.TH "grimshot" "1" "2021-02-23"
+.TH "grimshot" "1" "2021-05-20"
 .P
 .SH NAME
 .P
@@ -92,6 +92,13 @@ captures it.\&
 \fIoutput\fR
 .RS 4
 Captures the currently active output.\&
+.P
+.RE
+\fIselection\fR
+.RS 4
+Allows manually selection a single window (by clicking on it), an output (by
+clicking outside of all windows, e.\&g.\& on the status bar), or an area (by
+using click and drag).\&
 .P
 .RE
 .SH OUTPUT
--- a/contrib/grimshot.1
+++ b/contrib/grimshot.1
@@ -94,7 +94,7 @@ captures it.\&
 Captures the currently active output.\&
 .P
 .RE
-\fIselection\fR
+\fIanything\fR
 .RS 4
 Allows manually selection a single window (by clicking on it), an output (by
 clicking outside of all windows, e.\&g.\& on the status bar), or an area (by
--- a/contrib/grimshot.1.scd
+++ b/contrib/grimshot.1.scd
@@ -67,7 +67,7 @@ _window_
 _output_
 	Captures the currently active output.
 
-_selection_
+_anything_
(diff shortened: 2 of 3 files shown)


Ashe:
This sounds great, can't wait to get this!

RagnarGrootKoerkamp:
@Ashe Note that you could already just take the updated `grimshot` file and put it somewhere in your path (at least that's what I did).

fluix-dev:
> Ideally this needs some testing with multiple monitors, which I don't have.

Though I'm not using grimshot myself, I stole the snippet (`swaymsg ... | slurp -o`) and have been using it for many months with multiple monitors and it works without issue.

RagnarGrootKoerkamp:
@fluix-dev Oh, I didn't know of `slurp -o` before. That simplifies things since now we can use the same `jq` command as the existing `window` option, and just do `slurp -o` to add the rectangles for all outputs instead of only `slurp`.

RagnarGrootKoerkamp:
`selection` could maybe change to e.g. `anything`

fluix-dev:
> Oh, I didn't know of slurp -o before

Oh haha, looks like I got a similar snippet from somewhere else.

> `selection` could maybe change to e.g. `anything`

Hm, yeah `anything` is probably better.

contrib/grimshot.1
```diff
@@ -94,6 +94,13 @@ captures it.\&
 Captures the currently active output.\&
 .P
 .RE
+\fIselection\fR
```
kmARC, line 97: This should also be renamed to "anything"

contrib/grimshot.1.scd
```diff
@@ -67,6 +67,11 @@ _window_
 _output_
 	Captures the currently active output.
 
+_selection_
```
kmARC, line 70: This should also be renamed to "anything"

edward-shen:
Is there anything blocking this PR from being merged, or does this just need a reviewer? This is one of the few things I really missed switching from x11 to sway.

RagnarGrootKoerkamp:
I think it's ready and has just fallen of the radar. I've been using it for over a year with problems and having a single screenshot button that does everything is super convenient indeed.

WhyNotHugo:
Clicking on the status bar did not work for me. Without gaps I don’t see how you can screenshot the whole output.

RagnarGrootKoerkamp:
Hmm, interesting. That would probably be an issue with `slurp`. I'm running version 1.3.2, and with it the following works:
```
echo 0,0 500x500 | slurp -o
```
- click in the top left corner of the output to select it
- click outside the top left corner of the output to select the entire outpu
- click and drag anywhere to select an area.

WhyNotHugo:
Mind you, I'm just pointing out an issue I've found; I'm not saying that this should be a blocker to merging.

emersion:

RagnarGrootKoerkamp:
See https://github.com/OctopusET/sway-contrib/pull/11

PR #6287, opened 2021-05-20
Change link for IRC support chat

apt-ghetto:
Changed the link address to point to libera.chat instead of freenode.net
Added a note to the landing page.

### DIFF
--- a/_layouts/master.html
+++ b/_layouts/master.html
@@ -26,7 +26,7 @@
               <a class="nav-link" href="https://github.com/swaywm/sway/wiki">Wiki ≫</a>
             </li>
             <li class="nav-item">
-              <a class="nav-link" href="https://webchat.freenode.net/?channels=sway&uio=d4">Support Chat ≫</a>
+              <a class="nav-link" href="https://kiwiirc.com/nextclient/irc.libera.chat:+6697/#sway">Support Chat ≫</a>
             </li>
           </ul>
         </div>


index.html
```diff
@@ -43,3 +43,10 @@ <h1><img src="/assets/logo.png" id="logo" alt="Sway" /></h1>
     </p>
   </div>
 </div>
+<div class="row" style="margin-top: 1rem;">
+  <div class="col-md-12">
+    <p>
+      NOTE: The IRC channel #sway has been moved from Freenode.net to <a href="https://libera.chat">Libera Chat</a>.
```
fluix-dev, line 49:
```suggestion
      NOTE: The IRC channel #sway has been moved from freenode to <a href="https://libera.chat">Libera Chat</a>.
```
I believe this is the official capitalization and name. Also not really sure if this notice is necessary.

apt-ghetto: I am also not sure, if this is needed or not. I am not against dropping it.

I have added this only to mention the name of the channel.
With the old link to the webchat, the correct channel was part of the link.
The new link opens just the landing page of Libera Chat. New users might not know the correct channel name.
Another possibility is to change the link in the other file to something like "Support Chat (channel #sway) >>"

progandy:
Maybe use the upstream kiwiirc as a webchat link: ```https://kiwiirc.com/nextclient/irc.libera.chat:+6697/#sway```
You need to register your nick to access the channel, though.

The irc server has to be changed in all README files, the wiki and the github issue templates as well.

_layouts/master.html
```diff
@@ -26,7 +26,7 @@
               <a class="nav-link" href="https://github.com/swaywm/sway/wiki">Wiki ≫</a>
             </li>
             <li class="nav-item">
-              <a class="nav-link" href="https://webchat.freenode.net/?channels=sway&uio=d4">Support Chat ≫</a>
+              <a class="nav-link" href="https://kiwiirc.com/nextclient/irc.libera.chat:+6697/#sway">Support Chat ≫</a>
```
emersion, line 29: Hm, I'm not comfortable redirecting to kiwiirc.com, a third-party bridge. I'd rather wait for Libera Chat to gain an official web client.

I don't have many other ideas. Ideally we'd be using `ircs://irc.libera.chat/#sway`, but users without an IRC client installed will be confused.

Regardless, it would be a good idea to add `title="#sway on Libera Chat"`.

progandy: Maybe split it to the irc link plus a webchat link, or create a landing page similar to matrix.to.

Another option would be to host your own instance of The Lounge or Kiwi IRC.

apt-ghetto: I would say, there are different questions to resolve:

- How to inform users about possible ways to get help?
  - Currently, there is only one link to the "wrong" IRC channel. As far as I can tell, the channel is still available, but I don't know how many of the developers/supporters are still there.
  - There is also #sway-devel and it would be nice, if it can be found easier
  - Are there other (bridged) channels like Matrix or Telegram?
  - Any forums?
- Which webclient should be used?

Whatever decisions you take, I can update the PRs. Or you can reject them, if it is easier for you.

emersion: The fact that we've moved from Freenode to Libera Chat doesn't really change anything else. In other words, users can get help on IRC, and we have no other official discussion medium.

I'd rather not mention #sway-devel in the READMEs, as it's not for end users. It's already mentioned in CONTRIBUTING.md.

There is no official Libera Chat web client yet, and I'm not comfortable recommending a third-party one, because any third party web client is basically a MITM. I don't think I'm ready to start hosting a public web client myself.

Maybe we should just link to ircs:// for now, and switch to the web client once Libera Chat have one.

emersion: Maybe we should explicitly spell out the channel and network in the link text, e.g. "Support chat: #sway on Libera Chat".

emersion: The matter is becoming [more pressing](https://mastodon.sdf.org/@kline/106299403921451814), so I've pushed an updated commit. Thanks!

emersion:
Bleh, I messed up my push, and pushed directly to the destination branch.

This has been merged, even if the PR is closed.

PR #6301, opened 2021-05-31
Only defer seat attachment during initial startup

### LINKED ISSUE #6048: Input device ends up in wrong seat after reloading sway config
- **Sway Version:**
  - `sway version 1.5-4e02a598 (Feb 19 2021, branch 'master')`
  - wlroots `0.12.0.r345.gf3758d1d`

- **Debug Log:**
  - https://gist.github.com/FreeFull/7c626da5128d229a05e88a125482cae2

- **Configuration File:**
Default configuration, except with the following lines added:
```
seat tablet attach "1386:891:Wacom_One_by_Wacom_M_Pen"

input "1386:891:Wacom_One_by_Wacom_M_Pen" {
    map_to_output eDP-1
    map_from_region 0.0x0.0502196194 1.0x.9497803806
}
```

- **Description:**
When sway is started with the given configuration, everything works as expected. However, when sway's config is then reloaded with Super+Shift+c, the tablet pen device ends up in seat0 instead, and all programs that were running inside the sway session close/crash (including programs that aren't running under Xwayland). Further reloading won't make programs running inside sway crash again, unless the tablet is put into its own seat manually with `swaymsg seat tablet attach "1386:891:Wacom_One_by_Wacom_M_Pen"` and the config gets reloaded again.

- Output of `swaymsg -t get_seats` before and after reloading the config:
https://gist.github.com/FreeFull/6789a4ee8aa2d17737cd0c9d719f4c54

vilhalmer:
Deferred commands are only run once, during sway startup. This means that deferring seat attachment based on whether we are reading the config prevents devices from being reattached to the correct seat during a config reload. Instead, only defer if the config is not yet active.

Fixes #6048.

It sort of feels like there's still something more fundamental going on here, but this at least fixes the crash and gets devices where they're supposed to be. I think there might be some race with the device getting added to its intended seat before being removed from seat0 (or whatever fallback) based on how they happen to be iterated. It also seems like we tend to apply the config more times to a single device than is necessary based on the debug logs from the original issue. Finally, I'm also not sure why we can't set up seat configs during initial start given that they should all be referencing only the identifier strings. I'm not familiar enough with the seat code to make much more headway on any of these at the moment though.

I tested:
* Device is attached to the correct seat if present at sway startup.
* Device remains attached to the correct seat after reload, and nothing crashes.

### DIFF
--- a/sway/commands/seat/attach.c
+++ b/sway/commands/seat/attach.c
@@ -12,7 +12,7 @@ struct cmd_results *seat_cmd_attach(int argc, char **argv) {
 	if (!config->handler_context.seat_config) {
 		return cmd_results_new(CMD_FAILURE, "No seat defined");
 	}
-	if (config->reading) {
+	if (!config->active) {
 		return cmd_results_new(CMD_DEFER, NULL);
 	}
 


emersion (approved):
Thanks!

PR #6302, opened 2021-06-01
Iterate over subsurfaces below the parent surface

emersion:
Update for the breaking change in [1].

[1]: https://github.com/swaywm/wlroots/pull/2948


kennylevinsen (approved):
LGTM

Issue #6285, opened 2021-05-19
Consider bumping `RLIMIT_NOFILE`


emersion:
We're handling many client connections and many DMA-BUFs.

See http://0pointer.net/blog/file-descriptor-limits.html

Don't forget about the last bullet point.

PR #6306, opened 2021-06-02
Restore SIGPIPE before exec:ing

dnkl:
Sway ignores `SIGPIPE` (by installing a `SIG_IGN` handler), in order to [prevent IPC from crashing Sway](https://github.com/swaywm/sway/blob/9755684fb0fd665a65be2a3cbabc5d502244c459/sway/main.c#L356).

`SIG_IGN` handlers are the *only* signal handlers inherited by sub-processes. As such, we should be a good citizen and restore the `SIGPIPE` handler to its default handler.

One way to reproduce is to bind `foot bash` in Sway's config:
```
bindsym $mod+t exec foot bash
```
Activate the key combo, and then run e.g. this program:
```c
#include <stdio.h>
#include <signal.h>

int
main(int argc, const char *const *argv)
{
    for (int i = 1; i < 32; i++) {
        struct sigaction sig;
        if (sigaction(i, NULL, &sig) < 0)
            return 1;
        if (sig.sa_handler != SIG_DFL)
            printf("signal #%d is NOT the default handler\n", i);
    }

    printf("done\n");
    return 0;
}
```
The foot+bash combo is important, because neither foot, nor bash restores signal handlers they themselves haven't touched. Zsh and fish seem to reset all signals, thus hiding the issue.

Original bug report: https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1806907.html

Alternatively, we can loop **all** signals and reset their handlers to `SIG_DFL`. I thought it slightly cleaner to only reset the signals actually touched by Sway. But doing so opens up for the same thing happening again in the future, if Sway starts ignoring more signals. I'm fine with both solutions.

Finally, the same issue exists in swaynag. However, that fork+exec doesn't even bother to restore the signal mask. I can open up a separate PR to fix swaynag, it that's something that's wanted?

### DIFF
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -57,6 +57,7 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 		sigset_t set;
 		sigemptyset(&set);
 		sigprocmask(SIG_SETMASK, &set, NULL);
+		signal(SIGPIPE, SIG_DFL);
 		close(fd[0]);
 		if ((child = fork()) == 0) {
 			close(fd[1]);
--- a/sway/config/bar.c
+++ b/sway/config/bar.c
@@ -217,6 +217,7 @@ static void invoke_swaybar(struct bar_config *bar) {
 		sigset_t set;
 		sigemptyset(&set);
 		sigprocmask(SIG_SETMASK, &set, NULL);
+		signal(SIGPIPE, SIG_DFL);
 
 		pid = fork();
 		if (pid < 0) {


emersion:
Sounds like a good idea! Ideally we'd be saving the original `sigaction` on startup, and restoring it to whatever it was set to.

emersion:
(At some point, we should probably add helpers to save/restore the state inherited by our children -- especially since we'll probably add more stuff, e.g. https://github.com/swaywm/sway/issues/6285.)

ijackson:
Hi.  I'm the author of the program `dgit` which detected and complained about this situation, as reported in the original Debian bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=989274

> Sounds like a good idea! Ideally we'd be saving the original `sigaction` on startup, and restoring it to whatever it was set to.

I don't think that is necessary for `SIGPIPE`.  sway is entitled to assume that it's `DFL` on entry, just like sway's children are entitled to assume that.  Executing a nonconsenting program with a non-default `SIGPIPE` disposition is not a good idea :-).  I think the same applies to the other sigaction flags, so using plain `signal` is fine.

> (At some point, we should probably add helpers to save/restore the state inherited by our children -- especially since we'll probably add more stuff, e.g. #6285.)

Definitely!

emersion (approved):
Alright, sounds like we don't need to care too much about the original sigaction then.

Thanks!

PR #6313, opened 2021-06-07
Add support for linux-dmabuf surface feedback

emersion:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/1376

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -13,10 +13,12 @@
 #include <wlr/types/wlr_compositor.h>
 #include <wlr/types/wlr_data_control_v1.h>
 #include <wlr/types/wlr_drm_lease_v1.h>
+#include <wlr/types/wlr_drm.h>
 #include <wlr/types/wlr_export_dmabuf_v1.h>
 #include <wlr/types/wlr_gamma_control_v1.h>
 #include <wlr/types/wlr_idle.h>
 #include <wlr/types/wlr_layer_shell_v1.h>
+#include <wlr/types/wlr_linux_dmabuf_v1.h>
 #include <wlr/types/wlr_pointer_constraints_v1.h>
 #include <wlr/types/wlr_primary_selection_v1.h>
 #include <wlr/types/wlr_relative_pointer_v1.h>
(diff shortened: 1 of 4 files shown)


sway/server.c
```diff
@@ -9,14 +9,17 @@
 #include <wlr/backend/multi.h>
 #include <wlr/backend/session.h>
 #include <wlr/config.h>
+#include <wlr/render/interface.h>
```
bl4ckb0ne, line 12: I don't think this one is required

sway/tree/container.c
```diff
@@ -1058,6 +1072,75 @@ static void set_fullscreen(struct sway_container *con, bool enable) {
...
+	}
+
+	if (!con->pending.workspace || !con->pending.workspace->output) {
+		return;
+	}
+
+	struct sway_output *output = con->pending.workspace->output;
+	struct wlr_output *wlr_output = output->wlr_output;
+
+	// TODO: add wlroots helpers for all of this stuff
```
bl4ckb0ne, line 1092: Would be nice to have an issue on wlroots side to track it

sway/tree/container.c
```diff
@@ -1058,6 +1072,75 @@ static void set_fullscreen(struct sway_container *con, bool enable) {
...
+
+	const struct wlr_drm_format_set *renderer_formats =
+		wlr_renderer_get_dmabuf_texture_formats(server.renderer);
+	assert(renderer_formats);
+
+	int renderer_drm_fd = wlr_renderer_get_drm_fd(server.renderer);
+	int backend_drm_fd = wlr_backend_get_drm_fd(wlr_output->backend);
```
bl4ckb0ne, line 1099: Aren't those the same?

bl4ckb0ne (commented):
Just a few questions, otherwise LGTM

sway/tree/container.c
emersion: Indeed, opened https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3343

sway/tree/container.c
emersion: Not necessarily:

- The renderer FD points to the DRM render node, the backend FD points to the primary node.
- On split render/display setups, backend FD points to the display-only device, renderer FD points to the render-only FD (well… not yet, but soon)
- When the user overrides the rendering device (`WLR_RENDER_DRM_DEVICE`), then they might not match.
- When we'll support multi-GPU scanout (https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3331), they might not match.

sway/server.c
emersion: Done!

sway/tree/container.c
bl4ckb0ne: Ah yes, the infamous split render/display setup, forgot about that one.

bl4ckb0ne (approved):
LGTM, thanks :)

PR #6323, opened 2021-06-12
add persian README

AliChraghi:


emersion:
Thanks. I'll wait for another native Persian speaker to approve this before merging.

siruidops (approved):
I'm native Persian speaker. I approve this text

emersion:
Nice, thanks. @AliChraghi, can you add a link in the main README to the Persian translation?

mhdzli (changes requested):
You should use `zws` anywhere you used a separated `ها` or `می` and add html tags to set `rtl` text direction

PR #6331, opened 2021-06-17
Disable auto_back_and_forth for next_on_output

### LINKED ISSUE #6299: Unexpected `next_on_output` behavior with `workspace_auto_back_and_forth`
- **Sway Version:** 1.6

- **Debug Log:** [here](https://gist.github.com/RagnarGrootKoerkamp/4b7b56a9c1737a4c2230d092b09cd641)

- **Configuration File:**
```
set $mod Mod4
workspace_auto_back_and_forth yes
bindsym $mod+Tab workspace next_on_output

bindsym $mod+Return exec foot
bindsym $mod+1 workspace number 1
bindsym $mod+2 workspace number 2
bindsym $mod+3 workspace number 3
bindsym $mod+4 workspace number 4
bindsym $mod+5 workspace number 5
```

- **Description:**
- Set `workspace_auto_back_and_forth` to `yes`
- Execute `swaymsg 'workspace next_on_output'` on an output with only a single workspace. Instead of doing nothing, this will focus the last active workspace, which may have been on a different output (or could be a now empty workspace on the same workspace). Note that the last active workspace is completely ignored when there is more than one active workspace on the current output.

Expected behavior: `next_on_output` does nothing when a workspace is the only active workspace on its output, regardless of the `workspace_auto_back_and_forth` setting.

Actual behavior: `next_on_output` switches to the last active workspace, which can be on a different output or an empty workspace.

RagnarGrootKoerkamp:
Fix #6299 

This forces `no_auto_back_and_forth` to true for `workspace next_on_output` and `workspace prev_on_output` to keep parity with i3.
In i3, running `next_on_output` never changes focus to another output.
In Sway currently, with `workspace_auto_back_and_forth` set to `yes`, running `next_on_output` on an output with only a single active workspace will typically end up focussing the other output:
1. `next_on_output` focusses the current workspace, because it's the only
one
2. `auto_back_and_forth` focusses the last focussed workspace, because the
workspace to focus is the current one. This will usually be on the other monitor if the workspace there was last focused one

Looking at the code, I actually think it may be cleaner to not make `no_auto_back_and_forth` an argument of `workspace_switch` (which is only called from 1 place anyway). Instead the back-and-forth could be resolved in a separate function call from the workspace switching, since it's really only needed for switching to a named/numbered workspace, and for explicit calls to `workspace back_and_forth`. (But this would be a larger code change, so didn't do that for now.)

### DIFF
--- a/sway/commands/workspace.c
+++ b/sway/commands/workspace.c
@@ -215,8 +215,10 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 			ws = workspace_by_name(argv[0]);
 		} else if (strcasecmp(argv[0], "next_on_output") == 0) {
 			ws = workspace_output_next(current, create);
+			no_auto_back_and_forth = true;
 		} else if (strcasecmp(argv[0], "prev_on_output") == 0) {
 			ws = workspace_output_prev(current, create);
+			no_auto_back_and_forth = true;
 		} else if (strcasecmp(argv[0], "back_and_forth") == 0) {
 			if (!seat->prev_workspace_name) {
 				return cmd_results_new(CMD_INVALID,


emersion (approved):
Seems like this patch has the intended effect. Thanks!

> Looking at the code, I actually think it may be cleaner to not make no_auto_back_and_forth an argument of workspace_switch

Agreed. PRs welcome :)

PR #6332, opened 2021-06-18
Move auto_back_and_forth logic out of workspace_switch

RagnarGrootKoerkamp:
Followup on #6331

- Extract the auto back and forth code to a separate `workspace_auto_back_and_forth` function.
- Rename/repurpose `no_auto_back_and_forth` to `auto_back_and_forth` for easier understanding
- Only call `workspace_auto_back_and_forth` in the 2 cases it's actually needed, instead of disabling it for all other cases.

I had a closer look at the [i3 implementation](https://github.com/Airblader/i3-original/blob/next/parser-specs/commands.spec#L122-L136) now, and they only use the flag for changing to a named/numbered workspace, so I've made it similar here. Actually, `man 5 sway` already documents it this way as well.

Note that in both sway and i3 providing the flag is allowed in all cases. It's just ignored in some of them.

### DIFF
--- a/sway/commands/workspace.c
+++ b/sway/commands/workspace.c
@@ -178,9 +178,9 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 				"Can't switch workspaces while fullscreen global");
 		}
 
-		bool no_auto_back_and_forth = false;
+		bool auto_back_and_forth = true;
 		while (strcasecmp(argv[0], "--no-auto-back-and-forth") == 0) {
-			no_auto_back_and_forth = true;
+			auto_back_and_forth = false;
 			if ((error = checkarg(--argc, "workspace", EXPECTED_AT_LEAST, 1))) {
 				return error;
 			}
@@ -215,10 +215,10 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 			ws = workspace_by_name(argv[0]);
 		} else if (strcasecmp(argv[0], "next_on_output") == 0) {
 			ws = workspace_output_next(current, create);
-			no_auto_back_and_forth = true;
+			auto_back_and_forth = false;
 		} else if (strcasecmp(argv[0], "prev_on_output") == 0) {
 			ws = workspace_output_prev(current, create);
-			no_auto_back_and_forth = true;
+			auto_back_and_forth = false;
 		} else if (strcasecmp(argv[0], "back_and_forth") == 0) {
 			if (!seat->prev_workspace_name) {
 				return cmd_results_new(CMD_INVALID,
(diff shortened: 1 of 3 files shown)


sway/commands/workspace.c
```diff
@@ -209,16 +209,17 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 				ws = workspace_create(NULL, name);
 				free(name);
 			}
+			if(ws && auto_back_and_forth){
```
emersion, line 212: Style: missing space between `if` and `(`, missing space betwen `)` and `{`

sway/tree/workspace.c
```diff
@@ -572,14 +572,18 @@ bool workspace_switch(struct sway_workspace *workspace,
...
+	if (config->auto_back_and_forth && active_ws &&
+			active_ws == workspace && seat->prev_workspace_name) {
 		struct sway_workspace *new_ws =
-			workspace_by_name(seat->prev_workspace_name);
-		workspace = new_ws ?
-			new_ws :
-			workspace_create(NULL, seat->prev_workspace_name);
+				workspace_by_name(seat->prev_workspace_name);
+		workspace = new_ws ? new_ws
+						   : workspace_create(NULL, seat->prev_workspace_name);
```
emersion, line 580: Style: don't align, instead use one indentation level for continuation lines.

sway/commands/workspace.c
```diff
@@ -233,11 +234,14 @@ struct cmd_results *cmd_workspace(int argc, char **argv) {
 				ws = workspace_create(NULL, name);
 			}
 			free(name);
+			if(ws && auto_back_and_forth){
```
emersion, line 237: Ditto

emersion (changes requested):
LGTM apart from the style issues!

emersion:
Can you squash these two commits together, so that the Git history is clean?

sway/tree/workspace.c
RagnarGrootKoerkamp: Fixed. But `clang-format` insists that the aligned way of formatting is the right one, so maybe there is a setting to change that. (Running it on the entire file makes lots of other diffs as well.)

sway/commands/workspace.c
RagnarGrootKoerkamp: Seems that my editor didn't automatically pick up the `.clang-format`. I should fix that. (I was going to suggest adding one, only to realize it's already there :+1: )

emersion: Ah, right, I even forgot we had one. It's probably outdated and not complete, PRs welcome to improve it.

emersion (approved):
LGTM, thanks!

PR #6335, opened 2021-06-18
Update .clang-format

RagnarGrootKoerkamp:
I had a look at the `.clang-format` file. It seems that the current code isn't formatted very consistently. In particular, the `ContinuationIndentWidth` is usually 8 when continuing if statements (so that there is an indent relative to the following block), but 4 in other places. I don't think clang-format supports it that way though, so I chose to put it to 8 here.

Also, `BinPack{Arguments,Parameters}` were both `false`, but in practice arguments are packed almost everywhere, so I've changed those to `true`.

`AllowAllArgumentsOnNextLine: false` makes sure to prefer the current practice of
```
f(a,b,
 c);
```
over
```
f(
 a,b,c);
```
See [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) for the documentation and examples of all options.

To see some diffs, I have added formatted `tree/{container,workspace}.c` with these settings in this PR.
The biggest changes are wrapping of lines over 80 chars, and better packing of function arguments on the first line when some are wrapped to a second line.

Anyway, I don't want to force anything on you, so feel free to ignore this.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -8,9 +8,11 @@ IndentCaseLabels: false
 SortIncludes: false
 ColumnLimit: 80
 AlignAfterOpenBracket: DontAlign
-BinPackParameters: false
-BinPackArguments: false
+BinPackParameters: true
+BinPackArguments: true
 ContinuationIndentWidth: 8
 AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortLoopsOnASingleLine: true
 ReflowComments: false
+AllowAllArgumentsOnNextLine: false
+AlignOperands: DontAlign


emersion:
Can you drop the second commit? It worsens readability overall. Would be nice to have a clang-format option to keep existing continuation line breaks.

RagnarGrootKoerkamp:
Sure, I can drop the 2nd commit applying the formatting.
Clang-format seems to be mostly all-or-nothing, although I think it's also possible (not sure how common though) to run it only on changed parts of a file.

Not sure how useful the change to the .clang-format will be in this case, but at least it's a bit closer to the current practice than before.

emersion (approved):
Right. I'm not sure if/how `.clang-format` is used by contributors TBH.

I wonder if another tool would be a better fit.

In any case, let's merge this for now. Thanks!

Xyene:
For what it's worth, I've been happily using clang-format in vim for years when working on other projects. I kind of made my peace with not being able to really use it for Sway, but this gets us closer to that so +1 from me.

PR #6336, opened 2021-06-18
Implement surface-suspension-v1

emersion:
Depends on: https://github.com/swaywm/wlroots/pull/2978


Kommynct:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/99 was closed, this should probably be as well.

PR #6338, opened 2021-06-20
Fix wlroots symbol collision

kennylevinsen:
As plans are to temporarily revert our solution to symbol collisions (https://github.com/swaywm/wlroots/pull/2980), we'll have to instead work around the problem locally.

If we make a 1.6.1 for wlroots 0.14 support, we should cherry-pick this PR as well.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -38,3 +38,10 @@ tasks:
       cd sway
       meson configure build -Dxwayland=disabled
       ninja -C build
+  - build-static: |
+      cd sway
+      mkdir subprojects
+      ln -s ../../wlroots subprojects/wlroots
+      rm -rf build
+      meson build --default-library=static --force-fallback-for=wlroots
+      ninja -C build
--- a/common/util.c
+++ b/common/util.c
@@ -10,12 +10,6 @@
 #include "log.h"
 #include "util.h"
 
-uint32_t get_current_time_msec(void) {
-	struct timespec now;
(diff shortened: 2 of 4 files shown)


.builds/alpine.yml
```diff
@@ -38,3 +38,9 @@ tasks:
       cd sway
       meson configure build -Dxwayland=disabled
       ninja -C build
+  - build-static: |
+      cd sway
+      mkdir subprojects
+      ln -s ../../wlroots subprojects/wlroots
+      meson configure build --default-library=static
```
emersion, line 45: I think we need to force Meson to use the subproject, or else it'll pick up the one installed system-wide.

kennylevinsen: It appears that `meson configure` was unable to change dependency source, even with force flags. Now we wipe the build dir and set it up again, with a force-fallback for good measure.

emersion: Was `meson configure build/ --cleancache` not enough?

emersion: It doesn't seem like it's enough.

kennylevinsen: Same conclusion in my attempts (https://builds.sr.ht/~kennylevinsen/job/531097#bottom). :/

emersion (approved):
Not super happy with re-building the whole wlroots to switch to static libs. Opened https://github.com/mesonbuild/meson/issues/8922 to see if we can easily improve this without having to introduce a custom Sway option.

But this is good enough for now. Thanks!

PR #6354, opened 2021-06-30
Change 'Danish' to 'Dansk' in README

jgbyrne:
Danish is the only language in the README that is not named in its own language.

<sorry for the frivolity but the inconsistency bugs me>


emersion (approved):
Thanks!

PR #6360, opened 2021-07-01
add drag_mode command to move/resize window without floating_modifier

PJungkamp:
I added a drag_mode command which causes any BTN_LEFT or BTN_RIGHT to resize/move regardless of the floating modifier. This makes it possible to e.g. toggle a resize mode which supports touch resizing on a tablet when no keyboard is connected (my usecase).

Documentation is still missing. But I want to get Feedback on whether this is a wanted addition.


PJungkamp:
As I don't use sway on any of my machines anymore, so I don't want to leave this PR lingering around for longer as there was pretty much no interest in it.

PR #6365, opened 2021-07-03
container: ignore borders in fullscreen windows

### LINKED ISSUE #6080: Incorrect monitor geometry/refresh rates sent to floating mpv windows in multi-monitor configurations.
It seems like floating windows in sway can receive incorrect monitor information. It should happen to any wayland client, but mpv is a pretty easy way to see it go wrong.

- **Sway Version:** 1.5-d358aab8 (git master)
- **Debug Log:** https://0x0.st/-Pup.log
- **Configuration File:** Everything default, you just need `for_window [app_id="mpv"] floating enable` in there.
- **Description:**
  - Have a multi-monitor setup; preferably with different resolutions and refresh rates.
  - Launch an mpv window (it should be floating as per above config).
  - Set to fullscreen (press 'f').
  - Press "i" to look at the stats menu and observe badly rendered text (this is because mpv is told the resolution/dpi of the other monitor).
  - Observe that the reported display fps is incorrect (if you have two monitors, it will report the fps of the other one; no clue what happens if you have more).

I bisected this and it actually started with #5812 (ironic because that actually fixed another bug easily seen with mpv). It looks like the xdg_shell logic has changed since then but the issue still occurs. I'm not aware of the details but something `view_update_size` does in `handle_commit`[¹](https://github.com/swaywm/sway/blob/master/sway/desktop/xdg_shell.c#L302) seems to cause this. This only happens with floating windows. Nothing else is affected.

Dudemanguy:
~In the special case of fullscreening a floating window,
container_discover_outputs would detect that the container box
intersected with a sway_output that the window wasn't actually over.
This would then send bogus surface entrance events to clients which
would confuse them. The cause of this appears to be con_box using
x/y/width/height instead of content_x/content_y (and so on). Just the
"plain" x/y could return negative values at times which is clearly
incorrect. Since output_get_box uses layout coordinates, it makes more
sense to properly use the layout coordinates for the container as well
when we are comparing the geometry between the two. Fixes #6080.~

~I'm unsure if using the `content`coordinates here has any deeper implications, but just running through the logic for this, I saw `content` being used basically in every step of the process except for this part. From my testing, it seemed to return the correct and expected coordinates at least.~

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -894,7 +894,7 @@ void container_set_geometry_from_content(struct sway_container *con) {
 	size_t border_width = 0;
 	size_t top = 0;
 
-	if (con->pending.border != B_CSD) {
+	if (con->pending.border != B_CSD && !con->pending.fullscreen_mode) {
 		border_width = con->pending.border_thickness * (con->pending.border != B_NONE);
 		top = con->pending.border == B_NORMAL ?
 			container_titlebar_height() : border_width;


emersion:
Hm, `x`/`y`/`width`/`height` are also supposed to be in layout coords, but include the container borders as well. So it's a little bit confusing that this patch fixes your issues.

I don't think we can use the content geometry here, since we want the titlebar texture to use the right scale.

Dudemanguy:
Actually, that explanation makes sense to me. I think what's happening is simply that the coordinates for borders are taken into account even for fullscreen windows. Because the actual content of the fullscreen window is exactly equal to the whole monitor, this means that the coordinates for the borders are "outside" of the `wl_output` somewhere (hence the negative values). In a multimonitor setup, this causes the window to intersect with another output, thus causing the issue. To confirm this, I just set the border to none on a floating window and sure enough it behaved normally. If I set the border to even 1 pixel, the issue comes back.

Would the right approach be to simply find wherever x/y/width/height are set and ignore any borders for fullscreen windows (i.e. they would be exactly equal to the content)?

Dudemanguy:
This simple patch does the trick.
```
diff --git a/sway/tree/container.c b/sway/tree/container.c
index b928d069..872ceb41 100644
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -900,10 +900,18 @@ void container_set_geometry_from_content(struct sway_container *con) {
 			container_titlebar_height() : border_width;
 	}
 
-	con->pending.x = con->pending.content_x - border_width;
-	con->pending.y = con->pending.content_y - top;
-	con->pending.width = con->pending.content_width + border_width * 2;
-	con->pending.height = top + con->pending.content_height + border_width;
+	if (!con->pending.fullscreen_mode) {
+		con->pending.x = con->pending.content_x - border_width;
+		con->pending.y = con->pending.content_y - top;
+		con->pending.width = con->pending.content_width + border_width * 2;
+		con->pending.height = top + con->pending.content_height + border_width;
+	} else {
+		con->pending.x = con->pending.content_x;
+		con->pending.y = con->pending.content_y;
+		con->pending.width = con->pending.content_width;
+		con->pending.height = con->pending.content_height;
+	}
+
 	node_set_dirty(&con->node);
 }
 ```

Does that make sense?

emersion:
Ah, that makes sense! Yes, this patch LGTM. Other functions (e.g. `view_autoconfigure`) also seem to ignore borders in fullscreen mode.

cc @kennylevinsen, in case I missed something.

emersion (approved):
LGTM, waiting for a bit in case somebody else has feedback. Please ping me again if this gets lost.

kennylevinsen (approved):
LGTM

emersion:
Thanks!

PR #6375, opened 2021-07-07
Do not VTs switch, block tty access

tmpm697:


emersion:
What's the motivation for this?

tmpm697:
Xorg has this: `https://wiki.archlinux.org/title/xorg#Block_TTY_access` and I find it useful as I often left my physical machine there for a while and I paranoid even in a minute someone can break into my machine :P

emersion:
Access to other TTYs doesn't mean third parties can bypass the auth. They still need to enter a password. Additionally, you can disable all VTs except the one you're using for Sway.

I'm very likely to just NACK this if there's no good justification. Feel free to keep that patch locally or in a fork ofc.

kennylevinsen:
> I often left my physical machine there for a while and I paranoid even in a minute someone can break into my machine :P

As a general rule of thumb, you have lost the moment someone gets physical access to your machine while it is running. You can at most inconvenience the adversary.

Likewise, as emersion mentioned, blocking VT access does absolutely nothing with respect to security. All that matters is if you left an open shell session running somewhere that people can change to without needing credentials.

If the adversary can break through an agetty prompt, they can break through your lock screen too. If the adversary can break through a *blank VT with nothing running on it*, then there's nothing you can do and I'm pretty sure their sheer skill assigns them immediate legal ownership of your computer, or anything else they want for that matter.

emersion:
Closing since there's no good reason to have this.

PR #6381, opened 2021-07-08
common/ipc-client: stop using --get-socketpath

emersion:
`sway --get-socketpath` just returns SWAYSOCK if set, and nothing
otherwise. Thus it does nothing more than the simple env var check
right above. Let's remove it.

### DIFF
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -18,38 +18,10 @@ char *get_socketpath(void) {
 	if (swaysock) {
 		return strdup(swaysock);
 	}
-	char *line = NULL;
-	size_t line_size = 0;
-	FILE *fp = popen("sway --get-socketpath 2>/dev/null", "r");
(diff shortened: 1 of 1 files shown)


common/ipc-client.c
```diff
@@ -18,38 +18,10 @@ char *get_socketpath(void) {
...
-			}
-			return line;
-		}
-	}
 	const char *i3sock = getenv("I3SOCK");
 	if (i3sock) {
-		free(line);
 		return strdup(i3sock);
 	}
-	fp = popen("i3 --get-socketpath 2>/dev/null", "r");
```
progandy, line 40: For i3 the call returns the value of the X11 atom ```I3_SOCKET_PATH```, so this second change does change the behavior.

emersion: Right.

PR #6384, opened 2021-07-10
Fix wl_pointer.frame not sent on touch emulation

emersion:
When emulating touch, the simulating_pointer_from_touch field is
set to true. It's switched back to false when a touch_up event is
received. However we need to ensure we always send a wl_pointer.frame
event following a group of other wl_pointer events.

Since a touch_frame event is always guaranteed to come after a group
of touch events, unset simulating_pointer_from_touch in the touch_frame
handler instead of the touch_up handler. Add a new field to know whether
the touch_frame handler should stop emulation.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -507,7 +507,7 @@ static void handle_touch_up(struct wl_listener *listener, void *data) {
 
 	if (cursor->simulating_pointer_from_touch) {
 		if (cursor->pointer_touch_id == cursor->seat->touch_id) {
-			cursor->simulating_pointer_from_touch = false;
+			cursor->pointer_touch_up = true;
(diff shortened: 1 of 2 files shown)


sway/input/cursor.c
```diff
@@ -473,7 +473,7 @@ static void handle_touch_down(struct wl_listener *listener, void *data) {
 	seat->touch_x = lx;
 	seat->touch_y = ly;
 
-	if (surface && wlr_surface_accepts_touch(wlr_seat, surface)) {
+	if (surface && wlr_surface_accepts_touch(wlr_seat, surface) && false) {
```
David96, line 476: I think this slipped in by accident?

emersion: Oops, good catch! It's a remnant from my testing session. Removed.

PR #6385, opened 2021-07-11
swaybar/swaynag: Add meson options to enable/disable

dylanaraps:
- Both options default to on (no change in behavior).
- Users can toggle them off.
- Not sure if an option should also be added for `swaymsg`, let me know.
- Files related to swaybar/swaynag are still compiled (`sway/commands/bar/` for example). Shall I also make these conditional?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -140,9 +140,15 @@ if scdoc.found()
 		'sway/sway-output.5.scd',
 		'swaybar/swaybar-protocol.7.scd',
 		'swaymsg/swaymsg.1.scd',
-		'swaynag/swaynag.1.scd',
-		'swaynag/swaynag.5.scd',
 	]
+
+	if get_option('swaynag')
+		man_files += [
+			'swaynag/swaynag.1.scd',
+			'swaynag/swaynag.5.scd',
+		]
+	endif
+
 	foreach filename : man_files
 		topic = filename.split('.')[-3].split('/')[-1]
 		section = filename.split('.')[-2]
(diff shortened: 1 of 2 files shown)


emersion:
Thanks for submitting a pull request!

It would be nice to make the new Meson options `feature` instead of `boolean`. That way we can also stop requiring dependencies such as cairo and pangocairo if swaybar/swaynag are disabled. However it sounds like sway itself still depends on these, so maybe can keep this for later.

meson.build
```diff
@@ -225,8 +234,12 @@ subdir('sway')
 subdir('swaymsg')
 
 subdir('client')
```
emersion, line 236: Maybe we can skip this subdir if neither swaybar nor swaynag is enabled?

meson.build
```diff
@@ -138,11 +138,20 @@ if scdoc.found()
...
+
+	if get_option('swaynag')
+		man_files += [
+			'swaynag/swaynag.1.scd',
+			'swaynag/swaynag.5.scd',
+		]
+	endif
+
+	if get_option('swaybar')
+		man_files += [ 'swaybar/swaybar-protocol.7.scd' ]
```
emersion, line 152: The swaybar protocol is still supported by sway even if the swaybar client is disabled, so I think it makes sense to keep building this man page unconditionally.

dylanaraps:
Made the changes (except for using `feature`). Can also make this change if desired.

emersion (approved):
LGTM, thanks!

PR #6386, opened 2021-07-12
Fix #6297

### LINKED ISSUE #6297: hide_cursor disables mouse on keyboard input while gaming
- **Sway Version: 1.6**

- **Debug Log: <coming>**

- **Description:**
  - Play a video game while `hide_cursor` is enabled
  - Press mouse buttons and move the mouse (e.g. aim / shoot and look around)
  - Still pressing and moving the mouse, press a key on the keyboard (e.g. WASD for evading enemies)
  - The mouse now stops moving, and you can't look around until you release the key.

I suspect this has to do with XWayland, since I can't reproduce the bug in sway itself.

I am not alone with this issue: [reddit post](https://www.reddit.com/r/swaywm/comments/n7m5d2/a_script_to_disable_hide_cursor_whentyping_when/)

lelgenio:
This makes it so that `seat hide_cursor` no longer clears cursor focus when hidding.

Clearing focus casuses problems whenever keyboard and mouse are to be used in conjunction, like Games, video editors, 3D/CAD software.

A better alternative to removing the call to `wlr_seat_pointer_notify_clear_focus` would be to only run it when dwt is active, however that means mixing code related to `seat` and `input`.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -236,7 +236,6 @@ void cursor_update_image(struct sway_cursor *cursor,
 static void cursor_hide(struct sway_cursor *cursor) {
 	wlr_cursor_set_image(cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
 	cursor->hidden = true;
-	wlr_seat_pointer_notify_clear_focus(cursor->seat->wlr_seat);
 }
 
 static int hide_notify(void *data) {


emersion:
I don't like this approach. There's no reason why an invisible cursor should still give focus.

JustinLovinger:
@emersion hidden cursor clearing focus is my biggest issue with `seat hide_cursor`. It means I have to wiggle my cursor around on an element when I want to read a focus popup. It would be nice if that was an option at least.

kennylevinsen:
Mouse-over tooltips does behave in unfortunate ways with the current implementation (reading the xkcd alt texts is rather painful with a short timeout).

The question is whether that's worth a reduction in correctness of sorts.

JustinLovinger:
@kennylevinsen  if anything, I would argue the clearing focus is less "correct". If I click on an element, then unplug my mouse, I don't expect the element to "unclick". I don't see why an element should unfocus just because the device the focused it is hidden.

I understand why some users may want this behavior, but it seems like the less "correct" behavior to me.

kennylevinsen:
Pointer focus (a `wl_pointer.enter` event with no `wl_pointer.leave` event) means: The pointer from this particular seat is currently hovering over this particular surface. That statement cannot be true if there is no pointer, and I do not think there is a valid argument against it being correct to remove pointer focus.

It may be reasonable to consider a *hidden* pointer different from a non-existing one in order to improve some corner-case UX, but that would count as a (possibly justifiable) hack.

JustinLovinger:
@kennylevinsen is the pointer removed though? The option is called `hide_cursor`, not `remove_pointer`, and the `cursor_hide` function appears to make the pointer invisible by temporarily replacing the cursor image.

PR #6387, opened 2021-07-13
Move client helpers from common/ to client/

emersion:
The sway binary doesn't need to load images nor use the event loop.


mstoeckl:
It may be better to move them directly to their associated clients: `background-image.c` is only used by `swaybar`, and `loop.c` is only used by `swaynag`.

PR #6392, opened 2021-07-18
Add `output modeline` command

nowrep:


emersion:
What is the use-case for this?

nowrep:
> What is the use-case for this?

I need to use a mode with slightly adjusted values (from what cvt/gft generates), otherwise my GPU won't underclock memory in idle.
Also as someone else mentioned in the custom mode issue/pr, people might prefer mode created with `gft` instead of `cvt`.

sjnewbury:
> > What is the use-case for this?
> 
> I need to use a mode with slightly adjusted values (from what cvt/gft generates), otherwise my GPU won't underclock memory in idle.
> Also as someone else mentioned in the custom mode issue/pr, people might prefer mode created with `gft` instead of `cvt`.

This is even more important if you use AMD Overdrive since it doesn't lock the memory clock causing bad flickering with insufficient blanking.  For example, my TV reports a 3840x2160@50Hz mode, which keeps within the standard HDMI bandwidth constraints.  It also supports "Input Signal Plus" which then supports 3840x2160@60Hz which works fine.  But the 50Hz mode can't be used, while a cvt generated mode works fine.

@nowrep I do have a problem with the patch though.  After setting a custom mode, I can't then change mode again, without reloading sway!

nowrep:
> I do have a problem with the patch though. After setting a custom mode, I can't then change mode again, without reloading sway!

Should be fixed now

emersion:
Out of curiosity, does it help to switch the first `false` argument to `true` here?

https://github.com/swaywm/wlroots/blob/f2f3df9fb1ecc6d5d4fbda26ba5d57a8c04a3b62/backend/drm/drm.c#L1045

nowrep:
Not in my case.

```sh 
cvt 2560 1440 60
Modeline "2560x1440_60.00"  312.25  2560 2752 3024 3488  1440 1443 1448 1493 -hsync +vsync
```

```sh 
cvt -r 2560 1440 60
Modeline "2560x1440R"  241.50  2560 2608 2640 2720  1440 1443 1448 1481 +hsync -vsync
```

But I need

```sh
mode 248.227 2560 2608 2640 2720 1440 1443 1448 1521 +hsync -vsync
```

kennylevinsen:
Did you try with the reduced timing modeling?

The refresh rate difference between the reduced timing modeline and your custom modeline is negligible—58868mHz for reduced vs 58916mHz for custom—while the reduced timing modeline has a noticably lower pixel clock, which to sounds beneficial to your usecase.

emersion:
Ref https://github.com/swaywm/wlroots/pull/3130 which just switches it on unconditionally.

This implements reduced blanking timings v1 only, but if that's not big enough of a hammer CVT now has v2 and v3. Not sure how well-supported these are with current screen hardware.

nowrep:
> Did you try with the reduced timing modeline?

I did and it's not working - as in, gpu memory clock stays locked on max value.

I kinda don't get why not just allow the user to set the modeline (and make the user generate it as they see fit), instead of picking one algorithm and forcing it (it's also a lot more "unnecessary" code in wlroots). Changing how the algorithm generates the modeline fixes nothing, there will always be a case where it's not what the user wants.
You can set the modeline in most other wlroots based compositors (even rootston when the custom modes support landed in wlroots), so why not also in sway?

kennylevinsen:
> I kinda don't get why not just allow the user to set the modeline (and make the user generate it as they see fit), instead of picking one algorithm and forcing it (it's also a lot more "unnecessary" code in wlroots).

Because we absolutely need the "algorithms" to make the optimal modeline, and your case sounds more like a GPU driver bug. The reduced blanking modeline is clearly the better modeline for your usecase as it has a lower bandwidth requirement, so it not working sounds buggy.

If it's only a workaround for a fixable bug elsewhere, then the feature has no merit.

nowrep:
> Because we absolutely need the "algorithms" to make the optimal modeline

Isn't the optimal modeline supposed to come from the display device itself in EDID?

> If it's only a workaround for a fixable bug elsewhere, then the feature has no merit.

Of course it's amdgpu bug, but since it's been there at least since navi (or maybe even back to polaris?), it's unlikely to be fixed anytime soon. You can workaround it on Windows, on X11, on some Wayland compositors, but not sway.

If we are not workarounding buggy gpus/displays then why have a custom mode at all?

emersion:
> I kinda don't get why not just allow the user to set the modeline

Because modelines are annoying. I'd rather have the user input just the resolution and refresh rate.

emersion:
> If we are not workarounding buggy gpus/displays then why have a custom mode at all?

Because sometimes the screen will support a resolution or refresh rate which isn't advertised.

nowrep:
> Because sometimes the screen will support a resolution or refresh rate which isn't advertised. And that's not necessarily due to a buggy EDID.

Sure, and there still can be a situation where the screen won't accept the modeline generated by wlroots, but would accept a slightly different one.

kennylevinsen:
> Sure, and there still can be a situation where the screen won't accept the modeline generated by wlroots, but would accept a slightly different one.

We are not going to implement features for hypothetical problems.

If we can't avoid supporting modelines, it might happen, but trying CVT v2 or v3 modelines first would be a good start. also, please make sure that the GPU driver bug is properly filed for this.

sjnewbury:
The problem *is* the reduced blanking.  amdgpu uses dynamic power management, where the frequencies are adjusted according to the instantaneous bandwidth requirements and load.  This has to happen during blanking period to dynamically adjust memory clocks, if there is insufficient time during the blank interval the driver prohibits mclk DPM and locks the frequency to maximum unless OverDrive is enabled, whereupon the frequency changes even if it causes severe display artifacts.

If this is a bug, it's a bug in the concept of reduced blanking display modes.  AMD certainly have no other solution to offer to allow dynamic switching with active scanout, they're definitely aware of the issue.  The alternative is to disable DPM, which is what the driver usually does when the modeline is unsuitable.

kennylevinsen:
> The problem _is_ the reduced blanking.

We do not *yet* utilize reduced blanking, so if this was true there would not have been a problem. Is there a statement from an upstream bug report on this?

Correct me if I'm wrong, but I do not think @nowrep is trying to squeeze within bandwidth limitations to reach a higher mode like your case, but see the problem with both longer and shorter blanking periods than the particular mode.

kennylevinsen:
> If this is a bug, it's a bug in the concept of reduced blanking display modes.

No it would be a bug in the driver or firmware, or possibly even a hardware design flaw. It is a bug if the GPU fails to function correctly unless you stick to arbitrary modes designed around CRT display hardware limitatons. If the GPU can't handle a mode, it's driver should disallow it.

If it's a hardware design flaw like your comments suggest, then we probably end up having to support custom modelines, but it would be nice to have confirmation from amdgpu guys that this is something we need to work around in userspace.

sjnewbury:
> > If this is a bug, it's a bug in the concept of reduced blanking display modes.
> 
> No it would be a bug in the driver or firmware, or possibly even a hardware design flaw. It is a bug if the GPU fails to function correctly unless you stick to arbitrary modes designed around CRT display hardware limitatons. If the GPU can't handle a mode, it's driver should disallow it.
> 
> If it's a hardware design flaw like your comments suggest, then we probably end up having to support custom modelines, but it would be nice to have confirmation from amdgpu guys that this is something we need to work around in userspace.
> 
> After all, I'm sure everyone would agree that it's much nicer for things to just work than having to generate custom modelines (and for the community to have to deal with the associated support questions).

We are kind of stuck programming display hardware like it's a CRT though, like it or not.  This reminds me of when hard disks were still configured with CHS long after they ceased reflecting the actual hardware.   Maybe one day we'll stop pretending we're still using CRTs? ;-)

I really can't argue with your point about the lack of information from AMD around this issue. If it were down to me, I would have coded it to reject the mode instead of crippling the power management as you suggested.  I guess it comes down to "least surprise", having random display supported modes just not work could have been more of a support headache for AMD.  They can just pretend the issue is user error.

kennylevinsen:
Hmm, after doing some reading, making some observations and having some discussions, I'm leaning towards us actually needing this for the sole reason of monitor compatibility. It appears that the relevant EDID flags are poorly utilized, and as a monitor-incompatible mode likely just does bad things we might need it to be tunable.

I do still find it to be wrong for the sake of working around an amdgpu bug, and find such workarounds detrimental to ever getting root issues fixed.

> We are kind of stuck programming display hardware like it's a CRT though, like it or not.

Throughout all the monitors I've looked at EDID data from today, they all specifically advertise CVT-RB timings as part of their native modes, so while these Sharp, LG and Samsung panels are of course anecdotal, it does seem to indicate that the days of CRT timings at least are long gone. I wonder AMD fixed this issue on Windows or if they're ignoring monitor-specified reduced blanking timings and burning bandwidth.

emersion:
Yeah, after looking at the EDID bits it doesn't seem like we can do any useful auto-detection sadly… We _could_ look at the list of modes and see the versions of CVT used, but this is fragile and complicated.

So an option to set a custom modeline might be fine in the end, even if suboptimal.

emersion:
Maybe the syntax can be a bit different to be less ambiguous? Maybe something among these lines:

```
output HDMI-A-1 modeline "173.00 1920 2048 2248 2576 1080 1083 1088 1120 -hsync +vsync"
```

or `mode --line`?

RPigott:
If added, +1 for `output <output> modeline "..."` syntax.

nowrep:
When adding new command (`output modeline`), `commands/output/mode.c` would still need to touch the `output_config.drm_mode` to make the output_config merging work correctly, which would be weird?

Also, both commands will do the same thing (set output mode), it's just syntax that is different. So I think `mode --custom-line` is better, but your call of course.

sjnewbury:
> When adding new command (`output modeline`), `commands/output/mode.c` would still need to touch the `output_config.drm_mode` to make the output_config merging work correctly, which would be weird?
> 
> Also, both commands will do the same thing (set output mode), it's just syntax that is different. So I think `mode --custom-line` is better, but your call of course.

In that case, I would suggest calling it: "mode --custom-modeline", even though it's redundant, since that's what the string is known as and it removes all ambiguity as to what's expected.

kennylevinsen:
> When adding new command (output modeline), commands/output/mode.c would still need to touch the output_config.drm_mode to make the output_config merging work correctly, which would be weird?

I'd put them both into `commands/output/mode.c`. Similar commands often coexist - e.g. `bindcode`, `bindsym` and `bindswitch` all being in `commands/bind.c`. Nothing weird about that.

A separate subcommand makes the documentation simpler to write correctly (a --custom-modeline arg changes the format of the required positional argument, modeline is mutually exclusive with `--custom`, future arguments like `--timing-mode` will not have any effect when modeline is used), and results in nicer argument handling. It also seems like a fairly logical split: `mode` picks or generates a mode, `modeline` strictly specifies a pre-generated one.

emersion (approved):
LGTM, thanks!

PR #6396, opened 2021-07-21
Add `gaps <type> <scope> toggle <px>` command

bR3iN:
This command is part of i3-gaps since Airblader/i3#264 but is currently missing in sway.
It can be used to toggle gaps at runtime either via a keybinding or via `swaymsg`, e.g.:
```
bindsym $mod+g gaps inner current toggle 30
swaymsg gaps outer all toggle 15
```


emersion (approved):
LGTM, thanks!

PR #6399, opened 2021-07-23
Fix crash when starting without HOME

stoeckmann:
If HOME environment variable is not set, sway fails startup with a
segmentation fault due to null pointer dereference.

Also check calloc return value and only perform the fallback code when
really needed.

Signed-off-by: Tobias Stoeckmann <tobias@stoeckmann.org>


emersion (approved):
LGTM, thanks!

emersion:
Hm, CI is failing. Maybe you've based your patch off of an old checkout of Sway? Can you rebase?

stoeckmann:
Oops, sorry. I have rebased to master.

PR #6407, opened 2021-07-28
Don't send wl_pointer.motion from seatop_begin_default

nowrep:
The issue was introduced in 6ea45395c70939a6d855736cabfe75ad9cf4a0ae, followup 6b9a9b62462c0fae7b09294700112c569a3ccc19 helped but the issue can still be occasionally reproduced.

I don't fully understand the reasoning for 6ea45395c70939a6d855736cabfe75ad9cf4a0ae, but it talks about confine creation (so that would be on mouse button press?) and the problematic motion from seatop_begin_default is sent on mouse button release. @Xyene

You can see the issue in game here: https://www.reddit.com/r/linux_gaming/comments/osjtln/help_with_ffxiv_on_wayland_issue_explained_in/
Cannot be reproduced on weston.


Xyene:
Thanks for tracking this down. I'm not sure about the patch as-is (it appears to fix the symptoms, but not the cause), but having figured out how to make it not happen gets us most of the way there.

As far as I can tell, you're using `time_msec > 0` to check that we're coming from `seatop_begin_default`

https://github.com/swaywm/sway/blob/4e77bc293515ea2f00ebbc4395cb0eb4446d0195/sway/input/seatop_default.c#L792

and not sending a motion event if so. This doesn't seem quite correct to me. There is already a guard in wlroots that should prevent a duplicate motion event from being sent

https://github.com/swaywm/wlroots/blob/a93b18dbd5a43735709c59590fa0bff427dd888d/types/seat/wlr_seat_pointer.c#L233-L235

so what I think is happening is wlroots' cursor state diverges from Sway's during `seatop_down`, and "snapping back" into place when we go back to the default seatop. I'd take a closer look at

https://github.com/swaywm/sway/blob/6ffa4b1f7000cd90d63213820039ee432803e66a/sway/input/seatop_down.c#L41-L51

and try to figure out why the guard in wlroots isn't catching this.

nowrep:
> so what I think is happening is wlroots' cursor state diverges from Sway's during seatop_down, and "snapping back" into place when we go back to the default seatop. I'd take a closer look at

Yes, the issue only happens when the pointer motion handling goes to seatop_down.
The cursor locking seems to be delayed, so if I press the mouse button, then wait a second and then move the cursor, it will already be locked and won't go to seatop_down. If I press the mouse button and then immediately move the cursor, it will be handled by seatop_down and only then there is a chance the bug happens.

What seems to be wrong, as you said, is that the seatop_down will update the cursor position, but when the cursor lock happen, it won't be warped back to the committed cursor lock position. So when the mouse button is released, cursor will be rebased on position different from the cursor lock position.

So I tried another patch, and it seems to be working too (but I need more testing to be absolutely sure).

```diff
diff --git a/sway/input/cursor.c b/sway/input/cursor.c
index 2fe5b202..4242a55c 100644
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -837,6 +837,8 @@ static void handle_tool_button(struct wl_listener *listener, void *data) {
 		event->button, (enum zwp_tablet_pad_v2_button_state)event->state);
 }
 
+static void warp_to_constraint_cursor_hint(struct sway_cursor *cursor);
+
 static void check_constraint_region(struct sway_cursor *cursor) {
 	struct wlr_pointer_constraint_v1 *constraint = cursor->active_constraint;
 	pixman_region32_t *region = &constraint->region;
@@ -864,6 +866,10 @@ static void check_constraint_region(struct sway_cursor *cursor) {
 				cursor_rebase(cursor);
 			}
 		}
+
+		if (constraint->type == WLR_POINTER_CONSTRAINT_V1_LOCKED) {
+			warp_to_constraint_cursor_hint(cursor);
+		}
 	}
 
 	// A locked pointer will result in an empty region, thus disallowing all movement
```

Xyene:
Alright, just making sure I understand the sequence of events to reproduce this bug...

1. [seatop_default] user clicks at (x, y), button down event sent to application
2. [seatop_default] enter seatop_down
3. [seatop_down] user moves mouse around, e.g. to (x + 1, y + 1), application is notified of the move to (x + 1, y + 1)
4. [seatop_down] application gets around to processing the button press from 1), locks cursor to (x, y), sets constraint hint to (x, y)
5. [seatop_down] sway receives lock request, warps cursor back to (x, y), notifies application of the move
6. [seatop_down] application receives the motion event from 3) followed by the motion event from 5)

So far, nothing has gone wrong, is that right? Presumably, then,

7. [seatop_down] user moves mouse around, this generates relative motion events but not absolute motion events due to the clamping [here](https://github.com/swaywm/sway/blob/f4cda5157e40470da7a0e41685ff04a0d76e8be5/sway/input/cursor.c#L358-L376)
8. [seatop_down] user releases button, enter seatop_default
10. [seatop_default] we consider rebasing, but the cursor is still at (x, y) (it shouldn't have moved -- it was locked?) so we do nothing
10. [seatop_default] application releases lock, we consider warping to the hint of (x, y), but our cursor is already at (x, y) following 5), so we do nothing
11. [seatop_default] everything works?

Though of course, 11) doesn't actually happen. What am I missing in the flow of events here?

nowrep:
5) sway doesn't warp cursor back to (x, y), it stays at (x + 1, y + 1)
9) on rebase it sends move to (x + 1, y + 1)
10) here it warps back to (x, y)

nowrep:
And with the applied patch from my second comment, it goes:

5 - sway warp cursor back to (x, y)
9 - rebase noop, because cursor is already at (x, y)
10 - noop again

The problem is that sway won't warp the cursor when cursor gets locked.

Xyene:
> 5 - sway doesn't warp cursor back to (x, y), it stays at (x + 1, y + 1)

Thanks, I maybe-kind-of understand now. Well, not really -- I don't doubt you, but there is existing code that should be doing this already for both confines and locked regions:

https://github.com/swaywm/sway/blob/f4cda5157e40470da7a0e41685ff04a0d76e8be5/sway/input/cursor.c#L860-L862

Why is it not being hit?

(At any rate, I believe you should be using the locked region, not the hint, when deciding where to warp. The hint is optional, it just so happens that it's specified here.)

nowrep:
Application sets the constraint region as the whole window (I assume it's the Wine "lock cursor to window" feature).

So this condition https://github.com/swaywm/sway/blob/f4cda5157e40470da7a0e41685ff04a0d76e8be5/sway/input/cursor.c#L852-L853
is always false.

Xyene:
> Application sets the constraint region as the whole window (I assume it's the Wine "lock cursor to window" feature).

Thanks, I understand why we wouldn't be warping to (x, y) in 5) now; we're already in the region. I was assuming it was setting a 1x1 lock region at the point it got clicked.

This makes me confused about this bit, though:

> 5 - sway doesn't warp cursor back to (x, y), it stays at (x + 1, y + 1)
9 - on rebase it sends move to (x + 1, y + 1)
10 - here it warps back to (x, y)

Why is it sending a move on 9) then? If the motion to (x + 1, y + 1) had already been communicated to wlroots in 3), 9) should be a no-op here, since the cursor should not have moved.

nowrep:
> Why is it sending a move on 9) then? If the motion to (x + 1, y + 1) had already been communicated to wlroots in 3), 9) should be a no-op here, since the cursor should not have moved.

`seatop_down:handle_pointer_motion` only calls `wlr_seat_pointer_notify_motion` which just sends wl_pointer:motion to client, it doesn't update the cursor position (warp cursor).

Xyene:
`wlr_seat_pointer_notify_motion` ends up calling `wlr_seat_pointer_send_motion`, which has a guard

https://github.com/swaywm/wlroots/blob/a93b18dbd5a43735709c59590fa0bff427dd888d/types/seat/wlr_seat_pointer.c#L233-L235

before sending the motion. It also updates the internal state of the cursor,

https://github.com/swaywm/wlroots/blob/a93b18dbd5a43735709c59590fa0bff427dd888d/types/seat/wlr_seat_pointer.c#L246-L247

https://github.com/swaywm/wlroots/blob/a93b18dbd5a43735709c59590fa0bff427dd888d/types/seat/wlr_seat_pointer.c#L221-L224

nowrep:
Yes, I realized it's wrong too late after I sent the comment, sorry.

`wlr_cursor` and `wlr_seat::pointer_state` gets out of sync during `seatop_down`, presumably because this calculation is incorrect:

https://github.com/swaywm/sway/blob/f4cda5157e40470da7a0e41685ff04a0d76e8be5/sway/input/seatop_down.c#L45-L49

Replacing this code with 

```c
	double sx, sy;
	node_at_coords(seat,
			seat->cursor->cursor->x, seat->cursor->cursor->y, &surface, &sx, &sy);
```

as is done in `seatop_default` makes them both get in sync.

Xyene:
Oh, nice catch!

Out of curiosity, if you print out the values computed by the old code versus the new for the same motions, how much do they differ? I can't imagine they're too different, otherwise I'd have expected people to complain about things like their cursor not matching up with what they're drawing in e.g. GIMP?

Anyway, this seems like a (more) correct fix to me, conditional on you dogfooding it for a little while, since I don't have the means to reproduce the original bug 🙂. Remember to also change `handle_tablet_tool_motion`, since I just copy/pasted the pointer handler when writing that, and the original code long predates my involvement in this project.

nowrep:
Actually the real fix turned out to be much easier, it's just losing precision from doing `double -> int` conversion.

Xyene:
Ah, that's subtle, thanks!

As an aside, this also prompted me to write https://github.com/swaywm/wlroots/pull/3088. I don't know if it would have helped here or if the delta was too large, but it seemed worth doing.

Xyene (approved):
LGTM!

nowrep:
The delta was far too big for that change to matter - up to 1.0 pixels.

PR #6412, opened 2021-07-30
transaction: lock/unlock views

vyivel:
Supersedes https://github.com/swaywm/sway/pull/6362
Depends on ~https://github.com/swaywm/wlroots/pull/3081~, https://github.com/swaywm/wlroots/pull/3068 and on a to-be-written wlroots PR for Xwayland surface state caching support.

Notes:
- `wleird/slow-ack-configure` doesn't work anymore (it acks instantly), as the transaction system calls `wlr_surface_send_frame_done()` repeatedly to acquire the requested commit faster.
- There are still some rendering issues (just like in `master`) when a view's surface gets destroyed during a transaction. This will supposedly get fixed by https://github.com/swaywm/wlroots/pull/3024.

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -418,8 +418,9 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 	}
 
 	if (view->container->node.instruction) {
-		transaction_notify_view_ready_by_geometry(view,
-				xsurface->x, xsurface->y, state->width, state->height);
+		transaction_notify_view_acked_by_geometry(view,
+			xsurface->x, xsurface->y, state->width, state->height);
+		transaction_notify_view_ready(view);
 	}
 
 	view_damage_from(view);
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -388,12 +388,21 @@ static void get_geometry(struct sway_view *view, struct wlr_box *box) {
 	}
 }
 
+static void handle_cache(struct wl_listener *listener, void *data) {
+	struct sway_xwayland_view *xwayland_view =
+		wl_container_of(listener, xwayland_view, cache);
+	struct sway_view *view = &xwayland_view->view;
+	struct wlr_xwayland_surface *xsurface = view->wlr_xwayland_surface;
+	struct wlr_surface_state *cached = data;
+	transaction_notify_view_acked_by_geometry(view,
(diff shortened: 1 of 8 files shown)


sway/desktop/xwayland.c
```diff
@@ -388,12 +388,21 @@ static void get_geometry(struct sway_view *view, struct wlr_box *box) {
 	}
 }
 
+static void handle_cache(struct wl_listener *listener, void *data) {
+	struct sway_xwayland_view *xwayland_view =
+		wl_container_of(listener, xwayland_view, cache);
+	struct sway_view *view = &xwayland_view->view;
+	struct wlr_xwayland_surface *xsurface = view->wlr_xwayland_surface;
+	struct wlr_surface_state *cached = data;
+	transaction_notify_view_acked_by_geometry(view,
```
vyivel, line 397: Xwayland surfaces don't have `ack_configure` event, so we notify the transaction system here. Currently this is done incorrectly, as `xsurface->{x,y}` isn't cached. In the future, this should be changed to iterating over `cached->addons`, looking for corresponding `wlr_xwayland_surface` state, and taking the positing from there.

vyivel:
>wleird/slow-ack-configure doesn't work anymore (it acks instantly), as the transaction system calls wlr_surface_send_frame_done() repeatedly to acquire the requested commit faster.
This can be replaced by scheduling a frame on (every?) output. Not sure which is more suitable.

vyivel:
Right now the whole thing works just because the commit handlers calls damage iterator which schedules a frame even if the surface wasn't damaged.

vyivel:
For some reason this makes resizing `imv` unstable.

vyivel:
Even better, running `disobey-resize` froze my PC.

vyivel:
I'd say this is ready for a review (but not for a merge, not yet). I welcome everyone to test this branch with various layouts and clients.

Notes:
- Floating `imv` resizing doesn't work correctly, resulting to stutter. I believe this is caused by the fact that `imv` (and some other clients) acknowledges its configure and submits a new buffer in two different commits, making Sway think it's a client initiated resize. This *might* be related to https://github.com/swaywm/wlroots/issues/3026.
- Sway doesn't handle floating resize correctly (not adjusting the container according to the received state, but instead centering the surface inside the container, which doesn't make sense if it's floating). This also might cause issues with this branch.

Feel free to ask any questions.

PR #6413, opened 2021-07-30
Fallback to other output modes if preferred mode fails

emersion:
Sometimes the preferred mode is not available due to hardware
constraints (e.g. GPU or cable bandwidth limitations). In these
cases it's better to fallback to lower modes than to end up with
a black screen.

Tested with this wlroots patch:

```diff
diff --git a/backend/drm/drm.c b/backend/drm/drm.c
index cd161e02b777..550b450d68bf 100644
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -474,6 +474,10 @@ static bool drm_connector_test(struct wlr_output *output) {
 		}
 	}
 
+	if ((output->pending.committed & WLR_OUTPUT_STATE_MODE) && output->pending.mode->preferred) {
+		return false;
+	}
+
 	return true;
 }
```

Helps with some DP adapters.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -357,8 +357,26 @@ static void queue_output_config(struct output_config *oc,
 		set_mode(wlr_output, oc->width, oc->height,
 			oc->refresh_rate, oc->custom_mode == 1);
 	} else if (!wl_list_empty(&wlr_output->modes)) {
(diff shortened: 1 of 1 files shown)


RPigott (approved):
LGTM and works.

emersion:
Thanks for the review!

PR #6418, opened 2021-08-02
input: Use seatop_down on layer surface click

pinselimo:
This PR changes the seatop implementation to ```seatop_down``` after a layer-surface has been clicked.

## Issue

This solves an issue with layer-shell surfaces not receiving button release events. Especially when "mouse_warping container" is set.
On Waybar the issue leads to secondary interactions being necessary to "re-activate" the bar after changing workspaces with a mouse click.

However, the issue can also be triggered when moving the cursor away from the surface while being held pressed.

See https://github.com/Alexays/Waybar/issues/1047 for complaints and https://github.com/wmww/gtk-layer-shell/issues/109 for videos and a log. Maybe [this warning](https://codeberg.org/dnkl/foot/src/branch/master/input.c#L1773) from foot is also related.

## Proposed solution

The problem seems to be that on click the pointer is relocated and a different surface is focused immediately. Using ```seatop_down``` this re-focus is delayed until the button is released.
This is also why the problem does not occur if we do not use a layer-surface. Hence, the solution would be to make use of ```seatop_down```. The only additional changes necessary are null-pointer checks for the ```con``` field.

Thanks,
Simon

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -373,6 +373,9 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 			seat_set_focus_layer(seat, layer);
 			transaction_commit_dirty();
 		}
+		if (state == WLR_BUTTON_PRESSED) {
+			seatop_begin_down_on_layer_surface(seat, surface, time_msec, sx, sy);
+		}
 		seat_pointer_notify_button(seat, time_msec, button, state);
 		return;
 	}
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -374,7 +374,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 			transaction_commit_dirty();
 		}
 		if (state == WLR_BUTTON_PRESSED) {
-			seatop_begin_down_on_layer_surface(seat, surface, time_msec, sx, sy);
+			seatop_begin_down_on_surface(seat, surface, time_msec, sx, sy);
 		}
 		seat_pointer_notify_button(seat, time_msec, button, state);
 		return;
--- a/sway/input/seatop_down.c
+++ b/sway/input/seatop_down.c
@@ -10,8 +10,10 @@
 
 struct seatop_down_event {
 	struct sway_container *con;
+	struct sway_seat *seat;
+	struct wl_listener surface_destroy;
 	struct wlr_surface *surface;
-	double ref_lx, ref_ly;		 // cursor's x/y at start of op
+	double ref_lx, ref_ly;         // cursor's x/y at start of op
 	double ref_con_lx, ref_con_ly; // container's x/y at start of op
 };
 
@@ -71,11 +73,24 @@ static void handle_tablet_tool_motion(struct sway_seat *seat,
 	}
 }
 
+static void handle_destroy(struct wl_listener *listener, void *data) {
+	struct seatop_down_event *e =
+		wl_container_of(listener, e, surface_destroy);
+	if (e) {
+		seatop_begin_default(e->seat);
+	}
+}
+
 static void handle_unref(struct sway_seat *seat, struct sway_container *con) {
 	struct seatop_down_event *e = seat->seatop_data;
 	if (e->con == con) {
 		seatop_begin_default(seat);
 	}
+}
+
+static void handle_end(struct sway_seat *seat) {
+	struct seatop_down_event *e = seat->seatop_data;
+	wl_list_remove(&e->surface_destroy.link);
 }
 
 static const struct sway_seatop_impl seatop_impl = {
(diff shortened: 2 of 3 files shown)


sway/input/seatop_down.c
```diff
@@ -41,7 +41,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 static void handle_pointer_motion(struct sway_seat *seat, uint32_t time_msec) {
 	struct seatop_down_event *e = seat->seatop_data;
 	struct sway_container *con = e->con;
-	if (seat_is_input_allowed(seat, con->view->surface)) {
+	if (con && seat_is_input_allowed(seat, con->view->surface)) {
```
emersion, line 44: Hm. So this skips sending the motion events when we don't have a container? But it seems to me like we should still send the events to the layer surface?

pinselimo: True. Using ```!con ||``` would fix that but skip the check for an ```exclusive_client``` for layer surfaces. However, there is also no check in the initial button press event. Is it safe to assume a layer surface can always receive motion events if it was clicked?

sway/input/seatop_down.c
```diff
@@ -107,6 +107,8 @@ void seatop_begin_down(struct sway_seat *seat, struct sway_container *con,
 	seat->seatop_impl = &seatop_impl;
 	seat->seatop_data = e;
 
-	container_raise_floating(con);
+	if (con) {
+		container_raise_floating(con);
+	}
 	transaction_commit_dirty();
```
Xyene, line 113: I don't think we need to commit the transaction if we haven't raised the container. We should also leave a comment here reminding us that `con` is null when clicking a layer surface.

sway/input/seatop_down.c
```diff
@@ -74,7 +74,7 @@ static void handle_tablet_tool_motion(struct sway_seat *seat,
 
 static void handle_unref(struct sway_seat *seat, struct sway_container *con) {
 	struct seatop_down_event *e = seat->seatop_data;
-	if (e->con == con) {
+	if (!con || e->con == con) {
```
Xyene, line 77: I don't think `con` can ever be null here. The only caller is https://github.com/swaywm/sway/blob/acf946fe4c495b47c5a76230ba98f9b973bec96d/sway/tree/container.c#L1054

emersion: A layer surface has no `con`.

emersion: Ah, yeah. This is unref. Indeed, can't be null.

sway/input/seatop_down.c
Xyene: > However, there is also no check in the initial button press event.

That sounds like a bug, separate from this PR. There should be a check.

emersion: I wonder. My gut feelings says we should keep this check, maybe by replacing `e->con` with a `struct wlr_surface *`.

sway/input/seatop_default.c
```diff
@@ -373,6 +373,9 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 			seat_set_focus_layer(seat, layer);
 			transaction_commit_dirty();
 		}
+		if (state == WLR_BUTTON_PRESSED) {
+			seatop_begin_down(seat, cont, time_msec, sx, sy);
```
Xyene, line 377: Isn't `cont` guaranteed to be null here? I think we probably want a new function in `seatop_down` like `seatop_begin_down_on_layer_surface` that takes the layer surface, and a new field in `seatop_down_event`. That gives the state necessary for the `exclusive_client` check too.

sway/input/seatop_down.c
Xyene: That'd be ideal, if we could get something like `ref_con_l[x|y]` out of the `wlr_surface` to keep movement working.

sway/input/seatop_down.c
```diff
@@ -110,3 +110,23 @@ void seatop_begin_down(struct sway_seat *seat, struct sway_container *con,
...
+void seatop_begin_down_on_layer_surface(struct sway_seat *seat,
+		struct wlr_surface *surface, uint32_t time_msec, double sx, double sy) {
+	seatop_end(seat);
+
+	struct seatop_down_event *e =
+		calloc(1, sizeof(struct seatop_down_event));
+	if (!e) {
+		return;
+	}
+	e->con = NULL;
```
pinselimo, line 123: We only need ```e->con``` now for ```handle_unref```. Would it be feasible to compare ```e->surface == con->view->surface``` there as well or is the surface too volatile?

emersion: We should stop the seatop_down on surface unmap, it doesn't make sense to keep it going if the surface is not visible anymore. So maybe we can just add an unmap listener?

pinselimo: Unfortunately ```wlr_surface``` doesn't emit an unmap event. Only thing close would be destroy .. ?

emersion: Hm right, I suppose destroy is fine then... It's not that of a big deal, it just means the user will maybe need to release the pointer button for the events to be sent if the surface underneath is unmapped.

sway/input/seatop_down.c
```diff
@@ -72,15 +73,29 @@ static void handle_tablet_tool_motion(struct sway_seat *seat,
...
 }
 
+static void handle_end(struct sway_seat *seat) {
+	struct seatop_down_event *e = seat->seatop_data;
+	wl_list_remove(&e->surface_destroy.link);
+}
+
 static const struct sway_seatop_impl seatop_impl = {
 	.button = handle_button,
+	.end = handle_end,
```
Xyene, line 98: Nit: put this next to `.unref`, not in the middle of the pointer handlers.

sway/input/seatop_down.c
```diff
@@ -110,3 +129,26 @@ void seatop_begin_down(struct sway_seat *seat, struct sway_container *con,
 	container_raise_floating(con);
 	transaction_commit_dirty();
 }
+
+void seatop_begin_down_on_layer_surface(struct sway_seat *seat,
```
Xyene, line 133: There's no actual constraint as far as I can tell that this `wlr_surface` needs to be a layer surface. We can call this `seatop_begin_down_on_surface` and call it from `seatop_begin_down`, to reduce code duplication.

sway/input/seatop_down.c
```diff
@@ -10,7 +10,10 @@
 
 struct seatop_down_event {
 	struct sway_container *con;
-	double ref_lx, ref_ly;         // cursor's x/y at start of op
+	struct sway_seat *seat;
+	struct wl_listener surface_destroy;
+	struct wlr_surface *surface;
+	double ref_lx, ref_ly;		 // cursor's x/y at start of op
```
Xyene, line 16: Nit: the comments lost alignment ;)

Xyene (commented):
Otherwise I think this LGTM.

sway/input/seatop_down.c
pinselimo: Ah alright, though it was alphabetical ^^

Xyene (approved):
LGTM, thanks!

PR #6429, opened 2021-08-10
xdg-decoration: let floating clients set borders

### LINKED ISSUE #6414: SDL_WINDOW_BORDERLESS doesn't work on wayland sway, but works on xwayland sway
**Sway Version:** 1.6.1
**Debug Log:** https://gist.github.com/1ace/59d502f9c82ba2f2123f34752a04c48e#file-sway-log
**Configuration File:** https://gist.github.com/1ace/59d502f9c82ba2f2123f34752a04c48e#file-sway-config
**Description:**

As the title says, SDL's `SDL_WINDOW_BORDERLESS` only works when talking to sway through xwayland, but not when talking directly wayland to sway.

Expected behaviour: the window decorations are not shown
Actual behaviour on wayland sway: the window decorations are shown regardless of the flag (incorrect)
Actual behaviour on xwayland sway: the window decorations are not shown (correct)

Everything needed to reproduce the issue is in this gist: https://gist.github.com/1ace/59d502f9c82ba2f2123f34752a04c48e
In `run.sh`, replace `SDL_VIDEODRIVER=wayland` with `SDL_VIDEODRIVER=x11` to reproduce the xwayland behaviour.

Dudemanguy:
The xdg-decoration protocol allows clients to request whether they want
to use server side decorations or client side decorations. Currently,
sway ignores this and always enforces whatever the server is currently
set to. Although tiled clients cannot be allowed to set borders, there
is no harm in listening requests from floating clients. First,
handle_map needs to check if the client has requested client side
decorations and flip the csd bool to true if it has. Secondly,
xdg_decoration_handle_request_mode needs to also check if a client
requests a different mode and update the container if applicable.
Sidenote: also fixed an unrelated style error.

I didn't check, but I think this would fix #6414 assuming this is how SDL requests a borderless window.

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -443,12 +443,15 @@ static void handle_map(struct wl_listener *listener, void *data) {
 
 	bool csd = false;
 
-	if (!view->xdg_decoration) {
+	if (view->xdg_decoration) {
+		enum wlr_xdg_toplevel_decoration_v1_mode mode =
+			view->xdg_decoration->wlr_xdg_decoration->client_pending_mode;
+		csd = mode == WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE;
+	} else {
 		struct sway_server_decoration *deco =
 				decoration_from_surface(xdg_surface->surface);
 		csd = !deco || deco->wlr_server_decoration->mode ==
 			WLR_SERVER_DECORATION_MANAGER_MODE_CLIENT;
-
 	}
 
 	view_map(view, view->wlr_xdg_surface->surface,
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -20,6 +20,7 @@
 #include "sway/tree/arrange.h"
 #include "sway/tree/view.h"
 #include "sway/tree/workspace.h"
+#include "sway/xdg_decoration.h"
 #include "list.h"
 #include "log.h"
 #include "stringop.h"
@@ -835,7 +836,13 @@ void container_set_floating(struct sway_container *container, bool enable) {
 		if (container->view) {
 			view_set_tiled(container->view, false);
 			if (container->view->using_csd) {
+				container->saved_border = container->pending.border;
 				container->pending.border = B_CSD;
+				if (container->view->xdg_decoration) {
+					struct sway_xdg_decoration *deco = container->view->xdg_decoration;
+					wlr_xdg_toplevel_decoration_v1_set_mode(deco->wlr_xdg_decoration,
+							WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE);
+				}
 			}
 		}
 		container_floating_set_default_size(container);
@@ -873,6 +880,11 @@ void container_set_floating(struct sway_container *container, bool enable) {
 			view_set_tiled(container->view, true);
 			if (container->view->using_csd) {
 				container->pending.border = container->saved_border;
+				if (container->view->xdg_decoration) {
+					struct sway_xdg_decoration *deco = container->view->xdg_decoration;
+					wlr_xdg_toplevel_decoration_v1_set_mode(deco->wlr_xdg_decoration,
+							WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE);
+				}
 			}
 		}
 		container->width_fraction = 0;
--- a/sway/xdg_decoration.c
+++ b/sway/xdg_decoration.c
@@ -10,7 +10,7 @@ static void xdg_decoration_handle_destroy(struct wl_listener *listener,
 		void *data) {
 	struct sway_xdg_decoration *deco =
 		wl_container_of(listener, deco, destroy);
-	if(deco->view) {
+	if (deco->view) {
 		deco->view->xdg_decoration = NULL;
 	}
 	wl_list_remove(&deco->destroy.link);
@@ -23,8 +23,32 @@ static void xdg_decoration_handle_request_mode(struct wl_listener *listener,
 		void *data) {
 	struct sway_xdg_decoration *deco =
 		wl_container_of(listener, deco, request_mode);
+	struct sway_view *view = deco->view;
+	enum wlr_xdg_toplevel_decoration_v1_mode mode =
+		WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE;
+	enum wlr_xdg_toplevel_decoration_v1_mode client_mode =
+		deco->wlr_xdg_decoration->client_pending_mode;
+
+	bool floating;
+	if (view->container) {
+		floating = container_is_floating(view->container);
+		bool csd = false;
+		csd = client_mode ==
+			WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE;
+		view_update_csd_from_client(view, csd);
+		arrange_container(view->container);
+		transaction_commit_dirty();
+	} else {
+		floating = view->impl->wants_floating &&
+			view->impl->wants_floating(view);
+	}
+
+	if (floating && client_mode) {
+		mode = client_mode;
+	}
+
 	wlr_xdg_toplevel_decoration_v1_set_mode(deco->wlr_xdg_decoration,
-			WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE);
+			mode);
 }
 
 void handle_xdg_decoration(struct wl_listener *listener, void *data) {


emersion:
Testing with the `toplevel-decoration` example in wlroots, it seems like this doesn't behave correctly in tiled mode. `toplevel-decoration client` gets client-side decorations in tiled mode it seems, but Sway should force server-side decorations in this case.

Dudemanguy:
Oops, the old border needs to be saved in the floating toggle. It should work now.

sway/xdg_decoration.c
```diff
@@ -23,8 +23,28 @@ static void xdg_decoration_handle_request_mode(struct wl_listener *listener,
 		void *data) {
 	struct sway_xdg_decoration *deco =
 		wl_container_of(listener, deco, request_mode);
+	struct sway_view *view = deco->view;
+	bool wants_floating = view->impl->wants_floating;
```
emersion, line 27: `impl->wants_floating` is a function pointer, so need to call the function to get the `bool`. The function may be NULL.

If we have a container attached to the view, it's better to use `container_is_floating` to find out whether the view is floating. If we don't have a container (e.g. toplevel isn't mapped yet), using `wants_floating` sounds reasonable.

emersion:
With that comment fixed, it doesn't seem like this patch is enough to make floating views use CSD if they were initially tiled. Maybe something wrong with the CSD logic inside `container_set_floating`?

Dudemanguy:
I changed the floating check to an if/else based on if the view has a container. It seems to work.

Dudemanguy:
Actually, is the `wants_floating` check even needed? It seems like `view_update_csd_from_client` handles it fine and `wlr_xdg_toplevel_decoration_v1_set_mode` doesn't cause any ill effects if the window is tiled.

Dudemanguy:
Okay, I think I got it now. `view_update_csd_from_client` and such always needs to be called as long as there is a container but `wants_floating` needs to be checked so sway sends the correct mode on the configure event. Additionally, toggling floating needs to also send configure events if the container is using csd.

emersion:
This sounds to behave a lot better, nice! I'm getting segfaults with some clients, because not all clients support the xdg-decoration protocol. e.g.

```
../sway/tree/container.c:842:5: runtime error: member access within null pointer of type 'struct sway_xdg_decoration'
AddressSanitizer:DEADLYSIGNAL
==6451==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x555ecff8db9a bp 0x7ffce7388df0 sp 0x7ffce7388d80 T0)
==6451==The signal is caused by a READ memory access.
==6451==Hint: address points to the zero page.
    #0 0x555ecff8db9a in container_set_floating ../sway/tree/container.c:842
    #1 0x555ecff0db42 in cmd_floating ../sway/commands/floating.c:51
    #2 0x555ecfddaad3 in execute_command ../sway/commands.c:291
    #3 0x555ecff06ad4 in seat_execute_command ../sway/commands/bind.c:631
    #4 0x555ecfe8dfb0 in handle_key_event ../sway/input/keyboard.c:491
    #5 0x555ecfe8f115 in handle_keyboard_group_key ../sway/input/keyboard.c:563
    #6 0x7fa5fbd5b0a2 in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #7 0x7fa5fbcb5fa0 in wlr_keyboard_notify_key ../subprojects/wlroots/types/wlr_keyboard.c:93
    #8 0x7fa5fbcba0b6 in handle_keyboard_key ../subprojects/wlroots/types/wlr_keyboard_group.c:133
    #9 0x7fa5fbd5b0a2 in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #10 0x7fa5fbcb5fa0 in wlr_keyboard_notify_key ../subprojects/wlroots/types/wlr_keyboard.c:93
    #11 0x7fa5fbc08f96 in keyboard_handle_key ../subprojects/wlroots/backend/wayland/seat.c:267
    #12 0x7fa5f9ce4acc  (/usr/lib/libffi.so.7+0x6acc)
    #13 0x7fa5f9ce4039  (/usr/lib/libffi.so.7+0x6039)
    #14 0x7fa5fa34dfe3  (/usr/lib/libwayland-client.so.0+0x9fe3)
    #15 0x7fa5fa34a562  (/usr/lib/libwayland-client.so.0+0x6562)
    #16 0x7fa5fa34bcab in wl_display_dispatch_queue_pending (/usr/lib/libwayland-client.so.0+0x7cab)
    #17 0x7fa5fbbf7a6e in dispatch_events ../subprojects/wlroots/backend/wayland/backend.c:53
    #18 0x7fa5fb18a079 in wl_event_loop_dispatch (/usr/lib/libwayland-server.so.0+0xb079)
    #19 0x7fa5fb187be6 in wl_display_run (/usr/lib/libwayland-server.so.0+0x8be6)
    #20 0x555ecfe1b663 in server_run ../sway/server.c:261
    #21 0x555ecfe1692f in main ../sway/main.c:395
    #22 0x7fa5fa5e3b24 in __libc_start_main (/usr/lib/libc.so.6+0x27b24)
    #23 0x555ecfdd87ad in _start (/home/simon/src/sway/build/sway/sway+0x33b7ad)
```

sway/xdg_decoration.c
```diff
@@ -23,8 +23,33 @@ static void xdg_decoration_handle_request_mode(struct wl_listener *listener,
...
+	bool wants_floating;
+	if (view->container) {
+		wants_floating = container_is_floating(view->container);
+		bool csd = false;
+		if (wants_floating) {
+			csd = client_mode ==
+				WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE;
+		}
+		view_update_csd_from_client(view, csd);
+		arrange_container(view->container);
```
emersion, line 41: I think we're missing `transaction_commit_dirty` right after

sway/xdg_decoration.c
```diff
@@ -23,8 +23,33 @@ static void xdg_decoration_handle_request_mode(struct wl_listener *listener,
 		void *data) {
 	struct sway_xdg_decoration *deco =
 		wl_container_of(listener, deco, request_mode);
+	struct sway_view *view = deco->view;
+	enum wlr_xdg_toplevel_decoration_v1_mode mode =
+		WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE;
+	enum wlr_xdg_toplevel_decoration_v1_mode client_mode =
+		deco->wlr_xdg_decoration->client_pending_mode;
+
+	bool wants_floating;
```
emersion, line 32: Nit: maybe consider renaming to just `floating`.

sway/desktop/xdg_shell.c
```diff
@@ -443,12 +443,15 @@ static void handle_map(struct wl_listener *listener, void *data) {
 
 	bool csd = false;
 
-	if (!view->xdg_decoration) {
+	if (view->xdg_decoration) {
+		enum wlr_xdg_toplevel_decoration_v1_mode mode =
+			view->xdg_decoration->wlr_xdg_decoration->client_pending_mode;
+		csd = mode == WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE;
```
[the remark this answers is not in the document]
Dudemanguy: I'm not sure about this part. It seems like it is too early in the process to check for floating? At least if I do `view->impl->wants_floating && view->impl->wants_floating(view)`, that always returns false if the client is just started up (hacked an app-id into toplevel-decoration to easily make the window start up as floating).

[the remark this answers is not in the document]
emersion: At map-time, the view should always have a container, so `container_is_floating` could be used.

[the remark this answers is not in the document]
emersion: Ah. But only after the `view_map` call. Hmm.

sway/xdg_decoration.c
```diff
@@ -23,8 +23,34 @@ static void xdg_decoration_handle_request_mode(struct wl_listener *listener,
...
+		WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE;
+	enum wlr_xdg_toplevel_decoration_v1_mode client_mode =
+		deco->wlr_xdg_decoration->client_pending_mode;
+
+	bool floating;
+	if (view->container) {
+		floating = container_is_floating(view->container);
+		bool csd = false;
+		if (floating) {
+			csd = client_mode ==
```
emersion, line 37: Since `view_update_csd_from_client` already has a `container_is_floating` check, this `if` is unnecessary, we can set `csd` to the client's preference unconditionally.

emersion (commented):
Apart from this cleanup, LGTM!

sway/tree/container.c
```diff
@@ -826,6 +827,7 @@ void container_set_floating(struct sway_container *container, bool enable) {
 	struct sway_seat *seat = input_manager_current_seat();
 	struct sway_workspace *workspace = container->pending.workspace;
 	struct sway_container *focus = seat_get_focused_container(seat);
+	struct sway_xdg_decoration *deco = container->view->xdg_decoration;
```
emersion, line 830: Hm. I forgot one thing: at this point, there's no guarantee that the container has a view. Se we can't have this variable declaration here, it needs to be moved down the `if`s.

sway/desktop/xdg_shell.c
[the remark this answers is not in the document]
Dudemanguy: Oh oops, I missed the other comments on this. I'll put the logic back in `xdg_shell.c`.

Dudemanguy:
I double checked #6414 and the window is indeed now borderless if you toggle floating (had to modify the example to not instantly quit on keypress).

emersion (approved):
LGTM, thanks!

1ace:
Awesome @Dudemanguy, I just tested and can confirm this fixed it; thanks! :muscle:

PR #6430, opened 2021-08-11
Use fixed titlebar heights

### LINKED ISSUE #4992: Japanese/Chinese characters change title height which can cause screen flashing

WhyNotHugo:
Currently the height of titlebars can vary depending on their text.

If the text contains latin characters, then the height of those
characters (in whatever font applies) will define the titlebar height.
If a text contains CJK characters, the height of that font will
determine the height. The same applies for emoji and any other character
that's available only in the non-default font.

This issue is annoying in itself, but because all existing windows are
immediately resized, all applications need to rerender and this
frequently results in some of them crashing (Firefox, Chromium and
Signal are all curlpits here). I've only been able to reproduce this
with many windows open, so I suspect it's a matter of too much load for
the system to handle.

This quickly escalates to the point where visiting websites with emoji
in their title can kill all browser windows.

Additionally, when the titlebar of a window grows, it's first rendered
cropped, and the actual height only updates when there's a change in the
focused window (I'm sure other state changes trigger a redraw too), at
which point it's no longer cropped.

The approach taken here calculates the expected titlebar height based on
a short set of characters. Since these are likely to all use different
fonts, titlebars will be big enough to fit latin, CJK or emoji
characters without resizing.

I'm sure this still leaves out specific languages (e.g.: arabic
character bigger?). There's two ways to further improve this situation:

- Add a single character of each acceptable family to the same text
  used.
- Make the sample text user-configurable. This would allow a user to
  configure which font-families they expect to see, any anything bigger
  will be cropped. I don't feel the benefits are worth the effort here,
  but have nothing against it.

Fixes #4992

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -991,31 +991,18 @@ int workspace_output_cmp_workspace(const void *a, const void *b) {
 	return lenient_strcmp(wsa->workspace, wsb->workspace);
 }
 
-static void find_font_height_iterator(struct sway_container *con, void *data) {
-	size_t amount_below_baseline = con->title_height - con->title_baseline;
-	size_t extended_height = config->font_baseline + amount_below_baseline;
-	if (extended_height > config->font_height) {
-		config->font_height = extended_height;
-	}
-}
-
-static void find_baseline_iterator(struct sway_container *con, void *data) {
-	bool *recalculate = data;
-	if (*recalculate) {
-		container_calculate_title_height(con);
-	}
-	if (con->title_baseline > config->font_baseline) {
-		config->font_baseline = con->title_baseline;
-	}
-}
-
 void config_update_font_height(bool recalculate) {
 	size_t prev_max_height = config->font_height;
-	config->font_height = 0;
-	config->font_baseline = 0;
 
-	root_for_each_container(find_baseline_iterator, &recalculate);
-	root_for_each_container(find_font_height_iterator, NULL);
+	cairo_t *cairo = cairo_create(NULL);
+	int height;
+	int baseline;
+	get_text_size(cairo, config->font, NULL, &height, &baseline, 1,
+			config->pango_markup, "%s", "Panda 🐼 熊猫 판다");
+	cairo_destroy(cairo);
+
+	config->font_height = height;
+	config->font_baseline = baseline;
 
 	if (config->font_height != prev_max_height) {
 		arrange_root();


WhyNotHugo:
> Additionally, when the titlebar of a window grows, it's first rendered
> cropped, and the actual height only updates when there's a change in the
> focused window (I'm sure other state changes trigger a redraw too), at
> which point it's no longer cropped.

This "bug" can be reproduce by using this config:

```
exec alacritty
exec alacritty
exec alacritty
```

And, on of the three terminals, run `vim 熊猫`. You'll notice the title is first rendered cropped, though the height will update as soon as focus switches to another window.

emersion:
This approach has been rejected before, because there's no guarantee the chosen characters will be representative of the user's locale.

WhyNotHugo:
Would making the chosen characters configurable be acceptable?

emersion:
This has been discussed at length countless times. Unless something new has come up, https://github.com/swaywm/sway/issues/4992#issuecomment-610967189 is the latest thoughts we had.

WhyNotHugo:
The approach in @ddevault comment seems just way too complicated. Assuming the config setting were implemented, the user would have to figure out how to calculate what height they want.

This needs some understanding of how pango calculates heights, etc. Which is what the code in this PR does. It just seems weird to have the user do the complicated font-sizing-related math, instead of sway doing it itself.

This issue is becoming a huge nuisance to me, since it's ends up resulting in lots of application crashes just be clicking on link.

Sorry for bringing up an approach that's been brought up before (I must have missed it the previous time). But, again, is making the chosen characters configurable really not an acceptable change? That seems the most flexible approach (e.g.: can adapt to any locale), and doesn't burden users with manually calculating font metrics.

WhyNotHugo:
It's also worth noting that, while this may not cover some locales, users of those locales won't see resizing anyway, but still a fixed titlebar sized.

It's not perfect, but it's does fix the issues at hand. Support for that locale could easily be added by updating the chosen characters here.

fluix-dev:
Hey, just another user that gets occasionally bugged by having his titlebars resize or become too tall.

> The approach in ddevault comment seems just way too complicated.

While I agree that it takes some configuring, it's a very *clear* approach. Everybody knows what pixels are and how tall they want their titlebar to *generally* look. It matches the other settings like gaps and borders and I would guess that users will set it based on what they recall often seeing so you won't have issues with various locales.

However, I understand the usefulness of your approach so perhaps a `contrib/` script that takes some characters as input and gets the max height would be accepted? Not sure how one would really do this in a particularly portable way though.

WhyNotHugo:
Yeah, I’ve been thinking the same thing, porting this into a contrib script.

The code for such a script is basically on this PR. But instead of sway calling the function itself, you’d have to do it manually and provide the value back.

I honestly don’t think it’s a terrible approach, but it satisfies the requirement to get this merged.

WhyNotHugo:
My apologies for the "demanding" tone so far. The issue frustrates me, but that should not be reflected in my attitude towards you guys. I appreciate your patience.

I'd very much like to continue working on a viable fix for this, but I'll continue on-topic discussions in #6059 or #6433.

PR #6433, opened 2021-08-12
Manually computed fixed titlebar heights

WhyNotHugo:
This is an implementation of https://github.com/swaywm/sway/issues/4992#issuecomment-610967189

I don't expect this to be merge -- on the contrary, this PR is a prototype to further discuss this issue and next steps.

In its current form, this patch allows explicitly defining a fixed titlebar height (while falling back to the previous behaviour if this is not explicitly defined).

However, because titlebar height and baseline calculation and intrinsically related, a fixed titlebar height with a dynamic baseline means the text moves around depending on content (e.g.: as the title changes, if the character set used changes, so does the baseline).

As an extra consequence of that, the vertical alignment of title texts also depends on the character set of titles. However, because the baseline is calculated globally, the vertical alignment of a window's title depends on the content of _all_ open windows.

I think the best approach to work around this is to make both the height **and** baseline explicitly configurable. Do you think that would be acceptable?


WhyNotHugo:
Demonstration of the moving up-and-down due to the dynamic baseline:

https://user-images.githubusercontent.com/730811/129243120-e29738fa-af33-4641-9148-c5aa086cb0eb.mp4

vvrein:
Hmm, seems that configurable baseline (together with fixed height) could be easily added here -> https://github.com/swaywm/sway/pull/6059

WhyNotHugo:
The baseline that gets applied when rendering is `config->font_baseline`, here:

https://github.com/swaywm/sway/blob/4832fc937fae5a5e2af8828a7092c9026bd300e0/sway/desktop/render.c#L560-L564

It's the same value for all containers (which makes sense, else you'd get unaligned titles).

WhyNotHugo:
"Applied" is maybe not the right word. Pango itself applies baseline, etc. Sway itself adds a bit of padding which varies based on the `config->font_baseline`, which is, AFAIK, the `max()` of all the title's baseline.

The commit actually has a good explanation (which I wish I'd read first): https://github.com/swaywm/sway/commit/9215ca0f01f7d155191d11fd9caa2574387c6f84

Calculating this value per-container results in misaligned titlebars across containers, and titles for individual containers moving up and down as the applied fonts change. 

Calculating it globally means things will shift as the contents of _any_ titlebar changes, though everything is always aligned.

This value should likely be hardcoded when using fixed titlebar heights, probably as a second argument.

---

I wish I'd noticed #6059 sooner, since that seems to as close to fixing this.

I believe that calls to `find_font_height_iterator`  are unnecessary when a fixed titlebar height is being used, but that aside your approach seems to be bit somewhat cleaner so far.

WhyNotHugo:
I think it should be possible to use `pango_context_get_metrics` to get the font metrics for the configured fonts. Passing `language=NULL` will yield results based on the user's locale.

With that result, we can use `pango_font_metrics_get_ascent` to determine the height of the block we want to vertically centre. Adding padding on top equal to `pango_font_metrics_get_descent` should keep things vertically centred.

This means we wouldn't need to use the baseline to center titles any more.

emersion:
Right. We need to align the baselines across all titlebars.

I wonder if we can use `pango_font_metrics_get_height` to pick the fixed height? And then compute a fixed baseline offset from that (maybe for ascent/descent) so that the text is centered, and crop the text if it's too large?

WhyNotHugo:
Using `pango_font_metrics_get_height` for the fixed height was exactly what I was aiming for (apparently I mentioned the wrong function in my last comment).

I've pushed some changes that do exactly this.

I've used `baseline = height - descent`. This makes sense based on the definition of the [height](https://docs.gtk.org/Pango/method.FontMetrics.get_height.html) and [descent](https://docs.gtk.org/Pango/method.FontMetrics.get_descent.html).

The titlebar height is consistent what what I'd expect based on this, and the text is vertically centred.

However this is PR is not fully ready. The current patch "works" in that the titlebar has a fixed height, however content higher than the fixed height doesn't get trimmed and a grey box is rendered in its place. I still have to tweak some box-size calculations in `render_titlebar`.

I'll add the configuration command to change the fixed height on top of these changes, if this makes sense. Preliminary feedback on the approach is welcome.

WhyNotHugo:
Actually, my prior comment was correct:

`pango_font_metrics_get_ascent` is the distance between the baseline and the top. And `pango_font_metrics_get_descent` is the distance below the baseline. So we can use the ascent for our baseline, and the sum of both as a height.

`pango_font_metrics_get_height` is not what we'd expect: it's the total distance between one line's baseline and the following line's baseline (when wrapping text), so it's a bit higher that what we want in this case.

WhyNotHugo:
Alright, I've figured out how to properly crop titles.

Rather than crop them when rendering in `render_titlebar` (which is called very often), I do this in `update_title_texture`, which is only called when the titles change. The title text is centred before being drawn into the texture. This avoids rendering bits that won't be used and makes the positioning very straightforward; titles are all consistently centred inside the texture where they're initially rendered.

I tested this with my setup and all combinations seem to work:

- Titlebar is never resized.
- The default size is based on metrics for the current locale.
- If larger fonts are used, the titles keep their positioning (e.g.: never move around).
- If other fonts are too big, they're cropped, and we never overlap the titlebar border.

I'm not terribly happy with how the code looks, and welcome suggestions for improvements. I _am_ happy with the result on how it renders :relieved: 

Titlebar height is still not configurable -- I'm not sure if it's best to implement that in this PR, or a follow up on.

emersion:
Thanks for continuing this work. I didn't realize we had a sensible way to pick a good fixed height, without relying on rendering particular glyths.

> Titlebar height is still not configurable -- I'm not sure if it's best to implement that in this PR, or a follow up on.

Since the titlebar padding is configurable, maybe we can try this without introducing a new titlebar height option, and users can always increase the padding if something important gets cut off? We can consider adding a new option or changing our strategy if ascent/descent ends up causing issues.

I'll have a look at the code in the next few days.

WhyNotHugo:
Oh, I was not at all aware that the padding was configurable (I though it was hard-coded somewhere). Making the height configurable only made sense as a way to adjust that padding, and I think introducing that would be redundant and some pointless complexity.

I've now tried this branch configuring vertical padding `titlebar_padding 5 1` and `titlebar_padding 5 27`. The first is crops larger fonts as expected, and the second centres properly.

common/pango.c
```diff
@@ -109,6 +109,24 @@ void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
 	free(buf);
 }
 
+void get_text_metrics(const char *font, size_t *height, size_t *baseline) {
```
emersion, line 112: Nit: since `height` and `baseline` are not array indices, I don't think it makes sense for them to be `size_t`. Pango just uses `int`.

Edit: hmm, seems like we incorrectly use it elsewhere. Would be nice to clean this up (but it's unrelated to this PR).

sway/tree/container.c
```diff
@@ -536,6 +537,10 @@ static void update_title_texture(struct sway_container *con,
 		return;
 	}
 
+	if (height > (int) config->font_height) {
```
emersion, line 540: (Dropping `size_t` should allow us to drop these casts.)

emersion (changes requested):
Code looks pretty good!

Do we still need `container_calculate_title_height`? Do we still need `sway_container.title_height` and `title_baseline`?

Tested and this seems to work pretty well :+1:

WhyNotHugo:
We don't need `container_calculate_title_height`, `sway_container.title_height` and `title_baseline` any more, and I've pushed more changes dropping these.

I also just noticed that prior to dropping these, markers are not drawn correctly; with padding < 2, they render on top of the border and screw it up. _After_ dropping these couple of variables and references to it, it's just _slightly_ misaligned (1px under the title, but if you use no padding, it's noticeable since the square braces get cropped).

I'll update the markers-rendering logic to basically follow the same cropping/positioning rules as the titles themselves, since both should be doing the same thing anyway.

WhyNotHugo:
Alright, I've made markers get centred/cropped properly now too. The code for rendering titles and the code for rendering markers look **too** similar, and I think it would make sense to deduplicate a lot of it into a common function, but that's probably best in a follow-up PR.

Also `config_update_font_height` gets called each time a title's content changes, but should only be called when the configured font changes. I'd move that into `cmd_font`, but that doesn't get called for the _default_ font, so it's not that simple. And, again, I think that might make sense as a separate follow-up.

FrostKiwi:
https://github.com/swaywm/sway/pull/6059 was dropped in favor of this PR. As I understand, a direct titlebar_height setting is thus gone and this setup: https://github.com/swaywm/sway/pull/6059#issuecomment-856094475 which I really enjoy is thus not possible any more.
Am I understand it correctly?

PR #6438, opened 2021-08-14
cmd_bind_or_unbind_switch: fix copy/paste typo error message

1ace:
`[0]` is the switch type, not its state; we want `[1]` for that, and it's a string not an int :)


emersion (approved):
Thanks!

PR #6445, opened 2021-08-15
Add patch/black-clear_color

tmpm697:
To avoid double flickers when display wallpaper. Now you have one black screen -> wallpaper.


tmpm697:
Wrong repo, should be in my swea, sr.

Issue #6454, opened 2021-08-20
wl_touch and wl_pointer positions


stacyharper:
### Please fill out the following:

- sway version 1.6.1-50f979363e (Jun 26 2021, branch 'master')

My context: I'm implementing pointer and touch event on bemenu (https://github.com/Cloudef/bemenu/pull/200)

I come here after asking in #sway cause it initially could be a question. I now think there is an issue in the behavior of wl_touch motions. And someone in irc asked me to open an issue for futher discussion.

Lets say I got this window composition : 

 ![](http://dav.misterbanal.net/fb6b2521-8bf0-44c6-8f7c-7a6e8ab5da43.png)

Here, Clicking (and holding) on the Wev surface. I then move to other windows without releasing. I continue to receive motion events with the position relative to the initial surface position. Other windows dont grab focus and the initial surface got everything needed to understand and behave correctly.

Not take the exact same composition and touch the screen with you finger. Then you move your finger outside of wev. You will continue to receive motion event but here, the position will be relative to the new surface. The initial surface then think you entered again to this surface and cant behave accordingly to your real finger position.

I also suffer more issue with this bemenu works. I dont receive those pointer motion event when holding click when I leave the surface.

Is it a bug ? Is there something bemenu must to do to receive those event correctly ?

Thanks a lot

emersion:
>  I dont receive those pointer motion event when holding click when I leave the surface.

Ref https://github.com/swaywm/sway/pull/6418

> You will continue to receive motion event but here, the position will be relative to the new surface.

Sounds like a bug.

Eluminae:
I just edited a part :  

I also suffer more issue with this bemenu works. I dont receive those pointer motion event when holding click when I leave the surface IF THERE IS NO OTHER SURFACE BELLOW. I still dont know why. I received them exactly as Wev receive them.

Xyene:
> > You will continue to receive motion event but here, the position will be relative to the new surface.

> Sounds like a bug.

Ref https://github.com/swaywm/sway/issues/5630

I think this is just due to Sway not having a seatop_down implementation for touch input. If you (or someone else) wanted to take a stab at it, it shouldn't be too hard to get working. https://github.com/swaywm/sway/commit/5d13f647f9384e59012c0f829651911564bb5365 implemented this for tablet input, and I suspect a touch implementation would be very similar.

Eluminae:
Yeah but we receive event if there is another window under the touch. I feel there is just a bug on an output determination at one point.

[...]

PR #6455, opened 2021-08-21
Implement seatop_touch

stacyharper:
Atm we got issue with the touch position sent to the clients. While
holding contact, leaving the initial container will continue to send
motion event to the client but with the new local position from the new
container.

This seatop goal is to send the position of the touch event, relatively
to the initial container layout position.

This will solve: https://github.com/swaywm/sway/issues/6454

I'm pretty new to Wayland and even more with the Sway codebase.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -364,7 +364,7 @@ void cursor_unhide(struct sway_cursor *cursor) {
 	wl_event_source_timer_update(cursor->hide_source, cursor_get_timeout(cursor));
 }
 
-static void pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
+void pointer_motion(struct sway_cursor *cursor, uint32_t time_msec,
 		struct wlr_input_device *device, double dx, double dy,
 		double dx_unaccel, double dy_unaccel) {
 	wlr_relative_pointer_manager_v1_send_relative_motion(
@@ -479,43 +479,16 @@ static void handle_touch_down(struct wl_listener *listener, void *data) {
 	cursor_hide(cursor);
 
 	struct sway_seat *seat = cursor->seat;
-	struct wlr_seat *wlr_seat = seat->wlr_seat;
-	struct wlr_surface *surface = NULL;
 
 	double lx, ly;
 	wlr_cursor_absolute_to_layout_coords(cursor->cursor, &event->touch->base,
 			event->x, event->y, &lx, &ly);
-	double sx, sy;
-	struct sway_node *focused_node = node_at_coords(seat, lx, ly, &surface, &sx, &sy);
 
 	seat->touch_id = event->touch_id;
 	seat->touch_x = lx;
 	seat->touch_y = ly;
 
-	if (surface && wlr_surface_accepts_touch(wlr_seat, surface)) {
-		if (seat_is_input_allowed(seat, surface)) {
-			wlr_seat_touch_notify_down(wlr_seat, surface, event->time_msec,
-					event->touch_id, sx, sy);
-
-			if (focused_node) {
-			    seat_set_focus(seat, focused_node);
-			}
-		}
-	} else if (!cursor->simulating_pointer_from_touch &&
-			(!surface || seat_is_input_allowed(seat, surface))) {
-		// Fallback to cursor simulation.
-		// The pointer_touch_id state is needed, so drags are not aborted when over
-		// a surface supporting touch and multi touch events don't interfere.
-		cursor->simulating_pointer_from_touch = true;
-		cursor->pointer_touch_id = seat->touch_id;
-		double dx, dy;
-		dx = lx - cursor->cursor->x;
-		dy = ly - cursor->cursor->y;
-		pointer_motion(cursor, event->time_msec, &event->touch->base, dx, dy,
-			dx, dy);
-		dispatch_cursor_button(cursor, &event->touch->base, event->time_msec,
-				BTN_LEFT, WLR_BUTTON_PRESSED);
-	}
+	seatop_touch_down(seat, event, lx, ly);
 }
 
 static void handle_touch_up(struct wl_listener *listener, void *data) {
@@ -523,7 +496,7 @@ static void handle_touch_up(struct wl_listener *listener, void *data) {
 	struct wlr_touch_up_event *event = data;
 	cursor_handle_activity_from_device(cursor, &event->touch->base);
 
-	struct wlr_seat *wlr_seat = cursor->seat->wlr_seat;
+	struct sway_seat *seat = cursor->seat;
 
 	if (cursor->simulating_pointer_from_touch) {
 		if (cursor->pointer_touch_id == cursor->seat->touch_id) {
@@ -532,7 +505,7 @@ static void handle_touch_up(struct wl_listener *listener, void *data) {
 				event->time_msec, BTN_LEFT, WLR_BUTTON_RELEASED);
 		}
 	} else {
-		wlr_seat_touch_notify_up(wlr_seat, event->time_msec, event->touch_id);
+		seatop_touch_up(seat, event);
 	}
 }
 
@@ -543,19 +516,14 @@ static void handle_touch_motion(struct wl_listener *listener, void *data) {
 	cursor_handle_activity_from_device(cursor, &event->touch->base);
 
 	struct sway_seat *seat = cursor->seat;
-	struct wlr_seat *wlr_seat = seat->wlr_seat;
-	struct wlr_surface *surface = NULL;
 
 	double lx, ly;
 	wlr_cursor_absolute_to_layout_coords(cursor->cursor, &event->touch->base,
 			event->x, event->y, &lx, &ly);
-	double sx, sy;
-	node_at_coords(cursor->seat, lx, ly, &surface, &sx, &sy);
 
 	if (seat->touch_id == event->touch_id) {
 		seat->touch_x = lx;
 		seat->touch_y = ly;
-
 		struct sway_drag_icon *drag_icon;
 		wl_list_for_each(drag_icon, &root->drag_icons, link) {
 			if (drag_icon->seat == seat) {
@@ -572,9 +540,8 @@ static void handle_touch_motion(struct wl_listener *listener, void *data) {
 			pointer_motion(cursor, event->time_msec, &event->touch->base,
 				dx, dy, dx, dy);
 		}
-	} else if (surface) {
-		wlr_seat_touch_notify_motion(wlr_seat, event->time_msec,
-			event->touch_id, sx, sy);
+	} else {
+		seatop_touch_motion(seat, event, lx, ly);
 	}
 }
 
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1618,6 +1618,26 @@ void seatop_pointer_axis(struct sway_seat *seat,
 	}
 }
 
+void seatop_touch_motion(struct sway_seat *seat, struct wlr_touch_motion_event *event,
+		double lx, double ly) {
+	if (seat->seatop_impl->touch_motion) {
+		seat->seatop_impl->touch_motion(seat, event, lx, ly);
+	}
+}
+
+void seatop_touch_up(struct sway_seat *seat, struct wlr_touch_up_event *event) {
+	if (seat->seatop_impl->touch_up) {
+		seat->seatop_impl->touch_up(seat, event);
+	}
+}
+
+void seatop_touch_down(struct sway_seat *seat, struct wlr_touch_down_event *event,
+		double lx, double ly) {
+	if (seat->seatop_impl->touch_down) {
+		seat->seatop_impl->touch_down(seat, event, lx, ly);
+	}
+}
+
 void seatop_tablet_tool_tip(struct sway_seat *seat,
 		struct sway_tablet_tool *tool, uint32_t time_msec,
 		enum wlr_tablet_tool_tip_state state) {
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -261,7 +261,7 @@ static void handle_tablet_tool_tip(struct sway_seat *seat,
 
 		// Handle tapping on a container surface
 		seat_set_focus_container(seat, cont);
-		seatop_begin_down(seat, node->sway_container, time_msec, sx, sy);
+		seatop_begin_down(seat, node->sway_container, sx, sy);
 	}
 #if HAVE_XWAYLAND
 	// Handle tapping on an xwayland unmanaged view
@@ -374,7 +374,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 			transaction_commit_dirty();
 		}
 		if (state == WLR_BUTTON_PRESSED) {
-			seatop_begin_down_on_surface(seat, surface, time_msec, sx, sy);
+			seatop_begin_down_on_surface(seat, surface, sx, sy);
 		}
 		seat_pointer_notify_button(seat, time_msec, button, state);
 		return;
@@ -499,7 +499,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 
 	// Handle mousedown on a container surface
 	if (surface && cont && state == WLR_BUTTON_PRESSED) {
-		seatop_begin_down(seat, cont, time_msec, sx, sy);
+		seatop_begin_down(seat, cont, sx, sy);
 		seat_pointer_notify_button(seat, time_msec, button, WLR_BUTTON_PRESSED);
 		return;
 	}
@@ -647,6 +647,36 @@ static void handle_tablet_tool_motion(struct sway_seat *seat,
 	}
 
 	e->previous_node = node;
+}
+
+static void handle_touch_down(struct sway_seat *seat,
+		struct wlr_touch_down_event *event, double lx, double ly) {
+	struct wlr_surface *surface = NULL;
+	struct wlr_seat *wlr_seat = seat->wlr_seat;
+	struct sway_cursor *cursor = seat->cursor;
+	double sx, sy;
+	node_at_coords(seat, seat->touch_x, seat->touch_y, &surface, &sx, &sy);
+
+	if (surface && wlr_surface_accepts_touch(wlr_seat, surface)) {
+		if (seat_is_input_allowed(seat, surface)) {
+			cursor->simulating_pointer_from_touch = false;
+			seatop_begin_touch_down(seat, surface, event, sx, sy, lx, ly);
+		}
+	} else if (!cursor->simulating_pointer_from_touch &&
+			(!surface || seat_is_input_allowed(seat, surface))) {
+		// Fallback to cursor simulation.
+		// The pointer_touch_id state is needed, so drags are not aborted when over
+		// a surface supporting touch and multi touch events don't interfere.
+		cursor->simulating_pointer_from_touch = true;
+		cursor->pointer_touch_id = seat->touch_id;
+		double dx, dy;
+		dx = seat->touch_x - cursor->cursor->x;
+		dy = seat->touch_y - cursor->cursor->y;
+		pointer_motion(cursor, event->time_msec, &event->touch->base, dx, dy,
+				dx, dy);
+		dispatch_cursor_button(cursor, &event->touch->base, event->time_msec,
+				BTN_LEFT, WLR_BUTTON_PRESSED);
+	}
 }
 
 /*----------------------------------------\
@@ -1096,6 +1126,7 @@ static const struct sway_seatop_impl seatop_impl = {
 	.swipe_begin = handle_swipe_begin,
 	.swipe_update = handle_swipe_update,
 	.swipe_end = handle_swipe_end,
+	.touch_down = handle_touch_down,
 	.rebase = handle_rebase,
 	.allow_set_cursor = true,
 };
--- a/sway/input/seatop_down.c
+++ b/sway/input/seatop_down.c
@@ -2,11 +2,19 @@
 #include <float.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_tablet_v2.h>
+#include <wlr/types/wlr_touch.h>
 #include "sway/input/cursor.h"
 #include "sway/input/seat.h"
 #include "sway/tree/view.h"
 #include "sway/desktop/transaction.h"
 #include "log.h"
+
+struct seatop_touch_point_event {
+	double ref_lx, ref_ly;         // touch's x/y at start of op
+	double ref_con_lx, ref_con_ly; // container's x/y at start of op
+	int32_t touch_id;
+	struct wl_list link;
+};
 
 struct seatop_down_event {
 	struct sway_container *con;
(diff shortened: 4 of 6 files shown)


sway/input/seatop_touch.c
```diff
@@ -0,0 +1,110 @@
...
+		if (point_event->touch_id == event->touch_id) {
+			fprintf(stderr, "We found our point event\n");
+			break;
+		}
+	};
+
+	wl_list_remove(&point_event->link);
+
+	if (wl_list_empty(&e->point_events)) {
+		//seatop_end(seat);
```
Eluminae, line 64: This make sway to crash. I'm not sure yet how to release cleanly the seatop.

Eluminae:
For record, I firstly tried to rely on seatop_down but it use cursor as much as possible. As touches can be multiple, I think we have to implement a new dedicated seatop.

sway/input/seatop_touch.c
kennylevinsen: To change back to the default seatop, use `seatop_begin_default`. See `handle_button` in `sway/input/seatop_down.c` for an example.

Eluminae: Yeah, way better ! thanks !

Eluminae:
Oops, sorry I missclicked on ready for review, which is not !

kennylevinsen:
Just click "Convert to draft" in upper right near the reviewer section.

sway/input/seatop_touch.c
```diff
@@ -0,0 +1,105 @@
...
+	wlr_cursor_absolute_to_layout_coords(cursor->cursor, event->device,
+			event->x, event->y, &lx, &ly);
+	double sx, sy;
+	node_at_coords(seat, lx, ly, &surface, &sx, &sy);
+
+	if (!surface) { // Must start from the surface
+		return;
+	}
+
+	struct seatop_touch_point_event *point_event = calloc(1, sizeof(struct seatop_touch_point_event));
```
Xyene, line 37: Missing check for allocation failure.

sway/input/cursor.c
```diff
@@ -513,6 +515,7 @@ static void handle_touch_up(struct wl_listener *listener, void *data) {
 		}
 	} else {
 		wlr_seat_touch_notify_up(wlr_seat, event->time_msec, event->touch_id);
```
Xyene, line 517: I think this `wlr_seat_touch_notify_up` belongs in the seatop, now that we have one.

sway/input/seatop_touch.c
```diff
@@ -0,0 +1,105 @@
...
+	point_event->ref_lx = lx;
+	point_event->ref_ly = ly;
+	point_event->ref_con_lx = sx;
+	point_event->ref_con_ly = sy;
+	point_event->touch_id = event->touch_id;
+
+	wl_list_insert(&e->point_events, &point_event->link);
```
Xyene, line 44: We should handle sending `wlr_seat_touch_notify_down` here instead of in `cursor.c` now.

Xyene (changes requested):
Thanks for working on this. I've left a first round of comments.

sway/input/seat.c
```diff
@@ -1552,6 +1553,26 @@ void seatop_pointer_motion(struct sway_seat *seat, uint32_t time_msec) {
...
+	if (seat->seatop_impl->touch_motion) {
+		seat->seatop_impl->touch_motion(seat, event);
+	}
+}
+
+void seatop_touch_down(struct sway_seat *seat, struct wlr_event_touch_down *event) {
+	if (seat->seatop_impl->touch_down) {
+		seat->seatop_impl->touch_down(seat, event);
+	} else {
+		handle_touch_down_start(seat, event);
```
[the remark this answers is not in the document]
Eluminae: Oh nice ! Thats a point I was not understanding when looking at other seatop patches !

sway/input/seatop_touch.c
Eluminae: added

sway/input/cursor.c
Eluminae: I moved those notification and focus action in the seatop

sway/input/seatop_touch.c
Eluminae: This one too

Xyene (commented):
Just a bunch of small nits, thanks for working on this.

I think with the current implementation, the following things are true:

1. the cursor can be moved by pointer input, but while a finger is held down on the screen, the surface under the pointer will not receive movement events (see [this](https://github.com/swaywm/sway/blob/bb3fd0abc56f39e35dc7f4e86f25da1b4a6efbd7/sway/input/cursor.c#L378-L380))
2. pointer buttons likewise are no-ops while a finger is held down on the screen

2\) is fine, but I think we should resolve 1\). To do that, I think you'd want to merge the`seatop_down` and `seatop_touch_down` files, so that the touch down seatop can have pointer/tablet motion handlers. It's fine if `struct seatop_down_event` has a `struct wl_list point_events` that's only ever used for touch input.

Eluminae:
I pushed fixes you mentioned.

> is fine, but I think we should resolve 1). To do that, I think you'd want to merge theseatop_down and seatop_touch_down files, so that the touch down seatop can have pointer/tablet motion handlers. It's fine if struct seatop_down_event has a struct wl_list point_events that's only ever used for touch input.

I agree that those pointer move still must be send to the client.

I tried you solution. The thing is that the seatop_down pointer_motion think we pressed the pointer to begin. It rely on the cursor initial position. I think it make no sense to handle it that way cause this really is a touch event.

I think we should instead implement a pointer_motion in the seatop_touch to just notify with the current cursor position. 

What do you think about this ?

Eluminae:
Okay I thought a little more about this and in fact you right. I have to send cursor position relative to the cont position. I just have to setup the initial cursor position in my begin_touch_down correctly.

Eluminae:
Okay I tried to merge both. I let this change in a dedicated commit.

I dont think this is a good idea and it introduce more complexity to this simple seatop_touch. 

Plus anyway. the cursor disappear when I use the touch screen. I have to click with the mouse to display it again. And this action stop the previous seatop. I think we should just drop this commit :S

sway/input/seatop_down.c
```diff
@@ -110,3 +188,22 @@ void seatop_begin_down(struct sway_seat *seat, struct sway_container *con,
 	container_raise_floating(con);
 	transaction_commit_dirty();
 }
+
+void seatop_begin_touch_down(struct sway_seat *seat, struct sway_container *con,
+		struct wlr_event_touch_down *event, double sx, double sy) {
+	struct sway_cursor *cursor = seat->cursor;
+
+	struct seatop_down_event *e = calloc(1, sizeof(struct seatop_down_event));
+	sway_assert(e, "Unable to allocate seatop_down_event");
```
Xyene, line 197: We can reduce this duplication by:

```c
seatop_begin_down(seat, con, sx, sy);
handle_touch_down(seat, event);
```

You can either get rid of the `time_msec` parameter to `seatop_down` (it's not needed anywhere), or just pass 0 to it for now.

That'll also raise floating containers on touch, which we want.

sway/input/seatop_down.c
```diff
@@ -2,18 +2,95 @@
...
+	}
+
+	wlr_seat_touch_notify_up(seat->wlr_seat, event->time_msec, event->touch_id);
+}
+
+static void handle_touch_down(struct sway_seat *seat,
+		struct wlr_event_touch_down *event) {
+	double sx, sy;
+	struct wlr_surface *surface = NULL;
+	struct sway_node *focused_node = node_at_coords(seat, seat->touch_x, seat->touch_y, &surface, &sx, &sy);
```
Xyene, line 69: Nit: please keep lines < 80 columns wide. (Also line 75.)

sway/input/seat.c
```diff
@@ -1552,6 +1553,24 @@ void seatop_pointer_motion(struct sway_seat *seat, uint32_t time_msec) {
 	}
 }
 
+void seatop_touch_motion(struct sway_seat *seat, struct wlr_event_touch_motion *event) {
```
Xyene, line 1556: Nit: move after the pointer funcs.

Xyene (commented):
Thanks, code mostly looks good.

Regarding mixed pointer/touch input, there isn't really a consensus between other operating systems.

On Windows, the first touch point also updates the cursor position. If the pointer is moved at the same time as the first touch point, the cursor jitters around. This isn't the behavior Sway implements.

On Chrome OS, touch input does not update the cursor position, and the pointer is free to move the cursor. The cursor is hidden on touch down, but becomes visible again if the pointer moves. This is the intended behavior in Sway (see code snippets below, this allows e.g. scrolling a webpage with one hand while also using the pointer to click links).

https://github.com/swaywm/sway/blob/daf9ad1af73a0358fdabc29ca5817511c16e15a4/sway/input/cursor.c#L398
https://github.com/swaywm/sway/blob/daf9ad1af73a0358fdabc29ca5817511c16e15a4/sway/input/cursor.c#L316-L319

If it doesn't work like that something is broken in the code above (it should not require you to click the pointer to make the cursor visible, just move it), but that's not necessary to tackle in this PR unless you feel like it.

Eluminae:
Okay I've done everything you said. It looks to works at least for the touch part. 

The cursor sometime display itself when using the mouse, and sometime it stay hidden. I still dont know really why. Anyway, it does not send motion event to the client. There probably is a case we want to handle in the pointer_motion in cursor.c.

Clicking with the mouse break the seatop but that is probably what we expect. Starting the seatop with a mouse click, then add touches, then up touches will stop the seatop. It may be not what we expect.

Xyene:
> The cursor sometime display itself when using the mouse, and sometime it stay hidden.

That's weird, but likely unrelated to this PR. Unfortunately, I don't own a touchscreen device with which to test this myself.

> Anyway, it does not send motion event to the client.

Hmm, are you sure? `pointer_motion` will call into `seatop_pointer_motion`,

https://github.com/swaywm/sway/blob/bb3fd0abc56f39e35dc7f4e86f25da1b4a6efbd7/sway/input/cursor.c#L378-L380

which should be sending motion events here

https://github.com/swaywm/sway/blob/bb3fd0abc56f39e35dc7f4e86f25da1b4a6efbd7/sway/input/seatop_down.c#L49

> Clicking with the mouse break the seatop but that is probably what we expect.

Maybe. I don't have a good sense for how that is to use (in Chrome OS, it doesn't, you can click multiple times with the pointer while scrolling via touch), but it doesn't sound terrible.

> Starting the seatop with a mouse click hold, then add touches, then up touches will stop the seatop. It may not be what we expect.

Yeah, that's a little weird. Not sure we can do a lot about this with the current structure. The seatop split works well for pointer/tablet input which is single-point, but doesn't generalize well to multi-touch input. Fixing this would be a lot of work.

sway/input/seatop_default.c
```diff
@@ -638,6 +638,21 @@ static void handle_tablet_tool_motion(struct sway_seat *seat,
 	e->previous_node = node;
 }
 
+static void handle_touch_down(struct sway_seat *seat,
+		struct wlr_event_touch_down *event) {
+
```
Xyene, line 643: Nit: extra newline.

sway/input/seatop_down.c
```diff
@@ -85,28 +164,37 @@ static const struct sway_seatop_impl seatop_impl = {
...
 	seat->seatop_impl = &seatop_impl;
 	seat->seatop_data = e;
 
 	container_raise_floating(con);
 	transaction_commit_dirty();
 }
+
+void seatop_begin_touch_down(struct sway_seat *seat, struct sway_container *con,
+		struct wlr_event_touch_down *event, double sx, double sy) {
+
```
Xyene, line 197: Nit: extra newline.

Xyene (commented):
Otherwise LGTM.

sway/input/seatop_default.c
Eluminae: I really like those crazy guys...

Eluminae:
At least now client can implement correctly touch holded event :]

Xyene (dismissed):
Thanks! I'll wait for @emersion to merge, because he actually owns a touchscreen device 🙂

Eluminae:
Okay nice ! It is very easy to test with wev. 

Thanks a lot for your reviews and assistance on this @Xyene !

stacyharper:
Okay maybe it is not that ready yet. I got crashes while touching the swaybar.

stacyharper:
`con` may be NULL. I have to check it before `container_raise_floating`

Xyene:
This should be fixed in https://github.com/swaywm/sway/pull/6418 (there are some other cases that could crash too, see that PR for details). We should get that merged before this one.

stacyharper:
Oh yes, that's exactly that. I'll have both on my device then. Should I remove my check in this MR ?

Xyene:
Yeah, it'll prevent a merge conflict.

stacyharper:
Okay I dont have any crashes with both. But it will cause build conflicts cause I removed the time arg from the default seatop begin. 

I can push this MR with the other commit. We could then merge them in the correct order. 

Should I ?

Xyene:
Since https://github.com/swaywm/sway/pull/6418 has been merged, could you rebase this PR?

stacyharper:
Okay I just rebased, resolved conflict as better as I tried and pushed it back.

I now got crashed. I guess the `container_raise_floating(con)` from `seatop_begin_down` still is in cause. I tried to add back my check condition and I dit crashed as soon as before. But I still had a crash later on my tries.

What should I know about https://github.com/swaywm/sway/pull/6418 ?

sway/input/seatop_down.c
```diff
@@ -99,30 +178,37 @@ static const struct sway_seatop_impl seatop_impl = {
...
 	struct seatop_down_event *e = seat->seatop_data;
 	e->con = con;
 
 	container_raise_floating(con);
 	transaction_commit_dirty();
 }
 
+void seatop_begin_touch_down(struct sway_seat *seat, struct sway_container *con,
+		struct wlr_event_touch_down *event, double sx, double sy) {
+	seatop_begin_down(seat, con, sx, sy);
```
stacyharper, line 201: Here, should I try to use seatop_begin_down or the new seatop_begin_down_on_surface ?

stacyharper: Okay I used this seatop_begin_down_on_surface

stacyharper:
It looks like I fixed my crashes using directly seatop_begin_down_on_surface and not seatop_begin_down. The `con` in my context can be a `NULL` pointer so relying on it looks dangerous.

stacyharper:
Nope, it crash when playing too much with swaybar :(

stacyharper:
This line make Sway to crash `wl_signal_add(&e->surface->events.destroy, &e->surface_destroy);` when touching swaybar then draggin above when there is no window.

stacyharper:
Okay I think I fixed my crashes:

- I had to toggle off `simulating_pointer_from_touch` before starting my seatop in cursor.c `handle_touch_down`. The reason is that it is possible to have a finger or a touch on an empty workspace before a layer shell pop out. Which stuck the cursor in emulation mode. Then the new touch_down touches will behave wrongly and cause crashes.
- I fixed my surface determination in the default seatop `handle_touch_down`. I used the cursor x and y which is wrong and cause a NULL pointer as surface. Attaching destroy event to this pointer was causing a crash

stacyharper:
Hello there ! Someone got an idea of when this could be merged ? :D It prevent us in sxmo to chaim officialy the support of sway

sway/input/seatop_down.c
```diff
@@ -2,21 +2,100 @@
...
+		struct wlr_event_touch_up *event) {
+	struct seatop_down_event *e = seat->seatop_data;
+	struct seatop_touch_point_event *point_event, *tmp;
+
+	wl_list_for_each_safe(point_event, tmp, &e->point_events, link) {
+		if (point_event->touch_id == event->touch_id) {
+			wl_list_remove(&point_event->link);
+			break;
+		}
+	};
```
emersion, line 59: Style: unnecessary `;`

sway/input/seatop_down.c
```diff
@@ -2,21 +2,100 @@
...
+static void handle_touch_motion(struct sway_seat *seat,
+		struct wlr_event_touch_motion *event) {
+	struct seatop_down_event *e = seat->seatop_data;
+
+	struct seatop_touch_point_event *point_event;
+	wl_list_for_each(point_event, &e->point_events, link) {
+		if (point_event->touch_id == event->touch_id) {
+			break;
+		}
+	};
```
emersion, line 38: Style: unnecessary `;`

sway/input/seatop_down.c
```diff
@@ -2,21 +2,100 @@
...
 };
 
+static void handle_touch_motion(struct sway_seat *seat,
+		struct wlr_event_touch_motion *event) {
+	struct seatop_down_event *e = seat->seatop_data;
+
+	struct seatop_touch_point_event *point_event;
+	wl_list_for_each(point_event, &e->point_events, link) {
+		if (point_event->touch_id == event->touch_id) {
+			break;
```
emersion, line 36: What if the point wasn't found in the list? I'd prefer adding ` bool found` variable with `sway_assert(found)` if it's supposed to always be there.

sway/input/seatop_down.c
```diff
@@ -2,21 +2,100 @@
...
+}
+
+static void handle_touch_up(struct sway_seat *seat,
+		struct wlr_event_touch_up *event) {
+	struct seatop_down_event *e = seat->seatop_data;
+	struct seatop_touch_point_event *point_event, *tmp;
+
+	wl_list_for_each_safe(point_event, tmp, &e->point_events, link) {
+		if (point_event->touch_id == event->touch_id) {
+			wl_list_remove(&point_event->link);
```
emersion, line 56: This removes the point event but doesn't free it.

sway/input/seatop_down.c
stacyharper: fixed

sway/input/seatop_down.c
stacyharper: fixed

sway/input/seatop_down.c
stacyharper: okay I added this check

sway/input/seatop_down.c
stacyharper: I added a free(point_event), hoping its enough

stacyharper:
I pushed a rebase I had in my tree. I use this tree for some month now without crashes

sway/input/cursor.c
```diff
@@ -523,19 +519,15 @@ static void handle_touch_motion(struct wl_listener *listener, void *data) {
...
 	wlr_cursor_absolute_to_layout_coords(cursor->cursor, event->device,
 			event->x, event->y, &lx, &ly);
-	double sx, sy;
-	node_at_coords(cursor->seat, lx, ly, &surface, &sx, &sy);
 
-	if (seat->touch_id == event->touch_id) {
-		seat->touch_x = lx;
-		seat->touch_y = ly;
+	seat->touch_x = lx;
+	seat->touch_y = ly;
```
emersion, line 528: Why is this outside of the `seat->touch_id == event->touch_id` condition? It sounds like this potentially mixes up some touch points.

sway/input/seatop_down.c
emersion: Yes, should be enough.

sway/input/seatop_down.c
```diff
@@ -2,21 +2,104 @@
...
+			free(point_event);
+			break;
+		}
+	}
+
+	if (wl_list_empty(&e->point_events)) {
+		seatop_begin_default(seat);
+	}
+
+	wlr_seat_touch_notify_up(seat->wlr_seat, event->time_msec, event->touch_id);
```
emersion, line 69: Nit: should we call this before `seatop_begin_default`?

emersion:
Apart from these comments, this looks good to me (both code and testing).

sway/input/cursor.c
stacyharper: We use the seat->touch_x/y as storage to the default seatop seatop_touch_motion implementation. As the seat->touch_id is the last touch down touch id, moving this inside the condition make both touches to use the last touch point positions.

But that feels more like a comfort choice, so I decided to revert it. So now seat->touch_x/y will only represent the last touch positions (as before). In the seatop_down, I then recompute lx and ly using the event x and y.

I let this change in a dedicated commit to make it easier for review. I'll fixup if everything looks okay to you.

sway/input/seatop_down.c
stacyharper: To be honnest, I have no idea o_O

stacyharper:
> Apart from these comments, this looks good to me (both code and testing).

One things remains to me after some of my last tries:

I added a little bit of logic to prevent the seatop_down to handle touches initiated from other surface than the initial one. But it looks like it doesnt works as much as I expect it. If I touch two time in another surface, it looks like it reset the seatop :S I'm not sure of what I am missing atm

stacyharper:
> If I touch two time in another surface, it looks like it reset the seatop :S I'm not sure of what I am missing atm

to be more precise, my patch works for surfaces that handle touch events. By example, with two wev windows, I cannot interact with the second one when I have already a seatop_down on the first one. But in the case of web + foot, touching foot seems to stop the seatop_down. Maybe it is related to mouse simulation ? cause foot does not handle touch events yet

stacyharper:
sway/input/cursor.c:481 handle_touch_down

```c
→   if (surface && wlr_surface_accepts_touch(wlr_seat, surface)) { 
→   →   if (seat_is_input_allowed(seat, surface)) { 
→   →   →   cursor->simulating_pointer_from_touch = false; 
→   →   →   seatop_touch_down(seat, event); 
→   →   } 
→   } else if (!cursor->simulating_pointer_from_touch && 
→   →   →   (!surface || seat_is_input_allowed(seat, surface))) { 
→   →   // Fallback to cursor simulation. 
→   →   // The pointer_touch_id state is needed, so drags are not aborted when over 
→   →   // a surface supporting touch and multi touch events don't interfere. 
→   →   cursor->simulating_pointer_from_touch = true; 
→   →   cursor->pointer_touch_id = seat->touch_id; 
→   →   double dx, dy; 
→   →   dx = lx - cursor->cursor->x; 
→   →   dy = ly - cursor->cursor->y; 
→   →   pointer_motion(cursor, event->time_msec, event->device, dx, dy, dx, dy); 
→   →   dispatch_cursor_button(cursor, event->device, event->time_msec, 
→   →   →   →   BTN_LEFT, WLR_BUTTON_PRESSED); 
→   } 
```

We should find a way to prevent the `else if` when the surface does not accept touches if a seatop already exists (if it is not the seatop_default ?)

stacyharper:
> We should find a way to prevent the else if when the surface does not accept touches if a seatop already exists (if it is not the seatop_default ?)

Okay, in the last commit I moved the block inside the seatop_default. I also added the necessary check in the seatop_down (if seat can input or if the surface accept them). So now, while a seatop_down is running, we can only add new touches from the same initial surface until all touches are released.

stacyharper:
@emersion does the last patches seems still good to you ? (I'm sorry for pinging but I feels like the MR tends to stall after the moment I send multiple comments to myself as I did)

sway/input/cursor.c
```diff
@@ -463,50 +463,27 @@ static void handle_touch_down(struct wl_listener *listener, void *data) {
...
 	struct sway_seat *seat = cursor->seat;
-	struct wlr_seat *wlr_seat = seat->wlr_seat;
 	struct wlr_surface *surface = NULL;
 
 	double lx, ly;
 	wlr_cursor_absolute_to_layout_coords(cursor->cursor, event->device,
 			event->x, event->y, &lx, &ly);
 	double sx, sy;
-	struct sway_node *focused_node = node_at_coords(seat, lx, ly, &surface, &sx, &sy);
+	node_at_coords(seat, lx, ly, &surface, &sx, &sy);
```
emersion, line 472: The result of this functions seems unused.

emersion:
Not sure I'll have time for a full review soon, I'll defer to @Xyene for now.

sway/input/cursor.c
stacyharper: Yes you right ! I removed unessessary surface, sx and sy

stacyharper:
This is a kindly up :+1: As it works finely and the code looks fine, could it be finally merged ? :D

emersion:
We need someone to do a review first, as the code significantly changed since the last one.

stacyharper:
I rebased this, fixed minor conflicts and adapted the event usage. It compile and still looks fine. I'll run it for sometime to see if there is no issue.

stacyharper:
I bumped sxmo-sway to 1.8. This rebased patch continues to works nicely. I would love to drop sxmo-sway for the next release. Could someone plan to check this ?

emersion:
Had a quick look, acked-by me

sway/input/seatop_down.c
```diff
@@ -2,21 +2,114 @@
...
+		return;
+	}
+
+	double lx, ly;
+	wlr_cursor_absolute_to_layout_coords(seat->cursor->cursor, &event->touch->base,
+			event->x, event->y, &lx, &ly);
+
+	struct seatop_touch_point_event *point_event =
+		calloc(1, sizeof(struct seatop_touch_point_event));
+	sway_assert(point_event, "Unable to allocate point_event");
```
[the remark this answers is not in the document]
stacyharper: done

sway/input/seatop_down.c
```diff
@@ -99,30 +192,38 @@ static const struct sway_seatop_impl seatop_impl = {
...
-		struct wlr_surface *surface, uint32_t time_msec, double sx, double sy) {
+		struct wlr_surface *surface, double sx, double sy) {
 	seatop_end(seat);
 
 	struct seatop_down_event *e =
 		calloc(1, sizeof(struct seatop_down_event));
-	if (!e) {
-		return;
-	}
+	sway_assert(e, "Unable to allocate seatop_down_event");
```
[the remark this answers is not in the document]
stacyharper: done

stacyharper:
> Also `wlr_cursor_absolute_to_layout_coords` gets called twice in down/motion and their caller, this should be calculated once.

Add added lx and ly as arguments on touch_down and touch_motion. Also had to add them on the begin operation.

sway/input/seatop_down.c
```diff
@@ -99,28 +186,38 @@ static const struct sway_seatop_impl seatop_impl = {
...
+
 void seatop_begin_down_on_surface(struct sway_seat *seat,
-		struct wlr_surface *surface, uint32_t time_msec, double sx, double sy) {
+		struct wlr_surface *surface, double sx, double sy) {
 	seatop_end(seat);
 
 	struct seatop_down_event *e =
 		calloc(1, sizeof(struct seatop_down_event));
-	if (!e) {
+	if (!sway_assert(e, "Unable to allocate point_event")) {
```
bl4ckb0ne, line 220: You copy pasted the log from the other `sway_assert`

stacyharper: right, I didn't noticed the var name was used. Fixed

stacyharper: I don't even know why this patch touch those lines anyway. I guess multiple rebase caused this. I should probably restore this simple return.

bl4ckb0ne (approved):

PR #6461, opened 2021-08-25
Fix menu dropdown focus logic in xwayland

### LINKED ISSUE #6324: mouse and dropdown issues in Reaper after sway 1.6 [wlroots 13] in XWayland.
**Sway Version: 1.6**

**Debug Log:** [sway -d2>sway.log <PASTE>](https://paste.oddodd.org/) 

**Configuration File:** Default.

**Reproduce**
1. Open Reaper DAW.
2. Mouse to top drop down menu. 
3. Click __Options__ (drop down appears). 
4. Hover mouse downwards on drop down list.
5. Drop down disappears suddenly without cause. This makes it impossible to select (hover and click) an item "further" down in the list.

**Description:**
Recently, something has broken drop downs when running Reaper in Sway (XWayland that is). 

The symptom is: When you click something that produces a drop down, after scrolling/hovering past the first few items with the mouse, the drop down disappears, making it impossible to use them.
This seems to affect all drop downs in Reaper.

The bug is caused by something that was introduced in Sway version 1.6 and the accompanying wlroots 0.13. Downgrading to Sway version 1.5 and wlroots 0.12 and the issue is gone.

@madskjeldgaard and @kflak, please elaborate on these findings if needed.

salkin-mada:
Not sure what unforeseen results this will have. But it seems to close #6324.
I tested with `code` (vscode) running under `xwayland` and the menu's seemed to work fine there without setting seat focus to NULL before setting to previous.

https://github.com/swaywm/sway/blob/bb3fd0abc56f39e35dc7f4e86f25da1b4a6efbd7/sway/desktop/xwayland.c#L118

At least removing that line makes the drop down menu's in **Reaper** usable again.

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -111,10 +111,24 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
 			return;
 		}
 
+        // Try to find another unmanaged surface from the same process to pass
+        // focus to. This is necessary because some applications (e.g. Jetbrains
+        // IDEs) represent their multi-level menus as unmanaged surfaces, and
+        // when closing a submenu, the main menu should get input focus.
+        struct sway_xwayland_unmanaged *current;
+        wl_list_for_each(current, &root->xwayland_unmanaged, link) {
+            struct wlr_xwayland_surface *prev_xsurface =
+                    current->wlr_xwayland_surface;
+            if (prev_xsurface->pid == xsurface->pid &&
+                    wlr_xwayland_or_surface_wants_focus(prev_xsurface)) {
+                seat_set_focus_surface(seat, prev_xsurface->surface, false);
+                return;
+            }
+        }
+
 		// Restore focus
 		struct sway_node *previous = seat_get_focus_inactive(seat, &root->node);
 		if (previous) {
-			// Hack to get seat to re-focus the return value of get_focus
 			seat_set_focus(seat, NULL);
 			seat_set_focus(seat, previous);
 		}


kennylevinsen:
https://github.com/swaywm/sway/pull/2103/commits/74799937cf95205c2ff16da9dc7dcaea47ce7116

Xyene:
/cc @BrassyPanache

I have mixed feelings about this patch. I suspect we're missing something fundamental, because other compositors do not have such complex X11 refocusing logic, and presumably (has anyone tried Weston?) Reaper works there. The last focus issues we had with JetBrains IDEs ended up being due to us not implementing ICCCM focus hints; I wonder if that is at play here.

This "try to find another unmanaged surface from the same process to pass focus to" logic is a hack I initially wrote without understanding why it was necessary, so I'd be a little sad to see it get reintroduced. It may also be informative to trace through the code and understand why removing `seat_set_focus(seat, NULL);` is necessary in Sway, and why removing it fixes things in Reaper.

salkin-mada:
@Xyene Nice to see you joining in. We have been loosely throwing around the "lets find a universal fix" talk.
And Indeed. Making a more condense and real fix (understanding the inner workings of these focus issues/symptoms) seems way more intriguing than just adding back the "try to find another unmanaged surface from the same process to pass focus to" attitude.

BrassyPanache:
I can share how I approached the issue swaywm/wlroots#2604 (related pulls of swaywm/wlroots#2605 and swaywm/sway#5948).  The issue sounds very similar in nature - IntelliJ drop down menus were behaving incorrectly.

The initial fix I put in place was for swaywm/sway#3007 where I took a seemingly similar approach to what's taken here.  To make the fix I inserted logging statements until I could work out the specific change required to get IntelliJ to work and then made the change.  I then verified that other applications worked as expected and called it a day.

Of course this change broke with the next IntelliJ release.  I was not at the time privy to the [ICCCM](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.7) specifications regarding input focus.  I found the following table most insightful.

image

I would just check that everything is consistent to the above table from both the sway (and wlroots) side and the Reaper side.  If other window managers (e.g. GNOME) work it is a good indication that the change needs to occur in sway or wlroots.

I am by no means an expert on these matters - just sharing my prior experience :smile:

elmarsto:
I'm shocked that this work item is still open so long after a solving PR was offered.  REAPER in particular is broken.

I now have the choice of abandoning one of:
- Sway
- Wayland (falling back to i3 above)
- Linux (for the MacOS or Windows build)
- REAPER (for Bitwig)

madskjeldgaard:
> I'm shocked that this work item is still open so long after a solving PR was offered.  REAPER in particular is broken. While I understand this is volunteer effort, the treatment of this PR seems out-of-keeping with the rest of the project, which seems vibrant.
> 
> I now have the choice of abandoning one of:
> - Sway
> - Wayland (falling back to i3 above)
> - Linux (for the MacOS or Windows build)
> - REAPER (for Bitwig)

I abandoned Sway because of this. I need Reaper every day so this is a deal-breaker unfortunately.

emersion:
"Me too", "just merge it", "I won't use Sway unless this is fixed" type of comments don't help. If you want to help, read the discussion above and find a proper fix. A workaround like the one done currently in this PR won't get merged.

salkin-mada:
Fixed in #6764 🎉

martin-greentrax:
Hey everyone,
really appreciate your great work with swaywm, I found this issue after spending a long time trying to get popups in a java/xwayland app TWS of interactive brokers to not vanish and found it very similar. Could you kindly check out my bug report here:  https://github.com/swaywm/sway/issues/7823

Thank you :)

PR #6475, opened 2021-09-03
Add 'output render_bit_depth [8|10]' command

mstoeckl:
~Companion to [wlroots MR !3333](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3333).~ ~See commit message (and the documentation for the introduced command in the man page) for what this does.~

This PR provides an `output <name> render_bit_depth [8|10]` command which can be used to make sway render window surfaces onto a 10 bit-per-channel buffer, if sway's backend and rendering system support it. The default bit depth remains 8, because that is supported essentially everywhere.

This is a rather specific interface; it does not guarantee that 10 bpc will be used, and even if a 10-bpc render buffer is used, it is not guaranteed that a monitor will be sent 10 bpc color data. Also, direct scan out for fullscreen windows might temporarily change the format of the buffers being sent to an output.

I am not tied to the design of the `render_bit_depth` command; if anyone has a better alternative, please suggest it.

This command can easily be generalized to set 5-bit (RGB565) or 16-float-bit channel depths; however:
* Essentially all programs produce 8 bpc data, so 5 bpc will lose color information; also, as far as I know, all monitors already accept 8 bpc color data. There could in theory be monitors where sending 6bpc instead of 8bpc allows for higher frame rates, but I don't know of any.
* 16f used to work for me with the wlroots drm backend, but started failing (with frozen display) a few months ago, possibly after a kernel or Mesa change. This bit depth can be added later by someone who can test that it doesn't break anything in Sway or wlroots.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -1,5 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include <assert.h>
+#include <drm_fourcc.h>
 #include <stdbool.h>
 #include <string.h>
 #include <sys/socket.h>
@@ -67,6 +68,7 @@ struct output_config *new_output_config(const char *name) {
 	oc->subpixel = WL_OUTPUT_SUBPIXEL_UNKNOWN;
 	oc->max_render_time = -1;
 	oc->adaptive_sync = -1;
+	oc->render_bit_depth = RENDER_BIT_DEPTH_DEFAULT;
 	return oc;
 }
 
@@ -112,6 +114,9 @@ void merge_output_config(struct output_config *dst, struct output_config *src) {
 	}
 	if (src->adaptive_sync != -1) {
 		dst->adaptive_sync = src->adaptive_sync;
+	}
+	if (src->render_bit_depth != RENDER_BIT_DEPTH_DEFAULT) {
+		dst->render_bit_depth = src->render_bit_depth;
 	}
 	if (src->background) {
 		free(dst->background);
@@ -351,6 +356,23 @@ static int compute_default_scale(struct wlr_output *output) {
 	return 2;
 }
 
+/* Lists of formats to try, in order, when a specific render bit depth has
+ * been asked for. The second to last format in each list should always
+ * be XRGB8888, as a reliable backup in case the others are not available;
+ * the last should be DRM_FORMAT_INVALID, to indicate the end of the list. */
+static const uint32_t *bit_depth_preferences[] = {
+	[RENDER_BIT_DEPTH_8] = (const uint32_t []){
+		DRM_FORMAT_XRGB8888,
+		DRM_FORMAT_INVALID,
+	},
+	[RENDER_BIT_DEPTH_10] = (const uint32_t []){
+		DRM_FORMAT_XRGB2101010,
+		DRM_FORMAT_XBGR2101010,
+		DRM_FORMAT_XRGB8888,
+		DRM_FORMAT_INVALID,
+	},
+};
+
 static void queue_output_config(struct output_config *oc,
 		struct sway_output *output) {
 	if (output == root->noop_output) {
@@ -436,6 +458,22 @@ static void queue_output_config(struct output_config *oc,
 		sway_log(SWAY_DEBUG, "Set %s adaptive sync to %d", wlr_output->name,
 			oc->adaptive_sync);
 		wlr_output_enable_adaptive_sync(wlr_output, oc->adaptive_sync == 1);
+	}
+
+	if (oc && oc->render_bit_depth != RENDER_BIT_DEPTH_DEFAULT) {
+		const uint32_t *fmts = bit_depth_preferences[oc->render_bit_depth];
+		assert(fmts);
+
+		for (size_t i = 0; fmts[i] != DRM_FORMAT_INVALID; i++) {
+			wlr_output_set_render_format(wlr_output, fmts[i]);
+			if (wlr_output_test(wlr_output)) {
+				break;
+			}
+
+			sway_log(SWAY_DEBUG, "Preferred output format 0x%08x "
+				"failed to work, falling back to next in "
+				"list, 0x%08x", fmts[i], fmts[i + 1]);
+		}
 	}
 }
 
--- a/sway/sway-output.5.scd
+++ b/sway/sway-output.5.scd
@@ -157,6 +157,21 @@ must be separated by one space. For example:
 	adaptive sync can improve latency, but can cause flickering on some
 	hardware.
 
+*output* <name> render_bit_depth 8|10
+	Controls the color channel bit depth at which frames are rendered; the
+	default is currently 8 bits per channel.
+
+	Setting higher values will not have an effect if hardware and software lack
+	support for such bit depths. Successfully increasing the render bit depth
+	will not necessarily increase the bit depth of the frames sent to a display.
+	An increased render bit depth may provide smoother rendering of gradients,
+	and screenshots which can more precisely store the colors of programs
+	which display high bit depth colors.
+
+	Warnings: this can break screenshot/screencast programs which have not been
+	updated to work with different bit depths. This command is experimental,
+	and may be removed or changed in the future.
+
 # SEE ALSO
 
 *sway*(5) *sway-input*(5)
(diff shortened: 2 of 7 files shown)


sway/config/output.c
```diff
@@ -351,6 +356,22 @@ static int compute_default_scale(struct wlr_output *output) {
 	return 2;
 }
 
+/* Lists of formats to try, in order, when a specific render bit depth has
+ * been asked for. The last format in the list should always be XRGB8888,
+ * as a reliable backup in case the others are not available. */
```
emersion, line 361: I'd prefer to not fallback to 8-bit when the user explicitly asks for 10-bit. Similar to how we handle modes which cannot be enabled.

sway/config/output.c
```diff
@@ -351,6 +356,22 @@ static int compute_default_scale(struct wlr_output *output) {
...
+static const uint32_t *bit_depth_preferences[] = {
+	[RENDER_BIT_DEPTH_8] = (const uint32_t []){
+		DRM_FORMAT_XRGB8888,
+		0
```
emersion, line 365: Nit: use `DRM_FORMAT_INVALID` to make it more explicit that this magic value doesn't refer to a valid format

sway/config/output.c
```diff
@@ -437,6 +458,22 @@ static void queue_output_config(struct output_config *oc,
 			oc->adaptive_sync);
 		wlr_output_enable_adaptive_sync(wlr_output, oc->adaptive_sync == 1);
 	}
+
+	if (oc && oc->render_bit_depth != RENDER_BIT_DEPTH_DEFAULT) {
+		const uint32_t *fmts = bit_depth_preferences[oc->render_bit_depth];
+		assert(fmts);
+
+		for (int i = 0; fmts[i]; i++) {
```
emersion, line 466: Nit: `size_t`

sway/sway-output.5.scd
```diff
@@ -157,6 +157,20 @@ must be separated by one space. For example:
 	adaptive sync can improve latency, but can cause flickering on some
 	hardware.
 
+*output* <name> render_bit_depth 8|10
```
emersion, line 160: Config design question: do we know yet how this will interact with other related options? Like KMS "max bpc" (we don't have "force bpc" or "current bpc" yet unfortunately, but it's been discussed), potential HDR options, etc?

Is the bit depth a good thing to let users configure? Are there some other more meaningful or high level config options we could let the user configure and which would dictate low-level things such as render format and KMS "max bpc"?

Maybe it's a bit too soon to ask these questions. Also, I don't know a lot about color management and HDR. If everything is blurry still, maybe we can label this with an "experimental" disclaimer, saying it can go away or change in the future.

emersion: Some relevant comments were in https://github.com/swaywm/wlroots/pull/3162#issuecomment-913005351. I'm not sure I like the mode string extension idea too much, I'd prefer not to end up with hard-to-decipher mode strings, and instead use clearer output commands, even if multiple commands might change the hw limitations and such.

pjanx: On X11, DefaultDepth 30 just extends sRGB (or whatever profile the X11 display claims to have, I think it's a property, either way see https://www.freedesktop.org/software/colord/intro.html) all over the 30 bits. Only Chromium is being a jerk and needs to have the profile set, or it becomes super low contrast, anyway.

pjanx: I have no clue about HDR.

pjanx: Color management should not otherwise differ between 10-bit and 8-bit, I fail to see how.

sway/config/output.c
mstoeckl: Fixed.

sway/config/output.c
mstoeckl: Done.

sway/config/output.c
mstoeckl: > Similar to how we handle modes which cannot be enabled.

It looks like sway does different things here, depending on the backend.
* With the DRM backend, it picks the closest listed mode to the one asked for; `https://github.com/swaywm/sway/blob/master/sway/config/output.c#L257-L275`
* Setting a --custom mode, or a modeline, tries to set the specified mode (and possibly freezes the screen if the mode isn't allowed.)
* With the Wayland backend, it always accepts the mode.

>  not fallback to 8-bit when the user explicitly asks for 10-bit.

If I pick this behavior, then Sway will crash if no 10 bit buffer formats are supported, because it does not have any buffers to render onto. This _is_ an experimental command, so it might be fine to do this as long as the documentation warns about it.

sway/sway-output.5.scd
mstoeckl: > Maybe it's a bit too soon to ask these questions. Also, I don't know a lot about color management and HDR. If everything is blurry still, maybe we can label this with an "experimental" disclaimer, saying it can go away or change in the future.

I've added a short disclaimer.

> Config design question: do we know yet how this will interact with other related options? Like KMS "max bpc" (we don't have "force bpc" or "current bpc" yet unfortunately, but it's been discussed), potential HDR options, etc?

> Is the bit depth a good thing to let users configure? Are there some other more meaningful or high level config options we could let the user configure and which would dictate low-level things such as render format and KMS "max bpc"?

I think that `output <> render_bit_depth` is a command similar to `output <> modeline`, in that most people will never need to use it, and would instead use hypothetical commands like:
```
output <name> hdr disabled|enabled
output <name> wire_bit_depth 6|8|10|12|14|16
```
In this example, `output <name> hdr` would imply a render format with 16 or 16-float bpc; `wire_bit_depth` would correspond to "force bpc", and would set the most compact render format with the necessary precision. A backend/renderer combination which supports color management might require higher bpc formats. But there would still be cases where one might want to have a 16 bit render buffer when the other settings suggest 8-bit is sufficient (for higher precision colors when screen recording (?)), or force an 8-bit render buffer despite the renderer/backend preferring 16f for color management (to use less memory and power). These are where an `output <> render_bit_depth` override might be useful.

pjanx: > which supports color management might prefer higher bpc formats

Definitely. When there is a mismatch between the content colour space and the display, any transformation will typically lose noticeable information when using the low output bit depth of 8.

I'm curious, what reason do you have in mind with screen recording?

mstoeckl: > I'm curious, what reason do you have in mind with screen recording?

* As mentioned in https://github.com/swaywm/wlroots/pull/3162#issuecomment-913005351, sometimes programs produce 10 or 16 bit buffers; rendering to a higher depth, and taking a screenshot with the depth, will avoid losing color information from what the program provided.
* Screen sharing: if computer A has an 8 bit monitor, but computer B has a 10 bit monitor, then when looking at computer A's screen from computer B, it would be nice to have colors available at full precision.
* The RGB->YUV conversion when recording a video from a screen often loses a very small amount of information; capturing a 10 bpc screen, instead of an 8bpc screen, can slightly improve quality.

pjanx: I see, thanks for the explanation.

sway/config/output.c
emersion: > It looks like sway does different things here, depending on the backend.

Hm that's a good point. I forgot the bits about the DRM backend. We should look at a way to unify all of this. Since black screens aren't great, maybe we should always fallback to something that works, and integrate with swaynag to print warnings "your output mode couldn't be enabled" and "10-bit not available for this output".

Ideally we should also add a `render_bit_depth` field in `swaymsg -t get_outputs`, to let users query the current bit depth. But this can come later.

Also there's the wlr-output-management protocol, which never expects such fallbacks. It expects the configuration to fail if it's not possible. The wlr-output-management client itself will then try something else.

> If I pick this behavior, then Sway will crash if no 10 bit buffer formats are supported, because it does not have any buffers to render onto.

I wouldn't expect it to crash, but maybe it won't light up outputs configured with 10-bit render formats? But per the above, I've changed my mind.

sway/sway-output.5.scd
emersion: @mstoeckl, this design looks good to me, thanks!

emersion (approved):
LGTM, thanks!

emersion:
Opened #6681 and #6682 as follow-up issues.

sway/sway-output.5.scd
pjanx: By the way, I've found this, it seems to explain HDR well https://www.x.org/wiki/Events/XDC2016/Program/xdc-2016-hdr.pdf

PR #6478, opened 2021-09-03
commands: update split none command for a047b5ee4

RPigott:
@emersion https://github.com/swaywm/sway/pull/4935#issuecomment-912839599

Oops.


emersion (approved):
Eh, thanks for the fix!

RPigott:
Yeah, my bad. I stopped running this patch a while ago after compatibility with i3 improved, so I never noticed it staled again.

Issue #6147, opened 2021-03-30
Firefox popups not rendered on some multi-output setups


emersion:
1. Setup one output on the left with scale=1, and one output on the right with scale!=1
2. Start Firefox with its Wayland backend
3. From time to time some popups will not be displayed (I think it happens more often with the addon popups when Firefox is displayed on the scale!=1 output). The clicks will still go through just fine.

luispabon:
I made this recording here to show some weirdness related to this. The laptop display is scale 2, top display scale 1. I have a third scale 1 display to the right.

https://www.youtube.com/watch?v=63nUUXE6Qt0

If I position firefox very close to the bottom edge of the top display, the popup will display. 

There's also that weird thing on the bottom display once I move the floating FF window where a chunk of the ff window is rendered and flickers.

I'm on master sway (1d62d6bf) and wlroots (b89bcffe)

emersion:
> There's also that weird thing on the bottom display once I move the floating FF window where a chunk of the ff window is rendered and flickers.

Hm, this seems like a separate bug, I've seen this with other clients as well. Can you open a new bug report?

luispabon:
Of course.

minhduc0711:
This might be related https://bugzilla.mozilla.org/show_bug.cgi?id=1684197. 
In my case, the popups don't show up on the monitor with scale == 1.

emersion:
Is the scale=1 monitor on the right?

luispabon:
Yes. Top and top right scale 1, bottom scale 2
```
{
    output eDP-1 mode 3840x2160 scale 2 position 0,1081
    output "Iiyama North America PL2410HD 11046A0B03094" mode 1920x1080 scale 1 position 0,0
    output "Samsung Electric Company SMB1920NW HMDB101546" mode 1440x900 scale 1 position 1920,0
}
```

kennylevinsen:
Does anyone have a reproduction on another compositor?

luispabon:
I've just checked on Wayfire and the issue does not happen there.

luispabon:
Neither does it happen in gnome 3.38.

emersion:
This doesn't seem to be a damage tracking related issue, because I can still reproduce it with `-Ddamage=rerender`.

emersion:
Additional note: the coords in `xdg_popup.configure` seem to be reasonable. So probably something completely stops the popup (or a subsurface) from being rendered.

luispabon:
[...]

PR #6479, opened 2021-09-03
view: wait until surface is mapped to send enter

rpigott:
Based on observation, this is what weston does. https://github.com/swaywm/sway/issues/6147#issuecomment-912857546

I'm not sure if it's useful to send enter on surface creation before the surface is mapped, however the wayland protocol seems to encourage it, saying explicitly enter should be send on "creation".

On the one hand, it seems like common sense the application should know the required scale before it attaches a buffer, so the old way makes sense. But on the other hand the surface hasn't actually been shown so hasn't entered any outputs.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -1063,6 +1063,11 @@ static void view_child_handle_surface_map(struct wl_listener *listener,
 		wl_container_of(listener, child, surface_map);
 	child->mapped = true;
 	view_child_damage(child, true);
+
+	struct sway_workspace *workspace = child->view->container->pending.workspace;
+	if (workspace) {
+			wlr_surface_send_enter(child->surface, workspace->output->wlr_output);
+	}
 }
 
 static void view_child_handle_surface_unmap(struct wl_listener *listener,
@@ -1106,11 +1111,6 @@ void view_child_init(struct sway_view_child *child,
 
 	wl_signal_add(&view->events.unmap, &child->view_unmap);
 	child->view_unmap.notify = view_child_handle_view_unmap;
-
-	struct sway_workspace *workspace = child->view->container->pending.workspace;
-	if (workspace) {
-		wlr_surface_send_enter(child->surface, workspace->output->wlr_output);
-	}
 
 	view_child_init_subsurfaces(child, surface);
 }


sway/tree/view.c
```diff
@@ -1063,6 +1063,11 @@ static void view_child_handle_surface_map(struct wl_listener *listener,
 		wl_container_of(listener, child, surface_map);
 	child->mapped = true;
 	view_child_damage(child, true);
+
+    struct sway_workspace *workspace = child->view->container->pending.workspace;
+    if (workspace) {
+            wlr_surface_send_enter(child->surface, workspace->output->wlr_output);
+    }
```
jniedrauer, line 1070: Whitespace mismatch here, should use tabs to match the rest of the file.

RPigott: Oops, thanks.

kennylevinsen:
Maybe we should open an issue upstream to discuss this? Sounds like either gdk is broken or the protocol spec lacks clarification...

RPigott:
> Sounds like either gdk is broken or the protocol spec lacks clarification...

It's both imo. Gdk is being dumb even if you only consider the intended use and ignore the breakage that happens when firefox tries to hold onto the popup surface. I expect the issue would vanish if Gdk would just stop doing the show/hide nonsense so I'm going to see what becomes of https://gitlab.gnome.org/GNOME/gtk/-/issues/4198. I should probably explain this error there as well.

For this, I just wanted someone to test that it actually fixed the issue. I feel like if the protocol meant for enter events to be sent when the surface was mapped, it would have said mapped, but yeah a discussion is probably in order. I can start that discussion on the gitlab, but I've got to sleep soon 'cause it's really late here.

emersion:
Ref https://gitlab.freedesktop.org/wayland/wayland/-/issues/133

shone:
I can confirm that this commit fixes the issue described on #6147 for me:

- Sway built without this commit: a context menu did not appear after the first right click after opening Firefox. Firefox also got stuck in a state where it would no longer re-render its tabs/window decoration unless focus were switched between Firefox and another application.
- Sway built with this commit: context menus are appearing, and Firefox continues to render its tabs/window decorations. Still have a different issue where the damage/repainting of popups/context menus appears glitched-out though.

I have one laptop display at scale 2, and I'm opening Firefox on an external monitor monitor at scale 1. MOZ_ENABLE_WAYLAND=1 is set, and about:support indicates that the window protocol is wayland.

Thank you for your work on this issue :pray: It's been driving me bananas for months.

Jellyfrog:
This works for me too, context menus etc, and firefox is way snappier, related or not

kennylevinsen:
> This works for me too

Please use reactions for "can confirm"/"works for me"/etc., and reserve comments for development discussion and requested information to avoid noise. We get too many notifications as is...

hkupty:
Couldn't get it building in Arch. The error I get is [pasted here](https://pastebin.com/R3qEu99L)
I'm using:
-`wlroots-git`[AUR pkg](https://aur.archlinux.org/packages/wlroots-git/)
- `sway-git` [AUR pkg](https://aur.archlinux.org/packages/sway-git/), modified to use this branch:
```diff
diff --git a/PKGBUILD b/PKGBUILD
index 5069e20..b374996 100644
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -2,7 +2,7 @@
 # Contributor: Antonin Décimo <antonin dot decimo at gmail dot com>
 pkgname=sway-git
 _pkgname=sway
-pkgver=r6607.28cadf55
+pkgver=r6756.54e3c496
 pkgrel=1
 license=("MIT")
 pkgdesc="Tiling Wayland compositor and replacement for the i3 window manager"
@@ -39,7 +39,7 @@ optdepends=(
 backup=(etc/sway/config)
 arch=("i686" "x86_64")
 url="https://swaywm.org"
-source=("${pkgname%-*}::git+https://github.com/swaywm/sway.git"
+source=("${pkgname%-*}::git+https://github.com/RPigott/sway.git#branch=mapb4enter"
 	50-systemd-user.conf)
 sha512sums=('SKIP'
             '57590bc0d14c87289a4a9cd67991c6a841e54244d2a6186b5da5a08e633de2e8631959fa8c77ede211b0a5f315d920f2c1350951a53d6f2e9e81859056cb3c9e')
```

kennylevinsen:
> Couldn't get it building in Arch. The error I get is [pasted here](https://pastebin.com/R3qEu99L)

That's because the branch is out of date and needs to be rebased before it is merged.

RPigott:
Guys this isn't the patch you are looking for. It's marked draft/don't merge for a reason – it depends on the outcome of the wayland discussion.

The fix you want is https://github.com/swaywm/sway/issues/6147#issuecomment-920761907. They should work just as well if not better.

primalmotion:
> Guys this isn't the patch you are looking for. It's marked draft/don't merge for a reason – it depends on the outcome of the wayland discussion.
> 
> The fix you want is [#6147 (comment)](https://github.com/swaywm/sway/issues/6147#issuecomment-920761907). They should work just as well if not better.

Sure thing, but in the meantime, it makes sway + FF usable, so thanks for it anyway :)

@hkupty I simply got the patch from https://patch-diff.githubusercontent.com/raw/swaywm/sway/pull/6479.patch added it to the upstream PKGBUILD, without changing the URL pf the source

PR #6481, opened 2021-09-04
Fix overly-wide mark textures

danieldg:
The width of the texture needs to be calculated using the string that is
actually displayed in the texture.


emersion (approved):
Good catch, thanks!

PR #6482, opened 2021-09-04
swaybar: Prioritize hotspot events to bar bindings

joanbm:
This is consistent with i3bar's behaviour, and for example, allows binding a command to button1, while still being able to click on tray icons or other zones on the bar's status line which may have their own bindings.

E.g., in Sway, without this commit, this config. makes tray icons unclickable:

```
    bar {
        # ...
        bindsym button1 exec swaynag -m You_clicked_the_tray._Want_some_help?
    }
```

But the same configuration in i3 (with i3-nagbar) keeps tray items clickable.

### DIFF
--- a/swaybar/input.c
+++ b/swaybar/input.c
@@ -166,14 +166,13 @@ static void wl_pointer_button(void *data, struct wl_pointer *wl_pointer,
 		return;
 	}
 
-	if (check_bindings(seat->bar, button, state)) {
-		return;
+	if (state == WL_POINTER_BUTTON_STATE_PRESSED) {
+		if (process_hotspots(output, pointer->x, pointer->y, button)) {
+			return;
+		}
 	}
 
-	if (state != WL_POINTER_BUTTON_STATE_PRESSED) {
-		return;
-	}
-	process_hotspots(output, pointer->x, pointer->y, button);
+	check_bindings(seat->bar, button, state);
 }
 
 static void workspace_next(struct swaybar *bar, struct swaybar_output *output,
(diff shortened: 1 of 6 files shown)


joanbm:
Note that after this commit, if the bar's status command is handling click events (`click_events` in the protocol), and a status block on the bar is clicked/scrolled, and that status block don't have any binding associated, but there is a global bar binding, it will *not* be called. This is also the same behaviour as i3, and as far as I can tell it can not be easily avoided, because the status command doesn't tell you whether it handled the event or not.

swaybar/input.c
```diff
@@ -224,15 +223,14 @@ static void workspace_next(struct swaybar *bar, struct swaybar_output *output,
 static void process_discrete_scroll(struct swaybar_seat *seat,
 		struct swaybar_output *output, struct swaybar_pointer *pointer,
 		uint32_t axis, wl_fixed_t value) {
-	// If there is a button press binding, execute it, skip default behavior,
-	// and check button release bindings
 	uint32_t button = wl_axis_to_button(axis, value);
-	if (check_bindings(seat->bar, button, WL_POINTER_BUTTON_STATE_PRESSED)) {
-		check_bindings(seat->bar, button, WL_POINTER_BUTTON_STATE_RELEASED);
+	if (process_hotspots(output, pointer->x, pointer->y, button)) {
```
nmschulte, line 227: Should this path now also `check_bindings(seat->bar, button, WL_POINTER_BUTTON_STATE_RELEASED);`, as you see a number of the others later in the routine?

nmschulte (approved):
I haven't a suitable test scenario but the code changes look good.

swaybar/input.c
joanbm: Mmmm, actually as far as I can see there isn't anything wrong with this press/release code on `process_discrete_scroll` you commented on, since for scroll, the press/release come immediately after one another, and currently the hotspot callback do anything for the release case.

However, it prompted me to take a deeper look into how regular (pointer, non-scroll) press/release bindsyms are handled, and I found that actually, this PR was not handling *those* release events correctly.

Since hotspots (workspaces, status bar, tray, etc.) now handle pressed events, it makes sense that they also handle released events (even if they actually do nothing), in order to "block" a `bindsym --release buttonX` which is set in the bar from triggering. Otherwise weird ensues because if you simply press/release on a hotspot without moving the mouse, you'd just see the `bindsym --release buttonX` but not a `bindsym buttonX`. This is also how i3 handles this case, and **I added a commit to this PR fixing this**.

For example, suppose we define a `bindsym button1 exec something` and a `bindsym --release button1 exec somethingelse` on the bar, and then we left click on a tray icon. Then the behavior is:

|                | Tray event | Cmd for bindsym button1 | Cmd for bindsym --release button1 |
|----------------|------------|-------------------------|-----------------------------------|
| Latest Sway    | No         | Yes                     | Yes                               |
| This PR before | Yes        | No                      | Yes                               |
| Latest i3      | Yes        | No                      | No                                |
| This PR now    | Yes        | No                      | No                                |

I also tested the various combinations (press and release on bar, press and release on hotspot, press on bar and release on hotspot, press on hotspot and release on bar) and it matches how i3 handles those.

joanbm:
Ping - I have been using this since submitting the PR without issues.

BTW, my use case for this is binding some common applications (e.g. file manager, mute) to the free space in the middle of the bar, which makes them easily accessibly with the mouse only by Fitts' Law.

emersion (approved):
Thanks!

nmschulte:
Thanks for fixing this!

PR #6483, opened 2021-09-04
Safer screnlock support

danieldg:
See https://github.com/swaywm/wlroots/issues/2706 and https://github.com/swaywm/wlroots/pull/3165.

We may also want to add a command to automatically run something if the lockscreen crashes.  Currently you can use a `bindsym --locked` to relaunch swaylock, but that's not very obvious or user-friendly.  Avoiding a crash-relaunch 100% CPU loop will be a problem with the simple solution, however.

Related swaylock branch: https://github.com/danieldg/swaylock/tree/lock-protocol - no MR there yet.


danieldg:
Replaced by https://github.com/swaywm/sway/pull/6879

PR #6496, opened 2021-09-07
view: fix child position calc

vyivel:
Previously, the position was calculated incorrectly for nested
subsurfaces.

---

Test client: [subsurface-client/grandchild-damage](https://codeberg.org/vyivel/subsurface-client)

### DIFF
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -183,7 +183,7 @@ struct sway_xwayland_unmanaged {
 struct sway_view_child;
 
 struct sway_view_child_impl {
-	void (*get_root_coords)(struct sway_view_child *child, int *sx, int *sy);
+	void (*get_view_coords)(struct sway_view_child *child, int *sx, int *sy);
 	void (*destroy)(struct sway_view_child *child);
 };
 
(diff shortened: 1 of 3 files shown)


vyivel:
Actually, I forgot to test it for popups, but it _should_ work fine.

kennylevinsen:
Popups made of nested subsurfaces (firefox context menus) have been a weak spot before, so I'd recommend testing them.

RPigott:
Uh, the grandchild-damage client source appears to be missing.

vyivel:
>Popups made of nested subsurfaces (firefox context menus) have been a weak spot before, so I'd recommend testing them.

Tested with Firefox (without WebRender) and "normal" xdg-popups (gtk3-demo-application), both work correctly.

>Uh, the grandchild-damage client source appears to be missing.

Ah, forgot to add a file. Sorry about that, fixed now.

vyivel:
Updated to take view position in account.

emersion (approved):
This new version looks good to me. It moves the view geometry stuff to `view_child_damage` so that it's applied to all types of view children. The subsurface function is now always recursive, which allows to have the same logic whether the parent is a xdg-popup or a subsurface itself.

Thanks!

PR #6497, opened 2021-09-08
build: subproject support for wayland-protocols

lbonn:
As in wlroots: https://github.com/swaywm/wlroots/pull/2953.

To build sway on my system with outdated wayland-protocols (1.20):

- add wlroots, wayland-protocols, seatd as subprojects
- run `meson setup --force-fallback-for wayland-protocols build`
- ...

Without the force-fallback, sway finds wayland-protocols from the system, which does not satisfy wlroots (>=1.21) and the setup fails without a good error message.

I'll add a note in the wiki if this is merged.

### DIFF
--- a/protocols/meson.build
+++ b/protocols/meson.build
@@ -1,4 +1,4 @@
-wl_protocol_dir = wayland_protos.get_variable(pkgconfig: 'pkgdatadir')
+wl_protocol_dir = wayland_protos.get_variable('pkgdatadir')
 
 wayland_scanner_dep = dependency('wayland-scanner', required: false, native: true)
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM! Since Meson 0.58.0 we can use this syntax, and we already require 0.58.1.

> (since 0.58.0) added varname as first positional argument. It is used as default value for cmake, pkgconfig, configtool and internal keyword arguments.

emersion:
Thanks!

lbonn:
About the second problem, I realized that there is no point to not require the same version of wayland-protocols that wlroots needs, so I opened #6498.

PR #6498, opened 2021-09-08
build: require wayland-protocols>=1.21

lbonn:
Follow-up to #6497, to avoid the meson trickery.

This prevents unresolvable dependencies during meson setup if a system version is found and has a version >=1.14, <1.21.


emersion:
Hm, but Sway itself doesn't need 1.21.

What happens? Does Meson choose wayland-protocols <= 1.21 because of the Sway dep, then errors out in the wlroots subproject?

May be a good idea to fill a Meson issue for this, if there isn't one already.

emersion:
> I realized that there is no point to not require the same version of wayland-protocols that wlroots needs

Yeah, but it's a bit annoying to keep both in sync. Soon wlroots will require 1.22, and we'll need a new Sway update...

lbonn:
> What happens? Does Meson choose wayland-protocols <= 1.21 because of the Sway dep, then errors out in the wlroots subproject?

Yep, I believe that's what's going on.

> May be a good idea to fill a Meson issue for this, if there isn't one already.

That makes sense, I'll try to have another look at their bug tracker and investigate further.

RPigott:
Is there a meson issue for this?

emersion:
https://github.com/swaywm/sway/commit/0b4e3d39eb7af94617a3f95995e2b0ba8bb5513f bumps the dep version.

eli-schwartz:
> Hm, but Sway itself doesn't need 1.21.
> 
> What happens? Does Meson choose wayland-protocols <= 1.21 because of the Sway dep, then errors out in the wlroots subproject?
> 
> May be a good idea to fill a Meson issue for this, if there isn't one already.

It's a diamond dependencies problem. You want to resolve the same version of any one dependency across the entire project build, but the first one you resolve doesn't universally satisfy the constraints.

So the question then is, do you:
- error out
- re-resolve the entire build, taking into account that wayland-protocols must fallback
- artificially require the same version of the dependency for every subproject

One alternative would be to do `dependency('wayland-protocols', version: ...)` only after wlroots is looked up first.

Then, wlroots gets first chance at finding a suitable version of wayland-protocols, and if it has to fallback to a subproject fallback, that subproject fallback then globally overrides the dependency and the superproject (sway) is diverted into the subproject fallback too.

emersion:
I agree it's a tough problem given Meson's design. I've opened https://github.com/swaywm/sway/pull/6769.

PR #6504, opened 2021-09-10
Simplify swaybar/swaynag scaling code

### LINKED ISSUE #6486: Simplify HiDPI rendering code
Instead of having `* scale` all over the place, we can just use `cairo_scale` to set the scale factor.

Caduser2020:
Use `cairo_scale` to set the scale factor, removing redundant multiplications by `output->scale`.

Tested on my setup using fractional scaling, but also works properly for integer and no scaling (as tested with nested sway).

Fixes #6486

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -53,22 +53,21 @@ static uint32_t render_status_line_error(struct render_context *ctx, double *x) 
 		return 0;
 	}
 
-	uint32_t height = output->height * output->scale;
+	uint32_t height = output->height;
 
 	cairo_t *cairo = ctx->cairo;
 	cairo_set_source_u32(cairo, 0xFF0000FF);
 
-	int margin = 3 * output->scale;
-	double ws_vertical_padding =
-		output->bar->config->status_padding * output->scale;
+	int margin = 3;
+	double ws_vertical_padding = output->bar->config->status_padding;
 
 	char *font = output->bar->config->font;
 	int text_width, text_height;
 	get_text_size(cairo, font, &text_width, &text_height, NULL,
-			output->scale, false, "%s", error);
+			1, false, "%s", error);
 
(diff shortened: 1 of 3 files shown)


Caduser2020:
Also, swaynag and swaybar have multiple `-Wconversion` warnings, should a follow-up PR address these conversion warnings?

swaybar/render.c
```diff
@@ -165,11 +163,11 @@ static enum hotspot_event_handling block_hotspot_callback(
...
-			x - (double)hotspot->x / output->scale,
-			y - (double)hotspot->y / output->scale,
-			(double)hotspot->width / output->scale,
-			(double)hotspot->height / output->scale,
-			output->scale, button);
+			x - (double)hotspot->x,
+			y - (double)hotspot->y,
+			(double)hotspot->width,
+			(double)hotspot->height,
+			1, button);
```
emersion, line 170: Hm, not sure this part about input event handling is quite right. `process_hotspots` multiplies incoming pointer coordinates by the scale. If hotspot coords are now unscaled, maybe we need to remove that multiplication?

Also I think we should keep sending `output->scale` as the second-to-last parameter.

emersion:
Thanks for working on this! Nice to see all of this error-prone scaling stuff go away. :)

> Also, swaynag and swaybar have multiple -Wconversion warnings, should a follow-up PR address these conversion warnings?

Yeah, PRs very much welcome for this too! It might take some effort to enable this flag on the whole Sway codebase. I'm completely open to take an incremental approach and convert one binary at a time.

swaybar/render.c
Caduser2020: Thanks for pointing out the input code. I completely missed testing mouse events/buttons also (touch points untested but should work fine). Addressed in latest commit.

emersion (approved):
LGTM, thanks!

PR #6507, opened 2021-09-12
Log exec failures

emersion:
This can happen when the swaybar or its status command isn't found.
Instead of silently doing nothing, log the error.

### DIFF
--- a/sway/config/bar.c
+++ b/sway/config/bar.c
@@ -238,6 +238,8 @@ static void invoke_swaybar(struct bar_config *bar) {
 					bar->swaybar_command ? bar->swaybar_command : "swaybar",
 					"-b", bar->id, NULL};
 			execvp(cmd[0], cmd);
+
+			sway_log_errno(SWAY_ERROR, "Failed to start swaybar");
 			_exit(EXIT_FAILURE);
 		}
 		_exit(EXIT_SUCCESS);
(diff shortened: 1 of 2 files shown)


Emantor (approved):

sway/config/bar.c
```diff
@@ -238,6 +238,8 @@ static void invoke_swaybar(struct bar_config *bar) {
 					bar->swaybar_command ? bar->swaybar_command : "swaybar",
 					"-b", bar->id, NULL};
 			execvp(cmd[0], cmd);
+
+			sway_log_errno(SWAY_ERROR, "Failed to start swaybar");
```
kennylevinsen, line 242: Nit: "swaybar" vs "status command", even though they're both "xyz_command" in the config.

kennylevinsen:
Maybe change the swaybg exec error logging to match

sway/config/bar.c
emersion: Hm, I don't understand. This `execvp` is about starting swaybar, not the status command.

kennylevinsen: You add an error log for exec'ing status_command in the {R, and one for exec'ing swaybar_command (this line), but one message says "status command" while the other says "swaybar".

ammgws:
Why not merge it?

emersion:
I don't care too much about this PR, sorry.

PR #6511, opened 2021-09-13
readme: link to gamja for IRC

emersion:
gamja offers a better experience than Kiwi.

Obviously I'm 100% biased. Completely fine to not make the switch if someone feels differently.


kennylevinsen (approved):
LGTM

PR #6513, opened 2021-09-13
Rename pango_printf to render_text

emersion:
This avoids using the pango_ prefix, reserved for functions coming
from the Pango library.


kennylevinsen (approved):
LGTM

PR #6514, opened 2021-09-13
swaynag: fix pointer button events on scaled outputs

xdavidwu:
#6504 simplified rendering code by setting scaling at cairo, but
that also changed button position records into ones without scale
multiplication, breaking button events. This fixes it by not multiplying
scale on events handling as well.


emersion (approved):
Good catch, thanks!

PR #6515, opened 2021-09-13
readme: mention seatd

emersion:
When seatd is used, it isn't necessary to suid the sway binary.


kennylevinsen (approved):
LGTM, although maybe at some point in a distant future we should switch to recommending SUID seatd-launch over SUID sway.

PR #6516, opened 2021-09-13
Allow running as root when not using setuid bit

jbeich:
For debugging (bypass DRM_AUTH), recovery (single user mode) and in some embedded scenarios (root shell instead of sudo/doas) it maybe desirable to run the compositor (and everything else within) as root. Currently, this is not supported to prevent foot-shooting when using `builtin` server in libseat.

```
$ sudo sway -c /dev/null
00:00:00.090 [sway/main.c:166] Unable to drop root (we shouldn't be able to restore it after setuid), refusing to start
```

Similar to #5669 but setuid vs. root is detected automatically. This makes Sway behave like wlroots compositors that don't drop priveleges (e.g., river, phoc, gamescope). Alternatively, `drop_permissions()` can be removed in favor of seatd-launch(1).

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -150,7 +150,12 @@ static void log_kernel(void) {
 
 
 static bool drop_permissions(void) {
+	if (getuid() == 0 || getgid() == 0) {
+		sway_log(SWAY_INFO, "Running as root user, this is dangerous");
+		return true;
+	}
 	if (getuid() != geteuid() || getgid() != getegid()) {
+		sway_log(SWAY_INFO, "setuid/setgid bit detected, dropping permissions");
 		// Set the gid and uid in the correct order.
 		if (setgid(getgid()) != 0) {
 			sway_log(SWAY_ERROR, "Unable to drop root group, refusing to start");


emersion:
Yeah, now that we have seatd-launch I think it would make sense to just drop all of that suid logic. @kennylevinsen thoughts?

kennylevinsen:
My main worry would be users SUID'ing sway right now to use rhe direct/built-in backend, who would experience the change as having sway keep running as full root without as much as a warning. Principle of least surprise and all that.

If we're comfortable with just warning about it in the release notes, then I'd suggest killing it. Maybe wait a little for 0.6.1 rollout though.

emersion:
Hm right, the transition period is a bit dangerous indeed. Maybe a gradual migration would be better? e.g. error out if we're suid root for a release or two, then drop the logic completely?

Hjdskes:
Linking this related commit here for posterity: e1db1f8218998c428e8b087dda6660449ef2891a

emersion:
Superseded by #7044

PR #6517, opened 2021-09-14
swaybar: fix tray icons rendering on scaled outputs

xdavidwu:
#6504 simplified rendering code by setting scale at cairo, but not the code
for rendering tray icons, making it draws at scale of expected size.
This fixes by resetting scale before drawing icons, and restoring it
afterward.

### DIFF
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -520,7 +520,7 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
 	struct swaybar_hotspot *hotspot = calloc(1, sizeof(struct swaybar_hotspot));
 	hotspot->x = *x;
 	hotspot->y = 0;
-	hotspot->width = output->height;
+	hotspot->width = padded_surface_size;
 	hotspot->height = output->height;
 	hotspot->callback = icon_hotspot_callback;
 	hotspot->destroy = free;
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -494,23 +494,34 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
 	}
 
 	int padded_size = icon_size + 2*padding;
-	*x -= padded_size;
-	int y = floor((height - padded_size) / 2.0);
+	double padded_surface_size = (double) padded_size / output->scale;
+	double buffer_x = (*x * output->scale) - padded_size;
+	int buffer_y = floor((height - padded_size) / 2.0);
+
+	// Reset scale to draw with prepared icon as source surface. Drawing with
+	// source surface directly with scale applied will end up blurry in HiDPI.
+	cairo_save(cairo);
+	cairo_identity_matrix(cairo);
 
 	cairo_operator_t op = cairo_get_operator(cairo);
 	cairo_set_operator(cairo, CAIRO_OPERATOR_OVER);
-	cairo_set_source_surface(cairo, icon, *x + padding, y + padding);
-	cairo_rectangle(cairo, *x, y, padded_size, padded_size);
+	cairo_set_source_surface(cairo, icon, buffer_x + padding, buffer_y + padding);
+	cairo_rectangle(cairo, buffer_x, buffer_y, padded_size, padded_size);
 	cairo_fill(cairo);
 	cairo_set_operator(cairo, op);
+
+	// Restore scale
+	cairo_restore(cairo);
+
+	*x -= padded_surface_size;
 
 	cairo_surface_destroy(icon);
 
 	struct swaybar_hotspot *hotspot = calloc(1, sizeof(struct swaybar_hotspot));
 	hotspot->x = *x;
 	hotspot->y = 0;
-	hotspot->width = height;
-	hotspot->height = height;
+	hotspot->width = output->height;
+	hotspot->height = output->height;
 	hotspot->callback = icon_hotspot_callback;
 	hotspot->destroy = free;
 	hotspot->data = strdup(sni->watcher_id);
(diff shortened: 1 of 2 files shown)


emersion:
Hm, can't we draw tray icons without caring about the scale at all?

xdavidwu:
If I understand the code correctly, `tray_padding` config is currently defined in device pixels. I think with current config definition rendering that part in device pixels is more convenient, also I cannot find any way to fill with image source surface in device pixels without resetting the scale in cairo.

nmschulte:
See https://github.com/swaywm/sway/pull/6518/commits/7497911862ce04c9ff90ca2f185df2e60bb5b542

xdavidwu:
> See #6518 : 7497911.

That appoarch make an SVG icon blurrish for me. The image surface is scaled to smaller size (without respect the scale, to what should be rendered at 1x) with `cairo_image_surface_scale` call.

nmschulte:
> Hm, can't we draw tray icons without caring about the scale at all?

@emersion the reason this path needs to concern with output scaling is because the icons are loaded from pixmaps at which point the scale needs to be known.

swaybar/tray/item.c
```diff
@@ -494,23 +494,33 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
...
+	// restore scale
+	cairo_restore(cairo);
+
+	*x -= padded_surface_size;
+
 	cairo_surface_destroy(icon);
 
 	struct swaybar_hotspot *hotspot = calloc(1, sizeof(struct swaybar_hotspot));
 	hotspot->x = *x;
 	hotspot->y = 0;
```
nmschulte, line 521:
```suggestion
	hotspot->y = (double)y / output->scale;
	hotspot->width = padded_size;
	hotspot->height = padded_size;
```

nmschulte (commented):
Setting `.y` seems appropriate, and `.width`/`.height` are actually `padded_size` at this point (vs the original "`size`" (`height`...).

`padded_size` is the best icon integer-up-scaled toward (or fractional-down-scaled to) `target_size` ("full tray height without padding"; `icon_size`) with padding applied.

that is to say, in other words:
not all tray items are the same width (not all tray item icons are the same size): it is dependent on the available iconography

xdavidwu:
> Setting .y seems appropriate, and .width/.height are actually padded_size at this point (vs the original "size" (height...).

If you mean the hotspot (used in pointer button handling), before #6504, it was set to `height * height` square at `(x, 0)`, that is, all clicks inside that square box, with size of full bar height. This PR makes it does the same thing (#6504  also changed the pointer representation from device pixel size to surface size, thus the commit changes it to `output->height` instead of `height`). I don't think we should introduce unrelated behavior changes in this PR (like excluding the padding).

nmschulte:
> I don't think we should introduce unrelated behavior changes in this PR (like excluding the padding).

Addressing the issue separately is fine: the issue is a bug in the existing (and your) codes.
`x` is decremented by the width of the actual rendered/resolved icon (with padding), but the hotspot's width is set to the "ideal"icon width (`target_size`; the bar height minus padding), not the one actually determined and used (`padded_size`, or `icon_size` to exclude the padded region).  This can lead to overlapping hotspots for (adjacent) tray items.

xdavidwu:
Oh you meant changing to `padded_size * padded_size` ( / scale). I thought you want `icon_size * icon_size` ( / scale) (I mistook the y with y for icon (without padding), sorry).

Indeed this is a bug that can lead to unexpected behavior and need to be fixed.

swaybar/tray/item.c
xdavidwu: Fixed with a9ad38c, with following changes:

`padded_size` is in device pixels, so it also need to be divided by scale (`padded_source_size`).

I think using full height is more friendly, so I did not change y, height parts.

nmschulte:
@xdavidwu I have pushed another changeset that resolves all of the issues I found, and applies the scaling fix a little differently than this mechanism; I have updated the description with details: https://github.com/swaywm/sway/pull/6518#issue-733529597

emersion:
So, do we have a consensus which of the two PRs solves the issue at hand better?

> tray_padding config is currently defined in device pixels

That seems like an oversight. I wouldn't mind a follow-up PR to change it to surface-local coords. But yeah better to keep it as-is for this bugfix PR, so that we can more easily revert that change if needed.

nmschulte:
> So, do we have a consensus which of the two PRs solves the issue at hand better?

I believe the one here is more succinct, and the difference is elided when `tray_padding` is changed from device pixels. 

I'll split the other fixes from #6518 and help with the follow-up PR if @xdavidwu doesn't want to do it.

swaybar/tray/item.c
```diff
@@ -494,23 +494,33 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
 	}
 
 	int padded_size = icon_size + 2*padding;
-	*x -= padded_size;
+	double padded_surface_size = (double) padded_size / output->scale;
+	double x_in_pixels = (*x * output->scale) - padded_size;
```
emersion, line 498: Name nit: "in pixels" is a bit vague, as there are logical (surface-local) pixels and physical (buffer-local) pixels.

Maybe use something like `buffer_x` instead?

swaybar/tray/item.c
```diff
@@ -494,23 +494,33 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
...
 	int y = floor((height - padded_size) / 2.0);
 
+	// reset scale
```
emersion, line 501: Can we add a small comment here explaining why we're doing this? Something like "Icons need to be rendered in buffer-local space, otherwise they end up being blurry"

emersion:
Superseded by #6518

PR #6518, opened 2021-09-14
Swaybar scaling fixes

nmschulte:
Fixes some fallout from #6504:
- tray item rendering, due to scale changes, and creeping bad logic
- tray_padding vs min-height computation

---
This branch/solution; notice the icons are all crisp, they are all vertically centered and evenly spaced (but not evenly sized width-wise!):
(https://github.com/nmschulte/sway/commit/fa75a39295326aa54a32666b01c6d088ba8920a9)
swaybar-scaling-fixes-with-debug

One of my prior/first attempts, not really knowing how the scaling is applied or the intricacies of libcairo and wayland-protocols/sway; notice the aliasing/blur due to the reversal/removal of scale:
old-swaybar-scaling-fixes-with-debug

`master` with the same/similar diagnostic as my dev; note the large icons:
(https://github.com/nmschulte/sway/commit/8425299c80dac8e80d7f06739f4d9342c406bbe6)
master-with-debug

### DIFF
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -493,24 +493,36 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
 		cairo_destroy(cairo_icon);
 	}
 
-	int padded_size = icon_size + 2*padding;
-	*x -= padded_size;
-	int y = floor((height - padded_size) / 2.0);
+	double descaled_padding = (double)padding / output->scale;
+	double descaled_icon_size = (double)icon_size / output->scale;
+
+	int size = descaled_icon_size + 2 * descaled_padding;
+	*x -= size;
+	int icon_y = floor((output->height - size) / 2.0);
 
 	cairo_operator_t op = cairo_get_operator(cairo);
 	cairo_set_operator(cairo, CAIRO_OPERATOR_OVER);
-	cairo_set_source_surface(cairo, icon, *x + padding, y + padding);
-	cairo_rectangle(cairo, *x, y, padded_size, padded_size);
+
+	cairo_matrix_t scale_matrix;
+	cairo_pattern_t *icon_pattern = cairo_pattern_create_for_surface(icon);
+	// TODO: check cairo_pattern_status for "ENOMEM"
+	cairo_matrix_init_scale(&scale_matrix, output->scale, output->scale);
+	cairo_matrix_translate(&scale_matrix, -(*x + descaled_padding), -(icon_y + descaled_padding));
+	cairo_pattern_set_matrix(icon_pattern, &scale_matrix);
+	cairo_set_source(cairo, icon_pattern);
+	cairo_rectangle(cairo, *x, icon_y, size, size);
 	cairo_fill(cairo);
+
 	cairo_set_operator(cairo, op);
 
+	cairo_pattern_destroy(icon_pattern);
 	cairo_surface_destroy(icon);
 
 	struct swaybar_hotspot *hotspot = calloc(1, sizeof(struct swaybar_hotspot));
 	hotspot->x = *x;
 	hotspot->y = 0;
-	hotspot->width = height;
-	hotspot->height = height;
+	hotspot->width = size;
+	hotspot->height = output->height;
 	hotspot->callback = icon_hotspot_callback;
 	hotspot->destroy = free;
 	hotspot->data = strdup(sni->watcher_id);
--- a/swaybar/tray/tray.c
+++ b/swaybar/tray/tray.c
@@ -116,8 +116,8 @@ uint32_t render_tray(cairo_t *cairo, struct swaybar_output *output, double *x) {
 		}
 	} // else display on all
 
-	if ((int) output->height*output->scale <= 2*config->tray_padding) {
-		return 2*config->tray_padding + 1;
+	if ((int)(output->height * output->scale) <= 2 * config->tray_padding) {
+		return (2 * config->tray_padding + 1) / output->scale;
 	}
 
 	uint32_t max_height = 0;


swaybar/tray/item.c
```diff
@@ -493,24 +493,36 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
...
+	double descaled_padding = (double)padding / output->output_scale;
+	double descaled_icon_size = (double)icon_size / output->output_scale;
+
+	int size = descaled_icon_size + 2 * descaled_padding;
+	*x -= size;
+	int icon_y = floor((output->height - size) / 2.0);
 
 	cairo_operator_t op = cairo_get_operator(cairo);
 	cairo_set_operator(cairo, CAIRO_OPERATOR_OVER);
-	cairo_set_source_surface(cairo, icon, *x + padding, y + padding);
```
nmschulte, line 502: `cairo_set_source_surface` doesn't allow control of the CTM scaling directly, but the underlying pattern it obscures does allow it and could be retrieved and manipulated.  Instead, just do what `cairo_set_source_surface` does ourselves and setup the CTM as we desire (now with scaling).

---

https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-set-source-surface
>This is a convenience function for creating a pattern from surface and setting it as the source in cr with cairo_set_source().

https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-set-source
>Note: The pattern's transformation matrix will be locked to the user space in effect at the time of cairo_set_source(). **This means that further modifications of the current transformation matrix will not affect the source pattern.** See cairo_pattern_set_matrix().

nmschulte:
animation showing the slight discrepancy in vertical alignment (from `master`) on #6517, versus the corrected vertical middle-alignment logic included here in #6518.  this also demonstrates clearly that the items (due to their available symbols/size) are not equal size/width.

positioning-comparison

sorry for the mega images... :enhance:

swaybar/tray/item.c
```diff
@@ -493,24 +493,36 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
...
 
-	int padded_size = icon_size + 2*padding;
-	*x -= padded_size;
-	int y = floor((height - padded_size) / 2.0);
+	double descaled_padding = (double)padding / output->output_scale;
+	double descaled_icon_size = (double)icon_size / output->output_scale;
+
+	int size = descaled_icon_size + 2 * descaled_padding;
+	*x -= size;
+	int icon_y = floor((output->height - size) / 2.0);
```
xdavidwu, line 501: Note that this calcuates in surface coord, while #6517 calculates in device pixel coord (as before scale simplification). I think this is the source of that vertical alignment difference.

nmschulte: I agree. I think the maths in my branch here can be simplified/reduced with some careful thought but that seemed like yak-shaving compared to other things in `swaybar` design that could be improved.

The pixel-specified `tray_padding`, and there-in `output_scale` is necessarily spread through the icon-wrangling and painting works.

The logic is two-fold in that **after finding the best "base" icon for the current height** (an interative process repeating as the bar inits/populates/grows/operates), it is:
1. scaled as close as possible to the desired height avoiding artifacting/aliasing (using only power-of-two enlargement or resampling reduction), while account for and enabling pixel-specified `tray_padding`
1. aligned centered vertically within the (bar) height, again while respecting `tray_padding` (e.g. max-height implications, reaching all the way back to icon generation supporting "hiDPI"/scaling)

I don't know the implications of saving/restoring a cairo state, but I'm betting it's more than the added `cairo_matrix_t` operation.

The only thing I still don't quite wrap my head around is the needing the translation for both the source (surface) pattern and again in the `cairo_rectangle` call. Maybe it makes sense if the scale is 1 but .. I digress; this is why there's probably a (repeated) factor that can be removed in my maths if someone cares.

xdavidwu: >  the implications of saving/restoring a cairo state

side note: in #6517, they can be replaced by scale(1/scale) and scale(scale) back, but that involves float points, so I save the states and reset transformation matrix to identical matrix to avoid that. (An implication is that additional transformation, like translation, will also be ignored during icon rendering, but there is none and the current design passes `x` all the way thus translation is likely not needed)

My concern is that aligning icon+padding to surface coord may look bad on fractional scale, creating unneeded aliasing. Aligning to device pixels should be better (and possibly more accurate on making it center).

nmschulte: As well, I don't think the code-paths properly support fractional scaling; at least not in swaybar.

nmschulte:
updated; pruned to directly relevant commits (see #6588 and #6589)

nmschulte:
It would be nice to have this fixed upstream. [bump.]

swaybar/tray/item.c
```diff
@@ -493,24 +493,36 @@ uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
...
-	cairo_rectangle(cairo, *x, y, padded_size, padded_size);
+
+	cairo_matrix_t scale_matrix;
+	cairo_pattern_t *icon_pattern = cairo_pattern_create_for_surface(icon);
+	// TODO: check cairo_pattern_status for "ENOMEM"
+	cairo_matrix_init_scale(&scale_matrix, output->scale, output->scale);
+	cairo_matrix_translate(&scale_matrix, -(*x + descaled_padding), -(icon_y + descaled_padding));
+	cairo_pattern_set_matrix(icon_pattern, &scale_matrix);
+	cairo_set_source(cairo, icon_pattern);
+	cairo_rectangle(cairo, *x, icon_y, size, size);
```
emersion, line 513: Hmm, I'm not quite sure I understand this call.

The coordinates used here seem to be negated by `cairo_matrix_translate` above? Shouldn't `size` not include the padding?

nmschulte: Just as before, the pattern that is setup is painted into the padded region.  Now though, the padding is applied within the pattern to be painted, allowing for pixel-spec.  We want to remove allowing for pixel-spec though, but that was requested as a separate (post-this-fix) change.

I didn't and am not sure I do still fully understand the transform / translation setup, but it _is_ the same technique that `cairo_set_source_surface` does.  I was confused that the source surface needs the full x/y translation (origin setup) in addition to being passed to `cairo_rectangle`, but that's how it works I guess.

emersion: Hm, reading `_cairo_default_context_set_source_surface`, indeed you're right. Well, let's just merge this for now and see how we can simplify it in follow-up PRs.

emersion (approved):
LGTM, thanks!

PR #6519, opened 2021-09-14
config.in: use swaybar -B instead of -b

noocsharp:
'-b' spawns a terminal, which is unnecessary for this use case


nmschulte (changes requested):
The commit message is wrong: you mean swaynag, not swaybar.

PR #6520, opened 2021-09-15
swaybar: properly draw urgent block right border

### LINKED ISSUE #6510: status line blocks marked as urgent are missing the right border
- **Sway Version:**
  - `sway version 1.6.1`

- **Configuration File:**
The differences between my test config and the default one on arch:
```diff
--- /etc/sway/config	2021-06-24 12:26:44.000000000 -0500
+++ /home/jeremy/.config/sway/config	2021-09-12 13:28:42.176277391 -0500
@@ -202,15 +202,17 @@
 # Read `man 5 sway-bar` for more information about this section.
 bar {
     position top
+    separator_symbol " "
 
     # When the status_command prints a new line to stdout, swaybar updates.
     # The default just shows the current date and time.
-    status_command while date +'%Y-%m-%d %l:%M:%S %p'; do sleep 1; done
+    status_command ~/urgent-status.sh
 
     colors {
         statusline #ffffff
         background #323232
         inactive_workspace #32323200 #32323200 #5c5c5c
+        urgent_workspace #ffffff #ff0000 #ffffff
     }
 }
```

The simplest way I knew to generate the urgent condition was to use this script, which I put in `~/urgent-status.sh`:
```bash
#!/usr/bin/env bash

echo '{ "version": 1, "click_events": true, "cont_signal": 18, "stop_signal": 19 }'

echo '[[]'

while sleep 1; do
    # It's difficult to see the problem if it's the last item.
    echo ',[{"full_text": "foo", "urgent": true},{"full_text": "bar"}]'
done
```

- **Description:**
If the `urgent_workspace` border color is set and a statusline block is marked as urgent, the right border is missing. It sounds like that setting should only be applied to the workspace section of the bar, but it's also used for status line blocks.

status

nmschulte:
fixes #6510

introduced via #3287 -- https://github.com/swaywm/sway/pull/3287/files#diff-a1e918ce0bc71f4f7934767541319e724a51a34a5418ecdc286065e50921eda4L239
uncovered via #3394 -- https://github.com/swaywm/sway/pull/3394/files#diff-a1e918ce0bc71f4f7934767541319e724a51a34a5418ecdc286065e50921eda4R258

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -307,7 +307,7 @@ static uint32_t render_status_block(struct render_context *ctx,
 	render_text(cairo, config->font, 1, block->markup, "%s", text);
 	x_pos += width;
 
-	if (block->border && block->border_right > 0) {
+	if (border_color && block->border_right > 0) {
 		x_pos += margin;
 		render_sharp_line(cairo, border_color, x_pos, y_pos,
 				block->border_right, render_height);


swaybar/render.c
```diff
@@ -307,7 +307,7 @@ static uint32_t render_status_block(struct render_context *ctx,
 	render_text(cairo, config->font, 1, block->markup, "%s", text);
 	x_pos += width;
 
-	if (block->border && block->border_right > 0) {
```
nmschulte, line 310: zero is "transparent black" is a no-op. smart.

emersion (approved):
Good catch, thanks!

PR #6521, opened 2021-09-15
swaybar: properly draw blocks with transparent black border

nmschulte:
while the draw itself is a no-op, alignment must still be accounted

image

the blue bordered blocks demonstrate the transparent black (`0x0`) border support working (the background is blue)

```
bar {
    status_command ./top.sh
    position bottom
    separator_symbol '¦'
    icon_theme Adwaita
    colors {
        urgent_workspace #FF7F00 #1F1F1F #FFFFFF
    }
}
bar {
    status_command ./bottom.sh
    position bottom
    icon_theme Adwaita
    separator_symbol '¦'
}
```
```bash
#!/usr/bin/env bash
echo '{ "version": 1, "click_events": true, "cont_signal": 18, "stop_signal": 19 }'
echo '[[]'
while sleep 1; do
    echo ',[
    {"full_text": "UDD", "urgent": true,  "border": "#1F7F1FFF"},
    {"full_text": "UTB", "urgent": true,  "border": "#00000000"},
    {"full_text": "U--", "urgent": true},
    {"full_text": "N--", "urgent": false},
    {"full_text": "NTB", "urgent": false, "border": "#00000000"},
    {"full_text": "NDD", "urgent": false, "border": "#1F7F1FFF"}
    ]'
done
```

```bash
#!/usr/bin/env bash
echo '{ "version": 1, "click_events": true, "cont_signal": 18, "stop_signal": 19 }'
echo '[[]'
while sleep 1; do
    echo ',[
    {"full_text": "U--", "urgent": true},
    {"full_text": "N--", "urgent": false},
    {"full_text": "UTB", "urgent": true,  "border": "#00000000"},
    {"full_text": "NTB", "urgent": false, "border": "#00000000"},
    {"full_text": "UDD", "urgent": true,  "border": "#1F7F1FFF"},
    {"full_text": "NDD", "urgent": false, "border": "#1F7F1FFF"}
    ]'
done
```

this does not alter margin application in general: non-bordered blocks still receive no margin

### DIFF
--- a/swaybar/render.c
+++ b/swaybar/render.c
@@ -14,6 +14,7 @@
 #include "swaybar/ipc.h"
 #include "swaybar/render.h"
 #include "swaybar/status_line.h"
+#include "log.h"
 #if HAVE_TRAY
 #include "swaybar/tray/tray.h"
 #endif
@@ -215,11 +216,11 @@ static uint32_t render_status_block(struct render_context *ctx,
 	}
 
 	*x -= width;
-	if ((block->border || block->urgent) && block->border_left > 0) {
+	if ((block->border_set || block->urgent) && block->border_left > 0) {
 		*x -= (block->border_left + margin);
 		block_width += block->border_left + margin;
 	}
-	if ((block->border || block->urgent) && block->border_right > 0) {
+	if ((block->border_set || block->urgent) && block->border_right > 0) {
 		*x -= (block->border_right + margin);
 		block_width += block->border_right + margin;
 	}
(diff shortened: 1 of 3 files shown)


swaybar/render.c
```diff
@@ -286,6 +286,8 @@ static uint32_t render_status_block(struct render_context *ctx,
 		render_sharp_line(cairo, border_color, x_pos, y_pos,
 				block->border_left, render_height);
 		x_pos += block->border_left + margin;
+	} else if (block->urgent) {
+		x_pos += block->border_left + margin;
```
emersion, line 290: What if `block->border` is transparent? Don't we still need to add the offset even if the block is not urgent?

I'd be completely fine with dropping the `border_color` "optimization". Pretty sure it doesn't make any meaningful difference.

nmschulte: Correct, and this code does that: the only way to get a border without being urgent is to specify one on the block; `border_color` will reference `urgent_color` (if `urgent` set) or `border->color` otherwise.

It's not clear what value `border->color` has if the block has no specified border; I thought I covered all the permutations but I will verify this.

emersion (approved):
LGTM, thanks!

PR #6527, opened 2021-09-17
Avoid unecessary font metric calculations

WhyNotHugo:
Prior to 62d90a8e, titlebar's font height (and other related values)
would change any time any titlebar's content changed, so these values
were recalculated each time any titlebar's content changed (or a new
titlebar was created).

However, since the above was merge, these values no longer change so
often and we only need to recalculate them when the configured font
changes (and stop calling `config_update_font_height` each time
titlebars are rendered).

This commit removes all the unecessary calls to this function and avoids
all those unecessary calculations. Whenever the font strays from the
default value, the `font` command is called, and it calls
`config_update_font_height`, which is enough to keep the value always up
to date.

I've also added a default value to the `font_baseline` config, since
otherwise that's zero for setups that don't explicitly specify a font.

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -113,10 +113,8 @@ void get_text_metrics(const char *font, int *height, int *baseline) {
 	cairo_t *cairo = cairo_create(NULL);
 	PangoContext *pango = pango_cairo_create_context(cairo);
 	PangoFontDescription *description = pango_font_description_from_string(font);
-	PangoFontMetrics *metrics;
-
 	// When passing NULL as a language, pango uses the current locale.
-	metrics = pango_context_get_metrics(pango, description, NULL);
+	PangoFontMetrics *metrics = pango_context_get_metrics(pango, description, NULL);
 
 	*baseline = pango_font_metrics_get_ascent(metrics) / PANGO_SCALE;
 	*height = *baseline + pango_font_metrics_get_descent(metrics) / PANGO_SCALE;
--- a/sway/commands/reload.c
+++ b/sway/commands/reload.c
@@ -48,7 +48,6 @@ static void do_reload(void *data) {
 	}
 	list_free_items_and_destroy(bar_ids);
 
-	config_update_font_height();
 	root_for_each_container(rebuild_textures_iterator, NULL);
 
 	arrange_root();
--- a/sway/commands/title_format.c
+++ b/sway/commands/title_format.c
@@ -23,6 +23,5 @@ struct cmd_results *cmd_title_format(int argc, char **argv) {
 	}
 	view->title_format = format;
 	view_update_title(view, true);
-	config_update_font_height();
 	return cmd_results_new(CMD_SUCCESS, NULL);
 }
--- a/sway/config.c
+++ b/sway/config.c
@@ -236,8 +236,6 @@ static void config_defaults(struct sway_config *config) {
 	config->default_layout = L_NONE;
 	config->default_orientation = L_NONE;
 	if (!(config->font = strdup("monospace 10"))) goto cleanup;
-	config->font_height = 17; // height of monospace 10
-	config->font_baseline = 11; // baselint of monospace 10
 	config->urgent_timeout = 500;
 	config->focus_on_window_activation = FOWA_URGENT;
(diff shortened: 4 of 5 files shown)


WhyNotHugo:
A very minor caveat: [the default value for `font_height` is currently `17`](https://github.com/WhyNotHugo/sway/blob/583ef7375a9eb3750e972a42b11a94ebdecaf492/sway/config.c#L239).

When using the default font (e.g.: no font specified in the config file),`config_update_font_height` immediately sets this to `15` as soon as the first titlebar is rendered.

This patch has a small side effect for this particular scenario; if no font is specified we never call `config_update_font_height`, so that value will continue being `17`. Changing the hardcoded default to 15 will make sure users see no change in behavior.

However, I'm not 100% certain that this magic number isn't simply different on my setup due to my installed fonts -- so maybe it 17 _is_ the right value for you guys? I know the semantic meaning of `font_height` has changed a bit over time, but that default value has just remained the same.

Would you prefer if I update the default value to `15`, or just leave that as-is?

emersion:
"monospace" is not an actual font, it's a name which matches the preferred monospace font on the system. I think?

Hm, maybe we can drop the pre-computed values? And always compute the sizes on startup even if no custom font is set?

WhyNotHugo:
> "monospace" is not an actual font, it's a name which matches the preferred monospace font on the system. I think?

I believe you're right.

> Hm, maybe we can drop the pre-computed values? And always compute the sizes on startup even if no custom font is set?

Done. In reality, if the configuration file contains _any_ `font ...` command, then `config_update_font_height` is called by it already. It's only necessary after the font-reload because the _default_ value doesn't call the `font` command.

An alternative (cleaner, but more invasive) approach is to set `font = NULL` when loading the default values, and after parsing the whole file, if the font is still `NULL` run the comand `font monospace 10`. This seems cleaner because it means that the default font follows the same codepath as explicitly defined fonts. Thoughts?

RPigott:
`monospace` is a font alias. It's set in the various fontconfig files somehow, but I'm not totally sure how it works. You can use `fc-pattern -c monospace` to show the result of the configured substitutions on the pattern.

Also, don't quote me on this, but I _think_ the default font size comes from a dpi calculation, like so:

```shell
$ fc-pattern -c monospace -f '%{[]family{%{family}\n}}' \
    | parallel fc-match {} -f '"scale=2; \"%{family}\t\"; %{pixelsize} * 96 / %{dpi}\n"' \
    | awk '!font[$0]++' | bc | sed 's/,[^:]*\t/\t/' | column -ts $'\t' -o ': ' -R1
DejaVu Sans Mono: 16.00
     Inconsolata: 16.00
     Courier New: 16.00
       Luxi Mono: 16.00
  Nimbus Mono PS: 16.00
         Courier: 16.64
     DejaVu Sans: 16.00
       MS Gothic: 16.00
         NSimSun: 16.00
    DejaVu Serif: 16.00
  Noto Sans Mono: 16.00
```

I'm not sure how this works either. For example, I can influence the font sizes on my system by adding to the config

```shell
$ cat ~/.config/fontconfig/conf.d/50-dpi.conf
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
    <description>Set a wacky dpi</description>

    <match target="pattern">
        <edit name="dpi" mode="assign">
            <double>192</double>
        </edit>
    </match>
</fontconfig>
```

Everything stays the same at 16.0 except for Courier becomes 12.5. For the 16.0 size ones, the `pixelsize` property seems to always be reported as "dpi / 6" using the dpi value I set in the config, and setting dpi = 96 makes Courier match the others at 16. If I use the actual dpi of my screen, Courier is 16.9.

WhyNotHugo:
Sounds like dropping these hardcoded default is an improvement then: currently when tweaking sizes via fontconfig the default height becomes incorrect.

emersion:
> This seems cleaner because it means that the default font follows the same codepath as explicitly defined fonts. Thoughts?

I'm fine with either. It's just a single `config_update_font_height` call, so I don't feel strongly about it.

emersion (approved):
In any case, this LGTM! Thanks for the follow-up.

PR #6528, opened 2021-09-18
layer-shell: chase wlr layer surface refactor

vyivel:
https://github.com/swaywm/wlroots/pull/3201

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -598,14 +598,14 @@ void handle_layer_shell_surface(struct wl_listener *listener, void *data) {
 	sway_log(SWAY_DEBUG, "new layer surface: namespace %s layer %d anchor %" PRIu32
 			" size %" PRIu32 "x%" PRIu32 " margin %" PRIu32 ",%" PRIu32 ",%" PRIu32 ",%" PRIu32 ",",
 		layer_surface->namespace,
-		layer_surface->client_pending.layer,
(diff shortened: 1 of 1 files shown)


sway/desktop/layer_shell.c
```diff
@@ -659,13 +659,13 @@ void handle_layer_shell_surface(struct wl_listener *listener, void *data) {
...
-	wl_list_insert(&output->layers[layer_surface->client_pending.layer],
+	wl_list_insert(&output->layers[layer_surface->pending.layer],
 			&sway_layer->link);
 
-	// Temporarily set the layer's current state to client_pending
+	// Temporarily set the layer's current state to pending
 	// So that we can easily arrange it
 	struct wlr_layer_surface_v1_state old_state = layer_surface->current;
-	layer_surface->current = layer_surface->client_pending;
+	layer_surface->current = layer_surface->pending;
```
vyivel, line 668: Unrelated, but that's a hack and will stop working once https://github.com/swaywm/wlroots/pull/3151 is merged.

emersion: Ugh, yeah, that isn't great. We should definitely fix it, maybe by passing a "configuring" layer surface to `arrange_layers`. Also `arrange_layers` shouldn't actually update the positions of the other layers just yet… We should wait for `ack_configure`.

Oh well.

emersion (approved):
Thanks!

Issue #6501, opened 2021-09-09
Overlay surfaces don't get pointer events when there is a "fullscreen global" app.


tmccombs:
- **Sway Version:**
  - `sway version 1.6.1`

- **Debug Log:**
  - https://gist.github.com/tmccombs/7bf8c398c57b2f7d22047ac22f5d6dcf

- **Configuration File:**
  - Default configuration

- **Description:**
  - Open a terminal (or any other app)
  - Make the app "fullscreen global", for example by running `swaymsg fullscreen global`
  - Run `slurp` or another program that uses the `overlay` layer. 
  - The overlay surface will render above the fullscreen surface, and will capture keyboard presses, but won't capture pointer events.

See https://github.com/emersion/slurp/issues/92

PR #6545, opened 2021-09-23
Fix cursor input for layer-shell surfaces when a fullscreen global container is present

neonsea:
Previously, fullscreen global containers would grab cursor input, even if a shell-layer surface was on top of it. This would lead to issues like slurp not being able to select on top of full screen windows, see https://github.com/emersion/slurp/issues/92.

Related issue in this repo: https://github.com/swaywm/sway/issues/6501

This patch simply moves layer surface detection logic above fullscreen_global container detection logic. I've confirmed this fixes the issue, although I'm not sure if it introduces any other issues/bugs/unintended behaviour.


neonsea:
I've been using this patch for a bit now and I haven't detected any abnormalities with how fullscreen_global windows function.

vyivel:
This PR would fix that particular issue, but the main problem here is `node_at_coords()` not being in sync with rendering order. A "proper" way of doing things would be documenting the order in `render.c` and make `node_at_coords()` follow it in reverse.

With this PR, overlay layer is rendered on top of unmanaged surfaces, but `node_at_coords()` incorrectly checks it *after* unmanaged surfaces.

Issue #6523, opened 2021-09-15
Scaled-down GTK window leaves some of the screen stale


nwg-piotr:
### Please fill out the following:
- **Sway Version:**
  - `sway version 1.6-8d0a84f3 (Sep 15 2021, branch 'master')`

- **Debug Log:**
  - https://gist.github.com/nwg-piotr/6acf04fc325a4d336fe3cc540c07c7be

- **Description:**
  - When the window dimension decreases, the screen below the disappearing part won't get cleared. Instead, a part of the old window starts blinking. Focusing another container clears the effect. See the video: https://youtu.be/0F3vv_KoWAM

RPigott:
Is this a layer shell client?

nwg-piotr:
Yes. I should have mentioned, sorry.

RPigott:
Can you provide a WAYLAND_DEBUG log?

nwg-piotr:
> Can you provide a WAYLAND_DEBUG log?

As soon as I'm back from work.

nwg-piotr:
> Can you provide a WAYLAND_DEBUG log?

Available here: https://gist.github.com/nwg-piotr/41b78af5d4bf3a89c4f9e154845ee40a

Dunno if it's going to help, but remains of the window stay hidden below the focused window:

![obraz](https://user-images.githubusercontent.com/20579136/133692869-0a4a030a-6e35-428b-9b8a-27a2ba092c2d.png)

ErikReider:
I've seen similar "glitches" when showing multiple GTK windows in the same layer-shell layer

RPigott:
Can you provide a client and repro steps?

ErikReider:
Sadly it's been difficult to recreate... I'll try some more :)

PR #6551, opened 2021-09-24
layer-shell: fix commit handler

vyivel:
This commit makes sure the extents are kept up-to-date, fixes not
damaging the surface if its layer shell-specific state didn't change,
and adds a check if the layer shell-specific state didn't change but the
surface got mapped/unmapped, which could affect keyboard focus.

---

Fixes https://github.com/swaywm/sway/issues/6552

### DIFF
--- a/include/sway/layers.h
+++ b/include/sway/layers.h
@@ -22,6 +22,7 @@ struct sway_layer_surface {
 	struct wl_listener new_subsurface;
 
 	struct wlr_box geo;
+	bool mapped;
 	struct wlr_box extent;
 	enum zwlr_layer_shell_v1_layer layer;
 };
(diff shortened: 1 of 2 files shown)


RPigott (approved):
Kinda curious if moving the extent update to the commit handler has any effect on #6523.

Either way this change LGTM.

vyivel:
A user has reported that rofi doesn't get input focus immediately on start with this PR, so that needs to be looked into first before merging this.

vyivel:
This should work now; tested with `swaybg`, `swaybar`, `gtk-layer-demo` and `bemenu`.

RPigott (approved):
Oh, right. On the commit where the client becomes mapped the client may not have committed any actual layer-shell state. Makes sense and LGTM... again.

PR #6566, opened 2021-10-01
commands: Remove unused code after 1d3681f52135

ekhramtsov:
Clang 13 reports:

../sway/commands.c:470:23: error: variable 'context' set but not used
[-Werror,-Wunused-but-set-variable]
        enum command_context context = 0;
                             ^
Last use of was removed in commit 1d3681f52135.

Downstream PR: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=258813

--
I removed lines 470-495 and related because I couldn't find code that would pass "ipc"
and "binding" so they end up in argv[i]. This might be wrong.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -418,14 +418,6 @@ enum sway_popup_during_fullscreen {
 	POPUP_LEAVE,
 };
 
-enum command_context {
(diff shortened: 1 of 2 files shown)


emersion (approved):
Thanks!

PR #6575, opened 2021-10-03
idle_inhibit: Fix tree view after changes in commit 152a559e

bartelsielski:
Commit 152a559e replaced the view pointer in the inhibitor struct with a
pointer to the wlr_inhibitor for application inhibitors. But this was not
changed in the sway_idle_inhibit_v1_application_inhibitor_for_view function.
This caused a bug in the sway tree view where the application inhibitor is
always "none".


emersion:
Can you rebase against latest master? This should fix CI?

kennylevinsen (approved):
Needs a rebase for CI, but otherwise LGTM.

PR #6579, opened 2021-10-04
remove noop backend and noop output

bl4ckb0ne:
Fix for https://github.com/swaywm/wlroots/pull/3227

### DIFF
--- a/include/sway/tree/root.h
+++ b/include/sway/tree/root.h
@@ -31,7 +31,7 @@ struct sway_root {
 	list_t *scratchpad; // struct sway_container
 
 	// For when there's no connected outputs
-	struct sway_output *noop_output;
+	struct sway_output *fallback_output;
 
 	struct sway_container *fullscreen_global;
 
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -33,7 +33,6 @@ struct sway_server {
 	const char *socket;
 
 	struct wlr_backend *backend;
-	struct wlr_backend *noop_backend;
 	// secondary headless backend used for creating virtual outputs on-the-fly
 	struct wlr_backend *headless_backend;
 	struct wlr_renderer *renderer;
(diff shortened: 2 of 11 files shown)


emersion:
Well we still need to keep the noop output, but we can replace it with a headless output.

bl4ckb0ne:
Pushed a v2 of the patch. I put 800x600 as the headless output size randomly, let me know if you'd like a different size.

include/sway/tree/root.h
```diff
@@ -31,7 +31,7 @@ struct sway_root {
 	list_t *scratchpad; // struct sway_container
 
 	// For when there's no connected outputs
-	struct sway_output *noop_output;
+	struct sway_output *headless_output;
```
emersion, line 34: Hm. Sway now supports two types of headless outputs: this one used as a fallback in case there's no other output plugged in, and the ones created via `swaymsg create_output`, which behave just like regular outputs and can be used with e.g. VNC.

Maybe we can rename this one to `fallback_output` to avoid the confusion?

bl4ckb0ne: Done!

emersion:
Hm for some reason the headless output now shows up in `swaymsg -t get_outputs`. This was not the case for the noop one. It also shows up in `wayland-info` btw.

Also need to remove `#include <wlr/backend/noop.h>` from `sway/server.c`.

bl4ckb0ne:
Is it something we don't want?

emersion:
Yeah, we don't want this internal implementation detail to be exposed to our users. Users will be confused if they see this output that isn't supposed to exist.

emersion:
I think this happens because the `sway_output` is enabled for some reason? It shouldn't get enabled.

emersion:
Hm maybe this is an issue with wlroots, where `wlr_output.enabled == true` for headless outputs but not for noop. But I'm confused, I see noop outputs are marked as enabled in wlroots with `wlr_output_update_enabled`…

emersion:
Code LGTM! Can you squash these commits together? Starting Sway with just the first one applied will segfault.

bl4ckb0ne:
Done! I also added a bit of explaination in the commit message.

emersion (approved):
Excellent, thanks!

PR #6588, opened 2021-10-08
swaybar: tidy swaybar_output struct

nmschulte:
split from #6518; tidies `swaybar_output` to make the code a bit clearer

### DIFF
--- a/include/swaybar/bar.h
+++ b/include/swaybar/bar.h
@@ -69,14 +69,13 @@ struct swaybar_output {
 	bool focused;
 
 	uint32_t width, height;
-	int32_t scale;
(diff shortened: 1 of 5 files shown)


emersion:
I'm not a fan of adding `output_` fields to a struct already named "output".

nmschulte:
> I'm not a fan of adding `output_` fields to a struct already named "output".

Alright.  There's not much to be had in this request then?

> No, we shouldn't use the output mode here.
> We shouldn't store the output position here.
Okay.  I am unfamiliar and made these changes blindly as they seemed appropriate.  I would like to understand better but I guess that means you'd have to clarify; not a big deal as I'm sure it's involved.

Issue #6573, opened 2021-10-03
Move temporarily floated container back to original tiled position


siikamiika:
How to reproduce
1. Create a layout such as `H[1 2 3 V[4 5]]` or `T[1 2 3 4 5]`
1. `floating toggle` two times while focusing container `5`

Expected results (i3 behavior):
The layout stays the same

Actual results (in sway):
The layout becomes `V[H[1 2 3 V[4]] 5]` or `H[T[1 2 3 4] 5]`

This can also be reproduced by creating a new container belonging to the tiled layout when the floating container is focused.

----

Related issue: `focus mode_toggle` when focusing a floated container selects the entire tiled tree in sway instead of the container hinted by `client.focused_inactive` colors. If there are multiple such `focused_inactive` containers visible, i3 also focuses the previously active one and not the parent of everything like sway.

RPigott:
Wasn't this resolved in #6175? You should really report your sway version in issues.

siikamiika:
Sorry, I couldn't find anything related from issues when searching so I assumed this wasn't reported or noticed before. I'm running Arch build of sway 1.6.1 which seems to have been released after #6175 was merged, so I'm not sure if there was a regression or if that fix wasn't included in the release.

siikamiika:
https://github.com/swaywm/sway/commit/a558866f42de74f7dc0500c3fe29bb506cf03cf6 seems to be only on master, so I'll test on it and close this issue. If there are any remaining issues, they're probably nothing similar to the original report.

siikamiika:
Finally tested this and it seems to work great on master. Thanks for your work @RPigott

siikamiika:
Reopening to track a few small fixes

siikamiika:
Is there anything required from me to help with the review of the patches? So far I haven't noticed any issues using them.

PR #6594, opened 2021-10-09
container: restore next to sibling after floating

siikamiika:
I was testing the issue I reported in https://github.com/swaywm/sway/issues/6573 a bit further, and noticed a small difference in the behavior compared to i3. This PR attempts to fix it.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -821,8 +821,6 @@ void container_set_floating(struct sway_container *container, bool enable) {
 
 	struct sway_seat *seat = input_manager_current_seat();
 	struct sway_workspace *workspace = container->pending.workspace;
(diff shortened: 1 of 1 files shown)


siikamiika:
Sure, here are a few:
- `T[1 2 3 4]`: focus 3, focus 2, floating toggle twice: on master 2 becomes a sibling of 4, on i3 and this it goes between 3 and 4
- `H[1 V[2 3 4 5]]`: focus 3, focus 2, floating toggle twice: on master 2 becomes a sibling of 5, on i3 and this it goes between 3 and 4
- But interestingly, on a workspace that has never had its layout changed so it defaults to `H`: `H[1 2 3 4]`: focus 3, focus 2, toggle floating twice: on all 2 goes between 3 and 4

PR #6597, opened 2021-10-10
Fix a couple of typos

b1rger:


emersion:
Is _allows to_ incorrect wording?

ammgws:
> Is _allows to_ incorrect wording?

It feels strange to me.

PR #6598, opened 2021-10-10
focus inactive view on `focus` mode changes

siikamiika:
Related: #6573


siikamiika:
Thanks, I have now rewritten the commit message.

PR #6611, opened 2021-10-16
swaymsg: use INT_MAX max JSON depth when parsing IPC response

ammgws:
With the latest sway release (possibly also for sway-git and previous releases), `swaymsg -t get_tree` sometimes fails and returns "ERROR: Could not parse json response from ipc. This is a bug in sway."

In my case it turned out to be "failed to parse payload as json: nesting too deep", which is something that has come up before in other parts of the codebase (https://github.com/swaywm/sway/pull/6125)

This PR adopts the same solution as #6125, and also returns a better error message in case another issue comes up in the future.


emersion:
> Should we also use INT_MAX here, like with #6125?

Yeah, I think so.

PR #6616, opened 2021-10-17
Ignore modifiers on bindsym release, to match i3 behaviour

### LINKED ISSUE #6456: `keysym --release` is unreliable
- **Sway Version:** 1.6.1

- **Description:**
  - Press Alt_R key
  - Release Alt_R

k3d3:
Fixes #6456.

The i3 window manager ignores modifiers when a button is being released. Mimic this behaviour in Sway.

---

As this is my first commit to Sway, let me know if there's anything else you need!

Thanks


emersion:
I don't think this is how i3 behaves. Can someone check?

k3d3:
Changes made.

emersion:
The i3 logic is a bit more complicated: it only ignores modifiers if `release == B_UPON_KEYRELEASE_IGNORE_MODS`, which is set when pressing the keys.

PR #6629, opened 2021-10-21
Bump RLIMIT_NOFILE

emersion:
Wayland compositors handle many file descriptors: client
connections, DMA-BUFs, sync_files, wl_data_device pipes, and so
on. Bump the limit to the max.

Closes: https://github.com/swaywm/sway/issues/6285

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -137,6 +137,8 @@ void server_fini(struct sway_server *server);
 bool server_start(struct sway_server *server);
 void server_run(struct sway_server *server);
 
+void restore_nofile_limit(void);
+
 void handle_compositor_new_surface(struct wl_listener *listener, void *data);
 void handle_new_output(struct wl_listener *listener, void *data);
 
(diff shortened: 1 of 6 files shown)


WhyNotHugo (approved):
I'm running sway with this patch, and can no longer reproduce #6642.

Without this patch and the with my usual amount of windows (just restoring my Firefox session), this reproduces 100% of the times.

With this patch and more-than-usual amount of windows, the issue is no longer reproduced (I did 4 attempts).

bl4ckb0ne (approved):
LGTM, thanks :)

WhyNotHugo:
Thanks!

RPigott:
We should probably also log the nofile limit when setrlimit succeeds. The user might still have a low hard limit.

WhyNotHugo:
Showing an error when the limit is reached might be a good idea. Someone mentioned it's handled in a dependency/library, but I can't find the reference to that.

PR #6638, opened 2021-10-26
desktop: output: Set output damage on direct scanout

any1:


emersion:
Similar patch for wlroots' scene-graph: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3883

emersion:
We've switched to scene-graph now.

PR #6639, opened 2021-10-27
Implement startup notifications for workspace matching

RPigott:
Depends on swaywm/wlroots#3305.

This causes sway to export xdga tokens when it execs and use them for workspace matching. If a client doesn't request activation with a related token by the time it is mapped, we fall back to pid mapping just like before.

~~X startup notifications are not yet implemented, and at the moment sway only tracks the startup of applications it has exec'd. I plan to add support for these soon, or later in another pr.~~

~~Also, i3 has a `--no-startup-id` flag to disable this feature (or its equivalent in X). Sway currently parses and ignores that flag, but it should probably be implemented too.~~

EDIT: support for X startup notifications and `--no-startup-id` is in, support for external launchers is out for now.


RPigott:
Marked as WIP so I can split this up. The changes aren't too substantial, it's just probably cleaner to have the refactoring in the middle in it's own PR.

jkonecny12:
I'm not that used of code of Sway, so, if possible I would like to have an usage example. Could you please write some here?

RPigott:
@jkonecny12 did you see the foot patch?

emersion:
@RPigott, do you have plans to update this PR?

emersion:
Ah, it seems you've rebased it. Is it ready for review?

RPigott:
It should be ready for review. I had only planned to rearrange the commits/changes but I don't remember how anymore anyway.

sway/desktop/launcher.c
```diff
@@ -174,26 +184,21 @@ void root_record_workspace_pid(pid_t pid) {
...
-	struct pid_workspace *old, *_old;
-	wl_list_for_each_safe(old, _old, &pid_workspaces, link) {
-		if (now.tv_sec - old->time_added.tv_sec >= timeout) {
-			pid_workspace_destroy(old);
-		}
-	}
-
 	struct pid_workspace *pw = calloc(1, sizeof(struct pid_workspace));
+	struct wlr_xdg_activation_token_v1 *token =
+		wlr_xdg_activation_token_v1_create(server.xdg_activation_v1);
```
emersion, line 174: It seems like `server.xdg_activation_v1` is used here, but doesn't exist yet at this point in the Git history?

sway/desktop/launcher.c
```diff
@@ -151,12 +187,13 @@ void root_record_workspace_pid(pid_t pid) {
 	}
 
 	struct pid_workspace *pw = calloc(1, sizeof(struct pid_workspace));
-	pw->workspace = strdup(ws->name);
```
emersion: Hm. I think we record the workspace here in case the user goes to workspace 4 (no windows there, so it's created), then launches a client, then goes to another workspace (implicitly destroys workspace 4 since it's empty). Once the client starts up, it's desirable to re-create workspace 4 and place the new window there.

sway/desktop/launcher.c
emersion: I'm mistaken, we already have this in master for urgency.

sway/desktop/launcher.c
RPigott: I've only renamed this variable. It's now recorded as `pw->name` just below instead of `pw->workspace`, since that's what it contains. That scenario still works as intended.

Actually it looks like this strdup line really is unnecessary though, since I do the same in L141 in this commit, and don't even remember to free the old name str.

RPigott:
Rebased to resolve a conflict, and added two additional commits which fix issues.

RPigott:
Rebased to use my preferred email, and finally merge those fixup commits — they didn't apply cleanly in the previous history.

sway/desktop/launcher.c
```diff
@@ -79,45 +82,83 @@ struct sway_workspace *root_workspace_for_pid(pid_t pid) {
...
+	case N_CONTAINER:
+		// Unimplemented
+		break;
+	case N_WORKSPACE:;
+		struct sway_workspace *ws = pw->node->sway_workspace;
+		wl_list_remove(&pw->node_destroy.link);
+		wl_list_init(&pw->node_destroy.link);
+		// We want to save this ws name to recreate later, hopefully on the
+		// same output
+		pw->name = strdup(ws->name);
```
emersion: We've already done this in `root_record_workspace_pid()` I think?

sway/desktop/launcher.c
```diff
@@ -79,45 +82,83 @@ struct sway_workspace *root_workspace_for_pid(pid_t pid) {
...
+static void pw_handle_node_destroy(struct wl_listener *listener, void *data) {
+	struct pid_workspace *pw = wl_container_of(listener, pw, node_destroy);
+	switch (pw->node->type) {
+	case N_CONTAINER:
+		// Unimplemented
+		break;
+	case N_WORKSPACE:;
+		struct sway_workspace *ws = pw->node->sway_workspace;
+		wl_list_remove(&pw->node_destroy.link);
+		wl_list_init(&pw->node_destroy.link);
```
emersion: Can we move the `wl_list_remove()` + `wl_list_init()` out of the `switch`? Would make the logic a bit simpler.

include/sway/desktop/launcher.h
```diff
@@ -3,10 +3,30 @@
...
 
-struct sway_workspace *workspace_for_pid(pid_t pid);
+struct launcher_ctx {
+	pid_t pid;
+	char *name;
+	struct wlr_xdg_activation_token_v1 *token;
+	struct wl_listener token_destroy;
 
-void launcher_ctx_create(pid_t pid);
+	bool activated;
```
emersion, line 12: This is unused.

include/sway/desktop/launcher.h
```diff
@@ -3,10 +3,30 @@
...
+struct launcher_ctx *launcher_ctx_find_pid(pid_t pid);
+
+struct sway_workspace *launcher_ctx_get_workspace(struct launcher_ctx *ctx);
 
-void remove_workspace_pid(pid_t pid);
+void launcher_ctx_consume(struct launcher_ctx *ctx);
+
+void launcher_ctx_destroy(struct launcher_ctx *ctx);
+
+const char *launcher_ctx_get_name(struct launcher_ctx *ctx);
```
emersion, line 28: This is unused as well.

emersion:
Commits up to (and including) "[launcher: rename pid_workspace to launcher_ctx](https://github.com/swaywm/sway/pull/6639/commits/3add3febd0728f6cafda1b8ee22b070f42d1bbd9)" look good to me, with the comments addressed.

include/sway/desktop/launcher.h
RPigott: For my reference, this is first assigned to in e8deba2b (launcher: export xdga tokens [...])

TBH I can't 100% remember why I introduced this other than I think I was having trouble with some clients (firefox) that would map two surfaces at the same time with the same token. My intention is to make the token inert after I match something with it. I added this bool after my first approach didn't work in a later commit, and apparently the rebase isn't as simple as I had hoped if I want the history to be correct.

include/sway/desktop/launcher.h
RPigott: For reference, this is first used in e8deba2b (launcher: export xdga tokens [...])

sway/desktop/launcher.c
RPigott: Hm this is the same error as in the last review, but I didn't go far enough back when applying the fix. What's missing is a free-ing the old name, we still want the strdup I think.

What I'm doing here is updating pw->name in case it has changed since the process was launched. In order to preserve sway's existing behavior, I'm trying to accommodate the following scenario that currently works in sway:

1. User has workspace "A"
2. User launches process 123
3. User renames workspace "A" -> "B"
4. User switches to another workspace and an empty workspace "B" is reaped
5. Process 123 maps a window, it should be pid matched to workspace "B"
6. Workspace "B" is created

Previously all pid matches were handled by recording the name and looking it up when the window was mapped. When a workspace was renamed a helper function walked the list of pid matches and updated the tracked workspace names too. Now in b2133d968 (launcher: track workspaces by node) I dropped that and matches are handled directly by node, so it doesn't matter if the workspace name has changed ... unless it was changed before it was destoyed and we need to recreate a workspace with the updated name. So I'm recording it here for that purpose — this variable is now "name of the workspace to create in case the one we intended to match has been destroyed" .

sway/desktop/launcher.c
RPigott: Yeah seems good.

RPigott (commented):
Thanks, I'll try to fix up these bits and clean up the history.

sway/desktop/launcher.c
RPigott: Wasn't sure a neat way to write this since I want to re-init the listener before I add it to another node_destroy signal, like at the end of this N_WORKSPACE branch.

RPigott:
Ok I deferred the introduction of those attrs and functions until they are actually useful, and added the intended `free(pw->name)` to the earlier commits before it is renamed to `ctx->name`.

sway/desktop/xwayland.c
```diff
@@ -666,6 +669,27 @@ static void handle_set_role(struct wl_listener *listener, void *data) {
...
+
+	struct wlr_xdg_activation_token_v1 *token =
+		wlr_xdg_activation_v1_find_token(
+				server.xdg_activation_v1, xsurface->startup_id);
+	if (token == NULL) {
+		// We don't support external launchers in X
+		return;
+	}
+
+	struct launcher_ctx *ctx = token->data;
```
emersion, line 689: This may be NULL: clients can create activation tokens as well.

include/sway/desktop/launcher.h
```diff
@@ -0,0 +1,32 @@
...
+
+struct launcher_ctx *launcher_ctx_find_pid(pid_t pid);
+
+struct sway_workspace *launcher_ctx_get_workspace(struct launcher_ctx *ctx);
+
+void launcher_ctx_consume(struct launcher_ctx *ctx);
+
+void launcher_ctx_destroy(struct launcher_ctx *ctx);
+
+struct launcher_ctx *launcher_ctx_create();
```
emersion, line 28: We can use `(void)` to make it an error to call this function with arguments.

sway/desktop/xwayland.c
RPigott: view_assign_ctx treats NULL as just a request to cleanup the currently existing context. I can see how that's not desirable though because it interferes with pid matching and prior activation requests, forcing the view to be placed on the active workspace when mapped.

I'll change this to return early. In a future PR I think I will add contexts to client tokens as well to support external launchers.

include/sway/desktop/launcher.h
RPigott: Ok.

sway/desktop/launcher.c
```diff
@@ -59,7 +47,19 @@ static pid_t get_parent_pid(pid_t child) {
 	return -1;
 }
 
-static void launcher_ctx_destroy(struct launcher_ctx *ctx) {
+void launcher_ctx_consume(struct launcher_ctx *ctx) {
+	// The view is now responsible for destroying this ctx
+	wl_list_remove(&ctx->token_destroy.link);
+	wl_list_init(&ctx->token_destroy.link);
+
+	ctx->token = NULL;
```
emersion, line 57: Shouldn't we destroy the token here?

include/sway/desktop/launcher.h
```diff
@@ -0,0 +1,32 @@
...
+	char *name;
+	struct wlr_xdg_activation_token_v1 *token;
+	struct wl_listener token_destroy;
+
+	bool activated;
+
+	struct sway_node *node;
+	struct wl_listener node_destroy;
+
+	struct wl_list link;
```
emersion, line 17: Please add a comment to document which list head is used here.

include/sway/server.h
```diff
@@ -116,6 +116,8 @@ struct sway_server {
 	struct wlr_xdg_activation_v1 *xdg_activation_v1;
 	struct wl_listener xdg_activation_v1_request_activate;
 
+	struct wl_list launcher_ctxs; // launcher_ctx::link
```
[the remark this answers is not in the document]
emersion: Maybe `pending_launcher_ctxs` or `pid_launcher_ctxs`?

sway/commands/exec_always.c
```diff
@@ -90,8 +108,13 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 	waitpid(pid, NULL, 0);
 	if (child > 0) {
 		sway_log(SWAY_DEBUG, "Child process created with pid %d", child);
-		root_record_workspace_pid(child);
+		if (ctx != NULL) {
+			sway_log(SWAY_DEBUG, "Recording workspace for process %d", child);
+			ctx->pid = child;
+			wl_list_insert(&server.launcher_ctxs, &ctx->link);
```
emersion, line 114: Can we move this to `launcher_ctx_create()`? It doesn't seem like there's a case where `launcher_ctx_create()` is called without inserting in the list?

sway/desktop/launcher.c
```diff
@@ -0,0 +1,210 @@
...
+	wl_signal_add(&ctx->node->events.destroy, &ctx->node_destroy);
+
+	ctx->token_destroy.notify = token_handle_destroy;
+	wl_signal_add(&token->events.destroy, &ctx->token_destroy);
+
+	wl_list_init(&ctx->link);
+	return ctx;
+}
+
+const char *launcher_ctx_get_name(struct launcher_ctx *ctx) {
```
emersion, line 207: Nit: maybe `get_token_name` would be slightly better.

emersion (commented):
Cool, I think this is good to go after these last comments are addressed.

sway/desktop/launcher.c
RPigott: Hmm yes I think you're right.

sway/desktop/launcher.c
RPigott: done.

sway/commands/exec_always.c
RPigott: Yep, seems good.

include/sway/server.h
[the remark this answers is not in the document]
RPigott: `pending_launcher_ctxs` sounds good to me.

sway/desktop/launcher.c
RPigott: Ah hm I think you're right.

emersion (approved):
Thanks!

PR #6648, opened 2021-10-31
Add always_hide_tab command

shtayerc:
Titles on tabbed and stacked windows can be disabled, I was trying to get similar behavior to dwm monocle mode.

Command name can be changed if there is better idea.
I was also thinking about setting the flag when **default_border** is **none**, but that would change current behavior so I used new command.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -104,6 +104,7 @@ sway_cmd cmd_exec_validate;
 sway_cmd cmd_exec_process;
 
 sway_cmd cmd_assign;
+sway_cmd cmd_always_hide_tab;
 sway_cmd cmd_bar;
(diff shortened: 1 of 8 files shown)


emersion:
Sorry, we're not accepting any new window management features anymore. See https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#scope-of-future-changes-to-sway

shtayerc:
@emersion This is not window management, it is just hiding title bars. Similar to what **default_border** is doing. The same feature already exists with hide_lone_tab flag, except it only works for single window.

emersion:
The point is, this isn't in i3, so we (in general) don't accept it for Sway.

PR #6653, opened 2021-11-08
grimshot: Check if an instance is already running

SuperSamus:
Right now, it is possible to execute multiple instances of Grimshot, meaning you can have multiple overlays of Slurp or make a screenshot of its overlay. This is probably not a user's desired behavior.
Slurp itself doesn't intend to have this feature https://github.com/emersion/slurp/issues/96, but since this is a helper script I guess it's appropriate here.
Not sure if `pgrep` is a command that can be taken for granted like `echo` is. (I mean, I don't know if it should be added to the dependencies)

### DIFF
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -109,6 +109,9 @@ if [ "$ACTION" = "check" ] ; then
   check jq
   check notify-send
   exit
+elif [ "$(pgrep -u $USER grimshot)" != $$ ]; then
+  echo "An instance is already running"
+  exit 1
 elif [ "$SUBJECT" = "area" ] ; then
   GEOM=$(slurp -d)
   # Check if user exited slurp without selecting the area


emersion:
The user might have another instance on another compositor or another seat.

SuperSamus:
Do you know how I can check for that?

siikamiika:
> Do you know how I can check for that?

I guess it would be possible to use a pidfile that also stores the current `WAYLAND_DISPLAY` or `DISPLAY`

SuperSamus:
I tried checking `WAYLAND_DISPLAY` with the condition:
`elif [ $(pgrep grimshot | while read -r line; do grep -oPa "WAYLAND_DISPLAY=${WAYLAND_DISPLAY}" "/proc/${line}/environ"; done | grep "${WAYLAND_DISPLAY}" -c) = "2" ] ; then`
But for some reason it doesn't behave like I expect to. (It is false even with two instances)
Sorry if I'm new to Bash.

siikamiika:
The `elif` condition is quite stuffed so it can get hard to maintain, but it seems that your problem is counting multiple matches on the same line with `grep -c`

SuperSamus:
If I just run `pgrep grimshot | while read -r line; do grep -oPa "WAYLAND_DISPLAY=${WAYLAND_DISPLAY}" "/proc/${line}/environ"; done | grep "${WAYLAND_DISPLAY}" -c` however thr number is reported correctly.
Still, if you think it's not worth it feel free to close the PR.

siikamiika:
You're right, I made a mistake when testing it and included the `$(...)` too. I don't make the calls here and I'm just trying to help you get this forward. In general splitting a problem in smaller parts can make it easier to solve

siikamiika:
@SuperSamus you could try something like this
```bash
while read -r line; do
    if [ "$line" -eq "$$" ] ; then
        continue
    fi
    if grep -qFa "WAYLAND_DISPLAY=${WAYLAND_DISPLAY}" "/proc/${line}/environ" ; then
        echo "an instance is already running"
        exit 1
    fi
done <<<$(pgrep grimshot)
```

PR #6662, opened 2021-11-15
Improve man 5 sway

jkonecny12:
Recently I spent a few hours on making the `for_window` configuration option to work. The whole problem is that there are missing information in the man page. Let's add the missing info there.

However, question to someone knowing the code:
- should I add this also to `assign`?
- aren't these bugs?


jkonecny12:
Based on discussion on `#sway-devel` it seems that this is probably a bug. I'll try to investigate the behavior a bit more and create a bug report in case it's true.

emersion:
Marked as draft since more investigation is necessary.

PR #6665, opened 2021-11-15
sway: create wlr_renderer and wlr_allocator

bl4ckb0ne:
wlroots now required the compositor to create its own wlr_renderer and wlr_allocator to initialize the wlr_output

Matching wlroots PR: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3349

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -850,6 +850,12 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 		return;
 	}
 
+	if (!wlr_output_init_render(wlr_output, server->allocator,
+			server->renderer)) {
+		sway_log(SWAY_ERROR, "Failed to init output render");
+		return;
+	}
+
 	struct sway_output *output = output_create(wlr_output);
 	if (!output) {
 		return;
--- a/sway/server.c
+++ b/sway/server.c
@@ -73,12 +73,23 @@ static void handle_drm_lease_request(struct wl_listener *listener, void *data) {
 bool server_init(struct sway_server *server) {
 	sway_log(SWAY_DEBUG, "Initializing Wayland server");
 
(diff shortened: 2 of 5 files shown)


sway/desktop/output.c
```diff
@@ -850,6 +850,8 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 		return;
 	}
 
+	wlr_output_init_render(wlr_output, server->allocator, server->renderer);
```
emersion, line 853: Maybe we should bail out if this fails, to avoid hitting assertions down the line.

bl4ckb0ne: Sure

bl4ckb0ne: Let me know if the style is OK, not sure about the cut and number of tabs

sway/server.c
```diff
@@ -73,12 +73,17 @@ static void handle_drm_lease_request(struct wl_listener *listener, void *data) {
 bool server_init(struct sway_server *server) {
 	sway_log(SWAY_DEBUG, "Initializing Wayland server");
 
-	struct wlr_renderer *renderer = wlr_backend_get_renderer(server->backend);
-	assert(renderer);
+	server->renderer = wlr_renderer_autocreate(server->backend);
+	assert(server->renderer);
```
emersion, line 77: Before this wlroots breaking change, `wlr_backend_autocreate` was responsible for creating the renderer. That's no longer the case. That means `wlr_renderer_autocreate` will return NULL if the renderer couldn't be created.

Can we add proper error handling instead of assertions for `wlr_renderer_autocreate` and `wlr_allocator_autocreate`?

emersion (approved):
LGTM!

PR #6666, opened 2021-11-15
Clean up output destroy sequence

RPigott:
This is just the first 2 commits from #6639.

Previously an output had a destroy event that was emitted twice in a row (once in `output_begin_destroy` and again in the handler that is its only caller) and the output node destroy event was never signaled.

This renames the output destroy event to "disable" to better reflect it's usage and updates it to only emit once, and now the output node destroy event is emitted in `output_begin_destroy`.

There's no functional change to sway here, .

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -48,7 +48,7 @@ struct sway_output {
 	struct wl_listener damage_frame;
 
 	struct {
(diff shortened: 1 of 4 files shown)


emersion (approved):
LGTM, thanks!

PR #6672, opened 2021-11-18
Do not default to `off` for dpms command

ammgws:
Currently any argument passed to the dpms command that isn't in [`parse_boolean`](https://github.com/swaywm/sway/blob/cbecc5cbaed6b30c995d2c245def458e383b4e38/common/util.c#L41) will cause the command to default to `off`. While this behaviour/usage of parse_boolean might make sense for config options in order to [match i3](https://github.com/swaywm/sway/pull/2340/files#r204566486)'s lenient behaviour, it seems a bit out of place for this command in sway.

### DIFF
--- a/sway/commands/output/dpms.c
+++ b/sway/commands/output/dpms.c
@@ -35,8 +35,11 @@ struct cmd_results *output_cmd_dpms(int argc, char **argv) {
 
 	if (parse_boolean(argv[0], current_dpms == DPMS_ON)) {
 		config->handler_context.output_config->dpms_state = DPMS_ON;
-	} else {
+	} else if (strcmp(argv[0], "off") == 0) {
 		config->handler_context.output_config->dpms_state = DPMS_OFF;
+	} else {
+		return cmd_results_new(CMD_INVALID,
+				"Invalid dpms argument: %s", argv[0]);
 	}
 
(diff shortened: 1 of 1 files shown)


ammgws:
I was mulling over it for a while but couldn't accept the difference in outcomes: for example with `show_marks` which is documented to accept `yes|no`, it will silently fall back to "no" when the argument is invalid, which isn't as jarring as having your whole monitor blanked in the case of the dpms command.

This kind of subcommand argument checking is seen in other commands such as `transform`. Are we saying this should default to "normal" instead of throwing an error?

https://github.com/swaywm/sway/blob/cbecc5cbaed6b30c995d2c245def458e383b4e38/sway/commands/output/transform.c#L50

ammgws:
Alternative solutions instead of erroring out like `transform` could be:

* Simply no-op and return success

or

* Default to "on" instead of "off"

Either way seems more reasonable of a fallback than to blank the screen.

kennylevinsen:
That an erroneous dpms command turns the monitor off is only mildly surprising, as the user was explicitly playing with dpms.

parse_boolean is indeed unhelpful, but we probably can't fix that for i3 compat reasons. We do not want inconsistent per-config variations of a boolean config value.

PR #6677, opened 2021-11-21
swaybar: signal status command's process group

### LINKED ISSUE #5584: Sway reload leaves orphaned i3blocks instances
If I reload sway, recent i3block instance becomes orphaned and thus over time with frequent reloads there are too many of them. I checked i3 and it doesn't do that.

Sway: 1.5.2
i3blocks: 1.5.2
Kernel: x86_64 Linux 5.7.10_1
OS: Void Linux

LDVG:
Make the status command a process group leader and change the kill(2)
calls to target the new process group. Signals sent by swaybar will then
be received by both the status command and its children, if any. While
here, check the result of fork(2).

Without this, children spawned by the status command may not receive the
signals sent by swaybar. As a result, these children may be orphaned on
reload.

The issue could be shown by setting the bar to

    bar {
        status_command i3status | tee /tmp/i3status.out
    }

which would leave orphaned processes for each reload of sway

    $ ps o pid,ppid,cmd | grep i3status | grep -v grep
    43633   43624 sh -c i3status | tee /tmp/i3status.out
    43634   43633 i3status
    43635   43633 tee /tmp/i3status.out

    $ swaymsg reload

    $ ps o pid,ppid,cmd | grep i3status | grep -v grep
    43634       1 i3status
    43635       1 tee /tmp/i3status.out
    43801   43788 sh -c i3status | tee /tmp/i3status.out
    43802   43801 i3status
    43803   43801 tee /tmp/i3status.out

This fixes #5584.

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -172,7 +172,7 @@ bool determine_bar_visibility(struct swaybar *bar, bool moving_layer) {
 		if (bar->status) {
 			sway_log(SWAY_DEBUG, "Sending %s signal to status command",
 					visible ? "cont" : "stop");
-			kill(bar->status->pid, visible ?
+			kill(-bar->status->pid, visible ?
 					bar->status->cont_signal : bar->status->stop_signal);
 		}
 	}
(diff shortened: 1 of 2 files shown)


emersion (approved):
Thanks!

PR #6679, opened 2021-11-22
Treat fullscreen windows as 'tiled' for commands/focus

Synthetica9:
Currently, fullscreen containers are treated differently based on whether they were originally floating or tiled. This is strange from a user perspective, since there is no visual hint whether a container was originally floating or tiled. This change only affects multi-monitor setups.

To see the current effect, try the following:

1. Open a window in tiled mode
2. Fullscreen it
3. You can still move between monitors with `$mod-$left`/`$mod-$right`.
4. Open a window in floating mode
5. Fullscreen it
6. You can _not_ move between monitors with `$mod-$left`/`$mod-$right`
7. However, there is no visual difference between 3. and 6.

This PR makes the behaviour in 6. consistent with the behaviour in 3.

### DIFF
--- a/sway/commands/focus.c
+++ b/sway/commands/focus.c
@@ -451,7 +451,8 @@ struct cmd_results *cmd_focus(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "");
 	}
 	struct sway_node *next_focus = NULL;
-	if (container_is_floating(container)) {
+	if (container_is_floating(container) &&
+			container->pending.fullscreen_mode == FULLSCREEN_NONE) {
 		next_focus = node_get_in_direction_floating(container, seat, direction);
 	} else {
 		next_focus = node_get_in_direction_tiling(container, seat, direction, descend);


emersion:
How does i3 behave?

Synthetica9:
> How does i3 behave?

Not sure, don't have a multi-screen i3 setup to test on. If it _doesn't_ behave like this I'd argue that's a bug in i3, not in this PR.

markstos:
I installed i3 to test this. i3 behaves as this patch proposes to fix-- after fullscreening both a tiled and floating window, you can use Mod+arrow keys to navigate to windows on the other monitor. 

The proposed behavior change is both reasonable and aligned with i3. Merging is recommended.

emersion (approved):
Thanks for testing. I'm not super happy about calling a function called `_tiling` on a non-tiled container, but OTOH it's a function local to the file so not that of a big deal. (An alternative would be to add a condition to check for fullscreen views in the `_floating` variant.)

Looks good to me.

emersion:
Thanks for your contribution!

PR #6697, opened 2021-12-04
output: remove surface buffer damage check

vyivel:
A surface can have effective damage even without any buffer damage
committed.

---

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3389


emersion (approved):
LGTM, thanks!

PR #6698, opened 2021-12-04
github MD now supports bidi texts

mhdzli:


emersion:
Sweet. For some reason the first line seems to be right-aligned still?

https://github.com/swaywm/sway/blob/115ab1caafa166a5daa4b2f1b4c96fdf21aedbdb/README.ir.md

mhdzli:
It seems that the direction of every paragraph is based on the first word.

tiosgz:
> Sweet. For some reason the first line seems to be left-aligned still?

This issue is on `master` too (at least with my browser). GitHub makes everything `<p dir="auto">`, which bypasses that `<div>`.

emersion:
Maybe we can use a [right-to-left UTF-8 mark](https://en.wikipedia.org/wiki/Right-to-left_mark)? `&rlm;` should work I think.

mhdzli:
Yes that or as I added a few persian words to the first paragraph  which replaced "sway" with "sway windowmanager" in persian words "مدیر پنجره" means "window manager"

emersion:
There are other lines where the text direction is incorrect.

tiosgz:
> Yes that or as I added a few persian words to the first paragraph which replaced "sway" with "sway windowmanager" in persian words "مدیر پنجره" means "window manager"

I don’t know why, but with that workaround, `#sway sur irc.libera.chat` looks something like `sway sur# irc.libera.chat`.

emersion:
Maybe because the sentence is split on multiple lines?

Maybe it would help to use references for links just like the English version does, could shorten the text a bit.

mhdzli:
OK I think I covered all of them and there was a typo "sur" instead of "در"

emersion (approved):
Thanks

PR #5438, opened 2020-06-09
Handle border width and height on minimum floating sizes

RoastVeg:
Fixes #5337

PR #6700, opened 2021-12-04
Handle border width and height on minimum floating sizes

### LINKED ISSUE #5337: Take border into account for max floating window size
See https://github.com/swaywm/sway/pull/5299#discussion_r422660249

OctopusET:
This fixes #5337

Here's original author's PR. https://github.com/swaywm/sway/pull/5438
Seems like OP is not active for long time. So I fixed the build error, and squashed commits into one.

Thanks.


emersion (approved):
Thanks!

PR #6701, opened 2021-12-05
Implement `modifiers` array in swaybar click events

pogo5422:


sway/ipc-json.c
```diff
@@ -1000,6 +1001,20 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 					layout ? json_object_new_string(layout) : NULL);
 			}
 		}
+
+		json_object *modifiers_json = json_object_new_array();
+		uint32_t keyboard_modifiers = wlr_keyboard_get_modifiers(keyboard);
+		for (int i = 0; i < (int)(sizeof(modifiers) / sizeof(struct modifier_key)); i++) {
```
emersion, line 1007: Can we use `get_modifier_names` instead? That way we can also keep the `modifiers` global variable static in `sway/input/keyboard.c`.

swaybar/ipc.c
```diff
@@ -606,3 +606,23 @@ bool handle_ipc_readable(struct swaybar *bar) {
 	free_ipc_response(resp);
 	return bar_is_dirty;
 }
+
+json_object *ipc_get_keyboard_modifiers(struct status_line *status) {
+	uint32_t len = 0;
+	char *buf = ipc_single_command(status->bar->ipc_socketfd,
+			IPC_GET_INPUTS, NULL, &len);
```
emersion, line 613: We don't need to use the Sway IPC to get this information. Instead, we can just use Wayland. We'd need to add a `wl_keyboard` listener in `swaybar/input.c`.

emersion (changes requested):

swaybar/ipc.c
kraftwerk28: Will it work, since swaybar surface is not in focus? I tried attaching `wl_keyboard_listener` with a bunch of logging inside callbacks, but they aren't fired when I press keys.

kraftwerk28: Here's my current swaybar/input.c: https://gist.github.com/kraftwerk28/aedafb935d80638dc3f08e51a71a187c

https://user-images.githubusercontent.com/31807671/145615521-16854db0-6363-4d74-b152-23e988561b8a.mp4

leosunmo:
Any movement on this? It would be very nice to have modifiers in swaybar.

PR #6249, opened 2021-05-02
Tray D-Bus Menu

FlexW:
This pull request implements D-Bus menus for the tray icons. @ianyfan @nmschulte It seems like #5161 is dead, and I would like to take over. Please tell me if this is not the case, then I will close this pull request.

This pull request is based on #5161 and is a draft for now. Please don't test it yet. When the pull request is ready for testing, I will say so.

@ianyfan @nmschulte I created this early pull request to let you know I'm working on it. Just in case that we do not double the work. Maybe you could add some other points that should be addressed before this pull request can be merged. I would say keyboard navigation and touch support can be done in a second pr.

PR #6703, opened 2021-12-05
Delete .clang-format

emersion:
This file isn't accurate, and clang-format can't describe our code
style.

References: https://github.com/swaywm/sway/pull/6249#issuecomment-986214042


Xyene:
This is a little unfortunate, it was somewhere on my stack to look into making the .clang-format better represent Sway's codebase.

Since you seem to have looked into it already, what parts of Sway's code style are not describable in clang-format? It's unfortunate to have to manually deal with the minutiae of formatting; it's an extra step in development, and results in extra round-trips on otherwise fine PRs. Are the differences in capabilities major, or just a few points? Could we drop those formatting points and instead use something that's well-supported by an autoformatter? (Not necessarily clang-format, though I'm under the impression that's one of the more developed/maintained ones.) 

This should be weighed against the parts of the codebase that presently have inconsistent formatting, which would be consistent if we had an autoformatter description (hanging indents come to mind as something that are inconsistent throughout Sway and wlroots, which I've been fixing as I see them in code I'm modifying, but...).

emersion:
The code style is described in https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/CONTRIBUTING.md

The closest I've found is this, but it still doesn't do continuation lines properly and it doesn't retain existing continuation lines breaking points: https://paste.sr.ht/~emersion/6222bcb2a9f97d8e379672f5eedb6463085e2468

I think clang-format is just not the right tool for the job, because it completely ignores the input file formatting.

Xyene:
Just so we're on the same page...

> it still doesn't do continuation lines properly

Do you have an example in mind?

> it doesn't retain existing continuation lines breaking points

Are you referring to something like in this diff?

```diff
-void pango_printf(cairo_t *cairo, const char *font,
-               double scale, bool markup, const char *fmt, ...) {
+void pango_printf(cairo_t *cairo, const char *font, double scale, bool markup,
+               const char *fmt, ...) {
```

Would you expect this to remain unchanged after formatting?

emersion:
> Do you have an example in mind?

Hm I think the issue was

```c
if (this &&
        that) {
    do_thing();
}
```

vs.

```c
do_thing(this,
    that);
```

but I don't remember exactly.

> Would you expect this to remain unchanged after formatting?

Yeah. Sometimes clang-format did weird stuff and the manual breaks were much more natural.

In any case I wouldn't want to re-format three quarters of Sway's codebase just to make clang-format happy.

Xyene:
> Yeah.

Understood. Personally, I disagree -- I think there are cases where the manual argument formatting makes sense, where a function has many parameters and each line of parameters contains a "logical grouping" of parameters, like in the fictitious example below.

```c
void do_thing(struct sway_cursor *cursor,
              struct wlr_surface *a, int lx_a, int lx_y,
              struct wlr_surface *b, int lx_b, int ly_b) {
...
```

I'd weakly prefer that over what clang-format would generate. But, I think these sorts of cases are pretty rare in the codebase.  In

```diff
-void pango_printf(cairo_t *cairo, const char *font,
-               double scale, bool markup, const char *fmt, ...) {
+void pango_printf(cairo_t *cairo, const char *font, double scale, bool markup,
+               const char *fmt, ...) {
```

I don't see a "logical grouping" between {cairo, font} or {scale, markup, fmt}, so I think the autoformatting makes sense. Skimming through the diff with your settings, the majority of cases are like this.

---

> In any case I wouldn't want to re-format three quarters of Sway's codebase just to make clang-format happy.

Agreed, I wouldn't suggest reformatting all of Sway.

That said, I think there is value in having a reasonable approximation of the Sway style that works most of the time, rather than none at all. vim's clang-format integration (and I suspect that of other editors) allows only a specific (e.g. visually selected) block of code to be reformatted. A decent workflow is to:

- write some code
- clang-format it only the new code
- perform minor fixups to match Sway's style

(This is my workflow.)

The last step would frequently not be necessary for small-medium-sized edits.

Unfortunately, clang-format (at least in the current settings, as well as the ones in your paste) differs in hanging indentation from Sway's style, so even small edits tend to require fixups in the current world.

I think it would improve clang-format's usability in Sway a lot if we could avoid whatever causes it to perform this reformat:

```diff
                        dispatch_cursor_button(cursor, event->device, event->time_msec,
-                               BTN_LEFT, WLR_BUTTON_PRESSED);
+                                       BTN_LEFT, WLR_BUTTON_PRESSED);
```

As per the style guide this should be only indented once, not twice. If I could change a single thing in the style guide, it'd be to make this valid. Then both clang-format and JetBrains IDE defaults would "just work" here :)

I haven't dug deeply into whether this is possible to express in clang-format, otherwise that'd work too.

kennylevinsen:
We're not losing anything in this PR here as we never officially used clang-format so I don't see much reason to block this PR.

My two cents on *implementing* automatic formatting: In seatd I use clang-format in the build pipeline to save myself from making too many codestyle nits in reviews.

Indeed, the options are very limiting, and I had to make some compromises in style to be able to use it. I also only think it works if it has already been applied to the tree and is enforced by CI - the tool should remove work by guaranteeing that all code is always correctly styled, not create work by having me go through everything it changed (which is the entire file or entire repo) or remember what rules it always messes up.

But with that said having a formatting tool in CI is damn useful and saves reviewers and contributors time as contributors can run one command to fix their build themselves.

I don't know if there are other tools available that aligns more with our style guideline, or if the style change would be small enough to live with.

bl4ckb0ne:
Thanks o7

Xyene:
> we never officially used clang-format

Right, but this is similar to how [the kernel uses clang-format](https://www.kernel.org/doc/html/latest/process/clang-format.html): there's a best-effort .clang-format that developers use, possibly applying fixups afterwards. If the approximation is good enough, the number of fixups necessary is low. A decent approximation is better than none.

The result of merging this is that I'll revert it locally to continue using the approximation. I don't feel strongly about this; it's a minor inconvenience, but it doesn't feel like forward progress was made? I think committing @emersion's config instead would have been better.

Strong agree on every other point.

I've also seen clang-format checked on a per-hunk basis, upon commit. This allows the codebase to not have a big "reformat with clang-format" commit, but still require changed lines to be formatted. Eventual consistency and all. `.git-blame-ignore-revs` is a solution to the reformatting tree smash, but GitHub doesn't support it in the web viewer.

fluix-dev:
Not that I personally use it, but maybe placing this in `contrib` makes sense so that it's more clearly not "officially used" but useful for those that use it as Xyene does?

Only concern is here is do tools pick it up if it's not in the root directory?

emersion:
I think the clang-format file would be good `contrib/` material, but then developers would need to `ln` it for tools/editors to pick it up?

I was a little worried about tools/editors automatically picking up `.clang-format` and re-formatting the whole file when a small edit is made. But since the kernel has a top-level `.clang-format` file, maybe tools/editors don't do that?

Another thing is that I don't personally use it, so don't really want to maintain it. But if @Xyene volunteers to take care of it, that point is moot.

Relevant discussion on the Wayland repo: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/121

PR #6705, opened 2021-12-09
Fixup headless output names

emersion:
We use the headless backend to create a special fallback output
used when no other output is connected. However this messes up the
"real" headless output names users have come to expect (e.g.
currently the first headless output will be named "HEADLESS-2"
instead of "HEADLESS-1").

Fix this by setting the output name with [1].

[1]: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3395

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -5,6 +5,7 @@
 #include <time.h>
 #include <wayland-server-core.h>
 #include <wlr/backend/drm.h>
+#include <wlr/backend/headless.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_buffer.h>
 #include <wlr/types/wlr_drm_lease_v1.h>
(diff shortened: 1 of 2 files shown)


sway/desktop/output.c
```diff
@@ -833,13 +834,22 @@ static void handle_present(struct wl_listener *listener, void *data) {
...
+
 void handle_new_output(struct wl_listener *listener, void *data) {
 	struct sway_server *server = wl_container_of(listener, server, new_output);
 	struct wlr_output *wlr_output = data;
+
 	if (wlr_output == root->fallback_output->wlr_output) {
 		return;
 	}
 
+	if (wlr_output_is_headless(wlr_output)) {
```
bl4ckb0ne, line 847: Isnt this handleded by [36a2b19485ad299ee0039eb97c0f688d68205539] ?(https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/36a2b19485ad299ee0039eb97c0f688d68205539)

emersion: What do you mean?

That wlroots commit only allows compositors to set a custom output name. It doesn't change the fact that the name wlroots picks for the fallback output is "HEADLESS-1", which breaks users config files.

bl4ckb0ne: Oh right, the fallback input will be named `FALLBACK` after it has been named `HEADLESS-1` by wlroots, thus messing the order.

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, thanks :)

Issue #6546, opened 2021-09-23
Swaybg spinning up CPU


J0nnyMak0:
With [this](https://github.com/swaywm/wlroots/commit/59fa3637c348270cd1adb239296be1219538394b) commit, swaybg is loading CPU at 100%

stacyharper:
I think we suffer the same issue when using wayout (https://git.sr.ht/~proycon/wayout) (https://pkgs.alpinelinux.org/packages?name=wayout&branch=edge)

When wayout is running, swaybg burn 100% cpu, swaybar 20%. Virtual keyboard instantly crash.

~proycon recently reimplemented the drawing code to try to mitigate the issue.

PR #6709, opened 2021-12-13
Prevent a layer shell infinite looping case

stacyharper:
This commit try to fix a case we encounter when running wayout, a layer
shell to display text above the background.

When running wayout, other layer shell start to consume 100% of the CPU.
This seems related to a previous fix 744e85b. Then later, another commit
reworks this handle_surface_commit method 5fd5d643.

https://todo.sr.ht/~mil/sxmo-tickets/413

This is a tentative to fix the issue without missing the points of those
patches.

We basically try to not re-arrange layers if no layer changed.

Related to : https://github.com/swaywm/sway/issues/6546

This patch seems to solve our issue,  but am I missing something ?

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -308,8 +308,8 @@ static void handle_surface_commit(struct wl_listener *listener, void *data) {
 			wl_list_insert(&output->layers[layer_surface->current.layer],
 				&layer->link);
 			layer->layer = layer_surface->current.layer;
+			arrange_layers(output);
 		}
-		arrange_layers(output);
 	}
 
 	wlr_surface_get_extends(layer_surface->surface, &layer->extent);


emersion:
cc @vyivel 

The change looks fine to me, it seems like an oversight in the patches referenced above?

vyivel:
I think this will break cases when a layer surface changes its keyboard interactivity but not anything else.

emersion:
Shouldn't this be handled by `layer_surface->current.committed != 0`?

vyivel:
With this patch, `arrange_layers()` will only be called if the layer has changed (which implies `committed != 0`).
At least for me, GitHub screws up indentation up making the flow of logic a bit hard to see; I recommend opening the full file.

stacyharper:
This seems to cause some issues : 

- Sometime the swaybar do not display itself. Had to open my virtual keyboard to make all layer shell to pop
- bemenu --center is not centered (still verifying this)

vyivel:
Presumably what happens here is a loop of "Sway configures a layer surface <-> the client needlessly sets the same size/anchor/etc". While a client shouldn't really do this, it's perfectly valid behavior, so the fix would to be check not only `committed` mask but also if the values have actually changed.

emersion:
wlroots could have this logic, so that compositors don't need to manually check this.

vyivel:
That would make sense; also e.g. `wlr_output_enable()` unsets the flag if the value didn't change.
And for Consistency™ we would need to have the same logic everywhere. :P

stacyharper:
I tried this : https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3402

It seems to work. What do you guys think of that ?

emersion:
Superseded by the wlroots MR.

PR #6710, opened 2021-12-13
swaynag: remove xdg-output logic

emersion:
We can just get the output name from wl_output directly, now that
wl_output version 4 exists.


bl4ckb0ne (approved):
LGTM, thanks :)

PR #6712, opened 2021-12-14
meson.build: require wayland-protocols 1.24

colemickens:
As far as I can tell `ZWP_LINUX_DMABUF_FEEDBACK_V1_TRANCHE_FLAGS_SCANOUT` is introduced in wayland-protocols 1.24.

I noticed because of a build-time failure, rather than meson-prep-time failure.


emersion (approved):
Thanks for the fix!

PR #6716, opened 2021-12-14
swaybar: Fix multiple line on startup

### LINKED ISSUE #4496: swaybar shows multiple status lines at once

Hinara:
Fix #4496
Fix regression on swaybar displaying multiple lines from the
status_command introduced by 315b2bf.

### DIFF
--- a/swaybar/status_line.c
+++ b/swaybar/status_line.c
@@ -115,6 +115,14 @@ bool status_handle_readable(struct status_line *status) {
 		sway_log(SWAY_DEBUG, "Using text protocol.");
 		status->protocol = PROTOCOL_TEXT;
 		status->text = status->buffer;
+		newline = strrchr(status->buffer, '\n');
+		if (newline != NULL && newline[1] == '\0') {
+			newline[0] = '\0';
+			newline = strrchr(status->buffer, '\n');
+		}
+		if (newline != NULL) {
+			memmove(status->buffer, newline + 1, strlen(newline + 1) + 1);
+		}
 		// intentional fall-through
 	case PROTOCOL_TEXT:
 		errno = 0;


Hinara:
I did it as a quick fix, if it is enough for you good :)
If I would have fixed that more correctly I think I would have move the whole read logic only for i3 protocol, but as I don't really know that much about this i3 protocol, I have preferred to do the dirty trick of  only taking the last line read, but if the program does multiple write to output that the line might be broken.

swaybar/status_line.c
```diff
@@ -115,6 +115,14 @@ bool status_handle_readable(struct status_line *status) {
 		sway_log(SWAY_DEBUG, "Using text protocol.");
 		status->protocol = PROTOCOL_TEXT;
 		status->text = status->buffer;
+		newline = strrchr(status->buffer, '\n');
+		if (newline != NULL && newline[1] == '\0') {
+			newline[0] = '\0';
+			newline = strrchr(status->buffer, '\n');
```
emersion, line 121: If there is a single newline (e.g. "line 1 \n lin", with the second line cut off), `strrchr` will return NULL. I think we want to fallback to `status->buffer` here.

emersion: Hm. In case of partial reads the whole `getline` logic below will be broken anyways… I think to fix it properly we need to replace the logic above in `PROTOCOL_UNDEF` with a `getline` call.

emersion: Tried this and it broke the i3bar protocol parser, likely because `FILE*` has an internal buffer and the i3bar parser uses the FD directly. Oh well.

PR #6556, opened 2021-09-26
Ellipsize long window titles

FlexW:
cairo_image_surface_create() return CAIRO_STATUS_INVALID_SIZE on too
long window titles and creates then no surface. Therefore it makes
sense to limit the window title length to a reasonable length.

Fix: https://github.com/swaywm/sway/issues/6531

Signed-off-by: Felix Weilbach <felix.weilbach@t-online.de>

PR #6726, opened 2021-12-21
Add cairo_image_surface_create error handling

emersion:
cairo_image_surface_create can fail, e.g. when running out of
memory or when the size is too big. Avoid crashing in this case.

Closes: https://github.com/swaywm/sway/issues/6531


RPigott (approved):
LGTM and avoids the crash, just leaves the titlebar ugly. We'll still want ellipsizing or something like #6556 to keep things looking nice.

Issue #6337, opened 2021-06-20
Use after free in client→surface→subsurface destroy


numberZero:
Void Linux (amd64/musl), sway 1.6, wlroots 0.13.0.

Happens sometimes when closing swaylock, crashing sway with SEGV (under valgrind it reports the error but continues normally; under gdb it crashed at `struct sway_output *output = wlr_output->data;` as `wlr_output` was garbage [not even null]). Valgrind trace:
```
==5135== Invalid read of size 8
==5135==    at 0x1267E6: UnknownInlinedFun (layer_shell.c:387)
==5135==    by 0x1267E6: subsurface_handle_unmap (layer_shell.c:401)
==5135==    by 0x523D9CB: wlr_signal_emit_safe (signal.c:29)
==5135==    by 0x522D0CC: UnknownInlinedFun (wlr_surface.c:1015)
==5135==    by 0x522D0CC: UnknownInlinedFun (wlr_surface.c:1010)
==5135==    by 0x522D0CC: subsurface_handle_parent_destroy.lto_priv.0 (wlr_surface.c:1083)
==5135==    by 0x523D9CB: wlr_signal_emit_safe (signal.c:29)
==5135==    by 0x5233338: surface_handle_resource_destroy.lto_priv.0 (wlr_surface.c:677)
==5135==    by 0x51C09DE: ??? (in /usr/lib/libwayland-server.so.0.1.0)
==5135==    by 0x51C1072: wl_client_destroy (in /usr/lib/libwayland-server.so.0.1.0)
==5135==    by 0x51C14B2: ??? (in /usr/lib/libwayland-server.so.0.1.0)
==5135==    by 0x51C0B21: wl_event_loop_dispatch (in /usr/lib/libwayland-server.so.0.1.0)
==5135==    by 0x51C1204: wl_display_run (in /usr/lib/libwayland-server.so.0.1.0)
==5135==    by 0x11AD3C: UnknownInlinedFun (server.c:260)
==5135==    by 0x11AD3C: main (main.c:431)
==5135==  Address 0x1522fc60 is 0 bytes inside a block of size 216 free'd
==5135==    at 0x48BAA4B: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==5135==    by 0x523D9CB: wlr_signal_emit_safe (signal.c:29)
==5135==    by 0x522ACBE: layer_surface_destroy (wlr_layer_shell_v1.c:244)
[...]

PR #6728, opened 2021-12-23
Destroy sub-surfaces with parent layer-shell surface

emersion:
Closes: https://github.com/swaywm/sway/issues/6337


RPigott (approved):
Well I wasn't able to reproduce #6337, but these changes LGTM.

PR #6739, opened 2022-01-04
[IPC] Add repeat delay/rate info to keyboard

### LINKED ISSUE #6735: Get repeat delay and rate using get_inputs
# Description
Expose the current keyboard repeat delay and rate using

    swaymsg -t get_inputs "type:keyboard"

As far as I'm aware, there's currently no way to read those values. Writing them is easy though, for example:

    swaymsg input type:keyboard repeat_delay 1000

One specific use case is to temporarily reduce the `repeat_delay` in order to avoid a key press (in my case it's `XF86RotateWindows`) being registered multiple times and thus a script for rotating the screen being called too often:

    # TODO: Save current repeat delay to a variable

    # Reduce the repeat delay to avoid calling this script multiple times
    swaymsg input type:keyboard repeat_delay 1000

    if [ $(swaymsg -t get_outputs | jq -r '.[] | select(.name=="LVDS-1") | .transform') = "normal" ]; then swaymsg output LVDS-1 transform 180; else swaymsg output LVDS-1 transform normal; fi

    # TODO: Restore the original repeat delay value
    swaymsg input type:keyboard repeat_delay 250

SethBarberee:
Closes #6735

wlroots already has the info in the struct so let's access it and print it out.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -980,6 +980,11 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 		struct wlr_keyboard *keyboard = device->wlr_device->keyboard;
 		struct xkb_keymap *keymap = keyboard->keymap;
 		struct xkb_state *state = keyboard->xkb_state;
+		
+		json_object_object_add(object, "repeat_delay", 
+			json_object_new_int(keyboard->repeat_info.delay));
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

PR #6740, opened 2022-01-04
input: Fix next/prev layout switch for grouped devices

nfraprado:
Input devices in a seat can be grouped so that configuration changes on
one device, like the active layout, are automatically propagated to the
other devices.

The current logic for xkb_switch_layout wasn't accounting for this,
however. It was looping through all input devices and changing the
layout for each one, causing grouped devices to have their layouts
changed multiple times. This wasn't noticed when setting the layout to a
fixed index, but caused the wrong layout to be selected when using
the 'next' and 'prev' keywords.

In order to fix this issue, loop over each seat and, if the input
devices in the seat are grouped, change the layout only for the first
device in the seat.

### DIFF
--- a/sway/commands/input/xkb_switch_layout.c
+++ b/sway/commands/input/xkb_switch_layout.c
@@ -3,6 +3,7 @@
 #include "sway/config.h"
 #include "sway/commands.h"
 #include "sway/input/input-manager.h"
+#include "sway/input/seat.h"
 #include "log.h"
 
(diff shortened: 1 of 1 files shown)


emersion:
Hm, I don't think this works when there are multiple groups per seat, e.g. if one attaches two physical keyboards to a single seat.

emersion:
Superseded by https://github.com/swaywm/sway/pull/6966

PR #6742, opened 2022-01-06
input/seat: unset has_focus when focus_stack becomes empty

### LINKED ISSUE #6395: Segmentation fault on resume from S3 sleep
### **Sway Version:**

`sway version 1.6-96102184 (Jul 19 2021, branch 'master')`

### **Debug Log:**

https://gist.github.com/cpp-10/0444900ff1791f08999e6ee07f818907

### **Configuration File:**

The bug can be reproduced with the default configuration + these lines (complete config at https://gist.github.com/cpp-10/74b43cca6823cd5dcd009944b9370b68):
  
```
exec swayidle -w before-sleep 'swaylock -f'

set $laptop eDP-1
bindswitch --reload --locked lid:on output $laptop disable
bindswitch --reload --locked lid:off output $laptop enable
```

### **Stack Trace:**

Full `coredumpctl gdb sway` + `bt full` logs: https://gist.github.com/cpp-10/0caa8c80ed78eb76d9eaa8c8edd5c194

```
#0  0x00005605ba4105a4 in node_is_view (node=0x0) at ../sway/tree/node.c:41
#1  0x00005605ba3e0b67 in seat_set_exclusive_client (seat=0x5605bc0ea530, client=0x5605bc4de3c0) at ../sway/input/seat.c:1319
        focus = 0x0
        now = {tv_sec = 140736342280040, tv_nsec = 140736342280256}
        point = 0x7fffbbb04770
#2  0x00005605ba3d436e in handle_inhibit_activate (listener=0x5605bc0ea3d8, data=0x5605bc0ea460) at ../sway/input/input-manager.c:284
        input_manager = 0x5605bc0ea380
        seat = 0x5605bc0ea530
#3  0x00007f98e5289fc3 in wlr_signal_emit_safe (signal=0x5605bc0ea490, data=0x5605bc0ea460) at ../subprojects/wlroots/util/signal.c:29
        pos = 0x5605bc0ea3d8
        l = 0x5605bc0ea3d8
        cursor = {link = {prev = 0x5605bc0ea3d8, next = 0x7fffbbb04650}, notify = 0x7f98e5289f0d <handle_noop>}
        end = {link = {prev = 0x7fffbbb04630, next = 0x5605bc0ea490}, notify = 0x7f98e5289f0d <handle_noop>}
#4  0x00007f98e52655c8 in inhibit_manager_get_inhibitor (client=0x5605bc4de3c0, resource=0x5605bc6b44b0, id=3) at ../subprojects/wlroots/types/wlr_input_inhibitor.c:74
        manager = 0x5605bc0ea460
        wl_resource = 0x5605bc6b4390
#5  0x00007f98e428facd in  () at /usr/lib/libffi.so.7
#6  0x00007f98e428f03a in  () at /usr/lib/libffi.so.7
#7  0x00007f98e4b99124 in  () at /usr/lib/libwayland-server.so.0
#8  0x00007f98e4b9457c in  () at /usr/lib/libwayland-server.so.0
#9  0x00007f98e4b9707a in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#10 0x00007f98e4b94be7 in wl_display_run () at /usr/lib/libwayland-server.so.0
#11 0x00005605ba3c74a8 in server_run (server=0x5605ba441560 <server>) at ../sway/server.c:261
#12 0x00005605ba3c68c3 in main (argc=4, argv=0x7fffbbb04dd8) at ../sway/main.c:395
        verbose = 0
        debug = 1
        validate = 0
        allow_unsupported_gpu = 0
        long_options =
            {{name = 0x5605ba423080 "help", has_arg = 0, flag = 0x0, val = 104}, {name = 0x5605ba423085 "config", has_arg = 1, flag = 0x0, val = 99}, {name = 0x5605ba42308c "validate", has_arg = 0, flag = 0x0, val = 67}, {name = 0x5605ba423095 "debug", has_arg = 0, flag = 0x0, val = 100}, {name = 0x5605ba42309b "version", has_arg = 0, flag = 0x0, val = 118}, {name = 0x5605ba4230a3 "verbose", has_arg = 0, flag = 0x0, val = 86}, {name = 0x5605ba4230ab "get-socketpath", has_arg = 0, flag = 0x0, val = 112}, {name = 0x5605ba4230ba "unsupported-gpu", has_arg = 0, flag = 0x0, val = 117}, {name = 0x5605ba4230ca "my-next-gpu-wont-be-nvidia", has_arg = 0, flag = 0x0, val = 117}, {name = 0x0, has_arg = 0, flag = 0x0, val = 0}}
        config_path = 0x5605bb96fb60 "/home/charlie/sway-dbg/config"
        usage = 0x5605ba422af0 "Usage: sway [options] [command]\n\n  -h, --help", ' ' <repeats 13 times>, "Show help message and quit.\n  -c, --config <config>  Specify a config file.\n  -C, --validate         Check the validity of the config file, th"...
        c = -1
```

### **Description:**

This bug occurs with a combination of "clamshell mode" configuration for sway, a swayidle hook to start swaylock on system sleep and entering S3 sleep. Sway crashes on resume.

Steps to reproduce:
 - Start sway.
 - Shut laptop lid to induce S3 sleep. Wait for laptop to sleep.
 - Open lid to resume.
 - Sway has crashed.
 
Machine configuration at the time:
 - Thinkpad T14s (AMD Ryzen 7 PRO 4750U, open source amdgpu driver)
 - Kernel: 5.10.50-1-lts
 - No external devices (except power cable) plugged in.

tchebb:
We currently track the focus of a seat in two ways: we use a list called `focus_stack` to track the order in which nodes have been focused, with the first node representing what's currently focused, and we use a variable called `has_focus` to indicate whether anything has focus—i.e. whether we should actually treat that first node as focused at any given time.

In a number of places, we treat has_focus as implying that a focused node exists. If it's true, we attempt to dereference the return value of `seat_get_focus()`, our helper function for getting the first node in `focus_list`, with no further checks. But this isn't quite correct with the current implementation of `seat_get_focus()`: not only does it return NULL when `has_focus` is false, it also returns NULL when `focus_stack` contains no items.

In most cases, `focus_stack` never becomes empty and so this doesn't matter at all. Since `focus_stack` stores a history of focused nodes, we rarely remove nodes from it. The exception to this is when a node itself goes away. In that case, we call `seat_node_destroy()` to remove it from `focus_stack` and free it. But we don't unset `has_focus` if we've removed the final node! This lets us get into a state where `has_focus` is true but `seat_get_focus()` returns NULL, leading to a segfault when we try to dereference it.

Fix the issue both by updating `has_focus` in `seat_node_destroy()` and by adding an assertion in `seat_get_focus()` that ensures `focus_stack` and `has_focus` are in sync, which will make it easier to track down similar issues in the future.

Fixes #6395.

[1] There's some discussion in #1585 from when this was implemented about whether `has_focus` is actually necessary; it's possible we could remove it entirely, but for the moment this is the architecture we have.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -51,6 +51,16 @@ static void seat_device_destroy(struct sway_seat_device *seat_device) {
 static void seat_node_destroy(struct sway_seat_node *seat_node) {
 	wl_list_remove(&seat_node->destroy.link);
 	wl_list_remove(&seat_node->link);
+
+	/*
+	 * This is the only time we remove items from the focus stack without
+	 * immediately re-adding them. If we just removed the last thing,
+	 * mark that nothing has focus anymore.
+	 */
+	if (wl_list_empty(&seat_node->seat->focus_stack)) {
+		seat_node->seat->has_focus = false;
+	}
+
 	free(seat_node);
 }
 
@@ -1415,9 +1425,8 @@ struct sway_node *seat_get_focus(struct sway_seat *seat) {
 	if (!seat->has_focus) {
 		return NULL;
 	}
-	if (wl_list_empty(&seat->focus_stack)) {
-		return NULL;
-	}
+	sway_assert(!wl_list_empty(&seat->focus_stack),
+			"focus_stack is empty, but has_focus is true");
 	struct sway_seat_node *current =
 		wl_container_of(seat->focus_stack.next, current, link);
 	return current->node;


emersion (approved):
LGTM, thanks for the fix!

PR #6746, opened 2022-01-08
swaybar: tray_padding is scaled

nmschulte:
As @xdavidwu noted, and @emersion directed, re: fixes for #6504: -- https://github.com/swaywm/sway/pull/6517#issuecomment-922702239

> > tray_padding config is currently defined in device pixels
> 
> That seems like an oversight. I wouldn't mind a follow-up PR to change it to surface-local coords. But yeah better to keep it as-is for this bugfix PR, so that we can more easily revert that change if needed.

----

Unfortunately, no, it seems we cannot ignore scaling concerns when drawing the tray item icons.  This is because the displayed icon-size (e.g. 24px vs 48px size) is determined at the chosen scale, and so further this size must be scaled back, and also the scale imparted as part of the libcairo drawing operation.

### DIFF
--- a/swaybar/tray/tray.c
+++ b/swaybar/tray/tray.c
@@ -116,8 +116,8 @@ uint32_t render_tray(cairo_t *cairo, struct swaybar_output *output, double *x) {
 		}
 	} // else display on all
 
-	if ((int)(output->height * output->scale) <= 2 * config->tray_padding) {
-		return (2 * config->tray_padding + 1) / output->scale;
+	if (output->height <= 2 * (uint32_t)config->tray_padding) {
+		return 2 * config->tray_padding + 1;
 	}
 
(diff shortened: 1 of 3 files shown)


swaybar/tray/tray.c
```diff
@@ -116,8 +116,8 @@ uint32_t render_tray(cairo_t *cairo, struct swaybar_output *output, double *x) {
 		}
 	} // else display on all
 
-	if ((int)(output->height * output->scale) <= 2 * config->tray_padding) {
-		return (2 * config->tray_padding + 1) / output->scale;
+	if (output->height <= 2 * (uint32_t)config->tray_padding) {
```
nmschulte, line 119: Should `tray_padding` and the other swaybar configs be converted to `uint32_t`?

Issue #6562, opened 2021-09-29
Use wlroots' scene-graph API


emersion:
I think the wlroots' scene-graph API is now powerful enough for Sway to use.

Here's how I think we can organize the scene-graph nodes for Sway. The main hurdle is painting the fullscreen views and popups in the correct order.

- Root
  - Output 1 (used to offset output coords)
    - Background and bottom layers
    - Workspace 1 (enabled because focused)
      - View container (used to offset container coords)
        - Borders (rects)
        - View 1
          - Main surface + subsurfaces
          - Saved buffers, if any
          - Decorations (border rects + titlebar buffer)
        - View 2
          - …
      - Popups container (used to offset container coords)
        - Popups for view 1
        - Popups for view 2
    - Workspace 2 (disabled because unfocused)
      - View
        - …
    - Top and overlay layers
  - Output 2
    - Background and bottom layers (disabled because there's a fullscreen view)
    - Workspace 3 (disabled because there's a fullscreen view)
    - Top layer (disabled because there's a fullscreen view)
    - Fullscreen view
      - …
    - Fullscreen popups
    - Overlay layer
  - Floating
    - View
      - …

Xyene:
>   - Floating
    - View
      - …

Note that this isn't what i3 does. i3 has a stack of floating views _per output_, and when a view spans across another output, it'll either appear under all views of that output or over (depending on output order), but never in-between.

I'm personally in favor of not replicating this quirk when we move to the scene-graph API. It's a little janky and leads to bugs like https://github.com/swaywm/sway/issues/6073 (and likely others we haven't run into yet).

PR #6747, opened 2022-01-08
Fix inconsistent floating window z order across outputs

Nefsen402:
Currently floating windows when appearing across outputs (in a multi-monitor setup) will cover other floating windows unconditionally sometimes even if I try to focus the covered floating window. This happens when a window belongs to say workspace 1. If that window is strung across multiple monitors such that it also covers contents of workspace 2, floating windows in 2 will always be either behind or above floating windows in workspace 1. This makes working impossible sometimes unless you move the windows out of the way.

My fix is to maintain a counter that increments everytime a floating window wants to be shown on top. This counter is shared across the entire seat so that even across workspaces, floating windows maintain the expected z order.

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -113,6 +113,12 @@ struct sway_seat {
 				// sway_keyboard_shortcuts_inhibitor::link
 
 	struct wl_list link; // input_manager::seats
+
+	// Used by sway_container.floating_order. This value will increment as
+	// floating windows request to be shown on top, this makes sure that a
+	// value can be assigned to a floating window that is guaranteed
+	// to be highter (ignoring overflow) than other floating windows.
+	uint32_t floating_counter;
 };
 
 struct sway_pointer_constraint {
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -113,6 +113,12 @@ struct sway_container {
 	// Hidden scratchpad containers have a NULL parent.
 	bool scratchpad;
 
+	// Value specifing in which order floating windows should be shown
+	// higher values means that the floating window should appear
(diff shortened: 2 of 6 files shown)


Nefsen402:
Latest version fixes scratchpad windows not going to top when called upon

Xyene:
I'm in favour of adopting this behavior, but note:

- The current behavior is what i3 does, even if quirky and unintuitive. I would prefer to diverge from i3 here, but don't know how others feel.
- https://github.com/swaywm/sway/issues/6562 might be a better place to address this, as lots of stacking code will need to be replaced there. Any code we might merge in this PR is unlikely to survive the transition to scene-graph.

@emersion, thoughts?

Nefsen402:
> * [Use wlroots' scene-graph API #6562](https://github.com/swaywm/sway/issues/6562) might be a better place to address this, as lots of stacking code will need to be replaced there. Any code we might merge in this PR is unlikely to survive the transition to scene-graph.

It seems like the proposed plan in the issue tracker would make a global floating window list which would where each floating window would have a reference to its owned workspace so that the proper floating windows would be hidden if the workspace is not in view. This is functionally the same as what is being implement in this PR, but a bit more robust because there's no chance for a number to overflow, but I don't think that's a problem. A user would need to set a window to be on top 4 billion times before running into problems in the same session.

Nefsen402:
I spoke with @emersion on IRC, and we seem to be in agreement that the behavior of floating windows as outlined here is okay. This PR however is a hack, and probably shouldn't be merged. I will start working on the port to wlr_scene and the floating window behavior should be mirrored there.

vyivel:
Superseded by https://github.com/swaywm/sway/pull/6844.

Issue #6737, opened 2022-01-03
swaymsg "floating enable;move position 50 px" crashes


chmouel:
Version: sway version 1.6.1

Running : 

`swaymsg "floating enable;move position 50 px"`

 crash sway (50 is arbitrary any number will crash)

Crash Snippet 

```
0:00:29.364 [DEBUG] [sway/ipc-server.c:199] New client: fd 102
00:00:29.364 [DEBUG] [sway/ipc-server.c:219] Client 102 readable
00:00:29.364 [INFO] [sway/commands.c:257] Handling command 'floating enable'
00:00:29.364 [DEBUG] [sway/tree/arrange.c:263] Usable area for ws: 3440x1408@0,32
00:00:29.364 [DEBUG] [sway/tree/arrange.c:293] Arranging workspace '1' at 6.000000, 38.000000
00:00:29.364 [INFO] [sway/commands.c:257] Handling command 'move position 50 px'
(EE) failed to read Wayland events: Broken pipe
wl_display_dispatch: Broken pipe
[003 09:00:52.722745] [glfw error 65544]: Wayland: fatal display error: Connection reset by peer
Gdk-Message: 09:00:52.722: Error reading events from display: Broken pipe
[2022-01-03 09:00:52.722] [error] Mode: Unable to receive IPC header
[2022-01-03 09:00:52.722] [error] Mode: Unable to receive IPC header
```

This seems to be a string parsing issue, "50px will work". If you can't reproduce please let me know and i'll send a coredump/full log and test on the main branch.

emersion:
Please provide a stack trace. You can do so by compiling from source, reproducing the crash and then running `coredumpctl gdb` and then `bt full`.

Here are some instructions to compile from source: https://github.com/swaywm/sway/wiki/Development-Setup#compiling-as-a-subproject

ErikReider:
Same issue

mstoeckl:
I can reproduce this. The relevant portion of the stack trace:
```
Program terminated with signal SIGSEGV, Segmentation fault.
(gdb) bt full
#0  0x00007fa4545455a0 in ____strtoll_l_internal () at /usr/lib/libc.so.6
#1  0x0000562d571e76a2 in parse_movement_amount (argc=argc@entry=0, argv=0x562d5915a100, amount=amount@entry=0x7ffe85f41bf0) at ../sway/common/util.c:84
        err = 0x562d590a21b2 ""
[...]

PR #6749, opened 2022-01-09
Add safety assert in parse_movement_unit

emersion:
Let's add this just in case a caller passes argc == 0.

References: https://github.com/swaywm/sway/issues/6737#issuecomment-1008082540

cc @mstoeckl @David96


mstoeckl (approved):
LGTM; I tested this by rebasing onto a branch without the fix (1bf1d84b7535c3c132240ed7b18414dc6cfe7e8a) to 6737 .

emersion:
Thanks for the review!

PR #6751, opened 2022-01-10
Revert "build: use list for sdbus dep"

kchibisov:
The reverted commit introduced the use of list for sdbus deps, however
it was assuming that all packages which were in a list has a version
higher than 239. That is true for libsystemd and libelogind, since they
use the same versions, however basu is using version numbers which are
way lower than what libsystemd/libelogind are using, so basu only build
is failing.

This reverts commit 02b412a3d4e930237a1d16554af6e1e7d1855c89.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -92,15 +92,22 @@ if get_option('sd-bus-provider') == 'auto'
 	if not get_option('tray').disabled()
 		assert(get_option('auto_features').auto(), 'sd-bus-provider must not be set to auto since auto_features != auto')
 	endif
-	sdbus = dependency(['libsystemd', 'libelogind', 'basu'],
-		required: get_option('tray'),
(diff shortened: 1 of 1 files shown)


kchibisov:
Cloning wlroots on CI fails, so the failure is unrelated to reverted commit (given that alpine job managed to clone it).

emersion:
Can we keep using the dep list for `['libsystemd', 'libelogind']`?

Alternatively, we could drop the version requirement (or manually perform it, or convert to a symbol check).

kchibisov:
>Can we keep using the dep list for ['libsystemd', 'libelogind']?

Sure, done.

emersion:
Can you rebase against master to fix CI?

kchibisov:
>Can you rebase against master to fix CI?

Sure, done.

emersion (approved):
LGTM, thanks!

PR #6756, opened 2022-01-11
Print deprecation notice when running SUID

kennylevinsen:
SUID privilege drop is needed for the "builtin"-backend of libseat,
which copied our old "direct" backend behavior for the sake of
compatibility and ease of transition.

libseat now has a better alternative in the form of seatd-launch. It
uses the normal seatd daemon and libseat backend and takes care of SUID
for us.

Add a soft deprecation warning to highlight our future intent of
removing this code. The deprecation cycle is needed to avoid surprises
when sway no longer drops privileges.

We should put this in the release notes as well.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -153,6 +153,9 @@ static void log_kernel(void) {
 
 static bool drop_permissions(void) {
 	if (getuid() != geteuid() || getgid() != getegid()) {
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

PR #6757, opened 2022-01-11
meson: check=false on run_command

kennylevinsen:
Future meson releases will change the default and warns when the
implicit default is used, breaking builds.

Explicitly set check=false to maintain behavior and silence warnings.


emersion (approved):
Thanks!

PR #6764, opened 2022-01-17
xwayland: listen to `request_activate` event 

yrhki:
When REAPER submenu is closed `XCB_CLIENT_MESSAGE` with type
`NET_ACTIVE_WINDOW` is sent to set focus to parent menu.

Closes: #6324

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -121,6 +121,20 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
 	}
 }
 
+static void unmanaged_handle_request_activate(struct wl_listener *listener, void *data) {
+	struct wlr_xwayland_surface *xsurface = data;
+	if (!xsurface->mapped) {
+		return;
+	}
+	struct sway_seat *seat = input_manager_current_seat();
+	struct sway_container *focus = seat_get_focused_container(seat);
+	if (focus && focus->view && focus->view->pid != xsurface->pid) {
+		return;
+	}
+
+	seat_set_focus_surface(seat, xsurface->surface, false);
+}
+
 static void unmanaged_handle_destroy(struct wl_listener *listener, void *data) {
 	struct sway_xwayland_unmanaged *surface =
 		wl_container_of(listener, surface, destroy);
@@ -129,6 +143,7 @@ static void unmanaged_handle_destroy(struct wl_listener *listener, void *data) {
 	wl_list_remove(&surface->unmap.link);
 	wl_list_remove(&surface->destroy.link);
 	wl_list_remove(&surface->override_redirect.link);
+	wl_list_remove(&surface->request_activate.link);
 	free(surface);
 }
 
@@ -176,6 +191,8 @@ static struct sway_xwayland_unmanaged *create_unmanaged(
 	surface->destroy.notify = unmanaged_handle_destroy;
 	wl_signal_add(&xsurface->events.set_override_redirect, &surface->override_redirect);
 	surface->override_redirect.notify = unmanaged_handle_override_redirect;
+	wl_signal_add(&xsurface->events.request_activate, &surface->request_activate);
+	surface->request_activate.notify = unmanaged_handle_request_activate;
 
 	return surface;
 }
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -169,6 +169,7 @@ struct sway_xwayland_unmanaged {
 
 	int lx, ly;
 
+	struct wl_listener request_activate;
 	struct wl_listener request_configure;
 	struct wl_listener request_fullscreen;
 	struct wl_listener commit;


progandy:
What happens if you navigate the menu with the keyboard instead of the mouse?

emersion:
I don't understand why re-focusing is necessary here.

emersion:
Giving focus to an unmapped surface doesn't make sense, the surface isn't shown on screen.

yrhki:
REAPER submenus don't have parent surface so https://github.com/swaywm/sway/blob/0ffd8178febb110a9b4e2e5901c30b72288e5359/sway/desktop/xwayland.c#L108-L112 fails and when it reaches https://github.com/swaywm/sway/blob/0ffd8178febb110a9b4e2e5901c30b72288e5359/sway/desktop/xwayland.c#L118 whole menu goes away.

I also did some logging for swell (REAPER library)

	[swell-wnd-generic.cpp:456] DestroyWindow (menu) (__SWELL_MENU) S X:636+369 Y:99+92 0x12fb830
	[swell-menu-generic.cpp:1044] submenuWndProc WM_DESTROY (menu) (__SWELL_MENU) S X:636+369 Y:99+92 0x12fb830
	[swell-generic-gdk.cpp:279] swell_oswindow_focus (menu) (__SWELL_MENU) P X:204+432 Y:38+721 0x1262910
	[swell-generic-gdk.cpp:290] + gdk_window_raise (menu) (__SWELL_MENU) P X:204+432 Y:38+721 0x1262910
	[swell-generic-gdk.cpp:295] + gdk_window_focus (menu) (__SWELL_MENU) P X:204+432 Y:38+721 0x1262910
	[swell-wnd-generic.cpp:406] RecurseDestroyWindow (menu) (__SWELL_MENU) S X:636+369 Y:99+92 0x12fb830
	[swell-generic-gdk.cpp:243] swell_oswindow_destroy (menu) (__SWELL_MENU) S X:636+369 Y:99+92 0x12fb830
	[swell-generic-gdk.cpp:249] gdk_window_destroy (menu) (__SWELL_MENU) S X:636+369 Y:99+92 0x12fb830

Parent menu should be in focus before submenu is destroyed, but I didn't see sway logs didn't show anything happening.

yrhki:
Simplified code a bit. REAPER menus still work but pressing `esc` when mouse is over sub-menu closes whole menu.

yrhki:
@emersion I looked deeper into this and this seems to be proper solution.

Xyene:
This looks like a proper fix, thanks for looking into it!

Out of curiosity, does this remove the need to do https://github.com/swaywm/sway/blob/8daa6321cfe1134042613374fba0266b8671bb55/sway/desktop/xwayland.c#L105-L120? (In case it does, not asking you to make the change in this PR, just wondering if it does.)

emersion:
We should make sure this doesn't allow an X11 app to steal focus from a Wayland app. IIRC we already have some kind of client matching elsewhere, maybe we can do something similar?

yrhki:
@Xyene Removing https://github.com/swaywm/sway/blob/8daa6321cfe1134042613374fba0266b8671bb55/sway/desktop/xwayland.c#L108-L112 breaks menus in IntelliJ IDEA.

Also found that in IDEA `set_hints` event is called on sub-menu before `unmap` happens, but `sway_xwayland_unmanaged` dosen't have it. It would be helpful if `wlr_signal_emit_safe` would log when nothing is called.

Xyene:
Thanks for testing.

Yeah, I think that doesn't sound unreasonable for a debug-level message, it seems likely this issue would have been found quicker if we had it.

yrhki:
I think this will do it. Wouldn't it be better if it was limited to the same process similar to `xwm_handle_focus_in`?

emersion:
> Doesn't limiting to the same process apply here as well similar to xwm_handle_focus_in?

Yup, that would be nice.

sway/desktop/xwayland.c
```diff
@@ -121,6 +121,22 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
...
 }
 
+static void unmanaged_handle_request_activate(struct wl_listener *listener, void *data) {
+	struct wlr_xwayland_surface *xsurface = data;
+	if (!xsurface->mapped) {
+		return;
+	}
+	struct sway_seat *seat = input_manager_current_seat();
+	struct sway_container *focus = seat_get_focused_container(seat);
+	if (focus && focus->view->pid != xsurface->pid) {
```
emersion, line 131: It's possible that the current focus isn't a view. We should check for `focus->view` here as well.

sway/desktop/xwayland.c
```diff
@@ -121,6 +121,22 @@ static void unmanaged_handle_unmap(struct wl_listener *listener, void *data) {
...
+		return;
+	}
+
+	if (wlr_xwayland_or_surface_wants_focus(xsurface)) {
```
emersion, line 135: Maybe we can just remove this condition?

emersion (approved):
LGTM, thanks!

salkin-mada:
Fantastic! Thanks so much @yrhki

PR #6767, opened 2022-01-17
input/cursor: treat swipe begin as idle activity too

Xyene:
I accidentally overlooked this in fd53f80.

Thanks @iyzana for pointing it out.


emersion (approved):
Missed it as well. Thanks!

PR #6769, opened 2022-01-18
build: execute wlroots subproject before finding deps

emersion:
wlroots often requires dependencies more recent than Sway's.
Executing the wlroots subproject first will give Meson a chance to
find these newer dependencies, possibly via subprojects.

The subproject will override the "wlroots" dependency when executed,
so we don't need to use get_variable anymore.

References: https://github.com/swaywm/sway/pull/6498#issuecomment-1001746017

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -35,6 +35,15 @@ if is_freebsd
 	add_project_arguments('-D_C11_SOURCE', language: 'c')
 endif
 
+# Execute the wlroots subproject, if any
+wlroots_version = ['>=0.16.0', '<0.17.0']
+subproject(
+	'wlroots',
+	default_options: ['examples=false'],
+	required: false,
+	version: wlroots_version,
+)
+
 jsonc = dependency('json-c', version: '>=0.13')
 pcre = dependency('libpcre')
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -35,13 +35,23 @@ if is_freebsd
 	add_project_arguments('-D_C11_SOURCE', language: 'c')
 endif
 
+# Execute the wlroots subproject, if any
+wlroots_version = ['>=0.16.0', '<0.17.0']
+subproject(
```
bl4ckb0ne, line 40: When did meson allowed to this?

emersion: 0.54.0, since that's when `meson.override_dependency` was introduced.

bl4ckb0ne: The syntax feels a bit weird. So this will automatically use the dependency in `subprojects` when calling `dependency` if the wlroots subproject is there?

emersion: Yeah: executing the subproject will override the "wlroots" dep.

bl4ckb0ne: Good to know!

bl4ckb0ne (approved):
LGTM, thanks!

PR #6770, opened 2022-01-18
transaction: destroying nodes aren't hidden

### LINKED ISSUE #6473: Flicker of the background
sway 1.6.1

Default configuration with the background image.

1. Open two terminals (kitty in my case) using `Mod+Enter`. Vertically split.
2. Type `exit` in the second one.
3. Second terminal closes but the background flickers for ~15ms (1-2 frames).

It's very annoying. Flicker visible in >90 % times.

- Arch Linux stable
- kernel 5.13.13
- amdgpu

Emantor:
Commit 37d7bc69986f ("transaction: Only wait for ack from visible
views") introduced a check which uses view_is_visible() to check if a view
is still visible on the screen. However view_is_visible() will early
return in case the node is in the destroying state. This is incorrect
for transactions, since a destroying view which is visible will trigger
configure events for other clients. This bug was visible when repeatedly
opening and closing two views side by side, since we ignore the
destroying node we get a frame where the still open view is shown with
the old configure values and the rest is the desktop background. The
next frame is than correct again.

Fix this by considering destroying views as visible, we correctly wait
for them and send the configure events to other views in time, fixing
the background flicker.

Fixes #6473

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -402,7 +402,7 @@ static void transaction_commit(struct sway_transaction *transaction) {
 		struct sway_transaction_instruction *instruction =
 			transaction->instructions->items[i];
 		struct sway_node *node = instruction->node;
-		bool hidden = node_is_view(node) &&
+		bool hidden = node_is_view(node) && !node->destroying &&
 			!view_is_visible(node->sway_container->view);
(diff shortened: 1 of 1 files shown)


kennylevinsen (approved):
LGTM!

PR #6779, opened 2022-01-22
cmd/swap: error on swapping a container with itself

### LINKED ISSUE #6724: Sway crashes when attempting to swap focused fullscreen container with itself
### Steps To Reproduce
1. Open a terminal, then full-screen it
2. Run `sway mark "x"` then `sway swap container with mark "x"`

### Expected behavior
Sway crashes. I am running sway version 1.6.1 on void linux with musl libc and runit init. It also might be worth mentioning that I am not using elogind, and instead using seatd and setting the `XDG_RUNTIME_DIR` in my .profile.

[Debug log](https://gist.github.com/NumenHackerCracker/4d2ee6e75e618c56df4c0e6361cbb499)

RPigott:
Fixes #6724. Swapping a con with itself is an error in i3.

### DIFF
--- a/sway/commands/swap.c
+++ b/sway/commands/swap.c
@@ -126,10 +126,10 @@ void container_swap(struct sway_container *con1, struct sway_container *con2) {
 	}
 
 	enum sway_fullscreen_mode fs1 = con1->pending.fullscreen_mode;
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

Issue #6784, opened 2022-01-25
Undefined behaviour snprintf compiler warnings (errors) with GCC 12


tokyovigilante:
### Please read the following before submitting:
- Please do NOT submit bug reports for questions. Ask questions on IRC at #sway on Libera Chat.
- Proprietary graphics drivers, including nvidia, are not supported. Please use the open source equivalents, such as nouveau, if you would like to use Sway.
- Please do NOT submit issues for information from the github wiki. The github wiki is community maintained and therefore may contain outdated information, scripts that don't work or obsolete workarounds.
  If you fix a script or find outdated information, don't hesitate to adjust the wiki page.

### Please fill out the following:
- **Sway Version:**
sway HEAD@cd1ee0e/wlroots HEAD

- **Description:**
I upgraded to Fedora rawhide due to Sway 1.7+'s requirement for wayland >-1.20, which includes GCC 12. This raises the following build warnings:

```
ninja: Entering directory `build'
[493/539] Compiling C object sway/sway.p/commands_output_background.c.o
../sway/commands/output/background.c: In function ‘output_cmd_background’:
../sway/commands/output/background.c:115:25: warning: ‘sprintf’ argument 4 overlaps destination object ‘src’ [-Wrestrict]
  115 |                         sprintf(src, "%s/%s", conf_path, rel_path);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../sway/commands/output/background.c:81:23: note: destination object referenced by ‘restrict’-qualified argument 1 was declared here
   81 |                 char *src = join_args(argv, j);
      |                       ^~~
[535/539] Compiling C object sway/sway.p/server.c.o
../sway/server.c: In function ‘server_init’:
../sway/server.c:217:50: warning: ‘%d’ directive writing between 1 and 11 bytes into a region of size 8 [-Wformat-overflow=]
  217 |                 sprintf(name_candidate, "wayland-%d", i);
      |                                                  ^~
../sway/server.c:217:41: note: directive argument in the range [-2147483647, 32]
  217 |                 sprintf(name_candidate, "wayland-%d", i);
      |                                         ^~~~~~~~~~~~
../sway/server.c:217:17: note: ‘sprintf’ output between 10 and 20 bytes into a destination of size 16
  217 |                 sprintf(name_candidate, "wayland-%d", i);
[...]

PR #6802, opened 2022-01-31
Fix snprintf compiler warnings (errors)

mmhobi7:
https://github.com/swaywm/sway/issues/6784
Hello this replaces sprintf with snprintf to avoid
`../sway/server.c:217:50: warning: ‘%d’ directive writing between 1 and 11 bytes into a region of size 8 [-Wformat-overflow=]
`
and
`../sway/commands/output/background.c:115:25: warning: ‘sprintf’ argument 4 overlaps destination object ‘src’ [-Wrestrict]
`

### DIFF
--- a/sway/commands/output/background.c
+++ b/sway/commands/output/background.c
@@ -112,7 +112,7 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 						"Unable to allocate resources");
 			}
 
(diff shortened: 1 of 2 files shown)


emersion (changes requested):
Please squash these two commits together.

emersion (approved):
Thanks!

PR #6810, opened 2022-02-03
commands/resize: make resize consider all siblings

### LINKED ISSUE #5936: Resizing a window in a multi-window layout doesn't work like in i3
If I have (for example) 4 windows tiled horizontally, and I resize the rightmost window using resize mode (`$mod+r`), the behavior between i3 and sway differs.

In i3, if I increase the width of one window, the width of all other windows are evenly reduced to compensate.

In Sway, only the window immediately next to the window being resized gets its width reduced, all other windows are untouched.

I have a ultra-wide screen, and I usually start 5 chat applications horizontally. One of the chat applications is really just a list, so it doesn't need to be particularly wide, so I tend to reduce its width by about 50%.

In i3, I just select that window, reduce the width using resize mode, and afterwards all the other windows are equally wide. In Sway, the window(s) right next to the one I'm resizing end up being much larger than the rest.

timhallmann:
Fixes a compatibility issue with i3 where resizing works as described
hereafter:
> Direction can either be one of up, down, left or right. Or you can be
> less specific and use width or height, in which case i3 will take/give
> space from all the other containers.

Sway previously considered only the direct neighbours not all siblings.

Fixes #5936

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -80,61 +80,46 @@ void container_resize_tiled(struct sway_container *con,
 	}
 
 	// For HORIZONTAL or VERTICAL, we are growing in two directions so select
-	// both adjacent siblings. For RIGHT or DOWN, just select the next sibling.
-	// For LEFT or UP, convert it to a RIGHT or DOWN resize and reassign con to
-	// the previous sibling.
-	struct sway_container *prev = NULL;
-	struct sway_container *next = NULL;
+	// all adjacent siblings. For RIGHT or DOWN or LEFT or UP select just the
+	// previous or next sibling.
+	list_t *resize = create_list();
 	list_t *siblings = container_get_siblings(con);
 	int index = container_sibling_index(con);
 
 	if (axis == AXIS_HORIZONTAL || axis == AXIS_VERTICAL) {
-		if (index == 0) {
-			next = siblings->items[1];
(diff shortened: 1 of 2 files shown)


Ochiruko:
It's been three years. Has this been fixed elsewhere? If not, are any reviewers working on this?

timhallmann:
Seems unlikely.

While I wouldn't mind getting this merged, I do not care enough to further bother with finding a reviewer.

sway/commands/resize.c
```diff
@@ -76,105 +76,87 @@ void container_resize_tiled(struct sway_container *con,
...
 	if (is_horizontal(axis)) {
-		if (con->pending.width + amount < MIN_SANE_W) {
-			return;
-		}
-		if (next->pending.width - sibling_amount < MIN_SANE_W) {
-			return;
-		}
-		if (prev && prev->pending.width - sibling_amount < MIN_SANE_W) {
-			return;
+		for(int i = 0; i < resize->length; i++) {
```
emersion, line 106: Style nit: missing space between `for` and `(`

(Applies to other lines of code below as well.)

emersion (commented):
LGTM apart from the style issues.

emersion (approved):
Thank you!

timhallmann:
Thanks for merging! :)

PR #6811, opened 2022-02-03
input/seat: emit focus events when focusing parent

timhallmann:
Emit a focus event when a container or the entire workspace gets focused
(for example, when focus_parent is called).

Previously, focus events were only emitted when another view or
workspace was focused. This made it impossible to accurately follow the
focus by monitoring focus events.

This is a deviation from i3s behaviour but I'd consider that a bug as well.

I'm not sure if this change doesn't have some side effects or introduces bugs that I haven't noticed yet.

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1198,8 +1198,10 @@ void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
 
 	// emit ipc events
 	set_workspace(seat, new_workspace);
-	if (container && container->view) {
+	if (container) {
 		ipc_event_window(container, "focus");
+	} else if(last_workspace == new_workspace) {
+		ipc_event_workspace(NULL, new_workspace, "focus");
 	}
 
 	// Move sticky containers to new workspace


rpigott (commented):
I think I could be convinced it would be helpful if focus events were emitted in this case. The important question is how do you see those con focus events being used? The current focus event emits the container id, so it is possible to track the focus whenever a window really is focused. What do you need the other focus events for?

As I understand it, in i3 the focus event tracks the X server's focus. So it doesn't emit when a container/workspace is focused because the X server has no focus. Clearly this doesn't apply to sway, but I'm not certain if subtly changing the event's meaning in this way would have a negative effect on compatibility. E.g. the window focus event also contains app_id / window_properties which a script might expect to be present.

It may be more sensible to extend the ipc with another event type that emits such focus changes, and users that want to listen to both types could subscribe to both "window" events and "con" events, for example. That requires a bit more thought though and would probably be worth discussing with i3.

timhallmann:
> I think I could be convinced it would be helpful if focus events were emitted in this case. The important question is how do you see those con focus events being used? The current focus event emits the container id, so it is possible to track the focus whenever a window really is focused. What do you need the other focus events for?

I knew I forgot to mention something: My motivation is to get accurate window titles with waybar (window titles being the entire reason I use waybar over swaybar). waybar currently keeps the title of the last focused window which means I can't rely on the title to tell me which window (or container) is affected when I execute a sway command.

> As I understand it, in i3 the focus event tracks the X server's focus. So it doesn't emit when a container/workspace is focused because the X server has no focus.

Right, that explains why i3 doesn't emit a focus event. Then, however, containers and workspaces in i3 shouldn't have a focused property at all, if focus means solely the X server's focus.

> Clearly this doesn't apply to sway, but I'm not certain if subtly changing the event's meaning in this way would have a negative effect on compatibility. E.g. the window focus event also contains app_id / window_properties which a script might expect to be present.

The manual currently states that a window event has the affected view in the container property. In that sense, yep, this change would break compatibility. However, the current example has an app_id and name of null (okay, it's for a "new" view event) and other event types do emit containers (at least: fullscreen_mode, floating, mark).

Waybar, the scripts in contrib and the scripts linked in the wiki check whether these properties exist and have a value.
 
> It may be more sensible to extend the ipc with another event type that emits such focus changes, and users that want to listen to both types could subscribe to both "window" events and "con" events, for example. That requires a bit more thought though and would probably be worth discussing with i3.

The problem is that these events are already mixed. Either some "con" events remain "window" events (and possibly also emit "con" events) or backwards compatibility would have to be broken too.

It might be worth to update the manpage though, given that some events implementation already differs from their description.

Issue #4523, opened 2019-08-31
focus_on_window_activation doesn't work for Wayland applications


jonleighton:
* Sway Version: 1.2

https://github.com/swaywm/sway/pull/2560 implemented support for the `focus_on_window_activation` configuration option.

Unfortunately, it only seems to work for applications which are running under XWayland.

I ran Firefox via XWayland, and ran `firefox http://github.com` from a terminal, and the Firefox window was focused.

I then ran Firefox via Wayland, and ran `firefox http://github.com`, and the Firefox window was not focused.

jonleighton:
cc @RyanDwyer

RyanDwyer:
As far as I'm aware, Wayland doesn't have any way for clients to request window activation. So this feature is Xwayland only until a Wayland protocol is agreed on.

jonleighton:
Thanks. This kind of thing works under GNOME Shell (i.e. clicking a link in terminal raises the Firefox window) so I wonder how they do it there...

elcste:
Does the [recent xdg-activation-v1 implementation](https://github.com/swaywm/sway/pull/6132) make this possible?

emersion:
To an extent yes. After Sway support is added, Firefox/GTK need to be updated as well.

mdouglass:
I think https://bugzilla.mozilla.org/show_bug.cgi?id=1744002 indicates that support for xdg_activation_v1 was added to Firefox 96, which I am using, but it is still not working.

emersion:
> I think https://bugzilla.mozilla.org/show_bug.cgi?id=1744002 indicates that support for xdg_activation_v1 was added to Firefox 96, which I am using, but it is still not working.

That's not enough, the terminal needs to set the `XDG_ACTIVATION_TOKEN` env variable, and Firefox needs to parse it.

emersion:
Closing since Sway already implements everything needed for this to work.

elcste:
@emersion Should the [man page](https://github.com/swaywm/sway/blob/master/sway/sway.5.scd?plain=1#L623) be changed not to specify xwayland clients for `focus_on_window_activation`?

emersion:
Ah, yes, we should update the man page. Would you be willing to send a patch?

AdamantGarth:
> > I think https://bugzilla.mozilla.org/show_bug.cgi?id=1744002 indicates that support for xdg_activation_v1 was added to Firefox 96, which I am using, but it is still not working.
> 
[...]

PR #6816, opened 2022-02-05
Minor update to focus_on_window_activation

elcste:
Removed xwayland limitation since wayland clients are supported via xdg-activation. See https://github.com/swaywm/sway/issues/4523#issuecomment-1026194317.

(The only real change is removing "n xwayland", but I re-wrapped the block so the diff is messy. If there's a better way, let me know.)


emersion (approved):
Thanks!

Issue #6813, opened 2022-02-03
xdg-shell: use wlr_xdg_toplevel in views instead of wlr_xdg_surface


emersion:
https://github.com/swaywm/sway/pull/6664 has `s/surface/surface->toplevel/`. Ideally we'd store a `wlr_xdg_toplevel` in `sway_view` instead of storing a `wlr_xdg_surface`.

Issue #6817, opened 2022-02-05
Crash in xdg_shell.c/is_transient_for


David96:
- **Sway Version:**
  - master (30d27b59)
- **Wlroots Version:**
  - master (c1a2c09a)

- **Stack Trace:**
<details>

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  is_transient_for (child=0x5597d3526ff0, ancestor=0x5597d3526ff0) at ../sway/desktop/xdg_shell.c:226
226		struct wlr_xdg_toplevel *toplevel = child->wlr_xdg_surface->toplevel;
[Current thread is 1 (Thread 0x7fdab24b7980 (LWP 3786))]
(gdb) bt full
#0  is_transient_for (child=0x5597d3526ff0, ancestor=0x5597d3526ff0) at ../sway/desktop/xdg_shell.c:226
        toplevel = 0x2000000020
#1  0x00005597d16e64fc in view_is_transient_for (child=0x5597d3526ff0, ancestor=0x5597d3526ff0) at ../sway/tree/view.c:1428
#2  0x00005597d16e01be in container_is_transient_for (child=0x5597d3520050, ancestor=0x5597d3520050) at ../sway/tree/container.c:1622
#3  0x00005597d169eea9 in output_render (output=0x5597d2ed0930, when=0x7ffd0737e290, damage=0x7ffd0737e2a0) at ../sway/desktop/render.c:1088
        floater = 0x5597d3520050
        i = 0
        clear_color = {0, 0, 0, 1}
        nrects = 2
        rects = 0x5597d34cf340
        wlr_output = 0x5597d2ef3910
        renderer = 0x5597d24858b0
        workspace = 0x5597d3030210
        fullscreen_con = 0x5597d3520050
        seat = 0x7ffd073ae9fa <clock_gettime+90>
        focus = 0x5597d34db850
        width = 121103008
        height = 32765
        frame_damage = {extents = {x1 = 2, y1 = 32765, x2 = 121103632, y2 = 32765}, data = 0x0}
        transform = 32730
#4  0x00005597d169a325 in output_repaint_timer_handler (data=0x5597d2ed0930) at ../sway/desktop/output.c:545
        now = {tv_sec = 2082, tv_nsec = 6772452}
        output = 0x5597d2ed0930
        workspace = 0x5597d3030210
        fullscreen_con = 0x5597d3520050
        needs_frame = true
[...]

PR #6822, opened 2022-02-07
xdg-shell: use wlr_xdg_{toplevel,popup}

emersion:
Improved type safety.

Closes: https://github.com/swaywm/sway/issues/6813

cc @vyivel


vyivel (approved):
LGTM; it's much cleaner now, and storing the underlying object directly helps to avoid accidental NULL dereference when working with "inert" views, which should fix #6817.
Thanks!

PR #6823, opened 2022-02-08
Remove all sprintf calls

emersion:
Replace them with snprintf, which ensures buffer overflows won't
happen.


bl4ckb0ne (approved):
LGTM, thanks :)

PR #6825, opened 2022-02-09
input/libinput: make udev optional

illiliti:
See also https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3490

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -64,7 +64,7 @@ libinput = dependency('libinput', version: '>=1.6.0')
 xcb = dependency('xcb', required: get_option('xwayland'))
 drm_full = dependency('libdrm') # only needed for drm_fourcc.h
 drm = drm_full.partial_dependency(compile_args: true, includes: true)
-libudev = dependency('libudev')
+libudev = dependency('libudev', required: get_option('udev'))
 bash_comp = dependency('bash-completion', required: false)
 fish_comp = dependency('fish', required: false)
 math = cc.find_library('m')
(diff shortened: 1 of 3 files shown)


emersion:
We can probably make this depend on `wlroots.get_variable('have_session') == 'true'` as a first step now that [wlroots!3874](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3874) is merged.

emersion:
Actually since this is only used in the libinput backend, `have_libinput_backend` would be more appropriate.

illiliti:
What exactly should depend on `have_libinput_backend`? `dependency('libudev')`?

emersion:
Yeah

illiliti:
So udev will be assumed dependency if libinput is present, right? It doesn't make sense to me since I patched out udev dependency from libinput in my branch.

By the way, isn't sway has mandatory dependency on libinput and session backends?

emersion:
> So udev will be assumed dependency if libinput is present, right? It doesn't make sense to me since I patched out udev dependency from libinput in my branch.

It would be a first step.

> By the way, isn't sway has mandatory dependency on libinput and session backends?

Nope

illiliti:
The better first step would be cooperating with libinput to implement `sway_libinput_device_is_builtin` there. Then udev would just go away from sway.

But anyway, feel free to open another PR and do your own thing. I'll keep this PR as is since I don't want to make udev mandatory with libinput backend enabled.


> > By the way, isn't sway has mandatory dependency on libinput and session backends?
> 
> Nope

I see at least three `wlr/backend/session.h` header includes. As of libinput, it is literally everywhere.

PR #6835, opened 2022-02-15
Fix behaviour of container_move_to_container, add move to con_id

### LINKED ISSUE #6818: `container_move_to_container` doesn't behave like it's i3 equivalent
**Sway Version:** 1.7

**Logs:** [here](https://pastebin.com/pb6JcBbq)

I'm migrating here from i3 because X just slowly descended into graphical madness, I have a few keybindings which call a script. The relevant one to this is:
```bash
parent=$(nodePath | jq -rcM '.[-3] | if type == "null" then empty else .id end')
if [ "$parent" != '' ]; then
  swaymsg "[con_id=$parent] mark _parent; move window to mark _parent; unmark _parent"
fi
```
Where `nodePath` makes an array of each container down to the focused node, so `[-3]` of that is the "grandparent" (parent's parent). So it takes a window, and then makes it a child of it's parent's parent, thus moving it up the tree one level. This works fine on i3, and does absolutely nothing when used in sway. There's no error, the log says it's running the command, but nothing happens. When I go into a console and try to dispatch that set of commands manually with `swaymsg`, it also doesn't do anything. The issue doesn't appear to lie further down in the tree fetching or anything, because when inspecting the command as it runs the value of `$parent` is consistent and correct, the subsequent commands just aren't doing anything. (Well, *sometimes* it will actually place the mark, which can be seen by removing the unmark bit, but even then this extremely is inconsistent)

And, while we're here, when I try to move containers to marks in other ways, it's seemingly random and sporadic whether it will work or not. I think it might be failing to do anything whenever the marked container is a direct ancestor of the focused one, but I'm not entirely certain on that. But also, when it does work, the container being moved always loses focus, which is not how the same command behaves in i3.

Myrdden:
Closes #6818 
Okay so this PR does two things:

One, it removes this line
https://github.com/swaywm/sway/blob/f707f583e17cb5e8323ceb4bfd951ad0465b7d10/sway/commands/move.c#L235
In order to match the behaviour of i3's move functionality, which can be found [here](https://github.com/i3/i3/blob/c822eff1bfdeac30baf92b14a09d1c81a0193997/src/move.c#L72-L79), posted below:
```c
    /* We compare the focus order of the children of the lowest common ancestor. If con or
     * its ancestor is before target's ancestor then con should be placed before the target
     * in the focus stack. */
    Con *lca = lowest_common_ancestor(con, parent);
    if (lca == con) {
        ELOG("Container is being inserted into one of its descendants.\n");
        return;
    }
```
Basically, sway is checking on move that a destination is neither a descendant of the container to be moved, *and* that the container to be moved is not a descendant of the destination, whereas i3 only (rightly, I think) checks that the destination is not a descendant, as such a lot of valid moves in i3 simply don't work in sway and this fixes that.

The second, sort of incidental thing this PR adds is the command `move [container|window] [to] con_id <con_id>`. I use a script with a lot of "macros" to make navigating formerly-i3-and-now-sway easier, and I noticed that in decently upwards of half of them they all have the same repeated behaviour:
```
[con_id=$someDestination] mark _destination; move container to mark _destination; unmark _destination;
```
or something similar, so it seemed like there was a very obvious use case to just have that be
```
move container to con_id $someDestination
```
All of the rules surrounding whether a move is valid and how the move functions are identical to move-to-mark, this just saves some steps if you happen to have a con_id handy.

So yeah! Let me know what you think.

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -373,4 +373,9 @@ bool container_is_sticky_or_child(struct sway_container *con);
  */
 int container_squash(struct sway_container *con);
 
+bool container_has_con_id(struct sway_container *con, size_t *con_id);
+
+/** Returns a container with the given con_id */
+struct sway_container *container_find_con_id(size_t *con_id);
+
 #endif
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -25,7 +25,8 @@ static const char expected_syntax[] =
 	"Expected 'move <left|right|up|down> <[px] px>' or "
 	"'move [--no-auto-back-and-forth] <container|window> [to] workspace <name>' or "
 	"'move <container|window|workspace> [to] output <name|direction>' or "
-	"'move <container|window> [to] mark <mark>'";
+	"'move <container|window> [to] mark <mark>' or "
+	"'move <container|window> [to] con_id <con_id>'";
 
 static struct sway_output *output_in_direction(const char *direction_string,
 		struct sway_output *reference, int ref_lx, int ref_ly) {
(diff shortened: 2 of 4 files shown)


Myrdden:
Upon further inspection of i3's codebase it looks like it's implementing additional logic that isn't present here that also possibly solves the focus issues I was seeing in #6818, so simply removing the conditional is likely not sufficient to actually get this to work as expected. I suppose I should read into this more...

korreman:
Thanks for taking the time to work out how this works in both sway and i3. With the general reluctance towards adding new features to sway, it's probably best to give the new command its own PR. Matching the behavior of an existing command to that of i3 should be easier to merge. The most important thing is to have _some_ consistent way of moving containers to arbitrary positions in the tree.

PR #6839, opened 2022-02-19
output: update xcursor configuration on scale change

Nefsen402:
Sway will not correctly update the cursor glyph cache on an output
scale change. This results in sway falling back to a lower resolution
cursor glyph or just not rendering the correct cursor at all.

This is reproducible simply by using `swaymsg` to change
the scale sometimes, but I found 100% reproducibility when using
these patches[1].

[1] https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3418

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -824,6 +824,13 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		transaction_commit_dirty();
 
 		update_output_manager_config(output->server);
+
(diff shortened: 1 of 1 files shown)


emersion:
Hm, IIRC we already have output config changes updating the input config elsewhere…

emersion:
I think this is now handled by `wlr_cursor`?

Nefsen402:
Yes, this is an old bug.

PR #6842, opened 2022-02-21
Don't try to resize a client smaller than its constraints

Nefsen402:
Many clients will not ack & commit configures that make them smaller
or larger than their specified constraints. This means that the
transaction will timeout and poor perceived performance will endure.

Before:
https://user-images.githubusercontent.com/3621017/154896925-868b8713-8d92-4ed8-9586-3c965bce01a5.mp4

After:
https://user-images.githubusercontent.com/3621017/154896935-205b800f-5e27-485a-8b45-5ff223d48fd1.mp4

Is this the right solution to the problem?

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -149,6 +149,13 @@ static bool configure(struct sway_view *view, uint32_t *serial,
 		return false;
 	}
 
+	// avoid same size configures as clients may not ack & commit them
+	struct wlr_xdg_toplevel *top = view->wlr_xdg_toplevel;
+	if ((int) top->pending.width == width &&
+			(int) top->pending.height == height) {
+		return false;
+	}
+
 	*serial = wlr_xdg_toplevel_set_size(top, width, height);
 	return true;
 }
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -141,15 +141,16 @@ static const char *get_string_prop(struct sway_view *view,
 	}
 }
 
-static uint32_t configure(struct sway_view *view, double lx, double ly,
-		int width, int height) {
+static bool configure(struct sway_view *view, uint32_t *serial,
+		double lx, double ly, int width, int height) {
 	struct sway_xdg_shell_view *xdg_shell_view =
 		xdg_shell_view_from_view(view);
 	if (xdg_shell_view == NULL) {
-		return 0;
+		return false;
 	}
-	return wlr_xdg_toplevel_set_size(view->wlr_xdg_toplevel,
-		width, height);
+
+	*serial = wlr_xdg_toplevel_set_size(top, width, height);
+	return true;
 }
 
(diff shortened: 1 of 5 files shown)


Xyene:
> Many clients will not ack & commit configures that make them smaller or larger than their specified constraints. 

Not acking these requests is a client bug. The documentation of `xdg_toplevel::set_min_size` says:

> The client can specify a minimum size so that the compositor does not try to configure the window below this size.
> ...
> The client should not rely on the compositor to obey the minimum size. **The compositor may decide to ignore the values set by the client and request a smaller size.**

(Emphasis mine.)

Same for not acking configures with the same size, but it's reasonable from an efficiency perspective to not send those in the first place.

Client bugs notwithstanding, I don't see an issue with clipping surface contents like you've done here. Even if the client were to try to render below its desired size it'd probably look bad, and empirically apps don't even try, so might as well clip them.

> Also fix a bug where the transaction code will still wait for a client commit even though the instruction serial is 0 which is used to represent a unsuccessful configure

Could you split this up into a separate commit?

sway/tree/view.c
```diff
@@ -166,6 +167,17 @@ void view_get_constraints(struct sway_view *view, double *min_width,
 
 uint32_t view_configure(struct sway_view *view, double lx, double ly, int width,
 		int height) {
+	// Many clients don't play nicely if you try to reconfigure them out of
+	// their specified constraits. Clamp the values.
+	{
```
Xyene, line 172: Nit: don't nest this in it's own scope. Also, `s/constraits/constraints/`

emersion:
> Also fix a bug where the transaction code will still wait for a client commit even though the instruction serial is 0 which is used to represent a unsuccessful configure

0 is not used to indicate failures, serials will wrap-around once UINT32_MAX is reached.

sway/desktop/xdg_shell.c
```diff
@@ -148,7 +148,15 @@ static uint32_t configure(struct sway_view *view, double lx, double ly,
 	if (xdg_shell_view == NULL) {
 		return 0;
```
Nefsen402, line 149: @emersion If 0 does not mean an unsuccessful configure, then this code here is wrong. Should we give 0 a special meaning and document that or should we add a new `bool check_valid_configure()` function to check invalid conditions?

Nefsen402:
I redid the patch with a commit that fixes the ambiguity with a unsuccessful configure versus a 0 serial.

Xyene (approved):
This change LGTM, thanks! Will leave the merge up to @emersion in case he has other thoughts on this patch.

marcelohdez:
Gentle ping;

On my 13 inch laptop using Sway 1.9, when spawning large windows like Discord or Steam into an existing workspace there is a noticeable freeze. Then, when resizing to make them fit there is again a noticeable delay before each resize as seen in the **before** video, making Sway feel sluggish and this PR would at least mitigate these issues. Thank you all for your work!

PR #6854, opened 2022-02-25
swaynag: die on all allocation failures

noocsharp:
Ideally the number of allocations should be reduced, but until then, detecting failure is better than not.

### DIFF
--- a/swaynag/config.c
+++ b/swaynag/config.c
@@ -19,6 +19,10 @@ static char *read_from_stdin(void) {
 	ssize_t nread;
 	while ((nread = getline(&line, &line_size, stdin)) != -1) {
 		buffer = realloc(buffer, buffer_len + nread + 1);
+		if (!buffer) {
+			perror("realloc");
+			return NULL;
+		}
 		snprintf(&buffer[buffer_len], nread + 1, "%s", line);
 		buffer_len += nread;
 	}
@@ -152,6 +156,10 @@ int swaynag_parse_options(int argc, char **argv, struct swaynag *swaynag,
 				}
 				struct swaynag_button *button;
 				button = calloc(sizeof(struct swaynag_button), 1);
+				if (!button) {
+					perror("calloc");
+					return EXIT_FAILURE;
+				}
 				button->text = strdup(optarg);
 				button->type = SWAYNAG_ACTION_COMMAND;
 				button->action = strdup(argv[optind]);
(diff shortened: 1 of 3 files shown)


swaynag/config.c
```diff
@@ -19,6 +19,10 @@ static char *read_from_stdin(void) {
 	ssize_t nread;
 	while ((nread = getline(&line, &line_size, stdin)) != -1) {
 		buffer = realloc(buffer, buffer_len + nread + 1);
+		if (!buffer) {
+			perror("realloc:");
```
emersion, line 23: `perror` already prints a semicolon character.

emersion:
I wonder if that's really better than crashing with SIGSEGV.

noocsharp:
Deferencing a null pointer is undefined behavior. I think for the sake of reliability and robustness, relying on undefined behavior is a bad idea.

swaynag/swaynag.c
```diff
@@ -357,6 +361,9 @@ static void handle_global(void *data, struct wl_registry *registry,
 		if (!swaynag->output) {
 			struct swaynag_output *output =
 				calloc(1, sizeof(struct swaynag_output));
+			if (!output) {
+				return;
```
bl4ckb0ne, line 365: There should be an error message there

bl4ckb0ne (commented):
> Ideally the number of allocations should be reduced

Why not do that instead of checking every alloc? There's a few place where a static buffer with a big size could do the trick.

noocsharp:
> > Ideally the number of allocations should be reduced
> 
> Why not do that instead of checking every alloc? There's a few place where a static buffer with a big size could do the trick.

I think reducing the number of unnecessary allocations is for a separate pull request. For now, this is an easy way to improve correctness.

bl4ckb0ne:
> I think reducing the number of unnecessary allocations is for a separate pull request

2 commits instead of 1 is more work.

noocsharp:
I looked a bit closer, and I see two cases of potentially unnecessary allocations, but I don't see a way to remove them without more involved refactoring (not sure if its possible even with that). Also, I think it's best to avoid using a really large static buffer unless it can be guaranteed that the data will fit. The size of the buffers seem to be determined at runtime, so using a static buffer doesn't make much sense.

kennylevinsen:
As our policy is to handle allocation errors, I don't see why we should have a problem with this change. I also have no problem with changes from heap to stack being follow-up PRs.

However, many allocations are missed by this change, as e.g. `strdup(3)` also uses `malloc(3)`.

bl4ckb0ne (approved):
LGTM, thanks!

Issue #6861, opened 2022-03-02
sway_switch is never destroyed


bl4ckb0ne:
```
Direct leak of 280 byte(s) in 7 object(s) allocated from:
    #0 0x7f18aa588fb9 in __interceptor_calloc /usr/src/debug/gcc/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x5562660d724c in sway_switch_create ../sway/sway/input/switch.c:9
    #2 0x5562660d724c in seat_configure_switch ../sway/sway/input/seat.c:815
    #3 0x5562660d724c in seat_configure_device ../sway/sway/input/seat.c:881
```

`sway_switch` devices are created but never destroyed on teardown.

PR #6870, opened 2022-03-05
sway/input: destroy sway_switch properly

OctopusET:
Fix: #6861
There is sway_switch_destroy function in `input/switch.c`, but never used.
So, I added that function to seat_device_destroy function in `input/seat.c`

I've checked with ASAN and leak has gone.


emersion (approved):
Thanks!

PR #6871, opened 2022-03-06
input/cursor: don't allow resizing tiling fullscreen views

Xyene:
This check already exists for floating fullscreen views, just copying it
over. The views never truly got resized, but the cursor image would
erroneously change to the resize cursor.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -399,7 +399,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 
 	// Handle tiling resize via mod
 	bool mod_pressed = modifiers & config->floating_mod;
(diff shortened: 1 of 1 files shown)


emersion:
Does this prevent resizing a child of a fullscreen container, for instance a view in a fullscreen splith?

Xyene:
Actually... good catch, yes. But we already do just that for fullscreen splits that were originally floating, so there's inconsistency there. Converting to draft for now.

I think the behavior we want is probably more like:

* if a view is fullscreen and has no visible siblings, disallow resize, otherwise allow the resize (regardless of if it's a floating or tiling fullscreen)
* if a view is tiled and has no visible siblings, disallow resize
* other resize cases remain unchanged

What do you think?

emersion:
That sounds good to me.

PR #6484, opened 2021-09-04
Prevent segfault on keyboard connection

Yaroslav-95:
Sway was crashing on me everytime I connected or reconnected my Thinkpad usb keyboard. It turns out that for some reason, upon connection, the pointer to the keymap was NULL, therefore crashing Sway as it attempted to access the pointer in ipc_json_describe_input. I've added a check for that.

So far I've tested it and it doesn't crash on me anymore when connecting my keyboard. `swaymsg -t get_inputs` still shows the correct layouts for the keyboard, so I guess it sets the keymap pointer in a later event.

PR #6873, opened 2022-03-07
ipc-json: Fix NULL ptr dereference if layout NULL

jameswalmsley:


emersion:
Hm, why would `keyboard->keymap` be NULL?

jameswalmsley:
> Hm, why would `keyboard->keymap` be NULL?

Thats a good question :)
I'm trying to debug it at the moment.

This is if I unplug a keyboard and then plug it back in again.

I think the libinput_config event its being called too soon. 

Do you have any pointers?

jameswalmsley:
FYI this has been a problem in sway 1.6 and 1.7.
1.5 doesn't have this issue.

I havent had chance to bisect with more precision yet.

jameswalmsley:
Some debug logging:

```
00:00:09.970 [sway/ipc-server.c:505] Sending input event: removed
00:00:09.971 [sway/ipc-server.c:505] Sending input event: removed
00:00:10.227 [sway/ipc-server.c:505] Sending input event: removed
00:00:10.228 [sway/ipc-server.c:505] Sending input event: removed

00:00:13.445 [sway/ipc-server.c:505] Sending input event: libinput_config

00:00:13.446 [sway/ipc-json.c:1010] The keymap is NULL!!!

00:00:13.477 [sway/ipc-server.c:505] Sending input event: xkb_keymap
00:00:13.477 [sway/ipc-server.c:505] Sending input event: added
00:00:13.477 [sway/ipc-server.c:505] Sending input event: added

00:00:13.504 [sway/ipc-server.c:505] Sending input event: xkb_keymap
00:00:13.504 [sway/ipc-server.c:505] Sending input event: added
00:00:13.504 [sway/ipc-server.c:505] Sending input event: added
```

Looks like libinput_config event is sent before the added event.

jameswalmsley:
I've now moved

```
sway_input_configure_libinput_device(input_device);
```

To after the added event. This fixes the issue without the NULL pointer check. 

@emersion Would this break anything else?

bl4ckb0ne:
Devices should be properly configured when the backend emits the `new_input` event. There was recent changes in wlroots' input device code, have you tried with sway/wlroots master if the issue is still present?

jameswalmsley:
> Devices should be properly configured when the backend emits the `new_input` event. There was recent changes in wlroots' input device code, have you tried with sway/wlroots master if the issue is still present?

I saw the commits relating to this, and I've pulled the latest master this evening.

The problem still exists.

jameswalmsley:
@bl4ckb0ne Thanks for the information, I've moved the call to the first position where the problem goes away.
So that the added event comes when the device is completely configured.

emersion:
I don't understand why this fixes the bug -- the `added` event is emitted at the end of the function even with the old code…

jameswalmsley:
> I don't understand why this fixes the bug -- the `added` event is emitted at the end of the function even with the old code…

Hi emersion, I've switched my keyboard to bluetooth mode and the problem goes away, so this seems to be something for physical USB keyboards.

I was digging through the code for a few hours trying to understand why it was NULL but didn't quite get to a complete understanding of whats happening.

I'll try to have  another debug session to work out why this happens and why the patch fixes the problem.

I'll also try to get a minimum reproducer available.

jameswalmsley:
@emersion I think I have more of an understanding whats happening now.:

`sway_input_configure_libinput_device()` can call `ipc_event_input("libinput_config")` if the device has changed libinput config.
In my case it seems to have done so.

This in turn causes `ipc_json_describe_input(device)` to be called at this point. However keymap is NULL, so it causes the segfault.

If instead we call `sway_input_configure_libinput_device()` after the seat configuration, then then keymap won't be NULL.

The reason for this is that `seat_configure_keyboard()` gets called which in turn calls `sway_keyboard_configure()` which does the keymap compiling and setting the keymap pointer.

Let me know what you think.

jameswalmsley:
Some further information:

I've reduced this to the example configuration (which alone does not cause the issue).

The following conditions need to be present to trigger the issue:

  1. waybar configured with the sway/languages module.
  2. Logitech unified dongle, with a mouse and keyboard paired. (Pairing only the keyboard does not trigger the issue).
  3. sway config with swaybar_command waybar
  4. sway config with an input configuration for the paired mouse e.g:
  
  ```
  input 1133:16458:Logitech_MX_Anywhere_2 {
      accel_profile flat
      pointer_accel 0
      natural_scroll disable
  }
  ```

jameswalmsley:
Also just to confirm, I've tested this configuration on the latest master for wlroots / sway.

jameswalmsley:
@bl4ckb0ne I've managed to minimise the test-case as describe above, was wondering if you have any further feedback?

emersion:
I think we shouldn't emit a `libinput_config` event for devices which haven't gone through the `added` event.

Yaroslav-95:
> I think we shouldn't emit a `libinput_config` event for devices which haven't gone through the `added` event.

So basically `sway_input_configure_libinput_device` which is called when changing input configuration and on new device conections sends the `libinput_config` event which, in the case of a new device, should be sent after `added`. In that case a possible solution would be to have `sway_input_configure_libinput_device` just return a bool value indicating whether configuration has changed and so let the caller emit the event. What do you think @emersion?

emersion:
Yeah, sounds good.

Yaroslav-95:
> Yeah, sounds good.

Done here #6484

emersion:
Superseded by #6484

PR #6844, opened 2022-02-21
Use wlroots  scene graph api

Nefsen402:
Closes: #6562 #7120 #6914 #6821 #6960 #6148 #5867 #5497 #5460 #6469 #7458 #7463
Supersedes: #6820 #6747 

This should be enough where if most people were on this patch, they wouldn't know it. However, there are some features still missing from the port:  
- [x] Opacity support
     Scene graph doesn't have support for specifying the opacity of a `wlr_scene_surface`.
     However, implementing this would not be enough because we send in xdg surfaces
     into relevant helper functions that hide the surfaces / subsurfaces away from us. The easiest
     fix from the perspective of the user is the ability to specify the opacity of an entire subtree.
- [x] Surface scissor
      We're also missing support for specifying a scissor region for a surface node. This is useful
      because it will prevent applications that don't respect a configure from possibly rendering
      on top of other tiled applications. Again, the actual surface nodes are hidden from us, so it
      would be easiest to specify a scissor region for an entire subtree.
- [x] Buffer node output awareness
       Scenes already handle client output awareness for us (`wlr_surface_send_enter`, `wlr_surface_send_leave`)
       However, we don't have this information for buffer nodes so we can do more optimized
       text rending. For now text rendering is hard coded to a 2x screen scale and grey scale sub
       pixel handling. An appropriate solution to this problem is to support `wl_signal`s for
       output enter/exit events and the ability to overwrite the buffer in a `wlr_scene_buffer`.
- [x] scale_filter option
       Scenes do linear filtering only. We need some way to specify this.
- [x] Upstream issue for gamma correction during direct scanout: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3609

Wayland backend specific bugs (not regressions of this branch):
- Xwayland damage tracking issues, especially noticeable when entering/exiting full screen.
       This seems like a wlroots bug.
- Xwayland cropped cursor on HiDPI
       This is probably a bug I introduced but I'm not sure. This also happens sometimes in
       sway-git, maybe I messed the code up enough to make it 100% reproducible.
- Pointer lock doesn't work

Although there are xwayland damage tracking issues, this does not mean that current git is unaffected
by damage tracking issues. In fact, this branch fixes damage tracking issues with subsurfaces on a 
xdg_surface popup, so as far as I know, for wayland clients damage tracking is perfect.

PR #6874, opened 2022-03-08
arrange: Don't try to arrange a fullscreen container as a tiling/floating container

Nefsen402:
Causes the container to appear in random places in fullscreen view.
See: https://github.com/swaywm/sway/pull/6844#issuecomment-1061392867
See: https://github.com/swaywm/sway/pull/6844#issuecomment-1061691809

### DIFF
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -209,6 +209,34 @@ static void arrange_floating(list_t *floating) {
 
 static void arrange_children(list_t *children,
 		enum sway_container_layout layout, struct wlr_box *parent) {
+
+	// Don't include fullscreen containers. If we do, we run the risk
+	// of overwriting properties as if this container was tiled/floating
+	// but should still be fullscreen.
+	bool has_fullscreen = false;
+	for (int i = 0; i < children->length; ++i) {
+		struct sway_container *child = children->items[i];
+
+		if (child->pending.fullscreen_mode != FULLSCREEN_NONE) {
+			has_fullscreen = true;
+			break;
+		}
+	}
+
+	if (has_fullscreen) {
+		list_t *fs_purged = create_list();
+
+		for (int i = 0; i < children->length; ++i) {
+			struct sway_container *child = children->items[i];
+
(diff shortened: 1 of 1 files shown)


sway/tree/arrange.c
```diff
@@ -231,7 +231,13 @@ static void arrange_children(list_t *children,
 	// Recurse into child containers
 	for (int i = 0; i < children->length; ++i) {
 		struct sway_container *child = children->items[i];
-		arrange_container(child);
+
+		// If this container is fullscreen don't overwrite its properties
+		// here.
+		if (child->pending.workspace->fullscreen != child &&
+				root->fullscreen_global != child) {
```
emersion, line 238: Checking `child->pending.fullscreen_mode` is shorter.

emersion (commented):
We keep getting patches similar to this one, see for instance https://github.com/swaywm/sway/pull/6847 and https://github.com/swaywm/sway/pull/6679. It would be nice to find a more general solution.

Nefsen402:
> It would be nice to find a more general solution.

After #6844 is merged, I plan on making two more PRs cleaning up stuff I learned about while developing. Pr 2 is going to rewrite transactions. Pr 3 is going to focus on rewriting input stuff to use the scene-graph as the source of truth. This will remove all consumers of container->{current, pending}.{x, y, width, height} so I should be able to get rid of arrange.c all together. So that's my grand scheme vision that should get us to a system that is more "general". In fact #6844 already has the stepping stones for that as arranging the scene graph is done totally separately and I tried not to touch state that arrange.c generates as I know I want to move away from that.

Nefsen402:
I noticed this patch is wrong. `arrange_children` will still go through and overwrite pending.{x, y, width, height}. This patch just prevents it from further recusing committing that incorrect state into child views.

Nefsen402:
I fixed up the patch. This is no general solution and it's pretty nasty with its potential short lived heap allocation. It's the best I came up with.

Nefsen402:
Ping.

PR #6876, opened 2022-03-08
Remove WLR_SWITCH_STATE_TOGGLE usage

emersion:
Ref [1].

[1]: https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/4792446ee8f50104bd207d9ccd8558a7e4eb4514

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -70,12 +70,18 @@ struct sway_mouse_binding {
 	char *command;
 };
 
+enum sway_switch_trigger {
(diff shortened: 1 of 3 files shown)


bl4ckb0ne:
Do we really need the TOGGLE state?

emersion:
It's not a state, it's a trigger condition. It's used to execute a switch binding whenever its state changes (and not just when it becomes on, or just when it becomes off).

PR #6879, opened 2022-03-09
Implement ext-session-lock-v1

danieldg:
This prevents input from bypassing a lockscreen if the lockscreen crashes or has other problems (for example, https://github.com/swaywm/swaylock/issues/204).  Swaylock support is already present as of 1.6 (https://github.com/swaywm/swaylock/pull/219).

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -185,6 +185,10 @@ struct sway_workspace *seat_get_last_known_workspace(struct sway_seat *seat);
 
 struct sway_container *seat_get_focused_container(struct sway_seat *seat);
 
+// Force focus to a particular surface that is not part of the workspace
+// hierarchy (used for lockscreen)
+void sway_force_focus(struct wlr_surface *surface);
+
 /**
  * Return the last container to be focused for the seat (or the most recently
  * opened if no container has received focused) that is a child of the given
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -16,6 +16,7 @@
 #include <wlr/types/wlr_output_power_management_v1.h>
 #include <wlr/types/wlr_presentation_time.h>
 #include <wlr/types/wlr_relative_pointer_v1.h>
+#include <wlr/types/wlr_session_lock_v1.h>
 #include <wlr/types/wlr_server_decoration.h>
 #include <wlr/types/wlr_text_input_v3.h>
 #include <wlr/types/wlr_xdg_shell.h>
@@ -89,6 +90,19 @@ struct sway_server {
 	struct wl_listener output_manager_apply;
 	struct wl_listener output_manager_test;
 
+	struct {
+		bool locked;
+		struct wlr_session_lock_manager_v1 *manager;
+
+		struct wlr_session_lock_v1 *lock;
+		struct wl_listener lock_new_surface;
+		struct wl_listener lock_unlock;
+		struct wl_listener lock_destroy;
+
+		struct wl_listener new_lock;
+		struct wl_listener manager_destroy;
+	} session_lock;
+
 	struct wlr_output_power_manager_v1 *output_power_manager_v1;
 	struct wl_listener output_power_manager_set_mode;
 	struct wlr_input_method_manager_v2 *input_method;
@@ -148,6 +162,7 @@ void handle_new_output(struct wl_listener *listener, void *data);
 
 void handle_idle_inhibitor_v1(struct wl_listener *listener, void *data);
 void handle_layer_shell_surface(struct wl_listener *listener, void *data);
+void sway_session_lock_init(void);
 void handle_xdg_shell_surface(struct wl_listener *listener, void *data);
 #if HAVE_XWAYLAND
 void handle_xwayland_surface(struct wl_listener *listener, void *data);
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -275,6 +275,25 @@ static void for_each_surface_container_iterator(struct sway_container *con,
 
 static void output_for_each_surface(struct sway_output *output,
 		sway_surface_iterator_func_t iterator, void *user_data) {
+	if (server.session_lock.locked) {
+		if (server.session_lock.lock == NULL) {
+			return;
+		}
+		struct wlr_session_lock_surface_v1 *lock_surface;
+		wl_list_for_each(lock_surface, &server.session_lock.lock->surfaces, link) {
+			if (lock_surface->output != output->wlr_output) {
+				continue;
+			}
+			if (!lock_surface->mapped) {
+				continue;
+			}
+
+			output_surface_for_each_surface(output, lock_surface->surface,
+				0.0, 0.0, iterator, user_data);
+		}
+		return;
+	}
+
 	if (output_has_opaque_overlay_layer_surface(output)) {
 		goto overlay;
 	}
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -1056,6 +1056,41 @@ void output_render(struct sway_output *output, struct timespec *when,
 		wlr_renderer_clear(renderer, (float[]){1, 1, 0, 1});
 	}
 
+	if (server.session_lock.locked) {
+		float clear_color[] = {0.0f, 0.0f, 0.0f, 1.0f};
+		if (server.session_lock.lock == NULL) {
+			// abandoned lock -> red BG
+			clear_color[0] = 1.f;
+		}
+		int nrects;
+		pixman_box32_t *rects = pixman_region32_rectangles(damage, &nrects);
+		for (int i = 0; i < nrects; ++i) {
+			scissor_output(wlr_output, &rects[i]);
+			wlr_renderer_clear(renderer, clear_color);
+		}
+
+		if (server.session_lock.lock != NULL) {
+			struct render_data data = {
+				.damage = damage,
+				.alpha = 1.0f,
+			};
+
+			struct wlr_session_lock_surface_v1 *lock_surface;
+			wl_list_for_each(lock_surface, &server.session_lock.lock->surfaces, link) {
+				if (lock_surface->output != wlr_output) {
+					continue;
+				}
+				if (!lock_surface->mapped) {
+					continue;
+				}
+
+				output_surface_for_each_surface(output, lock_surface->surface,
+					0.0, 0.0, render_surface_iterator, &data);
+			}
+		}
+		goto renderer_end;
+	}
+
 	if (output_has_opaque_overlay_layer_surface(output)) {
 		goto render_overlay;
 	}
--- a/sway/lock.c
+++ b/sway/lock.c
@@ -1,0 +1,184 @@
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
+#include "log.h"
+#include "sway/input/keyboard.h"
+#include "sway/input/seat.h"
+#include "sway/output.h"
+#include "sway/server.h"
+
+struct sway_session_lock_surface {
+	struct wlr_session_lock_surface_v1 *lock_surface;
+	struct sway_output *output;
+	struct wlr_surface *surface;
+	struct wl_listener map;
+	struct wl_listener destroy;
+	struct wl_listener surface_commit;
+	struct wl_listener output_mode;
+	struct wl_listener output_commit;
+};
+
+static void handle_surface_map(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_surface *surf = wl_container_of(listener, surf, map);
+	sway_force_focus(surf->surface);
+	output_damage_whole(surf->output);
+}
+
+static void handle_surface_commit(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_surface *surf = wl_container_of(listener, surf, surface_commit);
+	output_damage_surface(surf->output, 0, 0, surf->surface, false);
+}
+
+static void handle_output_mode(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_surface *surf = wl_container_of(listener, surf, output_mode);
+	wlr_session_lock_surface_v1_configure(surf->lock_surface,
+		surf->output->width, surf->output->height);
+}
+
+static void handle_output_commit(struct wl_listener *listener, void *data) {
+	struct wlr_output_event_commit *event = data;
+	struct sway_session_lock_surface *surf = wl_container_of(listener, surf, output_commit);
+	if (event->committed & (
+			WLR_OUTPUT_STATE_MODE |
+			WLR_OUTPUT_STATE_SCALE |
+			WLR_OUTPUT_STATE_TRANSFORM)) {
+		wlr_session_lock_surface_v1_configure(surf->lock_surface,
+			surf->output->width, surf->output->height);
+	}
+}
+
+static void handle_surface_destroy(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_surface *surf = wl_container_of(listener, surf, destroy);
+	wl_list_remove(&surf->map.link);
+	wl_list_remove(&surf->destroy.link);
+	wl_list_remove(&surf->surface_commit.link);
+	wl_list_remove(&surf->output_mode.link);
+	wl_list_remove(&surf->output_commit.link);
+	output_damage_whole(surf->output);
+	free(surf);
+}
+
+static void handle_new_surface(struct wl_listener *listener, void *data) {
+	struct wlr_session_lock_surface_v1 *lock_surface = data;
+	struct sway_session_lock_surface *surf = calloc(1, sizeof(*surf));
+	if (surf == NULL) {
+		return;
+	}
+
+	sway_log(SWAY_DEBUG, "new lock layer surface");
+
+	struct sway_output *output = lock_surface->output->data;
+	wlr_session_lock_surface_v1_configure(lock_surface, output->width, output->height);
+
+	surf->lock_surface = lock_surface;
+	surf->surface = lock_surface->surface;
+	surf->output = output;
+	surf->map.notify = handle_surface_map;
+	wl_signal_add(&lock_surface->events.map, &surf->map);
+	surf->destroy.notify = handle_surface_destroy;
+	wl_signal_add(&lock_surface->events.destroy, &surf->destroy);
+	surf->surface_commit.notify = handle_surface_commit;
+	wl_signal_add(&surf->surface->events.commit, &surf->surface_commit);
+	surf->output_mode.notify = handle_output_mode;
+	wl_signal_add(&output->wlr_output->events.mode, &surf->output_mode);
+	surf->output_commit.notify = handle_output_commit;
+	wl_signal_add(&output->wlr_output->events.commit, &surf->output_commit);
+}
+
+static void handle_unlock(struct wl_listener *listener, void *data) {
+	sway_log(SWAY_DEBUG, "session unlocked");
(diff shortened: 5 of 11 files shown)


sway/lock.c
```diff
@@ -0,0 +1,174 @@
+#define _POSIX_C_SOURCE 200809L
+#include "log.h"
+#include "sway/input/keyboard.h"
+#include "sway/input/seat.h"
+#include "sway/lock.h"
+#include "sway/output.h"
+#include "sway/server.h"
+
+#define PERMALOCK_CLIENT (struct wl_client *)(-1)
```
emersion, line 9: Hm, really not a fan of this…

include/sway/input/seat.h
```diff
@@ -185,6 +185,10 @@ struct sway_workspace *seat_get_last_known_workspace(struct sway_seat *seat);
 
 struct sway_container *seat_get_focused_container(struct sway_seat *seat);
 
+// Force focus to a particular surface that is not part of the workspace
+// hierarchy (used for lockscreen)
+void sway_force_focus(struct wlr_surface *surface);
```
emersion, line 190: Can't we re-use the wlr-input-inhibitor implementation stuff for this?

include/sway/lock.h
```diff
@@ -0,0 +1,23 @@
+#ifndef _SWAY_LOCK_H
+#define _SWAY_LOCK_H
+#include <wlr/types/wlr_session_lock_v1.h>
+
+enum sway_lock_state {
+	SESSIONLOCK_STATE_UNLOCKED,
```
emersion, line 6: Style: `SESSION_LOCK`

sway/lock.c
```diff
@@ -0,0 +1,174 @@
...
+static void handle_session_lock_destroy(struct wl_listener *listener, void *data)
+{
+	struct sway_session_lock_manager *lock_state = server.session_lock->data;
+	wl_list_remove(&lock_state->session_locked.link);
+	wl_list_remove(&lock_state->session_lock_destroyed.link);
+	free(lock_state);
+}
+
+struct wlr_session_lock_manager_v1 *sway_session_lock_manager_create(void)
+{
```
emersion, line 157: Style: braces on the previous line

sway/lock.c
```diff
@@ -0,0 +1,174 @@
...
+	struct wlr_session_lock_manager_v1 *session_lock = wlr_session_lock_manager_v1_create(server.wl_display);
+	struct sway_session_lock_manager *lock_state = calloc(1, sizeof(*lock_state));
+	if (!lock_state)
```
emersion, line 160: Style: braces are mandatory

sway/lock.c
```diff
@@ -0,0 +1,174 @@
...
+	// redraw everything
+	for (int i = 0; i < root->outputs->length; ++i) {
+		struct sway_output *output = root->outputs->items[i];
+		output_damage_whole(output);
+	}
+}
+
+static void handle_session_lock(struct wl_listener *listener, void *data)
+{
+	struct sway_session_lock_manager *lock_state = server.session_lock->data;
```
emersion, line 113: I'd prefer to avoid using `data` fields if possible, because of the decreased type-safety. We should be able to use this instead:

```c
struct sway_session_lock_manager *manager = wl_container_of(listener, new_lock, session_locked);
```

sway/lock.c
```diff
@@ -0,0 +1,174 @@
...
+	session_lock->data = lock_state;
+	lock_state->new_surface.notify = handle_new_surface;
+	lock_state->unlock.notify = handle_unlock;
+	lock_state->abandon.notify = handle_abandon;
+	lock_state->session_locked.notify = handle_session_lock;
+	lock_state->session_lock_destroyed.notify = handle_session_lock_destroy;
+	wl_signal_add(&session_lock->events.new_lock,
+		&lock_state->session_locked);
+	wl_signal_add(&session_lock->events.destroy,
+		&lock_state->session_lock_destroyed);
```
emersion, line 171: Please use the signal name for all listeners.

include/sway/lock.h
```diff
@@ -0,0 +1,23 @@
...
+	SESSIONLOCK_STATE_LOCKED,
+	SESSIONLOCK_STATE_ABANDONED,
+};
+
+struct sway_session_lock_manager {
+	enum sway_lock_state state;
```
emersion, line 12: Maybe we could replace `state` with a `bool locked` field? Then we can just check `manager->lock == NULL` to figure out whether the locker has crashed. This ensures we only have a single source of truth.

sway/lock.c
```diff
@@ -0,0 +1,174 @@
...
+		output_damage_whole(output);
+	}
+}
+
+static void handle_session_lock(struct wl_listener *listener, void *data)
+{
+	struct sway_session_lock_manager *lock_state = server.session_lock->data;
+	struct wlr_session_lock_v1 *lock = data;
+	struct wl_client *client = wl_resource_get_client(lock->resource);
+	struct sway_seat *seat;
```
emersion, line 116: Style: this variable can be declared right before the `wl_list_for_each` invocation

sway/lock.c
```diff
@@ -0,0 +1,174 @@
...
+	for (int i = 0; i < root->outputs->length; ++i) {
+		struct sway_output *output = root->outputs->items[i];
+		output_damage_whole(output);
+	}
+}
+
+static void handle_abandon(struct wl_listener *listener, void *data)
+{
+	struct sway_session_lock_manager *lock_state = server.session_lock->data;
+	struct sway_seat *seat;
```
emersion, line 90: Style: this variable can be declared right before the `wl_list_for_each` invocation

sway/lock.c
```diff
@@ -0,0 +1,174 @@
...
+	surf->map.notify = handle_surface_map;
+	wl_signal_add(&lock_surface->events.map, &surf->map);
+	surf->destroy.notify = handle_surface_destroy;
+	wl_signal_add(&lock_surface->events.destroy, &surf->destroy);
+}
+
+static void handle_unlock(struct wl_listener *listener, void *data)
+{
+	struct sway_session_lock_manager *lock_state = server.session_lock->data;
+	struct sway_seat *seat;
```
emersion, line 58: Style: this variable can be declared right before the `wl_list_for_each` invocation

sway/lock.c
```diff
@@ -0,0 +1,174 @@
...
+	struct sway_session_lock_surface* surf = wl_container_of(listener, surf, destroy);
+	wl_list_remove(&surf->map.link);
+	wl_list_remove(&surf->destroy.link);
+	output_damage_whole(surf->output);
+	free(surf);
+}
+
+static void handle_new_surface(struct wl_listener *listener, void *data)
+{
+	//struct sway_session_lock_manager *lock_state = server.session_lock->data;
```
emersion, line 36: This can be removed if unused

sway/desktop/render.c
```diff
@@ -1056,6 +1058,40 @@ void output_render(struct sway_output *output, struct timespec *when,
 		wlr_renderer_clear(renderer, (float[]){1, 1, 0, 1});
 	}
 
+	if (lock_state->state != SESSIONLOCK_STATE_UNLOCKED) {
```
emersion, line 1061: Maybe we should extract this block to a helper function, to make the `goto`s less confusing.

emersion (changes requested):
Thanks for the patch! Here's a first round of comments.

include/sway/input/seat.h
danieldg: I mostly did (that interface wires up to `seat_set_exclusive_client` which the lock uses to block input to other clients) but that doesn't handle setting the focus to the client in the first place.  Normally, that's because a client already has focus when it requests the inhibitor be activated.

sway/lock.c
danieldg: I agree, but is there a better way to indicate "sway itself wants to own the lock"?  Adding a bunch of checks if the session is locked in addition to the input-grab is non-null seems worse.

include/sway/input/seat.h
danieldg: Note: at least swaylock v1.6 will run `lock` before creating any surfaces at all.  That requires setting up the inhibitor right away (which works just fine, dropping any input except `--locked` binds).  However, the `wl_keyboard` API seems to require a surface for input (and all the helpers that `sway_force_focus` calls require it too), so we need to postpone that until we have such a surface.

sway/lock.c
danieldg: Creating a wl_client with a dead or hanging file descriptor would also work, but that seems worse than changing all checks of `exclusive_client` to also check for the lock.

sway/lock.c
```diff
@@ -0,0 +1,170 @@
...
+		output_damage_whole(output);
+	}
+}
+
+static void handle_session_lock(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_manager *manager = wl_container_of(listener, manager, new_lock);
+	struct wlr_session_lock_v1 *lock = data;
+	struct wl_client *client = wl_resource_get_client(lock->resource);
+
+	if (manager->locked) {
```
mstoeckl, line 118: Changing this to `if (manager->locked && manager->lock) {` would make the `bindsym --locked [...] exec swaylock` trick to recover from a lock screen crash possible.

(I will probably be using Sway with this branch from now on, and will let you know if I find any bugs.)

danieldg: Ah, that was a mistake when converting the enum to bool; fixed now, thanks for finding it.

danieldg:
68a3c06bb adds support for output resolution changes.

danieldg:
https://github.com/danieldg/sway/tree/lock-scene-graph is a rebase of this on top of https://github.com/swaywm/sway/pull/6844 - once that is merged, I'll update the branch this request references.

L-as:
Just wanted to note that I'm daily-driving https://github.com/danieldg/sway/tree/lock-scene-graph (ae6ff23e9c51ed8a1625135f8cc27f12d2d16abf right now) with [wlroots d3a2ea](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/d3a2ea33e5308f46d4c148cc1de7c428de782536) and it works without any problems at all as far as I've noticed. `swaylock` works fine too, albeit I haven't tested the functionality of the protocol really.

L-as:
Thank you @danieldg

danieldg:
@WhyNotHugo:
> I've been using the same version for a while and notice a mildly annoying issue. After unlocking the screen, no window has focus.

See if the latest commit on that branch fixes this for you; it more closely matches how the layer-shell/input-manager locking works.

include/sway/server.h
```diff
@@ -89,6 +90,8 @@ struct sway_server {
 	struct wl_listener output_manager_apply;
 	struct wl_listener output_manager_test;
 
+	struct wlr_session_lock_manager_v1 *session_lock;
```
emersion, line 93: I'd prefer to store `sway_session_lock_manager` here (maybe the struct could even be embedded so that we don't need to deal with alloc errors). That way we don't need to go through a `void *data` pointer to get back the `sway_session_lock_manager`.

include/sway/lock.h
```diff
@@ -0,0 +1,17 @@
...
+#define _SWAY_LOCK_H
+#include <wlr/types/wlr_session_lock_v1.h>
+
+struct sway_session_lock_manager {
+	bool locked;
+
+	struct wlr_session_lock_v1 *lock;
+	struct wl_listener new_surface;
+	struct wl_listener unlock;
+	struct wl_listener abandon;
```
emersion, line 11: To make it clear all of these events are triggered by `wlr_session_lock_v1`, maybe we can slap a `lock_` prefix? That way no need to introduce new wording like "abandon".

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+
+	// redraw everything
+	for (int i = 0; i < root->outputs->length; ++i) {
+		struct sway_output *output = root->outputs->items[i];
+		output_damage_whole(output);
+	}
+}
+
+static void handle_session_lock_destroy(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_manager *manager = wl_container_of(listener, manager, manager_destroy);
```
emersion, line 156: Maybe we can `assert(manager->lock == NULL)` here?

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+		// copied from seat_set_focus_layer -- deduplicate?
+		struct sway_node *previous = seat_get_focus_inactive(seat, &root->node);
+		if (previous) {
+			// Hack to get seat to re-focus the return value of get_focus
+			seat_set_focus(seat, NULL);
+			seat_set_focus(seat, previous);
+		}
+	}
+
+
```
emersion, line 92: Style nit: stray newline

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+	wl_list_remove(&surf->destroy.link);
+	wl_list_remove(&surf->surface_commit.link);
+	wl_list_remove(&surf->output_mode.link);
+	output_damage_whole(surf->output);
+	free(surf);
+}
+
+static void handle_new_surface(struct wl_listener *listener, void *data) {
+	struct wlr_session_lock_surface_v1 *lock_surface = data;
+	struct sway_session_lock_surface* surf = calloc(1, sizeof(*surf));
```
emersion, line 47: Style: `*` on the right

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+	if (surf == NULL)
```
emersion, line 48: Style: braces are mandatory

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+		return;
+
+	sway_log(SWAY_DEBUG, "new lock layer surface");
+
+	struct sway_output *output = lock_surface->output->data;
+	wlr_session_lock_surface_v1_configure(lock_surface, output->width, output->height);
```
emersion, line 54: This configures the surface with the buffer size. I don't think this will work correctly on transformed or HiDPI outputs.

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+}
+
+static void handle_surface_commit(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_surface* surf = wl_container_of(listener, surf, surface_commit);
+	output_damage_surface(surf->output, 0, 0, surf->surface, false);
+}
+
+static void handle_output_mode(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_surface* surf = wl_container_of(listener, surf, output_mode);
+	wlr_session_lock_surface_v1_configure(surf->lock_surface, surf->output->width, surf->output->height);
```
emersion, line 32: Ditto

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+	sway_force_focus(surf->surface);
+	output_damage_whole(surf->output);
+}
+
+static void handle_surface_commit(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_surface* surf = wl_container_of(listener, surf, surface_commit);
+	output_damage_surface(surf->output, 0, 0, surf->surface, false);
+}
+
+static void handle_output_mode(struct wl_listener *listener, void *data) {
```
emersion, line 30: The logical size of the output can change even if the mode doesn't, for instance when the scale or transform changes. We should listen for `output.events.commit` instead and check the `committed` bitfield in the event struct.

sway/desktop/output.c
```diff
@@ -275,6 +276,23 @@ static void for_each_surface_container_iterator(struct sway_container *con,
...
 static void output_for_each_surface(struct sway_output *output,
 		sway_surface_iterator_func_t iterator, void *user_data) {
+	struct sway_session_lock_manager *lock_state = server.session_lock->data;
+	if (lock_state->locked) {
+		if (lock_state->lock == NULL) {
+			return;
+		}
+		struct wlr_session_lock_surface_v1 *lock_surface;
+		wl_list_for_each(lock_surface, &lock_state->lock->surfaces, link) {
+			if (lock_surface->output != output->wlr_output)
```
emersion, line 286: Style: braces are mandatory

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+}
+
+struct wlr_session_lock_manager_v1 *sway_session_lock_manager_create(void) {
+	struct wlr_session_lock_manager_v1 *session_lock = wlr_session_lock_manager_v1_create(server.wl_display);
+	struct sway_session_lock_manager *lock_state = calloc(1, sizeof(*lock_state));
+	if (!lock_state) {
+		return NULL;
+	}
+
+	session_lock->data = lock_state;
```
emersion, line 169: If the type of `server->session_lock` is changed as described above, we can drop this.

sway/lock.c
```diff
@@ -0,0 +1,179 @@
...
+	wl_list_remove(&lock_state->new_surface.link);
+	wl_list_remove(&lock_state->unlock.link);
+	wl_list_remove(&lock_state->abandon.link);
+
+	struct sway_seat *seat;
+	wl_list_for_each(seat, &server.input->seats, link) {
+		seat->exclusive_client = NULL;
+	}
+
+	// redraw everything
```
emersion, line 115: Maybe it's worth extracting this into a `desktop_damage_whole` function in `sway/desktop/desktop.c`?

sway/lock.c
danieldg: I've tested this with a scale=2 display and it sends the logical size, which matches my read of the protocol ("The width and height are in surface-local coordinates").  Also tested for a transformed surface.

sway/lock.c
danieldg: I just tested using `output.events.commit`, and the signal (ignoring `committed == 3`) is not sent when a nested wayland display's window is resized.  It is correctly sent when using `swaymsg output` to adjust the mode, which I had not previously tested.  Is the lack of this signal for this a bug in wlroots, or do I need to hook both here?

sway/lock.c
danieldg: Since #6844 will end up reworking that anyway, I'd prefer to leave it as is unless you prefer it moved.

danieldg:
I found some additional cases when the input grab would get re-assigned from the screenlock to the desktop, I think 528a72f covers that now.

sway/lock.c
emersion: Ack

sway/lock.c
emersion: Hm right, to work correctly with current wlroots we probably need to listen to both.

sway/lock.c
emersion: Oh, I mixed up `sway_output` and `wlr_output`. Sorry for the noise, your code is correct!

sway/desktop/render.c
```diff
@@ -1056,6 +1056,39 @@ void output_render(struct sway_output *output, struct timespec *when,
...
+
+		if (server.session_lock.lock != NULL) {
+			struct render_data data = {
+				.damage = damage,
+				.alpha = 1.0f,
+			};
+
+			struct wlr_session_lock_surface_v1 *lock_surface;
+			wl_list_for_each(lock_surface, &server.session_lock.lock->surfaces, link) {
+				if (lock_surface->output != wlr_output)
```
emersion, line 1080: Style nit: missing braces

emersion:
Apart from these last two remaining comments, LGTM!

emersion (approved):
LGTM, thanks!

WhyNotHugo:
I've been using this for a few weeks now and it's pretty solid. The usual glitches that used to appear when unplugging/plugging/resuming are gone. :+1:

PR #6880, opened 2022-03-09
Fix meson sandbox violation error

### LINKED ISSUE #6878: Meson sandbox violation for xdg-shell.xml
When building sway (HEAD = 04676936e71f6f) on Ubuntu 20.04 with the latest meson (0.61.2), the build fails due to the error:

```
protocols/meson.build:37:1: ERROR: Sandbox violation: Tried to grab file xdg-shell.xml from a nested subproject.
```

https://github.com/swaywm/sway/blob/04676936e71f6fccccb098f3232d16572b140902/protocols/meson.build#L37-L42

This can be fixed by changing `input: xml` to `input: files(xml)`. This is also an issue with the following two `custom_target()` functions.

stevenengler:
I don't know anything about meson so I don't know if this is the proper solution, but this fixes #6878 for me. (There is another sandbox violation in wlroots, but that's separate.)

### DIFF
--- a/protocols/meson.build
+++ b/protocols/meson.build
@@ -36,13 +36,13 @@ foreach p : protocols
 	xml = join_paths(p)
 	wl_protos_src += custom_target(
 		xml.underscorify() + '_server_c',
-		input: xml,
+		input: files(xml),
 		output: '@BASENAME@-protocol.c',
 		command: [wayland_scanner, 'private-code', '@INPUT@', '@OUTPUT@'],
 	)
 	wl_protos_headers += custom_target(
 		xml.underscorify() + '_server_h',
-		input: xml,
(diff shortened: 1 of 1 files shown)


stevenengler:
> @stevenengler, did you manage to fix the violation in `wlroots`?

I think I did the same thing as this PR, I just added `files()` to the input at [`wlroots/protocol/meson.build:65`](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/9de992b9/protocol/meson.build#L65).

According to my notes, the error was:

```
wlroots| subprojects/wlroots/protocol/meson.build:63:1: Exception: Sandbox violation: Tried to grab file xdg-shell.xml outside current (sub)project.
```

PR #6883, opened 2022-03-10
Replace pcre with pcre2

### LINKED ISSUE #6838: Replace pcre with pcre2
Originally pcre was used to increase compatibility with i3 in #1151 however recently i3 was updated to use pcre2 instead: https://github.com/i3/i3/commit/6c7efc4de87efb1cbdfe2acbaa5def1d1d5fa8a5.

Because of this it would be quite good if sway did the same as pcre was deprecated a few years ago, incidentally improving compatibility with i3. (Although it might be useful to wait for the next i3 release, since the commit has not been tagged yet.)

ndren:
Thank you @psychon for fixing the patch I've sent over email. This should fix #6838.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -12,6 +12,7 @@ packages:
   - mesa-dev
   - meson
   - pango-dev
+  - pcre2-dev
   - pixman-dev
   - scdoc
   - wayland-dev
--- a/.builds/archlinux.yml
+++ b/.builds/archlinux.yml
@@ -9,6 +9,7 @@ packages:
   - libxkbcommon
   - meson
   - pango
+  - pcre2
   - scdoc
   - wayland
   - wayland-protocols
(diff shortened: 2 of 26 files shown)


ndren:
I'm not too sure how to fix the build since pcre was not listed in any of the .yml build files. Any suggestions?

emersion:
I guess the dependency wasn't explicit because base system packages depend on pcre v1, so we got it "by chance". We need to make the pcre v2 dep explicit in `.builds/*.yml`.

emersion (commented):
Overall the code looks good to me!

bl4ckb0ne:
The package seems to have a different name on alpine. Maybe drop the suffix after the 2?

```
$ apk search libpcre2
libpcre2-32-10.39-r0
libpcre2-16-10.39-r0
pcre2-dev-10.39-r0
pcre2-10.39-r0
```

ndren:
Thank you, now it's building.

emersion (approved):
Thanks!

PR #6889, opened 2022-03-14
swaybar: set opaque region

emersion:
When the background color is fully opaque, set the surface's opaque
region to the whole surface.


bl4ckb0ne (approved):
LGTM, thanks!

PR #6890, opened 2022-03-14
swaybar: remove swaybar_output.input_region

emersion:
No need to keep the region around, we can immediately destroy it
after the wl_surface.set_input_region request.


bl4ckb0ne (approved):
LGTM, thanks :)

Issue #4511, opened 2019-08-29
Feature request: config option to disable primary-selection (clipboard) protocol


jaimet:
Feature request: I'd love some config option that would completely disable the primary-selection protocol message proxying that sway does. I've just asked on IRC if anything like this already exists, and Kenny Levinson kindly informed me that this isn't configurable at the moment, hence this feature request.

If you need any more info, please let me know. TIA, J

PS I've just found https://wiki.gnome.org/Initiatives/Wayland/PrimarySelection which says "Constraints for a Wayland primary selection: We should address the security concerns by making sure that the feature is off by default and has to be enabled by the user". I appreciate that it's a different DE, but the point is interesting nonetheless.

emersion:
Why do you want to disable it?

jaimet:
Because I don't like it ([Richard Feynman's explanation](https://www.youtube.com/watch?v=Tn6fE8PvVCE) may help)

emersion:
Then don't use it

jaimet:
Ha ha! Brilliant! Ok, I'll take that as a "This is not going to be implemented".

emersion:
If you can give a sensible reason to allow users to disable it, then I'll reconsider. If the reason is just "I don't like it", then no.

jaimet:
Ok, I'll try. (I'm not sure that I understand exactly what your criteria are for a "sensible reason", but I won't let that hold me back!)

Firstly, I often select stuff to make it stand out on the screen, and I don't want "the system" to store that selection because I think that's a security vulnerability. For example, "Hey, Joe, what do you think about this really sensitive word on this page, yeah, I'll select it so that you can see which sensitive word I'm talking about. Shit - now that I've selected it to show it to you, the system has recorded that sensitive word in primary, dammit! But I don't want that sensitive word hanging around in primary, so now I need to select something else and then verify that the second selection has replaced the sensitive word. Arghhhh."

Secondly (related to the first, but not the same), there's the higher "active threshold requirement" (I just made that up) of copying into the clipboard than there is to copy into primary. I don't want the "system" to remember things just because I selected them - I want to "have to" select *and then do something* before the system copies the selection.

[...]

PR #6897, opened 2022-03-17
Add primary_selection config option

t1lde:
See: https://github.com/swaywm/sway/issues/4511.

Adds a config option 'primary_selection', which explicitly
enables/disables the primary selection clipboard.

This is implemented as a config-option check in 'handle_request_set_primary_selection', which clears
the primary selection if the option is set to 'disabled'.

## Effect of Changes

With 'primary_config = enabled' in the config, the primary selection clipboard will not be filled.

## A Consideration

After setting the option with the cli: `sway primary_selection disabled`, there may still be a selection in the clipboard until the selection is cleared later. This could be addressed by clearing the clipboard after the option is set, but I'm not sure if that kind of effect would be desirable in the command function.

## Alternatives

- Existing workarounds suggested in the linked issue
- A better & more flexible config option that can instead disable the middle-mouse paste action, while retaining the other functionality of middle_emulation. It seems like this would require some deeper changes to wlr, or lipinput, so I didn't pursue this yet (but would be willing to, perhaps with some guidance)

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -130,6 +130,7 @@ struct input_config_tool {
 	enum sway_tablet_tool_mode mode;
 };
 
+
 /**
  * options for input devices
  */
@@ -468,6 +469,12 @@ enum xwayland_mode {
 	XWAYLAND_MODE_IMMEDIATE,
 };
 
+enum config_primary_selection {
+	PRIMARY_SELECTION_DEFAULT,
+	PRIMARY_SELECTION_DISABLED,
+	PRIMARY_SELECTION_ENABLED,
+};
+
 /**
  * The configuration struct. The result of loading a config file.
  */
@@ -527,6 +534,7 @@ struct sway_config {
 	bool auto_back_and_forth;
 	bool show_marks;
 	enum alignment title_align;
+	enum config_primary_selection primary_selection;
 
 	bool tiling_drag;
 	int tiling_drag_threshold;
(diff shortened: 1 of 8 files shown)


emersion:
It would be better to not create the `zwp_primary_selection_device_manager_v1` global if primary selection is disabled. This would let clients know that primary selection is disabled, instead of silently dropping the requests.

include/sway/config.h
```diff
@@ -527,6 +534,7 @@ struct sway_config {
 	bool auto_back_and_forth;
 	bool show_marks;
 	enum alignment title_align;
+	enum config_primary_selection primary_selection;
```
emersion, line 537: Do we really need a tri-state field here? Can't we just use a `bool`?

ivan-volnov:
@t1lde Thank you for your effort. I'm looking forward to get this feature in the next release.

emersion (changes requested):
(See comments above.)

t1lde:
Sorry I haven't gotten back to this, will give it another go with your suggestions when I get some time

openssl-rand:
Please get this merged, I cannot count the number of times I have accidentally pasted stuff that I didn't even mean to copy while scrolling around in VSCode.

I have read #4511 and have tried the workarounds, but they either don't work or are still not a permanent solution.

The code is already written, it just needs to be tested and merged. 

Thanks

emersion:
> The code is already written, it just needs to be tested and merged.

This code is not ready to be merged, see the comments above.

openssl-rand:
Yes, after that change is made. I don't know C but it seems like a quite simple change.

AidanGG:
Submitted https://github.com/t1lde/sway/pull/1 for @t1lde to merge into this PR.

emersion:
Superseded by https://github.com/swaywm/sway/pull/7312

PR #6898, opened 2022-03-18
Allow not having SWAYSOCK set if there is a single socket in XDG_RUNTIME_DIR

mazunki:
Quite often, I run into a scenario where I've set XDG_RUNTIME_DIR as a hardcoded path, but I haven't exported SWAYSOCK since it depends on the PID of sway.

If there is a single `sway-ipc.USER_ID.%d.sock` in the current `${XDG_RUNTIME_DIR}` there is a good chance this is the current sway session. Since some people may be running multiple instances of Sway, it should fail if more than one sway pid is found.

### DIFF
--- a/common/ipc-client.c
+++ b/common/ipc-client.c
@@ -13,6 +13,25 @@ static const char ipc_magic[] = {'i', '3', '-', 'i', 'p', 'c'};
 
 #define IPC_HEADER_SIZE (sizeof(ipc_magic) + 8)
 
+int query_for_swaypid(char *xdg_runtime_dir) {
+	if (xdg_runtime_dir != NULL) {
+		char pidline[1024];
+		char *pid;
+		FILE *fp = popen("pidof sway", "r");
+		fgets(pidline, 1024, fp);
+		pid = strtok(pidline, " ");
+		int sway_pid = atoi(pid);
+		if (strtok(NULL, " ") != NULL) {
+			return -1;
+		} else {
+			return sway_pid;
+		}
+	} else {
+		exit(EXIT_FAILURE);
+	};
+}
+
+
 char *get_socketpath(void) {
 	const char *swaysock = getenv("SWAYSOCK");
 	if (swaysock) {
@@ -46,10 +65,28 @@ char *get_socketpath(void) {
 			if (line[nret - 1] == '\n') {
 				line[nret - 1] = '\0';
 			}
+			free(line);
 			return line;
 		}
 	}
 	free(line);
+
+	if (getenv("XDG_RUNTIME_DIR") && getuid()) {
+		int swaypid = query_for_swaypid(getenv("XDG_RUNTIME_DIR"));
+		if (swaypid > 0) {
+			size_t path_sz = snprintf(NULL, 0, "%s/sway-ipc.%u.%i.sock", getenv("XDG_RUNTIME_DIR"), getuid(), swaypid);
+			char *swaysock_runtime = malloc(path_sz+1);
+			snprintf(swaysock_runtime, path_sz+1, "%s/sway-ipc.%u.%i.sock", getenv("XDG_RUNTIME_DIR"), getuid(), swaypid);
+			if (access(swaysock_runtime, F_OK) != -1) {
+				return strdup(swaysock_runtime);
+			} else {
+				fprintf(stderr, "sway socket not detected.\n"); // found a pid, but no socket file
+			}
+		} else {
+			fprintf(stderr, "Found more than one sway instance running.\n");
+			exit(EXIT_FAILURE);
+		}
+	}
 	return NULL;
 }
 
--- a/sway/main.c
+++ b/sway/main.c
@@ -87,24 +87,6 @@ void detect_proprietary(int allow_unsupported_gpu) {
 	fclose(f);
 }
 
-int query_for_swaypid(char *xdg_runtime_dir) {
-	if (xdg_runtime_dir != NULL) {
-		char pidline[1024];
-		char *pid;
-		FILE *fp = popen("pidof sway", "r");
-		fgets(pidline, 1024, fp);
-		pid = strtok(pidline, " ");
-		int sway_pid = atoi(pid);
-		if (strtok(NULL, " ") != NULL) {
-			return -1;
-		} else {
-			return sway_pid;
-		}
-	} else {
-		exit(EXIT_FAILURE);
-	};
-}
-
 void run_as_ipc_client(char *command, char *socket_path) {
 	int socketfd = ipc_open_socket(socket_path);
 	uint32_t len = strlen(command);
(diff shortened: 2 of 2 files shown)


emersion:
NACK. This will pick a random Sway instance, which isn't the right thing to do.

kennylevinsen:
> ... I haven't exported SWAYSOCK since it depends on the PID of sway.

You should not need to set SWAYSOCK in normal use. It is set by sway for its children.

mazunki:
> > ... I haven't exported SWAYSOCK since it depends on the PID of sway.
> 
> You should not need to set SWAYSOCK in normal use. It is set by sway for its children.

How is it set? It is not set on my system.

mazunki:
> NACK. This will pick a random Sway instance, which isn't the right thing to do.

It will only pick the socket if it's alone in the runtime directory. If there's more than one to pick from, it will fail saying there's more than one Sway instance running.

kennylevinsen:
> How is it set?

https://github.com/swaywm/sway/blob/440d0bc22d57b8b0b21a8acbf127243b8d08cfae/sway/ipc-server.c#L117

> It is not set on my system.

It most certainly is set for every child process of sway, unless you yourself overwrote or unset it.

It won't be set for anything started outside sway, such as over SSH or on a different VT.

mazunki:
> > How is it set?
> 
> https://github.com/swaywm/sway/blob/440d0bc22d57b8b0b21a8acbf127243b8d08cfae/sway/ipc-server.c#L117
> 
> > It is not set on my system.
> 
> It most certainly is set for every child process of sway, unless you yourself overwrote or unset it.
> 
> It won't be set for anything started outside sway, such as over SSH or on a different VT.

Thanks for the info. I know where the problem is now. Because I launch the `foot --server` as a user service through s6, it does not inherit the sway socket.

I'd like to solve this "the clean way", but I don't exactly know how to do that. Any suggestions on how to do that? Regardless, I think having the PR as an option of last resort, as an option, should be okay.

This is the result after I launch `footclient` through `wofi` or `on-click` in Waybar.

image

This is the result after I run `foot` through either Waybar or `wofi`.

image

kennylevinsen:
> I'd like to solve this "the clean way", but I don't exactly know how to do that. Any suggestions on how to do that?

Start the foot server as an `exec` in your sway config file instead of as a service, or run a sub s6 supervisor instance as an exec in your sway config file and use that to start anything that needs WAYLAND_DISPLAY or SWAY SOCK access.

mazunki:
> Start the foot server as an `exec` in your sway config file instead of as a service, or run a sub s6 supervisor instance as an exec in your sway config file and use that to start anything that needs WAYLAND_DISPLAY or SWAY SOCK access.

While running the foot server from the config would work, I don't think `exec`-ing a supervisor from Sway is the clean way to do it. It would be clean if I could pass the socket on to the already-running instance of the s6-supervisor, similar to how I can pass it on to dbus.

kennylevinsen:
Running things from within sway (supervisor or not) is the only clean way to do it for a few reasons:

1. The lifetime of anything needing Wayland, X11 or sway ipc is directly tied to the lifetime of sway
2. As a graphical shell, the lifetime of your entire session is usually tied to the lifetime of sway
3. Environments are inherited and can only be read by children of sway

Any other approach will result in duct tape (like guessing socket locations, waiting for a signal that sway started or handling crash loops until it is up, handling that things crash when sway closes), trying to make something started outside sway think it was started inside sway.

Nested supervisors is also specifically supported by s6 (and runit, and...) for this kind of use-case, although I usually don't bother with a supervisor for these kinds of small things - they tend not to crash.

mazunki:
> Running things from within sway (supervisor or not) is the only clean way to do it for a few reasons:
> 
>     1. The lifetime of anything needing Wayland, X11 or sway ipc is directly tied to the lifetime of sway
> 
>     2. As a graphical shell, the lifetime of your entire session is usually tied to the lifetime of sway
> 
>     3. Environments are inherited and can only be read by children of sway
> 
> 
> Any other approach will result in duct tape (like guessing socket locations, waiting for a signal that sway started or handling crash loops until it is up, handling that things crash when sway closes), trying to make something started outside sway think it was started inside sway.
> 
> Nested supervisors is also specifically supported by s6 (and runit, and...) for this kind of use-case, although I usually don't bother with a supervisor for these kinds of small things like mako, poweralertd or in this case foot server - they tend not to crash.

Good points, although I wouldn't want my spawned-from-inside-sway terminals to just die (let's say I have a server running on one of them, because i hadn't set up a service for it yet). By having the foot server be bound outside of sway this is not an issue. Passing on variables to footclient(s) (and removing them) would be a foot implementation detail though.

My main reason for supervision is not because i need the reliability, but rather because I enjoy the simplicity of being able to replace my wm without needing to reconfigure each daemon.

On a similar note, the Arch wiki suggests to run Pipewire through a sway-exec, but imo that's an awful solution since there's no reason Sway should have control over the audio server. Passing the socket path over to Sway is the best way to do it.

I agree with you in that apps which need a GUI require a wl/x11 instance to exist, and that `pidof`-ing is not the most elegant of solutions, and I will set up a chainloader for Waybar and whatnot. Let's see what I come up with.

Regardless, I still believe allowing fallback solutions to be okay, since it allows for initiates to tinker around with stuff with more ease, and less boilerplate.

kennylevinsen:
> Good points, although I wouldn't want my spawned-from-inside-sway terminals to just die

Your terminals will still die as foot server loses its Wayland display connection, like every other Wayland or X11 client (see point 1).

The purpose of the foot server is not persistence, it only exists to lower memory utilization and terminal open latency. Use tmux for persistence.

> On a similar note, the Arch wiki suggests to run Pipewire through a sway-exec, but imo that's an awful solution since there's no reason Sway should have control over the audio server.

Pipewire is also a video server and supposed to be run in your user session (see point 2). It is often dbus activated, but the obvious place to start it manually if needed would be from sway config.

If you use an s6 sub supervisor, then each wm only needs that single exec. I do not have enough wms for the number of execs to be a concern though, especially as the services needed in each differ.

> Regardless, I still believe allowing fallback solutions to be okay, since it allows for initiates to tinker around with stuff with more ease, and less boilerplate.

The only alternative is to have your sway config run something that sends the vars as they are discovered, but that is systemd-levels of silly roundabout design (see `systemctl --user update-environment`, and point 3).

mazunki:
> Your terminals will still die as foot server loses its Wayland display connection, like every other Wayland or X11 client (see point 1).
> 
> The purpose of the foot server is not persistence, it only exists to lower memory utilization and terminal open latency. Use tmux for persistence.

Yeah, tmux is a better solution. I wonder if it's even possible to "give foot --server away to another wm" through some socket magic.

> Pipewire is also a video server and supposed to be run in your user session (see point 2). It is often dbus activated, but the obvious place to start it manually if needed would be from sway config.
> 
> If you use an s6 sub supervisor, then each wm only needs that single exec. I do not have enough wms for the number of execs to be a concern though, especially as the services needed in each differ.

I don't understand why that's necessarily the case. I would often like to attach multiple users to the same speaker, say my siblings or a dedicated spotify account in daemon mode.

The desktop-portal interface, yes, that's a thing to care about, but those merely connect to pipewire through wlroots, don't they?

> 
> The only alternative is to have your sway config run something that sends the vars as they are discovered, but that is systemd-levels of silly roundabout design (see `systemctl --user update-environment`, and point 3).

What I have been doing for some months now is to add a silly roundabout in my terminals to run commands by exporting SWAYSOCK myself through a `$(find "${XDG_RUNTIME_DIR:?}" -name "sway-ipc*.sock")` maneuver since OpenRC didn't have any way to deal with user services. It was awful, hence this PR.

kennylevinsen:
> Yeah, tmux is a better solution. I wonder if it's even possible to "give foot --server away to another wm" through some socket magic.

No, recreating dead terminals would have to be implemented as a feature in foot. Updating the display location is the least problematic part.

> I don't understand why that's necessarily the case. I would often like to attach multiple users to the same speaker, say my siblings or a dedicated spotify account in daemon mode.

Your video output and input devices is run as a user session, so why is it weird to do the same for audio? The (valid) assumption is that each physical seat only has one physical user at a time. If you're sitting by the computer, the speakers, microphone, webcam etc. service you.

> What I have been doing for some months now is to add a silly roundabout in my terminals to run commands by exporting SWAYSOCK myself through a `$(find "${XDG_RUNTIME_DIR:?}" -name "sway-ipc*.sock")` maneuver since OpenRC didn't have any way to deal with user services. It was awful, hence this PR.

We're not going to add socket guessing as a feature. It's no less ugly to push the hack upstream.

The only clean option remains running stuff from the sway config, supervisor or not. Anything else will require additional work with no practical benefit.

mazunki:
> Your video output and input devices is run as a user session, so why is it weird to do the same for audio? The (valid) assumption is that each physical seat only has one physical user at a time. If you're sitting by the computer, the speakers, microphone, webcam etc. service you.

In most cases that's true, although having the option to be able to jump between two ttys (say games and work) sure is nice to have. If Pipewire or their siblings pules/jack are served from inside Sway that's problematic to deal with.

 > We're not going to add socket guessing as a feature. It's no less ugly to push the hack upstream.

In hindsight I agree. Let's keep it clean.

PR #6924, opened 2022-03-29
contrib: add active opacities and applications that ignore opacity

aelfsyg:
This pull request allows a user of the window opacity script to provide opacities for active and inactive windows, as well as overrides for specific applications, and a list of applications that should be ignored by the script.

The default active opacity is provided with the flag `-a`, and an override with `-A` as a JSON string of a dictionary of application names (`window.app_id`) and their opacity. The same can be done for inactive opacities with `-i`, and `-I`.

It is also common that a user will want certain applications to be completely ignored, such as VLC, or a browser, and this can be done with the `--ignore` flag, accepting the names of multiple . 

`~/src/sway/contrib/window-opacity.py -a 0.9 -A '{"emacs" : 0.95}' -i 0.8 -I '{"emacs" : 0.9}' --ignore firefox`


emersion:

PR #6929, opened 2022-04-05
wp-surface-scale-v1: Initial implementation

kennylevinsen:
Depends on https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3511 and https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/143

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -7,6 +7,7 @@
 #include <strings.h>
 #include <sys/stat.h>
 #include <wayland-server-core.h>
+#include <wlr/types/wlr_fractional_scale_v1.h>
 #include <wlr/types/wlr_linux_dmabuf_v1.h>
 #include <wlr/types/wlr_output_layout.h>
 #include <wlr/types/wlr_subcompositor.h>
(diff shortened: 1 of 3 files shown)


protocols/meson.build
```diff
@@ -20,6 +20,7 @@ protocols = [
 	['idle.xml'],
 	['wlr-input-inhibitor-unstable-v1.xml'],
 	['wlr-output-power-management-unstable-v1.xml'],
+	['wp-surface-scale-v1.xml'],
```
emersion, line 23: We shouldn't need the protocol here if generated enums aren't used in wlroots' public header.

kennylevinsen: Ah, good point.

sway/tree/container.c
```diff
@@ -1325,6 +1326,7 @@ static void surface_send_enter_iterator(struct wlr_surface *surface,
 		int x, int y, void *data) {
 	struct wlr_output *wlr_output = data;
 	wlr_surface_send_enter(surface, wlr_output);
+	wlr_fractional_scale_v1_notify_scale(surface, wlr_output->scale);
```
emersion, line 1329: Hm, should we iterate over all outputs and send the highest scale here?

emersion:
There are a few other places where we call `wlr_surface_send_enter()`, but don't update the fractional scale, e.g. layer-shell.

emersion:
This PR has some useful bits which could be leveraged here: https://github.com/swaywm/sway/pull/7443

emersion:
I've created an alternative PR here: #7448

PR #6930, opened 2022-04-06
contrib/term-menu: new script

emersion:
Not sure it's a good idea to add this one, but I use it extensively. Unfortunately requires per-terminal rules since there's no standard.


emersion:
Eh using `$TERM` here won't work, since this will typically be run outside of the terminal emulator. In my dotfiles I have a separate `$TERMINAL` env var for this, but it's not great. Maybe should just unconditionally use the default one for Sway (foot) and comment out the rest.

PR #6936, opened 2022-04-09
Gcc 12 fixes

danieldg:
Sway master has two compile errors when using GCC 12; one false positive and one due to an actual bug.  This adjusts the code slightly to fix the warnings.


emersion (approved):
LGTM, thanks!

PR #6940, opened 2022-04-15
Xkb switch layout fix

### LINKED ISSUE #6011: Cycling through layouts doesn't work
**Sway Version:** 1.5.1

**Debug Log:**
https://pbox.ca/4064245

**Relevant config:** 
```
input type:keyboard {
     xkb_layout us,us(intl)
}
```

**Steps to reproduce:**
Set two keyboard layouts under xkb_layout. Either with a bindsym or through swaymsg, run `input type:keyboard xkb_switch_layout next`. Specifying an index (ie, 0 or 1) rather than `next` or `prev` is able to switch the layout just fine, but the keyboard layout does not change if `next` or `prev` is used.

vitek:
First calculate absolute layout indexes and then apply

Fixes #6011

### DIFF
--- a/sway/commands/input/xkb_switch_layout.c
+++ b/sway/commands/input/xkb_switch_layout.c
@@ -5,6 +5,12 @@
 #include "sway/input/input-manager.h"
 #include "log.h"
 
+struct xkb_switch_layout_action {
+	struct wl_list link;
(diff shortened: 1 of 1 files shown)


emersion:
Closing in favor of https://github.com/swaywm/sway/pull/6966

PR #6945, opened 2022-04-17
Fix errors caused by `640f3b9f` in wlroots

kraftwerk28:


vyivel:
This requires Sway to link to `xcb-icccm`.

vyivel:
911c685 isn't buildable by itself, so please squash these two commits into one (something like `Chase wlroots X11 hints update`).

emersion (approved):
Thanks!

PR #6953, opened 2022-04-23
Update grimshot.1.scd

Pound-Hash:
Fixed typo. The object is **files**, which is plural. **image** modifies files; it's not countable.


emersion (approved):
Thanks

PR #6954, opened 2022-04-23
jq optimisation

0Karakurt0:
proof

### DIFF
--- a/contrib/grimshot
+++ b/contrib/grimshot
@@ -131,10 +131,9 @@ elif [ "$SUBJECT" = "area" ] ; then
   fi
   WHAT="Area"
 elif [ "$SUBJECT" = "active" ] ; then
-  FOCUSED=$(swaymsg -t get_tree | jq -r 'recurse(.nodes[]?, .floating_nodes[]?) | select(.focused)')
-  GEOM=$(echo "$FOCUSED" | jq -r '.rect | "\(.x),\(.y) \(.width)x\(.height)"')
-  APP_ID=$(echo "$FOCUSED" | jq -r '.app_id')
-  WHAT="$APP_ID window"
+    eval "$( swaymsg -t get_tree |\
+        jq -r '..|try select(.focused == true)|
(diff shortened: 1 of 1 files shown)


emersion:
I don't think the few saved milliseconds is worth damaging readability.

0Karakurt0:
@emersion that should be better

progandy:
I would not use eval with unsanitized text.

AdamantGarth:
To avoid eval you can use read:
```sh
rs=$(printf '\x1e') # ASCII Record Separator
IFS=$rs read -r GEOM APP_ID WHAT < <(swaymsg -t get_tree |\
        jq -r '..|try select(.focused == true) | [
            "\(.rect.x),\(.rect.y) \(.rect.width)x\(.rect.height)",
            "\(.app_id)",
            "\(.app_id) window"
        ] | join("\u001e")')
```
But I don't think you can handle newlines in app_id with this approach in POSIX shell, though in Bash it would be possible.

An alternative would be:
```sh
rs=$(printf '\x1e') # ASCII Record Separator
__oldIFS=$IFS
IFS=$rs
set -- $(swaymsg -t get_tree |\
        jq -r '..|try select(.focused == true) | [
            "\(.rect.x),\(.rect.y) \(.rect.width)x\(.rect.height)",
            "\(.app_id)",
            "\(.app_id) window"
        ] | join("\u001e")')
IFS=$__oldIFS
GEOM=$1
APP_ID=$2
WHAT=$3
```
Which can handle any app_id (except with ASCII record separators, of course), but the readability... Yeah.

AdamantGarth:
Oh, and also one can build jq from master, or use [gojq](https://github.com/itchyny/gojq) so that jq starts up ~4 times faster (on my machine). The savings should be roughly equivalent to this patch then :)

PJungkamp:
Just putting quotes around the strings you're substituting isn't properly escaping them, which is especially important in the case of a shell `eval`.

You may be able to use a shell `eval` safely using jq's `@sh` formatter which should properly escape the variable contents.

```shell
# a demonstration of proper POSIX string escaping
echo "you're not handling the '\"' character properly" | jq -rR '"\(. | @sh)"'
```

PR #6966, opened 2022-04-27
xkb_switch_layout: fix relative layout switches

### LINKED ISSUE #6011: Cycling through layouts doesn't work

vitek:
First calculate absolute layout indexes and then apply

Fixes #6011

### DIFF
--- a/sway/commands/input/xkb_switch_layout.c
+++ b/sway/commands/input/xkb_switch_layout.c
@@ -81,6 +81,9 @@ struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "Unable to allocate actions");
 	}
 
+	/* Calculate new indexes first because switching a layout in one
+	   keyboard may result in a change on other keyboards as well because
+	   of keyboard groups. */
 	struct sway_input_device *dev;
 	wl_list_for_each(dev, &server.input->devices, link) {
 		if (strcmp(ic->identifier, "*") != 0 &&
--- a/sway/commands/input/xkb_switch_layout.c
+++ b/sway/commands/input/xkb_switch_layout.c
@@ -6,6 +6,11 @@
 #include "sway/input/input-manager.h"
 #include "log.h"
 
+struct xkb_switch_layout_action {
+	struct wlr_keyboard *keyboard;
+	xkb_layout_index_t layout;
+};
+
 static void switch_layout(struct wlr_keyboard *kbd, xkb_layout_index_t idx) {
 	xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(kbd->keymap);
 	if (idx >= num_layouts) {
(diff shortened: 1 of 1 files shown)


sway/commands/input/xkb_switch_layout.c
```diff
@@ -67,6 +73,9 @@ struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
 		relative = 0;
 	}
 
+	struct wl_list actions;
+	wl_list_init(&actions);
```
emersion, line 77: We should be able to just use a `xkb_layout_index_t` array instead of a `wl_list` here

sway/commands/input/xkb_switch_layout.c
```diff
@@ -29,10 +35,10 @@ static xkb_layout_index_t get_current_layout_index(struct wlr_keyboard *kbd) {
 	return layout_idx;
 }
 
-static void switch_layout_relative(struct wlr_keyboard *kbd, int dir) {
+static xkb_layout_index_t get_layout_relative(struct wlr_keyboard *kbd, int dir) {
 	xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(kbd->keymap);
 	xkb_layout_index_t idx = get_current_layout_index(kbd);
-	switch_layout(kbd, (idx + num_layouts + dir) % num_layouts);
+	return  (idx + num_layouts + dir) % num_layouts;
```
emersion, line 41: Style nit: two spaces here

sway/commands/input/xkb_switch_layout.c
vitek: Do you suggest to create an array of pre calculated xkb_layout_index_t[] and then zip it with original input devices list server.input->devices? This approach requires to do input device filtration twice. That's why I've introduced xkb_switch_layout_action struct. 

As I understand sway/wl doesn't provide vector-like type array only array of pointers so I still have to manually free() every single element.

Perhaps I didn't get your idea.

emersion: Eh, right, I forgot that some filtering was necessary.

What I had in mind was something among these lines:

```c
struct xkb_switch_layout_action *actions = calloc(wl_list_length(&server.input->devices), sizeof(struct xkb_switch_layout_action));
size_t actions_len = 0;

// To add an entry:
actions[actions_len++] = (struct xkb_switch_layout_action){
    .keyboard = kbd,
    .layout = layout,
};

// To iterate:
for (size_t i = 0; i < actions_len; i++) {
    switch_layout(actions[i].keyboard, actions[i].layout);
}

// To cleanup:
free(actions);
```

vitek: Fixed thanks!

sway/commands/input/xkb_switch_layout.c
```diff
@@ -77,12 +91,23 @@ struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
...
-			switch_layout_relative(dev->wlr_device->keyboard, relative);
+			action->layout = get_layout_relative(
+				dev->wlr_device->keyboard, relative);
 		} else {
-			switch_layout(dev->wlr_device->keyboard, layout);
+			action->layout = layout;
 		}
 	}
 
+	for (size_t i = 0; i < actions_len; i++) {
```
emersion, line 107: Can we add a small comment somewhere explaining that we need a two-step process because switching a layout in one keyboard may result in a change on other keyboards as well because of keyboard groups?

emersion:
Code LGTM!

sway/commands/input/xkb_switch_layout.c
vitek: fixed

emersion (approved):
Thanks!

(Next time, please amend your commit instead of pushing an additional fixup commit.)

PR #6970, opened 2022-04-30
Session lock focus fixes

danieldg:
Keyboard focus tracking with session lock has some incorrect edge cases around output destroy/create; this changes them to more closely mirror those used by layer_shell.

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -96,6 +96,7 @@ struct sway_server {
 		struct wlr_session_lock_manager_v1 *manager;
 
 		struct wlr_session_lock_v1 *lock;
+		struct wlr_surface *focused;
 		struct wl_listener lock_new_surface;
 		struct wl_listener lock_unlock;
 		struct wl_listener lock_destroy;
--- a/sway/lock.c
+++ b/sway/lock.c
@@ -17,9 +17,20 @@ struct sway_session_lock_surface {
 	struct wl_listener output_commit;
 };
 
+static void set_lock_focused_surface(struct wlr_surface *focused) {
+	server.session_lock.focused = focused;
+
+	struct sway_seat *seat;
+	wl_list_for_each(seat, &server.input->seats, link) {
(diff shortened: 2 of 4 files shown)


danieldg:
@Nefsen402: this should fix the keyboard focus problems you mentioned

sway/lock.c
```diff
@@ -48,6 +55,13 @@ static void handle_output_commit(struct wl_listener *listener, void *data) {
 
 static void handle_surface_destroy(struct wl_listener *listener, void *data) {
 	struct sway_session_lock_surface *surf = wl_container_of(listener, surf, destroy);
+
+	// Move the seat focus to another surface if one is available
```
emersion, line 59: Hm. I'm not a fan of this hack.

danieldg: Replaced with tracking the focus in server state; that also simplifies the input changes and improves the stability of focus on output creation.

Nefsen402:
Here these patches rebased on the scene graph PR: https://github.com/Nefsen402/sway/commit/e46d3c2d8b4ba44cae4687ddc743c024d36ed593

include/sway/server.h
```diff
@@ -95,6 +95,7 @@ struct sway_server {
 		struct wlr_session_lock_manager_v1 *manager;
 
 		struct wlr_session_lock_v1 *lock;
+		struct wlr_surface *focused;
```
emersion, line 98: Instead of having this new field, can we use `wlr_seat.keyboard_state.focused_surface`?

danieldg: Not as easily - that field is overwritten by `swaymsg focus` and a number of other events, so it would need to be saved and restored to avoid the focus wandering to non-lockscreen surfaces.

emersion: Hm, I see. Let's go with this then.

emersion (approved):
Thanks!

Issue #6973, opened 2022-05-02
Redundant splits are not flattened sometimes


KogasaPls:
- **Sway Version:**
sway version 1.8-dev-a7898637 (May  2 2022, branch 'master')

- **Debug Log:**
https://gist.github.com/KogasaPls/ff282d47d536a0f2028dbde20e17297c

- **Configuration File:**
Default configuration with `set $mod Alt+Mod4` instead of `set $mod Mod4`.

**Description:**
Let A be a container with 2 children, B and C, where C also has children. 

    Representation: A[B C[x]]

Killing B leaves A as a redundant container:

    Representation: A[C[x]]

In this case, we should flatten to `C[x]`.

This bug can lead to deeply nested layers of containers over time, incurring a heavy performance cost. For example, if you open a sequence of containers with alternating splits:

    H[app1 V[app2 H[app3 V[app4 ... H[appN]...]]]]

then close them in the same order you opened them, you'll end up with

    H[V[H[V[...H[appN]...]]]]

This is very common when using `autotiling` or `autotiling-rs`, but can happen without any external scripts.

**How to replicate:**
Launch Sway
Spawn a single container and splith it (`H[app1]`)
Spawn a new container (`H[app1 app2]`)
splitv the child (`H[app1 V[app2]]`)
kill the parent (`H[V[app2]]`)
The workspace is now in the layout `H[V[app2]]` instead of the functionally identical `V[app2]`

KogasaPls:
Closing as this is an i3 bug (https://github.com/i3/i3/issues/3503), Sway is correctly reproducing i3's bad behavior (see https://github.com/swaywm/sway/pull/6979#issuecomment-1119390056)

ammgws:
@KogasaPls Maybe we could re-open this? https://github.com/i3/i3/pull/5469

ammgws:
FYI new PR here: https://github.com/swaywm/sway/pull/7961

ryym:
Hi, I'm running into this issue too. Is there a plan to merge that pull request?

PR #6979, opened 2022-05-05
Check for redundant splits after killing container

KogasaPls:
Fixes https://github.com/swaywm/sway/issues/6973. When a view is killed, check if it makes a container redundant, then squash that container if so.


Xyene:
Thanks for sending this in, however, the current behavior is not a bug and is how i3 behaves. See also https://github.com/swaywm/sway/pull/6124

Issue #6978, opened 2022-05-04
sway-master crashes when focusing parent in specific layout


WhyNotHugo:
### **Sway Version:**
  - `sway version 1.8-dev-519038a7 (Apr 29 2022, branch 'master')`

### **Debug Log:**

https://gist.github.com/WhyNotHugo/80a6a7ea3a06d591706d8039733529b6

##### **Configuration File:**

```
exec alacritty
```

### **Stack Trace:**

See gist above for coredump and backtrace.

### **Description:**

- Start out with a single terminal window.
- Open another window next to it
- `swaymsg split vertical` on this second window
- Run a third terminal window
- `swaymsg focus parent` on this third window
- `SIGSEGV`!

I had this happen a few times the last several days, and finally found a pattern on how to reproduce it! Happens with a nested sway and an empty config too.

vyivel:
Fixed by https://github.com/swaywm/sway/pull/6969.

PR #6980, opened 2022-05-06
seat: fix focus policy logic

vyivel:
Fixes: https://github.com/swaywm/sway/issues/6978


emersion:
Seems like the function hasn't been renamed in the header. Also needs a rebase!

vyivel:
Actually, that's already fixed on master.

PR #6984, opened 2022-05-11
Replace strncpy with memcpy

emersion:
memcpy is useless here, is dangerous because it doesn't guarantee
that the string is NUL-terminated and causes the following warning:

    ../sway/criteria.c: In function ‘criteria_parse’:
    ../sway/criteria.c:712:25: error: ‘strncpy’ destination unchanged after copying no bytes [-Werror=stringop-truncation]
      712 |                         strncpy(value, valuestart, head - valuestart);
          |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -682,7 +682,7 @@ struct criteria *criteria_parse(char *raw, char **error_arg) {
 		}
 		name = calloc(head - namestart + 1, 1);
 		if (head != namestart) {
-			strncpy(name, namestart, head - namestart);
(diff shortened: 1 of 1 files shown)


mschilli87:
@emersion:
> memcpy is useless here

Did you mean to write `strncpy` here?

emersion:
Indeed! Fixed, thanks.

bl4ckb0ne (approved):
LGTM, thanks!

PR #6985, opened 2022-05-11
man: Add descriptions for `stacking` and `tabbed` layouts

### LINKED ISSUE #5918: Containers + layouts are undocumented (can we document them or link to i3?)
I had trouble understanding some of the behavior for the different layouts, and when a tabbed/stacked layout is applied to a subset of windows or to all the windows on the screen. After asking on #sway IRC, the only documentation known was i3 documentation.

Can we update the man pages etc to describe container behavior instead of just the commands, or (if the semantics are intentionally undocumented to symbolize that you're supposed to look at i3 docs) can we link to the i3 docs?

charlesrocket:
Resolves #5918

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -176,6 +176,12 @@ set|plus|minus|toggle <amount>
 *layout* default|splith|splitv|stacking|tabbed
 	Sets the layout mode of the focused container.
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

PR #6989, opened 2022-05-12
Allow official open source Nvidia drivers after recent release

ZTube:
Nvidia recently released the source code of their gpu kernel driver codebase starting from version R515 for _Turing_ and newer architectures. [Source](https://developer.nvidia.com/blog/nvidia-releases-open-source-gpu-kernel-modules/)
As these drivers are licensed under MIT/GPLv2 they should no longer be viewed as being proprietary by sway and thus be supported without additional command line flags.

This commit implements tests for distinguishing between the closed- and open source driver by querying `/sys/module/nvidia/taint` for the **P** flag to indicate a proprietary license. If the flag is found the function _detect_proprietary_ works as usual and requires one to use the _--unsupported-gpu_ flag. In case the **P** flag is not present the open source driver is in use and therefore _--unsupported-gpu_ is not necessary.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -58,7 +58,25 @@ void detect_proprietary(int allow_unsupported_gpu) {
 	size_t line_size = 0;
 	while (getline(&line, &line_size, f) != -1) {
 		if (strncmp(line, "nvidia ", 7) == 0) {
-			if (allow_unsupported_gpu) {
+			FILE *taint_info = fopen("/sys/module/nvidia/taint", "r");
+			int taints_kernel = 1;
+			if (taint_info) {
+				char *flags = NULL;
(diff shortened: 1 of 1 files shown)


emersion:
NACK. The OpenGL implementation is still closed source, and I don't want to receive bug reports about it.

kennylevinsen:
Just to clarify, the reasons that the MIT/GPLv2 nvidia driver remains unsupported are:
- The driver is not yet supported by an open user space driver stack (e.g. mesa), and is only supported by its proprietary user space counterpart
- The driver is officially only alpha quality for the limited desktop use-cases it supports
- The driver is not upstreamed yet

It seems that these will all improve, but it will take time.

PR #6990, opened 2022-05-12
Fix segfault when connecting keyboard/mouse

kdarkhan:
Sometimes `device->wlr_device->keyboard->keymap` is NULL.
When that happens, the call to `xkb_keymap_num_layouts` segfaults.
This adds a guard around the handling of `keymap` to prevent that.

This should resolve issues #6502 #6484


kdarkhan:
Closing the PR since there is already an existing one which I missed.

PR #6992, opened 2022-05-13
sway: request SCHED_RR via RTKit

Emantor:
This tries to connect to DBUS to request RTTimeUSecMax from RTKit and
than changes the scheduling policy of sway to SCHED_RR. I tested the
impact using hackbench, with this patch I don't see any cursor
stuttering while running the following hackbench command:
```
  $ hackbench -f 100 -l 100000 -s 100000
```
This also fixes the "your compositor is too slow"-warnings from
libinput.

The realtime feature is currently auto, so if one of the SD-Bus
libraries is installed we will select it. However we will fail
graciously with only a debug note if we can't reach RTKit and won't
touch RLIMIT_RTTIME in this case. Users who don't want to have DBUS
within sway itself can disable the feature.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -6,6 +6,7 @@ option('swaybar', type: 'boolean', value: true, description: 'Enable support for
 option('swaynag', type: 'boolean', value: true, description: 'Enable support for swaynag')
 option('xwayland', type: 'feature', value: 'auto', description: 'Enable support for X11 applications')
 option('tray', type: 'feature', value: 'auto', description: 'Enable support for swaybar tray')
+option('rtkit', type: 'feature', value: 'auto', description: 'Enable support to request realtime policy via RTKit')
 option('gdk-pixbuf', type: 'feature', value: 'auto', description: 'Enable support for more image formats in swaybg')
 option('man-pages', type: 'feature', value: 'auto', description: 'Generate and install man pages')
 option('sd-bus-provider', type: 'combo', choices: ['auto', 'libsystemd', 'libelogind', 'basu'], value: 'auto', description: 'Provider of the sd-bus library')
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -173,4 +173,6 @@ void handle_pointer_constraint(struct wl_listener *listener, void *data);
 void xdg_activation_v1_handle_request_activate(struct wl_listener *listener,
 	void *data);
 
+void gain_realtime(void);
+
 #endif
--- a/meson.build
+++ b/meson.build
@@ -106,6 +106,12 @@ if get_option('tray').enabled() and not tray_deps_found
 endif
 have_tray = (not get_option('tray').disabled()) and tray_deps_found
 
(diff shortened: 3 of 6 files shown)


meson_options.txt
```diff
@@ -6,6 +6,7 @@ option('swaybar', type: 'boolean', value: true, description: 'Enable support for
 option('swaynag', type: 'boolean', value: true, description: 'Enable support for swaynag')
 option('xwayland', type: 'feature', value: 'auto', description: 'Enable support for X11 applications')
 option('tray', type: 'feature', value: 'auto', description: 'Enable support for swaybar tray')
+option('realtime', type: 'feature', value: 'auto', description: 'Enable support to request realtime policy via RTKit')
```
emersion, line 9: I'd prefer to name this "rtkit"

emersion (commented):
Do we really need to use D-Bus here? `CAP_SYS_NICE` sounds fine to request packagers to set. `SCHED_RR | SCHED_RESET_ON_FORK` should be what we need I think?

emersion:
cc @kennylevinsen

Emantor:
Hm, freebsd doesn't know about `RLIMIT_RTTIME` either. The reason I chose rtkit here is that I thought I could provide seamless `SCHED_RR` functionality if rtkit is available, however since this doesn't work for freebsd we can probably go the setcap route.

Even with setcap and one of `pthread_setschedparam` or `sched_setscheduler` I fear that this won't be portable to freebsd. `SCHED_RESET_ON_FORK` is not available there which would require us to juggle our priorities around the start of other processes.
Unconveniently `SCHED_RESET_ON_FORK` is AFAIR behind a `_gnu` if, so we can't simply use it with the current POSIX defines.

IMO the rtkit approach is cleaner. For most desktop environments targeted by sway the user should be running it either way to provide realtime capabilities to pipewire or pulseaudio.

kennylevinsen:
I'd generally want to avoid introducing dbus into the sway main process.

Options:
1. CAP_SYS_NICE + sched_setscheduler/setpriority. What permissions are required on FreeBSD for this?
2. External client talking to rtkit. This has a catch: setrlimit needs to be below `org.freedesktop.RealtimeKit1.RTTimeUSecMax` to allow the priority to be set. This appears to be [hardcoded to 200ms](https://github.com/heftig/rtkit/blob/c295fa849f52b487be6433e69e08b46251950399/rtkit-daemon.c#L107), so maybe it'll be fine to just set a reasonably low default (50ms?).
3. Make a PAM module set things up for us. A little finicky as PAM login managers will fork once after pam_open_session and run the target session from there, so the module needs to set realtime scheduling on first fork and clear it on next/use RESET_ON_FORK.
4. Bite the bullet and have dbus in sway

The first one is the simplest and least hacky. As for clearing the scheduling priority on fork, pthread_atfork might do the trick there for platforms without SCHED_RESET_ON_FORK. Otherwise we can just make a small "reset_child" helper.

Emantor:
Superseded by #6994

Issue #3987, opened 2019-03-28
Disable ptrace by default


schauveau:
I am running sway from git master on a Debian/Buster machine so with systemd and not suid-root. 

I noticed that the sway process does not disallow ptrace by default. I feel that this is a serious security issue since keyloggers are trivial to implement by *ptracing* the read() calls on /dev/input/event? devices. 

Wouldn't it make sense to call `prctl(PR_SET_DUMPABLE, 0)` at startup and maybe to add a command line (or build) option to allow developers to make coredump and to debug?

emersion:
Keyloggers can override your $PATH and make you run a rogue version of sway anyway.

The better solution is to run untrusted clients in sandboxes, be it Flatpak or another mechanism.

schauveau:
Of course but is that a reason to make their life easy?

emersion:
You could argue all applications should disable ptrace. In which case *you* should disable ptrace locally.

It's not about making keyloggers' life easy. It's about not making us feel like we're secure with a half-baked "solution", and think about a real solution instead.

In other words, securing the compositor against processes running as the current user in the same namespace is not going to work no matter what you do. It's not worth it to try to do it.

schauveau:
You are probably right and on second though a proper sway keylogger could have some practical uses like for instance debugging input issues. That should make a nice small github project.

schauveau:
In case someone would have a use for it: https://github.com/schauveau/sway-keylogger

Emantor:
FWIW, there is also [ptrace scope](https://www.kernel.org/doc/Documentation/security/Yama.txt)

emersion:
>So, protecting a compositor against a trivial attack is pointless. It is a bit like locking your house which is stupid because burglars can easily break your doors and windows.

This is quite ridiculous. It seems you haven't understood at all the problem.

Now are you going to ask Firefox to disable ptrace? Because this is clearly where people are going to type in their credit card numbers and account passwords.

Next should we unset `$PATH` for better security? (We used to unset `$LD_PRELOAD` and `$LD_LIBRARY_PATH` in the past, it's quite similar)

Now come on, instead of spreading FUD please try to help with an actual solution.

emersion:
>That other keylogger intercepts the communications between the compositor and a wayland client using the LD_PRELOAD trick and it can only be applied to new applications. My approach is different since it directly observes the communications between the compositor and the Linux kernel.

[...]

PR #6994, opened 2022-05-14
realtime: request SCHED_RR using CAP_SYS_NICE

Emantor:
Try to gain SCHED_RR (round-robin) realtime scheduling privileges before
starting the server. This requires CAP_SYS_NICE on Linux systems.
We additionally register a pthread_atfork callback which resets the
scheduling class back to SCHED_OTHER (the Linux system default).

Due to CAP_SYS_NICE, setting RLIMIT_RTPRIO has no effect on the process
as documented within man 7 sched (from Linux):
```
  Privileged (CAP_SYS_NICE) threads ignore the RLIMIT_RTPRIO limit;
  as with older kernels, they can make arbitrary changes to
  scheduling policy and priority. See getrlimit(2) for further
  information on RLIMIT_RTPRIO
```
Note that this requires the sway distribution packagers to set the
CAP_SYS_NICE capability on the sway binary.

Supersedes #6992

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -413,6 +413,8 @@ int main(int argc, char **argv) {
 		goto shutdown;
 	}
 
+	set_rr_scheduling();
+
 	if (!server_start(&server)) {
 		sway_terminate(EXIT_FAILURE);
 		goto shutdown;
--- a/sway/realtime.c
+++ b/sway/realtime.c
@@ -1,0 +1,40 @@
+#include <sys/resource.h>
+#include <sched.h>
+#include <unistd.h>
+#include <pthread.h>
+#include "sway/server.h"
+#include "log.h"
+
+static void child_fork_callback(void) {
+	struct sched_param param;
+
+	param.sched_priority = 0;
+
+	int ret = pthread_setschedparam(pthread_self(), SCHED_OTHER, &param);
+	if (ret != 0) {
+		sway_log(SWAY_ERROR, "Failed to reset scheduler policy on fork");
+	}
+}
+
+void set_rr_scheduling(void) {
+	int prio = sched_get_priority_min(SCHED_RR);
+	int old_policy;
+	int ret;
+	struct sched_param param;
+
+	ret = pthread_getschedparam(pthread_self(), &old_policy, &param);
+	if (ret != 0) {
+		sway_log(SWAY_DEBUG, "Failed to get old scheduling priority");
+		return;
+	}
+
+	param.sched_priority = prio;
+
+	ret = pthread_setschedparam(pthread_self(), SCHED_RR, &param);
+	if (ret != 0) {
+		sway_log(SWAY_INFO, "Failed to set scheduling priority to %d", prio);
+		return;
+	}
+
+	pthread_atfork(NULL, NULL, child_fork_callback);
+}
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -173,4 +173,6 @@ void handle_pointer_constraint(struct wl_listener *listener, void *data);
 void xdg_activation_v1_handle_request_activate(struct wl_listener *listener,
 	void *data);
 
+void set_rr_scheduling(void);
+
 #endif
--- a/sway/meson.build
+++ b/sway/meson.build
@@ -7,6 +7,7 @@ sway_sources = files(
 	'ipc-server.c',
 	'lock.c',
 	'main.c',
+	'realtime.c',
 	'server.c',
 	'swaynag.c',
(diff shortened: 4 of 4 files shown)


kennylevinsen:
Hmm. So on FreeBSD, [this requires `PRIV_SCHED_SET`](https://github.com/freebsd/freebsd-src/blob/f210e4fbc54ab78711bd5197b7097002ff3ae891/sys/kern/p1003_1b.c#L240).

From what I can tell that is primarily given to root, but from reading the code it seems like `mac(9)` can interfere with these decisions. Maybe it can be used to grant the privilege? Might need a FreeBSD expert to comment...

Emantor:
Apparently this can be done using `mac(9)`,  but requires it's own kernel [module](https://mysteriouscode.com/blog/granting-privileges-using-mac-framework/) ([code](https://github.com/pbiernacki/mac_rtprio/)).

kennylevinsen:
... oh. rtkit isn't a thing on FreeBSD either from what I can tell, so maybe this is just not something people do over there. We fail gracefully here anyhow.

(As an aside, I recently noticed that [rtkit advices one to try `sched_setscheduler` before rtkit](https://github.com/heftig/rtkit/blob/master/README#L100), so we need the manual `sched_setscheduler` code anyhow.)

sway/realtime.c
```diff
@@ -0,0 +1,40 @@
...
+	param.sched_priority = 0;
+
+	int ret = pthread_setschedparam(pthread_self(), SCHED_OTHER, &param);
+	if (ret != 0) {
+		sway_log(SWAY_ERROR, "Failed to reset scheduler policy on fork");
+	}
+}
+
+void gain_realtime(void) {
+	int prio = sched_get_priority_min(SCHED_RR);
```
kennylevinsen, line 20: Hmm, so pipewire ends up having priority 9, but `sched_get_priority_min(SCHED_RR)` is -2 so we're below pipewire which is a little annoying. I wonder where the pipewire's priority comes from - rtkit seems to ask for priority 0?

sway/main.c
```diff
@@ -413,6 +413,8 @@ int main(int argc, char **argv) {
 		goto shutdown;
 	}
 
+	gain_realtime();
```
kennylevinsen, line 416: Bikeshed: Something like `set_proc_scheduling` or something else to be clearer that we're fixing process scheduling, whatever that requires.

sway/realtime.c
Emantor: Pipewire uses the `MaxRealtimePriority`property from RTKit  to request the highest possible priority from RTKit. This sounds fine and also sway requesting the lowest sounds good. According to `man 2 sched_get_priority_min`:
```
Processes  with  numerically  higher  priority  values are scheduled before processes with numerically lower priority values. Thus, the value returned by sched_get_priority_max() will be greater than the value returned by sched_get_priority_min().
```
So pipewire should be scheduled before sway.

sway/main.c
Emantor: Sounds good.

Emantor: I renamed this to `set_rr_scheduling`.

sway/realtime.c
kennylevinsen: > Pipewire uses the `MaxRealtimePriority` property from RTKit to request the highest possible priority from RTKit.

Oh, I just realized that I was checking prio, not rtprio. That explains why I was seeing weird numbers. D'oh.

> So pipewire should be scheduled before sway.

Yeah, my thought was that it makes sense that we as the graphical window system would have either same or higher priority than audio and screen capture routing. It does seem a bit aggressive from pipewire's perspective to ask for the highest priority, outranking even a bunch of kernel threads...

(This is of course nowhere near as important as just being SCHED_RR in the first place - even if our priority is wrong, we're competing with a much smaller group of processes.)

kennylevinsen:
Other than being unsure about the priority we should use, LGTM.

Emantor:
IMO, starting with the minimum priority is fine. We can still adjust this or expose this as a user configurable option later.

kennylevinsen (approved):

kennylevinsen:
Thanks!

shibe2:
It was proposed in #3987 to make Sway process non-dumpable on Linux. Well, if sway executable file has `CAP_SYS_NICE` and calling process doesn't have it (which non-root users normally don't have), then dumpable attribute for Sway is taken from `fs.suid_dumpable`, which is 0 by default. So if anyone wonders why Sway process isn't dumpable, this may be why.

PR #7007, opened 2022-05-16
sway/lock: set session lock hints if systemd is available

roosemberth:
## Motivation for this MR: 

User automation can benefit from knowing whether a session is locked. For example, automation can stop a time-tracking software when the screen is locked and prompt the user whether they'd like to continue the tracker when the session is unlocked.

systemd-logind provides support for this, but it needs to be notified about this hint.

## Why implement this in sway?

From the [The D-Bus interface of systemd-logind documentation](https://www.freedesktop.org/software/systemd/man/org.freedesktop.login1.html):

> SetLockedHint() may be used to set the "locked hint" to locked, i.e. information whether the session is locked. This is intended to be used by the desktop environment to tell systemd-logind when the session is locked and unlocked.

## Alternatives

- User automation could listen whether the screenlocker programs is started, the problem is that if the user has many sessions in different ttys, distinguishing which session is locked becomes hairy.
- User automation could listen whether the screenlocker program is started, but then it would need to know a fixed list of screenlocking programs to detect, this seems rather unefficient.
- The screen locker software could set the hint on behalf of the desktop environment, but if the screen locker is killed, the hint would stay, leading to potentially unintended behaviour.
- The screen locker software could set the hint on behalf of the desktop environment, but then this would need to be implemented in all screenlockers.

## Downsides

- This feature is only available with systemd (to my knowledge).

## Future work

- Check whether a similar interface is implemented in environments with `libelogind` or `basu`. I don't know these environments and I can't test, therefore this is only enabled on systemd.

### DIFF
--- a/sway/lock.c
+++ b/sway/lock.c
@@ -6,6 +6,11 @@
 #include "sway/output.h"
 #include "sway/server.h"
 
+#include "../include/config.h"
+#if HAVE_LIBSYSTEMD
+#include <systemd/sd-bus.h>
+#endif
+
 struct sway_session_lock_surface {
 	struct wlr_session_lock_surface_v1 *lock_surface;
 	struct sway_output *output;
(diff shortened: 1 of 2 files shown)


roosemberth:
Tested on NixOS 22.05.20220507.f8a27ed (Quokka) with the following dependencies:
- wlroots 0.16@19896e7fb63db73c4a66b68a79d7cbd039198a1d
- waylock 0.3.5@1d267c79236ea30bef1e9b93a39f88b5f9fa5b61

(Mind that wlroots master (8fe3aa29da56be16faa73aca947647bd60cd4a94) did not compile for me at the time due to an unrelated problem with `wlr_shell_create`.)

vyivel:
Thanks for your work. However, Sway does not explicitly support any *specific* init system, systemd included.

> Running Sway as a systemd service is not supported, nor recommended, nor required for anything. It may break your setup if you're not familiar enough with systemd.

— https://github.com/swaywm/sway/wiki/Systemd-integration

kennylevinsen:
If we need lock hooks, we should just have a form of event bind - one my event go as far as calling the command `bindevent` - so you can trigger something when lock/unlock happens.

(One could also go the route of having a Wayland protocol to monitor things, but maybe that overcomplicates things.)

Integrating directly with logind to mediate the hint means that FreeBSD and Linux users without (e)logind won't be able to benefit from the feature. On the other hand, logind integration can be built upon a generic solution if one feels that it is necessary.

roosemberth:
Closing this MR since a more generic solution that can be used by users in other platforms seems preferable.

/close

PR #7017, opened 2022-05-22
ext-session-lock: disable direct scan-out when locked

danieldg:
Fixes https://github.com/swaywm/swaylock/issues/236.

@alextsits - please confirm this fixes it for you.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -453,6 +453,10 @@ static bool scan_out_fullscreen_view(struct sway_output *output,
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


Emantor (approved):
Makes sense to me.

emersion:
I'd prefer to add the check in `scan_out_fullscreen_view`, which already contains similar checks for the overlay layer etc

alextsits:
@danieldg thanks for this, I confirm it fixes https://github.com/swaywm/swaylock/issues/236 for me.

emersion (approved):
Thanks!

PR #7023, opened 2022-05-24
Use shell for spawning swaybar

pogo5422:
This adds ability to set any command in `swaybar_command`, not just path to binary, which is [default behavior](https://github.com/i3/i3/blob/next/src/main.c#L1186-L1191) in `i3`.

### DIFF
--- a/sway/config/bar.c
+++ b/sway/config/bar.c
@@ -236,10 +236,14 @@ static void invoke_swaybar(struct bar_config *bar) {
 			setenv("WAYLAND_SOCKET", wayland_socket_str, true);
 
 			// run custom swaybar
-			char *const cmd[] = {
-					bar->swaybar_command ? bar->swaybar_command : "swaybar",
-					"-b", bar->id, NULL};
+			const char *swaybarcmd =
+				bar->swaybar_command ? bar->swaybar_command : "exec swaybar";
+			int cmdlen = snprintf(NULL, 0, "%s -b %s", swaybarcmd, bar->id) + 1;
+			char *full_cmd = malloc(cmdlen);
+			sprintf(full_cmd, "%s -b %s", swaybarcmd, bar->id);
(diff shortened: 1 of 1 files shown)


sway/config/bar.c
```diff
@@ -236,9 +236,13 @@ static void invoke_swaybar(struct bar_config *bar) {
...
 			// run custom swaybar
-			char *const cmd[] = {
-					bar->swaybar_command ? bar->swaybar_command : "swaybar",
-					"-b", bar->id, NULL};
+			const char *swaybarcmd =
+				bar->swaybar_command ? bar->swaybar_command : "exec swaybar";
+			int cmdlen = snprintf(NULL, 0, "%s -b %s", swaybarcmd, bar->id);
+			char *full_cmd = malloc(cmdlen);
+			sprintf(full_cmd, "%s -b %s", swaybarcmd, bar->id);
+			full_cmd[cmdlen] = '\0';
```
Nefsen402, line 244: You're corrupting the heap here. The malloc is not large enough for the null byte here.

sway/config/bar.c
```diff
@@ -236,9 +236,13 @@ static void invoke_swaybar(struct bar_config *bar) {
 			setenv("WAYLAND_SOCKET", wayland_socket_str, true);
 
 			// run custom swaybar
-			char *const cmd[] = {
-					bar->swaybar_command ? bar->swaybar_command : "swaybar",
-					"-b", bar->id, NULL};
+			const char *swaybarcmd =
+				bar->swaybar_command ? bar->swaybar_command : "exec swaybar";
+			int cmdlen = snprintf(NULL, 0, "%s -b %s", swaybarcmd, bar->id);
+			char *full_cmd = malloc(cmdlen);
```
Nefsen402, line 242: I don't see a free for this malloc

PR #7033, opened 2022-05-27
chore: chase wlroots xdg-shell update

vyivel:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3514


emersion (approved):
Thanks!

PR #7038, opened 2022-05-30
build: link with -pthread

emersion:
Fixes the following FreeBSD error:

    ld: error: undefined symbol: pthread_getschedparam
    >>> referenced by realtime.c:25 (../sway/realtime.c:25)
    >>>               sway/sway.p/realtime.c.o:(set_rr_scheduling)

Fixes: a3a82efbf6b5 ("realtime: request SCHED_RR using CAP_SYS_NICE")

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -70,6 +70,7 @@ fish_comp = dependency('fish', required: false)
 math = cc.find_library('m')
 rt = cc.find_library('rt')
 xcb_icccm = dependency('xcb-icccm', required: get_option('xwayland'))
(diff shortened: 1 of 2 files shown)


kennylevinsen (approved):
LGTM - I wonder if we should just have used the non-pthread versions, but then we have to live without pthread_atfork.

emersion:
Also I'm not sure we want to change the priority of e.g. Mesa threads. Maybe, but maybe not, it depends what they do.

PR #7039, opened 2022-05-30
Improve tablet tool button handling.

### LINKED ISSUE #7036: Graphics tablet tool buttons stop working correctly after a while.
I plan to try to resolve this issue, I just wanted to document my thought process for posterity sake.

This is observed on `sway version 1.6.1` on NixOS, using a WACOM PTH-660 (Intuos Pro):

```
Bus 003 Device 007: ID 056a:0357 Wacom Co., Ltd PTH-660 [Intuos Pro (M)]
```

I'm pretty sure, however, that it's applicable to _all_ graphics tablets, since the advent of graphics tablet support. It was caused in my original PR (which, honestly, has proven to be very shoddy&mdash;sorry about that.) I strongly believe there are no relevant log messages, so I will not include any logs.

## Analyzing the Problem

The suspect code is [here](https://github.com/swaywm/sway/blob/440d0bc/sway/input/cursor.c#L821):

```c
	if (!surface || !wlr_surface_accepts_tablet_v2(tablet_v2, surface)) {
		// TODO: the user may want to configure which tool buttons are mapped to
		// which simulated pointer buttons
		switch (event->state) {
		case WLR_BUTTON_PRESSED:
			if (cursor->tool_buttons == 0) {
				dispatch_cursor_button(cursor, &event->tablet->base,
						event->time_msec, BTN_RIGHT, event->state);
			}
			cursor->tool_buttons++;
			break;
		case WLR_BUTTON_RELEASED:
			if (cursor->tool_buttons == 1) {
				dispatch_cursor_button(cursor, &event->tablet->base,
						event->time_msec, BTN_RIGHT, event->state);
			}
			cursor->tool_buttons--;
			break;
		}
		wlr_seat_pointer_notify_frame(cursor->seat->wlr_seat);
		return;
	}
```

My reading of the code is as follows:

- If there is no surface found at the given cursor's coordinates, _or_ if the surface found at the given cursor's coordinates does not accept tablet v2 events, we fall back to dispatching right click events for all tool button events.
- In order to keep things consistent, we keep a count of how many tool buttons are currently pressed, `tool_buttons`. That way, we can avoid generating multiple events when multiple tool buttons are pressed at once.
- If the count falls out of sync, we practically stop generating events. For example, if we manage to get two `WLR_BUTTON_RELEASED` events after just one `WLR_BUTTON_PRESSED` event, we wind up with a `tool_buttons` value of `0xffffffff`. If the events stay consistent after that, and you only press one tool button at a time, events will never get generated.
- In theory, the count can't fall out of sync as long as this can never happen.

**However**, there are a couple of glaring problems that I'm not sure how I didn't notice when I introduced this bug:

1. One interesting problem is that this, I believe, effectively breaks resizing surfaces that _do_ handle tablet v2 events, because the cursor will _never_ generate `BTN_RIGHT` events when on top of the surface in question.
2. The `tool_buttons` count can fall out of sync in _numerous_ ways. I'm sure it's plainly obvious to anyone reading, but the most trivial way for it to fall out of sync is for the condition to become false in between `WLR_BUTTON_PRESSED` and `WLR_BUTTON_RELEASED`. So even if we receive a consistent stream of events, it can still fall out of sync due to this conditional. (The logic was correct prior to my PR, #4570.)

## Solutions

In order to try to resolve the problem, I made a quick patch to my local Sway derivation that looks like this:

```diff
diff --git a/sway/input/cursor.c b/sway/input/cursor.c
index c89d5166..59d52a8a 100644
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -795,23 +795,21 @@ static void handle_tool_button(struct wl_listener *listener, void *data) {
 	node_at_coords(cursor->seat, cursor->cursor->x, cursor->cursor->y,
 		&surface, &sx, &sy);
 
-	if (!surface || !wlr_surface_accepts_tablet_v2(tablet_v2, surface)) {
+	struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(cursor->seat->wlr_seat);
+	uint32_t modifiers = keyboard ? wlr_keyboard_get_modifiers(keyboard) : 0;
+	bool mod_pressed = modifiers & config->floating_mod;
+
+	if (!surface || !wlr_surface_accepts_tablet_v2(tablet_v2, surface) || mod_pressed) {
 		// TODO: the user may want to configure which tool buttons are mapped to
 		// which simulated pointer buttons
 		switch (event->state) {
 		case WLR_BUTTON_PRESSED:
-			if (cursor->tool_buttons == 0) {
-				dispatch_cursor_button(cursor, event->device,
-						event->time_msec, BTN_RIGHT, event->state);
-			}
-			cursor->tool_buttons++;
+			dispatch_cursor_button(cursor, event->device,
+					event->time_msec, BTN_RIGHT, event->state);
 			break;
 		case WLR_BUTTON_RELEASED:
-			if (cursor->tool_buttons == 1) {
-				dispatch_cursor_button(cursor, event->device,
-						event->time_msec, BTN_RIGHT, event->state);
-			}
-			cursor->tool_buttons--;
+			dispatch_cursor_button(cursor, event->device,
+				event->time_msec, BTN_RIGHT, event->state);
 			break;
 		}
 		wlr_seat_pointer_notify_frame(cursor->seat->wlr_seat);
```

This attempt is very rough and was mainly just a one-off shot in the dark, but it _seems_ to do exactly what I wanted, though it has a couple of issues:

- This way of handling events makes no attempts to smooth over the case where multiple tool buttons are held down.
- If you let go of the modifier key first while on top of a tablet v2 capable surface, you will be stuck in resize mode as it will stop emulating mouse events.

Therefore, I think a more robust way to handle it would be to adjust the conditional _just a bit more_: In addition to turning on mouse emulation for tool buttons when holding down the floating mod key, we should also keep mouse emulation on until `tool_buttons` reaches zero. In addition, instead of allowing the `tool_buttons` value to underflow, we should log a warning and refuse to lower the `tool_buttons` value if `WLR_BUTTON_RELEASED` occurs when `tool_buttons` is already zero, probably in addition to dispatching the cursor button release. While this only handles half of the cases, this should be able to provide some diagnostics should there be a similar logic error introduced in the future, or issues with events being inconsistent.

(edit: It may also be worth refactoring the floating resize events so that they can simply understand the native tablet events instead, however, I think that this is worth holding off for a future enhancement. For now, simply enabling mouse click emulation is a relatively safe solution.)

jchv:
This change allows the tablet tool button to be used for floating mod resize. In addition, it attempts to ensure that tablet tool events are consistent such that tablet v2 events and pointer events will never be interleaved, and such that the tool buttons count will never fall out of sync and cause tool button emulation to break.

Some of this logic is similar to what is done for tablet tool tip, but not quite identical, because of the complication that we have to deal with multiple inputs that can overlap each-other.

The logic described to fix this in #7036 is not sufficient as it does not account for the inverse case (going from tablet events to pointer events.) Implementing the fix this way leads to much better behavior if you, for example, use a tool button to open a menu, then before releasing the tool button, hold the modifier key down.

(Note: I'm currently unable to compile SwayWM from `master` as I do not have the ability to run `meson>=0.60` right now, among other deps. So although I made a best effort to ensure this patch is right for SwayWM `master`, I could not actually test this yet on my own setup. Sorry.)

Fixes #7036.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -818,7 +818,34 @@ static void handle_tool_button(struct wl_listener *listener, void *data) {
 	node_at_coords(cursor->seat, cursor->cursor->x, cursor->cursor->y,
 		&surface, &sx, &sy);
 
-	if (!surface || !wlr_surface_accepts_tablet_v2(tablet_v2, surface)) {
+	// TODO: floating resize should support graphics tablet events
+	struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(cursor->seat->wlr_seat);
+	uint32_t modifiers = keyboard ? wlr_keyboard_get_modifiers(keyboard) : 0;
+	bool mod_pressed = modifiers & config->floating_mod;
+
+	bool surface_supports_tablet_events =
+		surface && wlr_surface_accepts_tablet_v2(tablet_v2, surface);
+
+	// Simulate pointer when:
+	// 1. The modifier key is pressed, OR
+	// 2. The surface under the cursor does not support tablet events.
+	bool should_simulate_pointer = mod_pressed || !surface_supports_tablet_events;
+
+	// Similar to tool tip, we need to selectively simulate mouse events, but we
+	// want to make sure that it is always consistent. Because all tool buttons
+	// currently map to BTN_RIGHT, we need to keep count of how many tool
+	// buttons are currently pressed down so we can send consistent events.
+	//
+	// The logic follows:
+	// - If we are already simulating the pointer, we should continue to do so
+	//   until at least no tool button is held down.
+	// - If we should simulate the pointer and no tool button is currently held
+	//   down, begin simulating the pointer.
+	// - If neither of the above are true, send the tablet events.
+	if ((cursor->tool_buttons > 0 && cursor->simulating_pointer_from_tool_button)
+		|| (cursor->tool_buttons == 0 && should_simulate_pointer)) {
+		cursor->simulating_pointer_from_tool_button = true;
+
 		// TODO: the user may want to configure which tool buttons are mapped to
 		// which simulated pointer buttons
 		switch (event->state) {
@@ -827,22 +854,35 @@ static void handle_tool_button(struct wl_listener *listener, void *data) {
 				dispatch_cursor_button(cursor, &event->tablet->base,
 						event->time_msec, BTN_RIGHT, event->state);
 			}
-			cursor->tool_buttons++;
 			break;
 		case WLR_BUTTON_RELEASED:
-			if (cursor->tool_buttons == 1) {
+			if (cursor->tool_buttons <= 1) {
 				dispatch_cursor_button(cursor, &event->tablet->base,
 						event->time_msec, BTN_RIGHT, event->state);
 			}
-			cursor->tool_buttons--;
 			break;
 		}
 		wlr_seat_pointer_notify_frame(cursor->seat->wlr_seat);
-		return;
+	} else {
+		cursor->simulating_pointer_from_tool_button = false;
+
+		wlr_tablet_v2_tablet_tool_notify_button(sway_tool->tablet_v2_tool,
+			event->button, (enum zwp_tablet_pad_v2_button_state)event->state);
 	}
 
-	wlr_tablet_v2_tablet_tool_notify_button(sway_tool->tablet_v2_tool,
-		event->button, (enum zwp_tablet_pad_v2_button_state)event->state);
+	// Update tool button count.
+	switch (event->state) {
+	case WLR_BUTTON_PRESSED:
+		cursor->tool_buttons++;
+		break;
+	case WLR_BUTTON_RELEASED:
+		if (cursor->tool_buttons == 0) {
+			sway_log(SWAY_ERROR, "inconsistent tablet tool button events");
+		} else {
+			cursor->tool_buttons--;
+		}
+		break;
+	}
 }
 
 static void check_constraint_region(struct sway_cursor *cursor) {
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -64,6 +64,7 @@ struct sway_cursor {
 	struct wl_listener tool_proximity;
 	struct wl_listener tool_button;
 	bool simulating_pointer_from_tool_tip;
+	bool simulating_pointer_from_tool_button;
 	uint32_t tool_buttons;
 
 	struct wl_listener request_set_cursor;


jchv:
I'm sure bandwidth is limited, but I'd love to get this good for merge since it is quite a massive quality-of-life improvement for using SwayWM with graphics tablets. Is there anything in particular I can do to help here? I just rebased since it has been a couple months.

sway/input/cursor.c
```diff
@@ -818,7 +818,34 @@ static void handle_tool_button(struct wl_listener *listener, void *data) {
...
+	struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(cursor->seat->wlr_seat);
+	uint32_t modifiers = keyboard ? wlr_keyboard_get_modifiers(keyboard) : 0;
+	bool mod_pressed = modifiers & config->floating_mod;
+
+	bool surface_supports_tablet_events =
+		surface && wlr_surface_accepts_tablet_v2(tablet_v2, surface);
+
+	// Simulate pointer when:
+	// 1. The modifier key is pressed, OR
+	// 2. The surace under the cursor does not support tablet events.
```
Xyene, line 831:
```suggestion
	// 2. The surface under the cursor does not support tablet events.
```

Small nit

sway/input/cursor.c
```diff
@@ -827,22 +854,38 @@ static void handle_tool_button(struct wl_listener *listener, void *data) {
...
 
-	wlr_tablet_v2_tablet_tool_notify_button(sway_tool->tablet_v2_tool,
-		event->button, (enum zwp_tablet_pad_v2_button_state)event->state);
+	// Update tool button count.
+	switch (event->state) {
+	case WLR_BUTTON_PRESSED:
+		cursor->tool_buttons++;
+		break;
+	case WLR_BUTTON_RELEASED:
+		if (cursor->tool_buttons == 0) {
```
Xyene, line 879: To make this logic a little easier to read, can we write it as

```c
if (cursor->tool_buttons == 0) {
   ...
} else if (cursor->tool_buttons <= 1) {
   ...
} else {
   ...
}
```

I think writing it this way also makes the `<= 1` case impossible to ever be `<` (because we won't be decrementing when `cursor->tool_buttons = 0` to get into this branch in the first place)?

In which case we might just need

```c
if (cursor->tool_buttons == 0) {
    sway_log(SWAY_ERROR, "inconsistent tablet tool button events");
} else {
    cursor->tool_buttons--;
}
```

What do you think?

Xyene (commented):
Thanks for working on this. I just gave it a whirl, and can confirm it mostly works.

One case that seems to not be handled is when starting with a layout like this (floating `gedit` on top of Firefox):

image

and then beginning a tool button mod resize from the bottom right of the `gedit` window. Things work fine if you resize slowly, but if you do so quickly (such that the cursor leaves the `gedit` window before `gedit` has a chance to resize its surface to under it), Firefox appears to receive the event and the resize is cancelled.

sway/input/cursor.c
jchv: Good catch, fixed now.

sway/input/cursor.c
jchv: Ahh, this feels obvious now that you point it out. Fixed.

jchv:
Thanks for taking a look! I've been a little busy myself lately, coincidentally. I've integrated the suggestions.

As for the bug with floating windows, I am oddly unable to reproduce this even if I get the cursor clearly off of the window. I'm actually not entirely sure _why_ this works: I'd assume if `node_at_coords` changed mid-resize, it would break. I tried both XWayland windows and native windows on top of each-other, nothing breaks. It also doesn't break if I try to resize it too small and drag the cursor entirely past the bounds of the window. I wonder what I'm missing....

Xyene:
If I had to guess, it's because https://github.com/swaywm/sway/blob/master/sway/input/seatop_resize_floating.c is unimplemented for tablets, so things only work if under pointer emulation. See https://github.com/swaywm/sway/blob/master/sway/input/seatop_move_floating.c for reference (moving floating windows with the tablet works). Being in `seatop_resize_floating` should prevent focus from changing until the tool is released.

jchv:
Yeah, that's why this PR forces pointer emulation when resizing. I'll try to take another look, but even using gedit and Firefox I can't get the same behavior—it's working as expected when I try.

jchv:
@Xyene I've messed around trying to figure out how to reproduce the issue you're having, but I'm just unable to. I've tried with gedit and Firefox in XWayland, gedit and Firefox in xdg_shell, tried the reverse, tried releasing mod while moving the cursor... I'm interested in fixing the issue, but I simply can't figure out what's going on. Do you have any ideas? Maybe your setup differs from mine in some fundamental way.

jchv:
New year, new rebase. Hello again.

I'm still daily-driving this patch. I still can't reproduce the issue with resizing. That said, especially since this fixes the issue where tool buttons stop entirely, I'd really like it if we could make some progress here somehow...

Of course I'd love to address the problems. I've tried to reproduce them to no avail. On my end, no matter what combination of floating windows, where I start the drag, or how violently I move the cursor, the resize doesn't break for me when using this patch. I can only reproduce that _without_ this patch. If there's more information that might enable me to reproduce an issue, I'd love to hear it.

That said, even if there was an issue with floating resize, I think the status quo in this PR is still a lot better than on master and in 1.8. I'd love to remove this from my local patchset by the time 1.9 comes out, and bring a much better graphics tablet experience to other Sway users who don't run patches. Can we make it happen?

Xyene (approved):
LGTM, thanks and sorry for the late reply. It's possible the floating resize thing is something buggy with my specific tablet model, so let's just merge this as-is.

jchv:
No problem, thanks for taking another look. I would like to still keep working on graphics tablet robustness, so I hope we can figure it out eventually. If it's happening for you, I'm sure it's going to happen for others. For now though, I'm just glad that 1.9 should ship without the tablet tool button bug. 👍

PR #7041, opened 2022-05-31
Enable security-context-v1

emersion:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3589


markstos:
What the security context feature does: "This lets sandbox engines such as Flatpak attach a security context to sandboxed clients. The compositor can then restrict which features are made available to that client."

PR #7044, opened 2022-06-01
Refuse to start when SUID is detected

kennylevinsen:
This ensures that those surprised by the deprecation of SUID operation
receive an error rather than accidentally having sway run as root.

This detection will be removed in a future release.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -150,27 +150,17 @@ static void log_kernel(void) {
 	pclose(f);
 }
 
-
-static bool drop_permissions(void) {
-	if (getuid() != geteuid() || getgid() != getegid()) {
-		sway_log(SWAY_ERROR, "!!! DEPRECATION WARNING: "
-			"SUID privilege drop will be removed in a future release, please migrate to seatd-launch");
-
-		// Set the gid and uid in the correct order.
-		if (setgid(getgid()) != 0) {
-			sway_log(SWAY_ERROR, "Unable to drop root group, refusing to start");
-			return false;
-		}
-		if (setuid(getuid()) != 0) {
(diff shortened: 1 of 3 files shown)


sway/main.c
```diff
@@ -150,27 +150,17 @@ static void log_kernel(void) {
...
-		if (setgid(getgid()) != 0) {
-			sway_log(SWAY_ERROR, "Unable to drop root group, refusing to start");
-			return false;
-		}
-		if (setuid(getuid()) != 0) {
-			sway_log(SWAY_ERROR, "Unable to drop root user, refusing to start");
-			return false;
-		}
+static bool detect_suid(void) {
+	if (geteuid() != 0) {
```
emersion, line 154: Do we need to check for `geteguid` as well?

emersion (commented):
Maybe it's time to rename `server_privileged_prepare`?

kennylevinsen:
> Maybe it's time to rename `server_privileged_prepare`?

The split into prepare and init was pretty redundant now, so I merged prepare into `server_init`.

sway/main.c
kennylevinsen: I assume you meant `getegid`? I added a check for that.

emersion: Ah yep!

emersion (approved):
LGTM, thanks!

psusi2:
Why would you refuse to start if suid rather than do whatever privileged things you need to do ( such as setting realtime scheduling priority ), then drop privileges?  Not everyone can enable CAP_SYS_NICE with an EA.

kennylevinsen:
We do not want SUID code as it is very hard to write safely and maintain.

> Not everyone can enable CAP_SYS_NICE with an EA.

SUID is far too big a hammer for this.

PR #7051, opened 2022-06-05
swaynag: move close_button up to fix SIGSEGV

greg904:
When swaynag_parse_options encounters '--dismiss-button' (or its
shorthand '-s'), it sets the text of the first button in the
swaynag.buttons list, which is expected to exist and to be the dismiss
button, to the one passed by the user.

Commit 4780afb68b4ee2cdf0e4925f40cf885819f8a74a ("swaynag: statically
allocate button_close, and move declaration") moved the list
initialization to after swaynag_parse_options is called which made that
code fail.

For example, the command 'swaynag --dismiss-button Dismiss' crashes and
'swaynag --message Message --button Yes "" --dismiss-button Dismiss'
shows the wrong buttons.

Move it back to before swaynag_parse_options is called.

### DIFF
--- a/swaynag/main.c
+++ b/swaynag/main.c
@@ -29,6 +29,11 @@ int main(int argc, char **argv) {
 	wl_list_init(&swaynag.outputs);
 	wl_list_init(&swaynag.seats);
 
+	struct swaynag_button button_close = { 0 };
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

PR #7054, opened 2022-06-06
contrib/autoname-workspaces: bugfix

hxr404:
Fix a bug in the contrib script autoname-workspaces.py + add argument to remove trailing whitespace


hxr404:
Sorry for all the force pushing btw. I should've used the usual (manual git) way for syncing my fork right-away instead of trying GitHub's built-in feature...

PR #7061, opened 2022-06-10
swaymsg: show non-desktop outputs

ForTheReallys:
Currently, `swaymsg -t get_outputs` doesn't report VR headsets.  This ultimately adds that support by adding non-desktop outputs to a list of the new type `sway_output_non_desktop` to `sway_root` since sway doesn't hold on to those outputs. I could simply create `sway_outputs` for non-desktop outputs, but I wasn't sure of the implications.  Let me know your thoughts

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -57,6 +57,12 @@ struct sway_output {
 	struct wl_event_source *repaint_timer;
 };
 
+struct sway_output_non_desktop {
+	struct wlr_output *wlr_output;
+
+	struct wl_listener destroy;
+};
+
 struct sway_output *output_create(struct wlr_output *wlr_output);
 
 void output_destroy(struct sway_output *output);
@@ -176,5 +182,7 @@ void handle_output_manager_test(struct wl_listener *listener, void *data);
 
 void handle_output_power_manager_set_mode(struct wl_listener *listener,
 	void *data);
+
+struct sway_output_non_desktop *output_non_desktop_create(struct wlr_output *wlr_output);
 
 #endif
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -240,10 +240,7 @@ static json_object *ipc_json_create_node(int id, const char* type, char *name,
 	return object;
 }
 
-static void ipc_json_describe_output(struct sway_output *output,
-		json_object *object) {
-	struct wlr_output *wlr_output = output->wlr_output;
-
+static void ipc_json_describe_wlr_output(struct wlr_output *wlr_output, json_object *object) {
 	json_object_object_add(object, "primary", json_object_new_boolean(false));
 	json_object_object_add(object, "make",
 			json_object_new_string(wlr_output->make ? wlr_output->make : "Unknown"));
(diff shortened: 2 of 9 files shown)


include/sway/output.h
```diff
@@ -177,4 +183,7 @@ void handle_output_manager_test(struct wl_listener *listener, void *data);
 void handle_output_power_manager_set_mode(struct wl_listener *listener,
 	void *data);
 
+// Non-desktop outputs
```
bl4ckb0ne, line 186: Useless comment

sway/tree/output.c
```diff
@@ -390,6 +391,34 @@ void output_get_box(struct sway_output *output, struct wlr_box *box) {
...
+
+	sway_log(SWAY_DEBUG, "Destroying non-desktop output '%s'", output->wlr_output->name);
+
+	int index = list_find(root->non_desktop_outputs, output);
+	list_del(root->non_desktop_outputs, index);
+
+	wl_list_remove(&output->destroy.link);
+
+	output->wlr_output->data = NULL;
+	output->wlr_output = NULL;
```
bl4ckb0ne, line 406: You forgot to free the output here, it should replace the need for those 2 lines.

sway/ipc-json.c
```diff
@@ -370,6 +374,21 @@ json_object *ipc_json_describe_disabled_output(struct sway_output *output) {
...
+	struct wlr_output *wlr_output = output->wlr_output;
+
+	json_object *object = json_object_new_object();
+
+	ipc_json_describe_wlr_output(wlr_output, object);
+
+	json_object_object_add(object, "non_desktop", json_object_new_boolean(true));
+	json_object_object_add(object, "type", json_object_new_string("output"));
+	json_object_object_add(object, "name",
+			json_object_new_string(wlr_output->name));
```
bl4ckb0ne, line 387: Nit: indend is only 2 tabs instead of 3

sway/tree/output.c
ForTheReallys: Good catch

bl4ckb0ne:
Can you split/squash the last commit into their respective commits please?

sway/tree/output.c
```diff
@@ -390,6 +391,33 @@ void output_get_box(struct sway_output *output, struct wlr_box *box) {
 	box->height = output->height;
 }
 
+static void handle_destroy_non_desktop(struct wl_listener *listener, void *data) {
+	struct sway_output_non_desktop *output =
+		wl_container_of(listener, output, destroy);
+
+	sway_log(SWAY_DEBUG, "Destroying non-desktop output '%s'", output->wlr_output->name);
+
```
bl4ckb0ne, line 399: I noticed that we never withdraw outputs, there should be a call to `wlr_drm_lease_v1_manager_withdraw_output` here to notify the clients if the output is leased.

emersion: Doesn't `wlr_drm_lease_v1_manager` listen for output destroy already?

bl4ckb0ne: Ah yeah, nevermind then.

bl4ckb0ne (approved):
LGTM, thanks!

PR #7064, opened 2022-06-13
seat: Fix reloading cursor theme after change

### LINKED ISSUE #6931: Changing mouse cursor theme using swaymsg has no effect
Sway version 1.7

I am trying to change the mouse cursor theme on a running instance of Sway. After some Googling, this seems to be the way to go:

```
$ swaymsg seat seat0 xcursor_theme Adwaita
```

However, this has no effect on a running instance of Sway. The mouse cursor theme on the desktop background and on window title bars remains the same.

However, if I put the lines in my Sway config:

```
seat seat0 xcursor_theme Adwaita
```

And reload my config using `$ swaymsg reload`, the cursor theme is set correctly.

Can the mouse cursor theme of Sway not be changed without restarting or reloading Sway's config?

nemuibanila:
This reverts commit afde6369
"seat: avoid unneeded reloading xcursor theme".
Always avoiding to reload the xcursor theme prevents reloading the
cursor even when this is desired. Instead seat_configure_xcursor
can determine whether a full reload is necessary.

To stay with the spirit of the reverted change, cursors are only fully
reloaded, if the theme has changed.

Fixes #6931

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -803,9 +803,7 @@ static void seat_apply_input_mapping(struct sway_seat *seat,
 
 static void seat_configure_pointer(struct sway_seat *seat,
 		struct sway_seat_device *sway_device) {
-	if ((seat->wlr_seat->capabilities & WL_SEAT_CAPABILITY_POINTER) == 0) {
-		seat_configure_xcursor(seat);
-	}
+	seat_configure_xcursor(seat);
 	wlr_cursor_attach_input_device(seat->cursor->cursor,
 		sway_device->input_device->wlr_device);
 	wl_event_source_timer_update(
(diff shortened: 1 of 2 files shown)


MithicSpirit:
Sorry for the bump but is there anything holding this from being merged? I am also being negatively affected by #6931 and would like to see it resolved.

The failed job seems to be due to an unchanged and unrelated file (ipc-json.c).

sway/input/seat.c
```diff
@@ -805,9 +805,7 @@ static void seat_apply_input_config(struct sway_seat *seat,
 
 static void seat_configure_pointer(struct sway_seat *seat,
 		struct sway_seat_device *sway_device) {
-	if ((seat->wlr_seat->capabilities & WL_SEAT_CAPABILITY_POINTER) == 0) {
-		seat_configure_xcursor(seat);
-	}
+	seat_configure_xcursor(seat);
```
emersion, line 810: `seat_configure_tablet_tool()` needs an update as well

sway/input/seat.c
```diff
@@ -1056,26 +1054,27 @@ void seat_configure_xcursor(struct sway_seat *seat) {
...
-	}
 
-	// Reset the cursor so that we apply it to outputs that just appeared
-	cursor_set_image(seat->cursor, NULL, NULL);
-	cursor_set_image(seat->cursor, "left_ptr", NULL);
-	wlr_cursor_warp(seat->cursor->cursor, NULL, seat->cursor->cursor->x,
-		seat->cursor->cursor->y);
+		// Reset the cursor so that we apply it to outputs that just appeared
+		cursor_set_image(seat->cursor, NULL, NULL);
+		cursor_set_image(seat->cursor, "left_ptr", NULL);
```
emersion, line 1083: Unrelated to this change, but would be nicer to handle this inside `wlr_cursor` instead of having this workaround...

GreyXor:
I'm interested in this fix to be able to change the cursor in relation to `prefers-color-scheme`.
If `prefer-dark` then my cursor is `light`, if `prefer-light`, then my cursor is `dark`.

@nemuibanila Gentle ping, this fix will allow me to accomplish this behavior, do you have any news? thanks :)

GreyXor:
@emersion Hello, do we miss something here for a review ? Thanks!

emersion (approved):
I think ideally we'd only reload the cursor theme once instead of each time we configure an input device… But this PR sounds like a step in the right direction regardless. Thanks!

PR #7067, opened 2022-06-15
Allocate enough space for `cmd_results->error`

kraftwerk28:
Sometimes (i.e. https://github.com/swaywm/sway/blob/master/sway/commands/move.c#L980-L990), 256 bytes is not enough to hold an error message.


emersion (approved):
Oh, yeah, calling `va_start` twice will work.

LGTM, thanks!

PR #7071, opened 2022-06-17
sway: prevent layer-shell-surfaces from receiving input when it is inhibited

### LINKED ISSUE #6800: security: rofi stays visible on top of screen locker
- **Sway Version:** sway version 1.8-dev-cd1ee0e1 (Jan 29 2022, branch 'master'

- **Description:**
If you leave rofi window open, then let swayidle lock the screen (I'm using swaylock), the rofi window stays visible and usable, over the lockscreen. This can be potentially be very dangerous, as raw commands from rofi are executed without any issue.

Molytho:
seat_set_focus_layer is missing an inhibition check allowing layer-shell surfaces to be focused even when it is inhibited.
layer-shell surfaces also don't lose keyboard focus in seat_set_exclusive_client.
This breaks assumptions made by input_inhibitor clients.

Partially fixes #6800 (it is shown but doesn't take input anymore)

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1326,7 +1326,7 @@ void seat_set_focus_layer(struct sway_seat *seat,
 			seat_set_focus(seat, previous);
 		}
 		return;
-	} else if (!layer || seat->focused_layer == layer) {
+	} else if (!layer || seat->focused_layer == layer || !seat_is_input_allowed(seat, layer->surface)) {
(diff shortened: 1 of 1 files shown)


emersion:
input-inhibitor is getting deprecated, replaced by ext-session-lock.

emersion:
input-inhibitor has been dropped.

PR #7083, opened 2022-06-22
Rename dpms output command to power

emersion:
The "dpms" command refers to VESA Display Power Management
Signaling, a deprecated standard. It's superseded by VESA DPM.

Instead of tying out command name to a particular standard, use the
neutral term "power".

### DIFF
--- a/sway/commands/output/dpms.c
+++ b/sway/commands/output/dpms.c
@@ -1,45 +1,8 @@
+#include "log.h"
 #include "sway/commands.h"
-#include "sway/config.h"
-#include "sway/output.h"
-#include "util.h"
-#include <strings.h>
 
 struct cmd_results *output_cmd_dpms(int argc, char **argv) {
-	if (!config->handler_context.output_config) {
-		return cmd_results_new(CMD_FAILURE, "Missing output config");
-	}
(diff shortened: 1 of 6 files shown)


sway/commands/output/dpms.c
```diff
@@ -1,45 +1,5 @@
 #include "sway/commands.h"
-#include "sway/config.h"
-#include "sway/output.h"
-#include "util.h"
-#include <strings.h>
 
 struct cmd_results *output_cmd_dpms(int argc, char **argv) {
```
bl4ckb0ne, line 3: Nit: should print an error message to the user saying it's a deprecated function.

sway/sway-output.5.scd
```diff
@@ -119,12 +119,20 @@ must be separated by one space. For example:
 	Enables or disables the specified output (all outputs are enabled by
 	default).
 
+	As opposed to the _power_ command, the output will loose its current
```
1ace, line 122: `s/loose/lose/` :wink:

emersion: Good catch!

https://github.com/swaywm/sway/pull/7342

PR #7084, opened 2022-06-23
Remove all remaining references to "DPMS"

emersion:
Follow-up for https://github.com/swaywm/sway/pull/7083

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -246,12 +246,6 @@ struct seat_config {
 	} xcursor_theme;
 };
 
-enum config_dpms {
-	DPMS_IGNORE,
-	DPMS_ON,
-	DPMS_OFF,
-};
-
 enum scale_filter_mode {
 	SCALE_FILTER_DEFAULT, // the default is currently smart
 	SCALE_FILTER_LINEAR,
@@ -273,6 +267,7 @@ enum render_bit_depth {
 struct output_config {
 	char *name;
 	int enabled;
+	int power;
 	int width, height;
 	float refresh_rate;
 	int custom_mode;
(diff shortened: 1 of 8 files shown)


include/sway/config.h
```diff
@@ -273,6 +267,7 @@ enum render_bit_depth {
 struct output_config {
 	char *name;
 	int enabled;
+	int power;
```
bl4ckb0ne, line 270: Why not make `power` a `enum zwlr_output_power_v1_mode`?

sway/desktop/output.c
```diff
@@ -1028,10 +1028,10 @@ void handle_output_power_manager_set_mode(struct wl_listener *listener,
 	struct output_config *oc = new_output_config(output->wlr_output->name);
 	switch (event->mode) {
```
bl4ckb0ne, line 1029: Could be simplified in `oc->power = event->mode` but it'll lose some readability.

tiosgz:
`grep` (after filtering out obvious false positives) still reports these, have you left them there intentionally?

```
config.in:40:#          timeout 600 'swaymsg "output * dpms off"' resume 'swaymsg "output * dpms on"' \
include/sway/output.h:35:       // last applied mode when the output is DPMS'ed
sway/desktop/output.c:766:              // We mark the output enabled even if it is switched off by DPMS
```

include/sway/config.h
emersion: 1. I don't want to tie this to a particular Wayland protocol
2. Consistency with the rest of the fields
3. Doesn't have a "default"/"unset" enum entry

sway/desktop/output.c
emersion: I'd rather not rely on the values meaning the same in the protocol and in the Sway codebase.

emersion:
Thanks, fixed!

vyivel (approved):

vyivel:
Thanks!

PR #7088, opened 2022-06-26
minor changes to sway.5.scd

quinnyo:
A couple of minor changes to the hide_edge_borders description.

- now refers to the correct smart_borders option ('smart' -> 'on')

- more consistent use of emphasis

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -744,8 +744,8 @@ The default colors are:
 	Hides window borders adjacent to the screen edges. Default is _none_. The
 	_--i3_ option enables i3-compatible behavior to hide the title bar on
 	tabbed and stacked containers with one child. The _smart_|_smart_no_gaps_
-	options are equivalent to setting _smart_borders_ smart|no_gaps and
-	_hide_edge_borders_ none.
+	options are equivalent to setting smart_borders _on_|_no_gaps_ and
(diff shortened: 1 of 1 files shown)


kennylevinsen:
Does this still render correctly through scdoc? The underscore sequence does look like something that could break it (`_on_|_no_gaps_`)...

quinnyo:
It seems to be fine, I borrowed the syntax from the line above (and elsewhere in the file) as I was unsure myself. Diffing the output from scdoc gives:

```
1086,1087c1086,1087
< options are equivalent to setting \fIsmart_borders\fR smart|no_gaps and
< \fIhide_edge_borders\fR none.\&
---
> options are equivalent to setting smart_borders \fIon\fR|\fIno_gaps\fR and
> hide_edge_borders \fInone\fR.\&
```

quinnyo:
I don't know. I just looked around the file to see how it was done.
Almost every underlined bit of text is a parameter.
There are a lot of mentions of commands without any markers.
I think there are more commands in bold than underlined, not including the ones used as headings.

emersion:
Bold is used for command definitions. Underline is used for references.

quinnyo:
Hi, hopefully that's OK now! Apologies for generating so much back and forth over this :blush:

Issue #6805, opened 2022-01-31
Text in title bars is not rendered when custom font used


Glaeqen:
- **Sway Version:**
  - `sway version 1.7`

- **Debug Log:**
  - [sway.log](https://gist.github.com/Glaeqen/820bba2d08408f1160f85babc803c6df)

- **Configuration File:**
  - Does not reproduce with default configuration/font
  - Used `JetBrains Mono` font (first clause in a configuration)
    - Arch Linux Package of a font: `ttf-jetbrains-mono 2.242-1`
  - [sway.config](https://gist.github.com/Glaeqen/ed955f68bb71f0549252da99e7a0fd1c)

- **Description:**
  - Open few windows in either stacked or tabbed mode
  - Expected result (screenshots from sway 1.6.1)
    - ![Screenshot-2022-01-30T22:58:40,169066684+01:00](https://user-images.githubusercontent.com/20447270/151800648-5a691613-ae1f-4249-88e6-76839f6e7198.png)
    - ![Screenshot-2022-01-30T22:59:04,528837887+01:00](https://user-images.githubusercontent.com/20447270/151800654-a137504e-1da2-49d7-9e8c-3de5d4c0cc5e.png)
  - Actual result (screenshots from sway 1.7)
    - ![Screenshot-2022-01-30T23:02:37,361355455+01:00](https://user-images.githubusercontent.com/20447270/151800843-baf3c94b-ba40-4d1f-a6c9-28a0412fe58b.png)
    - ![Screenshot-2022-01-30T23:02:41,826154779+01:00](https://user-images.githubusercontent.com/20447270/151800850-38d43339-b154-4e98-a98a-2c1080ef5a7f.png)
    - ![Screenshot-2022-01-30T23:03:10,797469390+01:00](https://user-images.githubusercontent.com/20447270/151800852-4b62d94f-e72c-4827-9789-c4601209ddb7.png)
    - ![Screenshot-2022-01-30T23:03:14,908904075+01:00](https://user-images.githubusercontent.com/20447270/151800854-f7026536-d273-4c98-84e3-46e3d061fc3b.png)

soltanmm:
`git bisect` for this winds up on commit 62d90a8e959c6edcc752e124a9928cfa2399fbd1.

Glaeqen:
Yes, explicit font size fixed it. Probably restoring the fallback mechanism or prohibiting the syntax in config file lacking a size is a way to go. The former option has an advantage of being backward compatible though.

[...]

PR #7092, opened 2022-06-30
Reject font values that are invalid for pango

WhyNotHugo:
Use pango to parse font configuration early, and reject the command as
invalid if the value is invalid for pango. Since we're already parsing
the font into a `PangoFontDescription`, keep that instance around and
avoid re-parsing the font each time we render text.

Fixes: https://github.com/swaywm/sway/issues/6805

### DIFF
--- a/sway/commands/font.c
+++ b/sway/commands/font.c
@@ -4,6 +4,7 @@
 #include "sway/config.h"
 #include "log.h"
 #include "stringop.h"
+#include <pango/pangocairo.h>
 
 struct cmd_results *cmd_font(int argc, char **argv) {
 	struct cmd_results *error = NULL;
@@ -22,6 +23,28 @@ struct cmd_results *cmd_font(int argc, char **argv) {
 		config->font = font;
 	}
 
+	// Parse the font early so we can reject it if it's not valid for pango.
+	// Also avoids re-parsing each time we render text.
+	PangoFontDescription *font_description = pango_font_description_from_string(config->font);
+
+	const char *family = pango_font_description_get_family(font_description);
+	if (family == NULL) {
+		pango_font_description_free(font_description);
+		return cmd_results_new(CMD_FAILURE, "Invalid font family.");
+	}
+
+	const gint size = pango_font_description_get_size(font_description);
+	if (size == 0) {
+		pango_font_description_free(font_description);
+		return cmd_results_new(CMD_FAILURE, "Invalid font size.");
+	}
+
+	if (config->font_description != NULL) {
+		pango_font_description_free(config->font_description);
+	}
+
+	config->font_description = font_description;
 	config_update_font_height();
+
 	return cmd_results_new(CMD_SUCCESS, NULL);
 }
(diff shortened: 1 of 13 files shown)


sway/commands/font.c
```diff
@@ -16,12 +17,34 @@ struct cmd_results *cmd_font(int argc, char **argv) {
 	if (strncmp(font, "pango:", 6) == 0) {
 		config->pango_markup = true;
 		config->font = strdup(font + 6);
+		free(font);
```
WhyNotHugo, line 20: In this branch, I think it makes sense to do:

```c
config->font = font + 6;
free(font[0..5]); // Pseudocode, not sure how to free just the first 5 bytes.
```

This avoids one unecessary `strdup`. Does this make sense? If so, how would you
do this?

sway/commands/font.c
```diff
@@ -16,12 +17,34 @@ struct cmd_results *cmd_font(int argc, char **argv) {
...
+	if (size == 0) {
+		pango_font_description_free(font_description);
+		return cmd_results_new(CMD_FAILURE, "Invalid font size.");
+	}
+
+	if (config->font_description != NULL) {
+		pango_font_description_free(config->font_description);
+	}
+
+	config->font_description = font_description;
```
WhyNotHugo, line 46: This has a risk that the value will (for only an instant) be NULL, so if
another thread tries to read it, it'll fail. Could this be an issue? Should I
reassign _and then_ `free`?

sway/config.c
```diff
@@ -243,6 +243,7 @@ static void config_defaults(struct sway_config *config) {
 	config->default_layout = L_NONE;
 	config->default_orientation = L_NONE;
 	if (!(config->font = strdup("monospace 10"))) goto cleanup;
+	config->font_description = pango_font_description_from_string(config->font);
```
WhyNotHugo, line 246: Can we call the `font` command here with arg `"monospace 10"`?

We'd be able to de-duplicate some font-configuration bits, and also drop the
extra call to `config_update_font_height` below (which is kind of a hack). In
fact, we can move the logic from `config_update_font_height` into the font
command itself, since the font height only changes when the font changes, and
drop the separate `config_update_font_height` entirely.

Does this seem reasonable?

sway/commands/font.c
emersion: That's not possible.

But it doesn't matter too much anyways, we shouldn't try micro-optimizing too hard.

sway/config.c
WhyNotHugo: I'll re-think how to optimise that further and address it separately then. It's a non-issue in this scope.

emersion:
This LGTM!

Follow-up issue: also pass a parsed `PangoFontDescription` to other text-related functions like `render_text`, but that's a more involved change.

WhyNotHugo:
I'd missed that one! That was quicker than expected (took less than my lunch in the oven :joy: ).

I wonder if `swaynag` and `swaybar` should fall back to `"monospace 10"` if the given input is invalid too. That's out of scope tho; the current versions will also misrender if the input font is invalid.

emersion (approved):
Nice, the new commit LGTM as well, thanks!

PR #7099, opened 2022-07-13
Fix bindswitch not executing all binds

varqox:
If there was bindswitch on lid:toggle and either lid:on or lid:off (or both).
Only one bind was executed instead of two. E.g. if the lid is closed sway should
execute lid:on and lid:toggle binds. But it executed only one of them (the one
that appears later in the config).

This is a regression introduced here: 6afb392823d27ec69bedc8fd74263c3d072cca29.

Additionally, made toggle event always happen before on or off event. This is
more convenient instead of an arbitrary order. It is easier to implement and
more intuitive because conceptually, toggling of the state happens before
the target state is reached.

### DIFF
--- a/sway/input/switch.c
+++ b/sway/input/switch.c
@@ -20,7 +20,12 @@ struct sway_switch *sway_switch_create(struct sway_seat *seat,
 	return switch_device;
 }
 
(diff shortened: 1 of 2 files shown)


emersion:
Hm it sounds like this was by design?

PR #7104, opened 2022-07-17
Set opacity of scratchpad windows on exit in inactive-windows-transparency script

anpandey:
`workspaces()` doesn't include windows in the scratchpad which means they're left transparent when the script exits. This just iterates over the scratchpad containers as well and resets their opacity.


emersion:

PR #7108, opened 2022-07-21
Add 16f option to render_bit_depth command

mstoeckl:
This change adds an option to let sway render onto XBGR16161616F buffers, assuming hardware and drivers support this. 

To test that this works, run `swaymsg output 'output-name' render_bit_depth 16f`. If hardware and drivers support this, then one can check that it has worked with 'drm_info' and search the output for 'framebuffer'; with Intel drivers, the framebuffer info for the primary plane being used should say `Format: XBGR16161616F `; with AMD, if I recall correctly, it is slightly different; a 'Bit depth" field should have value 64.

If Sway hangs or has corrupt output when `...  render_bit_depth 16f` is run, running Sway with the environment variable `WLR_DRM_NO_MODIFIERS=1` may resolve the issue.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -258,6 +258,7 @@ enum render_bit_depth {
 	RENDER_BIT_DEPTH_DEFAULT, // the default is currently 8
 	RENDER_BIT_DEPTH_8,
 	RENDER_BIT_DEPTH_10,
+	RENDER_BIT_DEPTH_16F,
 };
 
 /**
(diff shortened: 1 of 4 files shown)


emersion:
I am not a fan of using the "bit depth" command to set the full format.

mstoeckl:
> I am not a fan of using the "bit depth" command to set the full format.

This patch attempts only the XBGR16F format before falling back to XRGB8, because in any scenario where another format in the list (ABGR16F, ARGB16F, XRGB16F) is supported, then XBGR16F will also be supported. At the moment, Mesa only supports XBGR16F and ABGR16F formats, so the XRGB16F format will not work. Also, looking at https://drmdb.emersion.fr/formats, it seems that all driver/hardware combinations that allow any of (ABGR16F, ARGB16F, XRGB16F) as a plane format also allow XBGR16F.

The documentation changes in this patch do not refer to any specific buffer format, so future changes to the list of formats for 16F depth in the `bit_depth_preferences` table are possible. (For example, if hardware appears that only allows XRGB16F, or a more efficient 48-bit BGR161616F format is introduced.)

emersion:
Right, but "bit depth" is a misnomer, whether a format is floating-point or not has nothing to do with its bit depth.

PR #7109, opened 2022-07-23
ipc: fix segfault when reconnecting keyboard

Akaricchi:
This fix may not be correct, as I'm not familiar enough with the sway codebase to know if keymap is allowed to be NULL here or not. But it works for me, seemingly with no adverse effects.

FWIW, I'm using a custom `xkb_file`.


emersion:
Duplicate of https://github.com/swaywm/sway/pull/6484, see also https://github.com/swaywm/sway/pull/6873

PR #7113, opened 2022-07-25
ipc: make get_deco_rect check config->hide_lone_tab

baltitenger:
Without this, the `IPC_GET_TREE` ipc call would return false information about the container's `deco_rect` and `rect` properties if `hide_edge_borders --i3` was in effect.


emersion (approved):
LGTM, thanks!

PR #7127, opened 2022-08-03
allow no titlebar in tabbed mode

bgkillas:
i just want the option to have no titlebar and just use keybinds to switch between them and cant see reason to not give the option

### DIFF
--- a/sway/commands/titlebar_padding.c
+++ b/sway/commands/titlebar_padding.c
@@ -7,24 +7,13 @@
 
 struct cmd_results *cmd_titlebar_padding(int argc, char **argv) {
 	struct cmd_results *error = NULL;
-	if ((error = checkarg(argc, "titlebar_padding", EXPECTED_AT_LEAST, 1))) {
-		return error;
-	}
-
 	char *inv;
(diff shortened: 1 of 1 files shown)


WillPower3309:
I don't think this is the way to do this. Check out the tabbed container rendering function:
https://github.com/swaywm/sway/blob/master/sway/desktop/render.c#L776

If you want no titlebars in tabbed mode, I think the solution is to add a config option for ignoring them and then add a check for that option in the function I linked above.

markstos:
This is a niche feature not in i3, and features not in i3 are generally not being accepted.

sweenu:
See: https://github.com/swaywm/sway/issues/7350

markstos:
@sweenu #7350 is not about tabbed mode.

leinfink:
> @sweenu #7350 is not about tabbed mode.

It doesn't seem to be, at least it doesn't say so directly, but AFAIK the workaround they mention (setting font-size to 0) was only needed to remove titlebars in tabbed mode, not the "normal" ones. So I think that issue _is_ actually about tabbed mode.

PR #7132, opened 2022-08-08
sway-output.5: improve display of parameter

tbm:
Since "width" and "height" are separate parameters, show them as such.

### DIFF
--- a/sway/sway-output.5.scd
+++ b/sway/sway-output.5.scd
@@ -24,7 +24,7 @@ must be separated by one space. For example:
 
 # COMMANDS
 
-*output* <name> mode|resolution|res [--custom] <WIDTHxHEIGHT>[@<RATE>Hz]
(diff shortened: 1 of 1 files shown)


tbm:
Someone mentioned this formatting issue on Reddit: https://www.reddit.com/r/swaywm/comments/wguayr/i_cant_seem_to_change_my_monitors_refreshrate_to/ij45lsz/

I'm also not sure why the parameters are uppercase but I left that.

emersion:
Ah, I think they were upper-case to make them more readable with a lower-case "x" in the middle. I think I'd make sense to switch them to lower-case with this improved syntax.

tbm:
> I think I'd make sense to switch them to lower-case with this improved syntax.

Updated the PR accordingly.

emersion (approved):
LGTM, thanks!

PR #7143, opened 2022-08-22
sway/desktop/output.c: func handle_destroy add free

yuanye100:


emersion:
Output destroy is delayed in Sway. `output_destroy` takes care of free'ing memory already.

emersion:
Feel free to re-open if you think there is something which needs to be addressed.

PR #7178, opened 2022-09-19
allow pointer_constraints on layer_shell surfaces

### LINKED ISSUE #5634: layer-shell surfaces aren't capable of constraining the pointer.
Hello there, let me start by saying this is a very exciting time to be getting aware of sway! I went through the code and is very well written! nice!

I am an experienced Linux user and I am finally trying to move away from X11 on all my devices after realizing how better things are with wayland/sway.

While trying to mush wlroots examples together, I realized the layer-shell example doesn't go well with the pointer-constraint protocol. What I would like to do is to be able to lock the cursor if it goes beyond the viewport, and subsequently release it.

I tried various approaches on how to circumvent the problem but I was finally able to make the both nonstandard protocols work together when I removed this `if` statement on `sway/input/cursor.c` : https://github.com/swaywm/sway/blob/master/sway/input/cursor.c#L1294

I looks like overlays never have the "proper focus"... Is this a bug, or intentionally designed? Would I be able to constrain the cursor into an overlay in a different way?

ferdinandschober:
This could resolve #5634

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1462,6 +1462,18 @@ struct sway_node *seat_get_focus(struct sway_seat *seat) {
 	return current->node;
 }
 
+struct wlr_surface *seat_get_focused_surface(struct sway_seat *seat) {
+	struct sway_node *focus = seat_get_focus(seat);
+	if (focus && node_is_view(focus)) {
+		return focus->sway_container->view->surface;
+	}
+	struct sway_seat_node *current = wl_container_of(seat->focus_stack.next, current, link);
+	struct wlr_layer_surface_v1 *layer = current->seat->focused_layer;
+	if (layer && layer->surface)
+		return layer->surface;
+	return NULL;
+}
+
 struct sway_workspace *seat_get_focused_workspace(struct sway_seat *seat) {
 	struct sway_node *focus = seat_get_focus_inactive(seat, &root->node);
 	if (!focus) {
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1462,17 +1462,6 @@ struct sway_node *seat_get_focus(struct sway_seat *seat) {
 	return current->node;
 }
 
-struct wlr_surface *seat_get_focused_surface(struct sway_seat *seat) {
-	struct sway_node *focus = seat_get_focus(seat);
-	if (focus && node_is_view(focus)) {
-		return focus->sway_container->view->surface;
-	}
-	struct wlr_layer_surface_v1 *layer = seat->focused_layer;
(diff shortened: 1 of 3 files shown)


ferdinandschober:
Tested using [LAN Mouse](https://github.com/ferdinandschober/lan-mouse). It's working for me.

ferdinandschober:
This behaviour is the same on KDE.

ferdinandschober:
@emersion could you take a look?

sway/input/seat.c
```diff
@@ -1462,6 +1462,18 @@ struct sway_node *seat_get_focus(struct sway_seat *seat) {
 	return current->node;
 }
 
+struct wlr_surface *seat_get_focused_surface(struct sway_seat *seat) {
+	struct sway_node *focus = seat_get_focus(seat);
+	if (focus && node_is_view(focus)) {
+		return focus->sway_container->view->surface;
+	}
+	struct sway_seat_node *current = wl_container_of(seat->focus_stack.next, current, link);
+	struct wlr_layer_surface_v1 *layer = current->seat->focused_layer;
```
emersion, line 1471: Instead of going through `focus_stack`, can't we just read `seat->focused_layer` directly without the intermediate `current` variable?

sway/input/seat.c
```diff
@@ -1462,6 +1462,18 @@ struct sway_node *seat_get_focus(struct sway_seat *seat) {
 	return current->node;
 }
 
+struct wlr_surface *seat_get_focused_surface(struct sway_seat *seat) {
```
emersion, line 1465: Ideally this function would also handle the case where a lockscreen has focus.

sway/input/seat.c
ferdinandschober: yes that does indeed work, I copied the code from seat_get_focus without realizing I don't need the node

sway/input/seat.c
ferdinandschober: Is a lockscreen not covered by layershell? What do I need to do here?

emersion:
It occurs to me that we shouldn't even need this new `seat_get_focused_surface` function, we can instead look at `seat->wlr_seat->keyboard_state.focused_surface`. This would handle all cases.

ferdinandschober:
> It occurs to me that we shouldn't even need this new `seat_get_focused_surface` function, we can instead look at `seat->wlr_seat->keyboard_state.focused_surface`. This would handle all cases.

I implemented it this way and it's working for me

emersion (approved):
LGTM, thanks!

ferdinandschober:
Thank you, too!

PR #7183, opened 2022-09-23
Improve NULL title handling

puckipedia:
Any windows that have never had a title set visually behave closer to that of an empty title, but are unformattable, as the code bails out early on a NULL title.

(as an example, `hello-wayland` does this, as well as xwayland windows with an empty title.)

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -189,7 +189,7 @@ static bool criteria_matches_view(struct criteria *criteria,
 	if (criteria->title) {
 		const char *title = view_get_title(view);
 		if (!title) {
-			return false;
+			title = "";
 		}
 
 		switch (criteria->title->match_type) {
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -1282,20 +1282,22 @@ void view_update_title(struct sway_view *view, bool force) {
 
 	free(view->container->title);
 	free(view->container->formatted_title);
-	if (title) {
-		size_t len = parse_title_format(view, NULL);
+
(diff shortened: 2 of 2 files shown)


emersion:
What is the i3 behavior here?

puckipedia:
i3 seems to properly handle titles even if set to NULL, looking at the code.

https://github.com/i3/i3/blob/dba30fc9879b42e6b89773c81e1067daa2bb6e23/src/con.c#L2347
https://github.com/i3/i3/blob/dba30fc9879b42e6b89773c81e1067daa2bb6e23/src/match.c#L94-L110

sway/criteria.c
```diff
@@ -189,7 +189,7 @@ static bool criteria_matches_view(struct criteria *criteria,
 	if (criteria->title) {
 		const char *title = view_get_title(view);
 		if (!title) {
-			return false;
+			title = "";
```
WhyNotHugo, line 192: I don't think this is correct, this changes the title of windows when trying to match it to a criteria.

This is not the right place to do this.

WhyNotHugo:
Is the title actually a NULL pointer, or an empty string `""`?

sway/criteria.c
emersion: This doesn't change the title of windows. It just falls back to a value which can be matched.

emersion (approved):
LGTM!

I think we should do the same for the other criteria -- do you think you'd have time to submit a patch for this?

emersion:
Opened https://github.com/swaywm/sway/issues/7235

PR #7184, opened 2022-09-24
Dynamic max render time

any1:
This adds automatic detection of output `max_render_time`.

Requirement: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3737

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -21,6 +21,9 @@
 
 // TODO: Refactor this shit
 
+#define MAX_RENDER_TIME_OFF 0
+#define MAX_RENDER_TIME_AUTO -2
+
 /**
  * Describes a variable created via the `set` command.
  */
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -55,6 +55,14 @@ struct sway_output {
 	uint32_t refresh_nsec;
 	int max_render_time; // In milliseconds
 	struct wl_event_source *repaint_timer;
+
+	struct timespec render_begin_time;
+	struct wlr_render_timestamp *render_end_ts;
+
+	int auto_max_render_time; // in milliseconds
+	// TODO: Make this adjustable:
+	uint8_t render_time_window[60];
+	int render_time_window_index;
 };
 
 struct sway_output_non_desktop {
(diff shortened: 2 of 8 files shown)


emersion:
cc @mcoffin

any1:
## Thoughts on estimating `max_render_time`

Things that are known:
 * We can measure how long it takes to render individual frames with fairly
   decent accuracy.
 * We also know the vblank time for every previous frame.
 * We can approximate the next deadline for frame submission.
 * When a frame submission misses its deadline, the next frame is dropped.

Things that are unknown:
 * We do not know the **exact** deadline for the next frame submission.

Because we do not know the exact deadline, we need to give ourselves some
margins for error.

### How do you calculate the error margins?

One way to arrive at a good error margin is to increase the error margins every
time a frame is dropped. However, frames can also be dropped due to
fluctuations in rendering time. Thus, error margins should only be increased
while rendering time is stable. A good starting point for the increment might
be 100 µs.

### Method for estimating rendering time average, maximum deviation and stability:

 * A sliding window of previous rendering times can be maintained.
 * The running average, is simply the average over the window.
 * The maximum deviation is the difference between the lowest and highest value.
 * Rendering time can be considered stable if the maximum deviation is close
   enough to the standard deviation.
 * A suitable scaling factor for comparing standard and maximum devation should
   be found with experimentation.
 * I have used this method for estimating the speed stability of rotary encoders
   in the past and it worked well enough for that purpose.
 * Suggestions for alternate methods are encouraged.

Within this framework, maximum rendering time can be computed as the sum of
average rendering time, maximum deviation of rendering time and the error
margin.

any1:
I should mention that (if I remember correctly) I already tried the method described previously for estimating stability and it worked rather poorly.

The following are some of my thoughts after the meeting today:

Someone actually suggested on IRC a while back using a TCP bandwidth estimation algorithm, which might be an interesting approach. It is worth mentioning, though, that naive ones tend to be rather "saw-toothy".

The rendering deadline is basically a random variable and there are going to be outliers, so we are always going to drop some frames. It would be good to develop a statistical model for the distribution of this random variable. Then, the question becomes how many frame drops over a given period are we willing to accept. Then we can adjust the delay based on the pdf and the "acceptable average frame drop" variable.

It might be possible to identify a few different modes of operation and adjust the parameters for the pdf. E.g. when you have a full-screen application running, and also as someone mentioned in the meeting, things such as damage could be taken into account.

While you're operating in a particular mode, such as watching a video or playing a game (and you don't do anything else), the deadline should be fairly stable. It might be a good idea to try to detect when the deadline becomes unstable and add a larger delay while things are settling down.

PR #7185, opened 2022-09-25
man: Add XWayland information

mrusme:
I just tripped over this detail myself, not knowing that these parameters won't work anymore with X/XWayland disabled.


emersion:
Thanks!

PR #7187, opened 2022-09-26
Security context based criteria

puckipedia:
(builds on #7041. this requires a slightly updated version of the wlroots patches, which are currently only [on my fork](https://gitlab.freedesktop.org/puckipedia/wlroots/-/tree/security-context-fixes))

This adds `sandbox_engine` and `sandbox_app_id` to the supported window criteria, as well as `format_title` tokens and the JSON IPC representation for a view.


FelixPehla:
Since `security-context-v1` has been merged into wayland protocols and sway supports it, are there any plans to rebase these commits on the current implementation?

Issue #5522, opened 2020-07-06
Provide a way for headless outputs to be destroyed


RedSoxFan:
This is a follow up to #5483, where there was an attempt to remove a headless output using `output HEADLESS-1 disable`. Since disabling a headless output does not make sense and is not supported by wlroots, there should be a way to destroy these outputs.

We'll likely want something like `output HEADLESS-1 destroy`. The output command does not currently have any action-only subcommands so we'll likely want to add support for those. Unfortunately, the output command also has the unique property of allowing subcommands to be chained (for backwards compatibility) which may complicates things. We'll likely want to prevent this chaining for action-only subcommands and make them two different branches.

emersion:
Potential alternative solutions:

- Make disabling a no-op in the headless backend?
- Support disabling a Sway output even if wlroots can't?

Hubro:
Any workarounds available for this? What do I do when I'm no longer using the virtual output? Do I just have to reboot to avoid having a workspace permanently bound to an invisible display?

emersion:
Send a patch?

ldelossa:
> Any workarounds available for this? What do I do when I'm no longer using the virtual output? Do I just have to reboot to avoid having a workspace permanently bound to an invisible display?

Just spitballing, but you could assign some impossible dimension like 9999x0 to the headless and reconfigure sway. Then just move any workspaces that were there to your normal screens?

Hubro:
> > Any workarounds available for this? What do I do when I'm no longer using the virtual output? Do I just have to reboot to avoid having a workspace permanently bound to an invisible display?
> 
> Just spitballing, but you could assign some impossible position like 9999x0 to the headless and reconfigure sway. Then just move any workspaces that were there to your normal screens?  

I tried something similar a while back, I think I set it to 1x1 and placed it top right of my main monitor. Unfortunately it caused a slew of bugs, I think some applications would constantly crash, even though I didn't place them on the 1x1 monitor. It's also too easy to accidentally move a window to the invisible monitor, in which case they would usually crash/freeze indefinitely.

I've just stopped using virtual monitors until this feature is implemented, it's more trouble than it's worth.

PR #7192, opened 2022-09-29
sway/commands/output: Add command for unplugging non-physical outputs

any1:
Closes: https://github.com/swaywm/sway/issues/5522

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -294,6 +294,7 @@ sway_cmd output_cmd_scale_filter;
 sway_cmd output_cmd_subpixel;
 sway_cmd output_cmd_toggle;
 sway_cmd output_cmd_transform;
+sway_cmd output_cmd_unplug;
 
 sway_cmd seat_cmd_attach;
 sway_cmd seat_cmd_cursor;
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -27,6 +27,7 @@ static const struct cmd_handler output_handlers[] = {
 	{ "subpixel", output_cmd_subpixel },
 	{ "toggle", output_cmd_toggle },
 	{ "transform", output_cmd_transform },
+	{ "unplug", output_cmd_unplug },
 };
 
 struct cmd_results *cmd_output(int argc, char **argv) {
--- a/sway/commands/output/unplug.c
+++ b/sway/commands/output/unplug.c
@@ -1,0 +1,54 @@
+#include <strings.h>
+#include <wlr/config.h>
+#include <wlr/backend/headless.h>
+#include <wlr/backend/wayland.h>
+#if WLR_HAS_X11_BACKEND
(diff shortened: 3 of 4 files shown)


emersion (approved):
LGTM!

mauro3:
In case someone also wonders how to actually use this feature:
```
swaymsg output HEADLESS-1 unplug
```

fritzrehde:
I'm not sure if this is a bug or a feature, but after unplugging using `swaymsg output HEADLESS-1 unplug`, creating a new output again using `swaymsg create_output` will increment the number appended to `HEADLESS-<num>` everytime. So after unplugging `HEADLESS-1`, `HEADLESS-2` will be created next. This seems unideal for scripting purposes. Is there a way to properly remove `HEADLESS-1`, so I can continue using it once I use `swaymsg create_output` again? Or is there something more like `swaymsg output HEADLESS-1 plug`?

Riolku:
@fritzrehde Unfortunately it seems [part of wlroots](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/headless/output.c#L131). I think ill just use a small bash script with `swaymsg` and `jq`.

fritzrehde:
@Riolku Can you share the script you eventually come up with here then? Would be curious to see it

Riolku:
Haven't gotten a chance to iron it out, but you can get the active HEADLESS output with `swaymsg -t get_outputs | jq .[].name | grep HEADLESS | tr -d '"'`

"Get outputs and get the names of the outputs and look for the headless one and trim the double quotes"

fritzrehde:
I don't quite understand the reasoning behind incrementing though. Could someone explain that?

Riolku:
No chance of collision.

fritzrehde:
> Haven't gotten a chance to iron it out, but you can get the active HEADLESS output with `swaymsg -t get_outputs | jq .[].name | grep HEADLESS | tr -d '"'`
> 
> "Get outputs and get the names of the outputs and look for the headless one and trim the double quotes"

Ok, but the issue is more: How do I write e.g. a kanshi profile for the headless output? Or even just an output configuration in the sway config file?

Riolku:
I don't know what your use case is, but maybe this extended discussion is better in IRC. Feel free to ping me there.

rodhfr:
here's my script for using with wayvnc
```bash
#!bin/bash
pkill wayvnc
swaymsg create_output

# Get the list of output names containing "HEADLESS", ignoring the first one
headless_outputs=$(swaymsg -t get_outputs | jq -r '.[].name' | grep HEADLESS | tail -n +2)

# Loop through the outputs (starting from the second one)
for i in $headless_outputs; do
    swaymsg output "$i" unplug
done

swaymsg -t get_outputs
swaymsg output "HEADLESS-1" resolution 1920x1080
nohup wayvnc -o HEADLESS-1 &
```

PR #7206, opened 2022-10-06
commands: introduce non_desktop command for output

### LINKED ISSUE #7021: Create some way of dynamically setting displays as desktop or non-desktop
Two use-cases:
* I have a display that says it's non-desktop (like say an Index) and for some reason I want to extend the desktop to it anyway
* I have a display that looks to Sway like a normal monitor (like say a Project North Star AR headset) and I need to tell Sway to treat it as non-desktop so that Monado can do direct mode with it

bl4ckb0ne:
This command allows users to set a given output as non-desktop. The output is offered to the drm-lease-manager interface if available, for clients to lease.

Closes #7021

### DIFF
--- a/sway/commands/output/non_desktop.c
+++ b/sway/commands/output/non_desktop.c
@@ -1,0 +1,106 @@
+#include <strings.h>
+#include <wlr/types/wlr_output_damage.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "sway/output.h"
+#include "sway/server.h"
+#include "log.h"
+#include "util.h"
+
+static int find_non_desktop_output(const void *item, const void *data)
+{
+	const struct sway_output_non_desktop *output = item;
+	const char *name = data;
+	return strcmp(output->wlr_output->name, name);
+}
+
+static int find_output(const void *item, const void *data)
+{
+	const struct sway_output *output = item;
+	const char *name = data;
+	return strcmp(output->wlr_output->name, name);
+}
+
+struct cmd_results *output_cmd_non_desktop(int argc, char **argv) {
+	if (argc == 0) {
+		return cmd_results_new(CMD_INVALID, "Missing non_desktop argument");
+	}
+
+	struct output_config *oc = config->handler_context.output_config;
+
+	if (strcmp(oc->name, "*") == 0) {
+		return cmd_results_new(CMD_INVALID,
+			"Cannot apply non_desktop to all outputs");
+	}
+
+	if (parse_boolean(argv[0], true)) {
+		int index = list_seq_find(root->outputs, find_output, oc->name);
+		if (index == -1) {
+			return cmd_results_new(CMD_FAILURE, "unknown output %s ", oc->name);
+		}
+
+		struct sway_output *output = root->outputs->items[index];
+		struct wlr_output *wlr_output = output->wlr_output;
+		wlr_output->non_desktop = true;
+
+		sway_log(SWAY_DEBUG, "output %s (%p) non-desktop on", wlr_output->name,
+			wlr_output);
+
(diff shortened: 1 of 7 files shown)


sway/commands/output/non_desktop.c
```diff
@@ -0,0 +1,106 @@
...
+
+	struct output_config *oc = config->handler_context.output_config;
+
+	if (strcmp(oc->name, "*") == 0) {
+		return cmd_results_new(CMD_INVALID,
+			"Cannot apply non_desktop to all outputs");
+	}
+
+	if (parse_boolean(argv[0], true)) {
+		int index = list_seq_find(root->outputs, find_output, oc->name);
```
emersion, line 37: We already have a function to do this:

```
struct sway_output *sway_output = all_output_by_name_or_id(oc->name);
```

It also handles output identifiers.

sway/commands/output/non_desktop.c
```diff
@@ -0,0 +1,106 @@
...
+		if (index == -1) {
+			return cmd_results_new(CMD_FAILURE, "unknown output %s ", oc->name);
+		}
+
+		struct sway_output *output = root->outputs->items[index];
+		struct wlr_output *wlr_output = output->wlr_output;
+		wlr_output->non_desktop = true;
```
emersion, line 44: Hmm. Do we really need to do this? I don't like writing to the `wlr_output` struct.

sway/commands/output/non_desktop.c
```diff
@@ -0,0 +1,106 @@
...
+			wlr_output);
+
+		output_disable(output);
+		wlr_output_layout_remove(root->output_layout, wlr_output);
+
+		struct sway_output_non_desktop *non_desktop =
+			output_non_desktop_create(wlr_output);
+		list_add(root->non_desktop_outputs, non_desktop);
+		
+		if (server.drm_lease_manager) {
```
emersion, line 56: Maybe fail early in this case?

emersion:
I'd prefer to name this something else. "non desktop" is just a hint coming from the kernel. This command allows one to lease "desktop" outputs.

tl;dr something like "drm_lease" would be a better name IMHO.

sway/sway-output.5.scd
```diff
@@ -180,6 +180,11 @@ must be separated by one space. For example:
 	updated to work with different bit depths. This command is experimental,
 	and may be removed or changed in the future.
 
+*output* <name> non_desktop on|off
+	Enables or disables the non-desktop property of a given output. When
+	enabled, the output is disabled and offered to the drm-lease-manager
+	interface if it is available.
```
emersion, line 186: Maybe we should leave this command undocumented for a bit and see how it goes… A bit like `create_output` and `output XXX unplug`. If we do document it, then we're stuck with it ~forever.

sway/commands/output/non_desktop.c
bl4ckb0ne: `all_output_by_name_or_id` searches in `root->all_outputs`,  is it better? I didnt get the difference between the two lists.

sway/commands/output/non_desktop.c
bl4ckb0ne: I don't recally why I did that, maybe a leftover from a previous draft. I'll remove it.

bl4ckb0ne:
Now that you're saying it out loud it does make a lot of sense to have a `lease` command instead of using `non_desktop`. But that changes the scope of the whole patchset. 

Should true non-desktop output still be treated differently? Or should we treat them as regular output and already lease them unless the users requests to withdraw them?

bl4ckb0ne:
Closed in favour of #7207

PR #7207, opened 2022-10-07
commands: introduce output leasing command

### LINKED ISSUE #7021: Create some way of dynamically setting displays as desktop or non-desktop

bl4ckb0ne:
Supersedes https://github.com/swaywm/sway/pull/7206

Requires https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3323 for bugfix

Closes https://github.com/swaywm/sway/issues/7021

### DIFF
--- a/sway/commands/output.c
+++ b/sway/commands/output.c
@@ -8,6 +8,7 @@
 // must be in order for the bsearch
 static const struct cmd_handler output_handlers[] = {
 	{ "adaptive_sync", output_cmd_adaptive_sync },
+	{ "allow_drm_leasing", output_cmd_allow_drm_leasing },
 	{ "allow_tearing", output_cmd_allow_tearing },
 	{ "background", output_cmd_background },
 	{ "bg", output_cmd_background },
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -597,6 +597,16 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 		sway_session_lock_add_output(server->session_lock.lock, output);
 	}
 
+	struct output_config *oc = find_output_config(output);
+	if (oc && oc->allow_drm_leasing > 0) {
+#if WLR_HAS_DRM_BACKEND
+		if (server->drm_lease_manager) {
+			wlr_drm_lease_v1_manager_offer_output(server->drm_lease_manager,
+					wlr_output);
+		}
+#endif
+	}
+
 	request_modeset();
 }
 
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -284,6 +284,7 @@ sway_cmd input_cmd_xkb_switch_layout;
 sway_cmd input_cmd_xkb_variant;
 
 sway_cmd output_cmd_adaptive_sync;
+sway_cmd output_cmd_allow_drm_leasing;
 sway_cmd output_cmd_allow_tearing;
 sway_cmd output_cmd_background;
 sway_cmd output_cmd_color_profile;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -291,6 +291,7 @@ struct output_config {
 	bool set_color_transform;
 	struct wlr_color_transform *color_transform;
 	int allow_tearing;
+	int allow_drm_leasing;
 	int hdr;
 
 	char *background;
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -70,6 +70,7 @@ struct sway_output {
 	struct wl_event_source *repaint_timer;
 
 	bool allow_tearing;
+	bool allow_drm_leasing;
 	bool hdr;
 };
 
--- a/sway/commands/output/allow_drm_leasing.c
+++ b/sway/commands/output/allow_drm_leasing.c
@@ -1,0 +1,60 @@
+#include "sway/commands.h"
+#include "sway/output.h"
+#include "sway/server.h"
+#include "log.h"
+#include "util.h"
+
+#if WLR_HAS_DRM_BACKEND
+#include <wlr/backend/drm.h>
+#include <wlr/types/wlr_drm_lease_v1.h>
+#endif
+
+struct cmd_results *output_cmd_allow_drm_leasing(int argc, char **argv) {
+	if (!server.drm_lease_manager) {
+		return cmd_results_new(CMD_FAILURE,
+			"DRM lease manager interface not available");
+	}
+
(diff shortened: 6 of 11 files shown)


sway/desktop/output.c
```diff
@@ -922,9 +912,22 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
 
 	struct output_config *oc = find_output_config(output);
+	
 	apply_output_config(oc, output);
 	free_output_config(oc);
 
+	if (wlr_output->non_desktop) {
+		sway_log(SWAY_DEBUG, "Disabling non-desktop output");
+		output_disable(output);
+		wlr_output_layout_remove(root->output_layout, wlr_output);
```
bl4ckb0ne, line 922: Should `output_disable` take care of removing the output from the layout?

emersion:
What about the following behavior?

- Outputs available for leasing are working as usual until a client comes and leases them (will be easier with [wlroots!4771](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4771) maybe).
- Non-desktop outputs are disabled by default, but can still be explicitly enabled if the user wants to.

sway/commands/output.c
```diff
@@ -13,6 +13,7 @@ static const struct cmd_handler output_handlers[] = {
 	{ "disable", output_cmd_disable },
 	{ "dpms", output_cmd_dpms },
 	{ "enable", output_cmd_enable },
+	{ "leasing", output_cmd_leasing },
```
emersion, line 16: Obligatory naming bikeshed:

- Should we add "drm" in the name, to make it's clear it's about DRM leasing?
- Should we add an "allow_" prefix, like we have for "allow_tearing"? Might be a bit long and doesn't really improve clarity.

sway/commands/output/leasing.c
```diff
@@ -0,0 +1,66 @@
...
+		return cmd_results_new(CMD_FAILURE,
+			"Cannot apply leasing to unknown output %s", oc_name);
+	}
+
+	if (argc == 0) {
+		return cmd_results_new(CMD_INVALID, "Missing leasing argument");
+	}
+
+	bool status = true;
+	if (parse_boolean(argv[0], status)) {
```
emersion, line 33: We should be passing `sway_output->leasing` here instead of `status` I think (to make `toggle` work).

sway/commands/output/leasing.c
```diff
@@ -0,0 +1,66 @@
...
+		sway_output->leasing = true;
+		output_disable(sway_output);
+		wlr_output_layout_remove(root->output_layout, sway_output->wlr_output);
+		
+		sway_log(SWAY_DEBUG, "Offering output %s for leasing", oc_name);
+		wlr_drm_lease_v1_manager_offer_output(server.drm_lease_manager,
+			sway_output->wlr_output);
+	
+		config->handler_context.output_config->enabled = 0;
+
```
emersion, line 48: Style nit: stray newline

sway/desktop/output.c
```diff
@@ -922,9 +912,26 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
 	apply_output_config(oc, output);
 	free_output_config(oc);
 
+	/* TODO read from config for leasing new outputs */
+	if (wlr_output->non_desktop) {
+		if (output->enabled) {
+			sway_log(SWAY_DEBUG, "Disabling non-desktop output");
+			output_disable(output);
+			wlr_output_layout_remove(root->output_layout, wlr_output);
+		}
```
emersion, line 925: Hm it's a bit unfortunate to do this back-and-forth. I think with my behavior suggestion this will go away anyways.

sway/commands/output.c
bl4ckb0ne: I like `allow_drm_leasing`, seems pretty explicit to me

bl4ckb0ne:
>Outputs available for leasing are working as usual until a client comes and leases them (will be easier with [wlroots!4771](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4771) maybe).

That's what I had in mind yes, the user can continue to use it until it gets leased by a client

 > Non-desktop outputs are disabled by default, but can still be explicitly enabled if the user wants to.

What do you mean? Non-desktop outputs are always offered for leasing and always disabled? If so yes, the `leasing` option on those should be ignored. I can't find a use case for user to explicitly enable those tbh.

emersion:
My suggestion is to allow users to explicitly enable non-desktop outputs, and allow users to explicitly disable DRM leasing for these.

bl4ckb0ne:
What can the user do with an enabled non-desktop output? Use it as a regular output?

emersion:
Here are possible use-cases:

- A few HMDs (e.g. PSVR1) support transforming the HDMI stream to adapt it for the user (duplicating the stream for each eye, applying the lens distortion/correction/etc).
- Users may want to not use DRM leasing, and instead put a xdg-toplevel with the side-by-side rendered images on the HMD's output. This may be desirable e.g. when the VR engine doesn't support DRM leasing, or when the bandwidth is tight (DRM leasing doesn't allow for proper bandwidth-aware buffer allocation).
- The non-desktop flag is used for outputs that are _not_ HMDs as well, for instance Apple's touch bar. I don't think DRM leasing makes a lot of sense for such cases. It would also be possible for the non-desktop flag to be set by mistake by the kernel (because the EDID matching is too wide, for instance).

On top of that, I believe that it would simplify the codebase because it would remove the special cases. All outputs would go through the same codepath, only the default config would change.

bl4ckb0ne:
I dusted off the MR, revised the approach. There is now a `allow_drm_leasing` config that will offer a given output for lease, non-desktop output are kept the same (only up for lease, not part of the layout).

I have not tested this other than in a nested sway yet.

bl4ckb0ne:
@emersion ping for review pls

lheckemann:
I have this exact problem, and this addresses it nicely (tested successfully). Thanks for writing it!

lheckemann:
Just found myself updating to the latest version of this PR since the version I had no longer applied to sway 1.12 (but the current version is fine!). @emersion is there anything missing for this?

Issue #6926, opened 2022-04-01
client.sticky colorclass


JohnMertz:
Unclear if this is considered a 'window management feature' since it does not change the behaviour of windows, only basic color configuration. I posted the request for [i3](https://github.com/i3/i3/issues/4925), but I honestly only wanted it added there to meet the requirements of it being introduced there before Sway. I'm not confident in seeing any movement there, so I thought I'd try my luck. Sway does diverge from i3 here anyways since it ignores the 'placeholder' class, so hopefully having an additional class would also be acceptable. 

From the i3 issue request:

## Current Behavior
The decoration of a floating window does not change when the 'sticky' flag is toggled.

## Desired Behavior
The ability to define a different colorclass for when the window is stickied.

## Impact
Provide visual feedback to know whether the window is currently in sticky mode.
<pre>
[-] This feature requires new configuration and/or commands
</pre>
Sort of. It would ideally default to the same settings as the floating colorclass if no colorclass for stickied clients exists. An optional configuration with the same format as the existing colorclasses would be available for those who want it.

## Environment
<pre>
Sway version: 1.6.1
</pre>

<pre>
- Linux Distribution & Version: Debian Bookworm
</pre>

nashamri:
I would love to see this implemented.

qiu-x:
@JohnMertz please verify if my PR (#7208) implements this feature as intended.

qiu-x:
> It appears that the styles only apply when the focus/active status changes

Hmm, I thought I fixed this issue already. Well, I guess I missed something and I was fixed only on my setup(which is pretty heavily customized). I will see what I can do about this.

> Thanks a lot for the work so far!

Thanks for the idea!

qiu-x:
@JohnMertz Now the container should be redrawn properly. Please confirm if the issue is also fixed for your configuration.

JohnMertz:
Perfect! Works exactly as expected. My build is completely unmodified aside from those two commits and my config file is almost entirely i3 compatible other than the now client styles, so it should be a pretty faithful test case for a vanilla install.

PR #7208, opened 2022-10-08
sway: Implement `client.sticky` colorclass (#6926)

qiu-x:
This PR implements the feature mentioned in the issue #6926.

More precisely, this implements the following config commands:
`client.sticky`, `client.sticky_inactive` and `client.sticky_unfocused`.
If a option is not set the value is copied form the normal style, to preserve compatibility with i3.


qiu-x:
@emersion could you take a look?

emersion:
Sorry, I'm a bit short on time, and other PRs are already in my Sway review queue. Also we generally don't accept new WM features like this one -- just like i3.

Kommynct:
You should send this to [swayfx](https://github.com/WillPower3309/swayfx)

PR #7212, opened 2022-10-10
input: focus container when scrolling on titlebar

### LINKED ISSUE #6503: Scrolling on lone tab/titlebar doesn't focus the tab
In i3, scrolling on lone tab/titlebar has the same effect a click would.

Xyene:
Fixes #6503.

Tested to behave the same as i3 wrt. inactive containers in this layout:

image


emersion (approved):
Thanks!

PR #7221, opened 2022-10-15
Allow Sway to register Xorg fallbacks to Wayland

brianredbeard:
Approximately three years ago a mechanism was added to increase the reliability of session handling with Wayland (when handled GDM) by removing the need for timeout checks. ([1])

This mechanism utilizes the format extension scheme defined in the Freedesktop.org "Desktop Entry Specification" ([2]) allowing for the addition of new key/value configuration pairs.  This is designed to limit the scope of effect.

[1]: https://gitlab.gnome.org/GNOME/gdm/-/merge_requests/67
[2]: https://specifications.freedesktop.org/desktop-entry-spec/latest/

### DIFF
--- a/sway.desktop
+++ b/sway.desktop
@@ -3,3 +3,4 @@ Name=Sway
 Comment=An i3-compatible Wayland compositor
 Exec=sway
 Type=Application
+X-GDM-SessionRegisters=true


emersion:
> The mechanism is
that they specify X-GDM-SessionRegisters=true in their file, and then
call RegsterSession on the DisplayManager interface on the bus to say
that they've started up

Sorry, but we won't do that. This relies on D-Bus and it's GDM-specific.

brianredbeard:
This does not "rely" on dbus.  This is effectively a type-hint which _allows_ Sway to place nicely in the _presence_ of GDM.

kennylevinsen:
If I understand this correctly, `X-GDM-SessionRegisters=true` indicates that the started session will call `org.gnome.DisplayManager.Manager.RegisterSession` to indicate successful start, used to distinguish whether an exit is due to session start having failed or the user ending the session. When not set, GDM instead assumes that a session running more than 10 seconds is a success. If the option is set without calling the dbus method, any process exit would seemingly be seen as a failed start.

So yes, this "relies" on dbus, as we must cal a dbus method when set. Setting it without doing the dbus part means we *worsen* our behavior in the presence of GDM by lying to it and violating its subsequent expectations.

PR #7225, opened 2022-10-17
feat: set xdg-current-desktop envirenment in main.c

Decodetalkers:
Then the env will be sway as default, not needed to be set by people
I think it should be better

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -256,7 +256,7 @@ static const char usage[] =
 
 int main(int argc, char **argv) {
 	static bool verbose = false, debug = false, validate = false, allow_unsupported_gpu = false;
-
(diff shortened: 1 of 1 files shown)


emersion:

Decodetalkers:
> Duplicate of #4876

I have view the pr.But I does not use hardcode, if user has already set the variable, the variable will not be setted

Decodetalkers:
And I can check if the variable is set, is set, it will not load the default variable, do you think it is acceptable?

Decodetalkers:
> Duplicate of #4876

Now I will check the variable at first , if has already setted, I will not set the variable again. If now it will be acceptable?

emersion:
No. Sway is not responsible for setting this variable.

PR #7233, opened 2022-10-28
container_floating_set_default_size: Store workspace size box on the stack

Nefsen402:
A trivial patch I had floating around for a really long time.


emersion (approved):
Thanks!

PR #7238, opened 2022-11-04
feat: grimshot supports save and copy

orthros:
Adds new action savecopy which saves then copies

Fixes: 7116


emersion:

PR #7242, opened 2022-11-07
[DO NOT MERGE] Add wp-fractional-scale-v1 support

vyivel:
Add the support for https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/149

Requires https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3809

Built on top of https://github.com/swaywm/sway/pull/6844.


riskygit:
I applied [f167200](https://github.com/swaywm/sway/pull/7242/commits/f167200022eef116ace77ae6563617b3ce0367db) and [cb42051](https://github.com/swaywm/sway/pull/7242/commits/cb4205185c70894f95301196f30c832ffeddd5cd) to the latest `scene-graph` branch (for testing purposes) and didn't notice a difference, but then I did a diff on two screenshots to verify... it seems there is no difference at all.

PR #7245, opened 2022-11-08
Use wlr_damage_ring

vyivel:
wlr_output_damage is to be replaced with wlr_damage_ring, so use that.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -10,7 +10,6 @@
 #include <wlr/types/wlr_buffer.h>
 #include <wlr/types/wlr_drm_lease_v1.h>
 #include <wlr/types/wlr_matrix.h>
-#include <wlr/types/wlr_output_damage.h>
 #include <wlr/types/wlr_output_layout.h>
 #include <wlr/types/wlr_output.h>
 #include <wlr/types/wlr_presentation_time.h>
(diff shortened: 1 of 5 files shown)


Nefsen402:
LGTM

vyivel:
Added missing `wlr_damage_ring_finish()` and `wlr_damage_ring_rotate()` calls.

emersion:
Do we need to damage the whole ring when switching out of direct scanout?

Nefsen402:
> Do we need to damage the whole ring when switching out of direct scanout?

I think output_damage_whole is called when exiting direct scanout.

emersion:
> I think output_damage_whole is called when exiting direct scanout.

Ah, yes, that's correct

sway/desktop/output.c
```diff
@@ -848,6 +871,14 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 
 		update_output_manager_config(output->server);
 	}
+
+	if (event->committed & (WLR_OUTPUT_STATE_MODE |
+			WLR_OUTPUT_STATE_TRANSFORM |
+			WLR_OUTPUT_STATE_SCALE)) {
+		wlr_damage_ring_set_bounds(&output->damage_ring,
+			output->width, output->height);
```
emersion, line 879: It would be slightly more convenient to make `wlr_damage_ring_set_bounds` return a `bool` indicating whether the bounds were updated or not.

Or maybe add `wlr_output_event_commit_has_size_changed()`?

emersion (approved):
Regardless, LGTM, thanks!

sway/desktop/output.c
vyivel: `wlr_damage_ring_set_bounds()` returning a `bool` makes sense to me.

Issue #7120, opened 2022-07-29
Crash when disconnecting an output while the screen is locked


baltitenger:
**Sway Version:** 1.8-dev-a5a44ba9

In sway's `wlr_output->events.destroy` listener, it frees the `sway_output` related to the `wlr_output` being destroyed. Wlroots internally uses another listener for this event which calls `lock_surface_destroy`. Then sway's `lock_surface->events.destroy` handler tries to `output_damage_whole` the given `sway_output`. Depending on the order in which these handlers are called, the `sway_output` might be freed by the time `handle_surface_destroy` tries to damage it, causing a use-after-free.

ASan output:
```
ERROR: AddressSanitizer: heap-use-after-free on address 0x6150002c15c0 at pc 0x559ab2653418 bp 0x7ffc792a0fa0 sp 0x7ffc792a0f90
READ of size 8 at 0x6150002c15c0 thread T0
    #0 0x559ab2653417 in output_damage_whole ../sway/sway/lock.c:57
    #1 0x559ab2653417 in handle_surface_destroy ../sway/sway/lock.c:56
    #2 0x7f8002ff25fd in wlr_signal_emit_safe ../wlroots-git/util/signal.c:29
    #3 0x7f8002fe61de in lock_surface_destroy ../wlroots-git/types/wlr_session_lock_v1.c:188
    #4 0x7f8002ff25fd in wlr_signal_emit_safe ../wlroots-git/util/signal.c:29
    #5 0x7f8002fbc378 in wlr_output_destroy ../wlroots-git/types/output/output.c:392
    #6 0x7f8002fbc378 in wlr_output_destroy ../wlroots-git/types/output/output.c:383
    #7 0x7f8002fa67f3 in disconnect_drm_connector ../wlroots-git/backend/drm/drm.c:1548
    #8 0x7f8002fa67f3 in scan_drm_connectors ../wlroots-git/backend/drm/drm.c:1392
    #9 0x7f8002ff25fd in wlr_signal_emit_safe ../wlroots-git/util/signal.c:29
    #10 0x7f8002faf4c8 in handle_udev_event ../wlroots-git/backend/session/session.c:214
    #11 0x7f80030499e1 in wl_event_loop_dispatch (/usr/lib/libwayland-server.so.0+0xb9e1)
    #12 0x7f800304a196 in wl_display_run (/usr/lib/libwayland-server.so.0+0xc196)
    #13 0x559ab2635c96 in server_run ../sway/sway/server.c:307
[...]

PR #7248, opened 2022-11-10
lock: fix crash on output destroy

emersion:
Closes: https://github.com/swaywm/sway/issues/7120


emersion:
Maybe a slightly better approach is to destroy the `sway_session_lock_surface` with the `sway_output`.

emersion:
I've done that.

emersion:
Thanks for the review!

PR #7253, opened 2022-11-11
workspace_create: Don't allow NULL name

Nefsen402:
Through code inspection a null name shouldn't be given. Also, there is existing code that will segfault if the name is NULL, guard against this. The commit is quite old, I haven't hit this assert in running it for the past few months.


emersion (approved):
Thanks!

PR #7263, opened 2022-11-14
Fix PKGBUILD stable

ErikReider:


ErikReider:
Woops, wrong repo.....

Issue #7264, opened 2022-11-14
Build failure


J0nnyMak0:
Build is failing with the following error:

```
[358/582] Compiling C object sway/sway.p/server.c.o
FAILED: sway/sway.p/server.c.o
cc -Isway/sway.p -Isway -I../sway -Iinclude -I../include -Isubprojects/wlroots/include -I../subprojects/wlroots/include -Iprotocols -I/usr/include/cairo -I/usr/include/lzo -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/harfbuzz -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/usr/include/sysprof-4 -I/usr/include/pixman-1 -I/usr/include/libdrm -I/usr/include/json-c -I/usr/include/libevdev-1.0 -I/usr/include/pango-1.0 -I/usr/include/libmount -I/usr/include/blkid -I/usr/include/fribidi -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=c11 -O0 -g -DWLR_USE_UNSTABLE -Wno-unused-parameter -Wno-unused-result -Wno-missing-braces -Wundef -Wvla '-DSYSCONFDIR="//usr/local/etc"' '-DSWAY_VERSION="1.8-dev-85005b52 (" __DATE__ ", branch '"'"'master'"'"')"' -fmacro-prefix-map=../= -pthread -MD -MQ sway/sway.p/server.c.o -MF sway/sway.p/server.c.o.d -o sway/sway.p/server.c.o -c ../sway/server.c
../sway/server.c: In function ‘server_init’:
../sway/server.c:124:31: error: too few arguments to function ‘wlr_layer_shell_v1_create’
  124 |         server->layer_shell = wlr_layer_shell_v1_create(server->wl_display);
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../sway/server.c:21:
../subprojects/wlroots/include/wlr/types/wlr_layer_shell_v1.h:125:28: note: declared here
  125 | struct wlr_layer_shell_v1 *wlr_layer_shell_v1_create(struct wl_display *display,
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~
[375/582] Compiling C object sway/sway.p/desktop_xdg_shell.c.o
ninja: build stopped: subcommand failed.
```

Issue #7260, opened 2022-11-13
wlr-layer-shell-unstable-v1 v4 is advertised, v3 is supported


WhyNotHugo:
### Please fill out the following:
- **Sway Version:**
  - `sway version 1.8-dev-85005b52 (Nov 13 2022, branch 'master')`

- **Description:**

Sway advertises `wlr-layer-shell-unstable-v1` version 4:

```
> wayland-info | grep layer_shell
interface: 'zwlr_layer_shell_v1',                        version:  4, name: 12
```

However, only version 3 is implemented. In particular, `zwlr_layer_surface_v1::keyboard_interactivity:on_demand` is not implemented.

While it would be great to have v4 implemented, this issue is about sway advertising an unsupported version. If sway advertised v3 as supported, then clients could adapt accordingly.

bl4ckb0ne:
Fixed in #7265

WhyNotHugo:
Thanks!

PR #7265, opened 2022-11-14
sway/server: specify wlr-layer-shell version on creation

bl4ckb0ne:
Closes https://github.com/swaywm/sway/issues/7264


sway/server.c
```diff
@@ -49,6 +49,9 @@
 #include "sway/xwayland.h"
 #endif
 
+#define SWAY_XDG_SHELL_VERSION 2
+#define SWAY_LAYER_SHELL_VERSION 4
```
emersion, line 53: To fix https://github.com/swaywm/sway/issues/7260, we need to advertise 3 here.

bl4ckb0ne: Done

emersion (approved):
Thanks!

PR #7270, opened 2022-11-16
docs: Clarify how to use a custom XKB symbol file.

### LINKED ISSUE #7229: Allow specifying xkb include path
Heya,

I have some custom keyboard layouts which I'd like to use with sway. I have these described in `.xkb` files. On Xorg, I compile these to `.xkm` and then load them with `xkbcomp`.

To use custom symbol definitions (without modifying system directories), I run `xkbcomp` with `-I.`. But, I don't see an equivalent option to specify the xkb include path with sway, so sway fails to load the keymap:

```console
$ swaymsg 'input type:keyboard xkb_file "~/work/DvorakLayouts/layouts/xkb/layout-3.xkb"'
Error: Failed to compile keymap: Couldn't find file "symbols/my_ru" in include paths
```

An option equivalent to `xkbcomp`'s `-I` would allow this to work. Alternatively, a hack would be to implicitly add to the include search path the directory which contains the `.xkb` file.

Thanks!

markstos:
This was not covered in `man sway-input` nor `man xkeyboard-config`

Fixes #7229

### DIFF
--- a/sway/sway-input.5.scd
+++ b/sway/sway-input.5.scd
@@ -56,6 +56,10 @@ For more information on these xkb configuration options, see
 	Sets all xkb configurations from a complete .xkb file. This file can be
 	dumped from _xkbcomp $DISPLAY keymap.xkb_. This setting overrides
 	xkb_layout, xkb_model, xkb_options, xkb_rules, and xkb_variant settings.
+	Custom XKB symbol files may placed in _$XDG_CONFIG_HOME/xkb/symbols_ and
+	other locations. See
+	https://xkbcommon.org/doc/current/md_doc_user_configuration.html
+	for details.
 
 *input* <identifier> xkb_layout <layout_name>
 	Sets the layout of the keyboard like _us_ or _de_.


sway/sway-input.5.scd
```diff
@@ -56,6 +56,10 @@ For more information on these xkb configuration options, see
 	Sets all xkb configurations from a complete .xkb file. This file can be
 	dumped from _xkbcomp $DISPLAY keymap.xkb_. This setting overrides
 	xkb_layout, xkb_model, xkb_options, xkb_rules, and xkb_variant settings.
+	Custom XKB symbol files may placed in _$XDG_CONFIG_HOME/xkb/symbols_ and
+	other locations. See
+	https://xkbcommon.org/doc/current/md_doc_user_configuration.html
+	for details.
```
RPigott, line 62: I don't see the need to specifically mention the symbols files when all the xkb user components should work, so I would probably just word this as "xkb files".

rpigott (changes requested):
Sorry for the delay, but now with the secure_getenv issue almost ironed out, user xkb configs should actually work so I think this is a good change.

PR #7279, opened 2022-11-24
Use shm_open instead of mkstemp

mstoeckl:
This continues the replacement of mkstemp by shm_open; see also https://github.com/swaywm/swaybg/pull/49 .

(This change only affects `swaynag` and `swaybar`.)


emersion (approved):
Thanks!

Issue #7106, opened 2022-07-18
Chromium/Chrome 103+ prevents sway keybinds from activating when launched with --app and Ozone


nermolov:
- **Sway Version:** 1.7
- **Debug Log:** https://gist.github.com/nermolov/27af733d353db0f3daf72a686e0e23da
- **Configuration File:** Default
- **Description:**
  - Launch Chrome/Chromium version 103 or newer with `google-chrome --enable-features=UseOzonePlatform --ozone-platform=wayland --app="https://google.com/"` and attempt to use any sway keybindings.

Since Chrome/Chromium version 103 (problem does not occur with versions 102 and older), chrome (ozone) windows launched with the `--app` flag somehow capture all sway keybinds when they are focused. This issue does not happen with normal, non-app windows, nor does it happen when chrome is launched with xwayland.

This issue does not happen under Gnome (Gnome's keybinds work as expected).

PremiumUsername:
I can confirm this bug. For me:
`google-chrome-stable  --profile-directory=Default --app-id=ddifbjdgicpgdgphcocemmlofhjcpnph` 
...will result in the web app consuming all key presses.  All web apps I use are affected by this bug (home-assistant, jellyfin etc.). Just as nermolov said, a regular chrome instance works like normal.

cshei:
Seems like it also happens on KDE and is likely a Chrome regression: https://bugs.chromium.org/p/chromium/issues/detail?id=1338554

AdamantGarth:
Workaround:
```ini
for_window [app_id="^chrome-.*"] shortcuts_inhibitor disable
```

nermolov:
Thanks for the workaround! Works here. Seems to be intended behavior on Sway's side and it being a regression on the chrome side.

PR #7280, opened 2022-11-26
shortcuts_inhibitor: default to disabled

rpigott:
Feels a little odd to me that this one defaults to enabled. Defaulting to disabled will reduce surprises from abusive clients like chroium: #7106. Users can of course enable the inhibitor with shortcuts_inhibitor enable. Thoughts?

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -209,7 +209,7 @@ enum seat_config_allow_constrain {
 };
 
 enum seat_config_shortcuts_inhibit {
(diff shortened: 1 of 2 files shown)


emersion:
Following the discussion in https://bugs.chromium.org/p/chromium/issues/detail?id=1338554, it seems like Chromium will adopt a better behavior, only inhibiting the shortcuts when a webapp is fullscreen and has explicitly requested shortcut inhibition.

rpigott:
Ok, I'll shelve this one for now.

PR #7285, opened 2022-11-27
compositor: Use output renderer instead of global state

Nefsen402:
Some annoyance I had to work through to port sway to use wlroots!3887 where global renderers cannot be used for multigpu anymore.

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -7,6 +7,7 @@
 #include <strings.h>
 #include <sys/stat.h>
 #include <wayland-server-core.h>
+#include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_linux_dmabuf_v1.h>
 #include <wlr/types/wlr_output_layout.h>
 #include <wlr/types/wlr_subcompositor.h>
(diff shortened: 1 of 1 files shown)


emersion:
Hm, but here we actually want the main renderer, ie. the renderer used as a fallback when more optimized paths fail.

Nefsen402:
If you're worried about the dmabuf feedback tranche suggesting modifiers not supported by the primary GPU, the multigpu patches will blit to a staging buffer in this case. This may be slower, but only in rare cases where for whatever reason the client does not recreate the swapchain using a compatible modifier after coming out of fullscreen.

emersion:
The renderer tranche here (tranche 2 out of 2) is the fallback tranche. The optimized tranche (tranche 1 of 2) is the scanout tranche.

Nefsen402:
That should still be correct. Since renderers are allocated based on the local device for each connector, it's okay for the client to choose rendering modifiers only supported by the local GPU. In fact, we guarantee this to be the case b/c these tranches only get sent for fullscreen apps.

emersion:
I don't understand the above. In any case, I'm not comfortable merging this right now, too much unknowns regarding upcoming multi-GPU work. I don't want to shoot myself in the foot with this.

PR #7286, opened 2022-11-27
input/keyboard: enable user-specific keymaps

RPigott:
This adds the xdg dirs to the lookup path when compiling xkb keymaps. With these the user can put custom symbol/rules files in $XDG_CONFIG_HOME/xkb to define user specific xkb options and layouts without modifying /usr/share/X11/xkb.

I was surprised to find this did not already work. If I have to include them, they're not exactly defaults now are they, libxkbcommon?

Fixes : #7229 #7270


jkonecny12:
Sounds like a great improvement. Thanks!

RPigott:
We still miss these when running un-nested for some reason.

RPigott:
xkbcommon refuses to add search paths when sway is cap_sys_nice b/c of its use of secure_getenv.

PR #7287, opened 2022-11-27
Fix build on Debian Stable

n3rdopolis:


swaynag/meson.build
```diff
@@ -9,6 +9,7 @@ executable(
 	],
 	include_directories: [sway_inc],
 	dependencies: [
+		rt,
 		cairo,
```
Xyene, line 13: Nit: let's keep these in alphabetical order.

emersion:
Thanks!

PR #7288, opened 2022-11-27
Fix build on Debian Stable

n3rdopolis:


n3rdopolis:
I hope I did this correctly for 1.8

PR #7292, opened 2022-11-28
Fix typo in Czech README translation

AsciiWolf:


jkonecny12 (commented):
Maybe I'm wrong but this seems like incorrect correction to me. Was was there before seems like the correct variant?

Could you more explain the typo?

emersion (approved):
Ah yes, it's not a Czech typo, just a technical Markdown issue. This looks good to me, thanks!

PR #7300, opened 2022-11-30
Add support for external launcher clients with xdg activation

rpigott:
This allows a launcher client to request an activation token useful for startup notifications. For a compatible launcher, see fuzzel [1].

For the time being, we use these in exactly the same way we would our own internal launch contexts except they cannot be pid matched. If the launched client supports startup notifications I'd consider that a positive. We also get a little more info from the launcher than we have on our own, i.e. the intended app_id. We might consider restricting matches to the given app_id.

Depends on [2].

[1] https://codeberg.org/dnkl/fuzzel/pulls/195 (merged)
[2] https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3907 (merged)

### DIFF
--- a/sway/xdg_activation_v1.c
+++ b/sway/xdg_activation_v1.c
@@ -1,6 +1,7 @@
 #include <wlr/types/wlr_xdg_activation_v1.h>
 #include "sway/desktop/launcher.h"
 #include "sway/tree/view.h"
+#include "sway/tree/workspace.h"
 
 void xdg_activation_v1_handle_request_activate(struct wl_listener *listener,
 		void *data) {
@@ -36,5 +37,17 @@ void xdg_activation_v1_handle_request_activate(struct wl_listener *listener,
 	if (wlr_seat) {
 		struct sway_seat *seat = wlr_seat->data;
 		view_request_activate(view, seat);
+	}
+}
+
+void xdg_activation_v1_handle_new_token(struct wl_listener *listener, void *data) {
+	struct wlr_xdg_activation_token_v1 *token = data;
+	struct sway_seat *seat = token->seat ? token->seat->data :
+		input_manager_current_seat();
+
+	struct sway_workspace *ws = seat_get_focused_workspace(seat);
+	if (ws) {
+		launcher_ctx_create(token, &ws->node);
+		return;
 	}
 }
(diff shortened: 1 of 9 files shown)


RPigott:
Before:
1. Launch a foot server, `swaymsg exec -- foot -s` with workspace 1 active.
2. switch to workspace 2 and use unpatched fuzzel to launch footclient. sway erroneously pid-matches footclient to workspace 1.

After:
4. Kill and launch a new foot server with workspace 1 active.
5. switch to workspace 2 and use patched fuzzel to launch footclient. footclient uses the passed token to activate on workspace 2.

sway/xdg_activation_v1.c
```diff
@@ -35,5 +35,6 @@ void xdg_activation_v1_handle_request_activate(struct wl_listener *listener,
 		return;
 	}
 
-	view_request_activate(view);
+	struct sway_seat *seat = (struct sway_seat *)event->token->seat->data;
```
emersion: I think `event->token->seat` may be NULL here, if the client didn't set it.

emersion: Ah, it seems like the next commit contains the logic to handle this case.

emersion:
Overall looks pretty solid to me!

sway/xdg_activation_v1.c
rpigott: I don't think the client has the option to create a token w/o a seat, as it is ultimately derived from the seat associated with the serial provided. The seat may still be null if it was destroyed after a token was granted. I was trying not to change behavior yet, but yeah the null check rightfully belongs in this commit. I'll move it here.

rpigott:
I dunno what's the deal with the alpine ci.

sway/xdg_activation_v1.c
emersion: The client does have the option to not specify the seat when creating the activation token, FWIW.

emersion (approved):
LGTM, thanks!

CI is fixed by #7430.

sway/xdg_activation_v1.c
rpigott: For some reason I was under the impression wlroots would reject those but I think you're right. Should be good either way now anyhow.

rpigott:
Thanks for the CI fix, seems like this is good to go.

Issue #6654, opened 2021-11-08
Send fewer redundant keymap events


RPigott:
I think the following issues are probably all the same EMFILE error: #5757, #6310, #6563, #6640, #6642, and maybe more.

Sway currently sends a lot of redundant keymap events which tends to create a lot of fds, and I would assume this is the biggest offender causing EMFILE related bugs. Even with reasonable number clients/keyboards it's not hard to climb over 1k+.

#6629 probably helps, but need more feedback in case there is a fd leak somewhere. Also the user might still have a low hard limit.

RPigott:
Just starting to look into this and here's a quick debug log from foot truncated to contain only the events generated by a reload. Here the client is focused during the reload:

<details>
<summary>WAYLAND_DEBUG=1 foot</summary>

```
[2558429.106] wl_keyboard@20.modifiers(6629, 65, 0, 0, 0)
[2559008.379] wl_keyboard@20.keymap(1, fd 13, 64354)
[2559022.400] wl_keyboard@20.repeat_info(25, 600)
[2559022.425] wl_keyboard@20.modifiers(6631, 0, 0, 0, 0)
[2559022.444] wl_keyboard@20.leave(6632, wl_surface@3)
[2559022.456] wl_keyboard@20.enter(6634, wl_surface@3, array[0])
[2559022.467] wl_keyboard@20.modifiers(6635, 0, 0, 0, 0)
[2559022.483] wl_data_device@15.selection(nil)
[2559022.488] zwp_primary_selection_device_v1@16.selection(nil)
[2559022.493] wl_keyboard@20.keymap(1, fd 14, 64354)
[2559039.809] wl_keyboard@20.repeat_info(25, 600)
[2559039.833] wl_keyboard@20.modifiers(6636, 65, 0, 0, 0)
[2559039.853] wl_keyboard@20.keymap(1, fd 15, 64354)
[2559059.338] wl_keyboard@20.repeat_info(25, 600)
[2559059.361] wl_keyboard@20.modifiers(6637, 0, 0, 0, 0)
[2559059.380] wl_keyboard@20.keymap(1, fd 16, 64354)
[2559073.262] wl_keyboard@20.repeat_info(25, 600)
[2559073.287] wl_keyboard@20.modifiers(6638, 0, 0, 0, 0)
[2559073.306] wl_keyboard@20.keymap(1, fd 17, 64354)
[2559086.984] wl_keyboard@20.repeat_info(25, 600)
[...]

Issue #6637, opened 2021-10-26
Sometimes firefox and thunderbird are terminated on sway reload


532910:
```
% firefox 
Gdk-Message: 15:04:50.996: Lost connection to Wayland compositor.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
zsh: exit 1     firefox
```

sway 1.6 from debian experimental

RobertMueller2:
Hi,

this issue sounds familiar. I could reproduce this with Sway master, although I'd say it took a lot more tries to reproduce than with Debian's 1.6. And this time, only one application was affected. Usually I'm seeing a lot more apps exiting at the same time when reloading.

Please see attached log. I think line 343657 is where it happened.

[Wayland-Debug-Log.txt](https://github.com/swaywm/sway/files/7465606/Wayland-Debug-Log.txt)

I can try some more if I find some time on the weekend. As far as I recall, it's not just firefox and thunderbird that are affected but more GUI apps, but maybe it would be good to name the others as well. ;)

nicolasavru:
Also in the firefox bug tracker: https://bugzilla.mozilla.org/show_bug.cgi?id=1652820

RPigott:
I would bet this is also #6654.

torgiren:
Summary from firefox ticket:
Try to comment out "input * ..." lines from config.

I changed:
```
input * {
   xkb_layout "pl"
}
```
into
```
input 1133:16419:Logitech_Wireless_Keyboard_PID:4023 {
    xkb_layout "pl"
}
input 1:1:AT_Translated_Set_2_keyboard {
    xkb_layout "pl"
}
```
and that fixed my problem

WhyNotHugo:
This is, at least in my case, due to https://github.com/swaywm/sway/issues/6654

mindrunner:
Happens to me 100% on the sway-reloads. Same for thunderbird

hakanyi:
This used to happen to me 100% of reloads, but doesn't anymore. 
```bash
sway version 1.6.1
Mozilla Firefox 95.0.2
Thunderbird 91.4.1
```

mindrunner:
Any idea what changed on your side? Still an issue for me:

```
sway version 1.6.1
Mozilla Firefox 95.0.2
[...]

PR #7309, opened 2022-12-04
seat: Avoid sending redundant keymaps on reload

kennylevinsen:
Only update the currently active keyboard for the seat, and clear the active keyboard before reset. Vastly reduces keymap events WAYLAND_DEBUG output, and should avoid the file descriptor floods. Could use a bit of testing though.

Fixes: https://github.com/swaywm/sway/issues/6654


RPigott (approved):
Simple and effective. LGTM.

OctopusET:
Will this be backported to 1.8? It increases stability a lot.

tigerjack:
I confirm it fixes https://github.com/swaywm/sway/issues/6637 
Also, config reload seems way faster now.

emersion:
Yes, it will be backported.

PR #7311, opened 2022-12-04
input/keyboard: enable user xkb configs with cap_sys_nice

### LINKED ISSUE #7229: Allow specifying xkb include path

RPigott:
supercedes #7286, fixes #7229

This version partially duplicates the append_defaults behavior of libxkbcommon when AT_SECURE is set, enabling user configuration in the XDG configuration directories in case of cap_sys_nice. Kinda lame but it works.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -1,12 +1,14 @@
 #include <assert.h>
 #include <limits.h>
 #include <strings.h>
+#include <sys/auxv.h>
 #include <wlr/config.h>
 #include <wlr/backend/multi.h>
 #include <wlr/interfaces/wlr_keyboard.h>
 #include <wlr/types/wlr_idle.h>
 #include <wlr/types/wlr_keyboard.h>
 #include <wlr/types/wlr_keyboard_group.h>
+#include <xkbcommon/xkbcommon.h>
 #include <xkbcommon/xkbcommon-names.h>
 #include "sway/commands.h"
 #include "sway/input/input-manager.h"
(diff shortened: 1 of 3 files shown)


sway/input/keyboard.c
```diff
@@ -761,6 +802,10 @@ struct xkb_keymap *sway_keyboard_compile_keymap(struct input_config *ic,
 	xkb_context_set_user_data(context, error);
 	xkb_context_set_log_fn(context, handle_xkb_context_log);
 
+	if (getauxval(AT_SECURE) > 0) {
```
emersion, line 805: From the man page:

> This function is a nonstandard glibc extension.

RPigott: Oh hm, that's disappointing.

RPigott: Meh, secure_getenv is itself also a GNU extension or I'd consider just trying that. I think I'll investigate if xkbcommon is willing to drop secure_getenv here or maybe provide some way for us to get around it.

PR #7312, opened 2022-12-05
Disable primary clipboard option

AidanGG:
See https://github.com/swaywm/sway/issues/4511. This PR builds on https://github.com/swaywm/sway/pull/6897 which seems to have stalled, and I have kept @t1lde's contributions in the commit history.

I have attempted to address @emersion's feedback in https://github.com/swaywm/sway/pull/6897. Firstly, the `config` option is stored as `bool`. Secondly, I make the `primary_selection` only configurable at launch (similar to `xwayland`). This allows me to destroy the primary selection global after the config is loaded.

I couldn't think of a clean way to avoid creating the global at all, since `server_init()` creates the global and the default seat, and then registers listeners. After that, the configuration is loaded in `load_main_config()`. So I hope destroying the global after loading the config is an acceptable compromise.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -165,6 +165,7 @@ sway_cmd cmd_no_focus;
 sway_cmd cmd_output;
 sway_cmd cmd_permit;
 sway_cmd cmd_popup_during_fullscreen;
+sway_cmd cmd_primary_selection;
 sway_cmd cmd_reject;
 sway_cmd cmd_reload;
 sway_cmd cmd_rename;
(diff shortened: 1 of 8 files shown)


emersion:
We can move the global creation to `server_start`. The Xwayland option works in a similar way.

AidanGG:
I moved the global creation into `server_start` and it seems to work as intended. Even though listeners for `seat0` are still registered as before, they have no effect without the global.

emersion (approved):
LGTM!

Can you squash these two commits together?

emersion:
Feel free to use `Co-authored-by` to indicate the mixed authorship.

AidanGG:
Thank you, and done!

emersion:
Thanks for your contribution!

Issue #7303, opened 2022-12-01
Sway 1.8-dev crash on window close


hissssst:
- **Sway Version:**
  `sway version 1.8-dev`

- **Debug Log:**
  [sway2.log](https://github.com/swaywm/sway/files/10135158/sway2.log)

- **Configuration File:**
https://0x0.st/ok-4.txt

- **Stack Trace:**
No stack trace, just `(core dumped) exited with code 1`

- **Description:**
This bug rarely happens, but when I close firefox window (`Mozilla Firefox 102.5.0esr`), sway crashes

bl4ckb0ne:
```
11:36:33.556 [DEBUG] [wlr] [backend/drm/renderer.c:315] Failed to get DMA-BUF from buffer
11:36:33.556 [DEBUG] [wlr] [backend/drm/drm.c:423] connector eDP-1: Failed to import buffer for scan-out
```

Strange.

Do you run firefox with the wayland backend?

hissssst:
> Do you run firefox with the wayland backend?

Yes

hissssst:
> Are you able to provide a stacktrace?

How? The process just finishes up without any stacktrace and shell shows `(core dumped) exited with code 1`

bl4ckb0ne:
I don't know what distro you're using, coredumpctl, gdb trace, anything that can point to a problem in the code. We can't do much with a "that bug rarely happens" and a sway log that has no significant error.

hissssst:
[sway.coredump.txt](https://github.com/swaywm/sway/files/10144920/sway.coredump.txt)

Here you go

hissssst:
But I think this is a coredump from a latest crash. However, it is caused by the same actions (closing firefox)

hissssst:
@emersion , what does "waiting on reporter" label mean? Do I need to provide more information?

hissssst:
@bl4ckbone I've found a way to reproduce it
1. Open any window
2. Move it to scratchpad
3. Hide it
4. Kill the window process

hissssst:
Maybe next week. If you were able to reproduce the issue and this commit fixes it, just close it

bl4ckb0ne:
@hissssst i made some changes to the patch. If you could test it, it would be greatly appreciated.

emersion:
Can you confirm this is fixed in master?

Zeioth:
I just tested master and I got the same

[...]

PR #7126, opened 2022-08-02
container_get_siblings: handle NULL workspace

baltitenger:
Fixes #7123

Issue #7123, opened 2022-07-30
Crash when moving a fullscreen container to another workspace (ASan only)


baltitenger:
### Please fill out the following:
- **Sway Version:** 1.8-dev-9e879242 (Jul 30 2022, branch 'master')

- **Stack Trace:**
```
[ASan stuff]
#10 0x000055ff755e0c72 in container_get_siblings (container=0x616000103e80) at ../sway/sway/tree/container.c:1415
#11 0x000055ff7563d138 in get_deco_rect.constprop.0 (c=0x616000103e80, deco_rect=deco_rect@entry=0x7ffd4f7b0820) at ../sway/sway/ipc-json.c:461
        parent_layout = L_NONE
        siblings = <optimized out>
        tab_or_stack = <optimized out>
#12 0x000055ff75412bbc in ipc_json_describe_node (node=0x616000103e80) at ../sway/sway/ipc-json.c:693
        count = <optimized out>
        seat = 0x616000007580
        focused = <optimized out>
        name = 0x6030009dfcf0 "baltazar@pluto:~"
        focus = <optimized out>
        object = <optimized out>
#13 0x000055ff754183eb in ipc_json_describe_node_recursive (node=0x616000103e80) at ../sway/sway/ipc-json.c:730
        object = <optimized out>
        i = <optimized out>
        children = <optimized out>
#14 0x000055ff75437b52 in ipc_event_window (change=0x55ff75689a00 "fullscreen_mode", window=0x616000103e80) at ../sway/sway/ipc-server.c:332
        obj = 0x6040008819d0
        json_string = <optimized out>
#15 ipc_event_window (window=0x616000103e80, change=0x55ff75689a00 "fullscreen_mode") at ../sway/sway/ipc-server.c:324
#16 0x000055ff755dfcf4 in container_fullscreen_disable (con=0x616000103e80) at ../sway/sway/tree/container.c:1255
        __PRETTY_FUNCTION__ = "container_fullscreen_disable"
#17 0x000055ff75615858 in container_replace (container=0x616000103e80, replacement=0x616000184e80) at ../sway/sway/tree/container.c:1529
        fullscreen = FULLSCREEN_WORKSPACE
        scratchpad = false
        ws = 0x0
#18 0x000055ff75611836 in container_split (child=0x616000103e80, layout=L_TABBED) at ../sway/sway/tree/container.c:1603
        seat = 0x616000007580
        set_focus = true
        cont = 0x616000184e80
[...]

PR #7315, opened 2022-12-06
sway/ipc-json: don't get decoration box for containers in scratchpad

bl4ckb0ne:
Fixes https://github.com/swaywm/sway/issues/7303

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -1404,14 +1404,14 @@ enum sway_container_layout container_current_parent_layout(
 }
 
 list_t *container_get_siblings(struct sway_container *container) {
-	if (container->pending.parent) {
-		return container->pending.parent->pending.children;
-	}
-	if (container_is_scratchpad_hidden(container)) {
+	if (container_is_scratchpad_hidden_or_child(container)) {
 		return NULL;
 	}
 	if (!container->pending.workspace) {
 		return NULL;
+	}
+	if (container->pending.parent) {
+		return container->pending.parent->pending.children;
 	}
 	if (list_find(container->pending.workspace->tiling, container) != -1) {
 		return container->pending.workspace->tiling;


emersion:
Regressed by https://github.com/swaywm/sway/commit/a5a44ba95bdcc3adfe88e282fd206181528e60f6

cc @baltitenger

bl4ckb0ne:
I wasnt able to reproduce the issue from #7303, do the scratchpad containers have a geometry?

emersion:
Yes. In other words this patch introduces a regression: `deco_rect` was sent before, but isn't anymore.

RPigott:
Is #7303 not a dupe of #7123? Either way testing `con->scratchpad` here is not correct.

A scratchpad container has `con->scratchpad == true`

A scratchpad hidden container has `con->scratchpad == true && !con->pending.workspace`.

A scratchpad hidden split container child (parent is scratchpad hidden) has `!con->pending.workspace`, but only the top container in the tree has `scratchpad == true`.

#7126 added a `!con->pending.workspace` check but maybe it is better placed in this function.

We should keep the deco_rect field regardless, but its fine to return zeros for hidden containers imo.

emersion (dismissed):
Thanks, looks good now!

RPigott:
Scratchpad hidden split containers do have children, so those children have siblings, and I'm not sure its a good idea to say they don't. E.g. in sway it is possible to move a container into a scratchpad hidden parent as long as you can match it with criteria. It might be reasonable to refuse these actions, but we shouldn't crash on them:

<details>
<summary>move to child of scratchpad hidden split container w/ this patch</summary>

```
#0  0x00005555555d79dc in list_find (list=0x0, item=0x555555ba8100) at ../common/list.c:71
#1  0x00005555555c796e in container_add_sibling (fixed=0x555555ba8100, active=0x555555c61770, after=true) at ../sway/tree/container.c:1469
#2  0x00005555555ae85f in container_move_to_container (container=0x555555c61770, destination=0x555555ba8100) at ../sway/commands/move.c:250
#3  0x00005555555af733 in cmd_move_container (no_auto_back_and_forth=false, argc=2, argv=0x55555587ea48) at ../sway/commands/move.c:568
#4  0x00005555555b0dc1 in cmd_move (argc=2, argv=0x55555587ea48) at ../sway/commands/move.c:1045
#5  0x000055555556c3a6 in execute_command (_exec=0x555555c4fa20 "move container to mark test", seat=0x5555558a48e0, con=0x0) at ../sway/commands.c:294
#6  0x0000555555578f13 in ipc_client_handle_command (client=0x555555c63820, payload_length=27, payload_type=IPC_COMMAND) at ../sway/ipc-server.c:638
#7  0x0000555555577b16 in ipc_client_handle_readable (client_fd=55, mask=1, data=0x555555c63820) at ../sway/ipc-server.c:263
#8  0x00007ffff78349e2 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#9  0x00007ffff7835197 in wl_display_run () at /usr/lib/libwayland-server.so.0
#10 0x000055555557ca9f in server_run (server=0x5555555fcb60 <server>) at ../sway/server.c:325
#11 0x000055555557bad8 in main (argc=3, argv=0x7fffffffde78) at ../sway/main.c:415
```
</details>

^ caller gets confused because a container with a parent has no siblings.

The problem in #7123 is that we can't really rely on the invariant `!con->pending.workspace` <=> "container is scratchpad hidden or child" in most places we would like because the layout code breaks this property by detaching containers all the time. The fix is to just check what we mean, `!con->pending.workspace` instead of `container_is_scratchpad_hidden` in container_get_siblings so it's safe and correct to use everywhere, which is the effect of #7126.

Callers will still need to check the return value of this function if they didn't previously guarantee it isn't called on detached containers. Some existing calls already guard against this directly:

https://github.com/swaywm/sway/blob/c32a507303e38c7bf0b8054108bec45ff67e92c2/sway/tree/container.c#L1565-L1566

But I _think_ prior to #7113 all callers of `container_get_siblings` adhered to this unspoken rule in one way or another. I wouldn't be at all surprised if there were more similar bugs. E.g. just below in *describe_node using the parent layout effectively performs the same check:

https://github.com/swaywm/sway/blob/c32a507303e38c7bf0b8054108bec45ff67e92c2/sway/ipc-json.c#L775-L776

The scratchpad is special cased in a lot of places so this pitfall may have just been accidentally avoided. There is no point in keeping this footgun around so #7126 is a good change, except the `container_is_scratchpad_hidden` check is now redundant.

I'm guessing #7303 was already fixed by #7126 as I am not able to reproduce it on master. I think the current code in master is correct, except for the redundant check just mentioned, and that get_deco_rect should probably immediately return 0's when parent_layout == L_NONE instead of calling container_get_siblings and crashing (or waiting around to see that it is NULL on master):

https://github.com/swaywm/sway/blob/c32a507303e38c7bf0b8054108bec45ff67e92c2/sway/ipc-json.c#L529-L531

Yes this is all pretty annoying and ugly and could use a rework.

sway/tree/container.c
```diff
@@ -1404,15 +1404,15 @@ enum sway_container_layout container_current_parent_layout(
 }
 
 list_t *container_get_siblings(struct sway_container *container) {
-	if (container->pending.parent) {
-		return container->pending.parent->pending.children;
-	}
-	if (container_is_scratchpad_hidden(container)) {
+	if (container_is_scratchpad_hidden_or_child(container)) {
```
RPigott, line 1407: Returning NULL for children of scratchpad hidden split containers may crash in certain operations. See comment.

RPigott (changes requested):

emersion:
I believe per the comment above this PR is not needed.

PR #7318, opened 2022-12-08
ci: install Xwayland .pc file

emersion:
See https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3621


RPigott (approved):
LGTM.

PR #7278, opened 2022-11-24
Add libinput RotationAngle

zampierilucas:
This patch adds the libinput option RotationAngle to sway.
Fix: #7277 

Signoff-by: Lucas Zampieri <lzampier@redhat.com>

PR #7321, opened 2022-12-09
Fix input command order (rotation_angle)

baltitenger:
#7278 added the `rotation_angle` command but it didn't insert it in the proper place in the list, so the `repeat_delay` and `repeat_rate` commands became unusable.


emersion (approved):
Thanks!

PR #7326, opened 2022-12-15
input: enable user xkb configs with cap_sys_nice

rpigott:
replaces: #7311
fixes: #7229

This version uses a new xkb context flag to avoid secure_getenv in xkbcommon. We don't support suid operation so right now this flag is used unconditionally.

depends on: xkbcommon/libxkbcommon#312 (merged)


mhmdanas:
Just wondering, any reason this PR is still a draft? I would like to include this patch in Void Linux, but I want to confirm that there isn't some flaw first.

(libxkbcommon 1.5.0 is out with https://github.com/xkbcommon/libxkbcommon/pull/312 merged by the way)

rpigott:
Seems like all the CI platforms have this release now, so it should be good to go.

emersion (approved):
LGTM, thanks!

PR #7332, opened 2022-12-18
commands/move: Warp cursor after moving workspace to another output

### LINKED ISSUE #7027: Moving a workspace to a different output does not move the mouse
In i3 the `move workspace to output <direction>` command moves both the workspace and the mouse to the new output. In sway the workspace and focus is moved to the new output, but the mouse is left behind.

My suggestion would be to tie this behavior to the `mouse_warping` setting. If it is not set to `none`, the mouse should move to the new output when the workspace is moved.

anpandey:
This makes sway's behavior consistent with i3 when `mouse_warping` is set to any value besides `none`.

Fixes #7027.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -686,6 +686,9 @@ static struct cmd_results *cmd_move_workspace(int argc, char **argv) {
 	arrange_output(old_output);
 	arrange_output(new_output);
 
+	struct sway_seat *seat = config->handler_context.seat;
+	seat_consider_warp_to_focus(seat);
+
 	return cmd_results_new(CMD_SUCCESS, NULL);
 }
 


RPigott:
I don't have a spare monitor to test this on right now, but it appears i3 does something more like the `cursor_warp_to_workspace` function in sway:

https://github.com/i3/i3/blob/ed690c7ba0f3d4e972ebd5a40acf964f7b939754/src/workspace.c#L543-L547

Also, it appears sway allows the user to match a node with criteria to move its associated workspace. Does i3 allow this? Based on the above it seems like i3 will warp the mouse to the moved workspace regardless of the prior state of its focus. Again, I don't have a spare monitor to test this atm.

If i3 does allow moving inactive workspaces and does warp as described, that seems like a bug we don't need to painstakingly reproduce, so I would be fine with warping to focus here instead, but maybe we should consider skipping the warp altogether in that case, so that the cursor in the actually active workspace is undisturbed. Thoughts?

anpandey:
>it appears i3 does something more like the cursor_warp_to_workspace function in sway

i3 is using the focused container's box in the linked snippet. I tried testing to confirm this: i3 warps the cursor to the center of the workspace if it's empty or to the center of the focused container if it's not. This is consistent with what `seat_consider_warp_to_focus` does.

>it appears sway allows the user to match a node with criteria to move its associated workspace. Does i3 allow this?

i3 does, but the results from testing it makes me think it's unsupported. In both of the scenarios that I tried i3 does something wrong:

- When the matched workspace is moved to the current output, the cursor warps to the new focused container anyway (even though we didn't change outputs).
- When the matched workspace is moved to another unfocused output, both the focused and matched workspaces are moved to that output. This is definitely a bug.

This patch does what IMO is the right thing in both scenarios: it only warps the cursor if `mouse_warping` is set to `container` in the first scenario, and doesn't run into the bug in the second one.

>maybe we should consider skipping the warp altogether in that case, so that the cursor in the actually active workspace is undisturbed

I don't think there's a case where we need to skip the warp. Even when we match on an unselected workspace sway switches the focus to that workspace after it's moved. This is also what i3 does (but then again, see above).

RPigott (approved):
Ok, thanks for testing these scenarios. This patch LGTM.

RPigott:
Thanks!

Issue #7330, opened 2022-12-16
Sway fails to send wl_keyboard::enter events after vt switch


RPigott:
Reported in: https://codeberg.org/dnkl/foot/issues/1236
Regressed by: e1b268af98ed (#7309)

<details>
<summary>git bisect log</summary>

```
$ git bisect log
git bisect start
# status: waiting for both good and bad commits
# bad: [fbf9191fb8373e6718c61586329aaf362eeaa0c0] Fix input command order (rotation_angle)
git bisect bad fbf9191fb8373e6718c61586329aaf362eeaa0c0
# status: waiting for good commit(s), bad commit known
# good: [17ee30d07058bed7a5b5aaede3367a9e2f22d0af] build: bump wlroots dependency to 0.16.0
git bisect good 17ee30d07058bed7a5b5aaede3367a9e2f22d0af
# good: [1c69d0e72fe868d0fa2f6d7d9f9b4b3d7a3f78b4] config/output: use wlr_output_commit_state
git bisect good 1c69d0e72fe868d0fa2f6d7d9f9b4b3d7a3f78b4
# good: [6c3b35701d70c3ce5fc0ae719739d44b592b4143] ci: install hwdata
git bisect good 6c3b35701d70c3ce5fc0ae719739d44b592b4143
# good: [bdeb9f95651f6c99cc2f4cfb59020ddee202cf36] launcher: fudge the interface a bit
git bisect good bdeb9f95651f6c99cc2f4cfb59020ddee202cf36
# good: [f54d22ca2a53749291c74073c1dadfb504c979bb] build: remove trailing newline at end of file
git bisect good f54d22ca2a53749291c74073c1dadfb504c979bb
# good: [aa03a8fcb54a89752d154605c4b6064c84c9be44] sway/desktop/output: listen to output mode in commit handle
git bisect good aa03a8fcb54a89752d154605c4b6064c84c9be44
# bad: [c32a507303e38c7bf0b8054108bec45ff67e92c2] Add `primary_selection` config option
git bisect bad c32a507303e38c7bf0b8054108bec45ff67e92c2
# bad: [e1b268af98edeb09e570e8855ef64f0719cbafe2] seat: Avoid sending redundant keymaps on reload
git bisect bad e1b268af98edeb09e570e8855ef64f0719cbafe2
# good: [6b3245ac77a881ec2c35f9a420fcea0d000f99c3] sway/lock: handle mode in output commit handler
[...]

PR #7333, opened 2022-12-18
seat: Set keyboard if seat keyboard is NULL

kennylevinsen:
sway sends wl_keyboard.enter on seat focus change and when a keyboard active on a seat is configured. If all keyboards are removed and a keyboard is added back without changing the focused client, no new notify event would be sent despite having keyboard focus. This could lead to key events without notify, which is a protocol violation.

As a quick fix, when configuring a keyboard on a seat where no keyboard is currently active, activate the keyboard so that a focused surface will receive a notify event.

Regressed by: e1b268af98edeb09e570e8855ef64f0719cbafe2
Closes: https://github.com/swaywm/sway/issues/7330

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -1070,6 +1070,13 @@ void sway_keyboard_configure(struct sway_keyboard *keyboard) {
 		}
 	}
 
+	// If the seat has no active keyboard, set this one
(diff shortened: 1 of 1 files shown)


RPigott (approved):
This LGTM and works. Waiting to make sure @emersion has a chance to see it first.

Issue #7186, opened 2022-09-26
Swaynag config general settings not being used


cotsuka:
### Please fill out the following:
- **Sway Version:** 1.8-dev-b0fc8348

- **Debug Log:**
  - [sway.log](https://github.com/swaywm/sway/files/9642228/sway.log)

**Configuration File:**
~/.config/swaynag/config:
```
font=sans-serif 10
edge=top
layer=overlay

border-bottom-size=1
message-padding=4
details-border-size=1
button-border-size=1
button-gap=8
button-dismiss-gap=8
button-margin-right=8
button-padding=4

[warning]
button-background=EBCB8B
details-background=EBCB8B
background=EBCB8B
text=2E3440
button-text=2E3440
border=2E3440
border-bottom=2E3440
```

~/.config/sway/config
```
bindsym $mod+Shift+e exec swaynag -t warning -m 'Do you really want to exit sway? This will end your Wayland session.' \
        -b 'Shutdown' 'systemctl poweroff' \
        -b 'Reboot' 'systemctl reboot' \
        -b 'Logout' 'swaymsg exit'
```

- **Description:**
  - Create config file in `~/.config/swaynag/config`
  - Exec `swaynag -t warning -m 'Test'`
  - Swaynag isn't using the font and overlay settings as set in the config file. The type-related items all seem to be set correctly.

llyyr:
Regression caused by https://github.com/swaywm/sway/commit/80e386fd97c0da00970f0acc007574151048cfbf

reverting the commit fixes the issue

emersion:
Can you try https://github.com/swaywm/sway/pull/7336?

llyyr:
#7336 fixes it, thanks!

llyyr:
I misread the original issue, this only fixes the font config being ignored. The layer option still does nothing

PR #7336, opened 2022-12-21
swaynag: fix NULL font description 

emersion:
The font description was only set if provided on the CLI. It was
left NULL for the defaults and when reading from the config file.

Closes: https://github.com/swaywm/sway/issues/7186 

cc @WhyNotHugo

### DIFF
--- a/include/swaynag/types.h
+++ b/include/swaynag/types.h
@@ -4,7 +4,6 @@
 struct swaynag_type {
 	char *name;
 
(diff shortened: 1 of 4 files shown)


WhyNotHugo:
Sorry for breaking this, thanks for fixing it.

Changes LGTM. Haven’t tested the changes tho, and won’t be able to for a few days at least.

PR #7337, opened 2022-12-21
Added splith and splitv to workspace_layout command

mmhat:
This PR extends the `workspace_layout` command such that one can set `splith` or `splitv` as the default workspace layout. This is for example useful if you are on an ultra-width screen where you have enough space to position new containers next to each other by default.


anpandey:
This shouldn't be necessary, since `default_orientation` already controls whether new workspaces are split horizontally or vertically with `workspace_layout` being set to default.

mmhat:
I figured that what I was actually looking for was `workspace_layout default`. Hence closing this one.

PR #7351, opened 2023-01-04
Allow setting the font size to 0

### LINKED ISSUE #7350: [regression] The titlebar can't be hidden anymore since sway 1.8
- **Sway Version:** 1.8

The following configuration mentioned in https://github.com/swaywm/sway/issues/6946, https://github.com/swaywm/sway/issues/5050 and https://github.com/swaywm/sway/issues/1286:
```
font pango:monospace 0
titlebar_border_thickness 0
titlebar_padding 0
```
Worked just fine in Sway 1.7 but doesn't work anymore in 1.8 due to https://github.com/swaywm/sway/pull/7092

As a temporary workaround `font pango:monospace 1` can be used instead of `font pango:monospace 0` but it's not the best solution.

kit-ty-kate:
Fixes #7350 

*side note: i haven't tested locally, I don't have the necessary dependencies to build it at the moment.*

### DIFF
--- a/sway/commands/font.c
+++ b/sway/commands/font.c
@@ -33,10 +33,10 @@ struct cmd_results *cmd_font(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "Invalid font family.");
 	}
 
-	const gint size = pango_font_description_get_size(font_description);
-	if (size == 0) {
+	const PangoFontMask flags = pango_font_description_get_set_fields(font_description);
+	if ((flags & PANGO_FONT_MASK_SIZE) == 0) {
 		pango_font_description_free(font_description);
-		return cmd_results_new(CMD_FAILURE, "Invalid font size.");
+		return cmd_results_new(CMD_FAILURE, "Font size not given.");
 	}
 
 	if (config->font_description != NULL) {


bl4ckb0ne:
As stated in #7350, the size 0 font was a hack to begin with. I'd prefer to see a full feature to hide the title bar rather than bringing back the hack.

joelazar:
@bl4ckb0ne I completely agree, however if this hack was so widely used I would still consider adding it back until the proper feature gets implemented.

kennylevinsen:
> however if this hack was so widely used I would still consider adding it back

Reintroducing the hack just to remove it again later mean that those relying on it get to experience breakage twice.

Waiting for a proper PR sounds like the better option. This is also a good lesson in the reliability of hack solutions. :)

rpigott:
Agreed, if we support this it should just be a config option.

This feature was actually rejected by i3 several times [1] but I'm not very convinced by the stated reasons I could find. AFAICT i3 feels it results in poor visual clarity because borderless tabbed and stacked containers can result in an ambiguous visual representation of the tree.  I don't see this as an issue – sway/i3 have sufficiently rich ipc that users can develop any other indicator they like to disambiguate these cases, or even just use one layout over the other exclusively. I don't feel a possibly ambiguous visual layout is worth preventing (and is already possible without this feature anyway).

I wouldn't be opposed to adding this feature in sway as a new config option.

[1] i3/i3#4198

quackers19 (approved):

vwyu (approved):

emersion:
I've opened https://github.com/swaywm/sway/pull/7408 for the v1.8 branch.

PR #7364, opened 2023-01-08
swaynag: call swaynag_destroy on clean exit

mstoeckl:
This fixes a few memory leaks.

Quoting commit message:
> And fix the fallout of the swaynag_destroy having evolved without being tested:
> * wl_display_disconnect was called too early
> * `button_close` and `swaynag.details.button_details` needed to be heap allocated, since they are added to swaynag.buttons, > and all entries of swaynag.buttons are freed in swaynag_destroy
> * To keep things simpler, disconnect the lifetime of the 'Toggle details' button text config setting from the button itself.

### DIFF
--- a/swaynag/main.c
+++ b/swaynag/main.c
@@ -29,10 +29,12 @@ int main(int argc, char **argv) {
 	wl_list_init(&swaynag.outputs);
 	wl_list_init(&swaynag.seats);
 
-	struct swaynag_button button_close = { 0 };
-	button_close.text = strdup("X");
-	button_close.type = SWAYNAG_ACTION_DISMISS;
-	list_add(swaynag.buttons, &button_close);
+	struct swaynag_button *button_close = calloc(1, sizeof(struct swaynag_button));
+	button_close->text = strdup("X");
+	button_close->type = SWAYNAG_ACTION_DISMISS;
+	list_add(swaynag.buttons, button_close);
+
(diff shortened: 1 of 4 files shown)


swaynag/main.c
```diff
@@ -29,10 +29,12 @@ int main(int argc, char **argv) {
 	wl_list_init(&swaynag.outputs);
 	wl_list_init(&swaynag.seats);
 
-	struct swaynag_button button_close = { 0 };
-	button_close.text = strdup("X");
-	button_close.type = SWAYNAG_ACTION_DISMISS;
-	list_add(swaynag.buttons, &button_close);
+	struct swaynag_button *button_close = calloc(1, sizeof(struct swaynag_button));
```
bl4ckb0ne, line 32: Is it just me or this never gets freed?

swaynag/main.c
```diff
@@ -88,10 +88,11 @@ int main(int argc, char **argv) {
 	swaynag_type_merge(type, swaynag_type_get(types, "<args>"));
 	swaynag.type = type;
 
-	swaynag_types_free(types);
-
 	if (swaynag.details.message) {
-		list_add(swaynag.buttons, &swaynag.details.button_details);
+		swaynag.details.button_details = calloc(1, sizeof(struct swaynag_button));
```
bl4ckb0ne, line 92: Same here

bl4ckb0ne (approved):
LGTM!

Issue #7369, opened 2023-01-09
SIGSEGV in input/tablet.c sway_tablet_pad_notify_enter


ToxicMushroom:
- **Sway Version:**
  - 1.8

- **Debug Log:**
  - https://paste.sr.ht/~pixelhamster/bd2369292c3ecc93c7309306b2af57510ff26c3f

- **Configuration File:**
  - I don't have tablet configuration, so I assume this is irrelevant

- **Stack Trace:**
  - https://paste.sr.ht/blob/f378b7a494e21e02b7d2821bc50193b79ed23401

- **Description:**
  - Using sway with a tablet, I only got this crash once during months of tablet usage.
  - Maybe this is some race condition ?

ToxicMushroom:
I am now able to reproduce by launching sway (tablet plugged in) and hitting mod + l after sway started
`bindsym $mod+l exec --no-startup-id swaylock`
I'll provide more info and debug logs later today

ToxicMushroom:
sway log: 
https://paste.sr.ht/~pixelhamster/bd2369292c3ecc93c7309306b2af57510ff26c3f

bl4ckb0ne:
Could you try sway master to validate the fix in #7371?

ToxicMushroom:
I ran sway-git at 53c9a4a6438824abfb3d539bf55aec92074977e0 and was no longer able to reproduce the issue moments after being able to reproduce it on the community sway pkg.
Thanks a lot for the quick fix :D

bl4ckb0ne:
Great, thanks @vyivel for the fix

PR #7371, opened 2023-01-09
input/tablet: handle focusing NULL surface

vyivel:
Additionally, rename the function responsible for switching focus to match its behavior better.

---

Might fix https://github.com/swaywm/sway/issues/7369, which I can't reproduce.


bl4ckb0ne (approved):
LGTM!

Issue #7202, opened 2022-10-05
XDG_RUNTIME_DIR is not set in the environment (new Arch installation, with Sway)


brianbecsi:
Hello! New user here.

I have a fresh Arch install which I installed using ``archinstall``. I try to run Sway but I get some errors that there was no backend able to open a seat [libseat/libseat.c:79]. I try to sudo but I get an error XDG_RUNTIME_DIR not set. I tried running ``sudo -E sway`` but get the error "unable to drop root ... refusing to start".

Appreciate any help with this I don't know my way around X display.

beatplucker:
Hi there. 
You should add yourself to the `seat` group and start/enable the `seatd.service` and you should be good to go. I suggest you to check out the Sway page in the [Arch wiki](https://wiki.archlinux.org/title/Sway#Starting) for further details.
Good luck!

josteink:
I had this exact issue on both a previously working Ubuntu (22.10) install and a fresh Arch install. 

I can confirm the proposed solution worked on Arch.

josteink:
Ok. So after enabling seatd I tried disabling gdm and just boot straight into the tty. 

For some reason I still got the same error as before. Seemingly I need to keep start/boot into gdm and from there on switch to the tty. 

Then sway works. No idea why it is like that, or what dependency this triggers, but just leaving it here for other people with the same issue.

bl4ckb0ne:
Closing, as this is a distro related issue.

suhail-akhtar:
The XDG_RUNTIME_DIR environment variable is used to specify the directory for user-specific runtime files. If it’s not set, you might encounter the error you’re seeing.
Here’s how you can set it:

export XDG_RUNTIME_DIR=/run/user/$(id -u)

To make this change permanent, you can add the above line to your .bashrc

echo "export XDG_RUNTIME_DIR=/run/user/$(id -u)" >> ~/.bashrc

source ~/.bashrc


This should resolve the XDG_RUNTIME_DIR not set issue

Samueru-sama:
> Closing, as this is a distro related issue.

Please reopen this issue because the XDG Spec says that it can be unset: https://specifications.freedesktop.org/basedir-spec/latest/



> If $XDG_RUNTIME_DIR is not set applications should fall back to a replacement directory with similar capabilities and print a warning message. Applications should use this directory for communication and synchronization purposes and should not place larger files in it, since it might reside in runtime memory and cannot necessarily be swapped out to disk.

[...]

PR #7380, opened 2023-01-15
Provide more user-friendly error-message when XDG_RUNTIME_DIR is missing

josteink:
This should help avoid github-issues for cases like https://github.com/swaywm/sway/issues/7202.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -319,6 +319,8 @@ int main(int argc, char **argv) {
 	if (!getenv("XDG_RUNTIME_DIR") && optind == argc) {
 		fprintf(stderr,
 				"XDG_RUNTIME_DIR is not set in the environment. Aborting.\n");
+		fprintf(stderr,
+				"Please check that user is member of seat group and that the seatd.service is enabled.\n");
 		exit(EXIT_FAILURE);
 	}
 


emersion:
This is distro-specific, I'd rather not add this.

josteink:
Ok. Would have helped users like me though. 

Could/should one instead perhaps link to a wiki page or something where one can provide more distro-specific guidance or advice?

kennylevinsen:
The message printed relates to a seatd setup step, unrelated to XDG_RUNTIME_DIR.

For seatd debugging help, improving or linking to the seatd FAQ (https://man.sr.ht/~kennylevinsen/seatd/) might be useful.

sway/main.c
```diff
@@ -319,6 +319,8 @@ int main(int argc, char **argv) {
 	if (!getenv("XDG_RUNTIME_DIR") && optind == argc) {
 		fprintf(stderr,
 				"XDG_RUNTIME_DIR is not set in the environment. Aborting.\n");
+		fprintf(stderr,
+				"Please check that user is member of seat group and that the seatd.service is enabled.\n");
```
WhyNotHugo, line 323: `seatd` **does not** create or set `XDG_RUNTIME_DIR`.

I was faced with the error in the line above when setting up my current system, and pushing me to check `seatd` would have been a distraction from the real issue.

Samueru-sama:
> This is distro-specific, I'd rather not add this.

This isn't distro specific, see here: https://github.com/swaywm/sway/issues/7202#issuecomment-2438476988

sway/main.c
xplshn: I think providing a little solution would be best, just adding this to the ~/.profile does the trick: `[ -d "$XDG_RUNTIME_DIR" ] || { mkdir -p "$XDG_RUNTIME_DIR" && chmod 0700 "$XDG_RUNTIME_DIR"; }`

xplshn:
And it is not related to logind or seatd, the var must be set to point to a directory with 0700 permissions. It just happens to be that logind sets it by default and it creates the dir. While `seatd` does not do that, and instead, it is left up to the user to create the dir and set the permissions.

So adding that in the error message or maybe even showing a possible solution such as adding `[ -d "$XDG_RUNTIME_DIR" ] || { mkdir -p "$XDG_RUNTIME_DIR" && chmod 0700 "$XDG_RUNTIME_DIR"; }` to the ~/.profile would be best.

sway/main.c
WhyNotHugo: @xplshn This solution assumes that the environment variable is set; if the directory does not exist, it is very likely that the variable is unset too.

xplshn: Yes, just gotta set the variable. Can be done in the same line. You can even use `mktemp` or do it like I do in my ~/.profile (since I use Alpine)
```
export XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR:-/tmp/.$(id -u)-runtime-dir}"
[ -d "$XDG_RUNTIME_DIR" ] || { mkdir -p "$XDG_RUNTIME_DIR" && chmod 0700 "$XDG_RUNTIME_DIR"; }

```

image

Samueru-sama:
@kennylevinsen 

Sorry for spamming here, but since you locked the issue: [context](https://github.com/swaywm/sway/issues/7202#issuecomment-2453348095)

>then why are you here?

I came here because I was discussing with @xplshn about some sandboxing software and the topic about what do to when `XDG_RUNTIME_DIR` was unset came about. to which I was told that sway crashes when the variable is not set, which I explained him it is wrong and then came here also [explained you](https://github.com/swaywm/sway/issues/7202#issuecomment-2452851197) why it is wrong.

You never gave a reason as to `"We're not going to create a runtime di"` to which I told you that's fair after all you are not forced to comply with the spec at all anyway, however a warning about it would nice and not just logging some things, **getting kicked out tty when `XDG_RUNTIME_DIR` and having to look into logs is just very user unfriendly**. So adding warning like suggested here is the [right solution](https://github.com/swaywm/sway/pull/7380#issuecomment-2438509652) if you don't want to handle the unset variable.

kennylevinsen:
I locked the issue because the original user problem was resolved more than a year ago and the discussion was off-topic, causing unwanted notifications for others. It is *also* off-topic to post on random pull-requests like this one, which is just about changing a warning and *not* about creating folders or defaulting.

You can create a specific issue, but don’t expect us to change our opinion on creating runtime directories anytime soon. There is no technical reason for us to do so as it is always possible to have the directory automatically created before sway is started - it would purely be a (in my opinion, bad) workaround for a misconfigured system, and that’s not something we have a habit of doing. Instead we prefer FAQs or guides on how to set things up.

Now, please, no more noise. I don’t want to have to moderate more than absolutely necessary.

kennylevinsen:
I'm closing this pull request for now, not because we wouldn't want improved messaging, but because the idea of providing exact instructions is tricky and highly system dependent. Is `XDG_RUNTIME_DIR` missing because you have systemd-logind misconfigured? elogind? did you try to do it in `.profile`, but is running a shell that doesn't read it? Did you want this to be set by PAM automatically?

So far, this has been deemed a seatd-specific issue, by virtue of sway requiring either seatd or (e)logind to operate with the latter managing `XDG_RUNTIME_DIR` for you, and so the FAQ for how to deal with this has been within the domain of seatd. Improving the sway wiki itself and downstream wikis like Arch if one deems something is missing would not be a bad idea at all, and such guides is probably our preferred approach over hardcoding specific suggestions into errors.

Also remember, *sway is not i3*, but an entirely different project that replicates the window management of i3. `libwayland-client` gives a hard error if it is provided a relative display path without `XDG_RUNTIME_DIR` having been set, and many other things require it too.

xplshn:
Just do not provide instructions, provide a little snippet of the XDG spec, and that is:
`XDG_RUNTIME_DIR must be set in the user's environment to point to an existing directory with 0700 permissions`

PR #7381, opened 2023-01-15
focus_on_window_activation: raise if floating

xi:
If `focus_on_window_activation = focus` is set, an xwayland window should be focused when it requests activation. If there are multiple floating windows on the screen the one that receives focus may be obscured, so it makes sense to raise it.


emersion:
I think we need to add `transaction_commit_dirty()` at the end of that function.

emersion (approved):
Thanks!

bl4ckb0ne:
There seems to be an issue with the CI, all the builds passed

PR #7384, opened 2023-01-16
Switch default menu to wmenu

emersion:
Switch the default menu to wmeny (https://sr.ht/~adnano/wmenu), a Wayland-native alternative to dmenu. This removes the dependency on Xwayland for the default config.

wmenu is small (same scope as dmenu) and has the same flags as dmenu. One downside is that it's not widely packaged by distributions yet.

We still depend on dmenu_path.

### DIFF
--- a/config.in
+++ b/config.in
@@ -18,7 +18,7 @@ set $term foot
 # Your preferred application launcher
 # Note: pass the final command to swaymsg so that the resulting window can be opened
 # on the original workspace that the command was run on.
-set $menu dmenu_path | dmenu | xargs swaymsg exec --
+set $menu dmenu_path | wmenu | xargs swaymsg exec --
 
 ### Output configuration
 #


bl4ckb0ne:
Not widely packaged indeed [1]. Maybe we can wait for more distro packaging it before merging.

[1]: https://repology.org/project/wmenu/versions

kennylevinsen:
The included config's programs are not direct dependencies and package maintainers change them occasionally, so I am not overly concerned. Users lacking the programs referenced by the included config would not be a new failure mode either...

emersion:
> maybe we can wait for the arch package (not aur) to land?

There is no guarantee that it would ever land if we don't start depending on it.

I don't think we should wait for any distro here. It's just the default config file after all. Distro which don't like it can (and already do) patch the default config file.

keithbowes:
Why not use [rofi-wayland](https://github.com/lbonn/rofi) (e.g. `set $menu rofi -show run -run-command="swaymsg exec -- {cmd}" -theme dmenu`)?  It's widely-packaged and Wayland native, and doubtlessly has more features than wmenu.

emersion:
We specifically were looking for a simple, minimal, easy to build/understand/hack/package tool.

minsoehan:
To use a package as default, just looking at the package's quality is preferable, IMHO.

PR #7386, opened 2023-01-16
Fix pointer events for ext-session-lock surfaces

emersion:
We were never sending any pointer event to ext-session-lock surfaces.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -97,6 +97,24 @@ struct sway_node *node_at_coords(
 	double ox = lx, oy = ly;
 	wlr_output_layout_output_coords(root->output_layout, wlr_output, &ox, &oy);
 
+	if (server.session_lock.locked) {
+		if (server.session_lock.lock == NULL) {
+			return NULL;
+		}
+		struct wlr_session_lock_surface_v1 *lock_surf;
+		wl_list_for_each(lock_surf, &server.session_lock.lock->surfaces, link) {
+			if (lock_surf->output != wlr_output) {
+				continue;
+			}
+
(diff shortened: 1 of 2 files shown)


sway/input/seat.c
```diff
@@ -1090,8 +1090,19 @@ void seat_configure_xcursor(struct sway_seat *seat) {
 bool seat_is_input_allowed(struct sway_seat *seat,
 		struct wlr_surface *surface) {
 	struct wl_client *client = wl_resource_get_client(surface->resource);
```
Nefsen402, line 1092: The client declaration here can be moved down to where it's used. The lock code doesn't need to use it. Otherwise, LGTM

emersion: Fixed, thanks!

sway/input/cursor.c
```diff
@@ -97,6 +97,24 @@ struct sway_node *node_at_coords(
 	double ox = lx, oy = ly;
 	wlr_output_layout_output_coords(root->output_layout, wlr_output, &ox, &oy);
 
+	if (server.session_lock.locked) {
+		if (server.session_lock.lock == NULL) {
+			return NULL;
+		}
+		struct wlr_session_lock_surface_v1 *lock_surf;
+		wl_list_for_each(lock_surf, &server.session_lock.lock->surfaces, link) {
+			double _sx = ox, _sy = oy;
```
bl4ckb0ne, line 106: Why do you copy those values? `wlr_surface_point_accepts_input` should not modify them.

emersion: Right. I've switched the code to `wlr_surface_surface_at()` to account for sub-surfaces.

bl4ckb0ne (approved):
LGTM!

PR #7392, opened 2023-01-20
Check return value of wlr_renderer_begin()

emersion:
Since [1], wlr_renderer_begin() can fail. Check its return value and bail.

This fixes an assertion error (when begin() fails and then we try to render something) after a GPU reset.

[1]: https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/a541c9510a4cf544313bc9b0503d75820b42444e


kennylevinsen (approved):
LGTM

PR #7401, opened 2023-01-25
Allow windows in scratchpad to set the urgent flag

rjarry:
When a window in the scratchpad container requests for xdg_activation_v1, it is ignored no matter what the value of focus_on_window_activation is.

At least allow windows in the scratchpad to set the urgent flag.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -369,16 +369,13 @@ void view_set_activated(struct sway_view *view, bool activated) {
 
 void view_request_activate(struct sway_view *view, struct sway_seat *seat) {
 	struct sway_workspace *ws = view->container->pending.workspace;
-	if (!ws) { // hidden scratchpad container
-		return;
-	}
 	if (!seat) {
 		seat = input_manager_current_seat();
 	}
 
 	switch (config->focus_on_window_activation) {
 	case FOWA_SMART:
-		if (workspace_is_visible(ws)) {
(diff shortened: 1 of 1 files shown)


sway/tree/view.c
```diff
@@ -368,14 +368,11 @@ void view_set_activated(struct sway_view *view, bool activated) {
 
 void view_request_activate(struct sway_view *view) {
 	struct sway_workspace *ws = view->container->pending.workspace;
-	if (!ws) { // hidden scratchpad container
-		return;
-	}
 	struct sway_seat *seat = input_manager_current_seat();
 
 	switch (config->focus_on_window_activation) {
 	case FOWA_SMART:
```
rpigott, line 374: You need to handle the FOWA_FOCUS case as well. Revealing the scratchpad container is probably the way most consistent with the ususal behavior of sway. So it will need at least something like:

```c
if (container_is_scratchpad_hidden_or_child(view->container)) {
    root_scratchpad_show(view->container);
}
```

rpigott (changes requested):
Otherwise looks fine and I think allowing hidden views to set urgent is reasonable. Needs a rebase as well.

sway/tree/view.c
rjarry: Hi there,

I have added the suggested line. Please tell me if this is correct.

PR #7408, opened 2023-01-30
[1.8] Allow setting the font size to 0

emersion:
This is https://github.com/swaywm/sway/pull/7351 but targeting the v1.8 branch.

Maybe we can merge this for 1.8.1, then decide on a better fix for 1.9?


bl4ckb0ne:
Seems like a fair deal. I'll open the issue for 1.9.

emersion:
I think we can keep https://github.com/swaywm/sway/issues/7350 to discuss our plans :)

PR #7417, opened 2023-02-01
Added reference for event types to swaymsg manpage

ManDay:
*sway-ipc*(7) is mentioned under *SEE ALSO* but it helps to mention it explicity where the events can be looked up.


emersion (approved):
Thanks!

PR #7418, opened 2023-02-01
ipc: add support for output event

emersion:
For compatibility with i3 [1].

[1]: https://i3wm.org/docs/ipc.html#_output_event


bl4ckb0ne (approved):
LGTM

PR #7420, opened 2023-02-01
root: free non_desktop_outputs list on root_destroy

t2b3:
This fixes a memory leak because the non_desktop_outputs list was not freed when the root was destroyed.


emersion (approved):
LGTM

The build failures are unrelated: https://github.com/swaywm/sway/pull/7421

t2b3:
> 

Thanks for your feedback!

PR #7424, opened 2023-02-01
root: destroy fallback_output on root_destroy

t2b3:
This fixes memory leaks.

### DIFF
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -52,6 +52,9 @@ void root_destroy(struct sway_root *root) {
 	list_free(root->scratchpad);
 	list_free(root->outputs);
 	wlr_output_layout_destroy(root->output_layout);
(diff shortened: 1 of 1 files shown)


emersion:
We should be able to clean that up with `output_destroy()`.

t2b3:
Thanks for your reply. Where should I place the `output_destroy()`? I am not sure about the right place.

t2b3:
I tried several ways, but I always end up with some issues, like `wlr_output` of `fallback_output` must be destroyed first but if I do this `output_destroy()` tries to access this freed resource. There must be a proper way to remove the `fallback_output` from `root` but I cannot find out.

t2b3:
I still need to find a good approach to solve this problem. For the meantime, I withdraw this PR.

PR #7430, opened 2023-02-04
ci: install hwdata-dev on Alpine

emersion:
We're missing the pkg-config file so the DRM backend gets disabled.


rpigott (approved):
LGTM and the ci likes it :+1: .

PR #7432, opened 2023-02-05
ipc: Fix window_rect.y with hide_edge_borders

czak:
I'm using `border pixel` with `hide_edge_borders both` and noticed that `window_rect.y` was incorrectly offset for some windows.

## Issue

`window_rect.y` is currently set to `border_thickness` for windows with `border pixel`.

This is correct if there is a top border (exaggerated here with `border pixel 10`):

image

...but it is not correct if top border is not drawn:

hide-edge-borders-both

This means that windows adjacent to top screen edge return incorrect `window_rect.y` in this configuration.

## Solution

Compute the actual `y` offset into the container for windows with `border pixel`.

I also updated the man page section about `window_rect`. Previously it said that `borders are included` but that was not correct.

## What I tested

* With `hide_edge_borders both`:
    * Window with `border pixel 10`, adjacent to top screen edge
        * Expected: `window_rect.y == 0`
    * Window with `border pixel 10`, not adjacent to top screen edge
        * Expected: `window_rect.y == 10`
    * Window with `border normal`, placed wherever
        * Expected: `window_rect.y == 0`
* With `hide_edge_borders none`:
    * Window with `border pixel 10`, adjacent to top screen edge
        * Expected: `window_rect.y == 10`
    * Window with `border pixel 10`, not adjacent to top screen edge
        * Expected: `window_rect.y == 10`
    * Window with `border normal`, placed wherever
        * Expected: `window_rect.y == 0`

### DIFF
--- a/sway/sway-ipc.7.scd
+++ b/sway/sway-ipc.7.scd
@@ -337,8 +337,9 @@ node and will have the following properties:
    this, but borders are included.
 |- window_rect
 :  object
-:  The geometry of the contents inside the node. The window decorations are
-   excluded from this calculation, but borders are included.
+:  The geometry of the content inside the node. These coordinates are relative
+   to the node itself. Window decorations and borders are outside the
+   _window_rect_
 |- deco_rect
 :  object
 :  The geometry of the decorations for the node relative to the parent node
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -577,7 +577,7 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 
 	struct wlr_box window_box = {
 		c->pending.content_x - c->pending.x,
-		(c->current.border == B_PIXEL) ? c->current.border_thickness : 0,
+		(c->current.border == B_PIXEL) ? c->pending.content_y - c->pending.y : 0,
 		c->pending.content_width,
 		c->pending.content_height
 	};


rpigott:
Hi, thanks for the patch. Looks correct but I don't agree with the man page edit. As written its still confusing because "excluded from the calculation" is kinda ambiguous. Maybe "excluded from the region" or "outside the window_rect" or something else that clearly indicates what window content lies within the reported window_rect.

sway/sway-ipc.7.scd
```diff
@@ -337,8 +337,8 @@ node and will have the following properties:
    this, but borders are included.
 |- window_rect
 :  object
-:  The geometry of the contents inside the node. The window decorations are
-   excluded from this calculation, but borders are included.
+:  The geometry of the contents inside the node. The window decorations and
```
rpigott, line 340: suggested: "The geometry of the content inside the node. Window decoractions and borders are outside the window_rect."

rpigott (changes requested):

sway/sway-ipc.7.scd
czak: Thank you for taking a look! I agree the description was not great. In [74d42ce](https://github.com/swaywm/sway/pull/7432/commits/74d42ce5d7b964960af47b89701149c8ac1ac13f) I used your suggestion - I think it is an improvement.

I also checked how i3 [describes it](https://i3wm.org/docs/ipc.html):

> window_rect (map)
>
> The coordinates of the **actual client window** inside its container. These coordinates are relative to the container and do not include the window decoration (which is actually rendered on the parent container). So, when using the `default` layout, you will have a 2 pixel border on each side, making the window_rect `{ "x": 2, "y": 0, "width": 632, "height": 366 }` (for example).

And felt that the "relative to the container" part is important, because it is specific only to `window_rect`.

Current version after [7d7d56f](https://github.com/swaywm/sway/pull/7432/commits/7d7d56f941732c15a32db3c2d5e9a11ecd4a3a10):

> The geometry of the content inside the node. These coordinates are relative to the node itself. Window decorations and borders are outside the `_window_rect_`

I'm happy to work on this further if you think we can do better.
Thanks!

rpigott:
LGTM, just needs to be rebased to squash the fixup commits. We don't need 3 commits to edit the same line of the man page.

czak:
Rebased and squashed

rpigott (approved):
Thanks!

PR #7443, opened 2023-02-08
Send wl_surface.preferred_buffer_scale

emersion:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3696

### DIFF
--- a/sway/desktop/surface.c
+++ b/sway/desktop/surface.c
@@ -56,6 +56,7 @@ void surface_update_outputs(struct wlr_surface *surface) {
 		}
 	}
 	wlr_fractional_scale_v1_notify_scale(surface, scale);
+	wlr_surface_set_preferred_buffer_scale(surface, ceil(scale));
 }
 
 void surface_enter_output(struct wlr_surface *surface,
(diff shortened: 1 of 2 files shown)


rpigott:
I take it this obviates a need for #6479? Should we consider sending this event in advance of the first map?

emersion:
> I take it this obviates a need for https://github.com/swaywm/sway/pull/6479?

Kind of but not really: we use the list of entered outputs to compute the preferred scale.

> Should we consider sending this event in advance of the first map?

Yes. That's already the case AFAIK.

rpigott:
> Yes. That's already the case AFAIK.

I don't think so. #6479 only stopped us from sending it for view_child surfaces, like popups. IIRC, this was the only case where we sent an early enter, and it was never sent in weston or mutter hence https://gitlab.freedesktop.org/wayland/wayland/-/issues/133.

emersion:
Ah, sounds like a bug then.

emersion:
Since this also affects fractional-scale, I'll merge this and track the bug in a separate issue: https://github.com/swaywm/sway/issues/7668

PR #7446, opened 2023-02-09
xdg-activation: fix urgency when the client does not specify a seat

rpigott:
xdg-activation is now too strict in only allowing tokens with a seat to activate a surface. Clients may rely on this behavior for urgency hints. The seat argument is still useful in case the client does provide a seat so we can activate it on the desired seat.


emersion (approved):
LGTM, thanks!

Issue #1005, opened 2016-12-23
Layout save/restore


SirCmpwn:
Layout save/restore is an i3 feature that serializes your layout to JSON and attempts to arrange windows in the same way in a later session. First step is to research how i3 does it and propose some strategies for implementing it in sway.

cyphar:
I did a bit of research [using the documentation as a reference](https://i3wm.org/docs/layout-saving.html). Effectively the way this is implemented (I don't use this feature at all) is this:

* When generating the `i3-save-tree` output it includes all of the obvious information about container placement and so on -- which for us would just be a matter of recursively walking down the tree.

* However, it includes "swallowing" information which is effectively a regular expression that matches `WM_CLASS` (which is a tuple of (instance, class)), `_NET_WM_NAME` (title) and `WM_WINDOW_ROLE` (window role). When a program is started in an i3 session and it matches all of the "swallowing" information, it will place that program in the loaded pane.

* Note that the output of `i3-save-tree` is not actually standard-conformant JSON. IMO we should not follow their lead, and we should always output standard compliant JSON.

So from what I can see there are a few things we need to implement and/or consider:

0. Is there a nicer way of implementing "swallowing" in Wayland than in X? The above seems quite hacky to me.
1. What is the analogy for `WM_CLASS`, `_NET_WM_NAME` and ` WM_WINDOW_ROLE` in the Wayland (and wlc) protocols?
2.

SirCmpwn:
I've decided that we are just not going to support this i3 feature, ever.

mahmoudhossam:
@SirCmpwn Any specific reason why?

SirCmpwn:
It's too complicated and hacky for too little benefit. I don't like the design constraints it imposes upon sway.

hyiltiz:
Would you reconsider your decision? This feature is actually quite useful in some (could even say rare) professional settings, e.g. setting up a music production with a midi mixer, synthesizer, sequencer, and composition programs open in their specific layout every time, or just general software development with editors, compilers, debuggers and test windows. 

While there is no point in being i3 compliant on this, I think a good discussion in trying to come up with a better solution would be quite beneficial.

SirCmpwn:
No, I won't reconsider. However, you should be able to accomplish something similar with a script that starts up the software you need and arranges it appropriately via IPC.

hyiltiz:
[...]

PR #7450, opened 2023-02-12
Feature: targeted exec

rpigott:
Currently, new windows are workspace-matched to the ws that was active at the time they were launched, with some fallbacks in place in case we can't determine that. With this patch, we use the handler context workspace instead. By default this is the same behavior, but it enables criteria to affect exec statements.

```
[workspace=3] exec foot
```

launches foot on workspace 3. Sort of.

Criteria actually match nodes, not workspaces, so if the workspace is unpopulated you get an error instead. Even worse, if the workspace _is_ populated with more than one node, you launch a new foot client for each existing match, doubling the clients each execution, yielding a fun new variety of forkbomb.

But it feels kinda slick so maybe it should be allowed. I think it's genuinely useful to match marked containers, at least, which are guaranteed to be unique.

I also think there's a good case to be made the launch context should track the handler context node instead of workspace node. All the logic is there and it would make the behavior more consistent for slow clients, e.g. right now launching a slow client before switching workspaces puts that client on the original workspace, so perhaps launching a slow client before moving between containers should also target the original container when placing the new window.  Could potentially help scripting around #1005 too.

### DIFF
--- a/include/sway/desktop/launcher.h
+++ b/include/sway/desktop/launcher.h
@@ -26,7 +26,7 @@ void launcher_ctx_consume(struct launcher_ctx *ctx);
 
 void launcher_ctx_destroy(struct launcher_ctx *ctx);
 
-struct launcher_ctx *launcher_ctx_create_internal(void);
+struct launcher_ctx *launcher_ctx_create_internal(struct sway_node *node);
 
 struct launcher_ctx *launcher_ctx_create(
 	struct wlr_xdg_activation_token_v1 *token, struct sway_node *node);
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -63,7 +63,8 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 	}
 
 	pid_t pid, child;
-	struct launcher_ctx *ctx = launcher_ctx_create_internal();
+	struct sway_workspace *ws = config->handler_context.workspace;
+	struct launcher_ctx *ctx = launcher_ctx_create_internal(&ws->node);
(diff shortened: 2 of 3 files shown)


emersion:
I think the upstream i3 way of doing this is switching the worspace before exec'ing the command? (And then optionally switching back)

rpigott:
Yep, a good way to launch a single client on a specific workspace is:

```
swaymsg workspace 3, exec foot, workspace back_and_forth
```

Doesn't work for the scratchpad unfortunately but otherwise a good method. Criteria would let you place windows on the workspace of another container, or potentially in specific containers by crtieria if we track the node instead.

GreyXor:
> 

In some cases this does not work. For example at launch, GoLand displays a banner briefly before opening the application. While the banner is displayed on the good workspace 3, the actual GoLand application will appear on the wrong one

rpigott:
That's true but there's not a lot we can do about such clients atm. Sway's internal launcher will always fallback to pid matching so even if the client tried to map 2 windows and did use startup notifications for the main window we'd miss it. It happens that with an external launcher like fuzzel this is possible but chances are high the client doesn't support any startup notifications in the first place. If we want we could recreate this behavior with the internal launcher by adding a `--no-pid-match` similar to the existing `--no-startup-id`, to better support clients trying to use startup notifications in a non-trivial way. We should do our best to support clients that are actually trying to use the protocol.

For this PR I'm going to rework it so that execs target containers instead of workspaces and see if it has any unintended consequences. I have a few ideas that could be helped by this behavior even if we want to strictly keep the default behavior as is.

PR #7456, opened 2023-02-17
Fix map_to_output * turning off device

YellowOnion:
Untested fix for #7438

### DIFF
--- a/sway/input/libinput.c
+++ b/sway/input/libinput.c
@@ -220,7 +220,8 @@ bool sway_input_configure_libinput_device(struct sway_input_device *input_device
 
 	bool changed = false;
 	if (ic->mapped_to_output &&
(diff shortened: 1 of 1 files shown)


sway/input/libinput.c
```diff
@@ -219,8 +219,9 @@ bool sway_input_configure_libinput_device(struct sway_input_device *input_device
 			ic->identifier, input_device->identifier);
 
 	bool changed = false;
-	if (ic->mapped_to_output &&
-			!output_by_name_or_id(ic->mapped_to_output)) {
+	if (strcmp("*", ic->mapped_to_output) != 0 &&
+		ic->mapped_to_output &&
```
emersion, line 223: The `ic->mapped_to_output` check must come first, to avoid NULL pointer dereference.

emersion (approved):
Thanks!

PR #7471, opened 2023-02-22
Use wlr_output_configure_primary_swapchain()

emersion:
Depends on https://github.com/swaywm/sway/pull/7470 and https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4025


Nefsen402:
This is no longer needed.

PR #7475, opened 2023-02-25
fix old style functions definitions

ElyesH:


swaynag/swaynag.c
```diff
@@ -13,7 +13,7 @@
 #include "swaynag/types.h"
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
 
-static void nop() {
+static void nop(void) {
```
emersion, line 16: This function was intentionally not using `void` to allow any number of args to be passed to it.

emersion: Dropped that change.

PR #7476, opened 2023-02-26
sway{,bar}: use default font hint style

llyyr:
CAIRO_HINT_STYLE_FULL attempts to maximize contrast at the expense of fidelity, this makes most fonts that haven't been hand hinted, which makes up the majority of fonts out there, appear much worse.

In the absence of explicitly set hint style, cairo will default to CAIRO_HINT_STYLE_SLIGHT, which attempts to improve contrast while retaining fidelity to the original shapes, which is what we want.

Current text rendering on swaybar:
image

Current text rendering on sway: 
image

After this change:
image

There are, of course, downsides to this. This only really matters at small sizes, like anything less than 10pt. Text objectively looks (only) slightly worse at higher sizes, so there's a trade off here. However, sway only renders text in container title bars which most likely doesn't want to be super huge anyway, and there will probably be not many users with a status bar that covers 5-10% of their screen height.

Slight hinting, single-color antialiasing: 
image

Slight hinting, subpixel antialiasing:
image

Full hinting, single-color antialiasing:
image

Full hinting with subpixel antialiasing is actually the same as slight hinting weirdly though, the second deleted line in swaybar/render.c doesn't change the text in any way.

I wonder if it's better to add a toggle and let the user decide instead...

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -508,7 +508,6 @@ static void render_titlebar_text_texture(struct sway_output *output,
 	cairo_t *c = cairo_create(dummy_surface);
 	cairo_set_antialias(c, CAIRO_ANTIALIAS_BEST);
 	cairo_font_options_t *fo = cairo_font_options_create();
-	cairo_font_options_set_hint_style(fo, CAIRO_HINT_STYLE_FULL);
 	if (output->wlr_output->subpixel == WL_OUTPUT_SUBPIXEL_NONE) {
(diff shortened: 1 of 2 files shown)


emersion:
This was introduced in https://github.com/swaywm/sway/pull/2677 but the PR doesn't explain why FULL was used.

LGTM

bl4ckb0ne (approved):
LGTM

PR #7477, opened 2023-02-27
Handle xwayland OR surfaces with a parent as popups

9ary:
This is the first step towards making Fusion 360 more usable. Input is not handled yet.


9ary:
I don't think this is worth finishing. The more I dig, the worse it gets, and I don't think it's reasonable to spend so much time on backwards compat.
Wine is an *awful* X11 client, the native Wayland backend is making good progress and it's the way forward. Although it's very unstable, Fusion 360 actually runs there with none of the current problems, so that's promising.
Suggested workaround in the meantime: use rootful xwayland.

PR #7478, opened 2023-02-28
Output matching cleanup

emersion:
Centralize output matching in a single function. Will make it easier to add more logic to it later, e.g. to match port names or sink tags.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -162,13 +162,10 @@ static void merge_id_on_name(struct output_config *oc) {
 	char id[128];
 	output_get_identifier(id, sizeof(id), output);
 
-	size_t size = snprintf(NULL, 0, "%s on %s", id, name) + 1;
-	char *id_on_name = malloc(size);
+	char *id_on_name = format_str("%s on %s", id, name);
 	if (!id_on_name) {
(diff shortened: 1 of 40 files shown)


sway/config/output.c
```diff
@@ -153,28 +153,22 @@ static void merge_wildcard_on_all(struct output_config *wildcard) {
...
+	if (output == NULL) {
+		return;
 	}
 
+	const char *name = output->wlr_output->name;
+	char id[128];
+	output_get_identifier(id, sizeof(id), output);
+
+	size_t size = snprintf(NULL, 0, "%s on %s", id, name) + 1;
+	char *id_on_name = malloc(size);
```
bl4ckb0ne, line 166: Wouldnt it be easier to have `id_on_name` on the stack with a big size? Like `PATH_MAX` or whatever's big enough

bl4ckb0ne: Why not, it avoids calling `snprintf` twice to get the size, `id_on_name` would always be big enough and clear of any garbage if initialized with `= {0}`.

emersion: I've added helpers and converted everything to them.

PR #7480, opened 2023-02-28
Disable direct scanout for surfaces with popups

9ary:
Extracted from #7477. I couldn't find a native wayland client that was broken by this because subsurfaces are apparently very common which masks the problem.


emersion (approved):
LGTM, thanks!

PR #7484, opened 2023-03-02
Fix stacking and visibility for Xwayland unmanaged children

9ary:
I couldn't give up so easily after all. :P With lessons learned from #7477, here's a couple small fixes that will make Fusion 360 behave a little better. There are still two outstanding problems which hurt its usability:
- Xshape support is needed
- For some reason, the client will unset OR on the surfaces when it receives clicks, which breaks input because sway moves them. Even if I hack the event handler to ignore the change, the client will unmap and remap the surfaces a lot, causing flicker. I have no idea how to fix this at the moment but ~~it doesn't seem to happen in i3 nor mutter~~ it happens in i3 too but at least the window doesn't move there.

### DIFF
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -277,7 +277,9 @@ static void set_activated(struct sway_view *view, bool activated) {
 	}
 
 	wlr_xwayland_surface_activate(surface, activated);
-	wlr_xwayland_surface_restack(surface, NULL, XCB_STACK_MODE_ABOVE);
+	if (activated) {
+		wlr_xwayland_surface_restack(surface, NULL, XCB_STACK_MODE_ABOVE);
+	}
 
 	struct sway_xwayland_unmanaged *unmanaged;
 	wl_list_for_each(unmanaged, &root->xwayland_unmanaged, link) {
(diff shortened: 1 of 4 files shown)


sway/desktop/xwayland.c
```diff
@@ -274,7 +296,21 @@ static void set_activated(struct sway_view *view, bool activated) {
...
+
+	struct sway_xwayland_unmanaged *unmanaged;
+	wl_list_for_each(unmanaged, &root->xwayland_unmanaged, link) {
+		struct wlr_xwayland_surface *unmanaged_surface = unmanaged->wlr_xwayland_surface;
+		if (!is_surface_transient_for(unmanaged_surface, view)) {
+			continue;
+		}
+		unmanaged->visible = activated;
+		if (activated) {
+			wlr_xwayland_surface_restack(unmanaged_surface, NULL, XCB_STACK_MODE_ABOVE);
```
9ary, line 311: Alternatively we could ensure that no window ever gets restacked above anything that's override_redirect...

emersion:
Overall looks good.

Instead of introducing a new `visible` field I'd prefer to have a getter function instead, to avoid introducing a new source of truth.

9ary:
That can be fixed but I'd still question the correctness of messing with override_redirect windows when the point is that clients are supposed to do the right thing themselves. This only seems to affect wine, and it can be fixed there, so I'd favor doing that. This can always be picked up again if a client comes up that can't be fixed.

PR #7496, opened 2023-03-07
Add a .mailmap file

rpigott:
See `gitmailmap(5)`.

This is a way to keep contact info up to date in git blame/log etc. I'm not sure if anyone else has contributed under two emails but they of course could be added. There are also some old commits from @emersion using a different git user which could also be corrected with a mailmap line if desired.


emersion (approved):
Thanks!

Issue #5553, opened 2020-07-16
create_output is undocumented


RPigott:
Headlining the new features for 1.5 is the new `create_output` command, but it doesn't appear in sway(5).

Steps to reproduce:
```
$ swaymsg -t get_version
sway version v1.5-e215557b (Jul 16 2020, branch 'master')
$ MANPAGER=less LESS=+/create_output man 5 sway
Pattern not found
```
:(

RPigott:
Also could there maybe be a "docs" issue label? Wasn't sure where to file this.

emersion:
Ooops. It was originally a developer-only feature, so intentionally undocumented. But now that it can be useful to end-users, it should definitely be documented.

Zocker1999NET:
Is there somewhere a developer-doc already existing? It could be used as template for the end-user doc and maybe useful for all finding this issue.

For references:
- Introducing: #2669
- Implementation of backend: #5216
- Implementation of command: #5215

progandy:
I wonder if it should remain undocumented and the command merged with output, like

output "HL_MY_NAME" create (any name prefixed with HL that does not exist)
output "HL_MY_NAME" destroy
(output "*" create to create a sequential HEADLESS# name)

thecoder-001:
Just posting it here for anyone else who too wants to create a virtual display without looking at the commits to figure things out and has stumbled here.
Someone at reddit posted their setup [here](https://www.reddit.com/r/swaywm/comments/k1zl41/thank_you_devs_free_ipad_repurposed_as_a_second/) along with  an example config.

pradyparanjpe:
## I asked ChatGPT
Somehow, ChatGPT knows the syntax.

### DISCLAIMER:
I haven't checked if each of this works though (ChatGPT may have mis-learnt or may be writing a novel).

These are some extracts from its answer:

```sh
swaymsg create_output <output_name> [enable=<true/false>] [mode=<mode>] [position=<x> <y>] [scale=<scale>] [transform=<transform>] [background=<color/image>] [border=<normal/hide]] [layout=<layout>] [workspace=<workspace>]
```

```md
- <output_name>: The name you want to assign to the new output.
- [enable=<true/false>]: Whether to enable the output immediately upon creation. Default is true.
- [mode=<mode>]: The display mode (resolution and refresh rate) of the output.
- [position=<x> <y>]: The position of the output relative to the top-left corner of the screen.
- [scale=<scale>]: The scaling factor for the output.
[...]

PR #7497, opened 2023-03-08
Document create_output and unplug in man pages

sooraj-satheesh:
Documents the commands for using virtual inputs in the manual pages.

### DIFF
--- a/sway/sway-output.5.scd
+++ b/sway/sway-output.5.scd
@@ -131,6 +131,11 @@ must be separated by one space. For example:
 	As opposed to the _enable_ and _disable_ commands, the output keeps its
 	current workspaces and windows.
 
+*output* <name> unplug
+	Removes the specified headless output.
+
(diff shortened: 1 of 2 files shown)


emersion:
This is intentional, see https://github.com/swaywm/sway/issues/5553

CheerfulPianissimo:
Ah, sorry, did not see that issue. This [comment](https://github.com/swaywm/sway/issues/5553#issuecomment-659253656) of yours seems to indicate otherwise though?

emersion:
That's true.

I do wonder if we can find a better name than `create_output`. We could go for `output create` but that would be a bit inconsistent with the way it works right now.

omeid:
@emersion Is it possible to let the user select the display name so that something like `output ONTHEWALL create` could work?

emersion:
I'd rather not. The name picked by the user may clash with real output names (current or future), which would be tricky to handle.

Issue #5308, opened 2020-05-07
Scratchpad window size is different on several outputs


chep:
When moving a window to scratchpad, it is resized. When it is shown on another output, it is resized again, even if the resolution is the same. 
Then it seems to keep the new size on all monitors.

If you resize it, the size is kept until you display it on another output.

The size seems to be different regarding the application.

sway version: 1.4

Config:
output DVI-I-1 pos 0,0 res 1920x1080
output HDMI-A-1 pos 1920,0 res 1920x1080

c-cube:
I imagine it could be because of [this](https://github.com/swaywm/sway/blob/5501d14a2e7147b533df12b21368639cf01a7c9c/sway/tree/root.c#L150) where it probably resizes too eagerly?

chep:
It does not happen with every windows. For example it happens with kitty but not with terminator.

josefbachmann:
I have the same problem

sway version 1.5.1

### Description:
In my case: [termite](https://github.com/thestinger/termite)
On startup I open a terminal in my scratchpad like:

```
set $term termite
exec $term --name $scratchterm
for_window [app_id="(?i)$scratchterm"] move to scratchpad, resize set 1800 1000 px, opacity set 0.9;
```

I have two monitors:

```
set $display1 DP-1
set $display2 HDMI-A-1
output $display1 pos 0 0 res 2560x1440
output $display2 pos 2560 0 res 1920x1080
output $display1 mode 2560x1440@143.998001Hz
output $display2 mode 1920x1080@60.000000Hz
```

When open scratchpad (`bindsym $mod+Return scratchpad show`) on my main display1 everything is correct set up (size: 1800x1000px). If I close and reopen scratchpad on the second screen the size has completely changed and is much smaller for the scratchpad terminal window. it doesn't matter on which display I create the scratchpad window, when I open it on the other screen the size of the scratchpad changes.

is my config wrong or is this a bug?
thank you.

gilbertw1:
@josefbachmann I've encountered this exact same issue independently (using alacritty), so I don't believe it's an issue with your configuration. It even happens if you use the default keybindings to place a floating terminal in the scratchpad and recall it on a different monitor.

I have the same use case as you for using the scratchpad and had to end up writing a `run-or-raise` style script that manually enforces the dimension anytime the terminal is recalled from the scratchpad.

josefbachmann:
[...]

Issue #6561, opened 2021-09-29
Scratchpad doesn't remember window size when moving to another monitor


Zeioth:
**How to reproduce**:

* Send a window to the scratchpad.
* Move the cursor to another monitor and bring back the window from scratchpad.
* The size of the window changed, and is now super small.

This issue only seem to happen in multi monitor setups. If you always work in the same monitor, it does't happen.

boredland:
What are the screen resolutions of the different monitors? Are they identical or different?

Zeioth:
5 monitors, all of them UHD (3840 x 2160).

My config, if relevant:
```
output $screen2 mode 3840x2160@70.981Hz pos 0 0 scale 1
output $screen3 mode 3840x2160@70.981Hz pos 3840 0 scale 1
output $screen4 mode 3840x2160@70.981Hz pos 0 -2160 scale 1
output $screen5 mode 3840x2160@70.981Hz pos -3840 -2160 scale 1
output $screen6 mode 3840x2160@70.981Hz pos 3840 -2160 scale 1
```

boredland:
So this is different from the problem I experienced in the past, I think, when switching from a high-res display to a lower-res one.

Shfty:
I'm also seeing this on my 3-monitor setup - a central 2560x1440 monitor flanked by a pair of 1920x1200 ones.

In my case, restoring a scratchpad terminal on a different monitor causes it to briefly appear at its original size (i.e. the size it was hidden at), then shrink down to 69 columns x 24 rows.

I note that the window doesn't move during this process - when appearing at the original size, it's center-aligned as if it's already 69x24, after which the resize occurs.

paulodiovani:
I handle this issue by resizing the window after showing scratchpad. E.g.

```i3config
bindsym Ctrl+Escape exec swaymsg 'scratchpad show' \
  && sleep .1 \
  && swaymsg 'resize set 100ppt 40ppt, move position 0 0'
```

The `sleep .1` is necessary due to the time the window takes to render, otherwise, it could try resizing the underlying window, if any.
It is not pretty, as the window will flick before resizing/moving, due to the latency of using `exec swaymsg` instead of calling from sway process. I supposed it can be made better with use of `sway-ipc`, but it solves my issue for now.

I currently use this for my _dropdown terminal_.
Full config: https://github.com/paulodiovani/dot-files/blob/feature/sway/home/user/.config/sway/config.d/50-dropdownterminal

anpandey:
[...]

PR #7504, opened 2023-03-12
root: Try to preserve relative positions of floating containers

anpandey:
Should fix https://github.com/swaywm/sway/issues/5308 and https://github.com/swaywm/sway/issues/6561. The translation behavior also applies to all floating containers (not just those on the scratchpad), since that's what i3 does.

I've been dogfooding this for the past week on a setup with monitors of different aspect ratios without issue.

`floating_fix_coordinates` is adapted from [the same in i3](https://github.com/i3/i3/blob/d26ddcbfe5f0ead61cc55c1f596692d9a89e71f9/src/floating.c#L799).

### DIFF
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -712,6 +712,21 @@ void floating_calculate_constraints(int *min_width, int *max_width,
 
 }
 
+void floating_fix_coordinates(struct sway_container *con, struct wlr_box *old, struct wlr_box *new) {
+	if (!old->width || !old->height) {
+		// Fall back to centering on the workspace.
+		container_floating_move_to_center(con);
+	} else {
+		int rel_x = con->pending.x - old->x + (con->pending.width / 2);
+		int rel_y = con->pending.y - old->y + (con->pending.height / 2);
+
+		con->pending.x = new->x + (double)(rel_x * new->width) / old->width - (con->pending.width / 2);
+		con->pending.y = new->y + (double)(rel_y * new->height) / old->height - (con->pending.height / 2);
+
+		sway_log(SWAY_DEBUG, "Transformed container %p to coords (%f, %f)", con, con->pending.x, con->pending.y);
+	}
+}
+
 static void floating_natural_resize(struct sway_container *con) {
 	int min_width, max_width, min_height, max_height;
 	floating_calculate_constraints(&min_width, &max_width,
@@ -1025,6 +1040,13 @@ void container_floating_move_to(struct sway_container *con,
 		workspace_add_floating(new_workspace, con);
 		arrange_workspace(old_workspace);
 		arrange_workspace(new_workspace);
+		// If the moved container was a visible scratchpad container, then
+		// update its transform.
+		if (con->scratchpad) {
+			struct wlr_box output_box;
+			output_get_box(new_output, &output_box);
+			con->transform = output_box;
+		}
 		workspace_detect_urgent(old_workspace);
 		workspace_detect_urgent(new_workspace);
 	}
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -56,6 +56,16 @@ void root_destroy(struct sway_root *root) {
 	free(root);
 }
 
+static void set_container_transform(struct sway_workspace *ws,
+			struct sway_container *con) {
+	struct sway_output *output = ws->output;
+	struct wlr_box box = {0};
+	if (output) {
+		output_get_box(output, &box);
+	}
+	con->transform = box;
+}
+
 void root_scratchpad_add_container(struct sway_container *con, struct sway_workspace *ws) {
 	if (!sway_assert(!con->scratchpad, "Container is already in scratchpad")) {
 		return;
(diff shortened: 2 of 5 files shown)


rpigott (commented):
This causes my scratchpad windows to slowly float away up and left from their original positions. I reproduce like so:

1. open foot, then another floating foot and align it the top right corner with the top right of the screen
2. send floating foot to the scratchpad
3. show foot from the scratchpad alternately on the populated workspace and then on an empty workspace, same output, same workspace geometry
4. each round shifts the window up and to the left

possibly related to output scale 1.5. I can look into it a little more later.

Otherwise the code looks alright.

anpandey:
I can reproduce it without scaling, if the ratios of the heights and the widths of the two outputs are weird. Here's the log where I tried the same thing:

```
00:07:33.077 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (576.000000, 0.500000)
00:07:35.391 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1818.300000, 79.428058)
00:07:36.661 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (575.687296, -0.215078)
00:07:37.337 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1817.340625, 78.130216)
00:07:37.618 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (574.644951, -0.985588)
00:07:39.386 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1816.381250, 76.832374)
00:07:40.419 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (573.602606, -1.756098)
00:07:41.601 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1815.421875, 75.534532)
00:07:42.380 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (572.560261, -2.526608)
00:07:43.230 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1814.462500, 74.236691)
00:07:43.717 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (571.517915, -4.067627)
00:07:44.124 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1813.503125, 72.938849)
00:07:44.516 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (570.475570, -4.838137)
00:07:46.381 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (570.000000, -5.500000)
00:07:47.096 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1812.543750, 71.641007)
00:07:47.688 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (569.433225, -5.608647)
00:07:48.182 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1811.584375, 70.343165)
00:07:48.668 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (568.390879, -7.149667)
00:07:49.226 [DEBUG] [sway/tree/container.c:722] Transformed container 0x1c64b30 to coords (1810.625000, 69.045324)
```

rpigott:
I'm only using one monitor though. Maybe it's related to smart_gaps then? I don't really mind the behavior when switching between different shaped monitors, there probably isn't a perfect solution for that.

anpandey:
I was able to reproduce it using smart_gaps. This was happening because with smart_gaps, a workspace with no windows  still has gaps enabled (so it ends up being smaller than a workspace with a single container).

I've changed the scratchpad-related transforms so that they use the output box instead of the workspace box. The transforms should be stable as long as we're restricted to a single output.

sway/tree/container.c
```diff
@@ -712,6 +712,16 @@ void floating_calculate_constraints(int *min_width, int *max_width,
 
 }
 
+void floating_fix_coordinates(struct sway_container *con, struct wlr_box *old, struct wlr_box *new) {
```
rpigott, line 715: If the old box is zeros, e.g. because the transform is unset or taken from fallback output geometry, we should handle it such that the client gets centered like a new window, as it did before.

sway/tree/root.c
```diff
@@ -56,6 +56,17 @@ void root_destroy(struct sway_root *root) {
 	free(root);
 }
 
+static void set_container_transform(struct sway_workspace *ws,
+			struct sway_container *con) {
+	struct sway_output *output = ws->output;
+	if (!output) {
+		output = root->fallback_output;
```
rpigott, line 63: Maybe just explicitly set the transform to zeros. I'm not sure the fallback output is sensible here.

rpigott (changes requested):
Great, the new version works better.

rpigott (commented):
LGTM, just needs to be squashed.

anpandey:
Done.

rpigott (approved):
Thanks!

PR #7512, opened 2023-03-16
Set withdrawn state for xwayland views on hidden workspaces

9ary:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4057

The current implementation is a huge hack and some filtering may be needed for eg sticky views.

Mutter does this and I'm trying to get Wine to recognize withdrawn state as minimized. https://gitlab.winehq.org/wine/wine/-/merge_requests/2343

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -212,12 +212,25 @@ static void transaction_add_node(struct sway_transaction *transaction,
 	}
 }
 
+void container_iterator(struct sway_container *con, void *data) {
+	bool *is_current = data;
+	if (con->view) {
+		view_set_withdrawn(con->view, !*is_current);
+	}
+}
+
+void workspace_iterator(struct sway_workspace *ws, void *data) {
+	bool is_current = workspace_is_visible(ws);
+	workspace_for_each_container(ws, container_iterator, &is_current);
+}
+
 static void apply_output_state(struct sway_output *output,
 		struct sway_output_state *state) {
 	output_damage_whole(output);
 	list_free(output->current.workspaces);
 	memcpy(&output->current, state, sizeof(struct sway_output_state));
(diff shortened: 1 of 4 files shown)


sway/desktop/transaction.c
```diff
@@ -212,12 +212,25 @@ static void transaction_add_node(struct sway_transaction *transaction,
...
+	workspace_for_each_container(ws, container_iterator, &is_current);
+}
+
 static void apply_output_state(struct sway_output *output,
 		struct sway_output_state *state) {
 	output_damage_whole(output);
 	list_free(output->current.workspaces);
 	memcpy(&output->current, state, sizeof(struct sway_output_state));
 	output_damage_whole(output);
+	output_for_each_workspace(output, workspace_iterator, NULL);
```
kennylevinsen, line 233: Wouldn't it logically make more sense to do this on `apply_workspace_state` when a workspace node has been dirtied, or is there a reason for looking at the output node specifically?

When using `wlr_scene` (soon, hopefully), this would likely be implemented by tying into the scene node output visibility logic.

9ary: I'm not very familiar with the code here, but I noticed that moving workspaces between outputs or switching to a different workspace would dirty the output.

When moving a workspace across outputs, two workspaces change visibility, and they aren't the workspace that's being moved. Do those get marked as dirty?

Otherwise, this is pretty much the first thing that I tried and it happened to work in the following cases:
- moving a workspace across outputs
- switching workspaces on the same output (via the workspace command, or by focusing a container on another workspace)
- moving containers across workspaces

If `apply_workspace_state` catches those cases, then that works. Otherwise this already seems to catch everything, so I didn't bother hooking workspace state changes.

PR #7514, opened 2023-03-17
man: deprecate seat cursor move/set/press/release

emersion:
The Wayland protocol better serves this purpose, and is supported by more compositors.

### DIFF
--- a/sway/sway-input.5.scd
+++ b/sway/sway-input.5.scd
@@ -228,6 +228,8 @@ correct seat.
 	absolute coordinates (with respect to the global coordinate space).
 	Specifying either value as 0 will not update that coordinate.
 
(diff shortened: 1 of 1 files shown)


rpigott:
I don't think it matters, but virtual pointer can't really replace `cursor set`, can it?

@whot's libei stuff seems like an improvement over virtual pointer still if we could support it, but I never see any discussion about it. Is it dead?

emersion:
> virtual pointer can't really replace cursor set, can it?

The absolute motion events can.

I don't really like libei. I'd rather use the Wayland protocols we already have.

rpigott (approved):
Well deprecating this seems like a good idea either way.

PR #7517, opened 2023-03-18
root: Set inactive focus when scratchpad is moved to new workspace

anpandey:
Fixes behavior which can be replicated by these steps:

1. In a session with two outputs, open at least 2  tiling windows on each visible workspace and put a window in the scratchpad
2. While on output 1, show the scratchpad with `scratchpad show`
3. Move to output 2
4. Move the visible scratchpad window to the focused output with `scratchpad show`
5. Move back to output 1 with `focus output`

All of the containers in output 1 end up being focused instead of just the inactive one.


rpigott (approved):
LGTM, thanks!

PR #7524, opened 2023-03-26
Init the damage_ring bounds on output creation

ErikReider:
Otherwise the initial bounds would be `INT_MAX` until `handle_mode` or `handle_commit` is called :)


emersion:
Thanks!

ErikReider:
Seems like this is still an issue when unplugging and plugging the monitor back in.

@emersion Setting the damage ring bounds inside of the `sway_output` loop in `update_output_manager_config` fixes this but I'm not sure if it's a good fix due to the function being called from other places as well.

Sorry for the ping :)

PR #7527, opened 2023-03-27
swaybar: Set opaque region properly

Nefsen402:
The opaque region is set incorrectly if updated on-the-fly if switching from an opaque to a non opaque background.


kennylevinsen (approved):
LGTM

Issue #7528, opened 2023-03-27
Commit ac1ed63 - broken clipping on external monitor


quoing:
It seems that external monitor is clipped to default mode dimensions.

```
~ swaymsg -t get_version
sway version 1.9-dev-dee032d0 (Mar 27 2023, branch 'master')
```

```
~ swaymsg -t get_outputs
Output eDP-1 'AU Optronics 0x403D Unknown' (focused)
  Current mode: 1920x1080 @ 60.049 Hz
  Power: on
  Position: 0,0
  Scale factor: 1.000000
  Scale filter: nearest
  Subpixel hinting: unknown
  Transform: normal
  Workspace: 1
  Max render time: off
  Adaptive sync: disabled
  Available modes:
    1920x1080 @ 60.049 Hz

Output DP-3 'Ancor Communications Inc ROG PG279Q J8LMQS041702'
  Current mode: 2560x1440 @ 59.951 Hz
  Power: on
  Position: 1920,0
  Scale factor: 1.000000
  Scale filter: nearest
  Subpixel hinting: unknown
  Transform: normal
  Workspace: 2
  Max render time: off
  Adaptive sync: disabled
  Available modes:
    2560x1440 @ 59.951 Hz
    1920x1080 @ 50.000 Hz (16:9)
    1280x720 @ 60.000 Hz (16:9)
    1280x720 @ 59.940 Hz (16:9)
    1280x720 @ 50.000 Hz (16:9)
    1024x768 @ 60.004 Hz
    800x600 @ 60.317 Hz
    720x576 @ 50.000 Hz (16:9)
    720x480 @ 60.000 Hz (16:9)
    720x480 @ 59.940 Hz (16:9)
    640x480 @ 60.000 Hz (4:3)
    640x480 @ 59.940 Hz
    640x480 @ 59.940 Hz (4:3)
```
You can see Chrome cut left and bottom side on screenshot..
![broken-dp-3-full2](https://user-images.githubusercontent.com/13317708/227942591-7dc98bab-2404-438e-8a99-e94a03b8dbf0.png)

Replication: Update sway to post-ac1ed63, start sway,

quoing:
Not sure if it is interesting, but external monitor is plugged via docking station
https://support.lenovo.com/us/en/solutions/acc500106-thinkpad-usb-c-dock-gen-2-overview-and-service-parts

quoing:
[...]

PR #7530, opened 2023-03-30
Fix damage-ring bounds not being set when unplugging -> plugging in monitor

ErikReider:
#7524 was a partial fix. Seems like this is still an issue when unplugging and plugging the monitor back in.

I'm not sure if it's a good fix due to the `update_output_manager_config` function being called from other places though.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -825,7 +825,6 @@ static void update_output_manager_config(struct sway_server *server) {
 			config_head->state.x = output_box.x;
 			config_head->state.y = output_box.y;
 		}
-	
 	}
 
 	wlr_output_manager_v1_set_configuration(server->output_manager_v1, config);
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -825,6 +825,10 @@ static void update_output_manager_config(struct sway_server *server) {
 			config_head->state.x = output_box.x;
 			config_head->state.y = output_box.y;
 		}
+	
+		int width, height;
+		wlr_output_transformed_resolution(output->wlr_output, &width, &height);
+		wlr_damage_ring_set_bounds(&output->damage_ring, width, height);
 	}
 
 	wlr_output_manager_v1_set_configuration(server->output_manager_v1, config);
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -826,9 +826,6 @@ static void update_output_manager_config(struct sway_server *server) {
 			config_head->state.y = output_box.y;
 		}
 	
-		int width, height;
-		wlr_output_transformed_resolution(output->wlr_output, &width, &height);
-		wlr_damage_ring_set_bounds(&output->damage_ring, width, height);
 	}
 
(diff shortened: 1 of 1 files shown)


emersion:
Does it fix the bug to add `| WLR_OUTPUT_STATE_ENABLED` to this line instead?

https://github.com/swaywm/sway/blob/9cf66e8c7cfc06680a1af055bafd7e5cc01d6fe2/sway/desktop/output.c#L931

ErikReider:
> Does it fix the bug to add `| WLR_OUTPUT_STATE_ENABLED` to this line instead?

It sadly doesn't

emersion:
Well, I don't think `update_output_manager_config()` is a good place to do this. One would need to figure out which of the call sites fixes the issue.

quoing:
Could we rollback the original commit that actually broke this thing - or did it fix something else for majority of users? At this point I have to ignore any updates for sway-git or build sway manually (while rolling back the problematic commit).. I'm of course open to check fixes in separate devel branch in case something needs to be tested :)

ErikReider:
It's this call that's the caller.
https://github.com/swaywm/sway/blob/9cf66e8c7cfc06680a1af055bafd7e5cc01d6fe2/sway/desktop/output.c#L1030

Fixed in the latest commit

sway/desktop/output.c
```diff
@@ -1027,6 +1028,15 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
 
+	struct sway_output *tmp_output;
+	wl_list_for_each(tmp_output, &root->all_outputs, link) {
+		if (tmp_output == root->fallback_output) {
+			continue;
+		}
+		int width, height;
+		wlr_output_transformed_resolution(tmp_output->wlr_output, &width, &height);
+		wlr_damage_ring_set_bounds(&tmp_output->damage_ring, width, height);
+	}
```
emersion, line 1039: Hm. Do we really need to loop over all outputs?

Is it sufficient to move the `wlr_damage_ring_set_bounds()` call above, to a place after `apply_output_config()`?

ErikReider: I'll double check 👍

ErikReider: You're right! I can now also move the initial logic from the first PR down here :)

sway/desktop/output.c
```diff
@@ -825,6 +825,7 @@ static void update_output_manager_config(struct sway_server *server) {
 			config_head->state.x = output_box.x;
 			config_head->state.y = output_box.y;
 		}
+
```
emersion, line 828: Nit: stray newline

emersion (approved):
LGTM!

quoing:
tested the commit and it fixes my issue https://github.com/swaywm/sway/issues/7528 .. 

```
~ swaymsg -t get_version
sway version 1.9-dev-fa4aa79c (Apr  6 2023, branch 'makepkg')
```

emersion:
Thanks!

PR #7538, opened 2023-04-07
improve inactive-windows-transparency.py script

buyolitsez:
It's not convenient sometimes when you want to see two windows and both not transparent, so I've added a feature to script that you can send a USR1 signal and script won't make current window transparent(send USR1 again to remove that lock)

so I've just added `bindsym` to 
```
exec ps aux | grep 'inactive-windows-transparency.py ' | head -n 1 | awk '{print $2}' | xargs kill -s USR1
```
and now can make/unmake focused window with a simple keybind

### DIFF
--- a/contrib/inactive-windows-transparency.py
+++ b/contrib/inactive-windows-transparency.py
@@ -11,6 +11,23 @@ import signal
 import sys
 from functools import partial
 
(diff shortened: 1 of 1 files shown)


emersion:

Issue #5941, opened 2021-01-09
Support floating_modifier on touch


danieldg:
Currently, `floating_modifier` only handle tablet-tooltip and click events to start a window move.  It would be nice if it also worked on touch events so you can also move windows by touching.

I'm not sure what i3 does here, but it might treat a touch like a click - if so, this would also be an i3-compat request.

I would also like to be able to set Shift+Mod+touch to act like a right click for resize, but that would require new configuration (probably via an additional argument to floating_modifier).

frogamic:
I also would like this feature

PJungkamp:
I use sway on a surface pro which I occasionally take with me as a tablet.

Resizing windows by the borders is often really finicky, so I'd like to just be able to resize a window by touch when I'm in resize mode, since I do not have a keyboard with me to hold a modifier on.

Alternatively I'd like to be able to make any key the floating_modifier, to be able to map VolumeUp or VolumeDown as the floating modifier, since these are the only available keys when no keyboard is connected.

I commented on this issue, since here somebody already proposed augmenting the floating_modifier option for the sake of touch friendliness.

emersion:
Pull requests are welcome.

PJungkamp:
I think I don't need to touch the drag functionality itself.
 
For using any key combination as a floating mod, i need to parse floating_mod as modifiermask and/or key. 
Futhermore it seems like I need to modify the handle_button function initiating the drag and add a floating_mod_key to the config.

For implementing something like a drag_mode, I'd like to add a drag_mode command with enable/disable/toggle and an --inverse flag, where every mouse button is directly used for tiling without the need for a modifier.

I need to watch out for not conflicting with the tiling_drag config option.

Does this sound reasonable? Did I miss any potentially conflicting command?
(I've used sway for 1,5 months and linux as my OS for about 5 months, this is going to be fun!)

emersion:
Why can't we just handle touch just like we handle tablet tooltips?

tricepsbrachii:
I think this can be accomplished with using [keyd](https://github.com/rvaiya/keyd) to intercept input events of your choosing (eg tablet button or screen tap) and re-cast them to an unused key/code and map that to your floating_modifier.

danieldg:
> I think this can be accomplished with using [keyd](https://github.com/rvaiya/keyd) to intercept input events of your choosing (eg tablet button or screen tap) and re-cast them to an unused key/code and map that to your floating_modifier.

[...]

PR #7544, opened 2023-04-16
Tablet and Touchscreen Improvements

hmpthcs:
Have been loving Sway on my wacom tablet, but from day one of using it, I've been dreaming of a couple of what I consider essential features to make it a fully functional system. In this pull request, I would like to discuss and collect various minor changes that expose some of Sway's basic features to those working without a keyboard. These are not meant to be fundamental changes or additions of new features, but rather ways to extend existing features for use in the ever-growing world of touchscreen and tablet devices. Look forward to seeing some comments and suggestions.

First and foremost, window manipulation is basically impossible without being able to hold down the floating_modifier using a keyboard or, as I have done thus far, using a virtual keyboard to spoof the mod key..The latter is no solution at all and seems to overload sway at times.

I copied the changes from  #6360  to the latest master branch and have been testing on my tablet. So far this is a partial fix for #5941 (no touch yet, just stylus). 

**How to use:** 
1) "drag_mode enable" must be set in the config, or alternatively sent with "swaymsg drag_mode enable" at runtime. 
2) with the above set, the existing command "tiling_drag disable", causes the wacom stylus to act as a normal mouse; with "tiling_drag enable", the stylus resizes/moves windows in the same way as mouse+modifier normally do (i.e., drag anywhere on the window without needing to grab a titlebar/border/gap).
3) whether the stylus drags or resizes the window can be changed using the existing command, "floating_modifier <modifier> [normal|inverse]". As expected for a mouse-left-click, "normal" causes the stylus to move windows' positions, while "inverse" causes with stylus to resize windows.

**Limitations:**
1) Touchscreen input should do the same as the stylus does. In the attached code, only stylus events are working. Does #6455 provide any insight? I'm still trying to familiarize myself on on handling of pointers/inputs/etc. in sway.
2) using the "tiling_drag [enable|disable]" to toggle behavior may interfere with normal dragging of windows by titlebars (I don't have any titlebars, so haven't really tested this yet). In any case, it seems potentially confusing to have a chain of config settings that must be specified for stylus/touch drag to be enabled. If a user is ok with having  "tiling_drag enable" permanently set, this is no problem; otherwise two commands must be sent with swaymsg. There should be a single config setting to turn it on and off. Two approaches to fix this: a) remove the userspace "drag_mode [enable|disable]" entirely and instead have this command be effected automatically with "tiling_drag [enable|disable]" ; b) having "drag_mode enable" automatically send "tiling_drag enable". This second option would probably require a config error be sent if drag mode is on but tiling drag has been disabled. 
3) I don't think there is currently any way to get the current state of "drag_mode" via swaymsg. This would be necessary for integration with status bar indicators, notification apps, etc. Can we make this a "binding mode"? I haven't actually used any of these yet in my status bar or IPC scripts, but isn't this how swaybar and other shell apps get the info?


cccslater:
This looks like exactly what I'm looking for, shame it didn't get merged :-(

PR #7552, opened 2023-04-27
Use wlr_render_pass

Nefsen402:
Ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3631

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -32,6 +32,7 @@
 #endif
 
 struct render_data {
+	struct render_context *ctx;
 	const pixman_region32_t *damage;
 	float alpha;
 	struct wlr_box *clip_box;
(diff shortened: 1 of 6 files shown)


emersion:
Hm, it seems like with a 90-degree output rotation swaybar renders clipped somehow… Doesn't seem to be a damage issue…

sway/desktop/render.c
```diff
@@ -1020,32 +1019,39 @@ void output_render(struct render_context *ctx) {
...
 
 	if (debug.damage == DAMAGE_HIGHLIGHT) {
-		wlr_renderer_clear(renderer, (float[]){1, 1, 0, 1});
+		wlr_render_pass_add_rect(ctx->pass, &(struct wlr_render_rect_options){
+			.box = { .width = output->width, .height = output->height },
+			.color = { .r = 1, .g = 1, .b = 0, .a = 1 },
+		});
 	}
 
+	pixman_region32_t background;
```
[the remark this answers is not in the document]
Nefsen402: Yeah...

sway/desktop/render.c
```diff
@@ -103,18 +98,17 @@ static void set_scale_filter(struct wlr_output *wlr_output,
...
-	enum wl_output_transform inv_transform = wlr_output_transform_invert(transform);
-	wlr_matrix_project_box(matrix, &proj_box, inv_transform, 0,
-		output->wlr_output->transform_matrix);
+	struct wlr_fbox src_box = {0};
+	if (_src_box) {
+		src_box = *_src_box;
+	}
 
 	pixman_region32_t damage;
 	pixman_region32_init(&damage);
```
emersion, line 114: Maybe we need to use the scaled box here?

Nefsen402: Turns out the dst_box shouldn't be scaled to fix the issue. Callers are already responsible for scaling which seems odd to me. I'll leave it to try to reduce risk of regressions till all this gets blown away by scene.

Nefsen402:
> Hm, it seems like with a 90-degree output rotation and scale=1.5 swaybar renders clipped somehow… Doesn't seem to be a damage issue…

I think I worked through all the transform issues combined with scale

emersion:
It seems like the cursor is getting chopped off still when hovering swaybar with an output set to scale=1.5 transform=90.

emersion:
Ah, this is reproducible with scene, so unrelated to this patch!

Nefsen402:
The cursor??!! The rendering there isn't done by us. It's done by our fabulous friend `wlr_cursor`. I wonder if there is a bug in wlr_cursor_to_render_pass

emersion (approved):
LGTM, thanks!

PR #7561, opened 2023-05-02
Don't crash if there is no damage during render

Nefsen402:


sway/desktop/render.c
```diff
@@ -1018,7 +1018,7 @@ void output_render(struct render_context *ctx) {
 
 	if (!pixman_region32_not_empty(damage)) {
 		// Output isn't damaged but needs buffer swap
-		goto renderer_end;
```
emersion, line 1021: Hm, I don't think we even need to render software cursors in that case? So maybe we can just `return`?

Nefsen402: Software cursors submit damage on their own... Oh, but that should just go into the damage ring. I think a return here should be fine.

emersion (approved):
Thanks!

Issue #4090, opened 2019-04-25
Unable to set background


JonnyMako:
With latest sources, I'm not able to set background anymore. I see this in the log:
```
2019-04-25 11:22:19 - [sway/config/output.c:533] execvp failed: No such file or directory
```
```
sway version 1.0-rc1-257-gbd32b300 (Apr 25 2019, branch 'master')
```

RedSoxFan:
swaybg has been split into a standalone project and can be found at https://github.com/swaywm/swaybg
See #4088

PR #7563, opened 2023-05-03
man: make it explicit that swaybg is required for background config

abowen:
I couldn't work out why the background wasn't working with my arch setup with sway

Only after running it in debug and seeing this error message, I came across this post

https://github.com/swaywm/sway/issues/4090

Double checked the man page and nothing stated that `swaybg` was required.


emersion:
swaybg is a sway dependency. Distros should reflect this, end users shouldn't need to care.

emersion:
Superseded by https://github.com/swaywm/sway/pull/7851

PR #7565, opened 2023-05-04
Add support for touch cancel events

hrdl-github:
https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/37f42e2d introduces touch cancel events to wlroots. At the moment touch sequences that are cancelled by libinput are not terminated by a cancel or end event. The only way client applications can become aware of this is by tracking touch ids. When a new touch sequence is started that reuses a previously cancelled touch id, said application can assume that the previous touch sequence was cancelled. However, this approach introduces a range of other issues. See https://github.com/xournalpp/xournalpp/pull/4852 for the original motivation.

I've performed some rudimentary tests on a Surface Go running a 6.2 mainline kernel. Do you have any wishes for an exhaustive test plan?

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -53,6 +53,7 @@ struct sway_cursor {
 
 	struct wl_listener touch_down;
 	struct wl_listener touch_up;
+	struct wl_listener touch_cancel;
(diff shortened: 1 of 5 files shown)


emersion (approved):
This LGTM, thank you!

PR #7567, opened 2023-05-04
handle_touch_cancel: fix begin default

hrdl-github:
I forgot to call seatop_begin_default in e8f7551e46052a8df04b630bf06565ca77f830fb. This prevents e.g. stylus devices from being usable immediately after they cancelled a touch point.


emersion (approved):
Indeed, thanks for the fix!

PR #7571, opened 2023-05-05
Add support for cursor-shape-v1

emersion:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4106

~~TODO: animated cursors. Ideally `wlr_xcursor_manager` would be patched to support this.~~ `wlr_cursor` now supports animated cursors.


dnkl:
foot: https://codeberg.org/dnkl/foot/pulls/1389 (not yet tested against patched sway+wlroots).

Issue #7399, opened 2023-01-24
[Bug confirmed] Input freezes on games with certain key combinations (when "hide_cursor when-typing enable" is present)


Zeioth:
### Please fill out the following:
- **Sway Version:** Tested from 1.7.6 to 1.8.3

-  **Debug Log:**
[LOG.](https://pastebin.com/raw/yxRWuw8Y)

- **Configuration File:**

```
seat * hide_cursor 899000
seat * hide_cursor when-typing enable
bindsym --to-code Mod4+d exec wofi -p run
xwayland enable
```

- **Description:**
  - When certain inputs are introduced, mouse freezes (only in games, not on desktop).
  - Started like a month ago.
  - The bug doesn't appear in Hyprland or any other DE, only sway.

- **How to reproduce**
  - Open Apex/Quake, or any other FPS game.
  - While you move from left to right, click your mouse and drag.
  - The camera will freeze until click is released.

Apologies for being too vague, but this is such an obscure bug that I'm not sure how to start debugging it. Only thing I'm sure is it's only present on sway. [I recorded a video commenting the issue](https://www.youtube.com/watch?v=7W8581YQqqs).

Zeioth:
log added. Possible xwayland issue

    [xwayland/xwm.c:837] unhandled X11 property 404 (_NET_WM_BYPASS_COMPOSITOR) for window 37748737

Reverting xwayland version and testing.

Zeioth:
Reverted xwayland version and restarted sway. Bug still present.

Zeioth:
Running sway with this config solves the issue.

    bindsym --to-code Mod4+d exec wofi -p run
    xwayland enable

Looling for the line causing the issue on my original config.

Zeioth:
Got it. I'm able to reproduce 100% of the time. This line is the culprit:

    seat * hide_cursor 899000
    seat * hide_cursor when-typing enable

Specifically, the second line.

Very likely there is something [in this commit](https://github.com/swaywm/sway/commit/96578aa91e9856bfb3e2d26fb7a625ff7c9b79e3) that is doing things that it should not.

bl4ckb0ne:
It doesnt really feel like a bug, more like "by design". What maybe sway could not hide the cursor when a specific application is focused? Otherwise you should just disable this during your gaming sessions.

Zeioth:
[...]

PR #7574, opened 2023-05-07
hide_cursor when-typing: do not hide cursor when buttons are pressed

### LINKED ISSUE #7383: hide_cursor while-typing interrupts drag operations
As mentioned in #7382.

- **Sway Version:** 1.8
- **Debug Log:** unsure if relevant?

- **Configuration File:** `seat * hide_cursor while-typing enable`

- **Description:**
  - Set the option above
  - Hold mouse button down, start dragging
  - Press any key on the keyboard
  - Notice that the ongoing drag operation was interrupted

#4424 fixed this for timeout-based hiding, but apparently not for while-typing. Might be the root cause of #6297.

Fiirgen:
This was already done for timeout based hiding in #4424.
Now the `hide_cursor when-typing enabled` setting will also not hide the cursor unless all buttons are released.

This fixes #7383, #6297, #7399.

I'm not sure if doing that check every time a key is pressed is the best way though.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -225,7 +225,10 @@ void cursor_notify_key_press(struct sway_cursor *cursor) {
 	}
 
 	if (cursor->hide_when_typing == HIDE_WHEN_TYPING_ENABLE) {
-		cursor_hide(cursor);
+		if (cursor->pressed_button_count == 0) {
+			// Do not hide cursor unless all buttons are released
+			cursor_hide(cursor);
+		}
 	}
 }
 


ajulien-fr:
Hi,
Is there a reason why this hasn't been merged yet ?
As stated in the first comment, this would make the behavior more consistent with timeout based hiding and fix several bugs.
It's also only one line of additional code.
Thanks

ajulien-fr:
@ddevault,
I see you're the one who merged this change for timeout based hiding. Could you please do the same for this pull request ?
Thanks.

PR #7580, opened 2023-05-09
Fix wlr_render_pass regressions

kennylevinsen:
Two regression fixes:

1. Apply clip properly without affecting the projection matrix
2. Clear using the proper dimensions when scaling is in use

Seems to do the trick locally.

### DIFF
--- a/sway/desktop/render.c
+++ b/sway/desktop/render.c
@@ -1027,7 +1027,7 @@ void output_render(struct render_context *ctx) {
 
 	if (debug.damage == DAMAGE_HIGHLIGHT) {
 		wlr_render_pass_add_rect(ctx->pass, &(struct wlr_render_rect_options){
-			.box = { .width = output->width, .height = output->height },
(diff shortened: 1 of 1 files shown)


emersion:
"render: Apply clip to rendered texture correctly" LGTM!

emersion:
Setting scale=1.5 and transform=90, then killing `swaybg`, reveals that using `wlr_output.{width,height}` is the right thing to do. With `wlr_output_transformed_resolution()`, then the cleared rect is invalid:

screenshot

kennylevinsen:
> Setting scale=1.5 and transform=90, then killing swaybg, reveals that using wlr_output.{width,height} is the right thing to do. With wlr_output_transformed_resolution(), then the cleared rect is invalid:

Updated to just change from `sway_output->{width, height}` to `wlr_output->{width, height}`

emersion (approved):
Thank you so much! I couldn't wrap my head around this bug.

PR #7585, opened 2023-05-13
Make the login manager set XDG_CURRENT_DESKTOP

t2b3:
According to the FreeDesktop specification at
<https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html>, DesktopNames signals to the login manager what XDG_CURRENT_DESKTOP shall be set to.

This commit has been successfully tested with GDM.

### DIFF
--- a/sway.desktop
+++ b/sway.desktop
@@ -3,3 +3,4 @@ Name=Sway
 Comment=An i3-compatible Wayland compositor
 Exec=sway
 Type=Application
+DesktopNames=sway


rpigott:
Might as well I guess.

I don't think "sway" is the right value though. The desktop entry spec lets XDG_CURRENT_DESKTOP be a list of descriptive strings. I reckon it should at least be "sway:wlroots" or similar — ideally there are any number of compositors and we don't want projects like xdg-desktop-portal-wlr to have to list them all to appease the selection logic in xdg-desktop-portal.

emersion:
In the past we've rejected this

kennylevinsen:
Many of our initial concerns - e.g. user portal choice - should go away with the newer, unreleased, portal configuration system though.

PR #7587, opened 2023-05-13
Add support for layer-shell on_demand keyboard interactivity

ErikReider:
Supersedes #5974 

Tested (from original PR) with gtk-layer-demo:

- [x] xdg toplevel + layer-surface/top/on-demand
- [x] xdg toplevel + layer-surface/top/on-demand + layer-surface/top/exclusive
- [x] xdg toplevel + layer-surface/top/on-demand + layer-surface/overlay/exclusive
- [x] layer-surface/top/exclusive + layer-surface/top/none: Should be able to interact with none layer, popups, and subsurfaces(buttons, no keyboard input)
- [x] focus_follows_mouse
- [x] layer-surface/top/on-demand + layer-surface/top/exclusive, on-demand only focused when focused inside of bounds while exclusive is focused otherwise (also outside of its bounds)
- [x] layer-surface/top/exclusive + layer-surface/top/exclusive: Layer has focus until the cursor enters/clicks within the other layers bounds
- [x] Keyboard input is only available for EXCLUSIVE and ON_DEMAND surfaces
- [x] Interact buttons in layers/subsurfaces/popups while NONE exclusive
- [x] Clicking on xdg_popups/layer sub-surfaces works with any mode
- [x] return keyboard focus after destroying on-demand layer surface

More test cases would be appreciated :)

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -17,6 +17,39 @@
 #include "sway/tree/arrange.h"
 #include "sway/tree/workspace.h"
 
+struct wlr_layer_surface_v1 *toplevel_layer_surface_from_surface(
+		struct wlr_surface *surface) {
+	struct wlr_layer_surface_v1 *layer;
+	do {
+		if (!surface) {
+			return NULL;
+		}
+		// Topmost layer surface
+		if ((layer = wlr_layer_surface_v1_try_from_wlr_surface(surface))) {
+			return layer;
+		}
+		// Layer subsurface
+		if (wlr_subsurface_try_from_wlr_surface(surface)) {
+			surface = wlr_surface_get_root_surface(surface);
+			continue;
+		}
+
+		// Layer surface popup
+		struct wlr_xdg_surface * xdg_popup = NULL;
+		if ((xdg_popup = wlr_xdg_surface_try_from_wlr_surface(surface)) &&
+				xdg_popup->role == WLR_XDG_SURFACE_ROLE_POPUP) {
+			if (!xdg_popup->popup->parent) {
+				return NULL;
+			}
+			surface = wlr_surface_get_root_surface(xdg_popup->popup->parent);
+			continue;
+		}
+
+		// Return early if the surface is not a layer/xdg_popup/sub surface
+		return NULL;
+	} while (true);
+}
+
 static void apply_exclusive(struct wlr_box *usable_area,
 		uint32_t anchor, int32_t exclusive,
 		int32_t margin_top, int32_t margin_right,
@@ -218,7 +251,8 @@ void arrange_layers(struct sway_output *output) {
 	for (size_t i = 0; i < nlayers; ++i) {
 		wl_list_for_each_reverse(layer,
 				&output->layers[layers_above_shell[i]], link) {
-			if (layer->layer_surface->current.keyboard_interactive &&
+			if (layer->layer_surface->current.keyboard_interactive
+					== ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE &&
 					layer->layer_surface->surface->mapped) {
 				topmost = layer;
 				break;
@@ -231,10 +265,12 @@ void arrange_layers(struct sway_output *output) {
 
 	struct sway_seat *seat;
 	wl_list_for_each(seat, &server.input->seats, link) {
+		seat->has_exclusive_layer = false;
 		if (topmost != NULL) {
 			seat_set_focus_layer(seat, topmost->layer_surface);
 		} else if (seat->focused_layer &&
-				!seat->focused_layer->current.keyboard_interactive) {
+				seat->focused_layer->current.keyboard_interactive
+					!= ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE) {
 			seat_set_focus_layer(seat, NULL);
 		}
 	}
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1295,11 +1295,15 @@ static void seat_set_workspace_focus(struct sway_seat *seat, struct sway_node *n
 }
 
 void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
-	if (seat->focused_layer) {
+	// Prevents the layer from losing focus if it has keyboard exclusivity
+	if (seat->has_exclusive_layer) {
 		struct wlr_layer_surface_v1 *layer = seat->focused_layer;
 		seat_set_focus_layer(seat, NULL);
 		seat_set_workspace_focus(seat, node);
 		seat_set_focus_layer(seat, layer);
+	} else if (seat->focused_layer) {
+		seat_set_focus_layer(seat, NULL);
+		seat_set_workspace_focus(seat, node);
 	} else {
 		seat_set_workspace_focus(seat, node);
 	}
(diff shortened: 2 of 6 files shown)


kennylevinsen:
> This PR allows for entering text into textboxes while the layer is focused.

How was the bug reproduced? Is it with or without keyboard exclusivity?

The git history of this PR is a little foo. Having a commit for each of the fixes with a title and description that fits would be best.

ErikReider:
> How was the bug reproduced? Is it with or without keyboard exclusivity?

This was tested with a layer without any keyboard exclusivity (a notification from swaync with this [PR](https://github.com/ErikReider/SwayNotificationCenter/pull/221)). In the video below, the popup notification was tested while the second layer has keyboard exclusivity and covers the whole output (it's transparent).

https://github.com/swaywm/sway/assets/35975961/9198422b-9440-4a3d-a346-01a4ef2538fc

Prior to the fixes, you'd be able to focus and move views below an exclusive layer shown in the video below:

https://github.com/swaywm/sway/assets/35975961/bf94fbe4-2251-453f-8fc5-de7ee843e37e


> The git history of this PR is a little foo. Having a commit for each of the fixes with a title and description that fits would be best.

I'm on it!

ErikReider:
I removed the commit that disallowed non-keyboard exclusive layers from receiving mouse clicks while there being a exclusive layer beside it (ex: waybar receiving clicks while rofi is visible), so ignore that part in the videos above.

kennylevinsen:
Hmm I'm a little confused between the original PR description and what the video displays.

A few notes about keyboard interactivity:
- It is only meant to affect (and if exclusive, steal) keyboard input and focus. Pointer, touch and tablet input and focus is not constrained and should always work as normal. This seems to be changed here.
- Keyboard interactivity defaults to `none`, where keyboard focus should never be granted to the surface.
- For `exclusive`, given a surface in TOP or OVERLAY, the top-most surface should be granted unconditional keyboard focus as long as it is mapped.
- For `on_demand`, the layer surface should behave like a normal surface.

More info here: https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:enum:keyboard_interactivity

SwayNotificationCenter popups with a text field would definitely want `on_demand`. sway does not seem to currently handle `on_demand`, and might be treating it as exclusive by accident as it just checks `if (keyboard_interactivity)`, only distinguishing between 0 (`none`) and anything else. We need to make the handling of the two original types explicit checks, allowing us to fall back to normal focus logic for `on_demand`.

ErikReider:
Makes sense. So adding a check if the surface is requesting on_demand would be fine then?

ErikReider:
I've got the changes necessary to advertise `ON_DEMAND`, should I create a new PR and close this one or change the title, desc, and redo the commits?

kennylevinsen:
Either works!

ErikReider:
Updated everything, would you like a clean commit history too?

sway/desktop/layer_shell.c
```diff
@@ -234,7 +235,9 @@ void arrange_layers(struct sway_output *output) {
 		if (topmost != NULL) {
 			seat_set_focus_layer(seat, topmost->layer_surface);
 		} else if (seat->focused_layer &&
-				!seat->focused_layer->current.keyboard_interactive) {
+				seat->focused_layer->current.keyboard_interactive
+					!= ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE) {
+			seat->exclusive_layer = NULL;
```
kennylevinsen, line 240: Note that there can be  multiple layer surfaces asking for exclusive keyboard interactivity, with the topmost one getting the privilege.

Maybe a simple "update_exclusive_layer" that iterates through them and picks the first one?

sway/input/seatop_default.c
```diff
@@ -366,19 +368,61 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
...
+		}
+		seat_pointer_notify_button(seat, time_msec, button, state);
+		return;
+	}
+	// Handle clicking a layer subsurface
+	struct wlr_subsurface * subsurface = NULL;
+	if (surface &&
+			(subsurface = wlr_subsurface_try_from_wlr_surface(surface)) &&
+			(layer_surface = subsurface->parent) &&
+			(layer = wlr_layer_surface_v1_try_from_wlr_surface(layer_surface)) &&
```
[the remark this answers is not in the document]
ErikReider: > To throw a wrench in the wheel, the surface might be a popup whose parent is a layer surface... :)

Oh so a subsurface might have a popup as a parent which has a layer as a parent?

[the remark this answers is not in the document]
kennylevinsen: It would return the popup surface, as that is the topmost surface in the subsurface tree.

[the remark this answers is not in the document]
ErikReider: Should be resolved in [a522cd7](https://github.com/swaywm/sway/pull/7587/commits/a522cd711230b913e1672dcc1fda91522c18011d)

sway/desktop/layer_shell.c
ErikReider: Don't the lines above do the same thing? Or am I misunderstanding what you've meant? 😅

kennylevinsen: D'oh, misread. Testing with two exclusive interactivity layers would be a good test though.

ErikReider: > D'oh, misread. Testing with two exclusive interactivity layers would be a good test though.

Works fine, the layer focus remains until I move the cursor/click inside of the other exclusive layer, and vise versa :)

kennylevinsen (approved):
Looking better! Still a thing remaining though. :)

sway/input/seatop_default.c
```diff
@@ -380,6 +383,13 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 		return;
 	}
 
+	// Don't change focus and don't send any inputs if there's a focused layer
+	// with keyboard exclusivity
+	if (seat->exclusive_layer) {
```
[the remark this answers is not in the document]
ErikReider: > Hmm, this guard also means that pointer events (moving windows, clicking buttons) do not reach other surfaces when keyboard interactivity is exclusive, which is not the intention.

Oh ok, though that that was a bug. Will fix :)

ErikReider:
Anything else that needs to be done?

sway/input/seat.c
```diff
@@ -1347,14 +1351,16 @@ void seat_set_focus_layer(struct sway_seat *seat,
...
+	} else if (!layer) {
 		return;
 	}
 	assert(layer->surface->mapped);
 	seat_set_focus_surface(seat, layer->surface, true);
 	if (layer->current.layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP) {
 		seat->focused_layer = layer;
 	}
+	seat_set_focus_surface(seat, layer->surface, true);
+	seat->focused_layer = layer;
```
kennylevinsen, line 1363: I think you forgot to add something that sets `has_exclusive_layer`, as it is currently never set to true. Maybe that is what you *meant* to write here?

The check that the layer is TOP or above should also be respected, so it would likely look something like:

```c
if (layer >= TOP) {
   focused_layer = layer;
   if (interactivity == EXCLUSIVE) {
     has_exclusivity = true;
   }
 }
 ```

ErikReider: Oh, merged the wrong conflict hunk... Fixing :)

kennylevinsen:
@ErikReider Do you want to reword the git commits before we merge? They talk about the original version which had the wrong focus behavior. A conflict also snuck in in the meantime.

Will be merged when that is out of the way. :)

PR #7595, opened 2023-05-20
Add clang-tidy configuration file

t2b3:
clang-tidy is a C/C++ linter.

Meson automatically detects the configuration file and defines a clang-tidy target that runs clang-tidy if available.

The targets can be invoked by `ninja clang-tidy`.


emersion:
Sorry, but we are not interested in clang-tidy.

PR #7596, opened 2023-05-20
Enclose macro replacement list in parentheses

t2b3:


emersion:
Not a fan of these macro tricks. If the lack of parentheses is a bother, maybe we can switch these to proper `const` variables instead?

t2b3:
I tried this but some switch statements require constant expressions (`#define`) 
`const` variables do not fulfill this requirement

kennylevinsen:
I suppose C2X's `constexpr` would solve that.

t2b3:
I think it is not so important, so just leave it for now.

PR #7597, opened 2023-05-20
Fix layer old damage not being offset by the monitor layout coords

ErikReider:
Issue in question:

https://github.com/swaywm/sway/assets/35975961/236f6a6a-d03c-4560-9334-3982e9aaddcc

To reproduce:

1. Open sway with two outputs
2. Run rofi-wayland, `rofi -show drun`
3. Type something in so that there are no results and hit enter
4. If the bug doesn't appear, wait a few seconds


emersion (approved):
Thank you!

PR #7598, opened 2023-05-20
Add support for tearing-control-v1

RicArch97:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3871

Adds option to allow tearing per output, as well as an option to force enable or disable tearing for a specific application using a window rule. Only works with fullscreen applications.


Sid127:
this is no longer blocked by the wlroots MR, would love to see this finished and merged

RicArch97:
Marked as ready for review. Works on my end (i see tearing) when setting tearing allowed on the output and an override per view, with `WLR_DRM_NO_ATOMIC=1`. Not tested using Mesa git and automatically setting the hint though. 

I don't know if my approach is correct, since my knowledge about (Wayland) compositors is limited, so if anything can be improved let me know.

sway/commands/tearing_allowed.c
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  bool wants_tearing = parse_boolean(argv[0], true);
+
+  struct sway_view *view = container->view;
+  view->tearing_mode = wants_tearing ? TEARING_OVERRIDE_TRUE : 
+    TEARING_OVERRIDE_FALSE;
+
+  return cmd_results_new(CMD_SUCCESS, NULL);
+}
```
nekopsykose, line 25: need a trailing newline here and in `sway/commands/output/tearing_allowed.c`

FakeMichau:
After playing for some time it can cause few seconds long freezes when switching between workspaces, closing some apps, trying to type something into thunar. Tho there are times when it's fine, I didn't find a consistent way of reproducing this. Temporary solution I've found is switching to a different TTY and back. But usually the issues comes back quite quickly. dmesg doesn't say anything.
Potentially worth noting is that I'm also using VRR, I didn't try to use this MR without VRR.
I've also seen this on hyprland but I don't have anything to prove it's wlroots' fault.

ErikReider:
I've been using the initial PR since may and everything has been working as expected! :)

feschber:
One thing, I'd like to see (and I'm not sure, this may be more related to wlroots) is tearing on the mouse cursor.
My reasoning is:
- Actually seeing the tearing on such a tiny object is pretty unlikely (24 / 1080 = ~every 45th frame)
- It is one area, where it makes the largest perceivable difference

Seeing this as an option would be awesome!

emersion:
I don't see how cursor tearing would actually be useful. Can't think of any case where minimizing latency on the cursor image would be useful. (Note, tearing on cursor would only apply to the cursor image, not the cursor position.)

feschber:
> I don't see how cursor tearing would actually be useful. Can't think of any case where minimizing latency on the cursor image would be useful. (Note, tearing on cursor would only apply to the cursor image, not the cursor position.)

Okay I thought it would make it possible to render the cursor earlier. If it does not affect the cursor position then it makes no sense.

I guess I will wait for the new frame scheduler to be merged and see if that improves things.

Specifically on 60Hz the cursor feels noticeably delayed currently and I thought this might help because the latency improvement in games is VERY apparent.

zeakzzz (approved):

sad-goldfish:
I've set:
```
for_window [title=".*"] tearing_allowed yes
output DP-1 { tearing_allowed 1 }
```

Is that enough to enable tearing with this patch (and unstable wlroots git)? I still see `New presentation hint 0 received for surface`. Anyone know of a good way to test if this is working?

sway/desktop/output.c
```diff
@@ -601,11 +630,15 @@ static int output_repaint_timer_handler(void *data) {
 	pending.committed |= WLR_OUTPUT_STATE_DAMAGE;
 	get_frame_damage(output, &pending.damage);
 
-	if (fullscreen_con && fullscreen_con->view && !debug.noscanout) {
+	if (output_can_tear_fullscreen_view(output, fullscreen_view)) {
+		pending.tearing_page_flip = true;
```
emersion, line 634: Asking for tearing can cause the page-flip to fail for hardware-specific reasons. Ideally we'd have some kind of fallback for that case.

emersion:
Not sure about the naming of the options, e.g. `output allow_tearing` seems to actually be a way to force tearing on/off.

sway/desktop/output.c
RicArch97: how do we know if a page has failed, and what do we want to do in in that case?

emersion: > how do we know if a page flip has failed

The output commits fails.

> what do we want to do in in that case?

Probably fall back to a regular page-flip?

include/sway/config.h
```diff
@@ -285,6 +285,7 @@ struct output_config {
 	int max_render_time; // In milliseconds
 	int adaptive_sync;
 	enum render_bit_depth render_bit_depth;
+	int tearing_allowed;
```
OctopusET, line 288: How about following the name of the `allow_tearing` option as well?

sad-goldfish:
I'm not sure if you know this already but I don't think this builds against any wlroots commit anymore. I've tried e.g. `b80337a8f2bb5d032f533ed5e0feb6292670528f`, `ce89f49b7aab281198fad64e9a825a24dbf72e3d`, `1968ada2132237f5bf8e40b6bf903fbce76c0b40` and `da5f53b46176f05731e94dec5f079c1ce7a57ed5`.

include/sway/config.h
RicArch97: Done.

RicArch97:
> I'm not sure if you know this already but I don't think this builds against any wlroots commit anymore. 

Fixed by rebase.

sad-goldfish:
Just gave it a go. It does build against current wlroots (`4688a371e04787f5bb059e80b5de506f0a8f00ff`) and stuff works after switching to xwayland `d7f1909e7c198607794da5f4d778ce53139ff851`. I no longer see the tearing I previously saw in `vkcube` but this may have been a bug previously. A better way to test tearing would be nice. Also, I'll test the new non-legacy atomic tearing page flips at some point too.

sad-goldfish:
Testing with Linux 6.8-rc1 and libdrm 2.4.120 without setting `WLR_DRM_NO_ATOMIC=1`, I don't see any tearing nor any debug messages suggesting that something is broken.

RicArch97:
> Tested the same application on X11 and I can see tearing. It seems like, for me, tearing is broken with the current state of this patch. Though, of course, it could be because of Wlroots changes underneath.

I see. I'll have to investigate this some more. Recently some massive changes were made to the codebase. I made the changes based on a general outline of it, but overall I'm not involved enough with Sway and wlroots to know every detail of it. I don't have any trouble implementing it, but i may need some help with the approach based on recent changes.

vimproved:
I've fixed this PR in my own fork in this commit: https://github.com/vimproved/sway/commit/49590198febe073528677c80f2b33ec74ead1c75. It seems like tearing_page_flip needs to be updated every frame, and putting it behind `if (output->gamma_lut_changed || output->tearing_state_changed)` broke it. However, now it's calling wlr_output_commit_state twice every frame (once to apply tearing_page_flip, and once in wlr_scene_output_commit), which might be an issue (I'm not 100% sure).

sad-goldfish:
@emersion could you comment on what changes need to be made to this patch (with or without @vimproved's changes) before it can be merged? As Linux 6.8 will probably be released soon with the atomic tearing page flips and Sway 1.9 has been released (so timing for new features is good), it seems like a good time to work on this. I'd be happy to help work on this if there are any concerns that you have.

zeak-z:
any news on when this will be merged?

KyunLFA:
For those trying to use this MR with upstream and failing to see any tearing, there's a working version at https://github.com/vimproved/sway/tree/tearing_control (thanks vimproved for maintaining the MR and thanks RicArch97 for creating it!).

Test tearing with:

`vkcube --present_mode 0`

It should show very visible tearing.

Don't forget to fullscreen the window!

RicArch97:
> For those trying to use this MR with upstream and failing to see any tearing, there's a working version at https://github.com/vimproved/sway/tree/tearing_control (thanks vimproved for maintaining the MR and thanks RicArch97 for creating it!).
> 
> Test tearing with:
> 
> `vkcube --present_mode 0`
> 
> It should show very visible tearing.
> 
> Don't forget to fullscreen the window!

My latest changes already included the suggestion from @vimproved (thanks!). The difference is that i didn't rebase this branch in some time. Last time i tested it, the output commits were failing with the tearing flips enabled so i assumed the backend (wlroots, DRM) still needed work before the tearing flips could work correctly.

I definitely haven't forgotten about this MR, i just don't recommend using this yet until it's actually ready and reviewed. I just rebased it as well, and tested it with vkcube and overwatch 2. The commits no longer fail and the tearing works as expected (including using the hints provided by the application / game) without the need to launch with `WLR_DRM_NO_ATOMIC=1`. So we seem to be moving in the right direction.

KyunLFA:
@RicArch97 I understand 100%, I choose to use experimental stuff all the time (you'd probably cringe at my heavily experimental and patched custom compiled OS bits...) but I can get why you don't quite recommend it yet before the proper code review, but I think you'd agree people who can patch and compile code from github know enough about what they're doing to not throw pitchforks at you if their compositor crashes 😂. Thank you for the official rebase of this MR and hope it gets reviewed and merged in the near future!

feschber:
I'm sorry, but I'm pulling my hair out trying to get this to work.
```sway
output * allow_tearing yes

for_window [title=".*"] allow_tearing yes
```

Running an application that should tear just doesn't show tearing and resets the `allow_tearing` flag in
sway apparently:

```sh
$ swaymsg -t get_outputs | rg allow_tearing
    "allow_tearing": true,
    "allow_tearing": true,

$ vkcube --present_mode 0
Selected GPU 0: AMD Radeon RX 6800 (RADV NAVI21), type: DiscreteGpu

$ swaymsg -t get_outputs | rg allow_tearing
    "allow_tearing": false,
    "allow_tearing": true,
```

```sh
swaymsg version 1.10-dev-203ac891 (Mar 18 2024, branch 'tearing_control')
```

feschber:
To add to this: I'm on Kernel 6.9.5 and I'm definitely getting PRESENT_MODE_IMMEDIATE_KHR reported by Vulkan.

I also never got it to work in KDE ... so maybe its an issue with the kernel

RicArch97:
> I'm sorry, but I'm pulling my hair out trying to get this to work.
> 
> ```sway
> output * allow_tearing yes
> 
> for_window [title=".*"] allow_tearing yes
> ```
> 
> Running an application that should tear just doesn't show tearing and resets the `allow_tearing` flag in sway apparently:
> 
> ```shell
> $ swaymsg -t get_outputs | rg allow_tearing
>     "allow_tearing": true,
>     "allow_tearing": true,
> 
> $ vkcube --present_mode 0
> Selected GPU 0: AMD Radeon RX 6800 (RADV NAVI21), type: DiscreteGpu
> 
> $ swaymsg -t get_outputs | rg allow_tearing
>     "allow_tearing": false,
>     "allow_tearing": true,
> ```
> 
> ```shell
> swaymsg version 1.10-dev-203ac891 (Mar 18 2024, branch 'tearing_control')
> ```

Are you sure you are on the latest version of this branch? The date reported in your version seems odd. Anyway, when testing it 2 months ago i had a similar issue where the output commits seemed to fail. Currently for me that is no longer the case. Note that basically the first time that happens when tearing flips get enabled, we disable the tearing, otherwise Sway would stop working entirely.

I tested it with Linux 6.9.5, Mesa 24.1.1, libdrm 2.4.120 (GPU: Radeon 6900XT) and everything works as expected.

RicArch97:
Oops, it looks like i still had `WLR_DRM_NO_ATOMIC=1` set in my dev shell. So for now it looks like that is still required to set this to make the tearing work....

feschber:
> Oops, it looks like i still had `WLR_DRM_NO_ATOMIC=1` set in my dev shell. So for now it looks like it is still required to set this to make the tearing work....

Yes, its working with `WLR_DRM_NO_ATOMIC=1`, glad to here its nothing specific to my system I guess!

vimproved:
> My latest changes already included the suggestion from @vimproved (thanks!).

No problem, thank you for making this PR in the first place! I use the live version of sway on gentoo with this patch so I've mainly just been rebasing it for personal use but I'm glad other people have found it useful. For now `WLR_DRM_NO_ATOMIC=1` seems to still be required, I tried looking into why it wasn't working at some point but came up with inconclusive results.

somewhatfrog:
I have a question, will there be an option to enable tearing for everything, so there's no need to manually add each and every app?

ZakMcKrack3n:
> I have a question, will there be an option to enable tearing for everything, so there's no need to manually add each and every app?

I think this is already possible by just defining a window rule using some wildcards, e.g. everything that gets fullscreen should then be configured to tear, still learning the syntax here as well.

Wonder if there would be a global 'tear all fullscreen option' , if it could be toggled on the fly as well (binding keys etc.), not sure what is possible and what the limitations are, maybe it can only be set when getting to/from fullscreen.

somewhatfrog:
> > I have a question, will there be an option to enable tearing for everything, so there's no need to manually add each and every app?
> 
> I think this is already possible by just defining a window rule using some wildcards, e.g. everything that gets fullscreen should then be configured to tear, still learning the syntax here as well.
> 
> Wonder if there would be a global 'tear all fullscreen option' , if it could be toggled on the fly as well (binding keys etc.), not sure what is possible and what the limitations are, maybe it can only be set when getting to/from fullscreen.

In this case will sway count as fullscreen? What I meant is no sync at all for everything including tiles, not just apps which went into fullscreen.

ZakMcKrack3n:
> In this case will sway count as fullscreen? What I meant is no sync at all for everything including tiles, not just apps which went into fullscreen. Like a global toggle.

This should be the output option:

`output * allow_tearing yes`

if I read this correctly some posts back.

RicArch97:
The idea of the tearing hints is that the application can inform the compositor that it wants to tear. For games that would be when VSync is disabled in the game. By default we react to the hints if `allow_tearing` is set to true on the output.

The window rules `allow_tearing` can be using if you want to force it for a specific app (maybe if the app doesn't provide a hint for async) or force disable it for an app to get the current mailbox present mode (VSync off in game, but no tearing allowed by compositor).

somewhatfrog:
> The idea of the tearing hints is that the application can inform the compositor that it wants to tear. For games that would be when VSync is disabled in the game. By default we react to the hints if `allow_tearing` is set to true on the output.
> 
> The window rules `allow_tearing` can be used if you want to force it for a specific app (maybe if the app doesn't provide a hint for async) or force disable it for an app to get the current mailbox present mode (VSync off in game, but no tearing allowed by compositor).

Shouldn't the compositor be disabled completely when game is in fullscreen to utilize game's own sync approach? For instance right now in xorg without compositor a fullscreen game with vsync ON has no tearing and has lower input delay than the same game under wayland (no tearing) where input delay feels higher under the same conditions, disabling vsync ingame does help here while there's still no tearing. This is from my own testing with a 240fps camera on 60hz screen and frame counting from input to change on screen from about a year ago, game's built in sync always seem to do a better job than compositor.

Vinjul1704:
> The idea of the tearing hints is that the application can inform the compositor that it wants to tear. For games that would be when VSync is disabled in the game. By default we react to the hints if `allow_tearing` is set to true on the output.
> 
> The window rules `allow_tearing` can be used if you want to force it for a specific app (maybe if the app doesn't provide a hint for async) or force disable it for an app to get the current mailbox present mode (VSync off in game, but no tearing allowed by compositor).

So if I understand correctly, the `allow_tearing` option for outputs is that tearing is possible/allowed if requested, but the `allow_tearing` option for windows forces it? In that case, wouldn't it make sense to rename the window option to something like `force_tearing`?

amarakon:
> Shouldn't the compositor be disabled completely when game is in fullscreen to utilize game's own sync approach or whatever setting user decides to choose? For instance right now in xorg without compositor a fullscreen game with vsync ON has no tearing and has lower input delay than the same game under wayland (no tearing) where input delay feels higher under the same conditions, disabling vsync ingame does help here while there's still no tearing. This is from my own testing with a 240fps camera on 60hz screen and frame counting from input to change on screen from about a year ago, game's built in sync always seem to do a better job than compositor. Sorry if I am missing something, I have only brief understanding in what's happening under the hood in wayland.

What you call a window manager on X11 is a compositor on Wayland. You can't turn off the compositor on Wayland just like you can't turn off the window manager on X11. What you probably want is “direct scanout.” Direct scanout basically means the application contents are put directly on the screen. This improves latency on compositors that support it. I don't know if Sway does because I use Hyprland which supports both tearing and direct scanout.

somewhatfrog:
> > Shouldn't the compositor be disabled completely when game is in fullscreen to utilize game's own sync approach or whatever setting user decides to choose? For instance right now in xorg without compositor a fullscreen game with vsync ON has no tearing and has lower input delay than the same game under wayland (no tearing) where input delay feels higher under the same conditions, disabling vsync ingame does help here while there's still no tearing. This is from my own testing with a 240fps camera on 60hz screen and frame counting from input to change on screen from about a year ago, game's built in sync always seem to do a better job than compositor. Sorry if I am missing something, I have only brief understanding in what's happening under the hood in wayland.
> 
> What you call a window manager on X11 is a compositor on Wayland. You can't turn off the compositor on Wayland just like you can't turn off the window manager on X11. What you probably want is “direct scanout.” Direct scanout basically means the application contents are put directly on the screen. This improves latency on compositors that support it. I don't know if Sway does because I use Hyprland which supports both tearing and direct scanout.
> 
> EDIT: This [blog post](https://zamundaaa.github.io/wayland/2021/12/14/about-gaming-on-wayland.html) tests input latency on Wayland and X11. You can see that X without compositing is almost identical to Wayland with tearing. However, it was tested without direct scanout, so it's possible that Wayland with tearing and direct scanout has even less input latency than X.

sway has direct scanout but as of now it still adds extra input latency to fullscreen games, feels awful
i also tried hyprland, feels just as bad compared to xorg without compositor (ingame vsync on)

amarakon:
> sway has direct scanout but as of now it still adds extra input latency to fullscreen games, feels awful
i also tried hyprland, feels just as bad compared to xorg without compositor (ingame vsync on)

If I understand correctly, you compared X with ingame vsync to Hyprland with the compositor's vsync? It could be because the ingame vsync is better than the compositor's vsync. I play all games on Hyprland with tearing enabled and direct scanout for the lowest latency, and I don't notice a difference compared to X. Perhaps you're playing Proton games which run through Xwayland which adds a bit of latency compared to native Wayland.

sway/desktop/output.c
```diff
@@ -243,28 +266,36 @@ static int output_repaint_timer_handler(void *data) {
...
+				server.gamma_control_manager_v1, output->wlr_output);
+			if (!wlr_gamma_control_v1_apply(gamma_control, &pending)) {
+				wlr_output_state_finish(&pending);
+				return 0;
+			}
+		}
+
+		if (output->can_tear) {
+			pending.tearing_page_flip = true;
+			tearing_state_updated = true;
```
emersion, line 291: Hm, it seems like these changes try to add tearing logic on top of the gamma LUT stuff? I don't think that works: if tearing fails the gamma LUT fails to apply, and the other way around. Moreover, if tearing fails there is no retry step.

emersion: This might help a bit: https://github.com/swaywm/sway/commit/1e0031781fc9283db7096aba34deca5503c2ab91

vimproved: Just rebased my fork (https://github.com/vimproved/sway/tree/tearing_control) against https://github.com/swaywm/sway/commit/1e0031781fc9283db7096aba34deca5503c2ab91, so hopefully that should help.

> Moreover, if tearing fails there is no retry step.
Ah, seems I managed to accidentally get rid of the fallback somewhere in my fork. Will fix.

RicArch97: Changed.

When the output state test fails, we now try again but with tearing page-flip disabled. I hope this is what you meant.

It works fine with atomic disabled, but with atomic enabled, the page-flip seems to always fail with tearing enabled:

```
00:00:14.731 [ERROR] [sway/desktop/output.c:299] Output test failed on 'DP-1', retrying without tearing page-flip
00:00:14.737 [DEBUG] [wlr] [backend/drm/atomic.c:79] connector DP-1: Atomic commit failed: Invalid argument
00:00:14.737 [DEBUG] [wlr] [backend/drm/atomic.c:84] (Atomic commit flags: PAGE_FLIP_ASYNC | ATOMIC_TEST_ONLY)
00:00:14.737 [ERROR] [sway/desktop/output.c:299] Output test failed on 'DP-1', retrying without tearing page-flip
00:00:14.743 [DEBUG] [wlr] [backend/drm/atomic.c:79] connector DP-1: Atomic commit failed: Invalid argument
00:00:14.743 [DEBUG] [wlr] [backend/drm/atomic.c:84] (Atomic commit flags: PAGE_FLIP_ASYNC | ATOMIC_TEST_ONLY)
```

Any idea why this happens, or what the status is in case it's a known issue?

sway/desktop/output.c
kode54: It can also fail if you have hardware cursors enabled and a commit includes a cursor framebuffer ID change, ie. a cursor change. And this property change will persist until it succeeds. And from the kernel source, this sort of property change must be committed synchronously.

I proposed a change to wlroots, but Simon Ser said it should ideally be handled by the compositors, or at least, compositors should be aware of when flips fail and opt in to doing something different.

It may be possible to separate out the non-primary plane actions to a synchronous commit, but again, it should probably be the compositor opting into doing this.

sway/desktop/output.c
```diff
@@ -268,6 +291,22 @@ static int output_repaint_timer_handler(void *data) {
 			wlr_output_state_set_gamma_lut(&pending, 0, NULL, NULL, NULL);
 		}
 	}
+	
+	if (output->can_tear) {
+		pending.tearing_page_flip = true;
+		
+		if (!wlr_output_test_state(output->wlr_output, &pending)) {
+			sway_log(SWAY_ERROR, "Output test failed on '%s', retrying without tearing page-flip",
```
emersion, line 299: I'd suggest downgrading this to `SWAY_DEBUG`, because we will hit this quite often when the cursor plane is used.

sway/desktop/output.c
```diff
@@ -268,6 +291,22 @@ static int output_repaint_timer_handler(void *data) {
...
+				output->wlr_output->name);
+				
+			pending.tearing_page_flip = false;
+			
+			if (!wlr_output_test_state(output->wlr_output, &pending)) {
+				wlr_output_state_finish(&pending);
+				return 0;
+			}
```
emersion, line 307: This one can be removed: we're doing a real commit right after, and the real commit implicitly performs a test before being applied.

include/sway/output.h
```diff
@@ -73,6 +73,7 @@ struct sway_output {
 	int max_render_time; // In milliseconds
 	struct wl_event_source *repaint_timer;
 	bool gamma_lut_changed;
+	bool allow_tearing, can_tear;
```
emersion, line 76: Maybe rename `can_tear` to `enable_tearing`?

sway/commands/allow_tearing.c
```diff
@@ -0,0 +1,25 @@
...
+		return error;
+	}
+
+  struct sway_container *container = config->handler_context.container;
+  if (!container || !container->view) {
+    return cmd_results_new(CMD_INVALID,
+				"Tearing can only be allowed on views");
+  }
+
+  bool wants_tearing = parse_boolean(argv[0], true);
```
emersion, line 18: Should we have a way to reset this back to the default?

sway/commands/allow_tearing.c
```diff
@@ -0,0 +1,25 @@
+#include <sway/commands.h>
+#include "sway/config.h"
+#include "sway/tree/view.h"
+#include "util.h"
+
+struct cmd_results *cmd_allow_tearing(int argc, char **argv) {
+  struct cmd_results *error = NULL;
```
emersion, line 7: Style: indent contains spaces instead of tabs here (applies to elsewhere in the file as well)

sway/tree/view.c
```diff
@@ -1260,6 +1261,19 @@ bool view_is_transient_for(struct sway_view *child,
 		child->impl->is_transient_for(child, ancestor);
 }
 
+bool view_can_tear(struct sway_view *view) {
+	switch(view->tearing_mode) {
```
emersion, line 1265: Style: missing space between `switch` and `(`

sway/sway-output.5.scd
```diff
@@ -190,6 +190,19 @@ must be separated by one space. For example:
 	may have no effect or produce unexpected output when used together with future
 	HDR support features.
 
+*output* <name> allow_tearing yes|no
+	Allows or disallows screen tearing as a result of asynchronous page flips,
```
emersion, line 194: I'd avoid the term "asynchronous", because it's very ambiguous/overloaded. Maybe "immediate" would be better.

sway/sway-output.5.scd
```diff
@@ -190,6 +190,19 @@ must be separated by one space. For example:
 	may have no effect or produce unexpected output when used together with future
 	HDR support features.
 
+*output* <name> allow_tearing yes|no
```
emersion, line 193: Maybe we should document the default setting ("no").

sway/sway-output.5.scd
```diff
@@ -190,6 +190,19 @@ must be separated by one space. For example:
...
+	Allows or disallows screen tearing as a result of asynchronous page flips,
+	and an immediate presentation mode from a client.
+
+	With asynchronous page flips, frames from the client are presented as soon
+	as possible instead of synchronizing with the monitor's vblank interval 
+	(VSync). This prevents stutter and reduces latency in games.
```
emersion, line 199: "This prevents stutter and reduces latency in games."

This is pretty subjective and debatable, so I'd prefer to leave out that part.

sway/sway-output.5.scd
```diff
@@ -190,6 +190,19 @@ must be separated by one space. For example:
...
+
+	To adjust whether tearing is allowed for specific applications, see
+	*allow_tearing* in *sway*(5).
+
+	This setting only has effect on fullscreen windows.
```
emersion, line 204: Maybe "only has effect when a window is fullscreen on the output" would be a tad better, since this is about the output command?

sway/sway-output.5.scd
```diff
@@ -190,6 +190,19 @@ must be separated by one space. For example:
...
+*output* <name> allow_tearing yes|no
+	Allows or disallows screen tearing as a result of asynchronous page flips,
+	and an immediate presentation mode from a client.
+
+	With asynchronous page flips, frames from the client are presented as soon
+	as possible instead of synchronizing with the monitor's vblank interval 
+	(VSync). This prevents stutter and reduces latency in games.
+
+	To adjust whether tearing is allowed for specific applications, see
+	*allow_tearing* in *sway*(5).
```
emersion, line 202: Maybe we could repeat here that tearing will only be enabled when it's allowed for both the output and the application.

sway/desktop/output.c
RicArch97: Changed.

sway/desktop/output.c
RicArch97: Removed.

include/sway/output.h
RicArch97: Renamed.

sway/commands/output/allow_tearing.c
```diff
@@ -0,0 +1,22 @@
...
+
+struct cmd_results *output_cmd_allow_tearing(int argc, char **argv) {
+	if (!config->handler_context.output_config) {
+		return cmd_results_new(CMD_FAILURE, "Missing output config");
+	}
+	if (argc == 0) {
+		return cmd_results_new(CMD_INVALID, "Missing allow_tearing argument");
+	}
+
+	if (parse_boolean(argv[0], true)) {
```
emersion, line 13: I think we need to pass `config->handler_context.output_config->allow_tearing == 1` here instead of `true`, to make `toggle` work.

sway/commands/allow_tearing.c
RicArch97: I my mind, if you want the default behavior, you just remove the window rule. I can't think of a reason why you would `explicitly` set it to the default behavior?
 
 But if you do think we need it, maybe we could add an `auto` option?

sway/commands/allow_tearing.c
RicArch97: Fixed.

By default i use spaces as indentation in vscode, and when creating those new files, it used that setting without it really being noticeable 🙂

sway/tree/view.c
RicArch97: Fixed.

sway/sway-output.5.scd
RicArch97: Changed to "immediate".

sway/sway-output.5.scd
RicArch97: Removed.

sway/sway-output.5.scd
RicArch97: Changed to "This setting only has effect when a window is fullscreen on the output."

sway/sway-output.5.scd
RicArch97: Added an extra comment to explain that tearing will only be enabled when it's allowed for the output and application.

sway/commands/output/allow_tearing.c
RicArch97: Changed.

sway/desktop/output.c
vimproved: > It can also fail if you have hardware cursors enabled and a commit includes a cursor framebuffer ID change, ie. a cursor change. And this property change will persist until it succeeds. And from the kernel source, this sort of property change must be committed synchronously.
> 
> I proposed a change to wlroots, but Simon Ser said it should ideally be handled by the compositors, or at least, compositors should be aware of when flips fail and opt in to doing something different.
> 
> It may be possible to separate out the non-primary plane actions to a synchronous commit, but again, it should probably be the compositor opting into doing this.

Right, now I remember that was my end result of my investigation as to why atomic async page flips weren't working with this PR. Would it be possible to separate out the page flips for non-primary panes to a different synchronous commit, and if so, would that introduce any new issues?

vimproved: Just found your solution to this in your labwc AUR package and applied an equivalent solution in my tearing_control branch. Thanks for the help here!

kode54: I intend to file it as a PR there when the wlroots 0.18 changeover happens

sway/desktop/output.c
emersion: There are multiple issues with atomic on the kernel side. See:

- https://lore.kernel.org/dri-devel/20240629152204.666748-1-contact@emersion.fr/
- https://lore.kernel.org/dri-devel/20240702212215.109696-2-andrealmeid@igalia.com/

sway/desktop/output.c
```diff
@@ -330,9 +382,12 @@ static void handle_frame(struct wl_listener *listener, void *user_data) {
...
 	int delay = msec_until_refresh - output->max_render_time;
 
+	struct sway_view *fullscreen_view = output_get_fullscreen_view(output);
+	output->enable_tearing = output_can_tear_fullscreen_view(output, fullscreen_view);
+
 	// If the delay is less than 1 millisecond (which is the least we can wait)
-	// then just render right away.
-	if (delay < 1) {
+	// or if the output is allowed to tear, then just render right away.
+	if (delay < 1 || output->enable_tearing) {
```
emersion, line 390: Same here: not sure we want to mess with output timings here. The default settings should result in `delay == 0`.

emersion (commented):
Here's my last round of comments!

sway/commands/allow_tearing.c
emersion: I'm thinking of users experimenting with the new commands, toggling to `on` and then trying to go back to the defaults. But yeah, I agree that it's easy enough to just reload the config file, and a way to reset to the default can be introduced later if it turns out it's desirable.

sway/desktop/output.c
```diff
@@ -141,11 +141,26 @@ static struct buffer_timer *buffer_timer_get_or_create(struct wlr_scene_buffer *
 	return timer;
 }
 
+static bool output_can_tear_fullscreen_view(struct sway_output *output, 
+		struct sway_view *view) {
+	if (!view) {
+		return false;
+	}
+#ifdef WLR_HAS_DRM_BACKEND
+	if (wlr_backend_is_drm(output->wlr_output->backend) &&
```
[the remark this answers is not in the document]
RicArch97: Removed the DRM check. I renamed the `output_get_fullscreen_view` function to `output_can_tear` and moved the two remaining to conditions to it. The `output_get_fullscreen_view` function was only used for tearing anyways.

sway/desktop/output.c
```diff
@@ -170,6 +189,10 @@ static void send_frame_done_iterator(struct wlr_scene_buffer *buffer,
 	int delay = data->msec_until_refresh - output->max_render_time
 			- view_max_render_time;
 
+	if (view_can_tear) {
+		delay = 0;
```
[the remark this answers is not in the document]
RicArch97: I was not sure initially, and after testing it did not seem to make a difference, so the delay is no longer changed now.

sway/desktop/output.c
RicArch97: Unfortunately, changing this DID make a difference. When i removed the check for tearing enabled, and ran `vkcube --present_mode 0` in fullscreen, the cube looked like the upper half was vsync-ed, and the bottom half was tearing. Like a mix of both.

If this is unexpected, then maybe we have a bug somewhere else?

emersion: Can you print `msec_until_refresh` and `output->max_render_time`? Both should be zero with a default config.

RicArch97: Ah yes i did have `max_render_time` set in my config, oops. It does appear to work correctly with `max_render_time off`. 

The question is whether we want to ignore the setting for `max_render_time` when tearing is requested? As tearing would imply that we want to render as fast as possible?

RicArch97: Removed the extra check.

Perhaps we could mention in the doc that `max_render_time` should not be set for tearing to work correctly?

emersion: Hm, to me tearing works "correctly" when `max_render_time` is set, because I see these as completely orthogonal. For instance, when setting `max_render_time` to a very low value and combining it with tearing, one can render closer to the vsync and make it so in case rendering takes too long the frame tears with the old frame for the top and the new frame for the bottom.

+1 for specifying in the docs that disabling `max_render_time` is the behavior most people will want, by flipping as soon as a client updates.

RicArch97: Ah, yes I didn't think about that. This is what's called "FIFO relaxed" behavior, right?

emersion: It's a bit similar: if we're on time it shouldn't tear, if we're late it should tear. However, there is no logic to force vsync when we're on time, and it's not a FIFO because there's no queue in which client frames accumulate. So still a bit different.

RicArch97: Added an extra note about the `max_render_time` setting under the `allow_tearing` option in the `sway-output` man page.

emersion (approved):
Thank you very much, nice work!

feschber:
@emersion @RicArch97 could you comment on what is required to make this work without setting `WLR_DRM_NO_ATOMIC=1`?

Stoppedpuma:
@emersion Is there anyway we could get a "Currently tearing" or something similar in swaymsg or somewhere? I have a few windows where I can't tell if they are tearing or not, an indicator would be helpful.

emersion:
I agree that would be useful, I think it would make sense to add a bool to the IPC to indicate whether the last flip was tearing.

sway/desktop/output.c
llyyr: The second patch was merged last week, but the first one still hasn't been despite getting a R-b over a month ago. Perhaps it fell through the cracks and needs a ping?

emersion: v2 for second patch is here and waiting for feedback: https://lore.kernel.org/dri-devel/20240731191014.878320-1-contact@emersion.fr/

Issue #7593, opened 2023-05-18
Window resizing with "ppt" unit seems relative instead of absolute


Delgan:
Hi.

I observed a different behavior than i3 when resizing non-floating windows with percentage point values.

For both i3 and Sway my config is basically:

```ini
mode "resize" {
        bindsym $left       resize shrink width  10 px or 10 ppt
        bindsym $down       resize shrink height  10 px or 10 ppt
        bindsym $up         resize grow height  10 px ot 10 ppt
        bindsym $right      resize grow width 10 px or 10 ppt

        bindsym Left        resize shrink width 10 px or 10 ppt
        bindsym Down        resize shrink height 10 px or 10 ppt
        bindsym Up          resize grow height 10 px or 10 ppt
        bindsym Right       resize grow width 10 px or 10 ppt

        bindsym Return mode "default"
        bindsym Escape mode "default"
        bindsym $mod+r mode "default"
}

bindsym $mod+r mode "resize"
```

It seems i3 transforms `10 ppt` to "10% of screen size" while using Sway, it's "10% of container"?

https://github.com/swaywm/sway/assets/4193924/641eaed9-c7df-4c8e-87a6-f1aeebdf0013

[resize-sway.webm](https://github.com/swaywm/sway/assets/4193924/6300ad46-8891-4e1f-8a0d-0a8055ecd133)

Am I the only one experiencing this behavior?

```shell
~ $ i3 --version
i3 version 4.22 (2023-01-02) © 2009 Michael Stapelberg and contributors
~ $ sway --version
sway version 1.8.1
```

Rintse:
I wrote a little script that mimics the i3 behaviour before I found this issue.
I'll post it here for reference:

```
#!/bin/bash
shrink_or_grow=${1}
direction=${2}
percent=${3}

if [ "${percent}" -lt 0 ] || [ "${percent}" -gt 100 ]; then
    echo "Invalid percentage"
fi

parent=$(swaymsg -t get_tree | jq '
    .. 
    | select(.type?) 
    | . as $parent
    | .nodes 
    | select(any(.[]; .focused == true))
    | $parent'
)

parent_dir=$(echo "${parent}" | jq -r '.layout')

if [ "$parent_dir" == "splitv" ]; then
    parent_size=$(echo "${parent}" | jq '.rect.height')
elif [ "$parent_dir" == "splith" ]; then
    parent_size=$(echo "${parent}" | jq '.rect.width')
else
    echo "Parent is not a split"
[...]

PR #7624, opened 2023-06-05
Calculate tiled resizing amount from parent/workspace

nukoseer:
i3 shrinks/grows tiled windows according to screen size for ppt unit. So, for i3 compatibility, I've changed how sway calculates resizing amount for tiled windows.

Resolves: #7593

You can find the demonstration of default behaviour on issue #7593 and here is after this change:

https://github.com/swaywm/sway/assets/17731403/3e9a200c-10e7-4fb4-94d7-c1e9ee99a99a

Note: This is my first time trying to contribute to open source. Please feel free for suggestions/advices. 😇

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -253,12 +253,27 @@ static struct cmd_results *resize_adjust_tiled(uint32_t axis,
 		amount->unit = MOVEMENT_UNIT_PPT;
 	}
 	if (amount->unit == MOVEMENT_UNIT_PPT) {
+        struct sway_container *parent = current->pending.parent;
 		float pct = amount->amount / 100.0f;
 
 		if (is_horizontal(axis)) {
-			amount->amount = (float)current->pending.width * pct;
+            while (parent && parent->pending.layout != L_HORIZ) {
+                parent = parent->pending.parent;
+            }
+            if (parent) {
+                amount->amount = (float)parent->pending.width * pct;
+            } else {
+                amount->amount = (float)current->pending.workspace->width * pct;
+            }
 		} else {
-			amount->amount = (float)current->pending.height * pct;
+            while (parent && parent->pending.layout != L_VERT) {
+                parent = parent->pending.parent;
+            }
+            if (parent) {
+                amount->amount = (float)parent->pending.height * pct;
+            } else {
+                amount->amount = (float)current->pending.workspace->height * pct;
+            }
 		}
 	}
 


rpigott:
> i3 shrinks/grows tiled windows according to screen size for ppt unit.

I'm pretty sure i3 actually uses the size of the parent container. I suppose we should change to match.

nukoseer:
> I'm pretty sure i3 actually uses the size of the parent container. I suppose we should change to match.

Actually, I was thinking do it according to parent container, but as far as I remember parent was always NULL. So, I decided to do it relative to workspace size.

nukoseer:
Updated. If parent exists, sway should use parent size instead of workspace size for tiled resizing amount(PPT).

sway/commands/resize.c
```diff
@@ -256,9 +256,21 @@ static struct cmd_results *resize_adjust_tiled(uint32_t axis,
 		float pct = amount->amount / 100.0f;
 
 		if (is_horizontal(axis)) {
-			amount->amount = (float)current->pending.width * pct;
+            struct sway_container *parent = current->pending.parent;
```
rpigott, line 259: This logic is still incorrect, you will need to find the first parallel parent. There is a function `container_find_resize_parent` that does this already that is called by `container_resize_tiled`. Ideally we would need to call it only once to determine the parent.

rpigott (changes requested):

sway/commands/resize.c
nukoseer: When I try to find the parent with `container_find_resize_parent(current, axis)` call it returns the same `current` pointer back (even if parent exists, idk why yet) and resizing does not work as it should. So, I think I need to check if `parent == current` and if they are equal I should find parent manually again (`parent = current->pending.parent`). I also realized the `container_resize_tiled` function that you mentioned is called at the end of this function that I am working on.

nukoseer:
Update. I couldn't get the correct behaviour with `container_find_resize_parent`. Instead of using that function, I tried to find the parent with simple while-loop. Here is also video to check if behaviour is right (sorry for low resolution): @rpigott 

https://github.com/swaywm/sway/assets/17731403/e45427cb-1b0b-4f31-a037-7bccfacb1e9e

nukoseer:
Do you think this can be merged or is there still some incorrect logic for calculating size? @rpigott

rpigott (changes requested):
Sorry bout that, forgot about this one.

Logic looks correct, but please begin the else clauses on the same line as the closing brace for style.

Also, you should squash these commits into one before it can be merged.

N.B. Unlike public branches, it is usually proper etiquette to force-push to your PR branches so that the PR reflects the intended history to be merged upstream — not the intermediate work it took to get there. We can compare the aggregate diffs between versions if we want to know what changed between pushes.

rpigott (approved):

rpigott:
LGTM, thanks!

PR #7625, opened 2023-06-05
chase wlroots wlr_renderer_begin_buffer_pass change

Artturin:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4159

> ../sway/desktop/output.c:618:47: error: too few arguments to function 'wlr_renderer_begin_buffer_pass'
>   618 |         struct wlr_render_pass *render_pass = wlr_renderer_begin_buffer_pass(
>       |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

encountered in [nixpkgs-wayland](https://github.com/nix-community/nixpkgs-wayland) update job

log showed that people used chase in their messages relating to fixing wlroots changes

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -616,7 +616,7 @@ static int output_repaint_timer_handler(void *data) {
 	}
 
 	struct wlr_render_pass *render_pass = wlr_renderer_begin_buffer_pass(
-		wlr_output->renderer, buffer);
+		wlr_output->renderer, buffer, NULL);
 	if (render_pass == NULL) {
 		wlr_buffer_unlock(buffer);
 		return false;


emersion (approved):
Thanks!

Nefsen402:
> This fixes the build, but the options parameter isn't actually optional — a NULL value will cause sway to abort upon starting the gles2 backend:
> bt full
> 
> Presumably the options were intended to be optional though, so we should probably handle this in wlroots.

There were multiple problems with the code. Passing NULL here was only one of them. https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4173 to the rescue.

PR #7628, opened 2023-06-08
doc: add [--inhibited] to bindsym and bindcode

fshaked:
Also noticed bindcode is missing [--no-repeat], not sure if this is intentional or not.


emersion:
The description includes docs for both of these options.

> Also noticed bindcode is missing [--no-repeat], not sure if this is intentional or not.

Can we fix that too?

Though, maybe we should just get rid of that big list and use `[options...]` instead.

emersion:
Thanks!

PR #7640, opened 2023-06-19
render: Use wlroots scale filter

Nefsen402:
The old solution only worked with gles2. The new built in wlroots solution works with all renderers!


emersion:
We can probably cleanup the GL includes now, and maybe even the GL deps?

Nefsen402:
> We can probably cleanup the GL includes now, and maybe even the GL deps?

Done.

PR #7653, opened 2023-06-26
input: Move wlr_pointer_gestures_v1 to sway_input_manager

markbolhuis:
On multi-seat configurations a zwp_pointer_gestures_v1 global was created for every seat.

Instead, create the global once in the input manager, to be shared across all seats.


emersion (approved):
Good catch, thanks!

PR #7654, opened 2023-06-27
Update README.tr.md

Beryesa:
Fix some falsely put hyphens (there's no such thing here) and common typos, clarify non-native explanations.

@inspectorgadget03 @Tunay14335 :)


Beryesa:
Brought it back, can you review @mt190502 :)

PR #7655, opened 2023-06-27
swaybar: don't set current workspace as not visible

llyyr:
When there's only one workspace open, swaybar will mark it as not visible if the user scrolls on it and eventually incorrectly fail the `active->visible` assert. Fix this by making sure that new and current workspace aren't the same.

### DIFF
--- a/swaybar/input.c
+++ b/swaybar/input.c
@@ -207,7 +207,7 @@ static void workspace_next(struct swaybar *bar, struct swaybar_output *output,
 		}
 	}
 
-	if (new) {
+	if (new && new != active) {
 		ipc_send_workspace_command(bar, new->name);
 
 		// Since we're asking Sway to switch to 'new', it should become visible.


swaybar/input.c
```diff
@@ -207,7 +207,7 @@ static void workspace_next(struct swaybar *bar, struct swaybar_output *output,
 		}
 	}
 
-	if (new) {
+	if (new && new != active) {
```
rpigott, line 210: We could also just swap the order of the assignments within this block, but maybe that's too subtle. Up to you.

rpigott (commented):
I think it would be helpful to mention this bug only affects bars with wrap_scroll yes configured, for browsing. Changes LGTM.

swaybar/input.c
llyyr: Yeah there were various ways to do it but I figured it was best to avoid a pointless ipc command if we can

llyyr:
Changed the commit description to mention wrap_scroll

rpigott (approved):
LGTM, thanks!

PR #7657, opened 2023-06-30
input/seat: don't reset cursor in seat_configure_xcursor()

emersion:
This is handled by wlroots now.

References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4231


emersion:
Hm, just realized this might still be needed if the XCursor theme is updated...

PR #7664, opened 2023-07-08
swaybar: handle wayland-cursor failures

mstoeckl:
This fixes a rare crash when [`wl_cursor_theme_get_cursor`](https://gitlab.freedesktop.org/wayland/wayland/-/blob/6d3334657180f3663b8d83636f4d1bc2097492d2/cursor/wayland-cursor.c#L439-449) fails and returns NULL. This crash will not happen under normal usage -- I suspect the one computer I have for which this happens has a broken set of cursor config files. Still, even in that situation I think it is better to print an error than segfault or suddenly exit.


emersion (approved):
Thanks!

PR #7676, opened 2023-07-16
view: update wlr_toplevel size on client resizes

### LINKED ISSUE #5266: sway resizes floating windows back to their old size
Hi sway developers/community!

While writing a GUI application that resizes its window programmatically, I came across the following... weirdness:

When moving the mouse cursor out of a floating window's area, sway sends a resize event to the window, resizing it back to its original size.

I'm not sure if this is by design or an actual issue.

I have created a minimal reproduction program including a step-by-step guide [in this gist](https://gist.github.com/the-isz/4e1fb03d985c550b2db230b4b52a12c2).

```
$ swaymsg -t get_version
sway version 1.4
```

Thanks for looking into this and for gifting the world (including me) with this brilliant window manager!

Best regards

Dudemanguy:
If a floating client resizes itself, sway updates several of its internal dimensions to match but not wlr_toplevel. This means that the next time wlroots sends a toplevel configure event, it can have wrong coordinates that resize the client back to its old size. To fix this, let's just use wlr_xdg_toplevel_set_size so the wlr_toplevel has the same dimensions as sway. Fixes #5266.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -931,6 +931,11 @@ void view_update_size(struct sway_view *view) {
 	con->pending.content_width = view->geometry.width;
 	con->pending.content_height = view->geometry.height;
 	container_set_geometry_from_content(con);
+
+	// Update the next scheduled width/height so correct coordinates
+	// are sent on the next toplevel configure from wlroots.
+	wlr_xdg_toplevel_set_size(view->wlr_xdg_toplevel, view->geometry.width,
+		view->geometry.height);
 }
 
 void view_center_surface(struct sway_view *view) {


rpigott:
Why not just `wlr_xdg_toplevel_set_size`? Is there a reason to avoid sending a configure here?

emersion:
I'm worried about hacking into wlroots internals like this.

Dudemanguy:
>Why not just wlr_xdg_toplevel_set_size? Is there a reason to avoid sending a configure here?

No reason. I'm just dumb. Changed it to `wlr_xdg_toplevel_set_size` which works just fine.

rpigott (approved):
LGTM.

PR #7681, opened 2023-07-19
sway/commands/output: Add command to set color profile

mstoeckl:
This PR matches the **very WIP** wlroots MR [4280](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4280).

This requires the Vulkan renderer to be useful. (Use `WLR_RENDERER=vulkan` ; setting  `VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation`  to enable validation layers may be useful for testing.)

Example usage:
```
swaymsg output '*' color_profile icc /usr/share/color/icc/colord/ProPhotoRGB.icc
```

Known issues: 
* ~some race condition in which only some frames render correctly; causes flickering~ (fixed)
* ~Only applies with bit depth 10, at the moment~ (fixed)
* ~Having to copy the `struct wlr_color_transform` a number of times is awkward~ (now using reference counting)

### DIFF
--- a/sway/commands/output/color_profile.c
+++ b/sway/commands/output/color_profile.c
@@ -1,0 +1,101 @@
+#include <fcntl.h>
+#include <strings.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <wlr/render/color.h>
+#include "sway/commands.h"
+#include "sway/config.h"
+
+static bool read_file_into_buf(const char *path, void **buf, size_t *size) {
+	/* Why not use fopen/fread directly? glibc will succesfully open directories,
+	 * not just files, and supports seeking on them. Instead, we directly
+	 * work with file descriptors and use the more consistent open/fstat/read. */
+	int fd = open(path, O_RDONLY | O_NOCTTY | O_CLOEXEC);
+	if (fd == -1) {
+		return false;
+	}
+	char *b = NULL;
+	struct stat info;
+	if (fstat(fd, &info) == -1) {
+		goto fail;
+	}
+	// only regular files, to avoid issues with e.g. opening pipes
+	if (!S_ISREG(info.st_mode)) {
+		goto fail;
+	}
+	off_t s = info.st_size;
+	if (s <= 0) {
+		goto fail;
+	}
+	b = calloc(1, s);
+	if (!b) {
+		goto fail;
+	}
+	size_t nread = 0;
+	while (nread < (size_t)s) {
+		size_t to_read = (size_t)s - nread;
+		ssize_t r = read(fd, b + nread, to_read);
+		if ((r == -1 && errno != EINTR) || r == 0) {
+			goto fail;
+		}
+		nread += (size_t)r;
+	}
+	close(fd);
+	*buf = b;
+	*size = (size_t)s;
+	return true; // success
+fail:
+	free(b);
+	close(fd);
+	return false;
+}
+
+struct cmd_results *output_cmd_color_profile(int argc, char **argv) {
+	if (!config->handler_context.output_config) {
+		return cmd_results_new(CMD_FAILURE, "Missing output config");
+	}
+	if (!argc) {
+		return cmd_results_new(CMD_INVALID, "Missing color_profile first argument.");
+	}
+
+	if (strcmp(*argv, "srgb") == 0) {
+		wlr_color_transform_unref(config->handler_context.output_config->color_transform);
+		config->handler_context.output_config->color_transform = NULL;
+		config->handler_context.output_config->set_color_transform = true;
+
+		config->handler_context.leftovers.argc = argc - 1;
+		config->handler_context.leftovers.argv = argv + 1;
+	} else if (strcmp(*argv, "icc") == 0) {
+		if (argc < 2) {
+			return cmd_results_new(CMD_INVALID,
+				"Invalid color profile specification: icc type requires a file");
+		}
+		void *data = NULL;
+		size_t size = 0;
+		if (!read_file_into_buf(argv[1], &data, &size)) {
+			return cmd_results_new(CMD_FAILURE,
+				"Failed to load color profile: could not read ICC file");
+		}
+
+		struct wlr_color_transform *tmp =
+			wlr_color_transform_init_linear_to_icc(data, size);
+		if (!tmp) {
+			free(data);
+			return cmd_results_new(CMD_FAILURE,
+				"Failed to load color profile: failed to initialize transform from ICC");
+		}
+		free(data);
+
+		wlr_color_transform_unref(config->handler_context.output_config->color_transform);
+		config->handler_context.output_config->color_transform = tmp;
+		config->handler_context.output_config->set_color_transform = true;
+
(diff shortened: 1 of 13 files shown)


sway/commands/output/color_profile.c
```diff
@@ -0,0 +1,108 @@
...
+		size_t to_read = (size_t)s  - nread;
+		ssize_t r = read(fd, (void *)(b + nread), to_read);
+		if ((r == -1 && errno != EINTR) || r == 0) {
+			goto fail;
+		}
+		nread += (size_t)r;
+	}
+	*buf = b;
+	*size = (size_t)s;
+	return true; // success
```
emersion, line 47: `fd` is leaked here

sway/commands/output/color_profile.c
```diff
@@ -0,0 +1,108 @@
...
+	if (s <= 0) {
+		goto fail;
+	}
+	b = calloc(1, s);
+	if (!b) {
+		goto fail;
+	}
+	size_t nread = 0;
+	while (nread < (size_t)s) {
+		size_t to_read = (size_t)s  - nread;
```
emersion, line 38: Nit: double space here

sway/commands/output/color_profile.c
```diff
@@ -0,0 +1,108 @@
...
+		ssize_t r = read(fd, (void *)(b + nread), to_read);
```
emersion, line 39: Nit: unnecessary cast to `void *`

emersion (commented):
Mostly minor comments, overall looks good

sway/commands/output/color_profile.c
```diff
@@ -0,0 +1,103 @@
+#define _XOPEN_SOURCE 700
```
emersion, line 1: Can we `#undef _POSIX_C_SOURCE` if we need `_XOPEN_SOURCE`? Otherwise we provide two feature test macros and they potentially conflict (in which case what we get is implementation-defined).

sway/commands/output/color_profile.c
```diff
@@ -0,0 +1,103 @@
...
+		config->handler_context.leftovers.argc = argc - 2;
+		config->handler_context.leftovers.argv = argv + 2;
+	} else {
+		return cmd_results_new(CMD_INVALID,
+			"Invalid color profile specification: first argument should be icc|srgb");
+	}
+
+	return NULL;
+}
+
```
emersion, line 103: Style nit: stray newline it seems?

sway/sway-output.5.scd
```diff
@@ -178,6 +178,18 @@ must be separated by one space. For example:
 	updated to work with different bit depths. This command is experimental,
 	and may be removed or changed in the future.
 
+*output* <name> color_profile srgb|[icc <file>]
+	Sets the color profile for an output. The default is _srgb_. <file> should be a
+	path to a display ICC profile.
+
+	Not all renderers support this feature; currently it only works with the
+	the Vulkan renderer works. Even where supported, the application of the
```
emersion, line 186: Typo: "it only works with the Vulkan renderer."

emersion (commented):
CI is unhappy it seems like, I think we need to add #ifdefs (that's a bit cumbersome) and missing deps in CI manifests?

GreyXor:
Just an idea, would it be reasonable to be able to `unset` a profile as well? By unset, I mean returning to the same state as if I'd never called the color_profile command.

Maybe something like: `swaymsg output '*' color_profile unset`

Riptils:
Hello I've a little Issue...

With a ICC profile (took on NoteBookCheck website) 
...when I launch swaylock through swayidle I have a smal square around the circle, like this : [https://drive.google.com/file/d/1UENBeOQdPVz8q2elbuydGKVteBVeL1C6/view?usp=sharing](url)

And I havent this when launch swaylock just though the terminal.
Thanks for help !

emersion:
> I have not yet made it possible to build this without lcms2 support, and will update this PR eventually to do that.

I've been wondering whether we should always make `wlr_color_transform` available regardless of LCMS availability. I think it depends if we want to base all of our color pipeline on LCMS (Weston-style), or if we want to implement this ourselves.

> when I launch swaylock through swayidle I have a smal square around the circle

Weird, the ICC profile should only affect the final post-blending image. Maybe there is a small difference not noticable without the ICC profile applied, which becomes visible once the ICC profile is applied?

emersion:
Right. AFAIU, Weston plans to translate the CICP stuff into LCMS pipelines (with the float point number transforms that nobody uses in ICC). Other compositors seem to do math in shaders.

GreyXor:
> Hi, I correctly set my icc profile but when swaylock is triggered from swayidle (via `killall -USR1 swayidle` or when the swayidle timetout's reached). I see a little square: image
> 
> this not happening if I directly start `swaylock` from terminal (or if i'm not setting the icc profile). Thanks AMA if I can help with this issue

I have the same issue :

[recording.webm](https://github.com/swaywm/sway/assets/79602273/0b7347e3-93f0-4a72-abc5-b6fdc804a00a)

kennylevinsen:
@GreyXor, @Riptils, do you see the issue if you run sway with `WLR_SCENE_DISABLE_DIRECT_SCANOUT=1`? The color transformation does not apply to direct scanout.

GreyXor:
> @GreyXor, @Riptils, do you see the issue if you run sway with `WLR_SCENE_DISABLE_DIRECT_SCANOUT=1`? The color transformation does not apply to direct scanout.

Thanks, yes I still have the same issue

emersion:
Ref https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4692

Riptils:
Hello ! 
The ICC profile don't work when I use mpv in full-screen mode
My config is in `.config/mpv` on my [dotfiles](https://gitlab.com/Riptils/dotfiles)

emersion:
Now that the wlroots patch has been merged, just need to guard `wlr_color_transform_init_linear_to_icc()`.

emersion:
Oh right, didn't even remember about the patch I wrote.

emersion:
Thanks!

diederikdehaas:
> CI is unhappy it seems like, I think we need to add #ifs (that's a bit cumbersome) and missing deps in CI manifests?

Apologies if this is a dumb question, but:
I see that ``lcms2-dev`` is added to the list of packages for CI, but shouldn't they (then) also be listed as dependencies in ``mason.build`` (also to the v1.10 branch)?

emersion:
These are wlroots dependencies.

diederikdehaas:
Ok, thanks!

zDEFz:
I noticed that if you switch back and forth that the icc profile only gets applied once.
e.g from

`output DP-1 color_profile icc '/usr/share/color/icc/colord/XL2546K-R-120-cdm-D6500.icc' mode 1920x1080@144Hz position 1080 340`
to
`output DP-1 mode 1920x1080@144Hz position 1080 340 mode 1920x1080@144Hz position 1080 340`

is NOT possible.  It will not update unless I exit sway

PR #7690, opened 2023-07-28
Send pointer enter event in seat_set_focus_surface

OkamiW:
swaylock shows cursor after locked, and the cursor disappear after moving it. This is probably unintended behavior.

This commit tries to fix it by sending pointer enter event in seat_set_focus_surface, which in turn is invoked by set_lock_focused_surface


OkamiW:
> Hm I don't think this is correct, the pointer focus (ie, what's under the cursor) is different from the keyboard focus (ie, what gets keyboard events).

Hmm I agree, but it seems sway used to send pointer enter event to the lock process after screen is locked(swaylock is expecting this), but it didn't send this anymore, which lead to the mysterious cursor after locking.

I'm afraid I don't know sway's source code good enough to create a correct patch here, can you help to look into it? Thanks.

emersion:
Maybe we need a `cursor_rebase()` call in `handle_surface_map` in `sway/lock.c`?

OkamiW:
> Maybe we need a `cursor_rebase()` call in `handle_surface_map` in `sway/lock.c`?

This would fix the problem indeed(tested just now). The commit was updated to do this.

emersion (approved):
Thanks!

bitterhalt:
I still have this issue. I am using sway and swaylock from arch repos.

PR #7691, opened 2023-07-29
commands: lift restrictions on toggle all outputs

c31io:
Toggle all outputs on and off based on the state of the first output. Useful in key bindings, because toggling all outputs can't perform concurrently. The delay adds up.

Such command is semantically valid and useful in practice, so the restrictions should be lifted.

### DIFF
--- a/sway/commands/output/power.c
+++ b/sway/commands/output/power.c
@@ -16,8 +16,14 @@ struct cmd_results *output_cmd_power(int argc, char **argv) {
 	if (strcasecmp(argv[0], "toggle") == 0) {
 		const char *oc_name = config->handler_context.output_config->name;
 		if (strcmp(oc_name, "*") == 0) {
-			return cmd_results_new(CMD_INVALID,
-				"Cannot apply toggle to all outputs");
+			for (int i = 0; i < config->output_configs->length; i++) {
+				struct output_config *oc = config->output_configs->items[i];
+				oc->power = !oc->power;
+			}
+
+			config->handler_context.leftovers.argc = argc - 1;
+			config->handler_context.leftovers.argv = argv + 1;
+			return NULL;
 		}
 
 		struct sway_output *sway_output = all_output_by_name_or_id(oc_name);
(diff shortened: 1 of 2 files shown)


c31io:
`output X power toggle, output Y power toggle` is buggy.

`exec swaymsg output X power toggle && swaymsg output Y power toggle` and `output X toggle, output Y toggle` run sequentially, and the delay adds up linearly with the number of displays.

`output * power toggle` and `output * toggle` run blazingly fast but weren't allowed by the target lines.

emersion:
> toggling all outputs can't perform concurrently

This shouldn't result in any difference. Toggling outputs is done sequentially either way: in the client without this PR, in the compositor with this PR. The IPC shouldn't add any meaningful delay.

> Toggle all outputs on and off based on the state of the first output

If we were to merge this, I'd prefer to toggle each output individually: if output A is on and output B is off, the command would turn off A and turn on B.

c31io:
> If we were to merge this, I'd prefer to toggle each output individually

Implemented in the last commit.

emersion:
Why can't we iterate over outputs which match the pattern inside the `output power` command, instead of adding these new fields?

c31io:
Sure we can. No more new fields now.

emersion:
Hm, but now this mutates the current config directly…

kennylevinsen:
> Why can't we iterate over outputs which match the pattern inside the output power command, instead of adding these new fields?

Output commands currently generate a single configuration which is stored and applied, and so cannot cleanly store different on/off states for each output upon toggle.

The options as I see them are:
1. Ignore the existing handling of output configuration (handler_config.output_config). That is what this PR currently does, and is bit naughty.
2. Implement a new flag to signify that the configuration is being toggled, so that handler_context.output_config still works. This is what the PR used to do (https://github.com/swaywm/sway/commit/e239a9aee998ac60e48ce25cf40e0a303bc78857).
3. Let the toggle subcommand queue a new output configuration for each output. This will be more plumbing.

3 might be fanciest, but it's also more work. I'm leaning towards 2 - a bit hacky to have the extra flags, but it works. Thoughts?

c31io:
3 is a kind of over-engineering unless we implement list toggle `output {A,B} toggle` too.

PR #7704, opened 2023-08-12
Make floating_modifier respect shortcuts_inhibitor

ForTheReallys:
floating_modifier is currently triggered when the modifier is pressed,
and when a mouse's left or right button is pressed even if there is an
existing shortcuts_inhibitor.  This changes the default behavior to only
change the current seatop when there is no active shortcuts_inhibitor.
I've added an additional optional `--inhibited` option to the
`floating_modifier` command.  When `--inhibited` is passed, the seatop
is changed regardless of if there is an active shortcuts_inhibitor.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -507,6 +507,7 @@ struct sway_config {
 	struct bar_config *current_bar;
 	uint32_t floating_mod;
 	bool floating_mod_inverse;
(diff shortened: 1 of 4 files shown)


ForTheReallys:
Bump @emersion  @ascent12

ForTheReallys:
Bump @emersion

Issue #6681, opened 2021-11-23
Add render_bit_depth field to get_outputs IPC command


emersion:
Ref https://github.com/swaywm/sway/pull/6475

PR #7708, opened 2023-08-15
Add render_format to get_outputs IPC command

ianhattendorf:
Addresses https://github.com/swaywm/sway/issues/6681

I'm returning the `render_format` instead of `render_bit_depth` as that's what's ultimately used by `wlroots`.

Tested on top of `1.8.1` (I'm running/building on Fedora 38 and didn't want to rebuild all the dependencies) but nothing stands out as incompatible on `master`.

```
% swaymsg output HDMI-A-1 render_bit_depth 8
% ./build/swaymsg/swaymsg -t get_outputs
Output HDMI-A-1 'Guangxi Century Innovation Display Electronics Co., Ltd 27M2V 0000000000000'
  Current mode: 3840x2160 @ 60.000 Hz
  Position: 0,0
  Scale factor: 1.500000
  Scale filter: linear
  Subpixel hinting: unknown
  Render format: XRGB8888
  Transform: normal
  Workspace: 5
  Max render time: off
  Adaptive sync: disabled
  Available modes:
    3840x2160 @ 60.000 Hz
    3840x2160 @ 59.940 Hz
    ...
% swaymsg -t get_outputs | jq -r '.[] | .render_format'
XRGB8888

% swaymsg output HDMI-A-1 render_bit_depth 10
% ./build/swaymsg/swaymsg -t get_outputs
Output HDMI-A-1 'Guangxi Century Innovation Display Electronics Co., Ltd 27M2V 0000000000000'
  Current mode: 3840x2160 @ 60.000 Hz
  Position: 0,0
  Scale factor: 1.500000
  Scale filter: linear
  Subpixel hinting: unknown
  Render format: XRGB2101010
  Transform: normal
  Workspace: 5
  Max render time: off
  Adaptive sync: disabled
  Available modes:
    3840x2160 @ 60.000 Hz
    3840x2160 @ 59.940 Hz
    ...
% swaymsg -t get_outputs | jq -r '.[] | .render_format'
XRGB2101010
```

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -377,6 +377,18 @@ static const uint32_t *bit_depth_preferences[] = {
 	},
 };
 
+int sway_render_format_to_bit_depth(uint32_t render_format) {
+	switch (render_format) {
+		case DRM_FORMAT_XRGB2101010:
+		case DRM_FORMAT_XBGR2101010:
+			return 10;
+		case DRM_FORMAT_XRGB8888:
+			return 8;
+		default:
+			return -1;
+	}
+}
+
 const char *sway_render_format_to_string(uint32_t render_format) {
 	switch (render_format) {
 		case DRM_FORMAT_XRGB2101010:
@@ -389,7 +401,7 @@ const char *sway_render_format_to_string(uint32_t render_format) {
 			return "INVALID";
 		default:
 			return "UNKNOWN";
-		}
+	}
 }
 
 static void queue_output_config(struct output_config *oc,
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -377,6 +377,21 @@ static const uint32_t *bit_depth_preferences[] = {
 	},
 };
 
+const char *sway_render_format_to_string(uint32_t render_format) {
+	switch (render_format) {
+		case DRM_FORMAT_XRGB2101010:
+			return "XRGB2101010";
+		case DRM_FORMAT_XBGR2101010:
+			return "XBGR2101010";
+		case DRM_FORMAT_XRGB8888:
+			return "XRGB8888";
+		case DRM_FORMAT_INVALID:
+			return "INVALID";
+		default:
+			return "UNKNOWN";
+		}
+}
+
 static void queue_output_config(struct output_config *oc,
 		struct sway_output *output, struct wlr_output_state *pending) {
 	if (output == root->fallback_output) {
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -197,4 +197,6 @@ void handle_output_power_manager_set_mode(struct wl_listener *listener,
 
 struct sway_output_non_desktop *output_non_desktop_create(struct wlr_output *wlr_output);
 
+const char *sway_render_format_to_string(uint32_t render_format);
+
 #endif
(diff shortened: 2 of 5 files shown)


sway/config/output.c
```diff
@@ -377,6 +377,21 @@ static const uint32_t *bit_depth_preferences[] = {
 	},
 };
 
+const char *sway_render_format_to_string(uint32_t render_format) {
```
Nefsen402, line 380: I need we need to decide on a more robust way of printing all the formats. There are certainly common formats not filled out in this table. One solution to this problem is introducing a wlroots function to stringify this because it keeps track of all supported formats in a table it internally manages.

Nefsen402 (commented):
I think this MR is a good idea. @emersion what do you think about the introduction of a format stringify function in wlroots?

sway/config/output.c
Nefsen402: Maybe in the future printing the format modifier as well.

Nefsen402: Oh nevermind, libdrm already gives us `drmGetFormatName` and `drmGetFormatModifierName`!

ianhattendorf: Neat, I briefly looked for that but got sidetracked and went with this approach as a quick stopgap.

One question: looks like this was introduced in `drm` just under a year ago, do we need to worry about supporting versions of `drm` prior to its introduction? Or do new sway releases typically require fairly recent `drm` versions?

Nefsen402: No, wlroots uses this funciton. If someone were to try to build sway on these old versions, wlroots as a dependency wouldn't even compile.

emersion: There is no need to try to get this working with every existing format. The only formats which can be picked here are the ones passed in by Sway to wlroots by `wlr_output.render_format`. Sway has a fixed list of bit depth → format combinations, right above.

Also note, the libdrm functions don't return the same string as the `#define`: it returns the four character code. libdrm returns e.g. AR24 for `DRM_FORMAT_ARGB8888`.

I do not know whether this string is a good value to expose.

ianhattendorf: > There is no need to try to get this working with every existing format. The only formats which can be picked here are the ones passed in by Sway to wlroots via wlr_output.render_format. Sway has a fixed list of bit depth → format combinations, right above.

Agreed, but if a generic method exists it would be nice to use.

> Also note, the libdrm functions don't return the same string as the #define: it returns the four character code. libdrm returns e.g. AR24 for DRM_FORMAT_ARGB8888.

I noticed that after I commented, and personally would prefer the more verbose `ARGB8888` vs. `AR24`.

> I do not know whether this string is a good value to expose.

Is this in reference to `AR24` or the render format in general? If the latter, would you prefer just the bit depth or a different value?

emersion: > Is this in reference to AR24 or the render format in general?

A string representing the render format might tie out hands in the future...

ianhattendorf:
@emersion is there any interest in getting this merged? If so, I can resolve conflicts and make sure CI is passing.

emersion:
Sorry, not sure what to do here. `render_bit_depth` we expose as an output command but it turned out to be a bit of a mistake because it doesn't cover stuff such as unsigned ints vs. float for instance.

ianhattendorf:
> Sorry, not sure what to do here. `render_bit_depth` we expose as an output command but it turned out to be a bit of a mistake because it doesn't cover stuff such as unsigned ints vs. float for instance.

Interesting, any plan for moving away from `render_bit_depth` and to something else, like `render_format` or a bit depth/type pairing: `10u`, `16f`, etc.

PR #7723, opened 2023-08-28
xdg_shell: update wlr_toplevel size on client resizes

Dudemanguy:
Reverts #7676 and applies the logic to strictly xdg_shell and not xwayland to avoid some bad bugs there. cc @rpigott

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -294,6 +294,8 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		memcpy(&view->geometry, &new_geo, sizeof(struct wlr_box));
 		if (container_is_floating(view->container)) {
 			view_update_size(view);
+			wlr_xdg_toplevel_set_size(view->wlr_xdg_toplevel, view->geometry.width,
+				view->geometry.height);
 			transaction_commit_dirty_client();
 		} else {
 			view_center_surface(view);
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -931,11 +931,6 @@ void view_update_size(struct sway_view *view) {
 	con->pending.content_width = view->geometry.width;
 	con->pending.content_height = view->geometry.height;
 	container_set_geometry_from_content(con);
-
-	// Update the next scheduled width/height so correct coordinates
-	// are sent on the next toplevel configure from wlroots.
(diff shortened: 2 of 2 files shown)


Dudemanguy:
Ah, the build failures are because freedesktop is currently down.

rpigott:
Err, woops. And thinking a little more about it, sending a configure outside of a transaction seems like it might evoke #5857...

Do you have a reproducer for whatever behavior you observed in #7676? Maybe we can update the transactions to handle it well.

Dudemanguy:
Oh well that [https://gist.github.com/the-isz/4e1fb03d985c550b2db230b4b52a12c2](gist) in the linked issue should work. The way I've been testing is with mpv actually. You can try this [PR](https://github.com/mpv-player/mpv/pull/12260) here. To test the behavior, float the window, press the default binds to scale mpv (such as alt+0 or so), click on some other window to remove focus, then click back on mpv. Without the recent sway commits, it should resize on its own to the old size.

rpigott:
@emersion maybe we need wlroots api for this after all. I think updating the scheduled commit dimensions minus actually sending a configure is basically what we want.

rpigott:
Eh, actually Simon is on vacation atm.

Well I think this is better than the status quo and it appears to work. Let's merge it.

rpigott (approved):

emersion:
Please create a wlroots issue if you think this needs further discussion.

bitraid:
@Dudemanguy This prevents resize by command, eg:
`swaymsg 'for_window [app_id="foot"] floating enable; resize set 100ppt'`

Dudemanguy:
That doesn't work for me on sway 1.8.1 either.

bitraid:
For 1.8.1 you have to add it as a rule to config:
`for_window [app_id="foot"] floating enable; resize set 100ppt`

Dudemanguy:
OK now I see it. I didn't realize that was a newer feature. It's probably because the `wlr_xdg_toplevel_set_size` should not be done until after the view is mapped (or whatever the terminology is). I noticed something similiar in #7730 with fullscreen so this should basically be the same thing.

Dudemanguy:
@bitraid: try #7723 now.

bitraid:
It works!

Issue #7719, opened 2023-08-27
`meson build/` step fails on Debian in master branch, while v1.8 compiles fine


denppa:
# System
Debian Bookworm

# Description
All the dependencies necessary for v1.8 to compile has been installed, and v1.8 compiled successfully. While trying to compile the current HEAD (commit `bf44690ee`) during the `meson build/` phase, this error message pops up:
```
protocols/meson.build:4:0: ERROR: Could not get an internal variable and no default provided for <InternalDependency null: True>
```

Not sure what this means, but it doesn't sound like a dependency problem. Perhaps something is wrong with the `meson.build` file?

bl4ckb0ne:
Sounds like `wayland-protocols` was not found. Do you have the right version?

emersion:
Are you running with a Wayland subproject?

Can you try updating this line with `get_variable('wayland_scanner')`?

https://github.com/swaywm/sway/blob/bf44690ee8ce8438ffb1a84fbe6e8dbb6a851c04/protocols/meson.build#L5

denppa:
@bl4ckb0ne  and @emersion I suspect that `wayland-protocols` is not found because it did not compile correctly due to a missing dependency called `libdisplay-info`.

So I cloned the project into the `subprojects` folder, re-ran `meson build/`, and was greeted with this message:
```bash
wlroots| Dependency wayland-protocols found: YES 1.32 (overridden)
wlroots| Dependency wayland-scanner found: YES 1.22.90 (overridden)
wlroots| Program wayland-scanner found: YES (overridden)
wlroots| Has header "linux/dma-buf.h" : YES
wlroots| Run-time dependency egl found: YES 1.5
wlroots| Run-time dependency gbm found: YES 22.3.6
wlroots| Run-time dependency glesv2 found: YES 3.2
wlroots| Program ./embed.sh found: YES (/home/pilot/rsync/git/sway/subprojects/wlroots/render/gles2/shaders/./embed.sh)
wlroots| Run-time dependency vulkan found: YES 1.3.239
wlroots| Check usable header "vulkan/vulkan.h" with dependency vulkan: YES
wlroots| Program glslang glslangValidator found: NO
wlroots| Dependency pixman-1 found: YES 0.42.2 (cached)
wlroots| Dependency gbm found: YES 22.3.6 (cached)
wlroots| Checking for function "gbm_bo_get_fd_for_plane" with dependency gbm: YES
wlroots| Run-time dependency libudev found: YES 252
[...]

PR #7729, opened 2023-08-30
fixing `meson build/` fail on debian

denppa:
Without this change, the `meson build/` step of compiling sway cannot be completed successfully. This is related to #7719


emersion (approved):
Thanks!

PR #7730, opened 2023-08-30
xdg_shell: don't update wlr_toplevel if the container has no size yet

Dudemanguy:
3d5ae9813d390ea747462fc0026ee43b7c77d0f2 added logic to change the underlying wlr_toplevel size for floating containers, but it does it even when the pending state is fullscreen. This shouldn't be done there because sway's internal size will not be incorrect and there are cases when clients have no idea which output they are going to end up on (like setting fullscreen during initialization) and thus cannot know the correct size in advance.

cc @rpigott: Sorry, I didn't realize that my commit broke this edge case. Easy reproduction would be `mpv --fs video.mkv` and it should be the wrong size if the video size doesn't match your monitor's size. Should affect any other application that initially launches fullscreen as well.

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -294,8 +294,11 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		memcpy(&view->geometry, &new_geo, sizeof(struct wlr_box));
 		if (container_is_floating(view->container)) {
 			view_update_size(view);
-			wlr_xdg_toplevel_set_size(view->wlr_xdg_toplevel, view->geometry.width,
-				view->geometry.height);
+			// Only set the toplevel size the current container actually has a size.
+			if (view->container->current.width) {
+				wlr_xdg_toplevel_set_size(view->wlr_xdg_toplevel, view->geometry.width,
+					view->geometry.height);
+			}
 			transaction_commit_dirty_client();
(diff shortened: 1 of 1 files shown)


Dudemanguy:
I changed it so it doesn't update the toplevel if the current container state has no current size at all. Not sure if there was a better way to check this, but I presumed that `container->current.width` being 0 means it's not on the screen yet. This also fixes what was addressed here: https://github.com/swaywm/sway/pull/7723#issuecomment-1751739996.

rpigott:
Is there anything else required to reproduce? `mpv --fs video.mkv` seems to work alright for me, with a 4k screen and 480p video.

Dudemanguy:
Double checked just now with sway and wlroots from git master and still reproduced it. I honestly don't remember if there have been any mpv changes post 0.36 that would be directly relevant here. I don't *think* so. But even so, problems like in https://github.com/swaywm/sway/pull/7723#issuecomment-1751739996 as a result of that commit still occur.

rpigott (approved):
Ok, well I think the logic is sound anyway. It should be fine to merge. Thanks!

emersion:
Note, I've opened this related wayland-protocols issue: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/199

PR #7737, opened 2023-09-03
sway-ipc.7: typo stacked->stacking

hughobrien:
```
layout string  - The node's layout. It can either  be splith,  splitv, stacked, tabbed, or output
```

Correct to *stacking*

---
Edit: this may just be an inconsistency, as 'stacked' is used in the ipc response here https://github.com/swaywm/sway/blob/89f85312687b7fd1777c936e169e6400cee0a4b4/sway/ipc-json.c#L54

I defer to the experts.

### DIFF
--- a/sway/sway-ipc.7.scd
+++ b/sway/sway-ipc.7.scd
@@ -322,7 +322,7 @@ node and will have the following properties:
 :  Number of pixels used for the border width
 |- layout
 :  string
-:  The node's layout. It can either be _splith_, _splitv_, _stacked_,
+:  The node's layout. It can either be _splith_, _splitv_, _stacking_,
    _tabbed_, or _output_
 |- orientation
 :  string


emersion:
I'm not sure what this PR fixes -- it sounds like Sway actually uses "stacked"? How is this a typo?

hughobrien:
Either it's a typo or a known inconsistency, I thought the former at first since if you consult `sway-ipc` and then run 
```swaymsg layout stacked
Error: Expected 'layout default|tabbed|stacking|splitv|splith' or 'layout toggle [split|all]' or 'layout toggle [split|tabbed|stacking|splitv|splith] [split|tabbed|stacking|splitv|splith]...'
```

you see `stacking` is the preferred term. Perhaps `sway(5)` is the correct man page for that, which does indeed show `layout default|splith|splitv|stacking|tabbed`. 

Do you know if `stacked is in fact a load-bearing spelling?

emersion:
It seems like we're just mirroring i3 here. We need to return "stacked" instead of "stacking" in the IPC response to retain i3 compat.

hughobrien:
Makes sense, thanks for looking into it. I'll close now.

PR #7743, opened 2023-09-10
doc: clarify arguments for `map_from_region`

iffse:
The documentation didn't explain the meaning of the terms 0..1. This PR makes the implication explicit.

Also, adding a commonly used example as in #7724.


emersion (approved):
Thanks!

PR #7748, opened 2023-09-16
Send wl_surface.preferred_buffer_transform

stacyharper:
It looks like this has been forgotten in fc16fb65.

### DIFF
--- a/sway/desktop/surface.c
+++ b/sway/desktop/surface.c
@@ -49,14 +49,17 @@ void handle_compositor_new_surface(struct wl_listener *listener, void *data) {
 
 void surface_update_outputs(struct wlr_surface *surface) {
 	float scale = 1;
+	enum wl_output_transform transform;
 	struct wlr_surface_output *surface_output;
 	wl_list_for_each(surface_output, &surface->current_outputs, link) {
 		if (surface_output->output->scale > scale) {
 			scale = surface_output->output->scale;
+			transform = surface_output->output->transform;
 		}
 	}
 	wlr_fractional_scale_v1_notify_scale(surface, scale);
 	wlr_surface_set_preferred_buffer_scale(surface, ceil(scale));
+	wlr_surface_set_preferred_buffer_transform(surface, transform);
 }
 
 void surface_enter_output(struct wlr_surface *surface,


stacyharper:
I also remark those two preferred buffer values aren't sent on output changes (contrary to the fractional scaling value).

emersion:
I don't believe this is correct. The preferred buffer scale event needs to be sent when we can directly use pre-rotated pixel buffers, but wlroots I'm not sure wlroots' renderer would benefit from this.

emersion:
(Additionally this uses the transform of the output with the highest scale factor, there should be a better logic instead.)

stacyharper:
> (Additionally this uses the transform of the output with the highest scale factor, there should be a better logic instead.)


Okay this obviously doesn't works. Dunno what I had in mind..

But the issue remains: we never notice the preferred_buffer_transform.

Now I feels like it should be handled by wlroots itself, cause it store the primary_output attached to its wlr_scene_surface. But adding this is not enough:

```diff
diff --git a/types/scene/surface.c b/types/scene/surface.c
index 1fe0762b..51c4781c 100644
--- a/types/scene/surface.c
+++ b/types/scene/surface.c
@@ -17,6 +17,8 @@ static void handle_scene_buffer_outputs_update(
        double scale = surface->buffer->primary_output->output->scale;
        wlr_fractional_scale_v1_notify_scale(surface->surface, scale);
        wlr_surface_set_preferred_buffer_scale(surface->surface, ceil(scale));
+       wlr_surface_set_preferred_buffer_transform(surface->surface,
+               surface->buffer->primary_output->output->transform);
 }
 
 static void handle_scene_buffer_output_enter(
```

From the Sway side, we could also re-compute the primary output (if we don't store it somewhere), based on overlapping percents.

> I don't believe this is correct. The preferred buffer scale event needs to be sent when we can directly use pre-rotated pixel buffers, but wlroots I'm not sure wlroots' renderer would benefit from this.

I'm not sure to understand this

emersion:
The event is optional I believe.

kennylevinsen:
> Now I feels like it should be handled by wlroots itself, cause it store the primary_output attached to its wlr_scene_surface. But adding this is not enough: <snippet adding

Hmm, why not? Other than this not affecting sway before the scene PR is merged of course.

>> I don't believe this is correct. The preferred buffer scale event needs to be sent when we can directly use pre-rotated pixel buffers, but wlroots I'm not sure wlroots' renderer would benefit from this.
>
> I'm not sure to understand this

Basically, sending the preferred transform so that the client pre-rotates content only matters if we are trying to scan out the buffer. It does not matter if we are compositing, as our renderers - at least the hardware accelerated ones - would not care about having to rotate a texture as it is being positioned. It might have some effect on the pixman renderer though.

As a future concern, *if* we at some point end up utilizing KMS plane rotation (making the GPU rotate our output buffer on the way out), we would want to subtract the plane rotation - basically only sending a non-0° preferred transform when we could not rotate the plane correctly.

PR #7771, opened 2023-10-06
Clear held keybinding when clicking mouse buttons with modifiers

### LINKED ISSUE #4505: Treat $mod+mouse key same way as $mod+key
I occasionally use mouse in conjunction with $mod button to do certain tasks. Even the stock configuration comes with this functionality in mind:
```
# Drag floating windows by holding down $mod and left mouse button.
    # Resize them with right mouse button + $mod.
```
I'm currently using wldash as my launcher and I have it bound to:
```
set $menu eval "$(dot-desktop "$(dot-desktop | wldash start-or-kill)")"
bindsym --release Super_L exec $menu
```
However, Super_L is also my $mod key.
```
set $mod Mod4
```
The above `exec $menu` does not spawn wldash when I'm using combinations like `$mod+j`.
The issue is that it does spawn it if I initiate a drag or resize operation using mouse.

My guess is that the $mod gets consumed when pressed in combination with a key on a keyboard, but does not when a mouse button is used instead, leading to erroneous triggering of the launcher in my example.

I have used wldash purely as an example above. Same behavior can be observed using any other software (bemenu, krunner, albert to name a few that I've tried).

I'd like to propose the $mod+mouse interaction be made equivalent to $mod+key.

livanh:
When handling a button click with one or more modifiers, or using the mouse with the floating modifier, assume that the modifiers are not intended as a standalone keybinding, and clear the held_binding from the keyboard. This way it is possible to use a single modifier (e.g. Super) as main modifier, floating modifier, and as a standalone keybinding (with --release). Fixes #4505

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -9,6 +9,7 @@
 #include <wlr/util/edges.h>
 #include "sway/config.h"
 #include "sway/input/input-manager.h"
+#include "sway/input/keyboard.h"
 #include "sway/input/tablet.h"
 #include "sway/input/text_input.h"
 
@@ -368,5 +369,8 @@ keyboard_shortcuts_inhibitor_get_for_surface(const struct sway_seat *seat,
  */
 struct sway_keyboard_shortcuts_inhibitor *
 keyboard_shortcuts_inhibitor_get_for_focused_surface(const struct sway_seat *seat);
+
+struct sway_keyboard *sway_keyboard_for_wlr_keyboard(struct sway_seat *seat,
+		struct wlr_keyboard *wlr_keyboard);
 
 #endif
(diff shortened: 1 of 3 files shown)


emersion:
Does this break i3 compat?

goober0013:
Anything blocking this? Would like to see it merged.

emersion:
I'm still not sure this makes sense. `bindsym` is not supposed to interact with pointer events.

livanh:
The interaction consists in "consuming" the floating modifier when it's used to move or resize something with the mouse. If it's not consumed, it could also trigger an unintended bindsym action when released, as described in more detail in #4505.

Do you think that it's a valid use case? If so, could you do it in some other way?

PR #7774, opened 2023-10-10
protocol/action-binder: draft implementation

navi-desu:
adds support for wlr_action_binder_v1 on the wlroots MR https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4386
also adds a new command to trigger actions

currently this branch doesn't store the actions in any way, just passes along to wlroots.

ideally we could store binds so they could be outputted in commands, like `swaymsg -t get_actions` or similar

and then a client daemon using the ipc could possibly automatically creation of bind{sym,code,switch,gesture} based on the suggested trigger. problem being that there's still no format decided for the hint trigger string.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -104,6 +104,7 @@ struct sway_container *container_find_resize_parent(struct sway_container *con,
 sway_cmd cmd_exec_validate;
 sway_cmd cmd_exec_process;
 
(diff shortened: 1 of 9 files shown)


shiona:
Is this dead? Is there any solution that provides the same functionality?

Issue #7370, opened 2023-01-09
Adaptive sync / variable refresh broken when fullscreen


phaserblast:
Adaptive sync seems to be broken when going fullscreen on the Intel modesetting driver. When nothing is fullscreen, adaptive sync works and my FreeSync display reports the correct variable refresh rate. In fact, the display will sync correctly when there's only one thing being rendered via OpenGL in a single panel. However, as soon as an application goes fullscreen, the display changes to its default refresh rate (in my case, 75Hz) and ignores the variable timing. When running on a real X server and not sway, everything works fine so it's not my system.
The only workaround I've found is to launch sway setting the WLR_DRM_NO_ATOMIC=1 variable, however this only fixes apps that are Wayland native (X apps using Xwayland still won't work with adaptive sync even with the workaround). For example, launching MAME using -videodriver wayland works with the above workaround, however launching in X11 mode gets no adaptive sync.

abigrock:
This may be a duplicate of [Issue 5601](https://github.com/swaywm/sway/issues/5601). So far launching sway via `sway -Dnoscanout` or by setting the variable `WLR_DRM_NO_MODIFIERS=1` before launching sway fixes fullscreen adaptive sync with Xwayland applications for me. Both have their own drawbacks however.

phaserblast:
> This may be a duplicate of [Issue 5601](https://github.com/swaywm/sway/issues/5601). So far launching sway via `sway -Dnoscanout` or by setting the variable `WLR_DRM_NO_MODIFIERS=1` before launching sway fixes fullscreen adaptive sync with Xwayland applications for me. Both have their own drawbacks however.

Thanks. Setting WLR_DRM_NO_MODIFIERS=1 worked on my system with Intel graphics. Variable sync seems to be work with both Wayland and X applications. Can you elaborate on the drawbacks of this? I'm not even sure exactly what setting this variable does.

abigrock:
> Thanks. Setting WLR_DRM_NO_MODIFIERS=1 worked on my system with Intel graphics. Variable sync seems to be work with both Wayland and X applications. Can you elaborate on the drawbacks of this? I'm not even sure exactly what setting this variable does.

[...]

Issue #6832, opened 2022-02-14
FreeSync stops working in frame capped games when moving mouse


Elendil211:
sway version 1.7

Config is default with exception of `output $monitor mode 1920x1080@164.995Hz adaptive_sync on`

When I start a game (for example Overwatch, or Witcher 3 - both are running through wine) and I cap the framerate (for example with Overwatchs ingame limiter, MangoHud, or Gamescope), FreeSync is working, which I can see because my monitor displays the current refresh rate. But as soon as I move the mouse, the refresh rate jumps up to 165. Changing fullscreen or windowed mode does not change the behavior. Keyboard input does not have this effect. The effect does not occur when the game is GPU limited.

From what I understand, this behavior would be desired if a hardware cursor is displayed. However, it also occurs when there is no hardware cursor.

This is a problem, because limiting the frame rate in order to make use of FreeSync is desirable to get low input lag. The current behavior makes it impossible to use it.

exp80:
Probably not helpful, but I found this [Reddit post](https://www.reddit.com/r/swaywm/comments/k2xuvh/sway_and_adaptive_sync_moving_cursor_breaks_it/) from a **year ago** which describes the same issue. Has anybody looked into this yet? I am sorry if this comes out as rude and I understand this is an open source project and the contributors are all volunteers, but I have a hard time understanding why this issue is seemingly ignored. As I understand this completely breaks the core functionality of FreeSync in a lot of cases. Isn't this pretty important?

emersion:
Ref https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/44#note_1284481

exp80:
If I understand what you linked correctly, displaying hardware cursors with VRR is still being worked on and the behaviour is not ideal, as it currently always prioritises the cursor. This is an issue, but that should only happen when the hardware cursor is visible right, or am I misunderstanding something?

From this issue:
> From what I understand, this behavior would be desired if a hardware cursor was displayed. However, it also occurs when there is no hardware cursor.

Doesn't that mean that this is a seperate issue or a bug with sway? I also found this [comment](https://www.reddit.com/r/linux_gaming/comments/sohlnp/wayland_moving_mouse_breaks_freesync/hwsw73j/?utm_source=share&utm_medium=web2x&context=1), which seems to be in line with what I think.

emersion:
Oh, I missed the fact that no cursor was involved here.

exp80:
[...]

Issue #5076, opened 2020-03-06
Automatic VRR management


emersion:
With https://github.com/swaywm/sway/pull/5063, users have a way to unconditionally enable VRR. However this can cause some flickering on some monitors. Flickering seems to happen on higher-end monitors which have a larger VRR range. It seems like big enough refresh rate variations will cause flickering.

I think there are two ways to fix this. Both options aren't necessarily exclusive.

### Do like xf86-video-amdgpu: add a way for clients to opt-in

This would mean designing a Wayland protocol to let clients opt-in to VRR. Only clients with a mostly-fixed frame rate would opt-in. This excludes clients such as video players and web browsers, and includes games.

The compositor would only enable VRR if a client is fullscreen and has opted in.

The protocol should probably operate at the `wl_surface` level. We'll want to let Mesa opt-in if possible. Not sure how to let clients override Mesa's decision (this is also an issue with the current X11 implementation).

This is not great, because we miss some use-cases (power savings for web browsers, lowering the refresh rate to the video's frame rate for video players) and it relies on [Mesa's VRR blacklist](https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/src/util/00-mesa-defaults.conf#L403) (I don't really like the blacklist).

### Avoid flickering in the compositor

- Extract the VRR range from the EDID
- Add a configurable maximum refresh rate variation (perform experiment to find a good default value)
- Delay page-flips so that we prevent the refresh rate from varying too much
  - Make the page-flip happen later if a client commits a new buffer too early
  - Force a page-flip if no client commits a new buffer

This is a little bit more involved. However this allows us to take advantage of VRR for power savings and video playback. Also this doesn't require any cooperation with clients or Mesa, the compositor can do everything on its own.

aidanharris:
Does enabling/disabling adaptive sync work via `for_window` criteria (as a temporary measure, I agree with you that a blacklist is probably not the way to go)?

`for_window [app_id="firefox"] adaptive_sync off`?

ascent12:
The second certainly does seem like the more preferable option, if it works in practice.

>Not sure how to let clients override Mesa's decision

Probably would need to be an EGL and Vulkan extension, if Mesa wanted this at all.

emersion:
>Does enabling/disabling adaptive sync work via for_window criteria (as a temporary measure, I agree with you that a blacklist is probably not the way to go)?

[...]

PR #7780, opened 2023-10-16
Use wlr_frame_scheduler

emersion:
Update for https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4307

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -6,10 +6,10 @@
 #include <wayland-server-core.h>
 #include <wlr/config.h>
 #include <wlr/backend/headless.h>
+#include <wlr/interfaces/wlr_frame_scheduler.h>
 #include <wlr/render/swapchain.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_buffer.h>
-#include <wlr/types/wlr_frame_scheduler.h>
 #include <wlr/types/wlr_gamma_control_v1.h>
 #include <wlr/types/wlr_matrix.h>
 #include <wlr/types/wlr_output_layout.h>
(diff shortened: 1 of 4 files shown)


sway/desktop/output.c
```diff
@@ -758,18 +756,12 @@ static void handle_frame(struct wl_listener *listener, void *user_data) {
 	send_frame_done(output, &data);
 }
 
-static void handle_needs_frame(struct wl_listener *listener, void *user_data) {
-	struct sway_output *output =
-		wl_container_of(listener, output, needs_frame);
-	wlr_output_schedule_frame(output->wlr_output);
```
emersion, line 764: Note to self: this bit is now performed by `wlr_frame_scheduler` itself.

GrabbenD:
This is impressive! 🙂 

I know it's still work in progress but I found some interesting takeaways after comparing it back and forth a couple of times. 

In comparison to Sway 1.8.1, here's bugs which are **not** present in the current state of this PR:
- Direct Scanout doesn't have to be disabled to workaround VRR being stuck at output's refresh rate (bug: [#7370](https://github.com/swaywm/sway/issues/7370#issuecomment-1688612523)).
- Mouse movement speed/polling rate doesn't affect the displayed frame rate thus VRR doesn't break when moving the cursor (bug: [#6832](https://github.com/swaywm/sway/issues/6832)).

Differences:
- There's a performance uplift, from 96 FPS to 100 FPS with Halo Infinite (in comparison to Sway 1.9.0-dev [swaywm/sway@4326a26](https://github.com/swaywm/sway/commit/4326a26ad6fb64b8f8de7d32c1e7b6423e1f84f3) + Wlroots 0.17.0-dev [wlroots/wlroots@47bf87ad](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/47bf87ade2bd32395615a385ebde1fefbcdf79a2)).
- Poorly optimized games with high output latency like Halo Infinite are noticeably smoother due to less judder.

Problems
- Cursor pane is only updated when there's new frames being displayed (even with VRR disabled) meaning you can't move the mouse whilst in desktop with no open windows (for reference I'm not using a background/swagbg nor any bar/swaybar)

Setup:
- Sway: updated [emersion/sway@625c7d7](https://github.com/emersion/sway/commit/625c7d7a6c250015a5dd2b2d764ab99208742410)
- Wlroots: updated [rosefromthedead/wlroots@7a4e6861](https://gitlab.freedesktop.org/rosefromthedead/wlroots/-/commit/7a4e6861d9826aba782934f012b052145b2515a4)
- Environment variables: none other than https://aur.archlinux.org/cgit/aur.git/tree/50-systemd-user.conf?h=sway-git
- Mode: 3840x1600 @ 143.998 Hz
- GPU: RX 6800 XT
- Kernel: 6.5.7-2
- Distro: Arch Linux

I was hoping to provide more comprehensive numbers with [GPUVIS](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/docs/gpuvis.rst) but I haven't figured out good filters to make sense of the data.

emersion:
That sounds like unintended consequences. In theory once everything is properly implemented these changes shouldn't change the current behavior at all.

emersion:
This is now feature-complete.

GrabbenD:
Christmas came early this year!
WINE 9.0 with Wayland Vulkan driver and improved scheduler in Sway/Wlroots, can it get any better? 😄

I've been using this for a couple of hours now and I'm very happy with how it works.
Sway was the smoothest Wayland compositor way before this PR and it just keeps on getting better!

## Feedback

### My setup

Specifications:
- Hardware: same as above (https://github.com/swaywm/sway/pull/7780#issuecomment-1775642457)
- Kernel: `6.6.6`
- sway-git: `r7232.92ad6641-1`
- wlroots-git: `1:0.18.0.r6796.ecc134a6-1`
- mesa-git: `24.0.0_devel.181987.9ab59574ef1-1`
- My monitor has OSD which shows the current refresh rate

Kernel arguments:
```bash
# AMD (EEP)
'amd_pstate.shared_mem=1'
'amd_pstate=active'
'amd_prefcore=enable'

# GPU (VRR)
'video=DP-1:3840x1600@144'
'amdgpu.freesync_video=1'

# Performance
'mitigations=off'
'gpu_sched.sched_policy=0' # https://gitlab.freedesktop.org/drm/amd/-/issues/2516#note_2119750
'preempt=full'
```

### What has changed?

In comparison to Sway 1.8.1:

Improvements:
- There's absolutely zero micro-hiccups/micro-stuttering when playing games like Halo Infinite (once the map is fully rendered)!

- VRR finally works with Direct Scanout. I didn't have to disable it to fix frame jitter/stuttering. Reference: https://github.com/swaywm/sway/issues/7370#issuecomment-1688612523

- Fullscreen VRR workaround script is no longer needed to fix mouse sutttering as the desktop is using my full refresh rate (144Hz) instead of falling down to the minimum refresh rate (14Hz) when adaptive sync is enabled (not sure if this was a bug in damage tracking). Reference: https://github.com/swaywm/sway/issues/5076#issuecomment-1753499364

- I no longer need environment variables such as `WLR_DRM_NO_ATOMIC=1` (to prevent refresh rate from being stuck at highest value) and `WLR_DRM_NO_MODIFIERS=1` (to fix screen flickering).

- Cursor movement in game is no longer breaking VRR. Reference: https://github.com/swaywm/sway/issues/6832#issuecomment-1703684151

Noteworthy:
- VSync needs to be activated in game for VRR to even be triggered.

- I can no longer seem to find a way to use VRR when watching fullscreen videos in Chromium based browsers like Brave (this never made a noticeable difference though).


### Problem 1 - new "Failed to page-flip output" (error)

1. Start sway (without any environment variables):
   ```
   $ sway |& tee sway.log
   ```

2. Launch any program (like the `foot` terminal or just `VSCodium`)
   This spams `Failed to page-flip output` error which is repeating every time I move the mouse cursor around:
   ```log
   00:00:00.369 [ERROR] [wlr] [backend/drm/drm.c:807] connector DP-1: Failed to page-flip output: a page-flip is already pending
   00:00:01.195 [ERROR] [wlr] [types/wlr_layer_shell_v1.c:295] A configure is sent to an uninitialized wlr_layer_surface_v1 0x55b34e7b06c0
   00:00:01.196 [ERROR] [wlr] [backend/drm/drm.c:807] connector DP-1: Failed to page-flip output: a page-flip is already pending
   00:00:01.488 [ERROR] [wlr] [backend/drm/drm.c:807] connector DP-1: Failed to page-flip output: a page-flip is already pending
   00:00:03.064 [ERROR] [wlr] [backend/drm/drm.c:807] connector DP-1: Failed to page-flip output: a page-flip is already pending
   00:00:04.329 [ERROR] [wlr] [backend/drm/drm.c:807] connector DP-1: Failed to page-flip output: a page-flip is already pending
   00:00:04.329 [ERROR] [wlr] [backend/drm/drm.c:807] connector DP-1: Failed to page-flip output: a page-flip is already pending
   ```

* It doesn't matter if option `adaptive_sync on` or `adaptive_sync off` is used.

* I couldn't identify a way of stopping this error (e.g.`-Dnoscanout`, `WLR_DRM_NO_ATOMIC=1` nor `WLR_DRM_NO_MODIFIERS=1`).

### Problem 2 - can't toggle VRR on-the-fly (crash)

Toggling VRR causes compositor to crash instantly even though VRR is successfully activated.

* Here's my display config:
   ```ini
   # .config/sway/config
   output DP-1 {
      mode 3840x1600@143.998Hz
      position 0 0
      adaptive_sync off
      max_render_time 1
   }
   ```

* If I execute `$ swaymsg output DP-1 adaptive_sync 1` the compositor will exit instantly (and this is the entire `sway.log`):
   ```
   00:00:00.376 [ERROR] [wlr] [backend/drm/drm.c:807] connector DP-1: Failed to page-flip output: a page-flip is already pending
   00:00:02.380 [ERROR] [wlr] [types/wlr_layer_shell_v1.c:295] A configure is sent to an uninitialized wlr_layer_surface_v1 0x563ed38aec00
   ```

* Same issue happens when using `adaptive_sync on` in config and executing: `$ swaymsg output DP-1 adaptive_sync 0`

### Problem 3 - refresh rate isn't always updated

Monitor continues to use the _latest_ refresh rate instead of using the one specified in the config file when following these steps:

1. Start `$ sway` with `adaptive_sync off` config option.

2. Enable VRR: `$ swaymsg output DP-1 adaptive_sync 1` (which causes a crash even though VRR gets activated).

3. Since only one frame was displayed before the crash, the monitor uses its lowest refresh rate, `14Hz` in TTY/1.

4. When starting `$ sway` the next time, the refresh rate is stuck at `14Hz`.

* This cannot be reproduced if using config option `adaptive_sync on` (as the refresh rate gets updated from `14hz` automatically to `144hz`)

emersion:
Thanks for testing, however I'd like to reiterate that this PR isn't supposed to improve anything. This PR just replicates the existing Sway behavior with new APIs. IOW, it's a bug if this PR doesn't behave exactly as Sway master.

Nogesma:
I have been running wlroots master branch with the wlroots MR, and a modified version of this PR on top of sway master for about a week and it works great.

I am not using the `max_render_time` feature so I am just using a simple patch: https://gist.github.com/Nogesma/5ce59439fee0e8d7fc5454ce47d678f8

I haven't really noticed any major issues, and I can confirm that the bug with VRR not working when using direct scanout is fixed. I do not need to start sway/wlroots with any environment variable for it to work.

I still encounter the issue where moving the mouse cursor will break VRR, but I assume fixing it requires more work than just changing the frame scheduler.

If you still consider the change in behavior a bug, I can try to look into why the new frame scheduler fixes it, although I am not too familiar with the frame scheduling logic nor the drm part. 
This should also probably be addressed in the wlroots MR as most of the frame scheduling code seems to be gone from sway due to the switch to using the scene graph api.

### Specs

Distro: `archlinux`
Kernel: `6.7.9.arch1-1` + [patch to fix VRR for very high refresh rate monitors](https://gitlab.freedesktop.org/drm/amd/-/issues/3203)
Mesa: `1:24.0.3-1`
Sway: master + [patch](https://gist.github.com/Nogesma/5ce59439fee0e8d7fc5454ce47d678f8)
Wlroots: master + [MR](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4307#450e4b104c68fc951ebca8b9497a0293f761d004)

---
GPU: `AMD RX 6700 XT`
Monitor: `AW2725DF` (Display port, `48-360Hz`)

YellowOnion:
@Nogesma mouse movement doesn't actually break VFR, sway cursors are redrawn every time the mouse reports movement, which I assume, you with your 360hz monitor, you also have a gaming mouse doing 500hz/1000hz, if you set your mouse to 125hz polling it should drive the display at ~125hz, Sway treats all updates equally, I wrote a patch for wlr and sway to defer cursor updates to some max latency. if you want to try it.

patch-info: https://github.com/swaywm/sway/issues/6832#issuecomment-1698442549

kennylevinsen:
AMDGPU has some funny quirks where updating cursor position does not seem to drive the display faster and so cursor movements stutter at a lower refresh, but again note that this PR should have *no functional changes*.

sway/desktop/output.c
```diff
@@ -984,14 +1077,17 @@ static void handle_commit(struct wl_listener *listener, void *data) {
...
 		return;
 	}
 
-	output->last_presentation = *output_event->when;
-	output->refresh_nsec = output_event->refresh;
+	// Send frame done to all visible surfaces
+	struct send_frame_done_data frame_done_data = {0};
+	clock_gettime(CLOCK_MONOTONIC, &frame_done_data.when);
+	frame_done_data.msec_until_refresh = get_msec_until_refresh(event);
+	send_frame_done(output, &frame_done_data);
```
carbonXIII, line 1090: It seems strange for the compositor's `handle_present()` to also need to call `get_msec_until_refresh()` in order to determine the frame callback delay. `get_msec_until_refresh()` uses `output->max_render_time`, which is specific to the `timed_frame_scheduler` implemented here. For a generic frame scheduler, we don't seem to expose how long this delay should be, and assuming `output->max_render_time` seems wrong.

- Should frame callback timers be set by the scheduler in its present function instead?
- If it makes more sense to continue setting the callbacks on the compositor side, maybe this is better addressed in the frame scheduler MR?

I know this is on the back-burner, and things have changed a bit in the `wlr_output` events since the last rebase, so feel free to ignore if this no longer makes sense.

kennylevinsen: > It seems strange for the compositor's handle_present() to also need to call get_msec_until_refresh() in order to determine the frame callback delay

Why do you think so? `handle_present` notifies us of when presentation occurred some time after the fact and is not representative of the actual time of presentation. We need to do a bit of math to figure out how much time is actually left at this point, using the reported time of presentation and estimated time from there till the next frame.

> `get_msec_until_refresh()` uses `output->max_render_time`

Hmm? It doesn't in the PR as is. `output->max_render_time` is applied by `timed_frame_scheduler` itself to offset render from the value reported by `get_msec_until_refresh`.

> For a generic frame scheduler, we don't seem to expose how long this delay should be

I believe the intent was to later implement a "fancy" frame scheduler that dynamically decides on a suitable delay, which would more or less deprecate max_render_time.

carbonXIII: > Hmm? It doesn't in the PR as is. output->max_render_time is applied by timed_frame_scheduler itself to offset render from the value reported by get_msec_until_refresh.

Whoops, you're right, `output->max_render_time` is used by `send_frame_done_iterator` not in `get_msec_until_refresh`. That got mixed up at some point between this implementation, master, and my own patch :sweat_smile: 

Same point though, `output->max_render_time` is used outside of the frame scheduler.

> I believe the intent was to later implement a "fancy" frame scheduler that dynamically decides on a suitable delay, which would more or less deprecate max_render_time.

We're on the same page here. There are technically 2 delays here a frame scheduler implementation might want to configure separately. We're using `max_render_time` for both right now (ignoring `view->max_render_time`), but there is:
- A) the delay between present and the frame event (which triggers rendering for the compositor)
- B) the delay between present and the frame callbacks being called (which triggers rendering for clients)

(A) is already being handled entirely within this frame scheduler. My question is about how (B) should be handled by the frame scheduler, since in this implementation it's not.

carbonXIII: Having messed around with implementing a frame scheduler for a bit, I now realize that it doesn't make sense for (B) to be implemented in `wlr_frame_scheduler`. Even though it is very similar to (A), it is probably more appropriate to handle per-surface, rather than per-output.

So I guess this PR doesn't need to address that, sorry for the noise.

PR #7783, opened 2023-10-18
Allow approximate output refresh rate

### LINKED ISSUE #7599: Allow refresh rate flexibility on output modes
An example of various modes reported by a monitor:
- 2560x1440 @ 143.856 Hz
- 2560x1440 @ 119.881 Hz
- 2560x1440 @ 99.946 Hz
- 2560x1440 @ 59.951 Hz

Currently if you want to change the refresh rate, you have to use `swaymsg -t get_outputs` to see the exact refresh rate that is being reported and that you want. This is because "144Hz" won't be accepted, though it is what users likely know about their monitor. It would save people time and it would be more user friendly.

Secondly, different drivers/cards report different refresh rates for the same monitor. My AMD card reports "143.856 Hz", another machine with an Nvidia card reports "143.923 Hz", and a third machine reports "144.006 Hz". If I could enter "144Hz" in my config, the same config can be used for different machines connected to the same monitor.

What do you think of allowing rounded refresh rates to be inputted and let Sway automatically select the mode that, when rounded, would be said value?

taminob:
This PR adds a proposal to solve issue #7599.

It replaces the previous behavior which required an exact match of the output refresh rate by choosing the refresh rate that is closest to the specified target refresh rate (up to a difference of 1Hz). The limit of 1Hz is arbitrarily chosen, other suggestions are also welcome.

Otherwise, if there is no match, the fallback to the highest available refresh rate (for the same resolution) is kept as before.

As discussed in the issue, I also made the logging more verbose to provide additional information about the mode that gets activated in the end.

Additionally, I added the directory `.cache/` to the `.gitignore` which is the default cache directory generated by the `clangd` language server. If you want to have this change in a separate PR or don't want it at all, I can remove it from this PR.

Resolves #7599.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -249,6 +249,8 @@ static void set_mode(struct wlr_output *output, struct wlr_output_state *pending
 	// as (int)(1000 * mHz / 1000.f)
 	// round() the result to avoid any error
 	int mhz = (int)roundf(refresh_rate * 1000);
+	// If no target refresh rate is given, match highest available
+	mhz = mhz <= 0 ? INT_MAX : mhz;
 
 	if (wl_list_empty(&output->modes) || custom) {
 		sway_log(SWAY_DEBUG, "Assigning custom mode to %s", output->name);
@@ -258,23 +260,28 @@ static void set_mode(struct wlr_output *output, struct wlr_output_state *pending
 	}
 
 	struct wlr_output_mode *mode, *best = NULL;
+	int best_diff_mhz = INT_MAX;
 	wl_list_for_each(mode, &output->modes, link) {
 		if (mode->width == width && mode->height == height) {
-			if (mode->refresh == mhz) {
+			int diff_mhz = abs(mode->refresh - mhz);
+			if (diff_mhz < best_diff_mhz) {
+				best_diff_mhz = diff_mhz;
 				best = mode;
-				break;
-			}
-			if (best == NULL || mode->refresh > best->refresh) {
-				best = mode;
+				if (best_diff_mhz == 0) {
+					break;
+				}
 			}
 		}
 	}
-	if (!best) {
-		sway_log(SWAY_ERROR, "Configured mode for %s not available", output->name);
-		sway_log(SWAY_INFO, "Picking preferred mode instead");
-		best = wlr_output_preferred_mode(output);
+	if (best) {
+		sway_log(SWAY_INFO, "Assigning configured mode (%dx%d@%.3fHz) to %s",
+			best->width, best->height, best->refresh / 1000.f, output->name);
 	} else {
-		sway_log(SWAY_DEBUG, "Assigning configured mode to %s", output->name);
+		best = wlr_output_preferred_mode(output);
+		sway_log(SWAY_INFO, "Configured mode (%dx%d@%.3fHz) not available, "
+			"applying preferred mode (%dx%d@%.3fHz)",
+			width, height, refresh_rate,
+			best->width, best->height, best->refresh / 1000.f);
 	}
 	wlr_output_state_set_mode(pending, best);
 }
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ bin/
 test/
 build/
 build-*/
+.cache/
 .lvimrc
 config-debug
 wayland-*-protocol.*


sway/config/output.c
```diff
@@ -257,24 +257,43 @@ static void set_mode(struct wlr_output *output, struct wlr_output_state *pending
...
+					best_diff_mhz = diff_mhz;
+					best = mode;
+					if (best_diff_mhz == 0) {
+						break;
+					}
+				}
 			}
-			if (best == NULL || mode->refresh > best->refresh) {
-				best = mode;
+			if (highest == NULL || mode->refresh > highest->refresh) {
```
kennylevinsen, line 275: I wonder if we should do away with highest. Even if we're more than 1Hz away, picking the closest refresh rate to what the user specified is probably fine. When no refresh rate is specified, we could just set `mhz = INT_MAX`, in which the search for smallest diff also finds the largest value.

sway/config/output.c
```diff
@@ -257,24 +257,43 @@ static void set_mode(struct wlr_output *output, struct wlr_output_state *pending
...
-		sway_log(SWAY_DEBUG, "Assigning configured mode to %s", output->name);
+		sway_log(SWAY_ERROR, "Configured mode (%dx%d@%.3fHz) for %s not available",
+			width, height, mhz / 1000.f, output->name);
+		if (highest) {
+			best = highest;
+			sway_log(SWAY_INFO, "Picking mode (%dx%d@%.3fHz) with highest "
+				"available refresh rate instead", best->width, best->height,
+				best->refresh / 1000.f);
+		}
+		else {
```
kennylevinsen, line 292:
```suggestion
		} else {
```

kennylevinsen (changes requested):
Overall idea looks good!

sway/config/output.c
taminob: Sounds like a good solution to me, will implement it like this.

The only downside I see is that e.g. specifying `90Hz`, when only ~`60Hz` and ~`120Hz` are available, makes it less predictable which will be chosen.
The limit ensures that either around the user-specified or the highest refresh rate will be applied.

sway/config/output.c
```diff
@@ -257,24 +257,43 @@ static void set_mode(struct wlr_output *output, struct wlr_output_state *pending
...
-		sway_log(SWAY_INFO, "Picking preferred mode instead");
-		best = wlr_output_preferred_mode(output);
+	if (best && best_diff_mhz < 1000) {
+		sway_log(SWAY_INFO, "Assigning configured mode (%dx%d@%.3fHz) to %s",
+			best->width, best->height, best->refresh / 1000.f, output->name);
 	} else {
-		sway_log(SWAY_DEBUG, "Assigning configured mode to %s", output->name);
+		sway_log(SWAY_ERROR, "Configured mode (%dx%d@%.3fHz) for %s not available",
+			width, height, mhz / 1000.f, output->name);
+		if (highest) {
```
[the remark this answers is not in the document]
taminob: The reason was to avoid duplicating the log message. No longer relevant since I'll update the messages.

taminob:
I incorporated your feedback.

Just noticed during testing that `set_mode` is (after the first time) never called without a refresh rate since it will be merged with the previous config and thus the already configured refresh rate will be used for e.g. `swaymsg 'output * res 3840x2160`.

So the actual behavior is that, if no refresh rate is specified, the current one will be kept. If there is no current one, the highest will be used.

sway/config/output.c
```diff
@@ -258,23 +260,28 @@ static void set_mode(struct wlr_output *output, struct wlr_output_state *pending
...
-		best = wlr_output_preferred_mode(output);
+	if (best) {
+		sway_log(SWAY_INFO, "Assigning configured mode (%dx%d@%.3fHz) to %s",
+			best->width, best->height, best->refresh / 1000.f, output->name);
 	} else {
-		sway_log(SWAY_DEBUG, "Assigning configured mode to %s", output->name);
+		best = wlr_output_preferred_mode(output);
+		sway_log(SWAY_INFO, "Configured mode (%dx%d@%.3fHz) not available,"
+			"applying preferred mode (%dx%d@%.3fHz)",
+			width, height, refresh_rate / 1000.f,
```
kennylevinsen, line 283: I believe `refresh_rate` is already hertz, no division required.

sway/config/output.c
```diff
@@ -258,23 +260,28 @@ static void set_mode(struct wlr_output *output, struct wlr_output_state *pending
...
-		sway_log(SWAY_INFO, "Picking preferred mode instead");
-		best = wlr_output_preferred_mode(output);
+	if (best) {
+		sway_log(SWAY_INFO, "Assigning configured mode (%dx%d@%.3fHz) to %s",
+			best->width, best->height, best->refresh / 1000.f, output->name);
 	} else {
-		sway_log(SWAY_DEBUG, "Assigning configured mode to %s", output->name);
+		best = wlr_output_preferred_mode(output);
+		sway_log(SWAY_INFO, "Configured mode (%dx%d@%.3fHz) not available,"
+			"applying preferred mode (%dx%d@%.3fHz)",
```
kennylevinsen, line 282: nit: Missing space after comma in this message.

kennylevinsen (commented):
Two minor things, otherwise LGTM!

sway/config/output.c
taminob: good catch :)

kennylevinsen (approved):
LGTM, thanks!

PR #7792, opened 2023-10-24
Use locale time format for default bar command

mattfbacon:


emersion:
While I agree 24-hour format is objectively superior, half the rest of the world doesn't agree, so not sure this is worth changing.

mattfbacon:
Hmm, we could use `d_fmt`/`t_fmt` from the locale with something like `date +"$(locale d_t_fmt)"`.

emersion:
Maybe `%X`?

The full list of POSIX-compatible format options is available at https://pubs.opengroup.org/onlinepubs/009695399/utilities/date.html

mattfbacon:
Sounds good. And keep the fixed date format?

emersion:
We could use `%x`, but I don't have a strong opinion either way.

mattfbacon:
Changed to use %x, please squash to merge

emersion:
Can you rebase and squash?

mattfbacon:
OK

mattfbacon:
Sorry, I forgot we said that. I would prefer `%Y-%m-%d %X` because [ISO8601](https://en.wikipedia.org/wiki/ISO_8601) is pretty universal, and in most locales the ordering will not be purely hierarchical when combined with `%X` (e.g., D/M/Y H:M:S rather than strictly decreasing from Year to Second)

We may also want to consider `%F %X`.

> %F     full date; like %+4Y-%m-%d

Or just `%c`.

>  %c     locale's date and time (e.g., Thu Mar  3 23:05:25 2005)

emersion:
Okay, I don't mind either way.

emersion (approved):
Thanks!

PR #7799, opened 2023-10-29
xdg-shell: send maximized if tiled isn't supported

vyivel:
wlroots doesn't do it automatically anymore.

---

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4409


emersion (approved):
Thanks!

Issue #7173, opened 2022-09-16
Add ability to get current window opacity


acheronfail:
**Description:** `sway` can currently set `opacity` very easily in different ways, e.g., `swaymsg opacity 0.5` or `for_window [<criteria>] opacity 0.75`. But there's no way to _get_ the `opacity` of a window.

I would like to ability to bind a key to 'cycle through opacity values', but I cannot implement this without being able to query the current opacity value for a given window.

**Enhancement Request:** in `swaymsg -t get_tree`'s output, an `opacity` field should be returned with the window's current `opacity` value.

(I couldn't find a non-sway way to get opacity either, through other Wayland tools, etc. If such a way exists it could be a workaround, so please let me know. :slightly_smiling_face:)

bl4ckb0ne:
I don't get the use case, you can just keep the current opacity locally when cycling through it.

> (I couldn't find a non-sway way to get opacity either, through other Wayland tools, etc. If such a way exists it could be a workaround, so please let me know. slightly_smiling_face)

The opacity is a sway feature, Wayland has nothing to do with it.

RPigott:
Related: https://github.com/swaywm/sway/issues/6574#issuecomment-933070655

acheronfail:
> I don't get the use case, you can just keep the current opacity locally when cycling through it.

Maybe I'm missing something? As far as I can tell, it's not possible to have a binding that cycles through opacities... 

As an example, on `i3` (with `picom`) I have something similar to this:

```
bindsym $mod+t exec 'picom-trans --toggle 0.5'
```

Which, when used will toggle the transparency of the focused window.
With a script it's easy with `picom-trans` to get the transparency, and then use that to figure out what the next opacity should be if cycling through many. (This is what I have in `i3`, cycling though `1.0, 0.5, 0.75` opacity values).

What do you mean "keep the current opacity locally"? :thinking:

rickyrsyah:
> > I don't get the use case, you can just keep the current opacity locally when cycling through it.
> 
> Maybe I'm missing something? As far as I can tell, it's not possible to have a binding that cycles through opacities...
> 
> As an example, on `i3` (with `picom`) I have something similar to this:
> 
> ```
> bindsym $mod+t exec 'picom-trans --toggle 0.5'
> ```
> 
[...]

PR #7804, opened 2023-11-02
More options for querying and changing opacity

smason:
Similar to #7173 I wanted a nicer way to get control over opacity of windows.  These changes makes it easy to get the opacity of the current window, e.g., via:
```sh
swaymsg -r opacity get
```
as well as providing more options for altering it via `mul` and `div` subcommands, e.g.,:
```sh
swaymsg opacity mul 1.25
swaymsg opacity div 1.25
```
As per the commit message, human senses tend to work on a log-scale; hence multiplicative changes like this tend to feel smoother than linear changes (i.e. via the existing `plus` and `minus`).

This change doesn't expose the opacity in the `get_tree` output as I didn't need it, but could likely add that in if somebody cares about it.

### DIFF
--- a/sway/commands/opacity.c
+++ b/sway/commands/opacity.c
@@ -17,19 +17,46 @@ struct cmd_results *cmd_opacity(int argc, char **argv) {
 		return cmd_results_new(CMD_FAILURE, "No current container");
 	}
 
+	enum { SET, PLUS, MINUS, MUL, DIV } op = SET;
+	char *argval = argv[0];
+
+	if (argc == 1) {
+		if (!strcasecmp(argv[0], "get")) {
+			return cmd_results_new(CMD_SUCCESS, "%f", con->alpha);
+		}
+	} else {
+		if ((error = checkarg(argc, "opacity", EXPECTED_EQUAL_TO, 2))) {
+			return error;
+		}
+		if (!strcasecmp(argv[0], "set")) {
+			op = SET;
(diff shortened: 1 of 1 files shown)


emersion:
I don't think we'd want more commands to mutate the opacity. These div/mult operations can easily be done by the IPC client, e.g. in a shell.

I don't think a global "get opacity" command is the way to go. Returning the opacity in the get_tree response would probably make more sense.

All that said not sure we want to expand the opacity stuff at all. It doesn't work with sub-surfaces correctly so maybe we will axe it in a future release.

smason:
> All that said not sure we want to expand the opacity stuff at all. It doesn't work with sub-surfaces correctly

What's wrong with this?  Is this why, e.g., menus in firefox become transparent when the whole window does?

> so maybe we will axe it in a future release.

Meh, no great preference.  I did this because I thought it might me nice to have a transparent video player running over my text-editor in the background and got a bit carried away while configuring.  After using it a bit have decided that it doesn't look very good anyway! Ah well, interesting seeing more of the insides of Sway.

4fthawaiian:
> I don't think a global "get opacity" command is the way to go. Returning the opacity in the get_tree response would probably make more sense.
> 

is this something that may be implemented, or are you looking for someone to try their hand? I'd very much like to implement scroll-wheel opacity changing (via keybinds, etc) as seen in some other DE's, and this would be a good first start (tho i would have preferred the solution in this PR)

PR #7805, opened 2023-11-03
swaynag/config.c: fix build against gcc-14 (-Walloc-size)

trofi:
`gcc-14` added a new `-Walloc-size` warning that makes sure that size of an individual element matches size of a pointed type:

        https://gcc.gnu.org/PR71219

`sway` triggers it on `calloc()` calls where member size is used as `1` (instead of member count):

    swaynag/config.c:169:65: error: allocation of insufficient size '1'
     for type 'struct swaynag_button' with size '48' [-Werror=alloc-size]
      169 | struct swaynag_button *button = calloc(sizeof(struct swaynag_button), 1);


emersion (approved):
Thanks

PR #7819, opened 2023-11-15
input/keyboard: Skip configure on same config

kennylevinsen:
sway_keyboard_configure does a lot of keyboard setup, including loading the keymap (possibly twice in case of fallback), serializing both old and new keymap to compare them with a strcmp, and updating state that leads to events for all related clients.

This function is called a lot - output configuration will call it n_outputs x n_keyboard_devices times. Avoid doing work if the configuration for a particular keyboard has not actually changed.

Untested, hence draft. /cc @emersion

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -189,6 +189,8 @@ struct input_config {
 
 	bool capturable;
 	struct wlr_box region;
+
+	uint64_t seq_id;
 };
 
 /**
--- a/sway/config/input.c
+++ b/sway/config/input.c
@@ -6,12 +6,16 @@
 #include "sway/input/keyboard.h"
 #include "log.h"
 
+static uint64_t input_config_sequence = 0;
+
 struct input_config *new_input_config(const char* identifier) {
 	struct input_config *input = calloc(1, sizeof(struct input_config));
 	if (!input) {
 		sway_log(SWAY_DEBUG, "Unable to allocate input config");
 		return NULL;
 	}
+	input->seq_id = input_config_sequence++;
+
 	sway_log(SWAY_DEBUG, "new_input_config(%s)", identifier);
 	if (!(input->identifier = strdup(identifier))) {
 		free(input);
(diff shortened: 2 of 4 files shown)


emersion:
This seems to help a lot! Brings down a no-op `input '*' power on` execution time from 250ms to 40ms.

emersion:
I am not super clear on the lifetime of input config structs though -- is it safe to stash them like this, are we sure we won't compare a stale pointer (maybe re-used for a completely different config later on)?

kennylevinsen:
> I am not super clear on the lifetime of input config structs though -- is it safe to stash them like this, are we sure we won't compare a stale pointer (maybe re-used for a completely different config later on)?

It *seems* like we always reallocate the struct when making config changes, but there is of course some risk that we free and immediately malloc in the same spot.

We could have an input_config sequence number, incremented and assigned whenever we make a new input_config - that should be pretty safe.

include/sway/config.h
```diff
@@ -189,6 +189,8 @@ struct input_config {
 
 	bool capturable;
 	struct wlr_box region;
+
+	int seq_id;
```
emersion, line 193: Just to make it extra safe to not overflow can we make these `uint64_t`?

sway/config/input.c
```diff
@@ -6,12 +6,16 @@
 #include "sway/input/keyboard.h"
 #include "log.h"
 
+int input_config_sequence = 0;
```
emersion, line 9: Nit: this can be `static`

emersion:
Apart from these nits LGTM!

include/sway/config.h
kennylevinsen: Now `uint64_t`, just in case you managed to work your way through 4294967296 input configs... :)

sway/config/input.c
kennylevinsen: True

emersion:
Mirroring the IRC discussion here for posterity: actually I don't think this handles mappings correctly. If an input device has a mapping to a disconnected output, and then that output becomes connected, then this PR would skip the reconfiguration and not map the input device.

kennylevinsen:
Superseded by https://github.com/swaywm/sway/pull/7821

PR #7820, opened 2023-11-16
output: Only apply changes if needed

kennylevinsen:
Minimizes output commits where no change was needed. Handling of custom_mode and render_format is not particularly pretty.

Untested.


emersion:
I had similar patches locally, I posted the here for posterity: https://github.com/swaywm/sway/pull/7822

No particular preference which is better.

PR #7821, opened 2023-11-16
Only reconfigure input mappings on output change

emersion:
Fully reconfiguring all input devices on output change takes a
loooong time. Let's just reconfigure what we need: only mappings
depend on outputs.


sway/config/output.c
```diff
@@ -590,7 +590,7 @@ bool apply_output_config(struct output_config *oc, struct sway_output *output) {
 	// Reconfigure all devices, since input config may have been applied before
 	// this output came online, and some config items (like map_to_output) are
 	// dependent on an output being present.
-	input_manager_configure_all_inputs();
+	input_manager_configure_all_input_mappings();
```
kennylevinsen, line 593: This should also be used in `output_disable`, the only other user of `input_manager_configure_all_inputs`.

kennylevinsen (approved):
LGTM!

PR #7833, opened 2023-11-23
treewide: Sort includes

Nefsen402:
Folks here seem to care about include order a lot. At first I tried using `clang-format` but it results were undesirable.

```
// rules:
// 1. system includes (with <>) should go be sorted first
// 2. protocol inclusions sorted last
// 3. folders are sorted last
// 4. alphabetical order
```

Instead, I wrote a script in nodejs:
```
const fs = require('fs');

const matching = {
	"<": {prior: 0, match: ">"},
	"\"": {prior: 1, match: "\""},
};

const charset = 'utf-8';

for (let i = 2; i < process.argv.length; i++) {
	const file = process.argv[i];
	const contents = fs.readFileSync(file).toString(charset).split('\n');

	let start, end;
	const includes = [];

	for (let ii = 0; ii < contents.length; ii++) {
		if (contents[ii].startsWith("#include ")) {
			if (includes.length === 0) {
				start = ii;
			}

			end = ii + 1;

			let path = contents[ii].substring(9).trim();
			let char = path[0];
			if (path[path.length -1] !== matching[char].match) {
				throw new Error("Found no matching char");
			}
			path = path.substring(1, path.length - 1)

			includes.push({
				char,
				path: path.split('/'), 
				proto: path.endsWith('-protocol.h'),
			});
		} else if (includes.length && contents[ii].trim().length > 0){
			break;
		}
	}

	// ignore duplicate includes
	const mods = includes.filter((include, i) => {
		for (let ii = 0; ii < i; ii++) {
			if (JSON.stringify(include) === JSON.stringify(includes[ii])) return false;
		}
		return true;
	});

	// rules:
	// 1. system includes (with <>) sorted first
	// 2. protocol inclusions sorted last
	// 3. folders sorted last
	// 4. alphabetical order
	mods.sort((a, b) => {
		if (a.char !== b.char) return matching[a.char].prior - matching[b.char].prior;
		// protocols go at the end apparently
		if (a.proto !== b.proto) return a.proto - b.proto;

		for (let i = 0;; i++) {
			if (a.path[i] === b.path[i]) continue;

			if (a.path[i + 1] !== b.path[i + 1]) {
				if (a.path[i + 1] === undefined) return -1;
				if (b.path[i + 1] === undefined) return 1;
			}

			if (a.path[i] < b.path[i]) return -1;
			return 1;
		}
	});


	// only write file if the sort changed
	let changed = mods.length !== includes.length;
	if (!changed) for (let i = 0; i < mods.length; i++) {
		if (JSON.stringify(mods[i]) !== JSON.stringify(includes[i])) {
			changed = true;
			break;
		}
	}

	if (!changed) continue;

	const modified = [
		...contents.slice(0, start),
		...mods.map(include => {
			return `#include ${include.char}${include.path.join('/')}${matching[include.char].match}`
		}),
		...contents.slice(end, contents.length),
	];

	fs.writeFileSync(file, Buffer.from(modified.join('\n'), charset));
}
```


Nefsen402:
Neat, the sorter uncovered a missing include. First commit fixes it.

emersion:
Ref https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4449

PR #7834, opened 2023-11-23
Remove unused render_background_image and move load_background_image to swaybar

mstoeckl:
As swaybg/swaylock are no longer part of this repository, `render_background_image` can be dropped.

Since `load_background_image` is only used by swaybar, move it out of `common/` into `swaybar/` , and give it a better name.


emersion (approved):
LGTM, thanks!

PR #7837, opened 2023-11-24
Update README.de.md to match the EN one

### LINKED ISSUE #7812: German README.de.MD doesn't match README.md
- We are not accepting any new window management features unless they get implemented by i3.

- **Description:**
Please describe in plain English what the enhancement is and what the use case is.

The German README.de.MD doesn't match the English one. I'm guessing it just needs to be updated, but the German one contains information regarding maintaining distro packages which can isn't there on the English README.md

apreiml:
fixes #7812


emersion:
Thanks

PR #7854, opened 2023-12-06
output: Destroy when output layout is destroyed

Nefsen402:
Since output layout is destroyed when the wayland display is destroyed we run into a destroy listener order problem: Either the display starts destroying the outputs first, in which case we're good: The existing handling will clean up. However, things go wrong if the display decides to destroy the output layout first. In this case, sway will hold invalid references to the output layout as part of each output so that when it finally goes to destroy them, sway will dereference destroyed output layout bits.

Ref: https://github.com/swaywm/sway/pull/6844#issuecomment-1843599513

### DIFF
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -39,6 +39,7 @@ struct sway_output {
 
 	struct sway_output_state current;
 
+	struct wl_listener layout_destroy;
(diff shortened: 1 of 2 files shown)


emersion (approved):
Thanks!

PR #7860, opened 2023-12-12
Raise error on non-accessible background file

### LINKED ISSUE #7859: Setting non-existent background image results in success
- **Sway Version:**
  - `1.8.1`
  - `master`

- **Debug Log:**
  - `2023-12-12 14:12:31 - [swaylock-1.7.2/background-image.c:30] Failed to load background image (Failed to open file “/path/to/no/file”: No such file or directory).`

- **Description:**
While working on #7779 I noticed that `swaymsg -r 'output * bg /file/does/not/exist fill'` also delivers success although it will obviously fail:
```
[
  {
    "success": true
  }
]
```
Even though the logging is better than in the other issue, I'd still consider this a bug since the output is wrong.

I already had a look at the code and the cause is probably at https://github.com/swaywm/sway/blob/bf2b79b2845a070d84aafaf95e6cfcf9af7eeb9b/sway/commands/output/background.c#L126 where no error is returned.

taminob:
This changes the output of `swaymsg` to indicate the failure to set the background image:
```
[
  {
    "success": false,
    "parse_error": false,
    "error": "unable to access background file '/path/to/file'"
  }
]
```

This change will also lead to not applying other output options when a command like `swaymsg -r 'output * bg /non/accessible/file scale 1'` is executed.
However, I think that is an improvement in user experience - if they should be run independently of each others success, they could be split into two `swaymsg` calls.

Closes #7859

### DIFF
--- a/sway/commands/output/background.c
+++ b/sway/commands/output/background.c
@@ -123,7 +123,10 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 					src);
 			config_add_swaynag_warning("Unable to access background file '%s'",
 					src);
+			struct cmd_results *result = cmd_results_new(CMD_FAILURE,
+					"unable to access background file '%s'", src);
 			free(src);
+			return result;
 		} else {
 			output->background = src;
 			output->background_option = strdup(mode);


bl4ckb0ne (approved):
LGTM

PR #7861, opened 2023-12-12
Improve proprietary driver detection

emersion:
Use the DRM API instead of `/proc/modules`, and add DisplayLink.

Depends on https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4493

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -163,6 +163,8 @@ struct sway_debug {
 
 extern struct sway_debug debug;
 
+extern bool allow_unsupported_gpu;
+
 bool server_init(struct sway_server *server);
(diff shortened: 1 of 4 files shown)


bl4ckb0ne (approved):
LGTM! 

What's the issue with `evdi` ?

emersion:
See the third commit message. [evdi](https://github.com/DisplayLink/evdi) is an out-of-tree open-source driver developed by the DisplayLink company, which delegates everything to a closed-source user-space daemon. Essentially, it's a roundabout way to write a closed source driver.

bl4ckb0ne:
> but is just some condom for their proprietary user-space driver.

I laughed more than I should've

PR #7870, opened 2023-12-22
Buffer Pooling Optimization

emac07:
Ensure that the get_next_buffer function cycles through the pool and selects the next available buffer for use. Right now, it seems to select the first available buffer without considering if both buffers are busy.

### DIFF
--- a/client/pool-buffer.c
+++ b/client/pool-buffer.c
@@ -103,30 +103,29 @@ void destroy_buffer(struct pool_buffer *buffer) {
 }
 
 struct pool_buffer *get_next_buffer(struct wl_shm *shm,
-		struct pool_buffer pool[static 2], uint32_t width, uint32_t height) {
-	struct pool_buffer *buffer = NULL;
+                                    struct pool_buffer pool[static 2],
+                                    uint32_t width, uint32_t height) {
+    for (size_t i = 0; i < 2; ++i) {
+        if (!pool[i].busy &&
+            (pool[i].width == width && pool[i].height == height)) {
+            pool[i].busy = true;
+            return &pool[i];
+        }
+    }
 
-	for (size_t i = 0; i < 2; ++i) {
-		if (pool[i].busy) {
-			continue;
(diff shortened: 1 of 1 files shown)


client/pool-buffer.c
```diff
@@ -103,30 +103,29 @@ void destroy_buffer(struct pool_buffer *buffer) {
...
-
-	for (size_t i = 0; i < 2; ++i) {
-		if (pool[i].busy) {
-			continue;
-		}
-		buffer = &pool[i];
-	}
-
-	if (!buffer) {
-		return NULL;
```
kennylevinsen, line 117: This return should already handle when both buffers are busy, as buffer then remains NULL.

The only change I can see is when both buffers are available, and the second buffer matches the size while the first does not. Is this the scenario you intended to optimize?

emac07: Thank you for your review! The code snippet isn't specifically tailored to handle scenarios where both buffers are busy. It primarily addresses the condition where buffer remains NULL if unassigned. The optimization aims to improve buffer allocation efficiency, especially focusing on scenarios where both buffers are available but only one matches the required size. If that makes sense.

emac07: I misread your message originally. Yes that is the intended scenario
.

client/pool-buffer.c
```diff
@@ -103,30 +103,29 @@ void destroy_buffer(struct pool_buffer *buffer) {
...
+        if (!pool[i].busy &&
+            (pool[i].width == width && pool[i].height == height)) {
+            pool[i].busy = true;
+            return &pool[i];
+        }
+    }
+
+    for (size_t i = 0; i < 2; ++i) {
+        if (!pool[i].busy) {
+            if (pool[i].width != width || pool[i].height != height) {
```
kennylevinsen, line 118: This condition is always true, as we would have returned from the function during the first loop if a non-busy buffer had matching dimensions.

Issue #7855, opened 2023-12-08
Regression: zwlr_layer_surface_v1.configure event before surface is committed


WhyNotHugo:
### Please fill out the following:
- **Sway Version:**
  - current master, bf2b79b2845a070d84aafaf95e6cfcf9af7eeb9b

- **Debug Log:**
  - https://paste.sr.ht/~whynothugo/239130eff889f2cf99552c695969f989193af76d

- **Configuration File:**
  - `exec shotman --capture output`

Note: I used the following config when generating the log:

```
exec sh -c "shotman --capture window; swaymsg exit"
```

- **Stack Trace:**

No crash

- **Description:**

Sway is now sending a `configure` event to a surface before it has been committed.
This makes [shotman](https://shotman.whynothugo.nl/) crash because it never occurred to me to handle this, but this makes for a good reproduction example.

It looks like `wlroots` prints an error warning about this:

```
00:00:00.085 [wlr] [types/wlr_layer_shell_v1.c:295] A configure is sent to an uninitialized wlr_layer_surface_v1 0x7f332fac0390
00:00:00.085 [wlr] [types/wlr_layer_shell_v1.c:295] A configure is sent to an uninitialized wlr_layer_surface_v1 0x7f332169d6f0
00:00:00.085 [wlr] [types/wlr_layer_shell_v1.c:295] A configure is sent to an uninitialized wlr_layer_surface_v1 0x7f332fac0390
00:00:00.085 [wlr] [types/wlr_layer_shell_v1.c:295] A configure is sent to an uninitialized wlr_layer_surface_v1 0x7f332169d6f0
00:00:00.089 [wlr] [types/wlr_layer_shell_v1.c:295] A configure is sent to an uninitialized wlr_layer_surface_v1 0x7f332169d6f0
```

WhyNotHugo:
I haven't been able to bisect this properly yet. I started and realized halfway that I'd screwed up the associated wlroots version at some point, so have to start over.

rpigott:
I use a series of tags to bisect sway, by tagging sway commits with the wlroots commits they follow then using an alias to find a working wlroots commit for whatever sway commit I land on.

```
$ git config alias.wlrfind
describe --tags --abbrev=0 --match=wlroots/*
$ git wlrfind HEAD                                                                                                                                                                                                                
wlroots/c85838892d56
$ git wlrfind HEAD~20                                                                                                                                                                                                    
wlroots/1c24b1182ba3
```
[...]

PR #7871, opened 2023-12-22
desktop/layer-shell: don't configure uninitialized surfaces

emersion:
Fixes "A configure is sent to an uninitialized wlr_layer_surface_v1" errors.

Closes: https://github.com/swaywm/sway/issues/7855


kennylevinsen:
Hmm, so this skips the erroneous early configure, but I wonder if we're handling the initial layer surface commit correctly.

We only configure in `arrange_layer`, and we only arrange layers when:
1. A layer surface is created.
2. A layer surface is destroyed.
3. A layer surface is committed with a change, or if the mapped state changed.

The first is too early, the second is too late, and the third would not trigger for an empty commit without a buffer. We might need to add a `layer->initialized != layer_surface->initialized` check in `handle_surface_commit`...

PR #7886, opened 2023-12-28
Remove wlr_presentation in sway_server struct

Billli11:
It is no longer in use.


sway/server.c
```diff
@@ -259,9 +259,6 @@ bool server_init(struct sway_server *server) {
 	wl_signal_add(&server->pointer_constraints->events.new_constraint,
 		&server->pointer_constraint);
 
-	server->presentation =
-		wlr_presentation_create(server->wl_display, server->backend);
```
emersion, line 263: We still need to create the global.

emersion (approved):
Thanks!

PR #7888, opened 2023-12-28
ipc: consider all titlebar cases for window_rect

9ary:
Tabbed or stacked containers with more than one child always have a titlebar.

19df2e590602abe0b5e1b53bc11debdb37be3fbe changed `rect` to exclude the titlebar, but overlooked this case, causing my screenshot script to miscalculate the coordinates of the window contents.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -578,9 +578,10 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 	bool visible = view_is_visible(c->view);
 	json_object_object_add(object, "visible", json_object_new_boolean(visible));
 
+	bool has_titlebar = c->title_bar.tree->node.enabled;
 	struct wlr_box window_box = {
 		c->pending.content_x - c->pending.x,
-		(c->current.border == B_PIXEL) ? c->pending.content_y - c->pending.y : 0,
+		has_titlebar ? 0 : c->pending.content_y - c->pending.y,
 		c->pending.content_width,
 		c->pending.content_height
 	};


sway/ipc-json.c
```diff
@@ -578,9 +578,13 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 	bool visible = view_is_visible(c->view);
 	json_object_object_add(object, "visible", json_object_new_boolean(visible));
 
+	enum sway_container_layout parent_layout = container_parent_layout(c);
+	int siblings = container_get_siblings(c)->length;
+	bool has_titlebar = c->current.border != B_PIXEL
+		|| ((parent_layout == L_STACKED || parent_layout == L_TABBED) && siblings > 1);
```
Nefsen402, line 584: Won't stacked/tabbed containers with 1 sibling also have a title bar? Why do we only care if there are more than 1?

9ary: Good catch. I use `hide_edge_borders --i3 none` and forgot about sway's default behavior here.

9ary: Should be taken care of.

sway/ipc-json.c
```diff
@@ -578,9 +578,14 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 	bool visible = view_is_visible(c->view);
 	json_object_object_add(object, "visible", json_object_new_boolean(visible));
 
+	enum sway_container_layout parent_layout = container_parent_layout(c);
+	int siblings = container_get_siblings(c)->length;
+	bool has_titlebar = c->current.border != B_PIXEL
+		|| ((parent_layout == L_STACKED || parent_layout == L_TABBED)
+			&& (!config->hide_lone_tab || siblings > 1));
```
Nefsen402, line 585: I think this can all be replaced with

```
bool has_titlebar = c->title_bar.tree->node.enabled;
```

9ary: That seems to work correctly, though I ran into a regression with `hide_edge_borders --i3` (the window contents are positioned incorrectly).

9ary: #8231

Nefsen402 (approved):

PR #7904, opened 2024-01-09
add --ready-fd to signal readiness

anarcat:
This adds a new commandline option, --ready-fd, to send a notification when sway is ready to run.

This allows a process supervisor to notice sway's startup is complete which allows, for example, the session manager to start Wayland programs in the right order. Without this signal, users have to go through horrible hacks to order services.

For example, I've been using `NotifyAccess=all` in the `sway.service` file and `exec systemd-notify --ready` in my sway config to emulate this. Problem is it's racy and error-prone.

A particularly nasty bug triggered by `NotifyAccess=all` is when `podman` starts and then terminates a container. In that context, `conmon(8)` ends up notifying systemd it's the session master and takes over thee `MainPID` field in systemd. When it dies, systemd believes the session is over and proceeds to kill the entire session. This is explained in more details in:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1039857

This might not actually be the right place to do this in the sway startup sequence, that said. We call `server_run` right after, and maybe somewhere below that would be a better place. But `server_run` only calls `wl_display_run` and that's part of the core wayland library, so that seems a little too far down. I'm not sure Wayland itself is a place to do this, so for now I'm scratching my own itch and doing this in Sway itself.

Note that this approach was taken instead of using the proper `sd_notify` library call, as that approach was refused in #7659. The `--ready-fd` approach was accepted in swaywm/swaylock#281 so it is hoped it will be seen as acceptable here.

An alternative implementation would be to instead check the `NOTIFY_SOCKET` environment variable and use that, if present. That variable is used by systemd and at least the s6 supervisor to receive readiness notifications, so it might be less disruptive.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -201,6 +201,7 @@ static const struct option long_options[] = {
 	{"verbose", no_argument, NULL, 'V'},
 	{"get-socketpath", no_argument, NULL, 'p'},
 	{"unsupported-gpu", no_argument, NULL, 'u'},
+	{"ready-fd", required_argument, NULL, 'R'},
 	{0, 0, 0, 0}
 };
 
@@ -214,17 +215,19 @@ static const char usage[] =
 	"  -v, --version          Show the version number and quit.\n"
 	"  -V, --verbose          Enables more verbose logging.\n"
 	"      --get-socketpath   Gets the IPC socket path and prints it, then exits.\n"
+	"  -R  --ready-fd <fd>    Send readiness notification to the given file descriptor.\n"
 	"\n";
 
 int main(int argc, char **argv) {
 	static bool verbose = false, debug = false, validate = false;
 
 	char *config_path = NULL;
+	int ready_fd;
 
 	int c;
 	while (1) {
 		int option_index = 0;
-		c = getopt_long(argc, argv, "hCdD:vVc:", long_options, &option_index);
+		c = getopt_long(argc, argv, "hCdD:vVc:R:", long_options, &option_index);
 		if (c == -1) {
 			break;
 		}
@@ -264,6 +267,9 @@ int main(int argc, char **argv) {
 				fprintf(stderr, "sway socket not detected.\n");
 				exit(EXIT_FAILURE);
 			}
+			break;
+		case 'R': // --ready-fd
+			ready_fd = strtol(optarg, NULL, 10);
 			break;
 		default:
 			fprintf(stderr, "%s", usage);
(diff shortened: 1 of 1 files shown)


danieldg:
If I were integrating this into my setup scripts, I would want to have run `systemctl import-environment` to pull in `SWAYSOCK` and `WAYLAND_DISPLAY` before marking sway as ready, which is best done via a script that sway `exec`s from the config; this location is still a bit early for that to work (and due to the fact that `exec` in sway is asynchronous, there's actually no way to do this from within sway).

If other things are messing with the notify socket that sway hands them, then you should remove that environment variable from things you `exec`, perhaps by running them with `systemd-run` so they are in their own control groups and not sway's.  That also has advantages if you want to adjust process priorities or do other cgroup things.

Another solution would be to add an `unset-env` command to sway so you can remove `NOTIFY_SOCKET` after using it, but there are other ways to solve that: for example, have the systemd unit run a wrapper that unsets `NOTIFY_SOCKET` and exports it as `SWAY_STARTUP_NOTIFY_SOCKET`, then the post-startup `exec` script can just read that to send the notify instead (and other processes like conmon won't look at that variable).

YHNdnzj:
> A particularly nasty bug triggered by NotifyAccess=all is when podman starts and then terminates a container. In that context, conmon(8) ends up notifying systemd it's the session master and takes over thee MainPID field in systemd. When it dies, systemd believes the session is over and proceeds to kill the entire session. This is explained in more details in:
> 
> https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1039857

You can close down the notify access after signaling readiness, no? (https://github.com/systemd/systemd/pull/26214)

emersion:
Does this overlap with https://github.com/swaywm/sway/pull/8259?

In other words: if socket activation is implemented, is readiness notification support still useful?

danieldg:
If all 3 of the relevant sockets (wayland, X, sway-ipc) can be passed in via socket activation, the only thing I can think of is that you might want to wait for is claiming the ownership of the seat (from logind).  However, I can't think of a well-behaved process that would behave wrong if it got started prior to this happening, so I think it activation would be less useful at that point.  `systemd` supports passing in multiple sockets like this (although you'd need a wrapper script to translate them to command line arguments), so you would just need to add `--xwayland-fd` and `--sway-ipc-fd` to be complete.

kennylevinsen:
> the only thing I can think of is that you might want to wait for is claiming the ownership of the seat (from logind)

Taking control of the seat is done as [one of the first steps](https://github.com/swaywm/sway/blob/c5456be7506adece2cdf922ed6d919db597944ab/sway/server.c#L264) as it is required to be able to test and initialize the backends, whereas creating the Wayland display socket is [one of the last steps](https://github.com/swaywm/sway/blob/c5456be7506adece2cdf922ed6d919db597944ab/sway/server.c#L490).

mathstuf:
I believe that `NOTIFY_SOCKET` should still be unset in the environment to avoid leaking it to any programs `sway` also starts. In fact, it may make sense to have a complete environment modification setup for it so that environment settings for `sway` itself don't leak to spawned programs (e.g., logging or the like).

emersion:
We've talked about this on IRC and concluded that readiness notification is not very useful without passing back `WAYLAND_DISPLAY`, `DISPLAY` and `SWAYSOCK`. Readiness notification needs to happen after dependencies become aware of these env vars, or else dependencies cannot connect to Sway.

sertonix:
I am still hoping that these variable can some day be reliably set by the process calling sway.

emersion:
If the parent process can create these variables, then it also needs to create the sockets, and socket activation can be used instead of readiness notifications.

sertonix:
Right and readiness notification could be done with a small wrapper.

PR #7907, opened 2024-01-12
Extend swap command to support workspaces

fabianscode:
This extends the swap command to also support workspace swapping.
It solves an issue a friend and me had in sway, which is swapping the
contents of two workspaces efficiently. 

The syntax goes as follows:
`swap workspace with [number] <name>`

It will swap the content of the specified workspace with the content of
the currently focused workspace. The currently focused workspace stays
focused. If the other workspace doesnt exist, it will be created,
otherwise if the other workspace is empty afterwards, it will potentially
be destroyed.

It was achieved by simpy swapping relevant fields of the workspaces
and updating a few variables of the affected containers.


bonsaiiV:
Why add an entirely new command? The swap command already exists and intuitively it should work on the content of workspaces too.

To me it feels more like there is an oversight with swap, that needs to be fixed, as it is not working with workspaces.

simonfogliato:
I have been looking for a nice way to swap all windows from one workspace (on monitor one), to a second workspace (on monitor two). The swayr project looks promising but I'd love to see more stuff like this in native sway.

https://sr.ht/~tsdh/swayr/#menu-switchers

At the moment I have been using (swaymsg -t get_outputs) the following...

```
bindsym $mod+m output "HDMI-A-1" toggle
bindsym $mod+Control+Shift+Right move workspace to output right
bindsym $mod+Control+Shift+Left move workspace to output left
bindsym $mod+Control+Shift+Down move workspace to output down
bindsym $mod+Control+Shift+Up move workspace to output up
```

fabianscode:
> Why add an entirely new command? The swap command already exists and intuitively it should work on the content of workspaces too.
> 
> To me it feels more like there is an oversight with swap, that needs to be fixed, as it is not working with workspaces.

I guess that seems reasonable... any other opinions?

cartok:
I have a script on my Todo list as a workaround and think this is really a good feature, especially when working with more than one screen.

fabianscode:
> Why add an entirely new command? The swap command already exists and intuitively it should work on the content of workspaces too.
> 
> To me it feels more like there is an oversight with swap, that needs to be fixed, as it is not working with workspaces.

the latest commit implements exactly that. Any thoughts?

sway/commands/swap.c
```diff
@@ -11,7 +13,162 @@
...
+	cur_ws->gaps_inner = oth_ws->gaps_inner;
+	oth_ws->gaps_inner = cur_ws_cpy.gaps_inner;
+}
+
+static void set_new_focus(struct sway_workspace *ws, struct sway_seat *seat) {
+	if (ws->tiling->length) {
+		// this needs to be more specific (focus not just every container,
+		// but single windows
+		struct sway_container *container = ws->tiling->items[0];
+		struct sway_container *first_view = container_get_first_view(container);
```
bonsaiiV, line 91: Why focus first view? Shouldn't `workspace->focused_inactive_child` be the preferred  way?

bonsaiiV (changes requested):
This swaps the workspaces entirely. I was thinking about just inserting a container and swapping that, in case a workspace is selected to swap.
That's some different functionality.

Overall this approach looks fine. And basic testing showed, that it works.
I like it and think it is a nice addition. Maintainers seem reluctant about adding new stuff though. (to keep i3 compat)

Issues commented should get fixed though.

bonsaiiV:
I just noticed i3/i3#3952. This is what i described in my review. Since the maintainers want sway to be compatible to i3, it might be necessary to implement the feature for i3 first ... .

PR #7909, opened 2024-01-13
layer-shell: don't configure uninitialized surfaces

vyivel:


emersion:
Ref https://github.com/swaywm/sway/pull/7871, although your version seems better as it addresses @kennylevinsen's comment.

sway/desktop/layer_shell.c
```diff
@@ -708,13 +718,4 @@ void handle_layer_shell_surface(struct wl_listener *listener, void *data) {
...
 			&sway_layer->link);
-
-	surface_enter_output(layer_surface->surface, output);
-
-	// Temporarily set the layer's current state to pending
-	// So that we can easily arrange it
-	struct wlr_layer_surface_v1_state old_state = layer_surface->current;
-	layer_surface->current = layer_surface->pending;
-	arrange_layers(output);
-	layer_surface->current = old_state;
```
emersion, line 726: Hm, do you understand why we needed this hack before, and not anymore?

It seems like `arrange_layer()` operates on `current`… But I don't get why we had this workaround before.

vyivel: Honestly, no idea. It has been here since 0c8a64942e087038806b353949c900e03fd764a8 and I don't think it was required even at that time.

In any case, new_surface is now emitted immediately upon the layer surface creation, so we should wait until it has finished its initialization sequence before doing anything with it.

vyivel: Done.

emersion: Hm, I meant just the change to drop the `layer_surface->current` override...

Oh well, doesn't matter.

Issue #5519, opened 2020-07-05
Command "gaps" does not take effect immediately after "swaymsg reload"


whhone:
**Repro steps**:

1. Add or update the `gaps` command in the config file (for example, adding `gaps inner 20`)
2. Run `swaymsg reload`

**Expected**: The `gaps` command should take effect.

**Actual**: It does not change the existing workspaces. Note that the new `gaps` config work on new workspaces.

**Sway version**: It happens on both `1.4` and `master` (1d149230ea1fbe7bc5ac906e235db3c1fdc63afa)

kennylevinsen:
`gaps inner 20`, the `gaps` command that takes exactly 2 arguments, sets the default gaps for newly created workspaces. To set the *current* workspace, use the command that takes exactly 4 arguments: `gaps inner all set 20`.

Both commands and their behavior is documented in the manual (`man 5 sway`).

whhone:
Problem solved. By design,
- `gaps inner 20` is the command used in config file. It sets the **default** value. It won't change the value of existing workspaces
- `gaps inner all set 20` is the command used with `swaymsg`. It changes the value of existing workspaces but it is **runtime only**.

Hence, to change the gaps value for all existing workspaces. Run `swaymsg gaps inner all set 20`.

Issue #4714, opened 2019-11-09
Can't configure gaps to work


boobootat:
When entering commands `swaymsg gaps inner 10`, `swaymsg gaps outer 10` nothing happens. Tried to add these lines in .config/sway/config

`gaps inner|outer|horizontal|vertical|top|right|bottom|left <px>`
`gaps outer all set 20`
`gaps inner all set 20`

After reloading config and/or exiting sway session I have this:
![2019-11-09-164453_grim](https://user-images.githubusercontent.com/45947948/68531219-61006f80-0310-11ea-9ae9-468ded7691cd.png)

rdrey:
The syntax inside the config file is much simpler than the swaymsg syntax. (There is no `all` since it will apply to all future workspaces and there is no `set` either. This works in my config file `gaps inner 8`. Good luck

telamon:
Also worth mentioning since I had similar issue just a day ago.. 
The new  gap settings apply only to new workspaces. existing gaps are not retroactively updated.
So in order to get any kind of visual confirmation you need to reload your sway-conf and move your windows to an uninhabited workspace. (or reboot/log out)

RedSoxFan:
`gaps inner|outer|horizontal|vertical|top|right|bottom|left <px>` changes the default gaps for future workspaces.  This can be used in the config file or at runtime, but will not effect existing workspaces

`gaps inner|outer|horizontal|vertical|top|right|bottom|left all|current set|plus|minus <px>` changes the gaps for existing workspaces, but will not effect future workspaces. Since this affects existing workspaces, it can only be used at runtime

boobootat:
> This works in my config file: `gaps inner 8`.

Thanks a lot, it helped me.It's kinda funny that the solution was so simple. Loving sway so far

boobootat:
Thank you so much for your help <3

rileyrg:
Is there another resource for this as "man 5 sway" doesn't mention it?

Ramsfield:
> Is there another resource for this as "man 5 sway" doesn't mention it?

`man 5 sway` does have it, it's under the Commands section.
Here's the excerpt:
```
gaps inner|outer|horizontal|vertical|top|right|bottom|left all|current
       set|plus|minus <amount>
           Changes the inner or outer gaps for either all workspaces or the cur-
           rent workspace. outer gaps can be altered per side with top, right,
           bottom, and left or per direction with horizontal and vertical.
```
Cheers

rileyrg:
[...]

PR #7911, opened 2024-01-16
Update sway.5.scd - gaps

junelva:
Might help issues such as #4714 #5519.


bonsaiiV (changes requested):
This is confusing. I don't think it is helpful to explain how the configuration works in the section, where non-config commands are explained.

I think it is better to refer to the later explanation or delete this section and explain the additional non-config-syntax with the other part.

PR #7922, opened 2024-01-21
layer_shell: Fix typo of return instead of continue

Nefsen402:
Otherwise we would skip arranging the rest of the surfaces if one of them isn't initialized.


emersion (approved):
Thanks!

cc @vyivel

PR #7925, opened 2024-01-22
layer-shell: handle commits from uninitialized surfaces

vyivel:
Fixes: 2c69e19fd30c5ceac61d6af821c6c6875cd3f8b6


emersion:
Why should we handle commits for uninitialized surfaces?

vyivel:
In the latest release, the exclusive zone seems to be applied when the surface is created and destroyed; i.e. if there's a delay between creating and mapping, or unmapping and destroying, the user will see an empty gap. On latest commit, the exclusive zone is applied only when the surface is mapped (because of a corresponding check in the layer-shell scene helper), but if a surface commits a NULL buffer, the output isn't rearranged.

…we could just rearrange the output in the surface unmap handler, actually.

PR #7927, opened 2024-01-23
input: Propagate modifiers to clients with only pointer focus

danieldg:
When focus_follows_mouse is off, it is possible to hover and/or scroll windows without giving them keyboard focus.  However, attempts to use Ctrl+scroll on such clients will not work correctly because the keyboard modifiers are not available.  This patch propagates modifier state to the client with pointer focus ~~when it gains focus and when the state changes~~ before sending axis events.

https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/259 discusses permitting this behavior at the protocol level.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -795,6 +795,8 @@ static void handle_pointer_axis(struct sway_seat *seat,
 	free(dev_id);
 
 	if (!handled) {
+		wlr_seat_client_notify_modifiers(cursor->seat->wlr_seat,
+			cursor->seat->wlr_seat->pointer_state.focused_client);
 		wlr_seat_pointer_notify_axis(cursor->seat->wlr_seat, event->time_msec,
 			event->orientation, scroll_factor * event->delta, 
 			roundf(scroll_factor * event->delta_discrete), event->source,


sway/input/seat.c
```diff
@@ -504,6 +504,24 @@ static void collect_focus_container_iter(struct sway_container *container,
...
+
+	struct wl_client *kbd_client = kbd_focus ? kbd_focus->client : NULL;
+	struct wl_client *ptr_client = ptr_focus ? ptr_focus->client : NULL;
+
+	// Don't send modifiers if the client already has keyboard focus
+	if (kbd_client == ptr_client)
+		return;
+
+	// Switch keyboard focus temporarily to send the modifiers to the pointer's client
+	wlr_seat->keyboard_state.focused_client = ptr_focus;
```
emersion, line 520: Hm, would be nicer to find a solution for this in wlroots…

danieldg: The first idea I have there is to copy `wlr_seat_keyboard_send_modifiers` as `wlr_seat_keyboard_send_modifiers_raw` with an extra `struct wlr_seat_client *` argument (removing its first line). However, it might be nicer to put the whole `seat_sync_modifiers` function there (maybe called `wlr_seat_keyboard_send_modifiers_to_pointer`?) - although that's likely to be less flexible if we want to extend this logic to touch or tablet.  Any preference?

emersion: How about a `wlr_seat_client_notify_modifiers()` function which is a no-op if the client has keyboard focus, and a no-op if the last notified modifier is the same as the new one?

I wonder if we should send modifiers always when the surface has pointer focus, or just before a pointer button…

danieldg: Tracking the last-sent keyboard modifiers per-client seems like a lot of additional work for little gain - you'd have to track it for each client separately, including cleaning up when that client actually gets focus or exits.

Only sending just before a button/axis event is also possible, but if a client supports actions like changing the cursor in response to a modifier then sending it even without a press would be useful.  That's normally done for drag/drop, but I recall seeing a resize cursor appear when holding a modifier before - though I don't know what application it was.

danieldg: The latest version tracks the client-known modifiers in wlroots, which means it's now easy to only send on axis events.  I haven't found anything that would use modifiers with just hover, but if it exists it's pretty easy to add that too.

PR #7929, opened 2024-01-23
Add a common popup management struct

Nefsen402:
We had this `SWAY_SCENE_DESC_POPUP` type floating around that would only handle popup arrangement for xdg_shell. This was originally created to have popups follow the location of views as they move.  I had the same concern for layer_shell, but I only moved popups for a layer shell when that layer shell has been committed. This is wrong because if a layer_shell commits, we could potentially arrange other layer shells to move and those popups wouldn't move as well. This system ties it all together into a common system. 

This is also a soft requirement for input_method popups as they have a similar concern.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -612,13 +612,9 @@ void arrange_popups(struct wlr_scene_tree *popups) {
 		struct sway_popup_desc *popup = scene_descriptor_try_get(node,
 			SWAY_SCENE_DESC_POPUP);
 
-		// the popup layer may have popups from layer_shell surfaces, in this
-		// case those don't have a scene descriptor, so lets skip those here.
-		if (popup) {
-			int lx, ly;
-			wlr_scene_node_coords(popup->relative, &lx, &ly);
-			wlr_scene_node_set_position(node, lx, ly);
-		}
+		int lx, ly;
+		wlr_scene_node_coords(popup->relative, &lx, &ly);
+		wlr_scene_node_set_position(node, lx, ly);
 	}
 }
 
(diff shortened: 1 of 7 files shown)


Nefsen402:
If a later offset is needed, that can just be added to the struct as part of your PR or by just nesting more scene nodes.

sway/desktop/transaction.c
```diff
@@ -606,21 +606,15 @@ static void arrange_output(struct sway_output *output, int width, int height) {
...
-		// the popup layer may have popups from layer_shell surfaces, in this
-		// case those don't have a scene descriptor, so lets skip those here.
-		if (popup) {
-			struct wlr_scene_tree *tree = popup->view->content_tree;
-
-			int lx, ly;
-			wlr_scene_node_coords(&tree->node, &lx, &ly);
-			wlr_scene_node_set_position(&popup->scene_tree->node, lx, ly);
-		}
+		int lx, ly;
```
Decodetalkers, line 615: I think this should contain a part for offset, maybe in `sway_popup_desc`, add a field of 

* bool has_offset
* int x_offset
* int y_offset

emm, what do you thought about?

Nefsen402: For input_method popups, I think it would just be easier to nest an extra scene tree. The position of the popup will be managed by this code here, and then you can apply your own offset to the nested tree. This can make things easier because if you want to change the offset later, you don't need to worry about clobbering the original position.

Decodetalkers: I see, I should create a subsurface_tree of the view, set the position, and use this node as the relative

Decodetalkers (approved):
This pr will help a lot, hope it will be merged soon

vyivel (approved):
Sorry for the delay. This LGTM, thanks!

PR #7951, opened 2024-02-03
sway/config.c: only reset primary_selection at launch

vimproved:
Otherwise, an error will be shown whenever reloading due to the value of primary_selection being reset to true.

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -475,6 +475,11 @@ bool load_main_config(const char *file, bool is_active, bool validating) {
 				old_config->xwayland ? "enabled" : "disabled");
 		config->xwayland = old_config->xwayland;
 
+		// primary_selection can only be enabled/disabled at launch
+		sway_log(SWAY_DEBUG, "primary_selection will remain %s",
+				old_config->primary_selection ? "enabled" : "disabled");
+		config->primary_selection = old_config->primary_selection;
+
 		if (!config->validating) {
 			if (old_config->swaybg_client != NULL) {
 				wl_client_destroy(old_config->swaybg_client);


emersion:
Thanks for sending a patch!

Eh, I didn't consider that. I suppose the `xwayland` command has the same defect as well.

It also appears to me that the command won't complain if invoked from `swaymsg`, which seems wrong.

Hm, I wonder if there's a better way to fix it. ~~Instead of trying to read the current config value, can we check whether primary selection is currently enabled? e.g. store the return value of `wlr_primary_selection_v1_device_manager_create()` in `struct sway_server`, and check `server.primary_selection` to see if it's currently enabled.~~

Actually scratch that, I just noticed that there is `config->xwayland = old_config->xwayland;` up above for the xwayland case. Does mirroring this for primary selection fix it?

emersion:
> It also appears to me that the command won't complain if invoked from swaymsg, which seems wrong.

Maybe one way to fix it is to check for `config->active` instead of `config->reloading`?

vimproved:
> Actually scratch that, I just noticed that there is `config->xwayland = old_config->xwayland;` up above for the xwayland case. Does mirroring this for primary selection fix it?

Oh, I hadn't noticed that before. I'll try mirroring that real quick and see if it fixes the issue.

vimproved:
Yep, that works as well. I'll update this.

emersion (approved):
Thanks!

emersion:
> It also appears to me that the command won't complain if invoked from swaymsg, which seems wrong.

https://github.com/swaywm/sway/pull/7954

PR #7952, opened 2024-02-04
Add the DesktopNames key to the sway.desktop session file

nilninull:
According to
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html, to set the value of XDG_CURRENT_DESKTOP, a DesktopNames key is required in the session file.  And the value of XDG_CURRENT_DESKTOP is used to run xdg-desktop-portal*. 

If this value is not set, xdg-desktop-portal-wlr will not run at login.

The value of XDG_CURRENT_DESKTOP is set by the login manager, but from my research, some of them are supported and some are not.

In sddm, I confirmed that it is supported.

### DIFF
--- a/sway.desktop
+++ b/sway.desktop
@@ -3,3 +3,4 @@ Name=Sway
 Comment=An i3-compatible Wayland compositor
 Exec=sway
 Type=Application
+DesktopNames=sway;wlroots


kennylevinsen:
Duplicate of:
- https://github.com/swaywm/sway/pull/6227
- https://github.com/swaywm/sway/pull/5607
- https://github.com/swaywm/sway/pull/4876

With the new portal configuration system I'm more open to this. Clients that use this for anything is still horrible, but the .desktop file doesn't change anything there seeing that pretty much everyone runs portals these days and therefore has XDG_CURRENT_DESKTOP set regardless...

emersion:
Agreed.

nilninull:
OK

emersion:
We've just said that we are "more open to this", which means we are open to the idea!

DarkDefender:
Is there anything holding this up?

Vladimir-csp:
+1

emersion (approved):
I think the benefits of having xdg-desktop-portal just work on login managers using .desktop files outweigh our concerns about clients adding compositor-specific hacks.

@kennylevinsen ACKed this on IRC.

PR #7954, opened 2024-02-05
commands: make primary_selection a config-only command

emersion:
It's not possible to switch this at runtime.


kennylevinsen (approved):
LGTM

PR #7955, opened 2024-02-05
Workspace criterion also applies to containers

### LINKED ISSUE #7218: workspace=__focused__ criterion is ignored for non-leaf containers on focus command
- **Sway Version:** 1.7

- **Configuration File:**

Default config plus this:

```
bindsym $mod+1 [workspace=__focused__ con_mark="x"] focus
bindsym $mod+Ctrl+1 mark --toggle x
```

- **Description:**

The first keybindings above should focus a container with the mark "x", but only if it is on the current workspace.
This only works for me, if the container is a leaf (i.e. application window). But if I put the mark on a container higher up in the tree, the `workspace=__focused__` restriction is ignored. Then the container will be focused (and the workspace changed) anyway, despite a different workspace being active/shown.

bonsaiiV:
The workspace criterion did only apply to views, now it also applies to containers.

Fixes #7218

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -159,6 +159,31 @@ static bool has_container_criteria(struct criteria *criteria) {
 
 static bool criteria_matches_container(struct criteria *criteria,
 		struct sway_container *container) {
+	struct sway_seat *seat = input_manager_current_seat();
+	struct sway_container *focus = seat_get_focused_container(seat);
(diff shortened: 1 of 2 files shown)


emersion:
Is this consistent with i3 behavior?

bonsaiiV:
I don't know how i3 does this. I don't use i3 and do not understand how it handles criteria and the documentation i found is not precise.

PR #7961, opened 2024-02-07
reduces redundant containers

### LINKED ISSUE #7945: Avoid creating redundant containers
- **i3 PR:** https://github.com/i3/i3/pull/5469

- **Description:**
  - With https://github.com/i3/i3/commit/f8befe378ab3c9f8d035657ca0247c9b8c242b6b, WM should not create this insanity by an accident:
 
  redundant

bonsaiiV:
Applying layout changes to the parent of the parent, in case the parent only has a single child, stops the creation of a chain of single child containers. 

fixes #7945

### DIFF
--- a/sway/commands/layout.c
+++ b/sway/commands/layout.c
@@ -134,8 +134,14 @@ struct cmd_results *cmd_layout(int argc, char **argv) {
 	// Operate on parent container, like i3.
 	if (container) {
 		container = container->pending.parent;
+		// If parent has only a singe child operate on its parent and
+		// flatten once, like i3
 		if (container && container->pending.children->length == 1) {
-			container = container->pending.parent;
+			struct sway_container *child = container->pending.children->items[0];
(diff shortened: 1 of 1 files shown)


MartenBE:
Is there any update on this PR?

bonsaiiV:
I think that it is done, and working. If @emersion agrees, this could be merged.

wildwestrom:
If this does what I think it does, then I'd love to see this merged!

emersion (commented):
Could we use `container_flatten()` instead of this? I believe it's functionally equivalent.

bonsaiiV:
> Could we use `container_flatten()` instead of this? I believe it's functionally equivalent.

The difference is, that `container_flatten()` continues to flatten on the parent container (while vs. if). In i3 it is explicitly avoided to create new 1-child-containers, while preserving existing 1-child-containers.

emersion (approved):
Oh right. This LGTM then, thanks!

PR #7962, opened 2024-02-07
launcher: track the seat in the launcher ctx

rpigott:
This is a more suitable place to track the requesting seat, since we are able to respond appropriately to destroy notifications.

---

This is an alternative to #7959.

### DIFF
--- a/sway/desktop/launcher.c
+++ b/sway/desktop/launcher.c
@@ -67,6 +67,9 @@ void launcher_ctx_destroy(struct launcher_ctx *ctx) {
 	}
 	wl_list_remove(&ctx->node_destroy.link);
 	wl_list_remove(&ctx->token_destroy.link);
+	if (ctx->seat) {
+		wl_list_remove(&ctx->seat_destroy.link);
+	}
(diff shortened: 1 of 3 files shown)


sway/desktop/launcher.c
```diff
@@ -238,13 +243,15 @@ struct launcher_ctx *launcher_ctx_create_internal(void) {
...
-	token->seat = seat->wlr_seat;
 
 	struct launcher_ctx *ctx = launcher_ctx_create(token, &ws->node);
 	if (!ctx) {
 		wlr_xdg_activation_token_v1_destroy(token);
 		return NULL;
 	}
+	ctx->seat = seat;
+	ctx->seat_destroy.notify = launch_ctx_handle_seat_destroy;
+	wl_signal_add(&seat->wlr_seat->events.destroy, &ctx->seat_destroy);
```
emersion, line 254: We need to unregister this listener when the launcher ctx is destroyed or when the seat is destroyed.

rpigott: Err, yeah.

PR #7963, opened 2024-02-08
config: use format_str() instead of hand-rolled snprintf()

emersion:


sway/config.c
```diff
@@ -368,10 +362,7 @@ static char *get_config_path(void) {
 
 	const char *config_home = getenv("XDG_CONFIG_HOME");
 	if ((config_home == NULL || config_home[0] == '\0') && home != NULL) {
-		size_t size_fallback = 1 + strlen(home) + strlen("/.config");
-		config_home_fallback = calloc(size_fallback, sizeof(char));
-		if (config_home_fallback != NULL)
-			snprintf(config_home_fallback, size_fallback, "%s/.config", home);
+		config_home_fallback = format_str("%s/.config", home);
 		config_home = config_home_fallback;
```
bl4ckb0ne, line 366: `config_home_fallback` could go

```
config_home = format_str("%s/.config", home);
```

emersion: Unfortunately no, we still need to free that string.

bl4ckb0ne: Indeed, forgot about that.

bl4ckb0ne (approved):
LGTM!

PR #7971, opened 2024-02-11
ext-foreign-toplevel-list: Implement protocol

columbarius:
Requires: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4543

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -582,6 +582,14 @@ static struct sway_workspace *select_workspace(struct sway_view *view) {
 	return NULL;
 }
 
+static void update_ext_foreign_toplevel(struct sway_view *view) {
+	struct wlr_ext_foreign_toplevel_handle_v1_state toplevel_state = {
+		.app_id = view_get_app_id(view),
+		.title = view_get_title(view),
+	};
+	wlr_ext_foreign_toplevel_handle_v1_update_state(view->ext_foreign_toplevel, &toplevel_state);
+}
+
 static bool should_focus(struct sway_view *view) {
 	struct sway_seat *seat = input_manager_current_seat();
 	struct sway_container *prev_con = seat_get_focused_container(seat);
@@ -751,6 +759,15 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 		}
 	}
 
+	struct wlr_ext_foreign_toplevel_handle_v1_state foreign_toplevel_state = {
+		.app_id = view_get_app_id(view),
+		.title = view_get_title(view),
+	};
+	view->ext_foreign_toplevel =
+		wlr_ext_foreign_toplevel_handle_v1_create(server.foreign_toplevel_list, &foreign_toplevel_state);
+	wl_signal_add(&view->ext_foreign_toplevel->events.destroy,
+			&view->ext_foreign_destroy);
+
 	view->foreign_toplevel =
 		wlr_foreign_toplevel_handle_v1_create(server.foreign_toplevel_manager);
 	view->foreign_activate_request.notify = handle_foreign_activate_request;
@@ -828,6 +845,10 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 		input_manager_set_focus(&view->container->node);
 	}
 
+	if (view->ext_foreign_toplevel) {
+		update_ext_foreign_toplevel(view);
+	}
+
 	const char *app_id;
 	const char *class;
 	if ((app_id = view_get_app_id(view)) != NULL) {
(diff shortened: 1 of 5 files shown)


sway/tree/view.c
```diff
@@ -828,6 +834,10 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 		input_manager_set_focus(&view->container->node);
 	}
 
+	foreign_toplevel_state.app_id = (char *)view_get_app_id(view);
+	foreign_toplevel_state.title = (char *)view_get_class(view);
```
columbarius, line 838: Is this ok, or should the wlroots struct use the const qualifier?

emersion: IMHO would be better if the wlroots struct used the const qualifier

sway/tree/view.c
```diff
@@ -751,6 +751,12 @@ void view_map(struct sway_view *view, struct wlr_surface *wlr_surface,
 		}
 	}
 
+	struct wlr_ext_foreign_toplevel_handle_v1_state foreign_toplevel_state = {0};
+	view->ext_foreign_toplevel =
+		wlr_ext_foreign_toplevel_handle_v1_create(server.foreign_toplevel_list, &foreign_toplevel_state);
```
emersion, line 756: Hm, it's not nice to create the foreign toplevel handle with an empty initial state.

Is it too early to call `view_get_app_id()` here?

sway/tree/view.c
```diff
@@ -1060,6 +1075,14 @@ void view_update_title(struct sway_view *view, bool force) {
 	if (view->foreign_toplevel && title) {
 		wlr_foreign_toplevel_handle_v1_set_title(view->foreign_toplevel, title);
 	}
+
+	if (view->ext_foreign_toplevel) {
+		struct wlr_ext_foreign_toplevel_handle_v1_state foreign_toplevel_state = {
+			.app_id = (char *)view_get_app_id(view),
+			.title = (char *)title,
+		};
+		wlr_ext_foreign_toplevel_handle_v1_update_state(view->ext_foreign_toplevel, &foreign_toplevel_state);
```
emersion, line 1084: Suggestion: maybe we can extract this into a helper function?

emersion:
We also need to update the state when the app_id changes.

columbarius:
> We also need to update the state when the app_id changes.

This is currently not done for the toplevel_management protocol, right?

emersion:
>      This is currently not done for the toplevel_management protocol, right?

Correct, the old impl only sets app_id when mapping and then ignores any changes.

sway/tree/view.c
columbarius: Done

sway/tree/view.c
columbarius: Testing showed we can create the handle with app_id and title!

sway/tree/view.c
columbarius: Done

columbarius:
I've added a function updating the app_id analog to the title.

sway/tree/view.c
```diff
@@ -582,6 +582,14 @@ static struct sway_workspace *select_workspace(struct sway_view *view) {
 	return NULL;
 }
 
+static void update_ext_foreign_toplevel(struct sway_view *view) {
+	struct wlr_ext_foreign_toplevel_handle_v1_state toplevel_state;
+	toplevel_state.app_id = view_get_app_id(view);
+	toplevel_state.title = view_get_title(view);
+	sway_log(SWAY_ERROR, "Update ext foreign topleven: app_id %s, title %s", toplevel_state.app_id, toplevel_state.title);
```
emersion, line 589: Remnant from a debugging session?

sway/tree/view.c
```diff
@@ -582,6 +582,14 @@ static struct sway_workspace *select_workspace(struct sway_view *view) {
 	return NULL;
 }
 
+static void update_ext_foreign_toplevel(struct sway_view *view) {
+	struct wlr_ext_foreign_toplevel_handle_v1_state toplevel_state;
+	toplevel_state.app_id = view_get_app_id(view);
+	toplevel_state.title = view_get_title(view);
```
emersion, line 588: Please use initializers when possible: this ensures there are no undefined memory if a new field is added to the struct in a new wlroots version.

```c
const struct wlr_ext_foreign_toplevel_handle_v1_state toplevel_state = {
    .app_id = view_get_app_id(view),
    .title = view_get_title(view),
};
```

emersion:
With these fixed, LGTM!

sway/tree/view.c
columbarius: Ok, removed to logging.

sway/tree/view.c
columbarius: Done

emersion (approved):
Thanks!

columbarius:
Thanks for merging!

PR #7972, opened 2024-02-11
output: Fallback to sorting workspaces alphabetically

epontan:
Currently the order of unnumbered workspaces is the order they were created. This is a bit unintuitive for workflows where workspaces are created and named dynamically.

Instead of treating all unnumbered workspaces equal, fallback to sorting them alphabetically based on the name of the workspace. This will make the ordering more deterministic regardless of the order they were created.

This change will probably go unnoticed for a lot users since most people are using static workspace configuration connected to numbers. However it is a great improvement for workflows that create new workspaces on the fly with an associated name.

### DIFF
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -9,6 +9,7 @@
 #include "sway/tree/workspace.h"
 #include "sway/server.h"
 #include "log.h"
+#include "stringop.h"
 #include "util.h"
 
 enum wlr_direction opposite_direction(enum wlr_direction d) {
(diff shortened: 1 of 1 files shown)


Nefsen402:
@emersion Do you have any comments on this. I personally don't use numbered workspaces so this would also benefit me. If there are no comments about this, I'll merge after the weekend.

emersion:
What does i3 do here?

epontan:
> What does i3 do here?

I think it will keep the order it was created. The behavior was recently updated though it seems with this [PR](https://github.com/i3/i3/pull/4564).

I once had a local patch to do similar fallback sorting of unnumbered workspaces, but never got to create a PR for it as I was struggling a bit with their test system and eventually switched to sway.

PR #7975, opened 2024-02-12
view: drop ext_foreign_destroy

vyivel:
It's not used and causes a crash when a view is destroyed.


emersion:
Eh right, we shouldn't really need this.

cc @columbarius just FYI

PR #7986, opened 2024-02-16
Use backend commits for wlr-output-management-unstable-v1

emersion:
See https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4548


kennylevinsen:
Already handled in master.

PR #7988, opened 2024-02-17
input/text_input: fix dangling listeners

emersion:


Nefsen402:
There are also other signals like `input_method_grab_keyboard` and `input_method_commit` that don't seem to be cleaned up. But I just searched the single implementation file.

emersion:
Ah, yes, fixed.

PR #7989, opened 2024-02-17
docs: remove duplicate reference to SWAYSOCK

ammgws:


emersion:
The reason for the duplication is that `SWAYSOCK` both has an effect on Sway and is also set by Sway.

PR #7998, opened 2024-02-26
Add `floating` and `scratchpad_state` properties to GET_TREE to restore IPC parity with i3

llyyr:
See: https://i3wm.org/docs/ipc.html


sway/ipc-json.c
```diff
@@ -693,9 +696,18 @@ static void ipc_json_describe_container(struct sway_container *c, json_object *o
 	json_object_object_add(object, "urgent", json_object_new_boolean(urgent));
 	json_object_object_add(object, "sticky", json_object_new_boolean(c->is_sticky));
 
+	// Unlike X11, windows in Wayland cannot request to be floated, so they can
+	// never be "auto_on" or "user_off".
```
emersion, line 700: This isn't quite correct AFAIU:

- Wayland native programs are auto-floated if they are unresizable
- We support X11 windows via Xwayland

emersion (approved):
Thanks!

PR #8008, opened 2024-02-27
text_input: don't destroy scene_node twice

llyyr:


emersion (approved):
Ah, I've left a review comment about this in the original PR but seems like it got lost at some point…

Thanks!

Issue #8012, opened 2024-02-28
Focus on window activation doesn't work with xdg-open


martinsolberg:
### Please fill out the following:
- **Sway Version:**
  - 1.9

- **Configuration File:**
  - `focus_on_window_activation focus`

- **Description:**
  - Opening links in Firefox via `xdg-open https://url` does not focus the browser window. This is due to request being downgraded to urgency. See [discussion in PR](https://github.com/swaywm/sway/pull/7880) for details.

PR #8017, opened 2024-02-28
xdg-activation: launcher tokens are activation requests

rpigott:
If the launched client decides to pass it's token along as an activation request, allow that. This will make the behavior match tokens provided by an external launcher client.

### DIFF
--- a/sway/xdg_activation_v1.c
+++ b/sway/xdg_activation_v1.c
@@ -38,14 +38,14 @@ void xdg_activation_v1_handle_request_activate(struct wl_listener *listener,
 	}
 
 	// This is an activation request. If this context is internal we have ctx->seat.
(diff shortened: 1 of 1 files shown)


rpigott:
Reported in https://www.reddit.com/r/swaywm/comments/1b2ha6b/migrating_from_i3_for_windows_move_up_command_not/

rpigott:
I am marking this to fix #8012.

It may be required to use `swaymsg exec xdg-open` in some cases if the calling application does not provide it's own token, or the user might elect to just replace xdg-open with a version that uses swaymsg exec in sway.

rpigott:
@emersion any thoughts on this one?

emersion (approved):
Thanks!

PR #8020, opened 2024-02-29
commands/move: do not force focus on the moved container

rpigott:
My code archaeology isn't good enough to determine what this is here for, but it isn't correct. We should be able to move containers in a direction without focusing them. AFAICT i3 doesn't do this, so we shouldn't either.

This fixes ipc commands like move \<dir\> with criteria that apply to containers which are not the current focus.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -769,15 +769,6 @@ static struct cmd_results *cmd_move_in_direction(
 		ipc_event_window(container, "move");
 	}
 
-	// Hack to re-focus container
-	seat_set_raw_focus(config->handler_context.seat, &new_ws->node);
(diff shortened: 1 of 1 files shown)


rpigott:
Reported in https://www.reddit.com/r/swaywm/comments/1b2ha6b/migrating_from_i3_for_windows_move_up_command_not/

emersion:
Hmm, it seems like this was added here, but without explanation:

https://github.com/swaywm/sway/commit/7586f150c058997d9dde387ea7c091ffa7a3c3c7#diff-fcf5a35c88e94d24870f5242b57cc1120fd954e9849ed7dce9e05efed9be4748R728

emersion (approved):
Let's go with this and see if it causes any regressions. Thanks!

Issue #7409, opened 2023-01-30
Hide the titlebar


bl4ckb0ne:
Hiding the titlebar is done via a workaround to set the font size to 0 (see https://github.com/swaywm/sway/pull/7092 and #7408)

Font size 0 should not be allowed, instead there should be a proper way to hide the titlebars done via the config file.

Ref: 

#7350, https://github.com/swaywm/sway/issues/6946, https://github.com/swaywm/sway/issues/5050 and https://github.com/swaywm/sway/issues/1286

Inginn:
I am at my limit, I have the 0.01 height bandaid. but I keep pressing it with my mouse if I am using chrome and jumping to another window that is behind.
I know this is not part of i3 but I could use sway befor they "fixed" it.

WhyNotHugo:
I'd love to do this on a per-client basis. Some clients already draw functional titlebars, so I'd want to disable titlebars for those specifically.

Given that this needs to be designed from scratch, does applying it on a per-client basis sound reasonable? The all-windows use case can be covered by simply:

    for_window * titlebar hide

rpigott:
The client can already indicate when they draw their own decorations with xdg_decoration. Only allowed for floating clients atm (otherwise we enforce ssd), though I don't remember why.

MarcioGCosta:
Setting the default_border as pixel 0 i was able to get rid of the title bar
 
I am running version 1.8.2 in void

rpigott:
We restored this workaround in 1.8.2 until a proper feature is added.

hashworks:
A feature similar to `smart_gaps` would be great as well: `smart_titlebar on` would only show the titlebar when two or more windows are visible in a workspace.

tmpm697:
> Setting the default_border as pixel 0 i was able to get rid of the title bar
> 
> I am running version 1.8.2 in void

not work for me.

currently the only way to get rid of titlebar is setting font:
```
font pango:monospace 0
```

please do not remove this setting until we can provide a solution for it, my workflow adheres with it.

ShalokShalom:
> A feature similar to `smart_gaps` would be great as well: `smart_titlebar on` would only show the titlebar when two or more windows are visible in a workspace.

You can do that with
hide_edge_borders --i3 smart

hashworks:
Hm, that doesn't work for me. `hide_edge_borders --i3 smart` still shows a title bar even when only one tab is open. I guess I'll open a bug report later.

[...]

PR #8026, opened 2024-03-01
added 'disable_titlebar' option

neuromagus:
I fix my previous commit to master branch.

My patch fixed this issue https://github.com/swaywm/sway/issues/7409#issue-1563121281

and you fixed the rendering so my patch is more trivial. Amazing work, guys.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -124,6 +124,7 @@ sway_cmd cmd_create_output;
 sway_cmd cmd_default_border;
 sway_cmd cmd_default_floating_border;
 sway_cmd cmd_default_orientation;
+sway_cmd cmd_disable_titlebar;
 sway_cmd cmd_exec;
 sway_cmd cmd_exec_always;
 sway_cmd cmd_exit;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -532,6 +532,7 @@ struct sway_config {
 	bool validating;
 	bool auto_back_and_forth;
 	bool show_marks;
+	bool disable_titlebar;
 	enum alignment title_align;
 	bool primary_selection;
 
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -57,6 +57,7 @@ static const struct cmd_handler handlers[] = {
 	{ "client.urgent", cmd_client_urgent },
 	{ "default_border", cmd_default_border },
 	{ "default_floating_border", cmd_default_floating_border },
+	{ "disable_titlebar", cmd_disable_titlebar },
 	{ "exec", cmd_exec },
 	{ "exec_always", cmd_exec_always },
 	{ "floating_maximum_size", cmd_floating_maximum_size },
--- a/sway/commands/disable_titlebar.c
+++ b/sway/commands/disable_titlebar.c
@@ -1,0 +1,15 @@
+#include "sway/commands.h"
+#include "sway/config.h"
+#include "util.h"
+
+struct cmd_results *cmd_disable_titlebar(int argc, char **argv) {
+    struct cmd_results *error = NULL;
+    if ((error = checkarg(argc, "disable_titlebar", EXPECTED_EQUAL_TO, 1))) {
+        return error;
+    }
+
+    config->disable_titlebar = parse_boolean(argv[0], config->disable_titlebar);
+
+    return cmd_results_new(CMD_SUCCESS, NULL);
+}
+
--- a/sway/config.c
+++ b/sway/config.c
@@ -271,6 +271,7 @@ static void config_defaults(struct sway_config *config) {
 	config->auto_back_and_forth = false;
 	config->reading = false;
 	config->show_marks = true;
+	config->disable_titlebar = false;
 	config->title_align = ALIGN_LEFT;
 	config->tiling_drag = true;
 	config->tiling_drag_threshold = 9;
--- a/sway/meson.build
+++ b/sway/meson.build
@@ -50,6 +50,7 @@ sway_sources = files(
 	'commands/default_border.c',
 	'commands/default_floating_border.c',
 	'commands/default_orientation.c',
+        'commands/disable_titlebar.c',
 	'commands/exit.c',
 	'commands/exec.c',
 	'commands/exec_always.c',
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -726,6 +726,9 @@ The default colors are:
 	should be greater than titlebar_border_thickness. If _vertical_ value is
 	not specified it is set to the _horizontal_ value.
 
+*disable_titlebar* yes|no.
+	Remove titlebar. Default is _no_.
+
 *for_window* <criteria> <command>
 	Whenever a window that matches _criteria_ appears, run list of commands.
 	See *CRITERIA* for more details.
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -722,6 +722,10 @@ void container_update_representation(struct sway_container *con) {
 }
 
 size_t container_titlebar_height(void) {
+	if(config->disable_titlebar) {
+		return 0;
+	}
+	
 	return config->font_height + config->titlebar_v_padding * 2;
 }
 


neuromagus:
@WhyNotHugo 
> I would like to see the ability to toggle this on a per-window basis (much like we can toggle `floating` per-window).

Hmm, why not? Particularly this commit focus on total disable any titlebars. IMO, need add another option and way for hotkey with toggle behavior switcher. If need this stuff, again, why not?

For example - press Mod+t - disable titlebar on focused window? Am I right?

P.S. My English is so terrible, sorry for that.

y0nei:
We need this

Nefsen402:
`container_titlebar_height` is only used when using the `B_NORMAL` mode for borders. `B_PIXEL` will use a size determined by `container->current.border_top`. I think your patch should have correct behavior if instead of trying to change `container_titlebar_height` in any way. This should instead work:

```
diff --git a/sway/desktop/transaction.c b/sway/desktop/transaction.c
index 042141ab..4895add1 100644
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -392,6 +392,10 @@ static void arrange_container(struct sway_container *con,
                int border_top = container_titlebar_height();
                int border_width = con->current.border_thickness;
 
+               if (wants to disable title bar) {
+                       title_bar = false;
+               }
+
                if (title_bar && con->current.border != B_NORMAL) {
                        wlr_scene_node_set_enabled(&con->title_bar.tree->node, false);
                        wlr_scene_node_set_enabled(&con->border.top->node, true);
```

However, it might be beneficial to keep the current behavior. The top border size can be manually set to 0 by the user if they truly don't want any top decorations of any kind.

neuromagus:
Thanks for the quick and detailed answer. not everything is so clear here. For your recommendation top border did not showed if we remove titlebar, because rendering  happens below. ```titlebar = false``` doesn't allow me to use the ```if (title_bar... blah-blah)``` where rendering border_top.

So, I'm stuck, need copy render func in this construction
```c
if (wants to disable title bar) {
    title_bar = false;
    border_top = blah-blah
    ...
    wlr_scene_node_set_enabled(&con->border.top->node, true);
}
```
or insert inside "if" more checking? 
```c
} else if (con->current.border != B_NORMAL && config->disable-titlebar) {
   //copy here code for initial size and render border_top
}
```

neuromagus:
@Nefsen402

>However, it might be beneficial to keep the current behavior. The top border size can be manually set to 0 by the user if they truly don't want any top decorations of any kind.

After researching and testing TABBED mode with borders, I found this configuration to be unfortunate for this mode, the borders made it difficult to hit the scrollbar by mouse...

So maybe, but not necessary to be fixed manpages to recommend options with ```disable_titlebar``` (default_border none, default_floating_border none/pixel 2  etc) and that's it.

tmpm697:
I'm looking forward for at least to disable title on `tabbed` window as that the currently use case for me.

neuromagus:
use working solution today, @tmpm697  xD - https://github.com/neuromagus/disable_titlebar_in_sway

tmpm697:
tks @neuromagus , hope it merged soon, critical to one of my favourite workflow.

quincyf467:
Thanks for patch! Works as intended

vwyu (approved):

vwyu (approved):

neuromagus:
yes, master and 1.9 - are different beasts. Use [this](https://github.com/neuromagus/disable_titlebar_in_sway) version for normal work.

neuromagus:
>The term "titlebar" refers to the whole window decoration, including borders. Maybe using "title" instead would be more accurate? (This would be consistent with e.g. "title_format")

As you wish

>Do we want a new command, or do we want to special-case the empty string for "title_format"? It may be desirable for users to have a titlebar with no contents, however that could be achieved with a space character, for instance…

As you wish

>Do we want to extend "default_border" instead?

As you wish, I only want use Sway with no Titlebars (in any layouts).

neuromagus:
@emersion 
something wrong with Wlroots for testing (archlinux)!

```
Cloning into 'wlroots'...
fatal: unable to access 'https://gitlab.freedesktop.org/wlroots/wlroots.git/': The requested URL returned error: 504
[#1215763] 2024/05/08 16:03:16 Failed to clone git repository. If this a private repository, make sure you've added a suitable SSH key.
```

fossdd:
> something wrong with Wlroots for testing (archlinux)!

thats a issue on your side, working perfectly with my internet connection.

kennylevinsen:
> The term "titlebar" refers to the whole window decoration, including borders. Maybe using "title" instead would be more accurate? (This would be consistent with e.g. "title_format")

To nitpick, a titlebar has "padding", and a window has either a top border (`border pixel`) or a titlebar (`border normal`, or in stacking/tabbed mode).

I imagine users want the `border pixel` behavior, but for `stacking`/`tabbed`.

> > Do we want to extend "default_border" instead?
> 
> This is separate entity, IMHO. Someone want borders [without titlebars](https://github.com/neuromagus/disable_titlebar_in_sway/issues/1)... Should I add all this to my patch or you could make the adjustments yourself in Sway? I only want to use Sway with no "Titlebars" (on any layouts).

The `border` command currently controls the titlebar behavior for everything *but* stack/tabbed views. The options are:

1. `none`: No border and no titlebar
2. `normal`: Border (left, right, bottom) and titlebar
3. `pixel`: Border (left, right, top, bottom) with no titlebar
4. `csd`: No border and no titlebar, client-side decorations requested to replace titlebar

The problem is that stacking and tabbed do not respect `pixel`, and instead behave as if the border was set to `normal` instead.

In that sense, the goal is not to "disable the titlebar" - we have two toggles for that already in `border` - but to have a way to have a container that has `stacking` behavior that respects `border pixel`. We can either:

1. Have a new `always_pixel` border mode which also applies to tabbed/stacking.
2. Have a `grouped_titlebar always|follow_border` setting to control whether the titlebar for `stacking`/`tabbed` should respect `border pixel` or not. Or, if we wanted to get fancy, `titlebar splitv|splith|stacking|tabbed always|never|follow_border`.
3. Have a new container type like `stacking`/`tabbed`, but dedicated to the "swallow" use-case without navigation UI.

The second option would probably be the easiest to wire up. I don't mind the last option either, but it has IPC consequences and might not be worth it.

neuromagus:
>pixel: Border (left, right, top, bottom) with no titlebar

IMO, need a simpler config name. "pixel"? 

>The problem is that stacking and tabbed do not respect pixel, and instead behave as if the border was set to normal instead.

this is not a problem. If u disable titlebar, you create (or render) container with respects new rules. Sorry for my eng, how this stuff realized in Sway:
if u have titlebar - u add in container static pixel size (padding, font size, border size etc). I just add a hack - "disable titlebar" -> this size equals zero.

In my opinion - option "disable titlebar" need set alone, border size set to another config group. Maybe later need update titlebar render, some containers show titlebar, some not...

kennylevinsen:
> IMO, need a simpler config name. "pixel"?

Those are the existing names from i3, so changing those is not an option.

> this is not a problem. If you disable the titlebar, you create (or render) container with respects new rules. Sorry for my eng, how this stuff realized in Sway:
> if you have a titlebar - u add in container static pixel size (padding, font size, border size etc). I just add a hack - "disable titlebar" -> this size equals to zero.

The linked issues are all about having titlebar *also* be removed for `stacking`/`tabbed` layout, because you can already do this for everything else.

Bugs aside (including not reinstalling the top border when removing the titlebar), this also the only functional difference your PR provides over `border pixel`/`border none`.

> In my opinion - option "disable titlebar" need to set alone, border size set to another config group. Maybe later we need to update titlebar render, some containers show titlebar, some not...

The titlebar is already controlled by the `border` setting *except* for `stacking`/`tabbed`. I don't think we want *two* ways to show/hide the titlebar, just a way to have the setting also apply to stacking/tabbed layouts.

WhyNotHugo:
> I imagine users want the border pixel behavior, but for stacking/tabbed.

Personally, I wouldn't mind seeing the top border for tabbed windows too. It always felt kind of weird that tabbed windows only have three (right, left, bottom) borders).

> The titlebar is already controlled by the border setting except for stacking/tabbed. I don't think we want two ways to show/hide the titlebar, just a way to have the setting also apply to stacking/tabbed layouts.

Perhaps we want a setting to enable/disable titlebars for _containers_, since tabbed titlebars actually belong to the container itself, no?

neuromagus:
>The titlebar is already controlled by the border setting except for stacking/tabbed. I don't think we want two ways to show/hide the titlebar, just a way to have the setting also apply to stacking/tabbed layouts.

Nice to read this. Well, we need to add this functionality to "stacking/tabbed" or other layouts!.. And maybe add behavior like hotkey for enable/disable titlebar for this and only this container?

kennylevinsen:
> Perhaps we want a setting to enable/disable titlebars for containers, since tabbed titlebars actually belong to the container itself, no?

In case of stacked/tabbed, the titlebar for children is drawn by the parent container, whereas for splitv/splith/floating, the titlebar is drawn by the child containers.

That's also just an implementation detail, from the user perspective the titlebar is above the child container regardless of how it got there.

neuromagus:
>That's also just an implementation detail, from the user perspective the titlebar is above the child container regardless of how it got there.

yep.

PR #8033, opened 2024-03-03
tree/view: Do not clip to geometry if using CSD

thal:
As of commit 9da295c11, surfaces of `xdg_toplevel`s are clipped to a region defined by the geometry. This causes drop shadows of floating windows to be cut off, as seen with gnome-calculator against a white background:
image
This PR will avoid clipping if a `sway_view` is in a floating container and using CSD.

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -927,11 +927,14 @@ void view_update_size(struct sway_view *view) {
 void view_center_and_clip_surface(struct sway_view *view) {
 	struct sway_container *con = view->container;
 
+	bool clip_to_geometry = true;
+
 	if (container_is_floating(con)) {
 		// We always center the current coordinates rather than the next, as the
 		// geometry immediately affects the currently active rendering.
 		int x = (int) fmax(0, (con->current.content_width - view->geometry.width) / 2);
 		int y = (int) fmax(0, (con->current.content_height - view->geometry.height) / 2);
+		clip_to_geometry = !view->using_csd;
 
 		wlr_scene_node_set_position(&view->content_tree->node, x, y);
 	} else {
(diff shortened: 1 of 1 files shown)


sway/tree/view.c
```diff
@@ -940,12 +943,16 @@ void view_center_and_clip_surface(struct sway_view *view) {
...
-		wlr_scene_subsurface_tree_set_clip(&con->view->content_tree->node, &(struct wlr_box){
-			.x = con->view->geometry.x,
-			.y = con->view->geometry.y,
-			.width = con->current.content_width,
-			.height = con->current.content_height,
-		});
+		struct wlr_box clip = {0};
+		if (clip_to_geometry) {
+			clip = (struct wlr_box){
+				.x = con->view->geometry.x,
```
rpigott, line 949: I think it should just be using `con->current.content_x` instead of `con->view->geometry.x` here.

thal: Tried this but it did not work properly. Looks like `content_x` and `content_y` are absolute coordinates, while `subsurface_tree_set_clip` expects surface-local coordinates.

thal:
The new revision passes `con->view->geometry` to `subsurface_tree_set_clip` which is more consistent with the coordinates, and cleans up the code slightly.

Nefsen402:
I think the only difference between clipping with the geometry size instead of the `content_{width, height}` size is that if the application decides it shouldn't configure to the size of the window sway wanted. This is a problem when the application is bigger than the frame (this can be easy to trigger if you tile firefox and make it small - it will refuse to configure smaller than a certain size). This will allow firefox to overflow the bounds of the container. I believe we need to keep the `content_{width, height}` size.

thal:
I've restored the patch to use `content_width`, `content_height`.
@Nefsen402, I tested  with Firefox as you suggested, but I did not see any issues with either version of the patch, even when the configured size was smaller than the container.

emersion (approved):
LGTM, thanks!

PR #8044, opened 2024-03-07
Force bilinear scaling when scaling down

Nefsen402:
Untested

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -183,7 +183,15 @@ static void send_frame_done_iterator(struct wlr_scene_buffer *buffer,
 	}
 }
 
-static enum wlr_scale_filter_mode get_scale_filter(struct sway_output *output) {
+static enum wlr_scale_filter_mode get_scale_filter(struct sway_output *output,
+		struct wlr_scene_buffer *buffer) {
+	// if we are scaling down, we should always choose linear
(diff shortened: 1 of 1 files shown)


emersion (approved):
Works, and mirrors what Sway 1.8 was doing with MIN/MAG_FILTER.

Thanks!

Nefsen402:
Scaling mode won't make a difference if there is scaling unless there is a bug in a renderer that stretches the texture randomly by one pixel.

akelmanson:
downscaling with nearest filter (waybar): 👎
image

downscaling with bilinear filter (waylar): 👍
image

buffer already scaled by the client with bilinear filter (chrome): 👎 (blur)
image

buffer already scaled by the client with nearest filter (chrome): 👍
image

do you it's a bug in the renderer?

Nefsen402:
It might also be caused by a slightly shifted position causing the texture to be sampled between texels. But I don't feel like tracking down why a texture might be shifted by .5 pixels.

Nefsen402:
In that case, it might be a good idea to open a PR with your suggested change with this change:

```
	if ((buffer->dst_width == 0 && buffer->dst_height == 0) || (
			buffer->buffer_width == buffer->dst_width &&
			buffer->buffer_height == buffer->dst_height
	)) {
		return WLR_SCALE_FILTER_NEAREST;
	}
```

This PR simply restores prior behavior.

PR #8045, opened 2024-03-08
desktop/output: merge power management config on top of current one

emersion:
Don't replace an existing config with a blank one with just the
power field set.

Logic borrowed from apply_output_config_to_outputs().


emersion:
Nevermind, `store_output_config()` does the merging already…

PR #8046, opened 2024-03-08
config/output: drop enabling flag

emersion:
This was useful when wlroots backends were updating the current mode on their own. This is no longer the case.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

PR #8049, opened 2024-03-08
config/output: return merged config in merge_id_on_name()

emersion:
Fixes the case where store_output_config() is called with an output config referring to a name, and an existing config exists with an identifier. The existing config would get ignored.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -151,10 +151,10 @@ static void merge_wildcard_on_all(struct output_config *wildcard) {
 	}
 }
 
-static void merge_id_on_name(struct output_config *oc) {
+static struct output_config *merge_id_on_name(struct output_config *oc) {
 	struct sway_output *output = all_output_by_name_or_id(oc->name);
 	if (output == NULL) {
-		return;
+		return oc;
 	}
 
 	const char *name = output->wlr_output->name;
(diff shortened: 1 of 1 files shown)


sway/config/output.c
```diff
@@ -239,6 +241,12 @@ struct output_config *store_output_config(struct output_config *oc) {
 		oc->transform, oc->background, oc->background_option, oc->power,
 		oc->max_render_time);
 
+	if (wildcard) {
+		merge_wildcard_on_all(oc);
+	} else {
+		oc = merge_id_on_name(oc);
+	}
```
emersion, line 248: Not sure the ordering of this wrt. the logic above matters…? Maybe it does?

kennylevinsen: Store_output_config is still weird. I feel like it would make more sense to:
1. Merge the config with anything existing (once!) and return the resulting config for callers to use
2. Otherwise, update the name according to the `id_on_name` scheme, add the config to the list and return it.

(I haven't spent any braincycles on the wildcard side of things.)

kennylevinsen: I gave the cleanup I had in mind a go in https://github.com/swaywm/sway/pull/8050.

emersion:
Superseded by https://github.com/swaywm/sway/pull/8069

PR #8050, opened 2024-03-08
config/output: Fix store_output_config return

kennylevinsen:
When we receive a new output config, we merge it with existing configurations. Make sure that the caller of store_output_config always receives the exactly the configuration that was stored.

Based on https://github.com/swaywm/sway/pull/8049.

Not tested.


kennylevinsen:
Better solution in: https://github.com/swaywm/sway/pull/8069

PR #8063, opened 2024-03-14
Handle GPU resets

emersion:
Depends on https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4606 to work properly.

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -172,6 +172,45 @@ static void detect_proprietary(struct wlr_backend *backend, void *data) {
 	drmFreeVersion(version);
 }
 
+static void handle_renderer_lost(struct wl_listener *listener, void *data) {
+	struct sway_server *server = wl_container_of(listener, server, renderer_lost);
+
+	sway_log(SWAY_INFO, "Re-creating renderer after GPU reset");
+
+	struct wlr_renderer *renderer = wlr_renderer_autocreate(server->backend);
+	if (renderer == NULL) {
+		sway_log(SWAY_ERROR, "Unable to create renderer");
+		return;
+	}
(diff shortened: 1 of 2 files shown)


sway/server.c
```diff
@@ -172,6 +172,45 @@ static void detect_proprietary(struct wlr_backend *backend, void *data) {
...
 
+static void handle_renderer_lost(struct wl_listener *listener, void *data) {
+	struct sway_server *server = wl_container_of(listener, server, renderer_lost);
+
+	sway_log(SWAY_INFO, "Re-creating renderer after GPU reset");
+
+	struct wlr_renderer *renderer = wlr_renderer_autocreate(server->backend);
+	if (renderer == NULL) {
+		sway_log(SWAY_ERROR, "Unable to create renderer");
+		return;
```
kennylevinsen, line 183: Is there a reason to keep running if we fail to recreate the renderer? Unless the user has a way to recover, it might be better to just exit at this point on failure.

emersion: There are two reasons I can think of:

- Maybe the GPU is still in a bad state, and with a later reset event we can get into a better state
- Maybe the user has important unsaved work, and has ways to recover that work via other means (via an IPC, manually via GDB…)
- Maybe it's more important for the graphical apps to keep running even if there is no graphical output anymore (e.g. multiplayer LAN games)

But yeah, it's definitely a debatable choice.

kennylevinsen: Hmm, yeah fair enough. The failure mode won't be great regardless of what we do.

kennylevinsen (approved):

PR #8068, opened 2024-03-16
output: wlr_backend_commit and wlr_swapchain_helper

kennylevinsen:
Use the new `wlr_backend_commit` and `wlr_swapchain_helper` to light up outputs and apply configuration changes.

Works in local testing, but as sway's output configuration system is a bit complicated it wouldn't hurt with more testing.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -809,6 +809,34 @@ out:
 	return ok;
 }
 
+void apply_all_output_configs(void) {
+	size_t configs_len = wl_list_length(&root->all_outputs);
+	struct matched_output_config *configs = calloc(configs_len, sizeof(*configs));
+	if (!configs) {
+		return;
+	}
+
+	int config_idx = 0;
+	struct sway_output *sway_output;
+	wl_list_for_each(sway_output, &root->all_outputs, link) {
+		if (sway_output == root->fallback_output) {
+			configs_len--;
+			continue;
+		}
+
+		struct matched_output_config *config = &configs[config_idx++];
+		config->output = sway_output;
+		config->config = find_output_config(sway_output);
+	}
+
+	apply_output_configs(configs, configs_len, false);
+	for (size_t idx = 0; idx < configs_len; idx++) {
+		struct matched_output_config *cfg = &configs[idx];
+		free_output_config(cfg->config);
+	}
+	free(configs);
+}
+
 void apply_output_config_to_outputs(struct output_config *oc) {
 	size_t configs_len = wl_list_length(&root->all_outputs);
 	struct matched_output_config *configs = calloc(configs_len, sizeof(*configs));
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -79,7 +79,7 @@ struct output_config *new_output_config(const char *name) {
 	return oc;
 }
 
-void merge_output_config(struct output_config *dst, struct output_config *src) {
+static void merge_output_config(struct output_config *dst, struct output_config *src) {
 	if (src->enabled != -1) {
 		dst->enabled = src->enabled;
 	}
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -807,59 +807,6 @@ void apply_all_output_configs(void) {
 	free(configs);
 }
 
-void apply_output_config_to_outputs(struct output_config *oc) {
-	size_t configs_len = wl_list_length(&root->all_outputs);
-	struct matched_output_config *configs = calloc(configs_len, sizeof(*configs));
-	if (!configs) {
-		return;
-	}
-
-	// Try to find the output container and apply configuration now. If
-	// this is during startup then there will be no container and config
-	// will be applied during normal "new output" event from wlroots.
-	int config_idx = 0;
-	struct sway_output *sway_output;
-	wl_list_for_each(sway_output, &root->all_outputs, link) {
-		if (sway_output == root->fallback_output) {
-			configs_len--;
-			continue;
-		}
-
-		struct matched_output_config *config = &configs[config_idx++];
-		config->output = sway_output;
-		config->config = find_output_config(sway_output);
-
-		if (!output_match_name_or_id(sway_output, oc->name)) {
-			continue;
-		}
-
-		if (!config->config && oc) {
-			// No stored output config matched, apply oc directly
-			sway_log(SWAY_DEBUG, "Applying oc directly");
-			config->config = new_output_config(oc->name);
-			merge_output_config(config->config, oc);
-		}
-	}
-
-	apply_output_configs(configs, configs_len, false);
-	for (size_t idx = 0; idx < configs_len; idx++) {
-		struct matched_output_config *cfg = &configs[idx];
-		free_output_config(cfg->config);
-	}
-	free(configs);
-}
-
-void reset_outputs(void) {
-	struct output_config *oc = NULL;
-	int i = list_seq_find(config->output_configs, output_name_cmp, "*");
-	if (i >= 0) {
-		oc = config->output_configs->items[i];
-	} else {
-		oc = store_output_config(new_output_config("*"));
-	}
-	apply_output_config_to_outputs(oc);
-}
-
 void free_output_config(struct output_config *oc) {
 	if (!oc) {
 		return;
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -568,16 +568,6 @@ static bool finalize_output_config(struct output_config *oc, struct sway_output 
 	return true;
 }
 
-bool test_output_config(struct output_config *oc, struct sway_output *output) {
-	if (output == root->fallback_output) {
-		return false;
-	}
-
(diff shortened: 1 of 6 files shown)


sway/config/output.c
```diff
@@ -705,50 +677,109 @@ struct output_config *find_output_config(struct sway_output *output) {
...
+		sway_log(SWAY_ERROR, "Swapchain prepare failed");
+		goto out;
+	}
+
+	for (size_t idx = 0; idx < configs_len; idx++) {
+		struct matched_output_config *cfg = &configs[idx];
+		struct wlr_backend_output_state *backend_state = &states[idx];
+
+		if (!wlr_scene_output_build_state(cfg->output->scene_output, &backend_state->base, NULL)) {
+			sway_log(SWAY_ERROR, "Building output stat for '%s' failed", backend_state->output->name);
```
Nefsen402, line 712: typo: `stat`

kennylevinsen: Woops.

kennylevinsen:
Had entirely forgotten to use the new swapchain in the `wlr_scene_output_build_state` call, added.

sway/config/output.c
```diff
@@ -716,39 +717,126 @@ struct output_config *find_output_config(struct sway_output *output) {
...
+		}
+
+		struct matched_output_config *config = &configs[config_idx++];
+		config->output = sway_output;
+
+		if (!output_match_name_or_id(sway_output, oc->name)) {
+			continue;
+		}
+
+		config->config  = find_output_config(sway_output);
```
emersion: Nit: extra space before `=`

sway/config/output.c
```diff
@@ -716,39 +717,126 @@ struct output_config *find_output_config(struct sway_output *output) {
...
+	wl_list_for_each(sway_output, &root->all_outputs, link) {
+		if (sway_output == root->fallback_output) {
+			configs_len--;
+			continue;
+		}
+
+		struct matched_output_config *config = &configs[config_idx++];
+		config->output = sway_output;
+
+		if (!output_match_name_or_id(sway_output, oc->name)) {
```
emersion: Hm, so this will leave a NULL output config here? And I think a NULL config is equivalent to a config which disables outputs? But I'm not sure it's what we want?

kennylevinsen: Hmm, queue_pending_config actually attempts to enable the output when presented with a NULL config. My intent was to maintain the current, config (i.e., include an empty state).

Let's just always include the current known config. This function is removed later anyway.

sway/config/output.c
kennylevinsen: Woops!

sway/desktop/output.c
```diff
@@ -550,63 +550,73 @@ void handle_gamma_control_set_gamma(struct wl_listener *listener, void *data) {
...
-			ok &= apply_output_config(oc, output);
+		struct matched_output_config *cfg = &configs[config_idx++];
+		cfg->output = sway_output;
+
+		struct wlr_output_configuration_head_v1 *config_head;
+		wl_list_for_each(config_head, &config->heads, link) {
+			if (config_head->state.output == sway_output->wlr_output) {
+				cfg->config = output_config_for_config_head(config_head, sway_output);
+				break;
+			}
```
emersion, line 606: Hm. Again here, if an output wasn't included in the output management config, we should re-use the current output config...

sway/desktop/output.c
```diff
@@ -550,63 +550,73 @@ void handle_gamma_control_set_gamma(struct wl_listener *listener, void *data) {
...
-		oc->transform = config_head->state.transform;
-		oc->scale = config_head->state.scale;
-		oc->adaptive_sync = config_head->state.adaptive_sync_enabled;
-
-		if (test_only) {
-			ok &= test_output_config(oc, output);
+	bool ok = apply_output_configs(configs, configs_len);
+	for (size_t idx = 0; idx < configs_len; idx++) {
+		struct matched_output_config *cfg = &configs[idx];
+		if (ok) {
```
emersion, line 613: We're missing a `&& !test_only` here I think?

sway/desktop/output.c
```diff
@@ -550,63 +550,73 @@ void handle_gamma_control_set_gamma(struct wl_listener *listener, void *data) {
...
-		}
-		oc->x = config_head->state.x;
-		oc->y = config_head->state.y;
-		oc->transform = config_head->state.transform;
-		oc->scale = config_head->state.scale;
-		oc->adaptive_sync = config_head->state.adaptive_sync_enabled;
-
-		if (test_only) {
-			ok &= test_output_config(oc, output);
+	bool ok = apply_output_configs(configs, configs_len);
```
emersion, line 610: If `test_only` is true we should use `wlr_backend_test()` instead of `wlr_backend_commit()`…

sway/desktop/output.c
kennylevinsen: Ah yeah, need to find the old output config is none is included in the head. In that case we also need to make sure we don't store the found config in that case - that would be in issue in the output config change PR, where merge happens in find_output_config.

Will fix.

sway/desktop/output.c
```diff
@@ -550,63 +550,73 @@ void handle_gamma_control_set_gamma(struct wl_listener *listener, void *data) {
 	wlr_output_schedule_frame(output->wlr_output);
 }
 
+static struct output_config *output_config_for_config_head(
+	struct wlr_output_configuration_head_v1 *config_head,
+	struct sway_output *output) {
```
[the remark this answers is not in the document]
kennylevinsen: Fixed

sway/desktop/output.c
kennylevinsen: It now finds an existing config is no new one is provided, and does a check later to see if we need to store the config.

kennylevinsen:
I took the liberty of shortening some lines in the new apply_output_configs while I was addressing the comments - they had gotten a bit out of hand.

sway/desktop/output.c
kennylevinsen: Indeed, now there's a combined check for the cases where we need to store a new config.

sway/desktop/output.c
kennylevinsen: We now skip the render, commit and finalize on test-only. We shouldn't need to call `wlr_backend_test`, as the swapchain manager did that for us.

emersion:
By the way, we have some `wlr_output_test()` calls in `queue_output_config()` to lower the mode/BPC automatically. I don't think this plays very nice with the new logic? (Not saying we should necessarily fix it right now)

emersion:
I've looked through all of the commits and they all LGTM!

kennylevinsen:
> By the way, we have some wlr_output_test() calls in queue_output_config() to lower the mode/BPC automatically. I don't think this plays very nice with the new logic? (Not saying we should necessarily fix it right now)

Agree. In clay I retry with a new set of output states that progressively get less demanding until something sticks. While that approach isn't per-output like the old logic, I think we could have some fairly simple "graceful degradation" logic here too.

> I've looked through all of the commits and they all LGTM!

Thanks for the review!

sway/desktop/output.c
```diff
@@ -521,9 +521,7 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 		sway_session_lock_add_output(server->session_lock.lock, output);
 	}
 
-	struct output_config *oc = find_output_config(output);
-	apply_output_config(oc, output);
-	free_output_config(oc);
+	apply_all_output_configs();
```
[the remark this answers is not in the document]
kennylevinsen: In clay I run off an idle handler for new and destroyed outputs, and for `wlr_output_power_management` as those do not need any feedback. I also experimented with a timer, but it didn't seem to make much of a difference other than delaying bringup. Might be worth more experimentation.

It's a bit trickier when the trigger want feedback on success, e.g. dynamic configuration with `swaymsg` or `wlr_output_management`.

kennylevinsen:
re: our discussion about graceful fallback, I threw together something here: https://github.com/swaywm/sway/commit/18a1c26e7494e451d860ad0e4e235450aad09849

It is untested, but the idea is to first degrade render formats to 8 bit, then degrade automatically selected modes, starting with the most bandwidth-intensive one.

Problems:
- If one output supports 10-bit XRGB, while another 10-bit XBGR, then the current render format degradation will drop all the way to 8-bit as there is no per-output test.
- The VRR issue is not yet handled. It's not a fallback in the same sense and might need a different strategy.

sway/desktop/output.c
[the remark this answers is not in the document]
emersion: Ref https://github.com/swaywm/sway/pull/8091 :)

> It's a bit trickier when the trigger want feedback on success, e.g. dynamic configuration with swaymsg or wlr_output_management.

I don't think debouncing is particularly desirable in these cases.

emersion (approved):
Let's merge this for now and keep improving the fallback stuff in separate PRs.

Thanks a lot!

PR #8069, opened 2024-03-16
config/output: Refactor handling of tiered configs

kennylevinsen:
Output configuration can be applied to a particular output in three
ways: As a wildcard, by connector name and by identifier. This in turn
means that three different configurations must be handled at any given
time.

In the current model, this is managed by merging new configuration into
every other matching configuration. At the same time, an additional
synthetic configuration is made which matchehes both identifier and name
at the same time, further complicating logic.

Instead, manage and store each configuration independently and merge
them in order when retrieving configuration for an output. When changes
are made to a less specific configuration, clear these fields from more
specific configurations to allow the change to take effect regardless of
precedence.

Depends on: https://github.com/swaywm/sway/pull/8068

/cc @emersion

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -79,6 +79,71 @@ struct output_config *new_output_config(const char *name) {
 	return oc;
 }
 
+// supersede_output_config clears all fields in dst that were set in src
+static void supersede_output_config(struct output_config *dst, struct output_config *src) {
+	if (src->enabled != -1) {
+		dst->enabled = -1;
+	}
+	if (src->width != -1) {
+		dst->width = -1;
+	}
+	if (src->height != -1) {
+		dst->height = -1;
+	}
+	if (src->x != -1) {
+		dst->x = -1;
+	}
+	if (src->y != -1) {
+		dst->y = -1;
+	}
+	if (src->scale != -1) {
+		dst->scale = -1;
+	}
+	if (src->scale_filter != SCALE_FILTER_DEFAULT) {
+		dst->scale_filter = SCALE_FILTER_DEFAULT;
+	}
+	if (src->subpixel != WL_OUTPUT_SUBPIXEL_UNKNOWN) {
+		dst->subpixel = WL_OUTPUT_SUBPIXEL_UNKNOWN;
+	}
+	if (src->refresh_rate != -1) {
+		dst->refresh_rate = -1;
+	}
+	if (src->custom_mode != -1) {
+		dst->custom_mode = -1;
+	}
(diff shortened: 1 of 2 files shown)


kennylevinsen:
Made `store_output_config` a bit prettier with better comments, and moved the function documentation to the header file.

sway/config/output.c
```diff
@@ -142,96 +207,45 @@ static void merge_output_config(struct output_config *dst, struct output_config
...
-	char *id_on_name = format_str("%s on %s", id, name);
-	if (!id_on_name) {
-		return;
+	bool merged = false;
+	bool wildcard = strcmp(oc->name, "*") == 0;
+	struct sway_output *output = wildcard ? NULL : output_by_name_or_id(oc->name);
+	if (!output && !wildcard) {
+		// There is no config by this name, just add it in
+		goto done;
 	}
```
emersion, line 157: Hm, why do we need to grab the sway_output here? Won't the loop below do the right thing regardless?

sway/config/output.c
```diff
@@ -142,96 +207,45 @@ static void merge_output_config(struct output_config *dst, struct output_config
...
-			sway_log(SWAY_DEBUG, "Merging on top of output * config");
-			struct output_config *current = new_output_config(oc->name);
-			merge_output_config(current, config->output_configs->items[i]);
-			merge_output_config(current, oc);
-			free_output_config(oc);
-			oc = current;
+		// If the new config matches an output's name, and the old config
+		// matches on that output's identifier, supersede it.
+		if (strcmp(old->name, id) == 0 &&
+			strcmp(oc->name, output->wlr_output->name) == 0) {
```
emersion, line 182: Style nit: missing one tab here

emersion:
"config/output: Compress merge_output_config and co": not sure this is an improvement in readability. But I don't care too much.

sway/config/output.c
kennylevinsen: We need to know whether the output config matches an output's identifier or name, which requires us to look up the output. This info is needed to know which configs to supersede, with less specific configs wiping out settings in more specific configs in order for them to take effect.

I see that I had lost the call to `output_get_identifier` in a refactor, added back in.

kennylevinsen:
> "config/output: Compress merge_output_config and co": not sure this is an improvement in readability. But I don't care too much.

Yeah I wasn't sure either, just felt like those functions were a bit of an eye-sore so tried *something*. Removed.

emersion (approved):
LGTM, thank you!

PR #8070, opened 2024-03-17
Update for versioned wlroots files

vimproved:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4614

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -45,7 +45,7 @@ subproject(
 	required: false,
 	version: wlroots_version,
 )
-wlroots = dependency('wlroots', version: wlroots_version)
+wlroots = dependency('wlroots-0.18', version: wlroots_version, fallback: 'wlroots')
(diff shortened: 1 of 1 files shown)


bl4ckb0ne:
Does this handles using wlroots as a subproject? sway's master branch tracks wlroots master, we use a tagged version for releases.

emersion:
Maybe we need to add `fallback: 'wlroots'`?

emersion:
Can we add the `fallback`? Otherwise this PR breaks subprojects.

emersion:
I would prefer not to specify the `fallback` as an array, because this punches into the subproject's internal variables.

See https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4698

emersion (approved):
Thanks!

PR #7889, opened 2023-12-29
add --smart-titles flag to hide_edge_borders

nim65s:
Hi, 

This extend the `--i3` flag to hide title bars on any kind of container.

Issue #8075, opened 2024-03-18
waybar is not shown sometimes on master branch


svalaskevicius:
- **Sway Version:**
  - branch based on master commit 23389ebd

- **Configuration File:**
relevant bit:
```
bar bar1 {
    id bar1
    swaybar_command waybar -b bar1
    mode hide
    modifier $mod
}

exec waybar -b bar1
```

- **Description:**

when waybar is configured to be shown only on $mod key press, it fails to appear after a window is created in that workspace.

switching to another workspace and hitting the $mod summons waybar as expected. even if I switch back to the original workspace, waybar is still shown while I keep $mod pressed. pressing it again does not show waybar in this workspace. no logs nor errors are reported when the key is pressed.

svalaskevicius:
forgot to add - swaybar is working ok with regards to this behaviour. so my assumption was that this is related to rendering the top shell layer. but I wasn't able to find the fault in the short time I had

svalaskevicius:
update from commit 4d4c88f0  - waybar is showing now, but the region is not repainted after mod key is released and waybar is hidden - that region is black, until the underlying window invalidates it or parts of it (then it repaints parts)

Nefsen402:
I cannot reproduce this issue even when 4d4c88f was reverted on my local branch. I think there might me more to your configuration file? It sounds like the new symptoms have something to do with damage tracking, which at this point is unlikely since the renderer rewrite of sway. A demo video and a list of steps to reproduce would help.

svalaskevicius:
sure, what should I use to record the screen / what other info can I provide?

svalaskevicius:
my cfgs are also https://github.com/svalaskevicius/dotfiles/blob/gen-2024/.config/sway/config except the hide lone titlebar flag which should be commented out on this commit


there is also waybar config in a nearby file

svalaskevicius:
also, it doesnt happen _all_ the time - but very frequently. especially after:
1. hold mod - waybar appears
2. press number to switch workspace
3. release mod  - waybar disappears, with black background, until the underlying window invokes a repaint

Nefsen402:
Yeah, I've switched workspaces many times in my testing and have not seen it go wrong once. I am testing on the wayland backend with a nested session, but the backend should have nothing to do with this error. What kind of hardware are you running? Maybe it's a driver problem causing the screen not to repaint correctly? What version of waybar?

svalaskevicius:
sorry I missed the question.

[...]

Issue #8072, opened 2024-03-18
`hide_edge_borders --i3 smart` not hiding titlebar on tabbed/stacked containers anymore


nim65s:
- **Sway Version:**

1.9

- **Debug Log:**

https://paste.inpt.fr/?b020746019e5ca87#C1EYwxbmPZvtuiFZQkF6n1Lghxk9iTSnKitiKm5frZhW

- **Configuration File:**
```
hide_edge_borders --i3 smart
```

- **Description:**
  - the `--i3` flag is supposed to hide the titlebar on tabbed/stacked containers, ref. https://github.com/swaywm/sway/pull/3767. This used to work before 1.9, but is not working anymore on 1.9.

svalaskevicius:
I had some local commits to make this working and got to somewhat functioning state but I don't use master branch yet due to waybar not being drawn (shell top layer rendering issues it seems, and I don't have time to debug yet)

can make a pr when I'm back home

svalaskevicius:
@nim65s please see the linked PR ^, the main file I was focusing on is `transaction.c`

I hope this helps! it is really a needed feature of sway and I hope it would be merged to the main branch sooner rather than later. and so avoiding to keep rebasing the PR :)

hashworks:
@svalaskevicius You ever got around to that PR? :)

svalaskevicius:
hi. see #8076 but it has an issue that needs debugging - any help is appreciated as I'm not sure when can I allocate more time to it

tekstryder:
Still an issue on latest sway/wlroots git.

- sway @ https://github.com/swaywm/sway/commit/f4aba22582184c9a4a20fd7a9ffd70c63b4b393d 
- wlroots @ [abf80b5](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/abf80b529e48823e21215a6ccc4653e2c2a4a565)

ivanesmantovich:
Still an issue

PR #8076, opened 2024-03-18
Hiding lone titlebar scenegraph

svalaskevicius:
this is a WIP PR, for comments and hopefully others joining in taking this feature onwards :) 

the render.c file was removed and the logic is switching to the scenegraph based implementation.

during my brief testing this seems to have worked, but the testing was limited as I got stumbled on another issue of the master branch - waybar was not showing after I create windows (#8075) so I reverted to the previous version for now.

a step towards resolving #8072 - hopefully :)

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -404,8 +404,14 @@ void view_autoconfigure(struct sway_view *view) {
 			height = con->pending.height - y_offset
 				- con->pending.border_thickness * con->pending.border_bottom;
 		} else {
-			y = con->pending.y + container_titlebar_height();
-			height = con->pending.height - container_titlebar_height()
+			int titlebar_height;
+			if (config->hide_lone_title && view_is_only_visible(view)) {
+				titlebar_height = con->pending.border_thickness * con->pending.border_top + y_offset;
+			} else {
+				titlebar_height = container_titlebar_height();
+			}
+			y = con->pending.y + titlebar_height;
+			height = con->pending.height - titlebar_height
 				- con->pending.border_thickness * con->pending.border_bottom;
 		}
 		break;
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -569,6 +569,7 @@ struct sway_config {
 	enum edge_border_types hide_edge_borders;
 	enum edge_border_smart_types hide_edge_borders_smart;
 	bool hide_lone_tab;
+	bool hide_lone_title;
 
 	// border colors
 	struct {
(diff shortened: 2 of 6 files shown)


include/sway/tree/view.h
```diff
@@ -320,6 +320,8 @@ void view_execute_criteria(struct sway_view *view);
  */
 bool view_is_visible(struct sway_view *view);
 
+bool view_is_only_visible(struct sway_view *view);
+
```
nim65s, line 324: this change is no longer needed

sway/tree/view.c
```diff
@@ -229,7 +229,7 @@ bool view_ancestor_is_only_visible(struct sway_view *view) {
 	return only_visible;
 }
 
-static bool view_is_only_visible(struct sway_view *view) {
+bool view_is_only_visible(struct sway_view *view) {
```
nim65s, line 232: this change is no longer needed

nim65s (commented):
Thanks @svalaskevicius ! This is working for me, so I guess we could close #7889 in favor of this one.

include/sway/tree/view.h
svalaskevicius: removed, thanks :)

nim65s:
I found a little bug: when I open a first transparent windows (my terminal, kitty, in the current case), it doesn't have a title bar, which is expected. Then, I open a second one, and I get 2 side-by-side windows, both with titlebar, which is still expected. When I close one of the 2, the title bar of the lone window disappear, which is still the correct behavior.

But the space taken by the titlebar of the left window doesn't get its background back, and I have a black rectangle in place of my wallpaper there, on the top left corner of my screen. 

It's minor, as the windows itself is behaving as expected, and only the transparent background is not updated, but we might be missing something here. I guess it's not kitty's fault, because this issues did not appear on #7889.

svalaskevicius:
@nim65s I've rebased on upstream master and it seems I cannot reproduce your scenario anymore - please can you check?

nim65s:
I still have the same issue :/

iNecas:
Btw. I'm using this patch for some time now and it's really great feature. The only strange things seems with the behavior of `--smart-titles` flag, in my experience (and the code seems to confirm that),  `hide_edge_borders --i3 smart` seems enough to activate the feature. This would mean that one would not be able to have the previous behavior and it seem i3 is not hiding the titles (just guessiong, it's been a long time so can't tell). Also, command help at https://github.com/swaywm/sway/blob/6959a78a8f0d52f79ad7465135b3295307a5146a/sway/commands/hide_edge_borders.c#L7 would need to be updated.

Issue #8084, opened 2024-03-23
Sway does not recognize 'Super' as a modifier key name


Ferdi265:
## Description

Sway currently supports only the XKB modifier names and a few custom additions so that the familiar names "Ctrl" and "Alt" work as expected (as well as adding "Mod3" and "Mod5" that aren't defined as `XKB_MOD_NAME_*` macros).

What's missing is any human-readable name for the Logo (or Super) key, which is an issue that has confused many people starting out with sway that I know. Even the default sway config has to go out of its way to comment `# Logo key.` before `set $mod Mod4` in order to not confuse potential first-time users.

## Desired Feature

Allow using "Super", "Meta", or "Logo" as a modifier name for the logo key as well, instead of only "Mod4".

If that whole list is a little much, maybe decide on at least one human-readable name for the modifier key instead of just "Mod4".

## Prior Art

- Linux `input-event-codes.h` calls the logo key META
- Gnome calls the logo key "Super"
- KDE calls the logo key "Meta"
- Hyprland calls the logo key "SUPER", "LOGO", "WIN", or "MOD4"
- i3 calls the logo key "Mod4"

## Prior Art for deviating from what i3 does

Sway also supports "Alt" as a modifier name, despite i3 not supporting that.
This was implemented along with the "Ctrl" alias (which i3 does support) in response to #153.

WhyNotHugo:
Fixed in bc258a3be2f946c1c93bcbe40735b2db068e0ea8

Ferdi265:
> Fixed in [bc258a3](https://github.com/swaywm/sway/commit/bc258a3be2f946c1c93bcbe40735b2db068e0ea8)

right. I forgot the commit didn't have a "Closes" tag; Closed.

Issue #153, opened 2015-08-30
Control key is not recognized


aouelete:
I have this in my config file:
`bindsym Ctrl+$mod+Right workspce next`
and I get this error when starting sway: 
`bindsym - unknown key Ctrl`

aouelete:
Thanks, looks like it is fixed

PR #8085, opened 2024-03-23
input: add Super as alternative for Mod4

Ferdi265:
This PR implements alternative human-readable names for the logo key (Mod4) as proposed in #8084.

Implemented in this PR are the names "Super", "Meta", and "Logo".

This is the first time the supported modifier names have been touched since #153, almost 9 years ago (besides moving the code around into different files and switching from wlc to wlroots).

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -32,6 +32,7 @@ static struct modifier_key {
 	{ XKB_MOD_NAME_NUM, WLR_MODIFIER_MOD2 },
 	{ "Mod3", WLR_MODIFIER_MOD3 },
 	{ XKB_MOD_NAME_LOGO, WLR_MODIFIER_LOGO },
+	{ "Super", WLR_MODIFIER_LOGO },
 	{ "Mod5", WLR_MODIFIER_MOD5 },
 };
 
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -403,8 +403,8 @@ runtime.
 	For specifying modifier keys, you can use the XKB modifier names _Shift_,
 	_Lock_ (for Caps Lock), _Control_, _Mod1_ (for Alt), _Mod2_ (for Num Lock),
 	_Mod3_ (for XKB modifier Mod3), _Mod4_ (for the Logo key), and _Mod5_ (for
-	AltGr). In addition, you can use the aliases  _Ctrl_ (for Control) and _Alt_
-	(for Alt).
+	AltGr). In addition, you can use the aliases  _Ctrl_ (for Control), _Alt_
+	(for Alt), and _Super_ (for the Logo key).
 
 	Unless the flag _--locked_ is set, the command will not be run when a
 	screen locking program is active. If there is a matching binding with


emersion:
I'd be fine with adding a more user-friendly name, but would prefer not to add a whole bunch of aliases that mean the same thing. My preference would be to follow the libxkbcommon naming here.

Also this needs a man page update as well.

Ferdi265:
Right. So just "Logo", plus documentation updates IIUC.

I expected this and I'm fine with any of these, I mostly made this PR so broad so all options are on the table.

Will adjust accordingly.

Ferdi265:
updated to allow only "Logo" instead of "Super", "Meta", and "Logo".

I haven't found any explanation of modifier names in the man pages except a few examples. Would you accept a separate PR adding such a section that explains all of the supported modifier names or would you like this to be done in this PR? Should we update examples in the man pages or config to use symbolic names where possible?

(The branch name doesn't fit any more, but that's on me, and renaming a branch while keeping the PR attached doesn't work IIRC, so so be it.)

Ferdi265:
ahh, sorry, I was confused by `<xkbcommon-names.h>` calling the key "LOGO", while the rest of xkbcommon seems to call it Super or Meta. I'm not 100% sure which one to take.

I'm leving this choice to you. I'm fine with either of them, so please tell me which one you prefer.

Ferdi265:
Modifier documentation is now in branch [feature-modifier-docs](https://github.com/swaywm/sway/compare/master...Ferdi265:sway:feature-modifier-docs). I added it to the documentation for `bindsym`, which is the only place I found that explicitly talks about XKB keysyms with regards to keybindings.

rpigott:
Is there a way we can just get the virtual modifier names from the keymap to alias their real modifiers? That would basically implement this then, right?

emersion:
> Is there a way we can just get the virtual modifier names from the keymap to alias their real modifiers? That would basically implement this then, right?

You mean `xkb_keymap_mod_get_name()`? This returns the names used by `XKB_MOD_NAME_*` and we already have these listed.

> I haven't found any explanation of modifier names in the man pages except a few examples. Would you accept a separate PR adding such a section that explains all of the supported modifier names or would you like this to be done in this PR? Should we update examples in the man pages or config to use symbolic names where possible?

Oh, sorry, I thought we had these documented already but it doesn't seem like it. Either is fine!

Ferdi265:
> > I haven't found any explanation of modifier names in the man pages except a few examples. Would you accept a separate PR adding such a section that explains all of the supported modifier names or would you like this to be done in this PR? Should we update examples in the man pages or config to use symbolic names where possible?
> 
> Oh, sorry, I thought we had these documented already but it doesn't seem like it. Either is fine!

I made a new PR documenting the existing modifiers (#8087) and rebased this one on top of that in addition to documenting the Super modifier here.

Intended merge order is #8087 followed by #8085 (this PR).

rpigott:
Super, Meta, etc. are virtual modifier names. They're at least present in the (standard US) keymap I can extract from xwayland, along with their real modifier associations. So, IIUC, it is possible by manual inspection of the keymap to map those names uniquely to real modifiers (Mod1, Mod4). I don't see any xkbcommon api related to virtual modifiers though, so maybe I haven't fully understood their purpose.

Ferdi265:
> Yeah, I can see e.g. "Super" defined here:
> 
>     * https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/blob/022c3ca1db12e505cbd5ce8bf19c432d6a70c7e5/types/pc#L22
> 
>     * https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/blob/022c3ca1db12e505cbd5ce8bf19c432d6a70c7e5/compat/misc#L56
> 
> 
> But yeah, I don't see any way to extract this string from the XKB API.

`xkb_keymap_num_mods(keymap)` returns the number of modifiers and with `de` and `us` I see the virtual modifier names starting at modifier index 8. I don't know how to get the mappings to real modifiers though.

rpigott:
Seems like XKB (OG X Keyboard Extension) had something like this [1][2]

[1] https://man.archlinux.org/man/XkbGetVirtualMods.3
[2] https://man.archlinux.org/man/XkbVirtualModsToReal.3

Ferdi265:
rebased since #8087 was merged

emersion (approved):
Thanks!

WhyNotHugo:
Thank you!

PR #8091, opened 2024-03-27
desktop/output: Debounce modesets

kennylevinsen:
Add a 10 millisecond modeset debounce timer.

Depends on https://github.com/swaywm/sway/pull/8095

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -362,6 +362,26 @@ static void update_output_manager_config(struct sway_server *server) {
 	ipc_event_output();
 }
 
+static int timer_modeset_handle(void *data) {
+	struct sway_server *server = data;
+	wl_event_source_remove(server->delayed_modeset);
+	server->delayed_modeset = NULL;
+
+	apply_all_output_configs();
+	transaction_commit_dirty();
+	update_output_manager_config(server);
+
+	return 0;
+}
+
+static void request_modeset(struct sway_server *server) {
+	if (server->delayed_modeset == NULL) {
+		server->delayed_modeset = wl_event_loop_add_timer(server->wl_event_loop,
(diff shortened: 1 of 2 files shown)


emersion:
Before merging this we need to add fallback logic to disable outputs one by one (or else we risk ending up with no output lit up at all).

sway/desktop/output.c
```diff
@@ -385,9 +405,7 @@ static void begin_destroy(struct sway_output *output) {
 	output->wlr_output->data = NULL;
 	output->wlr_output = NULL;
 
-	transaction_commit_dirty();
-
-	update_output_manager_config(server);
+	request_modeset(server);
```
Nefsen402, line 408: Why do we want to modeset an output that we are in the process of destroying?

emersion: We might be able to light up an output that the user wanted to light up but we were missing a CRTC for, I think?

Nefsen402: In that case, wouldn't we want to move the modeset out of `begin_destroy` and into the place where the output is actually destroyed?

kennylevinsen: Yeah, we aren't modesetting *this* output, we're modesetting the entire backend in case this output was degrading other outputs (e.g., the usual bandwidth limits on Intel or not enough CRTCs).

> In that case, wouldn't we want to move the modeset out of begin_destroy and into the place where the output is actually destroyed?

`begin_destroy` seemed like the counterpart to `handle_new_output`, with them both ending in a similar epilogue - apply output configuration or evacuate, commit transaction, update `wlr_output_manager_v1`.

Just seemed easy to break both of those out into the same idle handler, and didn't think too much of it otherwise.

Nefsen402 (approved):

PR #8096, opened 2024-03-29
sway-ipc(7): Escape backslashes correctly in GET_CONFIG output

dkg:
Without this change, i see the following in the sway-ipc manpage:

```

   9. GET_CONFIG
       MESSAGE
       Retrieve the contents of the config that was last loaded

       REPLY
       An object with a single string property containing the contents of  the
       config

       Example Reply:
           {
                "config": "set $mod Mod4nbindsym $mod+q exitn"
           }
```


emersion (approved):
Thanks!

PR #8099, opened 2024-04-01
new directive `include_one` 

yashsinghcodes:
It would include all the files in the first directory and then from the second directory it would only import the files that weren't import before. So if the first directory contains a file called for example, "70_bar" then such file would be skipped when processing the second directory.

So currently I have only tested it my building the sway using following steps:

```
meson build/
ninja -C build/
```

I will mark it as draft for further testing on the command.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -144,6 +144,7 @@ sway_cmd cmd_fullscreen;
 sway_cmd cmd_gaps;
 sway_cmd cmd_hide_edge_borders;
 sway_cmd cmd_include;
+sway_cmd cmd_include_one;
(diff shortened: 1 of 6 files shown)


emersion:
Note, we're not accepting new config-related features in general: https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#scope-of-future-changes-to-sway

cfsmp3:
> Note, we're not accepting new config-related features in general: https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#scope-of-future-changes-to-sway

Would this diff be accepted if a similar one was first merged into i3?

This a really nice feature.

PR #8101, opened 2024-04-03
container: Only give a titlebar height for normal borders

### LINKED ISSUE #7409: Hide the titlebar

lanodan:
Closes: https://github.com/swaywm/sway/issues/7409

### DIFF
--- a/include/sway/tree/container.h
+++ b/include/sway/tree/container.h
@@ -193,7 +193,7 @@ void container_update_representation(struct sway_container *container);
 /**
  * Return the height of a regular title bar.
  */
-size_t container_titlebar_height(void);
+size_t container_titlebar_height(struct sway_container *container);
 
 void floating_calculate_constraints(int *min_width, int *max_width,
 		int *min_height, int *max_height);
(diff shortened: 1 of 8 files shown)


lanodan:
Note that while this respects the documented configuration, interaction with windows having other border settings is quite a mess.
screen

I guess sway should be forcing borders on childs of non-toplevel container with multiple childs.


Probably what's attempted at there but failing because a workspace always has a container (and title height being queried in a lot of other places): https://github.com/swaywm/sway/blob/dcb142bf5e390250939544075b5852ca21eaf721/sway/desktop/transaction.c#L288-L299

emersion:
Hm, why is this marked as fixing #7409? AFAIU that issue is about completely disabling titlebars everywhere. This PR doesn't implement this feature, right?

PR #8109, opened 2024-04-10
config/input: Implement libinput clickfinger_button_map option support.

tarcisioe:
Opening this as a draft since it depends on an unreleased libinput feature.

I recently closed [libinput issue #909](https://gitlab.freedesktop.org/libinput/libinput/-/issues/909) by [implementing](https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/988) a "clickfinger_button_map" option (mirrors "tap_button_map", but for clickmethod "clickfinger").

This implements support for setting such option in Sway. However, this would bump the required libinput version to 1.26, and the dependency comes through wlroots, which supports 1.14 and up. So maybe this code should be conditionally enabled depending on the libinput version during build, and I'm not sure if there is any project policy about this.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -249,6 +249,7 @@ sway_cmd input_cmd_seat;
 sway_cmd input_cmd_accel_profile;
 sway_cmd input_cmd_calibration_matrix;
 sway_cmd input_cmd_click_method;
+sway_cmd input_cmd_clickfinger_button_map;
 sway_cmd input_cmd_drag;
 sway_cmd input_cmd_drag_lock;
 sway_cmd input_cmd_dwt;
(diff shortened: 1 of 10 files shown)


emersion (approved):

tarcisioe:
@emersion since libinput 1.26 is out now, I'd like to see if we can merge this. However as I mentioned previously, I believe this means depending on 1.26. Would that be an issue for you guys? Maybe we can add some kind of version check/conditional compilation if a hard bump on the requirement is undesired.

emersion:
Yeah, I think it's fine to depend on 1.26.

tarcisioe:
Alright I'll rebase this, update the PR and remove draft.

emersion:
Thanks!

emersion:
Oh, I forgot - could we bump the minimum libinput version in `meson.build`?

tarcisioe:
Sorry, I forgot as well. #8236

PR #8112, opened 2024-04-12
tree/container: ensure pixman rect is valid in container_arrange_title_bar()

emersion:
Fixes "Invalid rectangle passed" errors printed by Pixman.


Nefsen402:
The reason why `sway_text_node_set_max_width()` with 0 doesn't work is because it will internally generate a `wlr_rect` with 0 area. `wlr_scene` special cases these rects and interprets them as having a source box of the entire input texture.

As a fixup, this MR LGTM.

Nefsen402:
Fixed up the max_width thing with https://github.com/swaywm/sway/pull/8140

Issue #8128, opened 2024-04-23
Scale gets reset after output power cycle when output_manager (kanshi) is in use


baltitenger:
**Sway Version:** sway version 1.10-dev-896d39bd3 (Apr 22 2024, branch 'master')

**Debug log:** http://0x0.st/XoCA.txt

**Configuration File:** default.

**Description:** If kanshi is configured to set a scale for an output, power cycling it (e.g. via `swaymsg output * power {off,on}` resets the scale to 1.

Kanshi config used:
```
profile laptop {
	output eDP-1 scale 1.25
}
```

<details><summary>Kanshi output with WAYLAND_DEBUG=1</summary>

```[3161855.407]  -> wl_display@1.get_registry(new id wl_registry@2)
[3161855.466]  -> wl_display@1.sync(new id wl_callback@3)
[3161855.758] wl_display@1.delete_id(3)
[3161855.786] wl_registry@2.global(1, "wl_shm", 1)
[3161855.795] wl_registry@2.global(2, "zwp_linux_dmabuf_v1", 4)
[3161855.801] wl_registry@2.global(3, "wl_compositor", 6)
[3161855.807] wl_registry@2.global(4, "wl_subcompositor", 1)
[3161855.813] wl_registry@2.global(5, "wl_data_device_manager", 3)
[3161855.820] wl_registry@2.global(6, "zwlr_gamma_control_manager_v1", 1)
[3161855.825] wl_registry@2.global(7, "zxdg_output_manager_v1", 3)
[3161855.831] wl_registry@2.global(8, "ext_idle_notifier_v1", 1)
[3161855.837] wl_registry@2.global(9, "zwp_idle_inhibit_manager_v1", 1)
[3161855.843] wl_registry@2.global(10, "zwlr_layer_shell_v1", 4)
[3161855.849] wl_registry@2.global(11, "xdg_wm_base", 5)
[3161855.855] wl_registry@2.global(12, "zwp_tablet_manager_v2", 1)
[3161855.860] wl_registry@2.global(13, "org_kde_kwin_server_decoration_manager", 1)
[3161855.867] wl_registry@2.global(14, "zxdg_decoration_manager_v1", 1)
[3161855.873] wl_registry@2.global(15, "zwp_relative_pointer_manager_v1", 1)
[3161855.879] wl_registry@2.global(16, "zwp_pointer_constraints_v1", 1)
[3161855.885] wl_registry@2.global(17, "wp_presentation", 1)
[...]

PR #8129, opened 2024-04-23
desktop/output: Fix check if config should be stored

kennylevinsen:
We want to check if a config_head existed for the current matched_output_config, so we should check cfg->output. sway_output is a temporary variable from a previous wl_list_for_each, and does not contain anything useful to us.

Fixes: https://github.com/swaywm/sway/issues/8128


emersion (approved):
Good catch, thanks!

Issue #8141, opened 2024-05-01
`swaymsg output eDP-2 enable` fails to re-enable internal laptop display after manually disabling it


colemickens:
### Please fill out the following:
- **Sway Version:**
646019c

- **Debug Log:**
https://gist.github.com/colemickens/a162d462c1c20f711aeeca00fbdf47d7

- **Configuration File:**
https://gist.github.com/colemickens/cc1c962daff934f594c652738cf816fb

(more or less the same config I've been using for ages)

- **Stack Trace:**
n/a

- **Description:**
   - plug in dock with an external display attached
   - kanshi fires up and enables the display
   - I manually disable the eDP-2 internal laptop display
   - I later try to manually enable the eDP-2 internal display, by running `swaymsg output eDP-2 enable`.

kennylevinsen:
> connector DP-1: Failed to enable adaptive sync: connector doesn't support VRR

Can you try with https://github.com/swaywm/sway/pull/8095?

PR #8142, opened 2024-05-01
config/output: Use all outputs for config merge

kennylevinsen:
When storing a config, we need to find the output that is being configured to extract its identifier. output_by_name_or_id does not return outputs that are disabled, and using this makes it impossible to merge configurations related to disabled outputs.

Switch to all_outputs_by_name_or_id.

Fixes: https://github.com/swaywm/sway/issues/8141


bl4ckb0ne (approved):
LGTM

PR #8155, opened 2024-05-09
remove unnecessary code

khufu0:
config_path's already a NULL pointer, no need to free it.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -229,6 +229,9 @@ int main(int argc, char **argv) {
 			exit(EXIT_SUCCESS);
 			break;
 		case 'c': // config
+			if (config_path != NULL) {
+				free(config_path);
+			}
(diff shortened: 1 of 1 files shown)


bl4ckb0ne:
Calling `free` on `NULL` is a no-op, doesn't hurt.

khufu0:
maybe add a guard if statement to check if its not equal to NULL for readability.

bl4ckb0ne:
As stated, it's a no-op. Why bother with a statement?

bl4ckb0ne:
From `man 3 free` [1]

       The free() function shall cause the space pointed to by ptr to be
       deallocated; that is, made available for further allocation. If
       ptr is a null pointer, no action shall occur.

Adding the `if` statement there is not useful.

[1]: https://man7.org/linux/man-pages/man3/free.3p.html

PR #8165, opened 2024-05-18
build: drop xwayland option

emersion:
Instead of having a build-time option to enable/disable xwayland support, just use the wlroots build config: enable xwayland in Sway if it was enabled when building wlroots. I don't see any use-case for disabling xwayland in Sway when enabled in wlroots: Sway doesn't pull in any additional dependency (just pulls in dependencies that wlroots already needs). We have a config command to disable xwayland at runtime anyways.

This makes it so xwayland behaves the same way as other features such as libinput backend and session support. This also reduces the build matrix (less combinations of build options).

I think we originally introduced the xwayland option when we didn't have a good way to figure out the wlroots build config from the Sway build system.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -38,8 +38,13 @@ tasks:
       cd sway
       ninja -C build
   - build-no-xwayland: |
-      cd sway
+      cd wlroots
       meson configure build -Dxwayland=disabled
+      ninja -C build
+      sudo ninja -C build install
+
+      cd ../sway
+      meson configure build --clearcache
       ninja -C build
   - build-static: |
       cd sway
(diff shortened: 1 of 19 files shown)


bl4ckb0ne:
You're missing a CI change https://github.com/swaywm/sway/blob/master/.builds/alpine.yml#L40

bl4ckb0ne (approved):
Thanks!

prometheanfire:
:( https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies#Why_automagic_dependencies_are_wrong  (not that I don't understand the desire to reduce the build matrix)

For reference I've kept the explicit xcb dep when using the `X` flag (along with pulling in `wlroots[X]` of course).

emersion:
It's not "automagic":  the build system doesn't check whether the dependency is available and then enables/disables the feature based on that. Instead, it forwards the wlroots xcb dependency if wlroots was built with xwayland support.

prometheanfire:
ah, makes sense

PR #8166, opened 2024-05-18
sway/config/output.c: fix null deref on output config

navi-desu:
If there's no config for the output, oc is null, but some screens might have a default rotation, causing the log call to dereference a null pointer.


emersion (approved):
Thanks!

PR #8174, opened 2024-05-22
Allow configuring unfocused tab title

shengdichen:
Hi sway devs:

This pull request allows configuration of the unfocused tab/stack container title (shortened to tab-title henceforth) independent from the generic unfocused tab-title. The resultant configuration option is called `client.unfocused_tab_title`.

A driving motivation for this is the quick visual distinction of an (unfocused) view and an (unfocused) container. 

The implementation closely mirrors that of the existing option of the _focused_ tab-title `client.focused_tab_title`. The documentation update is made in a separate, second commit.

Thanks for looking over this!

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -116,6 +116,7 @@ sway_cmd cmd_client_focused;
 sway_cmd cmd_client_focused_inactive;
 sway_cmd cmd_client_focused_tab_title;
 sway_cmd cmd_client_unfocused;
+sway_cmd cmd_client_unfocused_tab_title;
 sway_cmd cmd_client_urgent;
 sway_cmd cmd_client_placeholder;
 sway_cmd cmd_client_background;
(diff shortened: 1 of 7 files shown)


emersion:
Hi, thanks for sending a patch! Does this option exist in i3 as well, or is this something completely new? Note that in general our focus is not to add new window management features, see https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#scope-of-future-changes-to-sway

shengdichen:
Hi, thanks for the quick response!

No, sadly the option proposed here does not exist in i3 currently. I submitted this patch since:
- it was not a feature expansion to the core window management facilities;
- it seemed like a natural (to me) extension of an existing option.

I would be glad to see this merged, but of course I understand and respect sway's mission statement. In the case of not merging, should I try to submit a patch to i3 and see if they would accept this feature? And if it gets merged in i3, will this be open for discussion again here in sway?

Thanks!

PR #8178, opened 2024-05-24
server: hide xdg_output from unprivileged clients

emersion:
Regular Wayland clients shouldn't care about the position or size of outputs. Hide xdg_output from unprivileged clients to make sure they're not doing shenanigans with this information.


bl4ckb0ne (approved):
People are gonna complain and im fine with that. LGTM!

PR #8179, opened 2024-05-24
input/text_input: ensure keyboard is set before sending modifiers

emersion:
Clients get confused when modifier events are sent before the keymap.


bl4ckb0ne (approved):
LGTM

PR #8185, opened 2024-05-27
input/keyboard: don't send key release if we don't have focused surface

llyyr:
"The compositor must not send this event if the wl_keyboard did not have an active surface immediately before this event. The compositor must not send this event if state is pressed (resp. released) and the key was already logically down (resp. was not logically down) immediately before this event."

From https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/382


Nefsen402:
Ah, this must be why sometimes some xwayland surfaces would seem to have a key stuck! Thanks.

Nefsen402 (approved):

PR #8187, opened 2024-05-28
meson: use versioned wlroots for dependency()

YHNdnzj:
Adapt to upstream change at https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4614


emersion:
There is https://github.com/swaywm/sway/pull/8070 already FWIW

YHNdnzj:
> There is #8070 already FWIW

Ah, thanks. That one also seems neater. Closing.

Issue #8086, opened 2024-03-23
input-method popups are duplicated on screen, and then crash


sam-mccall:
[Here is](https://github.com/sam-mccall/bugreports/blob/main/sway-im-popup/src/main.rs) the world's simplest `input-method-unstable-v2`: it shows and hides the input popup panel on alternating keystrokes.

Sway duplicates the panel every time it is shown, rendering the same surface in both the current location and all the previous ones. Then when the input method is deactivated, sway crashes.

This only happens if I reuse the same surface for `get_input_popup_surface`. If I instead destroy it and use a new surface, things work as expected. I don't see anything in the protocol to indicate I need to use a new surface, though.

It should look like this:

https://github.com/swaywm/sway/assets/548993/c887ef0e-d4c3-4c1f-9341-a4e2dd4db264

But instead looks like this:

https://github.com/swaywm/sway/assets/548993/a2bfb8d5-7912-4a54-a26e-223a657a6cf9

cc @Decodetalkers, as #7226 landed recently

- **Sway Version:**
`sway version 1.10-dev-5a7477cb (Mar 23 2024, branch 'master')`

- **Debug Log:**
https://gist.github.com/sam-mccall/d4f6435a769983190e789c0e590631d8

- **Configuration File:**
Reproduces with empty config.

- **Stack Trace:**

<details>
<summary>Stack trace</summary>

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007f398da381cf in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007f398d9ea472 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f398d9d44b2 in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007f398d9d43d5 in __assert_fail_base
    (fmt=0x7f398db48dc8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x7f398e39653b "node", file=file@entry=0x7f398e396360 "types/scene/wlr_scene.c", line=line@entry=1059, function=function@entry=0x7f398e3967b0 <__PRETTY_FUNCTION__.4> "wlr_scene_node_coords") at ./assert/assert.c:92
[...]

PR #8196, opened 2024-06-05
input_method: Split up update method

Nefsen402:
The update method was a big hammer, it did a lot of things.
- This function would recreate the scene every time the input method client commits a new buffer

Instead, we create the scene when the input method surface maps / unmaps.
When the input method is unfocused, we simply disable the scene node.

Fixes: #8086

### DIFF
--- a/include/sway/input/text_input_popup.h
+++ b/include/sway/input/text_input_popup.h
@@ -15,6 +15,8 @@ struct sway_input_popup {
 
 	struct wl_listener popup_destroy;
 	struct wl_listener popup_surface_commit;
+	struct wl_listener popup_surface_map;
+	struct wl_listener popup_surface_unmap;
 
 	struct wl_listener focused_surface_unmap;
 };
--- a/include/sway/input/text_input_popup.h
+++ b/include/sway/input/text_input_popup.h
@@ -9,6 +9,7 @@ struct sway_input_popup {
 	struct wlr_scene_tree *scene_tree;
 	struct sway_popup_desc desc;
 	struct wlr_input_popup_surface_v2 *popup_surface;
+	struct wlr_output *fixed_output;
 
 	struct wl_list link;
 
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -612,9 +612,11 @@ void arrange_popups(struct wlr_scene_tree *popups) {
 		struct sway_popup_desc *popup = scene_descriptor_try_get(node,
 			SWAY_SCENE_DESC_POPUP);
 
-		int lx, ly;
-		wlr_scene_node_coords(popup->relative, &lx, &ly);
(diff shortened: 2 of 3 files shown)


include/sway/input/text_input_popup.h
```diff
@@ -14,7 +14,8 @@ struct sway_input_popup {
 	struct wl_list link;
 
 	struct wl_listener popup_destroy;
-	struct wl_listener popup_surface_commit;
```
emersion, line 16: Don't we need to still re-constrain the popup when its size changes, for instance?

include/sway/input/text_input_popup.h
```diff
@@ -9,6 +9,7 @@ struct sway_input_popup {
 	struct wlr_scene_tree *scene_tree;
 	struct sway_popup_desc desc;
 	struct wlr_input_popup_surface_v2 *popup_surface;
+	struct wlr_output *fixed_output;
```
emersion, line 12: Took me a bit to figure out that the popup is tied to the layer surface which is tied to the output, so it's OK to store an output pointer without a destroy listener. Maybe we could add a small comment to make this clear.

(Maybe it would be a tad better to fish back the layer surface when needed instead of adding more state?)

sway/input/text_input.c
```diff
@@ -431,18 +436,37 @@ static void handle_im_popup_destroy(struct wl_listener *listener, void *data) {
...
 
-static void handle_im_popup_surface_commit(struct wl_listener *listener,
-		void *data) {
+static void handle_im_popup_surface_map(struct wl_listener *listener, void *data) {
 	struct sway_input_popup *popup =
-		wl_container_of(listener, popup, popup_surface_commit);
-	input_popup_update(popup);
+		wl_container_of(listener, popup, popup_surface_map);
+	popup->scene_tree = wlr_scene_subsurface_tree_create(root->layers.popup,
+		popup->popup_surface->surface);
```
emersion, line 439: Hm, would it be simpler to toggle the scene's node `enabled` based on the `mapped` flag?

(Doesn't the scene surface tree already do this?)

sway/desktop/transaction.c
```diff
@@ -612,9 +612,11 @@ void arrange_popups(struct wlr_scene_tree *popups) {
...
 			SWAY_SCENE_DESC_POPUP);
 
-		int lx, ly;
-		wlr_scene_node_coords(popup->relative, &lx, &ly);
-		wlr_scene_node_set_position(node, lx, ly);
+		if (popup) {
+			int lx, ly;
+			wlr_scene_node_coords(popup->relative, &lx, &ly);
+			wlr_scene_node_set_position(node, lx, ly);
+		}
```
emersion, line 619: Nit: maybe should be in its own commit?

emersion:
Gentle ping :)

emersion (approved):
Thanks!

PR #8200, opened 2024-06-08
Set color transform when calling wlr_scene_output_build_state()

emersion:
We were only passing the color transform when calling wlr_scene_output_commit(). However when modesetting or pushing a new gamma LUT we render via wlr_scene_output_build_state(). Pass the color transform there as well.

cc @mstoeckl


GreyXor:
It fix:: https://github.com/swaywm/sway/pull/7681#issuecomment-2117033174

Nefsen402:
LGTM

kennylevinsen (approved):

PR #8201, opened 2024-06-08
Add mouse_warp to mark option

kwiesmueller:
This allows to better support certain pop-ups which require focus or they close (like certain password managers). The popup can be marked and only that window will get mouse_warped.

For example:
```
for_window [title="Enpass Assistant"] {
    move position mouse
    inhibit_idle visible
    urgent enable
    mark --add mouse_me
}

mouse_warping mark mouse_me
```


emersion:
Note, we're in general not accepting any new window management features anymore. See https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#scope-of-future-changes-to-sway

Issue #7946, opened 2024-02-01
swaylock crashing Sway in wlr_xdg_toplevel_set_size


J0nnyMak0:
With latest sources invoking swaylock crashes Sway. Stack trace attached.

[gdb.txt](https://github.com/swaywm/sway/files/14129531/gdb.txt)

bl4ckb0ne:
Relevant part of the log

```
#6  0x00007b9ae69932d4 in wlr_xdg_toplevel_set_size (toplevel=0x5b7d528d0640, width=-4, height=-29) at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_toplevel.c:558
        __PRETTY_FUNCTION__ = "wlr_xdg_toplevel_set_size"
#7  0x00005b7d51271e5c in configure (view=0x5b7d526da1b0, lx=10882, ly=27, width=-4, height=-29) at ../sway/desktop/xdg_shell.c:164
        xdg_shell_view = 0x5b7d526da1b0
#8  0x00005b7d512b37cc in view_configure (view=0x5b7d526da1b0, lx=10882, ly=27, width=-4, height=-29) at ../sway/tree/view.c:185
#9  0x00005b7d51271250 in transaction_commit (transaction=0x5b7d52f3f080) at ../sway/desktop/transaction.c:803
        instruction = 0x5b7d53098f40
        node = 0x5b7d5253d560
        hidden = false
        i = 21
#10 0x00005b7d51271451 in transaction_commit_pending () at ../sway/desktop/transaction.c:855
        transaction = 0x5b7d52f3f080
#11 0x00005b7d512717ee in _transaction_commit_dirty (server_request=true) at ../sway/desktop/transaction.c:932
#12 0x00005b7d51271804 in transaction_commit_dirty () at ../sway/desktop/transaction.c:936
#13 0x00005b7d5126e8d7 in handle_new_output (listener=0x5b7d512e8bc0 <server+96>, data=0x5b7d52be77f0) at ../sway/desktop/output.c:519
        server = 0x5b7d512e8b60 <server>
        wlr_output = 0x5b7d52be77f0
        scene_output = 0x5b7d52d1c270
        output = 0x5b7d5302b500
        oc = 0x5b7d52781580
 ```
 
 That negative size is odd. @vyivel didnt something about negative sizes got merged in wlroots recently?

Nefsen402:
[...]

PR #8217, opened 2024-06-24
tree/view: ensure view content size is positive

emersion:
Closes: https://github.com/swaywm/sway/issues/7946

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -365,8 +365,8 @@ void view_autoconfigure(struct sway_view *view) {
 
 	con->pending.content_x = x;
 	con->pending.content_y = y;
-	con->pending.content_width = width;
-	con->pending.content_height = height;
+	con->pending.content_width = fmax(width, 1);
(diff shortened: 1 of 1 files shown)


Nefsen402:
I think this is actually a wlroots bug, according to a coredump from a while ago the DRM backend generated an output with a 0x0 size. Unfortunately, I wasn't able to reproduce it since.

emersion:
It's completely possible/valid that an output has a zero size, for instance when it's configured to be powered off before it is plugged in.

emersion:
Besides, it's possible that the output size is too small for the number of tiled toplevels and results in negative sizes. For instance, try starting Sway nested, stacking vertically a lot of terminals with a titlebar, then resizing the Sway output to be smaller. I hit the same assert in that case.

Nefsen402:
Fair, LGTM

Nefsen402:
Thanks!

Nefsen402 (approved):
Thanks number two cause apparently I need to "finish a review" to merge

PR #8222, opened 2024-06-27
protocols: use tablet-v2 xml from stable/

vyivel:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4645


emersion (approved):
Thanks!

PR #8225, opened 2024-06-29
sway/commands/move.c: arrange new workspace

navi-desu:
When moving a container to a new workspace, the workspace's dimension are left unset. Usually this doesn't matter, but when moving a floating container to a new workspace on a different output, this leads to the position of the container being calculated with 0, so the container ends up halfway offscreen on the leftmost topmost monitor.


emersion:
No idea if that's a good way to fix it, let's just go with that for now…

PR #8230, opened 2024-07-01
config/output: Skip VRR tests when not supported

kennylevinsen:
Adaptive sync is a "soft" setting which we degrade of off when not supported. Some outputs types do not support turning it off (Wayland, X11), which makes for an awkward three-way test where we first enable, disable and finally unset the setting.

wlr_output.adaptive_sync_supported tells us whether the output definitely does not support making changes (backend without support, connector without the feature), or whether it might work.

Use this to avoid wasting time on adaptive sync test that can never succeed, and to avoid the Wayland/X11-backend specific unset step.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -721,21 +721,18 @@ static bool search_adaptive_sync(struct search_context *ctx, size_t output_idx) 
 	struct wlr_backend_output_state *backend_state = &ctx->states[output_idx];
 	struct wlr_output_state *state = &backend_state->base;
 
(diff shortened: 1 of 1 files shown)


Nefsen402:
LGTM, seems CI has issues though.

Nefsen402 (approved):

kennylevinsen:
> LGTM, seems CI has issues though.

arch linux builders are currently cooked, presumably by this: https://gitlab.archlinux.org/archlinux/packaging/packages/openssh/-/issues/5

PR #8231, opened 2024-07-03
transaction: fix missing top border with hide_lone_tab

9ary:
Regressed by scene graph.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -314,7 +314,7 @@ static void arrange_children(enum sway_container_layout layout, list_t *children
 
 			if (activated) {
 				arrange_container(child, width, height - title_bar_height,
-					false, 0);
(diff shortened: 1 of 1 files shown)


sway/desktop/transaction.c
```diff
@@ -284,7 +284,7 @@ static void disable_container(struct sway_container *con) {
 }
 
 static void arrange_container(struct sway_container *con,
-		int width, int height, bool title_bar, int gaps);
+		int width, int height, bool title_bar, bool lone_tab, int gaps);
```
Nefsen402, line 287: Why can't we set `title_bar = true` instead of adding the second parameter?

9ary: I was confused by an extra titlebar but now I realize it was unrelated to this change. Back to the original commit.

Nefsen402 (approved):

PR #8232, opened 2024-07-03
xdg_shell: don't send configure events to uninitialized surfaces

llyyr:
the surface isn't initialized yet when we first handle it in `handle_xdg_shell_toplevel`, move setting WM capabilities to handle_commit instead.

Fixes warnings from wlroots about a configure being scheduled for uninitialized surface


emersion (approved):
Thanks!

PR #8234, opened 2024-07-03
tree/container: drop decl for container_update_textures_recursive()

emersion:
This function doesn't exist anymore.


Nefsen402:
Thanks! (ps is there something in the liker that we can enable to disallow this?)

Nefsen402 (approved):

emersion:
> is there something in the liker that we can enable to disallow this?

Something _link_ this would be great, but I don't think the _liker_ can do that, it doesn't read headers.

PR #8235, opened 2024-07-03
sway_text_node: avoid unnecessary redraws

emersion:
When a floating container has a titlebar, render_backing_buffer() ends up being called each time the container is moved. Add some more checks for no-op changes in sway_text_node_set_max_width() and sway_text_node_set_background(). This makes the move smoother.


Nefsen402:
Thanks!

Nefsen402 (approved):

Issue #8205, opened 2024-06-11
Sway doesn't update window visibility when a container changes


WhyNotHugo:
### Sway Version

sway version 1.10-dev-40ca4150b (Jun 10 2024, branch 'master')

I've seen this bug for several weeks now, I think it was introduced with the scene graph changes.

### Debug Log

https://paste.sr.ht/~whynothugo/a7173261a6e1d238ddb739a040b70080921fe84b

### Configuration File

    exec foot
    exec foot
    # You'll need these mapping to easily reproduce the issue.
    bindsym Mod4+a focus parent
    bindsym Mod4+shift+s move container to scratchpad

### Stack Trace

No crash

### Description

- Start on an empty workspace
- Open a terminal
- Open another terminal
- `focus parent`
- `move container to scratchpad`
- **BUG: Both windows are still there**
- Switch to another workspace and back
- Windows are gone (they're now in the scratchpad).

Note that you can still interact with the windows in shrödinger's scratchpad until you switch to another workspace (when they disappear).

WhyNotHugo:
There is another sequence of step with similar results, and I suspect they are the same issue:

- Open FOUR terminals.
- Switch to a tabbed layout.
- Focus the second terminal
- Split vertically
- Focus the fourth terminal
- `move left` twice
- The tab-bar shows the second tab as active, but the third tab is visible (although we never unfocused the terminal that is not part of the split inside the second tab).
- `move left` again (nothing happens on screen, although the window moves to the left inside the split)
- `move left` again. The focused terminal (originally on the far right) has now escaped the split without it ever having been rendered inside the split.

WhyNotHugo:
I bisected the issue, but had to skip a lot of commits that either don't build, or render a black screen.

The first bad commit could be any of:

188811f80861caacd016b857b0d07f6d2d62d15a
5b8b505af5d5925ae9e617ee8f3c7a0f9c43409d
6d7b1321db54155cf78305dbafdcfc7de9b78415
946fc8094559801bc4be629368ac31025d28165a
08c484f46f130aa7e590ef4bcb39d2ceed7160f6
869baff25221e1a1881e9559453faa43f90da33e
b38ed8b4792928dca3e1580e8160792ea41e25c4
bac3ab552683516ee5a09632468cd8de529d311f
6e5fc4c2aafd211323c6037aa868c075852bfe15
9a579666068d62b9354a39941e1ac8c1f4a58093
[...]

PR #8240, opened 2024-07-10
transaction: Reparent all container children when disabling for scratpad

Nefsen402:
Fixes: #8205


WhyNotHugo:
I also left an alternative set of reproduction steps in https://github.com/swaywm/sway/issues/8205#issuecomment-2161090484 which triggers what I believed to be the same issue (but maybe it is a different one with similar symptoms?).

That sequence doesn't involve the scratchpad, but does involve a child switching to another container.

emersion (approved):
Thanks!

Issue #7995, opened 2024-02-21
Title bar text is cut off with scale > 1 and a variable width font


LawnGnome:
- **Sway Version:** sway version 1.10-dev-d6150b6b (Feb 21 2024, branch 'master')

- **Debug Log:** https://gist.github.com/LawnGnome/c4d81e2814cba34013c3972dd0cad117#file-sway-log

- **Configuration File:** https://gist.github.com/LawnGnome/c4d81e2814cba34013c3972dd0cad117#file-config

- **Description:**
  The title bar text is cut off when the following conditions are met:
  1. Output scale == 1.5 (not reproducible at scale == 1)
  2. Variable width font (doesn't reproduce with the default `monospace` font)
  3. A long title (short titles aren't obviously cut off)

  Please see the screenshot below, noting the cut off `g` at the end of the title:

  ![image](https://github.com/swaywm/sway/assets/229984/1b8fe495-78b8-436e-bfc1-081e4626081a)

  This doesn't occur on Sway 1.8.1.

Nefsen402:
Fixed by https://github.com/swaywm/sway/pull/8242

PR #8242, opened 2024-07-10
common/pango: Disable glyph position rounding

kennylevinsen:
Pango rounds glyph position and widths to nearest integer, which leads to font dimensions jumping around when rendering with a scale, causing text geometry to jump around when changing scale. This is disturbing when text buffers change scale, and also mean that the text geometry calculations in sway_text_node are incorrect.

Disable this rounding to make the geometry stable.

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -53,6 +53,8 @@ size_t escape_markup_text(const char *src, char *dest) {
 PangoLayout *get_pango_layout(cairo_t *cairo, const PangoFontDescription *desc,
 		const char *text, double scale, bool markup) {
 	PangoLayout *layout = pango_cairo_create_layout(cairo);
+	pango_context_set_round_glyph_positions(pango_layout_get_context(layout), false);
+
 	PangoAttrList *attrs;
 	if (markup) {
 		char *buf;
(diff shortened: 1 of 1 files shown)


common/pango.c
```diff
@@ -104,6 +106,7 @@ void get_text_size(cairo_t *cairo, const PangoFontDescription *desc, int *width,
 void get_text_metrics(const PangoFontDescription *description, int *height, int *baseline) {
 	cairo_t *cairo = cairo_create(NULL);
 	PangoContext *pango = pango_cairo_create_context(cairo);
+	pango_context_set_round_glyph_positions(pango, false);
```
Nefsen402, line 109: We also need to add this to the rendering path on line ~130

Nefsen402:
Fixes: #7995

kennylevinsen:
Easy test: `for i in $(seq 10); do for y in $(seq 10); do swaymsg output '*' scale "1.$i$y"; done; done`

common/pango.c
Nefsen402: Indeed! I didn't realize there was an entire abstraction between sway_text_node and pango!

Nefsen402:
LGTM

Nefsen402 (approved):

roshal:
fuzzy glyphs in window title and sway-bar

image

`~/.config/sway/config`

```
font pango: DMCA Sans Serif 9
```

`~/.config/fontconfig/fonts.conf`

```xml ~/.config/sway/config
<fontconfig>

	<match target='font'>
		<edit name='antialias'>
			<bool>true</bool>
		</edit>
	</match>

	<match target='font'>
		<edit name='hinting'>
			<bool>true</bool>
		</edit>
	</match>

	<match target='font'>
		<edit name='hintstyle'>
			<const>hintfull</const>
		</edit>
	</match>

	<match target='font'>
		<edit name='autohint'>
			<bool>false</bool>
		</edit>
	</match>

	<match target='font'>
		<edit name='rgba'>
			<const>unknown</const>
		</edit>
	</match>

	<match target='font'>
		<edit name='lcdfilter'>
			<const>lcdnone</const>
		</edit>
	</match>

</fontconfig>
```

PR #8248, opened 2024-07-13
Introduce grouped_titlebar command

kennylevinsen:
Stacking and tabbed layouts effectively override the titlebar control of the border command, always showing the titlebar to allow navigation.

Allow users to also hide the titlebar of stacking/tabbed layouts, through a new command that specify whether grouped titlebar configurations should always be visible (the default) or if they should follow the active container's border configuration.

A take on https://github.com/swaywm/sway/issues/7409 based on one of the suggestions in https://github.com/swaywm/sway/pull/8026.

Plenty of room for bikeshedding the name and shape of the command, but the wiring should be pretty much the same regardless. It's somewhat quirky that the titlebar behavior of a tabbed/stacking container is controlled by the child, but the same goes for the border behavior right now.


sway/sway.5.scd
```diff
@@ -663,6 +668,10 @@ The default colors are:
 	windows that are spawned in floating mode, not windows that become floating
 	afterwards.
 
+*default_grouped_titlebar* always_visible|follows_border
+	Set the default goruped titlebar setting for new windows. Config reload
```
WhyNotHugo, line 672:
```suggestion
	Set the default grouped titlebar setting for new windows. Config reload
```

WhyNotHugo (commented):
I like the reasoning behind this. It's a nice and simple approach. The code makes sense to me too.

Haven't had a chance to actually test it yet.

sway/desktop/transaction.c
```diff
@@ -291,6 +291,12 @@ static void arrange_children(enum sway_container_layout layout, list_t *children
 		int width, int height, int gaps) {
 	int title_bar_height = container_titlebar_height();
 
+	bool child_titlebar = title_bar_height == 0;
+	if (active && active->current.grouped_titlebar_follows_border && active->current.border != B_NORMAL) {
+		child_titlebar = true;
+		title_bar_height = 0;
```
Nefsen402, line 297: We are setting `title_bar_height` to 0 here, why not keep the existing logic to check the variable == 0 instead of adding a bool?

kennylevinsen: Yeah, good point - had some more complicated logic there temporarily and had missed that I had simplified it all the way back to just `title_bar_height == 0`.

emersion:
I wonder if "grouped" is the right terminology here? The i3 user guide only uses the terms "tabbed layout", "stacking layout" and "split layout".

kennylevinsen:
> I wonder if "grouped" is the right terminology here? The i3 user guide only uses the terms "tabbed layout", "stacking layout" and "split layout".

Definitely not! I just tried to come up with a term that could cover both `stacking` and `tabbed` at the same time without a horribly verbose `stacking_and_tabbed_titlebar` scenario. It feels awkward - suggestions welcome.

We could also split the setting up into stacking and tabbed versions to save us the pain of coming up with a name.

kennylevinsen:
Split it into stacking/tabbed commands.

emersion:
Seems like this is missing some files?

kennylevinsen:
> Seems like this is missing some files?

Heh, forgot to re-add after I renamed the command.

kanashimia:
Hello, do you think it would make sense to also allow changing type of the titlebar, as in tabbed/stacked with those options?

The difference between stacked and tabbed layouts isn't only in the titlebar but also in the focus behaviour, with stacked layout focus left always switches to the left neighbour but with tabbed layout it switches to the window in the tab group instead if there is one. I think it makes sense that some people like myself would want stacked layout but with a "tabbed" titlebar, so it takes less vertical space and looks nicer. When you have one master window and a tabbed group nearby it is harder to switch between them with tabbed layout.
That shouldn't be too complicated to add. 
I don't ask to do it in this PR, I may try to do it myself, but maybe should consider naming of the options because they fit the bill.
Thinking of something like this:
 `stacking_titlebar follows_border|always_visible [vertical|horizontal]`

Or maybe it is the focus behaviour that could be configured, or is it not worth adding? What do you guys think?

i3 had stack-limit option very long time ago which kinda did same thing, but that was axed sadly https://i3wm.org/docs/3.e/userguide.html#stack-limit
In Niri you can configure similar tabbed indicator for the stack, can also place it to the left or right.

Jacocococo:
After merging this into my own install to try it out, I found some issues related to window sizes, especially with some weird behavior when it comes to applying it to a tree of stacked/tabbed containers

Firstly, simply creating a new window in an empty workspace and setting the layout to tabbed or stacked without specifying any default_{tabbed/stacking}_titlebar or setting it for the window will cause this window to extend outside of its border by the height of the titlebar.

Secondly, setting the layouts of parent containers also causes weird issues. An example of doing so is this: In a new workspace, create 2 windows split horizontally, then split one of them vertically. After that, change the layout of the vertically split container to tabbed, then select this container's parent and change the layout of this to tabbed (effectively 'maximizing' a window)

The default_{tabbed/stacking}_titlebar setting will not apply to this parent. I think this is the right way to do it since layers of tabbed/stacked layouts are unintuitive to navigate. However, if the setting of this container is manually set as `tabbed_titlebar follows_border`, something strange will happen. Instead of the window getting bigger to fill out the extra space from the titlebar no longer being visible, there will be a gap at the bottom of the window, either within out outside the defined window border

I also somehow (unsure how) managed to make it have a gap both within and outside of the border (white: window, pink: window border, purple: wallpaper. Normal gaps are only set to 4 pixels as seen on the sides)
image

This seems to be a general issue about handling window sizes. Other than these bugs though, I think this is a good implementation that works nicely

PR #8249, opened 2024-07-14
Chase wlroots!2434 "Assorted borderline bikeshedding breaking API cleanups"

Billli11:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2434

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -27,7 +27,7 @@ packages:
 - x11/libX11
 - x11/pixman
 - x11/xcb-util-wm
-- x11-servers/xwayland-devel
+- x11-servers/xwayland
 - misc/hwdata
 sources:
 - https://github.com/swaywm/sway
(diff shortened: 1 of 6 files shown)


emersion:
Hm, the FreeBSD build is failing for some reason… Can you try pushing again? (With a new commit hash, e.g. `git commit --amend` and then leave commit message as-is)

Billli11:
Done but job still failed.

Look like package `x11-servers/xwayland-devel` is do longer in repo.  
FreeBSD Git repositories. [link](https://cgit.freebsd.org/ports/tree/x11-servers/xwayland-devel) return 404 error

emersion:
Ah, maybe we can just use the regular xwayland package now (without -devel)? Could you try that in a separate commit?

emersion (approved):
Thanks!

emersion:
If you're bored, would be helpful to submit that FreeBSD change back to wlroots :P

Billli11:
Done. [!4747](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4747)

PR #8251, opened 2024-07-14
build: use fs.relative_to() instead of hand-rolled logic

emersion:
Meson has introduced a relative_to() function [1] in its fs module since version 1.3.

[1]: https://mesonbuild.com/Fs-module.html#relative_to

Same as https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4724

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,7 @@ project(
 	'c',
 	version: '1.10-dev',
 	license: 'MIT',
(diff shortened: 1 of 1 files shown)


Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8259, opened 2024-07-19
sway/server: Implement wayland socket handover

Ferdi265:
This PR implements the compositor side of the Wayland socket handover protocol as described in the [KDE Wiki][KDE Wiki]. The CLI options are chosen so that they are compatible with Kwin (and Hyprland):

Sway now accepts the CLI options `--socket NAME` and `--wayland-fd FD`, which pass the name of the Wayland display socket and the Wayland display socket fd number respectively. If given, Sway will use these instead of finding a free Wayland display socket name and creating it.

This allows for Sway to be used with a helper tool that automatically restarts the compositor on crashes, and allowing clients to reconnect after the compositor restart and survive the crash. The reconnecting clients look like new clients to Sway and perform all the same steps as a fresh Wayland client would. No additional handling is required on the side of Sway.

The helper tool finds and creates the Wayland socket and launches Sway with the new CLI options, restarting it when it exits with a non-zero exit code. Two such helper tools currently exist as far as I know:

- [`kwin_wayland_wrapper`](https://invent.kde.org/plasma/kwin/-/blob/master/src/helpers/wayland_wrapper/kwin_wrapper.cpp?ref_type=heads) (only for KDE)
- [`wl-restart`](https://github.com/ferdi265/wl-restart) (developed by me)

[KDE Wiki]: https://invent.kde.org/plasma/kwin/-/wikis/Restarting

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -164,7 +164,7 @@ extern bool unsupported_gpu_detected;
 
 void sway_terminate(int exit_code);
 
-bool server_init(struct sway_server *server);
+bool server_init(struct sway_server *server, const char * socket_name, int socket_fd);
 void server_fini(struct sway_server *server);
 bool server_start(struct sway_server *server);
 void server_run(struct sway_server *server);
--- a/sway/main.c
+++ b/sway/main.c
@@ -208,6 +208,8 @@ static const struct option long_options[] = {
 	{"verbose", no_argument, NULL, 'V'},
 	{"get-socketpath", no_argument, NULL, 'p'},
 	{"unsupported-gpu", no_argument, NULL, 'u'},
+	{"socket", required_argument, NULL, 's'},
+	{"wayland-fd", required_argument, NULL, 'S'},
 	{0, 0, 0, 0}
 };
 
@@ -221,12 +223,16 @@ static const char usage[] =
 	"  -v, --version          Show the version number and quit.\n"
 	"  -V, --verbose          Enables more verbose logging.\n"
 	"      --get-socketpath   Gets the IPC socket path and prints it, then exits.\n"
+	"      --socket <name>    Sets the Wayland socket name (for Wayland socket handover)\n"
+	"      --wayland-fd <fd>  Sets the Wayland socket fd (for Wayland socket handover)\n"
 	"\n";
 
 int main(int argc, char **argv) {
 	bool verbose = false, debug = false, validate = false, allow_unsupported_gpu = false;
 
 	char *config_path = NULL;
+	char *socket_name = NULL;
+	int socket_fd = -1;
 
 	int c;
 	while (1) {
@@ -271,6 +277,13 @@ int main(int argc, char **argv) {
 				fprintf(stderr, "sway socket not detected.\n");
 				exit(EXIT_FAILURE);
 			}
+			break;
+		case 's': // --socket
+			free(socket_name);
+			socket_name = strdup(optarg);
+			break;
+		case 'S': // --wayland-fd
+			socket_fd = atoi(optarg);
 			break;
 		default:
 			fprintf(stderr, "%s", usage);
@@ -337,7 +350,7 @@ int main(int argc, char **argv) {
 
 	sway_log(SWAY_INFO, "Starting sway version " SWAY_VERSION);
 
-	if (!server_init(&server)) {
+	if (!server_init(&server, socket_name, socket_fd)) {
 		return 1;
 	}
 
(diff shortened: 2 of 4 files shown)


Ferdi265:
Update: Labwc is unsure whether the Kwin CLI options are the right path forward. Before merging this, we should consider alternative mechanisms before multiple compositors standardize on a subpar mechanism. See [this wl-restart issue](https://github.com/Ferdi265/wl-restart/issues/2).

emersion:
I think this would be a nice feature. I've dropped my thoughts in https://github.com/Ferdi265/wl-restart/issues/2#issuecomment-2419679363, and pinged KDE folks about it.

yurikhan:
> The reconnecting clients look like new clients to Sway and perform all the same steps as a fresh Wayland client would.

Does that mean they create new windows which are assigned new Sway container IDs that may or may not be the same as in the previous life?

If a custom script tracks windows, what attribute of the container must it use that is unique across the session, and stable over restarts? On i3 we had `window`, and I have observed `id` to change over i3 restart. On sway, `window` is always `null` and I see no other stable identifier attribute candidates.

kennylevinsen:
> If a custom script tracks windows, what attribute of the container must it use that is unique across the session, and stable over restarts? On i3 we had window, and I have observed id to change over i3 restart. On sway, window is always null and I see no other stable identifier attribute candidates.

There are no such thing as globally unique and persistent identifiers for a window. The only persistent properties will be whatever the client sets themselves, e.g., app_id and title. Some clients might use xdg-toplevel-tag though which could help in some cases.

If you build something external that snapshots the tree and tries to restore it after crash/restart, you'll have to use heuristics on the available, non-unique information. You probably won't have too many windows from the same app with the exact same title.

Ferdi265:
> > The reconnecting clients look like new clients to Sway and perform all the same steps as a fresh Wayland client would.
> 
> Does that mean they create new windows which are assigned new Sway container IDs that may or may not be the same as in the previous life?

Yes. Sway has no idea that a client came from a previous sway instance and therefore has to assume it's new.

> If a custom script tracks windows, what attribute of the container must it use that is unique across the session, and stable over restarts? On i3 we had `window`, and I have observed `id` to change over i3 restart. On sway, `window` is always `null` and I see no other stable identifier attribute candidates.

AFAIK this is still unsolved in Wayland generally. `xdg-session-management-v1`, which has an open PR in wayland-protocols would provide a session id + toplevel name to identify application toplevels accross restarts, though I'm not fully up-to-date if `xdg-session-management-v1` is the currently favored approach or how much consensus it has right now across desktops.

I think the best you can currently do is try to match window titles and app ids before and after restart, but that still leaves you open to confusing windows with identical titles.

yurikhan:
Thank you both.

Titles are neither stable (a Firefox window changes its title every time I switch tabs, and for some sites, even when I don’t) nor unique (two windows whose current tabs are looking at the same web page also yield a collision, even though their sets of tabs are otherwise completely distinct).

Okay, for now, I will tell myself to not worry about it.

Silent-Hunter:
I would like to say, this feature would solve an annoying issue I have. I use Sway when using my computer at my desk, and Labwc when using my computer at my TV. I have a cron job that changes the background once a minute, but it needs to be passed the WAYLAND_DISPLAY variable. Being able to choose it would mean being able to set the variable statically in the cron entry.

Ferdi265:
> I would like to say, this feature would solve an annoying issue I have. I use Sway when using my computer at my desk, and Labwc when using my computer at my TV. I have a cron job that changes the background once a minute, but it needs to be passed the WAYLAND_DISPLAY variable. Being able to choose it would mean being able to set the variable statically in the cron entry.

That can also be done in a different way: if you use systemd and use a systemd timer (or a systemd-based cron implementation), then you can have sway / labwc execute a script on startup that runs `systemctl --user import-environment WAYLAND_DISPLAY DISPLAY XDG_SESSION_TYPE XDG_CURRENT_DESKTOP SWAYSOCK` or similar, and all your systemd timers or services started from that point will have the correct environment.

Also, as implemented here and in `wl-restart`, this will still choose `WAYLAND_DISPLAY` dynamically. I have experimented with creating the Wayland socket statically or via a systemd `.socket` unit, but that still needs something else that handles the lock file next to the socket, which regular systemd socket units don't have. That is definitely solvable though, so what you want should in theory be implementable (that's more a feature for something like `wl-restart` though. If sway gains socket handover support, it doesn't care whether the socket comes from systemd, is chosen statically, or from `wl-restart`)

Silent-Hunter:
Ah no, I'm using Gentoo with OpenRC.

Issue #7936, opened 2024-01-26
Layer Shell regression: Layershell never releasing mouse again, once it is set to exclusive


feschber:
Setting the layershell keyboard interactivity to `exclusive` and then `none` makes it impossible to focus windows.
This must be a recent regression possibly related to the `scene_graph` api.

Might be  related: #7918

- **Sway Version:**
  - `swaymsg -t get_version` or `sway -v`
  
I tried to bisect but could not get an exact commit
it must be one of the following
```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
188811f80861caacd016b857b0d07f6d2d62d15a
5b8b505af5d5925ae9e617ee8f3c7a0f9c43409d
6d7b1321db54155cf78305dbafdcfc7de9b78415
946fc8094559801bc4be629368ac31025d28165a
08c484f46f130aa7e590ef4bcb39d2ceed7160f6
869baff25221e1a1881e9559453faa43f90da33e
b38ed8b4792928dca3e1580e8160792ea41e25c4
bac3ab552683516ee5a09632468cd8de529d311f
6e5fc4c2aafd211323c6037aa868c075852bfe15
9a579666068d62b9354a39941e1ac8c1f4a58093
0639bde9fb5637fbe63c65bfe85660745f159b96
ed2724bd6c83ad3fdc2010b3a1e2f5967f0e8b38
06ad734e70227ad0527fe11b88ad37e93005ce0c
c640c3015f3a7ea2987bd7854d13ff282f90804f
1e018e72b4d57c8f354b9be9686a7a75797cdcab
9c17cba0b29979ae23c4521b884f7419fd558770
0e1a02bf0aad3d743985602b55043c5de019d1f0
5f0801b6f245cc789ea93b9449dd0c573ef36e66
2e53de80bb0f4c93e74ae050fa07e78f18e909d9
We cannot bisect more!
```

- **Description:**
  - Run [Lan Mouse](https://github.com/feschber/lan-mouse) and configure an arbitrary client. Then move the mouse in one of the layershell surfaces on the configured screen edge.
  - There is now no way of focusing any other window.
  
  The relevant code can be found here: https://github.com/feschber/lan-mouse/blob/5cc8cda19d2bef25cff0d767361d0622a8fbf3e8/src/backend/producer/wayland.rs#L407-L487

feschber:
So I did some more testing, and my suspicion seems to be confirmed by https://github.com/wmww/gtk-layer-shell

[...]

PR #8032, opened 2024-03-02
seat: Drop has_exclusive_layer

Nefsen402:
Fixes: https://github.com/swaywm/sway/issues/7936

In the transition to wlr_scene, I accidentally dropped a call to reset this `has_exclusive_layer` so once it was set, nothing could reset it afterwards and the compositor basically unusable. Let's just drop the has_exclusive_layer bool for something simpler and more robust.

PR #8261, opened 2024-07-21
layer-shell: Restore interactive layer focus code

joanbm:
Commit 188811f80861 ("scene_graph: Port layer_shell") accidentally removed code in `arrange_layers` to handle focus on layer shell surfaces with keyboard interactivity.

Due to this, layer shell surfaces requesting exclusive keyboard interactivity may not get automatically focused, and layer shell surfaces giving up exclusive keyboard interactivity can remain focused.

Add the [previous code](https://github.com/swaywm/sway/blob/2c69e19fd30c5ceac61d6af821c6c6875cd3f8b6/sway/desktop/layer_shell.c#L247-L279) back to fix the problem.

Note the non-rename change included in b4d7e84d3852 ("desktop: Rename layers to shell_layers") is not included as it also seems accidental.

Fixes: #7936

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -90,6 +90,43 @@ void arrange_layers(struct sway_output *output) {
 	} else {
 		arrange_popups(root->layers.popup);
 	}
+
+	// Find topmost keyboard interactive layer, if such a layer exists
+	struct wlr_scene_tree *layers_above_shell[] = {
+		output->layers.shell_overlay,
+		output->layers.shell_top,
+	};
+	size_t nlayers = sizeof(layers_above_shell) / sizeof(layers_above_shell[0]);
+	struct wlr_scene_node *node;
+	struct sway_layer_surface *topmost = NULL;
+	for (size_t i = 0; i < nlayers; ++i) {
+		wl_list_for_each_reverse(node,
+				&layers_above_shell[i]->children, link) {
+			struct sway_layer_surface *surface = scene_descriptor_try_get(node,
+				SWAY_SCENE_DESC_LAYER_SHELL);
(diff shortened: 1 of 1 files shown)


joanbm:
This has the same purpose as #8032 but instead recovers the removed code as-is.

What I mention as "layer shell surfaces requesting exclusive keyboard interactivity may not get automatically focused" can be triggered by running `seq 3 | bemenu --grab`.

WhyNotHugo (commented):
I tested 35dd8d002acad34e63f28a6a2da2bf568d631c1b with shotman and it works for me: when shotman sends the request `set_keyboard_interactivity(none)`, focus is properly relinquished.

Nefsen402:
I don't like this because this is just papering over horrible and confusing input code. Ideally, during keypresses we decide what we are focusing in some scene native way, gone are `seat_set_focus_*()`.

But that requires a lot of work, and more importantly, an actual understanding of how focus in sway is supposed to work. Because every time I take a look at the code to try to gain some understanding, I blow a gasket. So as an interim solution, we can keep this.

Nefsen402 (approved):
Thanks!

joanbm:
> I don't like this because this is just papering over horrible and confusing input code. Ideally, during keypresses we decide what we are focusing in some scene native way, gone are `seat_set_focus_*()`.
> 
> But that requires a lot of work, and more importantly, an actual understanding of how focus in sway is supposed to work. Because every time I take a look at the code to try to gain some understanding, I blow a gasket. So as an interim solution, we can keep this.

Thanks for the review! Indeed, this comes from a place of "let's get this working again" [so I can bear testing the v1.10 branch on my PC] rather than from a place of "let's try to get to an elegant solution".

Issue #6436, opened 2021-08-13
Keymap not applied correctly for virtual keyboard


Xyene:
Sway config:

```
input * {
    xkb_layout "myus"
}
```

Then, connecting with wayvnc and opening up a shell, the keyboard layout is still `us`.

```
$ swaymsg -t get_inputs                                                                                                                                                                     
Input device: virtual pointer
  Type: Mouse
  Identifier: 0:0:virtual_pointer
  Product ID: 0
  Vendor ID: 0

Input device: virtual keyboard
  Type: Keyboard
  Identifier: 0:0:virtual_keyboard
  Product ID: 0
  Vendor ID: 0
  Active Keyboard Layout: English (US)
```

Switching it to `myus` is ineffective.

```
$ swaymsg input '*' xkb_layout myus   
$ swaymsg -t get_inputs                                                                                                                                                                     
Input device: virtual pointer
  Type: Mouse
  Identifier: 0:0:virtual_pointer
  Product ID: 0
  Vendor ID: 0

Input device: virtual keyboard
  Type: Keyboard
  Identifier: 0:0:virtual_keyboard
  Product ID: 0
  Vendor ID: 0
  Active Keyboard Layout: English (US)
```

To get things working, I first have to set the layout to `us` (this should be a noop, but appears to not be one), and _then_ to `myus`.

```
$ swaymsg input '*' xkb_layout us                                                                                                                                                                
$ swaymsg input '*' xkb_layout myus                                                                                                                                                           
$ swaymsg -t get_inputs 
Input device: virtual pointer
  Type: Mouse
  Identifier: 0:0:virtual_pointer
  Product ID: 0
  Vendor ID: 0

Input device: virtual keyboard
  Type: Keyboard
  Identifier: 0:0:virtual_keyboard
  Product ID: 0
  Vendor ID: 0
  Active Keyboard Layout: English (Mine)
```

At this point everything works.

Xyene:
/cc @any1 since I figure you're most familiar with this. Do you reckon the bug is:

- that the layout isn't applied at connection time, or
- that the layout of the virtual keyboard _can_ be changed at all?

The protocol doesn't seem to suggest what should happen in this case. Is the keymap specified in the virtual keyboard protocol supposed to always take precedence?

any1:
[...]

PR #8264, opened 2024-07-25
Add null-safety check for virtual keyboard keymaps

JingMatrix:
Even though there are already many issues / prs trying to address this problem, including #6484, #6873, #6990, I still get segfaults due to null keymap pointer.

Hence, I would simply suggest to add a check before we call `xkb_keymap_num_layouts`.

### DIFF
--- a/sway/commands/input/xkb_switch_layout.c
+++ b/sway/commands/input/xkb_switch_layout.c
@@ -95,10 +95,18 @@ struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
 			continue;
 		}
 
+		struct wlr_keyboard *keyboard =
+			wlr_keyboard_from_input_device(dev->wlr_device);
+		if (keyboard->keymap == NULL && dev->is_virtual) {
+			// The `sway_keyboard_set_layout` function is by default skipped
+			// when configuring virtual keyboards.
+			continue;
+		}
+
 		struct xkb_switch_layout_action *action =
 			&actions[actions_len++];
+		action->keyboard = keyboard;
 
-		action->keyboard = wlr_keyboard_from_input_device(dev->wlr_device);
 		if (relative) {
 			action->layout = get_layout_relative(action->keyboard, relative);
 		} else {
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -1133,7 +1133,9 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 		json_object *layouts_arr = json_object_new_array();
 		json_object_object_add(object, "xkb_layout_names", layouts_arr);
 
-		xkb_layout_index_t num_layouts = xkb_keymap_num_layouts(keymap);
+		xkb_layout_index_t num_layouts =
+			keymap ? xkb_keymap_num_layouts(keymap) : 0;
+		// Virtual keyboards might have null keymap
 		xkb_layout_index_t layout_idx;
 		for (layout_idx = 0; layout_idx < num_layouts; layout_idx++) {
 			const char *layout = xkb_keymap_layout_get_name(keymap, layout_idx);


kennylevinsen:
> ... I still get segfaults due to null keymap pointer.

Do you have a reproduction?

sway/ipc-json.c
```diff
@@ -1129,6 +1129,7 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 		json_object *layouts_arr = json_object_new_array();
 		json_object_object_add(object, "xkb_layout_names", layouts_arr);
 
+		if (keymap == NULL) return object;
```
kennylevinsen, line 1132: nit: coding style, we always use bracketed if blocks, not bracket-less or single-line ifs.

Note that we also cannot just return, as we still need to get past the libinput block to add libinput details to the object. If we consider a NULL keymap a valid state, then we should just make the keymap access conditional.

JingMatrix:
> Do you have a reproduction?

I cannot really reproduce it **on your device** since we may have different hardwares.
But logically, doing a null-safety check is a necessary and favorable coding style.

On my devices, without this commit, it always leads to a segfault.
Thanks for your review, changed as requested.

kennylevinsen:
> I cannot really reproduce it on your device since we may have different hardwares.

That's not what I asked - I need to know the order of events that as reliably as possible lead to this issue in order to reason about the issue and its fix, even if I can't use the reproduction directly.

(Whether it works on others' hardware directly or with adjustments also won't be known until the reproduction is shared.)

> But logically, doing a null-safety check is a necessary and favorable coding style.

The NULL check is good if the value is *intended* to sometimes be NULL, but then we need to ensure that every other place using the keymap also does an appropriate NULL check.

Doing a runtime NULL check on value that must always be non-null is bad practice. A safety check in this case would be an assert to _guarantee_ a clean crash with a coredump for debugging when impossible values occur. Skipping over impossible values is symptom treatment - the root cause (the bug that caused the NULL value) is then still present, and now less likely to be found as the symptom treatment hides it.

Knowing more about the issue will help here.

JingMatrix:
The following JSON is the result of IPC_GET_INPUTS (with many unrelated devices removed) after the above commit applied, and I guess that the keywap is null since it is a virtual_keyboard (see the first element in the list).
```json
[
  {
    "identifier": "0:0:wlr_virtual_keyboard_v1",
    "name": "wlr_virtual_keyboard_v1",
    "type": "keyboard",
    "repeat_delay": 600,
    "repeat_rate": 25,
    "xkb_layout_names": []
  },
  {
    "identifier": "0:0:HP_WMI_hotkeys",
    "name": "HP WMI hotkeys",
    "type": "keyboard",
    "libinput": {
      "send_events": "enabled"
    },
    "vendor": 0,
    "product": 0,
    "repeat_delay": 600,
    "repeat_rate": 25,
    "xkb_layout_names": [
      "French (Number row switched)",
      "French"
    ],
    "xkb_active_layout_index": 1,
    "xkb_active_layout_name": "French"
  },
  {
    "identifier": "9011:26214:ydotoold_virtual_device",
    "name": "ydotoold virtual device",
    "type": "pointer",
    "libinput": {
      "send_events": "enabled",
      "accel_speed": 0.0,
      "accel_profile": "adaptive",
      "natural_scroll": "disabled",
      "middle_emulation": "disabled"
    },
    "vendor": 9011,
    "product": 26214,
    "scroll_factor": 1.0
  },
  {
    "identifier": "9011:26214:ydotoold_virtual_device",
    "name": "ydotoold virtual device",
    "type": "keyboard",
    "libinput": {
      "send_events": "enabled",
      "accel_speed": 0.0,
      "accel_profile": "adaptive",
      "natural_scroll": "disabled",
      "middle_emulation": "disabled"
    },
    "vendor": 9011,
    "product": 26214,
    "repeat_delay": 600,
    "repeat_rate": 25,
    "xkb_layout_names": [
      "French (Number row switched)",
      "French"
    ],
    "xkb_active_layout_index": 1,
    "xkb_active_layout_name": "French"
  },
  {
    "identifier": "1:1:AT_Translated_Set_2_keyboard",
    "name": "AT Translated Set 2 keyboard",
    "type": "keyboard",
    "libinput": {
      "send_events": "enabled"
    },
    "vendor": 1,
    "product": 1,
    "repeat_delay": 600,
    "repeat_rate": 25,
    "xkb_layout_names": [
      "French (Number row switched)",
      "French"
    ],
    "xkb_active_layout_index": 1,
    "xkb_active_layout_name": "French"
  }
]
```
Being a wlr_virtual_keyboard, there could be no default keymap.

JingMatrix:
Note that in the `sway_keyboard_configure` function of sway/input/keyboard.c,
https://github.com/swaywm/sway/blob/4d4c88f0a73f6ee3da1c99355f04362ef2ad68c9/sway/input/keyboard.c#L1077-L1079
we have skipped the `sway_keyboard_set_layout` function for virtual
keyboards, which then have null keymaps.
Hence, a null-safety check is needed at runtime.

layercak3:
I run into this segfault right after sway boots as I automatically start both waybar with the sway/language module enabled (which sends IPC_GET_INPUTS) and fcitx5 (which creates a virtual keyboard) at launch. Reverting e32bdaa7bead5052dd32c12917ea8f74a9b14405 prevents the segfault from happening.

Somehow, it doesn't happen on my (slower) laptop. This lead me to add `sleep 5` before starting waybar in my waybar service. If I don't touch anything before waybar starts, then sway crashes after 5 seconds. However if I open a terminal, then waybar starts successfully. Other code paths seem to be causing the virtual keyboard's keymap to be compiled, that don't have the is_virtual check. Not that I notice any performance issues because of it, or have performance issues with e32bdaa7bead5052dd32c12917ea8f74a9b14405 reverted. I assume due to the reception of this PR that keymaps for virtual keyboards should have been compiled once so this null access should never be happening, meaning that e32bdaa7bead5052dd32c12917ea8f74a9b14405 expects the keymap to have been compiled already, which it isn't in some cases?

layercak3:
I found some wlr_keyboard_set_keymap calls related to "keyboard groups" but in a sway debug log I don't see the virtual keyboard added to any group, so I ignored that.

When fcitx5 starts, it immediately sends create_virtual_keyboard.
In wlroots wlr_keyboard_set_keymap is run when the virtual keyboard client sends zwp_virtual_keyboard_v1::keymap. Fcitx5 sends that request [here](https://github.com/fcitx/fcitx5/blob/6af78b6fbbb56751b788a710aab3455005fcf808/src/frontend/waylandim/waylandimserverv2.cpp#L424) in keymapCallback, however, this callback doesn't get run for the first time until the first done event. Looking at an fcitx5 debug log, when I opened that terminal, it receives a done event and runs keymapCallback for the first time.

If the ipc server receives IPC_GET_INPUTS before that first done-event/keymapCallback, sway crashes because it never called wlr_keyboard_set_keymap for the virtual keyboard yet.

layercak3:
After reading https://github.com/swaywm/sway/issues/6436 (compositors should not switch the layout of a virtual keyboard), I think that the approach in this PR and keeping e32bdaa7bead5052dd32c12917ea8f74a9b14405 is good. As stated above, there is a period of time where virtual keyboard keymaps can be NULL as the client may not immediately send over the keymap after creating the object.

Even if it's a rare case, I think the docs under GET_INPUTS in sway-ipc(7) should have a note stating that the xkb properties will not be included if the client did not yet send the keymap.

sway/commands/input/xkb_switch_layout.c
```diff
@@ -100,6 +100,10 @@ struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
 
 		action->keyboard = wlr_keyboard_from_input_device(dev->wlr_device);
 		if (relative) {
+			if (action->keyboard->keymap == NULL && dev->is_virtual) {
+				actions_len -= 1;
+				continue;
+			}
```
layercak3, line 106: This should be checked when given an index too, not just next/prev. The switch_layout code also calls xkb_keymap_num_layouts, not just the code to get the current layout's index.
```suggestion
		if (action->keyboard->keymap == NULL && dev->is_virtual) {
			actions_len -= 1;
			continue;
		}
		if (relative) {
```

JingMatrix: You are right, I changed it as requested.

sway/commands/input/xkb_switch_layout.c
```diff
@@ -99,6 +99,11 @@ struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
 			&actions[actions_len++];
 
 		action->keyboard = wlr_keyboard_from_input_device(dev->wlr_device);
+		if (action->keyboard->keymap == NULL && dev->is_virtual) {
+			actions_len -= 1;
```
emersion, line 103: Instead of incrementing right above and decrementing after-the-fact, can we store the result of `wlr_keyboard_from_input_device()` in an intermediate variable and `continue` if it's NULL?

sway/commands/input/xkb_switch_layout.c
```diff
@@ -99,6 +99,11 @@ struct cmd_results *input_cmd_xkb_switch_layout(int argc, char **argv) {
 			&actions[actions_len++];
 
 		action->keyboard = wlr_keyboard_from_input_device(dev->wlr_device);
+		if (action->keyboard->keymap == NULL && dev->is_virtual) {
```
emersion, line 102: Can we maybe leave it the `dev->is_virtual` bit, and add a comment saying that no keymap might happen with virtual devices?

sway/ipc-json.c
```diff
@@ -1133,6 +1146,9 @@ json_object *ipc_json_describe_input(struct sway_input_device *device) {
 		json_object *layouts_arr = json_object_new_array();
 		json_object_object_add(object, "xkb_layout_names", layouts_arr);
 
+		if (keymap == NULL && device->is_virtual) {
+			return object;
+		}
```
emersion, line 1151: Suggested change:

```c
xkb_layout_index_t num_layouts = keymap ? xkb_keymap_num_layouts(keymap) : 0;
```

emersion:
Thanks for the investigation.

JingMatrix:
Changed as requested @emersion , thank you for the advices.

emersion (approved):
Thanks!

Issue #7974, opened 2024-02-12
Multiple Xwayland windows are stacked in the wrong order


tkna91:
**Sway Version:**
- Sway version: `1.10-dev-88b2abf5 (Feb 11 2024, branch 'master')`
- wlroots version: `0.18.0-dev`
- OS: `Arch Linux x86_64`
- Kernel: `6.7.4-arch1-1`

**Screencast, Debug Log:**
- Screencast: 

https://github.com/swaywm/sway/assets/102382754/6aed8b7a-2c98-4b6a-99f8-6d352a942392

- sway debug log: https://paste.rs/IBnpw
- Console log of Chromium run: https://paste.rs/JgwkO

**Configuration File:**
- `/etc/sway/config`

**Description:**
Reproduction procedure
1. `sudo pacman -S chromium reaper` to Install Chromium and [REAPER](https://www.reaper.fm/)
2. `sway -c /etc/sway/config -d 2> sway-run-default_`date +%Y%m%d-%H%M%S`.log` to start Sway
3. `meta+enter` to start foot
4. `chromium<enter>` to start Chromium
5. `meta+w` to make tab layout
6. `meta+enter` to start foot
7. `reaper<enter>` to start REAPER
8. `escape` to close the floating window
9. click anywhere in the REAPER window

-> **I can't click on the REAPER window, the Chromium tab turns red. Layers are in the wrong order.**

Comment
- ~~I have heard that REAPER uses some sort of proprietary GUI library for games, so this may be related to that.~~
   - The GUI for REAPER was open source. It appears to use this. https://github.com/justinfrankel/WDL
       - https://www.askjf.com/index.php?q=5526s
       - https://www.askjf.com/index.php?q=6777s
- ~~The same problem occurs with [Renoise](https://www.renoise.com/), which is also an application for music and uses its own GUI library often used for games.~~
  * I checked again and it did not occur. It seems that since Renoise is also an Xwayland window like Chromium, this phenomenon occurs when used with REAPER at the same time.
- This is a multi-display environment.

Related Links
- #7608 
- https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4052

emersion:
Is this a regression? Does this work fine on previous Sway versions?

tkna91:
@emersion I think it was the same before, but let's check with the official release version, shall I ?

tkna91:
This does not seem to occur in the official release version.

Version:
```
$ pacman -Q sway wlroots0.16
[...]

PR #8268, opened 2024-07-29
desktop/xwayland: don't restack when marking window as inactive

emersion:
daaec72ac01f ("desktop/xwayland: restack surface upon activation") has updated Sway for wlroots commit bfc69decdd04 ("xwm: do not restack surfaces on activation"). However, it unconditionally restacks the window above all other windows even if marking the window as inactive.

Closes: https://github.com/swaywm/sway/issues/7974


Nefsen402:
Brilliant, thanks!

Nefsen402:
I wonder if we can make scene handle xwayland stacking in the future.

emersion:
Ah, that would be quite nice indeed!

PR #8273, opened 2024-08-03
build: avoid git repository discovery when determining version

jdknight:
When attempting to use Git to populate commit/branch information in a version string, it is possible through repository discovery that it uses Git information not relevant to project. For example, if repository content is extract into an interim build location when using an embedded build framework (e.g. Buildroot), the project will not have its Git repository to refer to. When it cannot find its repository, it will look into its parent folders and may find the Git repository of another project and use its branch/commit information.

This commit provides an explicit path to the project's Git repository when consider commit/branch information. This will prevent any repository discovery from occurring.


emersion (approved):
Thanks!

PR #8280, opened 2024-08-07
tearing: fix UAF, remove trailing whitespace

vyivel:
Fixes: 9a1c411abd8261c121dcd50dfe54132718768084


kennylevinsen (approved):
Hmm, we should have the pipeline fail on trailing whitespace...

Would be nicer if the bugfix was a commit after the trailing whitespace mess, but LGTM.

vyivel:
Split into two commits.

PR #8282, opened 2024-08-07
commands/output/color_profile: allows use of relative path for ICC profile

FelixPehla:
Allows using relative paths to specify the ICC file, and not just the absolute one.

I mostly took the code of `background.c` and adjusted it to fit into `color_profile.c`. Let me know if there are any issues.

### DIFF
--- a/sway/commands/output/color_profile.c
+++ b/sway/commands/output/color_profile.c
@@ -5,6 +5,7 @@
 #include <wlr/render/color.h>
 #include "sway/commands.h"
 #include "sway/config.h"
+#include "stringop.h"
 
 static bool read_file_into_buf(const char *path, void **buf, size_t *size) {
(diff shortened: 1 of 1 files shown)


emersion:
Hm, if possible, I'd like to leave the `join_paths` stuff out: I think we mostly have it in the `background` command for historical reasons, and I find it a bit confusing that arguments are automagically joined together into a single string like this. (Probably will need to `strdup()` the argument.)

FelixPehla:
That worked fine too. I wasn't sure why it was done like that in `background.c` but since I'm not that familiar with this codebase I thought reusing the code (that worked fine) would be the safe thing to do. Maybe the argument parsing has been different back then?

Anyway, does it look good now?

emersion:
Yeah, I think it's there for histerical raisins and now we can't remove it without breaking user configs.

emersion (approved):
LGTM, thanks!

PR #8287, opened 2024-08-08
input/keyboard: add support for pointer keys

emersion:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4775

Note, key repeat for motion is not yet implemented.

Simple way to test: `wtype -k Pointer_Right`

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -4,6 +4,7 @@
 #include <wlr/config.h>
 #include <wlr/backend/multi.h>
 #include <wlr/interfaces/wlr_keyboard.h>
+#include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_keyboard.h>
 #include <wlr/types/wlr_keyboard_group.h>
 #include <xkbcommon/xkbcommon-names.h>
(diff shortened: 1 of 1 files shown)


sway/input/keyboard.c
```diff
@@ -282,6 +284,35 @@ static bool keyboard_execute_compositor_binding(struct sway_keyboard *keyboard,
...
+			wlr_seat_pointer_notify_frame(cursor->seat->wlr_seat);
+			return true;
+		}
+
+		int dx, dy;
+		wlr_keyboard_keysym_to_pointer_motion(keysym, &dx, &dy);
+		if (state == WL_KEYBOARD_KEY_STATE_PRESSED && (dx != 0 || dy != 0)) {
+			wlr_cursor_move(cursor->cursor, NULL, dx * 10, dy * 10);
+			cursor_rebase(cursor);
+			wlr_seat_pointer_notify_frame(cursor->seat->wlr_seat);
```
kennylevinsen, line 308: Maybe we should just call `pointer_motion(cursor, time, &keyboard->wlr.base, dx, dy, dx, dy)`?

emersion: Indeed, that sounds better!

kennylevinsen (approved):
LGTM!

PR #8289, opened 2024-08-11
Switch default config to wmenu-run

emersion:
This removes the last dependency bit on dmenu. No need for "swaymsg exec" anymore: wmenu-run handles the xdg-activation shenanigans.


kennylevinsen (approved):
LGTM!

Issue #4114, opened 2019-05-01
swaybar disable changing workspaces with mouse wheel


uglywizkid:
Exists there any way to disable changing workspaces with mouse wheel? In i3 this can be established with

```
bar {
  wheel_up_cmd nop
  wheel_down_cmd nop
}
```

hmohamad:
According to `man sway-bar`:

`bindsym [--release] button[1-9]|<event-name> <command>`

Executes command when the mouse button has been pressed (or if released is given, when the button has been released). The buttons can be given as a x11 button number or an event name, which can be obtained from `libinput debug-events`. To disable the default behavior for a button, use the command nop.


So in this case its button 4 and 5 for vertical scrolling and 5 & 6 for horizontal scrolling.
```
bar {
    # diable vertical scrolling
    bindsym button4 nop
    bindsym button5 nop
    # diable horizontal scrolling
    bindsym button6 nop
    bindsym button7 nop
}
```

uglywizkid:
Thanks. This works charmfully.

larchenkov:
Is it possible to do the same for window tabs?

larchenkov:
Has the same question it's very annoying when window tabs switching with scroll

jayache80:
@larchenkov Yes. To disable mousewheel-over-titlebars-changes-tabs behavior, simply put the same `bindsym`s in the regular config section, outside of the `bar` block. So, in my config, I have:
```
bar {
    ...
    # Don't let mousewheel scrolling change workspaces
    # Diable vertical scrolling
    bindsym button4 nop
    bindsym button5 nop
    # Diable horizontal scrolling
    bindsym button6 nop
    bindsym button7 nop
}

# Don't let mousewheel scrolling change tabs
# Diable vertical scrolling
bindsym button4 nop
bindsym button5 nop
# Diable horizontal scrolling
bindsym button6 nop
bindsym button7 nop
```
You need it in both sections like that if you want to disable that behavior for both the status bar and title bars.

(Sorry for the noise, but this post comes up on google when searching for "swaywm mousewheel titlebars")

nermolov:
What would the config be to *enable* scrolling on title bars? I'm using default binding config and don't have it.

stacyharper:
I'm trying to disable the touch drag that change workspace as the mouse wheel does. The `libinput debug-events` show me the TOUCH_DOWN but it does not work from my sway config. Is it possible ?

stacyharper:
[...]

PR #8299, opened 2024-08-15
swaybar: prioritize workspace scroll over bindsyms

joanbm:
This adds handling for scroll events (e.g. mouse wheel) to the hotspot callback equivalent to the fallback in `process_discrete_scroll`.
While this may seem redundant (and in many cases, it is), it changes the priority when the user has Button4/Button5 bindsyms on the bar.

Before this commit:
* Scrolling on unused bar space calls the user's bindsyms.
* Scrolling over the workspaces *still* calls the user's bindsyms.

After this commit:
* Scrolling on unused bar space calls the user's bindsyms.
* Scrolling over the workspaces moves to the previous/next workspace.

If the user has no bindsyms for Button4/Button5, there is no change.

# Utility

This shows an example of why this is useful. Here, the user has set up a Button4/Button5 bindsym to change the volume (see: top right corner) when scrolling over the bar.

Before this change, the user could no longer use the mouse wheel to switch workspaces:

beforeY

After this change, the user can (*only* when scrolling over the workspaces):

afterY

# Discussion

This is consistent with the idea that workspaces are hotspots, and hotspot event handlers take priority over the user's bindsyms, see https://github.com/swaywm/sway/commit/53f9dbd424dc173a85c9f4cd30802259d38b1ef4 ("swaybar: Prioritize hotspot events to bar bindings").

However, note that **this strays further away from i3's behaviour**. On i3, user bindsyms take precedence over workspaces, even for Button1. In fact, it's [explicitly documented as such](https://web.archive.org/web/20240725173949/https://i3wm.org/docs/userguide.html#_mouse_button_commands).

### DIFF
--- a/include/swaybar/input.h
+++ b/include/swaybar/input.h
@@ -78,6 +78,9 @@ uint32_t event_to_x11_button(uint32_t event);
 
 void free_hotspots(struct wl_list *list);
 
+bool handle_workspace_button(struct swaybar_output *output,
+	uint32_t button, bool released, const char *ws);
+
 void swaybar_seat_free(struct swaybar_seat *seat);
 
 #endif
--- a/swaybar/input.c
+++ b/swaybar/input.c
@@ -246,6 +246,15 @@ bool handle_workspace_button(struct swaybar_output *output,
 		}
 		ipc_send_workspace_command(output->bar, ws);
 		return true;
+	} else if (button == SWAY_SCROLL_UP || button == SWAY_SCROLL_DOWN ||
+			button == SWAY_SCROLL_LEFT || button == SWAY_SCROLL_RIGHT) {
+		if (released) {
+			return true;
+		}
+
+		workspace_next(output->bar, output,
+				button == SWAY_SCROLL_UP || button == SWAY_SCROLL_LEFT);
+		return true;
 	}
 
 	return false;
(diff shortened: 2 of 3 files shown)


joanbm:
**Note 1:** Some people *want* to disable scrolling to switch workspaces: https://github.com/swaywm/sway/pull/1041, https://github.com/swaywm/sway/issues/4114. It's unclear to me if they don't like it being triggered when scrolling over free bar space (if so, this change doesn't make them unhappy) or with the entire concept in general (if so, this change will make them unhappy).

**Note 2:** I realized that with https://github.com/swaywm/sway/commit/53f9dbd424dc173a85c9f4cd30802259d38b1ef4, I broke compatibility with i3: After it, a `bar bindsym button1` does not prevent switching workspaces (this was working before, see https://github.com/swaywm/sway/pull/2798).

The underlying conceptual problem is that i3bar has three levels: Tray icons, bindsyms, then workspaces, in descending priority. In swaybar, we only have two: Hotspots (which include both tray icons and workspaces) and bindsyms. So, to make things consistent, we should either prioritize hotspots over bindsyms everywhere (this is what this PR does, at the cost of further breaking i3 compatibility), or make workspaces "low-priority hotspots" (i.e. be faithful to i3's behaviour).

Issue #8296, opened 2024-08-15
Regression: Adaptive Sync causes unsupported monitors to be disabled on startup


it-a-me:
**Sway Version:** sway version 1.10-dev-c30c4519 (Aug 14 2024, branch 'master')

- **Debug Log:**
[sway-err.log](https://github.com/user-attachments/files/16622442/sway-err.log)

- **Configuration File:**
modifications: add `output * adaptive_sync on` to the top of the config
[err-config.txt](https://github.com/user-attachments/files/16622460/err-config.txt)

- **Description:**
I started sway with the latest git version.  On startup, all monitors are disabled and I am left with 2 black screens(dual monitors)

It appears that when adaptive_sync is manually enabled on outputs with unsupported connectors, sway fails to initialize the monitor and disables the output. Reverting this commit fixes the issue: https://github.com/swaywm/sway/commit/4e38f93f367dfb7f1ec66060e6262b806cecf3a7  
Relevant portions of log

```
00:00:00.434 [DEBUG] [wlr] [backend/drm/drm.c:804] connector DP-3: Can't enable adaptive sync: connector doesn't support VRR
00:00:00.434 [DEBUG] [wlr] [types/wlr_output_swapchain_manager.c:181] Test commit for 2 outputs failed
00:00:00.434 [DEBUG] [wlr] [types/wlr_output_swapchain_manager.c:160] Preparing test commit for 2 outputs with implicit modifiers
00:00:00.434 [DEBUG] [wlr] [backend/drm/drm.c:804] connector DP-3: Can't enable adaptive sync: connector doesn't support VRR
00:00:00.434 [DEBUG] [wlr] [types/wlr_output_swapchain_manager.c:181] Test commit for 2 outputs failed
00:00:00.434 [DEBUG] [sway/config/output.c:848] Unable to find valid config with output DP-3, disabling
00:00:00.434 [DEBUG] [sway/config/output.c:675] Output state for DP-3
00:00:00.434 [DEBUG] [sway/config/output.c:677]     enabled:       no
```

etrnal70:
Wow thanks ! Finally found out why mine doesn't show up after recent commits, had `adaptive_sync` on kanshi config

emersion:
cc @kennylevinsen: maybe a consequence of your recent work?

kennylevinsen:
Can you check if https://github.com/swaywm/sway/pull/8301 fixes the issue?

it-a-me:
Yep, it works great now. Thanks

PR #8301, opened 2024-08-18
config/output: Do not set adaptive_sync if not supported

kennylevinsen:
After 4e38f93f367d ("config/output: Skip VRR tests when not supported"), the configuration search no longer touches VRR state for outputs that are known to not support it. This also means that it will not remove VRR if already set, which could cause output configuration to fail.

Ensure that VRR state is never set for outputs that do not support it by adding the same test for support to queue_output_config.

Fixes: 4e38f93f367d ("config/output: Skip VRR tests when not supported")
Fixes: https://github.com/swaywm/sway/issues/8296

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -491,7 +491,7 @@ static void queue_output_config(struct output_config *oc,
 		wlr_output_state_set_scale(pending, scale);
 	}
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

PR #8303, opened 2024-08-19
config/output: Skip search if config has a mode

kennylevinsen:
When doing an output configuration search, the intent is to only look for modes if the output's configuration does not contain a specific mode. This was done by testing if config_has_auto_mode returned false.

config_has_auto_mode had its return values backwards, leading to other modes being tested if the output configuration had specified modes or modelines, leading to unwanted modes being selected.

Invert the function to config_has_manual_mode to give it a clearer name, and fix the return values in the process.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -651,9 +651,9 @@ struct output_config *find_output_config(struct sway_output *sway_output) {
 	return result;
 }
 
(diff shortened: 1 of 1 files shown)


Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8311, opened 2024-08-22
sway/tree/container: don't trunc coords in `floating_fix_coordinates`

llyyr:
This can cause issues such as the window not being shown at the exact same coordinates when the old and new wlr_box aren't the same dimensions and the container is being moved back-and-forth between them.

For example, in the case where a floating window gets moved from one output to another but the outputs aren't the same resolution. For e.g. have two displays that aren't the same resolution then:

1. Open a floating window and set it to pos 0,0 on output 2
2. Send it to scratchpad then `scratchpad show` on output 1
3. `scratchpad show` on output 2 again

Observe that the window isn't at 0,0 on output 2 anymore.


emersion:
Thanks!

PR #8318, opened 2024-08-27
Use heading markdown to demarcate sections of commands

jonmon6691:
It's a little tough to notice that the COMMANDS section is actually 3 sections. Here is a simple change to the man page to give the sections their own headings.


emersion:
Can we maybe keep the COMMANDS section, and add some sub-sections?

jonmon6691:
Sure thing. Modified the file to used these sub-sections:

```
$ grep ^# sway.5.scd 
# NAME
# DESCRIPTION
# COMMAND CONVENTIONS
# COMMANDS
## Config only commands
## Runtime only commands
## Config or runtime commands
# CRITERIA
# SEE ALSO
```

emersion (approved):
Thanks!

PR #8320, opened 2024-08-30
Support for RGB565 render format for GUD

kennylevinsen:
The [GUD](https://github.com/notro/gud/wiki) driver needs to use RGB565 for performance reasons, but exposes XRGB8888 with the driver doing a conversion in that case, making a fallback insufficient.

Extend the bit depth option to allow lowering to a maximum of 6 bits per channel.

Depends on https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4824 to have any use.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -262,6 +262,7 @@ enum scale_filter_mode {
 
 enum render_bit_depth {
 	RENDER_BIT_DEPTH_DEFAULT, // the default is currently 8
+	RENDER_BIT_DEPTH_6,
 	RENDER_BIT_DEPTH_8,
 	RENDER_BIT_DEPTH_10,
 };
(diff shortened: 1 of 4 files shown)


mstoeckl:
Note: wlroots!4824 is not needed to test this, as RGB565 is often supported with the DRM and wayland backends for wlroots.

Setting bit depth to 6 worked fine with this PR; the only problem with the code that I noticed is that the error message in `sway/commands/output/render_bit_depth.c`, "Invalid bit depth. Must be a value in (8|10).", should be updated to say something like "(6|8|10)".

kennylevinsen:
> "Invalid bit depth. Must be a value in (8|10)."

Good point, fixed.

kennylevinsen:
Thanks for the reviews!

Issue #8298, opened 2024-08-15
Keybinding Deosn't Work When Change The Keyboard Layout


ezzobir:
When I change my keyboard layout from english to arabic the my sway keybinding deosn't working
I was using i3 before and it works normally

emersion:
Try using `--to-code` maybe?

ezzobir:
where I use it mester @emersion

Issue #8275, opened 2024-08-04
Bindings to keysyms are layout-dependent


chebykinn:
Hi, I've been using Sway on and off for a while already, but I wanted to understand the reason for making bindings layout-dependent.

Let's say for example a user sets a key binding to open an application menu the same way they did it in i3: 
```
bindsym $mod+d exec rofi -show run
```
This will only work with English keyboard layout. If they have a setup with multiple layouts like with an additional Cyrillic layout the menu wouldn't be called unless they switch back to English. To avoid this it is suggested to use `bindsym --to-code` in the  man page, but it's not straightforward for the people who come from i3 and I've found multiple related issues here before posting this.

So roughly I have these questions:
- What is the use case for binding keys to layouts? (AFAIK, other OSes, and environments don't do this)
- Can the default be changed?

chebykinn:
Or maybe this is an incorrect behavior on the i3 side, since they have `bindsym` and `bindcode`, the latter is supposed to work independently from the layout, but then I just don't understand the reason for having `bindsym`.

emersion:
A keysym (key symbol) doesn't identify a physical key. It identifies the result of translation a key according to a keyboard layout. A keycode identifies a physical key.

One could write a full config with keycodes rather than keysyms, but that's not what `bindsym` does.

chebykinn:
Right, but from my everyday experience on i3 `bindsym` somehow maps to physical keys, why does it work differently? 

I understand now the use case for binding to letters, French layout is a good example, do I understand that you have it as the only layout?

emersion:
I don't remember what i3 does but probably it does `--to-code` by default or something. We chose to not do that in Sway because it went against the principle of keysyms.

Some people have French as the only layout, some people have both French and English (personally I only have English).

chebykinn:
Got it, thanks! But what is the difference between `bindsym --to-code` vs `bindcode`?

emersion:
`bindsym --to-code` takes keysyms as input and converts them into keycodes according to the default layout. `bindcode` takes keycodes as input.

`bindsym` accepts symbol names e.g. "a", but for `bindcode` you'd need to look up the `KEY_*` constant in `/usr/include/linux/input-event-codes.h` and put the number in the config file.

chebykinn:
I understand now, thanks. I still think that for i3 compatibility it would've been best to keep the same behavior, but I get why you didn't do that. I'll leave this thread open for a while in case someone else would be interested.

[...]

PR #8323, opened 2024-09-01
input/keyboard: fix bindsym --to-code failing on duplicates

### LINKED ISSUE #8267: Bindsym --to-code not working with non-modifier keys.
- **Sway Version:**
1.9

- **Configuration File:**
`bindsym --to-code print+a exec firefox` won't work with other layouts.

- **Description:**
Currently Sway won't work if you use key combos with non-modifier keys (like shift, meta etc). Would be nice to support that as I use Print key as an extra modifier for Sway.

sahinf:
If a keysymbol maps to more than one keycode, we should choose the first keycode instead of refusing the binding. It is more likely that duplicates are coming from nonstandard keys such as the 'Internet' keys. 

After these changes, I can `bindsym --to-code Print`  as expected.

`libxkbcommon` API maps both keycodes 107 and 218 to the Keysym 0xff61 = 65377.
The only way to map between keysym to keycode or vice versa seems to be to loop
through all keycodes exposed by [`xkb_keymap_key_for_each`](https://xkbcommon.org/doc/current/group__components.html#gaa6d6a6d4aaf70ca0aa798e491d5616d8) and it starts looping from the lowest keycode. 

[Some more investigation here](https://github.com/swaywm/sway/issues/8267#issuecomment-2322812982)

To test, bind a nonstandard key to some visual changes eg: 
```config
# bindsym --to-code Print exec notify-send --expire-time=1000 "$(brightnessctl -m set 5%+)"
# bindsym --to-code Print+a exec notify-send --expire-time=1000 "$(brightnessctl -m set 5%-)"
```

Fixes #8267, #8298, maybe #8275

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -52,11 +52,12 @@ Throughout the documentation, *|* is used to distinguish between arguments for
 which you may only select one. *[...]* is used for optional arguments, and
 *<...>* for arguments where you are expected to supply some value.
 
-# COMMANDS - CONFIG ONLY
+# COMMANDS
 
 This section only lists general commands. For input and output commands, refer
 to *sway-input*(5) and *sway-output*(5).
 
+## Config only commands
 The following commands may only be used in the configuration file.
 
 *bar* [<bar-id>] <bar-subcommands...>
@@ -98,8 +99,7 @@ The following commands may only be used in the configuration file.
 	machines, it may be desirable to have Xwayland started immediately by
 	using _force_ instead of _enable_.
 
-# COMMANDS - RUNTIME ONLY
-
+## Runtime only commands
 The following commands cannot be used directly in the configuration file.
 They are expected to be used with *bindsym* or at runtime through *swaymsg*(1).
 
@@ -387,8 +387,7 @@ set|plus|minus|toggle <amount>
 
 	The default format is "%title".
 
-# COMMANDS - CONFIG OR RUNTIME
-
+## Config or runtime commands
 The following commands may be used either in the configuration file or at
 runtime.
 


emersion:
Well, the issue is that there is no way to tell whether the first one is the "good" one, or whether it's a different one. Maybe you want the first one but someone else will get confused because it expects the second one.

sahinf:
> no way to tell whether the first one is the "good" one

As far as my limited perspective regarding Internet keys, then yes there is a way to tell: Most people don't have internet keys. This fixes current Sway user issues without really taking away any feature or introducing new branch logic. It used to reject both keycodes, but now it would choose the first.  

Unless this is a `libxkbcommon` bug, I would argue that my compositor should absolutely make assumptions like choosing a real print key rather than being unable to map print key altogether, just because there was a duplicate entry or whatever with a key that literally doesn't exist.

sahinf:
@emersion Any other ideas to get `bindsym --to-code` to work as expected?

kennylevinsen:
> As far as my limited perspective regarding Internet keys, then yes there is a way to tell: Most people don't have internet keys.

Sway does not know what "internet keys" are, and the key people want is not guaranteed to be the first in the keymap, so this doesn't really seem like a solution - just a hack that works in this case.

Maybe we should just change the semantic to bind to *all* the matched keycodes. That might be more in line with what a user expects. If they want a specific one, they should be explicit with bindcode.

sahinf:
I assumed that we were rejecting duplicates for a reason but I can't find any reasoning in commit a09c144b8b5f9d0518e7239a27e2fb86e00644b3

I agree that would be more intuitive.

PR #8324, opened 2024-09-02
Move title_format to container

### LINKED ISSUE #8322: Support title_format for all containers
In i3 you can add title_format to all containers not only for windows. This allows to setting friendly names for tabs containing layouts in tabbed and stacked layout, making navigation easier/more intuitive.

i3 issue - resolved: i3/i3#2120
i3 PR resolving that issue - i3/i3#2143

bwskin:
Fixes #8322

This PR moves logic behind title_format support to container. This allows to add title_format to non-view containers, like i3 does.

Usage example: Formatting tab names in tabbed and stacked layouts which contain child layouts

Before:
before

After:
after

Related i3 issue: i3/i3#2120
Related i3 PR: i3/i3#2143

### DIFF
--- a/sway/commands/title_format.c
+++ b/sway/commands/title_format.c
@@ -1,3 +1,4 @@
+#define _POSIX_C_SOURCE 200809L
 #include <string.h>
 #include "sway/commands.h"
 #include "sway/config.h"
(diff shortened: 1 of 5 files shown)


sway/commands/title_format.c
```diff
@@ -11,16 +12,20 @@ struct cmd_results *cmd_title_format(int argc, char **argv) {
...
-		free(view->title_format);
+	if (container->title_format) {
+		free(container->title_format);
+	}
+	container->title_format = format;
+	if (container->view) {
+		view_update_title(container->view, true);
+	} else {
+		container_update_representation(container);
+		container_update_title_bar(container);
```
Nefsen402, line 28: I don't think this call is needed. I think `container_update_representation` will update the title bar itself.

Nefsen402 (changes requested):
Rest looks good

sway/commands/title_format.c
bwskin: Yeah, it does, thank you.

Nefsen402 (approved):
Thanks!

PR #8330, opened 2024-09-05
config/output: Fix interaction with output management

kennylevinsen:
Configuration applied by output management had an unfortunate interaction with configuration already present. Only a configuration based on the new config_head would be applied, rather than the result of merging this configuration with what was stored.

If the output manager provided a config_head that lead to `enable = 1`, this would disregard that the output had `power = 0` and light up an output that should have remained off. When committing any other changes later, the power state is sorted out, disabling the output again.

Rather than try to sort out merging at the call-site, we take a different approach: Just store the config temporarily while testing in a way that we can easily remove afterwards.

To reproduce the issue on master:
1. `WLR_WL_OUTPUTS=2 sway`
2. `swaymsg WL-2 power off`
3. `wlr-randr --output WL-2 --on` - Output turns on
4. `swaymsg WL-2 scale 2` - Output turns back off as stored config takes effect

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -544,9 +544,8 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 }
 
 static struct output_config *output_config_for_config_head(
-		struct wlr_output_configuration_head_v1 *config_head,
-		struct sway_output *output) {
-	struct output_config *oc = new_output_config(output->wlr_output->name);
+		struct wlr_output_configuration_head_v1 *config_head) {
+	struct output_config *oc = new_output_config(config_head->state.output->name);
 	oc->enabled = config_head->state.enabled;
 	if (!oc->enabled) {
 		return oc;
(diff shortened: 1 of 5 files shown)


sway/desktop/output.c
```diff
@@ -573,56 +572,31 @@ static struct output_config *output_config_for_config_head(
...
+		// config. We keep a record of it so we can remove it later.
+		struct output_config *oc = output_config_for_config_head(config_head);
+		store_temp_output_config(oc);
+		configs[config_idx++] = oc;
 	}
+	configs_len = config_idx;
 
-	sort_output_configs_by_priority(configs, configs_len);
-	bool ok = apply_output_configs(configs, configs_len, test_only, false);
+	bool ok = apply_all_output_configs(test_only, false);
```
Nefsen402, line 592: Maybe we can change this function to accept a pointer to an array of output configs (just pass configs directly) and a length, instead of storing it in output state. Not liking these `store_temp_output_config()`/`remove_temp_output_config()` tbh.

kennylevinsen: That's not a bad idea. It'll take a few signature changes, but should be fine.

kennylevinsen: Take a look at: https://github.com/swaywm/sway/pull/8341

Nefsen402:
The alternative PR has been merged. Thanks for all the work!

Issue #8328, opened 2024-09-04
crash after staying unused for a short time


svalaskevicius:
- **Sway Version:**
  - `swaymsg -t get_version` or `sway -v`
 sway version 1.10-dev-8eddb83f (Aug 31 2024, branch 'hiding-lone-titlebar-scenegraph')

  built using this master commit:
  ```
  * 292e7013 (HEAD -> hiding-lone-titlebar-scenegraph)  fix hiding lone titlebar (Sarunas Valaskevicius) 2024-03-08
  * 41c5ec96  fix hiding lone titlebar (Sarunas Valaskevicius) 2024-02-28
  * f38b6ad7  add --smart-titles flag to hide_edge_borders (Guilhem Saurel) 2023-12-29
  >> * 980a4e02 (upstream/master, upstream/HEAD)  use subheadings instead (Jon Wallace) 2024-08-27
  ```

stderr, stdout, and crash stacktrace: https://gist.github.com/svalaskevicius/1bb506cbbca62c20e8284f4924ac1399

- **Configuration File:**
https://github.com/svalaskevicius/dotfiles/blob/gen-2024/.config/sway/config

- **Stack Trace:**
see https://gist.github.com/svalaskevicius/1bb506cbbca62c20e8284f4924ac1399

- **Description:**
I left the PC be, could be related to any timeouts, swaylock etc

Nefsen402:
@kennylevinsen I believe you were running into a similar assert (negative size passed to wlr_scene_rect_create)

svalaskevicius:
hmm I suppose it _could_ be related to #8076 ...

svalaskevicius:
> hmm I suppose it _could_ be related to #8076 ...

I take that back `arrange_output (output=0x59efd45ac140, width=0, height=0) at ../sway/desktop/transaction.c:598` - the whole output is 0x0 ??

kennylevinsen:
So the issue is:
1. An output is alive and active with workspaces and windows on it.
2. The output is disabled.
3. The output disconnects from standby.
4. The output reconnects on its own, but is still disabled and so has no mode.
6. The output is added to the layout, but with a zero-sized box
7. Workspaces are attempted arranged on this layout, but everything is trying to calculate with the zero dimensions and things explode.

The only way I see to handle this is to ensure that:
1. When an output is not enabled, ensure that all workspace/scene nodes are disabled so they cannot spill over to neighboring outputs. I don't remember if we already do this - it's late.
[...]

PR #8332, opened 2024-09-05
transaction: Disable workspaces of disabled outputs

kennylevinsen:
Instead of attempting to arrange workspaces on a disabled output which may not have a valid geometry, treat workspaces on disabled output as disabled, ensuring that scene nodes are properly disabled and the missing geometry not accidentally relied on.

Fixes: https://github.com/swaywm/sway/issues/8328#issuecomment-2332747218

### DIFF
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -314,14 +314,6 @@ void arrange_output(struct sway_output *output) {
 	if (config->reloading) {
 		return;
 	}
-	struct wlr_box output_box;
-	wlr_output_layout_get_box(root->output_layout,
-		output->wlr_output, &output_box);
-	output->lx = output_box.x;
-	output->ly = output_box.y;
-	output->width = output_box.width;
-	output->height = output_box.height;
-
 	for (int i = 0; i < output->workspaces->length; ++i) {
 		struct sway_workspace *workspace = output->workspaces->items[i];
 		arrange_workspace(workspace);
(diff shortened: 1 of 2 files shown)


sway/tree/arrange.c
```diff
@@ -314,14 +314,9 @@ void arrange_output(struct sway_output *output) {
...
 	}
-	struct wlr_box output_box;
-	wlr_output_layout_get_box(root->output_layout,
-		output->wlr_output, &output_box);
-	output->lx = output_box.x;
-	output->ly = output_box.y;
-	output->width = output_box.width;
-	output->height = output_box.height;
-
+	if (!output->wlr_output->enabled) {
```
Nefsen402, line 317: We want to add this check to `transaction.c` as well. Look at the function `arrange_root()` in there. But it looks like the outright removal of `wlr_output_layout_get_box()` is fine here because we do it anyway when we finalize the output config.

kennylevinsen: So far my trick was to make `activated` in arrange_output dependent on `output->wlr_output->enabled`, so that the workspace trees get disabled. I think that works? It's a bit hard to see what happens on the screen before the output modesets...

We need all output specific nodes to be disabled while the output is so that not updating their position and size is okay, otherwise the unarranged content might overflow to neighbor outputs while our scene output geometry is zero-sized. :/

Nefsen402: I see, I missed that before.\

kennylevinsen:
The dirty node setup check out, as the `arrange_root` after the modeset will dirty them.

Nefsen402 (approved):
Thanks!

Issue #8333, opened 2024-09-06
Sway freezes while opening window with title change ruleset


Vityacv:
- **Sway Version:**
commit https://github.com/swaywm/sway/commit/be840f730e747a24106c8366ecb89e6b982cfa38 with latest wlroots

- **Configuration File:**
Just add ```include sway-font-awesome/icons-wl``` to config
OR
```for_window [app_id=^Alacritty$|^foot$|^footclient$|^kitty$|-terminal$] title_format "­<span size='100%' font-weight='bold'></span> %title"```

- **Description:**
If you are using some rules to change title (for example this set of rules https://github.com/iguanajuice/sway-font-awesome/blob/main/icons-wl) last commit be840f730e747a24106c8366ecb89e6b982cfa38 causes sway to freeze when opening window that matches one of the rules

PR #8335, opened 2024-09-06
container: Skip % char if it doesn't match a view property

Nefsen402:
The else condition was missed here and we would never skip the % char if it didn't end up matching with any property. Since we fail to skip we would re-evaluate the % in an infinite loop never achieving any forward-progress.

Fixes: https://github.com/swaywm/sway/issues/8333


emersion (approved):
Thanks!

PR #8338, opened 2024-09-08
sway/input/keyboard: always set active keyboard if there is none

llyyr:
Previously, we incorrectly only set active keyboard for non-virtual devices. 4c3c0602116c12c2821e1e505e7248b3c642b4ca incorrectly put unrelated code in `sway_keyboard_set_layout`.

Fixes: 4c3c0602116c12c2821e1e505e7248b3c642b4ca


any1:
Works for me.

emersion (approved):
Thanks!

PR #8339, opened 2024-09-08
input/keyboard: fix bindsym --to-code failing on duplicates

sahinf:
The `libxkbcommon` api does not prohibit multiple keycodes mapping to a single keysym. There is no reason to treat it as an error. With these changes, all of the matched keycodes will be bound the keysym.

Error variable is to track whether malloc failed while iterating. Can potentially keep calling a failing malloc but it will get logged and aborted once `xkb_keymap_key_for_each` returns control flow to sway. The only way to avoid malloc within the iterator is by preallocating enough memory to hold possible every keycode in the keymap, then copying over those matches->keycodes into the binding->keycodes. But that feels even more problematic.

Retains cleanup semantics of blame-commit a09c144

### DIFF
--- a/sway/commands/bind.c
+++ b/sway/commands/bind.c
@@ -22,9 +22,7 @@ void free_sway_binding(struct sway_binding *binding) {
 	if (!binding) {
 		return;
 	}
-
 	list_free_items_and_destroy(binding->keys);
-	list_free_items_and_destroy(binding->syms);
 	free(binding->input);
 	free(binding->command);
 	free(binding);
@@ -282,6 +280,7 @@ static struct cmd_results *binding_add(struct sway_binding *binding,
 		const char *keycombo, bool warn) {
 	struct sway_binding *config_binding = binding_upsert(binding, mode_bindings);
 
+	binding->order = binding_order++;
 	if (config_binding) {
 		sway_log(SWAY_INFO, "Overwriting binding '%s' for device '%s' "
 				"to `%s` from `%s`", keycombo, binding->input,
@@ -467,10 +466,11 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 	// sort ascending
 	list_qsort(binding->keys, key_qsort_cmp);
 
+	binding->command = join_args(argv + 1, argc - 1);
+
 	// translate keysyms into keycodes
 	if (!translate_binding(binding)) {
-		sway_log(SWAY_INFO,
-				"Unable to translate bindsym into bindcode: %s", argv[0]);
+		sway_log(SWAY_INFO,"Unable to translate bindsym into bindcode: %s", argv[0]);
 	}
 
 	list_t *mode_bindings;
@@ -481,13 +481,10 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 	} else {
 		mode_bindings = config->current_mode->mouse_bindings;
 	}
-
 	if (unbind) {
 		return binding_remove(binding, mode_bindings, bindtype, argv[0]);
 	}
 
-	binding->command = join_args(argv + 1, argc - 1);
-	binding->order = binding_order++;
 	return binding_add(binding, mode_bindings, bindtype, argv[0], warn);
 }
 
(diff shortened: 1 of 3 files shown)


sahinf:
Hey I forgot to use a separate branch and accidentally closed the original PR #8323 by force pushing my undo. 
These changes bind all the matching keycodes as discussed.

emersion:
Thanks for updating your patch!

Does this actually work? I would expect this to require all keycodes matching a keysym to be pressed. So if a single keysym can be produced by either keycode 42 and 43, I would expect this patch to require both to be pressed.

sahinf:
I thought it was working for me but I tested again and no it doesn't, oops. 
Looks like this would require the [`get_active_binding()`](https://github.com/swaywm/sway/blob/fb5eadc363a7f8b9eeeb0ba562ecb3c40e0e6e5a/sway/input/keyboard.c#L179) to additionally handle this specific case.

I left an issue on the libxkbcommon gh asking for clarification regarding duplicate keysymbols. Still not sure whether that's a bug or not. In the case that it is intentional, and multiple keycodes CAN be mapped to a unique keysym, then we should remove bindsym's --to-code flag because it's not possible to guarantee that mapping.

emersion:
It's not a libxkbcommon bug, it's an expected feature to be able to map multiple physical keys to a single keysym.

> Looks like this would require the [get_active_binding()](https://github.com/swaywm/sway/blob/fb5eadc363a7f8b9eeeb0ba562ecb3c40e0e6e5a/sway/input/keyboard.c#L179) to additionally handle this specific case.

Another way to implement this would be to duplicate the binding as many times as we find keycode combinations.

The annoying thing here is handling the combinatorial explosion, e.g. if keycodes (42, 43) both trigger keysym A and keycodes (45, 46) both trigger keysym B, then for an A+B binding we'd need to create 4 keycode bindings.

sahinf:
Thanks, yeah I got [confirmation](https://github.com/xkbcommon/libxkbcommon/issues/504#issuecomment-2341407172) that the duplicate mapping is correct.

> The annoying thing here is handling the combinatorial explosion, e.g. if keycodes (42, 43) both trigger keysym A and keycodes (45, 46) both trigger keysym B, then for an A+B binding we'd need to create 4 keycode bindings.

This seems clever, thanks. Can you specify what you mean by "trigger keysym"? I'm under the impression that keysyms aren't a factor when handling events, just keycodes. So the keycodes would trigger the command.

emersion:
> Can you specify what you mean by "trigger keysym"?

By "trigger", I mean "keycode is mapped to keysym". IOW, pressing on a physical key will emit one or more keysyms.

Reworded: if keycodes (42, 43) both are mapped to keysym A and keycodes (45, 46) both are mapped to keysym B, then for an A+B binding we'd need to create 4 keycode bindings.

sahinf:
@emersion Marking as ready for review. It introduces new functions, removes an existing function, and modifies some functions, but I have reasoning for everything. Please let me know if I left out any reasoning in the commit message or within comments.

sway/commands/bind.c
```diff
@@ -649,101 +646,144 @@ void seat_execute_command(struct sway_seat *seat, struct sway_binding *binding)
...
 
-		*keycode = matches.keycode;
-		list_add(binding->keys, keycode);
+		sym2code[i] = matches.keycodes;
 	}
 
-	list_qsort(binding->keys, key_qsort_cmp);
-	binding->type = BINDING_KEYCODE;
+	// If any keycode maps to more than one keysym, use all combinations.
+	xkb_keycode_t **combos = cartesian_product(sym2code, num_syms);
```
kennylevinsen, line 756: So the implemented strategy is to collect all codes in advance in a list of lists, and then create a new list-of-list containing the Cartesian product of all the codes, and then finally create bindings using them. I can't help but feel that this is a bit too complicated, with a lot of cleanup needed.

How about a recursive strategy, where we just register the keybindings as we iterate? Something along the lines of this totally untested psuedocode:

```c

struct translate_ctx {
        struct xkb_keymap *keymap;
        struct sway_binding *binding;
        xkb_keycode_t *codes;
        int depth;
        bool error;
};

static void recursive_translate(struct xkb_keymap *keymap, xkb_keycode_t keycode, void *data) {
        struct translate_ctx *ctx = data;
        ctx->codes[ctx->depth] = keycode;

        if (depth == binding->keys->length) {
                // We have a pick for each symbol, so add it.
                if (!add_binding(ctx->binding, ctx->codes)) {
                        ctx->error = true;
                }
                return;
        }

        ctx->depth++;
        xkb_keymap_key_for_each(ctx->keymap, recursive_translate, ctx);
        ctx->depth--;
}

bool translate_binding(struct sway_binding *binding) {
        xkb_keycode_t codes[binding->keys->length] = { 0 };
        struct translate_ctx ctx = {
                .keymap = xkb_state_get_keymap(config->keysym_translation_state),
                .binding = binding,
                .codes = codes,
                .depth = 0,
        };
        xkb_keymap_key_for_each(ctx->keymap, recursive_translate, ctx);
        return !ctx.error;
}
```

sway/commands/bind.c
```diff
@@ -649,101 +646,144 @@ void seat_execute_command(struct sway_seat *seat, struct sway_binding *binding)
...
+
+	for (int i = 0; i < num_syms; i++) {
+		struct sway_binding * copy_binding = malloc(sizeof(struct sway_binding));
+		binding->type = BINDING_KEYCODE;
+		*copy_binding = *binding;
```
kennylevinsen, line 761: This isn't safe as it just copies the pointer to strings and lists, which will then be free'd multiple times later.

sway/commands/bind.c
```diff
@@ -467,10 +466,11 @@ static struct cmd_results *cmd_bindsym_or_bindcode(int argc, char **argv,
 	// sort ascending
 	list_qsort(binding->keys, key_qsort_cmp);
 
+	binding->command = join_args(argv + 1, argc - 1);
+
 	// translate keysyms into keycodes
 	if (!translate_binding(binding)) {
-		sway_log(SWAY_INFO,
-				"Unable to translate bindsym into bindcode: %s", argv[0]);
+		sway_log(SWAY_INFO,"Unable to translate bindsym into bindcode: %s", argv[0]);
```
kennylevinsen, line 473: What happened here?

sway/commands/bind.c
```diff
@@ -649,101 +646,144 @@ void seat_execute_command(struct sway_seat *seat, struct sway_binding *binding)
...
+		sym2code[i] = matches.keycodes;
 	}
 
-	list_qsort(binding->keys, key_qsort_cmp);
-	binding->type = BINDING_KEYCODE;
+	// If any keycode maps to more than one keysym, use all combinations.
+	xkb_keycode_t **combos = cartesian_product(sym2code, num_syms);
+
+	for (int i = 0; i < num_syms; i++) {
+		struct sway_binding * copy_binding = malloc(sizeof(struct sway_binding));
```
kennylevinsen, line 759: Three nits:
1. Should be `*copy_binding`, not `* copy_binding`. Same goes if there are more asterisks.
2. Our preferred style is to sizeof the variable itself ,so `sizeof(*copy_binding)` rather than a named type. It's not unusual that one ends up accidentally typing the wrong type, but typing the wrong name tends to be easier to catch.
3. NULL check

Applies elsewhere too, especially the first one.

sway/commands/bind.c
sahinf: Doesn't this just choose the first 3 keycodes in the keymap for all the bindings? I'm not sure exactly what the output of that is but it doesn't look like it spits out a cartesian map of the keysyms.

sway/commands/bind.c
sahinf: Could we add a check when freeing? Or would it be better to duplicate the command per each binding.

sway/commands/bind.c
sahinf: oops, will undo.

sway/commands/bind.c
sahinf: 1. Agreed, oops
2. I can definitely do that. Every other sizeof call in `bind.c` does it by type though. Should I change those?  
image
image
image
image
3. Adding.

sway/commands/bind.c
sahinf: In fact, correct me if I'm wrong, but I don't think it is mathematically possible to map out a cartesian combination of Sets A x B x C while also adding elements to those sets at the same time in the same "process"

kennylevinsen: Let's say we 3 symbols (A, B), and each of them have 2 keycodes (A1, A2, etc.).

1. The first iterator is called at depth 0, which is looking for all codes matching symbol A.
2. The first iterator finds `A1`, and stores it at `keycodes[0]`. It then calls the iterator for the next level.
3. The second iterator finds `B1`, and stores it at `keycodes[1]`. It sees that we reached the last symbol, and stores the binding for `[A1, B1]`.
4. The second iterator finds `B2`, and stores it at `keycodes[1]. It sees that we reached the last symbol, and stores the binding for `[A1, B2]`.
5. The second iterator finishes.
6. The first iterator finds `A2`, and stores it at `keycodes[0]`. It then calls the iterator for the next level.
7. Second goes through the paces again, but with `A2` at the first position.
8. The first iterator finishes, at which point we have stored: `[A1, B1], [A1, B2], [A2, B1], [A2, B2]`.

This should scale to any number of symbols, although at the cost of iterating over the keymap a lot more than using the cartesian product approach. However, it is at least to me a lot simpler.

sway/commands/bind.c
kennylevinsen: It's not impossible - the most likely solution being a ref-counted substructure that can be shared between bindings - but doubt it's worth it over just doing a deep copy to avoid the issue. For strings it's just an strdup, and we presumably don't have the other list set so we can just explicitly NULL it.

sway/commands/bind.c
kennylevinsen: Yeah, we went through the paces of updating all sizeofs in wlroots (https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4367), but we haven't done such an effort in sway. :/

> I can definitely do that. Every other sizeof call in bind.c does it by type though. Should I change those

No, unless the PR is specifically a tree-wide cleanup, stick to modifying the code you need to change. Otherwise you end up with a diff where it's hard to tell format changes from logic changes.

> Actually some of those examples above are the first invocation to the specified type, so they should refer to types anyway.

The first invocation of the specified type is the declaration, and you can already reference the name in the right-hand side of the declaring assignment. The following is perfectly valid and is the preferred construct:

```c
struct sway_binding *binding = calloc(1, sizeof(*binding));
```

sway/commands/bind.c
sahinf: > which will then be free'd multiple times later.

If we set pointers to null after freeing them, this shouldn't be an issue though right.

sway/commands/bind.c
sahinf:
```
if (keysym != binding->syms->items[ctx->depth]) {
            return;
        }
```
Ok now I see what you're going for with this addition. 
I think that should work. I'll try it out later and report back.

sway/commands/bind.c
kennylevinsen: That would only set them to `NULL` in one copy of the struct, the other copies would still have the pointer set and attempt to free it.

sway/commands/bind.c
sahinf: I couldn't get your algorithm to not segfault. Can you get it to work?

kennylevinsen: Yeah, I have a local patch with the recursive approach that works with bind and unbind, but haven't changed the retranslation logic.

I suspect that the easiest solution is to find and remove all "derived" bindings and retranslate them from their original description (which will either have to be extracted and deduplicated, or stored separately).

sahinf: Does your patch work on translating `bindsym --to-code Print+a`?

sahinf: The easiest solution could be to merge my patch that works.

kennylevinsen: > The easiest solution could be to merge my patch that works.

Your patch also does not handle retranslation of existing bindings correctly - it removed that logic.

sway/commands/bind.c
```diff
@@ -649,101 +647,157 @@ void seat_execute_command(struct sway_seat *seat, struct sway_binding *binding)
...
-	// a bindsym to translate
-	case BINDING_KEYSYM:
-		binding->syms = binding->keys;
-		binding->keys = create_list();
-		break;
-	// a bindsym to re-translate
-	case BINDING_KEYCODE:
-		list_free_items_and_destroy(binding->keys);
-		binding->keys = create_list();
-		break;
```
kennylevinsen, line 712: For clarity of discussion, this is the retranslation logic that this (and my WIP) patch does not currently handle correctly. This serves to redo the `--to-code` conversion if the *primary* keymap changes, i.e., if you go from `us,ru` to `ru,us` or if you first set your non-standard keymap after all the bindyms. `--to-code` is not locked at the time of binding, but instead respect the configured primary keymap rather than the current keymap.

sahinf: I was already testing by switching between us,ru and it seemed to work in both cases. Like I would toggle the layout 

```config
### Input configuration
input 1:1:AT_Translated_Set_2_keyboard {
    xkb_layout us,ru
    xkb_options ctrl:swapcaps,grp:alt_shift_toggle
}
```

Which input did you test with and failed? So I can try on my machine too.

kennylevinsen:
```sh
swaymsg 'input * xkb_keymap us'
swaymsg 'bindsym --to-code mod1+a exec notify-send hi'
# test that alt+a sends hi according to a US QWERTY layout
swaymsg 'input * xkb_keymap fr'
# Test that alt+q now sends hi, as the French AZERTY layout is the new
# "first configured layout" that `--to-code` should respect
```

sahinf: That worked for me.

sahinf: > Test that alt+q now sends hi

Hold up I just saw this. That seems like a weird use case to me. If I bind something to code, I expect those keys to not change positions regardless of the layout.

sahinf: What's the purpose of even binding something to keycodes if you have to look up where the keycode is after changing layout? might as well bindsym.

kennylevinsen: There’s a long discussion in the original PR, but the TL;DR is that it makes it so that the bindsym behavior is not dependent on command ordering - whether you set keyboard layout first or last in the config, the result is the same.

This makes it consistent with bindcode and regular bindsym, which also give the same result regardless of ordering in respect to keyboard configuration.

sahinf: I see, but that still completely defeats the purpose of binding things to keycodes. Isn't the purpose of bindcode so that when you change layouts, it will be the same physical keys on the keyboard?

sahinf: If bindcode and bindsym --to-code do not stick with the layout that my keyboard is configured in, that makes both commands obsolete. What you really want to do in that example is bindsym.

kennylevinsen: `bindsym --to-code` is not the same as `bindcode`. Because it uses symbols and depends on a layout, it is implemented so that it makes the binding to match the keycodes of the at any time first configured layout.

Quoting the manpage:

```
	Bindings to keysyms are layout-dependent. This can be changed with the
	_--to-code_ flag. In this case, the keysyms will be translated into the
	corresponding keycodes in the first configured layout.
```

This means that when you have a keyboard configured for `xkb_layout us,fr`, a `bindsym --to-code` will  use the sequence according to the US keyboard layout even if you switch to French using `xkb_switch_layout` (`man 5 sway-input`) or an XKB-builtin layout switch toggle.

If the translation *only* happened once, then it would mean that the result depended on the order of `bindsym` and `xkb_layout` in your config file, which we do not want to be the case.

kennylevinsen: > If bindcode and bindsym --to-code do not stick with the layout that my physical keyboard is configured in, that makes both commands obsolete. What you really want to do in that example is bindsym.

This is how the functionality is defined and documented. The PR and issue is just about adding support for symbols emitted by more than one keycode.

kennylevinsen: > What's the "original PR" btw?

https://github.com/swaywm/sway/pull/3058

sahinf: > bindsym --to-code is not the same as bindcode

Thanks.

sway/commands/bind.c
sahinf: If you're going to merge that commit in, should I just close this PR? And timeline?

sway/commands/bind.c
sahinf: > This is how the functionality is defined and documented

> Test that alt+q now sends hi

I should have tested on default sooner but I just realized that this behavior doesn't work with the current code as the documentation says. Can you verify?

If so, I can start looking into making it work as intended.

sahinf:
@emersion @kennylevinsen I think adding multiple bindings per each found duplicate would require cleaning up each of those duplicates when re-translating, which seems excessively complicated given the problem I was trying to fix. 

But I realized someone may have already thought of this back in 2018 in the commit https://github.com/swaywm/sway/commit/a056419ad which introduced [state_keysyms_translated](https://github.com/swaywm/sway/blob/861dde100ab5536bea190b078c6c51adb6814be5/include/sway/input/keyboard.h#L64C1-L64C54). 

This [`get_active_binding` ](https://github.com/swaywm/sway/blob/861dde100ab5536bea190b078c6c51adb6814be5/sway/input/keyboard.c#L436C32-L436C56) tries to utilize it, but I don't believe we actually ever fill it:
```bash
0$ rg state_keysyms_translated
include/sway/input/keyboard.h
64:     struct sway_shortcut_state state_keysyms_translated;

sway/input/keyboard.c
373:            update_shortcut_state(&keyboard->state_keysyms_translated,
436:    get_active_binding(&keyboard->state_keysyms_translated,
468:            get_active_binding(&keyboard->state_keysyms_translated,
```


And this comment for `pressed_keycodes` says it's specifically for when a keycode generates multiple key ids. I assume that is referring to when a keycode is associated with multiple keysyms (I've never seen "key id" lingo in libxkbcommon) 
```
struct sway_shortcut_state {
	/**
	 * A list of pressed key ids (either keysyms or keycodes),
	 * including duplicates when different keycodes produce the same key id.
	 *
	 * Each key id is associated with the keycode (in `pressed_keycodes`)
	 * whose press generated it, so that the key id can be removed on
	 * keycode release without recalculating the transient link between
	 * keycode and key id at the time of the key press.
	 */
	uint32_t pressed_keys[SWAY_KEYBOARD_PRESSED_KEYS_CAP];
	/**
	 * The list of keycodes associated to currently pressed key ids,
	 * including duplicates when a keycode generates multiple key ids.
	 */
	uint32_t pressed_keycodes[SWAY_KEYBOARD_PRESSED_KEYS_CAP];
	uint32_t last_keycode;
	uint32_t last_raw_modifiers;
	size_t npressed;
	uint32_t current_key;
};
```

Do ya'll think it's worthwhile to try and handle duplicates through this? Might require a lot less changes.

sahinf:
In fact, it would make the most sense to be consistent with `wev` behavior. 
My original plan to simply choose the first keycode it discovered is consistent with `wl_keyboard` in wev.

kennylevinsen:
> @emersion @kennylevinsen I think adding multiple bindings per each found duplicate would require cleaning up each of those duplicates when re-translating,

Yeah, that's what I meant by:

> I suspect that the easiest solution is to find and remove all "derived" bindings and retranslate them from their original description (which will either have to be extracted and deduplicated, or stored separately).

I don't think it's particularly hard, but it _is_ somewhat annoying. It would require being able to recognize synthetic/derived/translated bindings as opposed to "real" ones to make them easy to remove, and would require storing the original binding in an inert way for reference for retranslation. That could be done with a `parent` pointer: Normal bindings don't have a parent, synthetic bindings point to the binding they were generated from. The parent is then stored in a different list not considered for actual execution of keybindings.

> But I realized someone may have already thought of this back in 2018 in the commit https://github.com/swaywm/sway/commit/a056419ad which introduced [state_keysyms_translated](https://github.com/swaywm/sway/blob/861dde100ab5536bea190b078c6c51adb6814be5/include/sway/input/keyboard.h#L64C1-L64C54).

Everything there is about the keyboard state in the currently active keymap, so that would not be immediately useful.

sahinf:
Closing this because #8358 offers a less complex solution

PR #8341, opened 2024-09-09
config/output: Fix interaction with output management, take 2

kennylevinsen:
Slightly different approach from https://github.com/swaywm/sway/pull/8330, as per review comment.


Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8343, opened 2024-09-12
lock: fix swaylock cursor remaining visible

### LINKED ISSUE #8342: Session lock surface receives no pointer enter until motion
- **Sway version:** 1.10-dev-be840f73 (Sep 11 2024, branch 'master')
- **Debug log:** [sway.log](https://github.com/user-attachments/files/16969216/sway.log) (nothing useful)
- **Description:** There seems to be a regression of https://github.com/swaywm/sway/pull/7690 (fixing https://github.com/swaywm/swaylock/issues/333) after the port to the scene graph API. To reproduce: launch sway (default config), launch swaylock (no arguments), observe visible cursor until it is moved. The cursor_rebase_all call in lock.c no longer triggers pointer enter.

This happens because at the time of the cursor_rebase_all call, while the wlr_surface for the lock surface has its client buffer initialized correctly, the wlr_scene_surface's scene buffer is null, so when sway/input/cursor.c traverses the scene tree it doesn't hit the lock surface.
  
The cursor rebase happens when the lock surface role's commit callback calls wlr_surface_map, triggering sway/lock.c's surface map callback. The scene buffer is initialized immediately *after* that, from the surface commit signal, by wlroots/types/scene/surface.c:204.

I don't know if this is a bug in sway or wlroots. Two small fixes which work but may be incorrect:

- Listen for surface commit rather than surface map in sway/lock.c for focus / cursor rebasing.
- Move the surface_reconfigure call from wlroots/types/scene/surface.c:204 to a surface map callback.

The change in lock.c is tiny and seems completely harmless; I'm happy to make it a PR if appropriate.

Luminiscental:
Fixes https://github.com/swaywm/sway/issues/8342 by listening to surface commit rather than surface map for focus and cursor rebasing on session lock.

### DIFF
--- a/sway/lock.c
+++ b/sway/lock.c
@@ -25,7 +25,7 @@ struct sway_session_lock_output {
 
 	// invalid if surface is NULL
 	struct wl_listener surface_destroy;
-	struct wl_listener surface_map;
+	struct wl_listener surface_commit;
 };
 
 static void focus_surface(struct sway_session_lock *lock,
@@ -59,8 +59,8 @@ static void refocus_output(struct sway_session_lock_output *output) {
 	}
 }
 
-static void handle_surface_map(struct wl_listener *listener, void *data) {
-	struct sway_session_lock_output *surf = wl_container_of(listener, surf, surface_map);
+static void handle_surface_commit(struct wl_listener *listener, void *data) {
+	struct sway_session_lock_output *surf = wl_container_of(listener, surf, surface_commit);
 	if (surf->lock->focused == NULL) {
 		focus_surface(surf->lock, surf->surface->surface);
 	}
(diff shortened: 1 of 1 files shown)


kennylevinsen:
Hmm, this will cause the cursor to be rebased on every single surface commit - for an animated lockscreen (e.g., hyprlock), that will be on every animated frame. That seems too aggressive.

Nefsen402:
It just rebases the cursor which should be nearly free.

kennylevinsen:
> It just rebases the cursor which should be nearly free.

```c
static void handle_rebase(struct sway_seat *seat, uint32_t time_msec) {
	struct seatop_default_event *e = seat->seatop_data;
	struct sway_cursor *cursor = seat->cursor;
	double sx = 0.0, sy = 0.0;
	e->previous_node = node_at_coords(seat,

	if (surface) {
		if (seat_is_input_allowed(seat, surface)) {
			wlr_seat_pointer_notify_enter(seat->wlr_seat, surface, sx, sy);
			wlr_seat_pointer_notify_motion(seat->wlr_seat, time_msec, sx, sy);
		}
	} else {
		cursor_update_image(cursor, e->previous_node);
		wlr_seat_pointer_notify_clear_focus(seat->wlr_seat);
	}
}
```

It sends pointer events, as it was only meant to be called when focus changes.

PR #8344, opened 2024-09-13
config/output: support DRM_FORMAT_ARGB8888

sdirkwinkel:
Some display output hardware [1] doesn't support any of the current formats, but works with ARGB8888. Fall back to it if available.

[1] https://github.com/torvalds/linux/blob/196145c606d0f816fd3926483cb1ff87e09c2c0b/drivers/gpu/drm/xlnx/zynqmp_disp.c#L313

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -783,6 +783,7 @@ static bool search_render_format(struct search_context *ctx, size_t output_idx) 
 		DRM_FORMAT_XRGB2101010,
 		DRM_FORMAT_XBGR2101010,
 		DRM_FORMAT_XRGB8888,
+		DRM_FORMAT_ARGB8888,
 		DRM_FORMAT_INVALID,
 	};
 	if (render_format_is_bgr(wlr_output->render_format)) {


emersion (approved):
Indeed, sounds good to me!

sdirkwinkel:
@kennylevinsen @emersion
Sorry, maybe I should have written this in the original description:

I also got it working by setting the default here (maybe it could be set to a format the hardware actually supports?):
https://salsa.debian.org/swaywm-team/wlroots/-/blob/debian/bookworm/types/output/output.c?ref_type=heads#L388
I tried something like this in wlroots first (with older sway, haven't tried yet with current sway):
```
diff --git a/types/output/render.c b/types/output/render.c
index 985b93a9..cc547fcf 100644
--- a/types/output/render.c
+++ b/types/output/render.c
@@ -17,6 +17,12 @@ bool wlr_output_init_render(struct wlr_output
*output,
        assert(output->allocator == NULL && allocator != NULL);
        assert(output->renderer == NULL && renderer != NULL);
 
+       const struct wlr_drm_format_set *display_formats =
wlr_output_get_primary_formats(output, allocator->buffer_caps);
+
+       if(display_formats->formats[0] != NULL) {
+               output->render_format = display_formats->formats[0]-
>format;
+       }
        uint32_t backend_caps = backend_get_buffer_caps(output-
>backend);
        uint32_t renderer_caps =
renderer_get_render_buffer_caps(renderer);
```
I didn't end up submitting that, because sway already iterates over format.
Could wlr_output_get_primary_formats be used in sway to avoid unnecessary tries?

kennylevinsen:
> Could wlr_output_get_primary_formats be used in sway to avoid unnecessary tries?

We already use `wlr_output_get_primary_formats` to skip unsupported formats:

https://github.com/swaywm/sway/blob/f957c7e658871c27935f88f6e1d18b9db67f3808/sway/config/output.c#L795-L805

The "problem" is that many GPUs support *all* the formats on the list, so if an output is failing to light up for some non-format reason (e.g., bandwidth limitation), we end up allocating and testing all those buffer types to no avail. This was already the case before this patch, with the 10-bit BGR test being quite bogus on most GPUs.

IMO, not something we need to worry about *yet*. If this ends up being enough of an issue (e.g., if the list grows longer or people experience noticeable delays), maybe we could add a heuristic to only try alternate format variations if the primary isn't supported to reduce the number of tested formats (e.g., BGR only if RGB isn't supported, ARGB only if XRGB isn't supported).

The "proper" fix is kernel support for doing a commit test for a buffer format without an actual buffer allocation, in which case we could just cycle through and test all supported formats without worry.

kennylevinsen:
Made a draft of the heuristic approach here for the heck of it: https://github.com/swaywm/sway/pull/8345

PR #8347, opened 2024-09-16
Include wlroots checkout helper for bisecting

WhyNotHugo:
I find this useful when bisecting, since the latest wlroots usually won't work on older sway commits.

Hopefully it can help others too.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -243,3 +243,29 @@ error_session:
 	return NULL;
 }
 ```
+
+## Bisecting
+
+When bisecting, it is often necessary to build older commits with their
+contemporary version of wlroots:
+
+```python
+#!/usr/bin/env python3
+#
+# Check out the wlroots commit that was the current master at the time that the current
+# sway HEAD was committed. Requires pygit2. E.g.: apk add py3-pygit2
+
+import pygit2
+
+sway = pygit2.Repository(".")
+head = sway.revparse_single("HEAD")
+sway_time = head.commit_time
+
+wlroots = pygit2.Repository("subprojects/wlroots")
+master = wlroots.revparse_single("master")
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -243,3 +243,29 @@ error_session:
...
+
+import pygit2
+
+sway = pygit2.Repository(".")
+head = sway.revparse_single("HEAD")
+sway_time = head.commit_time
+
+wlroots = pygit2.Repository("subprojects/wlroots")
+master = wlroots.revparse_single("master")
+for commit in wlroots.walk(master.id, pygit2.enums.SortMode.TIME):
```
bl4ckb0ne, line 266: I'm not sure if using the time is always the right thing here, have you encountered any issue with that so far?

WhyNotHugo: > I'm not sure if using the time is always the right thing here

In this context, `TIME` is the _commit_ time specifically, which is the time when the commit was merged into the master branch.

The time shown by `git log` is the _author_ time. Using that one does not work; it yields bogus results since some commits are authored months before being merged.

> have you encountered any issue with that so far?

When wlroots version N is released, the version is immediately bumped to N+1. From that point on, meson will complain that it has wlroots N+1 but expects version N.

In my experience bisecting sway, very few commits have this issue. Usually I skip these. If you need to dive into that range, you'll can just check out the wlroots tag.

CONTRIBUTING.md
```diff
@@ -243,3 +243,29 @@ error_session:
...
+    if commit.commit_time < sway_time:
+        break
+
+wlroots.checkout_tree(commit)
```
WhyNotHugo, line 270: For some edge cases, this produces less conflicts:

```suggestion
wlroots.set_head(commit.id)
wlroots.checkout_head()
```

PR #8349, opened 2024-09-17
ext-session-lock: Do not use commit listener to arrange

kennylevinsen:
Arranging lock surfaces rely on the sway_output width and height being updated, but these are only updated after the commit has been completed and all commit listeners have executed. This means that the lock surfaces will not be appropriately scaled to match a change in output dimensions, and may reveal what is under the lock background.

Replace the implicit arrange through the output commit listener with an explicit arrange after the output configuration is finalized.

This might have regressed by other transition away from output commit listeners for other arrange tasks, but even then it would have erroneously relied on signalling order.

Note: There is still a one-frame glitch on scale change as we only have the new output layout geometry for arrange *after* the output commits have gone through, as the output layout is feeding off commit listeners itself and cannot be updated in advance.

### DIFF
--- a/include/sway/lock.h
+++ b/include/sway/lock.h
@@ -1,0 +1,6 @@
+#ifndef _SWAY_LOCK_H
+#define _SWAY_LOCK_H
+
+void arrange_locks(void);
+
+#endif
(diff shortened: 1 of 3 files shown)


Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8331, opened 2024-09-05
output: More redundant reconfigures

kennylevinsen:
Two more places where we did too much stuff on output changes - these are covered by `apply_output_configs`.

PR #8326, opened 2024-09-03
Reduce output configurations and rearranges

kennylevinsen:
Whenever output configurations occurred we would end up re-arranging everything a gajillion times due to several components and listeners all responding to the change by calling `arrange_*`.

We now have a central location where changes occur from - by relying on that we can get down to a single `arrange_layers` per output and a shared `arrange_root` + output manager update... In most cases, as arrange_layers still calls arrange_output/arrange_popups on its own for now.

PR #8353, opened 2024-09-20
1.10 backport

kennylevinsen:
Backports: 
- #8349
- #8341
- #8332
- #8331
- #8326

cc @emersion


kennylevinsen:
If there's something we don't want, I can split it - just walked through my commits on master one by one, and it works in a quick local test.

I did have a one-time glitch where after testing on/off and output manager/power command interactions, one output was arranged with an odd, slightly smaller geometry (using an odd 3199 width instead of 3840), causing an overlap - but it wasn't immediately reproducible, so I suspect it might be a latent bug in master too.

emersion:
Thanks!

PR #8354, opened 2024-09-20
Reduce modesets and input configuration

kennylevinsen:
During a fresh sway start, we end up doing a bunch of input and output configuration, as each input command triggers input device configuration, and each output command triggers modeset. In addition, because output manager failed to clear the modeset timer, we could end up with another bogus modeset after output manager apply.

Batch input/output configuration during the initial config load, and skip the initial modeset altogether as we don't have any output devices at this point - those come later, and will trigger modeset on their own.

We still waste up to a few milliseconds per input command on my machine due to config store making some costly checks.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -704,6 +704,8 @@ struct output_config *find_output_config(struct sway_output *output);
 
 void free_output_config(struct output_config *oc);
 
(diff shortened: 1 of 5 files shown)


Nefsen402 (approved):
Thanks!

PR #8355, opened 2024-09-20
commands/gaps: Check config->reading instead

kennylevinsen:
Checking if the config is not active or is reloading is just a convoluted way of checking if the config is being read.


emersion (approved):
Thanks!

PR #8357, opened 2024-09-22
Print `SWAYSOCK` after it is set for this Sway instance

sahinf:
`log_env` attempts to query SWAYSOCK env variable before ipc server is initialized. Also cleans up unused files from main.c

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -13,7 +13,6 @@
 #include <unistd.h>
 #include <wlr/util/log.h>
 #include <wlr/version.h>
-#include "sway/commands.h"
 #include "sway/config.h"
 #include "sway/server.h"
 #include "sway/swaynag.h"
@@ -23,7 +22,6 @@
 #include "ipc-client.h"
 #include "log.h"
 #include "stringop.h"
-#include "util.h"
 
 static bool terminate_request = false;
 static int exit_value = 0;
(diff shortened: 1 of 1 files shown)


emersion:
The purpose of `log_env` is not to print env variables set by Sway itself. It's to log env vars set before Sway starts, for debugging purposes when we receive a bug report.

sahinf:
I see. Since it prints out an empty variable currently, would you be against moving it to after SWAYSOCK is set?

emersion:
I'd prefer to keep it the way it currently is, because it's sometimes useful to know whether the reporter ran Sway inside Sway or not.

emersion:
(More generally, I'd like to keep this clean of any env modification made by Sway itself.)

sahinf:
Could we rely on `00:00:00.088 [wlr] [wayland] unable to lock lockfile /run/user/1000/wayland-1.lock, maybe another compositor is running` instead for nested users? I currently watch my socket path to see the SWAYLOCK variable of nested containers.

emersion:
The point is, this particular log is not for end users. The only reliable way to grab the SWAYSOCK variable is to execute a command from inside Sway. Logs are not stable anyways, we can change them at any time.

sahinf:
I don't see the harm in making it a bit more useful for developers since it's current purpose still remains through the lockfile check, but I understand. I will close this.

PR #8359, opened 2024-09-25
swaybar: Fix 100% cpu usage if dbus dies.

sahinf:
Currently, swaybar does not gracefully die if it detects that the dbus connection was lost. Although it's not recommended to restart dbus without restarting the compositor, it can very easily happen. In the case it does, compositor's tray should not consume 100% cpu until it has to be force killed.

### DIFF
--- a/swaybar/bar.c
+++ b/swaybar/bar.c
@@ -508,7 +508,7 @@ void bar_run(struct swaybar *bar) {
 	}
 #if HAVE_TRAY
 	if (bar->tray) {
-		loop_add_fd(bar->eventloop, bar->tray->fd, POLLIN, tray_in, bar->tray->bus);
+		loop_add_fd(bar->eventloop, bar->tray->fd, POLLIN, tray_in, bar);
 	}
 #endif
 	while (bar->running) {
--- a/swaybar/tray/tray.c
+++ b/swaybar/tray/tray.c
@@ -1,4 +1,5 @@
 #include <cairo.h>
+#include <poll.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
@@ -90,9 +91,16 @@ void destroy_tray(struct swaybar_tray *tray) {
 }
 
 void tray_in(int fd, short mask, void *data) {
-	sd_bus *bus = data;
+	struct swaybar *bar = data;
 	int ret;
-	while ((ret = sd_bus_process(bus, NULL)) > 0) {
+
+	if (mask & (POLLHUP | POLLERR)) {
+        sway_log(SWAY_ERROR, "D-Bus connection closed unexpectedly");
+		bar->running = false;
+		return;
+    }
+
+	while ((ret = sd_bus_process(bar->tray->bus, NULL)) > 0) {
 		// This space intentionally left blank
 	}
 	if (ret < 0) {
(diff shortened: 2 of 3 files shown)


sahinf:
To reproduce:
  1. `systemctl --user restart dbus`
  2. `sudo htop` to see +100% cpu utilization by swaybar
  3. `perf record` & `perf report` show high cycles spent in `loop_poll`

swaybar/tray/tray.c
```diff
@@ -90,9 +91,17 @@ void destroy_tray(struct swaybar_tray *tray) {
...
 
 void tray_in(int fd, short mask, void *data) {
-	sd_bus *bus = data;
+	struct swaybar *bar = data;
 	int ret;
-	while ((ret = sd_bus_process(bus, NULL)) > 0) {
+
+	if (mask & (POLLHUP | POLLERR)) {
+        sway_log(SWAY_ERROR, "D-Bus connection closed unexpectedly");
+		sd_bus_unref(bar->tray->bus);
```
emersion, line 99: It sounds like this `sd_bus_unref()` call is something which should be done by `bar_teardown()` or similar, instead of being the responsibility of all spots we set `running = false`.

swaybar/tray/tray.c
```diff
@@ -37,7 +38,7 @@ struct swaybar_tray *create_tray(struct swaybar *bar) {
 	sway_log(SWAY_DEBUG, "Initializing tray");
 
 	sd_bus *bus;
-	int ret = sd_bus_open_user(&bus);
+	int ret = sd_bus_open_user_with_description(&bus, "Swaybar tray bus");
```
emersion, line 41: Since we have only a single `sd_bus *`, is it really useful to set the description?

emersion (changes requested):
Apart from this, looks good!

swaybar/tray/tray.c
sahinf: Nope. I was thinking dbus monitors might wanna know but that's just an assumption I made. Removed.

swaybar/tray/tray.c
```diff
@@ -90,9 +91,17 @@ void destroy_tray(struct swaybar_tray *tray) {
...
 
 void tray_in(int fd, short mask, void *data) {
-	sd_bus *bus = data;
+	struct swaybar *bar = data;
 	int ret;
-	while ((ret = sd_bus_process(bus, NULL)) > 0) {
+
+	if (mask & (POLLHUP | POLLERR)) {
+        sway_log(SWAY_ERROR, "D-Bus connection closed unexpectedly");
+		bar_teardown(bar);
```
emersion, line 99: `bar_teardown()` is already called when the main loop exits, no?

swaybar/bar.c
```diff
@@ -537,6 +537,7 @@ static void free_seats(struct wl_list *list) {
 void bar_teardown(struct swaybar *bar) {
 #if HAVE_TRAY
 	destroy_tray(bar->tray);
+	sd_bus_unref(bar->tray->bus);
```
emersion, line 540: Doesn't this cause a use-after-free?

Sounds cleaner to move `sd_bus_unref()` into `destroy_tray()`.

emersion:
Can you squash the commits together?

swaybar/bar.c
sahinf: Yes and also destroy_tray already calls `sd_bus_flush_close_unref` so I removed this line.

swaybar/tray/tray.c
sahinf: Yes. Just setting it to not running should be enough.

emersion (approved):
Thanks!

PR #8360, opened 2024-09-29
debug: Add dump_scene command

kennylevinsen:
The dump_scene commands dumps the scene tree to stderr.

Draft as it's a wee bit hacky and we don't currently have a concept of debug commands like this, but as it can be useful I thought I'd share it.


Nefsen402:
https://github.com/Nefsen402/sway/commit/ce56b75e88aeaeab6ac9d30e0a6ba54bc193cc6e

PR #8362, opened 2024-09-29
Add support for alpha-modifier-v1

emersion:


kennylevinsen (approved):
LGTM

Nefsen402:
This is not a correct implementation of alpha_modifier. output_configure_scene will overwrite the alpha that scene tries to set itself.

PR #8363, opened 2024-09-29
release: push tags before creating GitHub release

emersion:
Otherwise the GitHub release isn't attached to the Git tag.


kennylevinsen (approved):
LGTM, although a fun caveat of `follow-tags` is that it only pushes annotated tags, not simple tags. Not an issue for our use-case though.

emersion:
I think ideally we'd only push the specific tag that we just created, but that seems to require grabbing the remote name as well…

PR #8366, opened 2024-09-29
Fix alpha-modifier-v1

Nefsen402:
`output_configure_scene()` will overwrite any alpha that scene tries to set itself. We need to teach `output_configure_scene()` about alpha modifier.

Fixes: #8362


emersion (approved):
Oops, forgot about this… Thanks!

emersion:
Ah, seems like there's a missing include still.

PR #8370, opened 2024-10-01
input/mouse: bugfix button2 being interpreted as trying to move the

### LINKED ISSUE #8334: i3 inconsistency with "bindsym --release --border button2 kill"
**Description:**
Neither i3 nor Sway have close buttons on the titlebar, so I added the following to my config to be able to close windows just using my mouse.
`bindsym --release --border button2 kill`
The idea is to middle-click on the titlebar to close. This works perfectly on i3. Not on Sway.

**Problems on Sway:**
- This works fine with floating windows only.
- For tiling windows, it does not work with titlebars. To close a tiled window, I have to aim pixel perfect to a 2px thick border. (In my case it's also invisible due to my theming)
- Drag and drop in my file manager (Dolphin) stops working after doing a middle click on any titlebar/border.

sahinf:
container.

Man sway(5) specifies that when tiling_drag is enable, the floating_mod can be used to drag tiling, as well as floating containers. However the current code indiscriminately assumes any button press to be intended for moving the container, consequently causing an unintended call to `seatop_move_tilting:handle_button` rather than
`seatop_default:handle_button` to pass
`state=WL_POINTER_BUTTON_STATE_RELEASED` to `get_active_mouse_binding`

My idea was to make 'Handle moving a tiling container' follow the same path as 'Handle moving a floating container' because the initial call to handle moving a floating correctly exits that branch and ends up passing the RELEASED state to `get_active_mouse_binding`.

Fixes #8334

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -491,7 +491,9 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 	// Handle moving a tiling container
 	if (config->tiling_drag && (mod_pressed || on_titlebar) &&
 			state == WL_POINTER_BUTTON_STATE_PRESSED && !is_floating_or_child &&
-			cont && cont->pending.fullscreen_mode == FULLSCREEN_NONE) {
+			cont && cont->pending.fullscreen_mode == FULLSCREEN_NONE &&
+			button == (config->floating_mod_inverse ? BTN_RIGHT : BTN_LEFT)) {
+
 		// If moving a container by its title bar, use a threshold for the drag
 		if (!mod_pressed && config->tiling_drag_threshold > 0) {
 			seatop_begin_move_tiling_threshold(seat, cont);


emersion:
Hm, I wonder if we should use `floating_mod_inverse` for the titlebar case? The floating logic seems to use `floating_mod_inverse` only in the case of mod+click, not in the case of titlebars.

sahinf:
[`Handle floating move`](https://github.com/swaywm/sway/blob/c90cb37b2a0861548461daa9b75d75317e01b679/sway/input/seatop_default.c#L459C30-L459C56) works on both title bar and non title if the modifier key is pressed. The reason that `--release` bindings work when clicking on floating title bars is because of that check.    


I'm not sure how else to avoid entering that branch after returning from the initial [WL_POINTER_BUTTON_STATE_PRESSED call](https://github.com/swaywm/sway/blob/c90cb37b2a0861548461daa9b75d75317e01b679/sway/input/seatop_default.c#L358C1-L362C3).

sahinf:
** Currently, any mouse click on a tilting title bar is interpreted as moving the container. The changes force it into either left click or right click with the modifier inverse config.

emersion:
My suggestion is to check for `button == config->floating_mod_inverse ? BTN_RIGHT : BTN_LEFT` only for the `mod_pressed` case, and check for `button == BTN_LEFT` for the `on_titlebar` case.

sahinf:
Moving floating containers don't make that distinction though. Also, the manpages do not specify that floating mod should not work on titlebar.

```  
floating_modifier <modifier> [normal|inverse]
  When the modifier key is held down, you may hold left click to move windows,
  and right click to resize them. Setting modifier to none disables this feature.
  If inverse is specified, left click is used for resizing and right click for
  moving.
           
```

```
tiling_drag  enable|disable|toggle
  Sets whether or not tiling containers can be dragged with the mouse.
  If enabled (default), the floating_mod can be used to drag tiling,
  as well as floating, containers. Using the left mouse button on
  title bars without the floating_mod will also allow the container to
  be dragged. toggle should not be used in the config file.
```
Doesn't "Using the left mouse button on title bars **without the floating_mod** will also allow the container to be dragged." mean that it's only `BTN_LEFT` if floating_mod is not set?

emersion:
> Moving floating containers don't make that distinction though

Oh right. I was reading the resize logic. The move logic seems inconsistent with the resize logic, but that's orthogonal to this PR.

> the manpages do not specify that floating mod should not work on titlebar

Would need to check what i3 does.

emersion (approved):
Let's go with this for now. I still think we need to fix the inconsistency between move and resize, but that's for later.

Thanks!

PR #8374, opened 2024-10-06
Fix cursor events on screen lock/unlock

OkamiW:
After the scene graph refactor, node_at_coords no longer return the lock surface after the screen is locked, thereby cursor_rebase_all in handle_surface_map won't work.

This commit make node_at_coords return lock surface when locked.

I also found that the cursor would disappear after screen is unlocked, thereby adding a cursor_rebase_all in handle_unlock.

### DIFF
--- a/sway/lock.c
+++ b/sway/lock.c
@@ -59,12 +59,17 @@ static void refocus_output(struct sway_session_lock_output *output) {
 	}
 }
 
-static void handle_surface_map(struct wl_listener *listener, void *data) {
+static void update_focus(void *data) {
+	struct wl_listener *listener = data;
 	struct sway_session_lock_output *surf = wl_container_of(listener, surf, surface_map);
 	if (surf->lock->focused == NULL) {
 		focus_surface(surf->lock, surf->surface->surface);
 	}
 	cursor_rebase_all();
+}
+
+static void handle_surface_map(struct wl_listener *listener, void *data) {
+	wl_event_loop_add_idle(server.wl_event_loop, update_focus, listener);
 }
 
 static void handle_surface_destroy(struct wl_listener *listener, void *data) {
@@ -234,6 +239,7 @@ static void handle_unlock(struct wl_listener *listener, void *data) {
 		struct sway_output *output = root->outputs->items[i];
 		arrange_layers(output);
 	}
+	cursor_rebase_all();
 }
 
 static void handle_abandon(struct wl_listener *listener, void *data) {
--- a/sway/lock.c
+++ b/sway/lock.c
@@ -331,11 +331,31 @@ void sway_session_lock_add_output(struct sway_session_lock *lock,
 	}
 }
 
+struct is_lock_surface_data {
+	struct wlr_surface *surface;
+	bool is_lock_surface;
+};
+
+static void is_lock_surface(struct wlr_surface *surface, int sx, int sy, void *data) {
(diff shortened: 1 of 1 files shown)


sway/input/cursor.c
```diff
@@ -63,6 +63,11 @@ struct sway_node *node_at_coords(
 		}
 	}
 
+	if (server.session_lock.lock) {
+		*surface = server.session_lock.lock->focused;
```
emersion, line 67: `focused` represents the keyboard focus, but this codepath is about the pointer. Moreover, the surface-local coordinates `sx`/`sy` are left unset.

The session lock surfaces are added to `output->layers.session_lock`, so I wonder why this doesn't work?

cc @Nefsen402

kennylevinsen:
Hmm, this is a bit of an event ordering issue with `wlr_scene`.

The order of events here is:
1. Surface is committed
2. Surface role commit implementations are called
3. Surface role commit handlers call `wlr_surface_map`
4. Surface role map implementations are called
5. Map event fires
6. Commit event fires
7. Scene surface helper attaches new buffer to the surface

Dealing with this would require an "initial commit" sort of commit handler... :/

Nefsen402:
I've been pondering for a while if we should get rid of all the cursor rebase stuff and add some sort of callback or just go through with `wlr_scene_cursor` finally. Scene could implement cursor rebasing on its own and all these problems would go away similar to how we handled xwayland restacking.

kennylevinsen:
It could be useful, but there are some caveats.

One is that the effect of `cursor_rebase` is currently conditional on the ongoing seatop, with only the default seatop performing focus updates. It could also be argued that it's a bug that `cursor_rebase` doesn't also update keyboard focus when `focus_follows_mouse always` is set.

Another is that, even if scene have a helper, there should preferably still be a way for compositors to do things on their own.

OkamiW:
It seems that Labwc has a fix for the same problem: https://github.com/labwc/labwc/pull/1858
`wl_event_loop_add_idle` can be used here to delay the actual operations after the buffer has been attached.
This should be a proper fix for the issue.

OkamiW:
Gentle ping.

OkamiW:
Any update on this?

sway/lock.c
```diff
@@ -325,11 +332,26 @@ void sway_session_lock_add_output(struct sway_session_lock *lock,
...
+	return wlr_surface_has_subsurface(subsurface->parent, surface);
+}
+
 bool sway_session_lock_has_surface(struct sway_session_lock *lock,
 		struct wlr_surface *surface) {
 	struct sway_session_lock_output *lock_output;
 	wl_list_for_each(lock_output, &lock->outputs, link) {
-		if (lock_output->surface && lock_output->surface->surface == surface) {
+		if (lock_output->surface && (lock_output->surface->surface == surface
+					     || wlr_surface_has_subsurface(lock_output->surface->surface, surface))) {
```
emersion, line 354: This only handles one level of sub-surfaces. If we want to handle arbitrary nested sub-surfaces, we should use `wlr_surface_for_each_surface()`.

emersion:
In general the approach of delaying the focus updates are more of a workaround than a proper fix. However it's not clear what the proper fix could look like.

sway/lock.c
OkamiW: The code has been updated to use `wlr_surface_for_each_surface`.

Issue #8375, opened 2024-10-06
sway crashes if switch to another workspace with surface when IME popup is shown 


OkamiW:
- **Sway Version:**
  `sway version 1.10-dev-c90cb37b (Oct  6 2024, branch 'master')
`

- **Stack Trace:**
```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1   in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:78
#2   in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3   in __GI_abort () at abort.c:79
#4   in __assert_fail_base (fmt= "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry= "node", file=file@entry= "types/scene/wlr_scene.c", line=line@entry=1214, 
    function=function@entry= <__PRETTY_FUNCTION__.5> "wlr_scene_node_coords") at assert.c:94
#5   in __assert_fail (assertion= "node", file= "types/scene/wlr_scene.c", line=1214, function= <__PRETTY_FUNCTION__.5> "wlr_scene_node_coords") at assert.c:103
#6   in wlr_scene_node_coords (node=, lx_ptr=, ly_ptr=) at ../subprojects/wlroots/types/scene/wlr_scene.c:1214
#7   in arrange_popups (popups=) at ../sway/desktop/transaction.c:617
#8   in arrange_root (root=) at ../sway/desktop/transaction.c:687
#9   in transaction_progress () at ../sway/desktop/transaction.c:741
#10  in transaction_commit_pending () at ../sway/desktop/transaction.c:861
#11  in _transaction_commit_dirty (server_request=true) at ../sway/desktop/transaction.c:937
#12  in transaction_commit_dirty () at ../sway/desktop/transaction.c:941
#13  in seat_execute_command (seat=, binding=) at ../sway/commands/bind.c:648
#14  in handle_key_event (keyboard=, event=) at ../sway/input/keyboard.c:489
#15  in handle_keyboard_group_key (listener=, data=) at ../sway/input/keyboard.c:561
#16  in wl_signal_emit_mutable (signal=<optimized out>, data=) at ../wayland-1.23.1/src/wayland-server.c:2314
[...]

PR #8376, opened 2024-10-07
fix: sway crashes if switch to another workspace with surface when IME popup is shown

Decodetalkers:
resolve: https://github.com/swaywm/sway/issues/8375

in pr https://github.com/swaywm/sway/pull/8196, when im_popup_surface is unmapped, author set the popup->relative to NULL, butt popup is still in popup groups, where assert the relative is not NULL, this cause the panic

### DIFF
--- a/sway/input/text_input.c
+++ b/sway/input/text_input.c
@@ -454,6 +454,7 @@ static void handle_im_popup_surface_unmap(struct wl_listener *listener, void *da
 	struct sway_input_popup *popup =
 		wl_container_of(listener, popup, popup_surface_unmap);
 
+	scene_descriptor_destroy(&popup->scene_tree->node, SWAY_SCENE_DESC_POPUP);
 	// relative should already be freed as it should be a child of the just unmapped scene
 	popup->desc.relative = NULL;
 
--- a/sway/scene_descriptor.c
+++ b/sway/scene_descriptor.c
@@ -39,6 +39,9 @@ void *scene_descriptor_try_get(struct wlr_scene_node *node,
 void scene_descriptor_destroy(struct wlr_scene_node *node,
 		enum sway_scene_descriptor_type type) {
 	struct scene_descriptor *desc = scene_node_get_descriptor(node, type);
+	if (!desc) {
+		return;
+	}
 	descriptor_destroy(desc);
 }
 


emersion:
Hm, it seems like the `scene_descriptor_assign()` call is performed in `input_popup_set_focus()` when there is a non-NULL `surface`. It also seems like we're already calling `scene_descriptor_destroy()` when `input_popup_set_focus()` is called with a NULL `surface`.

Ah, but only when `relative != NULL`, which we reset before calling `input_popup_set_focus()`. So makes sense that we need to manually destroy the scene descriptor in that case. Would be nice to refactor the code to make this less awkward…

Can we guard that new `scene_descriptor_destroy()` call with a `popup->desc.relative` check?

sway/scene_descriptor.c
```diff
@@ -38,7 +38,9 @@ void *scene_descriptor_try_get(struct wlr_scene_node *node,
 
 void scene_descriptor_destroy(struct wlr_scene_node *node,
 		enum sway_scene_descriptor_type type) {
-	struct scene_descriptor *desc = scene_node_get_descriptor(node, type);
+	struct scene_descriptor *desc = scene_node_get_descriptor(node, type);	if (!desc) {
+		return;
```
Nefsen402, line 42: The formatting is off here. Please put a newline before the if.

Decodetalkers: ok, done

sway/input/text_input.c
```diff
@@ -454,9 +454,6 @@ static void handle_im_popup_surface_unmap(struct wl_listener *listener, void *da
 	struct sway_input_popup *popup =
 		wl_container_of(listener, popup, popup_surface_unmap);
 
-	// relative should already be freed as it should be a child of the just unmapped scene
-	popup->desc.relative = NULL;
-
```
Nefsen402, line 459: This isn't correct, we cannot just remove this because of the mentioned signal order. Both scene and text_input want to do stuff when the surface unmaps. If scene runs before text_input, text_input will be working with an invalid pointer.

Putting `scene_descriptor_destroy()` here like before was correct. Sorry for the confusion.

Decodetalkers: I see, thank you

Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8384, opened 2024-10-13
input/seatop_default: refactor move/resize button logic

emersion:
Make it so config->floating_mod_inverse only applies when pressing mod, not when clicking on titlebars.

Centralize logic into shared variables.


Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8385, opened 2024-10-14
layer_shell: Arrange exclusive zone clients first

Nefsen402:
This makes layer_shell more stable against the order of clients.


sway/desktop/layer_shell.c
```diff
@@ -78,10 +82,14 @@ void arrange_layers(struct sway_output *output) {
...
-	arrange_surface(output, &full_area, &usable_area, output->layers.shell_top);
-	arrange_surface(output, &full_area, &usable_area, output->layers.shell_overlay);
+	arrange_surface(output, &full_area, &usable_area, output->layers.shell_background, true);
+	arrange_surface(output, &full_area, &usable_area, output->layers.shell_background, false);
+	arrange_surface(output, &full_area, &usable_area, output->layers.shell_bottom, true);
+	arrange_surface(output, &full_area, &usable_area, output->layers.shell_bottom, false);
+	arrange_surface(output, &full_area, &usable_area, output->layers.shell_top, true);
+	arrange_surface(output, &full_area, &usable_area, output->layers.shell_top, false);
+	arrange_surface(output, &full_area, &usable_area, output->layers.shell_overlay, true);
+	arrange_surface(output, &full_area, &usable_area, output->layers.shell_overlay, false);
```
ikalnytskyi, line 92: **nit**: I think it'd make sense to order arrangement from _top_ to _bottom_, similar to what Sway 1.9 was doing[^1]. The rationale here is simple: the layer on the top most likely is more important to be the first one to be placed. This way, for instance, swaynag (which defaults to _top_) is going to be above of swaybar (which defaults to _bottom_).

[^1]: https://github.com/swaywm/sway/blob/v1.9/sway/desktop/layer_shell.c#L217-L225

kennylevinsen: tbh, not sure one is necessarily universally *better* than the other, but a separate PR to preserve sway 1.9's ordering probably wouldn't hurt.

sway/desktop/layer_shell.c
```diff
@@ -68,6 +68,10 @@ static void arrange_surface(struct sway_output *output, const struct wlr_box *fu
 			continue;
 		}
 
+		if ((surface->scene->layer_surface->current.exclusive_zone > 0) ^ exclusive) {
```
kennylevinsen, line 71: Maybe just use `!=`. :)

sway/desktop/layer_shell.c
Nefsen402: I added a commit for this

kennylevinsen (approved):
LGTM!

PR #8390, opened 2024-10-15
commands/destroy_output: Add new command

kennylevinsen:
destroy_output does the opposite of create_output, destroying any output by the specified name.

Neither create_output nor destroy_output are currently documented.


emersion:
We already have "unplug" for this purpose, I believe.

kennylevinsen:
I would *never* have guessed to look for it under that name.

Issue #8377, opened 2024-10-07
Urgent decorations do not change to focused


jpalus:
### Please fill out the following:
- **Sway Version:**
  - `sway version 1.10-rc2`

- **Debug Log:**
  - Let me know if needed, I can provide a log file.

- **Configuration File:**
  - Reproducible with default config

- **Description:**
In one workspace open window that will receive `urgent` mark and in the other workspace trigger it. In my case I open browser in one workspace, terminal in the other and spawn `xdg-open https://swaywm.org` that will trigger `urgent` in browser through default `focus_on_window_activation`. Switch to workspace with window in `urgent` state that will automatically become `focused`. Observe decorations:
In `sway 1.9`: urgent decorations remain for brief moment and switch to focused decorations
In `sway 1.10-rc2`: decorations remain as urgent

While at `urgent` state isn't following a bit misleading:
https://github.com/swaywm/sway/blob/c90cb37b2a0861548461daa9b75d75317e01b679/sway/sway.5.scd?plain=1#L600-L601

It does work for Wayland windows too.

emersion:
Sounds like a duplicate of https://github.com/swaywm/sway/issues/7912?

jpalus:
It's unrelated issue, problem is somewhere else.

Urgent flag might be discarded differently in two cases:
1. When switching between windows in same workspace (decorations are changed correctly and instantly)
2. When switching from another workspace directly to urgent window (broken but is supposed to transition to "focused" borders with a [delay of half a second](https://github.com/swaywm/sway/blob/63005491cfa374c1eb56f5ca6981a8f39961f67e/sway/config.c#L256) so user has a chance of noticing it was urgent)

https://github.com/swaywm/sway/blob/63005491cfa374c1eb56f5ca6981a8f39961f67e/sway/input/seat.c#L1229-L1247

The first case works because `view_set_urgent()` is executed as a part of key binding handling which ends with `transaction_commit_dirty()` which in turn triggers `container_update()` that is responsible for updating borders.

Now that is not the case with `view_set_urgent()` triggered from timer after half second delay. There's nothing left to trigger container update.

I'm not sure if it's correct but following change fixes it for me:
```diff
diff --git a/sway/input/seat.c b/sway/input/seat.c
index 9a00a3e2..0dd26290 100644
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
[...]

PR #8393, opened 2024-10-16
trigger container update after disabling urgent in timer

jpalus:
switching workspace directly to urgent window creates timer which delays reset of urgent state so user is able to notice it. make sure state change is reflected visually as well (border change) by triggering container update

Fixes: #8377

Created PR to get feedback on change proposed in #8377


Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8394, opened 2024-10-16
config/output: Fix default panel orientation and cleanup

kennylevinsen:
An issue was reported where default panel orientation no longer worked, and found it to be caused by the configuration `find_output_config` used as a starting point. This in turn was something sway always would apply during config loading, but it likely never had any effect as outputs tended to not show up during that time.

Instead, remove these initial values and make it so that `queue_output_config` fills out all output states even for unset config options, and without regard for the current output state.

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -692,6 +692,13 @@ static void dump_output_state(struct wlr_output *wlr_output, struct wlr_output_s
 		sway_log(SWAY_DEBUG, "    adaptive_sync: %s",
 			state->adaptive_sync_enabled ? "enabled": "disabled");
 	}
+	if (state->committed & WLR_OUTPUT_STATE_SCALE) {
+		sway_log(SWAY_DEBUG, "    scale:         %f", state->scale);
+	}
+	if (state->committed & WLR_OUTPUT_STATE_SUBPIXEL) {
+		sway_log(SWAY_DEBUG, "    subpixel:      %s",
+			sway_wl_output_subpixel_to_string(state->subpixel));
+	}
 }
 
 static bool search_valid_config(struct search_context *ctx, size_t output_idx);
@@ -906,9 +913,8 @@ static bool apply_resolved_output_configs(struct matched_output_config *configs,
 		backend_state->output = cfg->output->wlr_output;
 		wlr_output_state_init(&backend_state->base);
 
-		sway_log(SWAY_DEBUG, "Preparing config for %s",
-			cfg->output->wlr_output->name);
 		queue_output_config(cfg->config, cfg->output, &backend_state->base);
+		dump_output_state(cfg->output->wlr_output, &backend_state->base);
 	}
 
 	struct wlr_output_swapchain_manager swapchain_mgr;
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -285,7 +285,6 @@ static void set_mode(struct wlr_output *output, struct wlr_output_state *pending
 	mhz = mhz <= 0 ? INT_MAX : mhz;
 
 	if (wl_list_empty(&output->modes) || custom) {
-		sway_log(SWAY_DEBUG, "Assigning custom mode to %s", output->name);
 		wlr_output_state_set_custom_mode(pending, width, height,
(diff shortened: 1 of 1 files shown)


kennylevinsen:
Note: I don't have a panel with a drm panel orientation to test, would be nice with user validation.

diederikdehaas:
I was the one who reported that issue as I had that problem with my Pine64 PineTab2.

After I fixed my ``rot8`` invocation (by adding ``--display DSI-1`` as parameter), things started to work \o/

The changes I made to ``~/.config/sway/config`` were as follows:

```
input "1046:911:Goodix_Capacitive_TouchScreen" {
        map_to_output DSI-1
}
...
exec rot8 --display DSI-1
```

It does look like the PineTab2's mount-matrix DeviceTree param should be improved, but that's unrelated to this PR, but sway did react properly and changed the orientation of the screen/display when I changed the orientation of the device. So you can add my:

Tested-by: Diederik de Haas <didi.debian@cknow.org>

Thanks a lot for this and the extremely quick response to my issue :-D

diederikdehaas:
Possibly worth noting: I applied these commits on top of 1.10-rc3; luckily they applied cleanly.  
If this PR is accepted/merged, I'd very much like it backported to the v1.10 branch (so it can make it into Debian Trixie).

kennylevinsen:
Thanks for testing!

sway/config/output.c
```diff
@@ -527,24 +527,23 @@ static bool finalize_output_config(struct output_config *oc, struct sway_output
...
+			break;
+		case SCALE_FILTER_LINEAR:
+		case SCALE_FILTER_NEAREST:
+			output->scale_filter = scale_filter_new;
+			break;
+	}
+	if (scale_filter_old != output->scale_filter) {
+		sway_log(SWAY_DEBUG, "Set %s scale_filter to %s", oc->name,
+			sway_output_scale_filter_to_string(output->scale_filter));
+		wlr_damage_ring_add_whole(&output->scene_output->damage_ring);
```
Nefsen402, line 532: Bypassing scene and applying damage like this isn't ideal. The internal pending_damage thing also needs to be updated ideally. We aren't communicating this damage with the backend. But the scale filtering is already a hack...

Nefsen402:
"config/output: Remove initial values in find_output_config" has a typo in the message: "intiial"

sway/config/output.c
kennylevinsen: Yeah, it's not great - that part is still carried over from the original code though. It just got moved to this "finalize" helper when modeset got introduced, and now got a slight change to be more stable. :/

kennylevinsen:
> "config/output: Remove initial values in find_output_config" has a typo in the message: "intiial"

Ah, thanks!

sway/config/output.c
```diff
@@ -482,23 +479,20 @@ static void queue_output_config(struct output_config *oc,
 		// same value as the clients'.
 		float adjusted_scale = round(scale * 120) / 120;
 		if (scale != adjusted_scale) {
-			sway_log(SWAY_INFO, "Adjusting output scale from %f to %f",
-				scale, adjusted_scale);
 			scale = adjusted_scale;
 		}
+		wlr_output_state_set_scale(pending, scale);
```
Nefsen402, line 470: nit: wlr_output_state_set_scale is invoked last for both branches. Either move it out or declare the `float scale;` inside the branches.

sway/config/output.c
Nefsen402: I think I originally authored that. We might need to take a page out of labwc and vendor a damage function.

Nefsen402:
This cleanup is well appreciated. It makes sense to lock an output state to a sway output configuration instead of trying to append the state to the output.

Nefsen402 (approved):

sway/config/output.c
kennylevinsen: Ah yeah that section is still rather silly. Cleaned it up quite a bit.

kennylevinsen:
Thanks for the review!

diederikdehaas:
Awesome, thanks :-)  
Is there something I can/need to do to get this backported to the v1.10 branch?

PR #8399, opened 2024-10-20
backport:  config/output: Fix default panel orientation and cleanup

kennylevinsen:
Backports: config/output: Fix default panel orientation and cleanup (#8394)

Fixes an issue where initial panel orientation is not respected.


diederikdehaas:
\o/

PR #8400, opened 2024-10-20
backport:  layer_shell: Arrange exclusive zone clients first

kennylevinsen:
Backport of  layer_shell: Arrange exclusive zone clients first #8385

Restores 1.9 layer shell ordering behavior.


emersion:
Thanks, but I've already cherry-picked these.

PR #8408, opened 2024-10-25
desktop/output: chase wlroots private fields update

vyivel:
This will be replaced with a proper solution later.


emersion (approved):
Thanks!

Discussion about this:

```
17:28:13 <emersion> hm so sway uses wlr_scene_buffer.buffer_width
17:28:19 <emersion> not sure what to do here
17:28:45 <emersion> we can't grab from the buffer because it goes away once uploaded
17:28:59 <emersion> texture may not exist yet, and is private anyways
17:29:50 <emersion> maybe we should have separate mag_filter_mode/min_filter_mode
17:30:23 <emersion> or maybe just leave as a TODO to fixup later and use WLR_PRIVATE
17:35:32 <vyivel> maybe a node size getter
17:35:45 <vyivel> not very useful for trees though
```

PR #8414, opened 2024-10-28
desktop/output: Store output config on request_state 

kennylevinsen:
Fixes an issue where the size of fullscreened views is wrong when using the Wayland or X11 backends.

It makes `request_state` no longer be a hacky side-channel and instead uses the configuration system, but at the cost of no longer being able to apply completely arbitrary output states through this path. This is not currently an issue as the path is only used for parent window resizes.


ArenM:
I ran into a similar issue where swaylock wasn't being resized properly when running on a display using the wayland backend. This seems to fix that.

Nefsen402 (approved):
Thanks!

PR #8416, opened 2024-10-28
desktop/output: Modeset on session activation

kennylevinsen:
We need to modeset when our session is activated as the output state might have deviated from what we previously had active.

Depends on: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4878

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -6,6 +6,7 @@
 #include <wlr/backend.h>
 #include <wlr/backend/headless.h>
 #include <wlr/backend/multi.h>
+#include <wlr/backend/session.h>
 #include <wlr/config.h>
 #include <wlr/render/allocator.h>
 #include <wlr/render/wlr_renderer.h>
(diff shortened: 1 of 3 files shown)


sway/server.c
```diff
@@ -278,6 +279,9 @@ bool server_init(struct sway_server *server) {
 	wlr_scene_set_gamma_control_manager_v1(root->root_scene,
 		server->gamma_control_manager_v1);
 
+	server->session_active.notify = handle_session_active;
+	wl_signal_add(&server->session->events.active, &server->session_active);
+
```
Nefsen402, line 284: I think we could run into a event ordering problem. DRM will watch the same event which will trigger a scanning of connectors. What happens if this runs first?

kennylevinsen: It means that if connector state had changed since we were active last, we'd end up having to modeset multiple times in a row. I considered adding a new signal for ordering, but it felt annoying..

In the end, the new solution without needing to listen for session signals is much nicer.

Nefsen402:
This is no longer necessary after changes to the upstream wlroots MR.

PR #8418, opened 2024-10-30
Make move_to_container insert after the focused child of destination to match i3 behavior

ccdunder:
#6818 

See [this comment](https://github.com/swaywm/sway/issues/6818#issuecomment-2439372230) for an example scenario.

See [this code](https://github.com/i3/i3/blob/4661e74b5ebe8727d1b0f8c29b1697f1f42daf70/src/con.c#L1524) in i3 for reference.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -256,7 +256,26 @@ static void container_move_to_container(struct sway_container *container,
 	if (destination->view) {
 		container_add_sibling(destination, container, 1);
 	} else {
-		container_add_child(destination, container);
+		/* For split containers, we use the currently focused container within it.
+		* This allows setting marks on, e.g., tabbed containers which will move
+		* con to a new tab behind the focused tab. */
+		sway_log(SWAY_DEBUG, "target is a split container, descending to the currently focused child.");
+		struct sway_node *focused = seat_get_active_tiling_child(
+				config->handler_context.seat, &destination->node);
+		if (!focused || focused == &destination->node) {
+			// The container has no children
+			container_add_child(destination, container);
+			return;
+		}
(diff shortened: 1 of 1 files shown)


sway/commands/move.c
```diff
@@ -256,7 +256,25 @@ static void container_move_to_container(struct sway_container *container,
...
+			// The container has no children
+			container_add_child(destination, container);
+			return;
+		}
+		struct sway_container *target = focused->sway_container;
+		int index = container_sibling_index(target) + 1;
+		if (target->pending.parent) {
+			container_insert_child(target->pending.parent, container, index);
+		} else {
+			// TODO: This depends on missing support to mark workspaces.
```
ccdunder, line 275: My swat at next steps here are to file an i3-compat issue for making workspaces markable and link it here.

In the meantime, I'm cooking up a workaround for folks who are interested in this.

Nefsen402:
LGTM, thanks!

Nefsen402:
Ah, please squash your commits into one.

ccdunder:
@Nefsen402, thanks for the LGTM! I squashed my commits. PTAL at your convenience.

NOTE: build check failures are due to project-wide issue (#8471).

sway/commands/move.c
ccdunder: Updated TODO to reference an issue for tracking.

CyberShadow:
Hi, FYI I've opened #8980 which overlaps with this PR. It takes a different approach that also handles workspace targets - this PR modifies `container_move_to_container()`, but the `N_WORKSPACE` case in `cmd_move_container()` calls `container_move_to_workspace()` instead - so this fix can't apply to workspace targets even once workspaces can be marked.

Issue #8422, opened 2024-11-01
When the title gets long, the window title border appears distorted.


jiggol:
### Please fill out the following:
- **Sway Version:**
sway version 1.10

- **Debug Log:**
[logfile](https://gist.github.com/jiggol/ed38da9bfa651392d452a861cb4a682d#file-sway1-log)

- **Configuration File:**
[config](https://gist.github.com/jiggol/ed38da9bfa651392d452a861cb4a682d#file-config)

- **Description:**
The issue was noticed after upgrading sway to 1.10. It occurs when the window width is smaller than the actual length of the title.

![swappy-20241102-041101](https://github.com/user-attachments/assets/4c14e17b-7ed9-43dd-94c6-f9fdf7e1976c)

Nefsen402:
Fixed with https://github.com/swaywm/sway/pull/8427

PR #8427, opened 2024-11-03
container: Properly constrain title bar padding

Nefsen402:
Important for centered titles

Fixes: https://github.com/swaywm/sway/issues/8422


emersion (approved):
Thanks!

PR #8431, opened 2024-11-05
config/output: Update output position in two passes

kennylevinsen:
The modeset logic iterates over all outputs at the end, sets their new position in the layout and takes a copy of its geometry that is later referenced by layout and scene management code.

If one output is auto configured, then a later output that is manually configured can lead to the first output being moved without the stored geometry being updated.

Split this into two passes: The first pass finalizes the output config and makes updates to the layout, while the second pass updates the copy of the geometry and arranges things as a result of it.

I unfortunately don't have a reproduction of the original issue.


emersion (approved):
Thanks!

PR #8436, opened 2024-11-09
swaynag: fix null dereference on scale change

mstoeckl:
To reproduce the null dereference, run `swaynag` under recent sway or some other compositor supporting cursor-shape-v1, and then change the output scale.

Example:
```
swaymsg output '*' scale 1
swaynag -d -m "Test" &
sleep 1
swaymsg output '*' scale 2
```

Instead of crashing, the error `[swaynag/swaynag.c:161] Failed to get default cursor from theme` is also possible if your cursor theme does not have images for 2x scale; clearing the XCURSOR_THEME environment variable may help.

(The example sometimes produces  `Failed to get buffer. Skipping frame.` warnings for me and produces mis-sized buffers, but that is a separate issue.)


emersion (approved):
Thanks!

PR #8437, opened 2024-11-09
Improvements to swaynag rendering

mstoeckl:
Swaynag did not redraw buffers promptly when the output scale was changed or when scrolling the detail view. The root cause was wl_buffer exhaustion -- trying to commit too many updates before the compositor could respond. Fixing this (commit 3) required eliminating a roundtrip in `render_frame` (commit 2) which required adjusting the way the lower border is drawn (commit 1).

Observable improvements:
* With `swaynag --border-bottom-size 15 -m "Test" `, the bottom border no longer overlaps the message
* An intermediate wl_surface::commit() the detail view is toggled was eliminated, reducing "flicker" as swaynag no longer commits an under- (or over-) sized buffer followed by the correctly sized one.
* Given a long (>10kb) message in  `input.txt` , and running  `swaymsg -d -m 'test' -l < input.txt`: scrolling the detail view with the pointer axis should be less jerky, and there should be fewer "Failed to get buffer. Skipping frame." warning messages.

### DIFF
--- a/swaynag/render.c
+++ b/swaynag/render.c
@@ -235,12 +235,10 @@ static uint32_t render_to_cairo(cairo_t *cairo, struct swaynag *swaynag) {
 	}
 
 	int border = swaynag->type->bar_border_thickness;
-	if (max_height > swaynag->height) {
-		max_height += border;
-	}
+	max_height += border;
 	cairo_set_source_u32(cairo, swaynag->type->border_bottom);
 	cairo_rectangle(cairo, 0,
-			swaynag->height - border,
+			max_height - border,
 			swaynag->width,
 			border);
 	cairo_fill(cairo);
--- a/swaynag/render.c
+++ b/swaynag/render.c
@@ -260,38 +260,43 @@ void render_frame(struct swaynag *swaynag) {
 	cairo_paint(cairo);
 	cairo_restore(cairo);
 	uint32_t height = render_to_cairo(cairo, swaynag);
+	if (swaynag->height == 0) {
+		// for first commit, set required size but do not attach a buffer
+		zwlr_layer_surface_v1_set_size(swaynag->layer_surface, 0, height);
+		zwlr_layer_surface_v1_set_exclusive_zone(swaynag->layer_surface,
+			height);
+		wl_surface_commit(swaynag->surface);
+		goto cleanup;
+	}
 	if (height != swaynag->height) {
 		zwlr_layer_surface_v1_set_size(swaynag->layer_surface, 0, height);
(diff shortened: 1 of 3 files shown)


swaynag/render.c
```diff
@@ -260,39 +260,44 @@ void render_frame(struct swaynag *swaynag) {
...
-		if (!swaynag->current_buffer) {
-			sway_log(SWAY_DEBUG, "Failed to get buffer. Skipping frame.");
-			goto cleanup;
-		}
+		goto cleanup;
+	}
+	if (height != swaynag->height) {
+		zwlr_layer_surface_v1_set_size(swaynag->layer_surface, 0, height);
+		zwlr_layer_surface_v1_set_exclusive_zone(swaynag->layer_surface,
+			height);
```
emersion, line 275: In the resize case, shouldn't we wait for the compositor's configure event as well?

mstoeckl: I believe we do not need to and the cost is greater than the benefit.

I assume the layer shell configure event is fundamentally a hint, similar to xdg-shell. On rereading, I find I can't exactly parse the grammar of the protocol text for [the configure event](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:event:configure):
> The size is a hint, in the sense that the client is free to ignore it if it doesn't resize, pick a smaller size (to satisfy aspect ratio or resize in steps of NxM pixels).

At any rate, there is no "dimension_mismatch" error for ignoring the size as there is for ext-session-lock.

Consequently, if a layer shell client wants to use a specific size, it can just send a buffer (and, optionally, a set_size request), and acknowledge and disregard any following configure events. (The risk is that the compositor might clip the surface if it the client's provided size is infeasible.) In practice, sway appears to just reply with a configure event with height matching set_size, so there is no disagreement.

In fact, the documentation for the layer shell `set_size` request says
> If you pass 0 for either value, the compositor will assign it and inform you of the assignment in the configure event.

so it looks like the compositor should _always_ match the suggested height from set_size in its next configure event, unless there is some scale change, output layout change, or other unusual event (in which case the width might change and the compositor send a new configure event.) 

Therefore: waiting for the compositor's configure event introduces an extra roundtrip of delay before a new buffer can be committed, and its benefit (the client not committing a wrong-sized buffer) only occurs in the case where the configure width depends on the set_size height.

(Waiting also brings a theoretical risk of livelock, where the client could never commit a new buffer if the compositor never gives it the height it asked for -- but this should not happen per `set_size` documentation.)

Misc note: swaynag's current rendering code seems to be a weird mix of attempting to adapt to the recommended height value from the configure (see comparisons with swaynag->height in render_message, render_button) and ignoring it (render_detailed).

PR #8444, opened 2024-11-10
swaybar: Emit property changes for SNI watcher

joanbm:
Emit property change signals for the `IsStatusNotifierHostRegistered` and `RegisteredStatusNotifierItems` properties in `StatusNotifierWatcher`, so code relying on the `PropertiesChanged` signal, instead of signals such as `StatusNotifierHostRegistered`, can work properly.

A library that is affected by this is the `libappindicator-gtk3`* library and it can cause tray icons to be missing after starting swaybar due to a race condition, as follows:
* An application using libappindicator-gtk3 starts, e.g. nm-applet.
* Some time later, swaybar starts.
* swaybar creates the `StatusNotifierWatcher`.
* libappindicator-gtk3 observes the new watcher, but it sees that `IsStatusNotifierHostRegistered`=`false`, so it falls back to the Freedesktop System tray protocol.
* swaybar creates the `StatusNotifierHost`. At this point, libappindicator-gtk3 should "un-fallback" back to SNI. However, since swaybar does not emit the `PropertiesChange` signal on `IsStatusNotifierHostRegistered`, libappindicator-gtk3 doesn't get notified, and stays in fallback state forever.
* As a result, nm-applet will not show in the swaybar tray.

This race can be made reliable by inserting a 1-second long sleep [here](https://github.com/swaywm/sway/blob/03483ff3707a358d935e451d39748e58c205ce8a/swaybar/tray/tray.c#L57).

(*) Note that the libappindicator-gtk3 library has been mostly replaced by libayatana-appindicator, which is not affected by this. The affected version is still used by Arch Linux, source code [here](https://bazaar.launchpad.net/~indicator-applet-developers/libappindicator/trunk/files/298).

### DIFF
--- a/swaybar/tray/watcher.c
+++ b/swaybar/tray/watcher.c
@@ -38,6 +38,8 @@ static int handle_lost_service(sd_bus_message *msg,
 				list_del(watcher->items, idx--);
 				sd_bus_emit_signal(watcher->bus, obj_path, watcher->interface,
 						"StatusNotifierItemUnregistered", "s", id);
+				sd_bus_emit_properties_changed(watcher->bus, obj_path, watcher->interface,
+						"RegisteredStatusNotifierItems", NULL);
 				free(id);
 				if (using_standard_protocol(watcher)) {
 					break;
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

Issue #8433, opened 2024-11-07
Command that fail if no output is connected can fail on startup due to modeset delay


kennylevinsen:
A handful of commands have the following construct:

```c
		if (config->reading || !config->active) {
			return cmd_results_new(CMD_DEFER, NULL);
		} else if (!root->outputs->length) {
			return cmd_results_new(CMD_INVALID,
					"Can't run this command while there's no outputs connected.");
		}
```

These commands expect that once the deferred configuration is run, any outputs that we were expecting would have appeared. This is no longer true due to the modeset delay timer.

We can fix this three ways:

1. Remove the output requirement
2. Delay deferred commands further to happen after modeset
3. Force-trigger an immediate modeset once the configuration is done.

emersion:
> Remove the output requirement

Wouldn't the command then attempt to grab a `sway_output` or workspace, and then this fails because it's not been created yet?

emersion:
Would make sense to me to trigger an immediate modeset right after startup or reload.

Nefsen402:
I don't think special casing immediate modesets would fix this problem in the general case. It's possible I have a script that enables an output and then immediately changes a bunch of properties about it. It's not necessarily true that the only place we run into this is during configuration load/reload.

emersion:
Hm right. What about delaying specifically only modesets triggered by `new_output` events?

kennylevinsen:
> Wouldn't the command then attempt to grab a sway_output or workspace, and then this fails because it's not been created yet?

We always have a fallback output, so even though the focused output is very much not the one the user wants, all operations should be possible, no?

The only caveat is that focusing a particular output can't work before it's created. We might also want to have some special casing so that the fallback output doesn't end up being considered a preferred output to restore workspaces to and such.

But, fixing all that would be more work than just restoring the old behavior, even if it was always fragile, by doing a one-off immediate modeset. It does increase the likelihood of needing two modesets at startup though.

> It's possible I have a script that enables an output and then immediately changes a bunch of properties about it. 

Output configuration would just be stored and used when applicable, but trying to focus the output or move content to it immediately wouldn't work.

philiprhoades:
I have this problem too - upgrading to 1.10.1 did not fix it so I am currently starting my 10 workspaces from a shell script . .

emersion:
The fix is in master but not yet released.

philiprhoades:
> The fix is in master but not yet released.

[...]

PR #8446, opened 2024-11-11
config: Force modeset before running deferred configs

kennylevinsen:
Some commands require outputs to be enabled. These commands are deferred to allow outputs to be discovered, but the delayed modeset might only run some time later.

Force a modeset to occur before running deferred commands.

Fixes: https://github.com/swaywm/sway/issues/8433

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -705,6 +705,7 @@ struct output_config *find_output_config(struct sway_output *output);
 void free_output_config(struct output_config *oc);
 
 void request_modeset(void);
+void force_modeset(void);
 
 bool spawn_swaybg(void);
 
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -706,6 +706,7 @@ void free_output_config(struct output_config *oc);
 
 void request_modeset(void);
 void force_modeset(void);
+bool modeset_is_pending(void);
 
 bool spawn_swaybg(void);
 
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -408,6 +408,10 @@ void request_modeset(void) {
 	}
 }
 
+bool modeset_is_pending(void) {
+	return server.delayed_modeset != NULL;
+}
+
 void force_modeset(void) {
 	if (server.delayed_modeset != NULL) {
 		wl_event_source_remove(server.delayed_modeset);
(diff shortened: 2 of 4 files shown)


emersion:
Should we also force a modeset when a user has executed IPC commands? I don't see the value of delaying there, and it's quite surprising that you don't see the new state after the IPC command has finished execution.

kennylevinsen:
> Should we also force a modeset when a user has executed IPC commands? I don't see the value of delaying there, and it's quite surprising that you don't see the new state after the IPC command has finished execution.

Having everything batched is quite useful. An IPC script reacting to new outputs and running various output commands in response should preferably not end up forcing sequential modesets. I don't have a particular use in mind, but I imagine a kanshi-esque script isn't an unthinkable concept.

I have so far reserved immediate modesetting to where we have no choice as the new config must have taken immediate effect or where we must report on failure (output manager).

emersion:
Users can submit multiple commands in one go with IPC (`,` separator). We can keep the modeset grouped when this happens. But TBH I don't believe even that is really important, because the wlr-output-management protocol is much better suited for this use-case.

kennylevinsen:
Yeah, output manager is definitely a better tool in that regard. I also eventually want to figure out a way for a kanshi profile to be part of the first modeset, such as by figuring out a way to do away with this forced modeset for initial config load...

> We can keep the modeset grouped when this happens

I implemented a forced modeset this way at the end of IPC_COMMAND if a modeset is queued at this point.

I'm a little cautious of doing modesets on each output command as amdgpu can sometimes decide to take several wall clock seconds. It's easy to do if the current approach doesn't cut it though.

emersion:
> I also eventually want to figure out a way for a kanshi profile to be part of the first modeset, such as by figuring out a way to do away with this forced modeset for initial config load...

Yeah, that would be nice. Although I don't think getting rid of the "force modesets" is going to be reliable (it would be racy). Probably need a way to delegate output management to a particular Wayland client.

> I'm a little cautious of doing modesets on each output command as amdgpu can sometimes decide to take several wall clock seconds. It's easy to do if the current approach doesn't cut it though.

We were doing one modeset per output command in the previous release, so I'm not too worried.

emersion (approved):
Thanks!

PR #8447, opened 2024-11-11
Repaint fixes

kennylevinsen:
See individual commits.


emersion (approved):
Thanks!

Nefsen402:
I think it's surprising that wlr_scene_output_build_state() can leave `pending` in an unfinished state that requires a finish upon failure. Sounds like something we can improve on the wlroots side.

PR #8466, opened 2024-11-19
man: fix incorrect smart_borders documentation

alex-huff:
Currently, the documentation for smart_borders claims that when smart_borders is set to 'no_gaps', borders will only be enabled when the workspace has more than one visible child and gaps are equal to zero. However, the real behavior of 'no_gaps' is that borders are enabled when:
- The workspace has more than one visible child.

or,
- There are non-zero gaps between any of the workspace's edges and the corresponding edge of the output.

This commit adjusts the wording of the smart_borders documentation to reflect the true behavior of the option.

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -785,10 +785,11 @@ The default colors are:
 	Kills (closes) the currently focused container and all of its children.
 
 *smart_borders* on|no_gaps|off
-	If smart_borders are _on_, borders will only be enabled if the workspace
-	has more than one visible child. If smart_borders is set to _no_gaps_,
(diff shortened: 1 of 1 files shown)


alex-huff:
> 2\. The documentation for `hidge_edge_borders smart` seems wrong, referencing `smart_borders smart` (which does not exist)...

Yes the documentation regarding all the smart features was very confusing (and sometimes wrong) when I was first reading it. I had to resort to looking at the code to see what it did.

alex-huff:
After not looking at this for months and re-reading the changes I made to the documentation, I am not sure that my changes make the docs any less confusing. It might make things more confusing.

Issue #8468, opened 2024-11-20
Builtin input device detection is broken in `sway_libinput_device_is_builtin`


kuruczgy:
### Please fill out the following:
- **Sway Version:** sway version 1.10

- **Debug Log:** https://gist.github.com/kuruczgy/c332fdfe67075a22b4c176e3b520145d

- **Description:** Builtin input device detection is broken in [`sway_libinput_device_is_builtin`](https://github.com/swaywm/sway/blob/1.10/sway%2Finput%2Flibinput.c#L394), since https://github.com/swaywm/sway/pull/6591
```
kuruczgy: kennylevinsen: Oh wait, the logic in sway_libinput_device_is_builtin is that id_path has to pass all 3 tests, which is... impossible? Am I reading this wrong?
kennylevinsen: yeah that's borked
kennylevinsen: After getting id_path, we should be testing for specific matches and return true if any of them fit, false at the end
kennylevinsen: the original code just checked for paltform so it ended in a return strncmp..., but that doesn't work for multiple rules
```

PR #8469, opened 2024-11-20
input/libinput: fix builtin device detection logic

kuruczgy:
Fixes: #8468


kuruczgy:
It doesn't look like that archlinux failure is related :thinking:

emersion (approved):
Good catch, thanks!

emersion:
(Yeah, the CI failure is unrelated and fixed in https://github.com/swaywm/sway/pull/8470.)

PR #8472, opened 2024-11-22
tree/container: remove output_{enter,leave} listeners in destroy

vimproved:
https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/0d6cc471e95c1632b234fc9152659d24fe5982f1 added an assert that all signals are clear when destroying a wlr_scene_buffer, which is currently triggering due to sway not removing the output_enter and output_leave listeners on the container before calling wlr_scene_node_destroy on output_handler. Remove the listeners before wlr_scene_node_destroy is called.


vyivel:
Thanks!

vyivel (approved):

PR #8491, opened 2024-12-08
input/libinput: fix parsing input drag_lock command

baltitenger:
Regression introduced by by bbadf9b8b10d171a6d5196da7716ea50ee7a6062


emersion (approved):
Good catch, thanks!

PR #8497, opened 2024-12-11
Implement an emit command (#4937)

### LINKED ISSUE #4937: [Feature request] Allow forwarding of key events
Shortly put, I'd like to be able to do this:

`bindsym F24 forward [class="discord"]`

With this example, when F24 is pressed, the first window matching the criteria get a keypress event, exactly as if it was focused and F24 was pressed.

This would allow for a very easy way to handle global hotkeys.

Tert0:
Closes #4937

Command Syntax: `emit <press|release|press-release> <criteria> <keys>`
## Parameters
- **action**: specifies whether to press, release the keys or do both
- **criteria**: first window that matches the criteria will get the key input
- **keys**: modifiers and keys separated by +
## Examples
- `emit press-release [app_id=com.obsproject.Studio] ALT+F8`
- `bindsym --no-repeat Ctrl+T emit press [app_id=wev] Ctrl+A`
- `bindsym --no-repeat --release Ctrl+T emit release [app_id=wev] Ctrl+A`
## Test
1. Open `wev -f wl_keyboard` with a terminal
2. Move the `wev` window into another workspace
3. Run
```bash
swaymsg 'bindsym --no-repeat Ctrl+T emit press [app_id=wev] Ctrl+A'
swaymsg 'bindsym --no-repeat --release Ctrl+T emit release [app_id=wev] Ctrl+A'
```
4. Press and hold Ctrl+T
-> The terminal running `wev` should show that Ctrl+A was pressed
5. Release Ctrl+T
-> The terminal running `wev` should show that Ctrl+A was released

## Code Structure
I am not sure if I should move the code (for converting xkb keysym to key codes), that I copied from the `sway/commands/bind.c`, to `common/utils.c` or maybe somewhere else.

### DIFF
--- a/common/util.c
+++ b/common/util.c
@@ -141,3 +141,9 @@ bool sway_set_cloexec(int fd, bool cloexec) {
 	}
 	return true;
 }
+
+uint32_t get_current_time_msec(void) {
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	return now.tv_sec * 1000 + now.tv_nsec / 1000000;
+}
--- a/common/util.c
+++ b/common/util.c
@@ -142,7 +142,7 @@ bool sway_set_cloexec(int fd, bool cloexec) {
 	return true;
 }
 
-uint32_t get_current_time_msec(void) {
+uint32_t sway_get_current_time_msec(void) {
 	struct timespec now;
 	clock_gettime(CLOCK_MONOTONIC, &now);
 	return now.tv_sec * 1000 + now.tv_nsec / 1000000;
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -201,6 +201,7 @@ sway_cmd cmd_workspace;
 sway_cmd cmd_workspace_layout;
 sway_cmd cmd_ws_auto_back_and_forth;
 sway_cmd cmd_xwayland;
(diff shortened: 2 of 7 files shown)


emersion:
Now that we have Wayland protocols for this feature, I'm not sure adding IPC commands is really desirable…

Tert0:
Which wayland protocols are you referring to?

emersion:
virtual-keyboard and virtual-pointer.

kennylevinsen:
> As far as I understand the virtual-keyboard protocol, it does not allow sending keys to unfocused windows.

A Wayland surface must by definition have keyboard focus to receive key events. This is what enter/leave is.

The proposed command gives keyboard focus to the window matching the criteria (which is the bit "missing" from virtual-keyboard) and then issues a single synthetic key event. Unless all key events are sent at once, this would race with the keyboard state of the window the user currently focused and any keyboard inputs they are typing.

Tert0:
> A Wayland surface must by definition have keyboard focus to receive key events. This is what enter/leave is.

Using `swaymg` it additionally gives them visual focus (`wl_surface::enter` + shows them).

>Unless all key events are sent at once, this would race with the keyboard state of the window the user currently focused and any keyboard inputs they are typing.

Is there a way with wlroots to send all key events at once?

PR #8511, opened 2024-12-23
config: allow setting stdout/err for exec

MithicSpirit:
Introduces the `exec_output` command, which sets the file to use for
stdout and stderr in the `exec` and `exec_always` commands.

This change is currently undocumented as I am still polling interest (it was
originally just for personal use, but a friend encouraged me to submit it
upstream). If it seems like it might be merged, I can write up some
documentation.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -127,6 +127,7 @@ sway_cmd cmd_default_floating_border;
 sway_cmd cmd_default_orientation;
 sway_cmd cmd_exec;
 sway_cmd cmd_exec_always;
+sway_cmd cmd_exec_output;
 sway_cmd cmd_exit;
 sway_cmd cmd_floating;
 sway_cmd cmd_floating_maximum_size;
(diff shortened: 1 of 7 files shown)


kennylevinsen:
What does this do that `exec "some_command >OUT 2>ERR"` wouldn't already handle?

MithicSpirit:
> What does this do that `exec "some_command >OUT 2>ERR"` wouldn't already handle?

Nothing, this is just for convenience so you don't have to specify that for every command.

llyyr:
> Nothing, this is just for convenience so you don't have to specify that for every command.

I think you probably _want_ to specify it for each command, otherwise mixing up stdout/stderr output from multiple different commands into a single file doesn't seem very useful.

MithicSpirit:
I mainly don't want Sway's logs (that are going to be piped into systemd-cat whenever I get around to it) to be cluttered with the logs from everything else. I can still manually override the output for specific commands if I want to, but that's very rare.

This does feel very niche, so I don't have high hopes for it being merged.

emersion:
You can achieve the same thing with the following:

```
set $exec exec >>sway-exec.log 2>&1

bindsym $mod+Return $exec $term
```

MithicSpirit:
Yeah, that's also an option. I think my command is more convenient though.

PR #8521, opened 2025-01-04
Expose sandbox properties via IPC JSON, criteria, title format

jacobmcnamee:
Rebase and update of https://github.com/swaywm/sway/pull/7187.

Add sandbox properties to IPC JSON, criteria, and title format:
- `sandbox_engine`
- `sandbox_app_id`
- `sandbox_instance_id`

These properties may be set by sandbox engines (e.g. `flatpak`) via the security context protocol: https://wayland.app/protocols/security-context-v1.

### Example Usage

Prepend sandbox properties to window title when sandbox engine is set:
```
for_window [sandbox_engine="."] title_format "[%sandbox_engine/%sandbox_app_id/%sandbox_instance_id] %title"
```

### Testing

Launch an app via `flatpak`:
```
flatpak run org.gnome.clocks
```

Observe JSON layout tree via `swaymsg -r -t get_tree`:
```
"name": "Clocks",
"pid": 14309,
"app_id": "org.gnome.clocks",
...
"sandbox_engine": "org.flatpak",
"sandbox_app_id": "org.gnome.clocks",
"sandbox_instance_id": "2307446143",
```

Observe pretty layout tree via `swaymsg -p -t get_tree`:
```
#1: root "root"
  #2147483647: output "__i3"
    #2147483646: workspace "__i3_scratch"
  #3: output "WL-1"
    #4: workspace "1"
      #5: con "Clocks" (xdg_shell, pid: 14309, app_id: "org.gnome.clocks", sandbox_engine: "org.flatpak", sandbox_app_id: "org.gnome.clocks", sandbox_instance_id: "2307446143")
```

Observe window title when configured as shown above:
```
[org.flatpak/org.gnome.clocks/2307446143] Clocks
```

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -602,6 +602,18 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 	json_object_object_add(object, "inhibit_idle",
 		json_object_new_boolean(view_inhibit_idle(c->view)));
 
+	const char *sandbox_engine = view_get_sandbox_engine(c->view);
+	json_object_object_add(object, "sandbox_engine",
+			sandbox_engine ? json_object_new_string(sandbox_engine) : NULL);
+
+	const char *sandbox_app_id = view_get_sandbox_app_id(c->view);
+	json_object_object_add(object, "sandbox_app_id",
+			sandbox_app_id ? json_object_new_string(sandbox_app_id) : NULL);
+
+	const char *sandbox_instance_id = view_get_sandbox_instance_id(c->view);
+	json_object_object_add(object, "sandbox_instance_id",
+			sandbox_instance_id ? json_object_new_string(sandbox_instance_id) : NULL);
+
 	json_object *idle_inhibitors = json_object_new_object();
 
 	struct sway_idle_inhibitor_v1 *user_inhibitor =
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -372,12 +372,29 @@ set|plus|minus|toggle <amount>
 *title_format* <format>
 	Sets the format of window titles. The following placeholders may be used:
 
-		%title - The title supplied by the window ++
-		%app_id - The wayland app ID (applicable to wayland windows only) ++
-		%class - The X11 classname (applicable to xwayland windows only) ++
-		%instance - The X11 instance (applicable to xwayland windows only) ++
-		%shell - The protocol the window is using (typically xwayland or
-			xdg_shell)
+	*%title*
+		The title supplied by the window
+
+	*%app_id*
+		The wayland app ID (applicable to wayland windows only)
+
+	*%class*
+		The X11 classname (applicable to xwayland windows only)
+
+	*%instance*
+		The X11 instance (applicable to xwayland windows only)
+
+	*%shell*
+		The protocol the window is using (typically xwayland or xdg_shell)
+
+	*%sandbox_engine*
+		The associated sandbox engine
+
+	*%sandbox_app_id*
+		The app ID provided by the associated sandbox engine
+
+	*%sandbox_instance_id*
+		The instance ID provided by the associated sandbox engine
 
 	This command is typically used with *for_window* criteria. For example:
 
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -1058,6 +1058,22 @@ The following attributes may be matched with:
 	expression. If the value is \_\_focused\_\_, then all the views on the
 	currently focused workspace matches.
 
+*sandbox_engine*
+	Compare against the associated sandbox engine. Can be a regular expression.
+	If the value is \_\_focused\_\_, then the sandbox engine must be the same as
+	that of the currently focused window.
+
+*sandbox_app_id*
+	Compare against the app ID provided by the associated sandbox engine. Can be
+	a regular expression. If the value is \_\_focused\_\_, then the sandbox app
+	ID must be the same as that of the currently focused window.
+
+*sandbox_instance_id*
+	Compare against the instance ID provided by the associated sandbox engine.
+	Can be a regular expression. If the value is \_\_focused\_\_, then the
+	sandbox instance ID must be the same as that of the currently focused
+	window.
+
 # SEE ALSO
 
 *sway*(1) *sway-input*(5) *sway-output*(5) *sway-bar*(5) *sway-ipc*(7)
--- a/include/sway/criteria.h
+++ b/include/sway/criteria.h
@@ -53,6 +53,9 @@ struct criteria {
 	char urgent; // 'l' for latest or 'o' for oldest
 	struct pattern *workspace;
 	pid_t pid;
+	struct pattern *sandbox_engine;
+	struct pattern *sandbox_app_id;
+	struct pattern *sandbox_instance_id;
 };
 
 bool criteria_is_empty(struct criteria *criteria);
--- a/include/sway/tree/view.h
+++ b/include/sway/tree/view.h
@@ -221,6 +221,12 @@ const char *view_get_window_role(struct sway_view *view);
 
 uint32_t view_get_window_type(struct sway_view *view);
 
+const char *view_get_sandbox_engine(struct sway_view *view);
+
+const char *view_get_sandbox_app_id(struct sway_view *view);
+
+const char *view_get_sandbox_instance_id(struct sway_view *view);
+
 const char *view_get_shell(struct sway_view *view);
 
 void view_get_constraints(struct sway_view *view, double *min_width,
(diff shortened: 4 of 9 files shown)


sway/desktop/xdg_shell.c
```diff
@@ -139,6 +141,16 @@ static struct sway_xdg_shell_view *xdg_shell_view_from_view(
 	return (struct sway_xdg_shell_view *)view;
 }
 
+static const struct wlr_security_context_v1_state *security_context_from_view(
+		struct sway_view *view) {
+	const struct wl_client *client =
+		wl_resource_get_client(view->surface->resource);
```
emersion, line 147: It seems like this has nothing xdg-shell specific - do we really need to get through the `view->impl` abstraction, or can we just put this logic in generic view code?

sway/ipc-json.c
```diff
@@ -602,6 +602,18 @@ static void ipc_json_describe_view(struct sway_container *c, json_object *object
 	json_object_object_add(object, "inhibit_idle",
 		json_object_new_boolean(view_inhibit_idle(c->view)));
 
+	const char *sandbox_engine = view_get_sandbox_engine(c->view);
+	json_object_object_add(object, "sandbox_engine",
```
emersion, line 606: Seems like we should be adding these new props to the IPC man page as well.

sway/sway.5.scd
```diff
@@ -1058,6 +1075,22 @@ The following attributes may be matched with:
 	expression. If the value is \_\_focused\_\_, then all the views on the
 	currently focused workspace matches.
 
+*sandbox_engine*
+	Compare against the associated sandbox engine. Can be a regular expression.
+	If the value is \_\_focused\_\_, then the sandbox engine must be the same as
+	that of the currently focused window.
+
+*sandbox_app_id*
```
emersion, line 1083: I'm kind of wondering whether we should expose this, or whether we should always prefer the sandbox-provided app ID over the one provided by the sandboxed client.

(IIRC GNOME allows `org.libreoffice.Writer` if the sandbox app ID is `org.libreoffice`.)

emersion (commented):
On the whole this looks good to me, added a few comments!

(Thanks for rebasing on top of my criteria changes!)

jacobmcnamee:
I was curious how other compositors are using and/or exposing security context metadata and did some research.



### Naming Convention

- I think `sandbox_` is reasonable because restrictions imposed on a client by the compositor are trivially defeated unless the client is sandboxed (via Flatpak, Docker, LXC, etc.)
- I have found little to no conflicting precedent (see research below)



### Metadata Usage

- Per the protocol spec, `app_id` and `instance_id` are "opaque, sandbox-specific identifiers" with any meaning derived from engines.md
- Using `app_id` or `instance_id` without matching a well-known `sandbox_engine` violates the spec
- Consider system container use cases (e.g. Docker, LXC) where a security context applies to more than one application and `app_id` may be the name of the container (e.g. "docker-work-3")
- References:
  - https://wayland.app/protocols/security-context-v1
  - https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/security-context/engines.md



### Research Summary

- Most compositors restrict protocol access for clients with a security context
- Most compositors never use security context metadata (`sandbox_engine`, `app_id`, `instance_id`) directly or expose them externally
- KWin uses `app_id` regardless of `sandbox_engine` for .desktop file lookup (conflicts with spec)
- Mutter (draft PR) exposes `app_id` via existing API function `meta_window_get_sandboxed_app_id()` when `sandbox_engine` is "flatpak" (respects spec / engines.md)
- PipeWire exposes analogous security context properties on objects: `pipewire.sec.engine`, `pipewire.sec.app-id`, `pipewire.sec.instance-id`



<details>
<summary>Notes and References</summary>



### [COSMIC](https://github.com/pop-os/cosmic-comp)

- restricts protocol access for clients with a security context
- references:
  - https://github.com/pop-os/cosmic-comp/pull/156/commits/8dce518ba69762b776141a47c6ef98dd88a9d8cc
  - https://github.com/pop-os/cosmic-comp/pull/208
  - https://github.com/Smithay/smithay/pull/1106



### [Hyprland](https://github.com/hyprwm/Hyprland)

- restricts protocol access for clients with a security context
- references:
  - https://github.com/hyprwm/Hyprland/commit/da86aac0f5c5c3604351961b901f62cdf7186a10



### [Jay](https://github.com/mahkoh/jay)

- restricts protocol access for clients with a security context
- per-protocol permission flags implemented internally, not yet accessible
- references:
  - https://github.com/mahkoh/jay/pull/180



### [KWin](https://invent.kde.org/plasma/kwin)

- requires `sandbox_engine` and `app_id` to be non-empty (conflicts with spec)
- uses `app_id` regardless of `sandbox_engine` to find a .desktop file which may contain `X-KDE-Wayland-Interfaces=` (conflicts with spec)
- references:
  - https://invent.kde.org/plasma/kwin/-/merge_requests/4484
  - https://invent.kde.org/plasma/kwin/-/merge_requests/4630
  - https://phabricator.kde.org/D22589



### [mutter](https://gitlab.gnome.org/GNOME/mutter)

- draft PR:
  - requires `sandbox_engine` to be set
  - exposes `app_id` via existing API function `meta_window_get_sandboxed_app_id()` when `sandbox_engine` is "flatpak" (respects spec / engines.md)

- references:
  - https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/3107
  - https://gnome.pages.gitlab.gnome.org/mutter/meta/method.Window.get_sandboxed_app_id.html



### [niri](https://github.com/YaLTeR/niri)

- restricts protocol access for clients with a security context
- references:
  - https://github.com/YaLTeR/niri/commit/e4643c6dbea64304c9e8b074699227e8654c989d
  - https://github.com/Smithay/smithay/pull/1106



### [PipeWire](https://gitlab.freedesktop.org/pipewire/pipewire)

- implements an analogous security context interface
- exposes security context properties on objects: `pipewire.sec.engine`, `pipewire.sec.app-id`, `pipewire.sec.instance-id`
- references:
  - https://docs.pipewire.org/group__pw__security__context.html
  - https://docs.pipewire.org/page_man_pw-container_1.html



</details>

sway/ipc-json.c
jacobmcnamee: Already added to `sway-ipc.7.scd`

sway/sway.5.scd
jacobmcnamee: The sandbox `app_id` is defined to be engine-specific and may have an entirely different meaning.

For example, consider a system container (Docker, LXC, etc.) where the wayland socket / security context is associated with more than one application:
- `sandbox_engine` = "org.docker"
- `sandbox_app_id` = "my-docker-container-2"
- `app_id` = multiple, different per application

I'm not sure even that "Flatpak application ID" is equivalent to `app_id` (for that specific engine).

See https://github.com/swaywm/sway/pull/8521#issuecomment-2585655616

sway/desktop/xdg_shell.c
jacobmcnamee: Done.

sway/ipc-json.c
emersion: Oh, my bad, I missed it somehow.

sway/sway.5.scd
emersion: That's a good point! Perhaps one could argue that any sandbox engine sending metadata through the security-context protocol is Wayland-aware (because it has to add support for that specific protocol)? Docker/LXC don't support Wayland apps out-of-the-box, so will not provide these app IDs which don't match xdg-shell. However, as you said nothing prevents a sandbox engine from having a completely different scheme for app IDs.

emersion (approved):
Thanks a lot for doing the research!

This LGTM, will wait for a bit before merging just in case somebody else has objections.

sway/sway.5.scd
jacobmcnamee: My main point is that there exist use cases where a single wayland socket / security context must support multiple arbitrary applications, making it _impossible_ to set `sandbox_app_id` (a single value for the security context and all associated views) in a way which is compatible with `app_id` (potentially different per view).

For example, a system-level container (e.g. `docker run 'my-ubuntu-container-3'`) could start a terminal, browser, and media player via a single wayland socket / security context.

Flatpak (and other application-level container tools) are a special case where only one application is expected to be associated with a given security context (but again, I'm not sure that's completely accurate).

emersion: System-level containers are less secure than per-app containers, because granted permissions apply to the whole container. That is, if a sandboxed app obtains a permission, all other apps in the container can abuse it.

But yeah, indeed, we don't want to prevent one to run containers with multiple apps in it.

emersion:
Thanks!

PR #8526, opened 2025-01-07
Properly focus container when it changes parent

### LINKED ISSUE #8292: Sway does no properly update when a window moves into an existing container
## Version

`sway version 1.10-dev-40ca4150b (Jun 10 2024, branch 'master')`

## Configuration

Run `sway -c test.conf` with the following as `test.conf`:

```
workspace_layout tabbed
exec sh test.sh
```

And `test.sh` is:

```
# Open five windows
foot &
foot &
foot &
foot &
foot &

sleep 1

# focus the second one
swaymsg focus left
swaymsg focus left
swaymsg focus left

sleep 1
# split it vertically
swaymsg split v

sleep 1
# focus the rightmost one
swaymsg focus right
swaymsg focus right
swaymsg focus right

sleep 1
# move it left three times
swaymsg move left
sleep 1
swaymsg move left
sleep 1
swaymsg move left

# At this point, you'll note that we should be seeing the moved window inside
# the split container, but we're actually seeing the window on the right of the
# tabbed container. The titlebar on the other hand, shows the fourth window as
# active.
sleep 5

# This properly moves the window that we've been moving all along. It will no
# be on the left of the split container.
swaymsg move left
```

The `sleep`s are there to ensure that all commands execute before proceeding. They also help visualise the issue.

## Description

When I move a window into a split container, the view doesn't properly update, and I see the wrong window, while sway shows the wrong titlebar as active. Moving that window again restores sway to a consistent state.

I recommend running the above reproductions script while reading through the script as actions progress.

## Additional notes

I previously mentioned this in https://github.com/swaywm/sway/issues/8205, but it turns out it is a separate issue.

WhyNotHugo:
When a parented container changes parent, sway will render a sibling non-focused container.

Revert deletion of code which introduced this regression.

Fixes: 5e18ed3cf03eee9e83909fede46dd98dff652647
Fixes: https://github.com/swaywm/sway/issues/8292

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1417,8 +1417,11 @@ struct sway_node *seat_get_active_tiling_child(struct sway_seat *seat,
 	struct sway_seat_node *current;
 	wl_list_for_each(current, &seat->focus_stack, link) {
 		struct sway_node *node = current->node;
-		if (node_get_parent(node) != parent) {
+		if (node == parent || !node_has_ancestor(node, parent)) {
 			continue;
+		}
+		while (node_get_parent(node) != parent) {
+			node = node_get_parent(node);
 		}
 		if (parent->type == N_WORKSPACE) {
 			// Only consider tiling children


WhyNotHugo:
Kind bump? This is a really annoying regression since 1.10.

Nefsen402:
It seems this behavior is intended by the "regression" PR, which claims that this behavior is correct for i3.

WhyNotHugo:
This was not reproducible on i3 last I checked.

Have you reproduced the bug yourself? When moving windows through a tabbed container, sway will at some point render _another_ window, while indicating that the one being moved is still focused.

The commit that broke this says:

> My code archaeology isn't good enough to determine what this is here for, but it isn't correct.

I've found why this code is here, if fixes this particular issue. Maybe i3 doesn't need it because this corner case is handled elsewhere, I have no idea.

WhyNotHugo:
Here's a video reproducing the behaviour:

https://mirror.whynothugo.nl/t/sway-issue-8292.mkv

I start out with a tabbed container with:

- One terminal (1)
- A horizontal split with a single terminal (2)
- One terminal (3)
- One terminal (4)
- One terminal (wf-record)

I focus terminal (4), and move it left two times. It is now inside the horizontal split, sway marks the horizontal split as focused, but is rendered the surface from the `wf-record` window. 

I then type `swaymsg move left`. You can't see me type this, but it gets typed into the window that isn't being rendered (e.g.: it's not visible but still has keyboard focus). You can see it appear again after I move it left two more times. The first time it is moved to the left side of the horizontal split, the second time it is moved out of the split.

kennylevinsen:
> Huh, yeah this is definitely a real issue. I find it strange that it doesn't render anything at all, focusing/unfocusing shouldn't affect rendering to this extent.

When you just move plain left, no focus or render state is changed - only the parent's title updates. The issue is that a container is moved from being a direct child of its parent to being a grandchild.

https://github.com/swaywm/sway/blob/962e1e70a60e9f39d2fdb6fa1810017682fd1f7b/sway/commands/move.c#L125-L134

`container_insert_child` will amongst other things detach the container from the current parent, which causes the parent to pick its *last* child to render instead. It does *not* do anything to make the inserted container visible.

What we need to do in this case is change the container state to match the new parent, instead of being reset to the random child. This would also have to apply to the two reparenting scenarios below this one.

I'm not sure if i3 has any caveats if, say, you're moving unfocused containers - maybe visibility only changes if the container moved was initially visible or focused. The commit removing the workaround suggests we were doing too much refocusing compared to i3 at least. It's better to be functional than to be correct though.

alex-huff:
> What determines which child is chosen to render? The focus seems to remain correct (I can still keyboard interact with the non-visible window).

The child rendered is determined by `focused_inactive_child` in the parent's `sway_container_state`. `focused_inactive_child` is calculated with `seat_get_active_tiling_child` for every container during the transaction.

https://github.com/swaywm/sway/blob/8d3a52aa3076a34b008aed8ba142c955329a8874/sway/input/seat.c#L1420-L1434

So, a parent container's `focused_inactive_child` ends up being whichever *direct* child was most recently focused according to the focus stack.

When a container is focused all of its ancestors are moved directly above it on the stack. This is ensured by `seat_set_workspace_focus`.

https://github.com/swaywm/sway/blob/8d3a52aa3076a34b008aed8ba142c955329a8874/sway/input/seat.c#L1184-L1199

> I start out with a tabbed container with:
> 
> * One terminal (1)
> * A horizontal split with a single terminal (2)
> * One terminal (3)
> * One terminal (4)
> * One terminal (wf-record)

However, in this case you described the second container in the tabbed layout, `H[foot, foot]` (which ends up being the parent of the focused foot) is never bumped up on the focus stack which means the fourth container, `wf-recorder[~]` is the most recently focused direct child of the tabbed layout (the focused foot instance is now a grandchild) and so it becomes the `focused_inactive_child` when the transaction is committed which causes it to be rendered.

Your patch works because it re-focuses the already focused container which makes sure `H[foot, foot]` becomes the highest direct child of the tabbed container on the focus stack. It also has to focus the workspace first because `seat_set_workspace_focus` bails if the container is already focused.

The problem is that you should be able to move unfocused containers using IPC without the moved container erroneously gaining focus. Original reasoning for https://github.com/swaywm/sway/commit/5e18ed3cf03eee9e83909fede46dd98dff652647:
```
16:19:36 <djanz> Hi, I have an issue that appears like a slight bug, or at least a very different behaviour to i3. Suppose I set "no_focus [app_id='blah']". Then opening blah indeed does not give it focus. But if I also have the command "for_window [app_id='blah'] move up", either above or below the no_focus command, blah will get focus when opened. In i3 blah would not get focus. Sway 1.9.
```
However if the container being moved was already focused before the move command is executed, the it might be OK to refocus it after it is moved. If this is done it should also fix #8020.

andremarcais:
@WhyNotHugo, have you made any progress on this since? I've come across the same issue. (See #9217 and #9218.)

WhyNotHugo:
Updating the previous version of this patch to only update focus for seats that had focused the affected view was too much of a complex hack.

Instead, I've updated `seat_get_active_tiling_child` to find the currently active node based on the ancestry of the active view.

I think this is a lot more reliable: instead of patching around places that mutate the tree to then repair focus state, this simply derive it.

This also works for similar exotic cases, like moving a view from one unfocused output to another unfocused output.

I think with this change we can remove a workaround in container_split that exists for a bug similar to this one, but I'll keep things focused for now.

smlx (changes requested):
I've done some manual testing on this PR as it stands in 3aa7faab8da432ec14bc783b7cf201619ae46bb7 and it fixes the bug.

As far as I understand it, the behaviour change is correct too. Nice work.

My only comment is that the commit message is stale: the new approach does not revert the change from the previous commit. Once that is fixed, this will be good to go IMO.

WhyNotHugo:
Reworded commit. Code remains identical.

smlx:
In the new commit message:

> This is due to the recently focus stack having stale data when a focus window moves into an unfocused container, and other similar edge cases.

I'm confused: isn't the point of this change that the focus stack is valid, it's just that `seat_get_active_tiling_child` was previously only considering direct children as focus candidates and not deeper descendants?

WhyNotHugo:
The focus stack is a stack of most recently active views. When rendering, it’s queried to determine which tab was most recently active in a container, and that one is focused.

When an inactive window moves into an inactive tab, that tab isn’t the most recently active one, so another one gets focused.

I could have added and ugly workaround to fix the stack _for this particular scenario_, but there are others that yield the same result, and the prior expectation is essentially a huge pain to enforce everywhere.

So instead is using the most-recently-focused stack to determine which tab to focus, this change simply traverses the tree based on the focused window. This determines the correct tab reliably, rather than rely on a stack that we can’t reasonably keep in sync.

andremarcais:
On Sat Aug 8, 2026 at 11:43 AM CEST, Hugo wrote:
> So instead is using the most-recently-focused stack to determine which tab to focus, this change simply traverses the tree based on the focused window. This determines the correct tab reliably, rather than rely on a stack that we can’t reasonably keep in sync.

I disagree with this change. The focus stack must be kept in sync. This
is observable from IPC and this patch is not compatible with i3 (see
#9217, which is reproducible with this patch).

> I could have added and ugly workaround to fix the stack _for this particular scenario_, but there are others that yield the same result, and the prior expectation is essentially a huge pain to enforce everywhere.

True. I just found a similar issue with the `swap` command (I have yet
to create a ticket for it). The solution should be to encapsulate the
tree and its primitive operations in a way that does sync the focus
stack and the focus and then only use that safe interface from the code
for commands.

WhyNotHugo:
> I disagree with this change.

Please explain what this breaks.

> The focus stack must be kept in sync. This is observable from IPC

How do you query the focus stack from the IPC? I don't see it mentioned anywhere in `sway-ipc`, and as far as I understand, it's purely an internal implementation detail.

> I just found a similar issue with the `swap` command (I have yet
to create a ticket for it). The solution should be to encapsulate the
tree and its primitive operations in a way that does sync the focus
stack and the focus and then only use that safe interface from the code
for commands.

I found several other paths which trigger issues similar to the one reproduce by this issue. Trying to find all the sites which lead to both being out-of-sync sound like a sisyphean task.

What you propose is a major refactor. I'm not against it, but it's wildly beyond the scope of this PR.

andremarcais:
> How do you query the focus stack from the IPC? I don't see it mentioned anywhere in `sway-ipc`, and as far as I understand, it's purely an internal implementation detail.

If we look at the JSON produced by `swaymsg -t get_tree`, e.g.,

    {
      "id": 37,
      "focused": false,
      ...
      "nodes": [
        {
          "id": 36,
          "focused": false,
          ...
        },
        {
          "id": 11,
          "focused": false,
          ...
        },
        {
          "id": 38,
          "focused": true,
          ...
        }
      ],
      "floating_nodes": [],
      "focus": [
        38,
        11,
        36
      ],
      ...
    }

the `focus` array is built from the focus stack. (See
`focus_inactive_children_iterator` in `sway/ipc-json.c`.) Notice that
the first element of `focus` is either `focused` (as in the example) or
has a focused descendant. I3 upholds this expectation whereas Sway fails
to do so under certain circumstances. (See the issues I linked to
previously.)

To my knowledge, this is the only way to get the ID of the focused node
in sway/i3. Being able to traverse the tree this way is more generally
useful. (I have different bindings for focus left/right and focus
next/previous tab which walk the tree in this manner.)

> I found several other paths which trigger issues similar to the one reproduce by this issue. Trying to find all the sites which lead to both being out-of-sync sound like a sisyphean task.
>
> What you propose is a major refactor. I'm not against it, but it's wildly beyond the scope of this PR.

Indeed, I find it quite daunting myself, though I've been thinking about
it for a while now.

As it stands however, this PR disregards #9217 and related inconsistency
problems in IPC. I think your change in f0731de was closer to the "quick
fix" we should be going for. I just opened PR #9260 that looks quite a
bit like it but checks if the node being moved is focused or not.

Also worth noting is that i3 doesn't seem to have a global focus stack
like sway. Its internal representation more closely resembles what is
returned by IPC. Sway is faking it.

WhyNotHugo:
#9217 isn't in scope for this PR, it's fixing a separate issue.

f0731de2d38e36e61b6cbd1cde88a7a5c8c01d90 was problematic: it messed up focus when moving an inactive window, had issues with multi-seat scenarios, and emitted faulty IPC events.

andremarcais:
> #9217 isn't in scope for this PR, it's fixing a separate issue.

Ok. I misunderstood your previous comment.

What I understand now is that this is a change to the rendering logic.
Instead of fixing the `focus_stack` in `container_move_in_direction`,
you changed `seat_get_active_tiling_child` to work even with a broken
stack. Is this correct?

I though you were trying to get around the stack in `move.c` (as
`seat_get_active_tiling_child` is used in `container_move_in_direction`
as well). I'm still trying to wrap my head around why this is correct.

What is bothering me is this: the current version of
`seat_get_active_tiling_child` is pretty obviously equivalent to
`parent.focus[0]` for some IPC JSON `parent`; to me that is the
definition. For your version to be correct, it should return
`parent.focus[0]` when `focus_stack` is valid and a best effort
otherwise. Is that so? What even is a "valid" state for the stack? We
can agree that the state is invalid if `focus_stack[0]` is a descendant
of `focus_stack[i]`, `focus_stack[i]` and `focus_stack[j]` are siblings
but `j < i` (that is the subject of #8292 and #9218). Your
implementation will give a "best effort" more often than that: if
`focus_stack[k]` is a descendant of `focus_stack[i]`, and
`focus_stack[i]` and `focus_stack[j]` are siblings, and `k < j`, it may
return `focus_stack[i]` regardless of `i < j`.

Am I overthinking this? Maybe. FWIW, I haven't been able to find any
sort of regression in this patch.

> f0731de2d38e36e61b6cbd1cde88a7a5c8c01d90 was problematic: it messed up focus when moving an inactive window, had issues with multi-seat scenarios, and emitted faulty IPC events.

Ok. Good to know.

Issue #8527, opened 2025-01-07
Invalid group error in config


J0nnyMak0:
After updating to the latest commits today in Sway, I'm starting to get errors in my config that look like this:

```
[ERROR] [sway/config.c:830] Error on line 59 'bindsym Mod4+Return exec alacritty': Invalid group 
```

etc.

emersion:
Oops, should be fixed by https://github.com/swaywm/sway/pull/8528

PR #8528, opened 2025-01-07
Fix has_prefix() comparisons with 0

emersion:
has_prefix() returns a bool, unlike strncmp() which returns an int.

When writing the has_prefix() patch, I said to myself that I really should be careful to remove the integer comparisons, but still managed to miss these.

Fixes: 0c60d1581f7b ("Use has_prefix() instead of strncmp() throughout")
Closes: https://github.com/swaywm/sway/issues/8527


kennylevinsen (approved):
Ah, shit happens. LGTM!

PR #8529, opened 2025-01-08
config/output: skip format checks if all are supported

### LINKED ISSUE #8496: wlr_output_get_primary_formats returning NULL not handled (causing segfault)

- **Sway Version: 1.10**

- **Debug Log: [sway.log](https://github.com/user-attachments/files/18083880/sway.log)**

- **Configuration File: `xwayland disable`**

- **Stack Trace: [sway-bt.log](https://github.com/user-attachments/files/18084090/sway-bt.log)**

- **Description:**

If the wlroots documentation is correct `wlr_output_get_primary_formats` may return `NULL` ([doc](https://wayland.emersion.fr/wlroots/wlr/types/wlr_output.h.html#func-wlr_output_get_primary_formats)). It looks like this is not handled correctly at https://github.com/swaywm/sway/blob/f293418d9d8a9e71fc52bc75a83d24ec2cc934c9/sway/config/output.c#L776 which passes the return value to [`wlr_drm_format_set_get`](https://github.com/swaywm/sway/blob/f293418d9d8a9e71fc52bc75a83d24ec2cc934c9/sway/config/output.c#L786) without a `NULL` check. wlroots then tries to dereference the pointer and segfaults ([`drm_format_set.c#L32`](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/9649fbe443a96680f47e550ebaf37bba0fe44c9e/render/drm_format_set.c#L32)).

I wasn't able to create a reproducer that I can share but I hope the analysis is enough.

mTvare6:
Fixes #8496

### DIFF
--- a/sway/config/output.c
+++ b/sway/config/output.c
@@ -783,7 +783,8 @@ static bool search_render_format(struct search_context *ctx, size_t output_idx) 
 		if (needed_bits < format_bits) {
 			continue;
 		}
-		if (!wlr_drm_format_set_get(primary_formats, fmts[idx])) {
+		// If primary_formats is NULL, all formats are supported
+		if (primary_formats && !wlr_drm_format_set_get(primary_formats, fmts[idx])) {
 			// This is not a supported format for this output
 			continue;
 		}


emersion (approved):
Thanks!

PR #8530, opened 2025-01-08
config/output: don't hardcode DMA-BUF in search_render_format()

emersion:
We could be running with a backend which doesn't support DMA-BUFs, e.g. inside a parent Wayland compositor without GPU acceleration.


kennylevinsen (approved):
Ah, yeah that makes sense.

PR #8531, opened 2025-01-11
layer_shell: cleanup new_popup listener when destroying node

llyyr:
Found via https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4918


Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8542, opened 2025-01-23
Add support for mouse movement events through Sway IPC

Zai-Kun:
This patch lets you track your cursor's position through Sway ipc. You can subscribe to the 'cursor_movement' event and listen to cursor movements as they happen.
Checkout the demo on this reddit post: https://www.reddit.com/r/swaywm/comments/1i7zu4u/a_sway_patch_that_lets_you_listen_for_mouse/

### DIFF
--- a/include/ipc.h
+++ b/include/ipc.h
@@ -32,6 +32,7 @@ enum ipc_command_type {
 	IPC_EVENT_BINDING = ((1<<31) | 5),
 	IPC_EVENT_SHUTDOWN = ((1<<31) | 6),
 	IPC_EVENT_TICK = ((1<<31) | 7),
+	IPC_EVENT_CURSOR_MOVEMENT = ((1<<31) | 8),
 
 	// sway-specific event types
 	IPC_EVENT_BAR_STATE_UPDATE = ((1<<31) | 20),
--- a/include/sway/ipc-server.h
+++ b/include/sway/ipc-server.h
@@ -22,5 +22,6 @@ void ipc_event_shutdown(const char *reason);
 void ipc_event_binding(struct sway_binding *binding);
 void ipc_event_input(const char *change, struct sway_input_device *device);
 void ipc_event_output(void);
+void ipc_event_cursor_movement(double x, double y);
 
 #endif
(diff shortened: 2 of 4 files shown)


emersion:
Thanks for the patch! However, I'm not sure the Sway IPC is a good way to implement this. The Sway IPC is mostly for low-volume, low-frequency data. Wayland is better suited for relaying stuff such as input events.

Can you tell more about your use-case? Why do you need the cursor position?

poach3r:
> Thanks for the patch! However, I'm not sure the Sway IPC is a good way to implement this. The Sway IPC is mostly for low-volume, low-frequency data. Wayland is better suited for relaying stuff such as input events.
> 
> Can you tell more about your use-case? Why do you need the cursor position?

Having access to cursor position gives users the ability to create mouse position dependent programs such as global menus in a similar vein to Windows 11, Awesome, Openbox, etc. This is my primary use case, however I can imagine it being used for adding other features such as hot corners.

poach3r:
By "global menu" I mean a menu which appears when right clicking on your desktop which gives you various options. [This](https://upload.wikimedia.org/wikipedia/commons/f/f7/2010-04-24-133031_1280x800_scrot.png) is an example from Openbox.

kennylevinsen:
Context menus on right click on the background can already be implemented using wlr_layer_shell to make a fullscreen background surface. That surface will have regular mouse focus behavior and can observe right clicks on it, and in turn give you a surface to actually attach the context menu popups to.

Zai-Kun:
> Wayland is better suited for relaying stuff such as input events.

But, there is no inbuilt way to achieve this in wayland as far as I know. You can get the current mouse position through wlr-layer-shell protocol however, you can't "track" your mouse as it moves around without taking the focus which is not what I want for my use case. I've tried several things but nothing worked so I decided to make this patch in the end.

> Can you tell more about your use-case? Why do you need the cursor position?

I wanted to create a tool that would generate heat maps of your mouse movements and a "desktop mate" like oneko that followed your cursor around (using wlr-layer-shell to display it).

PR #8551, opened 2025-01-27
input/cursor: remove tool_proximity listener in destroy

layercak3:
Fixes crash when a transient seat client destroys its seat since https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/8f56f7ca43257cc05c7c4eb57a0f541e05cf9a79


Nefsen402:
Thanks!

Nefsen402 (approved):

PR #8553, opened 2025-01-27
Add support for BT2020 primaries

emersion:
References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4976

This is very basic and just enables BT2020 when 10-bit render depth is enabled and the output supports BT2020.


kennylevinsen:
https://github.com/swaywm/sway/pull/8617 contains a newer version of the BT2020 support than this branch through the hdr command.

emersion:
Yup, let's just continue this work there.

PR #8558, opened 2025-02-03
output/background: fix config ignoring fallback color

### LINKED ISSUE #8556: Fallback color does not fill the screen background when the background file does not exist.
- **Sway Version:**
  - sway version 1.10.1
  - Debian GNU/Linux trixie/sid

- **Configuration File:**
```
output HDMI-A-1 {
  background ~/Imagens/background.jpg fill #000000
  pos 0 0
}
```

- **Description:**
  - If the specified file cannot be accessed, a fallback color do not is provided to the output. Sway show config file error on top of screen, about the file unable to access:

The man page describes that this is possible, if I understand correctly it would not be necessary to generate an error if the file cannot be accessed and if there is a fallback color in the configuration to fill the background.

> output \<name\> background|bg \<file\> \<mode\> [\<fallback_color\>]
>            Sets the wallpaper for the given output to the specified file, using the given scaling mode (one of "stretch", "fill", "fit", "center", "tile"). If the specified file cannot be accessed  or  if  the image does not fill the entire output, a fallback color may be provided to cover the rest of the output. fallback_color should be specified as #RRGGBB. Alpha is not supported.

sahinf:
currently, the output background command handler prematurely returns with an error if the background file cannot be accessed. It should only error if user did not provide fallback color.

closes #8556

### DIFF
--- a/sway/commands/output/background.c
+++ b/sway/commands/output/background.c
@@ -3,10 +3,8 @@
 #include <string.h>
 #include <strings.h>
 #include <unistd.h>
-#include <errno.h>
 #include "sway/commands.h"
 #include "sway/config.h"
-#include "sway/swaynag.h"
 #include "log.h"
 #include "stringop.h"
 
@@ -42,14 +40,14 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 	}
 
 	struct output_config *output = config->handler_context.output_config;
-
+	char *src = NULL;
 	if (strcasecmp(argv[1], "solid_color") == 0) {
 		if (!validate_color(argv[0])) {
 			return cmd_results_new(CMD_INVALID,
 					"Colors should be of the form #RRGGBB");
 		}
-		output->background = strdup(argv[0]);
-		output->background_option = strdup("solid_color");
+		if (!(output->background = strdup(argv[0]))) goto cleanup;
+		if (!(output->background_option = strdup("solid_color"))) goto cleanup;
 		output->background_fallback = NULL;
 		argc -= 2; argv += 2;
 	} else {
@@ -77,37 +75,25 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 			return cmd_results_new(CMD_INVALID, "Missing background file");
 		}
 
-		char *src = join_args(argv, j);
+		if (!(src = join_args(argv, j))) goto cleanup;
 		if (!expand_path(&src)) {
 			struct cmd_results *cmd_res = cmd_results_new(CMD_INVALID,
 				"Invalid syntax (%s)", src);
 			free(src);
 			return cmd_res;
 		}
-		if (!src) {
-			sway_log(SWAY_ERROR, "Failed to allocate expanded path");
-			return cmd_results_new(CMD_FAILURE, "Unable to allocate resource");
-		}
 
 		if (config->reading && *src != '/') {
 			// src file is inside configuration dir
 
 			char *conf = strdup(config->current_config_path);
-			if (!conf) {
-				sway_log(SWAY_ERROR, "Failed to duplicate string");
-				free(src);
-				return cmd_results_new(CMD_FAILURE,
-						"Unable to allocate resources");
-			}
+			if (!conf) goto cleanup;
 
 			char *conf_path = dirname(conf);
 			char *real_src = malloc(strlen(conf_path) + strlen(src) + 2);
 			if (!real_src) {
-				free(src);
 				free(conf);
-				sway_log(SWAY_ERROR, "Unable to allocate memory");
-				return cmd_results_new(CMD_FAILURE,
-						"Unable to allocate resources");
+				goto cleanup;
 			}
 
 			snprintf(real_src, strlen(conf_path) + strlen(src) + 2, "%s/%s", conf_path, src);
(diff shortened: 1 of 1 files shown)


emersion:
cc @mstoeckl

sway/commands/output/background.c
```diff
@@ -117,23 +117,13 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
...
-			output->background_option = strdup(mode);
-		}
 		argc -= j + 1; argv += j + 1;
-
 		output->background_fallback = NULL;
-		if (argc && *argv[0] == '#') {
+
+		if (can_access) {
+			output->background = src;
+			output->background_option = strdup(mode);
```
mstoeckl, line 125: The background color should always be parsed, whether or not the file can be accessed.

sway/commands/output/background.c
```diff
@@ -142,11 +132,18 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
...
 			argc--; argv++;
 
-			if (!can_access) {
-				output->background = output->background_fallback;
-				output->background_option = strdup("solid_color");
-				output->background_fallback = NULL;
-			}
+			output->background = output->background_fallback;
+			output->background_option = strdup("solid_color");
+			output->background_fallback = NULL;
```
mstoeckl, line 137: Instead of doing `output->background_fallback = strdup(argv[0]);` and then later setting `output->background_fallback = NULL;`, I'd recommend introducing a new variable or reordering things so members of `output` are only written to once.

mstoeckl (changes requested):
I've tested this a bit; while it works when the file does not exist, I find that it fails to parse the color argument when the file _is_ accessible;  `swaymsg output '*' background ~/test.png center '#ff00ff'` gives an "Error: invalid output subcommand: #ff00ff".

Also, there's a preexisting issue in `sway/commands/output/background.c` that, while I don't think you need to fix it entirely, should not be exacerbated: when updating variables like `output->background_option`, the previous value should be `free()`d to avoid leaking it, and if  `strdup()` fails the function should return with an error message (like `return cmd_results_new(CMD_FAILURE, "Unable to allocate resource");`.)

sway/commands/output/background.c
```diff
@@ -117,23 +117,13 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
 		}
 
 		bool can_access = access(src, F_OK) != -1;
-		if (!can_access) {
-			sway_log_errno(SWAY_ERROR, "Unable to access background file '%s'",
```
mstoeckl, line 121: This error message, and maybe also the swaynag warning, would be useful to keep in the case where the file cannot be accessed and there there is a fallback color.

sway/commands/output/background.c
sahinf: Done in latest commit.

sway/commands/output/background.c
sahinf: Done.

sahinf:
To outline what I was going for, I present 4 cases:

|       | GOOD file | BAD file |
|-------|----------|----------|
| GOOD fallback |   same as before, use file     |    log that sway is using fallback, use fallback     |
| BAD fallback |    swaynag about malformed fallback, use file     |    swaynag about malformed fallback, return with error     |

sway/commands/output/background.c
```diff
@@ -117,40 +105,47 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
...
+		}
 
-			if (!can_access) {
-				output->background = output->background_fallback;
-				output->background_option = strdup("solid_color");
-				output->background_fallback = NULL;
+		free(output->background_option);
+		free(output->background_fallback);
+		if (!can_access) {
+			if (!fallback) {
```
mstoeckl, line 125: To avoid use-after-free I'd recommend setting `output->background_option` and `output->background_fallback` to NULL immediately after freeing them -- otherwise they may be used without being replaced if `can_access` is false and `fallback` is NULL.

sway/commands/output/background.c
```diff
@@ -117,40 +105,47 @@ struct cmd_results *output_cmd_background(int argc, char **argv) {
...
+											  "background file '%s' and no valid fallback provided", src);
+				free(src);
+				return res;
 			}
+			sway_log(SWAY_DEBUG, "Cannot access file '%s', using fallback '%s'", src, fallback);
+			output->background = fallback;
+			output->background_option = strdup("solid_color");
+		} else {
+			output->background = src;
+			output->background_option = strdup(mode);
```
mstoeckl, line 138: `output->background_fallback` should be set in this case?

sahinf: But `output->background_fallback` is currently left at `NULL`. I didn't want to alter that behavior.

screenshot_2025-02-03_19-58-10_757165636

https://github.com/swaywm/sway/blob/master/sway/commands/output/background.c

mstoeckl: The current code from git `master` sets `output->background_fallback = strdup(argv[0])` if a fallback color is specified and the background is accessible. It should continue to do so, so that a command like `swaymsg output '*' background ~/sway/assets/Sway_Tree.png center '#000080'` will end up having swaybg fill the background with the fallback color and then draw the (alpha-transparent) image on top of it.

mstoeckl: Err... 7259512ad22c08297acb5db242528b4a732ca7ad doesn't fix this: `output->background_fallback` always gets set to NULL on line 125.

sahinf: oops, fixed. Also freeing / initializing `output->background` now.

mstoeckl:
I've tested this, and think the logic is good now. Re code style, I don't have time today to look things over in detail. See [`CONTRIBUTING.md`](https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md) -- typical issues are indentation and where exactly `*` goes. Fixup commits can be squashed.

sahinf:
> I've tested this, and think the logic is good now. Re code style, I don't have time today to look things over in detail. See [`CONTRIBUTING.md`](https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md) -- typical issues are indentation and where exactly `*` goes. Fixup commits can be squashed.

Thank you for reviewing so quickly! I fixed the `*`. It now adheres to Macros, Error Codes, Construction/Destruction Functions, Names, Line Length, Indentation. 

It does not adhere to Brackets with `if (cond) goto cleanup`. I wanted to follow [sway/config.c](https://github.com/swaywm/sway/blob/d093c2e3583e785cb308bec74c08a61e8808faf6/sway/config.c#L218) because it looks so clean. Should I still break them up?

mstoeckl (approved):
At this point, the changes look good to me.

> Should I still break them up?

I think it's probably fine but am not certain; the project maintainers may or may not disagree.

emersion (approved):
Thank you for the review @mstoeckl!

PR #8567, opened 2025-02-11
Rework fork/exec strategy

kennylevinsen:
cmd_exec_process is used whenever sway is meant to execute a child process on behalf of the user, and had a lot of complexity.

In order to avoid having to wait on the user's process, a double-fork was used, which in turn required us to wait on the outer process. In order to track the child PID for launcher purposes, a pipe was used to transmit the PID back to sway.

This resulted in sway blocking for 5-6 ms per exec on my system, which is quite significant. The error handling was also quite lacking - the read loop did not handle errors at all for example.

Instead, teach sway to handle SIGCHLD and do away with the double-fork. This in turn allows us to get rid of the pipe as we can record the child's PID directly. This reduces the time we block to just 1.5 ms on my system. We'd be able to get down to just 150 µs if we could use posix_spawn(3), but posix_spawn(3) cannot reset NOFILE. clone(2) or vfork(2) would be alternatives, but that presents portability issues.

This change is replicated for swaybar, swaybg and swaynag handling, which had similar albeit less complicated implementations.

### DIFF
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -25,16 +25,6 @@ struct cmd_results *cmd_exec_validate(int argc, char **argv) {
 	return error;
 }
 
-static void export_xdga_token(struct launcher_ctx *ctx) {
-	const char *token = launcher_ctx_get_token_name(ctx);
-	setenv("XDG_ACTIVATION_TOKEN", token, 1);
-}
-
-static void export_startup_id(struct launcher_ctx *ctx) {
-	const char *token = launcher_ctx_get_token_name(ctx);
-	setenv("DESKTOP_STARTUP_ID", token, 1);
-}
-
 struct cmd_results *cmd_exec_process(int argc, char **argv) {
 	struct cmd_results *error = NULL;
 	char *cmd = NULL;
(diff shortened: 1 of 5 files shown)


sway/commands/exec_always.c
```diff
@@ -56,67 +46,42 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
...
-			if (ctx) {
-				export_xdga_token(ctx);
-			}
-			if (ctx && !no_startup_id) {
-				export_startup_id(ctx);
+
+		if (ctx) {
+			const char *token = launcher_ctx_get_token_name(ctx);
+			setenv("XDG_ACTIVATION_TOKEN", token, 1);
+			if (no_startup_id) {
```
emersion, line 64: Should probably be `!no_startup_id`?

emersion:
Looks like a good idea!

Previously this was not possible due to Xwayland (wlroots was doing a `waitpid()`), but that has been addressed in wlroots.

sway/commands/exec_always.c
kennylevinsen: ... and this is why we have reviews. :P

emersion (approved):
Thanks!

PR #8568, opened 2025-02-11
man: clarify criteria (incl. PCRE2 usage)

sandwm:
Replace the XWayland-only class attribute in the examples:
- The first example given should work for Wayland-native windows.
- The example 'Kill all windows with the title "Emacs"' should use title, not class. Also, it's a substring (regex) match.

There are many different implementations of regular expressions with incompatible syntax. For example, GNU grep alone provides three different ones. Clarify the use of PCRE2 by sway criteria.

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -972,7 +972,7 @@ The default colors are:
 A criteria is a string in the form of, for example:
 
 ```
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

Issue #7323, opened 2022-12-10
lacking documentation about containers and other concepts


PerBothner:
I'm playing around with Sway, and am impressed by what it can do. One problem is that the documentation covers the various commands, but it doesn't really explain the concepts. I'm willing to write up something and create a pull request for `sway.5.scd` (or perhaps `sway.1.scd`) but there are big holes in my knowledge. 

I suggest adding a `CONTAINERS` section to  `sway.5.scd`. To start with something like:

Sway organizes its windows as nested containers. A container is a rectangular area of the screen that contains one or more windows or nested sub-containers. Every container has a _layout mode_ which specifies how the sub-containers are displayed in a container: A layout mode is one of: `splith` (the sub-containers are arranged left-to-right); `splitv` (similar but top-to-bottom); `tabbed` (the title bars for all sub-containers are show left-to-right but otherwise only the active sub-container is displayed); `stacked` (similar to `tabbed` but the sub-container titlebars are shown one below each other). You can change the layout mode of a container with the `layout` command.

Each output (screen) has one or more top-level containers: The main tiling area, and zero or more floating (top-level) containers. (A floating container can be split into sub-containers.)

Question: It is unclear if/when a window is also a container. Many places in sway(5) uses the form "[container|window]" which implies a window is not a container. However, other places uses "container" in contexts that suggest a window also a container. Some of these seem to be errors. For example `focus up|right|down|left` is documented to "Moves focus to the next container" but it doesn't - it moves focus to the next _window_.

There seem to be what I call "invisible containers" - containers with a single child (window or sub-containers). These seem to be the reason `$mod+a` (focus parent) will sometimes seem to do nothing - presumably because they selected an invidible container. The `split` command seems to create an invisible container around the current window|container - but only if there isn't already a surrounding invisible container. More use of "indicators" (as in "The color used to indicate where a new view will open") might help.  Perhaps when an invisible container is focused we could change the border color or thickness?

Can an invisible container be nested inside another invisible container? I think the answer should be "no".

Question: sway(5) sometimes talks about "views" - but never defines it. Does it mean window, or container, or either?

emersion:
Yeah, I agree a general overview of Sway's concepts would be quite nice.

> It is unclear if/when a window is also a container

Yeah, a window is a kind of container.

> There seem to be what I call "invisible containers" - containers with a single child

These might be a bug. IIRC we try to eliminate intermediate single-child containers.

[...]

PR #8571, opened 2025-02-16
docs: use "window" instead of "view" throughout.

markstos:
"view" is an internal term, while the commonly understood
user-facing term is "window"

Ref: #7323


emersion (approved):
Thank you!

PR #8573, opened 2025-02-16
docs: define what "window" and "container" mean in Sway's context.

markstos:
Before, the Sway docs did not define what "container" meant.

Issue #7323 was raised to better define this concept in the docs.

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -40,6 +40,13 @@ output eDP-1 resolution 1920x1080
 These commands can be executed in your config file, via *swaymsg*(1), or via
 the bindsym command.
 
+# TERMINOLOGY
+
+Every window can be split horizontally or vertically. The terminology is
+"window" for a container that actually contains an Wayland window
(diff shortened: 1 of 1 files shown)


sway/sway.5.scd
```diff
@@ -40,6 +40,13 @@ output eDP-1 resolution 1920x1080
 These commands can be executed in your config file, via *swaymsg*(1), or via
 the bindsym command.
 
+# TERMINOLOGY
+
+Every window can be split horizontally or vertically. The terminology is
+"window" for a container that actually contains an Wayland window
+(like a terminal or browser) and "container" for containers that consist of
+one or more windows.
```
smlx, line 48: I think the wording could be improved slightly. Maybe something like this?

```suggestion
Every container can be split horizontally or vertically. The terminology is
"window" for a container that actually contains an application view
(like a terminal or browser) and "container" for a container that consists
of one or more windows or other containers.
```

smlx (changes requested):
This is a helpful addition to the docs!

PR #8578, opened 2025-02-17
Fix crashes on exit caused by wlroots listener checks

Ferdi265:
This PR fixes sway crashing on exit due to event listeners not being removed from wlroots objects on exit.

Listeners fixed:

- server: all objects created during `server_init()`, reworked `handle_renderer_lost()` to use an idle callback 
- input/input-manager: `wlr_backend.new_input`, `wlr_virtual_keyboard`, `wlr_virtual_pointer`, `wlr_keyboard_shortcuts_inhibit`, `wlr_transient_seat_manager`
- desktop/idle-inhibit: `wlr_idle_inhibit_manager`
- input/text-input: `wlr_text_input_manager` and `wlr_input_method_manager`
- tree/container:  `wlr_scene_buffer.output_enter` and `wlr_scene_buffer.output_leave`

If I haven't missed any, this fixes removing listeners at exit for all wlroots objects with listeners that sway uses. Other wlroots objects not touched in this PR already have some form of listener remove logic. I haven't individually tested those, but they do not lead to crashes on my system.

### DIFF
--- a/include/sway/input/text_input.h
+++ b/include/sway/input/text_input.h
@@ -25,8 +25,10 @@ struct sway_input_method_relay {
 	struct wlr_input_method_v2 *input_method; // doesn't have to be present
 
 	struct wl_listener text_input_new;
+	struct wl_listener text_input_manager_destroy;
 
 	struct wl_listener input_method_new;
+	struct wl_listener input_method_manager_destroy;
 	struct wl_listener input_method_commit;
 	struct wl_listener input_method_new_popup_surface;
 	struct wl_listener input_method_grab_keyboard;
--- a/sway/input/input-manager.c
+++ b/sway/input/input-manager.c
@@ -493,6 +493,14 @@ struct sway_input_manager *input_manager_create(struct sway_server *server) {
 	return input;
 }
 
+void input_manager_finish(struct sway_input_manager *input) {
+	wl_list_remove(&input->new_input.link);
+	wl_list_remove(&input->virtual_keyboard_new.link);
+	wl_list_remove(&input->virtual_pointer_new.link);
+	wl_list_remove(&input->keyboard_shortcuts_inhibit_new_inhibitor.link);
+	wl_list_remove(&input->transient_seat_create.link);
+}
+
 bool input_manager_has_focus(struct sway_node *node) {
 	struct sway_seat *seat = NULL;
 	wl_list_for_each(seat, &server.input->seats, link) {
--- a/sway/input/text_input.c
+++ b/sway/input/text_input.c
@@ -597,6 +597,34 @@ static void relay_handle_input_method(struct wl_listener *listener,
 	}
 }
 
+static void sway_input_method_relay_finish_text_input(struct sway_input_method_relay *relay) {
+	wl_list_remove(&relay->text_input_new.link);
+	wl_list_remove(&relay->text_input_manager_destroy.link);
+	wl_list_init(&relay->text_input_new.link);
+	wl_list_init(&relay->text_input_manager_destroy.link);
+}
+
+static void relay_handle_text_input_manager_destroy(struct wl_listener *listener, void *data) {
+	struct sway_input_method_relay *relay = wl_container_of(listener, relay,
+		text_input_manager_destroy);
+
+	sway_input_method_relay_finish_text_input(relay);
+}
+
+static void sway_input_method_relay_finish_input_method(struct sway_input_method_relay *relay) {
+	wl_list_remove(&relay->input_method_new.link);
+	wl_list_remove(&relay->input_method_manager_destroy.link);
+	wl_list_init(&relay->input_method_new.link);
+	wl_list_init(&relay->input_method_manager_destroy.link);
+}
+
+static void relay_handle_input_method_manager_destroy(struct wl_listener *listener, void *data) {
+	struct sway_input_method_relay *relay = wl_container_of(listener, relay,
+		input_method_manager_destroy);
+
+	sway_input_method_relay_finish_input_method(relay);
+}
+
 void sway_input_method_relay_init(struct sway_seat *seat,
 		struct sway_input_method_relay *relay) {
 	relay->seat = seat;
@@ -606,16 +634,22 @@ void sway_input_method_relay_init(struct sway_seat *seat,
 	relay->text_input_new.notify = relay_handle_text_input;
 	wl_signal_add(&server.text_input->events.text_input,
 		&relay->text_input_new);
+	relay->text_input_manager_destroy.notify = relay_handle_text_input_manager_destroy;
+	wl_signal_add(&server.text_input->events.destroy,
+		&relay->text_input_manager_destroy);
 
 	relay->input_method_new.notify = relay_handle_input_method;
 	wl_signal_add(
 		&server.input_method->events.input_method,
 		&relay->input_method_new);
+	relay->input_method_manager_destroy.notify = relay_handle_input_method_manager_destroy;
+	wl_signal_add(&server.input_method->events.destroy,
+		&relay->input_method_manager_destroy);
 }
 
 void sway_input_method_relay_finish(struct sway_input_method_relay *relay) {
-	wl_list_remove(&relay->input_method_new.link);
-	wl_list_remove(&relay->text_input_new.link);
+	sway_input_method_relay_finish_text_input(relay);
+	sway_input_method_relay_finish_input_method(relay);
 }
 
 void sway_input_method_relay_set_focus(struct sway_input_method_relay *relay,
--- a/sway/server.c
+++ b/sway/server.c
@@ -460,8 +460,29 @@ bool server_init(struct sway_server *server) {
 }
 
 void server_fini(struct sway_server *server) {
+	// remove listeners
+	wl_list_remove(&server->renderer_lost.link);
+	wl_list_remove(&server->new_output.link);
+	wl_list_remove(&server->layer_shell_surface.link);
+	wl_list_remove(&server->xdg_shell_toplevel.link);
+	wl_list_remove(&server->server_decoration.link);
+	wl_list_remove(&server->xdg_decoration.link);
+	wl_list_remove(&server->pointer_constraint.link);
+	wl_list_remove(&server->output_manager_apply.link);
+	wl_list_remove(&server->output_manager_test.link);
+	wl_list_remove(&server->output_power_manager_set_mode.link);
+#if WLR_HAS_DRM_BACKEND
+	wl_list_remove(&server->drm_lease_request.link);
+#endif
+	wl_list_remove(&server->tearing_control_new_object.link);
+	wl_list_remove(&server->xdg_activation_v1_request_activate.link);
+	wl_list_remove(&server->xdg_activation_v1_new_token.link);
+	wl_list_remove(&server->request_set_cursor_shape.link);
+
 	// TODO: free sway-specific resources
 #if WLR_HAS_XWAYLAND
+	wl_list_remove(&server->xwayland_surface.link);
+	wl_list_remove(&server->xwayland_ready.link);
 	wlr_xwayland_destroy(server->xwayland.wlr_xwayland);
 #endif
 	wl_display_destroy_clients(server->wl_display);
--- a/sway/server.c
+++ b/sway/server.c
@@ -182,11 +182,11 @@ static void detect_proprietary(struct wlr_backend *backend, void *data) {
 	drmFreeVersion(version);
 }
 
-static void handle_renderer_lost(struct wl_listener *listener, void *data) {
-	struct sway_server *server = wl_container_of(listener, server, renderer_lost);
+static void do_renderer_recreate(void *data) {
+	struct sway_server *server = data;
+	server->recreating_renderer = NULL;
 
 	sway_log(SWAY_INFO, "Re-creating renderer after GPU reset");
-
 	struct wlr_renderer *renderer = wlr_renderer_autocreate(server->backend);
 	if (renderer == NULL) {
 		sway_log(SWAY_ERROR, "Unable to create renderer");
@@ -219,6 +219,18 @@ static void handle_renderer_lost(struct wl_listener *listener, void *data) {
 
 	wlr_allocator_destroy(old_allocator);
 	wlr_renderer_destroy(old_renderer);
+}
+
+static void handle_renderer_lost(struct wl_listener *listener, void *data) {
+	struct sway_server *server = wl_container_of(listener, server, renderer_lost);
+
+	if (server->recreating_renderer != NULL) {
+		sway_log(SWAY_DEBUG, "Re-creation of renderer already scheduled");
+		return;
+	}
+
+	sway_log(SWAY_INFO, "Scheduling re-creation of renderer after GPU reset");
+	server->recreating_renderer = wl_event_loop_add_idle(server->wl_event_loop, do_renderer_recreate, server);
 }
 
 bool server_init(struct sway_server *server) {
--- a/sway/server.c
+++ b/sway/server.c
@@ -478,6 +478,7 @@ void server_fini(struct sway_server *server) {
 	wl_list_remove(&server->xdg_activation_v1_request_activate.link);
 	wl_list_remove(&server->xdg_activation_v1_new_token.link);
 	wl_list_remove(&server->request_set_cursor_shape.link);
+	input_manager_finish(server->input);
 
 	// TODO: free sway-specific resources
 #if WLR_HAS_XWAYLAND
--- a/include/sway/desktop/idle_inhibit_v1.h
+++ b/include/sway/desktop/idle_inhibit_v1.h
@@ -13,6 +13,7 @@ enum sway_idle_inhibit_mode {
 struct sway_idle_inhibit_manager_v1 {
 	struct wlr_idle_inhibit_manager_v1 *wlr_manager;
 	struct wl_listener new_idle_inhibitor_v1;
+	struct wl_listener manager_destroy;
 	struct wl_list inhibitors;
 };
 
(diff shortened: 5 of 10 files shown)


Ferdi265:
On second thought, it might be more useful to remove the bulk of the listeners in `server_fini()` before calling destroy, which avoids adding dozens of destroy listeners.

kennylevinsen:
> On second thought, it might be more useful to remove the bulk of the listeners in server_fini() before calling destroy, which avoids adding dozens of destroy listeners.

If we could have a 1:1 mapping between what is set up in `server_init` and what is torn down in `server_fini`, that might be fine - but it is probably still easier to understand if all structs with listeners must have an appropriate destroy handler to remove them.

One thing is what is easiest to implement, another is what is easiest to maintain and catch errors within in the future.

Ferdi265:
> > On second thought, it might be more useful to remove the bulk of the listeners in server_fini() before calling destroy, which avoids adding dozens of destroy listeners.
> 
> If we could have a 1:1 mapping between what is set up in `server_init` and what is torn down in `server_fini`, that might be fine - but it is probably still easier to understand if all structs with listeners must have an appropriate destroy handler to remove them.
> 
> One thing is what is easiest to implement, another is what is easiest to maintain and catch errors within in the future.

IMO components that have `_init()` and `_finish()` functions that create protocol objects that can't go away on their own should probably avoid extra `destroy` handlers and just remove the listeners in finish. Components where objects are created in reference to an outside lifetime (e.g. in response to a seat being created) should probably create `destroy` listeners to remove those.

This is sometimes hard to distinguish though; `text_input.c` for example uses global objects initialized in `server_init()`, but is created in response to seat objects which are destroyed after the wlroots ones. This lifetime mismatch makes this hard to deal with, since it needs to be cleaned up both  in response to a seat going away and in response to the wlroots objects being destroyed, with both happening at exit (which would cause a crash due to double `wl_list_remove()`).

I'm currently handling these by making the `_finish()` function robust against being called twice, but this doesn't feel optimal.

I think I'll try to get this PR far enough to exit sway without crashes with this method, keeping as much as possible in separate commits for reviewability, and we'll sort out what the best path forward is during review.

Ferdi265:
The current state makes sway exit correctly from a desktop containing a few konsole windows and a firefox window.

I'm not completely sure about the correct handling of all of these cases, and I'm sure there are a lot more objects with listeners that need to be removed before destroy, but those don't trigger on a regular exit in my setup. I've already started writing a script to search for all objects in wlroots that have listener checks and their corresponding usages in sway, but that might take a while.

In the meantime, I think it's better to mark this as reviewable now and see where we go afterwards.

Ferdi265:
> If we could have a 1:1 mapping between what is set up in `server_init` and what is torn down in `server_fini`, that might be fine

Addendum to this: this works fine for `server_init`, (and saves a ton of destro handlers) since all those objects are created/destroyed only at init and fini, but this doesn't work for some of the more dynamic objects or listeners created elsewhere in sway.

sway/server.c
```diff
@@ -457,6 +457,28 @@ bool server_init(struct sway_server *server) {
...
+	wl_list_remove(&server->drm_lease_request.link);
+#endif
+	wl_list_remove(&server->tearing_control_new_object.link);
+	wl_list_remove(&server->xdg_activation_v1_request_activate.link);
+	wl_list_remove(&server->xdg_activation_v1_new_token.link);
+	wl_list_remove(&server->request_set_cursor_shape.link);
+#if WLR_HAS_XWAYLAND
+	wl_list_remove(&server->xwayland_surface.link);
+	wl_list_remove(&server->xwayland_ready.link);
+#endif
```
llyyr, line 480: There should be quite a few more danging event listeners. I was doing this when the assert MR was first merged and ended up with this diff and probably still a few left. Feel free to check these and add them to your PR too.

```patch
commit 4757647868e2d14bf82cd5581da139f8cd6559d3
Author: llyyr <llyyr.public@gmail.com>
Date:   Sat Jan 18 00:14:21 2025 +0530

    XXX: remove listeners on exit

diff --git a/sway/server.c b/sway/server.c
index c7fc2a6df586..3dcbee23252f 100644
--- a/sway/server.c
+++ b/sway/server.c
@@ -458,10 +458,33 @@ bool server_init(struct sway_server *server) {
 
 void server_fini(struct sway_server *server) {
 	// TODO: free sway-specific resources
+	wl_list_remove(&server->idle_inhibit_manager_v1.new_idle_inhibitor_v1.link);
+	wl_list_remove(&server->layer_shell_surface.link);
+	wl_list_remove(&server->xdg_shell_toplevel.link);
+	wl_list_remove(&server->server_decoration.link);
+	wl_list_remove(&server->xdg_decoration.link);
+	wl_list_remove(&server->pointer_constraint.link);
+	wl_list_remove(&server->output_manager_apply.link);
+	wl_list_remove(&server->output_manager_test.link);
+	wl_list_remove(&server->output_power_manager_set_mode.link);
+	wl_list_remove(&server->request_set_cursor_shape.link);
+	wl_list_remove(&server->tearing_control_new_object.link);
+	wl_list_remove(&server->xdg_activation_v1_request_activate.link);
+	wl_list_remove(&server->xdg_activation_v1_new_token.link);
+	wl_list_remove(&server->renderer_lost.link);
+	wl_list_remove(&server->drm_lease_request.link);
 #if WLR_HAS_XWAYLAND
+	wl_list_remove(&server->xwayland_surface.link);
+	wl_list_remove(&server->xwayland_ready.link);
 	wlr_xwayland_destroy(server->xwayland.wlr_xwayland);
 #endif
 	wl_display_destroy_clients(server->wl_display);
+	wl_list_remove(&server->new_output.link);
+	wl_list_remove(&server->input->new_input.link);
+	wl_list_remove(&server->input->virtual_keyboard_new.link);
+	wl_list_remove(&server->input->virtual_pointer_new.link);
+	wl_list_remove(&server->input->keyboard_shortcuts_inhibit_new_inhibitor.link);
+	wl_list_remove(&server->input->transient_seat_create.link);
 	wlr_backend_destroy(server->backend);
 	wl_display_destroy(server->wl_display);
 	list_free(server->dirty_nodes);
```

llyyr: Ah, never mind. Just saw the rest of the commits. feel free to mark resolved

Ferdi265: I'll look into it; I might have missed some listeners.

Ferdi265: added `renderer_lost` to listeners removed in `server_fini()` and moved removal of the xwayland listeners into the existing WLR_HAS_XWAYLAND ifdef.

Ferdi265:
I found an interesting crash related to the `renderer_lost` listener when testing this PR today that seems non-trivial to fix:

The error occurs while sway is trying to destroy the old renderer after a GPU reset:

Call Trace: wlroots `begin_gles2_buffer_pass()` -> libwayland `wl_signal_emit_mutable()` -> sway `handle_renderer_lost()` -> wlroots `wlr_renderer_destroy()`

<details>
<summary>Detailed Stack Trace</summary>

## wlroots `begin_gles2_buffer_pass()`
```c
struct wlr_gles2_render_pass *begin_gles2_buffer_pass(...) {
	...
	if (...) {
                ...
		if (...) {
			wlr_log(WLR_ERROR, "GPU reset (%s)", reset_status_str(status));
			wl_signal_emit_mutable(&renderer->wlr_renderer.events.lost, NULL); // <--- HERE
			return NULL;
		}
	}
        ...
}
```
## libwayland `wl_signal_emit_mutable()`
```c
WL_EXPORT void
wl_signal_emit_mutable(...)
{
	struct wl_listener cursor;
	struct wl_listener end;

	/* Add two special markers: one cursor and one end marker. This way, we
	 * know that we've already called listeners on the left of the cursor
	 * and that we don't want to call listeners on the right of the end
	 * marker. The 'it' function can remove any element it wants from the
	 * list without troubles.
	 *
	 * There was a previous attempt that used to steal the whole list of
	 * listeners but then that broke wl_signal_get().
	 *
	 * wl_list_for_each_safe tries to be safe but it fails: it works fine
	 * if the current item is removed, but not if the next one is. */
	wl_list_insert(&signal->listener_list, &cursor.link); // <-- additional listeners outside of sway's control
	cursor.notify = handle_noop;
	wl_list_insert(signal->listener_list.prev, &end.link); // <-- additional listeners outside of sway's control
	end.notify = handle_noop;

	while (cursor.link.next != &end.link) {
		struct wl_list *pos = cursor.link.next;
		struct wl_listener *l = wl_container_of(pos, l, link);

		wl_list_remove(&cursor.link);
		wl_list_insert(pos, &cursor.link);

		l->notify(l, data); // <--- HERE
	}
        ...
}
```

## sway `handle_renderer_lost()`
```c
static void handle_renderer_lost(...) {
	...
	struct wlr_renderer *old_renderer = server->renderer;
        ...
	wlr_renderer_destroy(old_renderer);
}
```

## wlroots `wlr_renderer_destroy()`
```c
void wlr_renderer_destroy(struct wlr_renderer *r) {
	if (!r) {
		return;
	}

	wl_signal_emit_mutable(&r->events.destroy, r);

	assert(wl_list_empty(&r->events.destroy.listener_list));
	assert(wl_list_empty(&r->events.lost.listener_list)); <-- ASSERTION FAILS HERE

	if (r->impl && r->impl->destroy) {
		r->impl->destroy(r);
	} else {
		free(r);
	}
}
```
</details>

Basically, when the old `wlr_renderer` is destroyed in response to the `renderer_lost` event, there will always still be (fake) listeners added by `wl_signal_emit_mutable()` in the call stack below, triggering the assertion, regardless of what sway does. The  `wlr_renderer` would also be free'd, which I'm not sure works out without any UAF in the return path of `wl_signal_emit_mutable()`, I still need to check that.

The only way I see that this can be solved by sway without wlroots changes is that sway somehow schedules an event in the future where the old `wlr_renderer` can be destroyed without having a `wl_signal_emit_mutable()` on its own events in the call stack below.

kennylevinsen:
> I found an interesting crash related to the renderer_lost listener when testing this PR today that seems non-trivial to fix:

We can fix this in sway if we can break the stack without causing issues elsewhere by either:

1. Make the listener just queue an idle callback for the renderer lost handling.
2. Make the listener queue an idle callback just for destroying the old renderer at the end.

The first one is cleaner and is hopefully fine, but I'm not sure if there will be a bunch of things failing in between the renderer being lost and the idle callback running.

Ferdi265:
> > I found an interesting crash related to the renderer_lost listener when testing this PR today that seems non-trivial to fix:
> 
> We can fix this in sway if we can break the stack without causing issues elsewhere by either:
> 
>     1. Make the listener just queue an idle callback for the renderer lost handling.
> 
>     2. Make the listener queue an idle callback just for destroying the old renderer at the end.
> 
> 
> The first one is cleaner and is hopefully fine, but I'm not sure if there will be a bunch of things failing in between the renderer being lost and the idle callback running.

These seem like sensible approaches. I'll look into it and try to see if the first approach works without complications. IMO this could be done in a separate PR, since it doesn't really touch the exit logic and is more than just removing listeners.

Ferdi265:
> 1. Make the listener just queue an idle callback for the renderer lost handling.

I just implemented this in the latest push. I haven't fully tested it yet since I don't know of a way to force a GPU reset or renderer loss. I'm still searching if there's a simple way to do this (for info: I'm using amdgpu on this laptop), otherwise I'll hack something in to make sway think a GPU reset happened to test this.

Ferdi265:
Rebased onto latest master. I've also tested the new delayed renderer recreation by faking the return value of `glGetGraphicsResetStatusKHR()` in wlroots. The new reset handling works and doesn't crash. It's also robust against triggering multiple times before the idle handler gets called.

@kennylevinsen / @llyyr: could you look at the new renderer lost handling and confirm that nothing looks out of the ordinary?

Otherwise I think this PR is good to go. I've been running this for weeks now and haven't had any crashes on exit. There might be a few listeners still somewhere in sway, but for daily use this should be fine, and the rest can be handled when we come across them (IMO anything is better than the current state of master where sway just crashes on exit).

Ferdi265:
Rebased again onto latest master.

I also did a quick grep over the codebase and from a first look it seems like every remaining `wl_signal_add()` has at least one corresponding `wl_signal_remove()`, and at least for the ones I've looked at in more detail those seem to correctly handle exit (mostly via destroy listeners), but I haven't individually tested them. This seems to match with the fact that I haven't encountered any exit crashes these past weeks.

llyyr:
I've also tested this a bit and haven't seen any exit crashes

sway/input/input-manager.c
```diff
@@ -493,6 +493,14 @@ struct sway_input_manager *input_manager_create(struct sway_server *server) {
 	return input;
 }
 
+void input_manager_finish(struct sway_server *server) {
```
emersion, line 496: Nit: this could take a `struct sway_input_manager *` directly.

sway/input/text_input.c
```diff
@@ -606,16 +620,28 @@ void sway_input_method_relay_init(struct sway_seat *seat,
...
 		&relay->input_method_new);
+	relay->input_method_manager_destroy.notify = relay_handle_input_method_manager_destroy;
+	wl_signal_add(&server.input_method->events.destroy,
+		&relay->input_method_manager_destroy);
 }
 
 void sway_input_method_relay_finish(struct sway_input_method_relay *relay) {
+	// return early if finish was already called
+	// can be called due to seat or manager protocol object being destroyed
+	if (!relay->input_method_new.link.prev) return;
```
emersion, line 639: Hm, this sounds like a workaround.

Since `server.text_input` and `server.input_method` are only accessed in `sway_input_method_relay_init()`, we could instead do something like:

```c

static void relay_handle_text_input_manager_destroy(struct wl_listener *listener, void *data) {
	struct sway_input_method_relay *relay = wl_container_of(listener, relay,
		text_input_manager_destroy);
	wl_list_remove(&relay->input_method_new.link);
	wl_list_remove(&relay->input_method_manager_destroy.link);
	wl_list_init(&relay->input_method_new.link);
	wl_list_init(&relay->input_method_manager_destroy.link);
}
```

Perhaps this can be extracted to a helper function `sway_input_method_relay_finish_input_method()` and re-used in `sway_input_method_relay_finish()`.

sway/server.c
```diff
@@ -182,11 +182,11 @@ static void detect_proprietary(struct wlr_backend *backend, void *data) {
 	drmFreeVersion(version);
 }
 
-static void handle_renderer_lost(struct wl_listener *listener, void *data) {
-	struct sway_server *server = wl_container_of(listener, server, renderer_lost);
+static void do_renderer_recreate(void *data) {
+	struct sway_server *server = (struct sway_server *)data;
```
emersion, line 186: Style nit: unnecessary cast.

emersion (commented):
Thanks a lot for fixing this! Overall looks good to me, a few minor comments below.

sway/input/input-manager.c
Ferdi265: makes sense

sway/input/text_input.c
Ferdi265: I wrote this before I knew `wl_list_init()` could be used to make this work.

Thanks, I'll look into this.

sway/server.c
Ferdi265: Thanks, will change.

sway/input/input-manager.c
Ferdi265: done in 86aecdb3.

sway/server.c
Ferdi265: done in fe79c1a3

sway/input/text_input.c
Ferdi265: Implemented this exactly as suggested, using a new helper `sway_input_method_relay_finish_input_method()` that is called by `relay_handle_text_input_manager_destroy()` and `sway_input_method_relay_finish()`.

Done in 777ecd9e.

Ferdi265:
Fixed the nits from @emersion's review and rebased onto latest master. A quick test run seems to suggest everything is still working as before.

sway/input/text_input.c
```diff
@@ -597,6 +597,20 @@ static void relay_handle_input_method(struct wl_listener *listener,
...
+		text_input_manager_destroy);
+
+	sway_input_method_relay_finish_input_method(relay);
+}
+
+static void relay_handle_input_method_manager_destroy(struct wl_listener *listener, void *data) {
+	struct sway_input_method_relay *relay = wl_container_of(listener, relay,
+		input_method_manager_destroy);
+
+	sway_input_method_relay_finish(relay);
```
emersion, line 611: Probably we need to do the same here? ie, the input method manager may get destroyed before the text input manager?

include/sway/input/text_input.h
```diff
@@ -59,6 +61,7 @@ void sway_input_method_relay_init(struct sway_seat *seat,
 	struct sway_input_method_relay *relay);
 
 void sway_input_method_relay_finish(struct sway_input_method_relay *relay);
+void sway_input_method_relay_finish_input_method(struct sway_input_method_relay *relay);
```
emersion, line 64: Could we keep this `static` inside `text_input.c`, since it's not called elsewhere?

emersion:
Sorry, two more comments :P

sway/input/text_input.c
Ferdi265: Shouldn't `relay_handle_text_input_manager_destroy()` remove the listeners for `text_input_new` and `text_input_manager_destroy`, rather than the input method listeners?

emersion: Yes, my bad, my suggestion has a typo!

sway/input/text_input.c
Ferdi265: done in 8e75dad0.

Ferdi265:
> Sorry, two more comments :P

no problem :)

My last push changed the `text_input` handling to remove only the listeners related to `text_input` in `relay_handle_text_input_manager_destroy()` and only the listeners related to `input_method` in `relay_handle_input_method_manager_destroy()`. Both are done via static helper functions (`sway_input_method_relay_finish_text_input()` and `sway_input_method_relay_finish_input_method()`). `sway_input_method_relay_finish()` calls both.

emersion (approved):
Thank you!

Issue #8580, opened 2025-02-18
Choppy/slow window opacity updates in v1.10


fennerm:
### Please read the following before submitting:
- Please do NOT submit bug reports for questions. Ask questions on IRC at #sway on Libera Chat.
- Proprietary graphics drivers, including nvidia, are not supported. Please use the open source equivalents, such as nouveau, if you would like to use Sway.
- Please do NOT submit issues for information from the github wiki. The github wiki is community maintained and therefore may contain outdated information, scripts that don't work or obsolete workarounds.
  If you fix a script or find outdated information, don't hesitate to adjust the wiki page.

### Please fill out the following:
- **Sway Version:**
- swaymsg version 1.10.1

- **Debug Log:**
https://gist.github.com/fennerm/8f365f4d3ec19f28298540cc93cba794

- **Configuration File:**
  I can reproduce with the default config file but with the addition of:
```
default_border none
```

- **Stack Trace:**
NA

- **Description:**
I maintain a program which indicates the focused window by updating the window opacity temporarily - https://github.com/fennerm/flashfocus. In sway 1.10.1 the performance of updating window opacity seems to have degraded a lot. This seems to only happen when default_border=none, or at least its much more evident. 

Expected behavior from v1.9.5:
https://github.com/user-attachments/assets/b6186925-eefa-4003-9d9a-d3f2701ad975

Actual behavior in v1.10.1:
https://github.com/user-attachments/assets/ccfa9f49-468a-4cdd-8e79-e26693700d93

You can reproduce this using this script:
```
#!/usr/bin/env bash

for OP in 0.905 0.82 0.745 0.68 0.625 0.58 0.545 0.52 0.505 0.5 0.505 0.52 0.545 0.58 0.625 0.68 0.745 0.82 0.905 1
do
        swaymsg opacity $OP
        sleep .05
done
```

And updating your window navigation keys as follows:
```
bindsym $mod+h exec swaymsg focus left && <path-to-script>
bindsym $mod+l exec swaymsg focus right && <path-to-script>
bindsym $mod+j exec swaymsg focus down && <path-to-script>
bindsym $mod+k exec swaymsg focus up && <path-to-script>
```

fennerm:
Thanks for the quick fix!

PR #8581, opened 2025-02-18
Fix opacity command not updating contents sometimes

Nefsen402:
```
Calling container_update() wasn't enough: If there is no visible window
decorations (title bar, borders) container_update would basically no-op
and the scene wouldn't repaint with the update alpha. By also calling
output_configure_scene() we force a call to
wlr_scene_buffer_set_opacity() thus ensuring we update the scene.
```

Closes: #8580


kennylevinsen (approved):
LGTM

PR #8586, opened 2025-02-25
sway_text_node: Fix issues with large dimensions

kennylevinsen:
sway_text_node was not respecting max_width during rendering, which could lead to very large cairo surfaces and failures during rendering.

Minimal reproduction: https://gitlab.freedesktop.org/emersion/wleird/-/merge_requests/36

### DIFF
--- a/sway/sway_text_node.c
+++ b/sway/sway_text_node.c
@@ -64,18 +64,6 @@ static int get_text_width(struct sway_text_node *props) {
 	return MAX(width, 0);
 }
 
-static void update_source_box(struct text_buffer *buffer) {
-	struct sway_text_node *props = &buffer->props;
-	struct wlr_fbox source_box = {
-		.x = 0,
-		.y = 0,
-		.width = ceil(get_text_width(props) * buffer->scale),
-		.height = ceil(props->height * buffer->scale),
-	};
-
-	wlr_scene_buffer_set_source_box(buffer->buffer_node, &source_box);
-}
-
 static void render_backing_buffer(struct text_buffer *buffer) {
 	if (!buffer->visible) {
 		return;
@@ -147,7 +135,6 @@ static void render_backing_buffer(struct text_buffer *buffer) {
 
 	wlr_scene_buffer_set_buffer(buffer->buffer_node, &cairo_buffer->base);
 	wlr_buffer_drop(&cairo_buffer->base);
-	update_source_box(buffer);
 
 	pixman_region32_t opaque;
 	pixman_region32_init(&opaque);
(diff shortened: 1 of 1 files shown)


sway/sway_text_node.c
```diff
@@ -87,7 +87,7 @@ static void render_backing_buffer(struct text_buffer *buffer) {
 	}
 
 	float scale = buffer->scale;
-	int width = ceil(buffer->props.width * scale);
+	int width = ceil(get_text_width(&buffer->props) * scale);
```
emersion, line 92: The whole text was rendered to avoid rendering again when resizing a window.

I believe this change breaks `sway_text_node_set_max_width()`.

emersion: Oh, but we call `render_backing_buffer()` in `sway_text_node_set_max_width()`. According to git log it seems like it was there since the beginning.

This begs the question: do we need `update_source_box()` for anything?

sway/sway_text_node.c
```diff
@@ -60,6 +60,8 @@ static int get_text_width(struct sway_text_node *props) {
 	int width = props->width;
 	if (props->max_width >= 0) {
 		width = MIN(width, props->max_width);
+	} else {
+		width = MIN(width, 3840);
```
emersion, line 64: 8K monitors are 7680px large, so this seems a bit low?

kennylevinsen: I mean, it's the boundary for the title text *before* the max_width has been set by the container code.

We should always be setting the max_width from what I can tell, so I don't think it really makes sense for us to render anything before the max width is determined. I think the code has been written as too generalized for our needs...

sway/sway_text_node.c
kennylevinsen: Yeah I also thought that the source box was to optimize resize, but as you mention it's rendered ineffective.

It makes sense to do, but the question is if we want to fix that now or just whack it for now and try again later. From the perspective of fixing the crash I'm leaning towards the latter.

emersion: Yeah, makes sense to fix now and leave optimizations for later.

sway/sway_text_node.c
emersion: Ah yeah. I suppose `max_width` isn't a great name here.

sway/sway_text_node.c
emersion, line 78: I think we need to update the `pixman_region32_union_rect()` call below to match the new size.

kennylevinsen: Yeah good point, done.

emersion (approved):
LGTM! Maybe we should wait for a bit in case @Nefsen402 has something else to add :)

Nefsen402:
I have nothing to add. Nice and simple, should have done this since day 1.

Issue #8488, opened 2024-12-05
Custom Accel Profile


4e576rt8uh9ij9okp:
https://github.com/swaywm/sway/blob/1d783794b508e529bdc665296d690057c93997df/sway/commands/input/accel_profile.c#L7

Hi there,
I'm using Sway but how are the "flat" and "adaptive" mouse acceleration profiles defined?

I would like to create my own custom profile for it and would need some guidance where I can look up the existing acceleration profiles.

Thanks for making Sway, I ❤️ it :D

nazya:
@JayceFayne
In fact, libinput v1.23 introduced custom acceleration profiles for pointer control.
But according `man 5 sway-input` accel_profile can be either flat or adaptive.
Could the compositor support accel_profile custom for better customization?

nazya:
@mstoeckl 
I noticed your contributions in this area—could you comment on whether this could be implemented and if a PR prepared by me adding the support could be mergeable?

mstoeckl:
> could you comment on whether this could be implemented 

Yes.

> and if a PR prepared by me adding the support could be mergeable?

I'm not part of swaywm/sway and therefore cannot commit to anything: but I do think that such an option would fall within the project scope.

emersion:
I think it'd make sense to add such an option. Would probably end up being somewhat similar to `calibration_matrix`.

qiu-x:
For anyone in need of a custom acceleration curve in Sway, I recently made this libinput Lua plugin:
https://github.com/qiu-x/libinput-custom-plugins/blob/main/trackball-accel.lua

Important side-note: if you want to use this, you have to compile libinput with plugin auto-load for now (Sway doesn't load libinput Lua plugins on it's own yet)

PR #8592, opened 2025-03-04
input/mouse: add accel_profile custom

EBADBEEF:
Support libinput LIBINPUT_CONFIG_ACCEL_PROFILE_CUSTOM to define a custom mouse pointer acceleration curve. Implements #8488.

Custom acceleration curve is documented in libinput https://wayland.freedesktop.org/libinput/doc/latest/pointer-acceleration.html#the-custom-acceleration-profile.

Some thoughts:

Custom acceleration is also supported for scrolling, but this commit only implements pointer motion (LIBINPUT_ACCEL_TYPE_MOTION).

We can't detect if values changed because libinput does not provide a way to read back values set by libinput_config_accel_set_points/libinput_device_config_accel_apply. Sway seems to want to know if it changed to trigger a notification (ipc_event_input).

I stuffed the "step" and "points" values into one config option. Maybe it's better to have two like libinput.conf AccelStepMotion/AccelPointsMotion.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -263,6 +263,7 @@ sway_cmd input_cmd_map_to_region;
 sway_cmd input_cmd_middle_emulation;
 sway_cmd input_cmd_natural_scroll;
 sway_cmd input_cmd_pointer_accel;
+sway_cmd input_cmd_pointer_accel_custom;
 sway_cmd input_cmd_rotation_angle;
 sway_cmd input_cmd_scroll_factor;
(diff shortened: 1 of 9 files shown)


mmlb:
I've just come across this PR when looking into acceleration profiles for scrolling, not just pointer. I'll take this for a spin and see about extending so it supports scrolling too.

kassouni:
I too found myself reaching for this setting and came to this PR looking for it! Appreciate the work and would use it if merged.

PR #8597, opened 2025-03-08
sway/server: create ext-data-control manager

llyyr:
I kept the old wlr_data_control manager too because clients aren't done migrating yet due to [Kwin for instance](https://invent.kde.org/plasma/kwin/-/merge_requests/6606) not supporting ext-data-control yet.

Wine for example uses wlr-data-control, while mpv wants ext-data-control.


emersion (approved):
Thanks!

PR #8598, opened 2025-03-08
Use sigaction to establish signal handlers

danieldg:
The `signal` function may or may not re-establish the signal handler after delivering it once.  Switch to using `sigaction` to establish signal handlers so we can specify the desired behavior explicitly.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -159,6 +159,26 @@ void restore_nofile_limit(void) {
 	}
 }
 
+static void setup_signals(void) {
+	// handle SIGTERM signals by exiting
+	struct sigaction term_sig = {
+		.sa_handler = sig_handler,
+		.sa_flags = SA_RESETHAND,
+	};
+	sigaction(SIGTERM, &term_sig, NULL);
+	sigaction(SIGINT, &term_sig, NULL);
+
+	// handle SIGCHLD by reaping children; don't reset it after running it once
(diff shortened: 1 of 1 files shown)


danieldg:
https://man.freebsd.org/cgi/man.cgi?sigaction only refers to `<signal.h>` as a required `#include` and mentioned the constants it claims are undefined in the build.  Is there some other macro or build option interfering here?

andreaskem:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=275328

emersion:
We could add a workaround in the meantime:

```c
// TODO: remove this workaround once this fix is widely deployed:
// https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=275328
#if defined(__FreeBSD__) && __FreeBSD_version < 1500019
#undef _POSIX_C_SOURCE
#define _XOPEN_SOURCE 700
#endif
```

(1500019 was taken from https://docs.freebsd.org/en/books/porters-handbook/versions/)

kennylevinsen:
Hmm, we should definitely get our calls to `signal` switched out with `sigaction`...

For SIGCHLD specifically, I revisited the effort of using SIG_IGN and skipping having our own handler altogether: https://github.com/swaywm/sway/pull/8599

danieldg:
Superseded by #8601.

PR #8599, opened 2025-03-09
Use SIG_IGN for SIGCHLD instead of our own handler

kennylevinsen:
The behavior of handlers registered with signal(3p) is not well-defined  for signals delivered more than once, as laid out in the man page.

We should replace our use of signal with sigaction, but for SIGCHLD specifically we can also just skip the signals altogether by setting the handler to SIG_IGN which causes child reaping to not be required.

Fixes: https://github.com/swaywm/sway/pull/8567

We should still replace our other `signal` calls with `sigaction`, but only SIGCHLD is expected to be seen by sway more than once, and why let us get interrupted by sporadic signals if we can help it? Resetting the handler in 3 places before exec isn't too bad.

### DIFF
--- a/include/sway/server.h
+++ b/include/sway/server.h
@@ -163,6 +163,7 @@ bool server_start(struct sway_server *server);
 void server_run(struct sway_server *server);
 
 void restore_nofile_limit(void);
+void restore_signals(void);
 
 void handle_new_output(struct wl_listener *listener, void *data);
 
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -52,11 +52,8 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 	pid_t child = fork();
 	if (child == 0) {
 		restore_nofile_limit();
+		restore_signals();
 		setsid();
-		sigset_t set;
-		sigemptyset(&set);
-		sigprocmask(SIG_SETMASK, &set, NULL);
-		signal(SIGPIPE, SIG_DFL);
 
 		if (ctx) {
 			const char *token = launcher_ctx_get_token_name(ctx);
--- a/sway/config/bar.c
+++ b/sway/config/bar.c
@@ -213,13 +213,8 @@ static void invoke_swaybar(struct bar_config *bar) {
 		sway_log(SWAY_ERROR, "Failed to create fork for swaybar");
 		return;
 	} else if (pid == 0) {
-		// Remove the SIGUSR1 handler that wlroots adds for xwayland
-		sigset_t set;
-		sigemptyset(&set);
-		sigprocmask(SIG_SETMASK, &set, NULL);
-		signal(SIGPIPE, SIG_DFL);
-
 		restore_nofile_limit();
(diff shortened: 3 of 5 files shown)


emersion:
In general this approach sounds slightly better to me, because the kernel handles everything.

Since we now need to reset multiple signal handlers after forking, could we add a small wrapper to restore them (similar to `restore_nofile_limit()`)?

kennylevinsen:
> POSIX seems to define a separate SA_NOCLDWAIT flag. I'm not sure I'm reading the spec right.

Under [_Exit](https://pubs.opengroup.org/onlinepubs/9799919799/functions/_Exit.html), "Consequences of process termination", and [wait](https://pubs.opengroup.org/onlinepubs/009696899/functions/wait.html), the effect of SA_NOCLDWAIT and SIG_IGN is described as the same.

The unspecified behavior wording is weird, but it does follow up with "except as specified under "Consequences of Process Termination" in the description of the [_Exit()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/_Exit.html) function (see XSH [_Exit()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/_Exit.html#)). [Option End]".

When we switch to sigaction, using SA_NOCLDWAIT is definitely an option, but I *think* SIG_IGN is safe?

> Since we now need to reset multiple signal handlers after forking, could we add a small wrapper to restore them (similar to restore_nofile_limit())?

Will do. It was also my idea to have all three sites use a single common spawn utility function, but that's for later.

emersion:
Hmm, the Linux man page has some hints:

       POSIX.1-1990 disallowed setting the action for SIGCHLD to SIG_IGN.
       POSIX.1-2001 and later allow this possibility, so that ignoring
       SIGCHLD can be used to prevent the creation of zombies (see
       wait(2)).  Nevertheless, the historical BSD and System V behaviors
       for ignoring SIGCHLD differ, so that the only completely portable
       method of ensuring that terminated children do not become zombies
       is to catch the SIGCHLD signal and perform a wait(2) or similar.

Linux, FreeBSD, OpenBSD and NetBSD all prevent the creation of zombies when `SIGCHLD` is set to `SIG_IGN`.

The Linux source code reads:

```c
/*
 * We are exiting and our parent doesn't care.  POSIX.1
 * defines special semantics for setting SIGCHLD to SIG_IGN
 * or setting the SA_NOCLDWAIT flag: we should be reaped
 * automatically and not left for our parent's wait4 call.
 * Rather than having the parent do it as a magic kind of
 * signal handler, we just set this to tell do_exit that we
 * can be cleaned up without becoming a zombie.  Note that
 * we still call __wake_up_parent in this case, because a
 * blocked sys_wait4 might now return -ECHILD.
 *
 * Whether we send SIGCHLD or not for SA_NOCLDWAIT
 * is implementation-defined: we do (if you don't want
 * it, just use SIG_IGN instead).
 */
```

emersion (approved):
tl;dr I think this is correct

kennylevinsen:
Thanks for review!

PR #8601, opened 2025-03-10
Use wl_event_loop_add_signal for exit signals

danieldg:
This avoids calling non-async-signal-safe functions from within a signal handler.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -44,10 +44,6 @@ void sway_terminate(int exit_code) {
 	}
 }
 
-void sig_handler(int signal) {
-	sway_terminate(EXIT_SUCCESS);
-}
-
 void run_as_ipc_client(char *command, char *socket_path) {
 	int socketfd = ipc_open_socket(socket_path);
 	uint32_t len = strlen(command);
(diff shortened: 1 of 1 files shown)


sway/main.c
```diff
@@ -330,10 +326,6 @@ int main(int argc, char **argv) {
 
 	increase_nofile_limit();
 
-	// handle SIGTERM signals
-	signal(SIGTERM, sig_handler);
-	signal(SIGINT, sig_handler);
-
 	// avoid need to reap children
 	signal(SIGCHLD, SIG_IGN);
```
emersion, line 330: Could we keep all signal-related things in the same spot?

include/sway/server.h
```diff
@@ -157,6 +157,7 @@ extern struct sway_debug debug;
 
 extern bool allow_unsupported_gpu;
 
+void sway_terminate(int exit_code);
```
emersion, line 160: We should be able to get rid of the forward decl in `sway/commands/exit.c` now.

sway/main.c
emersion: Sounds good!

include/sway/server.h
danieldg: adding this declaration is no longer part of this series

emersion (approved):
Thanks!

PR #8606, opened 2025-03-12
Avoid crashing on too many containers

kennylevinsen:
If far too many containers are created, they can become so small that their size calculations come out negative, leading to crashes on asserts.

Instead, set a lower bound for sizes and disable the container entirely if it goes below it, giving whatever space it used to the last container.

The splits are not recalculated, so currently the effect is that if all containers have the same width fraction, they keep getting narrower until at some point they all round to zero and the last container will be given all the available space.

A better behavior would have been if the additional container did not contribute to size and fraction calculations at all, but it's an extreme edge-case, anything is better than crashing, and this is easier to implement.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -312,9 +312,9 @@ static void arrange_children(enum sway_container_layout layout, list_t *children
 			wlr_scene_node_set_position(&child->scene_tree->node, 0, title_bar_height);
 			wlr_scene_node_reparent(&child->scene_tree->node, content);
 
-			if (activated) {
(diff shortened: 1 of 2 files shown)


Nefsen402:
> A better behavior would have been if the additional container did not contribute to size and fraction calculations at all, but it's an extreme edge-case, anything is better than crashing, and this is easier to implement.

Eventually I want to get rid of `arrange.c` and just have the scene path calculate sizes itself. At that point, it would be trivial to implement this behavior.

Thanks!

Nefsen402 (approved):

Issue #6294, opened 2021-05-27
Ability to set default size for scratchpad windows


sujay1844:
As you see the title explains everything.
Is there a way to change the default size of scratchpad windows??
If I'm not wrong, the current default resolution is `86x35`. I want to change it to something like `135x35`.
Although I can manually change it once it's in the scratchpad but when I quit the window or reboot the computer, it changes back to `86x35`.
Has this feature been implemented??
Thanks in advance!! By the way, Sway is great!!

eigenbrot:
I agree that a default size would be great, but the same effect can be achieved through a simple modification of the config file. In your example:
```
# Open the scratch pad and resize it
bindsym $mod+Shift+a scratchpad show, resize 135x35
```
This has a nice benefit of ensuing the scratch pad has the same aspect ratio on monitors with different scale factors. A downside is that all scratchpad windows will be affected, but I suppose you could have another key combo to cycle them without resizing.

covert8:
When using this workaround with percentiles it gets applied to another window when i close the scratch pad with the same command

stelcodes:
Yeah I can confirm that the solution from @eigenbrot does not work since it can resize a random window when the scratchpad is closed. It would be nice to get a default size option since this seems like it requires a script which queries the window node tree in order to enforce a scratchpad size.

PR #8616, opened 2025-03-18
command for defining default floating height / width

Linkiablo:
This patch adds a command for defining the default floating height or width as fraction of the respective maximum.

This would be nice for setting the default scratchpad size #6294


emersion:
Note, we're in general not accepting any new window management features anymore. See https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#scope-of-future-changes-to-sway

Issue #8620, opened 2025-03-20
sway panic when try to call ime in session-lock program


Decodetalkers:
### Please fill out the following:
- **Sway Version:**
  - sway version 1.10.1

- **Debug Log:**

[sway.log](https://github.com/user-attachments/files/19365116/sway.log)

- **Configuration File:**
  - Please try to produce with the default configuration.
  - If you cannot reproduce with the default configuration, please try to find the minimal configuration to reproduce.
  - Upload the config to a pastebin such as gist.github.com.

- **Stack Trace:**
```
#0  0x000063bf9435c4cc in wl_signal_add (signal=0x118, listener=0x63bfa72a9440)
    at /usr/include/wayland-server-core.h:476
#1  0x000063bf9435d4d8 in input_popup_set_focus (popup=0x63bfa72a93a0, surface=0x63bfa7388e50)
    at ../sway/input/text_input.c:409
        view = 0x0
        layer_surface = 0x0
        relative_parent = 0x7e60cab4d5c9 <wlr_scene_subsurface_tree_create+35>
        relative = 0x63bfa738d700
#2  0x000063bf9435d9b4 in handle_im_new_popup_surface
    (listener=0x63bfa6f8bee8, data=0x63bfa734d2c0) at ../sway/input/text_input.c:521
        relay = 0x63bfa6f8be70
        popup = 0x63bfa72a93a0
        text_input = 0x63bfa738d700
#3  0x00007e60ca3cb47e in wl_signal_emit_mutable () at /usr/lib/libwayland-server.so.0
#4  0x00007e60cab7d575 in im_get_input_popup_surface
    (client=0x63bfa7135030, resource=0x63bfa7031b90, id=21, surface_resource=0x63bfa701ee20)
    at ../subprojects/wlroots/types/wlr_input_method_v2.c:266
        input_method = 0x63bfa701b8b0
        popup_surface = 0x63bfa734d2c0
        surface = 0x63bfa719d840
        popup_resource = 0x63bfa73644a0
#5  0x00007e60c98fb976 in ??? () at /usr/lib/libffi.so.8
#6  0x00007e60c98f812c in ??? () at /usr/lib/libffi.so.8
#7  0x00007e60c98faf0e in ffi_call () at /usr/lib/libffi.so.8
#8  0x00007e60ca3c9e85 in ??? () at /usr/lib/libwayland-server.so.0
#9  0x00007e60ca3ced22 in ??? () at /usr/lib/libwayland-server.so.0
[...]

PR #8621, opened 2025-03-20
 text_input: Fix ime panic in ext-session-lock 

Decodetalkers:
resolve: https://github.com/swaywm/sway/issues/8620

### DIFF
--- a/sway/input/text_input.c
+++ b/sway/input/text_input.c
@@ -10,6 +10,7 @@
 #include "sway/input/text_input_popup.h"
 #include "sway/layers.h"
 #include "sway/server.h"
+#include <wlr/types/wlr_session_lock_v1.h>
 
 static struct sway_text_input *relay_get_focusable_text_input(
 		struct sway_input_method_relay *relay) {
@@ -385,6 +386,8 @@ static void input_popup_set_focus(struct sway_input_popup *popup,
 
 	struct wlr_layer_surface_v1 *layer_surface =
 		wlr_layer_surface_v1_try_from_wlr_surface(surface);
+	struct wlr_session_lock_surface_v1 *lock_surface =
+		wlr_session_lock_surface_v1_try_from_wlr_surface(surface);
 
 	struct wlr_scene_tree *relative_parent;
 	if (layer_surface) {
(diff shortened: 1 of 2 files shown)


sway/input/text_input.c
```diff
@@ -404,8 +407,29 @@ static void input_popup_set_focus(struct sway_input_popup *popup,
...
+		popup->desc.view = NULL;
+
+		// we don't need to add an event here to NULL out this field because
+		// this field will only be initialized if the popup is part of a layer
+		// surface. Layer surfaces get destroyed as part of the output being
+		// destroyed, thus also trickling down to popups.
+		popup->fixed_output = lock->layer_surface->output;
 	} else {
 		struct sway_view *view = view_from_wlr_surface(surface);
+		// In the future there may be other lock surface been added, so we also need to check here.
```
emersion, line 429: Technically this is not just about lock surfaces, but shells.

sway/input/text_input.c
```diff
@@ -404,8 +407,29 @@ static void input_popup_set_focus(struct sway_input_popup *popup,
...
+		if (view == NULL) {
+			return;
```
emersion, line 431: Can we log a debug message here?

sway/input/text_input.c
```diff
@@ -385,6 +386,8 @@ static void input_popup_set_focus(struct sway_input_popup *popup,
 
 	struct wlr_layer_surface_v1 *layer_surface =
 		wlr_layer_surface_v1_try_from_wlr_surface(surface);
+	struct wlr_session_lock_surface_v1* lock_surface =
```
emersion, line 389: Style: `*` should be on the right side

emersion (commented):
Can you also fix the commit message to _not_ use Conventional Commits? We use a prefix to indicate the area/scope of the change, if any.

Looks good apart from these comments.

It seems like we don't handle correctly popups, or sub-surfaces inside layer-shell or ext-session-lock, or nested sub-surfaces, but that's a separate issue.

Decodetalkers:
> Can you also fix the commit message to _not_ use Conventional Commits? We use a prefix to indicate the area/scope of the change, if any.
> 
> Looks good apart from these comments.
> 
> It seems like we don't handle correctly popups, or sub-surfaces inside layer-shell or ext-session-lock, or nested sub-surfaces, but that's a separate issue.

done, will this kind of commit be right?

sway/input/text_input.c
Decodetalkers: sure

emersion: It seems like there are typos in the log message (double space, and "foget").

Suggested log message: "Unsupported IME focus surface"

(It can also happen with any example I gave earlier, e.g. nested sub-surface, so it's not purely about shells.)

emersion:
Yup, sounds good!

emersion (approved):
Thank you!

Issue #8625, opened 2025-03-24
Window borders for tabbed layout decrease in height with every successive tab


J0nnyMak0:
- **Sway Version:**
  - `sway version 1.10-dev-ab455bba (Mar 23 2025, branch 'master')`

- **Description:**
  - open any app.
  - place it in a tabbed layout (mod+w)
  - now open another app
  - the second app has window borders that are shorter than the height of the window
  - open a third app; and this one has even shorter length border and so on...

lbatalha:
This is likely to be related to https://github.com/swaywm/sway/commit/c2d6aff64c1e265c8f1d95b780b54193defae18a and `height -= title_bar_height;`, since `height` is used later on in a different context where subtracting the `title_bar_height` is incorrect

PR #8630, opened 2025-03-26
Fix regression that added visible waterline

claui:
Commit c2d6aff added a bounds check on `height - title_bar_height`, repurposing the local variable `height` in an attempt to DRY out the expression.

In https://github.com/swaywm/sway/issues/8625#issuecomment-2748546723, @lbatalha noted that the variable gets re-assigned inside the loop body, so its result would leak across loop iterations. This effect compounds and causes the artifact reported in issue #8625, where each child except the first in a tabbed container would acquire a visible waterline.

This PR introduced a second variable and reset it in each loop iteration to get rid of the waterline.

### Tabbed containers

Open Sway and create three tabbed children in the root container.  
Check for visible waterlines in child windows (except the first child):

image

### Stacking containers

Notably, I couldn’t reproduce the effect in stacking containers, even though those should have the regression, too, but I don’t fully understand why it doesn’t occur there. I have added the same fix there for consistency (and just in case).

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -312,9 +312,9 @@ static void arrange_children(enum sway_container_layout layout, list_t *children
 			wlr_scene_node_set_position(&child->scene_tree->node, 0, title_bar_height);
 			wlr_scene_node_reparent(&child->scene_tree->node, content);
 
-			height -= title_bar_height;
-			if (activated && width > 0 && height > 0) {
-				arrange_container(child, width, height, title_bar_height == 0, 0);
+			int net_height = height - title_bar_height;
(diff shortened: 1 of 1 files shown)


kennylevinsen (approved):
Heh, nice catch.

Minor git commit nit, could the summary line specify what is being fixed? E.g., "Fix  height calculation for tabbed/stacked containers". This makes it easier ot see what is going on when looking at the git history, especially in one-line mode when making release notes than having to pull up pull requests by number.

claui:
@kennylevinsen Sure thing! Fixed the summary line.

Credits for first discovery and figuring out the root cause go to @lbatalha.

PR #8640, opened 2025-04-01
sway/commands: Handle incorrect resize unit

### LINKED ISSUE #8619: Sway crashes if 'pt' used for resize set width/height units, instead of 'px'
- **Sway Version:**
sway version 1.9

- **Debug Log:**
https://gist.github.com/scott092707/1a9d3220123569157d85762c0698180d

- **Configuration File:**
- (minimally altered config)
- [See above gist]

- **Stack Trace:**
 [See above gist]

- **Description:**
  - The steps you took in plain English to reproduce the problem.
  - 
The file manager I use [Caja] exhibits odd behaviour, and the devs wanted me to see if it did so with windows that were floating, as well as tiled.
I wished to have it in exactly the same position as when tiled, so I tried to position/size it with percentages. This failed, for some reason.
I switched to using pixels, with the exact position/size as the tiled version (right half of a tiled 4K monitor, the other half being my text editor [featherpad]).
I goofed up, and wrote 'pt' instead of 'px' for the positioning/sizing units.
This causes Sway to crash (!).  [Dumped back at the login-manager [SDDM]]
After I booted with a live-usb and edited the for_window to remove all floating/position/size aspects, and rebooted, all was well.

I copied my usual config out of the way, and altered the standard 1.9 config minimally to use the bum for_window, and launch caja as usual,
I logged out/in, and Sway still crashed.

Odd, that such a minor syntax error could crash Sway...!

-------------------------------------------------------- snippets from config and sway log ------------------------------------------------------
for_window [app_id="[Cc]aja"] { 
	floating enable
	move position 1920 pt 65 pt
	resize set width 1920 pt
	resize set height 2095 pt
}

00:00:01.170 [DEBUG] [sway/tree/view.c:518] for_window '[app_id="[Cc]aja"]' matches view 0x55633253cce0, cmd: 'move position 1920 pt 65 pt'
00:00:01.170 [INFO] [sway/commands.c:261] Handling command 'move position 1920 pt 65 pt'
00:00:01.170 [DEBUG] [sway/tree/view.c:513] Checking criteria [app_id="[Cc]aja"]
00:00:01.170 [DEBUG] [sway/tree/view.c:518] for_window '[app_id="[Cc]aja"]' matches view 0x55633253cce0, cmd: 'resize set width 1920 pt'
00:00:01.170 [INFO] [sway/commands.c:261] Handling command 'resize set width 1920 pt'
00:00:01.171 [ERROR] [sway/commands/resize.c:405] resize_set_floating:invalid height unit

Odd, that when handling "width" it would ERROR using the word "height"

IMMEDIATELY after this, Sway crashes:

00:00:02.261 [ERROR] [common/ipc-client.c:87] Unable to receive IPC response
The Wayland connection broke. Did the Wayland compositor die?
00:00:00.000 [ERROR] [common/ipc-client.c:67] Unable to connect to /run/user/1000/sway-ipc.1000.361818.sock
Gdk-Message: 20:48:25.249: Error reading events from display: Connection reset by peer

sahinf:
problem: an invalid usage of the command `resize set` will cause sway to crash because it doesn't check for an invalid height.

solution:
1. validate height along with width
2. collect all the resize returns into a var in order to log potential errors before returning, so the user has more context as to why the resize config did not work.

fixes #8619

### DIFF
--- a/sway/commands/resize.c
+++ b/sway/commands/resize.c
@@ -457,7 +457,7 @@ static struct cmd_results *cmd_resize_set(int argc, char **argv) {
 		if (argc > num_consumed_args) {
 			return cmd_results_new(CMD_INVALID, "%s", usage);
 		}
-		if (width.unit == MOVEMENT_UNIT_INVALID) {
+		if (height.unit == MOVEMENT_UNIT_INVALID) {
 			return cmd_results_new(CMD_INVALID, "%s", usage);
 		}
 	}


sway/commands/resize.c
```diff
@@ -582,17 +581,16 @@ struct cmd_results *cmd_resize(int argc, char **argv) {
...
+	else if (strcasecmp(argv[0], "shrink") == 0) {
+		error = cmd_resize_adjust(argc - 1, &argv[1], -1);
 	}
-
-	const char usage[] = "Expected 'resize <shrink|grow> "
-		"<width|height|up|down|left|right> [<amount>] [px|ppt]'";
-
-	return cmd_results_new(CMD_INVALID, "%s", usage);
+	if (error->status == CMD_INVALID) {
+		sway_log(SWAY_ERROR, "Invalid resize command: '%s'", error->error);
```
emersion, line 593: Hm, in general we don't log anything on invalid commands. The returned error message is already printed to the user.

Isn't the error message descriptive enough?

sway/commands/resize.c
```diff
@@ -582,17 +581,16 @@ struct cmd_results *cmd_resize(int argc, char **argv) {
...
-		return cmd_resize_adjust(argc - 1, &argv[1], -1);
+	else if (strcasecmp(argv[0], "shrink") == 0) {
+		error = cmd_resize_adjust(argc - 1, &argv[1], -1);
 	}
-
-	const char usage[] = "Expected 'resize <shrink|grow> "
-		"<width|height|up|down|left|right> [<amount>] [px|ppt]'";
-
-	return cmd_results_new(CMD_INVALID, "%s", usage);
+	if (error->status == CMD_INVALID) {
```
emersion, line 592: This will segfault when there is no error.

sway/commands/resize.c
```diff
@@ -457,7 +456,7 @@ static struct cmd_results *cmd_resize_set(int argc, char **argv) {
 		if (argc > num_consumed_args) {
 			return cmd_results_new(CMD_INVALID, "%s", usage);
 		}
-		if (width.unit == MOVEMENT_UNIT_INVALID) {
+		if (height.unit == MOVEMENT_UNIT_INVALID) {
```
emersion, line 459: Good catch! This as a bugfix (without the error handling changes below) looks good to me.

sway/commands/resize.c
sahinf: Oops, didn't see. Thanks.

sahinf: I removed these changes and tested with `--buildtype=release` but I don't see the error message from the original issue after running with `-d`:
```
00:00:01.171 [ERROR] [sway/commands/resize.c:405] resize_set_floating:invalid height unit"
```
which is a call to `sway_assert`.
screenshot_2025-04-01_23-50-58_452064014

sahinf: | Isn't the error message descriptive enough?

So the reason I wasn't seeing the user's error message is because it gets logged in `resize_set_floating` through sway_assert, but it DOES NOT get logged in `cmd_resize_set` which is what gets called for the following `for_window` block:
```swayconfig
for_window [app_id="foot"] {
	floating enable
	move position 600 pt 200 pt
	resize set width 300 pt
	resize set height 300 pt
}
exec foot
```

`cmd_resize_set` additionally calls `resize_set_floating` on floating windows, which would resize it again.

So with all of this in mind, I think I should bring back these changes to bubble up the error from `cmd_resize_set` instead of returning with nothing because 
1. there will not be any error messages at all, unlike in the original issue
2. And although I have not tested this but rather read through the code, it looks like the other `cmd_resize_adjust` also does not log it's errors but returns the result of the command.

If you disagree, then this PR is ready to merge (already removed the error message)
but if you agree, I will bring back those changes and it should handle logging the other resizing errors.

common/util.c
```diff
@@ -99,7 +99,6 @@ int parse_movement_amount(int argc, char **argv,
 	// Try the second argument
 	amount->unit = parse_movement_unit(argv[1]);
 	if (amount->unit == MOVEMENT_UNIT_INVALID) {
-		amount->unit = MOVEMENT_UNIT_DEFAULT;
```
emersion, line 102: Why remove this? I think it was useful to process e.g. `move position 10 20` where the unit is optional twice in the same command.

sway/commands.c
```diff
@@ -295,6 +295,7 @@ list_t *execute_command(char *_exec, struct sway_seat *seat,
 			struct cmd_results *res = handler->handle(argc-1, argv+1);
 			list_add(res_list, res);
 			if (res->status == CMD_INVALID) {
+				sway_log(SWAY_ERROR, "%s", res->error);
```
emersion, line 298: We don't log command error results, these are displayed to the user by swaymsg.

sway/commands/move.c
```diff
@@ -916,8 +907,7 @@ static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
 	case MOVEMENT_UNIT_DEFAULT:
 		break;
 	case MOVEMENT_UNIT_INVALID:
-		sway_assert(false, "invalid x unit");
-		break;
```
emersion, line 920: It seems like this was unreachable with the previous code, because there was a `lx.unit == MOVEMENT_UNIT_INVALID` check above? Why change it?

sway/commands/resize.c
```diff
@@ -378,8 +378,8 @@ static struct cmd_results *resize_set_floating(struct sway_container *con,
 			con->pending.width = width->amount;
 			break;
 		case MOVEMENT_UNIT_INVALID:
-			sway_assert(false, "invalid width unit");
-			break;
+			return cmd_results_new(CMD_INVALID,
+						  "Invalid width unit, check man 5 sway for correct usage");
```
emersion, line 382: Style nit: we don't align to parentheses, we use one or two tabs.

sway/commands/resize.c
```diff
@@ -378,8 +378,8 @@ static struct cmd_results *resize_set_floating(struct sway_container *con,
 			con->pending.width = width->amount;
 			break;
 		case MOVEMENT_UNIT_INVALID:
-			sway_assert(false, "invalid width unit");
-			break;
```
emersion, line 382: I'm not sure how this was reachable before, because the only caller is `cmd_resize_set()` and that function already checks for invalid width/height amounts?

common/util.c
sahinf: I was testing and thought that it failed to catch `move position 600 pp 200 pp` because of that line, but I can't reproduce so turns out it was a PEBKAC error. Reverted that commit.

sway/commands.c
sahinf: But these errors are from for_window directives within the config, not from the user running swaymsg. They currently don't get displayed to the user at all. Should they not get printed if from for_window directives?
```swayconfig
for_window [app_id="Alacritty"] {
	floating enable
	move position 600 pt 200 pt
	resize set width 300 pt
	resize set height 300 pt
}
exec alacritty
```

sway/commands/resize.c
```diff
@@ -582,17 +582,13 @@ struct cmd_results *cmd_resize(int argc, char **argv) {
...
-		return cmd_resize_adjust(argc - 1, &argv[1], -1);
+	else if (strcasecmp(argv[0], "shrink") == 0) {
+		error = cmd_resize_adjust(argc - 1, &argv[1], -1);
 	}
-
-	const char usage[] = "Expected 'resize <shrink|grow> "
-		"<width|height|up|down|left|right> [<amount>] [px|ppt]'";
-
-	return cmd_results_new(CMD_INVALID, "%s", usage);
+	return error;
```
emersion, line 593: Seems like with this change we no longer report an error when executing `resize foo`.

sahinf: Weird on my end, it does report it. Which config did you test with?
screenshot_2025-04-20_11-54-35_709649975

emersion: `resize foo bar` crashes with the following ASan report, because it's left NULL:

```
../sway/commands.c:297:11: runtime error: member access within null pointer of type 'struct cmd_results'
AddressSanitizer:DEADLYSIGNAL
==18280==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x5e59b8c215b0 bp 0x7ffe45326960 sp 0x7ffe453267f0 T0)
==18280==The signal is caused by a READ memory access.
==18280==Hint: address points to the zero page.
    #0 0x5e59b8c215b0 in execute_command ../sway/commands.c:297
    #1 0x5e59b8c587c4 in ipc_client_handle_command ../sway/ipc-server.c:650
    #2 0x5e59b8c55012 in ipc_client_handle_readable ../sway/ipc-server.c:261
    #3 0x7c94a24f1111 in wl_event_loop_dispatch (/usr/lib/libwayland-server.so.0+0xa111) (BuildId: 2e487f08f7f64aa6e49c21f76a244c230d6c141f)
    #4 0x7c94a24f31f6 in wl_display_run (/usr/lib/libwayland-server.so.0+0xc1f6) (BuildId: 2e487f08f7f64aa6e49c21f76a244c230d6c141f)
    #5 0x5e59b8c6e584 in server_run ../sway/server.c:517
    #6 0x5e59b8c64c0b in main ../sway/main.c:389
    #7 0x7c949fc35487  (/usr/lib/libc.so.6+0x27487) (BuildId: 0b707b217b15b106c25fe51df3724b25848310c0)
    #8 0x7c949fc3554b in __libc_start_main (/usr/lib/libc.so.6+0x2754b) (BuildId: 0b707b217b15b106c25fe51df3724b25848310c0)
    #9 0x5e59b8c1eeb4 in _start (/home/simon/src/sway/build/sway/sway+0x38beb4) (BuildId: 529fc248538432f19e80e6784ffa9f320bf0aa04)
```

sway/commands.c
emersion: (But TBH, this should be a separate commit or PR.)

sway/commands/resize.c
emersion: (In general, I'm not sure why this whole code block has been modified - the old version looked fine to me.)

emersion (approved):
Thanks!

emersion:
Ref https://github.com/swaywm/sway/pull/8685 to log a message when a `for_window` command fails.

PR #8652, opened 2025-04-08
ipc: standardize pretty print with raw print

### LINKED ISSUE #8584: `type:mouse` doesn't work but `type:pointer` does
- **Description:**
This configs for type is pretty inconsistent, `type:touchpad` and `type:keyboard` works as expected as per `swaymsg -t get_inputs`. However, in `get_inputs` it shows the type as `Mouse`, while the configs require the type as `pointer` as per `libinput list-devices`.

I think it would be more user friendly to either setup up aliases for `type:mouse` -> `type:pointer`, or change the output for `swaymsg -t get_inputs` to show `Pointer` instead of `Mouse`.

sahinf:
`swaymsg -t get_inputs --raw` calls it a pointer but `--pretty` calls it a Mouse. Previous commit 6737b90cb that set this to pointer probably forgo to update the pretty one.

closes #8584

### DIFF
--- a/swaymsg/main.c
+++ b/swaymsg/main.c
@@ -99,7 +99,7 @@ static const char *pretty_type_name(const char *name) {
 		const char *b;
 	} type_names[] = {
 		{ "keyboard", "Keyboard" },
-		{ "pointer", "Mouse" },
+		{ "pointer", "Pointer" },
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

PR #8654, opened 2025-04-10
Add some tests about the sway config unit

michelepagot:
Create a test infrastructure in the sway subproject buildsystem
Create few tests about the config file activation and validation.
Add couple of checks in config.c to avoid null pointers exceptions when running tests.


emersion:
Hm, I'm not sure we want to build and maintain Sway unit tests…

PR #8656, opened 2025-04-12
Separate adjacent borders in swaybar

kupospelov:
Unlike `i3bar`, `swaybar` does not render the workspace/mode separator.

IMO stuck together borders are kind of ugly, especially in the default configuration, and even more so when the bar is on top.

For a long time I could ignore this by enabling thicker borders for all windows and effectively disabling workspace borders, but recently I made my configuration look more like the default with smart borders turned on, and now this issue is harder to ignore.

Before:
before

After:
after

### DIFF
--- a/include/swaybar/tray/item.h
+++ b/include/swaybar/tray/item.h
@@ -53,6 +53,6 @@ struct swaybar_sni {
 struct swaybar_sni *create_sni(char *id, struct swaybar_tray *tray);
 void destroy_sni(struct swaybar_sni *sni);
 uint32_t render_sni(cairo_t *cairo, struct swaybar_output *output, double *x,
-		struct swaybar_sni *sni);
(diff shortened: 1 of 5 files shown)


kupospelov:
I think an even better option might be to make this padding configurable, so that those who prefer how `i3` renders the buttons could, for example, set top+right button padding for bottom bars or bottom+right button padding for top bars.

A feature like this never existed in `i3` though, so given the new feature policy I just suggest making it consistent with `i3` instead.

grolongo:
It's reassuring to see that I'm not the only one obstinate with such small details, which indeed, was bothering me as much as you. I switched from i3 recently and it was driving me nuts when I tried to configure this, until I googled the issue and found this PR.

Thank you for this, hopefully this gets reviewed and merged soon.

PR #8661, opened 2025-04-14
server: fix shutdown crash when running on x11 backend

agorgl:
When running with X11 backend, DRM is not used, so deinitialization on fini should be skipped.
With this, after https://github.com/swaywm/sway/pull/8578 sway has no more crashes on shutdown.


emersion (approved):
Thanks!

PR #8663, opened 2025-04-14
Remove duplicate call to arrange_container

FilipVujicic:
Related to https://github.com/swaywm/sway/pull/8606
One of the calls to `arrange_container` was left outside the if block


kennylevinsen (approved):
LGTM, thanks!

PR #8668, opened 2025-04-16
layer_shell: destroy layer_surface on assigned output destruction

llyyr:
According to the spec, the closed event should be sent when the surface is no longer shown, because the output may have been destroyed or the user may have asked for it to be removed. In such cases, the clients should destroy the resource.

This fixes mako not being able to show notifications if the assigned output was destroyed while a notificataion was still visible

### DIFF
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -221,7 +221,7 @@ static void handle_output_destroy(struct wl_listener *listener, void *data) {
 		wl_container_of(listener, layer, output_destroy);
 
 	layer->output = NULL;
-	wlr_scene_node_destroy(&layer->scene->tree->node);
+	wlr_layer_surface_v1_destroy(layer->layer_surface);
 }
 
 static void handle_node_destroy(struct wl_listener *listener, void *data) {


llyyr:
I can't really understand what the previous code was trying to do here, so I don't know if this is completely incorrect but this behavior matches what _every_ other compositor does with on the output disable event.

emersion:
It seems like `wlr_layer_surface_v1_destroy()` was called on output destroy before 188811f80861caacd016b857b0d07f6d2d62d15a.

I think we're missing a layer shell surface destroy handler? Here the `struct sway_layer_surface` seems to be leaked when `wlr_layer_surface_v1_destroy()` is called? I'm actually surprised we don't hit asserts due to dangling listeners.

(There are other cases where a layer surface might get destroyed, e.g. when a client disconnects.)

emersion:
I think this still leaves the `struct sway_layer_surface` behind when a client destroys the layer shell surface.

llyyr:
It's handled here https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/layer_shell_v1.c#L15-20

emersion:
Oh right, my bad, I forgot that we were using the scene-graph layer shell helper now…

I think I preferred the previous version of this patch.

llyyr:
Changed it back

emersion (approved):
Thank you, and sorry for the noise!

PR #8670, opened 2025-04-17
sway_terminate() cleanups

emersion:
_See individual commits._

### DIFF
--- a/sway/commands/exit.c
+++ b/sway/commands/exit.c
@@ -1,8 +1,7 @@
 #include <stddef.h>
 #include "sway/commands.h"
 #include "sway/config.h"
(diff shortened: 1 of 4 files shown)


sway/commands/exit.c
```diff
@@ -2,8 +2,6 @@
 #include "sway/commands.h"
 #include "sway/config.h"
```
kennylevinsen, line 3: There's a missing include for `sway/server.h`.

emersion: Oops, fixed!

kennylevinsen (approved):
LGTM!

PR #8671, opened 2025-04-17
exec: export matched container id if present

### LINKED ISSUE #9038: sway does not export I3_WINDOW_ID
- **i3 PR:** https://github.com/i3/i3/pull/6160

- **Description:**
  - with the above (recent) PR, i3 provides `I3_WINDOW_ID` with exec whenever possible, which is useful for scripting.

Created #8671 to add this to sway.

RobertMueller2:
Closes: #9038 

Provides matched container id to each command execution via handler_context, if `--no-matched-container-id` is not set with exec
or exec_always. If there is no matched node, the handler_context has the focused node. The idea is that this can simplify scripting.

- Relates to [PR 6160 at i3](https://github.com/i3/i3/pull/6160). The i3 PR does not have `--no-matched-container-id`. After a while I decided for it because it felt better. [thought process](https://github.com/swaywm/sway/pull/8671#issuecomment-2820857830) (This was initially --matched-container-id, but after I3 added the PR, I decided to revert the logic)
- I decided against a function and for inline placement for no particular reason.
  - Since I made it inline though, `goto` seemed like a cleaner choice than guards or nesting more ifs
- I looked at `mark.c` for inspiration for multi argument handling. This could potentially do with redundancy reduction now, but that may be valid for even more command files. I did not feel comfortable addressing these things with this simple PR.
- I may accidentally have fixed a bug while doing this in line 35 (old). The below is what I observe _before patch_ for using `--no-startup-id` and then triggering an error due to no further args -- I believe this should say "exec" or "exec_always", not "--no-startup-id".
```
> swaymsg -- exec --no-startup-id
Error: Invalid --no-startup-id command (expected at least 1 argument, got 0)
```
- this is now 
```
Error: Invalid exec command (expected at least 1 argument, got 0)
```
and
```
Error: Invalid exec_always command (expected at least 1 argument, got 0)
```
for the tests below.

I'm using this change in my regular sway session.

Tested with (edited back to --matched-ids):

```
#!/bin/sh

i=1

echo "20 tests"
echo

for exec_variant in "exec" "exec_always" ; do
    for command_parameter in 'env | grep "\(DESKTOP_STARTUP_ID\|SWAY_EXEC_CON_ID\|I3_WINDOW_ID\)"' '' ; do
        for mode in 0 1 2 3 7; do
            HAVE_NODS=
            HAVE_CONID=
            # 1 = DESKTOP_STARTUP_ID in front, 4 = DESKTOP_STARTUP_ID in the back
            if [ $(($mode & 0x1)) -eq $((0x1)) -o $(($mode & 0x4)) -eq $((0x4)) ];then
                HAVE_NODS=1
            fi

            if [ $(($mode & 0x2)) -eq $((0x2)) ];then
                HAVE_CONID=1
            fi

            if [ -n "$command_parameter" ];then
                swaymsg -- exec "echo \"test ${i}: mode: ${mode}, exec_variant: ${exec_variant}, command_parameter: ${command_parameter:+set}\n---\nexpect: DESKTOP_STARTUP_ID ${HAVE_NODS:+not }set, SWAY_EXEC_CON_ID ${HAVE_CONID:-not set}${HAVE_CONID:+set}\""
            else
                swaymsg -- exec "echo \"test ${i}: mode=${mode}, exec_variant=${exec_variant}, command_parameter ${command_parameter:+set}\n---\nexpect: error message in terminal\""
            fi

            sleep .5
            swaymsg -- exec "echo \"actual command: ${exec_variant}${HAVE_NODS:+ --no-startup-id}${HAVE_CONID:+ --matched-ids}${command_parameter:+ }$command_parameter\""
            sleep .5
            swaymsg -- ${exec_variant}${HAVE_NODS:+ --no-startup-id}${HAVE_CONID:+ --matched-ids}${command_parameter:+ }$command_parameter
            sleep .5
            swaymsg -- exec 'echo "---"'
            i=$((i+1))
            read -p "press key to continue..." x
        done
    done
done
```

My demo use case:
```
[title="Picture-in-Picture"] floating enable, sticky enable, mark --add _newstickychange, mark --add _newnotitle, resize set 20ppt 20ppt, exec swayhelper move_window_to_corner 3 --title Picture-in-Picture
```

can now become

```
[title="Picture-in-Picture"] floating enable, sticky enable, mark --add _newstickychange, mark --add _newnotitle, resize set 20ppt 20ppt, exec --matched-ids swayhelper move_window_to_corner 3
```

While the former would even likely work without any matching in my swayhelper, just working on focused node, it is not guaranteed if this was matching several windows, in that case I have to duplicate the criteria as shown above. The latter is not shorter depending on the actual criteria, but has less redundancy.

### DIFF
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -10,9 +10,28 @@
 #include "sway/desktop/launcher.h"
 #include "sway/tree/container.h"
 #include "sway/tree/root.h"
+#include "sway/tree/view.h"
 #include "sway/tree/workspace.h"
 #include "log.h"
 #include "stringop.h"
+
+static void export_id(const char *env_var_name, size_t id) {
+	int id_len = snprintf(NULL, 0, "%zu", id);
+	if (id_len < 0) {
+		sway_log(SWAY_ERROR, "Unable to determine buffer length for %s", env_var_name);
+		return;
+	}
+	// accommodate \0
+	id_len++;
+	char* id_str = malloc(id_len);
+	if (!id_str) {
+		sway_log(SWAY_ERROR, "Unable to allocate buffer for %s", env_var_name);
+		return;
+	}
+	snprintf(id_str, id_len, "%zu", id);
+	setenv(env_var_name, id_str, 1);
+	free(id_str);
+}
 
 struct cmd_results *cmd_exec_validate(int argc, char **argv) {
 	struct cmd_results *error = NULL;
@@ -28,12 +47,12 @@ struct cmd_results *cmd_exec_validate(int argc, char **argv) {
 struct cmd_results *cmd_exec_process(int argc, char **argv) {
 	struct cmd_results *error = NULL;
 	char *cmd = NULL;
-	bool no_matched_container_id = false;
+	bool matched_ids = false;
 	bool no_startup_id = false;
 	int argv_counter = -1;
 	while (argc > 0 && has_prefix(*argv, "--")) {
-		if (strcmp(argv[0], "--no-matched-container-id") == 0) {
-			no_matched_container_id = true;
+		if (strcmp(argv[0], "--matched-ids") == 0) {
+			matched_ids = true;
 		} else if (strcmp(argv[0], "--no-startup-id") == 0) {
 			no_startup_id = true;
 		} else {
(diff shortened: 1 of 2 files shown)


RobertMueller2:
As I was optimising some of my scripts my eye got caught by SWAY_EXEC_CON_ID in a terminal window. Of course I was aware before the patch does this, it's the whole point of it. But just seeing the var while doing something entirely different triggered some new thought process, I guess.

If I start an application from the terminal, it'll have this env var from the terminal. I don't think that has consequences, but it would be an environment with an ID from another window, and that does not feel right. 

Secondly, while I'm not sure if I can construct a case where the variable is "too much information" (e.g. in sandboxing scenarios I'd always make sure the environment is clean), only exporting it on purpose in select scenarios feels better than doing it always.

So I think I'll make this a draft again and try and add a flag for exec.

RobertMueller2:
Now that i3 has implemented it, I'm thinking, it would be better to allow disable rather than having to enable the export. Also, for compatibility, this should use `I3_WINDOW_ID`.

kennylevinsen:
This PR uses the con ID, but the i3 feature uses the X11 window ID from what I can tell. This makes them incompatible, so setting `I3_WINDOW_ID` might be misleading at best.

RobertMueller2:
> This PR uses the con ID, but the i3 feature uses the X11 window ID from what I can tell. This makes them incompatible, so setting `I3_WINDOW_ID` might be misleading at best.

Ah yes, that's true. Sad :grin:  I suppose I should have tried that. Which I might have if I didn't have to exit my sway session or have to start Xephyr.

Anyway, I suppose we could export the X11 window ID when available. I'm not sure if it's useful but I wouldn't mind.

The question is, does this invalidate the idea of exporting the con id.

kennylevinsen:
> Ah yes, that's true. Sad 😁 I suppose I should have tried that. Which I might have if I didn't have to exit my sway session or have to start Xephyr.

You can run i3 under sway:

```
Xwayland :1234 -decorate &
sleep 1

unset WAYLAND_DISPLAY
unset SWAYSOCK
unset I3SOCK
DISPLAY=:1234 i3 -c some-i3-test-config
kill -TERM %1
```

Just prepare a config that uses e.g., alt as mod key.

RobertMueller2:
It's now exporting the actual X11 window id with `I3_WINDOW_ID` if available.

While testing this, I remembered why I had made the exports opt-in to begin with, so I went back to that. It's somewhat confusing when e.g. using a keybind to start applications (e.g. a terminal), and then SWAY_EXEC_CON_ID refers to the last focused node instead of the just started one. Logical, but not that intuitive. So, with the last commit (again), only when `--matched-ids` is given, `SWAY_EXEC_CON_ID` and `I3_WINDOW_ID` are exported.

PR #8674, opened 2025-04-20
Use pthread_atfork() to restore signals and NOFILE limit

emersion:
This ensures these functions are always called (even when a library such as wlroots or libc perform the fork) and removes the need to manually call them.


kennylevinsen (approved):
LGTM.

PR #8678, opened 2025-04-23
input/seatop_default: handle continuous scroll events

### LINKED ISSUE #8014: Touchpad Scrolling Not Working in Tabbed/Stacked Mode
I've encountered an issue while using Regolith with the Sway window manager, specifically when windows are in tabbed or stacked mode. The problem is that scrolling through the windows (to change the active one) with the touchpad does not work, whereas using a mouse's scroll wheel functions as expected.

**Environment:**
Sway Version: 1.8
Operating System: Ubuntu

**Steps to Reproduce:**
Open multiple windows in Sway under Regolith.
Change the window layout to either tabbed or stacked mode.
Attempt to scroll within a window title using the touchpad.

**Expected Behavior:**
Scrolling with the touchpad should allow me to change the active window, how it works when using a mouse.

**Actual Behavior:**
Scrolling with the touchpad has no change in which the active window is, while mouse scrolling works perfectly.

Thank you for your time and effort in maintaining Sway.

kupospelov:
The implementation is similar to how swaybar handles these events, except that swaybar does it in a pointer frame event handler.

Fixes #8014

### DIFF
--- a/include/sway/input/cursor.h
+++ b/include/sway/input/cursor.h
@@ -8,6 +8,7 @@
 #include "sway/input/seat.h"
 #include "config.h"
 
+#define SWAY_CONTINUOUS_SCROLL_TIMEOUT 1000
 #define SWAY_CURSOR_PRESSED_BUTTONS_CAP 32
 
 #define SWAY_SCROLL_UP KEY_MAX + 1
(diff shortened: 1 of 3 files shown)


kupospelov:
@emersion could you please take a look at this when you have a chance?

PR #8680, opened 2025-04-25
input/seatop_down: Update decorations for touchscreen inputs

### LINKED ISSUE #8675: Touching windows with a touchscreen updates keyboard focus but not client decoration focus
- **Sway Version:**
1.10 (from Void Linux's package) and 1.10.1 (from Chimera Linux's package) on different machines.

- **Debug Log:**
The logs were generated by sway 1.10 running on Void Linux.
[sway.log](https://github.com/user-attachments/files/19834119/sway.log)

- **Configuration File:**
I moved my configuration files out of the search path (renamed the .config/sway directory to swaybak) so that default configuration is applied.  The problem happens both with default configuration and custom configuration.
From testing, the `focus_follows_mouse` setting has no effect on the problem.

- **Description:**
  - Start sway on a machine featuring a touchscreen (tablet, 2-in-1 laptop, etc).
  - Open 2 or more windows (I opened two instances of foot for the debug logs).
  - Try typing on your keyboard: the window that is shown as focused by the window decorations (e.g. blue borders with default config, while unfocused windows have grey borders) should be receiving keyboard input.
  - Now, with a touchscreen, touch a window other than the one currently focused.
  - The window decorations are not updated to reflect a change in focus, the previously focused window still has blue borders while the others have grey ones (with default configurations).
  - Try typing on your keyboard: the window you touched now has keyboard focus and receives all keyboard inputs and is the container sway commands would act upon, e.g. sending a `kill` command through `$mod+Shift+q` (as is the default) will kill the window you last touched with the touchscreen, even though the window decorations indicate that this window you touched is not the one focused and should not receive keyboard input.

Edit: Additionally, if it's worth anything, i3 *does* update the window decorations to reflect a change in focus on touch, both with `focus_follows_mouse` on `yes` and `no` (haven't used X in a while so I don't know if it's a configuration I did or the default behaviour but touching the screen moves the cursor under i3 so I thought focus_follows_mouse could affect the behaviour there)

sahinf:
fixes #8675

### DIFF
--- a/sway/input/seatop_down.c
+++ b/sway/input/seatop_down.c
@@ -100,6 +100,7 @@ static void handle_touch_down(struct sway_seat *seat,
 
 	if (focused_node) {
 	    seat_set_focus(seat, focused_node);
+		transaction_commit_dirty();
 	}
 }
 


madeleine313:
Can confirm on my end it does fix the problem. (thanks for working on this, greatly appreciated)

emersion (approved):
Thanks!

PR #8684, opened 2025-04-27
Replace signal() with sigaction()

emersion:
The man page for signal(3) reads:

> new applications should use sigaction() rather than signal()


kennylevinsen (approved):
Partly duplicates https://github.com/swaywm/sway/pull/8598, but LGTM.

PR #8687, opened 2025-04-29
config/output: Fix missing output config supersedes

kennylevinsen:
color_transform and allow_tearing was not handled by supersede_output_config which could lead to configuration being incorrectly applied.


emersion (approved):
Thanks!

Issue #8607, opened 2025-03-13
swaybar: replace icon lookup with libsfdo


emersion:
We have our own icon lookup code, but it's not implementing the spec to the letter.

Switching to this library would allow us to get rid of this stuff:
https://gitlab.freedesktop.org/vyivel/libsfdo/

emersion:
IIRC the FDO notification spec states that the icon may be either an absolute path or an icon theme name. We already handle absolute path and can keep that logic (ie, only hand over the icon name to libsfdo if it's not an absolute path).

emersion:
Oh, I mxied things up and thought this was about mako and not about swaybar.

Anyways, I think it's completely fine and even desirable to have logic outside of the scope of XDG icon theme lookup in Sway instead of libsfdo. Absolute path handling falls into that category. File extension stripping sounds like a workaround for buggy apps, would be better to fix the apps to conform to the spec.

vyivel:
libsfdo-desktop doesn't concern itself with making sense of the `Icon` value, which means it's the user's responsibility to check whether it's an absolute path and act accordingly. libsfdo-icon is for finding an icon path by its name (and other parameters), which is not required if you already have a path. I don't see a problem here.

Regarding icon names with extensions, I too consider those to be application bugs.

PR #8689, opened 2025-04-30
Integrate libsfdo

myrslint:
This hopes to address https://github.com/swaywm/sway/issues/8607.

As of this PR, libsfdo is made an optional build-time dependency. Swaybar's own icon handling code is kept intact with libsfdo alternative code paths wrapped in preprocessor conditionals. It can be trivially refactored to make libsfdo a requirement followed by entirely removing own icon handling code.

There is as well a tiny amount of extra code which, if there is interest and permission to do so, I would want to use to add a new feature to swaybar where a workspace button optionally displays a list of the icons of the applications running on that workspace. With theme handling and icon caching by libsfdo this can be done efficiently. It would obviate the need for utilities such as sworkstyle or Waybar's workspace display rewrite rules.

### DIFF
--- a/common/sfdo.c
+++ b/common/sfdo.c
@@ -1,104 +1,123 @@
-#include <sys/stat.h>
-#include <stdlib.h>
 #include <sfdo-basedir.h>
 #include <sfdo-desktop.h>
 #include <sfdo-icon.h>
+#include <stdlib.h>
+#include <sys/stat.h>
 
 #include "log.h"
 #include "sfdo.h"
 
-// this extends libsfdo's behavior to also handle icons specified as absolute paths
-char *sfdo_icon_lookup_extended(struct sfdo *sfdo, char *icon_name, int target_size, int scale) {
-	char *icon_path = NULL;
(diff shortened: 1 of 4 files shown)


emersion:
Thanks for sending a patch!

A would prefer to make libsfdo a hard dependency for swaybar, so that we can drop all of the old code and avoid the conditionals.

myrslint:
> Thanks for sending a patch!
> 
> A would prefer to make libsfdo a hard dependency for swaybar, so that we can drop all of the old code and avoid the conditionals.

Hello there again. I think I botched both my local and remote copies of this branch when trying to do a rebase. Since I don't have the wherewithal to redo it all over again please consider the PR stale. If anyone else is willing to try their hand at this again please feel most welcome to take whatever useful bits you find in there. Thank you.

emersion:
I'd suggest reading this to understand how to rebase: https://git-rebase.io/

Here it should be possible to `git rebase -i origin/master` and then delete all lines except your own commits.

PR #8690, opened 2025-04-30
config/output: Use INT_MAX as x/y unset value

kennylevinsen:
We oftne use -1 to indicate unset values. In case of output (x, y), we would consider the fields set if they are not both -1. This means that (0, -1) and (-1, 0) are valid coordinates, but (-1, -1) is not.

We support negative output positioning, so we cannot use -1 to mean unset. Zero is also not an option as that would disallow reverting a set position back to (0, 0).

INT_MAX is an unreasonable output position, so use it to indicate unset values, and only use the value when both are set.


Emantor (approved):

emersion (approved):
Thanks!

PR #8694, opened 2025-05-02
fix #8681

### LINKED ISSUE #8681: Broken titlebar when container takes focus from fullscreen container
- **Sway Version:**
  - sway version 1.10.1

- log: https://gist.github.com/alex-huff/e566a3793ec6684f4200a984f1fa635c

- **Description:**
When a fullscreen container in a tabbed layout losses focus to one of its siblings, it's title bar will temporarily be broken. This bug happens with the default configuration.

To reproduce:
1. Go to empty workspace.
2. Switch workspace to tabbed layout.
3. Open two terminals.
4. Run `con_id=$(swaymsg -t get_tree | jq "recurse(.floating_nodes[], .nodes[]) | select(.focused==true).id"); sleep 2; swaymsg "[con_id=$con_id] focus"` in one of the terminals then quickly switch to the other terminal and put it in fullscreen.
5. After 2 seconds the original terminal should be refocused and the titlebar of the other terminal will be broken.

In this video you can see that after my fullscreen browser loses focus its titlebar isn't rendered.

https://github.com/user-attachments/assets/044fa6c8-d3ac-4bdb-8ff8-404da7e1f4fb

alex-huff:
Before this commit when a child of a tabbed or stacking container was taken out of fullscreen and a different sibling was focused in the same transaction, the titlebar of the previously fullscreen container would remain hidden. This commit makes sure that scene tree for decorations is enabled for all containers within a tabbed or stacking container when it is arranged.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -309,6 +309,7 @@ static void arrange_children(enum sway_container_layout layout, list_t *children
 			arrange_title_bar(child, title_offset, -title_bar_height,
 				next_title_offset - title_offset, title_bar_height);
 			wlr_scene_node_set_enabled(&child->border.tree->node, activated);
+			wlr_scene_node_set_enabled(&child->scene_tree->node, true);
 			wlr_scene_node_set_position(&child->scene_tree->node, 0, title_bar_height);
 			wlr_scene_node_reparent(&child->scene_tree->node, content);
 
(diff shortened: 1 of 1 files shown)


Nefsen402 (approved):

Nefsen402:
Thanks!

Nefsen402:
Hmm, it seems like CI failed. Try pushing the commit with a different hash to re-run CI. Also, while we're here, rename the commit to "transaction: Ensure all tabs are visible in tabbed mode"

alex-huff:
It looks like the CI is completing successfully in newer PRs so I'll try one more time to re-run the CI.

Nefsen402:
Thanks!

PR #8697, opened 2025-05-04
container not in scratchpad: return CMD_FAILURE

### LINKED ISSUE #8696: Matched node in scratchpad cannot show if another matched note exists outside of scratchpad
- **Description:**

I'm running
```sh
foot --app-id xyz
swaymsg '[app_id="xyz"]' move to scratchpad
```

Now I can use `swaymsg '[app_id="xyz"] scratchpad show'`, so far, so good.

If I start another (and do not move the second to scratchpad) as shown below, this no longer works.

The error message and exit code 2 is OK, but I'd at least expect that the one window matched that is on scratchpad would show, because i3 shows it if I try to replicate it there.

```
foot --app-id xyz
swaymsg '[app_id="xyz"] scratchpad show'
Error: Container is not in scratchpad.
```

I think this might be related to #8615 . If I change the line in question to return `CMD_FAILURE` instead of `CMD_INVALID`, I still get exit code 2 and an error message. I'll add a PR in a sec.

EDIT:
sway at f9945d81fb52c81ab60034dcfc41a2f36f0ed226
wlroots at 156d47c8665e7508623e3e96db0b0300742db937

RobertMueller2:
Closes #8696.

This was changed to CMD_INVALID in #8615, but for a scenario where there are criteria matched containers both inside and outside scratchpad, the container in scratchpad can no longer be shown via `[<criteria>] scratchpad show`.

Changing it to CMD_FAILURE still leads to an error message and exit code 2, but scratchpad show works again.

### DIFF
--- a/sway/commands/scratchpad.c
+++ b/sway/commands/scratchpad.c
@@ -121,7 +121,7 @@ struct cmd_results *cmd_scratchpad(int argc, char **argv) {
 		// we'll return an error. The same is true if the
 		// overridden node is not a container.
 		if (!con || !con->scratchpad) {
-			return cmd_results_new(CMD_INVALID, "Container is not in scratchpad.");
+			return cmd_results_new(CMD_FAILURE, "Container is not in scratchpad.");
 		}
 		scratchpad_toggle_container(con);
 	} else {


RobertMueller2:
Incorrect comparison of i3 behaivour: I used 4.24 instead of latest git, so no longer need this PR.

PR #8701, opened 2025-05-05
output: Minimize interaction with output after destroy

kennylevinsen:
When an output is destroyed, we go through the process of disabling it. This includes evacuating all content away from the output, which can lead to various modifications to the scene. With the scene_output still present, this can lead to things like output_enter events being emitted for the output currently being destroyed.

Ensure that the scene output is destroyed first and that the output is immediately considered disabled.

References: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3974

### DIFF
--- a/include/sway/tree/workspace.h
+++ b/include/sway/tree/workspace.h
@@ -96,7 +96,7 @@ void workspace_output_add_priority(struct sway_workspace *workspace,
 		struct sway_output *output);
 
 struct sway_output *workspace_output_get_highest_available(
-		struct sway_workspace *ws, struct sway_output *exclude);
+		struct sway_workspace *ws);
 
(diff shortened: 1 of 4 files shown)


kennylevinsen:
Currently only tested with some mild wayland backend tests which did not seem to implode, would be good if a reproducer for the original issue could be tested.

Nefsen402 (approved):
Waiting for issue reporter to confirm the fix.

emersion:
Looks good to me as well, thanks for the fix!

kennylevinsen:
Removed a no longer needed argument to workspace_output_get_highest_available.

kennylevinsen:
Poked original reporter, but even without reproduction the fix is probably a sound enough change.

emersion:
Thanks!

PR #8703, opened 2025-05-06
transaction: Fix size of children container decorations in stacking l…

### LINKED ISSUE #8686: in RC1 stacked window borders extend beyond the bottom of the window by a number of pixels equal to the top titlebar
- **Sway Version: 1.11-rc1-0e19d85d**

- **Debug Log:**
[sway_stacking_issue_2.log](https://github.com/user-attachments/files/19931676/sway_stacking_issue_2.log)

- **Configuration File:**
Mostly default configuration file: 
[sway_testconfig.txt](https://github.com/user-attachments/files/19931679/sway_testconfig.txt)

- **Description:**

When two or more windows are stacked the border on the outside of the window extends downwards past the bottom of the window by a distance equal to the height of the title bar.  You can see in the attached scree capture that when I stack windows there is a vertical border on top of the bar at the bottom of the screen.  I highlight the issue with my mouse cursor at about 25 seconds.

I suspect no one noticed this because the default configuration is to have the bar at the top, so the extra border height would be drawn past the bottom of the screen and not visible.  It's more noticeable with the bar on the bottom.

This is only an issue in the latest RC.  Not in stable.

https://github.com/user-attachments/assets/533e776c-e76a-461d-abdd-f95dccae3db0

Here's a screenshot from my actual configuration which has a bar height that is different from the titlebar height, so the line appears in the middle of the bar.  I couldn't do this with the default config but figured having the screengrab match the logs was more important.

Image

Thanks!

alex-huff:
…ayouts

Before this commit stacking containers with more than one child sized the active container's decorations as if there was only one titlebar. Commit a25645a introduced the local variable 'net_height' but incorrectly calculated it for stacking containers. Fixes #8686.

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -343,7 +343,7 @@ static void arrange_children(enum sway_container_layout layout, list_t *children
 			wlr_scene_node_set_position(&child->scene_tree->node, 0, title_height);
 			wlr_scene_node_reparent(&child->scene_tree->node, content);
 
-			int net_height = height - title_bar_height;
+			int net_height = height - title_height;
 			if (activated && width > 0 && net_height > 0) {
 				arrange_container(child, width, net_height, title_bar_height == 0, 0);
 			} else {


kennylevinsen (approved):
LGTM, thanks!

PR #8704, opened 2025-05-08
move: fix broken titlebar when moving child to new workspace

### LINKED ISSUE #8648: Bogus tab rendered after moving a container from a tabbed parent into another workspace
## Sway version

sway version 1.10.1

## Debug Log

https://gist.github.com/WhyNotHugo/5b4f792f6bbd09138ed3d63ef22e1346

## Configuration File

```
# Use this to `printenv SWAYSOCK` and then control sway from a non-child terminal.
exec foot
```

## Description

When a split container is a child of a tabbed container, if the split container is moved to another workspace, its tab remains visible, and overlap the tab bar indefinitely; the tabs for the remaining children will be occluded by the leftover "floating" tab.

## Reproduction steps

I suggest that you run a nested sway instance, run `printenv SWAYSOCK` and then run these command from non-child terminal. I use `foot` below, but any client which renders a single toplevel is good enough to reproduce this:

```
swaymsg exec foot
swaymsg exec foot
swaymsg exec foot
swaymsg layout tabbed
swaymsg focus left
swaymsg focus left
swaymsg focus right
swaymsg split h
swaymsg exec foot
swaymsg focus parent
swaymsg move container to workspace 9
```

The current workspace now has two tabs. The tab bar renders the two tabs, PLUS
the previous middle tab `H[]` overlapped with the two "real" tabs.

alex-huff:
Before this commit, when moving a non-leaf child of a tabbed or stacking container to a new workspace, the child would be detached from the parent container and the grandchildren would be sent to the new workspace but the child itself wouldn't be destroyed causing the titlebar to still be rendered as part of the parent container. Fixes #8648.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -222,6 +222,7 @@ static void container_move_to_workspace(struct sway_container *container,
 		container_detach(container);
 		if (workspace_is_empty(workspace) && container->pending.children) {
 			workspace_unwrap_children(workspace, container);
+			container_reap_empty(container);
 		} else {
 			container->pending.width = container->pending.height = 0;
 			container->width_fraction = container->height_fraction = 0;


emersion (approved):
LGTM, thanks!

alex-huff:
> Thanks for fixing this and for the clear description, quite interesting to finally understand what was going on.

Thank you for providing instructions for reproducing the issue. I encountered this exact problem around a week ago and I couldn't figure out how it happened which drove me a little crazy. I was glad to stumble upon the issue your created.

PR #8712, opened 2025-05-14
Remove the temporary SUID warning

kennylevinsen:
A temporary SUID detection that would cause sway to exit was introduced when SUID operation was deprecated, intended to avoid cases where a user would not heed the deprecation notice, continued to use SUID and ended up with sway accidentally running as root.

Remove the check, as the three years that have passed is sufficient time for users to discover the deprecation and adapt. We did not care if users intentionally want to run sway as root through SUID, we only wanted to avoid surprise root.

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -108,20 +108,6 @@ static void log_kernel(void) {
 	pclose(f);
 }
 
-static bool detect_suid(void) {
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM!

Maybe we can wait just a little bit more to make the new Sway release before merging this?

PR #8717, opened 2025-05-15
Raise scratchpad container in floating

odyxz:
When a scratchpad container is shown, it may appear underneath other floating containers, making it difficult for users to access it. This can lead to a frustrating user experience, as the scratchpad container is intended to be easily accessible.

### DIFF
--- a/sway/tree/root.c
+++ b/sway/tree/root.c
@@ -199,6 +199,8 @@ void root_scratchpad_show(struct sway_container *con) {
 	if (old_ws) {
 		workspace_consider_destroy(old_ws);
 	}
+
+	container_raise_floating(con);
 }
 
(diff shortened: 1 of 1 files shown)


emersion:
How does i3 behave here? Does it always raise floating containers when showing the scratchpad?

odyxz:
> How does i3 behave here?

It raises container. 
https://github.com/i3/i3/blob/cfa4cf16bea809c7c715a86c428757e577c85254/src/scratchpad.c#L208
https://github.com/i3/i3/blob/cfa4cf16bea809c7c715a86c428757e577c85254/src/con.c#L294

> Does it always raise floating containers when showing the scratchpad?
I use my small patch from January. No issues or undefined behavior.

emersion (approved):
Thanks!

PR #8722, opened 2025-05-15
transaction: ensure border scene is enabled for floating containers

### LINKED ISSUE #8721: Enabling floating on an unfocused child of a tabbed layout only shows titlebar
- **Sway Version:**
  - head

- **Configuration File:**
  - default

- **Description:**
  - switch to an empty workspace
  - open two terminals and put them in tabbed layout
  - switch to the first terminal
  - in the terminal, run `swaymsg "[con_id=$(swaymsg -t get_tree | jq 'first(.nodes | recurse(.[] | .nodes) | select(any(.focused))[-1].id) // "__focused__"')] floating enabled"`
  - notice only the titlebar of the second terminal is visible

Image

alex-huff:
When a container that was previously an inactive child of a tabbed or stacking layout becomes floating, it's border scene tree remains disabled. This results in only the titlebar being rendered for the container. This commit ensures the border scene tree is enabled when arranging floating containers.
Fixes #8721

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -534,6 +534,7 @@ static void arrange_workspace_floating(struct sway_workspace *ws) {
 		wlr_scene_node_set_position(&floater->scene_tree->node,
 			floater->current.x, floater->current.y);
 		wlr_scene_node_set_enabled(&floater->scene_tree->node, true);
+		wlr_scene_node_set_enabled(&floater->border.tree->node, true);
 
 		arrange_container(floater, floater->current.width, floater->current.height,
 			true, ws->gaps_inner);


sway/tree/container.c
```diff
@@ -99,7 +99,7 @@ struct sway_container *container_create(struct sway_view *view) {
 	//   - border
 	//     - border top/bottom/left/right
 	//     - content_tree (we put the content node here so when we disable the
-	//       border everything gets disabled. We only render the content iff there
+	//       border everything gets disabled. We only render the content if there
```
emersion, line 102: I think this means "if and only if": https://en.wikipedia.org/wiki/If_and_only_if

emersion:
LGTM apart from this!

alex-huff:
I have removed the change to the comment.

emersion:
Thanks!

PR #8725, opened 2025-05-18
ext-foreign-toplevel-request-v1: new protocol

navi-desu:
sway impl for https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/406
and matching https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5072


markstos:
The issue title of "ext-foreign-toplevel-request" doesn't match the titles of the links provided, which is "ext-image-capture-request".

navi-desu:
it did before the protocol got completely revamped, i did not have time to update the implementations yet

PR #8731, opened 2025-05-22
layer-shell: reclaim space from unmapped layer surfaces

alex-huff:
wlroots resets 'initialized' when a layer surface is unmapped and sway doesn't rearrange the layer surfaces in response to a commit of a surface where 'initialized' is false. This results in space not getting reclaimed from a recently unmapped layer surface until some other action causes 'arrange_layers' to get called. This commit makes sure all layer surfaces get rearranged when a layer surface is unmapped.


Nefsen402 (approved):
LGTM, Thanks!

PR #8732, opened 2025-05-22
transaction: fix floating fullscreen containers

### LINKED ISSUE #8729: Black fullscreen with tabbed floating container
- **Sway Version:**
swaymsg version 1.11-rc2-5cfcd1c (May 21 2025, branch 'master')

- **What happened:**
Black fullscreen and broken workspace
- **Steps to reproduce:**

1. Open first terminal in tiling mode
2. Open second terminal in tiling mode
3. Make tabbed container
4. Mod+A and make floating container
5. Fullscreen any window with Mod+F

- **Description:**
There are no errors in the stable version. Bugs are only in the master.
When you move a container to a scratchpad, there is a chance that the sway will crash with a segfault.

- **Stack Trace:**
https://gist.github.com/odyxz/87744af48d834c90586cc09c1247aaf2

alex-huff:
8fecf3a introduced a regression where fullscreening a child of a floating container would result in a black screen. This is because the order of 'arrange_fullscreen' and 'arrange_worksplace_floating' was swapped causing the fullscreen container's scene to get reparented after it was parented in the fullscreen layer.

Fixes #8729

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -596,6 +596,7 @@ static void arrange_output(struct sway_output *output, int width, int height) {
 
 				wlr_scene_rect_set_size(output->fullscreen_background, width, height);
 
+				arrange_workspace_floating(child);
 				arrange_fullscreen(child->layers.fullscreen, fs, child,
 					width, height);
 			} else {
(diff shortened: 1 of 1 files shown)


Nefsen402 (approved):
Oops

PR #8737, opened 2025-05-25
Add support for toplevel capture

emersion:
- [x] Handle xdg-popup

References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5078


kennylevinsen:
The TODO already seems taken care of, and the protocol is merged.

Should we undraft and merge this, or merge the first commit (https://github.com/swaywm/sway/pull/8766) first to unbreak build?

PR #8738, opened 2025-05-26
sway/tree: Simplify sway_node teardown

kennylevinsen:
A sway_node may end up being referenced in either a queued transaction, pending transaction or as a dirty node. To manage this, the transaction system has been responsible for destroying containers, workspaces and outputs at the end of their last referenced transaction.

This significantly complicates the teardown flow of surfaces and outputs. Instead, remove the node from transactions and dirty lists so that the callsite can remove and free the node immediately.

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -611,14 +611,16 @@ static struct cmd_results *cmd_move_container(bool no_auto_back_and_forth,
 	if (old_parent) {
 		container_reap_empty(old_parent);
 	} else if (old_ws) {
-		workspace_consider_destroy(old_ws);
+		if (workspace_consider_destroy(old_ws)) {
+			old_ws = NULL;
+		}
 	}
 
 	// arrange windows
 	if (root->fullscreen_global) {
 		arrange_root();
 	} else {
-		if (old_ws && !old_ws->node.destroying) {
+		if (old_ws) {
 			arrange_workspace(old_ws);
 		}
 		arrange_node(node_get_parent(destination));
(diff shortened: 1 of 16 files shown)


sway/tree/container.c
```diff
@@ -548,7 +518,6 @@ void container_begin_destroy(struct sway_container *con) {
 	container_end_mouse_operation(con);
 
 	con->node.destroying = true;
```
emersion, line 520: Do we still use `destroying` for anything?

sway/tree/view.c
```diff
@@ -928,7 +928,7 @@ void view_unmap(struct sway_view *view) {
 
 	struct sway_container *parent = view->container->pending.parent;
 	struct sway_workspace *ws = view->container->pending.workspace;
-	container_begin_destroy(view->container);
+	container_destroy(view->container);
```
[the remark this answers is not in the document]
kennylevinsen: I cleaned this up a bit so that `container_destroy` no longer contains view freeing code, getting rid of `view_begin_destroy` while at it.

sway/tree/container.c
kennylevinsen: I don't think so no - I have removed both it and `view->destroying` now.

sway/commands/move.c
```diff
@@ -618,7 +618,7 @@ static struct cmd_results *cmd_move_container(bool no_auto_back_and_forth,
 	if (root->fullscreen_global) {
 		arrange_root();
 	} else {
-		if (old_ws && !old_ws->node.destroying) {
+		if (old_ws) {
```
emersion, line 621: Hm, `workspace_consider_destroy()` might've destroyed the workspace… Should we make that function return a bool to indicate whether the workspace ended up being destroyed, and reset `old_ws` to NULL in that case?

sway/desktop/output.c
```diff
@@ -446,6 +445,8 @@ static void begin_destroy(struct sway_output *output) {
 	wl_event_source_remove(output->repaint_timer);
 	output->repaint_timer = NULL;
 
+	output_destroy(output);
```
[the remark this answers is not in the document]
emersion: Sounds good!

emersion (approved):
Thanks!

PR #8596, opened 2025-03-08
Add unit ppt to 'move' + direction

Absobel:
Attempt at solving https://github.com/swaywm/sway/issues/8566

Issue #8566, opened 2025-02-10
sway does not support ppt in `move left`


ElementalWarrior:
### Please fill out the following:
This is the command that processes move left for i3
https://github.com/i3/i3/blob/853b0d9161a5f99c8491751590229a449e4076b6/src/commands.c#L1589
This is the command that processes it for sway
https://github.com/swaywm/sway/blob/4852087e6166062d33372bd2869fce4033232e08/sway/commands/move.c#L703

- **Description:**
When issuing `[floating=true] move left 50 ppt`. The window only moves 50 px left.

Sway should move the window left 50% of the output size.

ElementalWarrior:
To be clear, I only found this trying to resize based on percent of the window rect. Which neither sway or i3 seem to support :(

PR #8743, opened 2025-05-30
Add unit ppt to 'move' + direction

Absobel:
Literally the same pull request as #8596 that tries to solve #8566 but i don't know how to use github and couldn't find a way to just change the branch name. Anyway sorry ! (I also cleaned up the commits to use rebases instead of merges)


emersion (commented):
Sorry for the delay, here are a few comments.

PR #8748, opened 2025-06-05
server: fix socket path memory leak

zzxyb:
The socket path allocated with strdup() in server_init() was not being freed in server_fini().
Remove const qualifier and add proper cleanup.


emersion (approved):
Thanks!

PR #8757, opened 2025-06-11
build: set wrap_mode=nodownload in default options

emersion:
This can be surprising (e.g. in CI, this can download source code instead of using system libraries) and users can easily turn it back on if desired.


kennylevinsen (approved):
LGTM

PR #8761, opened 2025-06-13
Make xkb_layout reset xkb_file and vice versa

### LINKED ISSUE #8329: xkb_file should accept an empty value
I want to use the same default layout for all keyboards, but devices like yubikeys (or barcode scanners, or any devices that pretend to be keyboards) use the US-QWERTY layout. Because my keyboard layout (for real keyboards) is hugely customized, I use a xkb file. My config thus looks like this:

``` 
input 4176:1031:Yubico_YubiKey_OTP+FIDO+CCID { 
  xkb_layout us 
} 
input type:keyboard { 
  xkb_file ~/.config/sway/my_weird_layout.xkb 
} 
```

with this config, the yubikey uses the layout described in my_weird_layout.xkb instead of US-QWERTY. I guess this happens because xkb_file gets merged from `input type:keyboard` into `input 4176:1031:Yubico_YubiKey_OTP+FIDO+CCID`. 

A nice solution for that issue would be for xkb_file to accept an empty value:

``` 
input 4176:1031:Yubico_YubiKey_OTP+FIDO+CCID { 
  xkb_layout us
  xkb_file 
} 
input type:keyboard { 
  xkb_file ~/.config/sway/my_weird_layout.xkb 
} 
```

I'm aware there are other solutions with the actual config syntax (use a xkb file for us qwerty as well, or configure each keyboard individually), but they're not as nice :)

not-my-profile:
Users that generally use a custom keyboard layout via xkb_file likely want to reset it to `us` for other devices that pretend to be keyboards like security keys, barcode scanners or shortcut remotes. For example:

    input 10429:2395:UGTABLET_Artist_Pro_16_(Gen2) {
      xkb_layout us
    }
    input type:keyboard {
      xkb_file ~/.config/sway/my_weird_layout.xkb
    }

Such configuration however didn't work because the config merge logic merged both commands together and then the xkb_file took precedence.

Fixes #8329.

### DIFF
--- a/sway/config/input.c
+++ b/sway/config/input.c
@@ -117,10 +117,13 @@ void merge_input_config(struct input_config *dst, struct input_config *src) {
 		free(dst->xkb_file);
 		dst->xkb_file = src->xkb_file ? strdup(src->xkb_file) : NULL;
 		dst->xkb_file_is_set = dst->xkb_file != NULL;
+		dst->xkb_layout = NULL;
 	}
 	if (src->xkb_layout) {
 		free(dst->xkb_layout);
 		dst->xkb_layout = strdup(src->xkb_layout);
+		dst->xkb_file = NULL;
+		dst->xkb_file_is_set = false;
 	}
 	if (src->xkb_model) {
 		free(dst->xkb_model);


sway/config/input.c
```diff
@@ -117,10 +117,13 @@ void merge_input_config(struct input_config *dst, struct input_config *src) {
 		free(dst->xkb_file);
 		dst->xkb_file = src->xkb_file ? strdup(src->xkb_file) : NULL;
 		dst->xkb_file_is_set = dst->xkb_file != NULL;
+		dst->xkb_layout = NULL;
```
kennylevinsen, line 120: `xkb_file` overrides not just `xkb_layout`, but also `xkb_model`, `xkb_options`, `xkb_rules`, and `xkb_variant`.

kennylevinsen (commented):
Note that you should already be able to do:

```
input 10429:2395:UGTABLET_Artist_Pro_16_(Gen2) {
  xkb_file -
  xkb_layout us
}
input type:keyboard {
  xkb_file ~/.config/sway/my_weird_layout.xkb
}
```

This does seem to lack documentation though.

EDIT: Submitted early.

sway/config/input.c
not-my-profile: Yes but I would have assumed that these other options don't take effect if `xkb_file` is set and `xkb_layout` isn't. Looking at `sway_keyboard_compile_keymap` that does seem to be the case?

In any case I think it would be nice to log or display an error in case conflicting xkb_ options are specified in the same config block. But I'd defer that to another PR.

PR #8766, opened 2025-06-16
Stop using wlr_scene_buffer_send_frame_done()

kennylevinsen:
That function now takes the output as input. We don't always have the output at hand, so use the function operating on a wlr_scene_surface instead.

Taken from https://github.com/swaywm/sway/pull/8737 to fix accidental breakage from merging https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5078. That PR is still draft and has an open TODO, so let's merge this single commit first. Not needed if the other PR is merged.

### DIFF
--- a/sway/desktop/output.c
+++ b/sway/desktop/output.c
@@ -97,11 +97,11 @@ struct buffer_timer {
 };
 
 static int handle_buffer_timer(void *data) {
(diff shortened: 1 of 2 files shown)


kennylevinsen:
No longer necessary.

PR #8780, opened 2025-06-25
New swaymsg message type: get_cursor that returns cursor position

OXDBXKXO:
Add an IPC command to Sway that allows swaymsg to request information about the cursor:

*Cursor over container:*
```json
{
	"x": 1024,
	"y": 768,
	"surface_x": 24,
	"surface_y": 48,
	"node_type": "con",
	"window_title": "Terminal",
	"app_id": "termite",
	"workspace_name": "1",
	"pid": 25370
}
```

*Cursor over workspace:*
```json
{
	"x": 1024,
	"y": 768,
	"node_type": "workspace",
	"workspace_name": "1"
}
```

### DIFF
--- a/completions/fish/swaymsg.fish
+++ b/completions/fish/swaymsg.fish
@@ -21,5 +21,6 @@ complete -c swaymsg -s t -l type -fra 'get_binding_modes' --description "Gets a 
 complete -c swaymsg -s t -l type -fra 'get_binding_state' --description "Get JSON-encoded info about the current binding state."
 complete -c swaymsg -s t -l type -fra 'get_config' --description "Gets a JSON-encoded copy of the current configuration."
 complete -c swaymsg -s t -l type -fra 'get_seats' --description "Gets a JSON-encoded list of all seats, its properties and all assigned devices."
+complete -c swaymsg -s t -l type -fra 'get_cursor' --description "Get JSON-encoded information about the current cursor position and the node under it."
 complete -c swaymsg -s t -l type -fra 'send_tick' --description "Sends a tick event to all subscribed clients."
 complete -c swaymsg -s t -l type -fra 'subscribe' --description "Subscribe to a list of event types."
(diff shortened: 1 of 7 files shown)


emersion:
Could you describe your use-case for this?

OXDBXKXO:
Hi,
Some tools need to be able to retrieve the cursor position.

For example, [Kando](https://github.com/kando-menu/kando/issues/452#issuecomment-2156749389), for which I plan to add Sway support (got it working on my forked Sway).

emersion:
I don't think Kando needs access to the global cursor position via the Sway IPC. Kando can open a full-screen layer-shell surface, receive the `wl_pointer.enter` event to get the position relative to the surface, and show the menu there.

OXDBXKXO:
Kando is just an example of a tool that would benefit from swaymsg being able to return information about the cursor position and the element under it.
It enables broader scripting/automation use-cases that relies on this information.

OXDBXKXO:
🥺 👉 👈

jirassimok:
I think emersion's point is correct; while adding this to Sway's IPC would be convenient, it isn't the proper way to get this information. It would be better to use the wlr-layer-protocol (or some other Wayland protocol), which would also be more portable to other WMs.

OXDBXKXO:
A pull-request was accepted today on the Kando project, which adds support for Niri. The contributor used the full-screen layer-shell trick.

I tried to use his work for the Sway backend, but the result is unsatisfactory: the trick does indeed allow Kando to receive the cursor position, but as no pointer-enter event is sent at surface creation, the capture of the cursor position hangs until the mouse moves.
The contributor claims that this is also the case on Hyprland, but as it provides a method for retrieving the cursor position through hyprctl, this is not problematic.

But then again, this new IPC command would have other uses than simply Kando compatibility: it would enable conditional actions based on cursor position.

emersion:
> no pointer-enter event is sent at surface creation

That's a Sway bug.

smlx:
I think the sway bug was possibly fixed by https://github.com/swaywm/sway/pull/9203. Do you think you could test if the full-screen layer shell trick now works on Sway master?

PR #8784, opened 2025-06-27
output: Log output information when connected

kennylevinsen:
The only logging we currently provide about outputs is debug logging
from wlroots about available modes, tested output states and errors
applying it.

Be more informative by logging some output details at INFO level when
they pop up, including optional features like HDR and adaptive sync.
This makes such output details more obvious from a quick glance, both
for users and in case logs are shared for support.

### DIFF
--- a/include/sway/config.h
+++ b/include/sway/config.h
@@ -680,7 +680,7 @@ struct seat_config *store_seat_config(struct seat_config *seat);
 int output_name_cmp(const void *item, const void *data);
 
 void output_get_identifier(char *identifier, size_t len,
-	struct sway_output *output);
+	struct wlr_output *output);
 
 const char *sway_output_scale_filter_to_string(enum scale_filter_mode scale_filter);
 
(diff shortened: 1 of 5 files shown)


emersion:
I don't think this is a good place to surface back output capabilities to users. `swaymsg -t get_outputs` is a better place.

This PR also duplicates some information with wlroots logs.

kennylevinsen:
> I don't think this is a good place to surface back output capabilities to users.

You can say that with most info logging though, including wlroots' own logging on the same topic of output details (detected modes, physical size).

> This PR also duplicates...

Physical size, yeah. I had missed that the wlroots logging was WLR_INFO, recalled it as being WLR_DEBUG.

Hmm, we could just add a few extra lines to the wlroots logging when discovering the connector... Then it's in one place for all wlroots compositors.

emersion:
Related: https://github.com/swaywm/sway/pull/8795

PR #8791, opened 2025-07-02
Integrate libsfdo (take 2)

myrslint:
This follows up on #8689.

Thanks to @emersion's advice [here](https://github.com/swaywm/sway/pull/8689#issuecomment-3025698388) I managed to salvage the work and rebase it, hopefully correctly this time.

In addition to #8689's objectives this makes libsfdo a hard dependency as requested [here](https://github.com/swaywm/sway/pull/8689#issuecomment-3009773073), dispenses with conditionally compiling based on availability of libsfdo, and removes the old code for locating and loading icons.

### DIFF
--- a/common/meson.build
+++ b/common/meson.build
@@ -8,6 +8,7 @@ lib_sway_common = static_library(
 		'loop.c',
 		'list.c',
 		'pango.c',
+		'sfdo.c',
 		'stringop.c',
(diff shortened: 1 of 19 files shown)


myrslint:
I think libsfdo needs to be added to the build containers now for build tests to succeed.

PR #8795, opened 2025-07-05
ipc-json, swaymsg: indicate when adaptive sync is unsupported

emersion:
A new "features" object is added to the output description. We'll add more entries in the future, e.g. HDR.


kennylevinsen (approved):
Seems good.

PR #8849, opened 2025-08-22
input/cursor: fix selection of all contaners when clicking in popup-menu and on the lockscreen

kkvark:
After mouse-down on an item in the popup-menus (for example, waybar tray-menus) all containers in the current workspace become selected. The same thing happens when clicking in the lockscreen.
I made this patch for myself to fix this behavior. Perhaps there is a better solution to this problem.

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -102,6 +102,10 @@ struct sway_node *node_at_coords(
 				return NULL;
 			}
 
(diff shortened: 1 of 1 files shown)


smlx:
I think this is the same bug that #9160 fixes? Could you check if that PR fixes your issue?

kkvark:
> I think this is the same bug that #9160 fixes? Could you check if that PR fixes your issue?

Yep you patch fix my issue. Then I'll close my request.

PR #8861, opened 2025-09-09
readme: mark display managers as supported

emersion:
It's been a long while since we've heard about issues related to display managers. Some used to be unreliable in the past, but by now most issues have been ironed out. Let's start supporting them officially.

PR #8869, opened 2025-09-11
readme: Update and reword Czech translation

Spiffyk:
Updated the Czech README to reflect changes in #8861 and also did some general rewording for a more natural flow of the language.

### DIFF
--- a/README.cs.md
+++ b/README.cs.md
@@ -1,23 +1,24 @@
 # sway
 
-sway je s [i3] kompatibilní [Wayland] kompozitor. Přečtěte si [FAQ]. Připojte se na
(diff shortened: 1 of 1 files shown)


Spiffyk:
I'm on the fence about how to translate "display manager" properly, to be honest. I'm not sure if there is a common translation for this, I'd personally just use the English "display manager" _in a conversation_ and be done with it, but it's not really correct to do that in _written_ Czech.

emersion:
Thanks for the PR!

@AsciiWolf @jkonecny12 @martymichal Would you have some time to review this patch?

AsciiWolf:
The rest looks good. Thanks! :)

PR #8870, opened 2025-09-11
Update Norwegian README to make it more in line with the English one, as well as improving grammar

mortie:
This PR does these 3 things, split into 3 commits:

* Mark display managers as supported in the Norwegian README, to reflect https://github.com/swaywm/sway/pull/8861
* Update the Norwegian translation to account for some other changes (see the message in https://github.com/swaywm/sway/commit/0bdf4f4312c6c5d63f67fba40c8601668ae4d3a8 for details)
* Improve the Norwegian translation's grammar. A common error in Norwegian text is to take words which should be written as one (for example "systempakker", meaning "system packages") and write them as two ("system pakker"), so I fixed some instances of that. I also did some other changes which make the text read more natural to me.

Feel free to take or leave individual commits from this PR.


emersion:
Thanks for the PR!

@Radiicall @JustTemmie Would you have some time to review this?

PR #8878, opened 2025-09-13
input/text_input: fix crash when using im

sahinf:
Problem: wlroots api changed to pass NULL for the callback and stores the input_method in the relay. sway tries to dereference the null ptr.

Solution: get im from relay, ignore `*data`


emersion:
This sounds like a subset of https://github.com/swaywm/sway/pull/8867?

sahinf:
oops, didn't check.

PR #8880, opened 2025-09-14
Fixes session lock surfaces not being able to switch focused output

### LINKED ISSUE #7790: SessionLock cannot switch focused surface

- **Sway Version:**
  - `sway version 1.9-dev-4326a26a (Oct 23 2023, branch 'master')`

- **Configuration File:**
https://github.com/waycrate/waycratelock

I made this program , but I found that the textInputFeild can only get focus when it is the first screen, and I figure the code is in
https://github.com/swaywm/sway/blob/master/sway/lock.c#L33, this means it can only focus on the first screen appeared, cannot focus on other screen. 

And the program is ok on river, which also impl ext-session-lock. so I think it is a bug, but I do not know how to fix it

- **Description:**
  - The steps you took in plain English to reproduce the problem.

https://github.com/swaywm/sway/assets/60290287/579bdba6-fb63-44fc-b578-112bf4e7f3b5

You can see that in 4 s, the text-input-area is not editable.

ErikReider:
Fixes: https://github.com/swaywm/sway/issues/7790

To reproduce the issue:
1. Start a nested sway session with two virtual outputs (`swaymsg create_output WL-2`)
2. Lock the session, preferably with a locker that shows if it receives focus like [gtklock](https://github.com/jovanlanik/gtklock) or [my locker](https://github.com/ErikReider/SwaySettings/pull/56)
3. Try switching focus to the non-focused input field on the other monitor 

It would also be nice if this could be backported to v1.11


ErikReider:
With the current code, when a seat wants to switch focus output, it forces the new surface to be the focus surface for every seat (in the `focus_surface` function). Let me know if that's not desired, and I'll fix it :)

ErikReider:
@emersion sorry for the ping, but is there anything stopping this from getting merged? :)

Decodetalkers:
Oh.. Thanks , there is a pr for it! @emersion can you help review it? I want it be merged before next release

Decodetalkers (approved):
Oh. I see, it can be fixed with such simple way. This solution is perfect. Hope it can be merged soon.

sway/input/seat.c
```diff
@@ -1283,6 +1283,24 @@ void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
...
+			case N_OUTPUT:
+				output = node->sway_output;
+				break;
+			case N_WORKSPACE:
+				output = node->sway_workspace->output;
+				break;
+			default:
+				break;
+			}
+		}
```
emersion: I'm confused about this piece of code. Lock surfaces are not `sway_node`s. When is this codepath taken?

sway/input/seat.c
```diff
@@ -1283,6 +1283,24 @@ void seat_set_focus(struct sway_seat *seat, struct sway_node *node) {
...
+		if (output != NULL) {
+			sway_session_lock_focus_output(server.session_lock.lock, output);
```
emersion: Shouldn't we only change focus for the seat passed as an argument, as opposed to all seats?

sway/lock.c
```diff
@@ -59,9 +60,34 @@ static void refocus_output(struct sway_session_lock_output *output) {
...
-	if (surf->lock->focused == NULL) {
+
+	struct sway_seat *seat = input_manager_current_seat();
+	struct sway_workspace *focused_ws = seat_get_focused_workspace(seat);
+	struct sway_output *focused_output = NULL;
+	if (focused_ws != NULL) {
+		focused_output = focused_ws->output;
+	}
+
+	if (surf->lock->focused == NULL || focused_output == surf->output) {
```
emersion: This will potentially result in initial focus being set on the wrong surface, then fixed later on. Can we avoid this?

sway/input/seat.c
Decodetalkers: I think this part is not related to lock surface. This part only check if in current time, these is a lock, if there is a lock, then it will focus the lock surface by the output that the sway_node is on. And there is node type of "N_OUTPUT" and "N_WORKSPACE" in it, I think this logic is no problem..

ErikReider: Okay so I double-checked now and `seat_set_focus` is called with the newly focused output's active workspace as the `sway_node`.

Just like @Decodetalkers mentioned above, it gets the (in this case) active workspace's output and tries to focus its lock-surface. I'll push a new commit simplifying this :)

sway/input/seat.c
ErikReider: In that case, the `focused` field in `struct sway_session_lock` (https://github.com/ErikReider/sway/blob/528b7af34fc906a097b339f99b5cc3e54dfc4bdb/include/sway/server.h#L16) should be removed and replaced with a per-seat `struct wlr_surface *focused_lock_surface` just like with layer-shell surfaces.

ErikReider (commented):
Let me know if you want me to squash the commits :)

include/sway/server.h
```diff
@@ -181,6 +180,8 @@ void sway_session_lock_add_output(struct sway_session_lock *lock,
 	struct sway_output *output);
 bool sway_session_lock_has_surface(struct sway_session_lock *lock,
 	struct wlr_surface *surface);
+void sway_session_lock_focus_output(struct sway_session_lock *lock,
+	struct sway_seat* seat, struct sway_output *output);
```
emersion: Style nit: `*` should be on the right

sway/lock.c
```diff
@@ -29,40 +30,68 @@ struct sway_session_lock_output {
 };
 
 static void focus_surface(struct sway_session_lock *lock,
-		struct wlr_surface *focused) {
-	lock->focused = focused;
-
-	struct sway_seat *seat;
-	wl_list_for_each(seat, &server.input->seats, link) {
+		struct sway_seat *seat, struct wlr_surface *focused) {
+	if (seat) {
```
emersion: Can we move the NULL check to the callers?

In general, we try to avoid functions which no-op when given NULL. It can be confusing to call a function named `focus_surface()` and have it not change any focus.

sway/lock.c
emersion, line 101: The lock surface can be destroyed before the session is unlocked. We need to reset `seat->focused_lock` in that case I think, to avoid leaving a dangling pointer behind?

It might be better to introduce a `seat_set_focus_lock()` function in `seat.c`, and install a destroy handler there…

sway/lock.c
```diff
@@ -29,40 +30,68 @@ struct sway_session_lock_output {
...
+	struct sway_seat *seat;
+	wl_list_for_each(seat, &server.input->seats, link) {
+		if (seat->focused_lock == output->surface->surface) {
+			focus_surface(output->lock, seat, next_focus);
+		}
+	}
+}
+
+void sway_session_lock_focus_output(struct sway_session_lock *lock,
+	struct sway_seat* seat, struct sway_output *output) {
```
emersion: Style nit: missing an extra indent here

include/sway/input/seat.h
```diff
@@ -99,6 +99,7 @@ struct sway_seat {
 	char *prev_workspace_name; // for workspace back_and_forth
 
 	struct wlr_layer_surface_v1 *focused_layer;
+	struct wlr_surface *focused_lock;
```
emersion: I wonder if we really need this field, since we already have `wlr_seat->keyboard_state.focused_surface`…

ErikReider: IMO it's good for cases where "refocusing" the locked surface is needed like in `seat_set_focus`. But I agree, other than the minor edge-cases, it shouldn't be needed right? Any thoughts? :)

ErikReider:
Everything should be up to snuff now. Sorry for the late response :)

Issue #8884, opened 2025-09-15
The IME popup in Chromium using text-input-v3 first appears in the top-left corner.


AbdAlRahmanGad:
- **Sway Version:**
  - sway version 1.10.1
- **Debug Log:**
  - Sway logs: https://gist.github.com/AbdAlRahmanGad/f5936c9cf0746b46d15e83781af450c8#file-sway-md
  - Chromium logs: https://gist.github.com/AbdAlRahmanGad/f5936c9cf0746b46d15e83781af450c8#file-chromium_log-md

- **Configuration File:**
  - Default
- **Description:**
  - Run `fcitx5 -d` and ensure you have a language like Chinese or Japanese
  - Run a snapshot build of chromium
  - When switching languages the popup will show in the top left corner
  - Type some text and the popup will start showing in the right place after that.
  - We are tracking the issue here: https://issues.chromium.org/issues/384531043. Please take a look at the upstream issue.
  - I attached a demo reproducing the issue.

- **Cause of the issue**
  - This happens because we expect a `done` event after `enable`, as we are trying to follow the protocol by not committing `set_cursor_rectangle` and similar requests before receiving a `done` event with a matching serial, but `sway` doesn't send one.
  - Please have a look at Orko's comment here for a better explanation: https://issues.chromium.org/u/1/issues/384531043#comment6

 
  https://github.com/user-attachments/assets/27dde334-b155-4973-b0a8-369e63553dcb

PR #8892, opened 2025-09-20
input/text_input: Fix crash on im destroy

sahinf:
Problem: Exiting input manager like fcitx5 causes sway to segfault

Solution: add null check.


tokyo4j:
FWIW, `wlr_input_method_keyboard_grab_v2` is still passed as `data` in `handle_im_keyboard_grab_destroy()`, because it keeps things simple in labwc and sway (ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5107#note_3042631).

Also, I'm not sure about the second commit. It's just a workaround to fix the specific problem in Chromium and not required by the protocol at all. I suggested a fix in chromium side: https://issues.chromium.org/issues/384531043#comment7.

sahinf:
The issue is not specific to Chromium, I see it in firefox for example, suggesting others may be affected too. Do we have any issues with sending done on enable?

Qutebrowser sets the position correctly, but I'm not sure if it's because qt apparently violates the protocol or because it calls https://wayland.app/protocols/text-input-unstable-v3#zwp_text_input_v3:request:set_cursor_rectangle correctly while others do not.

tokyo4j:
> The issue is not specific to Chromium, I see it in firefox for example, suggesting others may be affected too. Do we have any issues with sending done on enable?

Practically, no. But I'm not very happy about doing something not documented in the protocol as sway is often used as a reference of wlroots-based compositors. Also, sending `done` only after `enable` + `commit`, but not after usual `commit` looks a bit odd and specific to #8884.

But yeah, I see its practical benefit and this may be the best solution that doesn't complicate the compositor implementation. Let's see how sway maintainers react to this.

sahinf:
[input-method-v2: Destroy keyboard grab before input method (!5170) · Merge requests · wlroots / wlroots · GitLab](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5170)

PR #8894, opened 2025-09-21
sway/server: set color_manager for root scene

llyyr:
Chase https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5122/


include/sway/server.h
```diff
@@ -90,6 +90,8 @@ struct sway_server {
 	struct wlr_gamma_control_manager_v1 *gamma_control_manager_v1;
 	struct wl_listener gamma_control_set_gamma;
 
+	struct wlr_color_manager_v1 *color_manager_v1;
```
emersion, line 93: Nit: no need for this field, we can keep this as a local variable instead.

emersion (approved):

kode54:
Does this need an error check for the return value being NULL? Since I'm staging a version of it for my patches against labwc.

emersion:
Sway does not check for allocation errors at startup, so not adding the check keeps things consistent (see the rest of the global initialization). But yeah, the function can return NULL on allocation error, and other compositors might want to check for that.

emersion:
Thanks!

PR #8902, opened 2025-09-29
sway/input/cursor: send frame event when simulating pointer from tablet

llyyr:


emersion (commented):
Aren't we already sending the frame event from `handle_touch_frame()`?

llyyr:
> `handle_touch_frame`

That's for touch devices, this is for drawing tabs. In any case, I verified with WAYLAND_DEBUG=1 that we don't send the frame event when simulating pointer with tablet

emersion (approved):
Ah yes, good catch!

PR #8911, opened 2025-10-02
build: switch to explicit 'meson setup' syntax

nyxed:
Fixes: 
```
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```


emersion (approved):
Thanks!

PR #8698, opened 2025-05-04
Tray D-Bus Menu

rogueai:
This is just a rebase, along the lines of existing PRs (#6249, #8405)

I rebased #8405, then cherry-picked a PR https://github.com/grazzolini/sway/pull/1 from @grazzolini's own fork.

I tested locally and it seems to be working fine, at least with nm-applet and keepassxc.

Disclaimer: this is just a rebase, I know nothing about C so I won't be able to progress much further than this unfortunately.

PR #8405, opened 2024-10-24
Tray D-Bus Menu

grazzolini:
This PR takes all the patches from #6249 and I just did some code cleanup but other than that, did not touch any of the code. Also, incorporated some patches.

Issue #3799, opened 2019-02-28
Planned swaybar tray features


ianyfan:
[ ]  XEmbed proxy
[ ] Canonical DBus menu (currently blocked by layer shell popups)
[ ] caching icons correctly: https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#implementation_notes (ref https://github.com/swaywm/sway/issues/3789)
[ ] Locales when parsing index.theme file

Currently not being worked on (by me, at least) so contributions would be very welcome :)

artizirk:
KDE guys have a XEmbed proxy daemon https://github.com/KDE/plasma-workspace/tree/master/xembed-sni-proxy

m-khvoinitsky:
Consider adding: SVG icons support.
Currently, for example, Quassel shows sad face instead of icon because its icon is SVG.

ianyfan:
Hmm, I did allow it to load SVGs if it was built with gdk-pixbuf2 support, but I never actually tested it, so it might have gone wrong somewhere.

m-khvoinitsky:
I'm not 100% sure that the reason is SVG but that's the only difference that I've noticed between Quassel and working icons. And there is no any hint in logs.

emersion:
>However, it is currently blocked by layer shell popups

This can now be implemented, as layer-shell popups are supported by Sway.

agraven:
Should an issue be created for the implementing the DBus menu specification?

m-khvoinitsky:
> Consider adding: SVG icons support.
Currently, for example, Quassel shows sad face instead of icon because its icon is SVG.

It works since 1.4.

emersion:
PR implementing D-Bus menus: https://github.com/swaywm/sway/pull/5161

nmschulte:
I'm not sure what is spec'd, but is it possible to have tray icon tooltips, that helps me identify the tray item, some sort of name or title?

ianyfan:
Hm, I guess you can do it now that popups have been implemented, and there is a tooltip property in the SNI spec, but it's not something that I'm planning on doing, so someone else will have to implement it

Merith-TK:
how do i enable this? running archlinux on sway 1.4. but the only tray icon i have ever seen on my system through swaybar was albert, which didnt work due to the whole isolation that wayland does

LambLeeg:
[...]

PR #8920, opened 2025-10-13
Tray D-Bus Menu (fifth time the charm)

blinxen:
I recently switched from i3 to sway and found out about the missing dbus menu support in swaybar ([issue](https://github.com/swaywm/sway/issues/3799)).

From what I understood, the initial implementation was done in https://github.com/swaywm/sway/pull/5161 then continued in https://github.com/swaywm/sway/pull/6249 then continued in https://github.com/swaywm/sway/pull/8405 then continued in https://github.com/swaywm/sway/pull/8698 :thinking:. However, it seems that the last two PRs (https://github.com/swaywm/sway/pull/8405 and https://github.com/swaywm/sway/pull/8698) are just people trying to rebase the PR to master, with some minor changes suggested in comments. I would like this to get merged so I decided to take ownership of this change and see it through the end.

In addition to the previous state, the following changes were added:

* Fix a bug where swaybar would crash when the dbus menu is open but the underlying process (e.g. nm-applet)  was closed.
* Add recursion guard (https://github.com/swaywm/sway/pull/8405#discussion_r1815659380)
* Rename some global functions to adhere to the [style guide](https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md#style-reference)

I have not reviewed *all changes* that are included in this PR since I saw that a lot of people already put a lot of work into that.

Let me know if I can do anything to speed up the review process or if I should adjust / fix any issues with this PR.

### DIFF
--- a/include/swaybar/bar.h
+++ b/include/swaybar/bar.h
@@ -33,6 +33,7 @@ struct swaybar {
 	struct zxdg_output_manager_v1 *xdg_output_manager;
 	struct wp_cursor_shape_manager_v1 *cursor_shape_manager;
 	struct wl_shm *shm;
+	struct xdg_wm_base *wm_base;
 
 	struct swaybar_config *config;
 	struct status_line *status;
--- a/include/swaybar/input.h
+++ b/include/swaybar/input.h
@@ -15,6 +15,7 @@
 
 struct swaybar;
 struct swaybar_output;
+struct swaybar_seat;
 
 struct swaybar_pointer {
 	struct wl_pointer *pointer;
@@ -48,8 +49,8 @@ struct swaybar_hotspot {
 	struct wl_list link; // swaybar_output::hotspots
 	int x, y, width, height;
 	enum hotspot_event_handling (*callback)(struct swaybar_output *output,
-		struct swaybar_hotspot *hotspot, double x, double y, uint32_t button,
-		bool released, void *data);
+			struct swaybar_hotspot *hotspot, struct swaybar_seat *seat, uint32_t serial,
+			double x, double y, uint32_t button, bool released, void *data);
 	void (*destroy)(void *data);
 	void *data;
 };
--- a/include/swaybar/tray/dbusmenu.h
+++ b/include/swaybar/tray/dbusmenu.h
@@ -1,0 +1,27 @@
+#ifndef _SWAYBAR_TRAY_DBUSMENU_H
+#define _SWAYBAR_TRAY_DBUSMENU_H
+
+#include "swaybar/bar.h"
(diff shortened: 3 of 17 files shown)


blinxen:
One thing which I am unsure about is the namings of the functions in `dbusmenu.c`. A lot of them have `sway_` or `dbusmenu_` prefixes even though they are `static`.  I am inclined to remove these prefixes since the functions are just local. Any thoughts?

Thiago-Assis-T:
i would really like this in sway...

aurum77:
i NEED this in sway

radovim:
When can we expect this feature to be merged?

CyberShadow:
Thank you for working on this!

I've given this a try. Ran into some problems:
- segmentation faults when a menu item icon fails to load
- menu was extremely laggy because it kept trying to load the icon on every mouse movement
- also noticed that memory was freed incorrectly in one place
- SVG icons (which seem to be very common) don't work for me, but I think this is a problem with my environment

Here are patches for the first three:

[0001-fix-swaybar-dbusmenu-fix-null-pointer-dereference-wh.patch](https://github.com/user-attachments/files/24542856/0001-fix-swaybar-dbusmenu-fix-null-pointer-dereference-wh.patch)
[0002-fix-swaybar-dbusmenu-properly-destroy-cairo-surface-.patch](https://github.com/user-attachments/files/24542858/0002-fix-swaybar-dbusmenu-properly-destroy-cairo-surface-.patch)
[0003-fix-swaybar-dbusmenu-cache-icon-lookups-to-avoid-rep.patch](https://github.com/user-attachments/files/24542859/0003-fix-swaybar-dbusmenu-cache-icon-lookups-to-avoid-rep.patch)

blinxen:
I am still interested in this getting merged.

tatsumoto-ren:
@emersion @kennylevinsen could you take a look at this? I think this is one of the last remaining blockers for migrating from Xorg to Wayland. As far as I know, no other bar supports i3blocks, so switching to another bar like waybar, for example, is not an option. Plus, that would be an extra dependency to install.

lasombra:
This would be a great addition to sway 1.13

syntaxtea:
Could we please get a review on this? I recently switched from i3 to Sway, and the experience has been mostly seamless except for this rough edge. It's rather difficult to use apps that rely on tray icon functionality if I'm unable to right click. FWIW, single- and double-click work fine, right-click and hover do not.

amano-kenji:
Perhaps, sway developers hate dbus.

emersion:
This PR is quite large and I have other things on my plate. I'd like to review it but not sure when I'll have the chance.

PR #8921, opened 2025-10-14
sway/commands/layout: fix flatten parent once

bonsaiiV:
only flatten if grandparent also has only a single child

this is also the behavior in i3

i missed this check, when i first implemented it


emersion:
This seems to be consistent with i3: https://github.com/i3/i3/commit/f8befe378ab3c9f8d035657ca0247c9b8c242b6b

cyanreg:
Can confirm this fixes the issue.
Could you merge it?

emersion (approved):
Thank you all!

Issue #8918, opened 2025-10-11
Game running in wine wayland has weird gamma curve in current sway git


v44r:
- sway version 1.12-dev-90d32709 (Oct 10 2025, branch 'master')
- [swaylog](https://gist.github.com/v44r/bacb0ef0de00ddaaa27e1351e3a79aad#file-swaylog-txt)
- stock config

When I run a game under wine wayland (DISPLAY= wine game.exe) the gamma curve seems wrong. Gamma slider goes from too dark to too bright in just a couple of steps, and dark areas in videos are full of blocky artifacts independent of the gamma setting I choose (unless I black crush everything):

wayland gamma settings: https://streamable.com/oicoij (notice I just move the slider two or three pixels and it goes from too dark to too bright)

Under Xwayland (wine game.exe) the game looks perfect. 

xwayland gamma settings: https://streamable.com/2ouq1q  (much more gradual gamma adjustment)

This is with HDR disabled (DXVK_HDR=0).

This does not happen in sway 1.11 + wlroots 0.19. In 1.11+0.19 the game looks identical under wayland or xwayland and the gamma curve looks the same as in the above 1.12 xwayland video. In other compositors (tested in Hyprland) they also look the same in wayland and xwayland.

I've also attached the wine logs in the above swaylog gist, but a diff does not show anything relevant.

poisotf:
That would be wlroots@d8fb7adc (https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5141)

dxvk translating `DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709` to `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` [(here)](https://github.com/doitsujin/dxvk/blob/1c25c0a517d86e2bdea82730589a38eab17451c4/src/d3d11/d3d11_swapchain.cpp#L19) now causes a gamma correction that is arguably not wanted here

v44r:
Yes, I compiled wlroots@6978509f and Diablo 2 Resurrected (shown in the above video) looks fine, like the Xwayland version. The change introduced by the patch makes D2R unconfigurable (no gamma setting in the game looks right).

PR #8922, opened 2025-10-16
sway/server: stop advertising "srgb" color transfer function

poisotf:
In anticipation of https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/442

A brutish approach, clients that require or assume `wp_color_manager_v1::transfer_function::srgb` to always be available will die.

On recent mesa and wine, this gives the desired outcome of vulkan `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` using no color management messages. Closes: #8918

If this causes issues, a safer alternative would be a change in wlroots to quietly reinterpret "srgb" as an alias for "gamma22"

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -459,7 +459,6 @@ bool server_init(struct sway_server *server) {
 			WP_COLOR_MANAGER_V1_RENDER_INTENT_PERCEPTUAL,
 		};
 		const enum wp_color_manager_v1_transfer_function transfer_functions[] = {
-			WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_SRGB,
 			WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_ST2084_PQ,
 			WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_EXT_LINEAR,
 			WP_COLOR_MANAGER_V1_TRANSFER_FUNCTION_GAMMA22,


llyyr:
We should advertise both, there are high end displays out there which use the piecewise function and ITU-R BT.2380 has explicitly defined the piecewise function as the sRGB eotf.

We need to make this configurable instead and propagate the mess one level up to the end user, the MR you linked also doesn't remove the piecewise function anymore, just renames it instead.

emersion:
There are two issues with the color-management protocol `srgb` TF:

- It's ill-defined: some clients use it to mean "whatever sRGB content expects", which is a gamma 2.2 TF. See: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/442
- Historically monitors have been using gamma 2.2, but some monitors are using piece-wise sRGB now. We already have a  `output <name> color_profile srgb|icc` command, it would seem natural to add a new `gamma22` entry (used by default), and reuse or rename `srgb` for piece-wise sRGB monitors. (It would also be nice to add `edid` as an additional profile, to use primaries from the EDID.)

poisotf:
(I realize I've named this MR terribly : it changes the options passed by sway to wlroots, not sway's user-facing options)

emersion:
A new version of the protocol has been published with `srgb` replaced with a new variant. Ref https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/4032

Since fa81ce8ee64d690b881d97b734583e26cb2acb72 we now use wlroots helpers to figure out the supported TFs so this would need to be moved to wlroots.

PR #8927, opened 2025-10-23
xdg-shell: dont overwrite for_window scratchpad geometry

### LINKED ISSUE #8018: 'for_window resize set' not (or differently) working in sway 1.9
- **Sway Version:**
sway version 1.9

- **Debug Log:**
https://gist.github.com/528-491/65916752c34178f1bf2123cb473edf68

- **Configuration File:**
`swayminconf `:
```
exec alacritty --class popup_term --option window.opacity=0.9
for_window [app_id="^popup_term$"] floating enable, resize set 80ppt 70ppt, move position 10ppt -6px, mark popup_term, move scratchpad, sticky enable
bindsym Mod4+Return [con_mark="^popup_term$"] scratchpad show
```

- **Description:**
  - Start sway from tty with `sway -c swayminconf -d 2> ~/sway.log`
  - Press `Mod4+Return` to show pop-up terminal
  - Observe that the terminal window wasn't resized to 80ppt 70ppt

### LINKED ISSUE #8493: `for_window` `resize` fails on container moved to scratchpad without `scratchpad show`
- **Sway Version:**
  - sway version 1.10-dev-1d783794 (Dec  6 2024, branch 'master')

- **Debug Log:**
  - https://gist.githubusercontent.com/sahinf/d17049cefc7dbe95398ab23533d13126/raw/77d05ef2ce5a0f8e56dd4819bfedd2a77ab0cd6e/sway.log

- **Configuration File:**
  - https://gist.githubusercontent.com/sahinf/d17049cefc7dbe95398ab23533d13126/raw/77d05ef2ce5a0f8e56dd4819bfedd2a77ab0cd6e/config.in

- **Description:**

1. I followed man sway(5) `for_window` to resize a container and move it to scratchpad. 
```swayconfig
exec $term --app-id "float"
for_window [app_id="float"] floating yes, resize set width 1600

exec $term --app-id "scratch"
for_window [app_id="scratch"] resize set width 1600, move scratchpad
```
2. I ran `scratchpad show`

Then I saw that the size of the scratchpad terminal was not applied. 

I added a `scratchpad_show` in the `for_window` like
```swayconfig
exec $term --app-id "scratch"
for_window [app_id="scratch"] resize set width 1600, move scratchpad, scratchpad show
``` 
which resized the container as I wanted. However, I don't want that container visible from the start so it's not really a fix for me. 

I added a breakpoint right before returning in `resize_set_floating` and saw both the `container->current` and `container->pending` width/height values are set to the expected values in both instances. I also broke on `xdg_shell:handle_commit` and it looks like the the correct view->container->current width/height values do not propagate to the view->geometry.

### LINKED ISSUE #8711: `for_window` does not always execute `floating enable` even if criteria matches
When I popout the bitwarden extension view in firefox into its own window, that window does not become floating even though I have `for_window` with appropriate criteria to match the window. The bitwarden window only becomes floating once I change the focused window or change to a different workspace. The strange thing is that I have a similar config for gnome calculator but when I launch the calculator it always opens as a floating window. This seems to be the same as one mentioned in [this comment](https://github.com/swaywm/sway/issues/6137#issuecomment-2689939719) on different issue back in february. I'm opening this issue to raise the visibility of this issue since there was no reply to that comment

- **Sway Version:**
  - sway version 1.10.1

- **Debug Log:**
  - [logs](https://gist.github.com/skoved/cf79005fef933978654c7cf20369387c)

- **Configuration File:**
  - [minimal config](https://gist.github.com/skoved/58fa5cc8e8d84ea6ccfb5740ad41fca4) you can ignore the firefox crash at the end. It is not related to this issue

- **Description:**
  - add the bitwarden extension to firefox
  - run sway with the minimal config linked above
  - super+y to open firefox
  - click the extension button in the firefox toolbar and click on bitwarden
  - in the bitwarden popup click the popout window button
  - notice that the bitwarden window is focused and open in a tiled window **NOT** a floating window
  - change the focus back to the firefox window. **Now** the bitwarden window becomes floating

### LINKED ISSUE #8749: for_window rules are not executed immediately / at all

- **Sway Version:**
1.10.1

- **Debug Log:**
  [Link](https://gist.github.com/masscream/ea2346707fca8575eed95c3eb472c41d)

- **Configuration File:**
  - [Link](https://gist.github.com/masscream/817d4bd3285bb5fce651c78d007cf8e8)

- **Description:**
Hi, I'm trying to use Double Commander filemanager on sway (does not matter if gtk/xwayland or qt/wayland). As a common filemgr it usually opens additional progress window during Copy/Move/Delete/etc. operations. These windows are normally tiled in sway, which is not very handy. In i3 window manager, I can make them floating easily by
`for_window [window_role="TfrmFileOp"] floating enable`
without any problem. Since wayland does not support this kind of filtering, I need to filter these windows by a title, where the issue starts. I put this into my config
`for_window [title="(?:Copying|Moving|Deleting|Packing|Unpacking)"] floating enable`
but unfortunately, it does not have immediate effect on the appeared window. They start tiled, and they only detach if I move my mouse away from that window (changing windows) or switch to another/same workspace. It looks like, sway needs nagging to realise, that the window it has to execute rules for is actually running. If I do nothing during the whole operation, the window does not detach and just disappears in the end. What makes it even worse is, that sometimes the operation throws another question window (Really delete?, Follow symlinks? etc.) and usually such window appears right before the main operation window detaches from tiling, making it almost unreachable behind it. This is making sway very hard to use with such a handy filemgr as the DC is. Also, I tried to put rules as focusing on the window, switching to the same workspace, resizing or repositioning window but nothing helps. Sometimes even these rules got ignored - I found several threads that sway may have some issues with it. I also know that it may not be easy to fix it at the moment, but isn't there any workaround I could use in the meantime (except dropping DC). Thank you.

### LINKED ISSUE #9127: for_window rules with floating enable do not correctly update window geometry
- **Sway Version:** 1.11

- **Debug Log:**
[sway.1777042854.log](https://github.com/user-attachments/files/27057026/sway.1777042854.log)

- **Configuration File:**
[config.txt](https://github.com/user-attachments/files/27057037/config.txt)

- **Stack Trace:**
N/A

- **Description:**
- add `for_window [title="^pulsemixer$"] mark volume, floating enable` to my config
- open any containers
- open terminal and run pulsemixer
- containers retain a size as if the terminal had not been made floating
Image

sahinf:
fixes #8493, fixes #8018, fixes #8711, fixes #8749, fixes #9127

### DIFF
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -321,7 +321,8 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		// containers, we resize the container to match. For tiling containers,
 		// we only recenter the surface.
 		memcpy(&view->geometry, new_geo, sizeof(struct wlr_box));
-		if (container_is_floating(view->container)) {
+		bool pending_configures = xdg_surface->current.configure_serial < xdg_surface->scheduled_serial;
+		if (container_is_floating(view->container) && !view->container->node.instruction && !pending_configures) {
 			view_update_size(view);
 			// Only set the toplevel size the current container actually has a size.
 			if (view->container->current.width) {
--- a/sway/desktop/xdg_shell.c
+++ b/sway/desktop/xdg_shell.c
@@ -321,7 +321,8 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		// containers, we resize the container to match. For tiling containers,
 		// we only recenter the surface.
 		memcpy(&view->geometry, new_geo, sizeof(struct wlr_box));
-		if (container_is_floating(view->container)) {
+		bool pending_configures = xdg_surface->current.configure_serial != xdg_surface->scheduled_serial;
+		if (container_is_floating(view->container) && !view->container->node.instruction && !pending_configures) {
 			view_update_size(view);
 			// Only set the toplevel size the current container actually has a size.
 			if (view->container->current.width) {


GordianDziwis:
The [bash script](https://github.com/swaywm/sway/issues/8493#issuecomment-3426024977) works now without the second wait.

Maybe the issue runs deeper because `for_window [floating] resize set 1200 800` does not work. Resizing on `float` events with [swayipc-asyncx](https://crates.io/crates/swayipc-async) also does not work.

GordianDziwis:
I tested [4aaff7e2ef61b523629f80abfed0d784057c6411](https://github.com/swaywm/sway/pull/8927/commits/4aaff7e2ef61b523629f80abfed0d784057c6411) with `for_window [floating] opacity 0.5` and now all new (terminal) windows are matched and transparent.

sahinf:
Non-terminal windows should also get matched. lmk if that's not the case (it is for me).

sway/criteria.c
```diff
@@ -419,12 +419,6 @@ static bool criteria_matches_view(struct criteria *criteria,
 	}
 #endif
 
-	if (criteria->floating) {
-		if (!container_is_floating(view->container)) {
-			return false;
-		}
-	}
```
emersion, line 426: I don't understand this change. Doesn't this result in `criteria->floating` being always ignored and floating criteria matching tiled containers?

sway/desktop/xdg_shell.c
```diff
@@ -321,7 +321,7 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		// containers, we resize the container to match. For tiling containers,
 		// we only recenter the surface.
 		memcpy(&view->geometry, new_geo, sizeof(struct wlr_box));
-		if (container_is_floating(view->container)) {
+		if (container_is_floating(view->container) && !container_is_scratchpad_hidden(view->container)) {
```
emersion: Doesn't that result in a broken view size if a view in the scratchpad resizes itself?

sahinf: It didn't always but sometimes yeah. Just checking the node instruction and serial instead fixed all 3 issues tbh.

sahinf:
8927.in
```swayconfig
bindsym i scratchpad show
exec foot --app-id=1
for_window [app_id="1"] floating enable, resize set 300 px 300 px
exec foot --app-id=2
for_window [app_id="2"] floating enable, resize set 2400 1000, move scratchpad
exec foot --app-id=3
for_window [app_id="3"] floating enable, resize set 80 ppt 70 ppt, move scratchpad
```
8927.gdb
```gdb
break sway/desktop/xdg_shell.c:333
commands 
    printf "Acked serial: %u\n", xdg_surface->current.configure_serial
    print *xdg_surface
    continue
end

break root_scratchpad_add_container
commands
    printf "con=%p, width=%d, height=%d, floating=%d\n", con, con->pending.width, con->pending.height, container_is_floating(con)
    continue
end
break container_floating_set_default_size
commands
    printf "Before default: width=%d, height=%d\n", con->pending.width, con->pending.height
    continue
end
break configure if view->app_id && strcmp(view->app_id, "scratch3") == 0
commands
    printf "Sway requests: %dx%d\n", width, height
    continue
end
break get_constraints
commands
    printf "min: %.0fx%.0f, max: %.0fx%.0f\n", *min_width, *min_height, *max_width, *max_height
    continue
end
break floating_fix_coordinates
commands
    printf "con=%p, width=%d, height=%d, old_box=(%d,%d %dx%d), new_box=(%d,%d %dx%d)\n", \
        con, con->pending.width, con->pending.height, \
        old->x, old->y, old->width, old->height, \
        new->x, new->y, new->width, new->height
    continue
end
break container_floating_move_to_center
commands
    printf "Before center: width=%d, height=%d\n", con->pending.width, con->pending.height
    continue
end

run --config 8927.in
```

```sh
$ meson compile -C debug && WLR_RENDERER=pixman gdb -x 8927.gdb -q ./debug/sway/sway
press i
```

sway/desktop/xdg_shell.c
```diff
@@ -321,7 +321,8 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		// containers, we resize the container to match. For tiling containers,
 		// we only recenter the surface.
 		memcpy(&view->geometry, new_geo, sizeof(struct wlr_box));
-		if (container_is_floating(view->container)) {
+		bool pending_configures = xdg_surface->current.configure_serial < xdg_surface->scheduled_serial;
+		if (container_is_floating(view->container) && !view->container->node.instruction && !pending_configures) {
```
emersion, line 325: Are these two conditions necessary? If `pending_configures` is false, I was under the impression that the view couldn't have any pending instructions?

sway/desktop/xdg_shell.c
```diff
@@ -321,7 +321,8 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		// containers, we resize the container to match. For tiling containers,
 		// we only recenter the surface.
 		memcpy(&view->geometry, new_geo, sizeof(struct wlr_box));
-		if (container_is_floating(view->container)) {
+		bool pending_configures = xdg_surface->current.configure_serial < xdg_surface->scheduled_serial;
```
emersion: The configure serial can overflow the 32-bit integer. I think `xdg_surface->current.configure_serial != xdg_surface->scheduled_serial` should be safer?

emersion (commented):
I'm not exactly sure about the correctness of this fix… I don't think I fully grasp why we should ignore size changes when we have a pending configure.

If the check is applied to the inner `if` guarding `wlr_xdg_toplevel_set_size()` instead of the outer one, does the fix still work?

sahinf:
> I'm not exactly sure about the correctness of this fix… I don't think I fully grasp why we should ignore size changes when we have a pending configure.
> 


It does feel a bit hacky, but based on my testing, the for_window rule executes and sets a pending geometry and sends a configure event to the client with the new size. at this point, `xdg_surface->scheduled_serial` is incremented but the client hasn't acknowledged it, so `current.configure_serial` < `scheduled_serial` meaning a configure is pending. The client may stil be rendering the previous frame and hasn't processed the configure event yet so it sends a `wl_surface.commit` with dimensions matching the old size. Then `handle_commit` see's the buffers new size copies it, and calls the `view_update_size()` resizing the container back to the old client size, completely overwriting the *user's* intended `for_window` geometry. 

> If the check is applied to the inner `if` guarding `wlr_xdg_toplevel_set_size()` instead of the outer one, does the fix still work?

No because the `view_update_size()` call is what resizes the container to match the stale client geometry which is the bug. Moving the guard to the inner would still allow this to run with those stale values.

sway/desktop/xdg_shell.c
sahinf: Based on my testing, node.instruction addresses whether a transaction is in progress. It is set during the resize and cleared when the transaction commits. So any client commit during this phase should be ignored to prevent overwriting the pending size, if that's the case. 

`pending_configures` is for when after the transaction commits, the instruction is cleared and the compositor sends a configure event with the new size but hte client may not have processed it yet. The client can still commit a buffer with its old size and `pending_configures` will be true because the `current.configure_serial` (last acked) is less than `scheduled_serial`. 

This was just observed/deduced behavior though and may not be intended deterministic effects, I guess.

sahinf:
Also on a separate note, if you accept this, the `xwayland.c` commit handler may also want a similar or same pending_configures guard check? Not sure.

sway/desktop/xdg_shell.c
```diff
@@ -321,7 +321,8 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 		// containers, we resize the container to match. For tiling containers,
 		// we only recenter the surface.
 		memcpy(&view->geometry, new_geo, sizeof(struct wlr_box));
-		if (container_is_floating(view->container)) {
+		bool pending_configures = xdg_surface->current.configure_serial != xdg_surface->scheduled_serial;
+		if (container_is_floating(view->container) && !view->container->node.instruction && !pending_configures) {
```
smlx, line 325: Is the separate variable required? Maybe it is easier to read like this?

```suggestion
		if (container_is_floating(view->container) &&
				!view->container->node.instruction &&
				xdg_surface->current.configure_serial == xdg_surface->scheduled_serial) {
```

smlx (commented):
Am I understanding this patch correctly?

1. Scratchpad windows (not visible) don't get instructions attached to them which is why you need to check that the scheduled and configure serials match (so you don't update the view size based on a client's stale buffer size).
1. Visible windows that have an instruction are already part of a committed transaction so it is logically incorrect to update their size and start a new commit with `transaction_commit_dirty_client()`?

Is that accurate?

sahinf:
> Am I understanding this patch correctly?
> 
> 1. Scratchpad windows (not visible) don't get instructions attached to them which is why you need to check that the scheduled and configure serials match (so you don't update the view size based on a client's stale buffer size).
> 2. Visible windows that have an instruction are already part of a committed transaction so it is logically incorrect to update their size and start a new commit with `transaction_commit_dirty_client()`?
> 
> Is that accurate?

1. It's general for all floating windows, not specific to scratchpad. The issue is for all windows with for_window directive
2. Yes that's my experience so far

sway/desktop/xdg_shell.c
sahinf: I would say it's easier to read with the variable but we can def split up that 2nd line for exceeding the 80 char limit.

PR #8935, opened 2025-11-04
Make it possible to set name when using create_output

hw0lff:
The `create_output` command is really useful, e.g. for creating virtual screens for screensharing purposes.
The only thing that i missed was the ability to set a name by hand.
`swaymsg create_output name my-useful-name` is now able to do that.

Feel free to reject this PR.
I just wanted to put my changes out there if anyone may be in need of this feature.

### DIFF
--- a/sway/commands/create_output.c
+++ b/sway/commands/create_output.c
@@ -9,23 +9,28 @@
 #include "sway/server.h"
 #include "log.h"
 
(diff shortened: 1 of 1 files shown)


emersion:
This was suggested and discussed before: https://github.com/swaywm/sway/pull/8381

PR #8938, opened 2025-11-09
sway: spawn bar with posix_spawn instead of fork + exec

WavyEbuilder:


emersion:
What is the motivation for this?

Seems like this is incompatible with `pthread_atfork()` either way.

WavyEbuilder:
> What is the motivation for this?

fork + exec is slower, and just worse compared to posix_spawn (which as per the name is also in posix). I don't see a reason to *not* use posix_spawn where possible, unlike fork it doesn't dup the entire parent. Not a huge deal, but still wasting memory.

> Seems like this is incompatible with `pthread_atfork()` either way.

I didn't realise they were there, but in the cases for sched and signals `posix_spawn` can do that:
https://man7.org/linux/man-pages/man3/posix_spawnattr_setsigdefault.3p.html
https://man7.org/linux/man-pages/man3/posix_spawnattr_setschedpolicy.3p.html

WavyEbuilder:
If we end up doing it though, we may as well do it everywhere, except `exec` (as in sway's exec, not the syscall) currently. We need FreeBSD to support `setsid()` for posix spawn. There is `POSIX_SPAWN_SETSID` in glibc, but sadly this is not in FreeBSD yet. It is new in posix 2024 though, so it should get it at some point!

C-f for `POSIX_SPAWN_SETSID` on https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_spawn.html to see it in posix 2024 (I would link to the section directly but I can't seem to do that, so C-f will have to suffice, sorry).

Also https://austingroupbugs.net/view.php?id=1044

PR #8955, opened 2025-11-28
Drop org_kde_kwin_idle files

wineee:
Has been replaced by ext_idle_notifier_v1


emersion:
We forgot to drop these files here: https://github.com/swaywm/sway/pull/7642

PR #8956, opened 2025-11-28
Drop org_kde_kwin_server_decoration_manager

wineee:
This protocol has been superseded by xdg-decoration


kennylevinsen:
Last I checked, the main user of `org_kde_kwin_server_decoration_manager` was Gtk, who had problems implementing `xdg_decoration`. What's the current status?

wineee:
There is related discussion in https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5201
 
We may need to wait longer. The current situation is that xdg-decoration and kwin_server_decoration are not fully equivalent, and GTK developers do not accept compositors being able to force clients to use SSD.

PR #8968, opened 2025-12-12
[RFC/DRAFT] Support xdg-toplevel-drag-v1

tokyovigilante:
Hi, 

This PR implements support for xdg-toplevel-drag-v1, building on the wlroots PR [here](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5210). Largely based on Mutter's implementation, and tested with Chromium's wayland backend.

Somewhat opinionated in that all secondary toplevels are created floating, which makes sense as it won't break existing tiled layouts and seems a better fit for tool palettes etc. 

My first real deep dive into the scene graph and wlroots, so review and feedback appreciated.

https://github.com/user-attachments/assets/4f5f41fa-18c2-47f7-9483-6a70b80f422e

### DIFF
--- a/include/sway/input/seat.h
+++ b/include/sway/input/seat.h
@@ -74,10 +74,22 @@ struct sway_seat_node {
 	struct wl_listener destroy;
 };
 
+struct wlr_xdg_toplevel_drag_v1;
+
 struct sway_drag {
 	struct sway_seat *seat;
 	struct wlr_drag *wlr_drag;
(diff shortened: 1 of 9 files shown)


orowith2os:
I would propose a configuration option to change the default behavior for window modes for a toplevel-drag window; having an overlay of where the window would be on the screen in a tiled state would be nice.

PR #8974, opened 2025-12-19
man: clarify usage of --no-repeat in key bindings

adamanteye:
Some user finds that toggle command makes screen flashes while enable/disable works perfectly. This is bacause repeat_delay is low and and the binding does not use the --no-repeat flag.

See Arch Linux forum discussion for context:
https://bbs.archlinux.org/viewtopic.php?id=284229

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -463,7 +463,10 @@ runtime.
 
 	Unless the flag _--no-repeat_ is set, the command will be run
 	repeatedly when the key is held, according to the repeat
-	settings specified in the input configuration.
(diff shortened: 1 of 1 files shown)


emersion:
Arguably this is a Sway bug, and no command should ever block the main loop for a long time.

Ref https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/2239

adamanteye:
Yeah I did assume that sway handle all the input events asynchronously before I encountered this rare case. The toggle takes roughly 350ms to complete and that may not be significant for most users.

For now maybe documenting and providing a quick example is much more convenient.

PR #8975, opened 2025-12-19
Updated hindi readme

danishdtp:
Updated hindi readme to include installation process of three popular packages - apt, dnf and pacman


emersion:
Thanks for the patch, but we try to keep the translated READMEs aligned with the English one. We don't list all distro packages, we only provide generic instructions.

PR #8997, opened 2026-01-12
Fix logical error in criteria matching (Fixes #6753)

### LINKED ISSUE #6753: `workspace=__focused__` criteria misbehaves on empty workspaces
- **Sway Version:** 1.6.1
- **Debug Log:** https://gist.github.com/821382383c92756d99ee3e3d1d029ee4 (reproduced with my usual config, but I've confirmed this is reproducible on the most minimal config to ever exist, of 0 bytes in size)
- **Configuration File:** provided verbatim below:
  ```
  ```
  Yes, this is in fact an empty file. I reproduced it with /dev/null as a config.
- **Stack Trace:** n/a, no crash
- **Description:**
  1. Launch sway
  2. Launch an instance of `foot` (the exact app is not relevant here, but the example command is geared towards it) - and type something to make the window recognisable
  3. Move to a different, **empty** workspace (if you started on workspace 1 and only have one monitor, move to 2)
  4. Issue the following command (via keybind or swaymsg) exactly, word to word (the app_id and the tidbit after `||` can be adjusted for your app of choice):
     `exec swaymsg '[app_id="foot" workspace=__focused__] focus' || foot`
  5. Observe that instead of a new window on an empty workspace, you're being transferred to workspace 1 with your previous foot instance

  The exact command I listed above, as far as I understand Sway, is supposed to either find a window of my favorite terminal on the **current** workspace, or, if no such window is found, spawn a new window. This is a common thing in my workflow, where I keep an instance of a browser, an editor and a terminal on a workspace for each project I work on, and it's important for me to quickly jump between them. But additionally I also wanted to spawn a new window in case there is none and I accidentally closed the last, or this is a new workspace. I could work around that with a script that would check if the workspace is completely empty, but that complicates matters.

andremarcais:
When evaluating a `__focused__` pattern, sway's logic is to consider the pattern match to have failed for a given view and attribute if a view is focused and the value of attribute in question of the focused view is not that of the view in question. Expected behaviour is that the pattern matches if there is a focused view and that the values of the attribute for the focused view and the other view are equal. (See #6753, which is the case where the `workspace` attribute is matched against.). I.e., if we write out the function `criteria_matches_view` as a logical formula, expected behaviour is

    criteria_matches_view(criteria, view) <=>
        forall (attribute, pattern) in criteria,
        (pattern is __focused__ and
            focused view exists and view.attribute == focused.attribute)
        or
        (pattern is not __focused__ and view.attribute =~ pattern)

but it is actually (pay attention to the fourth line)

    criteria_matches_view(criteria, view) <=>
        forall (attribute, pattern) in criteria,
        (pattern is __focused__ and
            (no focused view or view.attribute == focused.attribute))
        or
        (pattern is not __focused__ and view.attribute =~ pattern).

Fix program logic to reflect (the disjunctive form of) the first formula to be compatible with i3. (In passing, this fixes #6753.)

### DIFF
--- a/sway/criteria.c
+++ b/sway/criteria.c
@@ -204,7 +204,7 @@ static bool criteria_matches_view(struct criteria *criteria,
 
 		switch (criteria->title->match_type) {
 		case PATTERN_FOCUSED:
-			if (focused && lenient_strcmp(title, view_get_title(focused))) {
+			if (!focused || lenient_strcmp(title, view_get_title(focused))) {
 				return false;
 			}
 			break;
@@ -224,7 +224,7 @@ static bool criteria_matches_view(struct criteria *criteria,
 
 		switch (criteria->shell->match_type) {
 		case PATTERN_FOCUSED:
-			if (focused && strcmp(shell, view_get_shell(focused))) {
+			if (!focused || strcmp(shell, view_get_shell(focused))) {
 				return false;
 			}
 			break;
@@ -244,7 +244,7 @@ static bool criteria_matches_view(struct criteria *criteria,
 
 		switch (criteria->app_id->match_type) {
 		case PATTERN_FOCUSED:
-			if (focused && lenient_strcmp(app_id, view_get_app_id(focused))) {
+			if (!focused || lenient_strcmp(app_id, view_get_app_id(focused))) {
 				return false;
 			}
 			break;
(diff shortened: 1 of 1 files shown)


vikanezrimaya:
Can confirm original issue is fixed on this commit. Note that a sleep or two seems to be required in the auto-reproduction example to alleviate the `foot` window possibly spawning too late, thus making the command as such:

```bash
swaymsg exec foot
sleep 1  # HACK: Ensure foot has appeared
swaymsg workspace 2
sleep 1  # HACK: ensure workspace switch happens (not sure if swaymsg exits on send-off completion or on command completion; should be latter but idk)
swaymsg "exec swaymsg '[app_id=\"foot\" workspace=__focused__] focus' || swaynag -m nonreproducible"
```

andremarcais:
Just rebased and clarified the commit message. The original message had several mistakes in the body and wasn't very clear. Also made the subject line more descriptive.

PR #9001, opened 2026-01-13
sway-ipc(7): Clarify that the reply shown for GET_BINDING_STATE is an example

j0le:
Fix documentation of sway-ipc. The example reply for GET_BINDIG_STATE is an example and not the exact reply. 

This bug seems to come from copy/paste from the ipc command SYNC directly above, where the the shown reply is the exact reply and not an example.


emersion (approved):
Thanks!

PR #9007, opened 2026-01-17
call disable container in arrange_root

### LINKED ISSUE #8972: fullscreened window moved to scratchpad is displayed when a different window is fullscreened
- **Sway Version:**
  - sway version 1.11

- **Debug Log:**
[log](https://gist.github.com/Ashbd9/99c8447a57da22718126138e5f8a142f)

- **Configuration File:**
default

- **Description:**
  - start sway with the default config
  - open a window, toggle full screen, and then move it to the scratchpad
  - open another window and toggle full screen

You will find that the first window displays over the second window. The 1st window is not in focus and the 2nd window receives all keyboard input

Lompik:
```disable_container``` is similar to previous code but adds an else branch to reparent the content_tree which fixes #8972. Unless recursion when ```!con->view``` is a problem?

### DIFF
--- a/sway/desktop/transaction.c
+++ b/sway/desktop/transaction.c
@@ -648,15 +648,7 @@ static void arrange_root(struct sway_root *root) {
 	for (int i = 0; i < root->scratchpad->length; i++) {
 		struct sway_container *con = root->scratchpad->items[i];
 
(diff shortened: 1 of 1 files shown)


Nefsen402:
Thanks!

PR #9017, opened 2026-01-28
config: add include_one directive for override-safe config includes

rajj28:
### Summary

This PR adds a new configuration directive called `include_one` to sway.

The directive allows multiple include globs to be processed left-to-right,
including all files from the first glob and skipping files from subsequent
globs whose basenames were already included earlier. This makes it possible
to layer configuration directories cleanly without duplicate overrides.

---

### Motivation

Sway (and i3) support including partial configuration files from directories,
which is commonly used to layer system-wide defaults and user overrides.

For example (Regolith use case):

include /usr/share/regolith/sway/config.d/*
include $HOME/.config/regolith2/sway/config.d/*

In practice, these directories often overlap, and the same filenames may
exist in both locations. While ordering helps, there is no way to express
"include system defaults unless the user already provided an override".

This PR introduces a new directive to solve that problem explicitly.

---

### New directive: `include_one`

include_one <glob1> <glob2> [<glob3> ...]

Behavior:
- Globs are processed from left to right.
- All files from the first glob are included.
- For subsequent globs, files are included **only if their basename**
  (filename without path) has not already been included.
- Ordering is preserved.
- Duplicate basenames are skipped silently.
- If a glob matches no files, behavior matches the existing `include` command.

Example:

include_one $HOME/.config/regolith2/sway/config.d/*
/usr/share/regolith/sway/config.d/*

If both directories contain a file named `70_bar`, the version from the
user directory is included and the system one is skipped.

---

### Implementation details

- The implementation reuses sway’s existing include logic and glob expansion.
- Deduplication is performed per-file using basenames only.
- Existing `include` behavior is unchanged.
- Error paths and memory ownership are handled explicitly to avoid undefined
  behavior on allocation failure.
- The new command is registered alongside the existing `include` directive.

---

### Notes

This change is isolated to configuration parsing and does not affect compositor
or wlroots internals. It was implemented and built against sway 1.9 using Meson;
rebasing to newer versions should only require minor mechanical adjustments due
to config code refactors.

---

### Status

So if the user has [~/.config/regolith2/sway/config.d/70_bar](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), the system-wide /usr/share/regolith/sway/config.d/70_bar will be skipped, allowing user overrides.

This change was implemented as part of a qualification task and is submitted
for review and discussion.

### DIFF
--- a/include/sway/commands.h
+++ b/include/sway/commands.h
@@ -145,6 +145,7 @@ sway_cmd cmd_fullscreen;
 sway_cmd cmd_gaps;
 sway_cmd cmd_hide_edge_borders;
 sway_cmd cmd_include;
+sway_cmd cmd_include_one;
 sway_cmd cmd_inhibit_idle;
 sway_cmd cmd_input;
 sway_cmd cmd_seat;
--- a/sway/commands.c
+++ b/sway/commands.c
@@ -103,6 +103,7 @@ static const struct cmd_handler handlers[] = {
 static const struct cmd_handler config_handlers[] = {
 	{ "default_orientation", cmd_default_orientation },
 	{ "include", cmd_include },
+	{ "include_one", cmd_include_one },
 	{ "primary_selection", cmd_primary_selection },
 	{ "swaybg_command", cmd_swaybg_command },
 	{ "swaynag_command", cmd_swaynag_command },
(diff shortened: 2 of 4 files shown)


SoumyaRanjanPatnaik:
This contributions should have been raised to https://github.com/regolith-linux/sway-regolith/. Please close this PR to avoid spamming upstream sway.

PR #9018, opened 2026-01-28
sway: advertise color_manager_v1 version 2

llyyr:
Ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5223


llyyr:
I noticed that sway no longer advertised srgb because after the aforementioned MR when creating a version 1 color manager, probably a wlroots bug but we should bump it here anyway. see: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5223?commit_id=68052f34d318a6f9b2cb77074abb1105043c240a#note_3306570

PR #9024, opened 2026-02-02
tree/workspace: fix crash on dragging container to workspace edge

### LINKED ISSUE #8939: Sway crashes when dragging container
- **Sway Version:** sway version 1.12-dev-055be4ec

- **Debug Log:** [swaycrash2.log](https://github.com/user-attachments/files/23463990/swaycrash2.log)

- **Configuration File:** [swayconfig](https://github.com/user-attachments/files/23464190/swayconfig.log)

- **Stack Trace:** [trace](https://github.com/user-attachments/files/23486404/swaydump.txt)

- **Description:**
  - open sway
  - press mod+w for tabbed layout
  - press mod+s for stacked layout
  - open 2 terminals (mod+enter)
  - press mod+w for tabbed layout
  - drag topmost container (T[foot foot]) to the right side of screen
  - observe crash

sahinf:
Problem: when dragging a parent container to the right of its children, sway crashes. The calculated insertion index could exceed the list bounds. The move logic would compute an index based on the pre-removal list state.

Solution: Clamp the insertion index to [0, length].

fixes: #8939

### DIFF
--- a/sway/input/seatop_move_tiling.c
+++ b/sway/input/seatop_move_tiling.c
@@ -373,7 +373,7 @@ static void finalize_move(struct sway_seat *seat) {
 		enum sway_container_layout new_layout = edge == WLR_EDGE_TOP ||
 			edge == WLR_EDGE_BOTTOM ? L_VERT : L_HORIZ;
 		workspace_split(new_ws, new_layout);
-		workspace_insert_tiling(new_ws, con, after);
+		workspace_insert_tiling(new_ws, con, after ? new_ws->tiling->length : 0);
 	}
 
 	if (old_parent) {


sway/tree/workspace.c
```diff
@@ -842,6 +842,8 @@ struct sway_container *workspace_insert_tiling(struct sway_workspace *workspace,
 	if (config->default_layout != L_NONE) {
 		con = container_split(con, config->default_layout);
 	}
+	if (index < 0) index = 0;
+	else if (index > workspace->tiling->length) index = workspace->tiling->length;
```
WillPower3309: oneline conditionals go against the standard style in sway:


```suggestion
	if (index < 0) {
		index = 0;
	} else if (index > workspace->tiling->length) {
		index = workspace->tiling->length;
	}
```

emersion:
Should we instead fix the caller to not specify an out-of-bounds index? Something among the lines of

```c
	int after = edge != WLR_EDGE_TOP && edge != WLR_EDGE_LEFT && new_ws->tiling->length > 0;
```

sahinf:
> Should we instead fix the caller to not specify an out-of-bounds index? Something among the lines of
> 
> ```c
> 	int after = edge != WLR_EDGE_TOP && edge != WLR_EDGE_LEFT && new_ws->tiling->length > 0;
> ```

Agreed and pushed changes. Though I made the length of the list instead.

sway/input/seatop_move_tiling.c
```diff
@@ -373,7 +373,7 @@ static void finalize_move(struct sway_seat *seat) {
 		enum sway_container_layout new_layout = edge == WLR_EDGE_TOP ||
 			edge == WLR_EDGE_BOTTOM ? L_VERT : L_HORIZ;
 		workspace_split(new_ws, new_layout);
-		workspace_insert_tiling(new_ws, con, after);
+		workspace_insert_tiling(new_ws, con, after ? new_ws->tiling->length : 0);
```
emersion, line 376: Before this PR, `workspace_insert_tiling()` was called with 0 or 1. Now it's called with 0 or `new_ws->tiling->length`. Are these two equivalent? Is there a guarantee that `new_ws->tiling->length` is 1?

sahinf: Yes the `new_ws` tiling length is guaranteed to be 1. Before `workspace_insert_tiling()`, `workspace_split()` calls `workspace_wrap_children()` which leaves exactly 1 child in the workspace. It's functionally the same but the intent is clearer to me. 

~/dev/sway/sway/tree/workspace.c
```c
struct sway_container *workspace_wrap_children(struct sway_workspace *ws) {
	struct sway_container *fs = ws->fullscreen;
	struct sway_container *middle = container_create(NULL);
	middle->pending.layout = ws->layout;
	while (ws->tiling->length) {
		struct sway_container *child = ws->tiling->items[0];
		container_detach(child);
		container_add_child(middle, child);
	}
	workspace_attach_tiling(ws, middle); // <-- 1 child. 
	ws->fullscreen = fs;
	return middle;
}
```

emersion: Makes sense, thanks!

emersion (approved):
Thanks!

Issue #8999, opened 2026-01-12
Nvidia warnings


h8d13:
Hello I wanted to say thanks for sway because it's awesome, use it a lot on my laptops

Next i wanted to know if you can somehow remove all of this? 

```
	bool is_unsupported = false;
	if (strcmp(version->name, "nvidia-drm") == 0) {
		is_unsupported = true;
		sway_log(SWAY_ERROR, "!!! Proprietary Nvidia drivers are in use !!!");
		if (!allow_unsupported_gpu) {
			sway_log(SWAY_ERROR, "Use Nouveau instead");
		}
	}

	if (!allow_unsupported_gpu && is_unsupported) {
		sway_log(SWAY_ERROR,
			"Proprietary drivers are NOT supported. To launch sway anyway, "
			"launch with --unsupported-gpu and DO NOT report issues.");
		exit(EXIT_FAILURE);
	}
```

Since it works perfectly fine on even the oddest of NVIDIA hardware, I don't see the value being added here other than blocking potential people who want to use/try Sway

I've seen this posted on reddit countless times where arch users need to go into their Login Manager and add the long `--unsupported-gpu` argument when it would have just worked without it. I do get the issues part but that seems like a non-problem for established projects like this one

Or at least make it a warn and not exit? <3

Anyways thanks, sincerely, a sway enjoyer

emersion:
Proprietary drivers are not officially supported because they make things extra hard to debug when a bug report comes in and they make it so we can't fix the bug if it turns out to originate from the driver. In short, it's a maintenance burden.

The goal is not to block people from using Sway. The goal is to avoid making (volunteer) Sway developers bear the cost.

I'm not necessarily against adjusting the behavior of the error message, as long as users running in this configuration are fully aware that they should not report issues and they are on their own.

h8d13:
I do see the value of it being there, just not an error: (would have worked fine most likely!) but instead a warning (non-fatal) ? 

Basically in `archinstall` we used to have a warning for this too (Sway + nVIDIA), but thought we could have removed it since their packaging changes to open drivers + their effort to have full support on more recent hardware. Eitherway I've also been using sway on a 1050Ti and get about 12k `glmark2` score using sway there

emersion:
> I do see the value of it being there, just not an error: (would have worked fine most likely!) but instead a warning (non-fatal) ?

Maybe. The warning needs to be very visible to users, and users need to explicitly acknowledge it.

> their packaging changes to open drivers

[...]

PR #9025, opened 2026-02-02
sway: change unsupported GPU message to swaynag

4e554c4c:
This commit shows a swaynag message when an unsupported GPU is detected which must be acknowledged by users. It also adds an environment variable (`ALLOW_UNSUPPORTED_GPU`) which may be used instead of the `--unsupported-gpu` argument.

Tested on a machine with an nvidia GPU and one without.

Resolves: #8999

### DIFF
--- a/sway/main.c
+++ b/sway/main.c
@@ -224,7 +224,7 @@ static const char usage[] =
 	"\n";
 
 int main(int argc, char **argv) {
-	static bool verbose = false, debug = false, validate = false;
+	bool verbose = false, debug = false, validate = false, allow_unsupported_gpu = false;
 
 	char *config_path = NULL;
 
@@ -284,6 +284,12 @@ int main(int argc, char **argv) {
 		fprintf(stderr,
 				"XDG_RUNTIME_DIR is not set in the environment. Aborting.\n");
 		exit(EXIT_FAILURE);
+	}
+
+	char *unsupported_gpu_env = getenv("SWAY_UNSUPPORTED_GPU");
+	// we let the flag override the environment variable
+	if (!allow_unsupported_gpu && unsupported_gpu_env) {
+		allow_unsupported_gpu = parse_boolean(unsupported_gpu_env, false);
 	}
 
 	// As the 'callback' function for wlr_log is equivalent to that for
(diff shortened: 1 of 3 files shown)


h8d13:
Tysm for this 

A lot of happy users that should make (:

CodecNomad:
yuh, fire

kennylevinsen:
Missing the flag does not cause sway to crash, it causes it to log the issue and refuse to start.

I imagine we may eventually drop the flag as nvidia drivers are becoming usable enough, even if we'll never be able to provide support for such driver issues.

4e554c4c:
@kennylevinsen yeah perhaps the commit wording is misleading, i can change it.

This is mostly following up on the discussion from #8999 with @emersion.

I find that as an end-user the current behavior might _seem_ like a crash if sway is started from a DM.

sway/main.c
```diff
@@ -286,6 +286,11 @@ int main(int argc, char **argv) {
 		exit(EXIT_FAILURE);
 	}
 
+	char *unsupported_gpu_env = getenv("ALLOW_UNSUPPORTED_GPU");
```
kennylevinsen: If we want an env var it should probably be more specific, e.g., `SWAY_ALLOW_UNSUPPORTED_GPU=1`. A bit verbose, but oh well.

sway/main.c
```diff
@@ -286,6 +286,11 @@ int main(int argc, char **argv) {
...
+	if (unsupported_gpu_env) {
+		allow_unsupported_gpu = parse_boolean(unsupported_gpu_env, false);
+	}
```
kennylevinsen: This setup means that `ALLOW_UNSUPPORTED_GPU=false sway --unsupported-gpu` would have `allow_unsupported_gpu` be set to false as `parse_boolean` overwrote the value.

Maybe just a quick `if (unsupported_gpu_env && !allow_unsupported_gpu) {`?

kennylevinsen (approved):
LGTM

sreddahc:
> I imagine we may eventually drop the flag as nvidia drivers are becoming usable enough, even if we'll never be able to provide support for such driver issues.

This would have been my preferred solution but appreciate the progress. Thank you ❤️

PR #9034, opened 2026-02-10
fix: container state stacked and tabbed y position and height

WillPower3309:
With the current logic in `arrange.c` the y position and height of staked and tabbed containers is incorrectly assigned to be the size + position of their parent container. This fix also allows us to simplify some placement and sizing logic in `transaction.c` and `view.c`.

### DIFF
--- a/sway/tree/arrange.c
+++ b/sway/tree/arrange.c
@@ -188,7 +188,7 @@ static void apply_tabbed_layout(list_t *children, struct wlr_box *parent) {
 	}
 	for (int i = 0; i < children->length; ++i) {
 		struct sway_container *child = children->items[i];
-		int parent_offset = child->view ? 0 : container_titlebar_height();
+		int parent_offset = child->view ? container_titlebar_height() : 0;
 		child->pending.x = parent->x;
 		child->pending.y = parent->y + parent_offset;
 		child->pending.width = parent->width;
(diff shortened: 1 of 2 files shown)


WillPower3309:
This looks like it will also need a change in view_autoconfigure as it previously compensated for the this incorrect position / size

WillPower3309:
Just need to test some edge cases prior to marking it as ready

Nefsen402:
When I wrote the layout code in transaction.c, the intention was to not use pending/current sizes. The idea was to move the transaction system to something scene native and therefore arrange.c can be removed someday.

WillPower3309:
> When I wrote the layout code in transaction.c, the intention was to not use pending/current sizes. The idea was to move the transaction system to something scene native and therefore arrange.c can be removed someday.

this makes sense. Given that the arrange.c logic is still incorrect and we could simplify view.c a bit since it relies on the position from arrange.c anyway, do the non transaction portions of this PR still seem desirable?

WillPower3309:
> When I wrote the layout code in transaction.c, the intention was to not use pending/current sizes. The idea was to move the transaction system to something scene native and therefore arrange.c can be removed someday.

one more thought here: we require knowing the pending size in order to do a view reconfigure correctly. In this case we need to do an "initial size pass" to determine the size of the view to configure prior to any arrangement in scene. Don't we need arrange.c then? I don't see how we can avoid this, and we'd save a lot of complexity just calculating positional values while we're at it. If we really want to do this based on relative positions, we can have the current / pending values of x and y be relative.

WillPower3309:
CC @Nefsen402 for your thoughts on the above. If you have an idea for that initial size pass that gets around the arrange.c requirement, I'd be happy to implement it but I can't see any other way without making transaction.c super hacky

PR #9043, opened 2026-02-22
commands/exec: redirect child stdio to /dev/null

### LINKED ISSUE #4214: `swaymsg exec nvim` corrupts Linux VT

OctopusET:
Children spawned via exec inherit sway's stdio pointing to the launch
TTY. wlroots puts the TTY in KD_GRAPHICS mode, where writes block,
causing programs that write to stdout/stderr on startup (e.g. Electron
apps like Signal, Discord) to hang when launched from a launcher like
fuzzel. Launching from a terminal works since it provides its own PTY.

Closes #4214
Supersedes #4221

### DIFF
--- a/sway/commands/exec_always.c
+++ b/sway/commands/exec_always.c
@@ -1,3 +1,4 @@
+#include <fcntl.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
@@ -52,6 +53,14 @@ struct cmd_results *cmd_exec_process(int argc, char **argv) {
 	pid_t child = fork();
 	if (child == 0) {
 		setsid();
+
+		int devnull = open("/dev/null", O_RDWR);
+		if (devnull > STDERR_FILENO) {
+			dup2(devnull, STDIN_FILENO);
+			dup2(devnull, STDOUT_FILENO);
+			dup2(devnull, STDERR_FILENO);
+			close(devnull);
+		}
 
 		if (ctx) {
 			const char *token = launcher_ctx_get_token_name(ctx);


kennylevinsen:
> wlroots puts the TTY in KD_GRAPHICS mode, where writes block, causing programs that write to stdout/stderr on startup

Eh? If writes blocked, sway itself would stall as we're writing our log output to stderr.

The referenced bugs only talk about stdin, mostly related to cases where one accidentally launched something with `swaymsg exec` that should probably have failed by virtue of having no interactive terminal like `nvim`, but due to having the original stdin available it launches just fine and just sits "in the background" for when sway terminates, as if you had run `nvim &` before starting sway.

As such, the fix they are discussing only relates to replacing stdin.

(and minor nit, it's logind/seatd, not wlroots, that set the mode to KD_GRAPHICS.)

OctopusET:
The real fix was only changing `exec sway | true` -> `sway` in `.zprofile`. It was my own bug.

It wasn't related to KD_GRAPHICS TTY nor sway. Electron apps were spinning on EPIPE in a loop due to `| true`

Still we can change this PR to only redirect stdin to `/dev/null` not stdout and stderr, to fix #4214 as suggested in https://github.com/swaywm/sway/pull/4221#pullrequestreview-246210085

WhyNotHugo:
This would be very annoying; I don't want logs for my applications to me implicitly sent to `/dev/null`.

PR #9045, opened 2026-02-22
Re-evaluate "default_orientation auto" when workspaces change outputs

mil-ad:
When `default_orientation auto` is set, workspace layout (horizontal vs vertical) is determined by output dimensions _at creation time_ but never re-evaluated. This means a workspace created on a landscape display keeps `L_HORIZ` even after moving to a portrait display, which defeats the
purpose of auto orientation in multi-monitor setups.

This patch adds `workspace_reorient_auto()` which re-evaluates the layout based on the _current_ output's aspect ratio. It is called when:

- A workspace is moved to a different output (`move workspace to output`)
- An output is disconnected and workspaces evacuate
- An output is reconnected and workspaces restore back
- An output's mode or transform changes (e.g. rotation)

The function is conservative and only acts when:

- `default_orientation` is `auto` (explicit horizontal/vertical is never touched)
- The workspace has at most 1 tiling child (existing multi-window layouts are preserved)
- The workspace is not tabbed or stacked

### DIFF
--- a/include/sway/tree/workspace.h
+++ b/include/sway/tree/workspace.h
@@ -166,4 +166,6 @@ bool sway_ext_workspace_init(void);
 void sway_ext_workspace_output_enable(struct sway_output *output);
 void sway_ext_workspace_output_disable(struct sway_output *output);
 
+void workspace_reorient_auto(struct sway_workspace *ws);
+
 #endif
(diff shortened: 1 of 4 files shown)


emersion:
How does i3 behave?

emersion:
(Sorry for closing, I fat fingered the button…)

mil-ad:
> How does i3 behave?

I believe i3 is more conservative than this: it only re-orients on mode/rect change on the same output, never on cross-output moves, and also skips multi-child workspaces. This PR re-evaluates in more situations (move-to-output, evacuation, reconnect, transform). If we're matching i3 behaviour exactly then this can be closed

PR #9049, opened 2026-02-24
layer-shell-v1: Bump to version 5

kennylevinsen:
This adds support for set_exclusive_edge.

References: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5275


kennylevinsen:
Missed the sway is vendoring the protocol file: https://github.com/swaywm/sway/pull/9050

emersion:
Does using an old vendored XML cause issues when running Sway? I would've expected this to not cause any bugs.

kennylevinsen:
```
[wlr] [wayland] wl_global_create: implemented version for 'zwlr_layer_shell_v1' higher than interface version (5 > 4)
```

Yeah I wasn't expecting this either.

emersion:
Hm, weird. wlroots also generates the header, I would've expected wlroots' copy of the generated code to be included in the dynamic library which uses `wl_global_create()`…

PR #9050, opened 2026-02-24
protocols: Bump vendored wlr-protocols

kennylevinsen:
Missed this when bumping to layer shell v5.

Not entirely sure why we need to vendor these two protocols definitions and not the rest. Would be nice if we could just remove them and rely on wlroots' deifnition.

### DIFF
--- a/protocols/wlr-layer-shell-unstable-v1.xml
+++ b/protocols/wlr-layer-shell-unstable-v1.xml
@@ -25,7 +25,7 @@
     THIS SOFTWARE.
   </copyright>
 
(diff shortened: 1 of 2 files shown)


emersion:
We mostly need these vendored protocols for (1) generated enums that wlroots headers use and (2) our own clients.

kennylevinsen:
I got a surprising:
 ```
 [wlr] [wayland] wl_global_create: implemented version for 'zwlr_layer_shell_v1' higher than interface version (5 > 4)
 ```
 
Maybe a symbol clash between sway and wlroot's versions considering i link statically.

PR #9051, opened 2026-02-26
layer-shell: handle popup reposition for unconstraining

### LINKED ISSUE #8518: Layer shell popups don't appear to be constrained to the screen
sway version 1.10

## Description

I am working on a branch to migrate waybar to gtk4. One issue I've run into is that if the bar is on the bottom , then tooltips are shown off screen, because they are drawn *underneath* the bar instead of above. 

This is partly due to the fact that GTK seems to always try to draw the tooltips below the target widget, however, It also seems that sway (or maybe wlroots?) isn't properly configuring the xdg_popup for the tooltip and constraining it to the screen. 

If I make a normal floating window, with a single label, and put it down at the bottom of the screen, the tooltip is correctly positioned above the mouse, and I see an xdg_popup.configure event from sway telling it that it has configured the popup above the window. 

However, with waybar, which uses the layer shell, the configure event says to place the popup below, even though that is off screen. 

I'm guessing the constraint logic just isn't being used for the layer shell. 

## Debug logs

Here are the logs from WAYLAND_DEBUG=1 of my version of waybar, showing the tooltips off screen:
https://gist.github.com/tmccombs/3b55e689d51abbbd7d2235e4e5c36035

haug1:
## Summary

  Handle `xdg_popup.reposition` for layer-shell popups so unconstrained geometry is re-applied after reposition requests.

  ## Problem

  For layer-shell popups, we unconstrain on initial commit only.
  When clients (eg. GTK4) later send `xdg_popup.reposition`, wlroots updates scheduled popup geometry from the positioner again. Without handling `reposition`, the previous unconstrained placement is lost and popups can end up off-
  screen (eg. tooltips below bottom edge).

  ## Root Cause

  Layer-shell popup code was missing a `wlr_xdg_popup.events.reposition` listener.

  Additionally, popup cleanup was tied to `wlr_popup->base->events.destroy` (xdg-surface destroy), which is too late for popup-scoped listeners: wlroots asserts popup listener lists are empty during popup destroy.

  ## Fix

  - Add `reposition` listener to layer-shell popups and call `popup_unconstrain()`.
  - Move destroy listener from `wlr_popup->base->events.destroy` to `wlr_popup->events.destroy`.
  - Remove the new `reposition` listener in popup cleanup.

  ## Why `wlr_popup->events.destroy`

  `reposition` is a popup-scoped signal (`wlr_xdg_popup.events.*`), so its listener must be removed during popup destroy, not base-surface destroy.

  ## Testing

  - Reproduced with GTK4 layer-shell tooltip near output edge.
  - Before: tooltip may render outside viewport after reposition.
  - After: tooltip is constrained/flipped/slid into visible area.
  - No crash when popup is destroyed.

  Fixes #8518.

### DIFF
--- a/include/sway/layers.h
+++ b/include/sway/layers.h
@@ -33,6 +33,7 @@ struct sway_layer_popup {
 	struct wl_listener destroy;
 	struct wl_listener new_popup;
 	struct wl_listener commit;
+	struct wl_listener reposition;
 };
 
 struct sway_output;
--- a/sway/desktop/layer_shell.c
+++ b/sway/desktop/layer_shell.c
@@ -321,6 +321,7 @@ static void popup_handle_destroy(struct wl_listener *listener, void *data) {
 	wl_list_remove(&popup->destroy.link);
 	wl_list_remove(&popup->new_popup.link);
 	wl_list_remove(&popup->commit.link);
+	wl_list_remove(&popup->reposition.link);
 	free(popup);
 }
 
@@ -356,6 +357,11 @@ static void popup_handle_commit(struct wl_listener *listener, void *data) {
 	}
 }
 
+static void popup_handle_reposition(struct wl_listener *listener, void *data) {
+	struct sway_layer_popup *popup = wl_container_of(listener, popup, reposition);
+	popup_unconstrain(popup);
+}
+
 static void popup_handle_new_popup(struct wl_listener *listener, void *data);
 
 static struct sway_layer_popup *create_popup(struct wlr_xdg_popup *wlr_popup,
@@ -376,11 +382,13 @@ static struct sway_layer_popup *create_popup(struct wlr_xdg_popup *wlr_popup,
 	}
 
 	popup->destroy.notify = popup_handle_destroy;
-	wl_signal_add(&wlr_popup->base->events.destroy, &popup->destroy);
+	wl_signal_add(&wlr_popup->events.destroy, &popup->destroy);
 	popup->new_popup.notify = popup_handle_new_popup;
 	wl_signal_add(&wlr_popup->base->events.new_popup, &popup->new_popup);
 	popup->commit.notify = popup_handle_commit;
 	wl_signal_add(&wlr_popup->base->surface->events.commit, &popup->commit);
+	popup->reposition.notify = popup_handle_reposition;
+	wl_signal_add(&wlr_popup->events.reposition, &popup->reposition);
 
 	return popup;
 }


haug1:
Here's a simple reproduction in python that I've ran with command `LD_PRELOAD=/usr/lib/libgtk4-layer-shell.so python3 app.py`:

```
#!/usr/bin/env python3

import gi

gi.require_version("Gtk", "4.0")
gi.require_version("Gtk4LayerShell", "1.0")
from gi.repository import Gtk4LayerShell, Gtk


def on_activate(app: Gtk.Application) -> None:
    window = Gtk.ApplicationWindow(application=app)
    window.set_decorated(False)

    Gtk4LayerShell.init_for_window(window)
    Gtk4LayerShell.set_layer(window, Gtk4LayerShell.Layer.TOP)
    Gtk4LayerShell.set_anchor(window, Gtk4LayerShell.Edge.LEFT, True)
    Gtk4LayerShell.set_anchor(window, Gtk4LayerShell.Edge.RIGHT, True)
    Gtk4LayerShell.set_anchor(window, Gtk4LayerShell.Edge.BOTTOM, True)
    Gtk4LayerShell.auto_exclusive_zone_enable(window)

    bar = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL)
    bar.set_margin_top(6)
    bar.set_margin_bottom(6)
    bar.set_margin_start(10)
    bar.set_margin_end(10)

    spacer = Gtk.Box()
    spacer.set_hexpand(True)
    bar.append(spacer)

    label = Gtk.Label(label="Simple Taskbar")
    label.set_tooltip_text(
        "This is a long multi-line tooltip from the taskbar label.\n"
        "Line 2: It should expand vertically so overlap behavior is obvious.\n"
        "Line 3: Hover and keep the pointer still to view the full tooltip.\n"
        "Line 4: GTK tooltip rendering size depends on compositor/theme limits."
    )
    bar.append(label)

    window.set_child(bar)
    window.present()


if __name__ == "__main__":
    app = Gtk.Application(application_id="com.example.bottom_bar")
    app.connect("activate", on_activate)
    app.run()
```

When building master, tooltip/popup will render outside screen viewport. When building this branch, tooltip/popup will appear inside screen viewport.

(Issue is also present in this more advanced app I'm working on: https://github.com/haug1/vibar. Only way to work around this currently is implementing your own gtk popover and handling it yourself)

haug1:
@emersion I'm not sure why my PR has not been merged, I'm worried it will be ignored, so I've tried to add more information to make it clear what these changes are doing and why they're needed. 
Please I hope you will have another look! :heart:

kennylevinsen:
> I'm not sure why my PR has not been merged, 

... only two days have passed since you posted it, normal etiquette is to poke after ~2 weeks. Also, please don't randomly single out and ping maintainers, we get notified by the creation and commentary on PRs already.

kennylevinsen (approved):
LGTM!

haug1:
Great news. Thank you

PR #9058, opened 2026-03-06
sway/commands/move.c: negative coords move to the bottom/right edge

cizra:
Add a possibility to move a window to the bottom or right edge.
Sample config, tested and working:
```
for_window [app_id="whatever"] move position -1 ppt -1 ppt
```

### DIFF
--- a/sway/commands/move.c
+++ b/sway/commands/move.c
@@ -809,9 +809,10 @@ static struct cmd_results *cmd_move_to_position_pointer(
 }
 
 static const char expected_position_syntax[] =
-	"Expected 'move [absolute] position <x> [px] <y> [px]' or "
+	"Expected 'move [absolute] position <x> [px|ppt] <y> [px|ppt]' or "
 	"'move [absolute] position center' or "
-	"'move position cursor|mouse|pointer'";
+	"'move position cursor|mouse|pointer' "
+	"(negative coordinates are relative to the workspace/screen bottom-right corner)";
 
 static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
 	struct sway_container *container = config->handler_context.container;
@@ -941,6 +942,26 @@ static struct cmd_results *cmd_move_to_position(int argc, char **argv) {
 	case MOVEMENT_UNIT_INVALID:
 		sway_assert(false, "invalid y unit");
 		break;
+	}
+	// Negative coordinates are measured from the bottom-right corner of the
+	// screen/workspace rather than the top-left corner, so that e.g.
+	// "move position -10px -10px" places the window's bottom-right corner
+	// 10px from the screen/workspace bottom-right corner.
+	if (lx.amount < 0) {
+		if (absolute) {
+			lx.amount = root->x + root->width + lx.amount
+				- container->pending.width;
+		} else {
+			lx.amount = ws->width + lx.amount - container->pending.width;
+		}
+	}
+	if (ly.amount < 0) {
+		if (absolute) {
+			ly.amount = root->y + root->height + ly.amount
+				- container->pending.height;
+		} else {
+			ly.amount = ws->height + ly.amount - container->pending.height;
+		}
 	}
 	if (!absolute) {
 		lx.amount += ws->x;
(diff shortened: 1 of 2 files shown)


cizra:
Who do I ping to get this reviewed? Let's randomly pick @emersion as the author of a recent commit - would you know who to ping?

emersion:
Is this something i3 does, or is this a new feature?

cizra:
@emersion I've no idea, I haven't used i3 almost at all.
I just wanted a way to place the window in bottom right corner, independent of window size or screen resolution. I didn't find any other way to do it.

Arnavion:
Outputs can be placed at negative positions and a negative move position would be valid in that case, which this PR would break.

Even with all outputs at non-negative positions, someone may want to move to a negative position to position the window slightly off-screen, which this PR would break.

---

>I just wanted a way to place the window in bottom right corner, independent of window size or screen resolution. I didn't find any other way to do it.

Parse `swaymsg -t get_outputs` with `jq` to get all the output positions and dimensions and then the math to get the bottom-right corner.

cizra:
@Arnavion thanks for the insightful comments.
Currently I have this one-liner in my `~/.config/sway/config`:
```
for_window [app_id="^$" shell="xdg_shell"] move position -1 ppt -1 ppt
```

How would I implement your suggestion? I would prefer not to change it to a daemon listening to events and moving windows around - it'd be an external system to manage, and it'd possibly introduce some flicker, if the window gets drawn somewhere, then moved elsewhere.

Would it be better to create a new command, movebr (for move-bottom-right) with this calculation logic?

gwhitney:
I, too, would like some feature very similar to this, as I have floating windows I want to place near the bottom/right of my display, and doing a round trip to an external executable to do the position computation and then back to sway with swaymsg seems much more cumbersome in addition to the drawbacks @cizra lists. After all, both Xwindows geometry specifications (`-100-100`) and css ('right' and 'bottom' properties) have ways of specifying offsets for the right and bottom edges of  rectangles from their enclosures.

Please could a maintainer provide guidance on a design for this type of feature that would be acceptable? (And @cizra I am happy then to help with implementation if you would like.)

Personally, I would suggest that the options need to be independent for vertical and horizontal positioning; one might want to put a window in the top right or bottom left corners as well.

Therefore I would suggest that the possibility of keywords `left` or `right` before the first dimension and `top` or `bottom` before the second dimension be introduced, so that the overall syntax for this command would be

`move [absolute] position [left|right] <x> [px|ppt] [top|bottom] <y> [px|ppt]`

where either the keyword `left` or no keyword in that position at all would produce the current horizontal positioning behavior, while `right` would place the right edge of the window the given dimension away from the right edge of its workspace/screen, and similarly the keyword `top` or no keyword in that position would give the current vertical positioning behavior, while `bottom` would place the bottom edge of the window the given dimension away from the bottom of its workspace/screen.

Would revising this PR to implement this suggested syntax allow it to go on to further review? Or is there a different forum that would be preferable to the PR queue for proposing such a feature and having it approved for implementation? Thanks so much for guidance.

cizra:
@gwhitney LGTM

emersion:
Sorry, but we can't break i3 compat.

There is no need for a daemon, a simple script can do this.

gwhitney:
> Sorry, but we can't break i3 compat.
> 
> There is no need for a daemon, a simple script can do this.

* OK, I totally understand that a current "invariant" of this project is that any valid config file for sway must be parseable without error by i3 and vice-versa, which makes a proposal like this one regrettably an automatic "won't implement" for sway, and that there are other projects to look at which might be amenable to backwards-compatible feature extensions.

* However, I don't see how it is possible to accomplish what the original submitter wants without spawning a long-running process which listens for sway IPC events, and when a "new" window event is issued, checks if that window meets the desired criteria, and if it does, parses the results of a get_tree message, pulls out the size of the workspace and the size of the window, calculates the necessary top/left coordinates for the desired placement, and then issues a swaymsg command to move that new window. I happen to agree with the OP that's a fairly heavy implementation/maintenance effort for a seemingly modest goal, but perhaps we are both missing some much simpler way to make certain new windows show up bottom right? If there is such a method and someone has two minutes to enlighten us, it would be very helpful and could possibly help future sway users who encounter this sort of thing as well. Thanks.

PR #8969, opened 2025-12-13
sway/desktop/transaction: null check node before accessing node

llyyr:
It's possible for this to be null when powering on display from a powered off state.

```
==1947040==ERROR: AddressSanitizer: heap-use-after-free on address 0x7ce3d56a7c60 at pc 0x00000048d0c4 bp 0x7ffc25f0a5b0 sp 0x7ffc25f0a5a8 READ of size 8 at 0x7ce3d56a7c60 thread T0
    #0 0x00000048d0c3 in transaction_add_node ../sway/desktop/transaction.c:174
    #1 0x00000048d0c3 in _transaction_commit_dirty ../sway/desktop/transaction.c:955
    #2 0x00000048d0c3 in _transaction_commit_dirty ../sway/desktop/transaction.c:941
    #3 0x000000522581 in apply_resolved_output_configs ../sway/config/output.c:1022
    #4 0x000000522581 in apply_output_configs ../sway/config/output.c:1070
    #5 0x00000047445f in timer_modeset_handle ../sway/desktop/output.c:407
    #6 0x7fa3d8b1e4bc in wl_timer_heap_dispatch ../src/event-loop.c:529
    #7 0x7fa3d8b1e4bc in wl_event_loop_dispatch ../src/event-loop.c:1052
    #8 0x7fa3d8b200b4 in wl_display_run ../src/wayland-server.c:1584
    #9 0x00000041fc45 in main ../sway/main.c:376
    #10 0x7fa3d702b2fa in __libc_start_call_main (/lib64/libc.so.6+0x2b2fa) (BuildId: 8523b213e7586a93ab00f6dd476418b1e521e62c)
    #11 0x7fa3d702b3ca in __libc_start_main_impl (/lib64/libc.so.6+0x2b3ca) (BuildId: 8523b213e7586a93ab00f6dd476418b1e521e62c)
    #12 0x000000421cb4 in _start ../sysdeps/x86_64/start.S:115
0x7ce3d56a7c60 is located 32 bytes inside of 424-byte region [0x7ce3d56a7c40,0x7ce3d56a7de8) freed by thread T0 here:
    #0 0x7fa3d93208eb  (/lib64/libasan.so.8+0x1208eb) (BuildId: cbfe49f3b7600c4f194d4c54774c977296e9d98a)
    #1 0x00000048701f in transaction_destroy ../sway/desktop/transaction.c:68
    #2 0x00000048701f in transaction_progress ../sway/desktop/transaction.c:766
[...]

PR #9060, opened 2026-03-07
tree/node: add and use node_finish in destroy functions

llyyr:
Otherwise we might try to modset on an output that is already removed. This happens because of the timer from request_modeset call in begin_destroy firing at inconvenient times

Correct fix to https://github.com/swaywm/sway/pull/8969

### DIFF
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -273,6 +273,7 @@ void output_destroy(struct sway_output *output) {
 	list_free(output->workspaces);
 	list_free(output->current.workspaces);
 	wlr_color_transform_unref(output->color_transform);
+	node_finish(&output->node);
 	free(output);
 }
 
--- a/include/sway/tree/node.h
+++ b/include/sway/tree/node.h
@@ -60,6 +60,8 @@ const char *node_type_to_str(enum sway_node_type type);
  */
 void node_set_dirty(struct sway_node *node);
 
+void node_finish(struct sway_node *node);
+
 bool node_is_view(struct sway_node *node);
 
 char *node_get_name(struct sway_node *node);
--- a/sway/tree/container.c
+++ b/sway/tree/container.c
@@ -530,6 +530,7 @@ void container_destroy(struct sway_container *con) {
 
 	scene_node_disown_children(con->content_tree);
 	wlr_scene_node_destroy(&con->scene_tree->node);
+	node_finish(&con->node);
 	free(con);
 }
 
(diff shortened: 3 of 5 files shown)


sway/desktop/transaction.c
```diff
@@ -60,6 +60,12 @@ static void transaction_destroy(struct sway_transaction *transaction) {
 			node->instruction = NULL;
 		}
 		if (node->destroying && node->ntxnrefs == 0) {
+			// Remove from dirty_nodes before freeing to prevent a later
+			// _transaction_commit_dirty call reading freed memory
+			int dirty_idx = list_find(server.dirty_nodes, node);
+			if (dirty_idx >= 0) {
+				list_del(server.dirty_nodes, dirty_idx);
```
emersion: I wonder if this is the right spot to remove the node from the dirty list. Maybe it would make more sense to introduce a `node_finish()` function which fires the `destroy` event and removes the node from the dirty list?

llyyr: We already have *_begin_destroy functions which fire the destroy event, so I should just remove the node from dirty list in output_destroy then.

But I'm not really sure if this is still the core issue, or if the request_modeset in begin_destroy is misplaced. See: https://github.com/swaywm/sway/pull/8091#pullrequestreview-2152146460

emersion: The output is removed from the scene and from the output list in `begin_destroy()`, so the modeset logic will never touch it AFAIU. I don't think moving the `request_modeset()` call elsewhere will change anything.

sway/tree/output.c
```diff
@@ -266,6 +266,10 @@ void output_destroy(struct sway_output *output) {
 		return;
 	}
 
+	int dirty_idx = list_find(server.dirty_nodes, &output->node);
+	if (dirty_idx >= 0) {
+		list_del(server.dirty_nodes, dirty_idx);
+	}
```
emersion: Hm, but this only handles the issue for outputs, and not for other kinds of nodes. That's why I've suggested to introduce a small function to finish nodes instead.

llyyr: I've done this but now I'm not sure if we should be setting nodes as dirty when destroying them in the first place.

emersion: Hm, I'm a bit confused because there are asserts right above these new calls asserting that the node isn't part of any transaction… So we're adding the destroying node to a transaction after that assert but before the node is destroyed?

llyyr: Hmm I thought I understood what was going on here but indeed I'm not sure I can explain why this commit fixes the problem... I'll try to better understand what is going on here.

Going by the backtrace though it does seem like a race related to the request_modeset timer because apply_output_configs is called from timer_modeset_handle after a previous transaction already freed the output.

PR #9066, opened 2026-03-13
follow wlroots 3879: Nuke buffer output_enter / output_leave

GreyXor:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3879


Nefsen402:
This isn't correct, we can't just change which signal we listen on. The real fix is going to be a bit more complex

emersion:
Indeed, we'll need to figure out which outputs are entered/left...

Nefsen402:
I finally got sway to compile so I'll write a patch.

Nefsen402:
Superseded by https://github.com/swaywm/sway/pull/9067

Issue #9000, opened 2026-01-13
wlr_foreign_toplevel_management: sway ignores output_enter/output_leave events for fullscreen window


Lompik:
- **Sway Version:**
  - sway version 1.12-dev-3c0588a4 (Jan  9 2026, branch 'master')

- **``` WAYLAND_DEBUG=1 foreign_toplevel -o 60 -m``` log**
  - note: the ```foreign_toplevel``` log comes from https://gitlab.freedesktop.org/wlroots/wlr-clients and specify ```-m $(wayland_info::screen::name)``` otherwise ```output_{enter,leave}``` events are not generated
  - start imv in fullscreen ```imv -f s.jpg``` then unfullscreen it by pressing f @  676948:
```
[ 675607.124] {Default Queue} zwlr_foreign_toplevel_manager_v1#4.toplevel(new id zwlr_foreign_toplevel_handle_v1#4278190100)
[ 675607.130] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190100.title("imv - [1/1] [0x0] [  100%] s.jpg  [scale to fit]")
[ 675607.132] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190100.state(array[4])
[ 675607.133] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190081.state(array[0])
[ 675607.134] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190100.state(array[8])
[ 675607.135] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190100.app_id("imv")
[ 675607.136] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190100.done()
[ 675607.143] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190081.done()
# 2 terminals leaving wl_output#6 (replace by imv fullscreen which i can see)
[ 675608.688] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190086.output_leave(wl_output#6)   
[ 675608.697] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190081.output_leave(wl_output#6)
[ 675608.700] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190100.title("imv - [1/1] [0x0] [-2147483648%] s.jpg  [scale to fit]")
[ 675608.701] {Default Queue} zwlr_foreign_toplevel_handle_v1#4278190086.done()
[...]

PR #9067, opened 2026-03-13
Chase wlroots!3879

Nefsen402:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3879


GreyXor:
Indeed better than mine 🫣, thanks!

Lompik:
I think this fixes https://github.com/swaywm/sway/issues/9000 as well

PR #9068, opened 2026-03-13
tree/container: use outputs_update instead of output_{enter,leave}

llyyr:
Ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3879


Nefsen402:
https://github.com/swaywm/sway/pull/9067

llyyr:
Ah, thanks! I didn't realize the active output was already tracked

Nefsen402:
I think this solution would UAF too if an output gets destroyed btw.

PR #9078, opened 2026-03-18
config: add include_one directive

Snorlax-011:
Adds `include_one <pattern1> <pattern2>`, which loads all files
matching pattern1, then loads files from pattern2 only if their
basename was not already loaded from pattern1.

This allows user configs in ~/.config/sway/conf.d/ to shadow
distro-provided defaults in /etc/sway/conf.d/ by filename,
a common pain point for packagers on NixOS, Fedora, etc.

**Test plan:**
1. `mkdir -p /tmp/test-sway/user /tmp/test-sway/system`
2. `echo 'exec notify-send user' > /tmp/test-sway/user/autostart.conf`
3. `echo 'exec notify-send system' > /tmp/test-sway/system/autostart.conf`
4. Add to sway config: `include_one /tmp/test-sway/user/*.conf /tmp/test-sway/system/*.conf`
5. Reload sway — only "user" notification appears, not "system"
6. Remove `/tmp/test-sway/user/autostart.conf`, reload — "system" notification now appears

**Edge cases verified:**
- pattern1 matches nothing: all pattern2 files load normally
- Same basename in both: pattern1 wins, pattern2 skipped
- Neither pattern matches: silently succeeds (matches `include` behavior)

### DIFF
--- a/sway/config.c
+++ b/sway/config.c
@@ -682,7 +682,7 @@ void load_include_one_configs(const char *pattern1, const char *pattern2,
 	list_free(loaded_basenames);
 
 	if (chdir(wd) < 0) {
-		sway_log(SWAY_ERROR, "failed to change working directory");
+		sway_log(SWAY_ERROR, "include_one: failed to change working directory");
 	}
 cleanup:
 	free(parent_path);


Snorlax-011:
- getcwd() failure handled with early return before any allocation
- create_list() OOM handled with goto cleanup
- basename() POSIX portability fixed. pattern2 loop uses strdup'd copy

Issue #9082, opened 2026-03-19
Crash on opening window


J0nnyMak0:
With latest source, I'm getting the following crash when trying to open any window on Sway.

[gdb.txt](https://github.com/user-attachments/files/26128969/gdb.txt)

Offending commit seems to be:  2c2a2ec38055092f368b44d4affefdfa8df17ff9

PR #9083, opened 2026-03-20
common/pango: get_text_size out pointers may be NULL

poisotf:
Fixes: 2c2a2ec38055092f368b44d4affefdfa8df17ff9
Closes: https://github.com/swaywm/sway/issues/9082


emersion (approved):
Oops, thanks for the fix!

PR #9086, opened 2026-03-20
pango: don't use cairo_create(NULL) for text metrics

### LINKED ISSUE #9084: Missing text for window titles
- sway version
  - 1.12-dev-82227d61 (Mar 21 2026, branch 'test') (same as master)
  - wlroots 4ca40004fd8f35a85f4f342c6b12d3792fbb23ea (current master)

- **Debug Log:**
  -  https://gist.github.com/Billli11/91fe035bfff11ab017ce91be3107b8cc

- **Configuration File:**
  - default config

- **Stack Trace:**
  - N/A

- **Description:**
  - start sway
  - open a window

Bisected to 40e1dcd29f1950236e184bd653ee5ede55d8202b

from output

```
00:00:01.359 [ERROR] [common/pango.c:110] pango_cairo_update_layout() failed: NULL pointer
```

Image

poisotf:
This returns a context in an error state, and pango won't work with it without also reporting errors.

Create a temporary 1x1 image instead

Close https://github.com/swaywm/sway/issues/9084

### DIFF
--- a/common/pango.c
+++ b/common/pango.c
@@ -123,7 +123,8 @@ out:
 }
 
 void get_text_metrics(const PangoFontDescription *description, int *height, int *baseline) {
-	cairo_t *cairo = cairo_create(NULL);
+	cairo_surface_t *dummy_surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, 1, 1);
+	cairo_t *cairo = cairo_create(dummy_surface);
 	PangoContext *pango = pango_cairo_create_context(cairo);
 	pango_context_set_round_glyph_positions(pango, false);
 	// When passing NULL as a language, pango uses the current locale.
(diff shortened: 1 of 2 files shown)


Billli11:
This fixed the issue on my system.

dawsers:
Thank you for the fix. I would like to add a couple of comments if you don't mind:

1. Now that we are checking for errors, when creating the dummy surface, maybe it would be good to also use [`cairo_surface_status()`](https://www.cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-status) to verify the surface is valid.
2. The early return in `text_calc_size()` doesn't destroy the dummy surface.

poisotf:
>     2. The early return in `text_calc_size()` doesn't destroy the dummy surface.

https://github.com/swaywm/sway/pull/9088 is more effective, and doesn't have this surface leak. I'd go for this one instead

PR #9093, opened 2026-03-23
Centre fullscreen surfaces smaller than output

### LINKED ISSUE #8845: Fullscreen windows with geometry smaller than the output are positioned incorrectly
...these windows are incorrectly positioned in the top left corner.

Can be reproduced with `weston-simple-egl -f -r`:

Image

This behavior is [out of specification](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml#L1096-1100):

```
	If the surface doesn't cover the whole output, the compositor will
	position the surface in the center of the output and compensate with
	with border fill covering the rest of the output. The content of the
	border fill is undefined, but should be assumed to be in some way that
	attempts to blend into the surrounding area (e.g. solid black).
```

So the window should be positioned in the center (this is what COSMIC, GNOME, Niri and Weston do). Optionally, protocol `viewporter` can be used to scale and fill the window for output resolution (~this is what Weston does~. Some windows, specifically Wine windows, seem to be scaled by Weston).

And a quick note: Personally, I don't use sway. The issue that this affects several older games running through `winewayland.drv`: https://bugs.winehq.org/show_bug.cgi?id=57456. The main developer of `winewayland.drv` [commented](https://bugs.winehq.org/show_bug.cgi?id=57456#c1) that Wine expects the compositor to comply with the specification, so I believe that Wine will not fix this issue on their end, which is why I am reporting this issue to the most popular compositors.

And the advantage of using protocol `viewporter` for this is that it will also fix this issue: https://bugs.winehq.org/show_bug.cgi?id=57925. In addition to not having black borders, which I personally think is better!

WhyNotHugo:
Sway renders fullscreen surfaces smaller than the output left-aligned.

From xdg-shell:

> If the surface doesn't cover the whole output, the compositor will
> position the surface in the center of the output and compensate with
> with border fill covering the rest of the output. The content of the
> border fill is undefined, but should be assumed to be in some way that
> attempts to blend into the surrounding area (e.g. solid black).

Render surfaces smaller than the output centred. Can be tested easily with:

    weston-simple-egl -f -r

Fixes: https://github.com/swaywm/sway/issues/8845

### DIFF
--- a/sway/tree/view.c
+++ b/sway/tree/view.c
@@ -988,7 +988,7 @@ void view_center_and_clip_surface(struct sway_view *view) {
 
 	bool clip_to_geometry = true;
 
-	if (container_is_floating(con)) {
+	if (container_is_floating(con) || con->pending.fullscreen_mode != FULLSCREEN_NONE) {
 		// We always center the current coordinates rather than the next, as the
 		// geometry immediately affects the currently active rendering.
 		int x = (int) fmax(0, (con->current.content_width - view->geometry.width) / 2);


llyyr:
Should this not be done at scene level instead?

kennylevinsen:
> Should this not be done at scene level instead?

Scene doesn't currently make policy decisions like buffer placement, so fullscreen only exists as the compositor deciding to make the scene node big, on top and blacking everything else out.

Nefsen402 (approved):
LGTM.

I believe this should also work with fullscreen global because current.content_width/height will be populated with the full viewport when a container is in this state.

emersion (approved):
Thanks!

WhyNotHugo:
This being a bugfix, can we include it in 1.12 too?

emersion:
This patch has already been cherry-picked on the v1.12 branch: https://github.com/swaywm/sway/commit/4e673a3515a86bbcd345e0f84d45704b2a36df45

WhyNotHugo:
Thanks!

Issue #9075, opened 2026-03-18
Active workspace lost when switching to another VT


WhyNotHugo:
### Please fill out the following:
- **Sway Version:**
  - sway version 1.11

- **Configuration File:**
  - n/a (reproducible regardless of config)

- **Stack Trace:**
  - n/a

- **Description:**

When switching to another TTY and back, the currently active workspace is if the current workspace is empty.

- **Step by step**

- Open any window
- Move it to workspace 2
- Remain in workspace 1, which is empty
- Switch to another TTY. E.g.: <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>F2</kbd>
- Switch back to sway.  E.g.: <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>F1</kbd>

The currently active workspace will be workspace 2, rather than the previously active workspace 1.

WhyNotHugo:
Possibly related to https://github.com/swaywm/sway/issues/9057

Issue #5057, opened 2020-02-29
How does sway do "i3-input" interactively take a command?


zdmgmail:
I want to rename like this:

bindsym $mod+n exec i3-input -F 'rename workspace to "%s"' -P 'New name for this workspace: '

RedSoxFan:
There is this script here that emulates i3-input: https://github.com/swaywm/sway/issues/2#issuecomment-305380165

You could also just do something similar to the default menu command
https://github.com/swaywm/sway/blob/7fea36750ea23906c41c18a2768acad2f51b2916/config.in#L21

For example:
```
bindsym $mod+n exec echo "" | dmenu -p "New name for this workspace" | xargs -r swaymsg rename workspace to
```

zdmgmail:
Thanks,perfectly solved!

hyunlee1o:
Hey, i was searching for this, thanks for answering. :)
I mostly use letters in workspace names so this helps me.

PR #9097, opened 2026-03-26
Preserving previous output

### LINKED ISSUE #9057: On output detach/reattach, workspaces do not return to their original output
# Description

## Expected Behavior
When outputs are removed and re-added in a few scenarios:
* When returning to a login manager (observed with lightdm) with multiple monitors
* When removing an external monitor for a laptop
* When switching to a virtual terminal via ctrl-alt-<num> and returning with multiple monitors
* (for repro purposes) when disabling and re-enabling a display through sway

You would expect workspaces that had been assigned to the previous output to return to that output. Practically, for the case of a login manager, this means on switching users, all workspaces should return to the outputs that they had been assigned to before the switch.

## Actual Behavior
In these scenarios, all workspaces return to the primary output and have to be moved back to the outputs.

## Simple Repro
Start sway with the wayland backend:

```
WL_BACKENDS=wayland WLR_WL_OUTPUTS=2 sway &                
INNER_DISPLAY=wayland-2
INNER_SWAYSOCK=$(ls $XDG_RUNTIME_DIR/*.sock | tail -1)
WAYLAND_DISPLAY=$INNER_DISPLAY foot &                      
WAYLAND_DISPLAY=$INNER_DISPLAY foot &                  
SWAYSOCK=$INNER_SWAYSOCK swaymsg move container to workspace 2
SWAYSOCK=$INNER_SWAYSOCK swaymsg -t get_tree
SWAYSOCK=$INNER_SWAYSOCK swaymsg output WL-2 disable                          
SWAYSOCK=$INNER_SWAYSOCK swaymsg output WL-2 enable                          
SWAYSOCK=$INNER_SWAYSOCK swaymsg -t get_tree       
```

The output of the first get_tree will show something like
```
#1: root "root"
  #4: output "WL-2"
    #5: workspace "1"
      #7: con "" (xdg_shell, pid: 63931, app_id: "foot", foreign_toplevel_id: "...")
  #3: output "WL-1"
    #6: workspace "2"
      #8: con "" (xdg_shell, pid: 63954, app_id: "foot", foreign_toplevel_id: "...")
```
And the second will show

```
#1: root "root"
  #3: output "WL-1"
    #5: workspace "1"
      #7: con "" (xdg_shell, pid: 63931, app_id: "foot", foreign_toplevel_id: "...")
    #6: workspace "2"
      #8: con "" (xdg_shell, pid: 63954, app_id: "foot", foreign_toplevel_id: "...")
  #4: output "WL-2"
    #6: workspace "3"
```

Showing that WL-2 has been reprovisioned with an empty workspace, and workspaces 1 & 2 are assigned to output WL-1.

andrewla:
When displays are removed, save the display name that the current workspace was allocated to, and restore to that display when the display is added again.

This means that switching VTs or running `dm-tool lock` will not scramble all workspaces back to the primary output. Similarly, when dealing with a laptop connecting to a monitor, the workspaces assigned to that monitor will go back to the monitor.

This only saves one level of history; this fixes the repro described in #9057 but might not be fully general if there are more than two displays involved. Arguably the "previous output" should also be reset if certain operations occur; for example if after display detach, if a workspace is selected and modified (i.e. windows resized or layout changed) then it is likely that the intent is no longer for it to revert to the old display when reattached. Open to discussion on these points.

Fixes #9057

### DIFF
--- a/include/stringop.h
+++ b/include/stringop.h
@@ -3,6 +3,7 @@
 
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdarg.h>
 #include "list.h"
 
 #ifdef __GNUC__
--- a/include/sway/output.h
+++ b/include/sway/output.h
@@ -39,6 +39,8 @@ struct sway_output {
 	struct wlr_scene_rect *fullscreen_background;
 
 	struct wlr_output *wlr_output;
+	char *previous_output_name;
+
 	struct wlr_scene_output *scene_output;
 	struct sway_server *server;
 	struct wl_list link;
--- a/include/sway/tree/workspace.h
+++ b/include/sway/tree/workspace.h
@@ -45,6 +45,7 @@ struct sway_workspace {
 	struct side_gaps gaps_outer;
 
 	struct sway_output *output; // NULL if no outputs are connected
+	char *previous_output_name; // NULL if no previous output
 	list_t *floating;           // struct sway_container
 	list_t *tiling;             // struct sway_container
 	list_t *output_priority;
--- a/sway/tree/output.c
+++ b/sway/tree/output.c
@@ -27,6 +27,7 @@ enum wlr_direction opposite_direction(enum wlr_direction d) {
 }
 
 static void restore_workspaces(struct sway_output *output) {
+	sway_log(SWAY_DEBUG, "Restore Workspaces: restoring workspaces for output %s", output->wlr_output->name);
 	// Workspace output priority
 	for (int i = 0; i < root->outputs->length; i++) {
 		struct sway_output *other = root->outputs->items[i];
@@ -36,9 +37,13 @@ static void restore_workspaces(struct sway_output *output) {
 
 		for (int j = 0; j < other->workspaces->length; j++) {
 			struct sway_workspace *ws = other->workspaces->items[j];
+
 			struct sway_output *highest =
 				workspace_output_get_highest_available(ws);
-			if (highest == output) {
+			if (highest == output || 
+				(ws->previous_output_name != NULL &&
+				!strcmp(ws->previous_output_name, output->wlr_output->name)))
+			{
 				workspace_detach(ws);
 				output_add_workspace(output, ws);
 				ipc_event_workspace(NULL, ws, "move");
--- a/sway/tree/workspace.c
+++ b/sway/tree/workspace.c
@@ -84,6 +84,7 @@ struct sway_workspace *workspace_create(struct sway_output *output,
 	}
 
 	ws->name = strdup(name);
+	ws->previous_output_name = NULL;
 	ws->prev_split_layout = L_NONE;
 	ws->layout = output_get_default_layout(output);
 	ws->floating = create_list();
@@ -149,6 +150,7 @@ void workspace_destroy(struct sway_workspace *workspace) {
 	wlr_scene_node_destroy(&workspace->layers.fullscreen->node);
 
 	free(workspace->name);
+	free(workspace->previous_output_name);
 	free(workspace->representation);
 	list_free_items_and_destroy(workspace->output_priority);
 	list_free(workspace->floating);
@@ -291,8 +293,6 @@ static void workspace_name_from_binding(const struct sway_binding * binding,
 }
 
 char *workspace_next_name(const char *output_name) {
-	sway_log(SWAY_DEBUG, "Workspace: Generating new workspace name for output %s",
-			output_name);
 	// Scan for available workspace names by looking through output-workspace
 	// assignments primarily, falling back to bindings and numbers.
 	struct sway_mode *mode = config->current_mode;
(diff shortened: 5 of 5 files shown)


andrewla:
I want to note here that the behavior as it stands is inconsistent with what i3 does in the same situation, so I feel that this is a reasonable feature in light of backwards compatibility.

WhyNotHugo:
Sounds like it might fix https://github.com/swaywm/sway/issues/9075, I'll try and remember to test this.

Arnavion:
I'm confused, because:

>When displays are removed, save the display name that the current workspace was allocated to, and restore to that display when the display is added again.

... is already what sway does (currently using 1.11) and is what it has always done in the few years that I've been using it. My monitors are connected using DP and when I turn one off, it vanishes from `swaymsg -t get_outputs` and all workspaces on it move to the other monitor. Then I turn that monitor on and all the workspaces that used to be on it move back to it. Just to be sure I can even unplug the DP cable from GPU after I turn it off and plug it back in before I turn it back on (so that it is guaranteed to have vanished from kernel's POV), and the workspaces still move back.

And:

>This means that switching VTs [...] will not scramble all workspaces back to the primary output.

does not repro for me either. All workspaces remain on whatever outputs they used to be.

The issue reported in #9075 does repro, though.

andrewla:
@Arnavion -- that's really interesting; I tried turning off one of my monitors just now, and got the behavior that I describe in #9057, where all the workspaces move and do not move back.

Would you mind giving the output of `swaymsg -t get_tree -p   | grep -v '     '` before, during, and after the single monitor shutdown?

For comparison, my flow looks like:

```
$ swaymsg -t get_tree -p   | grep -v '     '
#1: root "root"
  #2147483647: output "__i3"
    #2147483646: workspace "__i3_scratch"
  #49: output "HDMI-A-2"
    #44: workspace "3"
  #51: output "HDMI-A-1"
    #46: workspace "1"
~
$ swaymsg -t get_tree -p   | grep -v '     '
#1: root "root"
  #2147483647: output "__i3"
    #2147483646: workspace "__i3_scratch"
  #49: output "HDMI-A-2"
    #46: workspace "1"
    #44: workspace "3"
~
$ swaymsg -t get_tree -p   | grep -v '     '
#1: root "root"
  #2147483647: output "__i3"
    #2147483646: workspace "__i3_scratch"
  #49: output "HDMI-A-2"
    #46: workspace "1"
    #44: workspace "3"
  #53: output "HDMI-A-1"
    #54: workspace "2"
```

Did you try the explicit repro in #5057 (with the synthetic displays)?

Arnavion:
>Would you mind giving the output of `swaymsg -t get_tree -p   | grep -v '     '` before, during, and after the single monitor shutdown?

Before turning off right monitor:

```
#1: root "root"
  #2147483647: output "__i3"
    #2147483646: workspace "__i3_scratch"
  #4590: output "DP-1"
    #3527: workspace "8"
    #25: workspace "9"
  #4591: output "DP-2"
    #2421: workspace "1"
    #23: workspace "2"
```

While right monitor is turned off:

```
#1: root "root"
  #2147483647: output "__i3"
    #2147483646: workspace "__i3_scratch"
  #4590: output "DP-1"
    #2421: workspace "1"
    #23: workspace "2"
    #3527: workspace "8"
    #25: workspace "9"
```

After turning right monitor back on:

```
#1: root "root"
  #2147483647: output "__i3"
    #2147483646: workspace "__i3_scratch"
  #4590: output "DP-1"
    #3527: workspace "8"
    #25: workspace "9"
  #4607: output "DP-2"
    #2421: workspace "1"
    #23: workspace "2"
```

---

Just to be sure I also tested with `output disable / enable` and it still restored the workspaces to the correct outputs.

---

>Did you try the explicit repro in https://github.com/swaywm/sway/issues/9057 (with the synthetic displays)?

Yes it does misbehave with wayland backend. Both workspaces remain on the output that remained enabled instead of one of them moving back.

Arnavion:
And just to be sure, I did the same test as #9057 with sway in real VT, with a config that was just:

```
bindsym Mod4+Return exec foot
```

and doing the foot spawns from within that session. It still behaved correctly regardless of whether I disabled and re-enabled the DP-1 output or the DP-2 output - in either case the workspace that was on that output moved back to that output when it was re-enabled.

andrewla:
I just tried the minimal repro (with the config `bindsym Mod4+Return exec foot`, using sway 1.11, and it did the behavior from #9057 -- the workspaces moved and did not move back. Are you sure you're using vanilla sway and not the changes from the PR? Otherwise I'm completely baffled -- don't see anywhere in the code where there's even the possibility of moving a workspace during display attachment.

Arnavion:
>re you sure you're using vanilla sway and not the changes from the PR?

Yes, vanilla sway with no patches, and as I said it has behaved this way since I started using sway in 2021, long before your PR.

>Otherwise I'm completely baffled -- don't see anywhere in the code where there's even the possibility of moving a workspace during display attachment.

https://github.com/swaywm/sway/commit/5c9a917df9453f0463040b1164ba639b430f7833 I think.

emersion:
Yes, Sway already restores workspaces to outputs when reconnected. This is done via [`sway_workspace.output_priority`](https://github.com/swaywm/sway/blob/3774506bd092563538af2b6d2e332c84a8ac153c/include/sway/tree/workspace.h#L51).

PR #9103, opened 2026-04-05
Use wlr_frame_scheduler (Rebase of #7780)

kennylevinsen:
Rebase of https://github.com/swaywm/sway/pull/7780. Edited to suit a post-scene world.

Depends on https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5330


poisotf:
When testing with `output max_render_time`, this is giving me a flood of "Failed to page-flip output: a page-flip is already pending" errors.

I think `sway_timer_frame_scheduler` would need to listen to output commits to set its `frame_pending`, like wlr_present_scheduler does

PR #9105, opened 2026-04-09
Don't ignore initialisation errors

WhyNotHugo:
server_init ignores all errors. In practice, theses result in a segfault, potentially much later and losing any unsaved work.

Properly handle initialisation errors and bail immediately.

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -244,13 +244,25 @@ static void handle_new_foreign_toplevel_capture_request(struct wl_listener *list
 bool server_init(struct sway_server *server) {
 	sway_log(SWAY_DEBUG, "Initializing Wayland server");
 	server->wl_display = wl_display_create();
+	if (!server->wl_display) {
+		sway_log(SWAY_ERROR, "Failed to create wl_display");
+		return false;
+	}
 	server->wl_event_loop = wl_display_get_event_loop(server->wl_display);
 
 	wl_display_set_global_filter(server->wl_display, filter_global, NULL);
 	wl_display_set_default_max_buffer_size(server->wl_display, 1024 * 1024);
 
-	wlr_fixes_create(server->wl_display, 1);
+	if (!wlr_fixes_create(server->wl_display, 1)) {
+		sway_log(SWAY_ERROR, "Failed to create wp_fixes global");
+		return false;
+	}
 	root = root_create(server->wl_display);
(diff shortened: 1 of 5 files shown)


sway/server.c
```diff
@@ -244,13 +244,22 @@ static void handle_new_foreign_toplevel_capture_request(struct wl_listener *list
...
+	if (!server->wl_display) {
+		sway_log(SWAY_ERROR, "Failed to create wl_display");
+		return false;
+	}
 	server->wl_event_loop = wl_display_get_event_loop(server->wl_display);
 
 	wl_display_set_global_filter(server->wl_display, filter_global, NULL);
 	wl_display_set_default_max_buffer_size(server->wl_display, 1024 * 1024);
 
 	wlr_fixes_create(server->wl_display, 1);
```
emersion: Can we check this one too?

sway/server.c
emersion, line 289: Ditto

emersion: Ditto

emersion: Hm, I suppose we may be missing a `wlr_renderer_get_drm_fd(server->renderer) >= 0` check too here.

emersion (commented):
Thanks for doing this, looks good apart from these three comments!

sway/server.c
WhyNotHugo: > Hm, I suppose we may be missing a `wlr_renderer_get_drm_fd(server->renderer) >= 0` check too here.

That check is a 4 lines up.

WhyNotHugo:
Updated

sway/server.c
WhyNotHugo: What even is `wlr_fixes_create`?

WhyNotHugo: Is this failure truly fatal? I'm not familiar enough to judge here, I don't even know what this does.

sway/server.c
emersion: Hm, sorry, I don't see it?

emersion: (There is such a check for drm_syncobj below, but not for linux-dmabuf. The risk here is hard-failing on software renderers instead of starting without exposing some globals.)

WhyNotHugo: Fixed now.

emersion (approved):
LGTM, thanks!

Issue #8852, opened 2025-08-25
After reloading, sway sends bogus key events


WhyNotHugo:
### Please fill out the following:
- **Sway Version:**
  - `sway version 1.11` (although this has been an issue for years)

- **Configuration File:**
  - I don't believe that anything from my config is related, but it's available [here](https://git.sr.ht/~whynothugo/dotfiles/tree/a5621c908cad7b37aed0746b3e90e31140fbd385/item/home/.config/sway/config).

- **Stack Trace:**
  - n/a

- **Description:**

When I reload sway via a terminal, I type `swaymsg reload` and press <kbd>Enter</kbd>. The key event for this last keypress is sent multiple times (and my terminal prints a bunch of empty lines).

Here's the relevant portion of `WAYLAND_DEBUG`, with some comments of my own:

```
// This is the key event for when I press Enter after typing the command.
[ 739727.528] {Default Queue} wl_keyboard#26.key(66578, 82334358, 28, 1)
// Here foot changes its title to the command currently executing.
[ 739733.488] {Default Queue}  -> xdg_toplevel#35.set_title("swaymsg reload [~]")
[ 739734.414] {Default Queue}  -> wl_surface#3.damage_buffer(240, 36, 15, 36)
[ 739734.450] {Default Queue}  -> wl_surface#3.damage_buffer(0, 72, 15, 36)
[ 739734.466] {Default Queue}  -> wl_surface#3.frame(new id wl_callback#43)
[ 739734.479] {Default Queue}  -> wl_surface#3.set_buffer_scale(1)
[ 739734.490] {Default Queue}  -> wp_viewport#32.set_destination(1031, 836)
[ 739734.503] {Default Queue}  -> wl_surface#3.attach(wl_buffer#42, 0, 0)
[ 739734.514] {Default Queue}  -> wl_surface#3.commit()
[ 739734.525] {Default Queue}  -> zwp_text_input_v3#19.set_cursor_rectangle(0, 43, 9, 21)
[ 739734.536] {Default Queue}  -> zwp_text_input_v3#19.commit()
[ 739741.251] {Default Queue} wl_buffer#42.release()
[ 740622.922] {Default Queue} wl_keyboard#26.keymap(1, fd 15, 67281)
[ 740630.399] {Default Queue} wl_keyboard#26.repeat_info(25, 600)
[ 740630.407] {Default Queue} wl_keyboard#26.modifiers(66581, 0, 0, 16, 0)
[...]

PR #9106, opened 2026-04-09
Prevent bogus repeated keypresses when reloading config

WhyNotHugo:
When reloading the configuration, each time that a configuration applies to a keyboard, an additional bogus keypress is sent to clients.

This seems to happen due to handle_keyboard_group_leave updating the pressed keys, but handle_keyboard_group_enter not doing the reverse, so the state of pressed keys is inconsistent.

Extend handle_keyboard_group_enter to mirror
handle_keyboard_group_leave, ensure state is consistent and no spurious key events are sent.

### DIFF
--- a/sway/input/keyboard.c
+++ b/sway/input/keyboard.c
@@ -598,6 +598,25 @@ static void handle_keyboard_group_key(struct wl_listener *listener,
 	handle_key_event(sway_group->seat_device->keyboard, data);
 }
 
+static void keyboard_group_refocus(struct sway_keyboard *keyboard) {
+	// Refocus the focused node, layer surface, or unmanaged surface so that
+	// it picks up the current keyboard state.
+	struct sway_seat *seat = keyboard->seat_device->sway_seat;
+	struct sway_node *focus = seat_get_focus(seat);
+	if (focus) {
+		seat_set_focus(seat, NULL);
+		seat_set_focus(seat, focus);
+	} else if (seat->focused_layer) {
+		struct wlr_layer_surface_v1 *layer = seat->focused_layer;
+		seat_set_focus_layer(seat, NULL);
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Actually, I copied this from `handle_keyboard_group_leave` thinking I'd need to invert some booleans or replace some bits, but it ended up being identical. I think I can refactor.

WhyNotHugo:
I think that's tidier. The diff would have been smaller putting keyboard_group_refocus at the end, but that requires an additional forward declaration and didn't seem worth it.

emersion:
I can't reproduce the bug. I run Sway with `wev -f wl_keyboard & sleep 5; swaymsg reload`, then focus the wev window and press some keys. The behavior is the same with or without this patch.

WhyNotHugo:
The easiest reproducer is to just type `swaymsg reload` and press enter. The terminal will receive multiple Enter events.

I have an impression that having more `input` blocks affects this further, I have:

```
input type:keyboard {
  xkb_layout us
  xkb_variant intl
  xkb_numlock enable
}

input "13364:528:Keychron_Keychron_K1_Pro"                  xkb_options caps:escape,compose:ralt
input "13364:528:Keychron_Keychron_K1_Pro_Consumer_Control" xkb_options caps:escape,compose:ralt

# …
```

Can you try an equivalent configuration?

emersion:
Yeah, no dice. Could you find a way to make these "bogus repeated keypresses" visible in a wev or `WAYLAND_DEBUG` log?

WhyNotHugo:
I realise I forgot to link to the issue which includes the WAYLAND_DEBUG output: https://github.com/swaywm/sway/issues/8852

emersion:
Still cannot reproduce. With two foots running, I get key release/press pairs with or without the patch.

PR #9114, opened 2026-04-15
input/seat: end keyboard grab when clearing focus

### LINKED ISSUE #8919: BUG: Context Menu Results in Phantom Focus
- **Sway Version:**
1.11
 
- **Configuration File:**
repro provided, but `focus_follows_mouse no` if it matters 

- **Description:**

 Sway can get into a faulty state, where a window has focus, but keyboard events go to another windows. I've been triggering this by mistake for months now, but only today did I recognize the triggering condition.

Steps to reproduce:
1. open two terminals side by side
2. click left window to give it focus
3. press the mod button to bring up Waybar (hidden by default in my setup), 
4. while holding mod button to keep bar visible, right click on tray icon (flameshot in my case) to bring up its context menu.
5. with context menu visible, click on the right window

Result: the title bar of the right window shows it has received focus, but keyboard events still go to the left window. In my terminal, you can also see that the cursor doesn't change from an empty rectangle to a filled one, as it does when the window is ready to receive keyboard inputs.

Demonstration Video:

Image

Probably related to #8912. 
Possible dupe of #6967, but hopefully I've provided a reproducer.
Possibly related #6555

sahinf:
When focus leaves a surface, ensure any active keyboard grab is terminated. This prevents stale xdg-popup grabs from persisting after the popup is destroyed, which could otherwise cause keyboard input to remain stuck on the old window.

fixes #8919

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -14,7 +14,6 @@
 #include <wlr/types/wlr_tablet_v2.h>
 #include <wlr/types/wlr_touch.h>
 #include <wlr/types/wlr_xcursor_manager.h>
-#include "config.h"
 #include "list.h"
 #include "log.h"
 #include "sway/config.h"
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

emersion:
iioa has reported that this PR breaks drag-and-drop between apps.

sahinf:
oops, new PR to fix that regression inc.

PR #9117, opened 2026-04-17
input/seat: fix drag-and-drop regression and improve popup dismissal

sahinf:
Context: https://github.com/swaywm/sway/pull/9114#issuecomment-4271613916

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1071,6 +1071,7 @@ bool seat_is_input_allowed(struct sway_seat *seat,
 
 static void send_unfocus(struct sway_container *con, void *data) {
 	if (con->view) {
+		view_close_popups(con->view);
 		view_set_activated(con->view, false);
 	}
 }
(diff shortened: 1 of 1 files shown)


sway/input/seat.c
```diff
@@ -1071,6 +1071,7 @@ bool seat_is_input_allowed(struct sway_seat *seat,
 
 static void send_unfocus(struct sway_container *con, void *data) {
 	if (con->view) {
+		view_close_popups(con->view);
```
emersion, line 1074: A bunch of callers were calling `view_close_popups()` manually right before `seat_send_unfocus()`. A reckon we can drop these calls now?

sahinf: Oh nice catch, yeah. I think I see one caller that does that but lmk if otherwise: 
```sh
$ rg 'view_close_popups()' -A 5 | rg 'seat_send_unfocus'
sway/input/seat.c-1143-         seat_send_unfocus(last_focus, seat);
```

sahinf: actually you right, `seat_set_workspace_focus` was also calling it way after unfocus

emersion (approved):
Thank you!

PR #9125, opened 2026-04-23
Treat ext-workspace-v1 as privileged

### LINKED ISSUE #9120: ext-workspace-v1 issue

- **Sway Version:**
  - 1.12-rc2, also reproduces on current master (sway/tree/workspace.c
    and sway/server.c are byte-identical between the two).

- **Debug Log:**
  - Not applicable. This is a protocol-design issue, not a runtime
    crash. Behavior is visible by inspecting which globals are
    advertised to a security-context-v1 sandboxed client.

- **Configuration File:**
  - Default configuration reproduces the core issue.
  - The pango-injection sub-impact (see below) additionally requires
    `font pango:monospace 10` (or any `pango:` font) so that swaybar
    interprets workspace names as pango markup.

- **Stack Trace:**
  - Not applicable. No crash.

- **Description:**

  Sway 1.12 added support for the ext-workspace-v1 Wayland protocol,
  which lets a client create workspaces, activate a workspace, and
  reassign workspaces between outputs.

  Sway's `is_privileged` function in `sway/server.c` lines 98-130
  lists the Wayland globals that are filtered out for clients inside
  a `security_context_v1` sandbox. The `workspace_manager_v1` global
  is not in that list. As a result, any sandboxed Wayland client can
  bind `ext_workspace_manager_v1` and exercise the full protocol.

  This lets a sandboxed client do four things it is normally meant to
  be prevented from doing:

  1. Force-switch the user's active workspace at any time.
     `sway/tree/workspace.c` lines 54-58 call `workspace_switch`
     synchronously on an `ACTIVATE` request with no ownership or
     focus check. Sway already treats unsolicited focus changes as a
     security concern: `sway/xdg_activation_v1.c` line 47 requires
     `had_focused_surface` before an xdg-activation token can move
     focus. `ext-workspace-v1` bypasses that guarantee entirely.

  2. Create workspaces with attacker-chosen names. Lines 59-79 copy
     the client-supplied name straight into `ws->name`. When the
     user has pango markup enabled in swaybar, that name is passed to
     `pango_parse_markup` via `swaybar/render.c` line 430. Up to and
     including 1.11, Wayland clients could not set workspace names,
     so swaybar's workspace-label rendering was safe. 1.12 changes
     that without adjusting the escaping. A client that sets the
     name to `<span foreground="red" size="xx-large">FAKE</span>`
     gets it rendered as real red markup in the workspace buttons.
     Layout or DoS attacks via oversized spans apply too.

  3. Create an unbounded number of workspaces. Lines 59-79 have a
     duplicate-name check but no per-client cap and no total cap. A
     loop creating unique names exhausts memory and floods swaybar.

  4. Reassign workspaces between outputs without user consent. Lines
     80-89.

  Steps to reproduce sub-finding 1 (focus switch) with a minimal
  Wayland client:

  - Start sway, open two workspaces, focus workspace 1.
  - Run a test client that binds `ext_workspace_manager_v1`, finds
    workspace 2's handle, and sends `activate` followed by `commit`.
  - The active workspace jumps to workspace 2, no user input
    required.
  - If the test client is launched under a sandbox that creates a
    `security_context_v1` (e.g. via way-secure or bubblejail with
    security-context support), the attack still succeeds, which
    demonstrates the privilege gap.

  Suggested fix:

  Add `server->workspace_manager_v1->global` to the returns list in
  `is_privileged` in `sway/server.c`. That matches how every other
  protocol that modifies global compositor state is gated
  (`output_manager_v1`, `session_lock.manager`,
  `foreign_toplevel_manager`, etc.).

  If compositor-side workspace creation from sandboxed clients is
  ever wanted in the future, it would also be worth adding a
  per-client workspace-count cap and sanitizing the name against
  pango markup before storing it.

WhyNotHugo:
And do not expose it to clients in security contexts.

Fixes: https://github.com/swaywm/sway/issues/9120

### DIFF
--- a/sway/server.c
+++ b/sway/server.c
@@ -126,7 +126,8 @@ static bool is_privileged(const struct wl_global *global) {
 		global == server.input->virtual_keyboard->global ||
 		global == server.input->virtual_pointer->global ||
 		global == server.input->transient_seat_manager->global ||
-		global == server.xdg_output_manager_v1->global;
+		global == server.xdg_output_manager_v1->global ||
+		global == server.workspace_manager_v1->global;
 }
 
 static bool filter_global(const struct wl_client *client,


emersion (approved):
Thanks!

PR #9135, opened 2026-04-30
enable multi-key modifier with floating_modifier

carpetoctopus:
This commit allows a key combination to be used as the modifier for floating_modifier.

Example:

floating_modifier Shift+Alt normal

will work, whereas before it would throw an "Invalid modifier" error.


carpetoctopus:
It introduces some bugs so I am closing for now.

carpetoctopus:
See #9137 instead, I think it's a better approach.

PR #9142, opened 2026-05-05
swaybar/tray: replace icon lookup with libsfdo

### LINKED ISSUE #8607: swaybar: replace icon lookup with libsfdo

timkicker:
Drops the in-tree icon theme parser and lookup logic in favor of libsfdo-icon, which implements the icon theme spec.

Absolute icon paths are handled in swaybar itself with an access() probe, since libsfdo intentionally doesn't deal with them. Pixmap fallback still kicks in if load_image() fails on the absolute path.

KDE's IconThemePath is honored via sfdo_icon_theme_load_from() with a per-SNI override theme that gets dropped when the property changes.

The min/max size cache per SNI is replaced by re-lookup when the rendered size changes; libsfdo caches the parsed theme internally and reads gtk's icon-theme.cache files, so this is cheap.

Theme switching at runtime is preserved via a new tray_reload_icon_theme() called from parse_bar_config(). The old code worked here implicitly because all themes were preloaded.

Closes: https://github.com/swaywm/sway/issues/8607

First time contributer here, please tell me what to improve :)


timkicker:
gentle bump, would appreciate a review when someone has a moment. happy to adjust anything, especially the IconThemePath handling and the extension stripping i flagged.

PR #9145, opened 2026-05-09
man: update drag_lock default

### LINKED ISSUE #9119: sticky drag_lock no longer works in sway 1.12
- **Sway Version:**
  - `sway version 1.12-rc2`

- **Debug Log:**
  - https://gist.github.com/jpalus/b3085bf7ef1d118517cfd338ccf90b98

- **Configuration File:**
  - Default config + `swaymsg input type:touchpad tap enabled` invoked at runtime

- **Description:**
  - Try to double tap and keeping finger down select some text, then raise finger and try to select more text.
  - Text is no longer selected until next tap like it used to in sway 1.11. doing explicit `input type:touchpad drag_lock enabled_sticky` does not change it.

jpalus:
ref. #8800

Fixes #9119

### DIFF
--- a/sway/sway-input.5.scd
+++ b/sway/sway-input.5.scd
@@ -154,7 +154,7 @@ The following commands may only be used in the configuration file.
 
 *input* <identifier> drag_lock enabled|disabled|enabled_sticky
 	Enables or disables drag lock for specified input device. The default is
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

PR #9147, opened 2026-05-13
sway/input/cursor: implement pointer-warp-v1

llyyr:
Ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5367

### DIFF
--- a/sway/input/cursor.c
+++ b/sway/input/cursor.c
@@ -9,6 +9,7 @@
 #include <wlr/types/wlr_cursor_shape_v1.h>
 #include <wlr/types/wlr_pointer.h>
 #include <wlr/types/wlr_relative_pointer_v1.h>
+#include <wlr/types/wlr_pointer_warp_v1.h>
 #include <wlr/types/wlr_touch.h>
 #include <wlr/types/wlr_tablet_v2.h>
 #include <wlr/types/wlr_tablet_pad.h>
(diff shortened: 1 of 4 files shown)


sway/server.c
```diff
@@ -424,6 +425,14 @@ bool server_init(struct sway_server *server) {
 	wl_signal_add(&server->pointer_constraints->events.new_constraint,
 		&server->pointer_constraint);
 
+	server->pointer_warp_v1 = wlr_pointer_warp_v1_create(server->wl_display, 1);
```
bl4ckb0ne: The version needs to be in a define, see the top of the file right after the includes

bl4ckb0ne (changes requested):

sway/server.c
llyyr: I hadn't added it because only some versions are in define, but sure

sway/input/cursor.c
```diff
@@ -1415,3 +1416,43 @@ void handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
...
 }
+
+void handle_pointer_warp(struct wl_listener *listener, void *data) {
+	const struct wlr_pointer_warp_v1_event_warp *event = data;
+	struct sway_seat *seat = event->seat_client->seat->data;
+	struct sway_cursor *cursor = seat->cursor;
+
+	struct wl_client *focused_client = NULL;
+	struct wlr_surface *focused_surface =
+		cursor->seat->wlr_seat->pointer_state.focused_surface;
```
emersion: Nit: can be simplified to `seat` instead of `cursor->seat`.

sway/input/cursor.c
```diff
@@ -1415,3 +1416,43 @@ void handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
...
+	struct sway_view *view = view_from_wlr_surface(event->surface);
+	if (view == NULL) {
+		return;
+	}
+
+	struct sway_container *con = view->container;
+	double lx = event->x + con->pending.content_x - view->geometry.x;
+	double ly = event->y + con->pending.content_y - view->geometry.y;
+	wlr_cursor_warp(cursor->cursor, NULL, lx, ly);
+	wlr_seat_pointer_warp(event->seat_client->seat, event->x, event->y);
```
emersion: Do we need to send a pointer frame event here? Or is this unnecessary, or handled elsewhere?

sway/input/cursor.c
```diff
@@ -1415,3 +1416,43 @@ void handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
...
+		.width = event->surface->current.width,
+		.height = event->surface->current.height,
+	};
+
+	if (!wlr_box_contains_point(&surface_box, event->x, event->y)) {
+		sway_log(SWAY_DEBUG, "denying request to warp cursor out of surface");
+		return;
+	}
+
+	struct sway_view *view = view_from_wlr_surface(event->surface);
```
emersion, line 1447: This doesn't work for surfaces which aren't views (e.g. layer-shell, ext-session-lock). For xdg_popup, this doesn't account for the offset correctly.

Ideally we'd walk up the surface tree and accumulate delta coordinates. This could be left as a TODO for now.

sway/input/cursor.c
```diff
@@ -1415,3 +1416,43 @@ void handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
...
+	if (view == NULL) {
+		return;
```
emersion, line 1450: Can we log a message when this happens?

sway/input/cursor.c
```diff
@@ -1415,3 +1416,43 @@ void handle_request_set_cursor_shape(struct wl_listener *listener, void *data) {
...
+	}
+
+	struct sway_container *con = view->container;
+	double lx = event->x + con->pending.content_x - view->geometry.x;
+	double ly = event->y + con->pending.content_y - view->geometry.y;
```
emersion, line 1455: Should we use pending or current here?

sway/input/cursor.c
llyyr: I was mindlessly copying the `zwp_locked_pointer_v1` implementation for pointer warp. I think that implementation does `wlr_seat_pointer_warp` to not send a motion event to client, since it's inhibited by the spec. The spec for pointer-warp doesn't say the motion event should inhibited though, so I don't think we should do `wlr_seat_pointer_warp`

sway/input/cursor.c
llyyr: The same applies for pointer-constraints too

sway/input/cursor.c
llyyr: I think pending state is right here, as with zwp_locked_pointer_v1

PR #9150, opened 2026-05-16
input/seatop_default: fix focus on layer surface output switch

smlx:
When `focus_follows_mouse` is enabled and the cursor moves from one output to a layer surface (such as swaybar) on another output, Sway would previously focus the workspace node.

This updates `check_focus_follows_mouse` to use
`seat_get_focus_inactive()`, ensuring that the previously focused window on that workspace receives focus instead.

---

The bug this fixes is visible if you run your mouse pointer across the top of a two-output setup with swaybar at the opt of both outputs.

Reviewer note: this is the first time I have touched sway code in a long time so please let me know if something is amiss in my commit message or understanding of the bug and its fix.


smlx:
this has some issues. closing to do more testing.

PR #9153, opened 2026-05-18
Preserve IME ordering: forwarded key presses create an outstanding re…

daym:
…sponse count; text-input state/done snapshots wait behind those keys; later keys wait behind queued state.

Previously, Sway was letting zwp_input_method_v2.done advance the IM serial while a key already forwarded through zwp_input_method_keyboard_grab_v2.key could still answer with a commit using the older serial.  wlroots then (correctly) rejected that commit per protocol.

I had had programs lose some (like 10%) letters I typed when I typed fast. I had suspected my fingers, my keyboard hardware etcetc. But no, it was just an entirely predictable problem with message handling.

This fixes the problem by making Sway preserve the ordering between input-method keyboard grab keys and input-method done events. Forwarded key presses create an outstanding response count. Text-input state updates are snapshotted and queued behind outstanding key presses. If another key arrives while state is queued, the key is queued behind that state, so sequences such as K1, S1, K2, S2 are delivered to the input method as K1, then S1, then K2, then S2. A key response is considered observed when Sway receives either an input_method.commit or a same-client virtual-keyboard key press, which is the observable path used when the input method forwards an unhandled key instead of committing text.

See bug report https://github.com/swaywm/sway/issues/9154 , which this fixes.

### DIFF
--- a/include/sway/input/text_input.h
+++ b/include/sway/input/text_input.h
@@ -1,9 +1,11 @@
 #ifndef _SWAY_INPUT_TEXT_INPUT_H
 #define _SWAY_INPUT_TEXT_INPUT_H
 
+#include <stdint.h>
 #include <wlr/types/wlr_text_input_v3.h>
 #include <wlr/types/wlr_input_method_v2.h>
 #include <wlr/types/wlr_compositor.h>
+#include <wlr/types/wlr_virtual_keyboard_v1.h>
 
 /**
  * The relay structure manages the relationship between text-input and
@@ -22,7 +24,10 @@ struct sway_input_method_relay {
 
 	struct wl_list text_inputs; // sway_text_input::link
 	struct wl_list input_popups; // sway_input_popup::link
+	struct wl_list pending_im_events; // sway_pending_im_event::link
 	struct wlr_input_method_v2 *input_method; // doesn't have to be present
+	uint32_t pending_key_responses;
+	uint32_t logged_unmatched_key_responses;
 
 	struct wl_listener text_input_new;
 	struct wl_listener text_input_manager_destroy;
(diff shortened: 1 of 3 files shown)


daym:
Better to remove the following check (3bf9000a, "input-method-v2: validate commit serial", 2023-12-18, for wlroots#3782) from wlroots (types/wlr_input_method_v2.c, im_commit):

```c
    if (serial != input_method->current_serial) {
        input_state_reset(&input_method->pending);
        return;
    }
```

The check was introduced to fix wlroots#3782: While composing preedit in one text input and then switching focus to another, fcitx5 sends commit_text and commit carrying the serial of the deactivating transaction; if activation bumps the serial before those requests are processed, wlroots applies the commit after activation and the committed text lands in the newly focused input. The check drops those commits.

But the check can't distinguish "stale because the focused text input changed" from "stale because the focused client committed its text-input state while the input method was still producing this commit". The latter happens on essentially every keystroke, so commits for keys the user actually typed get dropped. Anything sway does to work around that--this PR's outstanding-response counter--is guessing, because zwp_input_method_keyboard_grab_v2.key has no ack.

Even better: support the new xx-keyboard-filter-v1 (merged to wayland-protocols' experimental/ in Dec 2025), which allows the input method to ack/reply per key: the compositor queues each intercepted key and the IM answers filter(serial, consume | passthrough), so the compositor always knows every key's fate. That makes serializing done against forwarded keys unnecessary, and it removes the zwp_virtual_keyboard_v1 re-injection hack--xx-input-method-v2 in the same series drops grab_keyboard entirely.

Even better if you have wl_keyboard v10 (wl_seat v10), because then repeats are supported too: v10 adds key_state.repeated and compositor-side key repeat (repeat_info rate 0), and the filter protocol notes it "works much better with compositor-side key repeat introduced in wl_seat version 10" --only wl_keyboard.key is filterable as of v10. wlroots still advertises SEAT_VERSION 9.

So I'm withdrawing this PR: the fix belongs in wlroots (remove the check) and ultimately in the new protocol.

PR #9156, opened 2026-05-22
seatop_move_tiling: fix stack overflow when dropping container into itself

### LINKED ISSUE #7949: Stack overflow when moving a container into itself
- **Sway Version:**: `sway version 1.10-dev-ba427a46 (Feb  2 2024, branch 'master')`
- **Debug Log:** [sway.log](https://github.com/swaywm/sway/files/14145819/sway.log)
- **Configuration File:** Default configuration
- **Stack Trace:** [gdb.txt](https://github.com/swaywm/sway/files/14145899/gdb.txt) (truncated)
- **Description:**
  - Change layout to tabbed
  - Create a window
  - Split horizontally
  - Create a window
  - Split horizontally
  - Tree should be `H[window H[window]]`
  - Using the mouse, move the outer container into the innermost container

smlx:
This change avoids a cyclic reference in the window tree when dragging a container and dropping it into one of its descendants.

The drop target resolution logic previously only checked if the target node was the container itself, but missed checking if the target was a descendant of the dragged container.

Fixes: #7949

### DIFF
--- a/sway/input/seatop_move_tiling.c
+++ b/sway/input/seatop_move_tiling.c
@@ -252,8 +252,9 @@ static void handle_motion_postthreshold(struct sway_seat *seat) {
 		}
 		if (edge) {
 			e->target_node = node_get_parent(&con->node);
-			if (e->target_node == &e->con->node) {
-				e->target_node = node_get_parent(e->target_node);
+			if (e->target_node && (e->target_node == &e->con->node ||
+					node_has_ancestor(e->target_node, &e->con->node))) {
+				e->target_node = node_get_parent(&e->con->node);
 			}
 			e->target_edge = edge;
 			update_indicator(e, &box);


sway/input/seatop_move_tiling.c
```diff
@@ -252,7 +252,8 @@ static void handle_motion_postthreshold(struct sway_seat *seat) {
 		}
 		if (edge) {
 			e->target_node = node_get_parent(&con->node);
-			if (e->target_node == &e->con->node) {
+			while (e->target_node && (e->target_node == &e->con->node ||
+					node_has_ancestor(e->target_node, &e->con->node))) {
 				e->target_node = node_get_parent(e->target_node);
```
emersion: Do we need a loop here?

Wouldn't the result always be `node_get_parent(&e->con->node)`?

smlx: Ah yes of course you are correct. I have pushed a change to drop the loop.

emersion (approved):
Thanks!

PR #9157, opened 2026-05-23
ipc: fix workspace `focused`/`visible` reported outside get_workspaces

orestisfl:
`ipc_json_describe_workspace` now sets `focused` and adds `visible` itself. The default `focused` from `ipc_json_create_node` compares the focused node to the workspace node, which misses the common case where a window of that workspace is focused.

As a result, `workspace` IPC events reported `focused: false` for every workspace and were missing `visible` entirely, even though `get_workspaces` overrode them.

### DIFF
--- a/sway/ipc-json.c
+++ b/sway/ipc-json.c
@@ -501,6 +501,14 @@ static json_object *ipc_json_describe_scratchpad_output(void) {
 
 static void ipc_json_describe_workspace(struct sway_workspace *workspace,
 		json_object *object) {
+	// Override the default focused indicator because it's set differently
+	struct sway_seat *seat = input_manager_get_default_seat();
+	json_object_object_del(object, "focused");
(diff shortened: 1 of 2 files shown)


emersion (commented):
This doesn't seem to be compatible with i3. In my testing, i3 returns `"focused": false` for workspaces with a focused window in the `get_tree` reply.

orestisfl:
> This doesn't seem to be compatible with i3. In my testing, i3 returns `"focused": false` for workspaces with a focused window in the `get_tree` reply.

Good catch. Not a good idea to upstream IMO, I thought there was a mismatch here. If we "fix" this, it might actually break IPC clients' "find focused container" logic.

The actual mismatch seems to be that i3 sends a workspace focus event on top of the move event. I'll close this PR and investigate a follow-up.

PR #9159, opened 2026-05-23
i3-compat: swaybar: default to first bar-id

orestisfl:
When launching swaybar from the command line without a --bar-id flag, default to the first configured bar.

This is a mirror of i3 PR https://github.com/i3/i3/pull/4231.

This makes it easier to call swaybar from the command line since the majority of swaybar users have only one bar configured.


emersion (commented):
LGTM, thanks!

PR #9160, opened 2026-05-25
input/seatop_default: do not focus workspace if clicking a layer surface

smlx:
Previously, the pointer handler in `seatop_default` would focus the workspace, regardless of whether a surface intercepted the click.

This commit adds a check for `surface == NULL`, allowing the click event to fall through to the layer surface handler.

### DIFF
--- a/sway/input/seatop_default.c
+++ b/sway/input/seatop_default.c
@@ -370,7 +370,7 @@ static void handle_button(struct sway_seat *seat, uint32_t time_msec,
 	}
 
 	// Handle clicking an empty workspace
-	if (node && node->type == N_WORKSPACE) {
+	if (node && node->type == N_WORKSPACE && surface == NULL) {
 		if (state == WL_POINTER_BUTTON_STATE_PRESSED) {
 			seat_set_focus(seat, node);
 			transaction_commit_dirty();


emersion:
Can the "fix()" part be dropped from the PR and commit title? We don't use Conventional Commits.

emersion (commented):
How does one reproduce the bug?

When clicking on swaybar, `node == NULL`.

smlx:
Sorry it wasn't so clear.

I'm working on fixing up the swaybar DBus menu PR and noticed a few focus bugs related to layer shell and popups. Before this change, clicking on the menu doesn't work and instead focuses the underlying workspace:

[layer.bug.webm](https://github.com/user-attachments/assets/53ceb3e3-5d0c-40fc-85ef-ce30d9b3c045)

After this change, clicking the menu actually works:

[layer.working.webm](https://github.com/user-attachments/assets/6c29f4fc-46b3-40c6-8c51-a927f6d56ff3)

smlx:
Gentle ping on this one.

> How does one reproduce the bug?

One easy way is:

1. Tile several windows on an output.
2. Position mouse cursor over one tiled window and click to focus it if necessary.
3. Do not move the mouse. Lock the screen with swaylock.
4. Click the left mouse button.
5. Unlock the screen.
6. Notice that the workspace node is focused.

emersion (approved):
LGTM, thanks!

PR #9162, opened 2026-05-27
input/seat: retain focus on sticky windows during workspace switch

### LINKED ISSUE #5089: Sticky windows lose focus when focusing another workspace
sway version 1.4

When changing the focused workspace, a focused sticky window will lose focus, even if that workspace is empty. A focused sticky window shouldn't lose focus when focusing different workspaces (this is the behavior of i3).

smlx:
I tried to add this functionality in the most minimally invasive way possible, but I'm not that familiar with the code so hopefully this makes sense. It does seem to fix the issue.

Fixes: #5089

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1127,6 +1127,28 @@ void seat_set_raw_focus(struct sway_seat *seat, struct sway_node *node) {
 	}
 }
 
+// Rebuild the focus stack for a given container by pushing its ancestors
+// and then the container itself to the top of the stack.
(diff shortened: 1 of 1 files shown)


smlx:
I realised the previous approach to fix this wasn't great because it resulted in focus flicker during the switch. So the new approach tries to avoid removing focus unnecessarily from a focused sticky window.

smlx:
I've reworked this patch locally and combined it with some other fixes. Will close this and send a new PR.

PR #9163, opened 2026-05-27
Fix sticky window focus and decorations when switching workspace

### LINKED ISSUE #5089: Sticky windows lose focus when focusing another workspace

### LINKED ISSUE #8924: fake focus on floating sticky windows in empty workspace

- **Sway Version:**
  - `swaymsg -t get_version` or `sway -v`
  - 1.11, but was always present

- **Description:**
  - create a couple of floating sticky windows
  - switch to an empty workspace
  - observe all floating windows are marked as focused, yet they are actually not since no keyboard inputs or shortcuts interact with them
  - the real focus is on the empty tiling workspace

This happens even if one of the sticky windows was focused before switching the workspace -> the window loses focus, although it is still in the exact same spot on the screen.

`swaymsg -t get_tree` reports that only the workspace is focused, no windows inside it

smlx:
See individual commits.

Fixes: #8924
Fixes: #5089

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1127,6 +1127,16 @@ void seat_set_raw_focus(struct sway_seat *seat, struct sway_node *node) {
 	}
 }
 
+// Rebuild the focus stack for a given container by pushing its ancestors
+// and then the container itself to the top of the stack.
+static void seat_rebuild_focus_stack(struct sway_seat *seat, struct sway_container *con) {
+	if (!con) {
+		return;
+	}
+	seat_rebuild_focus_stack(seat, con->pending.parent);
+	seat_set_raw_focus(seat, &con->node);
+}
+
 static void seat_set_workspace_focus(struct sway_seat *seat, struct sway_node *node) {
 	struct sway_node *last_focus = seat_get_focus(seat);
 	if (last_focus == node) {
@@ -1164,8 +1174,14 @@ static void seat_set_workspace_focus(struct sway_seat *seat, struct sway_node *n
 		node_set_dirty(&new_output->node);
 	}
 
+	struct sway_container *last_focus_con = last_focus && last_focus->type == N_CONTAINER ?
+		last_focus->sway_container : NULL;
+	bool sticky_has_focus_on_switch = last_focus_con && container_is_sticky_or_child(last_focus_con) &&
+		new_workspace && last_workspace && new_workspace != last_workspace &&
+		last_focus_con->pending.workspace == last_workspace;
+
 	// Unfocus the previous focus
-	if (last_focus) {
+	if (last_focus && !sticky_has_focus_on_switch) {
 		seat_send_unfocus(last_focus, seat);
 		node_set_dirty(last_focus);
 		struct sway_node *parent = node_get_parent(last_focus);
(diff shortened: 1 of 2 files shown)


smlx:
To reproduce the two issues this PR fixes:

1. Open two floating windows on a workspace. Make them sticky, and focus one of them.
2. Switch to another workspace.

On master: Both floating sticky windows will have blue active focus decorations (this is #8924) and the previously focused window will lose its focus (this is #5089).

With this patch: The focus state and decorations of neither window changes.

sway/input/seat.c
```diff
@@ -1127,6 +1127,29 @@ void seat_set_raw_focus(struct sway_seat *seat, struct sway_node *node) {
...
+	while (parent) {
+		list_add(ancestors, parent);
+		parent = parent->pending.parent;
+	}
+	// push ancestors onto the focus stack from the top down
+	for (int i = ancestors->length - 1; i >= 0; --i) {
+		struct sway_container *ancestor = ancestors->items[i];
+		seat_set_raw_focus(seat, &ancestor->node);
+	}
+	list_free(ancestors);
```
llyyr: We could just recurse here to achieve the order reversal, instead of allocating a list

sway/input/seat.c
llyyr: Shouldn't we only send ipc event if `!sticky_has_focus_on_switch`? Same for urgency timer below

sway/input/seat.c
```diff
@@ -1164,8 +1187,13 @@ static void seat_set_workspace_focus(struct sway_seat *seat, struct sway_node *n
 		node_set_dirty(&new_output->node);
 	}
 
+	struct sway_container *last_focus_con = last_focus && last_focus->type == N_CONTAINER ?
+		last_focus->sway_container : NULL;
+	bool sticky_has_focus_on_switch = last_focus_con && container_is_sticky_or_child(last_focus_con) &&
+		new_workspace && last_workspace && new_workspace != last_workspace;
```
llyyr: last_workspace is the previously active workspace on an output, not the last workspace containing focused container. 

Needs to check `last_focus_con->pending.workspace == last_workspace` as well.

smlx: Ah yes of course. Thank you!

sway/input/seat.c
smlx: Yes, absolutely.

sway/input/seat.c
smlx: I agree this would be simpler. My concern was that this opens the door to stack overflow on extreme tree depth edge cases, which is why I went for the list approach.

From what I can tell there is not limit on how deep the container tree can be in Sway?

smlx:
Updated with some correctness fixes based on code review comments.

sway/input/seat.c
llyyr: Those risks already exist elsewhere in the codebase, see container_find_child and container_for_each_child. If this is a concern it would require setting an explicit limit

smlx: Yeah okay fair point. I also checked i3 and there are similar recursive tree walks in that codebase. I'll rework this to use recursion.

smlx: Fixed, thanks!

llyyr (approved):
Both commits seem fine to me now

smlx:
gentle ping @emersion

PR #9166, opened 2026-05-28
docs: mark opacity as runtime-only command

### LINKED ISSUE #9100: docs: `opacity` command is in the wrong category in sway(5)
- **Sway Version:**
  - sway version 1.11

- **Description:**
  - The sway(5) man page says the `opacity` command is both available in config and runtime, but it only works in runtime. Trying to use it in config will yield `Unknown/invalid command 'opacity'` (unless it is behind a `for_window <criteria>` of course)

frapank:
`opacity` is registered in `command_handlers[]` and requires a container in
`handler_context`. Using it directly in the config file results in
"Unknown/invalid command 'opacity'". It must be used via `swaymsg`,
`bindsym`, or `for_window`.

Fixes #9100

### DIFF
--- a/sway/sway.5.scd
+++ b/sway/sway.5.scd
@@ -294,6 +294,10 @@ set|plus|minus|toggle <amount>
 	A no operation command that can be used to override default behaviour. The
 	optional comment argument is ignored, but logged for debugging purposes.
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

PR #9171, opened 2026-05-30
sway/server: advertise xdg-decoration-v1 version 2

llyyr:
Ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5385


emersion (approved):
Thanks!

PR #9172, opened 2026-05-31
input/seat: prevent layer surfaces from focusing under session lock

smlx:
When the screen is locked, `server.session_lock.lock` is active and
keyboard input must be forced to the screen lock surface. While
`seat_set_focus` checked for an active session lock and redirected
focus, `seat_set_focus_layer` was missing the same check. This allowed
layer surfaces to steal keyboard input while the screen lock was active.

This commit adds the missing session check to `seat_set_focus_layer`.

The bug this fixes can be reproduced on master by:

1. run `sleep 1 && fuzzel & swaylock`
2. type the name of some app e.g. `transmission` and hit enter. Notice that swaylock doesn't respond
3. type your password to unlock swaylock
4. upon unlock, notice the app has opened under the lockscreen

### DIFF
--- a/sway/input/seat.c
+++ b/sway/input/seat.c
@@ -1312,6 +1312,9 @@ void seat_set_focus_layer(struct sway_seat *seat,
 	} else if (!layer) {
 		return;
 	}
+	if (server.session_lock.lock) {
+		return;
+	}
 	assert(layer->surface->mapped);
 	if (layer->current.layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP &&
 			layer->current.keyboard_interactive


smlx:
I changed my approach with this patch to be much simpler and avoid the momentary focus flicker.

sway/input/seat.c
```diff
@@ -1321,6 +1321,9 @@ void seat_set_focus_layer(struct sway_seat *seat,
 	if (seat->focused_layer == layer) {
 		return;
 	}
+	if (server.session_lock.lock) {
```
emersion: Should we move this check up before `seat->has_exclusive_layer = true`?

It seems like e.g. `seat_set_focus()` assumes a layer surface has focus when that field is set.

smlx: Yes I think so. I'll move it up.

But looking at the `has_exclusive_layer` field it seems to be kind of redundant? It only seems to be checked in `seat_set_focus` and couldn't that check be replaced with something like this?

```C
bool has_exclusive_layer = seat->focused_layer &&
	seat->focused_layer->current.layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP &&
	seat->focused_layer->current.keyboard_interactive == ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE;

if (has_exclusive_layer) {
// ...
```

emersion (approved):
LGTM, thanks!

sway/input/seat.c
emersion: Yeah, I completely agree! Please feel free to submit a patch :)

PR #9175, opened 2026-06-02
swaybar/tray: fix heap buffer overflow in read_pixmap

### LINKED ISSUE #9174: Heap out-of-bounds write in swaybar `IconPixmap` parsing
**Sway Version:** `sway version 1.12`

**Debug Log:**

Not applicable. This is a memory-safety bug triggered by malformed D-Bus data. A debug log would not reveal it; an AddressSanitizer trace is provided under Stack Trace.

**Configuration File:**

Default configuration. The bug is in swaybar's tray/IconPixmap parsing code, independent of user configuration.

**Stack Trace:**

```
==2213==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7ba9b7be3500
READ of size 4 at 0x7ba9b7be3500 thread T0
    #0 0x555de3471ec8 in read_pixmap ../swaybar/tray/item.c:84
    #1 0x555de3472be3 in get_property_callback ../swaybar/tray/item.c:139
    #2 0x7f29ba30d534  (/usr/lib/libsystemd.so.0+0x86534)
    #3 0x555de347b35b in tray_in ../swaybar/tray/tray.c:103
    #4 0x555de3482e7b in loop_poll ../common/loop.c:83
    #5 0x555de34342d6 in bar_run ../swaybar/bar.c:519
    #6 0x555de3450bcd in main ../swaybar/main.c:95

0x7ba9b7be3500 is located 0 bytes after 96-byte region [0x7ba9b7be34a0,0x7ba9b7be3500)

SUMMARY: AddressSanitizer: heap-buffer-overflow ../swaybar/tray/item.c:84 in read_pixmap
```

This trace is from an ASan-instrumented swaybar binary. A malicious D-Bus StatusNotifierItem was registered with `width=32, height=32` but only 4 bytes of pixel data. When swaybar queried the `IconPixmap` property, `read_pixmap()` at line 84 performed an out-of-bounds read/write, crashing under ASan.

**Description:**

`swaybar` does not validate that the byte length of a StatusNotifierItem `IconPixmap` entry matches `width * height * sizeof(uint32_t)` before copying pixel data.

In `read_pixmap()` (`swaybar/tray/item.c`), the byte length from `sd_bus_message_read_array()` sizes the heap allocation, but the copy loop writes `width * height` uint32 values:

```c
// allocation uses npixels (attacker-controlled byte count)
struct swaybar_pixmap *pixmap = malloc(sizeof(struct swaybar_pixmap) + npixels);
pixmap->size = height;

// copy loop uses width*height (= width*height*4 bytes)
for (int i = 0; i < height * width; ++i) {
    ((uint32_t *)pixmap->pixels)[i] = ntohl(((uint32_t *)pixels)[i]);
}
```

When `npixels < width * height * 4`, this produces a heap out-of-bounds write. A malicious local process on the same D-Bus session bus can register a StatusNotifierItem with e.g. `width=32, height=32` and a 4-byte pixel array — swaybar allocates 4 bytes but writes 4096.

### Steps to reproduce

1. Build and run the D-Bus SNI PoC (source in attached `sway-iconpixmap-oob-proof.tar.gz`):
   ```bash
   gcc -g -o poc_sway_dbus_sni poc_sway_dbus_sni.c $(pkg-config --cflags --libs libsystemd)
   ./poc_sway_dbus_sni
   ```
2. The PoC registers a malicious StatusNotifierItem on the D-Bus session bus with `width=32, height=32` but only 4 bytes of pixel data.
3. When swaybar queries the `IconPixmap` property, `read_pixmap()` triggers the heap-buffer-overflow.
4. Under ASan, swaybar crashes: `heap-buffer-overflow in read_pixmap`.

### Attack surface

- Any local process on the same user's D-Bus session bus can expose a malformed `IconPixmap`
- No special privileges required beyond D-Bus session bus access
- Same-user local attack surface only

### Impact

Demonstrated impact:
- Reliable swaybar crash under AddressSanitizer
- Heap out-of-bounds write / adjacent-buffer corruption

Suggested severity: **Medium** (CVSS 3.1: AV:L/AC:L/PR:L/UI:N/S:U/C:N/I:N/A:H → 5.5)

### Suggested fix

Validate `npixels == (size_t)width * height * sizeof(uint32_t)` before allocation/copy, with overflow-safe arithmetic:

```c
if (height > 0 && width == height) {
    size_t w = (size_t)width;
    size_t h = (size_t)height;
    bool valid = true;
    size_t pixel_count = 0;
    size_t expected = 0;

    if (w > SIZE_MAX / h) {
        valid = false;
    } else {
        pixel_count = w * h;
        if (pixel_count > SIZE_MAX / sizeof(uint32_t)) {
            valid = false;
        } else {
            expected = pixel_count * sizeof(uint32_t);
            if (npixels != expected) {
                valid = false;
            }
        }
    }

    if (!valid) {
        sway_log(SWAY_DEBUG,
                "%s %s: discard malformed icon w:%d h:%d bytes:%zu expected:%zu",
                sni->watcher_id, prop, width, height, npixels, expected);
    } else {
        struct swaybar_pixmap *pixmap =
            malloc(sizeof(struct swaybar_pixmap) + expected);
        if (!pixmap) {
            ret = -ENOMEM;
            goto error;
        }
        pixmap->size = height;
        for (size_t i = 0; i < pixel_count; ++i) {
            uint32_t argb;
            memcpy(&argb,
                    (const unsigned char *)pixels + i * sizeof(argb),
                    sizeof(argb));
            ((uint32_t *)pixmap->pixels)[i] = ntohl(argb);
        }
        list_add(pixmaps, pixmap);
    }
}
```

Key changes: (1) validate `npixels == expected` before use, (2) overflow-safe `size_t` arithmetic, (3) `memcpy` instead of direct cast to avoid alignment UB, (4) check `malloc()` return.

### Regression test cases to consider

- 32×32 pixmap with 4096 bytes: accepted
- 32×32 pixmap with 4 bytes: rejected
- 32×32 pixmap with 4092 or 4100 bytes: rejected
- width ≤ 0 or height ≤ 0: skipped safely
- width ≠ height: skipped (existing square-only behavior)
- very large width/height causing multiplication overflow: rejected

### Attached files

`sway-iconpixmap-oob-proof.tar.gz` (attached to this issue) contains:
- `poc_sway_dbus_sni.c` — D-Bus StatusNotifierItem PoC
- `asan-swaybar-crash.log` — full ASan trace from real swaybar crash
- `evidence.txt` — key excerpt from the ASan trace
- `source-excerpt.txt` — vulnerable code excerpt

[sway-iconpixmap-oob-proof.tar.gz](https://github.com/user-attachments/files/28485155/sway-iconpixmap-oob-proof.tar.gz)

smlx:
Validate that the icon pixel data array size accommodates the provided
dimensions.

Fixes: #9174

### DIFF
--- a/swaybar/tray/item.c
+++ b/swaybar/tray/item.c
@@ -66,17 +66,17 @@ static int read_pixmap(sd_bus_message *msg, struct swaybar_sni *sni,
 		}
 
 		const void *pixels;
-		size_t npixels;
-		ret = sd_bus_message_read_array(msg, 'y', &pixels, &npixels);
+		size_t pixel_data_size; // size in bytes, each pixel is 4 bytes
+		ret = sd_bus_message_read_array(msg, 'y', &pixels, &pixel_data_size);
 		if (ret < 0) {
 			sway_log(SWAY_ERROR, "%s %s: %s", sni->watcher_id, prop, strerror(-ret));
 			goto error;
 		}
 
-		if (height > 0 && width == height) {
+		if (height > 0 && width == height && (size_t)width * height <= pixel_data_size / 4) {
 			sway_log(SWAY_DEBUG, "%s %s: found icon w:%d h:%d", sni->watcher_id, prop, width, height);
 			struct swaybar_pixmap *pixmap =
-				malloc(sizeof(struct swaybar_pixmap) + npixels);
+				malloc(sizeof(struct swaybar_pixmap) + pixel_data_size);
 			pixmap->size = height;
 
 			// convert from network byte order to host byte order


swaybar/tray/item.c
emersion, line 84: Completely orthogonal to this PR: it seems this line assumes `pixels` is aligned to 4 bytes, but it doesn't seem like anything provides that guarantee.

emersion (approved):
LGTM, thanks!

swaybar/tray/item.c
smlx: Hmm. I think you are right.

I think `pixmap->pixels` should be aligned though? So maybe copying the pixel data there and then doing the byte order conversion in place would be safe?

emersion: Looking into this some more, this is probably a mostly theoretical bug.

The `malloc(3p)` man page reads:

> The pointer returned if the allocation succeeds shall be suitably aligned so that it may be assigned to a pointer to any type of object

`pixmap->pixels` is a flexible array with no alignment constraint (`unsigned char`) that comes after an `int`. On systems where an int is smaller than 32 bits, `pixels` will not be aligned to 32 bits.

I generally just `memcpy()` when I need to read/write a `uint32_t` from/to a `uint8_t *` buffer.

